PR #8350, opened 2022-10-22
core: `Drawing` cleanup

relrelb:
Extract `MovieClip::drawing` and `Graphic::drawing`, which return a non-`Option` `Drawing`. This avoids many `.unwrap()`s in AVM1.

### DIFF
--- a/core/src/avm1/globals/movie_clip.rs
+++ b/core/src/avm1/globals/movie_clip.rs
@@ -849,66 +849,53 @@ pub fn duplicate_movie_clip_with_bias<'gc>(
         return Ok(Value::Undefined);
     }
 
-    let id = movie_clip.id();
     let movie = parent
         .movie()
         .or_else(|| activation.base_clip().movie())
         .unwrap_or_else(|| activation.context.swf.clone());
     let new_clip = if movie_clip.id() != 0 {
         // Clip from SWF; instantiate a new copy.
-        activation
-            .context
-            .library
-            .library_for_movie(movie)
-            .ok_or("Movie is missing!")
-            .and_then(|l| l.instantiate_by_id(id, activation.context.gc_context))
-    } else {
-        // Dynamically created clip; create a new empty movie clip.
-        Ok(MovieClip::new(movie, activation.context.gc_context).into())
-    };
-
-    if let Ok(new_clip) = new_clip {
-        // Set name and attach to parent.
-        new_clip.set_name(activation.context.gc_context, new_instance_name);
-        parent.replace_at_depth(&mut activation.context, new_clip, depth);
-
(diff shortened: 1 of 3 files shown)


core/src/display_object/graphic.rs
```diff
@@ -93,6 +93,12 @@ impl<'gc> Graphic<'gc> {
             },
         ))
     }
+
+    pub fn drawing(&self, gc_context: MutationContext<'gc, '_>) -> RefMut<'_, Drawing> {
+        RefMut::map(self.0.write(gc_context), |w| {
+            w.drawing.get_or_insert_with(Drawing::new)
```
adrian17, line 99: side observation: I feel like we should make Drawing either always-eagerly-constructed (I think there is only one place where we _don't_ construct it eagerly currently), or always-lazy (and, in fact, hidden behind an `Option<Box>`), not a mix. Given how relatively rare drawing ops are and Drawing takes 200+B per DO, the latter sounds

core/src/avm1/globals/movie_clip.rs
```diff
@@ -849,66 +833,51 @@ pub fn duplicate_movie_clip_with_bias<'gc>(
...
         .unwrap_or_else(|| activation.context.swf.clone());
     let new_clip = if movie_clip.id() != 0 {
         // Clip from SWF; instantiate a new copy.
-        activation
-            .context
-            .library
-            .library_for_movie(movie)
-            .ok_or("Movie is missing!")
-            .and_then(|l| l.instantiate_by_id(id, activation.context.gc_context))
+        let library = activation.context.library.library_for_movie(movie).unwrap();
```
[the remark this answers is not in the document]
relrelb: I'm pretty sure this `library_for_movie()` call cannot fail, hence the `.unwrap()`.

core/src/display_object/graphic.rs
relrelb: Agree, seems like currently `Drawing` is eagerly-initialized only in `Graphic::new_with_avm2()`, which is called from AVM2's `flash.display.Shape` constructor. Do you think changing to `Option<Box<Drawing>>` should be part of this PR?

adrian17: > Agree, seems like currently Drawing is eagerly-initialized only in Graphic::new_with_avm2(), which is called from AVM2's flash.display.Shape constructor. Do you think changing to Option<Box<Drawing>> should be part of this PR?

Up to you. I feel like it'd touch all the same spots in code, so might as well do it here?

relrelb: I won't have much time this week, so I prefer doing this in a separate PR.

PR #8351, opened 2022-10-22
render: Remove `RenderBackend::register_glyph_shape`

relrelb:
Use just `RenderBackend::register_shape` instead.

### DIFF
--- a/core/src/font.rs
+++ b/core/src/font.rs
@@ -2,6 +2,7 @@ use crate::html::TextSpan;
 use crate::prelude::*;
 use crate::string::WStr;
 use gc_arena::{Collect, Gc, MutationContext};
+use ruffle_render::backend::null::NullBitmapSource;
 use ruffle_render::backend::{RenderBackend, ShapeHandle};
 use ruffle_render::transform::Transform;
 use std::cell::{Ref, RefCell};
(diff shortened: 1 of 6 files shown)


adrian17:
> Also remove eager-loading of ASCII glyphs.

FYI I stand with the opinion that this is okay as-is.

core/src/font.rs
```diff
@@ -405,38 +409,35 @@ impl<'gc> Font<'gc> {
...
-                &self.swf_glyph,
-            ));
-        }
-        drop(write);
-        let read = self.shape.borrow();
-        Ref::map(read, |s| s.as_ref().unwrap())
+    pub fn shape_handle(&self, renderer: &mut dyn RenderBackend) -> ShapeHandle {
+        self.shape_handle.get().unwrap_or_else(|| {
+            let shape_handle =
+                renderer.register_shape((&*self.as_shape()).into(), &NullBitmapSource);
```
adrian17, line 437: just for the record, copied from Discord:
> definitely don't want shape_handle to call as_shape
> check out the comment at shape field
> not sure if that part of impl Glyph refactor is needed at all

I think we settled on idea that maybe another refactor to consume the swf::Glyph would avoid doubling memory usage.

relrelb: Just rebased this PR on top of current master. Can we merge it as-is and implement said refactor in a separate PR?

Issue #7419, opened 2022-07-15
Flash 3DX: Loading bar width does not increase


n0samu:
### Describe the bug

In the first frame of [Flash 3DX](https://www.lilgames.com/games/puzzle/flash-3dx/), there is a loading bar that is supposed to increase in width each time the frame runs: 
`barre._width += 1;`
The next frame checks the width of the loading bar to decide whether to start the game:
```
if(barre._width < 315)
{
   gotoAndPlay(1);
}
else
{
   gotoAndPlay(3);
}
```
In Ruffle, the loading bar width never increases even after being incremented. Inserting `trace(barre._width);`, you'll see that the width always stays zero. So the game never starts.

### Expected behavior

The loading bar width should gradually increase, like this:
![image](https://user-images.githubusercontent.com/71368227/179280917-38723269-426e-4778-a67b-27b6388b8b95.png)

### Affected platform

Desktop app

### Operating system

Windows 10

Toad06:
Interesting... and funny because in fact the width is equal to 1 _before_ it gets incremented. 😄

For instance, if you replace the whole code in frame 1 with this one:
```
trace("width: " + barre._width); // Flash Player: gradually incremented; Ruffle: 1 on the first call, then always 0
trace("height: " + barre._height); // always 0
barre._width = barre._width + 1;
```

The issue happens because the value of `_height` is 0. When the width is set and incremented, Ruffle calls [DisplayObject::set_width](https://github.com/ruffle-rs/ruffle/blob/4421fbff6aac95fcda478dcaec646ce394dab223/core/src/display_object.rs#L809): one of the calculations implies a division by zero (`value / object_height`) which ends up affecting the value of `_width` in an unexpected way (always 0 on the next steps).

It's possible to make the file work in Ruffle by using JPEXS: Shapes > DefineShape(29) > Raw edit > shapeBounds > Ymax (set a value greater than 0).

A few other observations:
- If Xmin/Xmax and Ymin/Ymax are all 0, Flash Player won't increment the width.
- In SWFv8+, it seems like the loading bar is no longer displayed but the width is still incremented, so the game can start.
- Same rules apply when setting the height.

I have [a branch](https://github.com/Toad06/ruffle/tree/set_width_height) that fixes the issue, as well as [Megaman Zero](https://uploads.ungrounded.net/313000/313265_zero15.swf) (cc. #1966) where platforms seem to rely on this behavior. Not sure if that's the correct way to do it though.

Issue #1966, opened 2020-12-14
Sprites not loading in


MaxThunderscore:
While testing some flash games prior to uploading them to archive.org I ran into this problem with the following games:

GoGo Happy & Smile: character falls through the ground; standing unusual screen ratio
Mega Man Zero: part of the level and sprites are missing

[Flash games.zip](https://github.com/ruffle-rs/ruffle/files/5687705/Flash.games.zip)

Toad06:
- **GoGo Happy & Smile**

Same issue when running this file in the official Flash Player projector. The game works as expected in Flash and Ruffle when playing it at https://www.newgrounds.com/portal/view/288598.

- **Mega Man Zero**

The character gets instant killed because he starts the level at an incorrect Y position.

In `\frame 883\DoAction`, we have this code:
```
if(sprite._hp <= 0 || sprite._y > 160)
{
	// die
}
```
=> When the level starts, `sprite._y` is equal to 292.

Removing the whole code in `\DefineSprite (546: terrain0000a)\frame 1\PlaceObject2 (3) Depth: 46 (bound)\onClipEvent(load)` prevents the issue, though positioning still remains incorrect:
![megaa](https://user-images.githubusercontent.com/14274782/102106357-c20e4700-3e30-11eb-88da-2ee081489c90.png)

PR #8353, opened 2022-10-22
core: Don't fail if height is 0 when calling `DisplayObject::set_width`

Toad06:
This PR is an attempt at fixing #7419 and the [Megaman game](https://uploads.ungrounded.net/313000/313265_zero15.swf?1147886105) in #1966.

Here's a list of things that I'm not sure about. I will need help for these.
- The behavior in AS3 (haven't checked).
- The behavior when `_rotation` is also used (usually won't match Flash but the behavior in FP is weird).
- `self.base().instantiated_by_timeline()` was added to prevent [this test](https://github.com/ruffle-rs/ruffle/tree/master/tests/tests/swfs/avm1/issue_2084) from failing.

### DIFF
--- a/core/src/display_object.rs
+++ b/core/src/display_object.rs
@@ -883,15 +883,30 @@ pub trait TDisplayObject<'gc>:
     /// Set by the ActionScript `_width`/`width` properties.
     /// This does odd things on rotated clips to match the behavior of Flash.
     fn set_width(&self, gc_context: MutationContext<'gc, '_>, value: f64) {
+        if value < 0.0 {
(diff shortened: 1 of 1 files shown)


kmeisthax:
We probably should have a test case specifically testing resizing 0-height and 0-width shapes at various rotations. I *suspect* AS2 and AS3 will behave identically, because basically everything else with basic display object properties is identical between the two. But it should be relatively easy to port whatever test we write for AS2 over to AS3 to check for any differences.

I can write such a test if you need me to.

PR #8356, opened 2022-10-23
core,avm2,desktop: Implement `SoundMixer.computeSpectrum()`

torokati44:
This should help with a number of z0r.de loops, such as: 2627, 5482, 6984, 7719, 7757, 7761, 7801, 7812, 7850, 7855, 7903, 7907

Here is a test file based on https://code.tutsplus.com/articles/render-an-mp3-audio-spectrum-in-flash-with-computespectrum--active-10808: (warning, loud beeps and boops after pressing the play button!)
[spectrumvis.zip](https://github.com/ruffle-rs/ruffle/files/9845577/spectrumvis.zip)

Some hints were also taken from:
https://web.archive.org/web/20120410195615/http://blog.benstucki.net/?p=60

Unfortunately the FFT still doesn't match what FP does, but at least the idea is the same.
If anyone has any ideas about how exactly their implementation is broken, that would be very much welcome.

The code is not particularly elegant or efficient yet either, but as a PoC I think it's fine.

### DIFF
--- a/core/src/backend/audio/mixer.rs
+++ b/core/src/backend/audio/mixer.rs
@@ -7,6 +7,39 @@ use std::io::Cursor;
 use std::sync::{Arc, Mutex, RwLock};
 use swf::AudioCompression;
 
+/// Holds the last 2048 output audio frames. Frames can be written to it one by
+/// one, and the last completely filled 1024-wide window can be read from it.
+struct CircBuf {
+    pub samples: [[f32; 2]; 2048],
+    pub pos: usize,
+}
+
+impl CircBuf {
+    /// Creates an empty circular buffer.
+    pub fn new() -> Self {
+        Self {
+            samples: [[0.0; 2]; 2048],
+            pos: 0,
+        }
+    }
+
+    /// Writes a value into the buffer, pushing the write position forward.
+    pub fn push(&mut self, sample: [f32; 2]) {
+        self.samples[self.pos] = sample;
+        self.pos = (self.pos + 1) % 2048;
+    }
+
+    /// Returns one half of the inner buffer, the one that is not currently
+    /// being written to.
+    pub fn get(&self) -> &[[f32; 2]; 1024] {
+        if self.pos < 1024 {
+            self.samples[1024..2048].try_into().unwrap()
+        } else {
+            self.samples[0..1024].try_into().unwrap()
+        }
+    }
+}
+
 /// An audio mixer for a Flash movie.
 ///
 /// `AudioMixer` manages the audio state for a Flash movie. This can be used by any backend that
@@ -29,6 +62,9 @@ pub struct AudioMixer {
 
     /// The sample rate of the output stream in Hz.
     output_sample_rate: u32,
+
+    /// The last two windows of output samples.
+    output_memory: Arc<RwLock<CircBuf>>,
 }
 
 /// An audio stream.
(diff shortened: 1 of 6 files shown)


torokati44:
I can now say with great relief that the output matches that of FP with all parameter combinations. :)
I'd be glad to provide any observations and test data for review.
Better names for variables and functions are appreciated.

core/src/backend/audio/mixer.rs
```diff
@@ -7,6 +7,39 @@ use std::io::Cursor;
 use std::sync::{Arc, Mutex, RwLock};
 use swf::AudioCompression;
 
+/// Holds the last 2048 output audio frames. Frames can be written to it one by
+/// one, and the last completely filled 1024-wide window can be read from it.
```
kmeisthax, line 11: What was the motivating factor for mixing 1024 samples at a time? Does Flash Player specifically do this or does AS3 code depend on this behavior?

kmeisthax (approved):

core/src/backend/audio/mixer.rs
torokati44: I tested this in the following way:

I created an audio file with a repeating 1024-wide section: 512 "high" values (something like half the max amplitude), followed by 512 zero samples. After 20 repetitions of this section, I inserted an extra zero sample. This sound is played in a loop.
This resulted in the captured sample windows to be completely static for a short time (until the 20 repetitions run out), then all of them ticking by one "to the right", when the extra zero is encountered.
This told me that the capture windows are not in random points in time, but taken from successive fixed length buffers, which some integer multiple of 1024 in length.

I also tested the same with 2048-long sections (1024 high + 1024 zero), but then the captured buffer was not static - it was oscillating between two different sets of samples.
This told me that the captured buffer is shorter than 2048. So only 1024 was left as a possible option.

I also know this, because as the first ("high") half 1024-long section slides "out out view" (so only zeroes are captured in between two "ticks"), then the high values immediately start rolling back in from the left.

Now, it's possible that this depends somewhat on the audio device or OS that FP runs on, but this is at least how it behaves on my (Linux) system.

torokati44:
Here are the files I used for testing (be careful not to damage your hearing or audio devices!):
[spectrumvis.zip](https://github.com/ruffle-rs/ruffle/files/9969252/spectrumvis.zip)
[spectrumvis_src.zip](https://github.com/ruffle-rs/ruffle/files/9969255/spectrumvis_src.zip)

It's easy to switch between a bunch of embedded sound files by editing just the `.as` file, or even the P-code in the SWF (in two places) with `ffdec`.
It also logs a bunch of captured arrays as well with `trace` for easier comparison. This is how I also envision some automatic regression testing could work.

The only mismatch with FP now is the play position within the sound when it's started, but that is outside the scope of this PR I think.

----

I also tried to substitute my _bazillion_ `cos` invocations with `rustdct`, but unfortunately none of the variants it offers matches what we need exactly. It might still be possible to incorporate some optimizations it uses, or to turn the naïve DCT algorithm into a "fast" one by splitting it in halves recursively. If it turns out that the current method is too slow, e.g. on web.

core/src/backend/audio/mixer.rs
kmeisthax: That sounds plausible enough.

torokati44:
Thank you for the review and the merge! 🥳

PR #8358, opened 2022-10-23
Fix #8315 by checking for relative URL without a base

### LINKED ISSUE #8315: desktop: loadMovie does not resolve URLs correctly
### Describe the bug

When running [The Grim Adventures of Billy and Mandy: Harum Scarum](https://www.arcadehole.com/games/harum-scarum/) in the Ruffle desktop app, an additional problem occurs that does not happen in web builds of Ruffle. You can see it by running this command: 
`ruffle https://www.getgames.io/games/TheGrimAdventuresOfBillyAndMandy/HarumScarum/1/game.swf`
After clicking the Start button to begin the game, the instructions screen will briefly appear, but then the main screen will reappear in the layer above. It is still possible to click the button to dismiss the instructions and start the game, though.
image

To understand the behavior better, you can download the SWF file and run it from a local server:
`python -m http.server`
`ruffle http://localhost:8000/game.swf`

When you click the Start button in the desktop app, the server will log the following line:
`"GET /game.swf?id=1666006614616 HTTP/1.1" 200 -`

Yet when you click the Start button the Ruffle browser extension, the server will log this line instead:
`"GET /?id=1666006749466 HTTP/1.1" 200 -`

This code attached to the PlaceObject2 tag under frame 8 loads the URL:
```
onClipEvent(load){
   this.holder.loadMovie(_root.trackerPath + "?id=" + (new Date()).valueOf());
}
```
In the desktop app, the loadMovie call **loads another copy of game.swf**, causing this behavior.

### Expected behavior

The instructions screen should load normally.
image

A URL like `/?id=1666007531881` should be requested by the loadMovie call.

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

Note that `_root.trackerPath` is undefined so it is not a factor at all. Removing it from the loadMovie call produces the same result.

cyraid:
Fix #8315 by changing Desktop Fetch to use "/" with relative URL's (without a base) to emulate flash.

### DIFF
--- a/desktop/src/navigator.rs
+++ b/desktop/src/navigator.rs
@@ -24,7 +24,7 @@ pub struct ExternalNavigatorBackend {
     event_loop: EventLoopProxy<RuffleEvent>,
 
     /// The url to use for all relative fetches.
-    movie_url: Url,
+    base_url: Url,
 
     // Client to use for network requests
     client: Option<Rc<HttpClient>>,
@@ -47,12 +47,22 @@ impl ExternalNavigatorBackend {
             .redirect_policy(RedirectPolicy::Follow);
 
         let client = builder.build().ok().map(Rc::new);
+        let mut base_url = movie_url;
+
+        // Force replace the last segment with empty. //
+
+        base_url
+            .path_segments_mut()
+            .unwrap()
+            .pop_if_empty()
+            .pop()
+            .push("");
 
         Self {
             channel,
             event_loop,
             client,
-            movie_url,
+            base_url,
             upgrade_to_https,
         }
     }
@@ -107,7 +117,7 @@ impl NavigatorBackend for ExternalNavigatorBackend {
 
     fn fetch(&self, request: Request) -> OwnedFuture<Response, Error> {
         // TODO: honor sandbox type (local-with-filesystem, local-with-network, remote, ...)
-        let full_url = match self.movie_url.join(request.url()) {
+        let full_url = match self.base_url.join(request.url()) {
             Ok(url) => url,
             Err(e) => {
                 let msg = format!("Invalid URL {}: {}", request.url(), e);


n0samu (approved):

cyraid:
I'm good with the updating now, all tests are green on my end for github actions. Tested by @n0samu and I (thanks again @n0samu )

desktop/src/navigator.rs
```diff
@@ -106,8 +106,20 @@ impl NavigatorBackend for ExternalNavigatorBackend {
...
+                if e == ParseError::RelativeUrlWithoutBase && !request.url().starts_with('/') {
+                    self.movie_url.join(format!("/{}", request.url()).as_str())
+                } else {
+                    self.movie_url.join(request.url())
+                }
+            }
+        };
+
         // TODO: honor sandbox type (local-with-filesystem, local-with-network, remote, ...)
-        let full_url = match self.movie_url.join(request.url()) {
```
relrelb, line 110: Thanks for your first contribution!

I think the root cause of #8315 is that the desktop backend stores the entire SWF URL, whereas the web backend stores just its containing directory. So rather I'd like to align to the web backend instead. The easiest way I found to convert a full `Url` to its containing directory is using [`url::PathSegmentsMut::pop`](https://docs.rs/url/latest/url/struct.PathSegmentsMut.html#method.pop):
```rs
let mut url = Url::parse("https://example.com/path/to/game.swf")?;
url.path_segments_mut().map_err(|_| "cannot be base")?.pop();
assert_eq!(url.as_str(), "https://example.com/path/to/");
```

cyraid: @relrelb 
The ? cannot be used on a non-result returning function, the constructor returns a `Self` (`ExternalNavigatorBackend`) and not a Result. In Desktop, the URL gets checked before the `ExternalNavigatorBackend::new` for an error, and shuts down if it there is one. The `::new` could be modified to accept only a base URL that is preprocessed that way? Or could just 'panic' (use unwrap) for it, 'cause the URL is checked anyway.

cyraid: To be compatible with Web, we could just do the 'panic' (`unwrap()`) route, which I've already tested and works.

cyraid: (And I'd go with `url.path_segments_mut().unwrap().clear();` if that's alright?)

desktop/src/navigator.rs
```diff
@@ -35,7 +35,7 @@ pub struct ExternalNavigatorBackend {
 impl ExternalNavigatorBackend {
     /// Construct a navigator backend with fetch and async capability.
     pub fn new(
-        movie_url: Url,
+        base_url: Url,
```
relrelb, line 38: This is still `movie_url`, right?

desktop/src/navigator.rs
```diff
@@ -48,11 +48,16 @@ impl ExternalNavigatorBackend {
 
         let client = builder.build().ok().map(Rc::new);
 
+        // We want just the base URL.
+
+        let mut url = base_url;
+        url.path_segments_mut().unwrap().clear();
```
relrelb, line 54: I'm afraid we don't want to use `clear`, since then the whole path is removed. Consider:
```rs
let mut url = Url::parse("https://example.com/path/to/game.swf").unwrap();
url.path_segments_mut().unwrap().clear();
println!("{}", url); // "https://example.com/"
```
Instead `.pop_if_empty().pop().push("");` is appropriate, as it preserves the containing path up to the SWF itself:
```rs
let mut url = Url::parse("https://example.com/path/to/game.swf").unwrap();
url.path_segments_mut().unwrap().pop_if_empty().pop().push("");
println!("{}", url); // "https://example.com/path/to/"
```
I think the `.pop_if_empty()` only matters when the SWF URL it actually a directory (i.e. ends with a slash), but I prefer to keep it to be on the safe side.

cyraid: Yup after trying a few things, I agree. ^_^ Let me know what you think @relrelb !

cyraid:
@relrelb You're welcome! Thanks for working together with me on it! :)

PR #8406, opened 2022-10-28
swf: Introduce and use `Rectangle::width()` and `Rectangle::height()`

relrelb:


render/canvas/src/lib.rs
```diff
@@ -816,20 +816,6 @@ fn swf_shape_to_canvas_commands(
     use ruffle_render::shape_utils::DrawPath;
     use swf::{FillStyle, LineCapStyle, LineJoinStyle};
 
-    // Some browsers will vomit if you try to load/draw an image with 0 width/height.
-    // TODO(Herschel): Might be better to just return None in this case and skip
-    // rendering altogether.
-    let (_width, _height) = (
```
adrian17, line 822: to be sure: this was dead from day 1?

Let's keep the comment, I'm pretty sure I saw this warning/error in the wild at some point still.

adrian17 (approved):

render/canvas/src/lib.rs
relrelb: Seems like this code was introduced in b611067bc575ed2d190f71a13afd1110d8a1bb96. I kept the comment but removed `_width` and `_height`.

PR #8449, opened 2022-11-01
AVM1: Some Activation cleanup

moulins:
- Remove unused `GcCell`s on `Activation` and `Scope`;
- Remove the `globals` parameter from `Activation::from_nothing`
- Reduce allocations in `Activation::from_nothing`:
  - Pre-allocate the global scope on the `Avm1` context so that it can be reused;
  - Reuse the global constant pool instead of allocating a new empty one; this doesn't matter as no bytecode is directly executed on a `from_nothing` activation without bringing its own constant pool.

------

`Activation::from_nothing` now does a single allocation (in `Scope::new_local_scope`), but I don't know if this can be removed.

Naïvely, an activation created with this method shouldn't be used to directly execute bytecode or to define locals, and so it should be fine to reuse directly the global scope (and indeed, the tests still pass after doing this change), but I'm not sure this reasoning is actually correct.

### DIFF
--- a/core/src/avm1/activation.rs
+++ b/core/src/avm1/activation.rs
@@ -193,7 +193,7 @@ pub struct Activation<'a, 'gc: 'a, 'gc_context: 'a> {
     scope: GcCell<'gc, Scope<'gc>>,
 
     /// The currently in use constant pool.
-    constant_pool: GcCell<'gc, Vec<Value<'gc>>>,
+    constant_pool: Gc<'gc, Vec<Value<'gc>>>,
 
     /// The immutable value of `this`.
     ///
@@ -255,7 +255,7 @@ impl<'a, 'gc, 'gc_context> Activation<'a, 'gc, 'gc_context> {
         id: ActivationIdentifier<'a>,
         swf_version: u8,
         scope: GcCell<'gc, Scope<'gc>>,
-        constant_pool: GcCell<'gc, Vec<Value<'gc>>>,
+        constant_pool: Gc<'gc, Vec<Value<'gc>>>,
         base_clip: DisplayObject<'gc>,
         this: Value<'gc>,
         callee: Option<Object<'gc>>,
@@ -315,7 +315,7 @@ impl<'a, 'gc, 'gc_context> Activation<'a, 'gc, 'gc_context> {
             context.gc_context,
             Scope::new_local_scope(global_scope, context.gc_context),
         );
-        let empty_constant_pool = GcCell::allocate(context.gc_context, Vec::new());
+        let empty_constant_pool = Gc::allocate(context.gc_context, Vec::new());
         avm_debug!(context.avm1, "START {}", id);
 
         Self {
@@ -834,7 +834,7 @@ impl<'a, 'gc, 'gc_context> Activation<'a, 'gc, 'gc_context> {
         &mut self,
         action: ConstantPool,
     ) -> Result<FrameControl<'gc>, Error<'gc>> {
-        self.context.avm1.set_constant_pool(GcCell::allocate(
+        self.context.avm1.set_constant_pool(Gc::allocate(
             self.context.gc_context,
             action
                 .strings
@@ -1744,14 +1744,14 @@ impl<'a, 'gc, 'gc_context> Activation<'a, 'gc, 'gc_context> {
                 }
                 SwfValue::Register(v) => self.current_register(v),
                 SwfValue::ConstantPool(i) => {
-                    if let Some(value) = self.constant_pool().read().get(i as usize) {
+                    if let Some(value) = self.constant_pool().get(i as usize) {
                         *value
                     } else {
                         avm_warn!(
                             self,
                             "ActionPush: Constant pool index {} out of range (len = {})",
                             i,
-                            self.constant_pool().read().len()
+                            self.constant_pool().len()
                         );
                         Value::Undefined
                     }
@@ -2886,11 +2886,11 @@ impl<'a, 'gc, 'gc_context> Activation<'a, 'gc, 'gc_context> {
         }
     }
 
-    pub fn constant_pool(&self) -> GcCell<'gc, Vec<Value<'gc>>> {
+    pub fn constant_pool(&self) -> Gc<'gc, Vec<Value<'gc>>> {
         self.constant_pool
     }
 
-    pub fn set_constant_pool(&mut self, constant_pool: GcCell<'gc, Vec<Value<'gc>>>) {
+    pub fn set_constant_pool(&mut self, constant_pool: Gc<'gc, Vec<Value<'gc>>>) {
         self.constant_pool = constant_pool;
     }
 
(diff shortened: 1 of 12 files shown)


moulins:
After a careful look at all callers of `Activation::from_nothing` and `Activation::from_stub`, I've concluded that none of them are directly defining new local variables on the returned `Activation`, so removing the extra local scope is fine, at the price of restricting what can be done with these `Activation`s.

As such, this PR now makes `Activation::from_nothing` fully allocation-free.

core/src/avm1/activation.rs
```diff
@@ -2834,7 +2809,6 @@ impl<'a, 'gc, 'gc_context> Activation<'a, 'gc, 'gc_context> {
             self.set_variable(name, value)
         } else {
             let scope = self.scope;
```
relrelb, line 2811: nit: Let's line this up to `self.scope.define_local(...)`

core/src/avm1/scope.rs
```diff
@@ -30,10 +29,10 @@ pub enum ScopeClass {
 }
 
 /// Represents a scope chain for an AVM1 activation.
-#[derive(Debug, Collect)]
+#[derive(Copy, Clone, Debug, Collect)]
```
relrelb, line 32: I prefer keeping large structs non-`Copy`, since cloning them isn't cheap.

core/src/avm1/scope.rs
```diff
@@ -60,67 +59,31 @@ impl<'gc> Scope<'gc> {
...
-            let grandparent = parent.read().parent;
-            if let Some(grandparent) = grandparent {
-                parent = grandparent;
-            } else {
-                break;
-            }
+        if scope.class == ScopeClass::Target {
+            scope.values = clip;
+        } else {
+            scope.parent = scope.parent.map(|p| Self::new_target_scope(p, clip, mc));
```
relrelb, line 71: I think recursion can be avoided here using a simple loop.

core/src/avm1/scope.rs
kmeisthax: How large is 'large', here? We have two GC pointers - one optional, and one enumclass'd; and a binary enum. I doubt it exceeds 32 bytes.

core/src/avm1/activation.rs
moulins: I did this at the start, but I ended up with weird borrowck errors; I could try changing it to `self.scope()...` though.

core/src/avm1/scope.rs
moulins: Unfortunately no :(
With a loop you'd need to allocate the scope chain and *then* replace the target (which is what the old code did), but now that `Scope` is immutable there's no way to replace the scope object once allocated (because it's behind a `Gc`).

core/src/avm1/scope.rs
moulins: With alignment, it's 24 bytes on x86-64, and 12 bytes on WASM, so definitely not a "big" struct.

core/src/avm1/activation.rs
moulins: Fixed!

core/src/avm1/scope.rs
relrelb: Actually I consider structs larger than a pointer size as "big" enough. In my opinion implicit clones for big structs hurt code readability (See `new_target_scope` for example, it's very unclear that `scope` is a clone of `parent`, at least for me).

core/src/avm1/scope.rs
```diff
@@ -30,10 +29,10 @@ pub enum ScopeClass {
...
 
 /// Represents a scope chain for an AVM1 activation.
-#[derive(Debug, Collect)]
+#[derive(Copy, Clone, Debug, Collect)]
 #[collect(no_drop)]
 pub struct Scope<'gc> {
-    parent: Option<GcCell<'gc, Scope<'gc>>>,
+    parent: Option<Gc<'gc, Scope<'gc>>>,
     class: ScopeClass,
     values: Object<'gc>,
```
relrelb, line 37: Might not be related, but I think `values` should be reduced to just `PropertyMap<'gc, Value<'gc>>`. `Object<'gc>` includes many functionality that doesn't make sense in the context of `Scope` (e.g. prototype, watchers, interfaces etc.). Though this might belong to a separate PR.

core/src/avm1/scope.rs
```diff
@@ -60,67 +59,31 @@ impl<'gc> Scope<'gc> {
...
     /// scope has been replaced with another given object.
     pub fn new_target_scope(
-        mut parent: GcCell<'gc, Self>,
+        parent: Gc<'gc, Self>,
         clip: Object<'gc>,
         mc: MutationContext<'gc, '_>,
-    ) -> GcCell<'gc, Self> {
-        let mut bottom_scope = None;
-        let mut top_scope: Option<GcCell<'gc, Self>> = None;
+    ) -> Gc<'gc, Self> {
```
relrelb, line 65: `new_target_scope` can actually return `Self`, then the `mc` parameter is redundant.

core/src/avm1/scope.rs
```diff
@@ -49,7 +48,7 @@ impl<'gc> Scope<'gc> {
     }
 
     /// Construct a child scope of another scope.
-    pub fn new_local_scope(parent: GcCell<'gc, Self>, mc: MutationContext<'gc, '_>) -> Scope<'gc> {
+    pub fn new_local_scope(parent: Gc<'gc, Self>, mc: MutationContext<'gc, '_>) -> Scope<'gc> {
```
relrelb, line 51: Let's use `Self` instead of `Scope<'gc>`.

core/src/avm1/scope.rs
kmeisthax: Gotcha. So the rule is `size_of::<T>() > size_of::<usize>()` should not be `Copy`.

I wonder if this could be a custom lint...

moulins: Well, by this rule `&[T]` shouldn't be `Copy` :D 
Will remove the Copy impl anyways, I don't think it's actually used much.

core/src/avm1/scope.rs
moulins: What about `with` blocks? Don't they trigger watchers and all that jazz?

core/src/avm1/scope.rs
moulins: `new_target_scope` needs the `mc` parameters to allocate the parent scopes. Will change for `new_with_scope` though.

core/src/avm1/scope.rs
relrelb: Since this isn't really related to this PR, let's defer the discussion to another time (which I don't have currently).

relrelb (approved):
@moulins Let's merge this. Ping me once conflicts are resolved.

moulins:
Rebased the PR on #8605.

kmeisthax (approved):
Will merge after the soundpeaks PR finishes going through CI

PR #8459, opened 2022-11-02
[AVM1] Remove manual bit-wrangling in declare_properties! macro

moulins:
Now that the relevant compiler ICE is fixed, we can use the Attribute constants directly.


core/src/avm1/property_decl.rs
```diff
@@ -126,66 +122,18 @@ macro_rules! declare_properties {
...
-                Here, everything is a proper `const fn` and so this should work correctly,
-                but NO!, we hit an ICE in the compiler :(
-                See:
-                    https://github.com/rust-lang/rust/issues/81899
-                    https://github.com/rust-lang/rust/issues/84957
-
-                TODO: use the `B)` desugaring once the above ICE is fixed.
-            */
-            attributes: 0 $(| declare_properties!(@__attr $attributes))* $(| declare_properties!(@__version $version))*,
+            attributes: crate::avm1::property::Attribute::from_bits_truncate(
```
relrelb, line 132: nit: Cannot this be simplified to the following?
```rs
attributes: crate::avm1::property::Attribute::empty() $(| crate::avm1::property::Attribute::$attributes)*,
```

moulins: It can't, because overloaded operators aren't available in `const` contexts.

relrelb (approved):

PR #8489, opened 2022-11-06
Remove impl_display_object_container! macro

moulins:
It is replaced by two accessor methods on the trait + default implementations of all other methods.

### DIFF
--- a/core/src/display_object/avm1_button.rs
+++ b/core/src/display_object/avm1_button.rs
@@ -414,11 +414,14 @@ impl<'gc> TDisplayObjectContainer<'gc> for Avm1Button<'gc> {
 }
 
 impl<'gc> TInteractiveObject<'gc> for Avm1Button<'gc> {
-    fn ibase(&self) -> Ref<InteractiveObjectBase<'gc>> {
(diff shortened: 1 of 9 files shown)


core/src/display_object/avm1_button.rs
```diff
@@ -401,7 +401,16 @@ impl<'gc> TDisplayObject<'gc> for Avm1Button<'gc> {
 }
 
 impl<'gc> TDisplayObjectContainer<'gc> for Avm1Button<'gc> {
-    impl_display_object_container!(container);
+    fn raw_container(&self) -> Ref<'_, ChildContainer<'gc>> {
```
kmeisthax, line 404: Minor note: We use the same pattern for `TInteractiveObject`, but the functions are `ibase` and `ibase_mut` respectively. Perhaps we should rename those to `raw_interactive` and `raw_interactive_mut` to match...

kmeisthax (approved):

core/src/display_object/avm1_button.rs
moulins: Good idea, will do :)

Aaron1011 (approved):

PR #8517, opened 2022-11-10
web: Allow chromium version detection

danielhjacobs:
See https://github.com/giggio/node-chromedriver/blob/main/README.md#include-chromium


danielhjacobs:
Don't know what the cause of the error might be...

danielhjacobs:
So, the reason this doesn't work with CI is https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#browsers-and-drivers. The Ubuntu image has both chromium and Google Chrome installed, but different versions. So the `.npmrc` works locally if just chromium is installed, but doesn't work with CI. Not sure anything can be done except recommending `INCLUDE_CHROMIUM=true npm install` if a user has chromium but not Chrome. Using a different .npmrc depending on local versus CI sounds obnoxious.

PR #8539, opened 2022-11-13
avm2,audio: Implement `SoundChannel.leftPeak` and `SoundChannel.rightPeak`

torokati44:
A bit of a WIP, but works well enough.
For example, it improves the following z0r loops at least: 6268, 6411, 6565.

Unlike with #8356, I didn't make sure to decouple the window of time over which this peak is taken from the output stream of the backend, because it didn't seem as important here.

### DIFF
--- a/core/src/avm2/globals/flash/media/soundchannel.rs
+++ b/core/src/avm2/globals/flash/media/soundchannel.rs
@@ -34,22 +34,40 @@ pub fn class_init<'gc>(
     Ok(Value::Undefined)
 }
 
-/// Stub `SoundChannel.leftPeak`
+/// Implements `SoundChannel.leftPeak`
 pub fn left_peak<'gc>(
-    _activation: &mut Activation<'_, 'gc, '_>,
-    _this: Option<Object<'gc>>,
+    activation: &mut Activation<'_, 'gc, '_>,
+    this: Option<Object<'gc>>,
     _args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
-    Err("Sound.leftPeak is a stub.".into())
+    if let Some(instance) = this
+        .and_then(|this| this.as_sound_channel())
+        .and_then(|channel| channel.instance())
+    {
+        if let Some(peak) = activation.context.audio.get_sound_peak(instance) {
+            return Ok(Value::Number(peak[0].into()));
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


torokati44:
Okay, on my part, I'm happy with the state of this now.

My findings are the following:

A constant (DC) value (regardless of its signum) always results in these properties being zero.
If the test sound is a 20Hz (or lower) square wave, these properties are (almost) always zero.
If the test sound is a 25Hz (or higher) square wave, these properties are (almost) always non-zero.
The "almost" is due to the quasi-nondeterministic relative timing of the edges of the square wave and

This means that the peak is computed not as the maximum of the absolute sample values, but from the minimum and maximum of the raw sample values. (The square wave is between a positive and a negative value, so the peak could never be zero if this wasn't the case.)

It also means that the window over which the peak is measured across a window of somewhere between `44100/25/2 = 1102.5` and `44100/25/2 = 882` frames in length. This is because the 20Hz square wave (often resulting in 0 peak) is long enough to fit the whole buffer inside one of its (half-length) constant regions, while the 25Hz one isn't.

This also nicely coincides with the 1024-long windows used for `computeSpectrum`. That same memory can't be used for this however, as the peaks are computed per-channel, not from the mixed output.

torokati44:
This also fixes/improves the following z0r.de loops:
6227
6268
6356
6456
6565
7162

torokati44:
Argh, there might be a bit too many new `unwrap`s in there... I might have to do something about those...

torokati44:
Okay, I got rid of most of the added `unwrap`s. The only one left just follows the pattern of surrounding code.

Here is a test SWF I repurposed from #8356: (again, be careful with your audio devices and hearing organs!)
[spectrumvis_peaks.zip](https://github.com/ruffle-rs/ruffle/files/10051085/spectrumvis_peaks.zip)

This time, the important bit is not the curves drawn, but the values `trace`d.

Here is a comparison plot I did with the above test file:
image

The sound here is as follows: 2 seconds of silence, then 2 seconds each of square waves increasing from 20 to 25 Hz. It then loops.

There are some slight differences due to the different relative timing of the square wave and the buffer windows (I think), but overall I think the match is good enough. I of course had to shift the two series in time to make them match up well.

kmeisthax (approved):

torokati44:
Thank you! \o/

PR #8561, opened 2022-11-17
[Draft] AVM2: Implement Stage.invalidate

golfinq:
Draft pull request implementing `Stage.invalidate` for AVM2 

- Tests need to be written to ensure behavior between flash player and this code is equivalent. 
  - Based on my reading of the docs, FP broadcasts a `Render` event to all child display objects to force redraws, which I am doing with the `dirty` parameter.
- Code also needs some cleaning up.

### DIFF
--- a/core/src/avm2/globals/flash/display/stage.rs
+++ b/core/src/avm2/globals/flash/display/stage.rs
@@ -11,7 +11,7 @@ use crate::avm2::Multiname;
 use crate::avm2::Namespace;
 use crate::avm2::QName;
 use crate::avm2::{ArrayObject, ArrayStorage};
-use crate::display_object::{StageDisplayState, TDisplayObject};
+use crate::display_object::{StageDisplayState, TDisplayObject, TDisplayObjectContainer};
 use crate::string::{AvmString, WString};
 use gc_arena::{GcCell, MutationContext};
 use swf::Color;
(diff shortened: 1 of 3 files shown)


kmeisthax:
This all seems plausible.

As for the tests... the approach I took for `DisplayObject` lifecycle events was to basically log everything in a bunch of tests that poked the timeline at various points during a frame. You can copypaste the event logger out of one of these tests - it already listens for render events because I wasn't sure what triggered them at the time.

I think all we need is a test for calling `Stage.invalidate()` per lifecycle event, i.e. a test that calls it on `exitFrame`, one for `enterFrame`, `frameConstructed`, and at least one frame script (those are the ones you get if you add code to an AS3 timeline). Make sure to construct an off-stage display object on each one so we can check if these are broadcasts or if they trickle down through the render tree.

The log output might be more than a little confusing to parse and I'll personally review diverging test output if necessary. My *guess* is that Flash Player will either immediately broadcast a render event on stack, or send the event at the end of frame.

danielhjacobs:
The rebase failure was caused by https://github.com/ruffle-rs/ruffle/pull/9050. I believe the clippy suggestion is correct.

adrian17:
I have some concerns here:

Does `invalidate` actually "invalidate"/"set cpu dirty" on the DOs? Is this observable? From the docs, my impression was that the main, if only, side effect was sending `render` event on renders.

> FP broadcasts a Render event to all child display objects to force redraws, which I am doing with the dirty parameter.

Correct me if I'm wrong, but I thought the point of `render` event wasn't to force redraws, it was to give the script an ability to hook up an event listener to fire just before a render. Like in Achievement Unlocked 2:
```
         if(stage)
         {
            addEventListener(Event.RENDER,this.renderHandler);
            stage.invalidate();
         }
```
Also in docs:

> The render event gives you an opportunity to make changes to the display list immediately before it is actually rendered. This lets you defer updates to the display list until the latest opportunity. This can increase performance by eliminating unnecessary screen updates.

(whether the event is fired before all DOs are drawn or just before this particular DO starts getting drawn is to be checked)

golfinq:
> Correct me if I'm wrong, but I thought the point of `render` event wasn't to force redraws, it was to give the script an ability to hook up an event listener to fire just before a render. Like in Achievement Unlocked 2:

I agree with you that the `render` event being used to hook behavior to renders, but in that case I don't see why the `stage.invalidate();` is necessary at all in your code snippet. The assumptions which went into this PR comes from a mix of the docs, specifically this sentence "Calling the invalidate() method signals Flash runtimes to alert display objects on the next opportunity it has to render the display list", and secondary material like this http://www.betriebsraum.de/blog/2007/11/03/stageinvalidate-and-eventrender-for-ui-elements/  which I reconciled to mean "Use stage.invalidate(); when you have changed something and you want that thing to be immediately drawn" and then translated that to "In ruffle, I need to indicate that these things should be immediately drawn regardless if they have been indicated elsewhere to not be". I have also had a lot of trouble getting `invalidate` to do anything in the many attempts at tests I have tried to write (without adobe animate, purely with mxmlc, ffdec, and actionscript). That being said, this is based off of assumptions and whether or not the code is causing flash games to work. If you have experience with this function I would be happy to change it to better match that experience.

adrian17:
Sorry for a late response :) 

> but in that case I don't see why the stage.invalidate(); is necessary at all in your code snippet

Because the RENDER event is documented to _not fire at all_ unless someone called `invalidate()`.

```
You must call the invalidate() method of the Stage object each time you want a render event to be dispatched
```

> which I reconciled to mean "Use stage.invalidate(); when you have changed something and you want that thing to be immediately drawn"

Rather, this means "when you want to manually draw something lazily just before it's about to be shown on screen". As far as I'm aware, it doesn't mean for the internal FP machinery to modify any internal drawings's state (and the change you did for Ruffle `drawing.set_dirty`, I also think would have no observable effect). If you have any example otherwise, I'll happily update that assumption ofc :D

golfinq:
> Because the RENDER event is documented to _not fire at all_ unless someone called `invalidate()`.

Oooooh, that helps my understanding a lot. I don't think this PR is correct in it's implementation then, as it's based on faulty assumptions. I am deferring to your understanding as I see that it matches the documentation better - I am going to close this PR then as in order to be salvaged it'll need to be rewritten anyway.

PR #8566, opened 2022-11-17
avm2: Stub DropShadowFilter

Calvin304:
seems to me that actual filter implementations will be in the DisplayObject.filters setter.

PR #8596, opened 2022-11-22
avm2: stub DropShadowFilter

Calvin304:
successor to #8566 (btw i'm starting to dislike git & github)


core/src/avm2/globals/flash/filters/DropShadowFilter.as
```diff
@@ -0,0 +1,54 @@
+package flash.filters {
+	public final class DropShadowFilter extends BitmapFilter {
+		public var alpha: Number;
+		public var angle: Number;
+		public var blurX: Number;
+		public var blurY: Number;
+		public var color: uint;
+		public var distance: Number;
+		public var hideObject: Number;
```
Aaron1011, line 9: This should be a `Boolean`

core/src/avm2/globals/globals.as
```diff
@@ -130,6 +130,7 @@ include "flash/filters/ColorMatrixFilter.as"
 include "flash/filters/DisplacementMapFilter.as"
 include "flash/filters/DisplacementMapFilterMode.as"
 include "flash/filters/GlowFilter.as"
+include "flash/filters/DropShadowFilter.as"
```
marcograss, line 133: maybe keep those alphabetically ordered?

PR #8603, opened 2022-11-23
avm1: Avoid code duplication in Array.splice

michiel2005:
Avoided code duplication in avm1 array.rs as asked
I learned rust for this project so I hope everything is working as expected.
no tests failed with cargo test --all.

### DIFF
--- a/core/src/avm1/globals/array.rs
+++ b/core/src/avm1/globals/array.rs
@@ -356,7 +356,7 @@ pub fn splice<'gc>(
         delete_count: i32,
         items: &[Value<'gc>],
         i: i32,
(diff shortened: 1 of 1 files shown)


michiel2005:
I see that the required checks are failing so I closed it

michiel2005:
Many checks are failing. is there a way I can run them on my side to test them before I do a pull request?

kmeisthax:
Yes. `cargo fmt` will fix formatting to match what the CI asks and `cargo clippy --all --tests` will run all the other non-formatting checks. `cargo test` actually runs the full regression test suite, but it's really CPU intensive.

kmeisthax (approved):

PR #8613, opened 2022-11-24
[AVM1] Replace most of impl_custom_object! macro by default impls on TObject

moulins:
This allows removing a lot of code duplication in "core" AVM1 object types.


adrian17 (approved):

Herschel:
Thank you!

PR #8623, opened 2022-11-26
avm2: implement get and set of displayobject filters

wsxarcher:


core/src/avm2/globals/flash/display/displayobject.rs
```diff
@@ -239,23 +239,38 @@ pub fn set_scale_x<'gc>(
...
-/// Implements `filters`'s getter.
 pub fn filters<'gc>(
     activation: &mut Activation<'_, 'gc, '_>,
-    _this: Option<Object<'gc>>,
+    this: Option<Object<'gc>>,
     _args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
-    log::warn!("DisplayObject.filters getter - not yet implemented");
+    if let Some(dobj) = this.and_then(|this| this.as_display_object()) {
+        return Ok(ArrayObject::from_storage(activation, dobj.filters())?.into());
```
adrian17, line 248: Interesting, just checked that indeed it returns a new Array every time.
(`trace(this.filters === this.filters); // false`). Could you please add a test for this?

core/src/avm2/globals/flash/display/displayobject.rs
```diff
@@ -239,23 +239,38 @@ pub fn set_scale_x<'gc>(
...
-    log::warn!("DisplayObject.filters setter - not yet implemented");
+    if let Some(dobj) = this.and_then(|this| this.as_display_object()) {
+        let new_filters = args
+            .get(0)
+            .cloned()
+            .unwrap_or(Value::Undefined)
+            .coerce_to_object(activation)?;
+
+        if let Some(filters_array) = new_filters.as_array_object() {
+            if let Some(filters_storage) = filters_array.as_array_storage() {
```
adrian17, line 266: FYI, AS3 does a type check here:
- it checks that the assigned value is an array (or null/undefined) and throws a TypeError otherwise (this ideally will be automatic once we add proper typed bindings to DisplayObject, so it can be skipped here)
- it checks that all array elements derive from Filter, and throws an ArgumentError otherwise (this should be implemented here).

If you want to keep it out of scope, it's okay, but leave a comment please.

wsxarcher: Done and addedd tests

core/src/avm2/globals/flash/display/displayobject.rs
wsxarcher: Done and checked on flash debug

adrian17 (approved):

PR #8625, opened 2022-11-27
Bump gc-arena to upstream's current master branch

moulins:
Because `Gc<'_, T>`'s Debug impl now prints the referenced value, (almost) all #[derive(Debug)] on types containing GC'd values were removed (otherwise, Ruffle would enter an infinite recursion when trying to debug-print objects).

This is almost certainly overkill, but I didn't want to take any risks, and useful Debug impls can be re-added again where needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,9 +33,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.20"
+version = "0.7.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc936419f96fa211c1b9166887b38e5e40b19958e5b895be7c1f93adec7071ac"
+checksum = "b4f55bd91a0978cbfd91c457a164bab8b4001c833b7f323132c0a4e1922dd44e"
 dependencies = [
  "memchr",
 ]
@@ -135,9 +135,9 @@ dependencies = [
 
 [[package]]
 name = "ash"
-version = "0.37.1+1.3.235"
+version = "0.37.0+1.3.209"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "911015c962d56e2e4052f40182ca5462ba60a3d2ff04e827c365a0ab3d65726d"
+checksum = "006ca68e0f2b03f22d6fa9f2860f85aed430d257fec20f8879b2145e7c7ae1a6"
 dependencies = [
  "libloading",
 ]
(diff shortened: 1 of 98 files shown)


adrian17:
> Because Gc<'_, T>'s Debug impl now prints the referenced value, (almost) all #[derive(Debug)] on types containing GC'd values were removed

If I'm reading this right, this does feel like a bit of a loss. With AVM debugging, previously we had access to logs like:
```
Stack push 1: Object(EventObject(EventObject { type: "keyDown", class: flash.events::KeyboardEvent, ptr: 0xdacacec530 }))
```
or
```
Stack push 1: Object(StageObject(StageObject { name: BLOCnog_fags101_fla::preloader_1,
    display_object: Some(MovieClip(MovieClip(GcCell(Gc { ptr: 0xdac9829628 })))), ptr: 0xdac98a8aa0 }))
```
Even though the pointers obscure 99% of information, a lot of the time this alone was enough to look for the same object in the log. If anything, I hoped for more explicit logs in the future, not more opaque :(

moulins:
> Now it would be just Object(_), right?
Yes, unfortunately :c

I guess the proper thing to do is to re-add manual `Debug` impls for all AVM1/2 `Objects`. It's more code than current `master`, but at least we could customize what information we show.

The lazier option would be ask upstream to revert the `Debug` change in upstream (or just do it in our fork), by arguing that it isn't actually useful because most `Gc`'d object will have circular references in some ways.

@adrian17 What do you think is the best way forwards?

Herschel:
My first instinct matches your argument that this is something `gc-arena` should slightly care about, as circular structures is a big use case of GC, but I don't have a great solution! I wonder how hard it would be to make something like a `GcDebug` macro that could omit or smartly print GCed fields to some depth.

It's a bit scary because I'm not sure how many of the gutted Debug impls are actually useful for us. We definitely want to keep a decent Debug impl of AVM1/AVM2 objects. I'd be okay with continuing to with our own fork, if that's an easier stepping stone.

moulins:
> It's a bit scary because I'm not sure how many of the gutted Debug impls are actually useful for us. We definitely want to keep a decent Debug impl of AVM1/AVM2 objects. I'd be okay with continuing with our own fork, if that's an easier stepping stone.

I suspect a bunch of them were actually never exercised; I think the proper solution is to revert all the changes to the format strings I made, and then put manual Debug implementations that reproduces the old behavior of stopping at the first `Gc`. Sure it means a little more boilerplate, but I don't think it's that bad, and with luck it will encourage drive-by changes to improve the output (e.g. use a 'short' representation by default, and switch to a more complete output if the alternate formatting is specified).

Herschel:
That sounds good to me! Maybe manually implementing the `Debug` impls could be its own PR(s).

moulins:
I'll rename this PR to be about `Debug` impls, I think.

moulins:
Opened #8964 to deal with derived `Debug` implementations.

I'll close this PR for the time being, and reopen it once the above PR is merged.

PR #8629, opened 2022-11-27
avm2: Implement filefilter filereference

wsxarcher:
* Implement FileFilter in AS3
* Implement FileReference and FileReferenceList stubs

The stubs will allow some games that initialize the file selectors before using them, to be able to run. Anyway most games file selection are only used on a user action.

/cc @marcograss


core/src/avm2/globals/flash/net.rs
```diff
@@ -1,5 +1,7 @@
 //! `flash.net` namespace
 
+pub mod file_reference;
```
adrian17, line 3: can't these two be defined in AS too?

wsxarcher: Temporary yes, but in the future when someone will implement the browse or download method they will need to do OS or browser operations afaik.

adrian17: Sure, but we still keep the class defined in ActionScript - when we actually implement it, we can annotate the methods with `native function`, so that they can still be implemented in Rust.
Baiscally, the .as class replaces `pub fn create_class(...)` boilerplate; ideally long-term we'd like to convert all our class declarations from Rust to AS.

wsxarcher: Ok now it's more clear because I did not see many in AS and I thought it was for perf reason

adrian17: No, the real reason is that we added AS classes way later, and some of these existing Rust-defined classes are slightly nontrivial to convert, so nobody did them yet ;)

wsxarcher: Ok! Fixed as requested

wsxarcher:
Yes everything is correct!

PR #8634, opened 2022-11-28
avm1: Various minor fixes in filters

Toad06:
[Here is the file (.SWF)](https://github.com/ruffle-rs/ruffle/files/10105540/filters.zip) I used for testing.

<details>
  <summary>View raw code</summary>
  
```
trace("ConvolutionFilter");
var x = new flash.filters.ConvolutionFilter();
trace(x.alpha);
x.alpha = NaN;
trace(x.alpha);

trace("");
trace("DisplacementMapFilter");
var x = new flash.filters.DisplacementMapFilter();
trace(x.alpha);
x.alpha = NaN;
trace(x.alpha);

trace("");
trace("DropShadowFilter");
var x = new flash.filters.DropShadowFilter();
trace(x.alpha);
x.alpha = NaN;
trace(x.alpha);
trace(x.strength);
x.strength = NaN;
trace(x.strength);

trace("");
trace("GlowFilter");
var x = new flash.filters.GlowFilter();
trace(x.alpha);
x.alpha = NaN;
trace(x.alpha);
trace(x.strength);
x.strength = NaN;
trace(x.strength);

trace("");
trace("GradientBevelFilter");
var x = new flash.filters.GradientBevelFilter();
x.colors = [16777215,13421772,0];
trace(x.alphas);
x.alphas = [2,NaN];
trace(x.alphas);
trace(x.strength);
x.strength = NaN;
trace(x.strength);

trace("");
trace("GradientGlowFilter");
var x = new flash.filters.GradientGlowFilter();
x.colors = [16777215,13421772,0];
trace(x.alphas);
x.alphas = [2,NaN];
trace(x.alphas);
trace(x.strength);
x.strength = NaN;
trace(x.strength);
```
</details>

### DIFF
--- a/core/src/avm1/globals/gradient_bevel_filter.rs
+++ b/core/src/avm1/globals/gradient_bevel_filter.rs
@@ -7,6 +7,7 @@ use crate::avm1::object::gradient_bevel_filter::GradientBevelFilterObject;
 use crate::avm1::property_decl::{define_properties_on, Declaration};
 use crate::avm1::{ArrayObject, Object, TObject, Value};
 use crate::string::{AvmString, WStr};
+use crate::types::F64Extension;
 use gc_arena::MutationContext;
 
 const PROTO_DECLS: &[Declaration] = declare_properties! {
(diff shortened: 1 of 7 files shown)


Toad06:
Clippy lint is kidding me. 🙃

danielhjacobs:
Not super familiar with Rust, but you should be able to do an allow above each of the functions, right?

Example here:
https://github.com/ruffle-rs/ruffle/blob/b8745f0ff1e278831269d190a4a86adaaa9201ea/core/src/avm2/value.rs#L455
```rust
#[allow(clippy::manual_clamp)]
```

Toad06:
Thank you, looks like this was indeed the solution. 😉

core/src/avm1/globals/gradient_bevel_filter.rs
```diff
@@ -347,7 +354,7 @@ pub fn set_strength<'gc>(
         .get(0)
         .unwrap_or(&1.into())
         .coerce_to_f64(activation)
-        .map(|x| x.clamp(0.0, 255.0))?;
+        .map(|x| x.max(0.0).min(255.0))?;
```
adrian17, line 357: why is the clamp change needed?

Toad06: If x is `f64::NAN`, using `.clamp()` returns NaN ; using `.max(0.0).min()` returns 0. In Flash Player, this property returns 0 in this situation. Properties like `blurX` and `blurY` return NaN, so they remain unchanged and still use `.clamp()`.

adrian17: How about putting it in a separate function, with a comment repeating what you wrote here? Either freestanding `nan_lossy_clamp(x, 0.0, 255.0)` or via trait `x.nan_lossy_clamp(0.0, 255.0)`. Name subject to change ;D

This also lets you have only a single clippy allow attribute instead of having it everywhere.

adrian17 (approved):

PR #8649, opened 2022-12-01
avm2: Implement `TextField` scrolling parameters

kmeisthax:
image

(Not pictured: significant [graphical corruption](https://pooper.fantranslation.org/@kmeisthax/109435857655768366) and 10 second pixel transition times)

This implements textfield scrolling parameters, which were the last thing needed by `FlxText`. Flixel games that use text can now go in-game (though they run slow).

No tests are included since this just implements properties that were already present in AVM1. I couldn't find any tests for them in the AVM1 tests directory; if there were any I would have ported them over.


adrian17 (approved):
LGTM. Can profile the perf issues after merging, but gotta push some of the JS version bumps first.

PR #8667, opened 2022-12-02
avm2: Fix no-op gotos

kmeisthax:
Tests courtesy @adrian17

This fixes a bug I introduced with the massive timeline fix PR: no-op gotos need to emit events and run frame scripts, but I decided to just make them go through the whole goto machinery assuming that it would just do nothing. Turns out it actually *does* do something: recreate objects that were moved or removed by AS3.

This also includes two ignored tests that won't work until I implement `RemoveObject` property nulling.


adrian17 (approved):
Looks safe, let's just squash it on merge.

PR #8676, opened 2022-12-03
Re-enable image tests on Linux

Aaron1011:


danielhjacobs:
So this works. But CI didn't pass until the image tests were disabled before. I'm very confused. What exactly happened, just out of curiosity?

danielhjacobs:
Actually CI ran without image tests despite the changes.

danielhjacobs:
Sorry for deleted comments. I kept reading things wrong. Anyway, do we know what happened before?

Issue #8324, opened 2022-10-18
Critter Pinball: Cannot launch ball


n0samu:
### Describe the bug

In [Critter Pinball](http://www.freewebarcade.com/game/critter-pinball/), holding down the enter or down arrow key pushes down the plunger, but releasing the key does not launch the ball. Also, Ruffle lags so much that you need to hold down the key for about 1 minute to push the plunger down all the way.

Link to SWF file: http://www.freewebarcade.com/game/critter-pinball/critter-pinball.swf

### Expected behavior

Holding down the enter or down arrow key, then releasing it should launch the ball.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

This error is repeatedly shown in the terminal:
`[ERROR ruffle_core::avm2::events] Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x2b391487cd0 }) to handler FunctionObject(FunctionObject(GcCell(Gc { ptr: 0x2b391108698 }))) : RustError("Non-public property https://ruffle.rs/AS3/impl/EventDispatcher/::dispatch_list not found on Object")`

n0samu:
Fixed by #8699, although the game is still intolerably slow in Ruffle.

PR #8699, opened 2022-12-05
avm2: Fix indirect event dispatching

adrian17:
When we use an intermediate event dispatcher, as in:
```
var dispatcher = new EventDispatcher(obj);
dispatcher.dispatchEvent(...);
```
We use the dispatcher's dispatch list, not the target's. The target doesn't need to be an EventDispatcher at all.
(When bubbling/capturing, we still do use the ancestors directly as both the dispatcher and target at the same time).

This improves the behavior in general, and fixes the error in https://github.com/ruffle-rs/ruffle/issues/8324 .


kmeisthax (approved):
How did I forget to account for this? Good catch.

PR #8701, opened 2022-12-05
core: debug: use ctrl+alt+f to dump DO render tree

adrian17:
Only enabled with `--features=avm_debug`.

Example dump for a very simple preloader. Shows the AS class, DO name and DO type.
```
 rel(0,0) abs(0,0) flash.display::Stage  Stage
  rel(0,0) abs(0,0) Main root1 MovieClip
   rel(360,260) abs(0,0) BLOCnog_fags101_fla::preloader_1 instance17 MovieClip
    rel(-360,-260) abs(0,0) BLOCnog_fags101_fla::bg_1000_2 bg MovieClip
     rel(0,0) abs(0,0) BLOCnog_fags101_fla::bgAni_10000_3 instance2 MovieClip
      rel(0,0) abs(0,0) flash.display::Shape instance1 Graphic
    rel(0,0) abs(170,359.55) flash.display::Shape instance3 Graphic
    rel(-150,-130) abs(210,130) BLOCnog_fags101_fla::adcon_4 adBox MovieClip
     rel(0,0) abs(210,130) flash.display::Shape instance4 Graphic
    rel(0,0) abs(170,90) flash.display::Shape instance5 Graphic
    rel(10,-206) abs(0,0) flash.display::SimpleButton redirFags Avm2Button
    rel(0,0) abs(200,390) flash.display::Shape instance10 Graphic
    rel(-150,170) abs(210,420) flash.display::MovieClip loaderBar MovieClip
     rel(0,0) abs(210,420) flash.display::Shape instance11 Graphic
     rel(300,0) abs(210,420) flash.display::MovieClip bar MovieClip
      rel(0,0) abs(210,420) flash.display::Shape instance12 Graphic
    rel(0,170) abs(0,0) flash.display::SimpleButton startBUT Avm2Button
```

PR #8702, opened 2022-12-06
core: `Loader` must pass enter/construct frame lifecycle events down

kmeisthax:
image

(Almost - this also requires a few method impls that @adrian17 also needs to PR)

Turns out our `Loader` display object was also missing `enter_frame`/`construct_frame` impls. All containers need to implement these methods or the objects put in them will freeze up.

### DIFF
--- a/core/src/display_object/loader_display.rs
+++ b/core/src/display_object/loader_display.rs
@@ -80,6 +80,18 @@ impl<'gc> TDisplayObject<'gc> for LoaderDisplay<'gc> {
     fn as_interactive(self) -> Option<InteractiveObject<'gc>> {
         Some(self.into())
     }
+
+    fn enter_frame(&self, context: &mut UpdateContext<'_, 'gc, '_>) {
+        for child in self.iter_render_list() {
+            child.enter_frame(context);
+        }
+    }
+
+    fn construct_frame(&self, context: &mut UpdateContext<'_, 'gc, '_>) {
+        for child in self.iter_render_list() {
+            child.construct_frame(context);
+        }
+    }
 }
 
 impl<'gc> TInteractiveObject<'gc> for LoaderDisplay<'gc> {


torokati44:
Here's an interesting thing: Take z0r.de/2772 for example (https://z0r.de/L/z0r-de_2772.swf), which uses `Loader.loadBytes`.

**On desktop**, this loop just showed a blank light grey background until now, nothing else.
With this PR, it now shows a loading "spinner" in the form of an animated "...".

**However, on web:** It has worked well (actually started playing beyond the loader) even before this PR, and still works with this PR.

(Keep in mind that #8701 has to be reverted at the time of writing for the web builds to succeed.)

kmeisthax:
I was not aware of that, and I have to wonder what's going on there. Will definitely take a look inside that SWF.

Also, what do you mean exactly by needing to revert #8701? Is CI not catching build errors in web?

torokati44:
> I was not aware of that, and I have to wonder what's going on there. Will definitely take a look inside that SWF.

Thanks for investigating!

> Also, what do you mean exactly by needing to revert https://github.com/ruffle-rs/ruffle/pull/8701? Is CI not catching build errors in web?

Something like that, yeah... :| We're not entirely sure yet. Could just be just due to the `avm_debug` feature being unexpectedly enabled or disabled... We had a short discussion about this on Discord in the `#dev-general` channel a couple hours ago.

EDIT: For what it's worth, z0r 2772 does use RSLPreloader, which checks if it's in the LOCAL_WITH_FILE sandbox and changes its behavior. That may explain some of the weirdness between web and desktop.

Ah, yes, that explains a lot.

adrian17:
> Also, what do you mean exactly by needing to revert https://github.com/ruffle-rs/ruffle/pull/8701? Is CI not catching build errors in web?

Locally, building without `avm_debug` feature fails as very clearly part of my ctrl+alt+f implementation is conditionally compiled and half is not. The big question is, how the heck did it pass CI then.

PR #8708, opened 2022-12-06
ci: Don't use a PPA for mesa (`lavapipe`) anymore on `ubuntu-22.04`

torokati44:
Since this PPA was only necessary back when we used `ubuntu-20.04` for CI, where `lavapipe` wasn't yet enabled in mesa.

This may or may not has had something to do with the SIGSEGVs we've seen recently, according to: https://github.com/lutris/lutris/issues/4581#issuecomment-1304638108

I hope this doesn't make the ongoing CI mess any worse. It could at least simplify and speed up tests a bit.

### DIFF
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -50,11 +50,7 @@ jobs:
 
       - name: Install Linux dependencies
         if: runner.os == 'Linux'
-        run: |
-          sudo add-apt-repository ppa:kisak/kisak-mesa -y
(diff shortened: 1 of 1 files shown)


torokati44:
I hope PRs updating the CI config are tested with the updated config, not with the one on `master`...

danielhjacobs:
They do use the updated config, as you can see in the screenshot.

Screenshot_20221206_081204_Chrome

Aaron1011 (approved):

PR #8728, opened 2022-12-09
web: Add version number requested in #8021

### LINKED ISSUE #8021: Access to Ruffle version outside Flash file
**Describe the feature**
Right now the only way to view which version of Ruffle a user has installed is by right clicking inside of a Flash window;
image
A way to check in the pop up or the settings page could be useful, especially with users that do not use nightly builds;
image

aidanbabo:
Change the compile time "macros" such as `%VERSION_NUMBER%` to be contained in only one file `build-info.ts`.

Closes #8021

### DIFF
--- a/web/packages/extension/assets/css/popup.css
+++ b/web/packages/extension/assets/css/popup.css
@@ -22,6 +22,10 @@ button {
     color: black;
 }
 
+#version-text {
+    text-align: center;
+}
+
 #status-container {
     text-align: center;
     background: white;
--- a/web/packages/core/src/build-info.ts
+++ b/web/packages/core/src/build-info.ts
@@ -1,0 +1,10 @@
+/**
+ * Stores build information. The string literals are replaces at compile time by `set_version.js`.
+ */
+export const buildInfo = {
+    versionNumber: "%VERSION_NUMBER%",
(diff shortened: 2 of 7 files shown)


aidanbabo:
The script that actually processes the macros (`set_version.js`) still processes every file. I am honestly not too sure how it works, but I am willing to try and make it only process `build-info.ts` if it would speed up compile time and someone wants to direct me a little bit.

web/packages/extension/assets/css/popup.css
```diff
@@ -22,6 +22,10 @@ button {
     color: black;
 }
 
+#version-text {
+    text-align: center;
```
n0samu, line 26: Hello and welcome - this looks great! A little thing I noticed here is that your `version-text` element is a `span`, not a block-level element like a `div`, so your `text-align: center` styling has no effect:
image
If you change it to a div, it looks like what you probably intended:
image

n0samu: Looks good to me! 👍

n0samu (approved):

adrian17 (approved):

adrian17:
Thanks!

PR #8740, opened 2022-12-12
core: Don't fire init events on AVM2 until the first `frameConstructed` event

kmeisthax:
Currently, the `LoaderInfo` `init` event can fire too early in AVM2 movies. This PR fixes this by delaying setting the `INITIALIZED` flag until we finish constructing the first frame for the first time.

According to the AS3 documentation, LoaderInfo is not supposed to fire `init` until after the first frame has constructed and its frame scripts have run. However, we reused the `INITIALIZED` flag from AVM1; checking every `exitFrame` to see if we need to fire `init` based on that flag. Because `run_frame` does *not* check `frames_loaded` (that's done later, in `run_frame_internal`), it always sets `INITIALIZED` regardless of if frame 1 has actually run - which causes the premature `init` event.

While we do also send AVM1 `onLoad` events too early as well, I don't believe you can actually add `onLoad` to the root timeline in AVM1, meaning that this bug probably doesn't affect any AS2 games.

### DIFF
--- a/core/src/display_object/movie_clip.rs
+++ b/core/src/display_object/movie_clip.rs
@@ -2288,6 +2288,7 @@ impl<'gc> TDisplayObject<'gc> for MovieClip<'gc> {
         // AVM1 code expects to execute in line with timeline instructions, so
         // it's exempted from frame construction.
         if context.is_action_script_3() && self.frames_loaded() >= 1 {
+            let is_load_frame = !self.0.read().initialized();
             let needs_construction = if matches!(self.object2(), Avm2Value::Undefined) {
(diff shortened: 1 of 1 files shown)


kmeisthax:
Oh, weird. I guess `this.onLoad` doesn't work but `function onLoad` does?

Herschel:
Can we add a test for this?

kmeisthax:
Unfortunately, we can't - our testing harness currently preloads everything all at once, which is why it didn't catch this bug. And simulating preload would probably break existing tests.

Herschel (approved):
Thank you!

Issue #8732, opened 2022-12-10
AVM2: HitTestPoint fails when child depth = 0


CUB3D:
### Describe the bug

In the attached swf, clicking on the blue shape should trace `HitTest=true` as it does in Flash Player. However in Ruffle it currently prints `HitTest=false`.

The file in question has been modified in jpexs to set the depth of the PlaceObject2 tag in DefineSprite(2) to 0. This clip is then moved from that sprite to the root by a script. 

Changing the comparison [here](https://github.com/ruffle-rs/ruffle/blob/f79208c75fe73acc6253aba1ceacc5a825334a64/core/src/display_object/movie_clip.rs#L2487), from `>` to `>=` fixes the actual hit test when depth = 0 (although I'm not sure if this depth should actually be 0 here).

Incidentally, that change also fixes bullet collision in Upgrade Complete.

[UGCbullet.zip](https://github.com/ruffle-rs/ruffle/files/10199100/UGCbullet.zip)

### Expected behavior

The file behaves the same in Ruffle as it does in Flash Player

### Affected platform

Desktop app

### Operating system

Linux

n0samu:
Fixed by #8744

Issue #8722, opened 2022-12-08
AVM2: HitTestObject with SimpleButton not working


CUB3D:
### Describe the bug

As shown by the attached sample, trying to perform a `hitTestObject`, where one value is a SimpleButton always fails.

Adding some additional logging around the current hit test logic, shows that the bounds of the button are always 
`BoundingBox { x_min: Twips(0), y_min: Twips(0), x_max: Twips(0), y_max: Twips(0), valid: false }`

This causes an issue in [Upgrade Complete!](https://www.newgrounds.com/portal/view/499812), where neither the Play button, nor the Back button in the Ship Builder work. This also might be the cause of the issue that prevents collecting money.

(In this SWF, click to move the green clip)
[UGC Bug.zip](https://github.com/ruffle-rs/ruffle/files/10182037/UGC.Bug.zip)

### Expected behavior

hitTestObject behavior matches Flash Player

### Affected platform

Desktop app

### Operating system

Linux

CUB3D:
After playing about with this for a while, I think I've found *a* solution to this issue, currently the bounds of SimpleButton are taken as empty [here](https://github.com/ruffle-rs/ruffle/blob/2bb4e2e549ab39d2f2cb7d8194ac21bb1ca2e682/core/src/display_object/avm2_button.rs#L643).

Changing this to 
```
BoundingBox {
    x_max: Twips::ZERO,
    y_max: Twips::ZERO,
    x_min: Twips::ZERO,
    y_min: Twips::ZERO,
    valid: true,
}
```
Makes the button clickable (somewhat, the mouse seems offset but I doubt that is related). This also fixes the issue in the linked swf.

n0samu:
Fixed by #8744

Issue #8412, opened 2022-10-29
fire boy and water girl light temple not working


ROBlNET13:
i cant press play on the swf i got from newgrounds https://www.mediafire.com/file/rnqlrgcon2u9n70/Fireboy_and_Watergirl_in_The_Light_Temple.swf/file 
here is the file

Calvin304:
it seems this is caused by Avm2ButtonData.hit_area having its parent set in Avm2ButtonData::create_state and not removed in this case. this causes incoming mouse events to get transformed twice; once because it's assumed hit_area has no parent in <Avm2Button as TInteractiveObject>::mouse_pick so the mouse event is transformed to local space and again through TDisplayObject::hit_test_shape.

Calvin304:
this is fixed in #8744

PR #8744, opened 2022-12-12
AVM2: Fix a plethora of hitTest issues

### LINKED ISSUE #8278: Samsara Room: Unable to Enter Game. (AS3)
### Describe the bug

After selecting "Play" you are unable to advance into the game, and unable to click to skip a sequence of text.

### Expected behavior

After clicking "Play" the intro sequence of waking up into the room should greet you to a grandfather clock, this sequence of text can also be skipped by a normal click.

### Affected platform

Desktop app

### Operating system

Windows 10 x64

### Browser

_No response_

### Additional information

Bit of a recent regression as the last working build was 2022-10-07.

In terms of no major graphical issues (there was still some) the last working version (to my knowledge) was 2022-09-11.

If you get past that and are unable to click the clock's hour/minute hand or the tealight you have to go left/right, and back, as you're unable to interact with anything immediately in Ruffle. And you can only clear out the first area, as going into the underwater section, you can only do the first section of the room, and nothing else.

After loading into a game, the cogwheel in the top right, will lead you to a "reset game", before you check other versions of Ruffle reset your progress.

- - -

As I was testing which was the last version able to run the title correctly, current versions would run it correctly again. This was not the behavior I was getting when I was writing this report and only now was it behaving correctly after looking for the regression.

### LINKED ISSUE #8694: HitBox: Cannot click Pause button
### Describe the bug

In [HitBox](http://www.notdoppler.com/hitbox.php), clicking the button in the top-right corner is supposed to pause the game and open a menu. Instead it does nothing.

I'm aware that this is probably the same mouse pick issue affecting many AVM2 objects and buttons. I just wanted to report this in the interest of thoroughness.

### Expected behavior

Clicking the button in the top-right corner of a level should pause the game and open a menu. Clicking the button again should close the menu and unpause the game.

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

_No response_

CUB3D:
This fixes a number of issues related to `hitTestObject`, `hitTestPoint` and `mouse_pick` behavior.

Fixes:
- HitTests for depth=0 children, fixes bullet collisions in [Upgrade Complete!](https://www.newgrounds.com/portal/view/499812), #8732 
- Fixes for bounds calculations for AVM2 buttons, fixes upgrade and sell buttons in UGC ship builder, #8722 
- MouseEnabled property shouldn't affect children, fixes #8694 
- Selecting the child of an AVM2 button is considered a selection of the button itself, fixes [BLOCnog](https://www.gameshot.org/?id=9121) in game menu
- hitTestPoint, when shape_flag=true, requires that the object is present on the stage. Fixes collision in Facuum
- MouseEvent.MOUSE_UP events don't fire due to AVM2 buttons not handing MouseUpInside events. Fixes #8278 
- Don't transform coordinates if button hit_area has a parent, fixes the play button in #8412
- Ignore mouse pick for static text. Fixes "NIGHTCORE" button in [z0r-de_6516.swf](https://z0r.de/L/z0r-de_6516.swf) (also big improvement to menus in [Adventure Story](https://www.newgrounds.com/portal/view/586861/format/flash?emulate=flash) )

Todos:
- [x] Tests


core/src/display_object/avm2_button.rs
```diff
@@ -643,6 +643,20 @@ impl<'gc> TDisplayObject<'gc> for Avm2Button<'gc> {
...
+        let state = self.0.read().state;
+        if let Some(child) = self.get_state_child(state.into()) {
+            let child_bounds = child.bounds_with_transform(matrix);
+            bounds.union(&child_bounds);
+        }
+
+        bounds
+    }
+
     fn hit_test_shape(
```
Calvin304, line 660: hit_test_shape has the same precondition that you just changed in mouse_pick.

> hit_area is not actually a child, so transform point into local space before passing it down.

should it be changed here too?

torokati44:
I was hoping that this would fix https://z0r.de/L/z0r-de_6516.swf as well.
The button on the bottom right can only be clicked next to its label.
Unfortunately I can't see any change in this with this PR... :/

CUB3D:
> I was hoping that this would fix [z0r.de/L/z0r-de_6516.swf](https://z0r.de/L/z0r-de_6516.swf) as well. The button on the bottom right can only be clicked next to its label. Unfortunately I can't see any change in this with this PR... :/

This should be fixed now

core/src/display_object/avm2_button.rs
CUB3D: Nice catch, it probably should

render/src/bounding_box.rs
```diff
@@ -11,6 +11,14 @@ pub struct BoundingBox {
 }
 
 impl BoundingBox {
+    /// Create a zero-sized, valid BoundingBox
+    pub fn empty() -> Self {
```
adrian17: How about make a custom impl Default that sets valid to true?
As an all-zero BoundingBox should be implicitly valid.

CUB3D: I tried doing this originally, however it causes some of our tests to fail. (Also this change isn't needed now that avm2 buttons overload `bounds_with_transform`)

adrian17: > however it causes some of our tests to fail

Oh, that's even weirder I think. Recently I was thinking that `valid` is actually redundant and just equivalent to doing validity check manually each time.

adrian17:
OK, I think I understand most parts of it, but the ones with SimpleButton using its child's transform sounds very weird - at least, if SimpleButton was implemented purely in AS, this couldn't have happened. It's possible that indeed FP is weird like that (not the first time), but this might also indicate that something in button's tree might be implemented wrong. Hence, pinging @kmeisthax just for a second opinion.
Same with the parent() logic.

adrian17:
Oh, and one more thing, do you think you could write some AS-only tests for this? Maybe not the mouse pick, but the hit tests should be doable. If you need help, I can try write them for you.

kmeisthax:
AS3 `SimpleButton` is such an odd collection of special cases and general weirdness that it using child transforms doesn't really seem all that bad.

Keep in mind that `SimpleButton` implements *some* of the capabilities of a `DisplayObjectContainer` (i.e. being able to host arbitrary children) without actually inheriting from that base class and thus not having all of its features (such as being able to display multiple children in a render list). The base implementation of `bounds_with_transform` already accounts for containers, but this is a non-container object host, so it needs to override that implementation.

I'm not entirely sure it's even possible to implement a "pure AS `SimpleButton`" - the only way to host child objects from AS3 is to be a `DisplayObjectContainer`, which `SimpleButton` is most certainly not.

We actually have test coverage for the AS3-side parent field in `simplebutton_childprops` and `simplebutton_childshuffle`. Buttons can never be in the parent field, because of typing (again, they're not containers, they just cosplay as one). However, Ruffle does set and unset the parent field on button states, since we aren't bound by those type restrictions. Specifically:

 * When a state is created from a SWF tag, we always parent it to the button.
 * When a state is set through AS3, we always parent it to the button.
 * When the active state changes, we parent the new active state to the button and unparent everything else.

I don't quite remember the justification for that last behavior, though - it was actually written by @Herschel. I'm assuming it's to keep inactive states from seeing `this.root`.

kmeisthax (approved):

PR #8749, opened 2022-12-12
demo: Add SPL for more info

p0008874:
Small detail because why not.


ActionWavele:
Did you make an SPL file for the Web Demo to test?

n0samu:
You say "why not", I say "why?" I don't think this is necessary.

Herschel:
Thank you! But I agree that this is unnecessary. Maybe "Flash File" instead of "SWF", but mentioning SPL seems like overkill outside of the documentation.

PR #8762, opened 2022-12-12
avm2: Fix aliasing in BitmapData.copyChannel

adrian17:
Fixes the panic in z0r 5360.

(this `disposed()` check is redundant, as `check_valid()` already did this earlier.)


torokati44:
Ah, so the key was to do the `.write()` after the `.read().clone()` in the case the latter is necessary?

adrian17:
Yeah, we can't have `.write()` and `.read()` active in the same scope at the same time. With `source_bitmap.read().clone();`, the "lock" is not held after the clone is finished.

PR #8768, opened 2022-12-13
avm2: Stub SharedObject.clear and SharedObject.size

n0samu:
Gets [Accurate Slapshot](http://www.notdoppler.com/accurateslapshot.php) and [Track the Ball](http://www.notdoppler.com/tracktheball.php) loading to the main menu.


adrian17 (approved):
OK, but I think we'll need to have these implemented soon, given that otherwise game saves can get in weird state and cause games to behave weirdly.

relrelb (approved):
Thanks!

PR #8787, opened 2022-12-14
chore: Fix `clippy::uninlined_format_args` lints

relrelb:


adrian17:
I mentioned it before, the reason I didn't fix it is that it makes it extremely inconsistent with `log::warn!` family. Would it be possible for the lint to apply to these too? Otherwise not sure if we wanna have this enabled.

relrelb:
> I mentioned it before, IIRC the reason I didn't fix it is that it makes it extremely inconsistent with `log::warn!` family. Would it be possible for the lint to apply to these too? Otherwise not sure if we wanna have this enabled.

Agree about `log::warn!`, but I'm not sure this means we should disable the lint completely.
Hopefully the `log` crate will somehow enforce this lint, but it seems non-trivial.

adrian17:
Personally disagree, but not gonna fight over it either ;D
Either way, how about also updating our logging (and avm_debug), while we're at it - even if tooling doesn't enforce it?

PR #7694, opened 2022-08-17
ci: Explicitly use `ubuntu-22.04` image for all CI workflows

torokati44:
Because GitHub is really slow in moving the `ubuntu-latest` alias over to `22.04`, and it's starting to cause issues with the desktop build.

This is an attempt to resolve #7222 according to the recent discussion on Discord.

PR #8789, opened 2022-12-14
ci: Use `ubuntu-latest`

relrelb:
Since `ubuntu-22.04` became default, we no longer need to mention it explicitly.

This basically reverts #7694.

### DIFF
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   create-nightly-release:
     name: Create Nightly Release
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     outputs:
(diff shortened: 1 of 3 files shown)


adrian17:
Hmm, maybe we should keep it locked to a specific version? This is basically dependency pinning, which sounds nice.
Also otherwise I'd expect a similar dance when the next version bump happens, right?

danielhjacobs:
This is still rolling out: https://github.com/actions/runner-images/issues/6399

relrelb:
Got it, closing.

danielhjacobs:
Apparently 4 hours ago the runner-images issue was updated to say the rollout would be completing today instead of March 2023 as previously announced. That said, I have no opinion either way with regard to Adrian's suggestion, so that's something that should be decided upon.

PR #8790, opened 2022-12-14
avm2: Support indexing with QNames

adrian17:
(draft, as I have yet to write some tests.)


adrian17:
Un-WIP'd.

core/src/avm2/multiname.rs
```diff
@@ -220,12 +221,37 @@ impl<'gc> Multiname<'gc> {
...
+    #[inline(never)]
+    #[cold]
+    pub fn try_replace_with_qname(
+        &self,
+        obj: Object<'gc>,
+        activation: &mut Activation<'_, 'gc, '_>,
+    ) -> Option<Self> {
+        if let Object::QNameObject(qname_object) = obj {
+            if self.has_lazy_ns() {
+                activation.pop_stack(); // ignore the ns component on stack
```
Herschel, line 233: style nit, `let _ = activation.pop_stack();` is a little more explicit about ignoring this

adrian17: In that case, should we add `#[must_use]` to the function?

Herschel: Yep! I agree

Herschel (approved):
Thank you!

PR #8797, opened 2022-12-15
ci: Do simple numerical actions updates

danielhjacobs:
https://github.com/actions/create-release and https://github.com/actions/upload-release-asset also need to be replaced with https://github.com/softprops/action-gh-release, but I'm less sure about how to do that.

I assume replacing create-release would just involve changing `release_name` to `name` when replacing the action, but replacing upload-release-asset may involve something like https://github.com/softprops/action-gh-release/issues/203#issuecomment-1029848473


Herschel (approved):
Thank you!

PR #8800, opened 2022-12-15
avm2: Collapse PackageNamespace to Namespace

adrian17:
This feels very weird, but appears to be correct. The simplest way you can observe this in FP without hacking around is
```js
class C{ public var x; }; // trait is PackageNamespace("")
var c = new C();
trace(c[new QName("x")]); // QName is Namespace("") - and it works
```
However, testing _this_ is mutually blocked by https://github.com/ruffle-rs/ruffle/pull/8790 :D 

So yeah, let's just assume that after parsing, PackageNamespace and Namespace are exactly the same interchangable thing.

For comparison, avmplus:
https://github.com/adobe/avmplus/blob/master/core/Namespace.h#L31 - note no `NS_Package`
https://github.com/adobe/avmplus/blob/master/core/AbcParser.cpp#L1459-L1484 - both `CONSTANT_Namespace` and `CONSTANT_PackageNamespace` generate a `Namespace::NS_Public`.


Herschel:
Thank you! Should we name our variant `Namespace::Public` instead of `Namespace::Namespace` to match avmplus? Not sure which is clearer in this case.

adrian17:
> Should we name our variant Namespace::Public instead of Namespace::Namespace to match avmplus?

It could be a bit confusing (and it's also confusing in avmplus) as there would be two meanings of "public":
- _a_ public namespace (any namespace of variant Public)
- _the_ public namespace (namespace where `.is_public() == true`)
So I'm not necessarily a fan of this change.

Herschel (approved):
Sounds good to me, thank you!

adrian17:
Oh, and btw:
https://joshblog.net/2009/05/13/packages-are-actually-namespaces-in-actionscript-3/

Has a cool demo code:
```
namespace flash_events = "flash.events";
trace( flash_events::MouseEvent ); //output: [class MouseEvent]
```
But unfortunately it doesn't appear to be accepted by mxmlc :/

adrian17:
> But unfortunately it doesn't appear to be accepted by mxmlc :/

I think this could have worked:
```
namespace flash_events = "flash.events";
trace( flash_events::["MouseEvent"] );
```

PR #8813, opened 2022-12-16
chore: Inherit cargo metadata from workspace

Herschel:
Use workspace inheritance added in Rust 1.64 to de-duplicate various settings across all packages.

Settings like `version`, `homepage`, etc. are shared among the entire workspace, so we can specify these once and inherit them.
 * Most of these packages are meant to only be used by Ruffle, so we should use one `version` # for the entire workspace. 
   * `swf` being an exception that is meant to be usable on its own.
 * I kept any unique `authors` field.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,12 @@ members = [
 resolver = "2"
 
 [workspace.package]
+authors = ["Ruffle LLC <ruffle@ruffle.rs>"]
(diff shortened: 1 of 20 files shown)


relrelb (approved):

Herschel:
We could try inheriting some of the dependencies as well, although that's a little scarier. Could be a future PR.

PR #8816, opened 2022-12-17
Improve code consistency by adding missing semicolomns

marcograss:
very basic PR to increase the code quality/consistency while I am still learning the project

I checked where semicolons were missing and added them while checking it was ok to do so and that tests were ok

### DIFF
--- a/core/build_playerglobal/src/lib.rs
+++ b/core/build_playerglobal/src/lib.rs
@@ -356,7 +356,7 @@ fn write_native_table(data: &[u8], out_dir: &Path) -> Result<Vec<u8>, Box<dyn st
                                 None,
                                 "",
                                 &native_instance_init_method_name,
-                            )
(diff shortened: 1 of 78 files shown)


Herschel:
Thanks for contributing! But personally I don't think this is necessary:
 * Omitting the trailing semicolon on a `()` expression is common in Rust, and cannot cause runtime errors like it can in JS.
 * Aiming for consistency is nice, but we use `rustfmt` to automatically handle style choices.
   * rustfmt currently doesn't have an opinion about this issue.
   * There's a [trailing_semicolon rustfmt option](https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#trailing_semicolon), but it's unstable and only concerns itself with `return`/`break`.
   * Since rustfmt can't enforce it, I don't think we should bother at the moment.

Issue #2435, opened 2021-01-12
Flash loop doesn't show "video" (which consists of DefineBitsJPEG3 tags) and lags the player pretty heavily


Diatonator:
Nightly 2021-01-11
Both Chrome and Windows desktop tested.

https://z0r.de/2021

No video content here (h263 or the like).
No AS3.

1) No video shown while music plays.
2) Either a Chrome tab, or a Desktop app - both hang heavily when they try to play this flash. Chrome tab can be closed only after a significant delay if this flash is being played, Windows desktop app - I had to kill it with the task manager.

DefineBitsJPEG3 tags:
![image](https://user-images.githubusercontent.com/75184268/104362511-f145d000-5524-11eb-899f-8438c4128547.png)

Those JPEGs are pretty small as you could see from the screen. Could it contribute to the issue?

Diatonator:
Still repro on Nightly 2021-12-30

Diatonator:
Still freaking out on Nightly 2022-07-31

n0samu:
This actually seems to be the same issue as https://github.com/ruffle-rs/ruffle/issues/6144#issuecomment-1051311291. I've added some trace statements to the SWF and attached it below:
[2021_trace.zip](https://github.com/ruffle-rs/ruffle/files/9510520/2021_trace.zip)
Frame 1:
```as
trace("starting movie. Frames loaded:");
trace(_framesloaded);
ifFrameLoaded(1)
{
   trace("going to frame 2");
   gotoAndPlay(2);
}
trace(this._currentFrame);
trace("going to frame 1");
gotoAndPlay(1);
trace(this._currentFrame);
trace("done with frame 1 script");
```
Frame 2:
```as
trace("Now on frame 2 script. _currentFrame value is:");
trace(this._currentFrame);
stop();
```
Here are the traces from Flash Player:
```
starting movie. Frames loaded:
0
1
going to frame 1
1
done with frame 1 script
Now on frame 2 script. _currentFrame value is:
2
```
When the frame 1 script runs, Flash Player thinks it has zero frames loaded. So `gotoAndPlay(2);` doesn't execute. It executes `gotoAndPlay(1);`, which goes to the same frame it's already on (basically does nothing). After the frame 1 script is done executing Flash Player moves onto frame 2 and stays there, and the animation plays.

On the other hand, Ruffle repeats these same traces over and over again:
```
starting movie. Frames loaded:
2
going to frame 2
2
going to frame 1
1
[...]

Issue #720, opened 2020-06-16
Multiple issues with z0r.de


ubitux:
I've been trying out ruffle to extract previews of all the swf from [z0r.de](https://z0r.de) (about 8000) using the `exporter`. Most seem to work (even though the first frame is white most of the time), but some of them fail badly:

- [1722](https://z0r.de/L/z0r-de_1722.swf): thread 'main' panicked at 'already mutably borrowed: BorrowError', /build/rust/src/rustc-1.44.0-src/src/libcore/cell.rs:798:9
- [1724](https://z0r.de/L/z0r-de_1724.swf): leaking infinite loop
- [1966](https://z0r.de/L/z0r-de_1966.swf): thread 'main' panicked at 'not implemented: 15-bit PNG', core/src/backend/render.rs:319:42
- [2021](https://z0r.de/L/z0r-de_2021.swf): leaking infinite loop
- [2279](https://z0r.de/L/z0r-de_2279.swf): thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: TryFromIntError(())', core/src/avm1.rs:418:19
- [3335](https://z0r.de/L/z0r-de_3335.swf): thread 'main' panicked at 'already mutably borrowed: BorrowError', /build/rust/src/rustc-1.44.0-src/src/libcore/cell.rs:798:9
- [3536](https://z0r.de/L/z0r-de_3536.swf): thread 'main' panicked at 'not implemented', core/src/display_object/movie_clip.rs:212:37 (defineFont4)
- [3845](https://z0r.de/L/z0r-de_3845.swf): thread 'main' panicked at 'not implemented', core/src/display_object/movie_clip.rs:212:37 (defineFont4)
- [5362](https://z0r.de/L/z0r-de_5362.swf): thread 'main' panicked at 'already mutably borrowed: BorrowError', /build/rust/src/rustc-1.44.0-src/src/libcore/cell.rs:798:9
- [5586](https://z0r.de/L/z0r-de_5586.swf): leaking infinite loop
- [6138](https://z0r.de/L/z0r-de_6138.swf): thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: AllocationError(OutOfMemory(Device))', /home/ux/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-core-0.5.4/src/device/mod.rs:309:22
[...]

PR #8820, opened 2022-12-17
core: Fix handling of some invalid JPEG data (fix #4209, #8775)

### LINKED ISSUE #4209: DefineBitsJpeg2 fails when JPEG data is truncated
https://z0r.de/259 ([SWF](https://z0r.de/L/z0r-de_259.swf)) does not display anything because the `jpeg-decoder` dependency fails to decode any of the 6 embedded JPEG images.

The actual error message is: `Error running definition tag: DefineBitsJpeg2, got failed to fill whole buffer`

After extracting the binary data of these tags into separate files, every image viewer/loader I tried opens them:
 - `convert` of ImageMagick outputs a warning: `convert: Premature end of JPEG file 'image1.jpg' @ warning/jpeg.c/JPEGWarningHandler/389.`, but converts the image nevertheless
 - Gwenview shows the same result as Flash Player, and what `convert` exports
 - VS Code shows the (incomplete) bottom row of blocks as blank (transparent)

Of course Adobe Flash Player is also willing to decode these images just fine.

There is a report for this already in the repo of the `jpeg-decoder` crate, but it's closed without a proper fix: https://github.com/image-rs/jpeg-decoder/pull/98

### LINKED ISSUE #8775: Lami images bug
### Describe the bug

most of the images doesn't appear correctly
Lami's head fucking gone
where's paul
liam's white with blood

### Expected behavior

lami
lami 3
lami 2
https://www.mediafire.com/file/l8cf84j17brfopa/487956_Lami.zip/file

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

Mozilla Firefox

### Additional information

it's on newgrounds

Herschel:
1) Previously we were removing the first occurrence of the invalid 0xFFD9FFD8 byte sequence in JPEG data, but this would break the JPEG if it happened to contain this byte sequence elsewhere (for example, EXIF data). Instead, properly parse the JPEG markers searching for the invalid marker sequence. Fixes #8775.

2) Sometimes JPEGs can be lacking the final 0xFFD9 EOI marker that marks the end of the image. I suspect some JPEG optimizers might remove this to save a few bytes. Flash and most JPEG viewers still display this, but `jpeg-decoder` errors. Glue on the EOI marker if its missing to convince `jpeg-decoder` to decode these cases. Fixes #4209.

### DIFF
--- a/render/src/utils.rs
+++ b/render/src/utils.rs
@@ -87,7 +87,7 @@ pub fn remove_invalid_jpeg_data(data: &[u8]) -> Cow<[u8]> {
     const SOI: u8 = 0xD8; // Start of image
     const EOI: u8 = 0xD9; // End of image
 
-    if data.starts_with(&[0xFF, EOI, 0xFF, SOI]) {
+    let mut data: Cow<[u8]> = if data.starts_with(&[0xFF, EOI, 0xFF, SOI]) {
         // Common case: usually the sequence is at the beginning as the spec says, so adjust the slice to avoid a copy.
         data[4..].into()
     } else {
@@ -127,6 +127,17 @@ pub fn remove_invalid_jpeg_data(data: &[u8]) -> Cow<[u8]> {
             jpeg_data = jpeg_data.get(payload_len + 2..).unwrap_or_default();
             pos += payload_len + 2;
         }
+    };
+
+    // Some JPEGs are missing the final EOI marker (JPEG optimizers truncate it?)
+    // Flash and most image decoders will still display these images, but jpeg-decoder errors.
+    // Glue on an EOI marker if its not already there and hope for the best.
+    if data.ends_with(&[0xFF, EOI]) {
+        data
+    } else {
+        log::warn!("JPEG is missing EOI marker and may not decode properly");
+        data.to_mut().extend_from_slice(&[0xFF, EOI]);
+        data
     }
 }
 
--- a/render/src/utils.rs
+++ b/render/src/utils.rs
@@ -65,23 +65,68 @@ pub fn glue_tables_to_jpeg<'a>(
 }
 
 /// Removes potential invalid JPEG data from SWF DefineBitsJPEG tags.
-///
-/// SWF19 p.138:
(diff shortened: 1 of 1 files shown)


torokati44:
With this new header parser, can it be easily detected if the data is too short? Then, say, extend it with zeroes to the appropriate length?
If so, that would also fix https://github.com/ruffle-rs/ruffle/issues/4209, and most likely https://github.com/ruffle-rs/ruffle/issues/2435 and https://github.com/ruffle-rs/ruffle/issues/720 too.

Herschel:
This code explicitly bails out before the image data (SOS marker), so I'm not sure if it can help with those cases easily. I'll look into it!

Herschel:
The z0r cases seems to simply be missing the final EOI marker, and gluing it on fixes them. I think this is safe to do, worst case the image still fails to decode.

CUB3D (approved):

render/src/utils.rs
```diff
@@ -65,23 +65,79 @@ pub fn glue_tables_to_jpeg<'a>(
...
+    // See https://github.com/ruffle-rs/ruffle/issues/8775 for various examples.
+
+    // JPEG markers
+    const SOF0: u8 = 0xC0; // Start of frame
+    const RST0: u8 = 0xD0; // Restart (we shouldn't see this before SOS, but just in case)
+    const RST7: u8 = 0xD7;
+    const SOI: u8 = 0xD8; // Start of image
+    const EOI: u8 = 0xD9; // End of image
+
+    let mut data: Cow<[u8]> = if data.starts_with(&[0xFF, EOI, 0xFF, SOI]) {
```
relrelb, line 90: nit: `strip_prefix` nicely fits here.

torokati44:
Oh, sorry, I mentioned https://github.com/ruffle-rs/ruffle/issues/2435 and https://github.com/ruffle-rs/ruffle/issues/720 above, but I misremembered, those are completely unrelated. :no_mouth:
And #4209 is indeed fixed.

render/src/utils.rs
torokati44: I think once this tiny thing is addressed, this should be good to go?

torokati44: See: #9153

Issue #8745, opened 2022-12-12
Sweet Drmzzz: Invisible sugar in "Sugar, Sugar" minigame


n0samu:
### Describe the bug

Link to SWF: https://web.archive.org/web/20210111155256id_/https://www.bartbonte.com/sweetdrmzzz/sweetdrmzzz.swf

In [Sweet Drmzzz](https://web.archive.org/web/20210127062810/https://bartbonte.com/sweetdrmzzz/), there is a minigame based on the developer's *Sugar, Sugar* game series. But in the levels with this minigame, the sugar is invisible, making the levels much more difficult than intended to complete.

Steps to reproduce:
1. Run the SWF file in the Ruffle desktop app.
2. Play through the first level and start the second level.
3. Draw a line as shown ingame.

### Expected behavior

Sugar is supposed to fall from the spaceship into the hole at the top of the planet, just like it falls into the cups in *Sugar, Sugar*.
![image](https://user-images.githubusercontent.com/71368227/206986275-712f698e-f3a8-45a8-a39d-e25ce48962ee.png)

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

It seems like the cause of the problem is that `BitmapData.threshold` is not implemented:
`ERROR ruffle_core::avm2::activation] ReferenceError: Error #1069: Property threshold not found on flash.display.BitmapData and there is no default value.`
Here is the relevant line of code in the game's `Main` source file (decompiled with FFDEC):
`this.finalBmp.threshold(this.suikerBmp,_loc6_,_loc7_,"==",this.backkleur,_loc4_,_loc5_,true);`

n0samu:
The same problem also affects [Sugar, Sugar 3](http://www.notdoppler.com/sugarsugar3.php).

MartySVK:
Fixed by https://github.com/ruffle-rs/ruffle/pull/10055

Issue #7958, opened 2022-09-12
retro unicorn attack: Stuck in Black screen.


p0008874:
### Describe the bug

https://user-images.githubusercontent.com/75534590/189677450-117d4dae-9e3c-4685-9662-a5475d82a292.mp4

Error Log Nightly 2022-09-21
```
[2022-09-21T02:49:19Z ERROR ruffle_core::display_object::movie_clip] Unexpected PlaceObject during goto: PlaceObject { version: 2, action: Modify, depth: 1, matrix: Some(Matrix { a: Fixed16(196608), b: Fixed16(0), c: Fixed16(0), d: Fixed16(196608), tx: Twips(2140), ty: Twips(20) }), color_transform: None, ratio: None, name: None, clip_depth: None, class_name: None, filters: None, background_color: None, blend_mode: None, clip_actions: None, has_image: false, is_bitmap_cached: None, is_visible: None, amf_data: None }
[2022-09-21T02:49:23Z ERROR ruffle_core::avm2::activation] AVM2 error: Cannot get undefined property "clone" on class Some(ClassObject { name: flash.display::BitmapData, ptr: 0x1e37d57a300 })
[2022-09-21T02:49:23Z ERROR ruffle_core::avm2::activation] AVM2 error: Cannot get undefined property "clone" on class Some(ClassObject { name: flash.display::BitmapData, ptr: 0x1e37d57a300 })
[2022-09-21T02:49:23Z ERROR ruffle_core::avm2::activation] AVM2 error: Cannot get undefined property "clone" on class Some(ClassObject { name: flash.display::BitmapData, ptr: 0x1e37d57a300 })
[2022-09-21T02:49:23Z ERROR ruffle_core::avm2::activation] AVM2 error: Cannot get undefined property "clone" on class Some(ClassObject { name: flash.display::BitmapData, ptr: 0x1e37d57a300 })
[2022-09-21T02:49:23Z ERROR ruffle_core::avm2::activation] AVM2 error: Cannot get undefined property "clone" on class Some(ClassObject { name: flash.display::BitmapData, ptr: 0x1e37d57a300 })
[2022-09-21T02:49:23Z ERROR ruffle_core::avm2::activation] AVM2 error: Cannot get undefined property "clone" on class Some(ClassObject { name: flash.display::BitmapData, ptr: 0x1e37d57a300 })
[...]

Issue #3086, opened 2021-02-04
Toxic II is softlocked in the first room


ItsMaximum:
[Toxic 2](https://github.com/ruffle-rs/ruffle/files/5926451/toxic_2.zip) is a flash game by nitrome that loads completely fine in ruffle, but the game softlocks in the first room due to certain game mechanics not working. Bombs don't deal any damage to the player, and the door doesn't open after hitting the computer panel with a bomb. I am not sure if this is due to a limitation with ruffle or the heavily obfuscated nature of the code.

n0samu:
Seeing this message when a bomb explodes:
`[WARN  ruffle_core::avm1::globals::bitmap_data] BitmapData.threshold - not yet implemented`
I think this may be the reason for these problems, but I'm not sure.

CUB3D:
After taking a deeper look into this, I think the last missing piece of this is `BitmapData.draw` with blend modes (specifically "difference"). 

In `Player::checkPickUps` it seems to use this combined with `getColorBoundsRect` to decide if something should be "collected". Hitting the computer with a bomb eventually calls `FirstInfoPoint::endCollect` which checks to see if the computer has been "collected" as well as "hacked" (the "hacked" flag is set earlier by this functions caller, `FirstInfoPoint::hack`). It looks like the threshold pr won't do much to help here, but it might be needed at a later point.

n0samu:
That makes sense. There are a lot of other games that need BitmapData.draw with blend modes, such as #8920.

waspennator:
Testing on the latest desktop nightly, I can confirm that you can kill yourself with your own bombs and bombing the computer now opens up the first door too.

There seems to be a new issue now, bombing yourself should only take off a chunk of HP, but in Ruffle it just straight up kills you playing the animation that plays when you fall into acid. Same goes for when you collide with an enemy robot.

Lord-McSweeney:
A minor issue that I noticed: the "sound fx" and "music" buttons are grayed out, although I'm not sure if that behavior occurs in Flash Player or not.

waspennator:
Tested out further after discussing with another user on the ruffle discord server, on level 2, you take damage as normal and you don't seem to die from everything instantly, but it goes right back to insta death on level 3. Then on level 4 your bombs cant hurt you. I can't test any further past that cause you can't get past a destructible tile wall that's normally bombable on flash, but on ruffle just leaves a solid green texture.

n0samu:
[...]

PR #8822, opened 2022-12-18
AVM1: Implement BitmapData.threshold

CUB3D:
This implements `BitmapData.threshold` for AVM1 (although the same implementation can likely be used in AVM2 as well).

This should help futher #3086, although it doesn't appear to fix the softlock on its own.
When the avm2 side is added, this should help with #8745 and #7958


core/src/bitmap/bitmap_data.rs
```diff
@@ -905,6 +906,144 @@ impl<'gc> BitmapData<'gc> {
...
+        // Pre-compute the masked threshold
+        let masked_threshold = threshold & mask;
+
+        // Define the test that will be performed for each pixel
+        match operation.to_utf8_lossy().as_ref() {
+            "==" => self.threshold_internal(
+                source_bitmap,
+                src_rect,
+                dest_point,
+                |v| v == masked_threshold,
```
adrian17, line 992: if instead you did `|v, threshold| v == threshold` , you wouldn't need the generic, right?

adrian17: In fact, you could do
```
let operation = match (...) {
    "==" => |v, threshold| v == threshold,
    ...
```
and deduplicate all these calls, right?

CUB3D: You can indeed, I forgot that non-capturing closures let you do that.

Dinnerbone (approved):

Issue #8827, opened 2022-12-18
1 day - black screen


sodscg:
### Describe the bug

just a black screen

`ERROR ruffle_core::avm2::events] Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x17392c4a840 }) to handler FunctionObject(FunctionObject(GcCell(Gc { ptr: 0x173929675b8 }))) : RustError("Cannot access property ::Update of null or undefined")`

### Expected behavior

game should work

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

[1day.zip](https://github.com/ruffle-rs/ruffle/files/10254707/1day.zip)

n0samu:
Fixed by #8836. And #9007 allows the New Game button code to get farther. But it still hangs - I'm opening a separate issue for that.

Issue #8766, opened 2022-12-13
Accurate Slapshot: Level select screen does not load


n0samu:
### Describe the bug

Link to SWF: http://i.notdoppler.com/files/accurateslapshot.swf
Note: Requires a stub of `SharedObject.size`.

In [Accurate Slapshot](http://www.notdoppler.com/accurateslapshot.php), there are 2 related issues:
- Clicking the Continue button on the main menu goes to a black screen.
- Clicking the X button in the top-right corner during gameplay softlocks the game.
![image](https://user-images.githubusercontent.com/71368227/207266942-9edf78e7-26c2-4bed-992e-04888df4f6e7.png)

Both buttons are supposed to go to the level select screen. And in both cases the following error is shown in the terminal:
`[ERROR ruffle_core::avm2::events] Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x29adda9bf20 }) to handler FunctionObject(FunctionObject(GcCell(Gc { ptr: 0x29ac48bac88 }))) : RustError("Cannot access property ::length of null or undefined")`

### Expected behavior

Both buttons should take you to the level select screen.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

Both problems are caused by code in the game's `TweenMax` package. Here is an SWF with many trace statements added to demonstrate: [accurateslapshot_trace.zip](https://github.com/ruffle-rs/ruffle/files/10216185/accurateslapshot_trace.zip)

After clicking the Continue button, the `showLevelManager()` function runs:
```
         if(contains(this._menu))
         {
            trace("About to run killChildTweensOf");
            TweenMax.killChildTweensOf(this._menu,true);
            trace("Just ran killChildTweensOf");
            removeChild(this._menu);
         }
```
Notice that in Ruffle, the first trace is shown right before the error, and the second trace never shows.

The `killChildTweensOf` function calls `getAllTweens()`:
```
         trace("About to run getAllTweens");
         var _loc3_:Array = getAllTweens();
         trace("Just ran getAllTweens");
```
Notice that in Ruffle, the first trace is shown right before the error, and the second trace never shows.

And finally in `getAllTweens`, we have the offending use of `length`:
```
         var _loc4_:Array = null;
         var _loc5_:int = 0;
         var _loc1_:Dictionary = masterList;
         var _loc2_:int = 0;
         var _loc3_:Array = [];
         for each(_loc4_ in _loc1_)
         {
[...]

PR #8836, opened 2022-12-19
AVM2: Stub flash.net.registerClassAlias

CUB3D:
This lets #8827 get to the menu, however clicking "new game" hangs, possibly due to the same issue as #8766


Herschel (approved):
Thank you!

PR #8840, opened 2022-12-19
web: Add CORS panic error message with FAQ link

n0samu:
Selfhosted builds of Ruffle will now detect CORS errors when loading Ruffle files or SWF files, and direct webmasters to the appropriate FAQ entry (which I just created).
This commit also updates the download link for the Ruffle desktop app to point to the Ruffle website.


Herschel (approved):
Thank you!

PR #8846, opened 2022-12-19
web: Don't unnecessarily borrow-mut instance on mouse-down

adrian17:
Fixes panic screen not popping up when panicking code is triggered by mouse click.


Herschel:
Thank you!

Issue #8855, opened 2022-12-19
lots of lag


ActionWavele:
### Describe the bug

Full moon lags a lot
![bont](https://user-images.githubusercontent.com/120395932/208501159-8a658a0b-ce09-4634-96bd-d8f3bdd9eb3b.jpg)

### Expected behavior

The game is normally really smooth

### Affected platform

Desktop app

### Operating system

Windows 10 22H2 update

adrian17:
Can you please link or attach the SWF in question?

ActionWavele:
here's a link
https://www.maxgames.com/game/full-moon.html

Lord-McSweeney:
`DoAction tag in AVM2 movie` spammed in console
`Property unload not found on flash.display.Loader and there is no default value.` is shown. This error appears in a lot of SWFs.

n0samu:
@ActionWavele what do you mean it "lags"? It's stuck on a black screen for me.

ActionWavele:
I couldn't attach a swf earlier, so I found a link to the game instead
probably didn't match the one I was ising, whi h loads into Ruffle, but the perfotmance is truly awful
like maybe 8 FPS
(would be handy if ruffle could tell me)

n0samu:
@ActionWavele please zip the SWF file and upload that. On Windows, you can just right-click the SWF file and click "Send to => Compressed (zipped) folder".

ActionWavele:
Oh, I never got back to this

ActionWavele:
[fullmoon.zip](https://github.com/ruffle-rs/ruffle/files/10416319/fullmoon.zip)
this should be the swf

waspennator:
Trying it out on desktop on my end, it doesn't seem too laggy on my end. I did have core temps up to check while running through the game. Did notice on desktop the game seems to drop frame wise when moving the rocks around on level 6 and level 7 with the wheel. My Comp's cpu is a AMD Ryzen 5 5600H and here's the general temps I noticed while running this game on ruffle and flash.

On Ruffle: Mostly hung around the low to mid 60's, would jump to the 70's during level 6 and 7

On Flash: Mostly hung around the low 50's, would jump up to the 60's during level 6.

ActionWavele:
this has since gotten better.
still laggy and slow, but slightly more playable.

Lord-McSweeney:
Seems like the lag is a combination of rendering (~35%) and scripts (~65%).

Issue #8851, opened 2022-12-19
Cardinal Quest -  Black Screen


sodscg:
### Describe the bug

```
ERROR ruffle_core::loader] Error during movie loading: FetchError("HTTP status is not ok, got 404 Not Found")
ERROR ruffle_core::avm2::activation] AVM2 error: Object(ErrorObject(ErrorObject(GcCell(Gc { ptr: 0x168464d5a18 }))))
ruffle_core::avm2::activation] ReferenceError: Error #1069: Property unload not found on flash.display.Loader and there is no default value.
ERROR ruffle_core::avm2::activation] AVM2 error: Object(ErrorObject(ErrorObject(GcCell(Gc { ptr: 0x168464d5a18 }))))
ERROR ruffle_core::avm2::activation] ReferenceError: Error #1069: Property unload not found on flash.display.Loader and there is no default value.
ERROR ruffle_core::avm2::activation] AVM2 error: Object(ErrorObject(ErrorObject(GcCell(Gc { ptr: 0x168464d5a18 }))))
ERROR ruffle_core::avm2::activation] ReferenceError: Error #1069: Property unload not found on flash.display.Loader and there is no default value.
ERROR ruffle_core::timer] Unhandled AVM2 error in timer callback: AvmError(Object(ErrorObject(ErrorObject(GcCell(Gc { ptr: 0x168464d5a18 })))))
```

### Expected behavior

the game should load

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

[cardinal-questMTQyOA==_0.zip](https://github.com/ruffle-rs/ruffle/files/10261793/cardinal-questMTQyOA._0.zip)

Lord-McSweeney:
`ReferenceError: Error #1069: Property unload not found on flash.display.Loader and there is no default value.` as well as some CORS issues.

n0samu:
I'm also seeing a huge number of symbol class assignment warnings like this:
`[WARN  ruffle_core::display_object::movie_clip] Got AVM2 error AvmError(Object(ErrorObject(ErrorObject(GcCell(Gc { ptr: 0x147c4de23c8 }))))) when attempting to assign symbol class cq.FontDungeon`

Lord-McSweeney:
Not fixed as of latest nightly.

Lord-McSweeney:
New error is `RustError("VerifyError: Trait color in class HxlText has same name as trait color in class HxlSprite, but does not override it")`.

[...]

PR #8871, opened 2022-12-19
Stub flash.display.Loader.unload

Lord-McSweeney:
Stub `flash.display.Loader.unload`.


n0samu:
Hello and welcome! I'm curious if there's a piece of content that this fixes are helps along?

Lord-McSweeney:
I think #8851 and #8855 both use MochiCrypt, which relies on `Loader.unload`. I'm not sure exactly how this pull request might make it better (cargo does not run for me), but I'm hoping it might help.

n0samu:
Sadly it didn't help either of those cases for me.

PR #8933, opened 2022-12-27
avm2: Fix hitTestPoint not transforming coordinates to world space

MrCheeze:
This behaviour is essentially copied from [avm1 movie_clip.rs's hit_test.](https://github.com/ruffle-rs/ruffle/blob/664c67ea27ac98337fb7cb01b976c3b34cdd2164/core/src/avm1/globals/movie_clip.rs#L200-L204)

Several AVM2/AS3 flashes I was testing had broken collision previously, but are fixed by this.

As an example, https://mrcheeze.github.io/homestuck-ruffle-tester/#03435 does not let you move before the change, because the hitTestPoint thinks that you're trapped inside a wall. Clicking some of the UI buttons didn't work either. Both issues are solved with this change.


n0samu:
Thanks so much for this! Unfortunately this is causing some regressions in other hitTest tests. I guess a correct fix must be more complicated :(

MrCheeze:
[Changed the logic for the case where avm2_root is None](https://github.com/ruffle-rs/ruffle/compare/6b54efc574fa6e702c583e84d0a1dfb7eb8eb0b1..4e2d90f5388e2e990d30a4424e10029e59ed0137). Not sure how confident to be in the change after my first attempt failed, but the flashes I was testing work correctly and the regression tests pass with this version.

n0samu (approved):
Seems like a logical fix and all the tests pass, so it gets an approve from me! Thanks again for the contribution! 😃

MrCheeze:
[Tried porting some existing AVM1 tests to AVM2 for the purpose of testing this branch.](https://github.com/MrCheeze/ruffle/commit/7a49f239b707f0d99690727d1c02e9313c109389)

Several tests fail even with my changes, but considerably less than were failing before...

Dinnerbone:
Looks good to me, and more correct is better than not more correct!

PR #8935, opened 2022-12-27
avm2: Add ctrlKey, altKey, shiftKey, controlKey properties to KeyboardEvent

MrCheeze:
Based on [similar logic](https://github.com/ruffle-rs/ruffle/blob/664c67ea27ac98337fb7cb01b976c3b34cdd2164/core/src/avm2/object/event_object.rs#L127-L136) already existing in MouseEvent.

I also included controlKey as always having the same value, which [is not strictly correct](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/events/KeyboardEvent.html) (there are differences on Mac), but I figured was better than leaving it out entirely. I did not include commandKey at all, though.

### DIFF
--- a/core/src/player.rs
+++ b/core/src/player.rs
@@ -948,6 +948,10 @@ impl Player {
                 if let PlayerEvent::KeyDown { key_code, key_char }
                 | PlayerEvent::KeyUp { key_code, key_char } = event
                 {
+                    let ctrl_key = context.input.is_key_down(KeyCode::Control);
+                    let alt_key = context.input.is_key_down(KeyCode::Alt);
+                    let shift_key = context.input.is_key_down(KeyCode::Shift);
+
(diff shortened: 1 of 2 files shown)


core/src/player.rs
```diff
@@ -965,6 +969,11 @@ impl Player {
                                 false.into(),                            /* cancelable */
                                 key_char.map_or(0, |c| c as u32).into(), /* charCode */
                                 (key_code as u32).into(),                /* keyCode */
+                                0.into(),                                /* keyLocationValue */
```
Aaron1011, line 972: Can you add a FIXME comment here?

MrCheeze: Added a TODO for that, for commandKey, and for refining ctrlKey.

PR #8939, opened 2022-12-27
avm2: Fix calling setTextFormat without providing a beginIndex and endIndex

MrCheeze:
Previous behaviour defaulted to undefined and applied the format to the range [0,0). The default should be -1, which applies the format to the full length of the TextField.

This fixes the text color in the windows that pop up when you click objects in [this flash](https://mrcheeze.github.io/homestuck-ruffle-tester/#00253). (Currently the text is displaying black and unreadable, with this fix it changes to white like it's supposed to.)

### DIFF
--- a/core/src/avm2/globals/flash/text/textfield.rs
+++ b/core/src/avm2/globals/flash/text/textfield.rs
@@ -840,11 +840,11 @@ pub fn set_text_format<'gc>(
             if let Some(tf) = tf.as_text_format() {
                 let mut begin_index = args
                     .get(1)
-                    .unwrap_or(&Value::Undefined)
+                    .unwrap_or(&(-1).into())
                     .coerce_to_i32(activation)?;
                 let mut end_index = args
                     .get(2)
-                    .unwrap_or(&Value::Undefined)
+                    .unwrap_or(&(-1).into())
                     .coerce_to_i32(activation)?;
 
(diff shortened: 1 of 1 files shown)


core/src/avm2/globals/flash/text/textfield.rs
```diff
@@ -838,14 +838,14 @@ pub fn set_text_format<'gc>(
...
-                    .unwrap_or(&Value::Undefined)
-                    .coerce_to_i32(activation)?;
+                let mut begin_index = match args.get(1) {
+                    Some(value) => value.coerce_to_i32(activation)?,
+                    None => -1,
+                };
+                let mut end_index = match args.get(2) {
+                    Some(value) => value.coerce_to_i32(activation)?,
+                    None => -1,
+                };
```
Bale001, line 848: I think that doing this is cleaner:
```rs
let mut begin_index = args
                    .get(1)
                    .unwrap_or(&-1.into())
                    .coerce_to_i32(activation)?;
let mut end_index = args
                    .get(2)
                    .unwrap_or(&-1.into())
                    .coerce_to_i32(activation)?;
```

MrCheeze: You are right, updated.

Bale001 (approved):

Issue #5926, opened 2021-12-25
[Scooby-Doo: Curse of Anubis - Pyramid of Doom!] The game cannot proceed because the character is stuck


Jmousy:
### Describe the bug

![scooby-doo](https://user-images.githubusercontent.com/75052171/147377904-3943f593-bd0f-4511-9ef0-31c036803139.gif)

When you start the game for the first time in **'Scooby-Doo: Curse of Anubis - Pyramid of Doom!'**, your character moves forward.
However, the Ruffle just keeps walking in place and the keyboard input is stuck.

The gif above is what happens when do nothing.

Ruffle Nightly 2021-12-22

[Scooby-Doo-Curse-of-Anubis-Pyramid-of-Doom.zip](https://github.com/ruffle-rs/ruffle/files/7775489/Scooby-Doo-Curse-of-Anubis-Pyramid-of-Doom.zip)

### Expected behavior

The character should initially move forward one step and then move by keyboard input.

### Affected platform

Self-hosted version

### Operating system

Windows 10 21H2 x64

### Browser

Google Chrome 96

Toad06:
Thanks for the report.

The issue is in `\DefineSprite (930)\frame 1\PlaceObject2 (745) Depth: 380 (Pers)\onClipEvent(enterFrame)`:

```
onClipEvent(enterFrame){
   if(_global.C[Nom].CanMove == true)
   {
      _global.C[Nom].Move();
   }
}
```
The variable `Nom` doesn't hold the same value in Flash Player and Ruffle:
- "Pers" in Flash.
- "undefined" in Ruffle.

---

The variable `Nom` is defined in `\DefineSprite (930)\frame 1\PlaceObject2 (745) Depth: 380 (Pers)\onClipEvent(load)`:
```
onClipEvent(load){
   with(_parent._parent._parent)
   {
      var Nom = this._name
      ...
```

Adding a line `var Nom = this._name;` outside of the `with` statement appears to be a workaround for Ruffle:
```
onClipEvent(load){
   var Nom = this._name;
   with(_parent._parent._parent)
   {
      var Nom = this._name
      ...
```

Toad06:
Fixed by #8953

Issue #3119, opened 2021-02-06
Player got Stuck after jump or move in Megaman vs Ghosts'n Goblins.


Jmousy:
**Describe the bug**
### [Issue A] Jump and Stuck
![megaman](https://user-images.githubusercontent.com/75052171/107117431-a1f18600-68bd-11eb-906d-058ad3ebd6c7.gif)

In Megaman vs Ghosts'n Goblins, the Arrow Key works without problems until you hit the jump (keyboard Z key).
**But After pressing the jump key, the following problem occurs.**

1. The player's position is fixed (stuck) at the last point and does not move.
2. Even if the player jumps or slides, it returns to where it stopped.

The player literally got stuck in the same spot after the jump. You can check the detailed symptoms through the gif at the top.

### [Issue B] Move and Stuck
![megaman-move](https://user-images.githubusercontent.com/75052171/107117787-06ade000-68c0-11eb-82a9-cb2a22346d78.gif)
(Graphics may look broken while reducing the gif size. This is not a problem with Ruffle.)

This time, simply press the Right Arrow Key to reproduce. The character seems to move well, but suddenly comes back to the starting point and gets stuck.

### Others...
Considering the results of both problems are the same, I think it is probably the same problem, but I was unable to determine where the problem was due to my inability to decompile the SWF. So, I'll also attach a file that might help.

Thank you for the hard work of the Ruffle team.

[megaman-vs-ghostsngoblins.zip](https://github.com/ruffle-rs/ruffle/files/5936788/megaman-vs-ghostsngoblins.zip)

**Is the problem with the Ruffle desktop app, extension, or self-hosted version?**
Probably happens on all platforms

**What platform are you using?**
Desktop: Windows 10 x64, Google Chrome (Latest), Ruffle Nightly 2021-02-06

n0samu:
Can confirm that these problems are still occurring!

adrian17:
Found the cause, https://github.com/ruffle-rs/ruffle/issues/3119 . (for the record, the variable is `newPos`)

Toad06:
Fixed by #8953

Issue #1550, opened 2020-11-13
Freezing the app allows you to go to next levels in Bubble Gum Run


Toad06:
When you complete a level in [Bubble Gum Run](https://www.y8.com/games/bubble_gum_run), the following panel should appear:
![panel](https://user-images.githubusercontent.com/14274782/99079358-21720080-25c0-11eb-81e0-f870e3c5967f.png)
In Ruffle, the panel never comes. The function `endLevel2` seems to be used to display it but for some reason this function is never called.

However, there's a very strange "workaround" to go to next levels in this game. Here are the different steps using the desktop build (Windows):
1. Start a new game and wait until the character completely appears on screen.
2. Right click somewhere on the window border (![window](https://user-images.githubusercontent.com/14274782/99079378-2931a500-25c0-11eb-871a-2a864f14cc6b.png)). This will "pause" the game. Wait a few seconds. In the meantime, notice the position of the enemy and the timer's value.
3. Left click anywhere on the game window to "resume" the game. Notice the enemy hasn't moved whereas the timer's value has dropped.
4. Now complete the level, by collecting all the bubble gums and going to the exit door (if you haven't enough time, please restart the game and proceed this step directly). You can wait a few seconds but the panel will not be displayed.
5. Repeat steps 2 and 3. The panel now appears (if it doesn't, repeat this step).
6. Go to the next level and repeat step 2.
7. Left click anywhere on the game window to "resume" the game. Notice that the panel unexpectely appears and that the timer's value hasn't dropped this time.
8. Steps 6 and 7 can be repeated for the next levels.

On the Web backend, right-clicking on the window border has no effect; instead you can switch between tabs.

Toad06:
The game still doesn't work but #6082 made the main issue (freezing Ruffle to go to the next level) impossible to reproduce. The panel still doesn't appear though, so it's now completely impossible to go to next levels. Pending on #6164 resolution to see if that changes anything.

Toad06:
Fixed by #8953

PR #8953, opened 2022-12-28
AVM1: Fix local variable definitions in with() scopes

### LINKED ISSUE #6164: avm1: Local variables defined inside `with` statements are incorrectly added as properties of the mentioned object
[Sample](https://github.com/ruffle-rs/ruffle/files/7964084/localwith.zip)

```
var o = {};
o.h = "";
with(o)
{
   h = "HELLO";
   e = " ";
   var w = "WORLD";
   trace(h + e + w);
};
trace(o.h + e + w);
trace(o.h + e + o.w);
```

Flash Player traces:

> HELLO WORLD
> HELLO WORLD
> HELLO undefined

Ruffle traces:

> HELLO WORLD
> HELLO undefined
> HELLO WORLD

It seems like the behaviour has always been wrong in Ruffle.

CUB3D:
In Flash Player, defining a local variable when in a with scope should actually define it on the first non-with parent scope, see #6164 for an example.

This fixes #6164, both issues in #3119 and the missing menu in #1550


Toad06:
Also fixes #5926. 👍

Dinnerbone (approved):

PR #8956, opened 2022-12-29
core: Make AVM2 frame labels case sensitive

MrCheeze:
Straightforward, other than needing to pass around the context to know if we're in an AS3 flash or not.

Apart for allowing for "LabelA" and "labela" to be different from each other, this _also_ fixes an issue where MovieClip.currentLabel always returned a lowercase string, which broke flashes that expected it to return a string with capitals, [such as this one.](https://mrcheeze.github.io/homestuck-ruffle-tester/#03318)


Lord-McSweeney:
This looks like this causes a regression in #9018: the game stops at the loading screen. In the console: "`ruffle_core::display_object::movie_clip] Error occured when running AVM2 frame script: RustError("Cannot access property ::addEventListener of null or undefined")`"

MrCheeze:
> This looks like this causes a regression in #9018: the game stops at the loading screen. In the console: "`ruffle_core::display_object::movie_clip] Error occured when running AVM2 frame script: RustError("Cannot access property ::addEventListener of null or undefined")`"

See my comment [here](https://github.com/ruffle-rs/ruffle/issues/9018#issuecomment-1374138308) - this change isn't what broke the movie, and I identified what did.

Dinnerbone (approved):

Aaron1011 (approved):

PR #8959, opened 2022-12-29
docs: Document Java in the web version README, too

kmeisthax:
Our build process has required Java for a while, but we don't document this requirement as good as we could. This is a problem for people who need to build from scratch in a container (such as, say, extension reviewers).


web/README.md
```diff
@@ -46,6 +46,13 @@ rust using the above instructions).
 
 For the compiler to be able to output WebAssembly, an additional target has to be added to it: `rustup target add wasm32-unknown-unknown`
 
+#### Java
+
+Follow the instructions [to install OpenJDK](https://openjdk.org/projects/jdk/19/) on your machine.
```
adrian17, line 51: we only require headless JRE (not full JDK), right? Can mention that too.
(for example, on my Ubuntu, `openjdk-8-jre-headless` is enough)

PR #8964, opened 2022-12-30
core: Replace some derived Debug impls with manual ones

moulins:
In future versions of `gc-arena`, the `Debug` impl. of `Gc` and `GcCell` will print the pointed-to value, which will cause derived `Debug` impls. to enter an infinite recursion.

As such, this manually implements `Debug` on types wrapping a `Gc/GcCell` to maintain the current behavior.

See #8625 for previous discussion.


core/src/avm1/object/bitmap_data.rs
```diff
@@ -12,23 +12,25 @@ use std::fmt;
...
+        let data = this.data.read();
         f.debug_struct("BitmapData")
-            .field("data", &this.data)
+            .field("ptr", &self.0.as_ptr())
+            .field("data", &data)
             .finish()
     }
 }
 
+#[derive(Clone, Collect)]
```
Aaron1011, line 26: Could you move this back to its original location, to shrink the diff?

core/src/avm1/object/sound_object.rs
```diff
@@ -44,17 +56,6 @@ pub struct SoundObjectData<'gc> {
     is_streaming: bool,
 }
 
-impl fmt::Debug for SoundObject<'_> {
```
Aaron1011, line 47: This can also be moved back to its original position in the file

core/src/avm1/object/bitmap_data.rs
moulins: Fixed

core/src/avm2/object.rs
```diff
@@ -93,7 +93,7 @@ pub use crate::avm2::object::xml_object::{xml_allocator, XmlObject};
 /// runtime.
 #[enum_trait_object(
     #[allow(clippy::enum_variant_names)]
-    #[derive(Clone, Collect, Debug, Copy)]
+    #[derive(Clone, Collect, Copy, Debug)]
```
Aaron1011, line 96: Can you change the order back?

core/src/avm2/value.rs
```diff
@@ -57,6 +58,20 @@ static_assertions::assert_eq_size!(Value<'_>, [u8; 16]);
 #[cfg(target_pointer_width = "64")]
 static_assertions::assert_eq_size!(Value<'_>, [u8; 24]);
 
+impl fmt::Debug for Value<'_> {
```
Aaron1011, line 61: I don't think a manual impl is needed now - this can become a `#[derive(Debug)]` again

core/src/display_object.rs
```diff
@@ -601,7 +601,7 @@ pub fn render_base<'gc>(this: DisplayObject<'gc>, context: &mut RenderContext<'_
 }
 
 #[enum_trait_object(
-    #[derive(Clone, Collect, Debug, Copy)]
+    #[derive(Clone, Collect, Copy, Debug)]
```
Aaron1011, line 604: This can also be re-ordered

core/src/avm2/multiname.rs
```diff
@@ -14,7 +14,7 @@ use swf::avm2::types::{
     AbcFile, Index, Multiname as AbcMultiname, NamespaceSet as AbcNamespaceSet,
 };
 
-#[derive(Clone, Copy, Debug, Collect)]
+#[derive(Clone, Copy, Collect, Debug)]
```
Aaron1011, line 17: The derive orders in this file can also be swapped back

core/src/avm2/value.rs
moulins: A `#derive(Debug)]` would print the more verbose `Object(FooObject { ... })` instead of `FooObject`, but if you don't think this is worth it I can change it back

core/src/avm1/object/stage_object.rs
```diff
@@ -34,6 +34,16 @@ pub struct StageObjectData<'gc> {
     text_field_bindings: Vec<TextFieldBinding<'gc>>,
 }
 
+impl fmt::Debug for StageObject<'_> {
```
Aaron1011, line 37: Can you move this back to its original position?

core/src/avm1/object.rs
```diff
@@ -64,7 +64,7 @@ pub enum NativeObject<'gc> {
 /// runtime.
 #[enum_trait_object(
     #[allow(clippy::enum_variant_names)]
-    #[derive(Clone, Collect, Debug, Copy)]
+    #[derive(Clone, Collect, Copy, Debug)]
```
Aaron1011, line 67: This order can also be swapped

core/src/avm2/value.rs
Aaron1011: I would prefer not to special-case the `Object` case like this (since the `Value::String` case gets printed with a wrapping `String`).

Aaron1011 (approved):

PR #8969, opened 2022-12-31
Remove & disallow unwraps in various crates

Dinnerbone:
Part of the "let's stop panicking" effort. Absolutely not exhaustive for the whole codebase, but this removes all use of unwraps and disallows them going forwards, in the following crates:
- render
- render/webgl
- render/canvas
- web
- desktop
- video

### DIFF
--- a/render/canvas/src/lib.rs
+++ b/render/canvas/src/lib.rs
@@ -1,3 +1,5 @@
+#![deny(clippy::unwrap_used)]
+
 use gc_arena::MutationContext;
 use ruffle_render::backend::null::NullBitmapSource;
 use ruffle_render::backend::{
--- a/render/canvas/src/lib.rs
+++ b/render/canvas/src/lib.rs
@@ -146,7 +146,7 @@ impl BitmapHandleImpl for BitmapData {}
 
 fn as_bitmap_data(handle: &BitmapHandle) -> &BitmapData {
     <dyn BitmapHandleImpl>::downcast_ref(&*handle.0)
-        .expect("Bitmap handle must be BitmapHandleImpl")
+        .expect("Bitmap handle must be a Canvas BitmapData")
 }
 
 impl BitmapData {
--- a/render/canvas/src/lib.rs
+++ b/render/canvas/src/lib.rs
@@ -143,7 +143,8 @@ struct BitmapData {
 impl BitmapHandleImpl for BitmapData {}
 
 fn as_bitmap_data(handle: &BitmapHandle) -> &BitmapData {
-    <dyn BitmapHandleImpl>::downcast_ref(&*handle.0).unwrap()
+    <dyn BitmapHandleImpl>::downcast_ref(&*handle.0)
+        .expect("Bitmap handle must be BitmapHandleImpl")
 }
 
 impl BitmapData {
@@ -154,8 +155,8 @@ impl BitmapData {
             ImageData::new_with_u8_clamped_array(Clamped(bitmap.data()), bitmap.width())
                 .into_js_result()?;
 
-        let window = web_sys::window().unwrap();
-        let document = window.document().unwrap();
+        let window = web_sys::window().expect("window()");
+        let document = window.document().expect("document()");
         let canvas: HtmlCanvasElement = document
             .create_element("canvas")
             .into_js_result()?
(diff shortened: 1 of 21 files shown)


n0samu:
Looking at the open issues, we have:
- 121 panics related to `navigator.rs`
- 91 panics related to `webgl/src/lib.rs`
- 28 panics related to `canvas/src/lib.rs`
- 14 panics related to `web/src/audio.rs`
It'll be interesting to see how many are resolved by this! On the other hand, I guess it might make it harder to catch related issues since users won't be reporting them as often. I guess we should try to triage more of the ones we have now!

Dinnerbone:
@n0samu any chance you have a list of (ideally unique) panics per file like that? I can take a look at the causes

adrian17:
> On the other hand, I guess it might make it harder to catch related issues since users won't be reporting them as often.

FWIW I agree, some of these like `Couldn't register shape` definitely should be something we should know about. It _could_ just so happen that the shape doesn't happen to be seen by user at any point and the game could otherwise work, yes. But I'd still consider it a more critical error than say `"Tried to instantiate non-registered character ID"`, which AFAIK _can_ happen naturally in the wild.

(Of course, I love the general effort of pushing errors to outside layers and adding context :) )

adrian17:
(I'm gonna read through the list more after new year's)

adrian17:
> FWIW I agree, some of these like Couldn't register shape definitely should be something we should know about

Update: I just looked at it more, didn't realize that these can only come from GL operation errors. Though... that also made me realize, looks like we currently completely swallow any errors coming from the lyon tessellator :(

render/webgl/src/lib.rs
```diff
@@ -161,7 +165,8 @@ impl Drop for RegistryData {
 impl BitmapHandleImpl for RegistryData {}
 
 fn as_registry_data(handle: &BitmapHandle) -> &RegistryData {
-    <dyn BitmapHandleImpl>::downcast_ref(&*handle.0).unwrap()
+    <dyn BitmapHandleImpl>::downcast_ref(&*handle.0)
+        .expect("Bitmap handle must be BitmapHandleImpl")
```
Aaron1011, line 169:
```suggestion
        .expect("Bitmap handle must be webgl RegistryData")
```

render/canvas/src/lib.rs
```diff
@@ -142,7 +144,8 @@ struct BitmapData {
 impl BitmapHandleImpl for BitmapData {}
 
 fn as_bitmap_data(handle: &BitmapHandle) -> &BitmapData {
-    <dyn BitmapHandleImpl>::downcast_ref(&*handle.0).unwrap()
+    <dyn BitmapHandleImpl>::downcast_ref(&*handle.0)
+        .expect("Bitmap handle must be BitmapHandleImpl")
```
Aaron1011, line 148:
```suggestion
        .expect("Bitmap handle must be canvas BitmapData")
```

Aaron1011 (changes requested):
A `BitmapHandle` always stores a `BitmapHandleImpl` trait object - the `expect` message should reference the backend-specific type being used as the downcast target.

n0samu:
> @n0samu any chance you have a list of (ideally unique) panics per file like that? I can take a look at the causes

I don't have a list - I got those numbers by searching the open GitHub issues with the `error-report` label for the filename in question, for example:
`is:open is:issue label:error-report web/src/audio.rs`

render/src/shape_utils.rs
```diff
@@ -195,7 +195,7 @@ impl PathSegment {
     fn to_draw_commands(&self) -> impl '_ + Iterator<Item = DrawCommand> {
         assert!(!self.is_empty());
         let mut i = self.points.iter();
-        let first = i.next().unwrap();
+        let first = i.next().expect("Points isn't empty");
```
Aaron1011, line 198:
```suggestion
        let first = i.next().expect("Points is empty");
```

scanner/src/execute.rs
```diff
@@ -21,7 +21,7 @@ fn execute_swf(file: &Path) {
     let frame_time = 1000.0 / movie.frame_rate().to_f64();
     let player = PlayerBuilder::new()
         .with_log(ScanLogBackend::new())
-        .with_navigator(NullNavigatorBackend::with_base_path(base_path, &executor))
+        .with_navigator(NullNavigatorBackend::with_base_path(base_path, &executor).unwrap())
```
Aaron1011, line 24: Should this be an `expect` instead?

Aaron1011 (changes requested):

Aaron1011 (approved):

adrian17:
(FWIW, my point about swallowing errors in shape registration still stands)

PR #8974, opened 2022-12-31
core: Drop Option<> from TDisplayObject::movie()

Dinnerbone:
This removes a bunch of unwraps!

### DIFF
--- a/core/src/avm1/activation.rs
+++ b/core/src/avm1/activation.rs
@@ -2682,7 +2682,7 @@ impl<'a, 'gc, 'gc_context> Activation<'a, 'gc, 'gc_context> {
             level
         } else {
             let level: DisplayObject<'_> =
-                MovieClip::new(self.base_clip().movie().unwrap(), self.context.gc_context).into();
+                MovieClip::new(self.base_clip().movie(), self.context.gc_context).into();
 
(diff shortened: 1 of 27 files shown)


Lord-McSweeney:
Can you remove unwraps from core/src/display_object/container.rs? I can count at least ten of them there.

Dinnerbone:
> Can you remove unwraps from core/src/display_object/container.rs? I can count at least ten of them there.

There's a lot, everywhere :D Trying not to do too many big hitters at once, but I can take a look.

Aaron1011 (approved):
The Clippy failures are unrelated to this PR, and will be fixed by https://github.com/ruffle-rs/ruffle/pull/8975

Aaron1011:
@Dinnerbone Would it make sense to movie the `Arc<SwfMovie>` field to `DisplayObjectBase`, instead of storing it in each `DisplayObject` struct?

kmeisthax (approved):
Looks good to me.

I'm not 100% sold on using `activation.context.swf` in non-movie (i.e. spawned from AS3) contexts, but I doubt this will break anything.

I also concur on moving `Arc<SwfMovie>` to `DisplayObjectBase`; but that can be a separate PR.

PR #8997, opened 2023-01-02
avm2: Fix edge case when setting fullYear on invalid Date object

Bale001:
Fixes the following case:
```as
var d = new Date(NaN);
d.fullYear = 1999;
trace(d);
```
Previously Ruffle gave:
```
Invalid Date
```

However, it turns out Flash actually allows setting the `fullYear` property of an invalid date object, by simply setting it to the first day of that year. So the correct output is actually:
```
Fri Jan 1 00:00:00 GMT-0800 1999
```
(the GMT part might differ depending on your timezone)

This also applies to `fullYearUTC`.


core/src/avm2/globals/date.rs
```diff
@@ -615,7 +611,20 @@ pub fn set_full_year<'gc>(
...
-        let timestamp = DateAdjustment::new(activation, &get_timezone())
+        let timezone = get_timezone();
+        if this.date_time().is_none() {
+            this.set_date_time(
+                activation.context.gc_context,
+                Some(
+                    timezone
+                        .with_ymd_and_hms(0, 1, 1, 0, 0, 0)
+                        .single()
+                        .expect("Unambiguous epoch time when constructing Date")
```
Aaron1011, line 622: Since this is an error message, shouldn't it say "Ambiguous"?

Dinnerbone:
I've pushed a fix for the error messages and we'll try and get this in now

PR #9013, opened 2023-01-04
Bump `gc-arena` to current master, take 2

moulins:
This required small changes to some Debug impls that were missed in #8964.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -12,7 +12,7 @@ byteorder = "1.4"
 bitstream-io = "1.6.0"
 flate2 = "1.0.25"
 fnv = "1.0.7"
-gc-arena = { git = "https://github.com/ruffle-rs/gc-arena" }
+gc-arena = { workspace = true }
 generational-arena = "0.2.8"
 indexmap = "1.9.2"
 tracing = "0.1.37"
--- a/render/wgpu/Cargo.toml
+++ b/render/wgpu/Cargo.toml
@@ -21,7 +21,7 @@ image = { version = "0.24.5", default-features = false }
 ouroboros = "0.15.5"
 typed-arena = "2.0.1"
 once_cell = "1.17.0"
-gc-arena = { git = "https://github.com/ruffle-rs/gc-arena" }
+gc-arena = { workspace = true }
 naga-agal = { path = "../naga-agal" }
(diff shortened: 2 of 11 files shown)


core/Cargo.toml
```diff
@@ -12,7 +12,7 @@ byteorder = "1.4"
 bitstream-io = "1.6.0"
 flate2 = "1.0.25"
 fnv = "1.0.7"
-gc-arena = { git = "https://github.com/ruffle-rs/gc-arena" }
+gc-arena = { git = "https://github.com/ruffle-rs/gc-arena", rev = "3195fbed9f2ba1683fb113e3d7d84794fe3465d5" }
```
danielhjacobs, line 15: Am I missing something or shouldn't this be kyren/gc-arena now?

Aaron1011: You're right - we no longer need our fork.

moulins: Oops, right, good catch :c

render/wgpu/Cargo.toml
```diff
@@ -21,7 +21,7 @@ image = { version = "0.24.5", default-features = false }
 ouroboros = "0.15.5"
 typed-arena = "2.0.1"
 once_cell = "1.17.0"
-gc-arena = { git = "https://github.com/ruffle-rs/gc-arena" }
+gc-arena = { git = "https://github.com/kyren/gc-arena", rev = "3195fbed9f2ba1683fb113e3d7d84794fe3465d5" }
```
Aaron1011, line 24: Can you make this a workspace dependency:
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace

Aaron1011 (changes requested):

render/wgpu/Cargo.toml
moulins: TIL, I didn't know that was possible.

Aaron1011: It's a pretty new feature, but we're using the latest stable Rust, so we can use it :)

Lord-McSweeney:
There are a lot of `Invalid session id` errors in the test results.

Aaron1011 (approved):

Issue #1839, opened 2020-12-04
Learn to fly 2 on armorgames


rustokomuska:
Playing Learn to fly 2 on armorgames.com
The game doesn't have sound after the first seconds of the opening cutscene, and when starting story mode, the sack doesnt move off the ramp like it's supposed to, so the game can't continue and is unplayable. It's also very laggy.
My browser is opera GX . My browser version (according to the opera GX update page) is LVL 2 (core: 71.0.3770.449)
I played it from this link https://armorgames.com/play/11967/learn-to-fly-2?tag-referral=flash
Screenshot of the sack not moving
![image](https://user-images.githubusercontent.com/54962938/101156550-b21c8900-3620-11eb-8fd7-4abc9de7aebb.png)

There are a lot of errors appearing on the opera extensions page, many referring to the sound.
This one says unable to register sound ID

```
Context
https://armorgames.com/play/11967/learn-to-fly-2?tag-referral=flash
Stack Trace
dist/core.ruffle.js:1 (it)
dist/ruffle.js:1 (__wbg_error_7f083efc6bc6752c)
dist/b48c4d894c3aec6bf23e.module.wasm:1 (anonymous function)
dist/b48c4d894c3aec6bf23e.module.wasm:1 (anonymous function)
dist/b48c4d894c3aec6bf23e.module.wasm:1 (anonymous function)
dist/b48c4d894c3aec6bf23e.module.wasm:1 (anonymous function)
dist/b48c4d894c3aec6bf23e.module.wasm:1 (anonymous function)
dist/b48c4d894c3aec6bf23e.module.wasm:1 (anonymous function)
dist/b48c4d894c3aec6bf23e.module.wasm:1 (anonymous function)
dist/b48c4d894c3aec6bf23e.module.wasm:1 (anonymous function)
dist/b48c4d894c3aec6bf23e.module.wasm:1 (anonymous function)
dist/b48c4d894c3aec6bf23e.module.wasm:1 (anonymous function)
dist/b48c4d894c3aec6bf23e.module.wasm:1 (anonymous function)
dist/b48c4d894c3aec6bf23e.module.wasm:1 (anonymous function)
dist/b48c4d894c3aec6bf23e.module.wasm:1 (_dyn_core__ops__function__FnMut__A____Output___R_as_wasm_bindgen__closure__WasmClosure___describe__invoke__h6b81efa5c5c5d026)
dist/core.ruffle.js:1 (k)
[...]

PR #9025, opened 2023-01-04
avm1: Implement BitmapData.draw with blend mode

Dinnerbone:
Mostly copied from the avm2 implementation, with addition of a u8 fallback per spec


Aaron1011 (approved):

n0samu:
Also fixes #1839! 🎉

PR #9026, opened 2023-01-04
Migrate from log to tracing

Dinnerbone:
Only touches core, desktop and web right now.

All functionality remains as before, with the following notable differences:
- Web now allows for different ruffle instances to be configured at different log levels, previously it ignored all but the first setting
- Web will use the browsers performance tab to record timings of tracing spans, but they're debug only right now. A future PR will expand on this.
- Desktop has more options available in the RUST_LOG environment variable than before, but as we don't add much new spans or fields, this isn't useful yet

Screenshot of an example of the performance metrics in firefox:
image

A future PR will add tracy support to desktop, you can see a screenshot of how that looks on discord. This PR is mostly just what's needed to do the switch with little fuss, to avoid dragging out conflicts.

Other planned changes in future PRs:
- Conditionally compile out Trace and maybe Debug output from release builds
- Restrict performance tracking in web builds behind a flag (not needed yet as they're debug anyway, so it won't be hit)
- More instrumentation!


core/src/player.rs
```diff
@@ -1421,6 +1421,7 @@ impl Player {
         })
     }
 
+    #[instrument(level = "debug", skip_all)]
```
adrian17, line 1424: could you also instrument `tick()`? It would mark the the "most uninterrupted non-render cpu time spend in a row"

Dinnerbone: Sure thing!

Aaron1011 (approved):

torokati44:
Oooh, this sounds very useful then, even to profile the video decoders on web more precisely! (Once it's no longer debug-only...)

adrian17:
> Once it's no longer debug-only...

This didn't mean debug builds, it meant debug log level :) So you just need to set the config to enable them.

torokati44:
Even better! \o/

adrian17 (approved):

Issue #8361, opened 2022-10-23
Autoplay behavior needs improvement


n0samu:
**Issue**
The behavior of web builds of Ruffle with `"autoplay": "on"` is less than ideal in several ways:
- An unmute overlay is *always* shown, even if the SWF doesn't play any audio before the first user interaction. [Here is an example.](https://n0samu.github.io/ruffle-testing/?swf=wonderputt.swf&autoplay=1)
- If animation with synced audio starts playing before the audio is unmuted, the animation will pause completely, rather than playing without audio. [Here is an example.](https://n0samu.github.io/ruffle-testing/?swf=ruffled.swf&autoplay=1)
- When the audio is unmuted, all of the sound effects that were supposed to play while the audio was muted will play at once. Notice what happens if you let Alien Hominid play through the intro, then click the unmute overlay. [Here is the link.](https://n0samu.github.io/ruffle-testing/?swf=alienhominid.swf&autoplay=1)

**Improvement**
- With `"autoplay": "on"`, an unmute overlay should only appear if the SWF tries to play audio before the first user interaction.
- SWFs should play as normal even if muted; the only difference should be that the user hears no audio. 
- When the audio is unmuted, only the sound that's supposed to be *currently* playing should play.

ActionWavele:
I thought browsers break autoplay?

Herschel:
The animation stops because the audio syncing code is trying to sync the animation to the non-playing sounds. The easiest fix is for the web audio backend to simply bail out of playing any sounds if it the audio context isn't playing. This would fix the all-sounds-playing-at-once issue on resume, too. The one downside is that any background music won't start playing once resumed.

n0samu:
I just noticed that part of this issue was already described in #6800.

n0samu:
> The animation stops because the audio syncing code is trying to sync the animation to the non-playing sounds. The easiest fix is for the web audio backend to simply bail out of playing any sounds if it the audio context isn't playing. This would fix the all-sounds-playing-at-once issue on resume, too. The one downside is that any background music won't start playing once resumed.

Actually I now think this is a good suggestion. To address the background music issue, I think we should check if the sound has looping enabled, and if so, we should store a reference to it for later when bailing. And then when the audio context resumes, we can call a method on the web audio backend to tell it to check if those looping sounds would still be playing, and play them now if so.

n0samu:
[...]

PR #9040, opened 2023-01-05
extension: Add autostart to toggle splashScreen/autoplay/unmuteOverlay

danielhjacobs:


n0samu:
I guess this PR was prompted by this tweet: https://twitter.com/rodneylives/status/1610345924462518273
With this PR, wouldn't it still show a bunch of orange Play buttons that you have to click one by one, though?
I guess we should probably add an autoplay option too.

danielhjacobs:
Thoughts of making it just one option: "Disable preloader and autoplay content on mute"?

autostartEnabled: Default false
```javascript
        config: {
            warnOnUnsupportedContent: options.warnOnUnsupportedContent,
            logLevel: options.logLevel,
            showSwfDownload: options.showSwfDownload,
            autoplay: options.autostartEnabled ? "on": "auto",
            unmuteOverlay: options.autostartEnabled ? "hidden" : "visible",
            preloader: !options.autostartEnabled,
        },
```

n0samu:
That would be alright, though you should keep the issues I mentioned in #8361 in mind.

danielhjacobs:
I have looked into that suggestion, but my Rust knowledge is too lacking to be able to implement it properly.

n0samu (approved):
I tried it and it seems to work well, given the limitations already discussed, thanks!

danielhjacobs:
The preloader name/option will be changed after https://github.com/ruffle-rs/ruffle/pull/9802 is merged.

Dinnerbone (approved):
I'm not convinced about hiding the unmute overlay but code wise it looks good to me

danielhjacobs:
Copy-pasting from Discord just to keep PR discussion here for any potential other reviewer:

> Fair enough. Personally I've found this PR's version of Ruffle useful when testing http://new.weedtowonder.org/. Those options and "warnOnUnsupportedContent": false are the configuration options I'd want to use on that site once I enable Ruffle on that site. But because of the current issues I haven't yet enabled Ruffle on that site, so I need to rely on the extension for testing. I can test the site more properly and without as many clicks by using this option (I don't like clicking an orange play button every time I navigate around the site while testing buttons). I know this is a personal use-case, but from my understanding even though Flash allowed audio autoplay, a lot of developers knew it was annoying so required at least one click on the content before the audio started anyway. So for all their websites that are no longer maintained, especially ones found on the web archive, this option is ideal.

nosamu:

> these options will be pretty useful for me as well. I’m already familiar with needing to click to get sound on other sites
> I don’t think that’s too much of a mystery for users, I just wish we could fix Ruffle’s autoplay behavior in general (at some point in the future)
> https://github.com/ruffle-rs/ruffle/issues/8361

web/packages/extension/assets/_locales/en/messages.json
```diff
@@ -11,6 +11,9 @@
     "settings_show_swf_download": {
         "message": "Show SWF download in context menu"
     },
+    "settings_autostart": {
+        "message": "Disable splash screen and autoplay (click to unmute)"
```
torokati44, line 15: It's ambiguous whether this is "disable(splash screen) AND disable(autoplay)" or "disable(splash screen) AND do(autoplay)". At least to me.

torokati44: And I don't know what to do with the "(click to unmute)" part.
I mean, intuitively, I can guess what this is supposed to do, but to a newcomer, it might be difficult to parse.

web/packages/extension/src/content.ts
```diff
@@ -173,6 +173,9 @@ function isXMLDocument(): boolean {
             warnOnUnsupportedContent: options.warnOnUnsupportedContent,
             logLevel: options.logLevel,
             showSwfDownload: options.showSwfDownload,
+            autoplay: options.autostart ? "on" : "auto",
```
torokati44, line 176: What's the difference between "autoplay" and "autostart"? Is this just a typo perhaps?

danielhjacobs: autostart controls unmuteOverlay, autoplay, and splashScreen options simultaneously.

torokati44: Right. Was just Ctrl+F-ing wrong, sorry.

web/packages/extension/assets/_locales/en/messages.json
danielhjacobs: The original phrasing of this option was "Disable splash screen and autoplay content on mute (click to unmute)". I changed it because I found it redundant and wordy (taking up several lines) but I can switch back to that to remove ambiguity.

danielhjacobs: Or maybe "Autoplay content on mute and disable splash screen (click content to unmute)"

torokati44: How about something like "Play automatically on click, without splash screen" or "Start on click, with no splash screen"?

danielhjacobs: But it plays automatically without click. It just needs clicks on the content for it to unmute.

torokati44: Oh, then I misunderstood...
In that case, how about something along the lines of "Play automatically without splash screen (then click to unmute)"?

n0samu: In the future we might want to add tooltips to provide further explanation for this and several other options that are difficult to describe succinctly. For now, let's not get too hung up on the wording.

danielhjacobs: I agree we shouldn't get too caught up with bike-shedding. But I do think torokati's last suggestion is probably worded better, so I went with it. It's a bit long, but still acceptable I suppose.

I can remove this from the popup panel if we feel this is more of a power user feature. Just to show the current length of the text, here are screenshots of the options page and popup panel. If this were removed from the popup panel and I switched from "Play automatically" back to "autoplay", it would fit on one line on the options page:
Screenshot from 2023-04-11 09-23-30
Screenshot from 2023-04-11 09-23-16

n0samu: Looks fine to me, and I think it's good to have on the popup because it's a widely useful feature. Autoplay is one of the features that one ripoff extension was making people pay for, after all 😉

danielhjacobs: Sounds good

torokati44: BTW if the width of the popup was increased a little bit, even this long option weren't broken into 3 lines, so it would look a lot less crowded. But this is a separate design thing...

adrian17 (approved):

PR #9053, opened 2023-01-07
avm2: Fix corner case when goto same frame multiple times (close #9018)

### LINKED ISSUE #9018: Segway of the Dead: Can't get moving
### Describe the bug

https://www.newgrounds.com/portal/view/614276?emulate=flash

Game seems to work fine all the way up until you actually get in game, at which point the main character gets stuck in place talking no matter how many times you press right.

Also a minor issue, but for some reason the main characters eyes stay permanently shut and his mouth doesn't open up at all when he talks in the garage.

Segway

### Expected behavior

You should be able to move forward when pressing D or right and the main character should be moving his mouth when talking and looking with his eyes in the garage.

### Affected platform

Browser's Extension

### Operating system

Windows 11

### Browser

Edge 108

### Additional information

_No response_

MrCheeze:
If you goto a specific frame, then goto the same fame again, frame scripts will run the first time but not the second. This is expected.

However, in the specific corner case where you goto a specific frame that does *not* have a frame script registered, then register a frame script for it, and then goto the same frame again, it looks like Flash Player *does* run the script, but not Ruffle. This PR changes Ruffle's behaviour to match.

[The movie in this issue](https://github.com/ruffle-rs/ruffle/issues/9018) appears to depend on this behaviour. It takes some fairly strange actionscript to hit this issue at all, though.


MrCheeze:
Unsure how to actually author a regression test flash for this one, as the structure of the original movie was fairly complex...

Dinnerbone (approved):

PR #9067, opened 2023-01-08
avm2: Do not remove an EditText's selection on unfocus (fix #9006) + implement selectionBegin/selectionEnd/caretIndex

### LINKED ISSUE #9006: Cannot type in EditText (focus related?)
### Describe the bug

Load the following flash: [html](https://mrcheeze.github.io/detritus/ruffle/00397_no_ctrlKey.html) / [swf](https://mrcheeze.github.io/detritus/ruffle/00397_no_ctrlKey.swf)

Press T to open a secret window. In flash player, you can type in the textbox that appears (and it also auto-focuses on the textbox). In ruffle you can't type at all.

Note that this is an AVM2 flash. Also, this may be related to a script that runs every frame, which removes focus from the input box and then gives it focus again. Here is the script (simplified):

```actionscript
public function enterFrame(param1:Event) : void
{
  stage.focus = stage;
  // ...
  stage.focus = this.secrets.promp.secretInput;
}
```

### Expected behavior

When you press T, a window opens with a textbox in it. You can type in the textbox. The textbox is auto-focused, so you don't have to click it.

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

Mozilla Firefox

### Additional information

_No response_

MrCheeze:
There are significant differences between how selection and caret info work between AVM1 and AVM2.

In AVM1, there is only a single, global selection, which applies to whichever element is currently focused. Therefore changing the focus necessarily erases any information of what was selected before. There can also be no active selection at all.

In AVM2, every text field has its own independent selection info. It is NOT optional, the default selection is just a caret at position 0. If a field loses focus, the selection is not rendered, but it is still present. Movies such as #9006 rely on the same selection still being there once you give back focus to the field.

Ruffle's model of selections (an Option per text field) is different from both the AVM1 model (an optional singleton) and the AVM2 model (mandatory per text field). This fix does not change that, it is only a narrow fix targeted at 9006.


MrCheeze:
Added implementation of the AVM2 selection index properties, for the sake of being able to actually add a regression test for the fixed issue.

adrian17 (approved):
Sorry for late review, I think these are fine.
Only two notes:
- can you please add a comment to the `selection` field with roughly what you wrote in the PR description? Just to keep a TODO in repo.
- tests need updating for the new system without `regression_tests.rs`.

MrCheeze:
@adrian17 Updated the test setup and added a comment to `selection` documenting the AVM1/AVM2/Ruffle differences, but didn't explicitly make it a TODO since I didn't want to take a stance on whether Ruffle should be changed or not.

core/src/focus_tracker.rs
```diff
@@ -59,12 +59,13 @@ impl<'gc> FocusTracker<'gc> {
         // This applies even if the focused element hasn't changed.
         if let Some(text_field) = focused_element.and_then(|e| e.as_edit_text()) {
             if text_field.is_editable() {
-                let length = text_field.text_length();
-                text_field.set_selection(
-                    Some(TextSelection::for_range(0, length)),
-                    context.gc_context,
-                );
-
+                if !context.is_action_script_3() {
```
Lord-McSweeney, line 62: `text_field.movie().is_action_script_3`, since mixed-avm is coming soon.

core/src/display_object/edit_text.rs
```diff
@@ -1861,10 +1883,10 @@ impl<'gc> TDisplayObject<'gc> for EditText<'gc> {
         self.set_avm1_removed(context.gc_context, true);
     }
 
-    fn on_focus_changed(&self, gc_context: MutationContext<'gc, '_>, focused: bool) {
-        let mut text = self.0.write(gc_context);
+    fn on_focus_changed(&self, context: &UpdateContext<'_, 'gc>, focused: bool) {
+        let mut text = self.0.write(context.gc_context);
         text.flags.set(EditTextFlag::HAS_FOCUS, focused);
-        if !focused {
+        if !focused && !context.is_action_script_3() {
```
Lord-McSweeney, line 1889: Use `self.movie().is_action_script_3`- see the above review

Lord-McSweeney (approved):

PR #9077, opened 2023-01-08
render: Make render_offscreen non blocking

Dinnerbone:
This can significantly improve performance of games that use `BitmapData.draw()`, especially if they don't immediately read from the bitmap afterwards.

However, this PR now only includes the possibility to use it async, and actually making AVM1/2 use it async will be in a future PR.


Lord-McSweeney:
With #9086, the game is faster (doesn't lag on drawing blood stains), but about 15 seconds into driving, the game sometimes randomly freezes. I haven't been able to reproduce this consistently, though.

Aaron1011 (approved):

PR #8961, opened 2022-12-29
Web/Desktop: Allow specifying the player version

CUB3D:
This updates both the desktop app and the web player to allow overriding the default player version that is reported to the running movie.

On desktop you can use `--player-version 5` to specify the version. On web, this is done via `playerVersion: 5` in the Ruffle config. 

This fixes #8472

PR #9090, opened 2023-01-09
swf: Accept Scaleform GFx header signatures

### LINKED ISSUE #9070: Autodesk's ScaleForm GFx Support?
Really out of the scope for ruffle, But this thing has similarity as Adobe Flash Player.

**Describe the feature**
Loading GFX and CFX Signature
https://en.wikipedia.org/wiki/Scaleform_GFx

n0samu:
Allows loading and playing Scaleform GFx files using Ruffle's existing functionality. Does not add support for any Scaleform-specific SWF tags, such as DefineSubImage (1008). I'm not sure if that is acceptable but I figured it's worth a shot.

Closes #9070.

### DIFF
--- a/swf/src/read.rs
+++ b/swf/src/read.rs
@@ -2559,8 +2559,8 @@ pub fn read_compression_type<R: Read>(mut input: R) -> Result<Compression> {
     let mut signature = [0u8; 3];
     input.read_exact(&mut signature)?;
     let compression = match &signature {
-        b"FWS" => Compression::None,
-        b"CWS" => Compression::Zlib,
+        b"FWS" | b"GFX" => Compression::None,
+        b"CWS" | b"CFX" => Compression::Zlib,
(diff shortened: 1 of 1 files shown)


kmeisthax (approved):
Seems reasonable for now.

Thinking-out-loud time: For things like Scaleform-specific tags, we probably should have a context variable to gate those off. The reason being that Flash-targeted movies might use the same tag IDs for code obfuscation. We could also use the same mechanism to allow picking a specific Flash Player version (i.e. to fix movies that were broken by Flash Player 10 having two digits in it). So something like:

```
enum PlayerVersion {
    Latest, //default
    AdobeFlash(String), //Provided string is the version string to hand to movies
    ScaleformGFx(String), //Enables Scaleform tags
}
```

n0samu:
Sounds good! And I think #8961 already moves us a bit in that direction.

PR #9096, opened 2023-01-09
wgpu: Implement Screen & Multiply blend modes as a trivial blends

Dinnerbone:
This is a 98% performance savings in rd_flying, and 37% in Heartcore

image
image


render/wgpu/src/blend.rs
```diff
@@ -4,8 +4,6 @@ use swf::BlendMode;
 
 #[derive(Enum, Debug, Copy, Clone)]
 pub enum ComplexBlend {
-    Multiply,   // Can't be trivial, 0 alpha is special case
```
Aaron1011, line 7: Were these comments incorrect?

Dinnerbone: Kiiinda. 0 isn't specialcase, "not written yet" was. By making the default pixel color to white, it handles this case

n0samu:
This is also a dramatic improvement in [Bejeweled 2](https://github.com/ruffle-rs/ruffle/files/10286637/bejeweled2.zip)! Tested in the desktop app on Windows 10 64-bit:
- ~65% GPU usage here vs. ~99% on master
- ~450MB of RAM vs. ~1GB on master
Very exciting! 🎉

Lord-McSweeney:
I see a noticeable performance increase on rd_flying, but it still has noticeable lag.

Lord-McSweeney:
Why are there different images uploaded for the tests? Shouldn't the output be the same?

Dinnerbone:
> Why are there different images uploaded for the tests? Shouldn't the output be the same?

They are visually the same, but an unfortunate issue with image tests is that they rely on pixel by pixel similarities.

The way that hardware blends something vs the way a shader blends something is the same on paper, but some pixels around the edges of shapes where MSAA comes into play are subtly different. Neither one is more correct than the other, it's just how hardware graphics works.

Aaron1011 (approved):

PR #9121, opened 2023-01-11
Dockerfile make dual-wasm optimized build

divinity76:
- compiling from Docker is a slow way to compile anyway (because "COPY . ruffle" and "npm install" is un-cached, and virtualization overhead), so people who want minimal compilation time shouldn't use Docker compile anyway. (switching to :dual-wasm increased build time from 8m20s to 11m10s on my ancient 2011 server, 4x xeon E7-4870)
- seems "docker cp" does not actually overwrite the firefox_unsiged.xpi file if it already exists, for reasons unclear to me, so i added "rm -rf" to the sample invocation
- <strike>git wasn't actually used in docker build process, removed git from apt-get (rustup use the git protocol internally but doesn't seem to use the actual git executable)</strike> nvm added git back


Dinnerbone:
> git wasn't actually used in docker build process, removed git from apt-get (rustup use the git protocol internally but doesn't seem to use the actual git executable)

Git is used by the build script though, to add version info. Does that still work?

divinity76:
@Dinnerbone i don't know how to check, on right-click it identifies as "Nightly {date-of-compilation}" and in Firefox Extensions page it identifies as "version 0.1.0" , here is a firefox_unsigned.xpi built without git being installed:
[optimized_firefox_unsigned.xpi.gz](https://github.com/ruffle-rs/ruffle/files/10429859/optimized_firefox_unsigned.xpi.gz)

However.. the build doesn't actually seem to work, when i try running it on random games on https://archive.org/download/armorgames , it just crashes the same way on all of them, with a 
```
# Error Info
Error name: RangeError
Error message: failed to grow table
Error stack:

966/getImports/imports.wbg.__wbindgen_init_externref_table@moz-extension://ef37fa67-6ac5-42e8-865a-f2c34bf8b8d7/dist/966.js:1796:30
fimport$299@:wasm-function[299]:0x0
finalizeInit@moz-extension://ef37fa67-6ac5-42e8-865a-f2c34bf8b8d7/dist/966.js:1823:10
init@moz-extension://ef37fa67-6ac5-42e8-865a-f2c34bf8b8d7/dist/966.js:1855:12

```
no idea why.. 

But in any case, if git is actually a requirement for building ruffle, it should be added back ofc!

n0samu:
@divinity76 can you check whether this problem is occurring with the current Dockerfile in master?

divinity76:
@Dinnerbone  you're right! lel i should have posted the full crash log, the crash log does -not- know the commit used to build ruffle when ruffle is built without git, the crash log contains `Commit: unknown`  - but adding git back fixes this! (already added git back to this PR)

@n0samu commit 5878bb0c5d6372758eb50b31bb701497205893ad is master right now, and i have confirmed that the Dockerfile builds a fully functional Ruffle firefox extension.
but when i patch it to use :dual-wasm builds, with the only diff being
```
$ git diff
diff --git a/web/docker/Dockerfile b/web/docker/Dockerfile
index f581ae9a9..9f171a90a 100644
--- a/web/docker/Dockerfile
+++ b/web/docker/Dockerfile
@@ -22,4 +22,4 @@ RUN cargo install wasm-bindgen-cli --version 0.2.83
 COPY . ruffle
 WORKDIR ruffle/web
 RUN npm install
-RUN npm run build
+RUN npm run build:dual-wasm
```
then it creates a ruffle that always crashes! full crash log (the ONLY diff from 5878bb0c5d6372758eb50b31bb701497205893ad is printed above):
```
# Error Info
Error name: RangeError
Error message: failed to grow table
Error stack:
```
966/getImports/imports.wbg.__wbindgen_init_externref_table@moz-extension://c996262f-61da-4fd9-8b41-e12dac9a0f05/dist/966.js:2343:30
fimport$464@:wasm-function[464]:0x0
finalizeInit@moz-extension://c996262f-61da-4fd9-8b41-e12dac9a0f05/dist/966.js:2370:10
init@moz-extension://c996262f-61da-4fd9-8b41-e12dac9a0f05/dist/966.js:2402:12

```

# Player Info
Allows script access: false
(Cached) null

# Page Info
Page URL: moz-extension://c996262f-61da-4fd9-8b41-e12dac9a0f05/player.html?url=https%3A%2F%2Fia800104.us.archive.org%2F1%2Fitems%2Farmorgames%2Fcatacombs-234817f.swf

# Browser Info
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:102.0) Gecko/20100101 Firefox/102.0
Platform: Win32
Has touch support: true

# Ruffle Info
Version: 0.1.0
Name: nightly 2023-01-16
Channel: nightly
Built: 2023-01-16T23:58:49.195Z
Commit: 5878bb0c5d6372758eb50b31bb701497205893ad
Is extension: true

# Metadata
```

so.. is dual-wasm builds actually broken right now? or is it a docker build issue? or perhaps an issue at my end?

divinity76:
anyway, until the dual-wasm-crash issue is resolved, this PR should probably not be merged anyway. converting to draft.

btw the builds submitted to addons.mozilla.org , are/were they dual-wasm builds or no?

n0samu:
Yes, our release builds are dual-wasm, so that's what was submitted to Mozilla. I still appreciate the Docker option though because it seemed like it was going to take weeks to walk the reviewer through all of the steps to build Ruffle normally. We'll see how things turn out now.

Dinnerbone:
> so.. is dual-wasm builds actually broken right now? or is it a docker build issue? or perhaps an issue at my end? (it's unlikely to be an issue at my end, that's kind-of the whole point of a Docker build, that is should be fully reproducible and.. build the same thing regardless of host environment )

I think it's using a broken version of `wasm-opt`, older versions had a bug when it came to wasm-with-extensions. It probably needs to build it from source if it can't get a newer version.

divinity76:
okay, updated the Dockerfile to explicitly install binaryen (and thus wasm-opt) version 111 - up from Ubuntu 22.04's bundled version 105,
and re-built with dual-wasm - btw seems v111 use WAY MORE CPU than v105 at building ruffle? like way way more? maxxing all 40 cores for several minutes, using **over 9 hours of CPU time**, pretty sure 105 didn't do that: https://i.imgur.com/t0bv9Fz.png - maybe v111 does much more than v105, or maybe i run a non-optimized/debug-build of binaryen now? (i'm not sure, just using the release file from https://github.com/WebAssembly/binaryen/releases/tag/version_111 ), in any case, build times increased significantly when i switched to that build of v111, over ubuntu's bundled v105, up from 11 minutes to 25 minutes or so (on a 2011-cpu 40-core server)

Anyway, that fixed it! now it builds a functional dual-wasm ruffle!

n0samu:
Hmm, that build time and CPU usage sounds quite worrisome. Can someone try it on a more typical PC?

divinity76:
i don't really have access to a typical-ish pc now, the closest thing i have is a gaming-laptop whose GPU died (it's being RMA-repaired) with an I7-10875H and another laptop rolling i7-8565U, both of these are weird laptop CPUs, and i7-8565U is old-ish (2018) and was weak- even on release (because its a low-power 15watt battery-optimized cpu~)

fwiw the instructions, at least on WSL/Ubuntu should be
```
sudo apt install git docker-compose wget;
git clone https://github.com/divinity76/ruffle;
cd ruffle;
wget 'https://github.com/ruffle-rs/ruffle/pull/9121.diff' -O- | git apply -;
rm -rf web/docker/docker_builds/*;
time docker build --tag ruffle-web-docker -f web/docker/Dockerfile .
docker cp $(docker create ruffle-web-docker:latest):/ruffle/web/packages web/docker/docker_builds/packages;
```
then there should be a `ruffle/web/docker/docker_builds/packages/extension/dist/firefox_unsigned.xpi`

adrian17:
> Can someone try it on a more typical PC?

On my local env (not docker), v111 takes ~20s to run for me. Didn't test on docker, and admittedly my v111 is custom built, so I may not even be on the same commit as the official release.

adrian17:
Actually, yeah, why not get it installed from conda (like our Actions workflow does), instead of from GH release?

adrian17:
Current PR version appears to be failing for me at:
```
Step 9/19 : RUN wget 'https://github.com/WebAssembly/binaryen/releases/download/version_111/binaryen-version_111-x86_64-linux.tar.gz' --quiet -O- | tar xfvz -
 ---> Running in 2e73dcc0bf8e

gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

divinity76:
>why not get it installed from conda (like our Actions workflow does), instead of from GH release?

I'm not familiar with conda, but i gave it a test-spin, seems the conda installer is an interactive installer. i guess we can make an [expect script](https://linux.die.net/man/1/expect) to install conda programmatically though..  or anyone know of an easier way? or anyone good at writing expect scripts? (i'm not good at expect-scripts)


>Current PR version appears to be failing for me here, as the url 404s (...) EDIT: wait no, url worked in browser after 3rd try. Huh. But still not from docker.

.. if that url -sometimes- 404's, maybe it's a github.com issue? 
[It's gotta be the ugliest hack I have ever done](https://memes.getyarn.io/yarn-clip/c96443da-ca51-4805-be27-6a97d202531b),  but.. added a bash-script to attempt to download it up to 10 times if the wget download fails for any reason.. can you try again?

kmeisthax:
So, I played around with this last night. On an i5-1240P Framework laptop, my build time from a completely fresh Docker environment is 14m 18s.

I ran the test *twice*, and compared the output for reproducibility. The WASM that had been optimized (`ruffle_web-wasm_extensions_bg.wasm`) was different by four bytes - specifically two function exports in the module are swapped. I'm not entirely sure *why* it's doing that.

I'd rather not turn off optimizations for the sake of reproducibility. My guess is that we could fix up the WASM module after-the-fact by sorting the exports or something; then it'd compare equal. But I don't have a tool that does that.

adrian17:
> I'm not familiar with conda, but i gave it a test-spin, seems the conda installer is an interactive installer

Don't think so?
https://docs.conda.io/en/latest/miniconda.html

I also checked what our GH Actions (https://github.com/conda-incubator/setup-miniconda) does to get Conda installed, and it appears that it also pulls from there.

divinity76:
@kmeisthax `I ran the test twice, and compared the output for reproducibility. The WASM that had been optimized (ruffle_web-wasm_extensions_bg.wasm) was different by four bytes`  - well, the firefox_unsiged.xpi differs by several megabytes depending whether or not `:dual-wasm` is being built

@adrian17 i didn't try the Miniconda installer, i tried the Anaconda installer, ill give the Miniconda a testrun tonight, anyhow skimming the documentation, it seems both Miniconda and Anaconda have an optional "silent mode", which sounds like exactly what we need in a Dockerfile build: https://conda.io/projects/conda/en/latest/glossary.html#silent-mode-installation
- unfortunately i couldn't find the documentation for how to invoke the silent mode, ill try again tonight (if nobody else wants to do it? :P )

divinity76:
anyone else thinks it's super strange that we're questioning github.com's reliability? anyway 

found (Ana|mini)conda silent installation documentation here https://docs.anaconda.com/anaconda/install/silent-mode/

and it seems this SHOULD have worked, but it doesn't: 
```
RUN wget 'https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh';
RUN chmod +x ./Miniconda3-latest-Linux-x86_64.sh;
RUN bash -c './Miniconda3-latest-Linux-x86_64.sh -b -p /miniconda;'
```
yields
```
Step 10/22 : RUN bash -c './Miniconda3-latest-Linux-x86_64.sh -b -p /miniconda;'
 ---> Running in 4c6d066814c3
PREFIX=/miniconda
Unpacking payload ...
                                                                                                               
Installing base environment...


Downloading and Extracting Packages


Downloading and Extracting Packages

Preparing transaction: ...working... done
Executing transaction: ...working... done

Installing * environment...

./Miniconda3-latest-Linux-x86_64.sh: 438: [[: not found
./Miniconda3-latest-Linux-x86_64.sh: 444: [[: not found

CondaFileIOError: '/miniconda/pkgs/envs/*/env.txt'. [Errno 2] No such file or directory: '/miniconda/pkgs/envs/*/env.txt'

The command '/bin/sh -c bash -c './Miniconda3-latest-Linux-x86_64.sh -b -p /miniconda;'' returned a non-zero code: 1
```
any idea what that's about?

n0samu:
> anyone else thinks it's strange that we're questioning github.com's reliability? anyway

Believe me, it's not strange given the problems we've been having with GH Actions reliability lately...
For example, we were recently affected by https://github.com/actions/runner-images/issues/6913

adrian17:
> anyone else thinks it's strange that we're questioning github.com's reliability?

Not sure what you mean, I don't think anyone mentioned it here. In my eyes, the only reason we need the dockerfile at all is so that Mozilla can build the extension locally (and according to some reports, `act` doesn't seem to work for us); so the main goal is for the docker setup to resemble github's as much as possible.

divinity76:
figured it out! to be perfectly clear, i'm positive there is a bug in the current Miniconda3-latest-Linux-x86_64.sh where it uses `#!/bin/bash`-exclusive syntax with the shebang `#!/bin/sh` , causing normal invocation to crash like above, but when the file is invoked with `bash Miniconda...sh` , the shebang will be ignored, and it will be parsed with bash-rules regardless of shebang, anyway this works:
```sh
RUN wget 'https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh';
RUN bash ./Miniconda3-latest-Linux-x86_64.sh -b -p /miniconda;
ENV PATH="/miniconda/bin:$PATH"
RUN conda install -y -c conda-forge binaryen
```
and it will install binaryen (and wasm-opt) from conda (-:
PR updated.

kmeisthax:
@divinity76 To be perfectly clear, it's not a blocker to this PR that there's 4 bytes different. I just wanted to let you know.

It *should* be possible to sort that particular section of the binary into a "canonical" ordering somehow, but I need to play with other WASM stuff first.

divinity76:
@kmeisthax can you try again with the updated Conda-based binaryen, is it still doing the 4 bytes thing?

divinity76:
perhaps wasm-opt does benchmarks of different function export orderings, and benchmark noise (like other background tasks, or cpu heat changing turbo boost clock speed) makes the export order non-deterministic?

n0samu:
> perhaps wasm-opt does benchmarks of different function export orderings, and benchmark noise (like other background tasks, or cpu heat changing turbo boost clock speed) makes the export order non-deterministic?

I hope not 😨

kmeisthax:
So, I've been trying to reproduce the 'four-byte difference' and haven't been able to. It only ever happened once when having my system clock deliberately set forward by five days. I've tried more builds using the reproducible ZIPs that the build system now generates and they all repro perfectly.

Note that this doesn't completely rule out the bug: I haven't been able to build with Docker plus my system clock set forward five days anymore, due to weird OCSP/TLS cert issues. But I have been able to build with smaller clock differences and they also all agree with the GitHub Actions compiled version of `firefox_unsigned.xpi`. If this turns out to be a "five days old" bug rather than just a "time dependent behavior" bug, I'm just going to tell Mozilla to set their system clock back a few days.

I'm going to merge this. Note that I will need to change `build:dual-wasm` to `build:repro` otherwise Mozilla will get the wrong version in their manifest.

PR #9133, opened 2023-01-12
wgpu: Detect if buffers are unusable on web and fail to create wgpu renderer if so

Dinnerbone:
Until https://github.com/gfx-rs/wgpu/issues/3371 is fixed, this will allow us to use wgpu "most of the time".

### DIFF
--- a/render/wgpu/src/backend.rs
+++ b/render/wgpu/src/backend.rs
@@ -5,6 +5,7 @@ use crate::surface::Surface;
 use crate::target::RenderTargetFrame;
 use crate::target::TextureTarget;
 use crate::uniform_buffer::BufferStorage;
+use crate::utils::detect_buffer_bug;
 use crate::{
     as_texture, format_list, get_backend_names, ColorAdjustments, Descriptors, Error,
     QueueSyncHandle, RenderTarget, SwapChainTarget, Texture, Transforms,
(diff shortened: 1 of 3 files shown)


render/wgpu/src/backend.rs
```diff
@@ -51,17 +51,25 @@ impl WgpuRenderBackend<SwapChainTarget> {
...
-            wgpu::PowerPreference::HighPerformance,
-            None,
+        let adapter = instance
+            .request_adapter(&wgpu::RequestAdapterOptions {
+                power_preference: wgpu::PowerPreference::HighPerformance,
+                compatible_surface: Some(&surface),
+                force_fallback_adapter: false,
+            })
+            .await
+            .ok_or_else(|| "Unsupported device".to_string())?;
```
Aaron1011, line 61: This makes us lose the nicer error message from `build_descriptors`. I think it would be better to better to call `detect_buffer_bug` from `build_descriptors` (possibly behind a flag).

Dinnerbone: I'll split up build_descriptors into the adapter finding stuff only, we need that anyway for the upcoming libretro stuff

Dinnerbone: Actually no nvm that means multiple `block_ons` at the caller, that can be a later change. I'll add a param!

Dinnerbone: Ended up with cfg'ing it since it's only for web

Aaron1011 (approved):

PR #9134, opened 2023-01-12
web: Enable wgpu-webgl by default

Dinnerbone:
Let's do it? Any objections?

Newgrounds has been running an old build pre-optimisations and as far as we've seen, it's been great and only had the buffer bug, which #9133 deals with (they'll fall back to regular webgl).

### DIFF
--- a/web/Cargo.toml
+++ b/web/Cargo.toml
@@ -13,7 +13,7 @@ publish = false # This crate is useless alone, people should use the npm package
 crate-type = ["cdylib", "rlib"]
 
 [features]
-default = ["canvas", "console_error_panic_hook", "webgl"]
+default = ["canvas", "console_error_panic_hook", "webgl", "wgpu-webgl"]
 
 # core features
 avm_debug = ["ruffle_core/avm_debug"]


torokati44:
Oh, I'd love the get those rendering goodies you put together in the past months to everyone... ☺️

One consequence, losing WebGPU support, doesn't matter the slightest, at this point in time, in my opinion.
Another is: Will this raise the minimum browser/HW requirements at all? I'm thinking of WebGL1, or any optional extensions that will be needed for this to work... If there's a runtime fallback to the classic WebGL backend, this is a moot point.
Also, has that seconds-long hang at the beginning experienced by Adrian on Windows (IIRC) been resolved?

PS.: I'm sorry if any or all of these were already addressed on Discord, I'm _really_ behind on my reading there...

Dinnerbone:
@torokati44 Right now there's a fallback to the "normal" WebGL backend if they can't support the wgpu one. These people will be:
- People with WebGL1.0
- People with the buffer bug (a range of intel igpus) - note that these people will have issues on webgl backend too, it's just less noticeable
- People with a broken or insufficient WebGL2.0 setup for our needs

So for those people, nothing changes, they'll use the old webgl-or-canvas fallback. That's something we should discuss in the future if we want to keep that backend or drop them to canvas entirely, but not something for today I think.

For the seconds-long hang, as far as I know it should be fast to startup now, it's significantly less pipelines. @adrian17 will have to confirm, but I haven't seen him mention it in his recent tests.

torokati44:
Alright, awesome! \o/

Aaron1011 (approved):
I'm ready to merge this, once we have confirmation from @adrian17 and other affected users that this no longer hangs on startup.

n0samu:
I still think we should do one more round of testing with NG before merging this. For the record, here's what Tom said about the previous build:
> Heads up, I've switched Newgrounds to the latest nightly because I've seen a decent amount of people at this point where the experimental build wasn't working. I'm down to test another build any time though!

But given the many improvements since the last build I sent him, I think it's pretty likely NG won't see problems this time. Just want to wait for confirmation of that, though.

n0samu:
Oh and along the lines of what Adrian17 was pointing out earlier, I'm seeing a significant new startup delay. Tested on Windows 10 in Firefox and Chromium with the demo and extension. I imagine this could get quite annoying for impatient users (which is probably many of them).


https://user-images.githubusercontent.com/71368227/212299822-d10a6c27-0b6b-483e-b0a3-0c2aa0544908.mp4


https://user-images.githubusercontent.com/71368227/212299838-bd60ba4f-dd8d-4f9d-9adc-5f4b48d9fb97.mp4

n0samu:
I can confirm that with https://github.com/Dinnerbone/wgpu/commit/f1fdd1ceb1b4518b627e70b682e6695448d625d7, the startup delay is drastically reduced! Still not as fast as WebGL, but I think this is good enough.


https://user-images.githubusercontent.com/71368227/212529574-12cfe211-2009-4de3-864e-e29a86350326.mp4

DustinBrett:
The WASM bundles increased from 6 MB to 10 MB after this change. Not sure if this was known or expected, but I went to upgrade Ruffle for my self hosted use and noticed since Jan 15th, 2023 it was around 6 MB and now it's at 11 MB. The biggest jump I saw was from this commit/release it seemed.

n0samu:
Yes that is expected, unfortunately

Issue #9137, opened 2023-01-13
Cascade: Huge 50MB flash just gives a white screen on web (but was working before)


MrCheeze:
### Describe the bug

[cascade.swf](https://github.com/MrCheeze/homestuck-ruffle-tester/blob/gh-pages/flash/cascade.swf) is an enormous 55MB movie that packs several sub-movies inside itself. Currently, it runs perfectly on desktop, but running it on web seems to overload the player, showing just a white box instead of the ruffle player, and spewing strange errors in the logs.

A few months ago it was working, for example you can still see it working in nightly 2022-11-19 [on archive.org](https://archive.org/details/cascade-homestuck).

Using git bisect, I was able to locate the commit where it stopped working, it was https://github.com/ruffle-rs/ruffle/commit/3bef8c4fd62bb52e99f3db4827ede41be7bc1899 .

### Expected behavior

The movie can be loaded in the web player, and actually plays instead of just showing a white screen.

### Affected platform

Online demo

### Operating system

Windows 11

### Browser

Firefox 108.0.2 (64-bit)

### Additional information

On the commit before 3bef8c4fd62bb52e99f3db4827ede41be7bc1899, the flash works.

On 3bef8c4fd62bb52e99f3db4827ede41be7bc1899, I get a white screen and this error in the logs:

```
New Ruffle instance created (WebAssembly extensions: OFF) ruffle-player.js:244:22
Loading SWF data ruffle-player.js:385:23
Uncaught RuntimeError: unreachable executed
    F ruffle_web-wasm_extensions.js:236
    c ruffle_web-wasm_extensions.js:213
    __wbg_requestAnimationFrame_4181656476a7d86c ruffle_web-wasm_extensions.js:820
    L ruffle_web-wasm_extensions.js:258
    __wbg_requestAnimationFrame_4181656476a7d86c ruffle_web-wasm_extensions.js:819
    F ruffle_web-wasm_extensions.js:236
    c ruffle_web-wasm_extensions.js:213
    __wbg_requestAnimationFrame_4181656476a7d86c ruffle_web-wasm_extensions.js:820
    L ruffle_web-wasm_extensions.js:258
    __wbg_requestAnimationFrame_4181656476a7d86c ruffle_web-wasm_extensions.js:819
    F ruffle_web-wasm_extensions.js:236
    c ruffle_web-wasm_extensions.js:213
[...]

PR #9143, opened 2023-01-13
core: Use BitmapData instead of separate `initial_data` field

Aaron1011:
Now that a `Bitmap` always stores a `BitmapData`, we can read the pixels directly from the `BitmapData`, instead of duplicating them in an `initial_data` field

### DIFF
--- a/core/src/avm1/globals/bitmap_data.rs
+++ b/core/src/avm1/globals/bitmap_data.rs
@@ -1309,7 +1309,6 @@ pub fn load_bitmap<'gc>(
 
     if let Some(Character::Bitmap {
         bitmap: bitmap_object,
-        initial_data,
     }) = character
     {
         let new_bitmap_data = BitmapDataObject::empty_object(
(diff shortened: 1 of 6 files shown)


MrCheeze:
This change does reduce memory use from 5 to 2-3 GB in ruffle_desktop, but unfortunately it still goes above 4GB in any web browser, meaning https://github.com/ruffle-rs/ruffle/issues/9137 does not get fixed.

MrCheeze:
> This change does reduce memory use from 5 to 2-3 GB in ruffle_desktop, but unfortunately it still goes above 4GB in any web browser, meaning #9137 does not get fixed.

Actually, although this PR does not fix Cascade on its own, it DOES fix Cascade when merged with https://github.com/ruffle-rs/ruffle/pull/9134 (memory usage reduced down to 3GB on web with both PRs merged).

So I guess... either the optimizations of this PR don't apply to the webgl renderer for some reason; or else the wgpu branch happens to have a lower memory footprint for unrelated reasons, and both optimizations combined are enough to bring cascade under the 4GB limit.

Dinnerbone (approved):

PR #9146, opened 2023-01-13
avm2: Correct name of rollover event

n0samu:
Due to a typo, the rollOver event was being fired as a mouseOut event instead. Also, the variable for the target of a rollOut event was misnamed as "rollover_target".

Fixes this little issue in Pursuit of Hat:

https://user-images.githubusercontent.com/71368227/212291896-5e8a0312-88e0-4074-aed7-1674c75ec390.mp4


adrian17 (approved):
nice catch :)

Dinnerbone (approved):

PR #9173, opened 2023-01-15
wgpu: Fix buffer bug detection

Dinnerbone:
This makes 8 passes of the test just to be safe, but from testing it appears that it always fails on pass 0. We can reduce the passes down to 4 or 2 or 1 later if we find it impacts performance, but let's be safe for now.


torokati44 (approved):
I haven't looked at the code at all (there would be no point to that, really), but I have tested it, and does what it's supposed to do!
That is, makes Ruffle work on my old laptop with an Intel HD4000, while it still works well on my newer desktop.

Aaron1011 (approved):

PR #9178, opened 2023-01-16
Delay reading image back from render backend using `SyncHandle`

Aaron1011:
This allows us to avoid blocking immediately after a `BitmapData.draw` call. Instead, we only attempt to use the `SyncHandle` when performing an operation that requires the CPU-side pixels (e.g. BitmapData.getPixel or BitmapData.setPixel).

In the best case, the SWF will never explicitly access the pixels of the target BitmapData, removing the need to ever copy back the render backend image to our BitmapData. If the SWF doesn't require access to the pixels immediately, we can delay copying the pixels until they're actually needed, hopefully allowing the render backend to finish processing the BitmapData.draw operation in the backenground before we need the result.

Now that the CPU and GPU pixels can be intentionally out of sync with each other, we need to ensure that we don't accidentally expose 'stale' CPU-side pixels to ActionScript (which needs to remain unaware of our internal laziness). To help enforce this, all access to the CPU-side pixels now goes through the new `PixelsWrapper` type. This type has a `Deref` impl which panics if a `SyncHandle` is present (e.g. `ensure_updated` has now been called).

This will prevent execution from continuing if we forget to call `ensure_updated` in one of the many places that require it, rather than reeturning incorrect data to ActionScript.

### DIFF
--- a/core/src/avm1/globals/bitmap_data.rs
+++ b/core/src/avm1/globals/bitmap_data.rs
@@ -159,6 +159,10 @@ pub fn get_pixel<'gc>(
 ) -> Result<Value<'gc>, Error<'gc>> {
     if let Some(bitmap_data) = this.as_bitmap_data_object() {
         if !bitmap_data.disposed() {
+            bitmap_data
+                .bitmap_data()
+                .write(activation.context.gc_context)
+                .ensure_updated(activation.context.renderer);
             if let (Some(x_val), Some(y_val)) = (args.get(0), args.get(1)) {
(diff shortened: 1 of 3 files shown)


Dinnerbone (changes requested):
Initial review - I think the design needs some work, sorry. I expected something like the PixelsWrapper to hold "cpu side vs gpu side" as an internal secret and to update when accessed, not expose it everywhere. That doesn't feel much better than the earlier design, but now has the capability to panic a few places :(

Dinnerbone:
Closing in favor of #9184

PR #9181, opened 2023-01-16
web: Cache the output of render debug info in case of panics

Dinnerbone:


render/wgpu/src/backend.rs
```diff
@@ -345,12 +345,10 @@ impl<T: RenderTarget + 'static> RenderBackend for WgpuRenderBackend<T> {
         let enabled_features = self.descriptors.device.features();
         let available_features = self.descriptors.adapter.features() - enabled_features;
         let current_limits = &self.descriptors.limits;
-        let available_limits = &self.descriptors.limits;
```
Aaron1011, line 348: Why did you remove this?

web/packages/core/src/ruffle-player.ts
```diff
@@ -1534,8 +1536,16 @@ export class RufflePlayer extends HTMLElement {
 
     protected debugPlayerInfo(): string {
         let result = `Allows script access: ${this.loadedConfig.allowScriptAccess}\n`;
+        let renderInfo = this._cachedDebugInfo;
         if (this.instance) {
-            result += `${this.instance.renderer_debug_info()}\n`;
+            try {
+                renderInfo = this.instance.renderer_debug_info();
+            } catch {
+                // ignored
```
Aaron1011, line 1544: Could you make the output include something like "(Cached)", and maybe even the error message, when this happens?

render/wgpu/src/backend.rs
Dinnerbone: Discord discussion - nosamu was worried it was too verbose.

(I also realized it's the same info twice, when removing it. It should have been `let available_limits = self.descriptors.adapter.limits();` oops)

web/packages/core/src/ruffle-player.ts
Dinnerbone: The error will be the whole "we're panicking and can't access stuff" that we currently spam the console with, I don't think it'll be valuable. I'll add Cached though!

PR #9184, opened 2023-01-16
Take two: Delay reading image back from render backend using `SyncHandle`

Aaron1011:
This allows us to avoid blocking immediately after a `BitmapData.draw` call. Instead, we only attempt to use the `SyncHandle` when performing an operation that requires the CPU-side pixels (e.g. BitmapData.getPixel or BitmapData.setPixel).

In the best case, the SWF will never explicitly access the pixels of the target BitmapData, removing the need to ever copy back the render backend image to our BitmapData. If the SWF doesn't require access to the pixels immediately, we can delay copying the pixels until they're actually needed, hopefully allowing the render backend to finish processing the BitmapData.draw operation in the backenground before we need the result.

Now that the CPU and GPU pixels can be intentionally out of sync with each other, we need to ensure that we don't accidentally expose 'stale' CPU-side pixels to ActionScript (which needs to remain unaware of our internal laziness). We now use a wrapper type `BitmapDataWrapper` to enforce that the `SyncHandle` is consumed before accessing the underlying `BitmapData.

### DIFF
--- a/core/src/avm1/globals/bitmap_data.rs
+++ b/core/src/avm1/globals/bitmap_data.rs
@@ -536,7 +536,7 @@ pub fn draw<'gc>(
             let source = if let Some(source_object) = source.as_display_object() {
                 IBitmapDrawable::DisplayObject(source_object)
             } else if let Some(source_bitmap) = source.as_bitmap_data_object() {
-                IBitmapDrawable::BitmapData(source_bitmap.bitmap_data())
+                IBitmapDrawable::BitmapData(source_bitmap.bitmap_data_wrapper())
             } else {
                 avm_error!(
                     activation,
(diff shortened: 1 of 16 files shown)


Dinnerbone:
I haven't tested it yet to confirm, but it seems like `BitmapData.draw()` would cause the bitmapdata to sync before drawing. If that's so, it's going to negate the performance gains when it comes to repeat drawing - we'll need to find a way to hint that we don't want to sync.

Aaron1011:
@Dinnerbone: Good catch. I've adjusted the AVM1 and AVM2 Bitmapdata.draw impls to ensure that we don't end up calling `sync`.

Dinnerbone (approved):

PR #9193, opened 2023-01-18
ci: Change activity check to something less error prone possibly

Dinnerbone:
Does it work? That's a great question and I'm glad you asked.


Aaron1011:
Let's give this a try

PR #9197, opened 2023-01-18
 Desktop + Demo: Add .gfx to file dialog filter

p0008874:
We already have Scaleform GFx Signature, How about file dialog filter make people shock.

Using For Scaleform GFx, Make 3A games modding People notice we're exist.

PS: The Original Flash Player can't open this up.


n0samu:
Ruffle doesn't support any Scaleform-specific tags yet, so I don't think it's reasonable to imply real support for these files by adding them to the dialog filters. We already have the "All Files" option for cases like this.

Dinnerbone:
> Ruffle doesn't support any Scaleform-specific tags yet, so I don't think it's reasonable to imply real support for these files by adding them to the dialog filters. We already have the "All Files" option for cases like this.

I agree with this. You can open them but that doesn't mean it's **supported**, whereas promoting opening them would imply that it is. I think we'll close this for now. Maybe down the road if our scope changes then we can reevaluate this.

Issue #3696, opened 2021-03-17
Enter Key from Numpad not working


q2apro:
**Describe the bug**
It is not a bug but a usability thing. Instead of the enter key I am using the enter key on the numpad pretty often.

Problem: Ruffle does not recognize this key.

**Is the problem with the Ruffle desktop app, extension, or self-hosted version?**

Self-hosted (web).

**What platform are you using?**
Windows 10, Chrome latest

Lord-McSweeney:
Fixed by #9204.

PR #9204, opened 2023-01-18
desktop+web: Add missing key codes

### LINKED ISSUE #8492: Caravaneer: Enter and Backspace keys don't work in the Operator Console
### Describe the bug

In [Caravaneer](https://www.kongregate.com/games/SugarFreeGames/caravaneer), there is a console in the military base where you can type commands. In Ruffle, typing the commands works, but you cannot send the command by pressing the Enter key. And deleting letters by pressing Backspace doesn't work either.

To make this easier to reproduce, the user who alerted me to the problem has provided a save file that takes you straight to the relevant area. And to demonstrate the source of the problem, I've modified the SWF to trace the keycodes that the user types. Here is some of the relevant ActionScript:
```
   _parent.KeyListener.onKeyDown = function()
   {
      if(_currentframe == 12 and LConMode == 1 or LConMode == 2)
      {
         KeyCode = Key.getAscii();
         trace(KeyCode); // Added by me
         if(KeyCode != 8 and KeyCode != 13)
         {
            Command += chr(KeyCode);
         }
         if(KeyCode == 8)
         {
            Command = Command.substring(0,Command.length - 1);
         }
```
Follow these steps to reproduce the issue:
1. [Download and extract this zip containing the SWF file and SOL save file.](https://github.com/ruffle-rs/ruffle/files/9947010/Caravaneer.zip)
2. Place the save file in the correct location depending on the SWF file's location. For example, mine is at:
`%LOCALAPPDATA%\ruffle\SharedObjects\localhost\Tech\RuffleIssues\CaravaneerKongregate_keycodeTrace.swf\Caravaneer.sol`
3. Run the SWF in Ruffle via the command line. (Make sure the log level is set to show trace statements.)
4. Click Load Game on the main menu and load the save from slot 1. 
5. Click Fort Goks to navigate to it. Click the military base near the center of the map. Click "Here it is" to show your pass. Finally click the Operator's Console in the top-left part of the military base map.
6. Type stuff into the console and notice Ruffle tracing the corresponding ASCII codes. Then press keys like Enter and Backspace, and notice that they trace `0` even though they shouldn't. (This is also true of other keys like Shift and Tab, but the game doesn't use these keys.)

### Expected behavior

Pressing Backspace should delete text and pressing Enter should send the command.
And in Flash Player, pressing Enter traces `13` and Backspace traces `8`. And other keys like Tab also have the correct key codes, not `0`. 

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

The user who originally reported this problem said that the Backspace key works for him, but the Enter key does not. He is on Linux - maybe there's something platform-dependent happening here as well?

Also, this problem seems similar to #1512 and related to #3696.

CUB3D:
Add key codes for Tab, Enter, Backspace and Numpad Enter
This fixes #8492 and #3696

### DIFF
--- a/desktop/src/main.rs
+++ b/desktop/src/main.rs
@@ -777,6 +777,7 @@ fn winit_key_to_char(key_code: VirtualKeyCode, is_shift_down: bool) -> Option<ch
         (VirtualKeyCode::NumpadSubtract, _) => '-',
         (VirtualKeyCode::NumpadDecimal, _) => '.',
         (VirtualKeyCode::NumpadDivide, _) => '/',
+
         (VirtualKeyCode::Numpad0, false) => '0',
         (VirtualKeyCode::Numpad1, false) => '1',
         (VirtualKeyCode::Numpad2, false) => '2',
@@ -787,6 +788,8 @@ fn winit_key_to_char(key_code: VirtualKeyCode, is_shift_down: bool) -> Option<ch
         (VirtualKeyCode::Numpad7, false) => '7',
         (VirtualKeyCode::Numpad8, false) => '8',
         (VirtualKeyCode::Numpad9, false) => '9',
+        (VirtualKeyCode::NumpadEnter, _) => '\r',
+
         (VirtualKeyCode::Tab, _) => '\t',
         (VirtualKeyCode::Return, _) => '\r',
         (VirtualKeyCode::Back, _) => '\u{0008}',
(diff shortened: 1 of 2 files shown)


Lord-McSweeney:
Also fixes #3696.

Dinnerbone (approved):

PR #9207, opened 2023-01-18
avm2: Implement TextField.maxChars

adrian17:
Handled with a simple check in text input handler. Will need to be careful around stuff like ctrl+v.
Only tested on the simplest possible case of
```as
         tf = new TextField();
         tf.maxChars = 5;
         tf.type = "input";
```


Dinnerbone (approved):
Looks good to me and works as I'd expect. We need to find a way to test this stuff though, adding key presses to the automation seems smart.

PR #9210, opened 2023-01-19
Move back to upstream wgpu repository

Aaron1011:
We're relying on unreleased changes, so we depend on the latest commit from their repository.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2621,14 +2621,16 @@ dependencies = [
 [[package]]
 name = "naga_oil"
 version = "0.4.2"
-source = "git+https://github.com/Dinnerbone/naga_oil?rev=c073f6fdf0560a74480140abc50295f9e0d9aa57#c073f6fdf0560a74480140abc50295f9e0d9aa57"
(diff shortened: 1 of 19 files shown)


Dinnerbone:
Upstream doesn't have the GL startup time fix yet, should do tomorrow when I get a chance to clean up the PR.

Does this run? Last time I tried to update us, the naga update broke our shaders, I suspect something with naga oil.

danielhjacobs:
https://github.com/gfx-rs/wgpu/pull/3380 is in master now. Probably best to switch to the latest commit to pull that in too.

danielhjacobs:
That commit was apparently after another naga bump, which is probably what's causing the new CI failures: https://github.com/gfx-rs/wgpu/pull/3405.

Aaron1011:
naga previously allowed duplicate struct definitions in wgsl input, but it now prevents that. `naga_oil` has been creating duplicate struct definitions due to the way it handles nested imports, but naga was accepting it until now.

I'm trying to come up with a simple way to fix this.

Dinnerbone (approved):

PR #9243, opened 2023-01-21
Implement AVM2 BitmapData.applyFilter for ColorMatrixFilter and BlurFilter

Dinnerbone:
Lays the groundwork for all filters too.


Dinnerbone:
Leaving as draft until after #9184 is in

core/src/bitmap/bitmap_data.rs
```diff
@@ -1277,6 +1277,40 @@ impl<'gc> BitmapData<'gc> {
...
+        context: &mut UpdateContext<'_, 'gc>,
+        source: BitmapHandle,
+        source_point: (u32, u32),
+        source_size: (u32, u32),
+        dest_point: (u32, u32),
+        filter: Filter,
+    ) {
+        let dest = self.bitmap_handle(context.renderer).unwrap();
+
+        self.update_dirty_texture(context.renderer);
```
Aaron1011, line 1291: This is almost identical to the BitmapData.draw implementation. I think it could be factored out into a method that takes a `FnOnce() -> Option<SyncHandle>` closure, which calls `update_dirty_texture`, calls the closure, and then processes the `SyncHandle`.

Dinnerbone: Good idea

Dinnerbone: @Aaron1011 Tried this and it's not that easy, we need to mutably borrow the renderer but we can't do that if the closure needs it too (which it will). Could probably make a function of the `match self.dirty_state` but that's about it

core/src/avm2/globals/flash/display/bitmapdata.rs
```diff
@@ -852,10 +853,249 @@ pub fn rect<'gc>(
...
+        [
+            args.get(0).expect("Infallible"),
+            args.get(1).expect("Infallible"),
+            args.get(2).expect("Infallible"),
+            args.get(3).expect("Infallible"),
+        ]
+    } else {
+        return Err("ArgumentError: Error #1063: Argument count mismatch on flash.display::BitmapData/applyFilter(). Expected 4, got 0.".into());
+    };
+    if let Some(dest_bitmap) = this.and_then(|this| this.as_bitmap_data()) {
```
Aaron1011, line 869: This should use `as_bitmap_data_wrapper` and `overwrite_cpu_pixels`, like the `draw` method in this file does. This will avoid blocking on any in-progress GPU -> CPU sync.

Aaron1011 (approved):

Aaron1011:
I noticed a pre-existing issue with the gpu->cpu sync code, so I think this is fine to merge in its current state. I'll submit a follow-up PR fixing the pre-existing and new issues that I've found.

core/src/avm2/globals/flash/display/bitmapdata.rs
Dinnerbone: Fixed - but I realized that we can't do this for `source_bitmap`. We don't need the copy to happen here, but we don't want to invalidate it.

Issue #8799, opened 2022-12-15
Firefox extension not being updated after 2023-02-22?


ousia:
I have the extension from https://addons.mozilla.org/en-US/firefox/addon/ruffle_rs/.

But the extension displays version as nightly 2022-12-07.

There have been other releases (https://github.com/ruffle-rs/ruffle/releases).

I have removed and installed the extension again and version remains the same.

I’m experiencing this both on Windows and Linux.

Could anyone confirm the issue?

Many thanks for your help.

torokati44:
Thanks for the report!
Yes, we are aware of this. The extension is stuck in a "manual review queue" on addons.mozilla.org ... again.
We're working on fixing it.

ousia:
I’m afraid that my browser cannot find https://addons.mozilla.org/en-US/firefox/addon/ruffle_rs/.

Is it just me?

colin-i:
> I’m afraid that my browser cannot find https://addons.mozilla.org/en-US/firefox/addon/ruffle_rs/.
> 
> Is it just me?

It is not just you. "Mozilla removed it." or "The developer removed it.".

divinity76:
any update? did the Ruffle devs intentionally remove it, or did Mozilla remove it? what was the rationale?

n0samu:
Hi, sorry for the late response. Unfortunately Mozilla removed the add-on because they asked us to prove that our build matches our source code, and we provided them with our source code and build instructions, but they were unable to follow our build instructions, even after multiple clarifications. The grace period expired and the add-on was delisted. We don't know when it will be reinstated, but we have sent followup messages to Mozilla to try to resolve the situation.

divinity76:
Can we make a dockerfile to build it? Then the instructions could just be
```git clone + cd + docker-compose -f docker-compose.yml build```

divinity76:
made a PR for a Dockerfile build: https://github.com/ruffle-rs/ruffle/pull/9066

n0samu:
Thanks so much! I think this will help a lot in the future. 👍

n0samu:
An update for anyone wondering: Thanks to the Dockerfile and clarified build instructions, Mozilla has succeeded in building Ruffle several times. But so far, each time they have claimed that our build does not match theirs. They have also complained about the minified plugin polyfill code within our source code, so we addressed the complaint in #9588. (This was an improvement to our build system that we had been meaning to make for years, anyway!)

[...]

PR #9244, opened 2023-01-21
core: Allow sealing build information into the source ZIP for reproducibility.

kmeisthax:
As per #8799 we need reproducible builds before Mozilla will let us back onto addons.mozilla.org. One of the largest differences between Mozilla's build and ours (save for the lack of dual-wasm) was version information, since Mozilla's reviewers are building from a source ZIP instead of a Git clone and they are building at a different time from us.

The solution is to... calculate all that information once and store it for later use. The resulting file is called `version_seal.json`. The plan is that source code submissions to Mozilla will be special "reproducible builds" archives with the version seal in them.

The version seal is only used when you specify reproducible builds with `npm run build:repro`, which also builds dual-WASM. The standard and dual-WASM builds will not use the version seal. Any future reproducibility fixes will be gated behind `build:repro` for now.

Note that this alone won't make builds reproducible - there's another major reproducibility problem I need to look into.

Intersects with #9121 - the `Dockerfile` will need to be changed to use `build:repro` once this PR lands, since that's the only build system Mozilla is able to run right now.

### DIFF
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -318,7 +318,7 @@ jobs:
         shell: bash -l {0}
         run: |
           npm ci
-          npm run build:dual-wasm
+          npm run build:repro
           npm run docs
 
       - name: Publish Firefox extension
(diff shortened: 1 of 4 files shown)


Aaron1011 (approved):

Dinnerbone (approved):

torokati44:
Were you able to sort out the unstable order of exports in the .wasm files?

PR #9261, opened 2023-01-22
web: Construct SwfMovie with spoofed URL when loaded from data

### LINKED ISSUE #8130: Got RustError when constructing AVM2 side of movie clip of type Preloader
### Describe the bug

I'm getting the following error on loading a game
`Got RustError("Cannot access property http://adobe.com/AS3/2006/builtin::split of null or undefined") when constructing AVM2 side of movie clip of type Preloader`
It also gives the ActionScript3 unsupported content [warning](https://github.com/ruffle-rs/ruffle/wiki/Frequently-Asked-Questions-For-Users#why-did-i-get-a-message-about-unsupported-content), but not sure if the above is also related to the unsupported content warning.

### Expected behavior

Do I wait for the ActionScript 3.0 completion of [tasks](https://github.com/ruffle-rs/ruffle/issues/1368)? Or am I missing something that can be fixed on my end?

### Affected platform

Self-hosted version

### Operating system

macOS

### Browser

Chrome

### Additional information

_No response_

### LINKED ISSUE #8448: web: Specify SWF filename via DataLoadOptions
**Describe the feature**
As discussed in #8130, loading SWFs via ArrayBuffer (DataLoadOptions) causes problems for many AVM2 games because they try to perform operations on the SWF URL, but Ruffle sets the SWF URL to null. Flash Player only supports loading root movies via URL or file path, so this is a situation that many games just are not capable of handling. And it's an especially significant problem because the Ruffle Web Demo and Archive.org's Emularity both use ArrayBuffer to load SWFs. [Here is an example game that is broken on Archive.org due to this issue](https://archive.org/details/flash_gravityduck).

I suggest adding an optional SWF filename field to DataLoadOptions that would be concatenated with the document base URI to form the SWF URL seen by ActionScript. For example, loading a movie via ArrayBuffer on the webpage `https://example.com/game.html`, with specified filename "game.swf", would produce the SWF URL `https://example.com/game.swf`.

If this optional field is not used, we could either keep the current behavior or use a default filename (maybe "movie.swf").

n0samu:
Implements the feature described in and closes #8448.
Fixes #8130.

To prevent abuse of this feature, the spoofed URL is constructed relative to the webpage's current URL, and the user can only specify the last path segment of the URL.

### DIFF
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -640,7 +640,8 @@ export class RufflePlayer extends HTMLElement {
                 console.log("Loading SWF data");
                 this.instance!.load_data(
                     new Uint8Array(options.data),
-                    sanitizeParameters(options.parameters)
+                    sanitizeParameters(options.parameters),
+                    options.swfFileName || "movie.swf"
                 );
             }
         } catch (err) {
--- a/web/packages/core/src/load-options.ts
+++ b/web/packages/core/src/load-options.ts
@@ -316,4 +316,11 @@ export interface DataLoadOptions extends BaseLoadOptions {
      * The data to load a movie from.
      */
     data: Iterable<number>;
+
+    /**
+     * The filename of the SWF movie to provide to ActionScript.
+     *
+     * @default "movie.swf"
+     */
+    swfFileName?: string;
 }
--- a/web/packages/demo/www/index.js
+++ b/web/packages/demo/www/index.js
@@ -144,7 +144,7 @@ async function loadFile(file) {
     }
     hideSample();
     const data = await new Response(file).arrayBuffer();
-    load({ data, ...defaultConfig });
+    load({ data: data, swfFileName: file.name, ...defaultConfig });
 }
 
 function loadSample() {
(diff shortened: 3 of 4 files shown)


web/packages/core/src/ruffle-player.ts
```diff
@@ -638,9 +638,18 @@ export class RufflePlayer extends HTMLElement {
...
+                        options.swfFileName || "movie.swf",
+                        document.baseURI
+                    )
+                );
+
                 this.instance!.load_data(
                     new Uint8Array(options.data),
-                    sanitizeParameters(options.parameters)
+                    sanitizeParameters(options.parameters),
+                    new URL(movieName, document.baseURI).href
```
Dinnerbone, line 652: We should do this in WASM and not JS, just to be safe. If the WASM side asks for the URL from JS, it's a simple JS script to call it with anything else (or even just search-and-replace for the lazy malicious host)

n0samu: I don't really know how to do that from WASM, but feel free to push that change to my PR

Dinnerbone (approved):

PR #9265, opened 2023-01-22
avm2: added Stage.fullScreenSourceRect stub

MartySVK:
Fracuum is playable thanks to Stage.fullScreenSourceRect Stub


n0samu:
Can confirm that this fixes Fracuum, thank you! 😄 
Looks like the linter is complaining though, you can run `cargo fmt` to fix it!

MartySVK:
done...thanks

Aaron1011 (approved):

Issue #8006, opened 2022-09-17
Ruffle can't load Adobe Flex SWZ files


danielhjacobs:
### Describe the bug

Since this is AS3, I had avoided reporting errors with it while they related to known unimplemented APIs. However, it currently has no error messages about unimplemented APIs in the console.

Mentioned first on Discord, this is most likely the current bug with http://new.weedtowonder.org/, SWF file http://new.weedtowonder.org/w2w-home.swf. It tries to load http://fpdownload.adobe.com/pub/swz/tlf/2.0.0.232/textLayout_2.0.0.232.swz, which does cause a console error due to CORS, but then it successfully loads the SWZ file from http://new.weedtowonder.org/textLayout_2.0.0.232.swz (according to the 200 status code in the Network tab). nosamu confirmed that this would be the fallback URL for the SWZ file: "usually Flex apps request from fpdownload.adobe.com first, and if that fails request from the same folder on the web as the rest of the assets".

So the most likely current issue is that Ruffle is not set up correctly to load SWZ files.

### Expected behavior

The swf should load the swz file and function properly.

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Google Chrome 105.0.5195.127

### Additional information

Info Log Level:
![Screenshot from 2022-10-05 10-06-23](https://user-images.githubusercontent.com/29206584/194081320-8150508e-f588-456f-b240-72bcd0cdbc5b.png)

Network Tab:
![Screenshot from 2022-10-05 10-06-43](https://user-images.githubusercontent.com/29206584/194081322-55ef8eaf-9a96-4842-bbde-6da539ab1b9c.png)

This remains unchanged after #8069. According to conversation on Discord, "Loader doesnt support the applicationDomain field in LoaderContext, which makes it so it cant add the SWZ to the correct domain."

Lord-McSweeney:
#8910 will add support for `applicationDomain` and the `context` parameter. Will it fix the issue?

EmperorBale:
> #8910 will add support for `applicationDomain` and the `context` parameter. Will it fix the issue?

It will allow the SWF to load the SWZ and use the definitions in it, however the loader that `weedtowonder` is using still seems broken for a possibly unrelated reason, so I'll leave it up to @danielhjacobs to decide whether he wants to close this issue or not after the PR is merged.

danielhjacobs:
[...]

PR #7653, opened 2022-08-12
avm2: Stub TimerEvent.updateAfterEvent

Aaron1011:
This is the last stub needed for Wonderputt to reach the
main game screen.

As far as I know, ActionScript cannot observe a frame being rendered,
so implementing this method isn't actually necessary for correctness.

The benefit of implementing this would be to make certain animations
appear smoother, since we'll render changes to the scene without
needing to wait for the next frame. However, actually rendering
*immediately* after the event would require some refactoring -
we have a `&mut UpdateContext` while running timers, but we'd need
to bail out and obtain a `&mut Player`.

PR #9279, opened 2023-01-23
avm2: Stub MouseEvent.updateAfterEvent

danielhjacobs:
Basically a direct copy of https://github.com/ruffle-rs/ruffle/pull/7653

Edit: Fixes an issue with http://new.weedtowonder.org/ (see https://github.com/ruffle-rs/ruffle/issues/8006)

### DIFF
--- a/core/src/avm2/globals/flash/events/MouseEvent.as
+++ b/core/src/avm2/globals/flash/events/MouseEvent.as
@@ -2,6 +2,7 @@
 package flash.events
 {
     import flash.display.InteractiveObject;
+    import __ruffle__.log_warn;
 
     public class MouseEvent extends Event
     {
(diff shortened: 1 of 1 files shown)


core/src/avm2/globals/flash/events/MouseEvent.as
```diff
@@ -75,6 +75,10 @@ package flash.events
             return this.formatToString("MouseEvent","type","bubbles","cancelable","eventPhase","localX","localY","stageX","stageY","relatedObject","ctrlKey","altKey","shiftKey","buttonDown","delta");
         }
 
+        public function updateAfterEvent():void {
+            // TODO - determine when we should actually force a frame to be rendered.
```
Aaron1011, line 79: Can you use `log_warn` (add `import __ruffle__.log_warn;`) to add a stub message to this function?

Aaron1011 (approved):

PR #9284, opened 2023-01-24
Fix DO render tree display and add info to CONTRIBUTING.md

n0samu:
#8964 changed the format of debug strings of display objects from this:
`Stage(GcCell(Gc { ptr: 0x19c9ab557c8 }))`
To this:
`Stage { ptr: 0x205a1342bb0 }`
The change caused Ruffle's DO tree dumping code to crash, so I fixed it to handle the new format. And also added info about the keyboard shortcut to CONTRIBUTING.md since it wasn't there already.

Thanks to @Bale001 for suggesting a more resilient approach.


Bale001 (approved):

Aaron1011 (approved):
I'd prefer that we not try to parse or manipulate `Debug` output at all, but I think this is fine for now.

PR #9301, opened 2023-01-25
wgpu: Ensure we clear before blending an object onto nothing

Dinnerbone:
Fixes the "motion blur" and crazy colours of https://z0r.de/7284


Aaron1011:
Would it be possible to make a small image-test SWF demonstrating this?

Dinnerbone:
> Would it be possible to make a small image-test SWF demonstrating this?

Added!

Aaron1011 (approved):

PR #9302, opened 2023-01-25
tests: Add support for comparing test output against real `flashplayer`

Aaron1011:
This is only supported on Linux for now. It should be straightforward to add macOS support, and Windows support should be possible with a little effort.

We don't store `flashplayer` in the repository:
I've added a script which downloads it from an external site.

There were two tricky issues that I needed to solve:
* Normally, flashplayer won't exit when it reaches the last frame. We could monitor the output and kill flashplayer when we see the expected number of trace lines, but this would make us unable to detect hangs that occur after the last line of output. However we can use `fscommand("quite")` from ActionScript to make flashplayer exit. This unfortunately requires modifying all of our SWFs, but this can be done gradually.
* Flashplayer always writes its trace output to the same file, even if there are multiple instances of the player running. In order to be able to run our tests in parallel, I used Frida (https://frida.re/) to hook into `flashplayer`, and redirect calls to libc `open` to a different file.

This is currently only enabled for the `try_catch` test.


kmeisthax:
So, in my personal experience I've never been able to find the `flashlog.txt` on Windows - but I suspect the problem might be something on my end. Having Linux support for CI is at least Good Enough for most of our tests.

I know there's at least a few tests that don't halt output once finished; those might be an issue as we roll this out.

adrian17:
BTW:

> So, in my personal experience I've never been able to find the flashlog.txt on Windows

I've been using it for 2 years, wish you mentioned this ;D For the record, it's `AppData/Roaming/Macromedia/Flash Player/Logs/flashlog.txt`, assuming it's enabled in mm.cfg.

PR #9305, opened 2023-01-25
Add detail on dont_warn_on_unsupported_content

p0008874:
Yeah, I can't solve what timedemo was used for.


n0samu (approved):
I think it was clear enough without the comment but sure, adding one doesn't hurt

Dinnerbone (approved):

Dinnerbone:
I do think we may want to expand the warning to other content like scaleform, but for now this is accurate I guess!

Issue #9287, opened 2023-01-24
avm2: Mouse events are not propagated correctly


n0samu:
### Describe the bug

Related to #8698.

Consider the following example (includes SWF and FLA): [ClickBlockTesting.zip](https://github.com/ruffle-rs/ruffle/files/10486951/ClickBlockTesting.zip)
![image](https://user-images.githubusercontent.com/71368227/214227786-a233dfcd-6467-4a6a-9f61-3d1118c437c0.png)
Code on the first frame adds a click handler to the green movieclip:
```as
click_mc.addEventListener(MouseEvent.CLICK, clickHandler);

function clickHandler(e) {
	trace("Clicked!");
}
```
The red square is a movieclip, the yellow square is a graphic, the blue square is a bitmap, and the pink shape is a morphshape.

In Flash Player, the only object that blocks the click event from getting to the green square movieclip (and thus the trace statement from appearing) is the red movieclip. Shapes, morphshapes, bitmaps, and text fields should not block the click event. But in Ruffle, all of the objects except the text field block the click event from getting propagated to the green movieclip.

For comparison, I've also created an AVM1 version of this project. Ruffle and Flash Player behave identically in the AVM1 version - nothing blocks the click event from getting propagated, not even the red movieclip.
[ClickBlockTestingAS2.zip](https://github.com/ruffle-rs/ruffle/files/10486989/ClickBlockTestingAS2.zip)

### Expected behavior

Click events should be propagated the same way as in Flash Player.

### Affected platform

Desktop app

### Operating system

Windows 10

n0samu:
Fixed by #9565

PR #9292, opened 2023-01-24
avm1: Don't ignore mouse_pick on static text (fix #9274)

danielhjacobs:

PR #9315, opened 2023-01-26
Add mouse hit tests for avm1 and avm2

Dinnerbone:
Copied from #9287 and adapted into test form (just made it print the coordinate and then wrote the json to follow the coordinates clicked).

Both are ignored as they currently fails. Once we fix that, we can unignore them.


Lord-McSweeney:
What about the fix from #9292 in AVM1?

Dinnerbone:
None of those changes seem to impact these tests, and we need to find another way to prove those changes.

Aaron1011 (approved):

PR #9319, opened 2023-01-27
Stub Loader.close

Lord-McSweeney:
Stubs flash.display.Loader.close.


Lord-McSweeney:
I hope not to make the same mistake that I did before. http://www.waternfire.com/games/sonic.swf has a class which extends `Loader` and overrides `close` (the game currently just stops in the preloader), but I don't think the class is used through the preloader. With this PR, the issue is now to do with unimplemented XML in AVM2.

relrelb (approved):

PR #9329, opened 2023-01-28
Various improvements to test framework

Dinnerbone:
New options:

- `player_options.with_renderer` which defaults to None (no renderer required), It's a struct that currently contains `optional` (whether to require a renderer, or just use it when available) and `sample_count` (default is 1, tests used to run with 4 but have been changed now)
- `output_path` which defaults to `output.txt`, used for custom tests that have multiple outputs
- `image_comparison` (replaces `image`), with fields `tolerance` and `max_outliers` to allow for known deviation across backends & drivers.

Image tests now just have `expected.png` instead of per-backend-per-OS expected images.

All tests now go through the same code so it's now very easy to extend what tests we perform, how we perform them and how we compare the results.


tests/tests/swfs/avm2/bitmapdata_draw/test.toml
```diff
@@ -1,2 +1,7 @@
 num_frames = 1
-image = true
+
+[image_comparison]
+tolerance = 0
```
kmeisthax, line 4: Does this actually work? The new `expected.png` is not aliased, unlike the old `expected-linux-Vulkan.png` etc.

tests/tests/swfs/avm2/displayobject_mask/test.toml
```diff
@@ -1,2 +1,7 @@
 num_frames = 1
-image = true
+
+[image_comparison]
+tolerance = 0
```
kmeisthax, line 4: Same concern as above.

kmeisthax (approved):
I'm a little confused as to why some of the new `expected.png` images work with `tolerance=0`, given that the new images are not antialiased anymore. Or at least the GitHub preview shows me aliased images.

Everything else looks good though.

PR #9339, opened 2023-01-29
Avm2: Implement Stage.invalidate

golfinq:
2nd Attempt at implementing `Stage.invalidate()` taking into account @adrian17's comments [here](https://github.com/ruffle-rs/ruffle/pull/8561#issuecomment-1407485695).


adrian17:
hmm, isn't this ordering backwards? As in, I understand that the test appears to pass, but there's a major distinction between what happens in this PR:
```
EXIT_FRAME ( test_stage =  7  )
<- frame is drawn on screen here (render() gets called)
     (this is optional, as we can actually call run_frame several times in a row without rendering)
<- ...player sleeps...
<- tick() gets called, run_all_phases_avm2() gets called
RENDER ( test_stage =  8  )
ENTER_FRAME ( test_stage =  8  )
```
vs what I assume happens in FP:
```
EXIT_FRAME ( test_stage =  7  )
RENDER ( test_stage =  8  )
<- DO is drawn on screen here
ENTER_FRAME ( test_stage =  8  )
```
As in, it should be triggered at the moment of the actual `player.render()`.
This should be easy to observe, with something like:
```
var tf = new TextField();
tf.text = "111";
addChild(tf);
tf.addEventListener("render", function(){ tf.text = "222"; });
```
In your version, if you reduce the framerate, there'll be a visible single frame where the text displays "111", in FP it'll immediately show "222".

(one thing that blocked me from fully understanding it is that in FP it appears to be triggered by player render, but not by a DO being drawn with `bitmapdata.draw(do)`; this feels weird to me and I keep thinking there could be a counterexample, but if we assume this to be true, this should make the implementation relatively easy)

golfinq:
I agree (after reading some more) that it should be moved. I think moving the logic to the end of  `run_all_phases_avm2` will best match flash. I am thinking this because: I don't see much code between `run_all_phases_avm2` and the next `stage.render`, I wrote up your test and flash showed a 111 then a 222 ([render_test.zip](https://github.com/ruffle-rs/ruffle/files/10543461/render_test.zip)). 

> FP it appears to be triggered by player render, but not by a DO being drawn with `bitmapdata.draw(do)`

I am not sure what you mean by this, do you mean that if someone calls `stage.invalidate` then the subsequent render event will not propagate to bitmap drawings? I haven't explored that part of the code so I can't speak to this at the moment.

P.S: I re-read the notes on `DisplayObject.render` and it seems I should be checking for a minimized or a non-rendering state, which I _think_ I can do with `player.needs_render`

> Note: This event is not dispatched if the display is not rendering. This is the case when the content is either minimized or obscured. 

P.P.S: doesn't setting `needs_render = true` at the end of `run_frame` mean the `player.needs_render` check will always be true?

adrian17:
>  I wrote up your test and flash showed a 111 then a 222

Wait, if that's the case then... ok, I don't understand how it's supposed to be sequenced anymore :sweat: Means my theory above doesn't fit anymore...
Also did another test: if I wrap my code in a setTimeout, then there isn't a "111" flash anymore, it instantly pops in as "222" as I'd expect. Maybe @kmeisthax would have some idea about the ordering.

> which I think I can do with player.needs_render

Yeah, but you may also call it from render(), which will only get called if needs_render anyway.
(that said, again: I don't get when exactly FP triggers it anymore :( )

> P.P.S: doesn't setting needs_render = true at the end of run_frame mean the player.needs_render check will always be true?

run_frame doesn't get called on every tick. Similarly, it's possible for something other than run_frame to trigger a render.

golfinq:
I just rebased and made some changes: namely I put it as close to `stage.render` as I could without running afoul of the borrow checker, I also set the `invalid` state to false every frame so that it wouldn't bleed between frames.

core/src/player.rs
```diff
@@ -1474,6 +1474,11 @@ impl Player {
 
     #[instrument(level = "debug", skip_all)]
     pub fn render(&mut self) {
+        self.update(|context| {
```
adrian17, line 1477: I'd consider avoiding always creating a full context and instead pull the check here:
```rs
        let invalidated = self.gc_arena.borrow().mutate(|_, gc_root| {
            gc_root.data.read().stage.invalidated()
        });
        if invalidated {
            self.update(|context| {
                context.stage.broadcast_render(context);
            }
        }
```
But maybe I'm overthinking it.

core/src/frame_lifecycle.rs
```diff
@@ -76,6 +76,8 @@ pub enum FramePhase {
 pub fn run_all_phases_avm2(context: &mut UpdateContext<'_, '_>) {
     let stage = context.stage;
 
+    stage.set_invalid(context.gc_context, false);
```
adrian17, line 79: I'm not sure we need to do it here, if broadcasting the event already disables it.

core/src/display_object/stage.rs
```diff
@@ -209,6 +213,16 @@ impl<'gc> Stage<'gc> {
         self.0.write(gc_context).loader_info = loader_info;
     }
 
+    // Get the invalidation state
+    pub fn invalidated(self) -> bool {
+        self.0.read().invalidated
+    }
+
+    // Set the invalidation state
+    pub fn set_invalid(self, gc_context: MutationContext<'gc, '_>, value: bool) {
```
adrian17, line 222: set_invalid -> set_invalidated :)

adrian17:
FYI, on Discord I did some more research about it. The consensus with kmeisthax was "it looks like FP avoids dispatching the `render` event until the second render (and equivalently: until the second frame lifecycle loop.)". Since Ruffle doesn't have such a neat way of synchronizing `run_all_phases_avm2` and renders, we think it's okay to keep the current simpler behavior of just "try to dispatch on every render". Let's just add a comment about it to the code, okay?

core/src/frame_lifecycle.rs
golfinq: I put this for situations where the render step is skipped between frames.

adrian17: Hmm, I think that in such a situation we should _not_ disable invalidation. `invalidated` flag should basically mean "we didn't fire the RENDER event yet".
With this line present, there's a risk that someone calls `stage.invalidate()` and the event never actually fires.

kmeisthax (approved):

PR #9351, opened 2023-01-30
desktop: Bump cpal to 0.15.0

Aaron1011:
`build_output_stream` now takes in a `timeout` parameter - I've passed in `None` to keep the current behavior.

cpal addded lots of new `SampleFormat` enum values. For now, I'm just returning an error if we encounter any of them - a quick test showed that desktop audio is still working on my Linux machine.


torokati44 (approved):
LGTM.

Dinnerbone (approved):

PR #9356, opened 2023-01-30
extension: Manifest v3 first draft - works minimally on FF, not Chrome

danielhjacobs:
Will close immediately, since this does not work. Meant to help exemplify what is needed for Manifest V3.

Issues on Firefox: Extension does not activate until Ruffle symbol is clicked (maybe just because it's an unsigned add-on though); opening an SWF in a new tab does not work.

Issues on Chrome: Does not work. Chrome should use `declarativeNetRequest` instead of `webRequestBlocking` (does not work on Firefox). Chrome should use `"service_worker": "dist/background.js"` instead of `"scripts": ["dist/background.js"]`. With those two changes, I still run into https://stackoverflow.com/questions/70474845/inject-javascript-from-content-script-with-a-chrome-extension-v3 and failure of dist/background.js to register as a service worker.

### DIFF
--- a/web/packages/extension/manifest.json5
+++ b/web/packages/extension/manifest.json5
@@ -1,18 +1,19 @@
 {
-    "manifest_version": 2,
+    "manifest_version": 3,
     "name": "Ruffle",
     "version": null, // Filled by Webpack.
     "default_locale": "en",
     "description": "__MSG_description__",
     "homepage_url": "https://ruffle.rs/",
 
-    "browser_action": {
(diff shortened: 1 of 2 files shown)


danielhjacobs:
For declaritiveNetRequest, see https://developer.chrome.com/docs/extensions/mv3/mv3-migration/#when-use-blocking-webrequest

For service worker, see https://developer.chrome.com/docs/extensions/mv3/mv3-migration/#man-sw

PR #9362, opened 2023-01-31
Stub opaque background for DisplayObject.

Lord-McSweeney:
Progresses Bloons 2.

### DIFF
--- a/core/src/avm2/globals/flash/display/displayobject.rs
+++ b/core/src/avm2/globals/flash/display/displayobject.rs
@@ -1077,7 +1077,7 @@ fn set_cache_as_bitmap<'gc>(
     Ok(Value::Undefined)
 }
 
-/// Implements `opaqueBackground`'s getter.
+/// Stubs `opaqueBackground`'s getter.
 pub fn opaque_background<'gc>(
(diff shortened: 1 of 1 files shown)


Dinnerbone:
This should be really easy to actually implement, no?

Lord-McSweeney:
Unfortunately I have practically no Rust experience. If you want to make changes to the branch, feel free to do so!

adrian17:
I'm perfectly happy with having it be stubbed, but let's use the stub macro Dinnerbone merged just earlier today :)

Dinnerbone (approved):

n0samu:
This does get rid of one error message from Bloons 2 but I don't think it's related to why the game is stuck. The error spam is coming from calls to `scene.update()`, since the `scene` variable seems to be declared but never initialized. Yet even if I remove all `scene` method calls, I get nothing but a blank screen, while Flash Player still plays the intro. In conclusion, I have no idea what's going on with this game, but I don't think your PR helps.

adrian17:
(for the future, I think future PRs like this can totally get squashed on merge :) )

Dinnerbone:
Yep regretted it the moment I saw the commits land!

PR #9364, opened 2023-01-31
desktop: Improve --help messages

n0samu:
Mention that SWFs can be loaded via URL as well as file path. And add a listing of the possible values for the `--load-behavior` option.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -48,6 +48,7 @@ enumset = "1.0.12"
 static_assertions = "1.1.0"
 rustversion = "1.0.11"
 bytemuck = "1.13.0"
+clap = { version = "4.1.4", features = ["derive"], optional=true }
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies.futures]
 version = "0.3.26"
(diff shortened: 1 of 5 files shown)


core/Cargo.toml
```diff
@@ -47,6 +47,7 @@ enumset = "1.0.12"
 static_assertions = "1.1.0"
 rustversion = "1.0.11"
 bytemuck = "1.13.0"
+clap = { version = "4.1.4", features = ["derive"] }
```
Dinnerbone, line 50: This should be optional and then enabled by a festure that desktop uses. I think we do this in wgpu create so you can see how it works there

Dinnerbone (changes requested):

Dinnerbone:
In core/Cargo.toml, mark clap as optional:
```toml
clap = { version = "4.1.4", features = ["derive"], optional = true }
```

In desktop/Cargo.toml add `"clap"` to the list of features we use from `core`.
```toml
ruffle_core = { path = "../core", features = ["audio", "mp3", "nellymoser", "clap"] }
```

Then change your derive to:
```rust
[cfg_attr(feature = "clap", clap::ValueEnum)]
[derive(Debug, Clone, Copy)]
```

n0samu:
Thanks! Sorry I didn't have time to figure that out myself yet.

Dinnerbone (approved):

PR #9369, opened 2023-01-31
The Great Restubbening - part 1

Dinnerbone:
We currently have a bunch of ways to stub things:
- Have a comment saying stub, return Undefined
- Have a (warning or error or info) saying (stub or not implemented), return Undefined
- Return an error saying stub

This changes all **rust stubs within AVM1 & AVM2 classes** to a new stub system, which we should use going forwards. It prints a warning when the first time an individual stub is encountered, and keeps track of the stubs encountered for a movie. It also changes things that previously errored, to return Undefined.

Later PRs will aim to achieve:
- A way to list all stubs encountered during a movie, maybe in crash reports or its own menu?
- A way to list all stubs that Ruffle has, in a format of our choosing (maybe make github issues for remaining stubs?)
- Converting AS stubs to use the same system

Individual PRs so as not to do a "change the world" PR, and we can benefit from the system right away.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2362,6 +2362,26 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0717cef1bc8b636c6e1c1bbdefc09e6322da8a9321966e8928ef80d20f7f770f"
 
 [[package]]
+name = "linkme"
+version = "0.3.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "22bcb06ef182e7557cf18d85bd151319d657bd8f699d381435781871f3027af8"
+dependencies = [
+ "linkme-impl",
+]
+
+[[package]]
+name = "linkme-impl"
+version = "0.3.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "83f2011c1121c45eb4d9639cf5dcbae9622d2978fc5e922a346bfdc6c46700b5"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "linux-raw-sys"
 version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3463,6 +3483,7 @@ dependencies = [
  "generational-arena",
  "indexmap",
  "instant",
+ "linkme",
  "lzma-rs",
  "nellymoser-rs",
  "num-derive",
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -8,6 +8,7 @@ repository.workspace = true
 version.workspace = true
 
 [dependencies]
+linkme = { version = "0.3", optional = true }
 byteorder = "1.4"
 bitstream-io = "1.6.0"
 flate2 = "1.0.25"
@@ -64,6 +65,7 @@ timeline_debug = []
 mp3 = ["symphonia"]
 nellymoser = ["nellymoser-rs"]
 audio = ["dasp"]
+known_stubs = ["linkme"]
 
 [build-dependencies]
 build_playerglobal = { path = "build_playerglobal" }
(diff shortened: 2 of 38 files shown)


Aaron1011:
It looks like `linkme` doesn't support WASM: https://github.com/dtolnay/linkme/issues/6

Dinnerbone:
The feature isn't enabled on web (or anywhere by default) - it'll be used by a tool that just extracts the list and dumps it out

Lord-McSweeney:
Can you also implement a native function in ActionScript (probably in the `__ruffle__` namespace) to warn about a stub? There are a couple stubs written in ActionScript, including `Loader.unload`.

Lord-McSweeney:
Also, shouldn't `set_tab_children` return `Value::Undefined` instead of `true`?

Dinnerbone:
> Can you also implement a native function in ActionScript (probably in the `__ruffle__` namespace) to warn about a stub? There are a couple stubs written in ActionScript, including `Loader.unload`.

That's what I meant by:

> Later PRs will aim to achieve:
> - Converting AS stubs to use the same system

Lord-McSweeney:
Ok, thanks! I also noticed that the PR returns `undefined` in places that previously errored, but functions using these may rely on a different type. Here are all the incorrect types I could find:

`DisplayObjectContainer.areInaccessibleObjectsUnderPoint` should return a boolean (should probably return False in this case)
`DisplayObjectContainer.getObjectsUnderPoint` should return an array
`SoundMixer.areSoundsInaccessible` should return a boolean (should probably return False in this case)
`Font.enumerateFonts` should return an array
`StaticText.text` should return a string (empty string in this case)
`Proxy.isAttribute` should return a boolean (I have no idea whether to return true or false here)

Bale001:
For stubbing AS3 methods, I think using metadata tags would look nice, like this:
```as
[Ruffle(Stub)]
public function registerClassAlias(a:String, b:Object):void {}
```

Dinnerbone:
> For stubbing AS3 methods, I think using metadata tags would look nice, like this:
> 
> ```actionscript-3
> [Ruffle(Stub)]
> public function registerClassAlias(a:String, b:Object):void {}
> ```

I have no idea how to get that working, but I'll give it a shot! My plan was just to call `__ruffle__.stubMethod` etc

Dinnerbone:
> Ok, thanks! I also noticed that the PR returns `undefined` in places that previously errored, but functions using these will probably expect a different type. Here are all the incorrect types I could find:
> 
> `DisplayObjectContainer.areInaccessibleObjectsUnderPoint` should return a boolean (should probably return False in this case) `DisplayObjectContainer.getObjectsUnderPoint` should return an array `SoundMixer.areSoundsInaccessible` should return a boolean (should probably return False in this case) `Font.enumerateFonts` should return an array `StaticText.text` should return a string (empty string in this case) `Proxy.isAttribute` should return a boolean (I have no idea whether to return true or false here)

Fixed!

Aaron1011 (approved):

Lord-McSweeney:
It appears that we are stubbing undocumented methods (such as `getMaxSize` and `getDiskUsage`) in AVM1's `SharedObject` (at least, according to https://open-flash.github.io/mirrors/as2-language-reference/index.html). Could we add a comment referencing that they are undocumented methods?

Dinnerbone:
> It appears that we are stubbing undocumented methods (such as `getMaxSize` and `getDiskUsage`) in AVM1's `SharedObject` (at least, according to https://open-flash.github.io/mirrors/as2-language-reference/index.html). Could we add comments indicating that they are undocumented methods?

There's a lot of conflicting documentation for AS1 & AS2, and we implement a whole bunch of stuff that's not well documented.
This PR doesn't change what's stubbed, only how they're stubbed, so can we leave this for a future PR or open an issue? There's a lot of stuff that it could affect but it's not the scope of this PR to address them.

PR #9377, opened 2023-02-01
core: Advance time by 1ms when doing catchup frames

### LINKED ISSUE #9325: Bubble Trouble: Score not shown
### Describe the bug

In [Bubble Trouble](https://www.newgrounds.com/portal/view/80114), after you beat the first level, your score disappears and never reappears.
image

It doesn't show on the Game Over screen, either.
image

### Expected behavior

Your score is supposed to be shown on both screens.
image
image

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

Firefox 109

### Additional information

_No response_

Dinnerbone:
This should fix #9325.

Basically, when we do catchup frames (actual frame rate is below target frame rate so we stop waiting and do multiple frames per tick), always pretend time advanced by 1ms to avoid games from seeing `oldTime == newTime`, preventing divide by zero bugs.

### DIFF
--- a/core/src/player.rs
+++ b/core/src/player.rs
@@ -488,7 +488,7 @@ impl Player {
                 let timer = Instant::now();
                 self.run_frame();
                 let elapsed = timer.elapsed().as_millis();
-                if frame > 0 && elapsed == 0 {
+                if elapsed == 0 {
                     // To make sure that `getTimer()` never returns the same value between frames,
                     // pretend that at least 1ms elapsed if we ran the frame too fast.
                     self.time_offset += 1;
--- a/core/src/player.rs
+++ b/core/src/player.rs
@@ -482,22 +482,30 @@ impl Player {
 
             let max_frames_per_tick = self.max_frames_per_tick();
             let mut frame = 0;
+            let mut last_time_offset = self.time_offset;
 
             while frame < max_frames_per_tick && self.frame_accumulator >= frame_time {
                 let timer = Instant::now();
                 self.run_frame();
-                let elapsed = timer.elapsed().as_millis() as f64;
+                let elapsed = timer.elapsed().as_millis();
+                if frame > 0 && elapsed == 0 {
+                    // To make sure that `getTimer()` never returns the same value between frames,
+                    // pretend that at least 1ms elapsed if we ran the frame too fast.
+                    self.time_offset += 1;
+                }
 
-                self.add_frame_timing(elapsed);
+                self.add_frame_timing(elapsed as f64);
 
                 self.frame_accumulator -= frame_time;
                 frame += 1;
-                // The script probably tried implementing an FPS limiter with a busy loop.
-                // We fooled the busy loop by pretending that more time has passed that actually did.
+
+                // We fooled the movie by pretending that more time has passed that actually did.
                 // Then we need to actually pass this time, by decreasing frame_accumulator
                 // to delay the future frame.
-                if self.time_offset > 0 {
-                    self.frame_accumulator -= self.time_offset as f64;
+                let delta_time_offset = self.time_offset - last_time_offset;
+                if delta_time_offset > 0 {
+                    self.frame_accumulator -= delta_time_offset as f64;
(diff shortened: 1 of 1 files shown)


n0samu:
Can confirm that the Bubble Trouble score is shown!

HaimZik:
Time should only be advanced when run_frame take 0ms.

Dinnerbone:
> Time should only be advanced when run_frame take 0ms.  

Measuring time to see how long it took actually costs perf in Web, we've learnt that before. I think it's completely fine to assume 1ms for now until its an issue

HaimZik:
> > Time should only be advanced when run_frame take 0ms.
> 
> Measuring time to see how long it took actually costs perf in Web, we've learnt that before. I think it's completely fine to assume 1ms for now until its an issue

Time is measured already before calling tick and also in
`let elapsed = timer.elapsed().as_millis() as f64;`
For the first frame you can just check that `dt>0`

HaimZik:
Looks fine now.
But note that the another cause of #9325 remain.  
There is still incorrect behavior under frame capping conditions: on desktop vsync or frame capping by settings and frame capping on chrome. Under such conditions frame catchup always happen and behavior is diverged from flash as I explained in the issue.  
On desktop with frame capping/vsync 60hz monitor and 120fps swf frame_accumulator is above frame time so there be one catchup frame, time_til_next_frame return the frame time, which mean ControlFlow::WaitUntil will return from sleep 8ms too early or worse not sleep at all.

core/src/player.rs
```diff
@@ -477,22 +477,30 @@ impl Player {
 
             let max_frames_per_tick = self.max_frames_per_tick();
             let mut frame = 0;
+            let mut last_time_offset = self.time_offset;
 
             while frame < max_frames_per_tick && self.frame_accumulator >= frame_time {
                 let timer = Instant::now();
+                let elapsed = timer.elapsed().as_millis();
```
HaimZik, line 484: elapsed() right after Instant::now will always result in 0.

HaimZik (changes requested):

HaimZik (approved):

core/src/player.rs
```diff
@@ -477,22 +477,30 @@ impl Player {
...
             let max_frames_per_tick = self.max_frames_per_tick();
             let mut frame = 0;
+            let mut last_time_offset = self.time_offset;
 
             while frame < max_frames_per_tick && self.frame_accumulator >= frame_time {
                 let timer = Instant::now();
                 self.run_frame();
-                let elapsed = timer.elapsed().as_millis() as f64;
+                let elapsed = timer.elapsed().as_millis();
+                if frame > 0 && elapsed == 0 {
```
relrelb, line 486: Why `frame > 0`?

core/src/player.rs
```diff
@@ -477,22 +477,30 @@ impl Player {
 
             let max_frames_per_tick = self.max_frames_per_tick();
             let mut frame = 0;
+            let mut last_time_offset = self.time_offset;
```
relrelb, line 480: Shouldn't `self.time_offset` be always 0 outside the `while` loop below?

core/src/player.rs
Dinnerbone: Removed, was leftover from the last approach

core/src/player.rs
Dinnerbone: In theory it could be changed outside the tick loop by things like ExternalInterface

core/src/player.rs
```diff
@@ -488,7 +488,7 @@ impl Player {
                 let timer = Instant::now();
                 self.run_frame();
                 let elapsed = timer.elapsed().as_millis();
-                if frame > 0 && elapsed == 0 {
```
HaimZik: Removing this check mean that if the first frame took 0ms then the next non catchup frame will start 1ms late(unless frame_accumulator is exactly frame_time).

HaimZik (changes requested):

PR #9378, opened 2023-02-01
avm2: Implement ScopeChain caching

Bale001:
Implements the ability for ScopeChain's to cache their results for better performance. Currently to avoid any problems with dynamic properties, caching is disabled for any ScopeChains that contain a `with` scope.

### DIFF
--- a/core/src/avm2/property_map.rs
+++ b/core/src/avm2/property_map.rs
@@ -134,6 +134,25 @@ impl<'gc, V> PropertyMap<'gc, V> {
         }
     }
 
+    pub fn insert_with_namespace(
+        &mut self,
+        ns: Namespace<'gc>,
+        name: AvmString<'gc>,
+        mut value: V,
+    ) -> Option<V> {
+        let bucket = self.0.entry(name).or_default();
+
+        if let Some((_, old_value)) = bucket.iter_mut().find(|(n, _)| *n == ns) {
+            swap(old_value, &mut value);
+
+            Some(value)
+        } else {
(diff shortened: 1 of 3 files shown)


Bale001:
BTW: If we ever implement an optimizer such as the one in #7920, we can actually do something even better than this since it would then be possible to resolve everything **before** the function is even called.

adrian17:
> BTW: If we ever implement an optimizer such as the one in https://github.com/ruffle-rs/ruffle/pull/7920, we can actually do something even better than this since it would then be possible to resolve everything before the function is even called.

I'm not 100% sure that's true.
This is related what we talked about in the past - a lot of lookups we do (op_coerce/op_astype/op_istype or resolve_type) shouldn't use the scope chain at all - in fact, it's incorrect to do so and it may be possible to construct some evil code that breaks current Ruffle due to this. Types (and more?) should instead use some global (not sure if fully-global or abc-global or what) registry that's not recursive at all.
(this would also let us remove the `resolve_class_private` weirdness)

The main(?) difference from optimization POV is that most/all the lookups using this global registry can be fully resolved eagerly (like, you can resolve function arg types and store them forever). However the "real" `findproperty` lookups might be able to optimize to getting a specific scope with `getglobal/outerscope` (not sure if even that it possible 100% of time), but you still need to `getproperty/getslot` the actual value at runtime, as the value can be modified at runtime.

Dinnerbone:
Profiled the box2d demo before (red) and after (yellow) this change:
image

About 8% faster!

PR #9385, opened 2023-02-01
The Great Restubbening - part 2: Actionscript

Dinnerbone:
@Bale001 wanted to look into using metadata for this but until then here's a direct approach!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -307,9 +307,11 @@ dependencies = [
  "convert_case",
  "proc-macro2",
  "quote",
+ "regex",
  "serde",
  "serde-xml-rs",
  "swf",
+ "walkdir",
 ]
 
 [[package]]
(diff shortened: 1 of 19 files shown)


Lord-McSweeney (approved):

danielhjacobs:
What about https://github.com/ruffle-rs/ruffle/blob/master/core/src/avm2/globals/flash/utils/Dictionary.as#L10 ?

Aaron1011 (approved):

Lord-McSweeney:
`TimerEvent.updateAfterEvent` doesn't appear to be stubbed: https://github.com/ruffle-rs/ruffle/blob/7c46fb207b2ec5e2a3e0d49f7050ef49575a1880/core/src/avm2/globals/flash/events/TimerEvent.as#L14

PR #9392, opened 2023-02-02
wgpu optimisations - vertex attributes and shared mesh buffers

Dinnerbone:
In games with lots of individual meshes, like box2d demo, rendering is about 33% faster
image

Registration (or replacement) of meshes is about 60% faster
image

Heartcore is about 8% faster, but it doesn't have that many individual meshes
image

### DIFF
--- a/desktop/src/main.rs
+++ b/desktop/src/main.rs
@@ -52,7 +52,7 @@ thread_local! {
 #[cfg(feature = "tracy")]
 #[global_allocator]
 static GLOBAL: tracing_tracy::client::ProfiledAllocator<std::alloc::System> =
-    tracing_tracy::client::ProfiledAllocator::new(std::alloc::System, 100);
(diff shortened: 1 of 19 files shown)


desktop/src/main.rs
```diff
@@ -52,7 +52,7 @@ thread_local! {
 #[cfg(feature = "tracy")]
 #[global_allocator]
 static GLOBAL: tracing_tracy::client::ProfiledAllocator<std::alloc::System> =
-    tracing_tracy::client::ProfiledAllocator::new(std::alloc::System, 100);
+    tracing_tracy::client::ProfiledAllocator::new(std::alloc::System, 0);
```
Aaron1011, line 55: Did you mean to commit this

Dinnerbone: Yes, it's its own commit.

Aaron1011 (approved):

PR #9400, opened 2023-02-03
Hook up Quality in wgpu and add --quality option to desktop

Dinnerbone:
The big caveat to this is that Medium quality is supposed to be 2x msaa, but that's not as well supported as 4x msaa - so it may end up looking like Low quality sometimes. WGPU also doesn't support 16x MSAA right now so those will turn into 8x.

In the future we may switch from MSAA to our own downscaling methods, which might solve those.


danielhjacobs:
The issue being since autohigh and autolow don't exist in the AS, they won't be set properly if set on the embed since quality is passed as-is. I assume it may be best to just treat autohigh as high and autolow as low.

Dinnerbone:
MSAA absolutely impacts speed, but I'm not sure I want to tackle somehow automatically monitoring and adjusting performance in this PR 😅 The behaviour should be identical to what it was before, at least

danielhjacobs:
Yeah, I think technically autohigh and autolow would be ignored (since they'd be considered invalid quality values). They'd both therefore function like high, since that's the default. They both have the potential to function as high already, so that's probably fine.

Dinnerbone:
Created https://github.com/gfx-rs/wgpu/pull/3454 for wgpu x16 support

render/wgpu/src/backend.rs
```diff
@@ -150,11 +147,7 @@ impl WgpuRenderBackend<crate::target::TextureTarget> {
 }
 
 impl<T: RenderTarget> WgpuRenderBackend<T> {
-    pub fn new(
-        descriptors: Arc<Descriptors>,
-        target: T,
-        preferred_sample_count: u32,
-    ) -> Result<Self, Error> {
+    pub fn new(descriptors: Arc<Descriptors>, target: T) -> Result<Self, Error> {
```
Aaron1011, line 150: Could this take in the quality as an argument? That would allow us to avoid re-creating the surface immediately when the command-line `quality` argument is something other than `StageQuality::Low`

Dinnerbone: We have to recreate it (and the actual wgpu Surface it's for) immediately for the size of the swf and other player configs that may affect it. I think there's a better way to do the whole bootstrapping process overall, probably by passing in ` render builder` to the PlayerBuilder, but I'm not sure.

Aaron1011 (approved):

PR #9402, opened 2023-02-04
AVM2 ByteArray fixes & error conversion

Dinnerbone:
This makes bloons2 get to the game (once @adrian17's refactoring is done)


core/src/avm2/activation.rs
```diff
@@ -1332,6 +1332,11 @@ impl<'a, 'gc> Activation<'a, 'gc> {
     ) -> Result<FrameControl<'gc>, Error<'gc>> {
         let args = self.pop_stack_args(arg_count);
         let multiname = self.pool_multiname_and_initialize(method, index)?;
+        avm_debug!(
```
adrian17, line 1331: This feels very duplicate with `avm_debug!(self.avm2(), "Opcode: {op:?}");`.
As in, I understand why you added it - to log the actual names, I also had this need several times. But still, this also makes the log double verbose :/
(and if we're already doing it this way, I'd consider logging them in a slightly more jpexs-like format, at least `Call property`->`callproperty`)

core/src/avm2/bytearray.rs
```diff
@@ -23,6 +25,17 @@ pub enum CompressionAlgorithm {
     Lzma,
 }
 
+pub struct EofError;
+
+impl EofError {
+    pub fn to_avm<'gc>(&self, activation: &mut Activation<'_, 'gc>) -> Error<'gc> {
```
adrian17, line 31: let's make this `#[inline(never)]`.
TBH I wish we didn't need to sprinkle this in every method. In a way it feels more "proper" for the Bytearray object itself to "throw the exception", instead of doing it in the binding. But both ways have their downsides :/

tests/tests/swfs/avm2/bytearray_compress/Test.as
```diff
@@ -0,0 +1,72 @@
...
+        public function Test() {
+            var ba = createByteArray();
+
+            ba.compress();
+            print("compressed", ba, false);
+
+            ba.uncompress();
+            print("uncompressed", ba, true);
+
+            //ba.compress("lzma");
```
adrian17, line 16: why are these commented?

core/src/avm2/activation.rs
Dinnerbone: I'll drop the commit for now. I originally wanted to replace the opcode debug with a custom one that would replace all the names, that feels more proper - it's just so much work 😅

tests/tests/swfs/avm2/bytearray_compress/Test.as
Dinnerbone: Couldn't test on version of flash

core/src/avm2/bytearray.rs
Dinnerbone: Done!

Aaron1011 (approved):

PR #9406, opened 2023-02-04
core: Add option to force stage scale mode

### LINKED ISSUE #5186: Allow overriding the stage scale mode
Hey,

Ruffle had a better "fit to screen" option in the beginning. Now it's more like the Flash Player and it's worse. It was a good feature because it scaled the games that couldn't before. Like "Arcuz" or "Submachine".

I guess the player zoomed better and used the + - mode like the browsers to scale which works on those games. Now you made it more like the Flashplayer and broke it. 

I would highly suggest to role it back to how it worked before as it was a good feature or make an option to the JavaScript api to use "standard" mode or "fit to screen / div" mode. Especially nowadays with big screens non-saleable games in 400x300 are not playable anymore!

n0samu:
Add a config option for desktop and web builds of Ruffle that prevents movies from setting the stage scale mode:
- Command line option `--force-scale` for the desktop app.
- `forceScale` boolean in `window.RufflePlayer.config` for web.
The player's `set_scale_mode` function now allows setting and locking the scale mode at the same time. Thanks to this, users can set a custom scale mode on web via the `scale` embed attribute and lock it via the `forceScale` option, preventing movies from changing it.

Implements the feature described in and closes #5186.

One example game that is helped by this change is Nanaca Crash, which by default uses `Stage.scaleMode = "noScale"`. Here are some screenshots what's possible with this PR:
image
image
image


relrelb:
Do we really need to introduce a dedicated `forceScale` option? Maybe we can just use the existing `letterbox` option instead?

core/src/player.rs
```diff
@@ -1932,6 +1923,8 @@ pub struct PlayerBuilder {
 
     // Misc. player configuration
     autoplay: bool,
+    scale: StageScaleMode,
```
relrelb, line 1926: IIUC `scale` and `force_scale` fields can be merged into a single `forced_scale_mode: Option<StageScaleMode>` field, where `None` means no scale mode is forced and `Some(...)` means, well, some scale mode is forced.
I see it This is more idiomatic and it prevents the situation where the value of `scale` is meaningless when `force_scale` is `false`.

web/src/lib.rs
```diff
@@ -554,6 +558,13 @@ impl Ruffle {
                     .and_then(|q| StageQuality::from_str(&q).ok())
                     .unwrap_or(default_quality),
             )
+            .with_scale(
+                config
+                    .scale
+                    .and_then(|s: String| StageScaleMode::from_str(&s).ok())
```
relrelb, line 564: nit: I think the `String` type annotation is not needed here (like it's not needed for `StageQuality` above).

core/src/player.rs
```diff
@@ -2083,6 +2078,14 @@ impl PlayerBuilder {
         self
     }
 
+    /// Sets the stage scale mode and optionally prevents movies from changing it.
+    #[inline]
+    pub fn with_scale(mut self, scale: StageScaleMode, force: bool) -> Self {
```
relrelb, line 2083: To avoid confusion, I prefer calling the introduced fields and methods `scale_mode` and not just `scale`.

core/src/player.rs
n0samu: `scale` sets the scale mode that the player starts with. ActionScript can change it at any time. `forceScale` prevents ActionScript from changing it. Flash Player never allowed forcing the scale mode, so whatever you set could be changed by the movie's ActionScript.

n0samu:
> Do we really need to introduce a dedicated `forceScale` option? Maybe we can just use the existing `letterbox` option instead?

No, letterbox is an entirely different feature that we created for Ruffle, separate from Flash's stage scale modes. It controls whether content outside of the stage area is shown or blocked out with the letterbox. Letterboxing can be applied to any stage scale mode except `exactfit`.

core/src/player.rs
n0samu: I've renamed the fields and methods accordingly, besides the user-facing options, which I wanted to name consistently with the Flash embed option (`scale`): https://helpx.adobe.com/flash/kb/flash-object-embed-tag-attributes.html

relrelb (approved):

PR #9422, opened 2023-02-05
web: add mobile virtual keyboard using prompt()

n42k:
Adds a simple virtual keyboard for mobile using `prompt()`.
Does not use the system keyboard, nor third party libraries.

Usage instructions:
Long press where you wish to type, then choose `Virtual keyboard (experimental)` in the context menu. Fill in the prompt and press OK. Whatever was typed in the prompt will be passed to the flash application.

Tested in [Adventure Quest](https://aq.battleon.com/game/web?launchtype=medium) in Chrome (Android & Windows) and Kiwi Browser (Android, as extension).
Tested in [Typing of the Living Dead](https://www.quickflashgames.com/games/typing-of-the-living-dead/) in Chrome (Android & Windows).

Some screenshots of it working on Android in Kiwi Browser using the extension:
Screenshot_20230205_173445_Kiwi Browser
Screenshot_20230205_173528_Kiwi Browser

### DIFF
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -810,6 +810,25 @@ export class RufflePlayer extends HTMLElement {
         }
     }
 
+    private async openVirtualKeyboard(): Promise<void> {
+        // Use simple prompt for now. Ideally the system keyboard would come up,
+        // possibly with a library for fallback (e.g. simple-keyboard)
+        const string = prompt("String to type:") ?? "";
+
+        const sleep = (ms: number) => new Promise((r) => setTimeout(r, ms));
+
+        for (const char of string) {
+            for (const type of ["keydown", "keyup"]) {
+                const event = new KeyboardEvent(type, {
+                    key: char,
+                    bubbles: true,
+                });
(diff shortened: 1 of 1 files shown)


danielhjacobs:
Copied from Discord:

> Started looking into this myself. Apparently this is a thing: https://javascript.info/keyboard-events#mobile-keyboards. Fun. 
> I can't find a working alternative to get the real virtual key pressed on Android Chrome. If alternatives work, apparently it's a bug.
> So I'm not sure this can be done.
> Tried keyup, keydown, and keypress. keyup and keydown have the same results: Nothing for e.code. "229" for e.keyCode and e.which. "Unidentified" for e.key. keypress does not fire. 
> Only way I can think to do this is to use a input event. Use its e.target.value value as the key to trigger the keyup and keybown KeyboardEvent (if its value is empty, send a Backspace key). Clear the input value at the end of the input event. 
> Converting character to https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key doesn't seem super simple though.

n0samu:
Superseded by #9461. Thanks for your help!

PR #9423, opened 2023-02-05
Log full avm2 errors when they fall through root level activations

Dinnerbone:
Split off from https://github.com/ruffle-rs/ruffle/pull/9402

### DIFF
--- a/core/src/avm2.rs
+++ b/core/src/avm2.rs
@@ -299,7 +299,8 @@ impl<'gc> Avm2<'gc> {
 
     /// Dispatch an event on an object.
     ///
-    /// This should be a root activation. As such, the error is a message and not something actionable.
+    /// This will become its own self contained activation.
+    /// As such, the error is a message and not something actionable.
     ///
     /// The `bool` parameter reads true if the event was cancelled.
     pub fn dispatch_event(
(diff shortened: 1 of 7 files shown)


core/src/avm2.rs
```diff
@@ -213,15 +213,18 @@ impl<'gc> Avm2<'gc> {
 
     /// Dispatch an event on an object.
     ///
+    /// This should be a root activation. As such, the error is a message and not something actionable.
```
adrian17, line 216: This creates a "new local root" activation, but the comment that it "should" be a root sounds weird. This isn't a VM-wide root activation, as `a.addChild(b)` is enough to dispatch a new event from inside an existing activation.

adrian17 (approved):
Comment aside, I think it's okay?
BTW: One thing I recently tripped on a lot is that we don't actually log anything if an error happened in a test, in a script initializer. So the trace just cuts off, without any error message despite us "throwing" an Err from a script. The usual path here is `symbol_class() -> get_defined_value() -> script.globals() -> runs the script initializer`
Do you think you could take a look at it? If not soon, I'm okay with having this PR merged as-is too.

Dinnerbone:
Pushed fix for the comment. I won't be able to do root-level errors in tests for this PR but I'll think on it, it'll need some changes to allow something external to fetch those errors.

PR #9426, opened 2023-02-05
avm2: addFrameScript allows null to unset the script

Dinnerbone:
This progresses https://www.kongregate.com/games/JavelinOU/tdp4-team-battle (which is now blocked on XML)


core/src/display_object/movie_clip.rs
```diff
@@ -2051,14 +2051,15 @@ impl<'gc> MovieClip<'gc> {
...
     ) {
-        let mut write = self.0.write(context.gc_context);
+        let frame_scripts = &mut self.0.write(context.gc_context).frame_scripts;
+        frame_scripts.retain(|fs| fs.frame_id != frame_id);
 
-        write
-            .frame_scripts
-            .push(Avm2FrameScript { frame_id, callable });
+        if let Some(callable) = callable {
+            frame_scripts.push(Avm2FrameScript { frame_id, callable });
```
adrian17, line 2061: if there can only be 1 of each, maybe we should turn this into a `Vec<Option<Callable>>?`
Then we wouldn't need an O(n) `retain` and instead just
```
frame_scripts[frame_id] = callable;
```

Dinnerbone: Done!

Aaron1011 (approved):

Issue #8417, opened 2022-10-31
Plant Pong: Cannot break glass bricks


n0samu:
### Describe the bug

Link to the game: http://www.thepencilfarm.com/games/plant_pong/
On level 6 of Plant Pong, hitting a glass brick 3 times is supposed to destroy it. But that does not happen. 

This error is shown in the terminal:
`[ERROR ruffle_core::avm2::events] Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x1640a452c00 }) to handler FunctionObject(FunctionObject(GcCell(Gc { ptr: 0x16448bd0ac8 }))) : RustError("Cannot access property ::pan of null or undefined")`

The relevant AS3 code is under the Brick class:
```
               _loc1_ = new GlassBreakSnd();
               _loc2_ = new SoundChannel();
               _loc3_ = _loc2_.soundTransform;
               _loc3_.pan = this.x / (Settings.SWF_W / 2) - 1;
               _loc2_.soundTransform = _loc3_;
               _loc2_ = _loc1_.play();
               trace("brick broken");
               addEventListener(Event.ENTER_FRAME,playGlassAnimation);
```
`soundTransform` seems to be implemented in Ruffle already, so I don't understand why this problem is occurring.

### Expected behavior

The glass brick should break after being hit 3 times.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

Tested with #8397, which fixes the previous issue with this game.

Lord-McSweeney:
Is `pan` used anywhere else?

n0samu:
Fixed

PR #9427, opened 2023-02-05
avm2: Make sure SoundChannel.soundTransform always returns a value

Dinnerbone:
This progresses Kingdom Rush Frontiers to the title screen.

### DIFF
--- a/core/src/avm2/globals/flash/media/soundchannel.rs
+++ b/core/src/avm2/globals/flash/media/soundchannel.rs
@@ -9,6 +9,7 @@ use crate::avm2::Error;
 use crate::avm2::Multiname;
 use crate::avm2::Namespace;
 use crate::avm2::QName;
+use crate::avm2_stub_getter;
 use crate::display_object::SoundTransform;
 use gc_arena::{GcCell, MutationContext};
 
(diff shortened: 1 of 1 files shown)


core/src/avm2/globals/flash/media/soundchannel.rs
```diff
@@ -88,15 +88,14 @@ pub fn sound_transform<'gc>(
...
-        .and_then(|this| this.as_sound_channel())
-        .and_then(|channel| channel.instance())
-    {
-        let dobj_st = activation.context.local_sound_transform(instance).cloned();
+    if let Some(channel) = this.and_then(|this| this.as_sound_channel()) {
+        let dobj_st = channel
+            .instance()
+            .and_then(|instance| activation.context.local_sound_transform(instance))
+            .cloned()
+            .unwrap_or_default();
```
adrian17, line 96: unfortunately this is a bigger issue, looks like the transform is somehow bound to the channel, not to the instance?
```
	        var a = channel.soundTransform;
	        a.volume = 0.5;
	        channel.soundTransform = a;
	        var b = channel.soundTransform;
	        trace(a == b); // false, so the getter always returns a copy
	        trace(b.volume); // 0.5
```

n0samu:
This fixes #8417! 🎉

Dinnerbone:
It looks like this commit accidentally got lumped in with #9426 which got merged. I'll replace this PR with marking it as stubbed as we don't know what to do with the correct behaviour yet.

Dinnerbone:
Dropping in favour of #9457

PR #9430, opened 2023-02-05
avm2: Convert most errors in get/set/initproperty to AVM errors

adrian17:
Also fixed final/sealed-ness of some core types, as without that my tests didn't pass.
I didn't update `coerce_to_object()` to always throw the exception, as I felt like this should use a better audit pass. Instead I only converted existing uses of `coerce_to_receiver()`.

One unfortunate result is that the new null/undefined errors don't have field names. If you think I should re-add them, let me know.

Also I didn't actually test it on any real content that actually required this - please let me know how it works now :)


core/src/avm2/value.rs
```diff
@@ -858,25 +859,14 @@ impl<'gc> Value<'gc> {
...
+    pub fn coerce_to_object_or_typeerror(
         &self,
         activation: &mut Activation<'_, 'gc>,
-        name: Option<&Multiname<'gc>>,
     ) -> Result<Object<'gc>, Error<'gc>> {
-        self.coerce_to_object(activation).map_err(|_| {
-            if let Some(name) = name {
-                format!(
-                    "Cannot access property {} of null or undefined",
-                    name.to_qualified_name(activation.context.gc_context)
```
Aaron1011, line 873: I think it's really important that we keep the ability to show field names in the error message. One approach would be to store extra data in `ErrorObjectData`, and only display it in the `Debug` impl.

adrian17: I readded it to every message; if we wanna split it to a separate private field of Error/TypeError, let's do it in another PR, okay?

Dinnerbone (approved):

Aaron1011 (approved):

PR #9434, opened 2023-02-06
core: Handle `DirtyState::CpuModified` in `overwrite_cpu_pixels`

Aaron1011:
If we're going to overwrite the CPU pixels with the result of a GPU operation, make sure the GPU texture is up to date with the latest CPU pixels. I've also renamed the method to `overwrite_cpu_pixels_from_gpu` to better reflect how it should be used.

### DIFF
--- a/core/src/bitmap/bitmap_data.rs
+++ b/core/src/bitmap/bitmap_data.rs
@@ -204,9 +204,9 @@ enum DirtyState {
 }
 
 mod wrapper {
-    use crate::avm2::Value as Avm2Value;
     use crate::context::RenderContext;
-    use gc_arena::{Collect, GcCell, MutationContext};
+    use crate::{avm2::Value as Avm2Value, context::UpdateContext};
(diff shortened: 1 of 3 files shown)


core/src/bitmap/bitmap_data.rs
```diff
@@ -271,14 +271,18 @@ mod wrapper {
         // Provides access to the underlying `BitmapData`.
         // This should only be used when you will be overwriting the entire
         // `pixels` vec without reading from it. Cancels any in-progress GPU -> CPU sync.
-        pub fn overwrite_cpu_pixels(
+        // If the CPU pixels are dirty, syncs them to the GPU.
+        pub fn overwrite_cpu_pixels_from_gpu(
```
adrian17, line 275: note: this desyncs the comment in `struct BitmapDataWrapper`.

Aaron1011: I updated the comment

Dinnerbone (approved):

Issue #6581, opened 2022-04-01
cannot read the glyph inside a button 


ousia:
### Describe the bug

The attached SWF file (remove the `.mov` extension, since it is a Flash animation) contains a full screen button using and edit text and a variable that changes values when full or normal screen .

https://user-images.githubusercontent.com/203767/161298330-e4c210bf-f89c-49ca-a4a1-82db9e363c2a.mov

It goes fullscreen and it resizes to normal size.

Ruffle isn’t able to display the glyphs from the font when used inside a button.

If the button contains an outline, it is displayed just fine.

I’m attach a sample file (again, `.mov` extension has to be removed first).

https://user-images.githubusercontent.com/203767/161299883-551dbabd-074b-45a0-814e-d3b952b7aadd.mov

### Expected behavior

Glyph from the button should be displayed, the same way the outline from the button is displayed.

The player from _Adobe_ displays both fine.

Many thanks for your help.

### Affected platform

Desktop app

### Operating system

Any of them.

### Browser

Any of them.

ousia:
Here is what the player from Adobe displays:

![flhp](https://user-images.githubusercontent.com/203767/161443410-49ca64b3-7321-409f-a6b3-dc4b607c19e9.png)

And here is what Ruffle displays:

![rffl](https://user-images.githubusercontent.com/203767/161443414-6ffc3267-eb89-4643-a30c-dae0cafa882a.png)

BTW, sample images come from a slightly modified [SWF file](https://github.com/ruffle-rs/ruffle/files/8405262/arrows.swf.txt) (`.txt` extension added to enable upload to GitHub).

Herschel:
This is the same issue as #2181. The glyph is a text field inside a button, with a variable binding of `arrow_fullscreen_text`. Even though this is a v7 SWF, resolving a text field variable path should act like SWFv5, where buttons are not valid display objects, so `arrow_fullscreen_text` should resolve to `_root.arrow_fullscreen_text`, but Ruffle resolves it on the button as `_root.arrow_fullscreen.arrow_fullscreen_text`.

ousia:
Many thanks for your reply, @Herschel.

So if I don’t get it wrong, this issue may be solved by #5492 (which would also fix #1561, among many others).

Many thanks for your excellent work.

Issue #6370, opened 2022-02-28
Ruffle fails to load levels and sounds SWF files when they are loaded in AS2 script


LukeSaward1:
### Describe the bug

I tried to load [Wubbzy's Amazing Adventure](https://www.speedrun.com/resourceasset/8lldq) in Ruffle and it does not load the levels and sounds required to play. The sounds and levels SWF files are usually loaded in the loading screen, but as the files cannot be loaded, the game does not progress past the title screen.

Steps to reproduce:
- Load **wubbzys-amazing-adventure.swf** in Ruffle.

### Expected behavior

I expected the game to load as it would in the Flash projector or online before Flash shut down and the Wubbzy's Amazing Adventure game page was taken offline.

### Affected platform

Desktop app

### Operating system

Ubuntu 20.04.3

Toad06:
Thanks for the report.

It seems like `MovieClip.loadMovie()` needs to delete/reset the properties that exist on the clip. Here's a sample:

```
_root.createEmptyMovieClip("mc",1);

mc.hello = "WORLD";
mc.getBytesTotal = -1; // <- what is done in Wubbzy's Adventure

trace(mc.hello); // "WORLD" in Flash and Ruffle
trace(mc.getBytesTotal); // "-1" in Flash and Ruffle

mc.loadMovie("empty.swf");

setTimeout(function()
{
   trace(mc.hello); // "undefined" in Flash, "WORLD" in Ruffle
   trace(mc.getBytesTotal); // "[type Function]" in Flash, "-1" in Ruffle
}
,1000);
```

[Sample](https://github.com/ruffle-rs/ruffle/files/8154435/sample.zip)

PR #5492, opened 2021-10-17
avm1: Display objects are indirectly referenced by string path (fix #1513)

Herschel:
Proof-of-concept of changing `avm1::StageObject` to reference its display object via string path instead of `GcCell`. This is the last big gotcha to tackle in AVM1, so making a rough draft to gain an understanding of the behavior. Let's hold off on merging this until after #5339.

Display objects in AVM1 should not actually hold pointers to their linked display object, but instead refer to it indirectly via absolute string path (for example, `_level0.clip.foo`). See https://moock.org/asdg/technotes/movieclipDatatype for more info.

This means:
 * Every property access on an AVM1 display object requires dereferencing the path to grab the actual display object.
 * If that display object is removed from the stage, the AVM1 object still exists, but the the path no longer resolves to a display object, so the reference is "dangling". Any operations on the AVM1 object are noops.
 * If a display object with the same path is later added to the stage, the AVM1 object will once again resolve, and all operations will work using the new display object.

Every movieclip property access `clip.foo` now requires a full resolve of the path, so this could significantly hurt performance -- measurement is needed.

TODO:
 - [ ] Properly handle unload events (#1535).
   * Currently `onUnload` tests fail because the path no longer resolves. 
   * Clips with unload handlers should switch to "off-screen" depths (<0), which allows them to continue to exist for an extra frame. 
 - [ ] Remove `base` from `StageObject` and instead change the `DisplayObject::object` into a `ScriptObject`/`PropertyMap`?
   * We have to resolve the path to a display object for every operation, so it makes sense that doing so should grab the base `ScriptObject` from the display object itself.
   * Forward gets/sets/other methods to this `ScriptObject` after dereferencing the path.
   * Test how watches etc. work on movieclip objects after removing + replacing a clip with the same path.
   * Be careful that `prototype`, `interfaces` etc. work correctly.
 - [ ] Add tests
 - [ ] Investigate performance and potential optimizations.
   * Having benchmarks would be helpful.
   * Idea: `StageObject::path` could store the parsed path segments instead of full string. Something like `{ level: Depth, segments: SmallVec<AvmString> }`.
   * #5339 will help a lot; eventually will let us cache string hashes, etc.

Fixes #993, #1017, #1029, #1114, #1140, #1415, #1513, #3839, #3916, #4404, #4793, #5275, others...

Issue #8213, opened 2022-10-06
Dropdown options not shown when SWF is loaded through loader SWF


n0samu:
### Describe the bug

1. Load this web activity with the Ruffle extension enabled: http://www.webgeo.de/modules/k_660/iframe_content.php?&pageid=4&folder=k_660
2. Click one of the dropdown menus. Notice that no options are shown.
![image](https://user-images.githubusercontent.com/71368227/194279889-dc670825-ac96-43f5-9a43-aacb8904b071.png)

Notice that the page uses a loader SWF: http://www.webgeo.de/modules/k_660/frame/flash/page_wrapper.swf
Which loads the main SWF: http://www.webgeo.de/modules/k_660/flash/k_660_04.swf 

Now load the main SWF using the Ruffle extension. Notice that this time, the options in the dropdown menus appear correctly:
![image](https://user-images.githubusercontent.com/71368227/194280606-e0f60925-3683-468b-af8e-038cdfe882d3.png)

### Expected behavior

In Flash Player, the correct dropdown options are shown regardless of whether the main SWF is loaded from the loader SWF or loader directly:
![image](https://user-images.githubusercontent.com/71368227/194281032-504f86f7-70fb-4b10-b91f-965c8859b222.png)

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Firefox 105

### Additional information

Found while triaging #2708.

ousia:
I guess this might be related to #5492.

Indirect references seem to be problematic (at least in loaders, #1561).

Or am I missing something here?

n0samu:
I think you're right. Feels like every AVM1 problem is a "display objects are referenced by string path" problem, doesn't it? 😭

ousia:
> I think you're right. Feels like every AVM1 problem is a "display objects are referenced by string path" problem, doesn't it?

@Herschel stated almost a year ago (in #5492):

> This is the last big gotcha to tackle in AVM1, so making a rough draft to gain an understanding of the behavior.

The real issue here is that there is nothing beyond the original draft (and it seems this won’t be anything different in the future 😭).

It would be great to finish AVM1 before AVM2 (but that seems to be my personal opinion).

Lord-McSweeney:
When Ruffle tries to execute a `CloneSprite` action to create the new text MCs in the menu, it does a lookup of the path `/my_ldr/contentHolder/me/menu2/mwin/item0` (code is in the inner SWF) to find the source MC to clone from.

[...]

PR #9447, opened 2023-02-07
avm1: Reference display objects indirectly by string path

### LINKED ISSUE #993: [Skeleton Park] Issue with a specific animation
In Skeleton Park, when two dogs meet each other, they begin to fight.
sk1

If you get into the brawl or if you get hit by another dog in the meantime, you will lose a life and a "Continue" button will appear:
- if you don't click this button quickly, the brawl will stop and the two dogs will become visible again. In Flash, they are hidden.
- if you click the button while the brawl still happens, the animation will still be displayed after the level restarts. This does not happen in Flash.
sk2

https://www.y8.com/games/skeleton_park

### LINKED ISSUE #1017: Destroy the evil squares targeting broken
https://www.newgrounds.com/portal/view/565062
The cannons lock on the first square then keep shooting at the same spot once it's destroyed. In the actual flash player, they lock on the following squares as well.

### LINKED ISSUE #1029: Buttons and text fields can not have instance names in SWFv4/5
Buttons and texts can not have instance names in SWF 4 or 5, and it is not possible(?) to access the button instance.

Try this code inside a movieclip that is inside a button:
```
trace("this:");
trace(this);
trace("_x:");
trace(_x);
trace("");

trace("_parent:");
trace(_parent);
trace("_parent._x:");
trace(_parent._x);
trace("typeof _parent:");
trace(typeof _parent);
```
in swf5, traces:
```
this:
_level0.instance1.instance2
_x:
71

_parent:
_level0
_parent._x:
0
typeof _parent:
movieclip
```

vs. swf6:
```
this:
_level0.instance1.instance2
_x:
71

_parent:
_level0.instance1
_parent._x:
68.75
typeof _parent:
object
```

FLA:
[button_v5.zip](https://github.com/ruffle-rs/ruffle/files/5111190/button_v5.zip)

### LINKED ISSUE #1140: Luigi's Mansion silently stops working
Edit 2020-11-29:
The game now works but there's still an unexpected behavior - please see the tests in the comments below.

----

After the ghost has reached the top of the screen, a sound effect plays and then nothing else happens anymore.
No error and no warning in the console.

Expected (Flash Player):
expe

Game:
[luigimansion.zip](https://github.com/ruffle-rs/ruffle/files/5189360/luigimansion.zip)

### LINKED ISSUE #1513: avm1: References to removed clips should be "resurrected" if the name is reused
Consider the following sequence:
1. On frame 1, there is a movieclip with an instance name of "clip". `var clipRef = clip;` on frame 1.
2. Remove the clip on frame 2. `trace(clipRef);` traces the empty string (clip is dead, but not implemented yet, cc #993).
3. Add a completely new clip named "clip" on frame 3. `trace(clipRef);` magically prints `_level0.clip` even though `clipRef` was never updated.

This seems like `avm1::StageObject` must store the canonical path to the movieclip, and will automatically point to a new clip if that path is ever reused. Some research is needed to determine the exact behavior, and hopefully figure out an efficient implementation (worst case is we have to parse the path on every use, which would be bad...!) My hope is that we can limit this check to only when the original clip is removed.

Example SWF using timeline:
[dead-mc-timeline.zip](https://github.com/ruffle-rs/ruffle/files/5504003/dead-mc-timeline.zip)

Example SWF using `createEmptyMovieClip` twice w/ same name and depth:
[dead-mc.zip](https://github.com/ruffle-rs/ruffle/files/5503990/dead-mc.zip)

I believe this causes #1415 and #1509. Related to #993.

### LINKED ISSUE #1989: Zelda Lampshade - Plug doesn't always pull
https://www.newgrounds.com/portal/view/253819/fomat/flash?emulate=flash

Enter the dungeon, then walk up to the outlet and press Space.

In Flash, the outlet is removed, boss defeated, then it says "Wow, that was easy..."

In Ruffle this works sometimes but other times the outlet doesn't move, the boss remains but it still says "Wow, that was easy..", at which point the player is stuck.

### LINKED ISSUE #2575: Error on Xeno Tactic 2 on jeu.cc
I tried to play Xeno Tactic 2 in-browser here: https://jeu.cc/xeno-tactic-2
Game loads and seems fine but enemies take no damage so it's pretty sad
Also the ad button "link open in a new window" in the top left doesn't open anything

EDIT: Also tried after downloading "Nightly 2021-01-16" and running on my computer but it's the same

### LINKED ISSUE #2732: Ruffle doesn't know how to count in "CoolioBeat2.swf", I got x2 combo when I should get x4, got x50 when I should get x100, etc...
This game works, apart from missing text outlines, the core gameplay mechanics (hit detection, etc.) works correctly. 
But I keep getting a very low score.

Turns out, the combo isn't counting correctly. 
When I hit 10 notes in a row, the combo count should be like:
x1 x2 x3 x4 x5 x6 x7 x8 x9 x10
But in Ruffle, it's like:
x0 x1 x1 x2 x2 x3 x3 x4 x4 x5

I don't find this problem happening in any other games, tried "CoolioBeat.swf" and "CoolioDJ.swf", the combo counter works correctly.
(In Coolio DJ the missing text outline makes it impossible to see the combo count, but I can just all combo a song and compare the score to using Adobe Flash Player, they are identical.)

(I didn't report this problem instantly, waited a week or two, tried few different night releases. This problem exists consistently. )
(I should try the online version and verify once more. but I can't get Ruffle.xpi to work at https://www.newgrounds.com/portal/view/424469 )

### LINKED ISSUE #2840: https://games.arantius.com/game.86.8+planet.html
At the end of level, the game does not end the game and does not pass to the next level.

image

### LINKED ISSUE #3118: The CPU is breaking the Gomoku rules in Wuzi Chess.
**Describe the bug**
wuzichess

When playing Wuzi Chess (same as Gomoku rules), on a CPU turn, the CPU stone is overlaid at the same location on the player's stone.

It was difficult to explain, so I attached a Gif that was slowed down.
In the Ruffle version, **you can see that the white stones are overlaid with black stones.**

There is no such problem in the Original Version, and it happens very frequently in Ruffle.
The problem can be easily reproduced by placing the stone diagonally along the line of the stone placed by the CPU.

[WuziChess.zip](https://github.com/ruffle-rs/ruffle/files/5936649/WuziChess.zip)

**Expected behavior**
The game should proceed according to the normal Gomoku rules.

**Is the problem with the Ruffle desktop app, extension, or self-hosted version?**
Probably happens on all platforms

**What platform are you using?**
 - Desktop: Windows 10 x64, Google Chrome (Latest), Ruffle Nightly 2021-02-06

### LINKED ISSUE #3604: Bejeweled.swf content works but isn't quite right ...
File location is https://thetruthabouthemp.com/hempity/flash/Bejeweled.swf - Working Demo Page is 
https://thetruthabouthemp.com/arcade/Bejeweled.html 
[Bejeweled.swf.zip](https://github.com/ruffle-rs/ruffle/files/6119495/Bejeweled.swf.zip)
Testing with Ruffle Nightly 3-10-2021 on Firefox

Problem #2710 Text Labels on Buttons Missing

And a little problem with the game logic. When the game reaches "No More Moves" and "Game Over" there is one button to click (I presume it is supposed to read "Continue") - when a new Board is Spawned it has "No More Moves" left ... and the game is over in 3 seconds with a score of zero. Click the button again to watch it spawn the exact same fatal board. Again. And again. 

If i had to make a stab of a guess I'd say there was a problem generating a 'random seed' for the new board.

This should be a popular game once the labels return and new games spawn OK. I play with it regularly even while broken in my little arcade / Ruffle Testbed at https://thetruthabouthemp.com/arcade/ 

(Chuck Jones last cartoon project in 2001 "Thomas Timberwolf" was 13 cartoons written in SWF and I found them and hosted them and ruffle WORKS PERFECTLY on ALL of Them. Kudos!!!! https://thetruthabouthemp.com/arcade/wolf/index.html )

### LINKED ISSUE #3839: AVM1, removeMovieClip should delete ALL props from target
Hi, i'm from AwayFL and we fix similar bug in our runtime.
I spent a lot of time to figure out why this was happened, and found one of source.

**Describe the bug**
`removeMovieClip` should destroy MC with deleting all properties from target.
`toString` after this should return empty string

**Example**
```
var mc = this.createEmptyMovieClip("test", 20);
trace('create mc:' + mc); // => 'create mc: _level0.test'
mc.filed = true;
trace('assign field:' +mc.filed); // => true
mc.removeMovieClip();
trace('remove mc:' + mc); // => ''
trace('read field:' + mc.field); // => read field:undefined
mc.filed = true;
trace('assign field again:' +mc.filed); // => assign field again:undefined
```

**Expected behavior**
All properties return 'undefined' when try to read it and assignments not update it.

**Is the problem with the Ruffle desktop app, extension, or self-hosted version?**
Ruffle demo site

**Why this is so important?**
To many games use a this behaviour for track a livetime of MC . They store a MSc or their containers in array and check a fields like `active` from MC in `update` loop, then `splice` MCs that is invalid.

(code from real game, i can't share game)
image
image

[removeMovieClipTest.zip](https://github.com/ruffle-rs/ruffle/files/6220381/removeMovieClipTest.zip)

### LINKED ISSUE #3916: Kodachrome - Looping Sound
https://www.newgrounds.com/portal/view/600078

Playing Kodachrome in Chrome on Win 10 with Ruffle, there is a persistent looping sound that starts at the very beginning of the game and it doesn't exist when playing in a traditional Flash player.

### LINKED ISSUE #4118: Nothing stands in the way of your victory in 'Super Pang'.
**Describe the bug**
[Super Pang.zip](https://github.com/ruffle-rs/ruffle/files/6371812/Super.Pang.zip)
super-pang

(Refer to the attached gif at the top)

'In Super Pang: The island tournament' game, bouncing balls do not appear on the stage.
Therefore, when you start the game, the game determines that all balls have been removed, and the stage is cleared.

You can immediately check the problem by pressing the space bar after starting the game.

**Expected behavior**
Bouncing balls should appear on the stage depending on the difficulty level.

**Is the problem with the Ruffle desktop app, extension, or self-hosted version?**
Checked in self-hosted version.

**What platform are you using?**
 - Windows 10 20H1 x64
 - Google Chrome (latest)
 - Ruffle Nightly 2021-04-25

### LINKED ISSUE #4404: (Little Soldiers) The game freezes when the first stage is completed.
## Describe the bug
image
After completing the first stage in Little Soldiers, the game freezes without a clear screen.

[little-soldiers.zip](https://github.com/ruffle-rs/ruffle/files/6527579/little-soldiers.zip)

## Expected behavior
image
The following screen is displayed and should continue

### Is the problem with the Ruffle desktop app, extension, or self-hosted version?
Tested on self-hosted.

## What platform are you using?
 - OS: Windows 10 20H2 x64
 - Browser: Chrome Latest
 - Ruffle: 2021-05-23 Nightly

### LINKED ISSUE #4793: Balls to the Brawl - Doesn't advance when enemy ._x == null
https://www.newgrounds.com/portal/view/734315/format/flash?emulate=flash

Once you defeat your first batch of enemies, the level fails to advance. I talked to FuturecopLGF about it and it appears to be how the game is checking for null values. Here is his description:

"In terms of playing through the game, yes, I'm seeing the same problem you're seeing where the screen doesn't unlock and progress with GO> when killing the first enemy. Not localized to the first level, either, seems to be happening for them all, whether enemies start on the initial screen or spawn within a bit of initial walking.

Let's see, digging through my old files...I'll try to sum it up. Each level is a frame on the timeline that the header goes to and stops on, where something like this is running on that frame:

---------

HandleEnemies = function()
{

  //Monsters in scene holds references to all monsters you're currently fighting against
  for (counter=0;counter<MonstersInScene.length;counter++)
  {
  
    //run the reference AI function that is assigned to them
    MonstersInScene[counter].gotoFunction();
    
    //determine if a monster has been killed but still exists in the array
    if(MonstersInScene[counter]._x == null)
    {
    
      //monster is dead and gone, remove from array
      MonstersInScene.splice(counter,1);
      counter = 0;
      HandleChainKill();
      
      if (MonstersInScene.length == 0)
      {
      
        //last enemy defeated, display GO mark
        _root.attachMovie("GOSIGN", "GOSIGN", 999993, {_x:860, _y:400});
        
        //determine if was a perfect round
        if (perfectHits == 0)
        {
        
          points = points + perfectpoints;
          _root.attachMovie("PARFECT", "PARFECT", 999995, {_x:80, _y:280});
        
        }
        perfectpoints = 0;
        perfectHits = 0;
        HandleVariation();

      }
    }
  }
}

onEnterFrame = function()
{
  //tons of handler functions which handle triggers, handle velocities, handle depth rearrangement, etc, which includes
  HandleEnemies();
}

---------

So basically, every frame the HandleEnemies function is being called, which runs the AI code for every monster (enemy) in the scene and also removes monsters from the scene when they have been beaten. Well, technically the monsters have already removed themselves from the scene by calling removeMovieClip on themselves, this is just the function noticing that they've been removed and doing cleanup by removing their null reference from the array. Once it detects that all monsters in the scene have been removed by seeing the empty array, it should spawn the GO sign and unlock progress. Obviously, something has gone wrong.

My guess as to what is going wrong...maybe the way that I was 'detecting' whether a monster referenced in the array no longer exists was really dumb and is being misinterpreted, being the line if(MonstersInScene[counter]._x == null). Other potential culprits might be the splice function being misinterpreted and not properly adjusting the array, as maybe splice doesn't exist in Ruffle? Also, there's the fact that I'm calling the monster's goToFunction when they might have already been reduced to null and Ruffle might be, instead of ignoring that bad code like Flash did, stopping the code in its tracks there with an error, not running the rest of the function."

### LINKED ISSUE #5095: web: Vector Tower Defense - Game bug
One of the more classic Tower Defense Games are not working correctly.

https://www.spiele-umsonst.de/vector-tower-defense-t107.html

You can place the tower but they all just shoot at a certain spot and does not hit or harm the enemies.

### LINKED ISSUE #5275: [Monoliths Mario World] A single fall kills Mario forever.
## **Describe the bug**
In the first stage of Monoliths Mario World, when Mario falls off a cliff, **his life is reduced to infinity and the game freezes.**
There are detailed symptoms in the video below.

[monoliths-mario-world.zip](https://github.com/ruffle-rs/ruffle/files/7147689/monoliths-mario-world.zip)

https://user-images.githubusercontent.com/75052171/132946521-0bb986e4-8649-4c9f-b9a6-71a2471a4ec0.mp4

## **Expected behavior**
After Mario dies, the stage should restart after the game over screen appears.

## **Is the problem with the Ruffle desktop app, extension, or self-hosted version?**
Tested on self-hosted & Desktop.

## **What platform are you using?**
 - OS: Windows 10 21H1 x64
 - Browser: Chrome Latest
 - Ruffle: 2021-09-11 Nightly

### LINKED ISSUE #5277: [Attack of the Puppybots] Laser attack fired from a fixed location
## **Describe the bug**
In 'The Powerpuff Girls: Attack of the Puppybots', a problem occurs when you switch characters (Using keyboard 'X' key) and then attack ('space' key).

The laser (bullet?) fired by the character is fixed at the position where the character was switched. After that, even if the character attacks while moving, the laser will only fire in place.

[powerpuff-girls-attack-of-the-puppybots.zip](https://github.com/ruffle-rs/ruffle/files/7147729/powerpuff-girls-attack-of-the-puppybots.zip)

### Original (Adobe Flash Player)
https://user-images.githubusercontent.com/75052171/132947149-03240da2-1486-4ed2-80d2-861400bf976e.mp4

### Ruffle Nightly 2021-09-11
https://user-images.githubusercontent.com/75052171/132947162-c39c0b12-ea15-41ce-b25e-ee3815a0beab.mp4

## **Expected behavior**
When the character attacks, the laser should be fired based on the character's position.

## **Is the problem with the Ruffle desktop app, extension, or self-hosted version?**
Tested on self-hosted & Desktop.

## **What platform are you using?**
 - OS: Windows 10 21H1 x64
 - Browser: Chrome Latest
 - Ruffle: 2021-09-11 Nightly

### LINKED ISSUE #5711: Weird behavior in Vector TD
### Describe the bug

In the game [Vector TD](https://www.coolstreaming.us/forum/arcade/game/2455.html), there seems to be an issue where the towers will destroy one enemy and then not stop firing, causing money and points to be racked up quickly in this glitched state. This is on the most recent web extension version of Ruffle, which is 2021-11-23 at the time of writing this.

https://user-images.githubusercontent.com/55714746/143133468-44d41ba6-8d3a-4e5f-9235-626f4f29d648.mp4

### Expected behavior

The tower is supposed to stop firing and target another enemy, awarding the player with a normal amount of points and money.

https://user-images.githubusercontent.com/55714746/143133874-31601950-44a6-4b07-b166-37d6376e97d3.mp4

### Affected platform

Browser's extension

### Operating system

Windows 10 21H1

### Browser

Chrome 96

### Additional information

_No response_

### LINKED ISSUE #5923: annoying sound issues BEE BOXING
### Describe the bug

just play the game and we have the issues with sound

but otherwise the game works just fine

https://user-images.githubusercontent.com/64484245/147374942-7ee7a2f0-9ee0-4438-8aac-25b1c236b755.mp4

https://www.bubblebox.com/play/95/bee-boxing.htm

### Expected behavior

for the game to play without annoying sounds  as it is in order to enjoy the game i have to mute my chromebook

### Affected platform

Browser's extension

### Operating system

ChromeOS

### Browser

Version 96.0.4664.111 (Official Build) (64-bit)

### Additional information

_No response_

### LINKED ISSUE #6185: [Bubble Bobble The Revival] Each time press the attack (Space Key), the stage always reset.
### Describe the bug

In **Bubble Bobble The Revival (By Free Online Games)**, after starting the game, if you press the attack key (space bar), the screen will scroll with the initializing of stage 1 at the same time as the attack.

Therefore, the game does not progress and stage 1 is reset.

[Bubble_Bobble_The_Revival.zip](https://github.com/ruffle-rs/ruffle/files/7978128/Bubble_Bobble_The_Revival.zip)

## ✔ Adobe Flash Player (Working)
https://user-images.githubusercontent.com/75052171/151962633-6bf81bb3-ee71-4a37-aefa-19192cafe626.mp4

## ⚠ Ruffle Nightly 2022-01-31 (Not Working)
https://user-images.githubusercontent.com/75052171/151962644-173da0e8-6090-4034-98f8-e9ade98301d4.mp4

image

### Expected behavior

The screen should not come out so the character can shoot the bubbles.

### Affected platform

Self-hosted version

### Operating system

Windows 10 21H2

### Browser

Google Chrome 97

### Additional information

_No response_

### LINKED ISSUE #6203: Trojan War - Bugged after first hit
### Describe the bug

Seems to be a problem with some Tower defense games as well as this one when hitting the first hit it gets in endless mode and doesn't harm anymore. Thought I've seen this behavior at least in Vector Tower Defense and somehwere else.

https://www.newgrounds.com/portal/view/604949?emulate=flash

### Expected behavior

Should hit and not endless loop stuck

### Affected platform

Self-hosted version

### Operating system

All

### Browser

_No response_

### Additional information

_No response_

### LINKED ISSUE #6336: "Extreme Pamplona" doesn't render the background, and crashes after losing.
### Describe the bug

When you start playing, the background (stadium) is briefly rendered for a couple of seconds and then completely disappears. 
image

When you lose, the game just displays a black screen, and doesn't load anymore.

You can play the game here [https://www.crazygames.com/game/extreme-pamplona](https://www.crazygames.com/game/extreme-pamplona)

Ruffle was build from tag nightly-2022-02-19 and was initialized with "trace" log level, and the self hosted version was built with the debug flag. The log contains the data since the load of the game, until the black screen crash.
[extreme-pamplona-log.log](https://github.com/ruffle-rs/ruffle/files/8115281/extreme-pamplona-log.log)

Here are the game files:
[extreme-pamplona.zip](https://github.com/ruffle-rs/ruffle/files/8115528/extreme-pamplona.zip)

### Expected behavior

- The background should be rendered.
- The game should load again after losing.

### Affected platform

Self-hosted version

### Operating system

Windows 11

### Browser

Google Chrome Version 98.0.4758.102 64-bit

### Additional information

_No response_

### LINKED ISSUE #6960: Minecraft tower defense has severe issues with ruffle
### Describe the bug

The bug happens some time after the game starts. The towers start shooting and the monsters are eliminated, however as soon as the time between waves start if there are still any mobs on the screen they become immortal. Forcing an early loose 

The skip wave button cam be used multiple times in-between waves instead than only once every wave.

link to the game: https://www.clickjogos.com.br/jogos/minecraft-tower-defense

### Expected behavior

The mobs should be vulnerable to tower fire at all times and traps too. And the skip wave button should only be capable of being used once every wave.

### Affected platform

Browser's extension

### Operating system

Ubuntu 20.04LTS

### Browser

Mozzila Firefox 100

### Additional information

URL to the game: https://www.clickjogos.com.br/jogos/minecraft-tower-defense

### LINKED ISSUE #7169: Error on https://www.coolstreaming.us/forum/arcade/game/6932.html
# Error Info
Error name: Error
Error message: panicked at 'called `Result::unwrap()` on an `Err` value: "Unable to create VAO"', render/webgl/src/lib.rs:499:50
Error stack:
```
Error: panicked at 'called `Result::unwrap()` on an `Err` value: "Unable to create VAO"', render/webgl/src/lib.rs:499:50
    at a.wbg.__wbg_new_df6e6ab7a65c4c4d (https://www.coolstreaming.us/forum/ruffle.js:1:7026)
    at std::sync::once::Once::call_once::{{closure}}::h6e694fc45f651103 (https://www.coolstreaming.us/forum/d044f0d011060fa3da19.wasm:wasm-function[1075]:0x283e42)
    at std::sync::once::Once::call_inner::h99e47643887ee5a5 (https://www.coolstreaming.us/forum/d044f0d011060fa3da19.wasm:wasm-function[949]:0x25c9fa)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::hb552ca3003b47989 (https://www.coolstreaming.us/forum/d044f0d011060fa3da19.wasm:wasm-function[3823]:0x407344)
    at std::panicking::rust_panic_with_hook::h03201bc01a0ed891 (https://www.coolstreaming.us/forum/d044f0d011060fa3da19.wasm:wasm-function[2971]:0x3a837e)
    at rust_begin_unwind (https://www.coolstreaming.us/forum/d044f0d011060fa3da19.wasm:wasm-function[4031]:0x40ce6e)
    at core::panicking::panic_fmt::hd44f353696296fec (https://www.coolstreaming.us/forum/d044f0d011060fa3da19.wasm:wasm-function[4179]:0x4138fe)
    at core::option::expect_none_failed::hf7dab2706f1a5551 (https://www.coolstreaming.us/forum/d044f0d011060fa3da19.wasm:wasm-function[3407]:0x3f5696)
    at ruffle_render_webgl::WebGlRenderBackend::register_shape_internal::hd0177d89d3ee36e6 (https://www.coolstreaming.us/forum/d044f0d011060fa3da19.wasm:wasm-function[596]:0x1a72c0)
    at <ruffle_render_webgl::WebGlRenderBackend as ruffle_core::backend::render::RenderBackend>::register_shape::h084fcfb1753704f4 (https://www.coolstreaming.us/forum/d044f0d011060fa3da19.wasm:wasm-function[2353]:0x37bb2e)
```

# Player Info
Allows script access: true
Player type: Object
SWF URL: ./arcade/vectortdxTh.swf?tgame=0&pn_gid=6932&pn_license=&pn_checksum=66eddbb4c08bc460ecf944732894d6ea&pn_domain=coolstreaming.us&pn_uname=
Param menu: false
Param movie: ./arcade/vectortdxTh.swf?tgame=0&pn_gid=6932&pn_license=&pn_checksum=66eddbb4c08bc460ecf944732894d6ea&pn_domain=coolstreaming.us&pn_uname=
Param type: application/x-shockwave-flash
Param pluginspage: https://www.macromedia.com/go/getflashplayer/
Param bgcolor: #666666
Param quality: high
Param width: 100%
Param height: 500
Param flashvars: location=./&gamename=vectortdxTh&hash=
Attribute 0: undefined
Attribute 1: undefined

# Page Info
Page URL: https://www.coolstreaming.us/forum/arcade/game/6932.html
SWF URL: https://www.coolstreaming.us/forum/arcade/game/arcade/vectortdxTh.swf?tgame=0&pn_gid=6932&pn_license=&pn_checksum=66eddbb4c08bc460ecf944732894d6ea&pn_domain=coolstreaming.us&pn_uname=

# Browser Info
Useragent: Mozilla/5.0 (X11; CrOS x86_64 14695.85.0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.0.0 Safari/537.36
OS: Linux x86_64

# Ruffle Info
Version: 0.1.0
Name: nightly 2021-05-20
Channel: nightly
Built: 2021-05-20T00:14:52.511Z
Commit: 41c4ac2834ec4c666989c6a745e68347fd4a6ca2

### LINKED ISSUE #7345: Error on https://www.gamesflow.com/jeux.php
# Error Info
Error name: Error
Error message: panicked at 'called `Result::unwrap()` on an `Err` value: "Unable to create VAO"', render/webgl/src/lib.rs:490:50
Error stack:
```
Error: panicked at 'called `Result::unwrap()` on an `Err` value: "Unable to create VAO"', render/webgl/src/lib.rs:490:50
    at c.wbg.__wbg_new_df6e6ab7a65c4c4d (https://www.gamesflow.com/include/p_jeux/ruffle/core.ruffle.f5e16f52204d9499c76d.js:1:7572)
    at std::sync::once::Once::call_once::{{closure}}::h8cc59a0b8a1b03e2 (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[1023]:0x2c3c5c)
    at std::sync::once::Once::call_inner::h51a70f81ef530383 (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[1457]:0x330fa0)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::haa76f48d1b3c2435 (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[4908]:0x4827fd)
    at std::panicking::rust_panic_with_hook::h12b2f0179ff84c92 (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[3229]:0x42594c)
    at core::panicking::panic_fmt::hf0d93ec1ac504cf9 (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[5191]:0x48e8ba)
    at core::result::unwrap_failed::hd5a60243cf946632 (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[4133]:0x46a34f)
    at ruffle_render_webgl::WebGlRenderBackend::register_shape_internal::h151e83d5ba38548e (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[738]:0x23ea43)
    at <ruffle_render_webgl::WebGlRenderBackend as ruffle_core::backend::render::RenderBackend>::register_shape::ha2a1f4d487d61a29 (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[3237]:0x42687e)
    at ruffle_core::drawing::Drawing::render::haf55c1d0bfe64a08 (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[660]:0x219863)
```

# Player Info
Allows script access: true

# Page Info
Page URL: https://www.gamesflow.com/jeux.php?id=2000811
SWF URL: https://www.gamesflow.com/jeux/4dceb19fea291.swf

# Browser Info
Useragent: Mozilla/5.0 (X11; CrOS x86_64 14816.82.0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/103.0.0.0 Safari/537.36
OS: Linux x86_64

# Ruffle Info
Version: 0.1.0
Name: nightly 2022-03-18
Channel: nightly
Built: 2022-03-18T00:17:54.702Z
Commit: e5b6e8c8b7198cd016bb0f37a683718f30a3bcac
Is extension: false

### LINKED ISSUE #7362: Error on https://www.gamesflow.com/jeux.php
# Error Info
Error name: Error
Error message: panicked at 'called `Result::unwrap()` on an `Err` value: "Unable to create VAO"', render/webgl/src/lib.rs:490:50
Error stack:
```
Error: panicked at 'called `Result::unwrap()` on an `Err` value: "Unable to create VAO"', render/webgl/src/lib.rs:490:50
    at c.wbg.__wbg_new_df6e6ab7a65c4c4d (https://www.gamesflow.com/include/p_jeux/ruffle/core.ruffle.f5e16f52204d9499c76d.js:1:7572)
    at std::sync::once::Once::call_once::{{closure}}::h8cc59a0b8a1b03e2 (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[1023]:0x2c3c5c)
    at std::sync::once::Once::call_inner::h51a70f81ef530383 (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[1457]:0x330fa0)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::haa76f48d1b3c2435 (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[4908]:0x4827fd)
    at std::panicking::rust_panic_with_hook::h12b2f0179ff84c92 (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[3229]:0x42594c)
    at core::panicking::panic_fmt::hf0d93ec1ac504cf9 (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[5191]:0x48e8ba)
    at core::result::unwrap_failed::hd5a60243cf946632 (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[4133]:0x46a34f)
    at ruffle_render_webgl::WebGlRenderBackend::register_shape_internal::h151e83d5ba38548e (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[738]:0x23ea43)
    at <ruffle_render_webgl::WebGlRenderBackend as ruffle_core::backend::render::RenderBackend>::register_shape::ha2a1f4d487d61a29 (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[3237]:0x42687e)
    at ruffle_core::drawing::Drawing::render::haf55c1d0bfe64a08 (https://www.gamesflow.com/include/p_jeux/ruffle/6eb258805834d3c235ef.wasm:wasm-function[660]:0x219863)
```

# Player Info
Allows script access: true

# Page Info
Page URL: https://www.gamesflow.com/jeux.php?id=14932
SWF URL: https://www.gamesflow.com/jeux/game-1254985785.swf

# Browser Info
Useragent: Mozilla/5.0 (X11; CrOS x86_64 14816.82.0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/103.0.0.0 Safari/537.36
OS: Linux x86_64

# Ruffle Info
Version: 0.1.0
Name: nightly 2022-03-18
Channel: nightly
Built: 2022-03-18T00:17:54.702Z
Commit: e5b6e8c8b7198cd016bb0f37a683718f30a3bcac
Is extension: false

### LINKED ISSUE #7711: Canyon Defense: Objects not resetted
### Describe the bug

On https://www.browsergames.fm/canyondefense/ game starts perfectly and thanks to the bugfix with regards to #7667 missile shots are fired. However, the missile turret either sticks on its last position and keeps firing on that one or it just sticks there. As a result, the game is not playable.

### Expected behavior

Missile turrets stop firing when an object is destroyed.
Missile turrets target new object when in range.

### Affected platform

Desktop app

### Operating system

macOS latest

### Browser

_No response_

### Additional information

Using nightly build from August 20

### LINKED ISSUE #7728: Intrusion: Repeat ERROR + The Rocket did not update.
### Describe the bug

`[2022-08-22T15:05:58Z ERROR ruffle_core::avm1::activation] Unknown AVM1 opcode: Unknown { opcode: 137, data: [213, 72, 143, 2, 3, 0] }`
https://streamable.com/9m4n63

### Expected behavior

https://streamable.com/hixlfh

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

Swf file: https://m.kongregate.com/games/vapgames/intrusion

### LINKED ISSUE #7846: Baker's Defense
### Describe the bug

https://user-images.githubusercontent.com/75534590/188166051-14cbed3b-241c-4e69-8283-805ecea82631.mp4

### Expected behavior

https://user-images.githubusercontent.com/75534590/188166130-30f1c431-bb2b-486d-9a90-cfd8b44c5d22.mp4

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

SWF FILE: http://farm.stickpage.com/bakersdefensegame.swf

### LINKED ISSUE #7928: Enemy did not remove completely.
### Describe the bug

https://user-images.githubusercontent.com/75534590/189338482-8e9fa942-05ff-4443-9b18-d55fd253fd86.mp4

### Expected behavior

https://user-images.githubusercontent.com/75534590/189338536-be082470-77a4-4bba-bdf0-8480ddd92c5f.mp4

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

Swf file: https://cdn.discordapp.com/attachments/722872096002408529/1014388465783275571/super_mario_world_3.swf

### LINKED ISSUE #8314: Harum Scarum: Characters do not respawn correctly
### Describe the bug

In [The Grim Adventures of Billy and Mandy: Harum Scarum](https://www.arcadehole.com/games/harum-scarum/), the player characters Billy and Mandy do not respawn correctly after dying or beating a level. They seem to respawn below the ground.

After **dying by falling down a pit**, these incorrect behaviors occur:
- Instead of losing just one life, two lives are lost instead. You can hear the "hurt" sound play twice as well.
- Characters respawn at 0 health.
image
- Initially Billy and Mandy are not visible or controllable, but after switching characters with the X button and pressing the Up arrow, they will jump out of the ground.
- After that, Billy and Mandy jump normally when pressing the Up arrow key, but they do not move when pressing the left and right arrow keys, and the spacebar does nothing.

After **beating a level** or **dying by losing health**, these incorrect behaviors occur:
- Billy and Mandy are invisible, but otherwise seem to interact and control normally.
- After switching characters with the X button and pressing the Up arrow, Billy and Mandy will jump out of the ground.
- If you finished the level with Billy selected, from then on the bar along the top will show the wrong character and weapon as selected. That's because every level starts with Mandy as the player-controlled character, but the indicator is not reset at the beginning of the level.
image

### Expected behavior

- Dying should only subtract 1 life
- Characters should respawn normally on the ground at the start of the level
- The indicator at the top should show the correct character and weapon as selected

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

Firefox 105

### Additional information

_No response_

### LINKED ISSUE #8463: Chronotron: Levels 32 and 33 are broken
### Describe the bug

In level 32 of [Chronotron](https://www.kongregate.com/games/scarybug/chronotron), there are some obstacles above the top of the screen that are supposed to periodically move downwards and slam into the floor. But in Ruffle, the obstacles seem to hit the top of the screen and never make it down to the floor. Basically it seems like the collision detection is treating the top of the screen as solid, preventing the obstacles from passing through.

https://user-images.githubusercontent.com/71368227/199654581-07513ca7-b21d-4955-8748-31367f214de6.mp4

Also, level 33 is impossible to complete because the collision between the box and the switch on the left side of the seesaw does not activate the switch. Since the switch does not activate, the platform does not move upwards like it's supposed to.

https://user-images.githubusercontent.com/71368227/199661834-b928f8c0-2ff9-464b-a3ff-c7caee77a606.mp4

Since it would take a lot of time to get all the way to level 33, I've uploaded both the SWF file and a .sol save file that unlocks all the levels up to 33: [chronotron.zip](https://github.com/ruffle-rs/ruffle/files/9925893/chronotron.zip)

### Expected behavior

**Level 32:**

https://user-images.githubusercontent.com/71368227/199655258-11b2070d-a736-40e6-b2da-de54e3eb2e35.mp4

**Level 33:**

https://user-images.githubusercontent.com/71368227/199661977-53f220cf-95c0-4a08-b2ad-57f242b73f8f.mp4

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

I'm guessing that this is a hitTest issue. But unlike some of the recent hitTest issues, this is not a recent regression; Ruffle builds from early 2022 have the same problem.

### LINKED ISSUE #8498: Dirk Valentine: Chain and Rocket
### Describe the bug

https://user-images.githubusercontent.com/75534590/200346482-4798371b-893e-475e-99bc-aa3630736138.mp4

https://user-images.githubusercontent.com/75534590/200346561-e4daf978-e726-40cd-b731-1dace6de5739.mp4

### Expected behavior

https://user-images.githubusercontent.com/75534590/200346779-4c5330be-ea7d-4e3d-9fe0-15b368eb929e.mp4

https://user-images.githubusercontent.com/75534590/200346852-9b8219f1-def9-4fa0-a642-bf8ae8456c68.mp4

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

SWF: http://www.nitrome.com/games/dirkvalentine/dirkvalentine.swf

### LINKED ISSUE #8576: "255 levels of function recursion were exceeded" in MotionBall
Link to the game: https://motion-ball.angelisium.fr/?challenge=1
_Note that the original game is from http://mb2.muxxu.com/ but requires registration._

```
255 levels of function recursion were exceeded in one action list. This is probably an infinite loop.
No more actions will be executed in this movie.
```

The videos below are showing two different ways to quickly trigger this error.

---

https://user-images.githubusercontent.com/14274782/202797389-236b6241-9ef6-4419-8e41-586034d65075.mp4

Note that the crash in this video doesn't happen because I move between rooms quickly. Instead this seems to depend on the number of "bouncy balls" (bb) I met during the game. So if you try to reproduce this step between rooms that don't hold a lot of those balls, this should take more time to crash.

---

https://user-images.githubusercontent.com/14274782/202796996-a222ebf6-ed06-43c6-96d8-fdc75d41cb50.mp4

---

So if you avoid entering too many rooms (so the number of bouncy balls you met remain low) or avoid colliding too much on walls, the game works as intended and it's even possible to complete levels.

Strange issue... Maybe a variant of #3839?

The source code of the game _should_ be available here: https://github.com/motion-twin/WebGamesArchives/tree/main/Frutiparc/Games/motionBall2

### LINKED ISSUE #8765: Bug with swarm towers in Desktop Tower Defence
### Describe the bug

Thanks for Ruffle, it's enabling a few blasts from the past! Using the SWF from #1807, Desktop Tower Defence is very playable using Ruffle's website demo (2022-12-13), except for one bug.

"Swarm Towers" won't stop firing after they first encounter a flying creep. They are useless after this and don't appear to target further creeps.

Easiest way to replicate this is select "Challenge" at main menu then "4K Set Maze", then place a few towers including swarm towers and click "Send next level" several times; flying creeps appear on the 7th wave.

### Expected behavior

Swarm towers should act like the other towers and only free when a creep is in range, then cease firing once creep is out of range or destroyed. They should then target fresh creeps in subsequent waves.

### Affected platform

Online demo

### Operating system

Linux

### Browser

Chromium 108.0.5359.98

### Additional information

_No response_

### LINKED ISSUE #8784: 破game: Enter the next scene rapidly
### Describe the bug

https://user-images.githubusercontent.com/75534590/207669284-e65ba027-c91e-48fc-b883-6518f553003d.mp4

### Expected behavior

Original Flash Player.

https://user-images.githubusercontent.com/75534590/207669664-14da442e-a520-4933-a8d1-3f2f79b42d3f.mp4

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

[1.3423_BreakGame.zip](https://github.com/ruffle-rs/ruffle/files/10230403/1.3423_BreakGame.zip)

### LINKED ISSUE #8857: Cube Colossus just keeps looping?
### Describe the bug

Here's a weird one
Cube Colossus will just jump back to the loading screen after starting the first mission, weird, right?

Flash Video

https://user-images.githubusercontent.com/103858019/208505982-6b713169-fa80-4962-b9a7-61c640415ee6.mp4

Ruffle Video

https://user-images.githubusercontent.com/103858019/208506031-37290dca-d9e2-4406-ac36-93bac152cfe4.mp4

### Expected behavior

the looping bit should never happen

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

Also take a look at the ship trails, not perfectly the same, I guess

### LINKED ISSUE #8872: Problems when playing Frontline Defense: First Assault by Max Games
### Describe the bug

[This link is one of the sites that hosts the Frontline Defense: First Assault.](https://www.gameflare.com/online-game/frontline-defense/)

The majority of the screen is filled with a yellow-tinted screen, such as the area selection map in this screenshot. That's not supposed to happen, and it makes viewing harder. Those borders on the edge are unaffected, though.
Screenshot (113)

Additionally, when a turret in-game stops detecting an enemy in range, it still continuously does the attacking animation, and when a new enemy is in range, the turret is still stuck like that, making it useless for the rest of the game.

### Expected behavior

The game didn't even have a yellow-tinted screen at the center, it was just clear.

When a turret doesn't detect an enemy anymore, it's supposed to stop and then attack again whenever a new enemy is in range.

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

Google Chrome Version 108.0.5359.125

### Additional information

_No response_

### LINKED ISSUE #9135: Pingu Snowball Shootout: snowball sound is repeatly playing and doesnt dissapear
### Describe the bug

I think this is something wrong with the emulator, but there are no errors for me when I open inspect element.
HTML LINK
https://theghostboi64.github.io/pinguwebarchive/uk/snowball-intro.htm

SWF LINK
https://theghostboi64.github.io/pinguwebarchive/uk/m/snowball.swf

Whenever I try to throw a snowball at something, the snowball hit sprite repeats, along with the sound.

I am using unpkg's latest build of ruffle.

### Expected behavior

The snowball with dissapear and the sound will only play one time when hits a character.

### Affected platform

Self-hosted version

### Operating system

ChromeOS 

### Browser

Version 108.0.5359.111 

### Additional information

_No response_

### LINKED ISSUE #9257: [AS1] Reference to MovieClip is incorrect in Ruffle when MC removed
### Describe the bug

I have found a scenario where Flash Player and Ruffle differ. I have produced a MWE [here](https://tmp.skylogic.ca/ruffle_MWE/) demonstrating the issue.

I first encountered this bug with my game, [Zeta Flow](https://zetaflow.skylogic.ca/game/play/14740). When a laser weapon is destroyed, the red laser that shoots out is also supposed to be removed. (in Flash it is removed; in Ruffle it remains).

This uses:
- Flash version 6 (Flash MX)
- Action Script 1

I tested desktop, self-hosted web, and CDN web versions of Ruffle (on Windows 10). I believe all versions of Ruffle affected.

I narrowed down the problem to a reference to a movie clip that doesn't get cleared when the movie clip is removed. In the MWE below, just follow the steps to reproduce the problem.

Online demo: https://tmp.skylogic.ca/ruffle_MWE/
SWF: https://tmp.skylogic.ca/ruffle_MWE/MWE.swf 
FLA: https://tmp.skylogic.ca/ruffle_MWE/MWE.fla

My game (Zeta Flow) relies on this behavior. The laser is supposed to remove itself when the gun is destroyed. In hindsight (15 years later), I could have programmed this differently (and more efficiently) to avoid this. Maybe I'll go and change the code now in order to fix the game! But nonetheless, this highlights a difference in behavior. 

### Expected behavior

The code boils down to this:

```
_root.attachMovie("gun", "gun", 0);
_root.attachMovie("laser", "laser", 1);
_root["laser"].gun_reference = _root["gun"];
_root["gun"].removeMovieClip();
```

In Flash Player, "gun_reference" is correctly empty, but in Ruffle it lives on.

Flash Player:
> _root["gun"] = 
> _root["laser"].gun_reference = 

Ruffle:
> _root["gun"] = 
> _root["laser"].gun_reference = _level0.gun

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

Chrome Version 108.0.5359.126

### Additional information

After writing this I looked at some other open issues, and now I'm realizing this issue may be the same as the one identified [here](https://github.com/ruffle-rs/ruffle/issues/993#issuecomment-675182938) and may possibly be resolved by https://github.com/ruffle-rs/ruffle/pull/5492.

I'll post this issue anyway in case it provides a useful test case.

### LINKED ISSUE #9504: Ricochet Kills: Completing a level will make the gun unable to fire
### Describe the bug

Tested on Ruffle Nightly 2023-02-10
Ricochet Kills SWF used: [509525_ricochet_kills.zip](https://github.com/ruffle-rs/ruffle/files/10708936/509525_ricochet_kills.zip)

This is what the game looks like in the Ruffle Demo. After beating the first level, I was spam clicking, but no bullets were coming out. Pressing Reset or starting a New Game doesn't fix the problem. Relaunching the game will make the gun fire again until a level is completed.

https://user-images.githubusercontent.com/36887711/218131843-d112c81f-7279-409c-8d8c-c76ddcd88aab.mp4

### Expected behavior

This is what the game looks like when I launch it using Flashpoint.

https://user-images.githubusercontent.com/36887711/218131949-006b78c8-17a2-440b-b4e0-4bf26dfa6e32.mp4

### Affected platform

Online demo

### Operating system

Windows 10

### Browser

Chrome 109

### Additional information

_No response_

### LINKED ISSUE #9514: Don't shoot the puppy - the puppy only dies once
### Describe the bug

https://user-images.githubusercontent.com/98996990/218255232-372d5f4e-0429-4040-9fe5-0bce7d62c49d.mp4

the gun can't kill the puppy after losing once
and the result screen will not appear until the puppy goes out of screen
[Don't shoot the puppy.zip](https://github.com/ruffle-rs/ruffle/files/10713489/Don.t.shoot.the.puppy.zip)

### Expected behavior

https://user-images.githubusercontent.com/98996990/218255252-eaeaf3d5-d5f1-4f12-8460-fe667bc9fb73.mp4

the puppy should die immediately when the gun fires

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

_No response_

### LINKED ISSUE #9556: Lettermania: Player stuck on standing animation
### Describe the bug

Tested on Ruffle Nightly 2023-02-13
Lettermania SWF used: [ltrman.zip](https://github.com/ruffle-rs/ruffle/files/10722275/ltrman.zip)

This is what the game looks like in the Ruffle Demo. The character can move perfectly fine, but the character does not change its movement animation.

https://user-images.githubusercontent.com/36887711/218460819-137adc45-76d7-41ec-a3c6-8f5008138006.mp4

### Expected behavior

This is what the game looks like when I launch it using Flashpoint.

https://user-images.githubusercontent.com/36887711/218461083-b3bff4df-8930-4f33-bedf-24c130a88643.mp4

### Affected platform

Online demo

### Operating system

Windows 10

### Browser

Chrome 110

### Additional information

_No response_

### LINKED ISSUE #9800: [AVM1] Sprite display bug
### Describe the bug

In this game, the character's hand floats separately from the body after the player moves for the first time (using the arrow keys).

Demo of expected vs. actual behavior: https://youtu.be/xdFtPl9c0H0
Game Link: https://jiggmin2.com/games/cooties
SWF Link: https://jiggmin2.com/games/cooties/cooties.swf

### Expected behavior

The hand stays attached to the character, following the player's mouse movement.

### Affected platform

Self-hosted version, nightly 2023-03-01.

### Operating system

macOS Mojave

### Browser

Brave (Chromium 109)

CUB3D:
In AVM1, `MovieClip`s are referenced by their path on the stage, which causes some differences in how they are handled compared to `Object`s, namely:
- Removing a clip causes its properties to become `Undefined` and the clip itself to become `""` (empty string)
- Re-creating another clip at the same path, causes existing references to "revive" (their properties become valid again)

This pr takes a different approach to #5492,  as per https://web.archive.org/web/20051224124649/http://moock.org:80/asdg/technotes/movieclipDatatype/: 
"Movie clips are implemented separately from objects internally in the Player, ..."

The combination of this strange behavior and clips having a unique result for `typeof` makes me think that this quote should be taken literally. This pr starts to separates `Value::MovieClip` from `Value::Object`, this new variant stores the path to the targeted object. When coerced to a `String`, this variant results in the path if the object is present on the stage, or an empty string if it is not. When coerced to an `Object` the result is either `Undefined` if the clip has been removed or the backing `Object` if it is not.

(Note that in order to simplify this, `Value::Object` is still used internally, and is only converted to a `Value::MovieClip` when pushed onto the avm stack, this isn't clean but it's a lot simpler. I don't intend for this "hack" to be permanent)

(This pr also fixes a minor issue with `SuperObject` string coercion found while creating tests)

TODO:
- [x] Add more tests
- [x] Why does 9052 preloader no longer work (unload handlers, fixed)
- [x] Why does nothing spawn in 1017 (`SuperObject`)
- [x] Why is mochi loader borked (unload handlers, fixed)
- [x] Bloons (mochi)
- [x] Why is 6203 / 8872 broken (broke `SuperObject`)
- [x] Monkey kick off 
- [x] Nanaca Crash
- [x] Numballs 
- [x] Ninja painter???????? (DO caching)
- [x] Cube Mayhem? (DO caching)
- [x] Cyber chaser (DO caching)
- [x] Kodachrome (audio loop is back)
- [x] Extreme Pamplona (DO caching)
- [x] SMF2 (DO caching)
- [x] Super Pang (This is the same scoping bug as wuzi chess with more indirection)
- [x] Bubble Bobble (events on removed clips)
- [x] Golphysics (unintentionally fixed?)
- [x] 1029 (swfv5 buttons)
- [x] Wuzi chess (scoping)
- [x] 4088 (Going to consider this out of scope for this pr, its big enough) 
- [x] 1561 (going to consider this out of scope for now)
- [x] Refactor
- [x] Dragon Fable regression (more resolving)

[Deep Breath](https://i.imgur.com/FzEz8iF.png)
Fixes #993, fixes #2840, fixes #1513, fixes #1140, fixes #5275, fixes #4793, fixes #4404, fixes #9257, fixes #9135, fixes #5923, fixes #2575, fixes #8765, fixes #7711, fixes #5711, fixes #5095, fixes #7169, fixes #7345, fixes #7362, fixes #6960, fixes #6203, fixes #8872, fixes #1017, fixes #3839, fixes #3604, fixes #6185, fixes #3916, fixes #6336, fixes #4118, fixes #9504, fixes #1029, fixes #2732, fixes #5277, fixes #8576, fixes #1989, fixes #3118, fixes #9514, fixes #9556, fixes #8857, fixes #8314, fixes #8463, fixes #7728, fixes #7846, fixes #7928, fixes #8498, fixes #8784, fixes #9800
(please link anything else that has the same bug as 993 and isn't in this list or the todos)


n0samu:
Although the game is far from working, this fixes the panic in #4932! 🎉 
This is going to be so exciting for so many people 😃

n0samu:
Looks like this also fixes #3839! Can you confirm?

n0samu:
Unfortunately, this seems to break the AVM1 Mochi loader for some games, including [Bloons TD](https://archive.org/details/bloonstd_202011) and [Desktop Tower Defense](https://www.kongregate.com/games/preecep/desktop-tower-defense).

n0samu:
Testing with a version of Desktop Tower Defense modified not to use a Mochi loader, #8765 is fixed!
And many, many other tower defense games with similar issues are also fixed! We'll be able to close all of these issues:
#7711, #5711, #5095, #7169, #7345, #7362, #6960.

On the other hand, Trojan War (#6203) seems to have regressed further. You can no longer place towers, and the panels on the top-left and bottom-right display incorrectly:
image
image

And Frontline Defense (#8872) has largely the same problems as Trojan War. You can't place towers and some info in textboxes shows up as "undefined".

n0samu:
Testing the games in my collection, here are some more regressions I'm noticing:
- In [Ninja Painter](http://www.notdoppler.com/ninjapainter.php), after you grab the paint, a flashing rainbow shape appears in the top-left corner of the screen, and slowly grows larger as time goes on, making the game very unpleasant to play. And the same problem also occurs in Ninja Painter 2.
- In [Cube Mayhem](https://archive.org/details/flash_cubemayhem), nothing happens when the cube hits the goal tile, so levels are impossible to complete.
- In [Cyber Chaser](http://www.notdoppler.com/cyberchaser.php), bullets stop after traveling barely any distance, making it impossible to shoot the boxes or enemies. And there are also bullets traveling across the top of the screen, for some reason.
- [Monkey Kick Off](https://archive.org/details/flash_monkeykickoff) shows a blue box near the top-left corner of the screen.
image
- In [Nanaca Crash](https://megami.starcreator.com/nanaca-crash/), after you click the Play button on the main menu it enters an infinite loop, so the game is unplayable.
- [Numballs](https://www.newgrounds.com/portal/view/563163/format/flash?emulate=flash) gets stuck on the loading screen with this error:
`ERROR ruffle_core::avm1::globals::shared_object: SharedObject::get_local: Invalid character in name`

n0samu:
The SWF provided in #3839 seems to log the correct output now. Some related issues are also fixed: #3604, #6185

But some are not fully resolved yet:
- Kodachrome (#3916) does not continuously loop the text/dialogue sound, but now the text doesn't show up at all and pressing the Down arrow does nothing:
image
- Super Pang (#4118) is still broken.
- Extreme Pamplona (#6336) no longer crashes after you lose, but now if you run for a bit the floor turns partially blue and you fall right through it.
image
- In Super Mario Flash 2 (#7260), Mario doesn't interact at all with Yoshi or any of the enemies anymore.
- Wuzi Chess (#3118) is still broken.

core/src/avm1/value.rs
```diff
@@ -391,6 +398,39 @@ impl<'gc> Value<'gc> {
...
             }
+            Value::MovieClip(path) => {
+                           let mut parts = path.as_wstr().split(b'.').into_iter();
+
+            let mut start = activation.root_object().coerce_to_object(activation).as_display_object();
+
+            // Handle the level id, to support multi-file movies
+            if let Some(root) = parts.next() {
+                if root.to_utf8_lossy().starts_with("_level") {
+                    if let Ok(level_id) = root[6..].parse::<i32>() {
```
moulins, line 409: This should use `WStr::strip_prefix`.

Lord-McSweeney:
Relevant: https://github.com/ruffle-rs/ruffle/pull/6548#discussion_r836713330

n0samu:
I can confirm that Trojan War, Frontline Defense, Monkey Kick Off, Nanaca Crash, and Numballs are working great with the latest changes! 🎉 

@ousia No, those two issues are not fixed. They seem to have somewhat different root causes that would likely need to be solved separately from this PR.
Your question also reminded me to check #8213. It's not fixed either, so I guess our theory about the root cause of the problem was incorrect.

n0samu:
@CUB3D your latest commit seems to have broken most of the tower defense games again to some degree. Vector TD is back to being totally broken, for instance.

ousia:
@n0samu, many thanks for your fast reply.

I thought #5492 was about indirect object references.

So, it seems that some issues with indirect references will need further investigation after this PR is merged.

Many thanks again for your help (and to @CUB3D for the PR).

core/src/avm1/value.rs
```diff
@@ -509,7 +509,7 @@ impl<'gc> Value<'gc> {
         if let Value::MovieClip(path, mc) = self {
             // Check if we can re-use the cached `DisplayObject`, if we can then take this fast path
             if let Some(mc) = mc.upgrade(activation.context.gc_context) {
-                if !mc.read().display_object.removed() || mc.read().display_object.pending_removal()
+                if !mc.read().display_object.removed()  || mc.read().display_object.pending_removal()
```
Lord-McSweeney: Is this an unrelated formatting change (accidental)?

Lord-McSweeney:
Could the fix for #1029 (already implemented in #5492) also be part of this PR? It fixes at least one issue and is pretty simple to implement.

n0samu:
Super Mario 63 is a whole other can of worms that we probably shouldn't touch yet 😆

Lord-McSweeney:
Toad06 commented that some issues may occur because of #3839, and it's one of the more complex AVM1 games, so it would be an interesting test case.

n0samu:
@CUB3D Thanks so much for all your work on this! I just retested all of the games I tested previously, and found that everything is working great except for Wuzi Chess (which I know you're working on) and Bubble Bobble (#6185, which you might have missed).

n0samu:
I'm seeing a regression in Golphysics (#5219). Now when I click the Start button to start the game, it goes to a blank screen instead.
`WARN run_frame: ruffle_core::avm1::globals::movie_clip: MovieClip.swapDepths: Invalid target -- in [Actions Parent] / [Frame]` 

Also, I just remembered to test #2732 and that issue is still occurring the same as before. Not sure if that issue is really related though.

Lord-McSweeney:
This also fixes #9504.

ousia:
> Could the fix for https://github.com/ruffle-rs/ruffle/issues/1029 (already implemented in https://github.com/ruffle-rs/ruffle/pull/5492) also be implemented in this PR? It fixes at least one issue and is relatively simple to implement.

If #1029 could be fixed, in https://github.com/ruffle-rs/ruffle/issues/1561#issuecomment-947371500 @Herschel stated that #1561 is a variant of it (although it needs work to consider graphics as well).

Just in case it could be considered in this or in a future PR.

danielhjacobs:
The fixes keyword is picky, but can be used to auto-close issues when used properly. You should probably copy paste the below to auto-close all the issues, as the current phrasing will only close the first listed issue automatically:

Fixes https://github.com/ruffle-rs/ruffle/issues/993, fixes https://github.com/ruffle-rs/ruffle/issues/2840, fixes https://github.com/ruffle-rs/ruffle/issues/1513, fixes https://github.com/ruffle-rs/ruffle/issues/1140, fixes https://github.com/ruffle-rs/ruffle/issues/5275, fixes https://github.com/ruffle-rs/ruffle/issues/4793, fixes https://github.com/ruffle-rs/ruffle/issues/4404, fixes https://github.com/ruffle-rs/ruffle/issues/9257, fixes https://github.com/ruffle-rs/ruffle/issues/9135, fixes https://github.com/ruffle-rs/ruffle/issues/5923, fixes https://github.com/ruffle-rs/ruffle/issues/2575, fixes https://github.com/ruffle-rs/ruffle/issues/8765, fixes https://github.com/ruffle-rs/ruffle/issues/7711, fixes https://github.com/ruffle-rs/ruffle/issues/5711, fixes https://github.com/ruffle-rs/ruffle/issues/5095, fixes https://github.com/ruffle-rs/ruffle/issues/7169, fixes https://github.com/ruffle-rs/ruffle/issues/7345, fixes https://github.com/ruffle-rs/ruffle/issues/7362, fixes https://github.com/ruffle-rs/ruffle/issues/6960, fixes https://github.com/ruffle-rs/ruffle/issues/6203, fixes https://github.com/ruffle-rs/ruffle/issues/8872, fixes https://github.com/ruffle-rs/ruffle/issues/1017, fixes https://github.com/ruffle-rs/ruffle/issues/3839, fixes https://github.com/ruffle-rs/ruffle/issues/3604, fixes https://github.com/ruffle-rs/ruffle/issues/6185, fixes https://github.com/ruffle-rs/ruffle/issues/3916, fixes https://github.com/ruffle-rs/ruffle/issues/6336, fixes https://github.com/ruffle-rs/ruffle/issues/4118, fixes https://github.com/ruffle-rs/ruffle/issues/9504

n0samu:
With the latest commits, #2732 is actually fixed! 🎉 

Also, I can confirm that Bubble Bobble is fixed and the Golphysics regression is resolved. 👍

Lord-McSweeney:
Herschel commented on #4088 that #1029 is more complicated; sometimes, the behavior is present in SWFv6 and above. Currently, this PR does not fix #6581, #5579, or #2181, but that might be better kept for another PR.

Toad06:
Also fixes #5277 and #8576.

n0samu:
I can confirm that Wuzi Chess is fixed! 😄 
image

And #9514 is fixed as well!

core/src/avm1/activation.rs
```diff
@@ -544,18 +546,30 @@ impl<'a, 'gc> Activation<'a, 'gc> {
         }
     }
 
+    fn stack_push(&mut self, mut value: Value<'gc>) {
+        if let Value::Object(Object::StageObject(s)) = value {
```
Aaron1011, line 550: Do we need the ability to push to the AVM1 stack (without* this logic running? If not, could we move this logic into the `push` method, and have it take an `Activation`?

itamar567:
This PR seems to break [DragonFable](https://play.dragonfable.com/game/web.asp) ([SWF](https://play.dragonfable.com/game/DFLoader.swf)), the character is spawning in the top left corner of the screen (and cannot move at all due to hitboxes)

Before this PR:
image

After:
image

n0samu:
This PR also fixes the new issue #9556!

n0samu:
I can confirm the issue with DragonFable, but I think this would be really difficult to debug 😕

itamar567:
Finally found and isolated the issue with DragonFable:
it seems like when setting a variable to an existing movieclip (say, `var _loc1_ = testclip`), then calling `eval` on it to get a variable from the clip (`eval("_loc1_." + _loc1_.varname)`, when `_loc1_.varname` is a name of a variable in `testclip`) will result in `undefined`.

Here is a simple example of the issue: 
[attachment.zip](https://github.com/ruffle-rs/ruffle/files/10731580/attachment.zip)

core/src/avm1/activation.rs
CUB3D: We don't really need to be able to push without this running, so moving it would be possible, but because we need a mutable borrow of activation inside of `push` to create a `MovieClipReference` as well as a mutable borrow of `self.context.avm1` to call push it would likely be a much more invasive change

danielhjacobs:
That attachment itamar567 provided may be good to add as a test.

Lord-McSweeney:
Was the stubbing of `GraphicsBitmapFill` and `GraphicsTrianglePath` accidental?

n0samu:
This PR also fixes #8857!

core/src/avm1/activation.rs
CUB3D: Just adding that the only solution I can think of, would be either removing `avm1.push` and switching everything to use `activation.push` or using interior mutability so that `avm1.push` can take `&self` over `&mut self`. Did you have another solution in mind?

adrian17: > I'd like to make it harder to accidentally skip running this logic (by using avm1.push instead of stack_push).

On the contary, I'd consider minimizing complexity of push whenever it's not needed (say `action_push`). In avm2, stack push keeps popping up on profiles with non-trivial %, while on paper it should be one of the cheapest things ever.

Lord-McSweeney:
This PR does not currently fix #6370, although that seems to be out-of-scope.

CUB3D:
> Just to confirm, once this PR is merged, issues #1561, #2181, #4088, #5579, #6370 #6581 and #8213 could be part of a follow-up?
> 
> Many thanks for your help.

Yes, I just think that this PR is large enough and difficult enough to review as it is and I don't really want to delay it getting merged by trying to fix every tangentially related issue.

n0samu:
This PR also completely fixes #8314! 🎉

n0samu:
And it also completely fixes #8463! 🎉 🎉

core/src/avm1/activation.rs
```diff
@@ -541,7 +541,14 @@ impl<'a, 'gc> Activation<'a, 'gc> {
 
     fn stack_push(&mut self, mut value: Value<'gc>) {
         if let Value::Object(Object::StageObject(s)) = value {
-            // TODO: this is wrong, because we will re-create cached references, once a clip switches to the slower path resolution, it shouldn't swich back
+            // Note that there currently exists a subtle issue with this logic:
+            // If the cached `Object` in a `MovieClipReference` becomes invalidated, causing it to switch back to path-based object resolution,
+            // it should *never* switch back to cache-based resolution
+            // However, currently if a `MovieClipReference` in this invalidated-cache state is converted back to an `Object`, such as when passed as an argument to a function,
+            // if it pushed back onto the stack then it will be converted into a new `MovieClipReference`, causing it to switch back to cache-based resolution
+            // Fixing this will require a thorough refactor of AVM1 to store `Either<MovieClipReference, Object>
```
adrian17, line 549: I think this isn't the only way, right? We could do it without changing storage, but refactoring the places where the to-from-object conversions are done.
Anyway, let's just make this comment more general I guess ;D

adrian17 (approved):
I think we should just get it in and look for regressions.

p0008874:
Fixes #7728 #7846

core/src/avm1/object_reference.rs
```diff
@@ -0,0 +1,216 @@
...
+    /// First tuple param indificates if this path came from the cache or not
+    pub fn resolve_reference(
+        &self,
+        activation: &mut Activation<'_, 'gc>,
+    ) -> Option<(bool, Object<'gc>, DisplayObject<'gc>)> {
+        let cache = self.cached_stage_object.read();
+        // Check if we have a cache we can use
+        if let Some(cache) = cache.as_ref() {
+            // Check if we can re-use the cached `DisplayObject`, if we can then take this fast path
+            if let Some(mc) = cache.upgrade(activation.context.gc_context) {
```
Aaron1011, line 130: Is it ever possible for an AVM1 movieclip to become unreachable (and get garbage collected) without first being marked as `removed()`? If this is possible, then the behavior will depend on exactly when our garbage collector runs. If it's not possible, can you add a comment?

Aaron1011 (approved):

n0samu:
This PR also fixes the new issue #9800!

Dinnerbone:
Thank you so much @CUB3D!

nachtklavier:
@CUB3D Thank you very much. Canyon Defense works like a charm now. Highly appreciate your work!

TheGhostboi64:
This now fixes the problem the swf i used! Good job and I wish y'all the best of luck on getting ruffle completed!

Herschel:
Thanks, @CUB3D !

n0samu:
Please add that as a comment to the original issue, thanks!

PR #9449, opened 2023-02-07
Stub Video.videoWidth and videoHeight and port Video to ActionScript

Lord-McSweeney:
Progresses #8006. 320 and 240 are the default width and height.

### DIFF
--- a/desktop/src/main.rs
+++ b/desktop/src/main.rs
@@ -151,16 +151,16 @@ fn trace_path(_opt: &Opt) -> Option<&Path> {
 }
 
 fn parse_url(path: &Path) -> Result<Url, Error> {
-    Ok(if path.exists() {
+    if path.exists() {
         let absolute_path = path.canonicalize().unwrap_or_else(|_| path.to_owned());
         Url::from_file_path(absolute_path)
-            .map_err(|_| anyhow!("Path must be absolute and cannot be a URL"))?
+            .map_err(|_| anyhow!("Path must be absolute and cannot be a URL"))
     } else {
         Url::parse(path.to_str().unwrap_or_default())
             .ok()
-            .filter(|url| url.host().is_some())
-            .ok_or_else(|| anyhow!("Input path is not a file and could not be parsed as a URL."))?
-    })
+            .filter(|url| url.host().is_some() || url.scheme() == "file")
+            .ok_or_else(|| anyhow!("Input path is not a file and could not be parsed as a URL."))
+    }
 }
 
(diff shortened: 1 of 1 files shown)


danielhjacobs:
I suspect moving this file to AS at the same time may be better.

Lord-McSweeney:
I agree. Marking as draft until #9385 is merged.

core/src/avm2/globals/flash/media/Video.as
```diff
@@ -0,0 +1,19 @@
+package flash.media {
+    public final class Video {
```
adrian17, line 2: it's missing a DisplayObject inheritance; also it's not `final`

core/src/avm2/globals/flash/media/Video.as
```diff
@@ -0,0 +1,19 @@
...
+        private var _videoWidth: int;
+        private var _videoHeight: int;
+
+        public function Video(videoWidth:int = 320, videoHeight:int = 240) {
```
adrian17, line 6: let's add a stub mention to the constructor

adrian17 (commented):
(Note to whoever merges: let's not forget to auto-squash)

core/src/avm2/globals/flash/media/Video.as
danielhjacobs: I think Adrian meant like https://github.com/ruffle-rs/ruffle/blob/master/core/src/avm2/globals/flash/utils/Dictionary.as#L10

Lord-McSweeney: `Video` doesn't actually do anything itself; it only serves as a container to be filled by `Video.attachCamera` or `Video.attachNetStream`. Adding a stub warning will unnecessarily warn about the Video stubbed on content that doesn't use `attachNetStream`, in which case the `Video` container is functioning correctly.

core/src/avm2/globals/flash/media/Video.as
relrelb: Agree on `DisplayObject ` inheritance, but actually `Video` is not declared as `final` as per https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/media/Video.html.

danielhjacobs:
Did you mean to close this and force-push and remove all commits?

core/src/avm2/globals/flash/media/Video.as
adrian17: >  actually Video is not declared as final

yeah, that's what I meant :) That comment was from the version of this PR that had `final`.

Issue #8947, opened 2022-12-28
Pursuit of Hat: Player character does not appear in level 6 after beating level 5


n0samu:
### Describe the bug

Link to SWF file: http://i.notdoppler.com/files/pursuitofhat.swf

The previous issue #8003 with [Pursuit of Hat](http://www.notdoppler.com/pursuitofhat.php) is now fixed. But after beating level 5 and clicking "next level", level 6 loads without the player character.
![image](https://user-images.githubusercontent.com/71368227/209775852-c032511f-ee4d-4c57-a2c7-90ff94b9b5fd.png)

The level loads normally if you click the "reset" button or load it from the level select screen instead.

The problem also happens after beating levels 6 and 7, but not after beating level 8 or levels 1-4.

There are some other minor issues with the game (also mentioned in #8003):
- You need to click inside each level to activate the keyboard controls.
- Restarting a level does not play the associated sound effect.
- The BigDino intro animation is skipped, though its audio plays.

### Expected behavior

Level 6 should load with the player character:
![image](https://user-images.githubusercontent.com/71368227/209775901-0e679df8-a075-40aa-8e35-3104f7bc55c7.png)

### Affected platform

Desktop app

### Operating system

Windows 10

n0samu:
Besides the BigDino intro being skipped, all of these problems are fixed now!

MartySVK:
looks like game is broken, I see spamming error:

``Error #1009: Cannot access a property or method of a null object reference. (accessing field: parent)
``

n0samu:
The BigDino intro skip has been fixed by #13389!

PR #9450, opened 2023-02-07
core: Run frames for 'orphan' AVM2 MovieClips

Aaron1011:
When a MovieClip is an 'orphan' (it has no parent), it still has frames run (including frame scripts). Some SWFS like SteamBirds and 'This is the Only Level TOO' rely on this behavior, so we need to implement it.

The overall idea is straightforward - we keep a global list of orphan movies, which we add to whenever we unset the parent for a movie. This list stores weak references for consistency with Flash. When we run a frame, we process entries in the root movie list, in addition to the normal recursive processing from the `Stage`.

However, exactly matching Flash's output turned out to be quite tricky. The particular sequence of calls I make in `run_all_phases_avm2` makes Ruffle pass two complicated test cases, but there could still be lurking bugs.

This is enough to get SteamBirds to the first level (which doesn't render due to a different error).


n0samu:
This seems to cause a regression in Epos (#9649) and Pursuit of Hat (#8947) where the intro sounds loop over and over after the intro segments are over. But both games have existing issues, so maybe the problem needs to be fixed elsewhere, not here?

core/src/display_object/avm2_button.rs
```diff
@@ -208,6 +208,9 @@ impl<'gc> Avm2Button<'gc> {
                     Ok(child) => {
                         child.set_matrix(context.gc_context, record.matrix.into());
                         child.set_depth(context.gc_context, record.depth.into());
+                        if let Some(clip) = child.as_movie_clip() {
+                            clip.set_is_button_state(context.gc_context);
```
kmeisthax, line 212: As currently written this flag doesn't strictly mean "this is a button state". SimpleButton groups multiple children for the same state under a parent clip that it instantiates. So that holder clip *doesn't* get the button-stage flag, but its children do, even though they're not "button states" per se. Is this correct?

core/src/display_object/movie_clip.rs
```diff
@@ -2189,6 +2211,26 @@ impl<'gc> MovieClip<'gc> {
         RefMut::map(self.0.write(gc_context), |s| &mut s.drawing)
     }
 
+    /// If `true`, this is clip is being used as one of the four
+    /// `ButtonState`s for an `Avm2Button`. Currently, this flag
+    /// is used to opt out of orphan handling in `avm2::valid_orphan`
```
kmeisthax, line 2216: If the above-mentioned flag behavior is correct this documentation should be changed from "is being used as a button state" to "was instantiated by `Avm2Button` from an SWF tag".

kmeisthax (commented):
I took a quick look at this. I'm not 100% sure we have the correct implementation, but it matches the tests at least. I did notice that some of the weirdness your implementation captures isn't entirely documented and have noted it here.

core/src/display_object/avm2_button.rs
Aaron1011: Honestly, I'm not really sure how simplebutton interacts with orphan movies - this was needed in order to keep the existing test spassing.

Issue #6673, opened 2022-04-10
miscalculated `Sound.duration()` (non-standard sample rate)


ousia:
The attached image shows how Flash Player and Ruffle display the `Sound.duration()` of the audio contained in <https://www.free-culture.tk/lfc-presentation.swf>:

![dm-rfflvfp](https://user-images.githubusercontent.com/203767/162624116-11542a71-e425-4fd2-9f61-58902c34e3dc.png)

As @Herschel has explained many times, the problem here is that the sound file has a non-standard sample rate 16kHz (although it is tagged in the SWF animation as being 11kHz).

For both `Sound.position()` and `Sound.duration()`, sample rate should be read from theMP3 file itself (not from SWF tagging).

BTW, I think there is no need to attach a sample file (it’s over 5MB). Let me know, if you need it.

Many thanks for your excellent work.

ousia:
Even after #4273, miscalculated `Sound.duration()` is still there.

n0samu:
As a workaround, why not modify the SWF file so the sound sample rate is tagged correctly?

ousia:
> As a workaround, why not modify the SWF file so the sound sample rate is tagged correctly?

Many thanks for your reply, @n0samu.

Well, in that case I’d have to modify the MP3 file itself. According to the [SWF File Format Specification](https://open-flash.github.io/mirrors/swf-spec-19.pdf#search=SoundRate):

> The sampling rate. This is ignored for Nellymoser and Speex codecs. 5.5kHz is not allowed for MP3. 0 = 5.5 kHz; 1 = 11 kHz; 2 = 22 kHz; 3 = 44 kHz.

In theory, 16kHz would be an invalid value. But in practice, the official player from Adobe had no problem with it.

If `Sound.position()` works right with this value for sampling rate, I think it would be consistent to make it work right with `Sound.duration()`.

Many thanks for your help.

n0samu:
Ah that makes sense, I agree with what you're saying then

ousia:
@n0samu,

I think this may be an easy fix, but I’m afraid I cannot code (in almost any programing language known to humankind 😅).

So, I will have to wait for the fix.

ousia:
BTW, this prevents having a timer with total time in sound.

ousia:
This is what I meant in my previous comment:

[...]

PR #9457, opened 2023-02-07
core: Queue up Sound and SoundChannel methods during loading

Aaron1011:
Flash supports calling `Sound.play`, `SoundChannel.stop`, and `SoundChannel.soundTransform` while a sound load is in progress (e.g. immediately after calling `Sound.load`).

To support this, we queue up information inside `SoundObject` and `SoundChannelObject` when a load is in progress. When a load completes, we trigger any queued `Sound.play` and `SoundChannel.stop` calls, and apply the most recent `SoundChannel.soundTransform`

### DIFF
--- a/core/src/avm2/globals/flash/media/sound.rs
+++ b/core/src/avm2/globals/flash/media/sound.rs
@@ -3,7 +3,7 @@
 use crate::avm2::activation::Activation;
 use crate::avm2::class::{Class, ClassAttributes};
 use crate::avm2::method::{Method, NativeMethodImpl};
-use crate::avm2::object::{sound_allocator, Object, SoundChannelObject, TObject};
+use crate::avm2::object::{sound_allocator, Object, QueuedPlay, SoundChannelObject, TObject};
 use crate::avm2::value::Value;
 use crate::avm2::Error;
 use crate::avm2::Multiname;
@@ -12,7 +12,7 @@ use crate::avm2::QName;
 use crate::backend::navigator::Request;
 use crate::character::Character;
 use crate::display_object::SoundTransform;
-use crate::{avm2_stub_getter, avm2_stub_method};
+use crate::{avm2_stub_constructor, avm2_stub_getter, avm2_stub_method};
 use gc_arena::{GcCell, MutationContext};
 use swf::{SoundEvent, SoundInfo};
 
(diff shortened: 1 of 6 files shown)


core/src/avm2/globals/flash/media/sound.rs
```diff
@@ -12,20 +12,24 @@ use crate::avm2::QName;
...
     this: Option<Object<'gc>>,
-    _args: &[Value<'gc>],
+    args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
     if let Some(this) = this {
         activation.super_init(this, &[])?;
 
-        if this.as_sound().is_none() {
+        if !args.is_empty() {
+            avm2_stub_constructor!(activation, "Sound", "with arguments");
```
Dinnerbone, line 29:
```suggestion
            avm2_stub_constructor!(activation, "flash.media.Sound", "with arguments");
```

SN902:
Kingdom rush, dash_fly_by_futzi01-d9flryk.swf and most htpot NSWF stuff, now crashes with errors in this commit:

Cannot access property:: close of null or undefined
Cannot access property:: addeventListener of null or  undefined
Cannot access property:: smoothing of null or not undefined

Aaron1011:
@SN902 Is that a regression, or were those swfs previously crashing earlier?

SN902:
@Aaron1011 on latest master works without issues

Aaron1011:
@SN902 I've updated the PR to fix the issue - I left in a `panic!` by accident.

SN902:
@Aaron1011 issue solved

ousia:
@Aaron1011,

I wonder whether this PR may also be considered to fix #6673 (or it is a totally unrelated issue).

Many thanks for your help.

Aaron1011:
@ousia I think that issue is unrelated - with this PR, loading is still all-or-nothing (we never report streaming/partial loads to AVM2), so it won't affect `Sound.duration`.

ousia:
@Aaron1011, many thanks for your reply.

It would be great that `Sound.position()` and `Sound.duration()` are consistent with non-standard frame rates.

I hope this can be achieved in the future.

Many thanks for your contributions to Ruffle (and to all contributors for their work).

Dinnerbone (approved):

PR #9461, opened 2023-02-08
web: Add virtual keyboard that opens when clicking an editable EditText

danielhjacobs:
Due to the way the code was written the phone keyboard re-defaults to capitalized for every character. Please make suggestions for improvement. The keyup and keydown events fire fake key codes on Chrome on Android so cannot be used (https://javascript.info/keyboard-events#mobile-keyboards). The keypress event does not fire on Chrome on Android.


danielhjacobs:
Due to the way the code was written the phone keyboard re-defaults to capitalized for every character. Please make suggestions for improvement. The keyup and keydown events fire fake key codes on Chrome on Android so cannot be used. The keypress event does not fire on Chrome on Android.

danielhjacobs:
Just thought of an idea for keyboard capitalization. Will implement it later.

n0samu:
For the capitalization issue, can't you just use the `autocapitalize="none"` attribute on the input element? https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize

danielhjacobs:
Didn't know about it worth a shot. It doesn't seem to exist in all browsers. Still it's better than nothing. I'm busy today though.

danielhjacobs:
Added autocapitalize attribute. It works as expected. Now even if you specifically turn on full capitalization on your phone keyboard it goes back to lowercase after you type a letter, but that's certainly better behavior, even if not ideal.

I personally would consider this ready for review in its current state. It works and it's a feature people have asked for forever.

If there's a better place to put the option to open the virtual keyboard on mobile, I'm open to suggestions, as the ability to find this option in the context menu isn't great and apparently the context menu doesn't work on iOS. But to be honest, that's a bit of a more fundamental Ruffle issue.

danielhjacobs:
https://user-images.githubusercontent.com/29206584/218123643-21be992a-ac8e-4553-8c09-4d1723eb2855.mp4

danielhjacobs:
https://user-images.githubusercontent.com/29206584/218125378-8a1136d2-e383-4112-92ce-2fd4bfc8ee95.mp4

danielhjacobs:
Unfortunately I have to hide the virtual keyboard when clicking anywhere, even if it hasn't been dismissed, because otherwise it seems to reappear no matter where you click on the content even after it has been dismissed.

web/packages/core/src/ruffle-player.ts
```diff
@@ -809,6 +809,49 @@ export class RufflePlayer extends HTMLElement {
             console.error("SWF download failed");
         }
     }
+    private openVirtualKeyboard(): void {
+        const initialValue = "Initial";
+        const initialValueLength = initialValue.length;
+        const existingInput = this.shadow.getElementById(
+            "ruffle-virtual-keyboard"
+        );
+        if (!existingInput) {
```
n42k, line 818: I think we can do an early return here:
```
const existingInput = this.shadow.getElementById(
    "ruffle-virtual-keyboard"
);

if (existingInput) {
  existingInput.focus();
  return;
}

const initialValue = "Initial";
const initialValueLength = initialValue.length;
const input = // ...
```

web/packages/core/src/ruffle-player.ts
```diff
@@ -809,6 +809,49 @@ export class RufflePlayer extends HTMLElement {
...
+            const input = document.createElement("input");
+            input.setAttribute("type", "text");
+            input.setAttribute("autocapitalize", "off");
+            input.style.position = "absolute";
+            input.style.top = "-100px";
+            input.style.width = "1px";
+            input.style.height = "1px";
+            input.id = "ruffle-virtual-keyboard";
```
n42k, line 826: Probably should make a variable to hold this magic string `"ruffle-virtual-keyboard"`.

web/packages/core/src/ruffle-player.ts
```diff
@@ -809,6 +809,49 @@ export class RufflePlayer extends HTMLElement {
...
+    private openVirtualKeyboard(): void {
+        const initialValue = "Initial";
+        const initialValueLength = initialValue.length;
+        const existingInput = this.shadow.getElementById(
+            "ruffle-virtual-keyboard"
+        );
+        if (!existingInput) {
+            const input = document.createElement("input");
+            input.setAttribute("type", "text");
+            input.setAttribute("autocapitalize", "off");
```
n42k, line 821: On GBoard, it helps to have autocomplete and autocorrect off:
```
["autocapitalize", "autocomplete", "autocorrect"]
  .forEach(key => input.setAttribute(key, "off"));
```

web/packages/core/src/ruffle-player.ts
```diff
@@ -809,6 +809,49 @@ export class RufflePlayer extends HTMLElement {
...
+            input.style.top = "-100px";
+            input.style.width = "1px";
+            input.style.height = "1px";
+            input.id = "ruffle-virtual-keyboard";
+            input.value = initialValue;
+            input.addEventListener("input", () => {
+                const isBackspace = input.value.length <= initialValueLength;
+                const string = isBackspace
+                    ? "-"
+                    : input.value.slice(initialValueLength);
```
n42k, line 832: Feels a bit hacky to use "-" for the backspace.

How about:
```
const isBackspace = input.value.length <= initialValueLength;
const chars = isBackspace ? ["Backspace"] : [...input.value.slice(initialValueLength];
for (const char of chars) {
   for (const eventType of ["keydown", "keyup"]) {
    // ...
```

web/packages/core/src/ruffle-player.ts
```diff
@@ -809,6 +809,49 @@ export class RufflePlayer extends HTMLElement {
...
+
+                for (const char of string) {
+                    const key = isBackspace ? "Backspace" : char;
+                    for (const eventType of ["keydown", "keyup"]) {
+                        const event = new KeyboardEvent(eventType, {
+                            key: key,
+                            bubbles: true,
+                        });
+                        this.dispatchEvent(event);
```
n42k, line 841: Nitpick:
```
this.dispatchEvent(new KeyboardEvent(eventType, {
  key: key,
  bubbles: true,
}));
```

n42k:
I consider this ready for review as well, looks quite good!
Suggested a few stylistic changes, but nothing major.

Thanks!

web/packages/core/src/ruffle-player.ts
danielhjacobs: Fair enough, I wasn't using "-" until I switched to loop over the characters so multi-character strings worked correctly, and after that without a change a backspace would have type "B" -> "a" -> "c".... This is a good idea instead, just not one I thought of myself.

web/packages/core/src/ruffle-player.ts
danielhjacobs: Yeah, I like early returns, just always forget them after refactors. I need to remember when I add an `else` to always consider if I can use one.

n0samu:
I just tested this along with #9575 on iOS, and I have a few small suggestions to improve the experience there:
- To hide the input box, you need to set `input.style.opacity = "0";`. Right now it's shown as a tiny box on the top-left.
- To prevent the browser from automatically scrolling to and zooming into the input box, you can call `input.focus({preventScroll: true});`

web/packages/core/src/ruffle-player.ts
```diff
@@ -808,6 +808,53 @@ export class RufflePlayer extends HTMLElement {
...
+        const existingInput = this.shadow.getElementById(virtualKeyboardId);
+
+        if (existingInput) {
+            existingInput.focus({ preventScroll: true });
+            return;
+        }
+
+        const initialValue = "Initial";
+        const initialValueLength = initialValue.length;
+        const input = document.createElement("input");
```
relrelb, line 822: Maybe define an `<input>` element statically in `shadow-template.ts`? Now it seems like the dynamically-created elements are "leaked" into the DOM (i.e. added but never removed).

relrelb:
Ideally we could open a virtual keyboard whenever an `EditText` is clicked, without making the user open it explicitly via the context menu. That probably needs to involve some Rust code to propagate clicks into the web side.

n0samu:
> Ideally we could open a virtual keyboard whenever an `EditText` is clicked, without making the user open it explicitly via the context menu. That probably needs to involve some Rust code to propagate clicks into the web side.

Do you want to work on that then? 😄 

I looked into implementing it already, but it is currently far beyond my abilities.

web/packages/core/src/ruffle-player.ts
danielhjacobs: Sorry, I've been busy with the iOS PR. I agree with your idea here, and it would probably be simpler to do that. Technically the input will exist on desktop too and just never be used, whereas here it only exists on touch devices since it only appears after choosing the context menu option, but that's probably fine. I'll try to implement this.

danielhjacobs:
> Ideally we could open a virtual keyboard whenever an `EditText` is clicked, without making the user open it explicitly via the context menu.

I wholeheartedly agree, but without trying to speak for others, just based on what they've told me, I think myself, @n42k, and @n0samu lack the requisite knowledge to implement that ourselves. This is the compromise until someone implements the proper logic.

web/packages/core/src/ruffle-player.ts
danielhjacobs: I imagine something like:
```css
#ruffle-virtual-keyboard {
  position: absolute;
  opacity: 0;
  top: -100px;
  width: 1px;
  height: 1px;
}
```
```html
<input id="ruffle-virtual-keyboard" type="text" autocapitalize="off" autocomplete="off" autocorrect="off" value="Initial" oninput="virtualKeyboardInput()">
```

web/packages/core/src/shadow-template.ts
```diff
@@ -275,12 +275,20 @@ ruffleShadowTemplate.innerHTML = `
                 transform: rotate(360deg);
             }
         }
+        #ruffle_virtual_keyboard {
```
relrelb, line 278: Personally I don't see a reason to call everything "Ruffle" - It doesn't add any meaning and only lengthens names unnecessarily. So how about staying just with `#virtual-keyboard`?

By the way, hopefully this will be extracted to a standalone CSS file soon (once `ruffle-core` will be bundled using Webpack, I'm working on this...). Then Stylelint will enforce following `kebab-case` pattern,

web/packages/core/src/ruffle-player.ts
```diff
@@ -821,6 +828,39 @@ export class RufflePlayer extends HTMLElement {
...
+                        key: char,
+                        bubbles: true,
+                    })
+                );
+            }
+        }
+        input.value = initialValue;
+    }
+    private openVirtualKeyboard(): void {
+        if (!("listener" in this.virtualKeyboard.dataset)) {
```
relrelb, line 853: My feeling is that adding an event listener unconditionally isn't much costly, but it would simplify things a lot.

web/packages/core/src/shadow-template.ts
danielhjacobs: Ok. This file currently uses kebab-case, camelCase, and snake_case, so i was unsure what convension to follow. Admittedly I think I added camelCase when adding the preloader, but the other two were there for a while. The ruffle part of the id was a remnant of before I added this to the shadow template, even with Javascript, as I added it to the document in the first version of this PR and used that I'd assuming it would not conflict with preexisting elements. I'll remove it.

web/packages/core/src/ruffle-player.ts
danielhjacobs: Would it not trigger all of them though? So opening the virtual keyboard with the context menu twice types each character twice?

danielhjacobs: And simplifying things "a lot" seems like a bit of an exaggeration, as wouldn't it just remove the conditional that checks for the data-* element and the line that sets the data-* element?

danielhjacobs: Just tested an unconditional statement here. It's fairly unpredictable what exactly it does because of race conditions. But usually if you open the virtual keyboard multiple times, it seem like typing will stop doing anything or type one or more backspaces. The input event needs to only fire once.

danielhjacobs: Or wait, you mean add the unconditional event listener to https://github.com/ruffle-rs/ruffle/pull/9461/files#diff-467a580c71d4e3b96032b54a2a292d485f3b83f87ba2478daf693bac6d532965R488

relrelb: Oh yeah, sorry if I wasn't clear. I meant adding the event listener during the one-time initialization.

web/packages/core/src/ruffle-player.ts
```diff
@@ -478,6 +482,13 @@ export class RufflePlayer extends HTMLElement {
         }
 
         this.unmuteAudioContext();
+        this.container.addEventListener("click", () =>
+            this.virtualKeyboard.blur()
+        );
+        this.virtualKeyboard.addEventListener(
+            "input",
```
relrelb, line 489: Isn't `"keydown"` more appropriate? Because then the pressed key is available directly from the event.

danielhjacobs: There was a long discussion about this in the #frontend_apps channel on Discord. The issue is https://javascript.info/keyboard-events#mobile-keyboards.

danielhjacobs: Tried keyup, keydown, and keypress. keyup and keydown have the same results: Nothing for e.code. "229" for e.keyCode and e.which. "Unidentified" for e.key. keypress does not fire.

danielhjacobs: Also, this logic works when opening your Clipboard with the virtual keyboard and choosing a previously copied line of text. keydown would not work for that.

danielhjacobs: image

relrelb (approved):

web/packages/core/src/ruffle-player.ts
relrelb: Got it.

relrelb (approved):

n42k:
Works flawlessly on Android, on all 3 keyboards I tested.

Swfs tested:
* https://aq.battleon.com/game/web?launchtype=medium
* https://github.com/ruffle-rs/ruffle/blob/master/tests/tests/swfs/avm1/edittext_default_format/test.swf

Keyboards:
* Samsung Keyboard 5.6.00.29
* GBoard 12.6.02
* Google Voice Typing **(?)** - no known version

Specifications: Galaxy M52 5G, Android 13, One UI 5.0.

n42k (approved):

n0samu (approved):
Just tested this on my iPhone and it's working perfectly, thank you!

relrelb:
Going to merge this. Thank you!

PR #9478, opened 2023-02-09
avm2: Make sure scripts get initialized after all scripts have been loaded when lazy init is disabled

Bale001:
Before this, SWF's that had lazy initialization disabled would initialize their scripts as they were being loaded. This is problematic because the script being initialized might be relying on a script that hasn't been loaded yet. This PR changes the behavior by loading all scripts first and then initializing them afterward (as long as lazy init is disabled). SWF's created by crossbridge rely on this behavior.


Aaron1011:
Could you write a test that demonstrates the new behavior?

Bale001:
> Could you write a test that demonstrates the new behavior?

Done

Aaron1011 (approved):

PR #9489, opened 2023-02-09
core: Support using URLVariables as a POST request body

Aaron1011:
This is needed by the Newgrounds API. We don't have the ability
to make fake requests to HTTP urls in our test frameworks,
so I haven't added any tests for this. However, I tested locally
that this allows the Newgrounds API to work (and got a medal
in Cloud Wars).


core/src/avm2/globals/flash/net/URLVariables.as
```diff
@@ -6,28 +6,28 @@ package flash.net {
...
 
         public function toString(): String {
-	    var acc : String = ""
-	    var sep :String = ""
-	    for (p in this) {
-		var pe : String = escapeMultiByte(p);
-		var val = this[p];
-		if (val is Array) {
-		    for (i in val) {
+			var acc : String = ""
```
adrian17, line 9: accidental whitespace, spaces<->tabs mismatch?

core/src/avm2/globals/flash/net/url_loader.rs
```diff
@@ -59,8 +60,55 @@ fn spawn_fetch<'gc>(
...
+
+    let data = url_request.get_property(&Multiname::public("data"), activation)?;
+
+    let data = if let Value::Null = data {
+        None
+    } else {
+        Some(data.coerce_to_object(activation)?)
+    };
+
     // FIXME - set options from the `URLRequest`
```
adrian17, line 75: Could you list the missing options here?

adrian17 (approved):

Aaron1011:
@adrian17 I've addressed your comments

Issue #8801, opened 2022-12-15
Canabalt: Cannot start game


xerz-one:
### Describe the bug

When pressing X or C to start a new game, nothing happens. This was tested in both the desktop client and the Firefox extension.

### Expected behavior

A new game starts, loading the corresponding world

### Affected platform

Found on desktop app and web extension

### Operating system

Arch Linux

### Browser

Firefox 108.0

### Additional information

Log example for the latest Linux desktop nightly:

```
[2022-12-15T22:10:46Z ERROR ruffle_core::avm2::events] Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x55f2329bae00 }) to handler FunctionObject(FunctionObject(GcCell(Gc { ptr: 0x55f22ffc1a58 }))) : RustError("VerifyError: Trait _p in class BG has same name as trait _p in class FlxSprite, but does not override it")
[2022-12-15T22:10:48Z ERROR ruffle_core::avm2::events] Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x55f232444bc0 }) to handler FunctionObject(FunctionObject(GcCell(Gc { ptr: 0x55f22ffc1a58 }))) : RustError("Resolved parameter type com.adamatomic.Canabalt::BG is unresolvable, not a class, null, or undefined")
```

Lord-McSweeney:
Fixed by #9506.

xerz-one:
Game still doesn't work with error `panicked at 'called `Option::unwrap()` on a `None` value', core/src/bitmap/bitmap_data.rs:1231:22`, should I open a new ticket?

waspennator:
Can confirm, this pops up in the error logs now, have a feeling one of the array.removeat PR's regressed it.

Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x1333c10 }) to handler FunctionObject(FunctionObject { ptr: 0x125df90 }) : RustError("VerifyError: Trait removeAt in class FlxArray has same name as trait removeAt in class Array, but does not override it")

waspennator:
Update, should be fixed by #10168

Issue #9126, opened 2023-01-12
Cell Shock: Doesn't start right


waspennator:
### Describe the bug

[cell-shock-1537169018.zip](https://github.com/ruffle-rs/ruffle/files/10397632/cell-shock-1537169018.zip)
(Uploaded to test on desktop since the website I tested it on was rendering it at a different resolution, causing it to not be sized properly.)
Game seems to load up to the title screen and works up until you actually start the game, You can move around and you auto fire, but the bullets just seem to stay in place after they get shot, no brussel sprouts show up, the meters, nothing else aside from you and the bullets.
![Core2](https://user-images.githubusercontent.com/78768499/211961899-0b466b8f-10dc-4486-a1b5-008eb0e4e791.png)

Checking on a website hosting this game, i'm noticing these errors in the logs.

Error dispatching event EventObject(EventObject { type: "mouseUp", class: flash.events::MouseEvent, ptr: 0x1ffbc90 }) to handler FunctionObject(FunctionObject { ptr: 0x1ff0e28 }) : RustError("VerifyError: Trait _looping in class Explosion has same name as trait _looping in class SpriteSheet, but does not override it")

Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x2014c80 }) to handler FunctionObject(FunctionObject { ptr: 0x20135c0 }) : RustError("Cannot access property http ://adobe.com/AS3/2006/builtin::pop of null or undefined")

### Expected behavior

The meters on top should appear, brussel sprouts should spawn from the sides and such to swarm you and the bullets shouldn't be stuck in place.

![Core1](https://user-images.githubusercontent.com/78768499/211962110-2e6c5d0a-b032-4185-9ee4-d39b6525cd4f.png)

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 108

waspennator:
Tested on desktop and on the ruffle web demo, game seems to work properly now, played up to wave 10 with no noticable issues now. closing since it was fixed by #9506

PR #9506, opened 2023-02-10
Avm2: Correctly support private namespaces

adrian17:
Problem description tldr: in FP, the following:
```
class B { private var x; } // QName("x", PrivateNamespace(""))
class C extends B { private var x; } // QName("x", PrivateNamespace(""))
```
Works without conflicts, as even though the namespaces "look" the same, they are actually two different namespace objects. This previously failed in Ruffle, as we only compared the contents.
(note: I couldn't add a test for it, as mxmlc doesn't generate such identical-looking qnames, and jpexs doesn't support precisely manipulating pool contents)

Unfortunately this required a massive refactor to propagate Activation everywhere, as now 1. creating a Namespace requires MC, 2. we want to avoid creating Namespaces on each use so I put the most common ones in `activation.context.avm2`, 3. this propagates into nearly every user of `.get_property(Multiname::public(...))`

This should hopefully fix most/all of the errors that look like:
```
Trait _p in class BG has same name as trait _p in class FlxSprite, but does not override it
```
But the only SWF I tested so far was https://www.newgrounds.com/portal/view/510303, which indeed becomes playable now.

Also removed the private_trait_scripts, as to the best of my knowledge, there's now nothing preventing us from just using the domain's definitions for everything. @Aaron1011 is this safe?

Todo:
- test more swfs?
- Do a perf comparison. (I checked that in basic content like box2d demo, we aren't creating any namespaces after init, so that's nice)

Also I'd really appreciate some feedback on the commit `avm2: Fix rust<->avm shared private fields`. I didn't really figure out a nicer way of doing this, but maybe someone else has something.
(and this commit needs to be squashed as otherwise the previous one doesn't pass tests)

Also to reviewers: in the first commit, 99% of changes in `avm2/globals` are identical, so you can skip through most of them. `avm2/object` are a bit more interesting, and top-level files are all interesting. That should still reduce the review volume by like 2-3x :)

### DIFF
--- a/core/src/avm2.rs
+++ b/core/src/avm2.rs
@@ -56,7 +56,7 @@ pub use crate::avm2::domain::Domain;
 pub use crate::avm2::error::Error;
 pub use crate::avm2::globals::flash::ui::context_menu::make_context_menu_state;
 pub use crate::avm2::multiname::Multiname;
-pub use crate::avm2::namespace::Namespace;
+pub use crate::avm2::namespace::{Namespace, NamespaceData};
 pub use crate::avm2::object::{
     ArrayObject, ClassObject, EventObject, Object, ScriptObject, SoundChannelObject, StageObject,
     TObject,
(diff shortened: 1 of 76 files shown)


Aaron1011:
> Also removed the private_trait_scripts, as to the best of my knowledge, there's now nothing preventing us from just using the domain's definitions for everything. @Aaron1011 is this safe?

This looks right to be - when I wrote `private_trait_scripts`, I assumed that skipping exports of private definitions was the right thing to do. Now that we export everything, `private_trait_scripts` should be unnecessary (and the `coerce_property` test is still passing.

Aaron1011 (approved):

n0samu:
I can confirm that this fixes #8801 and #9126! Additionally, [Georganism 2](https://www.notdoppler.com/georganism2.php) and [Axon](http://www.notdoppler.com/axon.php) now display their loading screens.

Aaron1011 (approved):

Aaron1011 (approved):

PR #2081, opened 2020-12-24
core: Implement TextField control characters

JMcKiern:
This pr implements some control character handling for text fields (eg CtrlC, CtrlV).

I'm submitting this as a draft to make sure that I'm going about this the right way.

## TODO

- [ ] - <kbd>Up</kbd>, <kbd>Down</kbd> - Should move caret between lines
- [ ] - <kbd>Home</kbd>, <kbd>End</kbd> - Should go to the start/end of the line
- [ ] - <kbd>Ctrl+Left</kbd>, <kbd>Ctrl+Right</kbd> - Should go to the start/end of the previous/next word

PR #9521, opened 2023-02-11
core: Implement TextField control characters

allancoding:
I think this should work.
This is the updated #2081


adrian17:
Uhh... can you please make a new PR? From what I see, this PR is reported to have 0 commits, and somehow GH got confused and decided that it's "merged" now.

adrian17:
In fact, from what I see, your branch `[allancoding:master]` has no commits.

allancoding:
Ok I fixed it here: #9557

PR #9536, opened 2023-02-12
avm2: Stub describeType to have .@name work

adrian17:
This should let NG API progress a bit further.
Note: I didn't yet test it on any real content using the API. Would like me or someone else to try this before merging, to see if this hack is worth it at all.


Dinnerbone:
Confirmed to progress Farm Rush (after the tag 72 PR)

Issue #8888, opened 2022-12-22
https://evilgames.eu/flashgames/frog-fable.htm not working


Lord-McSweeney:
### Describe the bug

On https://evilgames.eu/flashgames/frog-fable.htm, many AVM errors are reported in the console. They're all something like this: `Got AVM2 error AvmError(Object(ErrorObject(ErrorObject(GcCell(Gc { ptr: 0x20178cc }))))) when attempting to assign symbol class [something]`. Based on the quantity of errors, I would guess that this is unintentional. The game is stuck on a black screen. The only other warning shown in the console is `Unknown tag code: 72`

### Expected behavior

The thousands of warnings should not appear.

### Affected platform

Browser's extension

### Operating system

Ubuntu 20.04, Linux 5.13

### Browser

Chrome 108

Lord-McSweeney:
Looks like the next thing this needs is `flash.display.Shader`.

Lord-McSweeney:
The game is playable now, but is slow and has issues with the BigDino intro.

n0samu:
Original issue is fixed. I've opened #13110 to track the BigDino intro problem.

PR #9537, opened 2023-02-12
Support older DoAbc tag (72)

Dinnerbone:
No test, was too hard to construct 😅


Lord-McSweeney:
Progresses #8888, which now needs `TouchEvent` and `flash.utils.describeType`.

Aaron1011 (approved):

PR #9542, opened 2023-02-12
avm2: Implement `finddef` opcode

Bale001:
This implements a simple opcode that searches the domain for a multiname, and returns the script containing it.


core/src/avm2/activation.rs
```diff
@@ -1742,6 +1743,30 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
+        method: Gc<'gc, BytecodeMethod<'gc>>,
+        index: Index<AbcMultiname>,
+    ) -> Result<FrameControl<'gc>, Error<'gc>> {
+        let multiname = self.pool_multiname_static(method, index)?;
+        avm_debug!(self.avm2(), "Resolving {:?}", *multiname);
+        if let Some((_, mut found)) = self.domain().get_defining_script(&multiname)? {
+            let object = found.globals(&mut self.context)?;
+            self.push_stack(object);
+        } else {
+            return Err(Error::AvmError(crate::avm2::error::reference_error(
```
adrian17, line 1757: I think the impl should be next to Domain::get_defined_value, as most of the code there is nearly identical
(in fact, get_defined_value could use this new method under the hood, right? It's just find_def + get_property)

core/src/avm2/domain.rs
```diff
@@ -121,25 +121,36 @@ impl<'gc> Domain<'gc> {
...
     }
 
+    /// Resolve a Multiname and return the script that provided it.
+    ///
+    /// If a name does not exist or cannot be resolved, an error will be thrown.
+    pub fn find_defining_script(
+        self,
+        activation: &mut Activation<'_, 'gc>,
+        multiname: &Multiname<'gc>,
+    ) -> Result<(QName<'gc>, Script<'gc>), Error<'gc>> {
```
adrian17, line 131: just a question: do you think we still need this dance with `get_defining_script` returning a QName that we later use to `get_property`? As in, can't we just use the input multiname directly?
There might be some good motivation as to why we did it this way originally, but it could also be just "get_property used to take a QName arg, so we needed to grab the correct namespace".

adrian17: (this is a note for possible future simplification, don't need to simplify this in this PR :) )

adrian17 (approved):

PR #9547, opened 2023-02-13
avm2: Workaround haxe calling addChild(this) before super()

Dinnerbone:
This will progress haxe games.

Absolutely not an ideal fix and very much a hack, but I don't know what the proper solution would look like.


core/src/avm2/globals/flash/display/displayobjectcontainer.rs
```diff
@@ -192,7 +193,20 @@ pub fn add_child<'gc>(
                 .cloned()
                 .unwrap_or(Value::Undefined)
                 .as_object()
-                .and_then(|o| o.as_display_object())
+                .ok_or("ArgumentError: Child not a valid display object")?;
+            if child.as_display_object().is_none() {
+                let sprite = activation.resolve_class(&Multiname::new(
+                    Namespace::package("flash.display", activation.context.gc_context),
+                    "Sprite",
+                ))?;
```
Bale001, line 201: This can just be
```rs
let sprite = activation.avm2().classes().sprite;
```

adrian17 (approved):

PR #9561, opened 2023-02-13
Add a regression test for `SoundMixer.computeSpectrum`

torokati44:
The new audio backend simply mixes ~23ms worth of samples every ~1ms, then does nothing with them - it's only to leave the output history in the mixer to be accessed by AS.

I hope the .wav file is not too large...


torokati44:
Had to make both the test backend and the test run with "proper" timing... :/

tests/tests/util/options.rs
```diff
@@ -85,6 +86,7 @@ pub struct PlayerOptions {
     max_execution_duration: Option<Duration>,
     viewport_dimensions: Option<ViewportDimensions>,
     with_renderer: Option<RenderOptions>,
+    with_audio: Option<bool>,
```
Dinnerbone, line 89: This could just be bool I think, the other things are option because they have different behaviour for not being set at all

tests/tests/util/runner.rs
```diff
@@ -13,6 +18,31 @@ use std::rc::Rc;
...
+
+impl TestAudioBackend {
+    pub fn new() -> Self {
+        let mixer = AudioMixer::new(2, 44100);
+        let mixer_proxy = mixer.proxy();
+
+        std::thread::spawn(move || loop {
+            let mut buffer = [0f32; 2 * 1024];
+            mixer_proxy.mix::<f32>(&mut buffer);
+            std::thread::sleep(std::time::Duration::from_millis(1000 * 1024 / 44100));
```
Dinnerbone, line 33: Just for my own sanity, can you extract 44100 into a const, I'm guessing that's `SAMPLE_RATE`?
What's 1024?

tests/tests/util/options.rs
torokati44: Ah, alright. I thought they were `Option` so `serde` won't require them in the `.toml`. Changed to simply `bool`.

tests/tests/util/runner.rs
torokati44: Already did this yesterday, just didn't commit/push then... :)

Dinnerbone:
Looks good to me, but please update `tests/README.md` with the new field!

torokati44:
Done, thanks for the review!

Dinnerbone (approved):

torokati44:
Even with enabling sleeping to match movie FPS, and mixing at approximately real time, this test relies on two separate threads sleeping and waking up at the right times. This means that the test is not guaranteed to be 100% reliable, especially on resource-starved CI runners. There is enough slack in there that it should not cause problems in practice - which is why it takes a couple seconds to run -, but just FYI.

If it turns out to be problematic, we could also try changing the test backend so that mixing is optionally driven by the `computeSpectrum` calls themselves, instead of a separate thread - which would make it entirely deterministic, even without sleeping to match movie FPS.

PR #9572, opened 2023-02-14
desktop: Support flashvars passed via URL parameters

### LINKED ISSUE #9571: desktop: Support flashvars passed via URL parameters
**Describe the feature**
Quoting from [the Adobe site](https://helpx.adobe.com/flash/kb/pass-variables-swfs-flashvars.html), "Variables can also be passed to SWF files by appending them to the URL of SWF files." This works in the Flash projector too, not just in the browser. But it does not work in the Ruffle desktop app. For example, running `ruffle https://swf.example.com/example.swf?testvar=value1` would not pass the `testvar` value to `example.swf`. Passing `-Ptestvar=value1` does work, but it is cumbersome to manually separate each variable-value pair into its own `-P` argument this way.

Note that if Ruffle adds this feature, it will need to parse the query string into variable-value pairs, probably with the [query_pairs](https://docs.rs/url/latest/url/struct.Url.html#method.query_pairs) function. It would serve the same function as the [sanitizeParameters function](https://github.com/ruffle-rs/ruffle/blob/cfa3363416d3046baca48fe084cb675c0a786de0/web/packages/core/src/ruffle-player.ts#L93) on web, which might make the latter function redundant.

n0samu:
Closes #9571.

### DIFF
--- a/desktop/src/main.rs
+++ b/desktop/src/main.rs
@@ -314,9 +314,11 @@ impl App {
             let _ = event_loop_proxy.send_event(RuffleEvent::OnMetadata(swf_header.clone()));
         };
 
+        let mut parameters: Vec<(String, String)> = movie_url.query_pairs().into_owned().collect();
+        parameters.extend(parse_parameters(&opt));
         player.lock().expect("Cannot reenter").fetch_root_movie(
             movie_url.to_string(),
-            parse_parameters(&opt).collect(),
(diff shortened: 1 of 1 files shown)


Dinnerbone (approved):
Nicely done!

adrian17 (approved):

PR #9575, opened 2023-02-14
web: Show context menu on long-press on iOS after 0.8 seconds

danielhjacobs:


danielhjacobs:
Quoting nosamu on Discord:

> it seems like the click event is still getting fired after I release the long press, which either immediately selects an option on the menu or immediately dismisses the menu

Quoting myself:

> If it works, I think the best thing would be to set pointer-events to "none" (either directly in or in the timeout of) startLongPressTimer, and set the pointer-events to "" in clearLongPressTimer, maybe after a 0.05 second timeout.

I need someone with iOS to test if disabled pointer-events in this way works to stop the auto-click.

danielhjacobs:
Even though it has the same problems on iOS in its current state, I think I prefer the design of https://github.com/ruffle-rs/ruffle/pull/9598, so closing this.

PR #9588, opened 2023-02-16
web: Build plugin-polyfill.ts into content.ts the 'correct' way

### LINKED ISSUE #9569: web/packages/extension/src/content.ts contains minified code
Lines 145-147 specifically. This caused us to fail our extension review.

The intent of this code was to load `plugin-polyfill.ts` early; so we need a build system pass to compile this file separately and inject it back as a string into the compiled `content.ts`. The source of that file should have a clear explanation in the comments of how this process works.

kmeisthax:
Closes #9569.

The fix is to make Webpack compile just the plugin polyfill, and then add a new script to search-and-replace the content script with the contents of the compiled plugin polyfill. A comment is also added to the injection site for any curious Mozilla reviewers.


n0samu:
This will also be helpful for development going forward, thank you! I never could figure out how to generate that minified code in the past, so when I made PRs that touched the plugin polyfill, I had to edit the minified code manually. A real pain!

PR #9598, opened 2023-02-16
web: Allow context menu to work on iOS using PointerEvents

danielhjacobs:
Supersedes https://github.com/ruffle-rs/ruffle/pull/9575

### DIFF
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -133,6 +133,9 @@ export class RufflePlayer extends HTMLElement {
     // Whether this device is a touch device.
     // Set to true when a touch event is encountered.
     private isTouch = false;
+    // Whether this device sends contextmenu events.
+    // Set to true when a contextmenu event is seen.
+    private contextMenuSupported = false;
 
     // The effective config loaded upon `.load()`.
     private loadedConfig: Required<Config> = DEFAULT_CONFIG;
@@ -221,10 +224,21 @@ export class RufflePlayer extends HTMLElement {
         this.preloader = this.shadow.getElementById("preloader")!;
 
         this.contextMenuElement = this.shadow.getElementById("context-menu")!;
-        this.addEventListener("contextmenu", this.noContextMenu.bind(this));
-        this.addEventListener("pointerdown", this.pointerDown.bind(this));
-        this.addEventListener("pointerup", this.clearLongPressTimer.bind(this));
-        this.addEventListener(
+        window.addEventListener("contextmenu", this.contextMenu.bind(this), {
+            once: true,
+        });
+        window.addEventListener("pointerdown", this.pointerDown.bind(this));
+        window.addEventListener("pointerup", this.hideContextMenu.bind(this));
+        this.addEventListener("contextmenu", this.showContextMenu.bind(this));
+        this.container.addEventListener(
+            "pointerdown",
+            this.startLongPressTimer.bind(this)
+        );
+        this.container.addEventListener(
+            "pointerup",
+            this.checkLongPress.bind(this)
+        );
+        this.container.addEventListener(
             "pointercancel",
             this.clearLongPressTimer.bind(this)
         );
@@ -237,7 +251,6 @@ export class RufflePlayer extends HTMLElement {
             "webkitfullscreenchange",
             this.fullScreenChange.bind(this)
         );
-        window.addEventListener("click", this.hideContextMenu.bind(this));
 
         this.instance = null;
         this.onFSCommand = null;
@@ -778,19 +791,13 @@ export class RufflePlayer extends HTMLElement {
     }
 
     private pointerDown(event: PointerEvent): void {
-        if (event.pointerType === "mouse" && event.button === 2) {
-            this.showContextMenu.bind(this)(event);
-        }
-        // Give option to disable context menu when touch support is being used
-        // to avoid a long press triggering the context menu. (#1972)
         if (event.pointerType === "touch" || event.pointerType === "pen") {
             this.isTouch = true;
-            this.clearLongPressTimer.bind(this)(event);
-            this.longPressTimer = setTimeout(
-                () => this.showContextMenu.bind(this)(event),
-                this.longPressTimeout
-            );
         }
+    }
+
+    private contextMenu(): void {
+        this.contextMenuSupported = true;
     }
 
     /**
@@ -893,6 +900,8 @@ export class RufflePlayer extends HTMLElement {
                 window.open(RUFFLE_ORIGIN, "_blank");
             },
         });
+        // Give option to disable context menu when touch support is being used
+        // to avoid a long press triggering the context menu. (#1972)
         if (this.isTouch) {
             items.push(null);
             items.push({
@@ -903,18 +912,34 @@ export class RufflePlayer extends HTMLElement {
         return items;
     }
 
-    private clearLongPressTimer(e: PointerEvent): void {
-        if (this.longPressTimer && e.pointerType !== "mouse") {
+    private clearLongPressTimer(): void {
+        if (this.longPressTimer) {
             clearTimeout(this.longPressTimer);
             this.longPressTimer = null;
         }
     }
 
-    private noContextMenu(e: MouseEvent): void {
-        e.preventDefault();
+    private startLongPressTimer(): void {
+        this.clearLongPressTimer();
+        this.longPressTimer = setTimeout(
+            () => this.clearLongPressTimer(),
+            this.longPressTimeout
+        );
     }
 
-    private showContextMenu(e: PointerEvent): void {
+    private checkLongPress(event: PointerEvent): void {
+        if (this.longPressTimer) {
+            this.clearLongPressTimer();
+        } else if (
+            !this.contextMenuSupported &&
+            event.pointerType !== "mouse"
+        ) {
+            this.showContextMenu(event);
+        }
+        event.stopPropagation();
+    }
+
+    private showContextMenu(e: MouseEvent | PointerEvent): void {
         e.preventDefault();
 
         if (
@@ -961,7 +986,7 @@ export class RufflePlayer extends HTMLElement {
                 this.contextMenuElement.appendChild(menuItem);
 
                 if (enabled !== false) {
-                    menuItem.addEventListener("click", onClick);
+                    menuItem.addEventListener("pointerup", onClick);
                 } else {
                     menuItem.classList.add("disabled");
                 }
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -928,6 +928,18 @@ export class RufflePlayer extends HTMLElement {
     }
 
     private checkLongPress(event: PointerEvent): void {
+        // The is fired as a pointerup event on the container
+        // However, pointerup on the window will dismiss the context menu
+        // To stop the contextmenu from being dismissed immediately, we stopPropagation:
+        // 1) If a mouse did not fire this event and longPressTimer is not active
+        // (there was a previous event that cleared it)
+        // 2) If this pointerup event was fired by right-click on a mouse
+        if (
+            (!this.longPressTimer && event.pointerType !== "mouse") ||
+            (event.pointerType === "mouse" && event.button === 2)
+        ) {
+            event.stopPropagation();
+        }
         if (this.longPressTimer) {
             this.clearLongPressTimer();
         } else if (
@@ -936,7 +948,6 @@ export class RufflePlayer extends HTMLElement {
         ) {
             this.showContextMenu(event);
         }
-        event.stopPropagation();
     }
 
     private showContextMenu(e: MouseEvent | PointerEvent): void {
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -148,6 +148,8 @@ export class RufflePlayer extends HTMLElement {
     private _cachedDebugInfo: string | null = null;
 
     private isExtension = false;
+    private longPressTimer: ReturnType<typeof setTimeout> | null = null;
+    private longPressTimeout = 800;
 
     /**
      * Triggered when a movie metadata has been loaded (such as movie width and height).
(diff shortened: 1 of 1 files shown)


web/packages/core/src/ruffle-player.ts
```diff
@@ -220,8 +222,14 @@ export class RufflePlayer extends HTMLElement {
         this.preloader = this.shadow.getElementById("preloader")!;
 
         this.contextMenuElement = this.shadow.getElementById("context-menu")!;
-        this.addEventListener("contextmenu", this.showContextMenu.bind(this));
+        this.addEventListener("contextmenu", this.noContextMenu.bind(this));
```
relrelb, line 225: Wouldn't this break context menu on desktop (where `contextmenu` is triggered by a right-click)?

web/packages/core/src/ruffle-player.ts
```diff
@@ -772,10 +780,18 @@ export class RufflePlayer extends HTMLElement {
     }
 
     private pointerDown(event: PointerEvent): void {
+        if (event.pointerType === "mouse" && event.button === 2) {
+            this.showContextMenu.bind(this)(event);
```
relrelb, line 784: The `.bind(this)` is totally redundant (here and in other places).

web/packages/core/src/ruffle-player.ts
n0samu: No because pointer events cover mouse events and touch events. See the code near your other comment for how mouse pointer events are handled. The "contextmenu" event is blocked here simply to prevent the default browser context menu from appearing.

web/packages/core/src/ruffle-player.ts
n0samu: Ah right, here's some more info about that: https://javascript.info/bind
Basically it's only necessary when the function is being passed to something outside the object, like the setTimeout function.

web/packages/core/src/ruffle-player.ts
relrelb: Got it, but I still feel like using the `contextmenu` event is preferred when available, to avoid simulating its (not no complex, but still) logic ourselves. This can be accomplished by skipping the added functionality in the `pointerdown` handler if the `contextmenu` event is available, which can be checked via:
```js
"oncontextmenu" in window
```

n0samu: It's worth noting that Flash Player's context menu behavior is closer to this behavior than that of the default `contextmenu` event. Flash Player's context menu appears on rightMouseButtonDown, while the `contextmenu` event is fired on rightMouseButtonUp.

n0samu: Also, I just checked, and `"oncontextmenu" in window` returns `true` on my iPhone.

danielhjacobs: > "oncontextmenu" in window returns true on iPhone.

This is the issue and the reason I can't use feature detection even though I wanted to too. I asked Adrian to check on discord and he found that this was the case. It's because the context menu event can fire when you plug a Bluetooth mouse into an iPhone. Just not with long press.

danielhjacobs: It's also arguably better to use the same or similar logic for both desktop and phones. Also the noContextMenu function is because without this function on desktop right click would show the default browser context menu above the Ruffle context menu.

web/packages/core/src/ruffle-player.ts
danielhjacobs: Wasn't sure about this. I'll remove these instances.

danielhjacobs:
nosamu mentioned making changes to use the pointerUp event instead of the click event to dismiss the context menu, use stopPropagation when creating the contextMenu, and show the context menu on pointerup instead of pointerdown. Apparently without those changes, on iOS, it seems like the click event is still getting fired after nosamu releases the long press, which either immediately selects an option on the menu or immediately dismisses the menu. As I do not have iOS, I cannot test those kinds of changes, so nosamu is going to creating a commit on this branch on my fork with those changes when possible, and I'll merge them here.

web/packages/core/src/ruffle-player.ts
danielhjacobs: nosamu added the same type of feature detection isTouch has.

web/packages/core/src/ruffle-player.ts
```diff
@@ -772,13 +793,15 @@ export class RufflePlayer extends HTMLElement {
...
     private pointerDown(event: PointerEvent): void {
-        // Give option to disable context menu when touch support is being used
-        // to avoid a long press triggering the context menu. (#1972)
         if (event.pointerType === "touch" || event.pointerType === "pen") {
             this.isTouch = true;
         }
     }
 
+    private contextMenu(): void {
+        this.contextMenuSupported = true;
```
relrelb, line 802: Maybe move this to `showContextMenu()`?

web/packages/core/src/ruffle-player.ts
```diff
@@ -883,7 +908,45 @@ export class RufflePlayer extends HTMLElement {
...
+            this.clearLongPressTimer();
+        } else if (
+            !this.contextMenuSupported &&
+            event.pointerType !== "mouse"
+        ) {
+            this.showContextMenu(event);
+        }
+    }
+
+    private showContextMenu(e: MouseEvent | PointerEvent): void {
```
relrelb, line 949: While here, let's rename `e` to `event`.

web/packages/core/src/ruffle-player.ts
```diff
@@ -149,6 +152,8 @@ export class RufflePlayer extends HTMLElement {
     private _cachedDebugInfo: string | null = null;
 
     private isExtension = false;
+    private longPressTimer: ReturnType<typeof setTimeout> | null = null;
+    private longPressTimeout = 800;
```
relrelb, line 156: I think it's more appropriate to define this as `const` (even inside the body of `startLongPressTimer()` itself).

web/packages/core/src/ruffle-player.ts
```diff
@@ -149,6 +152,8 @@ export class RufflePlayer extends HTMLElement {
     private _cachedDebugInfo: string | null = null;
 
     private isExtension = false;
+    private longPressTimer: ReturnType<typeof setTimeout> | null = null;
```
relrelb, line 155: Actually, now it seems like `setTimeout` is not need at all, since we only actively check whether the given timeout has elapsed or not, rather than passively waiting for a callback to be triggered.
So, instead we can simply store the last timestamp a `pointdown` occurred (via `Date.now()`), then in `checkLongPress()` compare it against the current timestamp to tell whether enough time has elapsed.

relrelb:
I didn't test this, but I think that multi-touch isn't handled properly (i.e. touching with a second finger would "override" the timeout of the first one). Maybe we should track each "finger" separately?
See more info about multi-touch events at: https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events/Multi-touch_interaction

web/packages/core/src/ruffle-player.ts
relrelb: On a second thought maybe showing the context menu on `pointerup` can be inconvenient. It makes sense with a right-click, but with touch the user might not know how long they need to keep touching. Also context menu on Android is shown when still touching (and I believe on iOS as well).
If so, maybe we should return back to `pointerdown` and which case you can scratch my previous comment.

web/packages/core/src/ruffle-player.ts
danielhjacobs: `showContextMenu()` is not just fired on the contextmenu event. It is also fired in checkLongPress. Though this could be done if switching to nosamu's code from https://github.com/ruffle-rs/ruffle/compare/master...n0samu:ruffle:ios_pointer_event_3, which I actually happen to prefer. Namely, this can be added to the `if (e.type === "contextmenu")`  condition within `showContextMenu()`. However, this is currently a `once` event, and would not be with that change, not that that probably matters.

web/packages/core/src/ruffle-player.ts
danielhjacobs: > So, instead we can simply store the last timestamp a `pointdown` occurred (via `Date.now()`), then in `checkLongPress()` compare it against the current timestamp to tell whether enough time has elapsed.

I'll be honest, I prefer setTimeout, even without a callback. Time comparisons in JavaScript are finicky, while `setTimeout` is a well-documented standard function of the language.

Also, the contextmenu behavior is the exact behavior that every context menu follows on Android, from popping up after a little under a second to appearing before releasing touch, everything is exactly as it is with every other context menu functionality, and therefore exactly matches what an Android user expects. The idea here is just to match that behavior with the behavior on an iPhone. Popping up the context menu on any touch (pointerdown) without waiting severely messes with the functionality of touch-based games. Needing to hide the context menu to do anything on touch-based games on mobile severely detracts from the usefulness of the options within it.

danielhjacobs: Also, I think clearTimeout is needed so it can be cancelled when touch is released before the elapsed time has passed.

n0samu: @relrelb Here is my attempt to make the iOS context menu work like it does on Android (on pointerdown instead of pointerup): https://github.com/n0samu/ruffle/commit/4a655bf1be9251c6684597cb2a10120c3c0c4e39
With the context menu appearing on the pointerdown event, now there is no way to stop the pointerup event from propagating and immediately closing the context menu. The only solution I found was to add an event listener, which itself adds another event listener that finally closes the context menu. That way the context menu closes after the second pointerup event that is received, not the first (which fires immediately after the user lifts their finger).

However, I think this is a considerably worse user experience and would take far too much effort to improve. The context menu appears where the user's touch began, since that is the location of the pointerdown event. If we wanted the context menu to appear in the right place, we would need to store the user's current finger position and update it on every pointermove event. And then we have the problem of detecting "long presses": how do we detect that the user's finger moved "too far" for this to be a long press? It's pretty annoying for the context menu to appear while you're sliding your finger. It's comparatively less annoying for it to appear after you lift your finger (though it's still not ideal).

So can we please leave things as-is and proceed with merging this PR?

n0samu:
@danielhjacobs Thanks for your work on this! I tested the latest commit on iOS and it's still working great.

@relrelb Considering that the rest of Ruffle has no support for multitouch, I don't think it's quite reasonable to expect it here. It would be pretty complicated to implement.

web/packages/core/src/ruffle-player.ts
```diff
@@ -887,8 +903,49 @@ export class RufflePlayer extends HTMLElement {
...
+            longPressTimeout
+        );
+    }
+
+    private checkLongPress(event: PointerEvent): void {
+        if (this.longPressTimer) {
+            this.clearLongPressTimer();
+        } else if (
+            !this.contextMenuSupported &&
+            event.pointerType !== "mouse"
```
relrelb, line 927: nit: Do we can to check for `event.pointerType === "touch" || event.pointerType === "pen"` like we do in `pointerDown()`?

web/packages/core/src/ruffle-player.ts
relrelb: @n0samu Good arguments - I wasn't aware of them and they do make sense.
Indeed simulating UI logic can be an headache, but IIUC even under the current approach, if the user drags their finger "too far" for at least 800ms, then a context menu will be shown as well. The only difference is that this will happen when the finger is lifted, am I right?
So considering this, it seems like the "too far" case should be detected either way. I think this can implemented relatively easily - In `startLongPressTimer()` store the starting coordinates, then in `checkLongPress()` compare the distance between the starting and final coordinates against some threshold. If it's too far - don't show a context menu.
However this is only a partial implementation, since a context menu would still be shown when the finger is dragged a long distance but lifed just where it started... To handle this case the `pointermove` event should be used like you mentioned.

Which approach should be taken? I'm not sure. We just need to decide which is "good enough".

n0samu: > The only difference is that this will happen when the finger is lifted, am I right?

Yes that's right.

As for the "too far" behavior, I'm not familiar enough with Android to say how it works or if it even works that way at all. And tuning this behavior would require some experimentation. I'm not really up for all that.

relrelb: I'll try to summarize the conditions for a context menu on Android:
1. It is shown on `pointerdown` (doesn't wait for the finger to be lifted).
2. The finger must be kept "in place" for about 0.5-1 seconds.
3. If the finger is moved sooner, then the underlying text is selected.

I don't have an available iOS device, but I believe it behaves the same.
Anyway, I guess the current logic is fine enough, given that changing it further is going to be complex. Worst-case we'll get feedback from users and tune it accordingly.

So besides my 2 comments below, I see this PR ready to be merged.

danielhjacobs: Yeah, as nosamu mentioned switching from pointerup to pointerdown is complex. I also noticed the timing being somewhere between 0.5 and 1 seconds, which is why I guesstimated 0.8 seconds. And pointermove appears to fire at the slightest amount of movement, even the natural resting movement of your finger, so that's what makes the too far definition complex. I'll address the new comments in the next few hours.

web/packages/core/src/ruffle-player.ts
danielhjacobs: We could, though this check is arguably more inclusive of all possibilities: https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pointerType#value:

> If the device type cannot be detected by the browser, the value can be an empty string (`""`). If the browser supports pointer device types other than those listed above, the value should be *vendor-prefixed* to avoid conflicting names for different types of devices.

I don't actually own or know of any devices with other pointer types, but they are documented.

danielhjacobs: So with that in mind, I can change this, change pointerDown to be consistent with this instead, or change nothing. What do you think is best?

relrelb: Do we even need this condition at all? I can't think of a scenario where `!this.contextMenuSupported` and `event.pointerType` is not "touch".
Unless I miss something, I think it can be removed.

danielhjacobs: From a theoretical perspective, you're correct. However, according to @n0samu on Discord: "the 'feature detection' method doesn't work unless you actually right-click" and "the solution to this problem is simple, just don't trigger the context menu on pointerup if the type of pointer is mouse". Perhaps a more accurate variabloe name would be `contextMenuTriggered` and perhaps this could use a comment.

danielhjacobs: That's because `this.contextMenuSupported` is set whenthe contextmenu event fires, but this check occurs on `pointerup`, which can fire either before or after the `contextmenu` event on different devices (linux and Windows differ here).

relrelb: Got it. So let's keep the code as-is and add the comment you wrote.

relrelb (approved):
LGTM! Sorry for the long back and forth...

danielhjacobs:
Thanks and no problem, mimicking this logic for iOS is tricky and it's good to try to do it the best way possible. I don't have write access to the repo, so I'll need you to merge the PR unless you think it needs other reviewers.

PR #9608, opened 2023-02-17
desktop: min and max window size for ruffle

### LINKED ISSUE #9346: [bug][meme] Very Long Ruffle.exe Window (with a 17961x253 screenshot of Ruffle Window)
### Describe the bug

Very Long Ruffle Window

As you see here. Ruffle.exe Window does not have a size limit. I can increase its size indefinitely, by dragging the right border to the right, and dragging the window to the left, over and over again.

### Expected behavior

I expect TWO things, instead of ONE.
1. The obvious one: User should not be allowed to resize this window this much. 
2. Actually, I do see **a few .swf files defaults at a "larger than my screen resolution" resolution**. So a extra large ruffle window is achieved simply by opening these .swf files. This shouldn't happen as well.
_(Not going to tell you which .swf though, because it's a NSFW one, and because it should be pretty easy for you to create a test .swf file for this, just create a .swf that defaults at 2560x1440 resolution, and test it on a 1920x1080 screen, something like that.)

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

Actually, could you document this bug, like exactly how it happens? This is the first time ever I've seen a Windows app surpass the window size limit.

aria-aghaei:
Closes #9346 

My first PR, so any comments, suggestions, etc. are appreciated.

### DIFF
--- a/desktop/src/main.rs
+++ b/desktop/src/main.rs
@@ -225,16 +225,16 @@ fn load_movie(url: &Url, opt: &Opt) -> Result<SwfMovie, Error> {
     Ok(movie)
 }
 
-fn get_max_screen_size(event_loop: &EventLoop<RuffleEvent>) -> (u32, u32) {
+fn get_max_screen_size(monitors: impl IntoIterator<Item = MonitorHandle>) -> (u32, u32) {
     let mut max_screen_size = (0, 0);
 
-    for monitor in EventLoopWindowTarget::available_monitors(event_loop) {
+    for monitor in monitors {
         let size = MonitorHandle::size(&monitor);
         max_screen_size.0 += size.width;
         max_screen_size.1 += size.height;
     }
 
-    if max_screen_size.0 <= 32 || max_screen_size.1 <= 32 {
+    if max_screen_size.0 <= 64 || max_screen_size.1 <= 32 {
         return (32767, 32767);
     }
 
@@ -275,7 +275,8 @@ impl App {
         let title = format!("Ruffle - {filename}");
         SWF_INFO.with(|i| *i.borrow_mut() = Some(filename.to_string()));
 
-        let max_screen_size = get_max_screen_size(&event_loop);
+        let monitors = EventLoopWindowTarget::available_monitors(&event_loop);
+        let max_screen_size = get_max_screen_size(monitors);
 
         let window = WindowBuilder::new()
             .with_visible(false)
@@ -568,21 +569,38 @@ impl App {
                         let movie_width = swf_header.stage_size().width().to_pixels();
                         let movie_height = swf_header.stage_size().height().to_pixels();
 
+                        let monitors = Window::available_monitors(&self.window);
+                        let max_screen_size = get_max_screen_size(monitors);
+
                         let window_size: Size = match (self.opt.width, self.opt.height) {
-                            (None, None) => LogicalSize::new(movie_width, movie_height).into(),
+                            (None, None) => LogicalSize::new(
+                                movie_width.clamp(16.0, max_screen_size.0 as f64),
+                                movie_height.clamp(16.0, max_screen_size.1 as f64),
+                            )
+                            .into(),
                             (Some(width), None) => {
                                 let scale = width / movie_width;
                                 let height = movie_height * scale;
-                                PhysicalSize::new(width.max(1.0), height.max(1.0)).into()
+                                PhysicalSize::new(
+                                    width.clamp(16.0, max_screen_size.0 as f64),
+                                    height.clamp(16.0, max_screen_size.1 as f64),
+                                )
+                                .into()
                             }
                             (None, Some(height)) => {
                                 let scale = height / movie_height;
                                 let width = movie_width * scale;
-                                PhysicalSize::new(width.max(1.0), height.max(1.0)).into()
+                                PhysicalSize::new(
+                                    width.clamp(16.0, max_screen_size.0 as f64),
+                                    height.clamp(16.0, max_screen_size.1 as f64),
+                                )
+                                .into()
                             }
-                            (Some(width), Some(height)) => {
-                                PhysicalSize::new(width.max(1.0), height.max(1.0)).into()
-                            }
+                            (Some(width), Some(height)) => PhysicalSize::new(
+                                width.clamp(16.0, max_screen_size.0 as f64),
+                                height.clamp(16.0, max_screen_size.1 as f64),
+                            )
+                            .into(),
                         };
                         self.window.set_inner_size(window_size);
                         self.window.set_fullscreen(if self.opt.fullscreen {
--- a/desktop/src/main.rs
+++ b/desktop/src/main.rs
@@ -41,8 +41,7 @@ use winit::event::{
     ElementState, KeyboardInput, ModifiersState, MouseButton, MouseScrollDelta, VirtualKeyCode,
     WindowEvent,
 };
-use winit::event_loop::{ControlFlow, EventLoop, EventLoopBuilder, EventLoopWindowTarget};
-use winit::monitor::MonitorHandle;
+use winit::event_loop::{ControlFlow, EventLoop, EventLoopBuilder};
 use winit::window::{Fullscreen, Icon, Window, WindowBuilder};
 
 thread_local! {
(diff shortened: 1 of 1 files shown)


desktop/src/main.rs
```diff
@@ -254,9 +255,23 @@ impl App {
...
+        let mut max_resolution_y = 0;
+
+        for monitor in Window::available_monitors(&window) {
+            let size = MonitorHandle::size(&monitor);
+            max_resolution_x += size.width;
+            max_resolution_y += size.height;
+        };
+
+        if max_resolution_x > 0 && max_resolution_y > 0 {
+            Window::set_max_inner_size(&window, Some(LogicalSize::new(max_resolution_x, max_resolution_y)));
```
Dinnerbone, line 272: Can you run this before we build the window and use the value you calculate in the builder? Maybe extract it to a method that returns the size tuple

desktop/src/main.rs
```diff
@@ -43,6 +43,7 @@ use winit::event::{
 };
 use winit::event_loop::{ControlFlow, EventLoop, EventLoopBuilder};
 use winit::window::{Fullscreen, Icon, Window, WindowBuilder};
+use winit::monitor::{MonitorHandle};
```
Dinnerbone, line 46: Can you run cargo fmt please? Thank you!

Dinnerbone (approved):
Code and idea looks sound just some nitpicks, thank you so much!

Dinnerbone (changes requested):
Last one meant to be requested changes, sorry, miclick!

aria-aghaei:
It should be good now, I think

aria-aghaei:
Forgot to say thanks for suggestions, I'm pretty new at rust in general

aria-aghaei:
oops, I see those linting errors, I will fix

n0samu:
Even with this PR, it seems that SWFs with tiny/large dimensions can make the window open at tiny/large resolutions. It seems like the maximum window size restriction actually only applies to the user resizing the window. So I think we should limit the size that the window gets resized to when the movie is loaded, too: 

https://github.com/ruffle-rs/ruffle/blob/15435c6bea9468434271f5c36b748c9b07c5364d/desktop/src/main.rs#L538-L540

Here are the SWFs that I've been testing with: [ResolutionTests.zip](https://github.com/ruffle-rs/ruffle/files/10772045/ResolutionTests.zip)

aria-aghaei:
> Even with this PR, it seems that SWFs with tiny/large dimensions can make the window open at tiny/large resolutions. It seems like the maximum window size restriction actually only applies to the user resizing the window. So I think we should limit the size that the window gets resized to when the movie is loaded, too:

Hi, just pushed a commit that addresses this by clamping the window size on creation. Tested on both of those flashes in the zip as well as a few of my own and it seems to work.

Let me know any suggestions, comments, etc. Thanks!

desktop/src/main.rs
```diff
@@ -216,6 +217,22 @@ fn load_movie(url: &Url, opt: &Opt) -> Result<SwfMovie, Error> {
...
+    let mut max_screen_size = (0, 0);
+
+    for monitor in monitors {
+        let size = MonitorHandle::size(&monitor);
+        max_screen_size.0 += size.width;
+        max_screen_size.1 += size.height;
+    }
+
+    if max_screen_size.0 <= 64 || max_screen_size.1 <= 32 {
+        return (32767, 32767);
```
Dinnerbone, line 230: This is `(i16::MAX, i16::MAX)` right? Can you change it to those constants, just so we know it's max and not a value with any special significance?

desktop/src/main.rs
```diff
@@ -539,21 +560,38 @@ impl App {
                         let movie_width = swf_header.stage_size().width().to_pixels();
                         let movie_height = swf_header.stage_size().height().to_pixels();
 
+                        let monitors = Window::available_monitors(&self.window);
+                        let max_screen_size = get_max_screen_size(monitors);
+
                         let window_size: Size = match (self.opt.width, self.opt.height) {
-                            (None, None) => LogicalSize::new(movie_width, movie_height).into(),
+                            (None, None) => LogicalSize::new(
+                                movie_width.clamp(16.0, max_screen_size.0 as f64),
```
relrelb, line 568: To avoid repeating the same `clamp`s 4 times, winit's [`Size::clamp`](https://docs.rs/winit/latest/winit/dpi/enum.Size.html#method.clamp) can be used.
Also I think `max_inner_size` (and `min_inner_size`) should be stored in `App` itself so it can be retrieved here rather than computed again.

desktop/src/main.rs
relrelb: Also why this case is needed?

aria-aghaei: Just a sanity check, in case winit couldn't detect any monitors we don't want it to return a max size of 0,0. But if the monitor detection always succeeds we wouldn't need it.

desktop/src/main.rs
```diff
@@ -216,6 +217,22 @@ fn load_movie(url: &Url, opt: &Opt) -> Result<SwfMovie, Error> {
     Ok(movie)
 }
 
+fn get_max_screen_size(monitors: impl IntoIterator<Item = MonitorHandle>) -> (u32, u32) {
+    let mut max_screen_size = (0, 0);
+
+    for monitor in monitors {
+        let size = MonitorHandle::size(&monitor);
```
relrelb, line 224: nit: This can be written as `monitor.size()`.

desktop/src/main.rs
```diff
@@ -216,6 +217,22 @@ fn load_movie(url: &Url, opt: &Opt) -> Result<SwfMovie, Error> {
     Ok(movie)
 }
 
+fn get_max_screen_size(monitors: impl IntoIterator<Item = MonitorHandle>) -> (u32, u32) {
```
relrelb, line 220: How about returning `PhysicalSize<u32>` here?

desktop/src/main.rs
```diff
@@ -216,12 +217,29 @@ fn load_movie(url: &Url, opt: &Opt) -> Result<SwfMovie, Error> {
     Ok(movie)
 }
 
+fn get_max_screen_size(monitors: impl IntoIterator<Item = MonitorHandle>) -> PhysicalSize<u32> {
```
relrelb, line 220: Now that `get_max_screen_size` has a single caller, it can be made less generic: Accept an `event_loop: &EventLoop` parameter and call `event_loop.available_monitors()` here.

desktop/src/main.rs
```diff
@@ -216,12 +216,29 @@ fn load_movie(url: &Url, opt: &Opt) -> Result<SwfMovie, Error> {
     Ok(movie)
 }
 
+fn get_max_screen_size(event_loop: &EventLoop<RuffleEvent>) -> PhysicalSize<u32> {
+    let mut max_screen_size = (0, 0);
+
+    for monitor in event_loop.available_monitors() {
+        let size = monitor.size();
+        max_screen_size.0 += size.width;
```
relrelb, line 224: I think summing all widths and heights isn't the way to go.
Consider I have two monitors side-by-side. Now the max screen height will be calculated as the sum of their heights, which is obviously incorrect. What we really want is the *maximum* of their heights.
If one monitor is on top of the other, then we want the sum of their heights, but the maximum of their widths.

My point is that the calculation must take the monitor positions into account. Fortunately winit provides [`MonitorHandle::position()`](https://docs.rs/winit/latest/winit/monitor/struct.MonitorHandle.html#method.position). 
So I think the best way to use it is to accumulate the minimum and maximum of both X and Y coordinates:
```rs
fn get_max_screen_size(event_loop: &EventLoop<RuffleEvent>) -> PhysicalSize<u32> {
    let mut min_x = 0;
    let mut min_y = 0;
    let mut max_x = 0;
    let mut max_y = 0;
    for monitor in event_loop.available_monitors() {
        let size = monitor.size();
        let position = monitor.position();
        min_x = min_x.min(position.x);
        min_y = min_y.min(position.y);
        max_x = max_x.max(position.x + size.width as i32);
        max_y = max_y.max(position.y + size.height as i32);
    }
    let width = max_x - min_x;
    let height = max_y - min_y;
    (width, height).into()
}
```

relrelb (approved):
Thank you and welcome!

PR #7935, opened 2022-09-09
core: `ColorTransform` cleanup

relrelb:
Main changes:
* Merge `ColorTransformParams` into `ColorTransformObject`, as it's only relevant for AVM1.
* Make `BitmapData::color_transform` work with a generic `ColorTransform`, which uses fixed-point arithmetic.

Note that Ruffle still calculates color transforms slightly different from Flash. This is probably caused by inaccuracy of the current `ColorTransformObject` to `ColorTransform` conversion and/or the `ColorTransform` application logic itself. Since this requires further research, it'll be fixed in a future PR.

PR #5021, opened 2021-08-06
Move BitmapData implementation up into core::bitmap

torokati44:
This enables exposing it to AVM2 as well in the future.
A fairly mechanical change.
As mentioned in https://github.com/ruffle-rs/ruffle/pull/4940#issuecomment-892167336, and many times before on Discord.

PR #9620, opened 2023-02-17
swf: Introduce filter flags

relrelb:


torokati44:
FWIW: A similar change (by me, I admit...) used `...Params` instead of `...Flags` as names:
https://github.com/ruffle-rs/ruffle/pull/5021/files#diff-3950342c208fbe54cc6242ead16632fa89f950fd338d296560ab79be56965d98

Might not mean anything though, I'm fine with either variant.

relrelb:
> FWIW: A similar change (by me, I admit...) used `...Params` instead of `...Flags` as names: https://github.com/ruffle-rs/ruffle/pull/5021/files#diff-3950342c208fbe54cc6242ead16632fa89f950fd338d296560ab79be56965d98
> 
> Might not mean anything though, I'm fine with either variant.

Yeah, but now I added `bitflags` enums, which makes more sense to me to have `Flags` in their name.
And by the way, `ColorTransformParams` was removed in #7935.

Aaron1011 (approved):

PR #9627, opened 2023-02-18
web: Simplify custom context menu separators logic

relrelb:
Instead deduplicating separators in `RufflePlayer.showContextMenu()` using DOM attributes, do it right in `RufflePlayer.contextMenuItems()`, using a simpler approach.

### DIFF
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -885,7 +885,8 @@ export class RufflePlayer extends HTMLElement {
                 readonly separatorBefore: boolean;
             }[] = this.instance.prepare_context_menu();
             customItems.forEach((item, index) => {
-                if (item.separatorBefore) {
+                // Don't start with separators.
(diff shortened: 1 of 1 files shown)


adrian17 (commented):
TBH instead of now having every case of "adding a separator" check for its correctness, I'd either somehow encapsulate it, _or_ just replace the DOM-based checks in the final loop with index-based checks (`if (item === null && i > 0 && last !== null) { ...`)

relrelb:
> TBH instead of now having every case of "adding a separator" check for its correctness, I'd either somehow encapsulate it, _or_ just replace the DOM-based checks in the final loop with index-based checks (`if (item === null && i > 0 && last !== null) { ...`)

Personally I'd like to keep things simple rather than over-complicaing/over-designing them (that's the intent in this PR). Honestly I don't see a real scenario where said "correctness" would break without us noticing.

Dinnerbone (approved):

Herschel (approved):

Issue #9122, opened 2023-01-11
Steaml won't load


ActionWavele:
### Describe the bug

ok
![slnl](https://user-images.githubusercontent.com/120395932/211930068-3f9c7509-b757-40ac-b935-c7df4982dea7.png)

this game will only get stuck half way, then ruffle will crash.
I used 2023/01/11 Ruffle.

the log said:
```
thread 'main' panicked at 'wgpu error: Validation Error

Caused by:
    In Device::create_bind_group
    not enough memory left

', C:\Users\runneradmin\.cargo\git\checkouts\wgpu-53e70f8674b08dd4\630c12f\wgpu\src\backend\direct.rs:2962:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
AVM2 stack trace:

```

### Expected behavior

the game should load and play

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

Mozilla Firefox

### Additional information

this game works on adobe flash player projector sa 32
[Steaml_P_P_P.zip](https://github.com/ruffle-rs/ruffle/files/10396365/Steaml_P_P_P.zip)

ActionWavele:
I also do not recognise this user...
``C:\Users\runneradmin\``

was I hacked?

Lord-McSweeney:
That's a stack trace from the built Rust code. It's probably from a runner that built the file somewhere else, such as on Github's CI.

ActionWavele:
how did you get that error?
are you running the desktop version like I was?

Lord-McSweeney:
Yes, I was running on the desktop version. It seems like it crashed because you didn't have enough video memory left, did you have a lot of applications open or a slow GPU?

ActionWavele:
it seems to be that my hardware isn't powerful enough
I hope ruffle can become less agressive on my resources, and to fix the error you got
it looks related to XML issues

Lord-McSweeney:
This looks like it needs the AVM2 version of `BitmapData.compare`; the AVM1 version was added by #5684 already.

ActionWavele:
this is caused by the same issue as https://github.com/ruffle-rs/ruffle/issues/8844

it goes further on web, but not much
![steamlw](https://user-images.githubusercontent.com/120395932/220478003-e9428eea-af55-4bf4-9703-3a3c62b8374c.jpg)

this story screen can be skipped by clicking to make all text appear, and again to go to the main menu
clicking does not work.

[...]

Issue #3794, opened 2021-03-25
Fault Line From Nitrome is Impossible to Play


TRRESTGHGYTCFTGBCXSZD:
**Describe the bug**
It Keeps Looping the animation

**Expected behavior**
A Nice Game

**Is the problem with the Ruffle desktop app, extension, or self-hosted version?**

**What platform are you using?**
Windows Embedded 8.1 Industry Pro, Chrome 89.0.4389.90

llacb47:
There is no useful information here.

danielhjacobs:
Assuming http://www.nitrome.com/games/faultline/, the game won't load initially due to CORS see #2201. If you download http://cdn.nitrome.com/games/the-unit-v2.swf directly and load it into the Ruffle demo, you see the warning "This SWF contains ActionScript 3 which is not yet supported by Ruffle. The movie may not work as intended.", and the errors: "AVM2 error: TypeError: undefined is not an Object" and "Got TypeError: undefined is not an Object when constructing AVM2 side of display object". AS3 is unsupported. See #1368.

mero09-99:
The unit v2 swf isn't the game.
its a dummy swf, the game loads later.

n0samu:
Interestingly, this version of the game boots to the menu: http://www.notdoppler.com/faultline.php
But the levels on the level select screen are unselectable.

mero09-99:
this is because faultline, like a few other nitrome games, relies on external level data, with usually hashed names, in xml

n0samu:
> this website must have waited for nitrome's site to load the real swf, hacked past the sitelock and saw the title, and assumed the rest would work

No, the game works in Flash Player, so it's a Ruffle issue

mero09-99:
interesting
is Ruffle imcompatable with loading levels (xml files)?

n0samu:
The version of the game that I linked doesn't load levels via XML files, so it's not that.

Lord-McSweeney:
http://cdn.nitrome.com/games/the-unit-v2.swf directly now works a little better, but requires `encodeURIComponent`.

Lord-McSweeney:
As #9647 has been merged, the game can be played. There are some performance issues, though, and an error:
```
ERROR ruffle_core::avm2::activation: TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: ::length)
	at com.nitrome.folding::Level/getFoldedPoint()
[...]

PR #9647, opened 2023-02-19
avm2: Initial incomplete implementation of XML

Aaron1011:
This is a (very incomplete) implementation of the XML and XMLList classes. Basic parsing works, along with child and attribute lookup.

Namespacse, mutation, and the vast majority of XML/XMLList methods are not implemented. I doubt that this is enough to get any games working, but it should be enough to allow devs to start working on different parts of the XML implementation in parallel (now that the basic framework is somewhat in place).


Lord-McSweeney:
Would String-to-XML coercion be in-scope for this PR?

core/src/avm2/activation.rs
```diff
@@ -2620,7 +2622,17 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
-        self.push_stack(value1 == value2);
+        if value1 == value2 {
+            self.push_stack(true);
+        } else {
+            if let Some(xml1) = value1.as_object().and_then(|obj| obj.as_xml_object()) {
+                if let Some(xml2) = value2.as_object().and_then(|obj| obj.as_xml_object()) {
+                    self.push_stack(E4XNode::ptr_eq(*xml1.node(), *xml2.node()));
+                    return Ok(FrameControl::Continue);
+                }
+            }
```
Bale001, line 2633: Instead of adding this as a special case here, could we add this to Value's `PartialEq` implementation?

Aaron1011: This applies specifically to the strict_eq opcode - I don't think we'll necessarily want this in other places

Lord-McSweeney:
This significantly progresses Fault Line (#3794)- the first level can now be entered and played. The flickering text is still there, though, and a number of errors are thrown.

Lord-McSweeney:
Panics on https://github.com/ruffle-rs/ruffle/issues/9122 with `'called Option::unwrap() on a None value', core/src/avm2/e4x.rs:166:21`

core/src/avm2/activation.rs
adrian17: That's not necessarily true - in avmplus, Array `(last)indexof` uses `stricteq()` too. (and that's... the only exception apparently. But that's because `Vector.indexOf` is implemented in AS, so it implicitly already uses `stricteq` opcode).
That said, I'm also not convinced this should be in default `PartialEq` implementation.

core/src/avm2/e4x.rs
```diff
@@ -0,0 +1,401 @@
...
+            node: E4XNode<'gc>,
+            open_tags: &mut Vec<E4XNode<'gc>>,
+            top_level: &mut Vec<E4XNode<'gc>>,
+            depth: usize,
+            activation: &mut Activation<'_, 'gc>,
+        ) -> Result<(), Error<'gc>> {
+            if !open_tags.is_empty() {
+                open_tags
+                    .last_mut()
+                    .unwrap()
```
adrian17, line 158: `expect()` - I don't really know why it's supposed to be always-true

core/src/avm2/globals/XML.as
```diff
@@ -1,4 +1,16 @@
...
+    public final dynamic class XML {
+        public function XML(value:* = undefined) {
+            this.init(value);
+        }
+
+        private native function init(value:*): void;
+
+        AS3 native function name():Object;
+        AS3 native function localName():Object;
+        AS3 native function toXMLString():String;
```
adrian17, line 12: let's add prototype overrides in this PR too, I'm assuming this works in AS
(if not easily, let's add a comment so we don't forget)

core/src/avm2/globals/xml.rs
```diff
@@ -1,3 +1,81 @@
...
+        Ok(Value::Null)
+    }
+}
+
+pub fn local_name<'gc>(
+    _activation: &mut Activation<'_, 'gc>,
+    this: Option<Object<'gc>>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let node = this.unwrap().as_xml_object().unwrap();
```
adrian17, line 59: Same as in another review (I think?) - IMO let's do the `if let` dance like everywhere else and decide on a direction to refactor it all separately. At least that's my preference, but not a hard blocker if you leave it like this.

core/src/avm2/globals/xml_list.rs
```diff
@@ -1,4 +1,77 @@
...
+    activation: &mut Activation<'_, 'gc>,
+    this: Option<Object<'gc>>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let list = this.unwrap().as_xml_list_object().unwrap();
+    let children = list.children();
+    if has_simple_content_inner(&children) {
+        Ok(simple_content_to_string(children.iter().cloned(), activation)?.into())
+    } else {
+        Err("XMLList.toString() for non-simple content: not yet implemented".into())
```
Dinnerbone, line 65: Can we stub this and return ""?

core/src/avm2/globals/xml_list.rs
```diff
@@ -1,4 +1,77 @@
...
+
+    match E4XNode::parse(value, activation) {
+        Ok(nodes) => {
+            this.set_children(
+                activation.context.gc_context,
+                nodes.into_iter().map(E4XOrXml::E4X).collect(),
+            );
+        }
+        Err(e) => {
+            return Err(Error::RustError(
```
Dinnerbone, line 36: We should get into the habit of throwing avm errors where applicable now - there's definitely content that try/catches this

core/src/avm2/globals/xml.rs
```diff
@@ -1,3 +1,81 @@
...
+                        "XML constructor must be called with a single node: found {:?}",
+                        nodes
+                    )
+                    .into(),
+                ));
+            }
+            this.set_node(activation.context.gc_context, nodes[0])
+        }
+        Err(e) => {
+            return Err(Error::RustError(
```
Dinnerbone, line 30: Same here

core/src/avm2/globals/flash/utils.as
```diff
@@ -8,14 +8,7 @@ package flash.utils {
 	// made specifically to fool com.adobe.serialization.json.JsonEncoder.
```
Dinnerbone, line 8: The comment block here is no longer applicable

core/src/avm2/error.rs
```diff
@@ -239,3 +239,9 @@ impl<'gc> From<ruffle_render::error::Error> for Error<'gc> {
         Error::RustError(val.into())
     }
 }
+
+impl<'gc> From<quick_xml::Error> for Error<'gc> {
```
Dinnerbone, line 243: //TODO this?

Dinnerbone (changes requested):

core/src/avm2/globals/flash/utils.as
adrian17: It's mostly still applicable, but should be moved to the .rs file.

core/src/avm2/globals/xml_list.rs
Aaron1011: Given how incomplete the implementation is, I'd prefer not to add any stubs for now. Most SWFs that use XML will still have errors with this PR, and creating a `RustError` will make it easier to debug them.

Dinnerbone (approved):

Bale001 (approved):

adrian17 (approved):

core/src/avm2/globals/XML.as
```diff
@@ -1,4 +1,42 @@
...
+        
+        AS3 native function toString():String;
+    }
+}
+
+XML.prototype.name = function():Object {
```
relrelb, line 18: I think the following definitions can be moved inside the `class` body, like this:
```as
public final dynamic class XML {
    ...

    prototype.name = function(): Object {
        ...
    }
}
```

Aaron1011: Fixed

relrelb (approved):

PR #9648, opened 2023-02-20
core: Move `Stub` and stub macros to `ruffle_render`

Aaron1011:
This lets us `avm2_stub_` macros from render backends, which is useful for large feature like Stage3D. Since `linkme::distributed_slice` relies on the linker, we can collect stubs from across different crates.

The `avm2_stub_` macros now take in the `StubCollection` as their first argument, instead of an `activation`. This makes invoking them in `core` slightly more verbose, but lets us use the same macro in render backends.

We now store an `Rc<StubCollection>` in each render backend impl. The player calls into the render backend to obtain an `Rc<StubCollection>` during initialization.


Dinnerbone:
Not sure I agree with this one, sorry. I like the idea of using stubs elsewhere but they should be their own category, not an avm1 or avm2 stub - those should be limited to `core/avm1/` and `core/avm2/`, the actual API itself. Rendering things could be their own concept, or a general string "whatever" stub maybe.

For the issue of accessing the stub stuff from within render backends: there shouldn't be anything stopping us from depending on `core` from each individual backend, there's no circular dependency there.

Aaron1011:
There's no longer a pressing need for this

PR #9652, opened 2023-02-20
web: Use `@tsconfig/strictest`

relrelb:


web/packages/core/src/polyfills.ts
```diff
@@ -68,14 +68,13 @@ function polyfillFrames(): void {
     iframes = iframes ?? document.getElementsByTagName("iframe");
     frames = frames ?? document.getElementsByTagName("frame");
 
-    [iframes, frames].forEach((elementsList) => {
-        for (let i = 0; i < elementsList.length; i++) {
-            const element = elementsList[i];
-            if (element.dataset.rufflePolyfilled !== undefined) {
+    [iframes, frames].forEach((elements) => {
+        for (const element of elements) {
+            if (element.dataset["rufflePolyfilled"] !== undefined) {
```
adrian17, line 73: I thought `dataset.rufflePolyfilled` is the more idiomatic way of writing this?

web/packages/core/src/public-api.ts
```diff
@@ -72,10 +72,10 @@ export class PublicAPI {
                 prev.superseded();
             } else if (
                 prev.constructor === Object &&
-                prev.config instanceof Object
+                prev["config"] instanceof Object
             ) {
                 /// We're the first, install user configuration
-                this.config = prev.config;
+                this.config = prev["config"];
```
adrian17, line 78: let's be consistent and call it `this["config"]`

web/packages/core/src/ruffle-embed.ts
```diff
@@ -50,9 +50,11 @@ export class RuffleEmbed extends RufflePlayer {
                     allowScriptAccess,
                     src.value
                 ),
-                parameters: this.attributes.getNamedItem("flashvars")?.value,
-                backgroundColor: this.attributes.getNamedItem("bgcolor")?.value,
-                base: this.attributes.getNamedItem("base")?.value,
+                parameters:
+                    this.attributes.getNamedItem("flashvars")?.value ?? null,
```
adrian17, line 54: why is this needed? Isn't `Attr.value` typed as just a `string`?

web/packages/core/src/plugin-polyfill.ts
```diff
@@ -168,8 +152,7 @@ class RufflePluginArray {
 export const FLASH_PLUGIN = new RufflePlugin(
     "Shockwave Flash",
     "Shockwave Flash 32.0 r0",
-    "ruffle.js",
-    null
```
adrian17, line 172: No reason to have the null not be explicit anymore

web/packages/core/src/polyfills.ts
relrelb: `@tsconfig/strictest` turns on [`noPropertyAccessFromIndexSignature`](https://www.typescriptlang.org/tsconfig#noPropertyAccessFromIndexSignature), which bans "indexed" properties from being accessed via the dot operator.

web/packages/core/src/public-api.ts
relrelb: These are different - `this.config` is a class field while `prev` is an object defined on `window` (i.e. `window.RufflePlayer.config = { ... }`). Due to [noPropertyAccessFromIndexSignature](https://www.typescriptlang.org/tsconfig#noPropertyAccessFromIndexSignature), the latter must be `prev["config"]`, but I think we should write `this.config` where allowed.

web/packages/core/src/ruffle-embed.ts
relrelb: Yeah, but the optional chaning operator (`?.`) causes it to be `string | undefined`. `undefined` can happen if `getNamedItem("flashvars")` returns `null`.

web/packages/core/src/plugin-polyfill.ts
relrelb: Per discussion on Discord, seems like the `mimeTypes` parameter is unsued since https://github.com/ruffle-rs/ruffle/commit/5ff7168fe9c2dbe8988a1e7b777eeec8d3da04a8#diff-d8d3724115abb346d049c122ef9ff923fb69e5be9e9f7237896d86d955e575fcL135-L156, and now it's not even usable at all.
So I went ahead and removed it completely.

PR #9656, opened 2023-02-20
Remove ^M characters from Dictionary test

Lord-McSweeney:
Using Github's line ending conversion to CRLF. Github's online editor does this automatically, but it also shows the whole file as changed, even though the only manual change made was a single newline added in between two traces.


n0samu:
Can you please also fix the `dictionary_delete` and `dictionary_in` tests? They have the same problem.

relrelb (approved):
LGTM. I wonder if we can check for CRLF in CI, so future PRs will be rejected.

PR #9657, opened 2023-02-21
avm2: Use a lookup table for cosine in `SoundMixer.computeSpectrum()`

torokati44:
This results in 128 times fewer calls to the builtin `cosf` function.

While this is still the most naive implementation with two simple nested loops (so no recursive Cooley-Tukey-style `N*log(N)` complexity[^1]), and it does not even seem to autovectorize much; the main bottleneck was the `cos` computation, not the number of multiplications.

The recently added unit test still passes of course, as there was no meaningful change made to the main algorithm itself.

The improved performance can be seen best on web, with these loops:
 - https://z0r.de/L/z0r-de_7812.swf
 - https://z0r.de/L/z0r-de_7855.swf
 
For example, z0r 7812 before:
image

And after:
image

[^1]: https://www.youtube.com/watch?v=nmgFG7PUHfo

### DIFF
--- a/core/src/avm2/globals/flash/media/soundmixer.rs
+++ b/core/src/avm2/globals/flash/media/soundmixer.rs
@@ -155,27 +155,26 @@ pub fn compute_spectrum<'gc>(
             x.abs().ln().max(0.0) / 4.0
         }
 
-        // Need to make a copy of the samples used by the FFT, so they aren't
-        // modified in place.
-        let mut inp = [[0.0; 2]; 512];
-        inp.copy_from_slice(&hist[..512]);
-
-        // We could stop earlier here, depending on the value of 'stretch'.
-        for (freq, h) in hist.iter_mut().take(512).enumerate() {
-            let mut sum_left = 0.0;
-            let mut sum_right = 0.0;
-
-            for (i, sample) in inp.iter().enumerate() {
-                let freq = freq as f32;
-                let i = i as f32;
-                let coeff = (std::f32::consts::PI * freq * i / 1024.0).cos();
-
-                sum_left += sample[0] * coeff;
-                sum_right += sample[1] * coeff;
-            }
-
-            *h = [postproc(sum_left), postproc(sum_right)];
+        // Precompute a single period of the cosine function to be used as a lookup table.
+        let mut cos_lut = [0.0f32; 2048];
+        for (i, c) in cos_lut.iter_mut().enumerate() {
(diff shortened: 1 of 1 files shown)


core/src/avm2/globals/flash/media/soundmixer.rs
```diff
@@ -155,27 +155,26 @@ pub fn compute_spectrum<'gc>(
...
-            for (i, sample) in inp.iter().enumerate() {
-                let freq = freq as f32;
-                let i = i as f32;
-                let coeff = (std::f32::consts::PI * freq * i / 1024.0).cos();
-
-                sum_left += sample[0] * coeff;
-                sum_right += sample[1] * coeff;
-            }
+        // Precompute a single period of the cosine function to be used as a lookup table.
+        let mut cos_lut = [0.0f32; 2048];
```
adrian17, line 159: just to be sure, do we wanna have all these arrays to be on stack?

adrian17 (approved):

core/src/avm2/globals/flash/media/soundmixer.rs
torokati44: I did try to make the LUT:
 - An embedded constant, but:
   - I couldn't find a way to compute it at compile time with a `const fn` (loops not allowed and IIRC `cos` is not `const`).
   - I didn't want to embed it a big pile of literals just due to cosmetics.
   - Also this would have increased the binary size for no good reason.
 - A reused lazily computed global, but:
   - Rust is not very good with mutable global state without trickery.

All of these mostly to avoid any `cos` calls in the actual invocations (or at least in the after the first).
 
And for all the arrays, not just the LUT: I don't know, does it matter that much? Does it have a particular penalty on WASM? I personally didn't notice anything like that.

adrian17: > does it matter that much

Generally big arrays on stack (4b*2048 is not extreme, but also not small anymore) always increase the risk of accidental stack overflows in the long run. But I might be overreacting :)

torokati44: Oh, I see. Well, that is a valid concern, but I also think that these couple of KiBs shouldn't be that critical. Let's hope we're right... :)

torokati44:
Thank you for the review and merge @adrian17!

torokati44:
For future reference, I received a very nice (and accurate!) tip on how to substitute the two nested loops with a proper FFT:
https://github.com/ejmahler/rust_dct/issues/13#issuecomment-1465340496

PR #9659, opened 2023-02-21
render: Fix blending onto BitmapData contents

Aaron1011:
When rendering to an offscreen texture for `Bitmapdata.draw`, we first render to a temporary frame buffer, and then copy the contents of the frame buffer back to the target texture. However, this results in blend modes being incorrectly applied - for example, rendering with BlendMode.SUBTRACT will subtract against the framebuffer (which starts with each pixel as 0x00000000), instead of the previous BitmapData contents.

To fix this, we now use our texture target as the frame buffer
when performing `render_offscreen`. This ensure that we blend
over existing pixels (taking into account the `blendMode` provided
in the `BitmapData.draw` call).

When multisampling is enabled, we use a copy pipeline to copy
the existing contents of our texture to a fresh multisampled frame
buffer (the non-multisampled texture target becomes our resolve buffer).


SN902:
In this commit most blending parts broken (wind, rainbow, stars, water splashes)
[Rd_flying.zip](https://github.com/ruffle-rs/ruffle/files/10793779/Rd_flying.zip)

Aaron1011:
@SN902 I've reimplemented the logic - RD_Flying is now working for me locally

Lord-McSweeney:
Fixes #3086 fully, fixes the regression in #1839.

SN902:
@Aaron1011 all regressions is gone, i approve this commit. :)

Dinnerbone (approved):

PR #9660, opened 2023-02-21
avm2: Allow resolving interfaces before ClassObjects are available

Aaron1011:
The Adobe Animate compiler can emit a 'newclass' opcode for a concrete class before the 'newclass' opcodes for the interfaces it implements. As a result, we cannot rely on looking up an interface `ClassObject` when resolving a class's interfaces.

We now use `Domain.get_defining_script` to find the interface's `Script`, and iterate over the defined traits to find the `Trait` containing the `Class<'gc>`, which is what we actually need in `link_interfaces`.

Additionally, `link_interfaces` was failing to consider superinterfaces, which meant that methods from superinterfaces were not being copied into the vtable. I've fixed this along with the other changes.


Aaron1011:
We could introduce a `HashMap` to speed up 'early interface resolution', but I'm not sure if it's worth the additional memory usage (especially on wasm).

core/src/avm2/object/class_object.rs
```diff
@@ -368,6 +352,45 @@ impl<'gc> ClassObject<'gc> {
         Ok(())
     }
 
+    // Looks up an interface by name, without using `ScopeChain.resolve`
+    // This lets us look up an interface before its `ClassObject` has been constructed,
+    // which is needed to resolve interfaces when constructing a (different) `ClassObject`.
+    fn early_resolve_interface(
+        &self,
+        scope: ScopeChain<'gc>,
```
Bale001, line 360: Does this method need to be taking `scope` as an argument? It might be simpler for the method to get the class scope itself

Aaron1011: I've changed this to directly take in a `Domain`

adrian17 (approved):

PR #9661, opened 2023-02-21
core: Produce an error when BitmapData.draw call is unsupported

Aaron1011:
If our render backend doesn't support BitmapData.draw, then we shouldn't try to continue execution normally (in AVM2). Many SWFs rely on inspecting the resulting `BitmapData` to implement game logic, which will result in strange bugs if the draw wasn't actually performed.

### DIFF
--- a/core/src/bitmap/bitmap_data.rs
+++ b/core/src/bitmap/bitmap_data.rs
@@ -47,6 +47,11 @@ impl LehmerRng {
 #[collect(no_drop)]
 pub struct Color(i32);
 
+#[derive(Debug, Clone)]
+pub enum BitmapDataDrawError {
+    Unimplemented,
+}
+
 impl Color {
     pub fn blue(&self) -> u8 {
         (self.0 & 0xFF) as u8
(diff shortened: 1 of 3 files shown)


core/src/bitmap/bitmap_data.rs
```diff
@@ -1350,7 +1351,7 @@ impl<'gc> BitmapData<'gc> {
         blend_mode: BlendMode,
         clip_rect: Option<Rectangle<Twips>>,
         context: &mut UpdateContext<'_, 'gc>,
-    ) {
+    ) -> bool {
```
relrelb, line 1354: Wouldn't it be more appropriate to return a `Result` here?

core/src/avm2/globals/flash/system/Capabilities.as
```diff
@@ -9,5 +9,11 @@ package flash.system {
         public static function get version(): String {
             return "LNX 32,0,0,465"
         }
+        public static function get manufacturer(): String {
```
danielhjacobs, line 12: This file should not be included.

danielhjacobs: Your commit also has the wrong name.

Dinnerbone (approved):

Issue #6611, opened 2022-04-03
audio: An unexpected sound effect is played in Dexter's Laboratory - Web build only


Toad06:
Game: [ZIP](https://github.com/ruffle-rs/ruffle/files/8088528/dexter.zip) | [URL](https://6games.eu/games/10556-dexter-laboratory-when-stuffed-animals-attack.html)

---

When you complete a level in Dexter's Laboratory, points are added to the total score and a sound effect is played; this sound effect unexpectedly differs depending on the backend.

- Sound effect played in Flash Player and Ruffle Desktop:

https://user-images.githubusercontent.com/14274782/161447974-b8787e87-4185-4f66-ac2f-6a6c50a56fa2.mp4

- Sound effect played in Ruffle Web (incorrect):

https://user-images.githubusercontent.com/14274782/161447984-347edc87-b8b3-4498-947a-7d1310cfc68e.mp4

---

The issue occurs in Firefox and Chrome (Windows 10). Firefox logs the following warning into the console:

`Unexpected JavaScript error: JsValue(RangeError: AudioScheduledSourceNode.stop: The value for the stop time is outside the valid range.`
<details>
  <summary>See the full stack trace</summary>

```
966/e/c.wbg.__wbg_stop_c6532de1aeb2e765/<@moz-extension://abbac6ff-9469-462c-b376-24f9f0ad4a66/dist/966.js:1:19469
D@moz-extension://abbac6ff-9469-462c-b376-24f9f0ad4a66/dist/966.js:1:4037
966/e/c.wbg.__wbg_stop_c6532de1aeb2e765@moz-extension://abbac6ff-9469-462c-b376-24f9f0ad4a66/dist/966.js:1:19447
web_sys::features::gen_AudioBufferSourceNode::AudioBufferSourceNode::stop_with_when::hea6dfe877094f272@moz-extension://abbac6ff-9469-462c-b376-24f9f0ad4a66/dist/9d5fcf4ce4df1e4d4c96.wasm:wasm-function[7194]:0x58b606
ruffle_web::audio::WebAudioBackend::start_sound_internal::hb2458ee5add8c47e@moz-extension://abbac6ff-9469-462c-b376-24f9f0ad4a66/dist/9d5fcf4ce4df1e4d4c96.wasm:wasm-function[420]:0x14799b
[...]

Issue #1673, opened 2020-11-23
No Background Music on Sonic Shorts Vol. 1


digamejh:
Web page: https://www.y8.com/animation/sonic_shorts_volume_1
Flash file: https://img-hws.y8.com/cloud/y8-flash-animation/contents/item_versions/flash_animations/1519/original/sonic_shorts_volume_1.swf

Thank you!

Toad06:
Thanks for this bug report!

When clicking "Play Shorts" [`DefineButton2 (187)`], two distinct code blocks are executed in this order:

```
on(release){
   gotoAndPlay(111);
}
```

```
on(release){
   stopAllSounds();
}
```
Ruffle stops the audio from the title screen (as does the Flash Player), but it also stops the audio from the short (as we're already in frame 111 - and this frame has a `SoundStreamBlock`).

Reversing the order (or even removing the `SoundStreamBlock` in frame 111) fixes the issue in Ruffle.

Herschel:
Even after stopAllSounds, I'd expect the audio stream to start back up again from the next SoundStreamBlock, will have to investigate why this doesn't happen.

Toad06:
Audio is working in master. I assume this is fixed by #1719.
I'll let @Herschel confirm that.

digamejh:
Yep that worked, thank you!

Herschel:
Seems fixed, thanks for reporting!

digamejh:
However...

I just discovered a bug where after the video loads, if you click on "Scene Select" and pick a short, there is still no background audio!

Herschel:
Reopning :-)

Toad06:
Hmm, can't reproduce, audio also plays as expected when the short is started from the "Scene select" section.

There's however an issue on the web backend only: when selecting a short, the white noise sound effect is played, sometimes at really unexpected moments (you can check with the last short on the second page).

digamejh:
For reproducing the problem, try playing the "Chicken Boo" short (last one on the first page).

Toad06:
Fixed by #4273.

PR #9671, opened 2023-02-22
web/audio: Dynamic buffer size

torokati44:
~This is not done yet, just pushing for visibility.~

The goal is to:
 - Make demanding content (especially on mobile) as stutter-free as possible
 - Reduce latency on lighter content (or on powerful hardware) as much as possible

### DIFF
--- a/web/src/audio.rs
+++ b/web/src/audio.rs
@@ -16,14 +16,32 @@ use web_sys::AudioContext;
 pub struct WebAudioBackend {
     mixer: AudioMixer,
     context: AudioContext,
+    /// The current length of both buffers, in frames (pairs of left/right samples).
+    buffer_size: Arc<RwLock<u32>>,
     buffers: Vec<Arc<RwLock<Buffer>>>,
+    /// When the last submitted buffer is expected to play out completely, in seconds.
     time: Arc<RwLock<f64>>,
-    position_resolution: Duration,
+    /// How many consecutive times we have filled the next buffer "at a sufficiently early time".
+    num_quick_fills: Arc<RwLock<u32>>,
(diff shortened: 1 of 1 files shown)


torokati44:
Thank you for the suggestions @ousia! Once I get something that actually works, I might test those too!

torokati44:
At least in my limited testing, this now appears to do what it's supposed to:
 - z0r.de/690 no longer stutters sometimes
 - z0r.de/6268 now feels more "responsive" to music

More investigation is needed though, to make sure.

Dinnerbone (approved):
Looks good to me but I'm definitely not an expert in this area :D

torokati44:
Thanks for checking it out! And, well, neither am I! 😅

torokati44:
Code is one thing, but it'd be great if this could get some more testing, to make sure it doesn't cause any regressions by being too opportunistic with short buffers.

@Toad06 or @n0samu, could either of you please try it out on some of your content, if you have the time of course?

Toad06:
Forgot to mention it but I tried the PR right after it was no longer a draft. 

I tested many files (including #6611 and #1673 which had issues only affecting the web build) and haven't found any regression. So this looks good to me.  😉

torokati44:
That's great to hear, thank you! ☺️

n0samu:
I just did a few quick tests and this seems to be working exactly as intended! In Bloons 2, which previously had very choppy audio, now the audio is delayed but mostly smooth, except in the very most demanding sections. And games that ran smoothly seem unchanged, with no additional audio delay introduced. Nicely done! 😃

torokati44:
Thanks to you too, and also good to hear! ^^

PR #9680, opened 2023-02-23
avm2: Convert lots of classes to ActionScript

Aaron1011:
The vast majority of the changes are purely mechanical - defining the methods in `.as` files, renaming getters to use the `get_` prefix, and deleting the Rust-side class definition logic.

Non-trivial changes:
* We don't have support for native constructors, so classes that need them define a private native method called `init`, which we call from the ActionScript constructor.
* Instead of `ruffle_private_namespace`, we now use the package-specific `internal` namespace to access `internal` fields defined in Actionscript. `ruffle_private_namespace` had no remaining uses, so I removed it.
* A few methods were incorrectly checking if a particular argument was *present* in the args array, instead of non-null. With methods defined in ActionScript, our Activation code will ensure that the `args` slice always contains the correct number of arguments (with defaulted parameters filled in). Those methods were changed to explicitly check for `Value::Null`


core/src/avm2/globals/flash/display/BitmapData.as
```diff
@@ -1,5 +1,56 @@
...
+    import flash.geom.Rectangle;
+    import flash.geom.ColorTransform;
+    import flash.geom.Point;
+    import flash.geom.Matrix;
+    import flash.filters.BitmapFilter;
+    import flash.utils.ByteArray;
+
+    [Ruffle(InstanceAllocator)]
+    public class BitmapData implements IBitmapDrawable {
+        // FIXME - the first two arguments should not be defaulted, but it's currently
```
Lord-McSweeney, line 12: Couldn't this be fixed by manually setting width and height to 0 based on arguments.length in the ActionScript part of the constructor?

core/src/avm2/globals/flash/display/LoaderInfo.as
```diff
@@ -1,3 +1,35 @@
...
+	import flash.events.UncaughtErrorEvents;
+
+	
+	[Ruffle(InstanceAllocator)]
+	[Ruffle(NativeInstanceInit)]
+	public class LoaderInfo extends EventDispatcher {
+		public function LoaderInfo() {
+			throw new Error("LoaderInfo cannot be constructed")
+		}
+
```
Lord-McSweeney, line 14: nit: fix the indentation level for the rest of this class.

core/src/avm2/globals/flash/display/MovieClip.as
```diff
@@ -0,0 +1,27 @@
+package flash.display {
+    public dynamic class MovieClip extends Sprite {
+        public function MovieClip() {}
```
Lord-McSweeney, line 3: Is movieclip allowed to be constructed normally?

core/src/avm2/globals/flash/display/DisplayObject.as
```diff
@@ -1,5 +1,113 @@
...
-	public class DisplayObject {
-	}
+
+    import flash.geom.Rectangle;
+    import flash.geom.Transform;
+    import flash.display.DisplayObject;
+    import flash.display.DisplayObjectContainer;
+    import flash.display.LoaderInfo;
+    import flash.display.Stage;
+	import flash.geom.Point;
```
Lord-McSweeney, line 9: nit: fix the indentation level for the line

core/src/avm2/globals/flash/display/MovieClip.as
Aaron1011: Yes: https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/display/MovieClip.html#MovieClip()

core/src/avm2/globals/flash/display/BitmapData.as
Aaron1011: This PR is already quite large, so I wanted to do it separately.

Aaron1011: The real fix for this to create a native init method for `BitmapData`, and call that in `BitmapDataObject::from_bitmap_data`

core/src/avm2/globals.rs
```diff
@@ -478,96 +478,37 @@ pub fn load_player_globals<'gc>(
...
+        flash::display::graphics::create_class(activation),
         script
     );
-    class(
-        flash::display::interactiveobject::create_class(activation),
-        script,
-        activation,
-    )?;
+
+    // package `flash.geom`
```
Lord-McSweeney, line 489: nit: this comment can be removed

core/src/avm2/globals/flash/display/DisplayObject.as
```diff
@@ -1,5 +1,113 @@
...
+        public native function get mask():DisplayObject;
+        public native function set mask(value:DisplayObject):void;
+
+        public native function get opaqueBackground():Object;
+        public native function set opaqueBackground(value:Object):void;
+
+        public native function get cacheAsBitmap():Boolean;
+        public native function set cacheAsBitmap(value:Boolean):void;
+
+		public native function hitTestPoint(x:Number, y:Number, shapeFlag:Boolean = false):Boolean;
```
Lord-McSweeney, line 98: nit: fix the indentation level

relrelb:
@Aaron1011 Impressive work! But I kind of averse to such single massive commits. They are nearly impossible to review, and make it hard to blame/bisect changes.
Can you please split these changes to multiple commits? If needed, I'm here to help.

Aaron1011:
@relrelb I've split up the changes into smaller commits

relrelb:
nit: Seems like the removal of `ieventdispatcher.rs` happens on a wrong commit, not along with the addition of `IEventDispatcher.as`.

core/src/avm2/globals/flash/events/EventDispatcher.as
```diff
@@ -1,5 +1,19 @@
 // This is a stub - the actual class is defined in `eventdispatcher.rs`
 package flash.events {
-	public class EventDispatcher {
+	public class EventDispatcher implements IEventDispatcher {
```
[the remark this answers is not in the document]
Aaron1011: I think this is simpler - I'd prefer to not add another `Object` variant unnecessarily.

core/src/avm2/globals/globals.as
```diff
@@ -24,6 +24,19 @@ include "flash/utils/ByteArray.as"
 include "flash/utils/Dictionary.as"
 include "flash/desktop/ClipboardFormats.as"
 include "flash/desktop/ClipboardTransferMode.as"
+
+include "flash/events/IEventDispatcher.as"
```
relrelb, line 28: Is this not alphabetically-sorted for a reason?
If it does, please explain it in a comment (what other classes depend on it).
If not, please move it next to the other `flash.events` classes.

core/src/avm2/globals/flash/display/BitmapData.as
adrian17: Site note: mentioned before on Discord, but I was wondering if maybe we should make `instance_allocator` be able to receive args (to be a bit more like C++ constructor), so we can avoid two-phase initialization in classes like BitmapData (and other DOs) or `dummy` stuff in the XML PR.

core/src/avm2/globals/flash/events/EventDispatcher.as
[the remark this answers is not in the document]
adrian17: I don't think this annotation can work - our instance allocator is only called on a single layer (unlike C++ constructors), so a `MovieClip` will only call a `stage_allocator` and not call EventDispatcher's allocator at all.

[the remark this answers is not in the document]
adrian17: If we wanted EventDispatcher fields to be native (we do), it requires refactoring the entire object system - let's talk about it on Discord, as there were already some proposals throws over the years.
(For smaller-but-still-nontrivial refactors, maybe having stage_allocator initialize the dispatcher could work and remove the need for lazy init?)

core/src/avm2/globals/globals.as
adrian17: > Is this not alphabetically-sorted for a reason?

Interface before the class using it.
We never had them fully ordered and it's common enough to have dependency ordering that I'm not sure if it requires a comment.

core/src/avm2/globals/flash/display/Stage.as
```diff
@@ -0,0 +1,113 @@
...
+        override public native function set mouseEnabled(value:Boolean):void;
+
+        override public native function get name():String;
+        override public native function set name(value:String):void;
+
+        override public native function set opaqueBackground(value:Object):void;
+
+        override public native function set rotation(value:Number):void;
+
+        public native function set scale9Grid(value:Rectangle):void;
```
Lord-McSweeney, line 43: This should be an override, I believe.

Lord-McSweeney: Actually, it should also be implemented (or at least stubbed) on `DisplayObject`, otherwise this will `VerifyError` if it is added as an override. Pure Rust-defined classes do not `VerifyError`, whih is why the problem did not show up previously. For correct behavior, though, this needs to be stubbed on `DisplayObject`. The same problem is present with `loaderInfo` and `accessibilityProperties`, which need to be marked as `override` and stubbed in `DisplayObject`.

Aaron1011: We should eventually implement them on DisplayObject, but it's not really related to this PR - trying to call on on DisplayObject will give you an error both before and after this PR.

core/src/avm2/globals/globals.as
danielhjacobs: I'd say it's less so that it doesn't need a comment because it's common, but more so that it doesn't need a comment because the comment at the top of the file explains this situation, so another one in each location where this occurs seems unnecessarily verbose:
https://github.com/ruffle-rs/ruffle/blob/559ec8356ddf909d2bc0ca6561d2fff334eb3151/core/src/avm2/globals/globals.as#L1-L2.

Lord-McSweeney: That comment mentions superclasses, not interface/class dependencies.

core/src/avm2/globals.rs
Aaron1011: This is pre-existing

core/src/avm2/globals/globals.as
Aaron1011: I've expanded the comment to mention interfaces

Aaron1011:
I've addressed all of the comments

adrian17 (approved):
Approved, just unsure about the error-related commit.
(and we might wanna still squash the PR if we want to get building commits on bisects)

Lord-McSweeney:
LGTM

relrelb (approved):

PR #9679, opened 2023-02-22
avm2: Use avm errors for RangeErrors inside DisplayObjectContainer

Dinnerbone:
This progresses Papa's games (fixes #9539).

The error code is correct but the error messages are not the same as Flash's, I opted to keep our more descriptive ones until we discover that the exact text is needed one day.

PR #9683, opened 2023-02-23
Load filters from swf tag, lots of filter tests, and correct filter to/from avm2 conversions

Dinnerbone:
This makes Papa's Hot Doggeria playable (after #9679), and lays lots of groundwork for getting normal filters working.


render/src/filters.rs
```diff
@@ -66,3 +167,288 @@ impl Default for ColorMatrixFilter {
...
+            map_point: (0, 0),
+            mode: DisplacementMapFilterMode::Wrap,
+            scale_x: 0.0,
+            scale_y: 0.0,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct DropShadowFilter {
```
relrelb, line 259: I wonder if we could use `swf::DropShadowFilter` etc. directly without an intermediate struct.

Dinnerbone: That was what I used originally but there's not a 1:1 between swf filter tags and actual flash supported filters.

swf/src/types/drop_shadow_filter.rs
```diff
@@ -27,6 +27,11 @@ impl DropShadowFilter {
     pub fn num_passes(&self) -> u8 {
         (self.flags & DropShadowFilterFlags::PASSES).bits()
     }
+
+    #[inline]
+    pub fn hide_object(&self) -> bool {
```
relrelb, line 32: Per SWF19 CompositeSource should always be 1. Did you encounter real SWFs where it is 0?
By the way personally I prefer exposing a direct getter, i.e. `composite_source() -> bool` without a negation. The rational is to avoid abstraction where possible.

Dinnerbone: > Per SWF19 CompositeSource should always be 1. Did you encounter real SWFs where it is 0? By the way personally I prefer exposing a direct getter, i.e. `composite_source() -> bool` without a negation. The rational is to avoid abstraction as possible.

This is the only time it's used, and it's "hide object" in the editor and ActionScript. Because it's "hide object", I figured we should stick to the inversion here and not cause the callers to always invert.

render/src/filters.rs
relrelb: All filter types in `render/src/filters.rs` in this PR has a SWF counterpart. Am I missing something?

Dinnerbone: ShaderFilter (currently unimplemented) and DisplacementMapFilter do not exist in swf form.

There's also states that exist in swf that are somewhat nonsense to the renderer, like bevel filter that's both inside and on top, or convolution filter with a Vec that doesn't match rows*cols.

Technically the swf structs are lossy vs actionscript made filters, with Fixed16s, but I don't know if that'll actually matter in principal though.

Aaron1011 (approved):

render/src/filters.rs
Dinnerbone: I've pushed a commit that does that. I've had to lower the accuracy of the tests significantly to account for the usage of Fixed8/Fixed16 instead of f32 now though.

swf/src/read.rs
```diff
@@ -2123,7 +2123,31 @@ impl<'a> Reader<'a> {
...
+            blur_x: self.read_fixed16()?,
+            blur_y: self.read_fixed16()?,
+            angle: self.read_fixed16()?,
+            distance: self.read_fixed16()?,
+            strength: self.read_fixed8()?,
+            flags: GradientFilterFlags::from_bits_truncate(self.read_u8()?),
+        })
+    }
+
+    fn read_gradient_glow_filter(&mut self) -> Result<GradientGlowFilter> {
```
relrelb, line 2137: `read_gradient_bevel_filter` and `read_gradient_glow_filter` are exactly the same. Why both are needed?

Dinnerbone: I had to split up GradientFilter into two different types to allow for different impls on the types (for example, GradientGlowFilter.to_avm2() gives a different thing to GradientBevelFilter.to_avm2())

core/src/avm2/filters.rs
```diff
@@ -301,16 +310,22 @@ impl FilterAvm2Ext for ConvolutionFilter {
         let preserve_alpha = object
             .get_public_property("preserveAlpha", activation)?
             .coerce_to_boolean();
-        matrix.resize((matrix_x * matrix_y) as usize, 0.0);
+        let mut flags = ConvolutionFilterFlags::empty();
+        if clamp {
```
relrelb, line 314: How about converting this to `flags.set(ConvolutionFilterFlags::CLAMP, clamp);` (here and in other places)?

Dinnerbone: Personally I don't feel like that reads as easy as "if this then add this", but if you're sure I can do that change. It's functionally equivalent.

swf/src/read.rs
relrelb: I see `FilterAvm2Ext` for the `Filter` enum + each `swf` filter type. So maybe we can just implement `from_avm2_object` and `as_avm2_object` for the `Filter` enum, without `FilterAvm2Ext` at all? I prefer to avoid this code duplication...

core/src/display_object.rs
```diff
@@ -312,11 +313,11 @@ impl<'gc> DisplayObjectBase<'gc> {
         self.name = name;
     }
 
-    fn filters(&self) -> Avm2ArrayStorage<'gc> {
-        Avm2ArrayStorage::from_storage(self.filters.iter().collect())
+    fn filters(&self) -> Vec<Filter> {
+        self.filters.clone()
```
relrelb, line 317: Can we avoid the `.clone()` by returning a `&[Filter]`?

core/src/avm2/filters.rs
relrelb: I see it cleaner, and it aligns with the rest of the codebase.

swf/src/read.rs
relrelb: Actually this would be unnecessary if we implemented the filters as native objects with getter/setter for each field. I start to feel like the object system in avm2 needs a refactor, since currently it's not so easy to add new native objects.

Dinnerbone: That would be... very complicated.

These types do not exist in avm2 land, they are not exclusive to avm2, and they're defined in the renderer which has no concept of avmness at all. An `Ext` trait is required to put anything on them at all that interacts with Avm2.

You can have custom Object types for them, of course, but it'd need to store a **clone** of the filter object and nothing more. The objects avm2 gets are snapshots at that moment, not the actual direct filter on the stage.

core/src/display_object.rs
Dinnerbone: Not trivially, everything in DisplayObject returns owned types because of the lifetime complexities involved in returning a ref to a gc'd value

swf/src/read.rs
relrelb: Okay, so we can keep `FilterAvm2Ext` but implement it just for `Filter`, leaving the conversion of each specific filter type to a free function - `bevel_filter_to_avm2`, `glow_filter_to_avm2` etc.

Regarding the native objects, we can store just a reference to the underlying `DisplayObject`, then make the getters/setters operate directly on the `DisplayObject`, so any filter updates will be reflected immediately. Alternatively, we can use interior mutability to allow both the `DisplayObject` and avm2 `Object` "own" the filters. Ideally I think native objects are the way to go, as this pattern is used very widely in avmplus.

core/src/display_object.rs
relrelb: Yeah I remember something like that. I think nonetheless a `Ref<'gc, Vec<Filter>>` or something can be returned, but let's keep this as-is.

PR #9691, opened 2023-02-23
chore: Improve Renovate config

relrelb:
* Move under `.github/` to reduce amount of top-level files.
* Convert to JSON5 to allow comments and trailing commans.
* Include `js-sys` and `web-sys` in `wasm-bindgen` group.
* Remove `tracing-tracy` disabling since it's already pinned in `desktop/Cargo.toml`.


Dinnerbone:
> Remove `tracing-tracy` disabling since it's already pinned in `desktop/Cargo.toml`.

It was pinned before we switched to Renovate, but that didn't stop it. It's goal is to update the version number regardless of if it's pinned.

relrelb:
> > Remove `tracing-tracy` disabling since it's already pinned in `desktop/Cargo.toml`.
> 
> It was pinned before we switched to Renovate, but that didn't stop it. It's goal is to update the version number regardless of if it's pinned.

Good to know. I restored it.

PR #9692, opened 2023-02-23
avm2: Ignore duplicate definitions

Bale001:
Some super obfuscated SWF's may have 2 of the same trait. Previously we would throw an error when this happened, but it turns out flash silently ignores this.


core/src/avm2/domain.rs
```diff
@@ -182,35 +182,27 @@ impl<'gc> Domain<'gc> {
...
         mc: MutationContext<'gc, '_>,
-    ) -> Result<(), Error<'gc>> {
+    ) {
         if self.has_definition(name) {
-            return Err(format!(
-                "VerifyError: Attempted to redefine existing name {}",
-                name.local_name()
-            )
-            .into());
+            return;
```
relrelb, line 193: Can you point out how this situation is handled in avmplus? I wonder whether the first defined trait remains or later traits override each other.

Bale001: I've added a test that proves it ignores the second definition

relrelb (approved):

PR #9697, opened 2023-02-23
web: Change `ruffle-core` output directory to `dist`

relrelb:
This aligns with the other packages.


Dinnerbone (approved):
Not tested but looks good!

PR #9700, opened 2023-02-24
wgpu: Fix blown out colors when transforming them above 100%

Dinnerbone:
The test swf has a background of #CCFF00, a box of #66FFFF and a color transform of (59% alpha, 100% rgb, +0 alpha, +255 rgb)

The transform tries to set the rgb to (in 0..1 space) `0.8, 2, 2` which we then multiply by the alpha but it's already blown out at that time, so it's incorrectly `(0.8, 1, 1)`. This clamps it after applying the delta but before the multiplication, so it can't blow out.


n0samu:
The problem also occurs in the WebGL renderer - if it's not too much trouble, could you fix it there as well?

Dinnerbone:
> The problem also occurs in the WebGL renderer - if it's not too much trouble, could you fix it there as well?

Fixed but untested

n0samu:
> > The problem also occurs in the WebGL renderer - if it's not too much trouble, could you fix it there as well?
> 
> Fixed but untested

That completely breaks the WebGL renderer:
`ERROR web\src\lib.rs:1351 Error creating WebGL renderer: Couldn't link shader program: Error linking shader program: Some("Must have a compiled fragment shader attached:\nSHADER_INFO_LOG:\nERROR: 0:131: 'clam' : no matching overloaded function found\nERROR: 0:131: '=' : dimension mismatch\nERROR: 0:131: 'assign' : cannot convert from 'const mediump float' to 'highp 4-component vector of float'\n")`

To test the WebGL renderer, you can set `webgl.enable-webgl2` to False in `about:config` (in Firefox).

Dinnerbone:
> That completely breaks the WebGL renderer

Sorry, typo, fixed now! There shouldn't be any difference between webgl1 and 2 here

n0samu:
Oh it's just a typo, should be `clamp` not `clam`.

n0samu:
Great, works perfectly now! 👍

n0samu (approved):

adrian17 (approved):

PR #9701, opened 2023-02-24
web: Use TypeScript project references

relrelb:
This slightly improves TypeScript build times, and allows LSP to work across packages without building `ruffle-core`.

Official documentation at:
https://www.typescriptlang.org/docs/handbook/project-references.html


Dinnerbone (approved):
Neat!

PR #9709, opened 2023-02-24
web: Cleanup `build` script in `package.json`

relrelb:
* Wrap comments to maximum 100 characters per line.
* Take advantage of npm pre & post scripts to logically separate between WebAssembly (pre)build, TypeScript build, and `set_version.js` postbuild. More info at: https://docs.npmjs.com/cli/v9/using-npm/scripts#pre--post-scripts
* Rename some scripts to improve clarity:
    * `build:ruffle_web` -> `build:wasm-vanilla`
    * `build:ruffle_web-wasm_extensions*` -> `build:wasm-extensions*`
    * `build:cargo_bindgen_opt` -> `build:wasm`
* Reduce script duplication by introducing the `$WASM_BINDGEN_FLAGS` and `$WASM_OPT_FLAGS` environment variables.

### DIFF
--- a/web/packages/core/package.json
+++ b/web/packages/core/package.json
@@ -10,7 +10,7 @@
         "dist/"
     ],
     "scripts": {
-        "build": "npm run build:wasm-vanilla && npm run build:wasm-extensions && npm run build:ts",
+        "prebuild": "npm run build:wasm-vanilla && npm run build:wasm-extensions",
 
         "//1": "# Setting ENABLE_WASM_EXTENSIONS=true causes a second module to be built as well, that utilizes WebAssembly extensions, instead of it just being a 'fake' - a copy of the 'vanilla' one.",
         "//2": "# Unfortunately, we have to set $RUSTFLAGS here, instead of in .cargo/config.toml (for example), because it's not possible to specify them per-profile; see cargo issue #7878.",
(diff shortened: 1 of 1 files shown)


torokati44:
Most of the changes look good!

I have one worry regarding:
> * Reduce script duplication by introducing the `$WASM_BINDGEN_FLAGS`
> and `$WASM_OPT_FLAGS` environment variables.

IIRC the duplication was done - instead of your way of setting envvars - to make the whole thing also work properly on Windows, in `CMD.EXE`. I think it's not quite trivial to set envvars for a single command there...
Can you perhaps test this?

relrelb:
> Most of the changes look good!
> 
> I have one worry regarding:
> 
> > * Reduce script duplication by introducing the `$WASM_BINDGEN_FLAGS`
> >   and `$WASM_OPT_FLAGS` environment variables.
> 
> IIRC the duplication was done - instead of your way of setting envvars - to make the whole thing also work properly on Windows, in `CMD.EXE`. I think it's not quite trivial to set envvars for a single command there... Can you perhaps test this?
> 
> **EDIT**: _PS: It would have been a bit easier to review if it was actually split into 2-3 separate commits doing the different changes, but I know it doesn't always work out that way, and I was able to manage it this way too._

Thanks for looking!
I tested this branch on Windows and `ENABLE_WASM_EXTENSIONS=true` seems to work - It correctly builds two WebAssembly modules. In fact, this technique is already used with `$CARGO_FLAGS` (though it's dedicated to CI).
Also, I've split this PR into 4 separate commits. Indeed it's much better like this.

torokati44:
Yeah, you're right. It's probably `cross-env` doing it then. So it should be fine.
Thanks for splitting it into multiple commits!

torokati44 (approved):

PR #9718, opened 2023-02-25
i18n: Update Japanese locale files

creeper-0910:
Japanese translation files have been updated.


relrelb (approved):
Thank you, and welcome!

Issue #9592, opened 2023-02-16
What's New Pussycat: Stuck on loading screen


n0samu:
### Describe the bug

In [What's New Pussycat?](https://www2.b3ta.com/catgame/memory.swf), Ruffle gets stuck on the loading screen.

This was originally reported by a user on Discord.

### Expected behavior

The game should load.

### Affected platform

Browser's extension

### Operating system

WIndows 10

### Browser

Firefox 110

### Additional information

Frame 3 of the game includes the following code:
`_root.inst_loader.loadVariables("game.php?time=" + now_string,"");`

And frame 1 places sprite 8 with name `inst_loader` and the following ClipActionRecord:
```as
onClipEvent(data){
    _root.global_dataloaded = true;
    _root.str_debug = "data loaded!";
}
```
In Ruffle, the `data` clipEvent is never triggered, so the loading never completes.

This seems like the same issue as #7301.

n0samu:
Fixed by #9720 (but also see #9722)

Issue #7301, opened 2022-06-27
Flash5 Movie silently not starting (first frame visible)


andremichelle:
### Describe the bug

Open http://basementmx.andre-michelle.com/
<img width="212" alt="image" src="https://user-images.githubusercontent.com/6459974/175899098-e69af11c-8f36-4c18-8b20-15e75a737b6d.png">
When looking at the network tab, I see that all swf have been loaded. Testing in Chrome. The self-hosted version does also not work correctly.

### Expected behavior

This button usually shows preloading and then acts as a "drop-right" menu with 8 audio icons. 

### Affected platform

Browser's extension

### Operating system

mac monterey

### Browser

Chrome 102

n0samu:
I haven't found the specific issue, but @Toad06 pointed out that this makes heavy use of `onClipEvent`, some features of which are unimplemented in Ruffle, so that may be the direction to look into.

Toad06:
Yes, I think the issue is due to `\frame 2\PlaceObject2 (3) Depth: 2 (loops)\onClipEvent(data)` in "mixer.swf": when the call to `loops.loadMovie(bank + "/loopset.swf");` succeeds, Ruffle should trigger this data event but this never happens. 

Depending on the feedbacks on #7311, I could try to make a fix.

andremichelle:
That would be great. I was hoping to present this flash movie at WAC2022 as part of Flash audio history.

n0samu:
If needed, there are definitely ways to get your site up and running for the demo using the original Flash Player! Feel free to join our Discord server or contact me at `nosamu#3385` on Discord for details.

andremichelle:
Thanks. I sent you a short message on discord.

Toad06:
So I tried to implement `onData` but realized there was another issue that I wasn't able to fix. I'm posting my observations below.

---

[Sample 1](https://github.com/ruffle-rs/ruffle/files/9039827/sample_1.zip)
In `\frame 2\PlaceObject2 (3) Depth: 2 (loops)`, there are two clip events: `data` and `enterFrame`. The latter one should fire continuously and trace "frame" but this doesn't happen in Ruffle.

[Sample 2](https://github.com/ruffle-rs/ruffle/files/9039834/sample_2.zip)
This is almost the same sample as above, but the order of the clip events in the tree has been reversed (so `enterFrame` comes before `data`). This time, Ruffle succeeds to fire the event and trace "frame".

---

[...]

PR #9720, opened 2023-02-25
swf: Don't truncate `ClipEventFlag` in SWFv5

### LINKED ISSUE #8620: swf: `read_clip_event_flags` truncates data for `ClipEventFlag::DATA` in SWFv5
_This issue is extracted from https://github.com/ruffle-rs/ruffle/issues/7301#issuecomment-1173851524 to get more visibility._

Mike investigated this issue on Discord and found out the following:

> Looks like the mistake is in "swf/src/read.rs". `read_clip_event_flags` mistakenly does `as u8` which truncates it to 1 byte in SWFv5 case. but the "data" bit should still be set in that second byte.

Removing `as u8` indeed prevents the issue from happening in the following samples.
But could this cause any downside with other SWFs?

---

_The part below is the original comment made in #7301._

[Sample 1](https://github.com/ruffle-rs/ruffle/files/9039827/sample_1.zip)
In `\frame 2\PlaceObject2 (3) Depth: 2 (loops)`, there are two clip events: `data` and `enterFrame`. The latter one should continuously fire and trace "frame" but this doesn't happen in Ruffle.

[Sample 2](https://github.com/ruffle-rs/ruffle/files/9039834/sample_2.zip)
This is almost the same sample as above, but the order of the clip events in the tree has been reversed (so `enterFrame` comes before `data`). This time, Ruffle succeeds to fire the event and trace "frame".

---

https://github.com/ruffle-rs/ruffle/blob/49e8f4e939c6b0e388a707c33c48ea397be72109/core/src/display_object/movie_clip.rs#L1134-L1137

After this code, I added the following line: `if id == 3 { log::error!("{:?}", clip_actions); }`

This logs:
- Sample 1: `[]`
- Sample 2: `[ClipAction { events: ENTER_FRAME, key_code: None, action_data: [136, 2, 0, 0, 0, 150, 7, 0, 0, 102, 114, 97, 109, 101, 0, 38, 0] }]` 

In both cases, the data event is missing.

relrelb:
Although in SWFv5 the 2nd byte is documented to be reserved (all 0), Flash does read it and treats it the same as for SWFv6.

Fixes #8620 (might also #7301, and #9592).

### DIFF
--- a/swf/src/read.rs
+++ b/swf/src/read.rs
@@ -2057,10 +2057,10 @@ impl<'a> Reader<'a> {
         let bits = if self.version >= 6 {
             self.read_u32().unwrap_or_default()
         } else {
-            // SWF19 pp. 48-50: For SWFv5, the ClipEventFlags only had 2 bytes of flags,
-            // with the 2nd byte reserved (all 0).
-            // This was expanded to 4 bytes in SWFv6.
-            (self.read_u16().unwrap_or_default() as u8).into()
+            // SWF19 pp. 48-50: For SWFv5, ClipEventFlags only had 2 bytes of flags. This was
+            // expanded to 4 bytes in SWFv6.
+            // The 2nd byte is documented to be reserved (all 0), but it's not enforced by Flash.
+            self.read_u16().unwrap_or_default().into()
         };
 
         ClipEventFlag::from_bits_truncate(bits)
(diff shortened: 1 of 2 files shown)


n0samu:
I can confirm that this fixes #9592, although the game still isn't working due to another (maybe related?) issue: #9722.

Unfortunately this does not seem to help with #7301. That needs investigation.

adrian17 (approved):

PR #9726, opened 2023-02-25
Avm2: fix and refactor some PrimitiveObject usages

adrian17:
leading example that sparked this:
```
var vec = Vector.<Object>([true, false]);
vec = Vector.<Boolean>(vec);
trace(vec); // "true,true" in Ruffle
```

First commit just fixes the bug.
Second commit makes us not create a PrimitiveObject when doing a coercion to the `Object` type, as it's usually not what we really need. If I have function `function f(x: Object)`, there's no need for `x` to be boxed. The intended use is to call `.coerce_to_object()` manually if you actually need the thing boxed right there.

Maybe we should rename `coerce_to_object` to `nullcheck_and_box_primitives` :)

No tests, but the upcoming TextFormat PR requires this PR to have its test pass.


Aaron1011:
@adrian17 Could you add that example you posted as a test case?

adrian17:
Sure, can do that later.
Just a note: that's not actually the bug I originally observed, I had to invent this one. The original bug was
```js
class TextFormat {
    public native function set bullet(param1: Object): void;
```
and the internal setter doing a `coerce_to_boolean()`

core/src/avm2/activation.rs
```diff
@@ -2069,11 +2069,10 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
 
     fn op_convert_o(&mut self) -> Result<FrameControl<'gc>, Error<'gc>> {
-        let value = self
-            .pop_stack()
-            .coerce_to_object(self)
-            .map_err(|_| "Cannot convert null or undefined to object")?;
-
+        let value = self.pop_stack();
+        if matches!(value, Value::Null | Value::Undefined) {
+            return Err(make_null_or_undefined_error(self, value, None));
```
Lord-McSweeney, line 2074: I might be incorrect, but is `make_null_or_undefined_error` the correct error? Shouldn't this one be error #1034?

adrian17: let me check, sec

adrian17: nope:
undefined -> convert_o:
```
TypeError: Error #1010: A term is undefined and has no properties.
	at global$init()
```
null -> convert_o:
```
TypeError: Error #1009: Cannot access a property or method of a null object reference.
	at global$init()
```

adrian17: It's confusing yeah, but avmplus indeed uses a direct equivalent of our `make_null_or_undefined_error` in the same spot

Aaron1011 (approved):

PR #9727, opened 2023-02-25
Avm2: Convert Graphics, TextFormat to AS

adrian17:
Needs https://github.com/ruffle-rs/ruffle/pull/9726 for tests to pass.


core/src/avm2/globals.rs
```diff
@@ -478,23 +478,6 @@ pub fn load_player_globals<'gc>(
...
-    avm2_system_class!(
-        graphics,
-        activation,
-        flash::display::graphics::create_class(activation),
-        script
-    );
-
-    // package `flash.geom`
-
-    // package `flash.text`
```
Lord-McSweeney, line 491: This comment still applies to `flash::text::font` (the last remaining Rust `flash.text` class), so I don't think it should be removed.

Aaron1011 (approved):

PR #9728, opened 2023-02-25
avm2: Store interfaces from  superclases/superinterfaces in ClassObject

Aaron1011:
This lets loop over a single array to check for implemented interfaces in `has_class_in_chain`. It also fixes a bug where `obj is Interface` incorrectly returned `false` if the `Inteface` implementation came from a superinterface.


core/src/avm2/object/class_object.rs
```diff
@@ -304,75 +308,70 @@ impl<'gc> ClassObject<'gc> {
...
+                    queue.push(interface);
+                    interfaces.push(interface);
+                }
+            }
+
+            if let Some(superclass_name) = cls.read().super_class_name() {
+                queue.push(self.early_resolve_class(scope.domain(), superclass_name)?);
+            }
         }
+        write.interfaces = interfaces.clone();
```
adrian17, line 338: can we avoid the clone here?

core/src/avm2/object/class_object.rs
```diff
@@ -304,75 +308,70 @@ impl<'gc> ClassObject<'gc> {
...
-                            activation.context.avm2.public_namespace,
-                            interface_trait.name().local_name(),
-                        );
-                        self.instance_vtable().copy_property_for_interface(
-                            activation.context.gc_context,
-                            public_name,
-                            interface_trait.name(),
-                        );
-                    }
+        for interface in interfaces {
```
adrian17, line 344: note: avmplus only does it for newly implemented interfaces (as the old ones are already properly copied in the vtable).
That said, I can't tell for sure if this can be observed. They _do_ have some slightly fancier logic that limits when interface namespace is copied; I can try look for some edge cases later.

core/src/avm2/object/class_object.rs
```diff
@@ -435,15 +434,20 @@ impl<'gc> ClassObject<'gc> {
     pub fn has_class_in_chain(self, test_class: ClassObject<'gc>) -> bool {
         let mut my_class = Some(self);
 
-        while let Some(class) = my_class {
-            if Object::ptr_eq(class, test_class) {
+        // A `ClassObject` stores all of the interfaces it implements,
+        // including those from superinterfaces and superclasses (recursively).
+        // Therefore, we only need to check interfaces once, and we can skip
+        // checking them when we processing superclasses in the `while`
+        // further down in this method.
```
adrian17, line 441: can we run this loop only if `test_class` is itself an interface?

Aaron1011: No - if `test_class` is a normal class, then this loop will check the implemented interfaces.

core/src/avm2/object/class_object.rs
Aaron1011: AVM2 code can only observe interfaces through `is/instanceof` and property lookups, right?

core/src/avm2/object/class_object.rs
Aaron1011: Fixed

core/src/avm2/object/class_object.rs
Aaron1011: I misread your comment - I'll add the check.

adrian17 (approved):

PR #9730, opened 2023-02-25
avm2: Implement encodeURI and encodeURIComponent

Aaron1011:


tests/tests/swfs/avm2/escape/Test.as
```diff
@@ -4,38 +4,44 @@
...
+	trace("// " + fnName + "(null)");
+	trace(fn(null));
+	trace("");
+
+	var input = "test";
+	trace("// " + fnName + "(\"" + input + "\")");
+	trace(fn(input));
+	trace("");
+
+	var input = "!\"£$%^&*()1234567890qwertyuiop[]asdfghjkl;'#\zxcvbnm,./QWERTYUIOP{}ASDFGHJKL:@~|ZXCVBNM<>?\u0010";
```
Lord-McSweeney, line 33: `;'#\zxcvbnm`
Shouldn't this be `;'#\\zxcvbnm`? This was present in the original test for `escape`

Aaron1011: I copied this from line 28 in the original diff - there's only a single backslash there.

Dinnerbone (approved):
No preference on the extra `\`, can fix that up whenever

PR #9733, opened 2023-02-26
avm2: Correctly check when toJSON is missing, and use vtable properties

Aaron1011:
We were previously calling `get_property` to determine if a `toJSON` property exists, but that produces an error if the method is missing on a sealed class.

Additionally, JSON serialization wasn't taking into account properties from the vtable. All public properties (including fields, const fields, and getter methods) get serialized.

Unfortunately, our vtable property order currently doesn't match Flash's. I've hand-edited the test output for now (all of the actual properties are there, just in a different order), and added a note


adrian17:
> FIXME - the order doesn't currently match Flash Player

I don't think we should care tbh, for two reasons: 1. order in JSON shouldn't matter anyway, 2. it appears that the order in FP is unspecified, probably just related to hashtable ordering (or maybe they skip the hashtable and just ). The only consistency I have found is that `get` properties all appear after `var` fields - but I don't think we need to worry about reproducing that either.

(now that I look at it, avmplus uses the same infrastructure (`TypeDescriber`) that `flash.utils.describeType()` will probably use in the future, so we could take that into account to reduce future duplicate code)

adrian17 (approved):

core/src/avm2/vtable.rs
```diff
@@ -473,6 +474,19 @@ impl<'gc> VTable<'gc> {
...
+
+        let mut props = Vec::new();
+
+        for (name, ns, prop) in read.resolved_traits.iter() {
+            if ns.is_public() {
+                props.push((name, *prop));
+            }
+        }
+        props
+    }
```
Bale001, line 489: Instead of creating a new vector here, maybe we could return an iterator that only yields public properties

Aaron1011: That's tricky, because we we need to access them through the `Ref`

Bale001: Hm yeah, I thought it might be possible through `Ref::map`, but it doesn't seem like it

PR #9740, opened 2023-02-26
core: Require `SwfMovie::url`

relrelb:


core/src/avm2/globals.rs
```diff
@@ -543,8 +543,8 @@ fn load_playerglobal<'gc>(
     activation.avm2().native_instance_allocator_table = native::NATIVE_INSTANCE_ALLOCATOR_TABLE;
     activation.avm2().native_instance_init_table = native::NATIVE_INSTANCE_INIT_TABLE;
 
-    let movie =
-        SwfMovie::from_data(PLAYERGLOBAL, None, None).expect("playerglobal.swf should be valid");
+    let movie = SwfMovie::from_data(PLAYERGLOBAL, "file:///".into(), None)
```
Dinnerbone, line 546: This is a little odd but I guess maybe it'll be fine? If anything tries to perform any operations on this url it might have a funky time with it, though I don't know if anything actually will.

Dinnerbone (approved):

core/src/avm2/globals.rs
relrelb: Agree, I just followed the pattern in `core/src/loader.rs`.

PR #9751, opened 2023-02-26
avm1: Add opcode 137

Lord-McSweeney:
According to the Open Flash documentation, this was added to inform the interpreter that runtime type-checking was done. It served no actual purpose in the interpreter, though, and the person who documented this for Open Flash could not find a compiler that generated this opcode.
```
This AVM1 action only serves as a hint that the byte code was compiled using "strict mode". It has no runtime effect.
```


Lord-McSweeney:
Closing, this hasn't been implemented correctly

PR #9753, opened 2023-02-27
video: Make DecodedFrame able to hold frames in different formats

torokati44:
This will be useful for skipping conversion to RGBA where not absolutely needed - or even doing it in a hardware accelerated way by the renderer.

### DIFF
--- a/render/wgpu/src/backend.rs
+++ b/render/wgpu/src/backend.rs
@@ -515,16 +515,17 @@ impl<T: RenderTarget + 'static> RenderBackend for WgpuRenderBackend<T> {
     fn update_texture(
         &mut self,
         handle: &BitmapHandle,
-        rgba: Vec<u8>,
+        bitmap: Bitmap,
         region: PixelRegion,
     ) -> Result<(), BitmapError> {
         let texture = as_texture(handle);
 
         let extent = wgpu::Extent3d {
-            width: region.width(),
-            height: region.height(),
+            width: bitmap.width(),
+            height: bitmap.height(),
             depth_or_array_layers: 1,
         };
+        let bitmap = bitmap.to_rgba();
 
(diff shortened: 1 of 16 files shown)


torokati44:
For now I've removed the change from this:

```rust
trait VideoBackend {
     fn decode_video_stream_frame(
         &mut self,
         stream: VideoStreamHandle,
         encoded_frame: EncodedFrame<'_>,
        renderer: &mut dyn RenderBackend,
    ) -> Result<BitmapInfo, Error>;
}
```

To this:

```rust
trait VideoBackend {
     fn decode_video_stream_frame(
         &mut self,
         stream: VideoStreamHandle,
         encoded_frame: EncodedFrame<'_>,
    ) -> Result<Bitmap, Error>;
}
```

That will only be needed to allow skipping the colorspace conversion and texture upload for frames that won't end up being actually displayed (during precise seeking or when playing catch-up with wall clock time).
Since that is one of the main points of this set of changes, I do plan to address it in a separate PR soon - it just got a bit too messy for now...

This in itself will already allow the video backends to take over the colorspace conversion in whatever way they see fit instead of relying on `Bitmap::to_rgb[a]()`.

Dinnerbone (approved):

Herschel (approved):
Thank you!

render/wgpu/src/backend.rs
```diff
@@ -537,8 +538,7 @@ impl<T: RenderTarget + 'static> RenderBackend for WgpuRenderBackend<T> {
                 },
                 aspect: wgpu::TextureAspect::All,
             },
-            &rgba[(region.y_min * texture.width * 4) as usize
-                ..(region.y_max * texture.width * 4) as usize],
+            bitmap.data(),
```
torokati44, line 541: Oops, this is wrong.

PR #9755, opened 2023-02-27
web: Fix broken extensions

Toad06:
[Nightly 2023-02-27 extensions](https://github.com/ruffle-rs/ruffle/releases/tag/nightly-2023-02-27) are broken and throw this error:

>Extension is invalid
>Loading locale file _locales/zh_TW/messages.json: SyntaxError: JSON.parse: expected double-quoted property name at line 65 column 1 of the JSON data


relrelb (approved):
Great catch, thanks!
I wonder how we could have detected this in CI - We probably need some kind of extension tests.

PR #9757, opened 2023-02-27
Fixed BitmapData accuracy when converting from premultiplied to straight rgb

Dinnerbone:
This progresses Kingdom Rush which does a "tamper check" by testing every pixel of their studio logo.

I've had to disable avm2/blend_transform test because now that the inaccuracy is fixed, we're actually revealing that the test was giving incorrect data. You can see that the checked-in expected.png is different RGB values to Flash Players, so it's likely that blending on BitmapData is subtly incorrect somehow.


core/src/bitmap/bitmap_data.rs
```diff
@@ -80,12 +80,56 @@ impl Color {
...
+        let r = unmultiply(self.red());
+        let g = unmultiply(self.green());
+        let b = unmultiply(self.blue());
+
+        Self::argb(self.alpha(), r, g, b)
+    }
+
+    #[must_use]
+    /// Unlike to_un_multiplied_alpha which trades accuracy for speed,
+    /// this method is slower but slightly more correct
```
Lord-McSweeney, line 126: Isn't this method less correct?

relrelb: But all that matters is being accurate to Flash, no? I think that if Flash actually does trade accuracy for speed, then so should Ruffle. What are the benefits of being accurate to the math?

Dinnerbone: We only use this in one method where I don't think Flash does the same unmult-change-premult dance. I don't think we should either but didn't want to tackle changing that in this PR.

Dinnerbone: I've inlined this method as it's only used one place, no naming issue there :D

PR #9760, opened 2023-02-27
avm2: Shrink `Op` enum by using a `Box<[i32]>` in `LookupSwitch`

Aaron1011:
This shaves off one `usize` from the size of `Op`.


swf/src/avm2/write.rs
```diff
@@ -915,7 +915,7 @@ impl<W: Write> Writer<W> {
                 self.write_opcode(OpCode::LookupSwitch)?;
                 self.write_i24(default_offset)?;
                 self.write_u30(case_offsets.len() as u32 - 1)?;
-                for offset in case_offsets {
+                for offset in case_offsets.iter() {
```
relrelb, line 918: I prefer the more clean `for offset in &case_offsets`

adrian17 (approved):

PR #9773, opened 2023-02-27
avm2: Add Telemetry class

Bale001:
Adds a fake Telemetry class to fool SWF's that use it (like crossbridge generated SWF's). This implementation mimics when Telemetry is not connected to any server

### DIFF
--- a/core/src/avm2/globals/flash/profiler/Telemetry.as
+++ b/core/src/avm2/globals/flash/profiler/Telemetry.as
@@ -1,0 +1,17 @@
+package flash.profiler {
+    public final class Telemetry {
+        public static const connected: Boolean = false;
+        public static const spanMarker: Number = 0;
+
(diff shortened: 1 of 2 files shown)


Dinnerbone (approved):

mero09-99:
So, no telemetry is actually created or sent?
where would the telemetry go normally?

Dinnerbone:
> So, no telemetry is actually created or sent?
> where would the telemetry go normally?

That's right. To be clear, this class is about performance telemetry and not, say, usage data. I believe it would be used in official flash player with something like Adobe Scout to assist with profiling your own application, as a developer. It's not for sending random data over the Internet or anything like that.

This implementation does absolutely nothing though and purely exists to make sure things don't throw errors of they tried to use it.

mero09-99:
Alright, thanks for clearing that up 👍

PR #9780, opened 2023-02-28
desktop: handle errors before detaching from console on windows cli

aria-aghaei:
When running ruffle through cmd.exe or Powershell on Windows, ruffle desktop detaches from console before printing any errors encountered during initialization: 
image

This PR changes error handling on main so the error is printed on cli before the console detaches:
image

Tested on Windows (cmd.exe and powershell) as well as MacOS.


aria-aghaei:
Going to make another PR, commits are messed up on this one

PR #9781, opened 2023-02-28
desktop: Handle errors before detaching from console on Windows cli

aria-aghaei:
When running ruffle through cmd.exe or Powershell on Windows, ruffle desktop detaches from console before printing any errors encountered during initialization:
image

This PR changes error handling on main so the error is printed on cli before the console detaches:
image

Tested on Windows (cmd.exe and powershell) as well as MacOS.

### DIFF
--- a/desktop/src/main.rs
+++ b/desktop/src/main.rs
@@ -1006,7 +1006,7 @@ fn shutdown() {
     }
 }
 
-fn main() -> Result<(), anyhow::Error> {
+fn main() -> Result<(), Error> {
     init();
     let opt = Opt::parse();
     let result = if opt.timedemo {
--- a/desktop/src/main.rs
+++ b/desktop/src/main.rs
@@ -1006,7 +1006,7 @@ fn shutdown() {
     }
 }
 
-fn main() {
+fn main() -> Result<(), anyhow::Error> {
     init();
     let opt = Opt::parse();
     let result = if opt.timedemo {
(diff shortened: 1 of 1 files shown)


desktop/src/main.rs
```diff
@@ -1006,14 +1006,18 @@ fn shutdown() {
...
     init();
     let opt = Opt::parse();
     let result = if opt.timedemo {
         run_timedemo(opt)
     } else {
         App::new(opt).map(|app| app.run())
     };
+    if let Err(ref error) = result {
+        eprintln!("\n\n{:?}", error);
+        shutdown();
```
relrelb, line 1019: I think the `shutdown(); std::process::exit(1);` can be removed as it'll anyway happen right after the `if` (returning an `Error` from `main` causes the process to exit with a non-zero code). This effectively leaves just (with some modifications):
```rs
if let Err(error) = &result {
    eprintln!("{:?}", error);
}
```
And the rest can remain unchanged.

aria-aghaei: Hi, since returning an anyhow::Error from main causes it to be printed, the error is printed twice on a non-windows OS. That was why I had set main to not return anything and just had it print once, but not sure if there is a better way to achieve the same result.

aria-aghaei: Just remembered there is a better way, cleaned up main and got rid of those unnecessary lines.

relrelb (approved):
Thank you! I've changed the `if cfg!(windows)` to `#[cfg(window)]` to be consistent with the rest of the code.

PR #3511, opened 2021-03-02
POC: Navigator: disallow requests to some domains (currently mochiads.com)

adrian17:
Requested on discord, no GH issue :(

mochiads.com was used by some flash games for distributing ads (and other uses?) - currently the domain was taken over by what looks like malware distributor.

PR design questions:
- disable flag in options? (needed for MVP?)
- list of banned domains in options? (needed for MVP?)
- what logging level (browsers usually log blocked requests by default)?
- currently the error message is duplicated, because the Err() variant contents are pretty much just ignored. Maybe some better way to write this?

PR #9788, opened 2023-02-28
core: Introduce compatibility rules, use it for kongregate sitelock

Dinnerbone:
This may be controversial but some may remember discussing this in the past, a few years ago.

The idea is that some content simply cannot run anymore, either because they've lost control of the domain or some assets were lost or whatever it may be. In this PR, we introduce a general CompatibilityRules struct that we'll, over time, add new forms of compatibility rules to. Currently that's just SWF URL rewriting, but it could plausibly rewrite other URLs (some lost API), change some functionality, bytecode replacement, I don't know. It's a thing.

I'd also like to make the rules loadable from a toml file going forwards, which would allow a hypothetical project like Flashpoint to do things in Ruffle without DNS replacements.

When the feature `default_compatibility_rules` is enabled (default for web & desktop), we'll include our own rules that we have vetted to the best of our ability to be correct - morally and technically. Currently this is Kongregate's sitelock replacement, as it doesn't seem like they'll be in a position to fix this any time soon. They own both domains in question, so this one feels quite safe to do.

All additions to the default ruleset should be scrutinized.


Bale001:
I agree with this, another domain that would be helpful to rewrite is `fpdownload.adobe.com`, which is used to fetch SWZ files. This domain is broken by CORS, so we should setup a mirror and redirect from there.

Dinnerbone:
> I agree with this, another domain that would be helpful to rewrite is `fpdownload.adobe.com`, which is used to fetch SWZ files. This domain is broken by CORS, so we should setup a mirror and redirect from there.

Good idea, we can absolutely do this!

danielhjacobs:
While not directly related, this reminds me of https://github.com/ruffle-rs/ruffle/pull/3511. I only bring it up since this seems like a good place to think about that again.

Dinnerbone:
> While not directly related, this reminds me of #3511. I only bring it up since this seems like a good place to think about that again.

Yeah I think we'll want to consider bringing that in as a rule too

Aaron1011 (approved):

PR #9789, opened 2023-02-28
avm2: Stub TextField.condenseWhite

MartySVK:
this progresses Armor Mayhem


adrian17:
I personally like implementing the backing storage for these properties - so that even though the logic is unimplemented, the script actually sees a consistent value (`tf.condenseWhite = true; trace(tf.condenseWhite)`).
But it's not a big deal and it's fine to just get it in as-is :)

adrian17 (approved):

PR #9793, opened 2023-03-01
avm2: Add `attribute` and `attributes` methods to `XML/XMLList`

Aaron1011:


core/src/avm2/globals/xml.rs
```diff
@@ -101,3 +101,66 @@ pub fn children<'gc>(
...
+pub fn attributes<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Option<Object<'gc>>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let this = this.unwrap();
+    let xml = this.as_xml_object().unwrap();
+    let attributes = if let E4XNodeKind::Element { attributes, .. } = &*xml.node().kind() {
+        // FIXME - avoid clone
+        attributes.clone()
```
adrian17, line 114: Should be doable if we just move `attributes.iter().map(|node| E4XOrXml::E4X(*node)).collect()` here, right?

adrian17 (approved):

PR #9794, opened 2023-03-01
avm2: Get XMLList 'filter' syntax working

Aaron1011:
This requires the ability to do a limited 'set_property', as well as `get_enumerant_value`.

To prevent modification of XMLLists derived from queries, I've introduced a `target` field on `XMLList`. This is `None` for lists created with `new XMLList()`, and `Some` when the list was derived from a query on an existing `XML` /`XMLList`. We only allow `set_property` when `target` is `None`: this is enough for filtering to work, and prevents silent incorrect execution when trying to modify an existing node.


Aaron1011:
This requires https://github.com/ruffle-rs/ruffle/pull/9791 for the tests to pass

core/src/avm2/object/xml_list_object.rs
```diff
@@ -189,11 +208,37 @@ impl<'gc> TObject<'gc> for XmlListObject<'gc> {
...
-        _value: Value<'gc>,
-        _activation: &mut Activation<'_, 'gc>,
+        name: &Multiname<'gc>,
+        value: Value<'gc>,
+        activation: &mut Activation<'_, 'gc>,
     ) -> Result<(), Error<'gc>> {
-        Err("Modifying an XMLList object is not yet implemented".into())
+        let mut write = self.0.write(activation.context.gc_context);
+        if write.target.is_some() {
+            panic!(
```
Lord-McSweeney, line 217: This should probably be an `Err`, not a panic.

core/src/avm2/object/xml_object.rs
```diff
@@ -178,4 +180,16 @@ impl<'gc> TObject<'gc> for XmlObject<'gc> {
...
         read.base.has_own_dynamic_property(name)
     }
+
+    fn set_property_local(
+        self,
+        name: &Multiname<'gc>,
+        value: Value<'gc>,
+        _activation: &mut Activation<'_, 'gc>,
+    ) -> Result<(), Error<'gc>> {
+        panic!(
```
Lord-McSweeney, line 190: Same here, this should be an `Err`

Dinnerbone (approved):

PR #9797, opened 2023-03-01
core: Set opt-level=3 in profile.test

Aaron1011:


torokati44:
Is this to try avoiding the timeouts on the RT test in your Stage3D PR?

PR #9802, opened 2023-03-01
web: Change preloader name while keeping old name for compatibility

danielhjacobs:
Setting either option to false should cause the splash screen to not appear. Feel free to test both names. Also now allowing `--preloader-background` or `--splash-screen-background` as custom properties.

Feel free to test. Seems to work from my limited testing. We can't invalidate the old option as multiple people may rely on it, but it can be considered deprecated.


danielhjacobs:
While I was in that part of the code I changed all related classes to kebab-case (preferred according to https://github.com/ruffle-rs/ruffle/pull/9461#discussion_r1117771460)

relrelb:
Thinking out loud: Should we log a depreaction warning when `preloader` is used? This way people can know to move to `splashScreen`, which will eventually allow us to remove it.
Also, is it possible to have a telemetry server that collects usage statistics, so we can remove rarely-used features with confidence (and maybe also track usage of unimplemented features, collects crash reports etc.)? Though I guess this has some unwanted privacy implications...

relrelb (approved):

web/packages/core/src/ruffle-player.ts
```diff
@@ -121,7 +121,7 @@ export class RufflePlayer extends HTMLElement {
     private container: HTMLElement;
     private playButton: HTMLElement;
     private unmuteOverlay: HTMLElement;
-    private preloader: HTMLElement;
+    private splashScreen: HTMLElement;
```
relrelb, line 124: While here: I think most (if not all) fields can be marked as `readonly`. Let's do it.

danielhjacobs:
> Should we log a depreaction warning when preloader is used

You mean a console log? We could, but I doubt many people will read it. Do you want me to add it anyway? Basically, I'd just check if preloader is set to false specifically around https://github.com/ruffle-rs/ruffle/pull/9802/files#diff-467a580c71d4e3b96032b54a2a292d485f3b83f87ba2478daf693bac6d532965R423 and log "The configuration option preloader has been replaced with splashScreen. If you own this website, please update the configuration."

> Also, is it possible to have a telemetry server that collects usage statistics

This sounds like a larger, but useful, discussion. I wouldn't want to tie that discussion to this PR personally though.

relrelb:
> > Should we log a depreaction warning when preloader is used
> 
> You mean a console log? We could, but I doubt many people will read it. Do you want me to add it anyway? Basically, I'd just check if preloader is set to false specifically around https://github.com/ruffle-rs/ruffle/pull/9802/files#diff-467a580c71d4e3b96032b54a2a292d485f3b83f87ba2478daf693bac6d532965R423 and log "The configuration option preloader has been replaced with splashScreen. If you own this website, please update the configuration."

Yup, let's add a `console.warn` in this PR so people will know how to migrate. I think this is the least we can do without endeavoring too much.

> > Also, is it possible to have a telemetry server that collects usage statistics
> 
> This sounds like a larger, but useful, discussion. I wouldn't want to tie that discussion to this PR personally though.

Agree, just wrote down the idea. Definitely not directly related to this PR.

relrelb (approved):

PR #9811, opened 2023-03-02
avm2: Store Multiname instead of QName in QNameObject

### LINKED ISSUE #9635: Amuse Park: Unable to interact with almost all of the attractions while in-game
### Describe the bug

Tested on Ruffle Nightly 2023-02-19
Amuse Park SWF used: [amusepark.zip](https://github.com/ruffle-rs/ruffle/files/10776309/amusepark.zip)

This is what the game looks like in the Ruffle Demo. The only attraction I can interact with is the clean-up robot. Any button that does not relate to the attractions works perfectly fine.

https://user-images.githubusercontent.com/36887711/219933496-5e37faf0-517e-4426-ab10-b87dcb87e529.mp4

### Expected behavior

This is what the game looks like when I launch it using Flashpoint.

https://user-images.githubusercontent.com/36887711/219933475-f99b93cf-924f-4dfb-9e27-6519c5603c91.mp4

### Affected platform

Online demo

### Operating system

Windows 10

### Browser

Chrome 110

### Additional information

_No response_

Bale001:
This makes QNameObject store a `Multiname` internally, rather than a `QName`. This fixes issues with sending `QNameObject`'s over the proxy class, and is also a good first step in removing QName completely from Ruffle.

Fixes #9635 

~Will add some tests to this PR later~ done


core/src/avm2/globals/xml.rs
```diff
@@ -127,28 +127,30 @@ pub fn attribute<'gc>(
     let this = this.unwrap();
     let xml = this.as_xml_object().unwrap();
     let name = args[0];
-    let qname = match name {
-        Value::String(s) => QName::new(activation.avm2().public_namespace, s),
+    let multiname = match name {
+        Value::String(s) => QName::new(activation.avm2().public_namespace, s).into(),
```
adrian17, line 131: Note: all these uses of `QName::new(...).into()` could be simply converted to `Multiname::new(...)`

core/src/avm2/multiname.rs
```diff
@@ -341,11 +326,10 @@ impl<'gc> Multiname<'gc> {
 
     /// Indicates if this multiname matches any type in any namespace.
     pub fn is_any(&self) -> bool {
-        self.name.is_none()
-            && match self.ns {
-                NamespaceSet::Single(ns) => ns.is_any(),
-                NamespaceSet::Multiple(ns) => ns.iter().any(|ns| ns.is_any()),
-            }
+        match self.ns {
```
adrian17, line 329: why this change? Not checking for name doesn't seem like what we want.

core/src/avm2/object/qname_object.rs
```diff
@@ -43,15 +48,15 @@ pub struct QNameObjectData<'gc> {
...
 }
 
 impl<'gc> QNameObject<'gc> {
-    /// Box a QName into an object.
-    pub fn from_qname(
+    /// Box a Multiname into an object.
+    pub fn from_name(
         activation: &mut Activation<'_, 'gc>,
-        qname: QName<'gc>,
+        name: impl Into<Multiname<'gc>>,
```
adrian17, line 59: IMO let's keep being explicit here - the only other thing that can be turned into a Multiname are QNames, and I'd prefer them to be obvious.

core/src/avm2/multiname.rs
```diff
@@ -420,6 +404,53 @@ impl<'gc> Multiname<'gc> {
...
+            WStr::empty()
+        };
+
+        if ns.is_empty() {
+            if let Some(name) = self.name {
+                return name;
+            } else {
+                // For some reason when the namespace is empty but it is an `any` name,
+                // flash gets confused and prints that the namespace is also `any`.
+                return "*::*".into();
```
adrian17, line 441: how did you reproduce this? I'm getting a different result:
```
namespace nss = "";

class C extends Proxy {
    flash_proxy override function getProperty(name: *) {
        trace(name);
        trace(name.uri);
        trace(name.uri === "");
    }
}
var a = new C();
a.nss::*; // same with a.*
```
Traces
```
*

true
```

Bale001: Ah, so this is actually left over from an incorrect test I did & I didn't double check it. 

```as
var q = new QName("*");
trace(q.uri);
trace(q.localName);
trace(q);
```
prints...
```
null
*
*::*
```
But then,
```as
var q = new QName("test");
trace(q.uri);
trace(q.localName);
trace(q);
```
prints
```

test
test
```
The weird behavior turns out not to be in `as_uri`, but in QName's constructor. When "*" is passed as the local name of the qname, it makes the namespace `any` (it would usually be public by default).

core/src/avm2/multiname.rs
```diff
@@ -339,13 +324,17 @@ impl<'gc> Multiname<'gc> {
...
-                NamespaceSet::Single(ns) => ns.is_any(),
-                NamespaceSet::Multiple(ns) => ns.iter().any(|ns| ns.is_any()),
-            }
+    }
+
+    /// Indicates if this multiname matches any namespace.
+    pub fn is_any_namespace(&self) -> bool {
+        match self.ns {
+            NamespaceSet::Single(ns) => ns.is_any(),
+            NamespaceSet::Multiple(ns) => ns.iter().any(|ns| ns.is_any()),
```
adrian17, line 336: (I think it's actually not allowed for namespace-sets to contain Any names - we could validate and use that maybe, but that's not needed in this PR)

adrian17 (approved):

PR #9813, opened 2023-03-02
extension: Give precedence to toggles over default options (fix #5825)

### LINKED ISSUE #5696: Dragon Ball Devolution no download option in context menu
### Describe the bug

I manually installed the extension in Microsoft Edge, then opened the game. When I right clicked it, there was no download option in the context menu. I made sure to toggle it on in the settings menu. The game's official website is [https://www.txori.com/index.php?static5/dbdevolution](url).
Context menu:
screenshot (95)
My settings:
screenshot (96)
Someone on Ruffle's official discord (nosamu#3385) stated that they "think it must be a conflict between the selfhosted and extension versions of Ruffle, since the website is self-hosting" and that they "think that even though the extension version of Ruffle is loaded, the preferences are loaded from the webpage instead of from the extension".
You can see the full Discord conversation here: [https://discord.com/channels/610531541889581066/722872096002408529/911812753604378634](url).

### Expected behavior

I expected there to be a download option in the context menu.

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Microsoft Edge Version 96.0.1054.29 (Official build) (64-bit)

### Additional information

_No response_

### LINKED ISSUE #5825: Prefer extension options over default configuration for any options defined in both locations
For showSwfDownload, logLevel, and warnOnUnsupportedContent, on any sites which self-host Ruffle already, prefer the extension toggle over the self-hosted configuration (or default values, which are currently preferred if the values are not set in the configuration). Continue to obey configuration for options that are not toggleable with the extension, like base.

Up for debate: Decide whether to only ignore the default self-hosted configuration values that are not explicitly set, or to always ignore self-hosted configuration options in favor of extension toggles.

danielhjacobs:
It is possible to give them precedence over self-hosted options as well by merely rearranging the [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#spread_in_object_literals) for `loadedConfig`. But @Elementtair suggested that "I would feel it as disrepectfull to override self-host configuration for a stable release of the extension."

Edit: Also fixes #5696. Though that particular game is multi-asset, so downloading the base SWF is not very useful.

Edit 2: Note - this only works when the extension is the version of Ruffle that winds up taking precedence. If the self-hosted build takes precedence, the extension is not loaded until after RufflePlayer.load, and therefore the extension configuration is not available until then , so I can't take the toggles into account before loading content. To fix that we'd first need to fix https://github.com/ruffle-rs/ruffle/discussions/7505. But this PR is an improvement. If the extension is what's used, its toggles are respected.

### DIFF
--- a/web/packages/core/src/polyfills.ts
+++ b/web/packages/core/src/polyfills.ts
@@ -196,9 +196,17 @@ export function pluginPolyfill(): void {
  * Polyfills legacy Flash content on the page.
  *
  * @param isExt Whether or not Ruffle is running as a browser's extension.
+ * @param extensionConfig The configuration sent by the PublicAPI, stored if this is the extension.
  */
-export function polyfill(isExt: boolean): void {
+export function polyfill(
+    isExt: boolean,
+    extensionConfig: URLLoadOptions | DataLoadOptions | object
+): void {
     isExtension = isExt;
+    window.RufflePlayer ||= {};
+    if (isExtension) {
+        window.RufflePlayer.extensionConfig = extensionConfig;
+    }
     polyfillFlashInstances();
     polyfillFrames();
     initMutationObserver();
--- a/web/packages/core/src/public-api.ts
+++ b/web/packages/core/src/public-api.ts
@@ -12,7 +12,10 @@ declare global {
          * [[PublicAPI]] instance itself.
          */
         RufflePlayer?:
-            | { config?: DataLoadOptions | URLLoadOptions | object }
+            | {
+                  config?: DataLoadOptions | URLLoadOptions | object;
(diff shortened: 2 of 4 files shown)


Elementtair:
What I meant by disrecpecfull, is that people owning websites and offering flash content relies on a flux of visitors on their website. 
If you offer a too easy way to bypass their config and downloading content just because you can isn't helping those.
I can understand the frustation of websites not up to date with ruffle releases, but in the end, if not maintained, those sites are going to disapear. Will only survives those who are maintained, and you ll cut out part of their traffics.

PR #9818, opened 2023-03-02
avm2: Implement describeType

Aaron1011:
This includes all of the XML elements described in 'describeType' docs. Unfortunately, the order of elements produced by Flash depends on the iteration order of internal hashtables. As a result, the test manually stringifies an XML object, sorting the stringified children, to produce consistent output between Flash and Ruffle.


adrian17 (approved):
I wish we had a more convenient API for mutating the XML instead of formatting the string ;)

Dinnerbone (approved):

Lord-McSweeney:
There are a couple `panic!`s left in, not sure if those should be `Err`s

Aaron1011:
@Lord-McSweeney Those should be unreachable - they indicate that we've somehow constructed an invalid vtable.

Lord-McSweeney (approved):

PR #9821, opened 2023-03-03
wgpu: Cache gradient colors in a texture

Dinnerbone:
It's enough of an optimisation to allow us to merge all the gradient pipelines together again, and still be between 5-10% faster than before when rendering gradients.

This will speed up startup time a fair bit.


render/wgpu/shaders/gradient.wgsl
```diff
@@ -16,32 +16,22 @@ struct VertexOutput {
...
-}
-
-fn find_t(focal_point: f32, uv: vec2<f32>) -> f32 {
-    return 0.0;
-}
-
-/// FIXME: We should import VertexInput from 'common', but a naga_oil bug prevents
-/// us from importing 'common' in both this file and the invidual gradient shaders (like 'linear.wgsl')
-/// Currently, importing 'common' in that way will cause all of the definitions from 'common' to be duplicated,
-/// result in a WGSL parse error.
```
torokati44, line 44: Is this comment really not relevant anymore?

Dinnerbone: Not strictly, after we split up the one global VertexInput into different types

Aaron1011 (approved):

PR #9824, opened 2023-03-03
web: Fixed missing log messages when in futures, or using `log` crate

Dinnerbone:


web/src/lib.rs
```diff
@@ -494,6 +494,12 @@ impl Ruffle {
         js_player: JavascriptPlayer,
         config: Config,
     ) -> Result<Ruffle, Box<dyn Error>> {
+        // Redirect Log to Tracing if it isn't already
+        let _ = tracing_log::LogTracer::builder()
+            // wgpu crates are extremel verbose
```
torokati44, line 499: nit: typo: "extremely"

torokati44:
Do you think this is also why I wasn't seeing `tracing` log messages in the console from `Buffer::play` in #9671?

Dinnerbone:
> Do you think this is also why I wasn't seeing `tracing` log messages in the console from `Buffer::play` in #9671?

It makes sense! There might be other areas we're entering into Ruffle which isn't wrapped, but I think this was it from what I saw

Toad06:
Not specifically related to this PR, but I would like to mention that since #9026 it is possible to experience serious slowdowns in Firefox when a file logs a lot of messages into the browser's console - even if the user never opens the console (which is the difference with the previous behavior).

The issue is particularly noticeable in Critical Zone:
- Using the web extension, open the parameters page and set "log level" to "info".
- Browse to https://www.ferryhalim.com/orisinal/g2/criticalzone.htm without the console open (the server if often down, so I'm also attaching the game [here](https://github.com/ruffle-rs/ruffle/files/10883046/critical_zone.zip)).
- Play the game. After only some seconds, you should notice the issue.

So I was wondering if there is anything we can do on our end to prevent that or if this is just an issue related to Firefox (as Chrome doesn't seem affected).

Dinnerbone:
> Not specifically related to this PR, but I would like to mention that since #9026 it is possible to experience serious slowdowns in Firefox when a file logs a lot of messages into the browser's console - even if the user never opens the console (which is the difference with the previous behavior).
> 
> The issue is particularly noticeable in Critical Zone:
> 
>     * Using the web extension, open the parameters page and set "log level" to "info".
> 
>     * Browse to https://www.ferryhalim.com/orisinal/g2/criticalzone.htm without the console open (the server if often down, so I'm also attaching the game [here](https://github.com/ruffle-rs/ruffle/files/10883046/critical_zone.zip)).
> 
>     * Play the game. After only some seconds, you should notice the issue.
> 
> 
> So I was wondering if there is anything we can do on our end to prevent that or if this is just an issue related to Firefox (as Chrome doesn't seem affected).

Can you try the latest commit and see if it helps for you? It seems to for me but I don't want to rule out it being a fluke of me testing

Toad06:
Yes, I confirm there's no more slowdown on my end as well. Thank you for the fix. 😉

torokati44 (approved):
I see nothing obviously wrong with the code _(not that I'm an expert in this area... :wink:)_, and I can confirm it fixes at least one annoying issue (that of missing log messages from the audio callback), so LGTM I guess!

Aaron1011 (approved):

PR #9845, opened 2023-03-04
avm2: Implement Stage3D depth test, blend factors, and fix bugs

Aaron1011:
Context3D.setBlendFactors is now fully implemented. Context3D.setDepthTest is implemented, but the associated methods Context3D.setStencilActions and Context3D.setStencilReferenceValue are not yet implemented.

In the process of implementing this and adding test SWFs, I found and fixed some bugs:
* We now correctly handle setVertexBufferAt calls that use different vertex buffers.
* Context3D.drawTriangles now correctly handles a non-zero firstIndex

Implementing blend modes was somewhat tricky. The final Stage3D texture is draw without any transparency - that is, whatever values are in the buffer at the time of 'present()' get copied to the screen, with alpha set to 1.0. We didn't have the ability to do this before, so I added a new `bitmap_opaque.wgsl` shader that does this.


adrian17:
(more pipelines :( )

Aaron1011:
@adrian17 I think adding new pipelines here is unavoidable, unfortunately. However, Dinnerbone's suggestion allowed me to avoid adding an additional shader.

Dinnerbone (approved):

PR #9846, opened 2023-03-04
avm2: Implement KeyboardEvent.keyLocation getter/setter

Aaron1011:


Lord-McSweeney (approved):

Lord-McSweeney:
I think you need to add `private var _keyLocation` on line 9, next to the other two `private` `uint` values.

Aaron1011:
> I think you need to add `private var _keyLocation` on line 9, next to the other two `private` `uint` values.

Thanks - fixed

Bale001 (approved):

Issue #9562, opened 2023-02-14
Bloons TD 3: Cannot buy tower upgrades


n0samu:
### Describe the bug

In [Bloons Tower Defense 3](https://www.kongregate.com/games/ninjakiwi/bloons-tower-defense-3), you cannot buy the tower upgrades, even if you have enough money to afford them.
![image](https://user-images.githubusercontent.com/71368227/218601784-bf40d39b-ea84-4da9-8927-97bcd5f82198.png)

This seems to me like another mouse pick issue.

### Expected behavior

Clicking the tower upgrades when you have enough money should buy them.

### Affected platform

Desktop app

### Operating system

Windows 10

n0samu:
Fixed by #9867

Issue #9808, opened 2023-03-02
Totems Awakening: Cannot click Continue button on sponsor splash screen


n0samu:
### Describe the bug

Link to SWF: http://i.notdoppler.com/files/totemsawakening.swf

In [Totems Awakening](http://www.notdoppler.com/totemsawakening.php), the Continue button on the PlayHub splash screen is unclickable because Ruffle gives higher priority to the click handler attached to the entire splash screen. Thus clicking anywhere, even on the Continue button, opens the PlayHub website.

![image](https://user-images.githubusercontent.com/71368227/222375447-67faa128-c4a7-4c6b-83af-e6303f67aede.png)

### Expected behavior

Clicking the Continue button should go to the main menu.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

The relevant code is in the SplashScreen class:
```as
private function initClickListener() : void
{
    buttonMode = true;
    addEventListener(MouseEvent.CLICK,this.onClick,false,0,true);
}
```
```as
private function initContinueButton() : void
{
    this.continueButton = this.bg.getChildByName("ContinueButton") as MovieClip;
    this.continueButton.buttonMode = true;
    this.continueButton.addEventListener(MouseEvent.ROLL_OVER,this.onContinueButtonRollOver,false,0,true);
    this.continueButton.addEventListener(MouseEvent.ROLL_OUT,this.onContinueButtonRollOut,false,0,true);
    this.continueButton.addEventListener(MouseEvent.CLICK,this.onContinueButtonClick,false,0,true);
}
```
Some theories for why the Continue button handler is prioritized:
- SplashScreen ultimately inherits from Sprite, while the continueButton is a movieclip. Maybe movieclips are prioritized over sprites.
- Maybe the continueButton movieclip has a higher depth value than the overall SplashScreen sprite.
- SplashScreen is instantiated via ActionScript, while continueButton is placed within SplashScreenBackground by a PlaceObject tag. Maybe that gives continueButton higher priority.

n0samu:
Fixed by #9867

Issue #9738, opened 2023-02-26
Train Mania: Level select buttons are very difficult to click


n0samu:
### Describe the bug

Originally mentioned in #8874, the level select buttons in [Train Mania](http://www.notdoppler.com/trainmania.php) are extremely difficult to click:

https://user-images.githubusercontent.com/71368227/221402483-4741a2fe-ccb4-480f-a93f-8a64e77f22bd.mp4

Also, the buttons on the in-game Pause menu are only clickable the first time you open the Pause menu (by clicking the gear icon). The second time you open the menu, all of the buttons are unclickable.

### Expected behavior

Buttons should be easily clickable.

### Affected platform

Desktop app

### Operating system

Windows 10

n0samu:
Fixed by #9867

Issue #9534, opened 2023-02-12
Mario Forever Flash: Stuck at game loaded menu.


p0008874:
### Describe the bug

https://user-images.githubusercontent.com/75534590/218322788-2d2157fd-0585-4f28-8898-710b97e03846.mp4

```
2023-02-12T16:10:54.455329Z  WARN run_frame:run_all_phases_avm2: ruffle_core::stub: Encountered stub: AVM2 flash.system.Security.allowDomain getter
2023-02-12T16:10:54.457434Z  WARN run_frame:run_all_phases_avm2: ruffle_core::stub: Encountered stub: AVM2 flash.system.Security.allowInsecureDomain getter
2023-02-12T16:10:54.459540Z  WARN run_frame:run_all_phases_avm2: ruffle_core::stub: Encountered stub: AVM2 flash.net.LocalConnection.allowDomain()
2023-02-12T16:10:54.460050Z  WARN run_frame:run_all_phases_avm2: ruffle_core::stub: Encountered stub: AVM2 flash.net.LocalConnection.allowInsecureDomain()
2023-02-12T16:10:54.460515Z  WARN run_frame:run_all_phases_avm2: ruffle_core::stub: Encountered stub: AVM2 flash.net.LocalConnection.connect()
2023-02-12T16:10:55.001098Z ERROR ruffle_core::loader: Error during movie loading: FetchError("HTTP status is not ok, got 404 Not Found")
2023-02-12T16:10:57.468427Z  INFO run_frame:run_all_phases_avm2: avm_trace: [MochiAd] Couldn't load an ad, make sure your game's local security sandbox is configured for Access Network Only and that you are not using ad blocking software
2023-02-12T16:10:57.735374Z  WARN ruffle_core::stub: Encountered stub: AVM2 flash.net.LocalConnection.close()
```

### Expected behavior

https://user-images.githubusercontent.com/75534590/218322805-d8d281fc-d930-413f-ac9a-f38c3b4bf7d9.mp4

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

[marioforeverflash.zip](https://github.com/ruffle-rs/ruffle/files/10716543/marioforeverflash.zip)

n0samu:
This seems like another mouse pick issue. The game's onEnterFrame handler has this:
`gfx.addEventListener(MouseEvent.CLICK,clickman2);`
I can confirm that this code runs. And here is the clickman2 function:
```as
[...]

PR #9867, opened 2023-03-06
avm2: Fix mouse picking on MovieClip in 'button mode'

Aaron1011:
We were previously performing a redundant 'self.hit_test_shape' call in 'avm2_mouse_pick'. All of the logic in that function is handled in `avm2_mouse_pick.` Additionally, this call happened before we tested out children, which would result in us targeting a parent's drawing instead of aa child.


n0samu:
This fixes #9562, #9808, #9738, and #9534! 🎉

Dinnerbone (approved):

Minigugus:
This commit broke Happy Wheels demo (https://www.newgrounds.com/portal/view/547504):

before (not show on the video but clicks works):
https://user-images.githubusercontent.com/43109623/225154616-f90d9033-4df5-4fc2-b63a-097dd8e5bd7e.mp4

after:
https://user-images.githubusercontent.com/43109623/225154739-7350a782-dc7d-4e59-8e71-f2196d497be5.mp4

Issue #9564, opened 2023-02-14
Bloons TD 3: Road spikes, monkey glue, and pineapple are invisible


n0samu:
### Describe the bug

In [Bloons Tower Defense 3](https://www.kongregate.com/games/ninjakiwi/bloons-tower-defense-3), the road spikes, monkey glue, and pineapple items can be placed down and affect the bloons as they should, but they are invisible.

### Expected behavior

The items should be visible on the track when placed down.

### Affected platform

Desktop app

### Operating system

Windows 10

n0samu:
This has been fixed!

n0samu:
...Well, not quite yet, but it will be soon 😉

Issue #9563, opened 2023-02-14
Bloons TD 3: The tack shooter never shoots


n0samu:
### Describe the bug

In [Bloons Tower Defense 3](https://www.kongregate.com/games/ninjakiwi/bloons-tower-defense-3), the tack shooter tower never shoots, even when bloons are in range.

### Expected behavior

The tack shooter tower should... shoot tacks.

### Affected platform

Desktop app

### Operating system

Windows 10

PR #9874, opened 2023-03-06
core: Fix setting certain DisplayObject properties to NaN

Aaron1011:
This doesn't perfectly match Flash's behavior - I haven't been able to reproduce the values produces when the DisplayObject starts out with certain 'Matrix' values (a non-zero 'b' or 'd').

Howver, when the 'b' and 'd' matrix values are both 0, setting 'dobj.rotation = NaN' has no effect on the matrix, while 'dobj.scaleX = NaN' and 'dobj.scaleY = NaN' both treat 'NaN' as 0 for the purposes of updating the matrix.

This fixes the tack shooter in Bloons Tower Defense 3, which tries to set 'rotation = NaN' for spawned tacks.


n0samu:
Confirmed that this fixes #9563!

n0samu:
And this also fixes #9564!

adrian17 (approved):

PR #9879, opened 2023-03-06
avm2: Export Vector classes in public __AS3__.vec namespace

Aaron1011:
Previously, the Vector$ classes were only exported in the internal '__AS3__.vec' namespace, which is used by older ActionScript code. However, newer ActionScript code can also access these classes through the public '__AS3__.vec' namespace, via 'getDefintionByName'.

We now export these classes in both namespaces. In the public '__AS3__.vec' namespace, they are exported like 'Vector.<uint>' instead of 'Vector$uint'


core/src/avm2/class.rs
```diff
@@ -218,6 +220,26 @@ impl<'gc> Class<'gc> {
...
         new_class.class_initializer_called = false;
 
+        if params.len() > 1 {
+            panic!(
+                "More than one type parameter is unsupported: {:?}",
+                self.name()
+            );
+        }
+
+        // FIXME - we should store a `Multiname` instead of a `QName`, and use the
```
Lord-McSweeney, line 230: Does #9811 change this at all?

Aaron1011: No - that doesn't affect the internal `QName` and `Multiname` representations

adrian17 (approved):

PR #9893, opened 2023-03-07
avm2: Bring StageObject's DisplayObject creation into the same place

Dinnerbone:
This doesn't entirely solve the ordering issue, but it's a step towards it. Now there's a way to get and set up the DO regardless of what it is, and the next step is to figure out how to do that in initializer time.

I tried not to change the logic any, just moved things around.


iwannabethedev:
Is there any relation between this PR and #11084 ?

Lord-McSweeney:
Obsoleted by #11084.

PR #9894, opened 2023-03-07
avm2: Throw AVM error for null arg in removeChild()

Dinnerbone:
Apparently this was a common thing: `try { removeChild(thingThatMightNotExist} catch(e:Error) {}`

This makes Kingdom Rush a little playable, but it's then blocked by BitmapData.hitTest


Bale001 (approved):

Lord-McSweeney (approved):

Aaron1011:

Dinnerbone:
> Could you add a test for this?

I can, but, are you sure? We're going to need to slowly apply this to many many many places and testing every param of every method to see what happens if it's null/undefined is going to be a lot of tests.

adrian17:
>I can, but, are you sure? We're going to need to slowly apply this to many many many places and testing every param of every method to see what happens if it's null/undefined is going to be a lot of tests.

I agree in general, but as you said - if this particular case is something many swfs do, we should probably have it tested anyway.

core/src/avm2/globals/flash/display/display_object_container.rs
```diff
@@ -245,8 +245,8 @@ pub fn remove_child<'gc>(
             .get(0)
             .cloned()
             .unwrap_or(Value::Undefined)
-            .as_object()
-            .and_then(|o| o.as_display_object())
+            .coerce_to_object_or_typeerror(activation, None)?
```
adrian17, line 248: FYI, FP throws a different error here:
```
TypeError: Error #2007: Parameter child must be non-null.
```
While `coerce_to_object_or_typeerror` is for a more generic error handling, this error 2007 is more of checking arguments.
Avmplus uses it like:
```
toplevel()->checkNull(charSet, "charSet");
```
so I'm guessing we should have some separate method for it too?

That said, don't think the exact message is a priority, code 2007 is more important. We can also TODO it ofc.

Dinnerbone: Fixed!

Dinnerbone:
Added test + used new ParametersExt to give correct error code.

PR #9898, opened 2023-03-07
avm2: Run orphan framescripts that didn't get a chance to run

Aaron1011:
This is necessary to make Steambirds get past the preloader screen. All of the previous tests continue to pass with this change.

This commit modifies the existing test to start from within the symbol_class constructor, instead of a frame script. In this situation, a freshly-created orphan with a framescript will run directly after the constructor returns, *before* an enterFrame handler for the same orphan. I've verified that this modified test fails without my change.


core/src/frame_lifecycle.rs
```diff
@@ -85,6 +87,8 @@ pub fn run_all_phases_avm2(context: &mut UpdateContext<'_, '_>) {
 
     stage.enter_frame(context);
 
+    let mut ran_framescript = HashSet::new();
```
Dinnerbone, line 90: Would `FnvHashSet` make more sense here? I think we generally use FnvHash over Hash when we prefer speed over potential DOS attacks

Aaron1011: Good point - I've changed it

Dinnerbone (approved):

PR #9902, opened 2023-03-07
wgpu: Fix gradient regression

torokati44:
This fixes some regressions introduced in #9821.

There were two problems:
 - The computation if `a` (the lerp parameter) could underflow  `u8`.
 - The `last` variable was not set to the index of the last record after the ratio of the last record was surpassed.
 
~Regression tests to follow soon.~
Thanks to @AllinolCP for the repro case, which served as the basis of the visual regression test!


torokati44:
While we already have a regression test for simple radial gradients, the critaria for failures were:
The first record was not at ratio `0` and/or the last record was not at ratio `255`.
The test added now also covers linear/srgb interpolation.
But it seems a bit unorganized to just put it where it is, in itself.

AllinolCP:
off topic:
> Thanks to @Allinol (on Discord) for the repro case, which served as the basis of the visual regression test!
hi so @Allinol is actually my alt

n0samu:
I can confirm that this fixes gradient regressions in Nanaca Crash, Papa's Burgeria, and Bob the Robber - thank you!

However this does not fix the regression I observed in Ninja Painter. When hovering over the Main Menu button on the level select screen, here is what it is supposed to look like: 
image

But here's what it looks like ever since Dinnerbone's PR was merged:
image

Here is the SWF: http://i.notdoppler.com/files/ninjapainter.swf

torokati44:
I think ninjapainter issue is fixed, would you please be so kind as to re-test all regressions?

torokati44:
(Another set of changes pushed, sorry.)

Dinnerbone (approved):

torokati44:
Thanks for testing, @Bale001 and @n0samu (and others)!

Issue #9187, opened 2023-01-17
web: Add 'preferred renderer' config option


n0samu:
**Describe the feature**

With the switch to wgpu-webgl, I think it's even more important that we add a config option to the web version of Ruffle that allows choosing the render backend:
- It makes it much easier to test the behavior of the different render backends. No need to manually disable WebGL 1/2 in the browser's config to force a fallback to the WebGL/Canvas backends.
- Some content does not run well with wgpu-webgl yet. For example, [Bloons Super Monkey](http://www.notdoppler.com/bloonssupermonkey.php) performs badly and uses excessive GPU resources when you buy the plasma tentacle upgrade because the blend mode that it uses is not well-optimized in Ruffle yet. Switching to the WebGL backend resolves the problem. (The blend mode barely affects how the tentacle displays)
- Some content is drawn more accurately with the Canvas backend due to a stroke width issue with other backends. See #7369 for an example.

The behavior I'd suggest for this is to keep the same fallback behavior as usual, but just 'start' at the preferred renderer. So if WebGL is chosen as preferred, it still falls back to Canvas if WebGL is not available.

Issue #1905, opened 2020-12-10
Crash with Open Flash Charts


richdenton:
# Error Info
Error name: Error
Error message: panicked at 'called `Result::unwrap()` on an `Err` value: "Unable to create VAO"', render/webgl/src/lib.rs:465:50
Error stack:
```
D@http://XXXXXXXXXX/admin/js/ruffle/core.ruffle.563f1d11a2766abcb2a4.js:1:20381
__wbg_new_df6e6ab7a65c4c4d@http://XXXXXXXXXX/admin/js/ruffle/ruffle.js:1:34989
wasm-stub@[wasm code]
<?>.wasm-function[1056]@[wasm code]
<?>.wasm-function[1379]@[wasm code]
<?>.wasm-function[3696]@[wasm code]
<?>.wasm-function[2923]@[wasm code]
<?>.wasm-function[3993]@[wasm code]
<?>.wasm-function[4023]@[wasm code]
<?>.wasm-function[3898]@[wasm code]
<?>.wasm-function[4024]@[wasm code]
<?>.wasm-function[3427]@[wasm code]
<?>.wasm-function[398]@[wasm code]
<?>.wasm-function[1752]@[wasm code]
<?>.wasm-function[1033]@[wasm code]
<?>.wasm-function[1228]@[wasm code]
<?>.wasm-function[328]@[wasm code]
<?>.wasm-function[550]@[wasm code]
<?>.wasm-function[1560]@[wasm code]
<?>.wasm-function[2737]@[wasm code]
<?>.wasm-function[2476]@[wasm code]
<?>.wasm-function[4191]@[wasm code]
wasm-stub@[wasm code]
_dyn_core__ops__function__FnMut__A____Output___R_as_wasm_bindgen__closure__WasmClosure___describe__invoke__ha67c67d9b57d32de@[native code]
E@http://XXXXXXXXXX/admin/js/ruffle/core.ruffle.563f1d11a2766abcb2a4.js:1:18826
f@http://XXXXXXXXXX/admin/js/ruffle/core.ruffle.563f1d11a2766abcb2a4.js:1:17506
promiseReactionJob@[native code]
```

# Player Info
Allows script access: false

# Page Info
Page URL: http://XXXXXXXXXX/ #admin/exec_dash/

# Browser Info
[...]

PR #9904, opened 2023-03-07
web: Add "preferred renderer" config option

n0samu:
This allows users of web builds of Ruffle to choose a preferred render backend. Ruffle will prefer the specified backend while retaining the same fallback order as before. Includes a dropdown on the extension settings page.
image


SN902:
I think we should focus on WGPU, it will be faster in the future for most users with GTX 960 or better. If we fallback to old realization, it will be minor technical issue for most people with good hardware (as example how x86 applications slowdown performance at 5-10% and make trouble with memory barriers, they needed to patch with PAE to take advantage of 4GB RAM). Why 95% of people must shoot in own legs and helping 5% worldwide users with very outdated PC's, as i seen in some bug reports, they have GTX 650, 7xx or past gen intel integrated GPU's. Even if we stick on WebGL1, their systems may struggle with bad performance on some very complicated swf animations.

torokati44:
While development, this could make it easier to test, say, the Canvas2D backend. But there are ways to force either backend to be used on a dev machine. Otherwise I mostly agree with @Dinnerbone.
Also there's an issue where we can't have both the WebGPU and WebGL back-backends of the wgpu backend enabled in any single build AFAIK. - So unless we make a separate, direct WebGPU backend (skipping wgpu) - which I'm not sure why we would do -, those two options can't really coexist. ... (Unless the extension and a selfhosted build are configured with one and the other, but that's way too much of a hypothetical.)

Dinnerbone:
> Also there's an issue where we can't have both the WebGPU and WebGL back-backends of the wgpu backend enabled in any single build AFAIK. - So unless we make a separate, direct WebGPU backend (skipping wgpu) - which I'm not sure why we would do -, those two options can't really coexist. ... (Unless the extension and a selfhosted build are configured with one and the other, but that's way too much of a hypothetical.)

WGPU are working on that, and we can workaround that by hypothetically splitting off each renderer into its own build, switched the same way we do vanilla vs extensions wasm

adrian17:
> I think we should focus on WGPU, it will be faster in the future for most users with GTX 960 or better.

Not sure what you mean here; wgpu-webgl currently is definitely much heavier on low-spec devices. I cry inside every time I open https://ruffle.rs/demo/ in my VM to do any manual experimenting :/

n0samu:
> By adding this feature we are guaranteeing that WebGL (and for that matter, wgpu-webgl) sticks around for the future, as it's now part of our API. Our API can only be additive, really.

No, this PR adds a **preferred** renderer config option. Notice that the default value is actually `webgpu`, which is not actually available at the moment, and thus is never used. We aren't guaranteeing that Ruffle will honor the user's preference. We can still remove or change our list of renderers at any time, and we should make that clear in our documentation. (I can add a blurb about that in the doc comment as well if you like.)

> They'd need to know to expect things not working and that we're not going to fix it. The same can be said for canvas, too.

You're right. Although right now, using the Canvas backend is a way to avoid the stroke width issue and thus for some specific content, the Canvas backend is the best choice. See #9187 (which I forgot to link to this PR initially). Anyway, we can explain on our wiki page that wgpu is the only actively developed and supported backend at the moment.

>  I don't understand any possible case for users intentionally choosing it

Please do take a look at #9187.

> Say a web host chose "webgpu" - suddenly nobody's going to be able to access Ruffle because that's not working for the vast majority of people today.

What? Please test out my PR, that's not how it works at all!

Herschel:
This is more of a developer/power-user feature, not a public-intended API; so I don't think API concerns really apply here.  I've wanted this for quite so time, mainly because it's such a hassle to test specific backends. For example, this allows us to make the renderer selectable in the demo page, or when I'm testing dummy sites on a local web server.

 * Naturally the API should be documented appropriately: "unstable API; renderers are subject to change and may be added, removed, and broken; use at your own risk". We could also just make it `@internal`, but I don't think being public is a problem as long as the behavior is documented.
 * The behavior for an invalid renderer can simply choose the best available renderer, and this can be mentioned in the docs. A priority list is unnecessary.
 * Maybe use a string instead of enum for the renderer type to indicate non-exhaustiveness? Maybe that's overkill, not sure what the TypeScript idiom is.
 * I would love not having toggle on/off hardware acceleration when you want to test the canvas backend.

Herschel:
> * Maybe use a string instead of enum for the renderer type to indicate non-exhaustiveness? Maybe that's overkill, not sure what the idiomatic TypeScript is.

Thinking more about this, an enum is fine -- I think it's okay for us to codify which backends exist in the API without guaranteeing anything about their actual support or behavior, as long as this is documented. Removing a backend should be considered a semver breaking change, anyway.

web/packages/core/src/config.ts
```diff
@@ -58,4 +58,5 @@ export const DEFAULT_CONFIG: Required<Config> = {
     publicPath: null,
     polyfills: true,
     playerVersion: null,
+    preferredRenderer: RenderBackend.WebGpu,
```
Herschel, line 61: I think this would better default to `null`, to better indicate that this is subject to change and that the default behavior is to choose the best available renderer.

n0samu: Makes sense! I'll update it accordingly when I get a chance

Toad06:
Thanks for your work on this. I think this option is definitely useful for testing and could also be used as a workaround for users facing the VAO issues (#1905 / #9665).

I tried the PR, it works well with the self hosted build and also with the extension when the SWF is open directly. However, the option seems to be ignored when the SWF is played inside a webpage with the extension - maybe that the option needs to be added here:
https://github.com/ruffle-rs/ruffle/blob/1edcbe438dbd174e29da8d9a74c99e43ed4ab7dc/web/packages/extension/src/content.ts#L170-L178

I think it could also be worth to mention that this is quite an advanced option, to discourage users from changing it if they don't understand what this option is meant to be. Something like that for instance:
options

n0samu:
@Toad06 Thanks so much for pointing out that issue, I completely missed it. It should be fixed now!

As for your suggestion, I decided to rearrange the extension options from most to least commonly used, and by type:
image

I also removed the "Ignore website compatibility warnings" option from the extension popup since it is not very useful or commonly used.

Toad06 (approved):

Lord-McSweeney (approved):

Herschel:
I think the "wgpu" label should be either `wgpu (WebGL)` or `WebGL (wgpu)`. Otherwise this looks great to me.

Thanks! I'd like to get this in sooner than later so that we can integrate this before WebGPU goes live in Chrome 113. That way we can play with the different renderers more easily on the demo page.

@danielhjacobs Would you have any interest in adding a selector for this on the demo page? I guess to start this could be an additional dropdown somewhere, or maybe we turn the ℹ️ panel into a full options panel.

n0samu:
Alright, I've changed the dropdown item on the extension options page to say "Wgpu (via WebGL)".

@Dinnerbone Can you take a quick look at the doc comments I wrote and let me know if you think they're good?

n0samu:
As for the demo page, I think a renderer dropdown would be great to add to [Ruffle-Regressions](https://ruffle-regressions.github.io/). Considering its limited usefulness to most users, we may not want to add it to the official demo though.

web/packages/core/src/load-options.ts
```diff
@@ -117,6 +117,39 @@ export const enum WindowMode {
     Gpu = "gpu",
 }
 
+/**
+ * The render backend of a Ruffle player.
+ */
+export const enum RenderBackend {
+    /*
+     * A draft API that is currently unavailable, but will be preferred if available in the future.
```
iwannabethedev, line 125: I do not wish to butt in on the PR, but, short time-specifying additions to the documentation comments might be helpful. Like for *"A draft API that is currently **as of April 2023** unavailable, but will .."*, since "currently" depends on the current time, and specifying at which year it was written may help avoid having to look it up for any readers. In my personal projects I have added similar dates to specify the time when I wrote "currently" in comments.

web/packages/core/src/load-options.ts
```diff
@@ -117,6 +117,39 @@ export const enum WindowMode {
...
+    WebGpu = "webgpu",
+
+    /*
+     * The most featureful and currently preferred backend.
+     * Rendering is done the same way as in the desktop app, then translated to WebGL on-the-fly.
+     */
+    WgpuWebgl = "wgpu-webgl",
+
+    /*
+     * A vanilla WebGL backend. Was previously the default backend,
```
Dinnerbone, line 137: I'd change this to "Was the preferred backend until the switch to wgpu-webgl at the end of 2022"

Dinnerbone (changes requested):

web/packages/core/src/load-options.ts
```diff
@@ -117,6 +117,39 @@ export const enum WindowMode {
     Gpu = "gpu",
 }
 
+/**
+ * The render backend of a Ruffle player.
```
Dinnerbone, line 121: I'd probably put something here to the effect of "these options are non exhaustive and may change at any point"

web/packages/core/src/load-options.ts
```diff
@@ -319,6 +352,20 @@ export interface BaseLoadOptions {
      */
     playerVersion?: number | null;
 
+    /**
+     * The preferred render backend of the Ruffle player.
```
Dinnerbone, line 356: I'd add to this that a preference does not guarantee that specific backend, only request it if it's available.

Though, if we're using this as a testing feature, maybe changing that would actually be smart and having it be "force" instead of "prefer"; I know I've been caught a few times by using one backend instead of another when the preference wasn't doing what I expected during dev. This backend or error, so you know it's what you chose?

iwannabethedev: How easy is it to detect and determine at runtime which backend is actually used? I wonder if the current functionality could be kept, but then have both the "preferred" and the "actually used" backends being shown somewhere that is easy to access (I do not know where that "somewhere" would be, though).

beautiful_mspaint_illustration

Like in this image, where the two left-side labels do not change, and the two right-side labels change depending on which backend is preferred and which backend is currently used respectively.

I guess that this showing of "preferred" and "currently used" could come in a later PR, though.

iwannabethedev: I wonder whether a checkbox option could be added (either now or in a later PR), where you 'force' the backend used or 'give error' if the preferred backend is different from the currently used backend. Such that, in the regular behavior as it is implemented now, it is just preferred, but when that option is chosen, an error is given if the preferred backend is not the same as the currently used backend.

n0samu: > having it be "force" instead of "prefer"

Too late for that suggestion, sorry

> I guess that this showing of "preferred" and "currently used" could come in a later PR, though.

Yep and that PR won't be done by me 😛 

> I wonder whether a checkbox option could be added (either now or in a later PR), where you 'force' the backend used

Separate PR that will not be done by me

iwannabethedev: > Yep and that PR won't be done by me :stuck_out_tongue:

> Separate PR that will not be done by me

That is alright, sorry if it sounded like I implied that it should be you, I did not mean it like that. Piling substantially more work on you as a contributor would not be fair.

I am more interested if you think this proposed functionality might be doable (for instance by someone else). Because, if this proposed functionality can be done in a later PR, then that other functionality can always be decided upon later, for instance after this PR has been merged, and thus this discussion would not block merging this PR. I meant to be helpful, sorry if I came off as something else.

web/packages/core/src/load-options.ts
n0samu: Added the link. I will personally remind you guys to update the docs when we enable WebGPU.

web/packages/core/src/load-options.ts
n0samu: Actually the RenderBackend enum exhaustively lists all of Ruffle's web render backends. Added the "may change at any point" part though.

web/packages/core/src/load-options.ts
n0samu: No worries, I'm just a bit irritable about the whole process I've gone through to get this PR accepted lol.

Yes, a "force renderer or panic" config option would definitely be doable. Would just be an extra boolean that is checked in
web/src/lib.rs (fn create_renderer) and changes its behavior.

Dinnerbone (approved):

relrelb (approved):
@n0samu Sorry about conflicting with #10814. This PR looks good, I think it can be merged once rebased.

web/packages/core/src/load-options.ts
relrelb: @n0samu Wait, why not acutally implement "force renderer" in this PR? I agree with Dinnerbone that "preferred renderer" is not ideal for testing purposes.

adrian17: > Wait, why not acutally implement "force renderer" in this PR?

Because it's a separate functionality, with - in a way - completely separate philosophy from the original intent.

Can we please not restart this discussion this late in? It was basically hours from being merged after being stuck in limbo for over month.

n0samu: Main thing is it would be a lot of extra work and I'm not going to do it lol

n0samu:
Okay, rebased

web/packages/extension/src/content.ts
```diff
@@ -174,6 +174,7 @@ function isXMLDocument(): boolean {
             autoplay: options.autostart ? "on" : "auto",
             unmuteOverlay: options.autostart ? "hidden" : "visible",
             splashScreen: !options.autostart,
+            preferredRenderer: options.preferredRenderer,
```
relrelb, line 177: This line is no longer necessary due to `...options` above.

n0samu: Oh oops, I thought I removed that, one sec

Issue #9831, opened 2023-03-03
Diggy: Softlock after clicking Hall of Fame button


n0samu:
### Describe the bug

Link to game: https://www.kongregate.com/games/Vogd/diggy
To reproduce the bug:
1. Play through the first day of the game. (Takes less than a minute)
2. Start a new day.
3. Click the Hall of Fame button at the top of the screen.
4. The game will be softlocked; none of the buttons will respond to clicks.

![image](https://user-images.githubusercontent.com/71368227/222704429-3ee37eba-9473-420a-b28e-c1687b9231bc.png)

### Expected behavior

It should be possible to switch between the tabs of the Hall of Fame screen by clicking them, and to dismiss the screen by clicking the Hall of Fame button again.

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

Firefox 110

waspennator:
Can confirm issue is fixed by latest nightly.

Issue #9577, opened 2023-02-15
Accurate Slapshot: Unclickable buttons


n0samu:
### Describe the bug

Link to SWF: http://i.notdoppler.com/files/accurateslapshot.swf

In [Accurate Slapshot](http://www.notdoppler.com/accurateslapshot.php), various buttons are unclickable.
- After completing a level, the Main Menu, Replay, and Next Level buttons are clickable. But if you complete another level after that, the buttons will be unclickable this time.
- If you click the X button in the top-right corner of the gameplay screen, it will take you back to the level select screen, where all of the buttons will be unclickable.
![image](https://user-images.githubusercontent.com/71368227/218918084-32d295ae-f84b-4d99-b087-088e7efeff06.png)

### Expected behavior

The buttons on the level completion and level selection screens should always be clickable.

### Affected platform

Desktop app

### Operating system

Windows 10

rredford:
It worked for me, then on third level next buttons or really all buttons didn't work

Issue #9201, opened 2023-01-18
Papa games issues


MartySVK:
### Describe the bug

Since Papa games are very popular, I decided to make a special issue for them and test them regularly.

All games use similar logic and will also contain similar features. In addition to Papa games, Flipline Studios has similar games like Jacksmith.

Current issues:
- [ ] loading bar bug
- [ ] timer-volume  error
- [ ] addEventListener error
- [ ] maxChars error

From the discussion on discord, maxChars could be stubbed. Loading bar is probably just a visual issue and not every Papa game has this problem. The game is still loading, but you can click on "Loading" text and continue to the menu.

Error Logs:

`Error dispatching event EventObject(EventObject { type: "timer", class: flash.events::TimerEvent, ptr: 0x116ae670 }) to handler FunctionObject(FunctionObject { ptr: 0x11a02f10 }) : RustError("Cannot access property ::volume of null or undefined")`

`Error dispatching event EventObject(EventObject { type: "addedToStage", class: flash.events::Event, ptr: 0x1206cb48 }) to handler FunctionObject(FunctionObject { ptr: 0x12065480 }) : RustError("Cannot access property ::addEventListener of null or undefined")`

`Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x11f84410 }) to handler FunctionObject(FunctionObject { ptr: 0x11a1d620 }) : RustError("Cannot set undefined property ::maxChars on Some(flash.text::TextField)")`

### Expected behavior

Click the PLAY button to continue the game

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

Chrome 109

MartySVK:
> You should also mention the unclickable menu buttons in games like Papa's Burgeria on Flipline.com: https://www.flipline.com/games/papasburgeria/index.html
> 
> `ERROR core/src/avm2/events.rs:422 Error dispatching event EventObject(EventObject { type: "addedToStage", class: flash.events::Event, ptr: 0x1a24058 }) to handler FunctionObject(FunctionObject { ptr: 0x6d8bac8 }) : RustError("Cannot access property ::addEventListener of null or undefined")`

Actually, these kinds of issues are mostly connected with different issues. In this case, it is connected with Stage invalid. I tried @golfinq PR (https://github.com/ruffle-rs/ruffle/pull/8561) and you can click on the menu button, you can even reach the first customer 🥳. Hope, it will be merged soon

[...]

PR #9911, opened 2023-03-07
core: Only apply 'unload' and 'removed' logic to AVM1

Aaron1011:
I've also renamed these methods to 'avm1_unload' and 'avm1_removed', to make it clear that they don't
apply to AVM2.

This was causing us to incorrectly skip mouse picks, and remove masks.


core/src/display_object/avm2_button.rs
```diff
@@ -331,7 +331,9 @@ impl<'gc> Avm2Button<'gc> {
         }
 
         if let Some(old_state_child) = old_state_child {
-            old_state_child.unload(context);
+            if !context.is_action_script_3() {
```
adrian17, line 334: Is it actually possible to construct an avm2button from an as2 swf?

Aaron1011: Good point - I've removed all of the AVM1 logic from this file.

n0samu:
This fixes #9831, #9577, and https://github.com/ruffle-rs/ruffle/issues/9201#issuecomment-1423593609 🎉 🎉 🎉

adrian17 (approved):

PR #9913, opened 2023-03-07
avm2: Implement avmplus.getQualifiedClassName

Aaron1011:
This appears to be the same function as
'flash.utils.getQualifiedClassName'


Bale001 (approved):

core/src/avm2/globals/avmplus.rs
```diff
@@ -0,0 +1 @@
+pub use crate::avm2::globals::flash::utils::get_qualified_class_name;
```
adrian17, line 1: I'd consider just implementing it in AS (by calling the flash.utils one) instead of in Rust, but this is cool too :D

adrian17 (approved):

PR #9914, opened 2023-03-07
Add basic color conversion tests for video decoders

torokati44:
This should help with #9753, and future work building on it (color conversion in hardware).

TODO:
 - [ ] Make building the decoder crate optional somehow, only for tests that need it
 - [x] Document how these SWFs came to be
 - [ ] Add more test cases? (odd frame sizes, plus highlight chroma subsampling)
 
The test card image is from https://pixabay.com/vectors/tv-test-pattern-test-picture-display-146649/, they claim "no attribution required", see the license here: https://pixabay.com/service/license/


Dinnerbone:
> Make building the decoder crate optional somehow, only for tests that need it
I think we can bundle it with the imgtests feature maybe?

We will need to find a better way to build and filter tests soon 😅

torokati44:
> I think we can bundle it with the imgtests feature maybe?

For now, that should be still better than nothing, but even the majority of image tests don't need it...

> We will need to find a better way to build and filter tests soon :sweat_smile:

I guess... :sweat_smile:

Dinnerbone:
Ohh I thought you meant build as in the crate. Lock the dependency behind imgtests and then add a field to player_options to enable the backend, defaulting to false

torokati44:
Okay, I may have gone a bit overboard with the notes, but... How does this look now?

Dinnerbone:
Looks good, two small things:
- Can you make these tests require render backend? It's optional right now but I'm not sure they make sense without it
- Can you document the new with_video option in the tests readme?

Thanks!

torokati44:
Sure, thanks! Maybe not today though...
I also found some wording problems in NOTES since.

torokati44:
Okay, done! While the default for `required` is `false`, I left it in there to be explicit.

Dinnerbone (approved):

torokati44:
Thank you for the merge! I've updated the OP regarding the remaining unchecked TODOs.

PR #9919, opened 2023-03-08
avm2: Add Mutex.isSupported property

Bale001:
I missed this property because it isn't in the AS3 docs


Dinnerbone:
> I missed this property because it isn't in the AS3 docs

A tale as old as time :D At least we have a list of things now!

Dinnerbone (approved):

PR #9928, opened 2023-03-08
avm2: Improve handling of timeline-removed orphans

Aaron1011:
When a DisplayObject is removed from its parent by a RemoveTag, it still runs its framescript for the current frame (but with 'this.parent == null'). It then stops executing entirely, unlike ActionScript-removed orphans, which continue to execute indefinitely.

Additionally, objects created by ActionScript during a frame skip their next 'enterFrame' logic (but still receive an enterFrame event). This results in the currentFrame lagging one frame behind objects that were placed by the timeline during the same frame.

The combination of these two changes lets us greatly simplify frame lifecycle handling for orphan movies. Most of the orphan stages were unencessary, and the remaining ones run in the same phase as the normal Stage-descendant objects.


n0samu:
Can confirm that this fixes the repeating intro sounds in Pursuit of Hat and Epos!

core/src/frame_lifecycle.rs
```diff
@@ -83,70 +82,41 @@ pub fn run_all_phases_avm2(context: &mut UpdateContext<'_, '_>) {
...
+    stage.construct_frame(context);
     stage.frame_constructed(context);
 
     *context.frame_phase = FramePhase::Update;
-    stage.run_frame_avm2(context);
-
     Avm2::each_orphan_movie(context, |movie, context| {
         movie.run_frame_avm2(context);
     });
+    stage.run_frame_avm2(context);
```
Lord-McSweeney, line 101: Is the change in ordering between orphans and the stage correct?

Aaron1011: This method doesn't actually do anything, and will be removed in https://github.com/ruffle-rs/ruffle/pull/9944

core/src/frame_lifecycle.rs
```diff
@@ -73,63 +72,35 @@ pub fn run_all_phases_avm2(context: &mut UpdateContext<'_, '_>) {
...
     });
+    stage.run_frame_scripts(context);
 
     *context.frame_phase = FramePhase::Exit;
-    stage.exit_frame(context);
-
     Avm2::each_orphan_movie(context, |movie, context| {
         movie.on_exit_frame(context);
     });
+    stage.exit_frame(context);
```
Lord-McSweeney, line 97: Is the changed order of `exit_frame` compared to orphan movies' `exit_frame` correct? (this also applies to `run_frame_scripts`)

adrian17 (approved):

core/src/frame_lifecycle.rs
Aaron1011: I don't think the order is currently observable - all of the tests pass with this code.

kmeisthax (approved):
Looks good to me.

I'm wondering if the orphan behavior and the "skip enter frame" flag might also explain some of the weirdness with `SimpleButton` construction - i.e. the same mechanism that causes `enterFrame` to skip might also decide if it constructs late.

PR #7895, opened 2022-09-06
avm2: Add `#[native]` macro to allow defining methods with native types

Aaron1011:
This allows you to define a native method like:

```rust
#[native]
fn my_method(activation: &mut Activation, this: DisplayObject<'gc>, arg:
bool)
```

instead of needing to use `Option<Object<'gc>>` and `&[Value<'gc>]`
and manual coercions.

See the doc comments for more details.

PR #9930, opened 2023-03-08
avm2: Introduce ParametersExt and use it inside bitmapdata

Dinnerbone:
We currently have so many different ways of pulling out parameters from the function, and almost every single one is wrong. The "correct" way to do it tends to be extremely verbose.

This introduces common methods for **correctly** getting the value of a given type from the function parameters slice, and converts BitmapData to use it as an initial proof of concept.

It's still missing more useful functions like getting native object types (DisplayObject from an arg, for example), but it's already a significant improvement.

### DIFF
--- a/core/src/avm2/globals/flash/display/bitmap_data.rs
+++ b/core/src/avm2/globals/flash/display/bitmap_data.rs
@@ -20,6 +20,7 @@ use ruffle_render::transform::Transform;
 use std::str::FromStr;
 
 pub use crate::avm2::object::bitmap_data_allocator;
+use crate::avm2::parameters::ParametersExt;
 
 /// Copy the static data from a given Bitmap into a new BitmapData.
 ///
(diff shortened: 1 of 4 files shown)


core/src/avm2/globals/flash/display/bitmap_data.rs
```diff
@@ -85,23 +86,10 @@ pub fn init<'gc>(
...
-                    .coerce_to_boolean();
-                let fill_color = if let Some(value) = args.get(3) {
-                    value.coerce_to_u32(activation)?
-                } else {
-                    0xFFFFFFFF
-                };
+                let width = args.get_u32(activation, 0)?;
+                let height = args.get_u32(activation, 1)?;
+                let transparency = args.get_bool(2);
+                let fill_color = args.get_u32(activation, 3)?;
```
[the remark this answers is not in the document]
Dinnerbone: The default is provided from AS, that's a holdover from before this was converted

Lord-McSweeney:
Does this PR obsolete/supersede #7895?

Dinnerbone:
I'd say no. That PR is on hold until everyone can agree on the design, and I think makes a nicer end goal, but this works for today.

Aaron1011 (approved):

core/src/avm2/parameters.rs
```diff
@@ -0,0 +1,176 @@
...
+    fn try_get_object(
+        &self,
+        activation: &mut Activation<'_, 'gc>,
+        index: usize,
+    ) -> Option<Object<'gc>> {
+        match self[index] {
+            Value::Null | Value::Undefined => None,
+            Value::Object(o) => Some(o),
+            primitive => Some(
+                PrimitiveObject::from_primitive(primitive, activation)
```
adrian17, line 112: in most places we don't actually expect a primitive (matrix, color_transform, etc), so I'd consider making it a default? (just documenting it well).

On that note, I'd also add `get_value(idx)` just for a consistent api.

adrian17 (approved):

core/src/avm2/parameters.rs
Dinnerbone: Added get_value which is `this[index]` for now.
Good point about non-primitives, I think I'll need to consider that a bit though. Will handle it in a future PR if that's ok!

PR #9936, opened 2023-03-09
render: Make retrieve_offscreen_texture pass the raw buffer

Dinnerbone:
Previously we'd turn it into a vec and then put that in a struct and then copy that over manually to Pixels.
Now we just read straight from the buffer directly.

### DIFF
--- a/render/wgpu/src/lib.rs
+++ b/render/wgpu/src/lib.rs
@@ -8,13 +8,13 @@ use crate::pipelines::Pipelines;
 use crate::target::{RenderTarget, SwapChainTarget};
 use crate::uniform_buffer::UniformBuffer;
 use crate::utils::{
-    buffer_to_image, create_buffer_with_data, format_list, get_backend_names, BufferDimensions,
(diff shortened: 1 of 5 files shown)


render/wgpu/src/lib.rs
```diff
@@ -186,33 +185,27 @@ pub enum QueueSyncHandle {
...
-        let image = self.capture();
-        Ok(Bitmap::new(
-            image.dimensions().0,
-            image.dimensions().1,
-            ruffle_render::bitmap::BitmapFormat::Rgba,
-            image.into_raw(),
-        ))
+        with_rgba: RgbaBufRead,
+    ) -> Result<(), ruffle_render::error::Error> {
+        self.capture(|buf, buf_width| with_rgba(buf, buf_width));
```
Aaron1011, line 190: Can this be just `self.capture(with_rgba`)` ?

Aaron1011 (approved):

render/wgpu/src/lib.rs
Dinnerbone: Fixed, good catch!

PR #9949, opened 2023-03-10
avm2: Stub Stage.fullScreenWidth and fullScreenHeight

### LINKED ISSUE #9474: cat ninja not launching
i am aware that actionscript 3 content is not currently supported but I would like for it to be fixed. the problem is that it does not get past the loading screen. i am asking for a fix because it is a very simple game. thank you for your time.

n0samu:
Fixes #9474.

### DIFF
--- a/core/src/avm2/globals/flash/display/Stage.as
+++ b/core/src/avm2/globals/flash/display/Stage.as
@@ -130,8 +130,12 @@ package flash.display {
         public native function get frameRate():Number;
         public native function set frameRate(value:Number):void;
 
+        public native function get fullScreenHeight():uint;
+
         public native function get fullScreenSourceRect():Rectangle;
         public native function set fullScreenSourceRect(value:Rectangle):void;
+
+        public native function get fullScreenWidth():uint;
 
         public native function get scaleMode():String;
         public native function set scaleMode(value:String):void;
--- a/core/src/avm2/globals/flash/display/stage.rs
+++ b/core/src/avm2/globals/flash/display/stage.rs
@@ -522,3 +522,23 @@ pub fn set_full_screen_source_rect<'gc>(
     avm2_stub_setter!(activation, "flash.display.Stage", "fullScreenSourceRect");
     Ok(Value::Undefined)
 }
+
+/// Stage.fullScreenHeight's getter
+pub fn get_full_screen_height<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    _this: Option<Object<'gc>>,
+    _args: &[Value<'gc>],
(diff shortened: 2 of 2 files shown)


core/src/avm2/globals/flash/display/stage.rs
```diff
@@ -522,3 +522,23 @@ pub fn set_full_screen_source_rect<'gc>(
...
 }
+
+/// Stage.fullScreenHeight's setter
+pub fn get_full_screen_height<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    _this: Option<Object<'gc>>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    avm2_stub_getter!(activation, "flash.display.Stage", "fullScreenHeight");
+    Ok(0.into())
```
Lord-McSweeney, line 533: I suggest adding reasonable values here (1024x768 or 1920x1080), as this may control scaling on some games.

Lord-McSweeney (approved):

Cargo.lock
```diff
@@ -2774,9 +2774,9 @@ dependencies = [
 
 [[package]]
 name = "orbclient"
-version = "0.3.42"
+version = "0.3.43"
```
Lord-McSweeney, line 2777: Was this change intentional?

n0samu: No, that was an accident. Fixed.

core/src/avm2/globals/flash/display/stage.rs
```diff
@@ -522,3 +522,23 @@ pub fn set_full_screen_source_rect<'gc>(
     avm2_stub_setter!(activation, "flash.display.Stage", "fullScreenSourceRect");
     Ok(Value::Undefined)
 }
+
+/// Stage.fullScreenHeight's setter
```
torokati44, line 526: nit: "getter" in comment - also for the other method

torokati44 (approved):

core/src/avm2/globals/flash/display/stage.rs
n0samu: Not sure how I managed to miss that 😬

core/src/avm2/globals/flash/display/Stage.as
```diff
@@ -130,9 +130,13 @@ package flash.display {
         public native function get frameRate():Number;
         public native function set frameRate(value:Number):void;
 
+        public native function get fullScreenHeight():uint;
```
adrian17, line 133: for the future, it's alright to implement the stub in pure AS too :)

adrian17 (approved):

PR #9980, opened 2023-03-10
video/h263: Add a deblocking filter as a postprocessing step

torokati44:
This tries to alleviate https://github.com/ruffle-rs/h263-rs/issues/11.
I'm not sure if this code should be in that repo, or here, but it's not much, so maybe it's not so important.

Note that for comparison purposes, this code as is only applies this to the right half of the frames.

The algorithm was taken from section J.3 of ITU-T Rec. H.263 (01/2005) (https://www.itu.int/rec/T-REC-H.263-200501-I/en), just like the rest of the decoder.
Even though Flash (FLV1, Sorenson Spark, whatever) does not implement Annex J - where this is used as an in-loop filter - I thought the same idea could work well as a post-processing step. And it indeed appears so.
While in Annex J, the strength is controlled at the block-level by the quantization level, here it doesn't make sense, as the result of any deblocking applied to previous frames is not present, and we're not only applying this to "newly added" blocks.

~~TODO: Only use this if the video actually asked for it in the relevant flag, and investigate if  the current quality affects its use.~~ The deblocking flag is now taken into account.

Some z0r loop IDs that can be used for testing (as usual, some are disturbing/somewhat NSFW - I've marked some of the milder ones):

363 <- SFW
1109
1218
1277
1322
2143
2387
2812
3030
3703
4868
5282
5591 <- SFW
5927
7895 <- SFW
7980
7996 <- SFW
7999


torokati44:
I've moved the actual processing into `h263-rs` (https://github.com/ruffle-rs/h263-rs/pull/49), as I feel like it belongs there more.

torokati44:
With a visual test added, the only reason now that this is still a draft is that it refers to my `h263-rs` fork.

video/software/src/decoder/h263.rs
```diff
@@ -70,21 +75,49 @@ impl VideoDecoder for H263Decoder {
...
-        ))
+        let hdr = picture.as_header();
+
+        if self.1 == VideoDeblocking::Level1
+            || (self.1 == VideoDeblocking::UseVideoPacketValue
+                && hdr.options.contains(PictureOption::USE_DEBLOCKER))
+        {
+            let chroma_width = picture.chroma_samples_per_row();
+            let quantizer = hdr.quantizer as usize;
+            let strength = QUANT_TO_STRENGTH[quantizer];
```
Dinnerbone, line 86: I think it's *theoretically possible* for `quantizer` to be out of bounds, but that's more a bug inside the decoder than here

Dinnerbone (approved):

Issue #1955, opened 2020-12-13
Extremely thin/zoomed out lines look much thinner in Ruffle than they do in flash player.


MrCheeze:
There seems to be some subtle rendering difference when drawing extremely small lines. I have a screenshot and demo of the issue:

https://i.imgur.com/fffKIaX.png
https://mrcheeze.github.io/detritus/ruffle/zoomout.html

The beginning of this flash contains many lines that are shrunk to be extremely small (you can zoom in with flash player to see their true shape). When rendered in Ruffle, the lines are considerably smaller than a pixel, and so are on the verge of being totally invisible. They also flicker strongly when animated to move around.

In flash player, on the other hand, they render surprisingly thick - it seems to me like they always render at least 1 pixel wide when set to Low quality, half a pixel wide when set to Medium quality, and some other finer fraction of a pixel wide when set to High quality. In all cases, the end result ends up much thicker than what Ruffle itself draws.

I'm not sure if this difference in rendering behaviour is specific to only lines, or reflects some deeper difference between how rendering works between flash player and ruffle.

MrCheeze:
Done some searching through other issues, this seems to be related to the issue with "hairline" strokes discussed here: https://github.com/ruffle-rs/ruffle/issues/751#issuecomment-648412873

If my understanding is correct, stroke width is clamped to at least 1px _before_ scaling in Ruffle, but _after_ scaling is applied in Flash Player. This seems to match what I'm seeing in the demo above.

(And I think I was mistaken about quality being relevant here, the line itself is 1px in flash player no matter what. It just also has antialiasing applied if the quality is something other than low.)

ousia:
@MrCheeze, screenshot attached here:

![fffKIaX jpg](https://user-images.githubusercontent.com/203767/106101028-983d8500-613d-11eb-94a1-231252df6522.png)

MrCheeze:
Although this is still broken in the wgpu and webgl render backends, which means it's still broken by default on both desktop and web, it was actually fixed for only the canvas backend in https://github.com/ruffle-rs/ruffle/pull/7042 :

![image](https://user-images.githubusercontent.com/6541413/212784011-8ffacccc-5035-438d-89c4-8b0d2b18426a.png)

Based on Herschel's comments in that PR, the plan is to eventually make that fix available to the other render backends. But according to n0samu's comments in https://github.com/ruffle-rs/ruffle/issues/9044 , that may not happen any time soon.

[...]

PR #9981, opened 2023-03-11
Fix stroke widths and hairline strokes in wgpu and webgl

### LINKED ISSUE #751: [Mini-Putt 2] Differences with the Flash version
Flash: https://www.newgrounds.com/portal/view/38598/
Ruffle: https://www.newgrounds.com/portal/view/38598/format/flash?emulate=flash

**Graphics**

There are some differences with the Flash version. This is particularly noticeable on the title screen.

Flash:
1flash
Ruffle:
1ruffle

Flash:
2flash
Ruffle (borders are a bit blurry):
2ruffle

Flash:
3flash
Ruffle:
3ruffle

**Gameplay**

- (FIXED) ~~Collisions in triangle corners are different to the Flash version (the ball usually stops immediately after hitting it).~~

- (FIXED) ~~On level 2 (and on levels with similar gameplay), after you have put the ball in one of the three holes at the top, the ball will get out of the pipe way too fast.~~

**Other**

- When a level starts, the cursor sprite is always the handpoint until you drop the ball. On Ruffle, you only have that cursor in the area where you can drop the ball, otherwise you have the default one.

- (FIXED) ~~Music does not autoplay (already reported, see https://github.com/ruffle-rs/ruffle/issues/395).~~

Dinnerbone:
This fixes #751 (and probably a lot of other issues that should be tested)

This will come at a performance cost, but the size of that isn't known yet (I can't benchmark right now, PC is unhappy with me).


MrCheeze:
Can confirm it fixes https://github.com/ruffle-rs/ruffle/issues/1955 !

SN902:
Problems with outlines in next games:
 
[Rainbow Dash Shooter Project.zip](https://github.com/ruffle-rs/ruffle/files/10947421/Rainbow.Dash.Shooter.Project.zip)
[sponge_fall_game.zip](https://github.com/ruffle-rs/ruffle/files/10947423/sponge_fall_game.zip)

Toad06:
Thanks a lot for your work on this. 😉

This PR is a great improvement for all of the Steppenwolf games, for instance [Steppenwolf 2-1](https://www.y8.com/games/steppenwolf_chapter_2_episode_1). There are still some issues that I'm mentioning below.

---

Regression on the canvas renderer - looks wild now:
canvas

---

This one is less important but strokes look a bit smoother in Flash Player and in Nightly Ruffle using the canvas renderer:

fp
rpr
rcan

torokati44:
> This one is less important but strokes look a bit smoother in Flash Player and in Nightly Ruffle using the canvas renderer:
What's strange is that the overhead baggage shelf is missing on canvas backend... :D

Toad06:
> What's strange is that the overhead baggage shelf is missing on canvas backend... :D

Yeah, I noticed that too while posting, seems to be a very old bug (I tried with a release from November 2020 and the issue was already there).

Herschel:
Thanks so much for your work on this!

Some thoughts so far:

 * The code is assuming that all fills are drawn first, and then strokes, but this isn't necessarily the case. A ChangeStyle record with new styles basically indicates a "flush", where new fills will draw on top of previous strokes. This is how the Flash IDE exports raw vector art containing layers.
 
[Example SWF](https://github.com/ruffle-rs/ruffle/files/10986280/layer.zip)
FP vs this PR:
image

So the idea of splitting into one stroke part and one fill part doesn't handle all cases; you'd need to at least allow for multiple stroke shapes between multiple fill shapes.

* Because only strokes get re-tessellated, this can produce some unsightly gaps between the stroke and a jaggy fill:

[Example SWF](https://github.com/ruffle-rs/ruffle/files/10986310/jaggies.zip)
FP vs this PR
image

Naturally this is a pathological case, but just pointing out the obvious.

* In a perfect world we would also re-tessellate fills, to avoid jaggies on zoom even for stroke-less shapes. Obviously, this is even huger performance hazard, but I believe we'll need to it at some point. Instead of of "re-tessellate anytime scale changes" like strokes, I picture a stroke-less shape could be more coarse and only re-tessellate on some large threshold of change. And I'm sure this PR is only concerned about strokes at the moment.

* `RenderBackend::register_shape_fills`/`register_stroke_fills` seems like leaking implementation details into the renderer API. In a perfect world, core wouldn't care about retessellating at all, and only the tessellation backend itself would manage this, but I understand that this is probably the only sane way to do this at the moment, and we are focusing on WGPU for the near future. I point it out because a Vello/other-fancy-vector-rendering is always in the back of my mind.

* Given all of the above, I think I'd want to edge a bit more towards an API that is "re-tessellate the entire shape" vs. "re-tessellate strokes and shapes separately", since we'd want to re-tessellate fills eventually, and it seems a little simpler to manage assuming the performance isn't awful. This would help us deal with both the layered-stroke and the jaggy-fills cases above. But I could be wrong here, and simply allowing for multiple stroke-layers in the shape is good enough.

This is PR is a huge improvement and the above is me thinking out loud. Sorry if this has already been discussed ad nauseum on Discord, and I'm just out of the loop or completely off-base. 😄

Dinnerbone:
@Herschel yourr absolutely right and I realized that a while ago which is why I put this back to draft, sorry to make you type all that out 😅 I'm going to redo this with lessons learnt shortly, but trying to make more tests too.

Most likely it's going to redo the entire shape but only if there's non scaling strokes involved, otherwise the shape can stay static. And it still only needs to redo the shape once a scale changed, I believe. I think there's room to optimise further after that but we need a lot more tests before I'd feel comfortable, and that feels right to me for a basic start

Herschel:
> Most likely it's going to redo the entire shape but only if there's non scaling strokes involved, otherwise the shape can stay static.

This should probably also consider hairline strokes as non-scaling, as these act like non-scaling until the shape is scaled >20x, so I think they'd show the above artifacts as well.

Here's a good worst-case example for testing scaled strokes and shapes: Dad 'n Me used some weirdo optimizer to scale the assets way down and then back up, so huge strokes and jaggies in shapes will be very obvious here. (Tom has since fixed this in the SWF on Newgrounds, but this is the original SWF.) [DadNMeNg.zip](https://github.com/ruffle-rs/ruffle/files/10994645/DadNMeNg.zip)

torokati44:
Has this been completely superseded by https://github.com/ruffle-rs/ruffle/pull/23011?

Lord-McSweeney:
Superseded by #23011, thank you for your work on this!

Issue #9965, opened 2023-03-10
Resort Empire: Cannot create property alwaysShowSelection on flash.text.TextField


MartySVK:
### Describe the bug

The game is stuck on the menu scene.

URL - https://www.gameflare.com/online-game/resort-empire/
SWF - https://data.gameflare.com/games/3705/tQSKbxkA1IgxKe.swf

error log:
```
AVM2 error: Object(ErrorObject(ErrorObject { class: ReferenceError, ptr: 0x2a09598 }))
core.ruffle.3f25a242b7e474a088e3.js:1 ERROR core/src/avm2/activation.rs:942
ReferenceError: Error #1056: Cannot create property alwaysShowSelection on flash.text.TextField.
```

### Expected behavior

Click on PLAY

### Affected platform

Desktop app

### Operating system

Windows 10

PR #9990, opened 2023-03-11
avm2: Stub TextField.alwaysShowSelection

MartySVK:
progresses Resort Empire https://github.com/ruffle-rs/ruffle/issues/9965
now it is blocked by unimplemented BitmapData.threshold

### DIFF
--- a/core/src/avm2/globals/flash/text/TextField.as
+++ b/core/src/avm2/globals/flash/text/TextField.as
@@ -11,6 +11,9 @@ package flash.text {
 
         private native function init();
         
+        public native function get alwaysShowSelection():Boolean;
+        public native function set alwaysShowSelection(value:Boolean):void;
+
         public native function get autoSize():String;
         public native function set autoSize(value:String):void;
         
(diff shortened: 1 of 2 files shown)


core/src/avm2/globals/flash/text/TextField.as
```diff
@@ -11,6 +11,9 @@ package flash.text {
 
         private native function init();
         
+        public native function get alwaysShowSelection():Boolean;
```
adrian17, line 14: For the future, stubs can be written in pure AS too, without adding Rust code.

adrian17 (approved):

core/src/avm2/globals/flash/text/text_field.rs
```diff
@@ -32,6 +32,24 @@ pub fn init<'gc>(
     Ok(Value::Undefined)
 }
 
+pub fn get_always_show_selection<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    _this: Option<Object<'gc>>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    avm2_stub_getter!(activation, "flash.text.TextField", "alwaysShowSelection");
+    Ok(Value::Bool(false))
```
Lord-McSweeney, line 41: This can also be accomplished using `false.into()` rather than `Value::Bool(false)`.

PR #9995, opened 2023-03-11
Support click events on movie clips containing text fields

jayrlsw:
Remove code that prevents click events from working on text fields within movie clips


Aaron1011:
We should test that the SWFs that originally required this are still working (I suspect they needed mouseEnabled/mouseChildren to be implemented) : https://github.com/ruffle-rs/ruffle/pull/8744

jayrlsw:
> We should test that the SWFs that originally required this are still working (I suspect they needed mouseEnabled/mouseChildren to be implemented) : #8744

I have gone through all of the SWFs mentioned in that issue. I can confirm that the problems are still resolved after this change.

Aaron1011 (approved):

Lord-McSweeney (approved):

Lord-McSweeney:
As far as I know, the code that this PR removes was a hacky solution to fix some games that used `mouseChildren`, which was not yet implemented.

Issue #3294, opened 2021-02-17
Error on https://funkypotato.com/rollercoaster-creator-2/


perelowski:
# Error Info
Error name: Error
Error message: panicked at 'already borrowed: BorrowMutError', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/gc-arena-0.2.0/src/gc_cell.rs:62:29
Error stack:
```
Error: panicked at 'already borrowed: BorrowMutError', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/gc-arena-0.2.0/src/gc_cell.rs:62:29
    at c.wbg.__wbg_new_df6e6ab7a65c4c4d (<anonymous>:3:6695)
    at std::sync::once::Once::call_once::{{closure}}::hce6501f6f20767e4 (chrome-extension://cfffheakccjmlhnfhpdbdnbhdaeokmhb/dist/90797cff9fcf5439a21889280011890d.wasm:wasm-function[1327]:0x26ba91)
    at std::sync::once::Once::call_inner::hd3d269a4946f2df2 (chrome-extension://cfffheakccjmlhnfhpdbdnbhdaeokmhb/dist/90797cff9fcf5439a21889280011890d.wasm:wasm-function[1604]:0x29efac)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::h6ed7686ed76cb258 (chrome-extension://cfffheakccjmlhnfhpdbdnbhdaeokmhb/dist/90797cff9fcf5439a21889280011890d.wasm:wasm-function[4639]:0x382410)
    at std::panicking::rust_panic_with_hook::h4f753dc70b771d8e (chrome-extension://cfffheakccjmlhnfhpdbdnbhdaeokmhb/dist/90797cff9fcf5439a21889280011890d.wasm:wasm-function[3509]:0x357ce4)
    at std::panicking::begin_panic_handler::{{closure}}::h5c9fc29957309ae8 (chrome-extension://cfffheakccjmlhnfhpdbdnbhdaeokmhb/dist/90797cff9fcf5439a21889280011890d.wasm:wasm-function[4146]:0x37597c)
    at std::sys_common::backtrace::__rust_end_short_backtrace::h92808941d97e9b21 (chrome-extension://cfffheakccjmlhnfhpdbdnbhdaeokmhb/dist/90797cff9fcf5439a21889280011890d.wasm:wasm-function[4985]:0x387ae1)
[...]

PR #10004, opened 2023-03-11
Partly implement XML child and elements

evilpie:
This makes #3294 (rollercoaster-creator-2) fully playable.

Missing is any (*) matching for child()/elements() and the existing attributes() method. Also missing is support for number indexes with child().


core/src/avm2/globals/xml.rs
```diff
@@ -88,6 +88,49 @@ pub fn to_xml_string<'gc>(
...
+}
+
+pub fn child<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Option<Object<'gc>>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let xml = this.unwrap().as_xml_object().unwrap();
+    let multiname = name_to_multiname(activation, &args[0])?;
+    let children = if let E4XNodeKind::Element { children, .. } = &*xml.node().kind() {
```
Aaron1011, line 121: Could you add a FIXME here about checking when the name is an index?

tests/tests/swfs/avm2/xml_child/Test.as
```diff
@@ -0,0 +1,27 @@
+package {
+  import flash.display.Sprite;
+  public class Test extends Sprite { }
+}
+
+var xml = new XML("<x><foo>foo1</foo><bar>bar1</bar><foo>foo2</foo></x>")
```
Aaron1011, line 6: Could you also test `elements` in this file, or in a new test?

core/src/avm2/globals/xml.rs
```diff
@@ -88,6 +88,50 @@ pub fn to_xml_string<'gc>(
...
+}
+
+pub fn child<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Option<Object<'gc>>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let xml = this.unwrap().as_xml_object().unwrap();
+    let multiname = name_to_multiname(activation, &args[0])?;
+    // FIXME: Support numerical indexes.
```
Lord-McSweeney, line 121: I think this should also mention calling `XML.child` with any ("*") as the parameter

Aaron1011 (approved):

core/src/avm2/globals/xml.rs
Aaron1011: This is being addressed in a separate PR

Issue #9501, opened 2023-02-10
Numrows: The numbered keys on the keypad can only give "0" as an output


WyvernB:
### Describe the bug

Tested on Ruffle Nightly 2023-02-10
Numrows SWF used: [567278_numrows202c.zip](https://github.com/ruffle-rs/ruffle/files/10708065/567278_numrows202c.zip)

This is what the game looks like in the Ruffle Demo.

https://user-images.githubusercontent.com/36887711/218110302-64b6adc6-db42-4aa8-8f2d-1460ec10535e.mp4

### Expected behavior

This is what the game looks like when I launch it using Flashpoint.

https://user-images.githubusercontent.com/36887711/218110384-b298c1f8-e30a-408b-8e60-cc8d717e77b8.mp4

### Affected platform

Online demo

### Operating system

Windows 10

### Browser

Chrome 109

n0samu:
This is a similar problem to #9287. Here is the decrypted SWF with some trace statements added: [numrows_decrypted.zip](https://github.com/ruffle-rs/ruffle/files/10712717/numrows_decrypted.zip)

Here is the function that determines what number gets entered into the keypad when you push a key:
```as
private function funkceTlacitka(param1:MouseEvent) : void
{
    MovieClip(parent).playSound(MovieClip(parent).klikSound);
    param1.target.parent.gotoAndStop(3);
    this.pridejHodnotu(int(param1.target.text));
}
```
In Flash Player, `trace(param1.target)` outputs `[object TextField]` and `trace(param1.target.text)` outputs the number inside the key's text field. 
But in Ruffle, `trace(param1.target)` outputs `[object NumlockButton]` and `trace(param1.target.text)` outputs `undefined`. That means the button is trapping the mouse event instead of allowing it to propagate down to the text field within it. And `int(undefined)` becomes `0`, hence 0 is always entered into the keypad.

Lord-McSweeney:
Fixed by #10005.

PR #10005, opened 2023-03-12
Enable click events on non-selectable text fields

jayrlsw:
This change allows non-selectable text fields to trigger their click events. It also changes the cursor to be an arrow when the text field is not selectable.


n0samu:
Hmm, looks like these regresses the click_block test. The behavior of click events on text fields is a bit complex and we've had a really hard time figuring it out. This will take a bit of testing and tuning to get right!

n0samu (approved):
This looks perfect now, and it fixes #9501, thank you! 😃

Dinnerbone (approved):

PR #10030, opened 2023-03-12
chore: Upgrade Node.js dependencies

tsunamistate:
This is basically me doing Renovate's job, but with the added benefit of manual testing!
https://github.com/ruffle-rs/ruffle/pull/9878 only upgraded major versions and missed a lot of patch upgrades, and I just couldn't wait for a month

✋ I solemnly promise I will only do this once, and trust Renovate bot in further version upgrades


Dinnerbone:
It only missed them because the versions already allow for minor/patch updates through update, which is happening in less than 12 hours, no?

Regardless, thanks 😅

Dinnerbone (approved):

PR #10031, opened 2023-03-13
core/avm2: Use an actual FFT library in `computeSpectrum`

torokati44:
This should make it _even faster_ than #9657.
Now I'm really glad for #9561!
Benchmarking to be done.
Thanks to https://github.com/ejmahler/rust_dct/issues/13#issuecomment-1465340496!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3168,6 +3168,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "realfft"
+version = "3.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "93d6b8e8f0c6d2234aa58048d7290c60bf92cd36fd2888cd8331c66ad4f2e1d2"
+dependencies = [
+ "rustfft",
+]
+
+[[package]]
 name = "redox_syscall"
 version = "0.2.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3306,9 +3315,11 @@ dependencies = [
  "nellymoser-rs",
  "num-derive",
  "num-traits",
+ "once_cell",
  "percent-encoding",
  "quick-xml",
  "rand",
+ "realfft",
  "regress",
  "ruffle_macros",
  "ruffle_render",
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -49,6 +49,8 @@ static_assertions = "1.1.0"
 rustversion = "1.0.12"
 bytemuck = "1.13.1"
 clap = { version = "4.1.8", features = ["derive"], optional=true }
+realfft = "3.2.0"
+once_cell = "1.8.0"
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies.futures]
 version = "0.3.26"
--- a/core/src/avm2/globals/flash/media/soundmixer.rs
+++ b/core/src/avm2/globals/flash/media/soundmixer.rs
@@ -1,6 +1,7 @@
 //! `flash.media.SoundMixer` builtin/prototype
 
 use std::cell::RefMut;
+use std::sync::Arc;
 
 use crate::avm2::activation::Activation;
 use crate::avm2::bytearray::ByteArrayStorage;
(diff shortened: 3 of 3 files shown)


Dinnerbone:
Do you have a movie I can use to compare performance with?

torokati44:
I mostly use the ones linked in #9657:
- https://z0r.de/L/z0r-de_7812.swf
- https://z0r.de/L/z0r-de_7855.swf

Dinnerbone:
Benchmarked 7855 on Windows/Vulkan/Release:

`compute_spectrum` is about 82% faster.
image

`run_all_phases_avm2` is about 7% faster.
image

`run_frame` is about 7% faster.
image

torokati44:
Thank you for the benchmarking @Dinnerbone! I hope to do a quick second pass at this sometime soon, and I think it'll be good to go.

torokati44:
I saw similar speedups on these loops in Firefox as well (80-98% in `computeSpectrum` itself).
I switched the input filling loop from direct indexing to iterator-based.
While this looks uglier, it's supposed to be more idiomatic Rust, and could potentionally be faster due to fewer range checks?

torokati44:
Oh I guess I should probably do the same on the output side as well...

torokati44:
Okay, done!

ejmahler:
I have one (optional) suggestion for how to improve this: The `RealFftPlanner` caches a lot of setup, and when it gets dropped at the end of the function, that setup gets dropped too, so the next time `compute_spectrum` is called, it has to rebuild that setup.

If you have some global state cache (the `Activation` struct seems like this might be it, although idk enough about ruffle to know if that's an appropriate place to put it), you could store the planner in that local state, meaning the next time you plan a FFt with the global planner, you reuse the setup instead of computing it again. Alternatively, if you're not ok with storing the planner in global state, you could store an `Arc<dyn RealToComple>` in a static variable, and compute it from the planner with lazy_static.

With either approach, you'd only pay the setup cost the first time you called thefunction. It could be worth benchmarking to see if this gives meaningful speedups!

One other thing to note: RustFFT benefits from SIMD code on x86_64, and on arm, but not for wasm(yet), and it auto-selects them by cpu feature detection. So benchmarks will differ based on what platform you're testing this for - one reason for the huge speedup here might be because you teded on a machine with AVX for example, where rustfft is at its strongest. If ruffle has a wasm benchmark infrastructure, it might be worth benchmarking this inside of that to see if you get as much of a speedup without the SIMD.

ejmahler:
For the lazy static approach, it would be

```rust
lazy_static! {
    static ref STATIC_FFT: Arc<dyn RealToComplex> = {
        let mut planner = realfft::RealFftPlanner::new();
        let fft = planner.plan_fft_forward(2048);
        fft
    };
```

Then from the compute spectrum you can do something like
```rust
let fft = STATIC_FFT.as_ref()
let mut in_left = fft.make_input_vec();
// etc
```

You don't need that first line, but it would eliminate one layer of indirection and might make it easier to visually parse.

torokati44:
> I have one (optional) suggestion for how to improve this: The `RealFftPlanner` caches a lot of setup, and when it gets dropped at the end of the function, that setup gets dropped too, so the next time `compute_spectrum` is called, it has to rebuild that setup.

I was aware that this is its function, and have considered doing something similar with the `cos` LUT, just haven't managed to do it cleanly, and ultimately didn't consider it to be that important.

I don't want to lug it around everywhere in `Activation` though, as this method is not used that commonly.

> One other thing to note: RustFFT benefits from SIMD code on x86_64, and on arm, but not for wasm(yet), and it auto-selects them by cpu feature detection. 

We do enable the `simd128` target feature in one of our web builds, and have observed significant speedups in some small specific places (video decoding mostly) just due to the autovectorization of LLVM, without explicit intrinsics in the code. I'll leave it up to the compiler here to do as it best sees fit... :)

> For the lazy static approach, it would be

I went with `once_cell` instead, as that is what we already have as a dependency, and it's on its way into `std` after all...
But thank you again for the detailed example!

torokati44:
With the final version (with cached pre-planned FFT), the overall speedup of this method is 90%-95%, not counting the planning that only occurs once.

Dinnerbone:
Approximately 92% faster.
image

Dinnerbone (approved):

Herschel (approved):

PR #9497, opened 2023-02-10
desktop: Add --letterbox option

n0samu:
Closes #9492.

PR #10034, opened 2023-03-13
core: Use stage background color for letterbox

Aaron1011:
To help write a test for this, I've introduced
`AutomatedEvent::ReizeViewport`, which can be used to trigger a resize from input.json

Once we have support for capturing and comparing multiple frames in an image test, we can add another `ReizeViewport` to `stage_letterbox.json`, and test letterboxing in the other orientation.


n0samu:
Doesn't this kind of defeat the purpose of letterboxing? What is the rationale for this change?

Aaron1011:
This is the correct Flash Player behavior.

n0samu:
Flash Player actually has no concept of "letterbox" at all, as far as I'm aware. Like I said here: https://github.com/ruffle-rs/ruffle/pull/9406#issuecomment-1436108066
Is there something I'm missing? As I understand, we chose to letterbox content by default in the desktop app as a feature unique to Ruffle, because users tend to prefer it. But thanks to my PR #9497, the behavior can optionally be disabled, in which case Ruffle's behavior matches that of Flash.

Dinnerbone:
I agree with @n0samu here, I think if we have this it should be an option. I think personally I'd prefer black.

Flash Player doesn't have any letterboxing as far as I can tell; the closest thing is `StageScaleMode.SHOW_ALL` which *can look like* letterboxing but isn't, because it shows anything off-stage - not hidden by any bars at all.

Herschel:
Yes, the Flash Player generally not do letterboxing, and the letterbox feature was a QoL feature added for Ruffle (mainly desktop builds to make fullscreen nicer). I personally prefer black in this case as well.

The one place where the Flash Player does do letterboxing IIRC is when `stage.displayState` is fullscreen and `stage.fullScreenSourceRect` is set -- maybe this is what you're seeing @Aaron1011? IIRC in this case it will indeed letterbox using the stage color if the aspect ratio of the `fullScreenSourceRect` doesn't match the display.

Aaron1011:
Closing this, since it seems like I misunderstood how letterboxing works.

PR #10026, opened 2023-03-12
Don't panic when the movie root is removed

Dinnerbone:
So there's two key things to the whole rootless movie thing:
- AVM1 has, as far as I can tell, no way to remove the root movie. I don't know what the behaviour should be for AVM1 activations when there's no root movie as I haven't found a way to test it. I added a panic for when we try to do that, and we'll just have to watch for it and see how it was being used.
- AVM2 not only allows you to remove the root movie, but many games do this deliberately as either a sitelock or "lazy root swap" technique. It's not an error to operate entirely without a movie in AVM2, as things can execute as orphans (thanks Aaron!)

Every movie with this error needs to be retested to see if it's fixed, however from my own testing I can see that some movies error from unrelated things before they have a chance to reattach root, so some things might go from panicking to just being blocked on something else.

Confirmed fixes:
Fixes #9907
Fixes #5208
Fixes #5330 
Fixes #5832
Fixes #5948 (further blocked by TypeError #1009)
Fixes #6180 (no idea what that swf is supposed to do)
Fixes #6778 (same as above)
And then I got tired for now. It fixes a lot.

PR #10044, opened 2023-03-14
avm2: Add support for orphan Loader instances

Aaron1011:
Despite not being MovieClips, Loader instances appear to get the same kind of orphan handling - you can instantiate a Loader and call 'Loader.load' without ever adding it to a parent, and the loader will still run.

I've changed the movie code to work with a new `DisplayObjectWeak` enum. Currently, this just supports `MovieClip` and `Loader`, but it can easily be extended if we ever need other weak display objects.

This also fixes a bug where we were adding the loaded MovieClip as a child of the Loader slightly too early.


SN902:
Good news: previously before this PR two of my swf's only loaded trough command line --proxy=http://localhost, now they works without it

Bad news:  [Rainbow Dash Shooter Project.zip](https://github.com/ruffle-rs/ruffle/files/10963449/Rainbow.Dash.Shooter.Project.zip) previously only worked with avoiding option --proxy=http://localhost, now it start to panic after intro screens:

`ERROR ruffle_core::avm2::activation: AVM2 error: Object(ErrorObject(ErrorObject { class: Error, ptr: 0x25614bffe90 }))
ERROR ruffle_core::avm2::activation: Error: Parsed stage is not attached to document stage
thread 'main' panicked at 'already borrowed: BorrowMutError', core\src\avm2\object\function_object.rs:148:28
AVM2 stack trace:
        at playerio::PlayerIO$/proxy()
        at playerio::PlayerIO$/connect()
        at Function/apply()
        at MethodInfo-55()
        at playerio::PlayerIO$/emptyQueue()
`
        
So yeah, this is symptom of another issue, but no longer avoidable

Bale001:
Fixes that last issue in #8006 to get it past the preloader

Dinnerbone (approved):

Aaron1011:
@SN902 Can you re-test? I think https://github.com/ruffle-rs/ruffle/pull/10026 has improved things, and I'm not seeing a crash when I run it locally with your SWF.

PR #10046, opened 2023-03-14
avm2: Add several more Matrix3D methods

Aaron1011:
These implementations are ported from the MIT-licensed OpenFL code.


core/src/avm2/globals/flash/geom/Matrix3D.as
```diff
@@ -180,6 +180,96 @@ package flash.geom {
...
+			this._rawData[11] = m131 * m214 + m132 * m224 + m133 * m234 + m134 * m244;
+
+			this._rawData[12] = m141 * m211 + m142 * m221 + m143 * m231 + m144 * m241;
+			this._rawData[13] = m141 * m212 + m142 * m222 + m143 * m232 + m144 * m242;
+			this._rawData[14] = m141 * m213 + m142 * m223 + m143 * m233 + m144 * m243;
+			this._rawData[15] = m141 * m214 + m142 * m224 + m143 * m234 + m144 * m244;
+		}
+
+		public function copyFrom(other:Matrix3D):void {
+			// This makes a copy of other.rawData
```
Bale001, line 255: Just curious, why do we copy rawData on getter access? That seems unnecessarily expensive :/

Dinnerbone: That's a common thing in ActionScript API.

Flash Player:
```actionscript
var m: Matrix3D = new Matrix3D();
trace(m.rawData); // 1,0,0,0,0,1,0,0,0,0,1,0,0,0,0,1
m.rawData[0] = 0;
trace(m.rawData); // 1,0,0,0,0,1,0,0,0,0,1,0,0,0,0,1
```

So it's copying on read there.

Dinnerbone (approved):

core/src/avm2/globals/flash/geom/Matrix3D.as
Bale001: Hmm, it's concerning because that means the `prepend` (and `append`) method is way more expensive than it should be. In the OpenFL code, rawData is not copy on read so its cheaper there. We can deal with this in a future PR of course

Dinnerbone: That's true; it might be able to use `_rawData` instead

Herschel:
As a funny FYI, I have a memory of some version of the Flash Player having an outright broken `Matrix3D.multiply` where one of the values would be completely wrong. No clue which version it was, or if it's SWF-version-gated or etc., but it may pop up at some point, so we're not totally surprised :-)

core/src/avm2/globals/flash/geom/Matrix3D.as
Aaron1011: I've updated the methods to use `_rawData`

Herschel:
A quick example, `Matrix3d.determinant` being completely broken in a v12 (Flash Player 10.3) SWF:
```
var a:Matrix3D = new Matrix3D(Vector.<Number>([1,0,0,0, 0,1,0,0, 0,0,1,0, 0,0,0,1]));
trace(a.determinant); // Incorrectly -1 in a v12 SWF
```

[det.zip](https://github.com/ruffle-rs/ruffle/files/10976738/det.zip)
Changing the SWF version from 12 to 13 will give the correct answer. And I'm sure there are tons of other bugs that are version-gated as well with the other methods, so something we'll have to figure out some day.

Herschel (approved):

Issue #2666, opened 2021-01-20
the archive of old newspapers no longer works


INCALB:
Can you help me? the archive of old newspapers no longer works!
http://antico.erasmo.it/?page_id=11&capitoloid=1386

Lord-McSweeney:
Currently requires:
```
flash.utils.IExternalizable
flash.display.BitmapData.generateFilterRect
```

Lord-McSweeney:
Progressed by #10171.

Lord-McSweeney:
Loading images now works, thanks to #11045, but there are still a few glitches.

n0samu:
The main problem now is that the buttons to view the previous/next page don't work:
```
ERROR core/src/avm2/events.rs:419 Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0xa0624c }) to handler FunctionObject(FunctionObject { ptr: 0x13a436c }) : Error: flash.display.Loader.unload - not yet implemented
	at flash.display::Loader/unload()
	at mx.controls::SWFLoader/load()
	at mx.controls::SWFLoader/commitProperties()
	at mx.core::UIComponent/validateProperties()
	at mx.managers::LayoutManager/validateProperties()
	at mx.managers::LayoutManager/doPhasedInstantiation()
	at Function/apply()	at mx.core::UIComponent/callLaterDispatcher2()
	at mx.core::UIComponent/callLaterDispatcher()
```
So we need an implementation of Loader.unload now.

Lord-McSweeney:
Since Loader.unload has been partially implemented, the only issues left are the drop shadow effects and bad text centering.

n0samu:
I'm noticing a few other problems:
- The current page thumbnail on the left sidebar is not shown. It is supposed to let you click and drag to pan the view up and down.
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/714fee54-ccff-4b2a-9d62-fefb0f6c9913)
- Zooming in and out (using the controls on the left sidebar) doesn't work correctly.
- In the page thumbnail view, it is difficult to click the pages to move between them, because Ruffle doesn't treat the thumbnails as part of the buttons.
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/42c90b5b-a588-4d58-9116-bc375c493c01)

Lord-McSweeney:
The last issue seems to have been fixed- I can click on the thumbnails to switch the pages.

[...]

PR #10047, opened 2023-03-14
avm2: Stub DisplayObject.accessibilityProperties

Lord-McSweeney:
Progresses #2666 a little.


Dinnerbone (approved):

Dinnerbone:
On another look, this could literally just be a property, not a setter and getter. Right? 😅

Lord-McSweeney:
I'm keeping this as a getter/setter to match Flash's behavior, as well as making sure strange `override get accessibilityProperties` cases don't break.

PR #10050, opened 2023-03-14
Simplify parameter usage in bitmap, display_object, graphics & others

Dinnerbone:
Just a few classes for now, taking advantage of the new ParametersExt

### DIFF
--- a/core/src/avm2/globals/flash/display/bitmap.rs
+++ b/core/src/avm2/globals/flash/display/bitmap.rs
@@ -6,6 +6,7 @@ use crate::avm2::object::{BitmapDataObject, Object, TObject};
 use crate::avm2::value::Value;
 use crate::avm2::Error;
 
+use crate::avm2::parameters::ParametersExt;
 use crate::bitmap::bitmap_data::BitmapData;
 use crate::character::Character;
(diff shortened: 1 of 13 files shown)


core/src/avm2/globals/flash/display/bitmap.rs
```diff
@@ -22,22 +23,13 @@ pub fn init<'gc>(
...
-            .cloned()
-            .unwrap_or_else(|| "auto".into())
-            .coerce_to_string(activation)?;
-        let smoothing = args
-            .get(2)
-            .cloned()
-            .unwrap_or_else(|| false.into())
-            .coerce_to_boolean();
+            .try_get_string(activation, 1)?
+            .unwrap_or_else(|| "auto".into());
```
Lord-McSweeney, line 31: I think pixel snapping is automatically set to "auto" by the following AS code in Bitmap.as:
`public function Bitmap(bitmapData:BitmapData = null, pixelSnapping:String = "auto", smoothing:Boolean = false) {`

Dinnerbone: Removed it

Herschel (approved):

PR #3383, opened 2021-02-22
Implement BitmapData.paletteMap()

torokati44:
Given this test file: [BitmapPaletteMap2.as](https://gist.github.com/torokati44/db2d3653cff54251899cd18f272bf5d6) ([BitmapPaletteMap2.zip](https://github.com/ruffle-rs/ruffle/files/6021646/BitmapPaletteMap2.zip)) 

The output of Flash Player is:
image

And the difference of that and what Ruffle shows:
image

This time, the range of the difference image is **NOT** remapped, so those red/blue pixels are really that different!
I couldn't figure out the reason for this, it might be the not perfectly matching conversion of colors to/from premultiplied alpha representation.
In this test, the mapping is effectively applied many times over (because of the in-place operation, and the positive offset), so the errors might be amplified. After a single application, the mismatch is hopefully not quite this stark.

This helps a little bit with #2295, taking z0r.de/1906 from this:

image

To this:

image

Now only the "glow" filter effect is missing from this loop.
EDIT: And perhaps a blend mode or two...

PR #10055, opened 2023-03-14
avm2: Implement BitmapData.threshold

Dinnerbone:
Whilst developing this I noticed two fun things:
1 - Our AVM1 test for threshold is wrong, it needs to import `Point`, right now it's passing `undefined` as point for every argument
2 - When running a threshold on an image, with itself as the source image, and a positive destPoint, Ruffle is wrong as it needs to mutate whilst it reads.

Those can be addressed in a future PR. I've left part of the test commented for now, which shows the problem.

Otherwise, the test is ported from AVM1 and the code seems to work!


torokati44:
Oh, 2) is fun indeed, https://github.com/ruffle-rs/ruffle/pull/3383 also had that. Wrestling with aliasing in Rust is tricky...

Dinnerbone:
> Oh, 2) is fun indeed, #3383 also had that. Wrestling with aliasing in Rust is tricky...

Good to know, thanks!

Aaron1011 (approved):

Herschel (approved):
Thank you!

PR #10056, opened 2023-03-14
core: Cache 'width' and 'height' in Bitmap when BitmapData is set

Aaron1011:
If you call 'BitmapData.dispose()', any Bitmap objects using it will continue to report the original 'width' and 'height' values to ActionScript. The values only refresh if you explicitly do 'Bitmap.bitmapData = bitmapDataObject' (including with the same object).

Fancy Pants Adventure World 4 relies on this - it calls BitmapData.dispose(), and then uses the width and height from a previously-constructed Bitmap object.


Herschel (approved):
Thank you!

PR #10057, opened 2023-03-14
web: Don't let Config extend BaseLoadOptions

danielhjacobs:
Alternative to #10054.


danielhjacobs:
https://github.com/ruffle-rs/ruffle/pull/10054 is preferred.

PR #10018, opened 2023-03-12
avm2: Support any name in XML get_property_local

evilpie:
Fixes #9999

Issue #8330, opened 2022-10-19
New Club Penguin not supported ruffle player


deividbinev12:
### Describe the bug

https://user-images.githubusercontent.com/80902993/196750362-eb8831bc-c234-41c8-aa41-40d783b669d0.mp4

it's not working, because this is multi files, this is actionsupport 3.0
i know ruffle player multiplayer is not working

### Expected behavior

![Screenshot 2022-10-19 192944](https://user-images.githubusercontent.com/80902993/196750531-8067cbf3-45b9-403d-88e4-334f636bac67.jpg)
you know ''Download'' and ''Launch NewCP App'' it's can work with flash player, his server it's working

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge

### Additional information

from https://newcp.net/en/

Lord-McSweeney:
Still not working. Direct SWF: https://media1.newcp.net/play/v2/client/club_penguin.swf
Right now, it's erroring with `Error dispatching event EventObject(EventObject { type: "addedToStage", class: flash.events::Event, ptr: 0xb78470 }) to handler FunctionObject(FunctionObject { ptr: 0xb77fb0 }) : RustError("Attempted to call \"\" with 1 arguments (more than 0 is prohibited)")`

Lord-McSweeney:
With #9647, this errors on `ERROR ruffle_core::avm2::activation: ReferenceError: Error #1069: Property * not found on XMLList and there is no default value.`.

Lord-McSweeney:
Now errors: `Error: MainContext could not determine the locale for the specified langauge. Make sure that 'null' has been added to the switch.`.

AllinolCP:
> Now errors: `Error: MainContext could not determine the locale for the specified langauge. Make sure that 'null' has been added to the switch.`.

This is related to flashvars not being provided (e.g. lang=en)

evilpie:
I investigated this and made some progress, but got stuck at some point with something related to `contextView`. This game really needs a lot of stuff. What I've found:
- [ ] `-Plang=en` as @AllinolCP said
- [ ] `decodeURI`
- [ ] `ExternalInterface.call("window.location.href.toString")`
- [ ] XML.parent (#10244)
- [ ] `URLRequest.requestHeaders`

Lord-McSweeney:
This URL works for the page: https://play.newcp.net/eng/index.html#/login

Lord-McSweeney:
[...]

PR #10061, opened 2023-03-14
avm2: Support wildcard/any name in XMLList get_property_local

evilpie:
This is similar to #10018. Improves #8330.


Aaron1011:
Thanks!

PR #10063, opened 2023-03-14
chore: Port to bitflags 2.0.0

torokati44:
Since Renovate won't be able to deal with this on its own, I did it for it.
I don't actually know if there is a less painful way though (through a macro option or something).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1744,9 +1744,9 @@ dependencies = [
 [[package]]
 name = "h263-rs"
 version = "0.1.0"
-source = "git+https://github.com/ruffle-rs/h263-rs?rev=8c00d0e13892ee349f76eb4f2651cb2b921d3766#8c00d0e13892ee349f76eb4f2651cb2b921d3766"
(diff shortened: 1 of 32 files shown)


torokati44:
>  I don't actually know if there is a less painful way though (through a macro option or something).

Looks like This Is The Way: https://kodraus.github.io/rust/2022/10/07/bitflags2.html#upgrading-to-2x

torokati44:
Thank gosh for our fairly extensive test coverage! :relieved:

torokati44:
See also: https://github.com/ruffle-rs/h263-rs/pull/47

Herschel (approved):
Thank you!

Dinnerbone:
Maybe we should include h263-rs update in this PR so there isn't a period where we pull in two bitflags versions?

torokati44:
We can do that, sure. Even though many of our other dependencies might still pull 1.x in.

Also, other than the last one, these commits don't build, so should I squash them for the sake of bisecting later?

Dinnerbone:
Yeah that sounds smart. We can squash it when we merge the PR

torokati44:
Added the h263-rs bump, feel free to squash on merge.

PR #10066, opened 2023-03-15
avm2: Skip running enter_frame for children as well

Aaron1011:
When we skip running a frame for a MovieCilp, we skip all of its children as well. However, this skip 'counts' as a skip for any children that already wanted to skip their next frame. For example, say  we create three objects in ActionScript, and arrange them like 'obj1 -> obj2 -> obj3'.
The first 'obj1.enter_frame' call will not run a new frame for any of the objects, but next time, 'obj1.enter_frame' will run a new frame for all of the objects.

This fixes jacksmith, which was missing a frame1-framescript due to 'enter_frame' getting incorrectly run for a deeply nested child.


Herschel (approved):
Thank you!

PR #10078, opened 2023-03-15
chore: Update avm2 compatibility warning to be less severe

Dinnerbone:
I think we're very close to the point of removing it entirely, but until then let's be a little more confident in our support.


torokati44:
I agree with the changes to the first sentence, but highlighting interactivity in the second one no longer really applies.
I'd say something like: `If you choose to run it anyway, some parts of it may be missing or not work properly.`

Herschel (approved):
Thank you!

PR #10079, opened 2023-03-15
avm2: Add TextFormatDisplay class and stub TextFormat.display property

n0samu:
Along with #10044, this gets [Atomic Puzzle 2](http://www.notdoppler.com/atomicpuzzle2.php) into the first level.


Dinnerbone (approved):

Herschel (approved):
Thank you!

PR #10088, opened 2023-03-15
Partly implement XML.toString

evilpie:


tests/tests/swfs/avm2/xml_tostring/Test.as
```diff
@@ -0,0 +1,19 @@
...
+  public class Test extends Sprite { }
+}
+
+// FIXME: Implement indentation.
+XML.prettyPrinting = false;
+
+var xml = <animal id="1">Cow</animal>
+trace("xml.toString(): " + xml.toString());
+
+xml = <animals>
```
Aaron1011, line 12: Can you add a test that uses attribute values with characters that get escaped?

tests/tests/swfs/avm2/xml_tostring/Test.as
```diff
@@ -0,0 +1,22 @@
...
+  <animal id="1">Cow</animal>
+  <animal id="2">Pig</animal>
+</animals>;
+trace("xml.toString(): " + xml.toString());
+
+xml = <foo><bar a="x" b="y" c="z"/></foo>
+trace("xml.toString(): " + xml.toString());
+
+xml = <foo><bar x="a&quot;b">&gt;&amp;&lt;</bar></foo>;
```
Aaron1011, line 21: Sorry if it seems like I'm nitpicking, but could you also add a usage of `new XML` with an un-escaped characters that should be escaped? Eg. `new XML("<foo><bar attr="Quote: \" < & "</bar></foo>")` for all of the characters that your PR escapes.

evilpie: Ok

evilpie: That doesn't actually work, because that isn't valid XML, so the isn't parse-able. I did manually verify that we seem to have the decoded string in memory.

Aaron1011: Sorry, should have tested it first. I'm fine with merging this as-is - I'll do some local experimentation, and add more tests later.

Aaron1011 (approved):

PR #10090, opened 2023-03-15
avm2: Implement adding attribute to XML object

Aaron1011:
We only support values that are neither XML nor XMLList, since we can't yet properly stringify those.

Attempting to modify an existing attribute throws an error.


core/src/avm2/object/xml_object.rs
```diff
@@ -224,8 +224,40 @@ impl<'gc> TObject<'gc> for XmlObject<'gc> {
...
+                            "Cannot set an XML/XMLList object {:?} as an attribute",
+                            obj
+                        )
+                        .into());
+                    }
+                }
+                let value = value.coerce_to_string(activation)?;
+                let new_attr = E4XNode::attribute(
+                    activation.context.gc_context,
+                    name.local_name().unwrap(),
```
Bale001, line 253: This unwrap causes a panic when I do:
```as
var outer = <outer></outer>;
outer.@public::* = "test";
```

Bale001 (approved):

Herschel (approved):

Herschel (changes requested):
There's Untitled-3.fla in the commit, maybe that could have a better name (Test.fla?) or remove it if it's unnecessary?

Herschel:
Whoops, nevermind, the commit was removing the file. :-)

Herschel (approved):

PR #10092, opened 2023-03-15
avm2: Implement XML.appendChild for XML objects

Aaron1011:
This adds a reference to the passed-in XML object to the children list. The object will compare === with 'parent.children()[newChildIndex]`


Herschel (approved):

core/src/avm2/globals/xml.rs
```diff
@@ -197,3 +197,29 @@ pub fn attribute<'gc>(
...
+    this: Option<Object<'gc>>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let this = this.unwrap();
+    let xml = this.as_xml_object().unwrap();
+
+    let child = args
+        .get(0)
+        .unwrap_or(&Value::Undefined)
+        .coerce_to_object(activation)?;
```
Dinnerbone, line 212: `get_object` to handle nulls

Dinnerbone (approved):

Issue #10067, opened 2023-03-15
audio: Regression in closely spaced sound effects


n0samu:
#9671 seems to have caused a regression with sound effects that play in quick succession. Now the sound effects are playing at the same time as each other instead of one at a time and in order. It seems to be particularly happening at the start of movies before a stable audio buffer size has been established, but I can't 100% confirm that. The intro in this version of Electricman 2 provides a good demonstration of this problem: [electricman2_dx.zip](https://github.com/ruffle-rs/ruffle/files/10975223/electricman2_dx.zip) (First mentioned by @Zinc-Exclamation)

Using [Ruffle-Regressions](https://ruffle-regressions.github.io/), we can observe that the DX Interactive intro sounded fine in nightly 2023-03-11, but in nightly 2023-03-12 (the first build with the dynamic audio buffer), the intro sound effects don't play correctly. We can hear fewer of them because some are playing at the same time as each other rather than in sequence.

https://user-images.githubusercontent.com/71368227/225185634-58010a15-08dd-4d67-a05d-dd85d66bdf82.mp4

torokati44:
Yes, this is most likely due to the buffer length being chucked way up early in the content, when lots of initialization is taking place. And any change in which set of sounds to play can also only happen at the granularity of the buffer fill interval, which is around 370 ms (as configured currently) in the worst case.

This aligns well with the audio at the end of your screen capture:
![image](https://user-images.githubusercontent.com/288816/225364290-5152154f-db8b-402a-bc00-41fccc779f32.png)

Some possible solutions for this are:
  - Remember when any sound was requested to start, and play them all delayed by the current buffer duration, not right at the start of the next buffer window.
    - This might not be the best, especially in fast paced stuff like the above intro.
  - Cut off the beginning of the sounds that just started playing, so at least the next buffer contains "exactly what it should", with no delay.
    - I don't think this is acceptable in most cases.
  - Tune the buffer size algorithm somewhat, like by adding a warmup period where the buffer size is not increased, even if it causes some stutter.
    - I think this what I'm going to try to do.

n0samu:
Fixed by #10095 🎉

PR #10095, opened 2023-03-15
web/audio: Add a warmup period before increasing buffer size

torokati44:
This is meant to address #10067.


Herschel (approved):
Thank you!

Another thing that might be helpful in general would be not starting the audio context/buffers until the first instance where a sound actually plays, both as an optimization for silent SWFs and to avoid chugs during an initial SWF load.

torokati44:
> Another thing that might be helpful in general would be not starting the audio context/buffers until the first instance where a sound actually plays, both as an optimization for silent SWFs and to avoid chugs during an initial SWF load.

Yeah, that sounds like a good idea. But I'm not going to get to it now.

Dinnerbone (approved):

PR #10102, opened 2023-03-15
render: Support more Context3D texture formats

Aaron1011:
None of these formats can currently be implemented correctly with wgpu, so we just use Rgba8Unorm instead.

The handling of opaque compressed textures is a little sketchy - it should work for 'normal' SWFs that upload an opaque BitmapData, but we might need to manually adjust the alpha values if

### DIFF
--- a/core/src/avm2/object/context3d_object.rs
+++ b/core/src/avm2/object/context3d_object.rs
@@ -5,6 +5,7 @@ use crate::avm2::object::script_object::ScriptObjectData;
 use crate::avm2::object::{Object, ObjectPtr, TObject};
 use crate::avm2::value::Value;
 use crate::avm2::Error;
+use crate::avm2_stub_method;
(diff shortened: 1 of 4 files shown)


core/src/avm2/object/context3d_object.rs
```diff
@@ -470,3 +473,27 @@ impl std::fmt::Debug for Context3DObject<'_> {
...
+                "flash.display3D.Context3D",
+                "createTexutre",
+                "with BgrPacked"
+            );
+        }
+        Context3DTextureFormat::Compressed => {
+            avm2_stub_method!(
+                activation,
+                "flash.display3D.Context3D",
+                "createTexutre",
```
Lord-McSweeney, line 493: Typo ("createTexutre" --> "createTexture") here and on line 485.

Dinnerbone (approved):

PR #5340, opened 2021-09-20
Wire in DefineVideoStream smoothing flag to the renderer

torokati44:
This is how it's "supposed to" work, according to the docs (and common sense as well), but more experimentation is needed to determine how and why exactly Flash Player actually enables smoothing on videos.
In Flash Player, on my hand-crafted tiny test videos, this flag worked as expected; but when enabling this flag on videos in SWFs found in the wild, it did nothing, the picture stayed pixelated.

PR #10103, opened 2023-03-15
swf: Don't ignore Bitmap smoothing flags in SWF <8

torokati44:
I think this comment: `Bitmap smoothing only occurs in SWF version 8+.`
Refers to this section in the AS reference:

However, the SWF spec does not mention the version-gate for this flag:

As for `Lower versions use 0x40 and 0x41 type even when unsmoothed.` it would be great if someone could come up with a real-world example of this.
I already have a counter-example: https://z0r.de/L/z0r-de_56.swf
This is a V6 SWF, and the bitmaps should appear smoothed.

And at any rate, handling this at the SWF read/write level might not be the best idea - I think it would be in a better place near rendering.

### DIFF
--- a/swf/src/read.rs
+++ b/swf/src/read.rs
@@ -1635,8 +1635,7 @@ impl<'a> Reader<'a> {
             0x40..=0x43 => FillStyle::Bitmap {
                 id: self.read_u16()?,
                 matrix: self.read_matrix()?,
-                // Bitmap smoothing only occurs in SWF version 8+.
-                is_smoothed: self.version >= 8 && (fill_style_type & 0b10) == 0,
+                is_smoothed: (fill_style_type & 0b10) == 0,
                 is_repeating: (fill_style_type & 0b01) == 0,
             },
 
--- a/swf/src/read.rs
+++ b/swf/src/read.rs
@@ -1629,8 +1629,7 @@ impl<'a> Reader<'a> {
             0x40..=0x43 => FillStyle::Bitmap {
                 id: self.read_u16()?,
                 matrix: self.read_matrix()?,
-                // Bitmap smoothing only occurs in SWF version 8+.
-                is_smoothed: self.version >= 8 && (fill_style_type & 0b10) == 0,
+                is_smoothed: (fill_style_type & 0b10) == 0,
                 is_repeating: (fill_style_type & 0b01) == 0,
             },
 
--- a/swf/src/write.rs
+++ b/swf/src/write.rs
@@ -1534,9 +1534,7 @@ impl<W: Write> Writer<W> {
                 is_smoothed,
                 is_repeating,
             } => {
-                // Bitmap smoothing only an option in SWF version 8+.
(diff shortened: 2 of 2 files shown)


Herschel:
There is some complex behavior here to be careful of -- IIRC it has to do with a bitmap being in motion or not. In lower SWF version a one-frame movie would result in the smoothing behavior, while a multiple frame movie would be unsmoothed unless quality would be set to BEST.

I'll try to remember the exact behavior and get back to you!

torokati44:
Oh, this sounds similar to https://github.com/ruffle-rs/ruffle/pull/5340!

Herschel:
I see this PR is only changing `swf`, and indeed those comments are incorrect! So this is good to add. We do need to make sure we emulate the above behavior at some point, though.

Herschel (approved):

Herschel:
Actually thinking that this might be a little dangerous to merge right away as this will incorrectly cause a lot of classic SWFs to be smoothed.

torokati44:
Okay, I wasn't sure of it either, without any corresponding changes - hence my wish for relevant real-world content.
I meant this mostly to gather our collective thoughts about how to handle this.

Herschel:
No worries, I'll try to post some minimal example SWFs.

Herschel:
You can try Wasted Sky from a demo page -- renders as unsmoothed in FP but smoothed with this PR.

Here's some minimal examples:
[bitmap-smooth-swf7.zip](https://github.com/ruffle-rs/ruffle/files/10985336/bitmap-smooth-swf7.zip)
The singleframe SWF renders smoothed with default `_quality = "HIGH";`, but the multiframe one doesn't. I think the criteria may be as simple as checking `swf.num_frames > 1`.

To implement, perhaps we add a `RenderBackend::set_allow_bitmap_smoothing` accessor that will be set appropriately by `core::display_object::Stage`.

Herschel:
Some more examples:
[bitmap-smooth-flag.zip](https://github.com/ruffle-rs/ruffle/files/10985555/bitmap-smooth-flag.zip)

2 frame SWF with 2 bitmaps, one flagged smooth and one unsmoothed,
* bitmap-smooth-swf7-flag.swf: SWFv7, "HIGH" quality. Neither bitmap smoothed.
* bitmap-smooth-swf7-flag-best.swf: SWFv7, "BEST" quality. One bitmap smoothed.
* bitmap-smooth-swf8-flag.swf: SWFv8, "HIGH" quality. One bitmap smoothed.

The linked z0r SWF is 1 frame, so it honors the smoothing setting. The docs above indicate that the smoothing should take place "if the movie clip is static", which sounds like it should smooth if the timeline is stopped or something -- but so far from my testing only shows it being affected by the # of frames on the timeline, regardless if it is stopped or not.

Another point of confusion is the Flash IDE will always export the bitmap smoothing flag when SWF version is set to 7 and below, regardless of what is set in the UI, so I had to manually edit these SWFs.

PR #10105, opened 2023-03-15
avm2: Update SoundChannelObject::position on read (fix #9952)

### LINKED ISSUE #9952: lazer man - music reversing and looping issue
### Describe the bug

in this game, lazerman, I think all of the music is one big track, and the music is rewinded and fast forwarded depending on what you're doing (being on the menus, normal gameplay, rampage, etc) so when you complete a level, it will jump to the portion of the track that has the menu theme, and so on.
it seems that playing enough levels will make it reverse and fast forward incorrectly, causing the music to loop too early, and eventually making it sound like I've crashed my Wii console again, haha

(headphone warning!)

you can hear how the music tries to jump to the rampage portion  at the middle of the clip or so, but its looping that too early, and then jumps back to the main theme, which that loop had slowly been getting shorter and shorter, until it got stuck on one note, hence the lock-up like sound (at least, thats how I think its all happening)

https://user-images.githubusercontent.com/103858019/224254151-91c9a195-acb2-4f1f-afa2-fb1593a07a00.mp4

### Expected behavior

the music should not fail to work

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

[lazermanMTgzNw==.zip](https://github.com/ruffle-rs/ruffle/files/10939734/lazermanMTgzNw.zip)

Herschel:
`SoundChannel.position` was being updated by the audio manager each
tick, but the Flash Player only updates and caches the position
when `SoundChannel.position` is accessed. In contrast, an AVM1
`Sound` will constantly update its position.

This means accessing `position` only once after a sound has
finished playing will return 0.

Fixes #9952.

Also adds a test case to `soundchannel_position` test, although the test is currently ignored.


Herschel:
Here's some code to show that the position is actually grabbed on demand:

```
var s:Sound = new Snd();
var chan:SoundChannel = s.play();

trace("START");
for(var i = 0; i < 20000000; i++) {
	if( (i%100000) == 0 ) {
		trace(chan.position);
	}
}
trace("END");
```

Despite being a blocking loop, you'll see the position update in Flash Player, although Ruffle updates this way more often than FP.

Dinnerbone (approved):

adrian17 (approved):

PR #10109, opened 2023-03-16
render: Correctly handle BYTES_4 vertex data

Aaron1011:
Each byte gets normalized into a float in the range [0, 1]


Aaron1011:
This depends on https://github.com/ruffle-rs/ruffle/pull/10045

Herschel (approved):

PR #10118, opened 2023-03-16
avm2: Implement BitmapData.hitTest

Dinnerbone:


MartySVK:
You were right, only the first topping that touches it seems to stick. In Flash Player, all of them stick.
https://github.com/ruffle-rs/ruffle/issues/9201#issuecomment-1471674372

Dinnerbone:
I've investigated and can't find any reason related to hitTest why it wouldn't work :(

Herschel (approved):

adrian17 (approved):

Issue #9955, opened 2023-03-10
Let it Glow 2: Boards render incorrectly and only edges are clickable


n0samu:
### Describe the bug

Link to SWF: http://i.notdoppler.com/files/letitglow2.swf

In [Let it Glow 2](http://www.notdoppler.com/letitglow2.php), boards and floors and walls (all long and thin objects) render without their textures. So level 3 looks like this in Ruffle:
![image](https://user-images.githubusercontent.com/71368227/224299887-51cbd4f3-7c21-42ec-86cc-a53266870ef6.png)

Also, wooden boards are only clickable on their edges, making levels with wooden boards much more difficult than they should be.

### Expected behavior

Here is what level 3 is supposed to look like:
![image](https://user-images.githubusercontent.com/71368227/224300099-25856089-19b3-4b61-a0eb-537fbe3b7d17.png)

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

The rendering issues seem to be caused by `flash.display.GraphicsSolidFill`, `flash.display.Graphics.beginGradientFill`, and  `flash.display.Graphics.beginBitmapFill` being unimplemented.

n0samu:
Fixed by #10123

Issue #9829, opened 2023-03-03
Chainsaw Slasher: Visual issues and stuck on title screen


waspennator:
### Describe the bug

https://www.newgrounds.com/portal/view/627115?emulate=flash
Game has 1 visual issue on the preloader and one major issue keeping it from being playable on ruffle. On the preloader the progress bar seems to render behind the background of the progress bar, and on the title screen theres not only a bunch of missing text, but I cant seem to click on anything.
[www.newgrounds.com-1677833070577.log](https://github.com/ruffle-rs/ruffle/files/10880203/www.newgrounds.com-1677833070577.log)

On Preloader:
![CS1](https://user-images.githubusercontent.com/78768499/222674140-7b65ff33-d33e-404b-ba6e-0b0ef7be8c47.png)

On Title Screen:
![CS2](https://user-images.githubusercontent.com/78768499/222674203-c4a6f4e3-3c03-403a-a156-ca1f7ca44e84.png)

### Expected behavior

![CS3](https://user-images.githubusercontent.com/78768499/222674709-d90c2f01-1276-49d1-8ae7-682b8dfe60ae.png)

![CS4](https://user-images.githubusercontent.com/78768499/222674740-69bf9c95-3ded-4894-b5c1-d9fb0fa6ea99.png)

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 110

Lord-McSweeney:
#9857 fixes the display of the loading bar. The buttons look like they need `Graphics.beginBitmapFill()`.

waspennator:
[ruffle.rs-1679013858792.log](https://github.com/ruffle-rs/ruffle/files/10996837/ruffle.rs-1679013858792.log)

The preloader intro seems to have regressed from 2023-03-15 since it no longer plays anymore and the main menu buttons still don't show up, but the spots where they usually are do seem to be clickable now, meaning you can get ingame. However the game breaks instantly as soon as you get ingame.

![Chainsaw Ruffle](https://user-images.githubusercontent.com/78768499/225783487-99a12709-0a9a-42c3-a7fa-280258a99f5f.png)

Lord-McSweeney:
This also needs `BitmapData.setVector`.

waspennator:
Bitmapdata.setvector was merged in by #10275, game seems to run now into gameplay, but everyone seems to be invisible, doesn't seem to stop the player character from being attackable though.

[...]

Issue #9549, opened 2023-02-13
But RoboPolice are...: Invisible buildings


waspennator:
### Describe the bug

https://www.newgrounds.com/portal/view/598843?emulate=flash

The buildings you need to jump around on are entirely invisible on ruffle, making the game impossible to play.

Running this on the ruffle web demo, I see this pop up in the log as soon as I press play, so maybe it uses this to show the buildings?

Encountered stub: AVM2 flash.display.Graphics.beginBitmapFill()

![Robopolice Ruffle](https://user-images.githubusercontent.com/78768499/218359020-1a7987c6-0a8b-4a31-b252-905fba0bf414.png)

### Expected behavior

![Robopolice Flash](https://user-images.githubusercontent.com/78768499/218359031-a30fd4b6-c29c-445a-b731-e1fec3209844.png)

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 110

n0samu:
This happens because `beginBitmapFill()` is unimplemented:
`WARN ruffle_core::stub: Encountered stub: AVM2 flash.display.Graphics.beginBitmapFill()`

waspennator:
The buildings seem to show up now making this game actually playable now, but for some reason they don't have glowing windows.

![Run Ruffle](https://user-images.githubusercontent.com/78768499/225782650-16ba5e98-327e-4276-87ba-2e8809e21ee5.png)

Lord-McSweeney:
Fixed by #10630.

Issue #8543, opened 2022-11-14
House of Dead Ninjas: Doesn't Start


waspennator:
### Describe the bug

Game seems to get stuck preloading and never seems to move past it, even though it looks like it finished loading. I'm also seeing this in the console log.
![House of Dead Ninjas console log](https://user-images.githubusercontent.com/78768499/201563049-fd46710c-4f2c-4214-9f4d-c0baa042ffc1.png)

https://www.newgrounds.com/portal/view/566333?emulate=flash

### Expected behavior

It should load and start without locking up.

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 107

waspennator:
Starting to wonder if this has something to do with "NetConnection"

Lord-McSweeney:
`flash.net.NetConnection` is not yet implemented.

Lord-McSweeney:
If `NetConnection` is stubbed the error becomes: `RustError("ArgumentError: Cannot remove object from display list it is not a child of.")`

waspennator:
Tested the game out again on the latest nightly and things seem to be a little different, but still not booting to the title screen. Normally there is supposed to be a fancy little retro loading screen with retro boxart of the game in the background and a fancy dragon loading bar, then the game is supposed to flash white and transition to the intro and eventually the title screen.

But on the latest ruffle nightly, it shows a black screen for a little bit, then the white flash happens and then........... nothing.

[ruffle.rs-1678499214948.log](https://github.com/ruffle-rs/ruffle/files/10947307/ruffle.rs-1678499214948.log)

waspennator:
[House of Dead Ninjas.zip](https://github.com/ruffle-rs/ruffle/files/10947626/House.of.Dead.Ninjas.zip)
Providing a version of the game from Flashpoint since it seems to progress further compared to the one currently on NG.

Lord-McSweeney:
Looking into it, it seems that the XML has a UTF-16 BOM, which Ruffle will need to strip out.

Lord-McSweeney:
#10280 progresses this to `RustError("ArgumentError: Cannot remove object from display list it is not a child of.")`.

waspennator:
Actually I retested the NG version of the game again and I realized, I think most of Adult Swims uploaded games on NG including this one just redirect to a random page on Adult Swim now. Tested it on Flash player and it seems to redirect to Adult Swim games now.

[...]

PR #10123, opened 2023-03-16
avm2: Graphics stubs + implementations

Dinnerbone:
Implements:
- beginBitmapFill
- beginGradientFill
- lineBitmapStyle
- lineGradientStyle

Stubs:
- cubicCurveTo
- copyFrom
- drawPath
- drawRoundRectComplex
- drawTriangles
- drawGraphicsData
- readGraphicsData

Also adds tests for existing Graphics methods as I couldn't see any


waspennator:
For my reports, should effect
#9829 
#9549 
#8543 (Though I believe it's stuck on another issue)

n0samu:
As a bonus, could you possibly look at flash.display.GraphicsSolidFill? 😄
(#10126, #9955)

core/src/avm2/globals/flash/display/graphics.rs
```diff
@@ -41,13 +43,93 @@ pub fn begin_fill<'gc>(
...
+                is_smoothed,
+                is_repeating,
+            }));
+        }
+    }
+
     Ok(Value::Undefined)
 }
 
+fn swf_matrix_from_object<'gc>(
```
adrian17, line 99: could we reuse `fn object_to_matrix` here, at least to grab the AS fields?
(kinda weird that we have two different matrix types here, but I'm not questioning which one is right-er in which scenario)

adrian17 (approved):

Dinnerbone:
> As a bonus, could you possibly look at flash.display.GraphicsSolidFill? 😄 (#10126, #9955)

It's more involved because whilst the class is trivial to stub, it's part of `Graphics.drawGraphicsData` which is pretty complex. I'll take a look but no guarantees! Definitely not for this PR.

core/src/avm2/globals/flash/display/graphics.rs
Dinnerbone: Looks like we already added `From<ruffle_render::matrix::Matrix> for swf::Matrix`, so I've made that change!

swf Matrix is lossy and render one isn't... otherwise I don't know. It was a change made a very long time ago to split them up.

PR #10127, opened 2023-03-16
extension: Give precedence to toggles over default options (fixes #5825)

### LINKED ISSUE #5696: Dragon Ball Devolution no download option in context menu

### LINKED ISSUE #5825: Prefer extension options over default configuration for any options defined in both locations

danielhjacobs:
Also fixes #5696

Edit: This only works when the extension is the version of Ruffle that winds up taking precedence. If the self-hosted build takes precedence, the extension is not loaded until after RufflePlayer.load, and therefore the extension configuration is not available until then , so I can't take the toggles into account before loading content. The best way to ensure the extension loads early enough to take precedence in order to properly test this PR is to simulate a slow network connection using network throttling, as described in https://developer.chrome.com/docs/devtools/network/reference/#throttling and https://firefox-source-docs.mozilla.org/devtools-user/network_monitor/throttling/index.html.

### DIFF
--- a/web/packages/core/src/public-api.ts
+++ b/web/packages/core/src/public-api.ts
@@ -12,7 +12,10 @@ declare global {
          * [[PublicAPI]] instance itself.
          */
         RufflePlayer?:
-            | { config?: DataLoadOptions | URLLoadOptions | object }
+            | {
+                  config?: DataLoadOptions | URLLoadOptions | object;
+                  conflict?: Record<string, unknown> | null;
+              }
             | PublicAPI;
     }
 }
@@ -33,11 +36,11 @@ export class PublicAPI {
      * The configuration object used when Ruffle is instantiated.
      */
     config: DataLoadOptions | URLLoadOptions | object;
+    conflict: Record<string, unknown> | null;
 
     private sources: Record<string, typeof SourceAPI>;
     private invoked: boolean;
     private newestName: string | null;
-    private conflict: Record<string, unknown> | null;
 
     /**
      * Construct the Ruffle public API.
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -666,6 +666,13 @@ export class RufflePlayer extends HTMLElement {
         try {
             this.loadedConfig = {
                 ...DEFAULT_CONFIG,
+                ...(window.RufflePlayer?.conflict &&
+                window.RufflePlayer?.conflict["newestName"] === "extension"
+                    ? (window.RufflePlayer?.conflict["config"] as Record<
+                          string,
+                          unknown
+                      >)
+                    : {}),
                 ...(window.RufflePlayer?.config ?? {}),
                 ...this.config,
                 ...options,
--- a/web/packages/extension/src/ruffle.ts
+++ b/web/packages/extension/src/ruffle.ts
@@ -16,8 +16,8 @@ function handleMessage(message: Message) {
         case "load": {
             const api = window.RufflePlayer ?? {};
             api.config = {
-                ...api.config,
                 ...message.config,
+                ...api.config,
             };
             window.RufflePlayer = PublicAPI.negotiate(api, "extension");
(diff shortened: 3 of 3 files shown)


Herschel:
So that I understand, this allows explicitly set extension settings to take precedence over self-hosted settings, except in those cases where the self-hosted version registers itself first are still troublesome?

danielhjacobs:
> this allows explicitly set extension settings to take precedence over self-hosted settings

Not quite. As requested by Elementtair in https://github.com/ruffle-rs/ruffle/issues/5825#issuecomment-1432333275 and Txori in https://github.com/ruffle-rs/ruffle/issues/5696#issuecomment-1471713810, this will not override configuration options set by website owners. However, the extension toggles have never been taken into account on sites that self-host, even when the site owners don't set conflicting configuration options, because the default options end up in the config, while the extension toggles end up in the conflict's config.

Once I started taking the extension toggles into account, I realized the current order of the spread syntax for the extension config gave precedence to toggles over self-hosted configuration options (but that didn't matter because the extension config was not taken into account). In order to accommodate Txori's and Elementtair's requests that site owner configs should take precedence, I had to reorder the spread syntax (see changes to web/packages/extension/src/ruffle.ts)

> except in those cases where the self-hosted version registers itself first are still troublesome

Yes, because when that happens the `RufflePlayer.load` function is called before the extension's content script has even loaded, so I don't even have the toggles' values yet before the content loads. This ties into #7505. Most configuration options are useless if they aren't taken into account before the content loads. For the ones that aren't, such as `showSwfDownload`, there are ways to take them into account properly even if the self-hosted build takes precedence. I can add this to the `showContextMenu` function and check `combinedConfig.showSwfDownload` instead of `this.loadedConfig.showSwfDownload`:
```ts
const combinedConfig = {
    ...this.loadedConfig,
    ...(window.RufflePlayer?.conflict &&
    window.RufflePlayer?.conflict["newestName"] === "local"
        ? (window.RufflePlayer?.conflict["config"] as Record<
              string,
              unknown
          >)
        : {}),
    };
```
However, when I suggested that to nosamu on Discord he said "it would be nice for users but I think we should maybe just wait till we can get the underlying issues solved"

Herschel (approved):
LGTM, although I'm fairly dumb when it comes to the self-hosted+extension interactions, so someone more qualified may want to review!

Lord-McSweeney:
LGTM, though I'm not sure about making `conflict` public.

danielhjacobs:
It's the only way to check it outside the public-api file in ruffle-player.

Herschel:
Thank you! Going to go ahead and merge this.

PR #10129, opened 2023-03-16
avm2: Stub XML.namespace and Namespace.prefix/uri

evilpie:
Depends on #10116


core/src/avm2/globals/namespace.rs
```diff
@@ -52,6 +77,34 @@ pub fn class_init<'gc>(
...
 
+/// Implements `Namespace.prefix`'s getter
+pub fn prefix<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Option<Object<'gc>>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    if let Some(_) = this.and_then(|t| t.as_namespace_object()) {
+        avm2_stub_getter!(activation, "Namespace", "prefix");
+        return Ok(AvmString::new_utf8(activation.context.gc_context, "").into());
```
Bale001, line 88: You can do `"".into()` instead
```suggestion
        return Ok("".into());
```

core/src/avm2/globals/namespace.rs
```diff
@@ -52,6 +77,34 @@ pub fn class_init<'gc>(
...
+/// Implements `Namespace.uri`'s getter
+pub fn uri<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Option<Object<'gc>>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    if let Some(_) = this.and_then(|t| t.as_namespace_object()) {
+        avm2_stub_getter!(activation, "Namespace", "uri");
+        return Ok(AvmString::new_utf8(activation.context.gc_context, "").into());
+    }
```
Bale001, line 103: Does this really need to be stubbed? I'm thinking it can be as simple as:
```rs
if let Some(this) = this {
    if let Some(namespace) = this.as_namespace() {
        return Ok(namespace.as_uri().into());
    }
}
```
(I haven't tested this code, might have to modify it a bit)

Bale001 (approved):

Dinnerbone (approved):

Aaron1011 (approved):

torokati44 (commented):
There's a typo in one of the commit message tags (`avm22`), but oh well, it's not the end of the world.

PR #10136, opened 2023-03-16
avm2: Convert SoundTransform and SoundMixer to ActionScript classes

Aaron1011:


core/src/avm2/globals/flash/media/sound_transform.rs
```diff
@@ -0,0 +1,56 @@
...
+
+        if left_to_right != 0.0 || right_to_left != 0.0 {
+            return Ok(0.0.into());
+        }
+
+        let left_to_left = this
+            .get_public_property("leftToLeft", activation)?
+            .coerce_to_number(activation)?;
+
+        return Ok((1.0 - left_to_left.powf(2.0).abs()).into());
```
Herschel, line 30: I believe `abs` is unnecessary here because the value is squared

core/src/avm2/globals/flash/media/sound_transform.rs
```diff
@@ -0,0 +1,56 @@
...
+    activation: &mut Activation<'_, 'gc>,
+    this: Option<Object<'gc>>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    if let Some(mut this) = this {
+        let pan = args
+            .get(0)
+            .cloned()
+            .unwrap_or(Value::Undefined)
+            .coerce_to_number(activation)?;
```
Herschel, line 47: Can this use `ParametersExt` now?

Lord-McSweeney:
I think the entire `SoundTransform` class can be easily converted to pure AS:
```
public function get pan():Number {
    var LTR:Number = this.leftToRight;
    var RTL:Number = this.rightToLeft;
    
    if (LTR != 0 || RTL != 0) {
        return 0;
    }
    
    var LTL:Number = this.leftToLeft;
    
    return 1.0 - Math.pow(LTL, 2);
}

public function set pan(panAmount:Number):void {
    this.leftToLeft = Math.sqrt(1 - panAmount);
    this.rightToRight = Math.sqrt(1 + panAmount);
    this.leftToRight = 0;
    this.rightToLeft = 0;
}
```

PR #10142, opened 2023-03-17
`NetStream` foundation/stub impls for external media

kmeisthax:
This PR aims to implement all the foundational work necessary for playing video from an FLV/F4V/etc file. It will not actually implement any video or audio playback functionality as that will require writing demuxers and the like. Consider it a more-functional-than-average stub.

We hold a list of all playing streams, and provide a `NetStreamObject` for AVM1 and 2. Calling `play` will load the targeted media, but since we have no demuxers/decoders yet this is a no-op. We do however track streams that are playing/paused. You can also attach a stream to a `Video`, which will make it render frames from the associated `NetStream`, but that stream can't demux yet so it just does nothing. The underlying API is built for unlocked timestep though!

This PR also includes stubs for AVM2 `NetStream`.


core/src/avm2/globals/flash/net/NetStream.as
```diff
@@ -0,0 +1,381 @@
...
+    public class NetStream extends EventDispatcher {
+        public static const CONNECT_TO_FMS = "connectToFMS";
+        public static const DIRECT_CONNECTIONS = "directConnections";
+
+        public function NetStream(connection:NetConnection, peer:String = CONNECT_TO_FMS) {
+
+        }
+
+        public function appendBytes(bytes:ByteArray) {
+            __ruffle__.stub_method("flash.net.NetStream", "appendBytes");
```
Lord-McSweeney, line 23: I think these can use  `stub_method` rather than `__ruffle__.stub_method`.

Lord-McSweeney:
I think it would be better to add `NetStreamObject` as a `NativeObject`, right?

kmeisthax:
@Lord-McSweeney So, as far as I can tell from how it's used so far, `NativeObject` is for things in AVM1 that would ordinarily be `ScriptObject`s, but we want to hold Rust fields for so that we can access them faster. All the object types that actually need to hold pointers to external data types don't use `NativeObject`, they implement their own `TObject` impl. And our `NetStream` type is very much external data (it references an ongoing media stream) so it should be a `TObject` impl, not a `NativeObject` variant.

Am I correct in thinking about it this way?

kmeisthax:
I believe you're correct, I'll move `NetStreamObject` in there.

core/src/avm2/globals/flash/net/NetStreamInfo.as
```diff
@@ -0,0 +1,110 @@
...
+
+        public function toString():String {
+            __ruffle__.stub_method("flash.net.NetStreamInfo", "toString")
+            return super.toString();
+        }
+
+        public function get audioBufferByteLength(): Number {
+            return this._audioBufferByteLength;
+        };
+ 	 	public function get audioBufferLength(): Number {
```
Herschel, line 37: Copy+paste added extra indentation

core/src/avm2/globals/globals.as
```diff
@@ -241,6 +240,9 @@ include "flash/net/NetGroupReceiveMode.as"
 include "flash/net/NetGroupReplicationStrategy.as"
 include "flash/net/NetGroupSendMode.as"
 include "flash/net/NetGroupSendResult.as"
+include "flash/net/NetStreamInfo.as"
+include "flash/net/NetStreamMulticastInfo.as"
+include "flash/net/NetStream.as"
```
Herschel, line 245: NetStream should come first in alphabetical order

core/src/streams.rs
```diff
@@ -0,0 +1,164 @@
...
+        StreamManager::ensure_paused(context, self);
+    }
+
+    /// Resume stream playback.
+    pub fn resume(self, context: &mut UpdateContext<'_, 'gc>) {
+        StreamManager::ensure_playing(context, self);
+    }
+
+    /// Resume stream playback if paused, pause otherwise.
+    pub fn toggle_pause(self, context: &mut UpdateContext<'_, 'gc>) {
```
Herschel, line 161: `toggle_paused`?

Herschel (commented):
Mostly a bunch of naming nits.

core/src/display_object/video.rs
```diff
@@ -271,6 +308,7 @@ impl<'gc> Video<'gc> {
                     }
                 }
             },
+            VideoSource::NetStream { .. } => unimplemented!(),
```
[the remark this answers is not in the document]
kmeisthax: Now that I think of it, it can get hit: if you grab a `Video` off the timeline and call `attachNetStream` on it. I'm going to make this function return silently for now.

core/src/player.rs
```diff
@@ -526,10 +532,12 @@ impl Player {
             });
 
             self.update_timers(dt);
+            self.update(|context| {
```
adrian17, line 535: Hey, I know this is super duper old, but: any reason this uses `update`, and not `mutate_with_update_context`?
Asking because the former does way more things (like triggering GC, which in practice does nothing due to incremental logic), and on desktop I saw it being called thousands of times per frame. If this _is_ needed for some specific part (queued avm1 actions?), I could try refactoring this to only do that part. If it's not needed, I'll make a PR to replace it with `mutate_with_update_context`.

kmeisthax: I can't think of any reason to not use the lighter-weight method. I don't remember if AVM1 streams actually use the event queue or not though.

PR #10158, opened 2023-03-17
core: Fix BitmapData.colorTransform

Herschel:
* Pixels with 0 alpha are not affected by color transforms. This was the main cause of the bad rendering in VVVVVV.
 * Color channels should be clamped to the 0-255 range after being transformed. Previously they were truncated and could be wrapped on overflow (32767 as a redOffset should always result in bright red and never wrap to black, for example).
 * Flash bug: A color transform with only alpha multiplier > 1 has no effect.

Fixes garbled text in VVVVVV.

TODO for the future: There are precision issues where the resulting color is off by a bit or so. Need to investigate whether this is from the premultiply/unmultiply, or the transform math.


Herschel:
Could use a sanity check here, I think this a Flash bug. If a color transform only has alphaMultiplier > 1, but all other values are default, the color transform will have no effect. But if any other value is non-default, then the alphaMultiplier functions as you'd expect.

```
import flash.display.BitmapData;
import flash.geom.ColorTransform;

var bData:BitmapData = new BitmapData(1, 1, true, 0x20ffffff);
var ct:ColorTransform = new ColorTransform(1, 1, 1, 2, 0, 0, 0, 0);
bData.colorTransform(bData.rect, ct);
trace(bData.getPixel32(0, 0).toString(16)); // Should be 0x40ffffff, but no change.

ct.redMultiplier = 1.01;
bData.colorTransform(bData.rect, ct);
trace(bData.getPixel32(0, 0).toString(16)); // Now works as expected, 0x40ffffff
```

I think there is an early-out case that avoids doing the color transform work for a default transform, but this incorrectly checks `alphaMultiplier < 1` instead of `alphaMultiplier != 1`.

Toad06:
Thanks for your work on this and welcome back. 😉

For reference, one of the issues you mentioned and fixed was a regression introduced in #7935 (likely the clamp one?) - I haven't made a bug report as the affected game was made with haXe with an heavy obfuscation (JPEXS wasn't able to decompile and show a single line of ActionScript).

Dinnerbone (approved):
It certainly smells like a bug, thanks for the tests!

Issue #10126, opened 2023-03-16
Axon: Stuck on loading screen


n0samu:
### Describe the bug

Link to SWF: http://i.notdoppler.com/files/axon.swf

[Axon](http://www.notdoppler.com/axon.php) is stuck on the loading screen:
![image](https://user-images.githubusercontent.com/71368227/225649262-6071dd26-aec8-4049-a2f9-b9e2b6dff008.png)

The error is:
`ERROR run_frame:run_all_phases_avm2: ruffle_core::display_object::movie_clip: Got RustError("Property does not exist: Multiname { ns: Single(Namespace(Namespace(\"flash.display\"))), name: Some(\"GraphicsSolidFill\"), params: [], flags: (empty) }") when constructing AVM2 side of movie clip of type Preloader`

### Expected behavior

The game should load.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

GraphicsSolidFill is also needed by #9955.

n0samu:
Now the game reaches the menu, but when trying to begin gameplay, this error appears:
```
ERROR run_frame:run_all_phases_avm2: ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "HIDE_COMPLETE", class: com.preloaded.mvc.view.events::ViewEvent, ptr: 0x2c59fd3f7d8 }) to handler FunctionObject(FunctionObject { ptr: 0x2c5f9c35008 }) : TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: minSpawnRange)
        at axon.game.engine::AxonGameEngine/start()
        at axon.view.views::GameView/init()
        at axon.view.mediators::GameMediator/showView()
        at com.preloaded.mvc.view::AbstractLocationMediator/onLocationEnter()
        at com.preloaded.mvc.view::AbstractLocationMediator/handleNotification()
        at Function/apply()
        at org.puremvc.as3.multicore.patterns.observer::Observer/notifyObserver()
        at org.puremvc.as3.multicore.core::View/notifyObservers()
        at org.puremvc.as3.multicore.patterns.facade::Facade/notifyObservers()
        at org.puremvc.as3.multicore.patterns.facade::Facade/sendNotification()
        at org.puremvc.as3.multicore.patterns.observer::Notifier/sendNotification()
        at com.preloaded.mvc.model::LocationProxy/startEnteringStage()
        at com.preloaded.mvc.model::LocationProxy/onLocationExited()
[...]

PR #10161, opened 2023-03-17
Implement various IGraphicsDraw classes

Dinnerbone:
This progresses #10126

They don't actually *do* anything by themselves, they're part of `Graphics.drawGraphicsData()`, but I'll implement that later


core/src/avm2/globals.rs
```diff
@@ -70,6 +70,12 @@ pub struct SystemClasses<'gc> {
     pub textformat: ClassObject<'gc>,
     pub graphics: ClassObject<'gc>,
     pub igraphicsdata: ClassObject<'gc>,
+    pub graphicsbitmapfill: ClassObject<'gc>,
```
adrian17, line 73: note: we don't really need to add _every_ single class we're ever planning to use in Rust - we can also look them up on demand:
```rs
let filters_namespace = Namespace::package("flash.filters", activation.context.gc_context);
let filter_class = Multiname::new(filters_namespace, "BitmapFilter");
let filter_class_object = activation.resolve_class(&filter_class)?;
```
Not saying which one is correct in this scenario, just a thing to keep in mind when adding them to SystemClasses.

Dinnerbone: I know, but I added these because I know we're going to need them for `drawGraphicsData` and `readGraphicsData`

core/src/avm2/globals/flash/display/GraphicsPath.as
```diff
@@ -0,0 +1,57 @@
...
+        public var data : Vector.<Number>;
+        public var winding : String;
+
+        public function GraphicsPath(commands:Vector.<int> = null, data:Vector.<Number> = null, winding:String = "evenOdd") {
+            if (commands == null) {
+                commands = new Vector.<int>();
+            }
+            if (data == null) {
+                data = new Vector.<Number>();
+            }
```
Herschel, line 14: Seems like Flash Player doesn't initialize these if null:
```
var path:GraphicsPath = new GraphicsPath();
trace(path.commands); // null
trace(path.data); // null
```

Dinnerbone: Fixed!

Herschel (approved):

Issue #10162, opened 2023-03-17
Stacko: Cannot start game


n0samu:
### Describe the bug

Link to SWF: https://static.arcadespot.com/wp-content/uploads/2015/11/stacko.swf

After the Armor Games intro in [Stacko](https://arcadespot.com/game/stacko/), the game shows nothing but a black screen.

Here is a log file from running the game: [stacko.log](https://github.com/ruffle-rs/ruffle/files/11001570/stacko.log)

### Expected behavior

The game should display the main menu.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

The same problem occurs with all of the Kings Kollege games: [Nambers](https://arcadespot.com/game/nambers/), [Fillz](https://arcadespot.com/game/fillz/), [Hivex](https://arcadespot.com/game/hivex/). These games use the Starling framework.

n0samu:
With #10166 and #10171, all of these games are now stuck on `flash.display3D.Context3D.setRenderToTexture`.

n0samu:
Fixed by #10185 (but see #10217 for a new issue with 2 of the mentioned games).

PR #10185, opened 2023-03-17
avm2: Implement Context3D.setRenderToTexture/setRenderToBackBuffer

Aaron1011:
In the process, I fixed a bug where we were clearing the depth
and stencil buffers with the incorrect value.

This makes Fancy Pants World 4 Part 1 playable to completion
(though there are still some rendering issues that need
to be fixed).


core/src/avm2/globals/flash/geom/Matrix3D.as
```diff
@@ -274,5 +274,8 @@ package flash.geom {
 			return new Matrix3D(this.rawData);
 		}
 
+		public function copyToMatrix3D(other:Matrix3D):void {
+			other.rawData = rawData
```
Herschel, line 278: Should this clone the data?

Aaron1011: It goes through the getter, so it automatically copies it.

Herschel (approved):

tests/tests/swfs/avm2/stage3d_texture/Camera3D.as
```diff
@@ -0,0 +1,710 @@
+package
+{
+	import flash.geom.Matrix3D;
+	import flash.geom.Vector3D;
+	
+	/**
+	 *   A 3D camera using perspective projection
+	 *   @author Jackson Dunstan
+	 */
```
Herschel, line 9: Is this code under a compatible license?

Aaron1011: This comes from the zip file at https://www.jacksondunstan.com/articles/1998
Unfortunately, there doesn't seem to be an explicit license. However, we're only using this a test - it's not linked into any part of Ruffle, and it's not distributed as part of releases.

n0samu: I guess we'd need to ask him
https://www.jacksondunstan.com/articles/1655

n0samu:
With this PR, Fillz and Stacko from #10162 are working great, thank you!
However, Hivex and Nambers crash when trying to enter the first level:
`thread 'main' has overflowed its stack`

tests/tests/swfs/avm2/stage3d_texture/Camera3D.as
Herschel: I'm fine with including it, but it'd be nice get an explicit license as we do distribute source on GH releases. Maybe we could toss the author an e-mail?

Issue #8903, opened 2022-12-23
Adventure Ponies: Stuck loading


waspennator:
### Describe the bug

I think the first time I remembered testing this, it kept cycling across a bunch of random screens in a non stop loop, now it just gets stuck on the loading screen on the latest ruffle.

https://www.numuki.com/game/mlp-adventure-ponies/

Error dispatching event EventObject(EventObject { type: "complete", class: flash.events::Event, ptr: 0x277bca8 }) to handler FunctionObject(FunctionObject(GcCell(Gc { ptr: 0x2809a64 }))) : RustError("Property does not exist: Multiname { ns: Multiple(Gc { ptr: 0x1cdc8d4 }), name: Some(\"unescape\"), params: [], flags: (empty) }")
$web_sys::features::gen_console::console::error_1::hd0bae7ea2787150d	@	018f336e:0x4d3ca9 ﻿

### Expected behavior

It should actually load up to the title screen.

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 108

n0samu:
Can you copy the error as text instead of an image and edit your issue? Thanks

waspennator:
Is that better now?

Lord-McSweeney:
Does this have something to do with the global `escape()` and `unescape()` functions in ActionScript?

n0samu:
Good catch, adding the `unimplemented` label to this issue.

waspennator:
> Good catch, adding the `unimplemented` label to this issue.
> 
> And yes waspennator, that's better.

Gotcha, I think #8543 might need the unimplemented tag since I believe that requires "netconnection"

n0samu:
Done, thanks

waspennator:
Game is still broken as of the latest nightly spewing this out in the logs.

[...]

PR #10232, opened 2023-03-18
avm2: Implement String call handler

evilpie:
Makes #8903 playable


core/src/avm2/globals/string.rs
```diff
@@ -92,6 +92,18 @@ pub fn class_init<'gc>(
...
 }
 
+pub fn call_handler<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    _this: Option<Object<'gc>>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    Ok(args
+        .get(0)
+        .unwrap_or(&Value::Undefined)
```
adrian17, line 102:
```
var cls = String;
trace(cls()); // ""
```
So this should be a `.unwrap_or(&Value::String("".into()))`, I think

evilpie: Indeed, thanks!

adrian17 (approved):

Herschel (approved):
Thank you!

PR #10236, opened 2023-03-19
Fixed glyph rendering with self intersecting fonts

Dinnerbone:
Previously fonts with self-intersecting lines would render like this:
image

This fixes that, and also respects Shapes with the relevant flag set (which Flash itself likely doesn't produce, but tools may)


Herschel (approved):
Commit msg says "Rename HAS_FILL_WINDING_RULE to HAS_FILL_WINDING_RULE"

Dinnerbone:
> Commit msg says "Rename HAS_FILL_WINDING_RULE to HAS_FILL_WINDING_RULE"

Fixed!

Herschel:
Thank you!

PR #10240, opened 2023-03-19
web: Allow ExternalInterface.call to call methods of primitive types (close #9105)

### LINKED ISSUE #9105: ExternalInterface.call cannot call methods of non-object types.
### Describe the bug

I have [a movie](https://mrcheeze.github.io/homestuck-ruffle-tester/#captchas,880,400) that contains the code:

```actionscript
pageURL = ExternalInterface.call("window.location.href.toString");
var tempvars:Array = pageURL.split("&");
```` 

In flash player, this finds the method by running the javascript `window.location.href.toString` which works. In Ruffle [this actually runs](https://github.com/ruffle-rs/ruffle/blob/1b71e288fd984ace46f69c1a2e27e2e2ed31333b/web/src/lib.rs#L1137) the javascript `Reflect.get(window.location.href,'toString')` which throws an error.

These are not equivalent, because js Reflect.get will fail if you pass *anything* other than an object as the first argument. In this case it fails because href is a string (even though it's a string that has the property in question).

### Expected behavior

[This flash](https://mrcheeze.github.io/homestuck-ruffle-tester/#captchas,880,400) should not print
`ERROR core/src/display_object/movie_clip.rs:2344 Error occured when running AVM2 frame script: RustError("Cannot access property http://adobe.com/AS3/2006/builtin::split of null or undefined")`
in the logs.

### Affected platform

Self-hosted version

### Operating system

Windows 11

### Browser

Mozilla Firefox

### Additional information

_No response_

MrCheeze:
See the description of https://github.com/ruffle-rs/ruffle/issues/9105.

`ExternalInterface.call("window.location.href.toString")`, for example, is a method that is sometimes used to get the current webpage url.

This does not work in ruffle currently, because we trigger the javascript code `Reflect.get(window.location.href, "toString")` (which fails because href is a primitive and not an object) instead of `window.location.href["toString"]` (which works correctly).

The only way I could find to fix this was to expose a helper method from javascript - wasm_bindgen doesn't have anything built-in which is *actually* equivalent to target[key].

### DIFF
--- a/web/packages/core/src/ruffle-imports.ts
+++ b/web/packages/core/src/ruffle-imports.ts
@@ -59,3 +59,14 @@ export function copyToAudioBufferInterleaved(
         i += 2;
     }
 }
+
+/**
+ * Gets a property of an arbitrary JavaScript value.
+ * This is necessary because Reflect.get does not work for primitive targets.
+ *
+ * @internal
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function getProperty(target: any, key: string): any {
+    return target[key];
+}
--- a/web/src/lib.rs
+++ b/web/src/lib.rs
@@ -1174,6 +1174,12 @@ impl ExternalInterfaceMethod for JavascriptMethod {
     }
 }
 
+#[wasm_bindgen(raw_module = "./ruffle-imports")]
+extern "C" {
+    #[wasm_bindgen(catch, js_name = "getProperty")]
+    pub fn get_property(target: &JsValue, key: &JsValue) -> Result<JsValue, JsValue>;
+}
+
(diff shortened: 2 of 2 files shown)


Dinnerbone (approved):

Lord-McSweeney:
I think Ruffle's `Reflect` polyfill is now unnecessary, right?

MrCheeze:
> I think Ruffle's `Reflect` polyfill is now unnecessary, right?

Didn't even realize there was one... but no, Reflect is still used in many other places where the type of the first arg is known to be an object.

Herschel (approved):

Herschel:
TIL that `Reflect.get` won't work on primitives 😄

PR #10244, opened 2023-03-19
avm2: Implement XML.parent()

AllinolCP:
Hi! this is my first PR to ruffle 
Please let me know what I should improve

progresses #8330


Dinnerbone:
Welcome and thank you!
Can you please add a test case for when it's undefined?

AllinolCP:
added!

evilpie:
You can simplify your tests by adding `XML.prettyPrinting = false` at the top of your test. toString in Flash and Ruffle should then match.

AllinolCP:
> You can simplify your tests by adding `XML.prettyPrinting = false` at the top of your test. toString in Flash and Ruffle should then match.

great idea, i will try this

core/src/avm2/globals/xml.rs
```diff
@@ -147,6 +149,19 @@ pub fn children<'gc>(
...
 }
 
+pub fn parent<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Option<Object<'gc>>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let xml = this.unwrap().as_xml_object().unwrap();
+    let node = xml.node();
+    if node.parent().is_none() {
```
evilpie, line 159: Using is_none and unwrap doesn't feel very rusty to me. You can probably use Option::map_or here.

AllinolCP: should be resolved now!

Aaron1011 (approved):

Herschel (approved):
Thank you, and welcome!

PR #10246, opened 2023-03-19
avm2: Fix TextField.getTextFormat arguments

Lord-McSweeney:
These are used in the Rust handler, but were not correctly set in the AS bindings, leading to errors such as "Attempted to call flash::text::TextField::getTextFormat with 2 arguments (more than 0 is prohibited)"


Herschel (approved):
Thank you!

Issue #2007, opened 2020-12-18
Adding XML functionality for XMAS!!!!


dpoetry:
I am having a problem with pulling in an XML file on top of the SWF file. This might seem overly complicated. But I'll try to explain. The html file loads p1.swf. Then p1.swf uses LoadMovie to load a p1add.swf file. Then the p1add.swf file uses xml to load a bunch of images/jpgs. So either the p1add.swf isn't loading properly or the XML file inside the p1add.swf file isn't loading.

I am guessing the issue is it not loading the XML file. As other sections of this work use the LoadMovie function just fine and load external SWFs on top of other SWFs.

Here is the overall work: http://www.dpoetry.com/ruffletest/wittenoom2/starthere.html

And here is the page in with the XML files not loading: http://www.dpoetry.com/ruffletest/wittenoom2/artwork/p1.html

And here are the relevant files:

The HTML file: http://www.dpoetry.com/ruffletest/wittenoom2/artwork/p1.html
The primary SWF file: http://www.dpoetry.com/ruffletest/wittenoom2/artwork/p1.swf
The LoadMovie SWF file: http://www.dpoetry.com/ruffletest/wittenoom2/artwork/p1add.swf
The XML File: http://www.dpoetry.com/ruffletest/wittenoom2/artwork/images.xml
All the files: http://www.dpoetry.com/ruffletest/wittenoom2/artwork/

LASTLY: This is what the page above should look and function like: http://www.secrettechnology.com/wittenoom/artwork/p1.html. (will have to enable flash)

NOTE: The above URL from secrettechnology does not have Ruffles loaded. So everything still works within Flash.

Does any of that make sense?

cheers, Jason Nelson

Herschel:
Haven't investigated thoroughly yet, but we should support XML, but don't currently support loading external images.

jrob007:
I have a separate XML file that gets loaded and parsed into a frame as well. It relies primarily on `XML.load()`, `XML.onLoad()` and `XML.parseXML()`. However according to https://github.com/ruffle-rs/ruffle/issues/286 only the `parseXML` method is currently supported by Ruffle. My animation loads, but the textbox simply says "undefined" as it's not loading any external XML file.

Herschel:
The tracking issue is out of date; we should `XML.load/onLoad/parseXML`, so the issue is possibly elsewhere. If possible, could you post a sample SWF? Thank you!

dpoetry:
[...]

PR #10271, opened 2023-03-20
avm2: Fix TypeError message

Herschel:
Add missing `Error #2007: ` message to null parameter errors and edit a test to check this.

Perhaps we should add a `format_error!(id, msg, ...)` macro that automatically glues in the ID# to the front of the message.


Dinnerbone:
Just FYI there's quite a few places we've deliberately decided not to match the error message exactly, either because Flash had a wrong message entirely or it just wasn't as informative

Adding error code seems reasonable to me, but we may not be able to match things 1:1 in tests

Dinnerbone (approved):

adrian17:
> Perhaps we should add a format_error!(id, msg, ...) macro that automatically glues in the ID# to the front of the message.

We don't need a macro, we can format the string when constructing the error itself so we don't need to handle it manually in every call site.

Issue #10192, opened 2023-03-18
Wentworth: Stuck on loading screen


n0samu:
### Describe the bug

Link to SWF: http://i.notdoppler.com/files/wentworth.swf

[Wentworth](http://www.notdoppler.com/wentworth.php) is stuck on the loading screen with this error:
`ERROR run_frame:run_all_phases_avm2: ruffle_core::display_object::movie_clip: Got RustError("Cannot call null or undefined method copy of class XMLList") when constructing AVM2 side of movie clip of type MainSWF`

### Expected behavior

The game should load.

### Affected platform

Desktop app

### Operating system

Windows 10

PR #10273, opened 2023-03-20
avm2: Implement XMLList.copy()

yoganlava:
First PR, open for feedback.

Progresses: #10192


Dinnerbone:
Welcome and thank you!

Would you be able to make a test for this, or would you like some help with that?

yoganlava:
No problem. Would you like me to add it to the existing xml_basic test or have it exist on it's own?

n0samu:
I'm pretty sure either one would be fine!

n0samu:
That's fine, but could you keep the code inside the run() function of the Test class? The FLA has almost nothing in it, just this code on frame 1:
`Test.run();`
I think you could do the same in your AS file, thus avoiding the need to move all of the code outside the run() function.

core/src/avm2/globals/xml_list.rs
```diff
@@ -232,3 +232,13 @@ pub fn text<'gc>(
...
     Ok(XmlListObject::new(activation, nodes, Some(xml_list.into())).into())
 }
+
+pub fn copy<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Option<Object<'gc>>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let list = this.unwrap().as_xml_list_object().unwrap();
+    let children = (*list.children()).clone();
```
Bale001, line 242: I don't think this is correct - the [docs](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/XMLList.html#copy()) say `The copy is a duplicate of the entire tree of nodes.`, but this only clones the first layer.

yoganlava:
Closing this pull request as I might still be too new to impl a recursive deep_clone

PR #10278, opened 2023-03-20
avm2: Stub NetConnection.close

Lord-McSweeney:


n0samu:
Looks like you also removed the NetConnection constructor function for some reason? Why?

Lord-McSweeney:
The compiler should automatically add that, and no initialization of anything is actually done in the constructor, as far as I can see from the documentation.

Bale001 (approved):

PR #10281, opened 2023-03-20
avm1: Minimal implementation of `XML.getBytesTotal` and `XML.getBytesLoaded`

### LINKED ISSUE #2089: Error on http://extratoxic.com/_shared/Nintendo/06_NDO_013_MissionInSnowdriftland/
The console was spammed with these messages in this order.

```
Object method onLoading is not callable core.ruffle.js:1:30473
Object method getBytesLoaded is not callable core.ruffle.js:1:30473
Object method getBytesTotal is not callable core.ruffle.js:1:30473
```
The game stays on a white screen.
They did not stop until I closed the page.

Toad06:
These methods appear to work exactly like the ones in the `LoadVars` class, so we can support them by setting the `_bytesTotal` and `_bytesLoaded` properties on the XML instance. Note that when the load is complete, these properties are set in `core/src/loader.rs`, something which was already added in #8573 (but I haven't noticed this would also impact the XML class back then).

Fixes #2089 (the link mentioned in the issue is dead but the archive can be downloaded here: https://archive.org/details/06-ndo-013-mission-in-snowdriftland_202101).


Herschel (approved):
Thank you!

Issue #10156, opened 2023-03-17
Zombie Hooker XXX-mas: White Screen Softlock


waspennator:
### Describe the bug

https://www.newgrounds.com/portal/view/521563?emulate=flash
I can see the preloader for like half a second, then it vanishes to a white screen and tosses this up in the logs.

Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x45122d0 }) to handler FunctionObject(FunctionObject { ptr: 0x786db8 }) : RustError("Property does not exist: Multiname { ns: Single(Namespace(Namespace(\"flash.xml\"))), name: Some(\"XMLDocument\"), params: [], flags: MultinameFlags(0x0) }")

### Expected behavior

It should load normally without doing that

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 110

Lord-McSweeney:
Will be progressed by #10112 but then needs `XMLDocument.ignoreWhite`.

waspennator:
Game seems to be playable now with #10299 merged in.

PR #10299, opened 2023-03-21
avm2: Shim flash.xml.XMLNode using XML

evilpie:
This is more of a Proof-of-Concept. This approach won't preserve the XMLNode objects because they are recreated every time when accessing e.g. `childNodes`. It does work with #10156 though.


core/src/avm2/globals/flash/xml/XMLNode.as
```diff
@@ -1,6 +1,49 @@
 package flash.xml
 {
    public class XMLNode {
+      public var _xml: XML;
+      public var _parent: XMLNode = null;
+      public var _index: uint = 0;
```
Lord-McSweeney, line 6: I think these properties might be better made `internal`.

evilpie:
I actually think this might be good enough to land now. I introduced caching for the childrens, which avoids the problem of creating new (different of course) XMLNodes on every access.

core/src/avm2/globals/flash/xml/XMLNode.as
```diff
@@ -1,6 +1,82 @@
...
+         if (_isDocument) {
+            return null;
+         }
+         return _xml.localName();
+      }
+
+      public function get nodeType(): uint {
+         switch (_xml.nodeKind()) {
+            case "attribute":
+            default:
```
Lord-McSweeney, line 59: I think the two matches for the `switch` statement should be at the bottom of the switch statement, right? Otherwise the `default` case can be met every time and throw the Error.

core/src/avm2/globals/flash/xml/XMLDocument.as
```diff
@@ -1,6 +1,21 @@
...
+      public function XMLDocument(input: String = null) {
+         super(0, null);
+         _isDocument = true;
+         parseXML(input);
+      }
+
+      public function parseXML(input: String): void {
+         // Unlike the new AS3 XML class, XMLDocument allows parsing
+         // multiple elements like <a>1</a><b>2</b>. So introduce a wrapper.
+         _xml = new XML("<xml>" + input + "</xml>");
```
Lord-McSweeney, line 16: I think this should set `XML.ignoreWhite = this.ignoreWhite`- eventually Ruffle will implement it.

core/src/avm2/globals/flash/xml/XMLDocument.as
```diff
@@ -1,6 +1,21 @@
 package flash.xml
 {
    public class XMLDocument extends XMLNode {
```
Herschel, line 3: I'd add a comment explaining that this is a quick shim and that eventually we'd want to use a proper legacy XML implementation.

Herschel (approved):
Thank you! This is a clever way to get the legacy XML up and running.

Issue #10086, opened 2023-03-15
Totems Awakening 2: Cannot start game


n0samu:
### Describe the bug

Link to SWF: http://i.notdoppler.com/files/totemsawakening2.swf

After clicking the Play button in [Totems Awakening 2](http://i.notdoppler.com/files/totemsawakening2.swf), it goes to a black screen. This error appears several times in the terminal:
```
ERROR ruffle_core::avm2::activation: TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: http://adobe.com/AS3/2006/builtin::indexOf)
        at GameSounds/initIDs()
        at GameSounds/init()
        at Main()
        at Preloader/startup()
        at Preloader/onPlayButtonClick()
```

### Expected behavior

The main menu should appear.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

Here is the relevant function:
```as
private function initIDs() : void
{
    var _loc3_:XML = null;
    var _loc4_:String = null;
    this.soundIDs = [];
    this.musicIDs = [];
    var _loc1_:XML = describeType(GameSounds);
    var _loc2_:XMLList = _loc1_.elements("constant");
    for each(_loc3_ in _loc2_)
    {
        if((_loc4_ = String(GameSounds[_loc3_.attribute("name")])).indexOf("M_") == 0)
        {
            this.musicIDs.push(_loc4_);
        }
        else
        {
            this.soundIDs.push(_loc4_);
        }
    }
}
```

n0samu:
Fixed by #10306

PR #10306, opened 2023-03-21
avm2: Fully implement XML.elements()

yoganlava:
Might not be idiomatic rust, but will progress #10086.


core/src/avm2/globals/xml.rs
```diff
@@ -164,13 +164,25 @@ pub fn parent<'gc>(
...
-    _args: &[Value<'gc>],
+    args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
     let xml = this.unwrap().as_xml_object().unwrap();
+    let element_name = if args[0] == Value::Undefined {
+        AvmString::from("*")
+    } else {
+        args[0].coerce_to_string(activation)?
+    };
+    let is_asterisk = element_name == AvmString::from("*");
```
evilpie, line 175: You can use `name_to_multiname` to simplify your code.

core/src/avm2/globals/xml.rs
```diff
@@ -164,13 +164,26 @@ pub fn parent<'gc>(
 pub fn elements<'gc>(
     activation: &mut Activation<'_, 'gc>,
     this: Option<Object<'gc>>,
-    _args: &[Value<'gc>],
+    args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
     let xml = this.unwrap().as_xml_object().unwrap();
+    let element_name = if args[0] == Value::Undefined {
+        AvmString::from("*")
```
Lord-McSweeney, line 171: Are you sure this is the correct behavior if `XML.elements(undefined)` is called? Otherwise I don't think `XML.elements()` can be called without arguments.

core/src/avm2/globals/xml.rs
yoganlava: Done, thank you for pointing me to the function. I also made name_to_multiname resolve an undefined into a "*"

core/src/avm2/globals/xml.rs
```diff
@@ -105,11 +105,11 @@ pub fn name_to_multiname<'gc>(
         }
     }
 
-    let name = name.coerce_to_string(activation)?;
-    Ok(if &*name == b"*" {
+    let name_string = name.coerce_to_string(activation)?;
+    Ok(if &*name_string == b"*" || *name == Value::Undefined {
```
evilpie, line 109: I am afraid this is not correct. name_to_multiname roughly maps to ToXMLName (10.6 in [ECMA-357](https://www.ecma-international.org/publications-and-standards/standards/ecma-357/)), which says throw for `undefined`. (To be fair we don't correctly do that at the moment either)

Per spec we should be handling undefined in `elements`.

Herschel (approved):
Thank you!

PR #10308, opened 2023-03-22
avm2: Fix double borrow when setting local XMLList property

yoganlava:
Fixes double borrow error seen here: #10303
Prevented an error from being thrown due to a double borrow.


n0samu (approved):
Thank you! 😃

Bale001 (approved):

PR #10322, opened 2023-03-23
core: Make `ExternalInterface.call` support `undefined` values in browsers

### LINKED ISSUE #8621: `ExternalInterface.call` should support `undefined` values on web
[Sample (SWF)](https://github.com/ruffle-rs/ruffle/files/10096478/ext.zip)

AS2:
```
var text = flash.external.ExternalInterface.call("getVal");
trace(text);
```
JS:
```
function getVal() { return undefined; }
```

---

This sample logs:
- "undefined" in the Flash Player plugin.
- "null" in the Ruffle extension.

The Flash Player projector also logs "null" but that might be because `ExternalInterface.available` is `false`.

In general, `undefined` should be the default return value on web, this happens in all of the following cases:
- `flash.external.ExternalInterface.call()`
- `flash.external.ExternalInterface.call("jsFunctionThatDoesntExist")`
- `flash.external.ExternalInterface.call("jsFunctionWithNoReturnValue")`
- `flash.external.ExternalInterface.call("jsFunctionWithAnErrorInside")`

Toad06:
Fixes #8621. The behavior can be verified using [this sample](https://github.com/ruffle-rs/ruffle/files/11045693/externalUndefined.zip).

### DIFF
--- a/core/src/external.rs
+++ b/core/src/external.rs
@@ -20,6 +20,7 @@ use std::collections::BTreeMap;
 #[derive(Debug, Clone, PartialEq)]
 pub enum Value {
     Null,
+    Undefined,
     Bool(bool),
     Number(f64),
     String(String),
@@ -118,12 +119,26 @@ impl From<Vec<Value>> for Value {
 }
 
 impl Value {
+    pub fn check_avm1_value(value: Avm1Value) -> Avm1Value {
+        match value {
+            Avm1Value::Undefined => {
+                if cfg!(not(target_family = "wasm")) {
+                    Avm1Value::Null
+                } else {
+                    value
+                }
+            }
+            _ => value,
+        }
+    }
+
     pub fn from_avm1<'gc>(
         activation: &mut Avm1Activation<'_, 'gc>,
         value: Avm1Value<'gc>,
     ) -> Result<Value, Avm1Error<'gc>> {
         Ok(match value {
-            Avm1Value::Undefined | Avm1Value::Null => Value::Null,
+            Avm1Value::Null => Value::Null,
+            Avm1Value::Undefined => Value::Undefined,
             Avm1Value::Bool(value) => value.into(),
             Avm1Value::Number(value) => value.into(),
             Avm1Value::String(value) => Value::String(value.to_string()),
@@ -154,6 +169,7 @@ impl Value {
     pub fn into_avm1<'gc>(self, activation: &mut Avm1Activation<'_, 'gc>) -> Avm1Value<'gc> {
         match self {
             Value::Null => Avm1Value::Null,
+            Value::Undefined => Avm1Value::Undefined,
             Value::Bool(value) => Avm1Value::Bool(value),
             Value::Number(value) => Avm1Value::Number(value),
             Value::String(value) => {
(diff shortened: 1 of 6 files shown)


core/src/external.rs
```diff
@@ -186,9 +202,23 @@ impl Value {
         }
     }
 
+    pub fn check_avm2_value(value: Avm2Value) -> Avm2Value {
+        match value {
+            Avm2Value::Undefined => {
+                if cfg!(not(target_family = "wasm")) {
```
Herschel, line 208: Hmm, can we simply return `undefined` on desktop as well? i.e. is the `null` case taken care of by the fact that there is no `ExternalInterface`, like you said?

Herschel:
Haven't looked thoroughly, but is it possible for us to test this in `tests\tests\external_interface\test.rs` too?

core/src/external.rs
Herschel: I'm wondering if we can instead do something like `ExternalInterface::available` in external.rs that returns `!self.providers.is_empty()`, and then the AVM's ExternalInterface.call` it returns `null` in that case. That way we don't need the `check_avm1_value`/`check_avm2_value` I believe?

Herschel: Looked into this a bit more, and the docs say (https://open-flash.github.io/mirrors/as2-language-reference/flash/external/ExternalInterface.html#call()):
`If the desired function is not available, the call returns null;`

So seems like the behavior should be:
 * If the external interface is not available or the function does not exist in the external interface, return `null`.
 * Otherwise, return what the function returns, which may be `undefined`.
 

I suspect in the browser, every function "exists" because it's JavaScript land, so this will always try to execute the function, which returns `undefined` naturally when it calls a non-existent function. I guess `JavascriptInterface::get_method` should return a dummy function/object that evaluates to `undefined` even when the method doesn't exist.

Toad06: I made the suggested changes as they definitely make sense. This works like a charm on web, returning "undefined" in the expected cases. Unfortunately, I'm also getting test failures. Here's one of them:

```diff
/// callWith() end
-undefined
+null
```
The Rust part of this test is defined here:
https://github.com/ruffle-rs/ruffle/blob/86395f194fd48cfbd46e01367125a3140fc8b5f7/tests/tests/external_interface/mod.rs#L26-L37

The ActionScript part is in [avm1/external_interface](https://github.com/ruffle-rs/ruffle/tree/master/tests/tests/swfs/avm1/external_interface):
```
function callWith(name, payload) {
   // (...)
   trace("/// callWith() end");
}
```

So, from my understanding, `callback.call(context, "callWith", ...)` runs and calls `callWith` which returns `undefined` as the function has no return value. This then causes `do_reentry` to return `ExternalValue::Undefined` instead of `ExternalValue::Null` and causes the test failure.

Herschel: I think the test output needs to be changed now, as it was testing for the incorrect behavior. Instead it should the test ouput should have `undefined` there. I verified in the browser that calling `player.foo()` from JS for a function that returns nothing will result in `undefined` in the Flash Player.

Herschel (approved):
Thanks for your work on this!

Toad06:
Thank you for the review and help. 😉

PR #10330, opened 2023-03-23
core: Return false for shape hitTests on zero-scale graphics

### LINKED ISSUE #6906: Super Mario 63: problems when entering the castle
### Describe the bug

To reproduce the bugs:
1. Load the [SWF file](https://www.play-games.com/files/swf/Super-Mario-63_1402146483_1410902473.swf) with Ruffle
2. Play through the tutorial (takes 10-20 minutes or so) and reach the castle. Press the up arrow to enter the castle

You should notice these problems:
* The transition to the next level is extremely slow (takes 1-2 minutes even on high-end hardware)
* During this transition, memory usage climbs significantly (in my testing, up to 4GB)
* Once the level starts, many of the objects in the level are missing and part of the background is black/blank

It's still possible to reach the next door. But if you go through that door, there will be another extremely slow transition, and then Mario will end up floating in the sky and unable to move (see screenshot below)
image

### Expected behavior

None of these problems should occur.

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Firefox 100

### Additional information

I've attached the log from my browser console (log level: info) below, but it may not be very helpful
[sm63.log](https://github.com/ruffle-rs/ruffle/files/8644591/sm63.log)

Herschel:
If a graphic has a `scaleX` or `scaleY` of 0, a shape `hitTest(x, y, true)` should return false, even on the exact location of the object. Note that a bounds `hitTest(x, y, false)` still returns true on the exact location of the object. Bail out on the hitTest immediately when we detect no width or height.

Fix with the following changes:
 * Change `Matrix::invert` to `Matrix::inverse` and make it fallible, returning `None` for non-invertible matrices. (An object with zero scale will have a non-invertible matrix for its transform).
 * Change `DisplayObject::global_to_local_matrix` and `DisplayObject::global_to_local` fallible. These return `None` when the object has zero scale.
 * Adjust all places in `core` that used these to properly handle these methods returning `None`. All overrides of `hit_test_shape` will return false in this case.
 
Bonus:
 * Fix `clip.globalToLocal(point)` when the clip has zero scale: point will be unaffected.
 * Fix `clip._xmouse`/`_ymouse`/etc. when the clip has zero scale: a 1/20 scale matrix (twips-to-pixels) is used.
   * Added `DisplayObject::mouse_to_local` to handle these cases.

Fixes #6906.


Lord-McSweeney:
Does this happen in AVM2, too?

Herschel:
I decided on a more thorough way to solve this by making `Matrix::inverse`, `DisplayObject::local_to_global` return an `Option`. This way any caller has to explicitly handle the case when the object is zero scale and the matrix has no inverse.

core/src/display_object.rs
```diff
@@ -799,8 +798,20 @@ pub trait TDisplayObject<'gc>:
     }
 
     /// Converts a local position on the stage to a local position on this display object
-    fn global_to_local(&self, global: (Twips, Twips)) -> (Twips, Twips) {
-        self.global_to_local_matrix() * global
+    /// If the object has zero scale, this returns
```
Toad06, line 801: Not a big deal, but this comment seems incomplete (returns `None`?).

Toad06 (approved):
Thanks for your work on this. This looks good to me based on testing, I haven't found any issue or regression.

Mentioning that this game now works would be a nice addition to https://github.com/ruffle-rs/ruffle-rs.github.io/issues/132. 😉

szefkaf:
@Toad06 please note that the quality of this game dropped dramatically to 18-20fps sometimes after these fixes when more sprites are on the screen. In the version I fixed it stays at over 30 frames per second just like in the original Flash version. Please compare Level 1 Bob-omb Battlefield (left door in Castle) in both versions. In the new version of Ruffle, even the jump sound can't keep up with real Mario's jump on mid-range laptops. I'm a little afraid that it will have its consequences in other games that will simply be too slow to play especially on mobiles.

PR #10333, opened 2023-03-23
web: Don't polyfill with self-hosted Ruffle while extension is loading

danielhjacobs:
Finally found a way to make this work.

### DIFF
--- a/web/packages/core/src/plugin-polyfill.ts
+++ b/web/packages/core/src/plugin-polyfill.ts
@@ -150,6 +150,15 @@ export const FLASH_PLUGIN = new RufflePlugin(
     "ruffle.js"
 );
 
+/**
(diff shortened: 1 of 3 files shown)


Herschel:
Does this fix the various conflicts with self-hosted+extension loading in different orders?

danielhjacobs:
It stops the self-hosted build from loading if the extension is loaded (except in such cases where the plugin polyfill is slower to load than the self-hosted build, which can on rare occasions happen in Chrome). So it should make it predictable, and in conjunction with the extension config PR, it should load the correct configuration settings. If there are still issues, they can be fixed in the PublicAPI negotiate function.

Herschel:
Thanks for explaining!

Herschel (approved):

Herschel:
Thank you for your continued work!

Issue #9493, opened 2023-02-10
Strange graphic error in bubble tanks 2


rredford:
### Describe the bug

![image](https://user-images.githubusercontent.com/151014/217970687-2bbb2f40-bd3c-4bd7-9345-078b0b71b9da.png)

The bubble is very thick for some reason. 

### Expected behavior

See the image, right side for correct graphics.

### Affected platform

Desktop app

### Operating system

xfce mint 21.1

### Additional information

[bubble-tanks-2-1920.swf.zip](https://github.com/ruffle-rs/ruffle/files/10702676/bubble-tanks-2-1920.swf.zip)

rredford:
Note same thing happens on bubble tanks 1, I only reported this one as I feel that it is exact same thing anyway. I don't know if other versions have same issue also.

Lord-McSweeney:
See #7369.

n0samu:
This is not the stroke width issue. This is actually the same issue as #6784. On frame 15, the `PlaceObject2 (585)` tag has a clipDepth, which creates the more subtle bubble effects in the top-left and bottom-right corner of the screen while preventing the bubble sprite from blocking out the middle of the screen.

rredford:
Ah I think it's used for effect of exiting the bubble. Yeah think thats the case.

n0samu:
Unfortunately #9604 regressed this further. Here's what it looks like in Ruffle now:
![image](https://user-images.githubusercontent.com/71368227/220313549-1ffede77-0b72-4ab0-8879-7e8c2689d984.png)

n0samu:
Here is an SWF of just frame 15 of the original game, exported from JPEXS: [frame15.zip](https://github.com/ruffle-rs/ruffle/files/10793152/frame15.zip)
This one displays correctly in Ruffle. So it must be some script elsewhere in the movie that is confusing Ruffle, rather than the tags on frame 15.

Toad06:
I was able to fix the regression by moving the line `child.set_clip_depth(context.gc_context, 0);` at the beginning of the next `if` block:
https://github.com/ruffle-rs/ruffle/blob/d7c842a1393b9824ae9f455b60843dc76a341b4d/core/src/display_object/container.rs#L834-L840
With this change, the Garfield game in #9604 (https://archive.org/details/comiccreator_swf) still works. So that's better, but I still have no idea whether this is what Flash does or not. :)

---

About the initial report, the issue is here:
[...]

PR #10339, opened 2023-03-24
core: Correct `swapDepths`

Toad06:
Fixes the regression highlighted by nosamu in https://github.com/ruffle-rs/ruffle/issues/9493#issuecomment-1438194951 while making sure it doesn't break the game mentioned in #9604 (https://archive.org/details/comiccreator_swf).

### DIFF
--- a/core/src/display_object/container.rs
+++ b/core/src/display_object/container.rs
@@ -839,10 +839,10 @@ impl<'gc> ChildContainer<'gc> {
     ) {
         let prev_depth = child.depth();
         child.set_depth(context.gc_context, depth);
-        child.set_clip_depth(context.gc_context, 0);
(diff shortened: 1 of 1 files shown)


Lord-McSweeney:
Can you add a test for this?

n0samu:
It's a super simple fix but I don't know that crafting a test would be that simple. I think this PR is fine as-is, though adding tests is always appreciated. Tests can always be added in a followup, after all.

Toad06:
I gave a try at writing a test for this, but nothing successful. What causes the issue is likely a combination of factors and would require to look into the file more deeply (or maybe that I'm just missing something 😅).

Aaron1011 (approved):

Issue #9226, opened 2023-01-20
Phoenotopia - Unplayable


Kz-92:
### Describe the bug

https://user-images.githubusercontent.com/103858019/213604214-82cc8887-20e6-4e06-87bb-fa26d7543e43.mp4

This is how Phoenotopia should play (flash player 32)

https://user-images.githubusercontent.com/103858019/213604329-acae5154-73e7-4474-b1c7-587977555c6a.mp4

This is how Phoenotopia plays on Ruffle 2023-01-19

On the Desktop version of Ruffle, there is no log, strangely enough
On the Web version of Ruffle, this is the log:

```
WARN core/src/avm2/globals/flash/system/security.rs:41 Security.loadPolicyFile not implemented ruffle_web-wasm_extensions.js:671:20
WARN core/src/avm2/globals/toplevel.rs:42 LocalConnection.connect is not implemented ruffle_web-wasm_extensions.js:671:20
WARN core/src/avm2/globals/toplevel.rs:42 URLRequest.data setter is not yet implemented ruffle_web-wasm_extensions.js:671:20
WebGL warning: getSyncParameter: ClientWaitSync must return TIMEOUT_EXPIRED until control has returned to the user agent's main loop. (only warns once) 31
After reporting 32, no further warnings will be reported for this WebGL context.
WARN core/src/player.rs:591 AVM2 Context menu callbacks are not implemented
```
to reproduce the issue, just load the Phoenotopia SWF
https://farm.maxgames.com/phoenotopiaMjI1Mw==.swf

### Expected behavior

The game should load to the main menu and be playable.

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

Mozilla Firefox

ActionWavele:
wow, that save file, did you beat the game?

Lord-McSweeney:
Needs `unescape`.

waspennator:
Even with unescape implemented, ruffle still tosses this up in the logs as of the latest nightly.

Error dispatching event EventObject(EventObject { type: "commandComplete", class: com.newgrounds::APIEvent, ptr: 0x220da80 }) to handler FunctionObject(FunctionObject { ptr: 0x1d2d178 }) : RustError("Property does not exist: Multiname { ns: Single(Namespace(Namespace(\"\"))), name: Some(\"unescape\"), params: [], flags: (empty) }")

Lord-McSweeney:
The panic message is `Stage must always have a root movie`.

[...]

PR #10341, opened 2023-03-24
avm2: Trim xml text nodes when being parsed

yoganlava:
Trims XML text when parsing.
Makes #9226 playable.


tests/tests/swfs/avm2/xml_basic/Test.as
```diff
@@ -1,5 +1,6 @@
 ﻿package {
 	public class Test {
+		import flash.utils.*;
```
Lord-McSweeney, line 3: I think this line is unnecessary.

yoganlava: Yeah, you're right. Remnants from some other testing I was doing at the same time

Lord-McSweeney (approved):

core/src/avm2/e4x.rs
```diff
@@ -222,7 +222,15 @@ impl<'gc> E4XNode<'gc> {
                     let is_whitespace_char = |c: &u8| matches!(*c, b'\t' | b'\n' | b'\r' | b' ');
                     let is_whitespace_text = text.iter().all(is_whitespace_char);
                     if !(text.is_empty() || ignore_white && is_whitespace_text) {
-                        let text = AvmString::new_utf8_bytes(activation.context.gc_context, &text);
+                        let text = if matches!(event, Event::Text { .. }) {
+                            AvmString::new_utf8(
+                                activation.context.gc_context,
+                                std::str::from_utf8(&text).unwrap().trim(),
```
Herschel, line 228: It'd be nice to avoid `std::str::from_utf8` here and instead trim the byte slice directly (you could write something like [`trim_ascii_start`](https://doc.rust-lang.org/src/core/slice/ascii.rs.html#97), which is currently nightly only but can just be copied here). This would avoid the extra UTF-8 check and potential panic if the text isn't valid UTF-8.

Herschel (commented):
Thank you!

core/src/avm2/e4x.rs
```diff
@@ -278,7 +300,10 @@ impl<'gc> E4XNode<'gc> {
             let is_whitespace_char = |c: &u8| matches!(*c, b'\t' | b'\n' | b'\r' | b' ');
             let is_whitespace_text = text.iter().all(is_whitespace_char);
             if !(text.is_empty() || ignore_white && is_whitespace_text) {
-                let text = AvmString::new_utf8_bytes(activation.context.gc_context, text);
+                let text = AvmString::new_utf8_bytes(
+                    activation.context.gc_context,
+                    if is_text { trim_ascii(text) } else { text },
```
Herschel, line 305: I think needs to be `if is_text && ignore_white {` so that it wouldn't trim if `XML.ignoreWhitespace` is false. Although we don't have that hooked up yet (line 245 hardcodes it to true at the moment)

Herschel:
Oops, sorry I didn't notice this sooner, but this depends on the `XML.ignoreWhite` setting. Currently `ignore_white` is hardcoded to true -- would you want to try hooking that up?

yoganlava:
I only decided to hookup ignoreWhitespace. I can probably hook up the other static properties in another PR

core/src/avm2/e4x.rs
```diff
@@ -239,10 +239,14 @@ impl<'gc> E4XNode<'gc> {
...
 
+        let namespace = activation.avm2().public_namespace;
+        let xml_class = activation.resolve_class(&Multiname::new(namespace, "XML")).unwrap();
+
         // FIXME - look these up from static property and settings
         let ignore_comments = true;
         let ignore_processing_instructions = true;
-        let ignore_white = true;
+
+        let ignore_white = xml_class.get_public_property("ignoreWhitespace", activation).unwrap_or(Value::Bool(true)).coerce_to_boolean();
```
Aaron1011, line 249: This isn't quite right - the value should be read when an `XML` object is constructed, and stored in the `settings` object. For now, I think it's good enough to keep hardcoding `ignore_white = true`, and check it before trimming.

Aaron1011 (approved):

PR #10345, opened 2023-03-24
render: Replace `ColorTransform` with `swf::ColorTransform`

relrelb:


swf/src/types/color_transform.rs
```diff
@@ -1,39 +1,24 @@
+use crate::{Color, Fixed8};
 use std::ops::{Mul, MulAssign};
-use swf::{Color, Fixed8};
 
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq)]
```
Herschel, line 4: I'd prefer to keep `Copy`. `ColorTransform` is only 16 bytes, which is the size of a fat pointer.

Herschel (approved):
Thank you!

swf/src/types/color_transform.rs
```diff
@@ -77,59 +62,59 @@ impl Default for ColorTransform {
     }
 }
 
-impl Mul for ColorTransform {
-    type Output = Self;
+impl Mul for &ColorTransform {
```
Herschel, line 65: Actually now that this is `Copy`, can this change back to impl `for ColorTransform`?

Herschel (approved):
Thank you!

PR #6935, opened 2022-05-10
web: Fix some warnings

Herschel:
Fix warnings caused by desktop-only code.

PR #10347, opened 2023-03-24
ci: Deny warnings on Web

relrelb:
As suggested in https://github.com/ruffle-rs/ruffle/pull/6935#pullrequestreview-967181111.

### DIFF
--- a/exporter/src/main.rs
+++ b/exporter/src/main.rs
@@ -143,8 +143,7 @@ fn take_screenshot(
                     .renderer_mut()
                     .downcast_mut::<WgpuRenderBackend<TextureTarget>>()
                     .unwrap();
-                // Use straight alpha
-                renderer.capture_frame(false)
+                renderer.capture_frame()
             }) {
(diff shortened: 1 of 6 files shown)


exporter/src/main.rs
```diff
@@ -143,8 +143,7 @@ fn take_screenshot(
                     .renderer_mut()
                     .downcast_mut::<WgpuRenderBackend<TextureTarget>>()
                     .unwrap();
-                // Use straight alpha
-                renderer.capture_frame(false)
+                renderer.capture_frame()
```
Herschel, line 146: Are these changes mistakenly included in this PR?

relrelb: Actually no. I started with fixing a warning using `#[cfg(not(target_family = "wasm"))]`, and got carried away a bit...

I can split these changes to a separate PR if it makes sense.

Herschel (approved):

Herschel:
Thank you!

PR #10348, opened 2023-03-24
chore: Remove `static_assertions` dependency

relrelb:
`static_assertions` seems unmaintained, and anyway `assert!()` is usable in `const` contexts since Rust 1.57.0:
https://blog.rust-lang.org/2021/12/02/Rust-1.57.0.html#panic-in-const-contexts So simply use the suggested method instead.

Also the `rustversion` dependency is no longer needed because https://github.com/rust-lang/rust/pull/94075 already landed in stable.

### DIFF
--- a/core/src/avm2/error.rs
+++ b/core/src/avm2/error.rs
@@ -3,6 +3,7 @@ use crate::avm2::Activation;
 use crate::avm2::AvmString;
 use crate::avm2::Multiname;
 use crate::avm2::Value;
+use std::mem::size_of;
 
 use super::ClassObject;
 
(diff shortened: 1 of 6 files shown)


adrian17:
Can we keep the concrete values for the 2 platforms?

relrelb:
> Can we keep the concrete values for the 2 platforms?

Why? I thought it would be nicer to compute the sizes using current pointer size.

adrian17:
The formula happens to check out now, but it doesn't represent a consistent underlying rule. As in:
- on 32-bit, the size upper bound comes from `f64` alignment requirement (removing the Number variant would reduce the size to 12b)
- on 64-bit, the size upper bound comes from 2xpointer size plus 8b alignment (removing the `String` field would reduce the size to 16b).

So possible refactoring of the structs could throw the formula out of whack anyway.

And aside from being a "regression test", having specific values at hand makes it slightly easier to eyeball memory traffic, at least for me.

Herschel:
I agree, I prefer the explicit values in this case.

relrelb:
Done.

core/src/avm2/error.rs
```diff
@@ -34,17 +35,11 @@ impl<'gc> Error<'gc> {
...
 
 // This type is used very frequently, so make sure it doesn't unexpectedly grow.
-// For now, we only test on Nightly, since a new niche optimization was recently
-// added (https://github.com/rust-lang/rust/pull/94075) that shrinks the size
-// relative to stable.
+#[cfg(target_pointer_width = "32")]
+const _: () = assert!(size_of::<Result<Value<'_>, Error<'_>>>() == 24);
 
-#[rustversion::nightly]
-#[cfg(target_arch = "wasm32")]
```
adrian17, line 42: the reason we used wasm32 specifically was because wasm and x86 (specifically ` i686-unknown-linux-gnu`) can have different alignment requirements, and indeed in the past (when we added the assertions originally) the sizes for these two happened to be different because of that. So let's keep `wasm32` here.

adrian17 (approved):

Issue #2092, opened 2020-12-25
getURL has some issues depending on the backend


Toad06:
`getURL("https://ruffle.rs","_self","GET");`
Sample: [getURL.zip](https://github.com/ruffle-rs/ruffle/files/5742433/getURL.zip)

**Web**

- An extra character (`?`) is added to the destination URL.

- The panic screen briefly appears when this code is executed. This is a regression. I've run a bisect, the issue seems to have been introduced in commit 739af3f8d71eaf5981686823fa312715f23608a1. The following error is logged into the console:
`panicked at 'cannot recursively acquire mutex', /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/sys/wasm/../unsupported/mutex.rs:22:9`
(this issue doesn't happen if the target is `_blank`)

**Desktop (Windows)**

- Lots of characters are added to the destination URL: `?%24version=LNX+32%2C0%2C0%2C0&instance2=_level0.instance2&instance1=_level0.instance1`

- The official Flash Player projector seems to ignore this code and doesn't open the URL.

Herschel:
Maybe we should move the shared object flush from `beforeunload` to `unload` so that it doesn't run immediately when navigation occurs.

Toad06:
👍

I've tested that in Chrome and Firefox (Windows) and that did the trick: no more panic screen. This is also still good for #2046.

amrography:
I had the similar issue, my case `getURL("https://test.com?Correct=c","_self","GET");`. But after it only redirects me to `https://test.com?`

n0samu:
> The official Flash Player projector seems to ignore this code and doesn't open the URL.

Yes, the Flash projector tends to ignore these kinds of commands when playing a local file, for security reasons I assume. Let's look at the behavior when loading this SWF via a web server to get a more accurate test. I am running `python -m http.server` and running the SWF via this URL: `http://localhost:8000/getURL.swf`.

When I load the above URL into the Flash projector or [a Flash-enabled browser](https://archive.org/details/basilisk-portable-with-flash), it sends me to `https://ruffle.rs/`, as expected.

But when I load that into the Ruffle desktop app or web demo page, I get this URL instead:
`https://ruffle.rs/?%24version=LNX+32%2C0%2C0%2C0&instance2=_level0.instance2&instance1=_level0.instance1`

[...]

PR #10356, opened 2023-03-25
exclude __proto__ prop from object props enumeration

emgfc:


Herschel:
Could we add a test? Thank you!

n0samu:
Just a side note, I wonder if something similar would fix the issue where Ruffle sometimes adds extra query params to URLs opened via getURL (#2092)

Lord-McSweeney:
There are similar references in `globals/bitmap_filter.rs` and `object/script_object.rs`, but they already do this:
```
            object.define_value(
                gc_context,
                "__proto__",
                proto.into(),
                Attribute::DONT_ENUM | Attribute::DONT_DELETE,
            );
```
Should `__proto__` also have the `DONT_DELETE` attribute?

emgfc:
> Could we add a test for this? Thank you!
Spent almost all day trying to trigger this behavior with custom MovieClips in AS2, had no luck.

> Should __proto__ also have the DONT_DELETE attribute?
I'm pretty sure, yes.

Lord-McSweeney (approved):

Lord-McSweeney:
Seeing as the construction of the `__proto__` attribute in `script_object.rs` and `bitmap_filter.rs` don't have tests, I think this is fine without tests.

Herschel (approved):
Thank you!

Issue #9058, opened 2023-01-07
Scratch 2 doesn't run


Lord-McSweeney:
### Describe the bug

Scratch 2 (archived SWF at http://web.archive.org/web/20190714210604/https://cdn.scratch.mit.edu/scratchr2/static/__937f088a133dc228065bf12d2cbcd1d9__/ScratchFor10.2.swf) hits the script execution timeout on web.

### Expected behavior

Scratch 2 can run on web.

### Affected platform

Browser's extension

### Operating system

Linux

### Browser

Chrome 108

Lord-McSweeney:
This is now stuck on the AVM2 namespace constructor.

Lord-McSweeney:
Newest error is `Got AVM2 error RustError("Property does not exist: Multiname { ns: Multiple([Namespace(Namespace(\"flash.utils\")), Namespace(Namespace(\"grabcut.kernel\")), Namespace(Namespace(\"grabcut.vfs\")), Namespace(Namespace(\"grabcut\")), Namespace(Namespace(\"grabcut\")), Namespace(PackageInternal(\"grabcut\")), Namespace(Private(\"\")), Namespace(Namespace(\"\")), Namespace(Namespace(\"http://adobe.com/AS3/2006/builtin\"))]), name: Some(\"AlcDbgHelper\"), params: [], flags: MultinameFlags(0x0) }") when attempting to assign symbol class grabcut_locale_2D_inst_2E_o_3A_6758CAF9_2D_4E1F_2D_4AE8_2D_ADC8_2D_0E91C8BA3E91.DS28`, before which comes a "#1520 Mutex cannot be initialized" error.

Lord-McSweeney:
Now only requires `decodeURIComponent`.

An alternative version at http://web.archive.org/web/20190714210604/https://cdn.scratch.mit.edu/scratchr2/static/__937f088a133dc228065bf12d2cbcd1d9__/ScratchFor10.2.swf has a similar error (I believe also because of unimplemented `decodeURIComponent`).

Lord-McSweeney:
With #10360:
```
WARN ruffle_core::stub: Encountered stub: AVM2 flash.system.Security.loadPolicyFile()
WARN ruffle_core::stub: Encountered stub: AVM2 flash.text.Font.enumerateFonts()
WARN ruffle_core::stub: Encountered stub: AVM2 flash.display.ShaderData constructor
[...]

Issue #8116, opened 2022-09-28
Super Mech Not Supported AS3 with Ruffle Extension


deividbinev12:
### Describe the bug

https://user-images.githubusercontent.com/80902993/192844564-59ab3ec5-8436-4ddb-bc4d-76920aa3dc00.mp4

www.supermechs.com/:239 
        
      
        
        
      
        
      
       
        
       GET https://pagead2.googlesyndication.com/pagead/show_ads.js net::ERR_BLOCKED_BY_CLIENT
www.supermechs.com/:732 
        
      
        
        
      
        
      
       
        
       GET https://ads.lfstmedia.com/getad?site=298042 net::ERR_BLOCKED_BY_CLIENT
www.supermechs.com/:1 
        
       Tracking Prevention blocked access to storage for https://www.adobe.com/images/shared/download_buttons/get_flash_player.gif.
www.supermechs.com/:386 
        
      
        
        
      
        
      
       
        
       GET https://pagead2.googlesyndication.com/pagead/show_ads.js net::ERR_BLOCKED_BY_CLIENT
966.js:1 
        
       The AudioContext was not allowed to start. It must be resumed (or created) after a user gesture on the page.
(anonymous) @ 966.js:1
R @ 966.js:1
W.e.wbg.__wbg_new_e2e737399e087a85 @ 966.js:1
$<core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll::h58330bfd04d330d3 @ 017e887e:0x959fd
$wasm_bindgen_futures::queue::Queue::new::{{closure}}::h45d08e14315eeec0 @ 017e887e:0x3c4798
$<dyn core::ops::function::FnMut<(A,)>+Output = R as wasm_bindgen::closure::WasmClosure>::describe::invoke::hf95ac0264acb29b2 @ 017e887e:0x492031
$closure2920 externref shim @ 017e887e:0x49475f
S @ 966.js:1
c @ 966.js:1
Promise.then (async)
W.e.wbg.__wbg_then_18da6e5453572fc8 @ 966.js:1
$wasm_bindgen_futures::queue::Queue::schedule_task::hd172c6fcd7831cc1 @ 017e887e:0x42761e
$wasm_bindgen_futures::task::singlethread::Task::spawn::h4aad2114880c45f3 @ 017e887e:0x478613
$wasm_bindgen_futures::future_to_promise::{{closure}}::h4faf776db9cad08a @ 017e887e:0x44dca4
[...]

PR #10360, opened 2023-03-25
AVM2: Implement decodeURI and decodeURIComponent

AllinolCP:
progresses #8330


Lord-McSweeney:
Progresses #9058 into needing `flash.display.Shader`.
Progresses #8116 to the "connecting to server" screen (`NetConnection`).

core/src/avm2/globals/toplevel.rs
```diff
@@ -379,3 +379,90 @@ fn encode_utf8_with_exclusions<'gc>(
...
+
+        let mut bytes = Vec::new();
+        while let Some(b) = handle_percent(&mut chars) {
+            bytes.push(b);
+            if !matches!(chars.peek(), Some('%')) {
+                break;
+            }
+            chars.next();
+        }
+        let decoded = String::from_utf8_lossy(&bytes);
```
Herschel, line 447: I believe you can do `std::str::from_utf8` here and return the `URIError` immediately on failure.

Herschel (changes requested):
Doh, I had just done this myself as well! 😄 

The following give incorrect results:
```
trace(decodeURI("http%3A%2F%2F"));
trace(decodeURI("%u33"));
trace(decodeURI("%32%3"));
trace(decodeURI("%32%3"));
trace(decodeURI("%%3"));
```

core/src/avm2/globals/toplevel.rs
```diff
@@ -379,3 +379,95 @@ fn encode_utf8_with_exclusions<'gc>(
...
+    let mut bytes = Vec::new();
+
+    while let Some(c) = chars.next() {
+        let Ok(c) = c else {
+            return Err(Error::AvmError(uri_error(
+                activation,
+                &format!("Error #1052: Invalid URI passed to {func_name} function."),
+                1052,
+            )?));
+		};
```
Lord-McSweeney, line 430: fmt

tests/tests/swfs/avm2/decode_uri/Test.as
```diff
@@ -0,0 +1,56 @@
...
+
+	var input = "\x05";
+	trace("// " + fnName + "(\"\\x05\")");
+	trace(fn(input));
+	trace("");
+
+	var input = "😭";
+	trace("// " + fnName + "(\"" + input + "\")");
+	trace(fn(input));
+	trace("");
```
Lord-McSweeney, line 55: I would suggest adding one with input being `encodeURI("😭")`. I'm not sure if decoding that works in Flash, though.

core/src/avm2/globals/toplevel.rs
```diff
@@ -379,3 +379,90 @@ fn encode_utf8_with_exclusions<'gc>(
...
+    func_name: &str,
+) -> Result<Value<'gc>, Error<'gc>> {
+    let value = match args.first() {
+        None => return Ok("undefined".into()),
+        Some(Value::Undefined) => return Ok("null".into()),
+        Some(value) => value.coerce_to_string(activation)?,
+    };
+
+    let bs = value.as_wstr();
+    let mut output = String::new();
```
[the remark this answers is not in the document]
AllinolCP: Is there any point of using a `WString` instead of a `String`? based on the `encode_utf8_with_exclusions` it also uses a `String`

tests/tests/swfs/avm2/decode_uri/Test.as
AllinolCP: It supports it in flash but not in my implementation, as i have not supported 4 bytes utf-8 yet

core/src/avm2/globals/toplevel.rs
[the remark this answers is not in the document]
Herschel: `AvmString::new_utf8` involves converting from UTF-8 `String` to UCS-2 `WString`. That at least requires scanning the string to determine it's ASCII at best, or widening it to 16-bit code units at worst.

Instead you can build the `WString` directly in place. Very minor nit for a function where performance isn't important here at all, though!

AllinolCP:
Drafting this as I missed a thing in my implementation

core/src/avm2/globals/toplevel.rs
AllinolCP: I have `cargo fmt --all`'d this however its still showing this for no reason

tests/tests/swfs/avm2/decode_uri/Test.as
AllinolCP: Fixed now

core/src/avm2/globals/toplevel.rs
AllinolCP: I've tricked `rustfmt` to format it by making it a `if let` temporarily and that fixed it

Lord-McSweeney:
Here's another test case I could suggest: "~!%40%23%24%25%5E%26*()_%2B%5B%5D%5C%7B%7D%7C%3B'%2C.%2F%3C%3E%3F"

AllinolCP:
> Here's another test case I could suggest: "~!%40%23%24%25%5E%26*()_%2B%5B%5D%5C%7B%7D%7C%3B'%2C.%2F%3C%3E%3F"

Thanks! I added it

Lord-McSweeney:
I would also suggest adding some of the test cases that Herschel had mentioned:
>```trace(decodeURI("http%3A%2F%2F"));```
>```trace(decodeURI("%u33"));```
>```trace(decodeURI("%32%3"));```
>```trace(decodeURI("%32%3"));```
>```trace(decodeURI("%%3"));```
Other than that LGTM.

AllinolCP:
> I would also suggest adding some of the test cases that Herschel had mentioned:

```trace(decodeURI("http%3A%2F%2F"));```
This is tested using the latest added one
as for the others, it just errors out so it does not really matter as we already have a test for that

**Review by Lord-McSweeney — APPROVED** [review]

File: core/src/avm2/globals/toplevel.rs [hunk 5]
```diff
@@ -379,3 +379,147 @@ fn encode_utf8_with_exclusions<'gc>(
...
+        .and_then(|c| c.to_digit(16));
+    let low = chars
+        .next()
+        .and_then(|c| c.ok())
+        .and_then(|c| c.to_digit(16));
+
+    if high.is_none() || low.is_none() {
+        return None;
+    }
+    Some(low.unwrap() as u8 | ((high.unwrap() as u8) << 4))
```
Herschel, line 421:
```
    let high = chars.next()?.ok()?.to_digit(16)?;
    let low = chars.next()?.ok()?.to_digit(16)?;
    Some(low as u8 | ((high as u8) << 4))
```

Herschel (changes requested):
```
trace(decodeURI("%FFabcd")); // Should error, hits a overflow assert in this PR
```

core/src/avm2/globals/toplevel.rs
```diff
@@ -379,3 +379,147 @@ fn encode_utf8_with_exclusions<'gc>(
...
+                &format!("Error #1052: Invalid URI passed to {func_name} function."),
+                1052,
+            )?));
+        };
+        bytes.push(byte);
+        if (byte & 0x80) != 0 {
+            let mut n = 1;
+            while ((byte << n) & 0x80) != 0 {
+                n += 1;
+            }
```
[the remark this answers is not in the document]
AllinolCP: resolved in the latest commit!

Lord-McSweeney:
Some of these should probably be added as tests:
`"%FFabcd"`
Also, I feel like it would be better to add the original failing examples that Herschel commented as tests to make sure they don't, for example, panic or give incorrect results.

AllinolCP:
> Some of these should probably be added as tests:
> `"%FFabcd"`
I've added this to the tests

> Also, I feel like it would be better to add the original failing examples that Herschel commented as tests to make sure they don't, for example, panic or give incorrect results.

just in case i guess I'll add it

Herschel (approved):
Thank you!

PR #10362, opened 2023-03-25
fix ChildContainer.replace_at_depth() panic

emgfc:
Right now this method panics if a previous child is not in a render list, which causes https://github.com/ruffle-rs/ruffle/issues/10314. I'm totally unsure about this one, because I don't yet understand what's going on here, but at least that Sponge Bobbity thing works now.

Anyway, it would be nice if someone could explain me why there's a references in depth array to MovieClips that are not in render list.

### DIFF
--- a/core/src/display_object/container.rs
+++ b/core/src/display_object/container.rs
@@ -650,6 +650,7 @@ impl<'gc> ChildContainer<'gc> {
                     None
                 }
             } else {
+                tracing::error!("ChildContainer::replace_at_depth: Previous child is not in render list");
                 self.push_id(child);
                 None
             }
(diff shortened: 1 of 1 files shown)


adrian17:
> Anyway, it would be nice if someone could explain me why there's a references in depth array to MovieClips that are not in render list.

This has been a long-standing issue and I also could never figure it out :( But yeah, this might be the real underlying bug.

Herschel (commented):
Yep, anything on the depth list should be on the render list, so that's the real bug. The render list should be the canonical list of what the children are; depth list is only used to speed up operations that require searching for a depth.

That said, I think it's worthwhile to avoid the panic here for now.

Herschel (approved):
Thank you!

PR #10364, opened 2023-03-25
extension: Check tools using TypeScript

relrelb:
This would have caught #10351.


Herschel:
LGTM, although I'm dumb when it comes to this area and don't know all of what this entails!

Herschel:
@danielhjacobs Do you have any thoughts here? Looks good to me but want to check with people that work a lot with the web build.

danielhjacobs:
Yeah, from my perspective this can only be a good thing. From my understanding,  it just checks for issues on some files that weren't previously being checked, which is always a good idea.

Herschel:
👍 Feel free to rebase and merge @relrelb

PR #10367, opened 2023-03-25
avm2: Log error message instead of error object

Lord-McSweeney:
This makes the error message visible if the `Error` is an `AvmError` (e.g. `VerifyError`, `ReferenceError`).


Herschel (approved):

Herschel:
Thank you!

PR #10369, opened 2023-03-25
avm2: Implement XML set for simple element cases

evilpie:


Herschel (approved):
Thank you!

core/src/avm2/object/xml_object.rs
```diff
@@ -251,12 +255,49 @@ impl<'gc> TObject<'gc> for XmlObject<'gc> {
...
+        }
+
+        if name.is_any_name() {
+            return Err("Any name (*) not yet implemented for set".into());
         }
 
-        Err(format!("Modifying an XML object is not yet implemented: {name:?} = {value:?}").into())
+        let text = value.coerce_to_string(activation)?;
+
+        let matches: Vec<_> = children
```
Herschel, line 283: I realize this is temp first-pass code, but would be good to avoid allocating here and use the iterator directly when we implement replacing multiple elems.

PR #10373, opened 2023-03-25
render: Fix BitmapData.applyFilter with non-full sourceRect

Aaron1011:
Previously, we were scaling down the source image to fit into the smaller sourceRect, instead of cropping at the original scale. This broke the background textures in Fancy Pants World 4 Part 2, as the scaled-down output image resulted in a smaller rectangle being returned from 'getColorBoundsRect'

We now crop the image by properly constructing the UV-coordinate transformation matrix. We were also using the wrong value for the 'destPoint' y coordinate, which I fixed.

This slightly changes the image output of two tests - the new images now more closely match the Flash output.


Herschel (approved):

n0samu:
This caused a regression: #10380

PR #10378, opened 2023-03-26
avm2: Don't overwrite symbol class entry with a different SwfMovie

Aaron1011:
If you use a `Loader` to load an SWF containing a class that shadows an already-defined class, the class definition from the Loader SWF will be ignoredin favor of the already-defined class. This commit applies this log to symbol classes as well - the symbol registry for the class should continue to point to the existing MovieClip in the parent. This results in the child SWF instantiating the class from the parent SWF when the child places the affected movie clip on the timeline.

This fixes a bug in Fancy Pants World 4 Part 3, where the sub-level SWF was replacing the symbol class entry for the parent 'shipInteract' class with the dummy clip provided in the sub-level SWF (instead of continuing to use the correct clip from the parent SWF).


core/src/library.rs
```diff
@@ -90,6 +90,21 @@ impl<'gc> Avm2ClassRegistry<'gc> {
         movie: Arc<SwfMovie>,
         symbol: CharacterId,
     ) {
+        if let Some(old) = self.class_map.get(&class_object) {
+            // We should never have multiple symbols in the same movie pointing the same class
+            if Arc::ptr_eq(&movie, &old.0) {
+                panic!(
```
Lord-McSweeney, line 96: If this is theoretically possible through a malicious SWF, wouldn't a warning or error be better suited here?

Herschel:
This shouldn't apply when loading into a completely separate ApplicationDomain, i.e. `new ApplicationDomain(null)`, right? 

Sample:
[as3load.zip](https://github.com/ruffle-rs/ruffle/files/11070733/as3load.zip)

Aaron1011:
@Herschel If the domain is completely separate, then we'll instantiate a different `ClassObject`, which will gets its own symbol class entry.

Herschel:
> @Herschel If the domain is completely separate, then we'll instantiate a different `ClassObject`, which will gets its own symbol class entry.

But the sample I posted does not run correctly currently -- I'm not sure if this PR is meant to fix that, just clarifying for my own edification because the PR text was unclear whether it should.

SN902:
From this change, now few games panics with same error:

`thread 'main' panicked at 'Tried to overwrite class ClassObject { name: GGRPimgloading_imageClass, ptr: 0x2515ed94c50 } with symbol from same movie', core\src\library.rs:96:17`

[fluttershy demo camera.zip](https://github.com/ruffle-rs/ruffle/files/11070992/fluttershy.demo.camera.zip)

SN902:
Both swfs fixed

core/src/library.rs
```diff
@@ -90,6 +90,25 @@ impl<'gc> Avm2ClassRegistry<'gc> {
...
+            // We should never have multiple distinct symbols in the same movie linked
+            // with the same class.
+            // However, flash player *does* allow duplicate SymbolClass entires in the same
+            // movie - the same (symbol, class) pair can show up across multiple SymbolClass
+            // tags.
+            if Arc::ptr_eq(&movie, &old.0) && symbol != old.1 {
+                panic!(
+                    "Tried to overwrite class {:?} id={:?} with symbol id={:?} from same movie",
+                    class_object, old.1, symbol,
+                );
```
Herschel, line 103: We shouldn't `panic` here because a weirdo/malicious SWF file can hit this. Sample:
[symbolClassConflict.zip](https://github.com/ruffle-rs/ruffle/files/11073224/symbolClassConflict.zip)

This SWF has 2 symbols that register to the same class. This actually works in the FP, and stage instances of either symbol will instantiate the class(!). Instantiating via AS `new Child()` creates the first registered child. So probably this should gracefully bail out if the class->symbol map already has a registered entry.

Aaron1011: I've changed this to log a warning, and added your SWF as a test case.

Herschel (approved):
Thank you!

PR #10385, opened 2023-03-26
avm2: Allow more namespaces in XML

evilpie:
I am having a bit of hard time reconciling the spec for ToXMLName and our current implementation. I am not sure if having a Multiname as an argument is actually correct.

However what we can see is that we should probably omit the namespace every time it's actually empty because Multiname.as_uri also doesn't prefix the uri with a namespace in that case.


Herschel:
So I understand because I'm dumb at E4X, this will fix something like `xml.*::bar`, right? Would you mind adding a test case?

evilpie:
Good question, I didn't add a test for this because I am actually not sure what the syntax looks like.

The set_property_local call I was looking at came from `setproperty MultinameLA([PackageInternalNs("")])`, where I also don't know what that looks like in AS3.

core/src/avm2/object/xml_object.rs
```diff
@@ -226,7 +226,7 @@ impl<'gc> TObject<'gc> for XmlObject<'gc> {
         value: Value<'gc>,
         activation: &mut Activation<'_, 'gc>,
     ) -> Result<(), Error<'gc>> {
-        if !name.contains_public_namespace() {
+        if name.has_explicit_namespace() {
             return Err("Can not set non-public name yet".into());
```
Lord-McSweeney, line 230: Is this message still the correct unimplemented error message?

Herschel:
> Good question, I didn't add a test for this because I am actually not sure what the syntax looks like.

Yep, did a quick test myself and looks like this PR fixes code like this:
```
var xml:XML = <ns:foo xmlns:ns="*">
	<ns:bar>test</ns:bar>
	<bar>test2</bar>
</ns:foo>;

trace(xml.*::bar.toString());
```

evilpie:
Most games that I've tested don't actually use explicit namespaces (the :: syntax), but instead just have multinames without a public namespace.

evilpie:
This is probably not perfect, but it hopefully helps more than it hurts.

Aaron1011 (approved):

PR #10392, opened 2023-03-26
avm2: Implement ApplicationDomain constructor and fix parent handling

Aaron1011:
Previously, the `ApplicationDomain` constructor ignored its argument, instead of constructing a new domain with the specified domain as the parent.

Additionally, we were incorrectly executing code with `Activation::from_nothing` in several places, causing `ApplicationDomain.currentDomain` to return the system domain instead of the correct parent domain. I've introduced a new method `Activation::from_domain`, which allows explicitly passing in the domain. Internally, we now store an `Option<Domain>`, and panic when calling `caller_domain` with a `None` domain. Several places in the codebase have been adjusted to pass in the correct domain.


Herschel (approved):
Thank you!

PR #10393, opened 2023-03-26
avm2: Fire soundComplete events immediately

Aaron1011:
Instead of queueing up these events in the `Activation`, we can fire them immediately by making `AudioManager::update_sounds` a freestanding method that takes in an `UpdateContext`


Herschel (approved):
Thank you!

PR #10394, opened 2023-03-26
wgpu: Discard transparent bitmap pixels to fix using a Bitmap mask

Aaron1011:
When we draw a mask, we increment the stencil buffer for each pixel we draw. Flash supports using a Bitmap as a mask, and only considers pixels with alpha!=0 to be part of the mask.

This can be implemented by using 'discard' in our 'bitmap.wgsl' shader for pixels with 'alpha=0', which prevents the stencil buffer from getting updated.

This fixes using unlockable pants textures in Fancy Pants World 4 Part 3, which previously rendered as a large rectangle over the player.


Herschel:
I believe this doesn't happen by default and actually depends on various other factors (`cacheAsBitmap` being true for both masker and maskee, filters being added which also force bitmapcaching, possibly other scenarios involving BlendMode...)

Herschel:
Sample:
[bitmapMask.zip](https://github.com/ruffle-rs/ruffle/files/11073162/bitmapMask.zip)

Aaron1011:
@Herschel: I have a comment in the test about this. I think this moves us closer to Flash's behavior, though.

Herschel:
I'm not sure -- as the default behavior is for bitmaps not to mask based on alpha, I worry this will break existing content that uses bitmaps in masks in shapes etc., which I think is somewhat common in earlier content.

Aaron1011:
I've updated the PR to only enable this behavior when `cacheAsBitmap=true` is set on both the masker and maskee.

kjarosh:
Superseded by https://github.com/ruffle-rs/ruffle/pull/21497

Issue #10384, opened 2023-03-26
Mortlake Manison - No Menu after Loading


desertking:
### Describe the bug

https://www.newgrounds.com/portal/view/566028?emulate=flash - Hey, Playrix has a great audience for their games.. This is one of the games which is stuck showing the menu and throwing errors.

### Expected behavior

Show the menu after loading

### Affected platform

Desktop app

### Operating system

All

### Browser

All

evilpie:
>ERROR core/src/avm2/events.rs:419 Error dispatching event EventObject(EventObject { type: "addedToStage", class: flash.events::Event, ptr: 0x3ba1020 }) to handler FunctionObject(FunctionObject { ptr: 0x3ba0d08 }) : ReferenceError: Error #1069: Property pixelBounds not found on flash.geom.Transform and there is no default value. ruffle_web-wasm_extensions.js:677:20

Lord-McSweeney:
This seems to be fixed now, @desertking can you confirm?

Lord-McSweeney:
Closing as I can no longer reproduce this issue. Please feel free to reopen it if you can do so.

desertking:
I can confirm. I enabled it a while ago as it was loading

PR #10396, opened 2023-03-26
avm2: Implement pixelBounds for transform

yoganlava:
Progresses: #10384
Code taken from avm1 transform


core/src/avm2/globals/flash/geom/transform.rs
```diff
@@ -249,3 +249,29 @@ pub fn object_to_matrix<'gc>(
...
+    activation: &mut Activation<'_, 'gc>,
+    this: Option<Object<'gc>>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let this = this.unwrap();
+    let display_object = get_display_object(this, activation)?;
+    rectangle_to_object(display_object.world_bounds(), activation)
+}
+
+pub fn rectangle_to_object<'gc>(
```
Lord-McSweeney, line 263: Is there a reason for this function being public?

yoganlava: Nope, copied from similar function up the file and forgot to remove, thanks

Herschel (approved):
I was going to suggest adding a test, but looks like we're missing `tests/avm2/displayobject_transform/Test.as` in the repo... 😆 In any case, the behavior is already covered by the AVM1 test and we can always add it if we'd like later.

yoganlava:
Hmm seems to be failing for a test after tracing pixelBounds, put off the merge until I get to resolve it

Herschel:
Good thing we added the test 😄  Thank you!

yoganlava:
@Herschel Sorry haha, haven't resolved the issue yet. pixelBounds seems to be giving some wacky numbers but not sure where the issue lies

Herschel:
> @Herschel Sorry haha, haven't resolved the issue yet. pixelBounds seems to be giving some wacky numbers but not sure where the issue lies

Oops, thx! Haven't looked into it yet, but how far off is it? Is it a precision issue? It's possible that the issue is in core `pixel_bounds` as I think it is somewhat naively implemented, in which case I'm okay with deferring that to a future PR.

yoganlava:
I think it's due to stage bounds reporting the incorrect values.
e.g.
Flash
```
// stage.getBounds()
(x=0, y=0, w=0, h=0)
```
Ruffle:
```
// stage.getBounds()
(x=6710886.35, y=6710886.35, w=0, h=0)
```
I think the code in this PR is fine, as long as the bounds are fixed

yoganlava:
I think i found the reason, the value reported by ruffle `6710886.35` * 20 = 134217727 = 0x7ffffff. The default value for a Rectangle<Twips> is 0x7ffffff or Twips::INVALID. I'm guessing flash sets the Rectangle bounds as 0 while we set them as 0x7ffffff

Lord-McSweeney:
For the test I think you might need to add a `test.toml` and `output.txt` to actually run the test.

yoganlava:
> For the test I think you might need to add a `test.toml` and `output.txt` to actually run the test.

Don't worry, I was just editing the existing test which still has the output and test.toml

Herschel (approved):
Thanks for the patience!

PR #10397, opened 2023-03-26
wgpu: Implement MSAA support for Stage3D

Aaron1011:
When we receieve a nonzero 'antiAlias' parameter, we create create a non-multisampled resolve buffer to use with WGPU.

Several tests were already requesting antialiasing, so their output images are now anti-aliased without any changes to the tests themselves.


Herschel (approved):
Cool, thank you!

Aaron1011:
Looks like mesa doesn't support multisampled textures :/

PR #10399, opened 2023-03-27
web: Give proper precedence to extension config when it is used

danielhjacobs:
This should work when the extension config loads after the self-hosted config too:
```ts
            (window.RufflePlayer.conflict["newestName"] === "extension" ||
                (window.RufflePlayer as Record<string, unknown>)[
                    "newestName"
                ] === "extension")
```


Herschel (approved):
Thank you!

PR #10402, opened 2023-03-27
avm2: Convert Font class to ActionScript

Aaron1011:


core/src/avm2/globals/flash/text/Font.as
```diff
@@ -0,0 +1,13 @@
+package flash.text {
+    public class Font {
+        public function Font() {}
```
Lord-McSweeney, line 3: I don't think this constructor is necessary.

core/src/avm2/globals/Class.as
evilpie, line 1: This seems unrelated.

Aaron1011: This is needed by `Font.as`

Lord-McSweeney (approved):

Herschel (approved):
Thank you!

Issue #9997, opened 2023-03-11
Ruffle breaks XML formatting on Chrome


Lord-McSweeney:
### Describe the bug

If you go to https://www.w3schools.com/xml/simple.xml with the extension enabled the text will be bunched up together, without any formatting: 
![www w3schools com_xml_simple xml](https://user-images.githubusercontent.com/84632019/224509609-eee92ac6-eec0-4a8e-972d-5bef2d9c580b.png)

### Expected behavior

![www w3schools com_xml_simple xml (1)](https://user-images.githubusercontent.com/84632019/224509641-3ce2aa73-96f9-4e50-9872-bc31a4103b55.png)

### Affected platform

Browser's extension

### Operating system

(irrelevant)

### Browser

Chrome

### Additional information

I fixed the problem by inserting
```
if (!(document instanceof HTMLDocument)) {
    return;
}
```
at the start of the concatenated module `load-ruffle.js` in both the `ruffle.js` and the `content.js` final builds, but I don't think TypeScript would like this very much.

n0samu:
The strange thing is that the extension indicates that it is not even loading on the page:
![image](https://user-images.githubusercontent.com/71368227/224542151-e1891956-38e4-4c9c-831a-22e00d4dfff2.png)
And the problem occurs regardless of whether "Play Flash content in Ruffle" is enabled.

PR #5364, opened 2021-09-25
extension: Add log level option

relrelb:

PR #4103, opened 2021-04-24
extension: Avoid script injection in XML documents

relrelb:
#4092 was not proper because `document.xmlVersion` doesn't exist in Firefox.
So use another method to detect XML documents as described in [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/xmlVersion).

Really fixes #4090.

PR #10415, opened 2023-03-27
web: Fix XML document detection

Lord-McSweeney:
The code in `shadow-template.ts` would error if the document was an XML document, which means the code for detecting XML documents at the end of `ruffle-player.ts` would not run.

Fixes https://github.com/ruffle-rs/ruffle/issues/9997.

### DIFF
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -209,7 +209,7 @@ export class RufflePlayer extends HTMLElement {
      */
     constructor() {
         super();
-        
+
         const ruffleShadowTemplate = document.createElement("template");
         ruffleShadowTemplate.innerHTML = ruffleShadowTemplateHTML;
 
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -1,6 +1,6 @@
 import type { Ruffle } from "../dist/ruffle_web";
 import { loadRuffle } from "./load-ruffle";
-import { ruffleShadowTemplate } from "./shadow-template";
+import { ruffleShadowTemplateHTML } from "./shadow-template";
 import { lookupElement } from "./register-element";
 import { DEFAULT_CONFIG } from "./config";
 import {
@@ -209,6 +209,9 @@ export class RufflePlayer extends HTMLElement {
      */
     constructor() {
         super();
+        
+        const ruffleShadowTemplate = document.createElement("template");
+        ruffleShadowTemplate.innerHTML = ruffleShadowTemplateHTML;
 
         this.shadow = this.attachShadow({ mode: "open" });
         this.shadow.appendChild(ruffleShadowTemplate.content.cloneNode(true));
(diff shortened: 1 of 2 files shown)


relrelb:
I wonder how loading XML documents worked after #4103. Something regressed since?

n0samu:
I had the exact same thought. Here is what @Lord-McSweeney explained on Discord when I asked him about it:
> it's not the injected script, it's the content script
modifying any node even if it's not attached to the document causes the error
because `document.createElement` functions differently in XML vs HTML

As I understand, the `document` object in extension content scripts is actually somehow shared with the current page, so we have to avoid any operations on it when the current page is an XML document.

relrelb:
Okay, I think that `shadow-template.ts` is accidentally bundled into `content.js` due to an import of `ruffle-core`:
https://github.com/ruffle-rs/ruffle/blob/fc4b51b20f3b59aa38d65f4be15fedd93242249f/web/packages/extension/src/utils.ts#L2
which was introduced by me in #5364...

Ideally this `import "ruffle-core"` could be avoided completely, but sadly I don't see a nice way to import a `const enum` member without including its whole containing module. https://github.com/microsoft/TypeScript/issues/40344 seems to have some related discussions, but still without a clear conclusion/easy workaround.

I can think of 3 possible solutions:
1. As a really quick and direct fix, I think we can replace `LogLevel.Error` with `"error" as LogLevel`, requiring us to only `import type "ruffle-core"` without actually including any code. This will probably require some special annotation for TypeScript to accept that.
2. Make the `logLevel` field in `common.ts` optional, so `DEFAULT_OPTIONS` in `utils.ts` wouldn't need to specify it. Then, `DEFAULT_CONFIG` in `config.ts` will default to `LogLevel.Error` if it's absent.
3. Maybe the long-term approach is to split `BaseLoadOptions` etc. out to a separate package. Then importing it wouldn't pose a problem.

I'd like to avoid the third solution for now, as it's pretty heavyweight. The first two seem both good to me, but I tend to the second.

What do you think? By the way I didn't dig into this with a debugger, so I may have missed some details.

Lord-McSweeney:
I think they're bundled because of this code:
```
import { ruffleShadowTemplate } from "./shadow-template";
```
in `ruffle-player.ts`, which is included in `content.js`.

n0samu:
@relrelb I just tested option 1 and it worked perfectly without any extra annotations, thanks! I think we should go with that one.
```diff
--- a/web/packages/extension/src/utils.ts
+++ b/web/packages/extension/src/utils.ts
@@ -1,11 +1,11 @@
 import type { Options } from "./common";
-import { LogLevel } from "ruffle-core";
+import type { LogLevel } from "ruffle-core";

 const DEFAULT_OPTIONS: Options = {
     ruffleEnable: true,
     ignoreOptout: false,
     warnOnUnsupportedContent: true,
-    logLevel: LogLevel.Error,
+    logLevel: "error" as LogLevel,
     showSwfDownload: false,
 };
```
@Lord-McSweeney Would you like to amend your PR?

Lord-McSweeney:
Closing as @n0samu suggested a better fix.

PR #10426, opened 2023-03-28
desktop: Add GUI using egui

Herschel:
This is a rough WIP for discussion and collaboration. Code could use some cleanup as I'm still learning my way around egui.

image

 * Ruffle launches into an empty movie instead of showing the file picker dialog.
 * Main menu bar can be used to play a movie.
 * File menu
   * Open File
   * Open URL
   * Exit
 * Help menu
   * Discord
   * About
 * Right-click context menu

TODO:
 * ~Currently uses `fetch_root_movie` to change movies, but it'd be safer to recreate the entire `Player` when loading a new movie for now. Have to add a method to reset and re-use some of the backends.~
 * ~Pause and darken movie when UI is visible.~ no need anymore
 * Embed Ruffle logo animation and use it as default movie on startup.
 * ~Decide on rules for showing the menu bar.~
   * ~Currently it will hide whenever a movie is playing, and holding Esc for half a second will display it.~
   * ~This interacts badly with full-screen mode.~
   * ~Other possibilities:~
     * ~Always show~
     * ~Show when mouse is at top of screen~
 * In general, how should we handle keyboard shortcuts, because a game could also use the same keys?
 * Change egui styles to match Ruffle color scheme.
 * Settings menu
 * Improve About screen
   * Ruffle logo
   * Open source licenses/acknowledgements
 * ~Move file picker code out of `main` to `gui` (still using `rfd`)~
 * Move full-screen shortcuts into `gui`
 * ~Update navigator base URL when a new movie is loaded.~
 * Prevent GUI mouse event from interacting with the movie.
 * ~egui rendering is done directly to the final wgpu surface so no AA and possible color space issues.~
 * General polish of the layout.

**Other UI crate possibilities**

I'm not settled on egui as the best solution, this is experimental to see how the crate works. Open to opinions. Other possibilites:

 * [iced](https://github.com/iced-rs/iced) This is the most likely alternative. egui seemed a little more active, and at the time I started this work, iced didn't have a built-in menu bar widget. Looks like it does now in `iced_aw`.
 * [imgui-rs](https://github.com/imgui-rs/imgui-rs) Industry standard, very mature, but some awkwardness from C++<->Rust interop
 * [gtk](https://github.com/gtk-rs/gtk3-rs) 
 * [xilem](https://github.com/linebender/xilem) Formerly druid, in flux but maybe something to keep an eye on in the future


SN902:
My rough WIP, I think, all stuff must be in the center of screen, it will be useful in full-screen mode and on high res monitors 

image

waspennator:
Could touch input also be looked into for the revamp if possible? According to Nosamu it doesbt work on the original desktop version, but did on the flash player.

n0samu:
> Could touch input also be looked into for the revamp if possible? According to Nosamu it doesnt work on the original desktop version of ruffle, but did on the flash player.

I think that's quite unrelated to this PR - you should make a separate feature request for it.

Herschel:
Thanks for the feedback!

> * In **fullscreen** mode, the Flash projector hides the menu bar. I think we should do the same. I don't think we need a way to show the menu in fullscreen mode.

I like the idea of mimicking the standalone player, and it does keep things simple! The only thought I had is that it might be useful to be able to adjust settings without exiting fullscreen mode. For example, if we have a full settings menu with things like key rebinding, etc., it might be slightly annoying to have to exit to windowed mode to fiddle with this. But possibly this is thinking ahead too much!

I also have to properly resize the window and bump the player itself down to account for the menu size if the menu is always going to be visible.

> In general, how should we handle keyboard shortcuts, because a game could also use the same keys?

To answer my own question, I was just playing through Typing of the Dead via RetroArch recently, and I like how they handle it. Keyboard shortcuts can be captured by the UI normally, which is totally fine for most content. But you can press Scroll Lock to lock all keyboard input to the the game.

Dinnerbone:
> egui rendering is done directly to the final wgpu surface so no AA and possible color space issues.

Instead of rendering ui onto the wgpu ruffle backend, render ruffle onto a texture and project that onto the ui. This way we can have UI without a ruffle player loaded, and separation of MSAA

Herschel:
> Instead of rendering ui onto the wgpu ruffle backend, render ruffle onto a texture and project that onto the ui. This way we can have UI without a ruffle player loaded, and separation of MSAA

I agree! This was just the easiest way to get started.

Dinnerbone:
I've rebased this onto master (with wgpu 0.16, so I've had to change the egui version to a commit ref until they release a new one).

I can take a look at the TODOs too.

Dinnerbone:
@n0samu 

> The About Ruffle and Open URL dialogs have a resizable border when they probably shouldn't.

Fixed!

> Clicking the Open File button doesn't dismiss the submenu until after a file is selected, which isn't consistent with how the other menu buttons work.

Not easy to fix, it's because the window is "paused" (OS level) when you click it, so we aren't re-rendering and you're seeing the state of the screen before the dialog was opened.

> Yeah, right now the background music from one game persists after loading the next game, so this is definitely needed.

Should be fixed.

For keeping the menu bar open all the time and not overlaying the movie, it's not trivial right now but I can look into it. It'd certainly make the question of how to open it much simpler.

crumblingstatue:
This is really awesome.
Due to the immediate mode nature of egui, it should be fairly easy to walk the data structures of ruffle and expose a powerful debugging ui for the ruffle developers, and also enthusiasts to play with.

Dinnerbone:
Opening this up for review now. It's not great, there's a lot I'd like to do in further PRs, but it should now be a strict improvement over the current desktop!

n0samu:
Just tested this out and it's working great, thanks so much! From a UX standpoint I think it's definitely ready to merge.

Dinnerbone:
I'll take a look shortly, thank you!

Dinnerbone:
@thisisjonathan can you see if this commit fixes it? It's the same thing you tried but a correct width/height. I'm curious if that'll be enough or if that second crash is another issue entirely

thisisjonathan:
Starting now works correctly, but resizing still gives the same error as before. Going fullscreen with Alt-Enter does work however.

Dinnerbone:
@thisisjonathan and this one? Thanks for testing!

thisisjonathan:
I still get the same error.

Dinnerbone:
Crash on X11 should be fixed now, it was that winit's `window.get_inner_size()` can change at any point during execution. I've changed it to no longer rely on that.

desktop/assets/texts/en-US/main_menu.ftl
```diff
@@ -0,0 +1,21 @@
+# A note on ellipsis: they indicate when an action won't immediately happen, but needs more information from the user.
+# For example, "Open File..." won't open a file, it'll ask the user to choose a file and *then* open the file.
+# Whereas "Close" or "About" wouldn't need any information, they'd immediately happen as soon as you click it.
+# In general, if clicking the thing immediately does that thing, no ellipsis. If it asks for more information, ellipsis.
+# This is the case for English, but other languages may have other rules or ways of showing this.
+
+# Some items may be also be duplicated here, like "About Ruffle" in the menu and "About Ruffle" elsewhere as the header of the about menu.
+# This is because some languages may use different wording for one or the other, especially with things like capitalisation.
```
torokati44, line 8: nit: Some context for which is which would be nice then. But this is definitely not a blocker, just a note.

Dinnerbone: Is the file it's in and key name not enough?

`main_menu.ftl` `help-menu-about` is for the main menu item
`about_dialog.ftl` `about-ruffle` is for the about dialog
They both say "About Ruffle".

torokati44: It's probably fine, now that I think of it, translators can even request additional context on Crowdin itself. No need to get ahead of imagined problems needlessly...

Dinnerbone: Yeah, we can also put screenshots on the items in crowdin for super duper clarity. The comment here was mostly to future developers going "why not just a single text if they're both the same", because they might not be outside of English

desktop/src/app.rs
```diff
@@ -0,0 +1,488 @@
...
+use winit::event::{ElementState, KeyboardInput, ModifiersState, VirtualKeyCode, WindowEvent};
+use winit::event_loop::{ControlFlow, EventLoop, EventLoopBuilder};
+use winit::window::{Fullscreen, Icon, Window, WindowBuilder};
+
+pub struct App {
+    opt: Opt,
+    window: Rc<Window>,
+    event_loop: Option<EventLoop<RuffleEvent>>,
+    gui: Arc<Mutex<GuiController>>,
+    player: PlayerController,
```
torokati44, line 26: nit: _Slightly_ confusing name. (Same below where this is constructed.) No big deal though...

torokati44:
I have skimmed through the code and didn't see anything that looked really wrong with it. I like how the interface itself looks, it works intuitively, and it already covers the basics really well for a first iteration. Props for the verbose "About" menu/dialog! :)

One minor issue is the missing HIDPI scaling of the menu bar offset, but I wouldn't consider that a hard blocker - even though it's probably trivial to fix.

I guess the commit history is not as nice as it could be, given for how long it has been in the workings. Entirely squashing it wouldn't be ideal IMHO, and cleaning it up properly would probably be a major effort that's not worth the benefits. So it's probably fine the way it is.

The refactoring with moving the backends out into a separate folder, and splitting the massive `main.rs` into more easily digestible pieces is also nice to see. Let's hope it won't cause any major trouble when rebasing any preceding branches onto it.

A little explanation in some form would be nice about what `Player`, `PlayerController`, and `ActivePlayer` are, how are they different, what they do, why each is needed, and what's their relationship.
Perhaps the same thing with the new `MovieView` and `MovieViewRenderer`, and how it's different to the existing `RenderTarget` impls, and how and where the colorspace conversion is done.

So, given that this is something of a big-ticket item, which has been wanted for a while, and improves the UX of the desktop app significantly, I personally don't see any reason to hold it back any longer (modulo my slight concerns above).

torokati44 (approved):

Toad06:
Thank you for the quick fixes and the explanations. 😉 I also confirm this works on my end.

adrian17 (approved):

PR #10430, opened 2023-03-28
avm1: Support strings in `MovieClip.setMask(str)`

### LINKED ISSUE #9493: Strange graphic error in bubble tanks 2

Toad06:
Fixes #9493. Also found out that this method had a return value (the documentation states otherwise), so I implemented this part as well.


Toad06:
I added a visual test but it fails. Is there a specific way the snapshot should be made ("expected.png") or anyone has a clue on what is wrong here? For some reason, all visual tests pass on my machine even if the "expected.png" file contains unrelated data.

Toad06:
Thank you, I didn't know about that. Much appreciated!

Herschel:
Looks like we need to not unset the mask when doing `foo.setMask()`. Example:
[mask.zip](https://github.com/ruffle-rs/ruffle/files/11108535/mask.zip)

Toad06:
Good catch, thank you! It appeared that any value that couldn't be turned into a display object (other than `undefined` or `null`) also showed the same behavior (not unsetting the mask). I updated the PR and rewrote the test from scratch to cover as most cases a possible.

core/src/avm1/globals/movie_clip.rs
```diff
@@ -1045,19 +1045,29 @@ fn set_mask<'gc>(
...
+                let mask = activation.resolve_target_display_object(start_clip, *m, false)?;
+                if mask.is_none() {
+                    return Ok(Value::Bool(false));
+                }
+                mask
+            }
+        }
+    } else {
+        return Ok(Value::Undefined);
+    };
```
Herschel, line 1062: One more style nit. This might be cleaner with a single `match` instead of both `if` and `match`:
```
let mask = match args.get(0) {
   None => return Ok(Value::Undefined),
  Some(Value::Undefined | Value::Null) => None,
  Some(m) => // ... etc
}
```

Herschel (approved):
Thank you!

Toad06:
Thank you for the review and advices. 😉

PR #8188, opened 2022-10-04
core: Implement XMLSocket

Minigugus:
Fixes #905
Fixes #1016
Linked #288
Linked #5134
Relates to #3042
Rewrites #5160

Add `XMLSocket` class.
The default behavior is unchanged: without explicit user opt-in, XMLSocket always rejects connections.

* [X] `XMLSocket` class
* [X] sockets management and integration with backends
* [X] Desktop implementation
* [ ] Web implementation¹
* [ ] Tests

¹ Since web environments doesn't allows native TCP connections, the web implementation will let the user (the one instantiating the Ruffle player) provide its own XMLSocket transport, just like with the External Interface API (over websocket for instance).

PR #10433, opened 2023-03-28
avm2: Stub flash.net.XMLSocket

evilpie:
There is a draft PR #8188 for a real implementation, but in the meantime the stubbed version is useful for Coaster Racer and maybe other games using FlashConnect.


Lord-McSweeney:
I think a "correct" implementation of `XMLSocket` would be a wrapper around `Socket`; I think more games use the latter than the former.

Herschel (approved):
Can't hurt to stub it in the meantime -- thank you!

PR #10434, opened 2023-03-28
web: Give toggleable option to reload SWFs (fix #1885)

### LINKED ISSUE #1885: web: Add "Download SWF" and "Reload SWF" options to the context menu
Maybe these should just be in a debug build but I think it'd be useful to have a "Download SWF" and a "Reload SWF" option

Are there potential problems with facilitating easy downloading of SWFs? Would people get upset?

danielhjacobs:
Also does the same extension option reordering as #9904. This option may be better as extension only, however, I think it would actually surprisingly be harder to not include it for self-hosted.

### DIFF
--- a/web/packages/core/src/config.ts
+++ b/web/packages/core/src/config.ts
@@ -19,6 +19,7 @@ export const DEFAULT_CONFIG: Required<BaseLoadOptions> = {
     warnOnUnsupportedContent: true,
     logLevel: LogLevel.Error,
     showSwfDownload: false,
+    showReload: false,
     contextMenu: true,
     // Backwards-compatibility option
     preloader: true,
--- a/web/packages/core/src/load-options.ts
+++ b/web/packages/core/src/load-options.ts
@@ -223,6 +223,14 @@ export interface BaseLoadOptions {
     showSwfDownload?: boolean;
 
     /**
+     * If set to true, the context menu has an option to reload
+     * the content.
+     *
+     * @default false
+     */
+    showReload?: boolean;
+
+    /**
      * Whether or not to show a context menu when right-clicking
      * a Ruffle instance.
      *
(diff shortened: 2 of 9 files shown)


n0samu:
Why did you decide to make `showReload` set to `false` by default? What's the harm in showing the option by default?

danielhjacobs:
Mainly worried someone would accidentally click it while far into a game. Open for debate. If I show it by default, is it even worth making it a toggle, or should it just always appear in the context menu?

danielhjacobs:
Thinking about it, it may be a good idea to use https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm in reloadSwf

n0samu:
Sounds good. And yeah, I wouldn't have bothered to make a toggle at all, though I don't have any objection to it existing either.

HaimZik:
I think making it true by default will clutter the UI, I think the mindset of what can you remove is better.
Especially on mobile you want the context menu to be minimal as possible.

I don't see a use case for this outside the desktop app and the demo page.

Herschel:
Sorry if this is a dumb question! What's the motivation when the user can reload the page themselves? I also would like to avoid cluttering the UI with options unless there's a big use case.

n0samu:
It would be most useful in the web demo since it saves you from needing to re-select the file every time. I wasn't thinking much about clutter - maybe we should leave it as an off-by-default option and enable it in the web demo's default options?

Herschel:
I see! To me, this feels like specialized UI that the demo page should have, because it runs a locally uploaded file, and not necessary as an option in the player itself. Similar to how the demo page has the ℹ️ information button, there could be a 🔃 icon.

danielhjacobs:
I'll close this in favor of #10480. I agree with https://github.com/ruffle-rs/ruffle/pull/10426#issuecomment-1489865634 that this would be a good idea for the desktop GUI as well. Once #10480 is merged, #1885 can probably be closed as mostly completed and partially wontfix.

PR #10440, opened 2023-03-29
avm2: Add stub warning when loader loads images

Lord-McSweeney:


core/src/avm2/globals/flash/display/loader.rs
```diff
@@ -78,12 +79,18 @@ pub fn load<'gc>(
             )?
             .as_object()
             .unwrap();
+        
+        let stringified_url = url.to_string();
+        
+        if stringified_url.ends_with(".jpg") || stringified_url.ends_with(".png") {
```
n0samu, line 85: The AS3 docs say that Loader supports loading GIFs as well

core/src/avm2/globals/flash/display/loader.rs
```diff
@@ -78,12 +79,18 @@ pub fn load<'gc>(
...
+            avm2_stub_method!(activation, "flash.display.Loader", "load", "loading images");
+        }
 
         let future = activation.context.load_manager.load_movie_into_clip(
             activation.context.player.clone(),
             content.into(),
             // FIXME - set options from the `URLRequest`
-            Request::get(url.to_string()),
+            Request::get(stringified_url),
             Some(url.to_string()),
```
n0samu, line 94: For consistency this should use `stringified_url` as well, right?

Aaron1011:
Loading image is already implemented - if it's not working, that's a bug.

PR #10465, opened 2023-03-30
avm2: Convert RegExp to ActionScript

Aaron1011:


core/src/avm2/globals/reg_exp.rs
```diff
@@ -1,25 +1,17 @@
...
-    &[("exec", exec), ("test", test)];
+pub use crate::avm2::object::reg_exp_allocator;
 
-/// Implements `RegExp`'s instance initializer.
-pub fn instance_init<'gc>(
+/// Implements `RegExp`'s `init` method, which is called from the constructor
+pub fn init<'gc>(
     activation: &mut Activation<'_, 'gc>,
     this: Option<Object<'gc>>,
     args: &[Value<'gc>],
```
Lord-McSweeney, line 17: Is the `super_init` in this native constructor necessary anymore?

Aaron1011: Good catch

core/src/avm2/globals/RegExp.as
```diff
@@ -0,0 +1,31 @@
...
+
+        private native function init(re:*, flags:*):void;
+
+        public native function get dotall():Boolean;
+        public native function get extended():Boolean;
+        public native function get global():Boolean;
+        public native function get ignoreCase():Boolean;
+        public native function get multiline():Boolean;
+        public native function get lastIndex():Number;
+        public native function set lastIndex(value:Number):void;
```
Herschel, line 17: `lastIndex: int`

core/src/avm2/globals/RegExp.as
```diff
@@ -0,0 +1,31 @@
...
+        public native function get source():String;
+
+        AS3 native function exec(str:String):Object;
+        AS3 native function test(str:String):Boolean;
+
+        prototype.exec = function(str:String):Object {
```
Lord-McSweeney, line 23: `test` and `exec` should both take an empty string as a parameter by default (in both the `AS3` and the `prototype` methods)

Herschel (approved):

PR #7534, opened 2022-08-01
demo: Show metadata on information icon click

danielhjacobs:
Info icon created based on https://www.svgrepo.com/svg/24584/info-icon (which is licensed with a CC0 license). See also https://github.com/ruffle-rs/ruffle/wiki/SWF-version-chart

Fixes #7492

Look on a regular computer (changed uncompressedLength display to use bit-shift instead of Math.round and division since screenshot, so number is now one less for Fleeing the Complex):

Screenshot from 2022-08-03 09-40-31

Look on 320px x 320px screen (no screen is 320px x 320px, but the ability to switch between portrait and landscape orientation means I need to take a 320px height and 320px width into account):

Screenshot from 2022-08-03 09-41-44

PR #10480, opened 2023-03-31
demo: Add option to reload the current SWF

danielhjacobs:
Reload icon created based on https://www.svgrepo.com/svg/138915/reload (which is licensed with a CC0 license, which is the same license held by the information icon added in #7534).


n0samu (approved):
Works perfectly, thank you!

n0samu:
On second thought... reloading the SWF after a panic doesn't seem to work. Might there be a fix?

danielhjacobs:
Well, not with the code working as it does currently. But it may work to make data a global `let` variable and have this call `load({ data: data, swfFileName: file.name, ...defaultConfig }` instead of `player.load(player.loadedConfig)` when a local fileis reloaded. Though I haven't tested that:

https://github.com/ruffle-rs/ruffle/blob/d969358a200132c1d3bacaa4c2d690640cf836df/web/packages/demo/www/index.js#L147-L148

n0samu:
I think there are more fundamental issues preventing `player.load` calls from working after a panic, see the recent discussion in #dev-general on Discord

Herschel (approved):
Works great, thank you!

To gracefully deal with panic I suspect we'd have to trash and re-instantiate the entire wasm module, which might be painful but should theoretically be possible (clearing `lastLoaded` in `load-ruffle.rs`). We can always revisit that for a future PR.

PR #10482, opened 2023-03-31
avm2: Fixed BitmapData.setPixels() not respecting ByteArray.position

Dinnerbone:
The provided test fails before prior to the commit.

We assumed every ByteArray would start at 0, both in terms of position and where the data actually was. This fixes that, by making it use the typical read methods.

Also fixes the fact that we weren't setting the BMD as dirty after a `setPixels`.
Also adds an optimization to skip any GPU->CPU sync if we're filling the whole BMD.


Herschel (approved):
Thank you!

PR #10494, opened 2023-04-01
core: Don't overwrite font descriptor with a duplicate

Aaron1011:
If a SWF contains multiple DefineFont tags with the same font name (but different font IDs), the first tag will win when a font is looked up by *name*. This affects the behavior of EditText objects, which can have embedded HTML like `<font face="MyFontName">` which performs a font lookup by name.

This fixes Fancy Pants World 4 Part 3, which contains two DefineFont3 tags with the name FancyFont. The second font is missing many glyphs, so using it causes us to be unable to render the squiggle and life count text.


Herschel (approved):
Thank you! Eventually the font map needs to change to allow each font name to have one font of each style (bold, bold-italic, etc., per discussion at https://github.com/ruffle-rs/ruffle/pull/10181).

PR #10495, opened 2023-04-01
core: Log an error when a font is missing a glyph for a character

Aaron1011:
My hope is that this never happens for real SWFS (unless Ruffle has a bug). If this does get hit intentionally in real SWFs, then it's going to be very verbose (logging each time we render), so we'll need to adjust this to log only once for a given char.

### DIFF
--- a/core/src/font.rs
+++ b/core/src/font.rs
@@ -289,6 +289,12 @@ impl<'gc> Font<'gc> {
                 // Step horizontally.
                 transform.matrix.tx += twips_advance;
                 x += twips_advance;
+            } else {
+                tracing::error!(
+                    "Font {:?} is misssing glyph for character {:?}",
+                    self.0.descriptor,
+                    c
+                );
             }
         }
     }


Herschel:
In my experience, this happens quite often in real SWFs, because it's easy to forget to embed a font or all of the correct glyphs in the Flash IDE, so IMO I don't think it's particularly unexpected or requires an error message

Toad06:
Something I observed in a few AVM1 games is that if a glyph for a character is missing, then the Flash Player will try to select the closest glyph for the specified character. This behavior only seems to happen when dealing with uppercase/lowercase characters.

For instance, [in this math game](https://github.com/ruffle-rs/ruffle/files/5505498/mathgame.zip), there is no "m" glyph in the embedded font, but there is a "M": if you type "m" on the keyboard using Ruffle, nothing will be displayed but in the Flash Player, "M" will be displayed. Also worth noting that the font also includes "e" and "é" glyphs, but typing "ê" won't display "e" in Flash Player.

Herschel:
Wow, didn't know the uppercase/lowercase fallback behavior at all. Could we open an issue for that, @Toad06?

Also, in general, we allow inputting characters with missing glyphs into the text input, seen if you `trace(textfield.text)` even though there is no glyph to render, but in FP this input is ignored, and the textfield text doesn't change. (You can of course set `textfield.text = "whatever";` in AS and it will stick.)

Aaron1011:
I'm closing this for now.

PR #10497, opened 2023-04-01
avm2: QName and RegExp fixes

Lord-McSweeney:
1. `QName` does not define `valueOf` on its prototype.
2. `RegExp`'s prototype should not be enumerable, and it should have a `toString` defined on its prototype (but _not_ as an `AS3` one).

### DIFF
--- a/core/src/avm2/globals/RegExp.as
+++ b/core/src/avm2/globals/RegExp.as
@@ -29,23 +29,7 @@ package {
         }
 
         prototype.toString = function():String {
-            var res:String = "/" + this.source + "/";
-            if (this.ignoreCase) {
-               res += "i";
-            }
-            if (this.global) {
-               res += "g";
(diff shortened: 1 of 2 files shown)


Herschel:
Don't see any `RegExp` changes -- missing a commit?

core/src/avm2/globals/RegExp.as
```diff
@@ -27,5 +27,29 @@ package {
         prototype.test = function(str:String = ""):Boolean {
             return this.AS3::test(str);
         }
+
+        prototype.toString = function():String {
```
Aaron1011, line 31: I think this can just call `valueOf`, which is implemented in `regexp_object`

Aaron1011 (approved):

Herschel (approved):
Thank you!

PR #10499, opened 2023-04-01
avm2: Add X509 certificate classes

onkrot:


Herschel:
You can use `__ruffle__.stub_getter` to indicate that these are unimplemented ([example](https://github.com/ruffle-rs/ruffle/blob/47f594d1080e2fe54cc6f853dc5f64adca4475b4/core/src/avm2/globals/flash/media/Microphone.as#L6))

core/src/avm2/globals/flash/security/X500DistinguishedName.as
```diff
@@ -0,0 +1,38 @@
+package flash.security
+{
+    public final class X500DistinguishedName
```
Herschel, line 3: Looks like this class has a `toString` method as well

Lord-McSweeney:
I don't think `stub_getter` is needed for these since the current implementation is technically correct- it just needs to be hooked up to `SecureSocket` if that (ever) is implemented.

Herschel:
> I don't think `stub_getter` is needed for these since the current implementation is technically correct- it just needs to be hooked up to `SecureSocket` if that (ever) is implemented.

I double checked, and I agree -- I was worried that these getters should return some default value or throw on a default instance, but just checked in Flash and looks like these indeed return `null`s if you construct your own instance.

(Funnily doing `new X509Certificate().encoded` crashes Flash CS6 immediately!)

Issue #10498, opened 2023-04-01
XML regression in Flipline games


n0samu:
### Describe the bug

Due to an XML regression caused by #10471, Flipline games such as [Papa's Burgeria](https://www.flipline.com/games/papasburgeria/index.html) are no longer loading properly in the Ruffle desktop app. To test this:
1. [Download the SWF file for the game](https://i.flipline.com/gamefiles/papasburgeria/papasburgeria_v2.swf).
2. Run the SWF file in the Ruffle desktop app. The game will never finish loading.

The error is:
`ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "complete", class: flash.events::Event, ptr: 0x1b8653b7b80 }) to handler FunctionObject(FunctionObject { ptr: 0x1b8652cd120 }) : RustError(EscapeError(UnterminatedEntity(46..51)))`

### Expected behavior

The games should load.

### Affected platform

Desktop app

### Operating system

Windows 10

n0samu:
Fixed by #10500

PR #10500, opened 2023-04-01
avm2: Use custom_unescape wherever we unescape XML

Aaron1011:
When we were using a fork of quick-xml, we modified the actual unescape method. Now that we're using the crates.io release again, we need to go through our `custom_unescape` function.


Herschel (approved):

n0samu:
Fixes #10498 👍

PR #10515, opened 2023-04-02
wgpu: Slight optimization to color shader

Herschel:
The color is the same through the whole shape and doesn't change per pixel, so calculate the color in the vertex stage instead of the fragment stage.

### DIFF
--- a/render/wgpu/shaders/color.wgsl
+++ b/render/wgpu/shaders/color.wgsl
@@ -23,17 +23,14 @@ struct VertexOutput {
 fn main_vertex(in: VertexInput) -> VertexOutput {
     #if use_push_constants == true
         var transforms = pc.transforms;
+        var colorTransforms = pc.colorTransforms;
     #endif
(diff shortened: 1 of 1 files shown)


Aaron1011 (approved):

Aaron1011:
If I understand correctly, this will change how we do color interpolation. Before, the un-transformed input color would get interpolated before being passed to the fragment shader. Now, the vertex shader is performing the transformation, so the transformed color will get interpolated. Do we know which behavior is correct for Flash?

Herschel:
That's correct, the interpolation would be different if vertices had different colors, but this shader is only used for solid color fills where every vertex has the same color and the color is the same across the whole shape, so there'd be no difference -- there's not really any Flash behavior to compare against.

(The interpolation order does matter for gradients https://github.com/ruffle-rs/ruffle/issues/10517, which we currently handle incorrectly -- I'll fix this in a future PR).

Dinnerbone (approved):

Issue #10377, opened 2023-03-26
Coaster Racer: Cannot start game


n0samu:
### Describe the bug

Link to SWF: http://i.notdoppler.com/files/coasterracer.swf

After clicking the Continue button on the loading screen of [Coaster Racer](http://www.notdoppler.com/coasterracer.php), the game displays nothing but a black screen. This error appears in the terminal:
`ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "click", class: flash.events::MouseEvent, ptr: 0x22521ded210 }) to handler FunctionObject(FunctionObject { ptr: 0x22502b735d0 }) : RustError("Attempted to call \"\" with 2 arguments (more than 0 is prohibited)")`

### Expected behavior

The game should start.

### Affected platform

Desktop app

### Operating system

Windows 10

yoganlava:
Hopefully progressed by https://github.com/ruffle-rs/ruffle/pull/10299

n0samu:
Well, with that PR the error is:
```
ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "click", class: flash.events::MouseEvent, ptr: 0x280293f7880 }) to handler FunctionObject(FunctionObject { ptr: 0x2800ae7ce40 }) : ReferenceError: Error #1074: Illegal write to read-only property nodeName on flash.xml.XMLNode.
        at org.flashdevelop.utils::FlashConnect$/createMsgNode()
        at org.flashdevelop.utils::FlashConnect$/trace()
        at Lic$/buttonSkipCPMStarAdPressed()
```

yoganlava:
> Well, with that PR the error is:
> ```
> ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "click", class: flash.events::MouseEvent, ptr: 0x280293f7880 }) to handler FunctionObject(FunctionObject { ptr: 0x2800ae7ce40 }) : ReferenceError: Error #1074: Illegal write to read-only property nodeName on flash.xml.XMLNode.
>         at org.flashdevelop.utils::FlashConnect$/createMsgNode()
>         at org.flashdevelop.utils::FlashConnect$/trace()
>         at Lic$/buttonSkipCPMStarAdPressed()
> ```

Yea makes sense, seems like only the getters are defined

n0samu:
[...]

PR #10519, opened 2023-04-02
avm2: Improve the Sound::extract stub

evilpie:
With this we get #10377 playable.


Dinnerbone:
The docs suggest that it'll also set `position`. I know it's a stub but that might help other games

Dinnerbone (approved):

Herschel (approved):
Thank you!

Issue #1289, opened 2020-10-10
Zombotron — TypeError


elsandosgrande:
As of the tenth of October, when attempting to play the original Zombotron with Ruffle desktop, whether with a local build which was built from the latest committed code or with the latest nightly build, [the following occurs (warning, flashing imagery)](http://mellemws.my.to/direct/Screencast_from_2020年10月10日_17時26分41秒.webm), as opposed to [the expected result](https://kernel.ubuntu.com/~kernel-ppa/mainline/). My laptop is running Ubuntu Focal with [the oibaf graphics PPA](https://launchpad.net/~oibaf/+archive/ubuntu/graphics-drivers) and [the latest mainline kernel](https://kernel.ubuntu.com/~kernel-ppa/mainline/).

The terminal log continuously produce just one line: 

```
***
[2020-10-10T15:26:58Z ERROR ruffle_core::display_object::movie_clip] Got TypeError: undefined is not an Object when constructing AVM2 side of display object
[2020-10-10T15:26:58Z ERROR ruffle_core::display_object::movie_clip] Got TypeError: undefined is not an Object when constructing AVM2 side of display object
[2020-10-10T15:26:58Z ERROR ruffle_core::display_object::movie_clip] Got TypeError: undefined is not an Object when constructing AVM2 side of display object
***
```

The SFW in question: http://mellemws.my.to/direct/zombotron.swf

---

If this ticket breaches the following terms of the contributing guidelines, I appologise in advance and shall note this to avoid future incidents:

* A "this SWF doesn't work at all" report (what about it doesn't work?)
* Duplicate issues for each piece of content using an unimplemented feature

kmeisthax:
I retested this game and it still flashes, but the logged error is now that it's taking too long to execute. If I comment out the execution limit on desktop (entirely not an option, AVM2 needs optimization), then I instead get a bunch of errors about not being able to access `flash.display.SimpleButton` and `flash.media.Sound`. So this is now waiting on AS3 button and sound access.

Lord-McSweeney:
This game works better now, but has the following issues:
At the beginning, the button fills backwards (not sure if this is intentional behavior)
The SWF can't be resized (see #8921)
[...]

PR #10533, opened 2023-04-03
core: Focus EditText fields only if editable

n0samu:
In AVM2, clicking any EditText field would set the focus, which sometimes interfered with games' keyboard controls. By adding a check in the edit_text module, this changes AVM2's behavior to match AVM1's. Fixes https://github.com/ruffle-rs/ruffle/issues/1289#issuecomment-1492946983.


relrelb (approved):

Herschel (approved):

n0samu:
Why does my PR show up as having all these other commits from other PRs in it now?

Toad06:
@n0samu Not too important but I noticed this regressed the [`avm1/selection` test](https://github.com/ruffle-rs/ruffle/tree/master/tests/tests/swfs/avm1/selection) a bit: if you run the SWF in Flash Player, you will notice that it is possible to interact with the "Selectable text" but in Ruffle this no longer traces that the focus is on this EditText. Maybe `is_selectable` should be used instead of `is_editable`?

For reference (because Github made such a mess), this is commit 315a7c87e52d06aca418d2a764adc7183636c772.

n0samu:
Thanks for letting me know! That also made me curious about Flash's behavior when an EditText field is editable but not selectable. The Flash editor does not allow creating text fields like that, but here is one edited in JPEXS: [TextInputAVM2_EditableNonSelectable.zip](https://github.com/ruffle-rs/ruffle/files/11216962/TextInputAVM2_EditableNonSelectable.zip)
Flash Player actually handles this case in a logical way, allowing you to type into the text field but not allowing selection or showing the caret cursor.

PR #10542, opened 2023-04-04
avm2/naga-agal: Some naga-agal fixes, add fractal rendering test

Herschel:
Allows [this old Stage3D fractal renderer](https://github.com/Herschel/molehill-julia) to run, and adds it as a test case.

* Implemented `Context3d.setColorMask`
* Add support of `abs` AGAL op in `naga-agal`
* Allow multi-component boolean ops in `naga-agal` (verify me)
* Added stage3d_fractal image test
* Added complex_fractal test to `naga-agal`

@Aaron1011 Please verify my fix for multi-component boolean binary ops in `naga-agal`. The AGAL instruction in question is:
`slt ft1, ft1.xxxx, fc2.yyyy`
which was being written as a cast-to-scalar followed by trying to store it in a vector.


render/naga-agal/src/builder.rs
```diff
@@ -910,33 +910,49 @@ impl<'a> NagaBuilder<'a> {
...
+        let num_dest_components = [Mask::X, Mask::Y, Mask::Z, Mask::W]
+            .into_iter()
+            .filter(|mask| dest.write_mask.contains(*mask))
+            .count();
+
         // Optimization - use a Store instead of writing individual fields
         // when we're writing to the entire output register.
         if dest.write_mask.is_all() {
+            // Splat a scalar to the vector size, if necessary.
+            let expr = if source_is_scalar && num_dest_components != 1 {
```
Aaron1011, line 932: This is inside the `dest.write_mask.is_all()`, so `num_dest_components` will always be 4.

Herschel:
Took some more time to remember AGAL and learn more about naga. The boolean ops like `slt` were going through `Builder::first_components_binary_op` which seemed to assume that it only worked on scalars, checking for a `.xxxx` swizzle, but that doesn't seem right:  `slt vt0.xy, vt1.yx, vt2.zw` is valid, outtping 0.0 or 1.0 componentwise. I renamed this to `boolean_binary_op` and removed the single-component check.

Aaron1011 (approved):

Issue #9085, opened 2023-01-09
Youda Sushi Chef Demo loops


Lord-McSweeney:
### Describe the bug

https://www.shockwave.com/content/youdasushichef/sis/YoudaSushiChef_Demo.swf loops the creator logos endlessly, due to 2 errors:
`ERROR core/src/display_object/movie_clip.rs:2344 Error occured when running AVM2 frame script: RustError("Cannot access property ::init of null or undefined")`

which might be caused by the first error:

`ERROR core/src/display_object/movie_clip.rs:2020 Got RustError("Cannot access property ::loaderInfo of null or undefined") when constructing AVM2 side of movie clip of type game::App`

which _might_ be caused by the unimplemented `LocalConnection.connect` (`WARN core/src/avm2/globals/toplevel.rs:42 LocalConnection.connect is not implemented`)

### Expected behavior

The game does not loop and gets to the title screen.

### Affected platform

Browser's extension

### Operating system

Linux

### Browser

Chrome 108

Lord-McSweeney:
On further inspection, this also needs `flash.utils.describeType` and `flash.system.fscommand`.

n0samu:
Now there are a lot of other errors being spammed. I can't really go through them all but this game definitely needs full XML support at the very least.

Lord-McSweeney:
Newest error is `RustError("Cannot call null or undefined method ::child of class XML")`. The main menu flickers before going back to the creator logos.

Lord-McSweeney:
Next is `flash.system.fscommand`.

n0samu:
I'm getting this error:
`ERROR run_frame:run_all_phases_avm2: ruffle_core::display_object::movie_clip: Error occured when running AVM2 frame script: RustError("VerifyError: Trait concat in class ColorMatrix has same name as trait concat in class Array, but does not override it")`

Lord-McSweeney:
Seems to be working now, despite `fscommand` still missing (???). In the tutorial text, the text seems to be compressed incorrectly, though.

n0samu:
The text width issue looks similar to #9690. I'll open a new issue for the messed up graphics.

PR #10543, opened 2023-04-04
avm2: Array-related fixes

Lord-McSweeney:
1. toString and toLocaleString should only be defined on the prototype.
2. concat should only be defined as an as3 and proto property, not as an instace property.
3. Array doesn't have a valueOf defined directly on it.

Progresses #9085 a little.


Herschel (approved):
Thank you!

PR #10544, opened 2023-04-04
web: Avoid triggering cloudflare checks

### LINKED ISSUE #10037: Extension conflicts with Cloudflare site protection
### Describe the bug

When Ruffle extension is enabled in a browser, it breaks "Verify that you are human" bot blocker on some sites (e.g. `chat.openai.com`).

The verification process is stuck in an infinite loop of `"Click 'Verify'" -> "Wait for page to reload" -> "Click it again"`

### Expected behavior

1. Enable Ruffle
2. Go to Cloudflare-protected site
3. Verification passes, site works

### Affected platform

Browser's extension

### Operating system

Ubuntu 20.04

### Browser

Google Chrome 108.0

### Additional information

Checked in every possible case (e.g. with or without other extensions, with or without strict privacy browser prefrences, with or without chrome flags), everything leads to Ruffle extension breaking Cloudflare protected sites.

Lord-McSweeney:
This should fix #10037 (at least, for cloudflare protection).

### DIFF
--- a/web/packages/core/src/plugin-polyfill.ts
+++ b/web/packages/core/src/plugin-polyfill.ts
@@ -46,7 +46,9 @@ class RuffleMimeTypeArray implements MimeTypeArray {
     }
 
     item(index: number): MimeType {
-        return this.__mimeTypes[index]!;
+        // This behavior is done to emulate a 32-bit uint,
+        // which browsers use.
+        return this.__mimeTypes[index >>> 0]!;
     }
 
     namedItem(name: string): MimeType {
@@ -117,7 +119,10 @@ class RufflePluginArray implements PluginArray {
     }
 
     item(index: number): Plugin {
-        return this.__plugins[index]!;
+        // This behavior is done to emulate a 32-bit uint,
+        // which browsers use. Cloudflare's anti-bot
+        // checks rely on this.
+        return this.__plugins[index >>> 0]!;
     }
 
     namedItem(name: string): Plugin {


web/packages/core/src/plugin-polyfill.ts
```diff
@@ -46,7 +46,9 @@ class RuffleMimeTypeArray implements MimeTypeArray {
     }
 
     item(index: number): MimeType {
-        return this.__mimeTypes[index]!;
+        // This weird modulo behavior is done to emulate
+        // the behavior of browsers.
+        return this.__mimeTypes[index % 2 ** 32]!;
```
Herschel, line 51: I wonder if this would be more accurate as `index>>>0` to more closelt match the 32-bit int behavior.

Herschel:
Thanks for looking into this!

web/packages/core/src/plugin-polyfill.ts
n0samu: Hmm... testing this out, `navigator.plugins[-(2**32) + 1]` returns `undefined`, but if it behaved how you suggest, it would return the second element of the PluginArray.

Herschel: `navigator.plugins.items[-(2**32) + 1]` returns the second element on my end -- raw access of the array `navigator.plugins[2**32]` does not work

n0samu: Ah, makes sense! 👍

Herschel (approved):

Issue #2189, opened 2020-12-31
[Web] Cursor missing when selecting input field


x3fold:
On the web version, when selecting an input text field the blinking text cursor is missing (Nightly 2020-12-31)
The inputting of text is still working fine, just no cursor

Was working fine on 2020-12-21

Toad06:
Thanks for the report.

The caret color no longer matches the text color on the web backend (was implemented in https://github.com/ruffle-rs/ruffle/commit/6f3251a4e7ec3673b6d4706295a387994b5cbcce) and seems to be always white. On desktop, it's still incorrect when the field has no character at all.

I ran a bisect and it seems the issue started with commit 3ff6a37ff49ab0971aa612cd78c66db3638ec7f9.

A few affected games - all in their password section:
[Bouncy The Ball](https://www.newgrounds.com/portal/view/204189)
[Soap Bubble](https://www.newgrounds.com/portal/view/52537)
[Super Chick Sisters](https://www.newgrounds.com/portal/view/396372)

Toad06:
Reopening, as this is still an issue on some files:

[Super Chick Sisters](https://www.newgrounds.com/portal/view/396372) (password section)
[Ultimate Flash Sonic](https://www.newgrounds.com/portal/view/151706) (password section)
[Windows Doors](https://www.newgrounds.com/portal/view/512482) (click "Next")

(still only happening on the web backend with WebGL)

n0samu:
Can confirm that this problem is still occurring in Ultimate Flash Sonic and Windows Doors with the WebGL backend only. (With wgpu-webgl, the cursor shows up normally.) I wasn't able to check Super Chick Sisters.

PR #10551, opened 2023-04-04
avm1: Fix some issues with `Selection.getFocus()` and `setFocus()`

### LINKED ISSUE #2789: Selection.setFocus-function problem
Allthough input textFields are working fine, it seems that the selection.setFocus-function cannot be executed from via actionsript.

[Selection.setFocus_not_working.zip](https://github.com/ruffle-rs/ruffle/files/5859971/Selection.setFocus_not_working.zip)

Toad06:
- `Selection.getFocus()` now returns a string on success, instead of an object. Our test logs the correct values because of the string coercion when using `trace`.

- `Selection.setFocus(str)` now accepts strings like `"focus"` or `"_root.focus"`. This fixes #2789.

- `Selection.setFocus(focus)` to a text field was giving the focus internally, however it wasn't possible to type anything on the keyboard. This can also be observed in the issue mentioned above. Note that it can be difficult to notice that the field has focus because of #2189 (now also happens on desktop on my end).

- `Selection.setFocus(focus)` now returns `true` on success, `false` on failure. I updated the test output to that end. Note that it's possible for `setFocus` to fail and return `false` if the file itself doesn't have the focus (for instance on a webpage, if you click an element other than the Flash file). The onSetFocus event should also be triggered. This is not implemented by this PR.

### DIFF
--- a/core/src/avm1/globals/selection.rs
+++ b/core/src/avm1/globals/selection.rs
@@ -2,7 +2,7 @@ use crate::avm1::activation::Activation;
 use crate::avm1::error::Error;
 use crate::avm1::globals::as_broadcaster::BroadcasterFunctions;
 use crate::avm1::property_decl::{define_properties_on, Declaration};
-use crate::avm1::{Object, ScriptObject, TObject, Value};
+use crate::avm1::{Object, ScriptObject, Value};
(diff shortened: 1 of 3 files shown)


Herschel (approved):
Thank you!

PR #10557, opened 2023-04-05
web: Split Firefox extension upload into a separate process

kmeisthax:
This removes all extension upload related processes from nightlies - specifically addons.mozilla.org uploads. Unsigned nightlies will still be compiled and provided as normal.

A new weekly upload workflow has been added for actually submitting the latest nightly to Mozilla for addons.mozilla.org. Since we have automated source code submission, this should Just Work™ in one click, and I no longer need to add and remove the extension ID secret every time I want to turn extension upload off and on. The weekly upload process can also be triggered manually at any time if something should fail - it does not create any Git tags.

### DIFF
--- a/.github/workflows/release_amo.yml
+++ b/.github/workflows/release_amo.yml
@@ -1,0 +1,60 @@
+name: Release to addons.mozilla.org
+
+on:
+  #Run weekly every Friday
+  schedule:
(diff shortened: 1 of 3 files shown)


Herschel (approved):
Thank you!

PR #10562, opened 2023-04-05
avm2: stub URLRequest.requestHeaders

AllinolCP:


core/src/avm2/globals/flash/net/URLRequest.as
```diff
@@ -42,5 +45,18 @@ package flash.net {
 		public function get contentType():String {
 			return this._contentType;
 		}
+
+        public function get requestHeaders():Array
```
Lord-McSweeney, line 49: I think the getter and setter should be indented.

AllinolCP: Reolved in latest commit

Lord-McSweeney (approved):

Herschel (approved):
Thank you!

PR #10565, opened 2023-04-05
desktop: Add base and max execution duration options

n0samu:
These options already existed on web. This makes them available in the desktop app as well.


Herschel (approved):
Thank you!

PR #10572, opened 2023-04-05
Clean up and expand `ruffle_wstr`'s API

moulins:
Various `wstr`-related refactorings extracted from my experimental string interning branch.


Herschel (commented):
* I wonder if `decode_swf_str` might be better as an extension method, something like `s.decode(self.encoding)`
 * And/or, making `AvmString::from_swf_str(gc_context, s, encoding)`, because `decode_swf_str` -> `AvmString::new` is the most common pattern.

moulins:
> - I wonder if decode_swf_str might be better as an extension method, something like s.decode(self.encoding)
> - And/or, making AvmString::from_swf_str(gc_context, s, encoding), because decode_swf_str -> AvmString::new is the most common pattern.

Good idea! I've tried both routes, and the extension trait ended up being the cleanest one, so I've gone with that.

Dinnerbone (approved):

PR #10575, opened 2023-04-05
avm2: Call 'AS3::concat' in ColorMatrixFilter

### LINKED ISSUE #10552: Learn To Fly 3 - Stuck after Clicking on Story Mode
### Describe the bug

Loading SWF file https://www.spiele-umsonst.de/azad/downloads/flashgames/learntofly3.swf
ruffle_web-wasm_extensions.js:681 ERROR core/src/avm2/events.rs:419 Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x70f0d60 }) to handler FunctionObject(FunctionObject { ptr: 0x5a4bf80 }) : RustError("VerifyError: Trait concat in class ColorMatrix has same name as trait concat in class Array, but does not override it")

Here is the error I get when I want to start the story mode

### Expected behavior

Start it?

### Affected platform

Online demo

### Operating system

All

### Browser

All

### Additional information

_No response_

Aaron1011:
This ensures that we call the normal Array concat method (or a method that overrides it). Some SWFs may define a *public* concat method in an Array subclass, with a different signature.

Fixes #10552

### DIFF
--- a/core/src/avm2/globals/flash/filters/ColorMatrixFilter.as
+++ b/core/src/avm2/globals/flash/filters/ColorMatrixFilter.as
@@ -16,17 +16,20 @@
 
 		// From the Flash docs, we need to make a copy of the `Array`,
 		// as modifying the `filter.matrix` directly should have no effect.
+		// We call the method in the AS3 namespace, as some SWFS define a
+		// extend `Array` and declare a *public* 'concat' method with a
+		// different signature.
 
 		public function get matrix(): Array {
-			return this._matrix.concat();
+			return this._matrix.AS3::concat();
 		}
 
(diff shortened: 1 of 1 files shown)


n0samu (approved):

Herschel:
I think there could be the same problem if an Array subclass defined an `AS3 override function concat() { ... }`

Lord-McSweeney:
I don't think this is the correct solution; there seem to be some additional checks- see my comment in #10055:
1. The array should be padded/cropped to 20 items.
2. Every element of the array should be coerced to a `Number`.

Aaron1011:
@Lord-McSweeney That can be addressed in a follow-up PR

PR #10583, opened 2023-04-06
BitmapData optimisations: faster copies (including draws that are copies)

Dinnerbone:
Main optimizations:
- Removed in-loop checks for "is pixel inside bitmapdata" for most methods, only iterate pixels that are valid to begin with
- `BitmapData.copyPixels()` will do its best to find the fastest way to copy a region on the cpu, including just sometimes straight up replacing the entire `Vec<Color>` if it needs to
- `BitmapData.draw(anotherBitmapData)` will recognizes cases that are trivial copies and act like `copyPixels()`, and do those on the CPU efficiently too. This means we don't need to sync those draws back, so it can be really fast for games that just use BitmapData as a kind of staging texture.

A side effect of this is that **some** `BitmapData.draw()` will work on backends that don't current support it, as the renderer isn't involved.

I was originally trying to get a `copy_on_gpu` method in here but I hit some walls with that, so this has that stripped out.

### DIFF
--- a/core/src/bitmap/bitmap_data.rs
+++ b/core/src/bitmap/bitmap_data.rs
@@ -438,6 +438,14 @@ mod wrapper {
                 .render_bitmap(handle, context.transform_stack.transform(), smoothing);
         }
 
+        pub fn can_read(&self, read_area: PixelRegion) -> bool {
+            if let DirtyState::GpuModified(_, area) = self.0.read().dirty_state {
+                !area.intersects(read_area)
+            } else {
+                true
+            }
+        }
+
         pub fn is_point_in_bounds(&self, x: i32, y: i32) -> bool {
             x >= 0 && x < self.width() as i32 && y >= 0 && y < self.height() as i32
         }
@@ -594,6 +602,14 @@ impl<'gc> BitmapData<'gc> {
     #[inline]
     pub fn get_pixel32_raw(&self, x: u32, y: u32) -> Color {
         self.pixels[(x + y * self.width()) as usize]
+    }
+
+    pub fn raw_pixels_mut(&mut self) -> &mut Vec<Color> {
+        &mut self.pixels
+    }
+
+    pub fn raw_pixels(&self) -> &Vec<Color> {
+        &self.pixels
     }
 
     // Updates the data stored with our `BitmapHandle` if this `BitmapData`
--- a/core/src/bitmap/bitmap_data.rs
+++ b/core/src/bitmap/bitmap_data.rs
@@ -608,7 +608,7 @@ impl<'gc> BitmapData<'gc> {
         &mut self.pixels
     }
 
-    pub fn raw_pixels(&self) -> &Vec<Color> {
+    pub fn raw_pixels(&self) -> &[Color] {
         &self.pixels
     }
 
--- a/core/src/bitmap/operations.rs
+++ b/core/src/bitmap/operations.rs
@@ -355,8 +355,8 @@ pub fn perlin_noise<'gc>(
 pub fn copy_channel<'gc>(
     mc: MutationContext<'gc, '_>,
     target: BitmapDataWrapper<'gc>,
-    dest_point: (u32, u32),
-    src_rect: (u32, u32, u32, u32),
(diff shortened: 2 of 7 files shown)


core/src/bitmap/operations.rs
```diff
@@ -1201,6 +1201,168 @@ pub fn apply_filter<'gc>(
...
+        write.set_cpu_dirty(dest_region);
+    } else {
+        let dest = dest.sync();
+        let mut dest_write = dest.write(context.gc_context);
+        let source_read = source.read_area(source_region);
+
+        if !blend && (dest_write.transparency() || !source_read.transparency()) {
+            // Copying anything to a transparent texture,
+            // or copying an opaque texture to an opaque texture,
+            // means we can skip alpha premultiplication
```
Herschel, line 1248: This comment was a little confusing to me, because copying a transparent texture onto a transparent texture still causes blending when `mergeAlpha` -> `blend` is true. The logic is correct (checking `blend`), only the comment is confusing.

core/src/bitmap/operations.rs
```diff
@@ -1201,6 +1201,168 @@ pub fn apply_filter<'gc>(
...
+                        dest_write.set_pixel32_raw(
+                            dest_region.x_min + x,
+                            dest_region.y_min + y,
+                            color,
+                        );
+                    }
+                }
+            }
+        } else {
+            // Copying a transparent texture to an opaque texture, or blending
```
Herschel, line 1272: Same comment nit as above, this code path also happens when copying transparent onto transparent with `mergeAlpha`.

core/src/bitmap/operations.rs
```diff
@@ -1201,6 +1201,168 @@ pub fn apply_filter<'gc>(
...
+            // means we can skip alpha premultiplication
+
+            if dest_region == source_region
+                && dest_region.width() == dest_write.width()
+                && dest_region.height() == dest_write.height()
+                && dest_region.width() == source_read.width()
+                && dest_region.height() == source_read.height()
+            {
+                // Copying an entire texture that's the same size and type? Just replace the whole thing
+                *dest_write.raw_pixels_mut() = source_read.raw_pixels().to_owned();
```
Herschel, line 1257: Can this be something like `copy_from_slice` instead of allocing a new vec?

core/src/bitmap/bitmap_data.rs
```diff
@@ -596,6 +604,14 @@ impl<'gc> BitmapData<'gc> {
         self.pixels[(x + y * self.width()) as usize]
     }
 
+    pub fn raw_pixels_mut(&mut self) -> &mut Vec<Color> {
+        &mut self.pixels
+    }
+
+    pub fn raw_pixels(&self) -> &Vec<Color> {
```
Herschel, line 611: This can return `&[Color]`

core/src/bitmap/operations.rs
Dinnerbone: Yeah sorry, I wrote that before I introduced a blending path. I'll fix it up!

Dinnerbone: Actually looking at it, it's because I said "copying", which means no blending. I get that that can be confusing.

core/src/bitmap/operations.rs
Dinnerbone: I thought a clone would be faster than copy_from_slice, but I didn't benchmark it. If we go copy_from_slice we can expand the scope of this though to include any region within two same-sized BMDs.

core/src/bitmap/operations.rs
```diff
@@ -1201,6 +1201,171 @@ pub fn apply_filter<'gc>(
     }
 }
 
+#[allow(clippy::too_many_arguments)]
+fn copy_on_cpu<'gc>(
+    context: &mut UpdateContext<'_, 'gc>,
```
Lord-McSweeney, line 1206: This only ever uses a `MutationContext`- would it be (slightly) more performant to take a `MutationContext` over an `UpdateContext`?

Dinnerbone: Changed!

PR #10589, opened 2023-04-07
naga-agal: Implement remaining ops

Herschel:
First pass at implementing remaining AGAL opcodes.
 * log
 * exp
 * pow
 * ddx
 * ddy
 * sge
 * kil
 * m33
 * m34


Aaron1011:
Could you add a naga-agal test for these opcodes? You should be able to use AGALMiniAssembler to generate bytecode containing these opcodes

Aaron1011:
I've updated this PR to the latest `naga`, and added some basic tests.

Lord-McSweeney:
The first commit also adds the `min`, `rsq`, and `sge` ops, should that be noted in the commit message?

Aaron1011:
@Lord-McSweeney I've updated the commit message

render/naga-agal/src/builder.rs
```diff
@@ -1060,45 +1097,54 @@ impl<'a> NagaBuilder<'a> {
...
 
+                let (num_rows, ty, vec_size, out_size) = match opcode {
+                    Opcode::M33 => (
+                        3,
+                        self.matrix3x3f,
+                        VectorSize::Tri,
+                        VertexAttributeFormat::Float3,
+                    ),
+                    Opcode::M34 => (
+                        3,
```
iwannabethedev, line 1115: I do not understand the code or this area of the codebase or know about NAGA or AGAL, but, is it correct that this value is `3` instead of `4`?

https://github.com/ruffle-rs/ruffle/pull/10589/files?diff=unified&w=1#diff-de78bf7c3d3c1d8988c51e39816e4dd5d4bd731f862ce46da4afb97342464c1aR376 seems, at a superficial glance at least, like it maybe should be 4 - unless rows and columns are counted differently, maybe. But I know very little about this.

Dinnerbone (approved):
Looks ok to me!

PR #10594, opened 2023-04-07
avm1: Avoid cloning `NativeObject`

relrelb:
Previously `TObject::native` used to clone the underlying `NativeObject` each time, which prevented it from being mutated. This was worked-around by wrapping the various `NativeObject` variants with an extra `GcCell`, which provides an additional layer of 
 interior mutability.

To avoid this unnecessary complexity, make `TObject::native` return an `Option<Ref<NativeObject<'gc>>>`, without cloning the 
 underlying `NativeObject`. This allows replacing said `GcCell`s with `Box<RefCell<T>>`.

### DIFF
--- a/core/src/avm1/globals/bevel_filter.rs
+++ b/core/src/avm1/globals/bevel_filter.rs
@@ -449,8 +449,8 @@ fn method<'gc>(
         return Ok(this.into());
     }
 
-    let this = match this.native() {
-        NativeObject::BevelFilter(bevel_filter) => bevel_filter,
+    let this = match this.native().as_deref() {
+        Some(NativeObject::BevelFilter(bevel_filter)) => *bevel_filter,
         _ => return Ok(Value::Undefined),
     };
 
--- a/core/src/avm1/globals/bevel_filter.rs
+++ b/core/src/avm1/globals/bevel_filter.rs
@@ -5,7 +5,8 @@ use crate::avm1::object::NativeObject;
 use crate::avm1::property_decl::{define_properties_on, Declaration};
 use crate::avm1::{Activation, Error, Object, ScriptObject, TObject, Value};
 use crate::string::{AvmString, WStr};
-use gc_arena::{Collect, GcCell, MutationContext};
+use gc_arena::{Collect, MutationContext};
+use std::cell::RefCell;
 use swf::Color;
 
 #[derive(Copy, Clone, Debug, Collect)]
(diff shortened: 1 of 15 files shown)


moulins:
Isn't this vulnerable to double-borrows? Consider the following AS2 code:
```js
var filter = new flash.filters.BlurFilter();
filter.counter = 0;

var proxy = {
  valueOf: function() { return ++filter.counter; }
};

filter.blurX = proxy;
```

The native `blurX` setter will end up calling `proxy.valueOf` while holding a `cell::Ref` to the `filter` object, which will cause a double-borrow panic when it tries to increment the `counter` property.

-----

I think the `Gc` indirection around native data has to stay, so that you can hold a reference to it without locking the underlying `ScriptObject`.

However, what would be possible is something like this:
```rs
#[derive(Clone, Debug, Collect)]
#[collect(require_static)]
struct BlurFilterData {
    blur_x: Cell<f64>,
    blur_y: Cell<f64>,
    quality: Cell<i32>,
}

#[derive(Copy, Clone, Debug, Default, Collect)]
#[collect(no_drop)]
#[repr(transparent)]
pub struct BlurFilter<'gc>(Gc<'gc, BlurFilterData>);
```

relrelb:
> Isn't this vulnerable to double-borrows? Consider the following AS2 code:
> 
> ```js
> var filter = new flash.filters.BlurFilter();
> filter.counter = 0;
> 
> var proxy = {
>   valueOf: function() { return ++filter.counter; }
> };
> 
> filter.blurX = proxy;
> ```
> 
> The native `blurX` setter will end up calling `proxy.valueOf` while holding a `cell::Ref` to the `filter` object, which will cause a double-borrow panic when it tries to increment the `counter` property.

Good catch, thanks. So seems like this PR is inherently non-doable because of double-borrows. I think I'll just close it.

> I think the `Gc` indirection around native data has to stay, so that you can hold a reference to it without locking the underlying `ScriptObject`.
> 
> However, what would be possible is something like this:
> 
> ```rust
> #[derive(Clone, Debug, Collect)]
> #[collect(require_static)]
> struct BlurFilterData {
>     blur_x: Cell<f64>,
>     blur_y: Cell<f64>,
>     quality: Cell<i32>,
> }
> 
> #[derive(Copy, Clone, Debug, Default, Collect)]
> #[collect(no_drop)]
> #[repr(transparent)]
> pub struct BlurFilter<'gc>(Gc<'gc, BlurFilterData>);
> ```

I'm not so sure about wrapping each field with a `Cell`, since it's only possible for `Copy` fields. When we'll get to port more complex `NativeObject`s (e.g. `ConvolutionFilter`, `GradientBevelFilter`, `GradientGlowFilter`), this wouldn't be possible. I think it will be best to just stay with `GcCell` like done currently.

Issue #10573, opened 2023-04-05
avm1: Some properties seem to only need a setter


Toad06:
This affects at least the `MovieClip.focusEnabled` property. Consider the following sample [(SWF available here)](https://github.com/ruffle-rs/ruffle/files/11163341/focusEnabled.zip):

```
_root.createEmptyMovieClip("mc", _root.getNextHighestDepth());

trace("mc.focusEnabled: " + mc.focusEnabled);
trace("Selection.setFocus(mc): " + Selection.setFocus(mc));
trace("Selection.getFocus(): " + Selection.getFocus());

mc.focusEnabled = "whatever";
trace("mc.focusEnabled = 'whatever'");

trace("mc.focusEnabled: " + mc.focusEnabled);
trace("Selection.setFocus(mc): " + Selection.setFocus(mc));
trace("Selection.getFocus(): " + Selection.getFocus());

delete mc.focusEnabled;
trace("delete mc.focusEnabled");

trace("mc.focusEnabled: " + mc.focusEnabled);
trace("Selection.getFocus(): " + Selection.getFocus());
```

In Flash Player, this traces the following data:

```
mc.focusEnabled: undefined
Selection.setFocus(mc): false
Selection.getFocus(): null

mc.focusEnabled = 'whatever'

mc.focusEnabled: whatever
Selection.setFocus(mc): true
Selection.getFocus(): _level0.mc

delete mc.focusEnabled

mc.focusEnabled: undefined
Selection.getFocus(): _level0.mc
```

---

So it appears that the `focusEnabled` getter always returns the value set by the script (deleting the property also works), like any classic properties (using `delete` also works). In Ruffle, we assume this value to always be a boolean:
https://github.com/ruffle-rs/ruffle/blob/7a6e6cf21475225dc6066381eaf1404d43d4d726/core/src/avm1/globals/movie_clip.rs#L1429-L1434

The `focusEnabled` setter has an impact on `Selection.setFocus`, as Ruffle already implements:
https://github.com/ruffle-rs/ruffle/blob/7a6e6cf21475225dc6066381eaf1404d43d4d726/core/src/avm1/globals/movie_clip.rs#L1436-L1446
https://github.com/ruffle-rs/ruffle/blob/7a6e6cf21475225dc6066381eaf1404d43d4d726/core/src/avm1/globals/selection.rs#L143-L146

---

However, only mentioning a setter in `property()` doesn't seem allowed:
[...]

PR #10597, opened 2023-04-07
avm1: Correct `MovieClip.focusEnabled`

### LINKED ISSUE #2514: web: Governor of Poker - Game Pauses all the time
Hey,

this is also an AS2 game which could possibly work fine. I always enjoyed the game and played hours over hours. Sadly the game pauses all the time. Maybe you can fix it soon 👍 

https://www.newgrounds.com/portal/view/445178?emulate=flash

Toad06:
- `focusEnabled` is not a property of MovieClip, as @Herschel mentioned in https://github.com/ruffle-rs/ruffle/issues/10573#issuecomment-1498500673

- `is_focusable` returns true if the clip is in button mode (this is [mentioned here](https://open-flash.github.io/mirrors/as2-language-reference/MovieClip.html#focusEnabled) in the documentation). This behavior has the priority over `focusEnabled`. Fixes #2514.


Herschel (approved):
Thanks! I tested to see if this should call the property in case of something like `clip.addProperty("focusEnabled", getFocusEnabled, null)`, and indeed the property function does get called.

Toad06:
Thank you for the review. 😉

PR #10602, opened 2023-04-08
web: When building the extension, provide the GitHub token to GitHub CLI

kmeisthax:
This may or may not work depending on how the underlying Action deals with environment variables, or doesn't.


danielhjacobs:
https://stackoverflow.com/questions/2021982/awk-without-printing-newline for the latest failed action. The newline is printed in the awk output right now.

PR #10609, opened 2023-04-08
chore: Select "desktop" as the default workspace member

torokati44:
This makes the cargo command to run the player simpler:
`$ cargo run stuff.swf`

Instead of:
`$ cargo run --bin ruffle_desktop -- stuff.swf`

See: https://github.com/rust-lang/cargo/issues/7290

### DIFF
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -80,7 +80,7 @@ jobs:
           RUSTDOCFLAGS: -D warnings
 
       - name: Run tests with image tests
-        run: cargo test --locked --features imgtests
+        run: cargo test --all --locked --features imgtests
         env:
           XDG_RUNTIME_DIR: '' # dummy value, just to silence warnings about it missing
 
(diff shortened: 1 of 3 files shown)


torokati44:
The downside is that `cargo test` now needs to be invoked with `--all` added...

Herschel:
> The downside is that `cargo test` now needs to be invoked with `--all` added...

I was in the habit of always running with `--all` anyway, because usually my terminal is in some subfolder of the workspace.

Herschel:
Do we need to update any docs to make sure `cargo test --all` is spelled out?

Herschel (approved):
Looks like only CONTRIBUTING.md mentioned it, I tossed an `--all` in there for good measure. Thank you!

torokati44:
> Looks like only CONTRIBUTING.md mentioned it, I tossed an `--all` in there for good measure.

Yeah, that's the only relevant piece of doc' that I found too. Thanks for fixing it up! (I was doing Easter Family Stuff...)

Issue #10474, opened 2023-03-31
Mix Sudoku Light Vol 2 - Error Thrown on loading


desertking:
### Describe the bug

Some weird error. Maybe useful I post it here.

ruffle-player.js:419 Loading SWF file https://www.spiele-umsonst.de/azad/downloads/flashgames/mixsudokulightvol2.swf
ruffle_web-wasm_extensions.js:677 ERROR core/src/display_object/movie_clip.rs:2132 Got "RustError("Attempted to call \"\" with 2 arguments (more than 0 is prohibited)")" when constructing AVM2 side of movie clip of type com.conceptis.gen.loader.light::genLightPreloader

### Expected behavior

To load and run

### Affected platform

Online demo

### Operating system

All

### Browser

All

n0samu:
Progressed by #10299. The next error is:
```
ERROR run_frame:run_all_phases_avm2: ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x19af7725b20 }) to handler FunctionObject(FunctionObject { ptr: 0x19af6ff9760 }) : TypeError: Error #1088: The markup in the document following the root element must be well-formed.
        at XML/XML::init()
        at XML()
        at XML/XML()
        at com.conceptis.gen.loader.light::genLightInitializer/formatPuzzleName()
        at com.conceptis.gen.loader.light::genLightInitializer/initializeStep()
```

evilpie:
We are missing an implementation of the E4X "The Abstract Equality Comparison Algorithm" for XML/XMLList.

evilpie:
```actionscript
var a = new XML('<family><genealogy code="02" /></family>');
var b = new XML('<xxx family="02"/>');
trace("a.genealogy.@code", a.genealogy.@code);
trace("b.@code", b.@family);
trace("a.genealogy.@code == b.@family", a.genealogy.@code == b.@family); // Equal in Flash
```

evilpie:
The error now is 
```
ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "complete", class: flash.events::Event, ptr: 0x56367a718190 }) to handler FunctionObject(FunctionObject { ptr: 0x56367a8dc820 }) : Error: XML node 0 1 2 3 4 5 6 7 8\type is missing or is not in the correct format
```

Sadly it looks like the code is auto generated or hidden somehow?

desertking:
[...]

Issue #10183, opened 2023-03-17
Tilt: Cannot start game


n0samu:
### Describe the bug

Download SWF file: [tilt_decrypted.zip](https://github.com/ruffle-rs/ruffle/files/11006092/tilt_decrypted.zip)

After the intro splash screens, [Tilt](http://www.notdoppler.com/tilt-b.php) shows only a blank screen.

This error is repeated several times:
```
ERROR ruffle_core::avm2::activation: ReferenceError: Error #1069: Property text not found on String and there is no default value.
        at org.ascollada.core::ColladaElement$/parseStringArray()
        at org.papervision3d.objects.parsers::DAE/parseNode()
        at org.papervision3d.objects.parsers::DAE/parseVisualScene()
        at org.papervision3d.objects.parsers::DAE/parseScene()
        at org.papervision3d.objects.parsers::DAE/parseAfterMaterials()
        at org.papervision3d.objects.parsers::DAE/parse()
        at org.papervision3d.objects.parsers::DAE/load()
        at com.devm.puzzle.level::Level3D/init()
        at com.devm.puzzle::Game/init()
        at com.devm.puzzle::Main/initGame()
```

### Expected behavior

The main menu should load.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

Here is the relevant AS function:
```as
public static function parseStringArray(param1:XML) : Array
{
    var _loc2_:String = param1.text().toString();
    _loc2_ = _loc2_.replace(/^\s+/,"");
    _loc2_ = _loc2_.replace(/\s+$/,"");
    return _loc2_.split(/\s+/);
}
```
The error message does not make sense because it implies that Ruffle is trying to get the `text` of a String, even though it should be trying to get the text of an XML.

adrian17:
This could be a bug with https://github.com/ruffle-rs/ruffle/pull/9806

> When an XML object has simple content, you can call non-XML methods directly on it - it will internally be stringified, and the method will be called on the resulting string

This does sound like what's happening here, but not sure why it's not finding our `name()`.

adrian17:
Oh, well this is simply because we don't have `text()` implemented.

I think we should just stub all the methods (even have them throw exceptions instead of no-op), just to prevent the misleading fallback-to-string scenario.

evilpie:
This works now, but there is some text missing in the main menu.

Issue #10390, opened 2023-03-26
Army of Ages: copy is not a function of class XML


MartySVK:
### Describe the bug

The game is currently stuck on an error:
``
copy is not a function of class XML
``

SWF - https://data.gameflare.com/games/1884/6tkKc7XgIkdXg8.swf

### Expected behavior

progress the game

### Affected platform

Desktop app

### Operating system

Windows 10

evilpie:
I've got this working with #10412 and #10413 and some additional namespace related patch.

MartySVK:
The current error is:
``
ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "click", class: flash.events::MouseEvent, ptr: 0x2848db48a20 }) to handler FunctionObject(FunctionObject { ptr: 0x28488a80210 }) : RustError("Can not set non-public name yet")
``

MartySVK:
> #10385 makes it work

nice, game looks playable with this PR

PR #10613, opened 2023-04-08
Partially implement XML/XMLList abstract equality

evilpie:
Helps #10474 and  #10183. Quite likely also others, previously the `==` comparison would return wrong results which is often not an obvious failure.


core/src/avm2/value.rs
```diff
@@ -1138,6 +1138,31 @@ impl<'gc> Value<'gc> {
         other: &Value<'gc>,
         activation: &mut Activation<'_, 'gc>,
     ) -> Result<bool, Error<'gc>> {
+        // ECMA-357 extends the abstract equality algorithm with steps
+        // for XML and XMLList types. Because they are objects in Ruffle we
+        // have to do be a bit more complicated and factor out the code into
```
Herschel, line 1143: typo: `do be` -> `be`

n0samu:
Can confirm this makes #10183 playable! 🎉

core/src/avm2/value.rs
evilpie: Thanks!

Herschel (approved):
Thank you!

Lord-McSweeney:
Unfortunately this seems to have regressed #10390. The menu doesn't load anymore.

Issue #3144, opened 2021-02-07
Errors on AS2 Rollovers


brain-death:
https://astrologywizard.com/signs.htm

**Expected Behaviour**

Sign symbols change colour (I think) and play a sound on rollover.

**Describe The Bug**

Rollovers don't work. Sound doesn't play and colours only visible sporadically, else everything is grey.

All sounds appear to play on page load. Links work though...

**Platform Etc...**

Arch Linux. Firefox extension.

n0samu:
This problem occurs because Ruffle is not firing rollout events on objects within the movie when the mouse leaves the content area. Thus the button stays in the rollover state all the time instead of toggling between these states when the mouse enters/leaves the content area.

torokati44:
I think this is the same fundamental issue as https://github.com/ruffle-rs/ruffle/issues/7113 .

Lord-McSweeney:
Still occurring.

Lord-McSweeney:
Fixed by #10622.

Issue #1745, opened 2020-11-27
Differences with Flash Player in mouse interactions


Toad06:
Web and desktop backends are both affected.
Nothing critical though.

- When the cursor is over a button and you move the mouse outside of the canvas (or game window) quickly enough, then the button remains as if the cursor was still over it.
=> In [Mini-Putt 2](https://www.newgrounds.com/portal/view/38598), click "Practice", then check the issue with the "Continue" button.

- (FIXED) When you click a button (mouse down only) and release it outside of the button area (can also be outside of the canvas), then the action of the button is executed.
=> [Mini-Putt 2](https://www.newgrounds.com/portal/view/38598) can also be used here; or in [Luigi's Day](https://www.newgrounds.com/portal/view/44505), check the issue with the "Play" or "Controls" button.

- (FIXED) When you click a button, the cursor may not go back to the default one; on affected files, it would usually keep displaying the pointer cursor.
=> In [Bubble Gum Run](https://www.y8.com/games/bubble_gum_run), check the issue with the second "Play" button.

Herschel:
As of 10/2021, these all seem fixed except the third cursor-not-resetting issue.

relrelb:
Did #6262 solve this?

Toad06:
Not completely, the first item is still an issue but it only happens if you don't hold down any mouse button. If you press the left button and release it outside of the canvas, then the state of the "Continue" button will change as expected.

Toad06:
First item fixed by #10622.

Issue #10188, opened 2023-03-17
Insulin Dog banner running forever (dog does not sit)


rredford:
### Describe the bug

In picture, 2 banners is shown, one in ruffle and one in flash. As you can see, dog is correctly sitting when mouse is outside window. Ruffle does not do this so dog keeps running forever.

To repeat this:
In zip there is 3 files, run insulin_banner.swf, and wait for info block to go away, then mouse in and out of banner window. In Flash, dog sits after a second of mouse not inside window area. All 3 files is required as it wont function properly otherwise.

![image](https://user-images.githubusercontent.com/151014/226070340-a8d1de3e-4971-417f-bcab-b621c0c9ec07.png)

zip
[dogbanner.zip](https://github.com/ruffle-rs/ruffle/files/11006994/dogbanner.zip)

### Expected behavior

Doggy sits when mouse is outside window.

### Affected platform

Desktop app

### Operating system

xfce mint 21.1

SN902:
dog is correctly sitting when mouse on info button in left corner of the screen, pretty sure it's input issue

n0samu:
Same problem as #3144

Lord-McSweeney:
Issue still occurring (at least on desktop).

Lord-McSweeney:
Fixed by #10622.

Issue #10173, opened 2023-03-17
Errors on http://new.weedtowonder.org/introduction.html


danielhjacobs:
### Describe the bug

1) Text does not appear - Relates to the error ERROR core/src/avm2/activation.rs:953 ReferenceError: Error #1069: Property http://adobe.com/AS3/2006/builtin::settings not found on Class and there is no default value. (XML)
![Screenshot from 2023-03-17 11-18-09](https://user-images.githubusercontent.com/29206584/225946570-2113ce79-2ef7-49e3-bed4-4cdefb4f4d03.png)
2) Clicking a video name full-screens the content (it's supposed to play an FLV Video, and although those are obviously unsupported, it should do what http://new.weedtowonder.org/domestication.html when clicking a video name and show the FLV symbol, not full screen the content). Error messages when clicking the video name are:
    i) AVM2 error: Object(ErrorObject(ErrorObject { class: TypeError, ptr: 0x1489600 }))
    ii) TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: isWaiting)
    iii) The previous 2 errors 2 more times
    iv) Error dispatching event EventObject(EventObject { type: "click", class: flash.events::MouseEvent, ptr: 0x12dced8 }) to handler FunctionObject(FunctionObject { ptr: 0x8d7de8 }) : AvmError(Object(ErrorObject(ErrorObject { class: TypeError, ptr: 0x1489600 })))
[Screencast from 03-17-2023 11:19:03 AM.webm](https://user-images.githubusercontent.com/29206584/225946799-fcbb8d20-5981-4099-8ef2-d110541102d4.webm)
3) A picture of corn does not appear where the videos display, and a map of maize, wheat, and rice does not appear in the top-left of the content area. The map does load according to the network tab, but it is not displayed.
![Screenshot from 2023-03-17 11-20-19](https://user-images.githubusercontent.com/29206584/225947096-bb70d662-ed6d-4305-be09-cba8f0e2b152.png)
4) The "Jump to ..." menu bounces up and down a bit as you scroll between different options in the menu.
[Screencast from 03-17-2023 11:21:14 AM.webm](https://user-images.githubusercontent.com/29206584/225947353-250a1d83-cf36-47ea-b43d-5f2eac944e3b.webm)
5) As is completely expected, FLV videos do not work.

### Expected behavior

1/3. Text and images should appear:
[...]

PR #10622, opened 2023-04-09
core: Detect when the mouse moves out of the stage

### LINKED ISSUE #7113: AVM2: Mouse pointer leaving the window/canvas not handled properly
For example with https://z0r.de/L/z0r-de_7223.swf, when moving the mouse out of the window/canvas, the text on the top should always disappear.
However, if I move the mouse quickly, the text stays on. This does not happen with Flash Player.
Both the desktop application (tested on Linux/Wayland) and the web build (tested on the demo page in Firefox) are affected.
Just make sure not to do it on an edge with any black letterboxing, as that makes it work properly.

Toad06:
This PR targets both of the web and desktop apps.
- On web, this adds a new `pointerleave` event listener.
- On desktop, something similar was already implemented:
https://github.com/ruffle-rs/ruffle/blob/0643426b84e3658caae98e14f88509049e50ac49/desktop/src/main.rs#L515-L521

Fixes the first item in #1745 and closes the issue. Also fixes #7113.


n0samu:
Also fixes #3144 and #10188! 🎉

torokati44:
I'm so glad #7113 is being addressed, thanks! :relieved:

Just some nits:
- Isn't the `mouse_moved_out` name a bit "negative" (as opposed to "affirmative") [^1]? Perhaps something like `mouse_is_in` (but better) could be considered? It's not important though...
- Are we certain that `pointerenter` (and the corresponding on desktop) events don't need to be handled? It makes sense that the first move/click event will do what the enter event would do anyway, just a consideration for symmetry's sake...

[^1]: https://softwareengineering.stackexchange.com/questions/196830/boolean-method-naming-affirmative-vs-negative

n0samu:
Along those lines, here is a simple AS3 project with a movie clip that takes up the whole screen, with a rollover listener attached: [rollOverTest.zip](https://github.com/ruffle-rs/ruffle/files/11186763/rollOverTest.zip)
```as
function rollFn(evt:MouseEvent) {
	trace(evt);
}
mc.addEventListener(MouseEvent.ROLL_OVER, rollFn);
```
In Ruffle, moving the mouse into the Ruffle window does not trigger the trace statement (tested in the desktop app).

Toad06:
> Isn't the `mouse_moved_out` name a bit "negative" (as opposed to "affirmative")? Perhaps something like `mouse_is_in` (but better) could be considered?

This was an interesting read, thanks. `mouse_moved_out` also sounds too related to this PR. I renamed it to `mouse_in_stage`.

> Are we certain that `pointerenter` (and the corresponding on desktop) events don't need to be handled? It makes sense that the first move/click event will do what the enter event would do anyway, just a consideration for symmetry's sake...

From the ActionScript perspective, I don't think we need it but I could still try to change it.

> here is a simple AS3 project with a movie clip that takes up the whole screen, with a rollover listener attached

Your sample should work with this PR but there are some differences with Flash when the file launches:
- Ruffle always triggers the rollOver event (even if the mouse is outside) and reports `stageX` and `stageY` to be zero (same behavior as before this PR).
- The Flash projector does the same thing but reports `stageX` and `stageY` with values of the mouse position clamped to 0 and the width/height of the SWF.
- The Flash plugin doesn't log anything until a mouse interaction happens (this can be a click if the mouse hovered the SWF when it loaded).

danielhjacobs:
I haven't tested this PR with it, but #10173, number 4 may relate to either this or what nosamu mentioned.

Toad06:
@torokati44 I tried adding the `pointerenter` / `CursorEntered` event in another branch: https://github.com/ruffle-rs/ruffle/commit/d57748bdfe5b637e2977ddae9c2718da7551a222
Do you think it's better like this?

core/src/player.rs
```diff
@@ -1156,6 +1165,12 @@ impl Player {
                 }
             });
         }
+
+        if let PlayerEvent::MouseLeave = event {
```
torokati44, line 1169: Is there a reason why this didn't get a "MouseEnter" counterpart for symmetry?

Toad06: Sorry, late reply. `MouseLeave` was already present in `events.rs` and there's indeed no `MouseEnter`. The reason, [according to this answer](https://github.com/ruffle-rs/ruffle/pull/6006#discussion_r780883658), is that AS3 only has an [`Event.MOUSE_LEAVE`](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/events/Event.html#MOUSE_LEAVE).

torokati44: Ah, I see. Strange!

torokati44 (approved):

Dinnerbone (approved):

Toad06:
Thank you all for the reviews and feedbacks.

PR #10638, opened 2023-04-11
avm2: Implement XML.hasSimpleContent and hasComplexContent

n0samu:
Progresses [Jet Velocity](https://www.newgrounds.com/portal/view/553350/format/flash?emulate=flash)


core/src/avm2/globals/XML.as
```diff
@@ -44,6 +45,19 @@ package {
         AS3 native function text():XMLList;
         AS3 native function toString():String;
 
+        AS3 function hasComplexContent():Boolean {
+            return !this.hasSimpleContent();
+        }
+
+        prototype.hasComplexContent = function():Boolean {
+            return !this.hasSimpleContent();
```
Lord-McSweeney, line 53: I think it might be better if this uses an `AS3::` for consistency, but if it works this way, it's fine.

core/src/avm2/globals/XML.as
```diff
@@ -44,6 +45,19 @@ package {
         AS3 native function text():XMLList;
         AS3 native function toString():String;
 
+        AS3 function hasComplexContent():Boolean {
+            return !this.hasSimpleContent();
```
evilpie, line 49: This is not correct. hasComplexContent isn't simply the inverse of hasSimpleContent. (i.e. comments are neither complex nor simple).

See [ECMA-357](https://www.ecma-international.org/publications-and-standards/standards/ecma-357/)

evilpie (changes requested):

n0samu:
Okay I finally got around to implementing this properly. Please let me know how it looks now!

evilpie:
Looks good to me now, thanks!

core/src/avm2/e4x.rs
```diff
@@ -540,6 +540,18 @@ impl<'gc> E4XNode<'gc> {
         }
     }
 
+    pub fn has_complex_content(&self) -> bool {
+        match &self.0.read().kind {
+            E4XNodeKind::Element { children, .. } => children
+                .iter()
+                .any(|child| matches!(&*child.kind(), E4XNodeKind::Element { .. })),
+            E4XNodeKind::Text(_) | E4XNodeKind::CData(_) => false,
```
Aaron1011, line 548: Can you combine all of these match arms into one?

n0samu: `cargo fmt` makes that look kinda ugly IMO
```
            E4XNodeKind::Text(_)
            | E4XNodeKind::CData(_)
            | E4XNodeKind::Attribute(_)
            | E4XNodeKind::Comment(_)
            | E4XNodeKind::ProcessingInstruction(_) => false,
```
So given that, I think I'll leave it as-is. Definitely a good suggestion though!

Aaron1011:
Can you add a test for this?

n0samu:
Seems like we didn't need one this whole time for hasSimpleContent, plus I don't know much about writing XML... it would be a lot more work for me than the PR itself. So I don't think so, sorry

iwannabethedev:
I really think there should be a test for this, it is for multiple reasons.

Though, I ended up writing a possible core of a test for you.

Do be aware that I do not understand XML in ActionScript very well, and I do not understand XMLList.

```actionscript

var xmlAndXmlLists: Array = [
    (<a></a>).b,
    (<a><b/></a>).b,
    (<a><b x="3"/></a>).b,
    (<a><b/><b/></a>).b,
    (<a><b/>text<b/></a>).b,
    (<a><b><c/></b></a>).b,
    (<a><b><c/></b><b/></a>).b,
    (<a><b><c/><c/></b></a>).b,
    <a>text</a>,
    <a href="a"/>,
    <?instruction ?>,
    <a><?instruction ?></a>,
    <!-- comment -->,
    <a><!-- comment --></a>,
    <![CDATA[My
    Multiline
    CDATA
    ]]>,
    <a><![CDATA[My
    Multiline
    CDATA
    ]]></a>,
    
];

var i: int;
for (i = 0; i < xmlAndXmlLists.length; i += 1) {
    trace("");
    var which: String = "xmlAndXmlLists[" + i + "]";
    trace(which + ": " + xmlAndXmlLists[i].toXMLString());
    trace(which + ".hasSimpleContent(): " + xmlAndXmlLists[i].hasSimpleContent());
    trace(which + ".hasComplexContent(): " + xmlAndXmlLists[i].hasComplexContent());
    trace("");
}
```

Aaron1011 (approved):

PR #10648, opened 2023-04-11
core: Add frame rate config option

### LINKED ISSUE #3276: Feature Request: Ability to specify a framerate that overwrites the value in the SWF.
It has come up that some old games run too fast in Ruffle because the original developers had never been able to run them at the framerate the chose during development.

A handy option could be the ability to pass a framerate value to Ruffle, which would override the frame-rate specified in the original SWF.

n0samu:
Implements the feature described in and closes #3276.
Provides an easy workaround for #9325.

### DIFF
--- a/desktop/src/main.rs
+++ b/desktop/src/main.rs
@@ -150,6 +150,10 @@ struct Opt {
     /// The version of the player to emulate
     #[clap(long)]
     player_version: Option<u8>,
+
+    /// Set and lock the player's frame rate, overriding the movie's frame rate.
+    #[clap(long)]
+    frame_rate: Option<f64>,
 }
 
 #[cfg(feature = "render_trace")]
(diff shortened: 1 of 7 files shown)


web/packages/core/src/load-options.ts
```diff
@@ -300,6 +300,13 @@ export interface BaseLoadOptions {
      */
     forceScale?: boolean;
 
+    /**
+     * Sets the player's frame rate, overriding the frame rate set by the movie.
```
Dinnerbone, line 304: Can we make this say "Sets and locks ..." just to make it clear this makes it static and not dynamic?

desktop/src/main.rs
```diff
@@ -150,6 +150,10 @@ struct Opt {
     /// The version of the player to emulate
     #[clap(long)]
     player_version: Option<u8>,
+
+    /// Override the movie's default frame rate.
```
Dinnerbone, line 154: Here too, maybe add "and prevents the movie from changing it"

Dinnerbone (approved):
LGTM other than trivial doc nits

PR #10655, opened 2023-04-11
String interning MVP

moulins:
This PR builds on #10572.

- Adds a global string interner based on a custom, GC-aware `WeakSet`;
- Reworks the internal `AvmString` representation to distinguish between interned and non-interned strings, enabling faster equality checks for interned strings;
- Adds the `AvmAtom` type to represent guaranteed-interned `AvmString`s;
- Intern all strings loaded from the AVM1 & the AVM2 constant pools;
- Intern all AVM2 namespaces strings (but not the namespaces themselves, nor the strings in `QName`s or `Multiname`s);
- Intern all 'built-in' strings used in AVM1 globals initialization.

To facilitate these changes, I've added a new minimal `GcContext` struct (bundling the `MutationContext` with the interner), which can be used where passing a full `UpdateContext` would be impossible or undesirable.

-------

Performance-wise, I'm seeing ~10% FPS improvements on AVM2-heavy SWFs (like the box2d demo); unfortunately I don't have any good AVM1 benchmark so there's no concrete numbers, but I expect roughly similar figures.

-------

Possible work for future PRs:
- (AVM2) Optimize namespaces by using pointer tagging to encode the kind of namespace without an extra allocation (needs `Gc<T> -> *mut T` roundtrip support from `gc-arena`);
- Remove the need for a separate static `AvmString` variant (would also need `Gc<T> -> *mut T` roundtrip support);
- (AVM1) Avoid re-parsing `SwfStr`s into `AvmString`s on each action execution, maybe by keeping some kind of per-SWF string cache?
- Only store `AvmAtom`s for AVM1 & AVM2 properties; this would allow using the `AvmAtom` address directly as the map key and completely skip string hashing and comparisons in most cases (I've tried doing it in this PR, but had to bail out as the required changes were quickly growing out of control, and this PR is already big enough...);
- (AVM1) Add a way to intern integers as strings directly, to avoid transient string allocations on array accesses;
- etc...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1611,15 +1611,16 @@ dependencies = [
 [[package]]
 name = "gc-arena"
 version = "0.2.2"
-source = "git+https://github.com/kyren/gc-arena?rev=1a6310c0d5c98836fa9efb1c4773038ecfd5a92e#1a6310c0d5c98836fa9efb1c4773038ecfd5a92e"
+source = "git+https://github.com/kyren/gc-arena?rev=fcc8764362d25f8724912dd7f09f2405779ec053#fcc8764362d25f8724912dd7f09f2405779ec053"
 dependencies = [
  "gc-arena-derive",
+ "sptr",
 ]
 
 [[package]]
 name = "gc-arena-derive"
 version = "0.2.2"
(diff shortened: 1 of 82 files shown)


Dinnerbone (approved):
LGTM

moulins:
Sure, I'll take a look at fixing the conflicts this evening.

Given that this is based on #10572 (which doesn't seem to have any conflicts), maybe it's best to merge that PR beforehand? I don't remember if GitHub automatically closes existing PRs when merging a 'derived' PR such as this one.

Issue #10640, opened 2023-04-11
Crash on stacko.swf


n0samu:
### Describe what you were doing

Link to SWF: https://static.arcadespot.com/wp-content/uploads/2015/11/stacko.swf

The problem occurs after clicking the Play button on the main menu. This game used to work so this is a regression.

### What does the crash message say?

panicked at 'wgpu error: Validation Error

Caused by:
    In a RenderPass
      note: encoder = `<CommandBuffer-(2, 1317, Vulkan)>`
    In a pass parameter
      note: command buffer = `<CommandBuffer-(2, 1317, Vulkan)>`
    all attachments must have the same sample count, found 1 != 4

', C:\Users\runneradmin\.cargo\registry\src\index.crates.io-6f17d22bba15001f\wgpu-0.15.1\src\backend\direct.rs:3024:5

### Ruffle Version

0.1.0-nightly (81d907b 2023-04-09)

### Affected platform

Desktop app

### Operating system

Windows 10.0.19045 (Windows 10 Pro) [64-bit]

### Additional information

Filename: stacko.swf

### AVM2 Callstack
```
	at flash.display3D::Context3D/flash::display3D::Context3D::drawTriangles()
	at starling.display::QuadBatch/renderCustom()
	at starling.core::RenderSupport/finishQuadBatch()
	at starling.textures::RenderTexture/renderBundled()
	at starling.textures::RenderTexture/draw()
	at nepix.ui.panels.levels::LockedOverlay/show()
	at nepix.ui.panels.levels::LockedOverlay()
	at nepix.ui.panels.levels::LevelPack/initialize()
	at feathers.core::FeathersControl/initializeInternal()
	at feathers.core::FeathersControl/validate()
	at nepix.ui.panels.levels::LevelPack/validate()
	at feathers.layout::HorizontalLayout/validateItems()
	at feathers.layout::HorizontalLayout/layout()
	at feathers.controls::LayoutGroup/draw()
	at feathers.controls.supportClasses::LayoutViewPort/draw()
	at feathers.core::FeathersControl/validate()
	at feathers.controls::Scroller/refreshViewPortBoundsWithoutFixedScrollBars()
	at feathers.controls::Scroller/draw()
	at feathers.controls::ScrollContainer/draw()
	at feathers.core::FeathersControl/validate()
	at screens::LevelsScreen/initialize()
	at feathers.core::FeathersControl/initializeInternal()
[...]

PR #10657, opened 2023-04-12
wgpu: Fix MSAA and depth buffer handling in SetRenderToTexture

Aaron1011:
When multisampling is enabled, we should create a new multisampled texture, and use the existing texture as the resolve buffer. We also need to call `update_has_depth_texture` to keep our pipeline aware of whether or not we currently have a depth buffer attached.

Makes progress on #10641 (it has a stack overflow after this PR, due to an unrelated issue).


Dinnerbone (approved):
LGTM but is it possible to test this?

n0samu:
Also fixes #10640!

Aaron1011:
@Dinnerbone I've modified an existing test to use MSAA

Dinnerbone (approved):

Dinnerbone:
Thanks! You'll need to up the tolerance and outliers a bit to account for MSAA across devices now, sorry!

PR #10678, opened 2023-04-13
web: Use div element for save manager to support browsers without dialog

danielhjacobs:


Korne127:
This works and fixes the issue on Safari 14 :)

Korne127:
This works and fixes the issue on Safari 14 (macOS 11). :)

n0samu (approved):
Can confirm that this fixes the save manager for me on iOS Safari, thank you!

relrelb (approved):

PR #10691, opened 2023-04-14
web: No context menu on iOS if finger moves too far on long-press

danielhjacobs:


n0samu:
Unfortunately, this implementation is conceptually wrong because the context menu is activated if the long press timer is *absent*, not present. Clearing the timer tells Ruffle to activate the context menu on pointerup, which is exactly the opposite of what you want to be doing here. I've opened #10710 to supersede this.

PR #10709, opened 2023-04-16
avm2: stub xml.normalize()

AllinolCP:


n0samu:
Shouldn't this also be added to the prototype, like the other XML functions?

AllinolCP:
> Shouldn't this also be added to the prototype, like the other XML functions?

Right

Aaron1011 (approved):

PR #10710, opened 2023-04-16
web: No context menu on iOS if finger moves too far on long-press

n0samu:
Supersedes #10691.

### DIFF
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -104,6 +104,16 @@ function sanitizeParameters(
     return output;
 }
 
+class Point {
+    constructor(public x: number, public y: number) {}
+
+    distanceTo(other: Point) {
+        const dx = other.x - this.x;
+        const dy = other.y - this.y;
+        return Math.sqrt(dx * dx + dy * dy);
+    }
+}
+
 /**
  * The ruffle player element that should be inserted onto the page.
  *
@@ -148,6 +158,8 @@ export class RufflePlayer extends HTMLElement {
 
     private isExtension = false;
     private longPressTimer: ReturnType<typeof setTimeout> | null = null;
+    private pointerDownPosition: Point | null = null;
+    private pointerMoveMaxDistance = 0;
 
     /**
      * Triggered when a movie metadata has been loaded (such as movie width and height).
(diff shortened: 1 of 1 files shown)


web/packages/core/src/ruffle-player.ts
```diff
@@ -816,6 +822,21 @@ export class RufflePlayer extends HTMLElement {
...
+    /**
+     * Helper function to get the distance between two points
+     *
+     * @param x1 The x coordinate of point 1.
+     * @param y1 The y coordinate of point 1.
+     * @param x2 The x coordinate of point 2.
+     * @param y2 The y coordinate of point 2.
+     * @returns The distance between the two points.
+     */
+    getDistance(x1: number, y1: number, x2: number, y2: number): number {
```
relrelb, line 834: The logic looks good in general, thank you!

One nitpick - I think the more TypeScripty way to handle points would be to introduce:
```ts
class Point {
    constructor(public x: number, public y: number) {}

    distanceTo(other: Point) {
        const dx = other.x - this.x;
        const dy = other.y - this.y;
        return Math.sqrt(dx * dx + dy * dy);
    }
}
```
Then use it in the various places. What do you think?

n0samu: I agree that that's cleaner. Added, thanks!

web/packages/core/src/ruffle-player.ts
```diff
@@ -148,6 +158,8 @@ export class RufflePlayer extends HTMLElement {
 
     private isExtension = false;
     private longPressTimer: ReturnType<typeof setTimeout> | null = null;
+    private pointerDownPosition = new Point(-1, -1);
```
relrelb, line 161: I don't know if this makes sense, but initalizing this with `null` seems more clear to me. Then a non-null assertion will be needed in `checkLongPressMovement`.

n0samu: Ok, updated

relrelb (approved):

web/packages/core/src/ruffle-player.ts
n0samu: ... and resolved linter's complaint, oops

PR #10720, opened 2023-04-17
I18n: remove extra whitespace from JSON

yangyangdaji:
reduce file size
Signed-off-by: yangyangdaji <1504305527@qq.com>


adrian17:
I don't think there's any inherent advantage of 2-space over 4-space indent in this particular case (and we generally use 4 in most files), so I'm confused about reasoning for this PR.

(also some of the files are unevenly indented now - 4/2 space indents)

yangyangdaji:
I think that deleting excess space can reduce the file size 
 Up to you

Issue #9735, opened 2023-02-26
Fault Line: Visual Issues


n0samu:
### Describe the bug

Link to SWF: http://i.notdoppler.com/files/faultline.swf

In [Fault Line](http://www.notdoppler.com/faultline.php), after the folding the platforms, parts of them become invisible:
![image](https://user-images.githubusercontent.com/71368227/221389599-403d3b85-600c-4def-87ca-774e2ba991f8.png)

This is due to `beginBitmapFill` being unimplemented:
`WARN run_frame:run_all_phases_avm2: ruffle_core::stub: Encountered stub: AVM2 flash.display.Graphics.beginBitmapFill()`

Also, as pointed out in https://github.com/ruffle-rs/ruffle/issues/3794#issuecomment-1445229591, text is flickering in some places, for example:
![ruffle_flicker](https://user-images.githubusercontent.com/71368227/221390105-fdd569d2-af4b-4859-a768-f34431a4456c.gif)
(Note: it flickers much faster than it appears in the GIF)

### Expected behavior

Here is what the folded platform is supposed to look like:
![image](https://user-images.githubusercontent.com/71368227/221390154-d1c11d9d-eb6a-465c-b155-2442877f3d02.png)

Also, the text should not flicker.

### Affected platform

Desktop app

### Operating system

Windows 10

Lord-McSweeney:
Currently, it looks like there's been an error with the `beginBitmapFill` implementation:
![chrome-extension___donbcfbmhbcapadipfkeojnmajbakjdc_player html_url=http%3A%2F%2Fi notdoppler com%2Ffiles%2Ffaultline swf](https://user-images.githubusercontent.com/84632019/225806972-74cc2de8-8863-4a31-8568-ca507b9b2909.png)
(too zoomed in)
Also, parts of the level occasionally flicker and disappear before reappearing.

Lord-McSweeney:
Folding was fixed by #10630.

n0samu:
Fixed by #10734

PR #10734, opened 2023-04-18
avm2: Stop after first frame when movieclip doesn't extend MovieClip

### LINKED ISSUE #10721: Flickering on Astronomy Simulator (ActionScript3 Warning Message
Astronomy professor here. Using Ruffle to resurrect some excellent online flash simulations for students. Most work great, but one has a persistent flickering issue, which I have observed on: 
- Chrome OS running Chrome
- MacOS running Chrome
- MacOS running Safari
- Windows 10 running Chrome
- Windows 10 running Edge
- Android running Chrome

Ruffle gives the attached warning and the interactive elements in the simulation flicker as shown in the attached image. The simulations does respond to some actions. Here is the simulation: https://agl-initiatives.org/astroUNL/naap/habitablezones/animations/stellarHabitableZone.html

Is there any solution that I can implement? Can anyone point me in a direction that could fix this issue? 

Thanks,
AVH
Screenshot 2023-04-17 at 11 27 01 AM
Screenshot 2023-04-17 at 11 25 44 AM

Edit: Added link to simulation.

Aaron1011:
It's possible to have a DefineSprite tag with multiple frames, but with a corresponding SymbolClass that directly extends `Sprite` (and therefore does *not* extend `MovieClip`). When this happens, Flash Player stops after the first frame.

Fixes #10721


core/src/display_object.rs
```diff
@@ -1391,6 +1391,19 @@ pub trait TDisplayObject<'gc>:
                 }
             }
         }
+
+        if let Some(movie) = self.as_movie_clip() {
+            if let Some(obj) = movie.object2().as_object() {
+                let mut activation = Avm2Activation::from_nothing(context.reborrow());
+                let movieclip_class = activation.avm2().classes().movieclip;
```
adrian17, line 1398: could this be instead done by setting some flag at instantiation time? 

(I'm mostly not a fan of making a new reborrow on possibly-each MC construct phase; one other way to fix it could be to somehow remove `activation` requirement from `is_of_type`; that said, setting some MC flag still sounds like a "nicer" solution in general.)

Aaron1011: I changed it to take `UpdateContext` in https://github.com/ruffle-rs/ruffle/pull/10735

Lord-McSweeney:
Fixes the last issue in #9735! :tada:

SN902:
Fixes flickering in:

[PortageCS5.zip](https://github.com/ruffle-rs/ruffle/files/11268739/PortageCS5.zip)
[Pokemon Trainer Creator.zip](https://github.com/ruffle-rs/ruffle/files/11268748/Pokemon.Trainer.Creator.zip)
[slider1.zip](https://github.com/ruffle-rs/ruffle/files/11268763/slider1.zip)

in all htpot NSFW games (+80 swf's)

tests/tests/swfs/avm2/sprite_with_frames/test.toml
```diff
@@ -0,0 +1,7 @@
+num_frames = 2
+
+[image_comparison]
+tolerance = 0
+
+[player_options]
+with_renderer = { optional = true, sample_count = 1 }
```
Dinnerbone, line 7: This should be optional = false, unless you add some traces to show when it succeeds/fails

Dinnerbone (approved):

adrian17 (approved):
I still think it should ideally be a flag (maybe add a comment?), but I'd say it's not blocking.

PR #10745, opened 2023-04-19
avm1: shrink `Value` by boxing `MovieClipReference`s

moulins:
Brings `Value` from 48 bytes to 24 bytes on 64-bits targets (and similarly for 32-bits).

This doesn't actually cost an extra allocation as we can replace an inner `Gc<Vec<_>>` by a `Box<[_]>` directly.


Dinnerbone (approved):
Would it be worth adding a static assert for the size of Value to make sure this doesn't grow again later?

Issue #10705, opened 2023-04-16
AVM2 DO methods don't support Video DOs


Lord-McSweeney:
### Describe the bug

When passing a Video constructed with the ActionScript constructor to any `DisplayObject` methods, the argument passed will be an `Object(StageObject(StageObject { name: flash.media::Video (or something that extends Video), ptr: ... }))` instead of a `DisplayObject`.

This is the cause of (part of) #8228 and #10173.

### Expected behavior

The functions can be called without erroring out on `RustError("ArgumentError: Child not a valid display object")`.

### Affected platform

Desktop app

### Operating system

Linux

### Additional information

From my limited knowledge of Ruffle, `display_object::Video` should implement a `new_with_avm2` method, which can be called when the AS constructor is called. I tried doing this but ran into issues with the borrow checker.

Lord-McSweeney:
Fixed by #10756.

Issue #10175, opened 2023-03-17
Issues on http://new.weedtowonder.org/domestication.html


danielhjacobs:
### Describe the bug

I will mainly list issues that do not also apply to #10173. Though it's worth noting issues 1, 3, 4, and 5 from there also apply here:
1. When the content loads, it loops through every page before return to and showing the first.
[Screencast from 03-17-2023 11:43:15 AM.webm](https://user-images.githubusercontent.com/29206584/225952645-eb2557a8-33cb-45a9-a232-7cbdfa99d6d6.webm)
2. When clicking one FLV video name, the errors are this:
    i. `Got RustError("ArgumentError: Child not a valid display object") when constructing AVM2 side of movie clip of type fl.video::FLVPlayback`
    ii. `AVM2 error: Object(ErrorObject(ErrorObject { class: ReferenceError, ptr: 0x2c42350 }))`
    iii. `ReferenceError: Error #1069: Property http://adobe.com/AS3/2006/builtin::settings not found on Class and there is no default value`
    iv. `Got AvmError(Object(ErrorObject(ErrorObject { class: ReferenceError, ptr: 0x2c42350 }))) when constructing AVM2 side of movie clip of type ch2_fla::title_29`
[Screencast from 03-17-2023 11:44:53 AM.webm](https://user-images.githubusercontent.com/29206584/225953029-5a863e60-4dfb-42f1-a285-c515b4c688f6.webm)
3. Then, when clicking any more FLV video names, the errors match the ones listed in (2) in #10173, though the content does not go full screen.
[Screencast from 03-17-2023 11:45:25 AM.webm](https://user-images.githubusercontent.com/29206584/225953196-bbc6a6e2-c7af-430a-b7ec-05176dc68ca8.webm)
4. When clicking the 3rd left-hand side circle, the animation on that page loops instead of waiting for arrow clicks.
[Screencast from 03-17-2023 11:46:31 AM.webm](https://user-images.githubusercontent.com/29206584/225953426-4c4aabdc-5911-490a-b673-28d0a9c70301.webm)

### Expected behavior

1. The content is supposed to just start at the first page without looping.
[Screencast from 03-17-2023 11:48:17 AM.webm](https://user-images.githubusercontent.com/29206584/225953814-8821ad5c-9cc6-4caf-b76b-a58ebd224497.webm)
2. and 3. Clicking FLV videos is supposed to work, though obviously I know FLVs don't work in Ruffle, so this behavior is mostly expected.
[...]

Issue #10149, opened 2023-03-17
Bloom Defender: Broken FLV, title screen softlock


waspennator:
### Describe the bug

https://www.newgrounds.com/portal/view/583769?emulate=flash
Preloader seems to load up save for a flashing flv icon on the top left that shouldn't be there, then after the game intros, the game stops before it gets to the title screen.

[www.newgrounds.com-1679026182223.log](https://github.com/ruffle-rs/ruffle/files/10997823/www.newgrounds.com-1679026182223.log)

![Ruffle Bloom Defender 1](https://user-images.githubusercontent.com/78768499/225811415-10dc2855-c8b2-4d62-8883-489b11aad426.png)
![Ruffle Bloom Defender 2](https://user-images.githubusercontent.com/78768499/225811418-f429b1dc-1404-4780-b73c-a11c394d54f0.png)

### Expected behavior

The FLV should be playing normally and it should transition to the title screen.
![Flash Bloom Defender 1](https://user-images.githubusercontent.com/78768499/225811464-22799144-7e8a-49c6-a683-eb9346c02753.png)

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 110

n0samu:
Looks related to #10705 and to playerglobal versioning (there's the familiar `isPlaying` error)

olux997:
It turned out to be a problem with FLV Problem

Lord-McSweeney:
Unfortunately not fixed by #12967, though the `VerifyError`s no longer appear (instead there are now `Rejecting namespace match due to versions: Namespace("", SWF_13) Namespace("", FP_10_1)` logs)

waspennator:
Dunno when it got fixed, but seems to be working now

PR #10756, opened 2023-04-21
FLV support

kmeisthax:
This PR adds FLV support. Notably, only video: the existing streaming audio APIs want to work with either SWF bitstreams or single contiguous blocks of MP3 data. That needs refactoring big enough to justify a separate PR.

FLV also, bizarrely, has functionality for calling ActionScript methods from the video stream. Yes, the `onMetaData` event is actually triggered by one of these script calls; and no I did not implement it yet. I still do need to parse this block in order to not violate the video backend's existing API prerequisites, though.

~~This can't be merged until https://github.com/ruffle-rs/h263-rs/pull/50 is merged and this PR is updated to target the merged version.~~ I rolled that PR into this one so it can be merged in one go.


TheGhostboi64:
this is very cool! A defunct website I recreate mostly uses FLVs for the intros and others, so I can't wait to see it working! Nice!

Lord-McSweeney:
Fixes #10705.

waspennator:
Should fix the flv issue for #10149?

Lord-McSweeney:
I think the reason there's no functional difference currently is because the FLV PR isn't yet merged into the `h263-rs` repo.

danielhjacobs:
But wasn't the cargo lock changed to point to kmeisthax's latest commit?

kmeisthax:
Daniel is correct, `flv-rs` is being pulled in by commit ID separate from the other `h263-rs` packages. Hence why this is still in draft. (or it would be if I hadn't accidentally clicked the regular save button and had a clue as to how to turn on draft mode for a non-draft PR)

Lord-McSweeney:
You can turn it into a real draft by clicking the "Still in progress? Convert to draft" text below Reviewers - Suggestions, which is on the top-left of the page in Github.

kmeisthax:
I love how that link is buried where you can't find it

torokati44:
FWIW you can also switch the "Open PR" button to open as a draft right away.

core/src/avm2/globals/flash/media/video.rs
```diff
@@ -1,5 +1,42 @@
...
+use crate::display_object::{TDisplayObject, Video};
+
+/// Implements `flash.media.Video`'s `init` method, which is called from the constructor
+pub fn init<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Option<Object<'gc>>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    if let Some(this) = this {
+        activation.super_init(this, &[])?;
```
Lord-McSweeney, line 11: I don't think this line is necessary, since the AS constructor will call `super` anyway.

core/src/avm2/object/netstream_object.rs
```diff
@@ -15,15 +15,16 @@ pub fn netstream_allocator<'gc>(
...
-        NetStreamObjectData {
-            base,
-            ns: NetStream::new(activation.context.gc_context),
-        },
+        NetStreamObjectData { base, ns },
     ))
-    .into())
+    .into();
+
+    ns.set_avm_object(activation.context.gc_context, this.into());
```
iwannabethedev, line 25: Nit: I do not know if I like the cyclic reference between the `Object` and `NetStream`, but it might be fine, and alternatives, like a callback, might be overall worse.

core/src/display_object/video.rs
```diff
@@ -67,6 +64,12 @@ pub struct VideoData<'gc> {
     /// the prior keyframe. The first frame in the stream will always be
     /// treated as a keyframe regardless of it being flagged as one.
     keyframes: BTreeSet<u32>,
+
+    /// The movie whose tagstream or code created the Video object.
+    movie: Arc<SwfMovie>,
+
+    /// The self bounds for this movie.
+    size: (i32, i32),
```
iwannabethedev, line 72: Nit: Might this be better as `(u32, 32)`? Might be better as it currently is.

core/src/streams.rs
```diff
@@ -161,4 +262,357 @@ impl<'gc> NetStream<'gc> {
...
+                        Err(e) => {
+                            //TODO: Fire an error event to AS & stop playing too
+                            tracing::error!("FLV header parsing failed: {}", e);
+                            write.preload_offset = 3;
+                            return;
+                        }
+                    }
+                }
+                Some(_) => {
+                    //Unrecognized signature
```
iwannabethedev, line 301: Should this also have a `TODO` for firing an error event, like a few lines up?

core/src/streams.rs
```diff
@@ -161,4 +262,357 @@ impl<'gc> NetStream<'gc> {
...
+
+                let tag_needs_preloading = reader.stream_position().expect("valid position")
+                    as usize
+                    >= write.preload_offset;
+
+                match tag.data {
+                    FlvTagData::Audio(FlvAudioData { .. }) => {
+                        tracing::warn!("Stub: Stream audio processing");
+                    }
+                    FlvTagData::Video(FlvVideoData { codec_id, data, .. }) => {
```
iwannabethedev, line 350: The function is somewhat long. Would it be helpful (or worse) to maybe put parts of this function into their own functions that is called from here? Like, a function that might be called `handle_flv_tag_data_video()` for this match?

But, one problem with that is how it would interfere with the `loop`, in case there are `return` or `break` or `continue` statements and the like. The usage of a `loop`, together with a lot of code, feels a bit difficult to handle to me, though I could be wrong, both on this point and on possible alternatives on organizing the code being sufficiently better to change to one of them.

iwannabethedev:
It is obvious that a lot of time and effort went into this PR. I know too little of Rust, Ruffle and video playing to really give good feedback on this, but I have tried to leave a few comments. The PR is also rather long, making it more difficult to review. Parts that I think would be good to review is the overall type design and interface between the different parts of the codebase, and maybe also a light review of parts of the implementation, though this is of course only an initial PR and many parts have not been implemented yet or are works in progress.

I have not really reviewed this, so it is still in need of reviewing.

Lord-McSweeney:
https://www.nothingtodo.co.uk/view/3136/who-needs-a-glove.html logs "Video has no decoded frame to render" with this PR.

Lord-McSweeney:
Since DisplayObjects use instance allocators now, this needs to define an instance allocator on `Video`.

core/src/avm2/globals/flash/media/video.rs
```diff
@@ -1,5 +1,68 @@
...
+
+/// Implements `flash.media.Video`'s `init` method, which is called from the constructor
+pub fn init<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Option<Object<'gc>>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    if let Some(this) = this {
+        if let Some(video) = this.as_display_object().and_then(|dobj| dobj.as_video()) {
+            let width = args
```
Lord-McSweeney, line 48: I think it would be shorter to get `width` and `height` using `ParametersExt`.

kmeisthax: Good catch, I forgot `ParametersExt` existed in AVM2 land.

danielhjacobs:
Does this only work for H.263 FLV videos? I think https://github.com/ruffle-rs/ruffle/pull/10756#issuecomment-1517843846 may be because http://labcenter.dnalc.org/labs/dnaextraction/dnaextraction_d.html uses an FLV video with VP6A and MP3, for which this would need to interface with https://github.com/ruffle-rs/nihav-vp6. Additionally, the videos referenced by #10173 and #10175 are actually MP4 videos using the H.264 codec, which I know is unsupported.

kmeisthax:
@danielhjacobs This PR is already integrated with nihav-vp6. The DNA Extraction movie appears to be erroring out in nihav's validation checks; the FLV at http://labcenter.dnalc.org/labs/transformation/media/lauren_new_VP6_512K_Stream.flv has VP6 bitstreams with an invalid version number (16, as opposed to 6, 7, or 8).

Looking back at the FLV specification, they don't actually say what the video data payload is other than it being one of six opaque data types. In this case it would be a `VP6FLVALPHAVIDEOPACKET`, which isn't documented but I'm assuming works identically to the datastream in SWF files. That is, the alpha offset plus bitstreams for video and alpha in that order.

@Lord-McSweeney The Who Needs a Glove? video appears to be playing back correctly now, or at least as correctly as this PR permits.

torokati44:
Now that https://github.com/ruffle-rs/h263-rs/pull/50 got rolled into this PR, shouldn't this be "unmarked as draft" now?

torokati44:
>  the FLV at http://labcenter.dnalc.org/labs/transformation/media/lauren_new_VP6_512K_Stream.flv

I get a 404 for this file...

tests/README.md
```diff
@@ -14,11 +14,12 @@ As best practice, please also include any source used to make the swf - such as
 
 # Test Structure
 ## test.toml
-Except for `num_frames`, every other field and section is optional.
+Except for `num_ticks`, every other field and section is optional.
 
 ```toml
-num_frames = 1 # The amount of frames of the swf to run
-sleep_to_meet_frame_rate = false # If true, slow the tick rate to match the movies requested fps rate
+num_ticks = 1 # The amount of frames of the SWF to run.
```
Dinnerbone, line 20: Did you rename `num_frames` to `num_ticks`? If so, why do tests still have `num_frames`?

core/src/avm2/object/event_object.rs
```diff
@@ -144,6 +145,48 @@ impl<'gc> EventObject<'gc> {
...
+        let mut info_object = activation
+            .avm2()
+            .classes()
+            .object
+            .construct(activation, &[])
+            .unwrap();
+        for (key, value) in info {
+            info_object
+                .set_property(
+                    &Multiname::new(activation.avm2().public_namespace, AvmString::from(*key)),
```
Aaron1011, line 166: This can be `set_public_property`

tests/README.md
kmeisthax: A few tests failed when I tried switching everything over to ticks so it's opt-in for now.

danielhjacobs:
> > the FLV at http://labcenter.dnalc.org/labs/transformation/media/lauren_new_VP6_512K_Stream.flv
> 
> I get a 404 for this file...
> 
> EDIT: I think it's http://labcenter.dnalc.org/labs/dnaextraction/media/lauren_new_VP6_512K_Stream.flv instead...?

Correct

tests/README.md
Dinnerbone: Can you add `num_frames` back to the documentation and explain that it's one or the other then? I don't want the docs to not match what we're actually using

core/src/streams.rs
```diff
@@ -161,4 +262,364 @@ impl<'gc> NetStream<'gc> {
...
+                            Some(NetStreamType::Flv {
+                                stream, frame_id, ..
+                            }) => (stream, frame_id),
+                            _ => unreachable!(),
+                        };
+                        let codec = VideoCodec::from_u8(codec_id as u8);
+
+                        match (video_handle, codec, data) {
+                            (Some(video_handle), Some(codec), FlvVideoPacket::Data(data)) => {
+                                // NOTE: Currently, no implementation of the decoder backend actually requires
```
torokati44, line 363: Not for simple playback, but for discovering which frames are keyframes, and thus, for seeking to work at all.

tests/README.md
kmeisthax: Sure, but we ideally want all new tests to be in ticks, not frames. So I mentioned that `num_frames` is a deprecated option.

TemaisgameNoobas:
i guess that the audio playback will be on seperate PR's??
or it will be added along with this PR??

kmeisthax:
Audio playback requires some refactoring of how we handle data buffers. Everything in the existing audio backend wants an SWF data stream, or a contiguous MP3 data stream. In order to support FLV we need to be able to just hand it audio data chunks arbitrarily, which it currently can't do. So I have a separate branch for that which I plan to work on further once this PR merges.

core/src/streams.rs
```diff
@@ -359,16 +359,26 @@ impl<'gc> NetStream<'gc> {
...
                                 },
                             ) => {
+                                if codec == VideoCodec::ScreenVideo {
+                                    // ScreenVideo streams consider the FLV
+                                    // video data byte to be integral to their
+                                    // own bitstream.
+                                    let offset =
+                                        data.as_ptr() as usize - write.buffer.as_ptr() as usize;
+                                    let len = data.len();
+                                    data = &write.buffer[offset - 1..offset + len];
```
torokati44, line 379: Shouldn't this be `data = &write.buffer[offset - 1..offset + len + 1];` ?

kmeisthax: No, that would add an extra byte to the end of the bitstream

torokati44: Right, I confused `start:end` with `start:length`... Sorry about that.

core/src/streams.rs
```diff
@@ -359,16 +359,26 @@ impl<'gc> NetStream<'gc> {
...
                                 Some(video_handle),
                                 Some(codec),
                                 FlvVideoPacket::Vp6Data {
                                     hadjust: _,
                                     vadjust: _,
-                                    data,
+                                    mut data,
                                 },
                             ) => {
+                                if codec == VideoCodec::ScreenVideo {
```
torokati44, line 372: I think the same applies to V2 Screen Video. I plan to add the decoder for that back, and it's basically fused together with the current V1 one.

torokati44: But we can change this once we actually have that decoder... :)

kmeisthax: I'll add that condition in then

torokati44:
I posted a different Screen V1 test FLV on Discord, which will hopefully be less sensitive to platform differences.

PR #10766, opened 2023-04-21
Add Crowdin configuration

Dinnerbone:
This enables the crowdin-ci to work in the directory, and soon we can try the automatic PR integration too.

Crowdin project can be found at https://crowdin.com/project/ruffle

### DIFF
--- a/crowdin.yml
+++ b/crowdin.yml
@@ -1,0 +1,28 @@
+# To use the crowdin-cli, set an access token in CROWDIN_PERSONAL_TOKEN or a ~/.crowdin.yml
+# File structure documentation can be found at: https://developer.crowdin.com/configuration-file/
+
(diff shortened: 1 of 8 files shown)


crowdin.yml
```diff
@@ -0,0 +1,28 @@
+# To use the crowdin-cli, set an access token in CROWDIN_PERSONAL_TOKEN or a ~/.crowdin.yml
```
kmeisthax, line 1: Does this need me to add secrets to the organization or is this on Crowdin's end?

Dinnerbone: This is if a developer wants to the the CLI. The github integration does its own thing, not sure what it'll use exactly but we don't need to do anything more for that. It might use my account since I configured it?

Aaron1011 (approved):

Issue #10753, opened 2023-04-20
Potion Panic 2 - Cannot play game


n0samu:
### Describe the bug

Download SWF: [potionpanic2_decrypted.zip](https://github.com/ruffle-rs/ruffle/files/11286026/potionpanic2_decrypted.zip)

Clicking the Play button on the main menu of Potion Panic 2 does not start the game. The sound effect plays but nothing else happens.

### Expected behavior

The game should start.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

After clicking the Play button, first this error occurs:
```
ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "mouseDown", class: flash.events::MouseEvent, ptr: 0x270a43c9cd0 }) to handler FunctionObject(FunctionObject { ptr: 0x270a3cfaeb0 }) : TypeError: Error #1010: A term is undefined and has no properties. (accessing field: cacheAsBitmap)
        at Level/StopChildren()
        at Level/StopChildren()
        at Level/StopChildren()
        at Level/Resume()
        at Level/NextWave()
        at Level()
        at PotionPanic2/PlayGame()
        at MethodInfo-1415()
```
Then this error repeats many times:
```
ERROR run_frame:run_all_phases_avm2: ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x270c4a3b2e0 }) to handler FunctionObject(FunctionObject { ptr: 0x270d0dc4520 }) : TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: stage)
        at Level/Process()
```
Finally, this error occurs:
```
ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "timer", class: flash.events::TimerEvent, ptr: 0x270c4a67020 }) to handler FunctionObject(FunctionObject { ptr: 0x270d0dc6920 }) : TypeError: Error #2007: Parameter child must be non-null.
        at flash.display::DisplayObjectContainer/flash::display::DisplayObjectContainer::addChild()
        at mochi.as3::MochiServices$/setContainer()
        at mochi.as3::MochiSocial$/showLoginWidget()
        at Level/NextWave()
        at Level/Interwave()
        at flash.events::EventDispatcher/flash::events::EventDispatcher::dispatchEvent()
        at flash.utils::Timer/onUpdate()
```

[...]

PR #10767, opened 2023-04-21
avm2: Implement MorphShape class and use it for MorphShape displayobject

Aaron1011:
Makes progress on https://github.com/ruffle-rs/ruffle/issues/10753


core/src/display_object/morph_shape.rs
```diff
@@ -98,6 +104,31 @@ impl<'gc> TDisplayObject<'gc> for MorphShape<'gc> {
...
+
+    fn set_object2(&self, context: &mut UpdateContext<'_, 'gc>, to: Avm2Object<'gc>) {
+        self.0.write(context.gc_context).object = Some(to);
+    }
+
+    /// Construct objects placed on this frame.
+    fn construct_frame(&self, context: &mut UpdateContext<'_, 'gc>) {
+        if context.is_action_script_3() && matches!(self.object2(), Avm2Value::Null) {
+            let class = context.avm2.classes().morphshape;
+            let mut activation = Avm2Activation::from_nothing(context.reborrow());
```
Lord-McSweeney, line 123: Should this be `from_domain`?

Aaron1011: No, as nothing tries to access the domain. If this was wrong, it would panic when accessing the missing domain

Dinnerbone (approved):

PR #10783, opened 2023-04-22
avm2: Throw AVM errors for JSON operations (and add tests)

Lord-McSweeney:


core/src/avm2/globals/json.rs
```diff
@@ -261,13 +267,19 @@ pub fn parse<'gc>(
...
 ) -> Result<Value<'gc>, Error<'gc>> {
-    let input = args
-        .get(0)
-        .unwrap_or(&Value::Undefined)
-        .coerce_to_string(activation)?;
-    let reviver = args.get(1).unwrap_or(&Value::Undefined).as_object();
-    let parsed = serde_json::from_str(&input.to_utf8_lossy())
-        .map_err(|_| "SyntaxError: Error #1132: Invalid JSON parse input.")?;
+    let input = args.get_string(activation, 0)?;
+    let reviver = args.get_value(1).as_object();
```
Dinnerbone, line 271:
```suggestion
    let reviver = args.try_get_object(activation, 1);
```

core/src/avm2/globals/json.rs
```diff
@@ -277,21 +289,35 @@ pub fn stringify<'gc>(
     _this: Option<Object<'gc>>,
     args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
-    let val = args.get(0).unwrap_or(&Value::Undefined);
-    let replacer = args.get(1).unwrap_or(&Value::Undefined).as_object();
-    let spaces = args.get(2).unwrap_or(&Value::Undefined);
+    let val = args.get_value(0);
+    let replacer = args.get_value(1).as_object();
```
Dinnerbone, line 293:
```suggestion
    let replacer = args.try_get_object(activation, 1);
```

core/src/avm2/globals/json.rs
```diff
@@ -277,21 +289,35 @@ pub fn stringify<'gc>(
...
+    let spaces = args.get_value(2);
+
+    // If the replacer is None, that means it was a primitive (as_object returns None for primitives), and therefore not a valid replacer
+    if replacer.is_none() && !matches!(args.get(1).unwrap(), Value::Null) {
```
Dinnerbone, line 297: This then doesn't need the second half of the condition (null test)

Lord-McSweeney: This causes tests to fail because then a null replacer becomes invalid, and the default replacer is null.

Dinnerbone: I don't understand, `try_get_object` should return none for null

Dinnerbone (approved):

PR #10787, opened 2023-04-23
web: Add extension option for setting max. exec. time.

iwannabethedev:
Add an option to the 'options' menu for the extension for setting the maximum execution time for Actionscript code.

Since this option is likely to be rarely changed by users, it has not been added to the 'popup' UI, only the 'settings' UI.

The changes only modify the CSS layout to a minor degree. I might look into making the UI layout CSS more robust in a later pull request, though no promises.

### DIFF
--- a/web/packages/core/src/load-options.ts
+++ b/web/packages/core/src/load-options.ts
@@ -118,6 +118,16 @@ export const enum WindowMode {
 }
 
 /**
+ * Based on https://doc.rust-lang.org/stable/std/time/struct.Duration.html .
+ * Each member is an integer and non-negative.
+ */
+export interface Duration {
+
+    secs: number,
+    nanos: number,
+};
+
+/**
  * Any options used for loading a movie.
  */
 export interface BaseLoadOptions {
@@ -250,10 +260,7 @@ export interface BaseLoadOptions {
      *
      * @default { secs: 15, nanos: 0 }
      */
-    maxExecutionDuration?: {
-        secs: number;
-        nanos: number;
-    };
+    maxExecutionDuration?: Duration;
 
     /**
      * Specifies the base directory or URL used to resolve all relative path statements in the SWF file.
--- a/web/packages/core/src/load-options.ts
+++ b/web/packages/core/src/load-options.ts
@@ -122,10 +122,9 @@ export const enum WindowMode {
  * Each member is an integer and non-negative.
  */
 export interface Duration {
-
-    secs: number,
-    nanos: number,
-};
+    secs: number;
+    nanos: number;
+}
 
 /**
  * Any options used for loading a movie.
--- a/web/packages/extension/assets/_locales/en/messages.json
+++ b/web/packages/extension/assets/_locales/en/messages.json
@@ -20,6 +20,9 @@
     "settings_log_level": {
         "message": "Log level"
     },
+    "settings_max_execution_duration": {
+        "message": "How long can actionscript execute for before being forcefully stopped (in seconds)"
+    },
     "status_init": {
         "message": "Reading current tab…"
     },
--- a/web/packages/extension/assets/css/common.css
+++ b/web/packages/extension/assets/css/common.css
@@ -109,10 +109,8 @@ body {
 
 /* Number input. */
 
-.number_input input {
-
+.option.number-input input {
     width: 60px;
     height: 20px;
     margin: auto;
 }
-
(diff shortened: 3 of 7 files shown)


iwannabethedev:
I have tested it manually with a Flash application where one loading operation takes many seconds to complete (even if you use the old official Flash player).

iwannabethedev:
My apologies, I forgot to run `npm run lint` for the Typescript and CSS code. The linting now accepts the changes, and I have tested again that the options of the extension work as they should.

iwannabethedev:
Would it make sense to merge #9904 9904 before merging this? They touch some of the same files, and that PR deals with a more complex topic.

Dinnerbone:
FYI There was a want recently to change the Duration from `{secs, nanos}` to `secs` (number, not object), that might impact this too.

Though it'd allow both for backwards compatibility so maybe not?

iwannabethedev:
I think that would make sense, at the very least for `max_execution_duration`, since I do not see the fractional part being that useful. It would simplify this code, though, I could imagine that the change from type `{secs: number, nanos: number}` to `number` would not be difficult to change either now or later, especially with Typescript checking the types statically.

relrelb:
@iwannabethedev I've opened #10814 which should simplify this PR (and any future similar ones) a bit. So I prefer #10814 to be merged first, then this PR will need to be rebased, okay?

As for changing `{secs, nanos}` to just `secs`, I totally agree we should do it. For backward compatibility we can simply support both cases, and `console.warn` when the former is used, prompting developers to migrate. I also do prefer to do it before this PR so users won't have `{secs, nanos}` stored in their extension options. @iwannabethedev Would you like to give it a try?

relrelb:
> On rebasing this PR after #10814 is merged, that makes sense and sounds good to me.
> 
> On changing the type of `Duration` from `{secs, nanos}` to `secs`, if I understand you correctly, the PRs (in merge order) would be: #10814 (by you), then the PR with changing the type of `Duration` (by me, if I understand you correctly), then this PR (by me). If so, that sounds good to me, I can take a stab at changing the type of `Duration` and do it the way you propose (including backwards compatibility) as I understand it.

You understood right. Thank you and waiting for your PR!

iwannabethedev:
Wait, would #9904 be merged before or after #10814 ? I do not have a good feeling for which order of merging would be best, I have kind of been waiting for #10814 to be merged since it deals with a more complex issue than this PR and thus rebasing this PR would be simpler.

relrelb:
> Author

Yeah, #10814 is now merged. I think this PR doesn't have to wait for #9904, since #9904 is more complex and I don't want it to block things needlessly. But nevertheless it shouldn't conflict with this PR a lot, especially with #10814 in.

iwannabethedev:
Makes sense.

iwannabethedev:
> On rebasing this PR after #10814 is merged, that makes sense and sounds good to me.
> 
> On changing the type of `Duration` from `{secs, nanos}` to `secs`, if I understand you correctly, the PRs (in merge order) would be: #10814 (by you), then the PR with changing the type of `Duration` (by me, if I understand you correctly), then this PR (by me). If so, that sounds good to me, I can take a stab at changing the type of `Duration` and do it the way you propose (including backwards compatibility) as I understand it.

@relrelb Should I change the duration type only in the 'web' parts of the project?

If only the 'web' part: Only in Typescript or also in the Rust part of 'web'? Maybe with giving a custom deserialization function (like for the field `log_level` with `#[serde(deserialize_with = "deserialize_log_level")]`) in Rust for https://github.com/ruffle-rs/ruffle/blob/3cde3482f473cd76d210f1e8ffeee6fe6c7c52f2/web/src/lib.rs#L170 that handles both `{secs, nanos}` and `secs` (and gives a console.warn warning if the former is detected)?

Or should the duration type change be more extensive than that, also involving the non-web parts of the project?

I would guess the former with only being for the 'web' part, since the latter non-'web' Rust part uses `std::time::Duration`, which (while I do not know Rust) seems to make sense to continue using.

Dinnerbone:
> If only the 'web' part: Only in Typescript or also in the Rust part of 'web'? Maybe with giving a custom deserialization function (like for the field `log_level` with `#[serde(deserialize_with = "deserialize_log_level")]`) in Rust for

Rust side of the web crate. I wrote this recently as a custom deserializer but it's untested: https://gist.github.com/Dinnerbone/e9b4c6926dd5dc380a4b12c9c762c9bb

PR #10804, opened 2023-04-24
web: Don't inherit page CSS

n0samu:
Prevents Ruffle from inheriting page CSS, ensuring that it displays as expected. Prevents bugs like #8978 without needing to explicitly set every relevant style to its default value.


n0samu:
Seems inconsequential but sure I guess, done

relrelb (approved):

PR #6171, opened 2022-01-30
avm1: Fire `onReleaseOutside` event even when the button is invisible

Toad06:
Fixes #6169 and [this sample](https://github.com/ruffle-rs/ruffle/files/7966764/sample.zip).

PR #10807, opened 2023-04-24
avm1: Fix some issues with buttons

Toad06:
- If the mouse hovers a button which is then changed to be invisible, the `rollOut` event should trigger. This is a similar issue as 6171 (#6171).

- If the mouse hovers a button which is then disabled, the state of the button should only change with a mouse click or if the mouse moves away from the button. This can be observed in [DTunnel3D](https://www.newgrounds.com/portal/view/174308) when clicking the "Play" button.

- If the mouse hovers a disabled button which is then enabled, the `rollOver` event should trigger. This behavior can also be observed in DTunnel3D (click "Play" -> "Proceed", then don't move the mouse so it hovers "Menu").

I included a test that should cover all of the mentioned changes (inputs + visual).

The behavior will have to get verified for AVM2 buttons and movieclips in button mode. This will be for another PR.


Dinnerbone (approved):
Thank you!

PR #10808, opened 2023-04-24
web: Fix save manager styling

danielhjacobs:


n0samu:
This still looks weird when Ruffle is in a frame:
image

But if I remove the `max-height: 380px` style, that messes it up in other situations (like the web demo):
image

n0samu:
Also, here is the Flash project that I created to generate lots of dummy saves to test the save manager:
[junkGenerator.zip](https://github.com/ruffle-rs/ruffle/files/11317592/junkGenerator.zip)

web/packages/core/src/shadow-template.ts
```diff
@@ -336,6 +335,12 @@ ruffleShadowTemplate.innerHTML = `
 
         #local-saves {
             border-collapse: collapse;
+            overflow-y: scroll;
+            height: inherit;
```
n0samu, line 339: You're setting an actual height now so `height: inherit;` can be removed

web/packages/core/src/shadow-template.ts
```diff
@@ -336,6 +335,12 @@ ruffleShadowTemplate.innerHTML = `
 
         #local-saves {
             border-collapse: collapse;
+            overflow-y: scroll;
```
n0samu, line 338: Better to use `overflow-y: auto;` so the scrollbar is only shown when necessary, not all the time like this:
image

n0samu (approved):
This is nearly perfect now, thank you!

n0samu (approved):

PR #10811, opened 2023-04-24
avm2: Make `caller_domain` an `Option`, and call `expect` when actually needed

Aaron1011:
There are only a few places where we actually need to use the `caller_domain`, so we don't actually need it available for most native method calls. This means that `Activation::from_nothing` can be used in the vast majority of cases without causing a panic later on.

### DIFF
--- a/core/src/avm2/activation.rs
+++ b/core/src/avm2/activation.rs
@@ -580,7 +580,7 @@ impl<'a, 'gc> Activation<'a, 'gc> {
         this: Option<Object<'gc>>,
         subclass_object: Option<ClassObject<'gc>>,
         outer: ScopeChain<'gc>,
-        caller_domain: Domain<'gc>,
+        caller_domain: Option<Domain<'gc>>,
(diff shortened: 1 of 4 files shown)


Lord-McSweeney:
I think this will make it harder to catch cases where `from_domain` should have been used- most code that executes a function with `Executable::exec` can also run the Loader/ApplicationDomain/getDefinitionByName functions, and this will make it harder to catch that.

Aaron1011:
If we start out invoking a bytecode executable, then its down will get used as the caller domain.

This still panics when a domain is needed but missing, so I don't see why it would make it any harder to catch problems.

Bale001 (approved):

adrian17 (approved):

adrian17:
(Approved, though I'm thinking this hints that the "Activation" we use for using some AS state from native code, vs "Activation" that represents the interpreter itself, might have to be decoupled.)

Issue #10443, opened 2023-03-29
Flight: Upgrade screen invokes epilepsy 


HeyBanditoz:
### Describe the bug
SWF: [https://www.gameszap.com/files/swf/flight-1509798285.swf](https://www.gameszap.com/files/swf/flight-1509798285.swf)
(also attached to issue at [flight-1509798285.zip](https://github.com/ruffle-rs/ruffle/files/11097249/flight-1509798285.zip))

At the plane upgrade screen, all the upgrade and achievements cycles between different images (sprites?) very fast.
[Error Log](https://gist.github.com/HeyBanditoz/f50bebed936ad2acf71f248da8073e9c)
(the mouseOver and click errors only seem to happen when moving mouse over an upgrade, or clicking one, maybe a different issue?)
https://user-images.githubusercontent.com/7574664/228444224-11be736a-b8ee-4db8-b5fb-f6c6193b9467.mp4

#### Reproduction Steps
1. Load the game.
2. Create a new save.
3. Fly the plane, try to crash it.
4. Click next to skip the after-flight stats.
5. Observe the bug.

### Expected behavior
For the upgrades screen to look like this instead:
![image](https://user-images.githubusercontent.com/7574664/228444977-591ac50d-8373-4260-a200-585c7308677f.png)

### Affected platform
Linux build of [nightly-2023-03-29](https://github.com/ruffle-rs/ruffle/releases/tag/nightly-2023-03-29), but observed on web too.

### Operating system
Linux

### Additional information
```
$ lspci | grep VGA
00:02.0 VGA compatible controller: Intel Corporation TigerLake-LP GT2 [Iris Xe Graphics] (rev 01)
```

n0samu:
This is another issue caused by Class.hasOwnProperty being unimplemented. (And it will be very difficult to implement AFAIK.)

mara004:
+1

mara004:
#10864 sounds related

n0samu:
It does *sound* related, but that's XML.hasOwnProperty, not Class.hasOwnProperty. #10812 fixes Class.hasOwnProperty, but it unfortunately doesn't seem to help this game (besides getting rid of the error message). The new errors are:
```
ERROR ruffle_core::display_object::movie_clip: Error occured when running AVM2 frame script: TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: 0)
        at Flight_fla::UpgradeButton_156/setTip()
[...]

Issue #10009, opened 2023-03-12
Shadez 3: Black Screen Issue


waspennator:
### Describe the bug

https://www.crazymonkeygames.com/Shadez-3.html
Similar to Shadez 1, once you press play the game goes through the intro logos, then it goes to a black screen and doesn't progress further.

[www.crazymonkeygames.com-1678590976653.log](https://github.com/ruffle-rs/ruffle/files/10949906/www.crazymonkeygames.com-1678590976653.log)

### Expected behavior

Game should work without going to a black screen.

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 110

Lord-McSweeney:
I think this is the important error: `ReferenceError: Error #1069: Property http://adobe.com/AS3/2006/builtin::hasOwnProperty not found on Class and there is no default value.`

Lord-McSweeney:
Also needs `flash.net.Socket.connected` and `flash.geom.Matrix3D.interpolateTo`.

Lord-McSweeney:
This is still stuck on the same issue.

zohebnsr:
Any update on this?

Issue #9956, opened 2023-03-10
Mega Miner: Black screen


n0samu:
### Describe the bug

Link to SWF: http://i.notdoppler.com/files/megaminer.swf

[Mega Miner](http://www.notdoppler.com/megaminer.php) shows nothing but a black screen. As stated in https://github.com/ruffle-rs/ruffle/issues/1000#issuecomment-1459227097, the game needs TLF text support and `XML.settings()`.

### Expected behavior

The game should start.

### Affected platform

Desktop app

### Operating system

Windows 10

n0samu:
Now the error is:
```
ERROR run_frame:run_all_phases_avm2: ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "addedToStage", class: flash.events::Event, ptr: 0x1d33e2a3e80 }) to handler FunctionObject(FunctionObject { ptr: 0x1d362342050 }) : ReferenceError: Error #1069: Property perspectiveProjection not found on flash.geom.Transform and there is no default value.
        at MegaMiner_AG_fla::MainTimeline/__setPerspectiveProjection_()
```

Lord-McSweeney:
Also `ReferenceError: Error #1069: Property http://adobe.com/AS3/2006/builtin::hasOwnProperty not found on Class and there is no default value.`

n0samu:
With #10812, the new error is:
```
ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "addedToStage", class: flash.events::Event, ptr: 0x1d3df471638 }) to handler FunctionObject(FunctionObject { ptr: 0x1d3df6817d8 }) : ReferenceError: Error #1069: Property perspectiveProjection not found on flash.geom.Transform and there is no default value.
        at MegaMiner_AG_fla::MainTimeline/__setPerspectiveProjection_()
```

Lord-McSweeney:
This now spams out warnings repeatedly, still stuck on a black screen:
```
Display object container has no child with name __id19_
Display object container has no child with name __id1_
Display object container has no child with name __id19_
Display object container has no child with name __id1_
Display object container has no child with name __id19_
Display object container has no child with name __id1_
Display object container has no child with name __id19_
Display object container has no child with name __id1_
...
```

n0samu:
I actually fixed this with #11523, I just didn't notice because the Play button is invisible. Clicking its area goes to the main menu, from which you can click the invisible menu buttons and start the game. Not great but it works!

Issue #8952, opened 2022-12-28
Factory Balls 4 cannot start level


Lord-McSweeney:
### Describe the bug

After selecting the first level, https://uploads.ungrounded.net/581000/581575_factoryballs4.swf doesn't load the level. Instead, it spams `Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0xbd0090 }) to handler FunctionObject(FunctionObject(GcCell(Gc { ptr: 0xa2c2b4 }))) : RustError("Non-public property http://adobe.com/AS3/2006/builtin::hasOwnProperty not found on Object")` in the console, as well as `weak keys for Dictionary are not implemented`.

### Expected behavior

The game plays the first level.

### Affected platform

Browser's extension

### Operating system

Linux

### Browser

Chrome 108

### Additional information

There's also `ReferenceError: Error #1069: Property perspectiveProjection not found on flash.geom.Transform and there is no default value.` because `flash.geom.Transform.perspectiveProjection` is not implemented, but that error does not stop the game from running.

Lord-McSweeney:
Just asking- I've seen a lot of `RustError("Non-public property http://adobe.com/AS3/2006/builtin::something not found on Object")` errors, what are those caused by?

Lord-McSweeney:
This now needs `Transform.matrix3D`.

Lord-McSweeney:
Fixed by #11495.

PR #10812, opened 2023-04-24
avm2: Initialize `Object` instance vtable before other classes

Aaron1011:
This ensures that properties defined on the `Object` class (like `hasOwnProperty`) get correctly copied into the vtable for 'special' subclasses like `Class`.


Lord-McSweeney:
Progresses #8228 (now needs `flash.geom.Transform.matrix3D`).
Progresses #7866 (now needs TLF text).
Progresses #8889 (now needs `flash.geom.Transform.matrix3D`).
Progresses #8952 (now needs `flash.geom.Transform.matrix3D`).
Progresses #9956.
Progresses #10009.
Progresses #10443.

mara004:
Is this already sufficient to fix #10443, or just a start?

n0samu:
Just a start I guess - the issue isn't fixed by this.

adrian17 (approved):
Extreme spaghetti at this point 😅 
But looks just fine 👍

PR #10813, opened 2023-04-24
web: Update JSDoc lint rules

relrelb:
* Extend from `plugin:jsdoc/recommended-typescript-error`, which is more suitable for TypeScript and errors by default.
* Remove default and disabled but passing rules.
* Configure `jsdoc/tag-lines` to match JSDoc style of current codebase.


n0samu:
Sorry, I don't have any input to give on this PR; reading through the eslint-plugin-jsdoc readmes, I don't find their explanations understandable so I don't know what any of this stuff means. I assume it's fine though

Issue #8835, opened 2022-12-19
web: "Unable to lock Ruffle core" console spam after panic


n0samu:
### Describe the bug

Consider the Electricman panic (#4932, [SWF link](https://uploads.ungrounded.net/363000/363447_electricman2hs_fwg.swf)).
After the panic occurs, the console is spammed with thousands of "Unable to lock Ruffle core" error messages.
In fact, the message appears on every `requestAnimationFrame`. But why is Ruffle still calling requestAnimationFrame after `Ruffle instance destroyed.`?
As part of the panic handler, we can see that [the Ruffle instance gets destroyed](https://github.com/ruffle-rs/ruffle/blob/b323c086a943052e63ec2033f144650989189984/web/packages/core/src/ruffle-player.ts#L1450). And among other things, that code is supposed to [cancel the requestAnimationFrame calls](https://github.com/ruffle-rs/ruffle/blob/b323c086a943052e63ec2033f144650989189984/web/src/lib.rs#L394). So why isn't this working?

### Expected behavior

requestAnimationFrame should no longer be called after the Ruffle instance is destroyed. And the console spam should not happen.

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Firefox 108

### Additional information

This also occurs in Edge (Chromium) and in Ruffle selfhosted and web demo.

Lord-McSweeney:
This will probably be easier to debug now that the extension code is no longer minified

Lord-McSweeney:
Some basic debugging shows that requestAnimationFrame is repeatedly called by Ruffle.

n0samu:
I tried adding a bunch of trace statements to web/src/lib.rs to diagnose this. I found that the `destroy` function does in fact get called, but nothing inside the `if let` statement runs: https://github.com/ruffle-rs/ruffle/blob/d849e98902a3af64f3e00d4e74a54b2866a25526/web/src/lib.rs#L309

That line calls `remove_instance`, which never seems to return. Within that function, nothing after the `try_borrow_mut` call seems to execute:
https://github.com/ruffle-rs/ruffle/blob/d849e98902a3af64f3e00d4e74a54b2866a25526/web/src/lib.rs#L834

n0samu:
This is still happening for me in Firefox with the cases mentioned in #8427, as well as #11193. Looking for a way to reproduce it in Chromium.

n0samu:
[...]

PR #10819, opened 2023-04-25
Fix #10817

### LINKED ISSUE #10817: AVM2 stack traces in panics can cause Javascript execution
### Describe the bug

Because of this line of code:
https://github.com/ruffle-rs/ruffle/blob/940a8c6658d9f60db6d3e3d3c7623586de743720/web/packages/core/src/ruffle-player.ts#L1899

Panic text will not be sanitized. Putting a `</textarea>` inside an AS3 function name, and then panicking the player from inside the function will cause the HTML to "escape" the textarea. This textarea is only created when the "View Error Details" button is pressed, so the script can only execute then. This can allow a malicious SWF to execute Javascript. Note that this will not work on the extension page's player (the one shown when a SWF opens in a new tab), due to the strict content security policy forbidding script execution. POC SWF:
[avm2-malicious.swf.zip](https://github.com/ruffle-rs/ruffle/files/11318942/avm2-malicious.swf.zip)
```
const player = window.RufflePlayer.newest().createPlayer();
const container = document.createElement("div");
document.body.append(container);
container.appendChild(player);
player.load("avm2-malicious.swf");
```

### Expected behavior

Panic text is placed into the textarea instead of being set as `innerHTML`.

### Affected platform

Browser's extension

### Operating system

(irrelevant)

### Browser

_No response_

### Additional information

_No response_

Lord-McSweeney:
Create a child element instead of setting innerHTML.
Fixes #10817.

### DIFF
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -1898,7 +1898,10 @@ export class RufflePlayer extends HTMLElement {
                     this.container.querySelector("#panic-body")
                 );
                 panicBody.classList.add("details");
-                panicBody.innerHTML = `<textarea>${errorText}</textarea>`;
+                
+                let panicText = document.createElement("textarea");
+                panicText.value = errorText;
+                panicBody.appendChild(panicText);
                 return false;
             };
         }
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -1898,8 +1898,8 @@ export class RufflePlayer extends HTMLElement {
                     this.container.querySelector("#panic-body")
                 );
                 panicBody.classList.add("details");
-                
-                let panicText = document.createElement("textarea");
+
+                const panicText = document.createElement("textarea");
                 panicText.value = errorText;
                 panicBody.appendChild(panicText);
                 return false;


relrelb (approved):
LGTM. Please run `npm run format` in the `web` directory to appease ESLint.

divinity76:
textarea is kind-of an odd choice, do you want the stack trace to be editable for some reason? otherwise maybe
```js
const panicText = document.createElement("textarea");
panicText.setAttribute("disabled", "1");
```
or
```js
const panicText = document.createElement("span");
```

relrelb:
> textarea is kind-of an odd choice, do you want the stack trace to be editable for some reason? otherwise maybe
> 
> ```js
> const panicText = document.createElement("textarea");
> panicText.setAttribute("disabled", "1");
> ```
> 
> or
> 
> ```js
> const panicText = document.createElement("pre");
> panicText.textContent = errorText;
> ```

Yeah, good point. Funnily after a panic the parent `<ruffle-player>` still consumes keyboard events, making the `<textarea>` effectively "readonly" :)
Setting `disabled` grays out the `<textarea>`, so we better use `readOnly`, which only disallows editions.
A `<pre>` could also be used like you mentioned, but since it lacks border and background, some CSS will be needed.
For simplicity I'd like to stay with `<textarea>`, unless I missed some consideration. Would you like to submit a PR for this?

n0samu:
> Funnily after a panic the parent <ruffle-player> still consumes keyboard events, making the <textarea> effectively "readonly" :)

That's only because the Rust code that's supposed to clean up after Ruffle when it panics doesn't run correctly - see https://github.com/ruffle-rs/ruffle/issues/8835#issuecomment-1387307559. At some point I think we should refactor the panic-related code so the cleanup happens from the JS side instead of the Rust side, since calling back into the module after a panic doesn't work.

PR #10823, opened 2023-04-25
web: Parse `background_color` using a custom deserializer

relrelb:


iwannabethedev:
Though I have very little experience with Rust, the changes look very good and clean to me and like idiomatic Rust. Assuming you have tested with a few cases (both valid and invalid input), I think this is ready for merging (apart from rebasing).

Dinnerbone (approved):

PR #10829, opened 2023-04-26
web: Use default configuration in extension.

iwannabethedev:
I noticed by accident that some configuration settings had not been initialized in the extension, this PR should fix it.

Before:

before

After:

after

("Preferred renderer" not being set is correct, since its default value is `null`).

### DIFF
--- a/web/packages/extension/src/utils.ts
+++ b/web/packages/extension/src/utils.ts
@@ -1,6 +1,8 @@
 import type { Options } from "./common";
+import { DEFAULT_CONFIG as CORE_DEFAULT_CONFIG } from "ruffle-core";
 
-const DEFAULT_OPTIONS: Options = {
+const DEFAULT_OPTIONS: Required<Options> = {
+    ...CORE_DEFAULT_CONFIG,
     ruffleEnable: true,
     ignoreOptout: false,
     autostart: false,


iwannabethedev:
after_2

As proposed by @n0samu I believe , I have set the default value for `preferredRenderer` to `WebGpu`.

web/packages/extension/src/utils.ts
```diff
@@ -1,6 +1,8 @@
 import type { Options } from "./common";
+import { DEFAULT_CONFIG as CORE_DEFAULT_CONFIG } from "ruffle-core";
 
 const DEFAULT_OPTIONS: Options = {
+    ...CORE_DEFAULT_CONFIG,
```
relrelb, line 5: Great catch (my bad...)! Now I think `DEFAULT_OPTIONS` can be made `Required<Options>`.

iwannabethedev: Your changes fixed the previous duplication of initialization with default values, which can be a source of bugs and confusion, and therefore it is really nice that your changes fixed that duplication. I discovered (before your changes) that duplication, when I found out I had to update the initialization of a setting with a default value in multiple places.

I have added `Required<Options>`, and it did indeed compile and type-check. I have also rebased.

adrian17 (approved):

iwannabethedev:
The bug that this PR was meant to fix might have been caused by #10814 . This PR apparently caused bugs by itself, however. See also #10844 , #10850 and #10856 .

PR #10832, opened 2023-04-26
avm2: Use a make_error_2008 function

Lord-McSweeney:
Also converts a lot of error-2008s to AVM errors.

### DIFF
--- a/core/src/avm2/error.rs
+++ b/core/src/avm2/error.rs
@@ -135,6 +135,23 @@ pub fn make_reference_error<'gc>(
 
 #[inline(never)]
 #[cold]
+pub fn make_error_2008<'gc>(activation: &mut Activation<'_, 'gc>, param_name: &str) -> Error<'gc> {
+    let err = argument_error(
+        activation,
+        &format!(
+            "Error #2008: Parameter {} must be one of the accepted values.",
+            param_name
+        ),
+        2008,
+    );
+    match err {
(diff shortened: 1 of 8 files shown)


core/src/avm2/error.rs
```diff
@@ -133,6 +133,23 @@ pub fn make_reference_error<'gc>(
...
+pub fn make_error_2008<'gc>(activation: &mut Activation<'_, 'gc>, param_name: &str) -> Error<'gc> {
+    let err = argument_error(
+        activation,
+        &format!(
+            "Error #2008: Parameter {} must be one of the accepted values.",
+            param_name
+        ),
+        2008,
+    );
+    match err {
```
relrelb, line 147: Isn't the `?` operator can be used here? Something like:
```rs
Error::AvmError(argument_error(...)?)
```

Lord-McSweeney: I agree, but then the function has to return `Result<Error<'gc>, Error<'gc>>` and a further `?` has to be used when actually calling the function to get whichever of the `Error`s was the one part of the `Result`. This way is simpler and avoids an additional `?` every time the function is called.

relrelb: Got it. I didn't realize that `Result<Error<'gc>, Error<'gc>>` would be needed, so let's keep this as-is.

relrelb (approved):

Issue #7369, opened 2022-07-08
Excessive thickness (wide stroke) on shape outlines


hughmcmaster:
### Describe the bug

Outlines of basic shapes and embedded images are rendered with excessive thickness. In this Ruffle screenshot, this bug can be seen with the squares, flowers, caterpillar etc.

![image](https://user-images.githubusercontent.com/11186136/177934276-eee476b6-1cd1-4f53-bc1b-fccba8f2263d.png)

### Expected behavior

Outlines should be sized naturally. In most cases, this is a (very) thin stroke, as shown in this Flash Player screenshot.

![image](https://user-images.githubusercontent.com/11186136/177935523-e856810a-467a-4190-9132-acf35f1d8a6b.png)

### Affected platform

Online demo

### Operating system

Windows 10

### Browser

Mozilla Firefox 102

### Additional information

[Virtual Ecosphere](http://aws-website-diyecospherefromopenlearn-kldc0.s3-website-us-east-1.amazonaws.com/) (second Flash embed on the page)

n0samu:
I believe this is the same issue as #751 and #2023, with a fix for the Canvas backend being worked on in #7042

hughmcmaster:
Sadly, @Herschel's (merged) fix in #7042 had no effect in this case, but it is definitely a big step in the right direction.

n0samu:
Actually the problem is fixed now in the Canvas renderer only. To see the result, you would need to build Ruffle with only the Canvas renderer enabled, or manually disable WebGL in your browser by starting Chrome with the `--disable-webgl` option, or setting `webgl.disabled` to true in Firefox's `about:config`.

n0samu:
We have added a `preferredRenderer` config option which would allow you to use Ruffle's canvas renderer on your site without recompiling Ruffle. [See our wiki for more information](https://github.com/ruffle-rs/ruffle/wiki/Using-Ruffle#configuration-options).

PR #10835, opened 2023-04-26
web: Add `forceRenderer` config

relrelb:
Apart from `preferredRenderer`, which works in a "best effort" manner, is it useful to have a "mandatory" config for testing purposes. That is, if a forced renderer backend fails to initialized, then no further fallback is performed.

### DIFF
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -644,6 +644,10 @@ export class RufflePlayer extends HTMLElement {
             !("url" in options) || typeof options.url === "string",
             "`url` must be a string"
         );
+        check(
+            !("preferredRenderer" in options && "forceRenderer" in options),
+            "`preferredRenderer` and `forceRenderer` are mutual exclusive"
+        );
         return options;
     }
     /**
--- a/web/packages/extension/assets/_locales/en/messages.json
+++ b/web/packages/extension/assets/_locales/en/messages.json
@@ -21,7 +21,7 @@
         "message": "Log level"
     },
     "settings_max_execution_duration": {
-        "message": "How long can ActionScript execute for before being forcefully stopped (in seconds)"
+        "message": "Maximum allowed ActionScript execution duration (in seconds)"
     },
     "settings_player_version": {
         "message": "Flash player version number to emulate (range 1-32)"
--- a/web/packages/extension/assets/options.html
+++ b/web/packages/extension/assets/options.html
@@ -68,6 +68,16 @@
                         <input type="checkbox" id="ignore_optout" />
                         <label for="ignore_optout">Ignore website compatibility warnings</label>
                     </div>
+                    <div class="option select">
+                        <select id="force_renderer">
+                            <option value="">None</option>
+                            <option value="webgpu">WebGPU</option>
+                            <option value="wgpu-webgl">Wgpu (via WebGL)</option>
+                            <option value="webgl">WebGL</option>
+                            <option value="canvas">Canvas2D</option>
+                        </select>
+                        <label for="force_renderer">Force renderer</label>
+                    </div>
                     <div class="option number-input">
                         <input type="number" id="max_execution_duration" min="1" />
                         <label for="max_execution_duration">Maximum allowed ActionScript execution duration (in seconds)</label>
--- a/web/packages/extension/assets/options.html
+++ b/web/packages/extension/assets/options.html
@@ -19,7 +19,7 @@
                     class="logo" />
             </a>
         </div>
-        <div class="container">
+        <div class="options container">
             <div class="option checkbox">
                 <input type="checkbox" id="ruffle_enable" />
                 <label for="ruffle_enable">Play Flash content in Ruffle</label>
@@ -36,18 +36,6 @@
                 <input type="checkbox" id="warn_on_unsupported_content" />
                 <label for="warn_on_unsupported_content">Warn on unsupported content</label>
             </div>
-            <div class="option checkbox">
-                <input type="checkbox" id="ignore_optout" />
-                <label for="ignore_optout">Ignore website compatibility warnings</label>
-            </div>
-            <div class="option select">
-                <select id="log_level">
-                    <option value="info">Info</option>
-                    <option value="warn">Warning</option>
-                    <option value="error">Error</option>
-                </select>
-                <label for="log_level">Log level</label>
-            </div>
             <div class="option select">
                 <select id="preferred_renderer">
                     <option value="webgpu">WebGPU</option>
@@ -58,12 +46,6 @@
                 <label for="preferred_renderer">Preferred renderer</label>
             </div>
             <div class="option number-input">
-                <input type="number" id="max_execution_duration"
-                       min="1"
-                />
-                <label for="max_execution_duration">How long can ActionScript execute for before being forcefully stopped (in seconds)</label>
-            </div>
-            <div class="option number-input">
                 <input type="number" id="player_version"
                        min="1"
                        max="32"
@@ -71,6 +53,27 @@
                 />
                 <label for="player_version">Flash player version number to emulate (range 1-32)</label>
             </div>
+            <details>
+                <summary>Advanced Options</summary>
+                <div class="options">
+                    <div class="option select">
+                        <select id="log_level">
+                            <option value="info">Info</option>
+                            <option value="warn">Warning</option>
+                            <option value="error">Error</option>
+                        </select>
+                        <label for="log_level">Log level</label>
+                    </div>
+                    <div class="option checkbox">
+                        <input type="checkbox" id="ignore_optout" />
+                        <label for="ignore_optout">Ignore website compatibility warnings</label>
+                    </div>
+                    <div class="option number-input">
+                        <input type="number" id="max_execution_duration" min="1" />
+                        <label for="max_execution_duration">Maximum allowed ActionScript execution duration (in seconds)</label>
+                    </div>
+                </div>
+            </details>
         </div>
         <script src="dist/options.js"></script>
     </body>
--- a/web/packages/extension/src/options.ts
+++ b/web/packages/extension/src/options.ts
@@ -3,5 +3,10 @@ import { bindOptions } from "./common";
 
 window.addEventListener("DOMContentLoaded", () => {
     document.title = utils.i18n.getMessage("settings_page");
-    bindOptions();
+    const preferredRenderer = document.getElementById(
+        "preferred_renderer"
+    )! as HTMLSelectElement;
+    bindOptions((options) => {
+        preferredRenderer.disabled = !!options.forceRenderer;
+    });
 });
--- a/web/packages/core/src/config.ts
+++ b/web/packages/core/src/config.ts
@@ -36,4 +36,5 @@ export const DEFAULT_CONFIG: Required<BaseLoadOptions> = {
     polyfills: true,
     playerVersion: null,
     preferredRenderer: null,
+    forceRenderer: null,
 };
--- a/web/packages/core/src/load-options.ts
+++ b/web/packages/core/src/load-options.ts
@@ -393,6 +393,20 @@ export interface BaseLoadOptions {
     preferredRenderer?: RenderBackend | null;
 
     /**
+     * The forced render backend of the Ruffle player.
+     *
+     * This option should only be used for testing;
+     * the available backends may change in future releases.
+     * By default, Ruffle chooses the most featureful backend supported by the user's system,
+     * falling back to more basic backends if necessary.
+     * The available values in order of default preference are:
+     * "webgpu", "wgpu-webgl", "webgl", "canvas".
+     *
+     * @default null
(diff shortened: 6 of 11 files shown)


iwannabethedev:
Could it maybe make sense to mimic Phaser 3's similar behavior?

https://newdocs.phaser.io/docs/3.55.2/Phaser.Types.Core.GameConfig

https://newdocs.phaser.io/docs/3.60.0/Phaser.WEBGL

phaser_preferred_renderer

As I understand Phaser 3's behavior, if the setting is "AUTO", Phaser 3 tries WebGL and falls back to Canvas if WebGL is not supported. If the setting is not "AUTO" and instead is specific, like "WEBGL" or "CANVAS", then it does not fall back (maybe fails?).

Phaser 3 and Ruffle have different audiences - Phaser 3 is for game developers only, while Ruffle is for players, webmasters and game/application developers, as I understand Ruffle and Phaser 3. Phaser 3 is also a game engine/framework, while Ruffle is an emulator. And this setting for Phaser is usually set during development of a game, not during runtime by users. But I do imagine that similar behavior as Phaser 3 might make sense. "AUTO" seems to make best sense for players and probably webmasters and game developers, while the specific settings would be useful for players and both game and Ruffle-itself developers trying to debug and play a game that is not working (or test or debug Ruffle).

n0samu:
I'm not open to any change in the design of the preferred renderer feature, sorry. You can add your additional boolean if you like though

iwannabethedev:
I personally like the idea of an additional boolean, and I would not mind looking at providing a PR that implements a boolean that keeps the current behavior while the boolean setting is off and makes the choice strict when the boolean setting is on (with an exception for any "automatic" option, since I think that specific choice should always be flexible). But, I am a minor contributor, and not at all a core developer.

iwannabethedev:
> It is user-hostile to force a panic unnecessarily. I understand that "use a specific renderer or panic" is useful for development purposes. So let's provide an _additional_ forceRenderer boolean that triggers a panic if the chosen preferred renderer cannot be initialized. Don't force extension users and webmasters to do something dangerous just to specify their preferred renderer.

After thinking it over, I am warming further up to your argumentation. Comparing with Phaser 3, Phaser 3 has fewer kinds of audiences and use case situations - primarily game developers - and Phaser 3 does provide some flexibility such as in the form of "AUTO". Ruffle has more audiences and use case situations (including runtime situations - possibly primarily runtime situations, given that a very large number of games and applications have already been written), so, from an intuition point of view, providing extra flexibility feels right, at least to me.

torokati44:
FWIW I'm adding my +1 to the additional "force (no fallback)" boolean/checkbox.

iwannabethedev:
I feel like this is a somewhat complex situation, and I am not the one having to deal with the consequences down the road of whichever choice is made, so I do not feel like I have any right to dictate anything, honestly.

firoball:
View from outside:
Alternatively there could be a "default" list item which keeps the current fallback behaviour, so the common user does not need to bother. But I feel a "force" flag (disabled by default) is the clearer approach.

iwannabethedev:
I wrote some basic use case analysis on Discord.

I feel like there are strong use cases for at least an option like `Default`/`Auto`/`Automatic`, where there is a single setting where Ruffle falls back automatically depending on support, namely an option that users can use without worrying about or having to set the renderer backend used and where it tries to use the best possible renderer. This single setting would start with the "best" possible renderer backend and try down a list to find a renderer backend that is supported in the current environment. That list is currently: `["webgpu", "wgpu-webgl", "webgl", "canvas"]`, where the first attempted renderer backend is `"webgpu"`. The use would not choose a specific renderer backend, but an abstract "default" or "auto" renderer backend option. We could call this option **default**.

And there are also clear use cases for having mandatory/forced settings for each renderer backend, namely for Ruffle developers testing a specific renderer backend, or for users, game developers and webmasters for trying out different renderer backends for a given application as part of making that application work as workarounds. We could call these options **forced-each**.

The use case for having settings similar to `Default`/`Auto`/`Automatic`, but where the first attempted renderer backend is not `"webgpu"` but set by the user, is less clear to me. As far as I can figure out, the current solution supports this functionality through each option being preferred. I attempted to write some use cases for on Discord. I think if one or more clear use cases for this can be described, then the current solution with having a "preferred" or `Default` for each renderer backend, together with a forced/mandatory boolean setting, is much more clearly justified. We could call these options **preferred-each**.

In #9904 , support for **preferred-each** was implemented. #10831 looks at adding support for **default** (which with support for **preferred-each** and #9904 is rather straight-forward). This PR would change the supported functionality from **preferred-each** to **forced-each** (and I assume it could also support **default** with a future PR). Adding a boolean setting in a PR that forces a renderer backend, to the functionality of #9904 and #10831 , would provide support for both **preferred-each**, **default** and **forced-each**.

I believe it is clear-cut that at least **forced-each**, and also **default**, should be supported somehow given the strong use cases. (please do correct me if I am wrong on this point).

Issue #9187 describes a real-world use case using the game "Bloons Super Monkey" that supports setting a specific renderer backend by users or webmasters, and this real-world use case supports at least the **forced-each** options and arguably also the **preferred-each** options. Issue #7369 is also relevant. VAO issues #1905 and #9665 might also be relevant.

I do not have a good feel for what the drawbacks of supporting **preferred-each** might be. The discussion in https://github.com/ruffle-rs/ruffle/pull/9904 argues possible API issues. I think one possible take-away from that discussion is that **forced-each** and **preferred-each** (excluding **default**) might be helpfully described to users as an "Advanced" feature, or add a clear notice that any specific renderer backend is not guaranteed to be supported in the future. Such that the feature is not seen as something that users would normally set and that it does not become part of any API or expectation that Ruffle pledges to support. If this clear warning is added (both in the code and also in the UI facing users), like, maybe just "no single specific renderer backend is guaranteed to be supported in the future", is there any drawbacks to including/keeping support for **preferred-each**?

Finally, I am more or less a newcomer with little insight into Ruffle, so please do correct any and all errors, misconceptions, etc. I might have made in this comment.

relrelb:
Thanks all for your feedback! Sorry for my initial changes, it was I who misunderstood the situation.

Now `preferredRenderer` is restored and a separate `forceRenderer` is added. I made `preferredRenderer` and `forceRenderer` mutual exclusive (which is enforced by `checkOptions` in `ruffle-player.ts`) rather than making `forceRenderer` a boolean just to clarify things and avoid misconfigurations.

Let me know what do you think.

web/packages/extension/assets/options.html
```diff
@@ -55,7 +55,7 @@
                     <option value="webgl">WebGL</option>
                     <option value="canvas">Canvas2D</option>
                 </select>
-                <label for="preferred_renderer">Preferred renderer</label>
+                <label for="force_renderer">Force renderer</label>
```
n0samu, line 58: Looks like a leftover change from the previous version of this PR

danielhjacobs:
> Change the extension options page to use `forceRenderer`, since `preferredRenderer` is only expected to be used by webmasters, not Ruffle developers nor end users.

Isn't it the opposite? End-users should not using an option that can force a panic.

iwannabethedev:
Apart from merging in the most recent changes (including the changes that removed `MaxExecutionDurationOption`, which I think is a nice change given that `MaxExecutionDurationOption` was somewhat of an eyesore and rather special-cased, already when I originally wrote it), is there anything missing or needed in this PR? I believe that any "Default" or "Automatic" option, as in https://github.com/ruffle-rs/ruffle/pull/10831 , can always be adapted to the changes of this PR, since this PR deals with more complex issues than #10831 , and so I think merging this first makes sense. I do not mind at all adapting #10831 to this PR after this PR has been merged.

CC @relrelb @n0samu .

n0samu:
Please wait, I need some time to try my own approach, sorry

iwannabethedev:
I assume that you will coordinate together to figure out a good way to approach the feature (maybe or maybe not try out different approaches) and come to an all-things-considered good solution (a good solution, hopefully with both the UI and with the internal code), without spending lots of your time and energy.

iwannabethedev:
That sounds good, if you can coordinate with how to approach it, that will be very good. I imagine that two different approaches might create good grounds for taking the best of both.

web/packages/extension/assets/options.html
```diff
@@ -71,6 +53,37 @@
...
+                    </div>
+                    <div class="option select">
+                        <select id="force_renderer">
+                            <option value="">None</option>
+                            <option value="webgpu">WebGPU</option>
+                            <option value="wgpu-webgl">Wgpu (via WebGL)</option>
+                            <option value="webgl">WebGL</option>
+                            <option value="canvas">Canvas2D</option>
+                        </select>
+                        <label for="force_renderer">Force renderer</label>
```
iwannabethedev, line 79: Would it make sense to inform the user somehow of the exclusivity of "Preferred renderer" and "Force renderer", apart from "Preferred renderer" being disabled if "Force renderer" is set?

iwannabethedev: Like, maybe "Force renderer (exclusive with "Preferred renderer")?

That said, "Force renderer" is an "Advanced option", which helps a lot.

web/packages/extension/assets/_locales/en/messages.json
```diff
@@ -21,7 +21,7 @@
         "message": "Log level"
     },
     "settings_max_execution_duration": {
-        "message": "How long can ActionScript execute for before being forcefully stopped (in seconds)"
+        "message": "Maximum allowed ActionScript execution duration (in seconds)"
     },
```
iwannabethedev, line 25: Would it make sense to add translation strings for both `preferred_renderer` and `force_renderer`? I think it might have been forgotten in https://github.com/ruffle-rs/ruffle/pull/9904 (I forgot it myself in at least one PR until I got reminded of it in a different PR). Unless it was omitted on purpose in https://github.com/ruffle-rs/ruffle/pull/9904 .

From what I remember of the code, it is only necessary to make entries here, since the ID value and `for` value in the HTML is used for the translation key, so the entry keys would be `settings_preferred_renderer` and `settings_force_renderer` .

web/packages/extension/src/options.ts
```diff
@@ -3,5 +3,10 @@ import { bindOptions } from "./common";
 
 window.addEventListener("DOMContentLoaded", () => {
     document.title = utils.i18n.getMessage("settings_page");
-    bindOptions();
+    const preferredRenderer = document.getElementById(
+        "preferred_renderer"
+    )! as HTMLSelectElement;
+    bindOptions((options) => {
+        preferredRenderer.disabled = !!options.forceRenderer;
```
iwannabethedev, line 10: I am not certain, and this may be a bad idea, but I am wondering if this might make sense:

```
        preferredRenderer.disabled = !!options.forceRenderer && !options.preferredRenderer;
        forceRenderer.disabled = !!options.preferredRenderer && !options.forceRenderer;
```

The above is a bit involved, but, is basically meant to ensure that both options disables each other, but also allowing the user to unset one of them if the configuration somehow ends up with both being set.

web/packages/core/src/ruffle-player.ts
```diff
@@ -644,6 +644,10 @@ export class RufflePlayer extends HTMLElement {
             !("url" in options) || typeof options.url === "string",
             "`url` must be a string"
         );
+        check(
```
iwannabethedev, line 647: I do not understand the codebase well enough for this, but, will the user see a UI error if they mess up the settings?

Though, it might not be the responsibility of this PR to ensure that the user can see a UI error if that happens, but a new issue and PR.

I am not certain.

Is `check` meant to be used only for cases where there is a programmer error, or should it also be used for cases where UI-users have made a mistake? If the former, I am not certain that it is a good idea to use `check` here, unless the UI ensures that a UI-user cannot make a mistake and end up failing here.

web/packages/extension/src/options.ts
iwannabethedev: But will this require a default value for `preferredRenderer` to be set, which will be added in a different PR as I understand the plans? So it might not make sense to add this now, now that I think about it (if it makes sense in general to add it).

iwannabethedev:
I like the PR overall, there are some nice and very nice parts. There are a few nits (translation, I think). I have some questions, but I think they to some degree depend on the PR in https://github.com/ruffle-rs/ruffle/pull/10831 , which I think will be merged after this one - though, that PR might become just a one-liner given the additions and changes here.

web/packages/core/src/ruffle-player.ts
iwannabethedev: Will using the `in` operator here really work? Should it not instead be:

```!(options.preferredRenderer && options.forceRenderer),```

web/packages/extension/assets/_locales/en/messages.json
n0samu: I didn't know about adding translation strings, so yeah, that was an oversight on my part.

n0samu:
Thanks for your work on this! At this point I think each piece of this PR has been either superseded or moved to other PRs, so this one can be closed.

PR #10838, opened 2023-04-26
extension: Close `popup.html` when opening/reloading a window

Toad06:


web/packages/extension/src/popup.ts
```diff
@@ -141,12 +141,8 @@ window.addEventListener("DOMContentLoaded", () => {
...
     ) as HTMLButtonElement;
     reloadButton.textContent = utils.i18n.getMessage("action_reload");
-    reloadButton.addEventListener("click", async () => {
-        await utils.tabs.reload(activeTab.id!);
-        // TODO: Wait for tab to load?
-        setTimeout(() => {
-            displayTabStatus();
-        }, 1000);
+    reloadButton.addEventListener("click", () => {
+        utils.tabs.reload(activeTab.id!).then(() => window.close());
```
relrelb, line 145: nit: Let's keep this an `async` function, then `await` can still be used.

web/packages/extension/src/utils.ts
```diff
@@ -122,9 +122,9 @@ if (typeof chrome !== "undefined") {
     runtime = chrome.runtime;
 
     openOptionsPage = () =>
-        promisify((cb: () => void) =>
+        promisify((cb) =>
             chrome.tabs.create({ url: "/options.html" }, cb)
-        );
+        ).then(() => window.close());
```
relrelb, line 127: I'd rather move the `window.close()` to the caller of `utils.openOptionsPage()` (`popup.ts`), since this should be just a low-level compatibilily layer for extensions API.

relrelb (approved):
Thank you!

web/packages/extension/src/popup.ts
```diff
@@ -135,18 +135,16 @@ window.addEventListener("DOMContentLoaded", () => {
...
         "reload-button"
     ) as HTMLButtonElement;
     reloadButton.textContent = utils.i18n.getMessage("action_reload");
     reloadButton.addEventListener("click", async () => {
-        await utils.tabs.reload(activeTab.id!);
-        // TODO: Wait for tab to load?
-        setTimeout(() => {
-            displayTabStatus();
-        }, 1000);
+        await utils.tabs.reload(activeTab.id!).then(() => window.close());
```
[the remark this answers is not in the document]
Toad06: No worries (I'm to blame :D). Thank you for the review.

PR #10848, opened 2023-04-27
extension: Add option for setting `playerVersion`

### LINKED ISSUE #10840: Add setting `playerVersion` as an extension option.
**Describe the feature**
A user on Discord described a situation where a Flash application gave an error screen claiming that the Flash version was 3, and that Flash version at least 8 was required. However, the Flash version reported by Ruffle was considerably higher, maybe 32, and that the Flash application in question might have taken just the first digit of the Flash version when checking for version.

3nVK5kNTU5AH

The desktop and web player supports setting the `playerVersion` ( https://github.com/ruffle-rs/ruffle/blob/02b843464854f18307fdbd2b7cb50ba0ea02343e/web/packages/core/src/load-options.ts#L375 ) that is reported by Ruffle. This feature request is about adding it to the extension as an option as well, such that users can set it.

Thanks to @Dinnerbone and @danielhjacobs for figuring out and describing this feature request (this issue is made by me to avoid forgetting the feature request and to have something to refer to for a PR).

iwannabethedev:
Fixes #10840 .

The changes have been tested lightly manually.

### DIFF
--- a/web/packages/extension/assets/options.html
+++ b/web/packages/extension/assets/options.html
@@ -63,6 +63,13 @@
                 />
                 <label for="max_execution_duration">How long can ActionScript execute for before being forcefully stopped (in seconds)</label>
             </div>
+            <div class="option number-input">
+                <input type="number" id="player_version"
+                       min="1"
+                       max="32"
+                />
+                <label for="player_version">Flash player version number to emulate (range 1-32)</label>
+            </div>
         </div>
         <script src="dist/options.js"></script>
     </body>
--- a/web/packages/extension/src/common.ts
+++ b/web/packages/extension/src/common.ts
@@ -1,5 +1,5 @@
 import * as utils from "./utils";
-import type { BaseLoadOptions, Duration } from "ruffle-core";
+import type { BaseLoadOptions } from "ruffle-core";
 
 export interface Options extends BaseLoadOptions {
     ruffleEnable: boolean;
@@ -32,27 +32,43 @@ class CheckboxOption implements OptionElement<boolean> {
     }
 }
 
-/**
- * Number input option for 'maxExecutionDuration'.
- */
-class MaxExecutionDurationOption implements OptionElement<Duration> {
+class NumberOption implements OptionElement<number | null> {
     constructor(
         private readonly numberInput: HTMLInputElement,
         readonly label: HTMLLabelElement
-    ) {}
+    ) {
+        this.numberInput.addEventListener("input", () => {
+            this.numberInput.reportValidity();
+        });
+    }
 
     get input() {
         return this.numberInput;
     }
 
-    get value() {
-        return Math.max(1, Math.round(this.numberInput.valueAsNumber));
+    get value(): number | null {
+        const ni = this.numberInput;
+        const num = ni.valueAsNumber;
+        if (Number.isNaN(num)) {
+            return null;
+        }
+        if (ni.min) {
+            const min = Number(ni.min);
+            if (min > num) {
+                return min;
+            }
+        }
+        if (ni.max) {
+            const max = Number(ni.max);
+            if (num > max) {
+                return max;
+            }
+        }
+        return num;
     }
 
-    set value(value: Duration) {
-        // Ignoring the 'nanos' part from the old format type ObsoleteDuration.
-        const newValue = typeof value === "number" ? value : value.secs;
-        this.numberInput.value = newValue.toString();
+    set value(value: number | null) {
+        this.numberInput.value = value?.toString() ?? "";
     }
 }
 
@@ -89,9 +105,7 @@ function getElement(option: Element): OptionElement<unknown> {
         }
 
         if (input.type === "number") {
-            if (input.id === "max_execution_duration") {
-                return new MaxExecutionDurationOption(input, label);
-            }
+            return new NumberOption(input, label);
         }
     }
 
--- a/web/packages/core/src/load-options.ts
+++ b/web/packages/core/src/load-options.ts
@@ -166,6 +166,10 @@ export interface ObsoleteDuration {
     nanos: number;
 }
 
+/**
+ * Any new duration-based setting should use 'number' or 'SecsDuration' for its type,
+ * instead of this type.
+ */
 export type Duration = SecsDuration | ObsoleteDuration;
 
 /**
--- a/web/packages/extension/assets/_locales/en/messages.json
+++ b/web/packages/extension/assets/_locales/en/messages.json
@@ -23,6 +23,9 @@
     "settings_max_execution_duration": {
         "message": "How long can ActionScript execute for before being forcefully stopped (in seconds)"
     },
+    "settings_player_version": {
+        "message": "Flash player version number to emulate (range 1-32)"
+    },
     "status_init": {
         "message": "Reading current tab…"
     },
--- a/web/src/lib.rs
+++ b/web/src/lib.rs
@@ -133,7 +133,7 @@ where
     tracing::Level::from_str(&value).map_err(Error::custom)
 }
 
-fn deserialize_max_execution_duration<'de, D>(deserializer: D) -> Result<Duration, D::Error>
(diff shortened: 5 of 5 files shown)


web/packages/extension/assets/options.html
```diff
@@ -57,6 +57,11 @@
                 </select>
                 <label for="preferred_renderer">Preferred renderer</label>
             </div>
+            <div class="option number-input">
+                <input type="number" id="player_version"
+                />
+                <label for="player_version">Flash player version number to emulate</label>
```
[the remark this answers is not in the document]
iwannabethedev: Is there anything I should do in the code other than include `settings_player_version` in 'web/packages/extension/assets/_locales/en/messages.json' ? (I forced-pushed a new commit that included an entry there after I realized I forgot to include it previously after a very good and thorough review by relrelb in a different PR). The existing code appends `"settings_"` to the `id` used for the element, and that string, like `"settings_player_version"`, is then used to look up a translation. https://github.com/ruffle-rs/ruffle/blob/750431acec39117023c1f9a447e7fa9ce83cf0c7/web/packages/extension/src/common.ts#L135-L139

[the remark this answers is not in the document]
Dinnerbone: Ahh then it's all good, I didn't see the diff for that missing commit so I didn't make the association. We're good!

iwannabethedev:
I have added additional input validation logic and UI, since otherwise it was easy to panic the Ruffle player by inputting invalid values. I imagine there might be cases that could still panic the Ruffle player (maybe by inputting numbers with fractional parts for settings that requires integers), but minimum and maximum should be enforced and be reported to the user at least.

danielhjacobs:
I believe the max playerVersion that can actually be used is 32: https://github.com/ruffle-rs/ruffle/wiki/SWF-version-chart. I'm not sure if there's a point to going up to 255.

iwannabethedev:
@danielhjacobs That is a good point, and thank you for the link.

Looking at the table in https://github.com/ruffle-rs/ruffle/wiki/SWF-version-chart , the current setting in the Ruffle player is of type `u8`, and the chart includes version numbers with fractional parts. But, I imagine that it might not matter much in practice.

I will push a new change that decreases the maximum to 32.

danielhjacobs:
It might be good to also increase min to 1.

web/packages/extension/src/common.ts
```diff
@@ -32,14 +32,62 @@ class CheckboxOption implements OptionElement<boolean> {
...
+    }
+
+    get value(): number | null {
+        const ni = this.numberInput;
+        const num = ni.valueAsNumber;
+        if (Number.isNaN(num)) {
+            return null;
+        }
+        if (ni.min) {
+            const min = Number.parseFloat(ni.min);
```
relrelb, line 56: `Number.parseFloat()` is not really needed - The more basic `Number(ni.min)` is preferred.

web/packages/extension/src/common.ts
```diff
@@ -32,14 +32,62 @@ class CheckboxOption implements OptionElement<boolean> {
...
+            const max = Number.parseFloat(ni.max);
+            if (num > max) {
+                return max;
+            }
+        }
+        return num;
+    }
+
+    set value(value: number | null) {
+        if (value === null) {
```
relrelb, line 71: I find using the `?` and `??` operators here much more readable:
```ts
this.numberInput.value = value?.toString() ?? "";
```

web/packages/extension/src/common.ts
```diff
@@ -32,14 +32,62 @@ class CheckboxOption implements OptionElement<boolean> {
...
+
 /**
  * Number input option for 'maxExecutionDuration'.
  */
 class MaxExecutionDurationOption implements OptionElement<Duration> {
     constructor(
         private readonly numberInput: HTMLInputElement,
         readonly label: HTMLLabelElement
-    ) {}
+    ) {
```
relrelb, line 86: For code reuse, I suggest making `MaxExecutionDurationOption` a subclass of `NumberInputElement`:
```ts
class MaxExecutionDurationOption extends NumberInputElement implements OptionElement<Duration> { ... }
```

Also, how about renaming it to just `DurationOption`? At this point it doesn't really tied specifically to max execution time.

web/packages/extension/src/common.ts
iwannabethedev: I would like to avoid `NaN` here, since https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number claims that `NaN` is not a "valid floating point number", and that the `<input type="number"/>` tag required a "valid floating point number" for `value` to be considered valid:

> A number input is considered valid when empty and when a single number is entered, but is otherwise invalid. If the [required](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#required) attribute is used, the input is no longer considered valid when empty.
>
> Note: Any number is an acceptable value, as long as it is a [valid floating point number](https://html.spec.whatwg.org/multipage/infrastructure.html#valid-floating-point-number) (that is, not [NaN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN) or [Infinity](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Infinity)).

I am trying to use:

`this.numberInput.value = value === null ? "" : value.toString();`

instead to make that code more concise, though I have not yet tested it, I will test it before pushing.

web/packages/extension/src/common.ts
iwannabethedev: In regards to making it a subclass, the type `OptionElement<Duration>` cannot currently be a subtype of `OptionElement<number | null>` as long as `ObsoleteDuration` is still one possible case of `Duration`. If we take a chance and guess that `ObsoleteDuration` can never occur in the extension 'option' UI, then I would guess that we could use `NumberDuration` instead and then use `NumberOption` as the superclass.

In regards to renaming it for general duration usage, I believe that instead of using  `MaxExecutionDurationOption`, that the class `NumberOption` would be possible to use there instead for new duration-based settings that do not use `Duration` but only `number`/`SecsDuration`. I believe it would be a good idea if any new duration-based setting only use `number` for its type, not `Duration`. I am also using an element ID at https://github.com/ruffle-rs/ruffle/blob/d92e4c3487f07f3f0d0953e8af85440171481165/web/packages/extension/src/common.ts#L140 , but that should be possible to refactor, or alternatively just add another ID-check.

**Question:** Should we take a chance and assume that `ObsoleteDuration` can never occur in the extension 'option' UI? I do not at all have a good feel for whether it can be possible or not for that to occur. If that chance is taken, then I think `MaxExecutionDurationOption` can be removed.

**I have renamed the Rust deserialization function `deserialize_max_execution_duration` to `deserialize_duration` to make it clearer that it can be used for any duration deserialization.** I have also added a comment discouraging usage of `Duration` in favor of `number`/`SecsDuration`.

iwannabethedev:
I have lightly tested manually the most recently force-pushed changes.

web/packages/extension/src/common.ts
relrelb: Got it. In this case the code can be further reduced to:
```ts
this.numberInput.value = value?.toString() ?? "";
```
What do you think?

web/packages/extension/src/common.ts
relrelb: Great analysis. With that being said, I feel confidentally that `ObsoleteDuration` can never occour when configured from the extension UI, because `get value` always returns a number. So I think your conclusion is right and `MaxExecutionDurationOption` can be removed completely.

By the way, I like the renaming of `deserialize_max_execution_duration` to `deserialize_duration`!

relrelb (approved):
Thank you!

web/packages/extension/src/common.ts
iwannabethedev: I have removed `MaxExecutionDurationOption`, and `NumberOption` is used instead.

Issue #10844, opened 2023-04-27
Relative URLs in extension page are parsed relative to the extension page


Lord-McSweeney:
### Describe the bug

For example, opening this URL (chrome):
`chrome-extension://donbcfbmhbcapadipfkeojnmajbakjdc/player.html?url=http%3A%2F%2Fnew.weedtowonder.org%2Fw2w-home.swf`

And clicking "Domestication" will result in a navigation to
`chrome-extension://donbcfbmhbcapadipfkeojnmajbakjdc/domestication.html`.

The navigation should be to `http://new.weedtowonder.org/domestication.html`.

### Expected behavior

Paths on the extension page SWF viewer should be parsed correctly.

### Affected platform

Browser's extension

### Operating system

(irrelevant)

### Browser

Chrome

danielhjacobs:
Strangely, unless this bug was introduced in the last few days and therefore isn't in Firefox yet, this seems to not affect Firefox. I see you did mention Chrome in the browser section, but I thought it worth noting that this explicitly does not affect Firefox.

danielhjacobs:
In Firefox, the internal player page says for http://new.weedtowonder.org/w2w-home.swf says `document.getElementById('player').loadedConfig.base`. In Chrome, it says `null`.

danielhjacobs:
I think #5144 was bugged for Chrome.

relrelb:
Fixed by #10850.

PR #10850, opened 2023-04-27
extension: Fix popup setting changed reload detection.

iwannabethedev:
The problem seems to have been the inclusion of setting values that the previous equality function did not handle correctly. This function broadens the kinds of setting values that can be handled correctly.

### DIFF
--- a/web/packages/extension/src/popup.ts
+++ b/web/packages/extension/src/popup.ts
@@ -83,20 +83,36 @@ async function queryTabStatus(
     optionsChanged();
 }
 
-function objectsEqual<T>(x: NonNullable<T>, y: NonNullable<T>) {
-    for (const [key, value] of Object.entries(x)) {
-        if (y[key as keyof T] !== value) {
-            return false;
-        }
-    }
+/**
+ * Should only be called on data type objects without any "cyclic members" to avoid infinite recursion.
+ */
+function deepEqual(x: unknown, y: unknown): boolean {
+    if (
+        typeof x === "object" &&
+        typeof y === "object" &&
+        x !== null &&
+        y !== null
+    ) {
+        // Two non-null objects.
 
-    for (const [key, value] of Object.entries(y)) {
-        if (x[key as keyof T] !== value) {
-            return false;
+        for (const [key, value] of Object.entries(x)) {
+            if (!deepEqual(value, y[key as keyof typeof y])) {
+                return false;
+            }
         }
-    }
 
-    return true;
+        for (const [key, value] of Object.entries(y)) {
+            if (!deepEqual(value, x[key as keyof typeof x])) {
+                return false;
+            }
+        }
+
+        return true;
+    } else {
+        // Not two non-null objects.
+
+        return x === y;
+    }
 }
 
 function optionsChanged() {
@@ -104,7 +120,7 @@ function optionsChanged() {
         return;
     }
 
-    const isDifferent = !objectsEqual(savedOptions, tabOptions);
+    const isDifferent = !deepEqual(savedOptions, tabOptions);
     reloadButton.disabled = !isDifferent;
 }
 
--- a/web/packages/extension/src/player.ts
+++ b/web/packages/extension/src/player.ts
@@ -26,13 +26,12 @@ window.addEventListener("DOMContentLoaded", async () => {
     document.getElementById("main")!.append(player);
 
     const options = await utils.getOptions();
-    const config = {
-        letterbox: "on" as Letterbox,
-        ...options,
-    };
+
     player.load({
+        ...options,
+        // Override default value for 'letterbox' when playing in the extension player page.
+        letterbox: "on" as Letterbox,
         url: swfUrl,
         base: swfUrl.substring(0, swfUrl.lastIndexOf("/") + 1),
-        ...config,
     });
 });


Toad06:
This looks good to me based on testing. A review from @relrelb would still be appreciated.

relrelb:
While I don't 100% understand this fix (specifically for which inputs the new logic returns different results), I think a more proper (and simpler) approach would be to actually tune #10829 a bit - Instead of including `CORE_DEFAULT_CONFIG` in `DEFAULT_OPTIONS`, I suggest to move it inside `bindOptions()`, like this:
```ts
const options = {
    ...CORE_DEFAULT_CONFIG,
    ...await utils.getOptions(),
};
```
This way the full `CORE_DEFAULT_CONFIG` won't be actually saved in the extension storage, restoring the state before #10829, but still not-yet-set options will have their default value, preventing the empty selection boxes.

Note that I didn't test this yet, so I may be completely wrong after all.

relrelb:
> While I don't 100% understand this fix (specifically for which inputs the new logic returns different results), I think a more proper (and simpler) approach would be to actually tune #10829 a bit - Instead of including `CORE_DEFAULT_CONFIG` in `DEFAULT_OPTIONS`, I suggest to move it inside `bindOptions()`, like this:
> 
> ```ts
> const options = {
>     ...CORE_DEFAULT_CONFIG,
>     ...await utils.getOptions(),
> };
> ```
> 
> This way the full `CORE_DEFAULT_CONFIG` won't be actually saved in the extension storage, restoring the state before #10829, but still not-yet-set options will have their default value, preventing the empty selection boxes.
> 
> Note that I didn't test this yet, so I may be completely wrong after all.

Okay, I just confirmed this fixes the reload button always appearing, without regressing #10829:
image

This also fixes #10844, so I do think this is the preferred way. Going to toss a PR.

iwannabethedev:
> > While I don't 100% understand this fix (specifically for which inputs the new logic returns different results), I think a more proper (and simpler) approach would be to actually tune #10829 a bit - Instead of including `CORE_DEFAULT_CONFIG` in `DEFAULT_OPTIONS`, I suggest to move it inside `bindOptions()`, like this:
> > ```ts
> > const options = {
> >     ...CORE_DEFAULT_CONFIG,
> >     ...await utils.getOptions(),
> > };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This way the full `CORE_DEFAULT_CONFIG` won't be actually saved in the extension storage, restoring the state before #10829, but still not-yet-set options will have their default value, preventing the empty selection boxes.
> > Note that I didn't test this yet, so I may be completely wrong after all.
> 
> Okay, I just confirmed this fixes the reload button always appearing, without regressing #10829: image
> 
> This also fixes #10844, so I do think this is the preferred way. Going to toss a PR.

I do not understand #10844 , or why it is fixed by changing which and where the extension configuration settings are initialized to from the default settings, but I am getting the impression that the fix I made in #10829 was rather bad and wrong, and that I made some assumptions in it that I really should have checked before making.

I have looked a little bit at the code, and while I do not understand it as thoroughly as I should, when looking at 'web/packages/extension/src/player.ts', I wonder whether the following changes might make sense:

**_(Please be aware: I have not tested these changes or considered them in depth, and my understanding of and knowledge of this code base is very limited)._**

### First change

https://github.com/ruffle-rs/ruffle/blob/60ffe07ae7d810777c46b16df77ead92f1ce93d6/web/packages/extension/src/player.ts#L29-L32

is removed and `options` is used directly in the code below - the reasoning being that `letterbox: "on"` feels like setting a default value that is already set elsewhere - and in this case, the default value in player.ts is different from the default value in 'web/packages/core/src/config.ts' : https://github.com/ruffle-rs/ruffle/blob/60ffe07ae7d810777c46b16df77ead92f1ce93d6/web/packages/core/src/config.ts#L15 , https://github.com/ruffle-rs/ruffle/blob/60ffe07ae7d810777c46b16df77ead92f1ce93d6/web/packages/core/src/load-options.ts#L38-L53 , which feels wrong to me. Preferably, a default configuration value should only be set in one place. Though, it might just be me that have not looked into the code sufficiently, and I am not certain that the lines above are in fact a default configuration value.

### Second change

https://github.com/ruffle-rs/ruffle/blob/60ffe07ae7d810777c46b16df77ead92f1ce93d6/web/packages/extension/src/player.ts#L33-L37

is changed to:

```
    player.load({
        ...config,
        url: swfUrl,
        base: swfUrl.substring(0, swfUrl.lastIndexOf("/") + 1),
    });
```

, such that the configuration settings are overwritten by the new values for `url` and `base`, and not the other way around. I am not at all certain about this change, and it might introduce new bugs, but it feels as a starting approximation more correct to me - basically take the configuration, and overwrite it with relevant values.

I can attempt making these changes and see how they go, but I have not considered them in depth.

n0samu:
Letterbox being set to "on" in the extension's player page is intentional. Same reason we use it on the Ruffle Web Demo. The default value is "fullscreen" to provide the expected (Flash-like) behavior when Ruffle is used on websites, but for contexts where we're playing any arbitrary SWF with the same settings (the web demo and extension player page), we want to letterbox all the time to prevent offscreen elements from showing up and making the content look weird.

n0samu:
Also, if our recent changes are breaking everything we might want to consider reverting before considering making far-reaching changes to the codebase require extensive testing to rule out unintended consequences.

iwannabethedev:
> Also, if our recent changes are breaking everything we might want to consider reverting them before considering making far-reaching changes to the codebase requiring extensive testing to rule out unintended consequences.

I think the change in #10829 was in response to the change in #10814 , which might be what made the default options disappear from the extension's 'option' UI. Should we revert #10829 and #10814 if we go the revert route?

iwannabethedev:
I also wonder whether adding (more?) automated tests to this part of the code base might be helpful.

iwannabethedev:
I believe the most recent change fixes #10844 .

web/packages/extension/src/popup.ts
```diff
@@ -83,28 +83,48 @@ async function queryTabStatus(
...
 }
 
-function objectsEqual<T>(x: NonNullable<T>, y: NonNullable<T>) {
-    for (const [key, value] of Object.entries(x)) {
-        if (y[key as keyof T] !== value) {
-            return false;
+/**
+ * Should only be called on data type objects without any "cyclic members" to avoid infinite recursion.
+ */
+function dataEqual(x: unknown, y: unknown): boolean {
```
relrelb, line 89: Let's rename this function to `deepEqual()`, to better reflect its meaning.

web/packages/extension/src/popup.ts
```diff
@@ -83,28 +83,48 @@ async function queryTabStatus(
...
+    if (
+        typeof x === "object" &&
+        typeof y === "object" &&
+        x !== null &&
+        y !== null
+    ) {
+        // Two non-null objects.
+
+        for (const [key, value] of Object.entries(x)) {
+            const value2 = y[key as keyof typeof y];
```
relrelb, line 99: For slightly more compact code, I prefer to "inline" `value2` directly into the `dataEqual()` call:
```ts
if (!dataEqual(value, y[key as keyof typeof y])) {
    return false;
}
```
(here and right below)

relrelb:
Going to merge this. Thanks again!

PR #10856, opened 2023-04-27
extension: Fix options-related bug

### LINKED ISSUE #10844: Relative URLs in extension page are parsed relative to the extension page

relrelb:
With #10829, `DEFAULT_OPTIONS` contained all config properties, making it distinct (by `objectsEqual()` in `popup.ts`) from any other `Options` instance, due to each one having a unique `parameters` property. This caused the "Reload tab to apply changes" appear unconditionally in the popup page.

Moreover, it caused #10844 because the `base` config was always overridden to `null` in `player.ts`.

As a fix, move `CORE_DEFAULT_CONFIG` away from `DEFAULT_OPTIONS`, using it just for initial options binding in `common.ts`. This restores the state before #10829, where only extension-specific options are saved, but still not-yet-set options are bound to their default value, preventing empty selection boxes in the options page.

Supersedes #10850.
Fixes #10844.

### DIFF
--- a/web/packages/extension/src/common.ts
+++ b/web/packages/extension/src/common.ts
@@ -1,5 +1,6 @@
 import * as utils from "./utils";
 import type { BaseLoadOptions, Duration } from "ruffle-core";
+import { DEFAULT_CONFIG as CORE_DEFAULT_CONFIG } from "ruffle-core";
 
 export interface Options extends BaseLoadOptions {
     ruffleEnable: boolean;
@@ -124,7 +125,8 @@ export async function bindOptions(
 
     for (const [key, element] of elements.entries()) {
         // Bind initial value.
-        element.value = options[key];
+        element.value =
+            options[key] ?? CORE_DEFAULT_CONFIG[key as keyof BaseLoadOptions];
 
         // Prevent transition on load.
         // Method from https://stackoverflow.com/questions/11131875.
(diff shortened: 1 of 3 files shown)


iwannabethedev:
I believe the changes that caused these bugs and others are #10829 and #10814 .

iwannabethedev:
With the most recent change in #10850 , I would guess that this PR no longer supersedes it. Though, I do not have a good feeling for which of this PR and #10850 is the best one to go with. @n0samu also mentioned that given the changes (I believe in https://github.com/ruffle-rs/ruffle/pull/10829 and https://github.com/ruffle-rs/ruffle/pull/10814 ) that break the extension, that a revert of those PRs might make sense to ensure that the extension is not broken while we figure things out.

relrelb:
Closing in favor to #10850.

Issue #1108, opened 2020-09-04
Buttons are sometimes difficult to click after a goto


Herschel:
Imagine this frame sequence:
Frame 1: Button placed with rollOver/release handlers
Frame 2: Empty frame with no button, with gotoAndPlay(1) code

If the user hovers over the button, when the button is destroyed and recreated by the goto, Flash does not fire the rollover event a second time. The button remains clickable in its hover state.

However, a movie clip will act more as expected: it will fire its rollover event a second time as it gets recreated. It is also difficult to click (have to click and release faster than the movie framerate).

In Ruffle, buttons act like the movie clip case, becoming unclickable and firing their events twice. Some research is needed because I'm unsure of the exact Flash behavior -- the button should indeed get recreated, because if you put an MC with actions inside the button, those actions run twice as the button and MC get recreated. But I'm not sure how Flash decides not to fire the rollover events a second time.

The play button in Supersoldier (https://www.newgrounds.com/portal/view/284305) is very unresponsive because of this (cc #177). You can sometimes click it by mashing very quickly.

Minimal sample, with both an MC and a button:
[button-recreate.zip](https://github.com/ruffle-rs/ruffle/files/5172211/button-recreate.zip)

michiel2005:
We are currently recreating the button with the run_goto() in display object -> movie_clip.rs.
But we're not saving the state of the button before the goto. instead we create the button again like it is totally new. I am pretty sure that when we save the button state it is currently in, like clicked. And then recreate the button with that state. it will work.
Since OverDownToOverUp for example does not get called because the button is not in it's clicked state anymore after one goto has passed.

n0samu:
This bug also makes it extremely difficult to click the "Start" button in *Courage the Cowardly Dog: Nightmare Vacation*. (Pressing the spacebar to start the game works fine, at least.)
This is a multiasset game, so the easiest way to test it is to load it via the SWF URL in the Ruffle extension: https://www.getgames.io/games/swf/courage-the-cowardly-dog-nightmare-vacantion/game.swf

Toad06:
Here are [some samples](https://github.com/ruffle-rs/ruffle/files/11415130/_samples.zip) (edited with JPEXS) based on the file available in #4402. Epilepsy warning.

---

fileA (rollOver triggered once):
- frame 1: PlaceObject2 button
- frame 3: RemoveObject2 button + gotoAndPlay(1)

---

fileB (rollOver triggered as long as the mouse hovers the button):
[...]

Issue #10789, opened 2023-04-23
Zombie Rush: No response when clicking the Play button


Jmousy:
### Describe the bug

[zombie_rush.zip](https://github.com/ruffle-rs/ruffle/files/11302999/zombie_rush.zip)

When I click the 'Play' button in the **'Zombie Rush'** game, it is unresponsive and unable to start the game.
AVM1 & Checked on Ruffle 2023-04-22 Nightly.

![image](https://user-images.githubusercontent.com/75052171/233824273-52f2c730-148d-4162-84f5-198806d551a5.png)

### Expected behavior

When you click the Play button, the game should start.

### Affected platform

Self-hosted version

### Operating system

Windows 10 22H2

### Browser

Google Chrome 111

n0samu:
This happens because the game is quickly jumping between frames 94 and 95, and the actual clickable button (DefineButton2 37) is only placed on frame 95. This is yet another instance of #1108.

n0samu:
Fixed by #10862

Issue #9966, opened 2023-03-10
Super Soldier: Play button doesn't start game.


TomFulp:
### Describe the bug

https://www.newgrounds.com/portal/view/284305/format/flash?emulate=flash

Clicking the PLAY button at the top of the Super Soldier (AS2) menu doesn't advance to the game, you remain stuck on the menu.

### Expected behavior

Clicking the PLAY button should start the game.

### Affected platform

Self-hosted version

### Operating system

Windows 11

### Browser

Chrome version Version 110.0.5481.178

adrian17:
Known as https://github.com/ruffle-rs/ruffle/issues/1108 :/

n0samu:
Fixed by #10862

Issue #8482, opened 2022-11-05
Wrath II - Various UI buttons require several/random amount of clicks to function


Nicholas-Steel:
### Describe the bug

https://www.youtube.com/watch?v=BkFxhp3gs-8

The Scroll/Parchment UI seen when clicking Easy Game/Hard Game, between levels and when clicking the "i" button, has buttons that require a variable amount of mouse clicks to register that I'm clicking them. You can see & hear this in the video.

### Expected behavior

These UI elements should only need a single mouse click to register.

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Latest stable Firefox

### Additional information

Game is loaded directly, not embedded in a website.
http://home.exetel.com.au/franpa/Games/Wrath_II.swf

n0samu:
This might be related to #1108.

n0samu:
Fixed by #10862

PR #10862, opened 2023-04-28
avm1: fix buttons that were not clickable after a goto

michiel2005:
It is my first commit in this repo were I changed a serious amount of code.
This fixes issue #1108 and all underlying issues:
#8482, #8293, #9966, #10789, #7903, #7170, #6261, #4402
This is being done by placing removed displayObjects into a BTreeMap so we can use them again if there is a goto.
When the next frame begins the BTreeMap is emptied.


core/src/avm1/scope.rs
```diff
@@ -194,6 +194,7 @@ impl<'gc> Scope<'gc> {
             // Traverse the scope chain in search of the value.
             scope.set(name, value, activation)
         } else {
+            println!("it's the error");
```
Lord-McSweeney, line 197: Accidental insertion?

michiel2005: yes, was not meant to :P.

michiel2005:
I found that dragOut is not yet working after a goto, working on it.

michiel2005:
This can be reviewed now 👍 .

core/src/display_object.rs
```diff
@@ -1522,6 +1522,11 @@ pub trait TDisplayObject<'gc>:
     fn as_stage(&self) -> Option<Stage<'gc>> {
         None
     }
+    fn state(&self) -> Option<ButtonState> {
```
Lord-McSweeney, line 1525: nit: These functions should probably be placed higher up (around line 1322).

michiel2005: Fixed now, Thank you :)

adrian17:
Pinging @kmeisthax / @Herschel  , to me this feels like a hack that happens to work but isn't really how it's supposed to work under the hood - and doesn't feel like it matches Mike's original explanation from https://github.com/ruffle-rs/ruffle/issues/1108 either? I can't say with certainty, I just have a bad hunch about it.

Also, we definitely want some extra tests for this.

swf/src/types.rs
```diff
@@ -302,6 +302,7 @@ pub struct PlaceObject<'a> {
     pub version: u8,
     pub action: PlaceObjectAction,
     pub depth: Depth,
+    pub id: Option<CharacterId>,
```
kmeisthax, line 305: Isn't the character ID already part of `PlaceObjectAction`? I don't have my laptop open to check right now, but I could swear that was the case...

kmeisthax (commented):
So, the approach being taken here is to grab buttons out of the display tree but restore them if we need to recreate them. Mike explains this in terms of events being fired or not fired, but it doesn't have to be literally that. On the other hand, holding old buttons so we can bring them back does seem hacky.

Toad06's tests from issue #1108 would probably solve this dilemma. I don't see them incorporated into this PR - it might be helpful to see if they pass with these changes or not.

swf/src/types.rs
michiel2005: You were totally right. Tank you.

michiel2005:
I'm still a student so for any reviewers please tell me what I could have done better so I can learn from it.

iwannabethedev:
> I'm still a student so for any reviewers please tell me what I could have done better so I can learn from it.

I have not looked at the PR thoroughly, and I do not understand this part of the codebase, but, judging from review comments, it may be

This might also not be the best first PR (apart from learning from it, which is very good as a student), since it touches multiple parts of an existing, somewhat complicated codebase, and may also require both knowledge of that specific codebase and also software development and design, resource management, etc.

For this PR, there may be several aspects of it that may not be good or that may be uncertain. For this kind of PR, the codebase may be so large and complex that figuring all out may not be viable (also, the quality requirements for the application are not very high, there are after all not lives at risk for applications running on Ruffle or millions of dollars at stake if a bug occurs, like there for instance may be at for instance NASA, certain software in a bank, software for a pacemaker, the braking system of a car, etc. Therefore, it is not a general requirement that all aspects and parts of the system are understood thoroughly, since priorities are different). There is a bit of risk management here as well - what consequences might the changes of the PR have?

I do not know how you arrived at the given solution, but, if you did trial-and-error and just made changes that seemed to work without considering at least a bit of what the consequences might be, that is understandable but less good (though if you are a student, you cannot be expected to both know this codebase in depth, or have a good general grasp on how to figure out, analyze, mitigate, risk handle, and guess and estimate what the consequences might be of changes). UI systems, that this touches upon, can be rather complex and have multiple interacting parts, and have the inherent complicating aspect that external events can happen arbitrarily during execution (typically user input), in contrast to for instance in multiple aspects much simpler batch execution where all user input is given at the start of execution. There is also the aspect of resource management, life cycles and such - C++ has RAII, and Rust also has life-cycle management as a major part of the language (though I do not really know Rust) - that said, life-cycle management and resource management is in some ways relevant for programming tasks in general and is for many tasks necessary whether or not the programming language used has support and abstractions for handling and managing it more easily. One aspect that is important for larger projects like Ruffle is to ensure that the design and implementation is decent, also in the long term - otherwise, there might be consequences like a design or implementation being bad, other code over time building upon that bad design and implementation, and if problems are then discovered down the road, you might end up having to change a lot of code later, possibly having to change considerably more code than just the original bad changes. This can also create somewhat of an aversion to change, which is a problem, but there are different approaches to decrease the cost and risk of changes and the frequency issues because of changes - such as code review, communicating with other developers for figuring out approaches and designs, etc. But there are many aspects that are relative to the project in question, and programming problems are in many cases so complex that you cannot be expected to analyze everything throughout - some analysis can still go a long way, and again there is risk management and considering the potential consequences in each case, and also considering that there may be aspects and issues that you (and possibly others) might not have considered or know of.

I think you can learn something from both the comments here if you ponder and reflect upon them, and I would guess that you already have learned something from making the changes, the PR and the tests. If you would like to get the experience with also getting a PR merged, I think looking for an easier task might be better - if you want it also to be somewhat "meaty" with challenge in it, I would guess that a more isolated part of the system would be a good task. There are some issues that are designated "Good first issue": https://github.com/ruffle-rs/ruffle/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22 , which might be a good option (though I do not know). For this codebase, I would guess that implementing some of the API functionality (for AVM1, there are tracking issues for API functionality, like https://github.com/ruffle-rs/ruffle/issues/290 and https://github.com/ruffle-rs/ruffle/issues/15 . For AVM2, you can see https://ruffle.rs/avm2.html to see missing API functionality (do note that it may lag behind a couple of weeks since it is not currently automatically updated). On the other hand, you might be able to continue on with this and say that, you will put this PR into draft-mode, and then talk and learn about the possible issues, ponder, experiment and communicate with the main developers (Ruffle's Discord is a good place to discuss development issues, find the right channel), and then for instance open a new PR, one that might scavenge parts of this PR. For more complicated codebases and code that interacts with multiple different parts, communication with other developers can be very valuable for both oneself, others and the project in question. Which makes sense - complicated issues can take multiple people to figure out.

(thank you for reading my blog spam :smiley: ).

michiel2005:
Hey, iwannabethedev. At first I want to thank you for writing this. Really apreciate it.
This is how I came to this solution:
Months ago I saw this issue got posted: DR3i: "Clicking "Play next" doesn't proceed to next level" (8293), and I was interested in fixing it. At first I had no idea how Ruffle nor could I even write a hello world program in rust so I started to learn a bit of Rust and get a rough understanding of Ruffle. At first I tried to fix it by trail and error, that didn't work. I looked true the swf's under 1108 and then the problem clicked and why it happened. There were two problems, after the button got destroyed it would also be removed from the mouse down object and mouse hovered object. A other problem was when the button got recreated, the visual state would reset. At first I fixed it really hacky with the Name, that was not such a good idea after that I did it with the depth. It worked. But was still kind of ugly. Toad06 posted a few modified swf's as tests for this problem and then it clicked how I think it should all be done. I implemented it that way, removed the HashMap were I kept buttons when they were destroyed. And that is were my code is now.
I prefer discord. Can you please contact me there? I am Michiel#7283

P.S. This PR can be reviewed

iwannabethedev:
That sounds good, if you have made extensive changes, I can try to take a look at it, but I think adrian17 and kmeisthax should also look at it at some point if you have made extensive changes.

kmeisthax (commented):
It appears the new setup for this now is to check if the same object got reinstantiated; this seems viable. I'd approve and merge, but I did notice one flaw with how this is implemented: `CharacterId` isn't globally viable; you need to check both what character ID a display object was instantiated from *and* what movie that character came from. If you load two movies into the same stage, they both can have a "button 5", and this will falsely skip rollovers for them.

(If Flash actually *doesn't* check the source movie somehow, and multi-movie loads allow buttons from different movies to substitute for one another, ignore the following comments and instead add a `NOTE` comment documenting this.)

core/src/player.rs
```diff
@@ -1252,15 +1252,48 @@ impl Player {
...
 
+            let mut new_over_object_updated = false;
             // Cancel hover if an object is removed from the stage.
             if let Some(hovered) = context.mouse_over_object {
                 if !context.is_action_script_3() && hovered.as_displayobject().avm1_removed() {
                     context.mouse_over_object = None;
+                    if let Some(new_object) = new_over_object {
+                        if new_object.as_displayobject().id() == hovered.as_displayobject().id()
+                            && new_object.as_displayobject().depth()
+                                == hovered.as_displayobject().depth()
```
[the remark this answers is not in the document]
michiel2005: You were right, nice catch

core/src/player.rs
```diff
@@ -1441,6 +1489,29 @@ impl Player {
         needs_render
     }
 
+    ///This searches for a display object by it's id
+    //TODO: is there a better place to place next two functions
+    fn find_first_character_instance<'gc>(
+        obj: DisplayObject<'gc>,
+        character_id: &CharacterId,
+        movie_url: &str,
```
kmeisthax, line 1497: Likewise this should take an `Arc<SwfMovie>` instead of a `&str`.

core/src/player.rs
```diff
@@ -1252,15 +1252,54 @@ impl Player {
...
             // Cancel hover if an object is removed from the stage.
             if let Some(hovered) = context.mouse_over_object {
                 if !context.is_action_script_3() && hovered.as_displayobject().avm1_removed() {
                     context.mouse_over_object = None;
+                    if let Some(new_object) = new_over_object {
+                        if new_object.as_displayobject().id() == hovered.as_displayobject().id()
+                            && new_object.as_displayobject().depth()
+                                == hovered.as_displayobject().depth()
+                            && new_object.as_displayobject().movie().url()
+                                == hovered.as_displayobject().movie().url()
```
kmeisthax, line 1265: You'll want to use pointer equality instead of checking the URL, since it's possible there's multiple different movies loaded from the same URL. The syntax for actually checking that is `Arc::ptr_eq(&new_object.as_displayobject().movie(), &hovered.as_displayobject().movie())`.

kmeisthax (commented):
One more thing: you need to compare `SwfMovie`s by pointer instead of URL, since you can have two different movies with the same source URL.

Toad06:
> Toad06's tests from issue #1108 would probably solve this dilemma. I don't see them incorporated into this PR - it might be helpful to see if they pass with these changes or not.

@kmeisthax These tests are edited with JPEXS and based on the file available in #4402 (as I don't have a Flash editor to make them properly). I only made them for quick manual testing purposes, so I think we shouldn't add them as part of the Ruffle test suite (each file is about 2mb and includes data that will not be relevant here - without mentioning potential copyright issues).

kmeisthax (approved):
LGTM! I'm going to merge.

@Toad06 Gotcha on the tests, the new version of this PR has enough as is that I don't think we'll need yours anyway.

michiel2005:
The tests that Toad06 provided are actually in the tests right now.

core/src/player.rs
```diff
@@ -1441,6 +1494,27 @@ impl Player {
...
 
+    ///This searches for a display object by it's id
+    //TODO: is there a better place to place next two functions
+    fn find_first_character_instance<'gc>(
+        obj: DisplayObject<'gc>,
+        character_id: &CharacterId,
+    ) -> Option<DisplayObject<'gc>> {
+        if let Some(parent) = obj.as_container() {
+            for child in parent.iter_render_list() {
+                if &child.id() == character_id && Arc::ptr_eq(&child.movie(), &obj.movie()) {
```
kmeisthax, line 1505: While this is clever, it won't work, because we specifically need to check `child`'s movie against the movie that defined `character_id`, not the movie that defined it's parent. As it stands this code will break on the following display tree configuration:

 * Level 0 / root movie clip (Movie A, Character 0)
     * Button A (Movie A, Character 1)
     * Child movie's root clip (Movie B, Character 0)
       * Button B (Movie B, Character 1)

Calling `find_first_character_instance` looking for Button B will instead return Button A, because:

 * Button A and Level 0 are both defined in Movie A
 * Button A is character ID 1

You specifically need to provide the movie that defined the character ID you're interested in here as a separate function parameter.

kmeisthax (commented):
So... it appeared the rebase failed and @adrian17 tried to clean it up. In Discord, he had a question I couldn't exactly answer: why do we need to run `find_first_character_instance` on `down_object`?

I'm assuming there's some subtlety of how buttons survive gotos that requires searching the whole display tree for a matching copy of the object, but I can't exactly prove that. Furthermore, the `new_over_object` code doesn't actually do this check, it just checks the `id`, `depth`, and `movie` for equality and then copies the state if it matches. Is there any particular reason why `mouse_down_object` can't work the same way?

Also, I noticed a bug in `find_first_character_instance` while trying to answer Adrian's question.

michiel2005:
@kmeisthax we are seraching for the object because of this: when we are holding the button down and moving out of the button area. We should keep the button in it's mouse down state, so we need to check if the object is still there. We also can't check if the over object points at the button since the over object should remove it's reference when not hovered anymore.

michiel2005:
This pr is ready to be reviewed.

core/src/display_object.rs
```diff
@@ -1319,7 +1319,11 @@ pub trait TDisplayObject<'gc>:
     fn set_has_explicit_name(&self, gc_context: MutationContext<'gc, '_>, value: bool) {
         self.base_mut(gc_context).set_has_explicit_name(value);
     }
-
+    fn state(&self) -> Option<ButtonState> {
+        None
+    }
+    fn set_state(self, _context: &mut crate::context::UpdateContext<'_, 'gc>, _state: ButtonState) {
```
adrian17, line 1325: this `crate::context` shouldn't be necessary

core/src/player.rs
```diff
@@ -1252,15 +1252,59 @@ impl Player {
...
+                    if let Some(root_clip) = context.stage.root_clip() {
+                        display_object = Self::find_first_character_instance(
+                            root_clip,
+                            down_object.as_displayobject(),
+                        );
+                    }
+
+                    if let Some(new_object) = display_object {
+                        if new_object.id() == down_object.as_displayobject().id()
+                            && new_object.depth() == new_object.depth()
```
adrian17, line 1292: looks like a copy-paste error

core/src/player.rs
```diff
@@ -1252,15 +1252,59 @@ impl Player {
...
+
+                    if let Some(root_clip) = context.stage.root_clip() {
+                        display_object = Self::find_first_character_instance(
+                            root_clip,
+                            down_object.as_displayobject(),
+                        );
+                    }
+
+                    if let Some(new_object) = display_object {
+                        if new_object.id() == down_object.as_displayobject().id()
```
adrian17, line 1291: if `find_first_character_instance` can only return a matching DisplayObject, then are these conditions here still needed?

core/src/player.rs
```diff
@@ -1252,15 +1252,59 @@ impl Player {
                 ));
             }
 
+            let mut new_over_object_updated = false;
             // Cancel hover if an object is removed from the stage.
             if let Some(hovered) = context.mouse_over_object {
                 if !context.is_action_script_3() && hovered.as_displayobject().avm1_removed() {
                     context.mouse_over_object = None;
+                    if let Some(new_object) = new_over_object {
+                        if new_object.as_displayobject().id() == hovered.as_displayobject().id()
```
adrian17, line 1261: if these conditions "id, depth and movie match" are duplicated, we might wanna reuse these as some function that checks for "kinda-sorta-equivalence" of displayobjects

core/src/player.rs
michiel2005: I don't see why it looks like a copy paste error, changed the name new_object to new_down_object anyways.

core/src/player.rs
michiel2005: Added a function to checck for the same depth and id, but not to check for the same movie because that is only be done in one place.

core/src/player.rs
```diff
@@ -1252,15 +1252,57 @@ impl Player {
...
+
+                    if let Some(root_clip) = context.stage.root_clip() {
+                        display_object = Self::find_first_character_instance(
+                            root_clip,
+                            down_object.as_displayobject(),
+                        );
+                    }
+
+                    if let Some(new_down_object) = display_object {
+                        if new_down_object.depth() == new_down_object.depth()
```
Lord-McSweeney, line 1290: I think the condition `new_down_object.depth() == new_down_object.depth()` will always be true, did you mean `new_down_object.depth() == down_object.depth()`?

michiel2005: You were totally right. stupid mistake.

n0samu (approved):
I'm not experienced with reviewing PRs, but I read through the diff and tested it out, and everything is working well, plus the code looks sensible and fine. I think it's time to merge this! And thank you for your work!

core/src/player.rs
```diff
@@ -1364,10 +1398,18 @@ impl Player {
...
-                    let released_inside = InteractiveObject::option_ptr_eq(
+                    let mut released_inside = InteractiveObject::option_ptr_eq(
                         context.mouse_down_object,
                         context.mouse_over_object,
                     );
+                    if let Some(down) = context.mouse_down_object {
+                        if let Some(over) = context.mouse_over_object {
+                            if !released_inside {
+                                released_inside =
+                                    down.as_displayobject().id() == over.as_displayobject().id();
```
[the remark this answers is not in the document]
Lord-McSweeney: This still doesn't check `.movie`?

core/src/player.rs
```diff
@@ -1364,10 +1407,18 @@ impl Player {
...
```
Lord-McSweeney, line 1418: Shouldn't this check `.movie()`?

core/src/player.rs
```diff
@@ -1441,6 +1492,35 @@ impl Player {
         needs_render
     }
 
+    //Checks if two displayObjects have the same depth and id
+    fn check_display_object_equality(object1: DisplayObject, object2: DisplayObject) -> bool {
+        object1.depth() == object2.depth() && object1.id() == object2.id()
```
Lord-McSweeney, line 1497: This function should probably also check the movie, since everywhere this function is called, the movie is checked anyway.

kmeisthax: Almost certainly yes, and it'd also fix your other comment about `movie()` not being checked

michiel2005: Thanks for the feedback, I am away for a few days. But will fix them when I'm back.

michiel2005: Fixed those review comments and fixed some minor mistakes I made earlier. It can be reviewed again.

kmeisthax (approved):

PR #10864, opened 2023-04-28
avm2: Alternative implementation of XML hasOwnProperty @attribute

evilpie:
This is an alternative to #10695 without actually adding XML.prototype.hasOwnProperty and instead changing Object.prototype.hasOwnProperty to support `@attributes`.


Aaron1011 (approved):

core/src/avm2/e4x.rs
```diff
@@ -715,3 +715,30 @@ pub fn to_xml_string<'gc>(
...
+) -> Result<Multiname<'gc>, Error<'gc>> {
+    if let Value::Object(o) = name {
+        if let Some(qname) = o.as_qname_object() {
+            return Ok(qname.name().clone());
+        }
+    }
+
+    let name = name.coerce_to_string(activation)?;
+
+    if name.starts_with(b'@') {
```
relrelb, line 731: nit: `name.strip_prefix('@')` can be used here, avoiding the slice `name[1..]` right after.

evilpie: Nice. I didn't know of this method before.

PR #10865, opened 2023-04-28
avm1: `button.enabled = undefined` doesn't disable the button

### LINKED ISSUE #10788: Speedy Bubbles: Unable to click the Select Game button after a short period of time
### Describe the bug

In **'Speedy Bubbles'**(AVM1), when the game starts, the four game types appear one by one with an animation effect.
Clicking on each button should either bring up the demo version instructions or start the game.

However, in Ruffle (checked on 2023-04-22), all four buttons are unresponsive when clicked after the animation effect is complete. If you quickly click the button before the animation is complete, you can start the game.

The appearance of the mouse cursor is clickable, then changes to the default state.

image

If you click the "Options" or "Help" button in the bottom right corner, close the window, and then click any of the four buttons again, the issue should be resolved and you should be able to start playing.

[speedy_bubbles.zip](https://github.com/ruffle-rs/ruffle/files/11302845/speedy_bubbles.zip)

### Expected behavior

The four game buttons should be clicked at any time.

### Affected platform

Self-hosted version

### Operating system

Windows 10 22H2

### Browser

Google Chrome 111

### Additional information

_No response_

Toad06:
Fixes #10788. The file does `button.enabled = how;` but the variable `how` is undefined (not set before reading it). For some reason, Flash Player treats undefined as a truthy value when dealing with the `enabled` property.

This PR is a small fix, as `enabled` will likely have to get refactored:
- The property should return any value set by the user. This means that a code like `button.enabled = undefined; if(!button.enabled) { trace(button.enabled); }` will trace "undefined" though the button is enabled internally.
- `delete button.enabled` doesn't delete the property but changes its value to true (its default value), so the button gets enabled. This is another strange behavior that will have to get investigated (maybe using `delete` on properties that contain the `DONT_DELETE` flag should restore the default value?).

I added a test (currently ignored) that cover the cases I mentioned.


relrelb:
Looks like `enabled` behaves much like `focusEnabled`, as described in https://github.com/ruffle-rs/ruffle/issues/10573#issuecomment-1498500673:
```as
trace(MovieClip.prototype.hasOwnProperty("enabled")); // true
trace(someClip.hasOwnProperty("enabled")); // false, unless you had set it
```

So I guess the fix should be similar to #10597. The only difference I noticed between `enabled` and `focusEnabled` is that initially, `MovieClip.prototype.enabled = true`, whereas `MovieClip.prototype.focusEnabled` isn't even defined. This can be accomplished by modifying the definition of `enabled` in `PROTO_DECLS`:
```rs
"enabled" => bool(true);
```

Toad06:
Thanks for the feedback and help. This should be addressed now.

- I renamed the "enabled" field (`MovieClipData` struct) to "avm2_enabled" to improve clarity.

- While testing, I noticed that properties defined on buttons were enumerable (at least those that we currently implement), so I changed them while I was at it.

- The test I wrote is no longer ignored and passes.

- Other boolean properties are affected (`useHandCursor` and others that we don't implement yet). I suppose `fn avm1_enabled` could be turned into something more generic, by passing the name of the property as an argument.

relrelb (approved):
Looks good to me. Thank you!

Toad06:
Thank you for the review.

Issue #7903, opened 2022-09-07
cursor is stuck with the slider


zjk261:
### Describe the bug

![GIF3](https://user-images.githubusercontent.com/98996990/188764207-821ee540-df4e-4c2d-ac56-c6e1bc3ddae1.gif)
[blackbody.zip](https://github.com/ruffle-rs/ruffle/files/9501723/blackbody.zip)

### Expected behavior

![GIF4](https://user-images.githubusercontent.com/98996990/188764425-328c31ea-ecbe-4afe-82c1-4cf52382d2ba.gif)

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

![image](https://user-images.githubusercontent.com/98996990/188763937-2f94d18a-7580-460f-b1ee-7831a1959612.png)

n0samu:
Seems like the `EndDrag` action attached to the button is somehow not taking effect. I'm not sure why.

hatal175:
OK, so the issue is that the button being dragged keeps getting loaded and unloaded (id 32) constantly. After it gets unloaded the mouse update process removes the mouse_down_object since it is removed and then the release event has nothing to callback to.

You can actually make it work if you click fast enough so the button doesn't manage to unload.

Unfortunately, I'm not familiar enough with flash internals and frame behaviors to say what exactly is wrong in ruffle's implementation.

hatal175:
Maybe related to  #1108

n0samu:
Fixed by #10862

Issue #7170, opened 2022-06-06
[Breakdown at Santaland] Pipe puzzle doesn't progress when clicked


Jmousy:
### Describe the bug

In **Breakdown at Santaland**, the pipe puzzle doesn't work.

[Breakdown at Santaland.zip](https://github.com/ruffle-rs/ruffle/files/8842854/Breakdown.at.Santaland.zip)

Comparing the two videos below, in Ruffle (checked in Nightly 2022-06-05), if you click on any part, the sound effect is output, but the part does not move.

Based on the screen displayed in Adobe Flash Player, even if you progress through the puzzles in the order of 33-11-777-88-55, it seems that you cannot proceed to the next, so this is not likely to be a graphic display problem. (maybe)

### Adobe Flash Player

https://user-images.githubusercontent.com/75052171/172122133-ac1b50aa-17a2-4f2d-8f72-0cd271a33d66.mp4

### Ruffle Nightly 2022-06-05

https://user-images.githubusercontent.com/75052171/172122151-dfcbcd14-b250-441e-96fd-d5340385e868.mp4

### Expected behavior

The puzzle should proceed normally

### Affected platform

Self-hosted version

### Operating system

Windows 10 21H2

### Browser

Google Chrome 102

Toad06:
Thanks for the report.

I think this is the same issue as #1108: the fact that the mouse cursor changes from the pointer to the default one when the mouse button is down on a pipe is a good indicator (the button is deleted, which shouldn't happen).

It's still possible to interact with the pipes in the current build but you have to be super fast (clicking and releasing immediately)... Well, that's a challenge. 😄

michiel2005:
At the newest built there is a new bug were the pieces of the card in the first puzzle are disappearing when clicked on them.

michiel2005:
The bug I reported two days ago is gone, that is great. But I found that when clicking on the maze pieces the releaseOutside is being called in Ruffle even when the mouse gets up inside the button. I don't know if this is the right behaviour but I found it very strange.

n0samu:
Fixed by #10862

PR #10866, opened 2023-04-28
swf: Introduce `PointDelta`

### LINKED ISSUE #817: web: drag of cube is flickering
In Flash, you can drag the "1 kg air cube" smoothly. Only the selected values are updated.
(On the right side, quiz answers can be picked.)

In ruffle the drag also works, but stutters. All values are shown, selection is not possible.
(On the right side, answering the quiz is not possible.)

See http://www.webgeo.de/k_230_en/#3
(on http://www.webgeo.de you can switch between using ruffle or using flash)

Version: ruffle_web_2020-06-02
flash_k_230_en3
ruffle_k_230_en3
[k_230_03.zip](https://github.com/ruffle-rs/ruffle/files/4878746/k_230_03.zip)

relrelb:
Generally, when transforming a difference between two points, `p1` and `p2`, with a matrix `m`, we would like the following property to hold:

```
m * (p1 - p2) == m * p1 - m * p2
```

Unfortunately, it wasn't like this before, because matrices have a translation component, which is non-linear. In `m * p1 - m * p2`, the translations of `m * p1` and `m * p2` are the same and therefore cancel out each other. However, in `m * (p1 - p2)` the translation stays.

In order to preserve this property, introduce a new `PointDelta` type which is not subject to translation when transformed by a matrix.

For now, the following operations are supported:

* `Point - Point -> PointDelta`
* `Point + PointDelta -> Point`
* `Point += PointDelta`
* `Point - PointDelta -> Point`
* `Point -= PointDelta`

As a consequence, the expression `position + global_to_local_matrix * mouse_delta` in `update_drag()` now ignores translation, which fixes #817.

### DIFF
--- a/render/canvas/src/lib.rs
+++ b/render/canvas/src/lib.rs
@@ -962,8 +962,8 @@ fn create_linear_gradient(
         let end = matrix * Point::new(Twips::new(16384), Twips::ZERO);
         // If we have to scale the gradient due to spread mode, scale the endpoints away from the center.
         let delta = end - start;
-        let dx = 0.5 * (gradient_scale - 1.0) * delta.x.to_pixels();
-        let dy = 0.5 * (gradient_scale - 1.0) * delta.y.to_pixels();
+        let dx = 0.5 * (gradient_scale - 1.0) * delta.dx.to_pixels();
+        let dy = 0.5 * (gradient_scale - 1.0) * delta.dy.to_pixels();
         Ok(context.create_linear_gradient(
             start.x.to_pixels() - dx,
             start.y.to_pixels() - dy,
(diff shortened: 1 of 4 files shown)


Lord-McSweeney:
Is there any way to add a test for this?

relrelb:
> Is there any way to add a test for this?

Good point. I've expanded existing unit tests in `render/src/matrix.rs` to cover `PointDelta`s.

Toad06:
This will fix #10872 (as you already mentioned) as well as https://github.com/ruffle-rs/ruffle/issues/7170#issuecomment-1528107874 and https://github.com/ruffle-rs/ruffle/issues/7903#issuecomment-1528047513. I think it could be nice to merge this PR before the next nightly build.

PR #10868, opened 2023-04-28
avm2: Don't overwrite existing entry in `export_class`

Aaron1011:
This makes the behavior consistent with `export_definition`


relrelb (approved):

Lord-McSweeney:
Can you add a test for this?

core/src/avm2/domain.rs
```diff
@@ -245,6 +245,9 @@ impl<'gc> Domain<'gc> {
     }
 
     pub fn export_class(&self, class: GcCell<'gc, Class<'gc>>, mc: MutationContext<'gc, '_>) {
```
iwannabethedev, line 247: (Nit: Maybe refactor the key value into its own variable, like:

```rust
        let class_name = class.read().name();
        if self.0.read().classes.contains_key(class_name) {
            return;
        }
        self.0.write(mc).classes.insert(class_name, class);
```

?)

core/src/avm2/domain.rs
```diff
@@ -245,6 +245,9 @@ impl<'gc> Domain<'gc> {
...
+        if self.0.read().classes.contains_key(class.read().name()) {
+            return;
```
iwannabethedev, line 249: I do not know ActionScript, and I do not quite understand this code either (not because it is unclear, the code seems clear, but because I do not know much about this part of the codebase), but, if I had to guess, if we "include" a class/type (like `MyDuplicate`), and we then "include" it some other way (using for instance `Loader.load()`), we do not want to "include" what is the same class again, since it has already been "included", correct? If so, the change seems good. Though, is `name()` qualified with a package path (if that makes sense)? Like, if I had two different types `package1.MyDuplicate` and `package2.MyDuplicate`, would it handle that correctly? But I do not know ActionScript and I do not quite understand this part of the codebase.

iwannabethedev:
I do not understand this part of the codebase, and I know very little about ActionScript, so someone else should review it as well, I think.

core/src/avm2/domain.rs
relrelb: I think it's readable enough as-is.

core/src/avm2/domain.rs
```diff
@@ -245,6 +245,9 @@ impl<'gc> Domain<'gc> {
     }
```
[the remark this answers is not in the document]
relrelb: Actually that would be nice. Since I want to merge this PR already, would you like you add such a comment in a separate PR?

core/src/avm2/domain.rs
relrelb: You can see how the fix is relevant in the recently added test. Nothing actionable in this thread, so resolving.

relrelb (approved):

core/src/avm2/domain.rs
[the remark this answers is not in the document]
iwannabethedev: That sounds alright, though I am not exactly certain what it should say :smile: .

Maybe:

```rust
/// Export a class from a script into the current application domain.
```
? But I would feel more confident if someone who actually knows this part of the codebase made (or approved of) the comment.

core/src/avm2/domain.rs
iwannabethedev: That is not actually what I am asking about, sorry if I am being unclear. I did see the test, and I am referring to it in the comment. What I first asked about was if my understanding of the changes (and the test) is correct, and then I asked if a certain corner case is handled correctly. Which I am still in doubt about - do you know if `name()`  is qualified with a package path (if that makes sense)?

PR #10869, opened 2023-04-28
avm2: Implement URLRequest.requestHeaders

Aaron1011:
These are directly set on the underlying navigator's HTTP request type, and get printed out in our test navigator backend.

No validation of the header names is performed - on web, this will be enforced by the browser.


Dinnerbone (approved):

core/src/backend/navigator.rs
```diff
@@ -111,6 +121,14 @@ impl Request {
     pub fn set_body(&mut self, body: (Vec<u8>, String)) {
         self.body = Some(body);
     }
+
+    pub fn headers(&self) -> &[(String, String)] {
+        &self.headers
+    }
+
+    pub fn set_headers(&mut self, headers: Vec<(String, String)>) {
```
relrelb, line 129: nit: Maybe we should set a bit simpler interface? How about `add_header`?

desktop/src/navigator.rs
```diff
@@ -160,10 +161,20 @@ impl NavigatorBackend for ExternalNavigatorBackend {
...
 
-                let isahc_request = match request.method() {
+                let mut isahc_request = match request.method() {
                     NavigationMethod::Get => IsahcRequest::get(processed_url.to_string()),
                     NavigationMethod::Post => IsahcRequest::post(processed_url.to_string()),
                 };
+                if let Some(headers) = isahc_request.headers_mut() {
+                    for (name, val) in request.headers().iter() {
+                        headers.insert(
+                            HeaderName::from_bytes(name.as_bytes())
```
relrelb, line 171: I see `HeaderName` implements [`FromStr`](https://docs.rs/http/latest/http/header/struct.HeaderName.html#impl-FromStr-for-HeaderName). I think it can be used for a bit cleaner code.

desktop/src/navigator.rs
```diff
@@ -160,10 +161,20 @@ impl NavigatorBackend for ExternalNavigatorBackend {
...
                     client.ok_or_else(|| Error::FetchError("Network unavailable".to_string()))?;
 
-                let isahc_request = match request.method() {
+                let mut isahc_request = match request.method() {
                     NavigationMethod::Get => IsahcRequest::get(processed_url.to_string()),
                     NavigationMethod::Post => IsahcRequest::post(processed_url.to_string()),
                 };
+                if let Some(headers) = isahc_request.headers_mut() {
+                    for (name, val) in request.headers().iter() {
+                        headers.insert(
```
relrelb, line 170: Per the docs, it might be that [`append`](https://docs.rs/http/latest/http/header/struct.HeaderMap.html#method.append) is more suitable than [`insert`](https://docs.rs/http/latest/http/header/struct.HeaderMap.html#method.insert), as the former allows multiple headers of the same name.

Aaron1011: Testing shows that flash player only considers the last header in the array for a given name (later values overwrite earlier values). I'll add a comment explaining that.

core/src/backend/navigator.rs
Aaron1011: I've changed the code to perform header name deduplication in `loader.rs`. The simplest interface is now for `set_headers` to take in an `IndexMap`.

tests/tests/swfs/avm2/loader_load/actual.txt
```diff
@@ -0,0 +1,126 @@
+Test 0
```
moulins, line 1: Shouldn't this file not be committed?

Dinnerbone (approved):

PR #10887, opened 2023-04-30
avm2: Date related fixes

Lord-McSweeney:
1. Date should have a call handler.
2. Date should have a static length property.

There is one issue remaining with Date: `new Date(undefined)` should be `Invalid Date` instead of returning the same thing as `new Date()`.

### DIFF
--- a/core/src/avm2/globals/date.rs
+++ b/core/src/avm2/globals/date.rs
@@ -319,6 +319,19 @@ pub fn class_init<'gc>(
     Ok(Value::Undefined)
 }
 
+pub fn call_handler<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    _this: Option<Object<'gc>>,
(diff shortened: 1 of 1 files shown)


iwannabethedev:
I know too little of Rust and Ruffle to review this, but at the surface it looks good to me.

1. *Date should have a call handler.* I do not know what a call handler is (is it for conversion between types), so I cannot really give feedback on this. Have you tested it manually?
2. *Date should have a static length property.* I am convinced that this part is correct, since it is the same implementation as in #10914 and that PR has an automated test for it.

**Can someone else that has more insight into Ruffle and Rust review this?**

Lord-McSweeney:
The call handler is what gets called when code calls `Date(some parameters)` instead of using it as a constructor. From my tests, it always returns the current date, regardless of parameters.

Dinnerbone (approved):

PR #10900, opened 2023-05-01
web: Various improvements to options

n0samu:
Intended to supersede #10831, this PR does the following:
- Adds a collapsible "Advanced options" section to the extension settings page and moves less commonly used options there. (Thanks to @relrelb)
- Adds an "Automatic" option for preferred renderer in the extension settings page (Thanks to @iwannabethedev)
- Adds missing extension messages and improves wording for some existing ones. (Thanks to everyone who suggested these)
- ~~Adds an `enableFallbackRenderers` option that can be used to disable all fallback renderers. This is an attempt to address https://github.com/ruffle-rs/ruffle/pull/9904#discussion_r1174630515 in a way that I find more palatable than #10835. Even so, I would rather not add this option because I think its usefulness to developers is outweighed by the opportunities it creates for users and webmasters to cause problems for themselves. If you all agree, I would be very happy to remove this part of this PR!~~ Removed!

### DIFF
--- a/web/packages/extension/assets/_locales/en/messages.json
+++ b/web/packages/extension/assets/_locales/en/messages.json
@@ -21,7 +21,7 @@
         "message": "Log level"
     },
     "settings_max_execution_duration": {
-        "message": "How long can ActionScript execute for before being forcefully stopped (in seconds)"
+        "message": "Maximum allowed ActionScript execution duration (in seconds)"
     },
     "settings_player_version": {
         "message": "Flash player version number to emulate (range 1-32)"
--- a/web/packages/extension/assets/_locales/en/messages.json
+++ b/web/packages/extension/assets/_locales/en/messages.json
@@ -2,11 +2,8 @@
     "settings_ruffle_enable": {
         "message": "Play Flash content in Ruffle"
     },
-    "settings_page_ignore_optout": {
-        "message": "Play Flash content even on sites that disable Ruffle"
-    },
     "settings_ignore_optout": {
-        "message": "Ignore website compatibility warnings"
+        "message": "Play Flash content even on sites that disallow Ruffle"
     },
     "settings_show_swf_download": {
         "message": "Show SWF download in context menu"
@@ -26,6 +23,9 @@
     "settings_player_version": {
         "message": "Flash player version number to emulate (range 1-32)"
     },
+    "settings_preferred_renderer": {
+        "message": "Preferred renderer"
+    },
     "status_init": {
         "message": "Reading current tab…"
     },
@@ -39,13 +39,13 @@
         "message": "Checking Ruffle status on current tab…"
     },
     "status_result_running": {
-        "message": "Ruffle is loaded and running Flash content on the current tab."
+        "message": "Ruffle is loaded and ready to run Flash content on the current tab."
     },
     "status_result_optout": {
         "message": "Ruffle is not loaded because the current page has marked itself as incompatible."
     },
     "status_result_disabled": {
-        "message": "Ruffle is not loaded because it was disabled by the user."
+        "message": "Ruffle is not loaded because it is disabled."
     },
     "status_result_error": {
         "message": "An error occurred when querying the current tab's instance of Ruffle."
--- a/web/packages/extension/assets/_locales/de/messages.json
+++ b/web/packages/extension/assets/_locales/de/messages.json
@@ -2,11 +2,8 @@
     "settings_ruffle_enable": {
         "message": "Flash-Inhalte in Ruffle abspielen"
     },
-    "settings_page_ignore_optout": {
-        "message": "Flash-Inhalte abspielen, auch wenn die Seite Ruffle abgeschaltet hat"
-    },
     "settings_ignore_optout": {
-        "message": "Ignoriere Kompatibilitätswarnungen von Websites"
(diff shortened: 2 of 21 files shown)


iwannabethedev:
> Even so, I would rather not add this option because I think its usefulness to developers is outweighed by the opportunities it creates for users and webmasters to cause problems for themselves. If you all agree, I would be very happy to remove this part of this PR!

Is there an easy and relatively convenient way for Ruffle-developers (and maybe also players, game-developers and webmasters, though these groups of users seem less important to me in regards to this part of the topic) to view or determine which render-backend ended up getting chosen? Because if so, I imagine that Ruffle-developers could depend on that to see which render-backend is being used, and I believe that would lessen a need for supporting a forced option in any case.

n0samu:
Yes, you can right-click => Copy Debug Info and paste the text somewhere to see. There will be a line that says (for example):
`Renderer: wgpu`
Personally I find that convenient enough.

iwannabethedev:
I do not at all have a good feeling for whether that is convenient or not, I have never personally encountered it. I wonder if it might make sense to ask Dinnerbone if he thinks it would be convenient enough, since he described having multiple times encountered this issue: *"I know I've been caught a few times by using one backend instead of another when the preference wasn't doing what I expected during dev. "* .

For the log level method, I am not quite certain, since you would have to enable it before running or reload the page. Though, you could probably just set the log level to "Info", open a new tab with the application, see the first few lines which render-backend was chosen, and then revert.

For the context menu and copy-pasting debug information, is the right-click Flash context menu always available?

But, this is two different methods that one can use. And they seem decently straight-forward enough so that even users could also use them. They are not all that convenient, but not all that inconvenient either.

I am wondering if one could make the "forcedRenderer" more user-friendly by, when the selected render-backend is not available, to give a sort-of pop-up (

iwannabethedev:
But this might be extra work and effort that is not worth what advantages it might give.

iwannabethedev:
I personally think there are reasonable arguments for and against keeping something like "forceRenderer"/"enableFallbackRenderers".

iwannabethedev:
After some discussion and investigation elsewhere, one possible course of action might be to forgo the "forceRenderer" and "enableFallbackRenderers", and instead do something similar to Phaser 3, where which render-backend that has been chosen is always printed in the browser console:

Phaser 3 render-backend console printing

In the above image, you can see the "WebGL" string (as part of the nicely-colored and -formatted line), which indicates that "WebGL" (out of the two "WebGL" and "Canvas" render-backend options for Phaser 3) is used.

If that is always printed in the browser console, it would be even more convenient to figure out which render-backend was chosen instead of having to set the logging level to "Info".

I believe this could be a good substitute for "forceRenderer" and "enableFallbackRenderers", in particular since it would avoid any panic'ing as @n0samu describes, and it should make it relatively more convenient than setting the logging level to "info", since you would have access to this information no matter the logging level. And there is additionally also the possibility of getting debug info through the "Copy debug info" in the right-click context menu.

I will note that Phaser 3 additionally also has "forced" selection as well as "auto"/fallback selection, as the image in https://github.com/ruffle-rs/ruffle/pull/10835#issuecomment-1523170221 describes, but the situation for Phaser 3 is rather different from Ruffle in this regard, for in Ruffle, this forced selection setting is relevant only for game developers and while the game is being developed and maintained, which is very different from the situation and circumstances for Ruffle and applications run with Ruffle (Phaser 3 is also a game framework, while Ruffle is an emulator). Phaser 3 apparently also has no error handling if a forced selection is not supported, requiring the game developer to implement error handling. And Phaser 3 has WebGL-specific graphics features, while Flash and Ruffle's renderer-backends are meant to support all graphics features.

iwannabethedev:
I will open a PR that logs the currently used renderer-backend, since I think it would be a nice improvement in any case.

Dinnerbone:
@n0samu given the "enableFallbackRenderers" is debated topic right now, in idea and execution, can we remove it from this PR and pull it out into its own?

I'd like to get the rest of the fixes in, especially since you've touched translations.

n0samu:
Yep sounds good 👍

n0samu:
Alright I've rebased this and removed the enableFallbackRenderers change

Dinnerbone (approved):

web/packages/extension/assets/_locales/en/messages.json
```diff
@@ -21,11 +18,17 @@
...
-        "message": "How long can ActionScript execute for before being forcefully stopped (in seconds)"
+        "message": "Maximum allowed ActionScript execution duration (in seconds)"
     },
     "settings_player_version": {
         "message": "Flash player version number to emulate (range 1-32)"
     },
+    "settings_preferred_renderer": {
+        "message": "Preferred renderer"
+    },
+    "settings_enable_fallback_renderers": {
```
iwannabethedev, line 29: Should this be removed?

iwannabethedev:
I like the changes, I think it looks good, and I would believe that it is ready to be merged once the latest review comment has been looked at.

web/packages/extension/assets/_locales/en/messages.json
n0samu: Oh right I forgot about that

relrelb (approved):

PR #10903, opened 2023-05-01
extension: Log used renderer; Warn if not preferred.

iwannabethedev:
Log which renderer-backend was actually used in the browser console.

Add an option for toggling whether to warn or not
if the actually used renderer is not the same as
the preferred renderer, warning both in the console and also with a `window.alert()` popup. This option is meant first and foremost for Ruffle-developers.

@n0samu @relrelb

### DIFF
--- a/web/packages/core/src/config.ts
+++ b/web/packages/core/src/config.ts
@@ -36,4 +36,5 @@ export const DEFAULT_CONFIG: Required<BaseLoadOptions> = {
     polyfills: true,
     playerVersion: null,
     preferredRenderer: null,
+    warnIfPreferredRendererIsNotUsed: false,
 };
--- a/web/packages/core/src/load-options.ts
+++ b/web/packages/core/src/load-options.ts
@@ -393,6 +393,16 @@ export interface BaseLoadOptions {
     preferredRenderer?: RenderBackend | null;
 
     /**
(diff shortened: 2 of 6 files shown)


iwannabethedev:
warning dialog

iwannabethedev:
The popup dialog is rather ugly, but it is not really meant to be used by non-Ruffle-developers.

As has been discussed elsewhere, one advantage of this approach is that there is no panic if the preferred renderer is not actually used, while at the same time making it very easy for Ruffle-developers to determine if the preferred renderer was not used.

relrelb:
Honestly I'm quite disgusted by using `window.alert()`. Since it's meant to be used by Ruffle developers only, I don't see it much better than panicking. My preference here is simplicity over user experience, since there aren't really any "users".

iwannabethedev:
> Honestly I'm quite disgusted by using `window.alert()`. Since it's meant to be used by Ruffle developers only, I don't see it much better than panicking. My preference here is simplicity over user experience, since there aren't really any "users".

Good point, though if a user accidentally or foolishly select that option, is it not better to show `window.alert` than panicking? Like, if a user opens a pull request with a bot commit ... :smiley: .

n0samu:
I think we arrived at an optimal solution with #10909, so let's forget about the other stuff (panicking, window.alert, etc)

Issue #10699, opened 2023-04-15
Error on https://f.silvergames.com/ruffle/player.php


nguie2:
# Error Info
Error name: Error
Error message: panicked at 'called `Option::unwrap()` on a `None` value', core/src/bitmap/bitmap_data.rs:595:51
Error stack:
```
Error: panicked at 'called `Option::unwrap()` on a `None` value', core/src/bitmap/bitmap_data.rs:595:51
    at O.n.wbg.__wbg_new_15d3966e9981a196 (https://f.silvergames.com/ruffle/23042023/core.ruffle.83032e5cb08379aef3d6.js:1:47895)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::h927501cce3fc7417 (wasm://wasm/02acc5e6:wasm-function[8165]:0x8010b2)
    at std::panicking::rust_panic_with_hook::hc53aea0352e77326 (wasm://wasm/02acc5e6:wasm-function[5438]:0x735478)
    at core::panicking::panic_fmt::hf4a9df75710ece83 (wasm://wasm/02acc5e6:wasm-function[8047]:0x7fbd21)
    at core::panicking::panic::h54beb3f515dd3c6d (wasm://wasm/02acc5e6:wasm-function[7453]:0x7c8944)
    at ruffle_core::bitmap::bitmap_data::BitmapData::update_dirty_texture::hee3ddbdfbdc50e9d (wasm://wasm/02acc5e6:wasm-function[1254]:0x3bec50)
    at <ruffle_core::library::MovieLibrarySource as ruffle_render::bitmap::BitmapSource>::bitmap_handle::h0909caf8b3654249 (wasm://wasm/02acc5e6:wasm-function[3183]:0x6168dc)
    at <ruffle_render_wgpu::backend::WgpuRenderBackend<T> as ruffle_render::backend::RenderBackend>::register_shape::hd515b52f2b24d068 (wasm://wasm/02acc5e6:wasm-function[1040]:0x3550d5)
    at ruffle_core::display_object::movie_clip::MovieClipData::define_shape::h2a01979e41768b81 (wasm://wasm/02acc5e6:wasm-function[2992]:0x5f64b2)
    at ruffle_core::display_object::movie_clip::MovieClip::preload::h8b435e95ba744ef1 (wasm://wasm/02acc5e6:wasm-function[617]:0x1d75a3)
```

# Player Info
Allows script access: false

# Page Info
[...]

Issue #3299, opened 2021-02-17
Image stretched incorrectly on z0r.de/630 (texture larger than vulkan/webgl limit)


torokati44:
On Linux desktop (Vulkan backend) Ruffle shows this really stretched image for [z0r.de/630](https://z0r.de/L/z0r-de_630.swf):

![image](https://user-images.githubusercontent.com/288816/108217327-c0644680-7133-11eb-9a14-ed72ed48c5b9.png)

In the web version (like on the demo page), it's just solid black, with the following message in the console:
`WebGL warning: texImage: Requested size at this level is unsupported.`
This might be just about the size of the texture (image) itself, in pixels, and not the transformation (pan/stretch?) applied to it.

I haven't tested the canvas backend.

**Expected behavior**

Flash Player shows this in the very beginning, then scrolls it up for a long time:

![image](https://user-images.githubusercontent.com/288816/108216674-05d44400-7133-11eb-93b8-c1aaece70b20.png)

torokati44:
A completely wild guess, but this might also be related: https://z0r.de/6812
The video is in only one quadrant, as opposed to filling the entire player area.

torokati44:
New development:
In `WgpuRenderBackend::register_bitmap`, the RGBA data in the :cow: is correct before it is uploaded into the texture.
However, RenderDoc shows that at runtime, the captured texture itself is stretched.
It is the correct size (300x17700), and it is drawn with all the correct transforms. Only the image data itself is stretched. Only vertically, by a factor of around 13.5.

So this:

![image](https://user-images.githubusercontent.com/288816/108282895-ae5dc480-7182-11eb-9ccf-fee065a04baf.png)

Becomes (approximately) this:

![image](https://user-images.githubusercontent.com/288816/108282850-95551380-7182-11eb-963e-bd924cd8f923.png)

As a result, the texture only contains the top ~7.5% of the image content.

All of this is with **RADV**.
Whan switching to **Lavapipe**, the loop is displayed **correctly**.

This looks like a `wgpu-rs` bug?

torokati44:
Or worse, a RADV bug... :/
I had hoped that https://github.com/ruffle-rs/ruffle/pull/3066 would fix this, but just checked, and no.

torokati44:
Oh, I think I got it...

According to `vulkaninfo`, both RADV and Lavapipe report the `maxImageDimension2D` device limit to be `16384`.

[...]

Issue #10904, opened 2023-05-01
Error on https://rawcdn.githack.com/lioxryt/lioxryt-assets/a4041fc9b45936f148d3d7134977d5df4329e237/flash/index.html


UMADBRO233:
# Error Info
Error name: Error
Error message: panicked at 'called `Option::unwrap()` on a `None` value', core/src/bitmap/bitmap_data.rs:602:51
Error stack:
```
Error: panicked at 'called `Option::unwrap()` on a `None` value', core/src/bitmap/bitmap_data.rs:602:51
    at O.n.wbg.__wbg_new_15d3966e9981a196 (https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.312167b7d45670ed237b.js:1:48745)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::hd28f65a61f685e47 (wasm://wasm/02b74ace:wasm-function[8390]:0x81042c)
    at std::panicking::rust_panic_with_hook::h7f7102b82d51338f (wasm://wasm/02b74ace:wasm-function[5498]:0x732337)
    at core::panicking::panic_fmt::he4489d678d6570d5 (wasm://wasm/02b74ace:wasm-function[8245]:0x80acaa)
    at core::panicking::panic::hea78c4e8b318407d (wasm://wasm/02b74ace:wasm-function[7551]:0x7d095e)
    at ruffle_core::bitmap::bitmap_data::BitmapData::update_dirty_texture::h69e242bde85db84e (wasm://wasm/02b74ace:wasm-function[1279]:0x3d523d)
    at <ruffle_core::library::MovieLibrarySource as ruffle_render::bitmap::BitmapSource>::bitmap_handle::h218cbf3363c21bf3 (wasm://wasm/02b74ace:wasm-function[3247]:0x61c312)
    at <ruffle_render_wgpu::backend::WgpuRenderBackend<T> as ruffle_render::backend::RenderBackend>::register_shape::haa5f47cd0e691a1c (wasm://wasm/02b74ace:wasm-function[1120]:0x3868d1)
    at ruffle_core::display_object::movie_clip::MovieClipData::define_shape::h127073ca34759630 (wasm://wasm/02b74ace:wasm-function[3609]:0x66ba07)
    at ruffle_core::display_object::movie_clip::MovieClip::preload::h30e5eba151b7c315 (wasm://wasm/02b74ace:wasm-function[621]:0x1dece0)
```

# Player Info
Allows script access: false

[...]

Issue #10843, opened 2023-04-27
Error on https://www.minijuegos.com/helpers/game/sandbox.php


LuisUrbe:
# Error Info
Error name: Error
Error message: panicked at 'called `Option::unwrap()` on a `None` value', core/src/bitmap/bitmap_data.rs:602:51
Error stack:
```
Error: panicked at 'called `Option::unwrap()` on a `None` value', core/src/bitmap/bitmap_data.rs:602:51
    at O.n.wbg.__wbg_new_15d3966e9981a196 (https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.b23fa269e4f1e918d88d.js:1:48745)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::hd28f65a61f685e47 (wasm://wasm/02b61d02:wasm-function[8383]:0x80c16a)
    at std::panicking::rust_panic_with_hook::h7f7102b82d51338f (wasm://wasm/02b61d02:wasm-function[5498]:0x72f1b8)
    at core::panicking::panic_fmt::he4489d678d6570d5 (wasm://wasm/02b61d02:wasm-function[8238]:0x8069e8)
    at core::panicking::panic::hea78c4e8b318407d (wasm://wasm/02b61d02:wasm-function[7542]:0x7cc5f7)
    at ruffle_core::bitmap::bitmap_data::BitmapData::update_dirty_texture::h69e242bde85db84e (wasm://wasm/02b61d02:wasm-function[1281]:0x3d5f88)
    at <ruffle_core::library::MovieLibrarySource as ruffle_render::bitmap::BitmapSource>::bitmap_handle::h218cbf3363c21bf3 (wasm://wasm/02b61d02:wasm-function[3241]:0x61df78)
    at <ruffle_render_wgpu::backend::WgpuRenderBackend<T> as ruffle_render::backend::RenderBackend>::register_shape::haa5f47cd0e691a1c (wasm://wasm/02b61d02:wasm-function[1119]:0x38389f)
    at ruffle_core::display_object::movie_clip::MovieClipData::define_shape::h3ce1a0757a2ce061 (wasm://wasm/02b61d02:wasm-function[3597]:0x66c051)
    at ruffle_core::display_object::movie_clip::MovieClip::preload::h30e5eba151b7c315 (wasm://wasm/02b61d02:wasm-function[620]:0x1e0cd4)
```

# Player Info
Allows script access: false

# Page Info
[...]

PR #10908, opened 2023-05-01
wgpu: Resize textures if they're bigger than the system supports

Dinnerbone:
Tested by forcing the max size to 1024 and things work great, just not as pretty as they otherwise would be.

There's other approaches to solving this and they all have their pros and cons:
- This way (resize with cubic filter): not the prettiest for pixel art, mostly works as expected though
- Using a different filter: might make some art look better than others
- Using & stitching multiple textures: would make all art look better but significantly complicate rendering
- Trying to identify which region of the texture is used (ie individual draw calls in a mesh), and only using those: complicates things but may look better in some cases

Ultimately, this should be better than nothing for minimum fuss.

Note that webgl backend currently ignores the system limits and for some reason they don't complain, but that's not something I want to rely on.

### DIFF
--- a/render/wgpu/src/backend.rs
+++ b/render/wgpu/src/backend.rs
@@ -11,9 +11,12 @@ use crate::{
     as_texture, format_list, get_backend_names, ColorAdjustments, Descriptors, Error,
     QueueSyncHandle, RenderTarget, SwapChainTarget, Texture, Transforms,
 };
+use image::imageops::FilterType;
 use ruffle_render::backend::Context3D;
 use ruffle_render::backend::{RenderBackend, ShapeHandle, ViewportDimensions};
-use ruffle_render::bitmap::{Bitmap, BitmapHandle, BitmapSource, PixelRegion, SyncHandle};
+use ruffle_render::bitmap::{
+    Bitmap, BitmapFormat, BitmapHandle, BitmapSource, PixelRegion, SyncHandle,
+};
 use ruffle_render::commands::CommandList;
 use ruffle_render::error::Error as BitmapError;
 use ruffle_render::filters::Filter;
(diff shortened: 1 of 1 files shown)


firoball:
Supposedly progresses https://z0r.de/630 (Looooong cat).

torokati44:
Ref: https://github.com/ruffle-rs/ruffle/issues/3299

Lord-McSweeney:
Fixes #3299, #10843
Should fix (but not tested) #10666

#10904 no longer panics but instead hangs (at least for me on Linux, using desktop Ruffle)

n0samu:
Can confirm that #3299 works! And this fixes #10699 too!

render/wgpu/src/backend.rs
```diff
@@ -515,16 +541,22 @@ impl<T: RenderTarget + 'static> RenderBackend for WgpuRenderBackend<T> {
         &mut self,
         handle: &BitmapHandle,
         bitmap: Bitmap,
-        region: PixelRegion,
+        mut region: PixelRegion,
     ) -> Result<(), BitmapError> {
         let texture = as_texture(handle);
 
+        let mut bitmap = bitmap.to_rgba();
+        if self.clamp_bitmap(&mut bitmap) {
```
Aaron1011, line 549: Since we perform clamping when the bitmap is initially registered, will this ever do anything? The incoming `bitmap` can't be larger than the existing `BitmapHandle`, or the `write_texture` will fail.

Dinnerbone: Yes, things panic without this.
If a game registers a bitmap of 1000 and we internally resize it to 500, the game still thinks it's 1000 and will update it with a 1000 bitmap, which is wrong for the actual texture we're storing.

Aaron1011 (approved):

PR #10909, opened 2023-05-01
extension: Log used renderer.

iwannabethedev:
Log which renderer-backend was actually used in the browser console.

This is basically the same as https://github.com/ruffle-rs/ruffle/pull/10903 , but without the toggle option for whether to warn or not if the actually used renderer is not the same as the preferred renderer.

Related to: https://github.com/ruffle-rs/ruffle/pull/10835 , https://github.com/ruffle-rs/ruffle/pull/10831 , https://github.com/ruffle-rs/ruffle/pull/10900 .

CC: @n0samu @relrelb .

### DIFF
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -511,10 +511,17 @@ export class RufflePlayer extends HTMLElement {
             this.loadedConfig
         );
         this._cachedDebugInfo = this.instance!.renderer_debug_info();
+
+        const actuallyUsedRendererName = this.instance!.renderer_name();
+
         console.log(
-            "New Ruffle instance created (WebAssembly extensions: " +
+            "%c" +
+                "New Ruffle instance created (WebAssembly extensions: " +
                 (ruffleConstructor.is_wasm_simd_used() ? "ON" : "OFF") +
-                ")"
+                " | Used renderer: " +
+                (actuallyUsedRendererName ?? "") +
+                ")",
+            "background: #37528C; color: #FFAD33"
         );
 
         // In Firefox, AudioContext.state is always "suspended" when the object has just been created.
--- a/web/src/lib.rs
+++ b/web/src/lib.rs
@@ -378,6 +378,11 @@ impl Ruffle {
             .unwrap_or(JsValue::NULL)
     }
 
+    pub fn renderer_name(&self) -> JsValue {
+        self.with_core(|core| JsValue::from_str(core.renderer().name()))
+            .unwrap_or(JsValue::NULL)
+    }
+
     // after the context menu is closed, remember to call `clear_custom_menu_items`!
     pub fn prepare_context_menu(&mut self) -> JsValue {
         self.with_core_mut(|core| {
--- a/render/canvas/src/lib.rs
+++ b/render/canvas/src/lib.rs
@@ -486,6 +486,10 @@ impl RenderBackend for WebCanvasRenderBackend {
         Cow::Borrowed("Renderer: Canvas")
     }
 
+    fn name(&self) -> &'static str {
+        "canvas"
+    }
+
     fn set_quality(&mut self, _quality: StageQuality) {}
 }
 
--- a/render/src/backend.rs
+++ b/render/src/backend.rs
@@ -67,6 +67,10 @@ pub trait RenderBackend: Downcast {
     fn context3d_present(&mut self, context: &mut dyn Context3D) -> Result<(), Error>;
 
     fn debug_info(&self) -> Cow<'static, str>;
+    /// An internal name that is used to identify the render-backend.
+    /// For valid values, look at:
+    /// web/packages/core/src/load-options.ts:RenderBackend
+    fn name(&self) -> &'static str;
 
     fn set_quality(&mut self, quality: StageQuality);
 }
--- a/render/src/backend/null.rs
+++ b/render/src/backend/null.rs
@@ -93,5 +93,9 @@ impl RenderBackend for NullRenderer {
         Cow::Borrowed("Renderer: Null")
     }
 
+    fn name(&self) -> &'static str {
+        ""
+    }
+
     fn set_quality(&mut self, _quality: StageQuality) {}
 }
--- a/render/webgl/src/lib.rs
+++ b/render/webgl/src/lib.rs
@@ -1101,6 +1101,10 @@ impl RenderBackend for WebGlRenderBackend {
         Cow::Owned(result.join("\n"))
     }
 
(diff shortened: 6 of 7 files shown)


relrelb:
In fact this can be implemented completely on Rust side, and with greater simplicity, using just `tracing::info!()` to log the chosen backend (and I think this already the case for some backends). Am I missing a reason for it to pass through TypeScript?

n0samu:
I think we wanted to make the message appear regardless of the logLevel, like the "new Ruffle instance created" message. (My idea was to add the renderer name to that message specifically.)

iwannabethedev:
Yes, the concept was to make it a bit more convenient to view which renderer-backend was used, such that you would not have to switch the logging level to "Info" first. I spent some time seeing if I could do it just from Rust, but I did not figure out how to do printing to the browser console apart from using `tracing::info` or `tracing::error`, and using `tracing::error` would probably be confusing since the printing is not an error.

The printing is in many ways inspired by the image in https://github.com/ruffle-rs/ruffle/pull/10900#issuecomment-1529528513 , which is taken from a Phaser 3 application.

Additionally, another reason is that it would allow using an HTML-popup like with `window.alert` though that is only relevant if the changes in #10903 are considered.

web/src/lib.rs
```diff
@@ -81,6 +81,7 @@ struct RuffleInstance {
     has_focus: bool,
     trace_observer: Arc<RefCell<JsValue>>,
     log_subscriber: Arc<Layered<WASMLayer, Registry>>,
+    actually_used_renderer: String,
```
relrelb, line 84: Instead of storing `actually_used_renderer` in a field, I suggest adding a dedicated method for this to the `RenderBackend` trait, similar to `debug_info()`:
https://github.com/ruffle-rs/ruffle/blob/fe56f8d2120eb9d10aa72d9ff2898c46afe96b0f/render/src/backend.rs#L69

It could be something like:
```rs
fn name(&self): &'static str;
````

File: web/packages/core/src/ruffle-player.ts [hunk 2]
```diff
@@ -516,6 +520,20 @@ export class RufflePlayer extends HTMLElement {
                 (ruffleConstructor.is_wasm_simd_used() ? "ON" : "OFF") +
                 ")"
         );
+        {
+            const actuallyUsedRenderer =
+                this.instance!.get_actually_used_renderer() as
+                    | RenderBackend
```
relrelb, line 526: This type-cast isn't really necessary. Staying with `string` is good enough.

web/packages/core/src/ruffle-player.ts
```diff
@@ -516,6 +520,20 @@ export class RufflePlayer extends HTMLElement {
...
+                    | null
+                    | "";
+
+            if (actuallyUsedRenderer) {
+                console.log(
```
relrelb, line 531: I support n0samu's idea to display the used renderer in the `New Ruffle instance created` log above. I think we'd want to reduce the amount of explicit `console.log()`s to the minimum.

web/src/lib.rs
iwannabethedev: I like your suggestion, and I considered that multiple times, but I chickened out because I did not want to mess too much with the other crates of the codebase and I was in doubt about whether the name values for renderer-backends used in the `web` crate https://github.com/ruffle-rs/ruffle/blob/fe56f8d2120eb9d10aa72d9ff2898c46afe96b0f/web/src/lib.rs#L1437 would also be good to use in the rest of the codebase (like, should they be kept isolated or not). But, given that you suggest it, I would like to implement it, especially if @n0samu also likes the idea.

I also believe it would also simplify the code and make it cleaner.

Should I give it a go?

web/packages/core/src/ruffle-player.ts
iwannabethedev: It is good that you mention it again, because I somehow completely missed that it would fit well there :sweat_smile: (I think I was distracted or focused on other aspects). I will change it to be part of that.

web/src/lib.rs
relrelb: I wouldn't suggest it if I don't mean it :)
If you ask me, go for it!

web/packages/core/src/ruffle-player.ts
iwannabethedev: new browser console logging

web/src/lib.rs
```diff
@@ -373,6 +374,11 @@ impl Ruffle {
             .unwrap_or(JsValue::NULL)
     }
 
+    pub fn get_actually_used_renderer(&self) -> JsValue {
```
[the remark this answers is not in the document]
iwannabethedev: I have changed its name to `renderer_name` and will push once I have tested it.

In regards to its return type, I am calling that function from Typescript, so I do not know whether it can have a different type than `JsValue` .

web/packages/core/src/ruffle-player.ts
```diff
@@ -511,11 +511,20 @@ export class RufflePlayer extends HTMLElement {
             this.loadedConfig
         );
         this._cachedDebugInfo = this.instance!.renderer_debug_info();
-        console.log(
-            "New Ruffle instance created (WebAssembly extensions: " +
-                (ruffleConstructor.is_wasm_simd_used() ? "ON" : "OFF") +
-                ")"
-        );
+        {
```
relrelb, line 514: Is there a reason for opening an inner scope?

iwannabethedev: None really, apart from ensuring that the variable name `actuallyUsedRendererName` is not accessible later in the outer scope - I can remove it if you want.

web/src/lib.rs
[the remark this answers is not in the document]
relrelb: Can you give it a try at least? The `wasm-bindgen` docs doesn't mention whether returning `&str` is supported or not, but I do see `String` is documented: https://rustwasm.github.io/docs/wasm-bindgen/reference/types/string.html

web/packages/core/src/ruffle-player.ts
relrelb: Yeah, removing it would be nicer in my opinion.

web/src/lib.rs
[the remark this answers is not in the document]
iwannabethedev: Sorry, I tried, but I know too little of Rust to figure out the types.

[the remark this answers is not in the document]
relrelb: This seems to be https://github.com/rustwasm/wasm-bindgen/issues/1187. A workaround would probably be returning a `String`, but I think this can be skipped for now.

PR #10911, opened 2023-05-02
Localise core and web/core

Dinnerbone:
Currently only localises context menus in core, and all user facing texts in web.

I chose to use [Fluent](https://projectfluent.org/) as the localisation system, as it has nice support for both rust and typescript.

This PR doesn't include any actual translations, they need to be created after this is accepted and the new strings are uploaded to crowdin for people to work with.


Toad06:
Thanks for your work on this, this is a very nice improvement and hopefully this will make more people use Ruffle.

iwannabethedev:
In regards to selecting Fluent: The apparent lack of activity at

- https://github.com/projectfluent/fluent.js/graphs/contributors
- https://github.com/projectfluent/fluent-rs/graphs/contributors

makes me a little bit worried, though I might be looking at the wrong repositories, and a lack of activity is not necessarily bad, since it might mean that the code works. And there might be more activity in the future.

On the other hand, it appears that Mozilla/Firefox is committed to using Fluent internally:
- https://www.arewefluentyet.com/
- https://firefox-source-docs.mozilla.org/l10n/overview.html

Fluent appears a bit focused on Mozilla/Firefox, though Fluent also supports both Rust and Javascript/web, which fits very well with the main technologies used in Ruffle. And I imagine that Ruffle has the resources to change to a different translation technology in the worst case - I imagine that there are not a huge number of pieces of text to translate or places in the code to change translation handling of, and Ruffle is fairly active itself and have vested and varied interests in its development and success.

So, all in all I believe that Fluent is a good choice and that the worst-case (Fluent ending up dropped and not being maintained or developed) is not that bad either way.

iwannabethedev:
(On licensing:

No issues there, Firefox uses it, and Fluent appears to use Apache 2.0 in https://github.com/projectfluent/fluent.js and https://github.com/projectfluent/fluent-rs , which is one of the licenses that Ruffle uses as well).

web/packages/core/src/i18n.ts
```diff
@@ -0,0 +1,123 @@
...
+    const files = BUNDLED_TEXTS[locale];
+    if (files) {
+        for (const filename in files) {
+            const text = files[filename];
+            if (text) {
+                for (const error in bundle.addResource(
+                    new FluentResource(text)
+                )) {
+                    console.error(
+                        `Error in text for ${locale} ${filename}: ${error}`
```
iwannabethedev, line 31: Nit: I do not understand the code and I do not know Fluent, but might it make sense to change it to this (adding `text` to the printed string):

`Error in text for ${locale} ${filename} ${text}: ${error}`

?

Dinnerbone: It'll be a huge text file, that's a lot of spam. We can look up the text file ourselves.

web/packages/core/src/i18n.ts
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * @param id ID of the text to retrieve
+ * @param args Any arguments to use when creating the localised text
+ * @returns Localised text
+ */
+export function text(
+    id: string,
+    args?: Record<string, FluentVariable> | null
+): string {
+    const locales = negotiateLanguages(
```
iwannabethedev, line 81: Nit: Could it make sense to place the call to and result of `negotiateLanguages()` elsewhere? Like, instead of calling `negotiateLanguages()` on every call to `text()`, just save the result once and use it there? Though, that depends on whether the arguments to `negotiateLanguages()`, or relevant state of `Fluent`, can change after initialization.

Looking at for instance https://github.com/mozilla/fx-private-relay/blob/eb21d4b31f9f220e8d5d32d233ba4a6b6e728a30/frontend/src/functions/getL10n.ts#L38 (though this is a random file I found through searching on Github, `negotiateLanguages()` feels a bit to me like an initialization call of sorts, though that might not be appropriate in all cases.

On the other hand, while this https://github.com/common-voice/sentence-collector/blob/4b2b77299850b5570617cc8e64c0d11e34ee581d/web/src/l10n.tsx#L58 indicates that the function is rarely called, the localization appears to be able to change during runtime, like if the user changes locale, so as a first approximation, keeping it as it is is presumably fine and more correct than only calculating it once. A more granular method might be to do something similar to that line, where `negotiateLanguages()` is called whenever the locale (or other state?) changes.

Maybe add a TODO comment that this could be made more granular, such as calculating it whenever the locale changes?

But again, this comment is a nitpick, I will leave it up to you.

web/packages/core/src/i18n.ts
```diff
@@ -0,0 +1,123 @@
...
+ * The users preferred locales are used, in priority order, to find the given text.
+ *
+ * If no text is found for any preferred locale, en-US will be used.
+ * If en-US does not contain a text for this ID, an error will be logged and the ID itself will be returned.
+ *
+ * @param id ID of the text to retrieve
+ * @param args Any arguments to use when creating the localised text
+ * @returns Localised text with each line in a Paragraph element
+ */
+export function textAsParagraphs(
```
iwannabethedev, line 110: I have not looked at the code closely, and this seems like an improvement on the existing code, but I do not really like the general usage of assigning to `innerHTML`, like here: https://github.com/ruffle-rs/ruffle/blob/master/web/packages/core/src/ruffle-player.ts#L1890 . Important to note is that it is the existing code, not your PR, that uses this `innerHTML` . So your PR is not really introducing anything. But I still do not like it. I do not know how translations are gotten into the project, if they are committed to the repository or how - if translations are committed to the repository, they will have gone through review either way. Your way of doing it does seem fine, since it is not introducing anything that is not already being done, and your code does seem like an improvement. But I would not be sad if people make an issue or other PRs to change away from using `innerHTML` in 'ruffle-player.ts' and elsewhere.

Would it make sense for me to make an issue that describes that refactoring away from using `innerHTML` in 'web/packages/core/src/ruffle-player.ts' for translations (and other usage) would be nice (not required)? Or is there already a TODO for it in the code?

I am tempted to propose that you add a new function, `textAsParagraphs`, with return type `HTMLElement`, and rename this function `textAsParagraphsInnerHTML`, and then call `textAsParagraphsInnerHTML` from `textAsParagraphs`. Simply such that other users have the choice of using the `HTMLElement`-returning or using the `string`-returning variant.

I do not consider this a blocker, and I will let it be up to you what to do or not change anything in the PR, this is mostly me feeling that `innerHTML` is somewhat of a code smell.

web/packages/core/src/i18n.ts
Dinnerbone: It kinda depends. Do we want to support languages being changed on the fly? Browsers allow that, and `navigator.language`/`navigator.languages` may change at any time.

On the other hand, they *do* fire an event for when it changes. I just didn't think it needed the investment of time to figure that out for the first PR :D

web/packages/core/src/i18n.ts
iwannabethedev: Would it make sense for me to make an issue on Github for cleaning up the `innerHTML`-usage in 'web/packages/core/src/ruffle-player.ts' and elsewhere?

iwannabethedev:
I have lack of knowledge of both Rust and Ruffle, but I have reviewed the changes lightly, and the changes overall look good to me. Should there be more reviews by others? (did @Toad06 review it as well?). I would guess that the PR can be merged as is. I have some minor nits/suggestions here and there, but they are not important.

web/packages/core/src/i18n.ts
Dinnerbone: It doesn't make sense to me to have a method that returns a div when everything currently wants a html string of <p>s

iwannabethedev: That makes sense, it was more for future usage, in case it is refactored later to not use `innerHTML` - but it is easy to add such a function later. I saw it more as changing the internal "API" a little bit.

iwannabethedev: I have created https://github.com/ruffle-rs/ruffle/issues/10976 .

(I think all of my questions have been answered, please do mark the review lines here resolved as you see fit).

Toad06 (approved):
I'm not the most qualified to review code, but this looks good to me.

I just want to mention that [shadow-template.ts](https://github.com/ruffle-rs/ruffle/blob/master/web/packages/core/src/shadow-template.ts) has two strings which are not translated currently (not sure if that's intentional because these strings are easy to miss):

> Click to unmute

> Backup all saves (download all sols)

Dinnerbone:
@Toad06 thanks for catching those, I've made a commit to address them

PR #10912, opened 2023-05-02
make panicText readonly

divinity76:
it's not supposed to be editable. ref https://github.com/ruffle-rs/ruffle/pull/10819#issuecomment-1529977894

### DIFF
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -1900,7 +1900,7 @@ export class RufflePlayer extends HTMLElement {
                 panicBody.classList.add("details");
 
                 const panicText = document.createElement("textarea");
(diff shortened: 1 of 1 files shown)


n0samu (approved):

web/packages/core/src/ruffle-player.ts
```diff
@@ -1900,6 +1900,7 @@ export class RufflePlayer extends HTMLElement {
                 panicBody.classList.add("details");
 
                 const panicText = document.createElement("textarea");
+                panicText.setAttribute("readonly", "1");
```
relrelb, line 1903: nit: I think the `.readOnly` property can be set directly, without `.setAttribute()`.

relrelb (approved):
Thank you!

PR #10914, opened 2023-05-02
avm2: Add 'static length' for most types.

iwannabethedev:
The implementation is fairly straight-forward, just adding a public static constant member named `length` with the correct integer value to the different types. This should implement most of the missing `static length` members in https://ruffle.rs/avm2.html . The test results were gotten by running against a stand-alone Flash player. It was tested that all of them were constant values.

For one of the types, the member was not implemented, because the type itself has not yet been implemented, namely `UninitializedError`.


core/src/avm2/globals/ArgumentError.as
```diff
@@ -6,5 +6,7 @@ package {
             super(message, code)
             this.name = prototype.name
         }
+
+        public static const length:Number = 1;
```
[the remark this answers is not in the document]
iwannabethedev: Very good catch, when I decompile `playerglobal32_0.swc` I can see that the type of the field in at least some of the types is `int`, so I will change it to that. Thank you!

tests/tests/swfs/avm2/static_length/test.as
```diff
@@ -0,0 +1,34 @@
...
+trace("Boolean.length is: " + Boolean.length);
+trace("XML.length is: " + XML.length);
+trace("Function.length is: " + Function.length);
+trace("TypeError.length is: " + TypeError.length);
+trace("URIError.length is: " + URIError.length);
+trace("Array.length is: " + Array.length);
+trace("uint.length is: " + uint.length);
+trace("Date.length is: " + Date.length);
+trace("Error.length is: " + Error.length);
+//trace("UninitializedError.length is: " + UninitializedError.length);
```
adrian17, line 33: why is this commented out?

adrian17 (approved):

tests/tests/swfs/avm2/static_length/test.as
iwannabethedev: Because `UninitializedError` has not yet been implemented - should I add a comment that explains it to the file? Like:

```
// UninitializedError has not yet been implemented at the time of writing this test.
//trace("UninitializedError.length is: " + UninitializedError.length);
```

?

iwannabethedev: Maybe:

```
// Once the class UninitializedError has been implemented, please:
// - Implement UninitializedError.length .
// - Replace the file "output.txt" with "output_original.txt".
// - Comment in the line below.
//trace("UninitializedError.length is: " + UninitializedError.length);
```

?

iwannabethedev: Alternatively, I could look at implementing the class `UninitializedError` first (and also add `static length` for it and comment-in this line), but I do not know how involved and complex implementing that class would be, nor how to test it.

iwannabethedev: I added some comments to the ActionScript test file.

core/src/avm2/globals/boolean.rs
```diff
@@ -150,5 +150,12 @@ pub fn create_class<'gc>(activation: &mut Activation<'_, 'gc>) -> GcCell<'gc, Cl
         AS3_INSTANCE_METHODS,
     );
 
+    const CONSTANTS: &[(&str, u32)] = &[("length", 1)];
+    write.define_constant_uint_class_traits(
```
Lord-McSweeney, line 154: This should be `define_constant_int_class_traits` (and the above line should use `&[(&str, i32)]`).

iwannabethedev: Very good catch and thank you very much, I should have been more careful with the types, my knowledge of types in ActionScript is very limited. (I also got a bit confused by https://github.com/ruffle-rs/ruffle/pull/10914#discussion_r1182597426  with `uint`).

I have opened all the types using FFDEC in "playerglobal32_0.swc", and I can confirm that they are all `int` (apart from `Namespace`, where the type has not been set (FFDEC says "null" for the type in P-code source), but I have changed it to `int` there as well).

PR #10916, opened 2023-05-02
avm2: Add AS 'length()' to XML.

iwannabethedev:


core/src/avm2/globals/XML.as
```diff
@@ -49,6 +49,10 @@ package {
         AS3 native function text():XMLList;
         AS3 native function toString():String;
 
+        public function length(): int {
```
Aaron1011, line 52: This should be an `AS3` function instead of a `public` function, and should be declared on the prototype in the same way that `toXMLString` is.

iwannabethedev: Thank you, and sorry about the wrong implementation before, I know very little about ActionScript. Does the new implementation look correct?

Lord-McSweeney (approved):

Aaron1011 (approved):

Issue #2176, opened 2020-12-31
An issue with some prodigy site archives


sks2002Official:
I’ve found some problems with the old archived prodigy sites. Most of them the audio overlaps each over Which makes it turn to pure noise. The Their Law site has a problem where the clock doesn’t work when you move the hands. The archive for alwaysoutnumbered.com isn’t working probably. On the original one it would loop all of the samples plus it would play all 3 of the tracks making it play a variant of Memphis bells. Here’s the site for some help http://theprodigy.dk/archive/

n0samu:
The clock issue seems to be fixed but the audio layering issue is still occurring.

PR #10920, opened 2023-05-02
web: Support copying to clipboard

### LINKED ISSUE #6476: Super Mario Flash
### Describe the bug

Steps to reproduce
1. Go to https://ia601805.us.archive.org/18/items/super-mario-flash/[Mario_flash.swf](https://ia601805.us.archive.org/18/items/super-mario-flash/Mario_flash.swf
2. Press 'Level Editor'
3. Select a character
4. Press 'Create'
5. Press 'Select' 2x
6. Press 'Build'
7. Press 'Close'
8. Press the up arrow key on your keyboard
9. Build a level (it can even just be 1 piece)
10. Press the up arrow key on your keyboard
11. press 'save' 2x
12. press 'code'
13. press 'copy to clipboard'
14. It doesn't work!

### Expected behavior

I expected it to copy the code.

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Mozilla Firefox version 98.0.1

### Additional information

_No response_

Toad06:
Closes #6476.

I also tried to address the TODO for the AVM2 `set_clipboard` but I wasn't able to find a way to retrieve the type of event that triggered the function call.

### DIFF
--- a/core/src/avm2/globals/flash/system/system.rs
+++ b/core/src/avm2/globals/flash/system/system.rs
@@ -11,8 +11,17 @@ pub fn set_clipboard<'gc>(
     _this: Option<Object<'gc>>,
     args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
-    //TODO: in FP9+ this function only works when called from a button handler in the Plugin due to
-    // sandbox restrictions
+    // The following restrictions only apply to the plugin.
+    // TODO: Check the type of event that triggered the function call.
+    #[cfg(target_family = "wasm")]
+    if false {
+        return Err(Error::AvmError(crate::avm2::error::error(
+            activation,
+            "Error #2176: Certain actions, such as those that display a pop-up window, may only be invoked upon user interaction, for example by a mouse click or button press.",
+            2176,
+        )?));
+    }
+
     let new_content = args
         .get(0)
         .unwrap_or(&Value::Undefined)
--- a/web/Cargo.toml
+++ b/web/Cargo.toml
@@ -61,7 +61,7 @@ features = [
     "AddEventListenerOptions", "AudioBuffer", "AudioBufferSourceNode", "AudioContext",
     "AudioDestinationNode", "AudioNode", "AudioParam", "Blob", "BlobPropertyBag",
     "ChannelMergerNode", "ChannelSplitterNode", "Element", "Event", "EventTarget", "GainNode",
-    "HtmlCanvasElement", "HtmlElement", "HtmlFormElement", "HtmlInputElement", "KeyboardEvent",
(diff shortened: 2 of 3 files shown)


core/src/avm2/globals/flash/system/system.rs
```diff
@@ -11,8 +11,17 @@ pub fn set_clipboard<'gc>(
...
 ) -> Result<Value<'gc>, Error<'gc>> {
-    //TODO: in FP9+ this function only works when called from a button handler in the Plugin due to
-    // sandbox restrictions
+    // TODO: In the plugin, this function only works when called as a result of a user interaction
+    // due to sandbox restrictions.
+    #[cfg(target_family = "wasm")]
+    if false {
+        return Err(Error::AvmError(crate::avm2::error::io_error(
+            activation,
+            "Error #2176",
```
Lord-McSweeney, line 20: nit: Provide the error message as well ("Error #2176: Certain actions, such as those that display a pop-up window, may only be invoked upon user interaction, for example by a mouse click or button press.")


<br class="Apple-interchange-newline">")

Lord-McSweeney:
Would it be a good idea to fallback to `navigator.clipboard.writeText` (or vice-versa, falling back to `execCommand`) since `execCommand` is deprecated?

n0samu:
> Would it be a good idea to fallback to `navigator.clipboard.writeText` (or vice-versa, falling back to `execCommand`) since `execCommand` is deprecated?

No reason to complicate things unnecessarily. "Deprecated" is pretty meaningless in JS - generally these functions are never actually removed

core/src/avm2/globals/flash/system/system.rs
Toad06: Thanks, I added the full message. Still worth to mention that the FP plugin only displayed `Error #2176` but this behavior likely occurs for any other error.

Toad06:
Yeah, the specification for Clipboard Events even [mentions `document.execCommand`](https://w3c.github.io/clipboard-apis/#integration-with-rich-text-editing).

But we could still reconsider adding `navigator.clipboard.writeText` when this is no longer [an unstable API in web_sys](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.Navigator.html#method.clipboard).

Lord-McSweeney:
`execCommand` is available in insecure contexts, too.

n0samu:
Noted already: https://github.com/ruffle-rs/ruffle/blob/0930841fe3dea243fd2a05f48e53f5725bd025a4/web/src/ui.rs#L59-L61

Lord-McSweeney (approved):

Dinnerbone (approved):

PR #10921, opened 2023-05-02
web: Tiny refactor of preferred renderer code

n0samu:
Small changes mostly extracted from #10835. Git really messes up displaying the diff, but all this does is:
- Return the actual error if we fail to create a player for any reason (such as failing to create a renderer). This makes the panic screen slightly more informative.
- Move the `#[cfg` lines above their respective `match` arms (slightly cleaner style)
- Move the "Unrecognized renderer name" error to a more logical location

### DIFF
--- a/web/src/lib.rs
+++ b/web/src/lib.rs
@@ -295,7 +295,7 @@ impl Ruffle {
 
             let ruffle = Ruffle::new_internal(parent, js_player, config)
                 .await
-                .map_err(|_| JsValue::from("Error creating player"))?;
+                .map_err(|err| JsValue::from(format!("Error creating player: {}", err)))?;
             Ok(JsValue::from(ruffle))
         })
     }
@@ -1444,6 +1444,8 @@ async fn create_renderer(
         if let Some(pos) = renderer_list.iter().position(|&r| r == preferred_renderer) {
             renderer_list.remove(pos);
             renderer_list.insert(0, preferred_renderer.as_str());
+        } else {
+            tracing::error!("Unrecognized renderer name: {}", preferred_renderer);
         }
     }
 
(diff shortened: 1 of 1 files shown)


web/src/lib.rs
```diff
@@ -1444,6 +1444,8 @@ async fn create_renderer(
         if let Some(pos) = renderer_list.iter().position(|&r| r == preferred_renderer) {
             renderer_list.remove(pos);
             renderer_list.insert(0, preferred_renderer.as_str());
+        } else {
+            tracing::error!("Unrecognized renderer name: {}", preferred_renderer);
```
iwannabethedev, line 1448: What will happen if the `preferred_renderer` is the empty string?

n0samu: It says `Unrecognized renderer name: `. What do you expect?

iwannabethedev:
While I do not know Rust very well, the changes look good to me.

In regards to git diffing, what I find helps is to use the flag `-w` to ignore whitespace when using `git diff`, like `git diff -w`, after I have viewed and think the diff changes including whitespace look good and if I think the rest of the changes will be shown more clearly if I instruct git to ignore whitespace. Github also supports this when viewing differences: Under the "Files changed" tab in a PR, select "Hide whitespace", and hit the apply button.
github whitespace

web/src/lib.rs
iwannabethedev: I will run the changes and test my understanding.

iwannabethedev: @n0samu Nevermind, I was in doubt because the addition of `<option value="">Automatic</option>` in #10831 would have triggered this warning if that PR only consisted of that addition, because the valid values of `preferredRenderer` would effectively be both `null`, `""` and the enum values. However, #10831 also adds a check for the empty string at https://github.com/ruffle-rs/ruffle/pull/10831/files#diff-2e3b9a8dd1da21875eb3db17443d4cc1af124bf3fc185b78f9fc926377e5feb2R1431 , and I will have to change and rebase that PR either way. So this is all good.

iwannabethedev:
The changes look good to me.

web/src/lib.rs
n0samu: #10900 is intended to supersede #10831. It automatically converts the empty string to null and vice versa, so the Rust code will never see the empty string when the extension setting is set to Automatic.

iwannabethedev: You are completely right, I completely forgot about that. It is all good then.

n0samu:
Thanks for the tip, I never noticed the "Hide whitespace" option. That makes the diff look as it should!

relrelb (approved):

PR #10926, opened 2023-05-03
web: Configuration to disable context menu on touch

danielhjacobs:
image

### DIFF
--- a/web/packages/core/src/load-options.ts
+++ b/web/packages/core/src/load-options.ts
@@ -153,6 +153,35 @@ export const enum RenderBackend {
 }
 
 /**
+ * Represents the various context menu options that are supported.
+ */
+export const enum ContextMenu {
+    /**
+     * The context menu should appear when right-clicking or long-pressing
+     * the Ruffle instance.
+     */
+    On = "on",
+
+    /**
+     * The context menu should only appear when right-clicking
+     * the Ruffle instance.
+     */
+    RightClickOnly = "rightClickOnly",
+
+    /**
+     * The context menu should only appear when long-pressing
+     * the Ruffle instance.
+     */
+    TouchOnly = "touchOnly",
+
+    /**
+     * The context menu should not appear when right-clicking or long-pressing
+     * the Ruffle instance.
+     */
+    Off = "off",
+}
+
+/**
  * Non-negative duration in seconds.
  */
 export type SecsDuration = number;
@@ -334,17 +363,9 @@ export interface BaseLoadOptions {
      * Whether or not to show a context menu when right-clicking
      * a Ruffle instance.
      *
-     * @default true
+     * @default ContextMenu.On
      */
-    contextMenu?: boolean;
-
-    /**
-     * Whether or not to show a context menu when long-pressing
-     * a Ruffle instance on touch devices.
-     *
-     * @default true
-     */
-    touchContextMenu?: boolean;
+    contextMenu?: ContextMenu | boolean;
 
     /**
      * Whether or not to show a splash screen before the SWF has loaded with Ruffle (backwards-compatibility).
--- a/web/packages/core/src/load-options.ts
+++ b/web/packages/core/src/load-options.ts
@@ -169,12 +169,6 @@ export const enum ContextMenu {
     RightClickOnly = "rightClickOnly",
 
     /**
-     * The context menu should only appear when long-pressing
-     * the Ruffle instance.
-     */
-    TouchOnly = "touchOnly",
-
-    /**
      * The context menu should not appear when right-clicking or long-pressing
      * the Ruffle instance.
      */
--- a/web/packages/core/src/load-options.ts
+++ b/web/packages/core/src/load-options.ts
@@ -339,6 +339,14 @@ export interface BaseLoadOptions {
     contextMenu?: boolean;
 
     /**
+     * Whether or not to show a context menu when long-pressing
+     * a Ruffle instance on touch devices.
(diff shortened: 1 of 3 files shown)


n0samu (approved):

web/packages/core/src/load-options.ts
```diff
@@ -338,6 +338,14 @@ export interface BaseLoadOptions {
      */
     contextMenu?: boolean;
 
+    /**
+     * Whether or not to show a context menu when long-pressing
+     * a Ruffle instance on touch devices.
+     *
+     * @default true
+     */
+    touchContextMenu?: boolean;
```
relrelb, line 347: Quite debatable, but my own preference is to avoid introduction of new configs where possible. Rather, I prefer re-usage of existing ones. That's mainly in order to keep Ruffle's API simple and clear to external embedders.

In this case, I'd suggest expanding the existing `contextMenu` config to be an enum. Previous boolean values can be kept for backward compatibility. Something like:
```ts
export interface BaseLoadOptions {
    ...
    contextMenu?: ContextMenu;
    ...
}

export const enum ContextMenu {
    Always = "always",
    AlwaysBackwardsCompatibility = true,

    NoTouch = "noTouch",

    Never = "never",
    NeverBackwardsCompatibility = false,
}
```
The terminology might not fit exactly, I'm open to ideas (perhaps on/off, enable/disable). Anyway I hope this demonstrates the point in general.

What do you think?

danielhjacobs: I'd be a bit more keen if we could get rid of the old options, but obviously that's not possible due to backwards compatibility. Making `true` and `"always"` function the same way, as well as `false` and `"never"`, will complicate the code a bit, but I suppose it's fine, so I can do that.

In terms of terminology, what about `"on"`, `"off"`, and `"rightClickOnly"`? I think I prefer a more direct "this is when the context menu will appear" naming scheme as opposed to the option defining "this is when the context menu won't appear".

While adding these options, should I add `"touchOnly"`, or do we think that would never be useful?

danielhjacobs: `const enum member initializers must be constant expressions`

danielhjacobs: https://github.com/microsoft/TypeScript/issues/32192#issuecomment-507205898

relrelb: I like your suggested terminology, guess we should proceed with it.

Regarding backwards compatibility, as already done on previous API changes, we can `console.warn()` when the boolean values are used, in hope that embedders will eventually migrate to the new values, and after a grace period we will be able to remove the old ones.

danielhjacobs: Enums cannot contain booleans, so I don't know if this can be done.

relrelb: Hmm, in that case we can define `contextMenu?: ContextMenu | boolean;`, and omit the `BackwardsCompatibility` members from the `ContextMenu` enum. Alternatively, we can take inspiration from https://github.com/microsoft/TypeScript/issues/32192#issuecomment-507484261, but that feels to hacky to me.

danielhjacobs: Done. Something else occurred to me while working on this. The `checkIfTouch` function should have an `else`, `this.isTouch = false`, right? That way a device with touch support and mouse support will know that a mouse was used to open the context menu even if the first time the context menu was opened was done by touch.

danielhjacobs: If so, should I add that to this PR?

web/packages/core/src/ruffle-player.ts
```diff
@@ -463,6 +468,15 @@ export class RufflePlayer extends HTMLElement {
                     "'{secs: 15, nanos: 0}'."
             );
         }
+        if (
+            this.loadedConfig &&
+            (this.loadedConfig.contextMenu === false ||
```
relrelb, line 473: nit: This can be simplified to `typeof this.loadedConfig.contextMenu === "boolean"`.

relrelb (approved):
LGTM, thank you!

@n0samu Waiting for your re-review.

danielhjacobs:
I can't imagine it being particularly useful either. I guess my thought is that it's sort of just a natural antithesis to "rightClickOnly", so it felt natural to add it at the same time. However, since it wasn't very easy to disable the context menu in Flash Player (supposedly it could be done in a hacky way with ExternalInterface: https://gamedev.stackexchange.com/questions/18764/how-does-one-disable-the-right-click-context-menu-in-flash), I suppose it's unlikely that it could interfere with a game's natural controls like the touch context menu. It doesn't add much complication though, the only addition it makes to the logic is:
```js
            (!this.isTouch &&
                this.loadedConfig?.contextMenu === ContextMenu.TouchOnly) ||
```
However, if it seems unnecessary, I can remove it. I had asked about it in <https://github.com/ruffle-rs/ruffle/pull/10926#discussion_r1186081789> and didn't get a direct answer one way or the other.

danielhjacobs:
I removed the option.

relrelb:
> I had asked about it in [#10926 (comment)](https://github.com/ruffle-rs/ruffle/pull/10926#discussion_r1186081789) and didn't get a direct answer one way or the other.

Sorry, somehow I missed that question... Same opinion here that we should follow the KISS principle and avoid implementing features with no clear/obvious demand.

relrelb (approved):
Thanks again! Going to merge this.

PR #10931, opened 2023-05-03
web: Parse more configs using a custom deserializer

relrelb:
Introduce and use a generic `deserialize_from_str` function. The following configs are adapted:
* `salign`
* `quality`
* `scale`
* `wmode`

Additionally, `log_level` is adapted to `deserialize_from_str as well.

### DIFF
--- a/web/src/lib.rs
+++ b/web/src/lib.rs
@@ -261,7 +261,8 @@ struct Config {
     #[serde(deserialize_with = "deserialize_from_str")]
     salign: StageAlign,
 
-    quality: Option<String>,
+    #[serde(deserialize_with = "deserialize_from_str")]
+    quality: StageQuality,
 
     scale: Option<String>,
 
@@ -547,13 +548,6 @@ impl Ruffle {
             }
         };
 
-        let default_quality = if ruffle_web_common::is_mobile_or_tablet() {
-            tracing::info!("Running on a mobile device; defaulting to low quality");
-            StageQuality::Low
-        } else {
-            StageQuality::High
-        };
-
         let trace_observer = Arc::new(RefCell::new(JsValue::UNDEFINED));
         let core = builder
             .with_log(log_adapter::WebLogBackend::new(trace_observer.clone()))
@@ -568,12 +562,7 @@ impl Ruffle {
             } else {
                 CompatibilityRules::empty()
             })
-            .with_quality(
-                config
-                    .quality
-                    .and_then(|q| StageQuality::from_str(&q).ok())
-                    .unwrap_or(default_quality),
-            )
+            .with_quality(config.quality)
             .with_scale_mode(
                 config
                     .scale
--- a/web/src/lib.rs
+++ b/web/src/lib.rs
@@ -19,8 +19,8 @@ use ruffle_core::external::{
 };
 use ruffle_core::tag_utils::SwfMovie;
 use ruffle_core::{
-    Color, Player, PlayerBuilder, PlayerEvent, SandboxType, StageScaleMode, StaticCallstack,
-    ViewportDimensions,
+    Color, Player, PlayerBuilder, PlayerEvent, SandboxType, StageAlign, StageScaleMode,
+    StaticCallstack, ViewportDimensions,
 };
 use ruffle_render::quality::StageQuality;
 use ruffle_video_software::backend::SoftwareVideoBackend;
@@ -125,6 +125,18 @@ struct JavascriptInterface {
     js_player: JavascriptPlayer,
 }
 
+fn deserialize_from_str<'de, D, T, E>(deserializer: D) -> Result<T, D::Error>
+where
+    D: serde::Deserializer<'de>,
+    T: FromStr<Err = E>,
+    E: std::fmt::Display,
+{
+    use serde::de::Error;
+
+    let value: String = serde::Deserialize::deserialize(deserializer)?;
+    T::from_str(&value).map_err(Error::custom)
+}
+
 fn deserialize_color<'de, D>(deserializer: D) -> Result<Option<Color>, D::Error>
 where
     D: serde::Deserializer<'de>,
@@ -151,15 +163,6 @@ where
         (acc << 4) | digit
     });
     Ok(Some(Color::from_rgb(rgb, 255)))
-}
-
-fn deserialize_log_level<'de, D>(deserializer: D) -> Result<tracing::Level, D::Error>
-where
-    D: serde::Deserializer<'de>,
-{
-    use serde::de::Error;
-    let value: String = serde::Deserialize::deserialize(deserializer)?;
-    tracing::Level::from_str(&value).map_err(Error::custom)
 }
 
 fn deserialize_duration<'de, D>(deserializer: D) -> Result<Duration, D::Error>
@@ -255,7 +258,8 @@ struct Config {
     #[serde(rename = "menu")]
     show_menu: bool,
 
-    salign: Option<String>,
+    #[serde(deserialize_with = "deserialize_from_str")]
+    salign: StageAlign,
 
     quality: Option<String>,
 
@@ -269,7 +273,7 @@ struct Config {
 
     warn_on_unsupported_content: bool,
 
-    #[serde(deserialize_with = "deserialize_log_level")]
+    #[serde(deserialize_with = "deserialize_from_str")]
     log_level: tracing::Level,
 
     #[serde(deserialize_with = "deserialize_duration")]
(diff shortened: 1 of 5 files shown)


web/src/lib.rs
```diff
@@ -543,13 +550,6 @@ impl Ruffle {
             }
         };
 
-        let default_quality = if ruffle_web_common::is_mobile_or_tablet() {
```
iwannabethedev, line 546: Confirming comment: The default value for `quality` in 'web/packages/core/src/load-options.ts' is `"high"`, so I imagine that this business logic is almost never run, and removing it is a clear improvement in my opinion.

iwannabethedev:
Nit/question: I do not understand Rust, but, from what I can tell, when the empty string is deserialized using `T::from_str()`, the value used is based on `#[derive(Default)]`, correct?

In either case, I think the changes look very good and can be merged as-is.

web/src/lib.rs
iwannabethedev: (If someone really wanted to add this functionality in a future PR, I would guess that the way forward might be to have two types and fields: The "configuration field" would have the current type but with an option added, like `"auto"`, and when the "internal field", with the current type, gets its value from the "configuration field", it could then use this business logic. And the default value for the "configuration field" could then be changed from `"high"` to `"auto"`).

n0samu: Looks like an oversight by @Dinnerbone. All we need to do is make `quality` one of the configuration options that is allowed to be null, and make it null by default in load-options.ts. Then we can leave this code as-is instead of removing it. Can we do that please?

iwannabethedev: Should this business logic ever run when not using the extension? I guess it is not relevant currently, since, as I understand it, there is only the desktop builds outside of the extension. But, in a theoretical case that there comes builds for mobile platforms, should this business logic also be run then? The reasoning I have is that it might be better to place this business logic elsewhere than in the 'web' crate - however, this reasoning is most likely a bit too gold-plated, and placing the business logic here in the meantime most likely makes sense.

n0samu: The intention is to use low-quality mode on web builds of Ruffle when run on mobile, presumably to improve performance on those devices (Dinnerbone could explain better though). Web builds of Ruffle include the extension and selfhosted builds.

iwannabethedev: So:
- Keep the business logic.
- Change the default value of `quality` in 'web/packages/core/src/config.ts' to `null`
- Change the type of `quality` in 'web/packages/core/src/load-options.ts' to also include `null`
- Change what places in the code needs to be fixed to handle `null` for `quality`.
- Ensure that this business logic that uses `ruffle_web_common::is_mobile_or_tablet()` is called when `null` is the value for `quality`.
- Test that `null` is handled as intended on a mobile browser and on a desktop browser.
?

Should that be done in this PR or in a separate PR? Maybe create an issue to re-add this functionality that refers to this comment?

Toad06:
This is a nice improvement in code - but probably a bit too strict as an embed like this now causes a panic:


```
<object data="whatever.swf" type="application/x-shockwave-flash" width="100" height="100">
   <param name="wmode" value="NotValid" />
   <param name="scale" value="NotValidAsWell" />
</object>
```

> Error parsing config: Error: ParseEnumError

web/src/lib.rs
Toad06: I don't have a strong opinion on whether we should keep or remove this `is_mobile_or_tablet` check but the way this function is implemented is not terrible (and the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/orientation#browser_compatibility) mention that `window.orientation` was recently added to desktop Safari?):
https://github.com/ruffle-rs/ruffle/blob/2e972b8f32c73149e0a31b650dca0598ade47bf7/web/common/src/lib.rs#L43-L52

iwannabethedev:
> This is a nice improvement in code - but probably a bit too strict as an embed like this now causes a panic:
> 
> ```
> <object data="whatever.swf" type="application/x-shockwave-flash" width="100" height="100">
>    <param name="wmode" value="NotValid" />
>    <param name="scale" value="NotValidAsWell" />
> </object>
> ```
> 
> > Error parsing config: Error: ParseEnumError

Are there old Flash applications that pass those configuration settings like that? If so, I would guess that a logged warning instead of a panic would be better.

web/src/lib.rs
iwannabethedev: There is also another place in the code that uses `is_mobile_or_tablet()`, at:

https://github.com/ruffle-rs/ruffle/blob/master/render/webgl/src/lib.rs#L204-L211

Why does it call `is_mobile_or_tablet()` instead of checking `quality`? I do not know whether `quality` is even read or used in that file.

I feel like it might make sense to create an issue that has as tasks:

- **Adding an "auto"/"default" value for `quality` configuration setting, likely with concrete value `null`:** Look into adding a default value (such as `null` as @n0samu proposes) for `quality` in the web component. And, add in checks one (or more?) places when loading configurations that takes the "auto"/"default" value and changes it to one of the existing values in https://github.com/ruffle-rs/ruffle/blob/master/render/src/quality.rs , likely in 'web/src/lib.rs', possibly elsewhere as well where configurations are loaded. This check could potentially use hardware detection or try to detect and use whether the hardware being run on is a mobile device (like, if it is mobile, have "low" `quality` as default, if it is not mobile, have "high" `quality` as default).
- **Consider `is_mobile_or_tablet()`, its implementation and usage, and whether `quality` is sufficient or whether extra functions or values like `is_mobile_or_tablet()` is needed when deciding upon internal graphics settings:** `is_mobile_or_tablet()` is currently used in two places in the code, in one place it is/was used (at https://github.com/ruffle-rs/ruffle/blob/5b8e0df83fdf02b8abc5ddd342e05576019c9d67/web/src/lib.rs#L546 ) to help select a value for `quality`, and in another place it is used to decide upon a graphics quality level without using `quality` (at https://github.com/ruffle-rs/ruffle/blob/5b8e0df83fdf02b8abc5ddd342e05576019c9d67/render/webgl/src/lib.rs#L206 ). This seems like very different kinds of usages, namely: Configuration loading (deciding upon a `quality` configuration value when loading in configurations); Internal graphics settings reading (read the configuration fields and call detection functions to decide graphics settings like MSAA sample count). `is_mobile_or_tablet()`'s implementation may also have issues. If `quality` is not considered sufficient for "Internal graphics settings reading", consider what to do then.

I also feel like this issue is getting sufficiently complex to be beyond the scope of this PR, now that we have dug into it.

Would it be alright if:
- We accept the PR as it currently is.
- I make an issue with the above two tasks.
?

n0samu: Wait, so if `render/webgl/src/lib.rs` uses the `is_mobile_or_tablet()` function, and this PR removes that function... does this PR break the WebGL renderer in some way? Or am I missing something?

@Dinnerbone Is there a reason why the WebGL renderer is checking `is_mobile_or_tablet()` instead of the quality setting?

I think we should wait for Dinnerbone's input before deciding what to do because there's a fair bit of information that we're missing right now.

n0samu:
When polyfilling Flash embeds, our goal should be to accept everything that Flash accepts. @relrelb do you think the approach you're taking in this PR allows for this?

web/src/lib.rs
iwannabethedev: I think this PR only removes one usage of it here, not the function itself, though there is some discussion about removing the function itself. I think you are right about waiting for Dinnerbone.

relrelb:
> When polyfilling Flash embeds, our goal should be to accept everything that Flash accepts. @relrelb do you think the approach you're taking in this PR allows for this?

Agree we should be lenient as possible when polyfilling Flash embeds, and I wasn't aware of that. Unless anyone has a quick solution, I guess this PR needs re-thinking (also due to the `quality` discussion above).

danielhjacobs:
> This is a nice improvement in code - but probably a bit too strict as an embed like this now causes a panic:
> 
> ```
> <object data="whatever.swf" type="application/x-shockwave-flash" width="100" height="100">
>    <param name="wmode" value="NotValid" />
>    <param name="scale" value="NotValidAsWell" />
> </object>
> ```
> 
> > Error parsing config: Error: ParseEnumError

This is especially bad since "autohigh" and "autolow" are allowed quality values in Flash (https://helpx.adobe.com/flash/kb/flash-object-embed-tag-attributes.html) but Ruffle does not support them. Any embeds that use them will break with this PR, whereas before this PR they would default to high quality, which is fine.

Dinnerbone:
Superseded by #16612

PR #10935, opened 2023-05-04
avm2: Implement 'getQualifiedDefinitionNames' for ApplicationDomain

friedkeenan:
This adds a `get_defined_names` method to `Domain` which returns a `Vec` of `QName`s, which accesses the `defs` attribute, iterating over it and constructing the `QName`s. Then in the implementation of the `getQualifiedDefinitionNames` method, `get_defined_names` is called, and the result of that is filtered so only names with non-private namespaces are used, and then it gets shoved into a `VectorObject` and returned.

In my tests, even classes marked `internal` from other packages were returned from `getQualifiedDefinitionNames`, despite the documentation only stating that "public" definitions are returned. That is the reason why I check that the namespace is not private instead of just checking of it's public.

Something to note is that the documentation for `getQualifiedDefinitionNames` states that a `SecurityError` is raised if "The definition belongs to a domain to which the calling code does not have access", however ruffle does not seem to consider such things at present, and I was told not to worry too much about it. I have noted this deficiency in the rust implementation.

`getQualifiedDefinitionNames` is also normally only available for Flash Player 11.3+, however ruffle does not seem to have any mechanism for restricting its existence/use to only certain flash versions, and so I haven't restricted it at all with regards to that. I left a note in the documentation of the rust implementation of the method, I'm not sure if that's the best place for such a thing.

I should also say that this is my first real rust code so if things seem weird, that's probably why, and please let me know.


core/src/avm2/domain.rs
```diff
@@ -219,6 +219,15 @@ impl<'gc> Domain<'gc> {
         res
     }
 
+    pub fn get_defined_names(&self) -> Vec<QName<'gc>> {
```
Aaron1011, line 222: This could return an `impl Iterator<Item = QName<'gc>>` to avoid the need to allocate a `Vec`

friedkeenan: I had tried that, but the compiler told me that a type annotation was needed for the `collect` call and to consider specifying `Vec`. And when I tried to skip the `collect`, the compiler told me that I was capturing a lifetime that "does not appear in bounds", which I believe makes sense? So I settled for specifying `Vec`, despite agreeing that the allocation is unfortunate. Would you have more advice towards this? I feel my lack of rust experience is showing

Aaron1011: @friedkeenan I didn't realize that there was an intermediate `read()` call. It's probably more trouble that it's worth to make an `impl Iterator` work here.

Aaron1011 (approved):

PR #10938, opened 2023-05-04
avm2: Stub rest of Socket's methods

friedkeenan:
I need stubs for `flush` and `writeBytes` for my tool [friedkeenan/tfm-secrets-leaker](https://github.com/friedkeenan/tfm-secrets-leaker) which overrides those methods.

I looked through the project to see how to make a test for stubbed methods but from my searches it did not seem like stubs are tested at all, so I neglected to add any tests. I can confirm though that it works.

Also it seems like there were some lines with only whitespace that my editor removed upon saving.

### DIFF
--- a/core/src/avm2/globals/flash/net/Socket.as
+++ b/core/src/avm2/globals/flash/net/Socket.as
@@ -5,6 +5,7 @@ package flash.net {
 
     import __ruffle__.stub_method;
     import __ruffle__.stub_getter;
+    import __ruffle__.stub_setter;
 
     public class Socket extends EventDispatcher {
         private var _timeout:uint;
@@ -39,6 +40,11 @@ package flash.net {
             stub_method("flash.net.Socket", "close");
         }
 
+        public function get bytesAvailable():uint {
+            stub_getter("flash.net.Socket", "bytesAvailable");
+            return 0;
+        }
+
         public function get bytesPending():uint {
             stub_getter("flash.net.Socket", "bytesPending");
             return 0;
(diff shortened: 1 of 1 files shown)


relrelb (approved):
Looks good, thank you and welcome! While at it, would you like to stub the remaining `flash.net.Socket` methods (in the current or a future PR - for your preference)?
The full method list can be found in the docs: https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/net/Socket.html

friedkeenan:
I could stub the other methods/properties too. Should I include AIR-exclusive properties (there appears to be no AIR-exclusive methods)?

relrelb:
> I could stub the other methods/properties too. Should I include AIR-exclusive properties (there appears to be no AIR-exclusive methods)?

We generally don't try to emulate AIR-specific functionality. So just those included in Flash Player are fine.

friedkeenan:
Stubbed the rest of `Socket`. Wasn't quite sure what to return for `readObject`; I return `null` but would `{}` be more appropriate?

friedkeenan:
The error in the CI is because I named the parameter for `writeObject` as `object`, which is what the docs report, but changing it to `value` fixed the error locally. I committed and pushed this change, however Github was choking up at the time and the commit doesn't seem to have been pushed through to this PR despite [existing on my branch](https://github.com/friedkeenan/ruffle/commit/f00b41c0890cd80b0b6ce9e52670ecd9590f9f96)

relrelb:
Thank you!

PR #10939, opened 2023-05-04
extension: Localize more strings

Toad06:


web/packages/extension/src/options.ts
```diff
@@ -3,5 +3,7 @@ import { bindOptions } from "./common";
 
 window.addEventListener("DOMContentLoaded", () => {
     document.title = utils.i18n.getMessage("settings_page");
+    document.querySelector("#advanced_options")!.textContent =
```
relrelb, line 6: nit: Personally I prefer the good old `document.getElementById()` over `document.querySelector("#some_id")`. But that's debatable of course.

Dinnerbone:
If we're not using the enum values, it may make sense to use the plural form of these; "Errors", "Warnings", etc

PR #10941, opened 2023-05-04
[EXPERIMENT] `gc_extra`: Fine-grained interior mutability inside GC'd objects

moulins:
Introduce a new `ruffle_gc_extra` crate, which is an attempt at providing a more flexible API for mutable GC'd objects, without limiting each object to a binary choice of "fully immutable" (`Gc`) and "fully mutable" (`GcCell`). The design is very close to what I designed ~6 months ago in [this `gc-arena` PR](https://github.com/kyren/gc-arena/pull/33), but forgoes the complicated proc-macro machinery in favor of simpler (though slightly less ergonomic) declarative macros.

I've made the choice of doing development in-tree instead of separately in `gc-arena`, as the proposal can be fully implemented on top of the public `gc-arena` API, and using Ruffle's code as a test bed is a great way to iterate on the API design.

--------

In theory, the finer-grained mutability should allow us to:
- reduce the number of write barriers, and so the need for passing `MutationContext`s around (at least when modifying `'static` data inside a GC'd object);
- use single-field `Cell`s more extensively, reducing the number of runtime borrows;
- replace contorsions forced upon us by `GcCell` (e.g. double indirections) by more straight-forward designs.

The hope of this experiment is that this all pans out in practice, without requiring too much ceremony and/or unsafe code outside the core abstractions.

Once I'm happy with the results, I intend to adapt the content `ruffle_gc_extra` so that it can be integrated in `gc-arena` proper, and then mark this PR ready for merging.


moulins:
Closing this PR; the mutation APIs have been integrated in `gc-arena` (see #11182). The various "example" refactors of this PR will be the object of fresh new PRs in the future.

PR #10945, opened 2023-05-05
avm2: Work on flash.system.Capabilities

### LINKED ISSUE #9180: Player version config doesn't work in AVM2
### Describe the bug

Currently in AVM2, the player version is hard-coded. This means that the `playerVersion` config and `--player-version` flag do nothing in an AVM2 movie.

### Expected behavior

Player version config affects what the movie sees as the Flash player version.

### Affected platform

Desktop app

### Operating system

All

### Browser

_No response_

### Additional information

_No response_

Lord-McSweeney:
Eventually, code from #10469 can be reused to implement the rest of the `flash.system.Capabilities` getters.

Fixes #9180.

### DIFF
--- a/core/src/avm2/globals/flash/system/capabilities.rs
+++ b/core/src/avm2/globals/flash/system/capabilities.rs
@@ -1,0 +1,33 @@
+//! `flash.display.Capabilities` native methods
+
+use crate::avm2::{Activation, AvmString, Error, Object, Value};
+
+/// Implements `flash.system.Capabilities.version`
+pub fn get_version<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    _this: Option<Object<'gc>>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    // TODO: Report the correct OS instead of always reporting Linux
+    Ok(AvmString::new_utf8(
+        activation.context.gc_context,
+        format!("LNX {},0,0,0", activation.avm2().player_version),
+    )
+    .into())
+}
+
+/// Implements `flash.system.Capabilities.playerType`
+pub fn get_player_type<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    _this: Option<Object<'gc>>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    // TODO: When should "External" be returned?
+    let player_type = if cfg!(target_family = "wasm") {
+        "PlugIn"
+    } else {
+        "StandAlone"
+    };
+
+    Ok(AvmString::new_utf8(activation.context.gc_context, player_type).into())
+}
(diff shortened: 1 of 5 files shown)


core/src/avm2/globals/flash/system/capabilities.rs
```diff
@@ -0,0 +1,33 @@
...
+    activation: &mut Activation<'_, 'gc>,
+    _this: Option<Object<'gc>>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    // TODO: When should "External" be returned?
+    let player_type = if cfg!(target_family = "wasm") {
+        "PlugIn"
+    } else {
+        "StandAlone"
+    };
```
iwannabethedev, line 30: Would it be a bad idea or helpful to insert this link as a comment?:

https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/system/Capabilities.html#playerType

Or is it implied?

iwannabethedev: (I am not certain, but, I assume that string literals are fine to use here, since I would assume that the values are only used in one place in the codebase and only used by ActionScript applications).

core/src/avm2/globals/flash/system/capabilities.rs
```diff
@@ -0,0 +1,33 @@
...
+
+use crate::avm2::{Activation, AvmString, Error, Object, Value};
+
+/// Implements `flash.system.Capabilities.version`
+pub fn get_version<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    _this: Option<Object<'gc>>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    // TODO: Report the correct OS instead of always reporting Linux
```
iwannabethedev, line 11: (I would guess that one could make an issue for fully implementing this, but, the application developer should not be relying on this either way (it even says so in the official documentation), so I do not believe adding an issue for this is necessary. There are stubs for the other methods as well).

iwannabethedev:
I do not understand Rust or this piece of Ruffle well enough to really give good feedback, but it looks good to me overall.

I imagine that this can be complicated to test in an automated way. Have you tested the two new properties manually?

I think someone else should review as well, but otherwise this looks good to me as is. You are welcome to close the comments that I made once you have read them.

core/src/avm2/globals/flash/system/capabilities.rs
n0samu: > Would it be a bad idea or helpful to insert this link as a comment?

We don't usually do that and there's no particular reason to do that here

Lord-McSweeney:
I've tested both the properties, with `--player-type`, and compiling this branch and running a SWF in the browser/desktop app (which correctly reports `StandAlone` and `PlugIn`).

iwannabethedev:
I believe that @n0samu thinks this looks good, and I myself think it looks good (with my limited insight into Rust and this part of the codebase), so I believe it is ready to be merged once rebased.

core/src/avm2/globals/flash/system/capabilities.rs
Lord-McSweeney: It's just a TODO notice to state that "this is not fully implemented yet" without marking it as a stub because the most important functionality is implemented.

Aaron1011 (approved):

PR #10950, opened 2023-05-05
web: Add else to checkIfTouch for devices with mouse & touch support

danielhjacobs:


web/packages/core/src/ruffle-player.ts
```diff
@@ -883,6 +883,8 @@ export class RufflePlayer extends HTMLElement {
     private checkIfTouch(event: PointerEvent): void {
         if (event.pointerType === "touch" || event.pointerType === "pen") {
```
iwannabethedev, line 884: Given the addition of the `else` branch, could this be simplified to:

```typescript
    this.isTouch = event.pointerType === "touch" || event.pointerType === "pen";
```

?

iwannabethedev:
The changes look good to me.

web/packages/core/src/ruffle-player.ts
```diff
@@ -908,9 +908,8 @@ export class RufflePlayer extends HTMLElement {
     }
 
     private checkIfTouch(event: PointerEvent): void {
-        if (event.pointerType === "touch" || event.pointerType === "pen") {
-            this.isTouch = true;
-        }
+        this.isTouch =
```
relrelb, line 911: I'm not sure what scenario this is supposed to fix. Doesn't it mean that on devices with both touch and mouse support, a right-click will effectively hide the "Hide this menu" context menu item? If so, I wonder if we could just "inline" this condition into `showContextMenu()`, which is more direct and also allows the removal of the `isTouch` property.

danielhjacobs: It was mainly a follow-up to https://github.com/ruffle-rs/ruffle/pull/10926.  Without this change, I'm pretty sure if you use the configuration option rightClickOnly and then touch a Ruffle instance on a device with both touch and mouse support, then a right click will not open the context menu because isTouch is true and contextMenu is rightClickOnly. I don't have a touch device with mouse support to test that but it intuitively makes sense to me.

danielhjacobs: It also fixes hide this menu as mentioned though.

relrelb: If so, how about making `isTouch` a local in `showContextMenu()`? Since if I understand correctly, whether a context menu is opened by mouse or touch is completely local to that event, and actually no state across events needs to be preserved at all.

danielhjacobs: I could, yeah. That would require passing it with `contextMenuItems(isTouch)` too, right? I'll look into that tomorrow.

danielhjacobs: Property 'pointerType' does not exist on type 'MouseEvent'

That's because the contextmenu event is a MouseEvent, not a pointer event.

danielhjacobs: Though actually, perhaps `const isTouch = event instanceof PointerEvent;` is all that's really needed.

n0samu: > whether a context menu is opened by mouse or touch is completely local to that event, and actually no state across events needs to be preserved at all.

I didn't notice this until now, but for the record, this is false. `contextmenu` is a MouseEvent, and `pointerType` is a property of PointerEvent. So this suggestion completely broke touch device detection. I'm opening a PR that reverts this and goes back to Daniel's original approach.

danielhjacobs: > Property 'pointerType' does not exist on type 'MouseEvent'

I definitely should have thought a bit further about the fundamental reasons npm was complaining instead of bashing my head on the keyboard until npm stopped complaining lol.

PR #10953, opened 2023-05-05
wgpu: Allow unaligned writes to IndexBuffer3D

Aaron1011:
wgpu requires buffer copy sizes and offsets to be 4-byte aligned. Unfortunately, ActionScript can perform 2-byte aligned uploads into an IndexBuffer3D.

To support this, we now keep a copy of the IndexBuffer3D on the CPU. When performing an upload to the buffer, we round the offset down and the size up to the nearest 4-byte aligned value. The cpu buffer is used to fill out the write with existing data, so that we don't corrupt the contents of the GPU buffer.

To avoid introducing a new RefCell, I've changed IndexBuffer3D to use a `Box` instead of an `Rc` to store the trait object. This allows us to pass a mutable reference down to the backend.


Dinnerbone:
I'd appreciate a test that uses this, it feels like something that could break if we're not paying attention

Aaron1011:
@Dinnerbone I copied the triangle test, and changed it to upload to the index buffer several times (including at unaligned offsets).

tests/tests/swfs/avm2/stage3d_triangle_index_upload/test.toml
```diff
@@ -0,0 +1,7 @@
+num_frames = 1
+
+[image_comparison]
+tolerance = 1
+
+[player_options]
+with_renderer = { optional = true, sample_count = 1 }
```
Dinnerbone, line 7: nit: Make this required

Dinnerbone (approved):

PR #7577, opened 2022-08-04
web: Add metadata to error info and add Copy Debug Info context menu item

danielhjacobs:
Edit: No longer clashes with #7534. Now just adds a new context menu item, and adds metadata to the panic info. The function that displayed the orange warning screen was removed in the third commit. That commit can be reverted if that function is desired, but "Copy Debug Info" fulfills a similar purpose.

PR #2505, opened 2021-01-14
web: Remove event listeners when a Ruffle instance is destroyed

Toad06:
Fixes #1263. Should also resolve #876, as I believe this was the same problem.

PR #2318, opened 2021-01-07
web: Prevent "414 URI Too Long" error on Github when reporting bugs

Toad06:
githh

The stack error is sometimes very long, so long that the generated URL could trigger a 414 error on Github. This is especially true on Firefox: for instance, this problem happens with the game in #2280. With PR #2305, this will likely happen more often, so we need to strip the stack error from the URL.

PR #10958, opened 2023-05-06
web: Some panic-related cleanups

relrelb:


web/packages/core/src/ruffle-player.ts
```diff
@@ -148,7 +148,7 @@ export class RufflePlayer extends HTMLElement {
     private _readyState: ReadyState;
 
     private panicked = false;
-    private _cachedDebugInfo: string | null = null;
+    private debugInfo = "";
```
iwannabethedev, line 151: Nit/question: Might it make sense to rename this to `rendererDebugInfo`? Or would it be better to keep its current name, in case other debug information than rendering is added to it later?

web/packages/core/src/ruffle-player.ts
```diff
@@ -1588,37 +1588,40 @@ export class RufflePlayer extends HTMLElement {
      * @returns A string containing all the data included in the panic.
      */
     private getPanicData(): string {
-        const dataArray = [];
```
iwannabethedev, line 1591: (Nit/comment: I imagine that the original thinking behind creating an array of strings and then joining them at the end might be to avoid the cost of repeated string concatenation. But, I believe that the number and length of the string values here are limited, and that `getPanicData()` is a method that is called very infrequently if at all, so the performance of the string concatenation here does not really matter and I think the changes makes the code cleaner).

web/src/lib.rs
```diff
@@ -1286,6 +1167,116 @@ impl RuffleInstance {
...
+        self.canvas.remove();
+
+        // Stop all audio playing from the instance.
+        let _ = self.with_core_mut(|core| {
+            core.audio_mut().stop_all_sounds();
+            core.flush_shared_objects();
+        });
+
+        // Clean up all event listeners.
+        if let Some(mouse_move_callback) = self.mouse_move_callback.take() {
```
iwannabethedev, line 1181: (Nit/Idea: It might not be viable or a good idea, but, I feel like one could abstract away most of these statements.

Like, if you stored the different data in an array like this (not valid Rust code):

```rust
let event_listener_description = [
    (self.mouse_move_callback, self.canvas, "pointermove"),
    (self.mouse_enter_callback, self.canvas, "pointerenter"),
    (self.mouse_leave_callback, self.canvas, "pointerleave"),
    ...
]
```

And then implemented a method that iterated over it, like a `foreach` or similar, that might look like this (not valid Rust code):

```rust
event_listener_description.foreach(|(callback, target, event)| {
       if let Some(callb) = callback.take() {
            target
                .remove_event_listener_with_callback(
                    event,
                    callb.as_ref().unchecked_ref(),
                )
                .warn_on_error();
        }
})
```

Though, even if there may be merit to this refactoring, it might be dangerous to do now, and it might be better suited to a later PR. For instance, it would not be possible with just 3 parameters to do with `window_mouse_down_callback`, since it passes in an extra `true` argument, and during a refactoring, that might be lost. Care would at the very least be required.

Please close this comment once you have read it).

web/src/lib.rs
```diff
@@ -435,127 +435,8 @@ impl Ruffle {
...
-                instance
-                    .window
-                    .cancel_animation_frame(id.into())
-                    .warn_on_error();
-            }
-        }
-
-        // Player is dropped at this point.
+        let _ = self.remove_instance();
+        // Instance is dropped at this point.
```
iwannabethedev, line 439: (Nit/question: I do not know Rust, but, like for RAII in C++, is `drop()` called once the scope here is exited? If so, that is rather neat).

web/packages/core/src/ruffle-player.ts
```diff
@@ -1991,21 +1994,7 @@ export class RufflePlayer extends HTMLElement {
     }
 
     protected debugPlayerInfo(): string {
```
iwannabethedev, line 1996: Is it possible to have a `RufflePlayer` instance that is not also a `RuffleObject` instance or some other instance of a subtype? If so, I wonder if it might be nice to have some sort of string here that describes what it is (like `"Player type: Object\n"` in ruffle-object.ts). But, https://github.com/ruffle-rs/ruffle/blob/27738a492e3ea01d9e5fdd5a99147a15dac422ed/web/packages/core/src/ruffle-object.ts#L173 calls `super.debugPlayerInfo();`. I do not know what would be best here, but, your changes here are overall good as far as I can tell.

(I personally like to avoid inheritance, and prefer composition and the like).

iwannabethedev:
I mostly have nits, only one real question. I know very little Rust, so I cannot comment much on the Rust part. Overall, the changes look good to me, and I assume that you have tested them lightly manually (or through existing automated tests if they cover the changes).  I think it would be good to have another reviewer as well.

web/packages/core/src/ruffle-player.ts
Toad06: No, this was done to prevent issues when reporting a bug (#2318). Some parts later moved to the new function `getPanicData` (#7577).

web/src/lib.rs
Toad06: Mike suggested this a while ago (in https://github.com/ruffle-rs/ruffle/pull/2505#pullrequestreview-568626038):
> I think better is to make an explicit `EventDispatcher`/`EventListener` struct in Rust that organizes a lot of this code up, and can have a `Drop` impl that auto-removes the listener.

web/packages/core/src/ruffle-player.ts
iwannabethedev: Huh. I think I would have just made an interface like:

```typescript
interface ErrorData {
    errorText: string;
    stackIndex: number;
}
```

or something, instead of adding `stackIndex` to an array. But I have only skimmed through the code, and this might be way off.

But I would guess that this means that this part of the PR (still) looks good, I believe.

web/src/lib.rs
iwannabethedev: (Did you mean to reply to https://github.com/ruffle-rs/ruffle/pull/10958#discussion_r1186915849 ?)

relrelb: I had the same thought when touching that code, but I prefer to do that in a separate PR.

web/src/lib.rs
relrelb: Exactly. You can learn more about `Drop` [here](https://doc.rust-lang.org/std/ops/trait.Drop.html) and [here](https://doc.rust-lang.org/rust-by-example/trait/drop.html).

web/packages/core/src/ruffle-player.ts
relrelb: Renamed to `rendererDebugInfo`.

iwannabethedev:
The changes look good to me.

Toad06 (approved):

PR #10966, opened 2023-05-07
avm2: Implement PrintJobOptions

huwdp:
Implements PrintJobOptions class :)


Dinnerbone (approved):

Lord-McSweeney:
`pixelsPerInch` and `printMethod` are AIR-only properties, so I think those should be removed.

huwdp:
> `pixelsPerInch` and `printMethod` are AIR-only properties, so I think those should be removed.

This is done. I also implemented the getter/setter

Lord-McSweeney:
It seems that Flash doesn't actually implement a getter/setter for this (it's a normal property). Also, there are unused imports for `__ruffle__.stub_getter`/`setter`.

huwdp:
> It seems that Flash doesn't actually implement a getter/setter for this (it's a normal property since there are no restrictions on it). Also, there are unused imports for `__ruffle__.stub_getter`/`setter`.

Thankf for this. I've made the changes

Lord-McSweeney (approved):

PR #10970, opened 2023-05-07
desktop: Add volume option to CLI

RasmusMathiesen:
Simple addition to let desktop users adjust global audio volume when starting ruffle from the command line.

### DIFF
--- a/desktop/src/main.rs
+++ b/desktop/src/main.rs
@@ -20,6 +20,7 @@ use anyhow::{anyhow, Context, Error};
 use clap::Parser;
 use isahc::{config::RedirectPolicy, prelude::*, HttpClient};
 use rfd::FileDialog;
+use ruffle_core::backend::audio::AudioBackend;
 use ruffle_core::backend::navigator::OpenURLMode;
 use ruffle_core::{
     config::Letterbox, events::KeyCode, tag_utils::SwfMovie, LoadBehavior, Player, PlayerBuilder,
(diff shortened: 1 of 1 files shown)


desktop/src/main.rs
```diff
@@ -109,6 +110,10 @@ struct Opt {
     #[clap(long, short, default_value = "show-all")]
     scale: StageScaleMode,
 
+    /// Audio volume
```
n0samu, line 113: Let's add a little more information here - something like this:
`/// The audio volume as a number between 0 (muted) and 1 (full volume)`

n0samu (approved):
Welcome, and thank you for the contribution! Just have a little nit on the documentation.

RasmusMathiesen:
Thank you :-) I added your suggestion.

desktop/src/main.rs
```diff
@@ -109,6 +110,10 @@ struct Opt {
     #[clap(long, short, default_value = "show-all")]
     scale: StageScaleMode,
 
+    /// Audio volume as a number between 0 (muted) and 1 (full volume)
+    #[clap(long, short, default_value = "1.0")]
+    volume: f32,
```
relrelb, line 115: Thinking out loud - Does it make more sense to define this as an integer in range 0-100? Feels a bit more user-friendly to me.

n0samu: You would need to also make that change for the `volume` property of the player on web, and think about how to do that without causing unexpected behavior in people's existing sites... let's leave this PR as-is

relrelb (approved):
Thank you!

relrelb (approved):
Thank you!

PR #10971, opened 2023-05-07
avm1: `button.useHandCursor` can hold any value

Toad06:
This is a follow-up to #10865.

### DIFF
--- a/core/src/display_object/avm1_button.rs
+++ b/core/src/display_object/avm1_button.rs
@@ -43,7 +43,6 @@ pub struct Avm1ButtonData<'gc> {
     object: Option<Object<'gc>>,
     initialized: bool,
     has_focus: bool,
-    use_hand_cursor: bool,
 }
 
 impl<'gc> Avm1Button<'gc> {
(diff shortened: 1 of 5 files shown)


core/src/display_object/avm1_button.rs
```diff
@@ -195,31 +193,36 @@ impl<'gc> Avm1Button<'gc> {
         }
     }
 
-    pub fn enabled(self, context: &mut UpdateContext<'_, 'gc>) -> bool {
-        if let Some(object) = self.0.read().object {
+    fn get_property_as_bool(
```
relrelb, line 196: nit: How about renaming this function to `get_boolean_property`? (Same applies to `movie_clip.rs`)
Also I suggest renaming the `is_undefined_truthy` parameter to `default` - IMO it better reflects its meanings.

Toad06: Good idea, I've done the changes. Also, thank you for the review.

relrelb (approved):
Changes look good, thank you!
I start to think that `movie_clip.rs` and `avm1_button.rs` are more similar than different. They probably should share code somehow, but that's out of the PR scope.

PR #10979, opened 2023-05-08
web: Download all sols as single .zip

Toad06:
Clicking "Backup all saves (download all sols)" in the Save Manager currently triggers a download prompt for each available sol. This can be a bit annoying when you have dozens or hundreds of saves. This PR changes this behavior by placing the saves in a ZIP file.

### DIFF
--- a/web/packages/core/package.json
+++ b/web/packages/core/package.json
@@ -41,7 +41,7 @@
         "test": "cross-env TS_NODE_COMPILER_OPTIONS={\\\"module\\\":\\\"commonjs\\\",\\\"verbatimModuleSyntax\\\":false} mocha"
     },
     "dependencies": {
-        "@zip.js/zip.js": "^2.7.6",
+        "jszip": "^3.10.1",
         "wasm-feature-detect": "^1.5.1"
     },
     "devDependencies": {
--- a/web/packages/core/package.json
+++ b/web/packages/core/package.json
@@ -41,6 +41,7 @@
         "test": "cross-env TS_NODE_COMPILER_OPTIONS={\\\"module\\\":\\\"commonjs\\\",\\\"verbatimModuleSyntax\\\":false} mocha"
     },
     "dependencies": {
+        "@zip.js/zip.js": "^2.7.6",
         "wasm-feature-detect": "^1.5.1"
     },
     "devDependencies": {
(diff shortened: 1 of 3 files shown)


web/packages/core/package.json
```diff
@@ -41,6 +41,7 @@
         "test": "cross-env TS_NODE_COMPILER_OPTIONS={\\\"module\\\":\\\"commonjs\\\",\\\"verbatimModuleSyntax\\\":false} mocha"
     },
     "dependencies": {
+        "@zip.js/zip.js": "^2.7.6",
```
relrelb, line 44: Is there any particular reason for choosing this library? I see [`jszip`](https://www.npmjs.com/package/jszip) is much more popular (6M vs. 50K weekly downloads).

web/packages/core/src/ruffle-player.ts
```diff
@@ -897,12 +898,28 @@ export class RufflePlayer extends HTMLElement {
...
         this.instance?.set_fullscreen(this.isFullscreen);
     }
 
+    /**
+     * Prompt the user to download a file.
+     *
+     * @param href The content to download.
+     * @param name The name to give the file.
+     */
+    private saveFile(href: string, name: string): void {
```
relrelb, line 913: All `saveFile()` calls are preceded by `const blobURL = URL.createObjectURL(blob);` and followed by `URL.revokeObjectURL(blobURL);`. How about pulling that logic inside `saveFile()` itself, making it accept a `blob: Blob` in place of `href: string`?

web/packages/core/package.json
Toad06: Well... No, there's no particular reason. :) `jszip` is indeed much more popular, so I can switch to this library if you want. The development of `@zip.js/zip.js` seems more active though (last publish was a day ago, vs 9 months ago for `jszip`) but I don't know if that matters.

Toad06: I haven't noticed until now but this makes `ruffle.js` (when you compile) grow from ~114kb to ~252kb. Though the file size is still acceptable, this may be a bit too much for such a feature. `jszip` seems smaller in size, so it could be interesting to give it a try.

relrelb: Thanks for checking! My preference is to keep file sizes (both `.js` and `.wasm`) low as possible, because it directly affects download times on Web, so let's stay with `jszip`.

relrelb (approved):

relrelb:
@danielhjacobs This looks good to me, but since you have experience with this portion of the code, I'd appreciate your approval as well.

danielhjacobs (approved):
I think it looks good too.

Toad06:
Thank you guys for the review.

PR #10982, opened 2023-05-09
core: Remove most uses of context.is_action_script_3()

Lord-McSweeney:
Replace with `[current movie].is_action_script_3()`. This allows mixed AVM code to run, but mouse pick doesn't work in AVM1 and AVM2 movies don't seem to be loaded. However, it should no longer panic.

I'm also almost sure that the mouse pick implementation is incorrect.


Lord-McSweeney:
Closing in favor of #11005.

PR #10992, opened 2023-05-10
avm2: Repsect SKIP_INVISIBLE for DisplayObject

Aaron1011:
Previously, we allowed mouse picks of certain invisible DisplayObjects, even when HitTestOptions::SKIP_INVISIBLE was set.


tests/tests/swfs/avm2/click_invisible/Test.as
```diff
@@ -0,0 +1,46 @@
+﻿package {
```
iwannabethedev, line 1: (Nit: I do not know if there exists any auto-formatter for ActionScript, so I manually indented this file:

https://gist.github.com/iwannabethedev/00e73648a53f137873c2d8f229f6d30a

I did it quickly just by `Shift+Tab`'ing the whole selection a couple of times in a text editor like Gedit, and then `Tab` selections a few times as appropriate.)

tests/tests/swfs/avm2/click_invisible/Test.as
```diff
@@ -0,0 +1,46 @@
...
+			var shape = new Shape();
+			shape.graphics.beginFill(0xFF0000);
+			shape.graphics.drawRect(0, 0, 100, 100);
+			shape.graphics.endFill();
+			shapeContainer.addChild(shape);
+			this.addChild(shapeContainer);
+			
+			var bitmapContainer = new Sprite();
+			bitmapContainer.name = "MyBitmapContainer";
+			var bitmapData = new BitmapData(120, 120, false, 0);
```
iwannabethedev, line 24: (Very minor nit, consider ignoring: Is `0` transparent black or opaque black? Maybe write in hex-code?).

iwannabethedev:
The changes look good to me (though I have limited knowledge of Ruffle and ActionScript). Assuming that you tested temporarily with setting `visible == true` and that the mouse clicks gave hit results, then I think this is ready to be merged once someone else have reviewed as well.

It is up to you whether to reformat the test file, I have linked to a gist that should have fixed spaces in it.

Bale001 (approved):
This LGTM, I do think @iwannabethedev reformatted test looks better if you want to change it, but I don't really mind either way.

PR #11001, opened 2023-05-10
web: Simplify a source string

Toad06:
Some users are confused about what a "sol" is (see https://discord.com/channels/610531541889581066/1099002553661919232/1105851375126069260).

### DIFF
--- a/web/packages/core/texts/en-US/save-manager.ftl
+++ b/web/packages/core/texts/en-US/save-manager.ftl
@@ -7,4 +7,4 @@ save-reload-prompt =
 save-download = Download
 save-replace = Replace
 save-delete = Delete
-save-backup-all = Backup all saves (download all sols)
+save-backup-all = Download all save files


n0samu:
Can we say "save files" ?

iwannabethedev:
I did not know what SOL meant either, but, if it was renamed to: `Backup all saves (download all .sol Flash cookie files)`, it would be clearer to me.

For reasons I do not know, when I searched for "sol flash wikipedia", the Wikipedia page did not mention "SOL" in the text, only "LSO", despite several of the tools linked on that Wikipedia page using "SOL".

iwannabethedev:
But I think `Download all save files` is fine as well.

Maybe `Download all save files (.sol files)` ?

n0samu (approved):
Looks good, I don't think any further change is needed

danielhjacobs (approved):

Dinnerbone (approved):

PR #11002, opened 2023-05-10
avm2: Throw error for invalid Stage3D back buffer size

Aaron1011:


relrelb (approved):

Aaron1011:
It turns out that this only happens when t4h swf version is >=19

Aaron1011:
I'm not sure why I thought that - when I re-tested, the error message is just different with an SWF version < 19.

PR #11003, opened 2023-05-10
avm2: Minor refactoring of duplicated code.

iwannabethedev:
(Edited to also refactor 5 extra places in the code, to a total of 8 places refactored - Edited again to add 1 extra place to a total of 9, for `pixel_dissolve()`).

The code should contain no change in behavior. The ~~8~~ 9 places in the code that has been replaced should be exactly the same apart from variable names.

There are two functions that might use this new function that have not been implemented yet, namely `pixel_dissolve()` and `generate_filter_rect()`.

### Similar function APIs, slightly different behavior

There are two other places and functions in the code where similar pieces of code exist, and where the API for the functions are similar having the `sourceRect:Rectangle` (from [`BitmapData`](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/display/BitmapData.html)) parameter, namely:

`apply_filter()`

https://github.com/ruffle-rs/ruffle/blob/ea78bf2d3983b525b61861eae7e8fbb715ace272/core/src/avm2/globals/flash/display/bitmap_data.rs#L1043-L1052

and

`palette_map()`

https://github.com/ruffle-rs/ruffle/blob/ea78bf2d3983b525b61861eae7e8fbb715ace272/core/src/avm2/globals/flash/display/bitmap_data.rs#L1111-L1125

But they are not quite the same, and I have not looked into them in depth, and therefore have not refactored them.

### DIFF
--- a/core/src/avm2/globals/flash/display/bitmap_data.rs
+++ b/core/src/avm2/globals/flash/display/bitmap_data.rs
@@ -23,6 +23,26 @@ use ruffle_render::filters::Filter;
 use ruffle_render::transform::Transform;
 use std::str::FromStr;
 
+fn get_rectangle_x_y_width_height<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    rectangle: Object<'gc>,
+) -> Result<(i32, i32, i32, i32), Error<'gc>> {
+    let x = rectangle
+        .get_public_property("x", activation)?
+        .coerce_to_i32(activation)?;
+    let y = rectangle
+        .get_public_property("y", activation)?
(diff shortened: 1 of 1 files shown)


iwannabethedev:
After thinking it over, I think I will implement the refactoring for the "Slightly different APIs, same behavior" functions.

iwannabethedev:
I am closing this since there are plenty of other PRs to look at at the moment, and these changes are fairly straight-forward to do later on and can also fairly easily be adapted from this PR later on if desired.

Dinnerbone (approved):

PR #11011, opened 2023-05-11
i18n: sync my change from Crowdin

yangyangdaji:
sync my change from Crowdin
See for details https://crowdin.com/project/ruffle/zh-CN

Signed-off-by: yangyangdaji <1504305527@qq.com>


Dinnerbone:
They will be updated occasionally by a tool, we shouldn't update them by hand. This is to make sure we don't accidentally make any mistakes. I'll make a PR later today with all the crowdin changes since yesterday.

PR #11014, opened 2023-05-11
render: Use `swf::Point<Twips>` in more places

relrelb:
Follow-up to #10918.


core/src/display_object/edit_text.rs
```diff
@@ -660,26 +660,19 @@ impl<'gc> EditText<'gc> {
...
+                .draw_command(DrawCommand::LineTo(Point::new(Twips::ZERO, height)));
+            write
+                .drawing
+                .draw_command(DrawCommand::LineTo(Point::new(width, height)));
+            write
+                .drawing
+                .draw_command(DrawCommand::LineTo(Point::new(width, Twips::ZERO)));
+            write
+                .drawing
+                .draw_command(DrawCommand::LineTo(Point::new(Twips::ZERO, Twips::ZERO)));
```
iwannabethedev, line 675: Nit: Could `Point::new(Twips::ZERO, Twips::ZERO)` be replaced with `Point::ZERO` like above?

relrelb: Nice catch! Done.

PR #11016, opened 2023-05-11
agal: Correctly check vertex_input_expressions cache

Aaron1011:
We need to use the same logic that we use for `get_varying_pointer`, since the AGAL shader can access vertex inputs in any order.


iwannabethedev:
The changes look good to me.

Dinnerbone (approved):

PR #11023, opened 2023-05-12
extension: Fix encoding issues affecting translations

Toad06:
Currently, translations can look broken if the page is not UTF-8 encoded:
bug

### DIFF
--- a/web/packages/core/tools/bundle_texts.js
+++ b/web/packages/core/tools/bundle_texts.js
@@ -37,7 +37,20 @@ locales.forEach((locale) => {
 const options = {
     files: "dist/**",
     from: [/\{\s*\/\*\s*%BUNDLED_TEXTS%\s*\*\/\s*}/g],
-    to: [JSON.stringify(bundled_texts, null, 2)],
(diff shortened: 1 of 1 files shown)


Dinnerbone:
I wonder if we should forcefully escape any non ascii characters in the script to embed the translations, then anyone serving selfhosted as non utf8 will still work too

Toad06:
This might be a good idea - though others would say that all administrators have to do is add `charset="utf-8"` to their `<script>` tag.

Dinnerbone (approved):

web/packages/extension/src/content.ts
```diff
@@ -65,6 +65,7 @@ function injectScriptURL(url: string): Promise<void> {
         script.addEventListener("error", (e) => reject(e));
     });
     script.src = url;
+    script.charset = "utf-8";
```
relrelb, line 68: Actually with the escaping in `bundle_text.js`, I think this change is no longer needed :)

relrelb (approved):

Issue #11007, opened 2023-05-11
Crash on neave_tv.swf


TemaisgameNoobas:
### Describe what you were doing

Just launched the swf and Ruffle crashed.
The SWF could be found at Wayback Machine (by going to neave.tv in 2006)
[crashedswf.zip](https://github.com/ruffle-rs/ruffle/files/11451498/crashedswf.zip)

### What does the crash message say?

panicked at 'index out of bounds: the len is 43200 but the index is 43200', core/src/bitmap/bitmap_data.rs:599:9

### Ruffle Version

0.1.0-nightly (e488cc9 2023-05-10)

### Affected platform

Desktop app

### Operating system

Ubuntu 22.04 (jammy) [64-bit]

### Additional information

Filename: neave_tv.swf

### Render Info
Renderer: wgpu
Adapter Backend: Vulkan
Adapter Name: "NVIDIA GeForce GTX 1660 SUPER"
Adapter Device Type: DiscreteGpu
Adapter Driver Name: "NVIDIA"
Adapter Driver Info: "515.105.01"
Enabled features: Features(TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES | PUSH_CONSTANTS)
[...]

PR #11006, opened 2023-05-11
render/core: Fix of panic bug and rendering bug in #10955 .

iwannabethedev:
With help from and credit to @Dinnerbone and @n0samu , we have a probable fix for #10955 .

Credit for https://github.com/ruffle-rs/ruffle/pull/11006/commits/19b6cf68a11a1111d48d59358d6a6f2e7fe4d1ae goes fully to @Dinnerbone .

I have run the game "Dont escape 3", and the introduction part of it works now, without any panics or rendering issues as far as I can tell.

See also https://github.com/ruffle-rs/ruffle/pull/10583 .

Fixes #10955 .
Fixes https://github.com/ruffle-rs/ruffle/issues/11007 .
Might fix https://github.com/ruffle-rs/ruffle/issues/11017 .

Issue #10955, opened 2023-05-05
Error on https://uploads.ungrounded.net/ruffle_wrapper/ruffleembed.html


mari-does-not-dream-of-mental-stability:
# Error Info
Error name: Error
Error message: panicked at 'index out of bounds: the len is 39032 but the index is 39032', core/src/bitmap/bitmap_data.rs:604:9
Error stack:
```
339/O/n.wbg.__wbg_new_15d3966e9981a196@https://uploads.ungrounded.net/ruffle_wrapper/lib/core.ruffle.7d3ff466cfbd971981de.js:1:48914
core::ops::function::FnOnce::call_once{{vtable.shim}}::hd28f65a61f685e47@:wasm-function[8418]:0x811c6f
std::panicking::rust_panic_with_hook::h7f7102b82d51338f@:wasm-function[5512]:0x733897
core::panicking::panic_fmt::he4489d678d6570d5@:wasm-function[8271]:0x80c4b5
core::panicking::panic_bounds_check::h1213b10c1e002d73@:wasm-function[6633]:0x78d676
ruffle_core::bitmap::operations::copy_on_cpu::ha57d1f647286244e@:wasm-function[779]:0x2aafb6
ruffle_core::bitmap::operations::copy_pixels::hd834677fae0a78bc@:wasm-function[4235]:0x6c51e8
ruffle_core::avm2::globals::flash::display::bitmap_data::copy_pixels::h1320c5f91c02c9a2@:wasm-function[727]:0x28311e
ruffle_core::avm2::function::Executable::exec::hb5981ebd76854dce@:wasm-function[780]:0x2ab7ae
ruffle_core::avm2::object::TObject::call_property::hbef9b7a23152cbbb@:wasm-function[1116]:0x386444
<ruffle_core::avm2::object::Object as ruffle_core::avm2::object::TObject>::call_property::h1ff9616f1a2d6ab7@:wasm-function[2665]:0x5a55a9
ruffle_core::avm2::activation::Activation::run_actions::h8ac32777c1f985d5@:wasm-function[494]:0x2880d
ruffle_core::avm2::function::Executable::exec::hb5981ebd76854dce@:wasm-function[780]:0x2accb4
ruffle_core::avm2::object::TObject::call_property::hfa2ab3f33ae33aa2@:wasm-function[1084]:0x377981
[...]

Issue #10954, opened 2023-05-05
Error on https://game36911.konggames.com/games/Megadev/time4cat/frame/8aad93cd-f776-40b0-abef-ba86ae178ef7/


orion140:
# Error Info
Error name: Error
Error message: panicked at 'index out of bounds: the len is 4800 but the index is 4800', core/src/bitmap/bitmap_data.rs:604:9
Error stack:
```
Error: panicked at 'index out of bounds: the len is 4800 but the index is 4800', core/src/bitmap/bitmap_data.rs:604:9
    at O.n.wbg.__wbg_new_15d3966e9981a196 (https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.7d3ff466cfbd971981de.js:1:48914)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::hd28f65a61f685e47 (wasm://wasm/02b8133e:wasm-function[8418]:0x811c6f)
    at std::panicking::rust_panic_with_hook::h7f7102b82d51338f (wasm://wasm/02b8133e:wasm-function[5512]:0x733897)
    at core::panicking::panic_fmt::he4489d678d6570d5 (wasm://wasm/02b8133e:wasm-function[8271]:0x80c4b5)
    at core::panicking::panic_bounds_check::h1213b10c1e002d73 (wasm://wasm/02b8133e:wasm-function[6633]:0x78d676)
    at ruffle_core::bitmap::operations::copy_on_cpu::ha57d1f647286244e (wasm://wasm/02b8133e:wasm-function[779]:0x2aafb6)
    at ruffle_core::bitmap::operations::draw::hba4e79233cd20803 (wasm://wasm/02b8133e:wasm-function[800]:0x2c4c08)
    at ruffle_core::avm2::globals::flash::display::bitmap_data::draw::hae63a30bb1476f73 (wasm://wasm/02b8133e:wasm-function[807]:0x2cb7ac)
    at ruffle_core::avm2::function::Executable::exec::hb5981ebd76854dce (wasm://wasm/02b8133e:wasm-function[780]:0x2ab7ae)
    at ruffle_core::avm2::object::TObject::call_property::hbef9b7a23152cbbb (wasm://wasm/02b8133e:wasm-function[1116]:0x386444)
```
AVM2 stack:
```
    at flash.display::BitmapData/flash::display::BitmapData::draw()
    at Time4Cat/_tf71()
    at Time4Cat/_xe194()
    at Time4Cat/draw()
[...]

PR #11038, opened 2023-05-13
core: Fix wrong `blend_over()` - but with failing tests.

iwannabethedev:
This change seemingly fixes the new bug related to the SWF in https://github.com/ruffle-rs/ruffle/issues/10954 (the issue in #10954 describes a previous, already fixed bug, the new bug has not been described in any issue, but described and investigated on Discord). This new bug came to surface with the latest nightly (2023-05-13), for the SWF in #10954 , where the bug causes flashing with crazy colors once the game is started from the game's main menu (the bug was expressed and came to surface, but not introduced as far as I can tell, with the changes in https://github.com/ruffle-rs/ruffle/pull/11006 ). This PR is not regressing neither the panic bug nor the visual issue for https://github.com/ruffle-rs/ruffle/issues/10955 and https://github.com/ruffle-rs/ruffle/issues/11007 .

**Superseded by:** https://github.com/ruffle-rs/ruffle/pull/11044 .

~~The apparent cause of the bug is https://github.com/ruffle-rs/ruffle/pull/2488 , made some two years ago. Ruffle was much less mature back then (the code likely changed much more and unit tests were likely less cost-efficient overall development-wise), and it in my opinion makes sense that there were fewer unit tests back then for new code.~~

~~I believe that the primary cause of the bug was a lack of unit tests, which is understandable given that Ruffle was much newer back then. I suspect that a secondary cause of the bug was all the type-casts, which made the code more obscure and difficult to comprehend. I have made the code clearer, at the cost of making it more verbose, which I think is justified in this case.~~

~~For some reason, two tests are failing, so all the automatically run tests will presumably fail. I do not understand or know why they fail, I will need help from you to figure that out.~~

~~There are two `TODO`s in the code, which I would like to request reviewers to also consider.~~

~~Finally, I feel that it would be prudent for me to write some unit tests for `blend_over()`. Or at least create an issue, since it would be nice if this PR was figured out before the next nightly release. Though this fix is not that urgent, I believe, and it might be better to wait for the next nightly instead of hurrying.~~

### DIFF
--- a/core/src/bitmap/bitmap_data.rs
+++ b/core/src/bitmap/bitmap_data.rs
@@ -128,12 +128,24 @@ impl Color {
 
     #[must_use]
     pub fn blend_over(&self, source: &Self) -> Self {
-        let sa = source.alpha();
+        let tr = self.red() as u16;
+        let tg = self.green() as u16;
+        let tb = self.blue() as u16;
+        let ta = self.alpha() as u16;
 
-        let r = source.red() + ((self.red() as u16 * (255 - sa as u16)) >> 8) as u8;
-        let g = source.green() + ((self.green() as u16 * (255 - sa as u16)) >> 8) as u8;
-        let b = source.blue() + ((self.blue() as u16 * (255 - sa as u16)) >> 8) as u8;
-        let a = source.alpha() + ((self.alpha() as u16 * (255 - sa as u16)) >> 8) as u8;
+        let sr = source.red() as u16;
+        let sg = source.green() as u16;
+        let sb = source.blue() as u16;
+
+        let sa = source.alpha() as u16;
+        let neg_sa = 255 - source.alpha() as u16;
+
+        // TODO: Best to divide with 256/>>8? Or divide with 255?
+        let r = ((sr * sa + tr * neg_sa) >> 8) as u8;
+        let g = ((sg * sa + tg * neg_sa) >> 8) as u8;
+        let b = ((sb * sa + tb * neg_sa) >> 8) as u8;
+        // TODO: What should be done with alpha here?
+        let a = ((sa * sa + ta * neg_sa) >> 8) as u8;
         Self::argb(a, r, g, b)
     }
 }


iwannabethedev:
CC @n0samu @adrian17 . (Also informing  @Dinnerbone ).

Dinnerbone:
Can you please split this into "refactor" + "change" commits? It's hard to tell what changed when you refactor at the same time

What exactly is the issue you fixed? Were there values that the blend was wrong for? If so can you add those to the test?

It's likely if some values were right and some wrong, and now a different set is right and wrong, then the method used to calculate the values is wrong. Floating point math may be more correct, or something akin to the unpremultiply LUT, but we have to check with tests.

iwannabethedev:
As far as I can tell, the previous function implementation is obviously wrong, since it can result in values that are considerably larger than 255 despite being casted to `u8`. For instance:

```rust
source.red() + ((self.red() as u16 * (255 - sa as u16)) >> 8) as u8;
```

If `source.red()` can have the value `255`, `self.red()` can have the value `255` and `sa` can have the value `0` or `1`, then the result will be obviously larger than `255`. Even though the result is casted to `u8`. This is consistent with the game `Time4Cat` having crazy colors - the colors likely overflowed in strange ways.

What I did is what I in my experience is usually done with blending, namely where you take some proportion of one color, and take `1 - proportion` from the other color.

Dinnerbone:
> As far as I can tell, the previous function implementation is obviously wrong, since it can result in values that are considerably larger than 255 despite being casted to `u8`. For instance:

Yep, good catch!

I'd still like to amend the test with new values, to prove this and guard against regressions.

We do all blending in premultiplied alpha, so the formula in normalized 0..1 values would be `target = src + ((1 - src_alpha) * dst)`. If values aren't accurate to flash then we'll need to experiment if they did float-based, bitmath, or a LUT

iwannabethedev:
I do not feel that it was a good catch, since it seems like an obvious error, that either should have been caught in review or with unit testing. But, it was early in Ruffle's development, and you cannot expect to unit test or carefully code review everything when the code is presumably changing a lot - development-wise, it is less likely to be cost-efficient (though it of course depends on the kind of technology and quality level you are seeking and the goals - at places like NASA, there are of course other priorities).

The code currently fails, and I do not understand why. But, while the new fix visually works with the two SWFs I tested, I have not tested them specifically, so you are right that the new implementation, while not being obviously wrong, may not be accurate to the behavior of the Flash Player, and I agree (if I understand you correctly) that `blend_over()` really should be fully accurate in regards to what the Flash Player is doing with the Flash Player's version of `blend_over()`. So my current PR, if the tests are still accurate (and I think they are, because I have now run the AVM2 test script and I do indeed get the values that are in the test folder in output.txt), unless there is an issue or a bug other than in `blend_over()` elsewhere, then my new changes here are buggy (there might also be the possibility of both my changed implementation of `blend_over()` and elsewhere being buggy. It would be nice if we could test the corresponding function to `blend_over()` in the official Flash Player exactly, but I assume that is not possible).

I do not quite understand what you mean by premultiplied alpha, but I think I will have to ponder it a bit before I get it.

iwannabethedev:
Attaching the test result output when running with `cargo test copypix` in 'tests/tests':

```
running 2 tests
[WARN  wgpu_hal::vulkan::instance] Unable to find layer: VK_LAYER_KHRONOS_validation
[INFO  naga::back::spv::writer] Skip function Some("find_t")
[INFO  naga::back::spv::writer] Skip function Some("find_t")
[INFO  naga::back::spv::writer] Skip function Some("find_t")
[INFO  naga::back::spv::writer] Skip function Some("find_t")
[INFO  naga::back::spv::writer] Skip function Some("find_t")
[INFO  naga::back::spv::writer] Skip function Some("_naga_oil_mod_MZUWY5DFOI_membermain_vertex")
[INFO  naga::back::spv::writer] Skip function Some("_naga_oil_mod_MZUWY5DFOI_membermain_vertex")
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`: ruffle output != flash player output

Diff < left / right > :
<40142434 40142434 45122134 45122134 45122134 
<40142434 40142434 45122134 45122134 45122134 
>5d243147 5d243147 45122134 45122134 45122134 
>5d243147 5d243147 45122134 45122134 45122134 
 45122134 45122134 45122134 45122134 45122134 
 45122134 45122134 45122134 45122134 45122134 
 45122134 45122134 45122134 45122134 -7f564435 
 transparentSource: -ff1f1f2
 Non-transparent testing
 Original pixel: -1000000
<transparent source mergeAlpha=false -f3f3f4
<transparent source mergeAlpha=true -f3f3f4
>transparent source mergeAlpha=false -f2f2f3
>transparent source mergeAlpha=true -f2f2f3
 nontransparent source mergeAlpha=false -fdfbfa
 nontransparent source mergeAlpha=true -fdfbfa
 
 Transparent testing
 Original pixel: -1
 nontransparent source mergeAlpha=false -fdfbfa
 nontransparent source mergeAlpha=true -fdfbfa
 

', tests/tests/util/test.rs:134:13
test avm1/bitmap_data_copypixels ... FAILED
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`: ruffle output != flash player output

Diff < left / right > :
<40142434 40142434 45122134 45122134 45122134 
<40142434 40142434 45122134 45122134 45122134 
>5d243147 5d243147 45122134 45122134 45122134 
>5d243147 5d243147 45122134 45122134 45122134 
 45122134 45122134 45122134 45122134 45122134 
 45122134 45122134 45122134 45122134 45122134 
 45122134 45122134 45122134 45122134 80a9bbcb 
 transparentSource: f00e0e0e
 Non-transparent testing
 Original pixel: ff000000
<transparent source mergeAlpha=false ff0c0c0c
<transparent source mergeAlpha=true ff0c0c0c
>transparent source mergeAlpha=false ff0d0d0d
>transparent source mergeAlpha=true ff0d0d0d
 nontransparent source mergeAlpha=false ff020406
 nontransparent source mergeAlpha=true ff020406
 
 Transparent testing
 Original pixel: ffffffff
 nontransparent source mergeAlpha=false ff020406
 nontransparent source mergeAlpha=true ff020406
 Copy with alphaPoint=undefined:
<fe7f0000 fe7f0000 0 0 0 
<fe7f0000 fe7f0000 0 0 0 
>ff800000 ff800000 0 0 0 
>ff800000 ff800000 0 0 0 
 0 0 0 0 0 
 0 0 0 0 0 
 0 0 0 0 0 
 

', tests/tests/util/test.rs:134:13
test avm2/bitmapdata_copypixels  ... FAILED

failures:

---- avm1/bitmap_data_copypixels ----
test panicked: assertion failed: `(left == right)`: ruffle output != flash player output

Diff < left / right > :
<40142434 40142434 45122134 45122134 45122134 
<40142434 40142434 45122134 45122134 45122134 
>5d243147 5d243147 45122134 45122134 45122134 
>5d243147 5d243147 45122134 45122134 45122134 
 45122134 45122134 45122134 45122134 45122134 
 45122134 45122134 45122134 45122134 45122134 
 45122134 45122134 45122134 45122134 -7f564435 
 transparentSource: -ff1f1f2
 Non-transparent testing
 Original pixel: -1000000
<transparent source mergeAlpha=false -f3f3f4
<transparent source mergeAlpha=true -f3f3f4
>transparent source mergeAlpha=false -f2f2f3
>transparent source mergeAlpha=true -f2f2f3
 nontransparent source mergeAlpha=false -fdfbfa
 nontransparent source mergeAlpha=true -fdfbfa
 
 Transparent testing
 Original pixel: -1
 nontransparent source mergeAlpha=false -fdfbfa
 nontransparent source mergeAlpha=true -fdfbfa
 



---- avm2/bitmapdata_copypixels ----
test panicked: assertion failed: `(left == right)`: ruffle output != flash player output

Diff < left / right > :
<40142434 40142434 45122134 45122134 45122134 
<40142434 40142434 45122134 45122134 45122134 
>5d243147 5d243147 45122134 45122134 45122134 
>5d243147 5d243147 45122134 45122134 45122134 
 45122134 45122134 45122134 45122134 45122134 
 45122134 45122134 45122134 45122134 45122134 
 45122134 45122134 45122134 45122134 80a9bbcb 
 transparentSource: f00e0e0e
 Non-transparent testing
 Original pixel: ff000000
<transparent source mergeAlpha=false ff0c0c0c
<transparent source mergeAlpha=true ff0c0c0c
>transparent source mergeAlpha=false ff0d0d0d
>transparent source mergeAlpha=true ff0d0d0d
 nontransparent source mergeAlpha=false ff020406
 nontransparent source mergeAlpha=true ff020406
 
 Transparent testing
 Original pixel: ffffffff
 nontransparent source mergeAlpha=false ff020406
 nontransparent source mergeAlpha=true ff020406
 Copy with alphaPoint=undefined:
<fe7f0000 fe7f0000 0 0 0 
<fe7f0000 fe7f0000 0 0 0 
>ff800000 ff800000 0 0 0 
>ff800000 ff800000 0 0 0 
 0 0 0 0 0 
 0 0 0 0 0 
 0 0 0 0 0 
 




failures:
    avm1/bitmap_data_copypixels
    avm2/bitmapdata_copypixels

test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 4 filtered out; finished in 1.17s

error: test failed, to rerun pass `--test tests`

```

iwannabethedev:
> We do all blending in premultiplied alpha,

I do not understand this part. It would explain some cases where `to_premultiplied_alpha()` is called before-hand (though I feel like renaming and also documenting that `blend_over()` expects premultiplied alpha values would be helpful, like renaming it `blend_over_premultiplied_alpha()`). But, while I have only skimmed the code, there appears to be other places where `blend_over()` is called without `to_premultiplied_alpha()` being called before-hand. But I will have to take a closer look at the code to be certain I understand what is going on.

iwannabethedev:
Like, in this code piece, it is clear that `to_premultiplied_alpha()` is called before `blend_over()`.

https://github.com/ruffle-rs/ruffle/blob/master/core/src/bitmap/operations.rs#L1329-L1333

But in this code piece below, I am not certain whether `to_premultiplied_alpha()` is ever called before `blend_over()`.

https://github.com/ruffle-rs/ruffle/blob/master/core/src/bitmap/operations.rs#L1342-L1355

As well as here:

https://github.com/ruffle-rs/ruffle/blob/master/core/src/bitmap/operations.rs#L1235-L1245

https://github.com/ruffle-rs/ruffle/blob/master/core/src/bitmap/operations.rs#L1286-L1298

But I am not certain about those places, I have not looked into them carefully.

But; does this mean that the real fix is not this PR, but to carefully ensure that wherever `blend_over()` is called, ` `to_premultiplied_alpha()` has been applied to the values some time before when relevant? If so, I think it would be a really good idea to add documentation to `blend_over()`, and probably also rename that method, to clearly signal that pre-multiplication is required, since it is wrong for the range of the values of its type as it before this PR is implemented.

iwannabethedev:
@Dinnerbone I think that I will make a new PR, and in that PR, I will not touch the implementation of `blend_over()`, since if it expects pre-multiplied values, then my changes here are wrong and buggy. Instead, I will see if I can introduce `to_premultiplied_alpha()` in appropriate places where they appear to be missing, and then run the different SWFs that there have been bugs with - and, to help avoid issues in the future, then document and possibly rename `blend_over()`, since it feels too error-prone when (as far as I can see) multiple people misunderstand that it expects premultiplied values.

In either case, thank you very much for the help and guidance.

iwannabethedev:
Superseded by https://github.com/ruffle-rs/ruffle/pull/11044 .

PR #11040, opened 2023-05-13
extension: Don't say that Ruffle cannot load on "player.html"

Toad06:
In the current builds, when you open a SWF directly (say [Ultimate Flash Sonic](https://uploads.ungrounded.net/151000/151706_ultimate_sonic.swf)) and click the Ruffle icon in your browser, it says that "Ruffle cannot load on protected browser pages.". This PR fixes that minor issue.

A new string is also introduced to that end (let me know if the wording can be improved). I also wonder if we should change the string that mentions the "protected browser pages", since this notion may not be obvious for some users.

### DIFF
--- a/web/packages/extension/src/popup.ts
+++ b/web/packages/extension/src/popup.ts
@@ -14,14 +14,15 @@ let reloadButton: HTMLButtonElement;
 // prettier-ignore
 const STATUS_COLORS = {
     "status_init": "gray",
-    "status_no_tabs": "red",
-    "status_tabs_error": "red",
     "status_message_init": "gray",
-    "status_result_protected": "gray",
-    "status_result_error": "red",
-    "status_result_running": "green",
-    "status_result_optout": "gray",
+    "status_no_tabs": "red",
     "status_result_disabled": "gray",
+    "status_result_error": "red",
+    "status_result_optout": "gray",
+    "status_result_protected": "gray",
+    "status_result_running": "green",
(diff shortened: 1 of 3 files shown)


n0samu:
This has been bugging me for some time, so I'm happy you found a solution! But is it worth adding the extra permission? How does it change the permissions dialog that appears when installing the extension?

Toad06:
The "tabs" permission is unfortunately required for Chrome, otherwise `activeTab.url` is `undefined`.

Currently, when you click to install Ruffle from the Chrome Web Store, a dialog tells you that:
> Read and change all your data on the websites you visit

So I don't think Chrome will display a message scarier than that... 😅 but can't say for sure.

web/packages/extension/src/popup.ts
```diff
@@ -60,8 +61,17 @@ async function queryTabStatus(
             type: "ping",
         });
     } catch (e) {
-        listener("status_result_protected");
-        reloadButton.disabled = true;
+        const url = activeTab.url ? new URL(activeTab.url) : null;
```
relrelb, line 64: Nice improvement, thank you!
Can you please place the added logic outside the `catch`? My intuition here is that it doesn't have to do with any error-handling at all. I think it better fits right after `activeTab` is assigned.

Toad06: This definitely makes sense. I did the changes.

relrelb (approved):

PR #11044, opened 2023-05-14
core: Fix wrong color form used.

iwannabethedev:
Supersedes #11038 .

For the description of the bug, see #11038 .

The fix in #11038 was wrong and came about due to a lack of understanding of the codebase, however, it still identified the correct part of the code that was related to the bug. This PR fixes in the first commit the bug, and in the second commit adds documentation, including to a function that may otherwise appear as if it can be used with any kind of `Color` value, even though it cannot (the code also incorrectly passed in the wrong form of the `Color` value, which was the cause of the bug, and added to my confusion).

~~Fixes https://github.com/ruffle-rs/ruffle/issues/10954 .~~ Edit: Fixes the new flashing crazy colors bugs, in the SWF found in #10954 .
Has been tested against https://github.com/ruffle-rs/ruffle/issues/10955 and https://github.com/ruffle-rs/ruffle/issues/11007 .

Tasks and questions for the code reviewers:
- Is the documentation accurate and good?
- Would it in your estimation (for instance in an issue or future PR) be worth it to encode in the type system that there are two different forms of colors? One argument against is that it would complicate the codebase and take time to refactor and implement. One argument for is that it might prevent hours-long debugging sessions involving multiple people figuring out why there are strange rendering bugs (though this latter argument depends on how much more development there might be related to this code and how much the `Color` type is used).
- **New:** Is it fine in the test to have a `tolerance` of `1` in the `test.toml`?
- **New:** Should the test also be added to the AVM1 tests?

Credit to @Dinnerbone , @n0samu , @adrian17 and others for help and guidance.


Lord-McSweeney:
Could you add a test please?

iwannabethedev:
> Could you add a test please?

I am trying to compile the SWF used to run tests for the AVM2 version of `BitmapData.draw()` in 'tests/tests/swfs/avm2/bitmapdata_draw/' but I get the following error:

type error flash player

iwannabethedev:
> Could you add a test please?

A test has been added.

adrian17 (approved):

Issue #10463, opened 2023-03-30
Cannot load external images with AS3 Loader


n0samu:
### Describe the bug

Consider this sample: [AS3ImageLoader.zip](https://github.com/ruffle-rs/ruffle/files/11110181/AS3ImageLoader.zip)
```as
var loader:Loader = new Loader();
var url:URLRequest = new URLRequest("test.png");
loader.load(url);
addChild(loader);
```
Ruffle requests and reads the image file, but never displays it. Loading SWF files works fine, but loading image files does not.

### Expected behavior

The image should be displayed in the top-left corner of the player.
![image](https://user-images.githubusercontent.com/71368227/228800123-25b2bcef-f9a1-4a23-b55d-31ef45e70db7.png)

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

This problem was first mentioned by @Lord-McSweeney in #10440.

Lord-McSweeney:
Fixed by #11045.

Issue #9977, opened 2023-03-10
AstroFlyer: Cannot start game


n0samu:
### Describe the bug

Link to SWF: https://img.y8.com/cloud/y8-flash-game/contents/item_versions/flash_games/14068/original/astroflyer.swf

In [AstroFlyer](https://www.y8.com/games/astroflyer), clicking the Start New Game button doesn't work, so the game is unplayable.

### Expected behavior

Clicking Start New Game should start the game.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

This error seems related:
`ERROR run_frame:run_all_phases_avm2: ruffle_core::avm2::activation: TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: addEventListener)`

n0samu:
Here is the stack trace:
```
        at ASTROFLYER14_fla::mainMenu_7/frame12()
        at flash.display::MovieClip/flash::display::MovieClip::gotoAndStop()
        at ASTROFLYER14_fla::MainTimeline/nextWall()
        at ASTROFLYER14_fla::MainTimeline/getTimeDiff()
```
Here is the frame12 function:
```as
        internal function frame12() : *
        {
            stop();
            startGameButton.addEventListener(MouseEvent.CLICK,startIt);
        }
```
For some reason Ruffle does not link the startGameButton variable to the button object placed on frame 12:
![image](https://user-images.githubusercontent.com/71368227/227751073-d2117117-cef0-4fed-8854-0c9b0ae4f7bf.png)

n0samu:
The game actually works fine sometimes - this must be a streaming loading issue, since it seems timing-dependent.

n0samu:
I'm not sure why, but after testing this many times, it seems pretty certain that this is fixed now.

PR #11045, opened 2023-05-14
core: Refactor Loader to properly load images

Aaron1011:
We now run all of the completion logic (including adding the new DisplayObject as a child) in `Loader::movie_loader_complete`. Previously, some of this logic was run from `Loader::preload_tick`, which meant that loaded images did not have the logic run.

Also, `BitmapData` and `Bitmap` instances (with corresponding AVM2 objects) are now properly constructed for loaded images.


danielhjacobs:
Not at a computer to test this, does it fix number 3 on #10173?

Lord-McSweeney:
Fixes #10463.

Toad06:
This looks good to me based on testing.

I also wonder if this could be a good opportunity to turn the following AVM1 tests into visual tests (compare the image rendered with an expected image):
[avm1/movieclip_methods_with_loaded_image](https://github.com/ruffle-rs/ruffle/tree/master/tests/tests/swfs/avm1/movieclip_methods_with_loaded_image)
[avm1/mcl_giftarget](https://github.com/ruffle-rs/ruffle/tree/master/tests/tests/swfs/avm1/mcl_giftarget)
[avm1/mcl_jpgtarget](https://github.com/ruffle-rs/ruffle/tree/master/tests/tests/swfs/avm1/mcl_jpgtarget)
[avm1/mcl_pngtarget](https://github.com/ruffle-rs/ruffle/tree/master/tests/tests/swfs/avm1/mcl_pngtarget)

Toad06:
> Not at a computer to test this, does it fix number 3 on #10173?

It does. 👍

core/src/loader.rs
```diff
@@ -1604,38 +1609,116 @@ impl<'gc> Loader<'gc> {
...
+            if let Some(MovieLoaderEventHandler::Avm2LoaderInfo(loader_info)) = event_handler {
+                Some(*loader_info.as_loader_info_object().unwrap())
+            } else {
+                None
+            };
+
+        if let Some(loader_info) = loader_info {
+            // Store the real movie into the `LoaderStream`, so that
+            // 'bytesTotal' starts returning the correct value
+            // (we previously had a fake empty SwfMovie).
```
iwannabethedev, line 1638: Is this comment still relevant?

core/src/loader.rs
```diff
@@ -1604,38 +1609,116 @@ impl<'gc> Loader<'gc> {
...
+            // However, we still use `LoaderStream::NotYetLoaded`, since
+            // the actual MovieClip display object has not run its first
+            // frame yet.
+            loader_info.set_loader_stream(
+                LoaderStream::NotYetLoaded(movie.clone().unwrap(), Some(dobj.unwrap()), false),
```
iwannabethedev, line 1643: Nit: I am a bit confused by this expression: `Some(dobj.unwrap())` . `dobj` is an `Option` I believe, so I am wondering whether it could be just `dobj`. Is `unwrap()` called on it to ensure that it is present and otherwise panic?

core/src/loader.rs
```diff
@@ -1604,38 +1609,116 @@ impl<'gc> Loader<'gc> {
...
+            // clip does not yet exist. Any children added inside the movie
+            // frame constructor will see an 'added' event immediately, and
+            // an 'addedToStage' event *after* the constructor finishes
+            // when we add the movie as a child of the loader.
+            loader.insert_at_index(&mut activation.context, dobj.unwrap(), 0);
+        } else if let Some(dobj) = dobj {
+            // This is a load of an image into AVM1 - add it as a child of the target clip.
+            if dobj.as_movie_clip().is_none() {
+                let mut mc = target_clip.as_movie_clip().unwrap();
+                mc.replace_with_movie(uc, Some(movie.unwrap()), None);
```
Lord-McSweeney, line 1692: Same here

core/src/loader.rs
```diff
@@ -1604,38 +1609,116 @@ impl<'gc> Loader<'gc> {
...
+            loader_info.set_loader_stream(
+                LoaderStream::NotYetLoaded(movie.clone().unwrap(), Some(dobj.unwrap()), false),
+                uc.gc_context,
+            );
+        }
+
+        if let Some(dobj) = dobj {
+            if dobj.as_movie_clip().is_some() {
+                // We call these methods after we initialize the `LoaderInfo`, but before the
+                // add the the loaded clip as a child. The frame constructor should see
```
iwannabethedev, line 1651: (Nit: It would be nice if the wording in these comments from the previous version were fixed, or the comments are removed).

Nit: Are the comments still accurate?

core/src/loader.rs
```diff
@@ -1604,38 +1609,116 @@ impl<'gc> Loader<'gc> {
...
+                .unwrap();
+
+            // Note that we do *not* use the 'addChild' method here:
+            // Per the flash docs, our implementation always throws
+            // an 'unsupported' error. Also, the AVM2 side of our movie
+            // clip does not yet exist. Any children added inside the movie
+            // frame constructor will see an 'added' event immediately, and
+            // an 'addedToStage' event *after* the constructor finishes
+            // when we add the movie as a child of the loader.
+            loader.insert_at_index(&mut activation.context, dobj.unwrap(), 0);
```
iwannabethedev, line 1687: Is it the case that, whenever we have `Some(MovieLoaderEventHandler::Avm2LoaderInfo(loader_info))` and the other checks in this method, that we either should also have a `dobj` that is `Some`, or else we should panic? I am asking due to the `unwrap()` here, and that `dobj` seems like it might be `None` in other branches without panic'ing. It might well be the case that the changes here are good and relevant, I am asking due to me not understanding this part of the codebase.

core/src/loader.rs
```diff
@@ -1512,26 +1480,63 @@ impl<'gc> Loader<'gc> {
...
+
+                    // This will construct AVM2-side objects even under AVM1, but it doesn't matter,
+                    // since Bitmap and BitmapData never have AVM1-side objects.
                     let bitmap = ruffle_render::utils::decode_define_bits_jpeg(data, None)?;
-                    let bitmap_obj = Bitmap::new(uc, 0, bitmap)?;
 
-                    if let Some(mut mc) = clip.as_movie_clip() {
-                        let mut activation = Avm2Activation::from_nothing(uc.reborrow());
-                        mc.replace_with_movie(&mut activation.context, Some(movie), None);
-                        mc.replace_at_depth(uc, bitmap_obj.into(), 1);
```
iwannabethedev, line 1526: I do not understand or know this part of the codebase, but, further down in the changes, `replace_at_depth()` is apparently called with AVM1. Should it also be called with AVM2 somewhere?

iwannabethedev:
I have superficially reviewed the changes, though I do not understand this part of the codebase. The changes superficially look good to me, but I think there should be reviews done by others apart from my superficial review.

core/src/loader.rs
Aaron1011: Yes

core/src/loader.rs
Aaron1011: That's right - it should always be present here, so we should panic if it's not (instead of silently continuing).

core/src/loader.rs
Aaron1011: Yes, this comment is still accurate.

core/src/loader.rs
Aaron1011: Yes, this will always be present here.

core/src/loader.rs
Aaron1011: No - under AVM2, we manually add it as a child of the `Loader` instance.

core/src/loader.rs
iwannabethedev: (Nit: Even though this was written before your PR, I would really like it if the wording in this part of the comment: `but before the add the the loaded` was changed).

core/src/loader.rs
n0samu: This actually causes a panic in AstroFlyer (#9977) while the game is loading. The error is:
`thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', core\src\loader.rs:1643:73`

core/src/loader.rs
Aaron1011: What other comment is this referring to?

iwannabethedev: I believe that it is referring to `Some(movie.unwrap())`, similar as for the above comment about `Some(dobj.unwrap())`.

Aaron1011: Yes, this also asserts that the movie is actualy present.

core/src/loader.rs
Aaron1011: @n0samu I've updated the PR - AstroFlyer is working again.

Dinnerbone (approved):

PR #11046, opened 2023-05-14
avm2: Initial implementation of URLStream

Aaron1011:
This is based on URLLoader, and doesn't actualy 'stream' data - it all becomes available once the request finishes. However, this is good enough to get Sniper Team working.


core/src/avm2/globals/flash/net/URLStream.as
```diff
@@ -0,0 +1,115 @@
...
+        public function get bytesAvailable():uint {
+            return this._loader.data.bytesAvailable;
+        }
+
+        public function get endian():String {
+            return _endian;
+        }
+
+        public function set endian(value:String):void {
+            _endian = value;
```
Lord-McSweeney, line 49: See https://github.com/ruffle-rs/ruffle/pull/10879 for the validation:
```
        public function set endian(value:String):void {
            if (value === Endian.BIG_ENDIAN || value === Endian.LITTLE_ENDIAN) {
                this._endian = value;
                if (this._loader.data) {
                    this._loader.data.endian = value;
                }
            } else {
                throw new ArgumentError("Error #2008: Parameter endian must be one of the accepted values.", 2008);
            }
        }
```

core/src/avm2/globals/flash/net/URLStream.as
```diff
@@ -0,0 +1,115 @@
...
+    import flash.events.SecurityErrorEvent;
+    import flash.events.ProgressEvent;
+    import flash.net.URLLoaderDataFormat;
+    import flash.events;
+    import __ruffle__.stub_method;
+    import __ruffle__.stub_getter;
+    import __ruffle__.stub_setter;
+
+    public class URLStream extends EventDispatcher implements IDataInput {
+        private var _endian:String = Endian.LITTLE_ENDIAN;
```
Lord-McSweeney, line 17: The docs say the default endian should be `BIG_ENDIAN`.

Lord-McSweeney:
The `adobe.ac.pmd.rules.event.ConstructorDispatchingEvent` needs to be disabled here since it's assuming that the `dispatchEvent` logic is running inside the constructor, while it is instead running inside an event listener inside the constructor.

core/src/avm2/globals/flash/net/URLStream.as
```diff
@@ -0,0 +1,118 @@
...
+        private var _endian:String = Endian.BIG_ENDIAN;
+        private var _loader:URLLoader = new URLLoader();
+
+        public function URLStream() {
+            this._loader.dataFormat = URLLoaderDataFormat.BINARY;
+            var self = this;
+
+            this._loader.addEventListener(Event.COMPLETE, function(e:*):void {
+                self._loader.data.endian = self._endian;
+                self.dispatchEvent(new Event(Event.COMPLETE)); // NOPMD ConstructorDispatchingEvent
```
Dinnerbone, line 26: With #11123 landing, we're removing all NOPMD comments now

core/src/avm2/globals/flash/net/URLStream.as
```diff
@@ -0,0 +1,118 @@
...
+    import __ruffle__.stub_method;
+    import __ruffle__.stub_getter;
+    import __ruffle__.stub_setter;
+
+    public class URLStream extends EventDispatcher implements IDataInput {
+        private var _endian:String = Endian.BIG_ENDIAN;
+        private var _loader:URLLoader = new URLLoader();
+
+        public function URLStream() {
+            this._loader.dataFormat = URLLoaderDataFormat.BINARY;
```
Dinnerbone, line 21: Can you drop a comment in the constructor to explain that this is temporary implementation and not what we expect this to do in the future? Even just a `// HACK, see https://github.com/ruffle-rs/ruffle/pull/11046` would be sufficient

Dinnerbone (changes requested):

core/src/avm2/globals/flash/net/URLStream.as
Aaron1011: I added a comment to the `URLLoader` field.

Dinnerbone (approved):

Aaron1011:
@Dinnerbone I added a test, and made URLLoader fire a "progress" event.

PR #11047, opened 2023-05-14
agal: Add support for indirect register loads

Aaron1011:
These are poorly documented, but from looking at OpenFL and AGALMiniAssembler, they allow performing loads of the form `vc[va0.x + offset]` - that is, computing a dynamic register number, instead of using the register number present in the opcode.


render/naga-agal/src/builder.rs
```diff
@@ -808,43 +808,115 @@ impl<'a> NagaBuilder<'a> {
...
-                self.emit_const_register_load(source.reg_num as usize)?,
-                // Constants are always a vec4<f32>
-                VertexAttributeFormat::Float4,
-            ),
-            RegisterType::Temporary => {
-                let temp = self.get_temporary_register(source.reg_num as usize)?;
-                (
-                    self.evaluate_expr(Expression::Load { pointer: temp }),
+        let mut load_register = |register_type: &RegisterType, reg_num| {
+            match register_type {
```
iwannabethedev, line 812: (Very minor nit: Could move `reg_num` into its own `let reg_num = reg_num as usize;` before the match-branches, though this will likely be cleaner to do in a future PR).

render/naga-agal/src/builder.rs
```diff
@@ -808,43 +808,115 @@ impl<'a> NagaBuilder<'a> {
...
+                // Indirect loads allow loading from a dynamically computed register index.
+                // This dynamic index is computed as 'regN.X + offset', where 'regN' is a normal
+                // register (e.g. 'va0'), and 'X' is a component ('X, 'Y', Z', or 'W').
+                // Currently, we only support this when the 'outer' (non-index) register is
+                // a constant register, since we always access constant registers through
+                // an array access.
+                match source.register_type {
+                    RegisterType::Constant => {
+                        // Load the index register (e.g. 'va0') as normal, and access the component
+                        // given by 'index_select'. This is 'va0.x' in the above example.
```
iwannabethedev, line 856: (Very minor nit:

Replace:

```
// given by 'index_select'. This is 'va0.x' in the above example.
```

with

```
// given by 'index_select' (e.g. 'x'). This is 'va0.x' in the above example.
```

?)

render/naga-agal/src/builder.rs
```diff
@@ -808,43 +808,115 @@ impl<'a> NagaBuilder<'a> {
...
+
+                        let register_pointer = self.func.expressions.append(
+                            Expression::Access {
+                                base: self.constant_registers,
+                                index: index_with_offset,
+                            },
+                            Span::UNDEFINED,
+                        );
+
+                        // Perform the actual load, giving us 'vc[va0.x + offset]' in the above example.
```
iwannabethedev, line 902: (Is it correctly understood that `vc` is not actually read, and that the final value `register_pointer` already points to the correct place in "memory"?)

iwannabethedev:
I do not understand this part of Ruffle or AGAL or NAGA, and I have not investigated the code in depth, but somewhat superficially, the code looks good to me. Someone else should review it as well.

Is it viable and worth it to add a test case for this?

Dinnerbone:
Can you find a way to test this please? Or some content that uses it?

render/naga-agal/src/builder.rs
Aaron1011: I've changed the closure to take `reg_num` as a `usize`.

render/naga-agal/src/builder.rs
Aaron1011: Good idea - I've changed it

render/naga-agal/src/builder.rs
Aaron1011: We construct a pointer into the `vc` array, and load from it here. We don't load the entire `vc` array (that is, we don't make a copy of it).

Aaron1011:
@Dinnerbone I've added a test to `naga-agal` of some bytecode that uses an indirect register load.

Dinnerbone (approved):

PR #11058, opened 2023-05-15
web: Only minify ascii characters on selfhosted builds

Toad06:
This is a follow-up to #11023: I haven't paid attention earlier but the changes we made in this PR didn't actually affect self-hosted builds because Webpack default minification/obfuscation somehow reverts them.

### DIFF
--- a/web/packages/selfhosted/webpack.config.js
+++ b/web/packages/selfhosted/webpack.config.js
@@ -3,6 +3,7 @@
 const path = require("path");
 const json5 = require("json5");
 const CopyPlugin = require("copy-webpack-plugin");
+const TerserPlugin = require("terser-webpack-plugin");
 
 function transformPackage(content) {
     const package = json5.parse(content);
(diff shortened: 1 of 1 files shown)


relrelb:
Good finding! Sounds like Terser can handle escaping by itself, which is pretty neat. If so, I wonder if #11023 should be now reverted, since Terser would anyway escape the code for us.

Also, aren't similar changes needed for the extension as well? Or I'm missing something?

Toad06:
The code is not minified in the extension because of #8791, so Terser wouldn't run (I actually tried 😄). So I don't think the extension and the self-hosted builds can handle this behavior the same way.

It should be possible for #11023 to be reverted if we set `script.charset = "utf-8"` in `extension/src/content.ts`. But I think this could have consequences, for instance if a user wants to compile the self-hosted build without the minification (he then may face encoding issues).

relrelb (approved):

PR #11062, opened 2023-05-15
render: Stroke paths with tiny-skia-path

### LINKED ISSUE #7565: render: Self-overlapping semi-transparent strokes are rendered incorrectly
One great example of this is https://z0r.de/2245, where the changing colored overlay is made up of a really thickly stroked scribble, and it should appear as uniform in opacity and color. It does not.

Software rasterizers might have an easy time to skip any pixel when drawing the stroke that they already touched in this operation.
Otherwise, the usual solution is to turn the stroke outline into a fill path, then fill it with a non-zero fill rule.
This path conversion is not (yet?) provided by `lyon`.

It could also be possible to put the triangles outputted by the stroke tessellator through the fill tessellator again - as a path made up of distinct 3 sided polygons. This way any overlaps will disappear. However, this is likely to turn out to be massively inefficient.

It might also be possible to rasterize the overlapping triangles with some sort of depth or stencil buffer trick to discard the repeated paints at the overlaps.

### LINKED ISSUE #13656: Miter join should clip stroke when < miterlimit, uses bevel instead
### Describe the bug

Steps to reproduce:
- Create SWF file with a new stroke
- Set the stroke join style to miter
- Set enough line width for the effect to be seen
- Draw line segments that have too acute angle between them that the miter length is larger than miter limit

What I see: The stroke is converted to bevel and is not extended to the limit value.

Example SWF file + also Flash CS6 source code.
[miter_join.zip](https://github.com/ruffle-rs/ruffle/files/13061672/miter_join.zip)

Screenshot of the problem:
image
(I drew a black line on place where it should be clipped)

### Expected behavior

The stroke should be clipped at the miter limit value.

Flash uses behaviour that is called miter-clip in SVG 2 specs
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-linejoin#miter-clip

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

_No response_

### Additional information

_No response_

### LINKED ISSUE #13658: Square cap not rendered properly (only one side)
### Describe the bug

Steps to reproduce:
- Create SWF file with new red stroke that has enough width to see it
- Set stroke cap style to square
- Make an U shape
- Draw new 1 px black line on the top of the U

What I see in Ruffle: The red U left side does not reach the black line. It seems that the square cap is missing on one side.

SWF and its Flash CS6 source code:
[square_cap.zip](https://github.com/ruffle-rs/ruffle/files/13061795/square_cap.zip)

Screenshot:
image

### Expected behavior

The U shape should have both ends at the black line.

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

_No response_

### Additional information

_No response_

torokati44:
This is supposed to fix https://github.com/ruffle-rs/ruffle/issues/7565 by not using `lyon`'s `StrokeTessellator` - instead, for strokes, converting the Ruffle path to a `tiny_skia_path::Path` first, then running a `tiny_skia_path::PathStroker` on it, then converting that to a `lyon` path, and tessellating it as a fill with a non-zero fill rule.

It's dumb and probably slow, but hey, it seems to work.

Fixes #13656, and fixes #13658 too.


Dinnerbone:
> probably slow

Any metrics on this already or do you want me to benchmark something?

torokati44:
No metrics yet at all. I am currently just surprised that this approach even works as well as it appears to do (and first try, thanks Rust!).

I plan to do some benchmarking soon, but if you find the time and initiative, more eyes on it can't hurt.
Although I don't think this is in its final form yet.

Related: I suppose a wireframe rendering mode could come in handy for debugging/visualization/coolness purposes... I know it's not as simple now as back in the old `GL_LINE_LOOP` days, but any rudimentary solution would be fine for this.

torokati44:
Somewhat related, for debugging/verifying/inspecting purposes: https://github.com/ruffle-rs/ruffle/pull/11880

torokati44:
One small problem with this is that `tiny-skia-path` does not support different line cap styles at the start and the end of a path. I can try to add and PR this, but no promises. It's not that often needed, AFAIK only available through the `Graphics` drawing API (or something like that), and not from SWF tags.

Also, the exact dotting/dashing algorithm might be different (see the updated `avm2/graphics_direct_commands` expected test output) but I don't find this to be that big a deal.

And finally, this way of making a stroke mesh might, in some rare cases (such as in the linked z0r.de loop) produce highly suboptimal meshes, with lots of unnecessary internal vertices. While we could try to be clever and throw those away, I don't think the additional CPU overhead of the already slow tessellation justifies the handful of vertices saved for the GPU. I could be wrong though...

PR #11079, opened 2023-05-16
web: Minor improvements to the extension player page

Toad06:
- Use the `forceScale` option which makes games like [DTunnel 3D](https://uploads.ungrounded.net/174000/174308_nhsdtunnel3d.swf?1086971709) display as expected. This option is already in use on the web demo (cc. #10664).

- Pass parameters (flashvars) to the URL when Ruffle suggests to open the file in a separate tab due to CORS issues. This should allow more content to work in the extension player page. For instance, the two following embeds are now equivalent in this situation (in the current builds, only the second one would play):

```html
<embed type="application/x-shockwave-flash" src="http://data.hammerfest.fr/loader.swf?v=96" flashvars="$lang=fr&$out=/subscribe.html&swf=game.swf?v=101&$shake=1&$detail=1&$sound=1&$music=0&$volume=100&$mode=tutorial" />
```

```html
<embed type="application/x-shockwave-flash" src="http://data.hammerfest.fr/loader.swf?v=96&$lang=fr&$out=/subscribe.html&swf=game.swf?v=101&$shake=1&$detail=1&$sound=1&$music=0&$volume=100&$mode=tutorial" />
```

http://www.hammerfest.fr/try.html


n0samu (approved):

web/packages/core/src/ruffle-player.ts
```diff
@@ -1952,13 +1952,22 @@ export class RufflePlayer extends HTMLElement {
             this.isExtension &&
             window.location.origin !== this.swfUrl!.origin
         ) {
+            const url = new URL(this.swfUrl!);
+            if (this.loadedConfig?.parameters) {
+                const parameters = sanitizeParameters(
+                    this.loadedConfig.parameters
+                );
+                Object.keys(parameters).forEach((key) => {
```
relrelb, line 1960: nit: I think `Object.entries()` can be used here to iterate on both keys and values, then `parameters[key]!` turns into just `value`.

Toad06: Good point, I've made the changes.

relrelb (approved):

PR #11080, opened 2023-05-16
avm2: Add Matrix3D.[recompose,decompose,copyColumnTo,invert,determinant]

Aaron1011:
The Flash Player 'Matrix3D.recompose' method throws exceptions under certain circumstances with the "quaternion" orientation style. I haven't yet figured this out yet, so I've marked that case as a stub.

All of the implementations are based on the OpenFL code, with some tweaks to match Flash Player's behavior.


core/src/avm2/globals/flash/geom/Matrix3D.as
```diff
@@ -292,7 +295,292 @@ package flash.geom {
 		}
 
 		public function copyToMatrix3D(other:Matrix3D):void {
-			other.rawData = rawData
+			other.rawData = rawData;
+		}
+
+		// Based on OpenFL: https://github.com/openfl/openfl/blob/971a4c9e43b5472fd84d73920a2b7c1b3d8d9257/src/openfl/geom/Matrix3D.hx#L1437
```
iwannabethedev, line 301: If I understand it correctly, we are copy-pasting from sources where we do not necessarily understand the changes that we are copying. Normally, I really dislike copy-pasting code from a source where we do not necessarily understand the changes that were copied, especially so for libraries and foundational technologies like Ruffle. However, we are emulating existing technology that we can directly test against, so it is not nearly as bad as it could be. We still have some drawbacks from copy-pasting sources without understanding thoroughly what we are copy-pasting.

One potential issue is the quality of the code that we are copy-pasting, including whatever bugs they may have. Having some possibly buggy implementation is in many ways an improvement over not having any implementation (though we will still have to debug it later if there are bugs, while if it is not implemented, it is known for certain what its state is).

I tried to investigate where these changes come from in OpenFL, but I failed to do so. The closest I came was this file: https://github.com/openfl/openfl/blob/d29cba7a13c58a4090192cfef6b0e3f3314956d8/lib_esm/openfl/geom/Matrix3D.js , which only refers to ` "./../../_gen/openfl/geom/Matrix3D"`, and I would guess that the `_gen` in that path is some generated code.

I have only spent a tiny amount of time on reading this code and brushing up on this topic, and it is a long time since I last learned about linear algebra and related mathematics, and I have not worked a lot with 3D graphics, but, in regards to throwing exceptions, the original code from OpenFL may or may not be buggy if it does not throw exceptions, though this is guesswork on my part. For quaternions, 4x4 matrices, linear algebra, etc., there is a lot of literature out there, including at Wikipedia. Unity uses quaternions internally, and thus getting inspired by them might make sense (though, maybe open-source 3D graphics libraries are better sources). At https://docs.unity3d.com/ScriptReference/Matrix4x4-determinant.html , it says that *"You can not invert matrices with a determinant of zero."* . And I think the matrix of `testRecomposeWeird()` indeed has a determinant of 0, calculating from for instance https://www.wolframalpha.com/widgets/view.jsp?id=3c66255e08ebd4ee9c211ef0852ddc28 (though that calculator may be user-made and thus not reliable). But this is purely guesswork on my part and might be a coincidence or insufficient, etc., and I have not spent anywhere near the needed amount of time reading up on and understanding the relevant theory. A 3D graphics engine developer may be very helpful with figuring this stuff out.

iwannabethedev: Possibly relevant link: https://searchfox.org/mozilla-central/source/servo/components/style/values/animated/transform.rs#472-490 . In that function that mentions `recompose`, they seem to assume that the input is decomposed, or came from some decomposition.

iwannabethedev: I think it might be the case that the determinant plays some role in this: https://searchfox.org/mozilla-central/source/servo/components/style/values/animated/transform.rs#592-594 . In this function, where they decompose a 3D matrix, they calculate the determinant of `perspective_matrix`, and if it is `0`, they give an error. So, if the function for recomposition requires a matrix to be some sort of decomposed matrix or the result of decomposition; and in the method for decomposition, they give an error if the determinant of some intermediate result is `0`; then I think some check of the determinant might be what should throw an exception (though what the determinant should be calculated from specifically I am not certain about, since I have not really looked into the code or understood the theory really).

iwannabethedev:
(I have not reviewed the code, only looked at bits and pieces of it).

core/src/avm2/globals/flash/geom/Matrix3D.as
Aaron1011: The source for the OpenFL Matrix3D file is linked at the top of the file: https://github.com/openfl/openfl/blob/develop/src/openfl/geom/Matrix3D.hx

The important thing is to match Flash's behavior (including bugs), *not* to implement a generic matrix decomposition function. Flash does not throw an error for `testRecomposeWeird`, so neither should Ruffle.

iwannabethedev: > The source for the OpenFL Matrix3D file is linked at the top of the file: https://github.com/openfl/openfl/blob/develop/src/openfl/geom/Matrix3D.hx

Thank you, I overlooked that.

> The important thing is to match Flash's behavior (including bugs), not to implement a generic matrix decomposition function. Flash does not throw an error for testRecomposeWeird, so neither should Ruffle.

My apologies, I misunderstood what Flash was doing then, I thought Flash and this code was supposed to throw errors in certain cases.

Is the issue with `testRecomposeWeird()` different then? https://github.com/ruffle-rs/ruffle/pull/11080/files#diff-037d4e8ad2cf3cacd3fdafeb190793c9ec79d9627a76f97f5e405c8df9046920R10-R11 Like, it should return `NaN` instead of throw an exception? Given that you have commented the call to it out and mention that it should throw `NaN`.

I might have gotten it mixed up with the comment for https://github.com/ruffle-rs/ruffle/pull/11080/files#diff-037d4e8ad2cf3cacd3fdafeb190793c9ec79d9627a76f97f5e405c8df9046920R23-R33 , which does mention throwing an error. Might that throw due to some `determinant()` check?

iwannabethedev: I think OpenFL might have made the last main changes to `recompose()` in https://github.com/openfl/openfl/pull/270 .

core/src/avm2/globals/flash/geom/Matrix3D.as
```diff
@@ -584,3 +588,11 @@ package flash.geom {
 
 	}
 }
+
+import flash.geom.Orientation3D;
+
+function checkOrientation(orientationStyle:String) {
+	if (!(orientationStyle == Orientation3D.AXIS_ANGLE || orientationStyle == Orientation3D.EULER_ANGLES || orientationStyle == Orientation3D.QUATERNION)) {
+		throw new Error("Error #2187: Invalid orientation style badOrientation.  Value must be one of 'Orientation3D.EULER_ANGLES', 'Orientation3D.AXIS_ANGLE', or 'Orientation3D.QUATERNION'.", 2187);
```
Lord-McSweeney, line 596: Instead of using the string `badOrientation` this should use `orientationStyle`:
```
throw new Error("Error #2187: Invalid orientation style " + orientationStyle + ".  Value must be one of 'Orientation3D.EULER_ANGLES', 'Orientation3D.AXIS_ANGLE', or 'Orientation3D.QUATERNION'.", 2187);
```

core/src/avm2/globals/flash/geom/Matrix3D.as
```diff
@@ -292,7 +295,292 @@ package flash.geom {
...
+		}
+
+		// Based on OpenFL: https://github.com/openfl/openfl/blob/971a4c9e43b5472fd84d73920a2b7c1b3d8d9257/src/openfl/geom/Matrix3D.hx#L1437
+		public function recompose(components:Vector.<Vector3D>, orientationStyle:String = "eulerAngles"):Boolean {
+			if (orientationStyle == Orientation3D.QUATERNION) {
+				// Flash throws exceptions from 'recompose' certain values of 'components',
+				// which we need to reproduce. See the 'matrix3d_compose' test
+				stub_method("flash.geom.Matrix3D", "recompose", "Orientation3D.QUATERNION");
+			}
+			// RUFFLE - unlike in OpenFL, we continue on even if some of the 'scake' components are 0
```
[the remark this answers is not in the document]
Aaron1011: Fixed

Bale001 (approved):

PR #11082, opened 2023-05-17
Fix overflow bug in `shapeutils::winding_number_line()`.

### LINKED ISSUE #11077: First bowser fight in Super Mario 63 is not completeable
### Describe the bug

Mario spawns inside a platform near the pipe that takes you to bower and is stuck there. image

Steps to reproduce:

-Take the shortcut to the bowser 1 fight on a new save file (shortcut seen here: https://youtu.be/3ZnULk4CA60?t=41 )

### Expected behavior

Mario spawns at the start of the level and is able to complete the level and bowser fight
image

### Affected platform

Online demo

### Operating system

Manjaro Linux, x86_64, Linux 6.1.25-1-MANJARO

### Browser

Librewolf 102.0.1

### Additional information

This only occurs in ruffle versions 2023-05-15 and 2023-05-16, versions prior work fine, I am playing the 2009 version of Super Mario 63

The link to the swf is here https://cdn.discordapp.com/attachments/704769195749146634/1108132691246862406/Super_Mario_632009_Ver..swf

This same thing also happens in another flash implementation called waflash (likely not relevant or helpful but I thought I'd mention it)

iwannabethedev:
Fixes #11077 .

Related PR: #11014 .

An automated test has been added.

### DIFF
--- a/render/src/shape_utils.rs
+++ b/render/src/shape_utils.rs
@@ -532,123 +532,6 @@ impl<'a> ShapeConverter<'a> {
     }
 }
 
-#[cfg(test)]
-mod tests {
-    use super::*;
-    use swf::PointDelta;
-
-    const FILL_STYLES: [FillStyle; 1] = [FillStyle::Color(swf::Color {
-        r: 255,
-        g: 0,
-        b: 0,
-        a: 255,
-    })];
-
-    const LINE_STYLES: [LineStyle; 0] = [];
-
-    /// Convenience method to quickly make a shape,
-    fn build_shape(records: Vec<ShapeRecord>) -> swf::Shape {
-        let bounds = calculate_shape_bounds(&records);
-        swf::Shape {
-            version: 2,
-            id: 1,
-            shape_bounds: bounds.clone(),
-            edge_bounds: bounds,
-            flags: swf::ShapeFlag::HAS_SCALING_STROKES,
-            styles: swf::ShapeStyles {
-                fill_styles: FILL_STYLES.to_vec(),
-                line_styles: LINE_STYLES.to_vec(),
-            },
-            shape: records,
-        }
-    }
-
-    /// A simple solid square.
-    #[test]
-    fn basic_shape() {
-        let shape = build_shape(vec![
-            ShapeRecord::StyleChange(Box::new(swf::StyleChangeData {
-                move_to: Some(swf::Point::from_pixels(100.0, 100.0)),
-                fill_style_0: None,
-                fill_style_1: Some(1),
-                line_style: None,
-                new_styles: None,
-            })),
-            ShapeRecord::StraightEdge {
-                delta: PointDelta::from_pixels(100.0, 0.0),
-            },
-            ShapeRecord::StraightEdge {
-                delta: PointDelta::from_pixels(0.0, 100.0),
-            },
-            ShapeRecord::StraightEdge {
-                delta: PointDelta::from_pixels(-100.0, 0.0),
-            },
-            ShapeRecord::StraightEdge {
-                delta: PointDelta::from_pixels(0.0, -100.0),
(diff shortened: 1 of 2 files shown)


render/src/shape_utils.rs
```diff
@@ -1006,16 +1006,54 @@ fn winding_number_line(
...
+        && (d1.dx.get() as i64) * (d0.dy.get() as i64) < (d1.dy.get() as i64) * (d0.dx.get() as i64)
+    {
         return -1;
     }
 
     0
 }
 
+#[cfg(test)]
+mod test {
```
relrelb, line 1025: Can this be placed at bottom of the file please?

iwannabethedev:
Credit to @Toad06 , @swaggybaguette , @n0samu and others for reporting, debugging and helping fix this bug.

render/src/shape_utils.rs
iwannabethedev: I initially placed it at the bottom, but then I happened upon other tests in the same file that were placed further up:

https://github.com/ruffle-rs/ruffle/blob/ba3aa1b73b6ebcbe80f72ce6afa8ed680611667a/render/src/shape_utils.rs#L535-L571

Should I still move it down to the bottom? And should I leave the existing tests where they are (to keep the PR clean)?

iwannabethedev:
I think I could have very easily introduced that overflow bug as well, and I did not discover it at all when I reviewed the code carefully, which is the main reason why I added a documentation comment to `Twips`, since it is really non-obvious in my opinion.

render/src/shape_utils.rs
relrelb: Oh, I didn't notice that. So let's keep this as-is. However I did just notice that the vast majority of codebase uses `mod tests` and not `mod test` (without an `s`). Can you please add an `s` to align to the rest of the codebase?

iwannabethedev: I think I will have to move them all down:

```
error[E0428]: the name `tests` is defined multiple times
    --> render/src/shape_utils.rs:1020:1
     |
536  | mod tests {
     | --------- previous definition of the module `tests` here
...
1020 | mod tests {
     | ^^^^^^^^^ `tests` redefined here
     |
     = note: `tests` must be defined only once in the type namespace of this module

For more information about this error, try `rustc --explain E0428`.
error: could not compile `ruffle_render` due to previous error
warning: build failed, waiting for other jobs to finish...
```

iwannabethedev: I got the `mod test` from here: https://github.com/ruffle-rs/ruffle/blob/f969bdce710b7911e7fe6ad96a95ba11fa67acb2/render/src/bitmap.rs#L501 , I did not realize the difference.

iwannabethedev: I have moved all the tests down to the button, and ensured that the name is `mod tests`.

render/src/shape_utils.rs
```diff
@@ -1283,3 +1171,149 @@ impl<'a> LineScales<'a> {
...
+
+    #[test]
+    fn test_winding_number_line() {
+        fn test(
+            test_point: swf::Point<Twips>,
+            begin: swf::Point<Twips>,
+            end: swf::Point<Twips>,
+            expected: i32,
+        ) {
+            let result = crate::shape_utils::winding_number_line(test_point, begin, end);
```
relrelb, line 1301: nit: Actually I'm almost sure that a fully qualified name is unnecessary here. It can be referenced as simply as `winding_number_line()`.

iwannabethedev: You are right, when I first wrote it, I tried without any qualification, and that did not work for some reason, but when I try now, it works.

relrelb (approved):
Thank you!

PR #1111, opened 2020-09-04
core: Call post_instantiate on initial empty root (fix #946)

Herschel:
The `fake_root` did not have an object, which could cause the
player to panic if the SWF was not completely loaded when playing.
Calling `post_instantiate` ensures that this dummy root has an
object.

PR #965, opened 2020-07-31
Fix crash when user presses "play" way too quickly.

kmeisthax:
I fixed an unfortunate bug caused by some of the initialization changes I made in `movieorigin`. This ensures there is always a root movie, so the user can't catch the player half-initialized.

Issue #946, opened 2020-07-28
web: Ruffle's Play button appears too soon (regression?)


Toad06:
Using the nightly build of the day (2020-07-28), if you click the Play button as soon as it appears, you get the following error:

> panicked at 'root level', core/src/player.rs:598:26

... which in turn causes other errors, preventing the game from running.

This can be reproduced with any game (make sure to disable the cache while testing).

This build also adds some black borders around the white background.

https://www.newgrounds.com/portal/view/225944/
https://www.y8.com/games/streets_of_fire

Dinnerbone:
Ran a bisect.

| Status | SHA | Message |
| ------ | ----- | --------- |
| :x: | aed47d458d7b8b9c510b92fd39f9d56e049cb61c | Level loads in GetURL2 should also propagate origin information.|
| :heavy_minus_sign: | a34e81a70415bc8a83bf5dd3081918e0de30a483 | `_url` on desktop should always return a file URL for file-loaded movies.|
| :heavy_minus_sign: | f0e2c77c1f02255d27ea4c1fa7af86ea2aa47875 | URLs from paths is a desktop-only feature.|
| :heavy_minus_sign: | 9b9d4076fe770d58185d2067ce695dc736650f4f | Expose movie URL to ActionScript.|
| :heavy_minus_sign: | a8877ab63c0735f45a44752802a1fc430f047953 | Yield the correct error when the root movie load fails.|
| :heavy_minus_sign: | d1724416635a7cf2a51cb960e0461fca68c420ae | Also make absolute URLs for all movie loads passed through the load manager.|
| :heavy_minus_sign: | 5d15f5bfe3bcc5615ffd062c360f2caf5de10343 | When loading a movie from the filesystem outside of the core, ensure that the URL is properly |
| :heavy_minus_sign: | c926da8888cfbb33fa8ae289a3ff53cd7127677f | Refactor URL relativization into two utility methods that backends can provide base URLs and |
| :heavy_minus_sign: | 7433bfe28f34253ab6802da4ff0ed7ed576a4144 | Add a `NavigatorBackend` method to resolve relative URLs.|
| :heavy_minus_sign: | 4813942fe7ce07037bd6346eacf45ae29ebd41d9 | The player should always change the audio backend's framerate itself.|
| :heavy_minus_sign: | f56d16a68d7518359c896ee59da54a9f6d13b313 | Separate player creation from root movie setup, and allow users of Ruffle to load in movies |
[...]

PR #11083, opened 2023-05-17
core: Don't create a fake root clip when instantiating Stage

Aaron1011:
AVM2 can handle a missing root movie, and AVM1 doesn't seem to rely on having a root movie during Stage initialization.


Toad06:
For reference, the code that this PR removes was added in #965 and #1111 a while ago, to prevent a potential crash on the web builds (#946). However, many things have changed since then, and removing this code doesn't seem to cause any particular issue.

Dinnerbone (approved):

relrelb (approved):

Issue #8409, opened 2022-10-29
Game info not displaying in Canterlot Siege


Computerdude77:
### Describe the bug

In the game [Canterlot Siege](https://archive.org/details/canterlot_siege_by_futzi01-d5bfpyv), there is an issue with the in-game info (lives, waves, etc.). The numbers are not being shown at all, with the exception of the timer, which works perfectly. I am aware that there is also the same issue that all tower defense games seem to have at the moment that is likely related to #3839, but this is not specifically about that, and I thought I would mention it because it will likely be pointed out as well. Here is a screenshot demonstrating the issue:

![chrome_4r22g32tsD](https://user-images.githubusercontent.com/55714746/198820422-6ba90323-de50-4f2e-8c4c-6d759b12a79f.png)

### Expected behavior

This is a screenshot from native Flash, and you can see what is supposed to be displayed in the corner of the screen:

![flashplayer_32_sa_tPESrawyrN](https://user-images.githubusercontent.com/55714746/198820459-91645fac-e32d-443a-9177-4de78852ed63.png)

### Affected platform

Online demo

### Operating system

Windows 10 21H2

### Browser

Google Chrome 107

### Additional information

The version on Archive.org (2022-09-19) is older than the web demo this was tested on at the time of writing this (2022-10-28), but the issue is the same regardless of version.

Lord-McSweeney:
Console repeatedly errors "Attempted to add frame scripts to non-MovieClip this!".

Computerdude77:
As of the 2023-04-10 nightly, the issue with the UI numbers has been fixed. However, the issues documented above that involve the targeting still occur, so even though the original issue this was made for was fixed, I will leave this issue open until the above issues with the targeting are fixed too.

![chrome_e9RgCzn2ZF](https://user-images.githubusercontent.com/55714746/230843727-4b7488d8-95d9-4f6c-b5f1-a676a3a3785a.png)

Computerdude77:
As of the 2023-05-29 nightly, the targeting bug has been fixed (I can even see the commit where it was fixed referenced above). The only other issue that seems to occur with this game is the music peaking sometimes when looping. I have documented a seemly identical issue with Spike Toss, so I am going to go ahead and close the issue, even with this sound bug. If it turns out to be an unrelated issue, someone can open up a separate issue for it.

Issue #11051, opened 2023-05-14
Ruffle extension failed to load one of my favorite flash games.


KrzysztofoGuard:
On this page | https://supermariobros.online/super-goblin-war-machine | The Ruffle extension cannot run one of my favorite flash games and shows a black screen. I am asking the Ruffle team to look into this issue and bringing this game to life.

Lord-McSweeney:
This is due to unimplemented `SharedObject.setProperty()`.

n0samu:
The game appears to work now, but it is very slow. Tested via [Flashpoint's testing site](https://ooooooooo.ooo/static/?68dad594-9523-4afb-846d-873a7c821125) since the original link is dead.

Issue #10208, opened 2023-03-18
Gravity Master: Flickering menu buttons


waspennator:
### Describe the bug

[1_gravity-master.zip](https://github.com/ruffle-rs/ruffle/files/11007660/1_gravity-master.zip)
Game seems to work but the buttons have a major issue, they keep on flickering a bunch, making it hard to click on em. Just load up the SWF and you'll see it right away.

### Expected behavior

The buttons should be looking normal and not flickering a bunch.

### Affected platform

Desktop app

### Operating system

Windows 11

n0samu:
Fixed by #10734

PR #11084, opened 2023-05-17
avm2: Move DisplayObject assignment into an initializer

Aaron1011:


core/src/avm2/globals/flash/display/shape.rs
```diff
@@ -1,11 +1,47 @@
...
+
+            return initialize_for_allocator(activation, child, orig_class);
+        }
+        class_object = class.superclass_object();
+    }
+    unreachable!("A Shape subclass should have Shape in superclass chain");
+}
 
 /// Implements `flash.display.Shape`'s 'init' method, which is called from the constructor
 pub fn init<'gc>(
```
Lord-McSweeney, line 47: This init method can now be entirely removed.

core/src/avm2/globals/flash/utils/byte_array.rs
```diff
@@ -20,27 +19,6 @@ pub fn init<'gc>(
 ) -> Result<Value<'gc>, Error<'gc>> {
```
Lord-McSweeney, line 19: This init method can now be entirely removed.

Lord-McSweeney:
Progresses #11051.

core/src/player.rs
```diff
@@ -2342,9 +2342,9 @@ impl PlayerBuilder {
         let mut player_lock = player.lock().unwrap();
         player_lock.mutate_with_update_context(|context| {
             // Instantiate an empty root before the main movie loads.
-            let fake_root = MovieClip::new(fake_movie, context.gc_context);
-            fake_root.post_instantiation(context, None, Instantiator::Movie, false);
-            context.stage.replace_at_depth(context, fake_root.into(), 0);
+            //let fake_root = MovieClip::new(fake_movie, context.gc_context);
```
Lord-McSweeney, line 2345: Are these changes intentional, or left over from a previous PR?

Aaron1011: This is https://github.com/ruffle-rs/ruffle/pull/11083

core/src/avm2/globals/flash/display/shape.rs
```diff
@@ -15,12 +51,6 @@ pub fn init<'gc>(
 ) -> Result<Value<'gc>, Error<'gc>> {
     if let Some(this) = this {
         activation.super_init(this, &[])?;
```
Lord-McSweeney, line 53: This init method can be removed now, right?

core/src/avm2/globals/flash/text/text_field.rs
```diff
@@ -11,6 +12,41 @@ use crate::string::AvmString;
...
+
+            return initialize_for_allocator(activation, child, orig_class);
+        }
+        class_object = class.superclass_object();
+    }
+    unreachable!("A TextField subclass should have TextField in superclass chain");
+}
+
 /// Implements `flash.text.TextField`'s `init` method, which is called from the constructor.
 pub fn init<'gc>(
```
Lord-McSweeney, line 51: This init method can be removed now.

core/src/avm2/object/bytearray_object.rs
```diff
@@ -14,14 +15,34 @@ pub fn byte_array_allocator<'gc>(
...
+        .context
+        .library
+        .avm2_class_registry()
+        .class_symbol(class)
+    {
+        if let Some(lib) = activation.context.library.library_for_movie(movie) {
+            if let Some(Character::BinaryData(binary_data)) = lib.character_by_id(id) {
+                Some(ByteArrayStorage::from_vec(binary_data.as_ref().to_vec()))
+            } else {
+                None
```
Lord-McSweeney, line 28: This else is still reachable: for example, a class that extends `ByteArray` but doesn't correspond to a `BinaryData` in the symbol class.

Aaron1011: This only occurs if we're successfully looked up the `class_symbol`.

core/src/player.rs
Aaron1011: My changes previously needed this change, but with the latest version of this PR, all of the tests pass with the fake root in place. I've re-added this code.

core/src/avm2/globals/flash/display/display_object.rs
```diff
@@ -21,7 +20,43 @@ use crate::vminterface::Instantiator;
...
+use swf::Twips;
+use swf::{BlendMode, Rectangle};
+
+pub fn display_object_allocator<'gc>(
+    class: ClassObject<'gc>,
+    _activation: &mut Activation<'_, 'gc>,
+) -> Result<Object<'gc>, Error<'gc>> {
+    // todo we want an Err here
+    // as a script can do `new DisplayObject()`
+    panic!("This allocator should not be called! {:?}", class);
```
Lord-McSweeney, line 32: Error here: `ArgumentError: Error #2012: DisplayObject$ class cannot be instantiated.`

core/src/avm2/globals/flash/display/simple_button.rs
```diff
@@ -22,31 +62,39 @@ pub fn init<'gc>(
     if let Some(this) = this {
         activation.super_init(this, &[])?;
```
Lord-McSweeney, line 63: This use of `super_init` can be removed.

core/src/avm2/globals/flash/display/bitmap.rs
```diff
@@ -12,13 +13,68 @@ use crate::character::Character;
...
+
 /// Implements `flash.display.Bitmap`'s `init` method, which is called from the constructor
 pub fn init<'gc>(
     activation: &mut Activation<'_, 'gc>,
     this: Option<Object<'gc>>,
     args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
-    if let Some(mut this) = this {
+    if let Some(this) = this {
         activation.super_init(this, &[])?;
```
Lord-McSweeney, line 78: This use of `super_init` can now be removed.

core/src/avm2/globals/flash/display/display_object.rs
Aaron1011: This should never be hit - the actual error gets thrown from `DisplayObject.as`

Lord-McSweeney: This can still be hit by a class that extends DisplayObject, which is unintentional. See this SWF for example, which panics with this PR but runs fine (no errors) on master, logging `[object DO]`:
[avm2-tests.swf.zip](https://github.com/ruffle-rs/ruffle/files/11521182/avm2-tests.swf.zip)

Both this PR and master's state is incorrect, though- in FP, it errors with `ArgumentError: Error #2012: DO$ class cannot be instantiated.`

Aaron1011: Good catch - I've added a fix and a test case.

core/src/avm2/globals/flash/display/display_object.rs
```diff
@@ -21,7 +20,51 @@ use crate::vminterface::Instantiator;
...
+
+pub fn display_object_allocator<'gc>(
+    class: ClassObject<'gc>,
+    activation: &mut Activation<'_, 'gc>,
+) -> Result<Object<'gc>, Error<'gc>> {
+    let class_name = class
+        .inner_class_definition()
+        .read()
+        .name()
+        .to_qualified_name(activation.context.gc_context);
```
Lord-McSweeney, line 34: This is wrong because it prints the fully qualified class name: for example, in Ruffle, `new DisplayObject()` prints `ArgumentError: Error #2012: flash.display:DisplayObject$ class cannot be instantiated.`, while in Flash it prints `ArgumentError: Error #2012: DisplayObject$ class cannot be instantiated.`
(it's probably worth it adding a test that checeks the error message for `new DisplayObject()`.)

core/src/avm2/object/bitmapdata_object.rs
```diff
@@ -52,7 +52,16 @@ pub struct BitmapDataObjectData<'gc> {
...
+    // Constructs a BitmapData object from a BitmapDataWrapper.
+    // This is *not* used when explicitly constructing a BitmapData
+    // instance from ActionScript (e.g. `new BitmapData(100, 100)`,
+    // or `new MyBitmapDataSubclass(100, 100)`).
+    //
+    // Instead, this is used when constructing a `Bitmap` object,
+    // (from ActionScript or from the timeline), or when we need
+    // to produce a new BitmapData object from a `BitmapData` method
+    // like `clone()`
+    pub fn from_bitmap_data_interal(
```
Lord-McSweeney, line 64: Typo: "internal" instead of "interal"?

core/src/avm2/globals/flash/display/display_object.rs
Aaron1011: Fixed

core/src/avm2/object/bitmapdata_object.rs
Aaron1011: Fixed

core/src/avm2/globals/flash/text/static_text.rs
```diff
@@ -1,17 +1,18 @@
-use crate::avm2::{Activation, Error, Object, Value};
+use crate::avm2::error::argument_error;
+use crate::avm2::{Activation, ClassObject, Error, Object, Value};
 use crate::avm2_stub_getter;
 
-pub fn native_instance_init<'gc>(
+pub fn static_text_allocator<'gc>(
```
Lord-McSweeney, line 5: Won't a class extending `StaticText` hit this error when it tries to construct? IIRC that was the point of having a native instance initializer.

Aaron1011: `StaticText` is final, so that can't happen.

core/src/avm2/globals/flash/text/StaticText.as
```diff
@@ -1,11 +1,8 @@
 package flash.text {
     import flash.display.DisplayObject;
     
-    [Ruffle(NativeInstanceInit)]
+    [Ruffle(InstanceAllocator)]
     public class StaticText extends DisplayObject {
```
Lord-McSweeney, line 5: Note: `StaticText` should be a final class, should this be changed here?

Dinnerbone (approved):

core/src/avm2/globals/flash/text/StaticText.as
Aaron1011: Done

core/src/avm2/globals/flash/geom/sniper_team_notes
```diff
@@ -0,0 +1,179 @@
+Glow filter: InterfaceDeFreezeAnimationMC
```
Lord-McSweeney, line 1: Accidentally added?

core/src/avm2/globals/flash/media/sound.rs
```diff
@@ -19,8 +19,6 @@ pub fn init<'gc>(
     args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
```
Lord-McSweeney, line 20: Should Sound also be moved to an instance allocator?

SN902:
I found regression, this commit brings back flickering buttons/sliders in #10208 and [PortageCS5.zip](https://github.com/ruffle-rs/ruffle/files/11525396/PortageCS5.zip)

core/src/avm2/globals/flash/media/sound.rs
Aaron1011: I added a test - it looks like Sound is a 'normal' object (the initialization happens during the `super()` call, not during the allocator).

kmeisthax (commented):
I'm not sure why `super_init` got removed from `Sound` and `ApplicationDomain` - is there some subtlety of initialization I'm missing? I thought they were "normal" (i.e. no native-allocated fields). I try to keep `super_init` in native code since the AS3 compiler automatically inserts `constructsuper` everywhere anyway, but this isn't particularly important anymore.

The regression @SN902 found is worth looking into but I won't be able to do that for a few more days. Looking at the original PR #10734 it looks like you correctly carried forward everything from that, so it SHOULDN'T be broken, which suggests that we might need to move the check somewhere else, or set a flag here that play/goto checks for.

Aaron1011:
@kmeisthax: The 'super_init' removes were drive by fixes. Those calls were in `init` methods, which are just normal methods called from the constructor (the `super_init` call was left over when we defined the class in actionscript).

I discussed the regression on Discord, but forgot to leave a comment - I fixed it in [`ad15328` (#11084)](https://github.com/ruffle-rs/ruffle/pull/11084/commits/ad153288aaac26eb2d2c1df9e39b5bb7914a55c0)

core/src/display_object/graphic.rs
```diff
@@ -71,10 +71,7 @@ impl<'gc> Graphic<'gc> {
     }
 
     /// Construct an empty `Graphic`.
-    pub fn new_with_avm2(
-        context: &mut UpdateContext<'_, 'gc>,
-        avm2_object: Avm2Object<'gc>,
-    ) -> Self {
+    pub fn new_with_avm2(context: &mut UpdateContext<'_, 'gc>) -> Self {
```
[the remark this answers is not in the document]
Aaron1011: I've renamed this to `empty`

core/src/display_object/loader_display.rs
```diff
@@ -36,25 +36,22 @@ impl fmt::Debug for LoaderDisplay<'_> {
...
     movie: Arc<SwfMovie>,
 }
 
 impl<'gc> LoaderDisplay<'gc> {
-    pub fn new_with_avm2(
-        activation: &mut Activation<'_, 'gc>,
-        avm2_object: Avm2Object<'gc>,
-        movie: Arc<SwfMovie>,
-    ) -> Self {
+    pub fn new_with_avm2(activation: &mut Activation<'_, 'gc>, movie: Arc<SwfMovie>) -> Self {
```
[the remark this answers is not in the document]
Aaron1011: I've renamed this to `empty`

n0samu:
With this PR, #8947 is occurring again 😦 
```
ERROR run_frame:run_all_phases_avm2: ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "addedToStage", class: flash.events::Event, ptr: 0x218a7850388 }) to handler FunctionObject(FunctionObject { ptr: 0x218a27e0a88 }) : TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: visible)
        at classes::Level/holderOnAddedToStage()
        at flash.display::DisplayObjectContainer/flash::display::DisplayObjectContainer::addChildAt()
        at classes.app::Main/createLevel()
        at classes.app::Main/nextLevel()
        at classes::ShutterIn/onEnterFrame()
```

KrzysztofoGuard:
Tell me will the game I submitted for revival to the Ruffle team will work or are you already discouraged?  I am asking for a specific answer and I realize that there have been any problems recently.

n0samu:
@KrzysztofoGuard The answer to your question is here https://github.com/ruffle-rs/ruffle/issues/11051#issuecomment-1546905360. There some more things we need to figure out before the game will work at all.

waspennator:
Should close #8409 with pony towers being unable to damage targets.

SN902:
Found regression in [V8MuscleCars.zip](https://github.com/ruffle-rs/ruffle/files/11579989/V8MuscleCars.zip) after intro this game stuck on black screen with ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x2a000207ba8 }) to handler FunctionObject(FunctionObject { ptr: 0x2a07acda048 }) : TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: numChildren)

Aaron1011:
@SN902 I pushed a new commit which should fix the regression

SN902:
> @SN902 I pushed a new commit which should fix the regression
Fixed

adrian17 (approved):

PR #11095, opened 2023-05-17
extension: Add a button to reset settings

Toad06:


n0samu (approved):
Had to compile the PR just to see what the button would look like 😅 
Anyway, looks good and works as expected!
image

Toad06:
Thanks for testing and yeah, this is a simple button. 😄

By the way, the motivation behind this PR is of course to provide a way to reset the settings for users that messed up with the config, but also in case we need to use an older version of the extension as some settings may have changed and work differently.

This is for instance the case with the `maxExecutionDuration` setting (the value was an object, but is now a number): loading a file in the extension player page could fail and trigger the error (e.g. when using nightly-2023-02-20):
> Serious error occurred loading SWF file: Error parsing config: Error: invalid type: floating point `15`, expected struct Duration

iwannabethedev:
The changes look very good to me, and I believe that they are ready to be merged.

Dinnerbone (approved):

Issue #11098, opened 2023-05-18
Extension not updated on the Chrome Web Store after nightly 2023-05-15


n0samu:
The Ruffle extension is not being updated on the Chrome Web Store after nightly 2023-05-15 because the Chrome Web Store is rejecting the automatic upload with [the following error](https://github.com/ruffle-rs/ruffle/actions/runs/5009016209/jobs/8977478905#step:19:26):
```
Error: HTTPError: Response code 400 (Bad Request)
Error: publish error - You will need to access the Chrome Web Store Developer Dashboard and publish manually.
```
At the moment we don't know the cause of the error or what we can do about it. I am preemptively creating this issue since someone is going to notice it sooner or later. Personally I don't have access to the relevant Google account or Chrome Web Store listing so I don't have any insight.

n0samu:
After a bit of investigation, I'm willing to bet that Google is rejecting our updates because they are flagging the escaped Unicode localization strings as 'obfuscation'. Comparison below:
**2023-05-15**:
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/be9de126-43a0-4124-9216-c6fed4d17cb0)
**2023-05-16**:
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/a631ad5b-1f4c-4cd7-9554-edb7ab10f3ef)

We started escaping strings in #11023 because the Ruffle player was breaking on non-UTF8 pages. We could just flag the script as UTF-8 by adding the `charset` attribute to the script tag, but the escaping solution was favored because it was thought that it would also apply to selfhosted builds, while allowing site owners to avoid needing to add an extra attribute to their script tags. But we later found out that selfhosted builds needed a separate solution for escaping Unicode characters anyway (#11058). So even if this is not indeed the reason our extension updates are being rejected, I still think we should use the charset attribute in the extension's script tag, instead of going with this more complicated escaping method.

n0samu:
Actually it seems more likely that this problem was caused by #11040, which added an additional permission request that might raise suspicion. And for a relatively unimportant purpose too (as I tried to imply with my comment on the PR).

n0samu:
Fixed by #11103 (Google didn't like us adding the new permission, and once we reverted they started accepting our submissions again)

PR #11099, opened 2023-05-18
extension: Revert #11040

n0samu:
Reverts ruffle-rs/ruffle#11040 per https://github.com/ruffle-rs/ruffle/issues/11098#issuecomment-1552649360. This was an unnecessary addition to Ruffle's required permissions which may have tripped Google's systems and caused our updates to be rejected.

### DIFF
--- a/web/packages/extension/assets/_locales/en/messages.json
+++ b/web/packages/extension/assets/_locales/en/messages.json
@@ -56,9 +56,6 @@
     "status_result_running": {
         "message": "Ruffle is loaded and ready to run Flash content on the current tab."
     },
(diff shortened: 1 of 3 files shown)


adrian17 (approved):

Toad06:
Instead of reverting this completely (as Firefox doesn't need the permission to work), could we just remove the permission from `manifest.json` and add a TODO or FIXME in `popup.ts` explaining why it wouldn't work in Chrome?

n0samu:
Yes that sounds fine, can you open a new PR for that please?

n0samu:
Closing in favor of #11103

PR #11103, opened 2023-05-18
extension: Remove "tabs" permission

Toad06:
Attempt at fixing #11098 as per https://github.com/ruffle-rs/ruffle/issues/11098#issuecomment-1552649360.

Supersedes #11099.

### DIFF
--- a/web/packages/extension/manifest.json5
+++ b/web/packages/extension/manifest.json5
@@ -45,7 +45,6 @@
     "permissions": [
         "<all_urls>",
         "storage",
-        "tabs",
         "webRequest",
         "webRequestBlocking",
     ],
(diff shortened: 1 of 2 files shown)


n0samu (approved):
Can confirm that this works, thanks! We might want to consider using the [activeTab permission](https://developer.chrome.com/docs/extensions/mv3/manifest/activeTab/) for Chrome in a future PR, after the extension is back to publishing normally. (But we would need to make sure we only merge that PR when we know someone can monitor and respond to issues with Chrome Web Store publishing.)

adrian17 (approved):

Aaron1011 (approved):

Issue #10962, opened 2023-05-06
Development and tracking issue for `BitmapData.pixelDissolve`.


iwannabethedev:
Brief: `BitmapData.pixelDissolve` may be a more complicated method to figure out how to reverse engineer and implement. This issue is meant to help both coordinate and track its implementation and test.

Thanks and credit to @Dinnerbone and others for their main work as part of figuring out this issue.

### Reverse engineering

API descriptions:
- https://open-flash.github.io/mirrors/as2-language-reference/flash/display/BitmapData.html#pixelDissolve()
- https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/display/BitmapData.html#pixelDissolve()

(Warning: The example for AVM1 appears correct, while the example for AVM2 appears wrong, given that it does not appear to use the returned random seed).

On why the method might be challenging to reverse-engineer (reading at least the first article is recommended):

- https://fabiensanglard.net/fizzlefade/index.php
- http://antirez.com/news/113
- https://web.archive.org/web/20081217091012/http://www.quasimondo.com/archives/000681.php

In particular, it might well be desirable to achieve pixel accuracy, such that it is not only visual similarity that is achieved, but the same values that are generated for the same input between the official Flash player and Ruffle. This might be challenging to figure out, in particular since the implementation most likely uses some slightly involved algorithmic approach to ensure good performance, such as linear feedback shift register (to visit each pixel at most once).

See the below comments for the main resources and findings on this topic.

### Tasks

Testing:

- [x] Correct number of pixels are changed for different inputs.
- [x] Same image as `src` and `dst`.
- [x] Different image as `src` and `dst`.
- [x] Different colors (how does it handle alpha?).
- [x] Transparent `src` onto opaque `dst`.
- [x] Transparent `src` onto transparent `dst`.
- [x] What happens if the sizes of the images do not fit? Other similar dimensional testing?
- [ ] Performance testing? (Maybe just one basic test with a semi-large image).
- [x] Testing both AVM1 and AVM2 (the API descriptions are very similar, so the behavior might be the same, though this is not certain).

iwannabethedev:
Findings on prior open-source art, based on discussions elsewhere and https://github.com/ruffle-rs/ruffle/wiki/Helpful-Resources#similar-projects (**please be careful that any source code you find is appropriately licensed**):

### https://github.com/openfl/openfl
[...]

Issue #290, opened 2020-01-11
Actionscript Tracking: flash.display.BitmapData


Dinnerbone:
This is the tracking issue for **flash\.display\.BitmapData** in AVM1. We will check off each item of progress as appropriate, and fill in any relevant or missing information as we continue development of Ruffle.
# Legend
Each checkbox is independent of another. It's entirely possible for something to be tested but not exist yet, or for us to believe that it's completely functional but we haven't made enough tests to prove it.
## "Property Exists"
This means the item exists, but may not necessarily be fully implemented.
## "Has Test Coverage"
This means that we have believe that we have a good test coverage of this item, regardless of if those tests pass. It's okay to have tests available but not implement the item yet.
## "Completely Functional"
This means we believe that the item is completely implemented, and no more work needs to be done towards making it functional.

# flash\.display\.BitmapData
### `loadBitmap(id)`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional

### `width : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `height : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `transparent : Boolean`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `rectangle : Rectangle`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional

### `BitmapData (width, height, transparent, fillColor)`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `clone()`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `getPixel(x, y)`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `getPixel32(x, y)`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `setPixel(x, y, color)`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `setPixel32(x, y, color)`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `applyFilter (sourceBitmap, sourceRect, destPoint, filter)`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `colorTransform(rect, colorTransform)`
[...]

PR #11106, opened 2023-05-18
core: Implement `BitmapData.pixelDissolve()` for AVM1 and AVM2.

iwannabethedev:
The changes have automated tests for them.

`BitmapData.pixelDissolve()` has been implemented for both AVM1 and AVM2. For both AVM1 and AVM2, there is a visual test folder and a non-visual test folder.

Credit and thanks to: @Dinnerbone , @n0samu, @moulins and others.

Development issues:
- #10962 .
- #290 .

(Previous questions have been handled).

- ~~...~~ Support for AVM1 added.
- ~~...~~ `TODO`s have been handled.
- ~~...~~ Code organization has been changed.


core/src/avm2/globals/flash/display/bitmap_data.rs
```diff
@@ -1336,11 +1336,15 @@ pub fn compare<'gc>(
     {
         other_bitmap_data
     } else {
-        // The documentation says that -1 should be returned here, but -2 is actually returned.
+        // The documentation for AVM1 says that -1 should be returned here,
+        // but -2 is actually returned.
+        // TODO: Has this been tested for AVM2?
```
Lord-McSweeney, line 1341: I don't think this branch can be reached in AVM2 since AVM2 enforces type-checking.

iwannabethedev:
I ended up not implementing the algorithm that Flash Player uses, namely Linear Feedback Shift Register, for multiple different reasons, in particular that you would have to build up a table for each range of sizes of the bitmap. Also, this version has other implementation differences (I might not have listed them all here):

- It handles correctly `1*x` and `x*1` `BitmapData`s, which Flash Player does not handle correctly.
- It deviates from how Flash Player does rounding - Flash Player apparently does rounding for `BitmapData.pixelDissolve()` in an in my opinion really strange and peculiar way. To see this, comment in some of the tests that I commented out and see the test results and the differences.

core/src/avm2/globals/flash/display/bitmap_data.rs
iwannabethedev: Should I replace it with this?:

```rust
    let other_bitmap_data = args
        .get_object(activation, 0, "otherBitmapData")?
        .as_bitmap_data()
        .unwrap(); // The AVM2 enforces type-safety.
```

iwannabethedev: No wait, can that become `null`?

iwannabethedev: Would it be alright if I instead replace it with:

```rust

        // The documentation for AVM1 says that -1 should be returned here,
        // but -2 is actually returned.
        // TODO: For AVM2, this branch should never get reached, since
        //   AVM2 checks types.
```

? Since I would prefer not to change the behavior of that function in this PR (it might have been/be better to leave out these `TODO`s of the PR).

iwannabethedev:
I have found a bug that my tests did not catch.

iwannabethedev:
Postponing AVM1 support until later (likely someone other than me submitting a PR).

iwannabethedev:
I have implemented support for AVM1, and ported one of the test folders to AVM1. However, I did not succeed in porting the image test folder to AVM1, and I think I can not do that reasonably effectively without Adobe tooling, which I do not have access to. I therefore hope that others in the future will port that image test folder.

core/src/avm2/globals/flash/display/bitmap_data.rs
iwannabethedev: Does the `TODO` look good now?

iwannabethedev:
Attempt to porting visual test to AVM1: https://gist.github.com/iwannabethedev/2be00ac00264659c62e07d3d7c33be67 .

iwannabethedev:
Flash Player AVM2, Ruffle AVM1, Ruffle AVM2:

actual-linux-Vulkan

Flash Player AVM1:

expected

iwannabethedev:
Many thanks to @n0samu for helping me with porting the visual tests to AVM1.

core/src/avm2/globals/flash/display/bitmap_data.rs
```diff
@@ -1369,3 +1374,73 @@ pub fn compare<'gc>(
...
+                .coerce_to_i32(activation)?,
+        );
+
+        let random_seed = args.get_i32(activation, 3)?;
+
+        let num_pixels = args.get_i32(activation, 4)?;
+        if num_pixels < 0 {
+            return Err(Error::AvmError(range_error(
+                activation,
+                &format!("Parameter numPixels must be a non-negative number; got {num_pixels}."),
```
Dinnerbone, line 1420: Nit: I expect this should be prefixed with `Error #2027: `?

Dinnerbone (approved):
Thank you for this! Minor nit but the code looks good

core/src/avm2/globals/flash/display/bitmap_data.rs
iwannabethedev: I think I added that at some point, but then I looked at other code like https://github.com/ruffle-rs/ruffle/blob/master/core/src/avm2/globals/flash/display/display_object_container.rs#L60-L64 , and I got in doubt about whether to keep it. I can see that for some of the `AvmError()`, the error code is include in the message, and in others of them, the error code is not included. Should I include it? I think that I will make minor changes to the tests if I include it, since then I can include the `e:Error` in the message instead of printing a custom message.

iwannabethedev: Because I think the official Flash Player does include the number in the string version of the error that is thrown.

Dinnerbone: I think originally we somewhat assumed the error code to be part of tostring, but we discovered that wasn't the case. Places like that should be updated to include it.

Strictly speaking we don't *need* to be 1:1 with Flash error messages, and in some cases we deliberately add more info or give more correct messages than sometimes outright wrong ones used by Flash. However, error codes are useful for humans I guess!

iwannabethedev: I will add it to the code and to the tests.

iwannabethedev: I have added it :) .

core/src/bitmap/operations.rs
```diff
@@ -1616,3 +1617,222 @@ pub fn set_pixels_from_byte_array<'gc>(
...
+        // random.
+        fn f(num: u32) -> u32 {
+            num * num + 1
+        }
+        let result_before_xor = (f(halfpiece2)) % (1 << feistel_halfpiece_size);
+
+        let new_halfpiece1 = halfpiece2;
+        let new_halfpiece2 = halfpiece1 ^ result_before_xor;
+
+        (new_halfpiece2 << feistel_halfpiece_size) | new_halfpiece1
```
jackandrew95, line 1694: this must be wrong, it should be "(new_halfpiece1 << feistel_halfpiece_size) | new_halfpiece2", else the least bits are never changed no matter how many times feistel is applied. but the current code might produce more random results because of feistel only being applied once. either keep the code as it currently is or change the random function and how many rounds of feistel that is applied.

PR #11110, opened 2023-05-18
avm2: Allow passing `null` to `Context3D.setProgram`

Aaron1011:
This clears the vertex and fragment shader programs (bringing the Context3D back to its initial state).


Lord-McSweeney:
Can you add a test for this?

Aaron1011:
@Lord-McSweeney I've added a `setProgram(null)` call to an existing test.

render/wgpu/src/context3d/current_pipeline.rs
```diff
@@ -133,24 +133,14 @@ impl CurrentPipeline {
             sampler_override: [None; 8],
         }
     }
-    pub fn set_vertex_shader(&mut self, shader: Rc<ShaderModuleAgal>) {
-        if let Some(old_shader) = &self.vertex_shader {
-            // If we change the shader, we need to recompile the pipeline.
```
iwannabethedev, line 138: (Nit: Would it be correct to guess that you removed this because the check was/is always true, and thus `self.dirty.set(true)` should always be called?).

iwannabethedev:
The changes look good to me, though I do not know this part of the codebase, so at least one other developer should review this.

render/wgpu/src/context3d/current_pipeline.rs
Aaron1011: I'm going to do a broader refactor of this code in follow-up pr, to perform better tracking and allow re-using a compiled shader when unrelated parts of the pipeline change. It wasn't worth it to add handling for the `None` case in this PR, only to remove it later.

iwannabethedev: That makes good sense. Would it be good to also add a `TODO` to the code? Like:

`// TODO: Better tracking and re-usage of compiled shaders is planned to be added in a future PR.`

?

Dinnerbone (approved):

render/wgpu/src/context3d/current_pipeline.rs
Aaron1011: I don't think it's really necessary

PR #11119, opened 2023-05-18
tests: Make the audio backend tick every frame in the runner

torokati44:
With no separate thread and sleeping to worry about, this should make the `computespectrum` test 100% reproducible, and not take 10 seconds to run.

Should prevent spurious failures such as:
https://github.com/ruffle-rs/ruffle/actions/runs/4962589152/jobs/8880836781?pr=11027

Somehow I wasn't aware that this "push" style execution model is also available in the `AudioBackend` trait when I originally made that test... :upside_down_face:

### DIFF
--- a/tests/tests/swfs/avm2/audio_computespectrum/test.toml
+++ b/tests/tests/swfs/avm2/audio_computespectrum/test.toml
@@ -1,5 +1,4 @@
 num_frames = 240
-sleep_to_meet_frame_rate = true
 
 [player_options]
 with_audio = true
--- a/tests/tests/util/runner.rs
+++ b/tests/tests/util/runner.rs
@@ -22,26 +22,18 @@ use std::time::Duration;
 
 pub struct TestAudioBackend {
     mixer: AudioMixer,
+    buffer: Vec<f32>,
 }
 
 impl TestAudioBackend {
+    const NUM_CHANNELS: u8 = 2;
+    const SAMPLE_RATE: u32 = 44100;
+
     pub fn new() -> Self {
-        const NUM_CHANNELS: u8 = 2;
-        const SAMPLE_RATE: u32 = 44100;
-        const BUFFER_SIZE: usize = 1024;
-
-        let mixer = AudioMixer::new(NUM_CHANNELS, SAMPLE_RATE);
-        let mixer_proxy = mixer.proxy();
-
-        std::thread::spawn(move || loop {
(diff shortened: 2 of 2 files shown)


tests/tests/util/runner.rs
```diff
@@ -22,33 +22,35 @@ use std::time::Duration;
...
 }
 
 impl AudioBackend for TestAudioBackend {
     impl_audio_mixer_backend!(mixer);
     fn play(&mut self) {}
     fn pause(&mut self) {}
+
+    fn set_frame_rate(&mut self, frame_rate: f64) {
+        let new_buffer_size =
+            ((Self::NUM_CHANNELS as u32 * Self::SAMPLE_RATE) as f64 / frame_rate).round() as usize;
```
iwannabethedev, line 47: Nit: I do not understand this part of the codebase, but `Self::SAMPLE_RATE` seems a bit big. I do not know what kind of values that `frame_rate` can have, and what the buffer size ought to be, but, should the buffer size increase proportionally with `Self::SAMPLE_RATE`?

iwannabethedev:
The changes look overall good to me, though I do not understand this part of the codebase.

tests/tests/util/runner.rs
torokati44: 44100 is a fairly typical audio sample rate. It's almost a universal default. That's what CD audio used, and what many sound cards still use.
Frame_rate is typically somewhere around 24..60. The buffer is just there so we can tell the mixer to advance the audio stream, we don't do anything with it's contents in this test backend. And since a buffer in this setting holds all the audio data between consecutive frames, yes it should scale with the sample rate.

iwannabethedev: That makes sense, thank you.

tests/tests/swfs/avm2/audio_computespectrum/test.toml
```diff
@@ -1,5 +1,5 @@
 num_frames = 240
-sleep_to_meet_frame_rate = true
+sleep_to_meet_frame_rate = false
```
Dinnerbone, line 2: nit: You can just remove this line now

Dinnerbone (approved):

tests/tests/swfs/avm2/audio_computespectrum/test.toml
torokati44: Ah, true, thanks, I will soon!

torokati44: Okay, done!

PR #11120, opened 2023-05-19
Improve visual testing somewhat

torokati44:
See the commit messages.


torokati44:
To be honest, I don't even know when/if the expected image could have any translatency... So perhaps the alpha diff image makes no sense...? Or at least it could be saved only if it has any non-zero pixels?

n0samu:
Saving the alpha diff image only if there's an actual difference sounds good!

Dinnerbone (approved):

PR #11123, opened 2023-05-19
ci: Remove linter step

Bale001:
I think that while this was a good experiment, the trouble caused by this outweighs the good. It was good in theory, but had the following problems:
1. Majority of lints made no sense at all (like weird naming conventions)
2. The linter was clearly not designed to be used on something like playerglobals. It was instead clearly meant for a regular flash project.
3. Getting around the linter was too ugly. Those "NOPMD" comments are annoying and make the code uglier, and removing the lint entirely from the ruleset XML file takes too much effort.

In my opinion, what would be more useful to use would be a **formatter**, not a linter. This PR only removes the linter from the CI, I'm not sure if its a good idea to remove it entirely yet.


Dinnerbone (approved):
I totally agree

adrian17 (approved):
Do you wanna also remove the `NOPMD` comments?

relrelb:
> Do you wanna also remove the `NOPMD` comments?

And maybe also the linter code itself?

Aaron1011 (approved):

Bale001:
Alright, i've removed the linter code entirely and also all NOPMD comments

PR #11125, opened 2023-05-19
ci: Use GitHub CLI instead of deprecated actions

### LINKED ISSUE #11105: Nightly builds may stop working after May 31
As @danielhjacobs has explained multiple times on Discord, [GitHub is disabling the `save-state` and `set-output` workflow commands on May 31](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/). These commands are used by the unmaintained [create-release](https://github.com/actions/create-release) and [upload-release-asset](https://github.com/actions/upload-release-asset) actions that we use. We will need to switch to an equivalent maintained action, with the likely best choice being [action-gh-release](https://github.com/softprops/action-gh-release). This will likely require us to rewrite most of our [nightly release workflow](https://github.com/ruffle-rs/ruffle/blob/master/.github/workflows/release_nightly.yml). So far no one on the Ruffle team has indicated that they have the time and ability to do this, so any help would be greatly appreciated!

n0samu:
Resolves #11105. Tested on my branch and seems to be working perfectly: https://github.com/n0samu/ruffle/releases

### DIFF
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -20,7 +20,7 @@ jobs:
       tag_name: nightly-${{ steps.current_time_dashes.outputs.formattedTime }}
 
     # Only run the scheduled workflows on the main repo.
-    if: github.repository == 'ruffle-rs/ruffle'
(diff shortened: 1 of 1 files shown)


adrian17 (approved):
Nice :) IMO let's get it in and see how it behaves.

PR #11149, opened 2023-05-21
tests: Unconditionally save 'actual.txt' with Ruffle output

Aaron1011:
When a test has large amounts of output, the diff printed by Ruffle can be difficult to understand. Also, it can be useful to see the real floating point values for a passing test with approximate floating point comparison enabled.

I've modified the test runner to save the actual Ruffle trace output to 'actual.txt' (in the same directory as 'output.txt'). For simplicity, this is done regardless of whether or not the test passes or fails.

To prevent messy 'git status' output and accidental commits, I've added all of the test-generated files ('actual.txt', 'different-*.png', and 'actual-*png') to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,7 @@ target/
 
 # Flash recovery files
 RECOVER_*.fla
+
+tests/**/actual.txt
+tests/**/actual-*.png
+tests/**/difference-*.png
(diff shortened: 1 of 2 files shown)


adrian17:
This is a little bit awkward. IIRC other create a temporary directory somewhere in `/tmp/` and put the failure info there instead of in the working tree.

PR #11156, opened 2023-05-22
core: implement screen resolution for capabilities

AllinolCP:


core/src/avm1/globals/system.rs
```diff
@@ -348,7 +339,22 @@ impl SystemProperties {
         percent_encoding::utf8_percent_encode(s, percent_encoding::NON_ALPHANUMERIC).to_string()
     }
 
-    pub fn get_server_string(&self, avm: &mut Avm1) -> String {
+    pub fn screen_resolution(&self, context: &UpdateContext) -> (u32, u32, f64) {
```
iwannabethedev, line 342: (Nit: Would a documentation comment indicating which unit it is measured in be helpful?

Like, in case this is "pixels" (a guess on my part, I do not know what the unit is, and "pixels" is not necessarily all that accurate):

```rust
    /// Screen resolution, measured in pixels.
    pub fn screen_resolution(&self, context: &UpdateContext) -> (u32, u32, f64) {
```

?)

iwannabethedev: (Nit: I imagine that, in a future PR (not in this PR), it might be nice to define some basic type aliases ( https://doc.rust-lang.org/reference/items/type-aliases.html ) for different types like this, to decrease need for documentation and for making the meaning of the types clearer. But this would again be in a future PR from someone).

core/src/avm2/globals/flash/system/capabilities.rs
```diff
@@ -31,3 +31,43 @@ pub fn get_player_type<'gc>(
...
+}
+
+/// Implements `flash.system.Capabilities.pixelAspectRatio`
+pub fn get_pixel_aspect_ratio<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    _this: Option<Object<'gc>>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let viewport_dimensions = activation.context.renderer.viewport_dimensions();
+    Ok(viewport_dimensions.scale_factor.into())
```
iwannabethedev, line 62: I am not certain that this is correct, but I also do not understand this part of the codebase.

https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/system/Capabilities.html#pixelAspectRatio

https://en.wikipedia.org/wiki/Pixel_aspect_ratio

Is `scale_factor` really the same as the "pixel aspect ratio"/`pixelAspectRatio`? This is the documentation I found for `scale_factor`:

https://github.com/ruffle-rs/ruffle/blob/master/render/src/backend.rs#L457-L459 :

```rust
    /// The scale factor of the containing viewport from standard-size pixels
    /// to device-scale pixels.
    pub scale_factor: f64,
```

I also found this page, which may or may not be relevant and helpful: https://help.adobe.com/en_US/as3/dev/WS5b3ccc516d4fbf351e63e3d118a9b90204-7cc6.html .

iwannabethedev:
(Aside: I have not reviewed the whole PR, but, I feel like the existing codebase does not have a lot of documentation currently for unit types related to these changes. Of course, documentation can go out of date. I do not feel that it would be fair to request it of this PR, but I feel that if one or more people looked at making a PR (I am not asking you) that added "unit measurement types" (like seen in this Rust crate: https://docs.rs/typeables/latest/typeables/ ), like, type aliases for unit measurements and used it in the relevant parts of the codebase, then PRs like this would be easier to get correct and also somewhat "automatically" become documented through the use of unit measurement types. But this is not directly relevant for this PR).

core/src/avm2/globals/flash/system/capabilities.rs
AllinolCP: since scale_factor is devicePixelRatio in web I assumed that this was pixelAspectRatio in flash

iwannabethedev: I think `devicePixelRatio` comes from: https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio .

Also, I believe that `devicePixelRatio` describes some ratio between "physical" and "visual" pixels.

On the other hand, the "aspect ratio" in `pixelAspectRatio` describes (I believe) the relationship between width and height of some entity, in this case I believe the "aspect ratio" between the width and the height of a single pixel.

So, I would believe that these are two different abstractions.

AllinolCP: I see, `pixelAspectRatio` seems to be referring to the actual pixel ratio itself

iwannabethedev: I would believe that a decent first implementation of `devicePixelRatio` would simply be `1`, since I do not know if Ruffle even supports any other ratio than `1`, and the capability in https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/system/Capabilities.html#pixelAspectRatio is immutable. But, it would be best if someone else other than me weighed in. Does anyone have any thoughts on this?

I would propose that you simply change it to that. See also my comment on the Ruffle forum.

iwannabethedev:
While I do not know this part of the codebase or the topic well, I think the changes look overall good.

Requests:
- I would like to request that the ActionScript sources are attached to each test, for instance as a file named `Test.as` in each test folder, such that the source code can be viewed (or possibly compiled using for instance `mxmlc`) without opening the `*.fla` files or using a SWF editor like JPEXS FFDEC.

Suggestion:
- I wonder if an additional test case that changes in the `test.toml` file the `scale_factor` to for instance 2 or 3, such that it can be tested that the DPI does scale with `scale_factor`. This might be a bit bothersome, since you can only have one `test.toml` file per test folder.

Once the request with the `Test.as` files being included has been followed, I think a main developer should review the changes, and if that goes alright, that the PR might be ready to be merged.

core/src/avm2/globals/flash/system/capabilities.rs
```diff
@@ -31,3 +31,45 @@ pub fn get_player_type<'gc>(
...
+
+/// Implements `flash.system.Capabilities.screenResolutionX`
+pub fn get_screen_resolution_x<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    _this: Option<Object<'gc>>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let viewport_dimensions = activation.context.renderer.viewport_dimensions();
+    // Viewport size is adjusted for HiDPI.
+    let adjusted_width = f64::from(viewport_dimensions.width) / viewport_dimensions.scale_factor;
```
iwannabethedev, line 43: Is it correct that the implementation for AVM1 and AVM2 seem to differ here?

iwannabethedev:
Nit: Did you include the `Test.as` file for the AVM1 test?

iwannabethedev:
I do not know this part of the codebase well, but I think the changes look good, and I think after a main developer has reviewed the changes and the PR has been rebased, that it might be ready to be merged.

You can rebase in two main ways that I know of:
- Locally: For instance by being on your branch, doing `git fetch [name-of-remote-Ruffle]` and then doing `git rebase [name-of-remote-Ruffle]/master` (`name-of-remote-Ruffle` should point to `https://github.com/ruffle-rs/ruffle` , not your own fork of Ruffle).
- Through Github: Probably easier, see this screenshot, where you click "Update with rebase" highlighted in blue): update with rebase .

Dinnerbone (approved):

Korne127:
Sorry for this late comment, but I just noticed this PR / commit and from what I can tell, this doesn't implement the screen resolution correctly.

In Flash (both AVM1/AVM2), `System.capabilities.screenResolutionX` / `.screenResolutionX` are returning the resolution of the whole screen. E.g. when tracing it on my computer it returns `1440` / `900`.
However, this Ruffle implementation (at least on desktop) doesn't return the actual screen resolution but rather the resolution of the Ruffle window. When opening the same SWF, Ruffle returns `550` / `400` (which is the stage size).

PR #11182, opened 2023-05-22
Upgrade to new `gc-arena` API

moulins:
Upgrade `gc-arena` to latest `master`, which brings some breaking API changes:
- `MutationContext<'gc, '_>` was renamed to `&Context<'gc>`;
- `CollectionContext<'_>` was renamed to `&Collection`;
- `Gc::allocate` was renamed to `Gc::new`;
- `GcCell` was removed in favor of a more flexible API (TL;DR, the equivalent of `GcCell` is now `Gc<'gc, RefLock<T>>` a.k.a. `GcRefLock<'gc, T>`).

To accommodate these changes without extreme code churn, I've added a (hopefully temporary!) `ruffle_gc_arena` façade crate that re-adds the old APIs we're still using; namely:
- a `MutationContext` typedef;
- a reimplementation of `GcCell` on top of `GcRefLock`.

The goal is to slowly migrate these to their replacements, at which point the façade crate will be removed.

------

Additionally, this PR implements `Deref<Target=Mutation<'gc>>` for the various `Context` structs, which will allow using `&activation.context` as a shorthand for `activation.context.gc_context` in methods requiring a `Mutation` argument.


moulins:
Small update: I've decided to implement `Deref<Target=Mutation<'gc>>` for AVM1&2 `Activation` types too.  
Being able to write `Gc::new(activation, ...)` instead of `Gc::new(activation.context.gc_context, ...)` is really nice! (though sometimes the borrow checker doesn't like it, so one needs to fallback to the longer version)

moulins:
To give some idea of how code using the new mutability API looks,  [here](https://github.com/moulins/ruffle/compare/upgrade-gc-arena...gc-mutability-context3d) is a separate branch which replaces a few `GcCell`s in the codebase by the new `lock` types (the commits are fully independent).

Dinnerbone:
> (though sometimes the borrow checker doesn't like it, so one needs to fallback to the longer version)

I'm a bit scared of this. If we rely on the shortcut and then someone comes to make a change and suddenly the shortcut doesn't work, but they wouldn't know it's the shortcuts fault - that's going to cause a bit of frustration.

moulins:
> I'm a bit scared of this. If we rely on the shortcut and then someone comes to make a change and suddenly the shortcut doesn't work, but they wouldn't know it's the shortcuts fault - that's going to cause a bit of frustration.

A way to reduce surprise could be to mention the `Deref` impl in the type documentation, what do you think?  E.g.:
```rs
/// ...
/// As a convenience, this type can be deref-coerced to `&Mutation<'gc>`, but note that explicitely
/// writing `context.gc_context` can be sometimes necessary to satisfy the borrow checker.
pub struct UpdateContext<'a, 'gc> {
  /// ...
```

Dinnerbone (approved):
Fine with me!

Aaron1011 (approved):

adrian17 (approved):

Issue #11053, opened 2023-05-14
Gibbets 4: Illegal write to read-only property builtInItems on flash.ui.ContextMenu


MartySVK:
### Describe the bug

The game is stuck on loading with the error:

``
ERROR ruffle_core::display_object::movie_clip: Got "ReferenceError: Error #1074: Illegal write to read-only property builtInItems on flash.ui.ContextMenu." when constructing AVM2 side of movie clip of type Main::CurrentStateManager
``

SWF - https://data.gameflare.com/games/5092/x6HtXDrXIxUZ7r.swf
Game - https://www.gameflare.com/online-game/gibbets-4/

**Gibbets: Santa in Trouble** and **Gibbets 3** have the same problem

### Expected behavior

progress loading

### Affected platform

Desktop app

### Operating system

Windows 10

0xSharkyy:
this might be an issue with the games .swf file

Aaron1011:
We just need to define a setter for `builtInItems`

PR #11187, opened 2023-05-23
avm2: Stub ContextMenu.builtInItems setter

MartySVK:
Added setter for ContextMenu.builtInItems stub
fixes https://github.com/ruffle-rs/ruffle/issues/11053

### DIFF
--- a/core/src/avm2/globals/flash/ui/ContextMenu.as
+++ b/core/src/avm2/globals/flash/ui/ContextMenu.as
@@ -1,6 +1,8 @@
 package flash.ui
 {
     import flash.display.NativeMenu;
+    import __ruffle__.stub_getter;
+    import __ruffle__.stub_setter;
 
     public final class ContextMenu extends NativeMenu
     {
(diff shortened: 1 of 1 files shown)


core/src/avm2/globals/flash/ui/ContextMenu.as
```diff
@@ -17,9 +19,15 @@ package flash.ui
         private var _builtInItems: ContextMenuBuiltInItems = new ContextMenuBuiltInItems();
 
         public function get builtInItems(): ContextMenuBuiltInItems {
+            stub_getter("flash.ui.ContextMenu", "builtInItems");
```
Lord-McSweeney, line 22: I don't think you need to add a stub getter since this part is working correctly (returning the value of built-in-items).

MartySVK: I actually tried to keep the same format as in other parts of the code like here - https://github.com/ruffle-rs/ruffle/pull/10656/files
but I can remove it if it is useless

Lord-McSweeney: As done in that PR, the `stub_getter` is unnecessary (that PR doesn't add a `stub_getter`).

Aaron1011 (approved):

Issue #9940, opened 2023-03-09
Bloons TD 3 - Monkey storm freezes the game


danielhjacobs:
### Describe the bug

I used the monkey storm provided by the monkey beacon on round 49, at a time when all bloons were on screen so using it should have ended the round. Instead, the game "froze", with no option to continue to the next round and no bloons moving (I believe the bloons that appeared on screen are a visual bug, as the visibly frozen bloons are nowhere near the ice tower), though the beacon and ice tower were still doing their visual effects.

![monkey_storm](https://user-images.githubusercontent.com/29206584/224110145-cd73411a-b96a-4b61-9f08-2d27309ad502.png)

### Expected behavior

A monkey storm is meant to destroy all bloons on screen at the cost of $850. If that would result in the round ending, the round should end. No bloons should remain motionless on screen.

### Affected platform

Self-hosted version

### Operating system

Windows 11

### Browser

Chrome 110

### Additional information

Version at https://www.kongregate.com/games/Ninjakiwi/bloons-tower-defense-3

danielhjacobs:
Need to play the game to be sure, but supposedly fixed by #11195

PR #11195, opened 2023-05-23
core: Use clone-on-write behavior for render list iteration

Aaron1011:
When we iterate over a render list (in order to call `enter_frame`, `construct_frame`, etc.) we need to be sure that we iterate over all of the original `DisplayObjects` in the list, even if the list is modified during iteration (e.g. some ActionScript code calls `parent.addChildAt`).

Previously, `RenderIter` would repeatedly call `child_by_index` on the original `DisplayObjectContainer`, up to the original child count. If new DisplayObjects were inserted into the list during iteration, we could miss some of the original DisplayObjects in the list (as they are now at a greater position in the list).

To solve this, we now store the render list as an `Rc<Vec>`, and use `Rc::make_mut` to modify it. See the comments for more details.


Lord-McSweeney:
As I understand this will affect AVM1 too, is there a way to test that?

moulins:
> As I understand this will affect AVM1 too, is there a way to test that?

I don't think so. ActionScript 2 code cannot directly access the render list; `MovieClip`s can only be accessed by name and/or by depth.

danielhjacobs:
Should fix #9940 (not yet tested by me but Aaron said it was fixed on Discord).

adrian17 (approved):

Issue #11161, opened 2023-05-22
Nambers: Visual glitch


n0samu:
### Describe the bug

Link to SWF: https://static.arcadespot.com/wp-content/uploads/2015/10/nambers.swf

In [Nambers](https://arcadespot.com/game/nambers/), there is a visual glitch after selecting the numbers.

![nambers_ruffle_glitch](https://github.com/ruffle-rs/ruffle/assets/71368227/d31304c5-f6bf-4c4a-b5d8-f419c13ff34f)

### Expected behavior

Here is what it's supposed to look like.

![nambers_flashplayer_expected](https://github.com/ruffle-rs/ruffle/assets/71368227/73450337-718c-44b0-b05a-d3f23f9c8612)

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

Tested with #11080, which fixes previous issues with this game.

Aaron1011:
Fixed in https://github.com/ruffle-rs/ruffle/pull/11198

PR #11198, opened 2023-05-24
wgpu: Implement Context3D.setScissorRectangle

Aaron1011:


n0samu:
Can confirm that this fixes #11161, thank you!

render/wgpu/src/context3d/mod.rs
```diff
@@ -286,6 +290,14 @@ impl WgpuContext3D {
                 .as_ref()
                 .expect("Missing compiled pipeline"),
         );
+        if let Some(rect) = &self.scissor_rectangle {
+            pass.set_scissor_rect(
```
Dinnerbone, line 294: wgpu will panic if these bounds exceed the render attachment. I don't think we're checking for this, so what does Flash do?

Aaron1011: Flash throws an error during `drawTriangles`, but only if `Context3D.enableErrorChecking` is enabled. For now, I think we could just skip setting the scissor rect on the wgpu side, and log a warning.

Aaron1011: I've added in an error log when the rectangle is too large.

render/wgpu/src/context3d/mod.rs
```diff
@@ -286,6 +295,29 @@ impl WgpuContext3D {
                 .as_ref()
                 .expect("Missing compiled pipeline"),
         );
+        if let Some(rect) = &self.scissor_rectangle {
+            let current_size = self.current_texture_size.unwrap();
+            if rect.x_min.to_pixels() as u32 > current_size.width
+                || rect.x_max.to_pixels() as u32 > current_size.width
+                || rect.y_min.to_pixels() as u32 > current_size.height
+                || rect.y_max.to_pixels() as u32 > current_size.height
```
Lord-McSweeney, line 303: Should this check also happen for negative x/y?

Aaron1011: I've adjusted the check

Dinnerbone (approved):

PR #11202, opened 2023-05-24
avm2: Add legacy handling for configureBackBuffer with SWF version < 30

Aaron1011:
With SWF version < 30, Context3D.configureBackBuffer throws an error with a less informative message when the width/height are out of range. Additionally, it seems to special case the case width=0, height=0, antiAlias=0. enableDepthAndStencil=false, and does *not* throw an error. This is relied on by Sniper Team.


core/src/avm2/globals/flash/display3D/context_3d.rs
```diff
@@ -57,25 +57,39 @@ pub fn configure_back_buffer<'gc>(
     if let Some(mut context) = this.and_then(|this| this.as_context_3d()) {
         let width = args.get_u32(activation, 0)?;
         let height = args.get_u32(activation, 1)?;
+        let anti_alias = args.get_u32(activation, 2)?;
+        let enable_depth_and_stencil = args.get(3).unwrap_or(&Value::Undefined).coerce_to_boolean();
```
Dinnerbone, line 61:
```suggestion
        let enable_depth_and_stencil = args.get_bool(3);
```

adrian17 (approved):

PR #11206, opened 2023-05-25
web: Fix minor issues with the context menu

### LINKED ISSUE #11130: The mouse pointer position should frozen after right click.
### Describe the bug

https://github.com/ruffle-rs/ruffle/assets/75534590/498eea30-8766-4109-8802-ec5d26679582

### Expected behavior

https://github.com/ruffle-rs/ruffle/assets/75534590/57c6e6ab-8af5-437a-8a4c-9caaf7e6547f

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

http://eye.swfchan.com/flash.asp?id=5682&n=The+Maze.swf

Here's Html5 version can frozen after right click.
https://www.playscarymazegame.net/play-scary-maze-game/

Toad06:
- Adds a div element as an overlay between the canvas and the context menu to prevent interaction with the file when the context menu is open. Fixes #11130. It's worth mentioning that the behavior may not match Flash when you click back to the file (e.g. if you click a button, Flash hides the context menu and also propagates the event to the button). This is probably not too important.

- No longer opens Ruffle's context menu when the file hasn't started yet or when the save manager is open.

- No longer opens Ruffle's context menu when a panic occurs, enables the default context menu instead (allows opening links in new tab, etc.).

- Replaces the "click" event by "pointerup" for context menu items on desktop (pressing the mouse left button, moving the mouse and releasing on any item previously caused the context menu to hide instead of running the action of the cliked item).


n0samu:
Thank you for this - the first and third items sound great! 

> No longer opens Ruffle's context menu when the file hasn't started yet or when the save manager is open.

- Not showing the Ruffle context menu while the save manager is open is a good thing, but you should move the check before the code that adds the listener to disable the context menu. I'm thinking you should merge the check with the `this.panicked` check above - I understand that it might seem weird to allow the browser's default context menu in this situation, but right now it's suppressed even after selecting text in the save manager (to copy it, for example), which to me seems a little bit worse. But I'm not wholly opposed to moving the check right after the preventDefault() call if that's your preference. Or alternatively, you could try to make the browser's default context menu only appear when the user right-clicks within the save manager, but I'm not sure how to do that.
- `isPlaying` does not actually mean "has not started yet" - the Ruffle player exposes a `pause()` method that can be called at any time and changes `isPlaying` to false. So I think you should remove this condition. (Also, what if I just want to check the Ruffle version a site is using without playing the content on the page, shouldn't I be able to do that?)

> Replaces the "click" event by "pointerup" for context menu items on desktop (pressing the mouse left button, moving the mouse and releasing on any item previously caused the context menu to hide instead of running the action of the cliked item).

I'm really worried about this one because we carefully chose these events to work on desktop, Android and iOS, and the setup is very fragile. We'd need to test this thoroughly and I'm not sure it's worth the effort for such a minor improvement.

Toad06:
Thanks for the review.

> you should move the check before the code that adds the listener to disable the context menu. I'm thinking you should merge the check with the `this.panicked` check above 

Done!

> alternatively, you could try to make the browser's default context menu only appear when the user right-clicks within the save manager, but I'm not sure how to do that.

Not sure either, I think `event.originalTarget` would allow us to do that but it's not standard and only exists in Firefox.

> `isPlaying` does not actually mean "has not started yet" - the Ruffle player exposes a `pause()` method that can be called at any time and changes `isPlaying` to false. So I think you should remove this condition. (Also, what if I just want to check the Ruffle version a site is using without playing the content on the page, shouldn't I be able to do that?)

Right but even when the player is paused, I don't think it makes a lot of sense to show things like "Rewind", "Back", etc. I do agree however that it is relevant for some other elements to be shown (the ones that are specific to Ruffle), so I moved the check to the `contextMenuItems` function.

> I'm really worried about this one because we carefully chose these events to work on desktop, Android and iOS, and the setup is very fragile. We'd need to test this thoroughly and I'm not sure it's worth the effort for such a minor improvement.

I reverted this as you were concerned about it, but really there should be a comment in code explaining why this is necessary because the reason doesn't look obvious at all.

n0samu:
For the record, @danielhjacobs looked back at the original conversation on Discord, and it seems like using the click event instead of pointerup was only necessary on Android when we had a context menu item to open the virtual keyboard. (That functionality was never merged since it was superseded by #9461.) So we should actually be fine to switch to the click event if you still want to.

relrelb (approved):

Toad06:
> For the record, @danielhjacobs looked back at the original conversation on Discord, and it seems like using the click event instead of pointerup was only necessary on Android when we had a context menu item to open the virtual keyboard. (That functionality was never merged since it was superseded by #9461.) So we should actually be fine to switch to the pointerup event if you still want to.

This is good to know, thanks. I still think it would be nice to change this at some point, but after testing it would probably be best to do this when we find a way to disable the context menu from opening when right-clicking on... the context menu (another minor issue).

Otherwise this could affect the behavior on mobile devices that support `oncontextmenu`: holding one of the items in the context menu for some time before releasing would run the event for the selected item, but also immediately runs the `contextmenu` event; currently, this doesn't run the event for the selected item because we use the "click" event.

This wouldn't be a terrible change to do this now of course, but I'd prefer to keep things as-is for now.

n0samu:
Makes sense, we'll keep it as-is for now then. Thanks!

PR #11217, opened 2023-05-25
wgpu: Cache compiled naga-agal shader module

Aaron1011:
We use an `lru::LruCache` inside `ShaderModuleAgal`. This automatically gives us the proper garbage-collection behavior (when the Flash Program3D instance is garbage collected, we'll drop the `ShaderModuleAgal` and the cache).

The cache is keyed on the data needed to compile the shader (vertex attributes and sampler overrides). This lets us avoid shader recompilations when a Stage3D program repeatedly uses the same Program3D with different sampler overrides / vertex attribute formats.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2573,6 +2573,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "lru"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


Dinnerbone (approved):
I somewhat suspect we'll want a temporal element to the cache rather than a fixed small size, but this is good for now!

PR #11228, opened 2023-05-26
avm1: Correct `ActionCloneSprite` and `duplicateMovieClip()`

relrelb:
* `ActionCloneSprite` coerces `depth` before `target`.
* `duplicateMovieClip()` coerces `name` before `depth`.
* `duplicateMovieClip()` returns `undefined` on SWF<6.


core/src/avm1/globals/movie_clip.rs
```diff
@@ -761,86 +761,90 @@ fn duplicate_movie_clip<'gc>(
...
             );
             return Ok(Value::Undefined);
         }
     };
-    let init_object = args.get(2);
+    let depth = match args.get(1) {
+        Some(depth) => depth.coerce_to_i32(activation)?,
+        None => 0,
+    };
+    let init_object = args.get(2).map(|v| v.coerce_to_object(activation));
```
Toad06, line 778: Nit: Maybe it could be worth to add a comment mentioning that this argument (`initObject`) is not version-gated. The docs say it is supported only in FP6 and later, but this doesn't appear to be true. I suppose this should refer to the player version and not the SWF version, but the documentation also uses the same wording when talking about the return value (and you found out it was indeed version-gated).

https://open-flash.github.io/mirrors/as2-language-reference/MovieClip.html#duplicateMovieClip()

Toad06 (approved):
LGTM, just a nit. I'll rebase #11196 once this PR is merged.

PR #11233, opened 2023-05-27
docs: Add detailed instructions for running tests

n0samu:
This clarifies some things that confused me when I was starting out, and should help new contributors.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -157,13 +157,29 @@ Specific warnings and clippy lints can be allowed when appropriate using attribu
 
 ## Test Guidelines
 
-Most tests are SWF-based, with the SWFs stored in `tests/tests/swfs/`. They are configured in `tests/tests/regression_tests.rs`. To add a new test, create an `.swf` that runs `trace()` statements. You can do this in several ways, listed below.
+Most tests are SWF-based, with the SWFs stored in `tests/tests/swfs/`. They are configured in `tests/tests/regression_tests.rs`. Most test SWFs include `trace()` statements, the output of which is compared to the expected output from Flash Player. To view the output from Flash Player, first download the [debug Flash Player](https://web.archive.org/web/20220401020702/https://www.adobe.com/support/flashplayer/debug_downloads.html) for your platform. Then create a plain text file called `mm.cfg` with the following contents:
+```
+ErrorReportingEnable=1
+TraceOutputFileEnable=1
+```
+Place the file at the following location:
+* Windows: `%USERPROFILE%`
+* MacOS: `/Library/Application Support/Macromedia/`
+* Linux: `$HOME`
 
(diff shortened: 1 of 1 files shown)


Dinnerbone:
I like it but I'd like to keep the bit about how to run all tests in addition to just the specific one they're adding

CONTRIBUTING.md
```diff
@@ -157,13 +157,29 @@ Specific warnings and clippy lints can be allowed when appropriate using attribu
 
 ## Test Guidelines
 
-Most tests are SWF-based, with the SWFs stored in `tests/tests/swfs/`. They are configured in `tests/tests/regression_tests.rs`. To add a new test, create an `.swf` that runs `trace()` statements. You can do this in several ways, listed below.
+Most tests are SWF-based, with the SWFs stored in `tests/tests/swfs/`. They are configured in `tests/tests/regression_tests.rs`. Most test SWFs include `trace()` statements, the output of which is compared to the expected output from Flash Player. To view the output from Flash Player, first download the [debug Flash Player](http://web.archive.org/web/20220401020702/https://www.adobe.com/support/flashplayer/debug_downloads.html) for your platform. Then create a plain text file called `mm.cfg` with the following contents:
```
relrelb, line 160: Can HTTPS link can be used here?

CONTRIBUTING.md
```diff
@@ -157,13 +157,29 @@ Specific warnings and clippy lints can be allowed when appropriate using attribu
...
+```
+ErrorReportingEnable=1
+TraceOutputFileEnable=1
+```
+Place the file at the following location:
+* Windows: `%USERPROFILE%`
```
relrelb, line 166: I see it clearer to include the filename itself to form a full path. Like this:
```md
* Windows: `%USERPROFILE%\mm.cfg`
* MacOS: `/Library/Application Support/Macromedia/mm.cfg`
...
```
(Also applies to `flashlog.txt` below)

CONTRIBUTING.md
n0samu: Yep, good catch

CONTRIBUTING.md
n0samu: It kind of depends on how exactly you expect users to be following the instructions. When writing the instructions I was testing them by pasting the folder path into the Windows Explorer path bar, and into the "Go to folder" dialog in MacOS. So that's why I wrote them this way.

n0samu:
> I like it but I'd like to keep the bit about how to run all tests in addition to just the specific one they're adding

There never was anything about that as far as I understand? But ok, I added a mention of running `cargo test` by itself to run all the tests. (That's what you mean right?)

n0samu:
Sorry, I wasn't thinking of the tests in Ruffle's other workspaces like swf and web. I've added back `--all` to the instructions.

Dinnerbone (approved):

adrian17 (approved):

PR #11241, opened 2023-05-27
core: Store Color as an u32 instead of a i32

Aaron1011:
This is a packed ARGB value, so it doesn't make sense for it to be signed.

### DIFF
--- a/core/src/avm1/globals/bitmap_data.rs
+++ b/core/src/avm1/globals/bitmap_data.rs
@@ -97,8 +97,8 @@ fn constructor<'gc>(
         None => true,
     };
     let fill_color = match args.get(3) {
-        Some(fill_color) => fill_color.coerce_to_i32(activation)?,
-        None => -1,
+        Some(fill_color) => fill_color.coerce_to_u32(activation)?,
+        None => u32::MAX,
     };
 
     if !is_size_valid(activation.swf_version(), width, height) {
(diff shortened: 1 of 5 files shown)


torokati44:
I think the type names in the commit message and PR title are the wrong wy around.

Aaron1011:
@torokati44 Thanks, fixed

core/src/avm1/globals/bitmap_data.rs
```diff
@@ -97,8 +97,8 @@ fn constructor<'gc>(
         None => true,
     };
     let fill_color = match args.get(3) {
-        Some(fill_color) => fill_color.coerce_to_i32(activation)?,
-        None => -1,
+        Some(fill_color) => fill_color.coerce_to_u32(activation)?,
+        None => -1i32 as u32,
```
adrian17, line 101: `u32::MAX`?

core/src/avm1/globals/bitmap_data.rs
```diff
@@ -193,7 +193,8 @@ fn get_pixel<'gc>(
             if let (Some(x_val), Some(y_val)) = (args.get(0), args.get(1)) {
                 let x = x_val.coerce_to_u32(activation)?;
                 let y = y_val.coerce_to_u32(activation)?;
-                let col = operations::get_pixel(bitmap_data, x, y);
+                // AVM1 returns a signed int, so we need to convert it.
```
adrian17, line 196: can we have some test for the negative values here?

adrian17 (approved):

core/src/avm1/globals/bitmap_data.rs
Aaron1011: There are a lot of existing avm1 tests that fail without this.

PR #11242, opened 2023-05-27
avm2: Allow replacing and deleting XML attributes

Aaron1011:


core/src/avm2/object/xml_object.rs
```diff
@@ -303,41 +303,45 @@ impl<'gc> TObject<'gc> for XmlObject<'gc> {
...
         let mc = activation.context.gc_context;
-        let write = self.0.write(mc);
-        let mut kind = write.node.kind_mut(mc);
-        let E4XNodeKind::Element {
-            attributes, children
-        } = &mut *kind else {
-            return Ok(());
-        };
 
         #[allow(clippy::collapsible_if)]
```
adrian17, line 307: is this `allow` still needed?

adrian17 (approved):

core/src/avm2/object/xml_object.rs
Aaron1011: It's not - I've removed it

PR #11243, opened 2023-05-27
desktop: Add suspend/resume buttons

Dinnerbone:
Will come in handy with (not-yet-made) debugging overlays, to prevent values shifting all the time.
Also it's cool to just pause things.

I'm thinking a future option might be "pause in background" or something similar.

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -1,6 +1,6 @@
 use crate::cli::Opt;
 use crate::custom_event::RuffleEvent;
-use crate::gui::{GuiController, MovieView, MENU_HEIGHT};
+use crate::gui::{GuiController, MENU_HEIGHT};
 use crate::player::PlayerController;
 use crate::util::{
     get_screen_size, parse_url, pick_file, winit_key_to_char, winit_to_ruffle_key_code,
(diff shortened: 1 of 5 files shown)


desktop/src/app.rs
```diff
@@ -218,64 +214,74 @@ impl App {
...
+                                    let delta = match delta {
+                                        MouseScrollDelta::LineDelta(_, dy) => {
+                                            MouseWheelDelta::Lines(dy.into())
+                                        }
+                                        MouseScrollDelta::PixelDelta(pos) => {
+                                            MouseWheelDelta::Pixels(pos.y)
+                                        }
+                                    };
+                                    let event = PlayerEvent::MouseWheel { delta };
+                                    player.handle_event(event);
```
adrian17, line 264: could we instead wrap handle_event in an is-playing-checking function?

desktop/src/app.rs
```diff
@@ -137,15 +136,10 @@ impl App {
...
-                            let renderer = player
-                                .renderer_mut()
-                                .downcast_mut::<WgpuRenderBackend<MovieView>>()
-                                .expect("Renderer must be correct type");
-                            self.gui
-                                .lock()
-                                .expect("Gui lock")
-                                .render(Some(renderer.target()));
+                            if player.is_playing() {
+                                player.render();
```
adrian17, line 140: to be sure, we don't need to keep refreshing the render in case something draws on top of it?

desktop/src/app.rs
Dinnerbone: I'll take a look! It won't handle all of the cases but maybe it'll help

desktop/src/app.rs
Dinnerbone: We render ruffle to a texture, which shouldn't be able to change if the player is frozen. That texture is always rendered to the surface, and then UI after it, so it doesn't need any extra attention

adrian17 (approved):

Issue #11239, opened 2023-05-27
Error on https://game285222.konggames.com/games/jamboska/gift-clicker-mobile-edition/frame/d905e27e-3dff-4be4-a465-431e0ffcebee/


jexi27:
# Error Info
Error name: Error
Error message: panicked at 'capacity overflow', library/alloc/src/raw_vec.rs:525:5
Error stack:
```
Error: panicked at 'capacity overflow', library/alloc/src/raw_vec.rs:525:5
    at W.n.wbg.__wbg_new_87297f22973157c8 (https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.ee5716355123d2025211.js:1:50966)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::hafc7beb469c6cb04 (wasm://wasm/02c3637e:wasm-function[8620]:0x82e6b0)
    at std::panicking::rust_panic_with_hook::h7f7102b82d51338f (wasm://wasm/02c3637e:wasm-function[5613]:0x751612)
    at core::panicking::panic_fmt::he4489d678d6570d5 (wasm://wasm/02c3637e:wasm-function[8469]:0x829a1d)
    at alloc::raw_vec::capacity_overflow::h743526e95bf6b28b (wasm://wasm/02c3637e:wasm-function[8236]:0x8238d1)
    at <core::iter::adapters::flatten::FlatMap<I,U,F> as core::iter::traits::iterator::Iterator>::next::ha0792ffbf69c626d (wasm://wasm/02c3637e:wasm-function[2633]:0x59d8e3)
    at <ruffle_render_wgpu::context3d::WgpuContext3D as ruffle_render::backend::Context3D>::process_command::h54e9288cf0147e56 (wasm://wasm/02c3637e:wasm-function[531]:0xb148d)
    at ruffle_core::avm2::object::context3d_object::Context3DObject::copy_bitmap_to_texture::h19ac733f891021c5 (wasm://wasm/02c3637e:wasm-function[4661]:0x700453)
    at ruffle_core::avm2::globals::flash::display3D::textures::rectangle_texture::upload_from_bitmap_data::ha9949a70beb0757a (wasm://wasm/02c3637e:wasm-function[2689]:0x5ad8e3)
    at ruffle_core::avm2::function::Executable::exec::hba6a908a607c674f (wasm://wasm/02c3637e:wasm-function[799]:0x2871ad)
```
AVM2 stack:
```
[...]

PR #11245, opened 2023-05-28
wgpu: Fix calculation of image row padding

Aaron1011:


Lord-McSweeney:
Can you add a regression test for this?

n0samu:
With this PR, finally all of the Kings Kollege games work in Ruffle! 🎉

This also fixes #11239, although the game still crashes for a different reason: #11252

adrian17 (approved):

PR #11256, opened 2023-05-28
web: Allow extra MIME types in the polyfill

### LINKED ISSUE #11050: Ruffle extension does not work on Tooli's Classic Games site
### Describe the bug

Example game page: https://www.tooli.co.kr/flash/472106

The Ruffle extension does not load on pages of the Tooli's Classic Games site. It seems like some JavaScript on the page is interfering, but I have not been able to confirm/diagnose this yet.

### Expected behavior

Flash games should load and play like they do on other sites.

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Edge 113

### Additional information

_No response_

Toad06:
Closes #11050, by making the polyfill more lenient with the "application/octet-stream" and "binary/octet-stream" MIME types, as suggested by @n0samu in https://github.com/ruffle-rs/ruffle/issues/11050#issuecomment-1547507019.

### DIFF
--- a/web/packages/core/src/ruffle-embed.ts
+++ b/web/packages/core/src/ruffle-embed.ts
@@ -133,31 +133,36 @@ export class RuffleEmbed extends RufflePlayer {
      * Checks if the given element may be polyfilled with this one.
      *
      * @param elem Element to check.
-     * @returns True if the element looks like a flash embed.
+     * @returns True if the element looks like a Flash embed.
      */
     static isInterdictable(elem: Element): boolean {
+        const src = elem.getAttribute("src");
+        const type = elem.getAttribute("type");
+
+        // Don't polyfill when no file is specified.
+        if (!src) {
+            return false;
+        }
+
         // Don't polyfill if the element is inside a specific node.
         if (isFallbackElement(elem)) {
             return false;
         }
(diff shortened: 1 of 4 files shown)


web/packages/core/src/ruffle-embed.ts
```diff
@@ -133,31 +133,36 @@ export class RuffleEmbed extends RufflePlayer {
...
         // Check for MIME type.
-        const type = elem.getAttribute("type");
+        const isSwf = isSwfFilename(src);
         if (!type) {
             // If no MIME type is specified, polyfill if movie is an SWF file.
-            return isSwfFilename(elem.getAttribute("src"));
+            return isSwf;
         } else {
-            return isSwfMimeType(type);
+            return isSwfMimeType(type, isSwf);
```
relrelb, line 165: Looks like `isSwfMimeType()` is always called as `isSwfMimeType(mimeType, isSwfFilename(url))`. 

Does it make sense to simply unify them to a single function? Something like:
```ts
export function isSwf(url: string, mimeType?: string): boolean { ... }
```
Under the hoods there still can exist 2 separate functions, I talk about the exported interface.

n0samu (approved):
Looks perfect, thank you!

PR #11261, opened 2023-05-29
avm2: Run 'nested goto frame' for entire Stage and orphans

Aaron1011:
When we run a 'goto', a weird "nested frame" gets triggered. Previously, we were only calling `construct_frame` on the target MovieClip as part of this "nested frame". However, Flash Player seems to treat this (in some ways) like a normal frame - *all* objects on the Stage (and orphans) have `construct_frame` called.

In particular, `gotoAndStop`/`gotoAndPlay` is called during an "enterFrame" event handler, then unrelated objects on the Stage will have their children constructed during the execution of `gotoAndStop`/`gotoAndPlay`. The same logic holds for frame scripts.

This fixes a bug in Steamlands, which relies on children on the main timeline being constructed immediately following a call to `gotoAndStop` on an orphan (originally triggered from an "enterFrame" handler).


SN902:
Found gotoAndPlay regression in [Rainbow Dash Shooter Project.zip](https://github.com/ruffle-rs/ruffle/files/11588389/Rainbow.Dash.Shooter.Project.zip)

image

SN902:
MLP Tower defense performance issue fixed, and it also helps a lot #11346

actually this fix anything that has orphans with broadcast event listeners, now all similar games with orphans, which stay alive forever, no longer destroys performance

core/src/avm2.rs
```diff
@@ -256,16 +256,12 @@ impl<'gc> Avm2<'gc> {
...
         mut f: impl FnMut(DisplayObject<'gc>, &mut UpdateContext<'_, 'gc>),
     ) {
-        let mut i = 0;
         // FIXME - should we handle movies added while we're looping?
-        let total = context.avm2.orphan_objects.len();
-
         // We cannot use an iterator, as it would conflict with the mutable borrow of `context`.
-        while i < total {
-            if let Some(dobj) = valid_orphan(context.avm2.orphan_objects[i], context.gc_context) {
+        for orphan in context.avm2.orphan_objects.clone() {
```
adrian17, line 259: :(

adrian17 (approved):
Approved the goto-related parts; I'm assuming this is still blocked by the weak object PR?

core/src/avm2.rs
Aaron1011: This was a quick fix for a crash - I think I can introduce an `Rc` and use `Rc::make_mut`, like we already do for render list iteration

Aaron1011:
@adrian17 Yes, this needs https://github.com/ruffle-rs/ruffle/pull/11397

core/src/avm2.rs
Aaron1011: I split out this change into https://github.com/ruffle-rs/ruffle/pull/11414

PR #11271, opened 2023-05-29
Desktop: Stop busy looping when idling

adrian17:
Now, we only return `WaitUntil` if the player is running and actually can give us the wakeup time.


n0samu (approved):
Working perfectly for me, thank you!

Dinnerbone (approved):

Issue #10403, opened 2023-03-27
Runes of the Ancient Forest - Black Screen (Weird Error)


desertking:
### Describe the bug

Loading SWF file https://www.spiele-umsonst.de/azad/downloads/flashgames/runesoftheancientforest.swf
ruffle_web-wasm_extensions.js:677 ERROR core/src/avm2/events.rs:419 Error dispatching event EventObject(EventObject { type: "addedToStage", class: flash.events::Event, ptr: 0x2c4cdd0 }) to handler FunctionObject(FunctionObject { ptr: 0x26dc4b8 }) : Error: It looks like custom metadata is being ignored by your compiler. If you're compiling with the Flash IDE you need to open your "Publish Settings" and select "Publish SWC". See: https://github.com/robotlegs/robotlegs-framework/wiki/Common-Problems

### Expected behavior

Load and Run

### Affected platform

Online demo

### Operating system

All

### Browser

All

Lord-McSweeney:
Seems like the same underlying issue as #8330.

PR #11275, opened 2023-05-29
avm2: Support reading metadata through flash.utils.describeType

Lord-McSweeney:
This caught a bug in `swf`: `swf` would parse key-value pairs for metadata incorrectly.
Progresses New Club Penguin (#8330), and also anything else that uses the Robotlegs or SwiftSuspenders frameworks.

Unfortunately this doesn't progress #10403, even though I've confirmed that the class that was being checked has the same `describeType` output in FP and in Ruffle (so I would assume something's breaking with Robotlegs).


Aaron1011 (approved):

Dinnerbone (approved):

core/src/avm2/traits.rs
```diff
@@ -29,6 +31,58 @@ bitflags! {
...
+impl<'gc> TraitMetadata<'gc> {
+    // Represents this TraitMetadata as an XML string, as defined in describeType
+    pub fn as_xml_string(&self, mc: MutationContext<'gc, '_>) -> AvmString<'gc> {
+        let mut xml_string = AvmString::new_utf8(mc, "");
+        if self.items.is_empty() {
+            // This was in the form of [metadata]
+            xml_string = AvmString::concat(
+                mc,
+                xml_string,
+                AvmString::new_utf8(mc, format!("<metadata name=\"{}\"/>", self.name)),
```
adrian17, line 59: most/all of this should probably be in `utils.rs` together with the other DescribeType-related XML string creation code.

core/src/avm2/vtable.rs
```diff
@@ -349,18 +386,23 @@ impl<'gc> VTable<'gc> {
                         Some(Property::Virtual {
                             set: Some(disp_id), ..
                         }) => {
+                            // Override metadata
```
adrian17, line 389: (IMO we don't really need this comment, the line below is already implied to be an overriding. We might wanna have a more generic comment just saying that this block corresponds to method overriding.)

core/src/avm2/traits.rs
```diff
@@ -222,6 +335,7 @@ impl<'gc> Trait<'gc> {
                     disp_id: *disp_id,
                     method: unit.load_method(*method, false, activation)?,
                 },
+                metadata: metadata_from_abc_trait(activation, unit, abc_trait.metadata.clone())?,
```
adrian17, line 338: I thought we wanted these to be loaded lazily (at `describeType()` time) from the TU, rather than eagerly? Most metadata is not used so we'd prefer to just store the index and forget about it.

core/src/avm2/traits.rs
```diff
@@ -230,6 +344,7 @@ impl<'gc> Trait<'gc> {
...
```
adrian17, line 347: note: this should probably borrow the Vec, rather than copy it.

core/src/avm2/traits.rs
```diff
@@ -287,6 +410,10 @@ impl<'gc> Trait<'gc> {
         &self.kind
     }
 
+    pub fn metadata(&self) -> Vec<TraitMetadata<'gc>> {
+        self.metadata.clone()
```
adrian17, line 414: If we _do_ decide store it here, we might wanna store it behind a Gc, to minimize Trait struct size and vec copying.

Lord-McSweeney: Stored behind an `Option<Box<`

core/src/avm2/traits.rs
Lord-McSweeney: This will be resolved to None very quickly if there's no metadata

core/src/avm2/traits.rs
```diff
@@ -60,6 +81,54 @@ fn trait_attribs_from_abc_traits(abc_trait: &AbcTrait) -> TraitAttributes {
...
+    }
+
+    let abc = translation_unit.abc();
+    let mut trait_metadata_list = vec![];
+    for single_metadata in metadata.iter() {
+        // Lookup the Index<Metadata> to convert it into a Metadata.
+        let single_metadata = abc
+            .metadata
+            .get(single_metadata.0 as usize)
+            .ok_or_else(|| -> Error { format!("Unknown metadata {}", single_metadata.0).into() })?;
```
adrian17, line 101: Weird, we didn't need explicit typing before. Like:
```rs
.ok_or_else(|| format!("Unknown string constant {string_index}"))?
```
Can you check it out?

core/src/avm2/globals/flash/utils.rs
```diff
@@ -458,7 +472,23 @@ fn describe_internal_body<'gc>(
...
+
+                if let Some(get_disp_id) = get {
+                    let trait_metadata_getter = disp_metadata_table.get(get_disp_id);
+                    if let Some(metadata) = trait_metadata_getter {
+                        write_metadata(&mut xml_string, metadata);
+                    }
+                }
+
+                if let Some(set_disp_id) = set {
+                    let trait_metadata_setter = disp_metadata_table.get(set_disp_id);
```
adrian17, line 485: we probably don't need an additional var here?
```rs
if let Some(metadata) = disp_metadata_table.get(set_disp_id) {
```

core/src/avm2/traits.rs
```diff
@@ -29,6 +31,22 @@ bitflags! {
...
+#[collect(no_drop)]
+pub struct TraitMetadataItem<'gc> {
+    pub key: AvmString<'gc>,
+    pub value: AvmString<'gc>,
+}
+
+// Represents a single metadata item for a trait.
+#[derive(Clone, Collect, Debug, Eq, PartialEq)]
+#[collect(no_drop)]
+pub struct TraitMetadata<'gc> {
```
adrian17, line 45: Actually, what do you think about moving this code to `metadata.rs` and dropping the `Trait` from the struct name?
The structs themselves and `fn metadata_from_abc_trait` don't really have anything directly related to traits; and traits.rs in the past avoided directly interacting with TU.

core/src/avm2/vtable.rs
```diff
@@ -486,6 +551,10 @@ impl<'gc> VTable<'gc> {
         write
             .resolved_traits
             .insert(name, Property::new_slot(new_slot_id));
+        // Since there's no metadata (?), there's no need to insert
```
adrian17, line 554: (to be removed, as mentioned on Discord)

adrian17 (approved):

core/src/avm2/vtable.rs
```diff
@@ -494,6 +560,7 @@ impl<'gc> VTable<'gc> {
         write
             .resolved_traits
             .insert(name, Property::new_slot(new_slot_id));
+
```
adrian17, line 563: (some random whitespace change)

core/src/avm2/globals/flash/utils.rs
```diff
@@ -386,6 +387,8 @@ fn describe_internal_body<'gc>(
         if !ns.is_public() {
             continue;
         }
+        let slot_metadata_table = vtable.slot_metadata_table();
```
adrian17, line 390: instead of exposing the tables via `Ref`, how about encapsulating with
```
let trait_metadata = vtable.get_metadata_for_slot(slot_id);
```
?

adrian17 (approved):

core/src/avm2/globals/flash/utils.rs
Lord-McSweeney: I prefer it the way it currently is.

PR #11298, opened 2023-05-30
[desktop] use light menus on light theme & detect changes

tinaun:
if the user is on light mode, use light menu bars and windows within egui - also handles changing modes when ruffle is running


Dinnerbone (approved):
Neat! I'd like an option for the app when we eventually have a settings window, but respecting OS setting is best :D

PR #11307, opened 2023-05-30
desktop: rewrite AS3 warning popup in egui

adrian17:
Not a fan of how many layers I needed to go through hah ;D

Benefits:
- doesn't block the main thread anymore
- can be dismissed by clicking anywhere
- the link is actually clickable :)

Ideally this code can be later reused to replace the generic AS3 popup by a per-feature popup, but that's out of scope.
I didn't touch the other messages supported by the DesktopUiBackend.

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -452,6 +452,13 @@ impl App {
                     );
                 }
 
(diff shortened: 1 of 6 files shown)


adrian17:
How it looks like:

image

Aaron1011:
Could you add something to the message to indicate that you can click to close it?

Aaron1011 (approved):

PR #11309, opened 2023-05-31
avm2: Add PixelBender bytecode parsing to ShaderData

Aaron1011:
We now parse PixelBender bytecode, and populate the parameters from the bytecode on `ShaderData`. This is enough to progress Steamlands, which needs to access dynamically set properties on `ShaderData`

Bytecode execution is not implemented yet.


core/src/avm2/globals/flash/display/Shader.as
```diff
@@ -4,7 +4,7 @@ package flash.display {
     import __ruffle__.stub_setter;
 
     public class Shader {
-        private var _data:ShaderData;
+        private var _data:ShaderData = new ShaderData(new ByteArray());
```
Lord-McSweeney, line 7: I don't think this should be initialized- in Flash, `new Shader().data` returns `null`.

Dinnerbone (approved):

PR #11318, opened 2023-05-31
core: Mark stack-heavy DO functions as noinline

adrian17:
This should make the stack size commit of https://github.com/ruffle-rs/ruffle/pull/11261 unnecessary.
@Aaron1011 can you try it out please? I'm confident it'll work on desktop, but slightly worried about wasm-opt messing things up.

### DIFF
--- a/core/src/display_object.rs
+++ b/core/src/display_object.rs
@@ -1663,6 +1663,7 @@ pub trait TDisplayObject<'gc>:
     /// Since we construct AVM2 display objects after they are allocated and
     /// placed on the render list, these steps have to be done by the child
     /// object to signal to its parent that it was added.
+    #[inline(never)]
     fn on_construction_complete(&self, context: &mut UpdateContext<'_, 'gc>) {
         if !self.placed_by_script() {
             // Since we construct AVM2 display objects after they are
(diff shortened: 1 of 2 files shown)


Aaron1011:
@adrian17 Unfortunately, wasm still hits a stack overflow with bloons tower defense 5 on this PR (with the default stack size).

n0samu:
Is that both with and without wasm-opt?

n0samu:
I just tried testing this by adding the commit on top of #11261 and reverting 2233304f45d712001d6e74293f75288016b6f6f2. I actually get a stack overflow on both the desktop app (Windows 10) and the web demo.

Dinnerbone (approved):
Even if it doesn't fix web, let's help desktop

n0samu:
It didn't help on desktop or web for me, like I said already

adrian17:
There's no harm in this being merged, though indeed most of the pain comes from the UpdateContext copying.

torokati44:
> ... most of the pain comes from the UpdateContext copying.

@adrian17: Just curious, does this mean a literal `memcpy`-style copying of a large struct, or something more involved? If the former, is the `bulk-memory` intrinsic used (or at least considered by the codegen) for it? If not, does the `build-std` nightly compiler flag change anything?

I'd love to investigate all of this, once I have the time, but if anyone else can do it before me, that'd be great.

PR #11323, opened 2023-05-31
wgpu: Don't skip alpha/erase blend modes for BitmapData.draw

### LINKED ISSUE #10319: In Bloons TD 5, you can't put a monkey on the ground
### Describe the bug

In the first level of Bloons TD 5, you can't put a monkey on the ground. It worked before https://github.com/ruffle-rs/ruffle/commit/078672782ccac2eb38c23e739fc820e32e8bd052

image

Online Game - https://www.gameflare.com/online-game/bloons-tower-defense-5/
SWF - https://data.gameflare.com/games/3550/WDzRSvadTXBNOR.swf

### Expected behavior

Put a monkey on the ground

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

_No response_

Aaron1011:
We can skip these when we're drawing to a fresh buffer, but not when we're blending over an existing BitmapData texture.

Fixes #10319


Dinnerbone:
I think you've forgotten to commit the test image

render/wgpu/src/surface.rs
```diff
@@ -228,8 +231,13 @@ impl Surface {
                     let parent = match blend_mode {
                         ComplexBlend::Alpha | ComplexBlend::Erase => match nearest_layer {
                             LayerRef::None => {
-                                // An Alpha or Erase with no Layer above it should be ignored
-                                continue;
+                                if render_mode_existing_texture {
```
Dinnerbone, line 234: If I understand this right, this makes `LayerRef::None` act like `LayerRef::Current` for the deepest layer. Is that right? If there's things between, I somewhat expected it to become the the furthest ancestor?

Could you check that please? I think just passing in `LayerRef::Current` in the initial call to `draw_commands` for the case of bmd.draw() would do that behaviour, but we need to see which one is right first

Aaron1011: @Dinnerbone It looks like we only ever create `LayerRef::None` at the top level `draw_commands_to` (called from `render_offscreen` and `submit_frame`), so we'll never have any ancestor layers when we see it. However, I think you're right that it makes more sense to explicitly pass in `LayerRef::Current`.

Aaron1011:
@Dinnerbone This was originally not an image test (it just tested printing the pixels), but I've changed it to render the `BitmapData`s.

Dinnerbone (approved):

PR #11349, opened 2023-06-02
chore: Use Rust's `OnceCell`/`OnceLock`

relrelb:
They were stabilized in Rust 1.70.0, which was released yesterday. This removes the `once_cell` dependency.

Also, while working on this I noticed that `color_needs_clear` may never be set, possibly leading to some dead code. @Dinnerbone Can you confirm that?


Dinnerbone:
> Also, while working on this I noticed that `color_needs_clear` may never be set, possibly leading to some dead code. @Dinnerbone Can you confirm that?

Not looked at your diff, but that's not the case in the current code - the value doesn't matter, the set-ness does.

adrian17 (approved):

Dinnerbone (approved):

PR #11351, opened 2023-06-02
desktop: `Open Advanced` improvements

moulins:
- Remember options across movie loads;
  - using `Open`  will replace any options by the default options provided on the command-line;
- adds `Ctrl-Shift-O` as a shortcut;
- adds a `Clear all` button to the movie parameters section;
- The `Browse` button automatically opens in the current file's directory.

Also, adds a `File > Reload` menu command to quickly reload the current movie.

### DIFF
--- a/desktop/assets/texts/en-US/main_menu.ftl
+++ b/desktop/assets/texts/en-US/main_menu.ftl
@@ -11,6 +11,7 @@ file-menu = File
 file-menu-open-quick = Open...
 file-menu-open-advanced = Open Advanced...
 file-menu-close = Close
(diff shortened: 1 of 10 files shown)


Dinnerbone:
> Makes opening the same movie repeatedly less tedious.

Perhaps a Reload option would be useful, too?

moulins:
Good idea, I'll implement that!

Toad06 (approved):
Tested on Windows 10, works perfectly. 👍

Dinnerbone (approved):

PR #11361, opened 2023-06-03
avm1: Implement `MovieClip.lineGradientStyle`

Toad06:
This is mostly based on the implementation of `Graphics.lineGradientStyle` in AVM2 made by Dinnerbone (cc. #10123).

This PR also adjusts the implementation of `MovieClip.beginGradientFill`, mostly to avoid code duplication but also to fix some minor issues.

Tests are added for `MovieClip.beginGradientFill` and `MovieClip.lineGradientFill`, but are currently ignored due to the number of outliers (though at a glance, the differences vs Flash are not that obvious to notice).


tests/tests/swfs/avm1/movieclip_begin_gradient_fill/test.toml
```diff
@@ -0,0 +1,8 @@
+ignore = true
+num_frames = 1
+
+[image_comparison]
+tolerance = 0
+
+[player_options]
+with_renderer = { optional = true, sample_count = 1 }
```
Dinnerbone, line 8: If there's no output.txt to compare with, make the renderer required as it's only testing the visuals, nothing else. Same for the other test

tests/tests/swfs/avm1/movieclip_begin_gradient_fill/test.toml
```diff
@@ -0,0 +1,8 @@
+ignore = true
```
Dinnerbone, line 1: Ignored because there's still more unimplemented, or a mistake?

Toad06: We don't exactly match Flash here and this causes the number of outliers to be very high.

For instance, here's what the `difference-color-windows-Vulkan.png` file looks like when running this test (`avm1/movieclip_begin_gradient_fill`):

new

For the record, this is the same thing in master, with a few extra issues:
old

Dinnerbone: Ah, did you make expected.png from flash?

Toad06: Yes, this is made based on the Flash projector.

Dinnerbone: Was flash set to low quality?

If it visually looks the same, use the expected.png that ruffle generates (it's just actual.png), it's to stop regressions more than to make sure it's identical to flash.

Toad06: No, the quality was set to high. I tested setting it to low, this only affects the way borders around the squares are rounded (no alpha), otherwise everything else looks the same.

Dinnerbone: Understood. Can you please switch to using the image Ruffle generated for `expected.png` and unignoring the test though? The purpose isn't to test against flash (humans should do that), but to test against two versions of Ruffle and make sure we don't regress it.

Toad06: Alright, thanks for clarifying this. I made the changes (and also renamed the `source.txt` files to `source.as` as suggested in the other PR).

Dinnerbone: Looks like we'll need a tolerance of 3 on both tests to account for the test runners, and then we're good, thank you!

Toad06: Thank you for the assistance. 😉

Dinnerbone (approved):

PR #11370, opened 2023-06-04
desktop: load cjk fallback fonts using fontdb

### LINKED ISSUE #11234: EGUI: Can't display text on translated text. (Latest Nightly)
### Describe the bug

image

### Expected behavior

It should be normal text.

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

_No response_

tinaun:
im unsure where to put this, so everything is in GuiController

closes #11234


desktop/src/gui/controller.rs
```diff
@@ -286,3 +289,49 @@ impl GuiController {
...
+            Family::Name("Arial Unicode MS"), // macos
+            Family::Name("Noto Sans"),        // linux
+            Family::SansSerif,
+        ],
+        ..Query::default()
+    };
+
+    let id = font_database
+        .query(&system_unicode_fonts)
+        .ok_or(anyhow!("no unicode fonts found!"))?;
```
Dinnerbone, line 310: I think it'd be better to log an error and continue, rather than panic. Unless including `SansSerif` is supposed to make it infallible?

Dinnerbone:
At least on Windows, the font we include here doesn't support Korean:
image

Looking at https://en.wikipedia.org/wiki/List_of_CJK_fonts, we'll need to pick a few options for each OS to cover C, J and K separately. Possibly different fonts depending on which chinese locale specifically, but I'm not sure about that.

danielhjacobs:
I would expect the Mac and Linux options to work pretty well already, though a few people with each OS should probably test them. [Arial Unicode MS](https://en.wikipedia.org/wiki/Arial_Unicode_MS) is pan-unicode and included on Mac since Mac OS X v10.5. Noto Sans is a font-family with font files containing full CJK support, and has become the default on several Linux distros, including [Fedora](https://fedoraproject.org/wiki/Changes/DefaultToNotoFonts) and [Ubuntu (I think this link says that at least)](https://bugs.launchpad.net/ubuntu/+source/language-selector/+bug/1468027).

For Windows I'd suggest [MS Gothic](https://learn.microsoft.com/en-us/typography/font-list/ms-gothic) (Japanese), [Microsoft JhengHei](https://learn.microsoft.com/en-us/typography/font-list/microsoft-jhenghei) (Traditional Chinese), [Microsoft YaHei](https://learn.microsoft.com/en-us/typography/font-list/microsoft-yahei) (Simplified Chinese), and [Malgun Gothic](https://learn.microsoft.com/en-us/typography/font-list/malgun-gothic) (Korean).

desktop/src/gui/controller.rs
```diff
@@ -286,3 +289,49 @@ impl GuiController {
...
+
+    let system_unicode_fonts = Query {
+        families: &[
+            Family::Name("MS UI Gothic"),     // windows
+            Family::Name("Arial Unicode MS"), // macos
+            Family::Name("Noto Sans"),        // linux
+            Family::SansSerif,
+        ],
+        ..Query::default()
+    };
```
CatMe0w, line 306: Is it loading only `MS UI Gothic` on Windows? If so, it will always display Japanese characters, ignoring CJK settings (locale) of the system.

n0samu: Yeah, this was pointed out by Dinnerbone as well. Do you agree with Daniel's suggested font list above?

desktop/src/gui/controller.rs
tinaun: ok_or doesn't panic, I return immediately and load the default fonts if anything goes wrong- I should log it though, right now it does silently fail

desktop/src/gui/controller.rs
tinaun: I'll include whatever fonts you all deem acceptable for minimum viable coverage of all character systems commonly used. is there a easy way to refer to the system locale?

danielhjacobs: I assume that would be this:

https://github.com/ruffle-rs/ruffle/blob/fcd9935a8102717ce40d893f252da2723f343717/desktop/src/gui.rs#L21

https://github.com/ruffle-rs/ruffle/blob/fcd9935a8102717ce40d893f252da2723f343717/desktop/src/gui.rs#L80-L83

As the comment says "This should also be somewhere else so it can be supplied through UiBackend too". So right not I think it would just be by re-using that method.

Dinnerbone:
Given https://github.com/ruffle-rs/ruffle/issues/11538 we also need to look at arabic

Dinnerbone:
@tinaun will you be able to look into fixing this up, or do you want someone else to try and pick it up?

creeper-0910:
Fonts for Japanese may be easier to read here.
https://github.com/reindex-ot/RobotoNotoSansJP

n0samu:
Adding a custom font for each CJK language would significantly increase our binary size. So we should use system-installed fonts instead, as this PR currently does.

Dinnerbone (approved):
I think we're still missing languages such as Arabic, but this has waited long enough, let's get it in and add to it afterwards


@tinaun thank you!

Dinnerbone:
The contents of this PR are identical to #11924 which is rebasable because of the lack of merge commit, so I'm going to pull that one in - but it's the same contents, and the authors of the commits is unchanged.

PR #11371, opened 2023-06-04
core: Fix parsing of PixelBender sampleNearest and sampleLinear

Aaron1011:
This fixes a panic on startup for Sniper Team


core/src/pixel_bender.rs
```diff
@@ -367,11 +379,30 @@ fn read_op<R: Read>(
...
+                }
+                _ => unreachable!(),
+            }
+        }
         _ => {
             let dst = data.read_u16::<LittleEndian>()?;
             let mask = data.read_u8()?;
             let src = read_uint24(data)?;
-            assert_eq!(data.read_u8()?, 0);
+            assert_eq!(data.read_u8()?, 0, "Unepxected u8 for opcode {opcode:?}");
```
Lord-McSweeney, line 405: If this is possible in case of a corrupt shader, would it be a good idea to move this behind a warn/err?

Aaron1011: This caught a parsing bug, so I want to keep it in order to encourage bug reports

danielhjacobs: W.R.T. the same line, there's a typo: "Unepxected"

adrian17 (approved):

PR #11374, opened 2023-06-04
avm1: fix coercions for the `blendMode` property of display objects

moulins:
`mc.blendMode = 3` is a valid way of modifying the blend mode of a DO; on the other hand, booleans and objects *shouldn't* be coerced.


Toad06 (approved):

moulins:
Added a new commit to properly version-gate `blendMode` to SWF >= 8, which is what caused my confusion in #11375.

Dinnerbone (approved):

PR #11399, opened 2023-06-06
core: Correctly handle '.' and '<' in obfuscated QNames

Aaron1011:
QName can contain characters like '.' and '<' as part of the package or class name (though this requires editing the SWF or using a nonstandard compiler). This broke our attempt to parse generic type paramters by looking for '.<'

Instead, our 'Vector' special casing now operates on the unparsed 'AVMString' name, instead of attempting to construct a 'QName'. This means that we don't need to handle generic paramters at all for obfuscated names (which will never start with '__AS3__' or 'Vector.<')

This fixes a bug in Red Ball 4v3, which has an obfuscated class '!D.<H'


core/src/avm2/qname.rs
```diff
@@ -77,34 +77,16 @@ impl<'gc> QName<'gc> {
     /// NAMESPACE.LOCAL_NAME (Where the LAST dot is used to split the namespace & local_name)
     /// LOCAL_NAME (Use the public namespace)
     pub fn from_qualified_name(name: AvmString<'gc>, activation: &mut Activation<'_, 'gc>) -> Self {
```
adrian17, line 79: can you please add a comment to the method, saying that it explicitly doesn't support `.<T>`?

adrian17 (approved):

PR #11407, opened 2023-06-07
web: Don't try to use Ruffle on browsers with Flash support

danielhjacobs:


web/packages/core/src/plugin-polyfill.ts
```diff
@@ -216,6 +228,9 @@ declare global {
  * @param plugin The plugin to install
  */
 export function installPlugin(plugin: RufflePlugin): void {
+    if (isFlashEnabledBrowser()) {
```
relrelb, line 231: For symmetricity, can this be moved into `pluginPolyfill()` in `polyfills.ts`? Then `isFlashEnabledBrowser` can be moved there as well without an `export`.

danielhjacobs: It can, though that function is not called by the Ruffle extension, which calls installPlugin directly, but if you explicitly install the Ruffle extension on a Flash-enabled browser, you probably want to use Ruffle. In fact, perhaps, I should also change `if (!isFlashEnabledBrowser() && (isExtension || !usingExtension)) {` to `if (isExtension || (!isFlashEnabledBrowser() && !usingExtension)) {`

danielhjacobs:
I should note that on Discord Dinnerbone said "I'm still worried about adding flash-detection to our anti-flash-detection. I don't think I can articulate exactly why, but it feels like we're going to paint ourselves into a corner a bit"

I personally think it's fine but this should perhaps be discussed.

relrelb (approved):

Dinnerbone (approved):

adrian17 (approved):

PR #11417, opened 2023-06-07
avm2: Make LoaderInfo.content delegate to Loader.content

Aaron1011:


Lord-McSweeney:
I think the function should be this:
```
/// `content` getter
pub fn get_content<'gc>(
    this: Option<Object<'gc>>,
    _args: &[Value<'gc>],
    if let Some(this) = this {
        if let Some(loader) = this.as_loader_info_object().and_then(|o| o.loader()) {
            return loader.get_public_property("content", activation);
        } else {
            if let Some(loader_stream) = this
                .as_loader_info_object()
                .and_then(|o| o.as_loader_stream())
            {
                match &*loader_stream {
                    LoaderStream::Swf(_, root) | LoaderStream::NotYetLoaded(_, Some(root), _) => {
                        return Ok(root.object2());
                    }
                    _ => {
                        return Ok(Value::Null);
                    }
                }
            }
        }
    }

    Ok(Value::Undefined)
}
```

Aaron1011:
After further investigation, I think this is the wrong approach. I'm going to do some refactoring of `Loader` in separate PRs.

PR #11419, opened 2023-06-07
Many more debug UI features

Dinnerbone:
image
image

### DIFF
--- a/core/src/debug_ui/movie.rs
+++ b/core/src/debug_ui/movie.rs
@@ -1,0 +1,239 @@
+use crate::character::Character;
+use crate::context::UpdateContext;
+use crate::debug_ui::Message;
+use crate::tag_utils::SwfMovie;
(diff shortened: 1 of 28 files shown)


core/src/debug_ui/movie.rs
```diff
@@ -0,0 +1,320 @@
...
+
+pub fn open_movie_button(ui: &mut Ui, movie: &Arc<SwfMovie>, messages: &mut Vec<Message>) {
+    if ui.button(movie_name(movie)).clicked() {
+        messages.push(Message::TrackMovie(movie.clone()));
+    }
+}
+
+pub fn open_character_button(ui: &mut Ui, character: &Character) {
+    match character {
+        Character::EditText(_) => {
```
adrian17, line 261: do you expect some more complex per-type logic here? Otherwise, how about a much simpler enum->string literal match, and a single `label()` call at the end?

adrian17 (approved):

Issue #11252, opened 2023-05-28
Crash on GiftClicker.swf


n0samu:
### Describe what you were doing

Trying to run the game. This is the same game from #11239 - this is the error I get after merging in the upcoming PRs including #11245.

Link to SWF (for convenience): https://game285222.konggames.com/gamez/0028/5222/live/GiftClicker.swf

### What does the crash message say?

panicked at 'wgpu error: Validation Error

Caused by:
    In Device::create_render_pipeline
    Error matching ShaderStages(NONE | FRAGMENT) shader requirements against the pipeline
    Location[1] is provided by the previous stage output but is not consumed as input by this stage.

', C:\Users\nosamu\.cargo\registry\src\github.com-1ecc6299db9ec823\wgpu-0.16.0\src\backend\direct.rs:3019:5

### Ruffle Version

0.1.0-nightly (32069e63f09bfa3eee4743fe591be0acdb861af3 2023-05-28)

### Affected platform

Desktop app

### Operating system

Windows 10.0.19045 (Windows 10 Pro) [64-bit]

### Additional information

Filename: GiftClicker.swf

### AVM2 Callstack
```
	at flash.display3D::Context3D/flash::display3D::Context3D::drawTriangles()
	at starling.display.materials::StandardMaterial/drawTriangles()
	at starling.display.graphics::Graphic/render()
	at starling.display::DisplayObjectContainer/render()
	at starling.display::DisplayObjectContainer/render()
	at starling.display::Sprite/render()
	at starling.display::DisplayObjectContainer/render()
	at starling.display::Sprite/render()
	at starling.display::DisplayObjectContainer/render()
	at starling.display::Sprite/render()
	at starling.display::DisplayObjectContainer/render()
	at starling.core::Starling/render()
	at starling.core::Starling/nextFrame()
	at starling.core::Starling/onEnterFrame()
```

### Render Info
Renderer: wgpu
Adapter Backend: Vulkan
Adapter Name: "AMD FirePro W5100"
Adapter Device Type: DiscreteGpu
Adapter Driver Name: "AMD proprietary driver"
Adapter Driver Info: ""
Enabled features: Features(TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES | PUSH_CONSTANTS)
[...]

PR #11441, opened 2023-06-09
wpgu: Initial implementation of PixelBender shader execution

Aaron1011:
The implementation is split across four crates:
* `ruffle_render` now holds the main PixelBender bytecode parsing implementation (previously, this was in `ruffle_core`).
* `ruffle_core` holds some helper functions for converting between AVM2 `Value`s and the PixelBender vector types.
* `naga-pixelbender` (newly created) constructs a Naga `Module` from parsed PixelBender bytecode
* `ruffle_render_wgpu` sets up the render pipeline for the shader constructed by `naga-pixelbender`, and actually executes the shader.

The Actionscript-side shader parameters are passed in through uniforms. This allows us to cache the compiled `naga::Module` and associated wgpu types inside `ShaderData`, when it's first created. Each invocation of a `ShaderJob` only needs to create a bind group and render pass.

Limitations:

* Only a few of the PixelBender opcodes are implemented - however, this is enough to get Stemlands cannon rotation working, as well as a cool "donut" shader that I found and included as a test.
* PixelBender matrix types are not supported.
* Only BitmapData is supported as an input/output type - Flash Player also supports using Vector and ByteArray
* ShaderJob execution is always synchronous.


core/src/avm2/globals/flash/display/shader_job.rs
```diff
@@ -0,0 +1,165 @@
...
+    let shader = this
+        .get_public_property("shader", activation)?
+        .as_object()
+        .expect("Missing Shader object");
+    let shader_data = shader
+        .get_public_property("data", activation)?
+        .as_object()
+        .expect("Missing ShaderData object")
+        .as_shader_data()
+        .expect("ShaderData object is not a ShaderData instance");
```
Lord-McSweeney, line 39: Since these could be `null`, would it be better to do null handling rather than panic?

Aaron1011: This PR is already very large, so I'm postponing nicer error handling for a follow-up PR.

render/src/pixel_bender.rs
```diff
@@ -0,0 +1,636 @@
+//! Pixel bender bytecode parsing code.
+//! This is heavling based on https://github.com/jamesward/pbjas and https://github.com/HaxeFoundation/format/tree/master/format/pbj
```
Lord-McSweeney, line 2: nit: Typo: "heavling" --> "heavily"

core/src/avm2/globals/flash/display/shader_job.rs
Lord-McSweeney: Could you add some TODO notices in that case?

Lord-McSweeney:
Is there any chance #11252 could occur with the naga compilation?

Aaron1011:
@Lord-McSweeney Fortunately, no - PixelBender only allows you to write a fragment shader. The vertex shader is provided by us, and only has a single output (the fragment position).

core/src/avm2/globals/flash/display/ShaderJob.as
```diff
@@ -34,12 +37,21 @@ package flash.display {
...
 
         public function get shader():Shader {
-            stub_getter("flash.display.ShaderJob", "shader");
-            return null;
+            return this._shader;
         }
 
         public function set shader(value:Shader):void {
+            this._shader = value;
             stub_setter("flash.display.ShaderJob", "shader");
```
Dinnerbone, line 45: This isn't a stub anymore

core/src/avm2/globals/flash/display/ShaderJob.as
```diff
@@ -34,12 +37,21 @@ package flash.display {
...
         }
+
+        public function get target():Object {
+            return this._target;
+        }
+
+        public function set target(value:Object):void {
+            this._target = value;
+            stub_setter("flash.display.ShaderJob", "target");
```
Dinnerbone, line 54: Same here, if we're setting it, it's not a stub

desktop/src/main.rs
```diff
@@ -67,6 +67,9 @@ fn init() {
     let subscriber = tracing_subscriber::fmt::Subscriber::builder()
         .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
         .finish();
+
+    env_logger::init();
```
Dinnerbone, line 71: What's this for?

Aaron1011: This gives us logs from naga, which uses the `log` crate

Dinnerbone: We already should be seeing those, we redirect log to tracing. This would replace that redirection

Aaron1011: It looks like we're only using `tracing_log` on web. I'll remove `env_logger` from this pr, and enable `tracing_log` in a separate PR.

Dinnerbone: `tracing_subscriber` uses `tracing_log` by default, perhaps something's going wrong there? But I'm sure I've seen wgpu logs on desktop and those are log.

Dinnerbone (approved):

PR #11449, opened 2023-06-09
avm1: Don't create _levelN for unloadMovieNum(N)

### LINKED ISSUE #11136: AdventureQuest: Shops do not open when clicked
### Describe the bug

When you click on a link to a shop, it does not open a shop. Example:

Peek 2023-06-09 14-35

### Expected behavior

A Shop Menu should appear when you click on the shop link.

### Affected platform

Desktop app

### Operating system

Archlinux

### Browser

Brave Browser

### Additional information

_No response_

Dinnerbone:
I also found some other bugs, but I'm not tackling them now (someone else please feel free 😅). They all seem to be specific to Num:
- We make the `_levelN` immediately, but Flash makes it when the movie loads successfully
- We call `onUnload` but Flash doesn't
- We never remove `_levelN` (just empty it) but Flash does it on the next frame

This is enough to fix #11136


core/src/avm1/activation.rs
```diff
@@ -2879,6 +2889,14 @@ impl<'a, 'gc> Activation<'a, 'gc> {
         }
     }
 
+    /// Tries to resolve a level by ID. Returns None if it does not exist.
+    ///
+    /// If the level does not exist, then it will be created and instantiated
```
Lord-McSweeney, line 2894: Is this comment accurate?

Dinnerbone: Fixed!

core/src/avm1/activation.rs
```diff
@@ -2879,6 +2889,11 @@ impl<'a, 'gc> Activation<'a, 'gc> {
         }
     }
 
+    /// Tries to resolve a level by ID. Returns None if it does not exist.
+    pub fn try_resolve_level(&mut self, level_id: i32) -> Option<DisplayObject<'gc>> {
```
adrian17, line 2893: nit: AFAIK `try_F()` generally means "same as `F()`, but returns Result/Option instead of panic" - at least that's how I understand it from stdlib? So maybe some other name would be slightly better.

Maybe `try_resolve_level` -> `get_level()` and  `resolve_level` -> `get_or_create_level()`?

adrian17 (approved):

core/src/avm1/activation.rs
Dinnerbone: Good shout, renamed!

PR #11472, opened 2023-06-11
render: Add 'renderdoc' feature/module for manual frame capture

Aaron1011:
This feature is disabled by default. When enabled, you can use `ruffle_render::renderdoc::begin_frame_capture` and `ruffle_render::renderdoc::end_frame_capture` to manually trigger a RenderDoc frame capture (if Ruffle wasn't launched by RenderDoc, this logs an error).

This is very useful when debugging Stage3D/PixelBender bugs, as you can produce a capture containing only the relevant graphics calls.

### DIFF
--- a/render/Cargo.toml
+++ b/render/Cargo.toml
@@ -30,6 +30,11 @@ num-derive = "0.3"
 byteorder = "1.4"
 wgpu = { workspace = true, optional = true }
 
+# This crate has a `compile_error!` on apple platforms
+[target.'cfg(not(target_vendor = "apple"))'.dependencies.renderdoc]
+version = "0.11.0"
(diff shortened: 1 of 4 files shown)


render/Cargo.toml
```diff
@@ -36,4 +41,4 @@ approx = "0.5.1"
 [features]
 default = []
 tessellator = ["lyon"]
-web = ["wasm-bindgen"]
+web = ["wasm-bindgen"]
```
adrian17, line 44: nit: whitespace ;D

adrian17 (approved):

Dinnerbone (approved):

Issue #11460, opened 2023-06-10
Civilizations Wars - Soft locks after all huts are taken.


desertking:
### Describe the bug

https://www.newgrounds.com/portal/view/523882?emulate=flash - This game seems to work fine until you took over all the huts and stuff in the first level the level never ends and we can't proceed

There are also soft bugs like you can't drag & drop troops from more than one hut and some filters are missing. For gameplay video see:

https://www.youtube.com/watch?v=eaAA11VIi3g

### Expected behavior

To have the win screen after taken all and proceed

### Affected platform

Online demo

### Operating system

All

### Browser

All

Lord-McSweeney:
There is a trace in the console that said something like "`aGame.finished()`" (exact content unknown because I accidentally closed the tab) that appeared after capturing all the huts.

Aaron1011:
This is fixed on the latest `master`

PR #11476, opened 2023-06-11
avm2: Set named field in parent to null when child is removed

Aaron1011:
If a child named 'foo' is removed by the timeline (without having been previous added/removed from the timeline by ActionScript), then the 'foo' field in the parent will be set to null. This occurs even if the 'foo' field in the parent is not currently set to the child 'foo' (e.g. 'this.foo = someOtherObject' was executed by ActionScript).


Aaron1011:
Fixes https://github.com/ruffle-rs/ruffle/issues/11460

core/src/display_object/container.rs
```diff
@@ -597,13 +602,49 @@ impl<'gc> ChildContainer<'gc> {
...
+                let parent = child.parent().expect(
+                    "Parent must be removed *after* calling `remove_child_from_render_list`",
+                );
+                if child.has_explicit_name() {
+                    if let Avm2Value::Object(mut parent_obj) = parent.object2() {
+                        let mut activation = Avm2Activation::from_nothing(context.reborrow());
+                        parent_obj
+                            .set_public_property(child.name(), Avm2Value::Null, &mut activation)
+                            .unwrap_or_else(|err| {
+                                panic!(
```
adrian17, line 637: the sibling function that sets the child, doesn't panic here, it just logs and continues.  We probably don't want a panic here, since this can invoke a user setter that can throw.

core/src/display_object/container.rs
```diff
@@ -597,13 +602,49 @@ impl<'gc> ChildContainer<'gc> {
...
+            drop(this);
+
+            // Only set the parent's field to 'null' if the child was not placed/modified
+            // on the render list by AVM2 code.
+            if !child.placed_by_script() {
+                let parent = child.parent().expect(
+                    "Parent must be removed *after* calling `remove_child_from_render_list`",
+                );
+                if child.has_explicit_name() {
+                    if let Avm2Value::Object(mut parent_obj) = parent.object2() {
```
adrian17, line 632: (random thought: maybe we should put the setter and un-setter code in sibling functions next to each other, in the same file? But I don't have a good idea of which one would be the right one.)

core/src/display_object/container.rs
```diff
@@ -597,13 +602,56 @@ impl<'gc> ChildContainer<'gc> {
...
+                                    Avm2Value::Null,
+                                    &mut activation,
+                                );
+                                if let Err(e) = res {
+                                    tracing::error!("Failed to set child {} ({:?}) to null on parent obj {:?}: {:?}", child.name(), child, parent_obj, e);
+                                }
+                            }
+                            Err(e) => {
+                                tracing::error!("Failed to get current value of child {} ({:?}) on parent obj {:?}: {:?}", child.name(), child, parent_obj, e);
+                            }
```
Lord-McSweeney, line 650: The two error handlings should probably use `e.detailed_message` instead of `e`.

PR #11491, opened 2023-06-12
avm2: Add ignored test for subclass with 'overriding' field name

Aaron1011:
The Adobe Animate compiler rejects a subclass that contains a non-private field with the same name as a field in an ancestor class (e.g. 'pub var foo:String' in both the subclass and superclass).

Unfortunately, Flash Player accepts this code, and creates a distinct field for each class (even though they have the same namespace and name). I suspect that this is caused by the optimizer replacing the field accesses with internal slot accesses.

I've added an ignored test demonstrating this behavior - getting Ruffle to reproduce it will be tricky.


Lord-McSweeney:
Possibly the cause of #8851?

adrian17 (approved):
Looks good, I think. Later I can check if maybe my optimizer branch https://github.com/adrian17/ruffle/tree/avm2-peephole-optimizer-poc happens to fix it.
(as in, as-is it definitely won't fix it as I also need an extra commit to weaken trait validation; but without validation, it might achieve compatible behavior.)

Issue #2521, opened 2021-01-14
sub hunter not supported


gargle99:
https://www.flashgamesplayer.com/free/sub-hunter/play.html

n0samu:
Same problem as #8129 - this game uses MochiCrypt, and not all methods for it are implemented yet.

Also, this site is obnoxiously difficult to download files from. Here is the SWF file: [sub-hunter.zip](https://github.com/ruffle-rs/ruffle/files/9711999/sub-hunter.zip)

Lord-McSweeney:
Still not working as of latest nightly.

Lord-McSweeney:
With the latest PRs, this no longer happens, but there's an issue in `spark.primitives.supportClasses::GraphicElement/commitProperties()`:
```
      protected function commitProperties() : void
      {
         var _loc2_:mx.geom.Transform = null;
         var _loc1_:Boolean = false;
         if(this.displayObjectSharingMode != DisplayObjectSharingMode.USES_SHARED_OBJECT && Boolean(this.displayObject))
         {
            if(this.colorTransformChanged || this.displayObjectChanged)
            {
               this.colorTransformChanged = false;
               if(this._colorTransform)
               {
                  this.displayObject.transform.colorTransform = this._colorTransform;
               }
            }
            if(this.alphaChanged || this.displayObjectChanged)
            {
               this.alphaChanged = false;
               _loc2_ = this._transform as mx.geom.Transform; // mx.geom.Transform is undefined at this point, which throws an error.
```

Aaron1011:
With https://github.com/ruffle-rs/ruffle/pull/11494, https://github.com/ruffle-rs/ruffle/pull/11495, and https://github.com/ruffle-rs/ruffle/pull/11496, the level starts - but pressing Spacebar restarts the level immediately, instead of dropping a depth charge. There aren't any errors logged to the console when that happens.

Lord-McSweeney:
The only issue left is that Ruffle renders graphics in the SWF that are supposed to have a transparent background with a white background. This issue also happens in waflash.

PR #11494, opened 2023-06-12
avm2: Convert `has_class_in_chain` to use a `GcCell<Class>`

Aaron1011:
This lets us avoid looking up a `ClassObject` when performing parameter type coercion - we can instead use a `GcCell<Read>`. As a result, we no longer need all of the special logic for handling coercions during a script initializer.


adrian17:
I'm not sure if there's anything specific here that'd require a new test; this refactor should be decently covered by tests already, right?

Aaron1011:
Yeah, we have existing tests around the coercion logic and class lookup.

Lord-McSweeney:
You stated that this progresses #2521- could you add a test for the issue it fixes in the SWF?

Aaron1011:
I was mistaken - the progression is actually due to my PR implementing more methods in Transform, which lets the verifier pass when a class overrides them. Previously, this error was getting swallowed by the special-cased coercion logic.

adrian17 (approved):
Still not a fan of the diff, but let's just leave it for future us.

adrian17:
note: on Discord I've found an edge case regarding Vectors. Avmplus does some fancier stuff there, but I think our solution for now could just copy `resolve_type`:
```rs
        if !type_name.params().is_empty() {
            let mut param_types = Vec::with_capacity(type_name.params().len());

            for param in type_name.params() {
                param_types.push(match self.resolve_type(param)? {
                    Some(o) => Value::Object(o.into()),
                    None => Value::Null,
                });
            }

            return Ok(Some(class.apply(self, &param_types[..])?));
        }
```
just with resolve_type replaced by domain class lookup (...and maybe no silly Vecs? :D )

core/src/avm2/property.rs
```diff
@@ -64,56 +65,24 @@ impl<'gc> PropertyClass<'gc> {
...
-                let class = match outcome {
-                    ResolveOutcome::Class(class) => {
+                if name.is_any_name() {
+                    *self = PropertyClass::Any;
+                    (None, true)
+                } else {
+                    // Note - we look up the class in the domain by name, which allows us to look up private classes.
+                    // This also has the advantage of letting us coerce to a class while the `ClassObject`
+                    // is still being constructed (since the `Class` will already exist in the domain)
+                    let domain = unit.map_or(activation.avm2().stage_domain, |u| u.domain());
```
adrian17, line 75: correct me if I'm wrong: I think `unit` can only be `None` if this comes from a builtin Rust-defined trait, so I think this should use a builtin domain instead?

adrian17 (approved):

core/src/avm2/property.rs
Aaron1011: You're right - I've changed it.

PR #11501, opened 2023-06-12
core: Add options to set and force the stage alignment

n0samu:
This is basically the same as #9406, except for `salign` instead of `scale`. It can be annoying when games position themselves in a corner of the screen - this allows users to force them to be centered (or move them to any corner, if for some reason they want to).

**Before**
image

**After**
image

### DIFF
--- a/core/src/display_object/stage.rs
+++ b/core/src/display_object/stage.rs
@@ -95,6 +95,9 @@ pub struct StageData<'gc> {
     /// The alignment of the stage.
     align: StageAlign,
 
+    /// Whether to prevent movies from the changing the stage alignment
+    forced_align: bool,
+
     /// Whether or not a RENDER event should be dispatched on the next render
     invalidated: bool,
 
@@ -168,6 +171,7 @@ impl<'gc> Stage<'gc> {
                 },
                 invalidated: false,
                 align: Default::default(),
+                forced_align: false,
                 use_bitmap_downsampling: false,
                 view_bounds: Default::default(),
                 window_mode: Default::default(),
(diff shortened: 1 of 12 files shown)


core/src/display_object/stage.rs
```diff
@@ -1034,7 +1050,7 @@ bitflags! {
     ///
     /// This is a bitflags instead of an enum to mimic Flash Player behavior.
     /// You can theoretically have both TOP and BOTTOM bits set, for example.
-    #[derive(Clone, Copy, Default, Collect)]
+    #[derive(Clone, Collect, Copy, Debug, Default, Eq, PartialEq)]
```
adrian17, line 1053: nit: we usually put Collect on end, after stdlib derives. I know there are exceptions already, but this way it's slightly neater ;D

core/src/display_object/stage.rs
```diff
@@ -386,8 +390,20 @@ impl<'gc> Stage<'gc> {
     /// Set the stage alignment.
     /// This only has an effect if the scale mode is not `StageScaleMode::ExactFit`.
     pub fn set_align(self, context: &mut UpdateContext<'_, 'gc>, align: StageAlign) {
-        self.0.write(context.gc_context).align = align;
-        self.build_matrices(context);
+        if !self.forced_align() {
```
adrian17, line 393: just to be sure: this can be observed from AS2/3?
As in, if it's in forced align mode:
```as
stage.align = "B";
trace(stage.align); // does this trace "B" or the preset forced alignment?
```

n0samu: Yes it prevents setting the property so it will remain whatever it was locked to. And forceScale behaves the same way.

core/src/display_object/stage.rs
n0samu: Ah I didn't know about that, I'll move it back to the end then

Dinnerbone (approved):

Issue #6144, opened 2022-01-26
`gotoAndPlay()` frame off-by-one sometimes


torokati44:
A few z0r loops ([916](https://z0r.de/L/z0r-de_916.swf), [1115](https://z0r.de/L/z0r-de_1115.swf), [1875](https://z0r.de/L/z0r-de_1875.swf)) have this (or something similar) `DoAction` script in them:

![image](https://user-images.githubusercontent.com/288816/151079752-7db2da6e-747c-4e21-883a-1dd6e10a129b.png)

Frame 3 (or 2?) would start playing their sounds - but they are mute in Ruffle.

Removing this `+ 1` here does cause their sounds to play: https://github.com/ruffle-rs/ruffle/blob/354d7f0ce621a38bd238b98af5f88c22762fbfd9/core/src/avm1/activation.rs#L1303

...but then they either get a white frame flashing after each repetition, and/or their sounds are started again, playing more and more instances, every time they reset.

Looks like there is some more complicated (perhaps accidental) logic behind this 1-frame offset.

adrian17:
@kmeisthax I think this is AS2, but can you please check if this could be in any vicinity of your AVM2 frame ordering changes?

kmeisthax:
Most of the frame-ordering fixes regarding `gotoAndXyz` target gotos in the `Enter` and `Construct` frame phases, which don't exist in AVM1. I was also very careful to avoid making *any* frame ordering changes that would touch AVM1, since anything that did broke our test movies horribly.

FWIW the frame ordering PR does nothing to fix this movie.

Looking at your initial screenshot, the P-Code is `GotoFrame 2` but the disassembly is `gotoAndPlay(3)`. FWIW the SWF file format spec does not state if `ActionGotoFrame` is supposed to be 0-indexed or 1-indexed, but it appears both us and JPEXS are aligned on it being indexed from 1.

In the specific case of `z0r-de_916.swf`, frame 2 contains a `StartSound` tag which is skipped by playing frame 3. My impression was that non-streaming sounds don't get played by gotos, but streaming sounds are supposed to be rewound or fast-forwarded as necessary. In Ruffle, `run_goto` will stop streaming sounds, but it also runs the target frame, which is supposed to have `StartSound` or `SoundStreamBlock` tags we can work with.

However, this doesn't work here, because...

 * The movie is jumping one past the `StartSound` tag that kicks things off
[...]

PR #11508, opened 2023-06-12
avm1: Correct `WaitForFrame` and `WaitForFrame2` actions

Toad06:
This PR fixes the following cases:

`WaitForFrame` action:
- `ifFrameLoaded(n)` (where n is an integer equals to `_framesloaded`) evaluates to true.
- Any frame number above 16000 causes this action to evaluate to false.

`WaitForFrame2` action:
- `ifFrameLoaded(_framesloaded)` and `ifFrameLoaded(_framesloaded + 1)` evaluate to true.
- `ifFrameLoaded(2147483647 + 1)` evaluates to false.

---

I added two separate tests for these actions. A test already existed for `WaitForFrame2` but I replaced it because it didn't seem to test the correct values sometimes, for instance:

```
ifFrameLoaded(0) { trace("/// frame -43 marked as loaded"); }
ifFrameLoaded(NaN + 0) { trace("/// frame NaN marked as loaded"); }
```


core/src/avm1/activation.rs
```diff
@@ -2305,16 +2305,23 @@ impl<'a, 'gc> Activation<'a, 'gc> {
         action: WaitForFrame,
         r: &mut Reader<'_>,
     ) -> Result<FrameControl<'gc>, Error<'gc>> {
-        let loaded = self
-            .target_clip()
-            .and_then(|dobj| dobj.as_movie_clip())
-            .map(|mc| mc.frames_loaded() > min(action.frame, mc.total_frames() - 1))
-            .unwrap_or(true);
+        let frame_num = action.frame;
+        let loaded = if frame_num > 16000 {
```
Toad06, line 2309: I guess it would make sense to define this value as a constant. But not sure what could be best place for this.

torokati44:
I can't test right now, but does this perhaps fix some more of https://github.com/ruffle-rs/ruffle/issues/6144?

Toad06:
Unfortunately it doesn't, Ruffle still loads more frames than Flash Player on startup, which causes `ifFrameLoaded(n)` to evaluate differently.

Dinnerbone (approved):

adrian17 (approved):
Could that mean that in reality, `_framesloaded` is the one that lies by 1? As in, it sounds more likely that `_framesloaded` reports a lower number of loaded frames than the actual number, than that `ifFrameLoaded` lies about a frame being loaded that's not actually loaded.
(at least that's how I understand it)

Toad06:
> Could that mean that in reality, `_framesloaded` is the one that lies by 1? As in, it sounds more likely that `_framesloaded` reports a lower number of loaded frames than the actual number, than that `ifFrameLoaded` lies about a frame being loaded that's not actually loaded. (at least that's how I understand it)

This looks accurate. I added traces to z0r 916 ([edited file here](https://github.com/ruffle-rs/ruffle/files/11750517/z916.zip)) and this is the output I got from the Flash Player projector:

Frame 1:
> _framesloaded: 0
>_currentframe: 1

Frame 42:
> _framesloaded: 42
> _currentframe: 42

Issue #11510, opened 2023-06-13
avm2: Type coercion to Array is not correctly implemented


n0samu:
### Describe the bug

In ActionScript 3, `Array(value)` seems to coerce `value` to an Array by calling the `Array()` constructor with `value` as the argument. This is not implemented in Ruffle. Consider this simple AS3 code:
```as
var test = Array(undefined);
trace(test instanceof Array);
trace(test.length);
trace(test[0]);
```
Ruffle traces:
```
null
false
```
then throws this error:
```
Error occured when running AVM2 frame script: TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: length)
```

### Expected behavior

Flash traces:
```

true
1
undefined
```
`undefined` is successfully coerced into the array `[undefined]`. I also tried `Array(value)` with many other values (numbers, ints, strings, null, etc) and I found that the behavior is always exactly the same as calling `new Array(value)` (which Ruffle has properly implemented already). So it seems clear that coercing to an Array just calls the Array constructor in Flash.

### Affected platform

Desktop app

### Operating system

Windows 10

MartySVK:
I noticed this error in Hockey Legends or Halloween Basketball Legends

adrian17:
Fixed in https://github.com/ruffle-rs/ruffle/pull/11517 .

PR #11519, opened 2023-06-13
avm2: Throw AVM error in Array constructor

n0samu:
Just a tiny fixup for something I noticed while looking into #11510


Lord-McSweeney (approved):
LGTM, I don't think a test is necessary for something this minor.

Dinnerbone (approved):

adrian17 (approved):

PR #11520, opened 2023-06-13
avm2: Add void type

adrian17:
Still gotta add a test for this.
Note that this as-is lets you do much more than intended, like `new void()` or `var T = void`, which FP doesn't support. Can't really fix this without the class refactor :c  
I don't think this is a major issue, but if you think this is a major risk after all, then we should go with some more targeted hack in type-related opcodes.

### DIFF
--- a/core/src/avm2/globals.rs
+++ b/core/src/avm2/globals.rs
@@ -36,6 +36,7 @@ mod string;
 mod toplevel;
 mod r#uint;
 mod vector;
+mod void;
 mod xml;
 mod xml_list;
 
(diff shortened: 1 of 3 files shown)


core/src/avm2/globals.rs
```diff
@@ -498,6 +501,10 @@ pub fn load_player_globals<'gc>(
     );
     avm2_system_class!(array, activation, array::create_class(activation), script);
 
+    // TODO: this should _not_ be exposed as a ClassObject, getDefinitionByName etc.
+    // it should only be visible as an type for typecheck/cast purposes.
```
Lord-McSweeney, line 505: It might be worth pointing out the additional limitation that this errors if used for a purpose other than as a return type (you cannot coerce to void using the coerce opcode, for example)

adrian17: For the record, it _can_ be used for the `coerce` opcode (it's not much useful ofc - but it throws a runtime exception instead of an early VerifyError, so it's a perfectly normal coercion failure).

adrian17:
Rebased, un-drafted (without changes).

Aaron1011 (approved):

Issue #11386, opened 2023-06-05
String to XMLList cast fails


barisyild:
### Describe the bug

`Code:`

```as
var map:XMLList = XMLList('<?xml version="1.0"?><map></map>'); //Cast String to XMLList
trace(getQualifiedClassName(map)); //XMLList
```
`*Ruffle Only* Exception:`
```ex
Error occured when running AVM2 frame script: TypeError: Error #1034: Type Coercion failed: cannot convert "<?xml version="1.0"?><map></map>" to XMLList.
```

The above code can be cast in flash, but not ruffle.rs.

### Expected behavior

String must be cast to the XMLList type.

cast and creating new instance should have the same behavior and create a new XMLList.

```as
var map:XMLList = new XMLList('<?xml version="1.0"?><map></map>'); //Create XMLList from String
trace(getQualifiedClassName(map)); //XMLList
```

### Affected platform

Online demo

### Operating system

Windows 10

### Browser

Google Chrome

Aaron1011:
This is a simple fix - we need to add `[Ruffle(CallHandler)]` to the `XMLList` class (like we have for `XML`), and make the `call_handler` method call the `XMLList` constructor.

PR #11522, opened 2023-06-13
avm2: Implement call handler for XMLList

Aaron1011:
This delegates to the XMLList constructor, just as the XML call handler delegates to the XML constructor.


adrian17 (approved):

n0samu (approved):
Should fix #11386

PR #11526, opened 2023-06-13
wgpu: Cache Program3D bind group layout

Aaron1011:
The bind group layout only depends on the texture registers (and 2D/cubemap type) accessed by the fragment shader, not on the runtime texture bound with Context3D. This means that we can build and cache it when we compile the AGAL program to a Naga module.

Since the bind group layout is used for the overall pipeline, I've refactored the shader caching code into `ShaderPairAgal`, which holds both the vertex and fragment shader bytecode, and compiles both in the `compile` function.

### DIFF
--- a/core/src/avm2/object/context3d_object.rs
+++ b/core/src/avm2/object/context3d_object.rs
@@ -249,9 +249,8 @@ impl<'gc> Context3DObject<'gc> {
             .unwrap()
             .process_command(
                 Context3DCommand::UploadShaders {
-                    vertex_shader: program.vertex_shader_handle(),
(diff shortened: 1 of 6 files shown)


adrian17:
What content does this affect / which swfs are the best for observing the improvement?

Aaron1011:
I haven't been able to observe a noticeable improvement on desktop yet. However, in Sniper Team, the wgpu function `deduplicate_bind_group_layout` no longer shows up in a `perf record` profile with this PR.

Dinnerbone (approved):

PR #11544, opened 2023-06-14
avm2: Add error number in error message for BMD.drawWithQuality

Lord-McSweeney:


adrian17 (approved):
Ideally we should probably just be doing this in the constructor :D

Issue #11540, opened 2023-06-14
Error on https://uploads.ungrounded.net/ruffle_wrapper/ruffleembed.html


LimonGamesXD99:
# Error Info
Error name: Error
Error message: panicked at 'index out of bounds: the len is 0 but the index is 0', core/src/avm2/globals/flash/utils.rs:196:15
Error stack:
```
Error: panicked at 'index out of bounds: the len is 0 but the index is 0', core/src/avm2/globals/flash/utils.rs:196:15
    at U.n.wbg.__wbg_new_87297f22973157c8 (https://uploads.ungrounded.net/ruffle_wrapper/lib/core.ruffle.671e8308f154108396e3.js:1:51160)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::ha4de13118e9dd6b4 (wasm://wasm/02bfd5fe:wasm-function[8597]:0x827e52)
    at std::panicking::rust_panic_with_hook::h85b7f6628c291e12 (wasm://wasm/02bfd5fe:wasm-function[5581]:0x749d4c)
    at core::panicking::panic_fmt::hf5c4cd929d4aaa9e (wasm://wasm/02bfd5fe:wasm-function[8463]:0x823aa2)
    at core::panicking::panic_bounds_check::h0cfce459260a0a82 (wasm://wasm/02bfd5fe:wasm-function[6757]:0x7a2a92)
    at ruffle_core::avm2::globals::flash::utils::get_qualified_class_name::h3f4a78f0cba28c13 (wasm://wasm/02bfd5fe:wasm-function[2673]:0x5afa68)
    at ruffle_core::avm2::function::Executable::exec::h1c8f15e267a622e4 (wasm://wasm/02bfd5fe:wasm-function[810]:0x2ae452)
    at ruffle_core::avm2::object::class_object::ClassObject::run_class_initializer::h8a7e1242a959f795 (wasm://wasm/02bfd5fe:wasm-function[3678]:0x6850f6)
    at ruffle_core::avm2::object::class_object::ClassObject::into_finished_class::h2b50245ff64847dc (wasm://wasm/02bfd5fe:wasm-function[2574]:0x5980bd)
    at ruffle_core::avm2::object::class_object::ClassObject::from_class::h1bfe60530eca9ee0 (wasm://wasm/02bfd5fe:wasm-function[2964]:0x5f3a4e)
```
AVM2 stack:
```
[...]

PR #11545, opened 2023-06-14
avm2: Use a separate domain for playerglobals

Aaron1011:
We create a separate child domain associated with the Stage, which has its own child domain for the root MovieClip.

This prevents ActionScript from loading classes into the special playerglobals domain (Domain.parentDomain is modified to return null instead of the playerglobals domain when applicable), so the native method lookup logic will never run for user code.

Fixes https://github.com/ruffle-rs/ruffle/issues/11540


core/src/avm2/globals/flash/system/application_domain.rs
```diff
@@ -52,6 +52,9 @@ pub fn get_parent_domain<'gc>(
 ) -> Result<Value<'gc>, Error<'gc>> {
     if let Some(appdomain) = this.and_then(|this| this.as_application_domain()) {
         if let Some(parent_domain) = appdomain.parent_domain() {
+            if parent_domain.is_playerglobals_domain(activation) {
+                return Ok(Value::Null);
```
Lord-McSweeney, line 56: Is there a test for this specific part?

Aaron1011: We have tests for `stage.loaderInfo.applicationDomain`,but they're unaffected by this PR (since we previously had no parent, and we how have a hidden parent)

core/src/avm2/globals.rs
```diff
@@ -743,5 +743,9 @@ fn load_playerglobal<'gc>(
 
     // Domain memory must be initialized after playerglobals is loaded because it relies on ByteArray.
     domain.init_default_domain_memory(activation)?;
```
adrian17, line 745: do we still need domain memory on the builtins' domain?

adrian17 (approved):
LGTM, but yeah a test would be nice ;D

core/src/avm2/globals.rs
Aaron1011: We don't current use it, but I think we should keep it so that the playerglobals domain is co distent with other domains (all other domains will have that field initialized).

adrian17: OK 👍

core/src/player.rs
```diff
@@ -377,19 +377,14 @@ impl Player {
...
-                Avm2Activation::from_domain(global_activation.context.reborrow(), domain);
+            let stage_domain = context.avm2.stage_domain();
+            let mut activation = Avm2Activation::from_domain(context.reborrow(), stage_domain);
 
             activation
                 .context
                 .library
                 .library_for_movie_mut(activation.context.swf.clone())
-                .set_avm2_domain(domain);
+                .set_avm2_domain(stage_domain);
```
adrian17, line 387: this sounds slightly weird. In the scenario where we overwrite the root movie:
- the SWF loads with its own domain
- while the SWF loads, it sets all its classes into its domain
- when we make this SWF's root movie into the stage root movie, we overwrite its domain with stage domain
- ergo, all the definitions of the child movie are forgotten?
That doesn't sound right.

Are you sure that it's not the other way around - the child movie's domain becoming the new stage domain (and the old stage domain remaining as accessible parent)? I don't know if that's what FP does, but it sounds like this way less content would break.

Aaron1011: We never delete things from a domain, or overwrite the overall domain. The linkage is only from a movie to a domain, so replacing the root movie will continue to use the existing stage domain (and all of the loaded classes will stay loaded).

Dinnerbone (approved):

PR #11590, opened 2023-06-17
web: Allow setting `allowScriptAccess` on global config

Toad06:


relrelb (approved):

Dinnerbone:
We should probably (as a followup, doesn't need to block this PR) change it from a boolean but to Flash's 3 options (the important difference is sameDomain) as this invites exploitation a bit, but yay it's global now!

Dinnerbone (approved):

PR #11596, opened 2023-06-17
tests: Increased tolerance of stage3d tests

Dinnerbone:
This makes the tests pass on my machine

### DIFF
--- a/tests/tests/swfs/avm2/stage3d_fractal/test.toml
+++ b/tests/tests/swfs/avm2/stage3d_fractal/test.toml
@@ -1,7 +1,7 @@
 num_frames = 2
 
 [image_comparison]
-tolerance = 1
+tolerance = 2
 max_outliers = 25
 
 [player_options]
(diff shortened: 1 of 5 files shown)


Aaron1011 (approved):

torokati44:
I still get this failure:
```
---- avm2/stage3d_raytrace ----
Number of outliers (6) is bigger than allowed limit of 3. Max difference is 76
```

Dinnerbone:
@torokati44 if you increase the tolerance to 2 there, is it enough?

torokati44:
@Dinnerbone Nope.

torokati44:
This works though:
```toml
[image_comparison]
tolerance = 1
max_outliers = 6
```

Dinnerbone:
> @Dinnerbone Nope. The max difference of 76 hints at this.

Not really, it's easily misunderstood. There can be a bunch of 2-delta pixels and a single one of 76. The outliers should catch the crazy high value, and the rest we allow with a tolerance.

torokati44:
Oh, I see... That's really not intuitive...

Dinnerbone:
Nope it's not but that's statistics 😅

torokati44:
Couldn't find any lower limits than the above 1 - 6...

Dinnerbone:
@torokati44 all good now?

torokati44:
I was able to confirm now, all the tests (including visual) are indeed passing on my machine with these tolerances.

PR #11599, opened 2023-06-18
Update `zh-CN/messages.json`

steve02081504:
*Only a minor amendment to the translation

### DIFF
--- a/web/packages/extension/assets/_locales/zh_CN/messages.json
+++ b/web/packages/extension/assets/_locales/zh_CN/messages.json
@@ -87,7 +87,7 @@
         "message": "设置"
     },
     "description": {
-        "message": "将 Flash 放回网页中原本的位置。"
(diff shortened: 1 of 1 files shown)


n0samu:
Thanks for catching that! 👍

Lord-McSweeney:
Since these are automatically updated from Crowdin, won't the translation be overridden by the one stored on Crowdin?

Dinnerbone:
> Since these are automatically updated from Crowdin, won't the translation be overridden by the one stored on Crowdin?

Yes. Changes need to be done via crowdin, not here.

steve02081504:
> > Since these are automatically updated from Crowdin, won't the translation be overridden by the one stored on Crowdin?
> 
> Yes. Changes need to be done via crowdin, not here.

Thank you all!
I voted on Crowdin for what I think is a more Sindarin translation, so I think it's OK to close this pr.☺

PR #5251, opened 2021-09-08
avm1: Remove `TObject::get_local` in favor of `TObject::get_local_stored`

relrelb:
`get_local` is basically equivalent to `get_local_stored` that also handles virtual getters. So instead handle virtual getters in `search_prototype`. This allows to inline the `get_local_sub` helper methods into the implementations of `get_local_stored`.
The remaining usages of `get_local` were not exactly correct as they all should ignore virtual getters. This change solves this as well by using `get_local_stored` that ignores virtual getters.

PR #11631, opened 2023-06-19
avm1: Correct `matrix::object_to_matrix_or_default`

### LINKED ISSUE #3721: Onda Educa - Missing layer of objects
**Describe the bug**
Without going into too much detail, this is likely another example of #58
The puzzle activity in Animacuentos calls for the puzzle pieces (evidently on their own layer) to be sitting to the right of the empty puzzle board. They are not present when running in Ruffle. Also, the "Mezclar" button is supposed to shuffle the strewn positions of those puzzle pieces.  

**Expected behavior**
Layer with puzzles pieces should be present and able to be shuffled ("mixed").

**Is the problem with the Ruffle desktop app, extension, or self-hosted version?**
Desktop and web builds - 3/20/21 nightlies
.
**What platform are you using?**
Windows 10, Nvidia GPU, Chromium based browsers and Firefox.
toto_memoria
toto_memoria_ruffle
[puzle_micuento.zip](https://github.com/ruffle-rs/ruffle/files/6177033/puzle_micuento.zip)

Toad06:
The comment for this function says that it
> returns the identity matrix if any of the `a`, `b`, `c`, `d`, or `tx` properties do not exist

but that wasn't the case.

Fixes #3721 which calls `pz[n].beginBitmapFill(tmp2);` without specifying an optional matrix argument.

### DIFF
--- a/core/src/avm1/globals/matrix.rs
+++ b/core/src/avm1/globals/matrix.rs
@@ -107,41 +107,30 @@ pub fn object_to_matrix<'gc>(
 
 /// Returns a `Matrix` with the properties from `object`.
 ///
-/// Returns the identity matrix if any of the `a`, `b`, `c`, `d`, or `tx` properties do not exist.
+/// Returns the identity matrix if any of the `a`, `b`, `c`, `d`, `tx` or `ty` properties do not exist.
 pub fn object_to_matrix_or_default<'gc>(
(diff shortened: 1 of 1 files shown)


relrelb (approved):
Oops, my bad (I regressed this in #5251). Thank you!

Issue #10455, opened 2023-03-30
Some parts of Out of Wind show up with a black background


Lord-McSweeney:
### Describe the bug

SWF: http://i.notdoppler.com/files/outofwind.swf (https://www.notdoppler.com/outofwind.php)
The backgrounds should be transparent, not black. This is visible at the menu and when clicking play- after the latter, buttons in the top-right are obscured.

### Expected behavior

Lightspark renders them semi-correctly (though it has massive visual issues with the game)- see https://github.com/lightspark/lightspark/issues/1056.

### Affected platform

Browser's extension

### Operating system

Linux

### Browser

Chrome 110

n0samu:
I think I figured out a small piece of this puzzle, at least. Let's focus on just sprite 608, which Ruffle renders incorrectly: [sprite608.zip](https://github.com/ruffle-rs/ruffle/files/11108694/sprite608.zip)
![image](https://user-images.githubusercontent.com/71368227/228754583-719f6dd4-abb5-456d-9e45-e8a5979e8061.png)

The incorrect rendering is actually caused by the windmill sprite placed within it. The sprite consists of a PlaceObject 2 tag and a PlaceObject 3 tag which both place the same character (479). The PlaceObject3 tag uses a ColorTransform and the Overlay blend mode to enhance the colors of the original sprite. But similarly to #9304, the Overlay blend mode is rendered incorrectly by Ruffle. As seen in the attached `sprite608_windmillmod.swf`, if we change the blend mode to Normal, the problem goes away:
![image](https://user-images.githubusercontent.com/71368227/228756655-39fa8d21-7b07-42f8-adc9-0fbc1bedb37a.png)

n0samu:
The wrong rendering of sprite 608 started with commit 8db8e4669adb6bc5498b19af66db78a10ce0cc50 of #9096. Same regression as #9304.

n0samu:
The black background issue on the main menu is actually caused by a function inside the game that rasterizes sprites.
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/2b527b2d-f1b8-41fc-83ce-7e3fb9a6dbf9)
The function is in `menu.page.BasePage`:
```as
public function RasterizeChildren() : void
{
    var _loc1_:Sprite = null;
    var _loc2_:int = 0;
    while(_loc2_ < this.fVisual.numChildren)
    {
        if(this.fVisual.getChildAt(_loc2_) is Sprite)
        {
[...]

Issue #9186, opened 2023-01-16
Performance issues in Bloons Super Monkey


Lord-McSweeney:
### Describe the bug

Archived SWF available at http://web.archive.org/web/20201231062941/https://chat.kongregate.com/gamez/0007/2514/live/bloonssupermonkey.swf?kongregate_game_version=1266877460.

When getting the upgrade "plasma tentacle", visible lag will start.

### Expected behavior

The game runs smoothly with plasma tentacle upgrade.

### Affected platform

Browser's extension

### Operating system

Linux

### Browser

Chrome 109

n0samu:
Now I'm really confused because I'm not seeing any lag when buying the plasma tentacle on web builds of Ruffle. I'm only seeing it on desktop, so I assumed it was an issue with blend modes being inefficient. Are you sure this is happening for you on web?

n0samu:
Further discussion on Discord clarified that the testing had likely been done with today's nightly, which switched to wgpu-webgl. I didn't realize the new nightly was already out at the time this issue was made.

PR #11638, opened 2023-06-19
Implement cacheAsBitmap

Dinnerbone:
This is very hard to test programmatically by nature, but I've added what I could.

In cases like heartcore, where they cacheAsBitmap and then invalidate *every single frame* - we're about 50% slower. Flash itself always warns not to do this as it's likely slower, and so that holds true for us too.

In cases like rd_flying, we're about 92% faster!

image


core/src/display_object.rs
```diff
@@ -1180,6 +1429,7 @@ pub trait TDisplayObject<'gc>:
     /// Values other than the default `BlendMode::Normal` implicitly cause cache-as-bitmap behavior.
     fn set_blend_mode(&self, gc_context: MutationContext<'gc, '_>, value: BlendMode) {
         self.base_mut(gc_context).set_blend_mode(value);
+        self.invalidate_cached_bitmap(gc_context);
```
Lord-McSweeney, line 1432: An above comment references a bug in Flash where the bitmap is not invalidated when the blend mode is changed- would it be helpful to add that as a comment here (in case of some obscure edge case)?

render/canvas/src/lib.rs
```diff
@@ -451,7 +456,15 @@ impl RenderBackend for WebCanvasRenderBackend {
...
 
-    fn submit_frame(&mut self, clear: Color, commands: CommandList) {
+    fn submit_frame(
+        &mut self,
+        clear: Color,
+        commands: CommandList,
+        cache_entries: Vec<BitmapCacheEntry>,
+    ) {
+        if !cache_entries.is_empty() {
+            panic!("Bitmap caching is unavailable on the canvas backend");
```
Lord-McSweeney, line 466: Should log an error instead of panicking?

Dinnerbone: No, we shouldn't be able to get here and if we did a developer messed up (and will get this during development). You'll get graphical corruption if you proceed anyway, because it'll be using a texture of uninitialized values.

n0samu:
I'm seeing some regressions in Bloons Super Monkey:
- Some text is cut off now, specifically the Resume button in the Pause menu and the Next Wave button in the Wave Passed menu.
image
image
- The bloon progress meter on the right sidebar visually glitches as it fills up.
image
image

Also, the performance drop after buying the plasma tentacle is still happening (#9186) although it doesn't seem to have gotten any worse, at least.

n0samu:
This also causes a subtle regression in the rendering of sprite 608 of Out of Wind (https://github.com/ruffle-rs/ruffle/issues/10455#issuecomment-1489803032): the tiny wind turbine sprites now display with black backgrounds. You can see this in both the original sprite SWF and the "windmillmod" version:
image
image

SN902:
text is cut off in [433935_mpnew202c.zip](https://github.com/ruffle-rs/ruffle/files/11795516/433935_mpnew202c.zip)
image

n0samu:
This also causes a regression in [Bloons 2](http://www.notdoppler.com/bloons2.php)'s map screen:
ruffle_bitmapcache_bloons2

SN902:
[super_mario_world.zip](https://github.com/ruffle-rs/ruffle/files/11795993/super_mario_world.zip)

```thread 'main' panicked at 'already borrowed: BorrowMutError', core\src\display_object\movie_clip.rs:2361:28```

File: core/src/display_object.rs [hunk 3]
```diff
@@ -557,6 +673,97 @@ pub fn render_base<'gc>(this: DisplayObject<'gc>, context: &mut RenderContext<'_
         None
     };
 
+    let mut cache_info: Option<(BitmapHandle, Option<(u32, u32)>)> = None;
+    let base_transform = context.transform_stack.transform();
+    let bounds: Rectangle<Twips> = this.bounds_with_transform(&base_transform.matrix);
```
adrian17: IIRC in some edge cases I recall seeing `bounds_with_transform` on profiles (as it's tree-recursive), so maybe we could avoid calling it here in the common no-cache case?

Dinnerbone: It's already guarded on the later commits

Dinnerbone:
Text cutoff issues should be fixed, but bloons progress indicator + map isn't yet

n0samu:
With #11674 and #11649 merged in, the only remaining unresolved problem is the bloon progress meter in Bloons Super Monkey.

n0samu:
Looking at the bloon meter problem, here are 2 more things I noticed:
1. Elements of the meter shift up and down a little bit even in Flash Player and on current master. But the behavior on current master seems to be correct, whereas with this PR the wrong positioning is far more noticeable and persistent. For example, this never happens in FP or current master:
image
2. There is a related, less subtle bug that I somehow didn't catch before. The red bloon is supposed to slowly fill with the bronze color, but in this PR the red bloon doesn't change appearance at all until the bronze meter is completely full, at which point the meter changes to bronze all at once.

Dinnerbone:
All bugs should be fixed now. I'll address some reviews (from here + discord) later today.

Dinnerbone:
Everything should be fixed up now!

moulins:
Looks good to me!

Aaron1011 (approved):

Issue #11659, opened 2023-06-21
Dead Metal: Stuck on blank screen


n0samu:
### Describe the bug

Link to SWF: https://www.andkon.com/arcade/shooter/deadmetalupdated/deadmetalupdated.swf

[Dead Metal](https://www.andkon.com/arcade/shooter/deadmetalupdated/) is stuck on a blank screen and does not load. The error is:
```
ERROR run_frame:run_all_phases_avm2: ruffle_core::display_object::movie_clip: Got "TypeError: Error #1009: Cannot access a property or method of a null object reference.
        at MethodInfo-17()
        at Array/forEach()
        at PlayerProfile()
        at Main()" when constructing AVM2 side of movie clip of type Main
```

### Expected behavior

The game should load.

### Affected platform

Desktop app

### Operating system

Windows 10

Lord-McSweeney:
This gets into the game now, but nothing spawns.

n0samu:
Confirmed. Couldn't find the cause either 😦

PR #11673, opened 2023-06-21
avm2+tests: Fix Array.forEach with non-object this parameter, add tests

Lord-McSweeney:
Progresses #11659, though it needs a spoof url to work correctly, and for some reason, nothing is visible in the level.


Aaron1011:
Is this really specific to Array.forEach? I think this needs to apply to all unbound function invocation (e.g. standalone functions registered as event listeners or invoked manually from AS)

Lord-McSweeney:
Superseded by #11874.

Issue #11140, opened 2023-05-21
BlendMode.MULTIPLY does not handle alpha blending correctly


Aaron1011:
I've created a test case demonstrating the bug: https://github.com/Aaron1011/ruffle/tree/blend-mode-multiply-bug

Under Ruffle, this renders a large black square. Under Flash Player, it renders a small red square.

n0samu:
This was regressed by #9096 - before that the square was the correct size but an incorrect color.
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/36fcdb30-9660-4c18-b667-2c36500817ff)

And strangely enough, it rendered with the correct color before the initial blend mode implementation (#8262):
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/542f27c0-95e9-4612-bbf6-0808a00bac76)

So it seems like Flash is ignoring the blend mode completely. If I remove the "multiply" argument from the `draw()` call, the rendering in Flash is exactly the same.

Issue #11184, opened 2023-05-23
Black rectangles and `BlendMode::Multiply` bug.


iwannabethedev:
(Creating an issue based on investigations in Discord).

Possibly relevant issues:
- #10455 
- #10884 
- #9304 

Possibly relevant PRs:
- #9096 (as I understand the comments at Discord , the commit before that PR gave a different gray error, and that PR gives the black rectangle error). The old comment `// Can't be trivial, 0 alpha is special case` might be relevant.

Lord-McSweeney:
Possibly relevant: #11140

iwannabethedev:
@Aaron1011 I created this issue from the Discord investigations that you and others did, to avoid it being forgotten about.

Does it look accurate? Anything missing? Would it make sense to link to the branch that you mentioned to help debug the specific issue that a user had?

iwannabethedev:
CC @Dinnerbone .

Issue #10884, opened 2023-04-30
X-Girl: Black screen regression


n0samu:
### Describe the bug

Link to SWF: https://www.dolldivine.com/games/x-girl-forruffle.swf

[X-Girl: Comic Mutant Character Creator](https://www.dolldivine.com/x-girl) worked in early January, but now the screen displays mostly blank after clicking the Play button:
![image](https://user-images.githubusercontent.com/71368227/235332953-c68a3d3a-1de4-44c8-8b65-d75080488e3d.png)

I've bisected this regression to commit 8db8e4669adb6bc5498b19af66db78a10ce0cc50 from #9096.

### Expected behavior

Here's how the game looks in Nightly 2023-01-12:
![image](https://user-images.githubusercontent.com/71368227/235332996-e9f1f379-4c13-4c4b-90fb-63e79f0f9841.png)

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

Reported by @olamcbola on Discord.

n0samu:
This is actually the same regression as #9304.

n0samu:
Fixed by #11674

Issue #9304, opened 2023-01-25
Dinogeddon: Colors are incorrect on last frame 


n0samu:
### Describe the bug

Link to SWF: https://www.dolldivine.com/games/dinogeddon-premNoGallery.swf

In [Dinogeddon](https://www.dolldivine.com/dinogeddon-character-maker), Ruffle renders the colors incorrectly on the last frame of the menu. The colors are correct on all of the other frames. 

Steps to reproduce:
1. Run the SWF file in Ruffle
2. Click the right arrow button to start the game. 
3. Click the right arrow button repeatedly until you get to the last frame of the menu.

In Ruffle, it looks like this:
![image](https://user-images.githubusercontent.com/71368227/214485131-fe5f0dc3-9280-4116-a79f-790e7ff4173a.png)

### Expected behavior

The frame is supposed to look like this. The background colors are supposed to be the same as the previous frames. (Ignore the different-looking girl - her appearance is randomized)
![image](https://user-images.githubusercontent.com/71368227/214485290-3840d96c-90e3-4f51-afed-42f7be366df0.png)

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

This seems to be a problem with Ruffle's implementation of the Overlay blend mode. The menu sprite is sprite 1908, and its last frame is frame 12. On that frame, there are 3 PlaceObject3 tags with the Overlay blend mode:
![image](https://user-images.githubusercontent.com/71368227/214485875-4bba337e-4fc3-4446-8297-1eb3108c044a.png)
Changing their blend mode to Normal makes the problem go away. I think their blend mode is being applied to the entire screen instead of just the sprite they are placing.

Also, this seems like a similar issue to #4464. That game also uses PlaceObject3 tags with a blend mode (Erase).

Lord-McSweeney:
Seems to be fixed by #10251.

n0samu:
Uh, no it's not

n0samu:
This is a regression; I've bisected it to commit 8db8e4669adb6bc5498b19af66db78a10ce0cc50 of #9096.

n0samu:
Fixed by #11674

PR #11674, opened 2023-06-22
Fix blend mode Multiply with alpha blending

Dinnerbone:
This rolls it back to being a shader based blend, which is slower - but there's no real way around that.
Flash specialcased both src.a ==0 and dst.a == 0, which we can't account for without shaders :(

The difference to when it was a complex blend mode in the past is the addition of the `dst.a == 0` case.

This should fix https://github.com/ruffle-rs/ruffle/issues/11140 and https://github.com/ruffle-rs/ruffle/issues/11184 and also the remaining issues on my cacheAsBitmap PR afaik


render/wgpu/shaders/blend/multiply.wgsl
```diff
@@ -0,0 +1,52 @@
...
+}
+
+@fragment
+fn main_fragment(in: VertexOutput) -> @location(0) vec4<f32> {
+    // dst is the parent pixel we're blending onto
+    var dst: vec4<f32> = textureSample(parent_texture, texture_sampler, in.uv);
+    // src is the pixel that we want to apply
+    var src: vec4<f32> = textureSample(current_texture, texture_sampler, in.uv);
+
+    if (src.a > 0.0) {
```
Aaron1011, line 38: Could you add a comment here about the special-case in Flash?

Aaron1011 (approved):

n0samu:
Confirmed that this fixes #10455, #10884, and #9304!

render/wgpu/shaders/blend/multiply.wgsl
Dinnerbone: Added!

PR #11698, opened 2023-06-23
Add prependRotation, copyRowTo and copyRowFrom 

Fancy2209:
prependRotation is based on a suggestion by Aaron
copyRowTo and copyRowFrom implmentations are from OpenFL


core/src/avm2/globals/flash/geom/Matrix3D.as
```diff
@@ -106,6 +106,70 @@ package flash.geom {
 			}
 		}
 
+		//Based on https://github.com/openfl/openfl/blob/develop/src/openfl/geom/Matrix3D.hx#L542C1-L573
+		public function copyRowTo(row:uint, vector3D:Vector3D):void {
+			switch(row)
```
Aaron1011, line 111: These switches are missing `break` statements in each case.

core/src/avm2/globals/flash/geom/Matrix3D.as
```diff
@@ -106,6 +106,70 @@ package flash.geom {
...
+					vector3D.z = rawData[10];
+					vector3D.w = rawData[14];
+					break;
+				case 3:
+					vector3D.x = rawData[3];
+					vector3D.y = rawData[7];
+					vector3D.z = rawData[11];
+					vector3D.w = rawData[15];
+					break;
+				default:
```
Lord-McSweeney, line 137: Is this `default:` necessary?

core/src/avm2/globals/flash/geom/Matrix3D.as
```diff
@@ -106,6 +106,70 @@ package flash.geom {
...
+					rawData[10] = vector3D.z;
+					rawData[14] = vector3D.w;
+					break;
+				case 3:
+					rawData[3] = vector3D.x;
+					rawData[7] = vector3D.y;
+					rawData[11] = vector3D.z;
+					rawData[15] = vector3D.w;
+					break;
+				default:
```
Lord-McSweeney, line 169: same here

Fancy2209: I am not sure, should I remove it?

Fancy2209: Just tested, it isn't needed.

core/src/avm2/globals/flash/geom/Matrix3D.as
Fancy2209: Removed now

Aaron1011:
Could you add a test for these methods?

Fancy2209:
> Could you add a test for these methods?

I would love to but i don't have enough AS3 Knowledge to do so :/

Aaron1011:
I can write a test when I get access to my desktop again.

Fancy2209:
I see, thank you!

Aaron1011:
@Fancy2209 I added some tests, and fixed some bugs (the row index wasn't being checked, and you were writing to a copy of the array via the getter, not the underlying array).

PR #11702, opened 2023-06-23
Render filters on display objects

Dinnerbone:
This only hooks up existing filters, does not add new filters implementations.
AVM1 filters remain disconnected from `DisplayObject.filters`, this does not fix that.


render/wgpu/src/backend.rs
adrian17, line 1: commit nit: this change moves the code from `Sufrace::apply_filter` to `RenderBackend::apply_filter`, so the message doesn't really fit ;D

core/src/display_object.rs
```diff
@@ -768,6 +768,7 @@ pub fn render_base<'gc>(this: DisplayObject<'gc>, context: &mut RenderContext<'_
                 handle: handle.clone(),
                 commands: offscreen_context.commands,
                 clear: this.opaque_background().unwrap_or_default(),
+                filters: this.filters(),
```
adrian17, line 771: just to be sure: do you think us copying the filters vec (the call does a deep `.clone()` inside) is okay if dome each time here and in debug code?

adrian17 (approved):

core/src/display_object.rs
Dinnerbone: Debug code: I don't mind perf, I'd like to change `filters` to return a ref but that felt out of scope (already is vec today) which would be fine for debug

For render: It needs a clone ideally, I don't like the idea of refs leaking out to render, or things that can be interior mutable. I'd like to end up with a state that rendering can be done entirely on another thread on desktop (maybe web in a few years...) and trying to keep that design in mind, it submits work and moves on

render/wgpu/src/backend.rs
Dinnerbone: Good catch haha, I'll rename it

Aaron1011 (approved):

Issue #11513, opened 2023-06-13
Haxefilxel Demo


p0008874:
### Describe the bug

Filters:
https://github.com/ruffle-rs/ruffle/assets/75534590/8ddb2592-a5ff-433c-b4c9-424946cbdd08
FlxSpriteFilters:
![ruffle_yj9OGir2cJ](https://github.com/ruffle-rs/ruffle/assets/75534590/3b513d45-c642-401e-af6e-1d981959a3d8)
FileBrowse:
```
2023-06-13T11:47:09.566167Z  INFO ruffle_core::player: Loaded SWF version 21, resolution 640x480 @ 60 FPS
2023-06-13T11:47:09.663721Z  WARN run_frame:run_all_phases_avm2: ruffle_core::stub: Encountered stub: AVM2 flash.utils.Dictionary constructor with weak keys
2023-06-13T11:47:09.731684Z  WARN run_frame:run_all_phases_avm2: ruffle_desktop::backends::storage: Unable to read file Os { code: 3, kind: NotFound, message: "系統找不到指定的路徑。" }
2023-06-13T11:47:09.732215Z  WARN run_frame:run_all_phases_avm2: ruffle_desktop::backends::storage: Unable to read file Os { code: 3, kind: NotFound, message: "系統找不到指定的路徑。" }
2023-06-13T11:47:09.739447Z  WARN run_frame:run_all_phases_avm2: ruffle_core::stub: Encountered stub: AVM2 flash.media.Sound constructor with arguments
2023-06-13T11:47:13.421273Z  WARN run_frame:run_all_phases_avm2: ruffle_core::stub: Encountered stub: AVM2 flash.ui.MouseCursorData.hotSpot setter
2023-06-13T11:47:13.421591Z  WARN run_frame:run_all_phases_avm2: ruffle_core::stub: Encountered stub: AVM2 flash.ui.MouseCursorData.data setter
2023-06-13T11:47:13.422629Z  WARN run_frame:run_all_phases_avm2: ruffle_core::stub: Encountered stub: AVM2 flash.ui.Mouse.registerCursor()
```

### Expected behavior

Filters:
https://github.com/ruffle-rs/ruffle/assets/75534590/ecd42ba9-5f2e-4254-aa64-b5d6b70158d7
FlxSpriteFilters:
[...]

PR #11703, opened 2023-06-24
wgpu: Slight changes to blur filter

Lord-McSweeney:
See, for example, on https://github.com/ruffle-rs/ruffle/issues/11513.
Master:
image

With these changes:
image

This is visually almost indistinguishable compared to Flash (though Ruffle's blur strength seems to be a little higher, even with the blur amount halved in this PR).

Here's another comparison, of rd_flying with #11702.

#11702:
image

#11702 with these changes added:
image

As for the `bitmapdata_applyfilter_blur` test, the top-right one looks better, while the middle two look like they haven't been blurred enough (compared to FP), and the bottom two look improved (previously they were way more blurred than FP).


MrCheeze:
Some comparisons:

1. Flash player
2. Ruffle without this PR
3. Ruffle with this PR

image

image

image

Not perfect, but a major improvement. (Ignore the lack of yellow, that's GlowFilter, which isn't implemented at all.)

Dinnerbone (approved):

PR #11714, opened 2023-06-24
core: Apply filters when modifying an object, not just when instantiating

MrCheeze:
In https://github.com/ruffle-rs/ruffle/commit/fc00ae8eb6 , filters are applied in MovieClip.instantiate_child, which is called when creating an object. But the proper place for this is in TDisplayObject.apply_place_object, which is called both by instantiate_child _and_ when modifying an existing object with PlaceObject.

This change has no visible effect on master (since filters are still invisible). But when combined with https://github.com/ruffle-rs/ruffle/pull/11702 , which enables filter rendering, it fixes a bug where filters do not display unless they were set up at the same time as their object was created.

Note: I notice there are still [two TODOs](https://github.com/ruffle-rs/ruffle/blob/6d77ff1fc41aa34848188aa79ddb8c283671df4a/core/src/display_object/movie_clip.rs#L4343-L4400) in the codebase about applying/removing filters when running a GOTO. This PR doesn't touch those.

### DIFF
--- a/core/src/display_object.rs
+++ b/core/src/display_object.rs
@@ -1911,11 +1911,16 @@ pub trait TDisplayObject<'gc>:
                     self.set_opaque_background(context.gc_context, color);
                 }
             }
+            if let Some(filters) = &place_object.filters {
+                self.set_filters(
+                    context.gc_context,
+                    filters.iter().map(Filter::from).collect(),
+                );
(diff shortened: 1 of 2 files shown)


Dinnerbone (approved):
Thank you!

PR #11731, opened 2023-06-25
wgpu: Don't create cacheAsBitmap for textures that are too large

Dinnerbone:


Lord-McSweeney:
Can you add a test for this, using a very large shape?

Aaron1011 (approved):

Dinnerbone:
The limits vary based on the machine running the test, so not really - large for me might not be large for you.
I'd have to add an option per test to override the limits. I can do that if you think it's worth it

adrian17 (approved):

Lord-McSweeney (approved):

PR #11761, opened 2023-06-27
desktop,chore: Fix incorrect linker arguments on windows when using gnu linker(ld.exe)

lvyitian:
Fix incorrect linker arguments on windows when using gnu linker(ld.exe)
when not using msvc linker, the original build script will cause an error when building the desktop package


relrelb (approved):
Thank you and welcome!

torokati44:
(I think this PR should have been squashed into one commit, but oh well, no big deal.)

PR #11769, opened 2023-06-28
Switch blur to separable filter & respect quality setting

Dinnerbone:
It's not pixel perfect to flash but it should look a whole lot better - and be faster too!


TemaisgameNoobas:
does this PR boosts perfomance from the blur effect??
on my phone Ruffle mostly lags because of the blur effect

Dinnerbone:
The previous shader did X*Y texture lookups, this one does X+Y. It should be much faster, if the blur itself is the slow part.

adrian17 (approved):

Issue #11560, opened 2023-06-15
Mutilate-a-Doll 2: Failed to parse XML


MartySVK:
### Describe the bug

It is just a black screen with the error:

``ERROR ruffle_core::display_object::movie_clip: Got "RustError("Failed to parse XML: TypeError: Error #1088: The markup in the document following the root element must be well-formed.")" when constructing AVM2 side of movie clip of type Main``

SWF - http://f3.silvergames.com/mutilate-a-doll-2_v4.swf

### Expected behavior

progress the game

### Affected platform

Desktop app

### Operating system

Windows 10

Lord-McSweeney:
Fixed by #11782.

PR #11782, opened 2023-06-28
avm2: Correctly re-use objects in XML/XMLList call handlers and ctors

Aaron1011:
`XML(someXMLObj)` and `XMLList(someXMLList)` perform the normal cast behavior (returning the same object), instead of creating a new object like other arguments do.

This also applies to `new XMLList(someXMLObj)`
and `XML(singleElemXMLList)`


adrian17 (approved):

core/src/avm2/globals/xml.rs
```diff
@@ -10,6 +11,16 @@ use crate::avm2::string::AvmString;
 use crate::avm2::{Activation, Error, Multiname, Object, Value};
 use crate::avm2_stub_method;
 
+fn ill_formed_markup_err<'gc>(activation: &mut Activation<'_, 'gc>) -> Value<'gc> {
```
adrian17, line 14: btw, AFAIK the usual pattern we do here is: return the Result, then in the caller do `return Err(AvmError(ill_formed_markup_err(...)?));` instead of a panic.

n0samu:
Fixes #11560

PR #11783, opened 2023-06-28
wgpu: Remove 'TextureWrapper.format' field

Aaron1011:
We can already get the foramt from the wgpu::Texture

### DIFF
--- a/render/wgpu/src/context3d/mod.rs
+++ b/render/wgpu/src/context3d/mod.rs
@@ -405,7 +405,6 @@ pub struct ShaderCompileData {
 #[collect(require_static)]
 pub struct TextureWrapper {
     texture: wgpu::Texture,
-    format: wgpu::TextureFormat,
 }
 
 impl IndexBuffer for IndexBufferWrapper {}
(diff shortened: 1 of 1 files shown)


Dinnerbone (approved):

render/wgpu/src/context3d/mod.rs
```diff
@@ -405,7 +405,6 @@ pub struct ShaderCompileData {
 #[collect(require_static)]
 pub struct TextureWrapper {
```
relrelb, line 406: Doesn't this mean that `TextureWrapper` can be removed altogether? Since now it's totally equivalent to `wgpu::Texture`.

Aaron1011: This requires a little more work - the impl needs to go in the base `ruffle_render` crate behind a feature, since the orphan rule requires us to wrote `impl Texture for wgpu::Texture` in the crate where the `Texture` trait is defined.

I'm working a refactor of this as part of my pixelbender work, and I'll include that change in the future PR.

relrelb (approved):

PR #11798, opened 2023-06-29
Scale filter values when the filtered objects are scaled

Dinnerbone:
This will make blur_x/blur_y bigger (or smaller) when the object is scaled, which seems to match Flash (thanks @torokati44 for pointing that out!)

Also implemented @MrCheeze's idea of using accumulative values for the pass-scaling, so we can (later) calculate the size upfront correctly


swf/src/types/bevel_filter.rs
```diff
@@ -33,6 +33,11 @@ impl BevelFilter {
     pub fn num_passes(&self) -> u8 {
         (self.flags & BevelFilterFlags::PASSES).bits()
     }
+
+    pub fn scale(&mut self, x: f32, y: f32) {
```
relrelb, line 37: I think the more Rust-y way is to avoid mutation of structs, but instead consume and return a new struct. Something like:
```rs
pub fn scale(mut self, x: f32, y: f32) -> Self {
    self.blur_x *= Fixed16::from_f32(x);
    self.blur_y *= Fixed16::from_f32(y);
    self
}
```

Dinnerbone: I'm not sure. The enum method would look like this:

```rust
        match self {
            Filter::BevelFilter(filter) => filter.scale(x, y).into(),
            Filter::BlurFilter(filter) => filter.scale(x, y).into(),
            Filter::DropShadowFilter(filter) => filter.scale(x, y).into(),
            Filter::GlowFilter(filter) => filter.scale(x, y).into(),
            Filter::GradientBevelFilter(filter) => filter.scale(x, y).into(),
            Filter::GradientGlowFilter(filter) => filter.scale(x, y).into(),
            _ => self,
        }
```

It doesn't feel rusty to me

relrelb: Hmm, yeah. I guess avoiding those `.into()` is indeed preferred.

Aaron1011 (approved):

Issue #15, opened 2019-08-13
DisplayObject filter tracking issue


Herschel:
- [ ] Drop shadow
- [ ] Blur
- [ ] Glow
- [ ] Gradient glow
- [ ] Bevel
- [ ] Gradient bevel
- [ ] Convolution
- [ ] Color matrix

ivanpopelyshev:
I have all flash blur-based filters implementation in PixiJS WebGL if you need it ;)

Drop shadow / Blur / Glow / Bevel

Herschel:
Thank you, the shaders will definitely come in handy! I am interested in both pixel-perfect and approximations and allowing the user to choose (for example, in the canvas2d backend, I'd be happy with using standard SVG filter effects).

ivanpopelyshev:
I dont know parameters for svg blur :( 

I know which kernel and strength to use for different values of quality for approximation.

Pixel-perfect version is actually Fenwick tree :)

In Flash, blur filter is just "average premultiplied ARGB in a rectangle" , quality shows number of times applied.

wcbx:
Updates?

Herschel:
No updates yet. Blend modes will probably be implented first as they all rely on "cache as bitmap" behavior.

SelixGithub:
Is there a timeline for implementation? Are you planning to wait until everything else in AVM1 is complete?

n0samu:
Similarly to #5538, Pursuit of Hat needs ColorMatrixFilter - the cannons are always shown as red because their colors are supposed to be set via ColorMatrixFilter. This is particularly annoying on levels 15 and 17 because it's impossible to know in advance which button will fire which cannon.
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/9c5f0fcf-92b4-4c6f-9ea7-dc41bf76738d)

(Just adding another example, for whatever it's worth)

Aaron1011:
@ivanpopelyshev Ruffle is at the point where were implemting more filters. Could you share the source for the pixel-perfect flash filter implementations you mentioned? Thanks!

n0samu:
The problem with Pursuit of Hat is fixed by #11702 (and other DO/AVM2 ColorMatrixFilter issues will also be fixed)

ivanpopelyshev:
@Aaron1011  t.meSLASHhackerham  , write me in telegram, i'll share details. Cant post the source unfortunately )

Hexstream:
[...]

PR #11802, opened 2023-06-29
avm1: Slight cleanup to filters strength & angle

Dinnerbone:
We want precision in these objects, so stuffing a f64 into a u16 and converting it back isn't very useful.

### DIFF
--- a/core/src/avm1/globals/bevel_filter.rs
+++ b/core/src/avm1/globals/bevel_filter.rs
@@ -48,8 +48,7 @@ struct BevelFilterData {
     highlight: Color,
     shadow: Color,
     quality: i32,
-    // TODO: Introduce unsigned `Fixed8`?
-    strength: u16,
+    strength: f64,
     knockout: bool,
     blur_x: f64,
     blur_y: f64,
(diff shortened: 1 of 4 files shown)


adrian17 (approved):

core/src/avm1/globals/bevel_filter.rs
```diff
@@ -226,7 +225,7 @@ impl<'gc> BevelFilter<'gc> {
         value: Option<&Value<'gc>>,
     ) -> Result<(), Error<'gc>> {
         if let Some(value) = value {
-            let strength = ((value.coerce_to_f64(activation)? * 256.0) as u16).clamp(0, 0xFF00);
+            let strength = value.coerce_to_f64(activation)?.clamp(0.0, 255.0);
```
relrelb, line 228: Actually low precision is needed; Take for example:
```as
var f = new flash.filters.BevelFilter();
f.strength = 1 / 257;
trace(f.strength); // 0
```
Whereas now it would trace some non-zero number.

Lord-McSweeney: That should probably be added as a test.

TemaisgameNoobas:
wait, filters are now implemented??
i waited so long for them, since the summer of 2020!!

core/src/avm1/globals/bevel_filter.rs
Dinnerbone: Darn ok. Yes I agree that should be in a test!

danielhjacobs:
@TemaisgameNoobas To be clear, right now it's just Blur and Color Matrix: #15. But the infrastructure is now in place to add more of them.

PR #11815, opened 2023-06-30
avm2: Implement NetConnection.connect with a null `command`

Aaron1011:


core/src/avm2/globals/flash/net/net_connection.rs
```diff
@@ -0,0 +1,31 @@
...
+    avm2_stub_method,
+};
+
+pub fn connect<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Option<Object<'gc>>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let this = this.unwrap();
+    if let Value::Null = args[0] {
```
Lord-McSweeney, line 12: This should probably be `args.get_value(0)` with ParametersExt. Also, it is currently theoretically possible for `this` to be None, but Adrian is planning on fixing that soon.

Aaron1011: `args.get_value` is defined as `self[index]`, so it does the same thing.

Lord-McSweeney: The reason the ParametersExt functions are defined that way is so that it's easy to change/intercept every parameter lookup by changing just the `get_value` function. Otherwise there's no reason for `get_value` to be defined.

Aaron1011: I don't think that's true - these are just convenience methods.

Dinnerbone (approved):

n0samu:
Progresses Cardboard Box Assembler (#11921).

PR #11816, opened 2023-06-30
desktop: Use `SubscriberInitExt::init` to capture `log` output

Aaron1011:
This redirects output from the `log` crate to `tracing`

### DIFF
--- a/desktop/src/main.rs
+++ b/desktop/src/main.rs
@@ -24,6 +24,7 @@ use cli::Opt;
 use ruffle_core::StaticCallstack;
 use std::cell::RefCell;
 use std::panic::PanicInfo;
+use tracing_subscriber::util::SubscriberInitExt;
 use url::Url;
 
 thread_local! {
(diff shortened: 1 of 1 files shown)


desktop/src/main.rs
```diff
@@ -64,6 +64,13 @@ fn init() {
         panic_hook(info);
     }));
 
+    // Redirect Log to Tracing if it isn't already
+    let _ = tracing_log::LogTracer::builder()
+        // wgpu crates are extremely verbose
+        .ignore_crate("wgpu_hal")
```
Dinnerbone, line 70: We do this because there's no equivalent to `RUST_LOG` on web. It should be up to the user if they want, or don't want, to see wgpu things.

Dinnerbone:
It's really weird because `tracing-subscriber` should be doing this for us, and setting the right values based on filter etc. So I had a look.

It does it inside `SubscriberInitExt::try_init()` which we don't call because we do it ourselves, so let's just not do that.

Replace `    tracing::subscriber::set_global_default(subscriber).expect("Couldn't set up global subscriber");` with `subscriber.init();` and it'll be fixed, we don't need to do anything else or depend on tracing-log ourselves. It'll be correctly set up with the right filtering too.

Aaron1011:
@Dinnerbone I've updated the PR to use that method - I confirmed locally that I can see `wgpu` and `naga` output.

Dinnerbone (approved):

PR #11829, opened 2023-07-01
core: Expose `LoaderInfo.parameters` and fix other properties

Aaron1011:
When using a 'Loader', properties on the 'contentLoaderInfo' become set during specific events in the load sequence. In particular, 'LoaderInfo.bytesTotal' becomes available during the first 'progress' event.

Also, 'LoaderInfo.parameters' is now properly set from the URL query parameters. In Flash player, this work even with filesystem urls (e.g. 'file:///some/path/to/file.txt?paramOne=valOne' will load a file named 'file.txt', setting and expose the parameter 'paramOne' with value 'valOne' in `LoaderInfo.parameters`). This required some cleanup to the desktop and test NavigatorBackend impls to strip out query parameters when loading a parameter from disk.

Previously, we would set `SwfMovie.parameters` manually from the url. Now, the various `SwfMovie` constructors automatically extract query parameters from the provided url. Outside of `SwfMovie`, we only append *extra* parameters (e.g. those set from `flashvars`).

This makes CPMStar ads work, since the loaded SWF needs to access `LoaderInfo.parameters`


core/src/tag_utils.rs
```diff
@@ -101,27 +118,44 @@ impl SwfMovie {
...
+
+    fn append_parameters_from_url(&mut self) {
+        match Url::parse(&self.url) {
+            Ok(url) => {
+                for (key, value) in url.query_pairs() {
+                    self.parameters.push((key.into_owned(), value.into_owned()));
+                }
+            }
+            Err(e) => {
+                tracing::error!("Failed to parse LoaderInfo parameters from URL: {}", e);
```
Lord-McSweeney, line 157: This error message might make more sense as "Failed to parse self URL" or similar, since the error is to do with the url parsing, not the parameter parsing.

Dinnerbone (approved):

PR #11840, opened 2023-07-01
Sync package version updates in package-lock done after `npm install`

tsunamistate:
These are automatic changes done by NPM after doing `npm install`

Looks like some package versions in `devDependencies` block of `package-lock.json` has incorrect versions


torokati44:
This more or less reverts https://github.com/ruffle-rs/ruffle/pull/11819.

torokati44:
I don't mean to be rude or to overrule you with #11842, it just looked weird that the versions had to go backwards for the build to keep the repo clean.

tsunamistate:
OK, it was just weird that I was having `package-lock.json` that's supposed to be in sync kept falling out of sync after installing, so decided to commit the proposed change
Closing since #11842 supersedes this

PR #11871, opened 2023-07-03
avm2: Throw TypeError when trying to parse malformed XML

Aaron1011:
Scratch detects if a string is valid XML by trying to construct an XML object, and catching the thrown error.


tests/tests/swfs/avm2/xml_ctor_from_tostring/Main.as
```diff
@@ -69,6 +69,13 @@
 				trace("Caught error: " + e);
 				trace(e.errorID);
 			}
+		
+			try {
+				trace(new XML("<>"));
+			} catch (e) {
+				trace("Caught error: " + e);
+				trace(e.errorID);
+			}
```
Lord-McSweeney, line 78: Can you test for the error itself?

Aaron1011: I'm not sure what you mean

Lord-McSweeney: Can you add code to the test that parses invalid XML, then catches the error and checks if it the error is the same as FP (`Error #1090: XML parser failure: element is malformed.`)- the current test doesn't actually seem to parse invalid XML and check the error code/type.

Aaron1011: Fixed. It looks like `quick_xml` is very lenient in what it considers valid XML. This is a pre-existing issue, though, since we'll incorrectly create an `XML` instance in some cases both before and after this PR.

Lord-McSweeney (approved):

Dinnerbone (approved):

PR #11874, opened 2023-07-03
avm2: Set correct `this` values in function calls

adrian17:
Subsumes https://github.com/ruffle-rs/ruffle/pull/11673.

Changes:
- box primitives if they are used as `this`
- pass the correct `[object global]` if an unbound function is called
- pass the toplevel playerglobal `[object global]` if EventDispatcher calls a handler (note: there may be more cases like this)
- refactor function call apis to take `Value` instead of `Option<Object>`

This should also let us remove `Option` from `this: Option<Object>` in all native functions.

Downside of new design:
> this is now kinda awkward.
> a.f()
> we take a, turn it into an Object to do a vtable lookup of f, then turn it back into Value in .call(), then turn it back into an Object to pass the native function / Activation

Also not a fan of all these `let self_val: Value<'gc> = Value::from(self.into());`.

This would all eventually go away if we kill `PrimitiveObject`, but I that's a lot of work by itself.


Lord-McSweeney:
Progresses https://github.com/ruffle-rs/ruffle/issues/11659 (see #11673)

Dinnerbone (approved):

PR #11876, opened 2023-07-03
avm2: Support dependent strings, small string optimizations

adrian17:
This plugs in basic support for dependent strings, which are roughly similar to dependent strings in avmplus or Firefox.
This way, calling `LONG_STRING.substr(1)` is O(1) instead of O(n). And this way, scripts like
```
         while(data.length > 0)
         {
            char = data.charAt(0);
            data = data.substr(1);
            // ...code...
```
are no longer quadratic in space and time.

MVP scope:
- `substring()`, `substr()`, `slice()`
- `charAt()`, `split("")`
- cached, interned `""` and ASCII chars
- `a.substr().substr().substr()` points at the original parent string - no chain of depenednts

Out of MVP scope:
- other AVM2 APIs
- AVM1 APIs
- substring of a static string should be a static string (it's obvious on how to implement, but not obvious on how to nicely structure the code around it, it's already stretched across many layers)

Very out of scope:
- APIs that return `"".into()` and especially places that don't have direct access to `context.interner` at all.

### DIFF
--- a/core/src/string/interner.rs
+++ b/core/src/string/interner.rs
@@ -8,7 +8,7 @@ use std::ops::DerefMut;
 use gc_arena::{Collect, Gc, GcWeak, Mutation};
 use hashbrown::HashSet;
 
-use crate::string::{AvmString, AvmStringRepr, WStr};
+use crate::string::{AvmString, AvmStringRepr, WStr, WString};
 
 // An interned `AvmString`, with fast by-pointer equality and hashing.
 #[derive(Copy, Clone, Collect)]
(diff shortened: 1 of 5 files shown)


core/src/string/interner.rs
```diff
@@ -47,18 +47,40 @@ impl<'gc> AvmAtom<'gc> {
...
 
 impl<'gc> AvmStringInterner<'gc> {
-    pub fn new() -> Self {
-        Self::default()
+    pub fn new(mc: &Mutation<'gc>) -> Self {
+        let mut interned = WeakSet::default();
+
+        let mut intern_from_static = |s: &[u8]| {
+            let ch = Self::alloc(mc, Cow::Borrowed(WStr::from_units(s)));
+            ch.mark_interned();
```
moulins, line 65: `Self::alloc` already makes interned strings, so this call is useless.

core/src/string/interner.rs
```diff
@@ -91,14 +113,49 @@ impl<'gc> AvmStringInterner<'gc> {
...
     }
+
+    #[must_use]
+    pub fn empty(&self) -> AvmString<'gc> {
+        self.empty.into()
+    }
+
+    #[must_use]
+    pub fn get_char(&self, mc: &Mutation<'gc>, c: u16) -> AvmString<'gc> {
+        if c < 128 {
```
moulins, line 132: Using `if c < self.chars.len()` would avoid repeating the magic number 128.

core/src/string/interner.rs
```diff
@@ -91,14 +113,49 @@ impl<'gc> AvmStringInterner<'gc> {
...
+        s: AvmString<'gc>,
+        start_index: usize,
+        end_index: usize,
+    ) -> AvmString<'gc> {
+        if start_index >= end_index {
+            return self.empty.into();
+        }
+        if end_index == start_index + 1 {
+            if let Some(c) = s.get(start_index) {
+                if c < 128 {
```
moulins, line 152: Same as l.132.

core/src/string/interner.rs
```diff
@@ -91,14 +113,49 @@ impl<'gc> AvmStringInterner<'gc> {
...
+
+    #[must_use]
+    pub fn substring(
+        &self,
+        mc: &Mutation<'gc>,
+        s: AvmString<'gc>,
+        start_index: usize,
+        end_index: usize,
+    ) -> AvmString<'gc> {
+        if start_index >= end_index {
```
moulins, line 147: IMHO, this should panic if `start_index > end_index`, for consistency with slicing syntax (and maybe it should even take a `range: impl RangeBounds<usize>`?)

moulins:
LGTM!

torokati44:
Much like for #14631, I also made a somewhat "cleaned up" commit history for this.
https://github.com/torokati44/ruffle/commits/avm2-dependent-string-temp/
Feel free to take it over if you'd like, but I won't push it that much for this PR.

All this is just to satisfy my apparent newfound obsession with this kind of cleanup.
I really hope I'm not too annoying with it. :no_mouth:

Aaron1011 (approved):

PR #11813, opened 2023-06-30
Hook up AVM1 filters

Dinnerbone:
- [x] Blur
- [x] Bevel
- [ ] Glow
- [ ] DropShadow
- [x] ColorMatrix
- [ ] DisplacementMap
- [ ] Convolution
- [ ] GradientBevel
- [ ] GradientGlow

Issue #9011, opened 2023-01-03
Outlines created using GlowFilter don't work.


MrCheeze:
### Describe the bug

Load this flash: https://mrcheeze.github.io/homestuck-ruffle-tester/#02791_local

Hold up to walk up to the nearby computer, and then press space to interact with it.

A prompt with two options will appear. This prompt uses GlowFilter to have an orange and yellow box outlining it, but the filter seems to do nothing in Ruffle:

```actionscript
this.options = new MovieClip();
this.options.graphics.clear();
this.options.filters = [new GlowFilter(0xFFFF00,1,4,4,255),new GlowFilter(0xFF9900,1,6,9,255)];
this.options.addChild(/* stuff... */);
```

### Expected behavior

Flash player appearance:
![image](https://user-images.githubusercontent.com/6541413/210421870-119189e0-4095-4b19-ba7d-ad1832a15e4a.png)

Ruffle appearance:
![image](https://user-images.githubusercontent.com/6541413/210421914-dc472771-9a80-4acd-91f5-4b22aa70ccaa.png)

Ruffle should have the orange and yellow box outline.

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

Mozilla Firefox

### Additional information

I guess this is waiting on https://github.com/ruffle-rs/ruffle/issues/15 and https://github.com/ruffle-rs/ruffle/issues/299 .

MrCheeze:
Fixed by #11878 !

PR #11878, opened 2023-07-03
wgpu: Implement glow filter

Dinnerbone:
(texture sizes still don't grow with filters, that's a separate issue that applies to all filters)


n0samu:
Testing [Learn to Fly 2](https://www.kongregate.com/games/light_bringer777/learn-to-fly-2), this definitely improves some things, but not everything I expected.
- The main menu buttons use GlowFilter, but Ruffle doesn't seem to notice:
image
image
- The text glow in the Light Bringer Games logo animation looks wrong (I guess it's just the lack of filter rect expansion again)

https://github.com/ruffle-rs/ruffle/assets/71368227/f4fd6600-9e04-4d61-9c77-8f36b2011956

Some other filters in buttons seem to be missing as well.

Dinnerbone:
Can you see if the filter exists on the button using the debug menu? To narrow down if this is because of the filter not being applied (from timeline or avm) , or just not rendering correctly

n0samu:
Also, this strangely makes Ninja Painter look worse at its default resolution, while it looks fine with the window maximized (look at the "Main Menu" text and the Agame logo):
image
image

You can see the same problem on the level select screen too:
image

n0samu:
> Can you see if the filter exists on the button using the debug menu? To narrow down if this is because of the filter not being applied (from timeline or avm) , or just not rendering correctly

It looks like Ruffle doesn't notice the filter at all.
image

n0samu:
I'm actually running this with both this PR and #11813 pulled in

n0samu:
This also reminds me of the current issues with button blends like https://github.com/ruffle-rs/ruffle/issues/9201#issuecomment-1517500384

MrCheeze:
1. Filters do not render along an axis when that axis's scale is negative (NOT a problem introduced by this PR, just one I noticed while testing it).
image

2. The GlowFilters described in https://github.com/ruffle-rs/ruffle/issues/9011 still do not render (but according to the debug menu, they exist)

image
image

n0samu:
The text with glow filter also looks bad in Super Mario 63 at the default resolution (and fine maximized):
image

(Just another example of what I showed with Ninja Painter already)

Dinnerbone:
Filters now grow the texture size, so blurs and glows won't be cut off

Dinnerbone:
@n0samu
image
image

Dinnerbone:
All known issues are fixed afaik!

render/wgpu/src/filters/glow.rs
```diff
@@ -15,9 +15,9 @@ use wgpu::util::DeviceExt;
 struct GlowUniform {
     color: [f32; 4],
     strength: f32,
-    inner: u32,    // a wasteful bool, but we need to be aligned anyway
-    knockout: u32, // a wasteful bool, but we need to be aligned anyway
-    _pad: f32,
+    inner: u32,            // a wasteful bool, but we need to be aligned anyway
+    knockout: u32,         // a wasteful bool, but we need to be aligned anyway
+    composite_source: u32, // undocumented flash feature, another bool
```
Lord-McSweeney, line 20: Could this use a single `u32` with the three bools stored in the lower bits?

n0samu:
Thank you so much, Dinnerbone! 😃 

I noticed a tiny imperfection remaining - the blur on completed achievements in Ninja Painter's achievements screen is cut off at the edges for some reason:
image

Here's how it looks in Flash Player for comparison:
image

render/wgpu/src/filters/glow.rs
Dinnerbone: It could, but as the comment says we need to stay aligned. 16 bytes for webgl, which means the fields won't go anywhere and we'd just be cramming out data for no reason.

n0samu:
Just checked and this fixes #11950!

n0samu:
[Catchy Orbit](https://www.newgrounds.com/portal/view/630301/format/flash?emulate=flash) looks great now 😃
image

n0samu:
Another minor issue: the outlines around the letters of the [HitBox](http://www.notdoppler.com/hitbox.php) logo are considerably thinner than they should be:
image

PR #11880, opened 2023-07-04
Add wireframe mode toggle to debug menu

torokati44:
This adds a checkbox to toggle wireframe mode to the debug menu.
Doing so by switching to an entirely different set of pipelines, with `PolygonMode::Line`, as suggested by @adrian17 and @Dinnerbone.
Those are created lazily, so hopefully they won't add any additional startup delay.

I'm not at all sure how this is how it should be done.
My main concern is inflating the `RenderBackend` trait too much with less relevant stuff.

image


torokati44:
Now it's done with renderer backend downcasting (desktop with egui can only work with wgpu at the moment anyway), and not through a needlessly inflated RendererBackend trait.

Should the "not supported" tooltip be localized?

PR #11891, opened 2023-07-04
core: Correctly apply blend modes and filters from button records

Lord-McSweeney:
Makes some of the Papa's games look better (see https://github.com/ruffle-rs/ruffle/issues/9201#issuecomment-1517500384).

This should also fix the glow filter not applying to the buttons in https://github.com/ruffle-rs/ruffle/pull/11878#issuecomment-1619919537.


adrian17 (approved):

Dinnerbone:
A test would be nice but I'm not going to block on that. Thank you!

Dinnerbone (approved):

PR #11894, opened 2023-07-04
avm2: Implement XML.descendants for attributes

Aaron1011:
We were previously only ever checking children,
and not attributes.

In order to avoid matching both attributes and elements with a given name in 'descendants', `E4xNode::matches_name` now checks `is_attribute` on the provided `Multiname`. This requries changing several other parts of the codebase to properly set this flag on `Multinames` provided by ActionScript.


core/src/avm2/e4x.rs
```diff
@@ -524,7 +529,19 @@ impl<'gc> E4XNode<'gc> {
...
+        {
+            if name.is_attribute() {
+                for attribute in attributes {
+                    if attribute.matches_name(name) {
+                        out.push(E4XOrXml::E4X(*attribute));
+                    }
+                }
+            }
             for child in children {
                 if child.matches_name(name) {
```
adrian17, line 546: if we did `if !name.is_attribute()` here, then `matches_name` itself doesn't need to do the is-attribute check anymore, I think

adrian17 (approved):

PR #11895, opened 2023-07-04
avm2: Implement Graphics.copyFrom

Aaron1011:


Dinnerbone (approved):
Not strictly a problem with this PR but it'll be the first thing to trigger it; getting the drawing will trigger an invalidation on the source object because so far every use was mutable. We should rename the existing one to _mut and add a readonly one that doesn't invalidate

Lord-McSweeney (approved):

PR #11896, opened 2023-07-04
render: Implement more Pixelbender opcodes, and fix matrices

Aaron1011:
Matrices don't actually get their own registers - they are stored in normal float registers, with each column of the matrix getting its own register. The exception is float2x2, which is stored in a single vec4 (this is visible to PixelBender bytecode, which writes to a single vector, and then uses it as an M2x2 channel).


tests/tests/swfs/avm2/pixelbender_effect_twirl/twirl.pbk
```diff
@@ -0,0 +1,137 @@
...
+ * All Rights Reserved.
+ *
+ * NOTICE:  All information contained  herein is,  and remains the property of
+ * Adobe Systems Incorporated and its suppliers, if any.  The intellectual and
+ * technical  concepts  contained  herein  are  proprietary  to  Adobe Systems
+ * Incorporated  and  its suppliers  and may  be covered  by U.S.  and Foreign
+ * Patents, patents in process, and are protected by trade secret or copyright
+ * law.  Dissemination of this information or reproduction of this material is
+ * strictly forbidden  unless prior  written permission is obtained from Adobe
+ * Systems Incorporated.
```
Dinnerbone, line 13: 😬

Lord-McSweeney: This seems to be from https://github.com/nojvek/pixelbender/blob/master/twirl/twirl.pbk, which has no such license on it (and it's not like we're distributing it)- where did the license come from?

Dinnerbone:
Squash this on merge please; I don't want the old license in the history and have awkward questions if someone does a code scan in the future

Dinnerbone (approved):

Issue #11806, opened 2023-06-30
"Unsupported filter" spam in the console/terminal


n0samu:
Not a bug, but Ruffle outputs an "unsupported filter" warning every single time it tries to apply any unsupported filter to a display object, which can be many times per second. This is making debugging and triaging issues a lot more difficult than it used to be. Can we please output one warning per unsupported filter instead?

n0samu:
Fixed by #11915

PR #11915, opened 2023-07-05
wgpu: Only log 'unsupported' warning once per filter type

Aaron1011:
This eliminates the console spam produced when an object with an unsupported filter was onscreen.


n0samu:
Closes #11806.

Dinnerbone (approved):

PR #11916, opened 2023-07-05
Implement Bitmap pixel snapping

Dinnerbone:
This should make any content that renders bitmaps at not-entirely-pixel-values look much better, possibly helping to alleviate the text readability issues too.

Anything that's cacheAsBitmap has it forced on, which you can see through the updated tests.


core/src/avm2/globals/flash/display/bitmap.rs
```diff
@@ -80,15 +82,17 @@ pub fn init<'gc>(
         let bitmap_data = args
             .try_get_object(activation, 0)
             .and_then(|o| o.as_bitmap_data());
-        //TODO: Pixel snapping is not supported
-        let _pixel_snapping = args.get_string(activation, 1);
+        let Some(pixel_snapping) = PixelSnapping::from_wstr(&args.get_string(activation, 1)?) else {
+            return Err(Error::AvmError(argument_error(activation, "Error #2008: Parameter pixelSnapping must be one of the accepted values.", 2008)?));
```
Lord-McSweeney, line 86: Use `avm2::error::make_error_2008(activation, "pixelSnapping")` here

core/src/avm2/globals/flash/display/bitmap.rs
```diff
@@ -146,21 +150,35 @@ pub fn set_bitmap_data<'gc>(
...
+    this: Option<Object<'gc>>,
+    args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
-    avm2_stub_setter!(activation, "flash.display.Bitmap", "pixelSnapping");
+    if let Some(bitmap) = this
+        .and_then(|this| this.as_display_object())
+        .and_then(|dobj| dobj.as_bitmap())
+    {
+        let Some(value) = PixelSnapping::from_wstr(&args.get_string(activation, 0)?) else {
+            return Err(Error::AvmError(argument_error(activation, "Error #2008: Parameter pixelSnapping must be one of the accepted values.", 2008)?));
```
Lord-McSweeney, line 178: Same goes for here

Croworbit:
is this accurate what adobe flash does?

Dinnerbone:
> is this accurate what adobe flash does?

More or less. I suspect that they snap the width and height too but it's hard to tell 😅

n0samu:
Does this actually improve anything? So far I'm not seeing any

n0samu:
This fixes the display of the teagames.com logo in [Let it Glow 2](http://www.notdoppler.com/letitglow2.php), which uses pixelSnapping:
**Before:**
image
**After:**
image

I'm not seeing a difference in any content that doesn't explicitly use pixelSnapping though.

adrian17 (approved):
Can we copy the same logic to other backends? Maybe the webgl side could be the same 4-5 line change, I hope :D

PR #11936, opened 2023-07-07
avm1, avm2: Support onMetaData handler for AVM1, support onCuePoint and onMetaData NetStream.client function handlers for AVM2

### LINKED ISSUE #11921: Cardboard Box Assembler: Play button doesn't appear
### Describe the bug

Download: [CardboardBoxAssembler_LocalFLV.zip](https://github.com/ruffle-rs/ruffle/files/11963611/CardboardBoxAssembler_LocalFLV.zip)

**Note**: This is also affected by #11911, so you will need to delete the game's save data each time you test it, otherwise Ruffle will panic. And be sure to test this with #11815 merged in. Also, in Flash Player this will only work via a local web server because the SWF's security sandbox doesn't allow accessing local files. 

In Cardboard Box Assembler, the Play button doesn't appear after the intro video plays. The relevant code is in the `com.adultswim.Preroll` class:
```as
    public function onCuePoint(param1:Object) : void
    {
        trace("cuepoint: time=" + param1.time + " name=" + param1.name + " type=" + param1.type);
        switch(param1.name)
        {
            case "enterPlay":
                GlobalVarContainer.vars.bReady = true;
                this.nav = new Navigation();
        }
    }
```
onCuePoint does not trigger in Ruffle. We need an implementation like #11858 for AVM2.

### Expected behavior

The Play button should appear after the intro video finishes playing.

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

_No response_

Lord-McSweeney:
Fixes #11921


core/src/avm2/globals/flash/net/net_stream.rs
```diff
@@ -80,3 +82,37 @@ pub fn toggle_pause<'gc>(
...
+
+    Ok(Value::Undefined)
+}
+
+pub fn set_client<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Object<'gc>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let client = args.get_value(0).as_object();
```
Dinnerbone, line 103: Not `get_object` for proper error handling?

Lord-McSweeney: I've tested this and `null`, `undefined`, `3`, and `true` all throw the same Error # 2004. (meanwhile, `MovieClip` and `{}` work fine). Note that this is a setter function, not exactly a "do operation" function.

Lord-McSweeney: I've added the test demonstrating this.

core/src/streams.rs
```diff
@@ -697,22 +709,25 @@ impl<'gc> NetStream<'gc> {
         context: &mut UpdateContext<'_, 'gc>,
         variable_name: &[u8],
         variable_data: FlvValue,
-    ) {
+    ) -> Result<(), Avm2Error<'gc>> {
         match avm_object {
             Some(AvmObject::Avm1(object)) => {
                 match variable_name {
-                    b"onCuePoint" => {
+                    b"onCuePoint" | b"onMetaData" => {
```
kmeisthax, line 716: My assumption with FLV `ScriptData` was that it was a generic mechanism for calling methods on `NetStream`/`NetStream.client`. Any particular reason why there's a method name check here?

core/src/streams.rs
```diff
@@ -738,12 +750,32 @@ impl<'gc> NetStream<'gc> {
                 };
             }
             Some(AvmObject::Avm2(_object)) => {
-                tracing::warn!(
-                    "Stub: FLV stream data processing (name: {}) for AVM2",
-                    WStr::from_units(variable_name)
-                );
+                match variable_name {
+                    b"onCuePoint" | b"onMetaData" => {
```
kmeisthax, line 754: Same comments from the AVM1 version of this code also applies here.

kmeisthax (approved):
I'm actually really glad to see someone jump in and take on the script data stuff before I could get to it. Thanks a ton.

I have one question, though...

kmeisthax (approved):

PR #11943, opened 2023-07-07
avm2: Remove old hack now that DOs use an instance allocator

Lord-McSweeney:


adrian17 (approved):
LGTM, just squash it please ;D

Issue #11640, opened 2023-06-20
Giants and Dwarves TD - Can't place towers


desertking:
### Describe the bug

Dunno why but in the game I'm not able to place the tower. Doesn't recognize the click on the spot it seems.
https://www.newgrounds.com/portal/view/621555?emulate=flash

### Expected behavior

Able to place towers

### Affected platform

Online demo

### Operating system

All

### Browser

All

n0samu:
The game also breaks trying to load an existing save slot:
```
ERROR run_frame:run_all_phases_avm2: ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x22fb2a04988 }) to handler FunctionObject(FunctionObject { ptr: 0x22fb2a850f8 }) : TypeError: Error #1010: A term is undefined and has no properties. (accessing field: length)
        at gGame$/LoadGameData()
        at mcSaveSlotSelectCls$/InitSlot()
        at mcSaveSlotSelectCls/UpdateSlots()
        at mcSaveSlotSelectCls()
        at IGame/OpenSaveSlotSelect()
        at mcMainMenuCls/StartCampaign()
        at PageTurn/OnEnterFrame()
```

n0samu:
The saving/loading issue is due to unimplemented VectorObject serialization, like #9688
`WARN ruffle_core::avm2::amf: Serialization is not implemented for class other than Object: VectorObject(VectorObject { ptr: 0x1f69ffd5dc0 })`

n0samu:
And the tower-placing issue happens because `getObjectsUnderPoint` is unimplemented. This will be fixed by #11970.
`WARN run_frame:run_all_phases_avm2: ruffle_core::stub: Encountered stub: AVM2 flash.display.DisplayObjectContainer.getObjectsUnderPoint()`

serkonda7:
This issue can be closed.

Issue #10751, opened 2023-04-20
10 is again - Cannot click anything in menu


desertking:
### Describe the bug

https://www.newgrounds.com/portal/view/631141?emulate=flash - As stated above I cannot interact with the menu in any way

### Expected behavior

Click and do something

### Affected platform

Online demo

### Operating system

All

### Browser

All

n0samu:
This happens because `getObjectsUnderPoint()` is unimplemented:
`WARN ruffle_core::stub: Encountered stub: AVM2 flash.display.DisplayObjectContainer.getObjectsUnderPoint()`

PR #11970, opened 2023-07-09
avm2: Implement DisplayObjectContainer.getObjectsUnderPoint

Aaron1011:
I still need to write tests for this


n0samu:
Unfortunately this does not fix #10751.

n0samu:
This fixes #11640!

adrian17 (approved):
Can we get this in, maybe with a `// TODO` that this might be a naive not-perfect implementation?

(though, even a simplest test would be nice to have here)

techboyg5:
Hello?

Issue #9693, opened 2023-02-23
Crystal Crunch Story: Black Screen Issue


waspennator:
### Describe the bug

[Crystal Crunch Story.zip](https://github.com/ruffle-rs/ruffle/files/10815111/Crystal.Crunch.Story.zip)

Game seems to have started working recently as of now, but as soon as you start the game and get past the first cutscene, you are greeted with a black screen as the music plays.

![Crystal Crunch Ruffle](https://user-images.githubusercontent.com/78768499/220949764-5ab6783a-93b2-4f41-bb44-37a4eb2ffbbe.png)

### Expected behavior

It should display this screen after the first cutscene.

![Crystal Crunch Flash](https://user-images.githubusercontent.com/78768499/220949797-236cf796-c300-4527-891a-9616deb347cc.png)

### Affected platform

Desktop app

### Operating system

Windows 11

waspennator:
Still broken as of the latest nightly and nothing helpful is showing up in the logs save for this, unless this is what's breaking it.

Encountered stub: AVM2 flash.media.Sound constructor with arguments

waspennator:
Checking as of the latest nightly, my only guess is that it uses a combination of glow and dropshadow filters to render the playing field, looking at the logs while running the game.

waspennator:
![Crystal](https://github.com/ruffle-rs/ruffle/assets/78768499/609d4b59-0949-4199-8790-51da6561a7cb)

#11878 seems to make this swf almost perfectly playable, the only issue that i'm noticing is that the blocks dropping aren't making noises, though other noises still seem to play properly.

n0samu:
@waspennator can you open a new issue for the sound problem with this game?

Issue #9171, opened 2023-01-15
Pilgrimage: Broken animations, no music, softlock


waspennator:
### Describe the bug

https://www.newgrounds.com/portal/view/498177?emulate=flash
Game doesn't seem to lock up when pressing play anymore on ng, but it still has issues.

1: The animations and sound don't seem to play correctly on ruffle, the main character just slides around in place, the coins essentially become stuck solid white when you pick them up, and enemies get squished but don't seem to vanish correctly. Also the ingame music doesn't play.

2: Moving past this point causes the entire screen to freeze up, only being able to zoom in moving right or zoom out if you move left. It does fix itself if you move left enough away from the trigger spot, but the platforming further ahead makes this unplayable.

![212549230-57e7e34b-a40c-40a9-beda-c9f4b2291020](https://user-images.githubusercontent.com/78768499/212550874-a21b1adf-4ba3-4075-b9f8-f1b2caa8a259.png)

### Expected behavior

The animations and sound should play correctly and the screen shouldn't get stuck at that point and should correctly turn into pixelated garbage instead.

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 108

waspennator:
[www.newgrounds.com-1676519694262.log](https://github.com/ruffle-rs/ruffle/files/10751999/www.newgrounds.com-1676519694262.log)
Tested again and still happening, providing a log file with some of the errors I see. Apparently this SWF needs the "GetDescendants" opcode.

Lord-McSweeney:
Stubbing `Shader` does not fix the problem.

waspennator:
> Stubbing `Shader` does not fix the problem.

Tested it on latest nightly and noticed 2 new things.

1: The game seems to toss this up in the log, which i'm curious if the game is throwing a fit about this.
"Error occured when running AVM2 frame script: ReferenceError: Error # 1065: Variable Shader is not defined."

2: If you try to go past the spot where the screen is supposed to distort and glitch out, this error gets spammed a bunch in the logs until you go back.
"Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x3e148d8 }) to handler FunctionObject(FunctionObject { ptr: 0x2a91100 }) : TypeError: Error # 1009: Cannot access a property or method of a null object reference. (accessing field: data)"

waspennator:
Ran it through the ruffle website and my only guess is that one of these stubs are causing the issue.

[...]

PR #11972, opened 2023-07-09
avm2: Call Sound.load in constructor when passed a URLRequest

Aaron1011:


Lord-McSweeney:
Can you add a test for this?

waspennator:
Should effect #9171, #9693, and #12055?

n0samu:
Unfortunately this does not seem to fix any of those issues.

danielhjacobs (changes requested):
Not directly related, but change https://github.com/ruffle-rs/ruffle/blob/c829d797ce567666a5779a33121cf56c461d8fd4/core/src/avm2/globals/flash/media/sound.rs#L246 to `let _sound_context = args.try_get_object(activation, 1);`. It's currently never `None`, so the stub message always appears when `load` is called.

danielhjacobs:
Together with #12967, should fix https://dnasubway.cyverse.org/files/tour/tour-all.swf

danielhjacobs:
Test for this PR added in #13005.

danielhjacobs (approved):
If this is merged as-is, in a follow-up PR I'll rebase #13005 on it and remove `known_failure`, and in another follow-up I'll change this to `let _sound_context = args.try_get_object(activation, 1);` https://github.com/ruffle-rs/ruffle/blob/c829d797ce567666a5779a33121cf56c461d8fd4/core/src/avm2/globals/flash/media/sound.rs#L246

adrian17 (approved):

Issue #9785, opened 2023-02-28
Troyis Chess: Couldn't parse SWF


n0samu:
### Describe the bug

Link to SWF: https://www8.agame.com/mirror/flash/t/troyis_chess.swf

Loading the SWF file for Troyis Chess fails with this error:
`ERROR run_frame: ruffle_core::avm1::runtime: Couldn't parse SWF: Error parsing AVM1 action ConstantPool: Not enough data for string`
`ERROR run_frame: ruffle_core::avm1::runtime: No more actions will be executed in this movie.`

The game displays but is not interactive.

Note that the game also fails to load in the latest Adobe Flash Player. But it works in Flash Player 7, 8 and 9 from here: https://archive.org/details/MacromediaFlashProjectors.7z

### Expected behavior

The game should load as it does in Flash Player 7, 8 and 9.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

JPEXS Flash Decompiler is unable to disassemble the code in this movie.

I found this from the logs in #9047

Dinnerbone:
Even FFDEC can't read the scripts 🤔

n0samu:
The log from #9047 shows a few other instances of this error, but all of those are immediately following an unknown opcode with a huge amount of bogus data, so I'm not sure if those are a different problem. They similarly fail to load in latest Flash Player, though. An example is Tiny Battle: https://www8.agame.com/mirror/flash/t/tiny_battle.swf

Lord-McSweeney:
For future reference, the reason JPEXS cannot read the scripts was because the ConstantPool entries were invalid strings, which we now ignore.

PR #11980, opened 2023-07-10
avm1: Cache ConstantPools for constant pools with more than 1000 entries

Lord-McSweeney:
This makes Tiny Battle (see https://github.com/ruffle-rs/ruffle/issues/9785#issuecomment-1448073050) play at a reasonable speed after #11957. The game was using a constant pool with >18000 entries, which got re-interned in Ruffle every time the DoAction tag was iterated over.

### DIFF
--- a/core/src/avm1/runtime.rs
+++ b/core/src/avm1/runtime.rs
@@ -10,10 +10,11 @@ use crate::context::{GcContext, UpdateContext};
 use crate::frame_lifecycle::FramePhase;
 use crate::prelude::*;
 use crate::string::AvmString;
-use crate::tag_utils::SwfSlice;
+use crate::tag_utils::{SwfPosition, SwfSlice};
 use crate::{avm1, avm_debug};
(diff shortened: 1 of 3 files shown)


core/src/avm1/runtime.rs
```diff
@@ -27,6 +28,8 @@ pub struct Avm1<'gc> {
     /// don't close over the constant pool they were defined with.
     constant_pool: Gc<'gc, Vec<Value<'gc>>>,
 
+    constant_pool_cache: HashMap<SwfPosition, Gc<'gc, Vec<Value<'gc>>>>,
```
moulins, line 31: If SWFs are repeatedly loaded then unloaded, this will cause a memory leak; this should probably be a `weak_table::WeakKeyHashMap` instead.

moulins:
Why limit the caching to >1000 entries constant pools? I don't think always caching has any extra performance cost.

Issue #11967, opened 2023-07-09
Epic Battle Fantasy 3: Invisible Text


waspennator:
### Describe the bug

[548041_Epic_Battle_Fantasy_3_secu.zip](https://github.com/ruffle-rs/ruffle/files/11995808/548041_Epic_Battle_Fantasy_3_secu.zip)
(Maximize Flash Player and Ruffle in order to see it)
When you press play, normally there should be a warning on the side telling you not to resize the game as it can lead to visual issues, but for some reason, this isn't showing up on ruffle, which is weird cause ruffle debug does seem to detect it as existing and visible.

![Screenshot 2023-07-09 120255](https://github.com/ruffle-rs/ruffle/assets/78768499/01e01930-420f-48b2-aa45-e78c377aaa17)

### Expected behavior

![Screenshot 2023-07-09 120120](https://github.com/ruffle-rs/ruffle/assets/78768499/0659d274-5fe5-4e73-b1a2-62e232e8d7c9)
There should be a warning about resizing the game on the far left of the screen when maximized once you press play.

### Affected platform

Desktop app

### Operating system

Windows 11

n0samu:
This is not a bug - it happens because the text is placed outside the stage area, and by default Ruffle letterboxes the off-stage areas. If you run the desktop app with ` --letterbox=off` you'll see the text. This made me notice that the letterbox setting was somehow omitted from the "Open Advanced" menu though, I guess I'll fix that!

waspennator:
Closing since pr to turn off letterboxing is merged now and only happens when it's on.

PR #11981, opened 2023-07-10
desktop: Add Letterbox option to Open Advanced dialog

n0samu:
I guess we missed this one before. Addresses #11967.

I also made a minor change to make the wording for removing movie parameters more consistent.


Dinnerbone (approved):
Keep those translators on the their toes! 😉

PR #12002, opened 2023-07-11
avm1: Migrate XMLNode to NativeObject

### LINKED ISSUE #11999: Missing texts in HabboExperience animation
### Describe the bug

Hey there, I'm currently trying to port the "Habbo Experience" promotional animation from 2005. The original is still available at [http://images.habbo.com/downloads/experience/exp-20050704-1/index_uk.html](http://images.habbo.com/downloads/experience/exp-20050704-1/index_uk.html).

While porting, I've noticed that certain texts seem to be missing. There is no output in the console that would suggest a problem is present though.

Ruffle Version:
image
image
image

### Expected behavior

The texts are displayed.

Flash Version:
image
image
image

### Affected platform

Self-hosted version

### Operating system

Windows 11 latest

### Browser

Microsoft Edge (Chromium based) latest

### Additional information

The swf loads certain additional files on startup.
- scene3.swf
- scene4.swf
- scene5.swf
- scene6.swf
- data
  - conf.xml
  - [language-code]
    - console.swf
    - localization.xml
    - structure.xml

I've confirmed that the xml files are valid and the project is still working using native flash in Pale Moon. The missing texts would be in localization.xml and console.swf, however the texts in structure.xml are used properly.

This problem exists both in the self-hosted version and the browser plugin.

Dinnerbone:
This fixes #11999 

image


core/src/avm1/object.rs
```diff
@@ -60,6 +58,7 @@ pub enum NativeObject<'gc> {
     TextFormat(GcCell<'gc, TextFormat>),
     NetStream(NetStream<'gc>),
     BitmapData(BitmapDataWrapper<'gc>),
+    XmlNodeObject(XmlNode<'gc>),
```
Lord-McSweeney, line 61: The naming convention here is that this should be `XmlNode` instead of `XmlNodeObject`.

Dinnerbone: Yep sorry copy paste error, fixed!

Lord-McSweeney (approved):

PR #12018, opened 2023-07-12
wgpu: Implement Bevel filter

Dinnerbone:
Pixel perfect :)


n0samu:
Ninja Painter is totally messed up now 😱 
image

render/wgpu/shaders/filter/blur.wgsl
```diff
@@ -24,13 +23,30 @@ fn main_vertex(in: filter::VertexInput) -> filter::VertexOutput {
...
+        // If the width is even, we have a right edge of a fixed weight and offset
+        center_length -= 1.5;
+        total += textureSample(texture, texture_sampler, left_uv + (direction * (filter_args.full_size - 0.75))) * 1.5;
+    }
+
+    // At this point, the center_length must be a whole number, divisible by 2.
+    center_length /= 2.0;
+    for (var i = 0.0; i < center_length; i += 1.0) {
+        // The center of the kernel is always going to be 1,1 weight pairs. We can just sample between the two pixels.
+        total += textureSample(texture, texture_sampler, left_uv + (direction * (1.5 + (i * 2.0)))) * 2.0;
```
torokati44, line 47: Does it make sense to do silly stuff like:
 - Accumulate the center sum separately, and move the `* 2.0` out to after the loop.
 - Step `i` by 2 instead of multiplying it by `2.0` in the loop (would also spare the `center_length /= 2.0` above).

... or will the shader compilers deal with this sort of stuff on their own, so no need to hurt readability in source?
... or it doesn't matter either way as the bottleneck is memory bandwidth?

Dinnerbone: They're worth trying, and I had a few extra ideas:
- Offset the UV we give it to the left on the vertices, so we don't need to calculate it per fragment
- Precalculate `left_weight` and provide it from rust
- Switch to push constants where supported

Aaron1011 (approved):

PR #12028, opened 2023-07-13
web: Fix broken touch detection

n0samu:
This reverts ea78bf2d3983b525b61861eae7e8fbb715ace272 and restores @danielhjacobs's original, working approach from #10950.

### DIFF
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -139,6 +139,9 @@ export class RufflePlayer extends HTMLElement {
     // Allows the user to permanently disable the context menu.
     private contextMenuForceDisabled = false;
 
+    // Whether this device is a touch device.
+    // Set to true when a touch event is encountered.
+    private isTouch = false;
     // Whether this device sends contextmenu events.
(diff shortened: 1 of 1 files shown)


danielhjacobs:
For context for anyone who reviews this: https://github.com/ruffle-rs/ruffle/blob/009fab1889fe737c6dfafc5bca0ebaef927b8cd4/web/packages/core/src/ruffle-player.ts#L1356-L1358 This is the specific issue. The `contextmenu` event is a `MouseEvent` on most browsers so this is broken on most of them. It's a `PointerEvent` on Chrome though, so this does work on Chrome.

danielhjacobs (approved):
Note that if https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event#browser_compatibility ever reaches a point where it says that `contextmenu` is a `PointerEvent` on every browser that Ruffle supports, we can undo this revert if desired. However, browser manufactures are slow to implement changes made by the W3C, so this is unlikely to happen anytime soon.

relrelb (approved):
Confirming it was my oversight, sorry. Current changes look good to me.

Issue #12022, opened 2023-07-13
Spiral Out not working


Lord-McSweeney:
### Describe the bug

https://www.pouet.net/prod.php?which=55593 (see #11807).

The SWF needs `flash.geom.Matrix3D.transformVectors`; it currently gets stuck on a black screen.

### Expected behavior

SWF works correctly.

### Affected platform

Desktop app

### Operating system

Linux

Lord-McSweeney:
This is working a bit better now, but has performance and rendering issues. Reported stubs:
```
Encountered stub: AVM2 flash.geom.Transform.matrix3D setter
Encountered stub: AVM2 flash.display.DisplayObject.z setter
Encountered stub: AVM2 flash.utils.Dictionary constructor with weak keys
Encountered stub: AVM2 flash.display.NativeMenuItem constructor
Encountered stub: AVM2 flash.geom.PerspectiveProjection constructor
Encountered stub: AVM2 flash.geom.PerspectiveProjection.fieldOfView setter
Encountered stub: AVM2 flash.geom.PerspectiveProjection.toMatrix3D()
Encountered stub: AVM2 flash.geom.PerspectiveProjection.focalLength getter
Encountered stub: AVM2 flash.geom.PerspectiveProjection.focalLength setter
Encountered stub: AVM2 flash.display.Graphics.drawTriangles() winding behavior
Encountered stub: AVM2 flash.display.Graphics.drawTriangles() with uvt data
```

PR #12038, opened 2023-07-13
avm2: Implement `Matrix3D.transformVectors`

agdeneke:
This implements the Matrix3D.transformVectors() method required by Spiral Out in #12022.


core/src/avm2/globals/flash/geom/Matrix3D.as
```diff
@@ -190,6 +190,14 @@ package flash.geom {
 			return new Vector3D(x, y, z, w);
 		}
 
+		public function transformVectors(vin:Vector.<Number>, vout:Vector.<Number>):void {
+			var result3D:Vector3D;
```
Lord-McSweeney, line 194: Can you please add this code at the beginning:
```
if (vin == null) {
    throw new TypeError("Error #2007: Parameter vin must be non-null.", 2007);
}
if (vout == null) {
    throw new TypeError("Error #2007: Parameter vout must be non-null.", 2007);
}
```
Other than that, I think this PR is fine. Thanks for your contribution!

tests/tests/swfs/avm2/matrix3d/Test.as
```diff
@@ -56,6 +56,12 @@
 			vOut = mat.transformVector(v);
 			trace(vOut.x, vOut.y, vOut.z, vOut.w);
 
+			var vecs:Vector.<Number> = Vector.<Number>([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
```
Aaron1011, line 59: Could you add calls with null arguments, within catch blocks that print the exception? Thanks!

Lord-McSweeney (approved):

PR #12046, opened 2023-07-14
avm1: Migrate `Xml` to `NativeObject`

relrelb:


core/src/avm1/object.rs
```diff
@@ -578,11 +577,6 @@ pub trait TObject<'gc>: 'gc + Collect + Into<Object<'gc>> + Clone + Copy {
         None
     }
 
-    /// Get the underlying XML document for this object, if it exists.
-    fn as_xml(&self) -> Option<XmlObject<'gc>> {
-        None
-    }
-
     /// Get the underlying XML node for this object, if it exists.
     fn as_xml_node(&self) -> Option<XmlNode<'gc>> {
```
Dinnerbone, line 581: I suspect you'll need to override this to make xml_node operations work on xml objects, unless you've handled that another way?

relrelb: Hmm, I think you're right. Will update `as_xml_node` to handle both `XmlNode` and `Xml` objects.

Issue #8937, opened 2022-12-27
Gray screen on Transformice


Lord-McSweeney:
### Describe the bug

https://www.transformice.com/Transformice.swf fails to load, instead displaying a gray screen. The obfuscated game uses the unimplemented flash.profiler.showRedrawRegions is not yet implemented.

### Expected behavior

The game can load and run.

### Affected platform

Browser's extension

### Operating system

Linux

### Browser

Chrome 108

### Additional information

This game is also multiplayer, so until XMLSocket is implemented, the core functionality will not run.

Lord-McSweeney:
This issue will probably be fixed even if `flash.profiler` and `flash.profiler.showRedrawRegions` are stubbed.

Lord-McSweeney:
This also needs `flash.text.StyleSheet`.

Lord-McSweeney:
Here's a reverse-engineered version of the topmost-level DoABC:
```
package
{
    import flash.display.Sprite;
    import flash.utils.getDefinitionByName;
    import flash.utils.ByteArray;
    import flash.system.LoaderContext;
    import flash.display.Loader;
   
    public dynamic class CQNGLbjwtl extends Sprite
    {
      
      private static const MAINARRAY = "As4S29Fo_HjiWpZKyYv5mTqXlt7hr8C0uLRnBz.fd36IUVbMwexcDP1GNkJEaQgO"["split"]("");
       
      
      private var ajcxizualc:Class;
      
      private var iqkugpxnbvsc:Class;
      
      private var lfxpqk:Class;
      
      private var awgwprizmoyf:Class;
      
      private var drpqhvk:Class;
      
      private var qdagovprmbj:Class;
      
      private var crisfl:Class;
      
      private var dkpaobzeh:Class;
      
      private var zlfywsv:Class;
      
      private var rixfnsaj:Class;
      
      private var arwsrcv:Class;
      
      private var jmdtgnjhr:Class;
      
      public function CQNGLbjwtl()
      {
         this.ajcxizualc = CQNGLbjwtl_ajcxizualc;
         this.iqkugpxnbvsc = CQNGLbjwtl_iqkugpxnbvsc;
         this.lfxpqk = CQNGLbjwtl_lfxpqk;
         this.awgwprizmoyf = CQNGLbjwtl_awgwprizmoyf;
         this.drpqhvk = CQNGLbjwtl_drpqhvk;
[...]

Issue #2002, opened 2020-12-17
The Skullkid - Can't chainsaw through first desk


TomFulp:
https://www.newgrounds.com/portal/view/63747

I am pretty sure this was working in Ruffle before so something may have regressed. Walk up to the first table and hold the spacebar to chainsaw through it - part way through the cutting animation, something stops your progress and you can't complete the cutting.

Toad06:
You're right about the regression, there was no issue in nightly builds <= 2020-11-29.
However, this appears to have been fixed in the nightly 2020-12-10 or newer.

Herschel:
Looks like this is working on master.

PR #12047, opened 2023-07-14
core, avm2: Implement AS3 Socket

sleepycatcoding:
Based on #8188

Things left to do:

- [x] Use async connect in desktop
- [ ] Use correct exceptions/errors in Socket methods
- [x] Add command line flag to prevent, ask or always allow Socket connections
- [x] Implement all read*() and write*() methods
- [ ] Web implementation


core/src/avm2/globals/flash/net/socket.rs
```diff
@@ -0,0 +1,306 @@
...
+        let endian = args
+            .get(0)
+            .unwrap_or(&Value::Undefined)
+            .coerce_to_string(activation)?;
+        if &endian == b"bigEndian" {
+            socket.set_endian(Endian::Big);
+        } else if &endian == b"littleEndian" {
+            socket.set_endian(Endian::Little);
+        } else {
+            return Err(make_error_2008(activation, "value"));
```
Lord-McSweeney, line 89: nit: this should be `make_error_2008(activation, "endian")`

core/src/avm2/globals/flash/net/socket.rs
```diff
@@ -0,0 +1,306 @@
...
+    activation: &mut Activation<'_, 'gc>,
+    this: Object<'gc>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let socket = match this.as_socket() {
+        Some(socket) => socket,
+        None => return Ok(Value::Undefined),
+    };
+
+    let host = match args.get(0) {
```
Lord-McSweeney, line 17: This should be `let host = args.get_value(0);`, since the first parameter is guaranteed.

core/src/avm2/globals/flash/net/socket.rs
```diff
@@ -0,0 +1,405 @@
...
+pub fn close<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Object<'gc>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    if let Some(socket) = this.as_socket() {
+        // We throw an IOError when socket is not open.
+        let handle = socket
+            .get_handle()
+            .ok_or(Error::AvmError(io_error(activation, "TODO", 0)?))?;
```
Lord-McSweeney, line 56: `io_error(activation, "Error #2002: Operation attempted on invalid socket.", 2002)`

core/src/avm2/globals/flash/net/socket.rs
```diff
@@ -0,0 +1,405 @@
...
+pub fn flush<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Object<'gc>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    if let Some(socket) = this.as_socket() {
+        let UpdateContext { sockets, .. } = &mut activation.context;
+
+        // FIXME: Throw correct IoError
+        let handle = socket.get_handle().unwrap();
```
Lord-McSweeney, line 146: `io_error(activation, "Error #2002: Operation attempted on invalid socket.", 2002)`
Seeing as this error is going to appear often, it might be a good idea to create a function like:
```
fn invalid_socket_error<'gc>(
) -> Error<'gc> {
    io_error(activation, "Error #2002: Operation attempted on invalid socket.", 2002)
}
```

core/src/avm2/globals/flash/net/socket.rs
```diff
@@ -0,0 +1,405 @@
...
+    Ok(Value::Undefined)
+}
+
+pub fn write_bytes<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Object<'gc>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    if let Some(socket) = this.as_socket() {
+        let bytearray = args
```
Lord-McSweeney, line 292: `let bytearray = args.get_object(activation, 0)?;`

core/src/avm2/globals/flash/net/socket.rs
```diff
@@ -0,0 +1,405 @@
...
+            .get(0)
+            .unwrap_or(&Value::Undefined)
+            .coerce_to_object(activation)?;
+        let offset = args
```
Lord-McSweeney, line 296: `let offset = args.get_u32(activation, 1)?;`

core/src/avm2/globals/flash/net/socket.rs
```diff
@@ -0,0 +1,405 @@
...
+            .get(1)
+            .unwrap_or(&Value::Integer(0))
+            .coerce_to_u32(activation)? as usize;
+        let length = args
```
Lord-McSweeney, line 300: `let offset = args.get_u32(activation, 2)?;`

core/src/avm2/globals/flash/net/socket.rs
```diff
@@ -0,0 +1,405 @@
...
+    Ok(Value::Undefined)
+}
+
+pub fn write_byte<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Object<'gc>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    if let Some(socket) = this.as_socket() {
+        let byte = args
```
Lord-McSweeney, line 275: `let byte = args.get_i32(activation, 0)?;`

Lord-McSweeney:
The rest of the parameter lookups should also use `ParametersExt`.

core/src/avm2/globals/flash/net/socket.rs
```diff
@@ -272,11 +260,7 @@ pub fn write_byte<'gc>(
     args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
     if let Some(socket) = this.as_socket() {
-        let byte = args
-            .get(0)
-            .cloned()
-            .unwrap_or(Value::Undefined)
-            .coerce_to_i32(activation)?;
+        let byte = args.get_i32(activation, 0)?;
```
Lord-McSweeney, line 263: nit: Use `get_u32` here.

core/src/avm2/globals/flash/net/socket.rs
```diff
@@ -14,18 +15,8 @@ pub fn connect<'gc>(
...
-    }
-    .coerce_to_string(activation)?;
-
-    // FIXME: How do you get u16? there is no coerce_to_u16 method.
-    let port = args
-        .get(1)
-        .unwrap_or(&Value::Undefined)
-        .coerce_to_u32(activation)? as u16;
+    let host = args.get_string(activation, 0)?;
+    let port = args.get_i32(activation, 1)? as u16;
```
Lord-McSweeney, line 19: nit: Use `get_u32` here.

Lord-McSweeney:
Apart from my few nits, I think the AVM2-interface part of this looks good. Thank you for contributing!

core/src/avm2/globals/flash/net/socket.rs
```diff
@@ -0,0 +1,365 @@
...
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    if let Some(socket) = this.as_socket() {
+        let bytearray = args.get_object(activation, 0, "bytes")?;
+        let offset = args.get_u32(activation, 1)? as usize;
+        let length = args.get_u32(activation, 2)? as usize;
+
+        let ba_read = bytearray
+            .as_bytearray()
+            .ok_or("ArgumentError: Parameter must be a bytearray")?;
```
Lord-McSweeney, line 282: This is guaranteed to be a ByteArray by the AS-side bindings and the `get_object` filtering out any `nill`s, so you can just do `.expect("Parameter must be a bytearray!")`.

core/src/avm2/globals/flash/net/socket.rs
```diff
@@ -0,0 +1,365 @@
...
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let socket = match this.as_socket() {
+        Some(socket) => socket,
+        None => return Ok(Value::Undefined),
+    };
+
+    let host = args.get_string(activation, 0)?;
+    let port = args.get_i32(activation, 1)? as u16;
+
```
Lord-McSweeney, line 20: This should throw a `SecurityError`: `Error #2003: Invalid socket port number specified.`

core/src/socket.rs
```diff
@@ -0,0 +1,239 @@
...
+                        .0
+                        .get(handle)
+                        .expect("only valid handles in SocketAction")
+                        .target;
+
+                    let socket = target.as_socket().expect("only SocketObjects in handles");
+                    socket.read_buffer().extend(data);
+
+                    let socket_data_evt =
+                        EventObject::bare_default_event(&mut activation.context, "socketData");
```
Lord-McSweeney, line 233: This should be a `ProgressEvent`, not an `Event`. According to the Adobe docs:

Event Object Type: [flash.events.ProgressEvent](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/events/ProgressEvent.html)
property ProgressEvent.type = [flash.events.ProgressEvent.SOCKET_DATA](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/events/ProgressEvent.html#SOCKET_DATA)

core/src/socket.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// flush message not sent yet for instance.
+    fn poll(&mut self) -> Option<Vec<u8>>;
+}
+
+pub type SocketHandle = Index;
+
+#[derive(Collect)]
+#[collect(no_drop)]
+struct Socket<'gc> {
+    target: Object<'gc>,
```
Lord-McSweeney, line 63: Wouldn't it be better to have this as an `SocketObject`, since there's no way it could be any other kind of object?

sleepycatcoding: Good idea. I did not think of that.

core/src/avm2/globals/flash/net/socket.rs
sleepycatcoding: I used the TryFrom<u32> impl for u16 here. It errors when u32 is bigger than u16::MAX. Then I mapped the error to the correct SecurityError.

core/src/socket.rs
```diff
@@ -0,0 +1,217 @@
...
+                            .classes()
+                            .ioerrorevent
+                            .construct(
+                                &mut activation,
+                                &[
+                                    "ioError".into(),
+                                    false.into(),
+                                    false.into(),
+                                    "TODO".into(),
+                                    0.into(),
```
sleepycatcoding, line 151: Not sure what message or id to use here.

Lord-McSweeney: For now this can use `Error #2031: Socket Error.`

tests/tests/util/navigator.rs
```diff
@@ -125,4 +127,17 @@ impl NavigatorBackend for TestNavigatorBackend {
...
+
+    fn connect_socket(
+        &mut self,
+        _host: String,
+        _port: u16,
+        handle: SocketHandle,
+        _receiver: Receiver<Vec<u8>>,
+        sender: Sender<SocketAction>,
+    ) {
+        sender
```
Lord-McSweeney, line 139: Since this is for the test backend, I think it might be preferable to log a socket connection here, similarly to how it  was done for `navigate_to_url` and `fetch`, so that testing might be a bit easier. For example:
```
if let Some(log) = &self.log {
    log.avm_trace("Navigator::connect_socket");
    log.avm_trace(&format!("    Host: {}; Port: {}", host, port));
}
```

desktop/src/backends/navigator.rs
```diff
@@ -46,6 +60,8 @@ impl ExternalNavigatorBackend {
         proxy: Option<Url>,
         upgrade_to_https: bool,
         open_url_mode: OpenURLMode,
+        socket_allowed: HashSet<String>,
+        socket_behavior: SocketBehavior,
     ) -> Self {
```
sleepycatcoding, line 65: This new function triggers clippy::too_many_arguments lint. Should we allow it here?

core/src/avm2/object/socket_object.rs
```diff
@@ -24,6 +24,8 @@ pub fn socket_allocator<'gc>(
             // Default endianness is Big.
             endian: Cell::new(Endian::Big),
             object_encoding: Cell::new(ObjectEncoding::Amf3),
+            // Default is 20_000 milliseconds or 20 seconds.
+            timeout: Cell::new(20000),
```
Lord-McSweeney, line 28: For technical correctness with Flash, the timeout is set when instantiating the Socket, not when creating the object.  So the "default" should be `0`, and the constructor should do this:
```
        public function Socket(host:String = null, port:int = 0) {
            this.timeout = 20000; <-- timeout gets set here
            if (host != null) {
                this.connect(host, port);
            }
        }
```

desktop/src/backends/navigator.rs
```diff
@@ -277,4 +294,145 @@ impl NavigatorBackend for ExternalNavigatorBackend {
...
+            let mut stream = match stream {
+                Ok(stream) => {
+                    sender
+                        .send(SocketAction::Connect(handle, true))
+                        .expect("working channel send");
+
+                    stream
+                }
+                Err(err) => {
+                    warn!("Failed to connect to {}:{}, error: {}", host2, port, err);
```
Lord-McSweeney, line 366: AFAIK a timed out connection will throw a SecurityErrorEvent in Flash. It's probably not necessary to implement that right now, but it might be good to add a TODO notice here for that.

desktop/src/backends/navigator.rs
```diff
@@ -277,4 +294,145 @@ impl NavigatorBackend for ExternalNavigatorBackend {
...
+                        Err(_) => break,
+                    }
+                }
+
+                let mut buffer = [0; 4096];
+
+                match stream
+                    .read(&mut buffer)
+                    .or(async {
+                        Timer::after(Duration::from_millis(50)).await;
```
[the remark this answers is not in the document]
sleepycatcoding: The timeout here is for reading, not connecting. The problem is some protocols might not send us data before we send them data. As such if that timeout was not there, we would get into a infinite loop waiting for data from server.

core/src/avm2/object/socket_object.rs
sleepycatcoding: Done.

desktop/src/backends/navigator.rs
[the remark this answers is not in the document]
sleepycatcoding: I reread your comment and I also don't really know if the timeout should be increased.

sleepycatcoding:
> Is there a specific reason `timeout` is an `Option<Duration>` on the backend side? It seems unnecessary.

There is no specific reason. I removed Option from Duration.

desktop/src/backends/navigator.rs
sleepycatcoding: Is there a specific error message, id?

sleepycatcoding: I added a way to differentiate between a failed and timed out connection.

Lord-McSweeney:
Transformice (#8937) fails to connect with this PR. It works in FP, so the server must still be up. It might be worth looking into that.

sleepycatcoding:
It is very interesting as there seem to be two versions of the SWF. I could connect and play with this version https://www.transformice.com/TransformiceChargeur.swf, but the link in the linked issue https://www.transformice.com/Transformice.swf fails with message `Invalid version, try to reload the game.`, this happens both on Ruffle and Flash Player Projector, which leads me to believe they have updated the game since. The link they have on their website for the Old Browser based version (https://www.transformice.com/transformice-swf.html) uses the former SWF, not the latter.

sleepycatcoding:
Rebased.

Lord-McSweeney:
@sleepycatcoding you are correct. https://www.transformice.com/TransformiceChargeur.swf is the loader for the game, and the game has some magic values that are frequently changed in https://www.transformice.com/Transformice.swf; a different file is served every time. Therefore, the second one is not expected to work.

sleepycatcoding:
@Lord-McSweeney I think the problem you might have had with this is that by default socket connections are denied. You can enable them with `--socket unrestricted` or `--socket ask` flags or use the advanced open dialog which has socket behavior under network settings.

Lord-McSweeney:
I think the `--socket` option should match the `--open-url-mode` option (`--socket-mode allow/confirm/deny`), though I'd like to get others to weigh in on this. Also, I think setting the default to `deny` might be confusing to users- IMO, a default of `ask` would be better.

Dinnerbone (approved):
This is really cool. The code looks good, but, I'm a little hesitant about the lack of testing. However, I recognise there's no easy way to test this currently.

I've been pondering on that for a bit, and I think if we:
1. Make a new file `socket.toml` or something, which represents a list of what a mocked up socket should do. Receive X, send X, wait X, disconnect, Any deviation is an error.
2. Write a small utility that hosts a socket locally and runs that script, printing any deviation or progress it makes. Use this to get Flash output.
3. Add an option to the test config to mock a connection using this file, so it doesn't actually open sockets and introduce unpredictability to the tests. The mock socket would run through the script and fail the test if it didn't complete to the end, or received anything out of the ordinary.

I'm happy to not block this PR on that because this is really cool and something is better than nothing, but I think I'd want that for future improvements to this area. Does that sound like something you'd be able to do, or will you want help?

Thank you again for this PR!

torokati44:
> I think the --socket option should match the --open-url-mode option (--socket-mode allow/confirm/deny), though I'd like to get others to weigh in on this.

Agree. Naming it simply `socket` implies that the option itself would name a socket of some sort, which is not the case.

> Also, I think setting the default to deny might be confusing to users (because the connection silently fails)- IMO, a default of ask would be better.

Also agree. Keeping it silently on deny by default diminishes its out-of-the-box usability and discoverability of it's presence. I'd even go as far as to suggest setting it to allow by default on desktop.

Dinnerbone:
> Also agree. Keeping it silently on deny by default diminishes its out-of-the-box usability and discoverability of it's presence. I'd even go as far as to suggest setting it to allow by default on desktop.

I think one day when we know that we're secure and following whatever security practices existed at the time, but not today. Ask is a good compromise.

sleepycatcoding:
> Agree. Naming it simply `socket` implies that the option itself would name a socket of some sort, which is not the case.

What naming should we use for this? Maybe something along the lines like `socket-behavior` or `socket-mode`?

sleepycatcoding:
> This is really cool. The code looks good, but, I'm a little hesitant about the lack of testing. However, I recognise there's no easy way to test this currently.
> 
> I've been pondering on that for a bit, and I think if we:
> 
>     1. Make a new file `socket.toml` or something, which represents a list of what a mocked up socket should do. Receive X, send X, wait X, disconnect, Any deviation is an error.
> 
>     2. Write a small utility that hosts a socket locally and runs that script, printing any deviation or progress it makes. Use this to get Flash output.
> 
>     3. Add an option to the test config to mock a connection using this file, so it doesn't actually open sockets and introduce unpredictability to the tests. The mock socket would run through the script and fail the test if it didn't complete to the end, or received anything out of the ordinary.
> 
> 
> I'm happy to not block this PR on that because this is really cool and something is better than nothing, but I think I'd want that for future improvements to this area. Does that sound like something you'd be able to do, or will you want help?

I would prefer to have the testing implementation in a separate PR.

Hadaward:
I was testing this PR just now with Transformice and found that for some reason my ping was much higher in Ruffle than in Flash, something is slowing down the socket.


In flash the ping is 305, in ruffle it is 396, I used the /ping command of the game itself to verify. I also noticed a delay in synchronizing player movement. I tested it at least 10 times, and as much as my ping varied, it was never the same or close in the ruffle, always keeping a good distance between the loaders.

sleepycatcoding:
@Hadaward Can you retry with the latest commit? On my machine the change improved the latency difference to about 10 ms.

Lord-McSweeney:
I believe this still needs error checking in all the `readX`/`writeX` methods to check if the socket is actually connected.

sleepycatcoding:
> I believe this still needs error checking in all the `readX`/`writeX` methods to check if the socket is actually connected.

Added that behavior, also added a test to check that the IOErrors are correctly thrown.

Hadaward:
I just tested with the most recent commit, unfortunately I think my internet is not good. The ping in the ruffle and in the flash player are varying a lot, sometimes it is better in the ruffle than in the flash player, and sometimes it is better in the flash player than in the ruffle. But everything related to networking seems to be working. But regarding the previous delay, I almost don't feel it anymore, ruffle and flash player are behaving almost the same. If there's a delay now, it's because my ping isn't stable.

Lord-McSweeney (approved):
Looks good to me! Tests can be added in a follow-up PR.

PR #12056, opened 2023-07-14
wgpu: Fix off-by-one in blur size calculation

Dinnerbone:


Lord-McSweeney:
Can you add a test?

n0samu:
More important to get this in quickly and stop the crashes!

PR #12072, opened 2023-07-15
web: Move from many npm scripts, to a single build file

Dinnerbone:
Much more maintainable, and adding new options should be super easy now (such as webgpu).

It's also much faster, for me at least! There seems to be quite some seconds between every `npm run` on my machine.


web/packages/core/package.json
```diff
@@ -10,26 +10,7 @@
...
-        "//7": "# Dispatches to either building the real, or copying the fake (stand-in),",
-        "//8": "# 'with-extensions' module.",
-        "build:wasm-extensions": "node -e \"process.exit(process.env.ENABLE_WASM_EXTENSIONS == 'true' ? 0 : 1)\" && npm run build:wasm-extensions-real || npm run build:wasm-extensions-fake",
-        "build:wasm-extensions-real": "echo \"Building module with WebAssembly extensions\" && cross-env OUT_NAME=ruffle_web-wasm_extensions CARGO_PROFILE=web-wasm-extensions RUSTFLAGS=\"$RUSTFLAGS --cfg=web_sys_unstable_apis -Aunknown_lints -C target-feature=+bulk-memory,+simd128,+nontrapping-fptoint,+sign-ext,+reference-types\" WASM_BINDGEN_FLAGS=\"--reference-types\" WASM_OPT_FLAGS=\"--enable-reference-types\" npm run build:wasm",
-        "build:wasm-extensions-fake": "echo \"Copying the vanilla module as stand-in\" && shx cp dist/ruffle_web_bg.wasm dist/ruffle_web-wasm_extensions_bg.wasm && shx cp dist/ruffle_web_bg.wasm.d.ts dist/ruffle_web-wasm_extensions_bg.wasm.d.ts && shx cp dist/ruffle_web.js dist/ruffle_web-wasm_extensions.js && shx cp dist/ruffle_web.d.ts dist/ruffle_web-wasm_extensions.d.ts",
```
relrelb, line 26: `shx` is no longer used, thus can be removed from `devDependencies`.

relrelb (approved):
Looks good (apart from 1 minor comment)

web/packages/core/package.json
Dinnerbone: Good catch! Removed

web/packages/core/tools/build_wasm.js
```diff
@@ -0,0 +1,127 @@
...
+    });
+    console.log(`Running wasm-bindgen on ${flavor}...`);
+    runWasmBindgen({
+        path: `../../../target/wasm32-unknown-unknown/${profile}/ruffle_web.wasm`,
+        outName: filename,
+        dir: "dist",
+        flags: wasmBindgenFlags,
+    });
+    if (optimise) {
+        console.log(`Running wasn-opt on ${flavor}...`);
```
torokati44, line 88: Sorry, late, but typo: wasn-opt

Issue #10481, opened 2023-03-31
Homestuck font render issues (device font related)


MrCheeze:
### Describe the bug

[This movie](https://mrcheeze.github.io/homestuck-ruffle-tester/#01358) ([direct swf link](https://mrcheeze.github.io/homestuck-ruffle-tester/flash/01358.swf)) has multiple font rendering issues. Other flashes from Homestuck tend to use fonts in the same way and have similar issues.

1) Underlines are missing:
![image](https://user-images.githubusercontent.com/6541413/229163930-4100f0ea-2415-45ab-b027-8e8e8b5ec2e4.png)

2) Text is cut off because it uses the Ruffle default font instead of Courier New device font (see https://github.com/ruffle-rs/ruffle/issues/1862)
![image](https://user-images.githubusercontent.com/6541413/229164373-a10547c4-f449-4ce3-b095-d94f2cbe849d.png)

### Expected behavior

See the comparison above - underlines should be present in the text, and Courier New should be used instead of the Ruffle default font.

Underline presence is self-explanatory.

For getting Courier New, I would consider any solution that *doesn't* require editing the movie itself to be valid, for example:
- Ruffle actually loading the operating system's device fonts, as Flash Player does.
- Ruffle adding builtin support for [common fonts.](https://www.w3schools.com/cssref/css_websafe_fonts.php)
- Requiring the host website to provide an additional fonts file, as [proposed by n0samu here.](https://github.com/ruffle-rs/ruffle/issues/8387)

### Affected platform

Self-hosted version

### Operating system

Windows 11

### Browser

Mozilla Firefox

MrCheeze:
"Requiring the host website to provide an additional fonts file"  has now been implemented in https://github.com/ruffle-rs/ruffle/pull/13604, allowing for the correct font to be used!

PR #12088, opened 2023-07-15
core: Make underlines the same color as their text

MrCheeze:
This fixes issue 1 from https://github.com/ruffle-rs/ruffle/issues/10481 (underlines appear to be missing, because they are black on a black background, when they should be in the text color).


Dinnerbone:
Can you add a visual test for it please? We need more coverage for text stuff 😅

MrCheeze:
> Can you add a visual test for it please? We need more coverage for text stuff 😅

What's the protocol if there are significant other text differences that I don't want to fix?

expected
actual-windows-Vulkan
difference-color-windows-Vulkan

Dinnerbone:
Totally fine - the image is a snapshot to make sure we don't unintentionally regress. Whenever that stuff does get fixed up, the test will be reviewed by whomever and they'll see that either it's better, or it's worse and they come out with a bonus test case for their changes :D

Dinnerbone (approved):

PR #12089, opened 2023-07-15
wgpu: Blur optimisations

Dinnerbone:
It should be approximately 50% faster, as it's doing roughly half as much samples.

Here's the before-and-after of samples per kernel:
image

Left side is a naive but correct implementation of sampling one pixel per width of the kernel. This is our previous implementation, with the exception that we didn't support even-width kernels - we rounded them to odd width. Every sample lands in the center of the pixel and has a weight of 1 or 0.5 depending on if it's an edge piece which only wants "half" of that pixel to count. The numbers above each sample is the weight of that sample.

Right side is this new implementation, which tries to use the hardware to sample and mix two pixels at a time for us.

The center of every pixel is 0.5, and if you ask for the pixel at 0.5 you get the pixel's value - but if you sample it at, say, 0 (left edge) - you get a 50% mix of the left pixel and 50% mix of this pixel. By shifting where we sample, we can adjust the ratio of how much of each pixel the hardware will mix for us. The white small numbers indicate where we're sampling within that pixel.

The new algorithm will do something like this:
- Sample the left most pixel inside the blur, with its usual weight
- For every pair of (1, 1) weighted pixels, combine them into one sample with twice the weight
- If there's remaining pixels, sample it (if 1) or between them (if 2) with their combined weights

In addition to the half as much sampling, we switch to push constants where possible and do any impactful calculations ahead of time, so the overhead of blur itself should go down.

Finally, because this now supports even-width kernels, it should look more correct!


render/wgpu/shaders/filter/blur.wgsl
```diff
@@ -23,31 +23,34 @@ struct Filter {
...
+        total += textureSample(texture, texture_sampler, in.uv + (direction * (filter_args.full_size - 0.75))) * 1.5;
     }
 
     // At this point, the center_length must be a whole number, divisible by 2.
-    center_length /= 2.0;
-    for (var i = 0.0; i < center_length; i += 1.0) {
+    for (var i = 0.0; i < center_length; i += 2.0) {
         // The center of the kernel is always going to be 1,1 weight pairs. We can just sample between the two pixels.
-        total += textureSample(texture, texture_sampler, left_uv + (direction * (1.5 + (i * 2.0)))) * 2.0;
+        total += textureSample(texture, texture_sampler, in.uv + (direction * (1.5 + i))) * 2.0;
```
torokati44, line 53: Have you tried to move the last `* 2.0` to after the hot loop? (With a separate `center_total` accumulator of course .. or, well, we could pre-divide `total` by 2.0 so it all works out in the end...) If so, did it help at all?

torokati44:
I couldn't quite grok the benefits of the last commit (shifting UVs), or what it does - could you please explain it a little bit?

Dinnerbone:
> I couldn't quite grok the benefits of the last commit (shifting UVs), or what it does - could you please explain it a little bit?

It gets the uvs of the area on the source image it's going to operate on, usually 0 to 1. This works out such that the uv received by any single fragment is the center of the blur kernel.

We always operate on the left kernel edge and work our way right though, which we did by calculating the left uv from the center and then adding everything to that left uv.

By doing that same calculation inside the vertex shader, we're preshifting the entire coordinates by (distance between left edge and center), so that the uv received by each fragment is no longer the center of the kernel, but the left edge of the kernel.

It just means that we don't need to keep recalculating that shift every fragment, which frees up some cycles.

Another approach of course is to shift everything inside the vertices given by rust instead, but then we would always need to make a new vertex buffer instead of being able to use a constant quad like we do today.

render/wgpu/shaders/filter/blur.wgsl
Dinnerbone: I haven't been able to microbenchmark the nuances within the fragment shader itself, but I think that a *2 in the loop shouldn't make much of a difference with modern shader compilers. I think when it was *(i*2), maybe, but now it feels unnecessary 😅

Dinnerbone: Decided to do it anyway, can't hurt!

torokati44 (approved):

Issue #11821, opened 2023-07-01
Some errors (neave_tv.swf 2006, FLV works one time only)


TemaisgameNoobas:
### Describe the bug

Launched Neave.TV with new FLV support, at first it displays correctly, but when i try to switch the video this happens:
![image](https://github.com/ruffle-rs/ruffle/assets/60001674/494860c7-356c-4bd5-8b25-0c4e6f3c5cb1)
![image](https://github.com/ruffle-rs/ruffle/assets/60001674/7007c75b-e750-45a5-8ae6-0a724c32688e)
SWF:
[issue.zip](https://github.com/ruffle-rs/ruffle/files/11947306/issue.zip)

### Expected behavior

Playing videos fine

### Affected platform

Self-hosted version

### Operating system

Ubuntu 22.04

### Browser

UnGoogled Chromium Version 100.0.4896.127

### Additional information

Ruffle was unmodified, and it plays first video fine.

n0samu:
Note for people looking at this issue: `adhis.mp4` and `format.flv` are the FLV videos that it loads. Yes, adhis.mp4 is actually an FLV video. The problem is that selecting the second video ("FLV Sample") doesn't show the video (which is just text on a green background)
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/fd9cb7f9-807b-46c2-9cab-9fcc2339126f)

TemaisgameNoobas:
> This issue has progress. Now, a sign appears that reads "Sorry, it seems like _video_ is on slow conection at the moment". It switches the video every 10 seconds. ![image](https://user-images.githubusercontent.com/60001674/262835387-7e86ba0b-fda7-4328-b76e-5f0963fc36ec.png)

Im going to open a seperate issue for this.

Issue #5704, opened 2021-11-22
aperturescience.com: DefineVideoStream, got Invalid data: Invalid video codec.


FoxMcCloud45:
### Describe the bug

https://www.aperturescience.com uses an old SWF file made in 2010.

In Mozilla Firefox, the first frame of the video shows but nothing else happens.
In Google Chrome, the Flash window shows a PLAY button, clicking it causes a black render.

The website's SWF file is hosted at https://www.aperturescience.com/aperturescience.swf

Also provided below.
[aperturescience.zip](https://github.com/ruffle-rs/ruffle/files/7584238/aperturescience.zip)
 .

### Expected behavior

The SWF's video should play normally.

### Affected platform

Browser's extension

### Operating system

Microsoft Windows 10 LTSC 2019 (1809)

### Browser

Mozilla Firefox 91.3.0esr

### Additional information

Firefox's JavaScript console log is attached below.
[console-export-2021-11-22_21-27-39.txt](https://github.com/ruffle-rs/ruffle/files/7584230/console-export-2021-11-22_21-27-39.txt)

n0samu:
This SWF uses an external video file. Ruffle doesn't support loading external video/audio/image assets yet.

vladislavkoz:
Hello guys. I have SWF which uses some FLV video files. As I understand support for FLV is not added yet. And as I understand, we can't use any video files. Is it correct? In my case, I have the following error: 
![image](https://user-images.githubusercontent.com/40685761/160382402-d8cb1289-1dec-4840-9e4f-f1cd1611b3ab.png)

Is there any scheduled work on adding video files support then? Could you propose any workaround for this before video support is added?

danielhjacobs:
Embedded H.263 and VP6(A) videos work. FLV's/external videos do not work.

vladislavkoz:
> A)

So does it mean that we can't use external videos at all?

danielhjacobs:
Correct. You need to embed an H.263 or VP6 video in the swf, like this: https://dnalc.cshl.edu/view/962-Causes-Smoking-Prevention.html

Lord-McSweeney:
Video has been implemented- however, this seems to need something else to work, as it currently just shows a single frame. Stub warning: `Stub: Stream data processing (name: onXMPData)`

n0samu:
As I mentioned in https://github.com/ruffle-rs/ruffle/pull/12091#issuecomment-1674177630:
[...]

PR #12091, opened 2023-07-15
FLV audio +Buffer refactor

kmeisthax:
This PR adds a new set of types for managing data buffers in use by Ruffle's various backends:

 * `Buffer` stores the whole of a given data stream, whether that be a SWF file, FLV file, MP3 file, and so on. Data ownership is shared. Mutability is limited to appending data to the end of the buffer.
 * A `Slice` can be constructed from a range on the `Buffer`. Slices are guaranteed to always refer to the same data.
   * The `Slice` can be indexed as a `[u8]` by turning it into a `SliceRef<'_>`. This is equivalent to holding a `&[u8]`.
   * Data can be read from the `Slice` by turning it into a `SliceCursor` which implements `Read`.
 * Multiple `Slice`s can be appended to a single `Substream`, which represents a single embedded data stream in a `Buffer`. The slice list in the `Substream` is also shared and can be appended to.
   * A `Substream` can be iterated as a sequence of `Slice`s by turning it into a `SubstreamChunksIter`.
   * A `Substream` can be read from as a file by turning it into a `SubstreamCursor` which implements `Read`.

This mirrors the existing hierarchy of `SwfMovie` and `SwfSlice`, which will eventually be changed to use `Buffer` and `Slice` under the hood. The purpose of that is to enable progressive downloading of SWF files (eventually, pending implementation of streaming decompressors, etc). This switch has not been done yet of course.

The current motivation is to support FLV audio. Currently, the audio backend can only stream data from an SWF tagstream, because the stream reader logic is buried deep into the audio decoder. This PR actually clones all that logic; the intent is to replace the existing streaming sounds pipeline entirely with the `Substream` based pipeline.

The new audio API for all of this is actually super-simple: just a single `start_substream` method. `Substream` itself has an internally mutable chunks list, so appending to it will make playback happen as long as the sound is still playing. If it's been stopped, you can check for that with `is_sound_playing` and restart the stream if so.

This PR successfully plays the audio on https://www.nothingtodo.co.uk/view/3136/who-needs-a-glove.html without issue. Note that the transport controls will still not work since we don't support seeking.

### DIFF
--- a/core/src/backend/audio/decoders.rs
+++ b/core/src/backend/audio/decoders.rs
@@ -31,6 +31,9 @@ pub enum Error {
 
     #[error("Unhandled compression {0:?}")]
     UnhandledCompression(AudioCompression),
+
+    #[error("Too many sounds are playing")]
+    TooManySounds,
 }
 
 /// An audio decoder. Can be used as an `Iterator` to return stero sample frames.
--- a/core/src/backend/audio/decoders.rs
+++ b/core/src/backend/audio/decoders.rs
@@ -472,7 +472,7 @@ impl Read for SubstreamTagReader {
         buf[..len].copy_from_slice(&data[..len]);
 
         drop(data);
-        *chunk = chunk.to_start_and_end(chunk.start() + len, chunk.end());
+        *chunk = chunk.to_start_and_end(len, 0);
 
         if chunk.is_empty() {
             self.current_audio_data = None;
--- a/core/src/backend/audio/decoders.rs
+++ b/core/src/backend/audio/decoders.rs
@@ -14,6 +14,7 @@ pub use mp3::{mp3_metadata, Mp3Decoder};
 pub use nellymoser::NellymoserDecoder;
 pub use pcm::PcmDecoder;
 
+use crate::buffer::{Slice, SliceCursor, Substream, SubstreamChunksIter};
 use crate::tag_utils::{ControlFlow, SwfSlice};
 use std::io::{Cursor, Read};
 use swf::{AudioCompression, SoundFormat, TagCode};
@@ -238,9 +239,6 @@ pub trait SeekableDecoder: Decoder {
 /// `StreamTagReader` reads through the SWF tag data of a `MovieClip`, extracting
 /// audio data from the `SoundStreamBlock` tags. It can be used as an `Iterator` that
 /// will return consecutive slices of the underlying audio data.
-/// `StreamTagReader` reads through the SWF tag data of a `MovieClip`, extracting
-/// audio data from the `SoundStreamBlock` tags. It can be used as an `Iterator` that
-/// will return consecutive slices of the underlying audio data.
 struct StreamTagReader {
     /// The tag data of the `MovieClip` that contains the streaming audio track.
     swf_data: SwfSlice,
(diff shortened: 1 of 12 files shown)


TemaisgameNoobas:
When someone could review this??

core/src/streams.rs
```diff
@@ -285,10 +327,434 @@ impl<'gc> NetStream<'gc> {
...
+    fn commit_sound_stream(
+        self,
+        context: &mut UpdateContext<'_, 'gc>,
+        write: &mut NetStreamData<'gc>,
+    ) -> Result<(), NetstreamError> {
+        if !Self::sound_currently_playing(context, &write.sound_instance) {
+            if let Some((substream, sound_stream_head)) = &mut write.audio_stream {
+                let attached_to = write.attached_to;
+
+                if context.is_action_script_3() {
```
Lord-McSweeney, line 495: I'd prefer this to check if `avm_object` is `Some(Avm2Object(...))` instead of using `context.is_action_script_3`, for mixed AVM.

Lord-McSweeney: (Though I have no knowledge of the audio manager- if this does actually depend on the root movie type, `context.swf.is_action_script_3` is fine).

n0samu:
Can confirm that this successfully plays the FLV audio from #11921! But it fails to play the FLV audio from #12600; this error is repeated:
`ERROR ruffle_core::streams: Error committing sound stream: AVM1 NetStream not attached to MovieClip`

core/src/streams.rs
kmeisthax: I thought this was the correct way to check - and we check `context.is_action_script_3()` in at least 90 places. I specifically remember us removing `avm_object` checks for context type checks; if this is going to be a pain for mixed AVM we probably should be having a separate PR to remove these checks separately.

n0samu:
> `ERROR ruffle_core::streams: Error committing sound stream: AVM1 NetStream not attached to MovieClip`
This same problem also affects [Morningstar](https://ooooooooo.ooo/static/?7a673991-f6e5-4c0f-86e1-6dc1b14cfe80).

kmeisthax:
So I've decided to not require streams be attached to a movie clip in order to play sound. I'm not 100% sure if this'll cause problems, but it at least should fix those movies.

n0samu:
The problems I mentioned are indeed fixed now, thank you!
I think there is something unexpected going wrong in #5704 though. The video shows only the first frame, and more strangely, the audio plays but periodically glitches out. Might there be some script data interleaved with the audio that Ruffle is erroneously trying to play back?

TemaisgameNoobas:
Can someone test this SWF?? The SWF could be found in this issue: #11821

n0samu:
I already tested it. #12502 is solved but #11821 is not.

TemaisgameNoobas:
@n0samu Thanks, I need @kmeisthax 's attention on this issue: #11821

kmeisthax:
FLV switching *should* be fixed but Neave doesn't seem to work in desktop.

The problem was that nothing actually resets `NetStream` state if we load a second movie into it, so it tries to play it as if it both movies were part of the same file, which isn't how anything works. I've added that code so it should work but I haven't verified it yet.

kmeisthax:
Aperture Science isn't displaying video because that movie is waiting for an event from MX's video player that doesn't come. The event handler is responsible for actually putting the video on the screen. I suspect MX is waiting on a NetStream event we don't implement yet. The audio corruption I've still yet to figure out.

TemaisgameNoobas:
> FLV switching _should_ be fixed but Neave doesn't seem to work in desktop.
> 
> The problem was that nothing actually resets `NetStream` state if we load a second movie into it, so it tries to play it as if it both movies were part of the same file, which isn't how anything works. I've added that code so it should work but I haven't verified it yet.

Web version should be working!!

n0samu:
> FLV switching should be fixed but Neave doesn't seem to work in desktop.

Just tested it with your PR on desktop and it seems to be working fine!

Lord-McSweeney:
The issue with Aperture Science is that it expects a NetStream.Play.Stop `onStatus` event to come, but for some reason it never appears.

TemaisgameNoobas:
Is this PR is ready to be reviewed and merged?? I am actively waiting this to be merged because it is nice to hear audio in FLVs :)

core/src/streams.rs
```diff
@@ -500,9 +494,7 @@ impl<'gc> NetStream<'gc> {
                                     context.renderer,
                                 ) {
                                     Ok(bitmap_info) => {
-                                        let (_, position) = reader.into_parts();
```
torokati44, line 503: Ooh, it's nice to get rid of all this deconstructing-reconstructing reader.

core/src/buffer.rs
```diff
@@ -88,8 +91,8 @@ impl Buffer {
         if start <= s.len() && end <= s.len() && start <= end {
             Some(Slice {
                 buf: Self(self.0.clone()),
-                start: start,
-                end: end,
+                start,
+                end,
```
torokati44: This was probably clippy fallout from the previous commit, but I'm NOT going to nit-pick on this sort of thing here.

core/src/buffer.rs
```diff
@@ -242,6 +257,167 @@ impl Slice {
     }
 }
 
+/// A list of multiple slices of the same buffer.
+///
+/// `Substream` represents a substream of the underlying buffer. Slices can be
+/// appended to the chunks list
```
torokati44: This sentence looks unfinished ... ? Anyway, it's not in the final code, so it's not important.

TemaisgameNoobas:
This PR is still under fixing bugs?? I guess Aperture Science still doesn't work, but the PR is open

core/src/backend/audio/decoders.rs
```diff
@@ -235,9 +236,6 @@ pub trait SeekableDecoder: Decoder {
     }
 }
 
-/// `StreamTagReader` reads through the SWF tag data of a `MovieClip`, extracting
-/// audio data from the `SoundStreamBlock` tags. It can be used as an `Iterator` that
-/// will return consecutive slices of the underlying audio data.
```
torokati44, line 240: This deduplication confused me terribly... turns out, this doc stutter has been lurking in here for three years now... :sweat_smile:

core/src/streams.rs
```diff
@@ -393,14 +402,126 @@ impl<'gc> NetStream<'gc> {
...
+                                ..
+                            }) if context.audio.is_sound_playing(*audio_handle) => {
+                                substream.clone()
+                            }
+                            Some(NetStreamType::Flv {
+                                audio_stream, //None or not playing
+                                ..
+                            }) => {
+                                let substream = Substream::new(slice.buffer().clone());
+                                let audio_handle = (|| {
```
torokati44: This might be a technicality that I'm unaware of, but asking anyway out of curiosity: What's the point of an instantly called lambda here, as compared to a simple block?

core/src/display_object/movie_clip.rs
```diff
@@ -2485,6 +2493,25 @@ impl<'gc> MovieClip<'gc> {
...
+        context: &mut UpdateContext<'_, 'gc>,
+        netstream: Option<NetStream<'gc>>,
+    ) {
+        let mut write = self.0.write(context.gc_context);
+        if netstream != write.attached_audio {
+            if let Some(old_netstream) = write.attached_audio {
+                old_netstream.was_detached(context);
+            }
+
+            write.attached_audio = netstream;
```
torokati44, line 2508: I suppose the notification about "was detached" before actually detaching it is not likely to cause any weird unintended side effects... If it's possible at all.

core/src/streams.rs
kmeisthax: I wanted to use `?` in a function that does not return a `Result`. Didn't know you could have blocks that return like functions, I'll try that next time.

AFAIK this code is gone now so I'm just going to mark this as resolved.

core/src/buffer.rs
kmeisthax: I think you might have reviewed an old version of the PR - I caught and fixed this a little while ago and the fixed documentation is in my tree.

core/src/display_object/movie_clip.rs
kmeisthax: We don't fire a NetStream event for this, so the only code that can even observe this order of events is our own.

core/src/streams.rs
torokati44: > I wanted to use ? in a function that does not return a Result. Didn't know you could have blocks that return like functions, I'll try that next time.

Turns out, it's not a `return`, but a `break`. And it's not that old: https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html#break-from-labeled-blocks

> AFAIK this code is gone now so I'm just going to mark this as resolved.

I'm going commit-by-commit, as I find it easier to understand the changes that way. Also I'm interested in how the commit history looks.

core/src/display_object/movie_clip.rs
torokati44: Alright!

core/src/buffer.rs
torokati44: Like I said, I'm going commit-by-commit with the review... I know it has unfortunate side effects like this, sorry.

core/src/backend/audio/decoders.rs
```diff
@@ -472,7 +472,7 @@ impl Read for SubstreamTagReader {
         buf[..len].copy_from_slice(&data[..len]);
 
         drop(data);
-        *chunk = chunk.to_start_and_end(chunk.start() + len, chunk.end());
+        *chunk = chunk.to_start_and_end(len, 0);
```
torokati44, line 475: I'm confused, how is this ever going to be not be empty?

core/src/streams.rs
```diff
@@ -378,114 +383,147 @@ impl<'gc> NetStream<'gc> {
...
+    fn collect_sound_stream(
+        self,
+        context: &mut UpdateContext<'_, 'gc>,
+        write: &mut NetStreamData<'gc>,
+    ) {
+        if !Self::sound_currently_playing(context, &write.sound_instance) {
+            write.audio_stream = None;
+            write.sound_instance = None;
+        }
+    }
```
torokati44, line 529: The name confused me at first. I thought "collect" meant "pick back up" - since the comment says "we need to restart it". But it's actually collect as in GC, right? Wouldn't "cleanup" be a better verb in the name - after all, the doc comment literally starts with these words.

core/src/streams.rs
```diff
@@ -378,114 +383,147 @@ impl<'gc> NetStream<'gc> {
...
+
+    /// Ensure that if we have queued up audio into a sound stream, that said
+    /// stream gets sent over to the audio backend.
+    ///
+    /// Intended to be called at the end of tag processing. Audio processing
+    /// should occur only after a minimum number of tags have been processed to
+    /// avoid audio underruns.
```
torokati44, line 536: What determines this "minimum number" of tags to process? Is it the same as `max_lookahead_audio_tags`?

core/src/streams.rs
```diff
@@ -269,6 +271,44 @@ impl<'gc> NetStream<'gc> {
...
 
+    /// Initialize the `NetStream` buffer to accept new source data.
+    ///
+    /// This must be done once per source change and should ideally be done
+    /// immediately before the first `load_buffer` call for a particular source
+    /// file.
+    ///
+    /// Externally visible AVM state must not be initialized here - i.e. the
+    /// AS3 `client` doesn't go away because you played a new video file.
+    pub fn init_buffer(self, context: &mut UpdateContext<'_, 'gc>) {
```
torokati44, line 282: Isn't this actually a "clear" or a "reset" operation? :thinking:

torokati44:
Okay, I've read through all the changes here, and aside from my remaining comments above, plus one question in Discord, there's nothing standing out to me that should prevent merging. I trust @kmeisthax that this architecture is good.

The only thing is, that the commit history is not as clean as I would like (as in, it has a few "adding foo" and later "foo was buggy, fix foo"; or "add bar" and later "turns out, bar was not needed, remove it" moments), but I recognize that it's often much more effort to clear up, than what it's worth, and it's not that bad, nor important.

core/src/backend/audio/decoders.rs
kmeisthax: You're right, this throws out the rest of the current chunk. Unfortunately this isn't the cause of the audio glitching in Aperture Science :/

core/src/streams.rs
kmeisthax: There's no particular criteria, it just has to be 'enough'. And in practice it will be `max_lookahead_audio_tags` unless there's something strange going on. When I implement progressive download we may need to enforce this.

core/src/streams.rs
kmeisthax: Yes

kmeisthax:
The only architectural question I have remaining is how we handle the frame/sample counting logic that I didn't copy over to `SubstreamTagReader`. I suspect I'll have to add the ability to insert "frame gaps" in a `Substream` so we can do that. But that'll have to wait until we move SWFs over to the new system.

torokati44 (approved):

TemaisgameNoobas:
@kmeisthax this PR would not be merged untill the ability to insert "frame gaps" would be added??

kmeisthax:
No, because we're not moving SWFs over to the new audio API just yet. As far as I'm aware FLVs don't care about frame gaps... in fact, they don't even have frames, everything is indicated with time offsets. We'll discover all the fun with that in a later PR.

PR #12095, opened 2023-07-16
avm2: Fix handling of Vector.<*>

Aaron1011:
We were previously using the Vector.<Object> `Class` with two different `ClassObjects` (`Vector.<*>` and `Vector.<Object>`), which resulted in the class initializer getting skipped for a specialized class.

We also turned `Vector.<*>` into `Vector.<Object>` in the Vector class call handler, resulting in getQualifiedClassName printing `Vector.<Object>` instead of `Vector.<*>`. We now correctly apply the class in the call handler (but still use a value type of `Object` in `VectorStorage` for `Vector.<*>`)


core/src/avm2/class.rs
```diff
@@ -78,7 +78,7 @@ pub struct Class<'gc> {
     name: QName<'gc>,
 
     /// The type parameter for this class (only supported for Vector)
-    param: Option<GcCell<'gc, Class<'gc>>>,
+    param: Option<Option<GcCell<'gc, Class<'gc>>>>,
```
moulins, line 81: The double `Option` is pretty confusing here; I'd suggest adding a dedicated type, e.g.:
```rs
pub enum TypeParam<'gc> {
    Any,
    Class(GcCell<'gc, Class<'gc>>),
}

impl Eq for TypeParam<'gc> { ... }
```

adrian17:
Fully subsumed by https://github.com/ruffle-rs/ruffle/pull/12132 ,  so closing.

PR #11005, opened 2023-05-11
core: Support mixed AVM

Lord-McSweeney:
There are still some issues (see the unchecked checkbox), and I'm unsure about how accurate the mouse pick implementation is, but content should at least no longer panic.

Fixes #9330, fixes #9724, fixes #10400, fixes #8941, fixes #11668, fixes #10546, fixes #9937, fixes #12005, fixes #12568, fixes #10138, fixes #10740, fixes #12341, fixes #12407, fixes #11570, fixes #10002, fixes #9666, fixes #11278, fixes #14136, fixes #13917, fixes #13371, fixes #13548, fixes #13211, fixes #13506, fixes #9437, ...

The animation on #10175 plays correctly with this PR, though, as mentioned, it layers as a result of `Loader.unloadAndStop` being unimplemented.

KOF vs DNF (#9917) plays fine when this PR is combined with #12823, though it spews several harmless warnings and errors (a result of the obfuscation).

TODO/list of working mixed AVM games:

- [X] Anbot (#10546)
- [X] #9937
- [X] #8941
- [X] #9330
- [X] Tank Destroyer (#9724)
- [X] #10400
- [X] #11668
- [X] #12005
- [X] #12568
- [x] Mine Blocks (#10002)
- [x] Feeding Frenzy (#9666) (minor issue remaining- text that appears after eating a smaller fish isn't text, just a bar)
- [x] Christmas Bloons (#11570)
- [ ] Doodle Jump (#10759) (game over screen when touching an enemy is shifted (probably caused by `RemoveSprite: Source is not a display object` spam))
- [x] Mad Karate Man (#10740)
- [x] Traps and Sorcery (#12407)
- [x] Garden Raider (#12341)
- [x] Add test
- [ ] More tests? (not merge-blocking)

PR #12102, opened 2023-07-16
avm2: Remove focus from an EditText when it unloads (close #12092)

### LINKED ISSUE #12092: Homestuck "John: Enter Village" - Softlock after textbox disappears
### Describe the bug

https://mrcheeze.github.io/homestuck-ruffle-tester/#03077 / https://mrcheeze.github.io/homestuck-ruffle-tester/flash/03077.swf

To reproduce:

* Walk north, east, north, east, with WASD or arrow keys to get to this area:
image
* Press Ctrl-Shift-T to open the secret trickster code textbox.
* Type the code shoosh:doof
* As soon as you do, the textbox will instantly disappear (without having to press enter, etc) and you transform into a different character.

The bug is that after the textbox disappears, you are supposed to regain control and be able to move around again. Instead, all keyboard input is completely ignored after it disappears, and you are softlock.

### Expected behavior

As above, but after entering shoosh:doof and turning into a different character, you should be able to move with WASD/arrow keys and attack with space (as you were before).

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

Mozilla Firefox

### Additional information

_No response_

MrCheeze:
Since https://github.com/ruffle-rs/ruffle/pull/9911 , a textbox that unloads in AVM2 has incorrectly been holding the focus, preventing you from using keyboard input for anything else (see https://github.com/ruffle-rs/ruffle/issues/12092 ).

Since there is no avm2_unload function as such, I copied what [movie_clip](https://github.com/ruffle-rs/ruffle/blob/cd2d01bf6f2565e1aa442da671ca7e27ed186629/core/src/display_object/movie_clip.rs#L2718) and [loader_display](https://github.com/ruffle-rs/ruffle/blob/cd2d01bf6f2565e1aa442da671ca7e27ed186629/core/src/display_object/loader_display.rs#L145) do, of putting the AVM2 "unload-like" logic inside set_parent.

Extra note: I found that:
* parent.removeChild(textfield) removes the focus but does not fire a FOCUS_OUT event
* RemoveObject on the timeline removes the focus *and* fires a FOCUS_OUT event

Which doesn't impact this PR, since FocusEvent isn't implemented at all, but is somewhat surprising....


core/src/display_object/edit_text.rs
```diff
@@ -1681,6 +1681,21 @@ impl<'gc> TDisplayObject<'gc> for EditText<'gc> {
         self.0.write(context.gc_context).object = Some(to.into());
     }
 
+    fn set_parent(&self, context: &mut UpdateContext<'_, 'gc>, parent: Option<DisplayObject<'gc>>) {
+        let had_parent = self.parent().is_some();
+        self.base_mut(context.gc_context)
+            .set_parent_ignoring_orphan_list(parent);
+        let has_parent = self.parent().is_some();
+
+        if context.is_action_script_3() && had_parent && !has_parent {
```
Lord-McSweeney, line 1690: `self.movie().is_action_script_3` for mixed AVM

MrCheeze: Updated. That makes the condition here temporarily different from in movie_clip set_parent, but https://github.com/ruffle-rs/ruffle/pull/11005 makes the conditions the same again.

core/src/display_object/edit_text.rs
```diff
@@ -1681,6 +1681,21 @@ impl<'gc> TDisplayObject<'gc> for EditText<'gc> {
         self.0.write(context.gc_context).object = Some(to.into());
     }
 
+    fn set_parent(&self, context: &mut UpdateContext<'_, 'gc>, parent: Option<DisplayObject<'gc>>) {
+        let had_parent = self.parent().is_some();
+        self.base_mut(context.gc_context)
+            .set_parent_ignoring_orphan_list(parent);
+        let has_parent = self.parent().is_some();
```
Dinnerbone, line 1688: Nit: Is there any scenario where `parent` != `self.parent()` on this line? Save's us a bit of indirection 😅

Dinnerbone (approved):

Issue #10176, opened 2023-03-17
Rogue Soul 2: game is broken after second start


MartySVK:
### Describe the bug

Start the game and tutorial. Choose the control and close the game. When you start it again, the game is broken, and you cant jump.
On the first try, you can play the game in other levels as well, the game will break only when it is closed and reopened.
I see this error:

```
ERROR ruffle_core::avm2::activation: AVM2 error: Object(ErrorObject(ErrorObject { class: TypeError, ptr: 0x217a41ac4c0 }))
ERROR ruffle_core::avm2::activation: TypeError: Error #1010: A term is undefined and has no properties. (accessing field: bandit)
later
ERROR ruffle_core::avm2::activation: AVM2 error: Object(ErrorObject(ErrorObject { class: ReferenceError, ptr: 0x217a48e1020 }))
ERROR ruffle_core::avm2::activation: ReferenceError: Error #1069: Property format not found on flash.globalization.CurrencyFormatter and there is no default value.
```

But maybe there are some invalid save files ?

SWF - https://data.gameflare.com/games/5581/f4uZCsPtJl1zMz.swf

### Expected behavior

Controls should work after the game is re-opened.

### Affected platform

Desktop app

### Operating system

Windows 10

PR #12108, opened 2023-07-17
avm2: Stub flash.globalization.CurrencyFormatter.format

MartySVK:
fixes Rogue Soul 2 https://github.com/ruffle-rs/ruffle/issues/10176

### DIFF
--- a/core/src/avm2/globals/flash/globalization/CurrencyFormatter.as
+++ b/core/src/avm2/globals/flash/globalization/CurrencyFormatter.as
@@ -11,8 +11,8 @@ package flash.globalization {
         }
 
         public function format(value:Number, withCurrencySymbol:Boolean = false):String {
-            return value.toString();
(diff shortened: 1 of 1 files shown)


core/src/avm2/globals/flash/globalization/CurrencyFormatter.as
```diff
@@ -8,5 +9,10 @@ package flash.globalization {
         public function CurrencyFormatter(locale: String) {
             stub_constructor("flash.globalization.CurrencyFormatter");
         }
+
+        public function format(value:Number, withCurrencySymbol:Boolean = false):String {
+            return value.toString();
+            stub_method("flash.globalization.CurrencyFormatter", "format");
```
torokati44, line 15: Shouldn't this come before the `return` statement...? In other stubbed functions they are in that order.

adrian17 (approved):

PR #12132, opened 2023-07-17
avm2: Refactor Vector.<T> class creation

adrian17:
Intended to subsume https://github.com/ruffle-rs/ruffle/pull/12095 .

This rewrites creation of Vector classes.
Compared to before, where we constructed the entire class from scratch, now we use the standard ClassObject creation facilities and only do Vector-specific mutations (`param`, `applications`, `call_handler`) afterwards.
This also fixes the class hierarchy - `Vector.<int/uint/Number/*>` are toplevel classes, while any other `Vector.<T>` simply derives from `Vector.<*>`. And `Vector` is just empty.
This lets us remove `specialized_class_init` and all the incorrect logic in `has_class_in_chain` and `coerce_to_type`.

I wish there was less duplication (creating a class and setting a param/application in very different ways from two different places), but these two class groups (builtin vs runtime-applied) are so fundamentally different that I'm not sure if that's possible.


core/src/avm2/object/class_object.rs
```diff
@@ -908,10 +919,6 @@ impl<'gc> TObject<'gc> for ClassObject<'gc> {
             return Err(format!("Class {:?} is not generic", self_class.read().name()).into());
         }
 
-        if !self_class.read().param().is_none() {
-            return Err(format!("Class {:?} was already applied", self_class.read().name()).into());
-        }
-
         //Because `null` is a valid parameter, we have to accept values as
         //parameters instead of objects. We coerce them to objects now.
         let object_param = match nullable_param {
```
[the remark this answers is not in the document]
adrian17: Can do 👍

Dinnerbone:
This makes `specialized_at_runtime` begin to pass from #12293

core/src/avm2/globals.rs
```diff
@@ -83,7 +83,8 @@ pub struct SystemClasses<'gc> {
     pub sprite: ClassObject<'gc>,
     pub simplebutton: ClassObject<'gc>,
     pub regexp: ClassObject<'gc>,
-    pub vector: ClassObject<'gc>,
+    pub generic_vector: ClassObject<'gc>,
+    pub object_vector: ClassObject<'gc>, // Vector.<*>
```
Lord-McSweeney, line 87: Just checking- is this comment on the correct line? Isn't `Vector.<*>` the generic vector?

core/src/avm2/vector.rs
```diff
@@ -265,21 +285,22 @@ impl<'gc> VectorStorage<'gc> {
...
-            Some(v) => Ok(v),
-            None if Object::ptr_eq(self.value_type(), activation.avm2().classes().uint) => {
-                Ok(Value::Integer(0))
-            }
-            None if Object::ptr_eq(self.value_type(), activation.avm2().classes().int) => {
-                Ok(Value::Integer(0))
+            if let Some(value_type) = self.value_type() {
+                if Object::ptr_eq(value_type, activation.avm2().classes().uint) {
+                    Ok(Value::Integer(0))
+                } else if Object::ptr_eq(value_type, activation.avm2().classes().int) {
```
Lord-McSweeney, line 294: Should the `int` and `uint` if-statements be meged into a single statement?

core/src/avm2/globals.rs
adrian17: Actually no.
`generic_vector` refers to `Vector` itself. This is the only type with the `GENERIC` flag.
`object_vector` refers to `Vector<*>` (which is different but almost-identical to `Vector<Object>`).

Feel free to offer alternative naming :D

core/src/avm2/vector.rs
adrian17: They could, just thought it'd be slightly cleaner this way.

core/src/avm2/globals.rs
danielhjacobs: `untyped_vector` (matches `*` definition on https://web.archive.org/web/20171020231504/https://help.adobe.com/en_US/LiveCycleMosaic/9.0/actionscript/specialTypes.html) instead of `object_vector` maybe.

And just to avoid confusion, maybe `general_vector` instead of `generic_vector`, since in other langauges the star is called a generic.

Dinnerbone: Maybe `typed_vector` and `untyped_vector`? (If I understand the difference correctly!)

danielhjacobs: Scratch my suggestion I think, in the case of vectors the definition of the `*` doesn't match that page.

Aaron1011 (approved):

core/src/avm2/class.rs
```diff
@@ -78,7 +78,7 @@ pub struct Class<'gc> {
     name: QName<'gc>,
 
     /// The type parameter for this class (only supported for Vector)
-    param: Option<GcCell<'gc, Class<'gc>>>,
+    param: Option<Option<GcCell<'gc, Class<'gc>>>>,
```
Dinnerbone, line 81: This is a funny type and (without reading the usages, I admit) I'm struggling to come up with a meaning of the three potential states.

Can we maybe change this to a 3 state enum so we know what `Some(None)` vs `None` actually means?

core/src/avm2/object/class_object.rs
```diff
@@ -908,100 +919,48 @@ impl<'gc> TObject<'gc> for ClassObject<'gc> {
...
             Value::Null => None,
             Value::Undefined => return Err("Undefined is not a valid type parameter".into()),
```
Lord-McSweeney, line 926: The case for `Value::Undefined` should be removed: `Vector.<undefined>` has the same error as `Vector.<3>` in Flash.

core/src/avm2/class.rs
Dinnerbone: (not blocking PR on this, can fix up later!)

core/src/avm2/globals.rs
adrian17: Left the names as-is, expanded the comments.

Dinnerbone (approved):

Issue #12118, opened 2023-07-17
Test failure: shumway's acid-shapes-testing


Dinnerbone:
Test path: `tests/tests/swfs/visual/shumway_acid_tests/acid_shapes_testing`

# Expected
![image](https://github.com/ruffle-rs/ruffle/assets/317625/d2716912-4bfb-4c2f-9bed-814f9f3568f6)

# Actual
![image](https://github.com/ruffle-rs/ruffle/assets/317625/cc907861-b132-4dfb-8631-d021cd87b6d3)

# Difference
![image](https://github.com/ruffle-rs/ruffle/assets/317625/37817c45-f892-4564-82b1-0fb19c2d4338)

PR #12148, opened 2023-07-18
core: Calculate correct bounds for curves through the drawing API

### LINKED ISSUE #12117: Test failure: shumway's acid-shapes
Test path: `tests/tests/swfs/visual/shumway_acid_tests/acid_shapes`

# Expected
image

# Actual
image

# Difference
image

Dinnerbone:
This fixes #12117 
(and makes progress on #12118 but doesn't fix the trace output entirely)


core/src/drawing.rs
```diff
@@ -441,17 +445,53 @@ fn stretch_bounds(
...
+            let anchor_y = anchor.y.to_pixels();
+            let control_x = control.x.to_pixels();
+            let control_y = control.y.to_pixels();
+
+            let mut min_x = from_x.min(anchor_x);
+            let mut min_y = from_y.min(anchor_y);
+            let mut max_x = from_x.max(anchor_x);
+            let mut max_y = from_y.max(anchor_y);
+
+            if control_x < min_x || control_x > max_x || control_y < min_y || control_y > max_y {
```
adrian17, line 470: we could split the entire math to X and Y guarded scopes, right? As in
```
if control_x < min_x || control_x > max_x  {
    // calculate and update min/max_x
}
if control_y < min_y || control_y > max_y {
    // calculate and update min/max_y
}
```

adrian17 (approved):

Issue #12107, opened 2023-07-17
Ancient Odyssey - Can't click play button


desertking:
### Describe the bug

This happens for every Abroy game with the same intro.

Loading SWF file https://www.spiele-umsonst.de/azad/downloads/flashgames/ancientodyssey.swf

### Expected behavior

Click Play and start

### Affected platform

Online demo

### Operating system

All

### Browser

All

Lord-McSweeney:
Note that the object(s) used for the buttons are MovieClips- there are no Avm2Buttons.

PR #12150, opened 2023-07-19
avm2: Use invisible mask descendants for mouse picking and rendering

Aaron1011:
The 'DisplayObject.visible' flag seems to be ignored for mask objects (and their descendants) - mouse picking still takes them into account, and they get rendered as part of the mask.

This commit changes display object rendering and AVM2 mouse picking to ignore DisplayObject.visibility when dealing with mask objects/ descendants.

Fixes https://github.com/ruffle-rs/ruffle/issues/12107


core/src/display_object/container.rs
```diff
@@ -485,8 +485,10 @@ pub trait TDisplayObjectContainer<'gc>:
                 context.commands.push_mask();
                 child.render(context);
                 context.commands.activate_mask();
-            } else if child.visible() {
-                // Normal child.
+            } else if child.visible() || context.commands.drawing_mask() {
+                // Either a normal vivible child, or a descendant of a mask object
```
Dinnerbone, line 489: Nit: typo

Dinnerbone (approved):

core/src/display_object/movie_clip.rs
```diff
@@ -3050,6 +3050,7 @@ impl<'gc> TInteractiveObject<'gc> for MovieClip<'gc> {
             };
 
             if let Some(masker) = self.masker() {
+                // FIXME - should this really use `SKIP_INVISIBLE`? Advm2 doesn't.
```
torokati44, line 3053: Is `Advm2` a typo here?

PR #12157, opened 2023-07-19
avm2: Improve LocalConnection.send stub

Lord-McSweeney:
Progresses NewCP (#8330) after #11275.


adrian17 (approved):

SolsticeSpectrum:
@Lord-McSweeney What's the current status? All PRs that have been waiting for merge have been merged but the game still does not work. It still complains about missing `start_module.swf`. As mentioned before, a lot of clones simply don't include this file and original Flash would just skip it but Ruffle doesn't. I've also read in the Discord that Ruffle also needs something called `ExternalInterface`? I feel like we're closer than ever to run this bohemoth but how close we are actually?

Issue #11393, opened 2023-06-06
Error on https://www.disney--games.com/gameframe/338


K-REBO:
# Error Info
Error name: Error
Error message: panicked at 'wgpu error: Validation Error

Caused by:
    In Device::create_buffer
    Not enough memory left

', /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.16.1/src/backend/direct.rs:3019:5
Error stack:
```
339/U/n.wbg.__wbg_new_87297f22973157c8@https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.8d61eb08e3bb63255a98.js:1:51022
core::ops::function::FnOnce::call_once{{vtable.shim}}::h21b91a0c05b68a64@:wasm-function[8614]:0x82671e
std::panicking::rust_panic_with_hook::h85b7f6628c291e12@:wasm-function[5647]:0x74e548
core::panicking::panic_fmt::hf5c4cd929d4aaa9e@:wasm-function[8481]:0x82239f
core::ops::function::Fn::call::hae48aa1ce3e36936@:wasm-function[5965]:0x7629fc
wgpu::backend::direct::ErrorSinkRaw::handle_error::h59a1c73c3dbe9f4d@:wasm-function[4501]:0x6ed9d4
<T as wgpu::context::DynContext>::device_create_buffer::h5e4d5b9e0e95ae68@:wasm-function[6531]:0x78f725
<wgpu::Device as wgpu::util::device::DeviceExt>::create_buffer_init::hdbfc6cd743628523@:wasm-function[1431]:0x4131cd
core::cell::once::OnceCell<T>::get_or_try_init::outlined_call::ha050408a8bb7178b@:wasm-function[5003]:0x719045
ruffle_render_wgpu::surface::Surface::draw_commands_to::hac6028c1139ee180@:wasm-function[1069]:0x367ddc
<ruffle_render_wgpu::backend::WgpuRenderBackend<T> as ruffle_render::backend::RenderBackend>::submit_frame::haa2e087fc7e788ed@:wasm-function[911]:0x30bce7
ruffle_web::Ruffle::tick::ha8f939b4a8a82c36@:wasm-function[732]:0x26b831
ruffle_web::Ruffle::new_internal::{{closure}}::{{closure}}::{{closure}}::h2110eeadbc8a7319@:wasm-function[10392]:0x86a20b
[...]

PR #12167, opened 2023-07-19
render: Use hardcoded 4k limit for temp textures

adrian17:
See the comment:

```
        // NOTE: this used to compare with self.descriptors.limits.max_texture_dimension_2d
        // but in practice, some very large cached bitmaps introduce
        // slowness and extreme resource use, so for now we're using a hardcoded value.
        // This fixes slowdowns and (hopefully) crashes in Winnie the Pooh's Home Run Derby.
```

Should fix https://github.com/ruffle-rs/ruffle/issues/11393 , https://github.com/ruffle-rs/ruffle/issues/12156

EDIT: I didn't check if `max_texture_dimension_2d` can be lower than 4k, but I assumed it can't?

### DIFF
--- a/render/wgpu/src/backend.rs
+++ b/render/wgpu/src/backend.rs
@@ -991,9 +991,12 @@ impl<T: RenderTarget + 'static> RenderBackend for WgpuRenderBackend<T> {
         if width == 0 || height == 0 {
             return Err(BitmapError::InvalidSize);
         }
-        if width > self.descriptors.limits.max_texture_dimension_2d
-            || height > self.descriptors.limits.max_texture_dimension_2d
-        {
+
+        // NOTE: this used to compare with self.descriptors.limits.max_texture_dimension_2d
+        // but in practice, some very large cached bitmaps introduce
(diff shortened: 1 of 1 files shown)


Croworbit:
is this effectively a cap on quality that Ruffle will render?

adrian17:
No, it only caps two features: cacheAsBitmap and filters. The display objects above that size will still display, just without the extra effects. No other effect (blend modes) of display object is limited by this.
@Dinnerbone am I right in this assessment?

Dinnerbone:
> @Dinnerbone am I right in this assessment?

Yep!

The potential change when something stops being cacheAsBitmap is:
- Filters, if any, won't show
- Any objects that use the cacheAsBitmap as a mask, or the cacheAsBitmap using something else as a mask, will not mask correctly
- Nested blend modes may be different if they tried to blend onto "nothing" (transparent from the perspective of the cacheAsBitmap, but actually now would be the object behind the cacheAsBitmap)

Mostly only need to worry about the first one, I think.

adrian17:
superseded by https://github.com/ruffle-rs/ruffle/pull/12166 .

PR #12221, opened 2023-07-21
wstr: Don't panic on 0xFF in DecodeAvmUtf8

evilpie:
I was running into this panic:
```
thread 'main' panicked at 'attempt to shift right with overflow', wstr/src/utils.rs:139:18
```
caused by `to_string` in `byte_array.rs`.

### DIFF
--- a/wstr/src/tests.rs
+++ b/wstr/src/tests.rs
@@ -207,3 +207,9 @@ fn str_patterns() {
     test_pattern(wide, bstr!(b"aa"), &[(2, 4), (6, 8)], None);
     test_pattern(wide, wstr!('↓''a'), &[(1, 3), (5, 7)], None);
 }
+
+#[test]
+fn from_ff_byte() {
+    let s = WString::from_utf8_bytes(vec![0xff, 0xff]);
+    assert_eq!(s, wstr!(0xff 0xff));
+}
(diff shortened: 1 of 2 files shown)


moulins:
Hm, even with this fix, it seems the decoding is wrong. Hex-editing the string `"\xff\xff\xff\xff"` into some AVM1 bytecode, produces the empty string, whereas your fix would produce `"\xff\xff\xff\xff"`.

evilpie:
I was testing like this:
```
import flash.utils.ByteArray;

var bytes = [[0xff],
              [0xff, 0xff],
              [0xff, 0xff, 0xff],
              [0xff, 0xff, 0xff, 0xff],
              [0xff, 0x20],
              [0xff, 0xde],
              [0xff, 0x9e]]

for (var i = 0; i < bytes.length; i++) {
  var ba = new ByteArray();
  for (var j = 0; j < bytes[i].length; j++) {
    ba.writeByte(bytes[i][j]);
  }

  trace(bytes[i]);
  trace("length: " +ba.toString().length + " str: " + ba.toString());
}
```
Output:
```
255
length: 1 str: ÿ
255,255
length: 2 str: ÿÿ
255,255,255
length: 3 str: ÿÿÿ
255,255,255,255
length: 4 str: ÿÿÿÿ
255,32
length: 2 str: ÿ 
255,222
length: 2 str: ÿÞ
255,158
length: 2 str: ÿ
````

adrian17:
Could this be an AVM1/2 difference, then? Or a peculiarity in ByteArray specifically?
Do you get the same result with `ba.readUTF()`?

Issue #10600, opened 2023-04-07
Wentworth: Cannot start game


n0samu:
### Describe the bug

Link to SWF: http://i.notdoppler.com/files/wentworth.swf

#10192 is technically fixed, but [Wentworth](http://www.notdoppler.com/wentworth.php) still doesn't work at all. The game loads but is stuck unresponsive on this screen:
![image](https://user-images.githubusercontent.com/71368227/230659827-cfe055fe-dc36-445d-9e5b-2b35a24df551.png)

The background music that is supposed to play on this screen does not play either.

### Expected behavior

The background music should play and the buttons should be responsive to mouse hover and clicks.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

The game's logic is stored in the `xmls` object of the CompiledXML class. For example, here is the XML data that should tell the game to play the background music on the screen shown above:
```xml
  <if test="currentScene%28%29%20%21%3D%20%27loading%27">
    <if test="in%28currentScene%28%29%2C%27menu%27%2C%27intro%27%2C%27sponsor%27%2C%27splash%27%2C%27spacestation%27%2C%27wentworthfly%27%2C%27sponsor%27%29%20%7C%7C%20state%28%27lostduster%27%29%20%26%26%20currentScene%28%29%20%21%3D%20%27buckettree%27">
      <audio>WentworthStationMusic</audio>
    </if>
```

It seems like the XML data is either not loading or not having any effect, causing the screen to be unresponsive and silent.

n0samu:
@Dinnerbone found that this is occurring because we don't support `delete` on XMLList objects. Here is his minimal reproduction of the issue:
```as
var xml = <list><a>a</a><b>b</b><c>c</c></list>;
var list = xml.children();
var count = list.length();
while (count > 0) {
    trace(list[0]);
    delete list[0];
    count--;
}
```
And Aaron1011 says:
> We just need delete_property_local for XMLListObject
however, this is an XMLList derived from an XML object, which is more complicated
we need to make sure we match flash's logic for updating the parent object

n0samu:
Fixed by #12222

PR #12222, opened 2023-07-21
avm2: Implement delete_property_local for XMLList

Dinnerbone:
This progresses #10600 and allows you to start the game


core/src/avm2/object/xml_list_object.rs
```diff
@@ -438,4 +438,42 @@ impl<'gc> TObject<'gc> for XmlListObject<'gc> {
...
+                                parent.remove_child(activation.context.gc_context, &removed_node);
+                            }
+                        }
+                    }
+                    return Ok(true);
+                }
+            }
+        }
+
+        for child in write.children.iter_mut() {
```
Aaron1011, line 469: What happens if I do any any attribute name (e.g. `@*`)?

Dinnerbone: Works as you'd expect. I added that to the test to be safe.

Aaron1011 (approved):

PR #12225, opened 2023-07-21
wgpu: Towards accurate fractional sized blurs

### LINKED ISSUE #9832: Symphony: Unstable blur effect
### Describe the bug

Link to SWF: http://i.notdoppler.com/files/symphony.swf

In [Symphony](http://www.notdoppler.com/symphony.php), the entire game screen is represented as a big BitmapData object, with the blur filter applied to particles, the edge of the "battery" objects, etc. In Flash Player, the blur effect is stable; the ring around an uncharged battery stays the same width, and particles fade away at the same rate that they appear. But in Ruffle, the fading happens much more slowly. Particle colors accumulate on the screen and the ring around a battery expands until it fills the whole battery. You can see this pretty easily if you just move the force field up in the first level and wait:
image

### Expected behavior

Here's what it looks like in Flash Player:
image

The blur effect is stable and the red color is not accumulating over time.

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

This is likely the same problem as #9574.

Another thing to note is that the the force field object is not blurred like it's supposed to be. But I'm not sure if that's related.

torokati44:
The basic idea is from here: https://fgiesen.wordpress.com/2012/08/01/fast-blurs-2/

I also added a test that applies one round of horizontal blur with strengths `0:0.1:9.9`, to a bitmap having various length sequences of white pixels on black background. The reference image is an FP screenshot, and I'd say we are fairly close.
The "round_to_u8" thing (and its scalar equivalent) is specifically to approximate it better, and also to fix #9832.

The stage3d fractal test only breaks because the info text on it has a drop shadow which is based on blur.


torokati44:
Okay, I think I'm relatively happy with this now - modulo the fidgeting with the tolerances of some pre-existing visual tests. The changes in the expected outputs are visually imperceptible.
I hope I didn't get _too verbose_ with the comments.
In addition to Symphony, this also fixes the unreadable text in at least one NSFW My Little Pony game.

Dinnerbone (approved):
Incredible!

PR #12232, opened 2023-07-22
avm2: Don't change the length of a fixed output Vector

agdeneke:
This fixes Spiral Out not starting (see #12022).


Lord-McSweeney (approved):

core/src/avm2/globals/flash/geom/Matrix3D.as
```diff
@@ -198,10 +198,24 @@ package flash.geom {
...
 			var result3D:Vector3D;
-			for (var i = 0; i < Math.floor(vin.length / 3) * 3; i += 3) {
+			var resultVecs:Vector.<Number> = new Vector.<Number>();
+			for (var i = 0; i < resultVecsLength; i += 3) {
 				result3D = transformVector(new Vector3D(vin[i], vin[i + 1], vin[i + 2]));
-				vout.push(result3D.x, result3D.y, result3D.z);
+				resultVecs.push(result3D.x, result3D.y, result3D.z);
+			}
+
+			for (var i = 0; i < resultVecs.length; i++) {
```
Aaron1011, line 213: I don't think we need this loop - we can just directly use `vout[i]` or `vout.push` in the loop above.

Aaron1011 (approved):

PR #12234, opened 2023-07-22
core: Remove unsupported content warning

Dinnerbone:
We support more of AVM2 today than we did AVM1 back when it was added. I think it's time.

I thought about keeping the option there "just in case", being unused... but realistically the effort to detect and add new localized warnings feels too high.


adrian17 (approved):

danielhjacobs:
If someone has it set to false explicitly on web, will the config fail to parse when this PR merges if they update?

Aaron1011 (approved):

Issue #58, opened 2019-09-11
DisplayObject blend mode tracking issue


Herschel:
Desktop:
 - [x] Normal
 - [x] Layer
 - [x] Add
 - [x] Alpha
 - [x] Darken
 - [x] Difference
 - [x] Erase
 - [x] Hardlight
 - [x] Invert
 - [x] Lighten
 - [x] Multiply
 - [x] Overlay
 - [x] Screen
 - [x] Subtract
 - [ ] Shader (Pixel Bender)

Web canvas:
 - [x] Normal
 - [x] Layer
 - [x] Add
 - [x] Alpha
 - [x] Darken
 - [x] Difference
 - [x] Erase
 - [x] Hardlight
 - [x] Invert
 - [x] Lighten
 - [x] Multiply
 - [x] Overlay
 - [x] Screen
 - [x] Subtract
 - [ ] Shader (Pixel Bender)

ActionScript:
 - [x] `DisplayObject.blendMode`
 - [ ] `DisplayObject.blendShader`

wizzwizz4:
Some of those (Screen, I think, and some others) don't have a single definition. It'll take some snooping, and maybe reverse-engineering, to figure out which algorithm they're using.

karliss:
AS3 documentation https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/display/DisplayObject.html#blendMode has a bit more information than swf format specification.

Herschel:
All of the non-normal blend modes cause the display object to have "cache as bitmap" type behavior, where the display object is rendered into an offscreen render target first, and then rendered with the blend mode onto the stage. In fact, Layer blend mode is just Normal with the offscreen buffer behavior.

So the first step is to implement render target functionality in `RenderBackend`. It's already there for masking on web, but needs to be added to desktop.

ThisFlyingPotato:
Lacks blend effect

<details>
  <summary>Screenshots</summary>
First is how it is with Ruffle, second is how it is supposed to be.


![ruffle2](https://user-images.githubusercontent.com/69724149/116792286-fb81df00-aabf-11eb-9c7c-83fc0468a030.png)
![correct2](https://user-images.githubusercontent.com/69724149/116792310-15bbbd00-aac0-11eb-9d1f-ec5cadffba16.png)
![ruffle](https://user-images.githubusercontent.com/69724149/116792311-1b190780-aac0-11eb-94d5-912898330831.png)
[...]

PR #12238, opened 2023-07-22
avm2: Implement DisplayObject.blendShader

Aaron1011:
We already had almost all of the PixelBender support needed for this. Since this is run during displayobject rendering, we need to support MSAA, which requires us to create a `RenderPipeline` per sample count used.

This fixes Closure's rendering (previously, the game rendered as a solid white background with a few gray ovals).


Lord-McSweeney:
This also closes #58.

render/wgpu/src/pixel_bender.rs
```diff
@@ -39,6 +43,57 @@ pub struct PixelBenderWgpuShader {
...
+                            entry_point: naga_pixelbender::VERTEX_SHADER_ENTRYPOINT,
+                            buffers: &VERTEX_BUFFERS_DESCRIPTION_FILTERS,
+                        },
+                        fragment: Some(wgpu::FragmentState {
+                            module: &self.fragment_shader,
+                            entry_point: naga_pixelbender::FRAGMENT_SHADER_ENTRYPOINT,
+                            targets: &[Some(ColorTargetState {
+                                format: TextureFormat::Rgba8Unorm,
+                                // FIXME - what should this be?
+                                blend: Some(wgpu::BlendState {
```
Dinnerbone, line 72: `None`, I imagine? Unless we're expected to blend the blend result.

render/wgpu/src/pixel_bender.rs
```diff
@@ -29,7 +30,10 @@ use crate::{
 #[derive(Debug)]
 pub struct PixelBenderWgpuShader {
     bind_group_layout: wgpu::BindGroupLayout,
-    pipeline: wgpu::RenderPipeline,
+    pipeline_layout: PipelineLayout,
+    pipelines: RefCell<FnvHashMap<u32, RenderPipeline>>,
```
Dinnerbone, line 34: `SampleCountMap` exists for this

swf/src/types.rs
```diff
@@ -387,6 +387,8 @@ pub enum BlendMode {
     Erase = 12,
     Overlay = 13,
     HardLight = 14,
+    // FIXME - prevent this from being read from SWF
+    Shader = 15,
```
Dinnerbone, line 391: For cases like these, I think we should have a similar (but "enhanced") enum in ruffle_render, which has the new types that a swf can't know about. It feels like a big hack to have these exist in a serialization crate.

render/wgpu/src/pixel_bender.rs
Aaron1011: This is a pre-existing FIXME - I just moved some code around

Dinnerbone: Got it!

Dinnerbone (approved):

PR #12242, opened 2023-07-22
avm2: Throw ArgumentError when calling function with wrong arg count

Aaron1011:
Some SWFs rely on catching the error.

In order to reproduce Flash's error messages, we now store a `QName` in `FunctionObject`, which is set when we make a bound method.


tests/tests/swfs/avm2/wrong_arg_count/Test.as
```diff
@@ -0,0 +1,30 @@
...
+} catch (e) {
+	trace("Caught error: " + e);
+}
+
+
+try {
+	method("First", "Bad second", true);
+} catch (e) {
+	trace("Caught error: " + e);
+}
```
Lord-McSweeney, line 30: Can you add additional tests for calling functions with too few parameters, and also for variadic functions?

Lord-McSweeney (approved):

Molisson (approved):
Good enough, I guess.
Reviewed all of the files this time.

core/src/avm2/object.rs
```diff
@@ -209,11 +209,14 @@ pub trait TObject<'gc>: 'gc + Collect + Debug + Into<Object<'gc>> + Clone + Copy
         multiname: &Multiname<'gc>,
         activation: &mut Activation<'_, 'gc>,
     ) -> Result<Value<'gc>, Error<'gc>> {
-        match self.vtable().and_then(|vtable| vtable.get_trait(multiname)) {
-            Some(Property::Slot { slot_id }) | Some(Property::ConstSlot { slot_id }) => {
+        match self
```
adrian17, line 212: As written above, this requiring passing `ns` around feels very wrong. Both general design-wise, and perf-wise. Maybe cross-reference avmplus for this? They definitely didn't need to do this kind of thing. The approach in `fn display_method`, while probably also incorrect, managed to avoid it and pushed all work to error time.

adrian17 (changes requested):

core/src/avm2/activation.rs
```diff
@@ -393,15 +453,14 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
+        let signature: &[ParamConfig<'_>] = method.signature();
         if user_arguments.len() > signature.len() && !has_rest_or_args {
-            return Err(format!(
-                "Attempted to call {:?} with {} arguments (more than {} is prohibited)",
-                method.method_name(),
-                user_arguments.len(),
-                signature.len()
-            )
-            .into());
+            let mut dummy = Activation::from_nothing(context);
```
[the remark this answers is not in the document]
Lord-McSweeney: I don't think this needs to create a new Activation at all anyway

Aaron1011:
@adrian17 I was able to use `display_function` like you suggested, which let me get rid of lots of the extra logic that I added.

adrian17 (approved):

PR #12244, opened 2023-07-23
avm2: Clamp ByteArray.position when ByteArray.length is set

Aaron1011:
Note that ByteArray.position can manually be set out-of-range afterwards.


Aaron1011:
Fixes https://github.com/ruffle-rs/ruffle/pull/12244

Dinnerbone (approved):

tests/tests/swfs/avm2/bytearray/Test.as
```diff
@@ -105,4 +105,10 @@ trace(ba1.readBoolean());
...
 trace(ba1.readUnsignedInt());
-trace(ba1.readInt());
\ No newline at end of file
+trace(ba1.readInt());
+
+var ba2 = new ByteArray();
+ba2.position = 100;
+trace("ba2.position: " + ba2.position);
+ba2.length = 2;
+trace("ba2.position: " + ba2.length);
```
evilpie, line 114: This should be ba2.position

torokati44: Or perhaps the printed text should be `"ba2.length: "`?

Aaron1011: Good catch - I changed it to access `ba2.position`

torokati44: What's the point of tracing position again after only changing length? My impression was that the point of this last trace is to prove that length can be changed. But I don't mind either way.

Aaron1011: This PR clamps the position when `length` is changed - note that the two `ba2.position` values are different.

torokati44: Ah, alright, sorry for the trouble.

PR #12262, opened 2023-07-23
core+tests: Socket testing and improvements

sleepycatcoding:
Add support in SWF regression testing framework to mock a socket. Also some improvements to the general socket framework.

List of things left to do:

- [x] Better name for `socket_mock` utility.
- [x] Allow specifying socket mock file path in the utility.
- [x] Add tests for `Socket` now that the backend exists


torokati44:
> Better name for socket_mock utility

mocket

Dinnerbone:
> > Better name for socket_mock utility
> 
> mocket

Darnit I saw that text and immediately scrolled down to comment mocket too.

+1 for mocket.

torokati44:
Sadly we're not the first ones to make this portmanteau:
https://github.com/mindflayer/python-mocket

Dinnerbone (approved):
Awesome work once again!

I think I'd quite like some utility stuff in the socket format, for texts/floats/common data types, so we don't need to write out everything in bytes. But that's a nit and can come at any time later.

This all looks great to me, thank you!

PR #12276, opened 2023-07-24
wgpu: Fixes for blur filter quality

torokati44:
~This is based on, and currently includes, the changes from #12225.~

Turns out, the `PASS_SCALES` thing is not needed for actual filtering at all.


Dinnerbone:
If its not used in actually rendering, surely you'd see the bounds of a large blur be cut off by the max texture size which is now smaller than the kernel size?

torokati44:
Sorry, I don't understand what you mean...

The sizes of some of the rainbow squares on `bitmapdata_applyfilter_blur/expected.png`?

Dinnerbone (approved):

Issue #12078, opened 2023-07-15
Errors in Vyond's animation creator


c4artisan:
### Describe the bug

after a certain date in may-june (i don't know specifically) the vyond animation creator (or full videomaker, called that by vyond themselves) made it to the loading screen, however, 2 errors (one is related to xml, the other seems to be of a "prototype" thing?) keep it from getting past it. the creator is HEAVILY based upon xml strings (the swf being based around adobe flash builder flex stuff) (the theme list, files to list the themes, ALL of these things, except for, of course, the theme's actual .swf assets are based on xml), so it would make sense that it's not getting past the loading due to xml issues. due to it's infunctionality with no internet connection or without it's api (launching the swf directly isn't gonna work, even on flash player. it has to be on that specific website), here's an online host of the api, and the creator itself: https://instance.letsjoincdm.tk/go_full?tray=custom

and below are the 2 issues:
![image](https://github.com/ruffle-rs/ruffle/assets/41709262/28fc7548-d961-460c-9ae6-52412127ee03)

### Expected behavior

the creator will get past these errors, but perhaps encounter new errors, due to the creator being made in adobe flex/flash builder, and xml issues. my objective is to clear out these 2 errors, to progress to new ones, and have them fixed if possible

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Google Chrome 114

### Additional information

also known as the goanimate videomaker/animation creator (sorry if stuff in this issue isn't very well detailed, i tried)

Lord-McSweeney:
Most likely, the first error is causing the second one.

c4artisan:
i just realized the link died i'll be updating it soon

sleepycatcoding:
The 2nd error should be resolved by #12161.

c4artisan:
> The 2nd error should be resolved by #12161.

it was. now while there is still a second error, it's now another "can't set prototype of class to null or undefined"
![image](https://github.com/ruffle-rs/ruffle/assets/41709262/4fe0e559-0771-4854-a6ff-3fc82907f61a)

c4artisan:
as of the pull request above those errors are fixed, however, they're these now:
![image](https://github.com/ruffle-rs/ruffle/assets/41709262/9b23f63b-dfb0-4564-aea9-b5be1b54e60f)

c4artisan:
after #12330 the errors seem to be gone, but these pop up now
[...]

Issue #12054, opened 2023-07-14
BFXR - the editor does not load


Kz-92:
### Describe the bug

the program simply does not start
flash:

https://github.com/ruffle-rs/ruffle/assets/103858019/dd336546-ca25-498f-9534-23dd83e20728

Ruffle

https://github.com/ruffle-rs/ruffle/assets/103858019/90a89eed-cae9-4249-a306-52360fa4a64c

```
 ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "complete", class: flash.events::Event, ptr: 0x1d9b17673d0 }) to handler FunctionObject(FunctionObject { ptr: 0x1d9b171ec20 }) : RustError("Cannot set prototype of class to null or undefined")
 ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x1d9b1766330 }) to handler FunctionObject(FunctionObject { ptr: 0x1d9b171b3e0 }) : TypeError: Error #1034: Type Coercion failed: cannot convert "" to XMLList.
```

### Expected behavior

the program should start

### Affected platform

Desktop app

### Operating system

windows 10

### Additional information

the original site is still up, (make sure you have a flash plugin or ruffle)
https://www.bfxr.net/

or the SWF here
[Bfxr_Flash.zip](https://github.com/ruffle-rs/ruffle/files/12053009/Bfxr_Flash.zip)

evilpie:
This needs the undocumented XML::notfication and XML::setNotification functions and TLF support.

Issue #11789, opened 2023-06-29
Error Dispatching EventObject


cyberix-3d:
### Describe the bug

I'm experiencing an issue with the following link: 
https://www.gamemaker3d.com/editor/index_flash.html
When I access it, I receive the following error message:
`ERROR core/src/avm2/events.rs:419 Error dispatching event EventObject(EventObject { type: "complete", class: flash.events::Event, ptr: 0x3c405dc }) to handler FunctionObject(FunctionObject { ptr: 0x55236c }) : RustError("Cannot set prototype of class to null or undefined")`

Additionally, there are multiple errors related to accessing SWZ files from Adobe. 
These errors are occurring due to CORS restrictions:
```
Access to fetch at 'https://fpdownload.adobe.com/pub/swz/flex/4.6.0.23201/framework_4.6.0.23201.swz' from origin 'https://www.gamemaker3d.com' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
339.js:1348     GET https://fpdownload.adobe.com/pub/swz/flex/4.6.0.23201/framework_4.6.0.23201.swz net::ERR_FAILED 200

Access to fetch at 'https://fpdownload.adobe.com/pub/swz/tlf/2.0.0.232/textLayout_2.0.0.232.swz' from origin 'https://www.gamemaker3d.com' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
339.js:1348     GET https://fpdownload.adobe.com/pub/swz/tlf/2.0.0.232/textLayout_2.0.0.232.swz net::ERR_FAILED 200
```

I kindly request your assistance in resolving these issues. 
Thank you in advance.

### Expected behavior

I expected the link to load without any errors and for the event dispatching of the EventObject to be handled successfully without any issues.

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

Google chrome 114

n0samu:
Hi, it looks like you included the needed SWZ files on your site, and Ruffle is correctly falling back to those, so there is no problem there. As for the AVM2 error you mentioned, it's most likely caused by a problem that happened earlier. For example, a variable that is supposed to have a value may not have been properly initialized by Ruffle.

There are also a lot of warnings like this:
[...]

PR #12282, opened 2023-07-24
avm2+tests: Allow setting prototype to undefined

sleepycatcoding:
This is used internally by Adobe Flex in many places.

Progresses #12078, #12054, #11789


tests/tests/swfs/avm2/prototype_set_null/Test.as
```diff
@@ -0,0 +1,14 @@
...
+    import flash.display.Sprite;
+
+    public class Test extends Sprite
+    {
+    }
+}
+
+var val:Function = new Function();
+
+val.prototype = null;
```
adrian17, line 12: cool, I had no idea this was possible :)
Could you expand the test a bit more, to make sure that instantiation works and the resulting object is usable? Like
```as
function f(){}
f.prototype = null;
var x = new f();
trace(x); // [object Object]
trace(x.asdf); // undefined
```

sleepycatcoding: Can do.

adrian17:
BTW minor note for the future: if an avm2 feature PR includes a test for it, then I'd just mark it as `avm2:`, no need for `+ tests` :)

tests/tests/swfs/avm2/prototype_set_null/Test.as
sleepycatcoding:
```
2023-07-24T14:21:57.091519Z  WARN run_frame: ruffle_core::display_object::movie_clip: Got AVM2 error TypeError: Cannot convert <Unknown type> to primitive.
	at /trace()
	at global$init() [TU=Test] when attempting to assign symbol class Test
```
Interestingly it fails in Ruffle at the first `trace()` call, but works in Flash Player.

sleepycatcoding:
```
function f():void {};

f.prototype = null;

var x = new f();

trace(x.prototype); // undefined

trace(x.valueOf()); // [object Object]
trace(x.toString()); // [object Object]

trace(x); // [object Object]
trace(x.asdf); // undefined
```
It seems when `prototype` is set to undefined in Flash Player `valueOf` and `toString` methods remain, but in Ruffle they disappear?

Ruffle Output:
```
2023-07-24T14:45:23.087507Z  WARN run_frame: ruffle_core::display_object::movie_clip: Got AVM2 error TypeError: Error #1006: valueOf is not a function of class <Unknown type>.
	at global$init() [TU=Test] when attempting to assign symbol class Test
```

adrian17 (approved):

PR #9295, opened 2023-01-24
core: Propagate events to children in the reverse order

Toad06:
...when the event is not an event type used by Buttons.

Fixes #8357.

PR #12286, opened 2023-07-24
core: Push propagating `MovieClip` events in front of the action queue

mrb0nk500:
This is needed to match Flash Player's event order.

This allows Super Mario 63 to properly load custom levels from the level designer menu.

Also, this is my first time working with Rust, so if doesn't look too great, that's why.


mrb0nk500:
Also, for some reason, the test I made doesn't work with the test runner.
It almost seems like the test runner is using an older version, and not my version.

mrb0nk500:
It took way too many hours to get to this simple fix. lol

mrb0nk500:
Probably need to enable it only when using AVM1.

Lord-McSweeney:
I don't understand what you mean by that, but the test seems to be incorrectly written. The expected output differs from the output given by the test runner, so some part of the written test (not necessarily the actual bug fix for Ruffle) is probably incorrect.

Lord-McSweeney:
#9295 might intersect with this.

mrb0nk500:
Well, I know mine works with `onMouseDown`, since that's what I was trying to fix.

mrb0nk500:
I had also thought of optimizing event propagation, by caching which `MovieClip`s have propagating events, and only propagate if the parent has any children (including their children) with propagating events.

But I couldn't get it to work.

mrb0nk500:
> I don't understand what you mean by that, but the test seems to be incorrectly written. The expected output differs from the output given by the test runner, so some part of the written test (not necessarily the actual bug fix for Ruffle that you wrote) is probably incorrect.

The output was made from Flash Player, and done manually.

mrb0nk500:
I just checked, it's the coords I used in input.json

PR #12290, opened 2023-07-24
tests: Implement fscommand("quit")

Dinnerbone:
Also implements avm2 `fscommand()` function.

### DIFF
--- a/web/src/lib.rs
+++ b/web/src/lib.rs
@@ -124,6 +124,7 @@ extern "C" {
     fn is_virtual_keyboard_focused(this: &JavascriptPlayer) -> bool;
 }
 
+#[derive(Clone)]
 struct JavascriptInterface {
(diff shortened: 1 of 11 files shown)


web/src/lib.rs
```diff
@@ -1436,7 +1437,9 @@ impl ExternalInterfaceProvider for JavascriptInterface {
     fn on_callback_available(&self, name: &str) {
         self.js_player.on_callback_available(name);
     }
+}
 
+impl FsCommandProvider for JavascriptInterface {
```
Aaron1011, line 1442: We now need a `with_fs_commands` call for this, right?

Dinnerbone: Good catch, fixed!

Aaron1011 (approved):

PR #12169, opened 2023-07-19
chore: Silence warnings on latest Rust nightly regarding `wstr`'s impl of `PartialOrd`

kmeisthax:

PR #12298, opened 2023-07-25
chore: Appease nightly clippy

relrelb:


torokati44:
This is a clippy bug, see: https://github.com/ruffle-rs/ruffle/pull/12169

danielhjacobs:
Also, the one who worked on that lint suggested we should add `#[automatically_derived]` so this doesn't lint.

sleepycatcoding:
If we add #[automatically_derived] to this https://github.com/ruffle-rs/ruffle/blob/99ed2c6d4adfa8cbaace9b58e1116d2d170d32f7/wstr/src/common.rs#L497 and https://github.com/ruffle-rs/ruffle/blob/99ed2c6d4adfa8cbaace9b58e1116d2d170d32f7/wstr/src/common.rs#L624 then Clippy successfully passes.

Centri3:
Hmm that's unfortunate, I hadn't considered *other* lints also complaining...
It could be left this way or `#[allow]`s could be put wherever it's complaining, as it isn't really derived, though the sync is in a few days so it's likely best to wait for it (which will fix the original bug). You do usually want to put `automatically_derived` on macro-generated impls though.

PR #12299, opened 2023-07-25
avm1: Implement Button.filters

### LINKED ISSUE #12200: [HomestarRunner.com Main Page #24] Blur filter doesn't properly cover the AVM1Buttons
### Describe the bug

- Official pages:
  - https://homestarrunner.com/main/24 (updated site with self-hosted Ruffle player, though it's currently out of date by almost two months)
  - https://old.homestarrunner.com/main24.html (old-style SWF embeds, no included player)
- Direct link to SWF file: https://old.homestarrunner.com/main24.swf

This one has long been known as one of the few Ruffle compatibility gaps in old H*R content. Now it's been mostly fixed with the recently-added blur filters, the main joke of the page actually comes across now, but there are still a couple of issues. The main one is that the buttons on the left aren't being blurred at all, even though they would be in Flash Player:
image 1

~~Also, when trying to play this in an up-to-date web player, the blur effect is... just *way* off. The detail gets washed out far too much, and the radius of the blur effect gets clipped by what I'm guessing are the outside margins of the original objects underneath:~~ (fixed already, my browser was just using an outdated build)
image 1

### Expected behavior

The blur effect should apply to the entire displayed screen ~~and shouldn't be clipped at its edges~~. For comparison, here's what it looks like in Flash Player:
image
(Ignore the lack of top-bar buttons, those are provided by external SWFs that I can't easily load in Flash Player.)

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Mozilla Firefox 115

### Additional information

_No response_

Lord-McSweeney:
Fixes #12200.

### DIFF
--- a/core/src/avm1/globals/button.rs
+++ b/core/src/avm1/globals/button.rs
@@ -2,7 +2,9 @@
 
 use crate::avm1::activation::Activation;
 use crate::avm1::error::Error;
+use crate::avm1::globals::bitmap_filter;
 use crate::avm1::property_decl::{define_properties_on, Declaration};
+use crate::avm1::ArrayObject;
 use crate::avm1::{globals, Object, ScriptObject, TObject, Value};
 use crate::context::GcContext;
 use crate::display_object::{Avm1Button, TDisplayObject};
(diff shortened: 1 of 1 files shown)


Dinnerbone (approved):
There's probably a lot of stuff we need to copy from movieclip over :(

Lord-McSweeney:
For our current impl, only `cacheAsBitmap` needs to be added to Button.

Issue #11793, opened 2023-06-29
Doom Triple Pack: Cannot start game


n0samu:
### Describe the bug

Link to SWF: https://uploads.ungrounded.net/470000/470460_DoomGame.swf

In [Doom Triple Pack](https://www.newgrounds.com/portal/view/470460/format/flash?emulate=flash), selecting any of the 3 games goes to a blank screen. The error is:
```
ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "complete", class: flash.events::Event, ptr: 0x23cfc4f86c8 }) to handler FunctionObject(FunctionObject { ptr: 0x23cf5b50e28 }) : Error: Error #2030: End of file was encountered.
        at cmodule.doom::FSM__start/work()
        at cmodule.doom::CRunner/work()
        at cmodule.doom::CLibInit/init()
        at DoomGame/initGame()
        at DoomGame()
        at DoomContainer/startGame()
        at DoomContainer/startGameHandler()
        at flash.events::EventDispatcher/flash::events::EventDispatcher::dispatchEvent()
        at DoomMenu/gameClickedHandler()
```

### Expected behavior

The games should start.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

The second time you try to start the game, a panic will occur instead: #11792

This game was compiled with CrossBridge/Alchemy.

n0samu:
Fixed by #12308

Issue #12304, opened 2023-07-25
Renegade Racing: error - Cannot access a property or method of a null object reference. (accessing field: length)


MartySVK:
### Describe the bug

error:

``ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x1d237f52fc0 }) to handler FunctionObject(FunctionObject { ptr: 0x1d22a21e930 }) : TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: Mouse)``

SWF - https://data.gameflare.com/games/3538/MySOqH2GY5Eobf.swf
Game - https://www.gameflare.com/online-game/renegade-racing/

### Expected behavior

progress game

### Affected platform

Desktop app

### Operating system

Windows 10

Lord-McSweeney:
This is due to the following error:
```
Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0xb8bcf78 }) to handler FunctionObject(FunctionObject { ptr: 0xbcf24b0 }) : Error: Error #2030: End of file was encountered.
	at cmodule.Box2D::FSM__start/work()
	at cmodule.Box2D::CRunner/work()
	at cmodule.Box2D::CLibInit/init()
	at Box2DAS.Common::b2Base$/initialize()
	at Box2DAS.Dynamics::b2World()	at org.flixel.box2d::FlxBox2dWorld()
	at org.flixel.box2d::FlxBox2dLoadLevel/update()
```

Lord-McSweeney:
See also #11793, #10230, and #9681.

MartySVK:
ok, I will close the issue

Lord-McSweeney:
Having multiple reports of the same issue can be useful, and it will probably also be useful if later issues are encountered with the same framework.

Lord-McSweeney:
Progressed by #12308.
The main menu renders incorrectly:
![chrome-extension___donbcfbmhbcapadipfkeojnmajbakjdc_player html_url=https%3A%2F%2Fdata gameflare com%2Fgames%2F3538%2FMySOqH2GY5Eobf swf](https://github.com/ruffle-rs/ruffle/assets/84632019/6fbf3abf-74f1-480e-9099-1f9680d82ffc)
(screenshot is cropped)

Lord-McSweeney:
Also, you can't get into a level, due to this error:
```
[...]

Issue #11566, opened 2023-06-16
Tequila Zombies 2 - Can't start the level


desertking:
### Describe the bug

Hey, despite the heavy lag there are errors thrown when you get to level select menu and want to start it.

https://www.newgrounds.com/portal/view/600176?emulate=flash

ERROR core/src/avm2/events.rs:419 Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x521d37c }) to handler FunctionObject(FunctionObject { ptr: 0x3d86a1c }) : Error: Error #2030: End of file was encountered.
ruffle_web-wasm_extensions.js:716 ERROR core/src/avm2/events.rs:419 Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x68fd494 }) to handler FunctionObject(FunctionObject { ptr: 0x3d86a1c }) : TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: b2World_new)

### Expected behavior

Load & Run

### Affected platform

Online demo

### Operating system

All

### Browser

All

n0samu:
Same problem as #11426 and #9681

Lord-McSweeney:
(box2d/cmodule domain memory issue)

n0samu:
Fixed by #12308

Issue #11426, opened 2023-06-08
Controls Broken on Pretentious Game


Rno-1:
### Describe the bug

http://onemorelevel.com/static/games3/pretentious-game.swf
I click play and am unable to control the main character using keyboard controls.

The console gives these errors. The second one repeats ten more times with different pointer addresses.

`ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "render", class: flash.events::Event, ptr: 0x1867106fdd8 }) to handler FunctionObject(FunctionObject { ptr: 0x186704d7618 }) : Error: Error #2030: End of file was encountered.`

`ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "render", class: flash.events::Event, ptr: 0x18679c3e958 }) to handler FunctionObject(FunctionObject { ptr: 0x18679da5a08 }) : TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: b2World_new)`

### Expected behavior

The controls are supposed to work.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

Using Nightly 2023-06-07

n0samu:
Same problem as https://github.com/ruffle-rs/ruffle/issues/9681#issuecomment-1562620880

n0samu:
Fixed by #12308

PR #12308, opened 2023-07-25
avm2: Fix handling of default domain memory

Aaron1011:
Whenever `ApplicationDomain.domainMemory` is set to `null` (which is the default value), domain memory opcodes read from use hidden ByteArray in the domain. This hidden bytearray object is re-used if `ApplicationDomain.domainMemory` is ever set back to `null`.


sleepycatcoding:
Fixes #11793 (very very slow), #10230, #9681, #11426, #11566

Progresses #11408, #12304

core/src/avm2/domain.rs
```diff
@@ -325,12 +356,21 @@ impl<'gc> Domain<'gc> {
...
+        assert!(
+            write.domain_memory.is_none(),
+            "Already initialized domain memory!"
+        );
+        assert!(
+            write.default_domain_memory.is_none(),
+            "Already initialized domain memory!"
+        );
+
+        write.domain_memory = Some(domain_memory.as_bytearray_object().unwrap());
```
adrian17, line 372:
```
let bytearray = domain_memory.as_bytearray_object().unwrap();
write.domain_memory = Some(bytearray);
write.default_domain_memory = Some(bytearray);
```
To reduce duplication (esp of unwraps) a bit?

core/src/avm2/domain.rs
```diff
@@ -303,10 +319,25 @@ impl<'gc> Domain<'gc> {
...
         &self,
-        mc: MutationContext<'gc, '_>,
-        domain_memory: ByteArrayObject<'gc>,
-    ) {
-        self.0.write(mc).domain_memory = Some(domain_memory)
+        activation: &mut Activation<'_, 'gc>,
+        domain_memory: Option<ByteArrayObject<'gc>>,
+    ) -> Result<(), Error<'gc>> {
+        let mut write = self.0.write(activation.context.gc_context);
+        if let Some(domain_memory) = domain_memory {
```
adrian17, line 326: later `.unwrap_or_else` kinda duplicates this condition, so a bit simpler I think:
```rs
let new_memory = if let Some(domain_memory) = domain_memory {
    // check length
    domain_memory
} else {
    write.default_domain_memory.expect("Default domain memory not initialized")
}
write.domain_memory = Some(new_memory);

adrian17 (approved):

PR #12310, opened 2023-07-25
core,avm1: Dont reset fake time offset during busy loop detection

### LINKED ISSUE #3386: web: Open Doors : Slow movement
Hello,

the game movements are really slow. https://www.newgrounds.com/portal/view/429890?emulate=flash 
Please check it out. Dunno if it's a performance issue or it's a logic issue.

mrb0nk500:
This fixes #3386, but requires PR #9377 in order to do so.


mrb0nk500:
I really wanted to add a test for this, but due to the whole non deterministic nature of this, I don't think that'd be possible.

n0samu:
This would need to be tested with content such as real-time clocks or games that use the real system time, to make sure things aren't getting messed up there.

PR #12316, opened 2023-07-26
desktop: Change 'socket mode' text to 'TCP Connections'

Dinnerbone:


torokati44:
Are the string IDs good as is?

Dinnerbone:
Unless we deliberately want to invalidate every current translation. Do we?

torokati44:
Since the original language text (English) changes, they are pretty much already invalid, just not technically, no?

sleepycatcoding:
Shouldn't we change the command line flag too, no?

desktop/src/gui/open_dialog.rs
```diff
@@ -155,26 +155,26 @@ impl OpenDialog {
...
+                    .selected_text(match self.options.tcp_connections {
+                        SocketMode::Unrestricted => text(&self.locale, "tcp-connections-allow"),
+                        SocketMode::Ask => text(&self.locale, "tcp-connections-ask"),
+                        SocketMode::Deny => text(&self.locale, "tcp-connections-deny"),
                     })
                     .show_ui(ui, |ui| {
                         ui.selectable_value(
-                            &mut self.options.socket_mode,
+                            &mut self.options.tcp_connections,
                             SocketMode::Unrestricted,
```
sleepycatcoding, line 168: SocketMode::Unrestricted should be renamed to SocketMode::Allow, if we want full parity with command line flag and desktop GUI.

Dinnerbone: Good catch!

desktop/src/gui/open_dialog.rs
```diff
@@ -155,26 +155,26 @@ impl OpenDialog {
...
+                        SocketMode::Ask => text(&self.locale, "tcp-connections-ask"),
+                        SocketMode::Deny => text(&self.locale, "tcp-connections-deny"),
                     })
                     .show_ui(ui, |ui| {
                         ui.selectable_value(
-                            &mut self.options.socket_mode,
-                            SocketMode::Unrestricted,
+                            &mut self.options.tcp_connections,
+                            SocketMode::Allow,
                             text(&self.locale, "socket-mode-unrestricted"),
```
sleepycatcoding, line 169: Shouldn't this be tcp-connections-allow?

desktop/src/gui/open_dialog.rs
```diff
@@ -155,26 +155,26 @@ impl OpenDialog {
...
                         );
                         ui.selectable_value(
-                            &mut self.options.socket_mode,
+                            &mut self.options.tcp_connections,
                             SocketMode::Ask,
                             text(&self.locale, "socket-mode-ask"),
```
sleepycatcoding, line 174: Here too.

desktop/src/gui/open_dialog.rs
```diff
@@ -155,26 +155,26 @@ impl OpenDialog {
...
                         );
                         ui.selectable_value(
-                            &mut self.options.socket_mode,
+                            &mut self.options.tcp_connections,
                             SocketMode::Deny,
                             text(&self.locale, "socket-mode-deny"),
```
sleepycatcoding, line 179: This too.

Dinnerbone: Fixed all three, thank you!

sleepycatcoding (approved):

PR #12324, opened 2023-07-26
tests: Add known_failure as a test option, and migrate many tests to it

Dinnerbone:
The goal is to use `known_failure = true` instead of `ignore = true` for tests that we know don't work now, but may work one day.

When a test is marked `known_failure`, it'll be tested and the comparison logic is inverted. If it suddenly passes, it'll throw an error and the developer will be prompted to unflag it as the test now passes and should be checked going forwards for regressions.


torokati44:
Hmm, this name implies to me that a given test is a "negative test" and that the _expected result_ (ie. when Ruffle is working correctly) is that it fails. Sure, the message it gives explains that this is not the case, but in the test itself it's not there. How about "known_fails" (or something similar)? I like the basic idea though!

Dinnerbone:
`known_failure` might be good?

I just took the name Adrian said and rolled with it, didn't give much thought. Happy to change it!

torokati44:
> `known_failure` might be good?

Excellent IMHO!

Dinnerbone:
Renamed!

sleepycatcoding (approved):

adrian17 (approved):

PR #1506, opened 2020-11-06
Fix rendering order of DisplayObjects and drawings

relrelb:
Previously `DisplayObject`s were rendered before any drawings, which means that drawings were displayed in front of all `DisplayObject`s.
While I didn't find any evidence for that, it seems like the order should be the opposite, as observed in some games:
* [Shopping Cart Hero](https://chat.kongregate.com/gamez/0003/6396/live/ShoppingCartHero.swf)
* [Free Rider 2](http://www.onemorelevel.com/games3/freerider2x.swf)

As a drive by, I corrected some error messages in `MovieClip` to name the actually called method.

**EDIT**: By the time of fixing the issue, someone came with an identical fix in #1503, so this PR is only necessary for correcting the error messages.

PR #12348, opened 2023-07-27
tests: Port avmplus ByteArray test suite

Dinnerbone:
All tests pass with the `avm2:` commits

Also adds `lzma` feature to tests, so we can test lzma (de)compression


Lord-McSweeney (approved):

core/src/avm2/activation.rs
```diff
@@ -3025,7 +3027,8 @@ impl<'a, 'gc> Activation<'a, 'gc> {
 
         let address =
             usize::try_from(address).map_err(|_| "RangeError: The specified range is invalid")?;
```
Lord-McSweeney, line 3029: This should be `RangeError: Error #1506: The specified range is invalid.` (for negative addresses)

core/src/avm2/activation.rs
```diff
@@ -3059,7 +3063,8 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
```
Lord-McSweeney, line 3065: See above above comment

core/src/avm2/activation.rs
Dinnerbone: I haven't touched those errors in this PR, can we do that in a followup?

Aaron1011 (approved):

PR #12357, opened 2023-07-27
avm2: Special case Number.toString for Infinity and NaN

Lord-McSweeney:
This avoids getting into an infinite loop for some of the avmplus tests.

### DIFF
--- a/core/src/avm2/globals/number.rs
+++ b/core/src/avm2/globals/number.rs
@@ -260,6 +260,18 @@ pub fn print_with_radix<'gc>(
         return Value::from(number).coerce_to_string(activation);
     }
 
+    if number.is_nan() {
+        return Ok("NaN".into());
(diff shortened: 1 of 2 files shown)


core/src/avm2/globals/number.rs
```diff
@@ -260,6 +260,24 @@ pub fn print_with_radix<'gc>(
         return Value::from(number).coerce_to_string(activation);
     }
 
+    if number.is_nan() {
+        return Ok(AvmString::new_utf8(activation.context.gc_context, "NaN"));
```
Dinnerbone, line 264: I may be wrong but doesn't this allocate a new string every time, as opposed to `"NaN".into()` (or `AvmString::from("NaN")` which would not?

Dinnerbone (approved):

PR #12371, opened 2023-07-28
tests: Port many misc avmplus ecma3 tests

Dinnerbone:
Also limits Array's length to u16 in places, so we don't try to allocate 100GB.


core/src/avm2/globals/array.rs
```diff
@@ -88,6 +88,16 @@ pub fn instance_init<'gc>(
                     )?));
                 }
 
+                // [NA] temporarily limit this to u16. It may not be correct but it's better than 100GB arrays.
+                // TODO: sparse array support
+                if expected_len > u16::MAX as f64 {
+                    return Err(Error::AvmError(error(
```
Aaron1011, line 94: Could you make this a `RustError` instead? If an swf catches and ignores this error, it'll make debugging much harder.

Aaron1011 (approved):

Issue #12383, opened 2023-07-28
Spurious failures of socket tests with writeObject


adrian17:
Example 1:
https://github.com/ruffle-rs/ruffle/actions/runs/5694961606/job/15437188344
```
---- avm2/socket_write_little ----
test panicked: Received data did not match expected data
Expected: [1, 67, 255, 10, 20, 65, 246, 40, 92, 143, 194, 154, 191, 64, 205, 76, 153, 66, 174, 78, 58, 132, 72, 101, 108, 108, 111, 33, 10, 11, 1, 1, 3, 0, 0, 9, 255, 255, 158, 136, 8, 199, 115, 238, 25, 0, 72, 101, 108, 108, 111, 32, 102, 114, 111, 109, 32, 82, 117, 102, 102, 108, 101, 32, 83, 111, 99, 107, 101, 116, 33, 82, 97, 119, 32, 85, 84, 70, 32, 105, 115, 32, 99, 111, 111, 108]
Actual: [1, 67, 255, 10, 20, 65, 246, 40, 92, 143, 194, 154, 191, 64, 205, 76, 153, 66, 174, 78, 58, 132, 72, 101, 108, 108, 111, 33, 10, 11, 1, 1, 16, 0, 0, 0, 0, 9, 255, 255, 158, 136, 8, 199, 115, 238, 25, 0, 72, 101, 108, 108, 111, 32, 102, 114, 111, 109, 32, 82, 117, 102, 102, 108, 101, 32, 83, 111, 99, 107, 101, 116, 33, 82, 97, 119, 32, 85, 84, 70, 32, 105, 115, 32, 99, 111, 111, 108]
```

Example 2, ran locally:
```
---- avm2/socket_write_big ----
test panicked: Received data did not match expected data
Expected: [1, 67, 255, 10, 20, 65, 64, 191, 154, 194, 143, 92, 40, 246, 66, 153, 76, 205, 132, 58, 78, 174, 72, 101, 108, 108, 111, 33, 10, 11, 1, 1, 3, 0, 0, 9, 255, 255, 136, 158, 238, 115, 199, 8, 0, 25, 72, 101, 108, 108, 111, 32, 102, 114, 111, 109, 32, 82, 117, 102, 102, 108, 101, 32, 83, 111, 99, 107, 101, 116, 33, 82, 97, 119, 32, 85, 84, 70, 32, 105, 115, 32, 99, 111, 111, 108]
[...]

PR #12396, opened 2023-07-29
core,tests: Fix socket test backend

sleepycatcoding:
This was hiding the fact that the test technically did not run and failed, as it only happened very rarely in some scenarios.

Partially fixes the problem in #12383.


sleepycatcoding:
I redesigned the socket system to use async-channel when sending data out. This allows the removal of timeouts in test backend.

sleepycatcoding:
@Dinnerbone Now that the socket tests actually work, there is still something wrong AMF0 writeObject, so maybe we can split the Socket writeObject into a separate test and use `known_failure`?

Dinnerbone:
> maybe we can split the Socket writeObject into a separate test and use `known_failure`?

Will it always fail? If not it'll be flaky just with opposite chances. If it will always fail though, this sounds good!

sleepycatcoding:
This will always fail. The reason it was so unpredictable before was that the timeouts were too high, so basically the test did not run. It only ran on MacOS runner correctly is because the runner seems to be fairly slow.

sleepycatcoding:
There should probably be a flag to make test runner execute async tasks to completion after running all ticks as that would have shown that the socket write test did not run correctly.

sleepycatcoding:
There should probably be a flag to make test runner execute async tasks to completion after running all ticks as that would have shown that the socket write test did not run correctly.

Dinnerbone (approved):

SN902:
#12466

PR #12399, opened 2023-07-29
avm2: Allow several Vector tests to pass

Lord-McSweeney:
This also makes the first ("Vector fixed RangeError") and last ("Vector fixed RangeError checkReadIndex_d") tests in `from_avmplus/as3/Vector/fixed_length` pass.


core/src/avm2/activation.rs
```diff
@@ -1955,15 +1955,28 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
 
-        if args.len() > 1 {
-            return Err(format!(
-                "VerifyError: Cannot specialize classes with more than one parameter, {} given",
-                args.len()
-            )
-            .into());
+        let applied = base.apply(self, args[0])?;
+
+        if args.len() != 1 {
```
adrian17, line 1960: as mentioned on Discord, let's try moving this into `apply()`.

core/src/avm2/globals/vector.rs
```diff
@@ -260,12 +260,15 @@ pub fn concat<'gc>(
...
-            return Err(format!(
-                "TypeError: Cannot coerce argument of type {:?} to argument of type {:?}",
-                arg_class.read().name(),
-                my_base_vector_class.inner_class_definition().read().name()
-            )
-            .into());
+            return Err(Error::AvmError(type_error(
+                activation,
+                &format!(
+                    "Error #1034: Type Coercion failed: cannot convert {:?}@00000000000 to {:?}.",
```
adrian17, line 266: (long term we'll probably want to extract this into a function (not necessarily now), this error message is already duplicated...)
Oh BTW, the other implementation uses `instance_of_class_name()`, and `to_qualified_name_err_message()` are you sure that one isn't better?

Lord-McSweeney: Yeah, those are more correct. I didn't extract it into a `make_error_1034` function- if it ends up being needed in more places, we should definitely do that.

core/src/avm2/object/class_object.rs
```diff
@@ -911,17 +912,35 @@ impl<'gc> TObject<'gc> for ClassObject<'gc> {
...
         let self_class = self.inner_class_definition();
 
         if !self_class.read().is_generic() {
-            return Err(format!("Class {:?} is not generic", self_class.read().name()).into());
+            return Err(make_error_1127(activation));
+        }
+
+        if nullable_params.len() != 1 {
+            let class_name = self
+                .inner_class_definition()
```
adrian17, line 925: not necessarily in this PR, but maybe we should refactor these into `self.name()`

adrian17 (approved):
LGTM (with one thing to do)
Note: this should be squashed before merge (either before, or via github button)

PR #12432, opened 2023-07-30
Run npm update --save

tsunamistate:
Running `npm install` after https://github.com/ruffle-rs/ruffle/pull/12302 desyncs `package-lock.json` with reverts of some packages to previous versions
It seems that Renovate updates `package-lock.json` without updating `package.json`, and npm install reverts package versions to ones in `package.json`

I ran `npm update --save` like https://github.com/ruffle-rs/ruffle/pull/11842 to sync the files


torokati44:
Thank you!
I've amended the commit message a little bit (added the "chore" tag) - I hope you don't mind!

PR #12436, opened 2023-07-30
avm2: Fix property_is_enumerable on arrays

adrian17:


sleepycatcoding (approved):

torokati44:
Just searching for `>= index` in the codebase, I see a few really similar lines in `get_enumerant_name`, and also not only in `array_object.rs`, but in `dictionary_object.rs`, `vector_object.rs`, and `xml_list_object.rs` as well.
Should those be updated as well?

adrian17:
Will check these out tomorrow, thanks for pointing it out.

adrian17:
@torokati44 turns out the implementation for Vector shouldn't exist at all :D 

The remaining ones are in `get_enumerant_name/value` which I assume are correct.

torokati44:
> turns out the implementation for Vector shouldn't exist at all :D

Well, now we know! :D

> The remaining ones are in get_enumerant_name/value which I assume are correct.

Why the assumption? To me, any comparison that handles the `length == index` (which is usually over-indexing) the same way as `length > index` (not over-indexing), as opposed to as `length < index` (obviously over-indexing). Unless the thing is indexed starting from 1 for some reason.

adrian17:
It does a `>=`  comparison, but then subtracts 1 from the index.
I didn't try to understand exactly how enumeration works, but I checked that replacing the comparison by `>` breaks array enumeration so I'm assuming everything currently works as it should :D

torokati44:
> but I checked that replacing the comparison by > breaks array enumeration so I'm assuming everything currently works as it should :D

Okay, that's good enough for me! :D

Dinnerbone (approved):

Aaron1011 (approved):

Issue #2240, opened 2021-01-03
BitmapData.merge, BitmapData.perlinNoise, BitmapData.copyPixels  - not yet implemented


Diatonator:
Creating the issue similar to https://github.com/ruffle-rs/ruffle/issues/2073

copyPixels is present both in this and https://github.com/ruffle-rs/ruffle/issues/2073, but perlinNoise and merge are unique to this.

The link:

https://z0r.de/282

![image](https://user-images.githubusercontent.com/75184268/103491416-dabcbc00-4e34-11eb-9c25-73577f23c6e8.png)

The visual consequence of this not implemented is the absence of this sort of liquid motion:

Adobe:
![282_Flash](https://user-images.githubusercontent.com/75184268/103491537-aa295200-4e35-11eb-982c-3f685cd70e58.gif)

Ruffle:
![282_Ruffle](https://user-images.githubusercontent.com/75184268/103491539-adbcd900-4e35-11eb-90ae-d348660e616c.gif)

torokati44:
This loop also uses `BitmapData.draw()`, and will not work any better without it.
Unfortunately it passes a `MovieClip` to it as parameter (as opposed to a `BitmapData`), so a hacky temporary solution that simply copyes pixels around won't do either; this will need the actual full off-screen rendering.

torokati44:
Upon closer inspection, this loop also needs the `filters` property, and `DisplacementMapFilter`.

Lord-McSweeney:
Everything has been implemented for this SWF, but for some reason, the DisplacementMapFilter is static in Ruffle, while the filter on the text seems to change in FP.

torokati44:
See also: #12149

torokati44:
The initial post cropped the screenshots to their bottom parts.
The loop is square in Flash Player too, and has that top text:

![image](https://github.com/ruffle-rs/ruffle/assets/288816/f3215347-f9bd-40bf-9bb3-7e02ac7360fa)

firoball:
Oh, that's interesting :D
I thought inbetween some ruffle versions I noticed the exact window layout as initially shown, just without effects.

Nevermind then, sorry for the ping.

PR #2832, opened 2021-01-25
Implement BitmapData.perlinNoise()

torokati44:
This aims to help with #2240, although that is still blocked by `BitmapData.draw` (also by `merge`, but that doesn't seem to be enough in itself).
Also should improve #2295, but without `BitmapData.paletteMap`, it actually regresses a bit - at least visually.

I tested these changes with lots of different parameter combinations, using [this ActionScript code](https://gist.github.com/torokati44/c1b5a859d7c96bb4022ac36ed493b2f4), which I compiled with `mtasc` into [this .swf](https://github.com/ruffle-rs/ruffle/files/5865428/BitmapPerlinNoise.zip).

This produces the following image:

image

**EDIT:**
I managed to "fix" the rounding when converting the floating point values to `u8` in all cases.
Now, when taking the example `.swf` above, and making four variations of it: Setting the `transparency` property of the `BitmapData`; and the `fractalNoise` parameter of `perlinNoise()` to be always `true`/`false` (independent of each other); every single pixel matches perfectly between Ruffle and Flash Player in all four cases.

As usual, there are some ways performance could be improved, but for now, this is more than nothing.

PR #12439, opened 2023-07-30
core: Don't divide by zero in BitmapData::perlinNoise() if baseX or baseY is 0.0

torokati44:
This progresses https://z0r.de/282 .
Re: https://github.com/ruffle-rs/ruffle/issues/2240#issuecomment-1656176519


torokati44:
Do we have a visual test with the SWF from https://github.com/ruffle-rs/ruffle/pull/2832#issue-793199611 ? I can't seem to find it right now.

adrian17 (approved):

torokati44:
I have added the visual test from https://github.com/ruffle-rs/ruffle/pull/2832, without a .fla file (I have no Flash access at the moment), compiled with `mtasc`.
One thing I changed is that I set the `baseX` parameter in one of the calls to `0` to check for this fix as well.

Dinnerbone (approved):

Issue #11432, opened 2023-06-08
Stuck on Dragon Fist 3D's Second Loading Screen


Rno-1:
### Describe the bug

http://onemorelevel.com/static/games3/dragon-fist-3d.swf
The first loading screen works fine, but when you click "Play Game," it gets stuck on the second. Spoofing the URL to https://www.kongregate.com and https://benoldinggames.co.uk didn't work either.
The fist error message repeats about 50 times.

WARN ruffle_core::tag_utils: Unknown tag code: 255
WARN ruffle_core::tag_utils: Unknown tag code: 255
WARN ruffle_core::stub: Encountered stub: AVM2 flash.system.Security.allowDomain()
WARN ruffle_core::stub: Encountered stub: AVM2 flash.system.Security.allowInsecureDomain()
ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x1ec81d744b8 }) to handler FunctionObject(FunctionObject { ptr: 0x1ec8108c488 }) : TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: scaleX)
WARN ruffle_core::stub: Encountered stub: AVM2 flash.utils.Dictionary constructor with weak keys
ERROR ruffle_core::loader: Error during movie loading: FetchError("HTTP status is not ok, got 404 Not Found") 
ERROR ruffle_core::timer: Unhandled AVM2 error in timer callback: ReferenceError: Error #1069: Property prependScale not found on flash.geom.Matrix3D and there is no default value.

### Expected behavior

The second loading screen should finish loading and bring up the main menu.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

Running Nightly 2023-06-08

Lord-McSweeney:
The game still gets stuck on the second loading screen with the following error:
```
Unhandled AVM2 error in timer callback: TypeError: Error #2007: Parameter alphas must be non-null.
	at flash.display::Graphics/flash::display::Graphics::beginGradientFill()
	at flare.core::Light3D/setColors()
	at flare.core::Light3D/setParams()
	at flare.core::Light3D()
	at flare.basic::Scene3D()
	at ResourcesLoader()
	at DragonFist4/loadResources()
```

PR #12457, opened 2023-07-31
avm2: Implement `Matrix3D.prependScale`

agdeneke:
This implements the Matrix3D.prependScale() method required by Dragon Fist 3D in #11432.


Dinnerbone (approved):
Thank you!

Issue #11289, opened 2023-05-30
Neopets Loader Version 9 seems to present XML problem


sombraguerrero:
[np9_loader_v12.zip](https://github.com/ruffle-rs/ruffle/files/11597633/np9_loader_v12.zip)
### Describe the bug

Games in Neopets that utilize version 9 of the nploader SWF seem to present Ruffle with an XML loading problem.

ERROR core/src/avm2/events.rs:419 Error dispatching event EventObject(EventObject { type: "complete", class: flash.events::Event, ptr: 0xced544 }) to handler FunctionObject(FunctionObject { ptr: 0xce28f4 }) : TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: childNodes)

This version mostly applies to those games that were last using ActionScript 3 such as:
Destruct-O-Match III -- https://www.neopets.com/games/game.phtml?game_id=999&size=regular&quality=high&play=true
Jungle Raiders -- https://www.neopets.com/games/game.phtml?game_id=1064&size=regular&quality=high&play=true
Wheeler's Wild Ride -- https://www.neopets.com/games/game.phtml?game_id=1080&size=regular&quality=high&play=true

# Player Info
Allows script access: true
Renderer: wgpu
Adapter Backend: Gl
Adapter Name: "ANGLE (NVIDIA, NVIDIA GeForce RTX 3070 Ti Direct3D11 vs_5_0 ps_5_0, D3D11)"
Adapter Device Type: Other
Adapter Driver Name: ""
Adapter Driver Info: ""
Enabled features: Features(TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES | PUSH_CONSTANTS)
Available features: Features(TEXTURE_COMPRESSION_BC | CLEAR_TEXTURE | MULTIVIEW)
[...]

PR #12465, opened 2023-07-31
avm2: Implement XML settings & correct the XMLDocument implementation

Dinnerbone:
This should significantly progress (if not outright solve, I can't test yet) #11289


core/src/avm2/globals/flash/xml/XMLNode.as
```diff
@@ -1,112 +1,193 @@
...
-      }
-
-      public function get nextSibling(): XMLNode {
-         return this._parent.childNodes[this._index + 1];
-      }
-
-      public function get nodeName(): String {
-         if (_isDocument) {
+
+import __ruffle__.stub_getter;
```
Lord-McSweeney, line 4: nit: fmt (indent everything from here to end of file)

core/src/avm2/globals/flash/xml/XMLNode.as
```diff
@@ -1,112 +1,193 @@
...
+        }
+
+        node.previousSibling = before.previousSibling;
+        before.previousSibling = node;
+        node.nextSibling = before;
+        node.parentNode = this;
+    }
+
+    public function appendChild(node: XMLNode): void {
+        if(node.parentNode === this) {
```
evilpie, line 87: mega nit: space

core/src/avm2/globals/flash/xml/XMLNode.as
```diff
@@ -1,112 +1,193 @@
...
+        return null;
+    }
+
+    public function toString(): String {
+        if (nodeType != XMLNodeType.ELEMENT_NODE) {
+            return _escapeXML(nodeValue);
+        }
+
+        var result = "";
+        if (nodeName != null)
```
evilpie, line 149: nit: inconsistent braces (and below)

evilpie:
Very cool. LGTM

core/src/avm2/globals/flash/xml/XMLNode.as
Dinnerbone: Indented it all.

core/src/avm2/globals/flash/xml/XMLNode.as
Dinnerbone: 😱

core/src/avm2/globals/flash/xml/XMLDocument.as
```diff
@@ -1,21 +1,84 @@
...
+      }
+
+      private function _convertXmlNodeType(kind: String): uint {
+         if (kind == "text") {
+            return XMLNodeType.TEXT_NODE;
+         }
+         if (kind == "comment") {
+            return XMLNodeType.COMMENT_NODE;
+         }
+         if (kind == "processing-instruction") {
```
evilpie, line 75: I don't think this is possible. We don't allow processing-instructions during parsing. 

In general: I would  suggest we just throw for what I think is the impossible situation of an unexpected kind here. (Would also eliminate some null checks above)

Dinnerbone: We don't exactly have a way to panic or throw RustErrors during AS, and throwing a new type of AS error is weird and unprecedented.

evilpie: I don't think it's unreasonable to do `throw new Error("unexpected state")` for something that we believe is impossible. This seems a lot easier to debug later than the number of children being different or something like that.

Aaron1011: Could you add in a stub, so that there's some indication if this ever gets hit?

Dinnerbone: Stub will incorrectly mark this as a stub, which it isn't. I'll add an error.

PR #12466, opened 2023-07-31
avm2: Implement XMLSocket

sleepycatcoding:
Implements XMLSocket for AVM2.

I am not experienced with AS3, so comments and nitpicks are very welcome.

~~This is a draft, because #12396 is not merged yet.~~


core/src/avm2/globals/flash/net/XMLSocket.as
```diff
@@ -1,23 +1,85 @@
...
         }
 
-        public function close(): void {
-            stub_method("flash.net.XMLSocket", "close");
+        override public function connect(host:String, port:int):void
+        {
+            if (host == null)
+            {
+                // FIXME: We should take the movie "host" and connect to it here.
+                stub_method("flash.net.XMLSocket", "connect() with null host");
```
Lord-McSweeney, line 36: I think the implementation of `LocalConnection.domain` looks up the movie domain- you should be able to reuse the same logic here.

sleepycatcoding:
I don't know of any content to test this with. Testing is appreciated.

core/src/avm2/globals/flash/net/XMLSocket.as
```diff
@@ -1,23 +1,84 @@
...
+{
+    import flash.net.Socket;
+    import flash.events.ProgressEvent;
+    import flash.events.DataEvent;
+    import flash.utils.ByteArray;
+
+    // NOTE: Extending from Socket is a implementation detail. It is not possible for SWFs to
+    // use Socket methods as Flash's playerglobal doesn't include those methods, unless
+    // SWF author intentionally tried, which wouldn't make much sense.
+    public class XMLSocket extends Socket
```
Dinnerbone, line 11: We try to keep things as close to Flash as possible, as this is detectable (and therefor potentially bug-inducing) via as3.

I'd suggest keeping a `private var socket: Socket` instead, and wrapping that.

Dinnerbone (changes requested):

core/src/avm2/globals/flash/net/XMLSocket.as
sleepycatcoding: Done.

Dinnerbone (approved):
Thank you!

core/src/avm2/globals/flash/net/xml_socket.rs
```diff
@@ -0,0 +1,29 @@
...
+        if url.scheme() == "file" {
+            "localhost".into()
+        } else if let Some(domain) = url.domain() {
+            AvmString::new_utf8(activation.context.gc_context, domain)
+        } else {
+            // no domain?
+            "localhost".into()
+        }
+    } else {
+        tracing::error!("LocalConnection::domain: Unable to parse movie URL");
```
Lord-McSweeney, line 24: This should be `XMLSocket::connect` instead of `LocalConnect::domain`

sleepycatcoding: I forgot to change it when I copied it from LocalConnection 🤦‍♂️

tests/tests/swfs/avm2/xml_socket/Test.as
```diff
@@ -0,0 +1,67 @@
...
+    }
+}
+
+import flash.net.XMLSocket;
+import flash.errors.IOError;
+import flash.events.Event;
+import flash.events.DataEvent;
+
+// NOTE: prettyPrinting affects XMLSocket, so turn it off so we can pass.
+XML.prettyPrinting = false;
```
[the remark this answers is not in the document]
sleepycatcoding: I had to disable it, because Ruffle doesn't prettyPrint XML (yet).

adrian17 (approved):

PR #12467, opened 2023-07-31
debug_ui: Indent non-expandable items object- and display trees

torokati44:
This confused me terribly.

Before (single item, `instance2`, collapsed/expanded):

image image

After (same item collapsed/expanded):

image image

### DIFF
--- a/core/src/debug_ui/display_object.rs
+++ b/core/src/debug_ui/display_object.rs
@@ -583,7 +583,18 @@ impl DisplayObjectWindow {
                     }
                 });
         } else {
-            open_display_object_button(ui, context, messages, object, &mut self.hovered_debug_rect);
+            // This item is not expandable, but we want to keep
+            // the space empty where the expand button would be,
+            // so it doesn't look like a sibling of the parent.
+            ui.indent(ui.id().with(object.as_ptr()), |ui| {
+                open_display_object_button(
+                    ui,
+                    context,
+                    messages,
+                    object,
+                    &mut self.hovered_debug_rect,
+                );
+            });
         }
     }
 }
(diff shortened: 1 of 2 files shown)


torokati44:
Yes, this does mean that basically every level of the tree hierarchy is effectively two indentation levels.
One is to show the child relationship (which was already there), and one is to keep the alignment the same between collapsible and non-collapsible children.

I would prefer it the topmost level was omitted (with the expand buttons scooted to the left a bit), but saw no way to do that.

Dinnerbone (approved):
I have no strong feelings about this. Sure thing.

torokati44:
> I have no strong feelings about this. 

Huhh, that's so strange to hear. :D The second screenshot, when looked at in itself, at least to me, clearly means that `"water_mc"` has 4 children, 1 `MovieClip` (with a funky expand triangle, granted) and 3 `Graphic`s.

> Sure thing.

Anyway, I'm glad you don't oppose it either! ^^

Dinnerbone:
> Huhh, that's so strange to hear. :D The second screenshot, when looked at in itself, at least to me, clearly means that `"water_mc"` has 4 children, 1 `MovieClip` (with a funky expand triangle, granted) and 3 `Graphic`s.

Somewhat disagree. The line does that, to me 😅 Now there's clearly an extra indent that isn't associated with anything.

torokati44:
> The line does that, to me :sweat_smile:

Huh, you mean _this_ line? :D
image

I always consider whatever is there just meaningless decoration clutter. :sweat_smile:

torokati44:
Thank you for the merge! ^^

Issue #8534, opened 2022-11-13
Pipe Riders not supported AVM2


deividbinev12:
### Describe the bug

https://user-images.githubusercontent.com/80902993/201509664-fd7cb477-9173-4991-a92d-f37414c10a3b.mp4

Ruffle Extension 2022-11-04

https://www.y8.com/games/pipe_riders

### Expected behavior

https://user-images.githubusercontent.com/80902993/201509783-49f7e651-d142-4e64-88ff-a2edd4a17d50.mp4

Flash Version

https://flashstorage.games/pipe-riders.html

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge

### Additional information

it's 3D flash game but, AVM2 is not supported 3D, WAFlash is error but, pipe riders like friv

Lord-McSweeney:
I get an out-of-memory error in the console for me (Chrome 107): `WAFLASH> WAFLASH_NOTIFY_ABORT. Memory usage exceeded threshold!

Ar @ waflash.min.js?2022091801:1
`

n0samu:
> I get an out-of-memory error in the console (unrelated to Ruffe) for me (Chrome 107): `WAFLASH> WAFLASH_NOTIFY_ABORT. Memory usage exceeded threshold!`
> 
> `Ar @ waflash.min.js?2022091801:1`

That's because you're playing the game with WAFlash, not Ruffle. flashstorage.games has a bar below the game that lets you choose what emulator to use.
![image](https://user-images.githubusercontent.com/71368227/206959187-88b28b98-2948-4171-bc2c-00284f061294.png)

Lord-McSweeney:
The original issue is no longer present, but you cannot play levels, probably due to the following error: `Modifying an XML object is not yet implemented`.

Lord-McSweeney:
New error:
```
Error dispatching event EventObject(EventObject { type: "closed", class: flash.events::Event, ptr: 0x7c6250 }) to handler FunctionObject(FunctionObject { ptr: 0x2706b50 }) : TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: unlink)
	at sandy.core.data::Polygon/set appearance()
	at sandy.core.data::Polygon/destroy()
	at sandy.core.scenegraph::Shape3D/__destroyPolygons()
	at sandy.core.scenegraph::Shape3D/destroy()
	at PipeLevel/CreateCylinderClay()
	at PipeLevel/CreateScene()
[...]

Issue #12458, opened 2023-07-31
3 Pandas in Fantasy: Coercion failed: cannot convert Object@00000000000 to Array


MartySVK:
### Describe the bug

The game is stuck at the end of the first level

SWF - https://data.gameflare.com/games/6700/Corve3JZ8rUgtr.swf
Game - https://www.gameflare.com/online-game/3-pandas-in-fantasy/

Same error in the test for Vector - https://github.com/ruffle-rs/ruffle/issues/12321

### Expected behavior

progress the game

### Affected platform

Desktop app

### Operating system

Windows 10

Lord-McSweeney:
Note that this error also appears, but it's unrelated (stubbing it doesn't help): `ReferenceError: Error #1069: Property uncaughtErrorEvents not found on flash.display.Loader and there is no default value.`

Lord-McSweeney:
Should be fixed by #12470

PR #12470, opened 2023-07-31
avm2: Use RawTable to implement 'public index' iteration

Aaron1011:
This makes our implementation more closely aligned with avmplus. In particular, it's now possible to delete keys from an object while iterating without disturbing the iteration order (as long as those keys were already produced by the iterator).

This is based on @Bale001's work on RawTable-based iteration

A few tests had their output changed (they depend on the exact object iteration order, and don't neccessarily match Flash Player exactly).

Fixes https://github.com/ruffle-rs/ruffle/issues/8534, https://github.com/ruffle-rs/ruffle/issues/12458


core/src/avm2/dynamic_map.rs
```diff
@@ -0,0 +1,215 @@
...
+
+/// A HashMap designed for dynamic properties on an object.
+#[derive(Debug, Collect, Clone)]
+#[collect(no_drop)]
+pub struct DynamicMap<K: Eq + PartialEq + Hash, V> {
+    values: hashbrown::HashMap<K, DynamicProperty<V>, FnvBuildHasher>,
+    // The last index that was given back to flash
+    public_index: usize,
+    // The actual index that represents where an item is in the HashMap
+    real_index: usize,
```
moulins, line 30: It would be better if these were `Cell<usize>`s, to avoid requiring `&mut self` (and thus extra GC write barriers) in  `next/pair_at`.

core/src/avm2/dynamic_map.rs
```diff
@@ -0,0 +1,215 @@
...
+        None
+    }
+
+    pub fn pair_at(&mut self, index: usize) -> Option<&(K, DynamicProperty<V>)> {
+        // NOTE: AVM2 object enumeration is one of the weakest parts of an
+        // otherwise well-designed VM. Notably, because of the way they
+        // implemented `hasnext` and `hasnext2`, all enumerants start from ONE.
+        // Hence why we have to `checked_sub` here in case some miscompiled
+        // code doesn't check for the zero index, which is actually a failure
+        // sentinel.
```
moulins, line 139: This talks about a `checked_sub` operation that doesn't appear anywhere; is the comment outdated?

Aaron1011: I just deleted the whole comment, since we're now more closely based on avm2

Dinnerbone (approved):
LGTM!

torokati44:
Let's see if manually triggering a JS package update unblocks the checks...

Dinnerbone:
This PR needs a `cargo update`

torokati44:
Yeah I got confused by npm erroring out, but internally it's a cargo build failure.

torokati44:
Fixed the `hashbrown` version in `Cargo.lock`. Also snuck in the Node.js updates before, even though they are unrelated...

Lord-McSweeney:
Looks like more stuff needs to be adjusted to account for the changed iteration order of keys.

PR #12482, opened 2023-08-01
tests: Port avmplus as3 AMF test

Dinnerbone:
It currently fails.

> RustError("Error: Invalid object") when attempting to assign symbol class Test

First failure: 
> Array Index:Mixed Value:Mixed value check EXP vs ACT: 2 FAILED! expected: [object UserDefClassAllTypes] got: undefined


Lord-McSweeney:
(failure is because custom object serialization is currently unimplemented)

PR #12488, opened 2023-08-01
render: Calculate correct bounds for shapes containing curves

Dinnerbone:
Not likely to have any huge impact, but it'll tighten the bounds of anything glyphs with curves as we generate them this way


torokati44 (approved):
The order of parameters of `quadratic_curve_bounds` is a bit weird to me (the 3 `Point` params are not right after one another), but oh well, it doesn't really matter.

PR #5160, opened 2021-08-28
preleminary work on avm1 XMLSocket

hussein-aitlahcen:
This PR introduce basic avm1 XMLSocket operations; new, connect, onConnect, onData, close, send
The implementation is supporting Desktop only for now as I don't know how to handle it in web (perhaps websockets as @EmperorBale mentioned [here](https://github.com/ruffle-rs/ruffle/discussions/5134) ?).

Disclaimer; I am not familiar with the codebase nor I am familier with how Flash Player is working under the hood. I pretty much hacked ruffle based on this documentation http://demo.ligams.free.fr/AS2LR/XMLSocket.html to check whether or not we could run Dofus Retro with ruffle.
It ended up 'working' (requiring a bit more work, not part of  this PR, as there was some issues with SWF imports).

Related to #905 #288

PR #12489, opened 2023-08-01
avm1: Implement XMLSocket + Subsystem fixes

### LINKED ISSUE #288: Actionscript Tracking: XMLSocket
This is the tracking issue for **XMLSocket** in AVM1. We will check off each item of progress as appropriate, and fill in any relevant or missing information as we continue development of Ruffle.
# Legend
Each checkbox is independent of another. It's entirely possible for something to be tested but not exist yet, or for us to believe that it's completely functional but we haven't made enough tests to prove it.
## "Property Exists"
This means the item exists, but may not necessarily be fully implemented.
## "Has Test Coverage"
This means that we believe that we have a good test coverage of this item, regardless of if those tests pass. It's okay to have tests available but not implement the item yet.
## "Completely Functional"
This means we believe that the item is completely implemented, and no more work needs to be done towards making it functional.

# XMLSocket
### `setTimeout()`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional

### `XMLSocket()`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional

### `timeout : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional

### `close()`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `connect(url, port)`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `send(data)`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `onClose()`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `onConnect(success)`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `onData(src)`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `onXML(src)`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional

sleepycatcoding:
This PR implements XMLSocket for AVM1.

Supersedes #8188, #5160

Closes #288

~~This is a draft, because I haven't added tests and haven't tested this yet.~~

- [x] Add tests


core/src/socket.rs
```diff
@@ -14,15 +22,22 @@ use std::{
 
 pub type SocketHandle = Index;
 
+#[derive(Copy, Clone, Collect)]
+#[collect(no_drop)]
+enum SocketKind<'gc> {
+    Avm2(SocketObject<'gc>),
+    Avm1(Object<'gc>),
```
Lord-McSweeney, line 29: It might be nice to import `Object` as `Avm1Object`, to avoid confusion in other parts of this file.

sleepycatcoding:
Ready for review, I added some tests and they pass, but this has not been tested on any real content.

Dinnerbone (approved):
Looks good to me! Thank you!

core/src/socket.rs
```diff
@@ -111,100 +160,218 @@ impl<'gc> Sockets<'gc> {
...
+
+                                let message = AvmString::new_utf8_bytes(activation.gc(), &message);
+
+                                let _ = target.call_method(
+                                    "onData".into(),
+                                    &[message.into()],
+                                    &mut activation,
+                                    ExecutionReason::Special,
+                                );
+                            }
```
Minigugus, line 336: If more than 1 null byte arrives in the same message, you'll miss it until the next message from the server. In my change request I didn't had this issue because I was polling the socket back-end, I did it that way to have at most 1 message fired per update, but in your case this code runs only once per chunk of data received so you risk to miss some messages in some racy conditions.

```suggestion
                            let mut buffer = xml_socket.read_buffer();
                            buffer.extend(data);

                            // Check for a message.
                            while let Some((index, _)) =
                                buffer.iter().enumerate().find(|(_, &b)| b == 0)
                            {
                                let message = buffer.drain(..index).collect::<Vec<_>>();
                                // Remove null byte.
                                let _ = buffer.drain(..1);

                                let message = AvmString::new_utf8_bytes(activation.gc(), &message);

                                let _ = target.call_method(
                                    "onData".into(),
                                    &[message.into()],
                                    &mut activation,
                                    ExecutionReason::Special,
                                );
                            }
```

Minigugus (commented):
Well done, you clearly did better than me! :grin:

core/src/socket.rs
```diff
@@ -111,100 +160,218 @@ impl<'gc> Sockets<'gc> {
...
                         // Socket must have been closed before we could send event.
                         None => continue,
                     };
 
-                    let connect_evt =
-                        EventObject::bare_default_event(&mut activation.context, "connect");
-                    Avm2::dispatch_event(&mut activation.context, connect_evt, target.into());
+                    match target {
+                        SocketKind::Avm2(target) => {
+                            let mut activation = Avm2Activation::from_nothing(context.reborrow());
```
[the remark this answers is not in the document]
sleepycatcoding: It is used in multiple places in Ruffle for constructing events and firing them. For example: https://github.com/ruffle-rs/ruffle/blob/5f200dc2a9fd992575c4bb5c49973f203cccf43a/core/src/loader.rs#L1317 and https://github.com/ruffle-rs/ruffle/blob/5f200dc2a9fd992575c4bb5c49973f203cccf43a/core/src/loader.rs#L1714

core/src/socket.rs
```diff
@@ -111,100 +160,218 @@ impl<'gc> Sockets<'gc> {
...
-        if let Some(Socket { sender, .. }) = self.sockets.remove(handle) {
+        if let Some(Socket { sender, target }) = self.sockets.remove(handle) {
             drop(sender); // NOTE: By dropping the sender, the reading task will close automatically.
+
+            // Clear the read buffer if the connection was closed.
+            if let SocketKind::Avm1(target) = target {
+                let target = XmlSocket::cast(target.into()).expect("target should be XmlSocket");
+
+                target.read_buffer().clear();
+            }
```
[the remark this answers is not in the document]
sleepycatcoding: That is a oversight. Good catch!

core/src/socket.rs
sleepycatcoding: Good catch!

core/src/socket.rs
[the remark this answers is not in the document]
Minigugus: Does that mean the documentation is wrong, or theses samples should be fixed too? (Or I'm miss interpreting the documentation :sweat_smile:)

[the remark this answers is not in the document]
sleepycatcoding: To be honest, I don't know.

core/src/socket.rs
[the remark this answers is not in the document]
sleepycatcoding: Fixed in latest commit!

sleepycatcoding:
I found a bug myself where connected was returning `true` after server closed the connection, and it was because the socket was not removed from the Arena. I added a test case for this.

core/src/socket.rs
sleepycatcoding: Fixed in the latest commit.

core/src/socket.rs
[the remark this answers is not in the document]
sleepycatcoding: The same bug also existed in AVM2 XMLSocket.

core/src/avm1/globals/xml_socket.rs
```diff
@@ -0,0 +1,267 @@
...
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    if let Some(xml_socket) = XmlSocket::cast(this.into()) {
+        let timeout = args
+            .get(0)
+            .unwrap_or(&Value::Undefined)
+            .coerce_to_u32(activation)?;
+
+        // FIXME: Check if flash player clamps this to 250 like AS3 sockets.
+        xml_socket.0.timeout.set(timeout);
```
Minigugus, line 88: Consistency: there's a `socket.set_timeout()` in Avm2 code and a `timeout()` on `XmlSocket` just above, maybe a `xml_socket.set_timeout()` makes sens? By the way the clamping could be added in this function in the future if needed :slightly_smiling_face: 

```suggestion
        xml_socket.set_timeout(timeout);
```

core/src/socket.rs
```diff
@@ -111,100 +160,229 @@ impl<'gc> Sockets<'gc> {
...
+                SocketKind::Avm1(target) => {
+                    let target =
+                        XmlSocket::cast(target.into()).expect("target should be XmlSocket");
+
+                    target.read_buffer().clear();
+                }
+                SocketKind::Avm2(target) => {
+                    target.read_buffer().clear();
+                    target.write_buffer().clear();
+                }
```
Minigugus, line 177: It first I thought `target.write_buffer().clear()` was missing, but it's actually on purpose as the write buffer isn't used by the Avm1 XML socket. Is it worth trying to move this buffer to the Avm2-only code? Or to add a comment to explain the purpose of the writer buffer? I don't know if it's how much Ruffle team cares about that

Minigugus (approved):
Nice, thanks!

LGTM even with theses 2 comments :slightly_smiling_face:

core/src/socket.rs
sleepycatcoding: The reason that Avm2 Socket has a write buffer is because generally data is not sent out until the flush() method is called and at that point we empty out the buffer and send it out. But with XMLSocket we send the data out immediately, so there is no need to buffer the data.

Minigugus:
If found a panic while trying this PR:

```
thread 'main' panicked at 'already borrowed: BorrowMutError', core/src/avm1/globals/xml_socket.rs:40:28
stack backtrace:
   0: rust_begin_unwind
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/panicking.rs:593:5core/src/avm1/globals/xml_socket.rs:40:28
   1: core::panicking::panic_fmt
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/panicking.rs:67:14
   2: core::result::unwrap_failed
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/result.rs:1651:5
   3: ruffle_core::socket::Sockets::close
   4: ruffle_core::avm1::globals::xml_socket::close
   5: ruffle_core::avm1::function::Executable::exec
   6: ruffle_core::avm1::object::TObject::call_method
   7: <ruffle_core::avm1::object::Object as ruffle_core::avm1::object::TObject>::call_method
```

The problematic code is

https://github.com/sleepycatcoding/ruffle/blob/0864967b0d613dfbcaedf26a2694c6275b343484/core/src/avm1/globals/xml_socket.rs#L39-L41

I guess the swf I tried used to close the socket in the onData or onXML event, which can be quite frequent when using sockets. You should probably drop mutable references before running Avm code. I haven't looked at the code but Avm2 might be affected too :confused:

sleepycatcoding:
I think I understand where the issue is. Going to fix it later.

sleepycatcoding:
> I guess the swf I tried used to close the socket in the onData or onXML event, which can be quite frequent when using sockets. You should probably drop mutable references before running Avm code. I haven't looked at the code but Avm2 might be affected too :confused: 

AVM2 should not be affected, as the Socket API there only dispatches one socketData event with how many bytes were received. The event handler then must read the bytes manually with readByte, readBytes and etc

sleepycatcoding:
> If found a panic while trying this PR:

Can you try again with latest commit. It should be fixed.

Minigugus:
No panic this time, seams fixed, thanks :+1:

core/src/socket.rs
```diff
@@ -111,100 +160,238 @@ impl<'gc> Sockets<'gc> {
...
+                            drop(buffer);
+
+                            for message in messages {
+                                let _ = target.call_method(
+                                    "onData".into(),
+                                    &[message.into()],
+                                    &mut activation,
+                                    ExecutionReason::Special,
+                                );
+                            }
```
Minigugus, line 352: I was think of a better way to process messages without using 2 loops and an extra Vec, what about this code? I think it is more efficient as it only consider null bytes in `data`, in order to have a complexity of `O(data.len())` instead of `O(read_buffer.len())`

```suggestion
                            let mut data = data;

                            if let Some((index, _)) = data.iter().enumerate().find(|(_, &b)| b == 0)
                            {
                                // There's no null byte in the read buffer so we can be sure
                                // all remaining bytes will be consumed by the next message.
                                let mut buffer = xml_socket
                                    .read_buffer()
                                    .drain(..)
                                    .chain(data.drain(..index))
                                    .collect::<Vec<_>>();

                                loop {
                                    // Remove null byte.
                                    let _ = data.drain(..1);

                                    let message =
                                        AvmString::new_utf8_bytes(activation.gc(), &buffer);

                                    let _ = target.call_method(
                                        "onData".into(),
                                        &[message.into()],
                                        &mut activation,
                                        ExecutionReason::Special,
                                    );

                                    if let Some((index, _)) =
                                        data.iter().enumerate().find(|(_, &b)| b == 0)
                                    {
                                        // We already consumed everything in the buffer so arriving here
                                        // means the next message is fully in data only, no need to access
                                        // the buffer again then.
                                        buffer = data.drain(..index).collect::<Vec<_>>();
                                    } else {
                                        // No full message left in data, we'll wait for the next chunk
                                        // from the server.
                                        break;
                                    }
                                }
                            };

                            // Append what remain to XmlSocket read buffer.
                            // Since we consumed all null byes above, we can be sure
                            // the read buffer will never contain one.
                            if !data.is_empty() {
                                xml_socket.read_buffer().extend(data);
                            }

```

Obviously it's just a suggestion, your code works too :+1:

Minigugus, line 352: If the socket gets closed when multiple messages are still pending in `messages`, we'll still call `onData` even if user's code asked to close the socket.

```suggestion
                            for message in messages {
                                let _ = target.call_method(
                                    "onData".into(),
                                    &[message.into()],
                                    &mut activation,
                                    ExecutionReason::Special,
                                );
                                
                                // Drop remaining messages if the socket was closed.
                                if !context.sockets.sockets.contains(handle) {
                                    break;
                                }
                            }
```

Minigugus (commented):
I would have been happy to contribute myself but I don't have time yet :disappointed: Anyway you did an amazing job, thanks again! :slightly_smiling_face:

core/src/socket.rs
sleepycatcoding: Not sure what Flash's behavior here is at this point, but I think we can leave it like this for now and change it in the future if it becomes a problem.

sleepycatcoding: It does hinder code readability, but I think the  current way is performant enough.

sleepycatcoding: I thought about this, for a bit and implemented your suggestion in [8037eff](https://github.com/ruffle-rs/ruffle/pull/12489/commits/8037eff6e3755ed88dcfa9e164b29765cfbc6046)

Dinnerbone (approved):
Thanks for adding tests too!

PR #12496, opened 2023-08-02
avm2: Don't make descendants operator (..) call a method

Dinnerbone:
It's more correct, and should be faster (don't need all the function overhead)


Lord-McSweeney:
This will probably also be useful when implementing the `Proxy` `getDescendants` function.

Aaron1011 (approved):

PR #12499, opened 2023-08-02
avm2: Correct a bunch of errors

Dinnerbone:


core/src/avm2/globals/number.rs
```diff
@@ -320,7 +321,11 @@ fn to_string<'gc>(
                 .coerce_to_u32(activation)? as usize;
 
             if radix < 2 || radix > 36 {
-                return Err("toString can only print in bases 2 thru 36.".into());
+                return Err(Error::AvmError(range_error(
+                    activation,
+                    "Error #1003: The radix argument must be between 2 and 36; got 1.",
```
Lord-McSweeney, line 326: Same here

core/src/avm2/globals/function.rs
```diff
@@ -15,8 +16,16 @@ use gc_arena::GcCell;
 pub fn instance_init<'gc>(
     activation: &mut Activation<'_, 'gc>,
     this: Object<'gc>,
-    _args: &[Value<'gc>],
+    args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
+    if !args.is_empty() {
```
Lord-McSweeney, line 21: Function also has a call handler that just forwards to `new Function(...)` (`Function(...)` is equivalent to `new Function(...)`)- could you add that in this PR? It might fix some tests.

Lord-McSweeney, line 21: Function also has a call handler that just forwards to `new Function(...)` (`Function(...)` is equivalent to `new Function(...)`)- could you add that in this PR?

core/src/avm2/globals/number.rs
Dinnerbone: Fixed them

core/src/avm2/globals/function.rs
Dinnerbone: Added

Lord-McSweeney (approved):

Aaron1011 (approved):

PR #12514, opened 2023-08-02
tests: Port avmplus MOPS tests

Dinnerbone:
This was a pain to port, but let's hope nobody ever has to recompile it!

All passing, after some error cleanup!


Croworbit:
with all the tests from other projects, how can you be sure they are accurate to flash?
out of curiosity
will these help lead you there, or is this for compatibility between emulators somehow?

Dinnerbone:
They're ran in Flash Player and the output of Flash Player is used to compare against Ruffle.

For these tests specifically, they're built by Adobe and used to test the official AVM.

Lord-McSweeney (approved):

PR #12522, opened 2023-08-03
renovate: Remove separate lockFileMaintenance schedule, set rangeStrategy to bump.

torokati44:
This is supposed to fix the issue mentioned in https://github.com/ruffle-rs/ruffle/pull/12474#issuecomment-1660590171.

That is caused by the separate lockFileMaintenance PRs fast-forwarding the lockfiles in themselves, but not the version spec in package.json. So the lockfile will be rolled back locally. Let's try to make the lockfiles be bumped together with the version specs for both Rust and JS, with a different solution than https://github.com/ruffle-rs/ruffle/pull/9884.

The previous goal of grouping only patch updates together, and having minor and major updates be their own separate PRs per package (83f88a6f3e6757993d09b6bc97303739f86d4c30) also didn't work, as can be seen in https://github.com/ruffle-rs/ruffle/pull/12474.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -7,11 +7,14 @@
             "matchLanguages": ["rust"],
             "matchUpdateTypes": "patch",
             "groupName": "Rust dependency patches",
+            "rangeStrategy": "bump",
             "extends": ["schedule:weekly"],
         },
         {
             "matchLanguages": ["rust"],
             "groupName": "Rust dependencies",
+            "matchUpdateTypes": ["major", "minor"],
+            "rangeStrategy": "bump",
             "extends": ["schedule:weekly"],
         },
         {
@@ -26,6 +29,7 @@
         {
             "matchLanguages": ["js"],
             "groupName": "Node.js dependencies",
+            "rangeStrategy": "bump",
             "extends": ["schedule:monthly"],
         },
         {
@@ -34,7 +38,4 @@
             "extends": ["schedule:monthly"],
         },
     ],
-    "lockFileMaintenance": {
-        "enabled": true
-    }
 }


torokati44:
It would be great if someone who actually knows how all of this is supposed to work (unlike me), could verify that this will indeed do what's intended. O.O

Dinnerbone (approved):
I have no idea so let's just try it!

torokati44:
> I have no idea so let's just try it!

Now that's the spirit! :wink: But in all seriousness, we can always revert/adjust again later.
Since the dependency versions themselves will be bumped every week, obviously the lock files can't fall behind, as that would be invalid. Whether this will also update the transitive (indirect) dependencies in the lock files, we'll see. If not, maybe we'll have to re-add `lockFileMaintenance` schedule, but at least it won't cause the weird local rollback as it does now.

BTW I asked GPT4 to explain the old and new config files and how they are different. Since it's _(as we all know)_ an infallible, omniscient oracle, we shall obey it. Here is that it had to say:

----

Let's break down the configurations for RenovateBot one by one:

### **First Configuration**

1. **Base Configuration**:
   - This configuration extends from `config:base`, meaning it uses the base configuration settings of Renovate as a starting point.

2. **Rust Patches**:
   - This rule matches Rust language dependencies.
   - It only considers `patch` updates.
   - It groups these updates under the name "Rust dependency patches".
   - The pull requests for these updates will be created on a weekly schedule.

3. **Rust Dependencies**:
   - This rule matches Rust language dependencies (all update types since no specific `matchUpdateTypes` is set).
   - It groups these updates under the name "Rust dependencies".
   - The pull requests for these updates will also be created on a weekly schedule.

4. **wasm-bindgen Group**:
   - This rule matches specific packages: `wasm-bindgen`, `js-sys`, `web-sys`, and `wasm-bindgen-futures`.
   - These package updates are grouped under the name "wasm-bindgen".

5. **Disabling `tracing-tracy`**:
   - This rule matches the package named `tracing-tracy`.
   - Updates for this package are disabled, meaning no PRs will be created for updates to this package.

6. **Node.js Dependencies**:
   - This rule matches JavaScript language dependencies.
   - It groups these updates under the name "Node.js dependencies".
   - The pull requests for these updates will be created on a monthly schedule.

7. **GitHub Actions Dependencies**:
   - This rule matches dependencies that are managed by GitHub Actions.
   - These updates are grouped under the name "GitHub Actions dependencies".
   - The pull requests for these updates will also be created on a monthly schedule.

8. **Lock File Maintenance**:
   - This configuration will enable maintenance of lock files. When enabled, Renovate will regenerate the lock file(s) once a week (by default). This helps to ensure that the lock file remains up-to-date with the `package.json` file and with the latest package versions on the npm registry. It creates a separate PR specifically for this purpose.

### **Second Configuration**

This configuration is very similar to the first, but with some distinctions:

1. **Rust Patches**:
   - Similar to the first configuration, but with an added `rangeStrategy` set to `bump`. This means for patch updates, it will modify the existing range to include the patch (instead of just adding a new version).

2. **Rust Dependencies**:
   - This rule matches Rust language dependencies but only for `major` and `minor` updates.
   - Similar grouping and scheduling to the first configuration.
   - Again, a `rangeStrategy` of `bump` is set for these updates.

3. **Node.js Dependencies**:
   - Similar to the first configuration, but with a `rangeStrategy` of `bump`.

The other rules (for wasm-bindgen, `tracing-tracy`, and GitHub Actions) remain the same as in the first configuration.

### **Differences**

1. The second configuration specifies a `rangeStrategy` of `bump` for Rust patches, Rust dependencies (major & minor), and Node.js dependencies. This means that when these dependencies are updated, Renovate will modify the existing version range to include the new versions.

2. In the second configuration, Rust dependencies specifically mention major and minor updates, while the first configuration will include all updates (since no specific `matchUpdateTypes` is set).

3. The first configuration has the `lockFileMaintenance` setting enabled, ensuring that lock files are kept updated. This is absent in the second configuration.

In general, the impact of these configurations on PRs would be:
- Grouping related packages together for combined update PRs.
- Specific scheduling for when these PRs are opened.
- Disabling updates for specific packages.
- Maintenance of lock files in the first configuration.
- More precise control over version bumps in the second configuration using `rangeStrategy`.

PR #12524, opened 2023-08-03
avm2: Implement a few XML/XMLlist methods

Dinnerbone:
Nothing too exciting, just:

- `processingInstructions(name)` + make processingInstructions actually parse/stringify correctly
- `comments()`
- `valueOf()`
- `toJSON()`


core/src/avm2/e4x.rs
```diff
@@ -711,8 +742,13 @@ fn to_xml_string_inner<'gc>(xml: E4XOrXml<'gc>, buf: &mut WString) -> Result<(),
             buf.push_str(&escape_element_value(*text));
```
Lord-McSweeney, line 742: `to_xml_string_inner` will never return an Error now, so I think its return type can be removed.
This also means `to_xml_string` (previously returned an `Result<AvmString, Error>`) can now return an `AmString<'gc>`.

core/src/avm2/globals/XMLList.as
```diff
@@ -21,11 +21,24 @@ package {
...
         AS3 native function toString():String;
+        AS3 native function comments():XMLList;
+        AS3 native function processingInstructions(name:String = "*"):XMLList;
 
         // The following native methods are not declared in the documentation,
         // but still exist
         AS3 native function name(): Object;
 
+        AS3 function toJSON(k:String) : *
+        {
```
Lord-McSweeney, line 32: nit: The rest of the file puts the bracket on the same line as the function definition, like `AS3 function toJSON(k:String):* {`

core/src/avm2/e4x.rs
Dinnerbone: Good catch!

core/src/avm2/globals/XMLList.as
Dinnerbone: Fixed

core/src/avm2/e4x.rs
```diff
@@ -774,7 +808,7 @@ pub fn to_xml_string<'gc>(
     activation: &mut Activation<'_, 'gc>,
 ) -> Result<AvmString<'gc>, Error<'gc>> {
```
Lord-McSweeney, line 809: This can return an `AvmString<'gc>` now since it will always return `Ok`

Dinnerbone: Fixed (and a few other methods)

Lord-McSweeney (approved):

adrian17 (approved):

Issue #10493, opened 2023-04-01
Achievement Unlocked 2: Broken Achievements


waspennator:
### Describe the bug

https://www.newgrounds.com/portal/view/553791?emulate=flash
According to a review left on the page and testing on my end, these achievements are borked, copy pasting.

"129: Not Left Click (doesn't recognize when you right click)
189: Rollout! (doesn't recognize scrolling, move your mouse off the screen)
193: Won't Follow Directions (option doesn't show up when right clicked)
204-213: Two Browser Windows/The Final Straw (doesn't realize when you open it in two windows, making that whole chunk impossible since it relies on the first one to open a special thing)
233: I Don't Get It (walkthrough link doesn't work)"

There's also 2 achievements tied to the right click context menu that don't show up on ruffle (And are unobtainable currently on ruffle desktop until #10426 gets merged in)

204-213 also doesn't work on desktop even opening up two instances of ruffle.

### Expected behavior

All of the achievements listed should be obtainable, desktop should be able to open the context menu and earn the 2 achievements tied to it as well and opening up 2 ruffle instances should have the 2nd ruffle instance load up to this area.

![Achievement Unlocked 2 Hot Coffee](https://user-images.githubusercontent.com/78768499/229259928-0b20b492-65e7-4788-9565-97eb6d8662b5.png)

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 110

n0samu:
See also: #10287

waspennator:
Achievement 233 is working now (clicking on the walkthrough link button), but the other achievements are still borked.

waspennator:
According to discussion on discord, apparently this game uses LocalConnection for the hot coffee easter egg (communicates between 2 of itself. LocalConnection has been recently implemented, but it can't communicate with another instance of itself yet.

waspennator:
Still a problem

PR #12526, opened 2023-08-03
Implement context menu events for AVM2

Lord-McSweeney:
This allows you to complete Stage 7 in This Is The Only Level 3.

Unfortunately doesn't help with #10493.

### DIFF
--- a/core/src/avm2/globals/flash/ui/ContextMenuBuiltInItems.as
+++ b/core/src/avm2/globals/flash/ui/ContextMenuBuiltInItems.as
@@ -1,13 +1,97 @@
 package flash.ui
 {
-	public final class ContextMenuBuiltInItems {
-		public var forwardAndBack:Boolean = true;
-		public var loop:Boolean = true;
-		public var play:Boolean = true;
-		public var print:Boolean = true;
-		public var quality:Boolean = true;
-		public var rewind:Boolean = true;
-		public var save:Boolean = true;
-		public var zoom:Boolean = true;
-	}
+    public final class ContextMenuBuiltInItems {
+        private var _forwardAndBack:Boolean = true;
+        private var _loop:Boolean = true;
+        private var _play:Boolean = true;
+        private var _print:Boolean = true;
+        private var _quality:Boolean = true;
+        private var _rewind:Boolean = true;
+        private var _save:Boolean = true;
+        private var _zoom:Boolean = true;
+        
+        public function get forwardAndBack():Boolean {
+            return this._forwardAndBack;
+        }
+        
+        public function set forwardAndBack(value:Boolean):void {
+            this._forwardAndBack = value;
+        }
+        
+        
+        public function get loop():Boolean {
+            return this._loop;
+        }
+        
+        public function set loop(value:Boolean):void {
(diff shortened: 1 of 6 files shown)


core/src/avm2/globals/flash/ui/ContextMenuBuiltInItems.as
```diff
@@ -1,13 +1,97 @@
...
+        private var _save:Boolean = true;
+        private var _zoom:Boolean = true;
+        
+        public function get forwardAndBack():Boolean {
+            return this._forwardAndBack;
+        }
+        
+        public function set forwardAndBack(value:Boolean):void {
+            this._forwardAndBack = value;
+        }
```
Dinnerbone, line 19: Why did you convert from public fields to trivial getter/setters? Here and elsewhere

Lord-McSweeney: This matches FP. I know it's kinda weird, but there's no harm in doing it, and this way, `describeType` output will match as well.

Dinnerbone (approved):

core/src/avm2/globals/flash/ui/ContextMenuBuiltInItems.as
```diff
@@ -1,13 +1,97 @@
...
-		public var loop:Boolean = true;
-		public var play:Boolean = true;
-		public var print:Boolean = true;
-		public var quality:Boolean = true;
-		public var rewind:Boolean = true;
-		public var save:Boolean = true;
-		public var zoom:Boolean = true;
-	}
+    public final class ContextMenuBuiltInItems {
+        private var _forwardAndBack:Boolean = true;
```
adrian17, line 4: a bit late but: was there a reason to convert these?
I'm tempted to convert back ;D

core/src/player.rs
```diff
@@ -679,8 +698,40 @@ impl Player {
...
+                            let mut activation = Avm2Activation::from_nothing(context.reborrow());
+
+                            let menu_obj = root_dobj
+                                .as_interactive()
+                                .map(|iobj| iobj.context_menu())
+                                .and_then(|v| v.as_object());
+
+                            if menu_obj.is_some() {
+                                // TODO: contextMenuOwner and mouseTarget might not be the same (see above comment)
+                                let menu_evt = activation
```
adrian17, line 716: all of this should have been extracted to its own method, just like `run_context_menu_custom_callback` for AVM1

desktop/src/gui.rs
```diff
@@ -410,7 +410,9 @@ impl RuffleGui {
                     let clicked = if item.checked {
                         Checkbox::new(&mut true, &item.caption).ui(ui).clicked()
```
adrian17, line 411: this doesn't appear to support checked items?

core/src/avm2/globals/flash/ui/ContextMenuBuiltInItems.as
Lord-McSweeney: These are technically correct to flash (see the above review by Dinnerbone). I don't know why Flash does it this way.

PR #12539, opened 2023-08-03
core/desktop/web: Play unsupported videos with browser

danielhjacobs:
~~Draft since once a video is opened it cannot be closed or dismissed with current code.~~

[Screencast from 2023-08-04 11-02-41.webm](https://github.com/ruffle-rs/ruffle/assets/29206584/44a22311-0c91-4e8f-a28f-e708892afb1f)

### DIFF
--- a/core/src/backend/ui.rs
+++ b/core/src/backend/ui.rs
@@ -4,6 +4,7 @@ use fluent_templates::loader::langid;
 pub use fluent_templates::LanguageIdentifier;
 use std::borrow::Cow;
 use std::collections::HashSet;
+use url::Url;
 
 pub type FullscreenError = Cow<'static, str>;
 pub static US_ENGLISH: LanguageIdentifier = langid!("en-US");
@@ -36,6 +37,8 @@ pub trait UiBackend: Downcast {
     fn open_virtual_keyboard(&self);
 
     fn language(&self) -> &LanguageIdentifier;
+
+    fn display_unsupported_video(&self, url: Url);
 }
 impl_downcast!(UiBackend);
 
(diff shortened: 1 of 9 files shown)


danielhjacobs:
This is no longer a draft. However, note that an ideal version of this PR would add all videos in their proper positions with their proper dimensions/bounds and would give them an id corresponding to something unique (such as their handles), which would then allow them to be programmatically removed when their corresponding video element is destroyed. The video would not appear in a modal which takes focus away from the main content, but instead be overlaid on the content in its proper position.

This is meant as a stopgap until someone designs that ideal solution.

n0samu:
Testing [Axon](http://i.notdoppler.com/files/axon.swf) in the Ruffle extension's internal player page, I see some issues:
image
`Content-Security-Policy: The page’s settings blocked the loading of a resource at http://s3-eu-west-1.amazonaws.com/wellcome-collection/axon/video/sciencevid.mp4 (“default-src”).`
Also, each page of the "How to Play" section is supposed to show its own video, but the video element only pops up once. I'm not sure if that's strictly related to this PR though.

danielhjacobs:
I think the extension page may need `media-src: *`, bit I don't know if the extension stores will like that. I wonder if it would be possible to check for that error while the video loads, and if it's found, open it in a new tab if that's allowed.

danielhjacobs:
https://github.com/ruffle-rs/ruffle/blob/b98b0e88aafc2ddee4ff9c68217c2bd5d47f81cd/web/packages/extension/manifest.json5#L32

danielhjacobs:
The first "How to Play" video plays with that change. The others are interesting. They are fetched by the network tab, but for whatever reason the code doesn't get to "Unrecognized file signature" and therefore my workaround never fires.

danielhjacobs:
>  Also, each page of the "How to Play" section is supposed to show its own video, but the video element only pops up once. I'm not sure if that's strictly related to this PR though.

Should be fixed by latest commit.

Dinnerbone (approved):
LGTM!

kmeisthax (approved):
Seems fine to me

Issue #10639, opened 2023-04-11
Jet Velocity: Cannot start game


n0samu:
### Describe the bug

Link to SWF: https://uploads.ungrounded.net/553000/553350_jet_velocity_viral.swf

After the loading screen, [Jet Velocity](https://www.newgrounds.com/portal/view/553350/format/flash?emulate=flash) shows nothing but a black screen.

In the current version of Ruffle, the game is blocked by XML.hasComplexContent being unimplemented. With #10638, the error no longer occurs, but the game still doesn't load.

### Expected behavior

The game should load.

Note: when testing in the Flash projector, you'll need to load the game via the provided SWF URL (File => Open, then paste the URL). Due to Flash's security restrictions, the game doesn't load locally.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

Here is a decrypted version of the game with some trace statements added: [jetvelocity_trace.zip](https://github.com/ruffle-rs/ruffle/files/11195010/jetvelocity_trace.zip)

I recommend loading the game via a local web server so you can compare the trace output with the debug Flash projector.

Notice that both Flash and Ruffle trace "loading image bytes" many times, but only Flash ever traces "onLoaderComplete start!"
The relevant function is `onZipComplete` in `com.devm.util.misc.BitmapZipDispenser`.
```as
(_loc8_ = new Loader()).name = _loc5_.filename;
_loc8_.loadBytes(_loc5_.content);
_loc8_.contentLoaderInfo.addEventListener(Event.COMPLETE,onLoaderComplete,false,0,true);
```
The COMPLETE event of the Loader is never dispatched. I think this may have the same cause as #10463.

n0samu:
Progressed by #10638, #11045, and #11084. With all of those merged in, the main menu background gradient and music load correctly, but that's about it. The new error is:
```
ERROR run_frame:run_all_phases_avm2: ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "addedToStage", class: flash.events::Event, ptr: 0x13f0bf50df8 }) to handler FunctionObject(FunctionObject { ptr: 0x13f0d061cd8 }) : ReferenceError: Error #1069: Property http://www.collada.org/2005/11/COLLADASchema::asset not found on XML and there is no default value.
        at org.ascollada.core::DaeEntity/getNode()
        at org.ascollada.core::DaeDocument/read()
        at org.ascollada.core::DaeEntity()
        at org.ascollada.core::DaeDocument()
[...]

PR #12540, opened 2023-08-03
avm2: Limited XML namespace support

### LINKED ISSUE #10293: Soccer Skill Zone: Blue screen issue
### Describe the bug

https://www.newgrounds.com/portal/view/495537?emulate=flash
Game seems to work up to the point where you start the first level and the gate opens, but it quickly turns into this.

Free Kick Ruffle

Error dispatching event EventObject(EventObject { type: "EVENT_START_LEVEL", class: flash.events::Event, ptr: 0x31e01b8 }) to handler FunctionObject(FunctionObject { ptr: 0x344d6e8 }) : ReferenceError: Error # 1069: Property http ://www .collada.org/2005/11/COLLADASchema::asset not found on XML and there is no default value.

### Expected behavior

Game should work with the gate opening and a soccer ball should drop onto the field.

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 110

### Additional information

_No response_

evilpie:
This fixes #10293 and #10639.

Probably also needs some manual testing to make sure we aren't regressing games that use XML and currently work.


evilpie:
This almost passes `from_avmplus/e4x/Expressions/e11_1_1/`, but there is some unrelated issue with:
```
q = new QName(n, "attr1");
q2 = new QName(q, "attr1");
```

I need to figure out how to run all skipped tests and find out new failing and passing tests.

torokati44:
> I need to figure out how to run all skipped tests and find out new failing and passing tests.

Uh, don't we have a magic command for that?

Dinnerbone:
> I need to figure out how to run all skipped tests and find out new failing and passing tests.

It's automatic, if any test suddenly starts working it'll come up as "failing" and tell you to remove the `known_failure` flag.

No new tests pass because of this PR, sorry!

n0samu:
This also makes [Neopets Lost City Lanes](https://www.neopets.com/games/game.phtml?game_id=1108) playable!

evilpie:
Good to know that we still make sure "failing" tests don't suddenly start working.

What I actually meant was that I would be interested in basically the individual changes inside a test. For example for `e11_1_1` the output changed quite a lot, because a lot more asserts now pass.

Dinnerbone:
> What I actually meant was that I would be interested in basically the individual changes inside a test. For example for `e11_1_1` the output changed quite a lot, because a lot more asserts now pass.

Ah okay. No easy bulk way right just, just run the swf in ruffle desktop and see what it says!

core/src/avm2/e4x.rs
```diff
@@ -475,36 +490,58 @@ impl<'gc> E4XNode<'gc> {
...
-                attribute.key.into_inner(),
-            );
+            let (ns, local_name) = parser.resolve_attribute(attribute.key);
+            let namespace = match ns {
+                ResolveResult::Bound(ns) => Some(AvmString::new_utf8_bytes(
+                    activation.context.gc_context,
+                    ns.into_inner(),
+                )),
+                ResolveResult::Unknown(ns) if ns == b"xmlns" => continue,
+                // TODO: Throw error on unknown prefix
```
Aaron1011, line 508: Can you add an `avm2_stub_method!` call here?

evilpie: I just went ahead and added the correct errors in the second commit.

core/src/avm2/e4x.rs
```diff
@@ -475,36 +489,82 @@ impl<'gc> E4XNode<'gc> {
...
 
+        let (ns, local_name) = parser.resolve_element(bs.name());
+        let name =
+            AvmString::new_utf8_bytes(activation.context.gc_context, local_name.into_inner());
+        let namespace = match ns {
+            ResolveResult::Bound(ns) => Some(AvmString::new_utf8_bytes(
+                activation.context.gc_context,
+                ns.into_inner(),
+            )),
+            ResolveResult::Unknown(ns) => {
```
Aaron1011, line 551: Should this have the `xml` and `xmlns` cases as well?

evilpie: I am not sure if `<xmlns:dummy/>` is even valid XML, because `xmlns:` is used as an attribute usually, but when correctly binding the namespace like this it already works: `<x xmlns:xmlns="http://example.org"><xmlns:a></xmlns:a></x>`.

Interestingly Flash might have a bug related to `xml:`. `<xml:a />` works, but `<xml:a></xml:a>` throws
> The element type "a" must be terminated by the matching end-tag "&lt;/a>".

`<xml:a></a>` works! :exploding_head:

evilpie: So I added support for `xml`, but not `xmlns`, prefixes for elements.

Dinnerbone (approved):

Aaron1011 (approved):

Issue #5579, opened 2021-11-01
The terms on http://www.dnai.org/glossary/ aren't showing definitions when clicked


danielhjacobs:
### Describe the bug

When clicking any term, you see the word "hi". For ease of accessing the link, I will rewrite the website from the issue title: http://www.dnai.org/glossary/

### Expected behavior

e.g.: You are supposed to see the definition "One of the four bases that make up DNA.  Abbreviated with an 'A.'" when clicking "Adenine".

### Affected platform

Self-hosted version

### Operating system

Linux Fedora 34 (64-bit)

### Browser

Mozilla Firefox 93

### Additional information

Despite the "Affected Platform" being "Self-hosted version" and the self-hosted version being somewhat out-of-date, I did try the latest extension as well (ensuring it took priority), and the same issue is still present.

 Looking at the source code of http://www.dnai.org/glossary/glossary.swf in JPEXS, I see heavy use of _root, _parent, and _global. Possibly #768, or something related to iterating over "this"?

In DefineSprite (48: glossaryMC): frame1: DoAction (path based on JPEXS tags):

    a = new Object();
    a.Adenine = "One of the four bases that make up DNA.  Abbreviated with an \'A.\'";
    a.Allele = "An alternate version of a gene, e.g., Gregor Mendel\'s pea plants have flowers with two colors: white and reddish-purple.  The flower color gene in this case has two alleles, one for white and the other for reddish-purple.";
    a["Amino acids"] = "A class of molecules that are the building blocks of proteins.  There are 20 different amino acids used to make up proteins.";
    ...
    z = new Object();
    z.Zygote = "Refers to the cell formed by the union of two gametes; more broadly, refers to the organism that develops from such a cell.";
    _global.glossary = new Object();
    _global.temp = new Object();
    for(aa in this)
    {
       _global.temp[aa] = this[aa];
       _global.glossary[aa] = new Object();
       for(ab in _global.temp[aa])
       {
          _global.glossary[aa][ab] = _global.temp[aa][ab];
       }
       delete this.aa;
    }
    delete _global.temp;

Toad06:
I've added some `trace()` in the code you mentioned and everything looks ok, all the variables are set correctly.

I rather think the issue is in `\DefineSprite (42: bigDescriptionMC)\frame 1\DoAction`. A few observations:

[...]

Issue #7979, opened 2022-09-14
All text show as "test" in Phit and some rendering issues


Pranav-FP:
### Menu
![image](https://user-images.githubusercontent.com/110013681/190086556-d50e3b90-a35a-4ff9-b2c1-9e0ec5797f0f.png)
All the menu text show up as "test" and also rendering is not perfect.

### Level
![image](https://user-images.githubusercontent.com/110013681/190086959-0541bf4f-470a-43c3-b4a1-7842ae26b79e.png)
Only the outlibes are rendered properly

### Menu
![image](https://user-images.githubusercontent.com/110013681/190087115-5cafa729-2404-4da4-9e7f-10e840075e34.png)

#### Level
![image](https://user-images.githubusercontent.com/110013681/190087789-31d1854e-8660-4222-bdb2-0ffd9a68de80.png)

## Affected platform

Desktop app

## Operating system

Windows 11

## Browser

NA

## Additional information

SWF Link - https://www.addictinggames.com/sites/default/files/flash_games/phit_0.swf

Version - ruffle-nightly-2022_09_14 (latest as of writing this)

Pranav-FP:
Related to #5887 I guess

hatal175:
So the issue here is that there's a 'button sprite' -> button -> text heirarchy.
In the button sprite code CPhitButton it assigns a text to this._buttonText which is the variable name for the text. In flash this works fine but ruffle doesn't make that connection, I assume because of scoping issues (assigning to this._button._buttonText fixes this but ruins flash).
I assume #5492 will help? Not sure what is the pull request currently working on this.

n0samu:
Oh interesting, sounds very similar to #2181 then. As far as I understand, this would take quite a big refactor to fix and no one is working on it at the moment.

waspennator:
Blocks are rendering properly now, but the text is still somewhat buggy. I do notice that during gameplay, most of the button text shows up normally save for the quit button, but as soon as the buttons become active as you move pieces around, they start changing text to test.

![Screenshot_20230727_013323_Chrome](https://github.com/ruffle-rs/ruffle/assets/78768499/ab113135-0625-4604-b5a4-8f7f2d47db9d)
![Screenshot_20230727_013400_Chrome](https://github.com/ruffle-rs/ruffle/assets/78768499/e8bc3ddb-0094-49b7-82a4-1170dbb3a7e2)

Lord-McSweeney:
Fixed by #12541.

Issue #4088, opened 2021-04-22
Dynamic text variable location path


Tempy111:
I'm trying to fix this myself but it's... slghtly above my very basic RUST skills..

when you create a button, which has the text being dynamic via a variable which is NOT referenced to Root, it uncorrectly sets a path for the variable, and thus can't find it.

In flash, create a  button or maybe even just a normal clip called 'TextBoxClip'. Inside create a button with no instance name, inside that put a textbox called 'TextBox1', you could do:
`tellTarget (_root.TextBoxClip){
TextBoxText1 = "This is text";`

and it would display that text in the text field..

However, with Ruffle, it wants to use the uninstanced clip as part of the path. Flash has a weird quirk in AS1 and i think AS2 (i'm not sure about AS3) where it didn't even include the button name if it had an instance name.. I'm not quite sure what Ruffle is adding in it's place, probably a null value if the button is name instanced.

if the button is given an instance name, it works fine in ruffle, but since it should work without a name, (in flash didn't check the name anyway) it's a bit of a bug.

I was trying to find where in the ruffle code it works out and checks the path but the best I could find is from line 632 to 648 in the  'ruffle/core/src/display_object/edit_text.rs' file

Herschel:
Related to https://github.com/ruffle-rs/ruffle/issues/1029, button and text fields can't have instance names in SWFv4/5.
Also some what related to https://github.com/ruffle-rs/ruffle/issues/1513 as I think "fake" instance paths are returned if you try to be clever and actually grab the button, for example by doing `_level0.instance1` or `_parent` from a clip inside the button.

Herschel:
Looking at it a bit more, and the behavior is a little strange:

 * Even in SWFv6+, the behavior still occurs; if the text field is the child of a button, the binding will be registered on the parent instead. 
 * This also happens for a variable like `_root.button.foo`, because `foo` is a property of `button`. This actually binds to `_root.foo`.
 * It also includes `_root.button.button.button....button.foo`. This only binds to `_root.foo`, even if there is only one `button` clip...! I think the text field bindings must revert to an "SWF5" mode, and this is behavior is caused by the SWF5 buttons returning their parent, i.e. `_level0` instead of `_level0.button`, so repeated accesses of `_root.button` circle back to `_root`.
 * However, setting `textField.variable = "_root.button.foo"` does not exhibit SWF5 behavior. This binds to "foo" from the button object.

Herschel:
[...]

Issue #11317, opened 2023-05-31
objects with indirect reference (leftover from #9447)


ousia:
After #9447 was merged (and #5492 was subsequently closed), remaining issues with objects indirectly referenced are issues #1561, #2181, #4088, #5579, #6370 #6581 and #8213.

Excuse me for insisting on this, but they seem to be forgotten.

Many thanks for your excellent work.

n0samu:
Not forgotten, but CUB3D put a heroic effort into #9447, and he seems to be taking a well-deserved break! Not many people have what it takes to work on these difficult issues.

n0samu:
I don't think #5492 addressed any of the remaining open issues you mentioned, either

n0samu:
The simpler issue with text variable paths inside buttons (represented by #2181 et al) has been fixed by #12541. There are several issues with text variables remaining though - so far we've found #12588 and #12589, and I'm betting others will come up as well.

PR #12541, opened 2023-08-03
avm1: reference displayObjects parents if parent is a button

### LINKED ISSUE #2181: Prizma Puzzle Challenges always shows score of 99999
SWF URL: http://i.notdoppler.com/files/prizmapuzzlechallenges.swf

The level select screen always shows a score of 99999 for each level, regardless of whether the level has been played or not.
Below is an image showing the incorrect behavior.
image

To see this behavior, load the game and click "Start Game," then mouse over Level 1.

michiel2005:
avm1: reference displayObjects parents if parent is a avm1button. 
Fixes #2181, #4088, #5579, #7979 And possibly more.

This felt as a too easy fix.


danielhjacobs:
Does this fix #5579, which Toad06 thought might be the same issue?

michiel2005:
Indeed it does, thanks :)

waspennator:
Seems to fix the scores showing up as "9999999" when hovering over levels in the level select.

torokati44:
There's a collection of possibly related issues here: https://github.com/ruffle-rs/ruffle/issues/11317

Lord-McSweeney:
Related, but not the exact same issue: #1029
I wonder if this behavior is SWF-version gated, but the relevant issues include SWFv6, v7, and v9 (though noticeably not v4 and v5, which #1029 states as having different behavior because Button doesn't exist in those SWF versions).

Dinnerbone:
Does this specifically only apply to text variables, or are all our paths wrong when it comes to buttons?

michiel2005:
No problem, did that before :)

Dinnerbone (approved):

Dinnerbone:
Thank you!

PR #12555, opened 2023-08-04
build: Reduce disk use of reproducible docker builds

adrian17:
- JRE is now headless (and no JDK at all)
- use a minimal profile in rustup (as in: no docs/clippy)
- no `RUN rustup update` (anyone knows why it was there? Surely rustup installs recent version by default?)
- run `cargo clean` after the built artifact is moved to `dist/`

Result:
- ~0.7GB smaller image before we start building Ruffle
- ~1.1GB of space regained between vanilla and non-vanilla wasm builds (plus another 1.1GB at the end, but that shouldn't matter)

The next big thing would have been removing `miniconda` and pulling recent `binaryen` another way, that's another almost 1GB saved.

Would still like @kmeisthax to review and make sure it still reproduces with it (whether before or after merge).

### DIFF
--- a/web/docker/Dockerfile
+++ b/web/docker/Dockerfile
@@ -7,21 +7,19 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update -y 
 RUN apt-get -y full-upgrade
 RUN apt-get install -y apt-utils
-RUN apt-get install -y wget git openssl libssl-dev gcc gzip tar default-jdk default-jre pkg-config
+RUN apt-get install -y wget git openssl libssl-dev gcc gzip tar default-jre-headless pkg-config
 RUN wget 'https://deb.nodesource.com/setup_lts.x' --quiet -O- | bash
 RUN apt-get install -y nodejs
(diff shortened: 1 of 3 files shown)


Aaron1011 (approved):

relrelb (approved):
I think we can further reduce Docker size by switching over to an Alpine/Distroless image over Ubuntu, but I guess this can be part of another PR.

adrian17:
> I think we can further reduce Docker size by switching over to an Alpine/Distroless image over Ubuntu, but I guess this can be part of another PR.

The issue is that this might hard reproducible builds, as currently the docker tries to be roughly equvalent with the GH Actions builders (which are Ubuntu 22).

A way bigger and probably simpler gain would be removing `miniconda`, which AFAIK is only used to install `binaryen` (as in, `wasm-opt`) but still takes nearly 1GB by itself.
It was originally added from my suggestion in https://github.com/ruffle-rs/ruffle/pull/9121#issuecomment-1386184954 , as it appeared somewhat more reliable than other install methods.

PR #12557, opened 2023-08-04
web: Fix extension vs selfhost loading conflicts

Dinnerbone:
Well, as best as I can.

There's a few issues at play:
1. Things loaded from different sources are different JS types, so we always saw another version of ruffle as a "conflict" and didn't consider it at all. This means that whatever loaded last would take over config, but whatever loaded first would still polyfill because nothing told it to stop.
2. Extension set every single option explicitly, which meant that selfhosted stuff would get overwritten if loaded first.
3. Identical versions (we're still `0.1.0`) compared the same so it was pretty arbitrary which one would win - I made it include the build date as a tie breaker.

### DIFF
--- a/web/packages/core/src/polyfills.ts
+++ b/web/packages/core/src/polyfills.ts
@@ -214,10 +214,7 @@ export function pluginPolyfill(): void {
  * Polyfills legacy Flash content on the page.
  */
 export function polyfill(): void {
-    const usingExtension =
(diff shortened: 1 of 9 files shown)


Dinnerbone:
Got a tentative 👍 from Daniel on discord, merging this for now as it doesn't make things worse - but it seems that Polyfill might still be an issue, even though API based stuff is better.

PR #12579, opened 2023-08-05
Exporter: Updated the logic that gives a error message

michiel2005:
If you used exporter to export a frame and accidently gave a path that dis not exist it would say: Output directory is required when exporting multiple files. instead of giving the error that the path does not exist.

I found that confusing so fixed that in this little pr.


Dinnerbone (approved):

Dinnerbone:
Thank you!

PR #12580, opened 2023-08-05
avm2: Skip nested 'goto' frame for SWF 9

Aaron1011:
This fixes 'who_killed_travolta'


Dinnerbone (approved):

jarca0123:
This also fixes this game: https://www.superhry.cz/games/1692/

Aaron1011:
After some additional testing, this seems to be incorrect - we run the constructors for child objects at an odd time.

adrian17:
> After some additional testing, this seems to be incorrect - we run the constructors for child objects at an odd time.

For the record, Aaron's later comment from discord

> It's "less wrong" now
> it's still not correct, but I think we're going to need to do some invasive framscript refactoring to get it right for swf 9

adrian17 (approved):

n0samu:
For posterity, here is the download link for Who Killed Travolta: https://files.scene.org/view/parties/2006/assembly06/browser_demo/who_killed_travolta_by_evo_flash_bombsquad.zip

Issue #3156, opened 2021-02-08
Text shown in SWF is not in the right place and cut off, and text shown when hovering is inaccurate too


danielhjacobs:
**Describe the bug**

This is a visual bug and a bit hard to explain. The bug can be seen in action when using the browser extension here:

_(Note: not reproducible anymore, see below for a SWF)_
HTML Page: http://www.dnai.org/timeline/index.html
SWF File: http://www.dnai.org/timeline/timeline_header.swf

**Expected behavior**

The problem is a bit hard to explain, so I have an image that explains it better than I can:
![ruffle_header_differences](https://user-images.githubusercontent.com/29206584/107202000-adc96f00-69c7-11eb-9c91-d09d7452c55c.png)

**Is the problem with the Ruffle desktop app, extension, or self-hosted version?**

It applies to the extension and demo, and based on testing on my test server, also to the self-hosted version. Since it's a rendering/formatting issue, it seems to apply to every version

**What platform are you using?**

 I am using a Windows 10 computer, and tested both Chrome and Firefox

craniumcadoo:
I see there is a pull request out that would fix part of the issue, but that the author of said pull request said they don't fully understand the rest of the issue.

I doubted having more SWF files that are almost exactly the same and all have the exact same error on hover would help identify the other issue, which is why I did not include those files initially since I thought it would be redundant. but in case it helps, the following all have the exact same issue:

* http://www.dnai.org/a/ - http://www.dnai.org/a/header_mc1.swf
* http://www.dnai.org/b/ - http://www.dnai.org/b/header_mc2.swf
* http://www.dnai.org/c/ - http://www.dnai.org/c/header_mc3.swf
* http://www.dnai.org/d/ - http://www.dnai.org/d/header_mc4.swf
* http://www.dnai.org/e/ - http://www.dnai.org/e/header_mc5.swf

craniumcadoo:
Further testing:

http://www.dnai.org/timeline/, http://www.dnai.org/c/, and http://www.dnai.org/e/ - Upon moving the mouse to hover over (4), then (5), then back to (4), then ... repeatedly, the text just keeps moving further right, until it is eventually offscreen completely. 

The same applies to http://www.dnai.org/a/, however for that one, more often than not, the text starts so far to the right that it is already offscreen initially.

http://www.dnai.org/b/, http://www.dnai.org/d/ - In these two files, the same action moves the text to the left until  it is offscreen.

[...]

Issue #10660, opened 2023-04-12
Text width miscalculated when setting autoSize on centered or right-aligned text field


n0samu:
### Describe the bug

Consider the following sample: [NanacaScoreField.zip](https://github.com/ruffle-rs/ruffle/files/11209875/NanacaScoreField.zip). This is a simpler demonstration of #1440.

Here is the AS code:
```as
score_txt.autoSize = "right";
score_txt.text = "999.99";
```

In Ruffle, it looks like this:
![image](https://user-images.githubusercontent.com/71368227/231421473-81b8dbdb-eaca-4bf9-8acc-ebbc3047b3f2.png)

Notice that the score text is cut off.

### Expected behavior

Here is how it appears in Flash:
![image](https://user-images.githubusercontent.com/71368227/231421343-5a0a8223-6020-4d33-bc5c-bd11727325e7.png)

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

The score_txt field is set to right alignment in the Flash editor. The problem persists if the field is set to center alignment, and disappears if the field is set to left alignment. The problem disappears if the AS code that sets `autoSize` is removed, but persists if `autoSize` is set to "left", "right", or "center".

I tried adding the following line after https://github.com/ruffle-rs/ruffle/blob/aef3ed2fcb00c3df0ef21230c52971f30ade24de/core/src/display_object/edit_text.rs#L730
`println!("bounds: {:?}\nintrinsic bounds: {:?}", edit_text.bounds, intrinsic_bounds);`

According to my understanding of the concept of "intrinsic bounds", they should always start at (0, 0). I noticed that the intrinsic bounds have a nonzero `offset_x` when there is a problem, whereas `offset_x` is 0 when Ruffle is working correctly.

SuchAFuriousDeath:
https://github.com/ruffle-rs/ruffle/blob/690ce110d84dcf8925d60247f685587d8c53adfc/core/src/display_object/edit_text.rs#L746

If we swap intrinsic_bounds.width() for intrinsic_bounds.extent_x(), it seems to be fixed. Even the original #1440. No idea why though, I just messed around.

n0samu:
If it works and doesn't break other tests, I say go for it!

SuchAFuriousDeath:
I ran all tests and all succeeded.

n0samu:
[...]

PR #12598, opened 2023-08-06
core: fix center and right alignment of edit_text

### LINKED ISSUE #1440: Nanaca Crash: Misaligned 'record' numbers
image

SuchAFuriousDeath:
Fixes #1440 and its minimal reproduction #10660. I'm not really sure why this works, but it doesn't seem to break anything else

### DIFF
--- a/core/src/display_object/edit_text.rs
+++ b/core/src/display_object/edit_text.rs
@@ -743,7 +743,7 @@ impl<'gc> EditText<'gc> {
         if autosize != AutoSizeMode::None {
             if !is_word_wrap {
                 // The edit text's bounds needs to have the padding baked in.
-                let width = intrinsic_bounds.width() + padding;
+                let width = intrinsic_bounds.extent_x() + padding;
                 let new_x = match autosize {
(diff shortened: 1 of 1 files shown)


Dinnerbone:
Can you please add a test for this? We need more coverage of text stuff <3

SuchAFuriousDeath:
I can try :D

n0samu:
This also fixes #3156. 👍

core/src/display_object/edit_text.rs
```diff
@@ -743,7 +743,7 @@ impl<'gc> EditText<'gc> {
         if autosize != AutoSizeMode::None {
             if !is_word_wrap {
                 // The edit text's bounds needs to have the padding baked in.
-                let width = intrinsic_bounds.width() + padding;
+                let width = intrinsic_bounds.extent_x() + padding;
```
Lord-McSweeney, line 746: Based on nosamu's tests, this seems wrong. I have no experience in this part of Ruffle, but this seems like it's just adding padding to fix right alignment, and it breaks center alignment.

Issue #11928, opened 2023-07-06
Papas Pastaria - tomato sauce is invisible


MartySVK:
### Describe the bug

I noticed new regression in Pastaria, when I use tomato sauce, the texture is invisible while pouring. It will become visible at the end. This was working before.
The last working build is Nightly 2023-06-23. Probably caused by cacheAsBitmap - https://github.com/ruffle-rs/ruffle/pull/11638

SWF - https://data.gameflare.com/games/4643/LZA1w2C4888Z3a.swf

![image](https://github.com/ruffle-rs/ruffle/assets/10010569/5a12f53a-6939-422f-a573-875466a87956)
![image](https://github.com/ruffle-rs/ruffle/assets/10010569/61c77f99-c712-47ab-91d1-033334c8ea31)

### Expected behavior

visible tomato sauce while pouring.

### Affected platform

Desktop app

### Operating system

Windows 10

hoodamath:
i recently tested a new build of 8-6 and it is still not showing

PR #12664, opened 2023-08-09
core: Better handle scroll_rect and/or masks with cacheAsBitmap

### LINKED ISSUE #12030: Papa's series: Rendering regression in Jojo's Burger Slots minigame
### Describe the bug

In Ruffle Nightly 2023-06-24 and later, the Jojo's Burger Slots minigame in the Papa's games (such as [Pancakeria](https://www.flipline.com/games/papaspancakeria/index.html)) renders incorrectly, with the food graphics continuing to move outside of the bounds of the slot machine. See the video below:

https://github.com/ruffle-rs/ruffle/assets/71368227/9068f761-9953-40f1-b123-c7f45da54635

To reproduce the bug: Start a new game and play through the first ingame day (it's just the tutorial). Then play the minigame afterwards (it will always be the slots minigame).

### Expected behavior

The graphics should only be visible within the slot machine, not outside it.

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

I've bisected this to commit fe742194b106852a84cad78ba022d7abe194a4c0 from #11638.

Dinnerbone:
Instead of rendering the scrolled and masked object to a texture and reusing that for the CAB, it now renders the whole object to the texture and scrolls/masks that.

This should be much faster when using scroll rects as we no longer invalidate on every scroll, but instead scroll the texture we make.
Also a bit faster when using masks because we don't invalidate on mask changing.

This fixes #12030 and #11928


SN902:
[V8MuscleCars.zip](https://github.com/ruffle-rs/ruffle/files/12307792/V8MuscleCars.zip) now on the title screen we have invisible buttons (but clickable, you still can start the game)

image

adrian17 (approved):

PR #12689, opened 2023-08-11
tests: Fix running tests without img_tests feature

Dinnerbone:


torokati44 (approved):
Would've been nice if CI caught this, but oh well...

Dinnerbone:
Who tests the test framework? Probably the same people who watch the watchers.

Issue #11729, opened 2023-06-25
Adventure Story - Graphics and mechanics issues + some uplayable portions


Kz-92:
### Describe the bug

now that the chest softlock is gone, we can move on to some of the other issues
obviously, a lot of filters are needed. The background doesn't blur, the coins do not turn silver when having already collected them and returning to a level, almost all text in the game is unreadable (chests, signs, titlecards), etc

aside from that, there is also an issue specifically with how titlecards appear, when transitioning to other areas, (for example, 1-2) it will instead show the previous titlecard, (same for 1-3, it shows the 1-2 titlecard). 
additionally, magic attacks seem to be broken, as they act strangely and do not hit where they would normally (within the target reticule) and instead somehow hit targets way offscreen.

flash

https://github.com/ruffle-rs/ruffle/assets/103858019/3dcd83b0-0d7b-41ad-bd6a-0a025a09060a

Ruffle

https://github.com/ruffle-rs/ruffle/assets/103858019/01f6b1fb-7cb1-4ba8-9e4c-5ee4972b730b

and finally, level 2 is still unplayable

https://github.com/ruffle-rs/ruffle/assets/103858019/1d24b701-4b76-4e26-8ce9-89fec0975774

https://github.com/ruffle-rs/ruffle/assets/103858019/88a0d884-5349-409b-bb68-84953caad154

### Expected behavior

ruffle needs some improvment to match flash player

### Affected platform

Desktop app

### Operating system

windows 10

### Additional information

[586861_level01.zip](https://github.com/ruffle-rs/ruffle/files/11860170/586861_level01.zip)

Kz-92:
this is the logs I get when playing and doing all the actions mentioned 
```
ERROR ruffle_core::loader: Error during movie loading of "http://www.mochiads.com/static/lib/services/MochiLC.swf": HttpNotOk("HTTP status is not ok, got 404 Not Found", 404, true)
ERROR ruffle_core::loader: Error during movie loading of "http://www.mochiads.com/static/lib/services/services.swf?listenLC=%5F%5Fms%5F1687708433996%5F52175&mochiad%5Foptions=null&api%5Fversion=3%2E9%2E4%20as3": HttpNotOk("HTTP status is not ok, got 404 Not Found", 404, true)
ERROR ruffle_core::loader: Error during movie loading of "//apifiles.ngfiles.com/medals/21000/21866/10655_iceisslippery.webp?0": FetchError("The network path was not found. (os error 53)")
[...]

Issue #9621, opened 2023-02-18
Gotta Catch'em All one more time - commission: Text Box and missing glow issues


waspennator:
### Describe the bug

[Gotta Catch'em All one more time - commission.zip](https://github.com/ruffle-rs/ruffle/files/10772732/Gotta.Catch.em.All.one.more.time.-.commission.zip)

This SWF seems to work alright save for two issues, the first is that the animation when you hover over one of the ponies doesn't play correctly (They leave this weird afterimage of themselves and the texbox pops up instantly without text.) The second issue is that there's a missing glow effect throughout the swf.

[donbcfbmhbcapadipfkeojnmajbakjdc-1676683478753.log](https://github.com/ruffle-rs/ruffle/files/10772750/donbcfbmhbcapadipfkeojnmajbakjdc-1676683478753.log)

![Catchemall Ruffle](https://user-images.githubusercontent.com/78768499/219824436-d33825a8-e01d-4afd-b511-ceb909ac4d80.png)

### Expected behavior

The text box animation should play out properly when hovering over a pony and they should glow properly and not have this weird afterimage effect.

![Catchemall Flash](https://user-images.githubusercontent.com/78768499/219824447-938c5170-3e4a-4edc-8166-9387ffc11d42.png)

### Affected platform

Desktop app

### Operating system

Windows 11

waspennator:
Still an issue as of the latest nightly, ran it through the ruffle web demo and seeing some stuff, though I dunno what's causing the empty text box issues.

[ruffle.rs-1684780728893.log](https://github.com/ruffle-rs/ruffle/files/11535866/ruffle.rs-1684780728893.log)

waspennator:
Tried running it through the JPEXS decompiler to find out clues on how to get it working and posting the script of one of the description boxes as an example.

[terrafirma description.zip](https://github.com/ruffle-rs/ruffle/files/11780789/terrafirma.description.zip)

waspennator:
![image](https://github.com/ruffle-rs/ruffle/assets/78768499/54f2912f-91f5-43aa-982f-d0a51fed072d)
Filter situation looking better now, but text boxes when hovering are still borked and now the quality texture on the bottom left is cropped awkwardly from the left now.

waspennator:
Text boxes are still borked as of latest nightly. ran it through the decompiler again i'm thinking this might be a TLF Text issue, since searching up one of "Display object has no child with name" errors (I searched up id3) takes me to this code wise.

[...]

PR #12690, opened 2023-08-11
core: Fix filters being cut off when nested inside another cacheAsBitmap

Dinnerbone:
This should fix any remaining filter-cutoff issue, such as the glow in #11729 (https://github.com/ruffle-rs/ruffle/issues/11729#issuecomment-1632913648).

### DIFF
--- a/core/src/display_object.rs
+++ b/core/src/display_object.rs
@@ -794,7 +794,7 @@ pub fn render_base<'gc>(this: DisplayObject<'gc>, context: &mut RenderContext<'_
                 for filter in &mut filters {
                     // Scaling is done by *stage view matrix* only, nothing in-between
                     filter.scale(stage_matrix.a, stage_matrix.d);
-                    filter_rect = context.renderer.calculate_dest_rect(filter, filter_rect);
+                    filter_rect = filter.calculate_dest_rect(filter_rect);
                 }
                 let draw_offset = Point::new(filter_rect.x_min, filter_rect.y_min);
                 if cache.is_dirty(&base_transform.matrix, width, height) {
--- a/core/src/display_object.rs
+++ b/core/src/display_object.rs
@@ -785,10 +785,10 @@ pub fn render_base<'gc>(this: DisplayObject<'gc>, context: &mut RenderContext<'_
                 let width = width as u16;
                 let height = height as u16;
                 let mut filter_rect = Rectangle {
-                    x_min: 0,
-                    x_max: width as i32,
-                    y_min: 0,
-                    y_max: height as i32,
+                    x_min: Twips::ZERO,
+                    x_max: Twips::from_pixels_i32(width as i32),
+                    y_min: Twips::ZERO,
+                    y_max: Twips::from_pixels_i32(height as i32),
                 };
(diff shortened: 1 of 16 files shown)


waspennator:
Fixes the glow on the quality text being slightly cutoff on #9621

swf/src/types/blur_filter.rs
```diff
@@ -31,10 +31,10 @@ impl BlurFilter {
         self.num_passes() == 0 || (self.blur_x <= Fixed16::ONE && self.blur_y <= Fixed16::ONE)
     }
 
-    pub fn calculate_dest_rect(&self, source_rect: Rectangle<i32>) -> Rectangle<i32> {
+    pub fn calculate_dest_rect(&self, source_rect: Rectangle<Twips>) -> Rectangle<Twips> {
         let scale = PASS_SCALES[self.num_passes().clamp(1, 15) as usize - 1];
-        let x = (scale * self.blur_x.to_f32()).ceil().max(0.0) as i32;
-        let y = (scale * self.blur_y.to_f32()).ceil().max(0.0) as i32;
+        let x = Twips::from_pixels(scale * self.blur_x.to_f64());
+        let y = Twips::from_pixels(scale * self.blur_y.to_f64());
```
torokati44: Is the `.ceil().max(0.0)` here not necessary anymore? Or was this an accidental change?

torokati44:
Just curious, why the switch to `Twips`?

swf/src/types/blur_filter.rs
Dinnerbone: Hmm, max should stay, but ceil is handled by twips->int

Dinnerbone:
Because bounding boxes work in Twips, and that's where this is mostly called now

core/src/display_object.rs
```diff
@@ -1166,17 +1179,29 @@ pub trait TDisplayObject<'gc>:
     /// This differs from the bounds that are exposed to Flash, in two main ways:
     /// - It may be larger if filters are applied which will increase the size of what's shown
     /// - It does not respect scroll rects
-    fn render_bounds_with_transform(&self, matrix: &Matrix) -> Rectangle<Twips> {
+    fn render_bounds_with_transform(
+        &self,
+        matrix: &Matrix,
+        include_own_filters: bool,
+        view_matrix: &Matrix,
+    ) -> Rectangle<Twips> {
```
torokati44, line 1187: Function signature changes without doc comment changes are always _sus_, but in this case, I think the same comment still applies and is sufficient.

Dinnerbone: I wrote that comment just the other day whilst thinking about this change 😛

swf/src/types/blur_filter.rs
Dinnerbone: Fixed!

torokati44 (approved):
Mostly a rubber-stamp, but LGTM! And happy with all the stuff it fixes!

torokati44:
Hmh, I guess a visual test would have been nice... Maybe later?

Dinnerbone:
There was one...? 😅

torokati44:
Right, yes, sorry, my bad.

PR #12710, opened 2023-08-12
core: Fix avm2 button bounding box when using scroll rects

Dinnerbone:
Noticed this the other day (thus the comment),

Potentially the cause of some button clicking bug somewhere, but I don't know for sure.


core/src/display_object/avm2_button.rs
```diff
@@ -620,8 +620,17 @@ impl<'gc> TDisplayObject<'gc> for Avm2Button<'gc> {
     }
 
     fn bounds_with_transform(&self, matrix: &Matrix) -> Rectangle<Twips> {
-        // [NA]: The diff between this and the base impl is a lack of scroll_rect.
-        // Intentional, or bug?
+        // A scroll rect completely overrides an object's bounds,
+        // and can even the bounding box to be larger than the actual content
```
Lord-McSweeney, line 624: nit: grammatical mistake "and can even the bounding box"
(unless "even" is being used as a verb, but that also seems wrong)

Dinnerbone: Literally just copy pasted from the super method, didn't even read it 😅

I can change them both

adrian17 (approved):

PR #12712, opened 2023-08-12
avm1: Hook up Button.cacheAsBitmap and TextField.filters

Dinnerbone:
Just same impl as elsewhere


torokati44:
While I don't know enough about these parts to be confident enough to put a green checkmark on this, but I'm tentatively saying that this _looks good enough_ and that it _should be fine_.

adrian17 (approved):

PR #6119, opened 2022-01-24
extension: URL-encode the SWF URL for the player page

n0samu:
When the user clicks the "Open in a new tab" button, the extension's internal `player.html` page opens, with a URL parameter `?url=[the SWF URL]`. Right now, this URL parameter is not properly encoded. This causes a problem when we get to https://github.com/ruffle-rs/ruffle/blob/master/web/packages/extension/src/player.ts#L14:
> `const swfUrl = url.searchParams.get("url");` 

Because the value of the "url" parameter was not properly encoded, any `&` character in the SWF URL would be interpreted as the start of a *new* parameter, instead of being part of the actual SWF URL. So anything after the first `&` symbol in the SWF URL would be chopped off.
On Itch.io Flash pages such as https://rarykos.itch.io/hug-me-im-cold, the SWF URL contains several important parameters - if they are not included, the server returns 403 Forbidden (it's a type of hotlinking protection). So when the user clicked the "Open in a new tab" button on Itch.io pages, the SWF would refuse to load in the player.
This PR fixes the problem - Itch.io SWFs load when opened in a new tab after my change is applied.

Issue #448, opened 2020-03-16
[Web] Ruffle does not work with SWFObject


Toad06:
Ruffle will fail to run any Flash file using SWFObject. For example, the following code will not work (unless Flash is enabled... but that's not really useful 😜):

> swfobject.embedSWF("uniwar.swf", "file_div", "550", "400", "8");

In the attached sample:
- Add a nightly build in "ruffle/".
- Make sure Flash is disabled and run "test.html". You should see "Flash Player not detected!" instead of the "Click to play" by Ruffle.
- Eventually run "extension.html" using the browser's extension, you will notice the same issue.
- Run "workaround.html", things should work as expected because swfobject.embedSWF() has been rewritten and using "const swfobject" prevents SWFObject to run (but that will trigger an error in the console... 😅).

[swfobjectTest.zip](https://github.com/ruffle-rs/ruffle/files/4338003/swfobjectTest.zip)

Herschel:
https://github.com/swfobject/swfobject/blob/master/swfobject/src/swfobject.js#L41 is the SWFObject code that detects the Flash Player version. This runs before the Ruffle code that installs the fake plugin, because Ruffle will not install polyfills until the `DOMContentLoaded` event. We should at least install the plugin spoof immediately when Ruffle is initialized.

Secondly, Ruffle returns "Shockwave Flash 32.0 r0 (compatible; Ruffle 0.0.1)" for the plugin version, but this breaks SWFObject's version parsing at https://github.com/swfobject/swfobject/blob/master/swfobject/src/swfobject.js#L58-L60. We should just return the standard version string; people could detect a Ruffle install in other ways if they needed to.

ROBERT-MCDOWELL:
@Herschel 
whiy not to make a PR to make swfObject compatible to ruffle?

Herschel:
This should work now as long as the Ruffle script tag is above the SWFObject script tag (we should document this).

SWFObject does spit out a JS error in the console, but it seems okay -- it tries to call object.GetVariable on a fake flash embed for a more sophisticated version check, but Ruffle would have already removed the object.

Toad06:
Tested with today's JS build (2020-03-27), it works as described above in Firefox and Chrome.

However I've not been able to make it using the extension. I'm not sure if it's as intended, or if it requires some extra configuration.

Toad06:
@Herschel ,
This is still not working using the extension. I've found out that if you add this in "manifest.json":
`"run_at": "document_start"`
[...]

PR #12717, opened 2023-08-12
web: Update to manifest v3 where supported (Chrome/Edge)

Dinnerbone:
Doesn't take advantage of any new v3 stuff, just gets it working as-is


danielhjacobs:
In testing, mostly works, except the `injectScriptRaw` on Chromium on many websites, as mentioned on Discord.

n0samu:
Testing this in Edge, it definitely seems to have worsened the "extension loads too late" situation. On http://www.g2conline.org/ and Toad06's swfobject test from #448, navigating to and from the page using the browser's Back and Forward buttons almost always causes the Flash detection scripts to fail, whereas on the current nightly build, this almost never happens.

n0samu:
Also, this breaks the "Open in a new tab" feature for URLs not ending in .swf, such as Itch.io Flash pages. ([Here's an example](https://rarykos.itch.io/hug-me-im-cold) from #6119.) We should make the "Open in a new tab" button link directly to the extension's internal player page (for both the Chrome and FF extension, because why not).

Dinnerbone:
> Also, is there a reason to make the internal player page load SWFs either via the `url` parameter or the hash fragment? I think we should stick to one or the other.

The previous way was `?url=foo`, but that won't work with the redirect options we have available, as we can't uri encode the swf url. Hash works, and imo looks so much cleaner too.

I'll switch Firefox to using hash too soon, but didn't want to change how Firefox worked in this PR.

Dinnerbone:
> Also, this breaks the "Open in a new tab" feature for URLs not ending in .swf, such as Itch.io Flash pages. ([Here's an example](https://rarykos.itch.io/hug-me-im-cold) from #6119.) We should make the "Open in a new tab" button link directly to the extension's internal player page (for both the Chrome and FF extension, because why not).

I had a try at this, but Chrome (at least, I didn't try the others) blocks URLs that lead directly to extension pages.

Dinnerbone:
Chrome should now run the plugin polyfill immediately

n0samu:
> I had a try at this, but Chrome (at least, I didn't try the others) blocks URLs that lead directly to extension pages.

I guess the only solution then is to send a message to the background script and have it open the player page when it receives the message. I guess it would be ok to leave that for a separate PR but I hope we can get it done soon, so people don't start noticing stuff breaking.

danielhjacobs (approved):
Just tested this in Chromium and it seems to be working well. Also had a quick look over the code and it LGTM. Hopefully n0samu's suggestion can be implemented for non-swf direct links, but that can be a follow-up PR. Guess we should keep a close eye on https://bugzilla.mozilla.org/show_bug.cgi?id=1687755 so we can use this manifest for Firefox.

PR #12720, opened 2023-08-12
wgpu: Yeet the copy_srgb shader and pipeline

torokati44:
This workaround won't be needed at all after either we move to wgpu 17.0.0, or https://github.com/gfx-rs/wgpu/pull/3817 gets backported to wgpu 0.16.x.
And even until that, the desktop player is now fine even with OpenGL thanks to the indirect movie rendering through egui.
As for Android, I'm going to temporarily move to a patched wgpu 0.16 with the fix linked above.


Dinnerbone:
This surely will just make it crash when the format is different, no? 😅

You still gotta tell it to change the srgbness

torokati44:
> This surely will just make it crash when the format is different, no? sweat_smile

As in, what format is different from what? :no_mouth: The one we request vs. the one we end up getting?

> You still gotta tell it to change the srgbness

And how would I do that? Sorry, I still don't speak wgpu... :|

Dinnerbone:
If the texture you're copying is srgb but you're copying it to rgb, or vice verca.

When you make the texture view, you can specify that you want it in the same format but with/without srgb - that's how we should do it.

torokati44:
Aaah, right!
https://gpuweb.github.io/gpuweb/#dom-gputexturedescriptor-viewformats
> Two [GPUTextureFormat](https://gpuweb.github.io/gpuweb/#enumdef-gputextureformat)s format and viewFormat are texture view format compatible if:
>
>   - format equals viewFormat, or
>   - format and viewFormat differ only in whether they are srgb formats (have the -srgb suffix).

This will have to wait though. Unless you want to do it in the meantime...

Lord-McSweeney:
Superseded by #23827

Issue #12734, opened 2023-08-13
Regression: CABs with scrollRect cut off when scaled


n0samu:
### Describe the bug

Starting with commit a9c5bdfbab06c94cdd9bf543b63e1a9d148c1c64 (#12664), CABs (bitmap-cached display objects) with scrollRects set are cut off when scaled. This can be observed in many games; here are some examples:
- In [Bleach vs. Naruto](https://flashga.me/public/storage/games/bleach-vs-naruto-2.6-files/launcher.swf), scaling up the game cuts off all but the top left corner. That's because the main movieclip has CAB enabled with a scrollRect. Unchecking "Cache as bitmap" for the clip in the debug tools solves the issue.
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/977a00e3-a65c-4c81-aa8e-dceaa097adfb)
In [HitBox](http://i.notdoppler.com/files/hitbox.swf), the same problem happens only after beating a level, because a CAB movieclip with a scrollrect appears then.
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/8f94e045-3f3b-4e25-b873-f91e561449ff)
- In [Bloons Super Monkey](http://i.notdoppler.com/files/bloonssupermonkey.swf), when the game is scaled up, the bloon meter appears glitched. The bloon meter also uses scrollRect.
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/ed76cec4-a4e3-44b8-98bc-aa4fe475ac02)

### Expected behavior

These problems should not occur

### Affected platform

Desktop app

### Operating system

Windows 10

Lord-McSweeney:
Fixed by #12743.

PR #12743, opened 2023-08-13
core: Fix rendering masks + scrollRects when cacheAsBitmap

Dinnerbone:


n0samu:
Can confirm that this fixes #12734! I found a remaining regression in #12788 but this can be fixed separately.

PR #12770, opened 2023-08-14
avm2: XML objects must support lookups like "@attr"

### LINKED ISSUE #10228: Zap Apple Buckin': Doesn't start
### Describe the bug

https://arcade.equestriagaming.net/2018/08/zap-apple-buckin.html
Game doesn't seem capable of starting after the preloader, checked through the log and I have a feeling this might or might not be a culprit.

AVM2 error: String("snippet is missing attribute id")

[arcade.equestriagaming.net-1679177565702.log](https://github.com/ruffle-rs/ruffle/files/11009731/arcade.equestriagaming.net-1679177565702.log)

### Expected behavior

Game should start up and be playable.

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 110

### Additional information

_No response_

### LINKED ISSUE #12765: bullet hell adventure the game loads then goes to a black screen
### Describe the bug

http://www.onemorelevel.com/game/bullet_hell_adventure

https://github.com/ruffle-rs/ruffle/assets/136838819/1cc3da0c-3de0-400f-b864-c34d1d1490ed

### Expected behavior

for after the game loads it goes to the game instead of a black screen

### Affected platform

Browser's extension

### Operating system

windows 11

### Browser

microsoft edge

### Additional information

_No response_

evilpie:
So Haxe's uses `o.hasOwnProperty(field)` and `o[field]` to lookup attributes as strings on XML objects.

We can support this by manually re-parsing the multiname in get_property_local. `has_own_property` doesn't have an activation, which makes this difficult. I temporarily solved this by adding the hasOwnProperty method explicitly to the XML class, which was missing anyway.

I tried adding an activation parameter to has_own_property before, but gave up because it required a lot of other changes to pass an activation around.

Fixes #12765 (still unplayable slow) and fixes #10228

### DIFF
--- a/core/src/avm2/object/xml_object.rs
+++ b/core/src/avm2/object/xml_object.rs
@@ -1,7 +1,7 @@
 //! Object representation for XML objects
 
 use crate::avm2::activation::Activation;
-use crate::avm2::e4x::{name_to_multiname, E4XNode, E4XNodeKind};
+use crate::avm2::e4x::{string_to_multiname, E4XNode, E4XNodeKind};
 use crate::avm2::error::make_error_1087;
 use crate::avm2::object::script_object::ScriptObjectData;
 use crate::avm2::object::{ClassObject, Object, ObjectPtr, TObject, XmlListObject};
@@ -212,6 +212,19 @@ impl<'gc> TObject<'gc> for XmlObject<'gc> {
             }
         }
 
+        // Special case to handle code like: xml["@attr"]
+        let multiname = if !name.has_explicit_namespace()
+            && !name.is_attribute()
+            && !name.is_any_name()
+            && !name.is_any_namespace()
+        {
+            name.local_name()
+                .map(|name| string_to_multiname(activation, name))
+        } else {
+            None
+        };
+        let name = multiname.as_ref().unwrap_or(name);
+
         let matched_children = if let E4XNodeKind::Element {
             children,
             attributes,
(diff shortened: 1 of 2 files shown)


core/src/avm2/globals/XML.as
```diff
@@ -209,6 +215,10 @@ package {
             return "XML";
         };
 
+        prototype.hasOwnProperty = function(prop:String):Boolean {
+            return this.hasOwnProperty(prop);
```
Lord-McSweeney, line 219: Just checking- this doesn't need the `self.AS3::` logic, right? Could you add a comment noting that?

sleepycatcoding: If I were to guess it probably does not need it as the `hasOwnProperty`method is not in AS3 namespace.

core/src/avm2/object/xml_object.rs
```diff
@@ -196,6 +196,29 @@ impl<'gc> TObject<'gc> for XmlObject<'gc> {
             }
         }
 
+        let multiname = if !name.has_explicit_namespace() {
```
evilpie, line 199: I doubt this is really correct. I think it's likely that this could modify Multinames that are already names used for XML and we would e.g. strip two leading @@.

Aaron1011: I think this could be fixed by changing the check to `!name.has_explicit_namespace() && !name.is_attribute()`

evilpie:
I've removed the implementation of XML::hasOwnProperty from this patch, it's not really necessary since we have #10864.

Lord-McSweeney (approved):

core/src/avm2/object/xml_object.rs
```diff
@@ -212,6 +212,23 @@ impl<'gc> TObject<'gc> for XmlObject<'gc> {
...
+            && !name.is_any_name()
+            && !name.is_any_namespace()
+        {
+            name.local_name()
+                .map(|name| string_to_multiname(activation, name))
+        } else {
+            None
+        };
+
+        let name = match multiname {
```
Aaron1011, line 227: Can this just be `let name = multiname.or(name)`

evilpie: `or` takes an `Option`, but that made me realize that `as_ref.unwrap_or` works.

PR #12795, opened 2023-08-16
extension: Use internal player page for spl files too

danielhjacobs:
Unfortunately, https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/declarativeNetRequest/RuleCondition has no condition to allow matching the headers like https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/onHeadersReceived allows.

### DIFF
--- a/web/packages/extension/src/background.ts
+++ b/web/packages/extension/src/background.ts
@@ -52,7 +52,7 @@ async function enable() {
                     redirect: { regexSubstitution: playerPage + "#\\0" },
                 },
                 condition: {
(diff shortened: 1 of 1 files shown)


relrelb (approved):

danielhjacobs:
No longer relevant. We no longer do direct SWF link takeover because it doesn't work properly without MIME-type detection and honestly probably annoyed users as often as it helped them. See #12835.

danielhjacobs:
The replacement is an "Open SWF Player" option in the extension options popup. See #12836 for that.

PR #12804, opened 2023-08-17
avm2: nextScene() and prevScene() should fall back to current scene

### LINKED ISSUE #12797: works only with right click play https://www.hoodamath.com/games/waterballoons.html
### Describe the bug

it plays the loading screen but then shows black screen instead of menu
but if you right click play the game works

### Expected behavior

menu appear after loading screen

### Affected platform

Self-hosted version

### Operating system

macos

### Browser

google chrome

### Additional information

if you right click play it works

n0samu:
Fixes #12797

Note that even though `previous_scene` and `next_scene` are in `core`, they are only used by AVM2. In AVM1, nextScene() and prevScene() are compiled to `GotoFrame` instructions.


Dinnerbone (approved):

hoodamath:
but the game still doesn't play correctly?

Issue #9917, opened 2023-03-08
Error on https://kdata1.com/2016/02/kof-vs-dnf/


zqz979:
# Error Info
Error name: Error
Error message: panicked at 'called `Option::unwrap()` on a `None` value', core/src/library.rs:317:9
Error stack:
```
Error: panicked at 'called `Option::unwrap()` on a `None` value', core/src/library.rs:317:9
    at imports.wbg.__wbg_new_15d3966e9981a196 (chrome-extension://donbcfbmhbcapadipfkeojnmajbakjdc/dist/339.js:2154:21)
    at std::sync::once::Once::call_once::{{closure}}::h232d3d481290a400 (wasm://wasm/0294a55a:wasm-function[2037]:0x4f1ae7)
    at std::sys_common::once::generic::Once::call::h6b012ac601cc39bf (wasm://wasm/0294a55a:wasm-function[1972]:0x4e2c41)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::h68563d86acb3fcd8 (wasm://wasm/0294a55a:wasm-function[7469]:0x7b4d0f)
    at std::panicking::rust_panic_with_hook::hbafe3e603d331223 (wasm://wasm/0294a55a:wasm-function[5272]:0x70ca1c)
    at core::panicking::panic_fmt::h1d17fc068f528130 (wasm://wasm/0294a55a:wasm-function[7800]:0x7c3ff7)
    at core::panicking::panic::h27b5eefa3e4ff738 (wasm://wasm/0294a55a:wasm-function[7190]:0x78fe49)
    at ruffle_core::display_object::movie_clip::MovieClip::preload::hdaaba4814010f936 (wasm://wasm/0294a55a:wasm-function[600]:0x1d9e04)
    at ruffle_core::loader::Loader::preload_tick::hd2eecce57fe43998 (wasm://wasm/0294a55a:wasm-function[823]:0x2c9b45)
    at ruffle_web::Ruffle::tick::h0965383f85ea7d5c (wasm://wasm/0294a55a:wasm-function[672]:0x23d25c)
```

# Player Info
Allows script access: true
Player type: Object
SWF URL: main.swf
Param movie: main.swf
Attribute 0: undefined
Attribute 1: undefined
Attribute 2: undefined
Attribute 3: undefined

# Page Info
Page URL: https://kdata1.com/2016/02/kof-vs-dnf/
[...]

PR #12823, opened 2023-08-18
core: Don't special-case id 0 in SymbolClass-es

Lord-McSweeney:
This gets #9917 (and probably other SWFs using similar obfuscation) working, after #11005.

I still need to add tests.


Lord-McSweeney:
Superseded by #17275

PR #12831, opened 2023-08-18
fix(deps): update rust dependencies

torokati44:
Tale as old as time. This is https://github.com/ruffle-rs/ruffle/pull/12830 but not `wgpu` or `naga`.


torokati44:
Not needed now after #12832 and #12830.

PR #12835, opened 2023-08-18
web: Make open-in-new-tab open the local player directly

Dinnerbone:
This also disables the URL takeover all of SWF URLs


danielhjacobs:
Note: Since this switches Firefox to declarativeNetRequest, it increases the minimum supported Firefox version to 113: https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/113#changes_for_add-on_developers. As I said on Discord, that's pretty fine since there's an ESR out for 115 and Firefox can be easily updated.

Dinnerbone:
No longer uses declarativeNetRequest at all!

PR #12836, opened 2023-08-18
extension: Make the player page a local Ruffle demo

danielhjacobs:


web/packages/extension/assets/player.html
```diff
@@ -4,12 +4,86 @@
     <head>
         <meta charset="utf-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />
-        <title>Ruffle</title>
+        <title>Ruffle Local Demo</title>
```
Dinnerbone, line 7: I think we should call it "Player" instead of "Demo"

Dinnerbone (approved):

PR #12847, opened 2023-08-19
avm2: Various correctness improvements to Number, int, and uint

Lord-McSweeney:
Unfortunately this regresses one test because namespace versioning is not yet implemented.

### DIFF
--- a/core/src/avm2/globals/int.rs
+++ b/core/src/avm2/globals/int.rs
@@ -138,80 +138,13 @@ fn class_init<'gc>(
 }
 
 /// Implements `int.toExponential`
-fn to_exponential<'gc>(
-    activation: &mut Activation<'_, 'gc>,
-    this: Object<'gc>,
-    args: &[Value<'gc>],
-) -> Result<Value<'gc>, Error<'gc>> {
-    let number = Value::from(this).coerce_to_number(activation)?;
-
-    let digits = args
-        .get(0)
-        .cloned()
-        .unwrap_or(Value::Integer(0))
-        .coerce_to_i32(activation)?;
-
-    if digits < 0 || digits > 20 {
-        return Err(make_error_1002(activation));
-    }
-
-    let digits = digits as usize;
-
-    Ok(AvmString::new_utf8(
-        activation.context.gc_context,
-        format!("{number:.digits$e}")
-            .replace('e', "e+")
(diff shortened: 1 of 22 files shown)


sleepycatcoding (approved):

core/src/avm2/globals/int.rs
```diff
@@ -260,17 +206,7 @@ pub fn create_class<'gc>(activation: &mut Activation<'_, 'gc>) -> GcCell<'gc, Cl
     let class = Class::new(
         QName::new(activation.avm2().public_namespace, "int"),
         Some(Multiname::new(activation.avm2().public_namespace, "Object")),
-        Method::from_builtin_and_params(
-            instance_init,
-            "<int instance initializer>",
-            vec![ParamConfig {
```
adrian17, line 266: why did you drop the params? I can see them in `describeType()`:
```
    <constructor>
      <parameter index="1" type="*" optional="true"/>
    </constructor>
```

core/src/avm2/globals/int.rs
```diff
@@ -118,140 +130,74 @@ fn class_init<'gc>(
...
+    let number = if let Some(this) = this.as_primitive() {
+        match *this {
+            Value::Integer(o) => o,
+            _ => return Err(make_error_1004(activation, "int.prototype.toString")),
         }
+    } else {
+        return Err(make_error_1004(activation, "int.prototype.toString"));
+    };
+
+    let radix = args
```
adrian17, line 169: technically `toString` is also shared with `Number`, right? The only difference being in prototype on `to(Locale)String` checking the proto and throwing different errors. But if you don't think the after-type-checking part can be easily unified, feel free to ignore this, no pressure. Maybe just a comment?

adrian17 (approved):

core/src/avm2/globals/int.rs
Lord-McSweeney: I think it's better to keep it this way- the errors are different, and `(u)int` have an additional check in those methods- if `this == prototype`, return "0".

Lord-McSweeney: I think it's better to keep it this way; converting it would be more annoying. Additionally, we're getting close to having the primitive classes written in AS, too, so it will probably be removed soon-ish.

PR #12853, opened 2023-08-20
tests: Exclude WARP from running the acid_gradient_2 visual test (ported from Shumway)

torokati44:
This was the culprit for most of our recent Windows CI failures.

### DIFF
--- a/tests/tests/swfs/visual/shumway_acid_tests/acid_gradient_2/test.toml
+++ b/tests/tests/swfs/visual/shumway_acid_tests/acid_gradient_2/test.toml
@@ -6,4 +6,4 @@ num_frames = 1
 tolerance = 1
 
 [player_options]
-with_renderer = { optional = false, sample_count = 1 }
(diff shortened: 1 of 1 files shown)


Dinnerbone (approved):

Dinnerbone:
Should we just stop running visual tests on warp?

sleepycatcoding:
Probably a dumb question, but what is WARP?

Dinnerbone:
It's a software renderer for Direct X.

https://learn.microsoft.com/en-us/windows/win32/direct3darticles/directx-warp#what-is-warp

It's available on Windows systems where there's no dedicated gpu, such as our build runners. So they technically can run the visual tests, but with some issues 😅

sleepycatcoding (approved):

torokati44:
> Should we just stop running visual tests on warp?

I think with this `exclude_warp` feature, we'll manage just fine for now.

PR #12873, opened 2023-08-21
extension: Fix-ups for extension player page

danielhjacobs:
* Make pressing the Enter key after pasting/typing in an SWF URL load it.
* Automatically populate the SWF URL field with the current SWF URL if the player page is loaded directly or from clicking the "open in a new tab" button.
* Be consistent about how we space the buttons in the extension popup.
* If the hash is changed in the URL bar on the player page, load the SWF from the new hash.


web/packages/extension/src/player.ts
```diff
@@ -241,28 +243,32 @@ async function loadSwf(swfUrl: string) {
...
-            if ((webURL?.value || "") !== "") {
-                loadSwf(webURL.value);
-                window.location.href = "#" + webURL.value;
-            }
-        });
-    }
+    webFormSubmit.addEventListener("click", function () {
+        if (webURL.value !== "") {
+            loadSwf(webURL.value);
+            window.location.href = "#" + webURL.value;
```
Dinnerbone, line 268: I don't know if there's any strict difference between them, but `window.location.hash = webURL.value` does the same thing a bit more directly.
Not a big deal though!

danielhjacobs: Will change to `hash`. It also occurs to me, the line before this, `loadSwf(webURL.value);`, shouldn't be necessary with this firing the `hashchange` event which already does that.

danielhjacobs: The exceptions is that now if you don't change the URL before clicking "Load", the animation won't reload, but it's probably better to rely on the reload icon for reloading the swf anyway, as that gives a warning confirmation.

Dinnerbone (approved):

web/packages/extension/src/player.ts
```diff
@@ -241,28 +243,31 @@ async function loadSwf(swfUrl: string) {
...
+        webURL.value = swfUrl;
         await loadSwf(swfUrl);
     }
+}
+
+window.addEventListener("pageshow", async () => {
+    await loadSwfFromHash();
+});
+
+window.addEventListener("hashchange", async () => {
```
sleepycatcoding, line 260: Is it possible to pass the function in directly, or is the lambda required?

danielhjacobs: Opened #12904.

PR #12876, opened 2023-08-21
extension: Fix check for scripting permission

danielhjacobs:
The content script doesn't have access to the `scripting` API, but it does have the ability to get the extension's manifest: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Content_scripts#webextension_apis


Dinnerbone:
Didn't test it but makes sense!

Dinnerbone (approved):

PR #12888, opened 2023-08-22
avm2: Throw TypeError for null argument in TextField string property setters

n0samu:
I suspect this is an issue throughout the codebase, but I'm fixing it where I found it.

### DIFF
--- a/core/src/avm2/globals/flash/text/text_field.rs
+++ b/core/src/avm2/globals/flash/text/text_field.rs
@@ -95,7 +95,7 @@ pub fn set_auto_size<'gc>(
         .as_display_object()
         .and_then(|this| this.as_edit_text())
     {
-        let value = args.get_string(activation, 0)?;
+        let value = args.get_string_non_null(activation, 0, "autoSize")?;
         this.set_autosize(
             if &value == b"left" {
                 AutoSizeMode::Left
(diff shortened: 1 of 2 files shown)


core/src/avm2/globals/flash/text/text_field.rs
```diff
@@ -751,6 +766,9 @@ pub fn replace_text<'gc>(
         .as_display_object()
         .and_then(|this| this.as_edit_text())
     {
+        if matches!(args.get_value(0), Value::Null) {
+            return Err(make_error_2007(activation, "text"));
```
Lord-McSweeney, line 770: This should check `args.get_value(2)`

core/src/avm2/error.rs
```diff
@@ -166,6 +166,20 @@ pub fn make_error_1506<'gc>(activation: &mut Activation<'_, 'gc>) -> Error<'gc>
     }
 }
 
+#[inline(never)]
+#[cold]
+pub fn make_error_2007<'gc>(activation: &mut Activation<'_, 'gc>, param_name: &str) -> Error<'gc> {
+    let err = argument_error(
```
Lord-McSweeney, line 172: This should be a `type_error`

core/src/avm2/globals/flash/text/text_field.rs
n0samu: Good suggestion, thanks! I've done that

adrian17 (approved):

PR #12915, opened 2023-08-23
extension: Register listeners synchronously at top level of service worker

n0samu:
[From this blog post](https://adguard-vpn.com/en/blog/mv3-browser-extension.html):
> In Manifest V2 environment, it didn't matter where in the code the event listener was located – its registration could occur asynchronously because the background page was always running and never restarted.
> 
> In Manifest V3 the service worker is restarted when an event occurs. In this case, in order for the event to hit the listener, it must be registered at the top level of the code (synchronously).

This fix is needed for the "Open in new tab" button to work even when the service worker has been suspended (which happens after 30 seconds of inactivity)

### DIFF
--- a/web/packages/extension/src/background.ts
+++ b/web/packages/extension/src/background.ts
@@ -1,8 +1,15 @@
 import * as utils from "./utils";
 import { isMessage } from "./messages";
 
+async function contentScriptRegistered() {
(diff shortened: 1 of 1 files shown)


sleepycatcoding (approved):

sleepycatcoding:
lgtm

Issue #7915, opened 2022-09-08
Error on https://redesurftank.com.br


bobaoapae:
#Error Info
Error dispatching event EventObject(EventObject { type: "complete", class: flash.events::Event, ptr: 0x3ab498 }) to handler FunctionObject(FunctionObject(GcCell(Gc { ptr: 0x423694 }))) : "Cannot get undefined property \"close\" on class Some(ClassObject { name: flash.net::URLLoader, ptr: 0x409598 })"

# Player Info
Allows script access: true
Renderer: WebGL

# Page Info
Page URL: https://redesurftank.com.br/jogar&srv=surftank
SWF URL: https://redesurftank.com.br/loading-v29/Loading.swf?user=some&key=someKey&isGuest=&v=10950&rand=5079630&config=/config.php&area_id=1001
Aditional: Need create some random account to test and access the game

# Browser Info
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/105.0.0.0 Safari/537.36 Edg/105.0.1343.27
Platform: Win32
Has touch support: false

# Ruffle Info
Version: 0.1.0
Name: nightly 2022-09-08
Channel: nightly
Built: 2022-09-08T00:16:33.182Z
Commit: ece4b9a76dd825a1664f06c22622f1066cc69011
Is extension: true

# Metadata
width: 1000
height: 600
frameRate: 25
numFrames: 1
swfVersion: 28
backgroundColor: #000000
isActionScript3: true
uncompressedLength: 664832

Lord-McSweeney:
This currently needs `flash.globalization::DateTimeFormatter`.

bobaoapae:
The currently error changed, I removed the timeformatter for test. 

This is the error I have now. 

 Error dispatching event EventObject(EventObject { type: "timer", class: flash.events::TimerEvent, ptr: 0x933d80c }) to handler FunctionObject(FunctionObject { ptr: 0xae5c914 }) : RustError("Cannot set an XML/XMLList object XmlListObject(XmlListObject { ptr: 0x9b1106c }) as an attribute")

n0samu:
With #12941, the game loads most of the way, but gets stuck on loading game configuration (39%) with this error:
```
[...]

PR #12941, opened 2023-08-25
avm2: XML/XMLList is converted to a string when set as attribute value

sleepycatcoding:
Weirdly enough Flash converts the XML/XMLList object to a XML string when set as an attribute value.


Dinnerbone (approved):

n0samu (approved):
Progresses #7915. 👍

PR #12946, opened 2023-08-25
render: Add DrawCommand::CubicCurveTo

Dinnerbone:
This is the first DrawCommand that isn't available in the swf format. It's exposed to AS3 through `Graphics.cubicCurveTo` and `Graphics.drawPath`.

To avoid confusion, the existing `CurveTo` has been renamed to `QuadraticCurveTo`.

This is a hard dependency for device rendering support - a lot of fonts use cubic curves, not just quadratic curves.

For hit testing I just used lyon to generate the bounding box, and to split it to quadratic curves. I think I'd propose we switch all other hit testing to lyon:
- It's likely faster than our implementation, they have a lot more short-circuits and fast-approximations than we do. Not tested to verify this though.
- They have a *lot* more tests than we do :D


render/src/shape_utils.rs
```diff
@@ -676,10 +682,31 @@ pub fn draw_command_fill_hit_test(commands: &[DrawCommand], test_point: swf::Poi
...
+                control_b,
+                anchor,
+            } => {
+                lyon_geom::CubicBezierSegment {
+                    from: lyon_geom::Point::new(cursor.x.to_pixels(), cursor.y.to_pixels()),
+                    ctrl1: lyon_geom::Point::new(control_a.x.to_pixels(), control_a.y.to_pixels()),
+                    ctrl2: lyon_geom::Point::new(control_b.x.to_pixels(), control_b.y.to_pixels()),
+                    to: lyon_geom::Point::new(anchor.x.to_pixels(), anchor.y.to_pixels()),
+                }
+                .for_each_quadratic_bezier(0.01, &mut |quadratic_curve| {
```
moulins, line 700: Nit: the constant `0.01` is used twice, so it should probably be given a name.

render/src/shape_utils.rs
```diff
@@ -713,12 +740,41 @@ pub fn draw_command_stroke_hit_test(
...
+            } => {
+                let mut hit = false;
+                lyon_geom::CubicBezierSegment {
+                    from: lyon_geom::Point::new(cursor.x.to_pixels(), cursor.y.to_pixels()),
+                    ctrl1: lyon_geom::Point::new(control_a.x.to_pixels(), control_a.y.to_pixels()),
+                    ctrl2: lyon_geom::Point::new(control_b.x.to_pixels(), control_b.y.to_pixels()),
+                    to: lyon_geom::Point::new(anchor.x.to_pixels(), anchor.y.to_pixels()),
+                }
+                .for_each_quadratic_bezier(0.01, &mut |quadratic_curve| {
+                    if hit_test_stroke_curve(
```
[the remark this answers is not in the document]
Dinnerbone: Good idea, done!

render/src/shape_utils.rs
Dinnerbone: Done!

render/src/shape_utils.rs
```diff
@@ -1222,6 +1295,31 @@ pub fn quadratic_curve_bounds(
...
+        to: lyon_geom::Point::new(anchor.x.to_pixels(), anchor.y.to_pixels()),
+    }
+    .bounding_box();
+
+    let radius = stroke_width / 2;
+    Rectangle {
+        x_min: Twips::from_pixels(bounds.min.x) - radius,
+        x_max: Twips::from_pixels(bounds.max.x) + radius,
+        y_min: Twips::from_pixels(bounds.min.y) - radius,
+        y_max: Twips::from_pixels(bounds.max.y) + radius,
```
torokati44, line 1319: it's somewhat strange that `Rectangle<Coordinate>` doesn't have a `grow` or `expand` method for this, but alas, it doesn't.

Dinnerbone: We have so many pseudo math/geom types that all have different capabilities. I'd love to just make a ruffle-geom crate and get it over with... One day 😅

render/src/shape_utils.rs
```diff
@@ -713,12 +746,52 @@ pub fn draw_command_stroke_hit_test(
...
+                lyon_geom::CubicBezierSegment {
+                    from: lyon_geom::Point::new(cursor.x.to_pixels(), cursor.y.to_pixels()),
+                    ctrl1: lyon_geom::Point::new(control_a.x.to_pixels(), control_a.y.to_pixels()),
+                    ctrl2: lyon_geom::Point::new(control_b.x.to_pixels(), control_b.y.to_pixels()),
+                    to: lyon_geom::Point::new(anchor.x.to_pixels(), anchor.y.to_pixels()),
+                }
+                .for_each_quadratic_bezier(
+                    CUBIC_CURVE_TOLERANCE,
+                    &mut |quadratic_curve| {
+                        if !hit
```
torokati44, line 770: A comment here would be nice that this "doesn't make sense", but is for performance, in lieu of an early break from the `for_each_quadratic_bezier` - but this is such a minor nit...

render/src/shape_utils.rs
```diff
@@ -1222,6 +1295,31 @@ pub fn quadratic_curve_bounds(
...
 }
 
+pub fn cubic_curve_bounds(
+    start: swf::Point<Twips>,
+    stroke_width: Twips,
+    control_a: swf::Point<Twips>,
+    control_b: swf::Point<Twips>,
+    anchor: swf::Point<Twips>,
+) -> Rectangle<Twips> {
+    // [NA] Should we just move most of our math in this file to lyon_geom?
```
torokati44, line 1305: Honestly, I can't make up my mind on what to think about this... Many pros and cons on both sides...

render/src/shape_utils.rs
```diff
@@ -3,6 +3,9 @@ use enum_map::Enum;
 use smallvec::SmallVec;
 use swf::{CharacterId, FillStyle, LineStyle, Rectangle, Shape, ShapeRecord, Twips};
 
+/// Controls the accuracy of the approximated quadratic curve, when splitting up a cubic curve
```
torokati44, line 6: again, a minor nit: `... into quadratic ones` ?

Dinnerbone: Feels a little tautological to me 😅

torokati44: Also maybe a unit for this number, or "meaning of the magnitude" would be nice to mention, but not critical.

torokati44: > Feels a little tautological to me 😅
Holy moly, I forgot the middle of the sentence by the time I got to its end... Why am _I_ code reviewing _after midnight_?

torokati44:
Huh, did none of the ported ACID tests get fixed by this?

torokati44 (approved):

torokati44:
Left some review notes, nothing worthy to block the merge, I'll leave it up to you @Dinnerbone, whether you make any further changes based on them.

Dinnerbone:
They don't use it. I haven't actually seen any content online that uses it, outside of blog posts saying "you should use this!"

PR #12955, opened 2023-08-26
core: Various small Font and Glyph refactors

Dinnerbone:
Dependencies for TTF rendering

### DIFF
--- a/core/src/font.rs
+++ b/core/src/font.rs
@@ -127,6 +127,7 @@ impl<'gc> Font<'gc> {
                 let glyph = Glyph {
                     shape_handle: None.into(),
                     shape: None.into(),
+                    advance: Twips::new(swf_glyph.advance.into()),
                     swf_glyph,
                 };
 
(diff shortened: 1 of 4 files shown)


core/src/font.rs
```diff
@@ -60,36 +60,94 @@ impl EvalParameters {
...
     scale: f32,
 
-    /// Kerning infomration.
-    /// Maps from a pair of unicode code points to horizontal offset value.
-    kerning_pairs: fnv::FnvHashMap<(u16, u16), Twips>,
-
     /// The distance from the top of each glyph to the baseline of the font, in
     /// EM-square coordinates.
-    ascent: u16,
+    ascent: i16,
```
sleepycatcoding, line 146: Out of interest, why is this change necessary, can the ascent be negative?

core/src/font.rs
```diff
@@ -418,38 +469,78 @@ impl<'gc> Font<'gc> {
...
+        }
+    }
+}
+
 #[derive(Debug, Clone)]
 pub struct Glyph {
     // Handle to registered shape.
     // If None, it'll be loaded lazily on first render of this glyph.
-    shape_handle: RefCell<Option<ShapeHandle>>,
+    // It's a double option; the outer one is "have we registered", the inner one is option because it may not exist
```
sleepycatcoding, line 522: I think this should be doc comment.

sleepycatcoding (approved):
I don't have experience in handling fonts, but this looks good to me.

core/src/font.rs
Dinnerbone: For TTFs yes, apparently

core/src/font.rs
```diff
@@ -519,14 +519,15 @@ impl GlyphShape {
...
+    shape_handle: RefCell<Option<Option<ShapeHandle>>>,
```
adrian17, line 523: to be sure, there currently doesn't exist a case where we put a `None` here - it's for a future PR?

adrian17 (approved):

core/src/font.rs
Dinnerbone: Yeah, font files can define properties of a glyph without any way to display it - they'll be whitespace characters

torokati44: Wouldn't this be nicer as an `enum`?

PR #12967, opened 2023-08-27
avm2: Implement namespace versioning

Aaron1011:
This allows us to hide things like `MovieClip.isPlaying` from older SWFs, which may define an `isPlaying` method without marking it as an override.

This has the largest impact on public namespaces. There is no longer just one public namespace - we now have a public namespace per API version. Most callsites should use `Avm2.find_public_namespace()`, which determines the public namespace based on the root SWF API version. This ensures that explicit property access in Ruffle (e.g. calling "toString") are able to see things defined in the user SWF.

This requires several changes other:

* A new `ApiVersion` enum is introduced, storing all of the api versions defined in avmplus
* NamespaceData::Namespace now holds an `ApiVersion`. When we read a namespace from a user-defined ABC file, we use the `ApiVersion` from the root movie clip. This matches Flash Player's behavior - when a child SWF is loaded through `Loader`, its API version gets overwritten with the parent's API version, which affects definition visibility and 'override' requirements in the child SWF. Note that 'behavior changes' in methods like `gotoAndPlay` are unaffected.
* The `PartialEq` impl for `Namespace` has been removed - each call site must now choose to compare namespaces either by an exact version match, or by 'compatible' versions (a `<=` check) with `Namespace::matches_ns`
* `PropertyMap` now uses `Namespace::matches_ns` to check for a compatible version when looking up a definition.
* When compiling our playerglobal, we pass in the `-apiversioning` flag, which causes asc.jar to convert `[API]` metadata into a special Unicode 'version mark' suffix in namespace URLs. We parse this when loading namespaces to determine the API version to use for playerglobal definitions (unmarked definitions use the lowest possible version).

Unfortunately, this makes ffdec unable to decompile our playerglobal.swc - I've opened [an upstream issue](https://www.free-decompiler.com/flash/issues/2066-nullpointerexception-in-scriptinfo-getpacks-when-attempting-to-load-api-versioned-swf)


core/src/avm2/api_version.rs
```diff
@@ -0,0 +1,63 @@
+use gc_arena::Collect;
+
+// Based on https://github.com/adobe/avmplus/blob/master/core/api-versions.h
+// FIXME - if we ever add in AIR emulation, then implement the 'version series' logic.
```
Molisson, line 4: If you consider adding that, would Ruffle devs have to support Adobe AIR only features too?

Molisson (approved):
Looks fine to me, this is my first time reviewing the changes though I couldn't review all of them and couldn't find any mistakes since I'm unfamiliar with Rust.

core/src/avm2.rs
```diff
@@ -110,7 +114,12 @@ pub struct Avm2<'gc> {
     /// However, it's not strictly defined which items end up there.
     toplevel_global_object: Option<Object<'gc>>,
 
-    pub public_namespace: Namespace<'gc>,
+    /// The public namespace, versioned with `ApiVersion::ALL_VERSIONS`.
+    /// When calling into user code, you should almost always use `find_public_namespace`
+    /// instead, as it will return the correct version for the current call stack.
+    public_namespace_base_version: Namespace<'gc>,
+    /// Cached public namespaces for each `ApiVersion`
+    public_namespaces: FnvHashMap<ApiVersion, Namespace<'gc>>,
```
adrian17, line 122: Feels like we'd prefer a Vec<Option> here to a hashmap?

core/src/avm2.rs
```diff
@@ -628,6 +665,13 @@ impl<'gc> Avm2<'gc> {
 
     #[cfg(not(feature = "avm_debug"))]
     pub const fn set_show_debug_output(&self, _visible: bool) {}
+
+    /// Gets the public namespace, versioned based on the current root SWF.
+    /// See `AvmCore::findPublicNamespace()`
+    /// https://github.com/adobe/avmplus/blob/858d034a3bd3a54d9b70909386435cf4aec81d21/core/AvmCore.cpp#L5809C25-L5809C25
+    pub fn find_public_namespace(&self) -> Namespace<'gc> {
+        self.public_namespaces[&self.root_api_version]
```
adrian17, line 673: IIRC avmplus doesn't do this, but just making sure intuitively: what stops us from storing the one versioned namespace directly (next to `root_api_version`) instead of looking them up dynamically?

core/src/avm2/api_version.rs
```diff
@@ -0,0 +1,63 @@
...
+// For now, it's fine to just compare Flash player version against air versions directly.
+#[derive(Copy, Clone, Collect, PartialEq, Eq, PartialOrd, Ord, Hash, FromPrimitive, Debug)]
+#[collect(require_static)]
```
adrian17, line 7: If we can, let's put `#[collect(require_static)]` on struct fields instead of making trivial types GC-aware.

core/src/avm2/globals/README.md
```diff
@@ -63,6 +63,28 @@ class is loaded.
...
+To add versioning to an API:
+
+1. Determine the first version where it was added. This can be seen in the Flash Documentation (e.g. "Runtime Versions: Flash Player 11.4, AIR 3.4")
+2. Convert the Flash Player version to an SWF version number using [this chart](http://www.adobe.com/devnet/articles/flashplayer-air-feature-list.html)
+2. Determine the corresponding asc.jar version code for the SWF version. This can be found in avmplus in https://github.com/adobe/avmplus/blob/master/core/api-versions.as
+3. Add an `[API("VersionCode")]` metadata to the defintion. In the `Event.WORKER_STATE` example,
+   this looks like:
+
+   ```actionscript
+   [API("682")]
```
adrian17, line 81: not in this PR, but: do you think there's any point in making this less magical, via some preprocessing step maybe?

adrian17 (commented):
(Not a full review yet, just a quick look over the code)

core/src/avm2/api_version.rs
Aaron1011: It's the other way around - if we want to implement any AIR-only APIS, then we may need to add full support for AIR api-versioning.

core/src/avm2.rs
Aaron1011: We have multiple versioned namespaces in use (almost all items in playerglobals will have the lowest version, but specifically tagged `[API]` definitions will have different versions.

core/src/avm2/globals/README.md
Aaron1011: Maybe - I don't know if we'll actually need to implement enough versioned APIs for it to be worth it.

core/src/avm2/globals/README.md
```diff
@@ -63,6 +63,28 @@ class is loaded.
...
+## API Versioning
+
+Ruffle supports Flash's Api versioning, which hides newer playerglobal definitions
+(including methods/properties) from SWFs compiled with older API versions.
+For example, see `Event.WORKER_STATE`
+
+To add versioning to an API:
+
+1. Determine the first version where it was added. This can be seen in the Flash Documentation (e.g. "Runtime Versions: Flash Player 11.4, AIR 3.4")
+2. Convert the Flash Player version to an SWF version number using [this chart](http://www.adobe.com/devnet/articles/flashplayer-air-feature-list.html)
```
danielhjacobs, line 75: Dead link

core/src/avm2/globals/README.md
```diff
@@ -63,6 +63,28 @@ class is loaded.
...
+2. Determine the corresponding asc.jar version code for the SWF version. This can be found in avmplus in https://github.com/adobe/avmplus/blob/master/core/api-versions.as
```
Lord-McSweeney, line 76: It might be useful to note that this chart can also version based on FP version

Aaron1011: I'm not really sure what you mean

Lord-McSweeney: https://github.com/adobe/avmplus/blob/master/core/api-versions.as also has versioning based on FP minor version
(e.g. `CONFIG const FP_10_1 = 667;`)

Aaron1011: That's only for a few very old versions

Lord-McSweeney:
Can you also version-gate `rotationZ`? #13263 seems to need it.

tests/tests/swfs/avm2/cross_api_version_call_older/older/com/ruffle/Older.as
```diff
@@ -0,0 +1,28 @@
+﻿package com.ruffle {
+	import flash.display.MovieClip;
+	import flash.events.Event;
+
+	public class Older extends MovieClip {
+		public function Older() {
+			trace("Initialized Older with parent: " + this.parent + " Event = " + Event["WORKER_STATE"]);
+            // FIXME - enable this when our error messages match Flash Player
```
Lord-McSweeney, line 8: I think this should work correctly now

Aaron1011:
@Lord-McSweeney This pr is already pretty large, so I'd like to save adding more version-gating for follow-up prs.

Lord-McSweeney:
Does this implementation support version-gated classes? Some classes such as `Shader` are version-gated (e.g. Shader is only available in SWF >9), though I'm not sure what would actually rely on this.

Aaron1011:
Yes - all namespaces are now versioned, including those used in class QNames

core/src/avm2/namespace.rs
```diff
@@ -61,29 +84,79 @@ impl<'gc> Namespace<'gc> {
...
+                .is_playerglobals_domain(context.avm2);
+
+            let mut api_version = ApiVersion::AllVersions;
+            let stripped = strip_version_mark(namespace_name.as_wstr());
+            let has_version_mark = stripped.is_some();
+            if let Some((stripped, version)) = stripped {
+                assert!(
+                    is_playerglobals,
+                    "Found versioned url {namespace_name:?} in non-playerglobals domain"
+                );
```
Lord-McSweeney, line 129: Can't normal SWFs add a namespace with a version marker? If so, we should warn here, and ignore it (unless FP has a similar `assert!`, or FP actually obeys version markers in non-playerglobals SWFs)

Aaron1011: That should never happen, so I'd like to get an issue report if any real swf tries to include a version marker

Lord-McSweeney (approved):

core/src/avm2/api_version.rs
```diff
@@ -0,0 +1,60 @@
...
+    SWF_27 = 42,
+    AIR_16_0 = 43,
+    SWF_28 = 44,
+    AIR_17_0 = 45,
+    SWF_29 = 46,
+    AIR_18_0 = 47,
+    SWF_30 = 48,
+    AIR_19_0 = 49,
+    SWF_31 = 50,
+    AIR_20_0 = 51,
```
danielhjacobs, line 58: Re-stating what I said on Discord, this is probably not the highest ApiVersion we'll ever need to gate. https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/media/AudioDeviceManager.html was added in Flash Player 27, which according to https://github.com/ruffle-rs/ruffle/wiki/SWF-version-chart, is SWF version 38.

Lord-McSweeney: FP doesn't actually implement anything higher than that- see [the avmplus header that this file references](https://github.com/adobe/avmplus/blob/master/core/api-versions.h). I'm not sure that FP can actually gate anything higher than that- it might be faking it. It's possible that it uses 
SWF version to version-gate `AudioDeviceManager` and doesn't version-gate it based on AIR version.

core/src/avm2/namespace.rs
```diff
@@ -155,4 +232,49 @@ impl<'gc> Namespace<'gc> {
...
+        }
+        match (&*self.0, &*other.0) {
+            (
+                NamespaceData::Namespace(name1, version1),
+                NamespaceData::Namespace(name2, version2),
+            ) => {
+                let name_matches = name1 == name2;
+                let version_matches = version1 <= version2;
+                if name_matches && !version_matches {
+                    tracing::info!(
```
Lord-McSweeney, line 269: This `info!` feels like it's going to get triggered a lot- maybe an `avm_debug`?

core/src/avm2/namespace.rs
```diff
@@ -155,4 +232,49 @@ impl<'gc> Namespace<'gc> {
...
+        } else if self.is_private() || other.is_private() {
+            false
+        } else {
+            *self.0 == *other.0
+        }
+    }
+
+    /// Compares this namespace to another, considering them equal if this namespace's version
+    /// is less than or equal to the other (definitions in this namespace version can be
+    /// seen by the other). This is used to implement `ProperyMap`, where we want to
```
Lord-McSweeney, line 254: nit: typo: `ProperyMap` --> `PropertyMap`

Lord-McSweeney:
This isn't really important, but it would be nice to have the namespace version show up in debug tools.

core/src/avm2/namespace.rs
Aaron1011: This should be rare - it only logs when namespace versioning is actually causing something to be hidden, which should be very rare in practice

core/src/player.rs
```diff
@@ -373,6 +375,8 @@ impl Player {
         self.instance_counter = 0;
 
         self.mutate_with_update_context(|context| {
+            context.avm2.root_api_version = ApiVersion::from_u8(context.swf.version())
```
adrian17, line 378: This is not correct, right? We want SWF v23 to map to `SWF_23`, not `AIR_3_4`.

Aaron1011: I've fixed this

Lord-McSweeney:
If playerglobals code attempts to use version-gated methods (e.g. #13834 uses `Array.insertAt`), will it work with this PR?

PR #12977, opened 2023-08-27
naga-pixelbender: Stub out Opcode::Loop

Aaron1011:
Some experimentation with Pixel Bender Studio shows that Opcode::Loop has a 23-byte payload. I haven't tried to figure out how to interpet the payload yet, but we can now skip over the opcode instead of bailing out entirely.

### DIFF
--- a/render/src/pixel_bender.rs
+++ b/render/src/pixel_bender.rs
@@ -244,6 +244,9 @@ pub enum Operation {
     },
     Else,
     EndIf,
+    Loop {
+        unknown: Box<[u8]>,
+    },
 }
 
 #[derive(Debug, Clone)]
@@ -563,6 +566,13 @@ fn read_op<R: Read>(
                 }),
                 _ => unreachable!(),
             }
+        }
+        Opcode::Loop => {
+            let mut unknown = vec![0u8; 23];
+            data.read_exact(&mut unknown)?;
+            shader.operations.push(Operation::Loop {
+                unknown: unknown.into(),
+            });
         }
         _ => {
             let dst = data.read_u16::<LittleEndian>()?;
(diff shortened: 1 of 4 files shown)


render/src/pixel_bender.rs
```diff
@@ -564,6 +567,11 @@ fn read_op<R: Read>(
                 _ => unreachable!(),
             }
         }
+        Opcode::Loop => {
+            let mut unknown = vec![0u8; 23];
```
SuchAFuriousDeath, line 571: Am I missing something, or this doesn't have to be a Vec?

Aaron1011: I store a `Vec` in the `Operation::Loop` enum variant, to make it cheaper to move.

SuchAFuriousDeath: Cheaper to move than what? It feels like this could just be a slice

Aaron1011: Storing an array in `Operation::Loop` will make the entire enum larger. Since we're using a vec, there's no reason to start out with an array in this method and then copy it to a vec.

SuchAFuriousDeath: Doesn't a Vec have a size of 24 bytes? An array of 23 u8s should by 23 bytes, if I'm not mistaken

SuchAFuriousDeath: Forgive me if I'm missing something

Aaron1011: On 32-bit platforms like wasm, it's 12 bytes.

SuchAFuriousDeath: Right, my apologies then. But wouldn't a boxed array still save space?

Aaron1011: Good catch - I've changed it to `Box<[u8]>`

adrian17 (approved):

Issue #12613, opened 2023-08-07
File Browser Won’t Open in s2online.github.io


FlamedDogo99:
### Describe the bug

s2online.github.io contains a copy of the original scratchflash editor and player. Loading the website with ruffle causes no errors, however many features are broken, such as uploading a file or moving blocks in the editor.
Here is the swf file: https://github.com/FlamedDogo99/FlamedDogo99.github.io/blob/main/Scratch.swf

### Expected behavior

Expected behavior:
![Showing file picking menu](https://github.com/ruffle-rs/ruffle/assets/96555444/3c2006e5-ab8c-4cf0-8f2c-63f75f1f99e9)
Current behavior: (no prompt is displayed)
![Not showing file picking menu](https://github.com/ruffle-rs/ruffle/assets/96555444/66519324-d112-4663-a3de-e90f52deb441)

### Affected platform

Browser's extension

### Operating system

MacOS Ventura

### Browser

Brave 1.56.14

### Additional information

I’m assuming this is just due to AMV2 features that haven’t been implemented yet, but I thought it wouldn’t hurt to report this.

Lord-McSweeney:
Continuing from #9058 (original Scratch 2 issue). A list of issues:
`XML.setName` unimplemented, causing issues with drawing sprites
`DisplayObject.getObjectsUnderPoint` (#11970) unimplemented, causing the block dragging issues
`FileReference.browse` unimplemented in AVM2, causing the file browsing issue

FlamedDogo99:
Shoot didn't realize this was a duplicate, thanks!

Lord-McSweeney:
Currently this mostly works but starts up slowly due to `by.blooddy.crypto.image::PNG24Encoder.encode` taking a long time to run.

FlamedDogo99:
Also downloading just doesn't work, although I don't know if that was being addressed with this originally

FlamedDogo99:
Everything appears to be functional now.

PR #12979, opened 2023-08-27
avm2: Implement XML.setName

Lord-McSweeney:
Allows sprite costumes to be edited in Scratch 2 (progresses #12613).


Dinnerbone:
Just to be clear; is this intended as a stub or full implementation? I'm not sure it handles qnames or processing instructions?

Lord-McSweeney:
I didn't realize you could `setName` with a QName as the argument- I'll implement that now.

Dinnerbone:
The implementation should follow 13.4.4.35 from this PDF: https://www.ecma-international.org/wp-content/uploads/ECMA-357_2nd_edition_december_2005.pdf (or search for `setName`)

Lord-McSweeney:
I added support for QNames (but no namespaces).

sleepycatcoding (approved):
Looks good to me

tests/tests/swfs/avm2/xml_set_name/test.toml
```diff
@@ -0,0 +1 @@
+num_frames = 1
```
sleepycatcoding, line 1: A small nit, but I think according to the README.md in tests folder, the use of `num_frames` was deprecated and `num_ticks` is meant to be instead?

core/src/avm2/globals/xml.rs
```diff
@@ -81,6 +81,42 @@ pub fn name<'gc>(
...
 
+pub fn set_name<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Object<'gc>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let xml = this.as_xml_object().unwrap();
+    let node = xml.node();
+
+    let new_name = args.get_value(0);
```
adrian17, line 92: now that we have `isXMLName`, we should use it here and throw an exception if the name is illegal, right? (we can do this in a follow-up if you prefer)

adrian17 (approved):

core/src/avm2/globals/xml.rs
Lord-McSweeney: I would prefer to do it in this PR.

Issue #11634, opened 2023-06-19
Warzone Tower Defense: Cannot start game


n0samu:
### Describe the bug

SWF URL: https://www.andkon.com/arcade/missiledefender/warzonetowerdefense/warzonetowerdefense.swf
(Note: This is a multiasset game, so it needs to be loaded via its URL or all assets need to be downloaded)

The main menu buttons of [Warzone Tower Defense](https://www.andkon.com/arcade/missiledefender/warzonetowerdefense/) are all unresponsive, making it impossible to start the game. And the level name and preview image are not shown.

The errors/warnings are:
```
WARN core/src/stub.rs:126 Encountered stub: AVM2 XML.settings()
WARN core/src/stub.rs:126 Encountered stub: AVM2 XML.setSettings()
ERROR core/src/avm2/events.rs:419 Error dispatching event EventObject(EventObject { type: "complete", class: flash.events::Event, ptr: 0x2eb4a94 }) to handler FunctionObject(FunctionObject { ptr: 0x2e775e4 }) : RustError("ToXMLString: Not yet implemented node CData(\"Field of Broken Dreams\")")
```

### Expected behavior

Main menu buttons should be clickable and the level name & preview should be shown on the menu.

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Edge 114

### Additional information

This was originally reported by a user on Discord.

evilpie:
@Lord-McSweeney said in #11724 that this needs `XMLNode.nodeValue`.

I tried stubbing the getter, but as far as I can tell its not actually being called?

n0samu:
@evilpie This is a multiasset game so it needs to be loaded via its URL (don't try to download and open just the loader SWF). Here is the error:
```
ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "complete", class: flash.events::Event, ptr: 0x19c8486aac8 }) to handler FunctionObject(FunctionObject { ptr: 0x19c84db0198 }) : ReferenceError: Error #1069: Property nodeValue not found on flash.xml.XMLNode and there is no default value.
        at com.jgp.xml::XMLFetch/addNodeList()
        at com.jgp.xml::XMLFetch/addNodeList()
        at com.jgp.xml::XMLFetch/addNodeList()
        at com.jgp.xml::XMLFetch/addNodeList()
        at com.jgp.xml::XMLFetch/addNodeList()
        at com.jgp.xml::XMLFetch/parseDocument()
[...]

PR #12980, opened 2023-08-27
avm2: Fix XML/XMLDocument ignoreWhite handling for cdata

Aaron1011:
The ignoreWhite settings only affect text nodes, not cdata nodes. Additionally, there should be no special case for empty text.


Aaron1011:
Fixes https://github.com/ruffle-rs/ruffle/issues/11634

adrian17 (approved):

Issue #11630, opened 2023-06-19
Error on https://arcadespot.com/game/nambers/


n0samu:
# Error Info
Error name: Error
Error message: panicked at 'Tex storage 2D multisample is not supported', /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/glow-0.12.2/src/web_sys.rs:3141:9
Error stack:
```
Error: panicked at 'Tex storage 2D multisample is not supported', /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/glow-0.12.2/src/web_sys.rs:3141:9
    at imports.wbg.__wbg_new_d258248ed531ff54 (chrome-extension://donbcfbmhbcapadipfkeojnmajbakjdc/dist/339.js:2396:21)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::hcb6f72ebd1a9c8bd (wasm://wasm/02c6cd7e:wasm-function[8635]:0x83f65f)
    at std::panicking::rust_panic_with_hook::h85b7f6628c291e12 (wasm://wasm/02c6cd7e:wasm-function[5602]:0x767517)
    at core::panicking::panic_fmt::hf5c4cd929d4aaa9e (wasm://wasm/02c6cd7e:wasm-function[8499]:0x83b269)
    at <T as wgpu::context::DynContext>::device_create_texture::h2eb6e848288d8733 (wasm://wasm/02c6cd7e:wasm-function[6323]:0x79b06a)
    at wgpu::Device::create_texture::ha235fa98dfb3ff4e (wasm://wasm/02c6cd7e:wasm-function[4408]:0x702d1e)
    at <ruffle_render_wgpu::context3d::WgpuContext3D as ruffle_render::backend::Context3D>::process_command::h1e8733e3f3231cfb (wasm://wasm/02c6cd7e:wasm-function[531]:0xc2a67)
    at ruffle_core::avm2::globals::flash::display3D::context_3d::set_render_to_texture::had439fd829536164 (wasm://wasm/02c6cd7e:wasm-function[1539]:0x451401)
    at ruffle_core::avm2::function::Executable::exec::h3325d83591e2f464 (wasm://wasm/02c6cd7e:wasm-function[811]:0x2ae9e1)
    at ruffle_core::avm2::object::TObject::call_property::h03891e9d704e019e (wasm://wasm/02c6cd7e:wasm-function[1142]:0x3a5f81)
```
AVM2 stack:
```
[...]

PR #12982, opened 2023-08-27
wgpu: Disable Context3D.setRenderToTexture MSAA on web

Aaron1011:
This is not supported under webgl, and results in a panic "Tex storage 2D multisample is not supported" if we try to do it.

### DIFF
--- a/render/wgpu/src/context3d/mod.rs
+++ b/render/wgpu/src/context3d/mod.rs
@@ -751,6 +751,18 @@ impl Context3D for WgpuContext3D {
                 if sample_count == 0 {
                     sample_count = 1;
                 }
+                #[cfg(target_family = "wasm")]
+                {
+                    if sample_count > 1
+                        && matches!(
+                            self.descriptors.adapter.get_info().backend,
+                            wgpu::Backend::Gl
+                        )
+                    {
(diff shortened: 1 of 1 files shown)


Aaron1011:
Fixes https://github.com/ruffle-rs/ruffle/issues/11630

torokati44:
Sorry for nitpicking, but is there an easy way to test for WebGL directly, instead of for WASM? Keeping the (hopefully) nearing WebGPU in mind...

Dinnerbone:
I agree, cfg isn't the way we should do this.

(alrhoufh in reality we aren't going to be on webgpu until someone can figure out the async situation...)

Aaron1011:
I've adjusted this to check for `Backend::Gl` on wasm (as opposed to `Backend::BrowserWebGpu`)

Lord-McSweeney:
Will this disable it on OpenGL too? Does setRenderToTexture MSAA work on OpenGL? (asking because I often use `-g gl` to avoid unnecessary panics on my lower-end device)

Aaron1011:
No - this is still behind the wasm cfg, so it will only affect webgl

torokati44:
I think the check is fine now. Do you agree @Dinnerbone?

Dinnerbone (approved):

Issue #12354, opened 2023-07-27
Test Failure: avmplus e4x Global suite 


Dinnerbone:
Completely passing suites are not shown.

## Testsuite `from_avmplus/e4x/Global/e13_1_2_1`
> memory allocation of 137438953472 bytes failed

## Testsuite `from_avmplus/e4x/Global/isXMLNameTypeErr`
| Test | Passes? | Expected | Actual |
| ---- | ---- | ---- | ---- |
| Asserting for TypeError | :x: | ? | did not run |
| var object1 = { toString: function() { return this; } }; isXMLName(object1) | :x: | `Error #1050` | `TypeError not thrown` |
| Asserting for TypeError | :x: | ? | did not run |
| var object2 = { toString: function() { return String(object1); } }; isXMLName(object2) | :x: | `Error #1050` | `TypeError not thrown` |

Lord-McSweeney:
This is because we don't support isXMLName.

Lord-McSweeney:
The following tests are now passing:
- `Global/e13_1_2_1`
- `Global/isXMLNameTypeErr`

PR #12993, opened 2023-08-28
avm2: Implement isXMLName

sleepycatcoding:
TODO:

- [x] This probably should be moved to a submodule of `e4x`


sleepycatcoding:
Closes #12354

Lord-McSweeney (approved):

sleepycatcoding:
Rebased to solve merge conflicts.

core/src/avm2/e4x/is_xml_name.rs
```diff
@@ -0,0 +1,443 @@
...
+
+use crate::avm2::{Activation, Error, Value};
+use std::sync::OnceLock;
+
+// https://www.w3.org/TR/2004/REC-xml-20040204/#NT-Letter
+fn generate_letters() -> Box<[char]> {
+    let mut buf = vec![];
+
+    // Base char
+    buf.extend(0x0041..=0x005A);
```
adrian17, line 14: Two main questions:
- can we avoid making a 30kB vec (and iterating it) just for this? avmplus kinda got around it by only storing start-end ranges instead.
- any chance these ranges overlap with unicode, so we could use an existing (std)lib function to check for an unicode category?

adrian17: For the record: https://github.com/adobe/avmplus/blob/858d034a3bd3a54d9b70909386435cf4aec81d21/core/AvmCore.cpp#L3503

adrian17: And with ranges alone, the arrays are small enough that we could probably have them be just `const` instead of lazy-initialized.

sleepycatcoding: > Two main questions:
> 
>     * can we avoid making a 30kB vec (and iterating it) just for this? avmplus kinda got around it by only storing start-end ranges instead.
> 
>     * any chance these ranges overlap with unicode, so we could use an existing (std)lib function to check for an unicode category?

To be honest I do not know if any overlap with unicode, haven't looked into it.

sleepycatcoding: Pushed a commit to change it to static arrays instead.

sleepycatcoding: >     * any chance these ranges overlap with unicode, so we could use an existing (std)lib function to check for an unicode category?

There are methods with which ranges overlap, but if we would use them it would broaden the allowed characters, so compatibility with Flash would be lost. The reason for this is that Flash's characters were based on Unicode database version 2, but Rust's methods are based on the latest version.

evilpie:
I am not usually one for optimizing prematurely, but walking the whole table for every char seems quite slow.

I think we could at least bail out early when `*start > c`. Or maybe do  a binary search?

adrian17:
This feels like an extreme overkill. For 99.9% of cases, this will just validate ascii chars (and likely return true) and that's it - so it's not surprising that binary search is worse here. Further, it'll bail on the first non-matching char and input strings are usually very short, so I'd say vectorizarion also isn't important here.

I'd say that avmplus's original algo (maybe with that early bail? Or another kind of early check for ascii) is easily good enough here.

lvyitian:
> This feels like an extreme overkill. For 99.9% of cases, this will just validate ascii chars (and likely return true) and that's it - so it's not surprising that binary search is worse here. Further, it'll bail on the first non-matching char and input strings are usually very short, so I'd say vectorizarion also isn't important here.
> 
> I'd say that avmplus's original algo (maybe with that early bail? Or another kind of early check for ascii) is easily good enough here.

We can specialize it for ascii characters, just add a "match expression" for ascii characters before doing binary searching.

adrian17:
> The world isn't just ASCII, use cases in other languages (e.g. CJK) are equally important.

...not really, not in cases like XML. This is not generic code dealing with arbitrary text, this is relevant purely for XML element names. So this is not directly related to "language" - across the world, even in CJK language-speaking countries, it's likely they'll still exclusively use ASCII for raw identifiers in data formats like XML. Doesn't mean it _can't_ happen, but... again, the strings here are so short that the difference is still mostly completely irrelevant and not worth microoptimizing for.

(if you _did_ really want to microoptimize, I imagine some of these byte ranges for digits, letters etc are adjancent and could be merged, to avoid repeated scans; but again, I don't think this is worth the effort at all.)

BTW, in your benchmark, `'你'.is_ascii()` trivially optimizes to a constant.

lvyitian:
> > The world isn't just ASCII, use cases in other languages (e.g. CJK) are equally important.
> 
> ...not really, not in cases like XML. This is not generic code dealing with arbitrary text, this is relevant purely for XML element names. So this is not directly related to "language" - across the world, even in CJK language-speaking countries, it's likely they'll still exclusively use ASCII for raw identifiers in data formats like XML. Doesn't mean it _can't_ happen, but... again, the strings here are so short that the difference is still mostly completely irrelevant and not worth microoptimizing for.
> 
> (if you did want to _really_ microoptimize, I imagine some of these byte ranges for digits, letters etc are adjancent and could be merged, to avoid repeated scans; but again, I don't think this is worth the effort at all; I think we should just focus on the common case.)
> 
> BTW, in your benchmark, `'你'.is_ascii()` trivially optimizes to a constant.

But 

> > The world isn't just ASCII, use cases in other languages (e.g. CJK) are equally important.
> 
> ...not really, not in cases like XML. This is not generic code dealing with arbitrary text, this is relevant purely for XML element names. So this is not directly related to "language" - across the world, even in CJK language-speaking countries, it's likely they'll still exclusively use ASCII for raw identifiers in data formats like XML. Doesn't mean it _can't_ happen, but... again, the strings here are so short that the difference is still mostly completely irrelevant and not worth microoptimizing for.
> 
> (if you did want to _really_ microoptimize, I imagine some of these byte ranges for digits, letters etc are adjancent and could be merged, to avoid repeated scans; but again, I don't think this is worth the effort at all; I think we should just focus on the common case.)
> 
> BTW, in your benchmark, `'你'.is_ascii()` trivially optimizes to a constant.

But it is better than nothing right? If there's an extreme case that the string contains a lot of non-ascii characters and it is extremely long, now that you can improve the performance without the need of significant effort and without encountering performance regression in common situation, why not do it, just specialize it for ascii characters and do binary searching.

adrian17 (approved):
I think this is fine as-is, it's already better than avmplus's version. Let's get it in :)

PR #12994, opened 2023-08-28
avm2: Implement flash.media.Sound constructor with stream argument

danielhjacobs:
Together with #12967, fixes https://dnasubway.cyverse.org/files/tour/tour-all.swf.

According to https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/media/Sound.html#Sound(), this delegates to load.

Now if you use the `context` argument in the `Sound` constructor, you'll get a message about using `load` with `context`, not a message about using the constructor with `context`, because `load` has the actual stub.

### DIFF
--- a/core/src/avm2/globals/flash/media/sound.rs
+++ b/core/src/avm2/globals/flash/media/sound.rs
@@ -2,12 +2,13 @@
 
 use crate::avm2::activation::Activation;
 use crate::avm2::object::{Object, QueuedPlay, SoundChannelObject, TObject};
+use crate::avm2::parameters::ParametersExt;
 use crate::avm2::value::Value;
 use crate::avm2::Error;
 use crate::backend::navigator::Request;
 use crate::character::Character;
 use crate::display_object::SoundTransform;
-use crate::{avm2_stub_constructor, avm2_stub_getter, avm2_stub_method};
+use crate::{avm2_stub_getter, avm2_stub_method};
 use swf::{SoundEvent, SoundInfo};
 
 pub use crate::avm2::object::sound_allocator;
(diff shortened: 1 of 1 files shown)


core/src/avm2/globals/flash/media/sound.rs
```diff
@@ -18,8 +19,8 @@ pub fn init<'gc>(
     this: Object<'gc>,
     args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
-    if !args.is_empty() {
-        avm2_stub_constructor!(activation, "flash.media.Sound", "with arguments");
+    if args.try_get_object(activation, 0).is_some() {
+        let _ = load(activation, this, args);
```
Dinnerbone, line 23: Just to be clear, is this intentional that it ignores error?

danielhjacobs: I'm not really sure what Flash does here. The documentation said `If you do not pass a valid URLRequest object to the Sound constructor, you must call the load() function for the Sound object yourself, or the stream will not load.` Which would imply the Sound object still gets constructed to me. But I'm not sure.

danielhjacobs: Unlike the `load` method, which says it throws, this does not state that it throws in the documentation, so I feel like it probably ignores error.

danielhjacobs: I'm still not really familiar with Rust though, so this was really just a result of that. Would this be the right way to bubble the error up?
```rs
    if args.try_get_object(activation, 0).is_some() {
        let load_result = load(activation, this, args);

        let _ = match load_result {
            Ok(load) => load,
            Err(_) => return load_result,
        };
    }

```

sleepycatcoding:
Does this supersede #11972?

danielhjacobs:
> Does this supersede #11972?

Oh, no that's probably mostly better.

core/src/avm2/globals/flash/media/sound.rs
danielhjacobs: Closed in favor of #11972

PR #13018, opened 2023-08-30
avm2: Update TextField.as

pozirk:
Added getLineOffset() method.


n0samu:
Thanks for contributing! Unfortunately, this will not work because `native` methods need a corresponding Rust method. You could create a method stub fully in AS instead - search for other "stub" PRs here to see how it's done.

PR #13071, opened 2023-09-05
avm2: Implement avmplus.describeTypeJSON and use it for describeType

Aaron1011:
Some SWFS (in particular, anything using Unity) call avmplus.describeTypeJSON, and rely on the behavior of the various flags.

This PR changes our internal implementation to implement describeTypeJSON (producing an `Object` with dynamic fields). We then convert this to XML in `describeType`, using an implementation inspired by avmplus.

The existing describeType tests are passing - in a follow-up PR, I'll add tests for describeTypeJSON


core/src/avm2/globals/flash/display3D/textures/CubeTexture.as
```diff
@@ -1,6 +1,9 @@
 package flash.display3D.textures {
     import flash.display.BitmapData;
+    import flash.utils.ByteArray;
     public final class CubeTexture extends TextureBase {
-        public native function uploadFromBitmapData(source:BitmapData, side:uint, miplevel:uint = 0):void
+        public native function uploadFromBitmapData(source:BitmapData, side:uint, miplevel:uint = 0):void;
```
Lord-McSweeney, line 5: While I'm not complaining about this stub, I don't think it was added intentionally.

core/src/avm2/globals/avmplus.rs
```diff
@@ -1 +1,519 @@
...
+
+    let bases = activation
+        .avm2()
+        .classes()
+        .array
+        .construct(activation, &[])?
+        .as_array_object()
+        .unwrap();
+
+    let interfaces = activation
```
adrian17, line 102: we generally write this via `ArrayObject::from_storage(activation, ArrayStorage::new(0)`, no need to go through the constructor/allocator dispatch :D

adrian17: We could maybe wanna wrap it in `ArrayObject::empty(acativation)`?

(Alternatively, (though I don't think you'll prefer that - just throwing as a possibility), you could start with `Storage`s, fill them and create the `ArrayObject`s at the end.)

adrian17: we could do a similar thing for `classes().object` (tbh I'm surprised we didn't have it before):

```rs
impl<'gc> ScriptObject<'gc> {
    pub fn new<'gc>(
    ) -> Object<'gc> {
        let class = self.avm2().classes().object;
        ScriptObject(GcCell::new(activation.context.gc_context, ScriptObjectData::new(class))).into()
    }
```

adrian17 (approved):
Approved, but that Array/ScriptObject change would clean it up a decent bit, save like 50LOC.

Aaron1011:
I'll make a followup pr that adds helper methods

torokati44:
Clippy is not happy now, possibly due to #13072 having been merged.

adrian17:
> I'll make a followup pr that adds helper methods

FYI changing these

```
    let variables = activation
        .avm2()
        .classes()
        .array
        .construct(activation, &[])?
        .as_array_object()
        .unwrap();
```
Doesn't need new helpers, only `object` ones do (but sure, a separate PR is fine, alternatively I can do it).

PR #13072, opened 2023-09-05
avm2: Remove unnecessary '&mut self' from Object methods

Aaron1011:


adrian17 (approved):
weird, I distinctly remember not being able to remove these in the past.

PR #13077, opened 2023-09-05
core, avm2, web: Implement Geolocation

michaelwright235:
I'm excited to contribute to this amazing project! So I decided to implement Geolocation logic that is used by some mobile AIR games and apps. It's not present on desktop Flash Player. 

**This pull request consists of these parts:**
1. Geolocation related events. All of the implemented AS classes are AS3. 
  * `flash.events.PermissionEvent` (when users choices whether to grant permission or not) and `flash.permissions.PermissionStatus` that contains related constants.
  * `flash.events.GeolocationEvent` that is sent when the device's geolocation changes.
2. `flash.sensors.Geolocation` itself. One uses it to set up geolocation stuff in their scripts.
3. New geolocation backend. We need a separate backend to handle geolocation requests. NullGeolocationBackend is also included, it tells that device doesn't have a geo sensor.
4. Implementation of geolocation backend for web.

**Brief explanation**

Once a Geolocation class is created we immediately add it to GeolocationInstances struct. It should contain all class instances to dispatch events on all of them. Then the frontend is requested with a permission. When user choose a variant, a PermissionEvent and StatusEvent get sent to geo classes. If a permission is granted, we then start sending GeolocationEvents with the requested interval.

**Requesting permission problem**

Geolocation class was firstly introduced in AIR 2.5. Firstly it didn't require requesting user permission for geolocation use, since mobile devices back than didn't really care about privacy. That's why most of code examples on the web just add a listener for a geolocation update. If we try this on a modern Android device, we'll see that it doesn't work and it gets stuck waiting forever. Some versions of AIR even throw an error. So to address this issue I decided to request a permission every time a class gets created. Plus all modern browsers and OSs require user permission anyway. It makes my implementation compatible with old and new as3 code.

**Resources:**
- Adobe Flash docs ([Geolocation](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/sensors/Geolocation.html), [GeolocationEvent](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/events/GeolocationEvent.html), [PermissionEvent](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/events/PermissionEvent.html), [PermissionStatus](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/permissions/PermissionStatus.html), [StatusEvent](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/events/StatusEvent.html)).
- Real world code examples: [Detecting geolocation changes](https://help.adobe.com/en_US/as3/dev/WS144092a96ffef7cc-66bf4d0212658dde8c4-7fff.html), [Using the Geolocation Sensor in Flash Using AS3](https://web.archive.org/web/20160316232731/http://www.republicofcode.com/tutorials/flash/as3geolocation/).

**Current state:**
At this moment the web implementation works great. I made a [simple test swf projects](https://github.com/michaelwright235/swf_geolocation_test) to test this out.

**Current todo list:**
- [ ] Make GeolocationInstances struct accept multiple instances of Geolocation class. Currently only the last created one will receive events. I didn't quite figure out how to deal with garbage collection stuff, so this struct needs to be rewritten.


n0samu:
Thanks so much for contributing! However, I see Ruffle's main function as bringing Flash web games back to the web. Why should we support an AIR API on the web that Flash Player never supported? Are you working with some company that is trying to port their AIR app to the web or something like that? What about calling the browser's geolocation APIs via ExternalInterface?

michaelwright235:
It was just my project in my spare time. I've looked at the Ruffle features matrix and decided to implement this functionality.
Existing AIR apps for Android and iPhone become obsolete as well. For instance installing old android AIR app may simply fail because one could only produce only arm-v7 apks (arm-v8 was added only long after AIR was bought by Harman), some phones now don't them anymore. Running an app produced with AIR 2-ish will certainly fail on newer Android versions (checked it myself). iOS is even worse if we talk about API compatibility. In my opinion, supporting AIR apps will help preserving those kind of apps, plus extracting the inner swf from them is super easy.
Hmm, I didn't think of using ExternalInterface.

adrian17:
Indeed, it's a fact that all of these APIs are AIR-only, and inaccessible for the normal Flash Player. The SDK (and player, though I think they are usually just not compiled into the player at all) hides them via version metadata (like `[API("668")]`) in the .SWC files.
Further, we can't "just" make them accessible to normal SWFs (all the privacy, permissions and "why Adobe didn't want this API on web in the first place" questions aside), because some AS3 libraries check for API presence as a sort of feature detection, and us pretending to be AIR can break content that wasn't intended to be embedded in non-AIR contexts. Similarly (not applicable here, just a trivia), adding AIR-only fields or methods to classes can also affect class inheritance validation.

Our current consensus (quickly discussed on Discord moments ago :) ) is that if we do merge this, this will need to be disabled by default and inaccessible (via namespace versioning mechanisms) unless somehow an explicit switch "pretend to be an AIR player" is switched.

adrian17:
> supporting AIR apps

Just making sure, this requires someone to extract the SWF from the AIR app first, right?

adrian17:
> unless somehow an explicit switch "pretend to be an AIR player" is switched.

(we didn't decide on any specifics here btw. The prerequisite for the hiding mechanism is https://github.com/ruffle-rs/ruffle/pull/12967 .)

Lord-McSweeney:
I agree, this should only be available on AIR (which means the web backend should be unnecessary). If we ever do implement AIR APIs, it will likely be gated behind a flag (like lightspark's `--air`), which would _probably_ only be available on desktop Ruffle.

Aaron1011:
What would prevent air on web?

n0samu:
We subsequently agreed that an `air` configuration option, if implemented, should be available on both desktop and web.

Dinnerbone:
With #14159 and #14166 we are now bringing in AIR support!

Aaron1011:
Ruffle now has AIR support, which can be used to make these classes visible under AIR only

PR #13094, opened 2023-09-07
web: Support for Sockets with WebSocket

sleepycatcoding:
As WebSocket is not a normal TCP socket, it requires extra software like [`websockify`](https://github.com/novnc/websockify) on the server.

Unresolved questions:

- [ ] Make a interface so a custom socket connection method could be used (like integrated server running in web browser), but `ExternalInterface` exists so that kinda invalidates this point.
- [x] How do we determine if we use `wss` (HTTPS for WebSockets) or `ws` protocol?
- [x] An API to intercept to modify host, port, or cancel connection entirely?

### DIFF
--- a/web/packages/core/src/config.ts
+++ b/web/packages/core/src/config.ts
@@ -44,5 +44,5 @@ export const DEFAULT_CONFIG: Required<BaseLoadOptions> = {
     openUrlMode: OpenURLMode.Allow,
     allowNetworking: NetworkingAccessMode.All,
     openInNewTab: null,
-    wsProxy: [],
+    socketProxy: [],
 };
--- a/web/packages/core/src/config.ts
+++ b/web/packages/core/src/config.ts
@@ -44,4 +44,5 @@ export const DEFAULT_CONFIG: Required<BaseLoadOptions> = {
     openUrlMode: OpenURLMode.Allow,
     allowNetworking: NetworkingAccessMode.All,
     openInNewTab: null,
+    wsProxy: [],
 };
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2032,6 +2032,36 @@ dependencies = [
 ]
 
 [[package]]
+name = "gloo-net"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8ac9e8288ae2c632fa9f8657ac70bfe38a1530f345282d7ba66a1f70b72b7dc4"
+dependencies = [
+ "futures-channel",
+ "futures-core",
+ "futures-sink",
+ "gloo-utils",
+ "http",
+ "js-sys",
+ "pin-project",
+ "thiserror",
+ "wasm-bindgen",
+ "wasm-bindgen-futures",
+ "web-sys",
+]
+
+[[package]]
+name = "gloo-utils"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0b5555354113b18c547c1d3a98fbf7fb32a9ff4f6fa112ce823a21641a0ba3aa"
+dependencies = [
+ "js-sys",
+ "wasm-bindgen",
+ "web-sys",
+]
+
+[[package]]
 name = "glow"
 version = "0.12.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4217,8 +4247,10 @@ dependencies = [
  "base64",
  "chrono",
  "console_error_panic_hook",
+ "futures-util",
  "generational-arena",
  "getrandom",
+ "gloo-net",
  "js-sys",
  "ruffle_core",
  "ruffle_render",
--- a/web/Cargo.toml
+++ b/web/Cargo.toml
@@ -64,5 +64,5 @@ features = [
     "ChannelMergerNode", "ChannelSplitterNode", "ClipboardEvent", "DataTransfer", "Element", "Event",
     "EventTarget", "GainNode", "Headers", "HtmlCanvasElement", "HtmlDocument", "HtmlElement", "HtmlFormElement",
     "HtmlInputElement", "HtmlTextAreaElement", "KeyboardEvent", "Location", "PointerEvent",
-    "Request", "RequestInit", "Response", "Storage", "WheelEvent", "Window",
+    "Request", "RequestInit", "Response", "Storage", "WheelEvent", "Window", "ProgressEvent", "WebSocket", "MessageEvent", "BinaryType", "ErrorEvent", "CloseEvent",
 ]
--- a/web/Cargo.toml
+++ b/web/Cargo.toml
@@ -50,6 +50,8 @@ serde = { version = "1.0.188", features = ["derive"] }
 thiserror = "1.0"
 base64 = "0.21.4"
 async-channel = "1.9.0"
+futures-util = { version = "0.3.28", features = ["sink"] }
+gloo-net =  { version = "0.4.0", default-features = false, features = ["websocket"] }
 
 [dependencies.ruffle_core]
 path = "../core"
@@ -64,5 +66,5 @@ features = [
     "ChannelMergerNode", "ChannelSplitterNode", "ClipboardEvent", "DataTransfer", "Element", "Event",
     "EventTarget", "GainNode", "Headers", "HtmlCanvasElement", "HtmlDocument", "HtmlElement", "HtmlFormElement",
     "HtmlInputElement", "HtmlTextAreaElement", "KeyboardEvent", "Location", "PointerEvent",
-    "Request", "RequestInit", "Response", "Storage", "WheelEvent", "Window", "ProgressEvent", "WebSocket", "MessageEvent", "BinaryType", "ErrorEvent", "CloseEvent",
+    "Request", "RequestInit", "Response", "Storage", "WheelEvent", "Window",
 ]
--- a/web/packages/core/src/load-options.ts
+++ b/web/packages/core/src/load-options.ts
@@ -248,7 +248,7 @@ export const enum NetworkingAccessMode {
     None = "none",
 }
 
-export interface WebSocketProxy {
+export interface SocketProxy {
     host: string;
     port: number;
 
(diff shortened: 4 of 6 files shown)


Dinnerbone:
>  How do we determine if we use wss (HTTPS for WebSockets) or ws protocol?

There should be an object in the config for which WS server to use as a proxy, and that should be the ws address (`wss://example.org/proxy`). Ideally it could be configured per host/port being connected to, as well.

We shouldn't be directly connecting to the flash provided thing, as we need to proxy it.

sleepycatcoding:
> > How do we determine if we use wss (HTTPS for WebSockets) or ws protocol?
> 
> There should be an object in the config for which WS server to use as a proxy, and that should be the ws address (`wss://example.org/proxy`). Ideally it could be configured per host/port being connected to, as well.
> 
> We shouldn't be directly connecting to the flash provided thing, as we need to proxy it.

This could probably be used such as if no proxies are provided, we send connection failed message back to AVM, which would solve the third point entirely. Also yeah the proxy config idea solves the second point too.

web/packages/core/src/config.ts
```diff
@@ -44,4 +44,5 @@ export const DEFAULT_CONFIG: Required<BaseLoadOptions> = {
     openUrlMode: OpenURLMode.Allow,
     allowNetworking: NetworkingAccessMode.All,
     openInNewTab: null,
+    wsProxy: [],
```
sleepycatcoding, line 47: I do not like the name of the config option, but I couldn't think of a better name, so suggestions are welcome.

n0samu: I would call it `socketProxy`!

sleepycatcoding:
> * Make a interface so a custom socket connection method could be used (like integrated server running in web browser), but `ExternalInterface` exists so that kinda invalidates this point.

I just thought this could be a config option too, something like `socketBackend` that can be added in the future, and for backwards compatibility, this can just default to WebSocket.

web/packages/core/src/config.ts
sleepycatcoding: Renamed.

Minigugus:
> > * Make a interface so a custom socket connection method could be used (like integrated server running in web browser), but `ExternalInterface` exists so that kinda invalidates this point.
> 
> I just thought this could be a config option too, something like `socketBackend` that can be added in the future, and for backwards compatibility, this can default to WebSocket.

For me the Rust part shouldn't have to care about the actually proxy implementation, espacially since WebSocket isn't the only way to implement proxing in browser (mock server in the browser, WebRTC, [WebTransport](https://developer.mozilla.org/en-US/docs/Web/API/WebTransport) or even [Direct Sockets](https://wicg.github.io/direct-sockets/#abstract)). :confused: I agree WebSockets will be used most of the time but I don't think it should be the default.

Or at least if you only want to support WebSocket for now, why not just move the backend implementation to the JS part? (that's what I did in my previous PR by the way: https://github.com/Minigugus/ruffle/blob/d4ed4497f7ee363d97d393155dafdbcaa37ee4fd/web/packages/core/src/ruffle-player.ts#L61-L66)

sleepycatcoding:
> > > * Make a interface so a custom socket connection method could be used (like integrated server running in web browser), but `ExternalInterface` exists so that kinda invalidates this point.
> > 
> > 
> > I just thought this could be a config option too, something like `socketBackend` that can be added in the future, and for backwards compatibility, this can default to WebSocket.
> 
> For me the Rust part shouldn't have to care about the actually proxy implementation, espacially since WebSocket isn't the only way to implement proxing in browser (mock server in the browser, WebRTC, [WebTransport](https://developer.mozilla.org/en-US/docs/Web/API/WebTransport) or even [Direct Sockets](https://wicg.github.io/direct-sockets/#abstract)). 😕 I agree WebSockets will be used most of the time but I don't think it should be the default.
> 
> Or at least if you only want to support WebSocket for now, why not just move the backend implementation to the JS part? (that's what I did in my previous PR by the way: https://github.com/Minigugus/ruffle/blob/d4ed4497f7ee363d97d393155dafdbcaa37ee4fd/web/packages/core/src/ruffle-player.ts#L61-L66)
> 
> Regarding the API design, something like below seems the perfect world solution as it could support modern workflow, like [`WebSocketStream`](https://developer.chrome.com/articles/websocketstream/#introductory-example) (but I agree it's not the easiest API to implement in the Rust part 😅)
> 
> ```ts
> interface SocketConnectOptions {
>   host: string,
>   port: number,
>   timeout: number,
> 
>   // implements the TransformStream interface
>   readable: ReadableStream<Uint8Array>,
>   writable: WritableStream<Uint8Array>
> }
> 
> type SocketProxy = (options: SocketConnectOptions) => Promise<boolean>
> 
> // sample implementation:
> const gzipedWebSocketProxy: SocketProxy = async opt => {
>   const ws = new WebSocketStream(`wss://myproxyhost/${opt.host}:${opt.port}`)
>   const stream = await ws.opened
> 
>   stream.readable
>      .pipeThrough(new CompressionStream("gzip"))
>      .pipeThrough(opt) // possible since `opt` implements the `TransformStream` interface
>      .pipeThrough(new DecompressionStream("gzip"))
>      .pipeTo(stream.writable)
>      .catch(ex => console.warn('[%s:%d] connection error', opt.host, opt.port, ex))
> 
>   return true;
> }
> ```

I would have to look into how we could create a Readable/WritableStream with the current channels we have on the Rust side. And how we would drop them correctly, since dropping the receiver signals the AVM that Socket is to be closed.

sleepycatcoding:
> EDIT: found this crate, maybe it can help if you're still interested by Web Streams: https://crates.io/crates/wasm-streams 👍

Yeah, this crate looks nice, I will try to make a prototype later this evening.

marcussacana:
@sleepycatcoding I know this isn't the best place for this but since your fork has issues disabled, may I ask you an test build? I want test a game that I think the last detail to be compatible with ruffle is the Socket implementation, so I want give a try in my side early.

Dinnerbone (approved):
I think that for Ruffle and in the interests of making thing both "just work" and also "working sooner rather than later', this PR is absolutely perfect.

In the future we may investigate adding different backends or doing it in other ways, but I think this is perfectly reasonable for today and I'd be very happy to take this in ASAP.

robinheidrich:
Is it possible that if no proxy is configured, Ruffle will assume that the server also has WebSocket implemented and connect directly to it? Whether `wss` or `ws` could be either an option or based on the URL scheme.

If the WebSocket connection then fails, Ruffle could give an error that it has to be configured for web.

Should I create a new issue for this? @sleepycatcoding @Dinnerbone

Dinnerbone:
> Is it possible that if no proxy is configured, Ruffle will assume that the server also has WebSocket implemented and connect directly to it? Whether `wss` or `ws` could be either an option or based on the URL scheme.
> 
> If the WebSocket connection then fails, Ruffle could give an error that it has to be configured for web.
> 
> Should I create a new issue for this? @sleepycatcoding @Dinnerbone 

I don't think we should assume that, sorry, but if you have a good argument for it then we can discuss it.

The likelyhood is extremely low, and it'd force anyone who ran a server to shut down that server and spin up a websocket one in it's place - preventing actual flash from connecting to it (including ruffle desktop)

robinheidrich:
> > Is it possible that if no proxy is configured, Ruffle will assume that the server also has WebSocket implemented and connect directly to it? Whether `wss` or `ws` could be either an option or based on the URL scheme.
> > If the WebSocket connection then fails, Ruffle could give an error that it has to be configured for web.
> > Should I create a new issue for this? @sleepycatcoding @Dinnerbone
> 
> I don't think we should assume that, sorry, but if you have a good argument for it then we can discuss it.
> 
> The likelyhood is extremely low, and it'd force anyone who ran a server to shut down that server and spin up a websocket one in it's place - preventing actual flash from connecting to it (including ruffle desktop)

Ah I see, I thought you could do a WebSocket upgrade on the same port as the TCP server but unfortunately that doesn't work.

Nevertheless, this is a huge step forward for Ruffle and I hope NetConnection follows soon as well!

Thanks for the answer!

markhughes:
I just tried this out with the new build using the following:

```
  socketProxy: [
    {
      host: "127.0.0.1",
      port: 9000,
      proxyUrl: "ws://127.0.0.1:9000",
    },
  ],
  ```

However, I am not seeing any `ws` open attempt in the self-hosted version and connect is returning false. 
  
Also built the branch, but no luck.

Dinnerbone:
> I just tried this out with the new build using the following:
> 
> ```
>   socketProxy: [
>     {
>       host: "127.0.0.1",
>       port: 9000,
>       proxyUrl: "ws://127.0.0.1:9000",
>     },
>   ],
>   ```
> 
> However, I am not seeing any `ws` open attempt in the self-hosted version and connect is returning false. 
>   
> Also built the branch, but no luck.
> 

Anything in the browser console? What about the network tab?

markhughes:
Sadly no :( here is a minimal reproduction 
[as2_xmlsocket_test.zip](https://github.com/ruffle-rs/ruffle/files/12916398/as2_xmlsocket_test.zip)

sleepycatcoding:
> Sadly no :( here is a minimal reproduction [as2_xmlsocket_test.zip](https://github.com/ruffle-rs/ruffle/files/12916398/as2_xmlsocket_test.zip)

It seems the SWF does not try to connect even in Ruffle Desktop, so it does not seem web specific.

sleepycatcoding:
Ok, I think I know what is happening, it seems the reproduction you attached, creates a new Object and attaches the XmlSocket prototype methods to it, which makes this cast fail: https://github.com/ruffle-rs/ruffle/blob/b02038bcd25d933ed9f114954e9b59bd6c45d71f/core/src/avm1/globals/xml_socket.rs#L112-L117.

markhughes:
Ah very interesting, yes, it was a common thing for lots of frameworks (this one being Fortress Flash Development Kit) to use this method. Assuming this is something else then, should I open a new issue?

markhughes:
Have published the source reproduction here: https://github.com/markhughes/as2_xmlsocket_test

PR #13099, opened 2023-09-07
desktop: Change "Open URL Mode" to "Open Links"

Positron832Alt:
Change "Open URL Mode" to "Open Links" as I think it makes more sense. Also change the "Confirm" option to "Ask" to match the options for TCP connections. Sorry if I did things wrong.

### DIFF
--- a/core/src/backend/navigator.rs
+++ b/core/src/backend/navigator.rs
@@ -42,14 +42,14 @@ pub enum SocketMode {
 /// The handling mode of links opening a new website.
 #[cfg_attr(feature = "clap", derive(clap::ValueEnum))]
 #[derive(Clone, Copy, Debug, PartialEq, Serialize, Deserialize)]
-pub enum OpenURLMode {
+pub enum OpenLinks {
     /// Allow all links to open a new website.
     #[serde(rename = "allow")]
     Allow,
 
-    /// A confirmation dialog opens with every link trying to open a new website.
-    #[serde(rename = "confirm")]
-    Confirm,
+    /// Ask the user every time a link is trying to open a new website.
+    #[serde(rename = "ask")]
+    Ask,
 
     /// Deny all links to open a new website.
     #[serde(rename = "deny")]
--- a/core/src/backend/navigator.rs
+++ b/core/src/backend/navigator.rs
@@ -47,9 +47,9 @@ pub enum OpenURLMode {
     #[serde(rename = "allow")]
     Allow,
 
-    /// Ask the user every time a link is trying to open a new website.
-    #[serde(rename = "ask")]
-    Ask,
+    /// A confirmation dialog opens with every link trying to open a new website.
+    #[serde(rename = "confirm")]
+    Confirm,
 
     /// Deny all links to open a new website.
     #[serde(rename = "deny")]
--- a/core/src/backend/navigator.rs
+++ b/core/src/backend/navigator.rs
@@ -42,7 +42,7 @@ pub enum SocketMode {
 /// The handling mode of links opening a new website.
 #[cfg_attr(feature = "clap", derive(clap::ValueEnum))]
 #[derive(Clone, Copy, Debug, PartialEq, Serialize, Deserialize)]
-pub enum OpenLinks {
+pub enum OpenURLMode {
     /// Allow all links to open a new website.
     #[serde(rename = "allow")]
     Allow,
--- a/desktop/assets/texts/en-US/settings.ftl
+++ b/desktop/assets/texts/en-US/settings.ftl
@@ -16,10 +16,10 @@ tcp-connections-allow = Allow
 tcp-connections-ask = Ask
 tcp-connections-deny = Deny
 
-open-links = Open Links
-open-links-allow = Allow
-open-links-ask = Ask
-open-links-deny = Deny
+open-url-mode = Open Links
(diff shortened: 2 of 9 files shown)


web/src/navigator.rs
```diff
@@ -145,19 +145,19 @@ impl NavigatorBackend for WebNavigatorBackend {
         let window = window().expect("window()");
 
         if url.scheme() != "javascript" {
-            if self.open_url_mode == OpenURLMode::Confirm {
+            if self.open_links == OpenLinks::Ask {
                 let message = format!("The SWF file wants to open the website {}", &url);
                 // TODO: Add a checkbox with a GUI toolkit
-                let confirm = window
+                let ask = window
```
Lord-McSweeney, line 151: I feel like the variable name `confirm` should be kept as-is (instead of changing it to `ask`), here and in other places.

core/src/backend/navigator.rs
```diff
@@ -42,14 +42,14 @@ pub enum SocketMode {
 /// The handling mode of links opening a new website.
 #[cfg_attr(feature = "clap", derive(clap::ValueEnum))]
 #[derive(Clone, Copy, Debug, PartialEq, Serialize, Deserialize)]
-pub enum OpenURLMode {
```
Lord-McSweeney, line 45: Also, an `OpenLinks` enum feels weird. I think this should be kept as `OpenURLMode` (or at least `OpenLinkMode`).

core/src/backend/navigator.rs
```diff
@@ -42,14 +42,14 @@ pub enum SocketMode {
...
+pub enum OpenLinks {
```
Korne127, line 45: I would prefer to just change the translation files.

I think that it makes sense to keep the frontend settings consistent and therefore approve changing the translations. But I think that the underlying variables in the backend should stay as they currently are. Especially the enum should remain `OpenURLMode` since the enum is about the mode of how URLs can be opened. "Open Links" is good as a text displayed to a user in the settings, but it's not a good descriptive name for the underlying variables in the backend.

desktop/src/backends/navigator.rs
```diff
@@ -139,20 +139,20 @@ impl NavigatorBackend for ExternalNavigatorBackend {
...
                 .set_title("Open website?")
                 .set_level(MessageLevel::Info)
                 .set_description(&message)
                 .set_buttons(MessageButtons::OkCancel)
                 .show();
-            if !confirm {
+            if !ask {
                 tracing::info!("SWF tried to open a website, but the user declined the request");
                 return;
             }
```
Korne127, line 154: I don't think this rename makes sense. "Confirm" was meant as in the user confirming it and pressing the okay button to confirm it. So `if !confirm` makes sense as in "if the user didn't confirm it".
`ask` doesn't make sense here as a variable name (in this branch the user gets asked in any case, the condition checks whether the user confirmed to open the URL or not), therefore I wouldn't change it.

core/src/backend/navigator.rs
Korne127: Yeah, all the ftl files, essentially like `open-url-mode = Open Links`.
Although the part of changing the `Confirm` enum value to `Ask` also makes sense and is okay imo, my main problem is that OpenURLMode has been renamed everywhere.

Korne127:
Right now the enum is still called `OpenLinks` in many files (leading to this PR currently not compiling). So that needs to be changed as well, and I would also change the variables you changed to "open_links" back to "open_url_mode".
The enum variable being changed from `Confirm` to `Ask` does make sense though imo. I don't think it's necessary to change, but if you want to, I'm fine with also renaming that in the backend.

desktop/assets/texts/en-US/settings.ftl
```diff
@@ -16,10 +16,10 @@ tcp-connections-allow = Allow
 tcp-connections-ask = Ask
 tcp-connections-deny = Deny
 
-open-url-mode = Open URL Mode
-open-url-mode-allow = Allow
-open-url-mode-confirm = Confirm
-open-url-mode-deny = Deny
+open-links = Open Links
+open-links-allow = Allow
```
danielhjacobs, line 20: Same here https://github.com/ruffle-rs/ruffle/blob/53199a51164ce338a259f1dba5a850eb7280cf92/desktop/src/gui/open_dialog.rs#L196

desktop/assets/texts/en-US/settings.ftl
```diff
@@ -16,10 +16,10 @@ tcp-connections-allow = Allow
...
+open-links-ask = Ask
```
danielhjacobs, line 21: Also here https://github.com/ruffle-rs/ruffle/blob/53199a51164ce338a259f1dba5a850eb7280cf92/desktop/src/gui/open_dialog.rs#L201

desktop/assets/texts/en-US/settings.ftl
```diff
@@ -16,10 +16,10 @@ tcp-connections-allow = Allow
...
+open-links-deny = Deny
```
danielhjacobs, line 22: And here https://github.com/ruffle-rs/ruffle/blob/53199a51164ce338a259f1dba5a850eb7280cf92/desktop/src/gui/open_dialog.rs#L206

danielhjacobs (changes requested):

PR #13118, opened 2023-09-08
avm2: Fix avmplus.describeTypeJSON bugs and add test

Aaron1011:


core/src/avm2/class.rs
```diff
@@ -370,6 +372,7 @@ impl<'gc> Class<'gc> {
                     // A 'callable' class doesn't have a signature - let the
                     // method do any needed coercions
                     vec![],
+                    Multiname::any(activation.context.gc_context),
```
adrian17, line 375: not PR related: yikes, we really should make `Any` not allocate at all, maybe replacing it by an Option-like enum. (IIRC avmplus uses a nullptr to represent an Any namespace and Any name)

core/src/avm2/globals/object.rs
```diff
@@ -274,15 +274,43 @@ pub fn create_class<'gc>(activation: &mut Activation<'_, 'gc>) -> GcCell<'gc, Cl
...
-        ("propertyIsEnumerable", property_is_enumerable),
+    let as3_instance_methods: Vec<(&str, NativeMethodImpl, _, _)> = vec![
+        // These signatures are weird, but they match the describeTypeJSON output
+        (
+            "hasOwnProperty",
+            has_own_property,
+            vec![ParamConfig::optional(
+                "name",
+                Multiname::any(activation.context.gc_context),
+                true,
```
adrian17, line 285: wait, the default value of the param is `true`? not `undefined`?

adrian17 (approved):
Approved after clearing the question c:

core/src/avm2/globals/object.rs
Aaron1011: oops - fixed

PR #13152, opened 2023-09-11
avm2: Remove PrimitiveObject and use Value as a receiver

Aaron1011:
This PR removes PrimitiveObject entirely - all primitives are now only ever accessed through a `Value`.

I've tried to perform the minimal amount of refactoring needed to support this. Unfortunately, that turns out to be a very large number of changes, since we now need to take `Value` instead of `Object` in a number of places.

Summary of the main changes:

* Rust-defined methods now take `this: Value<'gc>` instead of `this: Object<'gc>`. However, the rust methods defined for `native` ActionScript methods still take in `Object` - this is accomplished by generating a shim in `build_playerglobal`. However, all methods defined purely in Rust need to be modified (all primitive classes, plus some things like `Date` and `Vector`)
* Property-related methods (`get_property`, `call_property`, etc.) are now defined on `Value`. When called on a primitive, we lookup the vtable for the corresponding class definition (e.g. `String`), and run the rest of the property logic as normal.
* Several places now take in an `Activation`, so that we can lookup primitive classes.
* `coerce_to_object` now panics if called on a primitive. We should eventually replace all these calls by changing the callsite to:
  1. Explicitly panic on primitives, if this is a native method retrieivng a type-checked argument
  2. Call methods on `Value`, if `coerce_to_object` was being used in order to do property lookups / method calls
  3. Throw an explicit error, if we really need a non-primitive.

  All of the tests are passing, so there should be little risk
  of breaking SWFS while these panics are still in place.


core/src/avm2/value.rs
```diff
@@ -1305,4 +1331,708 @@ impl<'gc> Value<'gc> {
...
+    pub fn call(
+        self,
+        receiver: Value<'gc>,
+        arguments: &[Value<'gc>],
+        activation: &mut Activation<'_, 'gc>,
+    ) -> Result<Value<'gc>, Error<'gc>> {
+        if let Some(this) = self.as_object() {
+            this.call(receiver, arguments, activation)
+        } else {
+            Err("Primitive is not callable".into())
```
Lord-McSweeney, line 1843: Can this use the correct error?

core/src/avm2/object.rs
```diff
@@ -881,15 +586,45 @@ pub trait TObject<'gc>: 'gc + Collect + Debug + Into<Object<'gc>> + Clone + Copy
         multiname: &Multiname<'gc>,
         args: &[Value<'gc>],
         activation: &mut Activation<'_, 'gc>,
-    ) -> Result<Object<'gc>, Error<'gc>> {
+    ) -> Result<Value<'gc>, Error<'gc>> {
```
Lord-McSweeney, line 589: I don't think the primitive construction process is correct- Booleans should default to false if constructed without arguments, and I don't think Number/int/uint/String should do a (potentially-failing) coercion.

Lord-McSweeney:
This is probably going to require looking through all the uses of `(try_)get_object` to make sure none of them could potentially encounter a primitive.

core/src/avm2/value.rs
Aaron1011: None of our tests hit it, so this can be done in a follow-up

core/src/avm2/object.rs
Aaron1011: This logic was moved from the instance initializer for each primitive class - all of our tests are still passing

Lord-McSweeney (approved):
I'm a little concerned about all the shims (hopefully we can correctly replace them soon), but the logic looks sound.

core/src/avm2/activation.rs
```diff
@@ -2899,17 +2890,29 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
+            }
+            Value::Bool(_) => {
+                class == Some(self.avm2().classes().boolean)
+                    || class == Some(self.avm2().classes().object)
+            }
+            Value::String(_) => {
+                class == Some(self.avm2().classes().string)
+                    || class == Some(self.avm2().classes().object)
+            }
+            Value::Integer(_) => panic!("How do we do instanceof for integers?"),
```
Lord-McSweeney, line 2913: I'm pretty sure this is `class == Some(self.avm2().classes().object)`- the other number/int/uint classes return `false`.

Lord-McSweeney: For future reference, this was wrong- it returns true for both `Number` and `Object`, just like `Value::Number`

torokati44:
Superseded by https://github.com/ruffle-rs/ruffle/pull/18995.
Thank you anyway!

PR #13168, opened 2023-09-13
FLV Seeking

kmeisthax:
This turned out easier than I thought it would. Hopefully it actually works as easy as I think it should.


core/src/avm2/globals/flash/net/net_stream.rs
```diff
@@ -116,3 +116,20 @@ pub fn set_client<'gc>(
...
     Ok(Value::Undefined)
 }
+
+pub fn seek<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Object<'gc>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    if let Some(ns) = this.as_netstream() {
+        let offset = args
```
Lord-McSweeney, line 126: I think `args.get_f64(activation)?` would be preferred here

core/src/streams.rs
```diff
@@ -331,6 +338,151 @@ impl<'gc> NetStream<'gc> {
...
     }
 
+    pub fn time(self) -> f64 {
+        self.0.read().stream_time
+    }
+
+    /// Queue a seek to be executed on the next frame tick.
+    ///
+    /// `offset` is in milliseconds.
+    pub fn seek(self, context: &mut UpdateContext<'_, 'gc>, offset: f64) {
```
Lord-McSweeney, line 348: Can this have an extra argument `notify` (like with `fn pause`) that determines whether or not to broadcast the event, instead of relying on `context.is_action_script_3`?

Dinnerbone (approved):
LGTM! Nice work!

core/src/streams.rs
```diff
@@ -331,6 +341,152 @@ impl<'gc> NetStream<'gc> {
...
+                    reader
+                        .seek(SeekFrom::Start(old_position))
+                        .expect("valid backseek position");
+                }
+
+                let tag = tag.unwrap();
+                write.stream_time = tag.timestamp as f64;
+
+                if skipping_back && write.stream_time > offset
+                    || !skipping_back && write.stream_time < offset
```
torokati44, line 455: Shouldn't this have either a `>=` or a `<=`, or both?
In case of equality here, we can stop, no matter which direction are we skipping, no?

core/src/streams.rs
```diff
@@ -331,6 +341,152 @@ impl<'gc> NetStream<'gc> {
...
+                    ("code", "NetStream.SeekStart.Notify".into()),
+                ],
+            );
+        }
+    }
+
+    /// Seek to a new position in the stream.
+    ///
+    /// All existing audio will be paused. The stream offset will be snapped to
+    /// the prior keyframe. If the stream is playing then new tag processing
```
torokati44, line 374: > The stream offset will be snapped to the prior keyframe.

Is it always "prior", even when seeking forwards?
I'm not saying it's not, it's just not immediately obvious from the first reading.

core/src/streams.rs
torokati44: No, sorry, scratch that. The comparison is the other way around than how I interpreted it initially, and it's the condition for `continue`, not `break`. Pardon me, it was quite late...

core/src/streams.rs
kmeisthax: You're right, it goes to the next keyframe in the direction of the seek.

Issue #10510, opened 2023-04-01
The Rambling Wheels Pinball - unable to play


Kz-92:
### Describe the bug

![rwpb](https://user-images.githubusercontent.com/103858019/229307156-c1b16429-ee78-4520-8462-51446d13b768.png)

when playing the game, the tables take a long time to load, and then when the game starts, the paddles are stuck and the ball launcher has no ball/ is also stuck.

### Expected behavior

the game should be playable

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

[The_Rambling_Wheels_Pinballv2011.swf.zip](https://github.com/ruffle-rs/ruffle/files/11130235/The_Rambling_Wheels_Pinballv2011.swf.zip)

n0samu:
It seems like the XMLs are no longer parsing correctly, which makes the game load faster but also not work 😛 

This seems to be a regression caused by #12540. @evilpie, pinging to let you know.

Lord-McSweeney:
Scratch 2 seems to have a similar SVG error- not sure if it's a regression though.

evilpie:
I am working on this. We need to improve `XML.namespace()` to actually return at least some namespaces.

Daisyfan1995:
Fixed By https://github.com/ruffle-rs/ruffle/pull/13179

Daisyfan1995:
i checked but only in desktop app Works But in Ruffle Demo & Ruffle Extension (Online Web) with open swf player still continue this Bug.

PR #13179, opened 2023-09-13
avm2: Support node namespace in XML.namespace

evilpie:
Improves #10510 (playable but slow). We still don't support [[InScopeNamespaces]] and actual non-empty prefixes.

### DIFF
--- a/core/src/avm2/globals/XML.as
+++ b/core/src/avm2/globals/XML.as
@@ -8,7 +8,7 @@ package {
             stub_method("XML", "normalize");
             return this;
         }
-        
+
         AS3 static function setSettings(settings:Object = null): void {
             if (settings == null) {
                 settings = XML.AS3::defaultSettings();
@@ -60,7 +60,10 @@ package {
         AS3 native function hasSimpleContent():Boolean;
         AS3 native function name():Object;
         AS3 native function setName(name:*):void;
-        AS3 native function namespace(prefix:String = null):*;
+        private native function namespace_internal_impl(hasPrefix:Boolean, prefix:String = null):*;
+        AS3 function namespace(prefix:String = null):* {
+            return namespace_internal_impl(arguments.length > 0, prefix);
+        }
         AS3 native function localName():Object;
         AS3 native function toXMLString():String;
         AS3 native function child(name:Object):XMLList;
@@ -120,7 +123,7 @@ package {
 
         prototype.namespace = function(prefix:String = null):* {
             var self:XML = this;
-            return self.AS3::namespace(prefix);
+            return self.AS3::namespace.apply(self, arguments);
         }
 
         prototype.localName = function():Object {
@@ -205,7 +208,7 @@ package {
             var self:XML = this;
             return self.AS3::text();
         };
-        
+
         prototype.normalize = function():XML {
             var self:XML = this;
             return self.AS3::normalize();
--- a/core/src/avm2/globals/xml.rs
+++ b/core/src/avm2/globals/xml.rs
@@ -8,6 +8,7 @@ use crate::avm2::object::{
 };
 use crate::avm2::parameters::ParametersExt;
 use crate::avm2::string::AvmString;
+use crate::avm2::Namespace;
 use crate::avm2::{Activation, Error, Multiname, Object, Value};
 use crate::avm2_stub_method;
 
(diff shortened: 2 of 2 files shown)


core/src/avm2/globals/xml.rs
```diff
@@ -128,15 +129,60 @@ pub fn set_name<'gc>(
...
 
-pub fn namespace<'gc>(
+pub fn _namespace<'gc>(
     activation: &mut Activation<'_, 'gc>,
-    _this: Object<'gc>,
-    _args: &[Value<'gc>],
+    this: Object<'gc>,
+    args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
     // FIXME: Implement namespace support (including prefix)
```
Lord-McSweeney, line 137: I think this FIXME can be removed now, since you added more detailed FIXMEs

core/src/avm2/globals/xml.rs
```diff
@@ -128,15 +129,60 @@ pub fn set_name<'gc>(
     Ok(Value::Undefined)
 }
 
-pub fn namespace<'gc>(
+pub fn _namespace<'gc>(
```
Lord-McSweeney, line 132: Non-blocking, but this name makes it sound like the function is unused. Maybe `internal_get_namespace` (or something similar)? Alternatively, we could keep it entirely implemented in Rust, and base the behavior off of `args.len()`.

Lord-McSweeney (approved):

core/src/avm2/globals/xml.rs
```diff
@@ -128,15 +129,60 @@ pub fn set_name<'gc>(
...
+        // b. Find a Namespace ns ∈ inScopeNS, such that ns.prefix = prefix. If no such ns exists, let ns = undefined.
+        // c. Return ns
+
+        // FIXME: Nodes currently either have zero or one namespace, which has the prefix "" (empty string)
+        Ok(match node.namespace() {
+            Some(ns) if prefix.is_empty() => {
+                let namespace = Namespace::package(ns, &mut activation.context.borrow_gc());
+                NamespaceObject::from_namespace(activation, namespace)?.into()
+            }
+            _ => Value::Undefined,
```
Lord-McSweeney, line 183: Did you test this case? From my testing, this returns `null` in FP, not `undefined`.

evilpie: It doesn't in my testing:

```
var xml:XML = <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://foobar.com">
    <path fill="none" />
</svg>

var ns1 = xml.namespace();
trace("xml.namespace(): " + ns1);

var ns2 = xml.namespace("");
trace('xml.namespace(""): ' + ns2);

var ns3 = xml.namespace("unknown");
trace('xml.namespace("unknown"): ' + ns3); // << undefined
```

core/src/avm2/globals/XML.as
```diff
@@ -60,7 +60,10 @@ package {
         AS3 native function hasSimpleContent():Boolean;
         AS3 native function name():Object;
         AS3 native function setName(name:*):void;
-        AS3 native function namespace(prefix:String = null):*;
+        private native function namespace_internal_impl(hasPrefix:Boolean, prefix:String = null):*;
+        AS3 function namespace(...args):* {
```
adrian17, line 64: we should keep the FP-compatible signature and instead check the magic `arguments` variable's length.

core/src/avm2/globals/XML.as
```diff
@@ -118,9 +121,13 @@ package {
             return self.AS3::name();
         };
 
-        prototype.namespace = function(prefix:String = null):* {
+        prototype.namespace = function(...args):* {
```
adrian17, line 124: same here, you could keep the correct signature and pass the `arguments` via `apply()`.

Lord-McSweeney: I can call `XML.namespace()` with any number of arguments in FP, so I'm not sure what you're stating is correct

adrian17: I don't know if we reproduce this correctly in Ruffle, but the presence of `arguments` prevents too-many-args error in FP.

```as
class C{ public static function f(a: String) { } }
var f = C.f;
f(); // error
f("a"); // ok
f("a", "b", "c"); // error

class C{ public static function f(a: String=null) {  } }
var f = C.f;
f(); // ok
f("a"); // ok
f("a", "b", "c"); // error

class C{ public static function f(a: String=null) { trace(arguments) } }
var f = C.f;
f(); // ok
f("a"); // ok
f("a", "b", "c"); // ok
```

adrian17: (also unrelated to this topic, but for completeness: a fully untyped function like `var f = function(a){}` also skips argument count checks. Adding a default `a=null` or type `a: String` reenables it. I don't know the exact rules here though.)

adrian17: So I think `arguments` is the only way we make it "variadic" while keeping `describeType`-compatible signature.

Lord-McSweeney: https://github.com/ruffle-rs/ruffle/pull/13179#discussion_r1328458053 this sounds like a function verification thing

adrian17: > This sounds like a function verification thing, which we haven't implemented.

I think we have this implemented? Didn't test but we have this logic:
```
        let has_rest_or_args = method.is_variadic(); // <- true if uses `arguemnts`

        if user_arguments.len() > signature.len() && !has_rest_or_args {
            return Err(format!(
                "Attempted to call {:?} with {} arguments (more than {} is prohibited)",
```

Lord-McSweeney: Whoops, my mistake. Apparently this exists: `AbcMethodFlags::NEED_ARGUMENTS`

evilpie: I don't think any of the correct signature or the number of arguments is really going to matter in practice. Considering that you have to do a bunch of tricks to even bypass the arguments length check of the compiler, I would just go for the correct signature?

Honestly I don't have any great insight here :/

adrian17 (approved):
LGTM. Some tests would be great too, but I'm gonna merge without them.

PR #13180, opened 2023-09-13
avm2: Add support for Stage3D bytearray/compressed textures

Aaron1011:
'Compressed' textures are Adobe Texture Format (ATF) wrappers around the obscure JPEG-XR image format. To implement this, I've forked the Rust 'jpegxr' crate, and added in a method to convert a JPEG-XR image to a TIFF (which can then be handled by the standard 'image' crate).

The 'jpegxr' crate wraps the C 'jxrlib' library. In order for this to work on web, I've modified the crate to include a 'fakelib' impl with stubs and copy-pasted musl method for various libc methods (based on szolo's previous work).

The jpexgr changes can be reviewed at https://github.com/ruffle-rs/jpegxr/tree/ruffle


core/src/avm2/globals/flash/display3D/textures/texture.rs
```diff
@@ -1,8 +1,200 @@
...
+
+pub fn upload_compressed_texture_from_byte_array<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Object<'gc>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let texture = this.as_texture().unwrap();
+    let data = args.get_object(activation, 0, "data")?;
+    let byte_array_offset = args.get_u32(activation, 1)? as usize;
+    let async_ = args.get_bool(2);
```
Lord-McSweeney, line 160: Should this be `_async` instead of `async_`?

core/src/avm2/globals/flash/display3D/textures/texture.rs
```diff
@@ -1,8 +1,200 @@
...
+
+pub fn do_copy<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    data: Object<'gc>,
+    texture: TextureObject<'gc>,
+    byte_array_offset: u32,
+    side: u32,
+    mip_level: u32,
+) -> Result<(), Error<'gc>> {
+    //eprintln!("uploadFromByteArray texture {:?} with miplevel {mip_level}", texture.handle());
```
Lord-McSweeney, line 26: I think this comment can be removed

core/src/avm2/globals/flash/display3D/textures/texture.rs
Aaron1011: It's not really unused, but it conflicts with the `async` keyword

render/src/utils.rs
```diff
@@ -155,7 +155,7 @@ fn validate_size(width: u16, height: u16) -> Result<(), Error> {
 
 /// Decodes a JPEG with optional alpha data.
 /// The decoded bitmap will have pre-multiplied alpha.
-fn decode_jpeg(jpeg_data: &[u8], alpha_data: Option<&[u8]>) -> Result<Bitmap, Error> {
+pub fn decode_jpeg(jpeg_data: &[u8], alpha_data: Option<&[u8]>) -> Result<Bitmap, Error> {
```
Lord-McSweeney, line 158: Does this need to be `pub`?

render/src/atf.rs
```diff
@@ -0,0 +1,119 @@
...
+impl ATFTexture {
+    pub fn from_bytes(mut bytes: &[u8]) -> Result<Self, Box<dyn std::error::Error>> {
+        // Based on https://github.com/openfl/openfl/blob/develop/src/openfl/display3D/_internal/ATFReader.hx
+        let bytes = &mut bytes;
+
+        let mut string_bytes = vec![0; 3];
+        bytes.read_exact(&mut string_bytes)?;
+        let signature = String::from_utf8(string_bytes)?;
+
+        if signature != "ATF" {
```
adrian17, line 39: This doesn't need a Vec or String at all, right?
```rs
    let mut string_bytes = [0; 3];
    bytes.read_exact(&mut string_bytes);
    if &string_bytes != b"ATF" {
```

render/src/atf.rs
```diff
@@ -0,0 +1,119 @@
...
+                    ATFFormat::CompressedLossyAlpha => 17,
+                };
+                let mut all_data = vec![];
+                for _ in 0..num_records {
+                    let len = if version == 0 {
+                        read_uint24(bytes)?
+                    } else {
+                        bytes.read_u32::<BigEndian>()?
+                    };
+                    // FIXME - make this more efficient
```
adrian17, line 95: hmm, there's no `.read_to_vec(len, &mut all_data)`? :(

adrian17 (approved):
LGTM, with the noted comments and assuming the current overhead is <= 0.5MB as reported on Discord.

render/src/atf.rs
Aaron1011: I changed this to first resize `all_data`, and then call `read_exact` with a slice of it.

PR #1089, opened 2020-09-02
Change build badges to github actions

Justin-CB:
The easy half of #584

PR #13209, opened 2023-09-16
avm2: Support Error #1089

Lord-McSweeney:


evilpie:
Interesting, ECMA-357 doesn't seem to call for this to throw.

Lord-McSweeney:
Maybe FP does it to ensure a set doesn't silently fail on an XMLList with multiple elements?

adrian17 (approved):

PR #13212, opened 2023-09-16
avm2: Minor fix to string_wrapping_index and String.substr

Lord-McSweeney:
This improves the `ecma3/String/substr` test, which next needs proper method signatures.


Lord-McSweeney:
I still need to add some proper tests for this.

core/src/avm2/globals/string.rs
```diff
@@ -529,12 +529,14 @@ fn substr<'gc>(
     let len = if len < 0. {
         if len.is_infinite() {
             0.
-        } else {
+        } else if len <= -1.0 {
```
adrian17, line 532: As said on discord: I think having this be `if (len as isize) < 0`... or casting `len` even earlier so we don't need to do `(len as isize)` again later sounds nicer here. I think FP logic would do the latter, make the cast earlier and have the indexing logic work on ints.

(There's also a question of whether the same should apply to Vector methods, as they have a negative index logic too, but that's a side note, out of scope)

Lord-McSweeney: I prefer `if len <= -1.0`, it makes more sense at a glance.

adrian17 (approved):

Issue #12008, opened 2023-07-12
Mega Miner: Invisible buttons


n0samu:
### Describe the bug

Link to SWF: http://i.notdoppler.com/files/megaminer.swf

[Mega Miner](http://www.notdoppler.com/megaminer.php) is playable now, but the Play button to start the game is invisible:
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/7e49a404-61ad-4494-a16d-3fcc6f9b040a)
Clicking its area goes to the main menu, from which you can click the invisible menu buttons and start the game. From that point on the game works basically fine, but it continuously spams "Display object container has no child with name" warnings as noted in https://github.com/ruffle-rs/ruffle/issues/9956#issuecomment-1596207531.

### Expected behavior

The buttons should be visible (and Ruffle probably shouldn't be spamming warnings; it may be related).

### Affected platform

Desktop app

### Operating system

Windows 10

n0samu:
These buttons use TLF text (DefineFont4), which is unimplemented in Ruffle.

n0samu:
With #13957, the main menu button text is visible (though it doesn't use the correct font yet). The Play button is still mostly invisible though.

evilpie:
See  #14749. The are still some rendering artifacts, but the button is definitely not invisible anymore.

n0samu:
Fixed by #14749

Issue #7866, opened 2022-09-04
https://z0r.de/3845 doesn't show a few features


Diatonator:
### Describe the bug

https://z0r.de/3845
This SWF seems to mostly work, but there are two strings of text absent compared to Adobe playback.

There are errors in the console:
![image](https://user-images.githubusercontent.com/75184268/188329707-f7a60ba6-cbfd-436a-a3e9-0c8736bd8dbd.png)

Nightly 2022-09-04

### Expected behavior

Those are the things absent from the Ruffle playback:
![image](https://user-images.githubusercontent.com/75184268/188329755-ec9ca0e2-25c7-4c2a-86d0-b383ee2b5545.png)

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Chrome 105.0.5195.102

Lord-McSweeney:
Needs `XML.settings`.

Lord-McSweeney:
Newest error:
```
 Got "ReferenceError: Error #1069: Property http://adobe.com/AS3/2006/builtin::hasOwnProperty not found on Class and there is no default value.
	at fl.motion::AnimatorFactoryBase/addTargetInfo()
	at _fla::btn_3()
	at flash.display::MovieClip/flash::display::MovieClip::gotoAndPlay()
	at _fla::MainTimeline/onEnterFrames()" when constructing AVM2 side of movie clip of type _fla::btn_3
```

Lord-McSweeney:
This now needs support for TLF text.

Lord-McSweeney:
This also has the `getFirstLeaf` error.

evilpie:
Currently the text isn't displayed correctly, because we can't load Times New Roman and it overflow the bounding box with the default font.

Issue #9232, opened 2023-01-20
Pony Creator: Start up freak out


waspennator:
### Describe the bug

https://www.numuki.com/game/mlp-pony-creator/
Game loads up to the preloader fine, but as soon as you click play, the game freaks out trying to load stuff in while spamming this in the console log. The buttons on top seem to work and the customization buttons with drop down menus that show up on the top left under create, accessorize, and pose, can also be clicked, but everything else doesn't seem to work.

AVM2 error: Object(ErrorObject(ErrorObject { ptr: 0x89e5988 }))

ReferenceError: Error # 1069: Property invalidate not found on flash.display.Stage and there is no default value.

Got AvmError(Object(ErrorObject(ErrorObject { ptr: 0x881f700 }))) when constructing AVM2 side of movie clip of type fl.controls::Button

![Pony Creator Ruffle](https://user-images.githubusercontent.com/78768499/213614458-fe2b32f0-46b3-4b84-9372-7dd42f03a296.png)

### Expected behavior

It should be loading up like this once you press play.

![Pony Creator Flash](https://user-images.githubusercontent.com/78768499/213614447-8cf6b34b-d90e-4478-a7d5-aaefe18f8327.png)

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 109

MartySVK:
Same as https://github.com/ruffle-rs/ruffle/issues/9131

waspennator:
Retested again and the issue is still happening, but there seem to be some different errors in the log this time.

[www.numuki.com-1676001871721.log](https://github.com/ruffle-rs/ruffle/files/10703585/www.numuki.com-1676001871721.log)

Lord-McSweeney:
This appears to need TLF text and XML.

waspennator:
Update, for some reason the damn site swapped over to waflash without warning, so i'm gonna slap a physical copy of the SWF here for easier testing.

[pony-creator.zip](https://github.com/ruffle-rs/ruffle/files/11007606/pony-creator.zip)

waspennator:
Flickering is now gone and its a step in the right direction, but it still doesn't seem to work.

Got "TypeError: Error # 1009: Cannot access a property or method of a null object reference. (accessing field: getFirstLeaf)" when constructing AVM2 side of movie clip of type fullfinal6_fla::help_link_100

[...]

PR #13226, opened 2023-09-18
avm2: Basic TLF support

Lord-McSweeney:
TLF was Adobe's more complex text rendering system. It supports adjustment of the baseline, easy grouping of graphics within text, and more advanced features that aren't available with `TextField`.

The rendering interface is done through the `TextLine` class, which extends `DisplayObjectContainer`. In this implementation, `TextLine` is a StageObject linked to an `EditText`.

Creating `TextLines` is done by creating a `TextBlock`, setting its `.content` to a `ContentElement` (e.g. `TextElement`), and then calling `createTextLine` on the `TextBlock`.

Unfortunately, I don't think Ruffle has a way to calculate the number of glyphs that fit within a certain width, nor can it calculate text justification, so we only create a single TextLine that contains the entire content of the `ContentElement`, and the height of the TextLine depends on the height of the text itself. We don't support calculating this either, so I just estimate it.

Combined with #13179, this improves rendering of #12008 (main menu buttons show up, but play button text is cut off), and allows the text in #7866 to render (but it's the wrong size, so it gets cut off offscreen).

This doesn't allow text to render in any of the Weed to Wonder SWFs, which use GroupElement instead of TextElement.


core/src/avm2/globals/flash/text/engine/TextBlock.as
```diff
@@ -48,83 +51,106 @@ package flash.text.engine {
...
+        public function get textLineCreationResult():String {
+            return this._textLineCreationResult;
+        }
+
+        public function get firstLine():TextLine {
+            return this._firstLine;
+        }
+
+        public function get lastLine():TextLine {
+            return this._firstLine;
```
adrian17, line 144: is this a copy error, or an intentional simplification? If the latter, can you add a comment, and maybe the stub marker?

Lord-McSweeney:
I still need to add tests for functionality (non-visual tests).

Lord-McSweeney:
I stubbed `InteractiveObject.accessibilityImplementation` because some `flashx.textLayout.compose` needs it to work.

Lord-McSweeney:
I'm unsure of what's going on with #9232. The game seems to be storing the length of the current text of a SpanElement, and calling `replaceText` with `(0, [that length], "whatever")`, but for some reason, the current length (`FlowElement._textLength` in the SWF) isn't updated after `replaceText` is called.

core/src/avm2/globals/flash/text/engine/text_block.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    display_object.set_text(text.as_wstr(), &mut activation.context);
+
+    let element_format = content
+        .get_public_property("elementFormat", activation)?
+        .as_object();
+
+    apply_format(activation, display_object, text.as_wstr(), element_format)?;
+
+    let instance = initialize_for_allocator(activation, display_object.into(), class)?;
```
adrian17, line 74: (copied from Discord)
I'd move this code next to `text_field_allocator`, smth like
```rs
/// Used for a fake TextLine class, that uses an EditText internally.
pub fn tlf_text_line_allocator<'gc>(
    class: ClassObject<'gc>,
) -> Result<Object<'gc>, Error<'gc>> {
    let textfield_cls = activation.avm2().classes().textline;

    let movie = activation.context.swf.clone();
    let display_object = EditText::new_tlf(&mut activation.context, movie, 0.0, 0.0, width, 15.0).into();
    let obj = initialize_for_allocator(activation, display_object, textfield_cls);
    (display_object, obj)
}
```

core/src/avm2/globals/flash/text/engine/text_block.rs
```diff
@@ -0,0 +1,147 @@
...
+    class.call_native_init(instance.into(), &[], activation)?;
```
adrian17, line 75: BTW I don't think `call_native_init` actually does anything here, right?

Lord-McSweeney: No, I included it because I saw other code do the same thing, but this feels more correct, since we should be calling the `native_init` when constructing a class. If we ever actually do something in the `native_init`, this will work correctly.

core/src/avm2/globals/flash/text/engine/text_block.rs
Lord-McSweeney: `TextLine` should actually throw in its allocator, similarly to `Worker` and `MessageChannel`, so I'm not sure that's the right place to put it. I think this would also result in the format code being extracted to the same file as `tlf_text_line_allocator`, and the format code is a bit...bulky.

core/src/avm2/globals/flash/display/InteractiveObject.as
```diff
@@ -1,13 +1,25 @@
...
+
 		public function InteractiveObject() {
 			throw new Error("You cannot directly construct InteractiveObject.")
 		}
 
+		public function get accessibilityImplementation():AccessibilityImplementation {
+		    return this._accessibilityImpl;
+		}
+		public function set accessibilityImplementation(value:AccessibilityImplementation):void {
+		    stub_setter("flash.display.InteractiveObject", "accessibilityImplementation");
```
Dinnerbone, line 19: Is this a stub if it's actually doing something?

Dinnerbone (approved):
LGTM, though in my head I imagined that TLF is basically exposing our internal layout classes and adding more functionality on top of them. I suspect that TLF is just "the stuff that happens automatically when you assign .text" (`relayout()` for us), but explicit and the onus moves onto AS instead of all being internal - and therefor giving them more power to do what they want with it, too

core/src/avm2/globals/flash/display/InteractiveObject.as
Lord-McSweeney: No, this isn't doing anything. Calling the setter should update the accessibility implementation (unlike `DisplayObject.accessibilityProperties`, which you must call `Accessibility.updateProperties()` to update)- this setter doesn't do anything right now.

evilpie:
#12054 also needs `flash.text.engine.TextJustifier: lineJustification` and `flash.text.engine.FontDescription: static isFontCompatible()`.

Dinnerbone:
image
This makes Gaia chat work :)

PR #13237, opened 2023-09-18
web: Only re-run polyfill when embed or object added to DOM

danielhjacobs:


torokati44:
Sorry if these questions are too trivial, but just to be sure:
 - Does `mutation.addedNodes` contain every single added node, even the children of those added directly?
 - Is the node name always UPPERCASE?

danielhjacobs:
>     * Does `mutation.addedNodes` contain every single added node, even the children of those added directly?
Apparently not, see https://stackoverflow.com/questions/61314922/mutationobserver-not-picking-up-on-child-nodes. I needed to use the [TreeWalker API](https://developer.mozilla.org/en-US/docs/Web/API/TreeWalker) as well.
>     * Is the node name always UPPERCASE?
Yes, for all HTML elements: https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeName#value

web/packages/core/src/polyfills.ts
```diff
@@ -189,11 +189,25 @@ async function injectRuffle(
...
     const observer = new MutationObserver(function (mutationsList) {
-        // If any nodes were added, re-run the polyfill to detect any new instances.
-        const nodesAdded = mutationsList.some(
-            (mutation) => mutation.addedNodes.length > 0,
+        // If any embed or object nodes were added, re-run the polyfill to detect any new instances.
+        const elements = ["EMBED", "OBJECT"];
+        const embedOrObjectAdded = mutationsList.some((mutation) =>
+            Array.from(mutation.addedNodes).some(
+                (node) =>
+                    elements.includes(node.nodeName) ||
```
danielhjacobs, line 197: This is needed because of https://stackoverflow.com/questions/28699733/why-does-treewalker-currentnode-default-to-root-node

web/packages/core/src/polyfills.ts
```diff
@@ -189,11 +189,25 @@ async function injectRuffle(
...
+                    document
+                        .createTreeWalker(
```
danielhjacobs, line 199: This is needed because of https://stackoverflow.com/questions/61314922/mutationobserver-not-picking-up-on-child-nodes

adrian17 (approved):

PR #13030, opened 2023-08-31
web: Fix 100% height when that equals 0

danielhjacobs:
`200px` is what it is on Pale Moon. Other browsers may vary.

Main Test (200px height) :
```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="utf-8">
  <title>Ruffle</title>
</head>
<body>
  <embed width="100%" height="100%" src="https://cdn.jsdelivr.net/gh/ruffle-rs/demo/logo-anim.swf"></embed>
</body>
</html>
```

Other tests:

50px height:
```html
<embed width="100%" height="50px" src="https://cdn.jsdelivr.net/gh/ruffle-rs/demo/logo-anim.swf"></embed>
```
0px height:
```html
<embed width="100%" height="0px" src="https://cdn.jsdelivr.net/gh/ruffle-rs/demo/logo-anim.swf"></embed>
```
0px height:
```html
<body style="height: 0px">
  <embed width="100%" height="100%" src="https://cdn.jsdelivr.net/gh/ruffle-rs/demo/logo-anim.swf"></embed>
</body>
```
50px height:
```html
<body style="height: 50px">
  <embed width="100%" height="100%" src="https://cdn.jsdelivr.net/gh/ruffle-rs/demo/logo-anim.swf"></embed>
</body>
```
200px height:
```html
<body style="height: 0px">
  <div>
    <embed width="100%" height="100%" src="https://cdn.jsdelivr.net/gh/ruffle-rs/demo/logo-anim.swf"></embed>
  </div>
</body>
```

Test on Flash browser

[Screencast from 2023-09-08 15-54-25.webm](https://github.com/ruffle-rs/ruffle/assets/29206584/b85ba861-ba6c-48b6-b33e-098a45d6affd)

Test using this PR:

[Screencast from 2023-09-08 15-55-28.webm](https://github.com/ruffle-rs/ruffle/assets/29206584/57eb64de-9238-4edd-8a6a-826a634a7081)

PR #13256, opened 2023-09-20
web: As appropriate use embed height mentioned in Mozilla bug 810082 c65

danielhjacobs:
See https://bugzilla.mozilla.org/show_bug.cgi?id=810082#c65, the test file, and https://github.com/ruffle-rs/ruffle/pull/13030#issuecomment-1716314418. Supersedes #13030.

### DIFF
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -177,6 +177,7 @@ export class RufflePlayer extends HTMLElement {
     private pointerMoveMaxDistance = 0;
 
     private volumeSettings: VolumeControls;
+    private resizeObserver: ResizeObserver;
 
     /**
      * Triggered when a movie metadata has been loaded (such as movie width and height).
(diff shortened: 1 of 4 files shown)


Dinnerbone (approved):
Nice work!

Benman2785:
this is bugged! (or my html that worked before is wrong)

i now have bugs in firefox and edge and chrome

i am now FORCED to specify height in px and cant use 100%

just check: https://hoffmeister.li/flash-games/
image

Benman2785:
https://hoffmeister.li/flash-games/index_test.html shows old behaviour

Benman2785:
can you fix your code to make it work again?

Benman2785:
https://hoffmeister.li/flash-games/The_last_Stand_2_2023-10-04.html
this logic is flawed even if not in iFrame...

danielhjacobs:
You're right, I found the real issue, and it's something I had actually noticed before. This logic only applies if the Flash embed/object is on a page or in an iframe with `<!DOCTYPE html>`. The sizing would be accurate to Flash if you added that to the game HTML page.

Benman2785:
thats not the problem:
https://hoffmeister.li/flash-games/The_last_Stand_2_2023-10-04.html
error persists

danielhjacobs:
That's not what I meant. I meant this bug exists in Ruffle because this matches Flash's behavior when using `<!DOCTYPE html>`.

Benman2785:
working ;)

PR #13284, opened 2023-09-23
avm2: Namespace improvements

evilpie:


Lord-McSweeney:
We should probably fix our weird `undefined` QName constructor implementation- in Ruffle, `new QName(undefined)` is treated the same as `ne w QName()` and `new QName(undefined, undefined)`, which FP handles correctly.

evilpie:
> We should probably fix our weird `undefined` QName constructor implementation- in Ruffle, `new QName(undefined)` is treated the same as `ne w QName()` and `new QName(undefined, undefined)`, while FP handles those two cases correctly.

As far as I can tell we match Flash for both `new QName(undefined)` and new `QName(undefined, undefined)` now.

Aaron1011 (approved):

adrian17 (approved):

evilpie:
Thanks for merging. For reference #13358 has some tests that shows some of the mistakes in Ruffle's QName implement.

PR #13302, opened 2023-09-24
wgpu: Rename DepthBuffer to StencilBuffer and stop providing depth_ops

Aaron1011:
This is a stencil-only texture, not a depth buffer. Setting `depth_ops` to `None` also makes us compatible with WebGPU (the wgpu crate doesn't yet validate this).


adrian17 (approved):

danielhjacobs:
We have a lot of "Shader blend should not need stencil buffer" error reports: https://github.com/search?q=repo%3Aruffle-rs%2Fruffle%20%22Shader%20blend%20should%20not%20need%20stencil%20buffer%22&type=issues

danielhjacobs:
They seem to all occur in Character Creator code (/cc or /character/creator seems to be in every URL)

danielhjacobs:
Sorry, that was added in https://github.com/ruffle-rs/ruffle/pull/12238, this just renamed it, as pointed out by kjarosh.

Issue #13130, opened 2023-09-10
AdventureQuest: Loading shops causes massive CPU spikes


secret-shober:
### Describe the bug

When loading a shop, this can cause a massive spike in CPU usage. When done enough times (such as farming void bosses) this can cause a crash as the following error stack describes:
```
# Error Info
Error name: Error
Error message: panicked at 'assertion failed: `(left == right)`
  left: `true`,
 right: `false`: cannot recursively acquire mutex', /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/std/src/sys/wasm/../unsupported/locks/mutex.rs:20:9
Error stack:
Error: panicked at 'assertion failed: `(left == right)`
  left: `true`,
 right: `false`: cannot recursively acquire mutex', /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/std/src/sys/wasm/../unsupported/locks/mutex.rs:20:9
    at imports.wbg.__wbg_new_d258248ed531ff54 (chrome-extension://donbcfbmhbcapadipfkeojnmajbakjdc/dist/339.js:2414:21)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::h2e7c106a4724653e (wasm://wasm/030be422:wasm-function[9327]:0x925500)
    at std::panicking::rust_panic_with_hook::h7601402c0a383194 (wasm://wasm/030be422:wasm-function[6094]:0x824f58)
    at core::panicking::panic_fmt::h6dad0405f48e39e2 (wasm://wasm/030be422:wasm-function[9163]:0x920b12)
    at core::panicking::assert_failed_inner::hb22561d3158be3ea (wasm://wasm/030be422:wasm-function[4039]:0x718721)
    at core::panicking::assert_failed::hc43d40ad7edb3a9d (wasm://wasm/030be422:wasm-function[8167]:0x8e1844)
    at ruffle_core::loader::Loader::movie_loader::{{closure}}::h8ae296ad3ac80621 (wasm://wasm/030be422:wasm-function[713]:0x238b0e)
    at <ruffle_web::navigator::WebNavigatorBackend as ruffle_core::backend::navigator::NavigatorBackend>::spawn_future::{{closure}}::hd49960977bc13f74 (wasm://wasm/030be422:wasm-function[1180]:0x3bbbd1)
[...]

PR #13319, opened 2023-09-25
avm1: Store childNodes on the node itself

adrian17:
I didn't do benchmarks yet. I can also add a test later.
This should help https://github.com/ruffle-rs/ruffle/issues/13130 .
In general, this should make repeated `.childNodes` a free op, at the cost of making mutations (`.appendChild` etc) more expensive (`O(n_children)`* but only if `.childNodes` was read at least once before).
This is not just an optimization, but also a correctness fix, as the array mutations can be observed in AS.

(*which implies that appending N elements is `O(n^2)`. I confirmed that FP has the same quadratic time complexity, though without benchmarks I can't tell whether we'll turn out faster or slower)

Unfortunately it's ugly due to current design. Ideally we'd want to refresh the array each time a mutation of a `XmlNode` (`xml/tree.rs`) happens, but this refresh is an AVM1 operation that technically can throw an AVM1 error, which makes actually calling this from `tree.rs` much harder.
Any advice? 😅


core/src/avm1/globals/xml.rs
```diff
@@ -222,6 +222,8 @@ impl<'gc> Xml<'gc> {
             }
         }
 
+        self.root().refresh_cached_child_nodes(activation).unwrap(); // :(
```
Lord-McSweeney, line 225: I don't think the AVM1 operations done by `refresh_cached_child_nodes` can throw an error?

Dinnerbone (approved):

PR #4584, opened 2021-06-16
avm1: Implement onDragOver, onDragOut, and _droptarget

Herschel:
* Implement the drag button states, triggered by clicking + holding on a button and dragging out and back over the button. (`onDragOver`, `onDragOut`, `onReleaseOutside`).
 * Improve general button state behavior when clicking and dragging.
 * Implement `_droptarget` property to return info about where an object is dragged.

Fixes #795, #1451, #1828, #2515, #2833, #3070, #3139, #3183, #3411, #3690, #3758, probably more.
Possibly fixes #3191?

PR #13330, opened 2023-09-27
core: Fix for mouse drag behavior

### LINKED ISSUE #13327: game not registering the mouse button being held down
### Describe the bug

The game responds appropriately to pressing the mouse button, but it doesn't register when the player is holding it down. 
[ark_of_treasure.zip](https://github.com/ruffle-rs/ruffle/files/12731728/ark_of_treasure.zip)

### Expected behavior

In this game, the player is expected to hold down their mouse button to select adjacent items that are identical in color. As the button is pressed down, a counter appears, showing the number of the selected items.

ark_of_treasure_highlight

### Affected platform

Desktop app

### Operating system

Windows 8.1

### Browser

Mozilla Firefox 115.0.2

### Additional information

I downloaded the .swf file from an online hosted version of the game, but the issue persists over various websites and the desktop app as well.

Kamina-sama:
Fixes #13327. Have yet to test on other games.

### DIFF
--- a/core/src/player.rs
+++ b/core/src/player.rs
@@ -1380,31 +1380,12 @@ impl Player {
                 // and fires "drag" events. Other objects are ignored.
                 if context.input.is_mouse_down() {
                     context.mouse_over_object = new_over_object;
-                    if let Some(down_object) = context.mouse_down_object {
-                        if InteractiveObject::option_ptr_eq(
-                            context.mouse_down_object,
-                            cur_over_object,
-                        ) {
-                            // Dragged from outside the clicked object to the inside.
-                            events.push((
-                                down_object,
-                                ClipEvent::DragOut {
-                                    to: new_over_object,
-                                },
-                            ));
-                        } else if InteractiveObject::option_ptr_eq(
-                            context.mouse_down_object,
-                            new_over_object,
-                        ) {
-                            // Dragged from inside the clicked object to the outside.
-                            events.push((
-                                down_object,
-                                ClipEvent::DragOver {
-                                    from: cur_over_object,
-                                },
-                            ));
-                        }
-                    }
+                    cur_over_object.map(|old_hovered_object| {
+                        events.push((old_hovered_object, ClipEvent::DragOut { to: new_over_object }));
+                    });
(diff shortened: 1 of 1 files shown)


Lord-McSweeney:
Thank you for contributing! Please run `cargo fmt` in your ruffle directory to fix the failing format tests.

Kamina-sama:
@Lord-McSweeney Got it.

core/src/player.rs
```diff
@@ -1380,30 +1380,21 @@ impl Player {
                 // and fires "drag" events. Other objects are ignored.
                 if context.input.is_mouse_down() {
                     context.mouse_over_object = new_over_object;
```
Kamina-sama, line 1382: By the way, this feels odd, shouldn't this value be updated regardless of mouse input? I might just not be seeing something here, so correct me if I'm wrong.

n0samu:
It looks like the code you changed was originally written to implement drag and drop events in AVM1 (#4584). I wonder if these events are different in AVM1 and AVM2. We need to be careful about changing core code because we don't want to regress AVM1.

Kamina-sama:
> It looks like the code you changed was originally written to implement drag and drop events in AVM1 (#4584). I wonder if these events are different in AVM1 and AVM2. We need to be careful about changing core code because we don't want to regress AVM1.

Understood, I will test further on some of those games.

n0samu:
> Understood, I will test further on some of those games.

Great! Particularly the games linked to that pull request would be great to test for regressions.

Kamina-sama:
@n0samu  Incriminati, Belly Biff, 3 in 1 checkers, koala checkers, tekken-miguel-caballero-rojo are okay.

Lord-McSweeney:
From a basic look at the code, this looks wrong. The original code made drag events only fire when the old mousedown object was the same as the new mousedown object- this code doesn't check that and fires them regardless. Are you sure this matches Flash Player?

torokati44:
This was replaced by https://github.com/ruffle-rs/ruffle/pull/21661 according to https://github.com/ruffle-rs/ruffle/pull/21661#issuecomment-3304633449. Thank you anyway!

PR #13348, opened 2023-09-28
core: Place jpegxr texture support behind feature flag

sleepycatcoding:
This probably needs changes to the workflow files, but I am not familiar with them so I would need some help.


desktop/Cargo.toml
```diff
@@ -60,6 +60,7 @@ avm_debug = ["ruffle_core/avm_debug"]
 lzma = ["ruffle_core/lzma"]
 software_video = ["ruffle_video_software"]
 tracy = ["tracing-tracy", "ruffle_render_wgpu/profile-with-tracy"]
+jpegxr = ["ruffle_core/jpegxr"]
```
Aaron1011, line 63: Can you enable this by default?

web/Cargo.toml
```diff
@@ -54,7 +54,7 @@ async-channel = "1.9.0"
 [dependencies.ruffle_core]
 path = "../core"
 default-features = false
-features = ["audio", "mp3", "nellymoser", "wasm-bindgen", "default", "default_compatibility_rules"]
+features = ["audio", "mp3", "nellymoser", "wasm-bindgen", "default", "default_compatibility_rules", "jpegxr"]
```
Aaron1011, line 57: same here

Aaron1011 (approved):

PR #13350, opened 2023-09-28
swf: Fix writing of an all-zeros curve ShapeRecord

adrian17:
I'm gonna think about a possible test later :)


adrian17:
Switched to draft, as the written SWF is still bad.

adrian17:
OK, now the sample SWF roundtrips into a valid SWF.

adrian17:
This is now reviewable.

Dinnerbone (approved):

PR #13357, opened 2023-09-29
web: Display a pop-up about hardware acceleration when it's disabled

danielhjacobs:
image

Edit: This is the original screenshot, new Screencast is here: https://github.com/ruffle-rs/ruffle/pull/13357#discussion_r1342766440

### DIFF
--- a/web/packages/core/src/shadow-template.ts
+++ b/web/packages/core/src/shadow-template.ts
@@ -321,7 +321,7 @@ export function applyStaticStyles(styleElement: HTMLStyleElement) {
             position: sticky;
             background: white;
             width: fit-content;
-            padding: 16px;
+            padding: 16px 28px 16px 16px;
             border: 3px solid black;
             margin: auto;
         }`,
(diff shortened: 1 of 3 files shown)


web/packages/core/src/shadow-template.ts
```diff
@@ -782,6 +782,26 @@ const videoModalClose = createElement("span", undefined, "close-modal");
...
+);
+const hardwareModalArea = createElement("div", undefined, "modal-area");
+const hardwareModalClose = createElement("span", undefined, "close-modal");
+hardwareModalClose.textContent = "\u00D7";
+const hardwareModalSpanOne = createElement("span");
+hardwareModalSpanOne.textContent =
+    "It looks like hardware acceleration is not enabled. While Ruffle may work, it could be unreasonably slow. You can enable hardware acceleration by ";
+const hardwareModalLink = document.createElement("a");
+hardwareModalLink.href =
+    "https://www.howtogeek.com/412738/how-to-turn-hardware-acceleration-on-and-off-in-chrome/";
```
danielhjacobs, line 799: We should probably not use this link.

n0samu (approved):

web/packages/core/src/shadow-template.ts
```diff
@@ -782,6 +782,26 @@ const videoModalClose = createElement("span", undefined, "close-modal");
...
+    "div",
+    "hardware-acceleration-modal",
+    "modal hidden",
+);
+const hardwareModalArea = createElement("div", undefined, "modal-area");
+const hardwareModalClose = createElement("span", undefined, "close-modal");
+hardwareModalClose.textContent = "\u00D7";
+const hardwareModalSpanOne = createElement("span");
+hardwareModalSpanOne.textContent =
+    "It looks like hardware acceleration is not enabled. While Ruffle may work, it could be unreasonably slow. You can enable hardware acceleration by ";
```
Dinnerbone, line 796: It'd be nice to localize this, but the way it's constructed won't be trivial to split up. Not sure what to do there 🤔

adrian17 (approved):

web/packages/core/src/shadow-template.ts
danielhjacobs: It can be: "It looks like hardware acceleration is not enabled. While Ruffle may work, it could be unreasonably slow. You can enable hardware acceleration by following this link: [link]". It would be a single string of text to translate then (since the link doesn't need translating), though it's a little uglier. Thoughts?

danielhjacobs: Updated, see screencast
[Screencast from 2023-10-02 10-22-41.webm](https://github.com/ruffle-rs/ruffle/assets/29206584/e1a72dd2-46a8-4949-8e78-52694d9697b8)

Dinnerbone (approved):
Thank you!

PR #13361, opened 2023-09-29
core: Font refactoring to allow backends to provide device fonts

Dinnerbone:
Peeled this out of my big font WIP.

The notable changes here are:
- There's an enum representing the "default" device fonts in Flash, ie `_sans` is `DefaultFont::Sans`.
- Each default font has a list of implementation font names, which currently is just `["Noto Sans"]` for each.
- Whenever a font is requested which doesn't already exist, it's treated as a device font and the UI backend is requested to load it
- The UI backend has a callback to register new fonts when prompted, and Player also has a method to register new fonts

I did a lot of back and forth on the designs of things before settling on this, and I fell down so many different rabbit holes. I'm brain dumping a bit below so that at least relevant research notes can be preserved before I inevitably try to forget the word "Font" and all its associated trauma.

## UI Backend callback vs return value
This totally *should* be a return value, ideally `Vec<FontDefinition>`. Unfortunately `swf::Font` has a lifetime, so... callback is easier.

## Font lookup should be Vec, not Option?
Yeah so here's a fun fact: nothing requires font names to be unique. If you request, say, Helvetica... well you may have Helvetica, Helvetica Bold, Helvetica Light, maybe a fancy version of Helvetica that has extra glyphs for CJK?

Now, that sounds like these are *variations* of a font... and that's true. But it doesn't have to be, that's just the most common scenario. Ultimately, they're named Helvetica and so they all need to be considered when we're looking for Helvetica.

## Font lookup
I wrote a todo for this in the code, I have no intention of fixing it right now because it's a bit of an invasive change and I'm already deep in it.

Due to the above point, when we look for fonts we need to look at **all** fonts to see if they match the one we're looking for.

We need to consider:
- Does this font have the glyphs we need for this text?
- Is this font capable of the style that we want to use it for, or do we need to emulate that style? (prefer ones that support it, but fallback to ones that don't)
- Is one of the names of this font, somewhat similar to the one we're looking for?

## Font names
Yeah so that one is a bit of a gotcha. Fonts have multiple names. Or they can have zero names! Their names can depend on your language, or your operating system!

We've seen a hint of this before with DefineFont tags having two names and this has caused a bit of confusion before.

I did some experimenting in Flash Player and found that:
- If the so called "full name" (name ID 4) of a font **starts with** the requested font name (case insensitive), that font is used
- If the so called "family name" (name ID 1) of a font **equals** (case insensitive) the requested font name, that font is used

We don't support anything like this yet. We currently store a single `FontIdentifier` inside a `Font`, but it should instead be that we ask each font "do you match this request".

## FontDefinition enum vs swf::Font directly
Right now it's a single variant enum, but it'll get at least 2 more variants when ttf support lands. Something like:
`FontCollection { bytes: Vec<u8>, index: u32 }` - used for `.ttc` font collections, index is the font within the collection
`FontFile { bytes: Vec<u8> }` for regular `.ttf`, `.otf` etc

And maybe for good measure:
`Swf(SwfMovie)` to allow multiple fonts packaged together, whilst benefiting from the compression of a swf over a straight up font tag.

On that note...

## DefineFont is amazing at compressing fonts.
It's a shame it's so lossy though.

NotoSans-Regular.ttf is 543 KB
The DefineFont3 tag for the same font is 356 KB
The SWF containing that tag is 211 KB.

Of course, if you limit it to just latin as we do today, then it's 107 KB tag / 72 KB SWF.

The lossy part, is that there's way more information in a font which isn't expressible through the SWF format. Most notably:
- GPOS is the modern day kerning, which has way more control over glyph placements. Newer fonts use it and not traditional kerning tables, so they'd just look weird in SWF format.
- GSUB makes things like ligatures work, and is required for some languages to be readable at all. No idea if it's possible to embed an arabic font, for example.

Honorable mention to variable fonts, which in theory could be much smaller by virtue of a single font for all style variations, instead of embedding a Regular, an Italic, a Bold, a Bold Italic, etc. Noto Sans TC (traditional chinese) is 64 MB across 9 variations... or a single 11MB variable font!

## Speaking of font styles
It's broken today. We don't emulate styles if there's no explicit matching font for them. We should fix that, but I'm leaving that for future us. Screw future us, what did they ever do for us?

## Speaking of our noto sans default
We should switch it to embed the swf rather than the tag. The swf is 44KB vs the tags 73KB. That's swf default speedy compression, too.

## Default fonts
Flash gives (in theory) 3 so called "default fonts". They're `_sans`, `_serif` and `_typewriter`.

I suspect there's more undocumented ones. I'm fairly sure I've seen a translation of `_typewriter` recently... and that apparently worked.

We currently just default everything over to NotoSans today, and I've preserved that in this PR by defining the implementation of each default font as "just noto sans", but this will change in the future:

- Desktop will default each to something appropriate to the OS
- Web will likely have a config to allow changing the implementations of each default

Note that it's a list of font implementations, not a single font. Maybe we can't actually find that font, or maybe we need to provide a latin font, a chinese font, a japanese font... etc.

## Multiple fonts per... font...
Oh god the word font has no meaning anymore. I'm aware I'm using it wrong in some places too, but whatever.

If a textbox has a mixture of latin and chinese text on the same line, it's very unlikely that a single font has both sets of glyphs. Fonts can only have a small amount of glyphs each, after all.

We need to change how we do text layouts to account for changing font per glyph - even if the font name is the same - so that we can handle this correctly. Today we'd just pick one and hope for the best. It's wrong before this PR, it's wrong after this PR. Future us can fix it. I believe in them.

## Fonts must exist on demand
So the way it works now is that when we need a font that doesn't exist to Ruffle, we ask the UIBackend for it. The font either must exist before the end of that call, or we're going to assume it doesn't exist ever. It sucks, it's going to mean more work on Web, but that's just how it is.

There can be actionscript code that makes a text box, puts some text in it, assigns a font, and then uses the width of the text to center it on the screen. It can only do that if the font existed right there and then, and so whatever font we end up using **must** exist at the time that we request it.

## Web config?
A future PR will make fonts configurable on web. It won't require any core changes to do so, this PR lays all the foundation for that.

Most likely it'll be something like two config sections:
- Default fonts implementations: a list of font names per default. A webmaster could change `_sans` to `["Comic Sans"]` if they wanted. (or realistically, add CJK fonts)
- A list of fonts sources (URL or blobs) to load

Because fonts must exist right away, web will have to resolve and download any URLs before starting playback of the movie. It'll then add it to the player through `player.register_device_font()`, and the UIBackend method just goes unimplemented here.

Maybe one day that theoretical low level font api on web browsers could be implemented, then we'd switch to the UIBackend method instead. Maybe. Doubt it.

## Default fonts default fonts feature
What even are words?

In the future, desktop won't use the NotoSans default that we use today. Web will, exporter probably will, android app may move away from it too one day.

It's a waste to bundle that font in executables that won't use it, but it's also kind of a waste to have every *other* executable need to copy it and the set up code for it.

As a compromise, it's a feature. It's on everywhere right now, and desktop can just turn it off later. We can revisit this later if we care.

### DIFF
--- a/core/src/html/layout.rs
+++ b/core/src/html/layout.rs
@@ -7,9 +7,11 @@ use crate::html::dimensions::{BoxBounds, Position, Size};
 use crate::html::text_format::{FormatSpans, TextFormat, TextSpan};
 use crate::string::{utils as string_utils, WStr};
 use crate::tag_utils::SwfMovie;
+use crate::DefaultFont;
 use gc_arena::Collect;
 use ruffle_render::shape_utils::DrawCommand;
 use std::cmp::{max, min};
+use std::ops::Deref;
 use std::sync::Arc;
 use swf::{Point, Twips};
 
@@ -462,20 +464,47 @@ impl<'a, 'gc> LayoutContext<'a, 'gc> {
         is_device_font: bool,
     ) -> Option<Font<'gc>> {
         let library = context.library.library_for_movie_mut(self.movie.clone());
+        let font_name = span.font.to_utf8_lossy();
 
-        // If this text field is set to use device fonts, fallback to using our embedded Noto Sans.
         // Note that the SWF can still contain a DefineFont tag with no glyphs/layout info in this case (see #451).
         // In an ideal world, device fonts would search for a matching font on the system and render it in some way.
-        if let Some(font) = library
-            .get_font_by_name(&span.font.to_utf8_lossy(), span.bold, span.italic)
-            .filter(|f| !is_device_font && f.has_glyphs())
-            .or_else(|| context.library.device_font())
-        {
-            self.font = Some(font);
-            return self.font;
+        if !is_device_font {
+            if let Some(font) = library
+                .get_font_by_name(&font_name, span.bold, span.italic)
+                .filter(|f| f.has_glyphs())
+            {
+                return Some(font);
+            }
         }
 
-        None
+        if let Some(font) = context.library.load_device_font(
+            &font_name,
+            context.ui,
+            context.renderer,
+            context.gc_context,
+        ) {
+            return Some(font);
+        }
+
+        // [NA] I suspect that there might be undocumented more aliases.
+        // I think I've seen translated versions of these used in the wild...
+        let default_font = match font_name.deref() {
+            "_serif" => DefaultFont::Serif,
+            "_typewriter" => DefaultFont::Typewriter,
+            _ => DefaultFont::Sans,
+        };
+
+        // TODO: handle multiple fonts for a definition, each covering different sets of glyphs
+        context
+            .library
+            .default_font(
+                default_font,
+                context.ui,
+                context.renderer,
+                context.gc_context,
+            )
+            .first()
+            .copied()
     }
 
     /// Append text to the current line of the ongoing layout operation.
@@ -505,15 +534,17 @@ impl<'a, 'gc> LayoutContext<'a, 'gc> {
     /// This function bypasses the text fragmentation necessary for justify to
     /// work and it should only be called internally.
     fn append_text_fragment(&mut self, text: &'a WStr, start: usize, end: usize, span: &TextSpan) {
-        let params = EvalParameters::from_span(span);
-        let text_size = Size::from(self.font.unwrap().measure(text, params, false));
-        let text_bounds = BoxBounds::from_position_and_size(self.cursor, text_size);
-        let mut new_text = LayoutBox::from_text(start, end, self.font.unwrap(), span);
+        if let Some(font) = self.font {
+            let params = EvalParameters::from_span(span);
+            let text_size = Size::from(font.measure(text, params, false));
+            let text_bounds = BoxBounds::from_position_and_size(self.cursor, text_size);
+            let mut new_text = LayoutBox::from_text(start, end, font, span);
 
-        new_text.bounds = text_bounds;
+            new_text.bounds = text_bounds;
 
-        self.cursor += Position::from((text_size.width(), Twips::default()));
-        self.append_box(new_text);
+            self.cursor += Position::from((text_size.width(), Twips::default()));
+            self.append_box(new_text);
+        }
     }
 
     /// Append a bullet to the start of the current line.
@@ -527,12 +558,8 @@ impl<'a, 'gc> LayoutContext<'a, 'gc> {
         span: &TextSpan,
         is_device_font: bool,
     ) {
-        let library = context.library.library_for_movie_mut(self.movie.clone());
-
-        if let Some(bullet_font) = library
-            .get_font_by_name(&span.font.to_utf8_lossy(), span.bold, span.italic)
-            .filter(|f| !is_device_font && f.has_glyphs())
-            .or_else(|| context.library.device_font())
+        if let Some(bullet_font) = self
+            .resolve_font(context, span, is_device_font)
             .or(self.font)
         {
             let mut bullet_cursor = self.cursor;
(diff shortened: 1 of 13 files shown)


core/src/html/layout.rs
```diff
@@ -505,15 +534,17 @@ impl<'a, 'gc> LayoutContext<'a, 'gc> {
...
 
-        new_text.bounds = text_bounds;
+            new_text.bounds = text_bounds;
 
-        self.cursor += Position::from((text_size.width(), Twips::default()));
-        self.append_box(new_text);
+            self.cursor += Position::from((text_size.width(), Twips::default()));
+            self.append_box(new_text);
+        }
     }
```
Deepfriedice, line 548: Sorry if I'm butting in.
Does it make sense for this function to silently do nothing (rather than failing loudly?) if there's no font set?

Dinnerbone: I'm not sure. There's a warning printed in the log when a font isn't found, and this situation here is only when there's no font at all - so someone compiled without the default font. Elsewhere in the file we also just ignore this scenario.

I supposed we could make a fallback-fallback font that's just tofu, but...

Deepfriedice: If there's a warning raised earlier, then that makes sense.

core/src/html/layout.rs
```diff
@@ -462,20 +464,47 @@ impl<'a, 'gc> LayoutContext<'a, 'gc> {
...
+        ) {
+            return Some(font);
         }
 
-        None
+        // [NA] I suspect that there might be undocumented more aliases.
+        // I think I've seen translated versions of these used in the wild...
+        let default_font = match font_name.deref() {
+            "_serif" => DefaultFont::Serif,
+            "_typewriter" => DefaultFont::Serif,
```
sleepycatcoding, line 493: I see that DefaultFont has a `TypeWriter` variant, should that be used instead?

core/src/library.rs
```diff
@@ -432,14 +452,87 @@ impl<'gc> Library<'gc> {
...
+        }
+
+        self.default_font_cache.insert(name, result.clone());
+        result
+    }
+
     /// Returns the device font for use when a font is unavailable.
-    pub fn device_font(&self) -> Option<Font<'gc>> {
-        self.device_font
+    pub fn load_device_font(
```
adrian17, line 481: minor nit: any chance we can rename one of the `load_device_font`? The duplicate naming adds to the confusion :c

core/src/html/layout.rs
adrian17: > so someone compiled without the default font

If it can only possibly be a programmer bug, I'd say it should still be an unwrap/panic/unreachable/etc.

core/src/html/layout.rs
Dinnerbone: Yep, thanks for catching it. Fixed!

core/src/library.rs
Dinnerbone: Made that rename, good idea!

core/src/html/layout.rs
Dinnerbone: I had a further look into what Flash did here just to make sure I wasn't missing something, and discovered that you can trigger the behavior in Flash with:

```actionscript
var text:TextField = new TextField();
text.text = "test";
text.embedFonts = true; // don't fall back to a device font

var format:TextFormat = new TextFormat(); 
format.size = 14;
format.font = "XXXXX";

text.setTextFormat(format); 

addChild(text);
```

So I think the current behavior would be correct for when no fallback was found, which ActionScript can sometimes force upon us.

PR #13372, opened 2023-09-30
avm2: Correctly fire `textureReady` event for async Texture uploads

Lord-McSweeney:
The Starling Benchmark was waiting for an event that never came.

### DIFF
--- a/core/src/avm2/globals/flash/display3D/textures/Texture.as
+++ b/core/src/avm2/globals/flash/display3D/textures/Texture.as
@@ -1,11 +1,24 @@
 package flash.display3D.textures {	
     import flash.display.BitmapData;
+    import flash.events.Event;
     import flash.utils.ByteArray;
-    import __ruffle__.stub_method;
+    import flash.utils.setTimeout;
     
     public final class Texture extends TextureBase {
         public native function uploadFromBitmapData(source:BitmapData, miplevel:uint = 0):void;
         public native function uploadFromByteArray(data:ByteArray, byteArrayOffset:uint, miplevel:uint = 0):void;
-        public native function uploadCompressedTextureFromByteArray(data:ByteArray, byteArrayOffset:uint, async:Boolean = false):void;
+        public function uploadCompressedTextureFromByteArray(data:ByteArray, byteArrayOffset:uint, async:Boolean = false):void {
+            if (async) {
+                var self = this;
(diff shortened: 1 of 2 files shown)


core/src/avm2/globals/flash/display3D/textures/Texture.as
```diff
@@ -1,11 +1,24 @@
...
     
     public final class Texture extends TextureBase {
         public native function uploadFromBitmapData(source:BitmapData, miplevel:uint = 0):void;
         public native function uploadFromByteArray(data:ByteArray, byteArrayOffset:uint, miplevel:uint = 0):void;
-        public native function uploadCompressedTextureFromByteArray(data:ByteArray, byteArrayOffset:uint, async:Boolean = false):void;
+        public function uploadCompressedTextureFromByteArray(data:ByteArray, byteArrayOffset:uint, async:Boolean = false):void {
+            if (async) {
+                var self = this;
+                setTimeout(function() {
+                    self.uploadCompressedTextureFromByteArrayInternal(data, byteArrayOffset);
```
Aaron1011, line 14: Can you move this out of the `setTimeout`? I've seen Stage3D code that breaks when events are fired from within the method (the listener gets registered afterward).

Lord-McSweeney: I don't see how this could break anything? Starling registers the event listeners after calling `uploadCompressedTextureFromByteArray`, so these need to be in a setTimeout.

Aaron1011: I mean, move just the `uploadCompressedTextureFromByteArrayInternal` call out of the `setTimeout`, and leave the `dispathEvent` call in the `setTimeout`

Lord-McSweeney: I don't think the current implementation can break anything? The Adobe docs say that the function immediately returns and dispatches the event after `uploadCompressedTextureFromByteArrayInternal` has been called.

Aaron1011: Oh, I see what you mean. Yeah, this works fine

Aaron1011 (approved):

adrian17 (approved):

PR #13382, opened 2023-09-30
avm2: Implement Sound.loadCompressedDataFromByteArray

Aaron1011:


adrian17 (approved):

core/src/avm2/globals/flash/media/sound.rs
```diff
@@ -264,14 +264,28 @@ pub fn load<'gc>(
...
 ) -> Result<Value<'gc>, Error<'gc>> {
-    avm2_stub_method!(
-        activation,
-        "flash.media.Sound",
-        "loadCompressedDataFromByteArray"
-    );
+    let bytearray = args.get_object(activation, 0, "bytes")?;
+    let bytes_length = args.get_u32(activation, 1)?;
+    let bytearray = bytearray
+        .as_bytearray_mut(activation.context.gc_context)
```
sleepycatcoding, line 273: I think you can just use `as_bytearray` here, as read_bytes does not need mutability. (Position is stored in a Cell)

Aaron1011: Good catch - I've changed it

sleepycatcoding (approved):

PR #13383, opened 2023-09-30
avm2: Implement some obscure SWFv9 XML behavior

sleepycatcoding:
This affects appendChild, prependChild, insertChildBefore, insertChildAfter.

### DIFF
--- a/core/src/avm2/e4x.rs
+++ b/core/src/avm2/e4x.rs
@@ -1183,3 +1183,29 @@ pub fn name_to_multiname<'gc>(
     };
     Ok(multiname)
 }
+
+// Based on https://github.com/adobe/avmplus/blob/858d034a3bd3a54d9b70909386435cf4aec81d21/core/XMLObject.cpp#L1543
+// This is needed to reproduce some weird behavior in SWFv9.
+pub fn maybe_escape_child<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    child: Value<'gc>,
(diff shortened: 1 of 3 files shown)


core/src/avm2/e4x.rs
```diff
@@ -1183,3 +1183,28 @@ pub fn name_to_multiname<'gc>(
...
     Ok(multiname)
 }
+
+// Based on https://github.com/adobe/avmplus/blob/858d034a3bd3a54d9b70909386435cf4aec81d21/core/XMLObject.cpp#L1543
+// This is needed to reproduce some weird behavior in SWFv9.
+pub fn maybe_escape_child<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    child: Value<'gc>,
+) -> Result<Value<'gc>, Error<'gc>> {
+    if activation.context.swf.version() <= 9 {
```
Lord-McSweeney, line 1193: Can you add a comment here stating that this depends on the root SWF? Otherwise we might end up accidentally changing it to `activation.caller_movie()` in the future

sleepycatcoding: I have no idea if this depends on the caller_movie or root SWF. (I didn't know that caller_movie existed).

Lord-McSweeney: avmplus has a comment saying that it uses the root movie.

Dinnerbone (approved):

PR #13389, opened 2023-10-01
avm2: Run DoAbc/DoAbc2/SymbolClass as part of their frame

Aaron1011:
We previously ran these tags during preloading - however, they are actually run as part of frame execution. This is observable by ActionScript - a SWF can load in a class from a stop()'d MoveClip, and then advance the clip to a frame with a SymbolClass referencing the loaded class.


adrian17 (approved):
Approved, sans minor test nitpicks on Discord

> btw, also the test .as is out of date? Or missing some scripts?
> I don't see the gotoAndStop frame scripts
> I also don't get testing getDefinitionByName("MyChild"), there's only a SecondFrameChild and FourthFrameChild, right?

PR #13447, opened 2023-10-06
avm2: Fix KeyboardEvent clone

sleepycatcoding:


core/src/avm2/globals/flash/events/KeyboardEvent.as
```diff
@@ -85,7 +85,8 @@ package flash.events
 
         override public function clone() : Event
         {
-            return new KeyboardEvent(this.type,this.bubbles,this.cancelable);
+            // FIXME: Constructor has command key value, but we dont have a field for it?
```
torokati44, line 88: There is no explanation in https://github.com/ruffle-rs/ruffle/pull/8935 as to _why_ `commandKey` was omitted, when `ctrlKey` and `controlKey` were both added.

sleepycatcoding: Should we just add the field?

torokati44: I don't see why not. It's a really minor thing, but technically increases compatibility a little bit, so...

torokati44 (approved):

core/src/avm2/globals/flash/events/KeyboardEvent.as
adrian17: That's because control/commandKey don't actually exist outside of AIR. If anything, we shouldn't have either.

torokati44: Argh, I deliberately looked for the AIR icon in the docs, couldn't find any. Are they lying again?

adrian17: might be just a documentation bug? They mention it in text:
```
commandKey	property	 
commandKey:Boolean
Language Version: 	ActionScript 3.0
Runtime Versions: 	AIR 1.0, Flash Lite 4
```

Issue #9954, opened 2023-03-10
Georganism 3: Progress not saved


n0samu:
### Describe the bug

Link to SWF: http://i.notdoppler.com/files/georganism3.swf

After beating a level of [Georganism 3](http://i.notdoppler.com/files/georganism3.swf), the following error appears several times in the terminal:
`ERROR run_frame:run_all_phases_avm2: ruffle_core::avm2::activation: ReferenceError: Error #1069: Property setProperty not found on flash.net.SharedObject and there is no default value.`

The game is still playable; you can go back to the level select screen after beating a level and play the next level. But progress is not saved between sessions.

### Expected behavior

The "level passed" screen should appear after beating a level, and progress should be saved between sessions.

### Affected platform

Desktop app

### Operating system

Windows 10

n0samu:
Thanks to #11052, you no longer need to go back to the level select screen after beating each level. But progress still isn't saved - I guess a full implementation will be needed.

n0samu:
Unfortunately this problem is not fully fixed by #13451. Progress is saved, but for some reason, clicking the Play button on the main menu resets progress on the first level, while leaving progress on the rest of the levels intact.

n0samu:
Just checked Flash Player and the problem happens there too. I guess this is a game bug. The Ruffle bug is fixed, so I will close this issue.

Issue #13426, opened 2023-10-04
Evoland (Linux Steam version)


Firlaev-Hans:
### Describe the bug

**(Read _Additional Information_ first)**

The game starts normally, but besides some small issues being a little slow and laggy and missing music, as well as playing sound effects in the initial starting area before sound effects are even unlocked, you can't progress very far at all, because as soon as you unlock 16 color graphics (~1-2 minutes into the game), Ruffle encounters this issue:
```
thread 'main' panicked at 'Found ignore_sampler in SamplerField { filter: Linear, mipmap: Disable, wrapping: Clamp, dimension: TwoD, special: Special { ignore_sampler: true }, texture_lod_bias: 0, reg_num: 0, reg_type: Sampler }', render/naga-agal/src/builder.rs:1148:21
stack backtrace:
   0:     0x5600f5ad7a31 - std::backtrace_rs::backtrace::libunwind::trace::h66dc1c6acf794faa
                               at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:     0x5600f5ad7a31 - std::backtrace_rs::backtrace::trace_unsynchronized::ha80d20099a67f790
                               at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x5600f5ad7a31 - std::sys_common::backtrace::_print_fmt::h7b959d43f35f16d4
                               at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library/std/src/sys_common/backtrace.rs:65:5
   3:     0x5600f5ad7a31 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hdaa196410d9ee0b9
                               at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library/std/src/sys_common/backtrace.rs:44:22
   4:     0x5600f5b08c1f - core::fmt::rt::Argument::fmt::h0ddfbbe8be3f80d0
                               at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library/core/src/fmt/rt.rs:138:9
   5:     0x5600f5b08c1f - core::fmt::write::h66b3c629f3d623e4
[...]

PR #13451, opened 2023-10-07
avm2: Implement SharedObject.setProperty

Lord-McSweeney:
Fixes save points not working on #13426. The whole file is marked as changed because I formatted it according to the majority of the AS files (`public native` instead of `native public`, 4 spaces instead of 3 spaces, etc)


n0samu:
Should also fix #9954!

adrian17 (approved):
LGTM.

BTW, for other reviewers:
> The whole file is marked as changed
There's a option to hide whitespace changes, though not immediately visible.

PR #13513, opened 2023-10-11
avm1: implemented Key.is toggled

### LINKED ISSUE #8579: AS2: Ruffle does not have Key.isToggled implemented
### Describe the bug

When using Key.isToggled in my project,
 I realized that it didn't work in Ruffle.
After talking with Nosamu on the discord, it turns out that the code I wrote shouldn't really work, according to documentation, lol

but it does indeed work in flash player, and has yet to be implemented in Ruffle
I've created an example you can try out below

[toggle example.zip](https://github.com/ruffle-rs/ruffle/files/10046047/toggle.example.zip)

### Expected behavior

Key.isToggle should be able to work, as in flash player

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

Its important to note that the docs only specify Key.isToggled is for caps lock and num lock, not other keys

however, it works with any key in flash, as far as I can tell.
here's a link to the documentation I was shown:

https://open-flash.github.io/mirrors/as2-language-reference/Key.html#isToggled()

michiel2005:
Implemented Key.is toggled.

Ready for review.
quick reference link: https://open-flash.github.io/mirrors/as2-language-reference/

It will close #8579


torokati44:
Thank you for the effort, but this approach will sadly not work for Ruffle at all.
We're relying on `winit` for all input processing, for the exact reason that it provides us with universal way to do these kinds of things, across all operating systems and platforms. What you did is really-really Windows-specific, and in a way, "cuts across abstraction levels" abruptly to a low-level OS function from a high-level application function.
So this method will have to utilize the functionality provided by `winit` as well.

michiel2005:
Cool, thanks for this info :)

Lord-McSweeney (approved):

torokati44:
Is there a possibility here that for CapsLock (and potentially NumLock and ScrollLock), this reports an inverse state than what the actual lock state is? Such as, when Ruffle is started with CapsLock being on. In either case, does it match what FP does in these cases?

michiel2005:
I couldn't find any way to make that possible with winit. The keypresses are not registered when ruffle is not in focus. This is correct for every key except caps lock and scroll lock.

torokati44:
Alright, that's fair, and definitely good enough. I don't have a better idea either. Just so we know that this is not _perfect_.

torokati44:
Also, do you mind if I clean up the commit history a bit for you?

michiel2005:
Yes, feel free to clean up the commit's. Or you could tell me how to do it. What you want :)

torokati44:
There you go. I've also adjusted the comment a little bit, I hope you don't mind!

torokati44 (approved):

PR #13537, opened 2023-10-13
avm2: Correctness fixes to AMF serialization and ByteArray object r/w

Lord-McSweeney:
This gets a `writeObject` and then a `readObject` of the following objects into a `ByteArray` to match Flash (both the round-trip of a `writeObject` and then `readObject`, and the actual binary contents of the `ByteArray`):
- `new Function()`
- `Vector.<Function>([new Function(), ...])`


core/src/avm2/amf.rs
```diff
@@ -281,7 +281,10 @@ pub fn deserialize_value<'gc>(
             }
             let storage = VectorStorage::from_values(
                 vec.iter()
-                    .map(|v| deserialize_value(activation, v))
+                    .map(|v| {
+                        deserialize_value(activation, v)
+                            .map(|value| value.as_object().map(Value::from).unwrap_or(Value::Null))
```
Lord-McSweeney, line 286: I need to fix this to only convert `Undefined` to `Null`- that should stop the failing tests (since there's no `Vector.<void>`)

Aaron1011 (approved):

sleepycatcoding:
Can you add the fixes to Socket.read/writeObject as well?

sleepycatcoding (approved):

PR #13576, opened 2023-10-16
ci: Set up cargo-deny

torokati44:
I'd like to get verification that the allowed/denied licenses are actually correct.
Also I haven't dared to deny duplicate crates, as there are a lot of them.


.github/workflows/test_rust.yml
```diff
@@ -76,6 +76,10 @@ jobs:
         # Don't fail the build for clippy on nightly, since we get a lot of false-positives
         run: cargo clippy --all --all-features --tests ${{ (matrix.rust_version != 'nightly' && '-- -D warnings') || '' }}
 
+      - name: cargo-deny
```
Dinnerbone, line 79: I'd maybe limit this to *one* of the Linuxes

torokati44: I've limited it to Linux and Rust stable.

Dinnerbone (approved):

PR #13591, opened 2023-10-17
core: Switch dasp from a git revision to v0.11.0

torokati44:
The previously referenced commit was originally part of https://github.com/RustAudio/dasp/pull/137.

The nightly compilation error it is supposed to fix (https://github.com/RustAudio/dasp/issues/136), isn't reproducible now, and it already wasn't in 2021: https://github.com/RustAudio/dasp/pull/155

This trims our dependency tree a little bit, since `dasp` `v0.11.0` is also an indirect dependency, as pointed out by #13576.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -857,7 +857,7 @@ dependencies = [
  "alsa",
  "core-foundation-sys",
  "coreaudio-rs",
- "dasp_sample 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "dasp_sample",
  "jni 0.19.0",
  "js-sys",
  "libc",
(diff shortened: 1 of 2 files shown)


torokati44:
Also note that `dasp` is a very low frequency project nowadays - not (necessarily...?) because it's "abandoned", perhaps only because it's "done". And the latest release, `v0.11.0`, doesn't include the fix in the previously referenced commit. Had there been any issues with it in the last three years, surely somebody would have fixed it, or at least have shouted/cried/cursed about it.

Dinnerbone (approved):

Benman2785:
this fixes audio problems in "Road of the Dead"

torokati44:
Huh, that's interesting! What problems exactly?

Benman2785:
before update the audio "lagged" in "Road of the Dead" at the start and when changing multiple lanes at once - audio output stopped then - now at first try it was fine (will test again now)

Benman2785:
yep its fixed now - also the horn - you can now press it continously - with old version it would have stopped

https://youtu.be/jbLGYwRtjus

torokati44:
Okay, great, thanks!

Benman2785:
wait - its not related to it - looks like it was fixed in an earlier version as 16-10-2023 also works fine (sry - havent tested the game in a while)

torokati44:
Yeah that makes a bit more sense logically... :sweat_smile:

Issue #13553, opened 2023-10-14
Seedling: "Variable Main is not defined" freeze during loading


madisonsilver:
### Describe the bug

When attempting to run Seedling, the game will freeze at 95% loaded (see attached image), and output the following error to the console:
```ERROR core/src/avm2/events.rs:424 Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x2968d38 }) to handler FunctionObject(FunctionObject { ptr: 0x806b30 }) : ReferenceError: Error #1065: Variable Main is not defined.```

Website: https://connorullmann.com/games/Seedling/Seedling.html
.swf: https://connorullmann.com/games/Seedling/Seedling.swf

Website: https://www.kongregate.com/games/TimeBlog/seedling?acomplete=seedling
.swf: [game150260.konggames.com/gamez/0015/0260/live/Seedling.swf?kongregate_game_version=1343066131](game150260.konggames.com/gamez/0015/0260/live/Seedling.swf?kongregate_game_version=1343066131)

(Note that Kongregate may also show the following error:
```ERROR core/src/loader.rs:800 Error during movie loading of "https://game150260.konggames.com/games/TimeBlog/seedling/frame/9a6a7a85-04ef-4abc-8fe4-64b9a7b0e89c/https%3A%2F%2Fchat.kongregate.com%2Fflash%2FAPI_AS3_d43c4b859e74432475c1627346078677.swf": HttpNotOk("HTTP status is not ok, got ", 404, false)```
This appears to be a Kongregrate-specific problem.)

![image](https://github.com/ruffle-rs/ruffle/assets/24858452/f6c6f8a6-5eb4-4e47-a690-df2135772734)

### Expected behavior

The game finishes loading and proceeds to the credits and title screen.

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Google Chrome 117, Firefox 118

### Additional information

This was tested on both Google Chrome 117 and Firefox 118.

~~I also tested the [2023-10-14 nightly build](https://github.com/ruffle-rs/ruffle/releases/tag/nightly-2023-10-14) on desktop, but it only showed a black screen.~~  Edit: It turned out that this was a result of Seedling's domain checking.  Trying it again with the correct link resulted in the same error as above after loading 0%.

```

# Player Info
Allows script access: true
Renderer: wgpu
Adapter Backend: Gl
[...]

Issue #13564, opened 2023-10-16
Give Up 2 - regression


zjk261:
### Describe the bug

game stuck after pressing play

https://github.com/ruffle-rs/ruffle/assets/98996990/4402c025-135b-40b5-94aa-f30042421673

[giveup2_viral.zip](https://github.com/ruffle-rs/ruffle/files/12909387/giveup2_viral.zip)

### Expected behavior

https://github.com/ruffle-rs/ruffle/assets/98996990/9e081c07-2cc2-4b29-96bf-847ba03bfa17

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

Chrome 117.0.5938.152

Lord-McSweeney:
I don't think this is a regression, but a very precise timing issue, similarly to other issues that have been recently reported. I can reproduce it on `10-02`, `10-05`, and `10-16`.

Issue #13563, opened 2023-10-15
Quietus II: "Variable Main is not defined" freeze


madisonsilver:
### Describe the bug

When attempting to run Quietus II, after pressing the play button, the game freezes and outputs the following error:
```
ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "mouseUp", class: flash.events::MouseEvent, ptr: 0x1f6bb066050 }) to handler FunctionObject(FunctionObject { ptr: 0x1f6b785c9c0 }) : ReferenceError: Error #1065: Variable Main is not defined.
```

url: https://www.newgrounds.com/portal/view/601856/format/flash?emulate=flash
.swf: https://uploads.ungrounded.net/601000/601856_QuietusII.swf?1346159466

### Expected behavior

The game progresses to the title screen.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

Also tested on Google Chrome 117 via the Newgrounds page. May be related to #13553 and #13562.

madisonsilver:
Stack:
```
ReferenceError: Error #1065: Variable Main is not defined.
        at global/flash.utils::getDefinitionByName()
        at Preloader/startup()
        at Preloader/load()
        at Preloader/onPlayClick()
```

Issue #13562, opened 2023-10-15
Praepoch: "Variable Main is not defined" freeze during loading


madisonsilver:
### Describe the bug

When attempting to run Praepoch, the game freezes while loading and outputs the following error:

```
ERROR core/src/avm2/events.rs:424 Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x12d3a40 }) to handler FunctionObject(FunctionObject { ptr: 0xb5cf80 }) : ReferenceError: Error #1065: Variable Main is not defined.
```

swf: https://uploads.ungrounded.net/601000/601861_Praepoch.swf?1346100357
url: https://www.newgrounds.com/portal/view/601861/format/flash?emulate=flash

### Expected behavior

The game finishes loading and proceeds to the title screen.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

Also tested on Google Chrome 117 via the Newgrounds page.  May be related to #13553 .

Lord-McSweeney:
Stack:
```
ReferenceError: Error #1065: Variable Main is not defined.
	at global/flash.utils::getDefinitionByName()
	at Preloader/startup()
	at Preloader/load()
	at Preloader/onEnterFrame()
```

PR #13595, opened 2023-10-18
avm2: Don't stop root movie clip, even if it doesn't extend `MovieClip`

Aaron1011:
Normally, Flash Player will ignores frames for a movie clip with a symbol class that doesn't extend `MovieClip` (e.g. it extends `Sprite`). However, the root movie appears to have frame run unconditionally, even if it only extends `Sprite`. This can be observed by adding a Graphics child in the second frame only - the child will flicker in and out as the player switches between frames, but only for the root movie clip.

Seedling relies on this behavior - it has `DoAbc2` tags in the second frame, and has a main `Preloader` class that extends `Sprite`.


Lord-McSweeney (approved):

madisonsilver:
I tested and can confirm that it resolves #13553, #13562, #13563, and #13564.  Thank you so much!

sleepycatcoding (approved):

PR #13604, opened 2023-10-18
web: Add defaultFonts and fontSources config options

Dinnerbone:
# Font sources
image

Each URL provided to this config option will *try* to eagerly load as a font source.
If the URL cannot load, it will be ignored (but a warning will print).
If the URL cannot parse into a valid font, it will be ignored (but a warning will print).

Right now it only supports fonts embeded inside SWFs. Any number of fonts can be inside a single SWF, but they need to be embedded with DefineFont2 and you need to make sure it's actually inside the swf (not stripped out by not being used).
A future PR will enable TTF/OTF support.

### 1. Create a new document in Flash Pro or Animate, make sure it's ActionScript 3.
image

### 2. On the right hand side, click the Library tab and right click somewhere in the big empty space, and hit New Font
image

### 3. Set the name, pick the font and style from the list, and then choose which characters you want to embed. Only these characters will be available to use from this font.
image

### 4.  In the ActionScript tab, check "Export for ActionScript" to make sure it stays inside the swf, and make sure it's "DF3" not "DF4".
image

### 5. Export the SWF. Make sure you export it as a SWF
image
image

### 6. Configure Ruffle to load that SWF as a font source.
```js
    fontSources: ["my-big-fonts.swf"]
}
```

# Default fonts
image

There are 3 "default" device fonts in Flash. `_sans`, `_serif`, `_typewriter`. In Ruffle they are our Noto Sans fallback unless configured otherwise.

To change that, set:
```js
    defaultFonts: {
        sans: ["Caveat"]
    }
}
```

The font in question needs to be loaded with fontSources or it won't be found.

### DIFF
--- a/web/src/lib.rs
+++ b/web/src/lib.rs
@@ -19,8 +19,8 @@ use ruffle_core::external::{
     ExternalInterfaceMethod, ExternalInterfaceProvider, FsCommandProvider, Value as ExternalValue,
     Value,
 };
-use ruffle_core::swf;
 use ruffle_core::tag_utils::SwfMovie;
+use ruffle_core::{swf, DefaultFont};
 use ruffle_core::{
     Color, Player, PlayerBuilder, PlayerEvent, SandboxType, StageAlign, StageScaleMode,
     StaticCallstack, ViewportDimensions,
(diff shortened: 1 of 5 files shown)


Dinnerbone:
(I spent a long time trying to get the internal config right but that ended up being a rabbit hole for another PR. We shouldn't pass the user set config all the way down to rust.)

web/src/lib.rs
```diff
@@ -460,6 +462,59 @@ impl Ruffle {
...
+                                tracing::warn!("DefineFont4 tag is not yet supported by Ruffle, inside font swf {font_name}");
+                            }
+                            _ => {}
+                        }
+                    }
+                    return;
+                }
+            }
+
+            tracing::warn!("Font source {font_name} was not recognise (not a valid SWF?)");
```
sleepycatcoding, line 493: Non-blocking, but this feels like a typo recognise -> recognised. P.S.  English is not my primary language.

sleepycatcoding (approved):
Looks good to me!

web/src/lib.rs
Dinnerbone: Fixed thank you!

PR #13605, opened 2023-10-18
web: Remove a whole bunch of internal stuff from public API + documentation, and simplify loading a bit

Dinnerbone:
Originally meant to change how config was passed over to rust but rabbit holes in rabbit holes 😅


sleepycatcoding (approved):
Looks good to me.

PR #13631, opened 2023-10-20
avm2: Replace culling panic messages with stub messages

sleepycatcoding:
A stub message is good enough.

### DIFF
--- a/core/src/avm2/globals/flash/display/graphics.rs
+++ b/core/src/avm2/globals/flash/display/graphics.rs
@@ -993,7 +993,7 @@ fn draw_triangles_internal<'gc>(
         let indices = &mut indices.iter();
 
         while let Some(triangle) = next_triangle(&vertices, indices, activation)? {
-            draw_triangle_internal(triangle, drawing, culling);
+            draw_triangle_internal(activation, triangle, drawing, culling);
         }
     } else {
         let mut vertices = vertices.iter();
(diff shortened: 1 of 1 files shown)


core/src/avm2/globals/flash/display/graphics.rs
```diff
@@ -1054,10 +1059,20 @@ fn draw_triangle_internal((a, b, c): Triangle, drawing: &mut Drawing, culling: T
             drawing.draw_command(DrawCommand::LineTo(a));
         }
         TriangleCulling::Positive => {
-            panic!("Positive culling not implemented")
+            avm2_stub_method!(
+                activation,
+                "flash.display.Graphics",
+                "drawTriangles",
+                "Positive culling not implemented"
```
Lord-McSweeney, line 1066: The way this stub message will appear is `Encountered stub: AVM2 flash.display.Graphics.drawTriangles() Positive culling not implemented`. I think it's better to change this to `"with positive culling"` rather than "Positive culling not implemented" (same with negative culling).

sleepycatcoding: Updated.

Lord-McSweeney (approved):

core/src/avm2/globals/flash/display/graphics.rs
```diff
@@ -1427,22 +1442,8 @@ fn handle_graphics_triangle_path<'gc>(
...
             let indices = &mut indices.iter();
 
-            while let Some((a, b, c)) = next_triangle(&vertices, indices, activation)? {
-                match culling {
-                    TriangleCulling::None => {
-                        drawing.draw_command(DrawCommand::MoveTo(a));
-
-                        drawing.draw_command(DrawCommand::LineTo(b));
-                        drawing.draw_command(DrawCommand::LineTo(c));
-                        drawing.draw_command(DrawCommand::LineTo(a));
```
torokati44, line 1437: Where did these `draw_command` calls go? 🤔

Lord-McSweeney: Into `draw_triangle_internal` :D!

sleepycatcoding: The code was exactly the same as in draw_triangle_internal so I just switched it over. I can switch back if that is preferred.

torokati44: Ah, right, I see! No need to switch back, it's nice this was. Just wasn't immediately obvious for me reading the diff.

PR #13674, opened 2023-10-22
avm2: Implement XMLList.parent

sleepycatcoding:


core/src/avm2/globals/XMLList.as
```diff
@@ -22,6 +22,7 @@ package {
         AS3 native function toXMLString():String;
         AS3 native function toString():String;
         AS3 native function comments():XMLList;
+        AS3 native function parent():Object;
```
Lord-McSweeney, line 25: I'm pretty sure this is also defined as a method on the `prototype`

Lord-McSweeney: Also, the return type should be `*`, to allow for returning `undefined` (also, that's the signature in playerglobals)

sleepycatcoding: Fixed.

Lord-McSweeney (approved):

Dinnerbone (approved):

torokati44:
I took the liberty of resolving a merge conflict manually.

PR #13675, opened 2023-10-22
avm2: Fix incorrect XMLList method signatures 

sleepycatcoding:


Lord-McSweeney:
Could you also fix some of the function parameter types (both for class methods and prototype methods)? `descendants()` should have argument `name:*="*"` instead of `name:Object`, and `processingInstructions()` should have `name:* = "*"` instead of `name:String = "*"`.

Lord-McSweeney (approved):

PR #13689, opened 2023-10-23
ci: Release nightlies automatically to Edge

Dinnerbone:
Secrets are already configured.

In theory this should just skip every release whenever we're already in review.
In practice... we need to try and see.


torokati44 (approved):
Alright, I'm all for trying and seeing!

PR #13492, opened 2023-10-10
avm2: Use root movie when running timer/event/broadcast closures

Lord-McSweeney:
Fixes #13490.

I think there's a few more places where this needs to be fixed, but it's unlikely that they will ever be called:

- FLV script data handlers
- `ExternalInterface` callback

For the future, I think a `Activation::from_context` that accepts a movie and checks the library for the AVM2 domain would be useful, but I think this is fine for now.

(I did confirm that this uses the root movie version.)

PR #13691, opened 2023-10-23
avm2: Implement Font.enumerateFonts and Font.registerFont

### LINKED ISSUE #8830: Air Transporter - Black Screen
### Describe the bug

`ERROR ruffle_core::avm2::events] Error dispatching event EventObject(EventObject { type: "addedToStage", class: flash.events::Event, ptr: 0x212255f83b0 }) to handler FunctionObject(FunctionObject(GcCell(Gc { ptr: 0x2120b778638 }))) : RustError("Font.enumerateFonts is a stub")`

### Expected behavior

should work

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

stub is a stub, but this will need to be implemented
[546844_2965.zip](https://github.com/ruffle-rs/ruffle/files/10254725/546844_2965.zip)

Dinnerbone:
Fixes #8830 (and possibly some others)

Probably makes some movies use their proper fonts, but none spring to mind at the moment (likely anything that logged about the `registerFont` stub would start looking better, at least)


Dinnerbone:
Confirmed that this makes the fonts in Habbo go from totally unreadable to perfect

sleepycatcoding (approved):
Looks good to me

core/src/avm2/globals/flash/text/font.rs
```diff
@@ -103,23 +64,9 @@ pub fn has_glyphs<'gc>(
...
-            .library
-            .library_for_movie_mut(movie)
-            .character_by_id(character_id)
-        {
-            return Ok(font.has_glyphs_for_str(&my_str).into());
-        }
+        return Ok(font.has_glyphs_for_str(&my_str).into());
     }
 
     Ok(Value::Undefined)
```
Lord-McSweeney, line 78: This would be better returning `false.into()` to be more clear about the return value (`hasGlyphs` returns a `bool`)

core/src/avm2/globals/flash/text/font.rs
```diff
@@ -129,18 +82,66 @@ pub fn has_glyphs<'gc>(
...
+            "enumerateFonts",
+            "with device fonts"
+        );
+    }
+
+    for font in activation.context.library.global_fonts() {
+        storage.push(FontObject::for_font(activation.context.gc_context, font_class, font).into());
+    }
+
+    if let Some(movie) = activation.caller_movie() {
```
Lord-McSweeney, line 104: This should always be `Some`- if it isn't, we have a bug. With #13492 it should always be set except in the strangest of cases (no SWF should be calling those cases).

Lord-McSweeney (approved):
Approved with a few minor nits

core/src/avm2/globals/flash/text/font.rs
Dinnerbone: Sure, it's existing code so I didn't consider it. Changing!

PR #13719, opened 2023-10-26
desktop: Add screenshot functionality to the desktop app

OctoCPPlus:
Allows you to take a screenshot of the stage using a button.

Feel free to suggest edits and changes, this code may be messy in places: I've never used rust before.


desktop/src/gui.rs
```diff
@@ -193,6 +195,13 @@ impl RuffleGui {
         player.set_volume(self.volume_controls.get_volume());
     }
 
+    fn is_taking_screenshot(&mut self) -> bool {
```
Lord-McSweeney, line 198: The name `is_taking_screenshot` sounds like this method doesn't change the state- could this just return `self.taking_screenshot`? (I think this approach would require adding a `set_taking_screenshot` function)

OctoCPPlus: Updated to use `set_taking_screenshot(bool state)` and `get_taking_screenshot() -> bool`

PR #13735, opened 2023-10-27
ci: Test that the Dockerfile in web is able to build the browser extensions

torokati44:
This should check weekly that https://github.com/ruffle-rs/ruffle/pull/13733 didn't regress, and the Firefox extension is buildable by the Dockerfile.

The question is, who is going to notice if this fails, and in what way?
Also, should we try passing in the environment variables containing the relevant repo secrets, so the signing step is also checked?

See for an example invocation: https://github.com/torokati44/ruffle/actions/runs/6661108191/job/18103430036

### DIFF
--- a/.github/workflows/test_extension_dockerfile.yml
+++ b/.github/workflows/test_extension_dockerfile.yml
@@ -1,0 +1,45 @@
+name: Test extension builder Dockerfile
+
+on:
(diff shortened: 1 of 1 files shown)


torokati44:
And indeed we got verification that this doesn't pass until #13733 is merged. :)

Dinnerbone (approved):

PR #13742, opened 2023-10-27
swf: Fix writing of a DefineButton2 with zero actions

sleepycatcoding:
Previously, `swf` crate always wrote out the action offset, but this should be 0 when there are no actions.

### DIFF
--- a/swf/src/write.rs
+++ b/swf/src/write.rs
@@ -973,7 +973,14 @@ impl<W: Write> Writer<W> {
                 }
                 writer_2.write_u8(0)?; // End button records
             }
-            writer.write_u16(record_data.len() as u16 + 2)?;
+
+            // Write action offset, this is explicitly 0 when there are no actions.
+            if button.actions.is_empty() {
+                writer.write_u16(0)?;
+            } else {
+                writer.write_u16(record_data.len() as u16 + 2)?;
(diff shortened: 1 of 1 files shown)


Dinnerbone:
```rust
#[test]
    fn write_define_button_2() {
        use crate::read::Reader;

        let button = Button {
            id: 3,
            is_track_as_menu: false,
            records: vec![ButtonRecord {
                states: ButtonState::UP
                    | ButtonState::OVER
                    | ButtonState::DOWN
                    | ButtonState::HIT_TEST,
                id: 2,
                depth: 1,
                matrix: Matrix::translate(Twips::from_pixels(2.0), Twips::from_pixels(3.0)),
                color_transform: ColorTransform::default(),
                filters: vec![],
                blend_mode: BlendMode::Normal,
            }],
            actions: vec![],
        };

        let mut buf = Vec::new();
        let mut writer = Writer::new(&mut buf, 15);
        writer.write_define_button_2(&button).unwrap();

        let mut reader = Reader::new(&buf, 15);
        let tag = reader.read_tag().unwrap();

        assert_eq!(tag, Tag::DefineButton2(Box::new(button)));
    }
```

Dinnerbone (approved):

PR #13745, opened 2023-10-27
swf: Fixed writing various tags correctly

Dinnerbone:


swf/src/write.rs
```diff
@@ -1821,8 +1821,9 @@ impl<W: Write> Writer<W> {
     }
 
     fn write_bevel_filter(&mut self, filter: &BevelFilter) -> Result<()> {
-        self.write_rgba(&filter.shadow_color)?;
+        // Note that the color order is wrong in the spec, it's hightlight then shadow.
```
torokati44, line 1820: Should we add this to our errata in the wiki?

torokati44: Also there's a typo here.

torokati44: Also I assume reading is already done in this order, and the actual appearance of the filter is expected. Right?

swf/src/write.rs
```diff
@@ -991,12 +991,7 @@ impl<W: Write> Writer<W> {
                 } else {
                     writer.write_u16(0)?;
                 }
-                let mut flags = action.conditions.bits();
-                if action.conditions.contains(ButtonActionCondition::KEY_PRESS) {
```
torokati44, line 995: Just curious - the writing bug was this condition here, right?

Dinnerbone: We pretended that the key press condition was `1<<9`, so when this bit wrote the `key<<9` in addition to the condition bit flags, it'd effectively write `(key | 1) << 9`

torokati44: Ah, okay, right, thanks.

swf/src/write.rs
```diff
@@ -2225,7 +2226,7 @@ impl<W: Write> Writer<W> {
         Ok(())
     }
 
-    fn write_define_text(&mut self, text: &Text) -> Result<()> {
+    fn write_define_text(&mut self, text: &Text, version: u8) -> Result<()> {
```
torokati44, line 2231: Nice to avoid the "boolean trap" by asking for the version as a number.
I guess this is a really internal function, so a note saying that only versions 1 and 2 are valid is not really necessary.

torokati44:
LGTM, pending that little note (and questions) regarding bevel filter. Really nice to see regression unit tests for every commit! :partying_face:

swf/src/write.rs
Dinnerbone: We do this in other similar write methods too, so I just copied the pattern :D

torokati44: Sure, it's perfectly fine this way.

swf/src/write.rs
Dinnerbone: Added to wiki and fixed typo! Reading is good yeah.

PR #13756, opened 2023-10-28
avm2: Add SWFv21 and above XML special case

sleepycatcoding:
Fixes SWF v21 and above XML appendChild test.


core/src/avm2/e4x.rs
```diff
@@ -1249,17 +1267,33 @@ pub fn maybe_escape_child<'gc>(
...
                 .construct(activation, &[string.into()])?;
-            Ok(xml.into())
+            return Ok(xml.into());
         }
-    } else {
-        Ok(child)
     }
+
+    // FIXME: This should use caller movie version.
+    if activation.context.swf.version() >= 21 {
```
Lord-McSweeney, line 1283: `activation.caller_movie().unwrap().version()`

sleepycatcoding: Originally used that, but some tests were panicking since caller_movie was missing?

Lord-McSweeney: I'm PRing a fix right now

Lord-McSweeney: Tests should no longer fail if you rebase

Lord-McSweeney:
Can you add a test that ensures that it's really the current SWF's version that's being used?

sleepycatcoding:
> Can you add a test that ensures that it's really the current SWF's version that's being used?

Added a test, I hope that it is sufficient. It uses `flash.display.Loader` to load a version 21 SWF, while the root is version 20.

Lord-McSweeney (approved):

Dinnerbone (approved):

PR #13760, opened 2023-10-28
avm2: Stub undocumented XML notification functions

evilpie:
Progresses https://github.com/ruffle-rs/ruffle/issues/12054. These function don't have `prototype` variants.

### DIFF
--- a/core/src/avm2/globals/XML.as
+++ b/core/src/avm2/globals/XML.as
@@ -96,6 +96,10 @@ package {
             return this.toJSON(k);
         }
 
+        // undocumented functions
+        AS3 native function notification():Function;
+        AS3 native function setNotification(f:Function):*;
+
         public static var ignoreComments:Boolean = true;
         public static var ignoreProcessingInstructions:Boolean = true;
         public static var ignoreWhitespace:Boolean = true;
--- a/core/src/avm2/globals/xml.rs
+++ b/core/src/avm2/globals/xml.rs
@@ -759,3 +759,29 @@ pub fn replace<'gc>(
     // 10. Return x
     Ok(xml.into())
 }
+
+pub fn set_notification<'gc>(
+    activation: &mut Activation<'_, 'gc>,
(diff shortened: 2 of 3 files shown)


evilpie:
Judging from the error stack trace I've seen, [mx.utils.XMLNotifier](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/mx/utils/XMLNotifier.html) must be using this under the hood.

core/src/avm2/globals/xml.rs
```diff
@@ -759,3 +759,29 @@ pub fn replace<'gc>(
...
+
+pub fn set_notification<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Object<'gc>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    avm2_stub_method!(activation, "XML", "setNotification");
+    let xml = this.as_xml_object().unwrap();
+    let node = xml.node();
+    let fun = args[0];
```
Lord-McSweeney, line 771: This should probably be
```
let fun = args.try_get_object(activation, 0);
```
which reduces the following code to 
```
    node.set_notification(
        fun.as_function_object(),
        activation.context.gc_context,
    );
```

evilpie: Nice, thanks.

Aaron1011 (approved):

core/src/avm2/globals/XML.as
```diff
@@ -96,6 +96,10 @@ package {
             return this.toJSON(k);
         }
 
+        // undocumented functions
+        AS3 native function notification():Function;
+        AS3 native function setNotification(f:Function);
```
Lord-McSweeney, line 101: Playerglobal signature says that this returns `:*`- can you add it here?

Lord-McSweeney (approved):

sleepycatcoding (approved):

PR #13772, opened 2023-10-29
avm2: Fix return value of SharedObject.flush

n0samu:
Fixes the "save failed!" problem in Phoenotopia (https://github.com/ruffle-rs/ruffle/issues/9226#issuecomment-1657242890)


Dinnerbone (approved):
Thanks for the test too!

PR #13781, opened 2023-10-29
core: Fix BG music not stopping in Samorost (close #963)

### LINKED ISSUE #963: Samorost: Strange behaviours after switching movies
In Samorost, each level exists as a different movie file. When switching to the next level, the background and music are meant to change. However, in Ruffle, the previous background and music track overlap with the next one.
Each level works properly if loaded independently.

**Video**: https://streamable.com/l4b2jz
**Video (rybar.swf, by itself)**: https://streamable.com/u3yd8p
**Game file**: https://amanita-design.net/samorost-1/

vadosnaprimer:
`MovieClip avm1_unload()` was already stopping the current stream at https://github.com/ruffle-rs/ruffle/blob/5bebebddbe3c39362ba520cb0c9c28b3cb7f72a3/core/src/display_object/movie_clip.rs#L2924C18-L2924C18 in Samorost, the BG music that keeps playing after switching SWFs in ruffle is not a stream but a sound.

I don't know how to make it stop only the sound that's currently playing, so here's a "fix" that just stops them all.

### DIFF
--- a/core/src/loader.rs
+++ b/core/src/loader.rs
@@ -768,6 +768,7 @@ impl<'gc> Loader<'gc> {
                 if let Some(mut mc) = clip.as_movie_clip() {
                     if !uc.is_action_script_3() {
                         mc.avm1_unload(uc);
+                        uc.stop_all_sounds();
                     }
 
                     // Before the actual SWF is loaded, an initial loading state is entered.


Lord-McSweeney:
Thank you for the contribution! Unfortunately this fix is (probably) not the right way to do it. Can you elaborate on what a "sound" is when you stated that the BG music "is not a stream but a sound"? I don't see any `StartSound` tags or Sound AS code in the SWF.

vadosnaprimer:
Are there examples of this in the code that I could use? I wouldn't be able to add it from scratch, only by copying an existing approach.

vadosnaprimer:
@Lord-McSweeney `audio_manager.stop_all_sounds()` does basically what you describe.

`start_sound_1()` calls `AudioManager start_sound()` which adds to this list:

https://github.com/ruffle-rs/ruffle/blob/fca17a0061d30342ab06f4b0443212ca93d1c3f8/core/src/backend/audio.rs#L346-L347

`AudioManager stop_all_sounds()` clears that list, and then `AudioMixer stop_all_sounds()` iterates through actively playing `AudioMixer sound_instances` and removes each of them from the arena, also clearing it after that.

https://github.com/ruffle-rs/ruffle/blob/fca17a0061d30342ab06f4b0443212ca93d1c3f8/core/src/backend/audio/mixer.rs#L636-L649

Lord-McSweeney:
Yes, except this implementation should only stop sounds that were played in a specific `MovieClip`, not every sound.

vadosnaprimer:
> I think in that case the right way to do it would be to add a list of handles of currently-playing sounds to `MovieClipData`, add handles to that list whenever the `StartSound` tag starts a sound (see `MovieClip::start_sound_1`), then iterate over all of those handles and call `context.stop_sounds_with_handle(...)` on them when the movie is unloaded.

If a sound has ended by itself or by the game before unloading the movie, would I also need to remove it from the list of per-clip handles? I was unable to find a place when sounds get removed when they simply end, so I don't know how I'd duplicate the removal regarding my duplicate list.

n0samu:
Superseded by #17016. Thank you for your work though, we appreciate it!

PR #13783, opened 2023-10-29
avm2: Fix E4X [[Replace]] implementation bug

### LINKED ISSUE #13780: Implementation mistake in ECMA-357 9.1.1.12 [[Replace]]
Our [implementation of [[Replace]]](https://github.com/ruffle-rs/ruffle/blob/5bebebddbe3c39362ba520cb0c9c28b3cb7f72a3/core/src/avm2/e4x.rs#L491) doesn't seem to match the spec;
> Step 5. If Type(V) is XML and V.[[Class]] ∈ {"element", "comment", "processing-instruction", "text"}

In our code the `if` only checks the first part `Type(V) is XML` and not the second:
```rust
if let Some(xml) = value.as_object().and_then(|x| x.as_xml_object()) {
```
The next line returns early for attributes, which doesn't seem to correspond to anything in the specification.

@sleepycatcoding

sleepycatcoding:
We were incorrectly returning for XML attributes. Fixes #13780.


core/src/avm2/e4x.rs
```diff
@@ -470,11 +470,17 @@ impl<'gc> E4XNode<'gc> {
...
 
         // 5. If Type(V) is XML and V.[[Class]] ∈ {"element", "comment", "processing-instruction", "text"}
-        if let Some(xml) = value.as_object().and_then(|x| x.as_xml_object()) {
-            if matches!(*xml.node().kind(), E4XNodeKind::Attribute(_)) {
-                return Ok(());
-            }
+        if let Some(xml) = value
+            .as_object()
+            .and_then(|x| x.as_xml_object())
+            .and_then(|x| {
```
evilpie, line 476: You can use `filter` instead.

sleepycatcoding: Thanks. I had no idea that `filter` also existed for `Option`.

Dinnerbone (approved):

PR #13792, opened 2023-10-30
avm2: XMLNode is_ kind helper methods

evilpie:


render/src/tessellator.rs
```diff
@@ -126,7 +126,7 @@ impl ShapeTessellator {
             } else if !self.is_stroke && next_is_stroke {
                 // Bake solid color fills followed by strokes into a single draw call, and adjust
                 // the index count to omit the strokes when rendering this shape as a mask.
-                debug_assert!(self.mask_index_count.is_none());
+                // debug_assert!(self.mask_index_count.is_none());
```
Dinnerbone, line 129: Debug, or just tired of this? :D

evilpie: This PR isn't ready yet, but I do get annoyed by this assert :frowning:

core/src/avm2/globals/xml.rs
```diff
@@ -477,7 +474,7 @@ pub fn text<'gc>(
     let nodes = if let E4XNodeKind::Element { children, .. } = &*xml.node().kind() {
         children
             .iter()
-            .filter(|node| matches!(&*node.kind(), E4XNodeKind::Text(_)))
+            .filter(|node| node.is_text())
```
Aaron1011, line 477: This now matches CData as well as text - is that intentional?

evilpie: Cdata is text. The previous code was wrong.

Aaron1011 (approved):

core/src/avm2/object/xml_list_object.rs
```diff
@@ -766,20 +766,14 @@ impl<'gc> TObject<'gc> for XmlListObject<'gc> {
                                 .xml_object_child(0, activation)
                                 .expect("List length was just verified");
 
-                            if matches!(
```
sleepycatcoding, line 769: According to avmplus this should not match CData? https://github.com/adobe/avmplus/blob/858d034a3bd3a54d9b70909386435cf4aec81d21/core/XMLListObject.cpp#L365

core/src/avm2/object/xml_list_object.rs
```diff
@@ -766,20 +766,14 @@ impl<'gc> TObject<'gc> for XmlListObject<'gc> {
...
-                                E4XNodeKind::Attribute(_) | E4XNodeKind::Text(_)
-                            ) {
+                            if xml.node().is_attribute() || xml.node().is_text() {
                                 value = Value::Object(xml.into())
                                     .coerce_to_string(activation)?
                                     .into();
                             }
                         }
                     } else if let Some(xml) = value.as_object().and_then(|x| x.as_xml_object()) {
-                        if matches!(
```
sleepycatcoding, line 779: Same here.

core/src/avm2/object/xml_list_object.rs
evilpie: Seems you are right. I guess we need to replicate the bug in avmplus.

evilpie: I added a note.

sleepycatcoding (approved):

Dinnerbone:
Any chance of a test to prove and cement the learnings of the avmplus behaviour?
Not a blocker, but it'd be nice!

torokati44:
I agree!

PR #13798, opened 2023-10-30
Suport uploading ATFTextureData::CompressedRawAlpha

Aaron1011:
Fortunately, this is very simple - we can just take the DXT5 data and upload it directly to the wgpu texture.


torokati44:
A test for this would be nice. 😶 Any chance of adding one?

Aaron1011:
I've added a test

tests/tests/swfs/avm2/stage3d_texture_bytearray_compressed_raw_alpha/test.toml
```diff
@@ -0,0 +1,15 @@
+num_frames = 1
+
+[image_comparisons.output]
+tolerance = 2
+max_outliers = 66
+
+[player_options]
+with_renderer = { optional = true, sample_count = 1 }
```
Lord-McSweeney, line 8: Shouldn't this be `optional = false`?

Dinnerbone (approved):

PR #13834, opened 2023-11-01
avm2: TLF GroupElement improvements

evilpie:
Depends on #13226. Makes #13803 somewhat playable, but the text is invisible.


core/src/avm2/globals/flash/text/engine/GroupElement.as
```diff
@@ -24,18 +32,22 @@ package flash.text.engine {
...
                 }
-            } else {
-                stub_method("flash.text.engine.GroupElement", "replaceElements", "with beginIndex != endIndex or newElements.length != 1");
-                return new Vector.<ContentElement>();
             }
+            return old;
+        }
+
+        public function splitTextElement(elementIndex:int, splitIndex:int): TextElement {
+            var textElement: TextElement = this._elements[elementIndex];
```
Lord-McSweeney, line 45: What happens here if `this._elements[elementIndex]` is not a `TextElement`? Does Flash also throw a coercion failure?

SN902:
[pony-creator.zip](https://github.com/ruffle-rs/ruffle/files/13235211/pony-creator.zip) now gets stuck in endless loading after clicking "Play Game" button

evilpie:
> [pony-creator.zip](https://github.com/ruffle-rs/ruffle/files/13235211/pony-creator.zip) now gets stuck in endless loading after clicking "Play Game" button

Do you mean after this PR? I tried with just #13226 and it also seemed to hang.

SN902:
Oh, the main root of the problem is #13226

I haven't tested it before, only on the last nightly build and this one.

core/src/avm2/globals/flash/text/engine/GroupElement.as
evilpie: Flash seems to throw:
```
ArgumentError: Error #2004: One of the parameters is invalid.
```

Lord-McSweeney:
Yeah, I think I figured out what it wants, but I'm not sure of the exact conditions under which it should be done. See my comment: https://github.com/ruffle-rs/ruffle/pull/13226#issuecomment-1732223322

core/src/avm2/globals/flash/text/engine/GroupElement.as
Lord-McSweeney: Can this also throw in that case?

danielhjacobs:
#13226 was merged so this will need a rebase.

core/src/avm2/globals/flash/text/engine/GroupElement.as
```diff
@@ -0,0 +1,85 @@
...
+                this._elements = elements.AS3::concat();
+            }
+        }
+
+        public function replaceElements(beginIndex:int, endIndex:int, newElements:Vector.<ContentElement>):Vector.<ContentElement> {
+            // This some sort of special case that doesn't throw.
+            if (beginIndex == endIndex && newElements == null) {
+                return null;
+            }
+            if (beginIndex < 0 || beginIndex > this._elements.length ||
```
Lord-McSweeney, line 42: Should these `>`s be replaced with `>=`s?

evilpie: I don't think so. AS seems to allow adding new elements at the end. Also tested by 
> /// replaceElements(3, 3, replacement)

Dinnerbone (approved):

Issue #13688, opened 2023-10-23
Videos on https://www.slaveryinnewyork.org are not starting


torokati44:
### Describe the bug

The videos on the Visitor Responses section don't start, such as this one:
https://www.slaveryinnewyork.org/gallery_2_responses.htm

All of the necessary files are thankfully there:
https://www.slaveryinnewyork.org/flash/gallery2.swf
https://www.slaveryinnewyork.org/flash/SkinUnderPlaySeekStop.swf
https://www.slaveryinnewyork.org/flash/Gallery2Responses2.flv

Source: https://news.ycombinator.com/item?id=37981663

### Expected behavior

The videos start playing, probably with a little video player interface, judging by the associated miscellaneous content.

### Affected platform

Browser's extension

### Operating system

-

### Browser

-

n0samu:
The audio from the videos plays now, and the controls are functional, but the video itself is not visible.

1bugproof:
Was doing tests, this happens because FLVPlayback constructor does `new VideoPlayer(0, 0)` so Video bounds are (0, 0).

Then, `_loc1_.setSize(320, 245)` -> super.width = 320, but Ruffle's generic set_width works by scaling `target_scale = value / current_width`

With `current_width = 0`, scale becomes 0, so the assignment has no effect.

Issue #11524, opened 2023-06-13
Steamlands: Menu borders are rendered incorrectly


n0samu:
### Describe the bug

[Download SWF](https://github.com/ruffle-rs/ruffle/files/10396365/Steaml_P_P_P.zip)

Steamlands is playable now, but as mentioned in https://github.com/ruffle-rs/ruffle/issues/9122#issuecomment-1589399613, there is a strange rendering issue with the menu borders. For example, here is how the world map menu looks in Ruffle:
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/a60558e5-d559-4de6-ac49-a516d940d80b)

### Expected behavior

Here is how the world map menu looks in Flash Player:
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/46e7b7bd-3c03-423f-bd98-789f3fc21b9b)

### Affected platform

Desktop app

### Operating system

Windows 10

n0samu:
This happens because 9-slice scaling is unsupported. This will be fixed whenever #10611 is finished and merged.

PR #13835, opened 2023-11-01
core,avm1,avm2: Stub 9-slice scaling (`scale9Grid`)

### LINKED ISSUE #3302: [NSFW] Lust planet - Text and shapes at the title screen barely make sense
Chrome 88.0.4324.150
Nightly 2021-02-16

http://eye.swfchan.com/flash.asp?id=182996&n=Lust+Planet+0.79b+by+Supersatanson+%28SssHentaiGameFlashPlatformerPornSexyAlienMonsterGirls%29.swf

The title screen is a hot mess - text is barely readable and the shapes are really different in the Ruffle playback:

Adobe:
image
Ruffle:
image

I know that the text glow/outlining isn't supported yet, but what about the shapes?

If you play the game further you will see the various problems with the text, but mostly it's either wrong font, no effects/glow, or the text is slightly out of bounds (a bit annoying but not that much). The title screen problems are looking the most severe so far, didn't play the game that much to tell otherwise.

torokati44:
This is a rebased version of the core, avm1, and avm2 parts of #10611, minus rendering.

~~This is a rebased version of #10611. Only minor changes done, like a typo fix, removing some back-and-forth changes, and turning the added visual test into an actually runnable test.~~

~~I'm not sure if `render_with_scaling_grid` is still needed with our current (caching, blending, filtering, masking, scrolling, etc.) renderer setup, or if only some part of it should be thrown out, or if it's even called from the right place.~~

~~Fixes #3302.~~
~~Improves, but does not fix, #13688.~~
~~Also related: #9829~~

### DIFF
--- a/core/src/avm2/globals/flash/display/DisplayObject.as
+++ b/core/src/avm2/globals/flash/display/DisplayObject.as
@@ -72,14 +72,9 @@ package flash.display {
 
         public native function get scaleZ():Number;
         public native function set scaleZ(value:Number):void;
-        
-        public function get scale9Grid():Rectangle {
-            stub_getter("flash.display.DisplayObject", "scale9Grid");
-            return null;
-        }
-        public function set scale9Grid(value:Rectangle):void {
-            stub_setter("flash.display.DisplayObject", "scale9Grid");
-        }
+
+        public native function get scale9Grid():Rectangle;
+        public native function set scale9Grid(value:Rectangle):void;
 
         public native function get name():String;
         public native function set name(value:String):void;
--- a/core/src/display_object/movie_clip.rs
+++ b/core/src/display_object/movie_clip.rs
@@ -598,6 +598,10 @@ impl<'gc> MovieClip<'gc> {
                     .0
                     .write(context.gc_context)
                     .define_morph_shape(context, reader, 2),
+                TagCode::DefineScalingGrid => self
+                    .0
+                    .write(context.gc_context)
+                    .define_scaling_grid(context, reader),
                 TagCode::DefineShape => self
                     .0
                     .write(context.gc_context)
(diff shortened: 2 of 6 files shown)


Dinnerbone:
It's worth noting that it sounded very much like the approach in that PR was not the approach we should be taking, just a learning exercise along the way.

torokati44:
Maybe we can now utilize our bitmap caching mechanism instead?

torokati44:
(Also I mostly just wanted to see how much this would improve https://github.com/ruffle-rs/ruffle/issues/13688, as opposed to just stubbing the AVM side, and since making it work on top of `master` wasn't entirely trivial, I thought I'd share what I have.)

n0samu:
Also worth testing #11524!

core/src/avm2/globals/flash/display/DisplayObject.as
```diff
@@ -1,5 +1,6 @@
-package flash.display {
-    
+package flash.display
```
Lord-McSweeney, line 1: The brackets in this file seem to have been moved to the next line each- to maintain formatting with the rest of the AS globals, they should remain on the previous line.

Dinnerbone (approved):

core/src/display_object/movie_clip.rs
```diff
@@ -3473,6 +3477,25 @@ impl<'gc, 'a> MovieClipData<'gc> {
         Ok(())
     }
 
+    #[inline]
+    fn define_scaling_grid(
```
Aaron1011, line 3481: Should we also add a stub here, like we do in the actionscript setter?

torokati44: Oh, this was a bit late. I don't know, maybe? Any second opinions...?

Dinnerbone: I'd say not, this is doing what it's supposed to

PR #13858, opened 2023-11-03
chore: Add 'content location' as a required field to issue reports

Dinnerbone:


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -30,6 +30,12 @@ body:
       description: A clear and concise description of what you expected to happen.
     validations:
       required: true
+  - type: textarea
+    attributes:
+      label: Content Location
+      description: Where can we find the Flash content related to this issue? You can attach it if you have the SWF, or link to a website that has it.
```
n0samu, line 36: This should replace the existing bullet point in the "describe the bug" section:
`If this is an issue with a specific SWF file, please provide a link to the SWF file, or you can attach the SWF by zipping it and dragging it onto the issue box.`
And we should keep the reminder about zipping the SWF file, since SWF files can't be attached directly. So I'd say:
> Where can we find the Flash content related to this issue? You can link to a webpage that has it, or you can attach the SWF by zipping it and dragging it onto the issue box.

Dinnerbone: Updated!

PR #13870, opened 2023-11-04
avm2: Small XML fixes

sleepycatcoding:


core/src/avm2/object/xml_list_object.rs
```diff
@@ -135,6 +137,18 @@ impl<'gc> XmlListObject<'gc> {
         )
     }
 
+    pub fn to_xml_string(&self, activation: &mut Activation<'_, 'gc>) -> AvmString<'gc> {
```
torokati44, line 140: I guess clippy would like this better as `as_xml_string`...

sleepycatcoding: Should I rename this or allow the lint?

torokati44: I think renaming it is fine.

core/src/avm2/globals/xml.rs
```diff
@@ -600,14 +599,22 @@ pub fn insert_child_after<'gc>(
     }
 
     // 3. Else if Type(child1) is XML
-    if let Some(child1) = child1.and_then(|x| x.as_xml_object()) {
+    if let Some(child1) = child1.as_object().and_then(|x| {
+        if let Some(xml) = x.as_xml_object() {
+            return Some(*xml.node());
+        } else if let Some(list) = x.as_xml_list_object() {
```
evilpie, line 605: Please add a note that this is non-standard avmplus behavior.

sleepycatcoding: Added.

core/src/avm2/object/xml_list_object.rs
sleepycatcoding: Renamed.

Dinnerbone (approved):

Issue #13876, opened 2023-11-05
Issues with depth (z-index)


ChaseWaylon:
### Describe the bug

Replaced the old (2021-09-06) self hosted ruffle version to a newest (2023-11-05) and some bugs surfaced in my SWFs.

All code is AS2.

First, loadClip() couldn't load the swf, because the container for it was created in a depth that new ruffle version is not supporting anymore:

createEmptyMovieClip("container", -99999);

I increased the depth to -9999 and swf is being loaded now (but now it covers all other MCs as they are by default lower than -9999).

So it looks like the problem is in the new max limit of the depth and it broke my original SWF created back in the Flash era.

### Expected behavior

Thank you for the good work!

### Affected platform

Self-hosted version

### Operating system

Win 11

### Browser

Chrome

Lord-McSweeney:
This is probably a regression caused by #8873.

n0samu:
@ChaseWaylon This should be fixed in the latest releases thanks to #13889. Can you check and let us know? Thanks!

Lord-McSweeney:
Closing as probably resolved.

Issue #9367, opened 2023-01-31
ひつまぶし


p0008874:
### Describe the bug

https://user-images.githubusercontent.com/75534590/215777571-9f5019f1-ba17-4fc1-9e02-403894490cbf.mp4

```
2023-01-31T13:30:02.938668Z  INFO ruffle_core::avm1::globals: clearInterval: Timer 0 does not exist
2023-01-31T13:30:02.938941Z  INFO ruffle_core::avm1::globals: clearInterval: Timer 0 does not exist
2023-01-31T13:30:02.939658Z  INFO ruffle_core::avm1::globals: clearInterval: Timer 0 does not exist
2023-01-31T13:30:02.939877Z  INFO ruffle_core::avm1::globals: clearInterval: Timer 0 does not exist
2023-01-31T13:30:07.076458Z  INFO ruffle_core::avm1::globals: clearInterval: Timer 0 does not exist
2023-01-31T13:30:07.076667Z  INFO ruffle_core::avm1::globals: clearInterval: Timer 0 does not exist
2023-01-31T13:30:07.077521Z  INFO ruffle_core::avm1::globals: clearInterval: Timer 0 does not exist
2023-01-31T13:30:07.077693Z  INFO ruffle_core::avm1::globals: clearInterval: Timer 2 does not exist
```

### Expected behavior

https://user-images.githubusercontent.com/75534590/215777784-298f41a3-92b6-46b0-80ad-651c7e3f511f.mp4

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

[1mlBj1H1Z5SmDfRcg5-GS2r-zWI-WYn9E.zip](https://github.com/ruffle-rs/ruffle/files/10546947/1mlBj1H1Z5SmDfRcg5-GS2r-zWI-WYn9E.zip)

p0008874:
![image](https://github.com/ruffle-rs/ruffle/assets/75534590/4b5421fc-7b52-4636-ad4e-16b6d47cb345)
Is seen Ruffle did not remove prouge2 on game start.
Frame 3 (Gamestart) in JPEXS:
```
Mouse.hide();
var i;
var j;
var k;
var m;
var n;
var xm;
var ym;
var val;
var str;
var AID;
var EID;
var JID;
var nID;
clearInterval(_root.JID);
clearInterval(_root.EID);
clearInterval(_root.AID);
clearInterval(_root.nID);
_root.RETRY.removeMovieClip();
[...]

PR #13889, opened 2023-11-07
avm1: Correctly mark unloaded movies as unloaded instead of relying on clip depth

### LINKED ISSUE #9522: avm1: Super Chick Sisters no longer works
### Describe the bug

- Run / Open https://uploads.ungrounded.net/396000/396372_superChickSisters.swf.
- Click "Enter code", type "gopam" and click "ok".
- Notice the screen is almost completely gray.

new

The issue was introduced in [nightly-2023-02-07](https://github.com/ruffle-rs/ruffle/releases/tag/nightly-2023-02-07). Bisection says that 755425ebfaa8f9bca381411309a157fe80fbbfd6 (#8873) is the first bad commit.

### Expected behavior

In nightly-2023-02-06 and older releases:
old

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

Web builds are also affected.

### LINKED ISSUE #9886: Negative depth MC does not call onUnload after unloadMovie
### Describe the bug

To reproduce:

Movie Clip is placed on stage manually (note depth is negative)
Movie Clip has onUnload function inside of it
Movie Clip is destroyed with unloadMovie (because depth is negative)
onUnload function is not called in Ruffle, but is called in Flash Player

Note: using swapDepths to make its depth positive and subsequently destroying it with removeMovieClip will cause onUnload to be called

### Expected behavior

onUnload is called when unloadMovie is called on an MC with negative depth

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

Chrome

### Additional information

_No response_

### LINKED ISSUE #10633: Sky Cops: Ocean background not shown
### Describe the bug

Link to SWF: https://www.playgb.com/swfs/1_3.swf

In [Sky Cops](https://www.playgb.com/play/sky-cops), after you click the Start Mission button, you'll be flying over the ocean towards an island. But since Ruffle version 2023-02-07, the ocean is not shown anymore.
image

### Expected behavior

The ocean should be visible.
image

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

This is a regression caused by commit 755425ebf from #8873.

Lord-McSweeney:
Fixes #9522 and fixes #10633. I'm pretty sure that this will also f​ix #9886 and #13876, but I haven't created test SWFs for those.

Logic is from [Toad06's branch](https://github.com/ruffle-rs/ruffle/compare/master...Toad06:ruffle:unload_test), mentioned in #10633.

When merged, #14276 should include enough test coverage.

### DIFF
--- a/core/src/display_object.rs
+++ b/core/src/display_object.rs
@@ -556,6 +556,11 @@ impl<'gc> DisplayObjectBase<'gc> {
         self.flags.contains(DisplayObjectFlags::AVM1_REMOVED)
     }
 
+    fn avm1_pending_removal(&self) -> bool {
+        self.flags
+            .contains(DisplayObjectFlags::AVM1_PENDING_REMOVAL)
+    }
+
     pub fn should_skip_next_enter_frame(&self) -> bool {
         self.flags
             .contains(DisplayObjectFlags::SKIP_NEXT_ENTER_FRAME)
@@ -568,6 +573,11 @@ impl<'gc> DisplayObjectBase<'gc> {
 
     fn set_avm1_removed(&mut self, value: bool) {
         self.flags.set(DisplayObjectFlags::AVM1_REMOVED, value);
+    }
+
+    fn set_avm1_pending_removal(&mut self, value: bool) {
+        self.flags
+            .set(DisplayObjectFlags::AVM1_PENDING_REMOVAL, value);
     }
 
     fn scale_rotation_cached(&self) -> bool {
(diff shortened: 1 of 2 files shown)


Dinnerbone (approved):
Code wise this looks okay but I'd very much appreciate some manual testing on AVM1 content to confirm no regressions. This has been a very sore area before.

Toad06:
> This has been a very sore area before.

With or without this PR, there still exist issues when dealing with movie clips at a negative depth, unfortunately. #9367 has a small game that illustrates the problem (I also commented the issue).

And that's actually the reason why I didn't open a PR: on one hand, I think the current approach is fine and safe; on the other hand, I have the feeling that adding this flag shouldn't be necessary as it may just hide the real cause of the bug (or an unimplemented behavior) somewhere else in the code base.

Lord-McSweeney:
Hmm, in that case I would guess you're more likely correct. If creating a new `MovieClip` with a negative depth _does_ result in the MC being destroyed on the next frame, then I would guess that the regressed games switch an MC to a negative depth- so we have to record the original depth, and destroy it on the next frame if the _original_ depth is negative. I'll do some further testing on this soon.

Toad06 (approved):

torokati44:
Since a consensus has been reached that this is okay, and is a rather simple change (the logic itself could be undone even just by looking a the code if it turns out to be wrong), and this is blocking another PR, I see no reason to postpone merging it any longer.

Issue #13079, opened 2023-09-05
Text is shown in wrong size(width of column)


sidmike:
### Describe the bug

I'm trying to run old runes.swf game through Ruffle, but Ruffle is displaying text at the wrong size. The column of text looks very narrow and long. It should not look like that.
[runes.zip](https://github.com/ruffle-rs/ruffle/files/12528651/runes.zip)

### Expected behavior

The column of text looks very narrow and long. It should not look like that. It should be wider for sure.

### Affected platform

Online demo

### Operating system

MacOs Ventura 13.5

### Browser

Google Chrome 116

### Additional information

Its written on Action Script 3

sidmike:
Forgot to mention what @n0samu already told regarding this bug:

 _...It looks similar to some AS2 issues I saw before but I don’t think I’ve seen it in AS3. Unless it’s related to unsupported TextFormat stuff. It’s worth checking..._

n0samu:
Looks like a similar issue to #13026, but it's a simpler reproduction, which is appreciated

sidmike:
yea, exactly same problem! Any ideas on what can be done here from Ruffle side?

Lord-McSweeney:
Fixed by #13923.

sidmike:
works like hell! thanks!

Issue #13026, opened 2023-08-31
Text paragraphs are cut off in Creeper World


Deepfriedice:
### Describe the bug

When playing the original Creeper World games, large paragraphs of text appear to be cut off at roughly the ~25% line. This is easiest to see on the "How To Play"->"Units" page. Note that the text is NOT just wrapping early - most of the text is not being displayed at all.
![Screenshot_20230901_021458](https://github.com/ruffle-rs/ruffle/assets/30456423/a20bc129-8b88-4dda-9e30-eec001466c66)

This doesn't seem to affect non-paragraph text, such as the "How To Play"->"Strategy Tips" page.
![Screenshot_20230901_024829](https://github.com/ruffle-rs/ruffle/assets/30456423/d4e4ddaa-0043-4fd1-9844-6e7cfa97c615)

The version of Creeper World I'm using is available for download here [here](http://knucklecracker.com/creeperworld/web/cwe-1023a-pack.zip
), but the same issue can be more easily seen on the (nearly identical) browser version [here](http://knucklecracker.com/creeperworld/playcwe.php).

### Expected behavior

![Screenshot_20230901_021648](https://github.com/ruffle-rs/ruffle/assets/30456423/7fc339d6-73a1-494c-aca0-8d074cc3a5e6)

### Affected platform

Desktop app

### Operating system

openSUSE Tumbleweed 20230823

### Additional information

These screenshots were taken from Ruffle nightly-2023-08-31, running with the "release" flag.
This seems very similar to the https://github.com/ruffle-rs/ruffle/issues/12603 and https://github.com/ruffle-rs/ruffle/issues/9690 issues, however in those issues the text is only wrapped early rather than actually hidden.

Issue #12603, opened 2023-08-06
Text on Shootorial does not work properly


Fancy2209:
### Describe the bug

I only tested Part 1 of Shootorial
https://game24537.konggames.com/gamez/0002/4537/live/FlashGameTutorialShell.swf?kongregate_game_version=1224203434
![imagem](https://github.com/ruffle-rs/ruffle/assets/64917206/a7e6d6e5-f849-4676-8e9b-c5c70f554ab3)

### Expected behavior

![imagem](https://github.com/ruffle-rs/ruffle/assets/64917206/2c986395-9832-481b-b051-88f69364efec)

### Affected platform

Self-hosted version

### Operating system

Windows 11

### Browser

Mozilla Firefox 116.0.1

Lord-McSweeney:
Fixed by #13923.

Issue #12387, opened 2023-07-29
Bloody Fun Day: Tutorial text formatting and Critter spawn RNG


waspennator:
### Describe the bug

https://www.newgrounds.com/portal/view/486608?emulate=flash

1: The text boxes that spawn during the tutorial don't seem to be formatted correctly, resulting in the text being hard to read.

2: When playing any 5 level or infinite (does not happen on tutorial), the critters all seem to spawn in as one singular color.

1:
![Screenshot_20230728_183154_Chrome](https://github.com/ruffle-rs/ruffle/assets/78768499/56a7e871-4658-4ef9-96b4-4adc4815e4b3)

2:
![Screenshot_20230728_182143_Chrome](https://github.com/ruffle-rs/ruffle/assets/78768499/81f6610b-3106-4e77-a866-718127b6483f)

### Expected behavior

The text should be properly formatted in the tutorial and the critter RNG shouldn't just spawn em all as the same color in 5 level or infinite mode.

### Affected platform

Browser's extension

### Operating system

Android 13

### Browser

Chrome 115

Aaron1011:
The critter spawn issue is caused by a bug in how we handle integer keys in a `Dictionary`.

This code:

```actionscript
import flash.utils.Dictionary;

var dict = new Dictionary();
dict[100] = "Test";
dict[1.2] = "Hello";

for (var key in dict) {
	trace("Key: " + key + " type: " + typeof(key));
}
```

prints this under flash:

```
Key: 100 type: number
Key: 1.2 type: string
```

but this under Ruffle:

```
Key: 100 type: string
Key: 1.2 type: string
```

The SWF tries to convert the key to a `uint,` which gives us `null` (and later `0`) when we have a string instead of a `number`.

waspennator:
Still happening as of latest nightly.

waspennator:
The 1st issue seems to be fixed now, but the critter rng is still borked.

Lord-McSweeney:
Fixed by #14721

PR #13923, opened 2023-11-09
core: Fix behaviour of set_width and set_height on text with AutoSize enabled

Deepfriedice:
This modifies EditText's `set_width` and `set_height` methods to store the requested size inside the `EditTextData` struct, and then call `relayout`. The `relayout` method can then switch between using the height and width provided, or those calculated by `LayoutBox::lower_from_text_spans`, without losing any information.

This still doesn't exactly replicate the behavior of Adobe Flash, but it seems to be significantly closer.

This resolves these issues:
- https://github.com/ruffle-rs/ruffle/issues/9690
- https://github.com/ruffle-rs/ruffle/issues/10100
- https://github.com/ruffle-rs/ruffle/issues/12387
- https://github.com/ruffle-rs/ruffle/issues/12603
- https://github.com/ruffle-rs/ruffle/issues/13026
- https://github.com/ruffle-rs/ruffle/issues/13079


Dinnerbone (approved):
Thank you very much!

PR #13950, opened 2023-11-10
core: Don't invalidate when setting filters to the same value

Dinnerbone:
There seems to be a lot of games that do `mc.filters = [....]` every single frame, so this should remove the performance penalty that brings.


Aaron1011 (approved):

torokati44:
Very much approving the concept and the intention, just haven't had time to read through the code changes yet...

torokati44:
Okay, two checkmarks should be enough, no need to block waiting for me.

Issue #13362, opened 2023-09-29
Skrillex Quest fails to display 3D content (Alternativa Engine)


skiesofravens:
### Describe the bug

https://ooooooooo.ooo/static/?1c1335c0-56af-3fce-2701-d90de13a730d
and [skrillexquest.zip because it has a ton of assets](https://github.com/ruffle-rs/ruffle/files/12772392/skrillexquest.zip)

When running in a browser, Skrillex Quest never displays the 3D model of the dust, and instead shows this in the console. Gameplay cannot begin without this cutscene playing.

![image](https://github.com/ruffle-rs/ruffle/assets/138401463/be329f98-beee-49b2-bfe0-9f055fc0b1bf)
Then TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: toString) errors endlessly appear, flooding the console with them.
![image](https://github.com/ruffle-rs/ruffle/assets/138401463/a1daf6ad-0f96-47d4-8cfd-a84ba3c170be)
Standalone Nightly is the same.

### Expected behavior

![image](https://github.com/ruffle-rs/ruffle/assets/138401463/a19234d2-07ea-4dae-a904-586de1670e1c)

### Affected platform

Browser's extension

### Operating system

Windows 10 22H2

### Browser

Mozilla Firefox 118.0.1 64-bit

### Additional information

Both extension and standalone are Nightly 9-29-2023 releases.

Aaron1011:
This needs improved XML namespace handling. XML child lookup current never matches a multiname with multiple namespaces: https://github.com/ruffle-rs/ruffle/blob/f140e7a14b379993625b07524c4453ae98592384/core/src/avm2/e4x.rs#L861-L871

However, this SWF defines an XML document with the namespace `http://www.collada.org/2005/11/COLLADASchema`, and expects to be able to do `getproperty` with a multiname containing (among other things) that namespace.

skiesofravens:
Displays speck of dust as of 2024-01-18, and is able to get into 3D gameplay on standalone! Its slow, audio obnoxiously loops and textures are broken (as would be expected from 3d in ruffle).
![Y6nPGIboq0](https://github.com/ruffle-rs/ruffle/assets/138401463/fff8936d-f961-4ffd-81eb-728acd551756)

2024-02-21 extension gets stuck on the title screen corruption, and throws endless Unable to lock Ruffle core errors:
[...]

PR #13954, opened 2023-11-10
avm2: Support enumeration for XML objects

sleepycatcoding:
Previously, iterating the XML object would iterate the prototype. This broke filtering expressions on XML objects, which should work.


Dinnerbone (approved):

Dinnerbone:
Love seeing all these fixes <3

Fancy2209:
This progressed #13362, new error pops up

PR #14015, opened 2023-11-14
desktop: Add window.location.href spoofing to the fake ExternalInterface

evilpie:
From what I can tell this `ExternalInterface.call("window.location.href.toString")` is probably the most common usage of the ExternalInterace API, for example in #12054, #8330 and others that I can't remember. I re-used the already existing `--spoof-url` parameter for this.

### DIFF
--- a/desktop/src/backends/external_interface.rs
+++ b/desktop/src/backends/external_interface.rs
@@ -1,10 +1,30 @@
-use ruffle_core::external::{ExternalInterfaceMethod, ExternalInterfaceProvider};
+use ruffle_core::context::UpdateContext;
+use ruffle_core::external::{
+    ExternalInterfaceMethod, ExternalInterfaceProvider, Value as ExternalValue,
+};
+use url::Url;
 
-#[derive(Default)]
-pub struct DesktopExternalInterfaceProvider;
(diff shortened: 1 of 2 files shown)


Dinnerbone:
> I re-used the already existing `--spoof-url` parameter for this.

FWIF this is wrong, but I'm not sure it matters in practice. The location would be something like `https://armorgames.com/play/13196/achievement-unlocked-3` rather than `https://cache.armorgames.com/files/games/achievement-unlocked-13196.swf?v=1373587527`

evilpie:
I see. I guess we could extract just the hostname? Or maybe hostname + path.  But like you said, it probably doesn't matter in practice, most games just try to extract the hostname anyway.

evilpie:
For `Security.pageDomain` (#14028) it would be useful to have the page URL as well, so I am going to look into adding a new parameter for that.

evilpie:
> For `Security.pageDomain` (#14028) it would be useful to have the page URL as well, so I am going to look into adding a new parameter for that.

We decided not to add a command line argument for page URL spoofing, so I am going to mark this as ready after all.

Dinnerbone (approved):

PR #14037, opened 2023-11-16
avm2: Add ContextMenu clipboard related properties

danielhjacobs:
Slightly progresses #10173.

Next error there is back to being TLF related.

### DIFF
--- a/core/src/avm2/globals/flash/ui/ContextMenu.as
+++ b/core/src/avm2/globals/flash/ui/ContextMenu.as
@@ -1,6 +1,7 @@
 package flash.ui
 {
     import flash.display.NativeMenu;
+    import __ruffle__.stub_getter;
 
     public final class ContextMenu extends NativeMenu
     {
@@ -11,6 +12,7 @@ package flash.ui
         }
 
         private var _customItems:Array;
+        private var _clipboardMenu:Boolean;
         
         public function get customItems():Array {
             return this._customItems;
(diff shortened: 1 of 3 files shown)


Dinnerbone (approved):

core/src/avm2/globals/flash/ui/ContextMenu.as
```diff
@@ -32,9 +35,25 @@ package flash.ui
...
         }
 
-        public static function get isSupported() : Boolean
-        {
+        public function get clipboardItems(): ContextMenuClipboardItems {
+            return this._clipboardItems;
+        }
+
+        public function set clipboardItems(value:ContextMenuClipboardItems):void {
+            this._clipboardItems = value;
```
Lord-McSweeney, line 43: This should have a stub warning?

core/src/avm2/globals/flash/ui/ContextMenu.as
```diff
@@ -32,9 +35,25 @@ package flash.ui
...
+        }
+
+        public function get clipboardMenu():Boolean {
+            return _clipboardMenu;
+        }
+
+        public function set clipboardMenu(value:Boolean):void {
+            _clipboardMenu = value;
```
Lord-McSweeney, line 51: This should have a stub warning

core/src/avm2/globals/flash/ui/ContextMenu.as
danielhjacobs: https://github.com/ruffle-rs/ruffle/pull/14038#discussion_r1395934331

Dinnerbone said that if things set what they should set, they shouldn't be stubs if if they do nothing. Difference in philosophy I guess. Probably a discussion for Discord.

Dinnerbone: Disagree, this code is doing what it's supposed to. The actual functionality is not here.

Lord-McSweeney (approved):

Issue #1074, opened 2020-08-30
Any conversation in the Club Penguin PSA Missions than has an endFunction fails and ends the conversation early


Thestickman391:
This can be seen at the start of Missions 8, 9, 10 and when Herbert hacks into the HQ in Mission 11.

![image](https://user-images.githubusercontent.com/4157860/91653128-f401d480-eaf1-11ea-85ca-5969fc311e14.png)
![image](https://user-images.githubusercontent.com/4157860/91653130-fb28e280-eaf1-11ea-8a3b-b7e42d36e599.png)
![image](https://user-images.githubusercontent.com/4157860/91653133-01b75a00-eaf2-11ea-840b-16ee017bd834.png)

To replicate use `wget --recursive --no-parent https://icer.ink/media1.clubpenguin.com/play/v2/games/quests/` to grab all the PSA missions and then run `ruffle /q8|9|10|11/quest.swf` and then start the opening conversation and see that it ends prematurely.

PR #14038, opened 2023-11-16
avm2: Make MouseEvent isRelatedPropertyInaccessible a getter/setter

danielhjacobs:
Slightly progresses #10173 after #14037. Without this PR but with #14037, the error that appears when clicking anywhere on the animation (which full-screens the content in Ruffle but not in browser Flash) is about writing to the read-only property isRelatedPropertyInaccessible. With this PR, it's instead TLF/FTE related.

### DIFF
--- a/core/src/avm2/globals/flash/events/MouseEvent.as
+++ b/core/src/avm2/globals/flash/events/MouseEvent.as
@@ -33,7 +33,7 @@ package flash.events
         public var shiftKey: Boolean;
         public var buttonDown: Boolean;
         public var delta: int;
-        public var isRelatedObjectInaccessible: Boolean;
+        private var _isRelatedObjectInaccessible: Boolean;
 
         public var movementX: Number;
         public var movementY: Number;
@@ -59,7 +59,6 @@ package flash.events
             this.shiftKey = shiftKey;
             this.buttonDown = buttonDown;
             this.delta = delta;
-            this.isRelatedObjectInaccessible = false; // unimplemented
 
             this.movementX = 0.0; // unimplemented
             this.movementY = 0.0; // unimplemented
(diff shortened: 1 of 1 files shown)


core/src/avm2/globals/flash/events/MouseEvent.as
```diff
@@ -76,6 +76,17 @@ package flash.events
...
         }
 
+        public function get isRelatedObjectInaccessible():Boolean {
+            return _isRelatedObjectInaccessible;
+        }
+
+        public function set isRelatedObjectInaccessible(value:Boolean):void {
+            // Technically this sets exactly what it should.
+            // However, it is a stub since it affects nothing.
+            stub_setter("flash.events.MouseEvent", "isRelatedObjectInaccessible");
```
Dinnerbone, line 86: Doesn't matter if it affects nothing, that's what it's meant to do, it's not a stub

Dinnerbone:
I'm a little confused, the property doesn't look readonly previously. You've just turned it from a read-write property into a getter/setter over the same property - which looks more correct with Flash's documentation, but it's the same thing with just extra work.

What exactly was the error you got before this PR?

danielhjacobs:
> What exactly was the error you got before this PR?

It wasn't specifically to do with this class, but a class that inherited from it:

`ReferenceError: Error #1074: Illegal write to read-only property isRelatedObjectInaccessible on RemappedMouseEvent.
	at flash.events::MouseEvent()
	at RemappedMouseEvent()
	at flashx.textLayout.container::TextContainerManager/mouseDownHandler()`

core/src/avm2/globals/flash/events/MouseEvent.as
danielhjacobs: Ok. I assume the same applies to the properties I added in #14037 too then so I should remove the stub_setters there too, correct?

Dinnerbone: Yep, thank you!

Dinnerbone:
Right okay, so it's a class overriding our MouseEvent, they define a getter without a setter, and then that fails because we try to call "our" setter (but it's really their lack of setter that we call).

This is good, thank you!

Dinnerbone (approved):

Issue #14028, opened 2023-11-15
World's End Chapter 1, 2, 3 won't load to the Main Menu


theExtramundane:
### Describe the bug

![pic-window-231115-1221-47](https://github.com/ruffle-rs/ruffle/assets/81713602/6e2a1797-30eb-4c06-8df4-147c61c9dfab)

I am trying to play this game and it seems to default to a blank screen. I've tried this on both GameJolt and Kongregate and neither seem to work properly. I also tested it with the following two chapters with the same results. I'm not exactly sure what causes this, but it appears to happen consistently wherever I test it, and most notably, it's the same error that appears when I try to launch it in the standalone Adobe Flash Player Projector (v32) as well. Seems like some sort of connectivity error.

### Expected behavior

Usually this would boot to the title screen and offer the option to play the game. It appears to work in BlueMaxima's Flashpoint as evidenced by [this video](https://www.youtube.com/watch?v=quw9dJtcBKI) made a bit over a year ago (it was confirmed it was Flashpoint in the comments), but since then the games have seem to have gone missing in Flashpoint's database, with only the "[story mode](https://flashpointproject.github.io/flashpoint-database/search/#9056b536-da1d-4ab4-a1bb-db3cd3bd197a)" equivalents remaining (which lead to dead links), so unfortunately I can't test it with that.

### Content Location

https://gamejolt.com/games/world-s-end-chapter-1/120686
https://www.kongregate.com/games/MezzanineStairs/worlds-end-chapter-1

I also have the SWF URL from the debug info here: https://download.gamejolt.net/a5ffa110211c0f9070f16b6ee7c3e45de4964bbe78b3b90b7b644a444343db02,1700159332,7/data/games/1/186/120686/files/58821a66ab818/worlds-end-chapter-1-gamejolt-1-20-2017.swf

### Affected platform

Browser's extension

### Operating system

Artix Linux

### Browser

Chromium: 119.0.6045.105 (Official Build) (64-bit)

### Player Info
Allows script access: false
Renderer: wgpu
Adapter Backend: Gl
Adapter Name: "ANGLE (AMD, AMD Radeon Graphics (renoir LLVM 16.0.6), OpenGL 4.6)"
Adapter Device Type: Other
Adapter Driver Name: ""
Adapter Driver Info: ""
[...]

PR #14054, opened 2023-11-17
avm2: Implement flash.system.Security pageDomain

evilpie:
Improves #14028. 

I don't have a good way to test real Flash in a browser, so it would be interesting to know what `Security.pageDomain` returns exactly. The documentation mentions `http://www.example.com` so without a trailing slash.

### DIFF
--- a/core/src/avm2/globals/flash/system/Security.as
+++ b/core/src/avm2/globals/flash/system/Security.as
@@ -1,6 +1,8 @@
 package flash.system {
 	public final class Security {
+		public static native function get pageDomain():String;
 		public static native function get sandboxType():String;
+
 		public static native function allowDomain(... domains):void;
 		public static native function allowInsecureDomain(... domains):void;
 		public static native function loadPolicyFile(url: String):void;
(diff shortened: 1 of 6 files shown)


n0samu:
Just checked in Flash Player and it does seem to include the trailing slash. For example, `http://localhost:8000/`

desktop/src/cli.rs
```diff
@@ -115,6 +115,10 @@ pub struct Opt {
     #[clap(long, value_parser)]
     pub spoof_url: Option<Url>,
 
+    /// Spoofs the URL of the page embedding the SWF.
+    #[clap(long, value_parser)]
+    pub spoof_page_url: Option<Url>,
```
Aaron1011, line 120: I don't think we need a separate option - we can extract the page url from `spoof_url`

evilpie: But maybe we want to spoof a domain for the SWF and the page? This option would be used for https://github.com/ruffle-rs/ruffle/pull/14015 as well, so I think we should add it.

n0samu: Do we know of any games that need different page and SWF URLs? If not, then I do think keeping it simple with one option is best

evilpie: I don't know any game off the top of my head that would need this. So let's remove it for now, adding it back is trivial.

Dinnerbone (approved):

PR #14069, opened 2023-11-19
avm1: Copy `Activation`'s local register set when running a `catch`

### LINKED ISSUE #14049: avm1: Attached files crash after setting FlashVars
This issue has only occured since #11497 was merged, but I don't see anything wrong with this PR.
It happens on both desktop and web.

[Files are available here.](https://github.com/ruffle-rs/ruffle/files/13391516/_files.zip)

```html
<embed src="viewer.swf" flashvars="avatarurl=avatar.swf?ThisWillCauseTheCrash=AnyValue" />
```
> Ruffle loads the file `avatar.swf?ThisWillCauseTheCrash=AnyValue` as expected, sets the FlashVars (after the `?`) as expected and... for some reason enters an infinite loop and crashes after reaching timeout.

---

```html
<embed src="viewer.swf" flashvars="avatarurl=avatar.swf" />
```
> No crash occurs in this situation.

No clue why this is happening. Note that there is an incorrect warning: `Encountered stub: AVM2 flash.display.MovieClip.goto() with SWF 9 movie`, but this is not what causes the issue (though this should be fixed too).

Lord-McSweeney:
Fixes #14049


Dinnerbone (approved):

Toad06:
Thank you very much for all your work on this issue. 😉

Issue #13990, opened 2023-11-12
Error on https://www.htfgames.com/games/after_amok.htm


Ignacioto3:
# Error Info
Error name: Error
Error message: panicked at core/src/avm2/globals/flash/display3D/textures/atf_jpegxr.rs:26:55:
Failed to parse ATF texture: Error { kind: UnexpectedEof, message: "failed to fill whole buffer" }
Error stack:
```
Error: panicked at core/src/avm2/globals/flash/display3D/textures/atf_jpegxr.rs:26:55:
Failed to parse ATF texture: Error { kind: UnexpectedEof, message: "failed to fill whole buffer" }
    at imports.wbg.__wbg_new_ab87fd305ed9004b (chrome-extension://donbcfbmhbcapadipfkeojnmajbakjdc/dist/339.js:2624:21)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::h2b09d9124792ee2e (wasm://wasm/0340036e:wasm-function[9968]:0x9cff40)
    at std::panicking::rust_panic_with_hook::h3aa054d35a0817d7 (wasm://wasm/0340036e:wasm-function[6528]:0x8cfe82)
    at core::panicking::panic_fmt::h3e1dd3d08288569e (wasm://wasm/0340036e:wasm-function[9765]:0x9caca7)
    at core::result::unwrap_failed::h8b3db0f11171b57b (wasm://wasm/0340036e:wasm-function[8207]:0x94dad9)
    at ruffle_core::avm2::globals::flash::display3D::textures::atf_jpegxr::do_compressed_upload::he98a59b979b7e0e9 (wasm://wasm/0340036e:wasm-function[720]:0x225363)
    at ruffle_core::avm2::globals::flash::display3D::textures::texture::upload_compressed_texture_from_byte_array_internal::h0db36283fb246ef1 (wasm://wasm/0340036e:wasm-function[3717]:0x76113a)
    at ruffle_core::avm2::function::Executable::exec::h5e19f618e1451095 (wasm://wasm/0340036e:wasm-function[922]:0x339fc9)
    at ruffle_core::avm2::object::TObject::call_property::ha173cc224400ee6c (wasm://wasm/0340036e:wasm-function[1263]:0x423639)
    at <ruffle_core::avm2::object::Object as ruffle_core::avm2::object::TObject>::call_property::h8fd0eed1dc8aaf8d (wasm://wasm/0340036e:wasm-function[2835]:0x69b12b)
```
[...]

PR #14082, opened 2023-11-20
avm2: Always run loadCompressedTextureFromByteArray synchronously

Aaron1011:
The caller might modify the ByteArray immediately after the call. Since we don't have a simple way of copying it from ActionScript, let's perform the actual upload synchronously, but still fire the event during the next tick.

Fixes https://github.com/ruffle-rs/ruffle/issues/13990


Lord-McSweeney:
Why not copy from AS? Shouldn't something like the following work?
```
var copiedData = new ByteArray();
data.readBytes(copiedData);
```
then passing the `copiedData` to `loadCompressedTextureFromByteArrayInternal`

Aaron1011:
Good idea - I'm closing this in favor of https://github.com/ruffle-rs/ruffle/pull/14140

PR #14089, opened 2023-11-20
core: Fix MorphShape inaccuracy on complex paths

rogual:
SWF stores edges as deltas in its ShapeRecords. When interpolating MorphShapes, Ruffle interpolates these deltas instead of interpolating the absolute positions of the control points. This can cause incorrect rendering of shape tweens.

### Explanation in detail:

- Imagine a complex path made of many small segments.

- The smallest deltas in the path might only be one twip long.

- Imagine we're tweening the shape with a shape tween that stretches it out to twice its size.

- Consider what happens to a single delta as the ratio smoothly increases from 0 to 1.

- The delta is quantized, so it's unable to smoothly stretch from being 1 twip long to being 2 twips long.

- Instead, it'll instantaneously "jump" to being 2 twips long as soon as the ratio exceeds 0.5.

- Add up all these "jumping" deltas and you get a very jittery tween.

### The solution:

MorphShapes should be interpolated using the absolute positions of the control points, not the deltas.

### Before:

https://github.com/ruffle-rs/ruffle/assets/520855/a128588f-2ffd-43f5-9782-6aecfa768aa2

### After:

https://github.com/ruffle-rs/ruffle/assets/520855/7d09ecb1-6212-4bff-99bc-921dacabf28f

I added this example SWF as a test case in `movieclip_hittest_shapeflag`.

### Affected SWF in the wild:

https://foon.uk/farcade/hapland2/ — The surface of the water is glitchy because of this.


Dinnerbone:
Thank you so much! Nice work, and thanks for the test :)

Dinnerbone (approved):

PR #14139, opened 2023-11-24
naga-agal: Zero-extend input vectors for cross-product

Aaron1011:
This matches the behavior of Flash Player

Fixes https://github.com/ruffle-rs/ruffle/issues/14011


torokati44 (approved):
I believe you when it comes to 3D stuff, and the added test is nice!

PR #14140, opened 2023-11-24
avm2: Copy data for async loadCompressedTextureFromByteArray

Aaron1011:
The caller might modify the ByteArray immediately after the call, so we need to copy the ByteArray.

Fixes https://github.com/ruffle-rs/ruffle/issues/13990

### DIFF
--- a/core/src/avm2/globals/flash/display3D/textures/Texture.as
+++ b/core/src/avm2/globals/flash/display3D/textures/Texture.as
@@ -1,24 +1,29 @@
-package flash.display3D.textures {	
+package flash.display3D.textures {
     import flash.display.BitmapData;
     import flash.events.Event;
     import flash.utils.ByteArray;
     import flash.utils.setTimeout;
-    
+
     public final class Texture extends TextureBase {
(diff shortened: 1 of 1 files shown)


core/src/avm2/globals/flash/display3D/textures/Texture.as
```diff
@@ -1,24 +1,28 @@
...
+                data.readBytes(copiedData);
+
                 setTimeout(function() {
-                    self.uploadCompressedTextureFromByteArrayInternal(data, byteArrayOffset);
-                    self.dispatchEvent(new Event("textureReady"));
-                }, 0);
-            } else {
+                        self.uploadCompressedTextureFromByteArrayInternal(copiedData, byteArrayOffset);
+                        self.dispatchEvent(new Event("textureReady"));
+                    }, 0);
```
Lord-McSweeney, line 19: These three lines seem to be double-indented

Aaron1011: This is the formatting used by my AS3 formatter in vscode

Lord-McSweeney (approved):

Bale001 (approved):

PR #14146, opened 2023-11-24
chore: Re-enable Dependabot with dependency groups

Aaron1011:
This should let us switch back from Renovate to Dependabot


torokati44:
Re: https://github.com/ruffle-rs/ruffle/pull/14863

Time to dust this off then?

Issue #4427, opened 2021-05-25
Flash Flash Revolution just won't work. https://www.flashflashrevolution.com/


9caseyu:
IN LEGACY ENGINE:
Every time I try to load a song, either it has no music, or it just doesn't load.

VELOCITY ENGINE: songs just don't load.

https://www.flashflashrevolution.com/

danielhjacobs:
It's ActionScript 3, see #1368.

kmeisthax:
So... confession time. I've been on a rhythm game kick for the better part of this year, and as a way to at least pull myself back into Ruffle work again, I decided to look into ancient Flash rhythm games. 😛

LEGACY ENGINE / `ffrndanotherway.swf` - AVM1 SWF. Has significant performance problems in the song select menu, but otherwise works. I can play through a song and get to score select. At some point in the menu I'd accidentally turned on a modifier that puts the step receptors on the side, which made it rather difficult to read 😅

VELOCITY ENGINE / `ffr_Wide.swf` - Also AVM1? I'm actually kind of confused - @danielhjacobs claims it's AS3, but it's not. For the record I got the SWF file from [here](https://www.flashflashrevolution.com/FFR_the_Game_Simple.php?ver=velo).

Regardless, the menus have even worse performance problems that Legacy Engine does. Definitely worth a profiling. But I'm still able to at least get into a song and play it.

At this point I should also point out that I'm playing on Ruffle Desktop. These movies attempt to connect to an external IP address which is something we don't support on Ruffle Web. If FFR wants to use Ruffle they may have to change their netcode to use a WebSocket shim or something.

R^3 engine / `R^3.swf` - This one actually is AVM2. It appears to hang at the loading screen at 0%. The only relevant log messages are:

```
[2m2023-11-13T02:43:01.377270Z[0m [33m WARN[0m [1mrun_frame[0m[2m:[0m[1mrun_all_phases_avm2[0m[2m:[0m [2mruffle_core::stub[0m[2m:[0m Encountered stub: AVM2 flash.system.Security.allowDomain()
[2m2023-11-13T02:43:01.380322Z[0m [33m WARN[0m [1mrun_frame[0m[2m:[0m[1mrun_all_phases_avm2[0m[2m:[0m [2mruffle_core::library[0m[2m:[0m Unknown device font "Times New Roman"
```

R^3 also has an AIR version that crashes immediately in Ruffle due to unimplemented APIs.

[...]

PR #14159, opened 2023-11-25
core: Add PlayerMode::{FlashPlayer, AIR} for Adobe AIR support

Aaron1011:
We've now had two different bug reports involving Adobe AIR SWFs, so I'm going to go ahead and start adding a framework for AIR support.

This commit just adds a command-line option
`--player-mode <flash-player|air>` (defaulting to `flash-player`), and passes it along to the `Player`. The actual value is currently unused - in a follow-up PR, I'm going to implement namespace versioning for AIR.

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -69,7 +69,7 @@ pub use events::PlayerEvent;
 pub use font::DefaultFont;
 pub use indexmap;
 pub use loader::LoadBehavior;
-pub use player::{Player, PlayerBuilder, StaticCallstack};
+pub use player::{Player, PlayerBuilder, PlayerRuntime, StaticCallstack};
 pub use ruffle_render::backend::ViewportDimensions;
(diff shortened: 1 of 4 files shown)


Dinnerbone:
Sorry to start the bikeshedding so soon, but I think "runtime" or "runtime environment" would be more appropriate a name.

kmeisthax:
Seems like a plan. In fact, I ran into this investigating #4427 - while two of the three versions of FFR works, R^3 engine doesn't, and being able to run the AIR version would aid in debugging and closing out that issue.

evilpie:
I would almost suggest just going with `--air` and no flag means FlashPlayer. That bypasses a bunch of the bikeshedding :laughing:.

Lord-McSweeney (approved):

Dinnerbone (approved):

PR #14160, opened 2023-11-25
avm2: Implement `XMLList.contains`

Lord-McSweeney:


sleepycatcoding:
This does not seem to be entirely correct. Take this following code:
```actionscript
var x = new XMLList("<item>A</item><item>B</item>");
var x1 = new XMLList("<item>A</item>");

trace(x.contains("A")); // False with this PR, true in Flash
trace(x.contains("B")); // False with this PR, true in Flash

trace(x.contains(x));
trace(x.contains(x1));

trace(x.contains(new XML("<item>A</item>")));
```

kmeisthax:
@sleepycatcoding On `xml_list.rs` line 221 it specifically returns false if the `XMLList` holds more than one node - i.e. multiple nodes in an `XMLList` isn't implemented yet.

sleepycatcoding:
> @sleepycatcoding On `xml_list.rs` line 221 it specifically returns false if the `XMLList` holds more than one node - i.e. multiple nodes in an `XMLList` isn't implemented yet.

Yeah, I know, this was just some code I wrote for testing and I copied all of it into the comment :sweat_smile:, I more meant that `x.contains("A")` did not work as expected.

Lord-McSweeney:
> On xml_list.rs line 221 it specifically returns false if the XMLList holds more than one node - i.e. multiple nodes in an XMLList isn't implemented yet.

@kmeisthax This seemed to match FP from my testings, `.contains([some multi-node XMLList])` will always return false.

Lord-McSweeney:
Superseded by #16525.

PR #14163, opened 2023-11-25
dekstop: Support `quit` and `fullscreen` fscommands

### LINKED ISSUE #9141: Kamen Rider Wizard v3,1: Can't exit or full screen
### Describe the bug

[_flash__kamen_rider_wizard_v3_1_by_crimes0n_d5zcwv3.zip](https://github.com/ruffle-rs/ruffle/files/10409035/_flash__kamen_rider_wizard_v3_1_by_crimes0n_d5zcwv3.zip)
Running this SWF through my flash player projector on desktop, the button on the top left is supposed to full screen the SWF (Infact the SWF starts in full screen when I launch it on the projector.), while the button on the top right exits the flash projector entirely. These 2 buttons on ruffle though seem to do absolutely nothing.

Kamen

### Expected behavior

The top left button should enable ruffle full screen when pressed and the button on the right should close ruffle when pressed.

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

_No response_

### Additional information

_No response_

Lord-McSweeney:
This closes #9141

### DIFF
--- a/desktop/src/backends/fscommand.rs
+++ b/desktop/src/backends/fscommand.rs
@@ -1,0 +1,33 @@
+use crate::custom_event::RuffleEvent;
+
+use ruffle_core::external::FsCommandProvider;
+use std::rc::Rc;
+use winit::event_loop::EventLoopProxy;
+use winit::window::{Fullscreen, Window};
+
+pub struct DesktopFSCommandProvider {
+    pub event_loop: EventLoopProxy<RuffleEvent>,
+    pub window: Rc<Window>,
+}
+
+impl FsCommandProvider for DesktopFSCommandProvider {
+    fn on_fs_command(&self, command: &str, args: &str) -> bool {
+        match command {
+            "quit" => {
+                let _ = self.event_loop.send_event(RuffleEvent::ExitRequested);
+            }
+            "fullscreen" => {
+                match args {
+                    "true" => self
+                        .window
+                        .set_fullscreen(Some(Fullscreen::Borderless(None))),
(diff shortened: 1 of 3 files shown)


desktop/src/backends/fscommand.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub event_loop: EventLoopProxy<RuffleEvent>,
+    pub window: Rc<Window>,
+    pub is_fullscreen: Cell<bool>,
+}
+
+impl FsCommandProvider for DesktopFSCommandProvider {
+    fn on_fs_command(&self, command: &str, _args: &str) -> bool {
+        let _ = match command {
+            "quit" => {
+                let _ = self.event_loop.send_event(RuffleEvent::ExitRequested);
```
Dinnerbone, line 19: Hmm I'm a bit unsure about this. There's cases where I can see that closing the window might be preferred - I guess if you opened it specifically for one swf. But every other time I think I'd expect that the movie just stops playing, and I can reopen something else. It's weird for a movie to have control over the host application like that.

Lord-McSweeney: I did this to match FP's behavior (it closes the window)- this is commonly used in conjunction with `fullscreen`, to make it easier for the user to exit out of a fullscreen window (probably more useful in Ruffle, since you can't press `Esc` to get out of fullscreen, you have to use the less-known `Alt`+`Enter`)

Dinnerbone (approved):

PR #14164, opened 2023-11-25
naga-agal: Fix handling of sampler overrides

Aaron1011:
After some testing, and looking at OpenFL, I believe I've determined the correct behavior for AGAL sampling:

Each time a Context3D.setProgram or Context3D.setSamplerStateAt call is made, the sampler config for the used texture slot(s) is updated with the new wrapping/filter behavior. For setProgram, this comes from all of the 'tex' opcodes used within the program.

However, when the 'ignoresampler' flag is set in a 'tex' opcode, the setProgram call does *not* override the existing sampler config. As a result, that program will sample with the behavior determined by the most recent setSamplerStateAt or setProgram call involving the used texture slot(s).

Previously, we were always overriding the opcode sampler config with the values from Context3D.setSamplerStateAt. However, I didn't realize that the order of the calls matter, so none of my tests ended up observing the effect of 'ignoresampler'.

We now need to process AGAL bytecode twice - a quick initial parse to determine the sampler configs (which need to be updated when we call 'setProgram'), and a second time when to build the Naga module (which needs to wait until we have the vertex attributes available, which can be changed by ActionScript after setting the program).


render/naga-agal/src/builder.rs
```diff
@@ -273,12 +283,14 @@ pub enum AgalVersion {
...
+    operations: Vec<(Opcode, DestField, SourceField, Source2)>,
+}
+
 impl<'a> NagaBuilder<'a> {
-    pub fn process_agal(
-        mut agal: &[u8],
-        vertex_attributes: &[Option<VertexAttributeFormat>; MAX_VERTEX_ATTRIBUTES],
-        sampler_overrides: &[Option<SamplerOverride>; 8],
-    ) -> Result<Module> {
+    fn parse_byecode(mut agal: &[u8]) -> Result<ParsedBytecode> {
```
Lord-McSweeney, line 293: typo: `parse_byecode` --> `parse_bytecode`

tests/tests/swfs/stage3d/shared/com/adobe/utils/AGALMiniAssembler.as
```diff
@@ -372,7 +372,7 @@ package com.adobe.utils
 								if ( optfound == null )
 								{
 									// todo check that it's a number...
-									//trace( "Warning, unknown sampler option: "+opts[k] );
+									trace( "Warning, unknown sampler option: "+opts[k] );
```
Lord-McSweeney, line 375: Intentional change or leftover from debugging?

Aaron1011: Intentional - this makes it easier to write agal shaders

Dinnerbone (approved):

PR #14166, opened 2023-11-26
avm2: Implement AIR playerglobal versioning

Aaron1011:
This builds on our existing playerglobal versioning support to add in AIR versioning. We closely follow the avmplus implementation:

* When an SWF is loaded, we chose either a FlashPlayer or AIR APIVersion for its SWF version, based on our configured player runtime.
* When loading playerglobals, we look at the player runtime. In AIR mode, we map FlashPlayer-versioned definitions to the closest AIR version. This ensures that all runtime APIVersions are in the same series (either AIR or FlashPlayer). In FlashPlayer mode, all AIR-versioned definitions get mapped to VM_INTERNAL, hiding them from user code.

Part of our existing api versioning code was implemented incorrectly. Within playerglobals, we need to treat all unmarked namespaces as VM_INTERNAL - this allows things like playerglobal script initializer "initproperty" opcodes to see any VM_INTERNAL AIR definitions (when we run under FlashPlayer mode). Previously, we were using AllVersions, which would result in those VM_INTERNAL definitions being hidden from other playerglobal code, which is not correct.

Using this support, I've added a stub for the AIR-only 'flash.net.DatagramSocket'. I've also extended the test framework with a new 'player_options.runtime' config option, which can be set to "AIR" or "FlashPlayer" to configure the test runtime mode.

I've also added two new tests:
* 'air_hidden_lookup' runs under the FlashPlayer runtime, and verifies that a list of classes (currently just "DatagramSocket" are inacessible).
* 'air_datagram_socket', which uses `player_options.runtime = "AIR"` to construct an instance of `flash.net.DatagramSocket`. We can extend this test once we implement more of `DatagramSocket`

With this commit, we have all of the needed infrastructure to start implementing and testing AIR-only classes and methods.


core/src/avm2/globals.rs
```diff
@@ -698,7 +698,7 @@ fn load_playerglobal<'gc>(
                 // Lookup with the highest version, so we we see all defined classes here
                 let ns = Namespace::package($package, ApiVersion::VM_INTERNAL, &mut activation.borrow_gc());
                 let name = QName::new(ns, $class_name);
-                let class_object = activation.domain().get_defined_value(activation, name)?;
+                let class_object = activation.domain().get_defined_value(activation, name).unwrap_or_else(|e| panic!("Failed to lookup {name:?}: {e:?}"));
```
Lord-McSweeney, line 701: Is this change leftover from debugging or generally useful?

Aaron1011: This gives a much more useful error - our normal error-printing logic doesn't really work during playerglobal loading, since the `Error` class might not have been initialized.

core/src/avm2/api_version.rs
```diff
@@ -59,37 +62,142 @@ pub enum ApiVersion {
...
+            (30, PlayerRuntime::FlashPlayer) => Some(ApiVersion::SWF_30),
+            (30, PlayerRuntime::AIR) => Some(ApiVersion::AIR_19_0),
+            (31, PlayerRuntime::FlashPlayer) => Some(ApiVersion::SWF_31),
+            (31, PlayerRuntime::AIR) => Some(ApiVersion::AIR_20_0),
+
             // We haven't yet created entries from higher versions - just map them
             // to the highest non-VM_INTERNAL version.
-            _ if val > 31 => Some(ApiVersion::SWF_31),
+            (32.., PlayerRuntime::FlashPlayer) => Some(ApiVersion::SWF_31),
+            (32.., PlayerRuntime::AIR) if val > 31 => Some(ApiVersion::AIR_20_0),
```
evilpie, line 200: `val > 31` is always true

core/src/avm2/api_version.rs
```diff
@@ -59,37 +62,142 @@ pub enum ApiVersion {
     VM_INTERNAL = 52,
 }
 
+static TRANSFER_TABLE: OnceLock<EnumMap<ApiVersion, (ApiVersion, ApiVersion)>> = OnceLock::new();
+
 impl ApiVersion {
-    pub fn from_swf_version(val: u8) -> Option<ApiVersion> {
-        match val {
+    pub fn to_valid_playerglobals_version(self, runtime: PlayerRuntime) -> ApiVersion {
+        // This maps an ApiVersion from our playerglobals SWF to the closest valid valid,
```
Lord-McSweeney, line 69: Typo- "closest valid valid"

Dinnerbone (approved):

Lord-McSweeney (approved):

PR #14178, opened 2023-11-26
avm2: Implement AIR flash.filesystem.File [desktop-only]

evilpie:
Preview. This only works on Desktop, because it uses the native Path APIs. Needs code for version gating and stubs for wasm.


core/src/avm2/globals/flash/filesystem/File.as
```diff
@@ -0,0 +1,46 @@
+package flash.filesystem
+{
+    import flash.net.FileReference;
+
+    public class File extends FileReference
+    {
+        private native function init(path: String = null);
```
Lord-McSweeney, line 7: `path: String = null`- a default parameter is unnecessary since the constructor will always call it with an argument

Dinnerbone:
I know this is just a draft for now, but we need to avoid using std::fs and std::path in core. This all needs to be a backend, or it won't be portable or testable.

evilpie:
@Mesteery would you be interested in making a pull request for your work? We now have FileReference and some foundational AIR support already. The missing `flash.filesystem.File` support currently prevents even the most basic AIR applications from working.

Mesteery:
Why not. I'll look into making a PR. There are still some problems with paths, I should rewrite it.

evilpie:
Nice. I look forward to your PR. It might make sense to make the actual path completely opaque and only have some kind of file handle in the avm2 code, then all methods like "resolve path", "get storage path" etc. would just forward that to the backend.

Fancy2209:
Any update on this?

JL2210:
> Personally I already implemented it on a fork, it's not perfect nor clean. And the branch not up to date with upstream for now but you can find it here if you want:
> 
> https://github.com/Mesteery/ruffle/tree/partial-air/core/src/avm2/globals/flash/filesystem

@Mesteery I rebased your fork against master on my fork (but I almost definitely did it wrong, I'm a rust/flash noob). https://github.com/JL2210/ruffle/tree/partial-air

Issue #13655, opened 2023-10-21
TU 46 - audio issues


Kz-92:
### Describe the bug

some sound effects in tu46 (the sequel to tu 95) are missing, tire screeches, engine noise, takeoff sounds
flash

https://github.com/ruffle-rs/ruffle/assets/103858019/3e811ac5-1be9-4cd6-8f8c-745a1f28a994

ruffle

https://github.com/ruffle-rs/ruffle/assets/103858019/8fcf4984-c8ad-4fa2-bac6-4d884740dc3a

### Expected behavior

all sounds should play normally

### Affected platform

Desktop app

### Operating system

windows 10

### Additional information

[590563_TU_-_46-working.zip](https://github.com/ruffle-rs/ruffle/files/13061517/590563_TU_-_46-working.zip)

n0samu:
Issue content: https://archive.is/zJu2W | [Megalodon](https://megalodon.jp/2026-0704-0159-28/https://yandexwebcache.net:443/yandbtm?fmode=inject&tm=1783097945&tld=com&lang=en&la=1697916800&text=url%3Ahttps%3A//github.com/ruffle-rs/ruffle/issues/13655&url=https%3A//github.com/ruffle-rs/ruffle/issues/13655&l10n=en&mime=html&sign=3ad45bafdc4a227937b1bd2f1e7ded4b&keyno=0)

Attached media:
https://user-images.githubusercontent.com/103858019/277127971-3e811ac5-1be9-4cd6-8f8c-745a1f28a994.mp4
https://user-images.githubusercontent.com/103858019/277127978-8fcf4984-c8ad-4fa2-bac6-4d884740dc3a.mp4

Issue #7879, opened 2022-09-05
https://z0r.de/7024 takes hella long to load and has no sound


Diatonator:
### Describe the bug

https://z0r.de/7024
This crap hangs the browser page and loads something for 15 seconds on my end, then animation plays but without sound.
![image](https://user-images.githubusercontent.com/75184268/188504882-db2e7e14-f5be-4c50-8fcd-6c5179d87594.png)
Nightly 2022-05-09
Also looking into the task manager, Chrome was eating memory for something during that 15 second loading time.

### Expected behavior

Animation with sound, instantaneous play (on modern machines of course).

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Chrome 105.0.5195.102

Diatonator:
I guess this issue https://github.com/ruffle-rs/ruffle/issues/6128 is worth mentioning, because this SWF was mentioned here with other 2, and it had some other issue I suppose.
Another 2 mentioned:
https://z0r.de/7312
https://z0r.de/5885

7312 - loads instantly, but no sound.
5885 - absolutely same stuff as for 7024 - long loading and no sound afterwards.

Herschel:
#7024 is using SampleDataEvent to play dynamic sound, playing a MOD file.

PR #14203, opened 2023-11-28
avm2,audio: Implement `SampleDataEvent`

### LINKED ISSUE #14156: Sion Soundobject Quartet - no sound!
### Describe the bug

This is an extremely fun little music synthesizer app. All of the buttons seem to click and the dials move correctly. However, there is no sound produced at all!

[9Xx7.swf.zip](https://github.com/ruffle-rs/ruffle/files/13464316/9Xx7.swf.zip)

### Expected behavior

When the app starts up there should be sound immediately playing that you can control with the buttons, knobs and dials that are included. 

### Content Location

see attached. 

### Affected platform

Desktop app

### Operating system

Mac OS X 14.1.1

### Browser

_No response_

### Additional information

I also tried on android with the same issue. I'm really excited to be able to play/have fun with this little toy again once it gets working. Happy to help any way I can!

Thank you!

torokati44:
Fixes: #14156
Also fixes z0r.de loops: 5885, 7024 (issue #7879), 7312
Probably fixes (to be tested): #9171, #13655

It's quite rough (full of `unwrap()`s for example, and overall unfinished) but it "just started working".

Mostly pushing to show that it's in progress - any ideas on how to improve the architecture are welcome.

An unexpected (to me at least) problem is that sound synthesis in AS3 needs to do quite a lot of math really fast, so on my machine, even in release mode, the content linked above could barely keep up with real time sound generation, or not at all. Profiling and optimization is yet to be done (probably related: #7920).


SN902:
Most NSFW video games depend on this commit, here still no any SFX, but now they panics after the end of the first loop (need to wait for the issue reproduction)

[Twi_Armor.zip](https://github.com/ruffle-rs/ruffle/files/13510058/Twi_Armor.zip)

`thread 'main' panicked at 'not too many sounds: TooManySounds', core\src\avm2\globals\flash\media\sound.rs:179:14
AVM2 stack trace:
        at flash.media::Sound/flash::media::Sound::play()
        at mp3pl::MP3Player/play()
        at appleblitz_fla::MainTimeline/playsounds()
        at appleblitz_fla::MainTimeline/loopend()
        at appleblitz_fla::MainTimeline/frame1()
error: process didn't exit successfully: `target\release\ruffle_desktop.exe` (exit code: 0xc0000409, STATUS_STACK_BUFFER_OVERRUN)`

core/src/backend/audio/mixer.rs
```diff
@@ -881,6 +899,68 @@ impl Stream for EventSoundStream {
     }
 }
 
+use std::collections::VecDeque;
+
+/// A stream for event sound instances with custom envelopes, start/end point, or loop settings.
+struct GeneratedSoundStream {
+    position: u32,
+    /// Contents of the shared `next_samples` are periodically moved into
+    /// to to reduce lock contention while yielding audio frames.
```
Lord-McSweeney, line 908: typo: "to to reduce"

torokati44: Thank you, fixed!

torokati44:
TBD:
 - verify that the architecture is correct - the `load_called` method, how `AudioManager` and `AudioBackend` and `Mixer` keep track of what, the new stream types
 - test all the games/animations mentioned above, and actually make them work properly
 - some kind of regression test
 - see if any more events need to be dispatched

core/src/backend/audio.rs
```diff
@@ -425,22 +441,72 @@ impl<'gc> AudioManager<'gc> {
...
+                let data = sample_data_evt
+                    .get_public_property("data", activation)
+                    .unwrap()
+                    .coerce_to_object(activation)
+                    .unwrap();
+                let ba = data.as_bytearray().unwrap();
+
+                ba.set_position(0);
+                for _ in 0..ba.len() / 4 {
+                    ns.push_back(ba.read_float().expect("float can be read"));
```
Lord-McSweeney, line 486: Shouldn't failing to read a float result in the read loop terminating early?

torokati44: that is to be tested :|

torokati44:
@copilot try to create a new PR inspired by this, implemented on top of current `master`. like a git rebase, but it will be complicated...

core/src/backend/audio.rs
torokati44: AFAIK `read_float` can only fail if there are not enough bytes (any four bytes can be reinterpreted as an f32), and the `ba.len() / 4` in the loop ensures this never happens.

core/src/avm2/globals/flash/media/sound.rs
```diff
@@ -174,6 +174,33 @@ pub fn play<'gc>(
...
 
+        // If no load has been initiated yet, this is a generated (synthesized) sound.
+        // Register it with the audio backend so it can receive SampleDataEvent callbacks.
+        if sound_object.is_empty() {
+            let sound_channel = SoundChannelObject::empty(activation);
+            let handle = activation
+                .context
+                .audio_manager
+                .start_generated_sound(activation.context.audio, sound_object)
+                .expect("not too many sounds");
```
Lord-McSweeney: Can't this `expect` be hit if too many sounds are generated?

core/src/avm2/object/sound_object.rs
```diff
@@ -161,6 +169,38 @@ impl<'gc> SoundObject<'gc> {
...
+        let mut sound_data =
+            unlock!(Gc::write(context.gc(), self.0), SoundObjectData, sound_data).borrow_mut();
+        match &*sound_data {
+            SoundData::Empty => {
+                *sound_data = SoundData::Loading {
+                    queued_plays: Vec::new(),
+                };
+            }
+            SoundData::Loading { .. } => {
+                panic!("Tried to load sound that is already Loading");
```
Lord-McSweeney, line 201: Is there really no way for these panics to be hit? If not, could you add explanatory comments?

core/src/avm2/globals/flash/media/sound.rs
torokati44: Yep, and we should return `null` to AS instead of panicking. Good call!

core/src/backend/audio.rs
torokati44: added a comment about this nevertheless (just not pushed yet)

core/src/avm2/object/sound_object.rs
torokati44: Ah, there are a couple of these cases that should be AS exceptions instead, will look into it!

core/src/avm2/object/sound_object.rs
```diff
@@ -120,14 +122,15 @@ pub enum SoundLoadingState {
     New,
     Loading,
     Loaded,
+    Generated,
 }
```
torokati44, line 126: Are `SoundData` and `SoundLoadingState` supposed to mirror each other like this? @kjarosh

I'm asking you because you added the latter in https://github.com/ruffle-rs/ruffle/pull/19586 .

~~(And now that I'm reading it, maybe `SoundData::NotLoaded` and `SoundData::Loading` should be collapsed into one - it's probably split as the result of another glitchy rebase from eons ago...)~~ EDIT: lol I read the diff wrong when writing this, there's no split like this

kjarosh: Well, it depends on whether sound behaves the same way in AVM1 and AVM2. If it does and we have a lot of evidence of that, perhaps it's a good idea to unify the structures between AVM1 and AVM2 to some extent?

torokati44: It's strange though because SoundData is used from both avm1 and avm2, while SoundLoadingState is only used for/by avm2.

torokati44:
I've incorporated some improvements into this from https://github.com/ruffle-rs/ruffle/pull/24762. (@cmdocmd)

core/src/avm2/object/sound_object.rs
Lord-McSweeney: Has this point been resolved?

core/src/avm2/object/sound_object.rs
Lord-McSweeney: Is this point still relevant?

tests/tests/swfs/avm2/generated_sound_play_twice/Test.as
```diff
@@ -0,0 +1,34 @@
...
+        var ch2:SoundChannel = s.play();
+        var ch1:SoundChannel = s.play();
+        trace(ch1 == ch2);
+        try {
+            // The problem is not that missing.mp3 doesn't exist, but that this
+            // sound is now procedurally generated, so nothing can be loaded into it.
+            s.load(new URLRequest("missing.mp3"));
+            trace("Load succeeded unexpectedly");
+        } catch (error:Error) {
+            trace("An error occurred: " + error.message);
```
Lord-McSweeney, line 23: Can this trace `error.getStackTrace()` instead of `error.message`?

Lord-McSweeney (commented):
What happens if `event.data` is set to null at the end of a SampleDataEvent event handler?

PR #14233, opened 2023-11-29
web: Mark the Firefox extension as compatible with Android upon AMO submission

torokati44:
See for reference:
https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#version-create
https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#version-compatibility-examples

The minimum version is from: https://github.com/ruffle-rs/ruffle/wiki#web

### DIFF
--- a/web/packages/extension/tools/sign_xpi.js
+++ b/web/packages/extension/tools/sign_xpi.js
@@ -59,6 +59,15 @@ async function sign(
             extensionId,
         )}/versions/${encodeURIComponent(version)}/`,
         json: {
+            compatibility: {
+                android: {
+                    min: "84",
+                    // "max" is undefined, so the manifest max or default will be used.
+                },
+                firefox: {
+                    min: "84",
+                },
+            },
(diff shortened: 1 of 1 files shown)


Dinnerbone (approved):

Dinnerbone:
Tried to set these values in the UI and got this error:
Invalid version range. For Firefox for Android, you may only pick a range that starts with version 121.0a1 or higher, or ends with lower than version 79.0a1.

torokati44:
Oh, dang, ouch, pardon.
Thanks for checking! And well in time. 😰
Well then let's set the android min version to "121.0a1"?

Dinnerbone:
Either "121.0a1" or "121" should work, from what I can tell!

torokati44:
Well 121 is a lot cleaner IMHO.

torokati44:
>Tried to set these values in the UI and got this error:
> Invalid version range. For Firefox for Android, you may only pick a range that starts with version 121.0a1 or higher, or ends with lower than version 79.0a1.

On the other hand, uBlock Origin has this set:

> Latest version
> Version 1.54.0
> Released Nov 22, 2023 - 3.46 MB
> Works with firefox 78.0 and later, *android 113.0 and later*

Issue #7660, opened 2022-08-12
Neopets Flash Game Issue - Advert Attack 


crystalglassxxx:
### Describe the bug

During the game Advert Attack, clicking on the "Go" button will eventually stop responding to clicks. 

### Expected behavior

Clicking "Go" is supposed to make the button move to a different spot and make they player move forward. 

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Google Chrome Version 104.0.5112.81

Toad06:
I made a minimal file that reproduces the issue (and also made one for AS3 as it is appears to be affected as well). I suppose the issue is in the "rounding" when setting the x and y properties on a movieclip. When running the test in Flash, x and y are always traced as integers, in Ruffle they can sometimes be floats (when this happens in the AS2 test, the fractional part is always 0.95). The traced values are likely not precise because of the string coercion, but hopefully this could give a hint on what is happening here.

[tests.zip](https://github.com/ruffle-rs/ruffle/files/11646515/tests.zip)

PR #14243, opened 2023-11-30
core: Fix mouse coords reported to ActionScript

rogual:
*Edit*: This patch now takes HiDPI into account and should work for any viewport scale factor.

The specific way Flash Player calculates local mouse coords (`_xmouse`, `_ymouse`, `mouseX`, `mouseY`) gives slightly different results from Ruffle. This patch updates Ruffle so it calculates the mouse position in the same way.

The important thing is to start with the global mouse position in (physical) pixels, and then transform it into local twip space with a matrix.

In order to make this change, I replaced `mouse_to_local` with a new `local_mouse_position` function, because it needs to get the stage scaling factor from the context, so it might as well read the mouse position from there too.


Dinnerbone:
This *might* fix or progress #7660, someone should test that

Dinnerbone (approved):
Thank you very much!

PR #14271, opened 2023-12-01
web: Fix the Firefox extension builder Dockerfile

torokati44:
After some experimentation in the wrong direction (thanks for the ideas anyway @Dinnerbone!), the solution was right here: https://github.com/torokati44/ruffle/actions/runs/7064760895/job/19233434777

That `${{ always() && github.repository == 'ruffle-rs/ruffle' }}` looks weird, but I _hope_ it does what I mean.

### DIFF
--- a/.github/workflows/test_extension_dockerfile.yml
+++ b/.github/workflows/test_extension_dockerfile.yml
@@ -13,8 +13,6 @@ jobs:
     name: Test the Dockerfile
     runs-on: ubuntu-22.04
 
-    if: github.repository == 'ruffle-rs/ruffle'
-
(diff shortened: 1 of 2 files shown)


torokati44:
Welp, I have a soft approval by @danielhjacobs on Discord, and it's getting close to the nightly release and AMO submission triggers, so I'm going ahead and merging. It can't be that bad... /famous last words/

torokati44:
For reference, the `--omit=optional` argument was put there in #12555, and it was fine until the rewrite of the demo page in React/Vite (#14115), but now it was causing problems.

PR #14278, opened 2023-12-02
web/extension: Raise minimum Firefox version on Android to 121

torokati44:
A follow-up for https://github.com/ruffle-rs/ruffle/pull/14233.

Reason: https://github.com/ruffle-rs/ruffle/actions/runs/7067394881/job/19240696121#step:7:200

### DIFF
--- a/web/packages/extension/tools/sign_xpi.js
+++ b/web/packages/extension/tools/sign_xpi.js
@@ -60,12 +60,11 @@ async function sign(
         )}/versions/${encodeURIComponent(version)}/`,
         json: {
             compatibility: {
-                android: {
-                    min: "84",
-                    // "max" is undefined, so the manifest max or default will be used.
-                },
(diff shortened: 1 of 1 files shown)


danielhjacobs (approved):
This Android version isn't stable until December 19th anyway, so while I would suggest we merge this PR before Friday I don't think we need to manually do an AMO submission after merge. We specifically avoid doing more than one AMO submission a week to not overwhelm their review process. I suggest we wait for CI to do it on Friday, so we don't submit two builds within 4 days of each other.

torokati44:
Thanks for the approval.

>  I suggest we wait for CI to do it on Friday, so we don't submit two builds within 4 days of each other.

I agree. Let's hope this week it will be completely successful. If not - we can decide then. But skipping submission on just one week is not the end of the world.

PR #14292, opened 2023-12-03
web: Simplify the `TS_NODE_COMPILER_OPTIONS` value for mocha in core test

torokati44:
While `npm test` in `web` works on CI, for some reason, on my local machine, I get this error:
`✖ ERROR: SyntaxError: Unexpected token 'v', "verbatimModuleSyntax" is not valid JSON`

Perhaps the `,` in that string confuses something. With this patch, I can test the `core` package successfully.

The removed option looks like a backward compatibility thing, that we're only trying to disable, but this _(leans on Chesterton's fence)_ doesn't seem necessary (anymore), and maybe it's not even enabled by default.

### DIFF
--- a/web/packages/core/package.json
+++ b/web/packages/core/package.json
@@ -14,7 +14,7 @@
         "build": "tsc --build --force",
         "postbuild": "node tools/set_version.js && node tools/bundle_texts.js",
         "docs": "typedoc",
-        "test": "cross-env TS_NODE_COMPILER_OPTIONS={\\\"module\\\":\\\"commonjs\\\",\\\"verbatimModuleSyntax\\\":false} mocha"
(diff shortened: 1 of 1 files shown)


torokati44:
Since that flag was added by @relrelb in https://github.com/ruffle-rs/ruffle/commit/479926cdba148193d5412ff875ae7d398768963a, perhaps you can tell if it's fine to remove now...?

Dinnerbone (approved):
LGTM. The default value is false as far as I can tell, too. It may have been left in from him testing to enable it, but decided not to?

torokati44:
> The default value is false as far as I can tell, too.

Yeah, I haven't found anything in the docs on the contrary.

> It may have been left in from him testing to enable it, but decided not to?

Yeah, that's plausible...

Thanks for the approval!

PR #14302, opened 2023-12-03
render: Support PixelBender ByteArray/Vector.<Number> input/output

Aaron1011:
When ActionScript uses a ByteArray/Vector.<Number> as a shader input or target, we create a temporary Rgba32Float texture, and copy the input float32 bytes to/from the texture.

Unfortunately, wgpu doesn't seem to support an Rgb32Float (3-channel) texture. When the shader uses 3 channels, we use a Rgba32Float (4-channel) texture, and manually insert/remove padding for the alpha channels. This isn't very efficient, but it's the simplest solution.

The temporary textures themselves aren't cached anywhere - if this becomes a performance issue, we could look into using some of our existing wgpu texture/buffer pooling code.


core/src/avm2/globals/flash/display/shader_job.rs
```diff
@@ -100,28 +104,75 @@ pub fn get_shader_args<'gc>(
...
-                        ))
+                        let input_texture = if let Some(bitmap) = input.as_bitmap_data() {
+                            ImageInputTexture::Bitmap(bitmap.bitmap_handle(
+                                activation.context.gc_context,
+                                activation.context.renderer,
+                            ))
+                        } else if let Some(byte_array) = input.as_bytearray() {
+                            let expected_len = (width * height * input_channels) as usize
+                                * std::mem::size_of::<f32>();
+                            assert_eq!(byte_array.len(), expected_len);
```
Lord-McSweeney, line 141: What happens here for FP when these asserts fail?

core/src/avm2/globals/flash/display/shader_job.rs
```diff
@@ -100,28 +104,75 @@ pub fn get_shader_args<'gc>(
...
+                                    .iter()
+                                    .flat_map(|val| {
+                                        (val.as_number(activation.context.gc_context).unwrap()
+                                            as f32)
+                                            .to_le_bytes()
+                                    })
+                                    .collect(),
+                            }
+                        } else {
+                            panic!("Unexpected input object {input:?}");
```
Lord-McSweeney, line 166: What's FP's error message here?

Aaron1011:
@Lord-McSweeney I didn't plan on investigating all of the Flash error messages for this PR - I was going to do it in a follow-up, or as we encounter them in the wild.

torokati44 (approved):
While I'm not at all familiar with this area of the codebase, I skimmed through the changes, and nothing stood out as obviously wrong or bad to my eyes.

Thank you for bringing down the outlier numbers as much as possible - this is a perfectly acceptable, given the nature and look of the test (which is: wow :open_mouth: ).

Please squash on merge though!

torokati44:
Fixed CI, rebased on `master`, and squashed the commits. I hope you don't mind.

torokati44:
I'm getting this error now (so does CI):

```
panicked at /home/attila/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.19.1/src/backend/wgpu_core.rs:3009:5:
wgpu error: Validation Error

Caused by:
    In Device::create_bind_group
    Texture binding 6 expects sample type = Float { filterable: true }, but given a view with format = Rgba32Float
```

Issue #14338, opened 2023-12-07
Error on https://www.disney--games.com/gameframe/1


paridhi16065:
# Error Info
Error name: Error
Error message: panicked at /home/runner/.cargo/git/checkouts/nihav-vp6-660fee9efdd4622a/83c7e10/nihav-duck/src/codecs/vp56.rs:1030:18:
index out of bounds: the len is 2 but the index is 2
Error stack:
```
Error: panicked at /home/runner/.cargo/git/checkouts/nihav-vp6-660fee9efdd4622a/83c7e10/nihav-duck/src/codecs/vp56.rs:1030:18:
index out of bounds: the len is 2 but the index is 2
    at H.n.wbg.__wbg_new_a64e3f2afc2cf2f8 (https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.62a8f0c82d1a76601817.js:1:56255)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::h2c4885bc68fd7d6d (wasm://wasm/03399e2a:wasm-function[9815]:0x9bac4c)
    at std::panicking::rust_panic_with_hook::hc20eadded6bfe687 (wasm://wasm/03399e2a:wasm-function[6301]:0x8acafe)
    at core::panicking::panic_fmt::h7a368385936888dc (wasm://wasm/03399e2a:wasm-function[9620]:0x9b5b68)
    at core::panicking::panic_bounds_check::ha1b736c04b755050 (wasm://wasm/03399e2a:wasm-function[7983]:0x933bd2)
    at nihav_duck::codecs::vp56::VP56Decoder::predict_dc::he097ff443be1aee4 (wasm://wasm/03399e2a:wasm-function[2222]:0x5e58de)
    at nihav_duck::codecs::vp56::VP56Decoder::decode_planes::h7dcc73d6830f38c7 (wasm://wasm/03399e2a:wasm-function[577]:0x839fc)
    at <ruffle_video_software::decoder::vp6::Vp6Decoder as ruffle_video_software::decoder::VideoDecoder>::decode_frame::he4824030118ae46d (wasm://wasm/03399e2a:wasm-function[776]:0x271c38)
    at <ruffle_video_software::backend::SoftwareVideoBackend as ruffle_video::backend::VideoBackend>::decode_video_stream_frame::hdb3a1b808c68f1f2 (wasm://wasm/03399e2a:wasm-function[2304]:0x5ffc4a)
[...]

Issue #14202, opened 2023-11-28
Error on https://www.slaveryinnewyork.org/gallery_2_responses.htm


n0samu:
This problem seems to happen unpredictably/randomly, not sure what's going on. Try clicking between the galleries several times and you should be able to trigger it.
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/127249a1-78d0-4a21-905d-49dad65331ae)

# Error Info
Error name: Error
Error message: panicked at /home/runner/.cargo/git/checkouts/nihav-vp6-660fee9efdd4622a/83c7e10/nihav-duck/src/codecs/vp56.rs:1030:18:
index out of bounds: the len is 2 but the index is 2

# Player Info
Allows script access: true
Renderer: wgpu
Adapter Backend: Gl
Adapter Name: "ANGLE (AMD, Radeon R9 200 Series Direct3D11 vs_5_0 ps_5_0)"
Adapter Device Type: Other
Adapter Driver Name: ""
Adapter Driver Info: ""
Enabled features: Features(TEXTURE_COMPRESSION_BC | TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES | PUSH_CONSTANTS | SHADER_UNUSED_VERTEX_OUTPUT)
Available features: Features(CLEAR_TEXTURE | MULTIVIEW)
[...]

Issue #13624, opened 2023-10-19
Error on Degree V12 Challenge


Computerdude77:
# Error Info
Error name: Error
Error message: panicked at /home/runner/.cargo/git/checkouts/nihav-vp6-660fee9efdd4622a/83c7e10/nihav-duck/src/codecs/vp56.rs:1030:18:
index out of bounds: the len is 2 but the index is 2
Error stack:
```
Error: panicked at /home/runner/.cargo/git/checkouts/nihav-vp6-660fee9efdd4622a/83c7e10/nihav-duck/src/codecs/vp56.rs:1030:18:
index out of bounds: the len is 2 but the index is 2
    at imports.wbg.__wbg_new_d258248ed531ff54 (chrome-extension://okdbphaljlmehkflmhmkaedhebdkdiej/dist/339.js:2479:21)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::h307f95dbfd4e7fe6 (wasm://wasm/032955f2:wasm-function[9716]:0x98c1a1)
    at std::panicking::rust_panic_with_hook::h3aa054d35a0817d7 (wasm://wasm/032955f2:wasm-function[6377]:0x890b79)
    at core::panicking::panic_fmt::h3e1dd3d08288569e (wasm://wasm/032955f2:wasm-function[9525]:0x98716a)
    at core::panicking::panic_bounds_check::h9245d4a825cc5107 (wasm://wasm/032955f2:wasm-function[8375]:0x933fb5)
    at nihav_duck::codecs::vp56::VP56Decoder::predict_dc::he755df5bc182e6f2 (wasm://wasm/032955f2:wasm-function[2147]:0x5b0ec2)
    at nihav_duck::codecs::vp56::VP56Decoder::decode_planes::hf1a32db45c1b2742 (wasm://wasm/032955f2:wasm-function[550]:0x837e1)
    at <ruffle_video_software::decoder::vp6::Vp6Decoder as ruffle_video_software::decoder::VideoDecoder>::decode_frame::hc2baecd7086898ae (wasm://wasm/032955f2:wasm-function[743]:0x270530)
    at <ruffle_video_software::backend::SoftwareVideoBackend as ruffle_video::backend::VideoBackend>::decode_video_stream_frame::hd2753e2339bc907f (wasm://wasm/032955f2:wasm-function[2276]:0x5d2805)
[...]

Issue #12327, opened 2023-07-26
VP6 videos not working on https://labcenter.dnalc.org/labs/restrictionanalysis/restrictionanalysis_d.html


danielhjacobs:
### Describe the bug

The FLV videos on this site do not play despite using the supported VP6 video codec.

### Expected behavior

[Screencast from 2023-07-26 15-02-52.webm](https://github.com/ruffle-rs/ruffle/assets/29206584/3cdc8933-7418-45c9-82e7-5b595a0f7c25)

### Affected platform

Browser's extension

### Operating system

Linux Fedora 37

### Browser

Mozilla Firefox 113

torokati44:
The actual problem here is the positioning of the video element. It is being flung rapidly to the right for some reason, and just barely appears on screen for a single frame:
![image](https://github.com/ruffle-rs/ruffle/assets/288816/c1f0b2c7-d0ee-4430-802f-9e943f878c75)

Just moments later, it is already far far away offscreen to the right:
![image](https://github.com/ruffle-rs/ruffle/assets/288816/dbbd9f67-ed00-455f-bd53-41a862ff997f)

So I'd chalk this up as an AVM1 interpreter issue, also based on some of the log output:

```
2023-12-09T10:55:57.249719Z  WARN run_frame: ruffle_core::library: Unknown device font "Myriad Pro"
2023-12-09T10:55:57.254087Z  WARN run_frame: ruffle_core::library: Unknown device font "Futura"
2023-12-09T10:55:57.294741Z  WARN run_frame: ruffle_core::library: Unknown device font "_sans"
2023-12-09T10:55:57.294863Z  WARN run_frame: ruffle_core::avm1::runtime: Avm1::pop: Stack underflow
2023-12-09T10:55:57.295169Z  WARN run_frame: ruffle_core::avm1::activation: Cannot enumerate Undefined
2023-12-09T10:55:57.296252Z  WARN run_frame: ruffle_core::avm1::object::stage_object: Unimplemented property _focusrect
2023-12-09T10:55:57.300381Z  WARN run_frame: ruffle_core::avm1::activation: Cannot enumerate Undefined
2023-12-09T10:55:57.300504Z  WARN run_frame: ruffle_core::avm1::object::stage_object: Unimplemented property _focusrect
2023-12-09T10:55:57.300611Z  WARN run_frame: ruffle_core::stub: Encountered stub: AVM2 flash.display.MovieClip.goto() with SWF 9 movie
[...]

PR #14339, opened 2023-12-08
FLV Seek Fixes

kmeisthax:
This fixes a handful of issues in the FLV seek logic that are collectively responsible for #14338, #13624, and possibly #14202. When we skip forward, we wind up saving a position that's one past the keyframe we *want* to be on, which is bad. This PR forces a backseek before we break out of the seek loop.

Furthermore, we don't execute no-op seeks anymore.

This PR is in draft because I need more testing. On Wireless Quest, the backseek fixed the panic, but the video didn't decode properly and I just got a blank screen with no error messages. Adding the no-op seek blocker completely fixes the issue. I still want to test this on other movies that do forward seeks to make sure I didn't break them (or, if they were broken, check to make sure they're fixed now).


danielhjacobs:
Might this improve #12327 or does that seem unrelated?

torokati44:
I too had hoped so, and it would make perfect sense, but unfortunately it doesn't seem to improve that... :/

torokati44 (approved):

PR #14345, opened 2023-12-08
wgpu: Fix overly strict assertion for setVertexBufferAt

Aaron1011:
The offsets for bound attributes can overlap, which caused us to compute the total size incorrectly. Fix this check to take overlapping into account.

### DIFF
--- a/render/wgpu/src/context3d/current_pipeline.rs
+++ b/render/wgpu/src/context3d/current_pipeline.rs
@@ -399,11 +399,25 @@ impl CurrentPipeline {
                     index_per_buffer.last_mut().unwrap()
                 };
 
-                // FIXME - assert that this matches up with the AS3-supplied offset
-                buffer_data.total_size += entry_size_bytes;
+                let offset_bytes = attr.offset_in_32bit_units * 4;
+
+                // ActionScript may call setVertexBufferAt with overlapping offsets, e.g.:
+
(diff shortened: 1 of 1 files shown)


torokati44 (approved):

render/wgpu/src/context3d/current_pipeline.rs
```diff
@@ -399,11 +399,25 @@ impl CurrentPipeline {
...
+                // setVertexBufferAt(1, buffer, 2, "float4"]
+                // setVertexBufferAt(2, buffer, 6, "float2"]
+                // ```
+                //
+                // To correctly compute the total size, we look at the furthest extent of any buffer
+                // (its offset plus its length), rather than simply adding up all of the lengths.
+                buffer_data.total_size = std::cmp::max(
+                    buffer_data.total_size,
+                    entry_size_bytes + offset_bytes as usize,
+                );
```
torokati44, line 417: I suppose this is written this way instead of `buffer_data.total_size = buffer_data.total_size.max(entry_size_bytes + offset_bytes as usize);` due to a stylistic preference? It is more symmetrical after all...

torokati44:
While a regression test for these kinds of things is always appreciated, I get that it's a lot more effort than the fix itself - which is fairly straightforward anyway - so I didn't block on that.

PR #14347, opened 2023-12-09
avm1: Support FLV date conversion

mtc-mlx:
This fixes a crash with Riva Web Player (https://rivavx.de/), which was used for some tutorial videos on LPKF plotters.

I could attach the files to demonstrate but I imagine it's probably not necessary for such a minor change.

It's my first time with Rust, so apologies in advance if I missed something obvious. Local timezone info doesn't seem to be stored in the ActionScript date object, so that property from the FLV date is unused.


torokati44:
Thank you, and welcome to the project! :wave:

core/src/avm1/flv.rs
```diff
@@ -24,6 +24,17 @@ fn avm1_object_from_flv_variables<'gc>(
     info_object.into()
 }
 
+fn avm1_date_from_flv_date<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    unix_time: f64,
+    _local_offset: i16,
```
kmeisthax, line 30: So, I'd have to write tests to check for certain how local offset winds up getting used, if at all. However, given that this takes UNIX timestamps, which are always UTC by default, having a local offset is silly, so my money's on unused.

kmeisthax (approved):

Issue #9220, opened 2023-01-19
Flatpak Support


Doomsdayrs:
**TL;DR Status**: the current Flathub submission:
* https://github.com/flathub/flathub/pull/5473

## Describe the feature

While ruffle is still in active development and not yet stable, it is still quite useful for users to play many flash games (such as BTD) that do operate well with the current support.

To make it easier for users on Linux distributions to use Ruffle, it would be nice to get ruffle bundled with flatpak and published to flathub.

[This](https://github.com/Doomsdayrs/flathub/tree/rs.ruffle.Ruffle) repository contains my current bundling of ruffle. There are issues I want to first solve before I can submit to flathub, of course with permission from ruffle-rs to do so.

## Issues

Most of these are all XDG compliance related.

- [x] Linux packaging requirements
These are some requirements I would like to get put upstream.
  - [x] Desktop file https://github.com/ruffle-rs/ruffle/pull/11227
  - [x] Metainfo file https://github.com/ruffle-rs/ruffle/pull/11227
  - [x] SVG Logo https://github.com/ruffle-rs/ruffle/pull/11226
- [x] Immediate crash due to clipboard errors, detailed below; Dealt with by using x11 fallback.

dginovker:
Clipboard crash is just missing permissions iirc. A good test would be to download Flatseal and enable all the permissions and see if it still crashes.

The SVG logo isn't required but highly encouraged, same thing with Desktop/metainfo file being upstreamed but it can be in your Flatpak repo. The only thing that is 100% required when submitting to Flathub is you have permission from the ruffle-rs maintainers

Doomsdayrs:
> Clipboard crash is just missing permissions iirc. A good test would be to download Flatseal and enable all the permissions and see if it still crashes.

Apologies, I doubt that the issue is with permissions. Please do follow the mention hole with https://github.com/YaLTeR/wl-clipboard-rs/issues/32 to see. It seems it might be improper usage of libraries.

Specifically flatpak support might have to wait on https://github.com/rust-windowing/winit/pull/2156 for proper clipboard handling.

> The only thing that is 100% required when submitting to Flathub is you have permission from the ruffle-rs maintainers

That is why I am here :D

Doomsdayrs:
One solution is to allow Clipboard to become null, and simply warn in console that it will not operate. But I despise this type of hacky solution and would ruin user experience.

n0samu:
[...]

PR #14362, opened 2023-12-11
Fix panic on Wayland

bi4k8:
When running current ruffle HEAD on Wayland, I get an error popup with the following contents:
> Ruffle has encountered a fatal error, this is a bug.
> 
> panicked at desktop/src/player.rs:147:22:
> Couldn't create ui backend: Couldn't get platform clipboard
> 
> Caused by:
>     Unknown error while interacting with the clipboard: Display parsing error
> 
> Please report this to us so that we can fix it. Thank you!

This appears to be caused by using the "arboard" crate without the feature flag required for its Wayland support: https://github.com/1Password/arboard/issues/98

This PR enables this feature and fixes the panic.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -223,6 +223,7 @@ dependencies = [
  "parking_lot",
  "thiserror",
  "winapi",
+ "wl-clipboard-rs",
  "x11rb",
 ]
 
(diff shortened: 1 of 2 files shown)


n0samu:
Thanks so much! I think this should also help unblock #9220.

bi4k8:
> Interestingly, I've been running Ruffle on Wayland for well over a year now, and never had any issues with it, even the clipboard works just fine (at least for text), both in Weston and KDE.

If you have XWayland around, clipboard integration probably worked for you through the X11 codepaths. I don't use XWayland so I noticed the missing native clipboard support.

> One thing is, please tag the commit messages as advised by `CONTRIBUTING.md`. I'll fix it this time, as I'm going to squash your commits on merge anyway.

My mistake--I'll follow convention if I have anything else to PR. Thanks for the project!

Issue #14294, opened 2023-12-03
(AS3 Game) Error dispatching event EventObject


Lujek69:
### Describe the bug

There is an EventObject error in the browser console when trying to enable the game

Full error text:
>ERROR core/src/avm2/events.rs:424 Error dispatching event EventObject(EventObject { type: "complete", class: flash.events::Event, ptr: 0x3005f28 }) to handler FunctionObject(FunctionObject { ptr: 0x2095aa0 }) : ReferenceError: Error #1056: Cannot create property needsSoftKeyboard on flash.text.TextField.
![screenshotbug](https://github.com/ruffle-rs/ruffle/assets/54140042/5a51e904-7fc4-460e-9a27-edfde4e28212)

### Expected behavior

After loading successfully, the game displays the character creation screen as in the screenshot
![runned](https://github.com/ruffle-rs/ruffle/assets/54140042/403512f2-28bd-43b6-a2cd-e3a0e7748f7c)

### Content Location

https://crazyisland.5v.pl/play/

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

Mozilla Firefox 120.0.1

Aaron1011:
This seems to be working now

PR #14397, opened 2023-12-14
avm2: Stub InteractiveObject needsSoftKeyboard and softKeyboardInputAreaOfInterest

evilpie:
This improves #14294. Creating an account doesn't work, but login seems to be possible, but I don't have any valid credentials.


torokati44 (approved):

Lujek69:
> This improves #14294. Creating an account doesn't work, but login seems to be possible, but I don't have any valid credentials.

After logging into the account, it doesn't load the rest of the graphics even though they are all available on the server (probably a similar issue to the character creator where faces are not visible).
Test account credentials: username: xTest password: test123
test
After logging in, it should be like this:
test2

PR #14409, opened 2023-12-15
swf: Make avm2 Writer::write_op public

friedkeenan:
This adds a test for `swf::avm2::write::Writer::write_op`, and makes it public, since it was private and unused before, and I myself will want to write ops in the future in order to patch the bytecode of SWF files.

The test is admittedly not all-encompassing (mainly in terms of explicitly testing that `write_u30` is used, since I don't use any values that would result in `write_u30` writing more than one byte), but there is an assert for every `Op` variant, and it was written without looking at the implementation of `write_op`, only at `Reader::read_op`.

I did consider changing the implementation of `write_op` by adding an `Op::opcode` method that would be used in `write_op` at the start to write the opcode, and then after would match on the op and write the extra data if applicable, but I wasn't sure if it was worth it, though I think an `Op::opcode` method could maybe be otherwise useful in general. If that's something that's desired, I could do that, though perhaps it's outside the scope of this PR.

### DIFF
--- a/swf/src/avm2/write.rs
+++ b/swf/src/avm2/write.rs
@@ -589,8 +589,7 @@ impl<W: Write> Writer<W> {
         Ok(())
     }
 
-    #[allow(dead_code)]
-    fn write_op(&mut self, op: &Op) -> Result<()> {
+    pub fn write_op(&mut self, op: &Op) -> Result<()> {
         match *op {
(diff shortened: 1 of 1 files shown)


kmeisthax (approved):

torokati44 (approved):

torokati44:
Thank you!

PR #14410, opened 2023-12-15
avm2: Remove ArrayObject `to_string` impl

sleepycatcoding:
This was causing the `Object.prototype.toString` to throw error 1050, instead of returning `[object Array]`, which was causing quite a few avmplus test failures.


n0samu (approved):

torokati44 (approved):
Huh, nice catch! :tada:

PR #14418, opened 2023-12-16
web/extension: Correct the version used for the Firefox extension

evilpie:
The version numbers must strictly equal one of those in these lists:
- https://addons.mozilla.org/api/v5/applications/firefox/
- https://addons.mozilla.org/api/v5/applications/android/

### DIFF
--- a/web/packages/extension/tools/sign_xpi.js
+++ b/web/packages/extension/tools/sign_xpi.js
@@ -61,10 +61,10 @@ async function sign(
         json: {
             compatibility: {
                 firefox: {
-                    min: "84",
+                    min: "84.0",
                 },
                 android: {
-                    min: "120",
+                    min: "120.0",
                 },
             },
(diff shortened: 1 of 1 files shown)


evilpie:
Aside: For some reason the latest public version uses `48.0` and not `84.0` in desktop? https://addons.mozilla.org/api/v5/addons/addon/ruffle_rs/versions/0.1.0.1051/. Probably doesn't matter, both versions are incredibly outdated.

torokati44:
> Aside: For some reason the latest public version uses 48.0 and not 84.0 on desktop? https://addons.mozilla.org/api/v5/addons/addon/ruffle_rs/versions/0.1.0.1051/. Probably doesn't matter, both versions are incredibly outdated.

I think this is because `"84"` was also taken as invalid, and this could be some sort of built-in fallback for that case...?

torokati44 (approved):

PR #14420, opened 2023-12-16
core/avm1: Improve some stubs

Lord-McSweeney:
Miscellaneous stub fixes

### DIFF
--- a/core/src/frame_lifecycle.rs
+++ b/core/src/frame_lifecycle.rs
@@ -124,7 +124,7 @@ pub fn run_inner_goto_frame<'gc>(
     removed_frame_scripts: &[DisplayObject<'gc>],
     initial_clip: MovieClip<'gc>,
 ) {
-    if initial_clip.swf_version() <= 9 {
+    if initial_clip.swf_version() <= 9 && initial_clip.movie().is_action_script_3() {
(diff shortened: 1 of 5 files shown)


core/src/frame_lifecycle.rs
```diff
@@ -124,7 +124,7 @@ pub fn run_inner_goto_frame<'gc>(
     removed_frame_scripts: &[DisplayObject<'gc>],
     initial_clip: MovieClip<'gc>,
 ) {
-    if initial_clip.swf_version() <= 9 {
+    if initial_clip.swf_version() <= 9 && initial_clip.movie().is_action_script_3() {
```
torokati44, line 127: Is this line also only about improving AVM1 stubs? If so, how?

Lord-McSweeney: This should fix the `Encountered stub: AVM2 flash.display.MovieClip.goto with SWF 9 movie` warning appearing on AVM1 movies that attempt to run a goto. The stub warning should only appear in AVM2 because it is only relevant to AVM2.

torokati44: Ah, I see, okay!

torokati44 (approved):

Issue #12604, opened 2023-08-07
catch the candy - regression and the display of the pause button when the mouse is hovering over it is not correct


zjk261:
### Describe the bug

can't pause the game

https://github.com/ruffle-rs/ruffle/assets/98996990/6a954c10-52a6-4d36-ba32-eada58276e8b

button
![ruffle](https://github.com/ruffle-rs/ruffle/assets/98996990/dfcd5fb0-1348-49f8-94a8-cdaa1aadf92e)

[catch-the-candy-no-sitelock.zip](https://github.com/ruffle-rs/ruffle/files/12270927/catch-the-candy-no-sitelock.zip)

### Expected behavior

https://github.com/ruffle-rs/ruffle/assets/98996990/bf773fe1-d7d7-4b7c-9308-12e6d9a91004

![fp](https://github.com/ruffle-rs/ruffle/assets/98996990/d1be2023-6433-44d2-8d90-b81add5a5f2c)

### Affected platform

Desktop app

### Operating system

Windows 10

n0samu:
This is a regression caused by #12157. The game's MouseWheel class is using LocalConnection for some reason, and the stub fools the game's logic into going into an infinite loop.

Lord-McSweeney:
This needs two things:
Proper LocalConnection impl (including cross-avm)
Mixed AVM

Lord-McSweeney:
Both issues seem to be fixed now

PR #14424, opened 2023-12-17
avm2/core: Partially implement `LocalConnection`

Lord-McSweeney:
This is an initial implementation of `LocalConnection` that only considers connections local to the specific SWF file. It adds a `LocalConnections` struct to `UpdateContext`, and implements the `LocalConnection` APIs for AVM2.

I also changed `LocalConnection.send` to not immediately fire the `status` event, to match Flash. #12604 is _technically_ fixed by this, but I would like to leave the issue open until AVM1/mixed AVM `LocalConnection` works (since that's what the SWF really needs).

### DIFF
--- a/core/src/avm2/globals/flash/net/local_connection.rs
+++ b/core/src/avm2/globals/flash/net/local_connection.rs
@@ -1,10 +1,12 @@
-use crate::avm2::error::type_error;
+use crate::avm2::error::{argument_error, type_error};
 use crate::avm2::object::TObject;
 use crate::avm2::parameters::ParametersExt;
 use crate::avm2::{Activation, Avm2, Error, Object, Value};
 use crate::string::AvmString;
 
 use crate::avm2_stub_method;
+
+pub use crate::avm2::object::local_connection_allocator;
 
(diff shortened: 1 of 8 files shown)


core/src/avm2/globals/flash/net/local_connection.rs
```diff
@@ -53,7 +57,24 @@ pub fn send<'gc>(
         )?));
     }
 
-    avm2_stub_method!(activation, "flash.net.LocalConnection", "send");
+    // FIXME: The rest of the logic in this function should actually run
```
Aaron1011, line 60: Since you're using `dispatchStatusEventWithDelay`, this comment is outdated, right?

Aaron1011 (approved):

core/src/avm2/globals/flash/net/local_connection.rs
Lord-McSweeney: I was referencing the entire rest of the function, which includes checking if there are any active listeners that have called `connect` on the connection. I think I can move that entire logic into a `setTimeout` now.

Issue #14446, opened 2023-12-18
Commando Assault no longer works


n0samu:
### Describe the bug

The nightly 2023-12-18 release broke [Commando Assault](https://ooooooooo.ooo/static/?d38fcefb-89c5-4977-91ba-14237ca5e6e4); it now shows nothing but a blank screen. The error is:
```
ERROR run_frame:run_all_phases_avm2: ruffle_core::display_object::movie_clip: Got "ReferenceError: Error #1069: Property flash.events:IEventDispatcher::addEventListener not found on com.miniclip._MiniclipGameManager and there is no default value.
        at com.macrojoy.zgd::Document/checkDomainAndStartGame()
        at com.macrojoy.zgd::Document()" when constructing AVM2 side of movie clip of type com.macrojoy.zgd::Document
```

### Expected behavior

The game should load.

### Content Location

Page URL: https://ooooooooo.ooo/static/?d38fcefb-89c5-4977-91ba-14237ca5e6e4
SWF: [Commando_Assault.zip](https://github.com/ruffle-rs/ruffle/files/13699067/Commando_Assault.zip)

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

Firefox 120

Lord-McSweeney:
Weird, this shouldn't be `VM_INTERNAL`: `Rejecting namespace match due to versions: Namespace("flash.events:IEventDispatcher", VM_INTERNAL) Namespace("flash.events:IEventDispatcher", AllVersions)`

ThrowAwaySDXR11:
I've been having similar issues loading Commando Assault. Inspecting the code of other websites that successfully load it "ex: Gameflare", the game seems to run on multiple SWF files "/game.swf" along with the five maps "/resource/map_easy.swf, /map_normal.swf, hard.swf, hell, final". I tried plugging all six SWFs into Ruffle but it still showed up as black. Maybe there is a way to get it to work this way, but I am not sure.

n0samu:
@ThrowAwaySDXR11 the game is sitelocked

ThrowAwaySDXR11:
Oh I see. So its impossible to get it working on another site. Good to know!

ThrowAwaySDXR11:
You have to modify the document file in the main swf to not crash when its not on miniclip.com. Also ruffle can only load the map files on (wgpu-webgl) so set that as the preferred renderer. It should work after that, I got it working on mine.

Issue #12810, opened 2023-08-17
SpongeBob: Halloween Horror - FrankenBob's Quest Part 1: Preloader Fail to Start on Main.


p0008874:
### Describe the bug

![image](https://github.com/ruffle-rs/ruffle/assets/75534590/cd0eaadb-7c38-4b8a-a0e7-b3c98405ee22)

```
2023-08-17T16:11:59.300284Z ERROR run_frame: ruffle_core::display_object::movie_clip: Got "ReferenceError: Error #1065: Variable Main is not defined.
        at global/flash.utils::getDefinitionByName()
        at com.sarbakan.sbdk.core::AbstractApplication()
        at Main()" when constructing AVM2 side of movie clip of type Main
```

### Expected behavior

![image](https://github.com/ruffle-rs/ruffle/assets/75534590/94ababf0-feec-4637-bec3-1d06a486d26a)

Ruffle nightly 2023-06-15

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

https://github.com/ruffle-rs/ruffle/files/11557190/spongebob-squarepants-halloween-horror-frankenbobs-quest.zip

p0008874:
```
2023-12-18T06:20:11.210613Z  INFO ruffle_core::player: Loaded SWF version 9, resolution 750x500 @ 35 FPS
2023-12-18T06:20:11.213087Z  WARN run_frame: ruffle_core::library: Unknown device font "_sans"
2023-12-18T06:20:11.213466Z  WARN run_frame: ruffle_core::library: Unknown device font "_sans"
2023-12-18T06:20:11.219353Z  WARN run_frame:run_all_phases_avm2: ruffle_core::stub: Encountered stub: AVM2 flash.utils.Dictionary constructor with weak keys
2023-12-18T06:20:11.219690Z  INFO run_frame:run_all_phases_avm2: ruffle_core::avm2::namespace: Rejecting namespace match due to versions: Namespace("flash.events:IEventDispatcher", VM_INTERNAL) Namespace("flash.events:IEventDispatcher", AllVersions)
2023-12-18T06:20:11.220441Z ERROR run_frame:run_all_phases_avm2: ruffle_core::display_object::movie_clip: Got "ReferenceError: Error #1069: Property flash.events:IEventDispatcher::addEventListener not found on MainPreloader and there is no default value.
        at com.sarbakan.sbdk.utils::EventManager/addEventListener()
[...]

PR #14449, opened 2023-12-18
avm2: Avoid marking interface method definitions with VM_INTERNAL

Aaron1011:
Our asc.jar doesn't seem to apply a version suffix to namespaces for interface method definitions. This was causing these methods to get marked as VM_INTERNAL when we loaded playerglobals, preventing SWF from invoking these methods through the interface (e.g. having a variable of type `IEventDispatcher`, and calling `dispatchEvent` on it)


Lord-McSweeney:
This fixes #14446 and fixes the regression in #12810

core/src/avm2/globals/flash/desktop/IFilePromise.as
```diff
@@ -0,0 +1,24 @@
+package flash.desktop {
+    import flash.utils.IDataInput;
+    import flash.events.ErrorEvent;
+
+    [API("668")] // AIR 2.0
+    public interface IFilePromise {
+
+        [API("668")]
```
Lord-McSweeney, line 8: Aren't these unnecessary since the interface itself is version-gated?

Aaron1011: Probably, but it doesn't hurt to be sure

evilpie: Personally I think we should avoid adding what is basically noise. If we have bugs with methods/properties being exposed unexpectedly those can be fixed.

Dinnerbone (approved):

PR #14461, opened 2023-12-18
ci: Switch to `cargo-nextest` for running Rust tests

torokati44:
This is supposed to avoid https://github.com/gfx-rs/wgpu/issues/4885, as `nextest` runs every test in its own process.
Supposedly for the same reason, WARP is a lot less delicate, therefore #14393 is mostly reverted.
There still seem to be some sort of random driver [de]initialization issue on Ubuntu, but with allowing a few retries, this should hopefully not actually cause failures often. EDIT: Allowed retrying test because of this.
It's also nicer to debug failures, because it's always obvious exactly which test case failed.
Run on 4 threads at once, to overlap CPU-bound and IO-bound work.

I have considered adding a file in `.config` instead of the 3-4 command line arguments, but I think this is still fine as is.

### DIFF
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -68,6 +68,7 @@ jobs:
           save-if: ${{ github.ref == 'refs/heads/master' }}
 
       - name: Check formatting
+        if: runner.os == 'Linux'
         run: cargo fmt --all -- --check
 
       - name: Check clippy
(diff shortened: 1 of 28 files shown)


torokati44:
Currently waiting on whether https://github.com/nextest-rs/nextest/issues/1172 gets done soon or not.

torokati44:
With that TODO comment I just added, I'm okay for mergint this now.
I'll leave the decision up to you whether this is acceptable as is.

Aaron1011 (approved):

PR #14489, opened 2023-12-23
avm2: Support more ElementFormat font options in TextBlock

evilpie:
TLF also seems to support device fonts, so I enabled that flag. (I am testing this with device font support)


Lord-McSweeney (approved):

Lord-McSweeney:
Have you checked the behavior when any of the `FontDescription` properties are null?

PR #14490, opened 2023-12-23
avm2: Micro-optimize coerce_to_i32/u32

evilpie:


Dinnerbone (approved):

core/src/avm2/value.rs
```diff
@@ -755,7 +755,15 @@ impl<'gc> Value<'gc> {
     /// Numerical conversions occur according to ECMA-262 3rd Edition's
     /// ToUint32 algorithm which appears to match AVM2.
     pub fn coerce_to_u32(&self, activation: &mut Activation<'_, 'gc>) -> Result<u32, Error<'gc>> {
-        Ok(f64_to_wrapping_u32(self.coerce_to_number(activation)?))
+        Ok(match self {
+            Value::Undefined | Value::Null => 0,
+            Value::Bool(b) => *b as u32,
+            Value::Integer(i) => *i as u32,
```
adrian17, line 761: Can we put them roughly in order of probability (as in, Integer first)? I know this isn’t guaranteed to impact codegen, but sounds like a good idea either way.

Lord-McSweeney (approved):

Issue #2044, opened 2020-12-19
Cannot Play Games on Friv


Fancy2209:
Every other part of friv like the sidebar and the menu works
But when I try to open A game it Fails to fetch the swf like seen in the video

https://user-images.githubusercontent.com/64917206/102698940-a50eb500-4230-11eb-8bb7-fa3e25bcedb4.mp4

these are the logs
```javascript
core.ruffle.js:1 Creating WebGL renderer...
core.ruffle.js:1 Falling back to WebGL1.
core.ruffle.js:1 WebGL graphics driver: ANGLE (Intel(R) HD Graphics Family Direct3D9Ex vs_3_0 ps_3_0)
core.ruffle.js:1 Minimum audio buffer sample rate: 5512
VM83:3 New Ruffle instance created.
core.ruffle.js:1 PLAY!
core.ruffle.js:1 Creating WebGL renderer...
core.ruffle.js:1 Falling back to WebGL1.
core.ruffle.js:1 WebGL graphics driver: ANGLE (Intel(R) HD Graphics Family Direct3D9Ex vs_3_0 ps_3_0)
core.ruffle.js:1 Minimum audio buffer sample rate: 5512
VM83:3 New Ruffle instance created.
core.ruffle.js:1 PLAY!
VM83:3 Loading SWF file ../../lb.swf?gametype=swf&invtype=normal&gamename=Fancypants-Adventure&gamecontrols=spcSQuadalaralmbmvmh
VM83:3 Loading SWF file http://l.funo.com/z/fancypantsadventure/fancypantsadventure.swf
core.ruffle.js:1 Loaded SWF version 9, with a resolution of 180x700
core.ruffle.js:1 [Violation] 'requestAnimationFrame' handler took 67ms
f.html?z=0l:1 Access to fetch at 'http://l.funo.com/z/fancypantsadventure/fancypantsadventure.swf' from origin 'http://www.frivclassic.com' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
core.ruffle.js:1 GET http://l.funo.com/z/fancypantsadventure/fancypantsadventure.swf net::ERR_FAILED
Oe @ core.ruffle.js:1
__wbg_fetch_f532e04b8fe49aa0 @ VM83:3
[...]

PR #14492, opened 2023-12-23
wgpu: Allow Stage3D texture upload where source is smaller than dest

Aaron1011:
Fancy Pants World 4 relies on this behavior.
Fixes https://github.com/ruffle-rs/ruffle/issues/14312


tests/tests/swfs/avm2/stage3d_sampler_partial_upload/test.toml
```diff
@@ -0,0 +1,8 @@
+num_frames = 1
+
+[image_comparisons.output]
+tolerance = 1
+max_outliers = 782
+
+[player_options]
+with_renderer = { optional = true, sample_count = 1, exclude_warp = true }
```
Lord-McSweeney, line 8: Shouldn't this be `optional = false`?

tests/tests/swfs/avm2/stage3d_sampler_partial_upload/test.toml
```diff
@@ -0,0 +1,8 @@
+num_frames = 1
+
+[image_comparisons.output]
+tolerance = 1
+max_outliers = 782
```
Lord-McSweeney, line 5: This is pretty high- can this be decreased by increasing `tolerance`?

Aaron1011: I'll try this in a follow up

torokati44:
The attached test only shows up as a blank white image for me on Linux on Flash Player (both regular and debugger) `32.0.0.465` on Linux... :/

And I get this this message in `flashlog.txt`: `Error #2044: Unhandled ErrorEvent:. text=Error #3702: Context3D not available.`

torokati44:
In fact, the same is true for the `avm2/stage3d_rotating_cube` test! :scream:
But `avm2/stage3d_fractal` and `avm2/stage3d_rotating_cube` for example work fine.

core/src/avm2/object/context3d_object.rs
```diff
@@ -353,12 +353,27 @@ impl<'gc> Context3DObject<'gc> {
     ) {
         let source = source.read();
 
-        assert_eq!(source.width(), dest.width(), "Mismatched width");
-        assert_eq!(source.height(), dest.height(), "Mismatched height");
+        // Note - Flash appears to allow a source that's larger than the destination.
+        // Let's leave in this assertion to see if there any real SWFS relying on this
+        // behavior.
```
torokati44, line 358: This sentence reads a bit weird to me. But I get the meaning, so it's not important.
Anyway, something this might sound a bit better IMHO:
> Let's leave this assertion in to see if there are any real SWFs relying on this behavior.

torokati44:
Otherwise, the changes themselves LGTM (note, again, that I'm not really familiar with these parts...).
I'd just like to be able to run the test in FP myself for comparison.

Aaron1011:
I've never been able to run those tests on Linux - I've always used a windows vm. Flash's stage3d support on linux is pretty bad, unfortunately.

Dinnerbone (approved):

torokati44:
> I've never been able to run those tests on Linux - I've always used a windows vm. Flash's stage3d support on linux is pretty bad, unfortunately.

Ah, I see. That's unfortunate indeed.

Issue #12303, opened 2023-07-25
Wizard Tower runs very slowly


Lord-McSweeney:
### Describe the bug

The SWF is available at https://lodev.org/wizardtower/wizardtower.swf

The game runs slowly on web, with the game sometimes taking up to 2-3 seconds to respond to key strokes.

### Expected behavior

The game runs at a playable pace.

### Affected platform

Browser's extension

### Operating system

Linux

### Browser

Chrome

Lord-McSweeney:
This works at a normal speed on desktop now

Issue #10560, opened 2023-04-05
Evoland: Severe performance issues


n0samu:
### Describe the bug

Link to SWF: http://web.archive.org/web/20171215200434oe_/http://evoland.shirogames.com/swf/classic/ld24.swf

The crash #9578 is fixed, but the game still has severe performance issues. After clicking the Start button, with default settings Ruffle hits a script timeout. With the timeout bumped to 100 seconds, the game eventually starts, but it takes multiple seconds to draw each frame, so it's not anything close to playable.

### Expected behavior

The game should run at something like normal speed.

### Affected platform

Desktop app

### Operating system

Windows 10

Lord-McSweeney:
The SWF does not play on web, it must be opened in the desktop version of Ruffle.

Lord-McSweeney:
For some reason this only works for me when I'm on the OpenGL graphics backend (on Linux)- the game just freezes up after clicking "Start" and doesn't play (even after increasing the max execution duration to 450 seconds).

Lord-McSweeney:
This also occasionally errors `RustError("Resolved parameter type void is unresolvable, not a class, null, or undefined")`.

Lord-McSweeney:
This is using very inefficient code that duplicates a number of coercions, which increases the size of functions. #14631's optimizer detects many of the duplicate coercions and replaces them with `nop`s, and it also pre-reads every function, which results in clearly visible (at least for me) performance improvements with that PR.

Lord-McSweeney:
I was wrong about the previous comment- the _one function_ that causes all the lag is `Game/applyMask`, which runs a loop on a bytearray every frame. Executing the function takes >200 ms on my computer.
(`output` is the `BitmapData` that everything is rendered to)
```
        public function applyMask(param1:int, param2:int) : void
        {
            var _loc7_:int = 0;
            var _loc8_:int = 0;
            var _loc9_:int = 0;
            var _loc10_:int = 0;
            var _loc3_:ByteArray = output.getPixels(output.rect);
            ApplicationDomain.currentDomain.domainMemory = _loc3_;
            var _loc4_:int = 0;
            var _loc5_:int = 0;
            var _loc6_:int = output.width * output.height;
            while(_loc5_ < _loc6_)
            {
                _loc7_ = _loc5_++;
                _loc8_ = li8(_loc4_) + param1;
                if(_loc8_ > 255)
                {
                    _loc8_ = 255;
                }
[...]

PR #14496, opened 2023-12-24
avm2: Optimize the interpreter loop

evilpie:
**Edit: Most of the changes described below have been abandoned in favor of https://github.com/ruffle-rs/ruffle/pull/14563**

Currently the very hot code in `do_next_opcode` is basically split into two parts:
- Reading/decoding the opcode via `Reader::read_op`
- Dispatching based on the decoded data

This doesn't seem to produce very good code. Most interpreters that I know fuse decoding and dispatching. In this proof of concept the compiler seems to generate jump tables for at least a subset of opcodes. 

I am not sure if I trust the numbers, but the time spent in a very tight loop like this seems to be reduced by 25%

```actionscript
var foo: int = 0xFFFFFFFF;
for (var i: int = 0; i < 100000; i++) {
    for (var j: int = 8; j >= 0; j--) {
        var bar: int = -(foo & 1);
        foo = (foo >> 1) ^ (0xEDB88320 & bar);
    }
}
```

### DIFF
--- a/core/src/avm2.rs
+++ b/core/src/avm2.rs
@@ -590,10 +590,16 @@ impl<'gc> Avm2<'gc> {
         self.call_stack
     }
 
+    #[cold]
+    fn stack_overflow(&self) {
+        tracing::warn!("Avm2::push: Stack overflow");
+    }
+
     /// Push a value onto the operand stack.
+    #[inline(always)]
     fn push(&mut self, value: impl Into<Value<'gc>>, depth: usize, max: usize) {
         if self.stack.len() - depth > max {
-            tracing::warn!("Avm2::push: Stack overflow");
+            self.stack_overflow();
             return;
         }
         let mut value = value.into();
@@ -610,11 +616,17 @@ impl<'gc> Avm2<'gc> {
         self.stack.push(value);
     }
 
+    #[cold]
+    fn stack_underflow(&self) {
+        tracing::warn!("Avm2::pop: Stack underflow");
+    }
+
     /// Retrieve the top-most value on the operand stack.
     #[allow(clippy::let_and_return)]
+    #[inline(always)]
     fn pop(&mut self, depth: usize) -> Value<'gc> {
         let value = if self.stack.len() <= depth {
-            tracing::warn!("Avm2::pop: Stack underflow");
+            self.stack_underflow();
             Value::Undefined
         } else {
             self.stack.pop().unwrap_or(Value::Undefined)
--- a/core/src/avm2/activation.rs
+++ b/core/src/avm2/activation.rs
@@ -921,6 +921,7 @@ impl<'a, 'gc> Activation<'a, 'gc> {
     }
 
     /// Run a single action from a given action reader.
+    #[inline(always)]
     fn do_next_opcode<'b>(
         &mut self,
         method: Gc<'gc, BytecodeMethod<'gc>>,


evilpie:
Okay. I was wrong, there still seem to be jump tables in the current code. So all the overhead must be from copying the Opcode and also the new code seems to trigger more of the op implementations to be inlined.

Aaron1011 (approved):

Aaron1011:
@adrian17 Does this conflict with any of your plans around bytecode verification/optimization?

Lord-McSweeney:
This optimization will have no effect after a verifier/optimizer implementation, as in those implementations, the bytecode is read and stored (and optimized) before actually running the code.

adrian17:
Yeah, the Vec<Op> change would fully replace it and it'd resemble the old implementation again.
That said, I _am_ curious to compare these two as-is, as in: I can compare this PR vs my branch with optimizations disabled (as in, measure just the Vec<Op> impact). But I'm still on leave until the new year's, so please give me just a couple more days ❤️ 
(I also had some other small ideas to compare, like changing the implementation to unpack the Ops after the call instead of in the main loop and calling the ops with separate arguments)

The change to make under/overflow cold is unrelated but sounds really nice, maybe adding inline-never would be even better? Not sure, would have to compare on some microbenchmark. (very long term the verifier would probably replace these too, but that doesn't mean these changes aren't good to have now.)

Lord-McSweeney:
I think you can do `e.into()` instead of `Error::RustError(Box::new(e))`

Lord-McSweeney:
I did some measurements on #10560 on my own verifier branch, with optimizations disabled, and (if I tested correctly) time spent running each frame was cut by more than 50%. I got similar, though smaller (30%) improvements on #12303, so there are a few games that rely on bytecode being cached and not re-read each time a function is executed.

evilpie:
> I think you can do `e.into()` instead of `Error::RustError(Box::new(e))`

I think we would need a new From implementation?
```
error[E0282]: type annotations needed
   --> core/src/avm2/activation.rs:964:36
    |
964 |                     .map_err(|e| e.into())?,
    |                                    ^^^^
    |
help: try using a fully qualified path to specify the expected types
    |
964 |                     .map_err(|e| <swf::error::Error as Into<T>>::into(e))?,
    |                                  +++++++++++++++++++++++++++++++++++++ ~

error[E0283]: type annotations needed
   --> core/src/avm2/activation.rs:964:36
    |
964 |                     .map_err(|e| e.into())?,
    |                                    ^^^^   - type must be known at this point
    |
note: multiple `impl`s satisfying `avm2::error::Error<'gc>: From<_>` found
   --> core/src/avm2/error.rs:542:1
    |
542 | impl<'gc, 'a> From<&'a str> for Error<'gc> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
548 | impl<'gc> From<String> for Error<'gc> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
554 | impl<'gc> From<ruffle_render::error::Error> for Error<'gc> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: and more `impl`s found in the following crates: `core`:
            - impl<T> From<!> for T;
            - impl<T> From<T> for T;
    = note: required for `Result<avm2::activation::FrameControl<'gc>, avm2::error::Error<'gc>>` to implement `FromResidual<Result<Infallible, _>>`
help: try using a fully qualified path to specify the expected types
    |
964 |                     .map_err(|e| <swf::error::Error as Into<T>>::into(e))?,
    |                                  +++++++++++++++++++++++++++++++++++++ ~

Some errors have detailed explanations: E0282, E0283.
For more information about an error, try `rustc --explain E0282`.
error: could not compile `ruffle_core` (lib) due to 2 previous errors
```

evilpie:
Okay, I am going to remove all the parts that change the opcode reading and we can focus on https://github.com/ruffle-rs/ruffle/pull/14563 instead.

torokati44 (approved):

adrian17 (approved):

Issue #14498, opened 2023-12-24
(NOT RUFFLE COMPATIBLE) Custom Pixel Bender Example | Ordered Dithering


FlashBacks1998:
### Describe the bug

Whenever I use Pixel Bender for Ruffle It doesent seem to work. I wrote a basic Hello World example with ordered dithering made though a Pixel Bender Kernel and it works just fine in Flash Player 11.2. However in Ruffle it doesent seem to work as *something* is being rendered since the screen is black and not the stage default background purple.

### Expected behavior

A dithered image. 
![Capture2](https://github.com/ruffle-rs/ruffle/assets/152461061/9022a04d-0db0-47ab-9281-d1b40e245dc4)
![Capture](https://github.com/ruffle-rs/ruffle/assets/152461061/52ea5ac0-b569-460e-91cd-7d19e0c99062)

### Content Location

Github Repository with compiled .binaries (swf and pixel bender) and source code (fla and pixel bender)
https://github.com/FlashBacks1998/Custom-Pixel-Bender-Example-Ordered-Dithering- 

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

Brave

PR #14503, opened 2023-12-24
render: Implement PixelBender Operation::Select (renamed from Loop)

Aaron1011:
This is not actually an loop opcode - it's a 'ternary if' opcode, which selects between two input registers based on a condition register, and writes to a destination register.

Fixes https://github.com/ruffle-rs/ruffle/issues/14498


evilpie:
I would just call it  `select` like llvm or mesa.

torokati44:
It's strange that `pbj` thinks it's a loop opcode[1], and `pbjas` simply skips it[2].

I guess it's not used that often...?

 [1]: https://github.com/HaxeFoundation/format/blob/39787764801f9e02c5b5ed771490e767a5488e65/format/pbj/Reader.hx#L258-L259
 [2]: https://github.com/jamesward/pbjas/blob/871e66d6289a089df6dc7511ef8921100a5bd630/pbjAS/pbjAS/PBJAssembler.as#L527-L546

tests/tests/swfs/avm2/pixelbender_dithering/test.toml
```diff
@@ -0,0 +1,10 @@
+# Based on  https://github.com/FlashBacks1998/Custom-Pixel-Bender-Example-Ordered-Dithering-
```
torokati44, line 1: absolute nit: double space

tests/tests/swfs/avm2/pixelbender_dithering/test.toml
```diff
@@ -0,0 +1,10 @@
...
+# The prints were sorted to avoid relying on object iteration order
+num_ticks = 1
+
+[image_comparisons.output]
+tolerance = 1
+max_outliers = 19813
```
torokati44, line 7: That's... um... a lot?
Perhaps a bit higher tolerance could be considered instead?

torokati44:
@evilpie
> I would just call it select like llvm or mesa.

I'm reading in at least two places (https://github.com/eclipse/omr/pull/668#issuecomment-273605969 and https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/Control_flow/Select) that in general, the difference between "ternary if" and "select" is that the former only evaluates the single branch whose value is to be returned, while the latter evaluates both (or all, in the general case) branches, and _selects_ a value to return afterwards.

Perhaps we should follow this convention with the naming - that is, if the difference is observable in PB at all.

lvyitian:
In https://web.archive.org/web/20100223210221/http://www.kaourantin.net/source/pbjtools/apbj.cpp, opcode 0x33 is called "select".

torokati44:
Ooh, nice find! The entire blog archive in fact!

Aaron1011:
I had based my changes off of `apbj.cpp`, but I preferred `TernaryIf` to `Select`. I can change it to `Select` to follow the convention, though (the opcode uses PixelBender registers, so both 'arguments' are always evaluated)

torokati44:
I'd say that's two reasons for going with `Select` then.

torokati44:
We were talking about `Select`, not `Switch`. 😶

Aaron1011:
Oops - I don't know why I did `Swich`

render/naga-pixelbender/src/lib.rs
```diff
@@ -1388,8 +1388,84 @@ impl<'a> ShaderBuilder<'a> {
...
+                } => {
+                    let src1_expr = self.load_src_register(src1)?;
+                    let src2_expr = self.load_src_register(src2)?;
+
+                    let expr_zero: Handle<Expression> = match condition.kind {
+                        PixelBenderRegKind::Float => self.zerof32,
+                        PixelBenderRegKind::Int => self.zeroi32,
+                    };
+                    if condition.channels.len() != 1 {
+                        panic!("Ternary if condition must be a scalar: {condition:?}");
```
evilpie, line 1405: s/ternary if/select/

evilpie:
I was wondering if we could use the https://docs.rs/naga/latest/naga/enum.Expression.html#variant.Select?

Aaron1011:
@evilpie Good idea - I've updated the pr

evilpie:
LGTM

torokati44:
I'd still like a bit higher tolerance and a lot fewer accepted outliers for the test.

Aaron1011:
Even with a toleranece of 10, there are still 19654 outliers. I'd really like to get this merged, and work on improving the tests as a follow-up.

torokati44:
That's fine by me.

Lord-McSweeney (approved):

torokati44:
To anyone who will end up merging - please squash! :)

PR #14506, opened 2023-12-24
avm2: Improve Graphics.drawGraphicsData stubs

Aaron1011:
We now produce stubs or clearer errors for unsupported argument combinations, instead of unclear `RustError`s


core/src/avm2/globals/flash/display/graphics.rs
```diff
@@ -1393,27 +1393,31 @@ fn handle_graphics_triangle_path<'gc>(
...
 
-    let indices = obj
-        .get_public_property("indices", activation)?
-        .coerce_to_object(activation)?;
+    let indices = obj.get_public_property("indices", activation)?.as_object();
 
-    let _uvt_data = obj
-        .get_public_property("uvtData", activation)?
-        .coerce_to_object(activation)?;
+    let _uvt_data = obj.get_public_property("uvtData", activation)?.as_object();
```
torokati44, line 1400: I suppose the change from `coerce_to_object` to `as_object` is to avoid boxing primitives, which would not be valid here anyway?

core/src/avm2/globals/flash/display/graphics.rs
```diff
@@ -1462,7 +1466,12 @@ fn handle_graphics_triangle_path<'gc>(
                 draw_triangle_internal(activation, triangle, drawing, culling);
             }
         } else {
-            panic!("Indices is not a vector");
+            avm2_stub_method!(
+                activation,
+                "flash.display.Graphics",
+                "drawGraphicsData",
+                "with null indices"
+            );
```
torokati44, line 1474: Is this accepted in FP? If so, what happens then?

torokati44: Also, what if `vertices` is also null?

core/src/avm2/globals/flash/display/graphics.rs
Aaron1011: `coerce_as_object` will produce an error if it's `null`, but we want to handle the `null` case here

torokati44: Gotcha.

core/src/avm2/globals/flash/display/graphics.rs
Aaron1011: If `indices` is null, then `vertices` is interpreted as a list of x,y coordinate pairs. I don't think `vertices` can be null - I'll improve te error message in a follow-up PR

sleepycatcoding (approved):

adrian17 (approved):

PR #14526, opened 2023-12-26
avm2: Check the text.engine.FontDescription properties before setting

evilpie:


core/src/avm2/globals/flash/text/engine/FontDescription.as
```diff
@@ -79,5 +104,12 @@ package flash.text.engine {
             return false;
         }
 
+        internal static function throwNonNull(name: String) {
```
Lord-McSweeney, line 107: These can be `private static`

Lord-McSweeney (approved):

PR #14204, opened 2023-11-28
core: Improve spacing of device font letters (close #14185)

MrCheeze:
The calculation implemented here seems to give exactly the right results for device fonts, except for the degenerate case where font size=1, which I wasn't able to match perfectly.

PR #14535, opened 2023-12-27
Implement support for loading device fonts

evilpie:
This work is substantially based on @dinnerbone's https://github.com/Dinnerbone/ruffle/tree/fonts branch. 

However I modified this in some key areas:
- Font matching was moved completely to `load_device_font`
- Font searching was improved by supporting built-in names and having a list of fallbacks (mostly for my Linux system ;)
- The glyphs now use the proper scale
- Renamed `DirectFont` because I didn't like that name, I think `FontFace` is at least slightly better. (Suggestions welcome)


MrCheeze:
Really exciting stuff. Looks like vertical spacing is not correct for fonts parsed from ttf, though. ([swf](https://mrcheeze.github.io/homestuck-ruffle-tester/flash/00253_local.swf))
image

evilpie:
Thank you! I fixed it by changing the code for computing the `leading` value to dinnerbone's previous `face.line_gap()` instead of `ascender + descender`. I still don't fully understand that specific property.

Dinnerbone:
One thing we're going to need for this to land is tests. We'll need a new config in the `test.toml` which lists device fonts to load (and default_font implementers) - defaulting to what we have atm. Then we can add some regression tests for font rendering :D

MrCheeze:
Note that there's already [a dozen+ visual regression tests](https://github.com/ruffle-rs/ruffle/pull/14204/files#diff-95e4e52b5a426fd370fb75d74d86d334a681f2bf4e6d6801d4366e379d54d35d) that happen to try to render device fonts. Those should all switch from rendering with Noto Sans to rendering with the correct device font with this change, right? So that makes them bonus regression tests.

Dinnerbone:
That's a fair point.

We could maybe make the default device font nothing (disabling device font rendering entirely unless enabled) - and then set something for each of the now failing tests. It'd save a bunch of work and make those tests less prone to changing whenever we happen to change the default.

MrCheeze:
I always assumed it would be the opposite - no test configuration needed, just load any device fonts that the test swf asks for, same as opening it in desktop ruffle normally. Then you can just directly compare ruffle and FP's rendering, instead of having to deal with the extra complication of ruffle using Noto Sans instead of the correct font.

Dinnerbone:
We can't do that - the tests can't assume any environmental factor, and must run everywhere. Maybe you have Times New Roman and I have a slightly different version, or an entirely different replacement font - the test should work on both our machines.

tests/tests/swfs/visual/fonts/device-font/test.toml
```diff
@@ -0,0 +1,19 @@
+num_frames = 1
+
+[fonts.regular]
+family = "Tinos"
```
evilpie, line 4: We could use any other font, maybe with an even more permissive license. I was just looking at Tinos anyway.

desktop/src/backends/ui.rs
```diff
@@ -249,11 +256,71 @@ impl UiBackend for DesktopUiBackend {
...
-        _is_bold: bool,
-        _is_italic: bool,
-        _register: &mut dyn FnMut(FontDefinition),
+        name: &str,
+        is_bold: bool,
+        is_italic: bool,
+        register: &mut dyn FnMut(FontDefinition),
     ) {
+        let families = &match name {
+            "_serif" | "Times New Roman" | "" => vec![
```
Dinnerbone, line 265: _serif/etc should be resolved by ruffle before it reaches the backend - opinions on what they should be should be set through `Player::set_default_font`

Dinnerbone (changes requested):
We should remove the `default_font` feature from `desktop/Cargo.toml` as it no longer needs it

desktop/src/backends/ui.rs
evilpie: Good idea.

Dinnerbone (approved):
Thumbs up from me!

PR #5767, opened 2021-11-30
avm2: Mouse event support

kmeisthax:
This PR aims to support dispatching mouse events into AVM2, which will at least hypothetically allow basic games to work.

It changes a number of internal things relating to event dispatch, much like the previous interactive object PR. One thing to explain first is that there are actually four patterns of event dispatch across three different representations. Those look like the following:

| Pattern | Recipient | Example | `InteractiveObject` | AVM1 | AVM2 |
| - | - | - | - | - | - |
| Broadcast | All interested objects | `enterFrame`, `exitFrame`, `frameConstructed` | Certain `DisplayObject` methods | Global objects such as `Mouse`, `Key`, etc<br/><br/>`onXYZ` on interested objects | `addEventListener` on any object that exposes the broadcast |
| Anycast | First object to handle the event | `mouseMove`, `mouseDown`, `mouseUp` | `handle_clip_event` | `onXYZ` on interested objects | Not available |
| Targeted | A specific DisplayObject only | `press`, `release` | `handle_clip_event` | `onXYZ` on interested objects | `addEventListener` |
| Bubble/&ZeroWidthSpace;Capture | The ancestor of a targeted event recipient | N/A | Not available | Not available | `addEventListener` | 

Notably, AVM2 expects certain events to be targeted, while AVM1 wants them to be broadcasts or anycast. We're using the existing `InteractiveObject` and `ClipEvent` machinery to deliver events to AVM2, which means that we need three new event types:

 * `MouseWheel`. In AVM1 this is a broadcast, so there's no `ClipEvent` equivalent.
 * `MouseMoveInside`. The existing `MouseMove` event has anycast semantics, so we need a targeted equivalent.
 * `MouseUpInside`. The existing `MouseUp` event has anycast semantics, so we need a targeted equivalent.

The remaining event types already available had compatible dispatch semantics, so we're reusing them. All AVM2 event dispatch for clip events currently occurs in `InteractiveObject.event_dispatch_to_avm2` and all existing object types now punt to it if they hold a valid AVM2 object to dispatch on. The new event mappings are as follows:

| `ClipEvent` | AVM2 | Notes |
| - | - | - |
| `Press` | `mouseDown` | |
| `MouseUpInside` | `mouseUp` | |
| `Release` | `click`, `doubleClick` | Double-click tracking is currently handled by `InteractiveObject` |
| `ReleaseOutside` | `releaseOutside` | |
[...]

PR #14550, opened 2023-12-29
core: Fix `mouseWheel` event not being dispatched in AVM2

Lord-McSweeney:
I still need to add a test for this.

I'm not sure why, but the wheel event being dispatched is jittery for me. I'd appreciate manual testing with games that use the mouse wheel (e.g. Scratch 2). Additionally, kmeisthax's comment on #5767 seems relevant- at least I can reproduce it:
> Flash Player appears to report a delta of 3/-3 on my machine when Ruffle reports 1/-1

If anyone else can reproduce that Ruffle scrolls less than Flash, I can make the change to multiply the reported delta by 3.

### DIFF
--- a/core/src/player.rs
+++ b/core/src/player.rs
@@ -1235,11 +1235,14 @@ impl Player {
                         || !over_object.as_displayobject().avm1_removed()
                     {
                         over_object.handle_clip_event(context, ClipEvent::MouseWheel { delta });
+
+                        over_object
+                            .event_dispatch_to_avm2(context, ClipEvent::MouseWheel { delta });
                     }
                 } else {
                     context
                         .stage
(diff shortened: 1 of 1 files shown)


MrCheeze:
Flash docs say delta "depends on the device and operating system and is usually configurable by the user" - guessing it depends on this setting?
image

Lord-McSweeney:
Is that setting at 3 by default? I'm on Linux and I assumed FP kept consistent behavior throughout the Windows and Linux builds.

n0samu:
I've opened an issue for the delta problem: #14795. It's a preexisting issue so it shouldn't block this PR.

Dinnerbone:
(waiting for the mentioned test :D)

Lord-McSweeney:
Superseded by #17129

PR #14563, opened 2023-12-31
WIP/POC AVM2 verifier

Lord-McSweeney:
This is based off @adrian17's original work on his peephole optimizer branch (see #7920 and [a newer branch](https://github.com/adrian17/ruffle/compare/avm2-peephole-optimizer-poc-faster))

This results in noticeable performance increases on #10560 (6-7% decrease in startup time and even more, clearly noticeable improvements in gameplay) and in #12303.

This supersedes the main part of #14496, the opcode reading optimizations.

TODO:
- [ ] Something in try..catch blocks is broken (I keep losing working reproductions, but FP is throwing verify errors where Ruffle is not)
  - [ ] Also only verify exception target if there are any potentially error-throwing opcodes present in the from..to
  - [ ] All error-throwing opcodes in the from..to should act as a potential jump to the exception target (just like `iftrue`, `iffalse`, etc)- example: [verify.zip](https://github.com/ruffle-rs/ruffle/files/13801680/verify.zip)
- [ ] [More brokenness](https://github.com/ruffle-rs/ruffle/files/13801695/verify.zip)
- [ ] Add many tests
- [x] Store `parsed_code` on the method itself rather than in `AbcMethodBody`
- [ ] Add a custom Op type in AVM2 that stores already-resolved constant pool entries
- [ ] Storing an entire verified `AbcMethodBody` seems unnecessary, and is mostly leftover from Adrian's original work

This allows for further optimizations that actually modify the bytecode, such as #7920.


evilpie:
Nice. I imagine shrinking the Opcode size would improve this even more.

Lord-McSweeney:
Closing in favor of #14631.

PR #14568, opened 2024-01-01
desktop: Introduce generic gamepad remapping support

evilpie:
This is mostly a proof-of-concept, because I totally fail at playing Super Mario without a gamepad. We probably need to do some kind of per-game/swf remapping logic to actually make this useful, considering how peculiar some of those keybinds are.

### DIFF
--- a/core/src/player.rs
+++ b/core/src/player.rs
@@ -29,6 +29,7 @@ use crate::display_object::{
     EditText, InteractiveObject, Stage, StageAlign, StageDisplayState, StageScaleMode,
     TInteractiveObject, WindowMode,
 };
+use crate::events::GamepadButton;
 use crate::events::{ButtonKeyCode, ClipEvent, ClipEventResult, KeyCode, MouseButton, PlayerEvent};
 use crate::external::{ExternalInterface, ExternalInterfaceProvider, NullFsCommandProvider};
 use crate::external::{FsCommandProvider, Value as ExternalValue};
@@ -355,6 +356,9 @@ pub struct Player {
 
     /// Any compatibility rules to apply for this movie.
     compatibility_rules: CompatibilityRules,
+
+    /// A map from gamepad buttons to key codes.
+    gamepad_button_mapping: HashMap<GamepadButton, KeyCode>,
 
     /// Debug UI windows
     #[cfg(feature = "egui")]
@@ -842,6 +846,7 @@ impl Player {
     /// Event handling is a complicated affair, involving several different
     /// concerns that need to resolve with specific priority.
     ///
+    /// 0. Transform gamepad button events into key events.
     /// 1. (In `avm_debug` builds)
     ///    If Ctrl-Alt-V is pressed, dump all AVM1 variables in the player.
     ///    If Ctrl-Alt-D is pressed, toggle debug output for AVM1 and AVM2.
@@ -861,6 +866,33 @@ impl Player {
     /// 8. Mouse state is updated. This triggers button rollovers, which are a
     ///    second wave of event processing.
     pub fn handle_event(&mut self, event: PlayerEvent) {
+        // Optionally transform gamepad button events into key events.
+        let event = match event {
+            PlayerEvent::GamepadButtonDown { button } => {
+                if let Some(key_code) = self.gamepad_button_mapping.get(&button) {
+                    PlayerEvent::KeyDown {
+                        key_code: *key_code,
+                        key_char: None,
+                    }
+                } else {
+                    // Just ignore this event.
+                    return;
+                }
+            }
+            PlayerEvent::GamepadButtonUp { button } => {
+                if let Some(key_code) = self.gamepad_button_mapping.get(&button) {
+                    PlayerEvent::KeyUp {
+                        key_code: *key_code,
+                        key_char: None,
+                    }
+                } else {
+                    // Just ignore this event.
+                    return;
+                }
+            }
+            _ => event,
+        };
+
         let prev_is_mouse_down = self.input.is_mouse_down();
         self.input.handle_event(&event);
         let is_mouse_button_changed = self.input.is_mouse_down() != prev_is_mouse_down;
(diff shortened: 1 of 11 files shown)


GTcreyon:
Doesn't seem to work out of the box for me. I press every button on my gamepad, nothing happens. I'm on Arch Linux, LXQt.

evilpie:
I've updated my patch to support defining custom remappings via a command line parameter.

Super Mario 63:
`-B north=x -B south=up -B west=z -B east=c -B d-pad-left=left -B d-pad-right=right -B d-pad-up=up -B d-pad-down=down`

Phoenotopia:
`-B south=z -B west=x -B d-pad-left=left -B d-pad-right=right -B d-pad-up=up -B d-pad-down=down`

GTcreyon:
That's very nice. I've never found joypad mappings very comfortable for SM63 due to the merging of certain actions. Down being used both for diving and for general downward movement was always a bit awkward. But perhaps with a little mod to the game SWF, this could be made more comfortable?

What are the options for mapping? The examples you've listed have `space`, `left`, `right`, etc. Is there an exhaustive list to reference?

evilpie:
> What are the options for mapping? The examples you've listed have space, left, right, etc. Is there an exhaustive list to reference?

Everything in the `KeyCode` enum https://github.com/ruffle-rs/ruffle/blob/b55db7083cad9b7e1276c811ab2d6cc2e185306e/core/src/events.rs#L385 should work. So `;` should be `semicolon` and `'` should be `apostrophe`. 

This is one the main aspects that make this awkward to actually land. We would need some simpler way of find out the right names etc.

evilpie:
I made some effort to add a better help message for this feature, including a way of listing all valid buttons/keys. I think that should be good enough to land this. I am also going to look at writing a wiki page for this soon.

--help:
```
  -B, --gamepad-button <GAMEPAD BUTTON>=<KEY NAME>
          Remaps a specific button on a gamepad to a keyboard key: <gamepad-button>=<key-name>.
          This can be used to add new gamepad support to existing games, for example -B d-pad-up=up.
          
          A case-insensitive list of supported <gamepad-button>s is:
          - north, east, south, west
          - d-pad-up, d-pad-down, d-pad-left, d-pad-right
          - left-trigger, left-trigger2
          - right-trigger, right-trigger2
          - select, start
          
          A case-insensitive (non-exhaustive) list of common <key-name>s is:
          - a, b, c, ..., z
          - up, down, left, right
          - return
          - space
          - comma, semicolon
          - key0, key1, ..., key9
          The complete list of supported <key-name>s can be found by using -B start=nonexistent.
```

-B start=nonexistent
```
error: invalid value 'start=nonexistent' for '--gamepad-button <GAMEPAD BUTTON>=<KEY NAME>': Could not parse <key name>: invalid variant: nonexistent
  The possible values are: unknown, mouse-left, mouse-right, mouse-middle, backspace, tab, return, command, shift, control, alt, pause, caps-lock, numpad, escape, space, pg-up, pg-down, end, home, left, up, right, down, insert, delete, key0, key1, key2, key3, key4, key5, key6, key7, key8, key9, a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p, q, r, s, t, u, v, w, x, y, z, numpad0, numpad1, numpad2, numpad3, numpad4, numpad5, numpad6, numpad7, numpad8, numpad9, multiply, plus, numpad-enter, numpad-minus, numpad-period, numpad-slash, f1, f2, f3, f4, f5, f6, f7, f8, f9, f10, f11, f12, f13, f14, f15, f16, f17, f18, f19, f20, f21, f22, f23, f24, scroll-lock, semicolon, equals, comma, minus, period, slash, grave, l-bracket, backslash, r-bracket, apostrophe
```

desktop/src/app.rs
```diff
@@ -99,6 +101,8 @@ impl App {
             loaded = LoadingState::Loaded;
         }
 
+        let mut gilrs = Gilrs::new().ok();
```
adrian17, line 104: Anything worth logging here?
In particular, if user explicitly asks for gamepad settings but we don't manage to initialize?

core/src/player.rs
```diff
@@ -2507,6 +2542,7 @@ impl PlayerBuilder {
                 load_behavior: self.load_behavior,
                 spoofed_url: self.spoofed_url.clone(),
                 compatibility_rules: self.compatibility_rules.clone(),
+                gamepad_button_mapping: self.gamepad_button_mapping.clone(),
```
adrian17, line 2545: Since the function consumes `self`, is `clone` needed here?

desktop/src/app.rs
```diff
@@ -449,6 +453,46 @@ impl App {
...
+                        Button::Start => Some(GamepadButton::Start),
+                        Button::DPadUp => Some(GamepadButton::DPadUp),
+                        Button::DPadDown => Some(GamepadButton::DPadDown),
+                        Button::DPadLeft => Some(GamepadButton::DPadLeft),
+                        Button::DPadRight => Some(GamepadButton::DPadRight),
+                        _ => None,
+                    }
+                }
+
+                match event {
```
adrian17, line 477: how about writing it in the same way as above, with 2 matches?
The nested function wouldn't be needed then.

(I noticed that these two blocks of code are super similar, but are written in different styles - no strong preference between them though)

core/src/player.rs
```diff
@@ -356,6 +357,9 @@ pub struct Player {
     /// Any compatibility rules to apply for this movie.
     compatibility_rules: CompatibilityRules,
 
+    /// A map from gamepad buttons to key codes.
+    gamepad_button_mapping: HashMap<GamepadButton, KeyCode>,
```
adrian17, line 361: Not a major issue or anything blocking ;D Just noticed that since the key is an enum, this could also be Vec<Option> or a perfect hash.
Not required to change, just an observation.

adrian17 (commented):
Do you have any future plans for this?

desktop/src/app.rs
evilpie: I didn't really like the approach with two nested matches, one which requires unreachable!, so I would rather not do that.

Dinnerbone (approved):

PR #14598, opened 2024-01-03
avm2: Use correct null parameter error in `registerClassAlias`

Lord-McSweeney:


n0samu (approved):

Aaron1011 (approved):

core/src/avm2/globals/flash/net.as
```diff
@@ -10,9 +10,13 @@ package flash.net {
     public native function navigateToURL(request:URLRequest, window:String = null):void;
 
     public function registerClassAlias(name:String, object:Class):void {
```
torokati44, line 12: Just curious - can the difference in parameter names compared to Flash (see: https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/net/package.html#registerClassAlias() ) be observed in any way?

torokati44: Anyway, this is not in the scope of this PR, so merging, but I'm still hoping to get an answer sometime regardless.

torokati44 (approved):

core/src/avm2/globals/flash/net.as
torokati44: @Dinnerbone checked for me on Discord, the parameter names can't be observed from AS, so they don't matter at all technically, the mismatch (with the error messages) just looks a bit strange here in the code.

PR #14603, opened 2024-01-03
avm2: Reduce the size of Op on 64bit

evilpie:
This reduces the size of `Op` from 24 to 16 bytes, at the cost of `LookupSwitch` having a bit more overhead. This is useful for #14563, because it would reduce the total bytecode size by 1/3.

### DIFF
--- a/core/src/avm2/activation.rs
+++ b/core/src/avm2/activation.rs
@@ -1102,12 +1102,9 @@ impl<'a, 'gc> Activation<'a, 'gc> {
                 Op::TypeOf => self.op_type_of(),
                 Op::EscXAttr => self.op_esc_xattr(),
                 Op::EscXElem => self.op_esc_elem(),
-                Op::LookupSwitch {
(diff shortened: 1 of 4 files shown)


torokati44:
This might be unrelated, but I faintly remember seeing something like `#[repr(u8)]` in an earlier commit... What was that about?

evilpie:
> This might be unrelated, but I faintly remember seeing something like `#[repr(u8)]` in an earlier commit... What was that about?

That was actually `OpCode`. I am not sure if that helped at all, the idea was to optimize the opcode reading by having it be the same type.

torokati44:
Ah, right, this is about `Op`, not `OpCode`.

> by having it be the same type.

Same as ... what?

torokati44:
Okay, I see, thanks.

Aaron1011 (approved):

torokati44 (approved):

PR #14604, opened 2024-01-03
 avm2: Optimize op_get_property with array-like index access

evilpie:


core/src/avm2/activation.rs
```diff
@@ -1467,16 +1467,32 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
+                        if let Some(dictionary) = object.as_dictionary_object() {
+                            let _ = self.pop_stack();
+                            let _ = self.pop_stack();
+                            let value = dictionary.get_property_by_object(name_object);
+                            self.push_stack(value);
+
+                            return Ok(FrameControl::Continue);
+                        }
+                    }
+                    Value::Integer(name_int) if name_int >= 0 => {
```
torokati44, line 1484: nit: I'd put this branch first, with a comment explaining that this is just a fast-path, with identical behavior intended.

evilpie:
I changed the approach a bit to also cover vector and bytearrays. Plus an added tests to verify the out-of-bounds behavior continues to be correct.

Aaron1011 (approved):

JL2210:
From bisecting between master and nightly-2023-07-31 I somehow got that this PR caused a regression on Clicker Heroes:
```
532279b4e63768dfef95b0171510d9a03bdea66d is the first bad commit
commit 532279b4e63768dfef95b0171510d9a03bdea66d
Author: Tom Schuster <evilpies@gmail.com>
Date:   Wed Jan 3 22:17:11 2024 +0100

    avm2: Inline Avm2::peek

 core/src/avm2.rs | 20 +++++++++++++-------
 1 file changed, 13 insertions(+), 7 deletions(-)
```
with the above commit the game renders a white screen and freezes when trying to exit by e.g. pressing the X on the corner of the Ruffle window
Reverting it on master doesn't fix the problem, though. I'll bisect again, because there was another change that allowed the game to run with audio but not render.

PR #14620, opened 2024-01-05
avm2: Stub flash.globalization.DateTimeFormatter

### LINKED ISSUE #14577: error at https://sites.google.com/view/super-math-game/platform-racing-2?authuser=0
### Describe the bug

on platform racing 2 fails to load past the first screen

image_2024-01-01_135940070

### Expected behavior

image_2024-01-01_135908052

### Content Location

https://sites.google.com/view/super-math-game/platform-racing-2?authuser=0

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

_No response_

### Additional information

_No response_

n0samu:
Closes #14577


torokati44:
Should parameter null checking not be done in all methods?

n0samu:
Not important, but yes. I added it to the remaining methods.

torokati44 (approved):

core/src/avm2/globals/flash/globalization/DateTimeFormatter.as
```diff
@@ -0,0 +1,101 @@
+package flash.globalization {
+    import __ruffle__.stub_constructor;
+    import __ruffle__.stub_getter;
+    import __ruffle__.stub_method;
+    import flash.globalization.LastOperationStatus;
```
Lord-McSweeney, line 5: This import is unnecessary

torokati44: It's used on line 32.

torokati44: If anything, I was wondering whether the string literal default parameter values should be their corresponding named constants ("enums", like this one) instead. But the AS3 docs also say string literals there, so...

n0samu: I tried that already and the AS3 compiler doesn't allow it.

PR #14621, opened 2024-01-05
core: Fix non-ASCII characters input in EditText

kjarosh:
Casting the character to u8 and back to char caused some non-ASCII non-control characters to be treated as control characters. For instance the letter "ą" (U+0105) after casting to u8 and back became ENQ (U+0005) which is a control character.
Some other letters worked, for instance the letter "ł" (U+0142) became "B" (U+0042) and was not classified as a control character.

The test edittext_input was added to verify this behavior.


Lord-McSweeney (approved):
Normally we'd like manual testing of the regression tests to ensure that they do actually work, but I see that that's not possible for some of the characters in this one. Thank you for contributing!

kjarosh:
@Lord-McSweeney thank you for your review! Could you elaborate on what do you mean by manual testing? I'm just curious because I tried to comply with CONTRIBUTING.md and I did not see anything like that :smiley:

danielhjacobs:
I think @Lord-McSweeney just means ideally you would have typed this in Flash Player manually using the test swf, and then checked the trace output itself:

```abcABC~!@#$%^&*()`-=_+/?|ąęćłþó→↓ß©ęœπąśðæŋ’ə…łźćż„”ńµ≤≥≠²³¢€½§·«»–—°±¾≈∧‰¼£¿¡ƇƌƣƱǈǍǛǭǼȉȥȹɌɏḐṁṤẈẟẮốỳỿⱧⱾꞦꞀꞾꟿꭧ𐞔𐞮𝼙𝼞စ```

I don't know if you did or didn't do that though, as the files committed give no such evidence either way. I know doing so would be difficult on a regular keyboard, which is why it's probably fine regardless. I assume Lord-McSweeney was confused about the purpose of the Python file and thought it generated the output, but it seems to be from reading it it was just used to programmatically convert input.txt to input.json. Other input tests skip having the input.txt altogether, just having input.json, but I think it's fine to include the files that created that.

If I'm wrong about anything anyone can feel free to correct me.

danielhjacobs (approved):
Just ran this myself in debug Flash Player, copy-pasted the text from input.txt, clicked escape, and checked the log. It matched output.txt. I could try typing the characters individually but that seemed good enough.

kjarosh:
Ohh I understand now, I apologize for the confusion. `output.txt` is the actual output from the flash player, not copy-paste of `input.txt`. I did check inputting individual characters (the ones that were supported by my keyboard, and others using the Ctrl+Shift+U input method). However I realized that copy-pasting the input from input.txt is identical to inputting them manually, and that's why in `test.as` I listened for ESC instead of listening for a change of the text field, and I left input.txt for easy reproduction. New characters may be easily added to input.txt, input.json may be regenerated by the Python script, and output.txt may be generated by pasting input.txt to flashplayer.

Thanks for merging!

n0samu:
Thank you for contributing! ❤️

PR #14628, opened 2024-01-05
extension: Localize strings in `player.html`

Toad06:


Dinnerbone:
I'm a bit torn :D I like this and very much want this, but we've just put a lot of work into making ruffle.rs/demo the same code as this, with the intent to reuse the demo page as a library.

We could just take this and figure it out later, but the strings will have to move to their own package later and on crowdin that means deleting and making new - losing any progress between now and then.

I'm not sure :(

torokati44:
> We could just take this and figure it out later, but the strings will have to move to their own package later and on crowdin that means deleting and making new - losing any progress between now and then.

Sure, but we had a similar thing after the extension redesign - then I simply copied the translated strings over. I can do that again if we go this route. Not the most pleasant thing to do manually, but not a big deal either.

Toad06:
For now, I could remove the translations for strings that are only meant for debbuging ("Uncompressed Length", "SWF Version", etc.). This should make the transition to their own package less painful. What do you guys think?

Sorry for the late reply, by the way.

PR #11954, opened 2023-07-08
[WIP] avm2: Store resolved function param types

adrian17:
It's an ugly WIP-POC-prototype thing. `ResolvedParamConfig` feels silly, maybe new gc-arena's Cell-like apis would help clean it up.
(I also wish the signature stuff was unified/shared between native and bytecode method types.)
Now that `has_class_in_chain` PR is merged, this passes all tests c:

This removes all runtime parameter type name lookups from function calls. In my microbenchmarks, for `function f(a: int, b: int, c: int, d: int, e: int)`, this speeds up calls by ~30%.

The remaining overhead is
- mostly: `UpdateContext` cloning
- secondly: copying the arguments vec (twice!)
- other than that: general push/pop overhead

PR #14631, opened 2024-01-06
AVM2 verifier

Lord-McSweeney:
This is #14563 but with state merging removed, since I couldn't figure it out. It also includes some basic optimizations, and it keeps verification of local registers in (since that's not dependent on the control flow).

TODO:
- [x] <s>Something in try..catch blocks is broken (I keep losing working reproductions, but FP is throwing verify errors where Ruffle is not)</s> Not relevant anymore
  - [x] <s>Also only verify exception target if there are any potentially error-throwing opcodes present in the from..to</s> Fixed
  - [x] <s>All error-throwing opcodes in the from..to should act as a potential jump to the exception target (just like `iftrue`, `iffalse`, etc)- example: [verify.zip](https://github.com/ruffle-rs/ruffle/files/13801680/verify.zip)</s> Fixed, like above
- [x] <s>[More brokenness](https://github.com/ruffle-rs/ruffle/files/13801695/verify.zip)</s> Not relevant anymore
- [x] Add <s>many</s> two tests
- [x] Store `parsed_code` on the method itself rather than in `AbcMethodBody`
- [x] Add a custom Op type in AVM2 that stores already-resolved constant pool entries
- [x] Storing an entire verified `AbcMethodBody` seems unnecessary, and is mostly leftover from Adrian's original work

The optimization framework that this PR implements should work well with #7920.

Supersedes #14563.

Future optimizations:
- Remove `Nop` opcodes in code and modify jump offsets
- Record more local types from parameters (relies on #11954)
- Optimize `setproperty`/`initproperty` and `callproperty`/`callpropvoid`


swf/src/avm2/types.rs
```diff
@@ -574,6 +572,3 @@ pub enum Op {
     Timestamp,
     URShift,
 }
-
-#[cfg(target_pointer_width = "64")]
```
evilpie, line 578: Undo this, please. Add add something similar to `avm2::op::Op`.

Lord-McSweeney: I moved the check to `avm2::op::Op` because we now copy all `swf` `Op`s into `avm2` `Op`s, and those are the ones that `Activation` iterates over.

core/src/avm2/activation.rs
```diff
@@ -1125,39 +1086,23 @@ impl<'a, 'gc> Activation<'a, 'gc> {
                 Op::Sxi8 => self.op_sxi8(),
                 Op::Sxi16 => self.op_sxi16(),
                 Op::Throw => self.op_throw(),
-                _ => self.unknown_op(op),
+                _ => unimplemented!("Encountered unimplemented AVM2 opcode {:?}", op),
```
adrian17, line 1087: Proposal: make unimplemented dxns stuff Err at runtime like before; as for PushConstant(PushFloat, really), make it fail verification, just like it does in FP with the feature disabled. Alternatively, remove it entirely, so we throw `"Error #1011: Method contained illegal opcode."` like we already would for PushFloat4.

Lord-McSweeney: The second approach won't work because some SWFs mention `Dxns`(`Late`) despite not actually using it.

adrian17: By „remove it entirely”, I only meant the PushConstant one.

core/src/avm2/verify.rs
```diff
@@ -0,0 +1,1126 @@
...
+        }
+    } else {
+        None
+    };
+
+    // Initial set of local types
+    let mut local_types = vec![None; method.body().unwrap().num_locals as usize];
+    local_types[0] = this_class;
+
+    // Invalidate local types if they should be invalidated
```
evilpie, line 554: This comment is a bit recursive.

It seems like this code currently just "invalidates" the remembered type of |this|, if that local is every modified. This approach doesn't really work for anything else.

core/src/avm2/verify.rs
```diff
@@ -0,0 +1,1126 @@
...
+                    local_types[*index_register as usize] = None;
+                }
+            }
+            _ => {}
+        }
+    }
+
+    let mut previous_op = None;
+    for (i, op) in code.iter_mut().enumerate() {
+        if let Some(previous_op_some) = previous_op {
```
evilpie, line 584: Can we unnest this a bit? (And definitely rename `previous_op_some`)

e.g. 
```rust
let previous_op = if let Some(previous_op) = previous_op {
  previous_op
} else {
  previous_op = Some(op.clone());
}
```

In theory we could even do something like
```rust
previous_op = code.first();
for (...) in code.iter_mut().enumerate().skip(1) {
```

core/src/avm2/verify.rs
```diff
@@ -0,0 +1,1126 @@
...
+
+                        let resolved_type = activation
+                            .domain()
+                            .get_class(&multiname, activation.context.gc_context)
+                            .unwrap();
+
+                        if let Some(class) = resolved_type {
+                            match previous_op_some {
+                                Op::PushNull => {
+                                    if !GcCell::ptr_eq(
```
evilpie, line 745: It would be good to explain this optimization.

core/src/avm2/verify.rs
```diff
@@ -0,0 +1,1126 @@
...
+                        | Op::GreaterEquals
+                        | Op::GreaterThan
+                        | Op::LessEquals
+                        | Op::LessThan
+                        | Op::PushTrue
+                        | Op::PushFalse
+                        | Op::Not
+                        | Op::IsType { .. }
+                        | Op::IsTypeLate => {
+                            previous_op = Some(op.clone());
```
evilpie, line 599: We could write this as `previous_op = mem::replace(op, Op::Nop)`, but I am not sure if that is really worth it.

core/src/avm2/verify.rs
```diff
@@ -0,0 +1,1126 @@
...
+                    _ => {}
+                }
+            }
+        }
+
+        previous_op = Some(op.clone());
+    }
+}
+
+fn ops_can_throw_error(ops: &[AbcOp], start_idx: u32, end_idx: u32) -> bool {
```
evilpie, line 795: This could take just a slice, without indexes.

core/src/avm2/verify.rs
Lord-McSweeney: I plan to add parameter types here in a later PR (though that depends on #11954).

core/src/avm2/verify.rs
Lord-McSweeney: I can't do `let Some(previous_op) = previous_op` because the optimizer needs to set the `Option` value of `previous_op`. I'll try unnesting it.

core/src/avm2/verify.rs
Lord-McSweeney: I think this current implementation is more readable.

core/src/avm2/verify.rs
adrian17: Either way, how about documenting the purpose? Something like `A conservative logic to only allow type-based optimizations we're 100% sure about - only trust a parameter's type if its variable was never written to. We invalidate stored types if we encounter a register-setting opcode.`

core/src/avm2/verify.rs
Lord-McSweeney: I tried unnesting it but I realized that there are likely going to be optimizations that we can do that would work despite `previous_op` being `None`, and/or those that would work even when the next op is on a jump target.

core/src/avm2/verify.rs
```diff
@@ -0,0 +1,1193 @@
...
+            | AbcOp::PopScope
+            | AbcOp::ReturnVoid => {}
+            _ => return true,
+        }
+    }
+
+    false
+}
+
+fn pool_int<'gc>(
```
adrian17, line 903: can we move these to `script.rs`? It already has a bunch of similar functions, so it'd be more consistent (and decrease verify.rs size a bit too).
Also, any chance this could be unified with `abc_int()` (and friends)? As they are pretty much duplicates, sans `0` treatment. (this'd also probably make function with invalid default values throw a correct VerifyError "for free"). Can be left for a follow-up though.

core/src/avm2/verify.rs
```diff
@@ -0,0 +1,1193 @@
...
+                        *offset,
+                        true,
+                        idx_to_byte_offset,
+                        byte_offset_to_idx,
+                    )?;
+
+                    if op_idx != i {
+                        // Replace last entry on worklist with current i (update i)
+                        worklist.pop();
+                        worklist.push(i + 1);
```
adrian17, line 350: I don't really understand what this (and `save_current_work` are for). Is there anything that makes the naive impl, like
```rs
    while let Some(mut i) = worklist.pop() {
        // no other popping inside
```
not work?

core/src/avm2/verify.rs
```diff
@@ -0,0 +1,1193 @@
...
+                        _ => {}
+                    },
+                    Op::CoerceD => match previous_op_some {
+                        Op::CoerceD
+                        | Op::PushDouble { .. }
+                        | Op::Multiply
+                        | Op::Divide
+                        | Op::Modulo
+                        | Op::Increment
+                        | Op::IncLocal { .. }
```
adrian17, line 627: IncLocal family isn't correct here, right? As it doesn't push any value

core/src/avm2/verify.rs
```diff
@@ -0,0 +1,1193 @@
...
+                    } => {
+                        let multiname = method
+                            .translation_unit()
+                            .pool_maybe_uninitialized_multiname(
+                                *name_index,
+                                &mut activation.context,
+                            );
+
+                        let resolved_type = if let Ok(multiname) = multiname {
+                            if !multiname.has_lazy_component() {
```
adrian17, line 728: I think a failed/lazy multiname should throw a VerifyError at an earlier stage, right?
(can be a follow-up)

core/src/avm2/verify.rs
Lord-McSweeney: This is mostly because on the conditional jumps, we need to push `i` back on the worklist.

core/src/avm2/verify.rs
Lord-McSweeney: I'm going to leave this for a follow-up to also fix more issues with namespace and string pooling

core/src/avm2/verify.rs
```diff
@@ -0,0 +1,1149 @@
...
+
+    // Handle exceptions
+    let mut new_exceptions = Vec::new();
+    for exception in body.exceptions.iter() {
+        // FIXME: This is actually wrong, we should be using the byte offsets, not the opcode offsets.
+        // avmplus allows for from/to (but not targets) that aren't on a opcode, and some obfuscated
+        // SWFs have them. FFDEC handles them correctly, stepping forward byte-by-byte until it
+        // reaches the next opcode. This does the same (stepping byte-by-byte), but it would be better
+        // to directly use the byte offsets when handling exceptions.
+        let mut from_offset = byte_offset_to_idx
```
Aaron1011, line 119: I think you could just initialize this to `None`, and let the `while` loop handle the initialiation

core/src/avm2/verify.rs
```diff
@@ -0,0 +1,1149 @@
...
+                .get(&((exception.from_offset + offs) as i32))
+                .copied();
+            offs += 1;
+            if offs as usize >= new_code.len() {
+                return Err(make_error_1054(activation));
+            }
+        }
+
+        // Now for the `to_offset`.
+        let mut to_offset = byte_offset_to_idx
```
Aaron1011, line 135: Same here

core/src/avm2/verify.rs
```diff
@@ -0,0 +1,1149 @@
...
+            .get(&(exception.to_offset as i32))
+            .copied();
+
+        let mut offs = 0;
+        while to_offset.is_none() {
+            to_offset = byte_offset_to_idx
+                .get(&((exception.to_offset + offs) as i32))
+                .copied();
+            if offs == 0 {
```
Aaron1011, line 144: `offs` is initialized to 0, so this branch will be taken on the very first iteration through the loop.

Aaron1011 (approved):

torokati44:
I have rebased this on top of `master`, fixed a few typos in the commit messages, and melted the changes in the `cleanup rebase fallout` and `chore: fmt + appease clippy` commits into their appropriate predecessors.
I could force-push it to this PR branch, but rather not without your informed consent. Feel free to take it over if you are fine with it, and I haven't broken anything: https://github.com/torokati44/ruffle/commits/avm2-verifier-no-merge/

torokati44 (changes requested):
See my comment above, I'm just leaving this review to draw attention to it.

Aaron1011 (approved):

Issue #280, opened 2020-01-11
Actionscript Tracking: TextField


Dinnerbone:
This is the tracking issue for **TextField** in AVM1. We will check off each item of progress as appropriate, and fill in any relevant or missing information as we continue development of Ruffle.
# Legend
Each checkbox is independent of another. It's entirely possible for something to be tested but not exist yet, or for us to believe that it's completely functional but we haven't made enough tests to prove it.
## "Property Exists"
This means the item exists, but may not necessarily be fully implemented.
## "Has Test Coverage"
This means that we have believe that we have a good test coverage of this item, regardless of if those tests pass. It's okay to have tests available but not implement the item yet.
## "Completely Functional"
This means we believe that the item is completely implemented, and no more work needs to be done towards making it functional.

# TextField
### `_listeners : Array`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional

### `getFontList()`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `broadcastMessage()`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional

### `_alpha : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_height : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_highquality : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_name : String`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_parent : MovieClip`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_quality : String`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_rotation : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_soundbuftime : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_target : String`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_url : String`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_visible : Boolean`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
[...]

PR #14634, opened 2024-01-06
avm1,avm2: Implement TextField.restrict

kjarosh:
I implemented `TextField.restrict` for both AVM1 and AVM2 (with minor differences). I also added extensive testing to make sure how the restrict actually works, because the documentation is very vague (and wrong in one case).

Currently restrict in Ruffle works only when inputting characters one by one, not when pasting a string (flashplayer does not differentiate between inputting and pasting characters in this case). ~I'm working on a second PR which will add support for restrict when pasting (it requires more complex testing).~ Support for restrict when pasting is included.

I guess it's a step towards completing #280.

If I see correctly it's also related to #12613.


core/src/avm1/globals/text_field.rs
```diff
@@ -859,3 +860,36 @@ fn set_filters<'gc>(
...
+) -> Result<(), Error<'gc>> {
+    match value {
+        Value::Undefined | Value::Null => {
+            this.set_restrict(None, &mut activation.context);
+        }
+        _ => {
+            let text = value.coerce_to_string(activation)?;
+            if text.is_empty() {
+                // According to docs, an empty string means that you cannot enter any character,
+                // but according to reality, an empty string is equivalent to null in AVM1.
```
torokati44, line 887: This is something that could be added to the spec errata in the Wiki perhaps.
We find lots of stuff like this all the time, and don't put half of it in there, but still...

core/src/avm2/globals/flash/text/text_field.rs
```diff
@@ -1325,18 +1325,41 @@ pub fn set_mouse_wheel_enabled<'gc>(
...
-    _args: &[Value<'gc>],
+    this: Object<'gc>,
+    args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
-    avm2_stub_setter!(activation, "flash.text.TextField", "restrict");
+    if let Some(this) = this
+        .as_display_object()
+        .and_then(|this| this.as_edit_text())
+    {
+        let value = args.get(0).cloned().unwrap_or(Value::Undefined);
```
torokati44, line 1353: We have `get_object` and `try_get_object` in `ParametersExt`, perhaps that could be used here?

core/src/display_object/edit_text.rs
```diff
@@ -1476,36 +1491,39 @@ impl<'gc> EditText<'gc> {
...
-                        &mut activation,
-                        "textInput",
-                        character_string,
-                        true,
-                        true,
-                    );
-                    Avm2::dispatch_event(&mut activation.context, text_evt, target);
+                let allowed = self.0.read().restrict.to_allowed(character);
+                if let Some(character) = allowed {
+                    if let Avm2Value::Object(target) = self.object2() {
```
torokati44, line 1496: Can't these two `if let`s be merged somehow? Not sure, nor if that would be more readable, it's just that things are getting more and more indenty here.

tests/tests/swfs/avm2/edittext_restrict/Test.as
```diff
@@ -0,0 +1,100 @@
...
+            trace("====================");
+            text.text = "";
+            currentRestrict += 1;
+            if (restricts.length <= currentRestrict) {
+                trace("No more restricts");
+                return;
+            }
+            text.restrict = restricts[currentRestrict];
+            trace("Restict set: '" + restricts[currentRestrict] + "'");
+            trace("Restict get: '" + text.restrict + "'");
```
torokati44, line 97: typo: "Restrict"

tests/tests/swfs/avm1/edittext_restrict/test.as
```diff
@@ -0,0 +1,93 @@
...
+    trace("====================");
+    text.text = "";
+    currentRestrict += 1;
+    if (restricts.length <= currentRestrict) {
+        trace("No more restricts");
+        return;
+    }
+    text.restrict = restricts[currentRestrict];
+    trace("Restict set: '" + restricts[currentRestrict] + "'");
+    trace("Restict get: '" + text.restrict + "'");
```
torokati44, line 82: typo: "Restrict"

torokati44:
This does a fair bit of stuff, including extending our testing capabilities, adding tests, adding a new type, implementing both avm1 and avm2 properties - all in a single commit.

Maybe it would be better to split it into a couple logically separate commits that build on each other...?

core/src/avm2/globals/flash/text/text_field.rs
kjarosh: I replaced it with `try_get_string`, should look better now!

core/src/display_object/edit_text.rs
kjarosh: Things were very indenty here, I managed to reduce the indent from 4 to mostly 1.

tests/tests/swfs/avm2/edittext_restrict/Test.as
kjarosh: Fixed, added some more test cases by the way

tests/tests/swfs/avm1/edittext_restrict/test.as
kjarosh: Fixed as well

core/src/avm2/globals/flash/text/text_field.rs
```diff
@@ -1325,18 +1325,39 @@ pub fn set_mouse_wheel_enabled<'gc>(
...
+    if let Some(this) = this
+        .as_display_object()
+        .and_then(|this| this.as_edit_text())
+    {
+        match args.try_get_string(activation, 0)? {
+            None => {
+                this.set_restrict(None, &mut activation.context);
+            }
+            Some(text) => {
+                this.set_restrict(Some(&text), &mut activation.context);
```
torokati44, line 1358: Could this be `Option::map`?

core/src/avm1/globals/text_field.rs
kjarosh: Added this information, hope it's okay: https://github.com/ruffle-rs/ruffle/wiki/SWF-AVM-Specification-Errata

core/src/avm2/globals/flash/text/text_field.rs
kjarosh: Yeah... I just replaced the expression without thinking. Should be good now!

Dinnerbone (approved):
Thanks so much, especially for the tests!

core/src/avm1/globals/text_field.rs
torokati44: It's perfect, thank you!

core/src/display_object/edit_text.rs
```diff
@@ -2312,3 +2328,210 @@ impl TextSelection {
...
+                            allowed.append(&mut current_intervals);
+                        }
+                    } else {
+                        disallowed.append(&mut current_intervals);
+                    }
+
+                    // Caret according to the documentation indicates
+                    // that we are now disallowing characters.
+                    // In reality it just switches allowing/disallowing.
+                    now_allowing = !now_allowing;
```
torokati44, line 2457: lol, nice misdocumentation, and great that you discovered it
(just babbling, no action necessary)

torokati44 (approved):

core/src/display_object/edit_text.rs
```diff
@@ -2312,3 +2328,210 @@ impl TextSelection {
...
+            && !self.intervals_contain(character, &self.disallowed)
+    }
+
+    fn intervals_contain(&self, character: char, intervals: &Vec<(char, char)>) -> bool {
+        for interval in intervals {
+            if self.interval_contains(character, interval) {
+                return true;
+            }
+        }
+        false
```
torokati44, line 2508: I guess this could have been a call to `any`, but this isn't any less readable, and no need to delay further just for this.

core/src/display_object/edit_text.rs
kjarosh: TBH it looks like an implementation detail leak. Someone at Macromedia did not think more than one caret may be included in restrict :smiley:

torokati44: Yep. And we have seen _a lot_ of these kinds of leaks.

torokati44: So, thanks, Hyrum's Law...
https://www.hyrumslaw.com/

PR #14635, opened 2024-01-06
extension: Close `popup.html` when clicking "Open SWF Player" in Firefox

Toad06:
It appears that the promise in the following code was never fulfilled, so `window.close()` was never called:
https://github.com/ruffle-rs/ruffle/blob/f0ed5c7211c22eda87425456e04e4d665068fa21/web/packages/extension/src/popup.ts#L176-L179

The same issue also occurred in Chrome, but it wasn't noticeable as this browser seems to close the popup by itself after calling `chrome.tabs.create`.

### DIFF
--- a/web/packages/extension/src/utils.ts
+++ b/web/packages/extension/src/utils.ts
@@ -131,8 +131,8 @@ if (typeof chrome !== "undefined") {
             chrome.tabs.create({ url: "/options.html" }, cb),
         );
     openPlayerPage = () =>
-        promisify((_cb: () => void) =>
-            chrome.tabs.create({ url: "/player.html" }),
+        promisify((cb: () => void) =>
+            chrome.tabs.create({ url: "/player.html" }, cb),
         );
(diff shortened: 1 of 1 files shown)


web/packages/extension/src/utils.ts
```diff
@@ -131,8 +131,8 @@ if (typeof chrome !== "undefined") {
             chrome.tabs.create({ url: "/options.html" }, cb),
         );
     openPlayerPage = () =>
-        promisify((_cb: () => void) =>
-            chrome.tabs.create({ url: "/player.html" }),
+        promisify((cb: () => void) =>
+            chrome.tabs.create({ url: "/player.html" }, cb),
         );
 } else if (typeof browser !== "undefined") {
```
evilpie, line 137: Drive-by: This `if` probably should be first, because Firefox supports both `chrome` and `browser`.

n0samu (approved):

Dinnerbone (approved):

PR #14643, opened 2024-01-06
Apply 'xml["@myattr"]' logic when setting/deleting properties

Aaron1011:
We previously only applied this when getting properties.


core/src/avm2/object/xml_object.rs
```diff
@@ -643,3 +635,23 @@ impl<'gc> TObject<'gc> for XmlObject<'gc> {
...
     }
 }
+
+fn handle_input_multiname<'gc>(
+    name: Multiname<'gc>,
+    activation: &mut Activation<'_, 'gc>,
+) -> Result<Multiname<'gc>, Error<'gc>> {
+    // Special case to handle code like: xml["@attr"]
+    // FIXME: Figure out the exact semantics.
+    let multiname = if !name.has_explicit_namespace()
```
evilpie, line 645: This logic can be simplified to an early return.

tests/tests/swfs/avm2/xml_attribute_name/Test.as
```diff
@@ -22,6 +22,29 @@ package {
             trace("xml[name2]: " + xml[name2]);
             trace('xml["bar"]: ' + xml["bar"]);
             trace('xml["@bar"]: ' + xml["@bar"]);
+			
+			// Test creating attribute
+			xml["@foo"] = "My new attr";
```
evilpie, line 27: Different indentation.

core/src/avm2/object/xml_object.rs
```diff
@@ -261,6 +261,8 @@ impl<'gc> TObject<'gc> for XmlObject<'gc> {
         // FIXME - implement everything from E4X spec (XMLObject::getMultinameProperty in avmplus)
         let read = self.0.read();
 
+        let name = handle_input_multiname(name.clone(), activation);
```
evilpie, line 264: We either need to move this under the index check, or change the check. Because otherwise `xml["@0"]` behaves like `xml["0"]`.

Aaron1011: I moved the check

evilpie (approved):

PR #14647, opened 2024-01-07
Progressive download of video files

kmeisthax:
Adds the ability to stream fetched data instead of an all-at-once download. Also adds the ability to get estimated download size. Most of `NetStream` was already built in anticipation of streaming download, so we're applying these new backend capabilities to it first.

This also reserves buffer capacity and sets `bytesTotal` based on the estimated file size length.

This does *not* add SWF progressive download capability. While we already have well tested implementations of *simulated* streaming download (specifically to hide shape processing latency), we need streaming-aware zlib/LZMA decompressors before we can actually download SWFs in chunks.

~~This PR is a draft because I haven't actually tested it in a browser yet.~~ I have been able to successfully do streaming playback on at least one video player (specifically old YouTube via Internet Archive).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4576,6 +4576,7 @@ dependencies = [
  "base64",
  "chrono",
  "console_error_panic_hook",
+ "futures",
  "futures-util",
  "generational-arena",
  "getrandom",
@@ -4599,6 +4600,7 @@ dependencies = [
  "url",
  "wasm-bindgen",
  "wasm-bindgen-futures",
+ "wasm-streams",
  "web-sys",
 ]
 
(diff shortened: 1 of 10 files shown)


torokati44:
_Ah, just when I was happy about us bot streaming videos, making adding MP4 (F4V) support a lot easier..._ 😛
But don't worry, surely I'll be able to shoehorn it in even with streaming.

torokati44:
Haven't had time yet to look at this closer, just curious, as to what should I expect - does this add seek-driven-chunk-downloads?

Also, there are some merge conflicts already... 😩

kmeisthax:
We don't have seek-driven downloads yet, this is just playing data that gets downloaded before the video finishes playing. I still have to figure out how the hell time-based seeks would work on an FLV...

Aaron1011 (approved):

sergo-zu:
Hello, guys! We all need this problem get successfully solved...

Issue #723, opened 2020-06-16
Tracking issue for video decoding


Dinnerbone:
We've had a number of people ask about this, and whilst we may not solve it soon - we can use this to let them be updated when we make progress.

There's 4 codecs (that I know of) that we'll need to support inside `DefineVideoStream`:

- [ ] Sorenson H.263 (codec 2)
- [ ] Screen Video (codec 3)
- [ ] VP6 (codec 4)
- [ ] VP6 with Alpha (codec 5)

I can also see that H.264 is required according to https://helpx.adobe.com/flash/kb/supported-codecs-flash-player.html

The biggest limitations will be that we need decoders that satisfy the following requirements:
- Must be usable on desktop on Mac, Windows and Linux
- Must be usable on browser either as built-in to the browser, or through Wasm
- Must be distributable legally
  - Video codecs and patents or licensing terms unfortunately go hand-in-hand
  - Asking end-users to install something would be a no-go for web

Issue #590 contains a relevant discussion on using ffmpeg to solve this.

Issue will be updated with more info as we progress.

ROBERT-MCDOWELL:
about h264 dec and enc, I think the best choice would be ffmpeg lib converted in wasm or
just https://github.com/cisco/openh264
already used by chrome in PPAPI and firefox in NPAPI

EZWrighter:
Adding a note here, that RTMP streaming is still used often in the browser for close to real time video delivery and flash is the only solution.  This would be a huge feature if we could still stream RTMP after flash is no longer officially supported in browsers.

ROBERT-MCDOWELL:
it looks like ffmpeg in wasm is more and more efficient and workers ready, which can open the door to a vast encoding decoding, including all Flash codecs of course
https://github.com/ffmpegwasm/ffmpeg.wasm

EZWrighter:
I actually looked at this problem and I believe socket security is a huge issue for the streaming libraries, so I am guessing RTMP Streaming will never be possible via wasm :-(

ROBERT-MCDOWELL:
this is not correct, Rust has already an RTMP library. 
https://users.rust-lang.org/t/rtmp-live-streaming-in-rust/16869
alos, with QUIC, future HTTP3, TCP and UDP low level protocol is available.

EZWrighter:
[...]

PR #14654, opened 2024-01-08
Add H.264 decoding by utilizing OpenH264 from Cisco

torokati44:
~~This is very much a WIP.~~ Progresses #723. ~~Basically pushing only for visibility.~~ Will partially replace #12539 (first only on desktop, in the future, using WebCodecs, entirely).

AFAIK this is the only legal way (without us or our users having to pay royalties to ~~MPEG LA~~ _VIA LA_) to do this with reasonable portability on desktop, but IANAL. See this little explainer video: https://vimeo.com/79578794

The code is somewhat based on the `openh264-rs` crate, see https://github.com/ralfbiedert/openh264-rs/issues/43.

YUV->RGB colorspace conversion is done by our trusty handcrafted code, like for H.263 and VP6[A]. This means that it will always assume BT.601 coefficients, even though it could _theoretically, possibly_ be BT.709 or something like that. But the difference is probably subtle enough anyway.

TODOs, in no particular order:
 - [x] Add a `configure_decoder` (or similar) method to the video backend/decoder traits. ~~Currently the preload method is misused to pass the "extradata" to the decoder.~~
   - Done.
 - [x] Figure out how to not create colliding `Index`es for proxied and handled videos in the ExternalBackend.
   - Added an enum to store in the new backend's arena. TODO: Add a warning comment.
 - [x] Optimize copying YUV data out, and possibly AVCC->AnnexB format conversion
 - [x] Handle downloading and hash-checking of the decoder shared libraries for the current platform/arch.
   - Done. TODO: Needs to do error handling better for proper fallback and hints for the user.
 - [x] Add the [Cisco license text](https://www.openh264.org/BINARY_LICENSE.txt), display it somewhere in the player.
   - Done. The new backend will write it into the current working directory as `OpenH264-license.txt` upon creation.
 - [x] Add a command line option to disable the decoder (as per Cisco's license terms, I hope this will be enough)
   - Done: `--enable-openh264 [true|false]`
 - [x] Switch to preferences for the enablement toggle
 - [x] Figure out how, or if, to test this (command line flag for implicit agreement to fetch the decoder library on every run?)
   - Download on first run without confirmation, even in CI.
   - For tests, this should be opt-in, just like "video" and "renderer" in general.
   - Command line flag will be only for opting out.
 - [x] Write up a method for updating OpenH264 versions in the future.
 - [ ] Check if undoing the "start code emulation prevention" for the decoder config (extradata) is needed, and undo it if so.
   - Left this as a TODO comment for now...
 - [x] Look at the presentation time offsets as provided by the FLV demuxer (later the MP4 container), and the fields in the `SBufferInfo` struct from the decoder.
 - [x] General code cleanup, including properly working features to disable this, comments, and error handling.
   - Almost done!
 - [x] Testing, testing, testing...
   - There is a visual test added, which is something...

TODOs for later:
 - WebCodecs implementation for web. Consider whether MSE plus a hidden `<video>` element would be a feasible fallback - probably not, but who knows.
   I plan to put this into the same `ExternalVideoBackend` added here. Also with the "other ideas" below. Spares us from duplicating the proxying logic. The Android backend will have to be different though, as it's somewhere else.

Somewhat less closely related TODOs for later:
 - MP4 (F4V) support: https://github.com/ruffle-rs/ruffle/pull/14655
 - AAC-LC decoding support (already available in Symphonia).

Other ideas:
 - Different, specific implementations for different platforms: Video Toolbox on macOS, Microsoft Media Foundation or DXVA[2] on Windows, VA-API (or VDPAU...) on Linux, and MediaCodec on Android. Maybe Vulkan Video anywhere applicable. (Loading FFmpeg or GStreamer on Linux would be iffy due to uncertain licensing of the used decoder.)


torokati44:
This is what I use to test this at the moment:
[f4vplayer_h264_flv.zip](https://github.com/ruffle-rs/ruffle/files/13855472/f4vplayer_h264_flv.zip)

I hope it's okay to upload here. Despite the name, it now loads an FLV file. It's from here: https://permadi.com/2010/02/flash-playing-flvf4v-videos-with-script/

danielhjacobs:
> Will partially replace #12539 (first only on desktop, in the future, using WebCodecs, entirely).

The [VideoDecoder](https://developer.mozilla.org/en-US/docs/Web/API/VideoDecoder) part of WebCodecs (and many other APIs) are only available in secure contexts, right (they're also not yet available in Firefox but I'm sure that will change)? Will this be the case in insecure contexts too? I suppose browsers are moving away from allowing insecure sites at all.

torokati44:
>  not yet available in Firefox but I'm sure that will change)?

Certainly, since it already works if enabled in `about:config`, it's just not yet stable/complete.

> The [VideoDecoder](https://developer.mozilla.org/en-US/docs/Web/API/VideoDecoder) part of WebCodecs (and many other APIs) are only available in secure contexts, right

I have no idea. Anyway, we can keep the current solution if VideoDecoder is not available for whatever reason.

torokati44:
I have decided that there is (currently) no need for any "internal" features such as `h263`, `vp6`, etc, within `software_video`, as this backend is currently only handling H.264 videos, on desktop. So, the single `external_video` added feature should suffice.
Later, when a WebCodecs backend is added, that can be selected automatically, based on the target platform.

Also, the OpenH264 decoder currently drops the shared library, and the corresponding license, into the current working directory. I hope this satisfies the licensing terms about reproducing the text itself.
The library is MD5 checksum checked if exist, but not repaired or replaced if it's not correct, it's simply not used.
The license file is not repaired either, but it's also not even checked for contents if it exists.

I also thought about renaming the CLI argument to `--enable-external-video` instead, but keeping the required licensing notice. Any opinions on this? It would break the abstraction of which codecs are in which backends a little less.

torokati44:
Also at the moment, the new, external video backend, when enabled as a feature, requires the existing software backend to be enabled too. There is no strict technical reason for this, it was just easier to do the fallback this way. For an MVP I think it's fine.

torokati44:
Some more notes:
 - There's still an assert or two left in there, I don't like that. No reason to panic when the decoder could also just stop working.
 - Perhaps a proper `H264Error` enum would be nice, either only for decode-time errors, or maybe also for library-acquisition-time ones.
 - When the NALU length is 3 or 4 bytes, the variant conversion could simple overwrite that with the start code. Even with 1 or 2 byte lengths, an in-vector shift would probably be better, as it can probably avoid a reallocation if there's spare capacity, plus it's nicer for cache locality.
 - I ended up reusing the `VideoDecoder` trait from the software backend directly, instead of duplicating it just to add the configuration method. But then that call is still cut short right away in the software backend. Now it could be passed on to the decoders, where most of which will do nothing anyway. Just less abstraction breaking...
 - What to so with the "stream end marker" packets? Do they matter? Maybe those too can be piped directly into the decoder?

core/src/streams.rs
```diff
@@ -968,11 +968,43 @@ impl<'gc> NetStream<'gc> {
...
+                    }
+                }
             }
-            (_, _, FlvVideoPacket::AvcNalu { .. }) => {
-                tracing::warn!("Stub: FLV AVC/H.264 NALU processing")
+            (
+                Some(video_handle),
+                Some(codec),
+                FlvVideoPacket::AvcNalu {
+                    composition_time_offset: _,
```
kmeisthax, line 986: I'm going to assume this means Flash didn't actually use the composition time offset in FLV H.264 streams either... AFAIK that sounds plausible to me (FLV is already variable framerate) but I wanted to just double-check.

desktop/src/cli.rs
```diff
@@ -164,6 +164,12 @@ pub struct Opt {
     #[clap(long)]
     pub dummy_external_interface: bool,
 
+    /// Whether to allow downloading and using an external H.264 decoder binary.
+    ///
+    /// OpenH264 Video Codec provided by Cisco Systems, Inc.
+    #[clap(long, default_value = "true")]
+    pub enable_openh264: Option<bool>, // TODO: rename to "enable_external_video_backend" or something like that...?
```
kmeisthax, line 171: Regarding the TODO: we're probably going to have other video backends eventually, at the very least for supporting hardware decoding of H.263 or H.264 streams. "External video backend" wouldn't be descriptive enough in that case.

video/external/src/BINARY_LICENSE.txt
```diff
@@ -0,0 +1,59 @@
...
+
+---------------------------------------------
+AVC/H.264 Patent Portfolio License Conditions
+---------------------------------------------
+
+In addition, the Cisco-provided binary of this Software is licensed under Cisco's license from MPEG LA only if the following conditions are met:
+
+1. The Cisco-provided binary is separately downloaded to an end user’s device, and not integrated into or combined with third party software prior to being downloaded to the end user’s device;
+
+2. The end user must have the ability to control (e.g., to enable, disable, or re-enable) the use of the Cisco-provided binary;
```
kmeisthax, line 49: While a command line flag is probably sufficient to comply with the license, we probably should have a menu item to turn OpenH264 on or off.

video/external/src/decoder/README.md
```diff
@@ -0,0 +1,10 @@
+To update to a new OpenH264 release:
```
kmeisthax, line 1: TODO: This should probably have a GitHub Actions workflow to check for and apply the latest release.

kmeisthax (approved):
I think we can merge this.

core/src/streams.rs
torokati44: > I'm going to assume this means Flash didn't actually use the composition time offset in FLV H.264 streams either...

Oh, don't assume that! This may be exactly what I asked about in Discord. I just haven't connected the dots...
Also at the beginning, didn't care about anything else than what's strictly necessary for "get picture on screen", and just never got back to this after

video/external/src/BINARY_LICENSE.txt
torokati44: Yes I thought about that too. But then found that it would break through the abstraction of the backend API even more, and didn't feel like widening that API a lot more, with things like registering custom options for core/ui, or reporting which codecs are actually supported. Though this latter one might be needed for the runtime fallback to the current "video tag popup" on web, in case there's no WebCodecs available in the browser.

torokati44: And for the menu option - I guess the persistence part of #15455 ties in here a bit. WDYT @Dinnerbone?

video/external/src/decoder/README.md
torokati44: That's a great idea, I just couldn't quite figure out how to best handle the built-in hardcoded library version matching sanity check - the `== (2, 4, 1)` part.

video/external/src/BINARY_LICENSE.txt
Dinnerbone: For sure, desktop has persistent preferences now and can easily have UI for this.

core/src/streams.rs
torokati44: https://github.com/ruffle-rs/ruffle/pull/15747 should help investigating this.

danielhjacobs:
At least for me on Fedora, this seems to have added a dependency, as I had to do this. If there's an equivalent dependency added to Ubuntu it should probably be noted in [the README](https://github.com/ruffle-rs/ruffle?tab=readme-ov-file#ubuntu):

`sudo dnf install libudev-devel`

Dinnerbone:
> At least for me on Fedora, this seems to have added a dependency, as I had to do this. If there's an equivalent dependency added to Ubuntu it should probably be noted in [the README](https://github.com/ruffle-rs/ruffle?tab=readme-ov-file#ubuntu):
> 
> `sudo dnf install libudev-devel`

`libudev-dev` is already listed, and (from my extremely ignorant Linux knowledge so I may totally be wrong) should be the equivalent package?

sleepycatcoding:
> At least for me on Fedora, this seems to have added a dependency, as I had to do this. If there's an equivalent dependency added to Ubuntu it should probably be noted in [the README](https://github.com/ruffle-rs/ruffle?tab=readme-ov-file#ubuntu):
> 
> `sudo dnf install libudev-devel`

If I am remembering correctly `libudev` was introduced as a dependency by the gamepad support (#14568).

danielhjacobs:
Ah, I may have just not build the Desktop package from source since that PR was merged, my mistake.

torokati44:
One last thing: I think I'll replace the test with a simpler, generated, "more test-y" video, that allows for accurate verification of frame order and timing.

torokati44:
Changed from `bzip2-rs`[^1] to `bzip2`[^2].
Initially I used the former because I thought that the latter is only a wrapper for the library installed on the system.
Actually, it thankfully bundles the `bzip2` source code, and links it in statically, so no issues there!

Thank you for the async code, @J0sh0nat0r!

[^1]: https://github.com/paolobarbolini/bzip2-rs
[^2]: https://github.com/alexcrichton/bzip2-rs

torokati44:
I'm not entirely sure how enablement should be handled w.r.t. #16310... :thinking: ...as this is entirely desktop-specific.

sleepycatcoding:
> I'm not entirely sure how enablement should be handled w.r.t. #16310... 🤔 ...as this is entirely desktop-specific.

Maybe this should be in preferences instead?

torokati44:
> Maybe this should be in preferences instead and have a option to also toggle it in the UI?

Hahah, yeah, you're right, there just weren't preferences yet when I started working on it, and they didn't come to my mind just now... :sweat_smile:

torokati44:
Made the toggle a preference now, instead of a command line option.

desktop/src/gui/dialogs/preferences_dialog.rs
```diff
@@ -261,6 +269,23 @@ impl PreferencesDialog {
         ui.end_row();
     }
 
+    fn show_video_preferences(&mut self, locale: &LanguageIdentifier, ui: &mut Ui) {
+        ui.label(text(locale, "enable-openh264"));
+
+        let previous = self.enable_openh264;
+        ui.checkbox(&mut self.enable_openh264, text(locale, "enable-openh264"));
```
sleepycatcoding, line 276: Hmm, this looks a bit weird with the label on the left and a checkbox on the right with the same label.

How about `ui.add(Checkbox::without_text(&mut self.enable_openh264))`?

torokati44: Good point, thank you for the suggestion, changed.

desktop/src/preferences/read.rs
```diff
@@ -51,6 +51,10 @@ pub fn read_preferences(input: &str) -> ParseDetails<SavedGlobalPreferences> {
         result.mute = value;
     };
 
+    if let Some(value) = document.get_bool(&mut cx, "enable_openh264") {
+        result.enable_openh264 = value;
+    };
```
sleepycatcoding, line 56: Can we get a trivial read/write test for this, so we don't accidentally break this in the future?

torokati44: Honestly, I'm not really seeing how _this_, of all things, could be broken "accidentally", but sure, more tests are more good! So I mostly just copied the test for `mute` - another boolean preference -, and substituted `enable_openh264`.

torokati44:
There's now a bit of a back-and-forth between `video/external: Add OpenH264 decoder` and `desktop: Add a preference to enable the OpenH264 decoder`, but it doesn't bother even me that much to go through and rebase it out...

desktop/src/preferences/read.rs
sleepycatcoding: Sorry for being so nit-picky :sweat_smile:, but we also have write tests in `preferences/write.rs` (mostly for round trip purposes), so if you could add one for this it would be great!

torokati44: No problem, please do be! I'll get right on it, just haven't noticed them myself.

torokati44: Done.

video/external/src/decoder.rs
```diff
@@ -0,0 +1,11 @@
+// bindgen ../openh264/codec/api/wels/codec_api.h --no-prepend-enum-name \
```
Aaron1011, line 1: Could this be done through a `build.rs` script, instead of committing the file?

torokati44: From a technical PoV, of course.
Then, we'd have to vendor (or perhaps obtain in some other way) the C header in question at build time.
I think the license within it allows vendoring, and the header in itself should not be problematic regarding patents or royalties.
That is a smaller file to check in, at the price of another dependency and a smidge longer build time.
Would you prefer that?

desktop/src/player.rs
```diff
@@ -202,7 +202,19 @@ impl ActivePlayer {
             RfdNavigatorInterface,
         );
 
-        if cfg!(feature = "software_video") {
+        if cfg!(feature = "external_video") && preferences.openh264_enabled() {
+            use ruffle_video_external::backend::ExternalVideoBackend;
```
Dinnerbone, line 206: I appreciate the effort to feature gate this, but this is a compile error if the feature is turned off.

video/external/src/backend.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    pub fn get_openh264() -> Result<PathBuf, Box<dyn std::error::Error>> {
+        // See the license at: https://www.openh264.org/BINARY_LICENSE.txt
+        const URL_BASE: &str = "http://ciscobinary.openh264.org/";
+        const URL_SUFFIX: &str = ".bz2";
+
+        let (filename, md5sum) = Self::get_openh264_data()?;
+
+        let filepath = std::env::current_dir()?.join(filename);
```
Dinnerbone, line 88: Let's not use current_dir please... that could be anywhere
Maybe our config storage place for now?
Or absolute worst case, the folder that the ruffle executable is in... but that's *not* likely to be the same as current_dir

Dinnerbone (changes requested):

video/external/src/backend.rs
torokati44: > Maybe our config storage place for now?

Ummh, I'd find it weird to put anything "executable" in there.

> Or absolute worst case, the folder that the ruffle executable is in...

That should be fine IMHO.

> but that's not likely to be the same as current_dir

Yeah, good point, I'm a bit too Linux-CLI-minded... :sweat_smile:

desktop/src/player.rs
torokati44: Good catch, thank you! Added an inner block that is conditionally compiled.

video/external/src/backend.rs
torokati44: Changed to:
```rust
        let filepath = std::env::current_exe()?
            .parent()
            .ok_or("Could not determine Ruffle location.")?
            .join(filename);
```

This way, we don't need to touch `.gitignore` either, as the lib will be put into `target`.

video/external/src/decoder.rs
torokati44: Could you please weigh in on this one too, @Dinnerbone?

torokati44:
Some further minor TODOs:
 - Remove or inactivate the toggle option in the preferences dialog if the feature is not enabled in the build.
 - The keyframe detection does not handle the cases when there are multiple actual NALUs clumped together in a chunk, and the IDR NALU is preceded by a SEI NALU (which is not that rare). Also the SPS/PPS NALUs could perhaps be considered "keyframes"...? Perhaps not?
 - Change the "decoder did not produce a frame yet" from an `Err` to instead return a `Result<Option<...>,...>` from `decode_frame`, and make this case a `None`,

desktop/src/player.rs
```diff
@@ -220,7 +220,22 @@ impl ActivePlayer {
...
+                        tracing::error!("Couldn't get OpenH264: {}", e);
+                        None
+                    }
+                };
+
+                builder = builder.with_video(ExternalVideoBackend::new(openh264_path));
+            }
+        } else if cfg!(feature = "software_video") {
             builder =
                 builder.with_video(ruffle_video_software::backend::SoftwareVideoBackend::new());
```
torokati44, line 240: Oh no, this still fails to compile if the `software_video` feature is disabled...

torokati44: Fixed.

Dinnerbone (approved):
Works on mac and windows.
Slow on my mac but profiled on windows and 2/3rds of the time is spent in openh264 - the other third is `to_rgba` which we can look at offloading more of to gpu later maybe

video/external/src/decoder.rs
Dinnerbone: I don't care enough to block this PR. If someone feels strongly about it, let's follow up in another PR later.

Benman2785:
https://hoffmeister.li/tv/stream1.html

</script>
<script src="[https://hoffmeister.li/flash-games/ruffle/ruffle.js](view-source:https://hoffmeister.li/flash-games/ruffle/ruffle.js)"></script>
</head>
<body>
<object type="application/x-shockwave-flash" data="https://hoffmeister.li/flash-games/swf/player_flv_mini.swf" width="100%" height="100%">
    <param name="movie" value="https://hoffmeister.li/flash-games/swf/player_flv_mini.swf" />
    <param name="allowFullScreen" value="true" />
    <param name="FlashVars" value="flv=https://hoffmeister.li/tv/test1.flv&amp;autoplay=1&amp;autoload=1" />
</object>

does NOT work :(

lvyitian:
> https://hoffmeister.li/tv/stream1.html
> </script> <script src="[https://hoffmeister.li/flash-games/ruffle/ruffle.js](view-source:https://hoffmeister.li/flash-games/ruffle/ruffle.js)"></script>
> 
> does NOT work :(
@Benman2785 This PR is **DESKTOP-ONLY**, it does **NOT** affect the web build of ruffle.

lvyitian:
> https://hoffmeister.li/tv/stream1.html
> </script> <script src="[https://hoffmeister.li/flash-games/ruffle/ruffle.js](view-source:https://hoffmeister.li/flash-games/ruffle/ruffle.js)"></script>
> 
> does NOT work :(

@Benman2785 This PR is **DESKTOP-ONLY**, it does **NOT** affect the web build of ruffle.

lvyitian:
> https://hoffmeister.li/tv/stream1.html
> </script> <script src="[https://hoffmeister.li/flash-games/ruffle/ruffle.js](view-source:https://hoffmeister.li/flash-games/ruffle/ruffle.js)"></script>
> 
> does NOT work :(

@Benman2785 The web implementation of H264 decoding will be based on WebCodecs API and is still work in progress.

lvyitian:
> https://hoffmeister.li/tv/stream1.html
> </script> <script src="[https://hoffmeister.li/flash-games/ruffle/ruffle.js](view-source:https://hoffmeister.li/flash-games/ruffle/ruffle.js)"></script>
> 
> does NOT work :(

@Benman2785 The reason is that if you don't use the OpenH264 prebuilt binaries provided by Cisco, you will have to pay for the patent fee to MPEG LA by yourself, and Cisco hasn't provided the WebAssembly-targeted prebuilt binaries of OpenH264, so we have to use the built-in decoders in the browser.

rafrafek:
Should I click Allow Anyway?

image

PR #14655, opened 2024-01-08
video: Add support for playing F4V (MP4) files

torokati44:
Similar to the somewhat related https://github.com/ruffle-rs/ruffle/pull/14654, this is in a really early stage.
 
 TODOs:
 - [ ] Don't hardcode track index
 - [ ] Precompute and store all frame ("sample" in MP4 lingo) offsets and lengths for all tracks?
 - [ ] Handle audio
 - [ ] Handle all supported video codecs
 - [ ] Seeking
 - [x] Consider whether `mp4` would be a better fit than `mp4parse`
   - Not really: It's larger (according to crates.io, and it includes writing as well as reading), and is really geared towards working with files on disk - unlike `mp4parse`, which is made for a really similar use case in Firefox.
 - [x] How about Symphonia?
   - Nope, sadly it won't even give me the codec type or extradata of video tracks: https://github.com/pdeljanov/Symphonia/issues/271
 - [x] Resolve inevitable conflicts with #14647
 - [x] Consider adding the ability to play MP4 files directly (as is on https://z0r.de/3099 - WARNING: content is not nice!)
   - Out of scope for this PR. Some other kinds of files (maybe images?) could also be opened this way.
 - [ ] Cleanup and testing, as usual


core/src/streams.rs
```diff
@@ -1189,6 +1224,168 @@ impl<'gc> NetStream<'gc> {
             }
         }
 
+        if let Some(NetStreamType::F4v {
+            context: media_context,
+            frame_id,
+            video_stream,
+        }) = &mut write.stream_type
+        {
+            println!("frame id: {:?}, end_time: {}", frame_id, end_time);
```
danielhjacobs, line 1233: Seems to now be called `max_time`: https://github.com/ruffle-rs/ruffle/blob/df0ac4395b84be317e39271a8f5a1c7d6227ada4/core/src/streams.rs#L1110

core/src/streams.rs
```diff
@@ -816,11 +830,32 @@ impl<'gc> NetStream<'gc> {
...
+            //  - moov (see “moov box” on page 19)
+            //  - mdat (see “mdat box” on page 32)
+            // And the first 4 bytes are the length of the first box.
+            Some([_, _, _, _, b'f', b't', b'y', b'p'])
+            | Some([_, _, _, _, b'm', b'o', b'o', b'v'])
+            | Some([_, _, _, _, b'm', b'd', b'a', b't']) => {
+                println!("F4V");
+
+                let mut cursor = slice.as_cursor();
+                let context = mp4parse::read_mp4(&mut cursor).unwrap();
```
danielhjacobs, line 850: I know his is still a draft, so I don't know if it's supposed to be panic-free yet, but on http://new.weedtowonder.org/, when click the play button on the video, this caused the panic :
```
panicked at core/src/streams.rs:850:63:
called `Result::unwrap()` on an `Err` value: InvalidData(CheckParserStateErr)
```

torokati44: Aand now the site seem to have changed...?

danielhjacobs: On web, and I changed the site to add https. It's now https://new.weedtowonder.org/.

torokati44:
Just found a cool web based MP4 dissector tool: https://gpac.github.io/mp4box.js/test/filereader.html

danielhjacobs:
Also note, this time testing with the attached files and a directory named mp4-1200 containing the files from https://github.com/danielhjacobs/danielhjacobs.github.io/tree/main/mp4-1200, I get this error when opening anastasia2017-teacher.swf with a Desktop app build from this PR:
```rs
panicked at core/src/streams.rs:1384:26:
called `Result::unwrap()` on an `Err` value: DecoderError(DecoderError(MiddleOfBitstream))
```

[anastasia.zip](https://github.com/ruffle-rs/ruffle/files/14824034/anastasia.zip)

danielhjacobs:
Replacing https://github.com/ruffle-rs/ruffle/pull/14655/files#diff-894b995744e59b9d8803f92eb6b185191b9018cd571feeaeb4a7c53dbd9a0bc5R1380-R1385 with this removes the panic:
```rs
                match context.video.decode_video_stream_frame(
                    video_handle,
                    encoded_frame,
                    context.renderer,
                ) {
                    Ok(bitmap) => {
                        write.last_decoded_bitmap = Some(bitmap);
                    }
                    Err(e) => {
                        tracing::error!(
                            "Got error when decoding FLV video stream: {}",
                            e
                        );
                        return;
                    }
                }
```

But that doesn't fix the underlying issue, as then there is the message `ruffle_core::streams: Got error when decoding FLV video stream: Decoder error` and since the video won't open in browser anymore as a workaround as it attempts to be decoded (and fails).

torokati44:
Hmm, maybe this should be rewritten using https://crates.io/crates/re_mp4

tests/tests/swfs/visual/video/f4v/test.toml
```diff
@@ -0,0 +1,17 @@
+num_frames = 75
+sleep_to_meet_frame_rate = true
```
kjarosh, line 2: Can we not do this?

torokati44: This was just another attempt to make the capture frame not be white...

torokati44:
Closing in favor of https://github.com/ruffle-rs/ruffle/pull/24038.

Issue #14340, opened 2023-12-08
Knightmare Tower does not go past the loading screen


flashliberty5467:
### Describe the bug

N/A as it does not go past the loading screen

![loading screen](https://github.com/ruffle-rs/ruffle/assets/64484245/468fc235-968e-420b-a19e-5a376b0a9b74)

### Expected behavior

to be able to test the game

### Content Location

https://www.maxgames.com/play/knightmare-tower.html

### Affected platform

Browser's extension

### Operating system

ChromeOS

### Browser

Version 118.0.5993.164 (Official Build) (64-bit)

### Additional information

# Player Info
Allows script access: false
Renderer: wgpu
Adapter Backend: Gl
Adapter Name: "ANGLE (Intel, Mesa Intel(R) UHD Graphics 600 (GLK 2), OpenGL ES 3.2)"
Adapter Device Type: IntegratedGpu
Adapter Driver Name: ""
Adapter Driver Info: ""
Enabled features: Features(TEXTURE_COMPRESSION_BC | TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES | PUSH_CONSTANTS | SHADER_UNUSED_VERTEX_OUTPUT)
Available features: Features(TEXTURE_COMPRESSION_ETC2 | TEXTURE_COMPRESSION_ASTC | CLEAR_TEXTURE)
[...]

Issue #10150, opened 2023-03-17
Knightmare Tower: Stuck before Title Screen


waspennator:
### Describe the bug

[www.newgrounds.com-1679026851203.log](https://github.com/ruffle-rs/ruffle/files/10997841/www.newgrounds.com-1679026851203.log)

https://www.newgrounds.com/portal/view/602131?emulate=flash
As soon as you click play, it will go through a screen transition or two, then get stuck on this.

![Knightmare Tower Ruffle](https://user-images.githubusercontent.com/78768499/225811948-4eda1924-5273-42de-9dc1-4317015262af.png)

### Expected behavior

The game should be starting up without issue.

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 110

Lord-McSweeney:
Relevant error: `RustError("VerifyError: Trait isPlaying in class CoreClip has same name as trait isPlaying in class MovieClip, but does not override it")`.

Lord-McSweeney:
This also seems to need `XML.copy`.

Lord-McSweeney:
Should be fixed by #14671

PR #14671, opened 2024-01-09
avm2: Run most of Loader.loadBytes immediately

Aaron1011:
This requires moving `set_root_movie` into `UpdateContext`.

Now, we preload the entire movieclip immediately - Flash Player does this regardless of the size of the SWF.
The 'Loader::load_complete' is delayed to the end of the frame (which is when the root class is constructed for the loaded clip).

Fixes https://github.com/ruffle-rs/ruffle/issues/14340, https://github.com/ruffle-rs/ruffle/issues/10150, #12484


core/src/loader.rs
```diff
@@ -2005,6 +2033,16 @@ impl<'gc> Loader<'gc> {
                 .as_container()
                 .unwrap();
 
+            // This isn't completely correct - the 'large_preload' test observes the child
+            // being set after an 'enterFrame' call. However, our current logic should
+            // hopefully be good enough.
+            avm2_stub_method!(
```
Lord-McSweeney, line 2039: Might `context_stub` be better for this?

Aaron1011: This is still associated with `Loader.loadBytes`, even though it's not directly called from that method

Dinnerbone (approved):

PR #14672, opened 2024-01-09
debug_ui: Add 'Show Domains' menu to view domains and classes

Aaron1011:
This opens a searchable list (similar to what we have for display objects), which shows a tree of Domains and their associated classes.

Currently, clicking on the domain/class buttons doesn't do anything. In a follow-up, I'm planning to add additional windows to display information about a class.

domain_class_search

### DIFF
--- a/core/src/avm2.rs
+++ b/core/src/avm2.rs
@@ -66,7 +66,8 @@ mod vtable;
 pub use crate::avm2::activation::Activation;
 pub use crate::avm2::array::ArrayStorage;
 pub use crate::avm2::call_stack::{CallNode, CallStack};
-pub use crate::avm2::domain::Domain;
+#[allow(unused)] // For debug_ui
+pub use crate::avm2::domain::{Domain, DomainPtr};
 pub use crate::avm2::error::Error;
 pub use crate::avm2::flv::FlvValueAvm2Ext;
 pub use crate::avm2::globals::flash::ui::context_menu::make_context_menu_state;
@@ -257,6 +258,10 @@ impl<'gc> Avm2<'gc> {
         let globals = context.avm2.playerglobals_domain;
         let mut activation = Activation::from_domain(context.reborrow(), globals);
         globals::load_player_globals(&mut activation, globals)
+    }
+
+    pub fn playerglobals_domain(&self) -> Domain<'gc> {
+        self.playerglobals_domain
     }
 
     /// Return the current set of system classes.
--- a/core/src/avm2/domain.rs
+++ b/core/src/avm2/domain.rs
@@ -1,5 +1,7 @@
 //! Application Domains
 
+use std::cell::Ref;
+
 use crate::avm2::activation::Activation;
 use crate::avm2::object::{ByteArrayObject, TObject};
 use crate::avm2::property_map::PropertyMap;
@@ -8,7 +10,7 @@ use crate::avm2::value::Value;
 use crate::avm2::Error;
 use crate::avm2::Multiname;
 use crate::avm2::QName;
-use gc_arena::{Collect, GcCell, Mutation};
+use gc_arena::{Collect, GcCell, GcWeakCell, Mutation};
 use ruffle_wstr::WStr;
 
 use super::class::Class;
@@ -21,6 +23,10 @@ use super::Avm2;
 #[derive(Copy, Clone, Collect)]
 #[collect(no_drop)]
 pub struct Domain<'gc>(GcCell<'gc, DomainData<'gc>>);
+
+#[derive(Copy, Clone, Collect)]
+#[collect(no_drop)]
+pub struct DomainWeak<'gc>(GcWeakCell<'gc, DomainData<'gc>>);
 
 #[derive(Clone, Collect)]
 #[collect(no_drop)]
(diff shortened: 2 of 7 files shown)


core/src/avm2.rs
```diff
@@ -66,7 +66,8 @@ mod vtable;
 pub use crate::avm2::activation::Activation;
 pub use crate::avm2::array::ArrayStorage;
 pub use crate::avm2::call_stack::{CallNode, CallStack};
-pub use crate::avm2::domain::Domain;
+#[allow(unused)] // For debug_ui
+pub use crate::avm2::domain::{Domain, DomainPtr};
```
torokati44, line 70: I'm not seeing any actual `DomainPtr` usages added to this file. Is this a leftover, hidden by the `allow`?

core/src/avm2/domain.rs
```diff
@@ -58,22 +68,49 @@ impl<'gc> Domain<'gc> {
...
+    pub fn classes(&self) -> Ref<'_, PropertyMap<'gc, GcCell<'gc, Class<'gc>>>> {
+        Ref::map(self.0.read(), |r| &r.classes)
     }
 
     pub fn is_playerglobals_domain(&self, avm2: &Avm2<'gc>) -> bool {
         avm2.playerglobals_domain.0.as_ptr() == self.0.as_ptr()
     }
 
+    pub fn children(&self, mc: &Mutation<'gc>) -> Vec<Domain<'gc>> {
+        // Take this opportunity to clean up dead children.
```
torokati44, line 101: So if the user never opens this new dialog, the children list will grow and grow with more and more dead weak refs?
Not that I assume domains come and go that often in large numbers, or that these refs would consume that much memory...
But am I getting this right?

core/src/avm2/domain.rs
```diff
@@ -383,8 +427,14 @@ impl<'gc> Domain<'gc> {
 
         Ok(())
     }
+
+    pub fn as_ptr(self) -> *const DomainPtr {
+        self.0.as_ptr() as _
+    }
 }
 
+pub enum DomainPtr {}
```
torokati44, line 436: I'm not seeing the entire codebase right now, but is this pattern used in other places as well? If so, that's fine, I'm just not familiar with it. Otherwise, a comment would be nice about what this is, and why is it the way it is.

core/src/debug_ui.rs
```diff
@@ -21,27 +23,33 @@ use std::sync::{Arc, Weak};
 use swf::{Color, Rectangle, Twips};
 use weak_table::PtrWeakKeyHashMap;
 
+use self::handle::DomainHandle;
```
torokati44, line 26: Wouldn't this be nicer up there on line 13 together with the other handles?

core/src/debug_ui/domain.rs
```diff
@@ -0,0 +1,88 @@
...
+                let domain = context.avm2.playerglobals_domain();
+                egui::ScrollArea::both().show(ui, |ui| {
+                    self.show_domain(ui, context, domain, messages, &search)
+                });
+            });
+        keep_open
+    }
+
+    #[allow(clippy::only_used_in_recursion)]
+    pub fn show_domain<'gc>(
```
torokati44, line 39: Is this `allow` still really necessary? It's clearly also used non-recursively just a few lines above. Does Clippy not detect that - and if so, is this a Clippy bug? Or am I missing something?

core/src/avm2/domain.rs
Dinnerbone: It's used in quite some places

core/src/avm2/domain.rs
Aaron1011: That's right - a domain should be created at most once per loaded swf (unless a game does something really weird), so I think this should be fine. We could skip adding it when the debug ui is disabled, but I think this could be useful for debugging

core/src/debug_ui/domain.rs
Aaron1011: This is for a parameter, not for the function itself

core/src/debug_ui.rs
Aaron1011: I guess, but I think things like import order should just be left up to `rustfmt`

torokati44: Absolutely, but it won't switch between `crate::` and `self::`, because that's semantically not the same. In this case, it won't matter (AFAIU), but switching would make it look nicer.

core/src/debug_ui/domain.rs
torokati44: Oh, it's about `self`. If you think it will be needed for when the `Class` buttons get wired up, it's fine to `allow` it - otherwise, it might be better to remove the `&mut self,`, and call as `Self::show_domain` instead, as Clippy suggests.
(Yeah I know it's nitpicking at this point, but maintaining code quality is finicky like this.)

Aaron1011: Yeah, I'm going to need to store some data when I implement the `Class` viewer

core/src/avm2.rs
Aaron1011: It's used by `handle.rs` through the re-export

torokati44: Ah, okay. I won't block on this then.

torokati44:
Thank you!

torokati44 (approved):

Issue #263, opened 2020-01-11
Actionscript Tracking: MovieClip


Dinnerbone:
This is the tracking issue for **MovieClip** in AVM1. We will check off each item of progress as appropriate, and fill in any relevant or missing information as we continue development of Ruffle.
# Legend
Each checkbox is independent of another. It's entirely possible for something to be tested but not exist yet, or for us to believe that it's completely functional but we haven't made enough tests to prove it.
## "Property Exists"
This means the item exists, but may not necessarily be fully implemented.
## "Has Test Coverage"
This means that we have believe that we have a good test coverage of this item, regardless of if those tests pass. It's okay to have tests available but not implement the item yet.
## "Completely Functional"
This means we believe that the item is completely implemented, and no more work needs to be done towards making it functional.

# MovieClip
### `attachVideo()`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional

### `_alpha : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_currentframe : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_droptarget : String`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_focusrect : Boolean`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_framesloaded : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_height : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_highquality : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_lockroot : Boolean`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_name : String`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_parent : MovieClip`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_quality : String`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_rotation : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_soundbuftime : Number`
- [ ] Property Exists
[...]

Issue #243, opened 2020-01-11
Actionscript Tracking: Button


Dinnerbone:
This is the tracking issue for **Button** in AVM1. We will check off each item of progress as appropriate, and fill in any relevant or missing information as we continue development of Ruffle.
# Legend
Each checkbox is independent of another. It's entirely possible for something to be tested but not exist yet, or for us to believe that it's completely functional but we haven't made enough tests to prove it.
## "Property Exists"
This means the item exists, but may not necessarily be fully implemented.
## "Has Test Coverage"
This means that we have believe that we have a good test coverage of this item, regardless of if those tests pass. It's okay to have tests available but not implement the item yet.
## "Completely Functional"
This means we believe that the item is completely implemented, and no more work needs to be done towards making it functional.

# Button
### `_alpha : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_focusrect : Boolean`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_height : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_highquality : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `menu : ContextMenu`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_name : String`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_parent : MovieClip`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_quality : String`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_rotation : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_soundbuftime : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_target : String`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_url : String`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_visible : Boolean`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_width : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
[...]

PR #14676, opened 2024-01-09
avm1: Add support for onSetFocus & onKillFocus

kjarosh:
Methods `onSetFocus` & `onKillFocus` are invoked when focus is changed for `TextField`, `Button`, and `MovieClip`.
Multiple SWFs use these methods to listen to a focus change, e.g. in order to implement placeholders for text fields.

Related to #243, #263, #280. If I understand it correctly, these issues falsely claim that `onSetFocus` is implemented (only `Selection.onSetFocus` is implemented).

I tried my best to implement it cleanly, please let me know if there's a better way.


Lord-McSweeney (approved):

core/src/display_object/avm1_button.rs
```diff
@@ -399,8 +401,33 @@ impl<'gc> TDisplayObject<'gc> for Avm1Button<'gc> {
...
-        self.0.write(gc_context).has_focus = focused;
+    fn on_focus_changed(
+        &self,
+        context: &mut UpdateContext<'_, 'gc>,
+        focused: bool,
+        other: Option<DisplayObject<'gc>>,
+    ) {
+        self.0.write(context.gc_context).has_focus = focused;
+
+        if let Value::Object(object) = self.object() {
```
Aaron1011, line 412: Could you extract the body of this `if let` into a separate function, and call it from all of the displayobjects that need it?

Aaron1011 (approved):

PR #14690, opened 2024-01-10
core: Fix caret and selection rendering

### LINKED ISSUE #10478: cursor missing in input textfield,  avm1
### Describe the bug

When an input textfield is selected, but nothing is written in it, the cursor doesn't show.
Is it possible to add this to the code? 
It wasn't a problem in earlier versions of ruffle,    i.e. one from 8th of july 2021.
[input_text_cursor_issue.zip](https://github.com/ruffle-rs/ruffle/files/11122504/input_text_cursor_issue.zip)

### Expected behavior

Input textfield works best if the cursor becomes visible the very moment an input textfield is selected, - before the user starts to write in it.

### Affected platform

Self-hosted version

### Operating system

windows 10

### Browser

Firefox

### Additional information

_No response_

kjarosh:
This commit fixes issues with caret and selection rendering:
1. They had the wrong height and were rendered lower than expected for some fonts and sizes.
2. The caret was not being rendered at all when there was no text, but only when the text was set earlier and then deleted.
3. The selection was rendered with translate_x=-1, which caused overlap over some glyphs.

This PR fixes #10478.

| Flashplayer   | Before        | After |
|:-------------:|:-------------:|:-----:|
| image | image | image |
| image | image | image |
| image | image | image |
| image | image | image |
|  |  |  |
|  |  |  |

### DIFF
--- a/core/src/display_object/edit_text.rs
+++ b/core/src/display_object/edit_text.rs
@@ -934,7 +934,7 @@ impl<'gc> EditText<'gc> {
                     && visible_selection.end() <= *end
                     && Utc::now().timestamp_subsec_millis() / 500 == 0
                 {
-                    Some((visible_selection.start() - start, end - start))
+                    Some(visible_selection.start() - start)
                 } else {
                     None
                 }
@@ -958,8 +958,11 @@ impl<'gc> EditText<'gc> {
         if let Some((text, _tf, font, params, color)) =
             lbox.as_renderable_text(edit_text.text_spans.displayed_text())
         {
-            let baseline_adjustment =
-                font.get_baseline_for_height(params.height()) - params.height();
+            let baseline = font.get_baseline_for_height(params.height());
+            let descent = font.get_descent_for_height(params.height());
+            let baseline_adjustment = baseline - params.height();
+            let caret_height = baseline + descent;
+            let mut caret_x = Twips::ZERO;
             font.evaluate(
                 text,
                 self.text_transform(color, baseline_adjustment),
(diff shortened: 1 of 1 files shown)


core/src/display_object/edit_text.rs
```diff
@@ -1032,6 +1014,43 @@ impl<'gc> EditText<'gc> {
...
+    ) {
+        let selection_box = context.transform_stack.transform().matrix
+            * Matrix::create_box(
+                width.to_pixels() as f32,
+                height.to_pixels() as f32,
+                0.0,
+                x,
+                Twips::ZERO,
+            );
+        context.commands.draw_rect(Color::BLACK, selection_box);
```
torokati44, line 1032: Is this really always `BLACK`?
I see this was similarly hardcoded before at the call site, but if it can change, perhaps adding a parameter and keeping the value there, or just adding a TODO comment, would be nice.

torokati44: Otherwise LGTM, thank you!

kjarosh: It's a good question. It seems like black is the default and it cannot be changed at all. I could find only the following.
* https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flashx/textLayout/edit/SelectionManager.html#focusedSelectionFormat
* https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flashx/textLayout/elements/Configuration.html#focusedSelectionFormat

From what I understand, it's a library which somehow implements a custom selection color on its own (https://github.com/apache/flex-tlf/blob/master/textLayout/src/flashx/textLayout/container/ContainerController.as#L2901).

core/src/display_object/edit_text.rs
```diff
@@ -1977,19 +1996,9 @@ impl<'gc> TDisplayObject<'gc> for EditText<'gc> {
...
-                                .size
-                                .unwrap_or_default() as f32,
-                            0.0,
-                            Twips::from_pixels(-1.0),
-                            Twips::from_pixels(2.0),
-                        );
-                    context.commands.draw_rect(Color::BLACK, caret);
+                    let format = edit_text.text_spans.default_format();
+                    let caret_height = format.size.map(Twips::from_pixels).unwrap_or_default();
+                    self.render_caret(context, Twips::ZERO, caret_height, Color::BLACK);
```
kjarosh, line 2001: This code is not perfect... It is executed when the editable text is being rendered without any layout_box in order to show the caret. But without any layout box we do not have information related to the font, etc. and `caret_height` is only an estimate which is often wrong.

From my testing this is the case when an empty editable text is spawned. But when any text is added and removed, now we do have an empty layout box, and the caret is rendered correctly. In my opinion it would be wise to remove this code altogether and replace it with the case of one empty layout box (forcing EditText to always have at least one layout box), but I'm not sure whether it's right and how to do that, and so I have left it as is.

core/src/display_object/edit_text.rs
torokati44: Ah, I see. But all of these are for `TextLayout`, and this PR is only about `EditText`, right?

core/src/display_object/edit_text.rs
torokati44: >  In my opinion it would be wise to remove this code altogether and replace it with the case of one empty layout box (forcing EditText to always have at least one layout box)

Yeah that sounds just about what I also would have suggested.

> but I'm not sure whether it's right and how to do that, and so I have left it as is.

It's okay, we can do that later as well.

core/src/display_object/edit_text.rs
kjarosh: If I understand it correctly, it's for `TLFTextField`, which is from an external library included inside the SWF (so Ruffle and flashplayer do not implement it at all). It uses some mechanisms (sprites, etc.) to model the field so it looks like `TextField` but has more features (like a custom selection color).

Ruffle only implements `TextField` (internally as `EditText`), and from my reconnaissance it seems like the selection color is always black and cannot be changed. The internet always redirects to `TLFTextField` when someone wants to change the selection color. It's also confirmed by my testing, where custom fonts, colors, colorTransforms, etc. do not change the behavior: the selection is always black.

torokati44 (approved):

torokati44:
Thank you!

PR #14697, opened 2024-01-11
core: Use sans instead of serif when missing a font

evilpie:
Seems like I was wrong and @Dinnerbone was right.

image


torokati44 (approved):
@Dinnerbone _is always right!!!_ :stuck_out_tongue_closed_eyes:

PR #14699, opened 2024-01-11
core: Support F13..F24 keys

### LINKED ISSUE #14698: Extra Function Keys are not Supported F13-F24
*tested in Desktop app, both the current Ruffle release and nightly build from Jan 11, 2024.

I noticed that Ruffle doesn't know what the F13-F24 keys are.
e.g. When pressing the F13 key, using Key.getCode() in AS2, Ruffle reports the key as '0'. But this has always been an undocumented feature of Flash Player and works in the official Flash Player.

In AS 1.0-2.0, the keycodes are 124-135 (for F13-F24 respectively). If you're wondering, while there are keyboards out there with F13-F24, I activate those keys on my StreamDeck.

This would be a helpful addition because the F13-F24 keys are rarely used and don't interfere with other programs in Windows. I use Ruffle with a stream overlay and other interactive experiences that are too slow for the official Flash Player to handle. Others using the Desktop Ruffle player could find this useful too!

Dinnerbone:
Closes #14698

### DIFF
--- a/core/src/avm2/globals/flash/events/KeyboardEvent.as
+++ b/core/src/avm2/globals/flash/events/KeyboardEvent.as
@@ -97,6 +97,10 @@ package flash.events
             return new KeyboardEvent(this.type,this.bubbles,this.cancelable,this._charCode,this._keyCode,this._keyLocation,this._ctrlKey,this._altKey,this._shiftKey,this._controlKey,this._commandKey);
         }
 
+        override public function toString(): String {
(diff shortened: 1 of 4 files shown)


torokati44:
Why split `web` into its own commit, but not `desktop`?

torokati44 (approved):

Dinnerbone:
> Why split `web` into its own commit, but not `desktop`?

Because I did desktop and then realized web can also support it. :D

torokati44:
But it's not consistent this way! 😭😄

Dinnerbone:
You have permission to rebase!

torokati44:
There you go!

Issue #8870, opened 2022-12-19
Window size does not match content on Wayland


torokati44:
### Describe the bug

When opening any content on Linux Desktop, the player window is always the default size.

Reverting to X11 (even via XWayland) fixes this.

Could be a `winit` or a `KDE` issue as well.

### Expected behavior

The player windows should match be the same size as the content - as it is with FP too.

### Affected platform

Desktop app

### Operating system

Fedora Linux 37

torokati44:
With https://github.com/rust-windowing/winit/pull/2676 just merged, this will soon be kinda fixable by doing this:

```diff
                         WindowEvent::Resized(size) => {
                             // TODO: Change this when winit adds a `Window::minimzed` or `WindowEvent::Minimize`.
                             minimized = size.width == 0 && size.height == 0;
 
                             let viewport_scale_factor = self.window.scale_factor();
                             let mut player_lock = self.player.lock().expect("Cannot reenter");
                             player_lock.set_viewport_dimensions(ViewportDimensions {
                                 width: size.width,
                                 height: size.height,
                                 scale_factor: viewport_scale_factor,
                             });
                             self.window.request_redraw();
+
+                            self.window.set_inner_size(LogicalSize::new(player_lock.movie_width(), player_lock.movie_height()));
+
                             if matches!(loaded, LoadingState::WaitingForResize) {
                                 loaded = LoadingState::Loaded;
                             }
                         }
```

But it's uh.... weird. A more sophisticated way should be found I guess.

torokati44:
Doing something like https://github.com/slint-ui/slint/pull/3841 could help here.

PR #13731, opened 2023-10-26
Update to winit 0.29

torokati44:
This is very much WIP:
  The reworked Keyboard handling is, for example, mostly just commented out.
  Also there is something wrong with resizing the window, at least on Linux+Wayland.

See the upstream changes here: https://github.com/rust-windowing/winit/compare/v0.28.7...v0.29.9
And here: https://github.com/rust-windowing/winit/blob/master/CHANGELOG.md#0299 (all the way from 0.28.7)

Feel free to force-push this branch if any of you continue working on this!

~~Currently blocked by: https://github.com/emilk/egui/pull/3649~~
~~Currently waiting for `egui` `0.25`: https://github.com/emilk/egui/milestone/3~~
It just got released, unblocking this! https://github.com/emilk/egui/releases/tag/0.25.0

PR #14616, opened 2024-01-04
chore: Upgrade `gc-arena` to 0.5.0

torokati44:
This was fairly painless - I hope nothing breaks.

PR #11059, opened 2023-05-15
core: Implement handling of text control input

n0samu:
This is a rebase and update of #2081. The same TODO items still apply, but those would be for a possible future PR. I've moved the left/right arrow key handling into the same code path as the rest of the text control events. ~~I'm marking this PR as a draft because I plan to implement pasting from the clipboard on web - that's the main feature still missing. In the meantime, please provide your reviews/feedback!~~ I've implemented clipboard pasting on web!

https://github.com/ruffle-rs/ruffle/assets/71368227/abf6f939-1f8d-4faf-ad75-829d4da7d8ad

Fixes the save/load menus in #4817, Fixes part of #1891.

PR #14703, opened 2024-01-11
desktop: Update winit to 0.29.10 and egui to 0.25.0

Dinnerbone:
Supersedes #13731

Testing of text and key inputs is requested, everything changed about those.


torokati44:
I think the resizing logic also changed, at least on Linux+Wayland: We no longer get a resize event after setting the window size ourselves. AFAIK, will test.
Also, there was that hack to set the initial window size (which didn't work on Linux+Wayland), maybe that is no longer necessary...?

torokati44:
And thank you for taking this over! 😊

torokati44:
Unfortunately, does not (currently) fix #8870.

n0samu:
Unfortunately this breaks the Ctrl-A, Ctrl-C and Ctrl-V text control inputs implemented in #11059 😦

Dinnerbone:
> Unfortunately this breaks the Ctrl-A, Ctrl-C and Ctrl-V text control inputs implemented in #11059 😦

Seems to have been an undocumented breaking change in egui. Opened https://github.com/emilk/egui/pull/3812 and rebased this PR to use that for now

torokati44:
I was wondering why `cargo-deny` didn't fail CI on this one - it's not supposed to allow new git crate sources to "randomly appear".
Turns out, it only warns about them. Which is fine, I guess.
Just for cleanliness' sake, I guess the `cargo-deny` config could be updated to allow the `Dinnerbone` GH organization as crate source.
While somewhat less related, we could also remove `kyren`, as we also get a warning about that, since we have since moved to a published release of `gc-arena` in #14616.
But all this can be a separate follow-up PR too. Maybe it _should_ be. And there we could even turn these warnings into errors, so they won't fall out of sync again.

torokati44:
`Cargo.lock` should be updated for the latest commit in your `egui` branch before merging - it's not up to date at the moment.

desktop/src/app.rs
```diff
@@ -100,10 +101,10 @@ impl App {
 
         // Poll UI events.
         let event_loop = self.event_loop.take().expect("App already running");
-        event_loop.run(move |event, _window_target, control_flow| {
+        event_loop.run(move |event, elwt| {
```
torokati44, line 104: Huh, it took me a while to figure out that `elwt` is for `EventLoopWindowTarget`.
It can stay like this (this is a common pattern after all, just for an unfamiliar - to me - struct name), or it can be perhaps `target`? Even I'm not sure which one would be better...

Dinnerbone: Yeah it felt funny to me too, I got it from the winit docs and felt I should be consistent with what they use 😅

torokati44: Okay, if they use it, maybe it's better that we do too, yeah.

desktop/src/app.rs
```diff
@@ -211,7 +213,9 @@ impl App {
                                 MouseButton::Left => RuffleMouseButton::Left,
                                 MouseButton::Right => RuffleMouseButton::Right,
                                 MouseButton::Middle => RuffleMouseButton::Middle,
-                                MouseButton::Other(_) => RuffleMouseButton::Unknown,
+                                MouseButton::Back
+                                | MouseButton::Forward
+                                | MouseButton::Other(_) => RuffleMouseButton::Unknown,
```
torokati44, line 218: Is it nicer to list all these variants instead of just `_`? I don't mind, and it's not that many lines, just caught my eye.

desktop/src/app.rs
```diff
@@ -112,9 +113,7 @@ impl App {
                 }
 
                 // Core loop
-                winit::event::Event::MainEventsCleared
-                    if matches!(loaded, LoadingState::Loaded) =>
-                {
+                winit::event::Event::AboutToWait if matches!(loaded, LoadingState::Loaded) => {
```
torokati44, line 116: Do we want to mention in a comment that `AboutToWait` is not _actually perfectly_ what we'd want - as discussed on Discord -, and that we just cross our fingers, and most of the time it does work out? Not that important, it's just in our collective thought contexts now, but may not be later, _if_ it ends up causing weird issues.

desktop/src/app.rs
```diff
@@ -304,38 +308,40 @@ impl App {
...
-                            if let Some(key) = input.virtual_keycode {
-                                let key_code = winit_to_ruffle_key_code(key);
-                                let key_char = winit_key_to_char(key, modifiers.shift());
-                                match input.state {
-                                    ElementState::Pressed => {
-                                        self.player.handle_event(PlayerEvent::KeyDown {
-                                            key_code,
-                                            key_char,
+                            let key_code = winit_to_ruffle_key_code(&event);
+                            let key_char = event.text.clone().and_then(|text| text.chars().last());
```
torokati44, line 312: Hm, what about the other chars that are not `last`? :thinking:

desktop/src/app.rs
```diff
@@ -385,7 +391,7 @@ impl App {
                         self.window.scale_factor(),
                     );
 
-                    self.window.set_inner_size(window_size);
+                    let _ = self.window.request_inner_size(window_size);
```
torokati44, line 394: I feel like caring about the return value here _could_ do something about #8870, but maybe not, and also, that can be done later. For now, I guess it doesn't matter, I'm not seeing any weird scaling behavior.

torokati44:
Oh, it seems like mouse events are being passed to the movie even when I'm navigating the `egui` menu.

torokati44:
Meanwhile, `wgpu` `0.29.10` just came out.

torokati44:
Yes, that.

desktop/src/app.rs
Dinnerbone: I think `last()` emulates the behaviour we'd have seen before.

Long term I don't know if this is something we should keep, but for now I think this is essentially the same as current

desktop/src/app.rs
Dinnerbone: Yeah good call, I'll put a comment here. Given nosamu saw behaviour changes, my immediate thought is that this is likely the culprit already and we'll wanna revisit this soon

desktop/src/app.rs
torokati44: Maybe TODO comment it then? However you feel like...

desktop/src/app.rs
Dinnerbone: Wasn't sure if we'd want to be informed on new enums or not :D

desktop/src/app.rs
Dinnerbone: I'd say let's keep current behaviour and consider coming back to try and fix that issue separately

desktop/src/app.rs
torokati44: Fair point, it's fine like this. (Although, _how many_ additional "named" buttons could a mouse gain?)

desktop/src/app.rs
Dinnerbone: Added!

desktop/src/app.rs
Dinnerbone: Added comment

desktop/src/app.rs
Dinnerbone: Decided to go for `_` :D that's a fair point and I can't argue it.

torokati44 (approved):

torokati44:
It's somewhat unfortunate that we have to switch to an `egui` fork, but it is what it is...

PR #14715, opened 2024-01-12
web: Properly panic when loading invalid SWF files (fix #14665)

### LINKED ISSUE #14665: Panic screen sometimes does not appear when it should on the extension's SWF player page
### Describe the bug

On the the Ruffle extension's internal player page, some error conditions do not make the panic screen appear, even though they should. Instead, either the loading spinner animation plays forever, or a blank screen appears, either of which is very confusing for users.
- If an invalid SWF is loaded via the "Select File" button, Ruffle encounters an error while trying to generate the panic message, and so a blank screen is shown instead of the panic screen:
`Uncaught (in promise) TypeError: this.swfUrl is undefined`
- If a non-SWF URL is loaded via the URL box, the loading spinner animation plays forever and the panic screen does not appear. That's despite this error appearing in the console:
`ERROR web/src/navigator.rs:377 Asynchronous error occurred: Invalid SWF: Couldn't read SWF: Invalid data: Invalid SWF`

### Expected behavior

The panic screen should appear.

### Content Location

Example of a non-SWF URL: `https://www.miniplay.com/game/renegade-racing`

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

_No response_

### Additional information

_No response_

danielhjacobs:
Note: The error when loading a page which is not an SWF is the same as the error when loading an SWF that does not exist for the purpose of code simplicity. This can be changed if desired.

### DIFF
--- a/core/src/backend/ui.rs
+++ b/core/src/backend/ui.rs
@@ -81,7 +81,7 @@ pub trait UiBackend: Downcast {
     /// Displays a message about an error during root movie download.
     /// In particular, on web this can be a CORS error, which we can sidestep
     /// by providing a direct .swf link instead.
-    fn display_root_movie_download_failed_message(&self);
+    fn display_root_movie_download_failed_message(&self, _invalid_swf: bool);
 
     // Unused, but kept in case we need it later.
     fn message(&self, message: &str);
@@ -275,7 +275,7 @@ impl UiBackend for NullUiBackend {
         Ok(())
     }
 
-    fn display_root_movie_download_failed_message(&self) {}
+    fn display_root_movie_download_failed_message(&self, _invalid_swf: bool) {}
 
     fn message(&self, _message: &str) {}
 
--- a/core/src/loader.rs
+++ b/core/src/loader.rs
@@ -838,7 +838,15 @@ impl<'gc> Loader<'gc> {
                 .map(|u| u.to_string())
                 .unwrap_or(swf_url);
 
-            let mut movie = SwfMovie::from_data(&response.body, spoofed_or_swf_url, None)?;
+            let mut movie =
+                SwfMovie::from_data(&response.body, spoofed_or_swf_url, None).map_err(|error| {
+                    player
+                        .lock()
+                        .unwrap()
+                        .ui()
+                        .display_root_movie_download_failed_message();
+                    error
+                })?;
             on_metadata(movie.header());
             movie.append_parameters(parameters);
             player.lock().unwrap().mutate_with_update_context(|uc| {
--- a/core/src/loader.rs
+++ b/core/src/loader.rs
@@ -821,7 +821,7 @@ impl<'gc> Loader<'gc> {
                     .lock()
                     .unwrap()
                     .ui()
-                    .display_root_movie_download_failed_message();
+                    .display_root_movie_download_failed_message(false);
                 error.error
             })?;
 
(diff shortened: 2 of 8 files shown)


n0samu (approved):

Lord-McSweeney:
I think having the message be the same for invalid SWF vs SWF download fail is confusing, but I'm not sure how differentiating would be implemented- adding a new UI function feels like overkill.

danielhjacobs:
> I think having the message be the same for invalid SWF vs SWF download fail is confusing, but I'm not sure how differentiating would be implemented- adding a new UI function feels like overkill.

Maybe this can take a boolean parameter `invalidSwf: boolean`. All other places calling it can switch to `display_root_movie_download_failed_message(false)`

https://github.com/ruffle-rs/ruffle/blob/3c115238fdf05a6965fde036f282d19730cfb07f/web/packages/core/src/ruffle-player.ts#L2244

Then before this line:
https://github.com/ruffle-rs/ruffle/blob/1b5eb2e8e258a81767652e3d6dfd8786284395e3/web/packages/core/src/ruffle-player.ts#L2277

I can add this:
```ts
} else if (invalidSwf) {
    error.ruffleIndexError = PanicError.InvalidSwf;
```

And before this line:
https://github.com/ruffle-rs/ruffle/blob/1b5eb2e8e258a81767652e3d6dfd8786284395e3/web/packages/core/src/ruffle-player.ts#L38

I can add `InvalidSwf,`

And before this:
https://github.com/ruffle-rs/ruffle/blob/1b5eb2e8e258a81767652e3d6dfd8786284395e3/web/packages/core/src/ruffle-player.ts#L2114

I can add:
```ts
            case PanicError.InvalidSwf:
                errorBody = textAsParagraphs("error-invalid-swf");
                errorFooter = this.createErrorFooter([new PanicLinkInfo()]);
                break;
```

And then I can add a message for `error-invalid-swf` before this:
https://github.com/ruffle-rs/ruffle/blob/1b5eb2e8e258a81767652e3d6dfd8786284395e3/web/packages/core/texts/en-US/messages.ftl#L33-L36

danielhjacobs:
I thought of all that before making this PR, which is why my initial comment said I did it the way I did "for the purpose of code simplicity."

Lord-McSweeney (approved):

core/src/backend/ui.rs
```diff
@@ -81,7 +81,7 @@ pub trait UiBackend: Downcast {
     /// Displays a message about an error during root movie download.
     /// In particular, on web this can be a CORS error, which we can sidestep
     /// by providing a direct .swf link instead.
-    fn display_root_movie_download_failed_message(&self);
+    fn display_root_movie_download_failed_message(&self, _invalid_swf: bool);
```
torokati44, line 84: Oh noo, the "boolean trap"! :sweat_smile: 

https://ariya.io/2011/08/hall-of-api-shame-boolean-trap

> it’s almost invariably a mistake to add a bool parameter to an existing function

Okay, don't take this that seriously, I'm also prone to falling victim to it - usually purposefully.
It's probably fine here.

torokati44: I think the least bad alternative would be an enum-ish parameter. Maybe named integer constants, or string literals. But given how any reasonable IDE nowadays will tell you the name of the parameter as which you are passing that `false` immediately on hover anyway, there's not much point.

Dinnerbone (approved):

PR #14721, opened 2024-01-13
avm2: Add weird parse-as-uint special case to Object keys

### LINKED ISSUE #14299: Monopoly: Bot can't buy a house
### Describe the bug

During the game, bots can not buy the houses. I tried the game in Flashpoint and it is working correctly. Unfortunately, I don't see any error in Ruffle.

image

But as you can see in the game log on the picture, the bot (Shoe) spends $0 to build a house "GO", which is wrong. "GO" is the name of the starting point, not the house. I see the same log every time they try to build the house. For some reason, this log is always shown twice (also the sound).
image
You can also see the money icon on the starting point, it should be on the house.
This log is called from MonopolyGame script (row 2873), so you can start here.
`case Board.BOARD_STATUS_SELECT_BUY`

Unfortunately, you have to play for a while to reproduce it. I can help with testing.
Probably the fastest way is to play with one bot. Set difficulty to max. If the bot has the same properties from the same category(color), he will buy the house. You can help him if you buy property and trade with him for a few bucks.

### Expected behavior

bots can build the houses

### Content Location

Game URL - https://www.gameflare.com/online-game/monopoly-online/
SWF 
[monopoly.zip](https://github.com/ruffle-rs/ruffle/files/13540197/monopoly.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

_No response_

Aaron1011:
When handling dynamic properties, avmplus will always try to parse the string key name as a uint. If it succeeds, then the key will be stored internally as a integer (via Atom), which is observable by property iteration. The intention appears to have been to support `obj[25] = someVal`, but it causes `obj["25"]` and `obj[25]` to map to the same key (though iterating over the object's keys will always produce a `number`).

Fixes #14299, #12387


core/src/avm2/dynamic_map.rs
```diff
@@ -14,8 +14,13 @@ pub struct DynamicProperty<V> {
 
 #[derive(Eq, PartialEq, Hash, Copy, Clone, Collect)]
 #[collect(no_drop)]
-pub enum StringOrObject<'gc> {
+pub enum DynamicKey<'gc> {
     String(AvmString<'gc>),
+    // When the name parses as a posiive integer, we use that integer as the key
```
Lord-McSweeney, line 19: "posiive" typo

Lord-McSweeney (approved):

PR #14722, opened 2024-01-13
avm2: Stub all remaining TextField methods and properties

Aaron1011:


core/src/avm2/globals/flash/text/TextField.as
```diff
@@ -154,5 +156,45 @@ package flash.text {
...
+        }
+
+        public function getParagraphLength(charIndex:int):int {
+            stub_method("flash.text.TextField", "getParagraphLength");
+            return 0;
+        }
+
+        public static function isFontCompatible(fontName:String, fontStyle:String):Boolean {
+            stub_method("flash.text.TextField", "isFontCompatible");
+            return true;
```
Lord-McSweeney, line 192: We should probably return `false` here

Aaron1011: I think it's better to return true - if we return `false` and a SWF checks for this every font it uses, it might end up thinking that there aren't any available fonts.

Lord-McSweeney (approved):

PR #14749, opened 2024-01-14
core: Implement DefineFont4

evilpie:
This more  a proof-of-concept. It kind of works with Mega Miner (#12008), but the there are still obvious rendering differences.

image


core/src/display_object/movie_clip.rs
```diff
@@ -3962,29 +3962,18 @@ impl<'gc, 'a> MovieClipData<'gc> {
...
-                id: font.id,
-                name: SwfStr::from_bytes(b"<Unknown Font4>"),
-                language: swf::Language::Unknown,
-                layout: None,
-                glyphs: Vec::new(),
-                flags: FontFlag::empty(),
-            },
+            font,
             reader.encoding(),
             FontType::EmbeddedCFF,
```
Lord-McSweeney, line 3971: This should always be `EmbeddedCFF`, can you hard-code it to be that in `from_font4_tag`?

Lord-McSweeney (approved):

core/src/library.rs
```diff
@@ -564,8 +564,13 @@ impl<'gc> Library<'gc> {
...
-                {
+                if let Ok(font) = Font::from_font_file(
+                    gc_context,
+                    descriptor,
+                    Cow::Owned(data),
+                    0,
+                    FontType::Device,
+                ) {
                     let name = font.descriptor().name().to_owned();
                     info!("Loaded new device font \"{name}\" from file");
```
Dinnerbone, line 575: This message is triggering but it's not actually a device font

evilpie: It is? We are inside `register_device_font`.

Dinnerbone: But DefineFont4 is embedding a font

evilpie: Correct and for DefineFont4 we use `register_character`.

Dinnerbone: Spooky. Loading this branch and the test I sent you - but deleting the Exo2 font from my computer - I see this in Ruffle's log:

```
2024-01-19T17:58:56.800120Z  INFO run_frame: ruffle_desktop::backends::ui: Loading device font "Exo2-Regular" for "Exo 2" (italic: false, bold: false)
2024-01-19T17:58:56.800415Z  INFO run_frame: ruffle_core::library: Loaded new device font "Exo 2" from file
```

It shouldn't be able to find that font (it physically doesn't exist) and it shouldn't even be looking for that font in the first place (it's embedded)

evilpie: We might be looking up the font as a device font, because `display_object.set_is_device_font` happens after `display_object.set_text_format`, so first relayout and assume the font is a device font?

Dinnerbone: Makes sense!

evilpie: Yep. Reordering fixed this.

Lord-McSweeney:
I think I know the reason the text is wrapping, I can resolve that issue after this PR is merged

tests/tests/swfs/visual/definefont4/test.toml
```diff
@@ -0,0 +1,7 @@
+num_frames = 5
+
+[image_comparisons.output]
+tolerance = 1
+
+[player_options]
+with_renderer = { optional = true, sample_count = 1 }
```
Dinnerbone, line 7: optional false - but otherwise approved!

Dinnerbone (approved):

PR #14750, opened 2024-01-14
avm2: Implement XML.setLocalName

sleepycatcoding:


core/src/avm2/globals/XML.as
```diff
@@ -255,6 +256,11 @@ package {
             return self.AS3::setChildren(value);
         }
 
+        prototype.setLocalname = function(name:*):void {
```
torokati44, line 259: typo: `setLocalname` -> `setLocalName`?

core/src/avm2/globals/xml.rs
```diff
@@ -830,6 +830,42 @@ pub fn set_children<'gc>(
     Ok(xml.into())
 }
 
+// ECMA-357 13.4.4.34 XML.prototype.setLocalName ( name )
```
torokati44, line 833: formatting nit: `in other places I see this written as `XML.prototype.setLocalName (name)`

core/src/avm2/globals/XML.as
sleepycatcoding: Oh right :sweat_smile:, fixed!

core/src/avm2/globals/xml.rs
```diff
@@ -830,6 +830,42 @@ pub fn set_children<'gc>(
...
+pub fn set_local_name<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Object<'gc>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let xml = this.as_xml_object().unwrap();
```
torokati44, line 839: I suppose this `unwrap` here can never fail, right?

core/src/avm2/globals/xml.rs
sleepycatcoding: Well, we have instances of both, for example: https://github.com/ruffle-rs/ruffle/blob/a638715bd87d818d134a70659e7124234b992447/core/src/avm2/globals/xml.rs#L489

torokati44: Okay then, never mind.

core/src/avm2/globals/xml.rs
sleepycatcoding: Since the method is defined on the XML object I don't think it can fail, but correct me if I am wrong. And we have other XML methods that use `unwrap` in the same way.

torokati44: No, that was my intuition as well.

torokati44 (approved):
Thank you!

Lord-McSweeney:
`globals::xml::set_name` currently manually constructs error 1117, can you switch that impl to use the new `make_error_1117`?

Lord-McSweeney (approved):

PR #14766, opened 2024-01-15
ci: Remove PR tracking action

Dinnerbone:
Turns out PR-responding actions won't run with secrets on forks.

Back to the drawing board. :(


torokati44 (approved):

Dinnerbone:
It was to be able to categorise them as we go, so we don't need to try and do that retroactively later when it's blog post time.

Labels can work but can get out of hand, not as clean as a board you can drag things around in :(

PR #14790, opened 2024-01-17
avm2: Don't hold GcCell read ref open for Loader.loadBytes

Aaron1011:
We were holding this open while dispatching an event, leading to a panic if anything attempted to modify the ByteArray.


Lord-McSweeney:
Can you add a test?

Aaron1011:
I updated an existing test to print out `bytearray.position` from event handlers.

Lord-McSweeney:
Are you sure that the test panics on current master? The panic issue was a result of a mutable borrow (`readUnsignedInt` or whatever it was needs to update `position`) after the immutable borrow in the `load_bytes`, but the test only calls `bytearray::get_position`, which immutably borrows the bytearray.

torokati44:
I had the same concerns, so I downloaded the updated test SWF, and started it in the desktop player and the web demo, got an error in both.

Aaron1011:
Yes, I confirmed that this errors on `master`. `bytearray.position` ends up mutating bytearray in order to attach the bound method.

Lord-McSweeney (approved):

Issue #14101, opened 2023-11-21
Regression: Shopaholic Hawaii - can't start the first level


MartySVK:
### Describe the bug

When I try to start the first level with OK button, I get this error:

```Error dispatching event EventObject(EventObject { type: "SHOPSTREET", class: zlong.Hawaii_Events::Hawaii_event, ptr: 0xb4a9860 }) to handler FunctionObject(FunctionObject { ptr: 0xac8ae88 }) : TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: removeEvents)	at zlong.hawaii_view::Hawaii_creditCard/removeEvent()	at zlong.hawaii_view::Hawaii_creditCard/destroy()	at zlong.manager::Hawaii_UIManager/clearScreen()	at zlong.manager::Hawaii_UIManager/toStreet()	at flash.events::EventDispatcher/flash::events::EventDispatcher::dispatchEvent()	at zlong.hawaii_view::Hawaii_DailyChallenge/toChooseChallenge()```

Last working build is 2023-11-12
SWF version 14

### Expected behavior

start the level

### Content Location

GAME URL - https://www.gameflare.com/online-game/shopaholic-hawaii/
SWF URL - https://data.gameflare.com/games/6633/7NoWsdgnMZbMlv.swf

### Affected platform

Desktop app

### Operating system

Windows 10

MartySVK:
looks like the problem is @Aaron1011 's PR https://github.com/ruffle-rs/ruffle/pull/13966

Lord-McSweeney:
From Aaron, in #14007:
> I think we need to implement the 'default namespace' logic for non-qname Multinames, which in this case will end up appending the public namespace to the incoming Multiname: https://github.com/adobe/avmplus/blob/858d034a3bd3a54d9b70909386435cf4aec81d21/core/Toplevel.cpp#L287

evilpie:
Fixed by #14801.

Issue #14719, opened 2024-01-12
Regression: Dolphin Pop


desertking:
### Describe the bug

This game used to run fine. Now it shows a black screen after selecting a level. sound still running thou.

### Expected behavior

work again

### Content Location

https://www.spiele-umsonst.de/azad/downloads/flashgames/DolphinPop.swf

### Affected platform

Online demo

### Operating system

All

### Browser

All

Lord-McSweeney:
The first error (at the title screen) is
```
Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x555f4ed367b0 }) to handler FunctionObject(FunctionObject { ptr: 0x555f4e6ae250, name: Ok("com.zlonggames.view::AbstractGeneralView/changeState()") }) : TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: http://adobe.com/AS3/2006/builtin::split)
	at com.zlonggames.control.coreconcrete.fromCoreApp.interfaces::TimerManage/initBestTime()
	at com.zlonggames.control.coreconcrete.fromCoreApp.interfaces::TimerManage/init()
	at com.zlonggames.view.level::LevelInGame/initDisplay()
	at com.zlonggames.view.level::LevelInGame()
	at com.zlonggames.view.menu::GatePassMenu/nextUIScene()
	at com.zlonggames.view::AbstractGeneralView/changeState()
```

n0samu:
Another regression caused by #13966

evilpie:
Fixed by #14801.

Issue #14596, opened 2024-01-03
Sandbox Hero no longer works


n0samu:
### Describe the bug

[Sandbox Hero](https://www.kongregate.com/games/kongregate/sandbox-hero) stopped working in Ruffle nightly 2023-11-13. Before then the main menu loaded fine, but now it shows only a black screen after clicking the Play button. And it seems to go into an infinite loop and uses more and more memory over time.

The game is sitelocked and needs to be run with `--spoof-url "http://chat.kongregate.com"` if run locally.

### Expected behavior

The game should load.

### Content Location

Page URL: https://www.kongregate.com/games/kongregate/sandbox-hero
SWF URL: https://chat.kongregate.com/gamez/0019/2277/live/SandboxHero.swf

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

Same problem as #14593. Another regression caused by #13966.

n0samu:
Fixed by #14801

Issue #14593, opened 2024-01-03
Pirateers 2 no longer works


n0samu:
### Describe the bug

[Pirateers 2](https://www.kongregate.com/games/labugames/pirateers-2) stopped working in Ruffle nightly 2023-11-13. Before then the main menu loaded fine, but now it shows only a black screen after the Armor Games intro. And it seems to go into an infinite loop and uses more and more memory over time.

### Expected behavior

The game should load.

### Content Location

Page URL: https://www.kongregate.com/games/labugames/pirateers-2
SWF URL: https://chat.kongregate.com/gamez/0021/9281/live/Pirateers2.swf

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

This seems to be another regression caused by #13966.

n0samu:
Fixed by #14801

PR #14801, opened 2024-01-17
avm2: Assume we have the default public namespace when matching XML names

evilpie:
I am optimistic that this will fix all regression introduced by https://github.com/ruffle-rs/ruffle/pull/13966. I am not sure how to test this.


Dinnerbone:
Copy the pcode from one of the swfs maybe? It seems a good test to have.

n0samu:
Fixes  #14101, #14206, #14593, #14596, #10651, and #14719! Thank you so much!

n0samu:
@evilpie Here is a test for you loosely based on #14719: [XMLTest_14719.zip](https://github.com/ruffle-rs/ruffle/files/13973172/XMLTest_14719.zip)
It traces `name: ` without your PR and `name: ruffle` with it. The game from #14719 has a super weird `getproperty` instruction in `com.zlonggames.model.MapData get firstGateId()` which looks maybe manually written/edited:
```
				pushstring "gate-id"
				getproperty MultinameLA([PackageInternalNs("com.zlonggames.model")])
```
I'm guessing they wanted to use the `.@` syntax to get the `gate-id` property but it didn't work when written in AS because of the dash. Anyway, for my test I just wrote a function that gets an attribute of an XML:
```as
function getProjectName(xmlObj:XML) {
	return xmlObj.@name;
}
```
And then I edited the `getproperty` instruction to remove all of the namespaces except `PackageInternalNs("XMLTest_14719_fla")`. But I noticed it even works if I remove *all* namespaces from the instruction.

evilpie:
Thanks @n0samu, I added the test.

PR #14802, opened 2024-01-17
desktop: Fix doubling inputted characters

kjarosh:
Fix a bug introduced by f65060e8 (#14703). The text input event was triggered two times: at key press and release. This patch makes sure that text input is triggered only on key press.


Dinnerbone (approved):
Thank you!

torokati44 (approved):

PR #14807, opened 2024-01-17
tests: Don't set logger in test framework

Dinnerbone:
It's the responsibility of the caller to do it how they want

### DIFF
--- a/tests/Cargo.toml
+++ b/tests/Cargo.toml
@@ -29,6 +29,9 @@ walkdir = "2.4.0"
 anyhow = "1.0.79"
 image = { version = "0.24.8", default-features = false, features  = ["png"] }
 futures = "0.3.30"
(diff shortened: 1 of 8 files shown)


tests/Cargo.toml
```diff
@@ -29,6 +29,9 @@ walkdir = "2.4.0"
 anyhow = "1.0.79"
 image = { version = "0.24.8", default-features = false, features  = ["png"] }
 futures = "0.3.30"
+env_logger = "0.10.1"
+tracing = { workspace = true}
```
sleepycatcoding, line 33: This is missing a space, but it was like that before, oh well.

Dinnerbone: Fixed

sleepycatcoding (approved):

PR #14820, opened 2024-01-18
Use inner mutability instead of `GcCell` in `Avm1Button` and `Avm2Button`

moulins:
Slowly getting rid of more `GcCell` usages.

### DIFF
--- a/core/src/display_object/avm2_button.rs
+++ b/core/src/display_object/avm2_button.rs
@@ -283,21 +283,35 @@ impl<'gc> Avm2Button<'gc> {
     pub fn set_state(self, context: &mut UpdateContext<'_, 'gc>, state: ButtonState) {
         self.invalidate_cached_bitmap(context.gc());
 
-        if let Some(state) = self.0.up_state.get() {
-            state.set_parent(context, None);
-        }
-        if let Some(state) = self.0.over_state.get() {
-            state.set_parent(context, None);
-        }
-        if let Some(state) = self.0.down_state.get() {
-            state.set_parent(context, None);
-        }
-        if let Some(state) = self.0.hit_area.get() {
+        for state in self.all_state_children(false) {
             state.set_parent(context, None);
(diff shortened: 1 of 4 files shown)


Dinnerbone (approved):
LGTM but I'm mostly just trusting the fact that you know this stuff :D

core/src/display_object/avm2_button.rs
```diff
@@ -283,23 +283,37 @@ impl<'gc> Avm2Button<'gc> {
...
         }
         if let Some(state) = self.get_state_child(state.into()) {
             state.set_parent(context, Some(self.into()));
         }
     }
 
+    /// Get all the display objects representing button states.
+    fn all_state_children(
+        &self,
+        weird_order: bool,
```
torokati44, line 297: I was going to say that this is not the most fortunate naming for a parameter, but:
 - I can't think of anything better...
 - I can see `weird_framescript_order` has been there before, and this is not worse...

torokati44 (approved):
I'm not seeing anything obviously wrong, but honestly - same as above: I trust @moulins with things like this, which may be a bit over my head.

moulins:
All of this is actually fairly mechanical; `GcCell<'gc, T>` is basically `Gc<'gc, RefCell<T>>`, so the changes do not touch the actual logic and mostly involve "splitting" the interiorly-mutable parts of the type into (mostly) one cell type for each field. This allows two things:
1. We can remove write barriers (and so the need for a `&Mutation<'gc>` parameters) when we modify `'static` data (e.g. boolean flags)
2. We reduce the need for long-lived runtime borrows of `RefCell`s: this reduces the risk of borrow panics (but we still need to be careful of partially-modified states when calling back into unknown AVM bytecode).

torokati44:
> All of this is actually fairly mechanical; 

Yeah, I assumed that much. 😅

Thank you for the explanation!

PR #14827, opened 2024-01-18
avm2: Fix possible overflow in address calculations

adrian17:
Should fix the test failure @Dinnerbone noticed on wasm.
(I didn't get to test it myself due to temporary pc difficulties)


Dinnerbone (approved):
Works! Thank you!

PR #14829, opened 2024-01-19
avm2: Add a bunch more constants

torokati44:
Browsing https://ruffle.rs/compatibility/avm2, these stood out as random missing pieces.
Easy wins for our API percentage.
_Goodhart's law has not been broken during the making of this PR._

### DIFF
--- a/core/src/avm2/globals/flash/desktop/ClipboardFormats.as
+++ b/core/src/avm2/globals/flash/desktop/ClipboardFormats.as
@@ -6,8 +6,20 @@
 package flash.desktop
 {
 
+    // The AS3 docs specify various API versions for the different members,
+    // as also indicated by their comments. The docs, however, are lying.
+    [API("662")]
     public class ClipboardFormats
     {
+        // Image data (AIR only).
+        public static const BITMAP_FORMAT:String = "air:bitmap";
+
+        // An array of files (AIR only).
+        public static const FILE_LIST_FORMAT:String = "air:file list";
+
+        // File promise list (AIR only).
+        public static const FILE_PROMISE_LIST_FORMAT:String = "air:file promise list";
+
         // HTML data.
         public static const HTML_FORMAT:String = "air:html";
 
@@ -16,6 +28,9 @@ package flash.desktop
 
         // String data.
         public static const TEXT_FORMAT:String = "air:text";
+
+        // A URL string (AIR only).
+        public static const URL_FORMAT:String = "air:url";
 
     }
 }
--- a/core/src/avm2/globals/flash/events/IOErrorEvent.as
+++ b/core/src/avm2/globals/flash/events/IOErrorEvent.as
@@ -3,6 +3,11 @@ package flash.events {
         // Defines the value of the `type` property of an `ioError` event object.
         public static const IO_ERROR:String = "ioError";
 
+        // These next three are undocumented.
+        public static const NETWORK_ERROR:String = "networkError";
+        public static const DISK_ERROR:String = "diskError";
+        public static const VERIFY_ERROR:String = "verifyError";
+
         // The standardErrorIoError event is dispatched when an error occurs while
         // reading data from the standardError stream of a NativeProcess object.
         [API("668")]
--- a/core/src/avm2/globals/flash/events/IOErrorEvent.as
+++ b/core/src/avm2/globals/flash/events/IOErrorEvent.as
@@ -1,6 +1,22 @@
 package flash.events {
     public class IOErrorEvent extends ErrorEvent {
+        // Defines the value of the `type` property of an `ioError` event object.
         public static const IO_ERROR:String = "ioError";
+
+        // The standardErrorIoError event is dispatched when an error occurs while
+        // reading data from the standardError stream of a NativeProcess object.
+        [API("668")]
+        public static const STANDARD_ERROR_IO_ERROR:String = "standardErrorIoError";
+
+        // The standardInputIoError event is dispatched when an error occurs while
+        // writing data to the standardInput of a NativeProcess object.
+        [API("668")]
+        public static const STANDARD_INPUT_IO_ERROR:String = "standardInputIoError";
+
+        //The standardOutputIoError event is dispatched when an error occurs while
+        // reading data from the standardOutput stream of a NativeProcess object.
+        [API("668")]
+        public static const STANDARD_OUTPUT_IO_ERROR:String = "standardOutputIoError";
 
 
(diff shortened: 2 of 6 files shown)


core/src/avm2/globals/flash/desktop/ClipboardFormats.as
```diff
@@ -8,6 +8,18 @@ package flash.desktop
 
     public class ClipboardFormats
     {
+        // Image data (AIR only).
+        [API(661)]
```
Lord-McSweeney, line 12: The way most other API-versioning metadata is, this would be `[API("661")]`, with quotes.

torokati44: Oh, you're right. Fixed.
Speaking of, maybe we should fix it instead, and do it without quotes everywhere? Just for simplicity. And that would make it "feel" more like an integer, which I think it is. Would that work?

core/src/avm2/globals/flash/ui/GameInputDevice.as
```diff
@@ -1,4 +1,9 @@
 package flash.ui {
     public final class GameInputDevice {
+        // Specifies the maximum size for the buffer used to cache sampled
+        // control values. If `startCachingSamples` returns samples that
+        // require more memory than you specify, it throws a memory error.
+        [API("689")]
```
Lord-McSweeney, line 6: This entire class should be `[API("688")]`

core/src/avm2/globals/flash/desktop/ClipboardFormats.as
```diff
@@ -8,6 +8,18 @@ package flash.desktop
 
     public class ClipboardFormats
```
Lord-McSweeney, line 9: From some testing, nothing in this class is version gated- the class itself, however, is version-gated to SWFv10 and above

core/src/avm2/globals/flash/ui/GameInputDevice.as
torokati44: Right - but not 689? It looks AIR-only to me.

core/src/avm2/globals/flash/events/StageVideoEvent.as
```diff
@@ -7,9 +7,27 @@ package flash.events
     
     public class StageVideoEvent extends Event
```
Lord-McSweeney, line 8: Nothing in this class is version-gated; however, the class itself is version-gated (docs say FP 10.2, playerglobals says FP 10.1)

core/src/avm2/globals/flash/ui/GameInputDevice.as
Lord-McSweeney: I can access it in normal FP

torokati44:
"Easy win" - haha, sure. If only there was such a thing. With the docs **lying** all the time!!

core/src/avm2/globals/flash/ui/GameInputDevice.as
danielhjacobs: image

evilpie:
AFAIK everything on https://ruffle.rs/compatibility/avm2 is actually available to Flash and can't be AIR-only.

torokati44:
That would make sense - isn't that how we got the list in the first place? By a dumper SWF?

Lord-McSweeney:
There are some AIR-only fields and functions in the version of `playerglobal.swc` that I have

core/src/avm2/globals/flash/ui/GameInputDevice.as
torokati44: Thanks, fixed!

core/src/avm2/globals/flash/desktop/ClipboardFormats.as
torokati44: Thank you, fixed!

torokati44:
Thank you for checking, and the suggestions, @Lord-McSweeney, @danielhjacobs, and @evilpie!

I've fixed all review comments. According to the logic above, should `IOErrorEvent` and `KeyLocation` also be changed?

Lord-McSweeney (approved):
The rest of it looks fine to me, thanks for the fixes!

evilpie (approved):

core/src/avm2/globals/flash/events/IOErrorEvent.as
```diff
@@ -1,7 +1,23 @@
 package flash.events {
     public class IOErrorEvent extends ErrorEvent {
```
evilpie, line 2: https://ruffle.rs/compatibility/avm2 also has `DISK_ERROR`, `NETWORK_ERROR` and `VERIFY_ERROR`, but they are not listed in Flash documentation.

torokati44: Huh, those are actually the reason I touched this class. Just didn't notice them missing from the docs... :sweat_smile:

torokati44: Well they are supposed to be `[Inspectable(environment="none")]` - whatever that means.
But I'm not seeing any way for us to denote or enforce that...

torokati44: I have added them. Thanks!

Lord-McSweeney: I'm pretty sure `[Inspectable(environment="none")]` is metadata that the Flash IDE uses

core/src/avm2/globals/flash/events/IOErrorEvent.as
```diff
@@ -1,7 +1,28 @@
...
+        // Defines the value of the `type` property of an `ioError` event object.
         public static const IO_ERROR:String = "ioError";
 
+        // These next three are undocumented, and are not supposed to be inspectable.
```
sleepycatcoding, line 6: Non-blocking, but is this comment a bit outdated since the Inspectable metadata was removed?

sleepycatcoding (approved):

core/src/avm2/globals/flash/events/IOErrorEvent.as
torokati44: I added this comment when I removed the metadata...
Oh, I have just now realized that the "... is metadata that the Flash IDE uses" comment was not just about it the metadata not being used by Flash Player, but also that it's entire intention doesn't make sense outside of the IDE. So I'll remove that part. Thanks!

PR #14839, opened 2024-01-19
swf: Show scaled value in `swf::FixedN`'s `Debug` implementations

moulins:
This makes the debug output more readable, e.g. in the debug UI filter's list.

Before:
image

After:
image


Dinnerbone (approved):

torokati44:
In some cases it might be surprising that what looks like an ordinary float (even when `dbg!`'d) is actually a `FixedN`, but I can accept that this is just an internal representation / implementation detail thing, and the readable number is the intended meaning.

torokati44 (approved):

Issue #4258, opened 2021-05-06
Support all Flash embed/object attributes


Herschel:
This is a tracking issue for supporting the attributes that could be placed on the Flash embed/object tag:
https://helpx.adobe.com/flash/kb/flash-object-embed-tag-attributes.html

The polyfill should honor these settings. They can be passed along to the Ruffle core via the config block, which also allows the JS API to use them.

In somehwat order of importance:
 [x] `src`/`movie`
 [x] `width`
 [x] `height`
 [x] `allowScriptAccess`: Disable JS calls from AS.
 [ ] `base`: Set the base URL for network requests
 [ ] `wmode`: Ignore all the "gpu" related ones, but `transparent` is often used (#1752)
 [ ] `allowFullScreen`: allow/disallow fullscreen requests from AS
 [ ] `fullScreenAspectRatio`: Control orientation on mobile in fullscreen mode
 [ ] `scale`: Stage.scaleMode
 [ ] `salign`: Stage.align
 [ ] `play`: Stop/play root timeline
 [ ] `loop`: Loop root timeline
 [ ] `menu`: Hide built-in context menu items (Should we make this disable the context menu completely?)
 [ ] `align`: Controls alignment if window is smaller than Flash viewport? Needs investigation
 [ ] `browserZoom`: Controls size changes when browser DPI changes? Needs investigation
 [ ] `quality`: Stage.quality

ousia:
@Herschel, as for AS2, `menu` might be already implemented.

I mean, https://www.mejorar-presentaciones.tk/mejorar-presentaciones-loader.swf contains (besides a `new ContextMenuItem()`):

``` js
new_context = new ContextMenu();
new_context.hideBuiltInItems();
```

Thanks to the work of @adrian17 with #4173, the last standard context menu items (“Enter fullscreen” and “About Ruffle…”) are displayed, besides the `new ContextMenuItem()`.

Flash player also removes the playing context menu items when `.hideBuiltInItems()` is invoked, but leaves “Settings…”, “Global settings…” and “About Adobe Flash Player…”.

Many thanks for your excellent work.

adrian17:
I think the point was about frontend (HTML, JS) API, not AS; I didn't add `menu` support to the embed API.

ousia:
Sorry, in that case I totally missed the point.

sombraguerrero:
Don't mind me, just tagging this for Onda Educa's interest.

danielhjacobs:
[...]

PR #14841, opened 2024-01-19
web: Add allowFullScreen embed/object option (part of #4258)

danielhjacobs:
Note 1:

I realized on `fullscreenchange`, `this.instance?.set_fullscreen` is called (this was to fix escape key behavior IIRC):
https://github.com/ruffle-rs/ruffle/blob/e0f5b8906dc1893416d9d836fd4458c8acc2817b/web/packages/core/src/ruffle-player.ts#L1140-L1142
That same function got called by clicking the context menu item "Enter/Exit Fullscreen", but with my change in `set_display_state` that Rust function doesn't actually full screen the content when `allowFullScreen` is false. So I switched "Enter/Exit Fullscreen" to use the pure JavaScript `setFullscreen` function, which will subsequently trigger a `fullscreenchange` event to call `this.instance?.set_fullscreen` to do the Rust portion of the logic. That way full-screening the content with the context menu always works, but only changes the `displayState` if `allowFullScreen` is true.

Note 2:

I'm not actually throwing the `SecurityError` that is supposed to be thrown when the `displayState` doesn't change in AS3, but that can be a future TODO.

### DIFF
--- a/core/src/display_object/stage.rs
+++ b/core/src/display_object/stage.rs
@@ -103,6 +103,9 @@ pub struct StageData<'gc> {
     /// Whether to prevent movies from the changing the stage alignment
     forced_align: bool,
 
+    /// Whether to allow the stage's displayState to be changed.
+    allow_full_screen: bool,
+
     /// Whether or not a RENDER event should be dispatched on the next render
     invalidated: bool,
 
@@ -174,6 +177,7 @@ impl<'gc> Stage<'gc> {
                 invalidated: false,
                 align: Default::default(),
                 forced_align: false,
+                allow_full_screen: true,
                 use_bitmap_downsampling: false,
                 view_bounds: Default::default(),
                 window_mode: Default::default(),
@@ -331,6 +335,16 @@ impl<'gc> Stage<'gc> {
         self.0.write(context.gc_context).forced_scale_mode = force;
     }
 
+    /// Get whether the Stage's display state can be changed.
+    pub fn allow_full_screen(self) -> bool {
+        self.0.read().allow_full_screen
+    }
+
+    /// Set whether the Stage's display state can be changed.
+    pub fn set_allow_full_screen(self, context: &mut UpdateContext<'_, 'gc>, allow: bool) {
+        self.0.write(context.gc_context).allow_full_screen = allow;
+    }
+
     fn is_fullscreen_state(display_state: StageDisplayState) -> bool {
         display_state == StageDisplayState::FullScreen
             || display_state == StageDisplayState::FullScreenInteractive
@@ -365,6 +379,7 @@ impl<'gc> Stage<'gc> {
     ) {
         if display_state == self.display_state()
             || (Self::is_fullscreen_state(display_state) && self.is_fullscreen())
+            || !self.allow_full_screen()
         {
             return;
         }
--- a/core/src/display_object/stage.rs
+++ b/core/src/display_object/stage.rs
@@ -104,7 +104,7 @@ pub struct StageData<'gc> {
     forced_align: bool,
 
     /// Whether to allow the stage's displayState to be changed.
-    allow_full_screen: bool,
+    allow_fullscreen: bool,
 
     /// Whether or not a RENDER event should be dispatched on the next render
     invalidated: bool,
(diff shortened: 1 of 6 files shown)


core/src/display_object/stage.rs
```diff
@@ -103,6 +103,9 @@ pub struct StageData<'gc> {
     /// Whether to prevent movies from the changing the stage alignment
     forced_align: bool,
 
+    /// Whether to allow the stage's displayState to be changed.
+    allow_full_screen: bool,
```
evilpie, line 107: I think we use `fullscreen` more than `full_screen`.

danielhjacobs: Then I need a `#[serde(rename)]` in `web/src/lib.rs`, since the name of the embed/object attribute is `allowFullScreen`, and the default way `serde` converts camelCase to snake_case makes this `allow_full_screen`.

core/src/display_object/stage.rs
```diff
@@ -365,6 +379,7 @@ impl<'gc> Stage<'gc> {
     ) {
         if display_state == self.display_state()
             || (Self::is_fullscreen_state(display_state) && self.is_fullscreen())
+            || !self.allow_full_screen()
```
evilpie, line 382: I think we should check for `is_fullscreen_state(display_state)` (probably move into the expression above)

core/src/display_object/stage.rs
evilpie: But we don't need to follow the `param` name, right? We already use `backgroundColor` instead of `bgColor`. Maybe it's not worth it.

core/src/display_object/stage.rs
danielhjacobs: Are you sure you can change the display state to Normal when `allowFullScreen` is not true in Adobe Flash Player? <https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/display/Stage.html#displayState> says a SecurityError is thrown when trying to set the display state "if the param or embed HTML tag's allowFullScreen attribute is not set to true", mentioning nothing about that only applying when setting it to a fullscreen state.

core/src/display_object/stage.rs
evilpie: Or we keep the name for loadOptions as well and just rename those internal ones?

danielhjacobs: True, I looked at how "bgcolor" does it. I didn't think of changing the name right after `getOptionString("allowFullScreen");` https://github.com/ruffle-rs/ruffle/blob/e0f5b8906dc1893416d9d836fd4458c8acc2817b/web/packages/core/src/ruffle-player.ts#L2483

core/src/display_object/stage.rs
danielhjacobs: I haven't tested it myself.

evilpie: Okay, I don't know either.

core/src/display_object/stage.rs
danielhjacobs: Changed the functions.

core/src/display_object/stage.rs
danielhjacobs: Ah, just read a bit more: "Full-screen mode is initiated in response to a mouse click or key press by the user; the movie cannot change `Stage.displayState` without user input." Since on web no scripts can make the SWF start in full screen without input, letting you change the `displayState` to normal would do nothing since the `displayState` can never start as anything else and can't be changed to full screen.

evilpie:
This is actually not really my area of expertise, so maybe a second reviewer would be good.

evilpie (approved):

core/src/display_object/stage.rs
danielhjacobs: I did `window.addEventListener("click", () => document.querySelector("embed, object").requestFullscreen())` in Pale Moon on http://new.weedtowonder.org/domestication.html. The content's full screen and I can use escape to exit full screen, but the real question is, did the `displayState` of the stage change? I don't have a good test SWF for this so I don't know. Ideally I'd have an SWF with a button that displays the current displayState when clicked, but I don't have Adobe Animate so making tests is a PITA.

danielhjacobs: [displayStateAvm1.swf.zip](https://github.com/ruffle-rs/ruffle/files/13994943/displayStateAvm1.swf.zip)

Tested this and I can't seem to cause the displayState to change without using AS, so I think my logic here is fine.

torokati44 (approved):
Thank you!

PR #14860, opened 2024-01-20
web: Support DefineFont4 in fontSources

evilpie:


Dinnerbone (approved):
We should also add ttfs here maybe

PR #14863, opened 2024-01-20
Update vite to 5.0.12 in `package.json`

tsunamistate:
Renovate PR to update `vite` to 5.0.12 (https://github.com/ruffle-rs/ruffle/pull/14851) updated only `package-lock.json`, but not `package.json`

This causes `vite`  inside `package-lock.json` to roll back to 5.0.10 after `npm install`. This PR fixes that :)


torokati44:
Thanks!

It seems DependaBot would have done better: https://github.com/ruffle-rs/ruffle/pull/14850/files

PR #14872, opened 2024-01-21
tests: More improvements to make them runnable on web

Dinnerbone:
Commit log explains the changes out, the full diff is a bit messy looking I guess :D

With these changes, tests run absolutely fine (though slow) on web. That's in a pending PR though as I'm still working on the actual web UI.

Main changes are:
- Testing is no longer just "run test", now it creates a runner and tick/test/tick/test until finished. This way, wasm can run tests that need async.
- Slightly less panic happy (panics on web are sad)
- Only one renderer per environment; the application running the tests can just run them again with a new environment

### DIFF
--- a/tests/framework/src/environment.rs
+++ b/tests/framework/src/environment.rs
@@ -14,24 +14,22 @@ pub trait Environment {
         false
     }
 
-    /// Creates any amount of render backends for a new test run.
+    /// Creates a render backend for a new test run.
     ///
-    /// This method must return both a [RenderBackend] and [RenderInterface] as pairs,
-    /// and they will be treated as pairs for the purposes of this test framework.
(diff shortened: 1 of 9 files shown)


tests/framework/src/environment.rs
```diff
@@ -14,24 +14,22 @@ pub trait Environment {
         false
     }
 
-    /// Creates any amount of render backends for a new test run.
+    /// Creates a render backends for a new test run.
```
torokati44, line 17: nit: "a render backend"

torokati44: (was just passing by, sorry)

Dinnerbone: Fixed!

PR #14879, opened 2024-01-21
avm2: Only lookup own properties in flash.net.getClassByAlias

evilpie:


torokati44:
nit: typo in commit message and PR title: amv2

core/src/avm2/globals/flash/net.as
```diff
@@ -20,13 +20,21 @@ package flash.net {
...
+    internal function _getClassByAlias(name:String):Class {
+        if (!this._aliasToClass.hasOwnProperty(name)) {
             return null;
         }
+
+        return this._aliasToClass[name];
+    }
+
+    public function getClassByAlias(name:String):Class {
+        var klass: Class = this._getAliasByClass(name);
```
Lord-McSweeney, line 33: You probably meant `_getClassByAlias` here, not `_getAliasByClass`

evilpie: :sweat: thanks. I added a test.

Aaron1011 (approved):

PR #14882, opened 2024-01-21
avm2: Integrate avm2 implementation.json generation into new `stub-report` crate

Aaron1011:
This brings Dinnerbone's implementation.json generation into master. You can now run `cargo run -p stub-report implementation.json` to write the implementation report to a file named `implementation.json`.

In a follow-up PR, I'm going to add a Github Action workflow to automatically run this and update the website when it changes.

### DIFF
--- a/core/src/avm2.rs
+++ b/core/src/avm2.rs
@@ -58,6 +58,7 @@ mod qname;
 mod regexp;
 mod scope;
 mod script;
+pub mod specification;
 mod string;
 mod stubs;
 mod traits;
--- a/core/src/avm2.rs
+++ b/core/src/avm2.rs
@@ -58,6 +58,7 @@ mod qname;
 mod regexp;
 mod scope;
 mod script;
+#[cfg(feature = "known_stubs")]
 pub mod specification;
 mod string;
 mod stubs;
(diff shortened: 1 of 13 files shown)


core/src/avm2/specification.rs
```diff
@@ -0,0 +1,525 @@
...
+        let prototype = class_object.prototype();
+        let prototype_base = prototype.base();
+        let prototype_values = prototype_base.values();
+        for (key, value) in prototype_values.as_hashmap().iter() {
+            let name = match key {
+                DynamicKey::String(name) => *name,
+                DynamicKey::Uint(key) => {
+                    AvmString::new_utf8(activation.context.gc_context, key.to_string())
+                }
+                DynamicKey::Object(object) => {
```
evilpie, line 309: I saw this while using this code yesterday and it didn't really make sense to me. I don't think this can happen.

Aaron1011: Probably not, but there's no real harm in having it here.

core/src/avm2.rs
```diff
@@ -56,6 +56,7 @@ mod qname;
 mod regexp;
 mod scope;
 mod script;
+pub mod specification;
```
evilpie, line 59: Should we put everything behind the feature flag instead?

torokati44:
For the record, @Dinnerbone and I have indicated our preference on Discord to having this as a separate executable (like `exporter` and `timedemo`), instead of as part of the regular desktop player.

Aaron1011:
I've moved this into a separate `stub-report` binary crate, and feature-flagged more definitions.

Dinnerbone (approved):

Issue #2178, opened 2020-12-31
Bug report: Click Jogos site


AnaBlue00:
Hello, 
First of all, I would like to congratulate you all on your initiative to keep Flash and try to recover games that will be damaged due to the removal of this extension. I came by this email to report a bug that happened when I entered a brazilian website of games called Click Joos, wich is looking for ways to keep games active and applied Ruffle extension as a try. I tried to play the game (the name of the game is Pack the house - Level 4 - Frenzy kitchen, if this is relevant) and it has not been reproduced. Instead, a screen appeared with the message that something went wrong and that it was not supposed to happen. The function of view error details said the following:
# Error Info
Error name: TypeError
Error message: Failed to execute 'compile' on 'WebAssembly': HTTP status code is not ok
Error stack:
```
TypeError: Failed to execute 'compile' on 'WebAssembly': HTTP status code is not ok
```

# Player Info
Allows script access: true
Player type: Object
SWF URL: https://img.clickjogos.com.br/dl/0/06cb453ce3748b7036b28ee4d1694a5b/src.swf
Param movie: https://img.clickjogos.com.br/dl/0/06cb453ce3748b7036b28ee4d1694a5b/src.swf
Param allowScriptAccess: always
Attribute 0: [object Attr]
Attribute 1: [object Attr]
Attribute 2: [object Attr]
Attribute 3: [object Attr]
Attribute 4: [object Attr]

# Page Info
Page URL: https://www.clickjogos.com.br/jogos/pack-the-house-level-4-frenzy-kitchen

# Browser Info
Useragent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/87.0.4280.88 Safari/537.36
OS: Win32

# Ruffle Info
Ruffle version: 0.1.0
Ruffle source: undefined

I would be very grateful if you could take action on this so that we can maintain our nostalgia. And I apologize if this error occurred due to my computer system and if that is the case, I would appreciate if you let me know what I could do to make it work. 
I'm sorry again for bothering you and thanking you for your work, it really means a lot to me and I'm sure I'm not the only one. 
Have a nice day,
Ana.

ousia:
@AnaBlue00, it seems that https://www.clickjogos.com.br/ hasn’t the server properly configured for _Ruffle_ (as explained in https://github.com/ruffle-rs/ruffle/wiki/Using-Ruffle#configure-webassembly-mime-type).

[...]

PR #14905, opened 2024-01-22
avm2: Stub flash.desktop.Clipboard

evilpie:
This makes the "Get Save Password" button in Phoenotopia work. (#9226)

### DIFF
--- a/core/src/avm2/globals/flash/desktop/Clipboard.as
+++ b/core/src/avm2/globals/flash/desktop/Clipboard.as
@@ -1,0 +1,54 @@
+package flash.desktop {
+    import __ruffle__.stub_getter;
+    import __ruffle__.stub_method;
+    import flash.system.System;
+
+    public class Clipboard {
+        private static var _generalClipboard = new Clipboard();
+
+        public static function get generalClipboard(): Clipboard {
+            return Clipboard._generalClipboard;
+        }
+
+        function Clipboard() {
+            // TODO: This should only be callable in AIR
+        }
+
+        public function get formats(): Array {
+            stub_getter("flash.desktop.Clipboard", "formats");
+            return new Array();
+        }
+
+        public function clear(): void {
+            stub_method("flash.desktop.Clipboard", "clear");
+        }
+
+        public function clearData(format: String): void {
+            stub_method("flash.desktop.Clipboard", "clearData");
+        }
+
+        public function getData(format: String, transferMode: String = ClipboardTransferMode.ORIGINAL_PREFERRED): Object {
+            stub_method("flash.desktop.Clipboard", "getData");
+            return null;
+        }
+
+        public function hasFormat(format: String): Boolean {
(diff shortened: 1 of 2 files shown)


core/src/avm2/globals/flash/desktop/Clipboard.as
```diff
@@ -0,0 +1,52 @@
+package flash.desktop {
+    public class Clipboard {
+        import flash.system.System;
+
+        private static var _generalClipboard  = new Clipboard();
```
Lord-McSweeney, line 5: nit: Double space before the equals

core/src/avm2/globals/flash/desktop/Clipboard.as
```diff
@@ -0,0 +1,52 @@
+package flash.desktop {
+    public class Clipboard {
```
Lord-McSweeney, line 2: This should import `__ruffle__.stub_method`/`getter` up here, and imports (the stub imports and the `flash.system.System` import) should be placed between the `package` block and the `public class` block

Lord-McSweeney (approved):

core/src/avm2/globals/flash/desktop/Clipboard.as
```diff
@@ -0,0 +1,54 @@
...
+
+    public class Clipboard {
+        private static var _generalClipboard = new Clipboard();
+
+        public static function get generalClipboard(): Clipboard {
+            return Clipboard._generalClipboard;
+        }
+
+        function Clipboard() {
+            // TODO: This should only be callable in AIR
```
torokati44, line 14: Why not add the `AIR` annotation here?

torokati44: Also, I guess this is another case of the docs lying then?
https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/desktop/Clipboard.html#Clipboard()

evilpie: The constructor exists in Flash, you just get an error in Flash when calling:
```
Error: Error #2178: The Clipboard.generalClipboard object must be used instead of creating a new Clipboard.
```

I didn't try to figure out how to implement this, because I needed a constructor for `generalClipboard`.

torokati44: Okay, fair enough.

lvyitian: If there's a way to get whether the current runtime is AIR runtime then it can be implemented by:
if(!isAirRuntime)
  if(_generalClipboard)
    throw new Error("Error #2178: The Clipboard.generalClipboard object must be used instead of creating a new Clipboard.");

evilpie: Cool idea.

PR #14907, opened 2024-01-22
avm2: flash.system.Capabilities#playerType should return "Desktop" when the current runtime is AIR runtime(unless the swf is loaded by an HTML page)

lvyitian:
See: https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/system/Capabilities.html#playerType

### DIFF
--- a/core/src/avm2/globals/flash/system/capabilities.rs
+++ b/core/src/avm2/globals/flash/system/capabilities.rs
@@ -27,7 +27,6 @@ pub fn get_player_type<'gc>(
     let player_type = if cfg!(target_family = "wasm") {
         "PlugIn"
     } else {
-        //See: https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/system/Capabilities.html#playerType
         match activation.context.avm2.player_runtime {
             PlayerRuntime::FlashPlayer => "StandAlone",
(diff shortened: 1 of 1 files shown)


core/src/avm2/globals/flash/system/capabilities.rs
```diff
@@ -26,7 +27,11 @@ pub fn get_player_type<'gc>(
     let player_type = if cfg!(target_family = "wasm") {
         "PlugIn"
     } else {
-        "StandAlone"
+        //See: https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/system/Capabilities.html#playerType
```
evilpie, line 30: It seems like we usually don't really refer to the documentation for specific cases like this.

lvyitian: For the documentation yes, but there're some places that reference the avmplus source like this, if you think this is unnecessary, you can remove it.

Lord-McSweeney: I think it should be removed, the documentation is an implicit source for most of these cases.

Aaron1011 (approved):

torokati44:
Thank you!

PR #14916, opened 2024-01-23
avm2: Implement flash.media.ID3Info

lvyitian:


Dinnerbone:
Thank you! Any chance of a test? :D

lvyitian:
I've never tried of making a test, and I don't know how to do it, I'd appreciate if you could write one for me.

core/src/loader.rs
```diff
@@ -1492,7 +1493,123 @@ impl<'gc> Loader<'gc> {
...
+                            .avm2()
+                            .classes()
+                            .id3info
+                            .construct(&mut activation, &[])
+                            .expect("failed to construct ID3Info object");
+                        if let Avm2Object::ScriptObject(id3) = id3 {
+                            let tag = Tag::read_from(response.body.as_slice());
+                            if let Ok(tag) = tag {
+                                call_id3_event = true;
+                                let _ = id3.set_property_local(
```
evilpie, line 1509: Can you use `set_public_property` here?

core/src/loader.rs
```diff
@@ -1492,7 +1493,123 @@ impl<'gc> Loader<'gc> {
...
+                        Avm2::dispatch_event(&mut activation.context, complete_evt, sound_object);
+
+                        let mut call_id3_event = false;
+                        let id3 = activation
+                            .avm2()
+                            .classes()
+                            .id3info
+                            .construct(&mut activation, &[])
+                            .expect("failed to construct ID3Info object");
+                        if let Avm2Object::ScriptObject(id3) = id3 {
```
evilpie, line 1505: I don't think this `if` is necessary at all. We should be able to use the original `id3` just fine.

core/src/loader.rs
```diff
@@ -1492,7 +1493,123 @@ impl<'gc> Loader<'gc> {
...
+                            let tag = Tag::read_from(response.body.as_slice());
+                            if let Ok(tag) = tag {
+                                call_id3_event = true;
+                                let _ = id3.set_property_local(
+                                    &Multiname::new(
+                                        activation.avm2().find_public_namespace(),
+                                        "album",
+                                    ),
+                                    tag.album().map_or(Avm2Value::Null, |v| {
```
evilpie, line 1514: Personally I would write this as 
```rust
if let Some(album) = tag.album() {
    id3.set_public_property("name", AvmString::new_utf8(activation.gc(), album).into(),
                            &mut activation).expect("failed set_public_property");
}
```

evilpie:
Great work implementing this, thanks!

evilpie:
The documentation claims that ID3Info also dynamically supports a lot of other tags, which this doesn't cover. Not something that needs to be implemented before landing to be clear.

> The following table describes ID3 tags that are supported but do not have predefined properties in the Sound class.

https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/media/Sound.html#id3

core/src/loader.rs
lvyitian: If I use set_public_property function, the public_namespace_vm_internal namespace will be used, is the VM_INTERNAL api version fine?

core/src/loader.rs
lvyitian: Is the VM_INTERNAL api version fine here?

core/src/loader.rs
lvyitian: I remember that this is because I'm not sure whether non-public function can be called from another module or not, so I unwrap it so that I can use public function instead.

lvyitian:
> The documentation claims that ID3Info also dynamically supports a lot of other tags, which this doesn't cover. Not something that needs to be implemented before landing to be clear.
> 
> > The following table describes ID3 tags that are supported but do not have predefined properties in the Sound class.
> 
> https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/media/Sound.html#id3

Yes, I've read this, but there isn't a direct way to get those fields, so I haven't implemented them.

core/src/loader.rs
evilpie: Should be fine, because the property exists already and `public_namespace_vm_internal` matches any namespace.

evilpie (approved):

core/src/loader.rs
```diff
@@ -1584,7 +1585,94 @@ impl<'gc> Loader<'gc> {
...
+                                )
+                                .expect("failed set_public_property");
+                            }
+                        }
+
+                        sound_object
+                            .as_sound_object()
+                            .expect("Not a sound object")
+                            .set_id3(activation.context.gc_context, Some(id3));
+                        if call_id3_event {
```
evilpie, line 1671: This could be written as `if tag.is_ok()` after changing the code above to `if let Ok(ref tag) = tag`.

core/src/avm2/globals/flash/media/sound.rs
```diff
@@ -296,3 +296,16 @@ pub fn load_pcm_from_byte_array<'gc>(
...
 }
+
+/// Implements `Sound.id3`
+pub fn get_id3<'gc>(
+    _activation: &mut Activation<'_, 'gc>,
+    this: Object<'gc>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    if let Some(id3) = this.as_sound_object().unwrap().id3() {
+        Ok(Value::Object(id3))
```
Lord-McSweeney, line 307: Does `id3.into()` not work here?

core/src/loader.rs
```diff
@@ -1584,7 +1585,92 @@ impl<'gc> Loader<'gc> {
                             &mut activation.context,
                             "complete",
                         );
-                        Avm2::dispatch_event(uc, complete_evt, sound_object);
+                        Avm2::dispatch_event(&mut activation.context, complete_evt, sound_object);
```
Lord-McSweeney, line 1588: Have you tested that the `id3` event is always dispatched after the `complete` event, and that `Sound.id3` is still null when accessing it the `complete` event handler?

core/src/loader.rs
```diff
@@ -1584,7 +1585,92 @@ impl<'gc> Loader<'gc> {
...
+                            .classes()
+                            .id3info
+                            .construct(&mut activation, &[])
+                            .expect("failed to construct ID3Info object");
+                        let tag = Tag::read_from(body.as_slice());
+                        if let Ok(ref tag) = tag {
+                            if let Some(v) = tag.album() {
+                                id3.set_public_property(
+                                    "album",
+                                    Avm2Value::String(AvmString::new_utf8(activation.gc(), v)),
```
Lord-McSweeney, line 1601: Does `AvmString::new_utf8(activation.gc(), v).into()` suffice here (and for all the other fields, below)?

core/src/loader.rs
lvyitian: As for mp3, it will be fired before complete event is fired. Actually, id3 event will be fired as soon as the id3 metadata was read completely.

lvyitian: Tested with mp3:
progress event
id3 event [object ID3Info]
progress event
complete event

Lord-McSweeney (approved):
Can you add a test?

lvyitian:
> Can you add a test?

Added.

tests/tests/swfs/avm2/stage3d_fractal/output.difference-color-windows-Vulkan.png
Lord-McSweeney, line 1: These files were accidentally added, can you delete them?

core/src/avm2/globals/flash/media/sound.rs
```diff
@@ -279,7 +283,96 @@ pub fn load_compressed_data_from_byte_array<'gc>(
     let handle = activation.context.audio.register_mp3(bytes).map_err(|e| {
         Error::RustError(format!("Failed to register sound from bytearray: {e:?}").into())
     })?;
-
+    let progress_evt = activation
```
Lord-McSweeney, line 286: Is this event really dispatched for `loadCompressedDataFromByteArray`? If so, can you check that the `complete` event is also dispatched? Also, might it be worth it extracting the ID3 logic into a function in `SoundObject`?

tests/tests/swfs/avm2/stage3d_fractal/output.difference-color-windows-Vulkan.png
lvyitian: OK

core/src/avm2/globals/flash/media/sound.rs
lvyitian: I've checked that, only progress and id3 event are called, complete event isn't called.

lvyitian: Okay, I've added a read_and_call_id3_event function and used it instead.

tests/tests/swfs/avm2/id3_info/test.as
```diff
@@ -0,0 +1,30 @@
...
...0x04, 0x45, 0x28, 0x01, 0x58, 0x8C, 0x08, 0xA0, 0x88, 0xA8, 0x00, 0x0F, 0x03, 0x00, 0x78, 0x11, 0x08, 0x00, 0x05, 0x60, 0x60, 0x26, 0x82, 0x22, 0xA0, 0x11, 0xBC, 0x0D, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xFF, 0xFB, 0x10, 0x64, 0xE3, 0x0F, 0xF0, 0x12, 0x00, 0xC3, 0x80, 0x00, 0x00, 0x08, 0x08, 0x00, 0x98, 0x85, 0x00, 0x62, 0x05, 0x40, 0x00, 0x01, 0xA4, 0x00, 0x00, 0x00, 0x20, 0x00, 0x00, 0x34, 0x80, 0x00, 0x00, 0x04, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA];
```
Dinnerbone, line 15: Is this "test_audio.mp3"? Why embed it here and not load it from a file, or embed it in the swf normally?

lvyitian: Yes, it is test_audio.mp3, because I use FlashDevelop and I have no idea how to embed it.

lvyitian: I can load it from my own github repository by https URL, but I think embed it into source is better than this. And thanks to this, I found loadCompressedDataFromByteArray doesn't call events correctly.

lvyitian:
@Dinnerbone So, what should I do?

Lord-McSweeney:
> there's no way of accessing filesystem directly

Ruffle's test backend supports doing a `new URLRequest("test_audio.mp3")`.

lvyitian:
@Dinnerbone Solved.

core/src/avm2/object/sound_object.rs
```diff
@@ -129,6 +137,89 @@ impl<'gc> SoundObject<'gc> {
...
+            if let Some(v) = tag.year() {
+                id3.set_public_property(
+                    "year",
+                    AvmString::new_utf8(activation.gc(), v.to_string()).into(),
+                    activation,
+                )
+                .expect("failed set_public_property");
+            }
+        }
+        self.set_id3(activation.context.gc_context, Some(id3));
```
Lord-McSweeney, line 217: With this implementation, `id3` will always be non-null in FP- have you confirmed that this is true? Especially for mp3s without ID3 info

tests/tests/swfs/avm2/id3_info/test.as
```diff
@@ -0,0 +1,24 @@
...
+		
+		public function Main()
+		{
+			var sound:Sound = new Sound();
+			sound.addEventListener(Event.ID3, function(event:Event):void{
+				trace("id3event id3:" + sound.id3);
+				var properties:Array = ["album", "artist", "comment", "genre", "songName", "track", "year"];
+				for (var i:String in properties) trace(properties[i]+":"+sound.id3[properties[i]]);
+			});
+			sound.load(new URLRequest("file:///test_audio.mp3"));
```
Lord-McSweeney, line 19: I think this can just be `new URLRequest("test_audio.mp3")`, no?

core/src/avm2/object/sound_object.rs
lvyitian: Yep, it is always non-null, for audio that doesn't contain id3 info, the fields inside it will be null.

tests/tests/swfs/avm2/id3_info/test.as
lvyitian: I just follow this: https://github.com/ruffle-rs/ruffle/blob/master/tests/framework/src/backends/navigator.rs#L243

Lord-McSweeney: >`Url::parse("file:///")?;`

This means that there is an assumed base URL of `file:///`, not that it needs to start with `file:///`.

Lord-McSweeney: Does `new URLRequest("test_audio.mp3")` not work?

lvyitian: It works, there was a misoperation that makes them unchanged.

Lord-McSweeney (approved):

Dinnerbone (approved):
Thank you!

Benman2785:
before that update i had at least audio when streaming rtmp flv to jwplayer - now i have nothing :/

Issue #9229, opened 2023-01-20
Crash on toxic_2.swf


Lord-McSweeney:
### Describe what you were doing

Trying to move on the first level.

### What does the crash message say?

panicked at 'Error in Queue::submit: parent device is lost', /home/uqers/.cargo/git/checkouts/wgpu-e4487c3622394142/a154700/wgpu/src/backend/direct.rs:306:9

### Ruffle Version

0.1.0-nightly ( )

### Affected platform

Desktop app

### Operating system

Ubuntu 20.04 (focal) [64-bit]

### Additional information

Filename: toxic_2.swf

### Render Info
Renderer: wgpu
Adapter Backend: Vulkan
Adapter Name: "Intel(R) UHD Graphics (ICL GT1)"
Adapter Device Type: IntegratedGpu
Adapter Driver Name: "Intel open-source Mesa driver"
Adapter Driver Info: "Mesa 21.2.6"
Enabled features: PUSH_CONSTANTS
[...]

PR #14919, opened 2024-01-23
wgpu: Some optimisations, some refactors, much less buffers

Dinnerbone:
This reduces the amount of buffers needed to run Heartcore from about 8k to 3k peak, which should improve performance and reduce memory usage. We also don't need a second command buffer per draw anymore. Filters, especially, should be cheaper to do now.

It also simplifies the codebase a lot since we no longer have to juggle lifetimes with our old staging belt system, and we also don't specialcase push constants since the performance should be comparable now. This opens up future improvements, especially to BMD draws (and making BMD operations on gpu viable, to avoid readbacks)


SN902:
[pony-creator-v3-generalzoi.zip](https://github.com/ruffle-rs/ruffle/files/14031204/pony-creator-v3-generalzoi.zip)

`thread 'main' panicked at render\wgpu\src\mesh.rs:413:10:
Mesh uniform buffer was too large!: BufferFull`

Dinnerbone:
@SN902 web or desktop? Do they look normal to you before this PR, no rendering artifacts or crashes??

SN902:
> @SN902 web or desktop? Do they look normal to you before this PR, no rendering artifacts or crashes?

I tested your PR on the desktop build, these games works fine on the latest nightly 2024-01-24

Dinnerbone:
Try now. I think I misunderstood the max_uniform_buffer_binding_size limit.

SN902:
> Try now. I think I misunderstood the max_uniform_buffer_binding_size limit.

Fixed, now works perfectly.

Lord-McSweeney:
This doesn't fix #9229 for me

Aaron1011 (approved):

PR #13615, opened 2023-10-19
core: Use correct embedded font lookup order

Dinnerbone:
Embedded fonts should properly fall back to the correct font for the style they have - and check for fonts case insensitive.

Also fixes that `<font>` in text should change the style to Regular.

This fixes #13602
This fixes Plazma Burst credits text

PR #1201, opened 2020-09-19
text: Default color for text spans should be black

Herschel:
Previously color defaulted to 0% alpha, causing text to be
invisible if the text field did not specify a defualt color.

PR #14924, opened 2024-01-24
core: Ignore font style in default TextFormat for HTML fields

kjarosh:
It seems that font styles in the default text format are ignored when dealing with an HTML field.

This patch revisits the fix from feacbdc1 (#13615), which assumed that `<font>` resets font style. That does not seem to be the case, but rather the bug was caused by the invalid default text format, which forced the text to be bold, due to the bold variant of the font being linked to the text field.

This patch reverts 2f84d468 (#1201), which assumed that the default color for a text span has 100% alpha. The test added here contradicts it and it seems that the default color is in fact rgba(0,0,0,0). Testing the original SWF suggests that the underlying problem has been fixed since that time.


Dinnerbone (approved):
Thank you!

PR #14931, opened 2024-01-24
avm2: Fix handling of Loader.loadBytes for images

Aaron1011:
I've switched back to the original code for creating the bitmap/bitmapdata, rather than relying on custom initialization logic that we only used in loader.

To make sure that the Bitmap/BitmapData are only exposed to ActionScript at the correct time, I've added a new flag to control when 'LoaderInfo.content' becomes non-null


core/src/avm2/globals/flash/display/loader_info.rs
```diff
@@ -146,13 +146,15 @@ pub fn get_content<'gc>(
...
-        .as_loader_info_object()
-        .and_then(|o| o.as_loader_stream())
-    {
+    let loader_info = this.as_loader_info_object().unwrap();
+
+    if let Some(loader_stream) = loader_info.as_loader_stream() {
         match &*loader_stream {
             LoaderStream::Swf(_, root) | LoaderStream::NotYetLoaded(_, Some(root), _) => {
                 if root.movie().is_action_script_3() || !root.movie().is_movie() {
+                    if !loader_info.expose_content() {
```
Lord-McSweeney, line 155: This if-statement should probably go before the `if root.movie().is_action_script_3() || !root.movie().is_movie()`

Dinnerbone (approved):

PR #14953, opened 2024-01-26
core: Improve HTML handling for EditText & add support for SWF 6,7

kjarosh:
This PR improves the logic of parsing and formatting HTML for EditText, and adds support for SWF versions 6 and 7.

Examples of bugs fixed:
* invalid HTML: When generating HTML, Ruffle sometimes generated mismatched tags, e.g. tried to close a tag which was never opened.
* text outside of tags: sometimes, especially in case of multiline fields with multiple paragraphs, Ruffle generated proper tags, but the text was placed outside of them.
* BR/SBR closing issues: When opened, Ruffle waited for a close tag and ignored every other closing tag. BR/SBR do not need to be closed, so Ruffle often waited indefinitely.
* P/LI behavior: P and LI have a very peculiar behavior, but a very important one, because it influences the number of paragraphs/bullets and thus newlines for multiline fields.

Support for SWF 6,7:
* whitespace in SWF 6,7: SWF versions 6,7 ignore witespace-only text. This significantly influences the behavior of newlines and paragraphs.
* kerning in SWF 6,7: Enabling kerning in `<font>` works only for SWF 8+.
* multiline in SWF 6: FP 6 defines the `multiline` property, but it completely ignores it, and the field behaves as if it's always multiline.

I've added comprehensive tests for SWF versions 6, 7, and 8 as they all differ in behavior. Some of the cases may seem a bit random at a first glance, but I was basically checking randomly generated HTML strings and I selected those for which tests were failing during development (and were semantically relevant!). Of course some sane standard test cases are included.

I also did a small refactoring related to `TextSpan` structures and made them easier to work with.

At first I just wanted to fix a small bug related to handling newlines in HTML, but things kept popping up! So I took a more universal approach.


core/src/html/text_format.rs
```diff
@@ -710,15 +736,13 @@ impl FormatSpans {
...
+                                if kerning == WStr::from_units(b"1") && swf_version >= 8 {
+                                    // Enabling kerning works only for SWF >=8
                                     format.kerning = Some(true);
                                 } else if kerning == WStr::from_units(b"0") {
                                     format.kerning = Some(false);
                                 }
                             }
-
-                            format.bold = Some(false);
-                            format.italic = Some(false);
```
kjarosh, line 721: Tests will probably fail due to this, #14924 ensures there's no regression.

kjarosh:
BTW I probably lost my sanity while implementing this

Dinnerbone:
> BTW I probably lost my sanity while implementing this

You're a true Ruffle developer now. Welcome to the club ❤️

Dinnerbone (approved):
Thank you very much!

PR #14963, opened 2024-01-26
debug_ui: Do not render container as collapsible if it has no children

sleepycatcoding:
Before:
before

After:
after


torokati44 (approved):
Thanks!

torokati44:
Agh, it just came to my mind, I guess the Domain list (https://github.com/ruffle-rs/ruffle/pull/14672) could also use this treatment (although I guess it's rare for a Domain to be empty), plus, more importantly, the indentation fix similar to https://github.com/ruffle-rs/ruffle/pull/12467. But those can be in a different PR, if any.

PR #14964, opened 2024-01-26
avm2: Only use embedded CFF fonts for TLF

evilpie:


evilpie:
This fixed a regression caused by #14749 in the SWF from #10173 . Because we were using the DefineFont3 version (with missing glyphs) of "Arial" for TLF text.

Dinnerbone (approved):

PR #14965, opened 2024-01-26
desktop: Fix context menu

Dinnerbone:
Previous approach was to immediately spawn a blank panel with a context menu attached, so egui would then see the right click event too and do its thing.

This was too unreliable, so now we just have our own context menu.

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -221,7 +221,7 @@ impl App {
                                 ElementState::Pressed => PlayerEvent::MouseDown { x, y, button },
                                 ElementState::Released => PlayerEvent::MouseUp { x, y, button },
                             };
-                            if state == ElementState::Pressed && button == RuffleMouseButton::Right
+                            if state == ElementState::Released && button == RuffleMouseButton::Right
                             {
                                 // Show context menu.
                                 // TODO: Should be squelched if player consumes the right click event.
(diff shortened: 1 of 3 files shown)


torokati44:
Just an idea, definitely not a blocker:
Do you think making "press-drag-release" activation would be feasible?
I know it's not that commonly used, but sometimes it's more convenient.
And it works in FP, at least on my Linux.

Dinnerbone:
> "press-drag-release" activation

Sorry, can you explain what you mean by this?

torokati44:
Certainly!
In most UI toolkits, it's possible to invoke the context menu, possibly open submenus, and select a menu option, with a single "right-click drag" gesture.
So, pressing down the right mouse button immediately shows the context menu, then hovering over items while still keeping it pressed highlights/opens them, then finally releasing it on top of an item triggers it.

Dinnerbone:
Ah right okay.
Flash does seem to do the menu on *press* for me, so that would be possible. But I tried a menu in literally every other application I had running at the time, and they did it on *release* - which also works around a conflicting behaviour with us vs egui if I do on mouse down.

So, not today :( Sorry

torokati44:
It might be a Windows thing/limitation then. Because it works for every single of _my_ applications! :P
But alright, no big deal. It works, which is the important thing!

torokati44 (approved):
It's kinda unfortunate that we have to replicate the context menu functionality from scratch; but really, it's not that bad at all!

PR #14967, opened 2024-01-26
desktop: Allow specifying WGPU instance flags from environment

arthomnix:
This is mainly useful to allow forcing usage of Vulkan on non-compliant drivers, such as Mesa's driver for Intel Haswell IGPUs, using the `WGPU_ALLOW_UNDERLYING_NONCOMPLIANT_ADAPTER` environment variable. Currently, wgpu will always reject adapters with non-compliant drivers, even though they will often still work fine.

### DIFF
--- a/desktop/src/gui/controller.rs
+++ b/desktop/src/gui/controller.rs
@@ -55,6 +55,7 @@ impl GuiController {
         }
         let instance = wgpu::Instance::new(wgpu::InstanceDescriptor {
             backends: backend,
+            flags: wgpu::InstanceFlags::default().with_env(),
             ..Default::default()
         });
         let surface = unsafe {


Dinnerbone (approved):

torokati44 (approved):
Thank you, and welcome to the project! :wave:

torokati44:
Actually, there are a couple more calls to `wgpu::Instance::new` here and there...
Don't we want this for all of them?

arthomnix:
Some of them are in tests or for wasm, but we possibly want this for the rest of them

torokati44:
Well, yeah, basically, the others are:
 - for the timedemo mode
 - for visual tests
 - for the exporter
 - for the Android app
 - for web

I don't think it's crucial for any of those... _Maybe_ for the timedemo, in some weird edge cases...? But eh. If somebody ends up needing it, they can add it. :upside_down_face:

PR #14982, opened 2024-01-28
avm2: Make Math.random() more closely match Flash Player

Aaron1011:
See the comments for details. Our previous implementation was 'too good', and broke Bloons Tower Defense 5 by generating `Number`s that Flash Player would never generate.


tests/tests/swfs/avm2/cryptscore/Test.as
```diff
@@ -0,0 +1,28 @@
...
+	public class Test extends MovieClip {
+	
+		
+		public function Test() {
+			var score = new CryptScore();
+			for (var i = 0; i < 10000; i ++) {
+				if (i % 1000 == 0) {
+					trace("i = " + i);
+				}
+				score.value = score.value;
```
torokati44, line 17: What does this line do? :thinking:

Aaron1011: This runs both the getter and setter.

torokati44: Ooh, right!

torokati44: Well at any rate, it wasn't immediately obvious...

tests/tests/swfs/avm2/cryptscore/test.toml
```diff
@@ -0,0 +1,7 @@
+# This file tests that our Math.random() implementation works
```
adrian17, line 1: The only thing that makes me hesitant from LGTM'ing this test is the possibility you mentioned on Discord, that it might randomly fail CI. Though I guess we can disable the test if it happens for the first time...?

Aaron1011: I believe we seed our rng when running tests, so this shouldn't be an issue

adrian17 (approved):

PR #14983, opened 2024-01-28
chore: Fix nightly Clippy lints

Aaron1011:


adrian17 (approved):

tests/mocket/src/main.rs
```diff
@@ -21,6 +21,7 @@ struct Opt {
     file_path: PathBuf,
 }
 
+#[allow(clippy::unused_io_amount)]
```
torokati44, line 24: Maybe this would be better on the smallest possible scope?
Perhaps right above ` SocketEvent::WaitForDisconnect => {` in the `match`?

PR #14985, opened 2024-01-28
Report OS as Windows instead of Linux for now

Fancy2209:
`flash.system.Capabilities.manufacturer` now reports Adobe Windows and `flash.system.Capabilities.os` Windows 8


evilpie:
You missed the Rust implemented `get_version`.

core/src/avm2/globals/flash/system/capabilities.rs
```diff
@@ -12,7 +12,7 @@ pub fn get_version<'gc>(
     // TODO: Report the correct OS instead of always reporting Linux
```
evilpie, line 12:
```suggestion
    // TODO: Report the correct OS instead of always reporting Windows
```

evilpie (approved):

PR #14991, opened 2024-01-29
debug_ui: Show ClassObject when clicking on domain class

Aaron1011:
This re-uses our existing infrastructure for displaying AVM2 Class objector. One minor limitation of this approach is the inability to view a `Class` that hasn't yet had its `ClassObject` created - however, this should be rare in practice.

### DIFF
--- a/core/src/avm2/class.rs
+++ b/core/src/avm2/class.rs
@@ -158,6 +158,12 @@ pub struct Class<'gc> {
     /// System defined classes are allowed to have illegal trait configurations
     /// without throwing a VerifyError.
     is_system: bool,
+
+    /// The ClassObjects for this class.
+    /// In almost all cases, this will either be empty or have a single object.
+    /// However, a swf can run `newclass` multiple times on the same class
+    /// to create multiple `ClassObjects`.
+    class_objects: Vec<ClassObject<'gc>>,
 }
 
 /// Allows using a `GcCell<'gc, Class<'gc>>` as a HashMap key,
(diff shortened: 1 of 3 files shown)


core/src/avm2/class.rs
```diff
@@ -280,6 +284,20 @@ impl<'gc> Class<'gc> {
         self.attributes = attributes;
     }
 
+    pub fn set_class_object(&mut self, class_object: ClassObject<'gc>) {
+        if self.class_object.is_some() {
```
Lord-McSweeney, line 288: Can't you call `newclass` twice on the same `Class` and get two `ClassObject`s created pointing to the same class?

Aaron1011: Yes, but that would be an extremely cursed swf, and I'd prefer to get a panic report if that ever happens.

Lord-McSweeney: I think the check should be removed, it's not that cursed of a case, and I feel like some SWF could end up tripping it. Besides, this should be unnecesary once the `ClassObject` refactor is completed.

Aaron1011: If this ever does happen, it's going to make the debug ui experience very confusing. If we don't want a panic, then I'd prefer to store a Vec, and show multiple ClassObjects in the ui somehow

Aaron1011: I've modified this to store a `Vec` and display a list in the debug ui

Lord-McSweeney (approved):

PR #14994, opened 2024-01-29
fix(deps): update rust dependencies

torokati44:
This is https://github.com/ruffle-rs/ruffle/pull/14990, but without the check that gets hung up on the fact that `gif` is now duplicated. `0.12.0` is still needed by `image`, while directly we're using `0.13.1` now. I don't think this is a bug deal, but we can skip the direct usage update.
 Plus an additional `cargo update`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -149,9 +149,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.4"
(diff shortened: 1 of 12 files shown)


torokati44:
The linked PR now also includes a `regress` `0.7` -> `0.8` bump, I will have to follow up with that here.

PR #15004, opened 2024-01-29
Fix GH saying Ruffle is written in AS

Fancy2209:


.gitattributes
```diff
@@ -9,3 +9,6 @@
 *.fla  binary
 *.swf  binary
 *.swc  binary
+
+# Fix GH saying Ruffle is written in ActionScript 
+tests\tests\swfs\** linguist-vendored
```
torokati44, line 14: I think forward slashes would be preferred.

torokati44 (approved):
Thanks!
Although a commit message with a tag (following all other commit messages, and adhering to `CONTRIBUTING.md`) would be nicer.
That can be adjusted by squash-merging too.

And maybe a newline character at the end of the file, but that is really just nitpicking.

torokati44:
The result is actually this, possible due to the `\` -> `/` change:
image

Croworbit:
can I ask why it showed such a high percent for actionscript before?

does this remove all of that?

torokati44:
Because it also counted our numerous tests written in AS, which are not really "part of" Ruffle itself.
So this PR makes GH ignore just the folder where we keep those.

Croworbit:
numerous tests indeed - thanks for the explanation!

Issue #14999, opened 2024-01-29
Missed keyboard controls for active TextFields (PageUp/Down, Ctrl + Left/Right/PageUp/PageDown, and Shift + Ctrl + Left/Right/PageUp/PageDown)


ReFaller:
### Describe the bug

In active text field:
PageUp/Down - should set the current text marker to begin/end
Ctrl + Left/Right should set the current text marker to the last space delimiter.
Shift + Ctrl + Left/Right doesn't work to select a text to the last space delimiter.
Shift + Ctrl + PageUp/PageDown doesn't work to select a text to the begin/end.

### Expected behavior

Inside the Flash player, basic keyboard control for text fields is provided, following the standards of all text editors.

### Content Location

[test.zip](https://github.com/ruffle-rs/ruffle/files/14087082/test.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

kjarosh:
I'm currently working on implementing this

kjarosh:
https://github.com/ruffle-rs/ruffle/pull/15116 would be helpful for PageUp/Down, as it effectively fixes the `screen_position_to_index` mapping, and PageUp/Down basically moves the cursor up/down by the height of the text field (according to my testing).

PR #15027, opened 2024-01-30
text: Advanced horizontal text controls for EditText

kjarosh:
This PR implements the following controls:
* <kbd>Ctrl</kbd> + <kbd>Delete</kbd> &mdash; delete a word to the right,
* <kbd>Ctrl</kbd> + <kbd>Backspace</kbd> &mdash; delete a word to the left,
* <kbd>Ctrl</kbd> + <kbd>→</kbd> &mdash; move a word to the right,
* <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>→</kbd> &mdash; select a word to the right,
* <kbd>Ctrl</kbd> + <kbd>←</kbd> &mdash; move a word to the left,
* <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>←</kbd> &mdash; select a word to the left,
* <kbd>Home</kbd> &mdash; move left to the beginning of the line,
* <kbd>Shift</kbd> + <kbd>Home</kbd> &mdash; select left till the beginning of the line,
* <kbd>Ctrl</kbd> + <kbd>Home</kbd> &mdash; move left to the beginning of the document,
* <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>Home</kbd> &mdash; select left till the beginning of the document,
* <kbd>End</kbd> &mdash; move right to the end of the line,
* <kbd>Shift</kbd> + <kbd>End</kbd> &mdash; select right till the end of the line,
* <kbd>Ctrl</kbd> + <kbd>End</kbd> &mdash; move right to the end of the document,
* <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>End</kbd> &mdash; select right till the end of the document.

Related to #14999. Note that this PR implements horizontal controls only. Vertical controls (arrows up/down page up/down) require more work, as we need to measure widths of characters and be aware of the text layout.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -64,6 +64,7 @@ image = { version = "0.24.8", default-features = false, features = ["tiff", "dxt
 enum-map = "2.7.3"
 ttf-parser = "0.20"
 num-bigint = "0.4"
+unic-segment = "0.9.0"
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies.futures]
 version = "0.3.30"
--- a/wstr/src/ops.rs
+++ b/wstr/src/ops.rs
@@ -462,6 +462,47 @@ impl<'a> WStrToUtf8<'a> {
         }
     }
 
+    /// Map the given UTF-16 code unit index to its corresponding UTF-8 code unit index.
+    pub fn utf8_index(&self, utf16_index: usize) -> Option<usize> {
+        self.translate_index(utf16_index, false)
+            .map(|(utf8_index, _)| utf8_index)
+    }
+
+    /// Map the given UTF-8 code unit index to its corresponding UTF-16 code unit index.
+    pub fn utf16_index(&self, utf8_index: usize) -> Option<usize> {
+        self.translate_index(utf8_index, true)
+            .map(|(_, utf16_index)| utf16_index)
+    }
+
+    fn translate_index(&self, index: usize, is_utf8: bool) -> Option<(usize, usize)> {
+        let ascii_prefix_len = self.head.len();
+        if index <= ascii_prefix_len {
+            return Some((index, index));
+        }
+
+        if self.tail.is_empty() {
+            return None;
+        }
+
+        let mut utf8_tail_pos = 0;
+        let mut utf16_tail_pos = 0;
+
+        while if is_utf8 {
+            utf8_tail_pos + ascii_prefix_len < index
+        } else {
+            utf16_tail_pos + ascii_prefix_len < index
+        } {
+            let c = self.tail[utf16_tail_pos..].chars().next()?.ok()?;
+            utf8_tail_pos += c.len_utf8();
+            utf16_tail_pos += c.len_utf16();
+        }
+
+        Some((
+            ascii_prefix_len + utf8_tail_pos,
+            ascii_prefix_len + utf16_tail_pos,
+        ))
+    }
+
     #[inline]
     pub fn prefix(&self) -> &str {
         self.head
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4270,6 +4270,7 @@ dependencies = [
  "thiserror",
  "tracing",
  "ttf-parser",
+ "unic-segment",
  "url",
  "wasm-bindgen-futures",
  "weak-table",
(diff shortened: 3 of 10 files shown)


wstr/src/ops.rs
```diff
@@ -462,6 +462,47 @@ impl<'a> WStrToUtf8<'a> {
...
+            utf8_tail_pos += c.len_utf8();
+            utf16_tail_pos += c.len_utf16();
+        }
+
+        Some((
+            ascii_prefix_len + utf8_tail_pos,
+            ascii_prefix_len + utf16_tail_pos,
+        ))
+    }
+
```
moulins, line 505: Note that similar code already exists in https://github.com/ruffle-rs/ruffle/blob/master/core/src/avm2/regexp.rs#L380.

Unfortunately, sharing code between this and `avm2/regexp.rs` isn't trivial as the ownership is different (a simple reference here, and an owned `AvmString<'gc>` in `regexp.rs`), but maybe there's a clever way of doing it?

kjarosh: That's right, sharing the code is not trivial, I thought about this.

Basically (1) I could just use `WStrToUtf8::new(&self.text).utf{8,16}_index(utf{16,8}_index)` when the indices are looked up, but that's not a perfect solution because it would force `O(n^2)`/`O(n^2)`/`O(n)` forward-pass/backward-pass/lookup complexity instead of the current `O(n)`/`O(n^2)`/`O(n)` and potential `O(n)`/`O(n)`/`O(1)+`. That's however the simplest solution which is not affected by the ownership problem.

If I decide to have state (which is required for anything better than `O(n)` lookup) I have the ownership problem. If I understand gc-arena well enough, `Collect` implementation is required to keep track of `Gc` references. The `derive(Collect)` automatically implements it hierarchically, but every non `'static` field also has to implement `Collect`. In this case I cannot implement `Collect` for `WStrToUtf8`, because (2) from the point of view of wstr it makes no sense, and (3) in regexp I have the E0117 problem (assuming that implementing `Collect` for `WStrToUtf8` is even possible at this point). I could make it `'static` by (4) not referencing head and tail but passing the string for every lookup, but that would be difficult to use, unsafe, and unnatural. I could (5) make it owned so that there is no pointer. That would require creating something like `WStrToUtf8Owned` and converting between these two types. There is also an issue of `WStrToUtf8` being a struct of pointers so implementing `ToOwned`/`Borrow` would be difficult? `WStrToUtf8Owned` would have to be implemented in wstr, as otherwise I would have to make the head/tail implementation detail public. I cannot replace `WStrToUtf8` with `WStrToUtf8Owned` as it would create owned instances every time a conversion to UTF-8 is done. I could also (6) implement `Collect` for `CachedText` manually, but that would require an unsafe implementation and I think more knowledge about gc-arena than I have.

It seems that the option (5) is the best, what do you think? Please point any mistakes or shortcomings in my analysis above — I'm a Rust beginner :)

Lord-McSweeney: > In this case I cannot implement Collect for WStrToUtf8
I'm not exactly sure of the context, but if `WStrToUtf8` is a field on a garbage-collected `Collect`-implementing object, you can mark the field as `#[collect(requires_static)]`, so you don't need to implement `Collect` for `WStrToUtf8` itself, it'll just be static, even if it references other pointers.

kjarosh: According to my understanding and testing for a field of type `WStrToUtf8<'a>` (it requires a lifetime) to be marked as `#[collect(require_static)]` the `'a` lifetime would have to outlive `'static`, which is not the case here. That's why I suggested making an owned version which does not require a lifetime. But I may be approaching it from a wrong direction

Dinnerbone (approved):
Looks great, thank you!

core/Cargo.toml
```diff
@@ -64,6 +64,7 @@ image = { version = "0.24.8", default-features = false, features = ["tiff", "dxt
 enum-map = "2.7.3"
 ttf-parser = "0.20"
 num-bigint = "0.4"
+unic-segment = "0.9.0"
```
torokati44, line 67: oh nyo: https://rustsec.org/advisories/RUSTSEC-2025-0098

https://github.com/ruffle-rs/ruffle/actions/runs/18682103319/job/53265402809?pr=21985#step:5:7043

torokati44: Maybe https://crates.io/crates/unicode-segmentation could also work instead...? :thinking:

kjarosh: Is there a reason other than this lib not supporting current unicode standard? Note that this behavior does not match Flash, it's a Ruffle feature atm (not to mention Flash wouldn't work with new unicode rules anyway).

danielhjacobs: According to evilpie in the issue where it was reported as unmaintained:

> The last commit has been 5 years ago. Outdated Unicode information causes compatibility issues. (See for example https://github.com/denoland/rust-urlpattern/issues/54)

danielhjacobs: https://github.com/rustsec/advisory-db/issues/2414

torokati44: 🤷 https://github.com/gfx-rs/wgpu/pull/8394

andyleiserson: I am waiving them in the wgpu repo because we use them via `deno` and `urlpattern`. It appears that `urlpattern` 0.4 has a replacement, but we can't switch to that until deno does and we update deno.

torokati44: https://github.com/ruffle-rs/ruffle/pull/21990

PR #15032, opened 2024-01-30
wgpu: Don't submit for every single BitmapData.draw()

Dinnerbone:
This should reduce the resource cost of BitmapData.draw() a fair bit, most impactful on webgl

### DIFF
--- a/core/src/bitmap/bitmap_data.rs
+++ b/core/src/bitmap/bitmap_data.rs
@@ -311,10 +311,10 @@ mod wrapper {
 
         /// Clones the underlying data, producing a new `BitmapData`
         /// that has no GPU texture or associated display objects
-        pub fn clone_data(&self) -> BitmapData<'gc> {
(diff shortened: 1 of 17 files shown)


core/src/bitmap/bitmap_data.rs
```diff
@@ -332,18 +332,21 @@ mod wrapper {
...
-                            copy_pixels_to_bitmapdata(&mut write, buffer, buffer_width, bounds)
-                        }))
+                    renderer
+                        .resolve_sync_handle(
+                            sync_handle,
+                            Box::new(|buffer, buffer_width| {
+                                copy_pixels_to_bitmapdata(&mut write, buffer, buffer_width, bounds)
+                            }),
+                        )
                         .expect("Failed to sync BitmapData");
```
Lord-McSweeney, line 350: There's no way this panic can be hit on the webgl/canvas backends, right?

Dinnerbone: Correct

Aaron1011 (approved):

PR #15044, opened 2024-01-31
Matching AVM1 to AVM2 for the sake of consistency

sombraguerrero:
Since we were a bit eager to get BTD5 solved, we merged the Math.random() AVMPLUS implementation to AVM2 ahead of considering including it in AVM1. This simply copies that code (and assumes we can apply the same test).


Dinnerbone:
AVM1 also has `action_random_number` in `activation.rs`, it needs to be changed too

sombraguerrero:
The Ubuntu runner does not seem to like me. :P

Dinnerbone (approved):

Dinnerbone:
> AVM1 also has `action_random_number` in `activation.rs`, it needs to be changed too

Turns out this is for integers only!

PR #15050, opened 2024-02-01
avm2: Add fast path for `Value::abstract_lt`-ing two integers

Lord-McSweeney:
Minor improvement to #10560. I couldn't find any content that used number comparison in a tight loop like Evoland did for integer comparison, so I didn't add the special-case.

### DIFF
--- a/core/src/avm2/value.rs
+++ b/core/src/avm2/value.rs
@@ -1315,21 +1315,26 @@ impl<'gc> Value<'gc> {
         other: &Value<'gc>,
         activation: &mut Activation<'_, 'gc>,
     ) -> Result<Option<bool>, Error<'gc>> {
-        let prim_self = self.coerce_to_primitive(Some(Hint::Number), activation)?;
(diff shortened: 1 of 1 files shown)


core/src/avm2/value.rs
```diff
@@ -1315,21 +1315,26 @@ impl<'gc> Value<'gc> {
         other: &Value<'gc>,
         activation: &mut Activation<'_, 'gc>,
     ) -> Result<Option<bool>, Error<'gc>> {
-        let prim_self = self.coerce_to_primitive(Some(Hint::Number), activation)?;
```
evilpie, line 1318: I like 
```rust
        if let (Value::Integer(a), Value::Integer(b)) = (self, other) {
            return Ok(Some(a < b));
        }
```

evilpie (approved):

core/src/avm2/value.rs
Lord-McSweeney: I think the current match syntax is more readable here, and it allows for easily adding more cases if/when we want them

PR #15055, opened 2024-02-01
avm2: Implement TextField::getLineText

### LINKED ISSUE #10975: Forty Thieves Solitaire - Click on the buttons does nothing
### Describe the bug

https://www.spiele-umsonst.de/azad/downloads/flashgames/fortythievessolitaire.swf - If you click play game in the menu or something else nothing happens, maybe has to do with the errors it throws?

core/src/avm2/events.rs:419 Error dispatching event EventObject(EventObject { type: "mouseUp", class: flash.events::MouseEvent, ptr: 0x103a8d4 }) to handler FunctionObject(FunctionObject { ptr: 0x3a63904 }) : ReferenceError: Error #1069: Property getLineText not found on flash.text.TextField and there is no default value.

### Expected behavior

Do something onclick...

### Affected platform

Online demo

### Operating system

All

### Browser

All

### Additional information

_No response_

evilpie:
This mostly fixes #10975. However I think we don't have the same layout / number of lines as Flash in some case, so we still have bugs.

image
This button in "Forty Thieves Solitaire" should be two lines, but is one in Ruffle.


tests/tests/swfs/avm2/textfield_getlinetext/test.toml
```diff
@@ -0,0 +1 @@
+num_frames = 5
```
Lord-McSweeney, line 1: Does this need 5 frames to run?

Aaron1011: Also, this should be `num_ticks`

core/src/avm2/globals/flash/text/text_field.rs
```diff
@@ -1151,6 +1151,31 @@ pub fn get_line_metrics<'gc>(
...
+pub fn get_line_text<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Object<'gc>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    if let Some(this) = this
+        .as_display_object()
+        .and_then(|this| this.as_edit_text())
+    {
+        let line_num = args
```
Lord-McSweeney, line 1163: This can use `ParametersExt` - `let line_num = args.get_i32(activation, 0)?;`

tests/tests/swfs/avm2/textfield_getlinetext/output.txt
```diff
@@ -0,0 +1,8 @@
+///
+hello world
+RangeError: Error #2006: The supplied index is out of bounds.
```
Lord-McSweeney, line 3: This is not the Flash-matching error message, Flash provides some additional information about the index- can you just print the error class name and code?

Lord-McSweeney (approved):

tests/tests/swfs/avm2/textfield_getlinetext/output.txt
evilpie: Interesting, I can do that. What error do you get? This is from Flash 32,0,0,314.

Lord-McSweeney: Maybe I'm mistaken- what I remember is that we omitted the index info from one of the index errors (or maybe it was a different error?)

evilpie: I think you might be thinking of the exact opposite, we produce a more exact error message, while Flash has just the generic out of bounds message: https://github.com/ruffle-rs/ruffle/blob/0df6bafdaf9cbbfcf0b1bb395f77d684c1d96cef/core/src/avm2/globals/flash/display/display_object_container.rs#L334.

PR #15066, opened 2024-02-02
github: Also exclude `.git` and `tests/tests/swfs` from `reproducible-source.zip`

torokati44:
This should reduce the unzipped(?) size from ~216 MB down to ~15MB, hopefully allowing the submission of it to AMO to not fail.

Also leave the `*` wildcard expansion up to `zip` instead of doing it in the invoking shell.

Plus a (now) much needed line wrap for bonus.

### DIFF
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -322,7 +322,9 @@ jobs:
       - name: Produce reproducible source archive
         if: ${{ !matrix.demo }}
         shell: bash -l {0}
-        run: zip -r reproducible-source.zip . -x /web/node_modules/**/* -x /web/**/node_modules/**/* && cp reproducible-source.zip "${{ needs.create-nightly-release.outputs.package_prefix }}-reproducible-source.zip"
+        run: |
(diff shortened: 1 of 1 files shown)


Dinnerbone (approved):

danielhjacobs:
Tangential for a separate PR, this should result in failure and we should add a Discord notification on failure:

https://github.com/ruffle-rs/ruffle/blob/8071a14d4dfb9f02895e3d737d90e229c744f864/web/packages/extension/tools/sign_xpi.js#L95

torokati44:
Yep, and eventually we should drop `sign-addon`, which is now deprecated (and brings in these two security bulletins: https://github.com/ruffle-rs/ruffle/security/dependabot/44, https://github.com/ruffle-rs/ruffle/security/dependabot/52); and use AMO's v5 submission API api, which is now complete and supported. And/Or maybe `web-ext`, if that's easier.

More info:
https://www.npmjs.com/package/sign-addon
https://blog.mozilla.org/addons/2022/03/17/new-api-for-submitting-and-updating-add-ons/
https://extensionworkshop.com/documentation/develop/getting-started-with-web-ext/#installation-section

Aaron1011 (approved):

danielhjacobs (approved):

PR #15077, opened 2024-02-02
text: Add support for relative font sizes in HTML

kjarosh:
Add support for relative font sizes in HTML, e.g. `<font size="+5">text</font>`.


core/src/html/text_format.rs
```diff
@@ -717,7 +717,34 @@ impl FormatSpans {
                             }
 
                             if let Some(size) = attribute(b"size") {
-                                format.size = size.parse().ok();
+                                // if the numer starts with + or -, the size is relative
+                                let (prefix, size) = if size.starts_with(&[b'+', b'-'][..]) {
```
adrian17, line 721: To be sure, how does this work? And what does `[..]` do here?
Intuitively, with slice/str APIs, pretty sure this would be equivalent to `starts_with(b"+-")`, which checks for literal prefix "+-", not "+ or -".

core/src/html/text_format.rs
```diff
@@ -717,7 +717,34 @@ impl FormatSpans {
...
+                                let first_foreign_char = size
+                                    .find(|c| c < '0' as u16 || c > '9' as u16)
+                                    .unwrap_or(size.len());
+                                let size = &size[0..first_foreign_char];
+
+                                let size: Option<f64> = size.parse().ok();
+
+                                if let Some(size) = size
+                                    .and_then(|size| match prefix {
+                                        b'+' => format.size.map(|last_size| last_size + size),
```
adrian17, line 738: Just to be sure, this means the size is relative to previous element, not to the parent?

kjarosh: That means the size is relative to the parent. `format` is the current stack entry, a clone of the last stack entry (https://github.com/ruffle-rs/ruffle/pull/15077/files#diff-0313d265738726ac783fa404fb5ffff0bf42894a2a806e514185620b9a63251fR673), so I'm just applying the size difference to the would-be size (basically doing `format.size += size`/`format.size -= size` but with more checks).

In order to make the size relative to the previous element I would have to use `spans` here instead of the format stack.

core/src/html/text_format.rs
kjarosh: According to the `Pattern` docs:

> - `&[u8]` searches for any of the given LATIN1 code units.

And `[..]` is to make a slice of `u8` (to fulfill the `&[u8]` requirement). I did it this way because I saw it somewhere, I'm not sure if there's a clearer way

adrian17: Oh OK, I think I get it now.
Anyway, I think `&[b'+', b'-']` is already a `&[u8]`? Does this work without the `[..]`?

core/src/html/text_format.rs
adrian17: 👍

core/src/html/text_format.rs
kjarosh: That's exactly my line of thought I had when I first saw this syntax :) Unfortunately `&[b'+', b'-']` is `&[u8, 2]` and it does not work.

I basically have to give a hint to rustc that it should be `&[u8]`. Alternatively, I could do `&[b'+', b'-'] as &[u8]`, but the `[..]` syntax was used multiple times throughout ruffle, and the `as` version has never been used.

core/src/html/text_format.rs
```diff
@@ -717,7 +717,34 @@ impl FormatSpans {
...
+                                        b'-' => format.size.map(|last_size| last_size - size),
+                                        _ => Some(size),
+                                    })
+                                    .map(|size| size.clamp(1.0, 127.0))
```
kjarosh, line 742: Also, as a side note, the font size seems to be an integer, and programmatically may be set to a value outside of the `[1,127]` range. I'll probably tackle this behavior in a near future

adrian17 (approved):

core/src/html/text_format.rs
```diff
@@ -717,7 +717,34 @@ impl FormatSpans {
                             }
 
                             if let Some(size) = attribute(b"size") {
-                                format.size = size.parse().ok();
+                                // if the number starts with + or -, the size is relative
+                                let (prefix, size) = if size.starts_with(&[b'+', b'-'][..]) {
+                                    (size.at(0) as u8, &size[1..])
+                                } else {
+                                    (0u8, &size[..])
```
evilpie, line 724: Let's use Some(char) and None.

kjarosh: Is `Option<u8>` okay? For char I would have to decode the text or do `as u8 as char`, but we know here that the first character is ASCII for sure

evilpie: Sounds good. `Option<u16>` would be an option, but that seems worse for matching later.

Dinnerbone (approved):
Thank you!

PR #15079, opened 2024-02-02
web: Explicit instruction how to use sudo for docker building

evilpie:
This was a problem in the last review. I am not super familiar with docker, but I assume this 
```bash
sudo docker cp $(sudo docker create ruffle-web-docker:latest):/ruffle/web/packages web/docker/docker_builds/packages
```
would be the right way to use the following with sudo:
```bash
docker cp $(docker create ruffle-web-docker:latest):/ruffle/web/packages web/docker/docker_builds/packages
```

### DIFF
--- a/web/packages/extension/tools/sign_xpi.js
+++ b/web/packages/extension/tools/sign_xpi.js
@@ -72,8 +72,12 @@ async function sign(
 We highly recommend using the Docker build workflow. You can invoke it using the following three commands:\n\
 \n\
 rm -rf web/docker/docker_builds/*\n\
+# Normally these commands:\n\
 docker build --tag ruffle-web-docker -f web/docker/Dockerfile .\n\
 docker cp $(docker create ruffle-web-docker:latest):/ruffle/web/packages web/docker/docker_builds/packages\n\
+# OR alternatively, if you have to use 'sudo docker', make sure to use $(sudo docker ...) for the second docker command:\n\
+sudo docker build --tag ruffle-web-docker -f web/docker/Dockerfile .\n\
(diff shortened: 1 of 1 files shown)


torokati44:
I hope `sudo` outputs to `stderr` if/when it asks for the password, otherwise, the `$()` substitution will not be reliable (if the previously entered password has already expired).

evilpie:
Maybe we could just split this up?

torokati44:
Plus I think the `docker build` command will also have to be `sudo`ficated, if the other two are?

torokati44:
Certainly, but if we ask the reviewer to do the container ID substitution, the instructions won't simply be "execute these commands in order" anymore, adding more points of potential human error.

torokati44:
> I hope `sudo` outputs to `stderr` if/when it asks for the password

It does, so nothing to worry about, it will keep working with the original nesting.

evilpie:
> > I hope `sudo` outputs to `stderr` if/when it asks for the password
> 
> It does, so nothing to worry about, it will keep working with the original nesting.

Okay. Let's keep this then.

torokati44:
Maybe the instructions could be something like:

> If you have to use `docker` via `sudo` - because your user is not part of the `docker` group, for example -, make sure to add the `sudo` prefix to **all three** `docker` commands, (`docker build`, `docker cp`, and the `docker create` within the `$(...)`)!


Whichever you like better.

torokati44:
> I would like to avoid the additional confusion

Fair. But I think we still shouldn't forget about mentioning the possibility of having to use `sudo` at the initial `docker build` command too...?

torokati44 (approved):
As always, let's hope for the best.

PR #15083, opened 2024-02-02
github: Add script and workflow to automatically update stub report

Aaron1011:
On every push to master, we now regenerate the stub report and check if it's changed from the one currently published on the website. If it has, we push a new commit to the website

### DIFF
--- a/.github/workflows/update_website_report.yml
+++ b/.github/workflows/update_website_report.yml
@@ -1,0 +1,24 @@
+name: Update AVM2 implementation report on website
+
+on:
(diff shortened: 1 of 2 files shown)


Dinnerbone:
I wonder if it'll be better to include the report.json as part of the release assets, and have the website just show that? Then we can even do cool things like graph the info over time - plus no commit spam in discord etc

Aaron1011:
Closing in favor of https://github.com/ruffle-rs/ruffle/pull/15084

PR #15096, opened 2024-02-03
text: Make enter add a new line in multiline fields

kjarosh:
For multiline fields, it is now possible to add a newline by pressing enter.


Dinnerbone (approved):
Thank you!

PR #15108, opened 2024-02-03
avm2: Implement selectedText in class flash.text.TextField

TFSMads:


Dinnerbone:
Hey, thank you for this! Is there any chance you can add a test, or would you like help with that?

PR #15115, opened 2024-02-04
core: Remove LoaderHandle when load is done

Aaron1011:
In almost all cases, we were letting completed LoaderHandles accumulate, leaking memory and wasting time in `preload_tick`

### DIFF
--- a/core/src/avm2/domain.rs
+++ b/core/src/avm2/domain.rs
@@ -24,6 +24,12 @@ use super::Avm2;
 #[collect(no_drop)]
 pub struct Domain<'gc>(GcCell<'gc, DomainData<'gc>>);
 
+impl std::fmt::Debug for Domain<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "Domain({:p})", self.0.as_ptr())
+    }
+}
+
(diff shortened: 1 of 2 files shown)


Dinnerbone (approved):

Dinnerbone:
Sorry, maybe reviewed too quickly.

I notice that some places have the potential to return early with `?` before getting to this line. Is that going to be an issue?

Is there another way we can do this?

core/src/avm2/domain.rs
```diff
@@ -24,6 +24,12 @@ use super::Avm2;
 #[collect(no_drop)]
 pub struct Domain<'gc>(GcCell<'gc, DomainData<'gc>>);
 
+impl std::fmt::Debug for Domain<'_> {
```
Lord-McSweeney, line 27: Accidentally added?

kmeisthax: My guess is Aaron added it, much like the `Debug` impls later on, while debugging the memory problems caused by not cleaning up loaders.

I'm personally fine with merging the debug impls alongside everything else. Not being able to trace something because someone forgot to add a `Debug` impl is frustrating.

kmeisthax:
I'm wondering if we can abuse RAII/`Drop` to make loaders close themselves regardless of whether or not they exit normally or not.

Issue #2351, opened 2021-01-09
HTML EditText with only an empty html tag can't be edited


MrCheeze:
The following flash shows this issue: https://homestuck.net/games/cursorstuck.swf.
Click New Game -> space four times -> Enter Name -> Click the text box on the top of the field and start typing.
In flash player, it's possible to type text in the field, but in Ruffle it's not possible.

The EditText field is defined as follows in the flash:
```[
xmin -40
ymin -40
xmax 11519
ymax 1059
wordwrap 1
multiline 1
html 1
font 57
height 900
color #ff000000
align left
leftmargin 0
rightmargin 0
indent 0
leading 40
]<p align="left"></p>
```

It seems only this specific combination triggers the issue. If the \<p> tag has non-empty text inside it, then it's possible to edit. If the tag is removed entirely so that the initial value is completely empty, then it's also possible to edit.

MrCheeze:
Narrowed down the issue somewhat.

In both FP and Ruffle, the textbox has two lines of text. The first is visible and the second is cutoff below. On Ruffle only, clicking the textbox incorrectly puts the cursor in the second line.

PR #15116, opened 2024-02-04
core: Improve placement of cursor when clicking in textbox

MrCheeze:
Currently when clicking in a textbox, the cursor will only be placed correctly if you click exactly on top of a letter. The basic purpose of this commit is to make it so that clicking in the surrounding margin will put the cursor in the closest available position, as in Flash Player.

The logic is a bit complex because a single row of text can contain multiple layout boxes, each with a different Y offset (but all with the same Y extent). To be accurate, we need to treat each layout box in the row "as if" it had the same Y offset that the tallest box in its row has.

This commit also contains a fix for an issue where lower_from_text_spans was passing the wrong text strings and indexes to fixup_line for newlines, which needed to be fixed in order to be able to click to place the cursor in an empty row that was created by newlines.

I don't know if tests are possible for this function, but [here's a swf](https://mega.nz/file/TMlGxLRb#9X1I-rywW7LQoYnhY8w1Daa2dAv5SxyfqeXF-nKk_Xw) that I manually tested with.

Oh yeah, and this fixes https://github.com/ruffle-rs/ruffle/issues/2351 (possibly others)


evilpie:
> I don't know if tests are possible for this function, but [here's a swf](https://mega.nz/file/TMlGxLRb#9X1I-rywW7LQoYnhY8w1Daa2dAv5SxyfqeXF-nKk_Xw) that I manually tested with.

You can simulate clicks at specific points using `input.json` e.g. https://github.com/ruffle-rs/ruffle/blob/master/tests/tests/swfs/avm2/mouse_pick_text/input.json.

MrCheeze:
Amended the PR by adding a test, the actual logic is unchanged. Should be reviewable now.

Dinnerbone (approved):

PR #15117, opened 2024-02-04
avm2: Implement TextField::GetRawText

TFSMads:


evilpie:
Hi! Thank you for your contribution. It seems like this method is not document and your current implementation is identical to the `text` getter. Did you investigate the difference between them? Could you provide a few tests for this? (I imagine especially around lines and text .vs html)

PR #1520, opened 2020-11-07
improve the Spanish translation

ousia:
@Justin-CB did a great work translating Ruffle into Spanish.

I hope some polishing could make it better.

I also replaced `...` with `…` (the proper Unicode character) in both English and Spanish.

BTW, are there other translation files than this one?

Many thanks for your good work.

PR #15119, opened 2024-02-04
avm2: Fix `in` operator not handling attributes in XML object

sleepycatcoding:
This makes expressions like
```
var x = <root val="hello"/>
"@val" in x; // Was false before this PR.
```
work.


core/src/avm2/object/xml_object.rs
```diff
@@ -365,6 +365,15 @@ impl<'gc> TObject<'gc> for XmlObject<'gc> {
         self.0.read().base.has_own_dynamic_property(name)
     }
 
+    fn has_property_via_in(
+        self,
+        activation: &mut Activation<'_, 'gc>,
+        name: &Multiname<'gc>,
+    ) -> Result<bool, Error<'gc>> {
+        let multiname = handle_input_multiname(name.clone(), activation);
+        Ok(self.has_own_property(&multiname))
```
evilpie, line 374: This doesn't seem correct. `in` needs to check whole chain. Consider `"propertyIsEnumerable" in xml` vs `xml.hasOwnProperty("propertyIsEnumerable")`.

evilpie (changes requested):

core/src/avm2/object/xml_object.rs
sleepycatcoding: Would changing to `self.has_property` be enough?

sleepycatcoding: > This doesn't seem correct. `in` needs to check whole chain. Consider `"propertyIsEnumerable" in xml` vs `xml.hasOwnProperty("propertyIsEnumerable")`.

At the moment `"propertyIsEnumerable" in xml` seems to return `false` on master too.

evilpie: > Would changing to `self.has_property` be enough, like how `has_property_via_in` does by default?

I think that might work? I think it should end up calling the XML's `has_own_property`.

> At the moment `"propertyIsEnumerable" in xml` seems to return `false` on master too.

It's `true` in Flash.

tests/tests/swfs/avm2/xml_has_property_via_in/Test.as
```diff
@@ -0,0 +1,38 @@
...
+trace("@a" in a);
+trace("");
+
+trace("0 in a");
+trace(0 in a);
+trace("");
+
+trace("1 in a");
+trace(1 in a);
+trace("");
```
evilpie, line 38: Can you add a test for isPropertyEnumerable?

sleepycatcoding: This currently fails, but seems to work with your PR #15120.

evilpie: We can land #1520 first and use this as a test for it.

PR #15120, opened 2024-02-04
 avm2: has_property should do proto.has_property instead of has_own_property

evilpie:
I noticed this while reviewing #15119. Need to test this though.


evilpie:
This now block #15119, because of a test added there.

Dinnerbone (approved):

PR #15129, opened 2024-02-05
avm2: Parse data for flash.net.navigateToURL (close #13898)

### LINKED ISSUE #13898: Ruffle does not append ? paremeters to urls properly
### Describe the bug

I tested [Kingdom Rush](https://armorgames.com/play/12141/kingdom-rush?flash) through the extension dinnerbone sent on the discord, who included fixes to make AGIv1 work
When I click the buy Premium buton, it sends me to 
https://armorgames.com/purchase/

### Expected behavior

The button to Buy Premium would send me to 
https://armorgames.com/purchase/?sku=KingdomRush%2DPremium&username=Fancy2209

### Content Location

https://cache.armorgames.com/files/games/kingdom-rush-12141.swf?v=1379178890

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Mozilla Firefox Developer Edition 120.0b7

### Additional information

_No response_

svitkin:
Fix bug where nothing was being passed as data to the optional last parameter of the navigator backend's `navigate_to_url` method. 

Adds a helper function to convert avm2 `Object` to `IndexMap` for compatibility with the backend's call signature. Do not think this will work when the `Object` is a `bytearray` but that would require changing the backend's call signature so that can be part of a different PR if required.

First PR here so a couple questions!
- Should I move `object_to_index_map` to a more reusable location? Should it just be a method on `Object`?
- Is there a regression test I can add? I tried running the related issue's `Test.swf` in a debug Flash player but I'm not sure what traces would be helpful to add to show that this is working in Ruffle like in the Flash Player - haven't found an easy way to capture and log the url after the navigation is complete


core/src/avm2/globals/flash/net.rs
```diff
@@ -39,10 +64,20 @@ pub fn navigate_to_url<'gc>(
         )?)),
         url => {
             let url = url.coerce_to_string(activation)?;
+            let method = request
+                .get_public_property("method", activation)?
+                .coerce_to_string(activation)?;
+            let method =
+                NavigationMethod::from_method_str(&method).unwrap_or(NavigationMethod::Get);
```
evilpie, line 71: After rebasing on #15145 this can just be `unwrap` or `expect`.

svitkin: Good call! Fixed in https://github.com/ruffle-rs/ruffle/pull/15129/commits/7fa96c36656b4d68bb9b721400d88f42fe6190b6

tests/tests/swfs/avm2/net_navigateToURL/Test.as
```diff
@@ -0,0 +1,34 @@
...
+	import flash.display.MovieClip;
+	public class Test extends MovieClip {
+        public function Test()
+        {
+            var request:URLRequest = new URLRequest("https://example.com/purchase/");
+			var variables:URLVariables = new URLVariables();
+            var TEST_PARAM_NAME:String = "sku";
+            var cases:Array = [
+                ["POST", "Test"],
+                ["POST", "sku=Test"],
```
evilpie, line 14: I think just doing `["POST", "Test"], ["POST, variables], ...` would be move obvious.

tests/tests/swfs/avm2/net_navigateToURL/Test.as
```diff
@@ -0,0 +1,34 @@
+package {
+    import flash.net.URLRequest;
```
evilpie, line 2: The indentation is off.

tests/tests/swfs/avm2/net_navigateToURL/Test.as
```diff
@@ -0,0 +1,34 @@
...
+                var method:String = case_tuple[0];
+                var data:String = case_tuple[1];
+                request.method = method;
+                if (data.indexOf(TEST_PARAM_NAME) != -1) { // pass data as string
+                    request.data = data;
+                } else { // pass data as object
+                    variables.sku = data;
+                    request.data = variables;
+                }
+                trace("// Method: ", method);
```
evilpie, line 28: Super nit, but the you remove the space at the end of the string to prevent the double spaced.

evilpie:
Does this test work for you? (I guess it might work on Windows with case-insensitivity?) It fails on CI, because it should be named `test.swf` and not `Test.swf`.

tests/tests/swfs/avm2/net_navigateToURL/Test.as
svitkin: Fixed in https://github.com/ruffle-rs/ruffle/pull/15129/commits/89da73ada2f1474c3179bea0ed321ae76b842f7f

tests/tests/swfs/avm2/net_navigateToURL/Test.as
svitkin: Fixed in [cbc36f7](https://github.com/ruffle-rs/ruffle/pull/15129/commits/cbc36f794828b024c059eb86929dad8fe6e1fbcd)

svitkin:
> Does this test work for you? (I guess it might work on Windows with case-insensitivity?) It fails on CI, because it should be named `test.swf` and not `Test.swf`.

Not sure why this is happening... locally it is named `test.swf`

core/src/avm2/globals/flash/net.rs
```diff
@@ -15,6 +17,47 @@ pub mod socket;
...
+    activation: &mut Activation<'_, 'gc>,
+    url: &String,
+    data: &Value<'gc>,
+) -> Result<(String, IndexMap<String, String>), Error<'gc>> {
+    let mut url = url.to_string();
+    let obj = data.coerce_to_object(activation)?;
+    let vars = object_to_index_map(activation, &obj).unwrap_or_default();
+    let str_data = data.coerce_to_string(activation)?.to_string();
+    if vars.is_empty() && !str_data.is_empty() {
+        if !url.contains('?') && !str_data.contains('?') {
```
evilpie, line 53: You are basically re-implementing https://github.com/ruffle-rs/ruffle/blob/master/core/src/avm2/globals/flash/display/loader.rs#L172 now.

In my local testing your `!str_data.contains('?')` check while somewhat logically, doesn't actually happen in Flash.

svitkin: And this can't fully use the loader.rs implementation because of the backend's call signature.

core/src/avm2/globals/flash/net.rs
```diff
@@ -15,6 +17,47 @@ pub mod socket;
...
+}
+
+fn parse_data<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    url: &String,
+    data: &Value<'gc>,
+) -> Result<(String, IndexMap<String, String>), Error<'gc>> {
+    let mut url = url.to_string();
+    let obj = data.coerce_to_object(activation)?;
+    let vars = object_to_index_map(activation, &obj).unwrap_or_default();
```
evilpie, line 50: Okay one last thing. I am pretty sure we are only supposed to handle URLVariables as a map and coerce everything else to a string.

```rust
let urlvariables = activation
                                .avm2()
                                .classes()
                                .urlvariables
if data.is_type_of(urlvariables.inner_class_definition(), &mut activation.context) {
...
  object_to_index_map()
...
} else {
 ...
  coerce_to_string()
  ...
}
```

evilpie:
Can you try doing this https://stackoverflow.com/questions/1793735/change-case-of-a-file-on-windows ? The pull request definitely has an upper case `Test.swf`, so that needs to be fixed.

core/src/avm2/globals/flash/net.rs
svitkin: good call! Was not aware you could check type explicitly like that. Changed in https://github.com/ruffle-rs/ruffle/pull/15129/commits/9472c52dcbb84b0911b2bf31105fe70b71e9c501

evilpie (approved):

evilpie:
:tada: Thanks for all your work! Could you squash these changes into two commits for `avm2:` and `tests:` before landing?

svitkin:
> 🎉 Thanks for all your work! Could you squash these changes into two commits for `avm2:` and `tests:` before landing?

Thank you for your help! Should be squashed (but let me know if not, have not worked with rebase this extensively before). Also I see `clippy` is failing but it seems unrelated to my changes.

lvyitian:
> > 🎉 Thanks for all your work! Could you squash these changes into two commits for `avm2:` and `tests:` before landing?
> 
> Thank you for your help! Should be squashed (but let me know if not, have not worked with rebase this extensively before). Also I see `clippy` is failing but it seems unrelated to my changes.

That's https://github.com/Smithay/wayland-rs/issues/698

PR #15169, opened 2024-02-08
Deserialize AMF dictionary

wsxarcher:
Deserialize dictionary

### DIFF
--- a/core/src/avm2/amf.rs
+++ b/core/src/avm2/amf.rs
@@ -1,6 +1,7 @@
 use std::rc::Rc;
 
 use crate::avm2::bytearray::ByteArrayStorage;
+use crate::avm2::object::dictionary_allocator;
 use crate::avm2::object::{ByteArrayObject, TObject, VectorObject};
 use crate::avm2::vector::VectorStorage;
 use crate::avm2::ArrayObject;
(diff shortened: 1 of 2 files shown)


core/src/avm2/amf.rs
```diff
@@ -434,8 +434,44 @@ pub fn deserialize_value<'gc>(
...
+                .classes()
+                .dictionary
+                .construct(activation, &[(*has_weak_keys).into()])?;
+            let dict_obj = obj.as_dictionary_object().unwrap();
+
+            for (key, value) in values {
+                let key = deserialize_value(activation, key)?;
+                let value = deserialize_value(activation, value)?;
+
+                match key {
```
adrian17, line 449: how about

```rs
if let Value::Object(key) = key {
    // set_property_by_object
} else {
    // coerce_to_string
}

File: core/src/avm2/amf.rs [hunk 2]
```diff
@@ -434,8 +434,44 @@ pub fn deserialize_value<'gc>(
...
+                    | Value::Number(..)
+                    | Value::Integer(..)
+                    | Value::Null
+                    | Value::Undefined => {
+                        let key_string = key.coerce_to_string(activation)?;
+                        dict_obj.set_public_property(key_string, value, activation)?;
+                    }
+                }
+            }
+            dict_obj.value_of(activation.context.gc_context)?
```
adrian17, line 464: `dict_obj.into()` should be enough here

adrian17 (approved):
Approved, assuming minor fixes

PR #15140, opened 2024-02-06
web,ci: Recommend and test with Node.js 20 LTS and 21 stable

torokati44:

PR #15217, opened 2024-02-14
ci: Remove Rust version from the web CI matrix - it was always stable

torokati44:
No change in functionality intended, only a little cleanup.


torokati44:
Ah, this will require another repo admin change, just like #15140.

adrian17 (approved):

PR #15233, opened 2024-02-15
ci: Fix failing visual tests

Dinnerbone:
Two failures:
- mesa-vulkan-driver just updated from 23.0.4 to 23.2.1, which brought in a new bug that fails any test with a blur filter. Update to a newer one from PPA (reverts #8708) to see if it helps.
- Don't upload artifacts with the same name, overwriting them

### DIFF
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -60,7 +60,6 @@ jobs:
         run: |
           sudo add-apt-repository ppa:kisak/kisak-mesa -y
           sudo apt-get update
-          sudo apt-get dist-upgrade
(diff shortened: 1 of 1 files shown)


torokati44:
Would the `upload-artifact/merge` action not work?
Or do we actually prefer this?

Dinnerbone:
> Would the `upload-artifact/merge` action not work? Or do we actually prefer this?

It downloads N artifacts and uploads 1. It's not going to work if we're failing to upload artifacts at all.

torokati44:
Ah, okay, I thought it adds new files into existing artifact zips.

torokati44 (approved):
Yay!

PR #15238, opened 2024-02-16
web: Fix version seal path

Dinnerbone:
I don't know where/when this broke, but we started reading the version seal from the wrong place - making it never used.

This is one small step towards getting reproducible builds again.


torokati44 (approved):
I don't know what `__dirname` is, or where it comes from, but I believe you.

Also - maybe the version seal gets _put_ to the wrong place now...? Just a guess...

Dinnerbone:
> Also - maybe the version seal gets put to the wrong place now...? Just a guess...

I thought that too but the extension script looks for it in the place it is now, so it's just reading that's wrong.

PR #15241, opened 2024-02-16
ci: Compile Rust tests in release mode

torokati44:
Let's see whether this is faster now...

Afterwards, I think it would still be favorable to create a custom Rust profile for CI, where optimizations are enabled, but debug checks are not disabled.

Also note that our cargo cache is primed with debug artifacts, so it's not trivial to accurately measure this.

### DIFF
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -77,13 +77,13 @@ jobs:
         if: runner.os != 'macOS'
         # TODO: Disallow retries in general once we can allow only after SIGABRT.
         # See: https://github.com/nextest-rs/nextest/issues/1172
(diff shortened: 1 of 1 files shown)


torokati44:
Current test times with empty cache:
image

Cache sizes:
770, 770, 770, 550, 720 MB

torokati44:
After adding a new build profile, as follows:
```toml
[profile.ci]
inherits = "release"
opt-level = 2
debug-assertions = true
overflow-checks = true
incremental = true
```

For reference, we already had this:
```toml
[profile.release]
panic = "abort"
```

I get this on an empty cache:

image

And on a primed cache:

image

Cache sizes:
580, 650, 570, 570, 490 MB

PR #15242, opened 2024-02-16
chore: Compress all image test expected.png images with oxipng

Dinnerbone:
This reduces the size of our tests folder, which will in turn make the upcoming swf-test-on-web wasm module smaller (goes from 77MB to 70MB)


evilpie (approved):
I visually compared a handful of the images and couldn't see any difference.

torokati44:
> I visually compared a handful of the images and couldn't see any difference.

I think it's expected that there is _no_ difference at all, PNG is lossless after all. The size difference is supposed to come from better compression strategies only.

I honestly don't know whether doing this is worth it - if you feel like the tester WASM size being a tiny bit smaller is that important, even if it makes the git repo quite a bit larger, I don't mind the latter. We're still far from it being a gigabyte or something...

MrCheeze:
What about new pngs saved by the test framework code - does it already use oxipng? It would be less than ideal to have the PR merged but then people still check in less-compressed pngs for future tests.

torokati44:
While it's sometimes easier to manipulate the expected images manually to average out the outputs on different platforms/backends, for the cases where this isn't necessary, that would be a great idea indeed!
Even if we decide not to merge this over repo size concerns, going forward.

evilpie:
I still don't understand what the repo size concerns are? Is there some GitHub limit we are worried about?

Dinnerbone:
I'll close this for now and in future come back and just make newly generated images crushed

PR #15243, opened 2024-02-16
web: Add swf-tests package, for running the swf test suite in a browser

Dinnerbone:
No way to run this automatically yet - but I decided I could iterate on this indefinitely, but it'd already be useful in this form today. I'll follow up in future PRs, but at least this won't bit rot if it lands in the repo.

This is part of a larger effort to eventually get webgl (and maybe canvas) tested, so we can start making those more stable and feature complete.

This tool has already found some bugs, which have been fixed so far (mostly 32-bit vs 64-bit ByteArray stuff)

Example is at https://dinnerbone.github.io/ruffle-swf-tests-on-web/


torokati44:
Top is Firefox, bottom is Chrome, on the same (Linux) machine:
image

PR #15251, opened 2024-02-16
web: Add `chromedriver` to `devDependencies`

torokati44:
...in an attempt to fix the umpteenth way CI poops itself, as suggested by the error message here:
https://github.com/ruffle-rs/ruffle/actions/runs/7936480638/job/21671854382?pr=15250


torokati44:
Sorry, this was not it.

Issue #15154, opened 2024-02-07
House (Newgrounds Version) is unable to boot past the "Play" screen


Spikestuff:
### Describe the bug

This is an issue with the [Newgrounds version](https://www.newgrounds.com/portal/view/683412?emulate=flash) specifically.
The Armor Games version can run the title as it doesn't have the splash.

![image](https://github.com/ruffle-rs/ruffle/assets/12412764/763add7c-6b70-42a2-ad9e-29f2edd880e5)

### Expected behavior

A still from the game booted up with the Armor Games version:
![image](https://github.com/ruffle-rs/ruffle/assets/12412764/1a559847-2597-424a-8361-46a40b2db833)

### Content Location

Newgrounds version: https://uploads.ungrounded.net//683000//683412_newgroundsversion.swf
Armor Games version: https://cache.armorgames.com/files/games/house-18108.swf

### Affected platform

Desktop app

### Operating system

Windows 10

Aaron1011:
This game actually does boot - however, it first tries to preload over 1000 enormous bitmaps (which I checked are all actually present in the SWF), allocating 16.6 GB of ram in the desktop player. If your system has enough ram, it will actually boot after this.

We need to do some kind of lazy loading of these bitmaps

PR #15270, opened 2024-02-18
core: Implement lazy decoding of bitmaps

Aaron1011:
We hit a pathological case in House
(https://github.com/ruffle-rs/ruffle/issues/15154), where eagerly decoding bitmaps during preloading results in over 10GB of ram being used.

With this PR, we store the compressed bitmap, and only decode it each time we instantiate it. In order to support bitmap fills, we store the decoded width/height and a lazily-initialized GPU handle in `Character::Bitmap`

### DIFF
--- a/core/src/character.rs
+++ b/core/src/character.rs
@@ -1,10 +1,14 @@
+use std::cell::RefCell;
+
 use crate::backend::audio::SoundHandle;
 use crate::binary_data::BinaryData;
 use crate::display_object::{
-    Avm1Button, Avm2Button, Bitmap, EditText, Graphic, MorphShape, MovieClip, Text, Video,
+    Avm1Button, Avm2Button, BitmapClass, EditText, Graphic, MorphShape, MovieClip, Text, Video,
 };
 use crate::font::Font;
-use gc_arena::Collect;
+use gc_arena::{Collect, GcCell};
+use ruffle_render::bitmap::{BitmapHandle, BitmapSize};
+use swf::DefineBitsLossless;
 
 #[derive(Clone, Collect, Debug)]
 #[collect(no_drop)]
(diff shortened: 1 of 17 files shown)


core/src/character.rs
```diff
@@ -1,18 +1,31 @@
...
     EditText(EditText<'gc>),
     Graphic(Graphic<'gc>),
     MovieClip(MovieClip<'gc>),
-    Bitmap(Bitmap<'gc>),
+    Bitmap {
+        #[collect(require_static)]
+        compressed: CompressedBitmap,
+        /// A lazily constructed GPU handle, used when performing fills with this bitmap
+        #[collect(require_static)]
+        handle: RefCell<Option<BitmapHandle>>,
```
Dinnerbone, line 24: I feel like this should be an enum. Either we have the handle and we don't care about compressed, or we have compressed and not yet a handle.

Or are there cases where we *do* care about compressed when we've already decompressed it?

Aaron1011: We could have a bitmap that's only ever used as a gradient fill, which would I ly require us to store the decompressed texture in gpu memory (via the BitmapHandle)

Dinnerbone: Sorry I don't understand (I tried unsuccessfully to decipher the typo - sorry! 😅)

Aaron1011: Oops, updated the comment

Dinnerbone: Sorry still confused :D

In that case why would you still want the `CompressedBitmap` around?

I'm thinking it's something like:

```rust
enum BitmapType {
    Compressed(CompressedBitmap),
    Decompressed{ handle: RefCell<Option<BitmapHandle>>, avm2_bitmapdata_class: GcCell<'gc, BitmapClass<'gc>> },
}
```

Aaron1011: How would we handle instantiating the bitmap in your proposal?

Dinnerbone (approved):

core/src/character.rs
Dinnerbone: I'm not sure, but it really feels weird that it'd need the compressed data when we've already uncompressed it...

Not blocker for this PR, but I'd like to come back to this

PR #15277, opened 2024-02-19
web: Don't generate random name of wasm files in extension

Dinnerbone:
Instead of [big random hash] it'll be the actual name of the wasm file.
This way `player.js` will not make a diff because the filename changed, and it'll also be clear which wasm file is which


torokati44:
I was heckin confuse as to what exactly are `[ext]` and `[query]` (assumed the first one to be `("_extensions"|"")`), but turns out, it's the file extension... 😄

torokati44 (approved):

PR #15280, opened 2024-02-20
core: Use latest fluent-templates `master` for reproducible builds

Aaron1011:


core/Cargo.toml
```diff
@@ -53,7 +53,7 @@ clap = { version = "4.5.1", features = ["derive"], optional=true }
 realfft = "3.3.0"
 hashbrown = { version = "0.14.3", features = ["raw"] }
 scopeguard = "1.2.0"
-fluent-templates = "0.8.0"
+fluent-templates = { git = "https://github.com/XAMPPRocky/fluent-templates", branch = "master" }
```
torokati44, line 56: It would be nice to leave a comment here, linking to the PR that we need it for, and that the we can switch back to the next release after 0.8.0.

torokati44:
`deny.toml` will also have to be updated to include this GH user as allowed source (at the end of the file).

core/src/i18n.rs
```diff
@@ -10,10 +10,7 @@ static_loader! {
 }
 
 pub fn core_text(language: &LanguageIdentifier, id: &str) -> String {
-    TEXTS.lookup(language, id).unwrap_or_else(|| {
```
adrian17, line 13: looking at how the non-fallible function works, are you sure we don't actually want a `.try_lookup().unwrap_or_else()`?

adrian17 (approved):
approved the version bump, assuming the comment is added and that failure path is readded (or veto'd ;D )

core/src/i18n.rs
Aaron1011: Fixed

PR #15293, opened 2024-02-22
ci: Match cargo features when testing the extension builder Dockerfile

torokati44:
Maybe this will bring us closer to reproducibility. Thanks for noticing, @Aaron1011!


torokati44:
This ended up being included in https://github.com/ruffle-rs/ruffle/pull/15294.

PR #15310, opened 2024-02-25
desktop: Add socket tests and fix flush on close

kjarosh:
This patch makes sure that any pending data is sent before the socket closes. Without this fix Ruffle ignored the data sent right before calling `socket.close`.

### DIFF
--- a/desktop/src/backends/navigator.rs
+++ b/desktop/src/backends/navigator.rs
@@ -574,20 +574,20 @@ impl<F: FutureSpawner> NavigatorBackend for ExternalNavigatorBackend<F> {
                 let mut pending_write = vec![];
 
                 loop {
-                    loop {
+                    let close_connection = loop {
                         match receiver.try_recv() {
                             Ok(val) => {
                                 pending_write.extend(val);
                             }
+                            Err(TryRecvError::Empty) => break false,
                             Err(TryRecvError::Closed) => {
                                 //NOTE: Channel sender has been dropped.
-                                //      This means we have to close the connection.
-                                drop(write);
-                                return;
+                                //      This means we have to close the connection,
+                                //      but not here, as we might have a pending write.
+                                break true;
                             }
-                            Err(_) => break,
                         }
-                    }
+                    };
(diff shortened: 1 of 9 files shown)


Dinnerbone:
Do you think it's possible to modify (or write) a test to show this, or is it too edge case?

Thank you though!

kjarosh:
You know... everything's possible, some things are just considerably more expensive :sweat_smile: 

According to my knowledge there are no existing tests (I searched for `test` inside the desktop package and found nothing). I'll look into writing a test and make this a draft for now.

kjarosh:
I managed to test this behavior (see `test_socket_flush_before_close`).

* I had to make the code more testable (using async channels & FutureSpawner)
* I used tokio for testing as I did not find any other runtime which supported local future spawning in tests

As a bonus I wrote some other tests and got rid of the busy waiting in `connect_socket` so that there's no 10 ms sleep anymore. (Can *you* imagine being woken up every 10 ms?? I would be very upset!)

desktop/src/backends/navigator.rs
```diff
@@ -621,3 +621,231 @@ impl<F: FutureSpawner> NavigatorBackend for ExternalNavigatorBackend<F> {
...
+    }
+
+    #[macro_rules_attribute::apply(async_test)]
+    async fn test_socket_deny() {
+        let addr = SocketAddr::from_str("127.0.0.42:65000").expect("dummy address");
+        let (_client_write, client_read) = connect_test_socket(addr, TIMEOUT, false);
+        assert_next_socket_actions!(
+            client_read;
+            Connect(dummy_handle!(), ConnectionState::Failed),
+        );
```
sleepycatcoding, line 804: Since this fails the same way if a server is not listening on that address, how about starting a server and making sure socket allow does what is expected.

kjarosh: That's a good catch, should be good now :)

Dinnerbone (approved):
You're amazing. Thank you!

PR #15316, opened 2024-02-25
 core,avm2: Make Timer.setDelay not change the timer ID

### LINKED ISSUE #15005: `flash.utils.Timer` not handling setting the `delay` properly
### Describe the bug

Seems the Timer class is not properly handling setting the `delay` property while it's running, in particular during the event handler.

Here's a simple repro

```
package
{
	import flash.display.Sprite;
	import flash.events.Event;
	import flash.utils.Timer;
	import flash.events.TimerEvent;
	
	public class Main extends Sprite 
	{
		private var timer:Timer;
		private var cnt:int = 0;
		
		public function Main() 
		{
			if (stage) init();
			else addEventListener(Event.ADDED_TO_STAGE, init);
		}
		
		private function timerEvent(e:TimerEvent):void
		{
			cnt++;
			trace("timerEvent", cnt);
			timer.delay = 500;         //  <<< Problem
		}

		private function timerComplete(e:TimerEvent):void
		{
			trace("timerComplete", cnt);
		}

		private function init(e:Event = null):void 
		{
			removeEventListener(Event.ADDED_TO_STAGE, init);

			// entry point
			trace("Starting");
			
			timer = new Timer(500, 3);
			timer.addEventListener(TimerEvent.TIMER, timerEvent);
			timer.addEventListener(TimerEvent.TIMER_COMPLETE, timerComplete);
			timer.start();
		}		
	}
}
```

In the FlashPlayer, this produces the expected sequence:
```
Starting
timerEvent 1
timerEvent 2
timerEvent 3
timerComplete 3
```

However, in Ruffle it generates this:
```
Starting
timerEvent 1
timerEvent 2
timerEvent 3
timerComplete 3
timerEvent 4
timerComplete 4
```

Note, if I don't set the `delay` in the code, everything works as expected.

I tried to debug it, and it seems it's related to the fact that `Timer::set delay` is actually running a `stop/start` sequence, which results in a new id assigned to Timer (through `timers.addTimer`), and that confuses the logic inside `update_timers` (`expected_id` is different). Or so I think 😄  I'm new to Ruffle so this is just based on some trace-based debugging.

### Expected behavior

The timer completion should be properly handled

### Content Location

Repro attached

### Affected platform

Desktop app

### Operating system

MacOS

### Browser

_No response_

### Additional information

_No response_

torokati44:
Fixes #15005.


torokati44:
To reviewers:
Should the entirely internal `updateDelay` function be called `_updateDelay` instead, for example?
Is the test sufficient as is, or should the non-changing ID also be explicitly verified?

Lord-McSweeney:
From what I can see our "standard" is to have `internal` functions and fields to be prefixed with an underscore, while `private` functions (called from AS) are not, so not putting a semicolon is fine.

torokati44:
> so not putting a semicolon is fine.

I assume you meant underscore. And okay, thanks!

BTW what's the difference between `internal` and `private`? 🤔

Lord-McSweeney:
Internal namespaces are only accessible within the package, while private namespaces are only accessible within the class (AFAIK both easy to circumvent with P-code editing). Internal fields are easily accessible from native Rust, so they're used for AS-Rust interaction where playerglobals AS needs to interface with native Rust.

Dinnerbone (approved):

core/src/avm2/globals/flash/utils/Timer.as
```diff
@@ -32,11 +32,12 @@ package flash.utils {
 			this.checkDelay(delay);
 			this._delay = value;
 			if (this.running) {
-				this.stop();
-				this.start();
+				this.updateDelay(delay);
```
torokati44, line 35: This parameter shouldn't be here. It's not used.

core/src/avm2/globals/flash/utils/Timer.as
```diff
@@ -32,11 +32,12 @@ package flash.utils {
...
 			}
 		}
 
+		private native function updateDelay(value:Number):void;
```
torokati44, line 39: TODO: drop that parameter

adrian17:
> Internal namespaces are only accessible within the package, while private namespaces are only accessible within the class (AFAIK both easy to circumvent with P-code editing).

Nit: internal namespaces can technically be accessed from anywhere (with P-code editing), while private namespaces can't be accessed by code from outside their translation unit at all, due to how private namespaces are accessed.
But I think AS3's intention is more or less as you wrote it :D

PR #15325, opened 2024-02-26
ci: Set CARGO_NET_GIT_FETCH_WITH_CLI=true on Windows (and incidentally on Linux)

torokati44:
This step takes almost 3 minutes on our Windows CI runs.
Inspiration: https://github.com/rust-lang/cargo/issues/9167

I could have also added this to `.cargo/config.toml`, as with https://github.com/ruffle-rs/ruffle/pull/15324 (or vice versa, set that using an envvar); but considered that to be more generally useful and less risky, while this may not be useful for everyone, and adds a dependency to `git` being available as a command.

### DIFF
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -31,7 +31,7 @@ jobs:
 
   build:
     needs: changes
-    if: needs.changes.outputs.should_run== 'true'
+    if: needs.changes.outputs.should_run == 'true'
     name: Test Rust ${{ matrix.rust_version }} / ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


torokati44:
Well, it only brought that step down from ~135 to ~110 seconds.
I'm not sure it's worth it then.
But maybe...?

torokati44:
No substantial speedup observed... :/
https://github.com/torokati44/ruffle/actions/runs/8051740050/job/21990316922

That step is either "near-instant", or takes about 2 minutes, as before.
Seems like any kind of FS operations involving lots of tiny files is indeed just hopelessly slow on Windows, no matter what library or program does them...
Anyway, it was worth a try. I hope at least it did in fact take effect.

PR #15338, opened 2024-02-27
desktop: Initial Preferences work, allow changing the graphics backend and persist it to disk

### LINKED ISSUE #12740: [Feature request] Portable .sol file directory
This is a weird request. but would it be possible to have either an option or a command line argument to allow us to set the directory of the .sol files. The default (of course) being "%localappdata%\ruffle" but what if we as users wanted all of our saves in another folder?.

Alternatively, let's assume that someone has a flash drive full of... ha.. Flash games (unintentional pun) and they're sick of grabbing their .sol files from the default dir. So they can either set it as a portable directory via Ruffle's settings or have a simple "portable.ini" file next to ruffle.exe. Say you set the directory to "Saves". Ruffle will then create a "Saves" directory next to "game.swf" and save the data right there.

This would also remove the subdirectories. so let's imagine I do this on a flash drive (again. pun not intentional). instead of the path being "C:\Users\User\Appdata\Local\ruffle\SharedObjects\localhost\Downloads\flashgame" it would be in "G:\Flash\flashgame\Saves" (or even "G:\Flash\flashgame\Saves\game.swf")

Doing so will also remove any potential "Large file path" bugs that some users may run into. If this is either impossible or never planned to be added then that's fine. I get it's something that barely anyone will care about when Symlinks exist for that exact thing

ruffleexample
[EDIT: here's a screenshot for people who don't understand my awkward writing]

Dinnerbone:
This is mostly foundational work, with the initial feature set being just graphics backend & power preference. More will come later.

If the given backend is unavailable, it'll fall back to something else (but warn in the console, at least). This fixes the situation where someone can pick Metal but they're on Windows.

If the preferences are given by CLI, they will be "readonly" in the UI with a tooltip explanation.
Otherwise, settings are read from (and saved to) a file on disk.

There's a new "config directory" where this is stored; it's a directory and not just a single file because I'm envisioning there'll be something like a `recent.toml`, maybe a bunch of game specific files, etc. Rather than one mega toml.

This also fixes #12740 by providing a save directory CLI argument.


desktop/src/gui/preferences_dialog.rs
```diff
@@ -0,0 +1,169 @@
...
+                        &mut self.graphics_backend,
+                        GraphicsBackend::Vulkan,
+                        text(&self.locale, "graphics-backend-vulkan"),
+                    );
+                    ui.selectable_value(
+                        &mut self.graphics_backend,
+                        GraphicsBackend::Metal,
+                        text(&self.locale, "graphics-backend-metal"),
+                    );
+                    ui.selectable_value(
```
evilpie, line 102: I think we should try to only offer usable graphicbackends.

desktop/src/config.rs
```diff
@@ -0,0 +1,32 @@
+use crate::cli::Opt;
```
evilpie, line 1: Maybe it's just me, but having `GlobalPreferences` in `config.rs` feels weird, do we need to mixup these synonyms?

Dinnerbone: That's fair, I can rename it to "preferences"

I started calling it "config" initially but "preferences" felt more appropriate for global ruffle-specific stuff, and "config" maybe for the per-swf stuff later

desktop/src/gui/preferences_dialog.rs
Dinnerbone: Made that change!

desktop/src/preferences.rs
```diff
@@ -0,0 +1,80 @@
...
+            toml::to_string(preferences.deref()).context("Could not serialize preferences")?;
+        std::fs::write(self.cli.config.join("preferences.toml"), serialized)
+            .context("Could not write preferences to disk")
+    }
+}
+
+// [NA] Deliberately not "deny_unknown" here, trying to keep this backwards & forwards compatible.
+// It's quite common to bisect, even users manually hop back and forth to test things.
+// Therefore, try to take care to not error out if something is totally wrong.
+// TODO: Maybe a custom deserializer?
```
evilpie, line 71: Forwards-compatibility is always going to be difficult. If we added a version number now, we might have an easier time later when trying to do a break change.

Dinnerbone: Switched to using a new (de)serializer, so we shouldn't need to worry about versioning in theory. Anything unknown is just temporarily ignored, unless changed by the user.

Dinnerbone:
Added language selection too:

https://github.com/ruffle-rs/ruffle/assets/317625/b7e3f414-8c9b-4b1b-8c33-f89e43f82dfe

PR #15377, opened 2024-03-01
web: Don't panic on browsers without ReadableStreamDefaultController

danielhjacobs:
Palemoon "added an initial implementation of the ReadableStreams API, improving web compatibility with sites that apparently use this API in utilitarian fashion," according to http://www.palemoon.org/releasenotes.shtml#v32.5.0.

Unfortunately, that decreased Ruffle's compatibility with Pale Moon, as we rely on the existence of the [ReadableStreamDefaultController](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultController) when ReadableStream is defined. Therefore, change the compatibility check for ReadableStream to check for the existence of ReadableStreamDefaultController specifically.


torokati44:
Agh, `prettier` thinks this should be two lines now...

danielhjacobs:
I'll fix prettier. And no, it's not explicitly used, it is implicitly used though (it's the `controller` in `async start(controller)`)

PR #15380, opened 2024-03-01
web: Close sockets properly

kjarosh:
This patch ensures that sockets are closed properly. Without it, when the application-end of the socket was closed, the other end was left intact, and the socket was not closed.

### DIFF
--- a/web/src/navigator.rs
+++ b/web/src/navigator.rs
@@ -1,7 +1,8 @@
 //! Navigator backend for web
 use crate::SocketProxy;
 use async_channel::{Receiver, Sender};
-use futures_util::{SinkExt, StreamExt};
+use futures_util::future::Either;
+use futures_util::{future, SinkExt, StreamExt};
 use gloo_net::websocket::{futures::WebSocket, Message};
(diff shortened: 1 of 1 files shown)


web/src/navigator.rs
```diff
@@ -404,42 +405,55 @@ impl NavigatorBackend for WebNavigatorBackend {
...
-                    Err(_) => {
-                        stream_sender
-                            .try_send(SocketAction::Close(handle))
-                            .expect("working channel send");
-                        return Ok(());
+            loop {
+                match future::select(ws_read.next(), std::pin::pin!(receiver.recv())).await {
+                    // Handle incoming messages.
+                    Either::Left((msg, _)) => {
+                        let Some(msg) = msg else {
```
Dinnerbone, line 418: I think you can probably match on `Some(msg)` specifically (here and below for outgoing), and then`_` to `break`?

kjarosh: that's right, looks better now :)

Dinnerbone (approved):
Awesome :D Thank you!

PR #15383, opened 2024-03-01
chore: Replace `generational-arena` with `slotmap`

torokati44:
Because `generational-arena` is unmaintained, and `slotmap` is the recommended alternative.

Most of it was a trivial search-and-replace, except for:
- That `.rev()` from `LoadManager` is gone, because this iterator type does not support it.
  - Note that iteration order was undefined before, and is undefined now. Therefore, we shouldn't have been relying on this already.
  - Replaced by a clumsy hand-cranked reversing solution, but it should behave the same (undefined, but accidentally usually correct) way.
- `loader.rs` has sometimes used `Index` instead of the `Handle` type it defines, in its API.
  - We should probably also rename that to match the convention of the rest of the handle types.
- The dummy handle in the socket tests was replaced by the default, I hope that's fine, @kjarosh.

In the future, we could consider:
- Using more specialized implementations for faster iteration
  - https://docs.rs/slotmap/latest/slotmap/index.html#choosing-slotmap-hopslotmap-or-denseslotmap
- Defining different key types for the different handles, as is the proper intended usage
  - Instead of using `DefaultKey` everywhere
- Introducing a named key/handle type (or alias, for now) in `desktop/executor.rs` and `web/lib.rs` as well.

### DIFF
--- a/web/Cargo.toml
+++ b/web/Cargo.toml
@@ -32,7 +32,7 @@ profiling = []
 
 [dependencies]
 console_error_panic_hook = { version = "0.1.7", optional = true }
-generational-arena = "0.2.9"
+slotmap = { workspace = true }
 js-sys = "0.3.68"
 tracing = { workspace = true, features = ["log"] }
 tracing-subscriber = { version = "0.3.18", default-features = false, features = ["registry"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2273,15 +2273,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "generational-arena"
-version = "0.2.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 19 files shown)


torokati44:
I'll try to ping @relrelb due to the iteration order thingy...

web/Cargo.toml
```diff
@@ -32,7 +32,7 @@ profiling = []
 
 [dependencies]
 console_error_panic_hook = { version = "0.1.7", optional = true }
-generational-arena = "0.2.9"
+slotmap = "1.0.7"
```
Aaron1011, line 35: Can you add this to the workspace dependencies?

torokati44: Good idea!

torokati44: Done.

Dinnerbone (approved):

PR #5250, opened 2021-09-08
chore: Bump to wgpu 0.10

Herschel:
Bump to latest wgpu 0.10.1. The main breaking change was the merge of `wgpu::SwapChain` into `Surface`.

PR #15387, opened 2024-03-01
web: No longer set the `web_sys_unstable_apis` config variable

torokati44:
At least `wgpu` no longer requires this as of `0.19.3` (even for WebGPU).
Let's see whether everything else keeps working without it...
Referencing: https://github.com/ruffle-rs/ruffle/pull/5250


torokati44:
CI will probably pass after https://github.com/ruffle-rs/ruffle/pull/15385 due to https://github.com/rustwasm/wasm-bindgen/pull/3791.

PR #15401, opened 2024-03-03
Define distinct key types for different kinds of slotmaps

torokati44:
Followup for https://github.com/ruffle-rs/ruffle/pull/15383.

I didn't dare touch `web/lib.rs`, because "scary JS interop", but that is the only `SlotMap` left with `DefaultKey`, so there probably won't be any confusion with that.

An ACK for the audio behavior change in `streams.rs` from @kmeisthax would be appreciated. I think this has been a bug so far, resulting from the exact same type of confusion that this PR is aiming to prevent in the future.


torokati44:
Added some more cleanup commits.
Hopefully this won't actually revert the bug that https://github.com/ruffle-rs/ruffle/pull/1395 fixed with the workaround.

Aaron1011 (approved):

adrian17 (approved):

PR #15455, opened 2024-03-05
desktop: Persist volume to disk & don't translate GPU backends

Dinnerbone:
I kept the `--volume` CLI argument though I'm not entirely convinced of its usefulness now. Regardless, if you pass in `--volume` then we don't persist any changes and consider it a temporary thing.


danielhjacobs (commented):
Should this string be removed?

desktop/assets/texts/en-US/preferences_dialog.ftl
```diff
@@ -3,11 +3,6 @@ preferences-dialog = Ruffle Preferences
 preference-locked-by-cli = Read-Only (Set by CLI)
 
 graphics-backend = Graphics Backend
-graphics-backend-vulkan = Vulkan
-graphics-backend-gl = OpenGL
-graphics-backend-dx12 = DirectX 12
-graphics-backend-metal = Metal
-graphics-backend-default = Default
```
danielhjacobs, line 10: Should this string be removed?

Dinnerbone: Hmm good point!

Dinnerbone: Added it back!

PR #15470, opened 2024-03-07
avm2: Lookup multinames in verifier and store them in `Op`

Lord-McSweeney:
Most common ops now no longer call `pool_maybe_uninitialized_multiname` or similar functions. As a result, `pool_maybe_uninitialized_multiname` practically disappears from profiles for most SWFs. 

TODO:
- [x] Lookup classes for `AsType` and `IsType`
- [x] Move all string lookups to `verify.rs`
- [x] Lookup exception classes and multinames

### DIFF
--- a/core/src/avm2/activation.rs
+++ b/core/src/avm2/activation.rs
@@ -717,17 +717,6 @@ impl<'a, 'gc> Activation<'a, 'gc> {
             })
     }
 
-    /// Retrieve a string from the current constant pool.
-    fn pool_string<'b>(
-        &mut self,
-        method: &'b BytecodeMethod<'gc>,
-        index: Index<String>,
-    ) -> Result<AvmAtom<'gc>, Error<'gc>> {
-        method
-            .translation_unit()
-            .pool_string(index.0, &mut self.borrow_gc())
-    }
-
     /// Retrieve a namespace from the current constant pool.
     fn pool_namespace(
         &mut self,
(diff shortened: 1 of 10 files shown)


core/src/avm2/activation.rs
```diff
@@ -1398,26 +1400,44 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
+            let object_value = self.context.avm2.peek(1);
 
-                    return Ok(FrameControl::Continue);
+            // FIXME: avmplus does this differently when the Multiname does not
+            // include the public namespace
+            if let Value::Object(object_value) = object_value {
+                match name_value {
+                    Value::Integer(name_int) if name_int >= 0 => {
+                        if let Some(mut array_storage) =
+                            object_value.as_array_storage_mut(self.context.gc_context)
```
adrian17, line 1418: This shouldn't be Array-specific, see `get_index_property`.
Also, can this go into a separate PR? It's really unrelated here.

core/src/avm2/verify.rs
```diff
@@ -408,12 +410,80 @@ pub fn verify_method<'gc>(
...
+
+            if namespaces.is_empty() {
+                // NOTE: avmplus segfaults here
+                panic!("Should have at least one namespace for QName in exception variable name");
+            }
+
+            let name = pooled_variable_name.local_name().expect("Just checked");
+
+            // avmplus uses the first namespace, regardless of how many namespaces there are.
+            Some(QName::new(namespaces[0], name))
```
adrian17, line 456: If this is a more generic rule (applying to all traits' names), we might wanna have a more generic function that converts a generic Multiname to a specific trait name.

adrian17 (commented):
LGTM, aside from the last array change.
Also I remember Dinnerbone not being a fan of raw `unwrap()s`, and I think I saw some without comments. EDIT: looks like these follow the existing pattern, so less pressing I guess.

core/src/avm2/verify.rs
Lord-McSweeney: I have _no idea_ where else this applies, and this is short, so I think this is fine for now. If it comes up later we can always change it.

PR #15341, opened 2024-02-28
avm2: Match verifier reading implementation with FP's

Lord-McSweeney:
Fixes #15314
Fixes #15181

Fixes #15345

Fixes the regression in #10148

With this implementation, all code in `new_code` is guaranteed to be reachable (and therefore also run through the verifier).

PR #15476, opened 2024-03-07
avm2: unify abstract type propagation

adrian17:
This unifies `ValueType::Number` and `ValueType::Class(classes.number)` (etc) to a single representation (that's analogous to avmplus's abstract optimizer's abstract value), so that primitive-based optimizations can happen more consistently.

This also refactors `astype` optimization to be more conservative, which replaces the previous fix from https://github.com/ruffle-rs/ruffle/pull/15341/commits/c6e3eaad02341f262b3a6b23510946bf79111595 .

There's obviously lots of relatively low-hanging fruit: remaining numerical ops, function return types etc. A lot of current type propagation tries to be careful and conservative where it probably doesn't need to be. In particular, we should be testing for inheritance instead of type equality more.

(One thing I still haven't fully figured out is the exact guaranteed semantics of the `(u)int` type. It appears that despite not guaranteeing int storage, the runtime double value is still guaranteed(?) to pass `is int` test, so maybe possibly the `CoerceI` optimization can happen for all `int`s and not only constant ones...?)

On testing with box2d, it increases ratio of successful CoerceD optimizations from 25% to 80%. This reduces the total % of executed CoerceD ops from 2.9% to 0.7%, and consequently increases Nop to 4.4%.
Sounds like we're nearing the point where either a "remove nops" pass, or an emit-opcodes refactor (in case we ever want one opcode to optimize into multiple) will be beneficial.

### DIFF
--- a/core/src/avm2/optimize.rs
+++ b/core/src/avm2/optimize.rs
@@ -10,48 +10,80 @@ use gc_arena::{Gc, GcCell};
 use std::collections::HashSet;
 
 #[derive(Clone, Copy, Debug)]
-enum ValueType<'gc> {
-    // Either a class, or null.
-    Class(ClassObject<'gc>),
-    Int,
-    Uint,
-    Number,
-    Boolean,
-    Null,
-    Any,
+struct OptValue<'gc> {
+    // This corresponds to the compile-time assumptions about the type:
+    // - primitive types can't be undefined or null,
+    // - Object (and any other non-primitive type) is non-undefined, but can be null
+    // - None (the * type) can be any value,
+    // - a value typed as int can be stored as a Number (and vice versa),
+    //   BUT an int-typed value should always pass `is int`
+    //   (say, a Value::Number above hardcoded int-range that's still representable as i32).
+    // Note that `null is Object` is still `false`. So think of this type more in terms of
+    // "could this value be a possible value of `var t: T`"
+    pub class: Option<ClassObject<'gc>>,
+
+    // true if the value is guaranteed to be Value::Integer
+    // should only be set if class is numeric.
+    pub contains_valid_integer: bool,
+    // true if the value is guaranteed to be Value::Integer AND is >=0
+    // should only be set if class is numeric.
+    pub contains_valid_unsigned: bool,
+
+    // true if value is guaranteed to be null.
+    // TODO: FP actually has a separate `null` type just for this, this can be observed in VerifyErrors
+    // (a separate type would also prevent accidental "null int" values)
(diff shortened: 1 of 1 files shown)


core/src/avm2/optimize.rs
```diff
@@ -11,48 +11,80 @@ use std::collections::HashSet;
...
-enum ValueType<'gc> {
-    // Either a class, or null.
-    Class(ClassObject<'gc>),
-    Int,
-    Uint,
-    Number,
-    Boolean,
-    Null,
-    Any,
+struct Value<'gc> {
```
Lord-McSweeney, line 14: We should probably name this something else to avoid conflicting with AVM2 `Value`

core/src/avm2/optimize.rs
```diff
@@ -11,48 +11,80 @@ use std::collections::HashSet;
...
+    // should only be set if class is numeric.
+    pub contains_valid_integer: bool,
+    // true if the value is guaranteed to be Value::Integer AND is >=0
+    // should only be set if class is numeric.
+    pub contains_valid_unsigned: bool,
+
+    // true if value is guaranteed to be null.
+    // TODO: FP actually has a separate `null` type just for this, this can be observed in VerifyErrors
+    // (a separate type would also prevent accidental "null int" values)
+    pub is_null: bool,
```
Lord-McSweeney, line 36: Can we name this `guaranteed_null` rather than `is_null`?

core/src/avm2/optimize.rs
adrian17: I wanted `Value` exactly because it only makes sense in optimizer's context ;D
And I wanted to keep it short, to reduce boilerplate's volume in optimizer's code.
That said, I'm open for suggestions; `AbstractValue`? `OptValue`?

core/src/avm2/optimize.rs
```diff
@@ -238,44 +274,49 @@ pub fn optimize<'gc>(
...
                     *op = Op::Nop;
                 }
-                stack.push_number();
+                stack.push_class_object(types.number);
             }
             Op::CoerceI => {
-                let stack_value = stack.pop();
-                if matches!(stack_value, Some(ValueType::Int | ValueType::Uint)) {
+                let stack_value = stack.pop_or_any();
+                // TODO: maybe the type check is safe here...?
```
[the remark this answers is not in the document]
adrian17: I thought we confirmed it is safe :D 
Let's continue this on discord, but for now the comment doesn't hurt - can we leave it as is?

PR #15491, opened 2024-03-08
avm2: Add support for sparse arrays

jarca0123:
This attempts to implement sparse arrays. There are (as of creating this pull request) big caveats:
- ~~dense arrays are turned into sparse arrays only if they go beyond current array length and conditions for conversion into sparse arrays from avmplus aren't implemented~~ dense arrays do not implement m_denseStart from avmplus, this might need a custom Vec 
- it uses BTreeMap, when it probably would be more "correct" to use the DynamicMap implementation
- I have no idea if splice works

It does pass tests and also fixes some known failures.

### DIFF
--- a/core/src/display_object/graphic.rs
+++ b/core/src/display_object/graphic.rs
@@ -49,13 +49,11 @@ impl<'gc> Graphic<'gc> {
         let static_data = GraphicStatic {
             id: swf_shape.id,
             bounds: swf_shape.shape_bounds.clone(),
-            render_handle: Some(context.renderer.register_shape(
-                (&swf_shape).into(),
-                &MovieLibrarySource {
-                    library,
-                    gc_context: context.gc_context,
-                },
-            )),
+            render_handle: Some(
+                context
+                    .renderer
+                    .register_shape((&swf_shape).into(), &MovieLibrarySource { library }),
+            ),
             shape: swf_shape,
             movie,
         };
--- a/core/src/display_object/morph_shape.rs
+++ b/core/src/display_object/morph_shape.rs
@@ -238,13 +238,9 @@ impl MorphShapeStatic {
             handle
         } else {
             let library = library.library_for_movie(self.movie.clone()).unwrap();
-            let handle = context.renderer.register_shape(
-                (&frame.shape).into(),
-                &MovieLibrarySource {
-                    library,
-                    gc_context: context.gc_context,
-                },
-            );
+            let handle = context
+                .renderer
+                .register_shape((&frame.shape).into(), &MovieLibrarySource { library });
             frame.shape_handle = Some(handle.clone());
             handle
         }
--- a/core/src/library.rs
+++ b/core/src/library.rs
@@ -346,7 +346,6 @@ impl<'gc> MovieLibrary<'gc> {
 
 pub struct MovieLibrarySource<'a, 'gc> {
     pub library: &'a MovieLibrary<'gc>,
-    pub gc_context: &'a Mutation<'gc>,
 }
 
 impl<'a, 'gc> ruffle_render::bitmap::BitmapSource for MovieLibrarySource<'a, 'gc> {
--- a/core/src/streams.rs
+++ b/core/src/streams.rs
@@ -164,6 +164,7 @@ impl<'gc> Eq for NetStream<'gc> {}
 pub enum NetStreamType {
     /// The stream is an FLV.
     Flv {
+        #[allow(dead_code)] // set but never read
         header: FlvHeader,
 
         /// The currently playing video track's stream instance.
--- a/render/naga-agal/src/builder.rs
+++ b/render/naga-agal/src/builder.rs
@@ -247,6 +247,7 @@ impl VertexAttributeFormat {
 pub struct ShaderConfig<'a> {
     pub shader_type: ShaderType,
     pub vertex_attributes: &'a [Option<VertexAttributeFormat>; 8],
+    #[allow(dead_code)] // set but never read
     pub sampler_configs: &'a [SamplerConfig; 8],
     pub version: AgalVersion,
 }


core/src/avm2/array.rs
```diff
@@ -33,49 +106,190 @@ impl<'gc> ArrayStorage<'gc> {
             .map(|v| Some(*v))
             .collect::<Vec<Option<Value<'gc>>>>();
 
-        Self { storage }
+        let storage_type = ArrayStorage::Dense(storage, values.len());
```
evilpie, line 109: This `let` seems superfluous to me. Can't you just return `ArrayStore::Dense`?

core/src/avm2/array.rs
```diff
@@ -85,44 +299,132 @@ impl<'gc> ArrayStorage<'gc> {
...
     pub fn append(&mut self, other_array: &Self) {
-        for other_item in other_array.storage.iter() {
-            self.storage.push(*other_item)
+        match self {
+            ArrayStorage::Dense(storage, dense_used) => match &other_array {
+                ArrayStorage::Dense(other_storage, _other_dense_used) => {
+                    for other_item in other_storage.iter() {
+                        storage.push(*other_item)
+                    }
+                    *dense_used = storage.iter().filter(|v| v.is_some()).count();
```
evilpie, line 308: Can't you just add `dense_used` and `other_dense_used`?

core/src/avm2/array.rs
```diff
@@ -85,44 +299,132 @@ impl<'gc> ArrayStorage<'gc> {
...
+                    if *dense_used < (storage.len() / 4) && MIN_SPARSE_LENGTH < storage.len() {
+                        self.convert_to_sparse();
+                    }
+                }
+            },
+            ArrayStorage::Sparse(storage, length) => match &other_array {
+                ArrayStorage::Dense(other_storage, ..) => {
+                    for (i, v) in other_storage.iter().enumerate() {
+                        match v {
```
evilpie, line 326: `if let Some()`

core/src/avm2/array.rs
```diff
@@ -85,44 +299,132 @@ impl<'gc> ArrayStorage<'gc> {
...
+        }
     }
 
     /// Push an array hole onto the end of this array.
     pub fn push_hole(&mut self) {
-        self.storage.push(None)
+        match self {
+            ArrayStorage::Dense(storage, dense_used) => {
+                storage.push(None);
+                if *dense_used < (storage.len() / 4) && MIN_SPARSE_LENGTH < storage.len() {
```
evilpie, line 369: I would add something like `maybe_convert_to_sparse` for this whole block.

core/src/avm2/array.rs
```diff
@@ -85,44 +299,132 @@ impl<'gc> ArrayStorage<'gc> {
...
-        let mut non_hole = None;
+        match self {
+            ArrayStorage::Dense(storage, dense_used) => {
+                let mut non_hole = None;
 
-        for (i, item) in self.storage.iter().enumerate().rev() {
-            if item.is_some() {
-                non_hole = Some(i);
-                break;
+                for (i, item) in storage.iter().enumerate().rev() {
```
evilpie, line 388: I think this is just `rposition`.

core/src/avm2/array.rs
jarca0123: I didn't realize that, thanks.

core/src/avm2/array.rs
```diff
@@ -2,17 +2,308 @@
...
+                    let value = storage[self.index];
+                    self.index += 1;
+                    Some(value)
+                }
+            }
+            ArrayStorage::Sparse(storage, _length) => {
+                if self.index >= self.index_back {
+                    None
+                } else {
+                    let value = storage.get(&self.index).cloned();
```
Lord-McSweeney, line 52: Is the `.cloned` necessary here (and in some other cases)? `Value` implements `Copy`.

core/src/avm2/array.rs
```diff
@@ -2,17 +2,308 @@
...
+                    Some(value)
+                }
+            }
+        }
+    }
+}
+
+impl<'a, 'gc> ExactSizeIterator for ArrayStorageMutableIterator<'a, 'gc> {
+    fn len(&self) -> usize {
+        match &self.storage {
```
Lord-McSweeney, line 149: Can this just be
```
    fn len(&self) -> usize {
        self.index_back - self.index
    }
```
?

core/src/avm2/object/array_object.rs
```diff
@@ -217,13 +208,12 @@ impl<'gc> TObject<'gc> for ArrayObject<'gc> {
...
-        while last_index < array_length {
-            if read.array.get(last_index as usize).is_some() {
-                return Ok(Some(last_index + 1));
-            }
-            last_index += 1;
+        if let Some(index) = read.array.get_next_enumerant(last_index as usize) {
+            return Ok(Some(index as u32));
         }
 
+        last_index = max(last_index, array_length);
```
Lord-McSweeney, line 215: Can this just use `std::cmp::max` here instead of doing a
```
use std::cmp::max;
```
at the top of the code?

core/src/avm2/array.rs
```diff
@@ -2,17 +2,308 @@
...
+            ArrayStorage::Dense(storage, ..) => {
+                if self.index >= self.index_back || self.index_back == 0 {
+                    None
+                } else {
+                    self.index_back -= 1;
+                    let value = storage[self.index_back];
+                    Some(value)
+                }
+            }
+            ArrayStorage::Sparse(storage, _length) => {
```
Lord-McSweeney, line 73: For consistency can this be `ArrayStorage::Sparse(storage, ..)`?

core/src/avm2/array.rs
```diff
@@ -2,17 +2,308 @@
...
+            ArrayStorage::Dense(storage, ..) => {
+                if self.index >= self.index_back {
+                    None
+                } else {
+                    let value = storage[self.index];
+                    self.index += 1;
+                    Some(value)
+                }
+            }
+            ArrayStorage::Sparse(storage, _length) => {
```
Lord-McSweeney, line 109: For consistency can this be `ArrayStorage::Sparse(storage, ..)`?

core/src/avm2/array.rs
```diff
@@ -2,17 +2,308 @@
...
+            ArrayStorage::Dense(storage, ..) => {
+                if self.index >= self.index_back || self.index_back == 0 {
+                    None
+                } else {
+                    self.index_back -= 1;
+                    let value = storage[self.index_back];
+                    Some(value)
+                }
+            }
+            ArrayStorage::Sparse(storage, _length) => {
```
Lord-McSweeney, line 134: Same here

core/src/avm2/array.rs
```diff
@@ -33,49 +326,185 @@ impl<'gc> ArrayStorage<'gc> {
...
+                let mut last_index = last_index;
+                while last_index < storage.len() {
+                    if storage[last_index].is_some() {
+                        return Some(last_index + 1);
+                    }
+                    last_index += 1;
+                }
+                None
+            }
+            ArrayStorage::Sparse(storage, _length) => {
```
Lord-McSweeney, line 364: `ArrayStorage::Sparse(storage, ..)`

core/src/avm2/array.rs
```diff
@@ -33,49 +326,185 @@ impl<'gc> ArrayStorage<'gc> {
...
+            }
         }
     }
 
     /// Get the length of the array.
     pub fn length(&self) -> usize {
-        self.storage.len()
+        match &self {
+            ArrayStorage::Dense(storage, ..) => storage.len(),
+            ArrayStorage::Sparse(_storage, length) => *length,
```
Lord-McSweeney, line 465: `ArrayStorage::Sparse(_, length)`?

core/src/avm2/array.rs
```diff
@@ -85,44 +514,130 @@ impl<'gc> ArrayStorage<'gc> {
...
+                if let Some((i, _item)) = storage_clone.iter().next_back() {
+                    non_hole = Some(i);
+                }
+                if let Some(non_hole) = non_hole {
+                    *length -= 1;
+                    storage.remove(non_hole).unwrap()
+                } else {
+                    let value = storage
+                        .get(&(*length - 1))
+                        .cloned()
```
Lord-McSweeney, line 634: Can this be `.copied()`?

Lord-McSweeney:
The uses of `ArrayStorage::XXX(something, ..)` and `ArrayStorage::XXX(something, _something_else)` can all be replaced with `ArrayStorage::XXX(something, _)`

core/src/avm2/array.rs
```diff
@@ -11,8 +14,229 @@ use std::{cmp::max, ops::RangeBounds};
...
+    index_back: usize,
+}
+
+impl<'a, 'gc> Iterator for ArrayStorageIterator<'a, 'gc> {
+    type Item = Option<Value<'gc>>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match &self.storage {
+            ArrayStorage::Dense(storage, _) => {
+                if self.index >= self.index_back {
```
Lord-McSweeney, line 40: Since this check is common to both branches, can it be moved outside the `match`?

core/src/avm2/array.rs
```diff
@@ -11,8 +14,229 @@ use std::{cmp::max, ops::RangeBounds};
...
+            }
+        }
+    }
+}
+
+impl<'a, 'gc> DoubleEndedIterator for ArrayStorageIterator<'a, 'gc> {
+    fn next_back(&mut self) -> Option<Self::Item> {
+        match &self.storage {
+            ArrayStorage::Dense(storage, _) => {
+                if self.index >= self.index_back || self.index_back == 0 {
```
Lord-McSweeney, line 65: Same here

core/src/avm2/array.rs
```diff
@@ -11,8 +14,229 @@ use std::{cmp::max, ops::RangeBounds};
...
+    }
+}
+
+impl<'a, 'gc> Iterator for ArrayStorageMutableIterator<'a, 'gc> {
+    type Item = Option<Value<'gc>>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match &self.storage {
+            ArrayStorage::Dense(storage, _) => {
+                if self.index >= self.index_back {
```
Lord-McSweeney, line 98: Same here

core/src/avm2/array.rs
```diff
@@ -11,8 +14,229 @@ use std::{cmp::max, ops::RangeBounds};
...
+            }
+        }
+    }
+}
+
+impl<'a, 'gc> DoubleEndedIterator for ArrayStorageMutableIterator<'a, 'gc> {
+    fn next_back(&mut self) -> Option<Self::Item> {
+        match &self.storage {
+            ArrayStorage::Dense(storage, _) => {
+                if self.index >= self.index_back || self.index_back == 0 {
```
Lord-McSweeney, line 123: Same here

core/src/avm2/array.rs
```diff
@@ -21,8 +245,10 @@ impl<'gc> ArrayStorage<'gc> {
     /// The length parameter indicates how big the array storage should start
     /// out as. All array storage consists of holes.
     pub fn new(length: usize) -> Self {
-        Self {
-            storage: vec![None; length],
+        if length > (1 << 28) {
```
Lord-McSweeney, line 248: Can this `(1 << 28)` be moved to the top of the file, like `const MAX_DENSE_LENGTH`?

Lord-McSweeney (approved):

Lord-McSweeney:
Not blocking, but I just realized that several places do `name.parse<usize>()` or `name.parse<u32>()`, but should instead do something like `name.parse::<u32>().ok().filter(|s| *s != 4294967295)`

core/src/avm2/array.rs
```diff
@@ -33,49 +224,189 @@ impl<'gc> ArrayStorage<'gc> {
...
-        if let Some(i) = self.storage.get_mut(item) {
-            *i = None;
+        match self {
+            ArrayStorage::Dense(storage, dense_used) => {
+                if let Some(i) = storage.get_mut(item) {
+                    *i = None;
+                    if *dense_used != 0 {
+                        *dense_used -= 1;
+                        if *dense_used == 0 {
+                            self.convert_to_dense();
```
Lord-McSweeney, line 348: How is this supposed to work? The array storage is already `Dense` here.

jarca0123: Oops, I don't know how I overlooked that, it's not supposed to be there. Now I've added conversion to dense if the sparse array is empty.

core/src/avm2/array.rs
```diff
@@ -21,8 +214,10 @@ impl<'gc> ArrayStorage<'gc> {
     /// The length parameter indicates how big the array storage should start
     /// out as. All array storage consists of holes.
     pub fn new(length: usize) -> Self {
-        Self {
-            storage: vec![None; length],
+        if length > MAX_DENSE_LENGTH {
+            ArrayStorage::Sparse(BTreeMap::new(), 0)
```
kmeisthax, line 218: Why does the inner ArrayStorage take a length of 0 if we have a length parameter?

core/src/avm2/array.rs
```diff
@@ -11,8 +15,197 @@ use std::{cmp::max, ops::RangeBounds};
 /// the prototype.
 #[derive(Clone, Collect)]
 #[collect(no_drop)]
-pub struct ArrayStorage<'gc> {
-    storage: Vec<Option<Value<'gc>>>,
+pub enum ArrayStorage<'gc> {
+    Dense(Vec<Option<Value<'gc>>>, usize),
```
kmeisthax, line 19: We need documentation here to explain why `Dense` now has a separate length parameter. Looking through the rest of the code, it *appears* that this length parameter counts the number of non-holes (akin to `BTreeMap.len()`.)

core/src/avm2/array.rs
```diff
@@ -11,8 +15,197 @@ use std::{cmp::max, ops::RangeBounds};
...
+    Sparse(BTreeMap<usize, Value<'gc>>, usize),
```
kmeisthax, line 20: Likewise the `usize` for `Sparse` should also be documented - it serves as an explicit ECMAScript Array.length, because we can't recover that information from what's in the `BTreeMap`. This notably is *different* from the `Dense` length parameter's purpose.

kmeisthax (commented):
After reading through most of the array code my main concern is just documenting what the `usize` parameters do, since that was the biggest thing I was confused about. That's the only thing I feel needs to be fixed before I can merge this.

The splice code seemed straightforward once I puzzled out what it was doing, and it's probably correct. I'm a bit iffy about "drop to splice" as an API but it shouldn't impact correctness in this case. (Remember that memory leaks are safe and that `Drop` doesn't always run!)

core/src/avm2/array.rs
evilpie: I think I agree. Probably better to change this to a struct like `Dense { storage: ..., dense_used: size }`. Actually the name `dense_used` feels a bit weird as well.

core/src/avm2/array.rs
```diff
@@ -11,8 +15,197 @@ use std::{cmp::max, ops::RangeBounds};
...
+}
+
+struct ArrayStorageIterator<'a, 'gc> {
+    storage: &'a ArrayStorage<'gc>,
+    index: usize,
+    index_back: usize,
+}
+
+pub struct ArrayStorageMutableIterator<'a, 'gc> {
```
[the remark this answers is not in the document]
jarca0123: It exists because I don't know how to implement it without a mutable iterator 😅

core/src/avm2/array.rs
adrian17: +1 on making it structs (assuming we wanna stay with a BTreeMap)
Also, I'm assuming this means that Dense can't have holes on either end (why not?)? If so, can this be documented too?

Lord-McSweeney (approved):

core/src/avm2/array.rs
```diff
@@ -33,49 +97,191 @@ impl<'gc> ArrayStorage<'gc> {
...
+            ArrayStorage::Dense {
+                storage,
+                occupied_count,
+            } => {
+                if size < MAX_DENSE_LENGTH {
+                    let mut num_of_new_holes = 0;
+                    if size > storage.len() {
+                        num_of_new_holes = size - storage.len();
+                    }
+                    if *occupied_count + num_of_new_holes < (size / 4)
```
adrian17, line 251: is this math correct? Not sure why it's adding holes to occupied.
Also, how about extracting the repeated logic to a function? As in, the
```
like, new_size >= MIN_SPARSE_LENGTH && new_used < new_size / 4
```

core/src/avm2/array.rs
```diff
@@ -85,44 +291,171 @@ impl<'gc> ArrayStorage<'gc> {
...
     /// Holes are copied as holes and not resolved at append time.
     pub fn append(&mut self, other_array: &Self) {
-        for other_item in other_array.storage.iter() {
-            self.storage.push(*other_item)
+        match self {
+            ArrayStorage::Dense {
+                storage,
+                occupied_count,
+            } => match &other_array {
+                ArrayStorage::Dense {
```
adrian17, line 299: could we be generic here? As in, just "for each item of `other_array`, add the item to our storage" - using either indices and `get`, or the iterator?

core/src/avm2/array.rs
```diff
@@ -85,44 +291,171 @@ impl<'gc> ArrayStorage<'gc> {
...
-        if let Some(non_hole) = non_hole {
-            self.storage.remove(non_hole).unwrap()
-        } else {
-            self.storage
-                .pop()
-                .unwrap_or(None)
-                .unwrap_or(Value::Undefined)
+                let mut non_hole = None;
+
+                let storage_clone = storage.clone();
```
adrian17, line 439: why does this require cloning the entire `storage`?

core/src/avm2/array.rs
```diff
@@ -85,44 +291,171 @@ impl<'gc> ArrayStorage<'gc> {
...
+                if let Some((i, _)) = storage_clone.iter().next_back() {
```
adrian17, line 440: Maybe this could use `last_entry` instead of an iterator?
Or maybe even try directly removing the last item with `pop_last`, so we don't need unwraps later?

core/src/avm2/array.rs
```diff
@@ -85,44 +291,171 @@ impl<'gc> ArrayStorage<'gc> {
...
+                if let Some(non_hole) = non_hole {
+                    *length -= 1;
+                    let value = storage.remove(non_hole).unwrap();
+                    self.maybe_convert_to_dense();
+                    value
+                } else {
+                    let value = storage
+                        .get(&(*length - 1))
+                        .copied()
+                        .unwrap_or(Value::Undefined);
```
adrian17, line 452: do I understand correctly that this only runs if the array is made entirely of holes? In that case, I think both this and the code for Dense (as in, the pre-existing one) sound overly complex, as `storage.get` must always be `None`; unless I'm missing some logic here.

core/src/avm2/array.rs
```diff
@@ -33,49 +97,175 @@ impl<'gc> ArrayStorage<'gc> {
...
+                        storage.resize(item + 1, None);
+                        *storage.get_mut(item).unwrap() = Some(value);
+                        let new_holes = (item + 1) - storage.len();
+                        *occupied_count = (*occupied_count + new_holes) + 1;
+                    }
+                } else {
+                    if storage[item].is_none() {
+                        *occupied_count += 1;
+                    }
+                    *storage.get_mut(item).unwrap() = Some(value);
```
adrian17, line 195: can this be just `storage[item] = Some(value)`?

core/src/avm2/array.rs
```diff
@@ -33,49 +97,175 @@ impl<'gc> ArrayStorage<'gc> {
...
+                    storage.resize(size, None);
+                }
+            }
+            ArrayStorage::Sparse { storage, length } => {
+                if size > *length {
+                    *length = size;
+                } else {
+                    let mut to_remove = Vec::new();
+                    for (i, _) in storage.range(size..) {
+                        to_remove.push(*i);
```
adrian17, line 259: I think this push-remove logic can be replaced by `storage.retain(|&k, _| k < size)`?

core/src/avm2/array.rs
```diff
@@ -198,8 +522,13 @@ where
     where
         I: IntoIterator<Item = V>,
     {
-        let storage = values.into_iter().map(|v| Some(v.into())).collect();
+        let storage: Vec<Option<Value>> = values.into_iter().map(|v| Some(v.into())).collect();
```
Lord-McSweeney, line 525: I think `Vec<_>` would be preferred here to maintain consistency

core/src/avm2/array.rs
```diff
@@ -33,49 +98,151 @@ impl<'gc> ArrayStorage<'gc> {
...
     /// array will be extended with holes.
     pub fn set(&mut self, item: usize, value: Value<'gc>) {
-        if self.storage.len() < (item + 1) {
-            self.storage.resize(item + 1, None)
+        match self {
+            ArrayStorage::Dense {
+                storage,
+                occupied_count,
+            } => {
+                if storage.len() < (item + 1) {
```
adrian17, line 162: I think this can be `item >= storage.len()`

core/src/avm2/array.rs
```diff
@@ -85,44 +252,138 @@ impl<'gc> ArrayStorage<'gc> {
...
+    }
+
+    /// Determine if the array should be converted to a sparse representation based on its size and the number of occupied slots.
+    fn should_convert_to_sparse(size: usize, occupied_count: usize) -> bool {
+        (occupied_count < (size / 4) && size > MIN_SPARSE_LENGTH) || size > MAX_DENSE_LENGTH
+    }
+
+    /// Convert the array storage to a sparse representation.
+    fn convert_to_sparse(&mut self) {
+        match self {
```
adrian17, line 290: Can these be rewritten without the match? Would be a bit shorter
```rs
if let ArrayStorage::Dense { storage, .. } = self {
```

core/src/avm2/array.rs
```diff
@@ -85,44 +252,138 @@ impl<'gc> ArrayStorage<'gc> {
...
+                    length: storage.len(),
+                };
+            }
+            ArrayStorage::Sparse { .. } => {}
+        }
+    }
+
+    /// Convert the array to a sparse representation if it meets the criteria.
+    fn maybe_convert_to_sparse(&mut self) {
+        match self {
```
adrian17, line 309: same as above

core/src/avm2/array.rs
```diff
@@ -85,44 +252,138 @@ impl<'gc> ArrayStorage<'gc> {
...
+                }
+            }
+            ArrayStorage::Sparse { .. } => {}
+        }
+    }
+
+    /// Convert the array to a dense representation if it meets the criteria.
+    fn maybe_convert_to_dense(&mut self) {
+        match self {
+            ArrayStorage::Dense { .. } => {}
```
adrian17, line 325: same

adrian17 (approved):

Issue #15487, opened 2024-03-08
Crash on Ena.swf


FlashBacks1998:
### Describe what you were doing

Working on a demo for the Ruffle Game jam. Failed to load a obj file with the Away3D Framework
[Ena.zip](https://github.com/ruffle-rs/ruffle/files/14538464/Ena.zip)

### What does the crash message say?

panicked at C:\Users\runneradmin\.cargo\registry\src\index.crates.io-6f17d22bba15001f\wgpu-0.19.3\src\backend\wgpu_core.rs:3006:5:
wgpu error: Validation Error

Caused by:
    In a RenderPass
      note: encoder = `<CommandBuffer-(3, 1, Vulkan)>`
    In a pass parameter
      note: command buffer = `<CommandBuffer-(3, 1, Vulkan)>`
    Multiview layer count must match

### Ruffle Version

0.1.0-nightly (571f60abff3e2e5d526cb62c74daafa91e3ebee4 2024-03-05)

### Affected platform

Desktop app

### Operating system

Windows 10.0.19045 (Windows 10 Home) [64-bit]

### Additional information

Filename: Ena.swf

### AVM2 Callstack
```
	at flash.display3D::Context3D/flash::display3D::Context3D::drawTriangles()
	at away3d.core.managers::Stage3DProxy/drawTriangles()
	at away3d.materials.passes::DistanceMapPass/render()
	at away3d.materials::MaterialBase/renderDepth()
	at away3d.core.render::DepthRenderer/drawRenderables()
	at away3d.core.render::DepthRenderer/draw()
	at away3d.core.render::RendererBase/executeRender()
	at away3d.core.render::RendererBase/render()
	at away3d.lights.shadowmaps::CubeMapShadowMapper/drawDepthMap()
	at away3d.lights.shadowmaps::ShadowMapperBase/renderDepthMap()
	at away3d.core.render::DefaultRenderer/updateLights()
	at away3d.core.render::DefaultRenderer/executeRender()
	at away3d.core.render::RendererBase/render()
	at away3d.containers::View3D/render()
	at scenes::Maze/onEnterFrame()
```

### Render Info
Renderer: wgpu
Adapter Backend: Vulkan
Adapter Name: "NVIDIA GeForce RTX 2080 Super with Max-Q Design"
Adapter Device Type: DiscreteGpu
Adapter Driver Name: "NVIDIA"
Adapter Driver Info: "551.52"
[...]

PR #15495, opened 2024-03-09
Implement surfaceSelector in Context3D.setRenderToTexture

Aaron1011:
This still needs a test, but I'm planning to add that in a follow-up PR
Fixes https://github.com/ruffle-rs/ruffle/issues/15487

### DIFF
--- a/core/src/avm2/globals/flash/display3D/context_3d.rs
+++ b/core/src/avm2/globals/flash/display3D/context_3d.rs
@@ -608,15 +608,6 @@ pub fn set_render_to_texture<'gc>(
         );
     }
 
-    if surface_selector != 0 {
-        avm2_stub_method!(
-            activation,
-            "flash.display3D.Context3D",
-            "setRenderToTexture",
-            "surfaceSelector != 0"
-        );
-    }
-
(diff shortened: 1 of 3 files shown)


core/src/avm2/globals/flash/display3D/context_3d.rs
```diff
@@ -612,15 +612,6 @@ pub fn set_render_to_texture<'gc>(
         );
     }
 
-    if surface_selector != 0 {
```
adrian17, line 615: Do we want the same validation as FP has?
> ArgumentError - for a mismatched surfaceSelector parameter. The value must be 0 for 2D textures and 0..5 for cube maps

adrian17 (approved):

Lord-McSweeney (approved):

moulins (approved):

core/src/avm2/globals/flash/display3D/context_3d.rs
torokati44: I hesitated to merge myself because I wasn't sure how strongly you wanted this to be addressed.

PR #15498, opened 2024-03-09
Request highest 'max_inter_stage_shader_components' supported by adapter

Aaron1011:
We're currently always requesting a maximum of 31 components, which is too low for some SWFs.


adrian17 (approved):
Out of curiosity, is there ever any downside in requesting the highest possible limit?

Aaron1011:
I'm not sure - it would be good to investigate

Dinnerbone:
> Out of curiosity, is there ever any downside in requesting the highest possible limit?

For this specific limit, no. We just need to make sure that we handle the case where the limit is lower than what we need - which we don't right now.

Issue #15454, opened 2024-03-05
non-ASCII chars in file name not properly displayed


ousia:
### Describe the bug

1. Rename [öæáßñ.txt](https://github.com/ruffle-rs/ruffle/files/14499392/oaeassn.txt) to `öæáßñ.swf`.
2. Open it with the desktop program.
3. At least in Linux (nightly 2024-03-05), file name is displayed as:

    ![wrong-non-ascii](https://github.com/ruffle-rs/ruffle/assets/203767/7223515e-2e0c-4855-a339-26a0884ae0bf)

### Expected behavior

The window title should read _Ruffle - öæáßñ.swf_.

### Content Location

https://github.com/ruffle-rs/ruffle/files/14499392/oaeassn.txt (as explained, replace the extension with `.swf`).

### Affected platform

Desktop app

### Operating system

Fedora 37

torokati44:
Huh, I guess we're supposed to `urldecode` it before setting it as a window title. :) Good catch!

TheDecimeter:
I was looking at this same issue, did you happen to come across: https://docs.rs/urlencoding/latest/urlencoding/

(you are correct about the location of the bug)

PR #15509, opened 2024-03-10
desktop: Add URL decode to desktop window's title (close: #15454)

TheDecimeter:
Fix so that desktop title properly renders non ASCII characters (the web player uses a static title so no fix is needed there)

issue: https://github.com/ruffle-rs/ruffle/issues/15454

image

### DIFF
--- a/desktop/src/player.rs
+++ b/desktop/src/player.rs
@@ -19,6 +19,7 @@ use ruffle_render::backend::RenderBackend;
 use ruffle_render::quality::StageQuality;
 use ruffle_render_wgpu::backend::WgpuRenderBackend;
 use ruffle_render_wgpu::descriptors::Descriptors;
+use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::path::PathBuf;
 use std::rc::Rc;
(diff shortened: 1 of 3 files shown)


desktop/src/player.rs
```diff
@@ -184,7 +185,9 @@ impl ActivePlayer {
             .unwrap_or_else(|| movie_url.as_str())
             .to_string();
 
-        window.set_title(&format!("Ruffle - {name}"));
+        let readable_name = decode(&name).expect("UTF-8");
```
Lord-McSweeney, line 188: Hmm, this probably shouldn't panic (a SWF file with an invalid name should still play). Can this be something like `unwrap_or(name)`?

Dinnerbone (approved):

Dinnerbone:
Thank you!

torokati44:
Thank you indeed, and welcome! :wave:

TheDecimeter:
> @TheDecimeter,
> 
> many thanks for your implementation.
> 
> Now animation filenames such as this are possible 😁:
> 
> ruffle-non-ascii-fs8-fs8

My pleasure, it was a fun intro to Rust

Issue #15213, opened 2024-02-14
Ruffle Crashes trying to run a game on Wayland


Doomsdayrs:
## Describe what you were doing

1. Select Balloons Tower Defense
![image](https://github.com/ruffle-rs/ruffle/assets/38189170/fbab7aba-e612-4af6-bbee-aaf076b89784)
2. Click Start
3. Notice Error
![image](https://github.com/ruffle-rs/ruffle/assets/38189170/b11796f5-a2c7-4d43-a8be-bf6768a2c40c)

## What does the crash message say?

```
panicked at desktop/src/player.rs:158:22:
Couldn't create ui backend: Couldn't get platform clipboard

Caused by:
    Unknown error while interacting with the clipboard: Display parsing error

Stack backtrace:
   0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
stack backtrace:
   0:     0x55c079bca4b6 - <unknown>
   1:     0x55c079bfb5f0 - <unknown>
   2:     0x55c079bc648f - <unknown>
   3:     0x55c079bca294 - <unknown>
   4:     0x55c079bcbd77 - <unknown>
   5:     0x55c079bcbad9 - <unknown>
   6:     0x55c0789df1b1 - <unknown>
   7:     0x55c079bcc396 - <unknown>
   8:     0x55c079bcc0e2 - <unknown>
   9:     0x55c079bca9b6 - <unknown>
  10:     0x55c079bcbe40 - <unknown>
  11:     0x55c07897b7f5 - <unknown>
  12:     0x55c07897bd43 - <unknown>
  13:     0x55c0789bf7c3 - <unknown>
  14:     0x55c078a8f0c3 - <unknown>
  15:     0x55c078a7ebbd - <unknown>
  16:     0x55c0789f7323 - <unknown>
  17:     0x55c0789f9bbb - <unknown>
  18:     0x55c078a8b826 - <unknown>
```

### Ruffle Version

0.1.0-nightly (555fcfa9c42ecf180a838bf2c9aca0d97ec45496 2024-02-12)

### Affected platform

Desktop app

### Operating system

Fedora 39

### Display Manager
Wayland

### Render Info
Renderer: wgpu
Adapter Backend: Vulkan
Adapter Name: "AMD Radeon Graphics (RADV NAVI22)"
Adapter Device Type: DiscreteGpu
Adapter Driver Name: "radv"
[...]

PR #15535, opened 2024-03-11
desktop: Make clipboard support optional

ColinKinloch:
On Wayland ruffle uses https://github.com/YaLTeR/wl-clipboard-rs which only supports [`wlr-data-control-unstable-v1`](https://wayland.app/protocols/wlr-data-control-unstable-v1) which is not intended to be supported by all compositors.

There are options with better support https://wayland.app/protocols/primary-selection-unstable-v1 but until it's implemented in `wl-clipboard-rs` ruffle will crash for want of clipboard access.

For most platforms this MR wont effect anything. On platforms without clipboard support it'll quietly fail to set and get the clipboard.

Fixes https://github.com/ruffle-rs/ruffle/issues/15213

Mutter doesn't intend to support it: https://gitlab.gnome.org/GNOME/mutter/-/issues/524

### DIFF
--- a/desktop/src/backends/ui.rs
+++ b/desktop/src/backends/ui.rs
@@ -1,5 +1,5 @@
 use crate::preferences::GlobalPreferences;
-use anyhow::{Context, Error};
+use anyhow::Error;
 use arboard::Clipboard;
 use chrono::{DateTime, Utc};
 use fontdb::Family;
@@ -117,7 +117,7 @@ impl FileDialogResult for DesktopFileDialogResult {
 pub struct DesktopUiBackend {
     window: Rc<Window>,
     cursor_visible: bool,
-    clipboard: Clipboard,
+    clipboard: Option<Clipboard>,
     preferences: GlobalPreferences,
     preferred_cursor: MouseCursor,
     open_url_mode: OpenURLMode,
@@ -136,7 +136,9 @@ impl DesktopUiBackend {
         Ok(Self {
             window,
             cursor_visible: true,
-            clipboard: Clipboard::new().context("Couldn't get platform clipboard")?,
+            clipboard: Clipboard::new()
+                .inspect_err(|err| tracing::error!("Failed to initialize clipboard: {err}"))
+                .ok(),
             preferences,
             preferred_cursor: MouseCursor::Arrow,
             open_url_mode,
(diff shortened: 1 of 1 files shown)


desktop/src/backends/ui.rs
```diff
@@ -136,7 +136,9 @@ impl DesktopUiBackend {
         Ok(Self {
             window,
             cursor_visible: true,
-            clipboard: Clipboard::new().context("Couldn't get platform clipboard")?,
+            clipboard: Clipboard::new()
+                .context("Couldn't get platform clipboard")
```
torokati44, line 140: Is this going to be logged anywhere, if followed by an `.ok()` immediately? :thinking:

ColinKinloch: I wrote this code late and I've not done much rust recently so I'm a little out of practice. Sorry.

I guess the "anyhow"s `context` call doesn't do anything in this context.

"arboard" spits out a pretty detailed error if that's sufficient, or I could check if `Clipboard::new().ok().is_none()` and warn.
```
2024-03-12T17:12:26.436744Z  WARN arboard::platform::linux: Tried to initialize the wayland data control protocol clipboard, but failed. Falling back to the X11 clipboard protocol. The error was: Unknown error while interacting with the clipboard: A required Wayland protocol (zwlr_data_control_manager_v1 version 1) is not supported by the compositor
```

evilpie: You could do `Clipboard::new().inspect_err(|err| tracing::error("Failed to initialize clipboard: {err}")).ok()`

ColinKinloch: Neat, thanks for the tip :)

evilpie (approved):

torokati44 (approved):
Thank you!
We really should get to the bottom of this Wayland protocol mess someday...

evilpie:
> Thank you! We really should get to the bottom of this Wayland protocol mess someday...

Second! I really hope `wl-clipboard-rs` or `arboard` can figure this out for us :smiley:

torokati44:
Well, of course... 😅
Or `smithay-clipboard`, that we're currently directly using.

torokati44:
Oh, and welcome, new contributor! :wave: :smile:

torokati44:
Speaking of - could you please raise this to https://github.com/Smithay/smithay-clipboard/issues as well, OP?
Seems like you have a better understanding of all these protocols and their implementation statuses than us...

ColinKinloch:
> could you please raise this to smithay-clipboard

Smithays fine, I'm currently looking into whether I can use it in place of arboard.

On that, do you have any example swf files that access the clipboard?

torokati44:
I meant the `wlr-data-control-unstable-v1` vs. `primary-selection-unstable-v1` thing.
Sounded to me like these are alternatives of each other...?

torokati44:
> On that, do you have any example swf files that access the clipboard?

Yeah, see: https://github.com/ruffle-rs/ruffle/issues/6476#issuecomment-1071468712
Proper link in this comment, steps to reach clipboard usage in the original description.

ColinKinloch:
In wayland the clipboard is defined in the core protocol by [`wl_data_device::selection`](https://wayland.app/protocols/wayland#wl_data_device:event:selection) and [`wl_data_device::set_selection`](https://wayland.app/protocols/wayland#wl_data_device:request:set_selection).

[`wlr-data-control-unstable-v1`](https://wayland.app/protocols/wlr-data-control-unstable-v1) seems to be for creating a "clipboard manager" which seems to be a client that is allowed to manage the clipboard.

"primary selection" refers to the ability to use middle click to paste the last highlighted text.

`egui_winit::clipboard::Clipboard` seems like an easy option as it's already in the dependency tree and wraps both arboard and smithay-clipboard. However in testing it in a flatpak sandbox it still seems to require x11 (XWayland) to access the system clipboard.

PR #15550, opened 2024-03-12
core: Remove time limit from ExecutionLimit

Aaron1011:
The time limit check was inherently non-deterministic. Since the 'progress' events depend on how often the execution limit is hit (which determines how many frames a movie will take to preload), runtime behavior could change between different machines.

In the case of Spinin, my lazy-image-decoding change made preloading fast enough (in terms of wall-clock time) that the entire movie preloaded at once, breaking the game.

### DIFF
--- a/core/src/limits.rs
+++ b/core/src/limits.rs
@@ -1,36 +1,19 @@
 use crate::context::UpdateContext;
-use std::time::Duration;
 
 /// Indication of how long execution is allowed to take.
 ///
-/// Execution is limited by two mechanisms:
-///
-/// 1. An *operation limit*, or oplimit, which is a certain number of actions
-///    executed, bytes processed, or other things that can be done before
-///    checking...
-/// 2. A *time limit*, which is checked every time the oplimit runs out.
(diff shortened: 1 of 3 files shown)


core/src/limits.rs
```diff
@@ -67,25 +42,15 @@ impl ExecutionLimit {
...
     /// executed. The ops will be deducted from the operation limit and, if
-    /// that limit is zero, the time limit will be checked. If both limits have
-    /// been breached, this returns `true`. Otherwise, this returns `false`,
-    /// and if the operation limit was exhausted, it will be returned to the
-    /// starting maximum.
+    /// that limit has been breached, this returns 'true'
     pub fn did_ops_breach_limit(
         &mut self,
-        context: &mut UpdateContext<'_, '_>,
+        _context: &mut UpdateContext<'_, '_>,
```
Lord-McSweeney, line 48: Can the `context` parameter can be removed from here now?

torokati44: Note that there's been quite some discussion about the point or validity of this PR on Discord.

Issue #14796, opened 2024-01-17
Mouse wheel does not scroll the UIScrollbar control


n0samu:
### Describe the bug

The UIScrollBar control is Flash's built-in way of creating a basic scrolling text field. The mouse wheel is supposed to scroll the text field when the cursor is hovered over the text field. But in Ruffle, the scroll wheel has no effect.

### Expected behavior

The mouse wheel should scroll the text field.

### Content Location

[UIScrollBarTest.zip](https://github.com/ruffle-rs/ruffle/files/13961400/UIScrollBarTest.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

PR #15556, opened 2024-03-13
avm1: implemented mouse_wheel_enabled to text fields and made text scrollable

Flawake:
added mouse_wheel_enabled to text fields, made text scrollable and implemente onScroller for avm1.
closes: #14796

CC: #280


core/src/display_object/edit_text.rs
```diff
@@ -1785,6 +1785,18 @@ impl<'gc> EditText<'gc> {
...
+    fn on_scroller(&self, activation: &mut Avm1Activation<'_, 'gc>) {
+        if let Avm1Value::Object(object) = self.object() {
+            let _ = object.call_method(
+                "broadcastMessage".into(),
+                &["onScroller".into(), object.into()],
+                activation,
+                ExecutionReason::Special,
+            );
+        }
+        //TODO: Implement this for Avm2
```
Lord-McSweeney, line 1797: Is this just the mouse wheel event?

michiel2005: It is indeed for the mouse scroll event. Your PR does the AVM2 side of this, right?

adrian17: > Is this just the mouse wheel event? 

AFAIK no, in AS3 a textfield has a separate `scroll` event.

That said, I don't think it's blocking this PR?
Though what probably should be investigated is the interaction of these two PRs, as in: whether capturing this event and `.preventDefault()`ing it is supposed to disable scrolling textfields.

Lord-McSweeney: `MouseEvent.WHEEL` and `Event.SCROLL` cannot be `preventDefault`ed, so this should be fine.

PR #15566, opened 2024-03-13
desktop: Use egui-winit clipboard

ColinKinloch:
The `egui-winit` clipboard implementation provides a nice interface to `smithay-clipboard` which correctly implements the core wayland selection protocol. It uses `arboard` as a fallback for other platforms.

It also provides a fake internal clipboard as a final fallback.

https://github.com/ruffle-rs/ruffle/pull/15535


torokati44:
What is the resolution of this? 🤔

> However in testing it in a flatpak sandbox it still seems to require x11 (XWayland) to access the system clipboard.

torokati44:
But: "Look at all those dropped dependencies from `Cargo.lock`! :star_struck:"

ColinKinloch:
> What is the resolution of this? 🤔

I was using `Clipboard::new(None)` which just falls back to arboard, to get it to use the wayland protocol it needs access to the `wl_display`. I was running on fumes at that point.

> Look at all those dropped dependencies

It's a little odd, the egui version still uses `arboard`. 

I considered trying to fix arboard but as far as I can see winit has plans for a clipboard interface and I think that's a better place for that code.

ColinKinloch:
Correct, ruffle can access the clipboard under flatpak with x11 access revoked.
Here's a little script I've been using to test:
```
#!/usr/bin/env sh

RUST_PREFIX="/lib/sdk/rust-stable"
RUST_ENABLE="${RUST_PREFIX}/enable.sh"
LLVM_ENABLE="/usr/lib/sdk/llvm17/enable.sh"
JAVA_ENABLE="/usr/lib/sdk/openjdk/enable.sh"

flatpak run \
  --device=dri \
  --share=network \
  --socket=pulseaudio \
  --socket=wayland \
  --nosocket=x11 \
  --nosocket=fallback-x11 \
  --share=ipc \
  --command=bash \
  --filesystem="$(pwd)" \
  org.freedesktop.Sdk//23.08 \
  -c "source ${RUST_ENABLE} && source ${LLVM_ENABLE} && source ${JAVA_ENABLE} && cargo ${*}"

```

torokati44:
Awesome - thank you so much!

PR #15571, opened 2024-03-14
breakpoints

colin-i:
The scope was to stop at instructions and to inspect variables there. This can help for a debugger. My lib is doing actionscript/avm1 so no avm2 plans at this moment for me.

Desktop example in file "a.txt". Is my first rust code, right now is not doing the same for web but for me desktop build is enough. Web example in file "test.txt".

I pulled with write access if you want to modify something.

[a.txt](https://github.com/user-attachments/files/16608408/a.txt)
[test.txt](https://github.com/ruffle-rs/ruffle/files/14600453/test.txt)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4489,6 +4489,7 @@ dependencies = [
  "url",
  "wasm-bindgen-futures",
  "weak-table",
+ "web-sys",
  "web-time",
 ]
 
@@ -5315,6 +5316,7 @@ dependencies = [
  "lzma-rs",
  "num-derive",
  "num-traits",
+ "serde",
  "simple_asn1",
 ]
 
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -69,6 +69,7 @@ id3 = "1.16.3"
 either = "1.15.0"
 chardetng = "0.1.17"
 tracy-client = { version = "0.18.2", optional = true, default-features = false }
+web-sys = { version = "0.3.70", features = ["console"] }
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies.futures]
 workspace = true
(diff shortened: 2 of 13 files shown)


torokati44:
Hi, and welcome!
Thank you for submitting the PR, and we're really sorry that you heard nothing back so far.
The reason for this is that unfortunately we don't think this approach is workable, and nobody felt like disappointing you with this news.
The main concerns were:
 - putting the console-interaction logic directly in the middle of the interpreter main loop function is not that nice
 - some kind of stop-the-world debugger that blocks the engine entirely is not inherently a blocker - if better designed, this could be usable for some cases. the issue is that this also blocks the UI so we can't use any other debugging infrastructure we already made
 - and while we're not currently sure exactly what this better design would be, it might be a bit closer to this: https://github.com/ruffle-rs/ruffle/pull/9373
 
 Thank you for your understanding, and again, sorry!

colin-i:
> Hi, and welcome! Thank you for submitting the PR, and we're really sorry that you heard nothing back so far. The reason for this is that unfortunately we don't think this approach is workable, and nobody felt like disappointing you with this news. The main concerns were:
> 
>     * putting the console-interaction logic directly in the middle of the interpreter main loop function is not that nice
> 
>     * some kind of stop-the-world debugger that blocks the engine entirely is not inherently a blocker - if better designed, this could be usable for some cases. the issue is that this also blocks the UI so we can't use any other debugging infrastructure we already made
> 
>     * and while we're not currently sure exactly what this better design would be, it might be a bit closer to this: [[POC] Debugger #9373](https://github.com/ruffle-rs/ruffle/pull/9373)
> 
> 
> Thank you for your understanding, and again, sorry!

I am writing the assembly and have to block Ruffle and see the state of the variables at whatever ACTIONRECORD I want, and the fastest way was this, of course with command line or web flags if the user want to do this.

I am only debugging my ActionSWF and my games, this pull is more like a draft and to see if I can find another approach and at this moment it is the best option for me.

danielhjacobs:
As per the comment, I have converted this PR to a draft as a middle ground between closing it and keeping it open where it can get confused with the pull requests that do need to be reviewed in their current state.

PR #15572, opened 2024-03-14
avm2: Implement XMLList.elements

sleepycatcoding:


core/src/avm2/globals/XMLList.as
```diff
@@ -24,6 +24,7 @@ package {
         AS3 native function comments():XMLList;
         AS3 native function parent():*;
         AS3 native function processingInstructions(name:* = "*"):XMLList;
+        AS3 native function elements(name:* = "*"):XMLList;
```
evilpie, line 27: Can you change XML.as/rs to use this `"*"` as well?

evilpie (approved):

core/src/avm2/globals/xml_list.rs
```diff
@@ -93,6 +93,40 @@ pub fn call_handler<'gc>(
...
+// ECMA-357 13.5.4.11 XMLList.prototype.elements ([name])
+pub fn elements<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Object<'gc>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let this = this.as_xml_list_object().unwrap();
+    // 2. Let name = ToXMLName(name)
+    let multiname = name_to_multiname(activation, &args[0], false)?;
+    let mut children = this.children_mut(activation.gc());
```
evilpie, line 105: nit: Move this right after `// 4. For i = 0 ...` and before `for`.

core/src/avm2/object/xml_object.rs
```diff
@@ -115,6 +115,36 @@ impl<'gc> XmlObject<'gc> {
...
+                .collect()
+        } else {
+            Vec::new()
+        };
+
+        let list = XmlListObject::new_with_children(
+            activation,
+            children,
+            Some(XmlOrXmlListObject::Xml(*self)),
+            None,
```
evilpie, line 137: Shouldn't this use `Some(name)`?

> 13.4.4.13 XML.prototype.elements
> 3. Let list be a new XMLList with list.[[TargetObject]] = x and **list.[[TargetProperty]] = name**

sleepycatcoding: I think this was because avmplus never set the target property here:
https://github.com/adobe/avmplus/blob/858d034a3bd3a54d9b70909386435cf4aec81d21/core/XMLObject.cpp#L1695

evilpie: Oh, fun. Add a note?

sleepycatcoding: Added a note.

PR #15576, opened 2024-03-14
core: Cull text under textfield

adrian17:
This should optimize cases where there's a lot of text left to be scrolled under the textfield. The speedup can be extreme if there are hundreds of pointlessly-drawn lines.

(However, note that this doesn't help with the opposite case - when you scroll to the bottom, it'll still draw all the text :/  )

The perfect solution might be to implement all 4 types of culling (lines under/above* textfield/screen), but with all the transforms, height metrics etc etc this was the only one I could confidently implement - and even in this one I'm not convinced I didn't skip any edge cases.

(Note that due to how `scrollV` works, lines above the textfield could be skipped without even iterating over them. In fact... maybe FP (and us) could deduce in-bounds for each glyph, thus making us not require masking at all?)

### DIFF
--- a/core/src/display_object/edit_text.rs
+++ b/core/src/display_object/edit_text.rs
@@ -953,12 +953,27 @@ impl<'gc> EditText<'gc> {
     /// Render a layout box, plus its children.
     fn render_layout_box(self, context: &mut RenderContext<'_, 'gc>, lbox: &LayoutBox<'gc>) {
         let origin = lbox.bounds().origin();
+
+        let edit_text = self.0.read();
+
+        // If text's top is under the textbox's bottom, skip drawing.
+        // TODO: FP actually skips drawing a line as soon as its bottom is under the textbox;
+        //   Current logic is conservative for safety (and even of this I'm not 100% sure).
+        // TODO: we should also cull text that's above the textbox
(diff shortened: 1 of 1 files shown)


adrian17:
Here's a demonstration.
This was recorded with text masking **disabled**, so text outside the textfield is visible - except lines culled by this PR.
Right textfield is normal, left textfield has `scrollV = 3`.

https://github.com/ruffle-rs/ruffle/assets/4729533/943716bd-3b97-4c88-afaf-60cc9fa14020

core/src/display_object/edit_text.rs
```diff
@@ -953,13 +953,28 @@ impl<'gc> EditText<'gc> {
     /// Render a layout box, plus its children.
     fn render_layout_box(self, context: &mut RenderContext<'_, 'gc>, lbox: &LayoutBox<'gc>) {
         let origin = lbox.bounds().origin();
+
+        let edit_text = self.0.read();
+
+        // If text's top is under the textbox's bottom, skip drawing.
+        // TODO: FP actually skips drawing a line as soon as it's bottom is under the textbox;
```
Lord-McSweeney, line 960: typo - "it's bottom" -> "its bottom"

Lord-McSweeney (approved):

PR #15579, opened 2024-03-14
avm2: Provide an option to disable the optimizer

adrian17:
Might be useful if we ever suspect a bug in the optimizer.

Feel free to bikeshed on naming or cli stuff :D

### DIFF
--- a/desktop/src/cli.rs
+++ b/desktop/src/cli.rs
@@ -195,6 +195,12 @@ pub struct Opt {
         value_name = "GAMEPAD BUTTON>=<KEY NAME"
     )]
     pub gamepad_button: Vec<(GamepadButton, KeyCode)>,
+
+    /// Disable AVM2 optimizer.
+    /// Note that some early opcode conversions
+    /// (like inlining constant pool entries) can't be disabled.
+    #[clap(long)]
+    pub no_avm2_optimizer: bool,
 }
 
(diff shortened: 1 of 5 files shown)


Lord-McSweeney:
LGTM, but note that this will have to be removed once we properly merge the verifier and optimizer (which will allow for, for example, eliminating `Nop`s)

adrian17:
I don't think this will necessarily need to be removed; this could still inhibit most of the optimization-only logic while keeping type propagation.

desktop/src/cli.rs
```diff
@@ -195,6 +195,12 @@ pub struct Opt {
         value_name = "GAMEPAD BUTTON>=<KEY NAME"
     )]
     pub gamepad_button: Vec<(GamepadButton, KeyCode)>,
+
+    /// Disable AVM2 optimizer.
+    /// Note that some early opcode conversions
+    /// (like inlining constant pool entries) can't be disabled.
+    #[clap(long)]
+    pub no_avm2_optimizer: bool,
```
torokati44, line 203: I know it's late, and not a big deal... But I have a similar situation going on with the disablization of OpenH264 over at #14654. And I deliberately chose the "positive meaning" to be the flag, and set the default value to `true`.
Has this been considered and decided against here, or was it done just so `false` can be kept as the default flag value?

Issue #5443, opened 2021-10-09
Tab key behavior from Adobe Flash Player not present in Ruffle


Sunlitspace542:
**Describe the feature**
Provide a clear and concise description of your feature request. Why would this feature be useful? Does the feature involve the desktop build, the web build, or both?

This was a feature the original Flash Player had, if you press the tab key, it cycles through each clickable item, highlighted with a yellow box around the item. pressing enter "clicks" on the item.
Screenshot attached with example of what I'm talking about using a Flash projector:
![image](https://user-images.githubusercontent.com/35207254/141229727-b9b7bcc5-b0bc-4a70-8a0a-977a76dc185f.png)

Erisfiregamer1:
Would be nice.

mero09-99:
I agree

battaglia01:
Kind of curious how hard this would be to implement. Has anyone looked into this? What does the tab key do now?

ActionWavele:
This is very important for accecibility

Atario:
> What does the tab key do now?

Nothing, looks like

VL4DST3R:
Any updates on this? This and the zoom functionality are some of the most noticeable pain-points with ruffle compared to native flash.

rikkoooo:
This is ultra-very important, and maybe even more important than the mouse support. And, this is a bug report, not an enhancement request.

battaglia01:
Ok. Does anyone have any idea how to implement this? Where would one poke around to do this?

kjarosh:
Implementation status:

- [ ] Tab key support (https://github.com/ruffle-rs/ruffle/pull/15542)
- [ ] AVM1 support for `tabEnabled`, `tabIndex` (https://github.com/ruffle-rs/ruffle/pull/15542)
- [ ] AVM1 support for `tabChildren`
- [ ] AVM2 support
- [ ] Analysis and implementation of a proper automatic order
- [ ] Visual highlight around the tabbed element

Croworbit:
does it work across other OSes?

kjarosh:
> does it work across other OSes?

This feature is platform-independent. Should work on every OS and on web (although I am not able test it on all OSes).

RavenAgentDev:
Note: It works on the browser version (as in the swf file can detect its being pressed)

kjarosh:
I think the tab key support is in a pretty good shape now. Referenced the two remaining TODOs in new issues.

Closing as completed!

PR #15580, opened 2024-03-14
avm1: Support `MovieClip.tabChildren`

kjarosh:
Related to #5443. Improves coverage for AVM1: #263.

Requires #15542.

The property `MovieClip.tabChildren` allows changing the behavior of tab ordering hierarchically. When set to `false`, it excludes the whole subtree represented by the movie clip from tab ordering.


core/src/focus_tracker.rs
```diff
@@ -100,6 +100,8 @@ impl<'gc> FocusTracker<'gc> {
             tab_order.first()
         };
 
-        self.set(next.copied(), context);
+        if next.is_some() {
+            self.set(next.copied(), context);
```
adrian17, line 104: To be sure, when can this happen? As in... only if previously the entire list was non-empty and we had a focused item, but now is entirely empty?

adrian17 (approved):

core/src/focus_tracker.rs
kjarosh: This happens when there is a focus, but tab ordering is empty, i.e. someone tabbed and tab ordering changed (e.g. setting `tabEnabled = false`), or the focus was set programmatically (`Selection.setFocus`).

This is tested by `tab_ordering_children`, when you comment out this `if` condition it will fail.

PR #15594, opened 2024-03-16
avm2: URLStream support for connected and HTTP_STATUS event

evilpie:


core/src/avm2/globals/flash/net/URLStream.as
```diff
@@ -74,10 +80,12 @@ package flash.net {
 
         public function load(request:URLRequest):void {
             this._loader.load(request);
+            this._connected = true;
```
adrian17, line 85: I'm surprised it's this easy. Does it work this way even if the request fails (for example, the URLRequest url 404s)?

evilpie: Yes. `connected` seems to stay true no matter if the request succeeds or fails.

adrian17 (approved):
> Yes. connected seems to stay true no matter if the request succeeds or fails.

Approved, but this is not really true - if the request throws immediately (say, a local file request on an swf without appropriate SandboxType), it stays `false`. I'm not sure if we already support this example or not.

evilpie:
> > Yes. connected seems to stay true no matter if the request succeeds or fails.
> 
> Approved, but this is not really true - if the request throws immediately (say, a local file request on an swf without appropriate SandboxType), it stays `false`. I'm not sure if we already support this example or not.

If `this._loader.load(request)` throws `connected` would stay `false`. However it seems like our implementation of the URLLoader will never throw synchronously, because we always spawn `load_data_into_url_loader`.

PR #15627, opened 2024-03-18
avm1: made the context menu work for all avm1 objects

Flawake:
Made the context menu work on all AVM1 objects, was only the root before. 
Shows the context menu when clicking instead of releasing on desktop like flash player did.
Don't add a context menu item when there already is a item with the same caption present. (avm1)

.swf and .fla used for testing:
[context menu.zip](https://github.com/ruffle-rs/ruffle/files/14652961/context.menu.zip)

### DIFF
--- a/core/src/player.rs
+++ b/core/src/player.rs
@@ -566,10 +566,9 @@ impl Player {
                 return vec![];
             }
 
-            // TODO: This should use a pointed display object with `.menu`
-            let root_dobj = context.stage.root_clip();
+            let display_obj = Player::get_context_menu_display_object(context);
 
-            let menu = if let Some(Value::Object(obj)) = root_dobj.map(|root| root.object()) {
+            let menu = if let Some(Value::Object(obj)) = display_obj.map(|obj| obj.object()) {
                 let mut activation = Activation::from_stub(
                     context.reborrow(),
                     ActivationIdentifier::root("[ContextMenu]"),
@@ -588,11 +587,11 @@ impl Player {
                     None
                 };
 
-                crate::avm1::make_context_menu_state(menu_object, &mut activation)
-            } else if let Some(Avm2Value::Object(hit_obj)) = root_dobj.map(|root| root.object2()) {
+                crate::avm1::make_context_menu_state(menu_object, display_obj, &mut activation)
+            } else if let Some(Avm2Value::Object(hit_obj)) = display_obj.map(|obj| obj.object2()) {
                 let mut activation = Avm2Activation::from_nothing(context.reborrow());
 
-                let menu_object = root_dobj
+                let menu_object = display_obj
                     .expect("Root is confirmed to exist here")
                     .as_interactive()
                     .map(|iobj| iobj.context_menu())
@@ -619,7 +618,7 @@ impl Player {
                     Avm2::dispatch_event(&mut activation.context, menu_evt, menu_object);
                 }
 
-                crate::avm2::make_context_menu_state(menu_object, &mut activation)
+                crate::avm2::make_context_menu_state(menu_object, display_obj, &mut activation)
             } else {
                 // no AVM1 or AVM2 object - so just prepare the builtin items
                 let mut menu = ContextMenuState::new();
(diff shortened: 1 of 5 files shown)


core/src/player.rs
```diff
@@ -628,14 +627,13 @@ impl Player {
                     ContextMenuCallback::Back => Self::back_root_movie(context),
                     ContextMenuCallback::Rewind => Self::rewind_root_movie(context),
                     ContextMenuCallback::Avm2 { item } => {
-                        // TODO: This should use the pointed display object (see comment on line 614)
-                        let root_dobj = context.stage.root_clip();
+                        let menu_item = *item;
 
-                        if let Some(root_dobj) = root_dobj {
-                            let menu_item = *item;
+                        if let Some(display_obj) = Player::get_context_menu_display_object(context)
```
adrian17, line 632: I don't think the object is re-picked once you click the option, right? I think we need to call the callback with the object that was originally right-clicked?

Flawake: It is indeed better if we can cache the picked object. But it is not certain that it is run on the right clicked object because it run's on the root if the right clicked object does not have a custom context menu defined. I can probably store the object in the GcRootData struct. But is that worth it?

adrian17: > because it run's on the root if the right clicked object does not have a custom context menu defined

No, this is normal - the "event" should propagate up the tree from the clicked DO until it encounters a DO with a menu (so at a top, the root).
Oh, I only now realized this isn't part of the PR. Then yeah, it needs to - otherwise the context menus will actually break in most cases, as usually when we right click, the cursor is on some specific DO that's a child of the menu owner; right?

Also you can experiment with FP events to observe this (we don't have these fields implemented on our side yet:
[menuSelect ](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/ui/ContextMenu.html#event:menuSelect)
[menuItemSelect](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/ui/ContextMenuItem.html#event:menuItemSelect)
And their fields contextMenuOwner, mouseTarget etc.

> It is indeed better if we can cache the picked object. 

It's not about caching, it's about correctness. when I right click on something and select an item but the mouse happens to hover over a different DO when clicking, then the code can run on a different object than expected.

> I can probably store the object in the GcRootData

We already do store related stuff in ContextMenuState.
ContextMenuCallback

Flawake: I can't find the displayObject stored in contextMenuState. I only seems to have a Vec of ContextMenuItem and a Vec of callbacks. Wich both don't seem to hold the clicked display object.

adrian17: > I can't find the displayObject stored in contextMenuState.

Yup, I meant to put the new field there instead of in GcRootData.

Flawake: Everything should have the correct behavior now. The function get_context_menu_display_object is now only called at a single place. I think that it makes the code more readable to have this functionality in a function. Or should I remove the funtion and just add the code where the function is being called?

Dinnerbone (approved):
Sorry about leaving this so long. Thank you very much!
I've force pushed to rebase onto master, as there was some conflicts.

core/src/player.rs
```diff
@@ -712,23 +708,54 @@ impl Player {
...
+    }
+
+    ///Returns the first display object that the mouse is hovering over that has a custom context menu. Returns root if none is found.
+    fn get_context_menu_display_object<'gc>(
+        context: &mut UpdateContext<'_, 'gc>,
+    ) -> Option<DisplayObject<'gc>> {
+        let mut display_object = run_mouse_pick(context, false).map(|obj| obj.as_displayobject());
+        loop {
+            if let Some(disp_obj) = display_object {
+                if disp_obj.hit_test_shape(
```
adrian17, line 738: wait, why is this hit test needed? Can the reason be documented? Or if you've found an edge case, let me know and I can try make a test.
Intuitively, if we've found a hit via `run_mouse_pick`, we shouldn't need to test for hit tests again while recursing upward.

Flawake: I don't exactly know what I thought 5 months ago, but I think I did indeed mess up.
I can't find a reason for this test after testing with my test swf I made back then.
I do remember using the flash player that came with macromedia flash 8 back then which did not give the same results using the latest flash player. So I might have taught this was needed because of inconsistencies with the old flash player I was using and the newest one.
Nevertheless, this does indeed not seem necessary.

PR #15634, opened 2024-03-18
chore: Bump `image` to `v0.25`, without the `dxt` feature

torokati44:
I hope some visual tests will fail on this, at least on Windows.
To be fixed by an alternative implementation based on https://github.com/jansol/texpresso, as recommended.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1529,7 +1529,7 @@ checksum = "3f4a6962241a76da5be5e64e41b851ee1c95fda11f76635522a3c82b119b5475"
 dependencies = [
  "egui",
  "enum-map",
- "image",
+ "image 0.24.9",
  "log",
  "mime_guess2",
  "serde",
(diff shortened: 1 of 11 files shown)


torokati44:
Uuuuhm. Huhh. How does this still work, @Aaron1011 ? :thinking: 
Do we simply not have a visual test for DXT-compressed textures?
I have no idea how ATF actually works together with the different algorithms... :/

Aaron1011:
I added the `dxt` feature early on when working on compressed alpha support, but we don't actually use it - we produce a TIFF image from the jpegxr crate, and then decode that using the `image` crate.

Aaron1011 (approved):

torokati44:
Oooh, I see. Thanks for the explanation!
So we're actually always taking the "compatibility fallback" option out of the ATF, and pre-decompress it on the CPU for the sake of portability, instead of making use of the other GPU-friendly compressed alternatives in there?

PR #15647, opened 2024-03-19
desktop: Improve backend selection

Dinnerbone:
We now try every single backend one-by-one to work around https://github.com/gfx-rs/wgpu/issues/2384 & https://github.com/gfx-rs/wgpu/issues/5324

### DIFF
--- a/desktop/src/gui/controller.rs
+++ b/desktop/src/gui/controller.rs
@@ -112,7 +112,6 @@ impl GuiController {
             initial_movie_url.clone(),
             PlayerOptions::from(&preferences),
             preferences.clone(),
-            descriptors.clone(),
         );
         let system_fonts =
             load_system_fonts(font_database, preferences.language().to_owned()).unwrap_or_default();
--- a/desktop/src/gui/controller.rs
+++ b/desktop/src/gui/controller.rs
@@ -48,30 +48,7 @@ impl GuiController {
         initial_movie_url: Option<Url>,
         no_gui: bool,
     ) -> anyhow::Result<Self> {
-        let mut backend: wgpu::Backends = preferences.graphics_backends().into();
-        if wgpu::Backends::SECONDARY.contains(backend) {
-            tracing::warn!(
-                "{} graphics backend support may not be fully supported.",
-                format_list(&get_backend_names(backend), "and")
-            );
-        }
-        let mut instance = wgpu::Instance::new(wgpu::InstanceDescriptor {
-            backends: backend,
-            flags: wgpu::InstanceFlags::default().with_env(),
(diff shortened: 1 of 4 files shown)


desktop/src/gui/controller.rs
```diff
@@ -361,6 +337,52 @@ impl GuiController {
...
+
+    tracing::warn!(
+        "Preferred backend(s) of {} not available; falling back to any",
+        format_list(&get_backend_names(preferred_backends), "or")
+    );
+
+    for backend in wgpu::Backends::all() {
+        if let Some(instance) = try_wgpu_backend(backend) {
+            tracing::info!(
+                "Using preferred backend {}",
```
sleepycatcoding, line 361: I think this should say "Using backend" or "Using any backend", since this is only reached if the preferred backend(s) weren't available.

Dinnerbone: Copy paste mistake, thanks :D

desktop/src/gui/controller.rs
```diff
@@ -361,6 +338,52 @@ impl GuiController {
...
+            return Ok((instance, backend));
+        }
+    }
+
+    tracing::warn!(
+        "Preferred backend(s) of {} not available; falling back to any",
+        format_list(&get_backend_names(preferred_backends), "or")
+    );
+
+    for backend in wgpu::Backends::all() {
```
adrian17, line 358: this is a `bitflags`, so we could subtract the preferred ones to not try them twice.

desktop/src/gui/preferences_dialog.rs
```diff
@@ -320,14 +314,29 @@ fn filename_pattern_name(locale: &LanguageIdentifier, pattern: FilenamePattern)
...
-    {
+fn backend_availability(instance: &wgpu::Instance, backend: wgpu::Backends) -> wgpu::Backends {
+    if instance.enumerate_adapters(backend).is_empty() {
         wgpu::Backends::empty()
     } else {
         backend
     }
 }
+
+fn find_available_graphics_backends() -> wgpu::Backends {
```
adrian17, line 325: earlier you mentioned that

> I think we can remove that code now, wgpu added an api to show which backends are (roughly) available

Just wanna make sure whether that was checked :D

adrian17 (approved):
approve, assuming review notes are checked

desktop/src/gui/preferences_dialog.rs
Dinnerbone: Yeah, turns out that's unreleased still :(
`wgpu::Instance::enabled_backend_features()` in the next major wgpu update

PR #15703, opened 2024-03-21
build: Silence clippy assigning_clones lint

adrian17:


torokati44 (approved):

torokati44:
I suppose the other lints can be cleaned up in separate PR(s).

torokati44 (approved):
Agh, messy-messy... But this is what I would have done too (for now), and at least we get back our green checkmarks!

PR #15717, opened 2024-03-22
core: Render yellow rectangular highlight on keyboard focus

kjarosh:

This patch implements rendering of the yellow rectangle around a focused element after pressing Tab.

| output 01_highlight_under expected | output 02_highlight_over expected |
|---|---|
| image | image |


kjarosh:
It seems that I accidentally improved Ruffle compatibility regarding the `visual/shumway_acid_tests/acid` test, which... made it fail.

Look at its left eye :monocle_face: 

| FP | Ruffle before | Ruffle after |
|---|---|---|
|  |  |  |

torokati44:
Plus the little rectangle on the bottom left is now rotated as it should be. The little block under its chin looks more different though... 🤔

kjarosh:
> Plus the little rectangle on the bottom left is now rotated as it should be. The little block under its chin looks more different though... 🤔

Ohh you shouldn't really look at that. Those elements are animated and I picked a random frame (I should've mentioned that!)

torokati44:
Ah, I see! Then perhaps we should compare more frames, to also test the animation? Now that there is support for that... :) Not in scope of this PR of course.

core/src/focus_tracker.rs
```diff
@@ -110,6 +148,43 @@ impl<'gc> FocusTracker<'gc> {
...
+        let Some(focus) = self.get() else {
+            return Highlight::Inactive;
+        };
+
+        if !focus.is_highlight_enabled() {
+            return Highlight::Inactive;
         }
+
+        let bounds = focus.world_bounds().grow(-Self::HIGHLIGHT_WIDTH / 2);
+        let mut drawing = Drawing::new();
```
Dinnerbone, line 169: Using Drawing for this is going to be a little expensive on the rendering side, but realistically this isn't going to be hit enough to be noticed (I hope?)
Ideally we have a draw command that's "draw box outline" later, and let the render backend optimise that. It can make it pixel perfect too, which would be nice.

Not something you need to change for this PR - just a comment!

Dinnerbone (approved):
Looks cool! Thank you!

PR #15741, opened 2024-03-23
core: Remove `Collect` from `Color` and `DirtyState` as it was unnecessary

Lord-McSweeney:


torokati44 (approved):
I mean, GC is not my area of expertise, but I'm not seeing anything obviously wrong here...?

PR #15747, opened 2024-03-23
core: Make sure that streams progress even with many small-dt ticks

torokati44:
This fixed the playback of FLV videos for me on Linux.

Rationale:
Currently, `stream_time` is only advanced when there was at least one tag processed in a tick.
However, in certain conditions (high rendering FPS, low video FPS, large audio chunks), it's possible that no tick ever has a `dt` large enough to bridge the time gap from the last processed tag, to the next. In this case, the stream got stuck.
With this change, any and all `dt` passed into `tick` will be accumulated into `stream_time`, no matter if it triggered any tag processing or not.

### DIFF
--- a/core/src/streams.rs
+++ b/core/src/streams.rs
@@ -1108,7 +1108,6 @@ impl<'gc> NetStream<'gc> {
         let buffer = slice.data();
 
         let max_time = write.stream_time + dt;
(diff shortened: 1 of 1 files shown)


kmeisthax (approved):
Can't believe I wound up writing a "less lag equals more lag" type of loop 😅

PR #15774, opened 2024-03-25
ci: Split cargo-deny into its own job, add cargo-machete

torokati44:
Contrary to the source branch name, I went with `cargo-machete` instead of `cargo-udeps`: https://blog.benj.me/2022/04/27/cargo-machete/

Unfortunately, both point to a couple false positives, so they would need some manual handholding, polluting our `Cargo.toml` files. ( https://github.com/bnjbvr/cargo-machete/issues/118 )

~~Do we even need this...? :/~~

### DIFF
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -153,6 +153,11 @@ jobs:
       - name: Check licenses, duplicates, and advisories
         uses: EmbarkStudios/cargo-deny-action@v1
 
+      - name: Check unused dependencies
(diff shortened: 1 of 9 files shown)


torokati44:
Turns out, machete only needed two ignores, for fairly reasonable ... reasons.
I still wish we could shove this into a single, specific config file somewhere...

torokati44:
Well it didn't turn out that bad IMHO... :)

Dinnerbone (approved):

Issue #15157, opened 2024-02-07
Ruffle Linux incorrectly scaling flash titles.


Spikestuff:
### Describe the bug

Something I noticed as it's more apparent in the --no-gui field compared to a standard open, is that ruffle incorrectly loads titles by not launching them at their native resolution.

[Samsara's Room](https://www.newgrounds.com/portal/view/624750) is being used here as an example, a 720x480 title. This will happen to any title.
But here in the Linux release of Ruffle it's being displayed at 800x534 (and that's not counting the 66 pixels total above and below the picture) a random 111% larger.

![image](https://github.com/ruffle-rs/ruffle/assets/12412764/66cc2689-2906-4bff-92ff-0e7aaefa5d8a)

### Expected behavior

Here's the Windows version running the --no-gui command correctly:
![image](https://github.com/ruffle-rs/ruffle/assets/12412764/003e6656-0bc4-4370-a8dd-f87366d41eaf)

### Content Location

The title tested was [Samsara's Room](https://www.newgrounds.com/portal/view/624750) but this issue persists with any other title.

### Affected platform

Desktop app

### Operating system

Ubuntu 22.04.3 LTS (Jammy Jellyfish)

### Additional information

I validated my Display Settings in Linux, and it's 2560x1440 @ 100% Window Scaling.
I also validated ScummVM for sanity reasons, and its Linux window matches Windows when launching a title that I had on hand being Riven.

torokati44:
I think this is #8870.

torokati44:
The window is 800x600, which is a sort of "factory default" of Ruffle, and the content is resized accordingly, with the black bars added as letterboxing to keep the original spect ratio.

PR #15781, opened 2024-03-26
desktop: Fix gui scale on launch on wayland

ColinKinloch:
On platforms that support synchronous resizing such as wayland `request_inner_size` returns the new window size and no `WindowEvent::Resize`  is generated ([documented here](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.request_inner_size)).

This MR resizes the gui if `request_inner_size` returns a value different from its original value.

| before | after |
|-|-|
| Screenshot from 2024-03-26 12-49-02 | Screenshot from 2024-03-26 12-44-55 |

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -419,7 +419,15 @@ impl App {
                         self.window.scale_factor(),
                     );
 
-                    let _ = self.window.request_inner_size(window_size);
+                    let viewport_size = self.window.inner_size();
+
+                    if let Some(new_viewport_size) = self.window.request_inner_size(window_size) {
+                        if new_viewport_size != viewport_size {
+                            self.gui.borrow_mut().resize(new_viewport_size);
+                        } else {
+                            tracing::warn!("Unable to resize window");
+                        }
+                    }
                     self.window.set_fullscreen(if self.start_fullscreen {
                         Some(Fullscreen::Borderless(None))
                     } else {
--- a/desktop/src/gui/controller.rs
+++ b/desktop/src/gui/controller.rs
@@ -141,31 +141,35 @@ impl GuiController {
         &self.descriptors
     }
 
+    pub fn resize(&mut self, size: PhysicalSize<u32>) {
+        if size.width > 0 && size.height > 0 {
+            self.surface.configure(
+                &self.descriptors.device,
+                &wgpu::SurfaceConfiguration {
+                    usage: wgpu::TextureUsages::RENDER_ATTACHMENT,
+                    format: self.surface_format,
+                    width: size.width,
+                    height: size.height,
+                    present_mode: Default::default(),
+                    desired_maximum_frame_latency: 2,
+                    alpha_mode: Default::default(),
+                    view_formats: Default::default(),
+                },
+            );
+            self.movie_view_renderer.update_resolution(
+                &self.descriptors,
+                self.window.fullscreen().is_none() && !self.no_gui,
+                size.height,
+                self.window.scale_factor(),
+            );
+            self.size = size;
+        }
+    }
+
     #[must_use]
     pub fn handle_event(&mut self, event: &WindowEvent) -> bool {
         if let WindowEvent::Resized(size) = &event {
-            if size.width > 0 && size.height > 0 {
-                self.surface.configure(
-                    &self.descriptors.device,
-                    &wgpu::SurfaceConfiguration {
-                        usage: wgpu::TextureUsages::RENDER_ATTACHMENT,
-                        format: self.surface_format,
-                        width: size.width,
-                        height: size.height,
-                        present_mode: Default::default(),
-                        desired_maximum_frame_latency: 2,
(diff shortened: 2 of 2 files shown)


torokati44:
Huh, I've never had this problem on KWin before.
I've only had #8870, which this still doesn't fix... Maybe KDE 6 will? :/

Can someone test on Windows please?

X11 (via XWayland) is unaffected, has the correct windows size both before and after.

torokati44:
Maybe this fixes https://github.com/ruffle-rs/ruffle/issues/15157 though...?

ColinKinloch:
I'm not sure, the window is the same size, but the image seems to have less gross filtering.
| before | after |
|-|-|
| Screenshot from 2024-03-26 13-26-01 | Screenshot from 2024-03-26 13-25-24 |

Is there a proper pixel art swf you can point me to?

ColinKinloch:
> Huh, I've never had this problem on KWin before.

I assume if you run it with `WAYLAND_DEBUG=1` you'll see a `xdg_toplevel@*configure` when ruffle resizes itself? This doesn't happen on mutter, I assume it's because it is a [request](https://wayland.app/protocols/xdg-shell#xdg_toplevel:event:configure) from the compositor to change size (drag resize) rather than a notification that it has resized.

torokati44:
Huh, weird.

Not quite pixel art, but https://z0r.de/L/z0r-de_420.swf for example is relatively small, and should be pixelated (possibly due to #10103, but whatever).
So it's easy to see if the UI is scaled incorrectly, and/or if the window content is stretched by the compositor.

torokati44:
Yes, I see those requests. So maybe it's KWin who insists on this "factory-default" size too much?

```bash
attila ~/projects/ruffle $ git checkout origin/master -f
HEAD is now at 9dfc6dc22 fix(deps): update rust dependencies
attila ~/projects/ruffle $ WAYLAND_DEBUG=1 cargo run -- 'https://z0r.de/L/z0r-de_420.swf' 2>&1 | grep -P 'xdg_toplevel@.*configure'
[2244621.129] xdg_toplevel@29.configure_bounds(1920, 1016)
[2244621.135] xdg_toplevel@29.configure(0, 0, array[0])
[2245301.342] xdg_toplevel@29.configure(800, 600, array[4])
[2245625.470] xdg_toplevel@29.configure(800, 600, array[4])
attila ~/projects/ruffle $ git checkout ColinKinloch/wl_resize_gui -f
Previous HEAD position was 9dfc6dc22 fix(deps): update rust dependencies
HEAD is now at 0cdde301a desktop: Fix gui scale on launch on wayland
attila ~/projects/ruffle $ WAYLAND_DEBUG=1 cargo run -- 'https://z0r.de/L/z0r-de_420.swf' 2>&1 | grep -P 'xdg_toplevel@.*configure'
[2287138.117] xdg_toplevel@29.configure_bounds(1920, 1016)
[2287138.120] xdg_toplevel@29.configure(0, 0, array[0])
[2287811.485] xdg_toplevel@29.configure(800, 600, array[4])
[2288157.708] xdg_toplevel@29.configure(800, 600, array[4])
attila ~/projects/ruffle $ 
```

ColinKinloch:
This bug doesn't manifest if the swf is passed as an argument. I assume ruffle creates the windows/gui if it's already got the file.

ColinKinloch:
| before | after |
|-|-|
| Screenshot from 2024-03-26 13-49-06 | Screenshot from 2024-03-26 13-48-00 |

torokati44:
On my machine, it looks and behaves exactly the same way before and after this PR, no matter if I supply the SWF on the command line, or in the "Open" dialog.
On your screenshots, the "after" really does look better, so I'm not against this change.
I'll only have time to look at the code later though.

ColinKinloch:
I had a look at it in a nested kwin instance.
Pretty weird, it goes from the initial 800x600 to the requested size (324x216) on load then returns to 800x600 for some reason. :shrug:

torokati44:
Yep, that's what I'm seeing too. Sad.
Thank you for checking!

ColinKinloch:
More testing. The size reset issue seems to also exist in egui.

And here's this MR a nested weston session:
| before | after |
|-|-|
| Screenshot from 2024-03-26 17-56-48 | Screenshot from 2024-03-26 17-57-32 |

parkerlreed:
Current nightly on Arch Plasma 6 seems to keep the regular 1x GUI scale. 

What's the test case that's supposed to trigger this?

torokati44:
What I said about Plasma 6 is not a matter of GUI scale, simply window size not matching the content size (as opposed to being reset to 800x600). So these are two related, but not exactly the same, issues.

torokati44:
Ah, right, testing in Weston, I get this before (currently):
image

And with this PR:
image

_It's hard to see_, but the content matches the window in size _just a little bit_ better! :smile:

torokati44 (approved):
Thank you yet again! :partying_face:

PR #15782, opened 2024-03-26
ci: Cleanup "skipped" test jobs

torokati44:
The non-expanded `${{ expressions }}` in the check names annoyed me incredibly.

Also, I don't think there is any need for the matrix to be there - but we'll see.

Cc: @midgleyc, who set this up IIRC. (Also Cc: @relrelb, just in case.)

### DIFF
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -145,17 +145,8 @@ jobs:
   check-required:
     needs: changes
     if: needs.changes.outputs.should_run == 'false'
-    name: Test Rust ${{ matrix.rust_version }} / ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


torokati44:
> in settings, you can mark some actions as "required" in order to be able to merge a PR

Oooh, riight, OMG.

I keep forgetting this silly anti-feature...

Sorry then.

:100: https://twitter.com/jessfraz/status/1771578781511438729

torokati44:
Also, thank you for the reminder and the explanation! ^^

torokati44:
BTW This is _so bad_, that some people have made a dedicated project to do properly what it's supposed to be doing: :facepalm: 

https://github.com/giantswarm/pr-gatekeeper

PR #15787, opened 2024-03-27
avm2: Add basic state merging for forward jumps to optimizer

Lord-McSweeney:
This gives some sizable improvements in Box2d:
```
Coerce -> Nop: +2.68%
SetProperty -> SetSlot: +9.06%
GetProperty -> GetSlot: +8.16%
CallProperty -> CallMethod: +6.05%
CoerceD -> Nop: +4.23%
CoerceB -> Nop: +2.70%
```

New stats:
```
InitProperty -> SetSlot : 74.39%
Coerce -> Nop : 46.58%
SetProperty -> SetSlot : 45.3%
GetProperty -> GetSlot : 62.97%
GetProperty -> CallMethod : 0.36%
CallProperty -> CallMethod : 31.05%
CoerceD -> Nop : 78.36%
CoerceB -> Nop : 8.11%
CoerceU -> Nop : 32.46%
CoerceI -> Nop : 2%
```


core/src/avm2/optimize.rs
```diff
@@ -251,15 +255,49 @@ pub fn optimize<'gc>(
...
     let mut local_types = initial_local_types.clone();
 
     for (i, op) in code.iter_mut().enumerate() {
-        if jump_targets.contains(&(i as i32)) {
+        if let Some(sources) = jump_targets.get(&(i as i32)) {
+            // Avoid handling multiple sources for now
+            if sources.len() == 1 {
+                // We can merge the locals easily, now
+                let source_i = sources[0];
+                // Because of the linear optimizer logic, this guarantees that the jump source came before the target
```
kmeisthax, line 271: Will the optimizer architecture change to the point where we'd need to worry about backwards jumps at some point? Or would we be splitting up optimization passes into separate functions by then?

kmeisthax (approved):

core/src/avm2/optimize.rs
Lord-McSweeney: Yes, we will definitely have to worry about backwards jumps (since we need to implement full state merging for the verifier/optimizer, e.g. to verify the stack and scope stack and optimize `findproperty`), but that will probably be after the verifier is merged with the optimizer, which will be a near-full rewrite of all this code.

Lord-McSweeney:
Drafting this as I found a bug

torokati44:
Two approvals should be enough, right...? :)

PR #15810, opened 2024-03-30
core: Improve order of automatic tab ordering

kjarosh:

This patch improves the order of automatic tab ordering, and makes it behave more like in FP. The order after this patch is far from being exactly the same as in FP, but is close enough.

In reality, FP sometimes jumps up with the focus or even ignores some arbitrary objects. However, the exact behavior is very complicated and requires a lot more testing and analysis.


Dinnerbone (approved):
LGTM, thank you!

Issue #310, opened 2020-01-11
Actionscript Tracking: Top Level


Dinnerbone:
This is the tracking issue for **Top Level** in AVM1. We will check off each item of progress as appropriate, and fill in any relevant or missing information as we continue development of Ruffle.
# Legend
Each checkbox is independent of another. It's entirely possible for something to be tested but not exist yet, or for us to believe that it's completely functional but we haven't made enough tests to prove it.
## "Property Exists"
This means the item exists, but may not necessarily be fully implemented.
## "Has Test Coverage"
This means that we have believe that we have a good test coverage of this item, regardless of if those tests pass. It's okay to have tests available but not implement the item yet.
## "Completely Functional"
This means we believe that the item is completely implemented, and no more work needs to be done towards making it functional.

# Top Level
### `_focusRect : Boolean`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_global : Object`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_highquality : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_quality : String`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_root : MovieClip`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_soundbuftime : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `_forceframerate : Boolean`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `$version : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `Infinity : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `NaN : Number`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### [`Accessibility`](https://github.com/ruffle-rs/ruffle/issues/239)`: Object`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### [`Array`](https://github.com/ruffle-rs/ruffle/issues/240)`: Object`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
[...]

PR #15838, opened 2024-04-02
core,avm1: Add support for controlling focus highlight with `_focusrect`

kjarosh:

This patch adds support for `_focusrect` in AVM1 along with tests:
* Tests for the `_focusrect` property for SWF 5,6,7
* Tests for the `_focusrect` behavior regarding focus highlight rendering for SWF 5,6

The property `_focusrect` allows specifying whether an object should be highlighted by keyboard focus. For SWF <=5 that was possible globally, and since SWF 6 the global option only has been the default, which may be overridden using a local `_focusrect` property. Regarding the behavior of the property itself, tests were added for SWF 5,6,7 to cover all differences, which include treating the property value as bool/number or using the property in different contexts.

The field `focus_rect` is available for each interactive object, and its value is used to control highlight rendering. The notable exception is the TextField, which despite having a focus, is incapable of rendering a highlight. That is why the `focus_rect_supported` field has been added to control this behavior.


kjarosh:
Forgot to mention that this PR improves coverage for https://github.com/ruffle-rs/ruffle/issues/310, https://github.com/ruffle-rs/ruffle/issues/263, https://github.com/ruffle-rs/ruffle/issues/243

PR #15839, opened 2024-04-02
docs: Update README.md

turicfr:
* Mention the Android application of Ruffle and link to its build instructions
* Add missing folders in the sructure section
* Misc changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,13 @@ Ruffle is an Adobe Flash Player emulator written in the Rust programming languag
 * [Project status](#project-status)
 * [Using Ruffle](#using-ruffle)
 * [Building from source](#building-from-source)
+  * [Prerequisites](#prerequisites)
+  * [Linux prerequisites](#linux-prerequisites)
   * [Desktop](#desktop)
-  * [Ubuntu](#ubuntu)
-* [Homebrew](#homebrew)
+    * [Build](#build)
+    * [macOS](#macos)
   * [Web or Extension](#web-or-extension)
+  * [Android](#android)
   * [Scanner](#scanner)
   * [Exporter](#exporter)
 * [Structure](#structure)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,22 +41,24 @@ Ruffle supports ActionScript 1, 2 and 3 pretty well, but it's still not finished
...
 
-You must also have Java installed, and available on your PATH as `java`.
+* Latest stable channel of [Rust](https://www.rust-lang.org/tools/install)
+* Java, available on your PATH as `java` (required for building the library containing the builtin Flash classes for ActionScript 3)
 
 ### Desktop
 
-If you are building for a Linux platform, the following are typical dependencies:
-#### Ubuntu
+#### Linux prerequisites
```
Dinnerbone, line 59: Strictly speaking, some of these are general linux prerequisites - not specifically desktop.
I know that's not new to this PR though :D

README.md
```diff
@@ -94,29 +100,38 @@ either the web or browser extension version of Ruffle.
...
 ruffle's parsing capabilities. Provided with a folder and an output filename, it will attempt to read
-all of the flash files and report on the success of such a task.
+all of the Flash files and report on the success of such a task.
 
 `cargo run --release --package=ruffle_scanner -- folder/with/swfs/ results.csv`
 
 ### Exporter
 
-If you have a swf and would like to capture an image of it, you may use the exporter tool.
+If you have an SWF file and would like to capture an image of it, you may use the exporter tool.
```
Dinnerbone, line 117: I know this can be a source of contention, but I'm a firm believe in "**a** SWF" - as "SWF" has an official pronunciation of "swiff"

Dinnerbone (changes requested):
Teeny tiny nit but otherwise thank you so much!

README.md
sleepycatcoding: Might be worth it to add a note that these are Ubuntu packages and that the names might not match with other flavors of Linux (e.g. Fedora), just for example https://github.com/ruffle-rs/ruffle/pull/14654#issuecomment-2029760510.

turicfr: I'm not a Linux expert, but I know that at least [Arch](https://aur.archlinux.org/packages/ruffle-nightly-bin) depends on `alsa-lib`, `openssl`, `zlib`, and `libxcb`. Might worth adding these as well?

sleepycatcoding: > I'm not a Linux expert, but I know that at least [Arch](https://aur.archlinux.org/packages/ruffle-nightly-bin) depends on `alsa-lib`, `openssl`, `zlib`, and `libxcb`. Might worth adding these as well?

`libasound2` should be equivalent to `alsa-lib`, I think it might be worth it to have separate lists for different flavors.

> Also, we might also want to delete `default-jre-headless` from the list as Java is already present in the general prerequisites.

I think we can keep it in, no harm mentioning it twice, just in case the reader missed it the first time.

PR #15844, opened 2024-04-02
web: Rewrite AMO submission script to use their v5 API directly

torokati44:
Using `axios` as an HTTP client instead of the deprecated `sign-addon` package.

See: https://github.com/mozilla/sign-addon/releases/tag/6.4.0

This will resolve https://github.com/ruffle-rs/ruffle/security/dependabot/44 and https://github.com/ruffle-rs/ruffle/security/dependabot/52, by dropping the outdated `request` package.

This will also unblock https://github.com/ruffle-rs/ruffle/pull/15731.


torokati44:
Changed terminology from "sign" to "submit", because we're always getting this result anyway:

> Your add-on has been submitted for review. It passed validation but could not be automatically signed because this is a listed add-on.

Apparently, signing only occurs after manual verification. Therefore, no need/way to download a new, signed XPI right away either.

Plus, Mozilla calls their old, and to be deprecated, v4 API the "signing" API, and the new v5 API the "submission" API.

torokati44:
Thankfully Axios automatically throws an error if the status code of any response is <200 or >299, and the AMO API is "well-behaved" in this manner, so no need to do much checking of that ourselves.

kmeisthax (approved):

Dinnerbone (approved):
Let's see what happens? :D

Issue #15045, opened 2024-01-31
Cannot access a property or method of a null object reference. (accessing field: toLowerCase)


BarbourSmith:
### Describe the bug

I am excited to see the AV3 file loading is working! 

I am trying to revive a flash based 2D modeling program, but when I go to open a file I encounter the error message:

`ruffle_web-wasm_extensions-8sjk2hps.js:2 ERROR core/src/avm2/events.rs:428 Error dispatching event EventObject(EventObject { type: "complete", class: flash.events::Event, ptr: 0x4254048 }) to handler FunctionObject(FunctionObject { ptr: 0x424d120, name: Ok("com.partkart::SVGLoader/onLoadComplete()") }) : TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: toLowerCase)`

To replicate the issue click `File -> Open SVG File`

<img width="1054" alt="image" src="https://github.com/ruffle-rs/ruffle/assets/9359447/7feb3b11-6032-487c-a6d8-e579a54680db">

And select the included .svg file (or any .svg file really)

<img width="381" alt="image" src="https://github.com/ruffle-rs/ruffle/assets/9359447/090bef85-60fa-4eef-a29d-cae0d5f2c506">

### Expected behavior

I would expect the .svg file to be opened and display in the window

### Content Location

[File loading issue.zip](https://github.com/ruffle-rs/ruffle/files/14116536/File.loading.issue.zip)

### Affected platform

Online demo

### Operating system

MacOS Sonoma

### Browser

Chrome and Firefox tested

evilpie:
So this SWF has code like:
```
         XML.ignoreWhitespace = false;
         this.svg = new XML(this.fr.data);
         XML.ignoreWhitespace = true;
```

If we just ignore what `ignoreWhitespace` says and force `ignore_white` to true in `E4XNode::parse`, the parsing succeeds. Obviously that isn't really correct. My theory right now is that for some reason Flash is reparsing the XML after the initial parse?

Lord-McSweeney:
I can't check right now but is it possible that this was regressed by #13106?

Aaron1011:
Calling `Xml.copy()` actually looks at some of the XML settings:

https://github.com/adobe/avmplus/blob/858d034a3bd3a54d9b70909386435cf4aec81d21/core/E4XNode.cpp#L713-L716

evilpie:
[...]

PR #15846, opened 2024-04-02
avm2: Add IS_QNAME flag to Multiname, improve XML name matching

adrian17:
Should fix https://github.com/ruffle-rs/ruffle/issues/15045 .

The main issue fix is just the `if self.local_name().is_none()` part, the IS_QNAME part is to prevent this single line fix from messing everything else up :P And improve accuracy in general.

It's a draft as I should still set (and maybe check?) the flag in more places, even though the tests pass already.
~~Also wanna try fix XML.copy() while I'm there.~~ ok we don't have the infrastructure to easily get the class fields :c


kmeisthax (approved):
Looks good to me!

Actually, it's kinda funny how we juggled between "QNames are separate from Multinames" to "everything is a Multiname and QName is just mislabeled in AS3" and now "Multinames pretend QNames are separate from Multinames"

adrian17:
FWIW, that's because in most cases qnames are not separate and are fully interchangable 😅 
I still think our QName should be on a chopping block, at least in some degree.

torokati44:
CI fails on Ubuntu because `Test.swf` is not `test.swf`.

core/src/avm2/e4x.rs
```diff
@@ -1055,14 +1055,18 @@ impl<'gc> E4XNode<'gc> {
...
+
         if !name.is_any_name() && self.local_name() != name.local_name() {
             return false;
         }
 
-        // The Multiname is not a QName, so an any name matches everything.
-        // See https://github.com/adobe/avmplus/blob/858d034a3bd3a54d9b70909386435cf4aec81d21/core/Multiname.cpp#L59
-        if name.is_any_name() && name.namespace_set().len() > 1 {
-            return true;
+        if self.local_name().is_none() {
```
evilpie, line 1068: It would be good to add a comment that explains what this does. I think in essence we/Flash allow matching non-elements nodes like text (that don't have a name) in specific instances, i.e. not a qname?

I would also move this above the previous if.

core/src/avm2/e4x.rs
```diff
@@ -1055,14 +1055,18 @@ impl<'gc> E4XNode<'gc> {
             return false;
         }
 
+        // The Multiname is not a QName, so an any name matches everything.
```
evilpie, line 1058: Or maybe specify here .. everything (including nodes without a name like text) ?

core/src/avm2/e4x.rs
adrian17: FWIW, the logic after this PR directly line-per-line corresponds to equivalent logic in avmplus:

https://github.com/adobe/avmplus/blob/858d034a3bd3a54d9b70909386435cf4aec81d21/core/Multiname.cpp#L59-L71

So IMO unless we have a really good reason, I don't think the check order should be changed here.

The comments I can expand if needed.

core/src/avm2/globals/q_name.rs
```diff
@@ -92,7 +92,8 @@ pub fn init<'gc>(
     };
 
     if let Some(namespace) = namespace {
-        this.set_namespace(activation.context.gc_context, namespace)
+        this.set_namespace(activation.context.gc_context, namespace);
+        this.set_is_qname(activation.context.gc_context, true);
```
evilpie, line 96: Why is this conditional on the namespace?

evilpie: @adrian17 I think you answered this on Discord, can you put the comment here so we can keep track of it?

adrian17: TBH I was waiting for some ack or response there, as the string->xml-name stuff distinction between avmplus and us really confused me :/
I'll get back to it after the '+' parsing regression.

BarbourSmith:
This is fantastic work!

If I want to play with it how often does the https://ruffle.rs/demo/ page update? If I wait a little bit will it automatically appear there or should I build a local copy?

torokati44:
It updates every night (UTC), next time in about 2.5 hours. :)

Issue #15851, opened 2024-04-03
Blox Forever: Incorrect tab highlighting behavior


waspennator:
### Describe the bug

Decided to randomly test out the tab shifting behavior now that ruffle has it somewhat implemented and i'm noticing a major difference between ruffle and flash where it doesn't seem to be highlighting the menu in the same way flash does. It takes around 7 tab clicks for the orange box to show up and it doesn't line up with how flash tab highlights this game, just seems to highlight a bunch of random stuff.

### Expected behavior

Each time I press tab, it should highlight each of the menu buttons and show a visual of them being hovered over each tab click.

### Content Location

https://www.newgrounds.com/portal/view/178172

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 123

torokati44:
You might be interested, @kjarosh... ?

danielhjacobs:
> Regarding the visual, that is perhaps covered by the "It seems that focusing elements by Tab triggers some additional events?" point from https://github.com/ruffle-rs/ruffle/issues/5443#issuecomment-1992075171.

Yeah, I'm guessing the onRollOver button event is triggered, at the least.

waspennator:
Retested with latest nightly. It's better now since it's only highlighting the 7 menu buttons and hovers over each one with each tab press, but it's still not playing the animation when hovering over the buttons. Another quirk i'm noticing is that it doesn't seem to be centered like it is on flash when tab highlighting.

How it looks on Ruffle:
![Screenshot 2024-04-05 194621](https://github.com/ruffle-rs/ruffle/assets/78768499/4b0a9c98-d51a-410a-9cf6-961f6fd46e2a)

How it looks on flash:
![Screenshot 2024-04-05 194649](https://github.com/ruffle-rs/ruffle/assets/78768499/eafaab4e-d4bf-428e-ab5f-69f9f60bab49)

kjarosh:
The centering issue is caused by the lack of the hover animation, adding support for `onRollOut`/`onRollOver` will fix both.

kjarosh:
> The centering issue is caused by the lack of the hover animation, adding support for `onRollOut`/`onRollOver` will fix both.

Well I _thought_ it will fix both, but apparently it's still off. The highlight will need some more tests and fine tuning ;)

PR #15864, opened 2024-04-03
core: Ignore non-tabbable objects from tab ordering

kjarosh:
Related to https://github.com/ruffle-rs/ruffle/issues/5443, https://github.com/ruffle-rs/ruffle/issues/15851.

These objects include:
* invisible objects along with their children,
* non-editable text fields.

The method `is_tab_enabled` is renamed to `is_tabbable`, as it no longer represents the `tabEnabled` property value, but also incorporates other logic.

This PR significantly improves the tab behavior and automatic ordering.


Dinnerbone (approved):

Dinnerbone:
"tabbable" is a bit funky but I'll allow it :D

torokati44:
https://github.com/w3c/aria/issues/225#issuecomment-184717739

PR #15865, opened 2024-04-04
frontend-utils: Made new crate and moved bookmarks & all toml serialization utils to it

Dinnerbone:
This crate will eventually serve as the shared code between desktop and android (and any other future frontend), so we don't need to keep reinventing things over and over.

This PR just moves all the serialization utils to it (with cleanups), but followup PRs will add more things such as per-swf launch settings.


Dinnerbone:
(side note: the name is kinda clunky as a commit prefix... ugh)

desktop/src/preferences/write.rs
```diff
@@ -1,146 +1,172 @@
...
-    fn get_underlying_table(&mut self) -> &mut ArrayOfTables {
-        if self.0.toml_document.contains_array_of_tables("bookmark") {
-            return self.0.toml_document["bookmark"]
+            tracing::warn!("missing or invalid bookmark array, recreating..");
+            toml_document.insert("bookmark", array());
+            toml_document["bookmark"]
                 .as_array_of_tables_mut()
-                .expect("type was just verified");
+                .expect("type was just created")
         }
```
sleepycatcoding, line 98: side note:  not related to the PR but I thought about making a WriteExt and have this as a `get_or_create_...` method there.

desktop/src/preferences/write.rs
```diff
@@ -1,146 +1,172 @@
 use crate::log::FilenamePattern;
 use crate::preferences::storage::StorageBackend;
-use crate::preferences::{Bookmark, BookmarksAndDocument, PreferencesAndDocument};
+use crate::preferences::{Bookmark, SavedGlobalPreferences};
+use ruffle_frontend_utils::bookmarks::Bookmarks;
```
sleepycatcoding, line 4: very small nit: import Bookmark directly?

frontend-utils/src/parse.rs
```diff
@@ -0,0 +1,273 @@
...
+}
+
+impl<T: Default> Default for DocumentHolder<T> {
+    fn default() -> Self {
+        Self {
+            toml_document: Default::default(),
+            inner: Default::default(),
+        }
+    }
+}
```
sleepycatcoding, line 38: clever 😄

frontend-utils/src/parse.rs
```diff
@@ -0,0 +1,273 @@
...
+    pub fn values(&self) -> &T {
+        &self.result
+    }
+}
+
+#[derive(Default)]
+pub struct ParseContext {
+    pub warnings: Vec<String>,
+    /// Path of the current item being parsed
+    pub path: Vec<&'static str>,
```
sleepycatcoding, line 104: should we make this private?

desktop/src/preferences/write.rs
Dinnerbone: That'd probably look quite nice! But I'll save it for a followup, yeah? :D

Issue #15460, opened 2024-03-06
meta: Dependencies declared in multiple crates should be workspace dependencies


torokati44:
This is to reduce the churn of updating them, and the possibility of them getting accidentally out of sync, therefore dupicated (see for example: #15396).

I have a preliminary list of the most duplicated dependency crates - I can comment it here later if you think it'd be easier to get started with that, instead of discovering them on your own.

Yes, this is shameless "new-contributor-bait". :wink:

homersimpsons:
If I understand correctly the purpose of this task, the following cargo subcommand may help: https://github.com/mainmatter/cargo-autoinherit

torokati44:
Yes, sounds like it!

PR #15868, opened 2024-04-04
chore: Dedupe a whole bunch of dependency version specifications

Dinnerbone:
Anything that we specified 4 times or more...

### DIFF
--- a/core/build_playerglobal/Cargo.toml
+++ b/core/build_playerglobal/Cargo.toml
@@ -17,4 +17,4 @@ quote = "1.0.35"
 swf = { path = "../../swf" }
 clap = { workspace = true }
 regex = "1.10.4"
-walkdir = "2.5.0"
+walkdir = { workspace = true }
--- a/core/build_playerglobal/Cargo.toml
+++ b/core/build_playerglobal/Cargo.toml
@@ -15,6 +15,6 @@ convert_case = "0.6.0"
 proc-macro2 = "1.0.79"
 quote = "1.0.35"
 swf = { path = "../../swf" }
-clap = {version = "4.5.4", features = ["derive"]}
+clap = { workspace = true }
 regex = "1.10.4"
 walkdir = "2.5.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,6 +67,7 @@ log = "0.4"
 num-derive = "0.4.2"
 num-traits = "0.2.18"
 serde = "1.0.197"
+thiserror = "1.0"
 
 [workspace.lints.rust]
 # Clippy nightly often adds new/buggy lints that we want to ignore.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,7 @@ futures = "0.3.30"
 image = { version = "0.25.0", default-features = false }
 js-sys = "0.3.69"
 log = "0.4"
+num-derive = "0.4.2"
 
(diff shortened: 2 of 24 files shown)


core/build_playerglobal/Cargo.toml
```diff
@@ -15,6 +15,6 @@ convert_case = "0.6.0"
 proc-macro2 = "1.0.79"
 quote = "1.0.35"
 swf = { path = "../../swf" }
-clap = {version = "4.5.4", features = ["derive"]}
+clap = {workspace = true, features = ["derive"]}
```
sleepycatcoding, line 18: Doesn't workspace declaration already set the ´derive` feature?

Dinnerbone: I guess it does!

torokati44:
Partially addresses #15460.

Some more honorable mentions, for a future tweak PR:
```
downcast-rs
fnv
walkdir
```
and
```
byteorder
indexmap
serde_json
```

torokati44:
Referencing https://github.com/mainmatter/cargo-autoinherit/issues/16 and https://github.com/mainmatter/cargo-autoinherit/issues/2 too.

core/build_playerglobal/Cargo.toml
torokati44: This one is still there, no?

Dinnerbone: Oh shoot the lack of a space made me miss it

torokati44: Maybe the same happened with `downcast-rs`, `fnv`, and `walkdir` then? :)

torokati44:
Also I think we could enable `serde/derive` globally, like `clap/drive`.

core/build_playerglobal/Cargo.toml
torokati44: Scratch that, my mistake.

torokati44:
Even ignoring my previous comments about this, I think `walkdir` is still declared four times.

Dinnerbone:
> Also I think we could enable `serde/derive` globally, like `clap/derive`.

I kinda feel like derive shouldn't be workspace wide for either package tbh :sweat_smile: If we later get a crate that doesn't need derive, we'd have to go and undo this to allow that to build without needing derive.

But we aren't in that case today so I'll just do that for consistency

torokati44:
Well then should we consider later making `clap/derive` not workspace-wide either? 🤔 Or is that definitely needed everywhere?

torokati44 (approved):
Awesome, thanks a lot!

(The last two commits are not in alphabetical order!! 😖🫠😵‍💫🫨😵😫😅😆😁😄)

Dinnerbone:
> Well then should we consider later making `clap/derive` not workspace-wide either? 🤔 Or is that definitely needed everywhere?

It's needed everywhere right now. It's less likely that we'll use clap without derive, than serde without derive - so I'm happy for it to stay this way right now

sleepycatcoding (approved):

sleepycatcoding:
The `walkdir` change is gonna create a local conflict on my save manager branch, but it is whatever. (since I also pulled it out into the workspace)

torokati44:
I suspect it's gonna cause a few conflicts, if anything, due to neighboring line changes... But they should be easy to sort out.

PR #15874, opened 2024-04-04
chore: Only do direct dependencies with dependabot

Dinnerbone:
This will hopefully make it not timeout

Kinda sucks but... oh well

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,6 @@
 version: 2
 updates:
   - package-ecosystem: "npm"
-    allow:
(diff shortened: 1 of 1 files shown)


Dinnerbone:
It'll update the lockfile for direct dependencies and that may happen to accidentally touch some indirect ones too, but it won't find if - for example - an indirect dependency gets an update but it's only referred to by one of our direct dependencies which *doesn't* get an update.

Dinnerbone:
(fwiw I think this is what renovate was doing too)

PR #15882, opened 2024-04-04
ci: run extra clippy step without tests

sleepycatcoding:
Currently to check if it would have caught the unused import on master.


sleepycatcoding:
Only adds about 2-3 seconds of extra runtime to the lint jobs, so it wasn't expensive after all!

Issue #15620, opened 2024-03-18
HitTest function does not properly translate clip coordinates relative to certain roots


nicko1337:
### Describe the bug

The `MovieClip` `hitTest` function does not properly translate its own movie clip's coordinates when all of the following conditions are met:

- The `_root` movie clip's `_lockroot` property is `true`.
- The `_root` movie clip's `_x` and/or `_y` value is not `0`.
- A `hitTest` function belonging to a descendant movie clip of `_root` is called, with three parameters provided: a global X coordinate, a global Y coordinate, and a shape flag.

Let's say the following structure is present:

- `_level0:MovieClip`: `_width` = 500, `_height` = 300
  - `child:MovieClip`: `_lockroot` = `true`, `_x` = 200, `_y` = 100, `_width` = 100, `_height` = 100
    - `bigBox:MovieClip`: `_x` = 0, `_y` = 0, `_width` = 100, `_height` = 100, completely filled in
    - `smallBox:MovieClip`: `_x` = 30, `_y` = 30, `_width` = 40, `_height` = 40, completely filled in
      - `coords:Object`: `x` and `y` values set to an initial value, then overwritten by `smallBox.localToGlobal(smallBox)`

Under these circumstances, `_level0.child.bigBox.hitTest(_level0.child.smallBox.coords.x, _level0.child.smallBox.coords.y, true)` returns `false`. The `shapeFlag` parameter can be set to `false` as well; the parameter doesn't really matter here. However, the small box (colored blue in this image) is clearly inside the larger box (colored red):

![ruffle_KoXnK32IVT](https://github.com/ruffle-rs/ruffle/assets/22125088/5f883a79-959d-4903-96b7-b046c83d7501)

From what I can tell, this occurs because `_level0.child.bigBox.hitTest` translates its movie clip's coordinates relative to the locked `_root` movie clip `_level0.child`, despite the coordinates provided by `_level0.child.smallBox.localToGlobal` being relative to `_level0`.

### Expected behavior

`_level0.child.bigBox.hitTest(_level0.child.smallBox.coords.x, _level0.child.smallBox.coords.y, true)` should return `true` in the scenario detailed above.

### Content Location

Here's an example ZIP that you can use to test the bug yourself. To try it out, simply load the `parent.swf` file into the Ruffle player. You can provide these parameters:

- `doNotLockRoot`: If truthy (e.g., `1`), `_level0.child._lockroot` will be `false`; it is `true` by default.
[...]

PR #15894, opened 2024-04-06
Use correct coordinates for hitTest

nicko1337:
Fix for issue #15620


Lord-McSweeney:
Thank you for the contribution! Can you add a test, as specified by https://github.com/ruffle-rs/ruffle/blob/master/tests/README.md?

Dinnerbone (approved):
Thank you very much!

Lord-McSweeney:
Oh, I stumbled across some Discord messages years ago, and there's apparently some weirdness with this:

mike:
> here's a fun Flash Player weirdness I just stumbled across...
> 
```
var clip = createEmptyMovieClip("clip", 0);

clip.beginFill(0xff0000);
clip.moveTo(200, 200);
clip.lineTo(400, 200);
clip.lineTo(400, 400);
clip.lineTo(200, 400);
clip.lineTo(200, 200);
clip.endFill();

trace(clip.hitTest(300, 300, true)); // true
_root._xscale = 50;
_root._x = 300;

trace(clip.hitTest(300, 300, true)); // false!?

function onEnterFrame() {
    trace(clip.hitTest(300, 300, true)); // turns to true a frame or two later?
}
```
> It should always print true, but after transforming root it prints false for some amount of time (actually varies based on the swf framerate!)


> Some kind of bug where it's not flushing the drawing API commands, or something?


> (If you just place some square art instead of using the drawing API, it works as expected -- hitTest is in root coords so transforming root should have no effect).


Might it be related to this @nicko1337 ?

nicko1337:
That's definitely weird, but I don't think it's related to this bugfix. I noticed the \_lockroot hitTest bug in Poptropica's AS2 islands, which do not use the drawing API or \_root scaling, but do use \_lockroot and non-zero coordinates for the locked \_root. The bug is now fixed in that game.

As for why I used the drawing API in the original bug report and this PR's test case, that's just because it was an easy way to reproduce the bug as I don't have access to Adobe Flash CS6 to draw anything (and Poptropica's AS2 content is not officially available anymore).

Unless this PR affects how Ruffle handles the bug you described (which unfortunately I can't check right now), I'm guessing that these are separate issues.

PR #15910, opened 2024-04-07
avm1: Fix `onKeyUp` and `onKeyDown` events for Button

kjarosh:
These two events were not firing properly when a key was pressed and the button had focus.


kjarosh:
It seems that it's not so easy, `KeyUp` & `KeyDown` are dispatched differently than others...

Dinnerbone (approved):
<3

PR #15949, opened 2024-04-10
frontend-utils: Move a bunch of backends from Desktop

Dinnerbone:
Most of this has been duplicated over on Android, but then diverged as each codebase went its own way. This lets us bring it all together again, and makes maintaining things so much easier.


sleepycatcoding (approved):
Looks good to me.

desktop/src/executor.rs
```diff
@@ -129,7 +132,7 @@ impl TaskHandle {
     );
 }
 
-pub struct WinitAsyncExecutor {
+pub struct WinitAsyncExecutor<R: PollRequester> {
```
Lord-McSweeney: Should this still be called `WinitAsyncExecutor`?

sleepycatcoding: It is renamed in a later commit: https://github.com/ruffle-rs/ruffle/pull/15949/commits/3ace44006b1fba5114bba0971848a7db98fd1c08

frontend-utils/src/backends/navigator.rs
```diff
@@ -0,0 +1,890 @@
+//! Navigator backend for web
```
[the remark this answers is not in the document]
Dinnerbone: Yes :D I'll remove it though

PR #15320, opened 2024-02-25
ci: Add a dedicated Cargo profile, test with that

torokati44:
This is a continuation of #15241, which I somehow messed up (the source branch was in this repo, not in my fork - oops, sorry, branch deleted).

PR #15950, opened 2024-04-10
tests: Schedule slow tests at the beginning

kjarosh:
By setting `size = "large"`, a test may be marked as large, which means that the duration of its execution is exceptionally long compared to other tests.

In order to minimize the duration of `cargo test`, slow tests should be executed at the beginning. This maximizes load on multithreaded CPUs by fully utilizing all threads and preventing late slow tests from stalling the whole suite.

Note: the following values are outdated.

| Test Command                  | Before  | After         | After with `fast` |
|-------------------------------|---------|---------------|-------------------|
| `cargo test`                  | ~~106.74s~~ | ~~77.45s **(-27%)**~~ | ~~43.55s **(-59%)**~~ |
| `cargo test -F imgtests`      | ~~115.94s~~ | ~~80.55s **(-31%)**~~ | ~~52.22s **(-55%)**~~ |
| `cargo nextest run`                  | ~~118.382s~~ | ~~80.872s **(-32%)**~~ | ~~49.526s **(-58%)**~~ |
| `cargo nextest run -F imgtests`      | ~~132.738s~~ | ~~91.500s **(-31%)**~~ | ~~69.857s **(-47%)**~~ |
| `cargo nextest run --cargo-profile ci` | ~~23.916s~~ | ~~13.178s **(-45%)**~~ | ~~10.070s **(-58%)**~~ |
| CI ubuntu      | ~~118.932s~~ | ~~82.319s **(-31%)**~~ | N/A |
| CI macos       | ~~69.940s~~  | ~~29.981s **(-57%)**~~ | N/A |
| CI windows     | ~~310.591s~~ | ~~217.358s **(-30%)**~~ | N/A |

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -25,6 +25,10 @@ known_failure = false # If true, this test is known to fail and the result will 
 output_path = "output.txt" # Path (relative to the directory containing test.toml) to the expected output
 log_fetch = false # If true, all network requests will be included in the output.
 
+# Used to mark larger tests as `size = "large"`. Larger tests will be executed
+# at the beginning to improve test execution time.
+size = "regular"
+
 # Sometimes floating point math doesn't exactly 100% match between flash and rust.
 # If you encounter this in a test, the following section will change the output testing from "exact" to "approximate"
 # (when it comes to floating point numbers, at least.)
--- a/tests/framework/src/test.rs
+++ b/tests/framework/src/test.rs
@@ -1,7 +1,7 @@
-use crate::environment::Environment;
 use crate::options::TestOptions;
 use crate::runner::TestRunner;
 use crate::util::read_bytes;
+use crate::{environment::Environment, options::TestSize};
 use anyhow::{anyhow, Result};
 use ruffle_core::tag_utils::SwfMovie;
 use ruffle_input_format::InputInjector;
@@ -13,6 +13,23 @@ pub struct Font {
     pub family: String,
     pub bold: bool,
     pub italic: bool,
+}
+
+#[derive(Clone, Copy)]
+pub enum TestKind {
+    Large,
+    RealTime,
+    Render,
+}
+
+impl TestKind {
+    pub fn name(self) -> &'static str {
+        match self {
+            TestKind::Large => "large",
+            TestKind::RealTime => "realtime",
+            TestKind::Render => "render",
+        }
+    }
 }
 
 pub struct Test {
@@ -41,6 +58,24 @@ impl Test {
             root_path: test_dir,
             name,
         })
+    }
+
+    pub fn kind(&self) -> Option<TestKind> {
+        // Real time overrides everything
+        if self.options.sleep_to_meet_frame_rate {
+            return Some(TestKind::RealTime);
+        }
+
+        match self.options.size {
+            TestSize::Regular => (),
+            TestSize::Large => return Some(TestKind::Large),
+        }
+
+        if self.options.player_options.has_renderer() {
+            return Some(TestKind::Render);
+        }
+
+        None
     }
 
     pub fn create_test_runner(&self, environment: &impl Environment) -> Result<TestRunner> {
--- a/tests/framework/src/test.rs
+++ b/tests/framework/src/test.rs
@@ -30,6 +30,21 @@ impl TestKind {
             TestKind::Render => "render",
         }
     }
+
+    pub fn priority(kind: &str) -> isize {
+        // NOTE: Make sure to synchronize this with the nextest config
+        match kind {
+            "realtime" => 3,
+            "large" => 2,
+            // Render tests are slower on Windows and sometimes on nextest
+            "render" => 1,
+            _ => 0,
+        }
+    }
+
+    pub fn ord(kind: &str) -> impl Ord {
+        -Self::priority(kind)
+    }
 }
 
 pub struct Test {
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -7,3 +7,17 @@ test(avm2/graphics_round_rects) or \
 test(avm1/netstream_play_flv) or \
 test(avm1/netstream_play_flv_screen)"""
 retries = 4
+
+# Run tests that take more time first
+# NOTE: Make sure to synchronize this with TestKind::priority
+[[profile.default.overrides]]
+filter = "test([realtime] )"
+priority = 3
+
+[[profile.default.overrides]]
+filter = "test([large] )"
+priority = 2
+
+[[profile.default.overrides]]
+filter = "test([render] )"
+priority = 1
--- a/tests/Cargo.toml
+++ b/tests/Cargo.toml
@@ -21,6 +21,8 @@ imgtests = [
 ]
 jpegxr = ["ruffle_test_framework/jpegxr"]
 lzma = ["ruffle_test_framework/lzma"]
+# Log how long each test took
+times = []
 
 [dependencies]
 ruffle_render_wgpu = { path = "../render/wgpu", optional = true }
--- a/tests/framework/src/options.rs
+++ b/tests/framework/src/options.rs
@@ -17,6 +17,14 @@ use std::time::Duration;
 use vfs::VfsPath;
 
 #[derive(Clone, Deserialize)]
+pub enum TestSize {
+    #[serde(rename = "regular")]
+    Regular,
+    #[serde(rename = "large")]
+    Large,
+}
+
+#[derive(Clone, Deserialize)]
 #[serde(default, deny_unknown_fields)]
 pub struct TestOptions {
     pub num_frames: Option<u32>,
@@ -32,6 +40,7 @@ pub struct TestOptions {
     pub log_fetch: bool,
     pub required_features: RequiredFeatures,
     pub fonts: HashMap<String, FontOptions>,
+    pub size: TestSize,
 }
 
 impl Default for TestOptions {
@@ -50,6 +59,7 @@ impl Default for TestOptions {
             log_fetch: false,
             required_features: RequiredFeatures::default(),
             fonts: Default::default(),
+            size: TestSize::Regular,
         }
     }
 }
@@ -226,6 +236,10 @@ impl PlayerOptions {
                 height: movie.height().to_pixels() as u32,
                 scale_factor: 1.0,
             })
+    }
+
+    pub fn has_renderer(&self) -> bool {
+        self.with_renderer.is_some()
     }
 
     pub fn create_renderer(
--- a/tests/fs-tests-runner/src/lib.rs
+++ b/tests/fs-tests-runner/src/lib.rs
@@ -61,6 +61,7 @@ pub struct TestLoaderParams<'a> {
 }
 
 pub type TestLoader = Box<dyn Fn(TestLoaderParams) -> Option<Trial>>;
+pub type TestComparator = Box<dyn FnMut(&Trial, &Trial) -> std::cmp::Ordering>;
 
 pub struct FsTestsRunner {
     root_dir: PathBuf,
@@ -68,6 +69,7 @@ pub struct FsTestsRunner {
     additional_tests: Vec<Trial>,
     test_loader: Option<TestLoader>,
     canonicalize_paths: bool,
+    comparator: TestComparator,
 }
 
 impl Default for FsTestsRunner {
@@ -84,6 +86,7 @@ impl FsTestsRunner {
             additional_tests: Vec::new(),
             test_loader: None,
             canonicalize_paths: false,
+            comparator: Box::new(|a, b| a.name().cmp(b.name())),
         }
     }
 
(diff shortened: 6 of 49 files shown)


Dinnerbone:
(For what it's worth, I highly recommend using nextest rather than test for local development!)

kjarosh:
Right! Did not think about using nextest. :smile:  Certainly has a better interface, but unfortunately tests are executing a bit longer :(

I added nextest benchmarks to the table in the description. At first glance, performance improvements are similar to `cargo test`. However, it seems that nextest does not execute tests marked as `slow` which is weird... :thinking: I'm going to investigate it tomorrow, and mark this PR as a draft for now.

torokati44:
You could try the profile dedicated to CI - it's a bit more optimized:
https://github.com/ruffle-rs/ruffle/pull/15320/files

torokati44:
Although, I do like the idea of somehow trying to sort the test cases in decreasing real time duration. But I wish it was somehow ✨automatical✨, with no need to manually annotate anything. That sort (pun may or may not be intended) of thing is bound to get outdated.
I'm on the fence about the `fast` feature... It feels like not the right tool to achieve this (tests can be filtered already anyway, at least by name, after all), but what do I know, maybe it is.

kjarosh:
This PR is just a suggestion, I tried speeding tests up somehow and noticed that at the end my CPU was sitting idle. Regarding the `fast` feature, I'm not sure about it either. It executes 98.86% of tests taking only 60% of the time. I'm not sure how useful it is, but decided to include it anyways for you to jugde.

I also thought about automating test annotation somehow. The idea of no annotations seems eerily close to the halting problem though ;) I was thinking about how to annotate the tests in order to schedule them properly. It could be the slow flag which divides tests into two categories, but it also could be a nondiscrete value (e.g. estimated execution time) which constitutes a total order. The latter seems to be an overkill, because (1) these values would be more dependent on the system, and (2) time improvement would be minimal compared to the former, which is a much simpler approach.

Regarding automation, I thought about giving a feedback to the user about slow tests which are fast and non-slow tests which are slow. That would require two time values in order to create hysteresis. They could be set manually (e.g. 1s,10s), or calculated at the end based on the probability distribution (e.g. σ,2σ). That of course is a bit complicated so I did not go in this direction as I wasn't sure whether we would want this approach at all. That also sounds like a problem that possibly could be solved at a toolchain/library level.

torokati44:
> This PR is just a suggestion,

Of course, and thanks! :)

> Regarding automation, I thought about giving a feedback to the user about slow tests which are fast and non-slow tests which are slow. 

`nextest` already warns about `[SLOW]` tests by itself, but it doesn't have the `is-slow` annotation of course, which would make this more or less acceptable for a given test case.

> The idea of no annotations seems eerily close to the halting problem though ;) I was thinking about how to annotate the tests in order to schedule them properly. It could be the slow flag which divides tests into two categories, but it also could be a nondiscrete value (e.g. estimated execution time)

Bah, it could also be just a little system-local cache of "the last 5-10 times this test case was ran, it took xx milliseconds". Then to schedule, look up these records, average per case, and sort. The averaging should even out any discrepancies caused by P-core vs. E-core, powersave/performance, dynamic clocks, overall system load, thermal throttling, etc... And how a given test case is identified isn't that crucial either, as long as it works "reasonably enough", since it only determines the order of test executions, which shouldn't affect anything :tm:, other than load balancing at the end, of course.

This could even be a generic `cargo-test` or `cargo-nextest` feature, maybe even enabled by default?

torokati44:
> This could even be a generic `cargo-test` or `cargo-nextest` feature, maybe even enabled by default?

Ahh, see: https://github.com/nextest-rs/nextest/issues/905
Also somewhat related: https://github.com/nextest-rs/nextest/issues/742

kjarosh:
Okay, I think I found out why nextest does not run tests marked as slow; I also found out why tests on nextest are slower compared to `cargo test`.

Turns out `libtest_mimic` adds the test kind as a prefix before its name (e.g. `[slow] test`). Nextest executes tests one by one matching them by name (with the prefix), but the filtering is done manually by our code (`is_candidate`), which does not take into the account the kind prefix as libtest-mimic does (see `TestInfo::test_name_with_kind` from libtest-mimic). That _silently_ ignores any test with a kind, as they are not executed but marked as passed.

The strategy of executing tests by nextest also explains why it's significantly slower than `cargo test`. Regression tests always perform a full walk over `tests/swfs` regardless of whether a test is matched exactly by its name or not. This full walk is performed of course for each test.

Removing the full walk on exact match makes nextest about 15s faster than before, which is closer to being as fast as `cargo test`.

kjarosh:
Created a new PR with these two commits: https://github.com/ruffle-rs/ruffle/pull/16031

adrian17:
I know I'm a bit late to the thread but

> Although, I do like the idea of somehow trying to sort the test cases in decreasing real time duration.

I'm not sure why I'd ever want that tbh. And I haven't really seen other projects do this kind of thing before either?
(If we wanted to sort (which I personally don't), if anything, I'd go for increasing time duration; if there's a failing test, this will minimize average time to it being reported.)

torokati44:
> I'm not sure why I'd ever want that tbh.

Because it minimizes the chance that a many-core CPU grinds through most tests in X seconds, and one or two really long tests happened to get started at  let's say X*0.8, extending the overall time to wait by another large fraction of X, while most cores are already idle. If the short and quick tests are left to the end, they can maximally utilize all of the cores all the way until all tests are done, therefore being quicker overall. It's a classic "packing problem" kind of thing...

kjarosh:
> if anything, I'd go for increasing time duration

That is better when we expect tests to fail instead of pass, but generally in CI it's way more probable for the tests to pass. When running locally we might have different expectations, although I usually have a positive attitude and expect them to pass ;)

kjarosh:
Refreshed this PR a bit to see how it performs as of today. I've removed the `fast` feature and did some tests (average of 3 retries):


| Test Command                  | Before  | After         | Difference |
|-------------------------------|---------|---------------|---------------|
| `cargo test`                  | 119.06s | 84.73s        | **−34.33 (-29%)** |
| `cargo nextest run`           | 133.123s | 97.44s       | **−35.68 (-27%)** |

torokati44:
Just a random idea on the concern of "manual annotations getting put of sync with reality over time": Is it feasible to at least output a little note at the end of a test "campaign" like this?:
"BTW test3428 is annotated as slow but was actually fsster than average (with some hysteresis margin), and test2638 and test2627 vica versa - you might want to check that."

Or is this perhaps also blocked on these:?
https://github.com/nextest-rs/nextest/issues/20
https://github.com/nextest-rs/nextest/issues/1152

kjarosh:
I think that would be easy to implement in `cargo test`, but I just don't see how I would do that in nextest: our code is executed only when tests are listed, and then for each test separately. We would have to persist the data for each test, detect when the last test is being executed, and somehow convey the message to the user. Of course unless we use some nextest-specific hooks or something (which I'm not knowledgeable enough to say anything about that).

I would argue that this "is_slow" annotation is only a heuristic: having outdated annotations also improves test times (not as much as having current annotations, but still). The outdating issue also mainly refers to new tests, i.e. tests not marked as slow when they should be, as it's far less likely for a slow test to become fast out of the blue.

If having a warning about wrongly marked tests after `cargo test` is okay, I could add that.

torokati44:
All fair points, and this isn't worth much effort even in my opinion - only if it was trivial, which it isn't then.

kjarosh:
It isn't trivial for nextest, but for `cargo test` I even remember having a PoC of something like that. But if we aren't planning on merging it anyway, taking the time to implement that does not make sense :smile:

tests/framework/src/test.rs
```diff
@@ -23,6 +43,7 @@ pub struct Test {
     pub output_path: VfsPath,
     pub root_path: VfsPath,
     pub name: String,
+    pub kind: Option<TestKind>,
```
torokati44, line 46: Why not `is_slow: bool` here?

tests/README.md
```diff
@@ -71,4 +72,4 @@ sleep_to_meet_frame_rate = false # If true, slow the tick rate to match the movi
 
 `fscommand("quit")` is enabled for tests, and will end the test at the end of this frame or tick.
 
-You can use this to end a test prematurely before the set number of iterations elapses, which may be useful for timer tests.
\ No newline at end of file
+You can use this to end a test prematurely before the set number of iterations elapses, which may be useful for timer tests.
```
torokati44, line 75: Huh, what is this change? A line ending?

tests/framework/src/test.rs
kjarosh: In fact, I can remove it altogether :)

tests/README.md
kjarosh: This change adds a missing newline at the end of file, but GitHub somehow cannot show it properly :thinking:

tests/framework/src/test.rs
```diff
@@ -15,6 +15,26 @@ pub struct Font {
     pub italic: bool,
 }
 
+#[derive(Clone, Copy)]
+pub enum TestKind {
+    Slow,
```
torokati44, line 20: What I meant before is, why is this "kind" type even needed?
Do you have any other "kinds" of tests you plan to add in the near future, that are mutually exclusive with them being "slow"? If not, I feel like this is just premature over engineering - YAGNI.
Or is this somehow a technical requirement to make test ordering with this attribute possible/easier?

kjarosh: Oh sure, using test kinds was actually the easiest way. It's not like "I created test kinds to possibly support more than one kind", but rather "libtest_mimic supports test kinds and I can utilize that for slow tests and test ordering". I think I could create a wrapper for a test trial that contains the information about the test ordering, but wouldn't _that_ be over engineering?

Compare that approach also to other frameworks, for instance:
* JUnit 4 test categories: https://github.com/junit-team/junit4/wiki/Categories (the example mentions `SlowTests.class` category),
* JUnit 5 tags: https://junit.org/junit5/docs/current/user-guide/#writing-tests-tagging-and-filtering (the example mentions a `fast` tag),
* pytest markers: https://docs.pytest.org/en/7.1.x/example/markers.html (docs mention a `slow` test mark).

The big disadvantage of libtest_mimic test kinds is that a test cannot have more than one kind, but it is possible to store multiple kinds by joining them e.g. with a comma (`kind1,kind2`).

Using test kinds also means that tests with a kind are shown as such in the output (i.e. slow tests are prefixed with `[slow]`).

If you're referring to the enum and not test kinds in general, well... that's only a high-level representation of a kind. It's better IMO than comparing strings, and it also encapsulates ordering logic. If you have any suggestions on improving it, I'm all ears!

torokati44: > but rather "libtest_mimic supports test kinds and I can utilize that for slow tests and test ordering".

Ooh, right, pardon my ignorance... It's been a while since we discussed this... ^^'

torokati44:
> However, it seems that nextest does not execute tests marked as slow which is weird... 🤔 I'm going to investigate it tomorrow, and mark this PR as a draft for now.

Is this still the case? If not, why is this still a draft?

kjarosh:
Updated slow tests, removed draft status. All nextest issues should be fixed already

torokati44:
Maybe just randomizing them also achieves something like that? :no_mouth: 
https://github.com/torokati44/ruffle/actions/runs/10334744158

Or, we could randomize first, and then sort the slow ones first... Just to be extra sure...?

kjarosh:
As @torokati44 noticed... this PR works totally by accident, see:
* https://github.com/ruffle-rs/ruffle/pull/17961#issuecomment-2356003955
* https://github.com/nextest-rs/nextest/issues/905#issuecomment-2355975586
* https://github.com/nextest-rs/nextest/blob/a4490cdb039c28b321fef6b43a1b71a0af517999/nextest-runner/src/list/test_list.rs#L635

torokati44:
cargo-nextest 0.9.91, released on 2025-02-13, now supports test priorities: https://nexte.st/docs/configuration/test-priorities/

kjarosh:
After rebasing and using the nextest priority feature, I don't see any significant time gains as before :/

It's probably because today's lexicographical order results in better ordering than before. The question is whether we want to merge it anyway or not.

kjarosh:
I guess those test kinds might be useful for test filtering (to skip large tests on Windows ARM, etc.).

torokati44:
Why the change from `slow` to `size`? IMHO it's not at all obvious, when looking at this option in isolation, what it refers to (disk usage, screen resolution, memory usage, number of checks done, etc...?).

kjarosh:
By doing experiments I noticed that ordering tests is not that simple. The main issue we have in this repo is that we use nextest which executes tests differently and results in vastly different test runtimes. Sometimes a test that takes 0.1s in `test` can take over 4s in `nextest`, sometimes it will take the same amount. I've  noticed that mostly render tests do that, that's why I created a separate kind for them. But I didn't feel comfortable leaving a parameter that we couldn't use in the future in case we need to include a third ordering option (size=medium?) for tests that are either slow or fast depending on the tool or in case we have _a lot_ of tests and we want to filter them more granularly.

I think the term "test size" is okay, it's also used in the Android ecosystem: https://developer.android.com/reference/androidx/test/filters/MediumTest, https://developer.android.com/reference/androidx/test/filters/LargeTest

PR #15959, opened 2024-04-12
extension: Remove Ruffle's own script tag onload

### LINKED ISSUE #15862: Extension should clean up after itself in the DOM
This is admittedly an esoteric request, but if possible could the Ruffle Addon remove the `<script>` elements it adds to the DOM after those scripts have ran?

The reason I request this is because I often use Ctrl+S to save web pages as "Web Page, Complete", which downloads to disk a snapshot of the page's DOM along with all linked resources in use. Including the two ruffle scripts (the inline one and the ruffle.js one). They take up a lot of disk space, and would probably interfere with Ruffle playback via the addon should any of the pages I download ever include flash content.

### Alternatives I have considered

- running a shell script to delete the downloaded ruffle.js and strip out the script tags: I did this in fact, I wrote a sed script to do this and it's awful and i hate it, and it sometimes fails to work (i think b/c my script assumes the inline script always precedes the src'ed one but this isn't always true? not sure)
- writing a userscript to strip out the script elements before saving. I haven't tried this yet, but it would probably work... but I'd prefer if the ruffle addon cleaned up after itself especially as that would benefit other people using the addon who may not be as technologically savvy enough to write or run a userscript

---

And btw, thank you all so much for creating Ruffle and keeping the dream of Flash alive. *Viva La Flash!*

danielhjacobs:
This is almost identical to "Method 1: Inject another file" from https://stackoverflow.com/questions/9515704/access-variables-and-functions-defined-in-page-context-from-an-extension/9517879#9517879, but uses `addEventListener` instead of `onload`.
When merged alongside #15958, fixes #15862.

### DIFF
--- a/web/packages/extension/src/content.ts
+++ b/web/packages/extension/src/content.ts
@@ -63,7 +63,10 @@ function injectScriptRaw(src: string) {
 function injectScriptURL(url: string): Promise<void> {
     const script = document.createElement("script");
     const promise = new Promise<void>((resolve, reject) => {
-        script.addEventListener("load", () => resolve());
+        script.addEventListener("load", function () {
+            resolve();
+            this.remove();
+        });
         script.addEventListener("error", (e) => reject(e));
     });
     script.charset = "utf-8";


web/packages/extension/src/content.ts
```diff
@@ -67,6 +67,9 @@ function injectScriptURL(url: string): Promise<void> {
     });
     script.charset = "utf-8";
     script.src = url;
+    script.onload = function () {
```
evilpie, line 70: We already add an event listener for "load" 5 lines above, can't we use that?

evilpie (approved):

PR #15966, opened 2024-04-12
desktop: Do not resize the window while it is maximized

### LINKED ISSUE #11364: Sometimes getting white screen after 'close' and opening a game
### Describe the bug

using 'close' on the Ruffle GUI and opening another or the same game, can sometimes cause a white screen instead of the game loading, even with working SWFs 
no game will ever load when it does this, so closing ruffle itself is the only way to fix it
It happened 3 times in a row, but when recording, it didn't happen, so not sure what cause is
It always seemed to happen if I press 'force' on advanced open for the shown all before opening the sef, not sure

also, when 'browse' for a swf to use, the file browser does not pop up after the first time

### Expected behavior

after closing a game, and opening a new one there should not be a white screen

### Affected platform

Desktop app

### Operating system

Linux Mint

### Browser

_No response_

### Additional information

_No response_

sleepycatcoding:
Fixes #11364.

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -408,78 +408,92 @@ impl App {
                 }
                 winit::event::Event::UserEvent(RuffleEvent::TaskPoll) => self.player.poll(),
                 winit::event::Event::UserEvent(RuffleEvent::OnMetadata(swf_header)) => {
-                    let movie_width = swf_header.stage_size().width().to_pixels();
-                    let movie_height = swf_header.stage_size().height().to_pixels();
                     let height_offset = if self.window.fullscreen().is_some() || self.no_gui {
                         0.0
                     } else {
                         MENU_HEIGHT as f64
                     };
 
-                    let window_size: Size = match (self.preferred_width, self.preferred_height) {
-                        (None, None) => {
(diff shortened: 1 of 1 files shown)


n0samu:
Testing the original issue now, I notice that on Windows, loading an SWF into a maximized window does result in a resize, but the window doesn't realize that it is now unmaximized, so it takes two clicks to maximize the window again:

https://github.com/ruffle-rs/ruffle/assets/71368227/d0069dae-a3ae-4c7e-931c-fb6183a7a025

I guess we should look into fixing this behavior, maybe in a followup. (I can open a new issue if this doesn't seem easy to solve)

sleepycatcoding:
Maybe we shouldn't try to resize the window at all when it is maximized?

n0samu:
Sounds good to me!

sleepycatcoding:
@n0samu could you test the changes on Windows to see if it fixes the window desync. (I don't have a Windows environment to test myself)

n0samu:
Just tested and it's working well now, thanks! 👍

Croworbit:
is this ready?

Lord-McSweeney (approved):

kjarosh (approved):

PR #15974, opened 2024-04-12
frontend-utils: Remove isahc in favour of reqwest

Dinnerbone:
isahc has unfortunately been seemingly unmaintained for 2 years now.

This brings in tokio in desktop - I don't really see that as a bad thing though. Future PRs may switch more things to tokio instead of our hand rolled versions of asyncness.

We also switch to using rustls instead of openssl.


torokati44:
On macOS:

```
        FAIL [   0.027s] ruffle_frontend_utils backends::navigator::tests::test_socket_basic_communication

--- STDOUT:              ruffle_frontend_utils backends::navigator::tests::test_socket_basic_communication ---

running 1 test
test backends::navigator::tests::test_socket_basic_communication ... FAILED

failures:

failures:
    backends::navigator::tests::test_socket_basic_communication

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 43 filtered out; finished in 0.02s


--- STDERR:              ruffle_frontend_utils backends::navigator::tests::test_socket_basic_communication ---
thread 'backends::navigator::tests::test_socket_basic_communication' panicked at frontend-utils/src/backends/navigator.rs:708:9:
assertion `left == right` failed
  left: Data(SocketHandle(4294967295v1), [72, 101, 108, 108, 111, 32])
 right: Data(SocketHandle(4294967295v1), [72, 101, 108, 108, 111, 32, 87, 111, 114, 108, 100, 33])
```

Strange...

Dinnerbone:
Huh. I'm not sure I touched anything relevant there?

Dinnerbone:
@kjarosh any immediate thoughts why that test would start failing? I think you made it, so before I dive into things I figure you might have an insight :D

kjarosh:
Huh... As I read the code now, it might actually be flaky due to the async call in `write_server`:

https://github.com/ruffle-rs/ruffle/blob/1bcc4d530a2440eca86f1d64db9106be8e8047fd/frontend-utils/src/backends/navigator.rs#L729-L735

It is an async call site and produces a potential switch point which may wake up the reader task and cause the two messages to be read independently (the reader is being executed concurrently in the background).

https://github.com/ruffle-rs/ruffle/blob/1bcc4d530a2440eca86f1d64db9106be8e8047fd/frontend-utils/src/backends/navigator.rs#L848-L854

I would suggest to join these two messages into one; it is not a logic in our code anyways to buffer them. (Although we _could_ be smarter here and not read eagerly.)

https://github.com/ruffle-rs/ruffle/blob/1bcc4d530a2440eca86f1d64db9106be8e8047fd/frontend-utils/src/backends/navigator.rs#L522

Aaron1011 (approved):

sleepycatcoding (approved):

PR #15981, opened 2024-04-14
frontend-utils: Get rid of deadlocks in AsyncExecutor

kjarosh:
Related to #15045: fixed a deadlock on desktop during opening a file. Also should fix deadlocks on Android.

3d04971a was the first approach to eliminating deadlocks caused by waking a task from another one. A task may also try to wake itself, and that case was unfortunately not covered.

This patch rewrites the `AsyncExecutor` to use an async channel for queuing tasks instead of a flag. Using a flag required holding a lock for a task which was problematic when a task needed to be awoken. A flag also required doing a full search through tasks each poll, which could be ineffective.

Changes from this PR should be tested thoroughly before merging. Things to watch out for:
1. Performance issues, although in theory this code should be faster, there could be some edge cases I did not think of.
2. Memory usage, the async channel is unbounded and tasks may queue duplicated entries. I do not think it should be a problem though, maybe some long running operations or an infinite loop?

cc @Dinnerbone

### DIFF
--- a/frontend-utils/src/backends/executor.rs
+++ b/frontend-utils/src/backends/executor.rs
@@ -7,7 +7,7 @@ use ruffle_core::loader::Error;
 use slotmap::{new_key_type, SlotMap};
 use std::future::Future;
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::{Arc, Mutex, RwLock, RwLockReadGuard, RwLockWriteGuard, Weak};
+use std::sync::{Arc, Mutex, MutexGuard, Weak};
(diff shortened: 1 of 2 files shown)


Dinnerbone (approved):

torokati44:
And now this has a merge conflict after https://github.com/ruffle-rs/ruffle/pull/16029. :no_mouth:

kjarosh:
Rebased, nothing changed, as this PR has removed the "Double polling" warning already (maybe we could have just merged this instead? :thinking: )

PR #15986, opened 2024-04-15
avm2: Store Class instead of QName as superclass of Class

Lord-McSweeney:
This is the first step necessary for the ClassObject refactor- eagerly looking up the superclass when loading a `Class` from the ABC.


core/src/avm2/globals.rs
```diff
@@ -378,7 +378,7 @@ fn class<'gc>(
     let class_read = class_def.read();
     let super_class = if let Some(sc_name) = class_read.super_class_name() {
```
adrian17, line 379: how about replacing this entire lookup by
```rs
     let super_class = if let Some(sc) = class_read.super_class() {
         Some(sc.read().class_object().ok_or_else(|| Error::from("Base class of a global class is corrupted"))?)
     } else {
         None
     };
 ```
?

adrian17 (approved):

Issue #16007, opened 2024-04-16
RustError("RangeError: Vector is fixed")


OceanBreezeGames:
### Describe the bug

In an old flash project I have the following code which works in flash without errors:
powerupTilePoints = new Vector.<cTilePoint>(gridW, true);
<other code ... >
powerupTilePoints.splice(0, powerupTilePoints.length);

But in Ruffle this Error is thrown when I try to splice the Vector:
RustError("RangeError: Vector is fixed")

### Expected behavior

I would expect that it should run without errors as flash does even though technically it is correct to throw that error.

### Content Location

Sorry nothing to publicly share at this time.
You should be able to just create a fixed size vector in flash and then splice it like in the description to replicate.

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

Microsoft Edge Version 123.0.2420.97 (Official build) (64-bit)

### Additional information

Using old Air SDK 29.

PR #16010, opened 2024-04-16
avm2: Allow splice() on a Vector with fixed=true

adrian17:
Fixes  https://github.com/ruffle-rs/ruffle/issues/16007 .


tests/tests/swfs/avm2/vector_splice_fixed_bug_compat/Test.as
```diff
@@ -0,0 +1,22 @@
...
+
+var v = new Vector.<int>(5, true);
+try {
+    v.push(5);
+} catch(e) {
+    trace("push exception caught");
+}
+v.splice(0, 3);
+trace("NO exception on splice")
+
```
evilpie, line 12: It would probably be good to trace the vector (size/contents) afterwards

adrian17: done

evilpie (approved):

torokati44:
(Yeah, stating the obvious, but: Looks like there's a "last empty line" mismatch between the expected and actual outputs.)

PR #16014, opened 2024-04-17
tests: Fix a typo in tab_ordering_tabbable

kjarosh:
Fortunately the implementation was right all along :eyes:


torokati44 (approved):
Huh, nice catch. Also, I believe you!

PR #16023, opened 2024-04-17
frontend-utils: Don't use tokio tasks inside ruffle's executor

Dinnerbone:
This fixes the deadlocks on desktop when enough requests are made.

Seems to be the best option I can think of, really. #15981 does kinda patch over it but the things will never end up executing due to the tokio budget system (we should still bring that in anyway though, it covers a different kind of deadlock)

### DIFF
--- a/frontend-utils/src/backends/executor.rs
+++ b/frontend-utils/src/backends/executor.rs
@@ -5,6 +5,7 @@ use async_channel::{unbounded, Receiver, Sender};
 use ruffle_core::backend::navigator::OwnedFuture;
 use ruffle_core::loader::Error;
 use slotmap::{new_key_type, SlotMap};
+use std::future::Future;
 use std::sync::atomic::{AtomicBool, Ordering};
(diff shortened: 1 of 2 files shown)


frontend-utils/src/backends/executor.rs
```diff
@@ -273,3 +274,16 @@ impl<R: PollRequester> FutureSpawner for AsyncFutureSpawner<R> {
...
+pub async fn spawn_tokio<F>(future: F) -> F::Output
+where
+    F: Future + Send + 'static,
+    F::Output: Send + 'static,
+{
+    let (sender, receiver) = tokio::sync::oneshot::channel();
+    tokio::spawn(async move { sender.send(future.await) });
+    tokio::task::unconstrained(receiver)
+        .await
+        .expect("Oneshout should succeed")
```
sleepycatcoding, line 288: nit: typo

sleepycatcoding (approved):

PR #16031, opened 2024-04-18
tests: Improve test lookup and add support for test kinds

kjarosh:
Regression tests were looked up using full directory walk regardless of whether there was an exact match or not. This patch looks up the test on exact match without scanning for all tests. This improvement is important due to how nextest executes tests: one by one using exact matches.

Additionally, this PR adds support for test kinds: `libtest_mimic` adds test kind as a prefix (e.g. `[kind] test` instead of `test`), which has to be taken into account when filtering for possible test candidates. This patch ensures compatibility with `libtest_mimic` when using test kinds.

Related comment: https://github.com/ruffle-rs/ruffle/pull/15950#issuecomment-2052141069

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5248,6 +5248,7 @@ dependencies = [
  "futures",
  "image 0.25.1",
  "libtest-mimic",
+ "regex",
  "ruffle_core",
  "ruffle_render_wgpu",
  "ruffle_test_framework",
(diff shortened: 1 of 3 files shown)


adrian17 (approved):

kjarosh:
Fixed path traversal detection, turns out I missed one `.canonicalize()` (it's a shame Rust does not have path normalization :( )

torokati44:
> Fixed path traversal detection,

CI disagrees. :V :sweat_smile:

Dinnerbone (approved):

PR #16035, opened 2024-04-19
avm2: Remove outdated TODO comments

Lord-McSweeney:


core/src/avm2/globals/flash/display/display_object.rs
```diff
@@ -903,9 +903,6 @@ pub fn set_scroll_rect<'gc>(
                 object_to_rectangle(activation, rectangle)?,
             );
 
-            // TODO: Technically we should accept only `flash.geom.Rectangle` objects, in which case
-            // `object_to_rectangle` will be infallible. Once this happens, the following line can
```
Lord-McSweeney, line 907: > the following line can be moved above the `set_next_scroll_rect` call.

I didn't see the point in doing that, so I left it as-is.

core/src/avm2/value.rs
```diff
@@ -34,8 +34,6 @@ pub enum Hint {
 }
 
 /// An AVM2 value.
-///
-/// TODO: AVM2 also needs Scope, Namespace, and XML values.
```
adrian17, line 38: this comment isn't _entirely_ irrelevant, as in: IIRC FP does have a primitive type for namespaces. I don't really understand _why_ or if it can ever matter, so yeah let's drop it :)

adrian17 (approved):

PR #16044, opened 2024-04-19
ci: Drop the `mathiasvr/command-output@v2.0.0` action from the `release_amo` workflow

torokati44:
We keep getting a note because of it about Node.js 16 actions being deprecated:
https://github.com/ruffle-rs/ruffle/actions/runs/8670264413

See: https://github.com/mathiasvr/command-output/issues/10

Also see: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

### DIFF
--- a/.github/workflows/release_amo.yml
+++ b/.github/workflows/release_amo.yml
@@ -1,54 +1,48 @@
 name: Release to addons.mozilla.org
 
 on:
-  #Run weekly every Friday
+  # Run weekly every Friday
   schedule:
    - cron: "0 2 * * 6"
-  
-  #Allow for manual releases if needed
+
+  # Allow for manual releases if needed
   workflow_dispatch:
 
 jobs:
   submit-amo:
     name: Submit to addons.mozilla.org
     runs-on: ubuntu-22.04
-  
+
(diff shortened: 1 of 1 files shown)


torokati44:
I still have to test this somehow, of course.

torokati44:
I'm now _reasonably_ sure that this will work:

Here's a run of this workflow on top of current-ish `master`, changed so it `echo`es the relevant variables in the scripts where they are used, before and after any bash string manipulation shenanigans (stripping a trailing newline from them):
https://github.com/torokati44/ruffle/actions/runs/8767230456/job/24060238513

And here's a run on top of the changes in this PR, where the newlines are not even there in the first place, and the `-` -> `_` substitution is done right at the start:
https://github.com/torokati44/ruffle/actions/runs/8767332491/job/24060455747

danielhjacobs (approved):
Looks good to me, and those runs look promising. Only one way to truly test it, and that's to let it run in two days and hope for the best.

torokati44:
Thank you for the review. Unless anyone objects, or wants to take another look, I'll merge this around tomorrow evening (UTC).

PR #16067, opened 2024-04-21
frontend-utils: Unify warning format strings; add warning enum

sleepycatcoding:
This lays the groundwork for translated warnings.

### DIFF
--- a/frontend-utils/src/parse.rs
+++ b/frontend-utils/src/parse.rs
@@ -2,7 +2,7 @@ use std::fmt;
 use std::fmt::Formatter;
 use std::ops::Deref;
 use std::str::FromStr;
-use toml_edit::{ArrayOfTables, DocumentMut, Item, Table, TableLike};
+use toml_edit::{ArrayOfTables, DocumentMut, Item, Table, TableLike, TomlError};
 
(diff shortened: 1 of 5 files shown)


frontend-utils/src/parse.rs
```diff
@@ -88,18 +88,46 @@ impl<T: fmt::Debug> fmt::Debug for ParseDetails<T> {
...
+    },
+    UnsupportedValue {
+        value: String,
+        path: String,
+    },
+}
+
+impl ParseWarning {
+    /// Creates a human readable string from the warning.
+    pub fn display(&self) -> String {
```
Dinnerbone, line 112: Why not impl std::fmt::Display?

sleepycatcoding: Mostly because if we do want to translate these in the future, its not trivial to pass a language identifier into `std::fmt::Display`, though we could keep `Display` as a wrapper defaulting to `en-US`.

sleepycatcoding: Changed to fmt::Display for now. I'll let someone else figure out the way to translate these :smile:

Dinnerbone (approved):

PR #16069, opened 2024-04-21
avm2: Fix regression to do with not registering superclasses in the right order

Lord-McSweeney:


core/src/avm2/script.rs
```diff
@@ -114,6 +113,23 @@ impl<'gc> TranslationUnit<'gc> {
...
+    pub fn load_classes(self, activation: &mut Activation<'_, 'gc>) -> Result<(), Error<'gc>> {
+        // Classes must be loaded in the order they appear in the constant pool,
+        // to ensure that superclasses are loaded before subclasses
+
+        let num_classes = self.0.read().classes.len();
+        for i in 0..num_classes {
+            let class = self.load_class(i as u32, activation)?;
+
+            // FIXME: This should actually export it in a way so that it's visible within the
+            // domain, but not visible when accessing from e.g. ApplicationDomain functions.
```
adrian17, line 125: I'm still not sure about the pool/domain edge cases, so maybe a more generic wording like "There are subtle differences between how a class is exported after creation, and how it's exported again when script traits are handled. TODO: investigate this"?

adrian17 (approved):

Issue #16072, opened 2024-04-22
 Boxhead The Nightmare: Red line around the text/buttons


MartySVK:
### Describe the bug

In the menu scene, you can see red lines around the text/ buttons.

In Ruffle:
![image](https://github.com/ruffle-rs/ruffle/assets/10010569/f82e55ec-b43b-4cfc-82f4-38ae8f23f355)

In Flash:
![image](https://github.com/ruffle-rs/ruffle/assets/10010569/1a33863a-b7ce-43a3-a601-1b8b9fa0a5e1)

### Expected behavior

Game online - https://www.gameflare.com/online-game/boxhead-the-nightmare/
SWF - https://data.gameflare.com/games/2038/NZgxE0rRkmU3GS.swf

### Content Location

Show text/buttons without red lines

### Affected platform

Desktop app

### Operating system

Windows 11

kjarosh:
My comment from Discord:

> Unfortunately it looks like it has been like that for a long time, it's difficult to tell whether the red box was introduced by some commit, as this game has just started working very recently. I verified that the red box was there as early as November 2023, the game worked very poorly before that

PR #16114, opened 2024-04-24
avm2: Send ADDED_TO_STAGE in SimpleButton at correct time

adrian17:
The biggest risk is that I don't really have a good understanding of what the motivation for the old implementation was for.
But looking at the new one by itself, it feels... right?

Fixes https://github.com/ruffle-rs/ruffle/issues/16072 - the errors in `Boxhead the Nightmare`, where the `ADDED_TO_STAGE` event fired before they were actually on stage.

This probably needs some more just-in-case testing.


danielhjacobs:
test.swf is named Test.swf

SN902:
I checked half of the games from my library, PR looks quite stable

Aaron1011 (approved):

PR #16150, opened 2024-04-28
avm2: Minor performance improvements

Lord-McSweeney:
This PR has been stripped of most of the larger changes. The only notable change remaining is a fast path for setting index properties on arrays.

### DIFF
--- a/core/src/avm2/activation.rs
+++ b/core/src/avm2/activation.rs
@@ -2858,7 +2858,7 @@ impl<'a, 'gc> Activation<'a, 'gc> {
         let dm = self.domain_memory();
         let mut dm = dm
             .as_bytearray_mut(self.context.gc_context)
-            .ok_or_else(|| "Unable to get bytearray storage".to_string())?;
+            .expect("Bytearray storage should exist");
 
         let Ok(address) = usize::try_from(address) else {
             return Err(make_error_1506(self));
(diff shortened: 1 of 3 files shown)


SN902:
I found a regression
[derpy FPS test.zip](https://github.com/ruffle-rs/ruffle/files/15142566/derpy.FPS.test.zip)

`thread 'main' panicked at core\src\avm2\property_map.rs:70:13:
internal error: entered unreachable code: Lookup on lazy Multiname should never happen (Multiname { ns: Multiple([]), name: Some("rslProgressHandler"), param: None, flags: MultinameFlags(HAS_LAZY_NS | IS_QNAME) })
AVM2 stack trace:
        at mx.managers::SystemManager/http://www.adobe.com/2006/flex/mx/internal::initialize()
        at mx.managers::SystemManager/initHandler()`

core/src/avm2/activation.rs
```diff
@@ -1418,25 +1420,43 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
-                        self.context.gc_context,
-                    );
+            let object_value = self.context.avm2.peek(1);
 
-                    return Ok(FrameControl::Continue);
+            if let Value::Object(object) = object_value {
+                match name_value {
+                    Value::Integer(name_int) if name_int >= 0 => {
+                        if let Some(mut array) =
+                            object.as_array_storage_mut(self.context.gc_context)
```
adrian17, line 1437: shouldn't this rather be made similarly to `op_get_property`, with a "virtual" `set_index_property`? As written, this doesn't support vectors or bytearray.

PR #16170, opened 2024-04-30
desktop: Skip waiting for resize event after resize fails

colin969:
Under hyprland tiling window manager, the resize event fails and the loading stops until the user manually resizes the window to trigger a resize event. 

This should fix that by not waiting for a resize event to continue loading if we know the resize failed.

The resize also fails even if the window is floating, so that may be something worth looking into separately.

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -440,8 +440,8 @@ impl App {
                     let viewport_size = self.window.inner_size();
 
                     // On X11 (and possibly other platforms), the window size is not updated immediately.
-                    // Wait for the window to be resized to the requested size before we start running
-                    // the SWF (which can observe the viewport size in "noScale" mode)
+                    // On a successful resize request, wait for the window to be resized to the requested size
+                    // before we start running the SWF (which can observe the viewport size in "noScale" mode)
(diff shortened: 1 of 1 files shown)


desktop/src/app.rs
```diff
@@ -440,7 +442,7 @@ impl App {
                     // On X11 (and possibly other platforms), the window size is not updated immediately.
                     // Wait for the window to be resized to the requested size before we start running
                     // the SWF (which can observe the viewport size in "noScale" mode)
```
torokati44, line 444: I feel like this comment could use a word or two's worth of updating with this change...?

colin969: Maybe swap out `Wait for the window to be resized to the requested size...` for `On a successful resize request, wait for...`?

torokati44: Yep, that sounds good!

torokati44 (approved):
No regressions found on Wayland (KDE or Weston), nor in XWayland (KDE), and it makes sense, plus I believe you that this fixes a use case, so in it goes, thanks!

torokati44:
It would have been better to test on Windows too, but I'm like 98.4% certain that it won't cause any issues there...

Issue #14949, opened 2024-01-25
web: display_file_save_dialog is non-functional


Victo-2350703:
# Error Info
Error name: Error
Error message: panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rfd-0.13.0/src/file_handle/web.rs:55:17:
This File Handle doesn't support reading. Use `pick_file` to get a readable FileHandle
Error stack:
```
339/z/n.wbg.__wbg_new_3a66822ed076951c@https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.2e2fb3ce7210bdd35fc9.js:1:70044
core::ops::function::FnOnce::call_once{{vtable.shim}}::h61e7c84c3ecc6317@:wasm-function[11302]:0xa78281
std::panicking::rust_panic_with_hook::h62090d3fc0630473@:wasm-function[6906]:0x938fe6
core::panicking::panic_fmt::hd79411a297d06dc8@:wasm-function[11024]:0xa719de
rfd::file_handle::web::FileHandle::read::{{closure}}::{{closure}}::h1e64924434865afa@:wasm-function[5393]:0x8a4d00
wasm_bindgen::convert::closures::invoke2_mut::h1cbe87ffa8499fd4@:wasm-function[12247]:0xa961e0
closure9655 externref shim@:wasm-function[12251]:0xa96271
339/z/n.wbg.__wbg_new_1d93771b84541aa5/t</<@https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.2e2fb3ce7210bdd35fc9.js:1:71434
339/z/n.wbg.__wbg_new_1d93771b84541aa5/t<@https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.2e2fb3ce7210bdd35fc9.js:1:71470
339/z/n.wbg.__wbg_new_1d93771b84541aa5@https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.2e2fb3ce7210bdd35fc9.js:1:71364
js_sys::Promise::new::h0dd111e06e5da0e5@:wasm-function[13654]:0xaccfca
ruffle_web::ui::WebFileDialogResult::new::{{closure}}::h00e737636dad6760@:wasm-function[3672]:0x76aeb8
[...]

PR #16178, opened 2024-04-30
web: Download the file on FileReference.save

kjarosh:
Fixes https://github.com/ruffle-rs/ruffle/issues/14949.

This PR adds support for saving files on web using FileReference. When writing data, a download is triggered with the default file name. Currently, there's no dialog that lets the user select save destination.

This PR also ensures that all implementations of FileDialogResult behave the same way: desktop, web, and tests. The methods `write` and `refresh` have been merged into one: `write_and_refresh`, which allows the tests and web implementations behave the same way as desktop.

Two tests were added:
1. `avm2/filereference_save_and_browse` — This test verifies how FileReference behaves when a browse (and load) operations are triggered after save. Additionally, it marks the commented out code related to reporting file size as FIXME, because the current implementation reports the size at all times, instead of only after saving.
2. `avm2/filereference_save_and_load` — This test ensures that the error 2037 is thrown when load is called after save. It's currently a known failure as Ruffle is not as strict as FP regarding throwing the error 2037.


web/src/ui.rs
```diff
@@ -116,6 +120,91 @@ impl FileDialogResult for WebFileDialogResult {
...
+        a.set_attribute("download", &data.file_name).ok()?;
+        a.dyn_into::<HtmlElement>().ok()?.click();
+        data.downloaded = true;
+        Some(())
+    }
+}
+
+impl FileDialogResult for WebDownloadFileDialogResult {
+    fn is_cancelled(&self) -> bool {
+        self.data().downloaded
```
evilpie, line 166: I am not sure if that is correct

web/src/ui.rs
```diff
@@ -116,6 +120,91 @@ impl FileDialogResult for WebFileDialogResult {
...
+        None
+    }
+
+    fn creator(&self) -> Option<String> {
+        None
+    }
+
+    fn contents(&self) -> &[u8] {
+        tracing::warn!("Reading downloaded file not implemented");
+        &[]
```
evilpie, line 195: I actually thought that using `load()` after everything from `save` is complete would allow access to data, but I currently can't reproduce that anymore. Something to look into.

web/src/ui.rs
```diff
@@ -116,6 +120,91 @@ impl FileDialogResult for WebFileDialogResult {
...
+        let blob = Blob::new_with_u8_array_sequence(&array).ok()?;
+        let window = web_sys::window()?;
+        let document = window.document()?;
+        let a = document.create_element("a").ok()?;
+
+        let url = JsUrl::create_object_url_with_blob(&blob).ok()?;
+        a.set_attribute("href", &url).ok()?;
+        a.set_attribute("download", &data.file_name).ok()?;
+        a.dyn_into::<HtmlElement>().ok()?.click();
+        data.downloaded = true;
```
evilpie, line 159: You should call `JsUrl::revoke_object_url`

web/src/ui.rs
kjarosh: That was not correct, it was my first assumption about the code

web/src/ui.rs
kjarosh: That's definitely on my TODO list

web/src/ui.rs
kjarosh: It's true, I was a bit hesitant to call it immediately after click, but it seems to work

web/src/ui.rs
kjarosh: TBH I also had some trouble testing it, so I decided to basically make every `FileDialogResult` implementation behave the same way. That was possible by joining `write` and `refresh` to `write_and_refresh`. This way the SWF tests also verify whether this behavior is correct.

Now I'm only wondering whether it's covered by SWF tests sufficiently :thinking:

kjarosh: Okay, I really tried to break it (call load after save, or something), but it seems that FP always throws error 2037 when the order is not right, it's much more stricter than Ruffle currently (we should probably model it as a state enum and disallow some transitions).

The only discrepancy I found was related to reporting the file size during saving (it was even commented out in tests), but that's a problem not related to this PR (and it affects all platforms).

evilpie: I had actually started experimenting with type states, but I wasn't really in the mood of rewriting a lot of the code after it already took weeks to land the initial version. I think it's not a bad idea, but testing all the FP edge cases will get annoying.

More test are always appreciated.

tests/tests/swfs/avm2/filereference_save_and_browse/Test.as
```diff
@@ -0,0 +1,81 @@
...
+
+function oncomplete(e) {
+    trace("complete event");
+    dump(e.target);
+
+    file.removeEventListener(Event.SELECT, onselect);
+    file.removeEventListener(Event.COMPLETE, oncomplete);
+    file.addEventListener(Event.SELECT, onselect2);
+    file.addEventListener(Event.COMPLETE, oncomplete2);
+    file.browse([new FileFilter("debug-select-success", "*.txt")]);
```
evilpie, line 50: Might be simpler to just do something like the following instead of duplicating all the events.

```rust
if (saving) {
   file.browse()
}
saving = false;
```

web/src/ui.rs
```diff
@@ -33,19 +36,51 @@ impl std::error::Error for FullScreenError {
...
         };
 
-        Self { handle, contents }
+        #[cfg(not(target_arch = "wasm32"))]
+        let file_name = None;
+
+        #[cfg(not(target_arch = "wasm32"))]
+        let modification_time = None;
+
+        #[cfg(target_arch = "wasm32")]
```
evilpie, line 59: Please don't interleave those.

web/src/ui.rs
```diff
@@ -56,9 +91,28 @@ fn get_extension_from_filename(filename: &str) -> Option<String> {
...
+    let array = Uint8Array::from(data);
+    let blob = Blob::new_with_u8_array_sequence(&Array::of1(&array))?;
+    let window = web_sys::window().ok_or(JsValue::from("no window"))?;
+    let document = window.document().ok_or(JsValue::from("no document"))?;
+    let a = document.create_element("a")?;
+
+    let url = JsUrl::create_object_url_with_blob(&blob)?;
+    a.set_attribute("href", &url)?;
+    if let Some(filename) = filename {
+        a.set_attribute("download", filename)?;
```
evilpie, line 104: Pretty sure you want to always set this attribute to force a download.

web/src/ui.rs
```diff
@@ -312,30 +350,13 @@ impl UiBackend for WebUiBackend {
...
         &mut self,
-        _file_name: String,
+        file_name: String,
         _title: String,
     ) -> Option<DialogResultFuture> {
-        None
-        /* Temporary disabled while #14949 is being fixed
-
-        // Prevent opening multiple dialogs at the same time
-        if self.dialog_open {
```
evilpie, line 322: You need to keep this.

evilpie (approved):

web/src/ui.rs
kjarosh: You mean the targets or the variables? I made the variables not interleaved now, but at the end one of them has to be :/

web/src/ui.rs
kjarosh: That's right, without it the downloaded file may be displayed, fixed

web/src/ui.rs
kjarosh: Oops, reverted

web/src/ui.rs
evilpie: Sorry, I am not sure if you were waiting on a response here. This looks exactly like what I had in mind. :+1:

BarbourSmith:
👀 😀

BarbourSmith:
❤️ ❤️ ❤️ ❤️ !!! Thank you! I will test as soon as the demo updates!

PR #16185, opened 2024-05-01
web: Don't append and subsequently remove link just to download files

danielhjacobs:
The `download_as_file` function in #16178, which is the `web_sys` version of this function just with `&[u8]` as input instead of a blob, made me realize this is unnecessary as it's not done in that function.


kjarosh (approved):

torokati44 (approved):
LGTM 🤷‍♂️

PR #16206, opened 2024-05-03
avm2: Implement XML.contains

cy-polo:
[XML.contains()](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/XML.html#contains()) method implemented.

Test:
```as
<?xml version="1.0" encoding="utf-8"?>
<mx:Application xmlns:mx="http://www.adobe.com/2006/mxml" layout="absolute" minWidth="955" minHeight="600" creationComplete="init()">
	<mx:Script>
		<![CDATA[			
			private function init():void {
				var xml:XML = 
					<order>
						<item>Rice</item>
						<item>Kung Pao Shrimp</item>
					</order>;
				trace(xml.item[0].contains(<item>Rice</item>)); // true
				trace(xml.item[1].contains(<item>Kung Pao Shrimp</item>)); // true
				trace(xml.item[1].contains(<item>MSG</item>)); // false
			}
		]]>
	</mx:Script>
</mx:Application>
```

### DIFF
--- a/core/src/avm2/globals/XML.as
+++ b/core/src/avm2/globals/XML.as
@@ -68,7 +68,7 @@ package {
         AS3 native function child(name:*):XMLList;
         AS3 native function childIndex():int;
         AS3 native function children():XMLList;
-        AS3 native function contains(name:*):Boolean;
+        AS3 native function contains(value:*):Boolean;
         AS3 native function copy():XML;
         AS3 native function parent():*;
         AS3 native function elements(name:* = "*"):XMLList;
@@ -172,9 +172,9 @@ package {
             return self.AS3::children();
         };
 
-        prototype.contains = function(name:*):Boolean {
+        prototype.contains = function(value:*):Boolean {
             var self:XML = this;
-            return self.AS3::contains(name);
+            return self.AS3::contains(value);
         };
 
         prototype.copy = function():XML {
--- a/core/src/avm2/globals/XML.as
+++ b/core/src/avm2/globals/XML.as
@@ -68,6 +68,7 @@ package {
         AS3 native function child(name:*):XMLList;
         AS3 native function childIndex():int;
         AS3 native function children():XMLList;
+        AS3 native function contains(name:*):Boolean;
         AS3 native function copy():XML;
         AS3 native function parent():*;
         AS3 native function elements(name:* = "*"):XMLList;
@@ -169,6 +170,11 @@ package {
         prototype.children = function():XMLList {
             var self:XML = this;
             return self.AS3::children();
+        };
+
+        prototype.contains = function(name:*):Boolean {
+            var self:XML = this;
+            return self.AS3::contains(name);
         };
 
         prototype.copy = function():XML {
--- a/core/src/avm2/object/xml_object.rs
+++ b/core/src/avm2/object/xml_object.rs
@@ -183,14 +183,6 @@ impl<'gc> XmlObject<'gc> {
         if let Some(xml) = value.as_object().and_then(|obj| obj.as_xml_object()) {
             let other = xml.node();
 
-            if node.is_text() || node.is_attribute() {
-                return node.xml_to_string(activation) == other.xml_to_string(activation);
-            }
-
-            if other.is_text() || other.is_attribute() {
-                return false;
-            }
-
             return node.equals(&other);
         }
 
--- a/core/src/avm2/object/xml_object.rs
+++ b/core/src/avm2/object/xml_object.rs
@@ -177,6 +177,46 @@ impl<'gc> XmlObject<'gc> {
         Ref::map(self.0.read(), |data| &data.node)
     }
 
+    pub fn contains(&self, value: &Value<'gc>, activation: &mut Activation<'_, 'gc>) -> bool {
+        let node = self.node();
+
+        if let Some(xml) = value.as_object().and_then(|obj| obj.as_xml_object()) {
+            let other = xml.node();
+
+            if node.is_text() || node.is_attribute() {
+                return node.xml_to_string(activation) == other.xml_to_string(activation);
+            }
+
+            if other.is_text() || other.is_attribute() {
+                return false;
+            }
+
+            return node.equals(&other);
+        }
+
+        if let Some(list) = value.as_object().and_then(|obj| obj.as_xml_list_object()) {
+            if list.length() == 1 {
+                let xml = list
+                    .xml_object_child(0, activation)
+                    .expect("List length was just verified");
+
+                if node.is_text() || node.is_attribute() {
+                    return node.xml_to_string(activation) == xml.node().xml_to_string(activation);
+                }
+
+                if xml.node().is_text() || xml.node().is_attribute() {
+                    return false;
+                }
+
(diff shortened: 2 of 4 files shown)


core/src/avm2/object/xml_object.rs
```diff
@@ -177,6 +177,46 @@ impl<'gc> XmlObject<'gc> {
...
+        if let Some(xml) = value.as_object().and_then(|obj| obj.as_xml_object()) {
+            let other = xml.node();
+
+            if node.is_text() || node.is_attribute() {
+                return node.xml_to_string(activation) == other.xml_to_string(activation);
+            }
+
+            if other.is_text() || other.is_attribute() {
+                return false;
+            }
```
sleepycatcoding, line 192: Any particular reason why we are manually comparing the text nodes and not delegating this to the `equals` method?

core/src/avm2/object/xml_object.rs
```diff
@@ -177,6 +177,46 @@ impl<'gc> XmlObject<'gc> {
...
+
+            return node.equals(&other);
+        }
+
+        if let Some(list) = value.as_object().and_then(|obj| obj.as_xml_list_object()) {
+            if list.length() == 1 {
```
sleepycatcoding, line 198: I am not sure if FP actually considers single-element XMLLists here, further testing would be required. (Though looking at avmplus, they seem to just check for a XML object and use `_equals`)

sleepycatcoding (commented):
Thank you and welcome! :wave:

core/src/avm2/object/xml_object.rs
cy-polo: **Not sure of what I'm saying, I'm not a XML expert**

The default equality check might simply compare the nodes' memory addresses, which would only return true if they are the exact same node. But in the context of XML, we might want to consider two nodes equal if they have the same type (text or attribute) and the same content, even if they are not the same node in memory.
And perhaps this method is faster on large XML documents than using `equal`.

sleepycatcoding: The `equals` method you are calling below is defined here: https://github.com/ruffle-rs/ruffle/blob/d1553a203e679d7bb6e8fc37979970087f344f73/core/src/avm2/e4x.rs#L213

This already does the necessary type matching and comparisons.

core/src/avm2/object/xml_object.rs
cy-polo: I did a test to see if FP supports single-element lists, and it looks like it does.
I hope I've answered your question correctly.

```as
<?xml version="1.0" encoding="utf-8"?>
<mx:Application xmlns:mx="http://www.adobe.com/2006/mxml" layout="absolute" minWidth="955" minHeight="600" creationComplete="init()">
	<mx:Script>
		<![CDATA[           
			private function init():void {
				// XML with single item
				var singleItemXML:XML = <order><item>Rice</item></order>;
				
				// Test if Flash Player considers single-element XML
				trace("Is singleItemXML a single element? " + singleItemXML.contains(<item>Rice</item>)); // true
				
				// Original XML with multiple items
				var xml:XML = 
					<order>
						<item>Rice</item>
						<item>Kung Pao Shrimp</item>
					</order>;
				trace("Is xml a single element? " + xml.contains(<item>Rice</item>)); // false
				
				// Test contains method with original XML
				trace(xml.item[0].contains(<item>Rice</item>)); // true
				trace(xml.item[1].contains(<item>Kung Pao Shrimp</item>)); // true
				trace(xml.item[1].contains(<item>MSG</item>)); // false
			}
		]]>
	</mx:Script>
</mx:Application>
```

FP result:
```
Is singleItemXML a single element? false
Is xml a single element? false
true
true
false
```

Ruffle result:
```
2024-05-04T09:36:59.023293Z  INFO avm_trace: Is singleItemXML a single element? false
2024-05-04T09:36:59.023318Z  INFO avm_trace: Is xml a single element? false
2024-05-04T09:36:59.023330Z  INFO avm_trace: true
2024-05-04T09:36:59.023336Z  INFO avm_trace: true
2024-05-04T09:36:59.023341Z  INFO avm_trace: false
```

sleepycatcoding:
```actionscript
var xml:XML = <order><item>Rice</item><item>Kung Pao Shrimp</item></order>;

trace(xml.item[0].contains(<item>Rice</item>)); // true, both in Ruffle and FP

var list:XMLList = new XMLList("<item>Rice</item>");
trace(xml.item[0].contains(list)); // true in Ruffle with this PR, false in FP
```

So from my testing we do not need to handle XMLLists in the contains method at all, so we can safely remove this.

nit: I think it might be cleaner to move the contains implementation into the `globals/xml.rs` file, it really does not need to be on the XmlObject.

cy-polo: I don't know how to move the `contains` implementation into the `globals/xml.rs` file, I let you do it, I'm not quite familiar with Rust syntax at the moment.

Anyway, according to our tests, the implementation of contains seems to work well!

sleepycatcoding: > I don't know how to move the `contains` implementation into the `globals/xml.rs` file, I let you do it, I'm not quite familiar with Rust syntax at the moment.

Something like this should work:
```rust
pub fn contains<'gc>(
    _activation: &mut Activation<'_, 'gc>,
    this: Object<'gc>,
    args: &[Value<'gc>],
    let xml = this.as_xml_object().unwrap();
    let value = args.get_value(0);

    if let Some(other) = value.as_object().and_then(|obj| obj.as_xml_object()) {
        let result = xml.node().equals(&other.node());
        return Ok(result.into());
    }
    
    Ok(false.into())
}
```

cy-polo: That's right, it works, thank you! PR ready to be merged 😉

core/src/avm2/globals/XML.as
```diff
@@ -68,6 +68,7 @@ package {
         AS3 native function child(name:*):XMLList;
         AS3 native function childIndex():int;
         AS3 native function children():XMLList;
+        AS3 native function contains(name:*):Boolean;
```
sleepycatcoding, line 71: One last thing, can we rename this argument to value instead?

core/src/avm2/globals/XML.as
```diff
@@ -171,6 +172,11 @@ package {
             return self.AS3::children();
         };
 
+        prototype.contains = function(name:*):Boolean {
+            var self:XML = this;
+            return self.AS3::contains(name);
+        };
+
```
sleepycatcoding, line 179: Here too.

sleepycatcoding (approved):
Just a few last nits, otherwise LGTM.

core/src/avm2/globals/XML.as
cy-polo: > Here too.

I renamed!

torokati44:
Can I ask for the commit history to be cleaned up a bit before merging? 😊

sleepycatcoding:
> Can I ask for the commit history to be cleaned up a bit before merging? 😊

I had the plan to squash it :sweat_smile:

Issue #16181, opened 2024-04-30
avm1: MovieClipLoader onLoad events use the incorrect clip


colin969:
### Describe the bug

Ruffle reuses the target clip when calling onLoad events (e.g `onLoadStart`) when it should be a different clip. This becomes very obvious when replacing the root movie.

In Flash, when replacing the root clip `_root` becomes identical to the target given and it's empty.

2 theories I could think of:
- Flash creates a new clip to pass as the target, and changes the behavior of `_root` to point to it during these events. After it's done it replaces the target.
- The target clip is used, but `_root` is cleared before the onLoad events are called somehow?

I couldn't find documentation for this but everything in my experimenting seems to lean this way.

This may relate to https://github.com/ruffle-rs/ruffle/issues/2132

### Expected behavior

Flash Player 32 Debug trace output for test below:
```
undefined 
Target Message
Target Message
```

Current Ruffle output:
```
Original Message
Target Message
undefined
```

### Content Location

test.as
```as
class Test {
  static function main() {
    _root.message = "Original Message";

    var loader = new MovieClipLoader();

    var o = new Object();
    o.onLoadStart = function(target) {
      trace(target.message);
      _root.message = "Target Message";
      trace(target.message);
    };

    loader.addListener(o);
    loader.loadClip("child.swf", "_level0");
  }
}
```

child.as
```as
class Child {
    static function main() {
        trace(_root.message);
    }
}
```

This behavior is used consistently by Kongregate's AS2 API - The API is passed the url of the swf to load, and sets up objects like `target.kongregateServices` before loading it into itself. Currently Kong's API isn't easy to test as it requires modification to communicate properly in Ruffle.

### Affected platform

Desktop app

### Operating system

Linux, Arch

Issue #9255, opened 2023-01-21
Clip variables not reset after replacing the clip


itamar567:
### Describe the bug

If you load a movie clip into a variable (for example, `loadClip("clip1.swf",testclip)`), and in that clip a variable is set (say, `var somevar = 3`), if you replace the clip in another clip (so now `testclip` points to a different clip), the new clip will have the variable set to the value it was set to in the old clip (`somevar` will be set to 3, instead of undefined).

Here is a simple example of it: [attachment.zip](https://github.com/ruffle-rs/ruffle/files/10473019/attachment.zip)

### Expected behavior

The new clip should have `somevar` undefined.

### Affected platform

Desktop app

### Operating system

Ubuntu 22.04.1

kjarosh:
It works fine, added a test to make sure we don't break this behavior

Issue #2132, opened 2020-12-28
[AVM1] MovieClipLoader and persistent variables


moulins:
Tested with commit 9e46757f of Ruffle.

The following code gives different results on Flash and Ruffle.
```as
var file = "inner.swf";

var mc = flash.Lib.current.createEmptyMovieClip(file, 1);
var loader = new MovieClipLoader();

mc.VALUE1 = "Hello!";
trace("inner.VALUE1=" + mc.VALUE1);
trace("inner.VALUE2=" + mc.VALUE2);

loader.onLoadComplete = function(mc) {
  mc.VALUE2 = "World!";
  trace("onLoadComplete: inner.VALUE1=" + mc.VALUE1);
  trace("onLoadComplete: inner.VALUE2=" + mc.VALUE2);
};

loader.onLoadInit = function(mc) {
  trace("onLoadInit: inner.VALUE1=" + mc.VALUE1);
  trace("onLoadInit: inner.VALUE2=" + mc.VALUE2);
}

trace("Loading " + file);
loader.loadClip(file, mc);
```

On Flash:
```
inner.VALUE1=Hello!
inner.VALUE2=undefined
Loading inner.swf
onLoadComplete: inner.VALUE1=undefined
onLoadComplete: inner.VALUE2=World!
onLoadInit: inner.VALUE1=undefined
onLoadInit: inner.VALUE2=World!
```

On Ruffle:
```
inner.VALUE1=Hello!
inner.VALUE2=undefined
Loading inner.swf
onLoadComplete: inner.VALUE1=undefined
onLoadComplete: inner.VALUE2=World!
onLoadInit: inner.VALUE1=Hello!
onLoadInit: inner.VALUE2=undefined
```

It seems that Ruffle incorrectly ignore all modifications made to the `MovieClip` in `onLoadComplete` and resets its state to before the loading started.

Files: 

[mc-loading.zip](https://github.com/ruffle-rs/ruffle/files/5747586/mc-loading.zip)

Lord-McSweeney:
Fixed, probably by #16210

PR #16210, opened 2024-05-04
avm1: Make target have correct properties when firing events after `loadClip`

colin969:
Changes the properties of the target clip used in `loadClip` to reflect what is available during and after the `onLoad` events when executed in Flash.

- Clears deletable properties of the target clip before starting the `onLoad` events
- When replacing root, preserves properties by copying them before calling `replace_root_movie` then pushing them back onto the root afterwards

Added test `mcl_loadclip_properties` to verify that deletable properties are cleared before `onLoadStart` and are correct in subsequent `onLoad` events.
Added test `mcl_loadclip_replace_root` to verify that properties set during `onLoadStart` remain available after replacing root.
Removed test `mcl_loadclip_onload_target` in favour of the 2 added ones being more appropriate.

Should solve the behaviour in these issues:
- https://github.com/ruffle-rs/ruffle/issues/2132
- https://github.com/ruffle-rs/ruffle/issues/16181
- https://github.com/ruffle-rs/ruffle/issues/9255

This fixes cases like Kongregate's AS2 API swf which relies on `onLoadStart` to add API properties to the loaded game which then replaces the root. At some point Kong stopped loading the API altogether for AS2 games and directly loads the game swf, so unfortunately this is difficult to test on the official site.

I'm not confident how / where I'm copying the root properties across is right even though it does work, so feedback there would be nice.


core/src/loader.rs
```diff
@@ -1019,7 +1035,43 @@ impl<'gc> Loader<'gc> {
...
+                        let mut root_properties: IndexMap<AvmString, Value> = IndexMap::new();
+                        if let Some(root) = uc.stage.root_clip() {
+                            let root_val = root.object();
+                            if let Value::Object(root_object) = root_val {
+                                let mut activation = Activation::from_nothing(
+                                    uc.reborrow(),
+                                    ActivationIdentifier::root("unknown"),
+                                    root,
+                                );
+                                for key in root_object.get_keys(&mut activation, true) {
```
Dinnerbone, line 1048: Do we know if it should include hidden keys?
Can probably test it using `ASSetPropFlags`

colin969: Yes, Kong's API makes all the properties it sets hidden. I'm still not certain why though.

colin969: For more context

```actionscript
   static function onAbstractLoadStart(target)
   {
      target._lockroot = true;
      for(var i in _global.kongregate_vars)
      {
         target[i] = _global.kongregate_vars[i];
         _global.ASSetPropFlags(target,i,7);
         _global[i] = _global.kongregate_vars[i];
         _global.ASSetPropFlags(_global,i,7);
      }
      com.kongregate.common.util.Log.init();
      com.kongregate.common.util.Log.info("Creating services...");
      target.kongregateServices = _global.kongregateServices = com.kongregate.client.KongregateServices.getInstance();
      target.kongregateScores = _global.kongregateScores = com.kongregate.client.HighScoreServices.getInstance();
      target.kongregateStats = _global.kongregateStats = com.kongregate.client.StatServices.getInstance();
      target.kongregateChat = _global.kongregateChat = com.kongregate.client.ChatServices.getInstance();
      target.kongregateMtx = _global.kongregateMtx = com.kongregate.client.MicrotransactionServices.getInstance();
      target.kongregate = _global.kongregate = {services:com.kongregate.client.KongregateServices.getInstance(),scores:com.kongregate.client.HighScoreServices.getInstance(),stats:com.kongregate.client.StatServices.getInstance(),chat:com.kongregate.client.ChatServices.getInstance(),mtx:com.kongregate.client.MicrotransactionServices.getInstance()};
      _global.ASSetPropFlags(_global,["kongregate","kongregateServices","kongregateScores","kongregateStats","kongregateChat","kongregateMtx"],7);
      _global.ASSetPropFlags(target,["kongregate","kongregateServices","kongregateScores","kongregateStats","kongregateChat","kongregateMtx"],7);
      AbstractAPIBootstrap._subclassListener.onLoadStart(target);
   }
   ```

Dinnerbone (approved):

PR #16227, opened 2024-05-06
docs: Fix scanner command and broken mtasc links

colin969:
Just some small problems that tripped me up recently.


torokati44:
Oh shoot, MTASC down...
BTW we also link to it here: https://github.com/ruffle-rs/ruffle/wiki/Helpful-Resources#tools

colin969:
Huh, somehow never found that wiki page before. Edited it there too.

adrian17 (approved):

PR #16232, opened 2024-05-06
docs: Replace mtasc link with archive.org snapshot in CONTRIBUTING.md

danielhjacobs:


torokati44:
This is already part of https://github.com/ruffle-rs/ruffle/pull/16227.

Issue #13250, opened 2023-09-19
"Prom Week" gets stuck at 100% loaded and has a graphical glitch


fpdotmonkey:
### Describe the bug

When I start the game [Prom Week](https://promweek.soe.ucsc.edu/play/), it begins to load and the loading bar seems to indicate it's done, but it never progresses past the loading screen.

As well, the content part of the window is wider than the title bar of the window.

![image](https://github.com/ruffle-rs/ruffle/assets/3139615/e9deea85-5b08-4ddd-88ff-bf80d12eb081)

### Expected behavior

I would expect the game to start after it's done loading and for the title bar and content of the window to be the same width.

### Affected platform

Desktop app

### Operating system

Ubuntu 22.04

### Additional information

After a while, this error message comes up when run from console.

```
2023-09-19T18:35:20.460396Z ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x556cb80187f0 }) to handler FunctionObject(FunctionObject { ptr: 0x556cb62a6690 }) : ReferenceError: Error #1056: Cannot create property cffHinting on flash.text.engine.FontDescription.
```

Lord-McSweeney:
This now takes a while to load (tested using desktop), but finally fails on an error "`null is not an Object`". Full log:
```
 INFO ruffle_core::library: Loaded new device font "Noto Sans" from swf tag
 INFO ruffle_core::player: Loaded SWF version 10, resolution 758x600 @ 30 FPS
 WARN ruffle_core::stub: Encountered stub: AVM2 flash.utils.Dictionary constructor with weak keys
 WARN ruffle_core::stub: Encountered stub: AVM2 flash.display.Graphics.drawRoundRectComplex()
 WARN ruffle_core::library: Unknown device font "Times New Roman"
 WARN ruffle_core::library: Unknown device font "Verdana"
 WARN ruffle_core::display_object::movie_clip: DefineFont4 tag (TLF text) is not implemented
 WARN ruffle_core::display_object::movie_clip: DefineFont4 tag (TLF text) is not implemented
 WARN ruffle_core::display_object::movie_clip: DefineFont4 tag (TLF text) is not implemented
 WARN ruffle_core::display_object::movie_clip: DefineFont4 tag (TLF text) is not implemented
 WARN ruffle_core::display_object::movie_clip: DefineFont4 tag (TLF text) is not implemented
[...]

PR #16235, opened 2024-05-07
Implement `describeType(null)` stub

### LINKED ISSUE #10143: Burrito Bison: Stuck before title screen
### Describe the bug

https://www.newgrounds.com/portal/view/562620?emulate=flash
Tested it out on the web page and tested it on desktop via spoofing the web page to get past the site lock, the results still the same, the game gets stuck on the screen here and nothing shows up in the console logs.

Burrito Bison Ruffle

### Expected behavior

The game should actually start

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 110

### Additional information

_No response_

Coder-256:
With this patch, Burrito Bison now works all but flawlessly in my testing, aside from some choppy music!: http://onemorelevel.com/static/games3/burritobison.swf

Fixes #10143.

This is my first contribution to this amazing project, I'm wondering what is the simplest way to create a test for this?

---

Note: the SWF on Newgrounds linked in #10143 issue gives this error screen when testing locally (haven't tried compiling the extension and testing on the web but that may resolve it):

Screenshot 2024-05-06 at 8 36 49 PM

The SWF I linked above works great.

### DIFF
--- a/core/src/avm2/globals/avmplus.as
+++ b/core/src/avm2/globals/avmplus.as
@@ -109,9 +109,6 @@ package avmplus {
     }
 
     public function describeType(value: *, flags: uint):XML {
-        if (value == null) {
-            return <type name="null" isDynamic="false" isFinal="true" isStatic="false"/>;
-        }
         var json = describeTypeJSON(value, flags);
         var xml = <type />;
         xml.@name = json.name;
--- a/core/src/avm2/globals/avmplus.as
+++ b/core/src/avm2/globals/avmplus.as
@@ -109,6 +109,9 @@ package avmplus {
     }
 
     public function describeType(value: *, flags: uint):XML {
+        if (value == null) {
+            return <type name="null" isDynamic="false" isFinal="true" isStatic="false"/>;
+        }
         var json = describeTypeJSON(value, flags);
         var xml = <type />;
         xml.@name = json.name;
--- a/core/src/avm2/globals/avmplus.rs
+++ b/core/src/avm2/globals/avmplus.rs
@@ -15,9 +15,12 @@ pub fn describe_type_json<'gc>(
     _this: Object<'gc>,
     args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
-    let value = args[0].coerce_to_object(activation)?;
     let flags = DescribeTypeFlags::from_bits(args.get_u32(activation, 1)?).expect("Invalid flags!");
+    if args[0] == Value::Null {
+        return describe_type_json_null(activation, flags);
+    }
 
+    let value = args[0].coerce_to_object(activation)?;
     let class_obj = value.as_class_object().or_else(|| value.instance_of());
     let object = activation
         .avm2()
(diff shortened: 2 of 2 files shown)


Coder-256:
I did a quick search for other issues that may be related and it does seem that "Prom Week" from #13250 now gets past the loading screen (but is still unplayable due to graphical glitches).

core/src/avm2/globals/avmplus.as
```diff
@@ -109,6 +109,9 @@ package avmplus {
     }
 
     public function describeType(value: *, flags: uint):XML {
+        if (value == null) {
+            return <type name="null" isDynamic="false" isFinal="true" isStatic="false"/>;
```
Lord-McSweeney, line 113: From what I can see this should actually be implemented in `describeTypeJSON`. Try comparing the output of `describeType(null,INCLUDE_BASES)` between this branch and FP (with `avmplus.describeType` and `avmplus.INCLUDE_BASES`)

Coder-256:
@Lord-McSweeney Ok, I've moved it to `avmplus.rs`. The new implementation is effectively identical, except it throws an error in a few cases where real FP throws an error.

My version isn't technically true to the original `avmplus`, but it behaves in effectively the same way from the perspective of an external user. Is this better?

core/src/avm2/globals/avmplus.rs
```diff
@@ -80,6 +83,48 @@ bitflags::bitflags! {
...
 
+fn describe_type_json_null<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    flags: DescribeTypeFlags,
+) -> Result<Value<'gc>, Error<'gc>> {
+    // These checks will effectively cause `describeType` to faithfully mirror FP's
+    // implementation. That is to say, it will cause it to throw an "Error #1009: Cannot access
+    // a property or method of a null object reference."
+    if flags.contains(DescribeTypeFlags::USE_ITRAITS)
+        || !flags.contains(DescribeTypeFlags::INCLUDE_BASES | DescribeTypeFlags::INCLUDE_TRAITS)
```
Lord-McSweeney, line 94: Where does this `!flags.contains(DescribeTypeFlags::INCLUDE_BASES | DescribeTypeFlags::INCLUDE_TRAITS)` come from? I haven't been able to get `describeTypeJSON` to return `null` except when `USE_ITRAITS` (the first condition) is set.

Lord-McSweeney: By the way you can directly test the output of `describeTypeJSON` with a SWF that uses a function in the `avmplus` package to interface with `describeTypeJSON`, like so:
```
package avmplus
{
    public function describeTypeHack(a:*, b:*) : *
    {
        return describeTypeJSON(a,b);
    }
}
```

core/src/avm2/globals/avmplus.rs
```diff
@@ -80,6 +83,48 @@ bitflags::bitflags! {
...
+
+    traits.set_public_property("bases", ArrayObject::empty(activation)?.into(), activation)?;
+    traits.set_public_property("interfaces", Value::Null, activation)?;
+    traits.set_public_property("variables", Value::Null, activation)?;
+    traits.set_public_property("accessors", Value::Null, activation)?;
+    traits.set_public_property("methods", Value::Null, activation)?;
+    traits.set_public_property("metadata", Value::Null, activation)?;
+    traits.set_public_property("constructor", Value::Null, activation)?;
+
+    object.set_public_property("traits", traits.into(), activation)?;
```
Lord-McSweeney, line 123: Seems like traits gets set to `null` if `INCLUDE_TRAITS` isn't set

core/src/avm2/globals/avmplus.rs
```diff
@@ -80,6 +83,48 @@ bitflags::bitflags! {
...
+    object.set_public_property("isFinal", true.into(), activation)?;
+    object.set_public_property("isStatic", false.into(), activation)?;
+
+    let traits = activation
+        .avm2()
+        .classes()
+        .object
+        .construct(activation, &[])?;
+
+    traits.set_public_property("bases", ArrayObject::empty(activation)?.into(), activation)?;
```
Lord-McSweeney, line 115: Seems like `bases` gets set to `null` if `INCLUDE_BASES` isn't set

Coder-256:
@Lord-McSweeney My reasoning was that since `describeTypeJSON` is marked as an internal method, any behavior that ends up identical from the perspective of the end user is acceptable (`describeType` is the only public way to call the method). Is it also a goal to make `describeTypeJSON` behave more closely to FP in case the user uses it via something like reflection? I'm trying to understand why the difference matters if it is an internal method.

Coder-256:
To your points that `traits` is set to null if `INCLUDE_TRAITS` isn't set, and that `bases` is set to null if `INCLUDE_BASES` isn't set, in both cases this triggers a null object error inside `describeType`. That is why I simply chose to short circuit rather than implement these methods more accurately since the fields aren't accessible to code outside the avmplus package if I understand correctly.

Coder-256:
Ok, in this version I now do believe it should fully work correctly even with something like `describeTypeHack`.

Lord-McSweeney (approved):

PR #16244, opened 2024-05-07
core: Invalidate cached bitmap on every newly decoded video frame

torokati44:
Not just when seeking manually.

Probably finally fixes https://github.com/ruffle-rs/ruffle/issues/12327.


torokati44:
I'm not actually sure those `*activation.context.needs_render = true;` lines are really necessary, but probably won't hurt either...?

Dinnerbone (approved):
Definitely don't think the needs_render is required, but sure!

torokati44:
My thinking was along the lines of "if for any reason an FLV packet is processed while the SWF timeline is stopped, or is just at a very low frame rate, and no input events, make sure we don't make the video any choppier than it has to be". But maybe these things can never actually happen...

PR #16250, opened 2024-05-08
core,render: Improve performance of "round to nearest, ties to even" float operations

kjarosh:
This patch improves performance of `ecma_conversions::round_to_even()` and `matrix::round_to_i32()`:
1. by using `f64::round_ties_even()`/`f32::round_ties_even()`, which have been stable since 1.77.0, instead of a custom algorithm; and
2. by removing an unnecessary comparison to `i32::MIN`, as casting a float to an integer automatically saturates values smaller than the minimum integer value to the minimum value of the integer type.

A primitive benchmark shows around 20% faster execution of these operations. In the case of `matrix::round_to_i32()` we can potentially expect noticeable improvements in overall performance as this operation is quite commonly used.

| Platform | Operation | Type | Improvement |
| -------- | --------- | ---- | ------- |
| amd64    | `ecma_conversions::round_to_even()` | `f64` | ~21% |
| WASM amd64 | `ecma_conversions::round_to_even()` | `f64` | ~20% |
| amd64    | `matrix::round_to_i32()` | `f32` | ~22% |
| WASM amd64 | `matrix::round_to_i32()` | `f32` | ~23% |

Those functions have been covered by tests to ensure the behavior hasn't changed.

### DIFF
--- a/core/src/ecma_conversions.rs
+++ b/core/src/ecma_conversions.rs
@@ -45,17 +45,11 @@ pub fn f64_to_wrapping_i32(n: f64) -> i32 {
 
 /// Implements the IEEE-754 "Round to nearest, ties to even" rounding rule.
 /// (e.g., both 1.5 and 2.5 will round to 2).
-/// Although this is easy to do on most architectures, Rust provides no standard
-/// way to round in this manner (`f64::round` always rounds away from zero).
-/// For more info and the below code snippet, see: https://github.com/rust-lang/rust/issues/55107
 /// This also clamps out-of-range values and NaN to `i32::MIN`.
-/// TODO: Investigate using SSE/wasm intrinsics for this.
 pub fn round_to_even(n: f64) -> i32 {
-    let k = 1.0 / f64::EPSILON;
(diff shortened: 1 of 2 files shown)


adrian17 (approved):
LGTM.
Out of curiosity, did you find any content that's particularly slow due to these, or is this PR more inspired by `round_ties_even` being stabilized?

kjarosh:
I remember analyzing performance of some SWF game where the results showed that sometimes `matrix::round_to_i32` took a noticeable amount of time. I tried optimizing matrix operations using SIMD intrinsics (`round_ties_even` was unstable at the time), but that was extremely complicated, and I basically would have had to implement portable SIMD.

Today I noticed that `round_ties_even` has been stabilized so that was a no-brainer optimization to do :)

torokati44:
> and I basically would have had to implement portable SIMD.

We use the `wide` crate for this over in https://github.com/ruffle-rs/h263-rs/. It maps neatly to intrinsics  on supported platforms, and expands to scalar stuff otherwise.

PR #16260, opened 2024-05-08
debug_ui: Exclude hidden objects properly when searching

kjarosh:
The option "Include Hidden" was not working properly, as it did not exclude all hidden objects when disabled. An object is hidden also when its ancestor is not visible.

| Before | After |
| ------ | ----- |
| <image src="https://github.com/ruffle-rs/ruffle/assets/1507072/b3255e1a-1382-409f-9e82-92866e498e9c" width="300"/> | <image src="https://github.com/ruffle-rs/ruffle/assets/1507072/b9fd565e-7638-427b-ab4e-9e833945bae5" width="300"/> |


torokati44:
Does the same thing apply with "Only Mouse Enabled Objects"?

kjarosh:
> Does the same thing apply with "Only Mouse Enabled Objects"?

Hmmm... I don't think _the same thing_ applies to mouse enabled, but there's `mouseChildren` that's not taken into account. I'm thinking I could also add "Only Tab Enabled Objects" by the way because the search logic is somewhat similar to that :eyes: 

But I would prefer to do that in a separate PR anyways as it requires more thinking and testing.

Dinnerbone (approved):

PR #16276, opened 2024-05-10
avm2: Optimize `findpropstrict`, `findproperty`, and `getlex`

Lord-McSweeney:
Also propagate types from `getslot`, `getglobalscope`, and `getglobalslot`.

A single `getlex QName(PackageNamespace(""), "Math")` executes about 2.5x faster with this PR.

New stats for box2d:
```
FindPropStrict -> GetScriptGlobals : 45.16%
FindPropStrict -> GetOuterScope : 7.87%
FindPropStrict -> GetScopeObject : 1.69%

FindProperty -> GetOuterScope : 2.83%
FindProperty -> GetScopeObject : 5.66%

InitProperty -> SetSlot : 38.15%
InitProperty -> SetSlotNoCoerce : 37.87%

SetProperty -> SetSlot : 15.03%
SetProperty -> SetSlotNoCoerce : 30.42%
SetProperty -> CallMethod : 1.57%

GetProperty -> GetSlot : 59.69%
GetProperty -> CallMethod : 0.34%

CallProperty -> CallMethod : 31.45%

CallPropVoid -> CallMethod : 61.74%

Coerce -> Nop : 46.95%

CoerceD -> Nop : 78.36%

CoerceB -> Nop : 8.11%

CoerceU -> Nop: 32.46%

CoerceI -> Nop : 2%

ReturnValue -> ReturnValueNoCoerce : 65.14%
```

(note: `getproperty` probably went down in percentage because we now split `getlex` into `findpropstrict` and `getproperty`, and the `getproperty` doesn't always optimize to a `getslot`- so the percentage looks like it went down, even though the getlex is faster)

Overall any optimization improvements other than the early scope resolution are tiny. Stockfish seems to start a little faster.

TODO:
- [x] Handle lazy multiname passed to `findproperty`/`findpropstrict`
- [x] Appease clippy

### DIFF
--- a/core/src/avm2/optimize.rs
+++ b/core/src/avm2/optimize.rs
@@ -265,6 +265,10 @@ pub fn optimize<'gc>(
         }
     }
 
+    if !method_body.exceptions.is_empty() {
+        has_simple_scoping = false;
+    }
+
     // TODO: Fill out all ops, then add scope stack and stack merging, too
     let mut state_map: HashMap<i32, Locals<'gc>> = HashMap::new();
 
@@ -769,17 +773,20 @@ pub fn optimize<'gc>(
                     }
 
                     if !stack_push_done {
-                        if let Some((class, index)) =
-                            outer_scope.get_entry_for_multiname(&multiname)
-                        {
-                            *op = Op::GetOuterScope {
-                                index: index as u32,
-                            };
+                        if let Some(info) = outer_scope.get_entry_for_multiname(&multiname) {
+                            if let Some((class, index)) = info {
+                                *op = Op::GetOuterScope {
+                                    index: index as u32,
+                                };
 
-                            stack_push_done = true;
-                            if let Some(class) = class {
-                                stack.push_class_object(class);
+                                stack_push_done = true;
+                                if let Some(class) = class {
+                                    stack.push_class_object(class);
+                                } else {
+                                    stack.push_any();
+                                }
                             } else {
+                                stack_push_done = true;
                                 stack.push_any();
                             }
                         }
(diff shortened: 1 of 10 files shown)


core/src/avm2/optimize.rs
```diff
@@ -769,17 +773,20 @@ pub fn optimize<'gc>(
                     }
 
                     if !stack_push_done {
-                        if let Some((class, index)) =
-                            outer_scope.get_entry_for_multiname(&multiname)
-                        {
-                            *op = Op::GetOuterScope {
-                                index: index as u32,
-                            };
+                        if let Some(info) = outer_scope.get_entry_for_multiname(&multiname) {
```
adrian17, line 777: can you add more comments here and in the function signature type?
Smth like "we hit a `with` scope, so we can't look deeper at all"

Alternatively, change `Option<Option<` to a 3-value enum (but I'm okay with just a comment here)

core/src/avm2/verify.rs
```diff
@@ -302,6 +330,12 @@ pub fn verify_method<'gc>(
                             1078,
                         )?));
                     }
+
+                    assert!(bytes_read > 1);
+                    byte_info[previous_position as usize] =
+                        ByteInfo::OpStart(AbcOp::FindPropStrict { index });
+                    byte_info[(previous_position + 1) as usize] =
```
adrian17, line 337: can you document that it's a slightly hacky way to emit multiple ops, with explanation how it works?

adrian17 (approved):
Approved, assuming that review notes about comments are addressed.

PR #16282, opened 2024-05-10
avm1: Implement NetConnection, more or less

Dinnerbone:
Tests are ported from avm2

Fixes #15823
Fixes #3317

PR #16284, opened 2024-05-10
avm1: Implement __resolve

### LINKED ISSUE #16182: prototype.__resolve conflict
### Describe the bug

Flash player 6 with AS1.o :: Self-Hosted
I am using prototype.__resolve which gets fired automatically just after onData event.

BTW - no external js or any file used here

it is working fine in compiled swf standalone move also works fine inside browser with flash player but doesn't works with Ruffle because "prototype.__resolve" never fired.
seems some conflict with Ruffle.js

### Expected behavior

if things works fine in compiled mode independent swf and it should work with Ruffle

### Content Location

my local machine

### Affected platform

Self-hosted version

### Operating system

win 10

### Browser

mozilla 115, MS Edge, GC

### Additional information

_No response_

Dinnerbone:
Fixes #16182

When combined with #16282, this fixes http://www.eugenicsarchive.org/eugenics/list2.pl (but not g2conline, sadly)


core/src/avm1/object.rs
```diff
@@ -696,8 +696,10 @@ pub fn search_prototype<'gc>(
     activation: &mut Activation<'_, 'gc>,
     this: Object<'gc>,
     is_slash_path: bool,
+    include_resolve: bool,
```
Lord-McSweeney, line 699: `include_resolve` is always true now, right?

core/src/avm1/object.rs
```diff
@@ -732,5 +734,36 @@ pub fn search_prototype<'gc>(
         depth += 1;
     }
 
+    if include_resolve {
+        if let Some(resolve) = find_resolve_method(orig_proto, activation)? {
+            let result =
+                resolve.call("__resolve".into(), activation, this.into(), &[name.into()])?;
+            return Ok(Some((result, 0)));
```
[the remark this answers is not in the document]
Dinnerbone: I'm not sure it's possible to know, and it likely doesn't matter. It only applies to `super` as far as I can tell, and that can't apply with something returned through `__resolve`.
However, `__resolve` tries to act like the thing existed on the property itself; so 0 is probably as correct as anything!

core/src/avm1/object.rs
Dinnerbone: You're right!

adrian17 (approved):

PR #16303, opened 2024-05-12
avm2: Implement StyleSheet

Dinnerbone:
This does **not** implement `TextField.stylesheet`, so you can't really do anything interesting with this stylesheet once it's made. But, the class itself should be fully implemented.

All the interesting logic is extracted out so that AVM1 can share it, too.


core/src/avm2/globals/flash/text/text_format.rs
```diff
@@ -185,20 +185,51 @@ pub fn set_color<'gc>(
...
+            Value::Undefined | Value::Null => {
+                text_format.display = None;
+                return Ok(Value::Undefined);
+            }
+            value => value.coerce_to_string(activation)?,
+        };
+
+        text_format.display = if value == WStr::from_units(b"block") {
+            Some(TextDisplay::Block)
+        } else if value == WStr::from_units(b"inline") {
```
adrian17, line 223: Could`value == b"inline"` or `&value == b"inline"` work here?

core/src/avm2/globals/flash/text/style_sheet.rs
```diff
@@ -0,0 +1,108 @@
...
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let document = args.get_string(activation, 0)?;
+    let result = activation
+        .avm2()
+        .classes()
+        .object
+        .construct(activation, &[])?;
+
+    if let Ok(css) = CssStream::new(document.as_wstr()).parse() {
```
adrian17, line 19: `&document`?

adrian17 (approved):

core/src/avm2/globals/flash/text/text_format.rs
Dinnerbone: > no implementation for `AvmString<'_> == &[u8; 6]`

I can do `*value == *b"block"` but that's getting a little... arcane

Issue #9736, opened 2023-02-26
Papa Louie 2 & 3: Freeze and high memory usage


n0samu:
### Describe the bug

When trying to play [Papa Louie 2](https://www.flipline.com/games/papalouie2/index.html) and [Papa Louie 3](https://www.flipline.com/games/papalouie3/index.html), the games freeze on the "Loading levels" screen:
![image](https://user-images.githubusercontent.com/71368227/221390400-5da353de-cd95-4677-be45-bdea875cdb59.png)

Ruffle's memory usage skyrockets, and then it crashes.

### Expected behavior

At the very least, we should try to fix the memory issue. Maybe by implementing #8416?

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Firefox 110

MartySVK:
Same probably for Cactus McCoy and Cactus McCoy 2

CreaZyp154:
In the devtools it shows this (tested with Papa Louie 2): 
```js
Uncaught RuntimeError: unreachable executed
    __wbg_adapter_64 moz-extension://58f0dc8b-0eca-495d-ae4d-43543be8493e/dist/339.js:294
    real moz-extension://58f0dc8b-0eca-495d-ae4d-43543be8493e/dist/339.js:271
    __wbg_requestAnimationFrame_d28701d8e57998d1 moz-extension://58f0dc8b-0eca-495d-ae4d-43543be8493e/dist/339.js:1340
    handleError moz-extension://58f0dc8b-0eca-495d-ae4d-43543be8493e/dist/339.js:309
    __wbg_requestAnimationFrame_d28701d8e57998d1 moz-extension://58f0dc8b-0eca-495d-ae4d-43543be8493e/dist/339.js:1339
    __wbg_adapter_64 moz-extension://58f0dc8b-0eca-495d-ae4d-43543be8493e/dist/339.js:294
    real moz-extension://58f0dc8b-0eca-495d-ae4d-43543be8493e/dist/339.js:271
    __wbg_requestAnimationFrame_d28701d8e57998d1 moz-extension://58f0dc8b-0eca-495d-ae4d-43543be8493e/dist/339.js:1340
    handleError moz-extension://58f0dc8b-0eca-495d-ae4d-43543be8493e/dist/339.js:309
    __wbg_requestAnimationFrame_d28701d8e57998d1 moz-extension://58f0dc8b-0eca-495d-ae4d-43543be8493e/dist/339.js:1339
[...]

PR #16308, opened 2024-05-13
core: Support dependent strings for concatenation

adrian17:
This makes affects all code that uses `AvmString::concat`, but the most important case is `"a" + "b"` in both AVM1 _and_ AVM2.
(also added it to AS3 `String::concat` while I was there)

With the PR, the code like
```
var a = "";
for (var i = 0; i < N; i += 1) { a += "x" };
```
Will allocate a new string buffer log(N) times, instead of N times. (note that it'll still allocate N string objects) This makes most code like this super fast instead of... crashing the player.

Should fix several SWFs, including https://github.com/ruffle-rs/ruffle/issues/9736 - though I didn't test it yet.

I still want to test the code a bit more (tbh adding test-only code to inspect string internals would be great 😢) and mostly get a safety review from @moulins if possible.
Also might still refactor it a bit, this entire code might be moved to `AvmStringRepr` to keep `AvmString` `unsafe`-free and avoid exposing internal pointers.


adrian17:
Forgot to add: for code that is _not_ egregiously written (so most code), this might slightly increase our memory usage.
Since if every string concat is only ever done once, like `data["c"+i]=1`, then the overallocation is never used.
That said, strings are usually not a huge % of total memory used, so it shouldn't make a noticeable difference on average.

MartySVK:
https://github.com/ruffle-rs/ruffle/issues/9736 is fixed with this PR 🥳

torokati44:
(After rebasing, I have amended this commit with a `cargo fmt` just to get rid of the red CI crosses, I hope you don't mind...)

core/src/string/avm_string.rs
```diff
@@ -38,6 +38,13 @@ impl<'gc> AvmString<'gc> {
         }
     }
 
+    pub fn as_managed(self) -> Option<Gc<'gc, AvmStringRepr<'gc>>> {
```
Aaron1011, line 41: Do we use the 'managed' terminology anywhere else? If not, could we call this `as_owned() `?

adrian17:
Note: the PR is still missing size overflow checks, but the pointer logic should be stable now.

core/src/string/avm_string.rs
adrian17: Oh sorry, forgot to mention (I also intended to add a comment).
Ideally the `Owned` naming is extremely misleading, since a `Source::Owned` string can either be dependent ("owned") or non-dependent ("owner"). IMO to resolve the overloaded usage of the term, the `Source::Owned` should become `Source::Managed` (in the "gc-managed" sense - feel free to propose alternatives), and owned/owner terminology will only relate to dependent strings.

I didn't want to refactor this naming in this PR, but I did "pave the way" with `.as_managed`, since otherwise having `left.owner().as_owned()` sounds like utter nonsense.

moulins (approved):

PR #16310, opened 2024-05-13
frontend-utils: Add bundle player options

sleepycatcoding:
Differences in the specification between the gist and this PR.

- Some examples in the gist referred to their old names, this has been fixed.
- Explicitly mentions that frame rate can be a floating-point value as well as an integer. (Changed for consistency, since script_timeout allows both)
- Fixed typo in `frame_rate` description (Maybe be -> May be).

cc @Dinnerbone

### DIFF
--- a/frontend-utils/src/player_options/read.rs
+++ b/frontend-utils/src/player_options/read.rs
@@ -1,0 +1,739 @@
+use crate::parse::{ItemExt, ParseContext, ReadExt};
+use crate::player_options::PlayerOptions;
+use std::time::Duration;
+use toml_edit::TableLike;
+
+pub fn read_player_options<'a>(
+    cx: &mut ParseContext<'a>,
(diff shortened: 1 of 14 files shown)


Dinnerbone (approved):
<3

frontend-utils/src/player_options/read.rs
```diff
@@ -0,0 +1,751 @@
...
+        // Values are case-sensitive.
+        let result = read("quality = \"High\"");
+        assert_eq!(&PlayerOptions::default(), result.values());
+        assert_eq!(
+            vec![ParseWarning::UnsupportedValue {
+                value: "High".to_string(),
+                path: "quality".to_string(),
+            }],
+            result.warnings
+        );
```
sleepycatcoding, line 270: I forgot to remove this when I merged case-sensitivity check into assert_variant. Will remove later.

PR #16324, opened 2024-05-14
debug_ui: Check alpha values for color transform

Aaron1011:
We were incorrectly displaying 'Default' for the color transform when only the alpha value was modified.


Dinnerbone (commented):
This doesn't feel right.

The code inside the branch you just changed specifically picks out RGB and A separately. Making it so they all have to be the same both removes the need to treat RGB vs A differently, but also makes it so it won't show shorthand for cases like `set rgb to 0, alpha 100%` or whatever.


If alpha is the only thing to be not-default, then `summary_color_transform_entry` should have returned `Some` - I don't understand how it could be giving "Default" in that scenario?

Aaron1011:
You're right - I misread this code while I was working on a different change

PR #16326, opened 2024-05-14
avm1: Implement most of StyleSheet

Dinnerbone:
`transform` is stubbed, but this is enough for tonight!


core/src/avm1/globals/style_sheet.rs
```diff
@@ -0,0 +1,207 @@
...
+    property_decl::Declaration, ArrayObject, ExecutionReason, NativeObject, ScriptObject,
+};
+use crate::avm1::{Activation, Error, Value};
+use crate::avm1_stub;
+use crate::context::GcContext;
+use crate::html::{transform_dashes_to_camel_case, CssStream, TextFormat};
+use crate::string::AvmString;
+use gc_arena::Gc;
+
+const PROTO_DECLS: &[Declaration] = declare_properties! {
```
Lord-McSweeney, line 13: Flash docs say that StyleSheet is version-gated to FP7, should these properties be marked as version-gated?

Dinnerbone: Done!

PR #16168, opened 2024-04-30
render,desktop: Port to `wgpu` `v0.20.0`, switch to `egui` `master`

torokati44:
~~Depends on https://github.com/emilk/egui/pull/4433 and https://github.com/bevyengine/naga_oil/pull/87/~~.
~~Also includes https://github.com/ruffle-rs/ruffle/pull/16077 because https://github.com/emilk/egui/pull/4394 is in `egui` `master`.~~

Builds on top of https://github.com/ruffle-rs/ruffle/pull/16330.
Fixes #15836 by switching to `egui` `master`.

Issue #15836, opened 2024-04-01
Ruffle logo missing in "About" dialog


torokati44:
### Describe the bug

![image](https://github.com/ruffle-rs/ruffle/assets/288816/b5d57c00-b33f-44de-a4de-b60e7f27146a)

### Expected behavior

![image](https://github.com/ruffle-rs/ruffle/assets/288816/f88386b9-9f7a-478a-b429-6027c2828398)

### Content Location

none/any

### Affected platform

Desktop app

### Operating system

indifferent

### Browser

N/A

### Additional information

Regressed by https://github.com/ruffle-rs/ruffle/pull/15634.
To be fixed by https://github.com/emilk/egui/pull/4160 - which is blocked on https://github.com/1Password/arboard/pull/142.
TODO after fix: Deny duplicating the `image` crate in deps using `cargo-deny`, similarly to `tracy-client`.

torokati44:
Related, but not a bug: This logo could be switched to `../assets/logo.svg` too, but it had some layout glitches when we last tried that IIRC.

PR #16330, opened 2024-05-15
`render/wgpu`: Drop `naga_oil` in favor of manual string manipulation

torokati44:
One fewer dependency. We've only used the very basics of its capabilities lately anyway.
Also, so we're not stuck waiting for the likes of https://github.com/bevyengine/naga_oil/pull/87.

Unblocks https://github.com/ruffle-rs/ruffle/pull/16168, and the update to `egui` `master` (temporarily), thereby the fixing of #15836.

TODO:
 - ~~sort out vertex colliding entry point names...?~~
 - ~~pipeline constants for `early_saturate`~~
   - After the switch to `wgpu` 0.20
   - Currently solved with a simple string substitution, the compiler will probably optimize out the dead piece of code.
 - investigate effect on shader compilation times (startup)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,7 +27,6 @@ updates:
         - "winit"
         - "wgpu"
         - "naga"
-        - "naga_oil"
         - "egui*"
         - "raw-window-handle"
       wasm-bindgen:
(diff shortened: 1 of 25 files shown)


torokati44:
> investigate effect on shader compilation times (startup)

Do you think you could take a look at this, @adrian17 ? You're the only one I remember having experienced any sort of noticeable startup delay...

Dinnerbone (approved):

PR #16335, opened 2024-05-15
`github`: Make Dependabot also ignore "profiling", add a group to bump `Cargo.lock`

torokati44:
"profiling" ignore:
The other two ignored crates are indirect through this.

`Cargo.lock` bump:
Hopefully, now that it runs on GHA infra, it won't timeout.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -45,3 +45,4 @@ updates:
     ignore:
       - dependency-name: "tracing-tracy"
       - dependency-name: "tracy-client"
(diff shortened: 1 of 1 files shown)


Aaron1011 (approved):

Dinnerbone:
fwiw we usually do `ci:`

torokati44:
> fwiw we usually do `ci:`, I think!

Fair enough, will read `CONTRIBUTING.md` next time, sorry! Too late to fix now... 😶

PR #16352, opened 2024-05-17
ci: Run Rust tests on Ubuntu 24.04

torokati44:
Currently available as beta on GHA: https://github.blog/changelog/2024-05-14-github-hosted-runners-public-beta-of-ubuntu-24-04-is-now-available/

Just want to see whether anything will break once we switch, plus whether we can clean up the tricks we did to get Lavapipe, and to tolerate SIGABRT.

A minor added benefit is that we won't lose CI if/when Launchpad is down (as has happened a couple times recently), and that nextest will alert us of genuinely flaky tests with higher probability (the retries hack won't hide them). Also, the Ubuntu jobs will be a bit faster.

Note that after merging this, the set of required checks will have to be updated by a repo admin to match.

### DIFF
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -13,7 +13,7 @@ concurrency:
 jobs:
   changes:
     name: Paths filter
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     outputs:
       should_run: ${{ steps.filter.outputs.src }}
     steps:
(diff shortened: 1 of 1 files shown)


torokati44:
Posted https://github.com/actions/runner-images/issues/10374 due to the Firefox WebDriver failures.

torokati44:
While the Edge people figure their shoot out, we could switch to 24.04 only for the Rust tests...?

torokati44:
> While the Edge people figure their shoot out, we could switch to 24.04 only for the Rust tests...?

Removed the web test workflow updates from this PR, will submit those separately.

torokati44:
This and #17176 together will make our CI setup "relatively" hack-free again (well, aside from #16682).

PR #16376, opened 2024-05-19
avm2: Refactor Executable->BoundMethod, introduce freestanding exec()

adrian17:
Fun fact, `Executable` has been almost unchanged since the very first AVM2 commit. Time for some code churn? :) 

The diff looks large, but most of it is just moving `exec()` implementation.

This merges several refactors:

- `NativeExecutable` and `BytecodeExcecutable` are basically identical, aside from the `method` field - so I collapsed them back to a single struct.
- At this point `Executable` is pretty much just a `BoundMethod`, so I renamed it to such.
- also did some internal renaming, like `superclass` is really just a `bound_class`, but I didn't do it outside of `Executable`.
- The only way to directly call a method is `Executable::from_method(...).exec(...)`, which is kinda pointless (and does pointless work of handling bound/unbound `this`), so I collapsed this to a single toplevel `exec(...)` function.

I checked for regressions on a microbenchmark (calls the `o.asdf` property getter repeatedly), the time on both wasm and x64 is within the margin of error.

### DIFF
--- a/core/src/avm2.rs
+++ b/core/src/avm2.rs
@@ -4,7 +4,6 @@ use std::rc::Rc;
 
 use crate::avm2::class::AllocatorFn;
 use crate::avm2::error::make_error_1107;
-use crate::avm2::function::Executable;
 use crate::avm2::globals::SystemClasses;
 use crate::avm2::method::{Method, NativeMethodImpl};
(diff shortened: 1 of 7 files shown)


kmeisthax (approved):
Looks good to me.

PR #16377, opened 2024-05-19
chore: Bump jpegxr dep to fix compiling with latest Clang

Aaron1011:


torokati44:
This is the change in `jpegxr` BTW: https://github.com/ruffle-rs/jpegxr/commit/db88651220688d2883a90d5477048071507b0493

Do you think this should be PR'ed upstream too?

Also I don't love the `libloading` downgrades in `Cargo.lock`, but they don't matter much either...

torokati44:
Reverted the downgrades. Thank you!

PR #16384, opened 2024-05-20
core: Fix stop() when executing attachMovie in DoInitAction

kjarosh:
This fixes a bug where the code in the first frame of a movie clip could have been executed after the second frame (so that instructions like stop did not work properly).

That is because when the first frame was executed in preload (e.g. for a sprite attached using attachMovie), the actions were queued, the second frame was executed in run_frame, and then the actions from the first frame were executed.

Ensuring that all actions are executed after preload fixes this issue.

This issue is related to https://github.com/ruffle-rs/ruffle/issues/5443 as it improves support for `mx.managers.FocusManager` (by preventing the `tabCapture` object from disappearing).

See the test `attach_movie_stop` for the MWE.


kjarosh:
> Could you please check whether this changes anything related to https://github.com/ruffle-rs/ruffle/issues/6144 ?

For future reference: it does not seem to change anything, for me it worked on the desktop even without this patch.

Dinnerbone (approved):

PR #16393, opened 2024-05-20
avm2: Implement Namespace constructor and prefix

Dinnerbone:
Needed to unblock proper namespace support in XML


evilpie:
```
from_avmplus/as3/RuntimeErrors/Error1098XmlIllegalPrefixFoNoNamespace
from_avmplus/e4x/Namespace/e13_2_1
from_avmplus/e4x/Namespace/e13_2_2
```
now pass :tada:

core/src/avm2/object/namespace_object.rs
```diff
@@ -65,7 +75,15 @@ impl<'gc> NamespaceObject<'gc> {
 
         let this: Object<'gc> = NamespaceObject(GcCell::new(
             activation.context.gc_context,
-            NamespaceObjectData { base, namespace },
+            NamespaceObjectData {
+                base,
+                namespace,
+                prefix: if namespace.as_uri_opt().is_some() {
```
sleepycatcoding, line 81: Any reason this is setting the prefix based on if the URI is `Some` and not like it is done in the constructor (checking if URI is empty)?

tests/tests/swfs/avm2/namespace_constr/Test.as
```diff
@@ -1,21 +1,104 @@
...
-}
-var n = new Namespace();
-
-trace("new Namespace()", n);
-
-n = new Namespace("prefix", "uri");
+	public class Test {
+		function Test() {
+			var otherNS = new Namespace("otherPrefix", "otherUri");
+			var values = [null, undefined, "test", "", "NOT A VALID PREFIX", otherNS];
```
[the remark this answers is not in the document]
Dinnerbone: Good thinking!

core/src/avm2/object/namespace_object.rs
Dinnerbone: Because I wrote that first and forgot to come back and update it :D

PR #16403, opened 2024-05-21
avm1: Move FunctionObject's base to FunctionObjectData

kjarosh:
This shrinks AVM1's `Object` (from 24 to 16 bytes on x64), and allows future shrinkage of AVM1's `Value` (e.g. from 24 to 16 bytes on x64 after putting all strings behind a `Gc` pointer).

### DIFF
--- a/core/src/avm1/function.rs
+++ b/core/src/avm1/function.rs
@@ -472,19 +472,12 @@ impl<'gc> From<Gc<'gc, Avm1Function<'gc>>> for Executable<'gc> {
 /// Represents an `Object` that holds executable code.
 #[derive(Clone, Collect, Copy)]
 #[collect(no_drop)]
-pub struct FunctionObject<'gc> {
-    /// The script object base.
-    ///
-    /// TODO: Can we move the object's data into our own struct?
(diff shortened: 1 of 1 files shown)


core/src/avm1/function.rs
```diff
@@ -472,26 +472,23 @@ impl<'gc> From<Gc<'gc, Avm1Function<'gc>>> for Executable<'gc> {
...
             .finish()
     }
 }
 
 #[derive(Clone, Collect)]
 #[collect(no_drop)]
 struct FunctionObjectData<'gc> {
+    /// The script object base.
+    ///
+    /// TODO: Can we move the object's data into our own struct?
```
Lord-McSweeney, line 490: I think the TODO can be removed now

Lord-McSweeney (approved):

core/src/avm1/function.rs
kjarosh: I originally removed this TODO but then noticed that the `FunctionObjectData` has been present and in use since the beginning and just assumed that I misunderstood the comment :eyes:

Removing it... (unless someone sees another meaning behind it)

PR #16404, opened 2024-05-21
avm2: Implement Namespace enumeration

sleepycatcoding:
Depends on #16393.


tests/tests/swfs/from_avmplus/e4x/Namespace/e13_2_5/test.toml
```diff
@@ -1,2 +1 @@
 num_ticks = 1
-known_failure = true
```
sleepycatcoding, line 2: This does not pass yet due to https://github.com/ruffle-rs/ruffle/pull/16393#discussion_r1608150947

Dinnerbone (approved):

Issue #15017, opened 2024-01-30
Game partially load (unplayable) - https://www.4399.com/flash/137953_4.htm


wsxarcher:
### Describe the bug

Game loads the board when selecting single player or multiplayer but it's unplayable because most objects are not loaded

```
ERROR core/src/library.rs:199 Tried to instantiate non-registered character ID 136
339.js:1133 ERROR core/src/display_object/movie_clip.rs:1700 Unable to instantiate display node id 136, reason being: Character id doesn't exist
339.js:1133 ERROR core/src/library.rs:199 Tried to instantiate non-registered character ID 242
339.js:1133 ERROR core/src/display_object/movie_clip.rs:1700 Unable to instantiate display node id 242, reason being: Character id doesn't exist
339.js:1133 WARN core/src/stub.rs:132 Encountered stub: AVM2 flash.display.Loader.unloadAndStop()
339.js:1133 WARN core/src/stub.rs:132 Encountered stub: AVM2 flash.display.Loader.unload()
339.js:1133 INFO web/src/log_adapter.rs:18 __________________--
339.js:1133 INFO web/src/log_adapter.rs:18 ctrl skin load OK !!!!
339.js:1133 INFO web/src/log_adapter.rs:18 ______________________________________
339.js:1133 INFO web/src/log_adapter.rs:18 __________  wmCtrl start  ____________
339.js:1133 INFO web/src/log_adapter.rs:18 ______________________________________
339.js:1133 INFO web/src/log_adapter.rs:18 _gameInfoStr = 712|490|137953|坦克动荡2
339.js:1133 WARN core/src/library.rs:523 Unknown device font "SimSun" (bold: false, italic: false)
339.js:1133 WARN core/src/library.rs:523 Unknown device font "Arial" (bold: true, italic: false)
339.js:1133 INFO web/src/log_adapter.rs:18 doSetStyle----------->100031854
339.js:1133 INFO web/src/log_adapter.rs:18 gameWid = 712
339.js:1133 INFO web/src/log_adapter.rs:18 gameHei = 490
339.js:1133 INFO web/src/log_adapter.rs:18 ************
339.js:1133 ERROR core/src/library.rs:199 Tried to instantiate non-registered character ID 136
339.js:1133 ERROR core/src/display_object/movie_clip.rs:1700 Unable to instantiate display node id 136, reason being: Character id doesn't exist
339.js:1133 ERROR core/src/library.rs:199 Tried to instantiate non-registered character ID 136
[...]

Issue #10305, opened 2023-03-21
The Last Dalek: Floor fails to load


waspennator:
### Describe the bug

http://downloads.bbc.co.uk/tv/isite-flash/doctorwho/games/lastdalek/isometrics.swf (Playable if you have the ruffle extension, otherwise yank it from flashpoint)
The game seems to work up until you start the level, at which point it loads into a room with missing assets like the floor, the hazmat suit characters walking around, and the music and sound effects.

![The Last Dalek Ruffle](https://user-images.githubusercontent.com/78768499/226740220-5abe3063-a025-470c-b00f-b0a9dea3f535.png)
[donbcfbmhbcapadipfkeojnmajbakjdc-1679432142501.log](https://github.com/ruffle-rs/ruffle/files/11033612/donbcfbmhbcapadipfkeojnmajbakjdc-1679432142501.log)

### Expected behavior

![The Last Dalek Flash](https://user-images.githubusercontent.com/78768499/226740271-19135dfa-c582-4f6a-a66b-a2e92cdb8eb0.png)
The game should be fully initializing like this.

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 110

waspennator:
Still occuring on latest nightly, either it's not rendering the assets cause checking the log, it can't "instantiate" non registered character ID's or it's something else i'm unsure of.

Lord-McSweeney:
This needs `ImportAssets`

TDGalea:
I have this game hosted [on my website](https://tdgalea.co.uk/o/Games/DoctorWho/TheLastDalek.html) with Ruffle embedded.
For me, the game doesn't even get past the first loading screen.
I'd be happy to help in any way I can (testing on-site, etc.) seeing the pull request from 22/05 as a start of hopefully getting this fixed.

waspennator:
Issue seems to be fixed up on my end and seems to play reasonably fine.

@TDGalea 
I tried running the game through your website and I think the problem is that the game is multiasset, seems like it's not properly retriving the assets for the game correctly.
[tdgalea.co.uk-1721454668639.log](https://github.com/user-attachments/files/16318981/tdgalea.co.uk-1721454668639.log)

waspennator:
Closing for now since the original issue of the floor not loading has been fixed though.

PR #16420, opened 2024-05-22
Implement ImportAssets/ImportAssets2

wsxarcher:
Load the ImportAssets(2) url async and copy the character requested.
This does not cover all cases because it requires extensive time of reversing the behavior of Flash but it fixes at least 2 real games I tested.

https://github.com/ruffle-rs/ruffle/issues/15017 (needs to set the HTTP referral if run on desktop, works perfectly as extension)
https://github.com/ruffle-rs/ruffle/issues/10305

### DIFF
--- a/core/src/avm1/property_map.rs
+++ b/core/src/avm1/property_map.rs
@@ -106,6 +106,10 @@ impl<'gc, V> PropertyMap<'gc, V> {
             self.0.shift_remove(&CaseInsensitive(key.as_ref()))
         }
     }
+
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
 }
 
 unsafe impl<'gc, V: Collect> Collect for PropertyMap<'gc, V> {
--- a/core/src/avm1/property_map.rs
+++ b/core/src/avm1/property_map.rs
@@ -106,10 +106,6 @@ impl<'gc, V> PropertyMap<'gc, V> {
             self.0.shift_remove(&CaseInsensitive(key.as_ref()))
         }
     }
-
-    pub fn len(&self) -> usize {
-        self.0.len()
-    }
 }
 
 unsafe impl<'gc, V: Collect> Collect for PropertyMap<'gc, V> {
--- a/core/src/display_object/movie_clip.rs
+++ b/core/src/display_object/movie_clip.rs
@@ -375,7 +375,6 @@ impl<'gc> MovieClip<'gc> {
                 flags: MovieClipFlags::PLAYING,
                 avm2_class: None,
                 drawing: Drawing::new(),
-                has_focus: false,
                 avm2_enabled: true,
                 avm2_use_hand_cursor: true,
                 button_mode: false,
--- a/core/src/display_object/movie_clip.rs
+++ b/core/src/display_object/movie_clip.rs
@@ -4214,17 +4214,15 @@ impl<'gc, 'a> MovieClipData<'gc> {
         &mut self,
         context: &mut UpdateContext<'_, 'gc>,
         id: CharacterId,
-    ) -> Result<Character<'gc>, Error> {
+    ) -> Option<Character<'gc>> {
         let library_for_movie = context.library.library_for_movie(self.movie());
 
         if let Some(library) = library_for_movie {
             if let Some(character) = library.character_by_id(id) {
-                return Ok(character.clone());
+                return Some(character.clone());
             }
         }
-        Err(Error::InvalidSwf(swf::error::Error::invalid_data(
-            "message",
-        )))
+        None
     }
 
     fn register_export(
@@ -4247,7 +4245,7 @@ impl<'gc, 'a> MovieClipData<'gc> {
                     .exported_name
                     .write(context.gc_context) = Some(*name);
             } else {
-                tracing::debug!(
+                tracing::warn!(
                     "Registering export for non-movie clip: {} (ID: {})",
                     name,
                     id
@@ -4273,26 +4271,32 @@ impl<'gc, 'a> MovieClipData<'gc> {
             let name = export.name.decode(reader.encoding());
             let name = AvmString::new(context.gc_context, name);
 
-            let character = self.get_registered_character_by_id(context, export.id)?;
+            if let Some(character) = self.get_registered_character_by_id(context, export.id) {
+                self.register_export(context, export.id, &name, self.movie());
+                tracing::debug!("register_export asset: {} (ID: {})", name, export.id);
 
-            self.register_export(context, export.id, &name, self.movie());
-            tracing::debug!("register_export asset: {} (ID: {})", name, export.id);
+                if self.importer_movie.is_some() {
+                    let parent = self.importer_movie.as_ref().unwrap().clone();
+                    let parent_library = context.library.library_for_movie_mut(parent.clone());
 
-            if self.importer_movie.is_some() {
-                let parent = self.importer_movie.as_ref().unwrap().clone();
-                let parent_library = context.library.library_for_movie_mut(parent.clone());
+                    if let Some(id) = parent_library.character_id_by_import_name(name) {
+                        parent_library.register_character(id, character);
 
-                if let Some(id) = parent_library.character_id_by_import_name(name) {
-                    parent_library.register_character(id, character);
-
-                    self.register_export(context, id, &name, parent);
-                    tracing::debug!(
-                        "Registering parent asset: {} (Parent ID: {})(ID: {})",
-                        name,
-                        id,
-                        export.id
-                    );
+                        self.register_export(context, id, &name, parent);
+                        tracing::debug!(
+                            "Registering parent asset: {} (Parent ID: {})(ID: {})",
+                            name,
+                            id,
+                            export.id
+                        );
+                    }
                 }
+            } else {
+                tracing::error!(
+                    "Export asset: {} (ID: {}) not found in library",
+                    name,
+                    export.id
+                );
             }
         }
         Ok(())
--- a/core/src/display_object/movie_clip.rs
+++ b/core/src/display_object/movie_clip.rs
@@ -19,7 +19,7 @@ use crate::avm1::{Activation as Avm1Activation, ActivationIdentifier};
 use crate::binary_data::BinaryData;
 use crate::character::{Character, CompressedBitmap};
 use crate::context::{ActionType, RenderContext, UpdateContext};
-use crate::context_stub;
+use crate::{context_stub, library};
 use crate::display_object::container::{dispatch_removed_event, ChildContainer};
 use crate::display_object::interactive::{
     InteractiveObject, InteractiveObjectBase, TInteractiveObject,
@@ -189,6 +189,12 @@ pub struct MovieClipData<'gc> {
 
     /// Attached audio (AVM1)
     attached_audio: Option<NetStream<'gc>>,
+
+    // TODO Consider moving this to InteractiveObject along with
+    //      TextField's and Button's tab indices after AVM2 analysis
+    tab_index: Option<i32>,
+
+    importer_movie: Option<Arc<SwfMovie>>,
 }
 
 impl<'gc> MovieClip<'gc> {
@@ -224,6 +230,8 @@ impl<'gc> MovieClip<'gc> {
                 tag_frame_boundaries: Default::default(),
                 queued_tags: HashMap::new(),
                 attached_audio: None,
+                tab_index: None,
+                importer_movie: None,
             },
         ))
     }
@@ -265,6 +273,8 @@ impl<'gc> MovieClip<'gc> {
                 tag_frame_boundaries: Default::default(),
                 queued_tags: HashMap::new(),
                 attached_audio: None,
+                tab_index: None,
+                importer_movie: None,
             },
         ));
         clip.set_avm2_class(gc_context, Some(class));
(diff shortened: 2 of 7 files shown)


core/src/library.rs
```diff
@@ -215,6 +225,12 @@ impl<'gc> MovieLibrary<'gc> {
         if let Some((id, character)) = self.character_by_export_name(export_name) {
             self.instantiate_display_object(id, character, mc)
         } else {
+            self.export_characters.iter().for_each(|(name, id)| {
+                tracing::error!("Exported character: {} -> {}", name, id);
```
kmeisthax, line 229: I'm assuming this is for debugging purposes, since it doesn't look like a normal error condition. You may want to remove these trace lines?

core/src/library.rs
```diff
@@ -215,6 +225,12 @@ impl<'gc> MovieLibrary<'gc> {
...
+            });
+
+            tracing::error!("registerd exports {}", self.export_characters.len());
```
kmeisthax, line 232: Same thing here.

core/src/loader.rs
```diff
@@ -350,6 +350,87 @@ impl<'gc> LoadManager<'gc> {
...
+                                .expect("Could not load movie");
+
+                            let _is_movie_as3 = movie.is_action_script_3();
+
+                            let movie = Arc::new(movie);
+
+                            player.lock().unwrap().mutate_with_update_context(|uc| {
+                                let stage_domain = uc.avm2.stage_domain();
+                                let mut activation =
+                                    Avm2Activation::from_domain(uc.reborrow(), stage_domain);
```
kmeisthax, line 396: If the loaded movie isn't AS3, shouldn't we be working with an AVM1 activation? I don't think you've written an AVM1 version of the import assets code yet, and I have no clue what FP would do if an AS3 movie tried to import AS2 assets (it'd probably fail), but it should at least detect the different version and print an error or something other than blindly trying to execute with AVM2.

core/src/display_object/movie_clip.rs
```diff
@@ -188,6 +188,9 @@ pub struct MovieClipData<'gc> {
 
     /// Attached audio (AVM1)
     attached_audio: Option<NetStream<'gc>>,
+
+    // If this movie was loaded from ImportAssets(2), this will be the parent movie.
+    importer_movie: Option<Arc<SwfMovie>>,
```
kmeisthax, line 193: As far as I can tell, the motivating idea for this field is that when A.swf imports B.swf, and then B.swf exports assets, B.swf's root `MovieClip` can start copying asset definitions into A.swf's library. This does seem like it'd work, though I'm not entirely sure if managing import/export relationships like this is the correct solution.

For example: if later on we add C.swf to our movie network, and C.swf imports B.swf again, the way this is currently implemented, we have to redownload and reexecute a second copy of the movie, creating two copies of every asset that won't compare equal. Flash Player might actually work this way, but if it didn't, then this solution wouldn't work. We'd need to have importer/exporter relationships centralized somewhere (probably in `Library`).

kmeisthax (commented):
What you have so far largely feels correct, though I absolutely do think this needs some test cases, if only to rule out some of the speculation I put in the comments.

core/src/loader.rs
wsxarcher: As far as I know both games are AS2 not AS3. I did not find any AS3 game using Import.
But if I do the activation of the avm1 it gets stuck completely so I skipped it.

core/src/display_object/movie_clip.rs
wsxarcher: Yes you guessed right. Unfortunately those stuff are not in avmplus and a harder to get it right. Having working games is the main priority for me, a future refactoring is definitely not excluded but I would say out of scope for this pr

wsxarcher:
> What you have so far largely feels correct, though I absolutely do think this needs some test cases, if only to rule out some of the speculation I put in the comments.

Testing can be tricky because of the async nature of the loading. I think flash locks the all vm to load something but in our case we cannot do it. So it works on these games because the loading is in the menu but if you make an artificial swf you need to put some arbitrary sleep and I'm not sure how to do it

core/src/display_object/movie_clip.rs
```diff
@@ -4156,36 +4318,94 @@ impl<'gc, 'a> MovieClipData<'gc> {
...
+    }
+
     #[inline]
     fn export_assets(
         &mut self,
         context: &mut UpdateContext<'_, 'gc>,
         reader: &mut SwfStream<'a>,
     ) -> Result<(), Error> {
         let exports = reader.read_export_assets()?;
+        //return Ok(());
```
Dinnerbone, line 4381: Nit: debug, remove this

core/src/display_object/movie_clip.rs
```diff
@@ -4156,36 +4318,94 @@ impl<'gc, 'a> MovieClipData<'gc> {
...
-                        .0
-                        .read()
-                        .static_data
-                        .exported_name
-                        .write(context.gc_context) = Some(name);
+
+            let character = self.get_registered_character_by_id(context, export.id)?;
+
+            self.register_export(context, export.id, &name, self.movie());
+            tracing::warn!("register_export asset: {} (ID: {})", name, export.id);
```
Dinnerbone, line 4389: Left over debugging? Make it `debug` or remove it

core/src/display_object/movie_clip.rs
```diff
@@ -4156,36 +4318,94 @@ impl<'gc, 'a> MovieClipData<'gc> {
...
+
+            if self.importer_movie.is_some() {
+                let parent = self.importer_movie.as_ref().unwrap().clone();
+                let parent_library = context.library.library_for_movie_mut(parent.clone());
+
+                if let Some(id) = parent_library.character_id_by_import_name(name) {
+                    parent_library.register_character(id, character);
+
+                    self.register_export(context, id, &name, parent);
+                    tracing::warn!(
```
Dinnerbone, line 4399: `debug` or maybe `info` if you think it's useful enough

core/src/display_object/movie_clip.rs
```diff
@@ -4156,36 +4318,94 @@ impl<'gc, 'a> MovieClipData<'gc> {
...
                 }
-            } else {
-                tracing::warn!(
-                    "Can't register export {}: Character ID {} doesn't exist",
-                    name,
-                    export.id,
-                );
             }
+
+            tracing::warn!("Exporting asset: {} (ID: {})", name, export.id);
```
Dinnerbone, line 4408: Debug or remove

core/src/display_object/movie_clip.rs
Dinnerbone: Flash *does* seem to work this way. From my testing, two imports to the same URL will trigger a download twice - which means it's *theoretically* possible to serve a different movie each time and get a different result.

I think we can probably pretend that's not the case and cache it in the future though, somehow

core/src/display_object/movie_clip.rs
```diff
@@ -4120,24 +4202,104 @@ impl<'gc, 'a> MovieClipData<'gc> {
...
+
+        let asset_url = url.to_string_lossy(UTF_8);
+
+        let request = Request::get(asset_url);
+
+        for asset in exported_assets {
+            let name = asset.name.decode(reader.encoding());
+            let name = AvmString::new(context.gc_context, name);
+            let id = asset.id;
+            tracing::warn!("Importing asset: {} (ID: {})", name, id);
```
Dinnerbone, line 4294: Another debug

core/src/display_object/movie_clip.rs
```diff
@@ -4156,36 +4318,94 @@ impl<'gc, 'a> MovieClipData<'gc> {
...
+        if let Some(character) = library.character_by_id(id) {
+            if let Character::MovieClip(movie_clip) = character {
+                *movie_clip
+                    .0
+                    .read()
+                    .static_data
+                    .exported_name
+                    .write(context.gc_context) = Some(*name);
+            } else {
+                tracing::warn!(
```
Dinnerbone, line 4359: Debug/info

Dinnerbone (changes requested):

core/src/display_object/movie_clip.rs
```diff
@@ -315,6 +321,75 @@ impl<'gc> MovieClip<'gc> {
         MovieClipWeak(GcCell::downgrade(self.0))
     }
 
+    pub fn new_import_assets(
+        activation: &mut Avm2Activation<'_, 'gc>,
```
Lord-McSweeney, line 325: Not blocking, but since `new_import_assets` doesn't need any state of the Activation, it should take an UpdateContext and construct the Activation itself. Also, it can be a `from_nothing` Activation, it doesn't need to be `from_domain`.

core/src/loader.rs
```diff
@@ -350,6 +350,87 @@ impl<'gc> LoadManager<'gc> {
         loader.movie_loader(player, request, loader_url)
     }
 
+    async fn wait_for_full_response(
```
adrian17, line 353: couldn't this use existing `Loader::wait_for_full_response`?

core/src/display_object/movie_clip.rs
```diff
@@ -315,6 +321,75 @@ impl<'gc> MovieClip<'gc> {
...
+                queued_tags: HashMap::new(),
+                attached_audio: None,
+                importer_movie: Some(parent.clone()),
+            },
+        ));
+
+        mc
+    }
+
+    pub fn get_importer_movie(&self) -> Option<Arc<SwfMovie>> {
```
adrian17, line 389: unused?
(if used, let's drop the `get_` prefix)

core/src/display_object/movie_clip.rs
```diff
@@ -315,6 +321,75 @@ impl<'gc> MovieClip<'gc> {
...
 
+    pub fn new_import_assets(
+        activation: &mut Avm2Activation<'_, 'gc>,
+        movie: Arc<SwfMovie>,
+        parent: Arc<SwfMovie>,
+    ) -> Self {
+        let num_frames = movie.num_frames();
+
+        let loader_info = if movie.is_action_script_3() {
+            // The root movie doesn't have a `Loader`
```
adrian17, line 332: This was copied from `fn player_root_movie`, right? I don't think the comment is relevant here.

But also, in general - is setting the `loader_info` actually observable here in any way? Could this work if the instantiation was much simpler, closer to `new_with_data`?

core/src/display_object/movie_clip.rs
wsxarcher: I removed the loader info therefore also the activation

core/src/display_object/movie_clip.rs
wsxarcher: Removed

core/src/display_object/movie_clip.rs
wsxarcher: I removed the loader info as it's not used

core/src/loader.rs
```diff
@@ -343,6 +343,64 @@ impl<'gc> LoadManager<'gc> {
...
+            match Loader::wait_for_full_response(fetch).await {
+                Ok((body, url, _status, _redirected)) => {
+                    let content_type = ContentType::sniff(&body);
+                    tracing::info!("Loading imported movie: {:?}", url);
+                    match content_type {
+                        ContentType::Swf => {
+                            let movie = SwfMovie::from_data(&body, url.clone(), Some(url.clone()))
+                                .expect("Could not load movie");
+
+                            let _is_movie_as3 = movie.is_action_script_3();
```
Dinnerbone, line 367: Unused?

Dinnerbone (commented):
One minor nit from me otherwise looks good, I'd like to get this in and then I want to work on a followup soon for testing + more edge cases

Dinnerbone (approved):

core/src/display_object/movie_clip.rs
```diff
@@ -4063,6 +4209,59 @@ impl<'gc, 'a> MovieClipData<'gc> {
...
+        if let Some(character) = library.character_by_id(id) {
+            if let Character::MovieClip(movie_clip) = character {
+                *movie_clip
+                    .0
+                    .read()
+                    .static_data
+                    .exported_name
+                    .write(context.gc_context) = Some(*name);
+            } else {
+                tracing::debug!(
```
Lord-McSweeney, line 4250: This `debug!` feels like it should be a `warn!` or at least `info!` unless it's common enough that that would be cause spam

core/src/loader.rs
```diff
@@ -343,6 +343,62 @@ impl<'gc> LoadManager<'gc> {
...
+                                tracing::debug!(
+                                    "Preloaded swf to run exports result {:?} {}",
+                                    url,
+                                    res
+                                );
+                            });
+                            Ok(())
+                        }
+                        _ => {
+                            tracing::warn!("Unsupported content type: {:?}", content_type);
```
Lord-McSweeney, line 389: Can this log message note that this was an unsupported content type specifically for an assets import?

core/src/avm1/property_map.rs
```diff
@@ -106,6 +106,10 @@ impl<'gc, V> PropertyMap<'gc, V> {
             self.0.shift_remove(&CaseInsensitive(key.as_ref()))
         }
     }
+
+    pub fn len(&self) -> usize {
```
Lord-McSweeney, line 110: Is this leftover or used somewhere?

core/src/display_object/movie_clip.rs
```diff
@@ -4063,6 +4209,59 @@ impl<'gc, 'a> MovieClipData<'gc> {
         Ok(())
     }
 
+    #[inline]
+    fn get_registered_character_by_id(
+        &mut self,
+        context: &mut UpdateContext<'_, 'gc>,
+        id: CharacterId,
+    ) -> Result<Character<'gc>, Error> {
```
Lord-McSweeney, line 4217: Can this return an `Option<Character>` instead, and log an error if the character isn't found?

core/src/display_object/movie_clip.rs
Dinnerbone: I asked him to change it to debug or info before 😅 whoops

PR #16424, opened 2024-05-22
Add HTTP Referer and Cookie

wsxarcher:
Add HTTP Referer and Cookie on desktop as CLI and UI options.
Useful to run 4399.com games which requires the Referer of the website.

Example:

```
ruffle TankTroublev40.swf --referer "https://www.4399.com/" --spoof-url "https://www.4399.com/flash/137953_4.htm" --base "https://sda.4399.com/4399swf/upload_swf/ftp14/cwb/20140611/y2/"
```


frontend-utils/src/player_options/read.rs
```diff
@@ -57,6 +57,12 @@ pub fn read_player_options<'a>(
     // Spoof Url
     result.spoof_url = table.parse_from_str(cx, "spoof_url");
 
+    // Referer
+    result.referer = table.parse_from_str(cx, "referer");
+
+    // Cookie
+    result.cookie = table.parse_from_str(cx, "cookie");
+
```
sleepycatcoding, line 65: I am not sure if we want these fields in bundle format just yet..

WDYT @Dinnerbone?

Dinnerbone: Hmmmm, okay. @wsxarcher maybe for now remove the deserialization and we can follow up on that later

wsxarcher: done!

PR #16426, opened 2024-05-22
core: Register character 0 with library in Loader

### LINKED ISSUE #14639: King's Island - White Screen after Play
### Describe the bug

King's Island has 3 parts in total. Each of it is not loading. I guess they share the same error, thou no error is thrown? What's wrong here?

### Expected behavior

Load & Run

### Content Location

https://www.spiele-umsonst.de/azad/downloads/flashgames/kingsisland.swf

### Affected platform

Online demo

### Operating system

All

### Browser

All

### Additional information

_No response_

Aaron1011:
This allows ActionScript code to instantiate the AVM2 class for the root MovieClip of a loaded movie. This instance will have its frames run, including timeline-placed children.

We now store `avm2_class` in `MovieClipStatic`, so that processing a `SymbolClass` tag from a loaded MovieClip will update the class used by the MovieClip stored in the Library.

Fixes #14639


Lord-McSweeney:
This is technically wrong- the root MovieClip of a loaded movie doesn't always have to be registered under character id #0 (see #12823)- it only has to be registered under a character id that no other character claims. Can the registration logic not hard-code id #0 as the character id that the root MC is registered under?

Aaron1011:
Are you sure? SymbolClass only links the AVM2 class name with a sprite id. Registering the loaded clip as character 0 will only have an effect if there's an explicit `SymbolClass` referencing character 0.

Lord-McSweeney:
The code is all correct, for our current implementation. However, FP allows linking the root MovieClip with a character id other than 0. If we fix that bug (like the PR I linked does), we'll also need to update this logic:
```
                    library.register_character(
                        0,
                        // Use 'instantiate' to clone movie clip data, so future instantiations
                        // don't reflect changes made to the loaded main timeline instance.
                        Character::MovieClip(mc.instantiate(gc_context).as_movie_clip().unwrap()),
                    );
```
to use the correct character id instead of always using character id 0.

PR #16433, opened 2024-05-23
core: Store MovieClip avm2_class in MovieClipStatic

Aaron1011:
This is preparation for allowing a class to be linked to the root movie clip and instantiated.

### DIFF
--- a/core/src/display_object/movie_clip.rs
+++ b/core/src/display_object/movie_clip.rs
@@ -159,7 +159,6 @@ pub struct MovieClipData<'gc> {
     frame_scripts: Vec<Option<Avm2Object<'gc>>>,
     #[collect(require_static)]
     flags: MovieClipFlags,
-    avm2_class: Option<Avm2ClassObject<'gc>>,
     #[collect(require_static)]
     drawing: Drawing,
     avm2_enabled: bool,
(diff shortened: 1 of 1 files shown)


core/src/display_object/movie_clip.rs
```diff
@@ -4611,6 +4611,7 @@ struct MovieClipStatic<'gc> {
     /// The last known symbol name under which this movie clip was exported.
     /// Used for looking up constructors registered with `Object.registerClass`.
     exported_name: GcCell<'gc, Option<AvmString<'gc>>>,
+    avm2_class: GcCell<'gc, Option<Avm2ClassObject<'gc>>>,
```
kmeisthax, line 4614: This will be the fourth GcCell in `MovieClipStatic`. And AFAIK there's also a preload step that copies the `MovieClipStatic` to change some of the other fields. As we find all these partial/delayed initialization cases I'm starting to wonder when it would make sense to make this mutable and call it `MovieClipShared` or `MovieClipDefinition` instead...

kmeisthax (approved):
Looks pretty straightforward, and it makes more conceptual sense for the class to live in `MovieClipStatic`.

core/src/display_object/movie_clip.rs
Aaron1011: That sounds reasonable - I can do that in a follow-up pr

PR #16453, opened 2024-05-25
avm2: Implement XMLList to XML proxy methods

sleepycatcoding:


evilpie (approved):

adrian17:
Hmm, how does it differ from the logic in https://github.com/ruffle-rs/ruffle/blob/bafc5d250e1d1c78da406cb9c0434fc9d0ce42e7/core/src/avm2/object/xml_list_object.rs#L571 ?
Don't they overlap at least partially?

sleepycatcoding:
> Hmm, how does it differ from the existing proxy logic in
> 
> https://github.com/ruffle-rs/ruffle/blob/bafc5d250e1d1c78da406cb9c0434fc9d0ce42e7/core/src/avm2/object/xml_list_object.rs#L573
> ?

```actionscript
var list = new XMLList("<a>B</a>");

// Compiles to callproperty (falls into existing proxy logic)
trace(list.nodeKind());

// Compiles to getproperty and call (fails without this PR)
trace(list.AS3::nodeKind());
```

Issue #15009, opened 2024-01-29
Does not load - https://www.4399.com/flash/205551_4.htm


wsxarcher:
### Describe the bug

Game keeps loading with black screen and loading circle

```
VM117:1  Uncaught TypeError: Cannot read properties of null (reading 'cloneNode')
    at <anonymous>:1:552
    at <anonymous>:1:945
```

### Expected behavior

Load the game

### Content Location

https://www.4399.com/flash/205551_4.htm

### Affected platform

Browser's extension

### Operating system

Window 11

### Browser

Edge/Chrome 120

### Additional information

This game is part of a bigger effort to support 4399.com games.
There might be other issues with this website, from how it handles Flash checks and many CORS issues.
Unfortunately, some games cannot be easily debugged on desktop because the swf download other swf or do requests while logged in.

**If Ruffle is not detected try to reload the page multiple times because of the known race condition on the Flash check. To avoid refreshing try this Tampermonkey user script [一键下载 flash 游戏 (greasyfork.org)](https://greasyfork.org/en/scripts/445006-%E4%B8%80%E9%94%AE%E4%B8%8B%E8%BD%BD-flash-%E6%B8%B8%E6%88%8F)**

To solve these issues refers to previous issues such as:

- https://github.com/ruffle-rs/ruffle/issues/13943#issuecomment-1835074616 - Race condition discussion
- https://github.com/ruffle-rs/ruffle/issues/14578#issuecomment-1873523939 - Ruffle detection issue
- Edge/Chrome flag to temporarily disable CORS `--disable-web-security` or use https://chromewebstore.google.com/detail/allow-cors-access-control/lhobafahddgcelffkeicbaginigeejlf
- Disable antidebugging with https://greasyfork.org/en/scripts/440060-anti-anti-debugger/code or Deactivate breakpoint
- Disable second antidebugging manually 
![fadb9686-1ce4-406b-9694-42b97842ee35](https://github.com/ruffle-rs/ruffle/assets/519077/bcf15e50-fc7a-4a31-a098-47be6d8b68c8)
- When running the downloaded .swf from desktop to debug, use `--referer "https://www.4399.com"`, the correct `--spoof-url` (usually the page link) and `--base` which is the relative URL where other resources are fetched.

wsxarcher:
Exception similar to number 2 of https://github.com/ruffle-rs/ruffle/issues/15013#issuecomment-2131575826

```
[...]

Issue #15013, opened 2024-01-29
Does not load - https://www.4399.com/flash/198660_4.htm 


wsxarcher:
### Describe the bug

Black screen after 4399 loading
```
VM244896:1  Uncaught TypeError: Cannot read properties of null (reading 'cloneNode')
    at <anonymous>:1:552
    at <anonymous>:1:945
```

### Expected behavior

Play the game

### Content Location

https://www.4399.com/flash/198660_4.htm 

### Affected platform

Browser's extension

### Operating system

Window 11

### Browser

Edge/Chrome 120

### Additional information

This game is part of a bigger effort to support 4399.com games.
There might be other issues with this website, from how it handles Flash checks and many CORS issues.
Unfortunately, some games cannot be easily debugged on desktop because the swf download other swf or do requests while logged in.

**If Ruffle is not detected try to reload the page multiple times because of the known race condition on the Flash check. To avoid refreshing try this Tampermonkey user script [一键下载 flash 游戏 (greasyfork.org)](https://greasyfork.org/en/scripts/445006-%E4%B8%80%E9%94%AE%E4%B8%8B%E8%BD%BD-flash-%E6%B8%B8%E6%88%8F)**

To solve these issues refers to previous issues such as:

- https://github.com/ruffle-rs/ruffle/issues/13943#issuecomment-1835074616 - Race condition discussion
- https://github.com/ruffle-rs/ruffle/issues/14578#issuecomment-1873523939 - Ruffle detection issue
- Edge/Chrome flag to temporarily disable CORS `--disable-web-security` or use https://chromewebstore.google.com/detail/allow-cors-access-control/lhobafahddgcelffkeicbaginigeejlf
- Disable antidebugging with https://greasyfork.org/en/scripts/440060-anti-anti-debugger/code or Deactivate breakpoint
- Disable second antidebugging manually 
![fadb9686-1ce4-406b-9694-42b97842ee35](https://github.com/ruffle-rs/ruffle/assets/519077/bcf15e50-fc7a-4a31-a098-47be6d8b68c8)
- When running the downloaded .swf from desktop to debug, use `--referer "https://www.4399.com"`, the correct `--spoof-url` (usually the page link) and `--base` which is the relative URL where other resources are fetched.

Dinnerbone:
Works in desktop, assuming this is ExternalInterface stubbiness:

1. `objectID` should be set
2. It does lots of `call("new function(){...}()")` which we don't support

[...]

PR #16505, opened 2024-05-27
web: Support ExternalInterface.call("function(){...}")

Dinnerbone:
Plus many tests for ExternalInterface on web

~~Depends on #16500 (drafted until that's merged)~~

Progresses #8330 (I assume)
Progresses #15013


sammypanda:
this is awesome (^^) thank u :heart:

evilpie:
I am not sure if this would pass the add-on review for addons.mozilla.org.

Dinnerbone:
> I am not sure if this would pass the add-on review for addons.mozilla.org.

Hmm, that's unfortunate. Do you know which guideline this would be violating and how we can fix it?

evilpie:
You are definitely going to get an automatic warning message. I think it might be possible to explain why the `eval` exists and why it is necessary. Some other extensions like content script messengers also use something like `eval` and can get accepted. Technically running Flash content is like eval but with extra steps :smile:

torokati44:
Given the risk factors mentioned above, and experienced elsewhere, shall we wait with this until the next AMO submission at least?

wsxarcher:
It also fix
https://github.com/ruffle-rs/ruffle/issues/15009
https://github.com/ruffle-rs/ruffle/issues/16482

web/packages/core/src/ruffle-imports.ts
```diff
@@ -61,12 +61,15 @@ export function copyToAudioBufferInterleaved(
...
  */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function getProperty(target: any, key: string): any {
-    return target[key];
+// @ts-expect-error defined but not used
+// eslint-disable-next-line @typescript-eslint/no-explicit-any,@typescript-eslint/no-unused-vars
+export function callExternalInterface(name: string, args: any[]): any {
+    // [NA] Yes, this is direct eval. Yes, this is a Bad Thing when it comes to security.
+    // In fact, yes this is vulnerable to an XSS attack!
+    // But plot twist: Flash allowed for this and many content *relies on it*. :(
```
Lord-McSweeney, line 73: "many content" typo?

Dinnerbone: I don't think so?

danielhjacobs: I'd say much content or many types of content (or some other grouping), content is treated as an uncountable noun. Alternatively, a lot of content.

torokati44:
Last week's AMO submission has been accepted, so apparently this didn't trip any extra alarms during review ... so far.

PR #16523, opened 2024-05-29
web: Make embed instead of Ruffle instance with the API when using Flash

danielhjacobs:
This should work, but it's arguable if it's something we should be doing. It's still possible to use the Ruffle API without it creating a Flash embed on Flash-enabled browsers by setting the favorFlash configuration option to false.

If you load the SWF as data instead of by URL, we still use Ruffle since Flash embeds don't have a way to support that.

### DIFF
--- a/web/packages/core/src/polyfills.ts
+++ b/web/packages/core/src/polyfills.ts
@@ -25,7 +25,7 @@ let embeds: HTMLCollectionOf<HTMLEmbedElement>;
  *
  * @returns Whether this browser has a plugin indicating pre-existing Flash support.
  */
-function isFlashEnabledBrowser(): boolean {
+export function isFlashEnabledBrowser(): boolean {
     // If the user sets a configuration value not to favor Flash, pretend the browser does not support Flash so Ruffle takes effect.
     if ("favorFlash" in globalConfig && globalConfig["favorFlash"] === false) {
         return false;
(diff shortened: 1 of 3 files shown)


danielhjacobs:
Things seem to be leaning towards not landing this PR. I can't say I necessarily disagree. The Ruffle API already has a few bits of functionality that differ from Flash. For example, you can do `player.volume = 0.5;` to set the volume of the SWF to half its initial value (on a logarithmic scale akin to human hearing). Another example is that when Ruffle is done loading an SWF it fires the `loadeddata` event, and some webmasters therefore wait for that event before doing something else. Additionally, there are plans to version the API and add even more QOL functionality that did not exist for Flash Player. If we fell back to a Flash embed some of a webmaster's code may not do anything if they used the API for functionality that didn't exist in Flash. Also, there is a configuration option, `playerRuntime`, which when set to "air" is meant to allow you to embed an AIR SWF in the browser. Currently, we still don't support most AIR APIs, but there are plans to improve that support, and if someone used the API to load an AIR SWF in the future, even Flash-enabled browsers wouldn't be able to load it without Ruffle.

Instead, we should just emphasize that webmasters should use the embed/object elements with the polyfill if their only purpose in using Ruffle is to embed Flash, which we already mention as the preferred method for non-advanced users.

PR #16531, opened 2024-05-30
 avm2: Parse XML namespace declaration and make ToXMLString output them 

evilpie:
This does not implement any of the XML API functions that interact with namespaces. It's primarily meant to produce XML ToString results that are again valid XML with namespaces.

In the specification `[[InScopeNamespaces]]` is a property of all nodes (not just elements), but only elements can introduce namespace declarations so I moved that to `E4XNodeKind::Element`. And similar to @Dinnerbone's WIP commit I also don't store all (ancestor) namespaces, but instead just calculate them in `in_scope_namespaces`.


Dinnerbone (approved):
<3

PR #16532, opened 2024-05-30
ci: Ignore just the Docker package files in the reproducible build

danielhjacobs:


torokati44:
I'm almost certain that this is a good solution, but at this point, I'm afraid of actually pushing the green button before quadruple cross-checking with an independent verification board of 7 independent experts, and an omniscient entity with transcendental consciousness. 🫠

torokati44:
Maybe the `approval_notes` in `submit_xpi.js` should also be amended to `rm` only this folder instead of `docker_builds/*`. It shouldn't matter in practice (as the `docker_builds` folder will still be there, empty), just for consistency's sake.

torokati44:
This would've been also useful: https://github.com/moby/moby/issues/20920#issuecomment-1710778527
But oh well: https://github.com/moby/moby/pull/20927

danielhjacobs:
Approval notes updated.

danielhjacobs:
Since that's a comment on a closed issue, opened https://github.com/moby/moby/issues/47880 as a new issue.

Dinnerbone (approved):
Thanks for dealing with this!

danielhjacobs:
Well, nothing to do but to merge and see I suppose.

PR #16536, opened 2024-05-30
avm2: Implement XMLList.normalize

sleepycatcoding:
As usual, adds a test for both `XML.normalize` and `XMLList.normalize`.


Lord-McSweeney:
Hmm, does this not allow any avmplus tests to pass?

sleepycatcoding:
> Hmm, does this not allow any avmplus tests to pass?

The avmplus test for this is `XMLList/e13_5_4_15`, unfortunately they tested with a single element lists, so it fell through to the XMLObject's normalize instead (this test was enabled with the XML's normalize impl)

Dinnerbone (approved):

Issue #13290, opened 2023-09-24
Make FLV Handler send playback status to the SWF


TemaisgameNoobas:
**Describe the feature**
Provide a clear and concise description of your feature request. Why would this feature be useful? Does the feature involve the desktop build, the web build, or both?

This feature is useful, because some SWFs may trigger functions when the FLV isn't playing. The FLV Handler actually plays the video fine, but the SWF thinks that the video is not playing. This feature involves both builds.

TemaisgameNoobas:
should be fixed by #16556 !!

PR #16556, opened 2024-06-01
avm1: Stub `bufferLength` and `bufferTime` on `NetStream`

### LINKED ISSUE #12934: A sign appears every 10 - 15 sec (Neave.TV 2006)
### Describe the bug

Hi everyone. There is no steps to reproduce the bug. Just launch the SWF, click on the sign in the center and it would appear. Here is the screenshot of it: 
image

The sign appears every 10 - 15 seconds, and after it appears it switches the video. If I press on "Continue watching anyway" a new video starts playing. I am sure the bug is in the SWF, but in the original Flash Player the sign does not appear. If I have any more info, I will edit the issue or post it in the comments.
Edit: The files are here: #11821 

### Expected behavior

The sign not appearing and not switching videos, as the videos play fine actually.

### Affected platform

Self-hosted version

### Operating system

Android 12

### Browser

Opera 76.2.4027.73374

### Additional information

_No response_

Lord-McSweeney:
This should fully fix #12934.

`bufferLength` always reports that the buffer is fully filled, while `bufferTime` can be set but has no effect.

### DIFF
--- a/core/src/streams.rs
+++ b/core/src/streams.rs
@@ -218,6 +218,10 @@ pub struct NetStreamData<'gc> {
     /// Seeks are only executed on the next stream tick.
     queued_seek_time: Option<f64>,
 
+    /// The number of seconds of video data that should be buffered. This is
+    /// currently unsupported and changing it has no effect.
+    buffer_time: f64,
+
     /// The last decoded bitmap.
     ///
     /// Any `Video`s on the stage will display the bitmap here when attached to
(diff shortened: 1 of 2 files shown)


core/src/streams.rs
```diff
@@ -218,6 +218,8 @@ pub struct NetStreamData<'gc> {
     /// Seeks are only executed on the next stream tick.
     queued_seek_time: Option<f64>,
 
+    buffer_time: f64,
```
adrian17, line 221: can you also document it as todo-currently-unused?

adrian17 (approved):

core/src/streams.rs
TemaisgameNoobas: > can you also document it as todo-currently-unused?

I think it requires buffer_time

TemaisgameNoobas:
should fix #13290

PR #16561, opened 2024-06-01
avm2: Implement a few avmplus XML edge cases

sleepycatcoding:


evilpie (approved):
Wow mind mending stuff, but looks good to me.

PR #16580, opened 2024-06-03
avm1: fix loadvars.send parameters

TheToto:
The `LoadVars.send` signature is `send(url:String, target:String, [method:String]):Boolean`

There was a small mistake in the code that read the second parameter for target AND method.


Dinnerbone (approved):
Nice catch! Clearly we're missing a test here 😅

PR #16587, opened 2024-06-04
avm1: Drop focus on EditText when something else is clicked

kjarosh:
In AVM1 text fields are somewhat special when handling focus. When a text field is clicked, it gains focus, when something else is clicked, it loses the focus. However, this logic only applies to text fields, other objects (buttons, movie clips) neither gain focus nor lose it upon press.

This behavior might have appeared inconsistent, as some SWFs lose focus on EditText when something else is clicked, some don't. Turns out that difference results from `mx.managers.FocusManager`. It automatically focuses the last focused EditText after click, so that it seemed that the field did not lose the focus. In reality it lost the focus on press, and gained it again on release. Since `mx.managers.FocusManager` now works (https://github.com/ruffle-rs/ruffle/issues/5443), Ruffle will mirror this behavior when this PR is merged.


Dinnerbone (approved):
Amazing!

PR #16591, opened 2024-06-04
avm2: Support namespaces in XML.setName

evilpie:


danielhjacobs:
It looks like from_avmplus/as3/RuntimeErrors/Error1117InvalidXmlName is passing now as well.

Dinnerbone (approved):
Thank you!

PR #16594, opened 2024-06-04
avm2: In Array.sortOn, behave better in presence of nonsense args

adrian17:
This fixes an issue originally reported on Discord (https://discord.com/channels/610531541889581066/610531541889581068/1239687519734599710, site was https://roguemon.com/) , I don't think there's a matching GH issue.

The SWF had code like
```
loadingTexts.sortOn(randomSort);

function randomSort(a:*, b:*):Number
{
    if (Math.random() < 0.5) return -1;
    else return 1;
}
```
The author probably didn't realize that `sortOn(function)` isn't supported (you're supposed to use `sort(function)`); it turns out that in FP it doesn't throw, and because of how FP's sort routine works, it _does_ reorder items in some... possibly nonsense way.
This PR doesn't make the sorting identical (that'd require us to copy FP's qsort algorithm, which is what we already did in avm1), it just prevents us from throwing a nonsense error.


adrian17:
(Turns out that the test passes before the PR, I think I need to use a class instead of a bare object in the test in order to induce the observable throw. Will fix that in 2-3h.)

adrian17:
Fixed, good to review now.

Dinnerbone (approved):

PR #16610, opened 2024-06-05
avm2: Minor optimizer improvements

Lord-McSweeney:
This is mostly just improvements to the `int` type in the optimizer.

Box2d stats:
```
FindPropStrict -> GetScriptGlobals : 62.18%
FindPropStrict -> GetOuterScope : 7.76%
FindPropStrict -> GetScopeObject : 8.24%

FindProperty -> GetOuterScope : 2.83%
FindProperty -> GetScopeObject : 94.34%

InitProperty -> SetSlot : 40.93%
InitProperty -> SetSlotNoCoerce : 44.51%

SetProperty -> SetSlot : 25.52% (-3.15%)
SetProperty -> SetSlotNoCoerce : 42.66% (+3.15%)
SetProperty -> CallMethod : 1.57%

GetProperty -> GetSlot : 67.31%
GetProperty -> CallMethod : 0.34%

CallProperty -> CallMethod : 41.13%
CallPropVoid -> CallMethod : 74.24%

Coerce -> Nop : 49.78%

CoerceD -> Nop : 81.34%

CoerceB -> Nop : 8.11%

CoerceU -> Nop : 50% (+17.54%)

CoerceI -> Nop : 33% (+31%)

ReturnValue -> ReturnValueNoCoerce : 71.56% (+5.50%)
```

### DIFF
--- a/core/src/avm2/optimize.rs
+++ b/core/src/avm2/optimize.rs
@@ -11,6 +11,14 @@ use crate::avm2::vtable::VTable;
 use gc_arena::Gc;
 use std::collections::HashMap;
 
+#[allow(clippy::enum_variant_names)]
+#[derive(Clone, Copy, Debug)]
+enum NullState {
+    NotNull,
+    MaybeNull,
+    IsNull,
+}
+
 #[derive(Clone, Copy)]
 struct OptValue<'gc> {
     // This corresponds to the compile-time assumptions about the type:
@@ -33,10 +41,9 @@ struct OptValue<'gc> {
     // should only be set if class is numeric.
     pub contains_valid_unsigned: bool,
 
-    // true if value is guaranteed to be null.
     // TODO: FP actually has a separate `null` type just for this, this can be observed in VerifyErrors
     // (a separate type would also prevent accidental "null int" values)
-    pub guaranteed_null: bool,
+    pub null_state: NullState,
 }
 impl<'gc> OptValue<'gc> {
     pub fn any() -> Self {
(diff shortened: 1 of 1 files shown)


core/src/avm2/optimize.rs
```diff
@@ -563,6 +559,8 @@ pub fn optimize<'gc>(
                         || value2.class == Some(types.number))
                 {
                     stack.push_class(types.number);
+                } else if value1.class == Some(types.string) || value2.class == Some(types.string) {
```
adrian17, line 562: I don't think it's always correct:
```
            var a: String = null;
            var b: String = null;
            var c = a + b;
            trace(c); // 0
            trace(getQualifiedClassName(c)); // int
```
In FP's verifier, this optimization is present, but also guarded by `notNull` abstract value metadata checks (which we don't model currently, we only model "guaranteed null".)

core/src/avm2/optimize.rs
```diff
@@ -401,8 +401,7 @@ pub fn optimize<'gc>(
             }
             Op::CoerceI => {
                 let stack_value = stack.pop_or_any();
-                // TODO: maybe the type check is safe here...?
-                if stack_value.contains_valid_integer {
+                if stack_value.class == Some(types.int) {
```
adrian17, line 404: This prevents optimizing `pushuint 5; coerce_i`, right? Could we check "type is int or value is statically known to be integral" instead?

adrian17 (changes requested):

core/src/avm2/optimize.rs
adrian17: We could add a trivial `guaranteed_not_null` that's `true` for `pushstring` opcodes, which would allow type propagation for `"asdf" + whatever`

adrian17 (approved):

PR #16612, opened 2024-06-05
web: Make config parsed by TS, not rust - and lots of other refactorings

Dinnerbone:
The main motivation here is to get all the parsing of config stuff out from Rust and into TypeScript, for a few reasons:
1. Duck typing is what JS/TS is good at - Rust is really, really bad at it.
2. As we evolve the config, potentially change which values are accepted (eg the duration once went from object to number), we can just translate that in TS and Rust only needs to know the "real", final value
3. We can easily add tests for config parsing, which wasn't quite as simple in Rust
4. We can change what Rust accepts on a whim now, as needed - without changing the actual config interface we expose publicly

Along with this, lots of refactoring to split up `lib.rs` some more. A new `RuffleInstanceBuilder` is responsible for actually constructing the `Player `and `RuffleInstance`.

I've also added a new default log subscriber to capture things that happen outside of a specific ruffle instance, as that's bit us in the butt before.

I feel like I can probably add another 10 more commits and continue the refactor but I'm losing steam, I accomplished what I wanted already :D

### DIFF
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -715,17 +715,13 @@ export class RufflePlayer extends HTMLElement {
             throw e;
         });
         configureBuilder(builder, this.loadedConfig || {});
-        this.instance = await builder.build(this.container, this).catch((e) => {
-            console.error(`Serious error loading Ruffle: ${e}`);
-            this.panic(e);
-            throw e;
-        });
+        builder.setVolume(this.volumeSettings.get_volume());
 
         if (this.loadedConfig?.fontSources) {
             for (const url of this.loadedConfig.fontSources) {
                 try {
                     const response = await fetch(url);
-                    this.instance!.add_font(
+                    builder.addFont(
                         url,
                         new Uint8Array(await response.arrayBuffer()),
                     );
@@ -739,25 +735,29 @@ export class RufflePlayer extends HTMLElement {
         }
 
         if (this.loadedConfig?.defaultFonts?.sans) {
-            this.instance!.set_default_font(
+            builder.setDefaultFont(
                 "sans",
                 this.loadedConfig?.defaultFonts.sans,
             );
         }
         if (this.loadedConfig?.defaultFonts?.serif) {
-            this.instance!.set_default_font(
+            builder!.setDefaultFont(
                 "serif",
                 this.loadedConfig?.defaultFonts.serif,
             );
         }
         if (this.loadedConfig?.defaultFonts?.typewriter) {
-            this.instance!.set_default_font(
+            builder!.setDefaultFont(
                 "typewriter",
                 this.loadedConfig?.defaultFonts.typewriter,
             );
         }
 
-        this.instance!.set_volume(this.volumeSettings.get_volume());
+        this.instance = await builder.build(this.container, this).catch((e) => {
+            console.error(`Serious error loading Ruffle: ${e}`);
+            this.panic(e);
+            throw e;
+        });
 
         this.rendererDebugInfo = this.instance!.renderer_debug_info();
 
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -734,23 +734,15 @@ export class RufflePlayer extends HTMLElement {
             }
         }
 
-        if (this.loadedConfig?.defaultFonts?.sans) {
-            builder.setDefaultFont(
-                "sans",
-                this.loadedConfig?.defaultFonts.sans,
-            );
-        }
-        if (this.loadedConfig?.defaultFonts?.serif) {
-            builder!.setDefaultFont(
-                "serif",
-                this.loadedConfig?.defaultFonts.serif,
-            );
-        }
-        if (this.loadedConfig?.defaultFonts?.typewriter) {
-            builder!.setDefaultFont(
-                "typewriter",
-                this.loadedConfig?.defaultFonts.typewriter,
-            );
+        for (const key in this.loadedConfig?.defaultFonts) {
+            const names = (
+                this.loadedConfig.defaultFonts as {
+                    [key: string]: Array<string>;
+                }
+            )[key];
+            if (names) {
+                builder.setDefaultFont(key, names);
+            }
         }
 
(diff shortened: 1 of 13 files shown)


web/src/builder.rs
```diff
@@ -0,0 +1,631 @@
...
+    #[wasm_bindgen(js_name = "setCredentialAllowList")]
+    pub fn set_credential_allow_list(&mut self, value: Vec<String>) {
+        self.credential_allow_list = value;
+    }
+
+    #[wasm_bindgen(js_name = "setPlayerRuntime")]
+    pub fn set_player_runtime(&mut self, value: &str) {
+        self.player_runtime = match value {
+            "air" => PlayerRuntime::AIR,
+            "flash_player" => PlayerRuntime::FlashPlayer,
```
sleepycatcoding, line 286: Uhh, isn't this a breaking change? (It was `flashPlayer` before)

Dinnerbone: You're right, thank you!

web/packages/core/src/ruffle-player.ts
```diff
@@ -734,25 +735,29 @@ export class RufflePlayer extends HTMLElement {
...
                 this.loadedConfig?.defaultFonts.serif,
             );
         }
         if (this.loadedConfig?.defaultFonts?.typewriter) {
-            this.instance!.set_default_font(
+            builder!.setDefaultFont(
                 "typewriter",
                 this.loadedConfig?.defaultFonts.typewriter,
             );
         }
```
sleepycatcoding, line 754: Is this exclamation syntax here necessary? (Doesn't seem to be used a few lines up)

web/src/builder.rs
```diff
@@ -0,0 +1,631 @@
...
+                                    .register_device_font(FontDefinition::SwfTag(*font, encoding));
+                            }
+                            swf::Tag::DefineFont4(font) => {
+                                let name = font.name.to_str_lossy(encoding);
+                                if let Some(data) = font.data {
+                                    tracing::debug!("Loaded font {name} from font swf {font_name}");
+                                    player.register_device_font(FontDefinition::FontFile {
+                                        name: name.to_string(),
+                                        is_bold: font.is_bold,
+                                        is_italic: font.is_bold,
```
sleepycatcoding, line 365: Shouldn't this be `font.is_italic` instead? (Though this seems to have existed before)

web/src/builder.rs
```diff
@@ -0,0 +1,631 @@
...
+
+    #[wasm_bindgen(js_name = "setDefaultFont")]
+    pub fn set_default_font(&mut self, default_name: &str, fonts: Vec<JsValue>) {
+        let default = match default_name {
+            "sans" => DefaultFont::Sans,
+            "serif" => DefaultFont::Serif,
+            "typewriter" => DefaultFont::Typewriter,
+            _ => {
+                return;
+            }
```
sleepycatcoding, line 309: Kind of surprised that rustfmt does not suggest  `_ => return,` here.

web/packages/core/test/config.ts
```diff
@@ -0,0 +1,30 @@
+import { strict as assert } from "assert";
+import { parseColor, parseDuration } from "../src/internal/builder";
+
+describe("Color parsing", function () {
+    it("should parse a valid RRGGBB hex, with hash", function () {
+        assert.strictEqual(parseColor("#A1B2C3"), 0xa1b2c3);
+    });
+
+    it("should parse a valid RRGGBB hex, without hash", function () {
+        assert.strictEqual(parseColor("#1A2B3C"), 0x1a2b3c);
```
sleepycatcoding, line 10: This should be testing without the hash `#`?

Toad06:
For some reason, it seems there are some issues when applying the global config to polyfilled elements:

```
    "backgroundColor": "#cccccc",
    "wmode": "transparent"
};
```
polyfill

The square on the left is an `embed` element, and the one on the right is an `object` element. As you can see, they produce different results (the `object` element applies the background color set in the global config), but neither of them is actually correct as they do not apply the wmode (which would make the file completely transparent).

[sample.zip](https://github.com/user-attachments/files/15613957/sample.zip)

Dinnerbone:
@Toad06 I can't reproduce, they're all black (transparent?) for me - looks identical to ruffle-master

web/packages/core/src/ruffle-player.ts
Dinnerbone: Fixed; I made it iterate the keys and do it automatically, too. It turns out we forgot to add japanese fonts to this list, which this also fixes.

web/src/builder.rs
Dinnerbone: Nice catch!

web/src/builder.rs
Dinnerbone: Fixed

web/packages/core/test/config.ts
Dinnerbone: Nice catch!

Toad06:
This is my fault, I compiled everything, but then somehow managed to copy the files from the wrong folder to the local server... Well, this is quite embarrassing (and I can't even use the excuse that "it was late"). I'm really sorry for wasting your time. 😳 I'll give the PR another try tomorrow... with the right files.

Toad06:
Testing the PR with the correct files, this looks good to me and also increases consistency across the different options. For instance, `letterbox: "incorrect"` causes a crash in the current builds, whereas `wmode: "incorrect"` falls back to the default value. With this PR, both options revert to their default values when given incorrect inputs.

A panic still occurs if the user provides a value of the wrong type. For instance, `wmode: 3.5` would produce this error:
- in the current builds:
> Error parsing config: Error: invalid type: floating point `3.5`, expected a string
- in this PR:
> RuntimeError: index out of bounds

This is not too important, but I'm not sure if this is still the intended behaviour, or if we should also be lenient in this situation?

By the way, once this PR is merged, I believe #10931 could be closed, as it was using a different approach.

Dinnerbone:
Good to know it works! My hope is to remove all "crash ruffle" states - things should be warnings and ignored, rather than any error condition. I'll see if I can expand on this and add more safeties (and tests) for it.

sleepycatcoding (approved):
LGTM

Issue #15003, opened 2024-01-29
Blurry vector rendering (especially with 1px borders) and font misalignment.


ReFaller:
### Describe the bug

![ruffle2](https://github.com/ruffle-rs/ruffle/assets/39001461/221fc575-4133-434b-a03c-340b475c26da)

1. Please note the blurriness of vector fonts in Ruffle. 
2. When using a system font in the text, it inexplicably shifts slightly upwards. 
3. The effect is very noticeable when rasterizing a 1px rectangular border (standard, around the font, and around a vector square).

### Expected behavior

Rendering feedback.

### Content Location

[test.zip](https://github.com/ruffle-rs/ruffle/files/14087742/test.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

Issue #1881, opened 2020-12-08
Device text should not be transformed


Herschel:
The Flash Player will not render device text as transformed; e.g. `txt._xscale = 200;` will make the textbox longer:
![image](https://user-images.githubusercontent.com/36278/101421363-fae67300-38a8-11eb-966e-089a32c421ef.png)
Note that the display object is still transformed if you look at the matrix, it's only the final rendering that ignores the scaling.

Ruffle actually transforms the text:
![image](https://user-images.githubusercontent.com/36278/101421379-08036200-38a9-11eb-9a06-791aabca6183.png)

This only applies to device text (cc #1862); outline text can be transformed as normal.

[devicetextransform.zip](https://github.com/ruffle-rs/ruffle/files/5656014/devicetextransform.zip)

First reported in #808.

(TODO: I have a vague recollection that device text rotation was a feature added in some version of Flash -- maybe sometime in the AS3 era?)

SecularSteve:
Happens on the loading screen of "Sonic Breaks his Neck" on Newgrounds. An explicit, but classic, animation.

https://www.newgrounds.com/portal/view/297089

![Screenshot 2023-12-13 at 23-11-47 Sonic Breaks his Neck.png](https://github.com/ruffle-rs/ruffle/assets/33793273/f944baef-3943-4b60-8883-18fc6ea02319)

PR #16619, opened 2024-06-06
render,core: Improve EditText's border and background rendering

kjarosh:
This PR removes the drawing used for rendering the border and
the background and renders them using primitives instead.

It provides two methods for drawing the border and the background:

* `draw_device_text_box` -- when device fonts are used,
* `draw_text_box` -- when fonts are embedded.

In order to draw the border properly new methods have been added:

* `draw_line` -- for drawing a single line (always 1px thick),
* `draw_line_rect` -- for drawing a rectangle using lines (always 1px thick).

These primitives have been implemented for wgpu, WebGL, and canvas.

| Before | After |
| ------ | ----- |
|  | <image src="https://github.com/ruffle-rs/ruffle/assets/1507072/e1916372-de57-4df2-b879-99835b612f2e" width="300"/> |
|  | <image src="https://github.com/ruffle-rs/ruffle/assets/1507072/95bd7fce-b8ee-4175-a5c5-74cb2981c679" width="300"/> |
|  | <image src="https://github.com/ruffle-rs/ruffle/assets/1507072/b728fdb7-49c3-4e54-9584-4309f62f4490" width="300"/> |
|  | <image src="https://github.com/ruffle-rs/ruffle/assets/1507072/9a9cbee3-ba26-4fdc-b0b1-34de5fd30803" width="300"/> |

Additional notes:
* `draw_line` added in this PR may also be used for rendering the selection caret in the future.

Related to https://github.com/ruffle-rs/ruffle/issues/15003, https://github.com/ruffle-rs/ruffle/issues/1881 (the border is not being transformed anymore).


Dinnerbone (commented):
Why is the "draw lines or emulate lines" the responsibility of core, and not the backend itself?

kjarosh:
> Why is the "draw lines or emulate lines" the responsibility of core, and not the backend itself?

Originally I thought about making that the responsibility of the backend, but wgpu required quite a refactor to implement that cleanly, and I was frustrated by DirectX at that time :sweat_smile: 

So anyway, I did the refactor now. The whole "drawing lines" thing came out bigger than expected, so if you wish I could split this PR into multiple ones (so that this one does not include the necessary refactors).

Dinnerbone (approved):
Thank you!

Issue #16474, opened 2024-05-27
The Clockmaker: Context menu copy paste issue


waspennator:
### Describe the bug

While it is possible to highlight text and even copy paste by ctrl+c and ctrl+v, there is another issue I noticed. If you highlight the text inside a text box and then right click, it incorrectly shows the context menu options as if you weren't clicking on the text box.

![Screenshot 2024-05-27 015115](https://github.com/ruffle-rs/ruffle/assets/78768499/0da03c07-a4c7-4fe1-810f-5d2e04aa297a)

### Expected behavior

When you left click or highlight text in a text box and then right click inside the text box, something like this should appear in the context menu.

![Screenshot 2024-05-27 015148](https://github.com/ruffle-rs/ruffle/assets/78768499/1858acf0-b294-4f7b-a10f-bee5a32875c2)

### Content Location

https://www.newgrounds.com/portal/view/873121?emulate=flash

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 123

Issue #1891, opened 2020-12-08
Can't select/copy/paste text in input box


Txori:
In Ruffle, a user can't use the mouse to select text inside an input box.
One cannot right-click inside the text input to open the copy / paste tool box.
Ctrl+c and ctrl+v don't work either.

You'll find the stripped down example here: https://www.txori.com/stuff/ruffle/debug/

Poopooracoocoo:
and the cursor keeps flashing when you move it with your arrow keys. you also can't use the up or down keys to navigate, or use other keyboard shortcuts such as ctrl+left/right. (i'm pretty sure there are issues for what i'm saying already)

Txori:
And I forgot about the useful double click to select an entire word!

rkontra:
I would be very excited to see this feature implemented. This is currently hindering the use of our flash application (located here: https://www.uniteddie.com/dct.html

Txori:
I believe it's being worked on here: https://github.com/ruffle-rs/ruffle/pull/2081

waspennator:
Testing with nightly 5/27/24, highlighting the text with the mouse now works and I can copy and paste using ctrl+C and ctrl+V, the only issue that's left now is the right click copy paste tool box.

waspennator:
Context menu for copy pasting should work now, which satisfies all the points in the first post. Ctrl+left or right also seems to work on my end, so the only issues left seem to be double clicking to highlight a word, not being able to use the up and down keys to move in editable text boxes, and another issue i'm seeing with double or triple highlight lines showing up when moving around to certain spots.

![Screenshot 2024-06-09 185225](https://github.com/ruffle-rs/ruffle/assets/78768499/36192799-b3fa-46ba-b382-3303c444428a)

PR #16622, opened 2024-06-06
core: Add copy/paste context menu for EditText

kjarosh:
This menu is shown when the EditText is focused and the mouse hovers over it. It has the following elements, which may be disabled when e.g. the field is not editable / selectable:
* Cut,
* Copy,
* Paste,
* Delete,
* Select All.

Additionally, this PR also adds handling for an empty clipboard and prevents copy/cut from password fields.

Related to https://github.com/ruffle-rs/ruffle/issues/1891. Fixes https://github.com/ruffle-rs/ruffle/issues/16474.

| Desktop | Web |
| ------- | --- |
| <image src="https://github.com/ruffle-rs/ruffle/assets/1507072/65c0abae-a196-40e6-990f-263871dabf54" width="300"/> | <image src="https://github.com/ruffle-rs/ruffle/assets/1507072/f639459a-08d2-4e33-807f-18488d42bad2" width="300"/> |
| <image src="https://github.com/ruffle-rs/ruffle/assets/1507072/cac80cee-4d28-4ea2-8621-7fe3fa9aa9f6" width="300"/> | <image src="https://github.com/ruffle-rs/ruffle/assets/1507072/c3f5d790-6eed-46c5-a128-9e6662a653d2" width="300"/> |


kjarosh:
I also noticed that the user should not be able to copy/cut the text from a password text field. So I cleaned up the code a bit to prevent logic duplication and added this case with a test.

Dinnerbone (approved):
😍

PR #16640, opened 2024-06-08
web: Remove jszip, use rust

### LINKED ISSUE #16633: Avoid overriding window.setImmediate / window.clearImmediate when adding ruffle.js
### Describe the bug

When using the self-hosted version and adding `<script src="path/to/ruffle/ruffle.js"></script>`, due to a downstream dependency, `window.setImmediate` and `window.clearImmediate` gets defined on the window.

This is generally fine, but other may conflict with other libraries that also set this method. This appears to be set from ruffle-core dependency on [jszip](https://www.npmjs.com/package/jszip) package which has a dependency on [setimmediate](https://www.npmjs.com/package/setimmediate).

We use ruffle with our web archive replay tools and inject it into every page to support Flash, and this unfortunately interferes with some pages.

### Expected behavior

Ideally, Ruffle won't inject anything into window if there is no Flash content found on the page, though most of the other injections, like window.RufflePlayer are ruffle specific. window.setImmediate / window.clearImmediate should ideally not be defined at all, as there doesn't seem to be a requirement for generally running Ruffle.

### Content Location

N/A (Not related to specific Flash content)

### Affected platform

Self-hosted version

### Operating system

Any

### Browser

Any

### Additional information

Was able to fix this by simply appending
```
window.setImmediate = undefined;
window.clearImmediate = undefined;
```
to the end of the ruffle.js. This may be the simplest/cleanest solution, given that this comes from a third-party dependency.

Dinnerbone:
Idea was from @adrian17 and imo it's how we should be doing "heavy" things anyway.

`ruffle.js` reduces from 393KB to 298KB (**-25%**)
regular wasm increases from 12,985KB to 13,104KB (**+0.9%**)

This fixes #16633

### DIFF
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -1,4 +1,4 @@
-import type { RuffleHandle } from "../dist/ruffle_web";
+import type { RuffleHandle, ZipWriter } from "../dist/ruffle_web";
 import { createRuffleBuilder } from "./load-ruffle";
 import { applyStaticStyles, ruffleShadowTemplate } from "./shadow-template";
 import { lookupElement } from "./register-element";
@@ -15,7 +15,6 @@ import type { MovieMetadata } from "./movie-metadata";
 import { swfFileName } from "./swf-utils";
 import { buildInfo } from "./build-info";
 import { text, textAsParagraphs } from "./i18n";
-import JSZip from "jszip";
 import { isExtension } from "./current-script";
 import { configureBuilder } from "./internal/builder";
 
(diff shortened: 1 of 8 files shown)


adrian17 (approved):
Too bad the typing gets increasingly incomprehensible :c

web/packages/core/src/ruffle-player.ts
```diff
@@ -1342,27 +1349,24 @@ export class RufflePlayer extends HTMLElement {
...
                 duplicateNames.push(solName);
                 if (duplicate > 0) {
                     solName += ` (${duplicate + 1})`;
                 }
-                zip.file(solName + ".sol", blob);
+                zip.addFile(solName + ".sol", array);
             }
         });
-        const blob = await zip.generateAsync({ type: "blob" });
+        const blob = new Blob([zip.save()]);
```
evilpie, line 1369: I would probably specify a `type: "application/zip"`.

PR #16668, opened 2024-06-10
filters: Fix blur scaling

kjarosh:
The blur of 1.0 is invariant, so that it should be preserved when scaling. Using `blur * scale` also scales the invariant, which is wrong. Tests show that `(blur - 1) * scale + 1` matches FP's behavior.

A test was added (`filters/any_blur_scales_with_screen`) which verifies this behavior. The test includes all known filters which have a blur, but some of them (gradient filters) are disabled (made invisible) because they are not yet fully implemented in Ruffle. They can be made visible when they are functional so that this test also covers them.

After this fix, the test `filters/any_blur_scales_with_screen` matches FP's output perfectly with `tolerance=0` and `max_outliers=0` (not in CI though :neutral_face:).


torokati44:
*tentative approval*

Dinnerbone (approved):

Croworbit:
does this fix any specific games?

kjarosh:
> does this fix any specific games?

I'm afraid not, it only improves appearance of blur when the content is scaled.

PR #16679, opened 2024-06-11
chore: Fix some beta clippy lints

torokati44:
Not all of them are fixed, see: https://github.com/rust-lang/rust-clippy/issues/12917

PR #16448, opened 2024-05-24
Revert "ci: Disable rust-lld on Rust nightly"

torokati44:
This reverts commit 1b3701e1aa17c46be9b077ecade6ae9d6e975af1 (#16390), as it should no longer be needed, according to https://github.com/cross-rs/cross/issues/1496#issuecomment-2130454741.

PR #16682, opened 2024-06-11
ci: Set the `-znostart-stop-gc` flag for lld when testing

torokati44:
Should fix CI failing with beta Rust (aside from the lints, for which there's https://github.com/ruffle-rs/ruffle/pull/16679 and https://github.com/ruffle-rs/ruffle/pull/16683).

Now that Rust beta also uses `rust-lld` by default on `x86_64-unknown-linux-gnu`, we can't use `-Z` as a workaround.

An alternative to https://github.com/ruffle-rs/ruffle/pull/16448.
Also reverts https://github.com/ruffle-rs/ruffle/pull/16390.
Referencing: https://github.com/dtolnay/linkme/issues/94
For more info, see: https://lld.llvm.org/ELF/start-stop-gc

### DIFF
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -62,10 +62,12 @@ jobs:
           sudo apt install -y libasound2-dev libxcb-shape0-dev libxcb-xfixes0-dev libgtk-3-dev mesa-vulkan-drivers libpango1.0-dev libudev-dev
 
       # Needed after: https://github.com/rust-lang/rust/pull/124129
-      # Based on: https://github.com/dtolnay/linkme/pull/88
-      - name: Disable rust-lld
-        if: matrix.rust_version == 'nightly'
-        run: echo RUSTFLAGS=${RUSTFLAGS}\ -Zlinker-features=-lld >> $GITHUB_ENV
+      # See also: https://github.com/dtolnay/linkme/issues/94
+      # Additionally: https://lld.llvm.org/ELF/start-stop-gc
+      - name: Disable linker start-stop-gc
+        # Note: We also use `rust-lld` on Windows, see `config.toml`.
+        if: runner.os != 'macOS'
+        run: echo RUSTFLAGS=${RUSTFLAGS}\ -Clink-args=-znostart-stop-gc >> $GITHUB_ENV
 
       - name: Cache Cargo output
         uses: Swatinem/rust-cache@v2


torokati44:
>  This conservative behavior works for existing code which does not take GC into fair consideration, but unnecessarily increases sizes for modern metadata section usage which desires precise GC.
>
> GNU ld 2.37 added `-z start-stop-gc` to restore the traditional behavior ld.lld 13.0.0 defaults to `-z start-stop-gc` and supports `-z nostart-stop-gc` to switch to the conservative behavior.

However, since this is only an issue for testing (we use `linkme` to collect the test cases IIRC), binary size shouldn't be much of an issue.

torokati44:
We may decide to wait for rustc to roll this change (switching to lld on linux by default) back from beta (https://github.com/dtolnay/linkme/pull/88#issuecomment-2161154145) instead of merging this.
Or for `linkme` to fix this some other way with the change in rustc.

danielhjacobs:
See https://github.com/rust-lang/rust/pull/126282 and https://github.com/rust-lang/rust/pull/126278

lqd:
When using `lld`, this is likely to me to be the expected fix for `linkme` users, as I wrote in https://github.com/dtolnay/linkme/pull/88#issuecomment-2119195142 and [the blog post announcing the change](https://blog.rust-lang.org/2024/05/17/enabling-rust-lld-on-linux.html#possible-drawbacks).

torokati44:
I see, thanks!

For the record, while looking into it, I had some other alternative ideas too, but I have no idea whether any of them are actually feasible. Such as:
  - Using the `KEEP` keyword in a linker script - `linkme` already has one for `cortex-m`, but I'm not sure it would be great to have one for x86_64 linux as well...
  - Adding a `retain` attribute to rustc, similar to `used`, which would affect the linker symbol...
  - Having `linkme` manually annotating the start/stop symbols to not be discarded...

bjorn3:
> Adding a retain attribute to rustc, similar to used, which would affect the linker symbol...

torokati44:
Considering that the next beta is coming out almost 6 weeks from now, I'm proposing merging this now, just to clean up CI, and unblock merging PRs without repo admin force.
As seen above, this will likely not be a bad fix long term either.

lqd:
> Considering that the next beta is coming out almost 6 weeks from now

AFAICT the new beta is already out and lld is disabled there

torokati44:
Oh, right, great! Anyway, I think leaving that option won't really hurt...

PR #16683, opened 2024-06-11
chore: Suppress a `clippy::doc_lazy_continuation` beta lint false positive

torokati44:
Split out from https://github.com/ruffle-rs/ruffle/pull/16679.
Referencing https://github.com/rust-lang/rust-clippy/issues/12917 again.


torokati44:
We may decide to wait for clippy to fix this lint instead of merging (and later reverting) it.

Dinnerbone:
That could be a while, I fear. Let's just do it.

Issue #16678, opened 2024-06-11
web: Cannot paste from context menu


n0samu:
### Describe the bug

Unfortunately, browsers only allow reading the clipboard contents while handling the [paste event](https://developer.mozilla.org/en-US/docs/Web/API/Element/paste_event), which is only fired after pressing Ctrl-V or pressing Paste in the browser's native context menu. So Ruffle handles the JS paste event, saving the pasted text to the web UI backend's `clipboard_content` variable; [here's how I implemented it in #11059](https://github.com/ruffle-rs/ruffle/pull/11059/files#diff-7e944172be0a590e61b7f2453657708dcc8233b5d19d6d2749a24fcfde682ab9R33). But clicking the Paste button in Ruffle's (non-native) context menu does not fire the JS paste event, so Ruffle's `clipboard_content` variable is not updated, and the text that shows up is the stale/outdated value of `clipboard_content`. This doesn't cause problems if you're pasting text copied previously from Ruffle, because Ruffle also updates the `clipboard_content` variable when you copy to the clipboard. But if you're pasting text from another location/tab/application, you'll notice that it does not work correctly:

https://github.com/ruffle-rs/ruffle/assets/71368227/93f33515-6fd6-41dd-a3ba-fec3ae7eec47

### Expected behavior

Pasting text using the context menu should always work.

### Content Location

I was testing with this SWF: http://i.notdoppler.com/files/golphysics.swf

### Affected platform

Online demo

### Operating system

Windows 10

### Browser

Firefox 126

### Additional information

The problem also occurs in Chromium/Edge. I found a [tutorial on web.dev](https://web.dev/patterns/clipboard/paste-text/) which seems to bring up a special native context menu in Firefox that allows clipboard pasting from a custom button (with an extra, somewhat nonintuitive step). But it doesn't seem to work in Chromium/Edge. On the other hand, pasting in Google Docs works in Chromium/Edge but not Firefox, but I haven't looked into how that is implemented.

Edit: It looks like this implementation works in all browsers, but in Firefox it still spawns the extra context menu you need to click, and in Chromium it requests clipboard access which the user needs to allow.
https://stackoverflow.com/questions/50633601/is-it-possible-to-paste-from-clipboard-onclick-in-javascript

Croworbit:
when it asks for clipboard, does one get to choose what in the clipboard is revealed? or is it all exposed?

kjarosh:
[...]

PR #16692, opened 2024-06-12
web: Implement pasting text using Clipboard API

kjarosh:
Fixes https://github.com/ruffle-rs/ruffle/issues/16678.

Ruffle does not have direct clipboard access on web, so the current paste implementation from the context menu does not work. This patch intercepts the paste context menu callback, and uses the Clipboard API to ask the browser for the clipboard and update it before calling the callback. When the Clipboard API is not available, a modal informing the user about cut, copy, paste shortcuts is displayed.

| Firefox | Chromium |
| ------- | -------- |
| <image src="https://github.com/ruffle-rs/ruffle/assets/1507072/dd3d576f-7ecc-4140-aa8b-ffba7e14ef8a" width="400"/> | <image src="https://github.com/ruffle-rs/ruffle/assets/1507072/7a5aeb40-3dab-4a0c-9d59-8598aff05d4c" width="400"/> |

| Clipboard modal |
| --------------- |
| <image src="https://github.com/ruffle-rs/ruffle/assets/1507072/3991ee24-29e5-4ea1-a7c6-d33a2b1a99b9" width="500"/> |
| A modal is displayed when the Clipboard API is not supported |

### DIFF
--- a/web/packages/core/src/ruffle-player.ts
+++ b/web/packages/core/src/ruffle-player.ts
@@ -70,7 +70,7 @@ interface ContextMenuItem {
      *
      * @param event The mouse event that triggered the click.
      */
-    onClick: (event: MouseEvent) => void;
+    onClick: (event: MouseEvent) => Promise<void>;
 
     /**
      * Whether this item is clickable.
@@ -1380,7 +1380,7 @@ export class RufflePlayer extends HTMLElement {
     /**
      * Opens the save manager.
      */
-    private openSaveManager(): void {
+    private async openSaveManager(): Promise<void> {
         this.saveManager.classList.remove("hidden");
     }
 
@@ -1467,7 +1467,7 @@ export class RufflePlayer extends HTMLElement {
                     // TODO: better checkboxes
                     text:
                         item.caption + (item.checked ? ` (${CHECKMARK})` : ``),
-                    onClick: () =>
+                    onClick: async () =>
                         this.instance?.run_context_menu_callback(index),
                     enabled: item.enabled,
                 });
@@ -1480,19 +1480,19 @@ export class RufflePlayer extends HTMLElement {
             if (this.isFullscreen) {
                 items.push({
                     text: text("context-menu-exit-fullscreen"),
-                    onClick: () => this.setFullscreen(false),
+                    onClick: async () => this.setFullscreen(false),
                 });
             } else {
                 items.push({
                     text: text("context-menu-enter-fullscreen"),
-                    onClick: () => this.setFullscreen(true),
+                    onClick: async () => this.setFullscreen(true),
                 });
             }
         }
 
         items.push({
             text: text("context-menu-volume-controls"),
-            onClick: () => {
+            onClick: async () => {
                 this.openVolumeControls();
             },
         });
@@ -1533,7 +1533,7 @@ export class RufflePlayer extends HTMLElement {
                 flavor: isExtension ? "extension" : "",
                 version: buildInfo.versionName,
             }),
-            onClick() {
+            async onClick() {
                 window.open(RUFFLE_ORIGIN, "_blank");
             },
         });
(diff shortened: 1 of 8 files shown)


torokati44:
As for the first commit, cf.: https://github.com/torokati44/ruffle/commit/df91395469aab494c7f1c6187ac6418e4af2dee9

I'm not saying that you should use this verbatim, but the comments therein might be worth considering.

torokati44:
> A modal is displayed when the Clipboard API is not supported

Does a plain `alert` have too much of "bad, crufty, old web" vibe?

kjarosh:
> Does a plain `alert` have too much of "bad, crufty, old web" vibe?

TBH I created a modal for pure consistency with other UI elements (also, I took inspiration from other sites which show a similar message), but now that you mentioned, it does have a bit of that vibe, yes :smile:

torokati44:
TBH if there's no compelling technical reason against it, I'd still prefer it I think. This "plain black text on white background" thing, just to tell the user a short message, is not the nicest IMHO.

kjarosh:
A version with `alert`:

<image src="https://github.com/ruffle-rs/ruffle/assets/1507072/ad76f130-db31-43a8-8838-082ce10638cc" width="500"/>

Just let me know which one to include, I'll update this PR accordingly.

The downside of using `alert` is that it's fairly intrusive and freezes temporarily the whole page where Ruffle is embedded. It also does not support setting a custom title and HTML inside.

Dinnerbone:
I'm fairly against alert for something so verbose, personally

n0samu (approved):
Looks great, thanks! I definitely prefer a modal over an alert as well. One small thing though, we might want to show the modal if the user denies the clipboard permission so they know there's an alternative way to paste without giving the site that extra permission they might not be comfortable with.

kjarosh:
Added a second version of the clipboard modal which is shown when user denies access to the clipboard:
<image width=500 src="https://github.com/ruffle-rs/ruffle/assets/1507072/5713cbd6-454e-4605-9702-26eeee50158e"/>

IMO This modal is non-intrusive, as clicking anywhere closes it, so it's not gonna annoy the user, even if they deny access intentionally or miss-click, but may give useful information for some.

torokati44:
FWIW I have accepted the overruling of my preference toward `alert`.

web/packages/core/src/ruffle-player.ts
```diff
@@ -1663,7 +1668,23 @@ export class RufflePlayer extends HTMLElement {
...
                     menuItem.addEventListener(
                         this.contextMenuSupported ? "click" : "pointerup",
-                        onClick,
+                        (event: MouseEvent) => {
+                            // Prevent the menu from being destroyed.
+                            // It's required when we're dealing with async callbacks,
+                            // as the async callback may still use the menu in the future.
+                            event.stopPropagation();
+
+                            const result = onClick(event);
```
evilpie, line 1682: I would `await onClick` here and make the function async of course.

evilpie:
We could also add some text to the context menu when e.g. pasting fails, but that only gives us a limited amount of space.

As a follow up: It would be nice if all our modals would use `color-scheme: light dark;` so that dark mode users aren't flash banged.

web/packages/core/src/shadow-template.ts
```diff
@@ -815,6 +815,35 @@ hardwareModalLink.target = "_blank";
...
+const clipboardModalClose = createElement("span", undefined, "close-modal");
+clipboardModalClose.textContent = "\u00D7";
+const clipboardModalHeading = createElement("h2", undefined);
+clipboardModalHeading.textContent = text("clipboard-message-title");
+const clipboardModalTextDescription = createElement(
+    "p",
+    "clipboard-modal-description",
+);
+const shortcutModifier =
+    navigator.userAgent.indexOf("Mac OS X") === -1 ? "Ctrl" : "Command";
```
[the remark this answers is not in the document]
kjarosh: fixed

evilpie (approved):

web/packages/core/src/ruffle-player.ts
kjarosh: Fixed. I tried to avoid changing every `onClick` callback to be asynchronous :sweat_smile:

adrian17 (approved):

PR #16694, opened 2024-06-12
 extension: Re-enable SWF takeover using declarativeNetRequest 

danielhjacobs:
This is based on https://github.com/w3c/webextensions/issues/460. I added feature detection. This should be tested on Chrome Beta or above.

See screencast:

[Screencast from 2024-06-28 09-56-35.webm](https://github.com/ruffle-rs/ruffle/assets/29206584/a25ef9f7-ac10-4141-9069-63f9e5273ab2)

### DIFF
--- a/web/packages/extension/assets/options.html
+++ b/web/packages/extension/assets/options.html
@@ -31,7 +31,7 @@
             </div>
             <div class="option checkbox">
                 <input type="checkbox" id="swf_takeover" />
-                <label for="swf_takeover">Redirect SWF links to SWF Player</label>
+                <label for="swf_takeover">Play SWF files in browser instead of downloading</label>
             </div>
             <div id="advanced-options">Advanced Options</div>
             <div class="option checkbox">
--- a/web/packages/extension/assets/options.html
+++ b/web/packages/extension/assets/options.html
@@ -29,6 +29,10 @@
                 <input type="checkbox" id="autostart" />
                 <label for="autostart">Autoplay Flash content (click to unmute)</label>
             </div>
+            <div class="option checkbox">
+                <input type="checkbox" id="swf_takeover" />
+                <label for="swf_takeover">Redirect SWF links to SWF Player</label>
+            </div>
             <div id="advanced-options">Advanced Options</div>
             <div class="option checkbox">
                 <input type="checkbox" id="ignore_optout" />
--- a/web/packages/extension/src/background.ts
+++ b/web/packages/extension/src/background.ts
@@ -102,7 +102,7 @@ async function enableSWFTakeover() {
                     redirect: { regexSubstitution: playerPage + "#\\0" },
                 },
                 condition: {
-                    regexFilter: "^.*\\.s(?:wf|pl)(\\?.*|#.*|)$",
+                    regexFilter: "^.*:\\/\\/.*\\/.*\\.s(?:wf|pl)(\\?.*|#.*|)$",
                     responseHeaders: [
                         {
                             header: "content-type",
@@ -128,7 +128,7 @@ async function enableSWFTakeover() {
                     redirect: { regexSubstitution: playerPage + "#\\0" },
                 },
                 condition: {
-                    regexFilter: "^.*\\.s(?:wf|pl)(\\?.*|#.*|)$",
+                    regexFilter: "^.*:\\/\\/.*\\/.*\\.s(?:wf|pl)(\\?.*|#.*|)$",
                     excludedResponseHeaders: [{ header: "content-type" }],
                     resourceTypes: [
                         chrome.declarativeNetRequest.ResourceType?.MAIN_FRAME ??
--- a/web/packages/extension/src/background.ts
+++ b/web/packages/extension/src/background.ts
@@ -244,12 +244,9 @@ function onMessage(
 }
 
 (async () => {
-    const { ruffleEnable, swfTakeover } = await utils.getOptions();
+    const { ruffleEnable } = await utils.getOptions();
     if (ruffleEnable) {
         await enable();
-    }
-    if (!swfTakeover) {
-        await disableSWFTakeover();
     }
 })();
 
--- a/web/packages/extension/src/background.ts
+++ b/web/packages/extension/src/background.ts
@@ -15,7 +15,9 @@ async function enable() {
             {
                 id: 1,
                 action: {
-                    type: chrome.declarativeNetRequest.RuleActionType.REDIRECT,
+                    type:
+                        chrome.declarativeNetRequest.RuleActionType?.REDIRECT ??
+                        "redirect",
                     redirect: { regexSubstitution: playerPage + "#\\0" },
                 },
                 condition: {
@@ -32,14 +34,17 @@ async function enable() {
                         },
                     ],
                     resourceTypes: [
-                        chrome.declarativeNetRequest.ResourceType.MAIN_FRAME,
+                        chrome.declarativeNetRequest.ResourceType?.MAIN_FRAME ??
+                            "main_frame",
                     ],
                 },
             },
             {
                 id: 2,
                 action: {
-                    type: chrome.declarativeNetRequest.RuleActionType.REDIRECT,
+                    type:
+                        chrome.declarativeNetRequest.RuleActionType?.REDIRECT ??
+                        "redirect",
                     redirect: { regexSubstitution: playerPage + "#\\0" },
                 },
                 condition: {
@@ -55,29 +60,39 @@ async function enable() {
                         },
                     ],
                     resourceTypes: [
-                        chrome.declarativeNetRequest.ResourceType.MAIN_FRAME,
+                        chrome.declarativeNetRequest.ResourceType?.MAIN_FRAME ??
+                            "main_frame",
                     ],
                 },
             },
             {
                 id: 3,
                 action: {
-                    type: chrome.declarativeNetRequest.RuleActionType.REDIRECT,
+                    type:
+                        chrome.declarativeNetRequest.RuleActionType?.REDIRECT ??
+                        "redirect",
                     redirect: { regexSubstitution: playerPage + "#\\0" },
                 },
                 condition: {
                     regexFilter: "^.*\\.s(?:wf|pl)(\\?.*|#.*|)$",
                     excludedResponseHeaders: [{ header: "content-type" }],
                     resourceTypes: [
-                        chrome.declarativeNetRequest.ResourceType.MAIN_FRAME,
+                        chrome.declarativeNetRequest.ResourceType?.MAIN_FRAME ??
+                            "main_frame",
                     ],
                 },
             },
         ];
-        await utils.declarativeNetRequest.updateDynamicRules({
-            removeRuleIds: [1, 2, 3],
-            addRules: rules,
-        });
+        try {
+            await utils.declarativeNetRequest.updateDynamicRules({
+                removeRuleIds: [1, 2, 3],
+                addRules: rules,
+            });
+        } catch (e) {
+            console.info(
+                "Failed to register rules: responseHeaders condition unsupported",
+            );
+        }
     }
     if (
         !utils.scripting ||
--- a/web/packages/extension/src/background.ts
+++ b/web/packages/extension/src/background.ts
@@ -64,83 +64,88 @@ async function isHeaderConditionSupported() {
 async function enableSWFTakeover() {
     // Checks if the responseHeaders condition is supported and not behind a disabled flag.
     if (utils.declarativeNetRequest && (await isHeaderConditionSupported())) {
-        const playerPage = utils.runtime.getURL("/player.html");
-        const rules = [
-            {
-                id: 1,
-                action: {
-                    type:
-                        chrome.declarativeNetRequest.RuleActionType?.REDIRECT ??
-                        "redirect",
-                    redirect: { regexSubstitution: playerPage + "#\\0" },
+        const { ruffleEnable } = await utils.getOptions();
+        if (ruffleEnable) {
+            const playerPage = utils.runtime.getURL("/player.html");
+            const rules = [
+                {
+                    id: 1,
+                    action: {
+                        type:
+                            chrome.declarativeNetRequest.RuleActionType
+                                ?.REDIRECT ?? "redirect",
+                        redirect: { regexSubstitution: playerPage + "#\\0" },
+                    },
+                    condition: {
+                        regexFilter: ".*",
+                        responseHeaders: [
+                            {
+                                header: "content-type",
+                                values: [
+                                    "application/x-shockwave-flash",
+                                    "application/futuresplash",
+                                    "application/x-shockwave-flash2-preview",
+                                    "application/vnd.adobe.flash.movie",
+                                ],
+                            },
+                        ],
+                        resourceTypes: [
+                            chrome.declarativeNetRequest.ResourceType
+                                ?.MAIN_FRAME ?? "main_frame",
+                        ],
+                    },
                 },
-                condition: {
-                    regexFilter: ".*",
-                    responseHeaders: [
-                        {
-                            header: "content-type",
-                            values: [
-                                "application/x-shockwave-flash",
-                                "application/futuresplash",
-                                "application/x-shockwave-flash2-preview",
-                                "application/vnd.adobe.flash.movie",
-                            ],
-                        },
-                    ],
-                    resourceTypes: [
-                        chrome.declarativeNetRequest.ResourceType?.MAIN_FRAME ??
-                            "main_frame",
-                    ],
+                {
+                    id: 2,
+                    action: {
+                        type:
+                            chrome.declarativeNetRequest.RuleActionType
+                                ?.REDIRECT ?? "redirect",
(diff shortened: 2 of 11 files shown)


web/packages/extension/src/background.ts
```diff
@@ -9,6 +9,75 @@ async function contentScriptRegistered() {
...
+                            values: [
+                                "application/x-shockwave-flash",
+                                "application/futuresplash",
+                                "application/x-shockwave-flash2-preview",
+                                "application/vnd.adobe.flash.movie",
+                            ],
+                        },
+                    ],
+                    resourceTypes: [
+                        chrome.declarativeNetRequest.ResourceType.MAIN_FRAME,
```
danielhjacobs, line 35: In `chrome`, this is part of an enum: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/chrome/index.d.ts#L12180. I can't use `utils.declarativeNetRequest.ResourceType.MAIN_FRAME` because in `browser` it's a `type` instead: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/webextension-polyfill/namespaces/declarativeNetRequest.d.ts#L21

web/packages/extension/src/background.ts
```diff
@@ -9,6 +9,75 @@ async function contentScriptRegistered() {
 }
 
 async function enable() {
+    if (utils.declarativeNetRequest) {
+        const playerPage = utils.runtime.getURL("/player.html");
+        const rules = [
+            {
+                id: 1,
+                action: {
+                    type: chrome.declarativeNetRequest.RuleActionType.REDIRECT,
```
danielhjacobs, line 18: In `chrome`, this is part of an `enum`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/chrome/index.d.ts#L12198. I can't use utils.declarativeNetRequest.RuleActionType.REDIRECT because in `browser` it's a `type` instead: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/webextension-polyfill/namespaces/declarativeNetRequest.d.ts#L413

web/packages/extension/src/background.ts
```diff
@@ -9,6 +9,75 @@ async function contentScriptRegistered() {
 }
 
 async function enable() {
+    if (utils.declarativeNetRequest) {
```
danielhjacobs, line 12: This needs feature detection, which I asked about here: https://github.com/w3c/webextensions/issues/638

danielhjacobs: Even if that issue is addressed, it won't help as it won't impact older versions of Chrome. My plan is to do something like the following:
```ts
if (utils.declarativeNetRequest && typeof browser !== "undefined") {
```
This will work if https://issues.chromium.org/issues/40556351 gets fixed after https://issues.chromium.org/issues/40727004, once both have been fixed.

web/packages/extension/manifest_firefox.json5
```diff
@@ -48,7 +48,8 @@
         "open_in_tab": true,
     },
     "permissions": [
-        "<all_urls>", // // To allow script injecting + the internal player to bypass CORS
+        "<all_urls>", // // To allow script injecting + the internal player to bypass CORS + SWF takeover
+        "declarativeNetRequest",
```
danielhjacobs, line 52: Completely unused on Firefox because Firefox does not yet support the `responseHeaders` `RuleCondition`, but adding it is not a problem because Firefox will throw within the `try` in the background script.

web/packages/extension/src/background.ts
```diff
@@ -9,6 +9,75 @@ async function contentScriptRegistered() {
...
+                    redirect: { regexSubstitution: playerPage + "#\\0" },
+                },
+                condition: {
+                    regexFilter: "^.*\\.s(?:wf|pl)(\\?.*|#.*|)$",
+                    responseHeaders: [
+                        {
+                            header: "content-type",
+                            values: [
+                                "application/octet-stream",
+                                "application/binary-stream",
```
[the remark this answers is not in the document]
danielhjacobs: Can be resolved very soon after https://crrev.com/c/5634852 is pushed on top of main.

web/packages/extension/src/background.ts
danielhjacobs: Currently doing this in the way suggested here: https://github.com/w3c/webextensions/issues/460#issuecomment-2169017147

> Perhaps not an ideal solution but checking the user agent of the browser for a Chrome version may help detect if response header conditions are supported

web/packages/extension/src/background.ts
```diff
@@ -9,6 +9,100 @@ async function contentScriptRegistered() {
 }
 
 async function enable() {
+    // Unlike other browsers, Chrome registers rules with an unsupported condition.
+    // Therefore, check that Chrome is a version that supports the responseHeaders condition,
+    // so the browser in use throws an exception or supports the responseHeaders condition.
+    // The alternative check here is to use feature detection to detect support for something
+    // that strictly greater Chrome versions support, but no such features exist yet.
+    const chrome127OrLess =
+        /Chrome\/([0-9]|[0-9][0-9]|1[0-1][0-9]|12[0-7])\./.test(
```
danielhjacobs, line 18: Chrome/0. - Chrome/9.
Chrome/00. - Chrome 99.
Chrome 100. - Chrome 119.
Chrome 120. - Chrome 127.

All Chromium browsers seem to have Chrome in their user-agent string, which is desired here.

The current Chrome Dev, Chrome 128, supports the responseHeaders condition.

web/packages/extension/src/background.ts
```diff
@@ -9,6 +9,100 @@ async function contentScriptRegistered() {
 }
 
 async function enable() {
+    // Unlike other browsers, Chrome registers rules with an unsupported condition.
+    // Therefore, check that Chrome is a version that supports the responseHeaders condition,
+    // so the browser in use throws an exception or supports the responseHeaders condition.
+    // The alternative check here is to use feature detection to detect support for something
+    // that strictly greater Chrome versions support, but no such features exist yet.
+    const chrome127OrLess =
+        /Chrome\/([0-9]|1[0-9]|1[0-1][0-9]|12[0-7])\./.test(
```
danielhjacobs, line 18: Chrome/0. - Chrome/9.
Chrome/10. - Chrome 99.
Chrome 100. - Chrome 119.
Chrome 120. - Chrome 127.

See https://www.useragentstring.com/pages/Chrome/

All Chromium browsers seem to have Chrome in their user-agent string, which is desired here.

The current Chrome Dev, Chrome 128, supports the responseHeaders condition.

web/packages/extension/src/background.ts
```diff
@@ -9,6 +9,99 @@ async function contentScriptRegistered() {
 }
 
 async function enable() {
+    // Unlike other browsers, Chrome registers rules with an unsupported condition.
+    // Therefore, check that Chrome is a version that supports the responseHeaders condition,
+    // so the browser in use throws an exception or supports the responseHeaders condition.
+    // The alternative check here is to use feature detection to detect support for something
+    // that strictly greater Chrome versions support, but no such features exist yet.
+    const chrome127OrLess = /Chrome\/([0-9]|1[0-9]|1[0-1][0-9]|12[0-7])\./.test(
```
danielhjacobs, line 17: Chrome/0. - Chrome/9.
Chrome/10. - Chrome 99.
Chrome 100. - Chrome 119.
Chrome 120. - Chrome 127.

See https://www.useragentstring.com/pages/Chrome/

All Chromium browsers seem to have Chrome in their user-agent string, which is desired here.

The current Chrome Dev, Chrome 128, supports the responseHeaders condition.

web/packages/extension/src/background.ts
danielhjacobs: Changed to the suggestion on https://github.com/w3c/webextensions/issues/638#issuecomment-2181124486

web/packages/extension/src/background.ts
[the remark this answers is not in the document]
danielhjacobs: Undid this removal temporarily since the currently available Chrome Dev for Fedora doesn't include this commit.

web/packages/extension/src/background.ts
```diff
@@ -8,7 +8,138 @@ async function contentScriptRegistered() {
...
+    try {
+        // Throws synchronously if not supported.
+        await utils.declarativeNetRequest.updateDynamicRules({
+            addRules: [
+                {
+                    id: ruleId,
+                    condition: { responseHeaders: [{ header: "whatever" }] },
+                    action: {
+                        type:
+                            chrome.declarativeNetRequest.RuleActionType
```
evilpie, line 23: Should we use `utils.declarativeNetRequest` consistently?

evilpie: And honestly, I don't see a point in doing this instead of just using strings.

danielhjacobs: TS yells about anything else due to type definitions by DefinitelyTyped.

danielhjacobs: See https://github.com/DefinitelyTyped/DefinitelyTyped/blob/223f242577e08bea526d40d53eee9fea9b13a9f1/types/chrome/index.d.ts#L12300 and https://github.com/DefinitelyTyped/DefinitelyTyped/blob/223f242577e08bea526d40d53eee9fea9b13a9f1/types/chrome/index.d.ts#L12196. For chrome this must be a defined enum member.

See https://github.com/DefinitelyTyped/DefinitelyTyped/blob/223f242577e08bea526d40d53eee9fea9b13a9f1/types/webextension-polyfill/namespaces/declarativeNetRequest.d.ts#L486 and https://github.com/DefinitelyTyped/DefinitelyTyped/blob/223f242577e08bea526d40d53eee9fea9b13a9f1/types/webextension-polyfill/namespaces/declarativeNetRequest.d.ts#L413. For browser this must be a string.

evilpie:
Can we at least wait until this is properly supported in Chrome so we don't have to do that annoying bug detection dance?

danielhjacobs:
> Can we at least wait until this is properly supported in Chrome so we don't have to do that annoying bug detection dance?

I don't disagree with that but we'll still need to decide what to do here. Supposedly Chromium does throw just like Firefox with a fully unsupported RuleCondition. The bug, as confirmed in https://issues.chromium.org/issues/347186592#comment9, is actually that features developed behind a flag are not fully hidden behind the flag in Chrome. Without this "annoying bug detection dance", with just a try and catch, we may end up in a situation where both older and newer versions of Chrome work properly with Ruffle but the versions that were implementing this feature behind a flag are broken. I'm not sure that's acceptable.

web/packages/extension/src/background.ts
evilpie: I think we should probably just use one namespace and stick with at some point in the future.

danielhjacobs: Yeah, but that should ideally be `browser` and Chrome still doesn't support that. For our use-case `chrome` does work fine but the WECG says that's not supposed to be the standard.

web/packages/extension/src/background.ts
```diff
@@ -8,7 +8,138 @@ async function contentScriptRegistered() {
...
+            {
+                id: 2,
+                action: {
+                    type:
+                        chrome.declarativeNetRequest.RuleActionType?.REDIRECT ??
+                        "redirect",
+                    redirect: { regexSubstitution: playerPage + "#\\0" },
+                },
+                condition: {
+                    regexFilter: "^.*\\.s(?:wf|pl)(\\?.*|#.*|)$",
```
evilpie, line 104: Technically this would match `http://abc.swf#foo` if the `.swf` existed. I guess it's unlikely to ever appear?

web/packages/extension/src/background.ts
```diff
@@ -8,7 +8,138 @@ async function contentScriptRegistered() {
     return matchingScripts?.length > 0;
 }
 
+async function isHeaderConditionSupported() {
```
evilpie, line 11: `// Copied from https://github.com/w3c/webextensions/issues/638#issuecomment-2181124486` would be appropriate.

evilpie:
> the versions that were implementing this feature behind a flag are broken. I'm not sure that's acceptable.

I didn't quite realize this could happen. I sort of assumed this problem would still be limited to Canary.

Do we want/need to have a setting for disabling this specific feature?

danielhjacobs:
Having a setting for disabling this feature may be nice in general since some people did find it frustrating when trying to just download SWFs. I added showSWFDownload to the player page for that reason.

However, I'm not sure that's enough. Try checking out f81fa15 and using the extension in Chrome stable. https://google.com redirects to the player page.

web/packages/extension/src/background.ts
danielhjacobs: If you can suggest a more accurate regex I can change this. This was the regex @Dinnerbone added here before it was devided that without the ability to check headers this was insufficient: https://github.com/ruffle-rs/ruffle/pull/12717/files#diff-b9e86d98769e020e68919846919692da73f818952ca6a5b93d84d449af40e916R55

danielhjacobs: This will also only match http://abc.swf#foo if that URL happens to have a Content-Type header of "application/octet-stream", "application/binary-stream", no Content-Type header (as that would match rule 3), or an empty string as a Content-Type header (after adding that to the `values` array). It's far more likely that URL will have a `text/html` Content-Type header, and that's still only if that TLD existed.

danielhjacobs:
> Do we want/need to have a setting for disabling this specific feature?

Latest commit adds this.

web/packages/extension/src/background.ts
danielhjacobs: Once https://crbug.com/798169 is addressed we can just use browser, but we'd have to wait until whatever version of Chromium supports that is old enough to be the oldest version we want to support.

web/packages/extension/src/background.ts
danielhjacobs: Thoughts on replacing this with this: `regexFilter: "^.*:\\/\\/.*\\/.*\\.s(?:wf|pl)(\\?.*|#.*|)$"`

That's [anything]://[anything]/[anything].swf(?[anything] or #[anything] or nothing)

web/packages/extension/src/background.ts
danielhjacobs: Added TODO comment for switching to a single namespace to utils.ts. As stated, I unfortunately can't change this with the current DefinitelyTyped type definitions, so just marking this as resolved.

danielhjacobs:
Marking this PR as ready for review. Whether we merge soon or wait until Chrome 128 reaches stable in August is an open question, but the code shouldn't actually be changing between now and then. Unfortunately, the feature detection method copied from https://github.com/w3c/webextensions/issues/638#issuecomment-2181124486 will be necessary to continue supporting Ruffle on Chrome 126 or 127, which were implementing this feature behind a flag, as noted in https://issues.chromium.org/issues/347186592#comment9.

web/packages/extension/src/background.ts
```diff
@@ -77,10 +235,13 @@ function onMessage(
 }
 
 (async () => {
-    const { ruffleEnable } = await utils.getOptions();
+    const { ruffleEnable, swfTakeover } = await utils.getOptions();
     if (ruffleEnable) {
         await enable();
     }
+    if (!swfTakeover) {
```
evilpie, line 242: Is this really necessary here? (We only conditionally enabled from `enable`.

web/packages/extension/assets/options.html
```diff
@@ -29,6 +29,10 @@
                 <input type="checkbox" id="autostart" />
                 <label for="autostart">Autoplay Flash content (click to unmute)</label>
             </div>
+            <div class="option checkbox">
+                <input type="checkbox" id="swf_takeover" />
+                <label for="swf_takeover">Redirect SWF links to SWF Player</label>
```
evilpie, line 34: I am not sure if this is really clear. Opening a page is not really a link? Maybe "Play SWF files in browser instead of downloading" or something?

evilpie (approved):
Thanks for adding the option. Just two notes.

web/packages/extension/src/background.ts
danielhjacobs: Yeah, this shouldn't be needed anymore, conditionally enabling in `enable` was the latest commit, after I realized without it if you toggled off both the SWF takeover and Ruffle itself the SWF takeover would be enabled when you re-enabled Ruffle even if you didn't toggle it back on.

web/packages/extension/assets/options.html
danielhjacobs: Ok, this works, I was slightly worried it was going to be really long and take up too much space, but at least in English, it's still just two lines, so I think it's fine.
Screenshot 2024-06-30 105005

danielhjacobs:
Official docs: https://developer.chrome.com/docs/extensions/reference/api/declarativeNetRequest#type-HeaderInfo and https://developer.chrome.com/docs/extensions/reference/api/declarativeNetRequest#property-RuleCondition-excludedResponseHeaders and https://developer.chrome.com/docs/extensions/reference/api/declarativeNetRequest#property-RuleCondition-responseHeaders

Dinnerbone (approved):
Thank you!

Issue #16576, opened 2024-06-03
Steam deck crashing on launch in game mode


monkeyx-net:
### Describe what you were doing

Texting using Ruffle on steamdeck in game mode. using nightly build 29/05/2024

Please also see the discussion on Discord dev general channel

### What does the crash message say?

panicked at desktop/src/gui/controller.rs:230:14:
Surface became unavailable: Timeout

### Ruffle Version

0.1.0-nightly (1712cd61e50f75f2bf4b153b168ed05a0056fe36 2024-05-29)

### Affected platform

Desktop app

### Operating system

Linux Rolling Release (Holo) [64-bit]

monkeyx-net:
Screen shot of crash

https://discord.com/channels/610531541889581066/614855396468719619/1247141956925263894

My initial discord message

I am in the process of adding ruffle to #RetroDeck. Which is a flatpak installation using ES-DE as the front end  emulator with pre confgured options etc

I am currently using the 29/05/2024 nightly ruffle as binary for testing.

So ruffle works without issue on Steam Deck in  Desktop mode with the --fullscreen option.

It only woks in gaming mode with the -g gl option though.

So I tried to run the 29/05 version directly in Gaming mode rather than via RetroDeck and ruffle panics and crashes and asks to send a crash report.

RetroDeck also works on my linux laptop with ruffle with all graphics modes apart from the obvious dx12.

Looking for advice on a good nightly to use for Steam Deck testing and to check other peoples experiences of using ruffle in gaming mode. (edited)

Issue #16304, opened 2024-05-13
Ruffle crash on launch Arch linux (Vulkan)


Nova-archx86:
### Describe what you were doing

Just opened the application and waited about half a second

### What does the crash message say?

panicked at desktop/src/gui/controller.rs:230:14:
Surface became unavailable: Timeout

### Ruffle Version

0.1.0-nightly (3d412c955ff2a84f02de7ef17be9eba70f3da5d2 2024-05-11)

### Affected platform

Desktop app

### Operating system

Arch Linux Rolling Release [64-bit]

### Additional information

Kernel version: Linux 6.8.9-arch1-2
CPU: AMD Ryzen 5 3600 (6 cores)
GPU: Radeon RX 7600 (8GB) (Vulkan)

Although this report is for the ruffle-nightly-bin aur package, I've experienced the same issue when compiling the source from ruffle-git (ruffle-adapter-git too) as well as when downloading the official linux binary on the website.

Logs:
2024-05-13T05:28:57.762423Z  WARN wgpu_hal::vulkan::instance: Unable to find extension: VK_EXT_swapchain_colorspace    
2024-05-13T05:28:57.781805Z  INFO ruffle_desktop::gui::controller: Using preferred backend Vulkan
2024-05-13T05:28:57.787678Z  INFO ruffle_desktop::gui::controller: Using graphics API vulkan on AMD Radeon RX 7600 (type: DiscreteGpu)
2024-05-13T05:28:57.810688Z  WARN egui_wgpu::renderer: Detected a linear (sRGBA aware) framebuffer Bgra8UnormSrgb. egui prefers Rgba8Unorm or Bgra8Unorm    
thread 'main' panicked at desktop/src/gui/controller.rs:230:14:
Surface became unavailable: Timeout
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Aborted (core dumped)

Nova-archx86:
From what I've tested ruffle seems to work just fine when only using opengl... Im wondering if this might be a vulkan issue?

Issue #16213, opened 2024-05-04
Instantly crashes when opened


unixportadmin22:
### Describe what you were doing

I was just opening the program, not even loading anything yet.

### What does the crash message say?

panicked at desktop/src/gui/controller.rs:230:14:
Surface became unavailable: Timeout

### Ruffle Version

0.1.0-nightly (d1553a203e679d7bb6e8fc37979970087f344f73 2024-05-04)

### Affected platform

Desktop app

### Operating system

Arch Linux [64-bit]

Issue #16009, opened 2024-04-16
Crash on Ubuntu 23.10 build from source @899c0bb


mtdudek:
### Describe what you were doing

I've build ruffle directly from source `cargo run --release --package=ruffle_desktop`.
Crash is instantaneous.

### What does the crash message say?

panicked at desktop/src/gui/controller.rs:230:14:
Surface became unavailable: Timeout

### Ruffle Version

0.1.0-nightly (899c0bb157e085a7cbc24e07ccda0291722df65c 2024-04-16)

### Affected platform

Desktop app

### Operating system

Ubuntu 23.10.0 (mantic) [64-bit]

torokati44:
Just curious, what GPU do you have, with which drivers?
Also, is this on Wayland or on X11?

mtdudek:
Wayland session with AMD iGPU 680M(rembrant).
Driver: amdgpu kernel 6.7.3

torokati44:
Just a random guess, might not do anything:
Could you try putting `WAYLAND_DISPLAY=""` before the command you pasted above? This should make it fall back to XWayland.

mtdudek:
Forcing XWayland still produces the same issue.
Full backtrace
```
stack backtrace:
   0:     0x5ecaebee9fc6 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h410d4c66be4e37f9
   1:     0x5ecaebf1a7d0 - core::fmt::write::he40921d4802ce2ac
   2:     0x5ecaebee52af - std::io::Write::write_fmt::h5de5a4e7037c9b20
   3:     0x5ecaebee9da4 - std::sys_common::backtrace::print::h11c067a88e3bdb22
   4:     0x5ecaebeeb6d7 - std::panicking::default_hook::{{closure}}::h8c832ecb03fde8ea
   5:     0x5ecaebeeb439 - std::panicking::default_hook::h1633e272b4150cf3
   6:     0x5ecaeaa44571 - ruffle_desktop::init::{{closure}}::ha70548078a1d4956
   7:     0x5ecaebeebcf6 - std::panicking::rust_panic_with_hook::hb164d19c0c1e71d4
   8:     0x5ecaebeeba42 - std::panicking::begin_panic_handler::{{closure}}::h0369088c533c20e9
   9:     0x5ecaebeea4c6 - std::sys_common::backtrace::__rust_end_short_backtrace::hc11d910daf35ac2e
[...]

PR #16699, opened 2024-06-13
wgpu: Prevent random crashes on desktop by handling SurfaceErrors properly

kjarosh:
This patch adds proper handling for `SurfaceError`s that may happen when calling `get_current_texture()`, instead of always panicking.

This fixes random crashes, e.g. when the GPU took too long to render a frame, or when the surface has unexpectedly changed.

Fixes https://github.com/ruffle-rs/ruffle/issues/16698, fixes https://github.com/ruffle-rs/ruffle/issues/16576, fixes https://github.com/ruffle-rs/ruffle/issues/16336, fixes https://github.com/ruffle-rs/ruffle/issues/16304, fixes https://github.com/ruffle-rs/ruffle/issues/16302, fixes https://github.com/ruffle-rs/ruffle/issues/16213, fixes https://github.com/ruffle-rs/ruffle/issues/16116, fixes https://github.com/ruffle-rs/ruffle/issues/16053, fixes https://github.com/ruffle-rs/ruffle/issues/16009, fixes https://github.com/ruffle-rs/ruffle/issues/15994, fixes https://github.com/ruffle-rs/ruffle/issues/15379, fixes https://github.com/ruffle-rs/ruffle/issues/15231, fixes https://github.com/ruffle-rs/ruffle/issues/15219, fixes https://github.com/ruffle-rs/ruffle/issues/14597, fixes https://github.com/ruffle-rs/ruffle/issues/14581, fixes https://github.com/ruffle-rs/ruffle/issues/14573, fixes https://github.com/ruffle-rs/ruffle/issues/14296, fixes https://github.com/ruffle-rs/ruffle/issues/14056, fixes https://github.com/ruffle-rs/ruffle/issues/13567, fixes https://github.com/ruffle-rs/ruffle/issues/13145, fixes https://github.com/ruffle-rs/ruffle/issues/12608, fixes https://github.com/ruffle-rs/ruffle/issues/11672.

### DIFF
--- a/desktop/src/gui/controller.rs
+++ b/desktop/src/gui/controller.rs
@@ -16,6 +16,7 @@ use std::sync::{Arc, MutexGuard};
 use std::time::{Duration, Instant};
 use unic_langid::LanguageIdentifier;
 use url::Url;
+use wgpu::SurfaceError;
 use winit::dpi::PhysicalSize;
 use winit::event::WindowEvent;
 use winit::event_loop::EventLoop;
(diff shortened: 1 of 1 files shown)


desktop/src/gui/controller.rs
```diff
@@ -224,10 +229,24 @@ impl GuiController {
...
     pub fn render(&mut self, mut player: Option<MutexGuard<Player>>) {
-        let surface_texture = self
-            .surface
-            .get_current_texture()
-            .expect("Surface became unavailable");
+        let surface_texture = match self.surface.get_current_texture() {
+            Ok(surface_texture) => surface_texture,
+            Err(e @ (SurfaceError::Lost | SurfaceError::Outdated)) => {
+                // Reconfigure the surface if lost or outdated
+                tracing::debug!("Surface became unavailable: {:?}, reconfiguring", e);
```
evilpie, line 236: Isn't this more at the level of `tracing::warn`?

kjarosh: I was wondering about this, on one hand it _may_ suggest that something's wrong, but on the other hand this is a fairly "standard" error, in a way that it just tells the application that it should recreate the swap chain. My logic was that "if it works, we don't need the logs, but if it's not working we can always debug this by enabling debug level".

So anyway, I'll switch to `warn` for all errors because now that I'm thinking about this, it was a panic before so these errors probably did not happen often :smile:

desktop/src/gui/controller.rs
```diff
@@ -224,10 +229,33 @@ impl GuiController {
     }
 
     pub fn render(&mut self, mut player: Option<MutexGuard<Player>>) {
-        let surface_texture = self
-            .surface
-            .get_current_texture()
-            .expect("Surface became unavailable");
+        let surface_texture = match self.surface.get_current_texture() {
+            Ok(surface_texture) => surface_texture,
+            Err(e @ (SurfaceError::Lost | SurfaceError::Outdated)) => {
```
Dinnerbone, line 234: Oh wow, I don't see the `@` used much!

PR #16704, opened 2024-06-13
core,desktop: Update `wgpu` to `v0.20.1`, bump `egui`

torokati44:
`egui` actually reverted to `wgpu` `0.19` for a while (https://github.com/emilk/egui/pull/4559), but now that https://github.com/gfx-rs/wgpu/pull/5681 fixed the reasons for the downgrade, they have updated again: https://github.com/emilk/egui/pull/4560

The little `egui` API change was: https://github.com/emilk/egui/pull/4556

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,16 +19,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c71b1793ee61086797f5c80b6efa2b8ffa6d5dd703f118545808a7f2e27f7046"
 
 [[package]]
-name = "accesskit"
-version = "0.12.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "74a4b14f3d99c1255dcba8f45621ab1a2e7540a0009652d33989005a4d0bfc6b"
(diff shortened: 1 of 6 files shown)


torokati44:
Trying to fix CI by updating WARP, as per: https://github.com/gfx-rs/wgpu/issues/5591#issuecomment-2073388373

torokati44:
In the meantime, https://github.com/emilk/egui/pull/4641 got merged too, so bumping egui to include that as well, so https://github.com/ruffle-rs/ruffle/pull/16639 can fully come to fruition!

Dinnerbone (approved):

torokati44:
Hmmm, it seems that this now drops `accesskit`. Is that a problem? Shall we add it back?
https://docs.rs/crate/egui/latest/features#accesskit

Dinnerbone:
It looks like it was always an optional dependency that we didn't enable... not sure where it came from before, to get enabled?

Dinnerbone:
Locally on master, accesskit is not being used

torokati44:
Well then, it should probably be fine.

PR #16713, opened 2024-06-14
ci: Delete broken MSVCRT DLLs on Windows

torokati44:
While waiting for the GHA runner image to be fixed, let's try some more workarounds...


torokati44:
Nope, this takes over 15 minutes, and doesn't work. *sigh*

PR #16725, opened 2024-06-14
core: Implement LocalConnection

Dinnerbone:
Oh boy, this was kind of a mess.

Technically, AVM1 and AVM2 `LocalConnection.send`, `connect`, `close` methods and associated callbacks are completely done. However, there's caveats to the underlying system.

1. This does **not** implement cross-player communication. You can't have two SWFs on the same page talk to eachother, or two instances of Ruffle desktop talk to eachother, etc. Maybe a future PR.
2. AVM1 has many AMF0 (de)serialization bugs. Some things just don't serialize well right now, and that may cause bugs. Notably, you can't serialize arrays.
3. AVM2 has many AMF0 and AMF3 (de)serialization bugs. Notably, you can't serialize something that's recursive, vectors, byte arrays, even arrays (though I kinda hack fixed that... may not be correct but it's better than before)
4. Domain security is **not** implemented, kinda deliberately. It does prefix connection names with domains, however I do not trust that our current domain logic is valid and am worried that by implementing that straight away, I'd just regress anything that uses even a stubbed LC. That can come later.

Progresses #8330 but it still gets stuck on a blue screen. Can see that it downloads more files though.


core/src/avm2/amf.rs
```diff
@@ -55,22 +55,27 @@ pub fn serialize_value<'gc>(
...
+                            sparse.push(elem);
+                        }
                     }
-                }
 
-                let len = o.as_array_storage().unwrap().length() as u32;
-                Some(AmfValue::ECMAArray(dense, sparse, len))
+                    Some(AmfValue::ECMAArray(dense, sparse, len))
+                } else {
+                    // TODO: is this right?
```
evilpie, line 76: Mhm did https://github.com/ruffle-rs/rust-flash-lso/issues/54 not make this work in AMF0?

Dinnerbone: That's for writing down to bytes, but we don't do that here.

(Writing down to bytes has a whole host of other issues, for AMF3-in-AMF0. Rather avoid the cost of that if we don't need to, as well)

core/src/local_connection.rs
```diff
@@ -1,82 +1,239 @@
...
+
 /// Manages the collection of local connections.
 pub struct LocalConnections<'gc> {
-    connections: SlotMap<LocalConnectionHandle, LocalConnection<'gc>>,
+    connections: FnvHashMap<WString, LocalConnectionKind<'gc>>,
+    messages: Vec<QueuedMessage<'gc>>,
 }
 
-unsafe impl<'gc> Collect for LocalConnections<'gc> {
+unsafe impl Collect for LocalConnections<'_> {
```
Lord-McSweeney, line 121: This should be able to `derive(Collect)` now

Lord-McSweeney (approved):

core/src/local_connection.rs
Dinnerbone: Can't, WString doesn't impl Collect :(

Issue #13401, opened 2023-10-02
Monopoly: Cannot create property http://www.w3.org/XML/1998/namespace::lang on Object


MartySVK:
### Describe the bug

The game was working in the 10-01-2023 build, it was probably broken by PR - https://github.com/ruffle-rs/ruffle/pull/13284

error:
```Error dispatching event EventObject(EventObject { type: "complete", class: flash.events::Event, ptr: 0x8918b0 }) to handler FunctionObject(FunctionObject { ptr: 0xc6d530 }) : ReferenceError: Error #1056: Cannot create property http://www.w3.org/XML/1998/namespace::lang on Object.```

Game Online - https://www.gameflare.com/online-game/monopoly-online/ (Ruffle was downgraded on the website)
SWF - https://data.gameflare.com/games/7027/pg9lUv19G9x30D/game.swf (but the game needs additional files)

### Expected behavior

progress the game

### Affected platform

Desktop app

### Operating system

Windows 10

MartySVK:
Also, I noticed from the comments that computer players are not able to build the houses. But that's a different issue

PR #16727, opened 2024-06-14
avm2: Implement namespace support for flash.xml

evilpie:


evilpie:
Damn I really made it hard on myself and used the very special namespace `xml` in the failing test: `<xml:a/>`.

evilpie:
So I removed the test for the special `xml:` namespace and replaced it with something that tests something actually useful. I think we need to actually make some changes to E4X's `XML` to support it properly.

The Monopoly game from #13401 still works though.

Dinnerbone (approved):

PR #2705, opened 2021-01-21
Restart MovieClips on End tags

torokati44:
This is an attempt to fix #2699, and help a little bit with #932.

Since I know very little about Flash, this might be entirely wrong, and might break much other content.
What I can see however, is that this fixes at least the case where the `frameCount` in the SWF header is too large.

So, the following loops, for example, now work correctly:
 - [z0r.de/838](https://z0r.de/L/z0r-de_838.swf)
 - [z0r.de/2304](https://z0r.de/L/z0r-de_2304.swf)
 - [z0r.de/2139](https://z0r.de/L/z0r-de_2139.swf)

**EDIT:** Well, these loops require the changes in the H.263 PR (#2173) as well to work _entirely_, but those are independent and orthogonal to this one. At least the looping also affects the sound, so it can be tested that way too.

PR #16730, opened 2024-06-15
avm2: Use actual ShowFrame tag count instead of header for most things

Aaron1011:
Some SWFS have 2 frames, but only declare 1 frames in the SWF header. This PR allows those swfs to advance to the next frame (either normally or via a goto).


core/src/display_object/movie_clip.rs
```diff
@@ -1850,8 +1861,22 @@ impl<'gc> MovieClip<'gc> {
...
+            // *not* check this against the frame count in the SWF header - flash player
+            // ignores that value and allows you to seek to any frame, even if the total
+            // number of frames exceeds the frame count in the header.
+            frame.min(max(mc.frames_loaded_ignore_header(), 0) as FrameNumber)
+        } else {
+            // Sanity; let's make sure we don't seek way too far.
+            frame.min(max(mc.frames_loaded(), 0) as FrameNumber)
+        };
+
+        tracing::debug!(
```
Lord-McSweeney, line 1875: Is this leftover or there for debugging?

core/src/display_object/movie_clip.rs
```diff
@@ -4649,12 +4689,35 @@ impl<'gc> MovieClipStatic<'gc> {
...
+                    return Ok(ControlFlow::Exit);
+                }
+                _ => {}
+            };
+            Ok(ControlFlow::Continue)
+        };
+
+        if is_root && !swf.is_empty() {
+            let mut reader = swf.read_from(0);
+            let _ = tag_utils::decode_tags(&mut reader, tag_callback);
```
Lord-McSweeney, line 4711: Does this mean we have to decode all the SWF tags twice now?

Aaron1011: Yes, but the actual decoding work here is just repeatedly reading tag codes and lengths. Also, it's only done for root clips.

core/src/display_object/movie_clip.rs
```diff
@@ -1850,8 +1861,17 @@ impl<'gc> MovieClip<'gc> {
...
-        let clamped_frame = frame.min(max(mc.frames_loaded(), 0) as FrameNumber);
+        let clamped_frame = if self.movie().is_action_script_3() {
+            // Make sure we don't allow a negative seek. Note that we do
+            // *not* check this against the frame count in the SWF header - flash player
+            // ignores that value and allows you to seek to any frame, even if the total
+            // number of frames exceeds the frame count in the header.
+            frame.min(max(mc.frames_loaded_ignore_header(), 0) as FrameNumber)
+        } else {
+            // Sanity; let's make sure we don't seek way too far.
+            frame.min(max(mc.frames_loaded(), 0) as FrameNumber)
```
Lord-McSweeney, line 1872: Does this mean gotos in AVM1 that go past the frames in the header are clamped? I made a small test and AVM1 also seems to ignore the header frame count in FP.

torokati44:
Possibly distantly related: #6144 and #2705.

adrian17:
Just for completeness, as researched on Discord, this doesn't feel correct; as in: in FP, I can observe that the playhead doesn't stop at the last ShowFrame, but instead it actually processes tags until the last one, and only then it either loops or stops. Ways to observe it:
- with DoABC2 and looping (the tag runs, the script initializer runs; I assume that it runs after the loop reaches the first ShowFrame, because the script can not be observed with a manual `gotoAndStop()`, just with a loop)
- with a control tag like `RemoveObject2`. The requested DO actually gets destroyed, even if after calling `.gotoAndStop()`, the `.currentFrame` looks as-if it stopped on the last "real" ShowFrame.

Also, intuitively, the count-in-advance logic would interact badly with possible async download implementation.
(and just personally, I'm not a fan of parsing tags twice ;) )

Aaron1011:
Closing in favor of https://github.com/ruffle-rs/ruffle/pull/17019 (which avoids counting `ShowFrame` tags)

PR #16737, opened 2024-06-15
core: Don't reborrow UpdateContext constantly, add a new lifetime to Activation to track the borrow

Dinnerbone:
The diff is mostly noise from adding one more lifetime to Activation, making the signature go from `Activation<'a, 'gc>` to `Activation<'player, 'update, 'gc>`.

Should reduce stack size of making activations, and ideally a small performance boost too.

Size of AVM1 `Activation` (on 64-bit) goes from 616 bytes to 120
Size of AVM2 `Activation` (on 64-bit) goes from 760 bytes to 296


Dinnerbone:
Closed in favour of #17345

PR #16741, opened 2024-06-16
avm2: Merge locals better in optimizer

Lord-McSweeney:


core/src/avm2/verify.rs
```diff
@@ -585,25 +588,31 @@ pub fn verify_method<'gc>(
...
-                    potential_jump_targets.insert(adjusted_result.0, JumpSources::Known(vec![i]));
+                    potential_jump_targets.insert(adjusted_result.0, vec![JumpSource::JumpFrom(i)]);
                 }
             }
             AbcOp::LookupSwitch(ref mut lookup_switch) => {
-                // TODO: Add i to possible sources, like in the branch ops
-
                 let adjusted_default = adjust_jump_to_idx(i, lookup_switch.default_offset, false)?;
                 lookup_switch.default_offset = adjusted_default.1;
-                potential_jump_targets.insert(adjusted_default.0, JumpSources::Unknown);
```
adrian17, line 601: do you remember why it was `Unknown` before?

core/src/avm2/verify.rs
```diff
@@ -585,25 +588,31 @@ pub fn verify_method<'gc>(
             | AbcOp::Jump { offset } => {
                 let adjusted_result = adjust_jump_to_idx(i, *offset, true)?;
                 *offset = adjusted_result.1;
-                if let Some(jump_sources) = potential_jump_targets.get_mut(&adjusted_result.0) {
-                    if let JumpSources::Known(sources) = jump_sources {
-                        sources.push(i);
-                    }
+                if let Some(sources) = potential_jump_targets.get_mut(&adjusted_result.0) {
```
adrian17, line 591: Could these all use the Entry api instead of manual `if` and `insert`?
As in, smth like
```rs
potential_jump_targets.entry(&adjusted_result.0)
    .or_default().push(JumpSource::JumpFrom(i));
```

adrian17 (approved):

core/src/avm2/verify.rs
Lord-McSweeney: I didn't know of the right way to implement it before

PR #16746, opened 2024-06-16
avm2: Fix and unify converting SWF's 1-based to 0-based slot indexing

adrian17:
Previously we assumed two possible cases:
- automatic slot assignment in SWF, `getproperty` in bytecode and relying on optimizer: 0-based in Ruffle. Usual case.
- manual slot assignments in SWF, `getslot` in bytecode: 1-based in Ruffle. Quite rare.

What we didn't realize (or at least, I was sure it's not allowed by FP) is that there exists a 3rd case:
- automatic slot assignment in SWF, `getslot` in bytecode. This is where we broke because of mixed indexing bases.

This PR unifies the code by making everything 0-based.


Lord-McSweeney (approved):

Lord-McSweeney:
We should add the verifier checks for `*slot 0`, too

Issue #13161, opened 2023-09-12
Ruffle cannot play the sample 01_HellWord from CrossBridge


Fancy2209:
### Describe the bug

```
Error: null function pointer called
        at Function/C_Run.as$0:anonymous()
        at global/F__ZNSt15basic_streambufIcSt11char_traitsIcEE5sputcEc()
        at global/F__ZSt16__ostream_insertIcSt11char_traitsIcEERSt13basic_ostreamIT_T0_ES6_PKS3_i()
        at global/F__ZStlsISt11char_traitsIcEERSt13basic_ostreamIcT_ES5_PKc()
        at global/F_main()
        at global/F__start1()
        at com.adobe.flascc::CModule$/callI()
        at com.adobe.flascc::CModule$/start()
        at com.adobe.flascc::CModule$/startAsync()
        at com.adobe.flascc::Console/init()
        at com.adobe.flascc::Console()
        at com.adobe.flascc.preloader::DefaultPreloader/onPreloaderComplete()
2023-09-12T14:20:54.082621Z ERROR run_frame:run_all_phases_avm2: ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "complete", class: flash.events::Event, ptr: 0x196602b4d10 }) to handler FunctionObject(FunctionObject { ptr: 0x19660ffc450 }) : Error: null function pointer called
        at Function/C_Run.as$0:anonymous()
        at global/F__ZNSt15basic_streambufIcSt11char_traitsIcEE5sputcEc()
        at global/F__ZSt16__ostream_insertIcSt11char_traitsIcEERSt13basic_ostreamIT_T0_ES6_PKS3_i()
        at global/F__ZStlsISt11char_traitsIcEERSt13basic_ostreamIcT_ES5_PKc()
        at global/F_main()
        at global/F__start1()
        at com.adobe.flascc::CModule$/callI()
        at com.adobe.flascc::CModule$/start()
        at com.adobe.flascc::CModule$/startAsync()
        at com.adobe.flascc::Console/init()
        at com.adobe.flascc::Console()
        at com.adobe.flascc.preloader::DefaultPreloader/onPreloaderComplete()
```
![imagem](https://github.com/ruffle-rs/ruffle/assets/64917206/b727375c-29c1-4c37-9318-5fb229f9a24f)

### Expected behavior

[...]

Issue #11787, opened 2023-06-29
Eager loading in ruffle loads in wrong order


Bale001:
(this issue is the root cause of #10851, and maybe some other crossbridge/alchemy related issues)
This is something we discovered a while ago, but I decided to open an issue for it so we can keep track of it better, and hopefully, we can use this issue to discuss a fix.

The issue is that, currently, in Ruffle, when we encounter a DoABC2 tag with eager init enabled, we initialize every script as we encounter them immediately. Why is this a problem? Because that script might be relying on a symbol class. For example, a class that extends ByteArray might have BinaryData that it's meant to be associated with. If we initialize the script before we link that class to the BinaryData, we will run into major issues.

So clearly, we need to initialize the script after we set SymbolClass's, but what's confusing here is exactly when we are supposed to initialize the script (based on testing, might be during the ShowFrame tag).

Issue #10851, opened 2023-04-27
Error on https://www.gamezhero.com/get-game-code/164bf317ea19ccfd9e97853edc2389f4


lalaisbaby:
# Error Info
Error name: Error
Error message: panicked at 'capacity overflow', library/alloc/src/raw_vec.rs:525:5
Error stack:
```
Error: panicked at 'capacity overflow', library/alloc/src/raw_vec.rs:525:5
    at imports.wbg.__wbg_new_15d3966e9981a196 (chrome-extension://donbcfbmhbcapadipfkeojnmajbakjdc/dist/339.js:2258:21)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::hd28f65a61f685e47 (wasm://wasm/02b61d02:wasm-function[8383]:0x80c16a)
    at std::panicking::rust_panic_with_hook::h7f7102b82d51338f (wasm://wasm/02b61d02:wasm-function[5498]:0x72f1b8)
    at core::panicking::panic_fmt::he4489d678d6570d5 (wasm://wasm/02b61d02:wasm-function[8238]:0x8069e8)
    at alloc::raw_vec::capacity_overflow::h743526e95bf6b28b (wasm://wasm/02b61d02:wasm-function[8014]:0x801def)
    at alloc::raw_vec::RawVec<T,A>::reserve::do_reserve_and_handle::h07da55bc9c349c86 (wasm://wasm/02b61d02:wasm-function[5377]:0x722fb0)
    at ruffle_core::avm2::globals::flash::utils::byte_array::set_length::ha3af8f98056fb6c9 (wasm://wasm/02b61d02:wasm-function[3403]:0x64a16a)
    at ruffle_core::avm2::function::Executable::exec::hb5981ebd76854dce (wasm://wasm/02b61d02:wasm-function[781]:0x2ac733)
    at ruffle_core::avm2::object::TObject::call_method::h058bcfcd584e2d7c (wasm://wasm/02b61d02:wasm-function[1929]:0x4c3f2d)
    at ruffle_core::avm2::object::TObject::set_property::hf7ac531dae35065d (wasm://wasm/02b61d02:wasm-function[2163]:0x510091)
```
AVM2 stack:
```
    at flash.utils::ByteArray/flash::utils::ByteArray::length()
    at global()
    at global()
    at global()
    at global()
    at global()
    at global()
    at global()
    at global()
    at global()
    at global()
    at global()
    at global()
    at global()
[...]

PR #16747, opened 2024-06-16
avm2: Match Flash behavior for DoAbc/SymbolClass script initializers

Aaron1011:
Flash Player handles SymbolClass tags and eager (non-lazy) DoAbc2 tags in an unusual way: During the first time that a given frame is executed:
1. All SymbolClass tags are processed in order, triggering ClassObject loading (and the associated script initializer execution, if it hasn't already been run)
2. All eager (non-lazy) DoAbc/DoAbc2 tags have their *final* script initializer executed.

To implement this, we now cache the needed `Script` and symbol class data during preloading. The first time we execute a frame, we run the symbolclass class lookup and assignment, followed by the final script initializer for each DoAbc2 tag in the frame.

This fixes 'minidash' (and probably other crossbridge games as well).


Dinnerbone:
When combined with #16730, this fixes #13161, #11787, #10851

Dinnerbone (approved):

Issue #441, opened 2020-03-15
Run app in Flex


eduardo12fox:
Hello Guys,
This solution run app make in Adobe Flex?

kmeisthax:
Not yet. I *just* made a very preliminary PR #404 which adds very basic AVM2/AS3 support. It doesn't even support display objects yet, or even coercions and math. It will take several months or more before Flex works in Ruffle.

bmiller-melink:
All,
Has there been any movement on the Flex/AVM2 front?  I've got a site that could be good for testing if you need/want one:
http://energy.melinkcorp.com/melink/html/mpower/MPowerFlash.html

Herschel:
Closing in favor of #1368.

Lord-McSweeney:
This currently needs:
```
flash.events.PressAndTapGestureEvent
flash.text.TextField.insertXMLText
```

Lord-McSweeney:
This now errors:
```
Error dispatching event EventObject(EventObject { type: "complete", class: flash.events::Event, ptr: 0x2a261b0 }) to handler FunctionObject(FunctionObject { ptr: 0xe831b0 }) : Error: The channel class 'mx.messaging.channels.AMFChannel' specified was not found.
	at mx.messaging.config::ServerConfig$/createChannel()
	at mx.messaging.config::ServerConfig$/getChannel()
	at com.belltracy.util::URLUtil$/get rootServerURI()	at com.belltracy.util::URLUtil$cinit()
	at global$init() [TU=com/belltracy/util/URLUtil]
	[massive stack trace]
```

Lord-McSweeney:
Still happening as of August 27.

Lord-McSweeney:
This has been significantly improved, but now throws a different error:
```
Error dispatching event EventObject(EventObject { type: "result", class: mx.rpc.events::ResultEvent, ptr: 0x670d428 }) to handler FunctionObject(FunctionObject { ptr: 0x45052b8 }) : TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: toString)
	at com.anychart.utils::XMLUtils$/merge()
	at com.anychart.utils::XMLUtils$/merge()
	at com.anychart.styles::Style/mergeState()
	at com.anychart.styles::Style/deserialize()
	at com.anychart.elements::BaseElement/deserializeStyleFromNode()
	at com.anychart.seriesPlot.elements.markerElement::MarkerElement/deserializeGlobal()
[...]

PR #16748, opened 2024-06-16
avm2: Allow ShaderParameter value to be null

Aaron1011:
This makes progresss in on #441


evilpie:
I am not sure if Flash actually copies the array.
```
var param = new ShaderParameter();
var array = [.1, .1, .1, .1];
param.array = array;
trace(param.array == array); // true
trace(param.array == param.array); // true
```

Lord-McSweeney:
Superseded by #20759

PR #16750, opened 2024-06-17
extension: Fix 4399 CORS and Flash detection issue  (close #13943)

### LINKED ISSUE #13943: Cannot run flash games under www.4399.com
### Describe the bug

Hi. The https://www.4399.com/ is a famous flash game website (https://www.similarweb.com/website/4399.com/#overview). I found the flash games under the site cannot play by using Ruffle. Most of them are using ActionScript 3. Do you have any idea to support them?

For example:
https://news.4399.com/aoyazhiguang/
https://www.4399.com/flash/18012.htm

Thank you.

### Expected behavior

The flash games can run by Ruffle.

### Content Location

https://news.4399.com/aoyazhiguang/
https://www.4399.com/flash/18012.htm

### Affected platform

Browser's extension

### Operating system

Windows10, Windows11

### Browser

Edge, Chrome

### Additional information

_No response_

riku-42:
Fixes #13943 

### Background
Currently site 4399.com doesn't support Flash game even if user already installed Ruffle extension. The site has some issues with its CORS config and Flash detection. This PR fix these issues by:

1. Setting `Access-Control-Allow-Origin` header to `https://www.4399.com` when requesting swf files.
2. Injecting script to make the Flash detection invalid.


Croworbit:
I don't think you can just forcibly enable CORS like this

riku-42:
I added some change. Now the header modification only works for XML request to swf file.

web/packages/extension/manifest_other.json5
```diff
@@ -50,11 +55,20 @@
     "permissions": [
         "storage",
         "scripting",
+        "declarativeNetRequest",
+        "declarativeNetRequestFeedback",
```
danielhjacobs, line 59: Is this permission needed in the release manifest? https://developer.chrome.com/docs/extensions/reference/api/declarativeNetRequest#permissions says it's used for debugging features.

web/packages/extension/4399_rules.json
```diff
@@ -0,0 +1,21 @@
+[
+  {
+    "id": 1,
+    "priority": 1,
+    "condition": {
+      "domains": ["4399.com"],
```
danielhjacobs, line 6: `domains` is deprecated: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/declarativeNetRequest/RuleCondition#domains. This should probably use `initiatorDomains` instead, and it may be a good idea to further restrict it by `requestDomains`

web/packages/extension/manifest_other.json5
riku-42: Thanks, I have removed it correspondingly.

web/packages/extension/4399_rules.json
riku-42: Updated to `initiatorDomains`. `requestDomains` may not work for this scenario because this site hosts swf files on many different CDNs. I have set `urlFilter` to filter out requests to swf files only.

Dinnerbone (approved):
Thank you! I think in the future we'll want to port this to Firefox too, and make it disableable in UI - but this is good for this PR.

PR #16759, opened 2024-06-17
ci: Switch back to Windows 2019 to avoid constant crashes

torokati44:
Others said it worked for them, let's see...


torokati44:
Aaand it's a big fat nopey nope again... :/

Issue #16715, opened 2024-06-14
OS level crash in Windows as of 6/14/24 when invoking the DX12 renderer


sombraguerrero:
### Describe the bug

As of the 6/14/24 nightly, I am experiencing a crash immediately upon trying to invoke the DX12 renderer.

The Ruffle logging outputs nothing beyond the same warnings it has about the downlevels since 1/22/24. There is absolutely no useful output, so that lead me to ultimately compile in debug mode and attach a debugger. The stacktrace is as follows:
```
[0x0]   ntdll!RtlInitAnsiStringEx+0x14   0x50ff5b4928   0x7ffe087f24ed   
[0x1]   KERNELBASE!DeleteFileA+0x5d   0x50ff5b4930   0x7ffe0883fbb9   
[0x2]   KERNELBASE!GetFullPathNameA+0x39   0x50ff5b4970   0x7ffe01327952   
[0x3]   d3dcompiler_47!D3DReturnFailure1+0xe012   0x50ff5b49f0   0x7ffe01327894   
[0x4]   d3dcompiler_47!D3DReturnFailure1+0xdf54   0x50ff5b4a20   0x7ffe013174ee   
[0x5]   d3dcompiler_47!D3DAssemble+0xc4e   0x50ff5b4a70   0x7ffe01317059   
[0x6]   d3dcompiler_47!D3DAssemble+0x7b9   0x50ff5b6e40   0x7ffe01317249   
[0x7]   d3dcompiler_47!D3DAssemble+0x9a9   0x50ff5b7160   0x7ffe0131856a   
[0x8]   d3dcompiler_47!D3DCompile2+0xea   0x50ff5b7220   0x7ffe0131846c   
[0x9]   d3dcompiler_47!D3DCompile+0x6c   0x50ff5b72a0   0x7ff7d95c3c27   
[0xa]   ruffle_desktop!wgpu_hal::dx12::shader_compilation::compile_fxc+0x297 [C:\Users\rsett\.cargo\registry\src\index.crates.io-6f17d22bba15001f\wgpu-hal-0.20.0\src\dx12\shader_compilation.rs @ 38]   0x50ff5b7320   0x7ff7da30a9ad   
[...]

PR #16764, opened 2024-06-18
ci: Release and test with Rust 1.78 instead of stable

torokati44:
At least until a release with https://github.com/gfx-rs/wgpu/pull/5831 is out.

Fixes https://github.com/ruffle-rs/ruffle/issues/16715

### DIFF
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -108,7 +108,9 @@ jobs:
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: stable
+          # TODO: Return to latest "stable" once this is released:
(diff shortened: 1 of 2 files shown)


danielhjacobs (approved):

kjarosh (approved):

torokati44:
Then we have to make sure to build the nightlies with this too.

torokati44:
I opted to switch all desktop releases to 1.78 because it's a smaller/simpler change, and doesn't really make a difference otherwise.

kjarosh:
Shouldn't we switch all tests to 1.78 too for consistency? Otherwise we will release using a different version than the one used for testing.

torokati44:
Fair point. That's also fine by me. Feel free to amend-commit-force-push. Any counter-opinions?

PR #16784, opened 2024-06-19
text: Restructure layout to include text lines

kjarosh:
This PR neither fixes any bugs nor implements new features. The main goal is to refactor the `layout` module, so that it's not _that_ painful to use.

This PR removes `LineData` & `get_line_data` and incorporates their logic into layout calculation adding `Layout` and `LayoutLine`. The goal is to simplify operations on the layout, as now line data is always present and integrated with the layout. Specialized methods may be added to `Layout` and `LayoutLine` in the future.

This refactor opens up possibilities for further improvements:
* providing binary-search-based layout box lookup,
* simplifying `screen_position_to_index`,
* implementing vertical text control,
* implementing advanced text API for AVM2.

This PR already includes some improvements:
* `EditText.layout_metrics` refactor — layout boxes no longer need to be traversed, as layout bounds and line bounds are calculated during layout,
* `EditText.line_text` refactor — only layout boxes from the given line need to be traversed, not all of them.

### DIFF
--- a/core/src/display_object/edit_text.rs
+++ b/core/src/display_object/edit_text.rs
@@ -23,7 +23,7 @@ use crate::events::{ClipEvent, ClipEventResult, TextControlCode};
 use crate::font::{round_down_to_pixel, FontType, Glyph, TextRenderSettings};
 use crate::html;
 use crate::html::{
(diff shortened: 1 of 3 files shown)


Dinnerbone (approved):
Thank you!

PR #16793, opened 2024-06-20
web: Core UI improvements

WumboSpasm:
| Before | After (Light) | After (Dark) |
|---|---|---|
| image | image | image |
| image | image | image |
| image | image | image |
| image | image | image |

Other changes of note:
* The context menu is no longer bound to the edges of the player
* Hovering over a context menu item no longer changes the cursor, which should bring it in line with most traditional context menus
* The player is now darkened and the browser context menu is suppressed while a modal is open
* Modals now stay in the center of the player more reliably, even if the player width is smaller than the modal width

Tasks:
- [x] Context menu
- [x] Volume controls modal
- [x] Save manager modal
- [x] Clipboard modal
- [x] Video modal
- [x] Hardware acceleration modal

I plan on making a separate PR for player overlay improvements.


evilpie:
Should we also add rounded corners to the context menu?

Dinnerbone:
Personally I like square context menu, the window rounding is quite strong. Maybe a subtle one would look okay, but we can discuss that for a followup PR?

Dinnerbone (approved):

torokati44:
> Personally I like square context menu, the window rounding is quite strong. 

Same. The radius that Windows 11 has on either, is too much for my taste. Not to mention Android, even the screens! 😵

The ones in KDE feel just right to me:
image

But I don't mind actual corners either, for sure.

Dinnerbone:
Thank you!

Issue #8891, opened 2022-12-22
not support one flv video


ivysrono:
### Describe the bug

https://zzb.wuxi.gov.cn/doc/2020/12/31/3158429.shtml

Click `Run Anyway`, get error:

`Error dispatching event EventObject(EventObject { type: "timerComplete", class: flash.events::TimerEvent, ptr: 0x2ad388 }) to handler FunctionObject(FunctionObject(GcCell(Gc { ptr: 0x6e18dc }))) : RustError("Property does not exist: Multiname { ns: Single(Package(\"\")), name: Some(\"unescape\"), params: [], flags: (empty) }") `

### Expected behavior

Play the video.

### Affected platform

Browser's extension

### Operating system

Win11

### Browser

Firefox 108

n0samu:
Ruffle does not support loading external video files yet. See #3801.

Also, the video file on this page uses the H.264 codec, which is unlikely to be supported by Ruffle anytime soon due to patent/licensing restrictions. Ruffle would need to use the browser's native video playback capabilities, which would be particularly difficult to implement.

Lord-McSweeney:
This also seems to need `NetStream`.

torokati44:
At least on desktop, this will be partially fixed by https://github.com/ruffle-rs/ruffle/pull/14654.
For web, I have a WIP (but already functional) branch on top of that PR that delegates H.264 decoding to the browser through the WebCodecs API where supported.

For audio, the AAC-LC audio codec will also have to be wired up (probably/preferably from Symphonia).

torokati44:
The video now plays, complete with sound, both on desktop and in the browser (given WebCodecs with H.264 support).

PR #16794, opened 2024-06-20
web,video: Add H264 decoding using the WebCodecs API

torokati44:
A continuation of https://github.com/ruffle-rs/ruffle/pull/14654.

Not really done yet, just pushing for visibility, and so that it doesn't get forgotten.
~_(Just look at the commit history lol.)_~

Due to this API being based on callbacks, this currently adds at least one additional (video) frame of delay as compared to a synchronous decoder, such as OpenH264 on desktop. I think this should be acceptable.
It would be possible to wire the callback straight from the browser through the video backend to the renderer backend, to update the texture immediately (off the regular frame pacing schedule), but I think that would be a bit too complicated, and could potentially result in less smooth playback.

Mostly works, in Chrome and in Firefox 129+.

I don't think this could be problematic at all from a patents/licensing/royalty standpoint.

Advances #8891.

### DIFF
--- a/video/external/src/decoder/webcodecs.rs
+++ b/video/external/src/decoder/webcodecs.rs
@@ -1,0 +1,287 @@
+use std::cell::RefCell;
+use std::rc::Rc;
+use std::sync::Arc;
+
+use crate::decoder::VideoDecoder;
+
+use ruffle_render::bitmap::BitmapFormat;
+use ruffle_video::error::Error;
+use ruffle_video::frame::{DecodedFrame, EncodedFrame, FrameDependency};
+
+use js_sys::Uint8Array;
+use tracing::{debug, trace, warn};
+use tracing_subscriber::{layer::Layered, Registry};
+use tracing_wasm::WASMLayer;
+use wasm_bindgen::prelude::*;
+use web_sys::{
+    DomException, EncodedVideoChunk, EncodedVideoChunkInit, EncodedVideoChunkType,
+    VideoDecoder as WebVideoDecoder, VideoDecoderConfig, VideoDecoderInit, VideoFrame,
+    VideoPixelFormat,
+};
+
+// Abbreviations used:
+//  - NAL: Network Abstraction Layer
+//  - NALU: NAL unit
+//  - VCL: Video Coding Layer
+//  - SPS: Sequence Parameter Set
+//  - PPS: Picture Parameter Set
+//  - IDR: Instantaneous Decoding Refresh
+//  - SEI: Supplemental enhancement information
+
+// NALU type 5 means IDR frame - basically a keyframe.
+const NALU_TYPE_IDR: u8 = 5;
+
+fn js_error_to_decoder_error(js_error: JsValue) -> Error {
+    Error::DecoderError(
+        js_error
+            .dyn_ref::<js_sys::Error>()
+            .unwrap()
+            .message()
+            .as_string()
+            .unwrap()
+            .into(),
+    )
+}
+
+pub struct H264Decoder {
+    /// How many bytes are used to store the length of the NALU (1, 2, 3, or 4).
+    length_size: u8,
+
+    /// The WebCodecs decoder object.
+    decoder: WebVideoDecoder,
+
+    /// The decoder output callback writes this, and the decode_frame method reads it.
+    ///
+    /// This in itself results in one frame of delay (because we can't block decode_frame
+    /// until the callback is invoked), but it shouldn't matter in practice.
+    last_frame: Rc<RefCell<Option<DecodedFrame>>>,
+
+    // Simply keeping these objects alive, as they are used by the JS side.
+    // See: https://rustwasm.github.io/wasm-bindgen/examples/closures.html
+    #[allow(dead_code)]
+    output_callback: Closure<dyn Fn(VideoFrame)>,
+    #[allow(dead_code)]
+    error_callback: Closure<dyn Fn(DomException)>,
+}
+
+impl H264Decoder {
+    /// `extradata` should hold "AVCC (MP4) format" decoder configuration, including PPS and SPS.
+    /// Make sure it has any start code emulation prevention "three bytes" removed.
+    ///
+    /// The log_subscriber is needed so that we have proper logging from within the callbacks.
+    pub fn new(log_subscriber: Arc<Layered<WASMLayer, Registry>>) -> Result<Self, Error> {
+        let last_frame = Rc::new(RefCell::new(None));
+        let lf = last_frame.clone();
+
+        let log_subscriber_for_output = log_subscriber.clone();
+        let output = move |output: &VideoFrame| {
+            let _subscriber = tracing::subscriber::set_default(log_subscriber_for_output.clone());
+            let visible_rect = output.visible_rect().unwrap();
+
+            match output.format().unwrap() {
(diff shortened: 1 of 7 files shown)


torokati44:
Rebased on top of https://github.com/ruffle-rs/ruffle/pull/17878.

video/external/src/decoder/webcodecs.rs
```diff
@@ -0,0 +1,277 @@
...
+                    }
+                    last_frame.replace(Some(DecodedFrame::new(
+                        visible_rect.width() as u32,
+                        visible_rect.height() as u32,
+                        BitmapFormat::Rgba,
+                        data,
+                    )));
+                }
+                _ => {
+                    panic!("unsupported pixel format: {:?}", output.format().unwrap());
```
danielhjacobs, line 99: Admittedly I'm not great with Rust, but does it make sense to panic if this is any other format?

https://developer.mozilla.org/en-US/docs/Web/API/VideoFrame/format and https://rustwasm.github.io/wasm-bindgen/api/web_sys/enum.VideoPixelFormat.html lists a lot more formats than just I420 and BGRX. Is it possible to embed a video in any of those other formats with Flash, and if so, wouldn't the whole thing panic when trying to play them.

torokati44: Yeah, good point, this is not nice.

torokati44: Replaced the panic with a simple error log. We can add conversion from other formats as needed.
I think what we get here depends on the browser more than the encoded content.
These two were chosen because they were returned by Chrome and Firefox in my testing, on my systems.

web/src/builder.rs
```diff
@@ -646,7 +646,9 @@ impl RuffleInstanceBuilder {
         let core = builder
             .with_log(log_adapter::WebLogBackend::new(trace_observer.clone()))
             .with_ui(ui::WebUiBackend::new(js_player.clone(), &canvas))
-            .with_video(SoftwareVideoBackend::new())
```
thy486, line 649: Is it possible to make a simple judgment to confirm whether the WebCodes API is supported?

These are my little thoughts, thank you for your contribution!


```rust
let is_webcodecs_support = { web_sys::window().and_then(|window| window.get("VideoDecoder")) }.is_some();
if is_webcodecs_support {
    core_builder = core_builder.with_video(ExternalVideoBackend::new_with_webcodecs(
        log_subscriber.clone(),
    ));
} else {
    core_builder = core_builder.with_video(SoftwareVideoBackend::new());
}
```

torokati44: The end result would be the same - `ExternalVideoBackend` has an internal `SoftwareVideoBackend` wrapped, that it uses for any non-H.264 video. And if WebCodecs is not available, both fail to create a decoder for H.264.

adrian17: Can you add it as a comment there?

video/external/src/decoder/webcodecs.rs
```diff
@@ -0,0 +1,277 @@
...
+                    continue;
+                }
+            };
+            trace!("frame type: {:?}", frame_type);
+
+            // The timestamp doesn't matter for us.
+            let init = EncodedVideoChunkInit::new(&Uint8Array::from(nalu), 0.0, frame_type);
+            let chunk = EncodedVideoChunk::new(&init).unwrap();
+
+            let _ = self.decoder.decode(&chunk);
```
adrian17, line 266: same here. What if it throws?

video/external/src/decoder/webcodecs.rs
```diff
@@ -0,0 +1,277 @@
...
+            configuration_data[1], configuration_data[2], configuration_data[3]
+        );
+        let config = VideoDecoderConfig::new(&codec_string);
+        trace!("decoder state: {:?}", self.decoder.state());
+        trace!("configuring decoder with: {:?}", &configuration_data[1..4]);
+
+        let data = Uint8Array::from(configuration_data);
+        config.set_description(&data);
+        config.set_optimize_for_latency(true);
+        let _ = self.decoder.configure(&config);
```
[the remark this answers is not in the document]
torokati44: (because I was lazy :face_with_peeking_eye: but you are right, it should be handled, or at least logged!)

video/external/src/decoder/webcodecs.rs
torokati44: who knows! :shrug: (will fix later)

torokati44:
Thank you for the review, @adrian17, I think I've resolved your notes.

torokati44:
Oh, and also, I guess a tracing subscriber for the other callback too can't hurt... :thinking: 

image

torokati44:
And with github.com/ruffle-rs/ruffle/pull/16526 having been long merged, this now fixes https://github.com/ruffle-rs/ruffle/issues/8891 entirely as far as I'm concerned...
(Trying to seeking past the currently buffered portion doesn't do anything, but I haven't tested what FP does then with this player. Also, for testing, I still found it helpful to block the network request for that one `/CA111...001.js` file that just wouldn't load for me.)

video/external/src/decoder/webcodecs.rs
```diff
@@ -0,0 +1,287 @@
...
+impl H264Decoder {
+    /// `extradata` should hold "AVCC (MP4) format" decoder configuration, including PPS and SPS.
+    /// Make sure it has any start code emulation prevention "three bytes" removed.
+    ///
+    /// The log_subscriber is needed so that we have proper logging from within the callbacks.
+    pub fn new(log_subscriber: Arc<Layered<WASMLayer, Registry>>) -> Result<Self, Error> {
+        let last_frame = Rc::new(RefCell::new(None));
+        let lf = last_frame.clone();
+
+        let log_subscriber_for_output = log_subscriber.clone();
```
adrian17, line 76: To be sure, putting the `log_subscriber` inside the `Decoder` wouldn't have made the boilerplate any simpler? Seems like what we're doing in most other backends.

Also, does it use a custom target by default? As in, is there any magic to enable `trace` level only for this decoder and other log sources - or the other way around?

adrian17 (approved):

video/external/src/decoder/webcodecs.rs
torokati44: > To be sure, putting the log_subscriber inside the Decoder wouldn't have made the boilerplate any simpler? Seems like what we're doing in most other backends.

The problem is that unlike other decoders, this one (unfortunately) has to operate via callbacks from JS. So, we need to give each closure its own `Arc` for the `log_subscriber` to keep anyway. If, by boilerplate, you mean that bunch of `.clone()`s.

> Also, does it use a custom target by default? As in, is there any magic to enable trace level only for this decoder and other log sources - or the other way around?

Well, since we don't supply any target explicitly, the macros use the module path as a default:
https://github.com/tokio-rs/tracing/blob/7a3ddfb02ac9b6e6f9c6bbe75feac2f2f3af91df/tracing/src/macros.rs#L89-L97 Again, if I understood your question right...

torokati44: Are you satisfied with my responses, @adrian17? Enough to press the Big Green Button maybe?

torokati44:
I think this has marinated enough. Merging, bearing the consequences.

Issue #16642, opened 2024-06-08
Crash on Aquanaut.swf


Akan9021:
### Describe what you were doing

Aquanaut just crashes on startup before even loading any level.

### What does the crash message say?

panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.20.0/src/backend/wgpu_core.rs:2996:5:
wgpu error: Validation Error

Caused by:
    In Device::create_render_pipeline
    Internal error in ShaderStages(VERTEX) shader: 0:42(30): error: syntax error, unexpected PACKED_TOK, expecting ')'

### Ruffle Version

0.1.0-nightly (6e53f980687cbc00ca5b505ff7a74df9a1a8c2a3 2024-06-08)

### Affected platform

Desktop app

### Operating system

EndeavourOS Rolling Release (rolling) [64-bit]

### Additional information

Filename: Aquanaut.swf

### Render Info
Renderer: wgpu
Adapter Backend: Gl
Adapter Name: "Mesa Intel(R) UHD Graphics 605 (GLK 3)"
Adapter Device Type: IntegratedGpu
Adapter Driver Name: "OpenGL"
Adapter Driver Info: "4.6 (Compatibility Profile) Mesa 24.1.0-arch1.1"
Enabled features: Features(TEXTURE_COMPRESSION_BC | FLOAT32_FILTERABLE | TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES | SHADER_UNUSED_VERTEX_OUTPUT)
Available features: Features(DEPTH_CLIP_CONTROL | DEPTH32FLOAT_STENCIL8 | TEXTURE_COMPRESSION_ETC2 | TEXTURE_COMPRESSION_ASTC | TIMESTAMP_QUERY | INDIRECT_FIRST_INSTANCE | TIMESTAMP_QUERY_INSIDE_ENCODERS | TIMESTAMP_QUERY_INSIDE_PASSES | PUSH_CONSTANTS | ADDRESS_MODE_CLAMP_TO_ZERO | ADDRESS_MODE_CLAMP_TO_BORDER | POLYGON_MODE_LINE | POLYGON_MODE_POINT | VERTEX_WRITABLE_STORAGE | CLEAR_TEXTURE | SHADER_PRIMITIVE_INDEX | SHADER_EARLY_DEPTH_TEST | DUAL_SOURCE_BLENDING)
[...]

Issue #12844, opened 2023-08-19
Crash on castaway-2-11261.swf


Lord-McSweeney:
### Describe what you were doing

Consistently panics after the Armor Games intro.
See #12843 for the SWF.
Note that this only occurs on the OpenGL backend for me, on Linux.

### What does the crash message say?

panicked at 'wgpu error: Validation Error

Caused by:
    In Device::create_render_pipeline
    Internal error in ShaderStages(VERTEX) shader: 0:34(30): error: syntax error, unexpected PACKED_TOK, expecting ')'

', /home/uqers/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.16.3/src/backend/direct.rs:3019:5

### Ruffle Version

0.1.0-nightly (d21520a4c76e6f226ad1550e0c6eee262527014f 2023-08-19)

### Affected platform

Desktop app

### Operating system

Ubuntu 20.04 (focal) [64-bit]

### Additional information

Filename: castaway-2-11261.swf

### AVM2 Callstack
```
	at flash.display::BitmapData/flash::display::BitmapData::applyFilter()
	at net.flashfilter.as3.effects::E01_raindrop/render()
	at net.flashfilter.as3.effects::E01_raindrop/init()
	at castaway2_fla::likwidIntroMC_1930/frame1()
	at flash.display::MovieClip/flash::display::MovieClip::gotoAndPlay()
	at castaway2_fla::ag_intro_mc_1925/frame225()
```

### Render Info
Renderer: wgpu
Adapter Backend: Gl
Adapter Name: "Mesa Intel(R) UHD Graphics (ICL GT1)"
Adapter Device Type: IntegratedGpu
Adapter Driver Name: ""
Adapter Driver Info: ""
Enabled features: Features(TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES | PUSH_CONSTANTS)
Available features: Features(DEPTH_CLIP_CONTROL | TEXTURE_COMPRESSION_BC | TEXTURE_COMPRESSION_ETC2 | TEXTURE_COMPRESSION_ASTC | ADDRESS_MODE_CLAMP_TO_ZERO | ADDRESS_MODE_CLAMP_TO_BORDER | VERTEX_WRITABLE_STORAGE | CLEAR_TEXTURE | SHADER_PRIMITIVE_INDEX | SHADER_EARLY_DEPTH_TEST)
[...]

PR #16805, opened 2024-06-21
wgpu: Do not use 'packed' as variable name in WGSL shaders

kjarosh:
Fixes https://github.com/ruffle-rs/ruffle/issues/12844, fixes https://github.com/ruffle-rs/ruffle/issues/16642.

Turns out that 'packed' is a keyword in some GLSL shaders, and naga currently preserves it, producing a syntax error.

As a workaround, this patch renames the variable to something else.


Dinnerbone:
This *should* be something that naga deals with, we should raise an issue there too.
A workaround is good for now though!

kjarosh:
Raised an issue about that: https://github.com/gfx-rs/wgpu/issues/5853

adrian17 (approved):

PR #16808, opened 2024-06-21
extension: Upgrade Firefox to Manifest V3

danielhjacobs:
This works. ~~The only reason it's a draft is because the onboarding page looks awful.~~

Manifests have been merged, please test on Chrome as well. If you have Firefox 126 (one below stable), test there too, as it treats host_permissions differently.


evilpie:
I thought that with Firefox 127 the installation prompt will now include all permissions?

> Beginning with Firefox 127, users will be prompted to grant MV3 host permissions as part of the install flow (similar to MV2 extensions). We’re excited to deliver this work as based on feedback from Firefox users and extension developers, this has been a major hurdle for MV3 extensions in Firefox.

https://blog.mozilla.org/addons/2024/05/14/manifest-v3-updates/

danielhjacobs:
Interesting. Surely we want to keep this working on older Firefox versions though, right? The onboarding page only appears if hasAllUrlsPermission returns false.

danielhjacobs:
Also the new permissionsButton popup is still nice because it handled the described situation where "MV3 extensions should still leverage the permissions API to ensure that the permissions required are already granted."

web/packages/extension/src/utils.ts
```diff
@@ -184,3 +91,32 @@ export async function getExplicitOptions(): Promise<Options> {
 
     return options;
 }
+
+export const hasAllUrlsPermission = async () => {
+    const allPermissions = await permissions.getAll();
+    if (Object.prototype.hasOwnProperty.call(allPermissions, "origins")) {
```
evilpie, line 97: What are you trying to protect against here? `allPermissions.origins?.includes("<all_urls>")` should work.

web/packages/extension/src/utils.ts
```diff
@@ -184,3 +91,32 @@ export async function getExplicitOptions(): Promise<Options> {
...
+};
+
+export async function hasHostPermissionForActiveTab() {
+    const [activeTab] = await tabs.query({
+        active: true,
+        currentWindow: true,
+    });
+
+    try {
+        return activeTab?.url
```
evilpie, line 113: I was under the impression that having access the `url` property would imply a host permission?

danielhjacobs: This was actually code I sort of swiped: https://discourse.mozilla.org/t/detect-whether-extension-has-host-permission-for-active-tab/120501/3. I guess it was unnecessarily complex.

danielhjacobs: Yeah, that was only needed with the tabs permission, which we don't grant.

danielhjacobs: Though as noted in popup.ts, `activeTab.url` returns `undefined` on Chrome as it requires the `tabs` permission, so my else is needed.

danielhjacobs: Reverted this back, as Firefox lets you access the current tab's URL at all times.

Dinnerbone (approved):
Thank you!

web/packages/extension/src/popup.ts
```diff
@@ -187,5 +186,30 @@ window.addEventListener("DOMContentLoaded", () => {
...
+    );
+    const url = activeTab?.url ? new URL(activeTab.url) : null;
+    if (
+        url &&
+        ["https:", "http:"].includes(url.protocol) &&
+        !(await utils.hasHostPermissionForActiveTab())
+    ) {
+        permissionsButton.classList.remove("hidden");
+        permissionsButton.addEventListener("click", async () => {
+            const grant = await utils.permissions.request({
```
evilpie, line 204: I think there is actually a way of doing this reliably by sending a message to the background script, which requests the permission and reloads the tab on success.

(I only tested this in Firefox)

web/packages/extension/assets/popup.html
```diff
@@ -34,6 +34,7 @@
             </div>
         </div>
         <div class="buttons-container">
+            <button class="hidden" id="permissions-button">Permanently enable for this site</button>
```
evilpie, line 37: I wonder if we should highlight this button a bit more, so it's at least more prominent than the "open SWF player" button?

evilpie (approved):
I think this is fine to land as-is. Users running without <all_urls> are probably used to pain.

web/packages/extension/src/popup.ts
evilpie: I thought that Firefox would forward the user activation via sendMessage, but it seems like it probably doesn't. Still not sure why I thought this worked in my testing ...

danielhjacobs: Discussion on Discord leads me to believe there's no easy way to do this in Firefox right now. If anyone thinks of something that can be a follow up PR.

danielhjacobs: See https://github.com/ruffle-rs/ruffle/pull/16808/commits/9152551e245509d4d33494c4e6754986ca2997e8

web/packages/extension/assets/popup.html
danielhjacobs: This can be a follow-up. I'm not good with design. I did move this button to a separate `buttons-container` div, so that it ends up on a separate line explicitly rather than just by a quirk of the CSS.

evilpie (approved):
Looks ok. If we get complaints about reloading tabs we can still fix it.

PR #16853, opened 2024-06-23
avm2: Parse and load DoAbc/DoAbc2 tags during frame execution

Aaron1011:
Parsing and loading the definitions from an DoAbc/DoAbc2 tag is visible to ActionScript - any definitions will become accessible, and the script initializer will be run when accessing any of those definitions.

We now delay the actual parsing/loading traits from DoAbc2 tags. This is run before we process any SymbolClass tags, and only runs the first time that we execute a frame.

This also fixes an issue where we would incorrectly produce a `VerifyError` during preloading when trying to load a superclass for a `Class` for a tag in a future frame. That frame might require other class definitions to be populated first (e.g. by a `Loader` in the first frame).


core/src/display_object/movie_clip.rs
```diff
@@ -4267,6 +4267,7 @@ impl<'gc, 'a> MovieClip<'gc> {
 
     // Flash Player handles SymbolClass tags and eager (non-lazy) DoAbc2 tags in an unusual way:
     // During the first time that a given frame is executed:
+    // 1. All Abc/DoAbc2 tags have their ABC files parsed and loaded. No script initializers are run yet.
```
Lord-McSweeney, line 4270: The list ordering should be updated to 1, 2, 3 here now.

Lord-McSweeney (approved):

Dinnerbone (approved):

Issue #16859, opened 2024-06-24
desktop app input regression since nightly 2024-01-17


zjk261:
### Describe the bug

while pressing shift whatever input has keycode 0
while pressing other keys, pressing shift causes keyup not detected

https://github.com/ruffle-rs/ruffle/assets/98996990/c2b03216-bfca-4cb9-bf59-ed2c1fe75e7e

### Expected behavior

https://github.com/ruffle-rs/ruffle/assets/98996990/d4b8411e-ab7f-4843-9a6c-a2f878c86966

### Content Location

[Main.zip](https://github.com/user-attachments/files/15955499/Main.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

browser extension is not affected
i think this may be caused by #14703

kjarosh:
The reason is that `winit_to_ruffle_key_code` translates only `Key::Character` with lowercase letters (e.g. `Key::Character("a")`), and the character inside is uppercase when Shift is down (e.g. `Key::Character("A")`).

zjk261:
https://github.com/ruffle-rs/ruffle/assets/98996990/68f971b1-4d50-4ad0-8db1-93d589feb4c1

i found another bug
when the window is not focused, focus on it sends a keycode 133
also a regression since nightly 2024-01-17

kjarosh:
> i found another bug
> when the window is not focused, focus on it sends a keycode 133
> also a regression since nightly 2024-01-17

That unfortunately I cannot reproduce, perhaps it's a Windows-specific bug. Could you file a separate issue for that?

PR #16868, opened 2024-06-24
desktop: Fix mapping input

kjarosh:
Fixes https://github.com/ruffle-rs/ruffle/issues/16859.

This PR fixes mapping input on desktop.

1. FP ignores <kbd>AltGr</kbd> as a key, and issues keys based on the character produced by its combination with other characters.
2. FP recognizes modifiers for numpad keys, but ignores them for non-numpad keys, so that pressing <kbd>Shift</kbd>+<kbd>3</kbd> produces key codes for `Shift` and `Key3`, but pressing <kbd>Shift</kbd>+<kbd>Numpad 7</kbd> produces keycodes for `Shift` and `Home`.
3. This PR fixes handling alphabetic inputs, so that uppercase letters are not ignored.


desktop/src/app.rs
```diff
@@ -341,13 +341,13 @@ impl App {
                                 _ => (),
                             }
 
-                            let key_code = winit_to_ruffle_key_code(&event);
+                            let key_code = winit_to_ruffle_key_code(dbg!(&event));
```
torokati44, line 344: Is this `dbg!` meant to be left in here?

kjarosh: Whoops, those `dbg!`s do be sneaky!

desktop/src/util.rs
```diff
@@ -118,22 +130,17 @@ pub fn winit_to_ruffle_key_code(event: &KeyEvent) -> Option<KeyCode> {
...
-        Key::Character("/") if event.location == KeyLocation::Numpad => KeyCode::NumpadSlash,
-        Key::Character("/") => KeyCode::Slash,
-        Key::Character("`") => KeyCode::Grave,
-        Key::Character("[") => KeyCode::LBracket,
-        Key::Character("\\") => KeyCode::Backslash,
-        Key::Character("]") => KeyCode::RBracket,
-        Key::Character("'") => KeyCode::Apostrophe,
-        Key::Character("*") => KeyCode::Multiply,
-        Key::Character("+") => KeyCode::Plus,
+        Key::Character(";") | Key::Character(":") => KeyCode::Semicolon,
```
evilpie, line 107: Does this make sense? We are mapping logical key codes to something totally different.

danielhjacobs: My main concern is that this is very much tied to a QWERTY keyboard layout. What happens with a different keyboard layout in Flash Player?

kjarosh: This does not make sense at all, but does FP make sense?

1. (English US) When pressing <kbd>Shift</kbd> + <kbd>3</kbd>, you get codes 16 (for Shift) and 51 (for 3).
2. (English UK) When pressing <kbd>Shift</kbd> + <kbd>3</kbd>, you get codes 16 (for Shift) and 163 (for £, because there is <kbd>£</kbd> on UK).
3. (English US) When pressing <kbd>Backslash</kbd>, you get 220 (for backslash).
4. (English UK) When pressing <kbd>Backslash</kbd>, you get 51 (for key 3, because there is <kbd>#</kbd> on UK).
5. (English US) When pressing <kbd>Shift</kbd> + <kbd>Backslash</kbd>, you get 16 (for Shift) and 220 (for backslash).
6. (English UK) When pressing <kbd>Shift</kbd> + <kbd>Backslash</kbd>, you get 16 (for Shift) and 192 (for `, because there is <kbd>~</kbd> on UK).

evilpie: I just tested with a German keyboard layout, and I can reproduce that `:` and `;`, which are totally different keys on the German keyboard, produce 186. :exploding_head:

Dinnerbone (approved):

PR #16905, opened 2024-06-27
render: Remove unused PartialEq impls from `shape_utils`

Aaron1011:
These impls were unused, and I don't think we should ever be doing a 'deep comparison' of things like `DistilledShape`


torokati44:
Well, at least the one on `DrawPath` seems to be needed for/by tests.

Aaron1011:
Oops - I ran `cargo test -p tests` locally, but forgot about unit tests.

evilpie:
You could only derive `PartialEq` for `cfg(test)`.

torokati44:
Superseded by https://github.com/ruffle-rs/ruffle/pull/22529.

Issue #14668, opened 2024-01-08
RPG Shooter Starwish: text box display bug


ku7698234:
### Describe the bug

Game location here: [https://www.gamesflow.com/jeux.php?id=2185451](https://www.gamesflow.com/jeux.php?id=2185451)
When the game is started by clicking "Game Start" then "New Game" then "Confirm Normal", after the text finishes scrolling, the dialog box text cannot be displayed.

### Expected behavior

All dialog box text should appear normally

### Content Location

https://www.gamesflow.com/jeux.php?id=2185451
or here:
[3935.zip](https://github.com/ruffle-rs/ruffle/files/13863937/3935.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

Mozilla Firefox 121.0

kjarosh:
Should be fixed now thanks to https://github.com/ruffle-rs/ruffle/pull/14722, after https://github.com/ruffle-rs/ruffle/pull/16906 it should work even better!

PR #16906, opened 2024-06-27
avm2: Implement `TextField.getLineIndexOfChar()`

kjarosh:
Implement `TextField.getLineIndexOfChar()`.

Related to https://github.com/ruffle-rs/ruffle/issues/14668.


core/src/avm2/globals/flash/text/text_field.rs
```diff
@@ -1444,3 +1444,28 @@ pub fn get_selected_text<'gc>(
...
+        return Ok(Value::Undefined);
+    };
+
+    let index = args.get_i32(activation, 0)?;
+    if index < 0 {
+        // Docs say "throw RangeError", reality says "return -1".
+        return Ok(Value::Number(-1f64));
+    }
+
+    return if let Some(line) = this.line_index_of_char(index as usize) {
```
evilpie, line 1466: Unnecessary return

evilpie (approved):

core/src/avm2/globals/flash/text/text_field.rs
torokati44: How come clippy didn't catch this then? :O

kjarosh: Removed some more unnecessary returns

PR #16914, opened 2024-06-28
extension: Only show onboarding page on first install

danielhjacobs:
See https://developer.chrome.com/docs/extensions/reference/api/runtime#type-OnInstalledReason


Dinnerbone (approved):
Thanks for looking into that so fast!

PR #16918, opened 2024-06-28
web/extension: Check last version on AMO, skip submission if not yet approved

torokati44:
See for relevant API docs:
https://mozilla.github.io/addons-server/topics/api/addons.html#versions-list

I did try this on one of my own extensions, it will probably do what it seems it should do.

Note to @evilpie:

> **file.status** _(int)_ - The [status](https://mozilla.github.io/addons-server/topics/api/addons.html#version-detail-status) for the file.

This line is wrong on the page above, it's a _(string)_.

### DIFF
--- a/web/packages/extension/tools/submit_xpi.ts
+++ b/web/packages/extension/tools/submit_xpi.ts
@@ -36,6 +36,39 @@ async function submit(
         baseURL: "https://addons.mozilla.org/api/v5/addons/",
     });
 
+    console.log("Checking the status of the last submitted add-on version...");
+    const versionsResponse = await client.get(
+        `addon/${extensionId}/versions/`,
+        {
+            headers: {
+                Authorization: `JWT ${getJwtToken(apiKey, apiSecret)}`,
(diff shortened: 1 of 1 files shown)


torokati44:
Eh, it probably would have been more pedantic to add a new function for checking, and only call `submit` based on the result of that, but eh...

Dinnerbone (approved):

torokati44:
I triggered a submission workflow run manually: https://github.com/ruffle-rs/ruffle/actions/runs/9722741599

I was confused at first as to why it didn't do the check, because I forgot that this workflow doesn't clone the repo, it just extracts the latest release, which didn't have this PR in yet. And the submission failed because this version already exists on AMO - but the workflow run still succeeded. Maybe we should change that?
I get that for Edge (and lately maybe even for Chrome) we should swallow some failures while the last submission is still pending, but perhaps it would be better to only allow this specific error, or even better, do this same kind of check for those stores too, if there's an easy way.

Issue #16574, opened 2024-06-03
Loved: Text is not clickable


n0samu:
### Describe the bug

In [Loved](https://www.newgrounds.com/portal/view/539211/format/flash?emulate=flash), text is not clickable, so the game is not playable.

### Expected behavior

On the first screen, "man" and "woman" should highlight when hovered and respond when clicked.

https://github.com/ruffle-rs/ruffle/assets/71368227/c385d3d0-b9da-4f90-b6b8-0690c5821dfa

### Content Location

Page URL: https://www.newgrounds.com/portal/view/539211/format/flash?emulate=flash
SWF URL: https://uploads.ungrounded.net/539000/539211_loved.swf

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

Firefox 126

### Additional information

This happens due to an unimplemented feature:
`WARN ruffle_core::stub: Encountered stub: AVM2 flash.text.TextField.getCharIndexAtPoint()`

PR #16924, opened 2024-06-29
avm2: Roughly implement `getLineIndexAtPoint` and `getCharIndexAtPoint` of `TextField`

kjarosh:
Fixes https://github.com/ruffle-rs/ruffle/issues/16574.

Both implementations currently use `screen_position_to_index`, which is inaccurate, and that's why these methods still produce a stub warning. However, this implementation makes some games work, is a lot better than a perfect implementation which would require a lot of work and time, and infinitely better than no implementation at all.

We will probably improve upon these implementations when laying out text and `screen_position_to_index` are improved.


Dinnerbone (approved):
<3

Issue #15911, opened 2024-04-07
Provide TypeScript declarations (or provide an example with TypeScript)


damil42:
The NPM package [@ruffle-rs/ruffle](https://www.npmjs.com/package/@ruffle-rs/ruffle) has no `TS` label (declarations / d.ts). Are we able to include the d.ts in this project? Otherwise you can't use it in TypeScript properly.

Add an exmaple using the package `@ruffle-rs/ruffle` with TypeScript. You use `ruffle-core` in [demo](https://github.com/ruffle-rs/ruffle/tree/master/web/packages/demo) (React TypeScript) but replacing it with the npm does not work. I wonder how you was able to work with it (`ruffle-core` package).

travisvn:
I am having this _exact_ issue, and I also tried to reference the demo's usage. It's very validating to see I wasn't the only one unable to reconcile `ruffle-core` with what was available.

Dinnerbone:
As of #16928, `ruffle-core` is now usable as a module. You need to run `installRuffle("local")` somewhere to initialize it though.

We aren't yet publishing the module on npm though - that will come shortly.

Issue #3952, opened 2021-04-10
Loading ruffle builds through webpack


hjri:
**Describe the bug**
I'm trying to use ruffle with webpack to add flash support in a vue-based website.
What i do is this:
```js
      //contents of a "create" function in vue component
      require('ruffle-mirror/ruffle') // using ruffle-mirror to avoid putting wasm/built js into source tree
      window.RufflePlayer = window.RufflePlayer || {};
      window.RufflePlayer.config = {
        publicPath: '/static/ruffle', // using copy-webpack-plugin to copy .wasm files into specified dir
        polyfills: false
      }
      const ruffle = window.RufflePlayer.newest()
      const player = ruffle.createPlayer()
      // element in a component
      const container = this.$refs.container
      container.appendChild(player)
      player.load(this.src) // loading sfw file from property
```
This however doesn't work, as Ruffle seems to completely ignore `publicPath` and still tries to load the wasm from current path. No matter if i try to set publicPath on `ruffle` or `player` the outcome is same - it tries to fetch from `./` and gets 404.

**Expected behavior**
Everything works just like when putting ruffle build in a source tree (into static dir), publicPath setting works

**Is the problem with the Ruffle desktop app, extension, or self-hosted version?**
self-hosted version (through npm and ruffle-mirror https://github.com/rwv/ruffle-mirror)

**What platform are you using?**
Webpack 4.44
Firefox 81 Linux x64

hjri:
Figured it out, i need to setup the config BEFORE loading ruffle.

However it's still strange that `publicPath` has no effect when setting it on player instance and others.

Another thing is that loading ruffle through webpack doesn't seem to be possible right now at all. No matter what I do I can't seem to be able to make webpack just generate url for wasm to make a publicPath out of it. It gets very complicated very quick.

Would be nice to have a separate builds for `ruffle-core` that can work with webpack.

adrian17:
> Another thing is that loading ruffle through webpack doesn't seem to be possible right now at all.

The demo page [1] is generated with webpack and uses `ruffle-core`; I know it doesn't substitute for good documentation, but for now it can serve as reference on usage.

[1] https://github.com/ruffle-rs/ruffle/blob/master/web/packages/demo/www/index.js

hjri:
I tried looking at it, but as far as I understand, `ruffle-core` there is built locally and uses likewise locally-built wasm file, there are no releases for just `ruffle-core` + wasm file.

[...]

PR #16928, opened 2024-06-29
web: Make ruffle-core actually work when not being used as a selfhosted zip

Dinnerbone:
Now things like ruffle website can *actually* depend on ruffle-core as a library, rather than only as an already built zip of assets and js files.

This does **not** include releasing ruffle-core to npm, I'll look at that later after reviewing the API.

API of selfhosted does not change. API of ruffle-core changes... but nobody can use it yet anyway, so it's all fine.


danielhjacobs:
Related issue: https://github.com/ruffle-rs/ruffle/issues/15911

danielhjacobs:
Another related issue: https://github.com/ruffle-rs/ruffle/issues/3952

travisvn (commented):
This is fantastic, much needed work. Thank you very much for this!

Issue #14996, opened 2024-01-29
TextFields mouse control for text select


ReFaller:
### Describe the bug

You can't select text with the mouse in text fields (for Input text fields and for Dynamic text fields), selection works only for the keyboard (Shift + Left/Right key buttons). Also, no useful features exist, such as double-clicking to select a word between space delimiters (available in Flash Player).
![image](https://github.com/ruffle-rs/ruffle/assets/39001461/afeb08de-be3a-4f6e-be0d-40113eef2826)

### Expected behavior

You should be able to highlight text by holding down the mouse anywhere and double clicks for worlds should work too for selecting it.

### Content Location

[test.zip](https://github.com/ruffle-rs/ruffle/files/14086781/test.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

ReFaller:
![image](https://github.com/ruffle-rs/ruffle/assets/39001461/e5319970-b75d-4937-8d22-e1f6f543370b)

waspennator:
I was able to highlight text using the mouse as of build 5/27/24, however it seems that double clicking to highlight text only works on flash still.

PR #16946, opened 2024-06-30
text: Implement double/triple click selection

kjarosh:
Fixes https://github.com/ruffle-rs/ruffle/issues/14996, progresses https://github.com/ruffle-rs/ruffle/issues/1891.

This PR implements selecting words by double-clicking and selecting lines by triple-clicking. It also includes support for selecting words and lines while dragging.

A test is added to verify this behavior. Example use cases covered by this test:
1. selecting characters,
2. selecting words by double-clicking,
3. selecting multiple words by double-clicking and dragging,
4. selecting lines by triple-clicking,
5. selecting multiple lines by triple-clicking and dragging,
6. double-clicking at word boundaries,
7. double-clicking between spaces,
8. changing the underlying text while clicking and dragging,
9. trying to select while the text field is not selectable.


Dinnerbone:
All that *and* tests?! You spoil us!

torokati44:
I want _quadruple_ click selection!! :stuck_out_tongue_winking_eye:

kjarosh:
Unfortunately the test `textbox_click` did not take into account that FP supports double/triple click and always expects to see a caret instead of selecting words/lines. That needs to be fixed

core/src/display_object/edit_text.rs
```diff
@@ -196,6 +200,11 @@ impl<'gc> EditText<'gc> {
     // This seems to be OS-independent
     const INPUT_NEWLINE: char = '\r';
 
+    /// Maximum duration between two clicks for them to be considered as a multi-click.
+    ///
+    /// Derived experimentally to be 250 ms.
+    const MULTI_CLICK_MAX_DURATION: TimeDelta = TimeDelta::milliseconds(250);
```
kjarosh, line 206: That value is pretty low, usually 500 ms is used. I'm wondering whether in the future we should add an option to increase that as an accessibility setting.

Dinnerbone: Seems to be 500ms for me in FP on Windows

Dinnerbone: Good shout - yes it is.

Dinnerbone: (I suspect this is actually not handled by edittext, but a general player event "double click" - which edittext listens to. Triple click is probably custom though.

kjarosh: That seems to be platform-dependent then, because on GNOME it's constant 250 ms irrespective of double click delay setting.

Unfortunately it seems that winit does not support double clicking though. Also, I wouldn't be able to test the behavior of triple clicking as I cannot change the delay 😐

kjarosh: I changed the delay to 500ms and added a TODO. I will address supporting double clicks at OS level in another PR. It should be easy to implement on the web, but on desktop I'm afraid we will need to add a setting until it's implemented in winit.

kjarosh: Never mind, I implemented a rough framework for double clicks, so that the click index may be supplied in player events, and when it's not, it's calculated.

Currently it's not supplied and TODOs are added both in desktop and web (tried `js_event.detail` but it doesn't work OOTB).

Dinnerbone (approved):
Thank you!

PR #16950, opened 2024-07-01
avm2: Implement Graphics.drawTriangles with UVT data

Aaron1011:
We now store a `uvt_data` vec in `DrawPath::Fill`. This gets passed through `lyon` as path attributes, stored in the vertex shader input, and passed along to a new `bitmap_uvt.wgsl` shader.

For now, this is only implemented in the wgpu backend.


kjarosh:
Have you thought about skipping tessellation at all and just rendering the triangles directly? From my understanding treating triangles as a drawing here makes everything more complicated:
* a winding rule is used during tessellation, whereas FP just renders triangles without winding,
* a render-backed culling cannot be used when drawing triangles as paths,
* going through a tessellator is more expensive,
* perhaps UVT could also be simplified without tessellation?

That would also get rid of `apply_temp_triangle_drawing`.

adrian17:
(marked as waiting-for-author, as Kamil's note counts as an important review and might cause the PR to be changed a lot)

Aaron1011:
We would still need `apply_temp_triangle_drawing`, since we want the normal draw commands for hit testing

kjarosh:
I've not tried to implement this and I do not have extensive knowledge related to hit testing, so what I'm saying might not make sense, but I rather thought about adding a new primitive for shapes. So now we have `DrawPath::Stroke` and `DrawPath::Fill`, and we could add a third one: `Triangles`, which skips tessellation, renders directly as triangles, and has a simpler hit test (and perhaps simpler UVT). So drawing triangles would be just adding a new instance of `Triangles` with vertices, like we add `DrawingPath::Fill` and `DrawingPath::Stroke` to `Drawing.paths`.

PR #16996, opened 2024-07-05
extension: Don't register 4399 content script if Ruffle is disabled

danielhjacobs:
Dual-purpose:
1) Don't register the 4399 content script if Ruffle is installed but disabled.
2) Don't show a separate permission for 4399 to avoid confusing people (see https://github.com/ruffle-rs/ruffle/discussions/16990). I tested on Firefox and it doesn't show after this PR.

Also has another benefit, which is that there's no manifest warning on Firefox versions without MAIN world support.

Ideally, the 4399 specific code could be toggled off in the extension settings, but between this and the declarativeNetRequestWithHostAccess change, at least its code doesn't generate additional permissions anymore, so I'll leave that for now. I could change the `contentScriptRegistered` function but since they're both registered at once it seems unnecessary.

Edit: Important note: this has not actually been tested on 4399.com

### DIFF
--- a/web/packages/extension/manifest.json5
+++ b/web/packages/extension/manifest.json5
@@ -26,12 +26,6 @@
             "all_frames": true,
             "run_at": "document_start",
         },
-        {
-            "matches": ["https://www.4399.com/flash/*"],
-            "js": ["dist/siteContentScript4399.js"],
-            "world": "MAIN",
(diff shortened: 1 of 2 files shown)


Dinnerbone (approved):
Tested 4399 and still works! Thank you!

Issue #7841, opened 2022-09-02
https://z0r.de/3537 doesn't loop


Diatonator:
### Describe the bug

https://z0r.de/3537 animation doesn't loop when it should.
Seems like a regression because on z0r's Ruffle version it loops.

Nightly 2022-08-31

### Expected behavior

Looping clip

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Chrome 105.0.5195.54

Diatonator:
Info from the z0r's Ruffle version (where the bug does not reproduce):

# Player Info
Allows script access: true

# Page Info
Page URL: https://z0r.de/3537?ruffle
SWF URL: https://z0r.de/L/z0r-de_3537.swf

# Browser Info
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/105.0.0.0 Safari/537.36
Platform: Win32
Has touch support: false

# Ruffle Info
Version: 0.1.0
Name: nightly 2022-09-05
Channel: nightly
Built: 2022-09-05T20:21:14.431Z
Commit: f152e99cc72de2a30b9ceda7197c3cf8315ad1ac
Is extension: false

# Metadata
width: 352
height: 288
frameRate: 15
numFrames: 9000
swfVersion: 6
backgroundColor: null
isActionScript3: false
uncompressedLength: 200319

torokati44:
FYI the z0r version includes #2705, which is known to "fix" some non-looping loops.

Bale001:
Fixed by #21526

PR #17019, opened 2024-07-06
core: Allow playhead to advance past SWF head frame count

Aaron1011:
The frame count declared in the SWF header is not required to match the actual number of frames present in the tag stream. Flash Player will execute frames until it reaches the end of the tag stream (reporting a `currentFrame` larger both `totalFrames` and `framesLoaded`), and *then* jump back to the start.

We now inspect the tag stream in `determine_next_frame` - if we've reached the end (either `TagCode::End` or we've run out of bytes), we jump back to the beginning. The declared SWF frame count is ignored.

I haven't addressed gotos yet, which are more complicated due to the 'frame clamping' behavior.


torokati44:
This fixes https://github.com/ruffle-rs/ruffle/issues/7841 plus supersedes https://github.com/ruffle-rs/ruffle/pull/2705, by also fixing all three loops linked therein.

core/src/display_object/movie_clip.rs
```diff
@@ -1276,10 +1276,18 @@ impl<'gc> MovieClip<'gc> {
...
-    pub fn frames_loaded(self) -> i32 {
+    // Gets the actual number of frames loaded (ignoring the SWF header)
+    pub fn real_frames_loaded(self) -> i32 {
         self.0.read().frames_loaded()
     }
 
+    // Gets the number of frames loaded, clamping it to the number of frames
+    // declared in the SWF header. This is the value we report to AVM code
+    // to match Flash Player's behavior
+    pub fn frames_loaded_for_avm(self) -> i32 {
```
adrian17, line 1287: can we rename these to be more consistent? As in, since both `frames_loaded_for_avm` and `total_frames` are based on what the header says, I'd say both should have the same naming scheme - either no suffix, or `header_` (for example).
`frames_loaded_for_avm` doesn't seem to fit, as it's used outside of pure AVM code.

core/src/display_object/movie_clip.rs
```diff
@@ -3404,10 +3426,7 @@ impl<'gc> MovieClipData<'gc> {
     }
 
     fn play(&mut self) {
-        // Can only play clips with multiple frames.
```
adrian17, line 3407: as researched on discord: there's some weirdness with `play()/stop()` that can cause `.isPlaying` to not always correspond to whether the playhead is _actually_ moving.
- a movie with 1 frame does `.play(); .isPlaying == false`
- a movie with 2 header frames but 1 real frame does `.play(); .isPlaying == true` (but there's nothing to loop to, so I don't know whether the playhead even attempts to move)
- a movie with 1 header frame but 2 real frames does `.play(); .isPlaying == false` but still loops (`enterFrame` event shows cycling `.currentFrame`)

core/src/display_object/movie_clip.rs
```diff
@@ -1470,12 +1478,25 @@ impl<'gc> MovieClip<'gc> {
...
-        if self.current_frame() < self.total_frames() {
+        let mc = self.0.read();
+        let mut reader = mc.static_data.swf.read_from(mc.tag_stream_pos);
+
+        // We ignore the frame count from the header, and instead continue
+        // until we reach the end of the stream or a `TagCode::End`.
+        // Flash Player ignores the frame count, and just executes the full
+        // tag stream before returning to the first frame.
+        if !reader.as_slice().is_empty()
+            && reader.read_tag_code().expect("Failed to read tag") != TagCode::End as u16
```
adrian17, line 1489: this can panic if the swf happens to have 1 byte hanging at the end, instead of 2. I was thinking it'd be better to fallback to either behavior in this case, but no clue what FP does in this case - does it stop or loop or crash.

Bale001:
Superseded by #21526

Issue #16991, opened 2024-07-04
Dead Maze: Right click opens context menu instead of triggering player action


JediFay:
### Describe the bug

Right click opens context menu
![image](https://github.com/ruffle-rs/ruffle/assets/174175668/7b0117cb-9ef5-4536-a5b6-2e58c716f77e)

### Expected behavior

It should trigger player action
![image](https://github.com/ruffle-rs/ruffle/assets/174175668/ab502c50-dde6-4f9b-957a-15f848516423)

### Content Location

http://deadmaze.com/alpha/chargeur-deadmeat.swf

### Affected platform

Desktop app

### Operating system

Windows 11

torokati44:
Hmm, interesting, I can't reproduce on Linux, neither on Wayland, nor on X11.
Is this perhaps due to some kind of accessibility setting that is turned on?

JediFay:
Only happens on Windows 10/11

torokati44:
I know you don't have Windows eother, @kjarosh, but is it possible that we _do_ get some double-click events after all, it's just misinterpreted/disguised as something else in a çūr$ėd way somewhere before it gets to us? 😱

kjarosh:
Unfortunately that game requires login, but by analyzing the code, the only possibility for the context menu to open is when the right mouse button is released. I'm not sure whether we are talking about LMB or RMB double-click, if LMB then I really do not see a possibility how that could happen, but if it's RMB then I suspect it might have something to do with this TODO:

https://github.com/ruffle-rs/ruffle/blob/23b7810a0761728c9e7c03114cb549cb309b3479/desktop/src/app.rs#L260

torokati44:
> Unfortunately that game requires login,

Nah, I was able to get in game just by leaving the input fields empty... 🤷‍♂️

JediFay:
Yes

kjarosh:
So it seems that in general right click events are not implemented (`rightClick`, `rightMouseUp`, `rightMouseDown`)

Issue #11810, opened 2023-06-30
Can not right click - debug window appears


jentak314:
### Describe the bug

This game is controlled by clicking mouse, both left and right button. If I right click, instead of game detecting it, a debug window appears.

It happens immediately in tutorial.

https://www.superhry.cz/hra/2842-stick-war-2

![stick_wars_ruffle](https://github.com/ruffle-rs/ruffle/assets/138238090/4e999334-dd8a-4c75-ac58-f3b79eb582a3)

### Expected behavior

.

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Chrome

Lord-McSweeney:
Right-click controls are unsupported; see later comments in #11009. Closing in favor of #10288.

Issue #10612, opened 2023-04-08
Memory.exe: Right click won't work


p0008874:
### Describe the bug

https://user-images.githubusercontent.com/75534590/230738850-2cdcfece-86d0-4dc5-92f1-fdc79cbbddf3.mp4

### Expected behavior

https://user-images.githubusercontent.com/75534590/230738854-99541bc6-1eb6-45d2-8afb-3e96546c8a78.mp4

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

https://www.newgrounds.com/portal/view/715381

p0008874:
![image](https://github.com/ruffle-rs/ruffle/assets/75534590/ff112786-423a-4000-8b9f-800b14531237)
Right click still won't remove and Context menu popup.

n0samu:
See #10288

Issue #10389, opened 2023-03-26
House of Wolves: right click does not work


MartySVK:
### Describe the bug

House of Wolves
- when you select a settler, with a right click you can do some action like the move. But it is not working in Ruffle (no errors). It is working in Flash.

SWF - https://data.gameflare.com/games/4772/W0CfhIUObjkuPk.swf

### Expected behavior

select settler, click with the right mouse button and move

### Affected platform

Desktop app

### Operating system

Windows 10

Lord-McSweeney:
Possibly related to this error: `Modifying an XML object is not yet implemented`.

MartySVK:
not sure, this error was there before, but it is not anymore

Lord-McSweeney:
In that case, the issue is probably #10288.

PR #17022, opened 2024-07-07
core: Implement right & middle click events, fix disappearing context menu on desktop

kjarosh:
Fixes https://github.com/ruffle-rs/ruffle/issues/16991, fixes https://github.com/ruffle-rs/ruffle/issues/10288, fixes https://github.com/ruffle-rs/ruffle/issues/10389, progresses https://github.com/ruffle-rs/ruffle/issues/11009, fixes https://github.com/ruffle-rs/ruffle/issues/10612, fixes https://github.com/ruffle-rs/ruffle/issues/11810.

This PR implements the following events:
* `rightMouseDown`,
* `middleMouseDown`,
* `rightMouseUp`,
* `middleMouseUp`,
* `rightClick`,
* `middleClick`.

Additionally this PR fixes disappearing context menu on desktop by changing its trigger from key release to key press (as in FP). Opening context menu on press is important, because events consume right click on press, not on release.


Dinnerbone:
I'm a little worried about web. We have no tests for context menu on web (as much as I constantly wish otherwise... one day!), and it's pretty convoluted because of ios (and mobile in general, but mostly ios iirc)

Testing this on web on a desktop and it works as expected - but testing it on firefox on android, the right click action doesn't seem to trigger and the context menu shows up instead :( Specifically testing #10389 and #10389

If you want to say mobile is for another time and this gets desktop working - then I'm okay with that, as long as we don't introduce any regressions.

kjarosh:
I'm aware that right clicking does not currently work on mobile, matter of fact not introducing regressions was my reason for not implementing it in this PR as it already includes a lot of changes.

I tried implementing it for mobile, but currently we handle the context menu a little differently for mobile and desktop, and unifying the behavior requires some more work. I've just added a TODO comment about that.

danielhjacobs:
The main issue is this one: https://bugs.webkit.org/show_bug.cgi?id=213953

Dinnerbone:
> I'm aware that right clicking does not currently work on mobile, matter of fact not introducing regressions was my reason for not implementing it in this PR as it already includes a lot of changes.

That's fine, thank you! Just wanted to know if that was intentional or not. I'm 100% okay with it not being here in this PR, as long as it's understood that that isn't the case.

yangyangdaji:
Maybe this can be fixed
 https://github.com/ruffle-rs/ruffle/issues/10578

yangyangdaji:
Maybe be fixed https://github.com/ruffle-rs/ruffle/issues/10578

Dinnerbone (approved):

PR #17025, opened 2024-07-07
chore: Make dependabot increase version requirements for npm dependencies

Dinnerbone:
It's completely ignoring things like new major versions, because it thinks we're a library and the default strategy for libraries is to stay semver compatible

However; the only library that we will actually publish is `ruffle-core` - which has no publicly accessible dependencies, so it's fine

Documentation: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,6 +2,7 @@ version: 2
 updates:
   - package-ecosystem: "npm"
     directory: "/web/"
+    versioning-strategy: increase
     schedule:
       interval: "weekly"
     groups:


Dinnerbone:
(also @torokati44 you might enjoy finding out about "increase-if-necessary")


(and then get sad when you read https://github.com/dependabot/dependabot-core/issues/4009)

torokati44:
> (also @torokati44 you might enjoy finding out about "increase-if-necessary", which can reduce accidental dependency dupes)

:star_struck:

> (and then get sad when you read https://github.com/dependabot/dependabot-core/issues/4009)

:sob:

torokati44 (approved):

PR #17028, opened 2024-07-07
web: Update ws manually until new webdriverio major update

Dinnerbone:
Removes our security alert (not that it actually affected us)

(https://github.com/webdriverio/webdriverio/issues/13056)

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -11097,17 +11097,16 @@
             }
         },
         "node_modules/puppeteer-core/node_modules/ws": {
-            "version": "8.13.0",
-            "resolved": "https://registry.npmjs.org/ws/-/ws-8.13.0.tgz",
-            "integrity": "sha512-x9vcZYTrFPC7aSIbj7sRCYo7L/Xb8Iy+pW0ng0wt2vCJv7M9HOMy0UoN3rr+IFC7hb7vXoqS+P9ktyLLLhO+LA==",
+            "version": "7.5.10",
+            "resolved": "https://registry.npmjs.org/ws/-/ws-7.5.10.tgz",
+            "integrity": "sha512-+dbF1tHwZpXcbOJdVOkzLDxZP1ailvSxM6ZweXTegylPny803bFhA+vqBYw4s31NSAk4S2Qz+AKXK9a4wkdjcQ==",
             "dev": true,
-            "license": "MIT",
             "engines": {
-                "node": ">=10.0.0"
+                "node": ">=8.3.0"
             },
             "peerDependencies": {
                 "bufferutil": "^4.0.1",
-                "utf-8-validate": ">=5.0.2"
+                "utf-8-validate": "^5.0.2"
             },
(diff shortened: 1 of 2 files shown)


web/package-lock.json
```diff
@@ -11023,17 +11023,16 @@
             }
         },
         "node_modules/puppeteer-core/node_modules/ws": {
-            "version": "8.13.0",
-            "resolved": "https://registry.npmjs.org/ws/-/ws-8.13.0.tgz",
-            "integrity": "sha512-x9vcZYTrFPC7aSIbj7sRCYo7L/Xb8Iy+pW0ng0wt2vCJv7M9HOMy0UoN3rr+IFC7hb7vXoqS+P9ktyLLLhO+LA==",
+            "version": "7.5.10",
```
torokati44, line 11026: This is... a downgrade here? Is that a problem?

torokati44:
Also are we not just gonna forget to undo this in a couple months when it's gonna be due?
I'm not opposing to it BTW, just asking.

web/package-lock.json
Dinnerbone: Not a problem

Dinnerbone:
I think it's fine, even if we forget 😅

web/package.json
```diff
@@ -48,5 +48,8 @@
         "lint": "npm run checkTypes --workspaces --if-present && eslint . && stylelint **.css",
         "format": "eslint . --fix && stylelint --fix **.css",
         "version-seal": "cross-env ENABLE_VERSION_SEAL=true tsx packages/core/tools/set_version.ts"
+    },
+    "overrides": {
+        "ws": "^7.0.0"
```
torokati44, line 53: A `TODO` comment to "remove this when ..." would be appreciated! :pleading_face:
Otherwise, I'm fine with this.

Dinnerbone: I agree it would be!


But alas, json

torokati44: Oh, frick, heck, right.
*looks up how much json5everywhere.org would cost...*

torokati44:
> I think it's fine, even if we forget 😅

Could you please elaborate? 😶
Won't this prevent updating to a 8.x.x version of ws?
Or it will be updated in the proper place, and this override will break other stuff no longer compatible with 7.x.x, and we'll be reminded that way? 🤔

Dinnerbone:
> Could you please elaborate? 😶

Well, mostly it just doesn't matter what version of ws we're using, this is such an inconsequential part of things as far as I can see. If we miss a new version... oh well? :D

danielhjacobs:
A new webdriverio major version was released. The upgrade is in https://github.com/ruffle-rs/ruffle/pull/17533, though we also need a new release of @wdio/browserstack-service with https://github.com/webdriverio/webdriverio/pull/13374 to fix the failures there.

danielhjacobs:
New PR: https://github.com/ruffle-rs/ruffle/pull/17539

PR #17030, opened 2024-07-07
swf: Remove 'libflate' feature

Aaron1011:
The 'flate2' crate already supports both C and pure Rust backends, so we don't gain anything by supporting libflate

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,12 +34,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
-name = "adler32"
-version = "1.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


torokati44:
I guess we could then also remove the `libflate` and `libflate_lz77` rows from `LICENSE.md` in both the project root and under `web` - not that they're up-to-date at all, but still.

torokati44:
Am I seeing correctly that this feature is no longer enabled at all anywhere by Ruffle, so only potentially by external users of `swf`?

torokati44:
Assuming the answer to the question above is "yes", and maybe that the two `LICENSE.md` files are also updated, this has my preliminary approval.

Aaron1011:
Yes, this only affects external users of the `swf` crate, not any Ruffle crates

torokati44:
There are two `LICENSE.md` files, one in web, the other in the project root.

torokati44 (approved):
This is a breaking change for the `swf` crate of course, but it's probably okay.

torokati44:
Merging due to lack of opposition.

PR #17036, opened 2024-07-07
web: Update eslint to v9

Dinnerbone:
Minimal functional changes, just updating to the new config structure + one commit that enables typescript 'strict' rules

In a future PR I'd like to try and consolidate/detangle these rules a bit, but this is how they were...


torokati44:
(I'm not really feeling like combing through all the changes, so don't take this as a review or an approval, but I'm inclined to believe this should be alright.)

web/package-lock.json
```diff
@@ -23,11 +23,11 @@
                 "chai-html": "^3.0.0",
                 "copy-webpack-plugin": "^12.0.2",
                 "cross-env": "^7.0.3",
-                "eslint": "^8.57.0",
+                "eslint": "^9.6.0",
                 "eslint-config-prettier": "^9.1.0",
                 "eslint-plugin-jsdoc": "^48.5.2",
                 "eslint-plugin-prettier": "^5.1.3",
-                "eslint-plugin-react-hooks": "^4.6.2",
+                "eslint-plugin-react-hooks": "^5.1.0-rc",
```
torokati44, line 30: Do we really want this? Is this mayhaps why Dependabot wasn't updating yet?

web/package-lock.json
```diff
@@ -38,7 +38,7 @@
                 "ts-loader": "^9.5.1",
                 "tsx": "^4.16.2",
                 "typescript": "^5.5.3",
-                "typescript-eslint": "^7.15.0",
+                "typescript-eslint": "^8.0.0-alpha.39",
```
torokati44, line 41: Same (especially) here.

web/package-lock.json
Dinnerbone: Possibly... I didn't know it checked peer dependencies though!

Unfortunately it's the first release with eslint9 support :(

web/package-lock.json
```diff
@@ -1181,46 +1220,6 @@
                 "npm": ">=7.0.0"
             }
         },
-        "node_modules/@humanwhocodes/config-array": {
-            "version": "0.11.14",
-            "resolved": "https://registry.npmjs.org/@humanwhocodes/config-array/-/config-array-0.11.14.tgz",
-            "integrity": "sha512-3T8LkOmg45BV5FICb15QQMsyUSWrQ8AygVfC7ZG32zOalnqrilm018ZVCw0eapXux8FtA33q8PSRSstjee3jSg==",
-            "deprecated": "Use @eslint/config-array instead",
```
torokati44, line 1170: Oh, nice to actually get these addressed.

web/package-lock.json
```diff
@@ -1235,13 +1234,18 @@
                 "url": "https://github.com/sponsors/nzakas"
             }
         },
-        "node_modules/@humanwhocodes/object-schema": {
-            "version": "2.0.3",
-            "resolved": "https://registry.npmjs.org/@humanwhocodes/object-schema/-/object-schema-2.0.3.tgz",
-            "integrity": "sha512-93zYdMES/c1D69yZiKDBj0V24vqNzB/koF26KPaagAfd3P/4gUlh3Dys5ogAK+Exi9QyzlD8x/08Zt7wIKcDcA==",
-            "deprecated": "Use @eslint/object-schema instead",
```
torokati44, line 1224: (Same here.)

web/package-lock.json
Dinnerbone: Same here; first version to support eslint9

They both seem pretty good though, and typescript-eslint got some upgraded rules from it

web/packages/selfhosted/js/ruffle.js
```diff
@@ -22,7 +22,8 @@ try {
 
         currentScriptURL = new URL(".", src);
     }
-} catch (e) {
+    // eslint-disable-next-line no-unused-vars
+} catch (_e) {
```
torokati44, line 26: Huh, is this lint disable comment needed even with the underscore prefix?

web/eslint.config.js
```diff
@@ -51,7 +51,15 @@ export default tseslint.config(
...
                 "error",
-                { argsIgnorePattern: "^_" },
+                {
+                    args: "all",
+                    argsIgnorePattern: "^_",
+                    caughtErrors: "all",
+                    caughtErrorsIgnorePattern: "^_",
+                    destructuredArrayIgnorePattern: "^_",
+                    varsIgnorePattern: "^_",
+                    ignoreRestSiblings: true,
```
torokati44, line 61: Is this something we actually want, or just something that would be too tedious to fix now?

web/packages/selfhosted/js/ruffle.js
Dinnerbone: That only works with typescript :(

torokati44: Oh, shoot, I see. Alright.

web/eslint.config.js
torokati44: Bah, I was specifically referring to `ignoreRestSiblings`. Why didn't it pick only that line... The rest (haha, puntended) I can understand, sure.

Dinnerbone: Oh, good question! I think so yes.

The syntax is this:
```typescript
let {x, y, ...rest} = {x: 1, y: 2, z: 3, w: 4};
console.log(x); // 1
console.log(y); // 2
console.log(rest); // Object { z: 3, w: 4 }
```

If you want to extract part of an object using the rest syntax, that's how you (can) do it. You'd need to ignore the the ones you specify, and then you can keep the ones you don't. Without this set to true, you need to do something like `let {x: _x, y: _y, ...rest} = ....`

I don't think we actually use this today though.

torokati44: Oh, I see. It's probably fine to set it then.

web/eslint.config.js
```diff
@@ -0,0 +1,150 @@
...
+            globals: globals.mocha,
+        },
+    },
+    {
+        files: ["packages/selfhosted/test/**"],
+        languageOptions: {
+            globals: {
+                ...globals.mocha,
+                ...globals.node,
+                ...globals.mocha,
```
torokati44, line 112: You probably meant `browser`, `node`, `mocha` here, instead of `mocha` twice.

web/packages/extension/tools/.eslintrc.yaml
```diff
@@ -1,4 +0,0 @@
-env:
-  node: true
-parserOptions:
-  ecmaVersion: 2022 # Needed for top-level-await.
```
torokati44, line 4: Umm, is this present in the flat config? I'm only seeing `packages/core/tools/**` there.

torokati44:
I _did_ end up combing through it all, and apart from the two remaining notes, everything looks okay.

web/eslint.config.js
Dinnerbone: Interesting that it worked, huh!

web/packages/extension/tools/.eslintrc.yaml
Dinnerbone: Good catch. It's defaulting to "`latest`" which is newer.

We *should* switch to `"latest"` but didn't want to make more changes than needed from base config.

Dinnerbone: Fixed! It should have been extension, not core

web/eslint.config.js
Dinnerbone: Turns out that we don't need it because typescript. This affects js. Removed!

torokati44 (approved):

Issue #17034, opened 2024-07-07
jinsei owata no daibouken 2 - yellow line in map 11


zjk261:
### Describe the bug

https://github.com/ruffle-rs/ruffle/assets/98996990/50529fe5-91b0-4909-9437-0d7f942da02f

### Expected behavior

there should be no line

### Content Location

[owata2.zip](https://github.com/user-attachments/files/16119930/owata2.zip)
[owata2_save.zip](https://github.com/user-attachments/files/16119931/owata2_save.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

I didn't press tab key

kjarosh:
The yellow line is a focus rect. The element is properly focused and has proper focus rect settings. The issue here is about timing – at the time the element is being focused, the focus rect should not be visible, but is. When the element is re-focused, FP also shows the line.

PR #17037, opened 2024-07-07
core: Hide focus highlight on invalid and empty bounds

kjarosh:
Fixes https://github.com/ruffle-rs/ruffle/issues/17034.

Flash Player has a fairly complicated, implementation-dependent behavior related to degenerate bounds and bounds that change after being highlighted. It seems that the highlight is hidden when there's nothing to be rendered:
  1. For a clip with untouched graphics, the highlight is hidden.
  2. For a clip with moveTo(0,0) lineTo(0,0), the highlight is not hidden.

Sometimes highlight is not updated after highlight bounds change, but this behavior is rare and inconsistent, it even differs depending on when the update happens.

However, the most common case of depending on the behavior above is when the SWF creates a movie clip, focuses it, and then adds some content to it. In that case FP will not render the highlight (even if in theory it should). Hiding focus highlight on invalid and empty bounds covers this case, but:
  1. In some rare cases, it will hide the highlight when it shouldn't (e.g. non-empty graphics degenerated into a point).
  2. It does not take into account that sometimes FP does not update highlight bounds.

However, I've never seen these cases in real SWFs, only during testing.


Dinnerbone (approved):
The majority of the (non-test) changes is a comment... classic Flash :D

PR #17042, opened 2024-07-08
desktop: Add View menu

kjarosh:
This PR adds the "View" menu, which is responsible for controlling the view:
1. options "100% (Unscaled)" and "Zoom to Fit" are analogous to FP's "100%" and "Show All",
2. options "Stretch to Fit" and "Crop to Fit" allow setting other scale modes which are not in FP,
3. option "Letterbox" controls letterbox rendering when "Zoom to Fit" is selected,
4. option "Full Screen" enter the full screen (as in FP),
5. option "Quality" selects the quality (as in FP, context menu does not have to be always shown).

image
image

### DIFF
--- a/desktop/src/gui/menu_bar.rs
+++ b/desktop/src/gui/menu_bar.rs
@@ -4,8 +4,10 @@ use crate::gui::{text, DebugMessage};
 use crate::player::LaunchOptions;
 use crate::preferences::GlobalPreferences;
 use egui::{menu, Button, Key, KeyboardShortcut, Modifiers, Widget};
-use ruffle_core::Player;
+use ruffle_core::config::Letterbox;
+use ruffle_core::{Player, StageScaleMode};
 use ruffle_frontend_utils::recents::Recent;
+use ruffle_render::quality::StageQuality;
 use unic_langid::LanguageIdentifier;
 use url::Url;
 use winit::event_loop::EventLoopProxy;
@@ -68,6 +70,7 @@ impl MenuBar {
 
             menu::bar(ui, |ui| {
                 self.file_menu(locale, ui, dialogs, player.is_some());
+                self.view_menu(locale, ui, &mut player);
 
                 menu::menu_button(ui, text(locale, "controls-menu"), |ui| {
                     ui.add_enabled_ui(player.is_some(), |ui| {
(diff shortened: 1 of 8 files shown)


desktop/src/gui/menu_bar.rs
```diff
@@ -275,6 +278,97 @@ impl MenuBar {
...
+                ui.menu_button(text(locale, "quality"), |ui| {
+                    let items = vec![
+                        ("quality-low", StageQuality::Low),
+                        ("quality-medium", StageQuality::Medium),
+                        ("quality-high", StageQuality::High),
+                        ("quality-best", StageQuality::Best),
+                        ("quality-high8x8", StageQuality::High8x8),
+                        ("quality-high8x8linear", StageQuality::High8x8Linear),
+                        ("quality-high16x16", StageQuality::High16x16),
+                        ("quality-high16x16linear", StageQuality::High16x16Linear),
```
kjarosh, line 357: I'm reusing here strings from `settings.ftl`, not sure whether it's a good idea

Dinnerbone:
I think I'd prefer for the 4 scale modes to be together, either in a submenu or just all together in the view menu with a separator below them.

I get that it's more Flash-like... but it's not intuitive 😅 We don't need to follow them in UI practices.

For wording, maybe "Stretch to Fit" and "Crop to Fit" would be more appropriate too? "100%" and "Show All" are confusing and honestly for all these years, until reading the code for this PR I never knew what they actually meant. Now I know! :D


Maybe:
- 100% (Unscaled)
- Zoom to Fit
- Stretch to Fit
- Crop to Fit


Or I guess just the same text that we show in Open Advanced today?

kjarosh:
Some variants to visualize:

| Submenu | No submenu | Submenu, integrated letterbox |
| ------- | ---------- | ----------------------------- |
|  |  |  |

Dinnerbone:
Thanks for that!

I think I'd prefer to leave letterbox as a separate thing - maybe grey it out if it doesn't make sense for the current mode. Having just "on/off" for letterbox is maybe a bit confusing and inconsistent if in Open Advanced we have 4 options instead of just 2 - but I don't think I care too much either way here.

Submenu looks ideal to me, but I'd call it "Scale Mode", not just "Mode".

Croworbit:
will this allow for these options to be changed in real time like FP?

kjarosh:
@Dinnerbone 

image

I think we have 3 options for a letterbox when opening: On, Off, Fullscreen Only. However as long as the third option makes sense when opening a SWF, it does not make too much sense when changing the value at runtime, because if I want to have letterbox enabled at fullscreen, I can just change the option and then enter fullscreen.

Regarding the scale mode options when opening a SWF, I think they are confusing — I had to test what they do, exactly as I had to test what FP's _Show All_ does. IMO your suggestions are a lot better (maybe we should even change the opening options?).

kjarosh:
> will this allow for these options to be changed in real time like FP?

Yes, that's the point of these options

PR #17051, opened 2024-07-08
core,web: Reset focus on player focus loss

kjarosh:
FP resets focus when the window (or the element on web) loses its focus.

The major change in this PR is how Ruffle manages focus on web.

Before this PR, `instance.has_focus` was unrelated to the current focus managed by the browser. When the user clicked anywhere on the window, it was set to `false`, and when the user clicked on the player, it was set to `true`. This had two major issues.

1. When the user clicked on the player, `has_focus` was set to `false`, and then again to `true`. That was problematic when listening to focus change events.
2. Not using browser's focus makes it harder to integrate with Ruffle, i.e. tab in/out of Ruffle.

This patch uses browser's focus management to detect focus and listen to focus events:

1. on `focusin` – FocusGained event is fired,
2. on `focusout` – FocusLost event is fired, and
3. on `pointerdown` – focus is manually set to the canvas.

The canvas has `tabindex` set to -1 in order to be focusable.


web/src/builder.rs
```diff
@@ -574,6 +574,10 @@ impl RuffleInstanceBuilder {
 
         let (renderer, canvas) = self.create_renderer().await?;
 
+        canvas
```
evilpie, line 577: I think you can `set_tab_index` for this.

evilpie (approved):

web/src/builder.rs
kjarosh: You're right! Fixed

web/Cargo.toml
```diff
@@ -74,7 +74,7 @@ features = [
     "EventTarget", "GainNode", "Headers", "HtmlCanvasElement", "HtmlDocument", "HtmlElement", "HtmlFormElement",
     "HtmlInputElement", "HtmlTextAreaElement", "KeyboardEvent", "Location", "PointerEvent",
     "Request", "RequestInit", "Response", "Storage", "WheelEvent", "Window", "ReadableStream", "RequestCredentials",
-    "Url", "Clipboard",
+    "Url", "Clipboard", "FocusEvent"
```
danielhjacobs, line 77: Not directly related to this PR but there seems to have once been a clear attempt to alphabetize this but once ReadableStream was added features were just added in the order they became needed and that has since continued.

kjarosh: The biggest problem of keeping the list in a lexicographical order with multiple entries per line is that most of the time adding/removing a single feature will cause multiple lines to be edited and rearranged

danielhjacobs:
Interesting timing for the new issue, might this fix #17064?

GonzalezAtWork:
This also applies to mx components?
I have an issue of mx:TextInput not working (only ctrl+v works) and inline focusOut not firing when changing focus.

`<mx:TextInput id="text" width="100%" focusOut="test()"/>`

Dinnerbone (approved):

PR #17066, opened 2024-07-09
core: Fix double caret rendering in justified text

kjarosh:
Progresses https://github.com/ruffle-rs/ruffle/issues/1891.

| Before | After |
| ------ | ----- |
|  |  |
|  |  |

### DIFF
--- a/core/src/display_object/edit_text.rs
+++ b/core/src/display_object/edit_text.rs
@@ -902,10 +902,11 @@ impl<'gc> EditText<'gc> {
 
         let caret = if let LayoutContent::Text { start, end, .. } = &lbox.content() {
             if let Some(visible_selection) = visible_selection {
+                let text_len = edit_text.text_spans.text().len();
                 if visible_selection.is_caret()
                     && !edit_text.flags.contains(EditTextFlag::READ_ONLY)
                     && visible_selection.start() >= *start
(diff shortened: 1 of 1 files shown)


adrian17 (approved):

torokati44:
My finger is hovering above the rebase button, but the fact that I don't understand the change (or event the code it changes) completely, keeps me from pressing it.
Would you mind explaining in a sentence or two how the logic produced two (and not more) carets to be rendered, and how the change results in at mos one being rendered, @kjarosh? :thinking:

kjarosh:
Added a description to the commit message which describes the fix.

The condition

```rust
visible_selection.start() >= *start && visible_selection.end() <= *end
```

was inaccurate, because the end of the selection is exclusive. That caused the condition to be true for two adjacent boxes. For instance:

```
  box 1: from 0 to 6,  "hello "
  box 2: from 6 to 11, "world"
```

The caret was rendered for both boxes when it was at position 6.

When applying a correct condition (i.e. treating the end as exclusive) there is a problem with rendering the caret at the very end of the text, because the condition will not be triggered for any box (position 11 in the example above).

That is why a condition specific to this case is added, i.e.

```rust
*end == text_len
```

When the box is the last box in the text, we are forcing the caret to be rendered.

PR #17078, opened 2024-07-11
extension: Use source-map devtool for debug build (close #15064)

### LINKED ISSUE #15064: Extension debug build eval error when loaded unpacked
The extension compiled with `npm run build:debug`, and loaded as unpacked extension in developer mode fails to do anything in the popup or the settings giving:

```
Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self'".

Context

extension://lecnpjmonjilblobmocglkbflfmggpdc/popup.html
```

The error is visible both in the extensions error page or in the inspector while the ruffle settings are open.

Browser:
Edge/Chrome 121

riku-42:
Fixes #15064 

The default value for Webpack `devtool` under development mode is `eval`, however, CSP `script-src: unsafe-eval` is banned from Chrome manifest v3. The PR fixes the issue by changing `devtool` to `source-map`.

### DIFF
--- a/web/packages/extension/webpack.config.js
+++ b/web/packages/extension/webpack.config.js
@@ -125,6 +125,7 @@ export default function (/** @type {Record<string, any>} */ env, _argv) {
         optimization: {
             minimize: false,
         },
+        devtool: mode === "development" ? "source-map" : false,
         plugins: [
             new CopyPlugin({
                 patterns: [


torokati44:
I'm not qualified to review, sorry - but I've tweaked the commit message a little bit while rebasing, I hope you don't mind! ^^

Dinnerbone (approved):
Thank you!

PR #17082, opened 2024-07-11
extension: Don't auto-reload tab when permissions are granted

danielhjacobs:


web/packages/extension/src/background.ts
```diff
@@ -59,13 +59,16 @@ async function disable() {
...
 
-function onAdded(permissions: chrome.permissions.Permissions) {
+async function onAdded(permissions: chrome.permissions.Permissions) {
     if (
         permissions.origins &&
         permissions.origins.length >= 1 &&
         permissions.origins[0] !== "<all_urls>"
     ) {
-        utils.tabs.reload();
+        const { permissionsSwitch } = await utils.getOptions();
```
evilpie, line 68: I don't really understand what `permissionsSwitch` is used for? It's seems to do nothing.

danielhjacobs: Perhaps a comment should be added. It does do nothing in this script, but it makes this value true if it's value changed since the tab loaded even if nothing else changed: https://github.com/ruffle-rs/ruffle/blob/master/web/packages/extension/src/popup.ts#L145

danielhjacobs: Clarifying comment added.

danielhjacobs: Coming up with a name for this variable was tough as I don't actually care about its truthiness, just whether it has changed. I actually asked ChatGPT: "How do I name a boolean variable that I only care if it changed and therefore just toggle between true and false anytime something happens but whose true/false value is irrelevant?"

It suggested:
- hasChanged (to me this would imply true means a change occurred and false means it did not)
- toggleFlag (too generic)
- stateChanged (same problem as hasChanged)
- flipFlop (unclear)
- changeNotifier (I would expect this to be a function that triggers on change)
updateToggle (I considered permissionsToggle, it was my second choice)
statusSwitch (my first choice)

evilpie: `forceChange` maybe?

danielhjacobs: After discussion with @kjarosh and @n0samu on Discord, they suggested using this as a boolean instead of a toggle and giving it the name `showReloadButton`. I set it to true when permissions are granted and set it to false when the tab loads.

evilpie (approved):

PR #17083, opened 2024-07-11
extension: Request permission for SWF origins loaded in player as needed

danielhjacobs:


web/packages/extension/assets/player.html
```diff
@@ -76,6 +76,10 @@
                 </div>
             </div>
         </div>
+        <dialog id="modal">
+            <button id="grant" type="button">Grant permissions</button>
+            <button id="close" type="button">Close</button>
```
danielhjacobs, line 81: I forgot i18n for this, I'll get to that later.

danielhjacobs: Actually, nevermind, just remembered we should save that for https://github.com/ruffle-rs/ruffle/pull/14628 once we determine what to do about the demo page too.

Dinnerbone (approved):

Issue #17067, opened 2024-07-10
Object World Beta: Cannot start game


n0samu:
### Describe the bug

[Object World Beta](https://ooooooooo.ooo/static/?c711ca1c-4baa-4869-97e7-d490a17ab859) shows nothing but a blank screen after the loading bar completes. The error is:
```
ERROR ruffle_core::timer: Unhandled AVM2 error in timer callback: ReferenceError: Error #1065: Variable flash.events::NetFilterEvent is not defined.
        at Utils::ibSSGlobals$/init()
        at ib2/placeStuff()
        at MethodInfo-31()
```

### Expected behavior

The game should load.

### Content Location

Link to game in Flashpoint's database: https://flashpointproject.github.io/flashpoint-database/search/#c711ca1c-4baa-4869-97e7-d490a17ab859
SWF file: [object_world_beta_by_shock_value_d2ky7l8.zip](https://github.com/user-attachments/files/16155291/object_world_beta_by_shock_value_d2ky7l8.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

Aaron1011:
It looks like Flash player defines an undocumented `NetFilterEvent` class that we need to implement

evilpie:
It seems like for some reason this SWF defines a wrapper for every? single class that flash supports. I highly doubt they are actually being used.

torokati44:
Finally a valid excuse to flash[sic!] out all the "useless" AS3 APIs! :partying_face:

Lord-McSweeney:
The function `Playground.Entity::ibEntity$/addIncompatibilities` does the following lookup of `Vector`:
```
[...]

PR #17089, opened 2024-07-12
avm2: Implement some more classes

Lord-McSweeney:
Progresses #17067


torokati44:
About `Trace`, I found this a while ago: https://sangupta.com/tech/magic-of-flash-trace-trace.html

torokati44 (approved):
Good enough for initial stubs I guess, especially since they allow real world content to progress - the added classes can be further fleshed out later.

PR #17109, opened 2024-07-13
avm2: Use small string cache when stringifying ints < 10

adrian17:
I checked that it speeds up a `vector[0] = 0; ...up to 10` microbenchmark by ~2x; will also reduce GC work, significantly if a movie relies on these kinds of accesses a lot.
This will become irrelevant for this benchmark when we optimize `setproperty` and `nextvalue` to not create strings at all, but it's a useful optimization anyway.

Created as draft as I still wanna play with it later today.


adrian17:
Ready for review, added one small commit.

Lord-McSweeney (approved):

PR #17119, opened 2024-07-14
avm2: Better trait verification

Lord-McSweeney:


Lord-McSweeney:
Undrafted as I still can't figure out where exactly avmplus does the eager return type lookup :/

tests/tests/swfs/avm2/sub_super_same_field/test.toml
```diff
@@ -1,3 +1 @@
 num_frames = 1
-# Ruffle does not currently reproduce this behavior
-known_failure = true
```
adrian17, line 3: just to be sure, I'm assuming that this is (the first) test that depends on the optimizer?

core/src/avm2/class.rs
```diff
@@ -678,34 +680,74 @@ impl<'gc> Class<'gc> {
...
 
-                            if !instance_trait.is_override() {
-                                return Err(format!("VerifyError: Trait {} in class {} has same name as trait {} in class {}, but does not override it", instance_trait.name().local_name(), self.name().local_name(), supertrait.name().local_name(), superclass_def.name().local_name()).into());
+                                (TraitKind::Const { .. }, TraitKind::Const { .. })
+                                | (TraitKind::Const { .. }, TraitKind::Slot { .. })
+                                | (TraitKind::Slot { .. }, TraitKind::Const { .. })
+                                | (TraitKind::Slot { .. }, TraitKind::Slot { .. }) => {
+                                    did_override = true;
+
+                                    // Const/Var traits override each other in avmplus
```
adrian17, line 705: I don't think this is entirely true, as in: FP is even more lax. It lets `var/const/class` (though ofc the last one won't happen in the wild) "override" any trait. Here's my old demo:
```as
class B {
    public var field1 = "base.field1";
    public function get field2(){ return "base.field2"; }
    public function field3(){ return "base.field3"; }
    public function printBase(){
        trace(this.field1); // base.field1
        trace(this.field2); // base.field2
        trace(this.field3); // child.field3
        trace(this["field1"]); // child.field1
        trace(this["field2"]); // child.field2
        trace(this["field3"]); // child.field3
    }
}
class C extends B {
    public var field11 = "child.field1";
    public var field21 = "child.field2";
    public var field31 = "child.field3";
}
new C().printBase();
```
(with `child.field3` showing up because optimizer couldn't optimize getting a bound method, so it just left a generic `getproperty`)

adrian17 (changes requested):

core/src/avm2/script.rs
```diff
@@ -230,11 +230,15 @@ impl<'gc> TranslationUnit<'gc> {
...
+        let c_class = class
             .c_class()
-            .expect("Class::from_abc_index returns an i_class")
-            .init_vtable(&mut activation.context)?;
+            .expect("Class::from_abc_index returns an i_class");
+
+        class.init_vtable(&mut activation.context)?;
+        c_class.init_vtable(&mut activation.context)?;
+
+        class.validate_class(activation)?;
```
adrian17, line 240: intuitively, shouldn't we do validation before init_vtable? Is there anything blocking reordering these?
(for the record: FP has override validation merged with slot-bindings creation, so reordering these isn't entirely correct, but I'd say better? :D )

core/src/avm2/script.rs
```diff
@@ -230,11 +230,15 @@ impl<'gc> TranslationUnit<'gc> {
...
+        let c_class = class
             .c_class()
-            .expect("Class::from_abc_index returns an i_class")
-            .init_vtable(activation.context)?;
+            .expect("Class::from_abc_index returns an i_class");
+
+        class.validate_class(activation)?;
+        c_class.validate_class(activation)?;
+
+        class.init_vtable(activation.context)?;
```
adrian17, line 240: Can we add a comment that in FP, finalizing the mapping is lazy?
(at least that's how I remember it, right?)

adrian17 (approved):

core/src/avm2/class.rs
```diff
@@ -648,20 +648,48 @@ impl<'gc> Class<'gc> {
...
+                    activation,
+                    &format!(
+                        "Error #1103: Class {} cannot extend final base class.",
+                        read.name.to_qualified_name(activation.gc())
+                    ),
+                    1103,
+                )?));
+            }
+
+            if superclass.is_interface() {
```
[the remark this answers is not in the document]
Lord-McSweeney: `validate_class` is called all classes

Issue #17107, opened 2024-07-13
Transformice: Bad mouse scroll behaviour and it doesn't work in text.TextFields


JediFay:
### Describe the bug

https://github.com/user-attachments/assets/9fad9e50-330c-4be3-8c61-c06ea3c211ef

https://github.com/user-attachments/assets/b55c047e-2189-4629-abcb-c994d20f3239

### Expected behavior

https://github.com/user-attachments/assets/5e079f3f-5a06-4263-9d8c-d9269ca96281

https://github.com/user-attachments/assets/80dbe136-bbeb-4692-83af-56a94b23a569

### Content Location

https://www.transformice.com/TransformiceChargeur.swf

### Affected platform

Desktop app

### Operating system

Windows 11

Issue #13466, opened 2023-10-08
AS3 Mouse Wheel Events Plz


KupoGames:
Mouse Wheel events would be nice to have for my new game.
It's not urgent, but I'll need it eventually.

I need to detect that the mouse wheel is being used, to add a zooming feature in the game.
Probably not a lot of browser games used it, but it's helpful for desktop games!

Thanks.

Lord-McSweeney:
AFAIK we've implemented mouse wheel events in AVM1, but not in AVM2.

PR #17129, opened 2024-07-15
avm2: Dispatch `mouseWheel` event on mouse wheel

kjarosh:
Fixes https://github.com/ruffle-rs/ruffle/issues/17107.

This PR makes sure that besides handling mouse wheel as a clip event, it's also dispatched to AVM2. The order of dispatching AVM2 event vs handling it internally is important, as it introduces observable behaviors in case of text fields.


Lord-McSweeney (approved):

Lord-McSweeney:
Fixes #13466

Dinnerbone (approved):

PR #17158, opened 2024-07-17
web: Use browser focus events for focus management, improve virtual keyboard support

kjarosh:
Reapplication of reverted https://github.com/ruffle-rs/ruffle/pull/17165, with fixed virtual keyboard.

Virtual keyboard on web is represented as an input inside the container. In order to preserve player focus when virtual keyboard is open, this PR moves the focus management to the container from the canvas. This way, when the input is focused, the container does not lose its focus.

This PR also integrates the virtual keyboard with the newly added focus management and removes Android-specific code, instead using generic logic which takes advantage of improved focus support in SWFs. It also fixes some issues related to the keyboard flickering when selecting text.

### DIFF
--- a/core/src/backend/ui.rs
+++ b/core/src/backend/ui.rs
@@ -91,8 +91,9 @@ pub trait UiBackend: Downcast {
     // Unused, but kept in case we need it later.
     fn message(&self, message: &str);
 
-    // Only used on web.
     fn open_virtual_keyboard(&self);
+
+    fn close_virtual_keyboard(&self);
 
     fn language(&self) -> LanguageIdentifier;
 
@@ -359,6 +360,8 @@ impl UiBackend for NullUiBackend {
     }
 
     fn open_virtual_keyboard(&self) {}
+
+    fn close_virtual_keyboard(&self) {}
 
     fn language(&self) -> LanguageIdentifier {
         US_ENGLISH.clone()
--- a/core/src/focus_tracker.rs
+++ b/core/src/focus_tracker.rs
@@ -107,6 +107,8 @@ impl<'gc> FocusTracker<'gc> {
         // Mouse focus change events are not dispatched when the object is the same,
         // contrary to key focus change events.
         if InteractiveObject::option_ptr_eq(old, new) {
+            // Re-open the keyboard when the user clicked an already focused text field.
+            self.update_virtual_keyboard(context);
             return;
         }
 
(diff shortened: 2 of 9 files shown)


core/src/backend/ui.rs
```diff
@@ -94,6 +94,9 @@ pub trait UiBackend: Downcast {
     // Only used on web.
     fn open_virtual_keyboard(&self);
 
+    // Only used on web.
+    fn close_virtual_keyboard(&self);
```
torokati44, line 98: It _may_ also end up getting used in the native Android app, eventually, maybe...

Croworbit:
will this be compatible with "two-in-one" laptops/ desktops with touchscreen monitors?

core/src/backend/ui.rs
kjarosh: You're right, it might also be useful for desktop when we start supporting touch, so I've removed the comments :)

core/src/focus_tracker.rs
```diff
@@ -157,7 +161,11 @@ impl<'gc> FocusTracker<'gc> {
                         .set_selection(Some(TextSelection::for_range(0, length)), context.gc());
```
danielhjacobs, line 161: I don't care too much, but this does more than just update the virtual keyboard. It also sets the text field's selection. So I'm not sure this is the best function name.

danielhjacobs (approved):
I think this is fine as long as it actually works on Android and iOS. I haven't actually tested it. It seems fine from the JavaScript and web_sys perspective, but I still don't fully trust my judgement when it comes to Rust. I will test this on Android as soon as I can, probably tonight. Maybe @n0samu can test it on iOS.

Croworbit:
> It works perfectly on Android! However I do not have a PC with a touchscreen to test on other OSes

when this gets merged, would there be any special steps to try it on desktop?

core/src/focus_tracker.rs
kjarosh: You're right, I just assumed this was keyboard-related code without looking too deep into this, I removed the method.

I also checked whether we need this at all, and I've found that this behavior is slightly inaccurate, so I added a comment. Perhaps this code could be moved somewhere else in the future.

kjarosh:
> when this gets merged, would there be any special steps to try it on desktop?

No special steps! The browser decides whether a virtual keyboard is needed and shows it if it is. The idea behind the keyboard mechanics hasn't changed, so it should work roughly the same as before https://github.com/ruffle-rs/ruffle/pull/17051.

Croworbit:
and final question for this, is it web only?

lvyitian:
> and final question for this, is it web only?

Currently, yes.

n0samu:
I'm not sure I'll have time to compile and test this, but I'd gladly test it on my phone if someone puts up a demo build on github.io or similar

kjarosh:
@n0samu @danielhjacobs you can test it here: https://kjarosh.github.io/ruffle/pr17158/

I've just noticed that removing characters does not work properly: when I input text with length X, I have to first click backspace X times before the text is being removed. But that behavior was present before the regression, so I think we could focus on it in a future PR.

web/packages/core/src/shadow-template.ts
```diff
@@ -77,6 +77,7 @@ export function applyStaticStyles(styleElement: HTMLStyleElement) {
 
         `#container {
             overflow: hidden;
+            outline: none;
```
danielhjacobs, line 80: This change now needs to be made here:
https://github.com/ruffle-rs/ruffle/blob/e25d97195eb6a2890d4015df9916ef16f0e1ed37/web/packages/core/src/internal/ui/static-styles.tsx#L50

kjarosh: Thanks!

Dinnerbone: Now it's in the .css file
Sorry :D

kjarosh: <image src="https://github.com/user-attachments/assets/eeac9487-cb89-4ff2-815a-fc10523c6094" width="300"/>

danielhjacobs (approved):

PR #17160, opened 2024-07-17
core: Add support for non-zero winding rule

kjarosh:
This PR adds support for non-zero winding rule, which is used for rendering fonts and paths in AVM2 using `drawPath` and `GraphicsPath`.

| FP | Before | After |
| -- | ------ | ----- |
|  |  |  |


kjarosh:
I meant of course non-zero, not even-odd, duh!

Dinnerbone (approved):
Thank you!

PR #17161, opened 2024-07-17
web: Use tsx for shadow template

danielhjacobs:


Dinnerbone (approved):
Thank you!

A future change maybe: Let's include the css from an actual css file, at build time. Ideally we'd use a loader so we can do the fancy `import classes from "foo.css"` but that would force all consumers to use the same loader. Build time inclusion would be best and then we can lint it and get proper highlighting etc

danielhjacobs:
Agreed on the CSS build time inclusion, but I think I'd rather leave that for a future PR.

PR #17169, opened 2024-07-18
ci: Temporarily pin to Node 22.4

danielhjacobs:
See https://github.com/nodejs/node/issues/53902

### DIFF
--- a/.github/workflows/test_web.yml
+++ b/.github/workflows/test_web.yml
@@ -35,7 +35,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node_version: ["20", "22"]
+        node_version: ["20", "22.4"]
         os: [ubuntu-22.04, windows-latest]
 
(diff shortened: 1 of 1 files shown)


Dinnerbone (approved):

torokati44:
Node 22.5.1 is now out with the fix, so I guess this can be reverted now?

danielhjacobs:
Not yet, it's not yet in https://github.com/actions/node-versions/releases/

torokati44:
If I see correctly, that repo is updated automatically every 12 hours, so it should be there in either 10 or 70 minutes.

Algiuxs:
It's there

torokati44:
Yes, see: https://github.com/ruffle-rs/ruffle/pull/17181#issuecomment-2239527084
Now we're waiting for https://github.com/actions/node-versions/pull/182.

PR #17171, opened 2024-07-19
wgpu: Directly copy BitmapData gpu buffer to Stage3D texture

Aaron1011:
We now immediately flush our in-progress buffer CommandEncoder after executing a `CopyBitmapToTexture` command, to ensure that we don't pick up any future avm2-side changes to the BitmapData gpu texture.

The Vec -> gpu buffer copy implementation is moved to a new `CopyBytesToTexture` command, which we use for copying AGAL textures.

### DIFF
--- a/core/src/avm2/globals/flash/display3D/textures/cube_texture.rs
+++ b/core/src/avm2/globals/flash/display3D/textures/cube_texture.rs
@@ -88,10 +88,11 @@ pub fn upload_from_bitmap_data<'gc>(
             let mip_level = args[2].coerce_to_u32(activation)?;
             if mip_level == 0 {
                 texture.context3d().copy_bitmapdata_to_texture(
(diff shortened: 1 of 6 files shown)


Fancy2209:
World 4 Part 1 seems to break with this
thread 'main' panicked at C:\Users\paulo\.cargo\registry\src\index.crates.io-6f17d22bba15001f\wgpu-0.20.1\src\backend\wgpu_core.rs:2996:5:
wgpu error: Validation Error

Caused by:
    In CommandEncoder::copy_texture_to_texture
    Copy error
    Copy of Y 0..4095 would end up overrunning the bounds of the Source texture of Y size 3571

Aaron1011:
This needs additional work to ensure that we submit pending bitmapdata gpu commands before copying from the texture

PR #17179, opened 2024-07-19
desktop: Detect light theme on Linux

kjarosh:
The function `start_dbus_theme_watcher` detects system theme on freedesktop-compliant systems and starts a watcher which listens for changes.

Tested on GNOME and KDE, Fedora and Manjaro.

| Before | After |
| ------ | ----- |
| <image src="https://github.com/user-attachments/assets/1703164c-8cda-43c0-a619-d188298d84cd" width="300"/> | <image src="https://github.com/user-attachments/assets/d2fd71f3-a395-4b8a-85cb-1c9c6a3935ae" width="300"/> |
| <image src="https://github.com/user-attachments/assets/2cd2f4fd-c9fd-419b-9d9c-9c14f803325c" width="300"/> | <image src="https://github.com/user-attachments/assets/6a8b6106-2b53-4e78-a134-110d173bb97a" width="300"/> |

Before vs after:

https://github.com/user-attachments/assets/9d671e23-8ad5-4b33-8cb7-05b04a29cf4f

### DIFF
--- a/desktop/src/dbus.rs
+++ b/desktop/src/dbus.rs
@@ -1,0 +1,82 @@
+#![cfg(target_os = "linux")]
+//! Types and methods utilized for communicating with D-Bus
+
+use futures::StreamExt;
+use zbus::export::futures_core::Stream;
+use zbus::zvariant::{OwnedValue, Value};
+use zbus::{proxy, Connection};
+
+type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;
+
+#[proxy(
+    interface = "org.freedesktop.portal.Settings",
+    default_service = "org.freedesktop.portal.Desktop",
+    default_path = "/org/freedesktop/portal/desktop",
+    gen_blocking = false
+)]
+trait FreedesktopSettingsInterface {
+    fn read(&self, namespace: &str, key: &str) -> Result<OwnedValue>;
+
+    #[zbus(signal)]
+    fn setting_changed(&self, namespace: &str, key: &str, value: Value<'_>);
+}
+
+pub enum ColorScheme {
+    Default,
+    PreferLight,
+    PreferDark,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum FreedesktopSettingsError {
+    #[error("Unexpected value for color scheme: {0}")]
+    UnexpectedColorScheme(Value<'static>),
+}
+
+pub struct FreedesktopSettings<'p> {
+    proxy: FreedesktopSettingsInterfaceProxy<'p>,
+}
+
+impl<'p> FreedesktopSettings<'p> {
+    const COLOR_SCHEME: (&'static str, &'static str) =
+        ("org.freedesktop.appearance", "color-scheme");
+
+    pub async fn new(connection: &Connection) -> Result<Self> {
+        Ok(Self {
+            proxy: FreedesktopSettingsInterfaceProxy::new(connection).await?,
+        })
+    }
+
+    pub async fn color_scheme(&self) -> Result<ColorScheme> {
+        let (namespace, key) = Self::COLOR_SCHEME;
+        let value = self.proxy.read(namespace, key).await?;
+        Self::parse_color_scheme(value.into())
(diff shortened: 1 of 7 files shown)


Dinnerbone (approved):
Neat!

torokati44:
Uhm, sorry but - why not use `zbus`, which we already have in our dep tree, instead of spawning a process which may do "anything"?

torokati44:
Also, maybe we could rely on https://github.com/frewsxcv/rust-dark-light instead?

kjarosh:
> Also, maybe we could rely on https://github.com/frewsxcv/rust-dark-light instead?

That unfortunately has a lot of issues, it does not work for me at all.

The zbus code from there errors for me during deserialization, and even if it didn't, it wouldn't report the proper theme because GNOME uses the value of `0` (default) when light theme is used, and that library tries to guess the theme in that case based on GTK theme name, which does not make sense since GNOME 43 (which shipped with dark theme support). And that's only GNOME, I would also have to test other DEs.

> Uhm, sorry but - why not use `zbus`, which we already have in our dep tree, instead of spawning a process which may do "anything"?

I'll try rewriting that to use `zbus`, that's a good idea!

torokati44:
> That unfortunately has a lot of issues, it does not work for me at all.

Ah, a legit reason to not add yet another dependency, yea! 😅

> I'll try rewriting that to use `zbus`, that's a good idea!

👌

kjarosh:
Rewrote the code and used `zbus` for communicating with D-Bus, made Ruffle listen for theme changes by the way (see the video from the description). I hope you like it!

torokati44 (approved):
I haven't looked through all the changes in detail, but it's probably fine - I like this approach a lot better than the process one.

Croworbit:
does the background turn white too, or just the menus?
does windows have this too?

kjarosh:
> does the background turn white too, or just the menus?

Just the menus, the background is always black.

> does windows have this too?

If you're referring to the OS, then according to winit's docs it should work there.

Croworbit:
how come the background is always black?
a limitation, or just preference?

kjarosh:
> how come the background is always black? a limitation, or just preference? its like a 'mixed' theme

That's not a limitation. Light theme does not mean that there are absolutely no dark elements, and black backgrounds are commonly used in apps to signify that no content is currently opened (e.g. in video players). If you think that the background should be light in light theme you are free to file an issue about that so we can discuss that idea!

Croworbit:
just a curiousity 
I see no problems here

torokati44:
To be fair, FP has a white background to signify "no content". But I prefer keeping it black nevertheless.

desktop/src/dbus.rs
```diff
@@ -0,0 +1,82 @@
...
+    #[error("Unexpected value for color scheme: {0}")]
+    UnexpectedColorScheme(Value<'static>),
+}
+
+pub struct FreedesktopSettings<'p> {
+    proxy: FreedesktopSettingsInterfaceProxy<'p>,
+}
+
+impl<'p> FreedesktopSettings<'p> {
+    const COLOR_SCHEME: (&'static str, &'static str) =
```
evilpie, line 40: Feels like this tuple actually requires more code to unpack every time it is used.

kjarosh: More code compared to what? :thinking:

evilpie: Just having two named consts.

kjarosh: Unfortunately using two constants will make the code two lines longer, because their names will be longer. It's true that using a tuple requires an additional line for unpacking but it makes it possible to name those constants locally (which would require two lines when using two constants), and ensures that they are always used together (they are basically inseparable, you always use both or none).

Croworbit:
> To be fair, FP has a white background to signify "no content". But I prefer keeping it black nevertheless.

I think this point has me realize that neither a black or white screen is a very good indication to show emptiness when some SWFs do exactly that when they break after loading

PR #17182, opened 2024-07-19
web: Include CSS at build time, from a CSS file

Dinnerbone:
This gives us the benefit of linting and IDE checking, since it's an actual CSS file.

Bonus changes:
- Use cssnano to compress the CSS at build time
- Use autoprefixer for the things that we need to support older browsers
- Use nesting in the CSS file, unrolled at build time

The diff of the old CSS and new (generated) CSS as of this commit is [here](https://gist.github.com/Dinnerbone/e0938e59668e7aaea068239463055ae4/revisions#diff-8066fb178be59f0a7d67c9e62c0910b9f651c91a3e140ea79cbd07033aa20476)


web/packages/core/.browserlistrc
```diff
@@ -0,0 +1,8 @@
+# Browsers that we support
+
+defaults
+chrome >= 87
+edge >= 87
+firefox >= 84
+iOS >= 14.1
+safari >= 14.1
```
danielhjacobs, line 8: Not opposed to the addition of this file, but is this change related to this PR? I can't find any reference to it in the new code, though I may have missed it.

Dinnerbone: autoprefixer uses it to decide which prefixes need to be added to support these browsers

Dinnerbone: Decided to check the css to see if this is *actually* impactful, and it did catch at least one thing:

We used: `width: fit-content;`
It replaced it with:
```css
width: -moz-fit-content;
width: fit-content
```

Turns out `fit-content` isn't supported in firefox until 94

danielhjacobs: On a related note, not sure why the generated CSS is adding `-moz-user-select` when full support for user-select was added in Firefox 69 and Firefox for Android 79, which is above our minimum supported version, but it's fine.

danielhjacobs: -webkit-user-select is still useful though, as Safari uses that: https://developer.mozilla.org/en-US/docs/Web/CSS/user-select#browser_compatibility

danielhjacobs (approved):
I reviewed this and https://gist.github.com/Dinnerbone/e0938e59668e7aaea068239463055ae4/revisions (but didn't directly review `static-styles.css` since it's hard to see the changes from there). This LGTM, though for the SVG CSS changes I just built the branch itself and checked if they still looked correct, since reviewing text changes to an SVG's path is hard.

PR #17193, opened 2024-07-20
ci: Add stable releases

kjarosh:
A PoC of stable releases for Ruffle.


web/packages/extension/webpack.config.js
```diff
@@ -41,11 +42,7 @@ function transformManifest(content, env) {
     // when it gets generated in web/packages/core/tools/set_version.js and then
     // load it in the code above.
 
-    // The extension marketplaces require the version to monotonically increase,
-    // so append the build number onto the end of the manifest version.
-    manifest.version = buildId
-        ? `${packageVersion}.${buildId}`
-        : packageVersion;
+    manifest.version = packageVersion;
```
Dinnerbone, line 45: Is this handled somewhere else now?

.github/workflows/release.yml
```diff
@@ -50,20 +68,125 @@ jobs:
...
+            package_prefix="ruffle-${{ steps.version.outputs.version }}"
+          fi
+
+          gh release create "$tag_name" \
+            --title "$release_name" \
+            --target "${{ steps.commit.outputs.ref }}" \
+            $notes_options
+
+          echo "tag_name=$tag_name" | tee -a $GITHUB_OUTPUT
+          echo "package_prefix=$package_prefix" | tee -a $GITHUB_OUTPUT
```
Dinnerbone, line 182: Could we maybe extract this into a `.sh` somewhere? I'm not a fan of how long and complex our workflow is currently... and this doesn't help 😅
I'm kinda of the opinion that any non trivial block of code should be its own unit

web/packages/extension/webpack.config.js
kjarosh: This will have to be rewritten, because I assumed I can use semver in the extension (these versions are unique and ordered). The problem is similar to wix's, we have to somehow translate nightly `a.b.c-nightly.d.e.f` versions to `a.b.c.d` here.

I can use `github.run_number` for that (like before) or something like number of days since a point in time. But I think it's wise to wait with that for @Korne127's PR and improved Windows metadata to make a decision.

Dinnerbone: We'll switch the extension to stable releases, so we don't need to handle nightly here

kjarosh: But we'll publish extensions anyway in nightly releases on GH, wouldn't that be a problem? (Probably not if the version requirement is only for extension stores and not extensions installed locally)

Dinnerbone: Yeah I don't think it'll be an issue, it's a store requirement. Will need to test it of course though.

And we *may* eventually want to make a separate nightly extension submission, to match what we have today - not sure. It's unlikely though.

kjarosh: Hmmm... will we need to provide `a.b.c.d` versions for wix though? You know, to preserve version ordering when users install nightly releases. If yes, then I guess it costs nothing to do the same for extensions.

Dinnerbone: Yeah for windows builds we need to stick to `a.b.c.d` number format (for wix/msi and also the exe resource data), and MSI specifically will only test the first three numbers to see "is the version different"

.github/workflows/release.yml
kjarosh: Moved a lot of release logic into a Python script

web/packages/extension/webpack.config.js
kjarosh: I added a version in format `a.b.c.d`, and called it `version4`. It's generated as follows:
1. for stable minor: `1.2.3` (current version) -> `1.3.0` (version) -> `1.3.0` (version4)
2. for nightly: `1.2.3` (current version) -> `1.3.0-nightly.1234.5.6` (version) -> `1.2.3.D` (version4)

Where `D` is a number related to the current day (`year - 2020` concatenated with zero-padded day of the year, so for today the number is 4221).

.github/workflows/release.yml
torokati44: Yay! Just noting that I wrote the AMO submission script in Node.JS instead of Python, because "we already use that". I too prefer Python personally, but this breaks consistency, and adds a(n admittedly ubiquitous) tooling requirement to CI.

kjarosh: I'm wondering if it is really. The AMO submission is a part of an npm package the same way e.g. build.rs is a part of a crate, so it makes sense it uses the same tooling. The release script is independent from any package, and Python is available OOTB on runners.

Dinnerbone: ... personally I've been pondering on replacing some of our node scripts with rust ones 😅 

I kinda don't strongly either way and I especially don't want to put kjarosh through rewriting everything another time, but I think I'd *prefer* to not have yet another language in the mix, if we didn't have to. That said, rust is verbose for something like this, and javascript sucks.

torokati44: >  I especially don't want to put kjarosh through rewriting everything another time, but I think I'd prefer to not have yet another language in the mix, if we didn't have to. That said, rust is verbose for something like this, and javascript sucks.

Same, agreed, yup, sure. :smile: I'm fine with leaving it in Python, just wanted to mention this precedent.

kjarosh:
Added updating metainfo in the release commit

kjarosh:
Some changes from this PR were merged in other PRs, the rest seem unlikely to be merged

n0samu:
Can we please keep this open unless/until we've made a formal decision on how/whether we're going to proceed with stable releases? I strongly support moving towards stable releases and appreciate your work in this PR ❤️

kjarosh:
Thanks, I appreciate it, but regardless, the code here is very obsolete and is unlikely to be merged even if we ever decide we go with stable releases

Korne127:
@n0samu I originally did #17488 to clean up and improve the packaging system, which was imo a requirement for stable releases. If wanted, I could still update and go through with that and / or split it up into several PRs (I've come to terms with that by now).

n0samu:
@Korne127 I think splitting it into several PRs would be very helpful! Each individual small change should be its own PR (for example, adding AppImage builds)

PR #17212, opened 2024-07-21
Small `AvmString` cleanup

moulins:
Small `AvmString`-related cleanups and optimisations:

- Rename `Source::Owned` to `Source::Managed`, to avoid confusion with the owned/depending distinction in `AvmStringRepr`
- Store a raw `Gc<AvmStringRepr>` instead of an `AvmString` in `AvmStringRepr`'s `owner` field; there's no reason to ever make a dependant string pointing to a static `AvmString`.


torokati44 (approved):
I haven't read through all the changes thoroughly, and I'm not even that familiar with this area of the code, but the changes make sense to me, and I trust the author in these matters. So here's a faint, cautious approval, just to perhaps help this move along.

PR #17213, opened 2024-07-21
web: Fix globals.d.ts file type declarations

danielhjacobs:
Apparently I don't know how this works. I need to import the old type declarations before defining new ones.

### DIFF
--- a/web/packages/core/src/globals.d.ts
+++ b/web/packages/core/src/globals.d.ts
@@ -1,56 +1,36 @@
-interface Error {
-    avmStack?: string;
-}
-// Just updating this module seems to disable type checking for tsx-dom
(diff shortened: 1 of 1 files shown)


web/packages/core/src/globals.d.ts
```diff
@@ -1,56 +1,30 @@
...
+
+declare global {
+    interface Error {
+        avmStack?: string;
+    }
+    interface SVGElement {
+        xmlns?: string;
+        fill?: string;
+        stroke?: string;
+        scale?: string | number;
```
danielhjacobs, line 21: This can be removed after #17224

Dinnerbone (approved):

PR #17214, opened 2024-07-21
Add flash.display.StageAspectRatio

Fancy2209:
also stubs stage.setAspectRatio
Currently a draft because I somehow made it impossible to load SWF


Fancy2209:
Somehow SWF can't load anymore, what did I do 😓

Fancy2209:
It seems API Versioning is what broke ruffle?????

core/src/avm2/globals/flash/display/StageAspectRatio.as
```diff
@@ -0,0 +1,13 @@
...
+{
+    [API("668")]
+    public final class StageAspectRatio
+    {
+        [API("681")]
+        public static const ANY:String = "any"
+        [API("668")]
+        public static const LANDSCAPE:String = "landscape"
+        [API("668")]
+        public static const PORTRAIT:String = "portrait"
```
kjarosh, line 11: You are missing semicolons, hence the errors. Also, I think annotating these constants as `[API("668")]` is unnecessary, as the whole class is annotated that way.

Fancy2209: Fixed

kjarosh (approved):

torokati44:
Thank you for cleaning up the commit history! :pray:

PR #17219, opened 2024-07-22
avm2: Stub orientation-related Stage API

kjarosh:
A follow-up of https://github.com/ruffle-rs/ruffle/pull/17214. This PR stubs the rest of API related to orientation. If something uses `Stage.setAspectRatio`, it's likely `Stage.supportsOrientationChange` and others will be useful.


torokati44 (approved):
Thank you!

PR #6843, opened 2022-05-01
web: Don't use inline CSS attributes

GhostlyDark:
The SVG files as part of `shadow-template.ts` use inline CSS which can be completely avoided.
This fixes potential issues with CSP configurations blocking `style-src-attr`.

Ruffle still requires `style-src-elem 'unsafe inline'` because of the other two style elements in `shadow-template.ts`.

PR #17224, opened 2024-07-22
web: Use CSS scale attribute instead of non-existent scale property

danielhjacobs:
This has been incorrect since https://github.com/ruffle-rs/ruffle/pull/6843. The SVG `scale` attribute only exists on \<feDisplacementMap>: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/scale


evilpie (approved):

evilpie:
Can you maybe post a screenshot before/after?

PR #17233, opened 2024-07-23
web: Improve context menu right-click behavior

WumboSpasm:
Right-clicking over the context menu no longer re-positions it, and can now trigger menu item actions.

As a result of these changes, the below issue is now fixed:

https://github.com/user-attachments/assets/95612841-4f76-4317-b678-f326a18bd354

In addition, left-clicking a non-clickable part of the context menu no longer closes it.

### DIFF
--- a/web/packages/core/src/internal/player/inner.tsx
+++ b/web/packages/core/src/internal/player/inner.tsx
@@ -260,6 +260,13 @@ export class InnerPlayer {
             "context-menu-overlay",
         )!;
         this.contextMenuElement = this.shadow.getElementById("context-menu")!;
+        const preserveMenu = (event: MouseEvent) => {
+            event.preventDefault();
+            event.stopPropagation();
+        };
+        this.contextMenuElement.addEventListener("contextmenu", preserveMenu);
+        this.contextMenuElement.addEventListener("click", preserveMenu);
+
         document.documentElement.addEventListener(
             "pointerdown",
             this.checkIfTouch.bind(this),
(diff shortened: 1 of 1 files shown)


n0samu:
On Windows at least, right-clicking over a context menu item seems to behave the same as clicking the item. Just pointing this out if we want to emulate proper context menu behavior.

WumboSpasm:
Done

web/packages/core/src/ruffle-player.tsx
```diff
@@ -1596,6 +1596,45 @@ export class RufflePlayer extends HTMLElement {
...
+            const [x, y] = getContextMenuOrigin(playerRect, contextMenuRect);
+
+            const minX = contextMenuRect.x - playerRect.x;
+            const minY = contextMenuRect.y - playerRect.y;
+            const maxX = minX + contextMenuRect.width;
+            const maxY = minY + contextMenuRect.height;
+
+            if (x! >= minX && x! <= maxX && y! >= minY && y! <= maxY)
+                return;
+        }
```
kjarosh, line 1636: We probably shouldn't perform low-level calculations of x,y positions, but use the event model instead. In this case the following should work in the constructor:
```ts
this.contextMenuElement.addEventListener("contextmenu", (event) => {
    event.stopPropagation();
    event.preventDefault();
});
```

kjarosh:
Also, as mentioned on Discord, would be nice to test it out on iOS

Korne127:
I kind of think it's more user-friendly to do nothing as right-click is something different, but I don't think it matters much, and both is definitely a big improvement to the current state! Thanks :)

kjarosh (approved):
Approved assuming it works on mobile Safari

kjarosh:
Deployed for testing: https://kjarosh.github.io/ruffle/pr17233/

PR #17237, opened 2024-07-23
chore: Fix new beta lint `clippy::manual_inspect`

torokati44:
See for example: https://github.com/ruffle-rs/ruffle/actions/runs/10058530166/job/27801842401?pr=17230

### DIFF
--- a/core/src/bitmap/turbulence.rs
+++ b/core/src/bitmap/turbulence.rs
@@ -229,15 +229,14 @@ impl Turbulence {
             let noise = self.noise2(color_channel, vec, stitch_info);
             sum += if fractal_sum { noise } else { noise.abs() } / ratio;
             ratio *= 2.0;
-            stitch_info.as_mut().map(|stitch_info| {
+            if let Some(ref mut stitch_info) = stitch_info {
                 // Update stitch values. Subtracting PerlinN before the multiplication and
                 // adding it afterward simplifies to subtracting it once.
                 stitch_info.width *= 2;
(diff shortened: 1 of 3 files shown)


core/src/bitmap/turbulence.rs
```diff
@@ -229,14 +229,15 @@ impl Turbulence {
             let noise = self.noise2(color_channel, vec, stitch_info);
             sum += if fractal_sum { noise } else { noise.abs() } / ratio;
             ratio *= 2.0;
-            stitch_info.as_mut().map(|stitch_info| {
+            stitch_info = stitch_info.map(|si| {
```
kjarosh, line 232: Can't we do the following here?

```rust
stitch_info.as_mut().inspect(|stitch_info| {
```

torokati44: That's what I tried first, but couldn't find a way to allow the closure to mutate its parameter. And exactly because it mutates, wouldn't it be a bit confusing that way?

kjarosh: Oh, you're right, I missed that. I believe we can do the following then, which is IMO more readable.
```rust
if let Some(ref mut stitch_info) = stitch_info {
    // ...
}
```

Because otherwise we're unnecessarily constructing a new object.

torokati44: Oh, yep, that works. Thanks!

kjarosh (approved):

PR #17253, opened 2024-07-24
ci: Don't test on Edge as the latest release is missing for Linux

danielhjacobs:

PR #17255, opened 2024-07-24
Revert "ci: Don't test on Edge as the latest release is missing for Linux"

torokati44:
Reverts ruffle-rs/ruffle#17253
To be merged once https://github.com/MicrosoftEdge/EdgeWebDriver/issues/156 is fixed.

### DIFF
--- a/.github/workflows/test_web.yml
+++ b/.github/workflows/test_web.yml
@@ -96,7 +96,7 @@ jobs:
         # Don't run browser tests on Windows because it's flaky for unknown reasons. :-(
         if: runner.os == 'Linux'
         working-directory: web
-        run: npm run wdio -- --headless --chrome --firefox
+        run: npm run wdio -- --headless --chrome --firefox --edge
 
   check-required:
     needs: changes


torokati44:
I wonder whether this is a `wdio` bug - why doesn't it try to use the installed Edge WebDriver 126.0.2592.102 (even though Edge is 126.0.2592.113).
Or maybe it's a GHA hosted Ubuntu image bug, since ideally the installed versions should match exactly?

danielhjacobs:
> Or maybe it's a GHA hosted Ubuntu image bug, since ideally the installed versions should match exactly?

I did open https://github.com/actions/runner-images/issues/10336

torokati44:
Looks like this works now, @danielhjacobs. Let's hope it doesn't reoccur soon...

danielhjacobs:
While it's fixed for now, runner-images will create a PR for a new image within the day. I want to see what Edge version and Edgedriver version it uses and see if that version has a Linux release yet.

danielhjacobs:
Yup: https://github.com/actions/runner-images/blob/bc9dc728e1e3b41ffed3cd950db959f3c2475fd2/images/ubuntu/Ubuntu2204-Readme.md#browsers-and-drivers

torokati44:
There's still a tiny mismatch: 

> Microsoft Edge 127.0.2651.74
> Microsoft Edge WebDriver 127.0.2651.72

danielhjacobs:
Yeah, and that will cause an issue:

https://msedgedriver.azureedge.net/127.0.2651.74/edgedriver_linux64.zip does not exist.

danielhjacobs:
See also: https://github.com/webdriverio/webdriverio/issues/13251

I now have issues open with all the relevant parties, so we'll see if any are willing to fix this.

PR #17278, opened 2024-07-27
avm2: Don't include interface methods in describeTypeJSON

Dinnerbone:


core/src/avm2/globals/avmplus.rs
```diff
@@ -355,6 +355,19 @@ fn describe_internal_body<'gc>(
                 let method = vtable
                     .get_full_method(*disp_id)
                     .unwrap_or_else(|| panic!("Missing method for id {disp_id:?}"));
+
+                // Don't include methods that also exist in any interface
+                let mut is_interface = false;
+                for interface in method.class.all_interfaces().iter() {
```
evilpie, line 361: I think `iter().any(|interface| .. )` would be a bit cleaner instead of a mutable variable.

core/src/avm2/globals/avmplus.rs
```diff
@@ -305,6 +305,10 @@ fn describe_internal_body<'gc>(
             Some(ns.as_uri())
         };
 
+        if !ns.matches_api_version(activation.avm2().root_api_version) {
```
evilpie, line 308: Maybe move this up to the first if?

evilpie (approved):

Lord-McSweeney (approved):

PR #17279, opened 2024-07-27
[AVM1 debug ui] Make some object properties editable

crumblingstatue:
Changing object properties can be useful for helping debugging games, or just cheating/general messing around.

Also adds a filter to quickly find fields of interest.

Todo
- [ ] ~~Attempt to add AVM2 support~~
   AVM2 debug ui code would require significant changes to add editing support, and I want to keep this pull request fairly simple. Perhaps a followup/intermediate step would be some refactoring and unification work on the AVM1 and AVM2 debug UI code.
- [x] Optionally try to add support to change type of a value
- [x] Try using `WString` as key instead of `String`

Screenshot of what it looks like:
image

### DIFF
--- a/core/src/debug_ui/avm1.rs
+++ b/core/src/debug_ui/avm1.rs
@@ -12,6 +12,8 @@ pub struct Avm1ObjectWindow {
     key_filter_string: String,
     edited_key: Option<String>,
     value_edit_buf: String,
+    /// True if the active text edit should be focused (after clicking 'edit', etc.)
+    focus_text_edit: bool,
 }
 
 impl Avm1ObjectWindow {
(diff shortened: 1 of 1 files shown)


core/src/debug_ui/avm1.rs
```diff
@@ -39,23 +47,273 @@ impl Avm1ObjectWindow {
...
+/// Dropdown menu indicating the type of the value, as well as letting the
+/// user set a new type.
+fn show_value_type_combo_box<'gc>(
+    ui: &mut Ui,
+    key: &WStr,
+    value: &Value<'gc>,
+    mutation: &Mutation<'gc>,
+) -> Option<Value<'gc>> {
+    let mut new = None;
+    egui::ComboBox::new(egui::Id::new("value_combo").with(key), "Type")
```
crumblingstatue, line 239: Perhaps the "Type" label is a bit redundant, and the dropdown menu is obvious enough without the label.

Dinnerbone (approved):
Thank you! LGTM

PR #17289, opened 2024-07-28
chore: Update translations

RuffleBuild:
New Crowdin pull request with translations


torokati44:
According to Google Translate, this says "link" in place of "URL"...? But of course, I wouldn't trust that entirely...

torokati44 (approved):
Meticulously translating all the translations back to English using Google Translate didn't turn up anything obviously wrong or problematic, so merging.
Individual languages can always be fixed/improved upon later on a case by case basis.
Thank you, translators!

PR #17299, opened 2024-07-29
avm2: stub desktop AIR APIs

Mesteery:


evilpie:
Can you remove the AIR argv handling from this PR? I think we shouldn't combine it.

core/src/avm2/globals/flash/display/ScreenMode.as
```diff
@@ -0,0 +1,40 @@
+package flash.display
+{
+  import __ruffle__.stub_getter;
+  import flash.geom.Rectangle;
+
+  // [API("733")] Ruffle doesn't support this API Version
```
evilpie, line 6: We have to use _some_ AIR version, otherwise it's available to normal Flash.

Mesteery: What version do you suggest?

Mesteery:
> Can you remove the AIR argv handling from this PR? I think we shouldn't combine it.

I put air argv handling in the same PR because it's needed by NativeApplication.
But yes I can split if you want.

core/src/avm2/globals/flash/display/ScreenMode.as
Fancy2209: This is from what I helped with the PR, my bad! didn't notice I left the class unversioned
Should probably use AIR 1.0 (`[API(661)]`)

core/src/avm2/globals/flash/desktop/native_application.rs
```diff
@@ -0,0 +1,17 @@
+use crate::avm2::{Activation, ArrayObject, ArrayStorage, Error, Object, Value};
+use crate::string::AvmString;
+
+pub fn get_argv_internal<'gc>(
```
evilpie, line 4: I am not a fan of the c-ism `argv`. Maybe `args` like Rust or `arguments` like Flash?

Mesteery: I agree, `arguments` is better, I will change that.

evilpie:
This PR is huge and does a lot of different stuff in one commit. @Mesteery if you split it up it will be easier and quicker to review.
Please add least split out the air_arguments and screen_size code.

Mesteery:
Splitted into #17609 and #17608

core/src/avm2/globals/flash/desktop/NativeApplication.as
```diff
@@ -0,0 +1,259 @@
...
+    public function set executeInBackground(param1:Boolean):void
+    {
+      stub_setter("flash.desktop.NativeApplication", "executeInBackground");
+    }
+
+    // [API("721")] Ruffle doesn't support this API Version
+    public function get isCompiledAOT():Boolean
+    {
+      stub_getter("flash.desktop.NativeApplication", "isCompiledAOT");
+      return true;
```
evilpie, line 250: I think `false` makes more sense here. 

>  This API returns (true) only for iOS for targets(ipa-app-store, ipa-test, ipa-debug, ipa-ad-hoc).

https://airsdk.dev/reference/actionscript/3.0/flash/desktop/NativeApplication.html#isCompiledAOT

core/src/avm2/globals/flash/display/NativeWindow.as
```diff
@@ -0,0 +1,317 @@
...
+    {
+      stub_method("flash.display.NativeWindow", "activate");
+      dispatchEvent(new Event(Event.ACTIVATE));
+    }
+
+    // Closes this window.
+    public function close():void
+    {
+      stub_method("flash.display.NativeWindow", "close");
+      if (!dispatchEvent(new Event(Event.CLOSING, false, true)))
```
evilpie, line 171: Isn't the negation wrong? I would think if it returns `false` we should not close?

evilpie (approved):

core/src/avm2/globals/flash/display/NativeWindow.as
Mesteery: Yes indeed, I don’t remember at all why negation is here
```suggestion
      if (dispatchEvent(new Event(Event.CLOSING, false, true)))
```

core/src/avm2/globals/flash/desktop/Icon.as
```diff
@@ -0,0 +1,21 @@
...
+  public class Icon extends EventDispatcher
+  {
+
+    // The icon image as an array of BitmapData objects of different sizes.
+    public var bitmaps:Array;
+
+    function Icon(bitmaps:Array)
+    {
+      this.bitmaps = bitmaps;
+    }
```
torokati44, line 19: I don't see this constructor in the docs... :thinking:

core/src/avm2/globals/flash/desktop/InteractiveIcon.as
```diff
@@ -0,0 +1,27 @@
...
+
+    public function get height():int
+    {
+      return this._height;
+    }
+
+    public function get width():int
+    {
+      return this._width;
+    }
```
torokati44, line 25: Since `_width` and `_height` don't seem to be set (or even be able to set), I don't see the point of them. The getters could just return a constant `0` or `-1` or `undefined`.
Also, a `stub_getter` call would be appropriate here, like in other places, I think.

core/src/avm2/globals/flash/events/NativeWindowBoundsEvent.as
```diff
@@ -0,0 +1,52 @@
...
+
+package flash.events
+{
+
+  import flash.geom.Rectangle;
+
+  [API("661")]
+  public class NativeWindowBoundsEvent extends Event
+  {
+    public static const RESIZE:String = "resize";
```
torokati44, line 14: There are 4 constants in the docs, only one of them here:
 - `MOVING`
 - `MOVE`
 - `RESIZING`
 - `RESIZE`

core/src/avm2/globals/flash/events/NativeWindowBoundsEvent.as
```diff
@@ -0,0 +1,52 @@
...
+    // [override] Creates a copy of the NativeWindowBoundsEvent object and sets the value of each property to match that of the original.
+    override public function clone():Event
+    {
+      return new NativeWindowBoundsEvent(this.type, this.bubbles, this.cancelable, this.beforeBounds, this.afterBounds);
+    }
+
+    // [override] Returns a string that contains all the properties of the NativeWindowBoundsEvent object.
+    override public function toString():String
+    {
+      return this.formatToString("NativeWindowBoundsEvent", "type", "bubbles", "cancelable");
```
torokati44, line 38: Missing fields, the documented format is this: `[NativeWindowBoundsEvent type=value bubbles=value cancelable=value previousDisplayState=value currentDisplayState=value]`

core/src/avm2/globals/flash/events/NativeWindowDisplayStateEvent.as
```diff
@@ -0,0 +1,49 @@
...
+// https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/events/NativeWindowDisplayStateEvent.html
+// by https://github.com/golfinq/ActionScript_Event_Builder
+// It won't be regenerated in the future, so feel free to edit and/or fix
+
+package flash.events
+{
+  [API("661")]
+  public class NativeWindowDisplayStateEvent extends Event
+  {
+    public static const DISPLAY_STATE_CHANGE:String = "displayStateChange";
```
torokati44, line 11: There are two constants documented, only 1 here:
 - `DISPLAY_STATE_CHANGING`
- `DISPLAY_STATE_CHANGE`

The event builder script can probably only handle one constant per event.

core/src/avm2/globals/flash/events/NativeWindowBoundsEvent.as
torokati44: NOTE that this is probably a copy-paste error in the docs (from `NativeDisplayStateEvent`, and the correct format is more likely something like this:
`[NativeWindowBoundsEvent type=value bubbles=value cancelable=value beforeBounds=value afterBounds=value]`

To be tested. Also, it's worth a comment about the doc mistake, if there is indeed one.

core/src/avm2/globals/flash/display/NativeMenuItem.as
```diff
@@ -1,5 +1,7 @@
 package flash.display {
     import flash.events.EventDispatcher;
+
+    [API("661")]
```
torokati44, line 4: The docs disagree with `airglobal.swc` here, this may actually be `667` in reality.

Also, BTW, this class has a bunch of properties that are not added. (This is not a blocker, just mentioning, for completeness.)

core/src/avm2/globals/flash/desktop/NativeApplication.as
```diff
@@ -0,0 +1,259 @@
...
+    {
+      stub_method("flash.desktop.NativeApplication", "publisherID");
+      return "";
+    }
+
+    public function get applicationDescriptor():XML
+    {
+      stub_method("flash.desktop.NativeApplication", "applicationDescriptor");
+      return null;
+    }
```
torokati44, line 105: Copy-paste mistake: `stub_method` -> `stub_getter`

core/src/avm2/globals/flash/desktop/NativeApplication.as
```diff
@@ -0,0 +1,259 @@
...
+    {
+      stub_getter("flash.desktop.NativeApplication", "paste");
+      return false;
+    }
+
+    public function clear():Boolean
+    {
+      stub_getter("flash.desktop.NativeApplication", "clear");
+      return false;
+    }
```
torokati44, line 203: Copy-paste mistake: `stub_getter` -> `stub_method`

core/src/avm2/globals/flash/desktop/NativeApplication.as
```diff
@@ -0,0 +1,259 @@
...
+
+    [API("681")]
+    public function set executeInBackground(param1:Boolean):void
+    {
+      stub_setter("flash.desktop.NativeApplication", "executeInBackground");
+    }
+
+    // [API("721")] Ruffle doesn't support this API Version
```
torokati44, line 246: I think it would be better to still use an `API` annotation, with the highest supported value; but also leaving the correct one as a comment.

core/src/avm2/globals/flash/desktop/NativeApplication.as
```diff
@@ -0,0 +1,259 @@
...
+    public function get isCompiledAOT():Boolean
+    {
+      stub_getter("flash.desktop.NativeApplication", "isCompiledAOT");
+      return false;
+    }
+
+    public function get isActive():Boolean
```
torokati44, line 253: Huh, I can't see this in the docs, only in `airglobal.swc` - but there it has a fairly large API version limit, I guess it should be handled like `isCompiledAOT`.

core/src/avm2/globals/flash/desktop/NativeApplication.as
```diff
@@ -0,0 +1,259 @@
...
+
+    public function get autoExit():Boolean
+    {
+      stub_getter("flash.desktop.NativeApplication", "autoExit");
+      return false;
+    }
+
+    public function set autoExit(param1:Boolean):void
+    {
+      stub_getter("flash.desktop.NativeApplication", "autoExit");
```
torokati44, line 115: Copy-paste mistake: `stub_getter` -> `stub_setter`

core/src/avm2/globals/flash/display/NativeWindow.as
```diff
@@ -0,0 +1,317 @@
...
+    public function startResize(edgeOrCorner:String = "BR"):Boolean
+    {
+      stub_method("flash.display.NativeWindow", "startResize");
+      return false;
+    }
+
+    public function get active():Boolean
+    {
+      return true;
+    }
```
torokati44, line 264: Is this really always correct, or worth a `stub_method` annotation?

core/src/avm2/globals/flash/display/NativeWindow.as
```diff
@@ -0,0 +1,317 @@
...
+      {
+        this._stage = _stage;
+        _stage.addEventListener(Event.RESIZE, function(e:Event):void
+        {
+          dispatchEvent(new NativeWindowBoundsEvent(NativeWindowBoundsEvent.RESIZE, false, false, _bounds, _bounds = new Rectangle(x, y, width, height)));
+        });
+      }
+      else
+      {
+        stub_constructor("flash.display.NativeWindow without stage");
```
torokati44, line 49: Minor nit: I think just (either) one of this, or the generic stub note at the top, should suffice.

Mesteery:
Thanks you for your review. I've normally addressed all the changes

torokati44:
Thank you! 🥳 I'll take another look, and still have two files left to read through - but gotta sleep first. 😌😴

core/src/avm2/globals/flash/display/NativeMenuItem.as
```diff
@@ -1,6 +1,32 @@
 package flash.display {
     import flash.events.EventDispatcher;
+    import __ruffle__.stub_constructor;
+    import __ruffle__.stub_getter;
+
+    [API("667")]
```
torokati44, line 6: nit: It might be worth a little comment that the docs say `661` here (AIR 1.0), but it's not super important, and definitely not a blocker.

core/src/avm2/globals/flash/display/NativeMenu.as
```diff
@@ -1,6 +1,141 @@
-package flash.display {
+package flash.display
+{
+
     import flash.events.EventDispatcher;
-    public class NativeMenu extends EventDispatcher {
+    import __ruffle__.stub_method;
+    import __ruffle__.stub_getter;
+    
+    [API("661")]
```
torokati44, line 8: `airglobal.swc` disagrees with the docs yet again - it seems like this is also technically versioned at `667`.

core/src/avm2/globals/flash/display/NativeMenu.as
```diff
@@ -1,6 +1,141 @@
...
+            return null;
+        }
+
+        // Moves a menu item to the specified position.
+        public function setItemIndex(item:NativeMenuItem, index:int):void
+        {
+            stub_method("flash.display.NativeMenu", "setItemIndex");
+        }
+
+        [API("668")]
```
torokati44, line 124: Yet another `airglobal.swc` - docs mismatch: This version gate is only in the docs, not in the `swc`... :/

core/src/avm2/globals/flash/display/NativeMenu.as
```diff
@@ -1,6 +1,141 @@
...
+    {
+
+        // Indicates whether any form of native menu is supported on the client system.
+        private var _isSupported:Boolean;
+
+        // The array of NativeMenuItem objects in this menu.
+        public var items:Array;
+
+        // The number of NativeMenuItem objects in this menu.
+        private var _numItems:int;
```
torokati44, line 19: I feel like instead of a separate field, the actual length of `items` could be returned from `get numItems`. That way, there's no possibility of inconsistency. And this member can't be set in any way anyway.

core/src/avm2/globals/flash/display/NativeMenu.as
Fancy2209: If this is 667, then NativeMenuItem gotta be too, no?

core/src/avm2/globals/flash/display/NativeMenu.as
```diff
@@ -1,6 +1,144 @@
...
+package flash.display
+{
+
     import flash.events.EventDispatcher;
-    public class NativeMenu extends EventDispatcher {
+    import __ruffle__.stub_method;
+    import __ruffle__.stub_getter;
+    
+    // According to the documentation, it should be [API("661")]
+    // but airglobal.swf disagrees with that:
```
torokati44, line 9: nit: It's `airglobal.swc`, not `swf`. :sweat_smile:

core/src/avm2/globals/flash/display/NativeMenu.as
Mesteery: Yes, this is what I did

core/src/avm2/globals/flash/display/NativeMenu.as
```diff
@@ -1,6 +1,144 @@
...
+        }
+
+        // Moves a menu item to the specified position.
+        public function setItemIndex(item:NativeMenuItem, index:int):void
+        {
+            stub_method("flash.display.NativeMenu", "setItemIndex");
+        }
+
+        // According to the documentation, it should be [API("668")]
+        // but there is no version gate in airglobal.swf
```
torokati44, line 127: `swc` here as well

core/src/avm2/globals/flash/display/NativeMenuItem.as
torokati44: (and `swc` here too)

core/src/avm2/globals/flash/display/NativeMenu.as
Mesteery: oops

core/src/avm2/globals/flash/display/NativeWindow.as
```diff
@@ -0,0 +1,314 @@
...
+      return false;
+    }
+
+    public function get supportsNotification():Boolean
+    {
+      stub_getter("flash.display.NativeWindow", "supportsNotification");
+      return false;
+    }
+
+    [API("661")]
```
torokati44, line 307: nit: Since the entire class is versioned at this level, there's no point in adding the same limit to any member.

core/src/avm2/globals/flash/display/NativeWindow.as
```diff
@@ -0,0 +1,314 @@
...
+    {
+      stub_getter("flash.display.NativeWindow", "bounds");
+      return _bounds;
+    }
+
+    public function set bounds(value:Rectangle):void
+    {
+      stub_setter("flash.display.NativeWindow", "bounds");
+      _bounds.height = height = value.height;
+      _bounds.width = width = value.width;
```
torokati44, line 112: Why set `height` and `width`, but not `x` and `y`?

Also, I feel like invoking other setters (and getters too) of other properties with which there is this tight coupling, might not be the best idea, as it can easily lead to infinite (indirect) recursion.

And, maybe `set x`, `set y`, `set width`, and `set height` should also affect the bounds, to keep things consistent - but again, perhaps directly via `_bounds` to avoid infinite property accessor recursion.

Mesteery: What do you recommend to do instead?

Mesteery: > Setting the bounds property of a window is equivalent to setting its x, y, width, and height properties. Likewise, setting any of the individual dimensions is equivalent to setting the bounds property. When you set all the dimensions at the same time by using the bounds property, **fewer events are dispatched.**

Should we dispatch these events too, and to which events the docs refer?

torokati44: I'd say `set bounds` should simply set `_bounds` - but probably still copying all four values within individually, to not "steal" the `Rectangle` object reference; and not touch any other setters of `this`.
And the 4 individual setters should also go directly into the appropriate member of `_bounds`.
Finally, the setters of `width`, `height`, and `bounds` could also set the size of the stage.
Perhaps this way, everything will be consistent, and there's no risk of infinite recursion.

torokati44: > Should we dispatch these events too, and to which events the docs refer?

I don't have any strong opinions: Since these are just stubs, I don't think event dispatching in general is that important yet, as nothing will work anyway.

torokati44 (approved):
I think this is all from me. Thank you for dealing with all my nitpicking!
I'll squash-merge later today unless anyone else objects, or finds something else.

core/src/avm2/globals/flash/desktop/NativeApplication.as
```diff
@@ -0,0 +1,264 @@
...
+
+    private var _idleThreshold:int = 300;
+
+    public function NativeApplication()
+    {
+      super();
+      setTimeout(function():void
+      {
+        dispatchEvent(new InvokeEvent(InvokeEvent.INVOKE, false, false, null, []));
+      }, 500);
```
adrian17, line 30: I'm assuming this is fake behavior - can you add some comment here?

core/src/avm2/globals/flash/display/NativeMenuItem.as
```diff
@@ -1,6 +1,33 @@
 package flash.display {
     import flash.events.EventDispatcher;
+    import __ruffle__.stub_constructor;
+    import __ruffle__.stub_getter;
+    // According to the documentation, it should be [API("661")]
+    // but airglobal.swc disagrees with that:
+    [API("667")]
     public class NativeMenuItem extends EventDispatcher {
```
adrian17, line 8: in FP, ContextMenu(Item) derives from NativeMenu(Item). Do we not want to do it here, or would it just complicate the current code too much?

core/src/avm2/globals/flash/desktop/NativeApplication.as
Mesteery: with #17609, `[]` will be replaced by the real arguments passed to the application, but it stay fake behaviour I think

core/src/avm2/globals/flash/display/NativeMenuItem.as
torokati44: I have a confuse.
We already have `class ContextMenuItem extends NativeMenuItem` and `class ContextMenu extends NativeMenu`. Do you mean that this version gate could cause problems for there?

core/src/avm2/globals/flash/desktop/NativeApplication.as
Mesteery: I don't know when INVOKE is triggered in AIR so I put a delay in case, so there are no issues.

Lord-McSweeney: > InvokeEvent objects are dispatched by the NativeApplication object (NativeApplication.nativeApplication). **To receive invoke events, call the addEventListener() method of the NativeApplication object. When an event listener registers for an invoke event, it will also receive all invoke events that occurred before the registration.**

And in airglobal I can see an `override public function addEventListener` for NativeApplication. If Dofus doesn't need this, can we remove it (or at least add a TODO comment) so we can properly implement it later?

core/src/avm2/globals/flash/display/NativeMenuItem.as
Lord-McSweeney: It shouldn't, playerglobal gets the highest versions

core/src/avm2/globals/flash/display/Stage.as
```diff
@@ -317,5 +326,13 @@ package flash.display {
         public function setAspectRatio(newAspectRatio:String):void {
             stub_method("flash.display.Stage", "setAspectRatio");
         }
+        
+        [API("661")]
+        public function get nativeWindow():NativeWindow {
+            if (!this._nativeWindow) {
```
Lord-McSweeney, line 332: This should probably do a stub warning

core/src/avm2/globals/flash/desktop/NativeApplication.as
```diff
@@ -0,0 +1,265 @@
...
+  import __ruffle__.stub_method;
+  import __ruffle__.stub_getter;
+  import __ruffle__.stub_setter;
+
+  [API("661")]
+  public final class NativeApplication extends EventDispatcher
+  {
+    private static var _instance:NativeApplication;
+
+    public var _openedWindows:Array = [];
```
Lord-McSweeney, line 20: Is there any way we could make sure that this isn't `public`?

torokati44:
Thank you for addressing all the review notes!

PR #17321, opened 2024-07-31
avm2: Switch all Object variants to use Gc instead of GcCell

Lord-McSweeney:


core/src/avm2/object/event_object.rs
```diff
@@ -343,45 +347,36 @@ impl<'gc> EventObject<'gc> {
...
-                .finish(),
-            Err(err) => f
-                .debug_struct("EventObject")
-                .field("type", &err)
-                .field("class", &err)
-                .field("ptr", &self.0.as_ptr())
-                .finish(),
-        }
+        f.debug_struct("EventObject")
+            .field("type", &self.0.event.borrow().event_type())
```
adrian17, line 377: technically the borrow is still fallible, can we make it a try_borrow like before?

core/src/avm2/object/xml_list_object.rs
```diff
@@ -67,62 +71,62 @@ impl<'gc> XmlListObject<'gc> {
...
     }
 
     pub fn target_object(&self) -> Option<XmlOrXmlListObject<'gc>> {
-        self.0.read().target_object
+        self.0.target_object.get()
     }
 
     pub fn target_property(&self) -> Option<Multiname<'gc>> {
-        self.0.read().target_property.clone()
+        self.0.target_property.borrow().clone()
```
adrian17, line 129: I think `target_property` can be a `Lock` too, if you `clone` it here?

Lord-McSweeney: Can't clone because `Cell<T>` only implements Clone for T: Copy

core/src/avm2/object/netstream_object.rs
```diff
@@ -31,54 +33,45 @@ pub fn netstream_allocator<'gc>(
...
-                .field("class", &obj.base.debug_class_name())
-                .field("ptr", &self.0.as_ptr())
-                .finish(),
-            Err(err) => f
-                .debug_struct("NetStreamObject")
-                .field("class", &err)
-                .field("ptr", &self.0.as_ptr())
-                .finish(),
-        }
+        f.debug_struct("NetStreamObject")
```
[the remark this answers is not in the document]
Lord-McSweeney: The type is practically always going to be `NetStream` though

PR #17324, opened 2024-08-01
text: Fix calculating line bounds

kjarosh:
This fixes a small regression introduced by https://github.com/ruffle-rs/ruffle/pull/17312. In some cases, having bullets or underlines in the text will negatively influence `screen_position_to_index` calculations due to their dummy bounds, and cause the result to be more off than before.

Unfortunately writing a test for that is currently very difficult, as this method (and generally laying out multiline text) is inaccurate.


torokati44 (approved):
You probably know what you're doing... ;)

PR #11227, opened 2023-05-26
Add freedesktop compliant appstream & metainfo

Doomsdayrs:
Validated to match requirements set by freedesktop via `desktop-file-check` and `flatpak run org.freedesktop.appstream-glib validate rs.ruffle.Ruffle.metainfo.xml`.

---

Related: https://github.com/ruffle-rs/ruffle/issues/9220

PR #17335, opened 2024-08-02
desktop: Add a windows .msi installer

Dinnerbone:
this was awful. xml and barely documented schemas... 😭

But this makes an msi as part of the nightly release, which installs ruffle and *hints* at a file association with swfs.

(It's a hint because windows will likely choose to ignore it if the user already has an association - best case it'll ask the user what they want as their default, the next time they open a swf)

I haven't actually tested the action changes on github, just locally 😓 Hard to test.


Dinnerbone:
[ruffle.zip](https://github.com/user-attachments/files/16463976/ruffle.zip)
This is the current master, if anyone wants to try it out.

danielhjacobs:
As discussed on Discord, this installer should probably add a Ruffle shortcut. It can be added to either or both of the Desktop and the Start Menu, but I expect most people will want it to be on the Desktop.

.github/workflows/release_nightly.yml
```diff
@@ -129,6 +136,15 @@ jobs:
           cp README.md package/README.md
           cp LICENSE.md package/LICENSE.md
 
+      - name: Package MSI
+        run: |
+          cd desktop/wix
```
kjarosh, line 141: IMO we should use `desktop/packages` directory for that. There's already macos-related files, and after https://github.com/ruffle-rs/ruffle/pull/11227 there will be Linux-related metadata, so it makes sense to have Windows-installer-related things there too: `desktop/packages/windows/wix`, or `desktop/packages/wix`, or `desktop/packages/windows`.

.github/workflows/release_nightly.yml
```diff
@@ -129,6 +136,15 @@ jobs:
...
+          wix build ruffle.wxs -ext WixToolset.UI.wixext -ext WixToolset.Util.wixext -arch x64 -o ../../package/setup.msi -pdbtype none
+        env:
+          RUFFLE_VERSION: "0.1.0"
```
kjarosh, line 144: Can we use version from cargo here? (see https://github.com/ruffle-rs/ruffle/pull/17193/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R93)

desktop/assets/ruffle_desktop.rc
```diff
@@ -1,4 +1,5 @@
 // winres.h cannot be included, so define manually
+#define VS_VERSION_INFO 1
```
kjarosh, line 2: I'm wondering about these version-related strings in this file (`VER_FILEVERSION`, `VER_PRODUCTVERSION`). Are they important? If they are, we should probably set them dynamically based on the version from cargo somehow?

Croworbit:
I don't really like stuff on my desktop, I assume the shortcut is optional?

desktop/assets/ruffle_desktop.rc
Dinnerbone: Yes we should! They're not critical by any means, but it's good practice to set them. It's the windows equivalent of tools giving `--version` as a standard CLI argument.

.github/workflows/release_nightly.yml
Dinnerbone: Good idea, yes we can!

But let's do that after your PR, since we're not using `cargo get` yet. I left it here to easily show future-us how to hook it up.

.github/workflows/release_nightly.yml
Dinnerbone: Sure, let's go with `packages/windows/wix`.

The macos files that are there currently aren't strictly related to distribution, it's just "the other half of the app" in mac terms, but... I'm fine with usurping the directory. :D

Dinnerbone:
Updated to add shortcuts, remove EULA screen, added a readme for how to build, and moved to `desktop/packages/windows/wix`.
[ruffle.zip](https://github.com/user-attachments/files/16475601/ruffle.zip)

> I don't really like stuff on my desktop, I assume the shortcut is optional?

Unfortunately it's really hard to make optional... :( Can delete it immediately, it won't recreate it.

Croworbit:
just for the record, that stinks.
but depending on the installer creation tool you're using, its probably painful and old, so fair enough

would it reappear at any time during an update, or is it only on first install?

desktop/assets/ruffle_desktop.rc
kjarosh: I guess we could set them dynamically in `build.rs` then. It seems that we can use the `macros` parameter for that, however I'm unsure how that works and I wasn't able to find any examples. We should also consider parameterizing `VER_DEBUG` and possibly the year.

desktop/packages/windows/wix/licence.rtf
```diff
@@ -0,0 +1,9703 @@
+{\rtf1\ansi\deff0{\fonttbl{\f0 \fswiss Helvetica;}{\f1 \fmodern Courier;}}
```
kjarosh, line 1: I think you forgot to remove that file? :sweat_smile:

Dinnerbone: 😭

desktop/packages/windows/wix/ruffle.wxs
```diff
@@ -0,0 +1,182 @@
...
+                    <Verb Id='open' Command='Open' TargetFile='DesktopEXE' Argument='"%1"'/>
+                </Extension>
+            </ProgId>
+            <ProgId Id='Ruffle.spl' Description='Flash Movie' Icon="DesktopEXE" IconIndex="1">
+                <Extension Id='spl' ContentType='application/x-shockwave-flash'>
+                    <Verb Id='open' Command='Open' TargetFile='DesktopEXE' Argument='"%1"'/>
+                </Extension>
+            </ProgId>
+            <ProgId Id='Ruffle.ruf' Description='Ruffle Bundle' Icon="DesktopEXE" IconIndex="1">
+                <Extension Id='ruf' ContentType='application/x-ruffle-bundle'>
```
kjarosh, line 105: I proposed one day `application/x.ruffle-bundle+zip`: according to my knowledge `x-` is obsolete, and `+zip` signifies that the format is based on a zip file.

desktop/packages/windows/wix/ruffle.wxs
```diff
@@ -0,0 +1,182 @@
...
+                <RegistryValue Key="SupportedTypes" Name=".swf" Value="" Type="string"/>
+                <RegistryValue Key="SupportedTypes" Name=".spl" Value="" Type="string"/>
+                <RegistryValue Key="SupportedTypes" Name=".ruf" Value="" Type="string"/>
+            </RegistryKey>
+        </Component>
+
+        <Feature
+                Id='Binaries'
+                Title='Application'
+                Description='Installs the Ruffle desktop player'
```
kjarosh, line 139: Should it be "player" here? Maybe just "application"? I thought that it was decided that Ruffle should be referred to as an emulator :sweat_smile:

kjarosh (approved):
LGTM, although note that I'm not able to test this and I do not have extensive knowledge about Windows applications

desktop/packages/windows/wix/ruffle.wxs
Dinnerbone: I tried to find your proposal but discord is awful to search, couldn't find it! I'll update to your suggestion

(I don't think this actually affects anything though, I have no idea why it's needed...)

Dinnerbone:
Made the desktop shortcut optional, and fixed the mime type & feature description

Croworbit:
> Made the desktop shortcut optional, and fixed the mime type & feature description

I appreciate this
sorry about the whining

PR #17340, opened 2024-08-02
core: Add missing At sign key codes for Android Apps

acelyakilic:
@ sign for swf files that require a mail account for the Android app

https://github.com/ruffle-rs/ruffle-android/pull/220

### DIFF
--- a/core/src/events.rs
+++ b/core/src/events.rs
@@ -505,6 +505,7 @@ pub enum KeyCode {
     Key7 = 55,
     Key8 = 56,
     Key9 = 57,
+    At = 64,
     A = 65,
     B = 66,
     C = 67,


torokati44:
In this repo, we use tag prefixes for commit messages, and I'm not sure how specific this is to Android in the end, just because it came up first in that context, so maybe this would be more fitting: `core: Add missing At sign (@) key code` - but don't worry, this can be fixed by the maintainers as well, by "squash merging".

acelyakilic:
Thank you for the information. Added to the title.

torokati44:
(The title of the PR is not necessarily the same thing as the commit message(s) contained within - but again, don't worry about it... :) )

torokati44:
Where did you get this numerical value from?
I can't find it in here: https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/ui/Keyboard.html

acelyakilic:
I found that the ascii equivalent of the horse sign is 64.

I didn't know that the flash player itself has a keyboard feature. It's up to you to decide whether Rufle itself has the At sign or not.

But I think the android app should have it on the keyboard.

kjarosh:
Unfortunately, I don't think that's right. FP's KeyCodes have little to do with ASCII:

- `;` has code 186, but it's 59 in ASCII,
- `,` has code 188, but it's 44 in ASCII,
- 97 is `a` in ASCII, but it's a code of `Numpad 1`,
- 33 is `!` in ASCII, but it's a code of `PageUp`.

The only keys that match ASCII codes are uppercase letters and non-numepad numbers (probably to make calculations easier in FP). Other character codes that match their respective "keys" are non-ASCII Unicode characters (we do not support them yet, but they are not required for text input).

The idea behind key codes is to identify a key on the keyboard, not a character. FP uses US QWERTY to map characters to keys, and US QWERTY does not have an <kbd>@</kbd> key. Instead, the character `@` is inputted using <kbd>Shift</kbd> + <kbd>2</kbd>, so it should use key codes 16 (for Shift) and 50 (for Number 2).

Do you have any reason other than the ASCII code to believe key code 64 is an `@`? If you want to make sure that pressing `@` on Android produces a proper key code, I think it should produce 16 and 50 (`Shift` and `Key2` respectively).

The following PR shows how it's been done in the desktop application:

- https://github.com/ruffle-rs/ruffle/pull/16868 

The truth is, we currently have to duplicate the mapping code in all frontends of Ruffle. We will have to make a bigger refactor of input-related code to address that.

torokati44:
Yes, I have since come to the same realization that `@` doesn't have it's own keycode, as it doesn't have a physical key of its own, even on English layout keyboard.

Yeah, we could copy that linked PR in the Android app, or even try to just emit a text entry event without any keypress events first, that might work for text fields.

My bad, sorry for the inconvenience on all parties.

PR #17343, opened 2024-08-02
avm2: Remove some dynamic-dispatch TObject logic

Lord-McSweeney:
All Object variants must now implement a method `gc_base`, which returns a `Gc` around `RefLock<ScriptObjectData>` (which will later be changed to just `ScriptObjectData`). Additionally, `TObject` methods that aren't overriden by the object variants are marked as `#[no_dynamic]`, which prevents them getting duplicated by `enum_trait_object`.

### DIFF
--- a/core/macros/src/lib.rs
+++ b/core/macros/src/lib.rs
@@ -3,8 +3,8 @@ extern crate proc_macro;
 use proc_macro::TokenStream;
 use quote::quote;
 use syn::{
-    parse_macro_input, parse_quote, FnArg, ImplItem, ImplItemFn, ItemEnum, ItemTrait, Pat,
-    TraitItem, Visibility,
+    parse_macro_input, parse_quote, FnArg, ImplItem, ImplItemFn, ItemEnum, ItemTrait, Meta, Pat,
+    TraitItem, TraitItemFn, Visibility,
 };
 
 /// `enum_trait_object` will define an enum whose variants each implement a trait.
@@ -63,9 +63,26 @@ pub fn enum_trait_object(args: TokenStream, item: TokenStream) -> TokenStream {
     let trait_methods: Vec<_> = input_trait
         .items
         .iter()
-        .map(|item| match item {
+        .filter_map(|item| match item {
             TraitItem::Fn(method) => {
                 let method_name = &method.sig.ident;
+
+                let is_no_dynamic = method.attrs.iter().any(|attr| match &attr.meta {
+                    Meta::Path(path) => {
+                        if let Some(ident) = path.get_ident() {
+                            ident == "no_dynamic"
+                        } else {
+                            false
+                        }
+                    }
+                    _ => false,
+                });
+
+                if is_no_dynamic {
+                    // Don't create this method as a dynamic-dispatch method
+                    return None;
+                }
+
                 let params: Vec<_> = method
                     .sig
                     .inputs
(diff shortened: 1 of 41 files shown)


core/macros/src/lib.rs
```diff
@@ -148,3 +166,14 @@ pub fn enum_trait_object(args: TokenStream, item: TokenStream) -> TokenStream {
 
     out.into()
 }
+
+#[proc_macro_attribute]
+pub fn no_dynamic(_args: TokenStream, item: TokenStream) -> TokenStream {
```
adrian17, line 171: as moulins proposed, this is probably not needed if you just strip the attribute in proc-macro

core/src/avm2/object.rs
```diff
@@ -153,6 +153,17 @@ pub use crate::avm2::object::xml_list_object::{
 pub use crate::avm2::object::xml_object::{xml_allocator, XmlObject, XmlObjectWeak};
 use crate::font::Font;
 
+fn borrow_data<'gc>(gc: Gc<'gc, RefLock<ScriptObjectData<'gc>>>) -> Ref<ScriptObjectData<'gc>> {
```
adrian17, line 156: why are these needed?

core/src/avm2/object/bitmapdata_object.rs
```diff
@@ -62,6 +61,11 @@ pub struct BitmapDataObjectData<'gc> {
     bitmap_data: Lock<Option<BitmapDataWrapper<'gc>>>,
 }
 
+const _: () = assert!(std::mem::offset_of!(BitmapDataObjectData, base) == 0);
```
adrian17, line 64: can these be moved to inside the method?
(Maybe even generated by a macro, to hide the repeated `unsafe`? ;D )

core/src/avm2/object.rs
Lord-McSweeney: I'm getting bad codegen when I remove these, though I'm not sure why, they're getting inlined. These are going to disappear as soon as ScriptObjectData gets interior mutability, though, so I don't think it's too big of a problem.

core/src/avm2/object/bitmapdata_object.rs
Lord-McSweeney: I think it makes more sense to have them here, as they're referencing the struct immediately above, and the TObject impl isn't right after the `struct AbcObjectData` definition for several object variants.

adrian17: My motivation for this was that them being directly above an `unsafe` block explains its safety better, but I'm not hell bent on this :)

core/macros/src/lib.rs
```diff
@@ -62,10 +62,44 @@ pub fn enum_trait_object(args: TokenStream, item: TokenStream) -> TokenStream {
...
                 let method_name = &method.sig.ident;
+
+                let mut is_no_dynamic = false;
+
+                let new_attrs = method
+                    .attrs
+                    .iter()
+                    .filter(|attr| match &attr.meta {
+                        Meta::Path(path) => {
+                            if let Some(ident) = path.get_ident() {
```
adrian17, line 77: `if path.is_ident("no_dynamic")`?

core/macros/src/lib.rs
```diff
@@ -62,10 +62,44 @@ pub fn enum_trait_object(args: TokenStream, item: TokenStream) -> TokenStream {
...
+                                    true
+                                }
+                            } else {
+                                true
+                            }
+                        }
+                        _ => true,
+                    })
+                    .map(|o| o.clone())
+                    .collect::<Vec<_>>();
```
adrian17, line 94: instead of copying and reassigning, can we call `method.attrs.retain(|attr| ...)`?

adrian17 (approved):
Approved, assuming notes are resolved.
I just wish there was a nicer way to wrap/hide/dedupe these `unsafe`s :/

I still think about smth like
`impl_base_cast!(ClassObjectData, ScriptObjectData, base)` that adds both the method and asserts in one go. (but we can leave that for the future)

PR #17346, opened 2024-08-02
web: Fix the config negotation between self-hosted and the extension

danielhjacobs:


web/packages/core/src/ruffle-player.tsx
```diff
@@ -966,6 +956,36 @@ export class RufflePlayer extends HTMLElement {
...
+                    ? {
+                        allowScriptAccess: parseAllowScriptAccess(
+                            "samedomain",
+                            this.options.url,
+                        )!,
+                    }
+                : {}),
+                ...(window.RufflePlayer?.config ?? {}),
+                // This will make extension config take precedence, if I move this one line higher self-hosted config will take precedence
+                // Let the heated discussions begin
```
danielhjacobs, line 979: Comment is here to facilitate discussion. Start discussing below. This has come up before. Site owners have come out against it, mainly arguing against showSwfDownload being able to be changed by end-users. However, it is true that `window.open(document.querySelector("ruffle-player, ruffle-embed, ruffle-object").swfUrl, "_self")` can already be used, making this just a convenience method.

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -858,6 +848,32 @@ export class InnerPlayer {
...
+                    ? {
+                          allowScriptAccess: parseAllowScriptAccess(
+                              "samedomain",
+                              this.options.url,
+                          )!,
+                      }
+                    : {}),
+                ...(window.RufflePlayer?.config ?? {}),
+                // This will make extension config take precedence, if I move this one line higher self-hosted config will take precedence
+                // Let the discussions begin
```
danielhjacobs, line 867: Comment is here to facilitate discussion. Start discussing below. This has come up before. Site owners have come out against it, mainly arguing against `showSwfDownload` being able to be changed by end-users. However, it is true that end users can usually just check the network tab, making this just a convenience method.

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -806,26 +809,13 @@ export class InnerPlayer {
...
-            };
+            this.options = options;
+            this.isPolyfillElement = isPolyfillElement;
 
             // Pre-emptively set background color of container while Ruffle/SWF loads.
             if (
-                this.loadedConfig.backgroundColor &&
-                this.loadedConfig.wmode !== WindowMode.Transparent
+                this.loadedConfig?.backgroundColor &&
+                this.loadedConfig?.wmode !== WindowMode.Transparent
```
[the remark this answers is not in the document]
danielhjacobs: See the new getter.

web/packages/extension/src/content.ts
```diff
@@ -198,14 +198,18 @@ function isXMLDocument(): boolean {
...
         }
     });
-
+    const autoStartOptions = options.autostart
+        ? {
+              autoplay: "on",
+              unmuteOverlay: "hidden",
+              splashScreen: false,
+          }
+        : {};
```
danielhjacobs, line 207: The idea here, much like `explicitOptions`, is to only set these options if autostart is not the default (false), so if a self-hosted config sets autoplay to "on", for example, it's not ignored when the extension is used.

danielhjacobs: This is not important if the self-hosted config takes precedence over the extension toggles as is currently the case.

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -858,6 +849,32 @@ export class InnerPlayer {
         }
     }
 
+    // The effective config loaded upon `load`, possibly changing if `window.RufflePlayer` changes later.
```
danielhjacobs, line 852: `loadedConfig` needs to be able to change after `load` in the event that `load` is run either before the extension loads or before a self-hosted script that sets `window.RufflePlayer.config` runs.

web/packages/core/src/public-api.ts
```diff
@@ -23,6 +23,7 @@ declare global {
  */
 export interface PublicAPILike {
     config?: DataLoadOptions | URLLoadOptions | object;
+    extensionConfig?: DataLoadOptions | URLLoadOptions | object;
```
danielhjacobs, line 26: Two separate configs, one for the extension and the other for self-hosted, ensures the values of one can not override the values of the other. We can ensure the values in the extension don't override the values of self-hosted without a separate config, as we control the way the values in the extension config are set. However, we can not do the same in reverse, as we can't dictate how self-hosted config values are set.

web/packages/extension/src/ruffle.ts
```diff
@@ -7,12 +7,8 @@ function handleMessage(message: Message) {
             if (window.RufflePlayer === undefined) {
                 window.RufflePlayer = {};
             }
-            if (window.RufflePlayer.config === undefined) {
-                window.RufflePlayer.config = {};
-            }
-            window.RufflePlayer.config = {
```
danielhjacobs, line 13: `window.RufflePlayer.config` is now explicitly the self-hosted config.

Korne127 (approved):

Dinnerbone:
I've thought about this before, and never added it because it just didn't feel right. Today looking at it in code, I'm still a little hesitant.

I *think* I'd prefer something like `ruffle.configs` as an array of configs + a new field "priority", where the existing `config` is an implicit entry at prio 0. We build `loadedConfig` by iterating the configs and merging them - sorted by prio ascending. 

We can then in the extension have some things that don't override the websites wishes (insert with low prio), and some that do (insert with high prio). Log level absolutely **should** change the log level, but "download swf" is the only debatable one afaik.

Websites like the wayback machine can then also use this to better give their preferences without a website replacing their wishes. Assuming, of course, a website doesn't just blindly go `ruffle.configs = []`. We should only document it in the tsdocs and give proper usage guidelines.
(Ofc the same issue is present with this pr - someone could just `ruffle.extensionConfig = {}` too.)

web/packages/core/src/internal/player/inner.tsx
n0samu: The extension config taking precedence is fine with me 👍

torokati44:
>  but "download swf" is the only debatable one afaik.

The browser is still the **user** agent. You host it publicly, I can access it. If I can access it to play it, my browser will store it anyway in it's cache, so if I want to, I will keep it to myself. No need/point to make it unnecessary complicated by me having to open the network tab in the developer tools or whatever...

PR #17347, opened 2024-08-02
ci: Fix Edge based tests

danielhjacobs:
Supersedes #17255

### DIFF
--- a/.github/workflows/test_web.yml
+++ b/.github/workflows/test_web.yml
@@ -92,11 +92,16 @@ jobs:
         working-directory: web
         run: npm test
 
+      - name: Set EDGEDRIVER_VERSION environment variable
+        if: runner.os == 'Linux'
+        run: |
+          echo "EDGEDRIVER_VERSION=$( $EDGEWEBDRIVER/msedgedriver --version | cut -f4 -d ' ' )" >> $GITHUB_ENV
+
       - name: Run browser-based tests
         # Don't run browser tests on Windows because it's flaky for unknown reasons. :-(
         if: runner.os == 'Linux'
         working-directory: web
-        run: npm run wdio -- --headless --chrome --firefox
+        run: npm run wdio -- --headless --chrome --firefox --edge
 
   check-required:
     needs: changes


torokati44:
Yay! But, just curious - where did this come from? Was it recommended by some wdio/GHA/Edge folks? Or have you just found it on your own.
Also, less directly related: I was wondering whether it would be beneficial to split the 3 browsers into 3 separate "steps", so it's easier to tell exactly which one has a headache in any given failed run (to be honest, I have never even attempted to parse anything useful out of the massive flood of log that these tests spew out - maybe that could be improved even?).

danielhjacobs:
> where did this come from? Was it recommended by some wdio/GHA/Edge folks? Or have you just found it on your own? (Not that it's really important...)

As I noticed and noted in https://github.com/actions/runner-images/issues/10380, https://github.com/webdriverio-community/node-edgedriver/blob/7b2b44946f422b86e4c937d277f412a380394a1c/src/install.ts#L24 relies on the `EDGEDRIVER_VERSION` enviroment variable.

https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#environment-variables-1 says `EDGEWEBDRIVER` is defined as `/usr/local/share/edge_driver`. I ran `echo "$($EDGEWEBDRIVER --version)"` in <https://github.com/danielhjacobs/ruffle/actions/runs/10220607522/job/28281327149> and it said "/usr/local/share/edge_driver: Is a directory".

So I then ran `ls $EDGEWEBDRIVER` in <https://github.com/danielhjacobs/ruffle/actions/runs/10220644237/job/28281446113> and it said "Driver_Notes" and "msedgedriver"

So I then ran `$EDGEWEBDRIVER/msedgedriver --version` in <https://github.com/danielhjacobs/ruffle/actions/runs/10220669515/job/28281539537> and it said "Microsoft Edge WebDriver 127.0.2651.72 (efc0f6216bd953a61b23596768d5f3dfbafa0b53)". The 4th word there, splitting on spaces, is "127.0.2651.72". `echo "EDGEDRIVER_VERSION=$( $EDGEWEBDRIVER/msedgedriver --version | cut -f4 -d ' ' )" >> $GITHUB_ENV` sets that to the GitHub environment variable `EDGEDRIVER_VERSION`.

> Also, less directly related: I was wondering whether it would be beneficial to split the 3 browsers into 3 separate "steps", so it's easier to tell exactly which one has a headache in any given failed run

Maybe, but I don't know if the wdio runs on the browsers in parallel. If so, that may be a lot slower.

Dinnerbone (approved):

.github/workflows/test_web.yml
```diff
@@ -92,11 +92,16 @@ jobs:
         working-directory: web
         run: npm test
 
+      - name: Set EDGEDRIVER_VERSION environment variable
```
torokati44, line 95: I think a comment explaining what this is and why it's there and what it's ultimately achieving, would be appreciated - just for future maintainability and bus factor and similar reasons...

torokati44:
Ah: https://github.com/webdriverio-community/node-edgedriver/issues/365#issuecomment-2267380878

torokati44:
Superseded by https://github.com/ruffle-rs/ruffle/pull/17378.

PR #17353, opened 2024-08-03
core: Allow clicking on a non-interactive AVM2 child of a Loader

Aaron1011:


core/src/display_object/loader_display.rs
```diff
@@ -240,6 +243,14 @@ impl<'gc> TInteractiveObject<'gc> for LoaderDisplay<'gc> {
                         return Avm2MousePick::Miss;
                     }
                 }
+            } else if child.as_interactive().is_none()
```
adrian17, line 246: to be sure, isn't this part of the condition already implied by the `else`?

Lord-McSweeney (approved):

PR #17385, opened 2024-08-06
avm1: Add a way to specify numeric `@epsilon` when comparing in `test_method!`

torokati44:
This is supposed to make nightly Rust CI green.

What do you think about the overall idea? I'm open to bikeshedding, as I'm not a macro wiz.


core/src/avm1/test_utils.rs
```diff
@@ -44,7 +44,18 @@ macro_rules! test_method {
...
+                            $(
+                                fn round_to_digits(x: f64) -> f64 {
+                                    let factor = f64::powi(10.0, $digits);
+                                    (x * factor).round() / factor
+                                }
+                                let ret = round_to_digits(ret.coerce_to_f64(activation)?);
+                                let out = round_to_digits(out);
+                            )?
+
+                            assert_eq!(ret, out, "{:?} => {:?} in swf {}", args, $out, version);
```
kjarosh, line 58: IMO `@digits` is a bit over-engineered here and we should just use `@delta` and assert `abs(expected - actual) < delta`, as that's the standard way of comparing FP numbers.

`@digits` compares x digits after the decimal point, but the precision of floating point numbers is not defined for the fractional part, but for the whole number. A better approach would be to compare x significant digits. Also `@digits` uses base 10 instead of base 2, which is kinda arbitrary and unnecessarily affects mantissa, but I'm not sure if it can cause any problems here.

torokati44: > IMO `@digits` is a bit over-engineered here 

Still not as over-engineered as `@transform(|x| (x * 1e12).round() / 1e12)` could have been! :V

> we should just use `@delta`

How about `@epsilon`? IIRC "delta" is more of a domain thing than a value thing (in functional analysis at least).

>  compares x digits after the decimal point, but the precision of floating point numbers is not defined for the fractional part, but for the whole number.

Yeah, fair, I know - this was just a quickly thrown together late night hack.

torokati44: > uses base 10 instead of base 2, which is kinda arbitrary and unnecessarily affects mantissa

Can you suggest a simple and clear way to write a binary exponent `f64` literal?

kjarosh: > How about `@epsilon`? IIRC "delta" is more of a domain thing than a value thing (in functional analysis at least).

Sure, seems alright!

> Can you suggest a simple and clear way to write a binary exponent `f64` literal?

Usually languages support hexadecimal floating point literals for this purpose, but I don't think Rust supports it (seems like it's a deliberate decision, but I don't know the reasoning). Currently you would have to convert the desired fp number to decimal and use that. There are alternatively some crates that can parse hex fp strings, or you could write the value as an expression.

Nonetheless, I was referring to mapping the original value (and potentially losing accuracy due to 5 not being a power of 2). For epsilon that should not matter, so we're safe here.

core/src/avm1/test_utils.rs
```diff
@@ -31,8 +31,9 @@ where
 }
 
 macro_rules! test_method {
-    ( $test: ident, $name: expr, $object: expr, $($versions: expr => { $([$($arg: expr),*] => $out: expr),* }),* ) => {
+    ( $test: ident, $name: expr, $object: expr, $($versions: expr => { $($(@epsilon($epsilon: literal))? [$($arg: expr),*] => $out: expr ),* } ),* ) => {
```
kjarosh, line 34:
```suggestion
    ( $test: ident, $name: expr, $object: expr, $($versions: expr => { $($(@epsilon($epsilon: expr))? [$($arg: expr),*] => $out: expr ),* } ),* ) => {
```

`epsilon` doesn't have to be a literal, someone might use an expression here

torokati44:
Any suggestions for where else to add this annotation? Or we'll do it gradually only as/when needed in the future?

kjarosh:
> Any suggestions for where else to add this annotation? Or we'll do it gradually only as/when needed in the future?

Correct me if I'm wrong, but these tests with `f64::<func>` do not make sense at all in their current form. The only thing they verify is whether `f64::<func>` returns the same value in runtime and as a constant expression. (Folks at rust should be grateful that we're testing such things for them! And when we find a regression in rustc, all our jobs start to fail—this is called a real sacrifice for Rust.)

IMO we should replace all `f64::<func>` in tests with constant values, the same as returned by Flash. And only then we'll be really testing our implementations. In that case, the annotation will have to be used when we deviate from the expected value a bit.

torokati44:
> IMO we should replace all `f64::<func>` in tests with constant values, the same as returned by Flash. And only then we'll be really testing our implementations. In that case, the annotation will have to be used when we deviate from the expected value a bit.

Yeah, I thought about this too, and fair point... My issue is - who says/knows that for all these methods, FP returns a single fixed value for a given parameter on all platforms and configurations and architectures, across all its versions...? :face_with_spiral_eyes:

kjarosh:
> My issue is - who says/knows that for all these methods, FP returns a single fixed value for a given parameter on all platforms and configurations and architectures, across all its versions...?

Probably no one, but that's also something that `@epsilon` should take care of. Just select a platform and update `@epsilon` accordingly (of course as long as it's relatively low, otherwise we've got a bug).

torokati44:
Okay, but if I replace `f64::exp(1.0)` with `std::f64::consts::E`, am I just testing the platform's `exp` implementation? :P Or mathematics itself? :sweat_smile:

kjarosh:
You're testing whether the definition of `std::f64::consts::E` matches the behavior of `f64::exp(1.0)` :smile:

Two questions:
1. How likely is that the definition of `std::f64::consts::E` would change? We know that Rust literals rather won't change. Now we also know that `f64::<func>` may change (as it did in nightly recently).
2. Why use `std::f64::consts::E` when it's actually harder to do so: you have to analyze what value is returned by Flash and that value has to match `std::f64::consts::E` exactly. You can skip step 2 and just use literals.

torokati44:
Should this literalization be done in this PR, or sometime in the future...?

kjarosh:
I'd go with doing it for `f64::exp` in this PR and adding TODOs for all other methods.

kjarosh:
You can replace the following:

1. `f64::exp(1.0)` -> `f64::from_bits(0x4005bf0a8b145769)`
2. `f64::exp(2.0)` -> `f64::from_bits(0x401d8e64b8d4ddae)`

Those are values (V) for which Flash will print `false`, `true`, `false` on my platform:

```as
trace(Math.exp(...) < V);
trace(Math.exp(...) == V);
trace(Math.exp(...) > V);
```

torokati44:
Great, okay! Can't we write them as base10 literals though? 😶

kjarosh:
We could do that, those values would be `2.718281828459045` and `7.38905609893065` respectively. However:
1. we should still preserve the hex values in a comment, as those are the original representations from the SWF,
2. `f64::from_bits` according to docs is extremely portable and its purpose it to preserve bits, so I don't think we should worry here (unless we're using NaNs which we don't),
3. some of these values as literals will throw an `approximate value of ... found` error.

torokati44:
> `f64::from_bits` according to docs is extremely portable and its purpose it to preserve bits, so I don't think we should worry here (unless we're using NaNs which we don't),

Yeah I'm not worried about portability, more about code readability.

> some of these values as literals will throw an `approximate value of ... found` error.

Okay, I think `from_bits` in code, and base10 form in comment for humans should suffice after all.

kjarosh (approved):

kjarosh (approved):

PR #17387, opened 2024-08-06
Managed `AvmStringRepr`s can now point to a `&'static WStr`

moulins:
Allows `&'static WStr`s to be converted into interned `AvmString`s without moving the string data into an heap-allocated buffer.

Slight progress towards getting rid of static `AvmString`s altogether :)

### DIFF
--- a/core/src/string/interner.rs
+++ b/core/src/string/interner.rs
@@ -3,7 +3,7 @@ use std::borrow::{Borrow, Cow};
 use std::cell::Cell;
 use std::hash::{BuildHasher, Hash, Hasher};
 use std::marker::PhantomData;
-use std::ops::DerefMut;
+use std::ops::{Deref, DerefMut};
 
 use gc_arena::{Collect, Gc, GcWeak, Mutation};
 use hashbrown::HashSet;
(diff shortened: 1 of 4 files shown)


core/src/string/interner.rs
```diff
@@ -53,80 +53,99 @@ pub struct AvmStringInterner<'gc> {
     interned: WeakSet<'gc, AvmStringRepr<'gc>>,
 
     empty: Gc<'gc, AvmStringRepr<'gc>>,
-    chars: [Gc<'gc, AvmStringRepr<'gc>>; 128],
+    chars: [Gc<'gc, AvmStringRepr<'gc>>; INTERNED_CHAR_LEN],
 }
 
+const INTERNED_CHAR_LEN: usize = 128;
+static INTERNED_CHARS: [u8; INTERNED_CHAR_LEN] = {
```
Lord-McSweeney, line 60: Can `INTERNED_CHARS` be declared as `const`?

Lord-McSweeney (approved):

core/src/string/interner.rs
moulins: Probably? (not 100% sure, depends on how const-promotion works exactly) I do feel that `static` is closer to the intent of the code here.

torokati44: It does work for me if I change it to `const`. I can amend if you'd like (I have no opinions on it, just offering to do the key pressing and mouse clicking).

torokati44:
Minor nit: This repo uses a different notation for commit message tags (`tag: Message`, not `[tag] Message`).
This can be fixed upon merging too, of course.

core/src/string/interner.rs
moulins: Huh, ok. Still prefer `static` though :D

moulins:
Fixed the commit message, everything should be good now :slightly_smiling_face:

core/src/string/interner.rs
evilpie: I was skeptical at first as well, but the Rust documentation makes it sounds like with `const` we could get the table duplicated every time it's used?

> Static variables, on the other hand, point to a single location in memory, which all accesses share. This means that, unlike with constants, they can’t have destructors, and act as a single value across the entire codebase.

https://doc.rust-lang.org/std/keyword.const.html

torokati44: Not necessarily the entire table, maybe only the elements one by one.

I've seen something similar in `idct.rs` in the `h263-rs` repo, with `BASIS_TABLE`.

torokati44: Those constants are used in fairly tight number crunching loops, and I believe it helped perf a lot to have them inlined.

torokati44: https://github.com/ruffle-rs/h263-rs/blob/f0fa94c366a1d0383df99aa835add175658d6bad/h263/src/decoder/cpu/idct.rs#L36-L65

https://godbolt.org/z/aETTWoncv

torokati44: TBF, in this example, changing `const` to `static` makes no difference whatsoever - I guess the optimizer passes know their stuff. (And this table never has its address taken or anything.)

torokati44: Makes sense. `static` is okay here then.

PR #17393, opened 2024-08-06
Always build LZMA support and remove the feature flag

evilpie:
From what I've heard LZMA support was not enabled by default because it used to require building C/C++ code.

### DIFF
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -84,7 +84,7 @@ jobs:
         if: runner.os != 'macOS'
         # TODO: Disallow retries in general once we can allow only after SIGABRT.
         # See: https://github.com/nextest-rs/nextest/issues/1172
-        run: cargo nextest run --cargo-profile ci --workspace --locked --no-fail-fast --retries 2 -j 4 --features imgtests,lzma,jpegxr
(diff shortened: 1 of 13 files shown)


evilpie:
https://crates.io/crates/swf also has an optional dependency on LZMA, I wasn't sure if we wanted to change that as well.

torokati44:
If there's no actual reason against having the feature flag, and the project builds and runs with it disabled, I see no point in removing it. It's just additional flexibility in case someone might want a smaller build. But I'm not really attached to it either.

evilpie:
All feature flags introduce complexity and LZMA is not an optional part of Actionscript. Personally I just keep forgetting to enable this flag.

torokati44:
Ah, I thought this was only for SWC (compressed SWF) support. (Is that not even LZMA?! 😵‍💫) Pardon my ignorance, never mind then.

n0samu:
SWC (CWS in the file header) is zlib compression. The rare LZMA-compressed SWFs are SWZ (ZWS in the header). But I still don't think this really needs a feature flag. Who wants Ruffle to not be able to open a few random SWFs?

evilpie:
@Dinnerbone also seems to like having it optional. How about adding LZMA to the default feature for ruffle_desktop at least?

PR #17396, opened 2024-08-06
core: Remove some unnecessary code added in #15627

Flawake:


core/src/player.rs
```diff
@@ -735,20 +735,14 @@ impl Player {
...
-                    context,
-                    *context.mouse_position,
-                    HitTestOptions::MOUSE_PICK,
-                ) {
-                    if let Some(Value::Object(obj)) = Some(disp_obj).map(|obj| obj.object()) {
-                        let mut activation = Activation::from_stub(
-                            context.reborrow(),
-                            ActivationIdentifier::root("[ContextMenu]"),
-                        );
+                if let Some(Value::Object(obj)) = Some(disp_obj).map(|obj| obj.object()) {
```
Lord-McSweeney, line 738: Can this line just be
```
                if let Some(Value::Object(obj)) = disp_obj.object() {
```
?

Flawake: Fixed that, also refactored the rest of the function.

adrian17 (approved):

Flawake:
> @Flawake do you want to resolve the conflict?


Currently somewhere between paris and the netherlands. But I will do it in the coming days.

torokati44:
I did a rebase for you.
Here's the range-diff:
image

The only meaningful change was the removal of `.reborrow()` after `context`, and of the first lifetime on `UpdateContext`. Plus some formatting.

PR #17399, opened 2024-08-07
ci: Fix desktop path for wix installer

Dinnerbone:


.github/workflows/release_nightly.yml
```diff
@@ -144,7 +144,7 @@ jobs:
           wix build ruffle.wxs -ext WixToolset.UI.wixext -ext WixToolset.Util.wixext -arch ${{ matrix.MSI_ARCH }} -o ../../../../package/setup.msi -pdbtype none
         env:
           RUFFLE_VERSION: "0.1.0"
-          CARGO_BUILD_DIR: target/${{ matrix.target }}/release
+          CARGO_BUILD_DIR: ../../../../target/${{ matrix.target }}/release
```
kjarosh, line 147: I'd go with setting working directory instead of running `cd` in this case, but that's something I can improve when adding support for stable releases

kjarosh (approved):

Issue #11336, opened 2023-06-01
Ruffle Desktop Gui: Basic file picker bug


waspennator:
### Describe the bug

If you open up the basic file picker through the browse option on the desktop client,  if you leave it open and try to click back onto the ruffle player itself, it will eventually go into a not responding state like this. It also just kinda freezes visually until you close the file picker , but I dunno if that's fixable or not (Moved file manager off screen for proper view)
![Ruffle_Player_2](https://github.com/ruffle-rs/ruffle/assets/78768499/70c070c4-d35f-4963-a498-42de52f0fd76)

It doesn't happen if you do it by clicking on the file tab on the menu bar, then press open, however it presents a new quirk where I can press the maximize, minimize, or exit buttons on the top right and it will initiate them as soon as you close the file picker, meaning you can accidentally close ruffle without warning this way.

### Expected behavior

I know other emulators prevent you from interacting with the gui until you open a file or close out of the file picker, so maybe temporarily prevent the player from being able to interact with the ruffle desktop client until the file manager browser is closed.

### Affected platform

Desktop app

### Operating system

Windows 11

d-air1:
On arch linux the window is frozen on launch and attempting to do anything sends the window into the not responding state.

PR #17403, opened 2024-08-07
desktop: Set main window as parent of file pickers

kjarosh:
Related to https://github.com/ruffle-rs/ruffle/issues/11336 a bit.

This patch makes sure that all file pickers are children of Ruffle. This way, file pickers will hover over the main window and will have to be dismissed before the user may interact with it.

Related functionality added in rfd:
* https://github.com/PolyMeilex/rfd/pull/209

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -251,9 +251,13 @@ dependencies = [
  "futures-channel",
  "futures-util",
  "rand",
+ "raw-window-handle",
  "serde",
  "serde_repr",
  "url",
+ "wayland-backend",
+ "wayland-client",
+ "wayland-protocols",
  "zbus",
 ]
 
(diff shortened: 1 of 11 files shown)


danielhjacobs:
If this were to be merged before a release, you'd need to add PolyMeilex here:

https://github.com/ruffle-rs/ruffle/blob/117df6e3c2fef4f47d0c2bfec97801558aee96e1/deny.toml#L78-L82

kjarosh:
The PR https://github.com/PolyMeilex/rfd/pull/209 has been merged, I think we can now consider switching to a commit from rfd's `master`.

evilpie (approved):
Not an expert on winit, but seems fine considering it compiles.

kjarosh:
Tested on both Wayland and X11, and it works flawlessly, merging

PR #17406, opened 2024-08-07
desktop: Enable LZMA feature by default

evilpie:
This is an alternative to #17393.


torokati44 (approved):
Yeah, since it's using a pure Rust crate (as the `swf` crate has been for years now), it feels weird to not have it on by default.

PR #17409, opened 2024-08-07
video: Fix race condition when downloading openh264. This fixes CI flakiness.

Dinnerbone:
The video backend downloads openh264 by:
- "do we have it already"
- if no:
  - download it to memory
  - extract it to disk
- check md5sum to make sure all is good

But, CI runs in parallel - so multiples of these were running at once. The "do we have it already" for the second instance would return true during the extraction operation of the first, but the md5sum would obviously fail if it hasn't finished extracting.

I work around this by downloading to a temporary file and renaming it in place once extraction has finished; the temp file is adjacent to the target file, so it should be an atomic move. This way - whenever the file does exist, it's guaranteed to exist whole and valid.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4565,6 +4565,7 @@ dependencies = [
  "ruffle_video_software",
  "slotmap",
  "swf",
+ "tempfile",
  "tracing",
 ]
 
--- a/video/external/src/backend.rs
+++ b/video/external/src/backend.rs
@@ -104,10 +104,15 @@ impl ExternalVideoBackend {
         // Regardless of whether the library was already there, or we just downloaded it, let's check the MD5 hash.
         let mut md5 = Md5::new();
         copy(&mut File::open(filepath.clone())?, &mut md5)?;
-        let result: [u8; 16] = md5.finalize().into();
+        let md5digest = md5.finalize();
+        let result: [u8; 16] = md5digest.into();
 
         if result[..] != hex::decode(md5sum)?[..] {
-            return Err(format!("MD5 checksum mismatch for {}", filename).into());
+            let size = filepath.metadata().map(|f| f.len()).unwrap_or_default();
+            return Err(format!(
+                "MD5 checksum mismatch for {filename}; expected {md5sum}, found {md5digest:x} (with a size of {size} bytes)",
+            )
+            .into());
         }
 
         Ok(filepath)
--- a/video/external/src/backend.rs
+++ b/video/external/src/backend.rs
@@ -85,20 +85,24 @@ impl ExternalVideoBackend {
 
         let (filename, md5sum) = Self::get_openh264_data()?;
 
-        let filepath = std::env::current_exe()?
+        let current_exe = std::env::current_exe()?;
+        let directory = current_exe
             .parent()
-            .ok_or("Could not determine Ruffle location.")?
-            .join(filename);
+            .ok_or("Could not determine Ruffle location.")?;
+        let filepath = directory.join(filename);
 
         // If the binary doesn't exist in the expected location, download it.
         if !filepath.is_file() {
             let url = format!("{}{}{}", URL_BASE, filename, URL_SUFFIX);
             let response = reqwest::blocking::get(url)?;
-            let bytes = response.bytes()?;
-            let mut bzip2_reader = BzDecoder::new(bytes.as_ref());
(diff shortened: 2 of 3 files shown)


torokati44:
Oh my gosh, how did you figure this out?! 😱 I'm so sorry. 🥺😭 Yeah, this was obviously problematic...

Dinnerbone:
Looking at the code and experience? It's a common mistake, no worries 😅

Cargo.lock
```diff
@@ -2837,7 +2837,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4979f22fdb869068da03c9f7528f8297c6fd2606bc3a4affe42e6a823fdb8da4"
 dependencies = [
  "cfg-if",
- "windows-targets 0.52.5",
+ "windows-targets 0.48.5",
```
torokati44, line 2840: lol

torokati44:
Just throwing this in here: https://docs.rs/tempfile/3.12.0/tempfile/struct.NamedTempFile.html#method.new_in

Not sure if it'd be better, but I dunno...

Also, this explains the "flv" test flakyness (or basically for any test that has "with_video" enabled), but what about the PB shader_data one...?

Dinnerbone:
> Just throwing this in here: https://docs.rs/tempfile/3.12.0/tempfile/struct.NamedTempFile.html#method.new_in

I have no real thoughts on hand rolling this vs using a crate, it's such a minor thing /shrug

torokati44:
Dunno, I like this a lil' better... :no_mouth: 
https://github.com/ruffle-rs/ruffle/compare/master...torokati44:ruffle:openh264_failures

It Just looks.... neat IMHO. (Note, I did cheat by using a yesterlatest version of the crate, so the windows dependency versions aren't touched...)

kjarosh:
I also do think using a library for managing temp files is better. 

> [4a5925d](https://github.com/ruffle-rs/ruffle/commit/4a5925de973b8904e361dc9ed0a54f23086322dc)

I think this code lacks removal of the temp file in case of errors (copying errors + persisting errors). When we handle the former error we could also skip downloading the file to memory and stream directly to disk using
```rust
let mut bzip2_reader = BzDecoder::new(response);
```
But the lib is 2MB so that probably does not matter much.

torokati44:
Good point about streaming decompression though! I don't know why `bytes` is stored like that. (Probably a remnant of the time when I was fighting `async`+`tokio`+`reqwest`.)

torokati44:
Updated both commits in my fork:
https://github.com/ruffle-rs/ruffle/commit/258919ff11ca632d56a509234c7e44f9a08f778a
https://github.com/ruffle-rs/ruffle/commit/39aa26f8bf921a74bc16a80b68d1b2a70668c715

kjarosh (approved):

torokati44:
https://github.com/ruffle-rs/ruffle/actions/runs/10316146288/job/28557743902?pr=17409#step:8:3496

What the...?!

torokati44:
```
---- avm1/netstream_play_flv ----
Couldn't get OpenH264: failed to persist temporary file: Access is denied. (os error 5)
```

:sob:

video/external/src/backend.rs
```diff
@@ -85,29 +85,37 @@ impl ExternalVideoBackend {
...
             let response = reqwest::blocking::get(url)?;
-            let bytes = response.bytes()?;
-            let mut bzip2_reader = BzDecoder::new(bytes.as_ref());
+            let mut bzip2_reader = BzDecoder::new(response);
 
-            let mut file = File::create(filepath.clone())?;
-            copy(&mut bzip2_reader, &mut file)?;
+            let mut tempfile = tempfile::NamedTempFile::with_prefix_in(filename, directory)?;
+            copy(&mut bzip2_reader, &mut tempfile)?;
+            // Let's assume that if this fails, it's because another process has already put it there
```
torokati44, line 102: I guess...?

torokati44: (Also see next line. Sounds good, @kjarosh?)

kjarosh: Yup, sounds good. I'd also mention in the comment that the MD5 will be checked nonetheless

torokati44 (approved):

torokati44:
(You can merge at will, I don't want to, since I took over.)

PR #17413, opened 2024-08-07
desktop: Fix Ruffle not responding when picking a file

kjarosh:
Fixes https://github.com/ruffle-rs/ruffle/issues/11336.

This PR fixes the problem of Ruffle not responding when picking a file.

This needs to be tested on Windows before merge, because I removed a Windows-specific workaround hoping that it won't be needed when picking files asynchronously.


torokati44:
So, having an async runtime (tokio) did end up coming in handy after all? (:

torokati44 (approved):
I think @danielhjacobs tested this on Windows, and found no issues without the workaround.

danielhjacobs:
Yeah, the issue doesn't exist without the workaround anymore even without this PR.

PR #17429, opened 2024-08-09
chore: Move `PKGBUILD` to `desktop/packages`

kjarosh:
It's a bit of housekeeping. Other platform-specific files used for packaging are in `desktop/packages`, so `PKGBUILD` should also be there.

No way of testing this, we'll have to check the next nightly whether it worked.


torokati44:
Speaking of AUR, our "official-ish" package on there, published by @relrelb, looks rather derelict: https://aur.archlinux.org/packages/ruffle-nightly-bin#comment-914876 :/

Hence the "unofficial" one: https://aur.archlinux.org/packages/ruffle-bin

torokati44 (approved):
LGTM, thank you!

kjarosh:
> Speaking of AUR, our "official-ish" package on there, published by @relrelb, looks rather derelict: https://aur.archlinux.org/packages/ruffle-nightly-bin#comment-914876 :/
> 
> Hence the "unofficial" one: https://aur.archlinux.org/packages/ruffle-bin

We can wait a while (I don't think this PR is urgent) and I can refresh the AUR package a bit in this PR by the way.

desktop/packages/linux/aur/ruffle-nightly-bin/PKGBUILD
```diff
@@ -8,7 +8,7 @@ url="https://ruffle.rs/"
 license=('Apache' 'MIT')
 depends=(zlib libxcb alsa-lib)
 provides=(ruffle)
-conflicts=(ruffle-git)
+conflicts=(ruffle)
```
kjarosh, line 11: Reading the docs, there should be `ruffle`, but TBH I do not see any difference in behavior, as packages conflict with `/usr/bin/ruffle` anyway.

kjarosh:
So I tried adding .desktop and icon, but that's currently a bit difficult. I would have to clone ruffle sources and take it from there, or download raw files from GitHub, both require passing additional hashes to PKGBUILD from GH actions.

The easiest way would be to package those files along with the Linux executable (similarly to license and readme).

desktop/packages/linux/aur/ruffle-nightly-bin/PKGBUILD
torokati44: Huh, it's weird to see that this package both provides, and conflicts with, `ruffle` - but it makes sense I guess, and the "unofficial" `ruffle-git` package has the same. So these names aren't package names then, but are in some other, more abstract namespace...?

torokati44 (approved):

desktop/packages/linux/aur/ruffle-nightly-bin/PKGBUILD
kjarosh: See https://wiki.archlinux.org/title/PKGBUILD#conflicts

> Note that conflicts are checked against pkgname as well as names specified in the provides array. Hence, if your package provides a foo feature, specifying foo in the conflicts array will cause a conflict between your package and all other packages that contain foo in their provides array (i.e., you do not need to specify all those conflicting package names in your conflicts array).

PR #17434, opened 2024-08-09
core: only add callback to context_menu_item if the callback function is not undefined

### LINKED ISSUE #17400: Context menu item should not be added if the callback function is non existent
### Describe the bug

There are 2 functions that edit the context menu on the movieclip.
Right clicking the movieclip gives the option to remove the context menu item from the movieclip.
Right clicking the root gives the option to add a context menu item to the movieclip.

This item should not be added if it's callback function does not exits, but it should be added if a menu item with the same name already exists. And it should use the callback function of that menu item in that case.

Ruffle does actually add the context menu, even if it's callback function does not exist and there is not already a item with the same name present

### Expected behavior

Ruffle should not add a context menu item if it's callback function does not exist and a item with the same name does not already exist.

### Content Location

[context_menu_issue.zip](https://github.com/user-attachments/files/16523035/context_menu_issue.zip)
These files where made in Macromedia Flash 8.

### Affected platform

Desktop app

### Operating system

Windows11

### Browser

_No response_

### Additional information

_No response_

Flawake:
only add callback to context_menu_item if the callback function is not undefined.
Fixes: #17400


core/src/avm1/globals/context_menu_item.rs
```diff
@@ -20,7 +20,10 @@ pub fn constructor<'gc>(
         .get(0)
         .unwrap_or(&Value::Undefined)
         .coerce_to_string(activation)?;
-    let callback = args.get(1).map(|v| v.coerce_to_object(activation));
+    let callback = args
+        .get(1)
+        .filter(|v| v.type_of() != Value::Undefined.type_of())
```
adrian17, line 25: to be sure, why can't we directly compare with `undefined` here? I haven't seen this pattern of comparing with `Undefined.type_of()` before.

Other than this, the PR looks fine.

adrian17:
Actually no, this doesn't look right to me, sorry :(
```
var item = new ContextMenuItem("a", undefined);
trace(item.caption);
trace(item.hasOwnProperty("onSelect")); // true in FP
trace(item.onSelect); // undefined
```
So the property is being set. Can you please show how to observe the wrong behavior in current Ruffle?

Flawake:
Interesting, did not know that.
This were my findings:
clipOne is a movieClip.
```
var menu:ContextMenu = new ContextMenu();
clipOne.menu = menu

menu.customItems.push(new ContextMenuItem("Remove item", notExist));
menu.customItems.push(new ContextMenuItem("Remove item", undefined));
menu.customItems.push(new ContextMenuItem("Remove item"));
menu.customItems.push(new ContextMenuItem("Remove item", exists));
trace(menu.customItems[0].onSelect); // fp: undefined,         Ruffle: [type Object]     PR: undefined
trace(menu.customItems[1].onSelect); // fp: undefined,         Ruffle: [type Object]     PR: undefined
trace(menu.customItems[2].onSelect); // fp: undefined,         Ruffle:  undefined         PR: undefined
trace(menu.customItems[3].onSelect); // fp: [type Function], Ruffle: [type Function]  PR: [type Function]

function exists():Void {
    trace("you selected: "+item.caption);
    menu.customItems.pop(new ContextMenuItem("Remove item"));
}
```

fp: When right clicking the movlieclip and selecting "Remove item". It removes the item and it is not clickable again.
It also prints: "you selected: undefined"
Ruffle: prints nothing, "Remove item" does not get deleted
Pr: does now behave the same as fp does

BUT:
I actually broke something with this pr:
```
trace(menu.customItems[0].hasOwnProperty("onSelect"));
trace(menu.customItems[1].hasOwnProperty("onSelect"));
trace(menu.customItems[2].hasOwnProperty("onSelect"));
trace(menu.customItems[3].hasOwnProperty("onSelect"));
```
fp: prints all true
Ruffle: only prints [2] as false
pr: prints [0], [1] and [2] as false

Running:

```
menu.customItems[0].onSelect()
menu.customItems[1].onSelect()
menu.customItems[2].onSelect()
menu.customItems[3].onSelect()
```
Does print "removeItemMenuItem: you selected: undefined" once in all versions.

Do you know what fp sets the value for trace(menu.customItems[3].hasOwnProperty("onSelect")); to return true
Since trace(menu.customItems[0].onSelect); is undefined

PR #17440, opened 2024-08-09
avm2: Don't allocate more memory doing regexp replacements, than is actually needed

Dinnerbone:


adrian17 (approved):

core/src/avm2/regexp.rs
```diff
@@ -159,6 +159,10 @@ impl<'gc> RegExp<'gc> {
         text: &AvmString<'gc>,
         m: &regress::Match,
     ) -> WString {
+        if !replacement.contains(b'$') {
+            // Nothing to do if there's no $ replacement symbols
+            return replacement.as_wstr().to_owned();
+        }
```
moulins, line 165: What is the intent here? If the goal is to avoid an allocation, it would be better to put this special case directly in the caller, to be able to return `replacement` directly.

Dinnerbone: Changed it to be a Cow to avoid the alloc here, does that look okay?

adrian17 (approved):

moulins (approved):

PR #17442, opened 2024-08-09
desktop: Reformat description in metainfo

kjarosh:
Some stores do not ignore whitespace and newlines, treating the description as preformatted text (e.g. GNOME Software, Mintinstall).

<image width="400" src="https://github.com/user-attachments/assets/bec7c7d7-91d2-4540-acbf-985c58ed220b"/>

### DIFF
--- a/desktop/packages/linux/rs.ruffle.Ruffle.metainfo.xml
+++ b/desktop/packages/linux/rs.ruffle.Ruffle.metainfo.xml
@@ -11,24 +11,9 @@
     <update_contact>ruffle@ruffle.rs</update_contact>
     <launchable type="desktop-id">rs.ruffle.Ruffle.desktop</launchable>
     <description>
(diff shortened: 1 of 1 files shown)


evilpie (approved):
I don't think I want to know how they are parsing HTML.

kjarosh:
I don't know what's going on here. When I load Ruffle metadata directly to GNOME Software as a preview, it's being parsed correctly. It's just when it gets the data from Flathub this weird parsing happens.

kjarosh:
I've been checking software managers and it seems that on clean installations there's no problem with the current description. I suspect our local builds somehow influenced that.

kjarosh:
Closing as it seems that it's an issue with the store frontend or Flathub. Other apps on Flathub and outside of it also format description the same way as us.

PR #17452, opened 2024-08-09
video: Check OpenH264 library downloads with SHA256 instead of MD5

kjarosh:
Continuation of https://github.com/ruffle-rs/ruffle/pull/17243.

We want to be better than Cisco and give our users more security by replacing MD5 with SHA256 hashes. Calculating a SHA256 hash on a safe network after verifying MD5 is a lot more secure than using HTTP+MD5 everywhere.

Thanks to @Fullmetal5 for noticing it and submitting a patch.

My comment from the commit:

> Cisco distributes the OpenH264 binaries and their MD5 hashes over HTTP.
Knowing that HTTP is insecure, MD5 hashes may easily collide,
and both are served over the same medium, saying that
this method is ridiculous is an understatement.
> 
> Take into account that these are binaries we download and execute, and
we for sure do not want to turn Ruffle into a remote code execution framework.
> 
> This patch changes MD5 to SHA256 in order to increase security
and protect our users from attacks resulting from this vulnerability.

### DIFF
--- a/video/external/Cargo.toml
+++ b/video/external/Cargo.toml
@@ -17,14 +17,8 @@ ruffle_video_software = { path = "../software" }
 
 # Needed for OpenH264:
 libloading = "0.8.5"
-md-5 = "0.10.6"
 reqwest = { version = "0.12.5", default-features = false, features = ["blocking"] }
 hex = "0.4.3"
 bzip2 = { version = "0.4.4", features = ["static"] }
 tempfile = "3.12.0"
-
-[package.metadata.cargo-machete]
-ignored = [
-    # Renames itself to "md5", see: https://github.com/bnjbvr/cargo-machete/issues/8
-    "md-5"
-]
+sha2 = "0.10.8"
--- a/video/external/src/backend.rs
+++ b/video/external/src/backend.rs
@@ -1,6 +1,5 @@
 use crate::decoder::VideoDecoder;
 use bzip2::read::BzDecoder;
-use md5::{Digest, Md5};
 use ruffle_render::backend::RenderBackend;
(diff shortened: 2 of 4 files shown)


video/external/Cargo.toml
```diff
@@ -17,11 +17,11 @@ ruffle_video_software = { path = "../software" }
...
 # Needed for OpenH264:
 libloading = "0.8.5"
-md-5 = "0.10.6"
 reqwest = { version = "0.12.5", default-features = false, features = ["blocking"] }
 hex = "0.4.3"
 bzip2 = { version = "0.4.4", features = ["static"] }
 tempfile = "3.11.0"
+sha2 = "0.10.8"
 
 [package.metadata.cargo-machete]
```
torokati44, line 26: I think this section can be removed now.

video/external/src/backend.rs
```diff
@@ -106,15 +106,15 @@ impl ExternalVideoBackend {
         }
 
         // Regardless of whether the library was already there, or we just downloaded it, let's check the MD5 hash.
```
torokati44, line 108: Outdated mention of MD5 (sorry, just added it recently).

video/external/Cargo.toml
kjarosh: Removed!

torokati44 (approved):

kjarosh:
I also checked them on GH runners for the record: https://github.com/kjarosh/openh264-hashes/actions/runs/10332759908

(Maybe I'll make that repository a bit more usable for the future in case hashes change or a new version is released.)

PR #17462, opened 2024-08-10
ci: Run nextest with profile "ci", add a test-specific override to allow retries for `avm2/pixelbender_shaderdata`

torokati44:
I'm not saying this is a nice solution (because it's not at all), but the flaky CI annoyed me too much...
Somewhat related: https://github.com/ruffle-rs/ruffle/pull/17385

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,4 @@
+# TODO: Figure out why this specific test is flaky on CI, fix it, then delete this file.
+[[profile.ci.overrides]]
+filter = "test(avm2/pixelbender_shaderdata)"
+retries = 3
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -81,7 +81,7 @@ jobs:
 
       - name: Run tests with image tests
         if: runner.os != 'macOS'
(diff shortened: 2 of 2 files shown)


Dinnerbone:
but it has hit other tests too... :(

torokati44:
But wasn't that super-duper rare...? :|

torokati44:
Okay I guess we could perturb the test order and see if anything changes...

Dinnerbone:
> But wasn't that super-duper rare...? :|

I mean, we've noticed this for a few days, and it's happened at least once. Scaling that up, that means at least one flake per 3 days.

maffs.

torokati44:
> Scaling that up, that means at least one flake per 3 days.

At _least_, or at _most_? :thinking:

Anyway:
https://github.com/torokati44/ruffle/commit/3841deb3e21ef26969e76d52c7404471ddf6af85
https://github.com/torokati44/ruffle/actions/runs/10334595741

torokati44:
In light of the previous comment, what do you think now, @Dinnerbone?

Dinnerbone:
Sure thing, go for it!

Dinnerbone (approved):

PR #17468, opened 2024-08-11
avm2: Declare 'namespace AS3' and validate bytecode in playerglobals

Aaron1011:
In order for asc.jar to generate efficient bytecode (a direct 'callproperty' instead of a dynamic namespace and method lookup), we need to have 'namespace AS3' in each package where we do a qualified 'obj.AS3::SomeMethod()' call.

I've adjusted `build_playerglobal` to scan all of our playerglobal method bodies to ensure that we don't have the bad bytecode. This uncovered several places where we were missing `namespace AS3'

Before this change, we were creating a new bound method within all of our prototype methods in `XML/XMLList`, since they used 'self.AS3::SomeXmlMethod()' calls to delegate to the AS3 native method.

This saves several hundred megabytes on
`https://s1106-r2game-sq.7road.net`, since we no longer create bound `FunctionObject`s for all of the distinct XML/XMLList objects that the swf creates.

### DIFF
--- a/core/build_playerglobal/src/lib.rs
+++ b/core/build_playerglobal/src/lib.rs
@@ -14,8 +14,10 @@ use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::str::FromStr;
+use swf::avm2::read::Reader;
 use swf::avm2::types::*;
 use swf::avm2::write::Writer;
+use swf::extensions::ReadSwfExt;
 use swf::{DoAbc2, DoAbc2Flag, Header, Tag};
 use walkdir::WalkDir;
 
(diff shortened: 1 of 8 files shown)


core/src/avm2/globals/README.md
```diff
@@ -32,6 +32,18 @@ In addition to potential copyright issues around redistributing Flash's `playerg
...
+
+Under some circumstances, it may be necessarily to call a method with an explicitly-qualified AS3 namespace:
+
+```
+var xml = <accessor />;
+xml.AS3::appendChild(elem);
+```
+
+In order for this to generate efficient bytecode, you must have `namespace AS3 = "http://adobe.com/AS3/2006/builtin";` inside
+your package declaration (see `GroupElement.as` for an example). If you forget to do this, you'll get an error at compile-time:
```
evilpie, line 45: `.` or did you want to show the error?

adrian17 (approved):

core/build_playerglobal/src/lib.rs
```diff
@@ -298,6 +300,43 @@ fn strip_metadata(abc: &mut AbcFile) {
     }
 }
 
+/// If we don't properly declare 'namespace ASC' in the input to asc.jar, then
```
torokati44, line 303: nit: typo: `namespace ASC` -> `namespace AS3`

Issue #17386, opened 2024-08-06
Feature Request: Resolve more Flash detection issues for 4399


riku-42:
**Bypassing Flash Detection on 4399.com**

4399.com is a popular Flash game website in China. The site has a method for detecting Flash support in the browser. Currently, Ruffle bypasses this detection by injecting a script called [4399-content-script.ts](https://github.com/ruffle-rs/ruffle/blob/07c9901efe278ea228a5a80467efdbe40bf1c33f/web/packages/extension/src/4399-content-script.ts#L4) at `document_start`. This content script successfully resolves Flash detection issues for most games under the URL pattern `www.4399.com/flash/*`.

However, 4399.com also hosts some Flash games that use different detection methods. For example, games found at `https://my.4399.com/yxmsdzls/` and `http://www.4399.com/flash/48399.htm` still display a block screen indicating that Flash is not supported, even after installing the Ruffle extension.

![image](https://github.com/user-attachments/assets/ba4d839f-fb80-479e-aebf-25319de32c1a)

riku-42:
Issues can be solved by this feature request:
#17318

PR #17473, opened 2024-08-12
extension: add more matched 4399 sites for content scripting

riku-42:
This PR is a part of work for #17386


Dinnerbone:
I'm a *little* hesitant here, but it's not your fault - this is good, it's just that we've discovered that browsers now show ruffle as "needing permissions on <all sites> and also <each domain listed here>". It's a little ridiculous, but it's already caused some people to make assumptions and go "what's 4399? Why is ruffle talking to some chinese site?"

Adding more will make that list even scarier, maybe? Can we get away with something like `*.4399.com/*`?

Dinnerbone (approved):

web/packages/extension/src/background.ts
```diff
@@ -193,7 +193,12 @@ async function enable() {
             },
             {
                 id: "4399",
-                matches: ["https://www.4399.com/flash/*"],
+                matches: [
+                    "*://www.4399.com/flash/*",
+                    "https://news.4399.com/qiu/",
+                    "https://my.4399.com/*",
+                    "http://sjsj.4399.com/",
```
torokati44, line 200: Is the presence/absence of a `*` at the end meaningful and intentional, or accidental? :thinking:

riku-42: The absence of a * is intentional because that is the game URL!😀

Korne127: Maybe group the general URLs with the wildcards and the game URLs together each? (basically switching lines 198 and 199)? But that's a minor nitpick; this generally looks good :)

riku-42: that's a nice nitpick!

Korne127 (approved):

torokati44 (commented):
I was only asking for clarification, but I agree that it's indeed nicer grouped this way, thank you!
I'm late with the "please squash on merge" note, but oh well. 🙃

PR #17481, opened 2024-08-12
web: Don't install Flash plugin if it already exists

danielhjacobs:


web/packages/core/src/plugin-polyfill.ts
```diff
@@ -201,6 +201,9 @@ declare global {
  * @param plugin The plugin to install
  */
 export function installPlugin(plugin: RufflePlugin): void {
+    if (navigator.plugins.namedItem("Shockwave Flash")) {
```
Aaron1011, line 204: It looks like we have a function `isFlashEnabledBrowser` in `polyfills.ts` that checks for Ruffle vs real Flash Player. We might want to call that here (and remove the now-duplicated call in `pluginPolyfill`)

danielhjacobs: So that function is a slight misnomer. Really it checks that favorFlash is not set to false and also that `navigator.plugins.namedItem("Shockwave Flash")` is defined by a file other than "ruffle.js".

danielhjacobs: But here, we want:

1) If `favorFlash` is false but the browser already has Flash support, don't polyfill `navigator.plugins` as it's unnecessary, but still polyfill Flash embed/objects and add the Ruffle API functionality
2) If `installPlugin` previously added "Shockwave Flash" to `navigator.plugins`, don't polyfill `navigator.plugins` again.

danielhjacobs: What we could do is remove this `if`:

https://github.com/ruffle-rs/ruffle/blob/fd9e7ff45266ff4ff5a7223b0cd17481e8d06f6f/web/packages/core/src/polyfills.ts#L216-L218

Aaron1011 (approved):

PR #17494, opened 2024-08-13
avm2: Declare correct signature for builtin rust-side getters/setters

Aaron1011:
This only affects the handful of classes that we still implement entirely in Rust (with only an ActionScript stub when compiling playerglobals).

We now expose the proper signatures to `describeType`, and have more type information for future use in the optimizer.

### DIFF
--- a/core/src/avm2/class.rs
+++ b/core/src/avm2/class.rs
@@ -1147,6 +1147,7 @@ impl<'gc> Class<'gc> {
     }
 
     #[inline(never)]
+    #[allow(clippy::type_complexity)]
     pub fn define_builtin_instance_properties(
         self,
         mc: &Mutation<'gc>,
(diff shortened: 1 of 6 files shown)


evilpie:
This should be even simpler after #17792 lands.

core/src/avm2/class.rs
```diff
@@ -1155,24 +1156,39 @@ impl<'gc> Class<'gc> {
             &'static str,
             Option<NativeMethodImpl>,
             Option<NativeMethodImpl>,
+            Option<Gc<'gc, Multiname<'gc>>>,
```
evilpie, line 1159: Wouldn't it be simpler if this accepted `Option<&' static str>` instead?

moulins: I think the simplest thing here would be to use the `CommonNamespaces` fields in the getter/setter definitions.

Lord-McSweeney:
Superseded by #20049

PR #17495, opened 2024-08-13
avm2: Make optimizer lookup callee return type

Aaron1011:
This provides us with more optimization opportunities. Since we optimize methods one at a time, we need to look up the return type in the domain rather than relying on the resolved return type in the target `BytecodeMethod`

### DIFF
--- a/core/src/avm2/method.rs
+++ b/core/src/avm2/method.rs
@@ -8,6 +8,7 @@ use crate::avm2::value::{abc_default_value, Value};
 use crate::avm2::verify::{resolve_param_config, VerifiedMethodInfo};
 use crate::avm2::Error;
 use crate::avm2::Multiname;
+use crate::context::UpdateContext;
 use crate::string::AvmString;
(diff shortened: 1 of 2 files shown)


Lord-McSweeney:
I don't think this works; for example
```
class A {
    public function someMethod():MyType;
}
class B extends A {
    public override function someMethod():OtherType;
}
```
If a type the compiler knows is of type `A`'s `someMethod` is called, the compiler assumes the return type is of type `MyType` but it could also be of type `OtherType`. We need to verify that overriding methods have the same return type as their supermethods.

adrian17:
To be clear: it is correct and works on real SWFs, we just don't validate them as strictly as FP does?

Lord-McSweeney:
The classes
```
class A {
    public function someMethod():MyType;
}
class B extends A {
    public override function someMethod():OtherType;
}
```
throws a VerifyError in FP. I'm working on getting it to do so in Ruffle too.

Issue #17292, opened 2024-07-28
CSS parser issue with comments and whitespaces


Mesteery:
### Describe the bug

1. When there is no semi-colon before block-end, Flash stop value before newline or return while Ruffle include everthing before block-end
```as
styleSheet.parseCSS("key { property: red \r\n }");
styleSheet.getStyle("key") // {"property"="red \r\n "}
```
2. When there is newline or return before semi-colon, the value is undefined on Flash
```as
styleSheet.parseCSS("key { property: red\r\n; }");
styleSheet.getStyle("key") // {"property"="red\r\n"}
```
3. CSS Comments are not supported as opposed to Flash
```as
styleSheet.parseCSS("/*comment*/ key { property: red\r\n }");
styleSheet.styleNames // []
```

### Expected behavior

1.
```as
styleSheet.parseCSS("key { property: red \r\n }");
styleSheet.getStyle("key") // {"property"="red "}
```
2.
```as
styleSheet.parseCSS("key { property: red\r\n; }");
styleSheet.getStyle("key") // {"property"=undefined}
```
3.
```as
styleSheet.parseCSS("/*comment*/ key { property: red\r\n }");
styleSheet.styleNames // ["key"]
```

PR #17502, opened 2024-08-13
avm2: support CSS comments and fix whitespace issues

Mesteery:
Fixes: https://github.com/ruffle-rs/ruffle/issues/17292

This PR add support for CSS comments and trim whitespaces after NEWLINE/RETURN when there is no semicolon.
However the second point is not fixed but I don't think it's important:
```as
styleSheet.parseCSS("key { property: red\r\n; }");
styleSheet.getStyle("key") // RUFFLE: {"property"="red\r\n"} FP: {"property"=undefined}
```
---
Should I delete `test.fla`?


Dinnerbone (approved):
Thank you!

torokati44:
Yay for squash merge! ^^

PR #17504, opened 2024-08-13
avm2: Implement Graphics::drawRoundRectComplex

adrian17:
I really hope I didn't make a typo somewhere 😅


torokati44 (approved):
No typo spotted. Thank you!

I suppose, it would be roughly half as many parameters and lines if we could use some sort of vector/point/size types instead of `_x`/`_y` pairs of `(f64, f64)` tuples...
But this is okay too.

torokati44:
Now I wonder why `UNIT_CIRCLE_POINTS` is not symmetric (as in, `0.41421356237309503` <-> `0.4142135623730951`, `0.0` <-> `0.00000000000000006123233995736766`)...
Might have something to do with fixed point precision maths and twips...? But this is a separate matter.

adrian17:
I'm honestly not sure if this is even related to FP - originally the table was generated with Rust methods (as seen in the comments below the `const`s), change in https://github.com/ruffle-rs/ruffle/commit/068e7acc6f39c5508d121ed2e3402e52774fefc9 .

And indeed, `(PI * 0.5).cos()` in Rust is 0.00000000000000006123233995736766.
It's entirely possible we could replace (maybe even should) this by intuitive symmetric values (0.0 etc :) ) with no loss in compatibility - but not in this PR.
I'm gonna add a comment about this though :)

PR #17513, opened 2024-08-14
core: Strip BOM in URLLoader text data

Mesteery:


core/src/loader.rs
```diff
@@ -1576,6 +1576,11 @@ impl<'gc> Loader<'gc> {
                             tracing::warn!("Invalid URLLoaderDataFormat: {}", data_format);
                         }
 
+                        // strips off an optional UTF8 BOM at the beginning
```
Lord-McSweeney, line 1579: `bytearray::to_string` covers UTF-16 BOMs, and we'd probably like to cover them too. Additionally, it seems that this also applies to url variables

Aaron1011:
Are you sure that this is done in URLLoader, and not in ByteArray? There's already some BOM-related logic in ByteArray, so I wonder if it just needs to be extended

Mesteery:
> Are you sure that this is done in URLLoader, and not in ByteArray? There's already some BOM-related logic in ByteArray, so I wonder if it just needs to be extended

According to uqers, in FP, URLLoader uses URLStream, and so it's done in URLStream/readUTFBytes like in ByteArray

core/src/loader.rs
Mesteery: done

core/src/loader.rs
```diff
@@ -1540,6 +1541,40 @@ impl<'gc> Loader<'gc> {
 
                 let mut activation = Avm2Activation::from_nothing(uc);
 
+                fn strip_bom<'a, 'gc: 'a>(
+                    activation: &mut Avm2Activation<'a, 'gc>,
```
Lord-McSweeney, line 1545: We should probably avoid duplicating this function across here and `byte_array`- could you define it in `byte_array.rs`?

Mesteery: sure

Dinnerbone:
I would really appreciate a test for this!

Lord-McSweeney (approved):

PR #17517, opened 2024-08-14
avm2: report more correct Capabilities.os

Mesteery:
Not very accurate but probably better than nothing

### DIFF
--- a/core/src/avm2/globals/flash/system/Capabilities.as
+++ b/core/src/avm2/globals/flash/system/Capabilities.as
@@ -1,10 +1,7 @@
 package flash.system {
     import __ruffle__.stub_getter;
     public final class Capabilities {
-        public static function get os(): String {
-            stub_getter("flash.system.Capabilities", "os");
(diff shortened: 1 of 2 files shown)


n0samu:
See also: #10469
I think it was never merged because we weren't able to agree on how much information we want to expose, how accurate it should be, etc :(

Mesteery:
Yes I can understand. Some AIR applications rely on `os` to for example, find an OS-specific path, or use FileStream instead of URLLoader for macOS platforms (in case of Dofus), so for theses applications a correct os value is important and can possibly break the game

evilpie:
We could do this for AIR only. I also think this changes to OS for everything running on the web/extension to Linux, which we most likely don't want to do. Windows is a much much larger market.

Mesteery:
What if we put Windows instead of Linux in the else?

Mesteery:
Closing in favor of https://github.com/ruffle-rs/ruffle/pull/17590

PR #17523, opened 2024-08-15
avm2: Added all missing `[API()]` to methods from the list

MartySVK:
Dinnerbone's list - https://gist.github.com/Dinnerbone/0cfad60f4587292005d41e566368c709

except one - "String static has extra method "fromCharCode"

### DIFF
--- a/core/src/avm2/globals/flash/display/GraphicsPath.as
+++ b/core/src/avm2/globals/flash/display/GraphicsPath.as
@@ -11,6 +11,7 @@ package flash.display {
             this.winding = winding;
         }
 
+        [API("694")]
         public function cubicCurveTo(controlX1:Number, controlY1:Number, controlX2:Number, controlY2:Number, anchorX:Number, anchorY:Number):void {
             if (commands == null) {
                 commands = new Vector.<int>();
--- a/core/src/avm2/globals/flash/display/BitmapData.as
+++ b/core/src/avm2/globals/flash/display/BitmapData.as
@@ -22,6 +22,7 @@ package flash.display {
         public native function get transparent():Boolean;
 
         public native function getPixels(rect:Rectangle):ByteArray;
+        [API("682")]
         public native function copyPixelsToByteArray(rect:Rectangle, data:ByteArray):void;
         public native function getVector(rect:Rectangle):Vector.<uint>;
         public native function getPixel(x:int, y:int):uint;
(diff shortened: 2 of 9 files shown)


torokati44:
Thank you!

Unfortunately, as it turns out, we can only deal with API version numbers that are present in the table on https://github.com/ruffle-rs/ruffle/wiki/SWF-version-chart (in the two rightmost columns).

Could you please adjust the ones that are too large, and leave the correct versions in a TODO comment next to them?
The version cap is 710 for stuff that's also available in Flash Player, and 711 for things that are AIR-only.

The same goes for https://github.com/ruffle-rs/ruffle/pull/17526.

(Also, minor nit: I first understood the PR titles and commit messages as if it added the actual class/method declarations - I think "API attributes", or "API annotations" would be less ambiguous. Or, maybe the `[API()]` way as you did in https://github.com/ruffle-rs/ruffle/pull/17500 is best.)

MartySVK:
Sure, I changed it to `[API()]`

I've removed them for now, I'll wait for the Discord discussion on whether they'll be removed completely or I'll add comments.

core/src/avm2/globals/flash/display/GraphicsPath.as
```diff
@@ -11,6 +11,7 @@ package flash.display {
             this.winding = winding;
         }
 
+        [API("674")]
```
torokati44, line 14: I think this one should be 672, for Flash Player 12.
https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/display/GraphicsPath.html#cubicCurveTo()

torokati44: After triple-checking it, I just went ahead and fixed it. I hope you don't mind. If you think I'm wrong, feel free to open a follow-up PR, or yell at me on Discord, or similar.

torokati44 (approved):

torokati44:
This looks good now, thank you!

PS.: Just a tip - I see you have created a topic-specific fork https://github.com/MartySVK/ruffle-add-missing-api, with two non-specific branches `part1` and `part2` in it. This is completely okay, it's just not the way it's usually done. That would be making a generic fork, and creating change-specific branches in it. It doesn't matter for the PR though. Also, at least you didn't make it from your `master` branch, which would have been a bit problematic!

evilpie (approved):

PR #17550, opened 2024-08-17
chore: Update to wasm-bindgen 0.2.93

torokati44:
Have to keep things consistent by hand with this one.

https://github.com/rustwasm/wasm-bindgen/blob/main/CHANGELOG.md#0293

### DIFF
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -329,9 +329,10 @@ jobs:
           registry-url: https://registry.npmjs.org
 
       # wasm-bindgen-cli version must match wasm-bindgen crate version.
-      # Be sure to update in test_web.yml, web/Cargo.toml and web/README.md.
+      # Be sure to update in test_web.yml, Cargo.toml, web/docker/Dockerfile,
+      # and web/README.md as well.
       - name: Install wasm-bindgen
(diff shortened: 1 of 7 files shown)


evilpie:
I am currently looking into updating web-sys and js-sys so that we can remove `--cfg=web_sys_unstable_apis`.

torokati44:
Okay, cool! Is that directly relevant here though? Sorry if I'm missing something, but isn't that a bit independent of this?

Also, wouldn't we have to re-add it for #16794 anyway?

evilpie:
> Okay, cool! Is that directly relevant here though? Sorry if I'm missing something, but isn't that a bit independent of this?

Updating wasm-bindgen is a requirement for doing that update.

>Also, wouldn't we have to re-add it for https://github.com/ruffle-rs/ruffle/pull/16794 anyway?

You are right, quite annoying.

torokati44:
> Updating wasm-bindgen is a requirement for doing that update.

Okay, so not the other way around at least - meaning, that doesn't block this PR.

> You are right, quite annoying.

Yes, we had to go back and forth on this a couple times already, as we were leapfrogging new features with stabilized APIs...

torokati44:
> I am currently looking into updating web-sys and js-sys

Is it okay to go ahead with this one until then?

PR #17556, opened 2024-08-18
 chore: Update web-sys and js-sys 

evilpie:


web/src/audio.rs
```diff
@@ -250,7 +251,8 @@ impl Buffer {
 impl Drop for Buffer {
     fn drop(&mut self) {
         if let Some(audio_node) = self.audio_node.take() {
-            audio_node.set_onended(None);
+            let scheduled: &AudioScheduledSourceNode = &audio_node;
```
torokati44, line 254: Is this a "JS downcast" of some sort? Or required by the borrow checker? 🤔 Sorry, not at PC right now.

web/src/lib.rs
```diff
@@ -354,11 +354,7 @@ impl RuffleHandle {
 
     async fn run_context_menu_callback_paste(&self, index: usize) {
         let window = web_sys::window().expect("Missing window");
-        let Some(clipboard) = window.navigator().clipboard() else {
-            tracing::warn!("Clipboard unsupported");
-            let _ = self.with_instance(|inst| inst.js_player.display_clipboard_modal(false));
```
torokati44, line 359: Is `display_clipboard_modal` still called with `false` anywhere? If not, I suppose it can be simplified by removing its parameter. Doesn't have to be in this PR though.

torokati44:
Relevant changelog entry: https://github.com/rustwasm/wasm-bindgen/blob/main/CHANGELOG.md#0293

web/src/audio.rs
evilpie: `AudioBufferSourceNode`'s `set_onended` method was deprecated in https://github.com/rustwasm/wasm-bindgen/pull/4020. This is a way of accessing the same method explicitly in its parent class.

torokati44: I see. So more of an upcast then.

web/src/lib.rs
evilpie: You are right, I think it's unused now.

I was actually wondering if we should somehow  manually (probably something like `"clipboard" in navigator`) keep checking if `clipboard` exists.

torokati44: And I was wondering - since accessing it no longer returns an `Option`, what happens if it's not available anyway? A JS exception? Errors on use? :thinking:

evilpie: I've added an explicit check for `"clipboard" in navigator`, because otherwise the `navigator.clipboard()` just panics when the property doesn't exist.

torokati44:
cargo clippy complains

torokati44:
fixed it for you

torokati44:
Oh, `render/canvas` was left out of this. Hopefully no API changes affect it.

PR #17563, opened 2024-08-19
desktop: Use F11 as fullscreen shortcut

kjarosh:
This PR removes <kbd>Alt</kbd>+<kbd>Enter</kbd> as the shortcut and replaces it with <kbd>F11</kbd>, additionally the shortcut hint is added to the menu.

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -120,7 +120,6 @@ impl App {
         let mut next_frame_time = None;
         let mut minimized = false;
         let mut modifiers = Modifiers::default();
-        let mut fullscreen_down = false;
 
         if self.initial_movie_url.is_none() {
             // No SWF provided on command line; show window with dummy movie immediately.
@@ -333,47 +332,21 @@ impl App {
                                 return;
                             }
 
-                            // Handle fullscreen keyboard shortcuts: Alt+Return, Escape.
-                            match event {
-                                KeyEvent {
-                                    state: ElementState::Pressed,
-                                    logical_key: Key::Named(NamedKey::Enter),
-                                    ..
-                                } if modifiers.state().alt_key() => {
-                                    if !fullscreen_down {
-                                        if let Some(mut player) = self.player.get() {
-                                            player.update(|uc| {
-                                                uc.stage.toggle_display_state(uc);
-                                            });
-                                        }
-                                    }
-                                    fullscreen_down = true;
-                                    return;
-                                }
-                                KeyEvent {
-                                    state: ElementState::Released,
-                                    logical_key: Key::Named(NamedKey::Enter),
-                                    ..
-                                } if fullscreen_down => {
-                                    fullscreen_down = false;
-                                }
-                                KeyEvent {
-                                    state: ElementState::Pressed,
-                                    logical_key: Key::Named(NamedKey::Escape),
-                                    ..
-                                } => {
-                                    if let Some(mut player) = self.player.get() {
-                                        if player.is_playing() {
-                                            player.update(|uc| {
-                                                uc.stage.set_display_state(
-                                                    uc,
-                                                    StageDisplayState::Normal,
-                                                );
-                                            })
-                                        }
+                            // Handle escaping from fullscreen.
+                            if let KeyEvent {
+                                state: ElementState::Pressed,
+                                logical_key: Key::Named(NamedKey::Escape),
+                                ..
+                            } = event
+                            {
+                                if let Some(mut player) = self.player.get() {
+                                    if player.is_playing() {
+                                        player.update(|uc| {
+                                            uc.stage
+                                                .set_display_state(uc, StageDisplayState::Normal);
+                                        })
                                     }
                                 }
-                                _ => (),
                             }
 
                             let key_code = winit_to_ruffle_key_code(&event);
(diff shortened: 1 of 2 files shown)


sleepycatcoding (approved):

torokati44:
I don't think there's any technical reason - it's just that apparently Ctrl+F matches Flash Player, and one shortcut is enough for the same function.

kjarosh:
> is therr any reason we can't have both?

Every shortcut gets consumed and its keys are not forwarded to the SWF. In Flash Player it's possible to bind to <kbd>Alt</kbd>+<kbd>Enter</kbd>, so using this shortcut means that SWFs will not be able to listen to those keys. I don't know how problematic consuming <kbd>Alt</kbd>+<kbd>Enter</kbd> is, but <kbd>Ctrl</kbd>+<kbd>F</kbd> is a safer bet, as FP also uses it for fullscreen.

torokati44 (approved):

Dinnerbone:
I was under the impression it was alt enter in *some* flash player, and that's why it was chosen for Ruffle. Is that not the case, and it's ctrl f everywhere?
Can we have someone test each OS and verify?

n0samu:
See https://github.com/ruffle-rs/ruffle/pull/10426#issuecomment-1487363749:
> Side note: the Flash projector has two ways of entering fullscreen mode: Alt-Enter and Ctrl-F. Weirder still, the Alt-Enter fullscreen mode can only be exited by pressing Alt-Enter again, and the Ctrl-F mode can only be exited by pressing Esc. And very weird problems occur when you try combining both fullscreen modes. This is not something we should emulate 😆 - our current keyboard shortcut setup for entering/exiting fullscreen is good.

torokati44:
On Linux player 32,0,0,465, it's Ctrl+F.

torokati44:
> This is on Windows - I don't know about other OSes.

I wonder - is this on every piece of content? Couldn't it be that the Alt+Enter was somehow handled by the SWF itself...? :thinking:

Dinnerbone:
I'm leaning towards keeping alt enter and maybe adding F11 or something, but ctrl f is... Not a great shortcut to add. Especially so if it's only for compatibility with the Linux version of flash player.

torokati44:
Yeah it's weird that it itsn't some kind of "Find", or at least kept unused for SWFs to do something similar for it themselves. Either way, I have no strong opinions on which hotkey we keep (as long as it matches FP "somewhere") - but the rest of the changes, about putting it in the menu, I like.

n0samu:
> > This is on Windows - I don't know about other OSes.
> 
> I wonder - is this on every piece of content? Couldn't it be that the Alt+Enter was somehow handled by the SWF itself...? 🤔

Yes it is a feature of the Windows projector itself. Not some specific game

n0samu:
And yes I support adding the F11 shortcut instead of Ctrl-F

torokati44:
Does that work on Windows FP?

n0samu:
No but some of our users expect that shortcut to work so it could be a nice feature to add

Dinnerbone:
Strangely, googling for how to fullscreen flash player gets results on using F11 on windows. I'm not at the pc to try this though

torokati44:
I see. But won't it make SWFs that rely on getting that keypress event not work? 😐
Or could/should we dispatch it to the player even after having handled it ourselves? Well I guess unless the SWF also toggles fullscreen mode on it, in which case they still undo each other... 😵‍💫

n0samu:
Realistically I don't think that is going to be important. If someone complains we can figure out what to do then

Dinnerbone:
Given it's the browser fullscreen key, I'm not sure content could realistically rely on being able to catch it any case

torokati44:
Good point, forgot about the browser thing for a minute... 🤪

kjarosh:
I did some tests and it seems that <kbd>Ctrl</kbd>+<kbd>F</kbd> is the general shortcut for fullscreen both on Windows and on Linux for all versions of FP and it has existed since the View menu appeared in FP3.

<kbd>Alt</kbd>+<kbd>Enter</kbd> is a hidden shortcut that works only on Windows on newer versions of FP (introduced after FP9, before FP32).

That's the reason I selected <kbd>Ctrl</kbd>+<kbd>F</kbd>, however <kbd>F11</kbd> is a good alternative as it's the modern standard shortcut for entering fullscreen.

torokati44:
So, Ctrl+F indeed seems to be the most universal, however weird. And Esc to exit.
So I wouldn't drop these.
But I also don't mind adding F11 or Alt+Enter to toggle either.

torokati44 (approved):

kjarosh:
Changed <kbd>Ctrl</kbd>+<kbd>F</kbd> to <kbd>F11</kbd>. I don't think having multiple shortcuts for one action is a good thing and we do not follow FP already with <kbd>Ctrl</kbd>+<kbd>P</kbd>, so I think we can just do it right and use <kbd>F11</kbd>.

n0samu:
Oh, I thought we agreed to keeping Alt-Enter :(
Can we please add it back? Windows users tend to expect that shortcut to work. On laptop keyboards, pressing F11 can be inconvenient.

Issue #17564, opened 2024-08-19
angry birds space regression since nightly 2024-02-28


zjk261:
### Describe the bug

the sounds of pigs are played abnormally frequenly

https://github.com/user-attachments/assets/3b1246d7-c07c-4ced-b538-806a0402d101

### Expected behavior

https://github.com/user-attachments/assets/423fef8a-f473-439b-ade2-592ac895d7c4

### Content Location

[angry-birds-space.zip](https://github.com/user-attachments/files/16661128/angry-birds-space.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

torokati44:
This was probably due to https://github.com/ruffle-rs/ruffle/pull/15316 then...

PR #17581, opened 2024-08-19
core: Recompute next tick time of timers when they are reset

torokati44:
Should fix https://github.com/ruffle-rs/ruffle/issues/17564.

It would be great to make a test for this - both to make sure it doesn't regress, and to make extra sure that it resets based on the current time, and not on the time of the last tick. Unfortunately I don't have the means to do that easily at the moment.


Dinnerbone (approved):

torokati44:
I'll try to come up with a test for this shortly.

torokati44:
There it is. Built with this beautiful one-liner:
` docker run -it --rm -v ${PWD}:/app instrumentisto/flex-sdk mxmlc -output test.swf -compiler.debug=true Test.as`
Thanks for it, @kjarosh!

Issue #17576, opened 2024-08-19
Copy text with ctrl+c breaks Ruffle and the focus of the browser


Txori:
### Describe the bug

When you select text in the SWF file and try to copy it using Ctrl+C, it breaks the focus on the Ruffle object with this error message:

ERROR web/src/lib.rs:921
ruffle_web-wasm_extensions.js:1225
Unable to borrow Ruffle instance

Now, the focus is stuck on the Ruffle emulator, and you can’t type anything in the browser. In the example I provide, you can't type in the html text form right under the swf. But you can't use F5 to reload, or Ctrl+Shift+i to open the DevTools...

This problem does not occur when copying the text using the right mouse click.

### Expected behavior

ctrl+c should copy the text to the clipBoard without breaking anything.

### Content Location

I made a strip down example here:
https://www.txori.com/stuff/ruffle/setClipboard.php

### Affected platform

Self-hosted version

### Operating system

Windows 11

### Browser

Google Chrome Version 127.0.6533.120 (Build officiel) (64 bits)

Lord-McSweeney:
Seems like after doing the ctrl+c on the text in the SWF you can't interact with the lower text field in any way

Txori:
There might be a problem with System.setClipboard(txt.text); too, as it automatically triggers the same error message.

kjarosh:
When copying text, a temporary textarea is created which temporarily gains focus. Currently, the textarea is removed without moving the focus, which causes the focus to get lost without any `focusout` event, which causes Ruffle to think it has the focus when it's not the case. https://github.com/ruffle-rs/ruffle/pull/17582 ensures that the focus is not lost and everything works as intended.

Txori:
Cool. Thanks for fixing that nasty bug :)

PR #17582, opened 2024-08-19
web: Ensure focus is not lost when copying text

kjarosh:
Fixes https://github.com/ruffle-rs/ruffle/issues/17576.

Copying text creates a temporary `textarea` which is selected and focused. In order to preserve the focus of Ruffle, we need to move the focus into the container before removing the temporary `textarea`.


torokati44:
Is the `<canvas>` getting the focus back from its parent automatically this way, or does it not have to...?

kjarosh:
> Is the `<canvas>` getting the focus back from its parent automatically this way, or does it not have to...?

The canvas is not focusable. By default the container has focus, and in case of virtual keyboard or copying text, an input inside the container gets the focus temporarily.

We could make the canvas focusable and it would work as it's a child of the container, but it would not make any difference in behavior as we have `focusin`/`focusout` callbacks on the container anyway.

torokati44 (approved):

kjarosh:
Added `textarea.focus()` before `select()` as per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/select#notes).

PR #17590, opened 2024-08-20
avm2: For AIR report the correct OS in flash.system.Capabilities

evilpie:
This is my alternative suggestion for #17517 @Mesteery that keeps reporting the OS as windows for FlashPlayer (so not AIR) usage.

### DIFF
--- a/core/src/avm2/globals/flash/system/capabilities.rs
+++ b/core/src/avm2/globals/flash/system/capabilities.rs
@@ -3,16 +3,50 @@
 use crate::avm2::{Activation, AvmString, Error, Object, Value};
 use crate::player::PlayerRuntime;
 
+/// Implements `flash.system.Capabilities.os`
+pub fn get_os<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    _this: Object<'gc>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let os = match activation.avm2().player_runtime {
+        // For most normal Flash Player usage, the OS should not matter,
+        // so let's pretend it's Windows for the broadest possible compatibility.
+        PlayerRuntime::FlashPlayer => "Windows 8",
+        PlayerRuntime::AIR => {
+            if cfg!(windows) {
+                "Windows 10"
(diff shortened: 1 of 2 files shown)


core/src/avm2/globals/flash/system/capabilities.rs
```diff
@@ -3,16 +3,50 @@
...
+    _this: Object<'gc>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let os = match activation.avm2().player_runtime {
+        // For most normal Flash Player usage, the OS should not matter,
+        // so let's pretend it's Windows for the broadest possible compatibility.
+        PlayerRuntime::FlashPlayer => "Windows 8",
+        PlayerRuntime::AIR => {
+            if cfg!(windows) {
+                "Windows 10"
```
Dinnerbone, line 18: Why windows 10 here but windows 8 above? :D

not a blocker, just curious

Dinnerbone (approved):

core/src/avm2/globals/flash/system/capabilities.rs
evilpie: No real reason. I just kept Windows 8 because we were already using it.

Dinnerbone: Yeah 8 as the general default is smart (I'm fairly sure some swfs are subject to the windows 9 problem lol)

torokati44: Then why report Windows 10 to AIR? It's a lie anyway...

evilpie: Good question. I just took that from #17517. It's possible that most AIR content might be newer than Flash in general and is more likely to be running on Windows 10? On the other hand wikipedia claims AIR was first released in 2008, which is much earlier than I had expected.

torokati44: I suppose it doesn't matter anyway. If anyone ever discovers that it causes problems to any content, we can fix it then.

PR #17592, opened 2024-08-20
desktop: Add "Exit Full Screen" option to context menu

kjarosh:
This option allows the user to exit full screen without knowing the <kbd>Esc</kbd> shortcut.

image

### DIFF
--- a/core/src/player.rs
+++ b/core/src/player.rs
@@ -787,6 +787,12 @@ impl Player {
         context.stage.root_clip()
     }
 
+    pub fn is_fullscreen(&mut self) -> bool {
+        self.mutate_with_update_context(|context| {
+            context.stage.display_state() != StageDisplayState::Normal
+        })
+    }
+
     pub fn set_fullscreen(&mut self, is_fullscreen: bool) {
         self.mutate_with_update_context(|context| {
             let display_state = if is_fullscreen {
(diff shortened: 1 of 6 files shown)


n0samu:
Might as well also add the "Enter fullscreen" context menu option that we have on web, right?

kjarosh:
> Might as well also add the "Enter fullscreen" context menu option that we have on web, right?

I'm not so sure about that. It's on the web because there's no other way of showing it, but on desktop there's the _View_ menu with _Full Screen_ option. And when fullscreen, GUI is not visible on desktop so the context menu is the only option in that case.

n0samu:
Sure, I figured it was best to keep the context menu consistent between desktop and web but that's valid reasoning as well. Just making the suggestion

core/src/player.rs
```diff
@@ -787,6 +787,12 @@ impl Player {
         context.stage.root_clip()
     }
 
+    pub fn fullscreen(&mut self) -> bool {
```
torokati44, line 790: nit: maybe `is_fullscreen`?

torokati44 (approved):

core/src/player.rs
kjarosh: done

sleepycatcoding (approved):

torokati44:
Also huhm: https://github.com/ruffle-rs/ruffle/issues/17617

PR #17593, opened 2024-08-20
avm2: Fix GraphicsPath.cubicCurveTo() [API()] version

torokati44:
Turns out, the docs were wrong again.
Making right the amendment I made at https://github.com/ruffle-rs/ruffle/pull/17523#discussion_r1720849567.
My apologies for doubting you, @MartySVK.


kjarosh (approved):
Tested it and in fact for SWF13 it works and for SWF12 it doesn't.

PR #17604, opened 2024-08-21
desktop: Internationalize file pickers & improve filters

kjarosh:
This PR adds internationalization to strings in file pickers and improves a bit their filters.

image

### DIFF
--- a/desktop/src/gui/dialogs.rs
+++ b/desktop/src/gui/dialogs.rs
@@ -66,7 +66,7 @@ impl Dialogs {
         window: Weak<winit::window::Window>,
         event_loop: EventLoopProxy<RuffleEvent>,
     ) -> Self {
-        let picker = FilePicker::new(window);
+        let picker = FilePicker::new(window, preferences.clone());
         Self {
             preferences_dialog: None,
             bookmarks_dialog: None,
--- a/desktop/src/gui/picker.rs
+++ b/desktop/src/gui/picker.rs
@@ -1,3 +1,5 @@
+use super::text;
+use crate::preferences::GlobalPreferences;
 use rfd::AsyncFileDialog;
 use std::{
     path::PathBuf,
@@ -16,14 +18,16 @@ pub struct FilePicker {
 struct FilePickerData {
     parent: Weak<Window>,
     picking: AtomicBool,
+    preferences: GlobalPreferences,
 }
 
 impl FilePicker {
-    pub fn new(parent: Weak<Window>) -> Self {
+    pub fn new(parent: Weak<Window>, preferences: GlobalPreferences) -> Self {
         Self {
             data: Arc::new(FilePickerData {
                 parent,
                 picking: AtomicBool::new(false),
+                preferences,
             }),
         }
     }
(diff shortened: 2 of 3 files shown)


Dinnerbone (approved):

desktop/assets/texts/en-US/dialogs.ftl
```diff
@@ -1,2 +1,9 @@
 dialog-ok = OK
-dialog-cancel = Cancel
\ No newline at end of file
+dialog-cancel = Cancel
+
+picker-title-open-file = Open a single file
+picker-filter-supported = All Supported Files
+picker-filter-swf = SWF (*.swf)
+picker-filter-spl = FutureSplash Animator (*.spl)
+picker-filter-ruf = Ruffle Bundle (*.ruf)
```
torokati44, line 8: I wonder - what's the chance that these indeed can be translated in any meaningful way to something else than what they currently are? And even if so, do we want that?

Dinnerbone: FutureSplash Animator probably can't be translated but could perhaps be transliterated

Ruffle Bundle could be translated ("Ruffle" should not, but "Bundle" I think could be)

torokati44, line 8: I wonder - what's the chance that these indeed can be translated in any meaningful way to something else than what they currently are? And even if so, do we want that?

Dinnerbone: Did we just do the time warp?

torokati44: Sorry, I recreated the note right when you replied to it... Because it seems the GH UI is buggy - this was meant to also correspond to `SWF (.swf)` above...

torokati44: Hmm, to me, even "bundle" is near that territory of "jargon" words that are not meant to be messed with... Like "shader" for example.

desktop/src/gui/dialogs.rs
```diff
@@ -42,32 +44,37 @@ impl Dialogs {
         window: Weak<winit::window::Window>,
         event_loop: EventLoopProxy<RuffleEvent>,
     ) -> Self {
+        let picker = FilePicker::new(window, preferences.language());
```
sleepycatcoding, line 47: This does not seem to allow for language changes at runtime (i.e. changing it in preferences)

desktop/assets/texts/en-US/dialogs.ftl
kjarosh: I would expect that in most languages they will be the same as in English, however I added entries for them too for the sake of consistency and in case some language will require a translation (e.g. by adding a "file" suffix to SWF because it would be ambiguous otherwise — I don't know, just guessing, but I would have to know every language to make sure it's not needed). Looking for instance at Inkscape, some words might be translatable ("bundle" perhaps?):

<image src="https://github.com/user-attachments/assets/eada8ea7-233a-498b-bae6-ddaf97a643fd" width="100"/>
<image src="https://github.com/user-attachments/assets/7c62ca26-e5af-4665-93d9-28fdb980cd1c" width="100"/>

desktop/src/gui/dialogs.rs
kjarosh: Should be good now!

torokati44:
For the record, I'm neutral about this change. Not approving, but not blocking either. :man_shrugging:

desktop/src/gui/picker.rs
```diff
@@ -34,10 +38,17 @@ impl FilePicker {
...
-            .set_title("Load a Flash File");
+            .add_filter(
+                text(locale, "file-picker-filter-supported"),
+                &["swf", "spl", "ruf"],
+            )
+            .add_filter(text(locale, "file-picker-filter-swf"), &["swf"])
+            .add_filter(text(locale, "file-picker-filter-spl"), &["spl"])
+            .add_filter(text(locale, "file-picker-filter-ruf"), &["ruf"])
+            .add_filter(text(locale, "file-picker-filter-all"), &["*"])
+            .set_title(text(locale, "picker-title-open-file"));
```
sleepycatcoding, line 51: Shouldn't this be `file-picker-title-open-file`?

sleepycatcoding (approved):
Apart from my one comment, LGTM!

desktop/src/gui/picker.rs
kjarosh: I have no idea how I missed that :scream:

PR #17605, opened 2024-08-21
tests: Add compilation scripts for tests

kjarosh:
I thought I'd include the scripts I used for compiling tests. They might come in handy when editing a test in the future!

### DIFF
--- a/tests/tests/swfs/avm2/edittext_always_show_selection/compile.sh
+++ b/tests/tests/swfs/avm2/edittext_always_show_selection/compile.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+docker run -it --rm -v ${PWD}:/app instrumentisto/flex-sdk mxmlc -output test.swf -compiler.debug=true Test.as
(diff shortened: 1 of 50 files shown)


torokati44:
This is great and awesome and cool! 🥳
But ... do we need 50 of it, in a random subset of test folders? 🤔

kjarosh:
> This is great and awesome and cool! 🥳 But ... do we need 50 of it, in a random subset of test folders? 🤔

Well... those tests are authored in different ways, some of them are hand crafted, some of them are made in JPEXS, some of them compiled with a script, some of them made with Flash.

I added these scripts for tests that were compiled with those scripts (they were originally .gitignore'd). Additionally, not all of those 50 scripts are identical, some have different CLI options. So considering that these scripts are not identical and they are only added for tests that use them, I do not see any other option to commit them :/

evilpie:
I am not going to use docker for compiling so this is just superfluous noise for me.

Lord-McSweeney:
47 of them are just
```
docker run -it --rm -v ${PWD}:/app instrumentisto/flex-sdk mxmlc -output test.swf -compiler.debug=true Test.as
```
and the remaining three are the exact same command but with a SWF version specified. Sorry, but seeing as the command is mostly intuitive, and this has even more potential to confuse people (we already sometimes have Test.as, test.fla, and other files in the test folders), I don't agree with adding these.

adrian17:
How about adding this docker invocation to CONTRIBUTING.md?

If there are any instructions relevant to building individual tests, I usually add them to .as (or .toml).

kjarosh:
Closing due to lack of interest

PR #17608, opened 2024-08-22
avm2: Add AIR Screen API

Mesteery:


core/src/avm2/globals/flash/display/Screen.as
```diff
@@ -0,0 +1,66 @@
...
+  import flash.geom.Rectangle;
+
+  [API("661")]
+  public final class Screen
+  {
+    private static var _screens:Array;
+    private var _width:int;
+    private var _height:int;
+
+    [API(661)]
```
Lord-McSweeney, line 13: We'd want the "661" in quotes for consistency with other versioning

Lord-McSweeney, line 13: I don't think this needs a versioning, since the entire Screen class is already versioned

core/src/avm2/globals/flash/display/Screen.as
```diff
@@ -0,0 +1,66 @@
...
+      return new Rectangle(0, 0, _width, _height);
+    }
+
+    public function get visibleBounds():Rectangle
+    {
+      stub_getter("flash.display.Screen", "visibleBounds");
+      return bounds;
+    }
+
+    [API(661)]
```
Lord-McSweeney, line 59: Same here

core/src/avm2/globals/flash/display/ScreenMode.as
```diff
@@ -0,0 +1,41 @@
+package flash.display
+{
+  import __ruffle__.stub_getter;
+  import flash.geom.Rectangle;
+
+  // [API("733")] Ruffle doesn't support this API Version
+  [API(661)]
```
Lord-McSweeney, line 7: Can this put the versioning in quotes like `[API("661")]` for consistency with other versioning metadata?

web/src/ui.rs
```diff
@@ -211,6 +211,10 @@ impl UiBackend for WebUiBackend {
         self.update_mouse_cursor();
     }
 
+    fn get_screens_sizes(&self) -> Vec<(u32, u32)> {
+        vec![(self.canvas.width(), self.canvas.height())]
```
kjarosh, line 215: What about `window.screen.width`/`window.screen.height`? https://developer.mozilla.org/en-US/docs/Web/API/Screen/width

core/src/backend/ui.rs
```diff
@@ -344,6 +348,10 @@ impl UiBackend for NullUiBackend {
         Ok(())
     }
 
+    fn get_screens_sizes(&self) -> Vec<(u32, u32)> {
```
kjarosh, line 351:
```suggestion
    fn get_screen_sizes(&self) -> Vec<(u32, u32)> {
```

core/src/avm2/globals/flash/display/Screen.as
```diff
@@ -0,0 +1,66 @@
...
+      _height = height;
+      mode = new ScreenMode(width, height);
+    }
+
+    public static function get mainScreen():Screen
+    {
+      return screens[0];
+    }
+
+    public static native function getScreensInternal():Array;
```
kjarosh, line 28: 1. I believe this method shouldn't be public.
2. Currently this method is being executed at most once, but the screen configuration may change in time. I'd go with making all methods native and reading always the most recent info (assuming it's what AIR does).

PR #17609, opened 2024-08-22
avm2: Support AIR application arguments

Mesteery:
Depends on #17299


Dinnerbone (changes requested):
Sorry for the delay in reviewing this. A few comments but otherwise looks good to me, thank you!

PR #17620, opened 2024-08-23
desktop: Change fullscreen fscommand to use event loop events

### LINKED ISSUE #17617: can't exit fullscreen with ESC in desktop app
### Describe the bug

can't exit fullscreen with ESC in some games that enter fullscreen at launch

### Expected behavior

exit fullscreen when pressing ESC

### Content Location

[pong.zip](https://github.com/user-attachments/files/16730534/pong.zip)
[china-checkers.zip](https://github.com/ruffle-rs/ruffle/files/12286235/china-checkers.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

_No response_

sleepycatcoding:
Previously the fullscreen state was set on the window directly causing a desync with the rest of the player code.

Fixes #17617

### DIFF
--- a/desktop/src/custom_event.rs
+++ b/desktop/src/custom_event.rs
@@ -19,6 +19,9 @@ pub enum RuffleEvent {
     /// The user requested to close the current SWF.
     CloseFile,
 
+    /// The user requested to enter full screen.
+    EnterFullScreen,
+
     /// The user requested to exit full screen.
     ExitFullScreen,
 
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -504,6 +504,14 @@ impl App {
                     self.player.destroy();
                 }
 
+                winit::event::Event::UserEvent(RuffleEvent::EnterFullScreen) => {
+                    if let Some(mut player) = self.player.get() {
+                        if player.is_playing() {
+                            player.set_fullscreen(true);
+                        }
+                    }
+                }
+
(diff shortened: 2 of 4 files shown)


kjarosh:
We should be able to write a quick test for that right? fscommand + get display state from stage?

sleepycatcoding:
> We should be able to write a quick test for that right? fscommand + get display state from stage?

The test harness and desktop do not share the fscommand backend so I am not sure how useful such a test would be.

kjarosh:
Oh you're right, no easy way to write a test then :(

desktop/src/custom_event.rs
```diff
@@ -19,6 +19,9 @@ pub enum RuffleEvent {
     /// The user requested to close the current SWF.
     CloseFile,
 
+    /// The user requrested to enter full screen.
+    EnterFullscreen,
+
     /// The user requested to exit full screen.
     ExitFullScreen,
```
kjarosh, line 26: There is an inconsistency here: `EnterFullscreen` vs `ExitFullScreen`.

kjarosh (approved):

desktop/src/custom_event.rs
sleepycatcoding: Fixed.

Dinnerbone (approved):

PR #17640, opened 2024-08-25
avm2: Improve XML delete

evilpie:


core/src/avm2/activation.rs
```diff
@@ -1514,9 +1514,25 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
+        let has_xml_list_name = if multiname.has_lazy_name() {
+            let name_value = self.context.avm2.peek(0);
+            matches!(name_value, Value::Object(Object::XmlListObject(_)))
+        } else {
+            false
+        };
         let multiname = multiname.fill_with_runtime_params(self)?;
         let object = self.pop_stack();
         let object = object.coerce_to_object_or_typeerror(self, Some(&multiname))?;
+        if has_xml_list_name {
```
evilpie, line 1526: I wasn't quite sure if it would be problematic to early return before everything is popped from the stack.

Lord-McSweeney: The stack will get cleared anyway, so it should be fine

Lord-McSweeney (approved):

PR #17659, opened 2024-08-26
desktop: Allow only one file picker open at a time

kjarosh:
With asynchronous file picking, the user may open multiple file pickers by accident or thinking that the previous click did not work when the picker takes significant time to open. This PR prevents that situation by disallowing multiple picker instances.

This PR shares one commit with https://github.com/ruffle-rs/ruffle/pull/17604.

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -3,7 +3,7 @@ use crate::gui::{GuiController, MENU_HEIGHT};
 use crate::player::{LaunchOptions, PlayerController};
 use crate::preferences::GlobalPreferences;
 use crate::util::{
-    get_screen_size, gilrs_button_to_gamepad_button, parse_url, pick_file, plot_stats_in_tracy,
+    get_screen_size, gilrs_button_to_gamepad_button, parse_url, plot_stats_in_tracy,
     winit_to_ruffle_key_code, winit_to_ruffle_text_control,
 };
 use anyhow::{Context, Error};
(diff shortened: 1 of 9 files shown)


torokati44:
Hmm, just wondering: Can't we make the picker dialog a "modal child" of the main window? That would make this impossible in the first place, plus would also solve the "file picker dialog opened in the background/disappeared" issue...

kjarosh:
> Hmm, just wondering: Can't we make the picker dialog a "modal child" of the main window? That would make this impossible in the first place, plus would also solve the "file picker dialog opened in the background/disappeared" issue...

It's already been done since https://github.com/ruffle-rs/ruffle/pull/17403. However there are some DEs (_ahem... X11 ahem..._) that allow interacting with the main window even when the picker is open, and on others it's possible to press the button twice before the picker even appears.

torokati44:
Yeah, I know about that PR, but a parent-child relationship can be either modal or non-modal, even on X11 - think of tool windows in GIMP for example.

kjarosh:
Right, on Linux specifically we use XDG portals which don't have an option to specify the kind of relationship, the responsibility of that is instead delegated to the portal implementation (which should always set the picker as a modal, and I guess it's what they do).

torokati44:
Anyway, I concede to the point that relying solely on this is not a good idea, as some display protocols are inherently asynchronous, etc...

torokati44:
FYI: This and https://github.com/ruffle-rs/ruffle/pull/17604 now have merge conflicts after #17642. :no_mouth:

torokati44 (approved):
nit: The word `picker` in itself may not always be self-explanatory - it could also be a "pick graphical object under mouse pointer" thingy, for example.
It might not be that much of an issue here, where the code has more context for it, but perhaps the string IDs in _the other PR_ could have "file-picker" in them to disambiguate a little.

PR #17664, opened 2024-08-26
chore: Bump some locked crate versions, deduplicate `nix`, move away from yanked `bytes` version

torokati44:
_beep boop_ :broom: :robot: :notes:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,12 +69,13 @@ checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
 
 [[package]]
 name = "alsa"
-version = "0.9.0"
+version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -2827,7 +2828,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4979f22fdb869068da03c9f7528f8297c6fd2606bc3a4affe42e6a823fdb8da4"
 dependencies = [
  "cfg-if",
- "windows-targets 0.52.6",
+ "windows-targets 0.48.5",
```
danielhjacobs, line 2831: Downgrade?

danielhjacobs (approved):
Looks good to me assuming the downgrade is fine.

Cargo.lock
torokati44: Agh, this version always keeps jumping all over the place, seemingly for no reason and with no consequences... Lemme fix it anyway.

PR #17668, opened 2024-08-26
Ability to load ttf/otf/ttc/otc as default fonts for ruffle extension

wsxarcher:


Dinnerbone (commented):
Can we just try to load it as a ttf/etc, and then fall back to swf?
I'm not sure this would work if they provide a link to (for example) `foo.ttf?ver=123` or a dreaded `fonts.php?file=1` (though I hope in 2024 nobody's doing that anymore)

web/src/builder.rs
```diff
@@ -340,46 +340,79 @@ impl RuffleInstanceBuilder {
...
-                                        "Font {name} from font swf {font_name} contains no data"
-                                    );
+            let font_name_lower = font_name.clone().to_lowercase();
+            let extension = font_name_lower.split('.').last();
+            match extension {
+                Some("ttf") | Some("otf") | Some("ttc") | Some("otc") => {
+                    tracing::debug!("Loading font {font_name} as TTF/OTF/TTC/OTC font");
+
+                    // Check if font collection
+                    let number_of_fonts = font::fonts_in_collection(&bytes[..]).unwrap_or(1u32);
```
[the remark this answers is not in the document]
wsxarcher: fonts_in_collection return Some only if it's a collection, it does not return 1 on normal ttf. That's why the unwrap_or.
Definitely I should use also another function to parse truetypes in general by creating a new Face but I will need to expose more internal functions

core/src/font.rs
```diff
@@ -249,6 +249,38 @@ impl FontFace {
 
         Twips::ZERO
     }
+
+    pub fn is_valid(bytes: &[u8]) -> bool {
```
evilpie, line 253: I don't think we need to add these as static methods, instead of calling them directly.

web/src/builder.rs
```diff
@@ -340,46 +340,79 @@ impl RuffleInstanceBuilder {
...
+                        if number_of_fonts > 1 {
+                            format!("{} {}", font_name, i + 1)
+                        } else {
+                            font_name.to_string()
+                        }
+                    };
+
+                    player.register_device_font(FontDefinition::FontFile {
+                        name: name.to_string(),
+                        is_bold: FontFace::is_bold(bytes_slice, i),
```
evilpie, line 364: After moving the code back here, stop reparsing the fonts for querying this.

core/src/font.rs
wsxarcher: Wanted to avoid to expose ttf_parser function to the web but is better to avoid multiple parsing and loading.

web/src/builder.rs
```diff
@@ -340,53 +341,119 @@ impl RuffleInstanceBuilder {
...
+        let full_name = face
+            .names()
+            .into_iter()
+            .find(|name| name.name_id == ttf_parser::name_id::FULL_NAME)
+            .and_then(|name| name.to_string());
+
+        let name = if let Some(full_name) = full_name {
+            full_name
+        } else if number_of_fonts > 1 {
+            format!("{} {}", url, index + 1)
```
evilpie, line 443: How likely is this every going to work? Like how often are SWFs using "Font 1", "Font 2" etc? I think this doesn't happen in practice. It would be simpler to just skip (and warn) unnamed fonts.

wsxarcher: Not sure if there are unnamed fonts just wanted to handle all cases

wsxarcher: Can more than one font be registered with the same name in ruffle?

wsxarcher: So maybe better to keep this logic, it's not hard imo, just a but ugly

evilpie: No. We should not accept obvious useless stuff.

wsxarcher: Ok I will remove, should I just warn when there is no full name and use the url or fail?

evilpie: I think a warning containing the URL and font index would be fine. (That's what we do for some other failures already)

wsxarcher: Done

web/src/builder.rs
```diff
@@ -340,53 +341,104 @@ impl RuffleInstanceBuilder {
...
+                tracing::debug!("Loading font {font_name} as TTF/OTF/TTC/OTC font");
+
+                // Check if font collection
+                let number_of_fonts = ttf_parser::fonts_in_collection(bytes_slice).unwrap_or(1u32);
+
+                Self::register_ttf_face_by_name(font_name, bytes.clone(), face, 0, player);
+
+                // Register all remaining fonts in the collection if it is a collection
+                for i in 1u32..number_of_fonts {
+                    let face = ttf_parser::Face::parse(bytes_slice, i)
```
evilpie, line 355: We should not panic for unparsable faces.

wsxarcher: Done

evilpie (approved):

torokati44:
Could you please rebase this branch on top of current `master`, resolving the conflicts?

torokati44:
I've recreated the final changes of this branch as a single commit, without merges.

Dinnerbone (approved):

PR #17670, opened 2024-08-26
core: Truncate socket host by null bytes

Dinnerbone:
Example:
```actionscript
socket.connect("127.0.0.1" + unescape("%00") + "this part is ignored, flash stops at the null byte")
```

Flash truncates at any null byte - likely a low level method thinking it's getting a cstring. This emulates the truncation behaviour, fixing a Habbo connection issue from Discord.

No tests because our test suite won't run into this, it ignores hosts and everything is local.

### DIFF
--- a/core/src/socket.rs
+++ b/core/src/socket.rs
@@ -103,7 +103,7 @@ impl<'gc> Sockets<'gc> {
 
         // NOTE: This call will send SocketAction::Connect to sender with connection status.
         backend.connect_socket(
-            host,
+            sanitize_host(&host).to_string(),
             port,
             Duration::from_millis(target.timeout().into()),
             handle,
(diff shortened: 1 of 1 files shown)


adrian17 (approved):
Can you add even just a small local unit `#[test]` with the `"127.0.0.1\0nonsense` example?

Dinnerbone:
Added unit test!

core/src/socket.rs
```diff
@@ -14,6 +14,7 @@ use crate::{
 use async_channel::{unbounded, Receiver, Sender as AsyncSender, Sender};
 use gc_arena::Collect;
 use slotmap::{new_key_type, SlotMap};
+use std::ops::Deref;
```
danielhjacobs, line 17: CI seems to think this is unnecessary.

core/src/socket.rs
```diff
@@ -435,3 +436,29 @@ impl<'gc> Sockets<'gc> {
         }
     }
 }
+
+/// Flash treats a socket host as a cstring, and stops reading at a null byte.
+/// We need to account for this here.
+fn sanitize_host(host: &str) -> &str {
+    host.split('\0').next().unwrap_or(host)
```
evilpie, line 443: AFAIK Just `unwrap` should be fine.

sleepycatcoding (approved):

PR #17686, opened 2024-08-27
core: use char slicing instead of byte slicing

oadq:
Fix a panic which can occur when pasting non-ASCII characters into a textbox. Use character slicing instead of byte slicing when limiting the length of the text being pasted in.

This fixes issue #17675.


adrian17 (approved):

adrian17 (approved):

torokati44:
Thank you, and welcome to the project! ^^

Issue #17432, opened 2024-08-09
Change default value for "Open Links" on the desktop app to "Ask"


Tiagoquix:
Hello there.

Currently, "Open Links" on desktop has "Allow" as the default value.
I want to suggest to change it to "Ask" for two reasons:

1. Security: you maybe wouldn't want to accesss a random link from 10 years ago and expose your IP address and other info to that website.
2. Annoyance: sometimes I click on Flash elements that I have no idea that they are links. A confirmation in this case would be nice.

Also, "Open Links" option does not follow the pattern of having a checkbox to be able to modify the default value.
![image](https://github.com/user-attachments/assets/98238913-8211-4141-8b65-194d40ce0d08)

Not sure if making "Upgrade HTTP to HTTPS" enabled by default would be a good idea, but since Ruffle runs 100% locally on the Desktop (AFAIK) then it wouldn't be necessary, but would be cool nonetheless.

For the "Open Links" request, I believe changing "allow" to "ask" [here](https://github.com/ruffle-rs/ruffle/blob/b9054c4f45240a46d1bc91659944b78c87a4348a/desktop/src/cli.rs#L169) would be sufficient.

Thanks in advance for your attention.

Korne127:
When I introduced that feature (in #10625), there was a discussion on which value to use as the default
The consensus in the end was that it's okay to change it to Ask, but that we should then add a "Save this preference" checkbox so users don't get the prompt each time. I planned on doing that, but other things got in the way and then I forgot.
That was mostly about the web version though, but I think we should keep it consistent, and if we resolve this, change the option and add such a checkbox in both desktop and web.

PR #17690, opened 2024-08-28
desktop: Change default value of Open Links to "Ask"

oadq:
Change the default value of the "Open Links" setting from "Allow" to "Ask".
Having the default value of "Allow" may raise security concerns.

More details are in issue #17432, which this PR resolves.

### DIFF
--- a/desktop/src/cli.rs
+++ b/desktop/src/cli.rs
@@ -166,7 +166,7 @@ pub struct Opt {
     pub frame_rate: Option<f64>,
 
     /// The handling mode of links opening a new website.
-    #[clap(long, default_value = "allow")]
+    #[clap(long, default_value = "confirm")]
     pub open_url_mode: OpenURLMode,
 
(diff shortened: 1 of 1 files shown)


n0samu:
Because the prompt can become annoying I would prefer to wait until we have a "don't ask again" checkbox in the dialog to do this

oadq:
> Because the prompt can become annoying I would prefer to wait until we have a "don't ask again" checkbox in the dialog to do this

I agree, but should the checkbox disable the prompt only for the URL, the SWF, or the entire session?

Korne127:
On web, @n0samu s main concern was old flash based navigation buttons which shouldn't get a prompt for each click. So on web, the checkbox would need to be for the entire session. On desktop, that's different though as the website doesn't replace the current SWF file, so I'd only make it for the current SWF file (if possible / easy to implement, I think the whole session is okay as well).

evilpie:
Replaced by #18233.

PR #17692, opened 2024-08-28
desktop: Use egui for Open URL dialogs

kjarosh:
This is the first step to make Open URL dialogs more user friendly. This PR uses egui for these dialogs instead of rfd which:

* fixes Ruffle not responding while the dialog is open,
* fixes the problem where the dialog was a different window that was independent from Ruffle,
* allows to add "do not ask again" checkbox in the future,
* adds internationalization to the dialog.

<image width="400" src="https://github.com/user-attachments/assets/d33d501b-4021-422e-98c3-045ce81740ff"/>

Related to https://github.com/ruffle-rs/ruffle/issues/17432.


adrian17 (approved):

oadq (approved):

frontend-utils/src/backends/navigator.rs
```diff
@@ -182,26 +182,13 @@ impl<F: FutureSpawner + 'static, I: NavigatorInterface> NavigatorBackend
...
-        // If the user confirmed or if in Allow mode, open the website
-
-        // TODO: This opens local files in the browser while flash opens them
-        // in the default program for the respective filetype.
-        // This especially includes mailto links. Ruffle opens the browser which opens
-        // the preferred program while flash opens the preferred program directly.
-        match webbrowser::open(modified_url.as_ref()) {
-            Ok(_output) => {}
-            Err(e) => tracing::error!("Could not open URL {}: {}", modified_url.as_str(), e),
-        };
```
torokati44, line 204: This was a nice built-in default behavior that I now have to duplicate in the Android app.
It's literally 3 lines of code, so it's not a big deal - just thought I'd mention it (to potentially continue encouraging the consideration of the "official" alternative - out-of-repo - front-end implementation).

kjarosh: I did it being fully aware that it'll duplicate this call on Android.

The alternatives of either introducing `async` here, spawning a future, or passing a callback seemed like an overkill. I also took into account that we would like to open an app if a link points to a local file, which I assumed would have a vastly different implementation on Android vs desktop. In that case orchestrating everything by making a call here and synchronizing on the response just to invoke another method shortly afterwards seemed silly.

I could have introduced a utility method that calls `webbrowser::open`, but its only responsibility would be to just call it and possibly log the error, which also seemed like an overkill to use frontend-utils for that.

kjarosh: Also I doubt you can make blocking UI on Android (which would have to be the case without this refactor), am I right?

torokati44: > I did it being fully aware that it'll duplicate this call on Android.

That's cool then! ^^

> The alternatives of either introducing `async` here, spawning a future, or passing a callback seemed like an overkill. 

Yeah, I agree.

> Also I doubt you can make blocking UI on Android (which would have to be the case without this refactor), am I right?

I have no idea honestly, but it's not relevant either then.

PR #17706, opened 2024-08-30
desktop: Implement message dialogs in egui

kjarosh:
This PR implements MessageDialog in egui and replaces some usages of rfd with it. It provides the following improvements:

1. current dialogs do not have parent/child relationship with Ruffle, the newly added MessageDialog uses egui which renders over the content,
2. when opening a dialog Ruffle is now responsive,
3. dialogs are internationalized,
4. message dialogs are now consistent with other dialogs.

<image width="400" src="https://github.com/user-attachments/assets/399bb5d0-38e4-4b51-b2c5-dd6f8633306e"/>

### DIFF
--- a/desktop/src/gui/dialogs/message_dialog.rs
+++ b/desktop/src/gui/dialogs/message_dialog.rs
@@ -1,0 +1,62 @@
+use crate::gui::{text, LocalizableText};
+use egui::{Align2, Ui, Window};
+use unic_langid::LanguageIdentifier;
+
+pub struct MessageDialogConfiguration {
+    title: LocalizableText,
+    body: LocalizableText,
+}
+
+impl MessageDialogConfiguration {
+    pub fn new(title: LocalizableText, body: LocalizableText) -> Self {
+        Self { title, body }
+    }
+}
+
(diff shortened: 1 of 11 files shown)


desktop/src/gui/dialogs/message_dialog.rs
```diff
@@ -0,0 +1,62 @@
...
+        keep_open && !should_close
+    }
+
+    pub fn render_window_contents(&mut self, locale: &LanguageIdentifier, ui: &mut Ui) -> bool {
+        let mut should_close = false;
+
+        ui.vertical_centered(|ui| {
+            ui.label("");
+            ui.label(self.config.body.localize(locale));
+            ui.label("");
```
sleepycatcoding, line 49: Why are these empty labels necessary?

sleepycatcoding (approved):

desktop/src/gui/dialogs/message_dialog.rs
sleepycatcoding: I see, it does look better with the padding, so this is probably fine.

PR #17707, opened 2024-08-30
avm2: Throw special XML parsing error for unmatched end tag at the root level

evilpie:
I am frankly not sure why this weird case needs a special error, but whatever.


adrian17 (approved):

adrian17:
Actually, can you squash them please? So we don't have a commit failing to test.

PR #17711, opened 2024-08-30
avm2: Correctly set XML attribute with namespace

evilpie:


core/src/avm2/object/xml_list_object.rs
```diff
@@ -722,6 +722,7 @@ impl<'gc> TObject<'gc> for XmlListObject<'gc> {
                                 // 2.c.iv.3. Let y.[[Class]] = "attribute"
                                 E4XNode::attribute(
                                     activation.gc(),
+                                    x.explicit_namespace().map(E4XNamespace::new_uri),
```
adrian17, line 725: (not relevant to this PR, but I think this shows why it'd be nicer for the same Namespace objects to be used between E4X and AS3 Namespace :c )

adrian17 (approved):
LGTM, just please squash the now-working tests.

core/src/avm2/object/xml_list_object.rs
evilpie: Yeah. Though XML's namespaces are much simpler than AS3's. I think `explicit_namespace` is also a bit of a hack from the beginning of E4X namespace support that we should look at at some point.

PR #17714, opened 2024-08-31
core: Add support for wider range of key codes

kjarosh:
This PR changes the definition of `KeyCode` from an enum, which was constraining the possible key codes, to a struct with a `u32` field, which covers all possible cases of a key code.

Key codes in Flash are not related to the physical buttons, but rather to logical button values. For instance, typing a non-ASCII key produces a key code related to that character, and using a non-ANSI keyboard or a layout different from US QWERTY also produces results matching that assumption.

This refactor opens up possibilities to match FP's behavior related to key codes and make this behavior configurable in the future.

CC @evilpie


core/src/events.rs
```diff
@@ -466,148 +466,153 @@ impl TextControlCode {
...
-    Period = 190,
-    Slash = 191,
-    Grave = 192,
-    LBracket = 219,
-    Backslash = 220,
-    RBracket = 221,
-    Apostrophe = 222,
-}
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub struct KeyCode(u32);
```
evilpie, line 470: Mostly curious, why did this change from `u8` to `u32`, are there any valid values that exceed the previous limit?

desktop/src/cli.rs
```diff
@@ -281,3 +281,125 @@ impl Opt {
         })
     }
 }
+
+// TODO The following enum exists in order to preserve
```
evilpie, line 285: Maybe we could just re-use the names used on the web for `KeyboardEvent` `code`. Or just go basic and use the integer key codes.

evilpie:
> This PR changes the definition of KeyCode from an enum, which was constraining the possible key codes, to a struct with a u32 field, which covers all possible cases of a key code.

Are there actual examples of this?

kjarosh:
> Are there actual examples of this?

Of course, you can just try inputting any non-ASCII character, for instance using an ANSI keyboard:

* (Polish layout) pressing <kbd>Alt Gr</kbd>+<kbd>A</kbd> (ą) produces key code 261, the value of `ą`,
* (Polish layout) pressing <kbd>Alt Gr</kbd>+<kbd>1</kbd> (≠) produces key code 8800, the value of `≠`,
* (German layout) pressing <kbd>;</kbd> (ö) produces key code 246, the value of `ö`,
* (German layout) pressing <kbd>-</kbd> (ß) produces key code 223, the value of `ß`,
* (German layout) pressing <kbd>Alt Gr</kbd>+<kbd>B</kbd> (“) using German layout produces key code 8220, the value of `“`,
* (Polish layout) pressing <kbd>Alt Gr</kbd>+<kbd>6</kbd> (½) produces key code 189, the value of `½`, and also the key code produced when pressing just <kbd>-</kbd>.

See also

https://github.com/ruffle-rs/ruffle/blob/556fd7306d4f8e49a728d5ea4fc182ff04e35f03/desktop/src/util.rs#L177-L181

desktop/src/cli.rs
kjarosh: I think the whole CLI needs to be reviewed before 1.0.0 and potentially changed in one PR. I believe our users would not appreciate it changing frequently. TBH I'd be aiming for preserving the behavior of CLI in this specific PR.

evilpie:
Thank you! It seems totally obvious in hindsight.

desktop/src/cli.rs
evilpie: We definitely should not change it in this PR. I was just trying to brainstorm ideas.

evilpie (approved):

desktop/src/cli.rs
kjarosh: I'd personally just use numeric codes in that case. Remapping buttons with CLI is obscure anyway and we'll have a nice GUI for that in the future :eyes:

PR #17716, opened 2024-08-31
ci: Fix failing doctests and run them in CI

kjarosh:
Unfortunately nextest does not support doctests yet (https://github.com/nextest-rs/nextest/issues/16), they need to be run using cargo test.

This adds ~10s on Linux and macOS and ~15s on Windows.

### DIFF
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -10,6 +10,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  FEATURES: lzma,jpegxr
+
 jobs:
   changes:
     name: Paths filter
@@ -67,6 +70,11 @@ jobs:
         if: runner.os == 'Linux'
         run: echo RUSTFLAGS=${RUSTFLAGS}\ --cfg=web_sys_unstable_apis\ -Clink-args=-znostart-stop-gc >> $GITHUB_ENV
 
+      - name: Enable image tests
+        if: runner.os != 'macOS'
+        shell: bash
+        run: echo FEATURES=${FEATURES},imgtests | tee -a $GITHUB_ENV
+
       - name: Cache Cargo output
         uses: Swatinem/rust-cache@v2
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/test_rust.yml
```diff
@@ -65,9 +84,14 @@ jobs:
       # Additionally: https://lld.llvm.org/ELF/start-stop-gc
       - name: Disable linker start-stop-gc
         # Note: We also use `rust-lld` on Windows, see `config.toml`.
-        if: runner.os != 'macOS'
+        if: runner.os == 'Linux'
```
torokati44, line 87: Doesn't that comment need some adjustment? While it's still true, it's now relevant somewhat differently now... As in, the interesting point is not that we also have to do this on Windows because `rust-lld` is also used there, but _not_ having to do it _despite_ that...? :thinking:

.github/workflows/test_rust.yml
```diff
@@ -10,6 +10,25 @@ concurrency:
...
+  # Supposedly makes "Updating crates.io index" faster on Windows.
+  # See: https://github.com/rust-lang/cargo/issues/9167
+  CARGO_NET_GIT_FETCH_WITH_CLI: 'true'
+
+  # Workaround for: https://github.com/nextest-rs/nextest/issues/1493
+  # See also: https://github.com/rust-lang/rustup/issues/3825
+  RUSTUP_WINDOWS_PATH_ADD_BIN: '1'
+
+  # dummy value, just to silence warnings about it missing
+  XDG_RUNTIME_DIR: ''
```
torokati44, line 30: Might be worth noting that this is only relevant on Linux.

.github/workflows/test_rust.yml
```diff
@@ -10,6 +10,25 @@ concurrency:
...
+  FEATURES: lzma,jpegxr
+  TEST_OPTS: --workspace --locked --no-fail-fast -j 4
+
+  # This is to counteract the disabling by rust-cache.
+  # See: https://github.com/Swatinem/rust-cache/issues/43
+  CARGO_INCREMENTAL: '1'
+
+  # Supposedly makes "Updating crates.io index" faster on Windows.
+  # See: https://github.com/rust-lang/cargo/issues/9167
+  CARGO_NET_GIT_FETCH_WITH_CLI: 'true'
```
torokati44, line 23: I'm leaning toward just removing this altogether, instead of also setting it for macOS. But it can also stay like this.

torokati44 (approved):
Sorry for my late response - none of my notes are actual blockers.

.github/workflows/test_rust.yml
kjarosh: done

.github/workflows/test_rust.yml
kjarosh: I can do that in a separate PR to compare speeds etc.

.github/workflows/test_rust.yml
kjarosh: removed the comment altogether as it seems it's an ELF issue

torokati44:
Thank you!

PR #17727, opened 2024-09-01
avm2: Implement Date using ActionScript

evilpie:


core/src/avm2/globals/Date.as
```diff
@@ -1,5 +1,461 @@
...
+        public static const length:int = 7;
+
+        public function Date(year:* = undefined, month:* = undefined, day:* = undefined, hours:* = undefined, minutes:* = undefined, seconds:* = undefined, ms:* = undefined) {
+            this.init(arguments);
+        }
+        private native function init(args:Array);
+
+        public static native function parse(param1:*):Number;
+
+        public static native function UTC(param1:*, param2:*, param3:* = 1, param4:* = 0, param5:* = 0, param6:* = 0, param7:* = 0, ... rest):Number;
```
adrian17, line 14: We usually try to keep the naming from docs/playerglobal. These names do show up in exception messages (though currently not sure if this can happen in Ruffle, better have correct names just in case)

core/src/avm2/globals/Date.as
```diff
@@ -1,5 +1,461 @@
...
+        AS3 native function toUTCString():String;
+
+        AS3 native function getUTCDay():Number;
+
+        AS3 native function getDay():Number;
+
+        AS3 native function getTimezoneOffset():Number;
+
+        AS3 native function getTime():Number;
+        AS3 native function setTime(time:* = undefined):Number;
```
adrian17, line 43: In docs/playerglobal, this isn't optional (but it is in the prototype method).

core/src/avm2/globals/Date.as
```diff
@@ -1,5 +1,461 @@
...
+            var d:Date = this;
+            return d.AS3::setTime(t);
+        }
+        prototype.valueOf = function():* {
+            var d:Date = this;
+            return d.AS3::valueOf();
+        }
+        prototype.toString = function():String {
+            var d:Date = this;
+            return d.AS3::toString();
```
adrian17, line 258: can you make sure that this generates proper bytecode? (I think it should, given globals.as, but let's make sure just in case)

core/src/avm2/globals/Date.as
```diff
@@ -1,5 +1,461 @@
...
+        prototype.setPropertyIsEnumerable("setUTCFullYear", false);
+        prototype.setPropertyIsEnumerable("setUTCMonth", false);
+        prototype.setPropertyIsEnumerable("setUTCDate", false);
+        prototype.setPropertyIsEnumerable("setUTCHours", false);
+        prototype.setPropertyIsEnumerable("setUTCMinutes", false);
+        prototype.setPropertyIsEnumerable("setUTCSeconds", false);
+        prototype.setPropertyIsEnumerable("setUTCMilliseconds", false);
     }
-}
+}
```
Lord-McSweeney, line 461: Seems like the newline was removed accidentally

core/src/avm2/globals/Date.as
```diff
@@ -1,5 +1,461 @@
...
+        public function get day():Number {
+            return this.AS3::getDay();
+        }
+
+        public function get dayUTC():Number {
+            return this.AS3::getUTCDay();
+        }
+
+
+        prototype.setTime = function(t:* = undefined):Number {
```
Lord-McSweeney, line 248: I think we'd prefer to have prototype methods defined in the class initializer at the top of the class to match FFDEC and the rest of the globals

core/src/avm2/globals/Date.as
evilpie: I think this is the name in the playerglobal.

core/src/avm2/globals/Date.as
evilpie: This seems to be optional in Flash.

core/src/avm2/globals/Date.as
evilpie: I can't open this class in JPEXS. I did test that `Date.prototype.toString.call()` works. (But somehow it doesn't in Flash, which I don't understand)

core/src/avm2/globals.rs
```diff
@@ -826,6 +822,7 @@ fn load_playerglobal<'gc>(
             ("", "TypeError", typeerror),
             ("", "URIError", urierror),
             ("", "VerifyError", verifyerror),
+            ("", "Date", date),
```
Lord-McSweeney, line 825: This should be in alphabetical order to match the classes around it

core/src/avm2/globals/date.rs
```diff
@@ -322,606 +253,567 @@ pub fn call_handler<'gc>(
...
+
+    let timestamp = DateAdjustment::new(activation, &Utc)
+        .millisecond(args.get(0))?
+        .apply(this);
+    Ok(timestamp.into())
 }
 
-/// Implements `secondsUTC` property's getter, and the `getUTCSeconds` method.
-pub fn seconds_utc<'gc>(
+/// Implements he `getUTCSeconds` method.
```
Lord-McSweeney, line 589: typo: "he" -> "the"

Lord-McSweeney (approved):
Thanks for working on this!

PR #17732, opened 2024-09-01
avm2: Make Sprite.startDrag/stopDrag available to SWF 9

Aaron1011:
The versioning in playerglobal is incorrect - Flash Player actually exposes this to SWF 9


adrian17 (approved):

core/src/avm2/globals/flash/display/Sprite.as
```diff
@@ -18,9 +18,7 @@ package flash.display {
         public native function get useHandCursor():Boolean;
         public native function set useHandCursor(useHandCursor:Boolean):void;
 
-        [API("667")]
         public native function startDrag(lockCenter:Boolean = false, bounds:Rectangle = null):void;
-        [API("667")]
         public native function stopDrag():void;
```
torokati44, line 22: Leaving a comment about the mismatch would have been nice for future reference.

torokati44:
Never mind my note - the versioning in (at least my version of) `playerglobal.swc` is as follows:

```as3
      public native function startDrag(param1:Boolean = false, param2:Rectangle = null) : void;
      public native function stopDrag() : void;
      
      [API("667")]
      public native function startTouchDrag(param1:int, param2:Boolean = false, param3:Rectangle = null) : void;
      [API("667")]
      public native function stopTouchDrag(param1:int) : void;
```

So, the `startDrag` and `stopDrag` methods are (correctly) unversioned, while the `startTouchDrag` and `stopTouchDrag` (note the `Touch` part) are the versioned ones. This was probably the cause of the confusion initially, when the `[API()]` annotations were added, and also when they were referred to as being incorrect.

PR #17733, opened 2024-09-01
core: Replace `on_exit_frame` with iteration over Loaders

Aaron1011:
This removes the need to traverse the entire display object tree - we instead just try to fire events (when ready) on corresponding `MovieClip`s for our `Loader::Movie` instances

### DIFF
--- a/core/src/avm2/object/loaderinfo_object.rs
+++ b/core/src/avm2/object/loaderinfo_object.rs
@@ -259,12 +259,15 @@ impl<'gc> LoaderInfoObject<'gc> {
         self.0.complete_event_fired.set(false);
     }
 
+    /// Fires the 'init' and 'complete' events if they haven't been fired yet.
+    /// Returns `true` if both events have been fired (either as a result of
+    /// this call, or due to a previous call).
     pub fn fire_init_and_complete_events(
         &self,
         context: &mut UpdateContext<'gc>,
         status: u16,
         redirected: bool,
-    ) {
+    ) -> bool {
         self.0.expose_content.set(true);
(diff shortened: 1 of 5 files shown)


wsxarcher:
Seems related to this
https://github.com/ruffle-rs/ruffle/issues/16490

core/src/display_object.rs
```diff
@@ -2066,19 +2067,12 @@ pub trait TDisplayObject<'gc>:
 
     /// Emit an `exitFrame` broadcast event.
     fn exit_frame(&self, context: &mut UpdateContext<'gc>) {
-        let exit_frame_evt = Avm2EventObject::bare_default_event(context, "exitFrame");
+        let exit_frame_evt: Avm2Object<'_> =
```
adrian17, line 2070: (why was this change needed?)

core/src/avm2/object/loaderinfo_object.rs
```diff
@@ -264,7 +264,7 @@ impl<'gc> LoaderInfoObject<'gc> {
         context: &mut UpdateContext<'gc>,
         status: u16,
         redirected: bool,
-    ) {
+    ) -> bool {
```
adrian17, line 267: can you add a comment on the return value?
Without some documentation, it took me a moment to understand what the `return true; } return false; } true` chain corresponds to

adrian17 (approved):

core/src/display_object.rs
Aaron1011: I accidentally inserted this with rust-analyer - removed

PR #17740, opened 2024-09-02
avm2: Misc fixes

Lord-McSweeney:


core/src/avm2/globals/vector.rs
```diff
@@ -249,12 +249,16 @@ pub fn concat<'gc>(
         return Err("Not a vector-structured object".into());
     };
 
+    let original_length = new_vector_storage.length();
+
+    let use_swf10_behavior = activation
+        .caller_movie()
```
adrian17, line 255: Pretty sure that this also is supposed to rely on declarer SWF (except for builtins, which propagate caller's context).
Can you document this with a `//TODO` here? I know we have more places like this already.

core/src/avm2/globals/vector.rs
```diff
@@ -284,21 +288,17 @@ pub fn concat<'gc>(
             continue;
         };
 
-        for val in old_vec {
-            if let Ok(val_obj) = val.coerce_to_object(activation) {
-                if !val.is_of_type(activation, val_class) {
-                    let other_val_class = val_obj.instance_class();
-                    return Err(format!(
-                        "TypeError: Cannot coerce Vector value of type {:?} to type {:?}",
```
adrian17, line 292: oh, it's not a real error in FP?

adrian17 (approved):
Can you also add a test before merging?

core/src/avm2/globals/vector.rs
Lord-McSweeney: The `coerce_to_type` should throw the error anyway; `Vector/typechecking` tests this

core/src/avm2/globals/vector.rs
Lord-McSweeney: `caller_movie` tracks the SWF that called the native function, and if it was called by another native function, it will use the new behavior, matching avmplus.

Lord-McSweeney:
All of the changes should be covered by tests

evilpie:
The XML namespace equality check doesn't seem to have a test.

evilpie:
I think the test for the namespace equality wasn't commited.

PR #17756, opened 2024-09-04
core: Use proper sandbox type per SWF

kjarosh:
It's the first step of implementing the security sandbox and its filesystem/network separation policy. Before that Ruffle assigned `localTrusted` sandbox type for all movies except web, where `remote` was used. The sandbox type also was assigned to the player, and not SWFs as it should. This PR does not introduce any sandbox policies based on the sandbox type, just the proper sandbox type detection. It is also not possible to specify trusted movies, and AIR applications don't use the `application` type yet.

The logic goes as follows. It is detected whether a SWF is local or remote based on its URL (URLs with scheme `file` are treated as local). Every remote SWF receives the `remote` sandbox type. For local SWFs, their FileAttributes are checked for `useNetwork`, which if true will assign `localWithNetwork` and if false will assign `localWithFile` sandbox type.

This behavior is covered by the following set of tests:
* `avm*/sandbox_type_local_network` &ndash; `localWithNetwork` type in AVM1/AVM2,
* `avm*/sandbox_type_local_file` &ndash; `localWithFile` type in AVM1/AVM2,
* `avm*/sandbox_type_remote` &ndash; `remote` type in AVM1/AVM2, a `localWithNetwork` SWF loads another one from the network.

The cases from the tests above were the only cases I found where implementing sandbox policies was not needed for the test to pass.


Croworbit:
this sounds good, how does this help?
(not saying it won't,  I know sandboxing can be good but don't actually know much)

evilpie:
I am really quite curious about how this is going to work in practice. e.g. can SWFs with localWithNetwork load SWFs with localWithFile? I am mostly thinking about the boundary between different SWFs with different sandbox levels.

core/src/sandbox.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    /// The movie is a trusted local movie with access to both filesystem and network.
+    LocalTrusted,
+
+    /// The movie is an AIR application with access to both filesystem and network.
+    Application,
+}
+
+impl SandboxType {
+    /// Detect sandbox type based on SWF URL and its header.
```
evilpie, line 37: Not sure I would go with `detect`. Maybe `infer`?

core/src/tag_utils.rs
```diff
@@ -92,30 +104,38 @@ impl SwfMovie {
     /// `LoaderInfo.bytesTotal` is set to the actual value, but no data is available,
     /// and `LoaderInfo.parameters` is empty.
     pub fn fake_with_compressed_len(swf_version: u8, compressed_len: usize) -> Self {
+        let url = "file:///".to_string();
```
evilpie, line 107: Pretty sure this is going to be a problem in the future. We need something like #17717.

core/src/sandbox.rs
kjarosh: I agree, changed to `infer`

core/src/tag_utils.rs
kjarosh: That's right, added TODOs for now, but this will need a lot more tests for sure

kjarosh:
> I am really quite curious about how this is going to work in practice. e.g. can SWFs with localWithNetwork load SWFs with localWithFile? I am mostly thinking about the boundary between different SWFs with different sandbox levels.

From my testing (if I remember correctly):
* `localWithFile` can load `localWithFile`,
* `localWithFile` cannot load `localWithNetwork`,
* `localWithNetwork` cannot load `localWithFile`,
* `localWithNetwork` can load `localWithNetwork` despite not having access to the filesystem,
* `localWithNetwork` can load `remote`,
* `localWithNetwork` cannot interact with objects from `remote`,
* `localWithFile` cannot load `remote`,
* `remote` cannot load `localWithFile` and `localWithNetwork`.

I'll be adding some tests (known failures) related to these behaviors along with the thrown exceptions so hopefully I'll cover these cases and even some more.

In AVM1 IIRC no exceptions are thrown, actions are canceled/prevented and security sandbox messages are traced.

evilpie (approved):

PR #17757, opened 2024-09-04
chore: Fix clippy errors on beta

kjarosh:


sleepycatcoding:
It seems there are more occurrences of the lint.

danielhjacobs:
Testing locally, there's a lot more then that. To test this locally:

- `rustup update`
- `rustup default beta`
- `cargo clippy --all --all-features --tests -- -D warnings`
- Make changes
- Run cargo command again

Don't forget to switch back to rustup default stable after fixing all occurences.

kjarosh:
:sob: I will fix other occurrences in a minute. I was wondering why only one error was raised in the CI...

torokati44:
Instead of switching the default, you can always: `cargo +beta clippy ...`

sleepycatcoding (approved):

kjarosh:
The remaining issue that fails is `std::panic::PanicInfo` renamed to `std::panic::PanicHookInfo` (https://releases.rs/docs/1.81.0/#compatibility-notes), which was released in 1.81.0 and deprecated in 1.82.0. Unfortunately beta is 1.82 already but stable is still 1.80.

torokati44:
That's supposed to change tomorrow... :)
https://www.whatrustisit.com/

swf/src/types.rs
```diff
@@ -1786,6 +1786,7 @@ pub type DoAction<'a> = &'a [u8];
 pub type JpegTables<'a> = &'a [u8];
 
 /// `ProductInfo` contains information about the software used to generate the SWF.
```
torokati44, line 1788: nit: Repeating the type name is redundant here, just like it was for `enum_trait_object`.

render/src/pixel_bender.rs
```diff
@@ -272,7 +272,9 @@ pub enum PixelBenderShaderArgument<'a> {
     },
 }
 
-/// An image input. This accepts both an owned BitmapHandle,
+/// An image input.
+///
+/// This accepts both an owned BitmapHandle,
 /// and a borrowed texture (used when applying a filter to
 /// a texture that we don't have ownership of, and therefore
 /// cannot construct a BitmapHandle for).
```
torokati44, line 280: nit: I probably would have re-wrapped this neatly into 3 lines, but oh well...

torokati44 (approved):
I think the `PanicInfo` rename can be dealt with in a separate PR later. Thank you!

swf/src/types.rs
kjarosh: fixed

render/src/pixel_bender.rs
kjarosh: fixed

PR #17771, opened 2024-09-05
avm2: Do not allocate for `Namespace::any()`

moulins:
`Namespace` now holds a `Option<Gc<_>>` internally, with `None` representing the Any namespace.

This is closer to how avmplus represents namespaces internally (they use a `NULL` pointer for the Any case), and allows creating the Any namespace without a `&Mutation<'gc>`.

### DIFF
--- a/core/src/avm2/specification.rs
+++ b/core/src/avm2/specification.rs
@@ -370,7 +370,10 @@ impl Definition {
     ) {
         for class_trait in traits {
             if !class_trait.name().namespace().is_public()
-                && class_trait.name().namespace() != avm2.as3_namespace
+                && class_trait
+                    .name()
+                    .namespace()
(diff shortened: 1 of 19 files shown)


Lord-McSweeney (approved):

torokati44:
Let's not allow this to start bit-rotting.

core/src/avm2/specification.rs
```diff
@@ -370,7 +370,10 @@ impl Definition {
     ) {
         for class_trait in traits {
             if !class_trait.name().namespace().is_public()
-                && class_trait.name().namespace() != avm2.as3_namespace
+                && class_trait
+                    .name()
+                    .namespace()
+                    .exact_version_match(avm2.as3_namespace)
```
torokati44, line 376: Oh, wasn't this condition negated...?

PR #17792, opened 2024-09-06
avm2: Port Namespace to ActionScript

Lord-McSweeney:
This slows down `new Namespace()` by about 220%, but that's not hot anyway.


evilpie:
This looks good to me, but I don't feel confident enough to review the `load_playerglobal` completely.

torokati44:
FWIW, Clippy is not happy about some lifetimes not being removed in this.

Issue #17776, opened 2024-09-06
Files without extension part in their name aren't shown to be opened


PavanLuca:
Any file without its extension format in its name isn't shown in the file picker.

Zorin OS Core 17.1, X11

kjarosh:
It appears that rfd just does not support filters for files without a dot in their name:

https://github.com/PolyMeilex/rfd/blob/42dcc7d61fc5e278b4ed76bb9720ba4d89266f01/src/backend/xdg_desktop_portal.rs#L31

PR #17793, opened 2024-09-06
deps: Bump RFD to the newest revision

kjarosh:
This fixes filters in XDG portals and allows showing files without an extension.

Fixes https://github.com/ruffle-rs/ruffle/issues/17776.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -248,9 +248,9 @@ dependencies = [
 
 [[package]]
 name = "ashpd"
-version = "0.8.1"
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -6060,6 +6060,18 @@ dependencies = [
  "wayland-scanner",
 ]
 
+[[package]]
```
torokati44, line 6063: :/

torokati44 (approved):

PR #17799, opened 2024-09-06
video: Prefer loading system OpenH264 library to downloading it

kjarosh:
Some systems do have OpenH264 installed and in order to get rid of the delay caused by downloading the library (and needlessly downloading it in the first place), an attempt is made first to load OpenH264 as a system library. When loading fails (or version is mismatched), only then Ruffle downloads OpenH264 and uses that version of the library.

This also will make it possible to support OpenH264 in Flatpak.

### DIFF
--- a/video/external/src/backend.rs
+++ b/video/external/src/backend.rs
@@ -26,6 +26,7 @@ enum ProxyOrStream {
 }
 
 struct OpenH264Data {
+    local_filenames: Vec<&'static str>,
     download_filename: &'static str,
     download_sha256: &'static str,
 }
@@ -48,6 +49,12 @@ impl ExternalVideoBackend {
     fn get_openh264_data() -> Result<OpenH264Data, Box<dyn std::error::Error>> {
         const OS: &str = std::env::consts::OS;
         const ARCH: &str = std::env::consts::ARCH;
+
+        let local_filenames = match OS {
+            "linux" => vec!["libopenh264.so.7", "libopenh264.so.2.4.1", "libopenh264.so"],
+            // TODO: investigate other OSes
+            _ => vec![],
+        };
 
         // Source: https://github.com/cisco/openh264/releases/tag/v2.4.1
         let (download_filename, download_sha256) = match (OS, ARCH) {
@@ -87,17 +94,19 @@ impl ExternalVideoBackend {
         };
 
         Ok(OpenH264Data {
+            local_filenames,
             download_filename,
             download_sha256,
         })
     }
 
-    pub fn get_openh264() -> Result<PathBuf, Box<dyn std::error::Error>> {
+    fn download_openh264(
+        openh264_data: &OpenH264Data,
+    ) -> Result<PathBuf, Box<dyn std::error::Error>> {
         // See the license at: https://www.openh264.org/BINARY_LICENSE.txt
         const URL_BASE: &str = "http://ciscobinary.openh264.org/";
         const URL_SUFFIX: &str = ".bz2";
 
-        let openh264_data = Self::get_openh264_data()?;
         let (filename, sha256sum) = (
             openh264_data.download_filename,
             openh264_data.download_sha256,
@@ -140,19 +149,32 @@ impl ExternalVideoBackend {
         Ok(filepath)
     }
 
-    pub fn new(openh264_lib_filepath: Option<PathBuf>) -> Self {
-        let h264_codec = if let Some(openh264_lib_filepath) = openh264_lib_filepath {
-            tracing::info!("Using OpenH264 at {:?}", openh264_lib_filepath);
-            OpenH264Codec::new(&openh264_lib_filepath)
-                .inspect_err(|err| tracing::error!("Error loading OpenH264: {:?}", err))
-                .ok()
-        } else {
-            None
-        };
+    pub fn load_openh264() -> Result<OpenH264Codec, Box<dyn std::error::Error>> {
+        let openh264_data = Self::get_openh264_data()?;
 
+        for filename in &openh264_data.local_filenames {
+            match OpenH264Codec::new(filename) {
+                Ok(codec) => return Ok(codec),
+                Err(err) => {
+                    tracing::warn!(
+                        "Failed to load system OpenH264 library {}: {}",
+                        filename,
+                        err
+                    );
+                }
+            }
+        }
+
+        tracing::info!("Downloading OpenH264 library");
+        let filename = Self::download_openh264(&openh264_data)?;
+        tracing::info!("Using OpenH264 at {:?}", filename);
+        Ok(OpenH264Codec::new(&filename)?)
+    }
+
+    pub fn new(openh264_codec: Option<OpenH264Codec>) -> Self {
         Self {
             streams: SlotMap::with_key(),
-            openh264_codec: h264_codec,
+            openh264_codec,
             software: SoftwareVideoBackend::new(),
         }
     }
--- a/video/external/src/backend.rs
+++ b/video/external/src/backend.rs
@@ -1,3 +1,4 @@
+use crate::decoder::openh264::OpenH264Codec;
 use crate::decoder::VideoDecoder;
 use bzip2::read::BzDecoder;
 use ruffle_render::backend::RenderBackend;
@@ -28,7 +29,7 @@ enum ProxyOrStream {
 /// except for H.264, for which it uses an external decoder.
 pub struct ExternalVideoBackend {
     streams: SlotMap<VideoStreamHandle, ProxyOrStream>,
-    openh264_lib_filepath: Option<PathBuf>,
+    openh264_codec: Option<OpenH264Codec>,
     software: SoftwareVideoBackend,
 }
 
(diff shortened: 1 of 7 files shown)


torokati44:
Just a random observation on my Fedora 40 system:
```
md5sum /usr/lib64/libopenh264.so.2.4.1 
a266478faa47e70dae142920e162aea4  /usr/lib64/libopenh264.so.2.4.1
```

Meanwhile:

http://ciscobinary.openh264.org/libopenh264-2.4.1-linux64.7.so.signed.md5.txt -> `19c561386a9564f8510fcb7586b9d402`

I don't know where this file is coming from, or why it's different... :/

torokati44:
Oh...

https://fedoraproject.org/wiki/OpenH264

>  It contains OpenH264 binary built inside the Fedora infrastructure, but distributed by Cisco, so that the all licensing fees are still covered by them.

torokati44:
Never mind, it mattersn't, as discussed on Discord... :upside_down_face:

video/external/src/decoder/openh264_sys.rs
```diff
@@ -1359,11 +1359,11 @@ pub struct OpenH264 {
         Result<unsafe extern "C" fn(pVersion: *mut OpenH264Version), ::libloading::Error>,
 }
 impl OpenH264 {
-    pub unsafe fn new<P>(path: P) -> Result<Self, ::libloading::Error>
+    pub unsafe fn new<P>(filename: P) -> Result<Self, ::libloading::Error>
     where
         P: AsRef<::std::ffi::OsStr>,
```
torokati44, line 1364: Is it just me, or does this feel inconsistent now, with the `P` not following the rename...? :thinking:

video/external/src/decoder/openh264.rs
```diff
@@ -1,13 +1,62 @@
...
 use crate::decoder::openh264_sys::{self, videoFormatI420, ISVCDecoder, OpenH264};
 use crate::decoder::VideoDecoder;
 
 use ruffle_render::bitmap::BitmapFormat;
 use ruffle_video::error::Error;
 use ruffle_video::frame::{DecodedFrame, EncodedFrame, FrameDependency};
+use thiserror::Error;
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct H264Version(u32, u32, u32);
```
torokati44, line 16: I feel like `OpenH264Version` would be a better name, lest it be confused with https://en.wikipedia.org/wiki/Advanced_Video_Coding#Versions .

video/external/src/decoder/openh264.rs
```diff
@@ -1,13 +1,62 @@
...
+
+impl Display for H264Version {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "({},{},{})", self.0, self.1, self.2)?;
```
torokati44, line 20: Why make it look like a tuple, instead of a "normal" version number? :thinking:
I mean, `(2,4,1)` -> `2.4.1`?

video/external/src/decoder/openh264.rs
```diff
@@ -1,13 +1,62 @@
...
+        Ok(())
+    }
+}
+
+#[derive(Debug, Error)]
+pub enum H264Error {
```
torokati44, line 26: Again, since this is also pub, maybe `OpenH264Error`?

video/external/src/decoder/openh264.rs
```diff
@@ -1,13 +1,62 @@
...
+    #[error("OpenH264 version mismatch, expected {0}, was {1}")]
+    VersionMismatchError(H264Version, H264Version),
+}
+
+/// H264 codec representation.
+pub struct H264Codec {
+    openh264: Arc<OpenH264>,
+}
+
+impl H264Codec {
```
torokati44, line 39: The reason I'm so insistent on qualifying everything in here as "OpenH264", is mostly #16794, but also, later, possibly, alternative decoders using Vulkan Video or VA-API, or what have you.

video/external/src/decoder/openh264.rs
```diff
@@ -1,13 +1,62 @@
...
+    const VERSION: H264Version = H264Version(2, 4, 1);
+
+    pub fn new<P>(filename: P) -> Result<Self, H264Error>
+    where
+        P: AsRef<::std::ffi::OsStr>,
```
torokati44, line 44: Again, feels strange to call the type of `filename` `P`, of all letters, when it's not even a `PathBuf`.

video/external/src/backend.rs
```diff
@@ -28,7 +29,7 @@ enum ProxyOrStream {
 /// except for H.264, for which it uses an external decoder.
 pub struct ExternalVideoBackend {
     streams: SlotMap<VideoStreamHandle, ProxyOrStream>,
-    openh264_lib_filepath: Option<PathBuf>,
+    h264_codec: Option<H264Codec>,
```
torokati44, line 38: Yet again, maybe a name more specific to OpenH264?

video/external/src/backend.rs
```diff
@@ -25,6 +25,11 @@ enum ProxyOrStream {
     Owned(VideoStream),
 }
 
+struct OpenH264Data {
+    download_filename: &'static str,
+    download_sha256: &'static str,
+}
```
torokati44: Is the benefit of this purely code readability/clarity, or something more technical too? The former is also a fine reason to have it, just curious!

torokati44: Also, maybe instead of adding the `download_` prefix to both fields, maybe the struct name could carry this information...? But it's not `pub`, so it doesn't matter much...

video/external/src/backend.rs
```diff
@@ -49,6 +50,11 @@ impl ExternalVideoBackend {
         const OS: &str = std::env::consts::OS;
         const ARCH: &str = std::env::consts::ARCH;
 
+        let local_filenames = match OS {
+            "linux" => vec!["libopenh264.so.7", "libopenh264.so.2.4.1", "libopenh264.so"],
+            _ => vec![],
```
torokati44, line 55: I get that it's far less likely to have this library already installed on any other OS - but do we just assume that it's basically impossible, or should at least a `// TODO: investigate other OSes` comment be left here? I don't know, just wondering...

video/external/src/decoder/openh264.rs
torokati44: Although, there's already `openh264_sys::OpenH264Version`. I know it's not `Display` but it`s at least `Eq`...

video/external/src/decoder/openh264_sys.rs
kjarosh: Huh, you're right, but that's how libloading does it:
```rust
pub unsafe fn new<P: AsRef<OsStr>>(filename: P) -> Result<Library, Error> {
```

I could change that to `F`, but in reality it might be either path or filename, so I don't know what's best here. My goal was to match the signature with libloading in order not to suggest that we are supporting paths only.

video/external/src/decoder/openh264.rs
kjarosh: I don't think I should use `openh264_sys::OpenH264Version` as `openh264_sys` is private, I changed the name to `OpenH264Version`

video/external/src/decoder/openh264.rs
kjarosh: :facepalm: of course, I just replaced the tuple with the struct without really thinking about it

video/external/src/decoder/openh264.rs
kjarosh: Done

video/external/src/decoder/openh264.rs
kjarosh: Sure, I just originally wrote the code to be consistent with `H264Decoder`

video/external/src/backend.rs
kjarosh: Done

video/external/src/backend.rs
kjarosh: This change is a bit future-oriented, as I will add `local_filenames` to this struct in the next commit, and that's the real motivation of this struct :sweat_smile:

video/external/src/backend.rs
kjarosh: Of course, the reason I added a `match` here was to make it easy to add support for other OSes, it's just that I have no idea what names I should use or how OpenH264 is installed there.

(Added the TODO)

video/external/src/decoder/openh264_sys.rs
torokati44: Being consistent with `libloading` is fine then.

video/external/src/decoder/openh264.rs
torokati44: Resolving for the same reason as above, consistency with `libloading`.

video/external/src/decoder/openh264.rs
torokati44: It could still be `pub use`'d, no? Anyway, if you decided not to use it, that's okay, I just wanted to make sure it's a deliberate thing. Thanks for the rename!

torokati44 (approved):
LGTM, thank you very much! :partying_face: 

Now, to rebase #16794 again, on top of this... :weary:

kjarosh:
Just replaced one more occurrence of `h264` -> `openh264` in a variable name

PR #17803, opened 2024-09-07
video: Use cache dir for storing downloaded OpenH264 library

kjarosh:
Using the executable's parent directory is a bad idea in most cases, as the place where Ruffle is installed should not be modified (often it's not even writable). This PR changes the directory where the OpenH264 library is saved to the cache directory, which should always be writable.


tests/framework/src/options.rs
```diff
@@ -177,8 +177,11 @@ impl PlayerOptions {
         if self.with_video {
             #[cfg(feature = "ruffle_video_external")]
             {
+                let current_exe = std::env::current_exe()?;
+                let directory = current_exe.parent().expect("Executable parent dir");
```
torokati44, line 181: Huh, why still use this? :thinking:

torokati44: Ah, this is the test framework, there is no preferences here...

torokati44 (approved):
LGTM, thanks!

PR #17804, opened 2024-09-07
desktop: Install Ruffle metadata in PKGBUILD

kjarosh:
This most importantly adds a way of opening Ruffle using GUI.

Tested on Manjaro GNOME, would appreciate if someone else also tested it!


torokati44 (approved):
I say let's go ahead with this. It works on at least one system, it's a small enough change, it'll be easier to test once it's in, and can always be fixed/reverted later.

PR #17810, opened 2024-09-07
build: Freeze rustc for nightly builds to 1.80.1

adrian17:
Note: not tested.

PR #17812, opened 2024-09-07
Revert "build: Freeze rustc for nightly builds to 1.80.1"

torokati44:
Reverts ruffle-rs/ruffle#17810

Merge once the Vector.sort situation is _sorted out_.

### DIFF
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -108,9 +108,9 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install Rust toolchain
-        uses: dtolnay/rust-toolchain@master
+        uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: 1.80.1
+          toolchain: stable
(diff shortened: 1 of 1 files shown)


torokati44:
Related: https://github.com/ruffle-rs/ruffle/issues/17785 and https://github.com/rust-lang/rust/issues/129561#issuecomment-2336413186

danielhjacobs:
Rebased. This is probably good to merge, but it should be tested to make sure https://github.com/ruffle-rs/ruffle/issues/17785 works properly still.

torokati44:
I actually can't reproduce that even before https://github.com/ruffle-rs/ruffle/pull/17846... :/ On desktop at least.

danielhjacobs:
Just to be sure, though, I assume so, you're testing with building with rust 1.81, right?

torokati44:
Of course. The AVM trace is really helpful in suggesting that this is supposed to happen near the "end game screen", but for me, it didn't.

adrian17 (approved):

PR #17817, opened 2024-09-08
core/avm2: Remove some unnecessary warning logs

Lord-McSweeney:


moulins:
Why not keep the logs as `tracing::info!`? You never know when these could be useful.

evilpie:
Why not change it to `tracing::debug` or `tracing::trace` instead?

evilpie:
Haha collision. `tracing::info` is still our default log level, so we probably shouldn't use that.

Lord-McSweeney:
Seems like this was worked out over Discord.

PR #17834, opened 2024-09-09
avm2: Cache method body index on Method

adrian17:
This fixes quadratic behavior observable with >>1000 classes. On Habbo SWF, on desktop it shaves like 100ms and on FF it improves load time by 0.5-1s.


swf/src/avm2/read.rs
```diff
@@ -67,8 +67,10 @@ impl<'a> Reader<'a> {
 
         let len = self.read_u30()?;
         let mut method_bodies = Vec::with_capacity(len as usize);
-        for _ in 0..len {
-            method_bodies.push(self.read_method_body()?);
+        for body_idx in 0..len {
+            let body = self.read_method_body()?;
+            methods[body.method.0 as usize].body = Some(Index::new(body_idx));
```
Lord-McSweeney, line 72: If FP verifyerrors here when there is already a Method attached to the MethodBody can we check and panic?

Lord-McSweeney (approved):

swf/src/avm2/read.rs
adrian17: Oh, this is in `swf`, so can't really log from there either...
And we ideally shouldn't panic in `swf` as it's more of a library :/
Gonna `Err` then.

PR #17836, opened 2024-09-09
avm2: Remove Copy impl and Gc field from PropertyClass

Aaron1011:
We now produce return a new PropertyClass instance
(instead of updating in-place and returning `true`) in order
to cache class name lookups. I've also adjusted the call sites to
ensure that we're always storing an resolved `PropertyClass` back
in the vtable.

To avoid growing the property maps, I've replaced the `Gc` with a `Box` in the less-common `PropertyClass::Name` variant


Lord-McSweeney:
We should be able to keep `Copy` and remove the `Gc` entirely, if that would be preferable, now that the Multiname is behind a Gc

Lord-McSweeney:
Superseded by #17890.

PR #17846, opened 2024-09-10
AVM2: Implement avmplus' sort algorithm instead of using the standard Rust sort.

moulins:
Replace the use of Rust's standard sort in `Array.sort` and `sortOn` by a port of avmplus' QuickSort algorithm. This avoid panics on Rust >=1.81 when `Array.sort` is called with a non-Ord comparison function, and will always produce the same result as Flash Player.

This also adds a test for AVM1 and AVM2 for "sorting" an array with a random comparison function, that check both the sorted result and the exact sequence of comparisons done by the algorithm. Currently, this test only passes for AVM2; I wasn't able to find which modifications to apply to the AVM1 quicksort to produce the correct results.


Aaron1011 (approved):

adrian17 (approved):

Voultapher:
I wonder if this change is a good fit for running arbitrary user-code. It's trivial to hit quadratic run-time, e.g:

```rust
let len = 1_000_000;
let mut v = (0..len as i32).map(|val| val % 2).collect::<Vec<i32>>();
let _ = qsort(&mut v, &mut |a, b| -> Result<Ordering, ()> { Ok(a.cmp(b)) });
```

I guess if you are going for bug-for-bug compatibility with the avmplus code this is the correct choice.

PR #17866, opened 2024-09-11
avm2: Return actual VerifyErrors for constant pool errors

Lord-McSweeney:
This requires passing around an Activation in some places. The PR also adds better verification for `Class` and slot names.

`Any` namespaces can now only be found in Multinames with a `NamespaceSet::Single`.

### DIFF
--- a/core/src/avm2/multiname.rs
+++ b/core/src/avm2/multiname.rs
@@ -1,19 +1,18 @@
 use crate::avm2::activation::Activation;
+use crate::avm2::error::{make_error_1032, make_error_1080};
 use crate::avm2::script::TranslationUnit;
 use crate::avm2::Error;
 use crate::avm2::Namespace;
 use crate::avm2::QName;
 use crate::avm2::{Object, Value};
-use crate::context::{GcContext, UpdateContext};
+use crate::context::GcContext;
 use crate::string::{AvmString, WStr, WString};
 use bitflags::bitflags;
 use gc_arena::Gc;
 use gc_arena::{Collect, Mutation};
 use std::fmt::Debug;
 use std::ops::Deref;
-use swf::avm2::types::{
-    AbcFile, Index, Multiname as AbcMultiname, NamespaceSet as AbcNamespaceSet,
-};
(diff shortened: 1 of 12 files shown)


core/src/avm2/multiname.rs
```diff
@@ -163,33 +160,57 @@ impl<'gc> Multiname<'gc> {
...
+        }
+
         let abc = translation_unit.abc();
-        let abc_multiname = Self::resolve_multiname_index(&abc, multiname_index)?;
+
+        let abc_multiname = abc
+            .constant_pool
+            .multinames
+            .get(multiname_index.0 as usize - 1)
+            .ok_or_else(|| format!("Unknown multiname constant {}", multiname_index.0))?;
```
adrian17, line 204: there's no corresponding error in FP?

core/src/avm2/qname.rs
```diff
@@ -40,52 +42,36 @@ impl<'gc> QName<'gc> {
...
-                            .into(),
-                    }
-                } else {
-                    return Err(
-                        "Attempted to pull QName from multiname with multiple namespaces".into(),
-                    );
-                }
-            }
-            _ => return Err("Attempted to pull QName from non-QName multiname".into()),
+        // For any non-QName not in the playerglobals domain, return an error.
```
adrian17, line 61: When can this happen?
Or is this avmplus logic without an example?

core/src/avm2/multiname.rs
Lord-McSweeney: There is, I'm going to fix all the remaining constant pool errors in a follow-up PR (should just be replaced with `make_error_1032`, like in `verify.rs`).

core/src/avm2/qname.rs
Lord-McSweeney: This is from avmplus, but when not implementing it, playerglobal failed to load- it seems that ASC incorrectly generates single-namespace `Multiname`s for some declarations that could just be `QName`s.

adrian17 (approved):

PR #17873, opened 2024-09-12
tests: Test exposed RufflePlayer API

evilpie:
This test ensures we notice when accidentally adding or removing something from the public RufflePlayer API.


danielhjacobs:
displayMessage was made public once more so that should be tested here too. We should avoid adding anything else if possible until `Player.ruffle()` exists though.

PR #17875, opened 2024-09-12
desktop: Add filesystem access dialog

kjarosh:
This is an attempt to address the issue of movies having unrestricted filesystem access.

When a movie wants to access a local file, a dialog is shown which asks for permission. They user may allow access to the file or to one of the parent directories.

<image src="https://github.com/user-attachments/assets/52a961d2-9b8c-449d-b6e7-112ba3129331" width="500"/>

This behavior can be changed with:
* `--filesystem-access-mode allow`
* `--filesystem-access-mode deny`
* `--filesystem-access-mode ask` (default)

Comments and suggestions are welcome!

### DIFF
--- a/desktop/assets/texts/en-US/filesystem_access_dialog.ftl
+++ b/desktop/assets/texts/en-US/filesystem_access_dialog.ftl
@@ -1,0 +1,7 @@
+filesystem-access-dialog-title = Requesting Filesystem Access
+
+filesystem-access-dialog-message = The current movie is attempting to read the following file. Do you want to allow it?
+filesystem-access-dialog-allow-remember-message = Do not ask again for files from the following directory for the duration of this session:
+
+filesystem-access-dialog-allow = Allow
+filesystem-access-dialog-allow-remember = Allow and Remember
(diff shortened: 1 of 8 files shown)


desktop/assets/texts/en-US/filesystem_access_dialog.ftl
```diff
@@ -0,0 +1,7 @@
+filesystem-access-dialog-title = Requesting Filesystem Access
+
+filesystem-access-dialog-message = The current movie is attempting to access the following file. Are you sure you want to allow it?
```
Dinnerbone, line 3: nit: Change "Are you sure you want to allow it?" to "Do you want to allow it?"

The former assumes intent and feels like you're questioning the users intent, rather than asking the user what their intent is.


(perhaps change "access" to "read", too? These are read-only operations, right?)

torokati44: (If so, I suppose, the network/web permission dialogs should also be updated to be consistent.)

kjarosh: Done, and yes, this is read-only access to the file (unless we implement AIR API)

kjarosh: Network access / open url dialogs are updated here: https://github.com/ruffle-rs/ruffle/pull/18066

adrian17 (approved):

PR #17876, opened 2024-09-12
AVM2: Intern all (non-private) namespaces

moulins:
This adds an interner for `Namespace` objects (stored in the new `CommonNamespaces` struct), and use it to intern all namespaces (except private ones). This allows using a simple pointer comparison to implement namespace equality.

The PR is split in roughly three parts:
1. Extract the `WeakSet` implementation used in the existing string interner into a new `utils` modules; ~~also move `core::tag_utils` here for consistency~~.
2. Add a new `CommonNamespaces` struct that will hold the interner (but not yet);  mirroring #17842, move the cached namespaces on `Activation` to it.
3. Finally, add the namespace interner to `CommonNamespaces` and use it.


moulins:
Closing this; interning namespaces doesn't seem to bring significant speed increases or memory savings, so the extra complexity isn't warranted.

PR #17878, opened 2024-09-12
video/external: Make OpenH264 support optional at build time

torokati44:
This is the first (more ready and conflicty, but less risky and controversy) part of https://github.com/ruffle-rs/ruffle/pull/16794 split out. The rest can be more or less additive after this, once I finally get back to it.

I know that `external_video` itself is already a blanket feature that disables this; and having that enabled, but `openh264` not, is somewhat silly, but ... bear with me? :grinning:

### DIFF
--- a/video/external/src/backend.rs
+++ b/video/external/src/backend.rs
@@ -1,3 +1,4 @@
+#[cfg(feature = "openh264")]
 use crate::decoder::openh264::OpenH264Codec;
 use crate::decoder::VideoDecoder;
 
@@ -26,21 +27,42 @@ enum ProxyOrStream {
 /// except for H.264, for which it uses an external decoder.
 pub struct ExternalVideoBackend {
     streams: SlotMap<VideoStreamHandle, ProxyOrStream>,
+    #[cfg(feature = "openh264")]
     openh264_codec: Option<OpenH264Codec>,
     software: SoftwareVideoBackend,
 }
 
 impl Default for ExternalVideoBackend {
     fn default() -> Self {
-        Self::new(None)
+        Self::new()
     }
 }
 
 impl ExternalVideoBackend {
-    pub fn new(openh264_codec: Option<OpenH264Codec>) -> Self {
+    fn make_decoder(&mut self) -> Result<Box<dyn VideoDecoder>, Error> {
+        #[cfg(feature = "openh264")]
+        if let Some(h264_codec) = self.openh264_codec.as_ref() {
+            let decoder = Box::new(crate::decoder::openh264::H264Decoder::new(h264_codec));
+            return Ok(decoder);
+        }
+
+        Err(Error::DecoderError("No OpenH264".into()))
+    }
+
+    pub fn new() -> Self {
         Self {
             streams: SlotMap::with_key(),
-            openh264_codec,
+            #[cfg(feature = "openh264")]
+            openh264_codec: None,
+            software: SoftwareVideoBackend::new(),
+        }
+    }
+
+    #[cfg(feature = "openh264")]
+    pub fn new_with_openh264(openh264_codec: OpenH264Codec) -> Self {
+        Self {
+            streams: SlotMap::with_key(),
+            openh264_codec: Some(openh264_codec),
             software: SoftwareVideoBackend::new(),
         }
     }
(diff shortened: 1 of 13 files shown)


video/external/src/decoder/openh264.rs
```diff
@@ -67,7 +73,136 @@ pub struct H264Decoder {
...
+            return Err(format!(
+                "SHA256 checksum mismatch for {filename}; expected {sha256sum}, found {sha256digest:x} (with a size of {size} bytes)",
+            )
+            .into());
+        }
+
+        Ok(filepath)
+    }
+
+    pub fn load_openh264(directory: &Path) -> Result<OpenH264Codec, Box<dyn std::error::Error>> {
```
kjarosh, line 184: I believe these methods would fit better in `OpenH264Codec`, this one for instance would be a constructor

video/external/src/backend.rs
```diff
@@ -25,154 +24,52 @@ enum ProxyOrStream {
...
-            (os, arch) => return Err(format!("Unsupported OS/arch: {}/{}", os, arch).into()),
-        };
-
-        Ok(OpenH264Data {
-            local_filenames,
-            download_filename,
-            download_sha256,
-        })
+    #[cfg(feature = "openh264")]
+    pub fn load_openh264(directory: &Path) -> Result<OpenH264Codec, Box<dyn std::error::Error>> {
```
kjarosh, line 44: Hmmm... Do we need this method? I think we can call `load_openh264` directly from `H264Decoder`/`OpenH264Codec`.

video/external/src/backend.rs
```diff
@@ -25,154 +24,52 @@ enum ProxyOrStream {
...
-            let size = filepath.metadata().map(|f| f.len()).unwrap_or_default();
-            return Err(format!(
-                "SHA256 checksum mismatch for {filename}; expected {sha256sum}, found {sha256digest:x} (with a size of {size} bytes)",
-            )
-            .into());
-        }
-
-        Ok(filepath)
+        #[allow(unreachable_code)]
+        return Err(Error::DecoderError("No OpenH264".into()));
```
kjarosh, line 56:
```suggestion
        Err(Error::DecoderError("No OpenH264".into()))
```

video/external/src/backend.rs
torokati44: The reason I put this "proxy" here, is that I don't _really_ want to make much more than `ExternalVideoBackend` public. So, _explicitly_ reaching in from `desktop/player.rs` (incl. conditional compilation) right to `openh264.rs` would seem maybe a bit strange to me...? Even though the (new) `OpenH264Codec` type (now) has to be used implicitly out there.

But maybe I'll see it differently once I get some sleep.

video/external/src/backend.rs
torokati44: Right, nice, thanks! In some earlier incarnation of this code, this wasn't possible without clippy freaking out.

video/external/src/decoder/openh264.rs
torokati44: Good point, moved them there. I still need to bikeshed the names and their order. Any suggestions?

video/external/src/backend.rs
kjarosh: I see that, that's okay of course, but it feels strange having openh264 module pub in that case. 

Although I would recommend making the OpenH264 constructor public (and only that, the rest pub(crate) if needed), as OpenH264 is a bit of an exception here—we need to download it, and make all kinds of weird things with it. So it makes sense that `ExternalVideoBackend` does not want to have to do anything with instantiating OpenH264 and just wants the instance, moving the burden of instantiating it to the caller; compared to acting as a proxy to first instantiate it, and then require to pass the instantiated object. The current flow that we have fits more to the former scenario IMO.

So I'm leaving the decision to you, just sharing my thoughts.

video/external/src/decoder/openh264.rs
kjarosh: IMO names are okay (`get_openh264_data`, `download_openh264`, `load_openh264`, only one of them is pub anyways), I don't see how we can improve here. The current order is bottom-up, you could reverse it to obtain top-down order, but I'm not sure if we have any preference here.

torokati44: > IMO names are okay `get_openh264_data`, `download_openh264`, `load_openh264`

The current names are not these - but since they are within `OpenH264Codec`, and as you said, mostly not `pub`, do you still think they should contain `openh264`?

kjarosh: Nope, they don't have to :) it's all good

video/external/src/backend.rs
torokati44: Moved decoder instantiation out to caller, using the single public "constructor", and removed the proxy method from the backend implementation.

kjarosh (approved):

torokati44:
Thanks for the review! ^^

kjarosh:
Thank YOU for working on H264 ;)

PR #17885, opened 2024-09-12
chore: Update translations

RuffleBuild:
New Crowdin pull request with translations


torokati44:
I have manually pasted every translation into Google Translate, and all of them seemed to make sense when turned back into English. I don't think there's much more review that we could do, especially for the Arabic texts.

PR #17890, opened 2024-09-13
avm2: Correctly differentiate between Any and null Multiname

Lord-McSweeney:
For example, we now throw a class lookup VerifyError ("`Class *::* could not be found`") when trying to call a method with return type `QName(null, null)` instead of incorrectly assuming that the name is the Any name.

This removes the `any` Multiname from `CommonMultinames`.


Lord-McSweeney:
The last commit in this PR supersedes the Gc-removal part of #17836.

adrian17 (approved):

PR #17905, opened 2024-09-14
desktop: Do not store logs in config directory

kjarosh:
Logs are non-essential and can be removed, contrary to config files. This PR changes the log directory from config dir to cache dir.

| Platform | Old Path | New Path |
| -------- | -------- | -------- |
| Linux    | `~/.config/ruffle/*.log` | `~/.cache/ruffle/log/*.log` |
|macOS     | `~/Library/Application Support/ruffle/*.log` | `~/Library/Caches/ruffle/log/*.log` |
| Windows  | `C:\Users\Alice\AppData\Local\ruffle\*.log` | `C:\Users\Alice\AppData\Local\ruffle\log\*.log` |

Old log files will be migrated to the new path on startup.

### DIFF
--- a/desktop/src/main.rs
+++ b/desktop/src/main.rs
@@ -160,6 +160,9 @@ async fn main() -> Result<(), Error> {
     if let Some(parent) = log_path.parent() {
         std::fs::create_dir_all(parent)?;
     }
+    if let Err(err) = migrate_logs(&preferences, logs_path) {
+        tracing::warn!("Failed to migrate logs: {}", err);
+    }
 
     // [NA] `_guard` cannot be `_` or it'll immediately drop
     // https://docs.rs/tracing-appender/latest/tracing_appender/non_blocking/index.html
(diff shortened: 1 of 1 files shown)


Dinnerbone:
I like it, my only nit is why `log` and not `logs`?

kjarosh:
> why `log` and not `logs`?

I guess the reason is the subconscious desire to stick to conventions used in the Filesystem Hierarchy Standard: `/var/log` not `/var/logs`, `/usr/include` not `/usr/includes`, `/usr/lib` not `/usr/libs`, etc.

Dinnerbone (approved):

kjarosh:
> _Some say_ that logs rather belong to `$XDG_STATE_HOME`

I know that there's no obvious place to store log files on Linux (except `/var/log` for system services). Usually you'd use `syslog` for the purpose of persisting logs, but I'm not sure about flatpak though.

However, the goal of this PR is not to find the best place to persist logs, but rather to remove them from the obviously wrong place. If they are stored in the config dir, the user has no way of removing them (except manually knowing which files may be removed). That will get exceptionally bad if someone chooses "with timestamp" which will make the config directory grow indefinitely in size.

> The migration part is a nice touch! Although, maybe if the destination file already exists, it shouldn't be overwritten?

I don't think that matters much, logs are migrated before we even try to write to the new directory, and the goal of the migration is not strictly to preserve the logs, but rather to get them out of the config dir.

The file `ruffle.log` is truncated on start anyway and overwriting logs with timestamps is virtually impossible, unless you're Flash (_the other one_) and you have an exceptionally mischievous computer.

torokati44:
Thank you! ^^

PR #17909, opened 2024-09-14
desktop: Update Freedesktop categories

kjarosh:
1. Change order of categories. — Apparently the order matters when desktop environments decide to include the app only in one category. It seems that we should list more important categories first.
2. Add `VectorGraphics` Freedesktop category — Ruffle handles SWF files, which are generally vector graphic animations.

### DIFF
--- a/desktop/packages/linux/rs.ruffle.Ruffle.desktop
+++ b/desktop/packages/linux/rs.ruffle.Ruffle.desktop
@@ -6,5 +6,5 @@ Comment=Play Flash games & movies
 Icon=rs.ruffle.Ruffle
 Exec=ruffle %u
 MimeType=application/x-shockwave-flash;application/vnd.adobe.flash.movie
-Categories=Game;Utility;Graphics;AudioVideo;Player;Viewer
+Categories=AudioVideo;Player;Graphics;Viewer;Utility;Game
 Keywords=flash;swf;player;emulator;rust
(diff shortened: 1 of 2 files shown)


adrian17 (approved):

evilpie:
Is Audio/Video really the best category? Flash is still primarily interactive and not just playback.

kjarosh:
I considered the following main categories [1]:
1. AudioVideo
2. Graphics
3. Utility
4. Game

Ad 4. it's hard to argue that the main use of Ruffle is gaming, Ruffle supports far wider content than games.

Ad 3. it seems that Utility is a category that includes random utilities (basic text editors, tools related to files, archiving & compression) [2]. I don't think this category describes what Ruffle does well, it can be argued whether Ruffle even fits in this category.

Ad 2. Graphics generally assumes non-interactive, static multimedia. Main content in Ruffle is interactive and dynamic.

Ad 1. AudioVideo is for "[...] presenting, creating, or processing multimedia". Granted, it does not mention that the multimedia may be interactive, but it also does not say or imply it's not.

I agree that there's no category that describes what Ruffle does in 100%, but I don't see a better main main (?) category than AudioVideo.

[1] https://specifications.freedesktop.org/menu-spec/latest/category-registry.html#main-category-registry
[2] https://specifications.freedesktop.org/menu-spec/latest/additional-category-registry.html

evilpie:
You are definitely correct that there doesn't seem to be any category that is a perfect fit. Just terminology wise, and not directly based on the description, the two interesting ones are: _Viewer_ is a sub-category of Graphics/Office and _Player_ of Audio/Video. So, I guess I can agree with your ordering. Looking at the other examples (Clock and Calculator) and the description, we might actually want to remove the _Utility_ category completely.

evilpie (approved):

kjarosh:
I'll do a separate PR for the removal of _Utility_ so we can have a discussion there.

PR #17916, opened 2024-09-15
 avm2: Fix String.replace crash when regex is used in replacement 

### LINKED ISSUE #17899: String.replace is not re-entrant on a used regular expression
### Describe the bug

```
package {
        public class Test {}
}

var str: String = "Hello, this is a simple example.";
var pattern: RegExp = /simple/;

trace("Starting...");

var replaced: String = str.replace(pattern, function(contents) {
    return contents.replace(pattern, "complicated");
});

trace("Original String: " + str);
trace("Modified String: " + replaced);
```

The above script crashes due to `replace`'s implementation doing a mutable borrow on the regular expression object, and holding onto that during the replacement operation.

### Expected behavior

The following trace
```
Starting...
Original String: Hello, this is a simple example.
Modified String: Hello, this is a complicated example.
```

### Content Location

The above AS3 snippet compiled with Apache Flex (attached the SWF for reference)

[test.swf.zip](https://github.com/user-attachments/files/17000298/test.swf.zip)

### Affected platform

Desktop app

### Operating system

macOS Sonoma (14.6)

### Browser

_No response_

### Additional information

_No response_

rtpg:
This scopes the locking a bit down to avoid holding a lock on the
 underlying RegExp object when calling a callback in AS3-land.

Fixes #17899, though would appreciate guidance on how to add the test SWF into the project.


lvyitian:
Test guide: https://github.com/ruffle-rs/ruffle/blob/master/tests/README.md

torokati44:
Thank you, and welcome to the project! Please take a look at https://github.com/ruffle-rs/ruffle/blob/master/CONTRIBUTING.md, and tag/label your commit message and PR accordingly - for this one, I'd suggest `avm2`.

core/src/avm2/regexp.rs
```diff
@@ -214,12 +214,12 @@ impl<'gc> RegExp<'gc> {
     /// Implements string.replace(regex, replacement) where the replacement is
     /// a function.
     pub fn replace_fn(
-        &mut self,
+        regexp: &Object<'gc>,
```
rtpg, line 217: This change is to allow for finer-grained locking when calling back into AS code

rtpg:
Thanks for the guidance (and sorry for not reading the contributing guidelines!), I modified the existing `string_replace` test suite to crash with the existing locking behavior, and now it is should be good for a review IMO

Dinnerbone (approved):
Thank you!

core/src/avm2/globals/string.rs
```diff
@@ -356,17 +356,26 @@ fn replace<'gc>(
     let pattern = args.get(0).unwrap_or(&Value::Undefined);
     let replacement = args.get(1).unwrap_or(&Value::Undefined);
     // Handles regex patterns.
-    if let Some(mut regexp) = pattern
+    // doing this a bit awkwardly to avoid holding a lock on the underlying
+    // regexp (if it exists)...
+    let is_regexp = pattern
```
adrian17, line 361: I think there's one slightly cleaner (less unwraps) way to write it:
```rs
let regexp = pattern.as_object().and_then(|o| o.as_regexp_object());

if let Some(regexp) = regexp {
    // ...
```

Here, `regexp` becomes `RegExpObject`, so you can pass it to `RegExp::replace_*` methods already type-checked instead of generic `Object`.

core/src/avm2/regexp.rs
```diff
@@ -241,21 +241,22 @@ impl<'gc> RegExp<'gc> {
     /// Implements string.replace(regex, replacement) where the replacement may be
     /// a string with $-sequences.
     pub fn replace_string(
-        &mut self,
+        regexp: &Object<'gc>,
```
adrian17, line 244: Let's pass `Object` by value, it's just a GC pointer.
(`RegExpObject` with the suggestion above)

rtpg: So I changed this up to `RegExpObject`, but I still have to get to the underlying data  with `regexp.as_regexp_mut(activation.gc()).unwrap()` yeah? I'm not missing some obvious `impl` that can lean on knowing it's a `RegExpObject` a bit more?

Obviously having the signature be more precise here is nicer overall though.

core/src/avm2/globals/string.rs
rtpg: Changed this up to do as such!

core/src/avm2/regexp.rs
adrian17: Yeah. We could add a new method that returns a non-Option (in fact, I'd claim `as_regexp_mut` being defined on `Object` is a bit of an anti-pattern), but there isn't right now. (and it's not critical to add in this PR specifically)

rtpg:
alright I believe this is good for another review

PR #4813, opened 2021-07-09
web: Add a loadeddata event, which triggers when loadedmetadata does for now

danielhjacobs:
This can be used by someone who expects it to indicate when the swf is fully loaded, without fear that that functionality will change in the future (which means it will need to be changed when streaming support is added). Based on https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/loadeddata_event.

**Edit:** This builds on top of #4084, so there are some important things to note. First, that PR said RufflePlayer.readyState was supposed to mimic the media APIs. However, https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/readyState has 5 possible values, not just 3. I'm not saying Ruffle should duplicate that, but loadeddata triggers when the ready state is HAVE_CURRENT_DATA for media elements, so it may not be the best name for the event for the RufflePlayer, since people familiar with the media APIs will be confused by it.

I''m not sure what the best event name would be.

For media elements, loadeddata triggers when the ReadyState is 2 (HAVE_CURRENT_DATA). canplay triggers, I believe, when the ReadyState is 3 (HAVE_FUTURE_DATA), and canplaythrough triggers, I believe, when the ReadyState is 4 (HAVE_ENOUGH_DATA).

For RufflePlayer, the ReadyState can hold the values 0 (HAVE_NOTHING), 1 (LOADING), or 2 (LOADED).

PR #4084, opened 2021-04-22
web: Expose SWF metadata to JavaScript

Herschel:
* Adds `RufflePlayer.metadata` to the JS API that exposes the SWF header data to JavaScript.
 * Add `RufflePlayer.readyState` and fire a `loadedmetadata` event once the metadata is available, mimicking the HTML5 media APIs.
 * Add some some very basic tests for the JS API to test loading a movie and grabbing the metadata.

PR #17939, opened 2024-09-15
web: Add `ruffle()` (Version 1) API

Dinnerbone:
Bikeshedding time, here's the chance to decide what the "v1 api" looks like before we lock it in.

The only change so far is that I've added a specific method for calling EI, vs the very easy-to-break and implicit "the method will exist, trust me". I'm awaiting feedback before touching anything else.

Whilst this is "version 1" (`PlayerV1` interface), if we never add a V2 that's totally fine. `ruffle()` is the same as `ruffle(1)`.

Things to consider for each existing member of the API:
- Is anything a property that should be a method?
- Is anything a method that should be a property?
- Should we rename anything, or change arguments?

### DIFF
--- a/web/packages/core/src/internal/player/impl_v1.ts
+++ b/web/packages/core/src/internal/player/impl_v1.ts
@@ -1,0 +1,105 @@
+import { PlayerV1 } from "../../public/player";
+import { InnerPlayer, ReadyState } from "./inner";
+import type { DataLoadOptions, URLLoadOptions } from "../../public/config";
+import type { MovieMetadata } from "../../public/player";
+
+export class PlayerV1Impl implements PlayerV1 {
+    #inner: InnerPlayer;
+
+    constructor(inner: InnerPlayer) {
+        this.#inner = inner;
+    }
+
+    get onFSCommand(): ((command: string, args: string) => boolean) | null {
+        return this.#inner.onFSCommand;
+    }
+
+    set onFSCommand(
+        value: ((command: string, args: string) => boolean) | null,
+    ) {
+        this.#inner.onFSCommand = value;
+    }
+
+    get readyState(): ReadyState {
+        return this.#inner._readyState;
+    }
+
+    get metadata(): MovieMetadata | null {
+        return this.#inner.metadata;
+    }
+
+    get loadedConfig(): URLLoadOptions | DataLoadOptions | null {
+        return this.#inner.loadedConfig ?? null;
+    }
+
+    async reload(): Promise<void> {
+        await this.#inner.reload();
+    }
+
+    async load(
+        options: string | URLLoadOptions | DataLoadOptions,
+        isPolyfillElement: boolean = false,
+    ): Promise<void> {
+        await this.#inner.load(options, isPolyfillElement);
+    }
+
+    play(): void {
+        this.#inner.play();
+    }
+
+    get isPlaying(): boolean {
+        return this.#inner.isPlaying;
+    }
+
+    get volume(): number {
+        return this.#inner.volume;
+    }
+
+    set volume(value: number) {
+        this.#inner.volume = value;
+    }
+
+    get fullscreenEnabled(): boolean {
+        return this.#inner.fullscreenEnabled;
+    }
+
+    get isFullscreen(): boolean {
+        return this.#inner.isFullscreen;
+    }
+
+    setFullscreen(isFull: boolean): void {
+        this.#inner.setFullscreen(isFull);
+    }
+
+    enterFullscreen(): void {
+        this.#inner.enterFullscreen();
+    }
+
+    exitFullscreen(): void {
+        this.#inner.exitFullscreen();
+    }
+
+    async downloadSwf(): Promise<void> {
+        await this.#inner.downloadSwf();
+    }
+
+    displayMessage(message: string): void {
+        this.#inner.displayMessage(message);
+    }
+
+    pause(): void {
+        this.#inner.pause();
+    }
+
+    set traceObserver(observer: ((message: string) => void) | null) {
+        this.#inner.traceObserver = observer;
+    }
+
+    get config(): URLLoadOptions | DataLoadOptions | object {
+        return this.#inner.config;
+    }
+
+    set config(value: URLLoadOptions | DataLoadOptions | object) {
+        this.#inner.config = value;
+    }
+}
--- a/web/packages/core/src/internal/player/impl_v1.ts
+++ b/web/packages/core/src/internal/player/impl_v1.ts
@@ -10,14 +10,8 @@ export class PlayerV1Impl implements PlayerV1 {
         this.#inner = inner;
     }
 
-    get onFSCommand(): ((command: string, args: string) => boolean) | null {
-        return this.#inner.onFSCommand;
-    }
-
-    set onFSCommand(
-        value: ((command: string, args: string) => boolean) | null,
-    ) {
-        this.#inner.onFSCommand = value;
+    addFSCommandHandler(handler: (command: string, args: string) => void) {
+        this.#inner.addFSCommandHandler(handler);
     }
 
     get readyState(): ReadyState {
--- a/web/packages/core/src/internal/player/impl_v1.ts
+++ b/web/packages/core/src/internal/player/impl_v1.ts
@@ -102,4 +102,8 @@ export class PlayerV1Impl implements PlayerV1 {
     set config(value: URLLoadOptions | DataLoadOptions | object) {
         this.#inner.config = value;
     }
+
+    callExternalInterface(name: string, ...args: unknown[]): unknown {
+        return this.#inner.callExternalInterface(name, args);
+    }
 }
--- a/web/packages/core/src/internal/player/impl_v1.ts
+++ b/web/packages/core/src/internal/player/impl_v1.ts
@@ -71,7 +71,7 @@ export class PlayerV1Impl implements PlayerV1 {
         this.#inner.setFullscreen(isFull);
     }
 
-    enterFullscreen(): void {
+    requestFullscreen(): void {
         this.#inner.enterFullscreen();
     }
 
--- a/web/packages/core/src/internal/player/impl_v1.ts
+++ b/web/packages/core/src/internal/player/impl_v1.ts
@@ -43,7 +43,7 @@ export class PlayerV1Impl implements PlayerV1 {
         await this.#inner.load(options, isPolyfillElement);
     }
 
-    play(): void {
+    resume(): void {
         this.#inner.play();
     }
 
@@ -87,8 +87,12 @@ export class PlayerV1Impl implements PlayerV1 {
         this.#inner.displayMessage(message);
     }
 
-    pause(): void {
+    suspend(): void {
         this.#inner.pause();
+    }
+
+    get suspended(): boolean {
+        return !this.#inner.isPlaying;
     }
 
     set traceObserver(observer: ((message: string) => void) | null) {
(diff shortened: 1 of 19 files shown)


web/packages/core/src/internal/player/impl_v1.ts
```diff
@@ -0,0 +1,109 @@
...
+import type { MovieMetadata } from "../../movie-metadata";
+
+export class PlayerV1Impl implements PlayerV1 {
+    #inner: InnerPlayer;
+
+    constructor(inner: InnerPlayer) {
+        this.#inner = inner;
+    }
+
+    get onFSCommand(): ((command: string, args: string) => boolean) | null {
```
Dinnerbone, line 13: 3 possibilities here:
1. As today, we set **the** callback for **all** FS commands.
2. We change it to register **a** callback for **all** FS commands.
3. We change it to register **the** callback for **a specific named** FS command.

I think I kinda prefer 2.

web/packages/core/src/internal/player/impl_v1.ts
```diff
@@ -0,0 +1,109 @@
...
+        return this.#inner.onFSCommand;
+    }
+
+    set onFSCommand(
+        value: ((command: string, args: string) => boolean) | null,
+    ) {
+        this.#inner.onFSCommand = value;
+    }
+
+    get readyState(): ReadyState {
```
Dinnerbone, line 23: We should probably use an actually public enum, maybe string based

web/packages/core/src/public/v1.ts
```diff
@@ -0,0 +1,173 @@
...
+export interface PlayerV1 {
+    /**
+     * A movie can communicate with the hosting page using fscommand
+     * as long as script access is allowed.
+     *
+     * @param command A string passed to the host application for any use.
+     * @param args A string passed to the host application for any use.
+     * @returns True if the command was handled.
+     */
+    onFSCommand: ((command: string, args: string) => boolean) | null;
```
Dinnerbone, line 14: 3 possibilities here:

As today, we set the callback for all FS commands.
We change it to register a callback for all FS commands.
We change it to register the callback for a specific named FS command.

I think I kinda prefer 2.

web/packages/core/src/public/v1.ts
```diff
@@ -0,0 +1,173 @@
...
+     * If no such call has been made, this will be `null`.
+     */
+    readonly loadedConfig: URLLoadOptions | DataLoadOptions | null;
+
+    /**
+     * Indicates the readiness of the playing movie.
+     *
+     * @returns The `ReadyState` of the player.
+     */
+    get readyState(): ReadyState;
```
Dinnerbone, line 33: We should probably use an actually public enum, maybe string based

web/packages/core/src/public/v1.ts
```diff
@@ -0,0 +1,173 @@
...
+     *
+     * The options will be defaulted by the [[config]] field, which itself
+     * is defaulted by a global `window.RufflePlayer.config`.
+     */
+    load(options: string | URLLoadOptions | DataLoadOptions): Promise<void>;
+
+    /**
+     * Plays or resumes the movie.
+     */
+    play(): void;
```
Dinnerbone, line 72: This is actually Ruffle's suspend/resume, not Flash's play/pause. I suggest we rename this to `resume()` or `suspended = false` or something similar

web/packages/core/src/public/v1.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
```
Dinnerbone, line 74: This is actually Ruffle's suspend/resume, not Flash's play/pause. I suggest we rename this to `isSuspended`

web/packages/core/src/public/v1.ts
```diff
@@ -0,0 +1,173 @@
...
+     * Whether this player is currently playing.
+     *
+     * @returns True if this player is playing, false if it's paused or hasn't started yet.
+     */
+    get isPlaying(): boolean;
```
Dinnerbone, line 79: This is actually Ruffle's suspend/resume, not Flash's play/pause. I suggest we rename this to isSuspended

web/packages/core/src/public/v1.ts
```diff
@@ -0,0 +1,173 @@
...
+     */
+    get isFullscreen(): boolean;
+
+    /**
+     * Exported function that requests the browser to change the fullscreen state if
+     * it is allowed.
+     *
+     * @param isFull Whether to set to fullscreen or return to normal.
+     */
+    setFullscreen(isFull: boolean): void;
```
Dinnerbone, line 119: Maybe a setter?

web/packages/core/src/public/v1.ts
```diff
@@ -0,0 +1,173 @@
...
+     * Requests the browser to make this player fullscreen.
+     *
+     * This is not guaranteed to succeed, please check [[fullscreenEnabled]] first.
+     */
+    enterFullscreen(): void;
+
+    /**
+     * Requests the browser to no longer make this player fullscreen.
+     */
+    exitFullscreen(): void;
```
Dinnerbone, line 131: These are identical to `setFullscreen(true)` / `setFullscreen(false)`, seems a bit redundant

web/packages/core/src/public/v1.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * Pauses this player.
+     *
+     * No more frames, scripts or sounds will be executed.
+     * This movie will be considered inactive and will not wake up until resumed.
+     */
+    pause(): void;
```
Dinnerbone, line 139: This is actually Ruffle's suspend/resume, not Flash's play/pause. I suggest we rename this to `suspend` or `suspended = true` or something

web/packages/core/src/public/v1.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * Sets a trace observer on this flash player.
+     *
+     * The observer will be called, as a function, for each message that the playing movie will "trace" (output).
+     *
+     * @param observer The observer that will be called for each trace.
+     */
+    set traceObserver(observer: ((message: string) => void) | null);
```
Dinnerbone, line 148: Do we want to set **the** trace observer, as today - or set **a** trace observer?

web/packages/core/src/public/v1.ts
danielhjacobs: Probably, keep in mind the idea here may generally be to match https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/readyState#value (though HAVE_CURRENT_DATA, HAVE_FUTURE_DATA, and HAVE_ENOUGH_DATA don't necessarily all make sense for Ruffle, though depending how streamed loading of content is implemented maybe they will).

Dinnerbone: Is this to match it for compatibility, or just for being a similar thing? Because if it's the latter; there's also document.readyState which _is_ a string, for precedence

danielhjacobs: Context is in https://github.com/ruffle-rs/ruffle/pull/4084 and https://github.com/ruffle-rs/ruffle/pull/4813#issuecomment-892177759

web/packages/core/src/public/v1.ts
evilpie: Why do we have this method and enter/exitFullscreen?

web/packages/core/src/public/v1.ts
```diff
@@ -0,0 +1,173 @@
...
+     * @param value The volume. 1.0 is 100% volume.
+     */
+    set volume(value: number);
+
+    /**
+     * Checks if this player is allowed to be fullscreen by the browser.
+     *
+     * @returns True if you may call [[enterFullscreen]].
+     */
+    get fullscreenEnabled(): boolean;
```
evilpie, line 104: Isn't this name misleading? I think this could easily be confused to with "isFullscreen".

web/packages/core/src/public/v1.ts
evilpie: Oh sorry, I just saw your comment below.

Maybe we should have a getter and setter `fullscreen`, and a getter for `fullscreenAllowed`?

web/packages/core/src/public/v1.ts
evilpie: We could also consider switching the player to be a proper `EventTarget` and use `addEventListener` for this.

web/packages/core/src/public/v1.ts
danielhjacobs: Doesn't it match https://developer.mozilla.org/en-US/docs/Web/API/Document/fullscreenEnabled?

Dinnerbone:
I've gone ahead and changed `pause()`/`play()`/`isPlaying` to `suspend()`/`resume()`/`suspended` as that seems fairly agreed upon. Comment if anyone disagrees with the specifics before this PR is finished.

web/packages/core/src/public/v1.ts
Dinnerbone: I think I'd prefer actions/verbs to be methods, not just a property (it feels weird, not sure I can articulate it, it's like a side effect even if it's documented as such?)

How about we just go with `requestFullscreen()`/`exitFullscreen` to match the same methods on `Element`?

web/packages/core/src/public/v1.ts
Dinnerbone: tbh I don't entirely agree with having this method. It's a document thing, not at all related to this player or element. The source is `return document.fullscreenEnabled || document.webkitFullscreenEnabled` and whilst I guess webkit prefix is certainly a consideration... is it our place? It just feels weird to have a method like that.

Dinnerbone:
Renamed `enterFullscreen` to `requestFullscreen`, and removed `setFullscreen`.
Decided to keep `fullscreenEnabled` as if we still support old ios, it's likely users will forget to check the prefixed stuff too.

web/packages/core/src/public/v1.ts
Dinnerbone: If we go `EventTarget` we'll lose the warning of unhandled commands, unless we require people to flag that they've handled it. Right now we ask that the `onFSCommand` returns true when handled.

I think that's probably fine, but what do you think?

Dinnerbone: I thought about this a bit and decided that event may not be the way to go. A few issues:
1. Versioning will be difficult
2. It's a bit weird to have the element have the event when everything else is behind `ruffle()` - people may accidentally `ruffle().addEventListener`
3. Typescript really does not seem to like custom events... makes it super verbose to use :(

I'll go with option 2 for now and we can reevaluate in the future for a V2.

Dinnerbone:
Changed `onFSCommand` (getter/setter) to `addFSCommandHandler` (method).

Dinnerbone:
Made `ReadyState` a string

web/packages/core/src/public/v1.ts
Dinnerbone: I think setting is fine.

web/packages/core/src/public/player/legacy.ts
```diff
@@ -1,56 +1,106 @@
 import { DataLoadOptions, URLLoadOptions } from "../config";
 import { MovieMetadata } from "./movie-metadata";
-import { ReadyState } from "../../internal/player/inner";
+
+/**
+ * Describes the loading state of an SWF movie.
+ *
+ * @deprecated Use {@link PlayerV1.readyState} to get a {@link ReadyState} instead.
+ */
+export enum LegacyReadyState {
```
danielhjacobs, line 9: Can we remove this enum and just compare any legacy ready states to numbers instead of enum members? This only adds confusion.

Dinnerbone: I'm not sure that having people just need to know what arbitrary numbers means is better than code constants (which an enum basically is). That seems even more confusing to me.
Adrian argued in favour of just keeping the legacy enum and not having the new one, which whilst I'm not personally in favour of, I think I'd prefer over switching and entirely not documenting the old style :(

Dinnerbone: Reverted to old style API, and made the old enum public

adrian17 (approved):

Issue #17895, opened 2024-09-13
Add option to request GameMode on Linux


Tiagoquix:
Would be cool to have a checkbox to automatically request [GameMode](https://github.com/FeralInteractive/gamemode) whenever Ruffle is running.

kjarosh:
We can implement this fairly easily using ashpd or zbus

PR #17940, opened 2024-09-15
desktop: Add gamemode support

kjarosh:
Fixes https://github.com/ruffle-rs/ruffle/issues/17895.

This PR adds gamemode support through ashpd (Linux-only).

* There are 3 values for the gamemode setting:
  * `Off`—gamemode is disabled,
  * `On`—gamemode is enabled,
  * `Default`—gamemode is enabled only on high power preference.
* A CLI option `--gamemode` is added.
* A GUI is added to the preferences.

Gamemode is requested only when a SWF file is loaded.

<image src="https://github.com/user-attachments/assets/6ac4f052-7daf-4396-82cd-44dbc4452d73" width="400"/>

TODO:
* [x] Add a tooltip which describes what `Default` does.
* [ ] Switch away from ashpd git when https://github.com/bilelmoussaoui/ashpd/pull/234 is released.


kjarosh:
> just curious, what made you go with `ashpd` over `zbus` directly? I assume the readily available (even if buggy) high level support for this specific use case - in which case, great!

That's true.

> But technically, `ashpd`, outside of a Flatpak jail, just does a DBus call or two: and `zbus` also works over Flatpak, right?

That's also true.

I could go with `zbus` and write the proxy myself, but we depend on `ashpd` anyway (through `rfd`) and using `ashpd` we don't have to worry about D-Bus names and types, which is a value added.

However, if we have good reasons to stick with `zbus`, I can also switch to it.

(I'm also considering rewriting our `FreedesktopSettings` to use `ashpd` for the same reasons.)

torokati44:
> However, if we have good reasons to stick with `zbus`, I can also switch to it.

No, quite the opposite, in agreement with you. I was just checking my assumptions.

Dinnerbone:
CI only lints on linux so I'll pretend to be a bot for a moment.

```
warning: field `gamemode_preference_readonly` is never read
  --> desktop\src\gui\dialogs\preferences_dialog.rs:24:5
   |
11 | pub struct PreferencesDialog {
   |            ----------------- field in this struct
...
24 |     gamemode_preference_readonly: bool,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: method `show_gamemode_preferences` is never used
   --> desktop\src\gui\dialogs\preferences_dialog.rs:304:8
    |
52  | impl PreferencesDialog {
    | ---------------------- method in this implementation
...
304 |     fn show_gamemode_preferences(
    |        ^^^^^^^^^^^^^^^^^^^^^^^^^

warning: function `gamemode_preference_name` is never used
   --> desktop\src\gui\dialogs\preferences_dialog.rs:556:4
    |
556 | fn gamemode_preference_name(
    |    ^^^^^^^^^^^^^^^^^^^^^^^^
```

kjarosh:
Dinnerbot? :eyes:

Cargo.lock
```diff
@@ -246,6 +246,22 @@ dependencies = [
...
+ "futures-channel",
+ "futures-util",
+ "rand",
+ "serde",
+ "serde_repr",
+ "tokio",
+ "url",
+ "zbus",
+]
+
```
torokati44, line 264: Could we maybe `[patch.crates-io]`, so it doesn't get duplicated?
I suppose it's not a big deal, and that would be just one more piece of debris to clean up afterward...

torokati44: Or we could just wait a few days...: https://github.com/bilelmoussaoui/ashpd/pull/234#issuecomment-2357986893 :)

Dinnerbone: A few days passed; let's get this in?

torokati44: I don't object.

kjarosh: It should be ready for merge now

desktop/src/player.rs
```diff
@@ -259,6 +264,18 @@ impl ActivePlayer {
...
+        let gamemode_enable = match preferences.gamemode_preference() {
+            GameModePreference::Default => {
+                preferences.graphics_power_preference() == PowerPreference::High
+            }
+            GameModePreference::On => true,
+            GameModePreference::Off => false,
+        };
+
+        if cfg!(not(target_os = "linux")) && gamemode_enable {
+            tracing::warn!("Cannot enable GameMode, as it is supported only on Linux");
```
adrian17, line 276: Now that I look at it, this ends up adding a pointless warning on windows by default :(

kjarosh: https://github.com/ruffle-rs/ruffle/pull/18242

PR #17946, opened 2024-09-16
core: Optimize BMD.fillRect()

adrian17:
This removed repeated range checks on each pixel, replacing them by possibly-memcpy.
I originally only added `set_pixel32_row_raw`, which improved 640x480 fills by 3x (FF) - 5x (desktop).
Then I tried adding a full-fill fast path, which (as expected) didn't do a noticeable difference on desktop, but somehow improved FF times by another 2x.

~~EDIT: I just remembered that I tested browser perf without wasm extensions, but bulk-memory might also do a difference here. Will check it soon.~~
With wasm-opt and bulk-memory enabled, the line-fill path and full-fill path are equally fast, and another 2x faster than the best result without extensions. Just slightly slower than desktop now (though weirdly, the FF profiler doesn't show `memory.fill` like it used to).

This only helps the CPU side, it won't help is most time is spent waiting for GPU.

### DIFF
--- a/core/src/bitmap/bitmap_data.rs
+++ b/core/src/bitmap/bitmap_data.rs
@@ -711,6 +711,19 @@ impl<'gc> BitmapData<'gc> {
     }
 
     #[inline]
+    pub fn set_pixel32_row_raw(&mut self, x1: u32, x2: u32, y: u32, color: Color) {
+        let p1 = (x1 + y * self.width) as usize;
+        let p2 = (x2 + y * self.width) as usize;
+        let slice = &mut self.pixels[p1..p2];
+        slice.fill(color);
+    }
+
+    #[inline]
+    pub fn fill(&mut self, color: Color) {
+        self.pixels.fill(color);
+    }
+
+    #[inline]
     pub fn get_pixel32_raw(&self, x: u32, y: u32) -> Color {
         self.pixels[(x + y * self.width()) as usize]
     }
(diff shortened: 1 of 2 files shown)


torokati44:
Ooh, really nice! 

I think you could perhaps, instead of detecting a "full fill", you could, with a _little bit_ more code, make it _ever so slightly_ more generic, by detecting "full width fill from rows `i` through `j` - then the full fill would be just an accidental edge case of this, while still performing the same, I reckon.

nir: I think this is `A-core`, not `A-render`... :P Maybe both...? 👀

adrian17:
> I think, instead of detecting a "full fill", you could, with a little bit more code, make it ever so slightly more generic, by detecting "full width fill of rows from i through j" - then the full fill would be just an accidental edge case of this, while still performing the same, I reckon.

Yeah, but I'd say at this point the difference between 1 big copy and H width-sized copies is already quite small, so I'm not sure if it's worth it. I was already on a fence about adding the `fill()` path.

torokati44:
> Yeah, but I'd say at this point the difference between 1 big copy and H width-sized copies is already quite small, so I'm not sure if it's worth it. I was already on a fence about adding the fill() path.

That's fair - it's not like I can find any real world content right off the top of my head that does "fill all but the last 3 lines of a 5x1700 BitmapData repeatedly". And even if such content does exist ... hopefully we'll notice...?

PR #17950, opened 2024-09-16
chore: Improve iOS support

madsmtm:
Fix a few instances of `target_os = "macos"`. Required for https://github.com/madsmtm/ruffle-ios.

### DIFF
--- a/core/src/avm2/globals/flash/system/capabilities.rs
+++ b/core/src/avm2/globals/flash/system/capabilities.rs
@@ -16,7 +16,7 @@ pub fn get_os<'gc>(
         PlayerRuntime::AIR => {
             if cfg!(windows) {
                 "Windows 10"
-            } else if cfg!(target_os = "macos") {
+            } else if cfg!(target_vendor = "apple") {
                 "Mac OS 10.5.2"
             } else {
                 "Linux 5.10.49"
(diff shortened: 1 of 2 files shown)


video/external/src/backend.rs
```diff
@@ -82,6 +82,10 @@ impl ExternalVideoBackend {
                 "libopenh264-2.4.1-mac-arm64.dylib",
                 "213ff93831cfa3dd6d7ad0c3a3403a6ceedf4ac1341e1278b5b869d42fefb496",
             ),
+            ("ios", _) => (
+                "libopenh264-2.4.1-ios.a",
```
torokati44, line 86: Isn't this a static library? If so, what good is downloading it at runtime? 🤔

madsmtm: You're right, that makes no sense, I didn't think when I migrated that part

torokati44:
I think these changes look fine, and will likely end up being accepted; but it looks like they can be split into two main parts - even though they are serving the same goal. Would you mind splitting out the audio backend refactor into its own PR please, just for easier review discussion?
Also, make sure to tag/label the commits/PRs according to `CONTRIBUTING.md`.
Thank you!

madsmtm:
Yeah, that was lazy of me, sorry, I've made https://github.com/ruffle-rs/ruffle/pull/17951 for the `cpal` moving

PR #17965, opened 2024-09-17
core: Make applyFilter with zero-sized rects noops

adrian17:
Verified that it fixes https://github.com/ruffle-rs/ruffle/issues/17964 . Hopefully should fix most/all of "dimension X is zero" in applyFilter.

I don't know whether this is the best place (it can be done a layer above and below, and after more clamping). @Dinnerbone can you please check it out?

### DIFF
--- a/core/src/bitmap/operations.rs
+++ b/core/src/bitmap/operations.rs
@@ -1210,6 +1210,17 @@ pub fn apply_filter<'gc>(
     dest_point: (u32, u32),
     filter: Filter,
 ) {
+    // Prevent creating 0x0 textures.
+    // FIXME: this is not correct.
+    // Currently at minimum, applyFilter(blur) is bugged in that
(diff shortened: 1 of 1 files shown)


Dinnerbone (approved):
Looks good to me, thank you!

I was initially concerned that we might not catch things that are *effectively zero*, but I think in practice filters grow the size so an effective zero might end up being a positive value, and I think our rendering should handle that fine as long as the source size isn't zero; so this should be good, I think!

torokati44 (approved):

adrian17:
On hold until the Discord questions are resolved.

torokati44:
(converted to draft accordingly)

Dinnerbone (approved):

n0samu:
Strangely, #17964 seems to have been deleted. I've archived it from Google's cache for future reference: https://archive.is/szqJP

Issue #17986, opened 2024-09-19
Theme detection on startup stopped working


kjarosh:
### Describe the bug

When you start Ruffle it's always in dark mode

### Expected behavior

It should be in light mode when the system theme is light

### Content Location

N/A

### Affected platform

Desktop app

### Operating system

Linux

### Additional information

Theme changes dynamically when system theme changes, it just does not work on startup

miguellords:
This might help:This file might fix it

https://mega.nz/file/1vkGlJTL#yU-SJLlIilF0mx5qwpPOd3BM8iKZqOJLEQUStR76rrQ
I put the necessary dlls in the archive

kjarosh:
That was regressed by https://github.com/ruffle-rs/ruffle/commit/490c02bb58d2c8ebc11233dbf9d6dacfd8f025cc in https://github.com/ruffle-rs/ruffle/pull/17401

CC @torokati44

torokati44:
Yeah, that PR was my immediate suspect... :/

kjarosh:
More info: it also does not work when the theme is set to light in the preferences. When debugging I can see that the proper theme is detected, it just seems that setting the theme on start in egui does nothing.

PR #17987, opened 2024-09-19
desktop: Bump `egui` to latest `master`, set theme instead of visuals

torokati44:
Aiming to fix https://github.com/ruffle-rs/ruffle/issues/17986.

Please verify, @kjarosh.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1430,7 +1430,7 @@ checksum = "f25c0e292a7ca6d6498557ff1df68f32c99850012b6ea401cf8daf771f22ff53"
 [[package]]
 name = "ecolor"
 version = "0.28.1"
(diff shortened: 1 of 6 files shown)


kjarosh:
Confirmed working, thank you!

Reading the rationale behind these changes in egui I see that we now should be able to easily integrate with the accent color too, not only dark/light mode!

kjarosh (approved):

torokati44:
Whoo! Should our accent color be dark blue or orange? The former for dark theme, latter for light? Or the other way around?

kjarosh:
I was rather thinking about https://github.com/flatpak/xdg-desktop-portal/blob/main/data/org.freedesktop.portal.Settings.xml#L43 with a possibility to set a custom one (similarly to the dark/light mode).

torokati44:
Oh, so it's a global setting, chosen by the user, rather than a way for applications to subtly customize an otherwise cohesive UI design... gotit!

PR #18007, opened 2024-09-20
AVM2: move cached namespaces to a new `CommonNamespaces` struct

moulins:
Extracted from #17876; this mirrors #17842.

### DIFF
--- a/core/src/avm2/globals.rs
+++ b/core/src/avm2/globals.rs
@@ -370,10 +370,7 @@ fn define_fn_on_global<'gc>(
     script: Script<'gc>,
 ) {
     let (_, global, domain) = script.init();
-    let qname = QName::new(
-        Namespace::package("", ApiVersion::AllVersions, &mut activation.borrow_gc()),
-        name,
-    );
+    let qname = QName::new(activation.avm2().namespaces.public_all(), name);
     let func = domain
         .get_defined_value(activation, qname)
         .expect("Function being defined on global should be defined in domain!");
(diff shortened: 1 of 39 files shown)


core/src/avm2/globals.rs
```diff
@@ -557,30 +555,28 @@ pub fn load_player_globals<'gc>(
...
-    let public_ns = activation.avm2().public_namespace_base_version;
-    let vector_public_ns = activation.avm2().vector_public_namespace;
-    let vector_internal_ns = activation.avm2().vector_internal_namespace;
-
     // Unfortunately we need to specify the global traits manually, at least until
     // all the builtin classes are defined in AS.
     let mut global_traits = Vec::new();
 
+    let namespaces = activation.avm2().namespaces;
+    #[rustfmt::skip]
```
Lord-McSweeney, line 563: Can we keep the above
```
    let public_ns = activation.avm2().public_namespace_base_version;
    let vector_public_ns = activation.avm2().vector_public_namespace;
    let vector_internal_ns = activation.avm2().vector_internal_namespace;
```
and avoid skipping fmt here?

core/src/avm2/namespace.rs
```diff
@@ -336,3 +336,77 @@ impl<'gc> Namespace<'gc> {
...
+
+    // These are required to facilitate shared access between Rust and AS.
+    pub(super) flash_display_internal: Namespace<'gc>,
+    pub(super) flash_utils_internal: Namespace<'gc>,
+    pub(super) flash_geom_internal: Namespace<'gc>,
+    pub(super) flash_events_internal: Namespace<'gc>,
+    pub(super) flash_text_engine_internal: Namespace<'gc>,
+    pub(super) flash_net_internal: Namespace<'gc>,
+
+    pub(super) __ruffle__: Namespace<'gc>,
```
Lord-McSweeney, line 360: Can this be called `ruffle_internal` instead?

Lord-McSweeney (approved):
LGTM besides nits

core/src/avm2/namespace.rs
moulins: This namespace is actually public, because it's purpose for Ruffle-specific functions we want to call in tests.

PR #18037, opened 2024-09-22
desktop: Add Alt+Enter back as a hidden shortcut on Windows

kjarosh:
Previously <kbd>Alt</kbd>+<kbd>Enter</kbd> was a hidden shortcut that was used to toggle full screen, and it is used by Flash Player to toggle full screen on Windows too.

As Windows users sometimes expect this shortcut to work, this patch brings it back as a hidden shortcut (until we add some kind of preferences).

CC @n0samu

### DIFF
--- a/desktop/src/gui/menu_bar.rs
+++ b/desktop/src/gui/menu_bar.rs
@@ -23,6 +23,8 @@ pub struct MenuBar {
 
 impl MenuBar {
     const SHORTCUT_FULLSCREEN: KeyboardShortcut = KeyboardShortcut::new(Modifiers::NONE, Key::F11);
+    const SHORTCUT_FULLSCREEN_WINDOWS: KeyboardShortcut =
+        KeyboardShortcut::new(Modifiers::ALT, Key::Enter);
     const SHORTCUT_OPEN: KeyboardShortcut = KeyboardShortcut::new(Modifiers::COMMAND, Key::O);
     const SHORTCUT_OPEN_ADVANCED: KeyboardShortcut =
         KeyboardShortcut::new(Modifiers::COMMAND.plus(Modifiers::SHIFT), Key::O);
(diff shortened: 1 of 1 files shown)


n0samu (approved):

torokati44:
> and it is used by Flash Player to toggle full screen on Windows too.

Then why have it only enter fullscreen for Ruffle, and not toggle?

kjarosh:
> Then why have it only enter fullscreen for Ruffle, and not toggle?

Wdym? It should toggle full screen the same way F11 does.

desktop/src/gui/menu_bar.rs
```diff
@@ -64,7 +66,14 @@ impl MenuBar {
                 player.set_is_playing(!player.is_playing());
             }
         }
-        if egui_ctx.input_mut(|input| input.consume_shortcut(&Self::SHORTCUT_FULLSCREEN)) {
+        let mut fullscreen =
+            egui_ctx.input_mut(|input| input.consume_shortcut(&Self::SHORTCUT_FULLSCREEN));
+        if cfg!(windows) && !fullscreen {
```
torokati44, line 71: But it doesn't.

torokati44: Wait, I think I misunderstood something, sorry.

kjarosh: The variable `fullscreen` tells whether the fullscreen shortcut was pressed, the code that toggles fullscreen is few lines below: https://github.com/ruffle-rs/ruffle/pull/18037/files#diff-f8a52a3fe6f612b8157bf85e7e2fb1cfaaaf771d54d44e6878a9349805dcb120R79

kjarosh: And done, it's now `fullscreen_pressed`

desktop/src/gui/menu_bar.rs
```diff
@@ -64,7 +66,14 @@ impl MenuBar {
                 player.set_is_playing(!player.is_playing());
             }
         }
-        if egui_ctx.input_mut(|input| input.consume_shortcut(&Self::SHORTCUT_FULLSCREEN)) {
+        let mut fullscreen_pressed =
+            egui_ctx.input_mut(|input| input.consume_shortcut(&Self::SHORTCUT_FULLSCREEN));
+        if cfg!(windows) && !fullscreen_pressed {
+            // TODO We can remove this shortcut when we add some kind of preferences.
+            fullscreen_pressed = egui_ctx
+                .input_mut(|input| input.consume_shortcut(&Self::SHORTCUT_FULLSCREEN_WINDOWS));
```
torokati44, line 74: It irks me a little that the second shortcut is only consumed if the first one hasn't been (instead of consuming both unconditionally and combining the results after), but it's such a minor thing that is very unlikely to matter, that it's not worth blocking for.

torokati44 (approved):
Thank you!

Issue #17902, opened 2024-09-14
Error on https://www.minijuegos.com/helpers/game/sandbox.php


mateoelproo:
# Error Info
Error name: Error
Error message: panicked at /home/runner/.cargo/git/checkouts/rust-flash-lso-7db697fbf6bd96f3/98679aa/flash-lso/src/amf3/read.rs:435:14:
Unable to get Object
Error stack:
```
Error: panicked at /home/runner/.cargo/git/checkouts/rust-flash-lso-7db697fbf6bd96f3/98679aa/flash-lso/src/amf3/read.rs:435:14:
Unable to get Object
    at n.wbg.__wbg_new_796382978dfd4fb0 (https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.eff553c567a7ead39869.js:1:83857)
    at ruffle_web.wasm.js_sys::Error::new::hfb561c222a4e70eb (wasm://wasm/ruffle_web.wasm-03212586:wasm-function[12740]:0x985756)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once{{vtable.shim}}::he535d08ebd1aba14 (wasm://wasm/ruffle_web.wasm-03212586:wasm-function[9796]:0x915500)
    at ruffle_web.wasm.std::panicking::rust_panic_with_hook::h33fe77d38d305ca3 (wasm://wasm/ruffle_web.wasm-03212586:wasm-function[6360]:0x805dca)
    at ruffle_web.wasm.core::panicking::panic_fmt::hde8b7aa66e2831e1 (wasm://wasm/ruffle_web.wasm-03212586:wasm-function[9515]:0x906186)
    at ruffle_web.wasm.core::option::expect_failed::hacfbd4e0f8d6ca3b (wasm://wasm/ruffle_web.wasm-03212586:wasm-function[8896]:0x8e472c)
    at ruffle_web.wasm.flash_lso::amf3::read::AMF3Decoder::parse_single_element::h1c9def9900e85592 (wasm://wasm/ruffle_web.wasm-03212586:wasm-function[795]:0xe5e4d)
    at ruffle_web.wasm.flash_lso::amf3::read::AMF3Decoder::parse_single_element::h1c9def9900e85592 (wasm://wasm/ruffle_web.wasm-03212586:wasm-function[795]:0xe32a1)
[...]

Issue #17782, opened 2024-09-06
Error on https://www.megaigry.ru/embed/zombidle/


CraMbic444:
# Error Info
Error name: Error
Error message: panicked at /home/runner/.cargo/git/checkouts/rust-flash-lso-7db697fbf6bd96f3/98679aa/flash-lso/src/amf3/read.rs:435:14:
Unable to get Object
Error stack:
```
Error: panicked at /home/runner/.cargo/git/checkouts/rust-flash-lso-7db697fbf6bd96f3/98679aa/flash-lso/src/amf3/read.rs:435:14:
Unable to get Object
    at n.wbg.__wbg_new_796382978dfd4fb0 (https://www.megaigry.ru/emulators/ruffle/core.ruffle.1fad5110710476fd8724.js:1:83857)
    at ruffle_web.wasm.js_sys::Error::new::hfb561c222a4e70eb (wasm://wasm/ruffle_web.wasm-0320469e:wasm-function[12753]:0x98635f)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once{{vtable.shim}}::h81099db94754a517 (wasm://wasm/ruffle_web.wasm-0320469e:wasm-function[9827]:0x917b42)
    at ruffle_web.wasm.std::panicking::rust_panic_with_hook::h33fe77d38d305ca3 (wasm://wasm/ruffle_web.wasm-0320469e:wasm-function[6388]:0x80ffe8)
    at ruffle_web.wasm.core::panicking::panic_fmt::hde8b7aa66e2831e1 (wasm://wasm/ruffle_web.wasm-0320469e:wasm-function[9548]:0x908861)
    at ruffle_web.wasm.core::option::expect_failed::hacfbd4e0f8d6ca3b (wasm://wasm/ruffle_web.wasm-0320469e:wasm-function[8930]:0x8e731c)
    at ruffle_web.wasm.flash_lso::amf3::read::AMF3Decoder::parse_single_element::h1c9def9900e85592 (wasm://wasm/ruffle_web.wasm-0320469e:wasm-function[795]:0xe5a5d)
    at ruffle_web.wasm.ruffle_core::avm2::globals::flash::utils::byte_array::read_object::h9393bade97bee456 (wasm://wasm/ruffle_web.wasm-0320469e:wasm-function[902]:0x1e410f)
[...]

Issue #17709, opened 2024-08-30
Error on http://www.hackedfreegames.com/game/7462/tequila-zombies-3


Anony2043:
# Error Info
Error name: Error
Error message: panicked at /home/runner/.cargo/git/checkouts/rust-flash-lso-7db697fbf6bd96f3/f9e3de5/flash-lso/src/amf3/read.rs:435:14:
Unable to get Object
Error stack:
```
Error: panicked at /home/runner/.cargo/git/checkouts/rust-flash-lso-7db697fbf6bd96f3/f9e3de5/flash-lso/src/amf3/read.rs:435:14:
Unable to get Object
    at push.69.imports.wbg.__wbg_new_796382978dfd4fb0 (chrome-extension://donbcfbmhbcapadipfkeojnmajbakjdc/dist/69.js:3784:21)
    at ruffle_web.wasm.js_sys::Error::new::h88ecd9455fcda73d (wasm://wasm/ruffle_web.wasm-031f09b6:wasm-function[12706]:0x981f36)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once{{vtable.shim}}::h62db458ea18d0144 (wasm://wasm/ruffle_web.wasm-031f09b6:wasm-function[9778]:0x910f41)
    at ruffle_web.wasm.std::panicking::rust_panic_with_hook::h47bd3d747ed79dc3 (wasm://wasm/ruffle_web.wasm-031f09b6:wasm-function[6254]:0x7f078d)
    at ruffle_web.wasm.core::panicking::panic_fmt::hb859252f4b513814 (wasm://wasm/ruffle_web.wasm-031f09b6:wasm-function[10187]:0x92cab4)
    at ruffle_web.wasm.core::option::expect_failed::h9d0c8588efc786aa (wasm://wasm/ruffle_web.wasm-031f09b6:wasm-function[8893]:0x8e11e6)
    at ruffle_web.wasm.flash_lso::amf3::read::AMF3Decoder::parse_single_element::hf93725722d921b93 (wasm://wasm/ruffle_web.wasm-031f09b6:wasm-function[795]:0xef507)
    at ruffle_web.wasm.flash_lso::amf3::read::AMF3Decoder::parse_single_element::hf93725722d921b93 (wasm://wasm/ruffle_web.wasm-031f09b6:wasm-function[795]:0xec676)
[...]

Issue #14329, opened 2023-12-06
Error on http://media-r03ru.rjgplay.com/inconflict-loader.swf


petr590:
# Error Info
Error name: Error
Error message: panicked at /home/runner/.cargo/git/checkouts/rust-flash-lso-7db697fbf6bd96f3/2f976fb/flash-lso/src/amf3/read.rs:417:14:
Unable to get Object

# Player Info
Allows script access: false
Renderer: wgpu
Adapter Backend: Gl
Adapter Name: "Intel(R) HD Graphics"
Adapter Device Type: IntegratedGpu
Adapter Driver Name: ""
Adapter Driver Info: ""
Enabled features: Features(TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES | PUSH_CONSTANTS | SHADER_UNUSED_VERTEX_OUTPUT)
Available features: Features(TEXTURE_COMPRESSION_ETC2 | CLEAR_TEXTURE)
Current limits: Limits { max_texture_dimension_1d: 4096, max_texture_dimension_2d: 4096, max_texture_dimension_3d: 2048, max_texture_array_layers: 256, max_bind_groups: 4, max_bindings_per_bind_group: 1000, max_dynamic_uniform_buffers_per_pipeline_layout: 8, max_dynamic_storage_buffers_per_pipeline_layout: 0, max_sampled_textures_per_shader_stage: 16, max_samplers_per_shader_stage: 16, max_storage_buffers_per_shader_stage: 0, max_storage_textures_per_shader_stage: 0, max_uniform_buffers_per_shader_stage: 11, max_uniform_buffer_binding_size: 16384, max_storage_buffer_binding_size: 0, max_vertex_buffers: 8, max_buffer_size: 268435456, max_vertex_attributes: 16, max_vertex_buffer_array_stride: 255, min_uniform_buffer_offset_alignment: 32, min_storage_buffer_offset_alignment: 256, max_inter_stage_shader_components: 60, max_compute_workgroup_storage_size: 0, max_compute_invocations_per_workgroup: 0, max_compute_workgroup_size_x: 0, max_compute_workgroup_size_y: 0, max_compute_workgroup_size_z: 0, max_compute_workgroups_per_dimension: 0, max_push_constant_size: 96 }
Surface quality: high
Surface samples: 4
Surface size: Extent3d { width: 550, height: 400, depth_or_array_layers: 1 }

# Page Info
[...]

PR #18046, opened 2024-09-23
core: Fix "Failed to get Object" panics

CUB3D:
This constitues the ruffle-side changes of https://github.com/ruffle-rs/rust-flash-lso/pull/67 and https://github.com/ruffle-rs/rust-flash-lso/pull/68. 

This fixes the following (previously would panic on load): 
https://github.com/ruffle-rs/ruffle/issues/17902 - loads
https://github.com/ruffle-rs/ruffle/issues/17709 - loads
https://github.com/ruffle-rs/ruffle/issues/13748 - loads
https://github.com/ruffle-rs/ruffle/issues/10520 - loads

This prevents panics on the following:
https://github.com/ruffle-rs/ruffle/issues/17782 - (and possible dupe https://github.com/ruffle-rs/ruffle/issues/17336 )
Now errors with "Property uploadFromBitmapData not found on flash.display3D.textures.RectangleTexture and there is no default value."

https://github.com/ruffle-rs/ruffle/issues/14329
Fails to load past 54%, some URL load failures might be related

https://github.com/ruffle-rs/ruffle/issues/10174
Same as 17782 above

These are not fixed, but one possible underlying issue is that a new `AMF3Decoder` is created for each call to `ByteArray.readObject` which could result in missing references that are replaced with `Undefined` when an element references a prior value in the stream. In the simple case, we could just cache the `AMF3Decoder` in the `ByteArray` object, but this won't handle more complex cases where the `ByteArray`s position is changed, so I'm leaving this for future work.
Update: I'm actually unsure if the above guess is true anymore, a missing change progressed the above and removed the invalid reference warnings but it still seems odd.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ RECOVER_*.fla
 
 # Jetbrains tools (CLion, IntelliJ, etc)
 /.idea
+
+# Mac junk
+.DS_Store
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -43,7 +43,7 @@ serde = { workspace = true }
 serde_json = { version = "1.0", features = ["preserve_order"] }
 nellymoser-rs = { git = "https://github.com/ruffle-rs/nellymoser", rev = "754b1184037aa9952a907107284fb73897e26adc", optional = true }
 regress = "0.10"
-flash-lso = { git = "https://github.com/ruffle-rs/rust-flash-lso", rev = "98679aa0cdcad5fecc5c7321b2eac35b69dbcd5f" }
+flash-lso = { git = "https://github.com/ruffle-rs/rust-flash-lso", rev = "cbd18e1a79cf902f8ff1d2bf551801c4021b3be6" }
 lzma-rs = {version = "0.3.0", optional = true }
 dasp = { version = "0.11.0", features = ["interpolate", "interpolate-linear", "signal"], optional = true }
 symphonia = { version = "0.5.4", default-features = false, features = ["mp3"], optional = true }
--- a/core/src/avm2/amf.rs
+++ b/core/src/avm2/amf.rs
@@ -306,7 +306,7 @@ pub fn deserialize_value_impl<'gc>(
             // First let's create an array out of `values` (dense portion), then we add the elements onto it.
             let mut arr: Vec<Option<Value<'gc>>> = Vec::with_capacity(values.len());
             for value in values {
-                arr.push(Some(deserialize_value(activation, value)?));
+                arr.push(Some(deserialize_value_impl(activation, value, object_map)?));
             }
             array
                 .as_array_storage_mut(activation.context.gc_context)
@@ -317,7 +317,7 @@ pub fn deserialize_value_impl<'gc>(
             for element in elements {
                 array.set_public_property(
                     AvmString::new_utf8(activation.context.gc_context, element.name()),
-                    deserialize_value(activation, element.value())?,
+                    deserialize_value_impl(activation, element.value(), object_map)?,
                     activation,
                 )?;
             }
@@ -330,7 +330,7 @@ pub fn deserialize_value_impl<'gc>(
 
             let mut arr: Vec<Option<Value<'gc>>> = Vec::with_capacity(values.len());
             for value in values {
-                arr.push(Some(deserialize_value(activation, value)?));
+                arr.push(Some(deserialize_value_impl(activation, value, object_map)?));
             }
 
             array
@@ -352,7 +352,7 @@ pub fn deserialize_value_impl<'gc>(
 
             for entry in elements {
                 let name = entry.name();
-                let value = deserialize_value(activation, entry.value())?;
+                let value = deserialize_value_impl(activation, entry.value(), object_map)?;
                 // Flash player logs the error and continues deserializing the rest of the object,
                 // even when calling a customer setter
                 if let Err(e) = obj.set_public_property(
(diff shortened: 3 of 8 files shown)


core/Cargo.toml
```diff
@@ -43,7 +43,7 @@ serde = { workspace = true }
 serde_json = { version = "1.0", features = ["preserve_order"] }
 nellymoser-rs = { git = "https://github.com/ruffle-rs/nellymoser", rev = "754b1184037aa9952a907107284fb73897e26adc", optional = true }
 regress = "0.10"
-flash-lso = { git = "https://github.com/ruffle-rs/rust-flash-lso", rev = "98679aa0cdcad5fecc5c7321b2eac35b69dbcd5f" }
+flash-lso = { git = "https://github.com/ruffle-rs/rust-flash-lso", rev = "016946cef2392f6b0bf79f79d692880ab76204a5" }
```
torokati44, line 46: This is the head of the currently unmerged https://github.com/ruffle-rs/rust-flash-lso/pull/68, which seems to be unfinished? This seems a bit unfortunate...

Would it be possible to update to the current `master` of that repo first?

.gitignore
```diff
@@ -13,3 +13,6 @@ RECOVER_*.fla
 
 # Jetbrains tools (CLion, IntelliJ, etc)
 /.idea
+
+# Mac junk
+.DS_Store
```
torokati44, line 18: nit: missing trailing linefeed

core/src/avm2/amf.rs
```diff
@@ -288,6 +294,15 @@ fn get_or_create_element<'gc>(
 pub fn deserialize_value<'gc>(
     activation: &mut Activation<'_, 'gc>,
     val: &AmfValue,
+) -> Result<Value<'gc>, Error<'gc>> {
+    let mut x = BTreeMap::new();
+    deserialize_value_impl(activation, val, &mut x)
```
Lord-McSweeney, line 299: `deserialize_value_impl` only seems to be called here, can't it be inlined into `deserialize_value`?

core/src/avm2/amf.rs
```diff
@@ -460,6 +507,14 @@ pub fn deserialize_value<'gc>(
         }
         AmfValue::AMF3(val) => deserialize_value(activation, val)?,
         AmfValue::Unsupported => Value::Undefined,
+        AmfValue::Amf3ObjectReference(r) => {
+            if let Some(o) = object_map.get(r) {
+                (*o).into()
+            } else {
+                tracing::error!("Got an object reference {r:?} to an object we've not seen yet");
```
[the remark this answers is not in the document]
CUB3D: Changed

core/src/avm2/amf.rs
CUB3D: Good catch, I forgot the change the other recursive calls to `deserialize_value_impl`, which has progressed 17782/17336/10174

.gitignore
CUB3D: Fixed

core/Cargo.toml
CUB3D: That pr isn't unfinished it's just waiting for review? I don't mind this pr being stalled until that pr is merged, but I'd rather not revert this to a buggier commit

torokati44: I see, sorry then. I interpreted "Note this is not as clean as I would like, but I want to fix the panics first and worry about the refactor later" as you still working on it.

torokati44:
Since the act of merging https://github.com/ruffle-rs/rust-flash-lso/pull/68 changed its HEAD hash (and there were some changes added to it even before the merge), I amended this branch accordingly. While at it, I also squashed a few "fixup" type commits into the original commits they fix, for cleaner commit history. I hope you don't mind, @CUB3D.

core/src/avm2/amf.rs
```diff
@@ -404,39 +436,54 @@ pub fn deserialize_value<'gc>(
...
+                        } else {
+                            value
+                        }
+                    })
+                })
+                .collect::<Result<Vec<_>, _>>()?;
+
+            // Swap in the actual values
+            obj.as_vector_storage_mut(activation.context.gc_context)
+                .unwrap()
```
torokati44, line 470: nit: I believe we generally prefer `expect`, with a meaningful message.

torokati44 (approved):
Other than that one note, I vote on merging this soon. It looks fairly non-problematic.

Dinnerbone (approved):
LGTM (but mostly trusting you as the expert here :D)

core/src/avm2/amf.rs
torokati44: Actually, I don't think it's worth blocking on this, I see no way for this to fail, `obj` is created locally, right above this line.

torokati44: If you are also okay with leaving this as is, @Dinnerbone, I'd say go for the merge.

core/src/avm2/array.rs
```diff
@@ -114,6 +114,23 @@ impl<'gc> ArrayStorage<'gc> {
...
+
+        match self {
+            ArrayStorage::Dense {
+                storage,
+                occupied_count,
+            } => {
+                *occupied_count = new_occupied_count;
+                *storage = new_storage;
+            }
+            ArrayStorage::Sparse { .. } => {}
```
Lord-McSweeney, line 130: Can this panic, just to ensure that it doesn't accidentally get called on a Sparse array?

Lord-McSweeney (approved):
Approved except for above comment

core/src/avm2/array.rs
CUB3D: Done

core/src/avm2/amf.rs
CUB3D: Done

core/src/avm2/array.rs
```diff
@@ -114,6 +114,25 @@ impl<'gc> ArrayStorage<'gc> {
         }
     }
 
+    /// Replace the existing dense storage with a new dense storage
+    /// Does nothing if this `ArrayStorage` is sparse
```
torokati44, line 118: This comment is now outdated due to the added `panic!`.
I'm just gonna fix this and merge.
Thank you!

torokati44: ... also squashed this last "chore" commit into the first one.

PR #18061, opened 2024-09-24
chore: Bump some indirect dependencies in `Cargo.lock`

torokati44:
Since https://github.com/ruffle-rs/ruffle/pull/18048 got replaced by https://github.com/ruffle-rs/ruffle/pull/18065 after https://github.com/toml-rs/toml/pull/788, this is just a careful `Cargo.lock` bump now.


torokati44:
This also drops the `tower` and `tower-sys` crates from the tree due to https://github.com/hyperium/hyper-util/pull/151.

danielhjacobs (approved):

PR #18066, opened 2024-09-24
desktop: Update dialog texts to ask for intent

kjarosh:
We do not want to question the user's intent, we want to ask what their intent is.

That will unfortunately invalidate all translations :(

CC @Dinnerbone @torokati44


n0samu (approved):
Invalidating translations does suck, but I think it's worth it for this. Thank you!

torokati44:
One of the messages is one sentence, the other is two. :no_mouth:

n0samu:
Sorry, I guess we missed our chance to fix that 😅

PR #18080, opened 2024-09-26
desktop: Update app icon on Linux

kjarosh:
This new icon conforms with icon guidelines of Flathub, GNOME, and KDE. Mainly it introduces margin so that the icon does not fill the whole canvas.

See also https://github.com/flathub/flathub/issues/5633


torokati44 (approved):
I like it, thank you! c:

kjarosh:
Just fixed some sub-pixel inconsistencies, should be even better now!

PR #18083, opened 2024-09-26
web: Support gamepad button input

evilpie:
Mostly just a proof of concept right now. This doesn't actually support defining the button to keycode mappings yet. I suspect we want to think a bit deeper about how that configuration would look like before adding it  to the Ruffle API.

### DIFF
--- a/core/src/events.rs
+++ b/core/src/events.rs
@@ -1,4 +1,5 @@
 use crate::display_object::InteractiveObject;
+use std::str::FromStr;
 use swf::ClipEventFlag;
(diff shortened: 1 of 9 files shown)


evilpie:
`gamepadButtonMapping` is now a new  _experimental_ option. I think in the future we might want to support multiple gamepads and also some handling for stick axis. Maybe it would even be better to name the option something like `gamepadButtonMapping`.

Lord-McSweeney (approved):

Issue #18103, opened 2024-09-29
Request - readGraphicsData()


olamcbola:
A request for the implementation of readGraphicsData
https://airsdk.dev/reference/actionscript/3.0/flash/display/Graphics.html#readGraphicsData()

It is used to draw repeating fabric patterns onto vector shapes, in games such as this one:
https://www.dolldivine.com/fairy-tale-high

Also drawGraphicsData and GraphicsBitmapFill if those don't exist yet.

PR #18104, opened 2024-09-29
avm2: Partially implement Graphics.readGraphicsData

evilpie:
Improves #18103

### DIFF
--- a/core/src/display_object/graphic.rs
+++ b/core/src/display_object/graphic.rs
@@ -104,7 +104,8 @@ impl<'gc> Graphic<'gc> {
 
     pub fn drawing(&self, gc_context: &Mutation<'gc>) -> RefMut<'_, Drawing> {
         RefMut::map(self.0.write(gc_context), |w| {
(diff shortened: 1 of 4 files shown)


core/src/display_object/graphic.rs
```diff
@@ -104,7 +104,8 @@ impl<'gc> Graphic<'gc> {
 
     pub fn drawing(&self, gc_context: &Mutation<'gc>) -> RefMut<'_, Drawing> {
         RefMut::map(self.0.write(gc_context), |w| {
-            w.drawing.get_or_insert_with(Drawing::new)
+            w.drawing
+                .get_or_insert_with(|| Drawing::from_swf_shape(&w.static_data.shape))
```
evilpie, line 108: I am curious if this is going to break anything. But this is required to get the drawing commands for the SWF shapes in #18103.

evilpie:
Ah, this doesn't really work with some of the more complex patterns like the arrow on pink. I don't really see any "unimplemented" message though.

PR #18105, opened 2024-09-29
core: Improve ergonomics by replacing `GcContext` by `StringContext`

moulins:
Supercedes #18081.

Embeds a `&Mutation<'gc>` directly into the string interner context (which is renamed from `GcContext` to `StringContext`, for clarity), and makes it accessible in `UpdateContext` without the need for an explicit reborrow method.

This simplifies most usages, as we don't need to pass a `&Mutation<'gc>` separately.

### DIFF
--- a/core/src/avm2/value.rs
+++ b/core/src/avm2/value.rs
@@ -521,7 +521,7 @@ pub fn abc_default_value<'gc>(
         AbcDefaultValue::Uint(u) => abc_uint(translation_unit, *u).map(|v| v.into()),
         AbcDefaultValue::Double(d) => abc_double(translation_unit, *d).map(|v| v.into()),
         AbcDefaultValue::String(s) => translation_unit
-            .pool_string(s.0, &mut activation.borrow_gc())
+            .pool_string(s.0, activation.strings())
             .map(Into::into),
         AbcDefaultValue::True => Ok(true.into()),
         AbcDefaultValue::False => Ok(false.into()),
@@ -841,6 +841,7 @@ impl<'gc> Value<'gc> {
                 if *i >= 0 && *i < 10 {
                     activation
                         .context
+                        .strings
                         .interner
                         .get_char(activation.context.gc_context, '0' as u16 + *i as u16)
                 } else {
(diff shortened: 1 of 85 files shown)


core/src/string/context.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    #[must_use]
+    pub fn get_char(&self, c: u16) -> AvmString<'gc> {
+        if let Some(s) = self.interner.chars.get(c as usize) {
+            (*s).into()
+        } else {
+            AvmString::new(self.gc(), WString::from_unit(c))
+        }
+    }
```
Lord-McSweeney, line 85: Can we keep `get_ascii_char`? It's likely to be used in many more places, and we could pass a `char` directly to `get_ascii_char`, so we can do stuff like `activation.strings().get_ascii_char('0')` without the additional u16 conversion. Also, an allocation path slows down the AVM2 stack, so I would guess (though I haven't actually tested) this function without the allocation path would be faster too.

core/src/avm2/value.rs
```diff
@@ -839,10 +839,7 @@ impl<'gc> Value<'gc> {
             }
             Value::Integer(i) => {
                 if *i >= 0 && *i < 10 {
-                    activation
-                        .context
-                        .interner
-                        .get_char(activation.context.gc_context, '0' as u16 + *i as u16)
+                    activation.strings().get_char('0' as u16 + *i as u16)
```
Lord-McSweeney, line 842: For example, this could use `get_ascii_char`

Lord-McSweeney (approved):

core/src/string/context.rs
moulins: Good point, I'll keep it then (but probably change it to take an `u8` instead, it makes more sense IMHO)

Lord-McSweeney: The point of `char` was to reduce the `.into()` needed anywhere else, but taking `u8` is fine too, i guess

core/src/avm2/value.rs
moulins: I prefer leaving this change for a future PR :)

PR #18136, opened 2024-10-01
ci: Create PR for downloading translations then add labels

danielhjacobs:
There is no reason this should be necessary, but it seems like it probably is :shrug:


kjarosh (approved):

torokati44:
I know I'm late as usual, but I believe that any time a hack or a workaround like this is implemented, it's crucial that a comment is also added to explain what it does, what circumstances made it necessary, and what was the straightforward way that didn't work - so that in the future, somebody doesn't unknowingly submit a PR as "simplifying" things, essentially reverting this.

Issue #18142, opened 2024-10-02
Unnecessary permission dialog when opening an SWF file


n0samu:
### Describe the bug

To reproduce:
1. Run the Ruffle desktop app
2. Click File => Open and choose any SWF file

Ruffle will ask for permission to access the file that you just chose to open. This is annoying and unnecessary.
![image](https://github.com/user-attachments/assets/5c500d75-6872-42ae-9b46-1a730457936d)

### Expected behavior

Ruffle should only ask for permission before accessing other files that the user didn't explicitly choose to open.

### Content Location

N/A

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

Related to #17875. According to @Lord-McSweeney, this issue is *not* reproducible on Linux. So it may be a separate issue from #18141.

kjarosh:
This bug has been fixed and the dialog should not show for the SWF file you're opening. For other files it's not possible yet to allow access and remember it for more than the session.

kjarosh:
You can file an issue about that, I think eventually we want it done, but it requires also a UI that lists all persisted permissions and allows removing them.

Currently the dialog should be shown only when the SWF tries to access a weird location.

hamzaelbarmaki1994:
@kjarosh  done : https://github.com/ruffle-rs/ruffle/issues/19790

Issue #18141, opened 2024-10-02
Cannot run SWFs from the command line


n0samu:
### Describe the bug

As of nightly 2024-09-30, running SWFs in Ruffle from the command line no longer works; the Ruffle window never appears. For example, running `ruffle test.swf` results in the Ruffle window never appearing.

### Expected behavior

Ruffle should open and play the specified SWF.

### Content Location

N/A

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

I've bisected this regression to commit 4eb66c0485ccc3a8965bbd51842b9a1b8d54fcb5 (#17875)

@Lord-McSweeney confirms that this bug is reproducible on Linux.

PR #18147, opened 2024-10-02
desktop: Fix filesystem access dialog

kjarosh:
This patch fixes the following two issues:

1. On Windows the allowed path check was not always working, due to Windows paths being Windows paths, now paths are canonicalized before comparison.

2. On some systems the dialog was not rendered when the main SWF wasn't loaded, and the dialog was not able to allow the main SWF. Now the dialog allows the SWF instantly when constructed.

That needs testing from @n0samu and @Lord-McSweeney, as I cannot reproduce these issues.

Fixes https://github.com/ruffle-rs/ruffle/issues/18141, fixes https://github.com/ruffle-rs/ruffle/issues/18142.

### DIFF
--- a/desktop/src/backends.rs
+++ b/desktop/src/backends.rs
@@ -6,4 +6,5 @@ mod ui;
 pub use external_interface::DesktopExternalInterfaceProvider;
 pub use fscommand::DesktopFSCommandProvider;
 pub use navigator::DesktopNavigatorInterface;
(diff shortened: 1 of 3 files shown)


n0samu:
This does not fix #18141 for me 😕 
However, it does fix #18142.

adrian17 (approved):

PR #18158, opened 2024-10-03
core: Use the canonical empty `AvmString` instead of `"".into()`

moulins:
Some more progress towards removing `'static` strings.

This mostly involves passing a `&StringContext` in a few more places, together with some local refactors to make the borrow-checker happy.

Notable changes:
- `avm2::TObject::value_of` now takes a `&StringContext`, and has a default impl that returns `self` as a `Value`;
- `Namespace::as_uri` takes a `&StringContext` too, in some cases I rewrote the calling logic to use `as_uri_opt` instead.

### DIFF
--- a/core/src/avm2/object/namespace_object.rs
+++ b/core/src/avm2/object/namespace_object.rs
@@ -24,11 +24,8 @@ pub fn namespace_allocator<'gc>(
         NamespaceObjectData {
             base,
             namespace: Lock::new(namespace),
-            prefix: Lock::new(if namespace.as_uri().is_empty() {
-                Some(activation.strings().empty())
-            } else {
-                None
-            }),
+            // The public namespace has no prefix
+            prefix: Lock::new(Some(activation.strings().empty())),
         },
     ))
     .into())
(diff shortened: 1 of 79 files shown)


Lord-McSweeney (approved):

core/src/avm2/object/namespace_object.rs
```diff
@@ -84,8 +81,8 @@ impl<'gc> NamespaceObject<'gc> {
             NamespaceObjectData {
                 base,
                 namespace: Lock::new(namespace),
-                prefix: Lock::new(if namespace.as_uri().is_empty() {
-                    Some("".into())
+                prefix: Lock::new(if namespace.as_uri(activation.strings()).is_empty() {
+                    Some(activation.strings().empty())
```
Lord-McSweeney, line 85: This seems a bit redundant, but I don't see a better solution

moulins: We could have a dedicated method `namespace.has_empty_uri()`, but I'm not sure this is worth it.

PR #17401, opened 2024-08-07
core,desktop: Update to `winit` `v0.30.5` and latest `egui` from git

torokati44:
Still needs some cleanup and porting and migrating away from deprecated stuff to the new and shiny new trait-based APIs I think.

~~Not latest minor version, because:~~
~~https://github.com/emilk/egui/pull/4849#issuecomment-2258111041~~

~~But soon it probably can be:~~
~~https://github.com/rust-windowing/winit/pull/3849~~

Supersedes https://github.com/ruffle-rs/ruffle/pull/16835.

PR #18193, opened 2024-10-05
desktop: Migrate to winit 0.30's ApplicationHandler

Dinnerbone:
This is one of the TODOs from #17401

Needs testing please! Largest change is to ControlFlow (specifically, how often we redraw things). In the previous update model, we'd force it to wait until the next frame time after every event we receive. Now, we conditionally do it at the end of a batch of events.

There's still another pending TODO (make the window later, instead of before event loop) - but that's a larger and messier change.
We can also clean up the ControlFlow usage further by using a callback in egui, but I didn't want to risk more changes than needed.

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -15,29 +15,38 @@ use std::rc::Rc;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 use url::Url;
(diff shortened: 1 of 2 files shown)


kjarosh:
Tested on Wayland & X11, and everything seems to work properly

PR #18201, opened 2024-10-07
web: Fix programmatic focusing and context menu paste

kjarosh:
This is a follow up to the original patch that changes how Ruffle is focused (https://github.com/ruffle-rs/ruffle/pull/17158). It changes the focusable element from the container to the shadow host:

1. It fixes the context menu, which is outside the container. Without it for instance paste wasn't working properly, because Ruffle lost its focus when clicking "Paste" and had nowhere to paste the text.

2. It fixes programmatic focusing, which may be used by websites for e.g. implementing custom input overlays and simulating input.

These two cases (and more) are covered by the newly added web integration tests: `context_menu` & `programmatic_events`.

CC @n0samu @danielhjacobs


kjarosh:
The context menu test is failing on Chrome, I'll need to investigate that.

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -201,7 +201,7 @@ export class InnerPlayer {
         this.debugPlayerInfo = debugPlayerInfo;
         this.onCallbackAvailable = onCallbackAvailable;
 
-        this.shadow = this.element.attachShadow({ mode: "open" });
+        this.shadow = this.element.attachShadow({ mode: "open", delegatesFocus: true });
```
kjarosh, line 204: This requires Firefox 94

danielhjacobs: That's fine, it's almost three years old and it's not like Safari where it's tied to OS version.

danielhjacobs: The bigger concern is that this requires Safari 15 but that's probably fine too.

kjarosh: Update: `delegatesFocus` from `attachShadow` has been supported since Safari 13.1, so it's alright (it's only the getter that requires 15).

Dinnerbone:
The programmatic test fails on ios 17 - I think that's a browser bug so maybe we should just ignore the test there. The only warning I can see is "The requested resource could not be found, or a request was received using an HTTP method that is not supported by the mapped resource" - which isn't very helpful.

The context menu test fails most of the time with either random wdio bugs (inserting the text `WDIO_CONTROLv`) or that the browser suddenly asks for permission to use the clipboard. I think let's ignore that one too, because damn I don't have energy to babysit browsers 😓

Dinnerbone (approved):

PR #18205, opened 2024-10-07
desktop: Use ashpd for dbus communication

kjarosh:
That's a refactor that simplifies our code a lot.

I would appreciate some tests to make sure it does not break anything.

### DIFF
--- a/desktop/src/dbus.rs
+++ b/desktop/src/dbus.rs
@@ -4,83 +4,28 @@
 use std::mem;
 use std::sync::{Arc, Mutex};
 
-use futures::StreamExt;
-use zbus::export::futures_core::Stream;
-use zbus::zvariant::{OwnedValue, Value};
-use zbus::{proxy, Connection};
+use ashpd::desktop::settings::{ColorScheme, Settings};
+use futures::Stream;
 
 type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;
 
-#[proxy(
-    interface = "org.freedesktop.portal.Settings",
-    default_service = "org.freedesktop.portal.Desktop",
-    default_path = "/org/freedesktop/portal/desktop",
-    gen_blocking = false
-)]
-trait FreedesktopSettingsInterface {
-    fn read(&self, namespace: &str, key: &str) -> Result<OwnedValue>;
-
-    #[zbus(signal)]
-    fn setting_changed(&self, namespace: &str, key: &str, value: Value<'_>);
+pub struct FreedesktopSettings {
+    proxy: Settings<'static>,
 }
 
-pub enum ColorScheme {
-    Default,
-    PreferLight,
-    PreferDark,
-}
-
-#[derive(thiserror::Error, Debug)]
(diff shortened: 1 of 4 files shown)


desktop/src/gui/theme.rs
```diff
@@ -171,10 +160,13 @@ impl ThemeController {
 }
 
 #[cfg(target_os = "linux")]
-fn scheme_to_theme(color_scheme: ColorScheme) -> Theme {
-    use crate::dbus::ColorScheme;
+fn scheme_to_theme(color_scheme: ashpd::desktop::settings::ColorScheme) -> Theme {
+    use ashpd::desktop::settings::ColorScheme;
     match color_scheme {
-        ColorScheme::Default => Theme::Light,
+        // We do not select the theme of window decorations,
```
torokati44, line 166: Could you please clarify what you mean by this?

torokati44: I mean... we don't select it "here", we don't _want_ to select it, we "can't" select it...?

torokati44: Note, on a Wayland compositor with no support for server-side decorations, I believe we have the `sctk` decorations, which does have a dark mode:
https://github.com/PolyMeilex/sctk-adwaita

kjarosh: > Could you please clarify what you mean by this?
>
> I mean... we don't select it "here", we don't _want_ to select it, we "can't" select it...?

It means that we just don't do it. We *can* do it for sure, but I haven't checked whether it is feasible or how easy it is.

When we don't select it, it means we have to assume it's light in order not to desync decorations with GUI. There are apps that use dark mode without dark preference, and will use only light mode when strictly required (e.g. GNOME terminal, it's dark even when dark mode is disabled).

Note that this does not change the behavior, just adds a comment explaining the situation.

kjarosh: Turns out it's fairly simple: https://github.com/ruffle-rs/ruffle/pull/18213

desktop/src/dbus.rs
```diff
@@ -4,83 +4,28 @@
...
-    Default,
-    PreferLight,
-    PreferDark,
-}
-
-#[derive(thiserror::Error, Debug)]
-pub enum FreedesktopSettingsError {
-    #[error("Unexpected value for color scheme: {0}")]
-    UnexpectedColorScheme(Value<'static>),
-}
```
torokati44, line 37: Yay for less boilerplate!

torokati44 (approved):
The code looks fine, I appreciate the simplification.
I assume you have tested it and nothing regressed.
Seems like the `ashpd` release cadence is a bit "meh" _(no worries, it's FOSS after all!)_, and at least this lessens the likelihood of `zbus` dep duplication due to pulling it in both directly and indirectly.
Thank you! ^^

torokati44:
This is on this branch - beautiful!

https://github.com/user-attachments/assets/226c20b2-0e00-49c8-9b22-946ec7a9faef

PR #18213, opened 2024-10-08
desktop: Synchronize window theme on Linux

kjarosh:
Set window decorations theme so that they are always synchronized with the GUI theme on Linux. Note that this does not apply to other OSes, as we rely on window theme changing automatically there.

This fixes window decorations desync on Linux Wayland, which is caused by sctk-adwaita not watching on system theme changes.


torokati44 (approved):
While I can only see this actually taking effect in Weston (where we're forced to use client-side decorations), it's still nice! ^^

kjarosh:
Oh yeah, forgot to mention that this affects only CSD!

torokati44:
And I guess it's up to `winit` to make it work with SSD, if they want to, and if it's possible at all, etc...

PR #18218, opened 2024-10-08
chore: Preserve translations (but unapprove them) when strings change

Dinnerbone:
Previously if a string changed, all translations get removed.
With this, they'll stay - but it'll unapprove the translations as they may not be strictly accurate anymore.


kjarosh:
What about the web extension messages file? (Can't comment in code unfortunately)

kjarosh (approved):

PR #18223, opened 2024-10-09
core: Clean up ExternalInterface a little

Dinnerbone:
Instead of getting a method then calling it, we just call things directly now. We also can now only have one Provider instead of many, but realistically there's no case in which multiple providers can peacefully coexist as they'll be conflicting with methods.

### DIFF
--- a/core/src/avm1/globals/external_interface.rs
+++ b/core/src/avm1/globals/external_interface.rs
@@ -4,7 +4,7 @@ use crate::avm1::activation::Activation;
 use crate::avm1::error::Error;
 use crate::avm1::property_decl::{define_properties_on, Declaration};
 use crate::avm1::{Object, ScriptObject, Value};
-use crate::external::{Callback, Value as ExternalValue};
+use crate::external::{Callback, ExternalInterface, Value as ExternalValue};
 use crate::string::StringContext;
 
 const OBJECT_DECLS: &[Declaration] = declare_properties! {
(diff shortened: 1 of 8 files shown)


core/src/avm1/globals/external_interface.rs
```diff
@@ -64,23 +64,16 @@ pub fn call<'gc>(
...
-        .context
-        .external_interface
-        .get_method_for(&name.to_utf8_lossy())
-    {
-        let mut external_args = Vec::with_capacity(external_args_len);
-        if external_args_len > 0 {
-            for arg in &args[1..] {
-                external_args.push(ExternalValue::from_avm1(activation, arg.to_owned())?);
-            }
+    let mut external_args = Vec::with_capacity(external_args_len);
```
evilpie, line 67: I would probably write this using an iterator an map.

core/src/avm2/globals/flash/external/external_interface.rs
```diff
@@ -12,21 +12,14 @@ pub fn call<'gc>(
...
-        for arg in &args[1..] {
-            external_args.push(ExternalValue::from_avm2(arg.to_owned()));
-        }
-        Ok(method
-            .call(activation.context, &external_args)
-            .into_avm2(activation))
-    } else {
-        Ok(Value::Null)
+    let mut external_args = Vec::with_capacity(args.len() - 1);
+    for arg in &args[1..] {
```
evilpie, line 16: Same here.

desktop/src/backends/external_interface.rs
```diff
@@ -1,62 +1,49 @@
...
 fn is_location_href(code: &str) -> bool {
     matches!(
         code,
         "document.location.href" | "window.location.href" | "top.location.href"
     )
 }
 
-struct FakeEval(Option<Url>);
-
-impl ExternalInterfaceMethod for FakeEval {
```
evilpie, line 28: Nice cleanup :tada:

evilpie (approved):

core/src/avm1/globals/external_interface.rs
Dinnerbone: This was old code moved up, so I didn't really touch it - but good idea, changed :)

PR #18225, opened 2024-10-09
desktop: Create the window after the event loop

Dinnerbone:
This resolves the remaining TODO from #17401

Winit wants us to create the window after the event loop starts, and that wasn't very easy - but here we are.

Kindly requesting that someone tests on Wayland & X11 to make sure it's fine still :D


torokati44:
(eager to see whether this finally resolves #8870 👀)

torokati44:
Couldn't spot any regressions in a quick test on X11 (XWayland) or Wayland (KDE and Weston). :tada:

adrian17 (approved):

PR #18229, opened 2024-10-10
web: PluginsArray and MimeTypeArray prototype spoofing

evilpie:
This is enough to fool https://fingerprintjs.github.io/BotD/main/, but doesn't require a huge refactoring.

The only annoying part is faking `MimeType`, which is not a constructable type in the DOM, so we have to create a new wrapper for it.

This is a replacement for #18157.


Dinnerbone:
Do you think we can add a test with that script on it, like how we check that we fool swfobject?

Dinnerbone (approved):
Thanks (and thanks @lihe07!)

PR #18233, opened 2024-10-10
desktop: Move open URLs mode from launch options to preferences

kjarosh:
This PR moves open links behavior preference from launch options (which was settable before running each SWF) to the preferences (which is settable at any point in time and is persistent).

Additionally, this PR changes the default to "confirm". Users who don't want the confirmation popup to be displayed all the time can disable it, and others will have an additional layer of security on by default.

Opening links without confirmation is unsafe because:

1. It shares your IP address and other information.

2. It might be used to leak sensitive information in the URL.

By having the confirmation dialog on by default, users will have a chance to check if the link is safe before opening it.

Fixes https://github.com/ruffle-rs/ruffle/issues/17432.

CC @n0samu

@torokati44 This PR changes a bit the frontend utils API which gives more control to the implementations over how links are opened, but requires adding a simple check by them to preserve the same behavior as before.

### DIFF
--- a/desktop/src/cli.rs
+++ b/desktop/src/cli.rs
@@ -192,8 +192,8 @@ pub struct Opt {
     pub frame_rate: Option<f64>,
 
     /// The handling mode of links opening a new website.
-    #[clap(long, default_value = "allow")]
-    pub open_url_mode: OpenURLMode,
+    #[clap(long)]
+    pub open_url_mode: Option<OpenUrlMode>,
 
     /// How to handle non-interactive filesystem access.
     #[clap(long, default_value = "ask")]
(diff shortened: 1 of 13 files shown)


torokati44:
> @torokati44 This PR changes a bit the frontend utils API which gives more control to the implementations over how links are opened, but requires adding a simple check by them to preserve the same behavior as before.

I assume you noted this due to the Android app - thanks for the heads-up!

desktop/src/cli.rs
```diff
@@ -192,8 +192,8 @@ pub struct Opt {
     pub frame_rate: Option<f64>,
 
     /// The handling mode of links opening a new website.
```
evilpie, line 194: I don't think it's a great style to have an implicit default mode (confirm) that can't be explicitly selected.

kjarosh: What do you mean? It can be selected by the user here or in preferences

evilpie (approved):

desktop/src/cli.rs
kjarosh: That's right!

Dinnerbone (approved):
I think I was initially hesitant about changing the default, but, I'm fine with it. It's good.

Thanks!

Issue #1486, opened 2020-11-04
Add setting to spoof URLs in network requests


Herschel:
Add a setting to pattern match and replace URLs in network requests. By default we currently auto-upgrade HTTP to HTTPS, but a more general URL-rewrite mechanism would be useful.

This could be used as a workaround for issues with CORS, allowing files to be loaded from a different location than was originally coded into the SWF, and allow for redirecting dead services to a new reimplementation (Mochi high scores, etc.).

Herschel:
Re-opening because it would still be useful to allow more advanced URL spoofing, although it's not as pressing with `upgradeToHttps` implemented.

AyrA:
I'm also interested in this.

Just saying, that as a temporary solution until this is properly implemented, you can simply hijack the fetch calls:

```js
(function (originalFetch) {
	const changeUrl = function (url) {
		//TODO: Check url parameter and modify as needed
		return url;
	};
	window.fetch = function () {
		let a = Array.from(arguments);
		if (typeof(a[0]) === "string") {
			//Argument to fetch() call is a raw URL string
			a[0] = changeUrl(a[0]);
		} else if (a[0] && typeof(a[0].url) === "string") {
			//Argument to fetch() call is a request object
			//HACK: This replaces the entire request object with an URL
			//      because the Request.url property is readonly.
			//      This is appropriate for GET requests only.
			const changedUrl = changeUrl(a[0].url);
			if (changedUrl !== a[0].url) {
				a[0] = changedUrl;
			}
		}
		return originalFetch.apply(window, a);
	};
})(window.fetch);
```

This code is good enough to rewrite simple GET requests which is usually sufficient to redirect requests that pull resources at runtime. For true interception you would need to reproduce the request object, but I'm not sure if you can reliably do that for POST requests with body data.

woeterman94:
> I'm also interested in this.
> 
> Just saying, that as a temporary solution until this is properly implemented, you can simply hijack the fetch calls:
> 
> ```js
> (function (originalFetch) {
> 	const changeUrl = function (url) {
> 		//TODO: Check url parameter and modify as needed
> 		return url;
> 	};
> 	window.fetch = function () {
> 		let a = Array.from(arguments);
> 		if (typeof(a[0]) === "string") {
> 			//Argument to fetch() call is a raw URL string
> 			a[0] = changeUrl(a[0]);
> 		} else if (a[0] && typeof(a[0].url) === "string") {
[...]

PR #18252, opened 2024-10-13
web: Implement URL rewrite rules

kjarosh:
These rules are configurable and they allow rewriting URLs used in SWFs. The main rationale behind them is to rewrite obsolete links to working ones in order to preserve old content.

In order to add a rewrite rule, add a `urlRewriteRules` config option which is an array of rewrite rules.  Each rule consists of a matcher (either a string or a regex), and a replacement, e.g.:

```js
    "urlRewriteRules": [
        // Rewrites to a relative URL
        [/^https?:\/\/example.com\/(.*)\.swf$/, "$1.swf"],
        // Rewrites to an absolute URL
        [/^(.*)\.test\.swf$/, "$1.swf"],
        // Rewrites using a string (exact match, not a regex)
        ["http://old.example.com/site.html", "http://new.example.com/other_site.html"],
    ],
};
```

Related to https://github.com/ruffle-rs/ruffle/issues/1486.


MrCheeze:
Love to see it! Just one thing - can we say more explicitly whether this rewrites URLs in urlrequests, when navigating to a URL, or both? (I assume both, since that would be the most useful.)

kjarosh:
Yes, it will rewrite URLs in both network requests and links. Added that to the doc.

I originally needed URL rewrite for links only, but I figured that there must be content that needs to have all URLs rewritten in order to work.

adrian17 (approved):

web/packages/selfhosted/test/utils.ts
```diff
@@ -200,3 +200,28 @@ export function loadJsAPI(swf?: string) {
...
+) {
+    await browser.execute(
+        (modals) => {
+            for (const modal of modals) {
+                const m = modal as unknown as HTMLElement;
+                const cl = m.querySelector(".close-modal")! as HTMLElement;
+                cl.click();
+            }
+        },
+        await player.shadow$$(".modal:not(.hidden)"),
```
danielhjacobs, line 216: webdriverio v9 automatically pierces the shadow DOM.

kjarosh: Should be good now!

farves:
This code works fine, but it doesn't Rewrite https://old.mysite.com/2.swf to https://mysite.com/2.swf It gets https://old.mysite.com/2.swf with Status Code: 200 OK, but blocks it by CORS policy.  Where is the problem?

```
const script=document.createElement('script');
script.src='/ruffle/ruffle.js';
script.onload=function() {
play();
}
document.body.appendChild(script);
function play(){
const ruffle=window.RufflePlayer.newest();
const player=ruffle.createPlayer();
player.config={
urlRewriteRules": ["https://old.mysite.com/2.swf", "https://mysite.com/2.swf"],
contextMenu:"off",
autoplay:"on",
warnOnUnsupportedContent:false,
allowScriptAccess:true,
maxExecutionDuration:30,
base:"https://mysite.com",
openUrlMode:"deny",
unmuteOverlay:"hidden"
};
const container = document.getElementById('container');
container.appendChild(player);
player.load('https://mysite.com/1.swf');
```

kjarosh:
@farves `urlRewriteRules` is an array of arrays, see the example from the description

farves:
@kjarosh even if I do 
```
....other code
player.config={
urlRewriteRules: [
["https://old.mysite.com/2.swf", "https://mysite.com/2.swf"],
["https://old.mysite.com/3.swf", "https://mysite.com/3.swf"],
],
....other code
```
it doesn't Rewrite https://old.mysite.com/2.swf to https://mysite.com/2.swf

PR #18257, opened 2024-10-13
render: Skip blending if there's nothing to draw

adrian17:
Noticed on `The Last Stand: Union City` that it tries to draw out-of-frame street lights; the draw is culled, but the blending logic doesn't know that.
This won't improve max FPS if everything is in frame, but should save some GPU resources when there's blended stuff out of frame.


Dinnerbone (approved):
Nice!

PR #18258, opened 2024-10-13
debug_ui: Improve layout debug boxes

kjarosh:
Displaying all at once would be a mess, so this patch adds a way of enabling and disabling specific types of layout debug boxes.

Hopefully we'll be able to reduce the number of different kinds of bounds soon, but unless it's done, displaying any of them is very useful for debugging.

image


Dinnerbone (approved):
Love debug improvements :D Thank you!

PR #18262, opened 2024-10-13
text: Cull text lines above text field

kjarosh:
We do not need to render text lines which are above the text field, as they are masked anyway.

### DIFF
--- a/core/src/display_object/edit_text.rs
+++ b/core/src/display_object/edit_text.rs
@@ -943,6 +943,12 @@ impl<'gc> EditText<'gc> {
         }
     }
 
+    fn render_layout_line(self, context: &mut RenderContext<'_, 'gc>, line: &LayoutLine<'gc>) {
+        for layout_box in line.boxes_iter() {
+            self.render_layout_box(context, layout_box);
+        }
+    }
+
     /// Render a layout box, plus its children.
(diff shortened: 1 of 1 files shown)


adrian17 (approved):

core/src/display_object/edit_text.rs
```diff
@@ -953,9 +959,8 @@ impl<'gc> EditText<'gc> {
         // If text's top is under the textbox's bottom, skip drawing.
         // TODO: FP actually skips drawing a line as soon as its bottom is under the textbox;
         //   Current logic is conservative for safety (and even of this I'm not 100% sure).
-        // TODO: we should also cull text that's above the textbox
-        //   (instead of culling, this can be implemented as having the loop start from `scrollY`th line)
         //   (maybe we could cull-before-render all glyphs, thus removing the need for masking?)
+        //   [KJ] FP always displays the first line, culls any line which is not fully visible
```
torokati44, line 963: What do you mean by "the first line" here? :thinking:

kjarosh: Elaborated a bit about it in the comment, I mean the first visible line (taking into account vscroll).

torokati44: Yeah, that was what I guessed as only this makes logical sense to me; but the comment didn't clearly mean that to me - thank you for clarifying it!

PR #18272, opened 2024-10-14
text: Take the gutter into account when masking text

kjarosh:
Flash Player does not render text over the gutter.


kjarosh:
Turns out we'll need to render the caret a bit differently, as it shouldn't be affected by the mask.

Lord-McSweeney (approved):

PR #18286, opened 2024-10-16
chore: Fix errors reported by clippy beta

kjarosh:
This fixes errors like

    error: the following explicit lifetimes could be elided: 'a
    error: empty line after doc comment
    error: unneeded `return` statement

Plus others related to doc comments.

### DIFF
--- a/core/src/bitmap/turbulence.rs
+++ b/core/src/bitmap/turbulence.rs
@@ -1,7 +1,8 @@
-/// This file is a Rust port of the C reference implementation of the
-/// feTurbulence element in the SVG specification. It's the usual Perlin noise.
-/// See: https://www.w3.org/TR/SVG11/filters.html#feTurbulenceElement
-/// The `octave_offsets` parameter of `turbulence` was added after porting.
+//! This file is a Rust port of the C reference implementation of the
+//! feTurbulence element in the SVG specification. It's the usual Perlin noise.
+//!
+//! See: <https://www.w3.org/TR/SVG11/filters.html#feTurbulenceElement>.
+//! The `octave_offsets` parameter of `turbulence` was added after porting.
 
 // Copyright © 2015 W3C® (MIT, ERCIM, Keio, Beihang).
 // This software or document includes material copied from or derived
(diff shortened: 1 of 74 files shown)


torokati44:
There are a few more with `'gc`, and there's also a `clippy::empty-line-after-doc-comments`.

kjarosh:
Whoops, I forgot to push \;)

torokati44:
Another one... :)

torokati44:
They just won't run out... 😅

torokati44:
About the (current) second commit: https://github.com/rust-lang/rust-clippy/issues/12917 may be related.

kjarosh:
> About the (current) second commit: [rust-lang/rust-clippy#12917](https://github.com/rust-lang/rust-clippy/issues/12917) may be related.

To be honest, we should have just commented out that doc comment. It does not apply to anything as the original code has been commented out too.

torokati44:
Yeah, but nobody suggested it until now, and even when I saw at first what you did now, it struck me as really weird, but given that "doc comments" are almost like proper language elements in Rust, it does make some weird sense to "actually comment them out", now that I think about it...

torokati44:
_Holy mother of wide-reaching nit diffs, Crabman!_

core/src/bitmap/turbulence.rs
```diff
@@ -38,9 +39,11 @@ fn random(seed: i64) -> i64 {
...
 #[derive(Copy, Clone)]
 struct StitchInfo {
-    width: i32, // How much to subtract to wrap for stitching.
+    /// How much to subtract to wrap for stitching.
+    width: i32,
     height: i32,
-    wrap_x: i32, // Minimum value to wrap.
+    /// Minimum value to wrap.
+    wrap_x: i32,
     wrap_y: i32,
```
torokati44, line 47: Isn't this going to be weird that the doc comments are only for the first half of each pair? :thinking:

kjarosh: Oh, I didn't notice those comment referred to more than one field, TBH it was weird before too... :/

I can move those comments to the doc comment of the struct or add x/y width/height variants (or something else perhaps?), which one do you prefer?

kjarosh: Done

torokati44: Thanks! ^^

torokati44 (approved):
LGTM, thank you! :tada:
I guess this is better merged sooner rather than later, so it doesn't start bitrotting on us...

torokati44:
Oh, sorry, there's a missing space in one of the commit messages: ` out,othwerise` :P

torokati44:
That stray `EditText` comment could also be a merge conflict resolution artifact BTW.

PR #18299, opened 2024-10-17
extension: Disable Ruffle on Duo Security

danielhjacobs:
My organization uses https://outlook.office.com secured by Duo Security, and they use Duo Security's check for Flash installation to ensure you don't log in with Flash installed. Unfortunately or fortunately depending on how you look at it, Ruffle is detected as Flash. We can't avoid that without fooling fewer plugin detection scripts used to enable Flash content, so we instead have to disable Ruffle on this site.

### DIFF
--- a/web/packages/extension/manifest.json5
+++ b/web/packages/extension/manifest.json5
@@ -22,6 +22,7 @@
                 "https://*.time4learning.com/*",
                 "https://*.edgenuity.com/*",
                 "https://www.chewy.com/*",
+                "https://*.duosecurity.com/*",
             ],
             "js": ["dist/content.js"],
             "all_frames": true,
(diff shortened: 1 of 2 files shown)


danielhjacobs:
More info about this is available here: https://duo.com/docs/policy#plugins

Dinnerbone (approved):
We should probably start marking all these exceptions with relevant issues/PRs explaining them 😅 the list is getting long

danielhjacobs:
I can link to this PR or directly to https://duo.com/docs/policy#plugins. Which do you think is better for this?

Dinnerbone:
I think the PR

danielhjacobs:
> We should probably also find a way to merge the lists...
> But this doesn't have to be done for this PR at all!

I'll save that part for the future, since combining ts and json5 sounds difficult, but I added the links to issues/PRs

Issue #18038, opened 2024-09-22
There will be a large offset when you click on the Sprite


aschord:
### Describe the bug

Because SpriteSheet technology is used in the game as a display, it is normal to click the flag when playing demo game.swf with FlashPlayer, but it will be greatly offset when playing the flag with RufflePlayer (presumably because the Bitmap not displayed is also considered as the click range).  The click range is not accurate

### Expected behavior

I want it to be as accurate when I click on the flag as it is in FlashPlayer

### Content Location

**BUG example file source code :** [demo game.swf](https://github.com/aschord/RuffleTest/blob/master/demo%20game.swf)   AND  [demo game src.7z](https://github.com/aschord/RuffleTest/blob/master/demo%20game%20src.7z)

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

Google Chrome 93

kjarosh:
Could you describe in detail how to reproduce the problem? It's not obvious what "the flag" means. A screenshot with a specific point where to click would be also appreciated.

aschord:
> Could you describe in detail how to reproduce the problem? It's not obvious what "the flag" means. A screenshot with a specific point where to click would be also appreciated.

When starting the game, the flag is not accurate in RufflePlay, because the flag is a Bitmap object, the part that can not be seen when clicked will also have click events, resulting in a lot of click accuracy offset! You can test with RuffePlayer by clicking on the flag and see that it doesn't turn over the flag where the mouse is.

aschord:
When will this problem be solved? thank you

kjarosh:
@torokati44 FYI after clicking a flag in this SWF, the following debug assert fails (https://github.com/ruffle-rs/ruffle/pull/14352):

https://github.com/ruffle-rs/ruffle/blob/7ed23b04b9bae0b39ee2754f1e8b99e3d6790863/core/src/timer.rs#L179-L182

PR #18311, opened 2024-10-18
avm2: Fix mouse pick on non-interactive objects with masks

kjarosh:
Fixes https://github.com/ruffle-rs/ruffle/issues/18038.

Mouse pick should take into account that a non-interactive object may have a mask, in which case the hit test should also be performed on the mask.


core/src/display_object/movie_clip.rs
```diff
@@ -3248,8 +3248,10 @@ impl<'gc> TInteractiveObject<'gc> for MovieClip<'gc> {
                             Avm2MousePick::Miss
                         }
                     }
-                } else if child.as_interactive().is_none()
-                    && child.hit_test_shape(context, point, options)
+                } else if child
+                    .masker()
+                    .unwrap_or(child)
+                    .hit_test_shape(context, point, options)
```
Dinnerbone, line 3254: This needs to test both the mask _and_ the clip. If you modify the test to have the mask be 75px wide, click just to the right of the visible object and Flash says stage but Ruffle (this PR) says clip.

kjarosh:
Just fixed the commit message so that it doesn't suggest we should hit test the mask only :eyes:

Lord-McSweeney (approved):

PR #18324, opened 2024-10-19
avm1: Set ConvolutionFilter's alpha to 1 when initialized with a color

britt-j:
In AVM1, initializing a ConvolutionFilter with a substitute color will result in the substitute alpha being set to 1.
This doesn't happen in AVM2.

Tested on SWFv8 and SWFv43.


Dinnerbone (approved):
Thank you, nice catch!

kjarosh:
Fun fact: the documentation claims the default value of `alpha` is 0, but in the example a line below that claim, it says that the example changes alpha from the default value of 1 to 0.35. The constructor doc doesn't mention anything about the default value.

kjarosh (approved):

PR #18330, opened 2024-10-20
tests: Print comparison when approximation assertion fails

kjarosh:
This makes it easier to compare values and see which one failed.


torokati44 (approved):
LGTM *rubber-stamp*

PR #18332, opened 2024-10-20
text: Fix layout measurements

kjarosh:
This patch fixes various measurements made during laying out text.

1. After laying out a line, the cursor should be advanced by the value of `ascent + descent + leading`, not `font_size + font_leading + leading`.

2. Font leading is ignored.

3. Proper line bounds (not interior bounds) are now used for calculating exterior text bounds.

Additionally, it updates the following tests, as their output now is more similar to the output produced by the Flash Player:

1. `avm2/away3d_advanced_shallow_water_demo`
2. `visual/definefont4`
3. `visual/fonts/device-font`
4. `visual/shumway_acid_tests/acid`
5. `visual/shumway_acid_tests/acid_text`

In general, this PR should make a lot of text look _so much better_, especially multiline text fields.


kjarosh:
After this PR I was able to make epsilons smaller in 6 tests, and remove approximations altogether in 3 tests :tada:

PR #18348, opened 2024-10-22
tests: Improve edittext_getcharboundaries test

kjarosh:
This patch removes y precision constraint, as now Ruffle produces output consistent with Flash Player (https://github.com/ruffle-rs/ruffle/pull/18332). Additionally some test cases are added.


torokati44 (approved):
Yay, thanks! LGTM.

Issue #14713, opened 2024-01-12
Cursor selection no longer appears over whitespace


Lord-McSweeney:
### Describe the bug

Whitespace in the selection is never rendered. For example, if the text "Hello world" is selected, "Hello" and "world" will have an opaque black rectangle rendered under the text, but the space will not have it. I suspect this is due to the device font changes (not #14690), but I'm not sure.

### Expected behavior

The selection is rendered on whitespace.

### Content Location

Any basic reproduction works

### Affected platform

Desktop app

### Operating system

Ubuntu 20.04

### Browser

Irrelevant

kjarosh:
I ran bisect over this, which reports 4db87a9643e23e3f0ae91694d89a67c9b566074d as the culprit (which appears to be from #14535). Also it does not always happen, I guess for some fonts only (perhaps the default ones?).

PR #18353, opened 2024-10-22
core: Include a regular subsetted Noto Sans as the default font

### LINKED ISSUE #14761: minus sign does not appear
### Describe the bug

hi!
in a serie of questions for pupils, minus sign does never appear . Plus, divide or multiply are ok...
Here is a picture of it, and the swf zipped !
Thanks a lot
Franck

### Expected behavior

Show the sign minus, so that kids could choose right answer!

### Content Location

[exo307.zip](https://github.com/ruffle-rs/ruffle/files/13938003/exo307.zip)

### Affected platform

Browser's extension

### Operating system

Windows 11 

### Browser

Brave

### Additional information

_No response_

evilpie:
The idea is to include a regular TTF font file for font rendering, instead of first creating a SWF with the Noto Sans as an embedded font. I am using `pyftsubset` together with a list of unicode ranges matching specific Unicode blocks.
This should make it quite easy to extend the included font in the future. Sadly `ttf_parser` doesn't support any of the newer compressed formats like WOFF. That is why I decided to manually apply deflate compression, which reduces the size by almost 50%.

Fixes #14761


tests/tests/swfs/avm2/stage3d_texture/output.expected.png
evilpie, line 1: Not quite sure what happened here. I can't reproduce this when running the test in Firefox using the extension. (The colors seem wrong anyway)

image

tests/tests/swfs/visual/cache_as_bitmap/edittext_selection/output.02.expected.png
evilpie, line 1: Oh this looks bad.

evilpie: Not sure what is going on here. The test actually includes a `DefineFont3` for `Noto Sans`, but it seems like we might still be falling back to the device font. I actually get the same rendering on desktop (I don't have `Noto Sans` installed).

evilpie: Apparently a known issues: https://github.com/ruffle-rs/ruffle/issues/14713. Thanks @kjarosh

evilpie:
This should be ready to review. I don't see any text rendering change that seems problematic.

torokati44:
> Sadly ttf_parser doesn't support any of the newer compressed formats like WOFF.

Just referencing https://github.com/RazrFalcon/ttf-parser/issues/122, in case it's added in the future.

torokati44:
There are also crates like [allsorts](https://crates.io/crates/allsorts) and [webtype](https://crates.io/crates/webtype), have you considered relying on them (or anything else)? Not that there's anything wrong with simply deflating (it may be better even, as no new dependency brought in), I'm just curious.

evilpie:
We need `ttf_parser` anyway for TLF's `DefineFont4`, so adding another crate with a different API would just be additional work.

tests/tests/swfs/visual/cache_as_bitmap/edittext_selection/output.02.expected.png
kjarosh: This shouldn't be needed after https://github.com/ruffle-rs/ruffle/pull/18437.

tests/tests/swfs/avm2/stage3d_texture/output.expected.png
kjarosh: What do you mean? Note that this test clicks the "Green Only" button (see its `input.json`).

evilpie: Oh sorry. Somehow I thought it's obvious. But the = sign (e.g. in `rows=5`) now looks like just a dash (-).

kjarosh: I believe you would have to render it in 1:1 scale and quality 1 to observe that. If we up the quality to 4 the second dash would probably be shown, but it's interesting it's missing on quality 1 :thinking:

Dinnerbone (approved):
Looks good to me but I'd really like to consider using metric compatible fonts at some point.

tests/tests/swfs/visual/cache_as_bitmap/edittext_selection/output.02.expected.png
torokati44: Dropped this change.

evilpie:
> Looks good to me but I'd really like to consider using metric compatible fonts at some point.

Which is also a kind of fun point, because from what I can tell Flash on Linux never used the "standard" fonts from Windows either.

PR #18370, opened 2024-10-24
avm2: Improve `TextField.getLineMetrics()`

kjarosh:
This method was inaccurate in various ways, this PR makes it return correct values and throw for invalid arguments.


Dinnerbone (approved):
Neat!

PR #18390, opened 2024-10-27
avm2: Fix string.replace capture group 0 handling

aidanbabo:
`RegExp::effective_replacement` would panic given a capture group of 0. It now fails to recognize 0 as a valid capture group.

The code now also handles an edge case involving capture groups with a leading 0 having the second digit appended after the replacement text. This behavior is observed in the debug flash player and I'm not sure why it behaves this way. If I'm getting strange behavior please let me know!


Dinnerbone (approved):
Thank you!

Issue #18386, opened 2024-10-26
Ruffle crashes in Safari 15 and Pale Moon


n0samu:
### Describe the bug

Since nightly 2024-10-19, Ruffle is crashing shortly after starting any game/content in Safari 15. The error is:
```
RuntimeError: Out of bounds table access (evaluating 'b._dyn_core__ops__function__FnMut__A____Output___R_as_wasm_bindgen__closure__WasmClosure___describe__invoke__h281f2f93d3d6cf28(n,e,t)')
```
This happens on the Ruffle Web Demo and [my testing site](https://n0samu.github.io/ruffle-testing/). Ruffle freezes and no panic screen is shown. On some other sites/pages like Kongregate, I do get a panic screen and a "cannot recursively acquire mutex" error.

Additionally, Ruffle fails to load in Pale Moon with the following error:
```
Error name: CompileError
Error message: wasm validation error: at offset 28: bad type
```

### Expected behavior

Ruffle should work in these browsers.

### Content Location

https://n0samu.github.io/ruffle-testing/

### Affected platform

Self-hosted version

### Operating system

Windows 10, iOS 15.8.3

### Browser

Pale Moon 33.4.0.1, Safari 15

n0samu:
Our fallback WASM now adheres to the WebAssembly MVP - it does not use any extensions (see #18528). This allows us to support browsers like Pale Moon, Basilisk, and old versions of Safari.

PR #18397, opened 2024-10-27
ci: Freeze rustc for nightly web builds to 1.81.0

danielhjacobs:
***Note:** Tested locally, but not via CI.

Rust 1.82.0 enables the `reference-types` and `multivalue` target features for the WebAssembly target by default: https://releases.rs/docs/1.82.0/#compatibility-notes. wasm-bindgen 0.2.94+ enables reference type and multivalue transformations if the module already makes use of the corresponding target features.  The `reference-types` WebAssembly extension is not available in Safari 14.1, which we claim to support: https://caniuse.com/wasm-reference-types. It also, from testing, appears to not work in Pale Moon. Additionally, there seems to be some issues with support in Safari 15, which should support the reference-types extension but may have some bugs (which I don't expect will ever be fixed since Safari 15 isn't really supported anymore).

Hopefully, this can be reverted after a wasm-bindgen release with https://github.com/rustwasm/wasm-bindgen/pull/4213 is merged.

### DIFF
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -332,9 +332,9 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install Rust toolchain
-        uses: dtolnay/rust-toolchain@stable
+        uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: stable
+          toolchain: 1.81.0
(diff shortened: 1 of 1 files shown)


adrian17 (approved):

n0samu:
Should fix #18386

Issue #16877, opened 2024-06-25
text invisible regression since nightly 2024-03-19


zjk261:
### Describe the bug

![image](https://github.com/ruffle-rs/ruffle/assets/98996990/873a8863-f9e7-447f-a519-03330b095e7b)

### Expected behavior

text should be visible

### Content Location

https://n0samu.github.io/flash-player-check

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Chrome 126.0.6478.114

Lord-McSweeney:
When I bisect I get that nightly-2023-08-07 was the first commit where the text was invisible.

evilpie:
The text seems to invisible due to this early return:
https://github.com/ruffle-rs/ruffle/blob/dd776d9a8183f52ecdb3775d7e969201c65b0ecc/core/src/display_object/edit_text.rs#L946-L958

`origin.y()` is some positive number and `edit_text.bounds.y_max` is -505.

I think this was original added by @adrian17 in https://github.com/ruffle-rs/ruffle/commit/3c2d629e0d6b8b12b62634cf592ce0f9b48d0c3c. 
/cc @kjarosh

kjarosh:
There was a discussion about it on Discord recently. I believe @britt-j is working on a fix

britt-j:
yeah, I'll be working on getting a test made for this by tomorrow hopefully

kjarosh:
The text is now visible, but text selection is broken. That's the case for every non-left-aligned text, so nothing special in this SWF. This should be fixed after refactoring `screen_position_to_index` so that it uses proper bounds from the layout.

PR #18418, opened 2024-10-28
core: Fix text culling check

britt-j:
The check relied on `y_max` being the same as `height`, which isn't the case since `y_max` can be negative
(close https://github.com/ruffle-rs/ruffle/issues/16877)


tests/tests/swfs/visual/edittext/edittext_negative_bounds/test.toml
```diff
@@ -0,0 +1,8 @@
+num_ticks = 1
+
+[image_comparisons."output"]
+trigger = 1
+tolerance = 0
```
britt-j, line 5: Not sure what value this should have

adrian17 (approved):
LGTM, but wanna make sure kjarosh approves too :)

kjarosh:
Thank you for your contribution!

There are several pitfalls that we can encounter in text-based tests.

It's important to make sure the test produces the same output as Flash Player. Unfortunately that's not the case for `edittext_negative_bounds`. If you open the SWF in FP, you can see that the displayed text differs a lot.

This is specifically caused by the fact that device fonts are used to render the text. It's the case despite having a font set, because `useOutlines` is `false`. It has to be set to `true` in order for embedded fonts to work. When you set it to `true`, note that it will not work properly because currently, the embedded font has 0 ascent and descent set.

In general, I recommend using fonts that are vastly different from the default ones. I personally use my own crafted fonts and commit them (which you can also just embed to the SWF), see for instance: https://github.com/ruffle-rs/ruffle/blob/e9926f9a2e4f518a4095d9c1643440031033f14f/tests/tests/swfs/visual/fonts/leading_define_font/TestFont.ttf. I recommend doing that because of several reasons:
1. a test font does not have complex shapes,
2. it's easy to spot that a wrong font is used,
3. it's possible to modify properties of the test font and provide custom values.

Regarding the output, it's best if the output comes from the Flash Player, and not Ruffle. It's enough to screenshot FP and replace the `output.expected.png` with it. In that case, you'll need to set `tolerance` and `outliers`. I usually set `outliers = 0` and `tolerance = 128` (along with `quality = 4`), which is enough for black and white text-based tests. Setting `scaleMode` to `noscale` helps a lot when taking a screenshot.

kjarosh:
> Do you mean the `sample_count` in `with_renderer`?, because I can't find anything about `quality` in the test readme nor in the other tests in the `edittext` folder

Yes, I'm sorry, I meant `sample_count`.

> I'm sorry if I've misunderstood what image tests are for, but I thought they were moreso regression tests for Ruffle specifically rather than accuracy tests? I'm still going to take a screenshot from FP for this test, I'm just wondering

Well, I guess there's no universal definition. Some tests in fact do that due to the fact that it's often _extremely_ hard to get pixel-perfect output in Ruffle. In the case of simple text tests, I've put a lot of effort into making text near-pixel-perfect, so it should be fairly easy. Not to mention that we won't have to update the test when something changes in Ruffle, it should automatically pass (provided that something changes in the right direction ;) ).

> So it'd be alright if I used that font you linked, or should I make a different one for this?

Sure thing, you can simply embed this TTF in JPEXS and you're good to go. No need to create a custom font. Creating a custom font would be beneficial if you wanted to test something specific related to that font, but here we just want to check whether the text renders. Just remember that this font defines only 4 characters if I remember correctly (a, b, c, d).

If you need some help with something just let me know, I can polish up the test :)

britt-j:
@kjarosh don't know if the force pushes sent a notification, but if not, this should finally be ready now

kjarosh (approved):
Perfect! Thank you!

PR #18419, opened 2024-10-28
core: Handle the case where frame rate is set to 0

Toad06:
In the current builds, any SWF file with a frame rate set to 0 in its header fails to run. This actually makes some sense, but Flash disagrees.

This PR fixes this issue and makes the following SWF work: [`telepoche040110.swf`](https://www.lejusteprixtv.com/fichiers/presse/telepoche040110.swf), created with the third-party tool Print2Flash.

I also made a few changes to the `Stage.frameRate` property in AVM2 to align with Flash's behavior.


Toad06:
This also happens for `NaN`, so this case is covered as well.

Lord-McSweeney (approved):

kjarosh (approved):

kjarosh:
Thank you!

Toad06:
Thank you for the reviews. 😉

PR #18426, opened 2024-10-29
chore: Update translations

RuffleBuild:

### DIFF
--- a/desktop/packages/linux/rs.ruffle.Ruffle.desktop
+++ b/desktop/packages/linux/rs.ruffle.Ruffle.desktop
@@ -23,6 +23,7 @@ GenericName[ja]=Flashプレイヤー
 GenericName[ko]=플래시 플레이어
 GenericName[nl]=Flash-speler
 GenericName[pl]=Odtwarzacz Flash
+GenericName[pt_PT]=Leitor Flash
 GenericName[tr]=Flash Oynatıcı
 GenericName[zh_CN]=Flash 播放器
 GenericName[zh_TW]=Flash 播放器


torokati44:
Do you think this is ok, @kjarosh ?
image

kjarosh:
I see that now there are various words that are used for "player":

1. reprodutor (used in keywords & description)
2. jogador (used in description)
3. leitor (now used in generic name)

I have no idea which one is correct to use. It seems that "jogador" is used more for games (see https://en.wiktionary.org/wiki/jogador), "reprodutor" for movies & recordings (https://en.wiktionary.org/wiki/reproduzir), and "leitor" for digital readers (https://en.wiktionary.org/wiki/leitor).

I have zero knowledge about Portuguese, but if I had to guess I would go for "reprodutor" due to:

* https://en.wiktionary.org/wiki/player#Translations-software_application
* https://pt.wikipedia.org/wiki/Windows_Media_Player
* https://pt.wikipedia.org/wiki/Adobe_Flash_Player

Maybe something like "Reprodutor Flash"? No idea how to join these two words together.

100tpt:
> I see that now there are various words that are used for "player":
> 
> 1. reprodutor (used in keywords & description)
> 2. jogador (used in description)
> 3. leitor (now used in generic name)
> 
> I have no idea which one is correct to use. It seems that "jogador" is used more for games (see https://en.wiktionary.org/wiki/jogador), "reprodutor" for movies & recordings (https://en.wiktionary.org/wiki/reproduzir), and "leitor" for digital readers (https://en.wiktionary.org/wiki/leitor).
> 
> I have zero knowledge about Portuguese, but if I had to guess I would go for "reprodutor" due to:
> 
> * https://en.wiktionary.org/wiki/player#Translations-software_application
> * https://pt.wikipedia.org/wiki/Windows_Media_Player
> * https://pt.wikipedia.org/wiki/Adobe_Flash_Player
> 
> Maybe something like "Reprodutor Flash"? No idea how to join these two words together.

Hi! I've been asked to address your notes, as I’m one of the main Portuguese translators for Ruffle.

You should definitely discard the word jogador; it means "player" as in someone who plays a game.

As for reprodutor vs. leitor, the difference is more nuanced—both can describe a program that plays content. The literal translations are "reproducer" vs. "reader".

To break the tie, leitor is much more popular. It’s commonly used for CD/DVD players (leitor de CD/DVD) and, therefore, for digital audio/video players (leitor de áudio/vídeo).

While reproduzir [um áudio/vídeo] is commonly used for "to play [an audio/video]," the noun reprodutor isn’t as common, is longer, and can seem more technical.

TL;DR: use 'leitor' for all cases; 'jogador' for a player in a game.

> I'll fix it. I suggest becoming a coordinator, so I can keep it consistent.
> Also, in some cases, "Flash Player" has been kept as is. Should it be translated in all instances?

kjarosh:
Thank you @100tpt for the comprehensive explanation! In this case please feel free to fix other translations with the word "reprodutor" (I guess it's used in keywords only).

> Also, in some cases, "Flash Player" has been kept as is. Should it be translated in all instances?

I believe we would have noticed if "Flash Player" was badly translated. Usually "Flash Player" means _the_ Flash Player.

kjarosh (approved):

PR #18434, opened 2024-10-30
core: Exclude `_root` from button mode

Toad06:
I noticed in [`box_present.swf`](https://github.com/Toad06/CafeJeux-Archive/blob/main/swf/box_present.swf) that this file sets `_root.onPress = func`: in Flash Player, this does nothing other than set the property, but in Ruffle, it causes the clip to enter button mode.

This only affects button events; equivalent `onMouse*` events are supported, as shown in the included test.


Dinnerbone (approved):
Thanks!

Issue #14326, opened 2023-12-06
Sara's Cooking Class: Gingerbread House - isFontCompatible is not a function of class Class


MartySVK:
### Describe the bug

The menu is broken, probably because of an error:

```
ERROR core/src/avm2/events.rs:424 Error dispatching event EventObject(EventObject { type: "addedToStage", class: flash.events::Event, ptr: 0x8a10c20 }) to handler FunctionObject(FunctionObject { ptr: 0x8a04b28 }) : TypeError: Error #1006: isFontCompatible is not a function of class Class.
```
![image](https://github.com/ruffle-rs/ruffle/assets/10010569/ed11b056-5840-4df5-97cb-f1c7f8203fee)

This problem also has different games from the Sara's Cooking Class series

### Expected behavior

working menu

### Content Location

Game URL - https://www.gameflare.com/online-game/gingerbread-house/
SWF URL - [agame.zip](https://github.com/ruffle-rs/ruffle/files/13581540/agame.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

evilpie:
Not sure why all the text is shifted, but at least it's mostly usable.

![image](https://github.com/ruffle-rs/ruffle/assets/296445/0263bb49-b4e2-47b1-a99d-618863a41773)

MartySVK:
maybe something for @kjarosh to check 🥺

Lord-McSweeney:
Text is shifted because we don't implement TLF positioning correctly (all TLF text should be one line up).

PR #18435, opened 2024-10-31
debug_ui: Add Avm2Button tab

kjarosh:
This tab shows important information of the button, e.g. its states.

image

This helps in debugging https://github.com/ruffle-rs/ruffle/issues/14326.


Dinnerbone (approved):
Neat!

PR #18439, opened 2024-11-01
chore: Reformat AS3 code using asformat

evilpie:
This is the result of formatting our current AS3 code using the [Apache Royale Formatter](https://apache.github.io/royale-docs/formatter). To be honest, I'm not a big fan of the result, simply because I don't like it when braces are added to a new line. It also doesn't fit in with our current style.

If we do decide to land this, we could add a CI lint check that ensures all AS3 in pull requests is correctly formatted.


core/src/avm2/globals/flash/text/engine/GroupElement.as
```diff
@@ -1,85 +1,105 @@
...
+					this._elements.AS3::insertAt(beginIndex + i, newElements[i]);
+				}
+			}
+			return old;
+		}
+
+		public function splitTextElement(elementIndex:int, splitIndex:int):TextElement
+		{
+			var element = getElementAt(elementIndex);
+			if (!(elementinstanceof TextElement))
```
evilpie, line 75: And it apparently produces wrong code ...

evilpie: https://github.com/apache/royale-asjs/issues/1249

Fancy2209: This issues has been fixed by 
https://github.com/apache/royale-compiler/commit/975a04b9073b6f14ddc5ab4ddc609ceb724f9bab

PR #18446, opened 2024-11-02
text: Improve accuracy and performance of caret placement in text

kjarosh:
This PR fixes `screen_position_to_index` and makes some refactors:

* the gutter is now subtracted instead of added when translating local position to layout coordinates,
* text bounds are now taken into account in `screen_position_to_index` (it was broken when the bounds did not start at `(0,0)`),
* binary search is now used for finding the line in `screen_position_to_index`.

This makes `screen_position_to_index` a lot more accurate.

Four tests are added:

* `text_caret_placement_leading` – it checks how the caret is being placed in text depending on text leading,
* `text_caret_placement_align` – it checks how the caret is being placed in text depending on text align,
* `text_caret_placement_translated_bounds` – it checks how the caret is being placed when bounds do not start at `(0,0)`,
* `text_caret_placement_scroll` – it checks how the caret is being placed with non-zero horizontal and vertical scroll.


adrian17 (approved):
Just to be sure, this also corrects for scrolling? (I’m not familiar what the abstractions cover there)

kjarosh:
Yes, scrolling worked okay, nonetheless I've added a test `text_caret_placement_scroll` while I was at it.

PR #18489, opened 2024-11-05
debug_ui: Show text format on hover in EditText

kjarosh:
This patch adds the possibility to view the whole text format for a given text span, or the default one of the text field.

<image src="https://github.com/user-attachments/assets/493e9fc1-8561-4b0d-984d-be83d5fdda76" height="300"/>
<image src="https://github.com/user-attachments/assets/aa8f99ca-514d-42a8-a7ee-3284912e9df6" height="300"/>


torokati44 (approved):
Nice, thank you!

Issue #18030, opened 2024-09-21
Rogue Soul 2: Text alignment issues


waspennator:
### Describe the bug

Text on certain buttons (like the play button) and on shop item prices (agility upgrade) are misaligned

### Expected behavior

Text should be aligned

### Content Location

https://www.newgrounds.com/portal/view/646521?emulate=flash

### Affected platform

Browser's extension

### Operating system

Android 13

### Browser

Chrome 128

PR #18494, opened 2024-11-06
text: Fix text align for autosize fields

kjarosh:
Fixes https://github.com/ruffle-rs/ruffle/issues/18030, progresses https://github.com/ruffle-rs/ruffle/issues/18490, fixes https://github.com/ruffle-rs/ruffle/issues/10660, fixes https://github.com/ruffle-rs/ruffle/issues/1440, fixes https://github.com/ruffle-rs/ruffle/issues/3156.

Supersedes https://github.com/ruffle-rs/ruffle/pull/18092, supersedes https://github.com/ruffle-rs/ruffle/pull/12598.

Basically, when horizontal autosize is effective, we don't know the width of the text to be laid out and we have to calculate it dynamically.

Before this PR, the width was assumed to be the width of the field, which made non-left-aligned text display weirdly.

This PR lays out text two times when autosize is on: the first time to calculate the proper text width, and the second time to provide the final layout knowing the proper width.

Note that we could do it more efficiently by doing multi-pass layout, first pass – basic horizontal layout + vertical layout, second pass – full horizontal layout; but that requires rewriting most of the current layout code. We could do that in the future when we have more test coverage for text fields though (and we're absolutely sure that's how it works)!

Note 2: this PR moves some code around, so make sure to review commit-by-commit and don't be afraid of the number of lines changed.


Dinnerbone (approved):
Nice work!

tests/tests/swfs/avm2/edittext_autosize_align/test.toml
```diff
@@ -0,0 +1,7 @@
+num_ticks = 1
+
+[image_comparisons.output]
+tolerance = 128
```
Dinnerbone, line 4: To be sure, why such a high tolerance?

kjarosh: It might seem that the tolerance is pretty high, but for tests which check the placement of black rectangles over white background, it basically means a half-pixel tolerance. It's pretty common that we're off by half a pixel, e.g. in cases of border corners (different line rasterization algorithms), or glyph placement (font hinting probably?).

Dinnerbone: But is it _everywhere_ that's wrong, or just some places? Can we reduce the tolerance and increase the outliers? I'm a little worried by "every pixel can be half wrong" 😅

kjarosh: I could alternatively do low tolerance + max_outliers, but TBH after making so many text tests I got bored of counting these outliers and making sure they are at the right places.

kjarosh: The solution would be to provide max outliers with max tolerance for each one, so that we could say "there are max 144 outliers, but each one shouldn't be more off than 128".

PR #18515, opened 2024-11-07
web: Use tsx for save manager rows and use more accurate check for SOL

danielhjacobs:
This includes a few vaguely related changes:

1) Many functions that relate to the save manager and don't rely on any parts of the `InnerPlayer` have been pulled out into external functions, taking a queue from already existing external functions like `parseBoolean` and `parseAllowScriptAccess`. These functions include `saveFile`, `base64ToArray` (which has also been slightly simplified), `base64ToBlob`, and `isB64SOL`.

2) `isB64SOL` has been changed to more accurately check for SOL data, using logic from https://www.sans.org/blog/local-shared-objects-aka-flash-cookies/

3) Each row of the save table has been changed to a tsx component called SaveRow, instead of using so much direct DOM manipulation.

### DIFF
--- a/web/packages/core/src/internal/player/inner.tsx
+++ b/web/packages/core/src/internal/player/inner.tsx
@@ -2246,16 +2246,7 @@ function isSolData(data: string): boolean {
         if (data.charCodeAt(10 + i) !== markers[i]) return false;
     }
 
-    // Two bytes length of the object name
-    // First byte code * 256 + second byte code
-    const nameLength = data.charCodeAt(16) * 256 + data.charCodeAt(17);
-
-    // Three bytes of padding after the object name, always 0x00 0x00 0x00
-    if (data.slice(18 + nameLength, 18 + nameLength + 3) !== "\x00\x00\x00") {
-        return false;
-    }
-
-    // The structure makes this look like an SOL file up until the data
+    // This looks like an SOL file up until the end of the header
     return true;
 }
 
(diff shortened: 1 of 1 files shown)


web/packages/core/src/internal/player/inner.tsx
```diff
@@ -2245,6 +2177,80 @@ export function isFallbackElement(elem: Element): boolean {
...
+       data.charCodeAt(0) === 0x00 && data.charCodeAt(1) === 0xbf &&
+       // Next four bytes give the length of the LSO file excluding the first six bytes.
+       // First byte code * 256^3 + second byte code * 256^2 + third byte code * 256^1 + fourth byte code * 256^0 (1) + 6
+       [0, 1, 2, 3].reduce((acc, i) => acc + data.charCodeAt(i + 2) * (256 ** (3 - i)), 6) === data.length &&
+       // Next four bytes are another magic value (ASCII value of TCSO)
+       data.slice(6, 10) === "TCSO" &&
+       // Next six bytes are padding (0x00 0x04 0x00 0x00 0x00 0x00)
+       [0x00, 0x04, 0x00, 0x00, 0x00, 0x00].every((v, i) => data.charCodeAt(10 + i) === v)
+    );
+}
```
Dinnerbone, line 2252: I think just checking the first two bytes should be enough. to be honest.
I'm a little worried that bugs in our encoder would make a save file not appear (perhaps we write more than we say we do), and in such a situation I'd rather see (and debug) an invalid save file than just have it silently not show up (even though ruffle will still try to use it)

danielhjacobs: I think I'd rather check the first two and TSC0, just the first two seems like too little since this loops through the entire local storage so it doesn't seem that unlikely for there to be false positives on sites that use local storage extensively if just checking the first two.

Dinnerbone: That sounds good. I like the hardcoded magic values, but anything calculated feels like it could cause issues one day

Dinnerbone (approved):
Thank you!

PR #18523, opened 2024-11-08
text: Return text width only in `Font::measure`

kjarosh:
This patch simplifies code, as text height is not needed anymore after recent text refactors.


Lord-McSweeney (approved):

torokati44:
Thank you! 🚀

Issue #18412, opened 2024-10-28
Potentially using wasmv1-none


workingjubilee:
Issue opened to discuss the possibility of using the wasmv1-none target in 1.84, as raised by https://github.com/rust-lang/rust/pull/131487#issuecomment-2441608673

danielhjacobs:
To copy some things over from Discord:

The benefits:

- Support for Safari 14.1 continues (not really necessary, very much EOL)
- Support for Pale Moon continues: https://forum.palemoon.org/viewtopic.php?f=70&t=31740#p256587, Moonchild says "these extensions in Firefox were tied to a specific Rust codegen methodology (cranelift) i.e. using Rust to generate wasm machine code; and now as a result requiring clients to interpret this extended (not baseline) wasm code, and guess what Mozilla used for this? That's right, their Rust implementations in Firefox. Since we don't support Rust in our development I don't think we will ever be able to follow that particular trail of breadcrumbs as it fully relies on processing wasm with Rust on the client side. I may still be misunderstanding how all of these pieces fit together but it does look like these are extensions to wasm that aren't possible to implement as they are. On top the cranelift codegen has, apparently, not been continued in Firefox/mozilla-central (if I go by the scattered BZ bugs) and the state of the code interpreter at Mozilla's client side is unknown as a result. If this is simply a matter of pushing optional extensions because Rust wants everything to be Rust (typical) then I'd appreciate it if someone helps us either to make a shim..."
- Other non-mainstream browsers may be able to more easily support Ruffle as they won't need to support the ever-evolving feature-set of the wasm32-unknown-unknown target.
- @n0samu reports Safari 15 seems to have some bugs with `reference-types`. It theoretically is meant to support them but it seems in practice there are issues. Without this, we may need to drop support for Safari 15 as well, which is slightly more unfortunate as it's still supported by security updates.

Concerns from @adrian17, for which the answers seem to all be no:

> the questions relevant for us are

> - is std:: accessible (even if it only re-exports stuff from core::)
> - is std::sync:Mutex available
> - are functional parts (that don't touch the OS themselves) like std::path::Path or std::time::Duration available

The major issue:

- We would need changes to https://github.com/search?q=repo%3Aruffle-rs%2Fruffle+std%3A%3A+path%3Aweb%2F&type=code and https://github.com/search?q=repo%3Aruffle-rs%2Fruffle+std%3A%3A+path%3Acore%2F&type=code: That's a lot of files and may not be worth it.

[...]

PR #18528, opened 2024-11-08
web: Make npm run build:dual-wasm create an MVP vanilla WASM module and make the default WASM module be with extensions

danielhjacobs:
~~I'm not sure whether this itself will be controversial. By itself, this update changes nothing in the development or CI pipelines by default.~~

~~Instead, it adds an environment variable, `BUILD_WASM_MVP`, which when set will cause `npm run build:dual-wasm` and/or `npm run build:repro` to create a truly vanilla WASM module by using nightly Rust to disable all the WebAssembly extensions. With this set, the version of Ruffle that gets built can run on Pale Moon and Safari versions as old as 14.1.~~

~~What this doesn't do (yet) is use this environment variable on CI or in the Firefox Docker script. Doing that is a matter of debate. Once we unpin CI from Rust 1.81, if we don't use this environment variable on CI, Ruffle will stop working on those older browsers. We can use this on CI temporarily as long as we want to keep supporting those browsers.~~

Edit: New scope of this PR is to make `npm run build:dual-wasm` and `npm run build:repro` build the vanilla WASM modules as MVPs so they can support more browsers. It does this by building std, which it does stably by setting `RUSTC_BOOTSTRAP` to `1` to use Nightly features with stable Rust when building the vanilla module. Any CI scripts that use either of those commands (Nightly release, Dockerfile, dockerfile test) also now need to install the rust-src component. The environment variable used by those commands is now called BUILD_WASM_MVP=true instead of ENABLE_WASM_EXTENSIONS=true, to reflect the new purpose.

This also changes `npm run build` from building only the vanilla WASM module to building only the extensions WASM module.

### DIFF
--- a/.github/workflows/test_web.yml
+++ b/.github/workflows/test_web.yml
@@ -52,7 +52,6 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           targets: wasm32-unknown-unknown
-          components: rust-src
 
       - name: Cache Cargo output
         uses: Swatinem/rust-cache@v2
@@ -118,7 +117,6 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           targets: wasm32-unknown-unknown
-          components: rust-src
 
       - name: Cache Cargo output
         uses: Swatinem/rust-cache@v2
--- a/.github/workflows/test_web.yml
+++ b/.github/workflows/test_web.yml
@@ -52,6 +52,7 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           targets: wasm32-unknown-unknown
+          components: rust-src
 
       - name: Cache Cargo output
         uses: Swatinem/rust-cache@v2
@@ -117,6 +118,7 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           targets: wasm32-unknown-unknown
+          components: rust-src
 
       - name: Cache Cargo output
         uses: Swatinem/rust-cache@v2
--- a/web/README.md
+++ b/web/README.md
@@ -45,6 +45,7 @@ to update to the latest stable version of rust. You may run `rustup update` to d
 rust using the above instructions).
 
 For the compiler to be able to output WebAssembly, an additional target has to be added to it: `rustup target add wasm32-unknown-unknown`
+Because we use `RUSTC_BOOTSTRAP` in order to use certain Nightly flags with stable Rust you will also need to run `rustup component add rust-src`.
 
 #### Java
 
@@ -106,6 +107,7 @@ In this project, you may run the following commands to build all packages:
     -   There is `npm run build:dual-wasm` as well, to build a second WebAssembly module that makes use of some WebAssembly extensions,
         potentially resulting in better performance in browsers that support them, at the expense of longer build time.
     -   `npm run build:repro` enables reproducible builds. Note that this also requires a `version_seal.json`, which is not provided in the normal Git repository - only specially-marked reproducible source archives. Running this without a version seal will generate one based on the current state of your environment.
+    -   You will also need to run `rustup component add rust-src` with either of the previous two commands since we rebuild std for the vanilla WASM module.
 
 From here, you may follow the instructions to [use Ruffle on your website](packages/selfhosted/README.md),
 run a demo locally with `npm run demo`, or [install the extension in your browser](https://github.com/ruffle-rs/ruffle/wiki/Using-Ruffle#browser-extension).
--- a/web/README.md
+++ b/web/README.md
@@ -45,7 +45,6 @@ to update to the latest stable version of rust. You may run `rustup update` to d
 rust using the above instructions).
 
 For the compiler to be able to output WebAssembly, an additional target has to be added to it: `rustup target add wasm32-unknown-unknown`
-Because we use `RUSTC_BOOTSTRAP` in order to use certain Nightly flags with stable Rust you will also need to run `rustup component add rust-src`.
 
 #### Java
 
@@ -104,8 +103,8 @@ In this project, you may run the following commands to build all packages:
     -   This will build the wasm binary and every node package (notably selfhosted and extension).
     -   Output will be available in the `dist/` folder of each package (for example, `./packages/selfhosted/dist`).
     -   You may also use `npm run build:debug` to disable Webpack optimizations and activate the (extremely verbose) ActionScript debugging output.
-    -   There is `npm run build:dual-wasm` as well, to build a second WebAssembly module that makes use of some WebAssembly extensions,
-        potentially resulting in better performance in browsers that support them, at the expense of longer build time.
+    -   There is `npm run build:dual-wasm` as well, to build a second WebAssembly module that disables all supported WebAssembly extensions,
+        potentially resulting in support for more browsers, at the expense of longer build time.
     -   `npm run build:repro` enables reproducible builds. Note that this also requires a `version_seal.json`, which is not provided in the normal Git repository - only specially-marked reproducible source archives. Running this without a version seal will generate one based on the current state of your environment.
     -   You will also need to run `rustup component add rust-src` with either of the previous two commands since we rebuild std for the vanilla WASM module.
 
--- a/web/docker/Dockerfile
+++ b/web/docker/Dockerfile
@@ -11,7 +11,7 @@ RUN wget --progress=:giga https://github.com/WebAssembly/binaryen/releases/downl
     mv wasm-opt /usr/local/bin
 
 # Installing Rust using rustup:
-RUN wget 'https://sh.rustup.rs' --quiet -O- | sh -s -- -y --profile minimal --target wasm32-unknown-unknown
+RUN wget 'https://sh.rustup.rs' --quiet -O- | sh -s -- -y --profile minimal --target wasm32-unknown-unknown --component rust-src
 ENV PATH="/root/.cargo/bin:$PATH"
 # wasm-bindgen-cli version must match wasm-bindgen crate version.
 # Be sure to update in test_web.yml, release_nightly.yml, Cargo.toml, and web/README.md as well.
--- a/web/packages/core/tools/build_wasm.ts
+++ b/web/packages/core/tools/build_wasm.ts
@@ -42,12 +42,19 @@ function cargoBuild({
     profile,
     features,
     rustFlags,
+    extensions,
 }: {
     profile?: string;
     features?: string[];
     rustFlags?: string[];
+    extensions?: boolean;
 }) {
     let args = ["build", "--locked", "--target", "wasm32-unknown-unknown"];
+    if (!extensions) {
+        args.push("-Z");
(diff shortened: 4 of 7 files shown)


danielhjacobs:
We had wanted to wait for a wasm-bindgen update, but https://github.com/rustwasm/wasm-bindgen/pull/4213 was replaced by https://github.com/rustwasm/wasm-bindgen/pull/4235, which only disables reference type transformations for the bundler target (and updates CI to Rust v1.82). https://github.com/rustwasm/wasm-bindgen/issues/4227 was closed as not planned. On that issue, daxpedda  suggested `wasm32v1-none`. I pointed to https://github.com/ruffle-rs/ruffle/issues/18412. daxpedda said 

> Arc and Mutex should be replaced with Rc and RefCell unless you are using multi-threading. HashMap can be replaced with hashbrown and so on.
> If Ruffle intends to continue supporting older versions of Safari with newer versions of Rust and therefore LLVM, it should definitely consider supporting wasm32v1-none.

However, getting wgpu `no_std` compatible would require quite some work, and I don't know about other dependencies.

With that in mind, we're back to the issue of either using nightly Rust, or dropping support for Safari versions 15 or below and Pale Moon

daxpedda doesn't want to disable the transformations because:

> disabling those would only be a temporary solution, e.g. if multi-value gets properly enabled by LLVM it would be emitted by Rust and not wasm-bindgen. This will continue as more and more proposals will be enabled by default which older versions of Safari don't support.

torokati44:
> or dropping support for Safari versions 15 or below and Pale Moon

The former will only get less and less relevant over time, while the latter may (?) end up having this support added...?

danielhjacobs:
If we're dropping support for Safari 15 officially, there's an extra consideration we haven't talked about. All the current features are Safari 16+, right? Do we need dual-wasm then?

torokati44:
Interestingly, if we go the `nightly` route (which I'm not a fan of, BTW) the `-Z build-std=std,panic_abort` options would also be fairly beneficial for the build with extensions enabled, because a lot of naive `memcpy`-ing and `memset`-ing in `std` (in `Vec` for instance, while decoding videos) would be a lot faster using the `bulk-memory` extension.

danielhjacobs:
> while the latter may (?) end up having this support added

That's down to whether https://repo.palemoon.org/MoonchildProductions/UXP/issues/2647 gets fixed. Also worth considering, although I said "all the current features are Safari 16+", I don't know which other ones are unsupported by Pale Moon.

danielhjacobs:
> The former will only get less and less relevant over time

The general consensus is that this should at most be a temporary solution. MacOS Catalina and iPhone models 7 and below can't upgrade to Safari 16+. Do we have any ideas on market share there?

danielhjacobs:
https://www.coolest-gadgets.com/macos-statistics/

As of February 2024, MacOS Catalina holds the largest market around the world with 94.99% share. Other leading versions of MacOS are High Sierra (1.37%) and Sierra (0.93%).

danielhjacobs:
It's impossible to know market shares:

https://gs.statcounter.com/os-version-market-share/macos/desktop/worldwide
https://bugs.webkit.org/show_bug.cgi?id=216593

torokati44:
I guess linking this here couldn't hurt: https://blog.rust-lang.org/2024/09/24/webassembly-targets-change-in-default-target-features.html#disabling-on-by-default-webassembly-proposals

danielhjacobs:
Pale Moon logs this when I log individual extension support:
```
WebAssembly extension "Bulk Memory" is NOT supported.
WebAssembly extension "SIMD" is NOT supported.
WebAssembly extension "Saturated Float-to-Int" is NOT supported.
WebAssembly extension "Sign Extensions" is supported.  
WebAssembly extension "Reference Types" is NOT supported.
```

```diff
diff --git a/web/packages/core/src/load-ruffle.ts b/web/packages/core/src/load-ruffle.ts
index 6b474f61a..f0aadac41 100644
--- a/web/packages/core/src/load-ruffle.ts
+++ b/web/packages/core/src/load-ruffle.ts
@@ -35,15 +35,31 @@ async function fetchRuffle(
     setPolyfillsOnLoad();
 
     // NOTE: Keep this list in sync with $RUSTFLAGS in the CI build config!
-    const extensionsSupported: boolean = (
-        await Promise.all([
-            bulkMemory(),
-            simd(),
-            saturatedFloatToInt(),
-            signExtensions(),
-            referenceTypes(),
-        ])
-    ).every(Boolean);
+    const extensions = [
+        { name: "Bulk Memory", check: bulkMemory },
+        { name: "SIMD", check: simd },
+        { name: "Saturated Float-to-Int", check: saturatedFloatToInt },
+        { name: "Sign Extensions", check: signExtensions },
+        { name: "Reference Types", check: referenceTypes },
+    ];
+
+    const extensionsSupport = await Promise.all(
+        extensions.map(async (extension) => ({
+            name: extension.name,
+            supported: await extension.check(),
+        }))
+    );
+
+    let extensionsSupported = true;
+
+    extensionsSupport.forEach(({ name, supported }) => {
+        if (!supported) {
+            console.log(`WebAssembly extension "${name}" is NOT supported.`);
+            extensionsSupported = false;
+        } else {
+            console.log(`WebAssembly extension "${name}" is supported.`);
+        }
+    });
 
     if (!extensionsSupported) {
         console.log(
```

danielhjacobs:
Note that in testing, iOS 16.3 and below don't work with Rust 1.82+ without this workaround, so it's not just iOS 15-.

web/packages/core/tools/build_wasm.ts
```diff
@@ -42,12 +42,27 @@ function cargoBuild({
...
     rustFlags?: string[];
+    extensions?: boolean;
 }) {
-    let args = ["build", "--locked", "--target", "wasm32-unknown-unknown"];
+    let args =
+        !extensions && process.env["BUILD_WASM_MVP"]
+            ? [
+                  process.env["NIGHTLY_VERSION"]
+                      ? `+nightly-${process.env["NIGHTLY_VERSION"]}`
+                      : "+nightly",
```
danielhjacobs, line 57: The idea is the following commands should work in CI:

```
rustup install nightly-2024-12-02
rustup target add wasm32-unknown-unknown --toolchain nightly-2024-12-02
BUILD_WASM_MVP=1 NIGHTLY_VERSION=2024-12-02 npm run build:repro
```

danielhjacobs:
It appears that following the most recent wasm-bindgen update this PR is only necessary on browsers which do not have reference-types support. From what I know, of the browsers we currently support, then only eliminates support for Safari 14.1, Pale Moon, and Basilisk.

Pale Moon and Basilisk can currently support Flash anyway. On sites which use the polyfill they won't use Ruffle. Sites which use the Ruffle API or sites which use the polyfill if the user hasn't installed Flash will break on those browsers without this PR if we bump Rust back to stable. We can merge #18838 to add an appropriate error message.

Safari 14.1 on iOS has an ~1% market share among iOS users.

danielhjacobs:
This now updates CI as well and doesn't use Nightly, instead using `RUSTC_BOOTSTRAP` for the vanilla WASM module.

web/docker/Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN wget --progress=:giga https://github.com/WebAssembly/binaryen/releases/downl
     mv wasm-opt /usr/local/bin
 
 # Installing Rust using rustup:
-RUN wget 'https://sh.rustup.rs' --quiet -O- | sh -s -- -y --profile minimal --target wasm32-unknown-unknown
+RUN wget 'https://sh.rustup.rs' --quiet -O- | sh -s -- -y --profile minimal --target wasm32-unknown-unknown && rustup component add rust-src
```
torokati44, line 14: I think the installer script for rustup can take the `--component` flag immediately, no need to invoke it separately.

.github/workflows/test_web.yml
```diff
@@ -52,6 +52,7 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           targets: wasm32-unknown-unknown
+          components: rust-src
```
torokati44, line 55: Is this still needed then?

.github/workflows/test_web.yml
```diff
@@ -117,6 +118,7 @@ jobs:
...
```
torokati44, line 121: Same here.

web/README.md
```diff
@@ -103,9 +103,10 @@ In this project, you may run the following commands to build all packages:
     -   This will build the wasm binary and every node package (notably selfhosted and extension).
     -   Output will be available in the `dist/` folder of each package (for example, `./packages/selfhosted/dist`).
     -   You may also use `npm run build:debug` to disable Webpack optimizations and activate the (extremely verbose) ActionScript debugging output.
-    -   There is `npm run build:dual-wasm` as well, to build a second WebAssembly module that makes use of some WebAssembly extensions,
-        potentially resulting in better performance in browsers that support them, at the expense of longer build time.
+    -   There is `npm run build:dual-wasm` as well, to build a second WebAssembly module that makes use of only MVP WebAssembly extensions,
```
torokati44, line 106: Doesn't MVP mean "no extensions at all"?

web/README.md
```diff
@@ -103,9 +103,10 @@ In this project, you may run the following commands to build all packages:
...
+        potentially resulting in support for more browsers, at the expense of longer build time.
     -   `npm run build:repro` enables reproducible builds. Note that this also requires a `version_seal.json`, which is not provided in the normal Git repository - only specially-marked reproducible source archives. Running this without a version seal will generate one based on the current state of your environment.
+    -   Since we use `RUSTC_BOOTSTRAP` in order to use certain Nightly flags with stable Rust when building dual-wasm, you will also need to run `rustup component add rust-src` with either of the previous two commands.
```
torokati44, line 109: It might be worth mentioning that this is only needed for `build:dual-wasm` now.

web/packages/core/tools/build_wasm.ts
```diff
@@ -144,7 +155,7 @@ function detectWasmOpt() {
         return false;
     }
 }
-const buildExtensions = !!process.env["ENABLE_WASM_EXTENSIONS"];
+const buildWasmMVP = !!process.env["BUILD_WASM_MVP"];
```
torokati44, line 158: If WASM is Wasm, why isn't MVP Mvp?

web/README.md
danielhjacobs: I do say "with either of the previous two commands"

What would you suggest?

torokati44: Right, I didn't see enough context as for what the previous two commands actually are (I'm on mobile ATM), sorry.

Not nitpicking, and maybe I'm not right, but the direct reason for needing `rust-src` is that we do `build-std`. The `RUSTC_BOOTSRAP` envvar just happens to be necessary for it, since it's not stable yet.

adrian17 (approved):

torokati44 (approved):
Thank you! :tada:

torokati44:
Just spotted in the Rust 1.84 release notes: https://github.com/rust-lang/rust/pull/131080

PR #18581, opened 2024-11-12
video/external: Bump OpenH264 to 2.5.0

torokati44:
I suppose we can wait with this until FlatHub also updates `org.freedesktop.Platform.openh264`, so our Flatpak app doesn't regress.

Also note that this adds (untested) Windows on ARM64 support!

See for changelog: https://github.com/cisco/openh264/releases/tag/v2.5.0


kjarosh:
Filed an issue: https://gitlab.com/freedesktop-sdk/openh264-extension/-/issues/20

torokati44:
Also I think 2.5.0 yields frames later than 2.4.1, will have to investigate further and report upstream...

normalcea:
Hi, I am one of the maintainers of the ruffle package in [nixpkgs](https://github.com/NixOS/nixpkgs/), this version bump would make our package code much cleaner as nixpkgs master has openh264 2.5.0. 

Thank you.

torokati44:
I was thinking of creating a small standalone C program that decodes the frames in the h264 visual test FLV using 2.4.1 and 2.5.0, to use it to file a bug report to OpenH264 about the increased latency... But tiiimeee... :face_exhaling:

torokati44:
The raw config and frame data are already available in byte array literal form here: https://bugzilla.mozilla.org/show_bug.cgi?id=1896869

torokati44:
I finally got around to writing that test program and reporting the issue to Cisco: https://github.com/cisco/openh264/issues/3837

torokati44:
To be fair, maybe we could/should manually flush the decoder after each submitted NALU to get the decoded frames immediately (out of order), then do the reordering ourselves, based on the presentation time offset from the FLV...

tsunamistate:
Will probably need to update OpenH264 to 2.6.0, since 2.5.0 and lower versions have a vulnerability: https://github.com/cisco/openh264/security/advisories/GHSA-m99q-5j7x-7m9x

torokati44:
@kjarosh could the Flatpak app still do the regular "try to load the system lib, but inevitably fail, so just fetch the .so and load that" dance, even without that built-in framework that's being abandoned?

torokati44:
:skull: (I mean, it's understandable, but still...)
https://github.com/normalcea/nixpkgs/commit/9146c4c0089d003e43bc045a148728a1f05ccc93

kjarosh:
> @kjarosh could the Flatpak app still do the regular "try to load the system lib, but inevitably fail, so just fetch the .so and load that" dance, even without that built-in framework that's being abandoned?

Yup, it'll work, it's not a blocker.

But downloading binaries in runtime is a bad design—I know, not our choice—but we could think about using the same mechanism as the extension used.

Nonetheless, I thought that we're waiting for some OpenH264 fixes? I forgot what's the status

torokati44:
We're waiting for _backporting of_ said fixes. But if we give up on the "factory provided" OpenH264 extension for the Flatpak, and roll our own, we're good to go ahead.

torokati44:
Incidentally: I looked into VA-API, and the bindings themselves should be trivial, even though we might have to roll our own, like for OpenH264, to make the dynamic loading work (instead of requiring the .so to always be present to even launch the player, because it's linked to it), the issue is that VA-API is "stateless", so we would have to wire up a lot of the state tracking and buffer and frame order management that OpenH264 and WebCodecs conveniently provide for us... 🫠

torokati44:
So I think I'm just gonna go to 2.6.0, get rid of the FreeDesktop-provided Flatpak extension, rely on runtime downloading of it, and then maybe add back our own hand-rolled "extension" for Flatpak via "the same mechanism" as was before - sounds good?

kjarosh:
Sounds good, I don't think Flatpak is a blocker here, we can always rely on downloading OpenH264 at runtime (with a startup delay). Preferably we would migrate to some alternative to OpenH264 (as the default)

torokati44:
> Preferably we would migrate to some alternative to OpenH264 (as the default)

Yes, but what? 🫠 I'd love to see VA-API, but can't take it on myself anytime soon. FFmpeg (libavcodec) could work, I already had a working decoder with it before OpenH264. But the legal stuff keeps bothering me... I'll probably slowly let it go though...

kjarosh:
Let me know if you need some convincing! :eyes:

torokati44:
I feel like we'd just be running in circles.

PR #18598, opened 2024-11-14
chore: Update idna to 1.0, drop unicode-normalization

torokati44:
Now that all of `url`, `cookie_store`, and `publicsuffix` are out with releases depending on `idna` `1.0.x`, we can switch to that.

With this, the `idna_adapter` crate is introduced, which lets us globally control (in a hacky, but workable way), which IDNA backend we want, and allows dropping support for it entirely.

See: https://crates.io/crates/idna_adapter

It should probably be locked to `1.0.x` with some mechanism. (Adding it directly  to `Cargo.toml`, and/or configuring `cargo-deny`, since we already have it...)

Cc: @adrian17, since I remember you complaining a lot about the binary size bloat caused by these encoding tables... :sweat_smile:


evilpie:
We should definitely add something to (probably) `deny`, because this mechanism seems very brittle. I guess we could deny `unicode-normalization` as well now?

torokati44:
The size of the vanilla WASM module after wasm-bindgen and wasm-opt:

| `idna_adapter` | Bytes | Diff | Relative |
|--------|--------|-------- | --------|
| (base) | 12,915,705 | 0 | 100.0% |
| `1.0` | 12,707,466 | -208,239 | 98.4% |
| `1.1` | 12,964,026 | +48,321 | 100.4% |
| `1.2` | 12,821,717 | -93,988 | 99.3% |

deny.toml
```diff
@@ -63,6 +63,14 @@ deny = [
     # Having multiple versions of this can silently cause images (such as
     # the logo in the about dialog) in egui to not appear.
     { name = "image", deny-multiple-versions = true },
+    # Force using the 1.0.x series, with no actual IDNA support. We don't
+    # need it (Flash Player didn't support it), so it would just add bloat.
```
torokati44, line 67: This statement is yet to be verified still.

torokati44:
On top of adding bans in `deny.toml`, it would be great if somehow it could be added as a permanent exception to `cargo update`, so it won't try to update (and certainly not by duplicating it!) `idna_adapter`... Any ideas? :/

deny.toml
torokati44: IDNA domain names seem to work on (my) Linux, but not on Windows (10)... So I think it's fine it they don't work in Ruffle.

torokati44:
See this for testing:

```actionscript
package {
    import flash.display.Loader;
    import flash.display.Sprite;
    import flash.net.URLRequest;

        var nextY = 0;

        function loadFromDomain(domain: String) {
            var imageURL:String = domain + "/img/navi/top.gif";
            var loader:Loader = new Loader();
            var req:URLRequest = new URLRequest(imageURL);
            trace("Loading from " + req.url);

            loader.load(req);
            loader.y = this.nextY;
            this.nextY += 50;
            addChild(loader);
        }
        public function Test() {
            loadFromDomain("https://www.schoen.de");
            loadFromDomain("https://www.schön.de");
            loadFromDomain("https://www.xn--schn-7qa.de");

        }
	}
}
```

SWF here: [test.zip](https://github.com/user-attachments/files/17968751/test.zip)

On Linux, I see all 3 instances of the same image, on Windows, the middle one is missing (with the FP debugger showing an IO exception about it).

torokati44:
This will also be needed to resolve this advisory:
https://rustsec.org/advisories/RUSTSEC-2024-0421

deny.toml
evilpie: What would happen if a IDNA domain embedds rust? Does that break?

adrian17: I don't know how to test it, but my understanding was that we'd propagate the `http://xn--stuff` string all the way to the `fetch` api, which would handle it correctly?
Do I get this right?

adrian17:
> 98.4%

To be fair, -200kB was a way bigger deal when our wasm bundle was like 2MB :(

adrian17:
Anyway, as long as evilpie's concern is resolved, I think this is safe-ish?
Worst case, if something breaks, we can just trivially revert this.

adrian17 (approved):

torokati44:
> Worst case, if something breaks, we can just trivially revert this.

I don't think that would be necessary even, we can easily switch between the backends now via the `idna_adapter` version series.

deny.toml
torokati44: I also have no idea, nor about how to check. :/ I guess I could either buy an `xn--` domain, try to find one with embedded flash, or inject an embedding into a site on one...?
But yeah, my guess is the same: since FP on Windows desktop doesn't understand IDNA (as shown below), probably the browser plugin didn't either... And whatever transformations the browser did for it (if any), we could probably utilize as well.

torokati44: Oh... Guess what `/etc/hosts` is for... :man_facepalming: :sweat_smile:

evilpie: I test this using /etc/hosts and it seems to work. :tada:

evilpie (approved):

torokati44:
Thanks to both of you! 😊

torokati44:
@dependabot unignore url

Issue #18668, opened 2024-11-19
Regression in vertical text alignment


123jjck:
### Describe the bug

1. Open https://xarium.cc/playtest.php
2. Text is not displayed correctly
3. ![Image](https://github.com/user-attachments/assets/5cc14ee2-5dbf-4694-97c9-98cc585d707a)

### Expected behavior

Normal messagebox:
![Image](https://github.com/user-attachments/assets/a972e3d9-7f0e-4f32-b225-9fa4c635bdc7)
(you can test at https://xarium.cc/playtestold.php with [nightly-10-03](https://github.com/ruffle-rs/ruffle/releases/tag/nightly-2024-10-03))

### Content Location

https://xarium.cc/playtest.php

### Affected platform

Self-hosted version

### Operating system

Windows 11

### Browser

Google Chrome 131.0.6778.70 

### Additional information

Script that creates MessageBoxes: (method `CreateTextField`):
[MessageBoxConfig.txt](https://github.com/user-attachments/files/17812638/MessageBoxConfig.txt)

Probably due to https://github.com/ruffle-rs/ruffle/pull/18162, since versions before [nightly-2024-10-04](https://github.com/ruffle-rs/ruffle/releases/tag/nightly-2024-10-04) works fine

kjarosh:
No easy way of reproducing it on desktop. SWF requires a lot of files, the website somehow detects Ruffle Desktop and returns 403.

123jjck:
> website somehow detects Ruffle Desktop and returns 403

This is due to CloudFlare's JS challenge.
I modified the game client to display a messagebox every time it initializes (don't pay attention to the "undefined" button)
[serial_experiments_lain.zip](https://github.com/user-attachments/files/17822990/serial_experiments_lain.zip)

![Image](https://github.com/user-attachments/assets/37ad3497-2293-4b73-b893-b32504d1f86f)

kjarosh:
@123jjck thank you! That makes debugging a lot easier.

So I don't believe the main reason this renders wrong is text alignment issues (before https://github.com/ruffle-rs/ruffle/pull/18162 it must have looked better by accident). The text field is generated programmatically and the result is different in Flash and in Ruffle. Most notably, the font size is 20 instead of 19, and the height is 48.95 instead of 61.75.

123jjck:
> the font size is 20 instead of 19

Isn't the font size 19.5 according to the code? (maybe different rounding in Flash and in Ruffle?)
```as
[...]

PR #18701, opened 2024-11-21
text: Fix rounding in TextFormat

kjarosh:
Progresses https://github.com/ruffle-rs/ruffle/issues/18668.

Turns out `TextFormat` rounds values differently in setters and in the constructor + the behavior differs for SWF7 vs SWF8.


core/src/avm1/globals/text_format.rs
```diff
@@ -535,7 +551,10 @@ pub fn constructor<'gc>(
     this: Object<'gc>,
     args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
-    fn get_arg_as_i32<'gc>(activation: &mut Activation<'_, 'gc>, arg: Option<&Value<'gc>>) -> Result<Option<f64>, Error<'gc>> {
+    fn get_arg_as_i32<'gc>(
```
adrian17, line 554: so to be sure, the behavior only difference for setters, not the constructor?

adrian17 (approved):

core/src/avm1/globals/text_format.rs
kjarosh: That's right. Flash had a somewhat consistent behavior before SWF8, and then changed it, but only for setters.

Issue #7957, opened 2022-09-12
[Dot Action 1 & 2] The cursor display is out of range of the text.


Jmousy:
### Describe the bug

![image](https://user-images.githubusercontent.com/75052171/189578631-061b7c8c-3c8c-4fd2-b884-5bd3cd503976.png)

![image](https://user-images.githubusercontent.com/75052171/189578592-59fe36af-a00b-40e0-a65c-650f303b5b0f.png)

Recent Nightly is finally starting to show the text of Dot Action! 😍

In Dot Action 1 (with 2 also having the same problem), the text selection box is out of text as shown in the image above.

It seems to be because the text display size is different from the original file.

Confirmed on Ruffle Nightly 2022-09-12.

DotAction: https://cdn.flasharch.com/share/dot-action.swf
DotAction 2: https://cdn.flasharch.com/share/dot-action-2.swf

### Expected behavior

The text highlight should display correctly.

### Affected platform

Self-hosted version

### Operating system

Windows 10 21H2

### Browser

Google Chrome 105

Jmousy:
Thank you for resolving the issue!
I haven't checked the unmerged PRs above yet, but it appears to be fixed as of Nightly 2024-11-23.

Below are the text issues that remain as of that build (except for #18708). We will close this issue if it is resolved in Nightly in the future.

## Issue 1

![Image](https://github.com/user-attachments/assets/5b6ea53d-cca8-4d88-a926-c92314a141ba)

## Issue 2

![Image](https://github.com/user-attachments/assets/40b444b6-3b45-4881-a049-8cc4309fd9c7)

## Issue 3

![Image](https://github.com/user-attachments/assets/ef5c5f94-4e9c-4328-ae49-2935f441c245)

Issue #18651, opened 2024-11-18
Sugar's DressUp World - Cropped Text


TomFulp:
### Describe the bug

After clicking "Play", you will see a text prompt box where the text is getting cropped:

![Image](https://github.com/user-attachments/assets/8cb74a77-bc4f-4b64-a85a-4d7b87e0f08d)

### Expected behavior

The text box should look like this:

![Image](https://github.com/user-attachments/assets/d93fb5d2-93de-426d-9393-49748ae271f0)

### Content Location

https://www.newgrounds.com/portal/view/628422/format/flash?emulate=flash

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

Version 130.0.6723.119 (Official Build) (64-bit)

adrian17:
Looks like this also relies on `flash.text.engine.ElementFormat::getFontMetrics`.

Issue #3610, opened 2021-03-11
Problem entering in name text in game leon wars - self hosted version of Ruffle


freeworldgroup:
**Self hosted version of Ruffle - playing on google chrome - Windows 10 - Alienware laptop** 

Not sure if this is a bug, or simply a feature that has not been added yet - when trying to play the game 'leon wars', it is not possible to enter the players name at the beginning - making it impossible to play. 

http://www.freeworldgroup.com/games6/gameindex/theleonwars.htm

thanks
[theleonwars.zip](https://github.com/ruffle-rs/ruffle/files/6125640/theleonwars.zip)

PR #18708, opened 2024-11-22
text: Fix calculation of text size

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/18668
* Fixes https://github.com/ruffle-rs/ruffle/issues/7957
* Progresses https://github.com/ruffle-rs/ruffle/issues/18651
* Does NOT fix (yet) https://github.com/ruffle-rs/ruffle/issues/3610 (probably an edge case, I would leave it for a future PR)

This patch fixes how the text size (its width, height) is calculated. It makes sure that empty lines at the beginning, leading, and empty lines at the end are taken properly into account.

Additionally, this PR simplifies some code and renames `exterior_bounds` to `text_size`.

Fixing text size calculation results in autosize behaving properly.


core/src/html/layout.rs
```diff
@@ -104,13 +104,8 @@ pub struct LayoutContext<'a, 'gc> {
     /// should be used.
     bounds: Option<BoxBounds<Twips>>,
 
-    /// The exterior bounds of all laid-out text, including left and right
-    /// margins.
-    ///
-    /// None indicates that no bounds have yet to be calculated. If the layout
```
adrian17, line 110: Is the part about `None` not correct anymore? (what's the new meaning of `None`?)

core/src/html/layout.rs
```diff
@@ -469,16 +459,21 @@ impl<'a, 'gc> LayoutContext<'a, 'gc> {
...
 
-        self.flush_line(end);
+        if self.current_line_index == 0 {
+            // The very first line always gets the leading.
+            line_size_bounds += Size::from((Twips::ZERO, self.max_leading));
+        }
 
-        if let Some(eb) = &mut self.exterior_bounds {
-            *eb += line_bounds;
+        if self.boxes.first().unwrap().start() == end && last_line {
```
adrian17, line 470: is it possible that `self.boxes.is_empty()` here?

adrian17:
Also FYI,
> Does NOT fix

I think github only saw the word "fix" and will try to auto-close that issue anyway.

core/src/html/layout.rs
kjarosh: Removed too much :D should be good now

core/src/html/layout.rs
kjarosh: It's not possible, at the beginning of the method a box is added in case there are no boxes. I've moved this `unwrap` higher so it's more obvious.

adrian17: Oh, I see. I wasn't familiar with the abstractions and didn't connect that `append_text()` makes  `self.boxes` no longer empty.

adrian17 (approved):

PR #18709, opened 2024-11-22
chore: Rename `pt` locale for extension to `pt_BR`

kjarosh:
This should have been `pt_BR` all along, and the rename won't happen automagically, so here we go!


torokati44 (approved):
I believe you!

PR #18721, opened 2024-11-23
tests: Fix tests which were missing outputs/test.toml

kjarosh:
Some tests were missing `test.toml`, some were also missing their outputs.


adrian17 (approved):

Aaron1011:
Should we adjust ci to warn/error for test directories missing a test.toml?

kjarosh:
That's not trivial: first of all it's difficult to know which directories should be tests. Assuming we'll also check for test.swf, there are subdirectories of tests that have test.swf, and there are valid tests without test.toml (files used in Rust code directly). So we would have to reorganize existing tests a bit.

PR #18744, opened 2024-11-24
avm2: Add some more stubs

kjarosh:
New stubs:

* `DisplayObject.globalToLocal3D`
* `DisplayObject.local3DToGlobal`
* `InteractiveObject.requestSoftKeyboard`

### DIFF
--- a/core/src/avm2/globals/flash/display/DisplayObject.as
+++ b/core/src/avm2/globals/flash/display/DisplayObject.as
@@ -1,5 +1,7 @@
 package flash.display {
-    
+
+    import __ruffle__.stub_method;
+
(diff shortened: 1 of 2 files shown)


core/src/avm2/globals/flash/display/DisplayObject.as
```diff
@@ -127,6 +130,18 @@ package flash.display {
 
         public native function globalToLocal(point:Point):Point;
 
+        [API("662")]
```
torokati44, line 133: Not directly related, but just noticed that `blendShader` has the same `[Version("10")]` metadata in `playerglobal.swc` as these functions (which should indeed translate to `[API("662")]`), but it's missing for us.

PR #18746, opened 2024-11-24
avm2: Don't activate fast-path for non-public Multiname

Lord-McSweeney:
```
newarray 1
pushbyte 0
pushbyte 0
setproperty MultinameL([Namespace("notpublic")])
```

Now throws error #​1056, as expected.


core/src/avm2/activation.rs
```diff
@@ -1264,12 +1262,17 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
-                        if let Some(value) = object.get_index_property(name_int as usize) {
-                            let _ = self.pop_stack();
-                            let _ = self.pop_stack();
-                            self.push_stack(value);
-
-                            return Ok(());
+                        // Note that this path doesn't activate when the
+                        // public namespace isn't included in the multiname's
+                        // namespace set
+                        if multiname.contains_public_namespace() {
```
adrian17, line 1268: Can you document (in all places) that in FP the behavior switches whether it's run in JIT or interpreter mode? I don't want it to be missed accidentally.

Lord-McSweeney: I'll add this and the test in the latest PR since that also implements the correct behavior

PR #18750, opened 2024-11-25
audio: Move `AdpcmSubstreamDecoder` into the `adpcm` module

torokati44:
As suggested in: https://github.com/ruffle-rs/ruffle/pull/16526#discussion_r1855263374


kjarosh (approved):
Looks good!

PR #18754, opened 2024-11-25
core: Rename BitmapData->BitmapRawData, wrap it in new BitmapData

adrian17:
This does make `bitmap_data.rs` uglier (it now has three copies of `fn height()` and similar methods), but on the other hand:

- The new names are clearer
- Most two-phase construction and `GcCell` handling are abstracted away:

```rs
// before
let bitmap_data = BitmapData::new_with_pixels(...);
let bitmapdata_wrapper = BitmapDataWrapper::new(GcCell::new(mc, bitmap_data));
let bitmapdata_avm2 = BitmapDataObject::from_bitmap_data_internal(..., bitmapdata_wrapper);
// after
let bitmapdata = BitmapData::new_with_pixels(mc, ...);
let bitmapdata_avm2 = BitmapDataObject::from_bitmap_data_internal(..., bitmapdata);
```

- It's a nice middle step towards future BitmapData refactors (including possibly data sharing), possibly without having to touch all these places again.

### DIFF
--- a/core/src/bitmap/bitmap_data.rs
+++ b/core/src/bitmap/bitmap_data.rs
@@ -1,10 +1,14 @@
 use crate::avm2::{Object as Avm2Object, Value as Avm2Value};
+use crate::context::RenderContext;
 use crate::display_object::{DisplayObject, DisplayObjectWeak, TDisplayObject};
 use bitflags::bitflags;
-use gc_arena::{Collect, Mutation};
+use gc_arena::{Collect, GcCell, Mutation};
 use ruffle_render::backend::RenderBackend;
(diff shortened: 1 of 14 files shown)


Lord-McSweeney (approved):

core/src/bitmap/bitmap_data.rs
```diff
@@ -192,9 +196,151 @@ bitflags! {
     }
 }
 
+#[derive(Copy, Clone, Collect, Debug)]
+#[collect(no_drop)]
+pub struct BitmapData<'gc>(BitmapRawDataWrapper<'gc>);
```
Lord-McSweeney, line 224: Is there a reason to have both `BitmapData` and `BitmapRawDataWrapper`?

PR #18770, opened 2024-11-26
avm2: Fix EditText and Graphic linkage

Lord-McSweeney:


core/src/display_object/movie_clip.rs
```diff
@@ -4554,7 +4557,9 @@ impl<'gc, 'a> MovieClip<'gc> {
                             .library
                             .library_for_movie_mut(movie.clone());
 
-                        match library.character_by_id(id) {
+                        let character = library.character_by_id(id);
```
Lord-McSweeney, line 4560: This needs to be duplicated (here and in `lookup_class_for_character`) to satisfy the borrow checker

core/src/display_object/movie_clip.rs
```diff
@@ -4611,7 +4604,7 @@ impl<'gc, 'a> MovieClip<'gc> {
                         }
                     }
                     Err(e) => tracing::error!(
-                        "Got AVM2 error {e:?} when attempting to assign symbol class {name:?}",
```
adrian17, line 4614: Why not print the error anymore?

core/src/avm2.rs
```diff
@@ -579,6 +579,39 @@ impl<'gc> Avm2<'gc> {
...
+            .ok_or_else(|| {
+                make_error_1014(
+                    activation,
+                    Error1014Type::ReferenceError,
+                    name.to_qualified_name(activation.gc()),
+                )
+            })?;
+
+        // See first comment
+        if class_object.inner_class_definition() != class {
```
adrian17, line 613: To be sure, what does FP do here? Does it also attempt both and compares, or does it try just one of these?

core/src/display_object/movie_clip.rs
adrian17: I'm guessing `Character` should just become a `Copy`, if/when all its variants become `Gc` wrappers - that should solve it, right?

core/src/avm2.rs
```diff
@@ -609,6 +614,53 @@ impl<'gc> Avm2<'gc> {
...
+                ) {
+                    return Err(Error::AvmError(type_error(
+                        activation,
+                        &format!("Error #2022: Class {}$ must inherit from DisplayObject to link to a symbol.", name.to_qualified_name(activation.gc())),
+                        2022,
+                    )?));
+                }
+            }
+        } else if movie_clip.avm2_class().is_none() {
+            // If this ID doesn't correspond to any character, and the MC that
```
adrian17, line 646: What happens if multiple entries don't correspond to characters? (like 0, 12345, 23456)

core/src/avm2.rs
```diff
@@ -609,6 +614,53 @@ impl<'gc> Avm2<'gc> {
...
+            // we're processing doesn't have an AVM2 class set, then this
+            // ClassObject is going to be the class of the MC. Ensure it
+            // subclasses Sprite.
+            if !class
+                .has_class_in_chain(activation.avm2().classes().sprite.inner_class_definition())
```
adrian17, line 651: BTW, maybe these can be added to `ClassDefs` to make this slightly shorter?

adrian17:
Oh, and I think the first commit should move to the end, since you said it's technically broken without the refactor, right?

core/src/avm2.rs
Lord-McSweeney: This was already resolved in #17275 - FP silently ignores them

core/src/display_object/movie_clip.rs
Lord-McSweeney: We still print the error, but not the name of the symbolclass, since that should already appear in the stack trace

core/src/avm2.rs
Lord-McSweeney: I have no idea; I couldn't test with FFDEC and I don't want to try hex-editing a SWF to figure it out- it's very difficult to trigger anyway, and if we do find a SWF that tries it, we'll get an error report

Lord-McSweeney: I'll add the proper behavior when the ClassObject wasn't initialized properly (set to `null`) to avoid a panic in that case, since it's easy to cause

torokati44:
_(Could someone add a `T-*` and an `A-*` label please? 😶)_

adrian17 (approved):

PR #18782, opened 2024-11-27
chore: Update tracy and allow bots to update it from hereon

Dinnerbone:
It was pinned before because conflicts would sometimes introduce multiple tracy clients, but we have cargo-deny for that now...


Cargo.lock
```diff
@@ -2744,7 +2744,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4979f22fdb869068da03c9f7528f8297c6fd2606bc3a4affe42e6a823fdb8da4"
 dependencies = [
  "cfg-if",
- "windows-targets 0.52.6",
+ "windows-targets 0.48.5",
```
torokati44, line 2747: 😑

Dinnerbone: 😭

adrian17 (approved):

PR #18810, opened 2024-11-30
avm2: Implement Matrix3D with 2D support only

### LINKED ISSUE #8033: Chirumiru: Display Bug
### Describe the bug

image
image
image
image
image

```
[2022-09-21T02:55:36Z ERROR ruffle_core::avm2::activation] AVM2 error: Could not find script for class trait Multiname { ns: Single(Package("flash.geom")), name: Some("Matrix3D"), params: [], flags: (empty) }
[2022-09-21T02:55:36Z ERROR ruffle_core::avm2::activation] AVM2 error: Could not find script for class trait Multiname { ns: Single(Package("flash.geom")), name: Some("Matrix3D"), params: [], flags: (empty) }
[2022-09-21T02:55:36Z ERROR ruffle_core::avm2::activation] AVM2 error: Could not find script for class trait Multiname { ns: Single(Package("flash.geom")), name: Some("Matrix3D"), params: [], flags: (empty) }
[2022-09-21T02:55:36Z ERROR ruffle_core::avm2::activation] AVM2 error: Could not find script for class trait Multiname { ns: Single(Package("flash.geom")), name: Some("Matrix3D"), params: [], flags: (empty) }
[2022-09-21T02:55:36Z ERROR ruffle_core::display_object::movie_clip] Got Could not find script for class trait Multiname { ns: Single(Package("flash.geom")), name: Some("Matrix3D"), params: [], flags: (empty) } when constructing AVM2 side of movie clip of type nineball2_fla::MainTimeline
[2022-09-21T02:55:36Z ERROR ruffle_core::avm2::activation] AVM2 error: Cannot get undefined property "perspectiveProjection" on class Some(ClassObject { name: flash.geom::Transform, ptr: 0x20745267920 })
[2022-09-21T02:55:36Z ERROR ruffle_core::avm2::events] Error dispatching event EventObject(EventObject { type: "addedToStage", class: flash.events::Event, ptr: 0x207453ea2a0 }) to handler FunctionObject(FunctionObject(GcCell(Gc { ptr: 0x20745325fd8 }))) : "Cannot get undefined property \"perspectiveProjection\" on class Some(ClassObject { name: flash.geom::Transform, ptr: 0x20745267920 })"
[2022-09-21T02:56:47Z ERROR ruffle_core::avm2::activation] AVM2 error: Could not find script for class trait Multiname { ns: Single(Package("flash.geom")), name: Some("Matrix3D"), params: [], flags: (empty) }
[2022-09-21T02:56:47Z ERROR ruffle_core::avm2::activation] AVM2 error: Could not find script for class trait Multiname { ns: Single(Package("flash.geom")), name: Some("Matrix3D"), params: [], flags: (empty) }
[2022-09-21T02:56:47Z ERROR ruffle_core::avm2::activation] AVM2 error: Could not find script for class trait Multiname { ns: Single(Package("flash.geom")), name: Some("Matrix3D"), params: [], flags: (empty) }
[2022-09-21T02:56:47Z ERROR ruffle_core::avm2::activation] AVM2 error: Could not find script for class trait Multiname { ns: Single(Package("flash.geom")), name: Some("Matrix3D"), params: [], flags: (empty) }
[2022-09-21T02:56:47Z ERROR ruffle_core::display_object::movie_clip] Got Could not find script for class trait Multiname { ns: Single(Package("flash.geom")), name: Some("Matrix3D"), params: [], flags: (empty) } when constructing AVM2 side of movie clip of type nineball2_fla::A_85
```
On Nightly 2022-09-22
```
[2022-09-22T05:18:46Z ERROR ruffle_core::display_object::movie_clip] Got RustError("Could not find script for class trait Multiname { ns: Single(Package(\"flash.geom\")), name: Some(\"Matrix3D\"), params: [], flags: (empty) }") when constructing AVM2 side of movie clip of type nineball2_fla::MainTimeline
[2022-09-22T05:18:46Z ERROR ruffle_core::avm2::activation] AVM2 error: Object(ErrorObject(ErrorObject(GcCell(Gc { ptr: 0x28450ea5498 }))))
[2022-09-22T05:18:46Z ERROR ruffle_core::avm2::events] Error dispatching event EventObject(EventObject { type: "addedToStage", class: flash.events::Event, ptr: 0x28450e38e40 }) to handler FunctionObject(FunctionObject(GcCell(Gc { ptr: 0x28450a8f088 }))) : AvmError(Object(ErrorObject(ErrorObject(GcCell(Gc { ptr: 0x28450ea5498 })))))
[2022-09-22T05:19:59Z ERROR ruffle_core::display_object::movie_clip] Got RustError("Could not find script for class trait Multiname { ns: Single(Package(\"flash.geom\")), name: Some(\"Matrix3D\"), params: [], flags: (empty) }") when constructing AVM2 side of movie clip of type nineball2_fla::A_85
``` 

### Expected behavior

image
image
image
image
image

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

SWF:
https://web.archive.org/web/20100707002957/http://www.l-room.net/product/movie/swf/chirumiru.swf

cookie-s:
Partially resolves #8033 .

## Screenshots

The comparison of screenshots for the first two frames mentioned in #8033 . (Please ignore background differences because those are not at the exact same frames.)

| Ruffle website demo (2024-11-30)   | This PR | Adobe Flash Player 32 |
| -------- | ------- | ------- |
|  |  | |

image
image

## Description

This PR replace stubs for `flash.geom.Transform.matrix3D` getter/setter with an actual implementation of Matrix3D with limited support.

This implementation is just a proxy to the existing 2D matrix implementation. Therefore transformations beyond 2D transformation works differently from the expected result.


core/src/avm2/globals/flash/geom/transform.rs
```diff
@@ -173,6 +199,41 @@ pub fn color_transform_to_object<'gc>(
...
+    Ok(object.into())
+}
+
+pub fn object_to_matrix3d<'gc>(
+    object: Object<'gc>,
+    activation: &mut Activation<'_, 'gc>,
+) -> Result<Matrix3D, Error<'gc>> {
+    let raw_data = object
+        .get_public_property("rawData", activation)?
+        .coerce_to_object(activation)?;
```
Lord-McSweeney, line 221: This should use `.as_object().expect("rawData cannot be null");` instead of `coerce_to_object`

cookie-s: Hi thank you. I basically followed the way of `object_to_matrix` and it seems to use `coerce_*` functions.

But actually, in the line 223-224, I looked for any `() -> Result<_,_>` function but I couldn't find a good one. Do you know any good way?

core/src/avm2/globals/flash/geom/transform.rs
```diff
@@ -45,6 +45,32 @@ pub fn set_color_transform<'gc>(
...
 }
 
+pub fn get_matrix_3d<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Object<'gc>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    // FIXME: This Matrix3D is generated from the 2D Matrix.
+    // It does not work when the 2D matrix contains any transformation in 3D.
+    // Support native Matrix3D.
```
kjarosh, line 55: We should preserve the stub warnings using `avm2_stub_getter!`, `avm2_stub_setter!`, as the 3D matrix is not yet fully supported. In order to do that, add the following under this comment (+ the other one in the setter).

```rust
avm2_stub_getter!(activation, "flash.geom.Transform", "matrix3D");
```

cookie-s: Thank you, true. I didn't know that we can produce the warning from Rust, too! Fixed: [62d6f79](https://github.com/ruffle-rs/ruffle/pull/18810/commits/62d6f794ee02b0f22a9e5d41d5d1917e9bb77c8c)

core/src/avm2/globals/flash/geom/transform.rs
Lord-McSweeney: `coerce_to_object` is the one function we're trying to phase out, `as_object` should always be used instead (returns either `Some(object)` for an object or `None` for null, except `rawData` can never be null). What do you mean in your second point? ` -> Result<Matrix3D, Error<'gc>>` looks fine to me.

core/src/prelude.rs
```diff
@@ -3,6 +3,7 @@ pub use crate::display_object::{
     DisplayObject, DisplayObjectContainer, HitTestOptions, TDisplayObject, TDisplayObjectContainer,
 };
 pub use ruffle_render::matrix::Matrix;
+pub use ruffle_render::matrix3d::Matrix3D;
```
Lord-McSweeney, line 6: We don't actually use this as a prelude, so you don't need to add this line here

cookie-s: True, thank you. Updated: [31c6e11](https://github.com/ruffle-rs/ruffle/pull/18810/commits/31c6e11e08c5ccd907bff829bb6fc5aaf18251a9)

core/src/avm2/globals/flash/geom/transform.rs
cookie-s: > coerce_to_object is the one function we're trying to phase out

Oh, I see, that makes sense. Fixed: [01c9101](https://github.com/ruffle-rs/ruffle/pull/18810/commits/01c9101479e676070dfba67ec387cc59403d4019)

> What do you mean in your second point?

Ah, sorry. I thought I needed to avoid `.expect()`, but it's probably not the case. Please forget it.

core/src/avm2/globals/flash/geom/transform.rs
```diff
@@ -173,6 +205,42 @@ pub fn color_transform_to_object<'gc>(
     Ok(object.into())
 }
 
+pub fn matrix3d_to_object<'gc>(
```
kjarosh, line 208: Just a nitpick: I think `matrix3d_to_object` and `object_to_matrix3d` don't have to be public

kjarosh (approved):
LGTM, thank you and welcome!

kjarosh:
This PR unfortunately introduces a regression by breaking the getter of `matrix3D`, because an array of 16 arguments is passed to the constructor instead of a vector with 16 elements.

The following patch fixes that but the getter still always returns an object instead of returning `null` by default.

```diff
diff --git a/core/src/avm2/globals/flash/geom/transform.rs b/core/src/avm2/globals/flash/geom/transform.rs
index 104a34a09..759238484 100644
--- a/core/src/avm2/globals/flash/geom/transform.rs
+++ b/core/src/avm2/globals/flash/geom/transform.rs
@@ -1,5 +1,7 @@
 use crate::avm2::globals::slots::*;
+use crate::avm2::object::VectorObject;
 use crate::avm2::parameters::ParametersExt;
+use crate::avm2::vector::VectorStorage;
 use crate::avm2::{Activation, Error, Object, TObject, Value};
 use crate::display_object::TDisplayObject;
 use crate::prelude::{DisplayObject, Matrix, Twips};
@@ -209,12 +211,17 @@ fn matrix3d_to_object<'gc>(
     matrix: Matrix3D,
     activation: &mut Activation<'_, 'gc>,
 ) -> Result<Value<'gc>, Error<'gc>> {
-    let args = matrix.raw_data.map(Into::into);
+    let value_type = activation.avm2().class_defs().number;
+    let mut raw_data_storage = VectorStorage::new(16, true, Some(value_type), activation);
+    for (i, data) in matrix.raw_data.iter().enumerate() {
+        raw_data_storage.set(i, Value::Number(*data), activation)?;
+    }
+    let vector = VectorObject::from_vector(raw_data_storage, activation)?.into();
     let object = activation
         .avm2()
         .classes()
         .matrix3d
-        .construct(activation, &args)?;
+        .construct(activation, &[vector])?;
     Ok(object.into())
 }
 
```

kjarosh:
I'm sorry @cookie-s! We have to consider reverting it :(

* https://github.com/ruffle-rs/ruffle/pull/18869

To be merged again, it will need a test + the patch posted by me above + 2D/3D switch to cover the following property (source: https://docs.ruffle.rs/en_US/FlashPlatform/reference/actionscript/3/flash/geom/Transform.html#matrix3D).

> If the matrix property is set to a value (not null), the matrix3D property is null. And if the matrix3D property is set to a value (not null), the matrix property is null.

adrian17:
> but the getter still always returns an object instead of returning null by default

More specifically: after some research, turns out that `transform.matrix3D` behaves exactly opposite to `transform.matrix` - while the latter can be implemented with just getter/setter returning a new object, the former is a real object with a similar DisplayObject binding as `Transform` itself - `transform.matrix3D` always returns the same object, and its values are bound to the DisplayObject, as in:
```as
var mat = movieclip.transform.matrix3D;
movieclip.x = 50;
trace(mat.rawData); // shows 50
```

cookie-s:
@kjarosh Thank you for letting me know and providing the patch!
I have made a new PR to address your [points](https://github.com/ruffle-rs/ruffle/pull/18810#issuecomment-2521637205): https://github.com/ruffle-rs/ruffle/pull/18888

@adrian17 Thank you, that's very interesting observation. After your comment, I checked the reference closely and found
https://airsdk.dev/reference/actionscript/3.0/flash/display/DisplayObject.html#transform
>Each of the transform object's properties is itself an object. This concept is important because the only way to set new values for the matrix or colorTransform objects is to create a new object and copy that object into the transform.matrix or transform.colorTransform property.
>
>For example, to increase the tx value of a display object's matrix, you must make a copy of the entire matrix object, then copy the new object into the matrix property of the transform object:
> ```
>    var myMatrix:Matrix = myDisplayObject.transform.matrix;  
>    myMatrix.tx += 10; 
>    myDisplayObject.transform.matrix = myMatrix;  
> ```
> You cannot directly set the tx property. The following code has no effect on myDisplayObject:
> ```
>    myDisplayObject.transform.matrix.tx += 10;
> ```

It mentions  `the transform.matrix or transform.colorTransform` but `transform.matrix3D` is probably not included in this sentence?
Sorry I couldn't support `movieclip.transform.matrix3D.appendTranslation(5, 5, 0);` in https://github.com/ruffle-rs/ruffle/pull/18888 but this information was helpful to understand a possible underlying design.

Issue #18811, opened 2024-11-30
Error on https://chasovnya.msk.ru/test.php


Kicksback:
# Error Info
Error name: Error
Error message: panicked at core/src/html/layout.rs:399:48:
called `Option::unwrap()` on a `None` value
Error stack:
```
Error: panicked at core/src/html/layout.rs:399:48:
called `Option::unwrap()` on a `None` value
    at n.wbg.__wbg_new_70a2f23d1565c04c (https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.1bb0d7819817410a0695.js:1:85360)
    at ruffle_web.wasm.js_sys::Error::new::hc7430be370242701 (wasm://wasm/ruffle_web.wasm-03229c4a:wasm-function[12865]:0x98663a)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once{{vtable.shim}}::h1a8fe10950133898 (wasm://wasm/ruffle_web.wasm-03229c4a:wasm-function[9790]:0x9041a4)
    at ruffle_web.wasm.std::panicking::rust_panic_with_hook::h33fe77d38d305ca3 (wasm://wasm/ruffle_web.wasm-03229c4a:wasm-function[6416]:0x82a155)
    at ruffle_web.wasm.core::panicking::panic_fmt::hde8b7aa66e2831e1 (wasm://wasm/ruffle_web.wasm-03229c4a:wasm-function[9502]:0x8fd257)
    at ruffle_web.wasm.core::panicking::panic::hcaca2598a27ec0fc (wasm://wasm/ruffle_web.wasm-03229c4a:wasm-function[9972]:0x915ae8)
    at ruffle_web.wasm.core::option::unwrap_failed::h9aa82eb71128b127 (wasm://wasm/ruffle_web.wasm-03229c4a:wasm-function[12351]:0x96344d)
    at ruffle_web.wasm.ruffle_core::html::layout::LayoutContext::fixup_line::hb75da25946570994 (wasm://wasm/ruffle_web.wasm-03229c4a:wasm-function[890]:0x192f58)
    at ruffle_web.wasm.ruffle_core::html::layout::lower_from_text_spans_known_width::h3bf6d46fd7b8ee5d (wasm://wasm/ruffle_web.wasm-03229c4a:wasm-function[982]:0x25eb60)
[...]

PR #18814, opened 2024-11-30
text: Always resolve a font, fall back to an empty font

kjarosh:
This PR ensures that we always resolve _some_ font, which is required for laying out text. When no font could be resolved, an empty font is returned that has no glyphs and has metrics with all values equal to 0.

* Fixes https://github.com/ruffle-rs/ruffle/issues/18811
* Fixes https://github.com/ruffle-rs/ruffle/issues/18816 (probably, no way of testing that)


core/src/html/layout.rs
```diff
@@ -655,7 +676,7 @@ impl<'a, 'gc> LayoutContext<'a, 'gc> {
             "_明朝" => Some(DefaultFont::JapaneseMincho),
             _ => None,
         } {
-            return context
+            if let Some(&font) = context
```
evilpie, line 679: Did this happen? I am not sure we really want to even support a missing default font. (I guess this could happen on custom Linux installs, without Noto Sans)

evilpie: Ah. I see, on the web with a custom config.

Maybe upgrade this to error, and immediately return an empty font?

kjarosh: This has also happened on a Windows desktop: https://github.com/ruffle-rs/ruffle/issues/18816 :/

We also have to "support" missing fonts at least by not panicking :smile: We also shouldn't skip any text or text layout when a font is missing and should assume metrics of value 0 (FP does that). So making `resolve_font` always resolve to something seems very appropriate.

I was wondering whether to error out here or fall back, but I thought it would be best to do everything to display _something_, especially that (according to my knowledge) we don't have a lot of tests related to device font resolution and don't match FP 100% in this regard.

But I personally don't have that much experience with fonts here, so if you think it's best to error out I'll happily change that, just let me know.

evilpie: I think it would make more sense to:
- In `default_font` try harder to find any available font? i.e. `device_fonts.all()`
- Otherwise return the empty font and log an error (So we don't keep creating empty fonts / logging errors)
- Revert the changes to this if branch

evilpie (approved):
Sorry, I was on a work trip. Looks good to me. I think I might try to do some simplification of `default_font`.

kjarosh:
No worries, thanks for the review!

PR #18829, opened 2024-12-02
avm2: Group generated AS3 field slot constants by package

moulins:
A follow-up to #18747, as proposed on Discord.

The generated code now looks like this:

```rs
mod slots {
  mod flash_utils_timer {
    pub const _timerId: u32 = ...;
    pub const _delay: u32 = ...;
  }
  // ...
 }
```

This makes usage terser, as most files only need access to the slots of a single AS3 class, which can then be imported as `use avm2::globals::slots::the_class as slots`.

### DIFF
--- a/core/build_playerglobal/src/lib.rs
+++ b/core/build_playerglobal/src/lib.rs
@@ -6,6 +6,7 @@ use proc_macro2::TokenStream;
 use quote::quote;
 use regex::RegexBuilder;
 use std::borrow::Cow;
+use std::collections::HashMap;
 use std::ffi::OsStr;
 use std::fs;
 use std::fs::File;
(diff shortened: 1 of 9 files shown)


Lord-McSweeney:
Can we keep Rust-style names for the constants, like `TIMER_ID` and `DELAY`?

adrian17 (commented):
LGTM, and I'm neutral on the constant name, I'm okay with it going either way.
(not sending approve until this is decided either way)

torokati44:
_(Could someone add a `T-*` and an `A-*` label please? 😶)_

moulins:
I personally prefer the "Flash-style" names, but I don't care that much so I just changed it to `UPPER_CASE`.

Lord-McSweeney (approved):

Lord-McSweeney:
I forgot this was going to conflict with #18757- @moulins could you change the EventDispatcher code to use the grouping?

torokati44:
Is that the right PR? 🤔

moulins:
> I forgot this was going to conflict with #18770- @moulins could you change the EventDispatcher code to use the grouping?

Done!

PR #18851, opened 2024-12-03
text: Fix `getCharBoundaries` for justified text

kjarosh:
`getCharBoundaries()` should report stretched space width, not the original space width.


core/src/html/layout.rs
```diff
@@ -1128,9 +1128,20 @@ impl<'gc> LayoutLine<'gc> {
...
         let layout_box = self.boxes.get(box_index)?;
-        let origin_x = layout_box.bounds().origin().x();
-        let (start, end) = layout_box.char_x_bounds(position)?;
-        Some((origin_x + start, origin_x + end))
+        let (start, mut end) = layout_box.char_x_bounds(position)?;
+
+        // If we're getting bounds of a space in justified text,
+        // we have to take into account that the character is stretched,
+        // and its bounds will end where the next character starts.
+        if layout_box.end() == position + 1 {
```
adrian17, line 1136: Just to be sure, can you clarify how `if layout_box.end() == position + 1 {` indicates that it's a space? I'm not nitpicking and this isn't blocking, I'm still trying to better wrap my head around the abstractions.

adrian17 (approved):

core/src/html/layout.rs
kjarosh: Sure thing!

It doesn't check for a space at all. It just uses the next character's start position as this character's end position when the character is at the end of a box. For non-spaces or non-justified text it just won't change the value.

Now thinking about this, the logic might be different when we take letter spacing or kerning into account, but there are no tests for that yet so we'll see. I'm just trying not to touch every possible aspect I can think of at the same time though :sweat_smile: 

Updated the comment

Dinnerbone (approved):

Issue #14740, opened 2024-01-14
Tether Together - wrong font


zjk261:
### Describe the bug

![1](https://github.com/ruffle-rs/ruffle/assets/98996990/bd3b70bf-da9d-458f-b3c5-7bae37fa89fb)

### Expected behavior

![2](https://github.com/ruffle-rs/ruffle/assets/98996990/2e653449-614a-48e2-9359-6429d823e2ad)

### Content Location

[TetherTogether.zip](https://github.com/ruffle-rs/ruffle/files/13929990/TetherTogether.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

PR #18856, opened 2024-12-04
text: Do not show text that's missing an embedded font

kjarosh:
Flash Player does not show any text that requires an embedded font that is missing.

As we now support DefineFont4, we're safe to remove the fallback device font.

* Related to https://github.com/ruffle-rs/ruffle/issues/18877
* Breaks https://github.com/ruffle-rs/ruffle/issues/14740
* Breaks https://github.com/ruffle-rs/ruffle/issues/18945


core/src/html/layout.rs
```diff
@@ -862,9 +862,9 @@ impl<'a, 'gc> LayoutContext<'a, 'gc> {
     fn left_alignment_offset(span: &TextSpan, is_first_line: bool) -> Twips {
         if span.bullet {
             if is_first_line {
-                Twips::from_pixels(35.0 + span.left_margin + span.block_indent + span.indent)
+                Twips::from_pixels(36.0 + span.left_margin + span.block_indent + span.indent)
```
adrian17, line 865: Was this just to make the tests pass better? The change is kinda out of nowhere ;D

core/src/html/layout.rs
```diff
@@ -660,10 +660,10 @@ impl<'a, 'gc> LayoutContext<'a, 'gc> {
             {
                 return font;
             }
-            // TODO: If set to use embedded fonts and we couldn't find any matching font, show nothing
-            // However - at time of writing, we don't support DefineFont4. If we matched this behaviour,
-            // then a bunch of SWFs would just show no text suddenly.
```
adrian17, line 665: just to be sure, did you test some SWFs like this?

adrian17 (approved):

core/src/html/layout.rs
kjarosh: Yes, I added a test that verifies how the bullet behaves when there's no font (to make sure I'm not breaking anything), and it turned out the bullet was misaligned by one pixel :sweat_smile:

core/src/html/layout.rs
kjarosh: Generally yes, I made sure that Flash behaves this way and most SWFs from my text collection work.

However, I've just remembered one SWF that used a wrong font: https://github.com/ruffle-rs/ruffle/issues/14740, and unfortunately checking it now, it turns out the text disappears in it. Which makes sense, because now Ruffle somehow fails to load the embedded font and falls back to a device font \:/

So I guess there are (and probably will be) cases when Ruffle fails to load an embedded font. Some probably are hard to spot.

PR #18859, opened 2024-12-04
misc: Add devcontainer config

torokati44:
This was done on a smartphone using GitHub Copilot.
So that I can do Rust stuff on a smartphone using GitHub Codespaces.
Judge accordingly.


torokati44:
I guess doing this instead would be better...?

torokati44:
Can confirm, this looks neater, and does add `cargo` to the tools in the created Codespace.
https://github.com/torokati44/ruffle/commit/e42d50b22238093b018c36c5a0256fd3da0735e5

torokati44:
See: https://containers.dev/overview

PR #18860, opened 2024-12-04
text: Fix `getCharBoundaries()` for 0-width characters

kjarosh:
Basically, `getCharBoundaries()` returns null for 0-width characters. This PR also fixes a bug where some characters (with missing glyphs) were not taken into account properly because `Font::evaluate` skipped them.


core/src/font.rs
```diff
@@ -613,6 +619,9 @@ impl<'gc> Font<'gc> {
                 // Step horizontally.
                 transform.matrix.tx += twips_advance;
                 x += twips_advance;
+            } else {
+                // No glyph, zero advance
+                glyph_func(pos, &transform, &Glyph::empty(), Twips::ZERO, x);
```
adrian17, line 624: What's the difference between processing a "noop" glyph and skipping it entirely?

adrian17 (approved):

core/src/font.rs
kjarosh: I've described it in my commit message:

```
This is a natural assumption of Font::evaluate and makes writing code easier.
Code that renders glyphs will work the same way as non-renderable characters
will produce empty glyphs, but code that evaluates the font for measurement
purposes will use this property.
```

I'll add a short comment about it here too.

kjarosh: Done

adrian17 (approved):

PR #18861, opened 2024-12-04
avm2: Make `TextField.getCharIndexAtPoint` accurate

kjarosh:
This patch fixes `getCharIndexAtPoint()` so that it's accurate for all inputs.


core/src/display_object/edit_text.rs
```diff
@@ -2109,6 +2109,38 @@ impl<'gc> EditText<'gc> {
         )
     }
 
+    pub fn char_index_at_point(self, position: Point<Twips>) -> Option<usize> {
```
adrian17, line 2112: can you please add some docstrings to better distinguish this from `screen_position_to_index`? From just signature and used coordinate space they sound super similar, but name style and implementation are super different.

kjarosh: Done

Lord-McSweeney (approved):

Issue #18848, opened 2024-12-03
Misaligned text on NYTimes graphics


martgnz:
### Describe the bug

We integrated Ruffle in our archive last year to recover past flash graphics. This has gone very well and lots of them are working without problems (thanks so much for your work on Ruffle!).

Now, we are trying to update Ruffle from 2023.08.07 nightly to latest and we're seeing some regressions in the text rendering related to position on the y axis. We were wondering if this is a known problem:

![Image](https://github.com/user-attachments/assets/77d2bcc8-2ada-4c50-b732-dd1c8c03dbf2)
![Image](https://github.com/user-attachments/assets/f11f00d2-69de-4ff9-84b3-bb5a5690b880)
![Image](https://github.com/user-attachments/assets/5e2f0f24-8732-4989-9e9e-a480c1688245)

### Expected behavior

Labels are positioned correctly, as on 2023.08.07:

![Image](https://github.com/user-attachments/assets/d00380a5-64ae-4c6f-89a8-311fc7a86298)
![Image](https://github.com/user-attachments/assets/16f07063-51de-4ae1-9917-bd966e9d9fd0)
![Image](https://github.com/user-attachments/assets/07edb973-06d8-484f-a2ae-5bcb97f1e2be)

### Content Location

- https://archive.nytimes.com/www.nytimes.com/interactive/2010/02/26/sports/olympics/20100226-olysymphony.html?_r=1
- https://archive.nytimes.com/www.nytimes.com/interactive/2009/11/06/business/economy/unemployment-lines.html
- https://archive.nytimes.com/www.nytimes.com/elections/2008/results/president/explorer.html

Since this is self-hosted and we haven't updated Ruffle publicly yet you can grab the SWF source and play it on the extension:

- https://static01.nyt.com/packages/flash/newsgraphics/2010/0225-olytimes/BasicChartDataProject.swf
- https://static01.nyt.com/packages/flash/newsgraphics/2009/1102-lines/BasicChartDataProject.swf
- https://static01.nyt.com/packages/flash/elections/2008/maps/o_ElectionTrends.swf

Not all graphics are affected by this, e.g. these ones work with newest Ruffle and labels look good:

- https://archive.nytimes.com/www.nytimes.com/interactive/2009/07/31/business/20080801-metrics-graphic.html
- https://archive.nytimes.com/www.nytimes.com/interactive/2010/01/10/nyregion/20100110-netflix-map.html
[...]

Issue #16357, opened 2024-05-17
Incorrect textfield align


pngdrift:
### Describe the bug

the label near textinput does not align to the right
![image](https://github.com/ruffle-rs/ruffle/assets/98256152/8d70d877-6516-433c-bec6-cae3845a4d57)

### Expected behavior

![image](https://github.com/ruffle-rs/ruffle/assets/98256152/1e9f5b51-33ea-44ec-9b1f-b1079cc7cd29)

### Content Location

[Tanksguitest.zip](https://github.com/ruffle-rs/ruffle/files/15352017/Tanksguitest.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

Issue #14988, opened 2024-01-29
Regression: Versions 2023-11-10+ Have Broken Input Textfields 


astroUNL:
### Describe the bug

Starting with 2023-11-10 and up to present (2023-01-29) some input textfields are broken:

1. The fields do not display text.
2. The user is unable to enter text.

The particular fields with the problem are created with the `createTextField` method. However, other text fields created with that method work (such as the degree symbol fields). I have been unable to determine what it is about these fields that cause them to not appear.

<img width="1137" alt="ruffle-text-bug" src="https://github.com/ruffle-rs/ruffle/assets/49266991/26b790af-a91a-461b-9dae-a69c048f4bab">

### Expected behavior

The textfields should display text and allow input, as shown in the picture.

### Content Location

Source SWF:
https://astro.unl.edu/classaction/animations/coordsmotion/altazimuth.swf

Working (2023-11-09, self-hosted):
https://astro.unl.edu/classaction/animations/coordsmotion/altazimuth.html

Broken (2023-11-10, unpkg):
https://astro.unl.edu/classaction/animations/coordsmotion/altazimuth.html?ruffleVer=0.1.0-nightly.2023.11.10

(The URL query string can be used to specify specific versions hosted at unpkg.com/@ruffle-rs, otherwise a locally hosted copy of the 2024-11-09 version of Ruffle is used. Sometimes using unpkg is slow.)

There are other Flash sims with the same problem on this page:
https://astro.unl.edu/animationsLinks.html

### Affected platform

Self-hosted version (also desktop version)

### Operating system

Windows, macOS

### Browser

Chrome

### Additional information

I am the author of this SWF and I still have the FLA.

britt-j:
The regression seems to have happened in https://github.com/ruffle-rs/ruffle/commit/924e5ef632253c711f36d37ebad9758eafd4d24d, which was a part of https://github.com/ruffle-rs/ruffle/pull/13923

britt-j:
The code responsible for the regression is: https://github.com/ruffle-rs/ruffle/blob/6059031d71d4480c5f861b6b7c44adbb74bb9684/core/src/display_object/edit_text.rs#L814-L815 The SWF sets the height of the textfield to 0 on initialization and never explicitly sets it to something else, meaning it'll be reset to zero when the autosize mode is none, which it does set it to.
[...]

Issue #11511, opened 2023-06-13
Renegades - Text errors 


desertking:
### Describe the bug

https://www.newgrounds.com/portal/view/642311?emulate=flash - Basically the game seems to work fine but there are text errors and they are only 50% readable. Dunno why but maybe can be fixed rather easy?

### Expected behavior

See the complete text and messages. 

### Affected platform

Online demo

### Operating system

All

### Browser

All

PR #18870, opened 2024-12-05
text: Set autosize bounds lazily

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/18848
* Fixes https://github.com/ruffle-rs/ruffle/issues/14988
* Fixes https://github.com/ruffle-rs/ruffle/issues/16357
* Fixes https://github.com/ruffle-rs/ruffle/issues/11511
* Fixes https://github.com/ruffle-rs/ruffle/issues/19114
* Progresses https://github.com/ruffle-rs/ruffle/issues/15413
* Progresses https://github.com/ruffle-rs/ruffle/issues/16880
* Requires https://github.com/ruffle-rs/ruffle/pull/18861
* Supersedes https://github.com/ruffle-rs/ruffle/pull/7749
* Related to https://github.com/ruffle-rs/ruffle/pull/13923

Auto size does not set the bounds immediately, but lazily after some selected operations, independently of relayout.

This patch removes the `requested_width/height` and replaces it with proper lazily set autosize bounds.

Some tests are added which prove that relayout and lazy autosize bounds are independent of each other, and prove _when_ autosize bounds are lazily updated.


adrian17:
Can you also please test https://github.com/ruffle-rs/ruffle/issues/11632 on this?

kjarosh:
> Can you also please test #11632 on this?

It seems that the text boxes there already work on master (probably due to https://github.com/ruffle-rs/ruffle/pull/18494, but just guessing)

tests/tests/swfs/text/auto_size/return/output.txt
```diff
@@ -1,12 +1,12 @@
-150
+103
```
adrian17, line 1: was the old test output created wrongly?
It the new one equal to FP?

adrian17 (approved):

tests/tests/swfs/text/auto_size/return/output.txt
kjarosh: That's correct. Somehow the `output.txt` wasn't from FP

adrian17:
To be sure, is anything blocking this?

kjarosh:
One test fails due to https://github.com/ruffle-rs/ruffle/pull/18861 missing

PR #18827, opened 2024-12-01
avm1: Try calling properties on root if the current scope isn't callable

ydalton:
Thanks to @Herschel for the fix, and @Dinnerbone for suggesting the commit message.

In the final boss of Super Mario 63, Mario hangs in the air, and thus cannot complete this battle and finish the game. This commit fixes that bug. See issue #10582 for the technical details.

Closes: #10582

PR #18885, opened 2024-12-06
avm1: Try calling properties on root if the current scope isn't callable

ydalton:
Redo of #18827, I did a thing (put the commit on a topic branch on master, then discarded the extra commits on main) and it auto-closed my PR.

### DIFF
--- a/core/src/avm1/activation.rs
+++ b/core/src/avm1/activation.rs
@@ -2818,7 +2818,17 @@ impl<'a, 'gc> Activation<'a, 'gc> {
 
         // Finally! It's a plain old variable name.
         // Resolve using scope chain, as normal.
-        self.resolve(path)
+        match self.resolve(path)? {
+            CallableValue::UnCallable(Value::Undefined) => {
+                let obj = self.target_clip_or_root().object().coerce_to_object(self);
+                if obj.has_property(self, path) {
+                    obj.get(path, self).map(|v| CallableValue::Callable(obj, v))
+                } else {
+                    Ok(CallableValue::UnCallable(Value::Undefined))
+                }
+            }
+            val => Ok(val),
+        }
     }
 
     /// Sets the value referenced by a target path string.


Molisson:
Could you add a test for this? You can use this from the SM63 issue:
https://github.com/ruffle-rs/ruffle/files/13326666/deleted-clip-ef.zip

thatsprettygood:
@ydalton

Tachytaenius:
Tried to implement this on modern downloaded source. Got:
```error[E0599]: no method named `object` found for enum `display_object::DisplayObject<'gc>` in the current scope```

**Comment by ChrisCPI** [comment]
These are my findings regarding this behavior that I dug into some months ago, but only shared in the Ruffle Discord:

(all of what is described below comes from flash player)

In a case i've made, there's the `_root`, and then the MC `instance1` inside of it, and then `instance2` which is inside of `instance1`.

- the root has a variable called `rootVar` set to `true`
- `instance1` has two frames: the 1st where `instance2` is added, the 2nd where it is removed. and instance1 also has a variable called `rootVar` set to the string `"PARENT"`.

Now, inside of `instance2`, there is this code:
```as
_root.onEnterFrame = function() {
    trace("from the inner, rootVar is is " + rootVar);
    trace("from the inner, target is: " + _target);
    delete onEnterFrame;
}
_parent.gotoAndStop(2)
```

When this is ran, the first trace statement prints `true`, so its from the root. likewise `_target` is also that of the root. this is not new information.

But then I discovered that if I changed it to `_parent.onEnterFrame` instead of `_root.onEnterFrame`, then it started tracing the `rootVar` and `_target` belonging to the parent (instance1), implying that it just doesn't default to the root every time like what this pr does(?).

Now, here's behavior coming from **Ruffle** (NOT with this PR's changes):

At the root:
```as
function foo() {
    trace("foo:");
    trace(_target);
    trace(this);
    delete _root.onEnterFrame;
}
stop();
```

inside `instance1`, child of root:
```as
_root.onEnterFrame = function() {
    delete _root.onEnterFrame;
    foo();
}
_root.gotoAndStop(2);
```

In ruffle, this doesn't call foo() at the root, as we know. However, inside of `instance1`, if `foo` is defined *in any way*, like this:
```as
var foo = true;
_root.onEnterFrame = function() {
    delete _root.onEnterFrame;
    foo();
}
_root.gotoAndStop(2);
```
*then* it actually calls the function at the root. So in Super Mario 63's case, defining `CharCode` in the affected movieclip does indeed fix the final boss.

With the earlier case of _root -> instance1 -> instance2, if i put a `foo()` function in instance1, and then this in instance2:
```as
var foo = true;
_parent.onEnterFrame = function() {
    foo();
    delete onEnterFrame;
}
_parent.gotoAndStop(2);
```
in ruffle, it still goes to call the foo() at the _root, while in flash player it calls the one at the parent (instance1). Also in both foo() functions, it traces `this` followed by `_target`.

Output from Flash Player:
```
calling foo from inner
FOO FROM PARENT:
_level0.instance1
/instance1
```

Output from Ruffle:

```
calling foo from inner
FOO FROM ROOT:

/
```

Lastly, I then tried a timeout:

```as
var foo = true;
function doTrace() {
    trace("calling foo from inner");
    foo();
}
setTimeout(doTrace, 500);
_root.gotoAndStop(2);
```
tracing this in flash player:
```
calling foo from inner
FOO FROM ROOT:
_level0
/
```
And Ruffle tracing the same thing as before:
```
calling foo from inner
FOO FROM ROOT:

/
```

Adamillo:
Any updates with this branch?

relrelb:
Superseded by #24725.

Issue #18895, opened 2024-12-07
Nightly 2024-11-26 creates bug in FlashChess III


Benman2785:
### Describe the bug

since [Nightly 2024-11-26](https://github.com/ruffle-rs/ruffle/releases/tag/nightly-2024-11-26) every pop-up-screen is shown when loading game = its now unusable

### Expected behavior

that the swf is working correctly

### Content Location

latest ruffle:
https://hoffmeister.li/flash-games/Flash-Chess_3.html

latest working ruffle:
https://hoffmeister.li/flash-games/Flash-Chess_3_w.html

### Affected platform

Self-hosted version

### Operating system

latest Win11

### Browser

Firefox, Chrome

### Additional information

sry for late reporting -> had stressfull 4 weeks :/

n0samu:
There are some errors related to perspectiveProjection:
```
ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "addedToStage", class: flash.events::Event, ptr: 0x1f65eda6fd0 }) to handler FunctionObject(FunctionObject { ptr: 0x1f65ea27f90, name: Ok("flashCHESS3_fla::MainTimeline/__setPerspectiveProjection_()") }) : TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: fieldOfView)
        at flashCHESS3_fla::MainTimeline/__setPerspectiveProjection_()
ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x1f67153f240 }) to handler FunctionObject(FunctionObject { ptr: 0x1f671a37e40, name: Ok("md.flashchess::Chess/initApplication()") }) : TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: projectionCenter)
        at md.flashchess::Chessfull/prepareInterface()
        at md.flashchess::Chess/initApplication()
```
@kjarosh I guess this is probably related to one of your stub PRs (#18744 or #18753)

PR #18908, opened 2024-12-08
avm2: Fix `perspectiveProjection`, `matrix3D` stubs

kjarosh:
Fixes stubs broken by https://github.com/ruffle-rs/ruffle/pull/18753 (now with tests!).

* Fixes https://github.com/ruffle-rs/ruffle/issues/18895


adrian17 (approved):

Lord-McSweeney:
Does this interact with `matrix` at all? If you set `matrix3D`, can you get back a corresponding `matrix` from the `matrix` getter or will it return null?

kjarosh:
These stubs do not change the behavior of `matrix` at all. It's just the original behavior + returning `null` when appropriate.

PR #18911, opened 2024-12-08
avm2: Use direct slot access in more native code

Lord-McSweeney:
This also fixes a bug with `BitmapData.paletteMap` not using the right coordinates for the destination point, and another bug with `PressAndTapGestureEvent::tapStageX`/`Y` not working at all.

Notable changes:
- `Object::is_instance_of` has been replaced with the new method `Value::is_instance_of`
- Uses of `FunctionObject::from_function` have been replaced with `FunctionObject::from_method`. Previously, the former allocated a prototype while the latter did not; now, `from_method` always allocates a prototype.


core/src/avm2/object/function_object.rs
```diff
@@ -105,23 +105,6 @@ impl<'gc> FunctionObject<'gc> {
     ///
     /// This associated constructor will also create and initialize an empty
```
adrian17, line 106: leftover docs

core/src/avm2/object.rs
```diff
@@ -229,6 +234,16 @@ pub trait TObject<'gc>: 'gc + Collect + Debug + Into<Object<'gc>> + Clone + Copy
         self.base().get_property_local(name, activation)
     }
 
+    /// Same as get_property_local, but constructs a public Multiname for you.
+    fn get_string_property_local(
```
adrian17, line 238: BTW, I think all three of these `*_string_property_local` could be `#[no_dynamic]`?

core/src/avm2/object.rs
```diff
@@ -1429,6 +1402,7 @@ pub enum WeakObject<'gc> {
     FontObject(FontObjectWeak<'gc>),
     LocalConnectionObject(LocalConnectionObjectWeak<'gc>),
     SharedObjectObject(SharedObjectObjectWeak<'gc>),
+    SoundTransformObject(SoundTransformObjectWeak<'gc>),
```
adrian17, line 1405: Any strong reason for doing this? I think I'd prefer this to keep it on the AS3 side, especially if others like `Point` are also on the AS3 side.

core/src/avm2/globals/flash/display/Shader.as
```diff
@@ -4,7 +4,9 @@ package flash.display {
     import flash.utils.ByteArray;
 
     public class Shader {
+        [Ruffle(InternalSlot)]
```
adrian17, line 7: (note: some shader-related changes got mixed in with the Rectangle commit)

core/src/avm2/object.rs
Lord-McSweeney: There aren't any accessible slots, unlike in Point, and we need to access the properties from `display_object.rs`, so I thought this worked better than the repeated `get_slot`s in there.

adrian17 (approved):

PR #18933, opened 2024-12-10
text: Treat empty font name the same way as `null`

kjarosh:
Flash Player ignores setting font with an empty name, and treats it the same way as `null`.  However, it does differentiate between `""` and `null` inside `TextFormat`.

* Fixes https://github.com/ruffle-rs/ruffle/issues/14740


adrian17 (approved):

Benman2785:
does nothing for: https://github.com/ruffle-rs/ruffle/issues/18945 Space Race

Issue #3725, opened 2021-03-21
Keypresses Get Missed Due to the Lack of an Input Buffer


ItsMaximum:
Both the Official Adobe Flash Player and Ruffle can miss short keypresses due to the lack of an input buffer. Using a keyboard with an average scan rate of around 50hz (19-20ms shortest key press, tested with http://blog.seethis.link/scan-rate-estimator/ ), inputs that start and stop in-between the framerate of the game being played (usually 30fps, updating every 33.333ms) can be missed. An input buffer that stores inputs to be processed on each frame of gameplay would solve this issue.

Herschel:
This is intentional to match Flash behavior, and dependent on the specific code used; for example, the usual `Key.isDown` in an `enterFrame` loop naturally misses sub-frame presses because the code only runs on frame intervals. Key listeners such as `on(keyPress)` will correctly catch the press and fire a handler regardless of framerate.

That said, I do agree with adding this as an optional quality-of-life feature. We could provide an option to ensure that any key press is seen as down for at least one frame.

ActionWavele:
good idea

pinkgothic:
On Terry Cavanaugh's "Naya's Quest" (original game here: https://terrycavanaghgames.com/nayasquest/ - I downloaded the `.swf` some years ago and used to use Adobe Flash Player locally to play it once my browser stopped supporting it), I run into either #this issue or key duplication (maybe #12319?), depending on luck. I'm using the Desktop version of Ruffle (Windows 11).

Due to this, it's very hard to complete the game, the core game mechanic of which depends on allowing you to walk yourself off a cliff. I need to repeat keypresses a lot and hope that once one registers, animations are slow enough so potential multiple keypresses come to a stop before the animation cycle for a single keypress ends.

I never observed not-registering key presses in Adobe Flash Player, but it might just be the emulation speed making it visible. (Ruffle _is_ at least an order of magnitude slower at running the game then the standalone Flash Player was.)

PR #18974, opened 2024-12-15
core: Add compatibility flags

kjarosh:
Compatibility flags help configure Ruffle by enabling and disabling specific behaviors.

They are meant to be used for instance in the following situations.

1. They fix bugs in Flash Player that make some content misbehave. Note that in general we don't fix bugs in Flash Player -- we are bug compatible. However, there are examples where a bug is so sneaky, some content could have been created with an assumption that the bug doesn't exist and the bug affects it.

2. They genuinely improve the experience of using Ruffle at the cost of lowering compatibility with Flash Player.

3. They improve the "perceived" compatibility at the cost of the "real" compatibility. For instance, something does not work in Flash Player, and we make it work.

The main goal of this PR is to prepare for adding new flags in the future. This way we'll have a simple interface to configure such behaviors in the future. The most notable example would be https://github.com/ruffle-rs/ruffle/pull/18856 where we could add a compatibility flag which enables font fall back again for content that is not yet fully working in Ruffle due to various bugs.

<image src="https://github.com/user-attachments/assets/16f03e82-2ac6-4732-8ac9-14ce5518fefa" height="200"/>
<image src="https://github.com/user-attachments/assets/279c3873-8da6-4c1f-a8d1-19d3b0601d80" height="200"/>
<image src="https://github.com/user-attachments/assets/3dc1e3c8-a1cc-4086-b131-9742cd59d412" height="200"/>
<image src="https://github.com/user-attachments/assets/922ee927-1c9b-4d2f-8435-471ef3ba134d" height="200"/>


n0samu:
As a matter of UX, we should definitely give the right column a more descriptive name like "FP default" so people understand what it means

kjarosh:
Forgot to add a screenshot!

<image src="https://github.com/user-attachments/assets/6c0d2165-ac1e-47a3-9002-510b69a8cb99" height="200"/>

The "FP" label does have a tooltip, I decided to keep the name short in order not to waste horizontal space (it has only a checkbox).

emergence75:
can this also used in the actionsscript classes ?
if this is possible, it would be a nice solution for the correct hack here
https://github.com/ruffle-rs/ruffle/pull/18495

danielhjacobs:
Another good one would be https://github.com/ruffle-rs/ruffle/issues/3725

thisisjonathan:
Would it make sense to make this part of the "Open advanced..." options instead of being a global preference? Imagine someone who changes a flag to make a game work and then later plays a different game forgetting that they changed it and then that game doesn't work right because of the flag. You'd also have to ask about the flags each time each time you're troubleshooting.

Another advantage of making it part of the launch options is that you could make it part of a Ruffle bundle, so you could bundle the game with the flags that are optimal for that game.

I'm working on a Flash editor based on Ruffle. Part of the appeal of Flash for me is that it (with some exceptions, like device fonts) works the same on my PC as on someone else's. It would be annoying if my games didn't work for some people because they turned on a flag a long time ago and forgot about it. It would also be bad if someone used my editor to make a game, forgot that they had a flag turned on and accidentally relied on the behavior with the flag turned on. Then when other people try the game, it doesn't work right for them.

kjarosh:
> Would it make sense to make this part of the "Open advanced..." options instead of being a global preference?

I think it would. I wouldn't say that this PR is like the ultimate implementation of the GUI for compatibility flags, I'd rather treat it as the starting point and ability to put something behind a flag at all.

The arguments you mentioned got me thinking and I'm rethinking whether saving compatibility flags to preferences is a good idea as of now. 

> Another advantage of making it part of the launch options is that you could make it part of a Ruffle bundle, so you could bundle the game with the flags that are optimal for that game.

I agree with that. But also note that the content deciding about stuff going on within Ruffle is not always the best idea. That's something we have to approach very carefully in order not to give too much control to the content.

I'd say if we go with this, each flag should be assessed individually whether it can be set within the content.

Issue #15485, opened 2024-03-08
Chess Demons not loading


Benman2785:
### Describe the bug

Error dispatching event EventObject(EventObject { type: "addedToStage", class: flash.events::Event, ptr: 0xc2a580 }) to handler FunctionObject(FunctionObject { ptr: 0xc2a508, name: Ok("Main/init()") }) : RustError("A script in this movie has taken too long to execute and has been terminated.")

### Expected behavior

that it loads the game

### Content Location

https://hoffmeister.li/flash-games/ChessDemons.html
https://hoffmeister.li/flash-games/swf/ChessDemons.swf

### Affected platform

Self-hosted version

### Operating system

Win11 latest

### Browser

Firefox & Chrome

Benman2785:
now loads - but copy-protection is also on ;)

Benman2785:
doesnt load anymore in newer versions
last working is: ruffle-nightly-2024_10_25-web-selfhosted

Benman2785:
https://youtu.be/-eQKBSNCsI4

Benman2785:
with [Nightly 2024-12-14](https://github.com/ruffle-rs/ruffle/releases/tag/nightly-2024-12-14) ChessDemons load again :)

Benman2785:
works (since a while now)

PR #18995, opened 2024-12-16
avm2: Remove PrimitiveObject

### LINKED ISSUE #15517: RegExp valueOf returns a string instead of object
RegExp.prototype.valueOf currently returns a string (similar to toString) instead of returning the RegExp object. Actually I think we need to add an actual RegExp.prototype.valueOf method, instead of just implementing the [`value_of`](https://github.com/ruffle-rs/ruffle/blob/0466605b62609f47f69a2a97b25e2a9b3a37cb64/core/src/avm2/object/regexp_object.rs#L90) method of the TObject trait.

Testcase: `typeof /abc/.valueOf()`

Lord-McSweeney:
Supersedes #13152.

Testing on #15485 shows no significant change in its startup time.

Closes #15517.

This also refactors event dispatch code to accept an EventObject instead of a Value or Object


core/src/avm2/globals/array.rs
```diff
@@ -334,7 +333,7 @@ pub fn to_locale_string<'gc>(
 /// array while this happens would cause a panic; this code exists to prevent
 /// that.
 pub struct ArrayIter<'gc> {
-    array_object: Object<'gc>,
+    array_object: Value<'gc>,
```
adrian17, line 336: Still think this should be an Object, especially considering its main use point might be converted to `get_index_property` eventually.

core/src/avm2/parameters.rs
```diff
@@ -107,25 +106,21 @@ impl<'gc> ParametersExt<'gc> for &[Value<'gc>] {
...
     ) -> Option<Object<'gc>> {
         match self[index] {
-            Value::Null | Value::Undefined => None,
+            Value::Null => None,
             Value::Object(o) => Some(o),
-            primitive => Some(
-                PrimitiveObject::from_primitive(primitive, activation)
-                    .expect("Primitive object is infallible at this point"),
-            ),
+            _ => panic!("try_get_object should read Object or Null"),
```
adrian17, line 123: why not keep the `Undefined` variant support?
As is, this mostly just forced changing a bunch of `try_get_object` to more verbose `get_value().as_object()`. (and also feels out of scope?)

tests/tests/swfs/avm2/verify_abnormal_loop/output.txt
```diff
@@ -1 +1 @@
-ReferenceError: Error #1069: Property field not found on int and there is no default value.
+ReferenceError: Error #1069: Property field not found on Number and there is no default value.
```
adrian17, line 1: oh, nice catch.

core/src/avm2/object.rs
```diff
@@ -245,72 +241,6 @@ pub trait TObject<'gc>: 'gc + Collect + Debug + Into<Object<'gc>> + Clone + Copy
...
-    /// Retrieve a property by Multiname lookup.
-    ///
-    /// This method should not be overridden.
-    ///
-    /// This corresponds directly to the AVM2 operation `getproperty`, with the
-    /// exception that it does not special-case object lookups on dictionary
-    /// structured objects.
-    #[allow(unused_mut)] //Not unused.
-    #[no_dynamic]
-    fn get_property(
```
adrian17, line 257: why not keep these, if only as
```
    fn get_property(...) {
        Value::from(self).get_property(...)
    }
```
?
This would prevent several dozen `Value::from(self).get_property(...)` changes in other places, and feels like a nice QoL wrapper to have in general.

adrian17 (approved):

Issue #19011, opened 2024-12-18
(NSFW) Animation #6 - Shade - Can only type three numerical numbers instead of four


Noxious45:
### Describe the bug

When trying to type the password into the keypad at the beginning to get a secret NSFW scene, (code is 2141) only three numbers can be put in instead of four 

https://github.com/user-attachments/assets/609b5ffd-801b-4080-98d9-11b3e7f8675b

### Expected behavior

The player should be able to type a maximum of four numbers into the keypad

https://github.com/user-attachments/assets/57043423-eb56-4c36-ad0b-565cf09a8122

### Content Location

https://www.newgrounds.com/portal/view/697775

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Chrome version 131.0.6778.140 

### Additional information

# Player Info
Allows script access: false
Renderer: wgpu
Adapter Backend: Gl
Adapter Name: "ANGLE (NVIDIA, NVIDIA GeForce RTX 3060 (0x00002504) Direct3D11 vs_5_0 ps_5_0, D3D11)"
Adapter Device Type: Other
Adapter Driver Name: ""
Adapter Driver Info: "WebGL 2.0 (OpenGL ES 3.0 Chromium)"
Enabled features: Features(TEXTURE_COMPRESSION_BC | FLOAT32_FILTERABLE | TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES)
Available features: Features(DEPTH32FLOAT_STENCIL8 | TEXTURE_COMPRESSION_BC_SLICED_3D | PUSH_CONSTANTS | CLEAR_TEXTURE | MULTIVIEW)
[...]

PR #19012, opened 2024-12-18
text: Skip setting HTML text when it hasn't changed

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/19011

Flash Player checks whether we're setting `htmlText` to a different value before doing it. It not only prevents text relayout, but also has observable effects, because not every set of spans can be represented as HTML.

This is especially affecting input text fields with bound variables, as text propagation will set the value back, and we can produce spans not representable as HTML during input.

Additionally, this PR fixes `<p>`'s align case sensitivity.


adrian17:
> This is especially affecting input text fields with bound variables, as text propagation will set the value back,

What do you mean? Aren't they supposed to be always synced anyway?

kjarosh:
> > This is especially affecting input text fields with bound variables, as text propagation will set the value back,
> 
> What do you mean? Aren't they supposed to be always synced anyway?

They are, but due to the fact that HTML cannot represent all possible contents, it gets tricky. I'm talking specifically about the following scenario:

1. The text field has value `1\n` (represented as `<p align="center">1</p>`)
2. The user deletes `\n`, leaving `1` (represented as `<p align="center">1</p>`)
3. Now the binding kicks in, it sets the value of the variable
4. The binding synchronizes the value back, setting the text field to `<p align="center">1</p>`
5. The text field now has value `1\n`, which is wrong (it should be `1`)

Now the fact that the binding synchronizes the value back doesn't matter, because when you do the same without binding, it behaves the same way (i.e. the value is `1`, not `1\n`). This case is covered by the test.

Unfortunately, the tests now started to fail, which means I have found yet another bug related to when the relayout takes place (yay! I guess?).

adrian17:
To be sure, is the full roundtrip ("binding synchronizes the value back") needed, or is it more like "it's simpler this way and doesn't impact any behavior by itself"?

kjarosh:
> To be sure, is the full roundtrip ("binding synchronizes the value back") needed, or is it more like "it's simpler this way and doesn't impact any behavior by itself"?

So generally, I did it this way not because it's simpler, but because it's correct. The binding has nothing to do with this behavior, it just happens to occur with a binding.

I don't think the roundtrip is needed (but I'm not sure), I think after this PR it will just be a useless string comparison + property lookup. I guess that could be optimized, but I'm not sure if that's worth it.

adrian17:
> So first of all, I did it this way not because it's simpler, but because it's correct.

I wasn't questioning the PR at all, sorry if it seemed that way. Was just making sure about the roundtrip.

adrian17 (approved):

adrian17 (changes requested):
Ah, sorry, reversing the approve, since you mentioned that bug 😅

kjarosh:
> I wasn't questioning the PR at all, sorry if it seemed that way. Was just making sure about the roundtrip.

No worries! I wasn't thinking that, just wanted to emphasize (for everyone) that the binding thing is not that important here :smile:

kjarosh:
@adrian17 Should be good now, added some more test cases here too

adrian17 (approved):

PR #19022, opened 2024-12-20
tests: Rebuild AAC decoding test without network permission to make it work in FP

torokati44:
I had to add the `-use-network=false` argument to MXMLC, so that Flash Player allows the SWF to access the local FLV files.
Ruffle doesn't care about this yet, so the test was technically passing.


kjarosh (approved):

kjarosh:
Just curious: how did you make sure it works in FP with network enabled? :thinking:

torokati44:
The SWF was probably built with Flash first, when I made the original test, then later I made some minor cleanups, and rebuilt it with MXMLC. I think.

Issue #12124, opened 2023-07-17
Test failure: shumway's acid-textfield-scroll


Dinnerbone:
Test path: `tests/tests/swfs/visual/shumway_acid_tests/acid_textfield_scroll`

# Expected
![image](https://github.com/ruffle-rs/ruffle/assets/317625/e10dbf9c-6ce4-4621-9dec-61380a6e9327)
```
Correct max horizontal scrolling position: yes
Max vertical scrolling position: 3
Current vertical scrolling position: 3
Current bottomScrollV: 4
Current horizontal scrolling position: 50
```

# Actual
![image](https://github.com/ruffle-rs/ruffle/assets/317625/3dc08f69-2701-4166-a0aa-776f4a8128a5)
```
Correct max horizontal scrolling position: no
Max vertical scrolling position: 4
Current vertical scrolling position: 3
Current bottomScrollV: 4
Current horizontal scrolling position: 0
```

# Difference
![image](https://github.com/ruffle-rs/ruffle/assets/317625/bebd1642-b3c6-44b2-a326-3b2543936388)

PR #19023, opened 2024-12-20
avm2: Handle negative `TextField.scrollH` properly

kjarosh:
Flash Player does an abs on `scrollH` in AVM2, contrary to AVM1.

* Fixes https://github.com/ruffle-rs/ruffle/issues/12124


core/src/avm2/globals/flash/text/text_field.rs
```diff
@@ -1466,7 +1466,7 @@ pub fn set_scroll_h<'gc>(
             .cloned()
             .unwrap_or(Value::Undefined)
             .coerce_to_i32(activation)?;
-        let clamped = input.clamp(0, this.maxhscroll() as i32);
+        let clamped = input.abs().clamp(0, this.maxhscroll() as i32);
```
Lord-McSweeney, line 1469: Could the `clamp(0, this.maxhscroll())` call be changed to `min(this.maxhscroll())` now?

kjarosh: Certainly! Nice catch

Lord-McSweeney (approved):

Issue #12120, opened 2023-07-17
Test failure: shumway's acid-text-2


Dinnerbone:
Test path: `tests/tests/swfs/visual/shumway_acid_tests/acid_text-2`

# Expected
![image](https://github.com/ruffle-rs/ruffle/assets/317625/475be8b1-0dea-4b5e-87a6-0bb3d66f95d7)
```
Text is: AABBBBBBBBBBBBBBBBBBBBAB
```

# Actual
![image](https://github.com/ruffle-rs/ruffle/assets/317625/90cae875-7ad4-44fb-a3a1-dd0ccb93b5ca)
```
Text is: 
```

# Difference
![image](https://github.com/ruffle-rs/ruffle/assets/317625/3c3f93ba-f818-4529-bca5-d2a5194d2b9c)

PR #19024, opened 2024-12-20
avm2: Implement `StaticText.text`

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/12120


core/src/font.rs
```diff
@@ -806,6 +810,9 @@ pub struct Glyph {
 
     shape: GlyphShape,
     advance: Twips,
+
+    // The character this glyph represents.
+    character: char,
```
evilpie, line 815: Maybe `Option<char>` instead?

core/src/avm2/globals/flash/text/static_text.rs
```diff
@@ -16,9 +17,17 @@ pub fn static_text_allocator<'gc>(
...
 ) -> Result<Value<'gc>, Error<'gc>> {
-    avm2_stub_getter!(activation, "flash.text.StaticText", "text");
-    Ok(activation.strings().empty().into())
+    if let Some(this) = this
+        .as_object()
+        .and_then(|this| this.as_display_object())
+        .and_then(|this| this.as_text())
+    {
+        let text = this.text(activation.context);
+        return Ok(AvmString::new(activation.context.gc_context, text).into());
```
evilpie, line 29: nit: `activation.gc()`

evilpie (approved):

core/src/avm2/globals/flash/text/static_text.rs
kjarosh: Right!

core/src/font.rs
kjarosh: If you're referring to the replacement character, I don't believe `Option<char>` is appropriate here, as it seems that Flash just doesn't care whether `u16` is a surrogate code point or not (which is the requirement behind Rust's `char`). However, the replacement character is used everywhere, so I just did the same and added a TODO comment (just now). I believe we need a wider refactor of Flash characters here, but that's outside of this PR.

PR #19029, opened 2024-12-20
web: Do not scroll the page when mouse wheel is handled in AVM2

kjarosh:
This patch prevents scrolling the page when the users scrolls content inside the SWF (i.e. when the content somehow handles the scroll).

Based on my testing this behavior is a lot smarter than Flash originally had, because Flash just trapped mouse wheel events. We might want to add an option to simulate this behavior too in the future.

You can easily test the behavior here: https://www.newgrounds.com/portal/view/922049.


Dinnerbone:
Oh no this is going to bring up the "where's the line" debate again :D

This does _not_ match Flash's behaviour, AS3 mouse wheel events explicitly do not prevent the page from scrolling. It was a known trick to use AVM1 for scroll wheel stuff, which _did_ prevent scrolling.

The linked page in original Flash does not prevent scrolling.

kjarosh:
So... I tested this a bit more, and I originally did tests with default params and assumed the result is always the same. Turns out the mouse wheel is trapped only when `wmode` is `window`, `opaque`, or `transparent` (not on `direct`, `gpu`). It's `direct` on NG, and it was `window` in my tests.

torokati44:
Is this FP behavior consistent across browsers and platforms? 🤔

kjarosh:
> Is this FP behavior consistent across browsers and platforms? 🤔

Turns out this behavior differed across browsers and varied in time. We had a discussion about it on Discord, and I guess the bottom line was not to follow Flash Player in this regard.

Pasting some links from @n0samu:
* https://stackoverflow.com/questions/5477426/issues-with-wmode-opaque-and-issues-with-wmmode-window
* https://petethepig.blogspot.com/2012/02/handling-mouse-wheel-event-with-wmode.html

kjarosh:
Updated the PR and added a `scrollingBehavior` setting:
* `never` — scrolling Flash content never scrolls the page,
* `always` — scrolling Flash content always scrolls the page,
* `smart` (default) — scrolling Flash content scrolls the page only when scrolling is not handled.

Is that acceptable @Dinnerbone ?

Dinnerbone:
> Updated the PR and added a `scrollingBehavior` setting:
> 
>     * `never` — scrolling Flash content never scrolls the page,
> 
>     * `always` — scrolling Flash content always scrolls the page,
> 
>     * `smart` (default) — scrolling Flash content scrolls the page only when scrolling is not handled.
> 
> 
> Is that acceptable @Dinnerbone ?

To be clear, `smart` is Flash Player's default yeah? Could you add that to the documentation? ("This matches Flash Player's behaviour" or such)

Otherwise - yes, thank you so much! Sorry for the fuss 😅

kjarosh:
> To be clear, `smart` is Flash Player's default yeah? Could you add that to the documentation? ("This matches Flash Player's behaviour" or such)

That's the Ruffle default. I think the issue is that Flash doesn't have a default, the behavior differed across versions, browsers, and wmodes (it was either `always` or `never` for AVM2, `smart` for AVM1 according to my knowledge).

Dinnerbone:
Okay, sure. Let's go with this!
Thanks again! <3

Dinnerbone (approved):

kjarosh:
(added a sentence about the behavior in FP)

PR #19053, opened 2024-12-24
web: Reduce docker build IO

divinity76:
- binaryen xz is no longer stored on disk, instead being piped directly to tar
- only extract wasm-opt (the only file we're interested in)
- locate wasm-opt using wildcard, instead of keeping path in sync with version number.

we're doing the same with sh.rustup.rs below.

### DIFF
--- a/web/docker/Dockerfile
+++ b/web/docker/Dockerfile
@@ -6,9 +6,9 @@ FROM node:22
 
 # Installing wasm-opt from GitHub:
 # Keep the version number in sync with the ones in the Actions workflows!
-RUN wget --progress=:giga https://github.com/WebAssembly/binaryen/releases/download/version_121/binaryen-version_121-x86_64-linux.tar.gz
-RUN tar xzf binaryen-version_121-x86_64-linux.tar.gz binaryen-version_121/bin/wasm-opt
-RUN mv binaryen-version_121/bin/wasm-opt /usr/local/bin
+RUN wget --progress=:giga https://github.com/WebAssembly/binaryen/releases/download/version_121/binaryen-version_121-x86_64-linux.tar.gz -O- | \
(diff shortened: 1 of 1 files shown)


torokati44:
Then I assume `--quiet` would make more sense instead of `--progress=:giga`, like below.

Also please tag the commit message (and PR title) according to https://github.com/ruffle-rs/ruffle/blob/master/CONTRIBUTING.md#commit-message-guidelines, and add a `T-*` and `A-*` label.

divinity76:
> Then I assume `--quiet` would make more sense instead of `--progress=:giga`, like below.

progress=:giga is there for a reason. The download can take a long time, it's 100MB compressed and sometimes github's download servers are randomly slow. Without the progress indicator, the docker build may appear to hang. Progress indicator is printed to stderr and does not corrupt the data stream for tar. Also it wasn't introduced by me, it appears to be introduced by you in a66ba1505196d5fd77e79c32d5c8abb13fbd6db8.

> Also please tag the commit message (and PR title) according to https://github.com/ruffle-rs/ruffle/blob/master/CONTRIBUTING.md#commit-message-guidelines

done

>, and add a `T-*` and `A-*` label.

I don't understand, please elaborate

danielhjacobs:
Regarding the labels those can only be added if you have at least triage access to the repo which I don't think you have. I'll add them.

torokati44:
> Without the progress indicator, the docker build may appear to hang. Progress indicator is printed to stderr and does not corrupt the data stream for tar. 

> Regarding the labels those can only be added if you have at least triage access to the repo which I don't think you have. I'll add them.

Both good points, and thank you!

torokati44 (approved):

danielhjacobs (approved):

PR #19061, opened 2024-12-25
refactor: Use constants from `Twips` where appropriate

kjarosh:
It's a refactor that replaces `Twips` instantiations with respective constants. Some instantiations in tests (by `Twips::new`) were left intact as they seem to focus on the pure numerical value instead of semantically the number of twips.

### DIFF
--- a/core/src/avm1/globals/transform.rs
+++ b/core/src/avm1/globals/transform.rs
@@ -146,10 +146,10 @@ fn method<'gc>(
             // If the bounds are invalid, the pixelBounds rectangle consists only of zeroes.
             let bounds = if world_bounds == Rectangle::default() {
                 Rectangle {
-                    x_min: Twips::new(0),
-                    x_max: Twips::new(0),
-                    y_min: Twips::new(0),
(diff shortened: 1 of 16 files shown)


torokati44:
For some reason, reading `Twips::ONE` in itself, it's not obvious to me that it's in fact one pixel's worth of twips, and not one twip...

kjarosh:
> For some reason, reading `Twips::ONE` in itself, it's not obvious to me that it's in fact one pixel's worth of twips, and not one twip...

Yeah... That's a bit confusing. I don't have a proposal of a better name though... :/ `ONE_PIXEL` is kinda long

n0samu:
I had the same confusion and was thinking of bringing it up too. How about ONEPX? (Or ONE_PX if you prefer)

danielhjacobs (approved):
Bikeshedding aside, LGTM

evilpie:
I like `ONE_PX` as well. Let's get this done.

evilpie (approved):

Issue #19042, opened 2024-12-23
Text display overlaps in https://www.maxgames.com/play/robot-wants-kitty.html


JimenezLi:
### Describe the bug

The title screen button text overlaps in Ruffle.
![Image](https://github.com/user-attachments/assets/57961fb2-1026-47dc-91af-d1577d659e8a)

### Expected behavior

The text shouldn't overlap.
![Image](https://github.com/user-attachments/assets/1ec7ff6b-9425-4061-83a4-2645c69274db)

### Content Location

https://farm.maxgames.com/robotkittymaxmar25OTcw.swf
[robotkittymaxmar25OTcw.zip](https://github.com/user-attachments/files/18226881/robotkittymaxmar25OTcw.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

Both desktop and browser have this problem

kjarosh:
This game uses flixel, everything is rendered on a bitmap

kjarosh:
It seems that the root cause is the following set of operations on a `BitmapData`:

1. `.draw()` with `tx=0,0`
2. `.fillRect()` with `ARGB=0`
3. `.draw()` with `tx!=0,0`

After 3, the bitmap contains old pixels from 1. When I do a read operation (e.g. `getPixel`, `getPixel32`) between 1 and 2, everything works properly.

kjarosh:
So I suspect something's wrong with 

https://github.com/ruffle-rs/ruffle/blob/c9afa5b7c1a0b6303770a1a9fa4f1b7cb0174b10/core/src/bitmap/bitmap_data.rs#L392

being used here

https://github.com/ruffle-rs/ruffle/blob/c9afa5b7c1a0b6303770a1a9fa4f1b7cb0174b10/core/src/bitmap/operations.rs#L51

When I replace `overwrite_cpu_pixels_from_gpu` with `sync`, everything works properly. If I understand correctly, `overwrite_cpu_pixels_from_gpu` should discard any GPU buffer as we're overwriting everything on the CPU side. Later, when syncing we should overwrite the GPU buffer with the CPU one, but something must be going wrong there, as the discarded GPU buffer is somehow used later (and perhaps we're drawing over it).

PR #19066, opened 2024-12-25
render: Flush draws before updating bitmap texture

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/19042

This fixes an issue where new draws were drawn over an old GPU buffer. For instance, consider the following scenario:

    1. draw (GPU)
    2. full fill (CPU)
    3. draw (GPU)

When full fill is executed, the GPU buffer is discarded without a sync, but we have to make sure we wait for any queued draws in order to truly discard 1.

Thank you @Dinnerbone for providing the fix!


Dinnerbone (approved):

torokati44:
Is this perhaps related to the regressions caused by https://github.com/ruffle-rs/ruffle/pull/15032?

kjarosh:
> Is this perhaps related to the regressions caused by #15032?

Yes, it is!

PR #18399, opened 2024-10-28
Revert "ci: Freeze rustc for nightly web builds to 1.81.0"

torokati44:
Reverts ruffle-rs/ruffle#18397.

To be merged once https://github.com/rustwasm/wasm-bindgen/pull/4213 is merged and released, so https://github.com/rustwasm/wasm-bindgen/issues/4227 can be fixed properly.

Also referencing https://github.com/rustwasm/wasm-bindgen/issues/4211.

PR #19075, opened 2024-12-27
Revert "chore: Fix a nightly `clippy::unnecessary_map_or` lint"

danielhjacobs:
Reverts ruffle-rs/ruffle#19073

Can only be re-merged after https://github.com/ruffle-rs/ruffle/pull/18399, which itself needs other PR(s): https://github.com/ruffle-rs/ruffle/pull/18399#issuecomment-2514682267

The suggested code has only been available in stable since Rust 1.82: https://doc.rust-lang.org/std/option/enum.Option.html#method.is_none_or

### DIFF
--- a/core/src/avm2/e4x.rs
+++ b/core/src/avm2/e4x.rs
@@ -1230,7 +1230,7 @@ impl<'gc> E4XNode<'gc> {
         };
 
         // 2.a. If N.prefix == "" and x.[[Name]].uri == "", return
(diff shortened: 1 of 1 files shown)


torokati44:
> The suggested code has only been available in stable since Rust 1.82

Oh for crying out loud, I even went and looked up the docs and saw this, and separately I was aware of the 1.81 lock, just not at the same time... 😭

Sorry and thank you!

PR #19093, opened 2024-12-30
Removed "rust" as a keyword in the .desktop

RubisetCie:
First of all, I know how this might looks, and I don't wish to sound "insulting" or petty. :)
I know you're all proud rustaceans, and you should be, because Rust is a great programming language. Be proud of your project, Ruffle is amazing! :D

But... was it necessary to put "rust" as a keyword in the .desktop file? To me the end user shouldn't care about what programming language was used to develop Ruffle... besides, it is already mentionned in the package description and it messes up my start menu when I type "rust"...

However don't worry, I wouldn't mind much if you choose to reject my PR. Have a great day.

### DIFF
--- a/desktop/packages/linux/rs.ruffle.Ruffle.desktop.in
+++ b/desktop/packages/linux/rs.ruffle.Ruffle.desktop.in
@@ -7,4 +7,4 @@ Icon=rs.ruffle.Ruffle
 Exec=ruffle %u
 MimeType=application/x-shockwave-flash;application/vnd.adobe.flash.movie
 Categories=AudioVideo;Player;Graphics;Viewer;VectorGraphics;Game
-Keywords=flash;swf;player;emulator;rust
+Keywords=flash;swf;player;emulator
--- a/desktop/packages/linux/rs.ruffle.Ruffle.metainfo.xml.in
+++ b/desktop/packages/linux/rs.ruffle.Ruffle.metainfo.xml.in
@@ -43,7 +43,6 @@
         <keyword>swf</keyword>
         <keyword>player</keyword>
         <keyword>emulator</keyword>
-        <keyword>rust</keyword>
     </keywords>
     <url type="homepage">https://ruffle.rs</url>
     <url type="bugtracker">https://github.com/ruffle-rs/ruffle/issues</url>


Croworbit:
I'll assume you're having this problem with the flatpack release
have you tried downloading the standalone versions, or compiling from source?
Ruffle doesn't necessarily need to be installed, and that should avoid this issue

RubisetCie:
No in fact, I'm already building Ruffle from source using cargo...

kjarosh:
Hi @RubisetCie and thank you for your contribution! I think you are making a great point—keywords are about searching, and we should ask the question whether the user wants to see Ruffle when they are searching for "rust".

I believe it's a good idea to rethink keywords in general, as modifying each keyword means we have to redo all of the translations (we'd modify them only in `*.in` files, as the rest is generated).

Croworbit:
that's odd. I was only ever able to replicate your problem with the start menu after installing the flatpack version, as I compile ruffle myself often and don't encounter this issue.

kjarosh:
> have you tried downloading the standalone versions, or compiling from source?
> Ruffle doesn't necessarily need to be installed, and that should avoid this issue

I don't think that's particularly helpful, as you give up major features that way, which are important to some people (automatic updates, easy management, sandboxing).

Not mentioning the fact that we should strive to improve the experience of using Ruffle.

Croworbit:
you make regular ruffle sound crippled or unsafe, which I don't agree to be the case

kjarosh:
Just to make sure @RubisetCie, the issue is fixed after applying changes from this PR? How do you install Ruffle after building?

RubisetCie:
@kjarosh Yes, just removing "rust" in the keyword list is enough to make everything perfect to me. :)

@Croworbit I think there may be a misunderstanding, I'm mainly talking from a "semantic" point of view; when the users are typing "rust" in the start menu, I think they should find developer tools (like RustRover for example), but not applications developped in Rust... don't you agree?

Croworbit:
> @Croworbit I think there may be a misunderstanding, I'm mainly talking from a "semantic" point of view; when the users are typing "rust" in the start menu, I think they should find developer tools (like RustRover for example), but not applications developped in Rust... don't you agree?

yes, I do agree
that's wh\y I said what I said in parenthesis in my earlier comment

adrian17 (approved):

kjarosh (approved):

kjarosh:
Thank you!

PR #19134, opened 2025-01-04
desktop: Use async IO for file browsing in AVM

kjarosh:
Using async IO means we're not blocking the main thread and the content doesn't freeze while reading.

In theory, Ruffle should use only async IO, as otherwise the main thread will be blocked by every bit of IO performed. According to my knowledge this is the first instance of using async IO in Ruffle.

Things to consider:
* apparently `tokio::fs` is slow: https://github.com/tokio-rs/tokio/issues/3664,
* do we want to use `tokio::fs`, or `async-fs` (or something else) so that we are compatible with other runtimes (referring specifically to `frontend_utils`).

### DIFF
--- a/desktop/Cargo.toml
+++ b/desktop/Cargo.toml
@@ -44,7 +44,7 @@ chrono = { workspace = true }
 fluent-templates = "0.12.0"
 toml_edit = { version = "0.22.22", features = ["parse"] }
 gilrs = "0.11"
-tokio = { workspace = true, features = ["rt-multi-thread", "macros"]}
+tokio = { workspace = true, features = ["rt-multi-thread", "macros", "fs"] }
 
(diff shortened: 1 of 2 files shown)


Aaron1011 (approved):

torokati44:
Do you think this is mergeable as is, despite the points of doubt mentioned above, @kjarosh?

kjarosh:
In my opinion:

1. This particular case of IO is caused by user interaction, so slight performance drop is perfectly acceptable especially if it makes the application responsive.

2. I think all async APIs are drop in replacements of `std::fs`, so we can easily switch to a different backend in the future if we decide.

IMO it's fine to merge, we should be aware of those issues when migrating the rest of IO.

torokati44:
Alright, thank you!

Issue #17143, opened 2024-07-16
Dead Maze/Transformice missing TextField.styleSheet


JediFay:
### Describe the bug

![ruffle_deadmazestylesheet](https://github.com/user-attachments/assets/008c82ad-6808-4f6f-a9c8-e880118d6fbc)
![ruffle_tfmstylesheet1](https://github.com/user-attachments/assets/6902a647-d71d-4348-9076-44197350b9b5)
![ruffle_tfmstylesheet2](https://github.com/user-attachments/assets/a5fe161a-665f-4185-9ae8-df3ec664b118)

### Expected behavior

![flash_deadmazestylesheet](https://github.com/user-attachments/assets/fc384bf4-f41d-4d7b-9ebe-e76fadd2ad52)
![flash_tfmstylesheet1](https://github.com/user-attachments/assets/e5856418-01f3-4db3-bac8-449b65e4c71e)
![flash_tfmstylesheet2](https://github.com/user-attachments/assets/a588845a-afaa-4f5d-bd56-32b1ecdc58ae)

### Content Location

https://www.deadmaze.com/alpha/chargeur-deadmeat.swf
https://www.transformice.com/TransformiceChargeur.swf

### Affected platform

Desktop app

### Operating system

Windows 11

Issue #16880, opened 2024-06-25
Text isn't correct Galaxy Harvest


Charumu:
### Describe the bug

Text in Menu and In-game is black and not bright green as it should be causing it to be barely readable if at all
![obraz](https://github.com/ruffle-rs/ruffle/assets/170186282/5ea1dd17-61fd-4999-8730-8732fe3f8072)

### Expected behavior

Text like in screenshots of it from before flash shutdown
![obraz](https://github.com/ruffle-rs/ruffle/assets/170186282/c2d23848-852d-4760-8dad-b9a5e14ea1af)

### Content Location

You can use SWF:
http://chat.kongregate.com/gamez/0022/4673/live/BiomassRelease.swf
or just play it in kongregate or any other site that has it and uses ruffle

### Affected platform

Desktop app

### Operating system

Nobara 39 x86_64

### Additional information

terminal logs from running the game and shutting it down after Menu loads:
`2024-06-25T12:16:38.459597Z  INFO ruffle_desktop::gui::controller: Using preferred backend Vulkan
2024-06-25T12:16:38.493929Z  INFO ruffle_desktop::gui::controller: Using graphics API vulkan on NVIDIA GeForce RTX 3050 Laptop GPU (type: DiscreteGpu)
2024-06-25T12:16:38.513999Z  WARN egui_wgpu::renderer: Detected a linear (sRGBA aware) framebuffer Bgra8UnormSrgb. egui prefers Rgba8Unorm or Bgra8Unorm    
2024-06-25T12:16:38.514471Z  INFO ruffle_desktop::gui::controller: loaded cjk fallback font "NotoSans-Regular"
[Legacy] Serving exact file: /home/maciej/Pobrane/flashpoint-nano-master/games/legacy/chat.kongregate.com/gamez/0022/4673/live/BiomassRelease.swf
2024-06-25T12:16:38.587745Z  INFO ruffle_core::context: Loaded SWF version 11, resolution 800x600 @ 30 FPS
2024-06-25T12:16:38.603999Z  WARN ruffle_core::stub: Encountered stub: AVM2 flash.display.DisplayObject.scale9Grid setter
2024-06-25T12:16:38.604117Z  INFO ruffle_desktop::backends::ui: Loading device font "TimesNewRomanPSMT" for "Times New Roman" (italic: false, bold: false)
2024-06-25T12:16:38.604392Z  INFO ruffle_core::library: Loaded new device font "Times New Roman" from file
[...]

Issue #14832, opened 2024-01-19
The Amazing Spider-Man Online Movie Game: Missing letters in some text fields


n0samu:
### Describe the bug

In [The Amazing Spider-Man Online Movie Game](https://www.numuki.com/game/spider-man-online-movie-game/), letters are missing from some text fields:
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/a9288a2b-5d6c-4273-a7b3-6451e70dacc3)
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/ff76ad34-428c-4ca3-a851-eae96c62abc9)
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/09efd6de-82b3-49fc-842b-3c2455a3ea5d)

Due to CORS and ExternalInterface errors, you'll need to test the game in the Ruffle desktop app:
`ruffle https://media.numuki.com/spider-man/games/online-movie-game/main.swf --dummy-external-interface`

### Expected behavior

Text fields should not be missing letters.

### Content Location

Page URL: https://www.numuki.com/game/spider-man-online-movie-game/
Main SWF URL: https://media.numuki.com/spider-man/games/online-movie-game/main.swf
This is a multiasset game, so downloading the main SWF is not enough to run it locally. If you want to inspect the files, I recommend using Flashpoint.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

The game uses a separate [fontLib.swf](https://media.numuki.com/spider-man/games/online-movie-game/nullassets/swf/fontLib.swf) file to store the complete copies of its fonts. Most of the buttons and text fields with this problem are stored in the [shellAssets.swf](https://media.numuki.com/spider-man/games/online-movie-game/nullassets/swf/shellAssets.swf) file. This file has incomplete copies of the required fonts with only enough characters to display the text fields' default/initial text. The fontLib.swf file gives custom names to the fonts: AKZIGBEB, AKZIGBER, etc. The text seen in the game is loaded via the [config.xml](https://media.numuki.com/spider-man/games/online-movie-game/xml/config.xml) (or `config.preloader.php` in the Flashpoint version). The font is set via CSS. For example, here's the PLAY NOW button:
`<play style="tall_btn" ><![CDATA[PLAY NOW]]></play>`
And later in the XML, here's the applicable CSS:
`.tall_btn {font-family:AKZIGBEM;color:#bcbcbc;font-size:22px;}`
So I think this needs CSS support in AVM2.

[...]

Issue #12257, opened 2023-07-23
Text Cave: Wrong text color, text doesn't fit on screen


jmousy:
### Describe the bug

This is an AVM2 game. The known title is **'Text Cave'**.

[text-cave.zip](https://github.com/ruffle-rs/ruffle/files/12138257/text-cave.zip)

I know there are a lot of text improvements that still need to be implemented in Ruffle, but I thought this game would be a good text rendering example, so I submit an issue.

As you can see in the photo, the stage is not displaying as intended because the text is not positioned correctly on the screen due to the font color, font family, font weight, and (possibly) font width of the text.

![image](https://github.com/ruffle-rs/ruffle/assets/75052171/5a79e7f3-0dfc-4720-be04-2aeb9c8ef2d1)

### Expected behavior

The text should appear in the correct position.

### Affected platform

Desktop app

### Operating system

Windows 10 22H2

### Browser

Google Chrome 115

kjarosh:
Device fonts are used here (Consolas)

Jmousy:
Thanks for the additional information. I probably missed it, as the font-related functionality was missing at the time I created this issue.

However, after applying Consolas (Windows 11 default installed), I was able to see the difference from the original as shown below: the colors are still not being applied and the font size seems to be wrong.
(2024-12-09 Nightly)

![Image](https://github.com/user-attachments/assets/725addcb-7efa-4515-99d6-4038cf7864e1)

(But I may have missed something more....)

Reproduction link: https://flasharch.com/archive/play/9a87fabfd1fcf1ffea60d3ee4450996c

kjarosh:
> However, after applying Consolas (Windows 11 default installed), I was able to see the difference from the original as shown below: the colors are still not being applied and the font size seems to be wrong. (2024-12-09 Nightly)
> 
> Reproduction link: https://flasharch.com/archive/play/9a87fabfd1fcf1ffea60d3ee4450996c

The site detects dev tools being open and redirects, which makes it difficult to debug what's wrong, unfortunately.

In general, the SWF is highly environment-dependent, that's what it looks like in Flash Player for me:

![Image](https://github.com/user-attachments/assets/2a79fdf2-2c9d-4a45-899d-4d5fa0b0f2b8)

We're for sure missing colors in the text and we should assume monospace for device font `Consolas`.

jmousy:
Thank you for your investigation.

[...]

PR #19143, opened 2025-01-06
avm2: Add style sheet support for TextField

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/17143
* Fixes https://github.com/ruffle-rs/ruffle/issues/16880
* Fixes https://github.com/ruffle-rs/ruffle/issues/14832
* Progresses https://github.com/ruffle-rs/ruffle/issues/12257
* Progresses https://github.com/ruffle-rs/ruffle/issues/8937

This patch adds `TextField.styleSheet` support.  Now when parsing HTML, the attached style sheet is used to provide styles for HTML elements.

This patch also adds support for:
* `class` attributes in HTML,
* `<span>` HTML elements,
* custom HTML elements.

CSS support is spanned across all existing HTML elements and custom ones.

Additionally, some observable behaviors related to `.styleSheet` are implemented, e.g. performing relayout after setting `.styleSheet`, or .text behaving as `.htmlText` with style sheet set.

Currently, the implementation is coupled with AVM2 objects, that should change when AVM1 support is added.


n0samu:
Somebody should check #14832!

core/src/avm2/globals/flash/text/StyleSheet.as
```diff
@@ -131,5 +119,8 @@ package flash.text {
         private native function innerParseCss(css: String): Object;
         private native function innerParseColor(color: String): Number;
         private native function innerParseFontFamily(fontFamily: String): String;
+
+        private native function getStyle0(selector: String): Object;
+        private native function setStyle0(selector: String, styleObject: Object, textFormat: TextFormat): void;
```
Lord-McSweeney, line 124: I think `getStyleInternal`/`setStyleInternal` would be better for consistency with similar native functions

kjarosh: done!

Dinnerbone (approved):
Amazing!

kjarosh:
Before merging, I'm thinking I unnecessarily rewrote `_styles` to Rust. I could probably simplify the code a lot by moving it back to AS 🤔 (I'll look into it tomorrow)

kjarosh:
Simplified the code in the second commit a lot. Sorry for the trouble @Dinnerbone ! (but I think it was worth it)

Dinnerbone (approved):

PR #19147, opened 2025-01-06
 extension: Fix types which are changed in latest @types/chrome 

danielhjacobs:
Needed for #19144


torokati44:
This doesn't seem like a pure formatting change to me, but thank you!

danielhjacobs:
Updated PR title and commit message.

PR #19198, opened 2025-01-13
web: Don't unnecessarily duplicate the module for single WASM builds

danielhjacobs:


web/packages/core/tools/set_version.ts
```diff
@@ -72,8 +77,16 @@ const options = {
...
         /%COMMIT_HASH%/g,
+        /%WASM_MVP%/g,
+    ],
+    to: [
+        versionNumber,
+        versionName,
+        versionChannel,
+        buildDate,
+        commitHash,
+        wasmMvpName,
```
torokati44, line 88: It's a bit strange to call this `%WASM_MVP%`/`wasmMvpName`, when it is only _maybe_ MVP.
I'm not saying I have a better name, but later down the line, this can get confusing. At least a comment about this here and there would be nice IMHO.

torokati44: Maybe the words "other", "fallback", or "compatibility" could inspire a better name...?

danielhjacobs: `fallbackWasmName` and `%FALLBACK_WASM%` sound good to me. `other` has the same issue imo; it implies there are always more than one (one and an**other**). `compatibility` isn't quite the same issue, but it does imply to me that it refers to the WASM module with the most compatibility at all times (so the MVP module).

`fallback` works best, because it just means, to me, that this is the WASM module we fall back to, which may or may not be the MVP module.

danielhjacobs: https://github.com/ruffle-rs/ruffle/blob/23ed90e42d0e6c820abe80babd8cb09057c241b9/web/packages/core/src/load-ruffle.ts#L48-L52

Opinion question: Is it worth changing this to `if (!extensionsSupported && "%FALLBACK_WASM%" === "ruffle_web") { `?

Currently, that logs on Pale Moon even when doing a single WASM build.

torokati44: I suppose, yeah. With a short comment about the reason for it maybe.

(I always try to imagine either us, or someone else, looking at every independent piece of the added/changed code independently, a year so from now - having forgotten every context discussed here or Discord, and not looking at the entire cohesive diff -; and try to make sure it will still make sense then. That's why I'm so insistent on commenting every piece of logic like this.)

torokati44: And maybe later we can also qualify the module name of the one currently called simply "ruffle_web" as '"ruffle_web-mvp" or something like that.

torokati44 (approved):
The diff looks okay, though I suppose this can be squashed into a single commit. Thank you!

Dinnerbone:
We can remove the copying code from build_wasm.ts now, right?

torokati44:
Hah, good point, this is only the loading part so far... 😅

danielhjacobs:
Ah, the copying part complains about these:

https://github.com/ruffle-rs/ruffle/blob/0f4ecc70ca24537906130941497b3091aeeb6bc1/web/packages/core/src/load-ruffle.ts#L12

https://github.com/ruffle-rs/ruffle/blob/0f4ecc70ca24537906130941497b3091aeeb6bc1/web/packages/core/src/internal/builder.ts#L1

https://github.com/ruffle-rs/ruffle/blob/0f4ecc70ca24537906130941497b3091aeeb6bc1/web/packages/core/src/internal/player/inner.tsx#L1

torokati44:
You mean, if it's not copied, then the build complains about it missing?

danielhjacobs:
Yes, but I can `@ts-expect-error` and switch the type imports to come from `%FALLBACK_WASM%`.

Dinnerbone:
Those should be pointing to the one that always exists, which is now `-wasm_extensions`, right?

danielhjacobs:
Fixed

torokati44:
Thank you!

PR #19216, opened 2025-01-14
extension: Update declarativeNetRequest rules using the chrome namespace

danielhjacobs:
I prefer using `utils` whenever possible, but sometimes DefinitelyTyped makes that harder than it should be. This should let #19206 be merged I think.

### DIFF
--- a/web/packages/extension/src/background.ts
+++ b/web/packages/extension/src/background.ts
@@ -141,7 +141,7 @@ async function enableSWFTakeover() {
                     },
                 },
             ];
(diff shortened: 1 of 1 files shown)


torokati44:
I know less about this than you, so ... sure?

torokati44:
Ooh, https://github.com/ruffle-rs/ruffle/pull/18995 was almost done auto-merging... :sweat_smile: Now it has to run CI again. Doesn't matter, we don't pay for the wasted CPU cycles (directly). :D

PR #19219, opened 2025-01-14
avm2: Fix URShift in optimizer and hasnext in interpreter when passed a negative index

Lord-McSweeney:
`op_has_next` should push `false`, not `0` when `cur_index` is negative, and Op::URShift should push the uint type in the optimizer, not the int type.

### DIFF
--- a/core/src/avm2/optimize.rs
+++ b/core/src/avm2/optimize.rs
@@ -951,7 +951,7 @@ pub fn optimize<'gc>(
                 Op::URShift => {
                     stack.pop(activation)?;
                     stack.pop(activation)?;
-                    stack.push_class(activation, types.int)?;
+                    stack.push_class(activation, types.uint)?;
                 }
                 Op::PushDouble { .. } => {
                     stack.push_class(activation, types.number)?;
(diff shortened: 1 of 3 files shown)


core/src/avm2/optimize.rs
```diff
@@ -1252,7 +1252,7 @@ pub fn optimize<'gc>(
                 Op::HasNext => {
                     stack.pop(activation)?;
                     stack.pop(activation)?;
-                    stack.push_any(activation)?;
+                    stack.push_class(activation, types.uint)?;
```
adrian17, line 1255: Proxy's `nextNameIndex()` is also declared to return an int, so you can manufacture a case where `hasnext` yields -5. This means our `get_next_enumerant` is buggy and should return an i32...?
Anyway, the optimizer should be typing this as int.

adrian17 (approved):
Approved, assuming this gets merged with the haxnext change removed or changed to int.

core/src/avm2/optimize.rs
Lord-McSweeney: This was surprisingly hard to do, so I just changed it to push `Number`.

adrian17: Can you add a comment that ideally it should be an int?

PR #19227, opened 2025-01-14
chore: Update translations

RuffleBuild:


desktop/assets/texts/ar-SA/preferences_dialog.ftl
```diff
@@ -23,4 +23,10 @@ theme = المظهر
 theme-system = الإعدادات المبدئية
 theme-light = فاتح
 theme-dark = داكن
+# See for context https://github.com/FeralInteractive/gamemode
+gamemode = وضع اللعبة
```
kjarosh, line 27: I don't think that's right. IMO it should be transliterated or left as-is, not translated, similarly the translations below

evilpie (approved):

PR #19229, opened 2025-01-15
web: Qualify the vanilla (MVP) WASM module with a name suffix

torokati44:
Instead of the one with extensions enabled.

A continuation of https://github.com/ruffle-rs/ruffle/pull/19198.


danielhjacobs (approved):
Didn't test it but the code LGTM.

torokati44:
_CI is green, therefore nothing can break_.

PR #19236, opened 2025-01-15
avm2: Access Event.target in dispatchEvent

Lord-McSweeney:
Like #19127, but clones the Event in `EventDispatcher.dispatchEvent` instead of `events::dispatch_event`.


adrian17 (approved):
To be sure, this means the clone logic will never trigger if _we_ (as in, native code) are the ones doing dispatching. I'm assuming we're considering it safe?

Can you please also copy the test from my PR? It test a more real-world case and a getter being hit.

torokati44:
GitHub had a brainfart and thought that a no-op rebase would fail, so I force-pushed the last commit with the same changes, hopefully it will work this time.

torokati44:
Narrator: It did not.
What the heck?

torokati44:
Third time's the charm! :tada:
BTW the issue was https://www.githubstatus.com/incidents/78wybrzyv0wf, referencing https://github.com/orgs/community/discussions/149282.

PR #19248, opened 2025-01-16
avm2: Store locals on AVM stack instead of per-Activation RegisterSets

Lord-McSweeney:
This speeds up `getlocal`/`setlocal` by about 10%

### DIFF
--- a/core/src/avm2/activation.rs
+++ b/core/src/avm2/activation.rs
@@ -24,7 +24,6 @@ use crate::context::UpdateContext;
 use crate::string::{AvmAtom, AvmString, StringContext};
 use crate::tag_utils::SwfMovie;
 use gc_arena::Gc;
-use smallvec::SmallVec;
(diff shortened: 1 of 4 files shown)


core/src/avm2/activation.rs
```diff
@@ -214,23 +175,20 @@ impl<'a, 'gc> Activation<'a, 'gc> {
         let (method, global_object, domain) = script.init();
 
         let num_locals = match method {
-            Method::Native { .. } => 0,
+            Method::Native { .. } => 1,
```
adrian17, line 178: To be sure, is this even reachable?

adrian17 (approved):

core/src/avm2/activation.rs
Lord-McSweeney: The script initializer of the builtins script is a native method, so without this the builtin script fails to initialize

PR #19259, opened 2025-01-17
build: Build binaries for Linux AArch64

kjarosh:
GitHub added new runners with ARM processors, and we are able to easily build binaries for AArch64 that way!

Let's see how well this works!

### DIFF
--- a/tests/tests/swfs/avm2/pixelbender_dithering/test.toml
+++ b/tests/tests/swfs/avm2/pixelbender_dithering/test.toml
@@ -4,7 +4,7 @@ num_ticks = 1
 
 [image_comparisons.output]
 tolerance = 1
-max_outliers = 19813
+max_outliers = 19929
 
 [player_options]
 with_renderer = { optional = false, sample_count = 1 }
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -71,6 +71,9 @@ jobs:
           - build_name: linux-x86_64
             os: ubuntu-24.04
 
+          - build_name: linux-aarch64
+            os: ubuntu-24.04-arm
+
           # Mac does two Rust builds to make a universal binary
           - build_name: macos-x86_64
             os: macos-14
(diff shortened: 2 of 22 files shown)


colin-i:
I made a pull request for this fork at [https://github.com/kjarosh/ruffle/pull/8](url)
I don't know what values you are expecting for max_outliers, so I separated that when matrix.os is ubuntu-24.04-arm and added a separate max_outliers_arm value at few toml files.

torokati44:
> added a separate max_outliers_arm value at few toml files.

I don't think this is a good idea... 😶
It's not like the ARM runners are less accurate compared to Fladh Player, _because_ they are ARM... They are just inaccurate in a different way than x86, and the current expected outputs match that more.

colin-i:
The arm desktop builds are somehow workable at my games. Last arm build that was working was with the last commit from October. In November there is a commit that is causing this:
`thread 'main' panicked at desktop/src/app.rs:466:14:
GUI controller should be created: Limit 'max_color_attachments' value 8 is better than allowed 4`
Same through  Jan 17(this fork time).
Build logs are at my colin-i/ruffle actions.
I will resolve that at a future time if no one will do that.

kjarosh:
Those ARM pipelines are a bit flaky, but they mostly succeed. WI think we can merge them without marking them as required

tests/tests/swfs/avm2/pixelbender_dithering/test.toml
```diff
@@ -4,7 +4,7 @@ num_ticks = 1
 
 [image_comparisons.output]
 tolerance = 1
```
torokati44, line 6: I wonder whether allowing a bit more tolerance and fewer outliers would make more sense. Or adding a second check, with higher tolerance, now that it's plssible... 😉 It might be the case that the nature of the test conteadicts this, dunno... But either way, it may be out of scope for this PR...

kjarosh: Well... I don't think that works, I've tried it and even with tolerance=128 we have 8k outliers.

It looks like the output is just very very far off (but the visual effect looks similar).

torokati44:
I guess we could also add links to these builds to the website.

PR #19263, opened 2025-01-17
text: Limit font name length to 64 characters in TextFormat

kjarosh:
That's what Flash Player does ¯\\\_(ツ)\_/¯


adrian17 (commented):
Should every place where we use/construct `text_format` natively in (even font tags and avm1) have this logic?

kjarosh:
> Should every place where we use/construct `text_format` natively in (even font tags and avm1) have this logic?

Did more tests and it seems that no, it's just the AVM1/AVM2 setters which clamp it to 64 (e.g. it's not the case in HTML).

The reason I implement this is CSS, where you can specify a list of fonts and the list can get quite large, but as AVM2 setters are used there, its length is clamped to 64.

Regarding SWF tags, we have a buggy logic there anyway, because Flash treats each font class as a reference to a SWF tag, and not a font family per se (which we treat it as).

I pushed fixes for AVM1 too (with a test).

core/src/avm1/globals/text_format.rs
```diff
@@ -80,7 +80,10 @@ fn set_font<'gc>(
 ) -> Result<(), Error<'gc>> {
     text_format.font = match value {
         Value::Undefined | Value::Null => None,
-        value => Some(value.coerce_to_string(activation)?.as_wstr().into()),
+        value => {
+            let font = value.coerce_to_string(activation)?;
+            Some(font.slice(0..64).unwrap_or(&font).to_owned())
```
evilpie, line 85: Because it confused me: `slice` returns `None` when the lenth is less than 64.

evilpie (approved):

PR #14463, opened 2023-12-19
avm1: Don't allow for looking up with an empty name when children don't have an explicit name set

Lord-McSweeney:
Fixes #14369
Fixes #13210
Fixes #12961
Fixes #12877
Fixes #12253

PR #19272, opened 2025-01-18
core/avm1/avm2: Remove some static AvmStrings

Lord-McSweeney:
*Only* about 120 static strings left in AVM2!

### DIFF
--- a/core/src/avm1/object/stage_object.rs
+++ b/core/src/avm1/object/stage_object.rs
@@ -341,7 +341,7 @@ impl<'gc> TObject<'gc> for StageObject<'gc> {
             // Button/MovieClip children are included in key list.
             for child in ctr.iter_render_list().rev() {
                 if child.as_interactive().is_some() {
-                    keys.push(child.name());
+                    keys.push(child.name().expect("Interactive DisplayObjects have names"));
                 }
             }
         }
@@ -657,8 +657,10 @@ fn frames_loaded<'gc>(
         .map_or(Value::Undefined, Value::from)
 }
 
-fn name<'gc>(_activation: &mut Activation<'_, 'gc>, this: DisplayObject<'gc>) -> Value<'gc> {
-    this.name().into()
+fn name<'gc>(activation: &mut Activation<'_, 'gc>, this: DisplayObject<'gc>) -> Value<'gc> {
+    this.name()
+        .unwrap_or_else(|| activation.strings().empty())
+        .into()
 }
 
 fn set_name<'gc>(
--- a/core/src/avm2/globals.rs
+++ b/core/src/avm2/globals.rs
@@ -9,6 +9,7 @@ use crate::avm2::traits::Trait;
 use crate::avm2::value::Value;
 use crate::avm2::vtable::VTable;
 use crate::avm2::{Avm2, Error, Multiname, Namespace, QName};
+use crate::string::WStr;
 use crate::tag_utils::{self, ControlFlow, SwfMovie, SwfSlice, SwfStream};
 use gc_arena::Collect;
 use std::sync::Arc;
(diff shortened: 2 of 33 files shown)


core/src/display_object.rs
```diff
@@ -1556,10 +1556,7 @@ pub trait TDisplayObject<'gc>:
         }
     }
 
-    fn name(&self) -> AvmString<'gc> {
-        self.base().name().unwrap_or_default()
-    }
-    fn name_optional(&self) -> Option<AvmString<'gc>> {
+    fn name(&self) -> Option<AvmString<'gc>> {
```
adrian17, line 1559: fwiw, from the API pov, it does make sense for a DO to always have a name, I think?
We'd need to check, but maybe it's possible that a DO is always suppposed to have _some_ name (see `set_default_instance_name`)?

With that it mind, maybe eventually this still should return an `AvmString` and unwrap internally?

That said, I haven't tested any of this. I'm not suggesting to make the refactor bigger, rather want to avoid refactoring a dozen places to support an Option, which might be reverted in the future.

core/src/avm2/object/event_object.rs
```diff
@@ -261,12 +251,17 @@ impl<'gc> EventObject<'gc> {
 
     pub fn net_status_event(
         activation: &mut Activation<'_, 'gc>,
-        info: Vec<(impl Into<AvmString<'gc>>, impl Into<AvmString<'gc>>)>,
+        info: Vec<(&str, &str)>,
     ) -> EventObject<'gc> {
+        let event_type = AvmString::new_utf8(activation.gc(), "netStatus");
```
adrian17, line 256: We should do interning in most of these places, also for `key`.

core/src/avm2/globals.rs
```diff
@@ -827,11 +831,14 @@ macro_rules! avm2_system_class_defs_playerglobal {
     ($activation:expr, [$(($package:expr, $class_name:expr, $field:ident)),* $(,)?]) => {
         let activation = $activation;
         $(
+            let package = AvmString::new_utf8(activation.gc(), $package);
+            let class_name = AvmString::new_utf8(activation.gc(), $class_name);
```
adrian17, line 835: And definitely intern here.

core/src/avm1/object/stage_object.rs
```diff
@@ -341,7 +341,9 @@ impl<'gc> TObject<'gc> for StageObject<'gc> {
             // Button/MovieClip children are included in key list.
             for child in ctr.iter_render_list().rev() {
                 if child.as_interactive().is_some() {
-                    keys.push(child.name());
+                    if let Some(name) = child.name() {
```
Lord-McSweeney, line 344: This was unnecessary, all interactive DOs have names

core/src/display_object.rs
Lord-McSweeney: No, not all DOs have names. `Graphic`s that were placed without a name, for example, don't have names. This was what caused the bug that #14463 fixed: you cannot look up a `Graphic` child with _any_ name, even though if you somehow got access to a Graphic in AVM1 its `.name` would be an empty string.

core/src/avm2/globals.rs
```diff
@@ -812,9 +813,16 @@ macro_rules! avm2_system_classes_playerglobal {
     ($activation:expr, [$(($package:expr, $class_name:expr, $field:ident)),* $(,)?]) => {
         let activation = $activation;
         $(
+            // Package and class names are ASCII
+            let package = WStr::from_units($package.as_bytes());
+            let class_name = WStr::from_units($class_name.as_bytes());
+
+            let package = activation.strings().intern_wstr(package);
```
adrian17, line 820: `intern_static` should be possible here

adrian17 (approved):

core/src/avm2/globals.rs
Lord-McSweeney: Thanks, fixed!

PR #19274, opened 2025-01-18
text: Allow specifying multiple device fonts for text fields

kjarosh:
When device fonts are used, you can specify a list of font names for a text field, in which case the first one available will be used. No glyph fallback mechanism between fonts on the list is applied.

This is true for both CSS and pure `TextFormat`s.

Additionally some refactors are included.


core/src/html/layout.rs
```diff
@@ -666,48 +666,51 @@ impl<'a, 'gc> LayoutContext<'a, 'gc> {
...
-                    span.style.bold,
-                    span.style.italic,
-                    context.ui,
-                    context.renderer,
-                    context.gc_context,
-                )
-                .first()
-            {
+        // Specifying multiple font names is supported only for device fonts.
+        let font_names: Vec<&str> = font_name.split(",").collect();
```
adrian17, line 670: Would this work if you iterated on the split without collecting into a Vec?

adrian17 (approved):

core/src/html/layout.rs
kjarosh: I could do that, sure, but then I'm using the vec to get the fallback font name and IMO it's clearer when I do `font_names.first()` compared to not collecting into a vec and doing something else.

PR #19277, opened 2025-01-19
avm2: Stub more methods from FTE

kjarosh:
Stubbed methods:
* `ElementFormat.getFontMetrics`
* `TextLine.getAtomBidiLevel`
* `TextLine.getAtomCenter`
* `TextLine.getAtomGraphic`
* `TextLine.getAtomTextBlockBeginIndex`
* `TextLine.getAtomTextBlockEndIndex`
* `TextLine.getAtomTextRotation`
* `TextLine.getAtomWordBoundaryOnLeft`

### DIFF
--- a/core/src/avm2/globals/flash/text/engine/ElementFormat.as
+++ b/core/src/avm2/globals/flash/text/engine/ElementFormat.as
@@ -1,4 +1,8 @@
 package flash.text.engine {
+    import __ruffle__.stub_method;
+
+    import flash.geom.Rectangle;
+
     [API("662")]
     public final class ElementFormat {
         private var _alignmentBaseline:String;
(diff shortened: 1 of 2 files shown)


core/src/avm2/globals/flash/text/engine/ElementFormat.as
```diff
@@ -199,5 +203,13 @@ package flash.text.engine {
         public function set typographicCase(value:String):void {
             this._typographicCase = value;
         }
+
+        public function getFontMetrics():FontMetrics {
+            stub_method("flash.text.engine.ElementFormat", "getFontMetrics");
+            var emBox:Rectangle = new Rectangle(0, _fontSize * -0.8, _fontSize, _fontSize);
+            return new FontMetrics(
+                emBox, -5, 1.2, 1.8, 1.2, 0.075, 0.6, -0.35, 0.6, 0.0
```
adrian17, line 211: are these completely random numbers, or taken from some existing font? :D

adrian17 (approved):

core/src/avm2/globals/flash/text/engine/ElementFormat.as
kjarosh: Those are numbers taken from an existing font (roughly). I don't think their values matter, but their signs and orders of magnitude might.

PR #19301, opened 2025-01-22
avm2: Access proxy methods by dispatch id

Lord-McSweeney:
This way we don't need to construct a Multiname and use a static string at the call site

### DIFF
--- a/core/build_playerglobal/src/lib.rs
+++ b/core/build_playerglobal/src/lib.rs
@@ -488,7 +488,13 @@ fn write_native_table(data: &[u8], out_dir: &Path) -> Result<Vec<u8>, Box<dyn st
 
                         let (trait_name, const_name) =
                             rust_path_and_trait_name(&abc, trait_, parent);
-                        let const_name = quote::format_ident!("{}", const_name);
+
+                        // Make getters prefixed with GET_ and setters prefixed with SET_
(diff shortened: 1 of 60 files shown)


core/build_playerglobal/src/lib.rs
```diff
@@ -479,9 +475,39 @@ fn write_native_table(data: &[u8], out_dir: &Path) -> Result<Vec<u8>, Box<dyn st
...
+                            rust_path_and_trait_name(&abc, trait_, parent);
+
+                        // Make getters prefixed with GET_ and setters prefixed with SET_
+                        if matches!(trait_.kind, TraitKind::Getter { .. }) {
+                            const_name = "GET_".to_owned() + &const_name;
+                        } else if matches!(trait_.kind, TraitKind::Setter { .. }) {
+                            const_name = "SET_".to_owned() + &const_name;
+                        }
+
+                        let const_name = quote::format_ident!("{}", const_name);
```
moulins, line 499: This could be written more succintly as:
```rs
let const_name = match trait_.kind {
    TraitKind::Getter => format_ident!("GET_{}", const_name),
    TraitKind::Setter => format_ident!("SET_{}", const_name),
    _ => format_ident!("{}", const_name),
};

moulins (approved):
LGTM :)

PR #19304, opened 2025-01-22
Add "StartupNotify=false" to Linux .desktop file

Tiagoquix:
This fixes two windows appearing in the KDE Plasma taskbar when starting Ruffle under Wayland (the first one takes some time to disappear, and serves no purpose).

This probably should not be done this way, but for now this fixes it.

See https://specifications.freedesktop.org/desktop-entry-spec/latest/recognized-keys.html#id-1.7.6 and https://www.freedesktop.org/wiki/Specifications/startup-notification-spec/ for more details.

### DIFF
--- a/desktop/packages/linux/rs.ruffle.Ruffle.desktop
+++ b/desktop/packages/linux/rs.ruffle.Ruffle.desktop
@@ -73,3 +73,4 @@ Keywords[uk]=flash;swf;програвач;емулятор
 Keywords[zh_CN]=flash;swf;播放器;模拟器
 Keywords[zh_TW]=flash;swf;播放器;模擬器
 Keywords=flash;swf;player;emulator
+StartupNotify=false


kjarosh:
Welcome, and thanks for your contribution!

The desktop file is generated, you'd need to add this property in `.desktop.in`.

I'm not sure if that's a proper fix, based on the documentation I read it might actually work by accident (startup notification protocol is an X11 thing right?).

Can you check whether https://github.com/ruffle-rs/ruffle/pull/19308 fixes it?

kjarosh:
> Can you give me the link to the binary or publish a temporary test Flathub build for it?

I'm sorry, I'm afraid I won't have time to do that soon (+ I prefer not to share binaries). We can merge that fix and then you can test afterwards, that's okay too!

kjarosh:
The fix has been released @Tiagoquix

Issue #19303, opened 2025-01-22
Add icon to Wayland window (Linux)


Tiagoquix:
When using Ruffle on Wayland (Flatpak), a generic Wayland icon is shown instead of the Ruffle icon. It spawns the normal Ruffle window and afterwards a generic Wayland one appears (maybe XWayland? IDK).

Window class also doesn't exist for the Wayland window.

kjarosh:
Ruffle does support native Wayland, it shouldn't go through XWayland

PR #19308, opened 2025-01-22
desktop: Associate Ruffle window with Ruffle

kjarosh:
This patch associates the main window created by Ruffle with the Ruffle application by its ID.

This fixes issues on some desktops (KDE, Wayland) where the window is not being recognized as Ruffle.

* Fixes https://github.com/ruffle-rs/ruffle/issues/19303
* Supersedes https://github.com/ruffle-rs/ruffle/pull/19304

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -439,11 +439,17 @@ impl ApplicationHandler<RuffleEvent> for App {
             let preferred_height = self.preferences.cli.height;
             let start_fullscreen = self.preferences.cli.fullscreen;
 
-            let window_attributes = WindowAttributes::default()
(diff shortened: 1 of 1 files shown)


torokati44:
What does this require to be installed? The `.desktop` file?

kjarosh:
> What does this require to be installed? The `.desktop` file?

I believe so, although maybe the icon itself is enough? Who knows, these things differ for various environments, GNOME might do it differently than KDE, etc.

The good thing is that the class is unique, so we won't cause any conflicts + it should be ignored when there's no such application.

torokati44 (approved):
Cool, works for me too, thanks!

PR #19314, opened 2025-01-24
exporter: add a frame cache

fries1234:
This commit adds a frame cache which limits the amount of frames that can be rendered into memory before writing to disk and getting more frames. This can save memory on big 4k frames, especially if you have alot of them as it can fill up 32 gigs of ram pretty fast.

I also split the options structs into its own file.


adrian17 (commented):
Hey, sorry for looking at this with such a delay.

I'm trying to understand exactly what this is doing, and - `cache` is a bit of a bad name, right? It's more of batching control?

To be sure, is there any advantage of storing these together at all, and not saving and throwing away each frame as soon as it's captured? It's possible we shouldn't have been storing frames in the Vec in the first place.

And assuming there is, I think the new logic is kinda confusing. If you call `take_screenshot` several times, won't it skip `skipframes` frames each time?

EDIT: Talking on discord, looks like it really might make sense to just stop storing multiple frames entirely and just do something like this instead (pseudocode):

```rs
for _ in 0..opt.skipframes { run_frame(); }
for i in 0..opt.frames {
    run_frame();
    let frame = take_screenshot();
    write(frame, i, opt.frames);
}
```

fries1234:
> I'm trying to understand exactly what this is doing, and - `cache` is a bit of a bad name, right? It's more of batching control?

Yeah, I'd say it is doing batching stuff. I coded this because I was trying to export 4k frames from a SWF file and I ran out of memory because 4k frames are big and thousands of them is not kind to 32 gigs of ram.


> To be sure, is there any advantage of storing these together at all, and not saving and throwing away each frame as soon as it's captured? It's possible we shouldn't have been storing frames in the Vec in the first place.

I don't know if there's any advantage to doing this. I would say just writing frames to the disk after every screenshot would be fine. Maybe there might be a reason to do a batching thing if IO is a concern?

> And assuming there is, I think the new logic is kinda confusing. If you call `take_screenshot` several times, won't it skip `skipframes` frames each time?

I think the `capture_swf` function does compensate for that when calling the `take_screenshot` function.

PR #19316, opened 2025-01-24
extension: Don't build vanilla WASM module for web extensions

danielhjacobs:
This won't support Safari 16.3- with the Safari extension, but the Safari extension doesn't work anyway, and by the time someone fixes that we can probably just drop Safari extension support for older Safari versions.

The selfhosted and npm packages are now being built as part of the "Build web demo" job. The "Build web demo" job still uses dual-wasm. It is now renamed the "Build web demo and selfhosted" job, and the "Build web" job is now the "Build web extensions" job (though it does also do things like cloning the JS docs still).


danielhjacobs:
Note: I can't really test this properly. I did test that `npm run build:repro` now creates the expected version of Ruffle that fails on Pale Moon, while `npm run build:dual-wasm-repro` works on Pale Moon.

danielhjacobs:
Notes:

May increase minimum Chrome and Edge versions for the extension to Chrome 96, released over 3 years ago.
May increase minimum Firefox version for the extension to Firefox 89, released over 3 and a half years ago.

torokati44:
Okay but can we assume that 16.0, 16.1, and 16.2 are not in use anywhere anymore, because 16.3 is out? 🤔

danielhjacobs:
16.4*

At my last check, almost 5% of ios users are still using 16.0-16.3 for some reason, but unlike major versions, we can tell people to upgrade if they use those versions. Any device that supports 16.0 supports 16.4 too. People just don't upgrade unless they have to do so it seems.

danielhjacobs:
https://github.com/ruffle-rs/ruffle/pull/18528#issuecomment-2465776897

"Percent iOS 16.0-3"

danielhjacobs:
As of December, the market share is down to exactly 4%.

thy486:
> //v8.dev/blog/v8-release-96

This might be relevant [here](https://github.com/ruffle-rs/ruffle/blob/6ba533db6c1cda5043961e0d9b384d4481f2a36d/web/packages/core/src/load-ruffle.ts#L44).

.github/workflows/release_nightly.yml
```diff
@@ -416,12 +417,12 @@ jobs:
...
 
       - name: Package selfhosted
-        if: ${{ !matrix.demo }}
+        if: matrix.demo
         run: zip -r "${{ needs.create-nightly-release.outputs.package_prefix }}-web-selfhosted.zip" .
         working-directory: web/packages/selfhosted/dist
 
       - name: Upload selfhosted
-        if: ${{ !matrix.demo }}
+        if: matrix.demo
```
torokati44, line 425: Why is the negation removed here? 🤔

danielhjacobs: So selhosted continues working on Safari 16.3-. It's also why the workflows long name is changed to "Build web demo and selfhosted"

danielhjacobs: See the second paragraph of https://github.com/ruffle-rs/ruffle/pull/19316#issue-2809964632

torokati44: Then, since it now means "demo and selfhosted", maybe this flag should be replaced with a different flag with inverse values, for "extensions"? :thinking:

torokati44: Oh but it should be clear that this flag governs "whether we're building the browser extensions, or everything else", and not "whether WASM extensions are enabled".

torokati44: Hmm, looking at it again, more closely: Since (almost) _all_ steps of the `build-web` job in the `Release Nightly` workflow are gated by the value of `matrix.demo`, maybe it would make more sense to just split it, getting rid of all the conditionals? :thinking:

torokati44: Sure, a couple of setup steps at the beginning will have to be duplicated this way, but maybe the increased clarity is worth it...?
BTW still waiting for https://github.com/actions/runner/issues/1182 or something similar...

torokati44: Something like this: [`3299304` (#19767)](https://github.com/ruffle-rs/ruffle/pull/19767/commits/3299304031f9db97dab46a5594b2aa99039add35)

danielhjacobs:
For posterity, copying my message from Discord:

> Assuming macOS Safari numbers are similar to iOS version numbers listed on https://gs.statcounter.com/os-version-market-share/ios/mobile-tablet/worldwide, this last month (February 2025), 9.5% of Safari users used versions 14.1-16.3. Assuming the Ruffle extension currently works on Safari 14.1-15.8, after https://github.com/ruffle-rs/ruffle/pull/19664 is merged and works, that is how may Safari users that would otherwise be able to use the extension would be unable to use it after [this PR].

Also note that I think I prefer the #19767 version of this PR.

danielhjacobs:
I'm going to treat this as superseded by #19767. When that is merged, this can be closed. I don't see a point to merging them separately.

torokati44:
I guess this can be closed now then.

Issue #19269, opened 2025-01-18
Allow specifying font name in fontSources config


danielhjacobs:
**Describe the feature**
Provide a clear and concise description of your feature request. Why would this feature be useful? Does the feature involve the desktop build, the web build, or both?

When using a TTF file as a font source, can there be a way to specify the name (and boldness and italicness) of the font for which to use that TTF file? Right now, it's based on the font name or falls back to the URL if no font name exists within the TTF file, but if you didn't create the TTF file and don't have the ability to edit the SWF file the font name in the TTF file may not match the font name requested by the SWF.

https://github.com/ruffle-rs/ruffle/blob/41e373c5aaee802658c0a5f40a9b9fd307902fbb/web/src/builder.rs#L422-L433

danielhjacobs:
Side note: It's pretty useless that it falls back to the as-is URL. It may be at least slightly useful if it fell back to the URL slug sans file extension.

danielhjacobs:
Also note, because fontSources is defined as an array of strings currently, support for this may require letting fontSources be an array of strings or a structure like this and doing different logic depending which it is:

```ts
fontSources: [
    ""fonts/Noto-Sans-Bold.swf": {
        "name": "Noto Sans",
        "bold": true,
        "italics": false
     }
]
```

danielhjacobs:
I fed my current code and desired behavior to AI and the changes look right to my untrained eyes:

Here: https://github.com/ruffle-rs/ruffle/blob/41e373c5aaee802658c0a5f40a9b9fd307902fbb/web/src/builder.rs#L33

Add this:
```rs
#[wasm_bindgen(inspectable)]
#[derive(Debug, Clone)]
pub struct FontSource {
    pub name: Option<String>,
    pub bold: Option<bool>,
    pub italic: Option<bool>,
    pub data: Vec<u8>,
}
```

Here: https://github.com/ruffle-rs/ruffle/blob/41e373c5aaee802658c0a5f40a9b9fd307902fbb/web/src/builder.rs#L63

Change to this:
```rs
    pub(crate) custom_fonts: Vec<(String, FontSource)>,
```

Here: https://github.com/ruffle-rs/ruffle/blob/41e373c5aaee802658c0a5f40a9b9fd307902fbb/web/src/builder.rs#L341-L342

Change to this:
```rs
        for (font_name, font_source) in &self.custom_fonts {
            let bytes_slice = &font_source.data[..];
```

[...]

PR #19321, opened 2025-01-25
web: Allow defining custom names for TTF fontSources

danielhjacobs:
Per https://github.com/ruffle-rs/ruffle/issues/19269#issuecomment-2613619265, draft PR to try to fix that issue. Note, it is mostly human-reviewed AI generated content. For full transparency, the prompt was this:

> I have this current code in web/src/builder.rs:
> https://github.com/ruffle-rs/ruffle/blob/41e373c5aaee802658c0a5f40a9b9fd307902fbb/web/src/builder.rs#L34-L36
> ...
> https://github.com/ruffle-rs/ruffle/blob/41e373c5aaee802658c0a5f40a9b9fd307902fbb/web/src/builder.rs#L63-L67
> ...
> https://github.com/ruffle-rs/ruffle/blob/41e373c5aaee802658c0a5f40a9b9fd307902fbb/web/src/builder.rs#L72
> ...
> https://github.com/ruffle-rs/ruffle/blob/41e373c5aaee802658c0a5f40a9b9fd307902fbb/web/src/builder.rs#L99-L102
> ...
> https://github.com/ruffle-rs/ruffle/blob/41e373c5aaee802658c0a5f40a9b9fd307902fbb/web/src/builder.rs#L339-L360
> ...
> https://github.com/ruffle-rs/ruffle/blob/41e373c5aaee802658c0a5f40a9b9fd307902fbb/web/src/builder.rs#L406-L407
> ...
> https://github.com/ruffle-rs/ruffle/blob/41e373c5aaee802658c0a5f40a9b9fd307902fbb/web/src/builder.rs#L412-L442
> ...
> https://github.com/ruffle-rs/ruffle/blob/41e373c5aaee802658c0a5f40a9b9fd307902fbb/web/src/builder.rs#L687
> ...
> https://github.com/ruffle-rs/ruffle/blob/41e373c5aaee802658c0a5f40a9b9fd307902fbb/web/src/builder.rs#L295-L298
> In a TS file, I have this code that uses that addFont wasm_bindgen function:
> https://github.com/ruffle-rs/ruffle/blob/32f6322efba15ba1d37fbc64ba6cda3b3d0ff7c5/web/packages/core/src/internal/player/inner.tsx#L599-L614
> fontSources is defined in a different file as this:
> https://github.com/ruffle-rs/ruffle/blob/c15c273b75b2e4782e5c480a58cf6a8a6900af96/web/packages/core/src/public/config/load-options.ts#L637
> An example of fontSources is this:
> ```js
> fontSources: ["NotoSans-Bold.ttf", "NotoSans-Italic.ttf"]
> ```
> Adjust the code's logic to allow either this structure with this current logic or this structure:
> ```js
> fontSources: [
>    "fonts/Noto-Sans-Bold.ttf": {
>        "name": "Noto Sans",
>        "bold": true,
>        "italics": false
>     }
> ]
> ```
> With that structure the logic should use the name as the name in register_device_font, the bold boolean in the is_bold of register_device_font, and the italics boolean as the is_italic in register_device_font
> It should also allow not providing any of the three of those (name, bold, or italics), and should fall back to the current values from the the ttf file if they are not provided.

### DIFF
--- a/web/packages/core/src/internal/player/inner.tsx
+++ b/web/packages/core/src/internal/player/inner.tsx
@@ -615,17 +615,19 @@ export class InnerPlayer {
                     }
                 } else if (typeof fontSource === "object") {
                     // Handle new format (object with additional properties)
-                    try {
-                        const response = await fetch(key);
-                        builder.addFont(
-                            key,
-                            typeof fontSource['name'] === "string" ? fontSource['name'] : null,
-                            typeof fontSource['bold'] === "boolean" ? fontSource['bold'] : null,
-                            typeof fontSource['italics'] === "boolean" ? fontSource['italics'] : null,
-                            new Uint8Array(await response.arrayBuffer()),
-                        );
-                    } catch (error) {
-                        console.warn(`Couldn't download font source from ${key}`, error);
+                    for (const [url, fontInfo] of Object.entries(fontSource)) {
+                        try {
+                            const response = await fetch(url);
+                            builder.addFont(
+                                url,
+                                fontInfo?.['name'] || null,
+                                fontInfo?.['bold'] ?? null,
+                                fontInfo?.['italics'] ?? null,
+                                new Uint8Array(await response.arrayBuffer()),
+                            );
+                        } catch (error) {
+                            console.warn(`Couldn't download font source from ${key}`, error);
+                        }
                     }
                 }
             }
--- a/web/packages/core/src/internal/player/inner.tsx
+++ b/web/packages/core/src/internal/player/inner.tsx
@@ -597,37 +597,55 @@ export class InnerPlayer {
         builder.setVolume(this.volumeSettings.get_volume());
 
         if (this.loadedConfig?.fontSources) {
-            for (const key in this.loadedConfig.fontSources) {
-                const fontSource = this.loadedConfig.fontSources[key];
-                if (typeof fontSource === "string") {
-                    // Handle old format (array of strings)
-                    try {
-                        const response = await fetch(fontSource);
-                        builder.addFont(
-                            fontSource,
-                            null, // No custom name
-                            null, // No bold override
-                            null, // No italic override
-                            new Uint8Array(await response.arrayBuffer()),
-                        );
-                    } catch (error) {
-                        console.warn(`Couldn't download font source from ${fontSource}`, error);
-                    }
-                } else if (typeof fontSource === "object") {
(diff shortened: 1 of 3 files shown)


danielhjacobs:
Allowed `fontSources` config options:

```ts
    "fontSources": [
        "fonts/Times New Roman.ttf",
        {
            "fonts/centurygothic.ttf": {
                name: "Century Gothic",
            },
            "fonts/arial.ttf": {},
        }
    ]
```ts
    "fontSources": [
        {
            "fonts/centurygothic.ttf": {
                name: "Century Gothic",
            }
        },
        {
            "fonts/arial.ttf": {},
        }
    ]
```
```ts
    "fontSources": ["fonts/centurygothic.ttf", "fonts/arial.ttf"]
```
```ts
    "fontSources": {
        "fonts/centurygothic.ttf": {
            name: "Century Gothic",
        },
        "fonts/arial.ttf": {},
    }
```

evilpie:
This really seems like too many different ways of defining fonts. We should at most have two, keeping the old one for compatibility of course.

danielhjacobs:
1) Re: https://github.com/ruffle-rs/ruffle/pull/19321#issuecomment-2614186680, the 1st and 2nd options are technically identical, that said I think I'd prefer to keep the 3rd and 4th instead of the 1st-3rd. That will mean when using some fonts that need their name/bold/italics values overridden and others that don't, you'll need an empty object value for the ones that don't.
2) TTC files and SWF files can define multiple fonts per file. In that case, the current config idea can't really work. I can avoid using the name/bold/italics values from the config if the number of fonts in the TTF collection is greater than 1 or if using an SWF font source, but that doesn't solve the issue, it just skips dealing with it when it doesn't make sense.
3) Instead, we can leave the current `fontSources` config option as it is now and add a new config option, `fontMapping`, a `Map` that has a key of the name, bold, and italics values requested by the SWF TextField and a value of the name, bold, and italics value that was registered by the `fontSources`. The layout code would then need adjustments to check the config's `fontMapping`, and if the font name, bold value, and italics value the SWF is requesting in that part of the layout code is a key in the `fontMapping` config, it should load the device font matching the name, bold, and italics value from that part of the `fontMapping` config instead. If not, it should load the device font it does not (the key values that we looked for).

danielhjacobs:
4. It would probably be better if the font fell back to the URL slug sans file extension if the TTF file has no full name, instead of the as-is URL.

PR #19329, opened 2025-01-25
avm2: Replace complex ops with simpler ops

Lord-McSweeney:
This should shrink `Activation::run_actions`

This removes the ops `getglobalscope`, `getglobalslot`, `callsupervoid`, `pushnan`, and `pushbyte`- instead, we now use `getouterscope`/`getscopeobject`, `getouterscope`/`getscopeobject`+`getslot`, `callsuper`+`pop`, `pushdouble`, and `pushshort`.

### DIFF
--- a/core/src/avm2/verify.rs
+++ b/core/src/avm2/verify.rs
@@ -326,6 +326,20 @@ pub fn verify_method<'gc>(
                     }
                 }
 
+                AbcOp::GetGlobalSlot { index } => {
+                    // Split this `GetGlobalSlot` into a `GetGlobalScope` and a `GetSlot`;
+                    // this is possible because a `GetGlobalSlot` is guaranteed
+                    // to take up at least 2 bytes.
+
+                    // See the comment on GetLex for more information.
+
+                    assert!(bytes_read > 1);
+                    byte_info[previous_position as usize] =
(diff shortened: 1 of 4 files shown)


core/src/avm2/verify.rs
```diff
@@ -326,6 +326,34 @@ pub fn verify_method<'gc>(
...
+                    // this is possible because a `CallSuperVoid` is guaranteed
+                    // to take up at least 2 bytes.
+
+                    // See the comment on GetLex for more information.
+
+                    assert!(bytes_read > 1);
+                    byte_info[previous_position as usize] =
+                        ByteInfo::OpStart(AbcOp::CallSuper { index, num_args });
+                    byte_info[(previous_position + 1) as usize] =
+                        ByteInfo::OpStartNonJumpable(AbcOp::Pop);
```
adrian17, line 340: FWIW, since we already have `push_return_value` for CallMethod, IMO it makes sense to add it here too, if only for consistency. But up to you.

adrian17 (approved):

core/src/avm2/verify.rs
Lord-McSweeney: I only added `push_return_value` to CallMethod because that's the only way to optimize `setproperty` on caller- we could also add here and to CallProperty, but I don't think it's worth adding the complexity.

PR #19349, opened 2025-01-26
desktop: Set LoadingState to Loaded even when window is maximized

### LINKED ISSUE #19343: [Linux] "Open with Ruffle" does not work
### Describe the bug

Double-clicking a SWF file or right-clicking and using "Open with Ruffle" on Linux, KDE Plasma, Wayland, using the Flatpak distribution, does not work.

If I open a SWF file in this way, Ruffle only gives me a white screen and becomes a bit unresponsive. Closing the SWF (not the Ruffle window) and opening another SWF also does not work. It only works when I restart Ruffle and open a SWF using the "Open..." submenu in the app (or the "Recents" submenu).

My SWF files are stored in an external drive, and the Flatpak app has read+write access to the entire folder where the SWFs are located.

It works normally when I use the app's built-in option, but the desktop integration doesn't.

I tried changing the `Exec=` line of the `.desktop` file to `%U` instead of `%u`, but it didn't work. (https://specifications.freedesktop.org/desktop-entry-spec/latest/exec-variables.html)

Since this happens with any SWF, I think I don't need to provide a specific SWF for testing, but let me know if you need one.

### Expected behavior

N/A

### Content Location

N/A

### Affected platform

Desktop app

### Operating system

Linux

### Browser

N/A

### Additional information

N/A

sleepycatcoding:
Should fix #19343

This bug did not affect opening files after application startup since LoadingState is never set back to Loading again.

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -257,7 +257,7 @@ impl MainWindow {
         // do not resize while window is maximized.
         let should_resize = !self.gui.window().is_maximized();
 
-        let viewport_size = if should_resize {
+        let (viewport_size, state) = if should_resize {
             let movie_width = swf_header.stage_size().width().to_pixels();
             let movie_height = swf_header.stage_size().height().to_pixels();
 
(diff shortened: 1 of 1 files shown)


desktop/src/app.rs
```diff
@@ -320,6 +320,7 @@ impl MainWindow {
 
             viewport_size
         } else {
+            self.loaded = LoadingState::Loaded;
```
kjarosh, line 323: Maybe it would be better if this `if` returned not only viewport size, but also new loading state? We could update it in one line below the `if`

sleepycatcoding: Changed.

kjarosh (approved):
LGTM, thanks!

PR #19371, opened 2025-01-29
core,avm1,avm2: Add and use istr macro, remove most static strings

Lord-McSweeney:
This PR introduces a macro `istr!` that retrieves a string from the common string store (`CommonStrings`). This helps a lot to reduce static string usage everywhere, especially in AVM1. Currently, this PR removes about half of all static strings we use- removing more will require some additional changes, which I'm leaving for future PRs.


core/src/string/common.rs
```diff
@@ -1,28 +1,48 @@
...
+use super::{AvmString, AvmStringInterner, AvmStringRepr, WStr};
 use gc_arena::{Gc, Mutation};
 
+#[allow(non_snake_case)]
 pub struct CommonStrings<'gc> {
-    pub str_x: AvmAtom<'gc>,
-    pub str_y: AvmAtom<'gc>,
-    pub str_width: AvmAtom<'gc>,
-    pub str_height: AvmAtom<'gc>,
+    pub str_false: AvmString<'gc>,
```
adrian17, line 51: can't these stay being AvmAtoms? It definitely feels more "right", plus lighter.

(I'm also assuming that eventually apis like `get_public_property` will switch to requiring `AvmAtom`s?)

Lord-McSweeney: This was my initial approach too, but then we end up needing to add a lot more `.into()` for places that require an `AvmString`- and AvmString is going to shrink to 8 bytes very soon anyway.

core/src/string/interner.rs
```diff
@@ -52,6 +52,9 @@ impl AvmAtom<'_> {
 pub struct AvmStringInterner<'gc> {
     interned: WeakSet<'gc, AvmStringRepr<'gc>>,
 
+    /// Strings used across both AVMs and in core code.
+    pub common: Gc<'gc, CommonStrings<'gc>>,
```
adrian17, line 56: Couldn't this stay a `CommonStrings`, without a Gc (and have `common()` return the reference)?
This would mean one less pointer hop.

I just tried making this change and it almost went perfectly smoothly, in `op_type_of` to satisfy the borrow checker I needed to access them directly, without storing `common_strings` in a variable. (alternatively, we can add `self.common_strings()` that returns a non-mut reference, and same with `self.avm2_nomut()` - since we don't actually need a `&mut` anywhere there.)

adrian17 (approved):

PR #19385, opened 2025-01-30
avm2: Port String and Array to Actionscript

Lord-McSweeney:


core/src/avm2/globals/string.rs
```diff
@@ -158,17 +149,13 @@ pub fn last_index_of<'gc>(
...
     let this = this.coerce_to_string(activation)?;
 
-    let pattern = match args.get(0) {
-        None => return Ok(Value::Undefined),
-        Some(s) => s.clone().coerce_to_string(activation)?,
-    };
+    let pattern = args.get_string(activation, 0)?;
 
-    let start_index = match args.get(1) {
-        None | Some(Value::Undefined) => this.len(),
```
adrian17, line 238: to be sure, this one was just plain wrong?
(or happened to give the same end result?)

core/src/avm2/globals/Array.as
```diff
@@ -1,5 +1,207 @@
...
+        public static const DESCENDING:uint = 2;
+
+        public static const UNIQUESORT:uint = 4;
+
+        public static const RETURNINDEXEDARRAY:uint = 8;
+
+        public static const NUMERIC:uint = 16;
+
+        // Some of these functions work on all Array-like objects in avmplus
+        // (i.e. Vector, ByteArray), but we only support calling them on Arrays
```
adrian17, line 16: as in, it's a bug? If so, can you make it clearer (`are supposed to work (...) but currently we only support`) and mark it as a TODO?

core/src/avm2/globals/flash/xml/XMLNode.as
```diff
@@ -70,7 +70,10 @@ package flash.xml
                 }
                 var index = parentNode.childNodes.indexOf(this);
                 if (index > -1) {
-                    parentNode.childNodes.removeAt(index);
+                    // For some reason ASC doesn't add the AS3 namespace here;
+                    // we need to add it manually since removeAt is only defined
+                    // in the AS3 namespace
+                    parentNode.childNodes.AS3::removeAt(index);
```
adrian17, line 76: to be sure, does this still generate a `callproperty` rather than a runtime lookup of `AS3`?

adrian17 (approved):

core/src/avm2/globals/flash/xml/XMLNode.as
Lord-McSweeney: Seems like it does- the check Aaron added in `build_playerglobal` would've caught it, but I checked FFDEC to make sure and it's still a `callproperty`

core/src/avm2/globals/string.rs
Lord-McSweeney: The implementation before this PR was wrong, and now the AS bindings will coerce to Number anyway so an `undefined` appearing here is impossible

PR #19389, opened 2025-01-31
avm1: Implement TextFormat.display

kjarosh:
This PR implements `TextFormat.display` and adds tests for it.


core/src/avm1/globals/text_format.rs
```diff
@@ -425,17 +424,36 @@ fn set_bullet<'gc>(
...
+    value: &Value<'gc>,
 ) -> Result<(), Error<'gc>> {
-    avm1_stub!(activation, "TextFormat", "display");
+    if matches!(value, Value::Undefined | Value::Null) {
+        text_format.display = Some(crate::html::TextDisplay::Block);
+        return Ok(());
+    }
+
+    let value = value.coerce_to_string(activation)?;
+    let display = if value.eq(WStr::from_units(b"inline")) {
```
adrian17, line 449: this could be a `&value == b"inline"`, I think?

adrian17 (approved):

core/src/avm1/globals/text_format.rs
kjarosh: you're right, thanks!

PR #19393, opened 2025-01-31
avm2: Port Vector classes to AS

Lord-McSweeney:
This unblocks porting Number, uint, and int to AS


core/build_playerglobal/src/lib.rs
```diff
@@ -279,6 +284,13 @@ fn rust_method_name_and_path(
     let name = resolve_multiname_name(abc, trait_name).to_string();
 
     path += &flash_to_rust_string(&name, false, "::");
+
+    match &trait_.kind {
+        TraitKind::Getter { .. } => flash_method_path += "get ",
```
adrian17, line 289: To be sure, what was this added for?

adrian17 (approved):

core/build_playerglobal/src/lib.rs
Lord-McSweeney: The changes around naming were to have correct method names in stack traces. Without the two changes, the DisplayObject blendMode setter would display as `flash.display::DisplayObject/flash::display::DisplayObject::blendMode()`. Now it displays as `flash.display::DisplayObject/set blendMode()`, matching FP.

PR #19402, opened 2025-02-01
desktop: Make the main window at minimum wider than the menu bar

### LINKED ISSUE #16700: Linux: Main application window should have a minimum size
### Describe the bug

If you move the mouse cursor on one of the 4 corners and click + drag to resize the window, it's possible to reduce the application to only a small strip, even hiding the buttons for window control.

[Screencast from 2024-06-13 05-11-11.webm](https://github.com/ruffle-rs/ruffle/assets/4840355/4b12210b-50c2-4029-aff2-2dea9c0d4de1)

Screenshot from 2024-06-13 05-12-25

### Expected behavior

I think it would be better to have a fixed minimum width and height for the window, so at least the full titlebar and some part of the application itself remains visible. I think at least every modern GTK / QT application does this nowadays, and it's more professional looking.

Here's for example how the smallest possible window for WPS Office looks like. It's not possible to reduce it further.

339048712-3fed8bea-f155-4188-9775-a3334b4169ac

### Content Location

not relevant

### Affected platform

Desktop app

### Operating system

Arch Linux, Gnome 46 on Wayland

### Browser

_No response_

### Additional information

Tested on latest Nightly: ruffle-nightly-2024_06_13-linux-x86_64.tar.gz

evilpie:
Fixes #16700

This is how the main window looks at its smallest size:
image

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -436,7 +436,12 @@ impl ApplicationHandler<RuffleEvent> for App {
                 Icon::from_rgba(icon_bytes.to_vec(), 32, 32).expect("App icon should be correct");
 
             let no_gui = self.preferences.cli.no_gui;
-            let min_window_size = (16, if no_gui { 16 } else { MENU_HEIGHT + 16 }).into();
+            let min_window_size = if no_gui {
+                (16, 16)
+            } else {
+                (350, MENU_HEIGHT + 16)
+            }
+            .into();
             let preferred_width = self.preferences.cli.width;
             let preferred_height = self.preferences.cli.height;
             let start_fullscreen = self.preferences.cli.fullscreen;


adrian17 (approved):

torokati44:
In at least a couple places (I guess where it's more obviously noticeable), `MENU_HEIGHT` is multiplied by the window `scale_factor`. Maybe it should be everywhere, and similarly, this width too?

evilpie:
> In at least a couple places (I guess where it's more obviously noticeable), `MENU_HEIGHT` is multiplied by the window `scale_factor`. Maybe it should be everywhere, and similarly, this width too?

How would I test this?

torokati44:
I suppose you could try setting the `WINIT_X11_SCALE_FACTOR` environment variable to `2` or something like that.

See: https://docs.rs/winit-gtk/latest/winit/dpi/index.html#how-is-the-scale-factor-calculated

evilpie:
It seems like the window size takes the scale factor into account automatically and we don't need to do anything else.

Minimum window size with `WINIT_X11_SCALE_FACTOR=2`:
image

PR #19423, opened 2025-02-03
desktop: Fix busy loop in winit

kjarosh:
This patch simplifies the code for setting control flow by always requesting a wait till the next frame time.

The busy loop was caused by the fact that winit was leaving wait requests with timestamps in the past, which caused us to not update the control flow as a wait was requested earlier.

Thanks @madsmtm for helping us with this fix.

Testing shows that we're now producing a reasonable number of ticks per second, and idle CPU usage has dropped significantly.


n0samu (approved):

adrian17 (approved):
Let's try it.

PR #19443, opened 2025-02-05
core: Use a declarative macro to define the list of common strings

moulins:
An alternate version using `unsafe` to work with arrays instead of field-by-field would save ~12kB on the final (stripped) executable, but would increase code complexity:

```rs
macro_rules! define_common_strings {
    (@count_one $tt:tt) => { 1 };
    ($($field:ident: $str:literal,)*) => {
        #[allow(non_snake_case)]
        #[repr(C)]
        pub struct CommonStrings<'gc> {
            $(pub $field: AvmString<'gc>,)*
        }

        const COMMON_STRINGS_LEN: usize = 0 $(+ define_common_strings!(@count_one $field))*;
        static COMMON_STRINGS: [&[u8]; COMMON_STRINGS_LEN] = [$($str),*];
    };
}

impl<'gc> CommonStrings<'gc> {
    pub fn new(mut intern_from_static: impl FnMut(&'static [u8]) -> AvmString<'gc>) -> Self {
        let strings = COMMON_STRINGS.map(intern_from_static);
        // SAFETY: `Self` is `repr(C)`, and contains only `AvmString` fields, so the layouts are compatible.
        unsafe { std::mem::transmute(strings) }
     }
 }


unsafe impl<'gc> Collect for CommonStrings<'gc> {
    fn trace(&self, cc: &gc_arena::Collection) {
        // SAFETY: `Self` is `repr(C)`, and contains only `AvmString` fields, so the layouts are compatible.
        let array: &[AvmString<'gc>; COMMON_STRINGS_LEN] = unsafe { std::mem::transmute(self) };
        array.trace(cc)
    }
}

define_common_strings! {
    str___constructor__: b"__constructor__",
    str___proto__: b"__proto__",
    // etc
}
```


core/src/string/common.rs
```diff
@@ -1,364 +1,191 @@
...
-    pub str_xMax: AvmString<'gc>,
-    pub str_xMin: AvmString<'gc>,
-    pub str_xml: AvmString<'gc>,
-    pub str_y: AvmString<'gc>,
-    pub str_yMax: AvmString<'gc>,
-    pub str_yMin: AvmString<'gc>,
-    pub str_zoom: AvmString<'gc>,
-}
+macro_rules! define_common_strings {
+    (@count_one $tt:tt) => { 1 };
```
Lord-McSweeney, line 5: This seems unused

Lord-McSweeney (approved):

PR #19445, opened 2025-02-05
video/external: Reorder frames decoded by OpenH264 manually

torokati44:
This should unblock https://github.com/ruffle-rs/ruffle/pull/18581, despite https://github.com/cisco/openh264/issues/3837.

### DIFF
--- a/core/src/display_object/video.rs
+++ b/core/src/display_object/video.rs
@@ -312,6 +312,7 @@ impl<'gc> Video<'gc> {
                         codec: streamdef.codec,
                         data: &read.movie.data()[*slice_start..*slice_end],
                         frame_id,
+                        composition_time: None,
                     };
                     context
                         .video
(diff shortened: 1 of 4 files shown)


torokati44:
Hm, I wonder if instead we should use the composition time of the decoded frames (assuming OpenH264 carries them over correctly from the encoded ones, but _surely it does_), to govern the "lead time" with which the FLV streaming feeds the decoder... 🤔
Then we could even better match FP with which frame is presented when, getting the delay lower than it currently has to be.

torokati44:
... and/or separate the "here's an encoded chunk" and "gimme a decoded frame" calls, like they are in FFmpeg (libavcodec), so they don't have to be 1:1 with each other...

torokati44:
Note to self: if we go forward with this, then TODO clear the frame queue on seek (especially on backward ones), decide whether it's signalled explicitly from the outside or detected based on `frame_id`s passed in, and see how those work with FLV sources

PR #19456, opened 2025-02-06
desktop: Call `window.pre_present_notify()` before presenting

kjarosh:
This patch fixes a weird issue on Wayland, which caused Ruffle to stop ticking when the window wasn't visible.  When using `Fifo` present mode, Wayland winded down FPS to 0 when window wasn't visible, and it caused present() method to block indefinitely.

Calling `window.pre_present_notify()` before `present()` fixes this issue.


Dinnerbone (approved):

kjarosh:
BTW This is probably the most expensive code I've ever seen. It required a lot of engineering to produce. :eyes:

Issue #18952, opened 2024-12-13
Error on https://uunnblockedgames.weebly.com/drag-racer-v3.html


Pwhippy:
# Error Info
Error name: Error
Error message: panicked at core/src/avm1/runtime.rs:449:41:
called `Option::unwrap()` on a `None` value
Error stack:
```
Error: panicked at core/src/avm1/runtime.rs:449:41:
called `Option::unwrap()` on a `None` value
    at n.wbg.__wbg_new_4d81617a04bc1b5b (https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.98308ff2359089abacc0.js:1:57779)
    at ruffle_web.wasm.js_sys::Error::new::h5021eff635514cb9 (wasm://wasm/ruffle_web.wasm-031ae2ca:wasm-function[12844]:0x9898fe)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once{{vtable.shim}}::h192da10450332470 (wasm://wasm/ruffle_web.wasm-031ae2ca:wasm-function[9764]:0x9065f1)
    at ruffle_web.wasm.std::panicking::rust_panic_with_hook::h33fe77d38d305ca3 (wasm://wasm/ruffle_web.wasm-031ae2ca:wasm-function[6410]:0x82d516)
    at ruffle_web.wasm.core::panicking::panic_fmt::hde8b7aa66e2831e1 (wasm://wasm/ruffle_web.wasm-031ae2ca:wasm-function[9476]:0x8ff690)
    at ruffle_web.wasm.core::panicking::panic::hcaca2598a27ec0fc (wasm://wasm/ruffle_web.wasm-031ae2ca:wasm-function[9942]:0x917def)
    at ruffle_web.wasm.core::option::unwrap_failed::h9aa82eb71128b127 (wasm://wasm/ruffle_web.wasm-031ae2ca:wasm-function[12329]:0x9664e9)
    at ruffle_web.wasm.ruffle_core::player::Player::tick::hc0665d68e38eef42 (wasm://wasm/ruffle_web.wasm-031ae2ca:wasm-function[900]:0x1afc2a)
    at ruffle_web.wasm.ruffle_web::RuffleHandle::tick::h2d3c3b01b979a0f6 (wasm://wasm/ruffle_web.wasm-031ae2ca:wasm-function[973]:0x25a2b0)
[...]

Issue #17072, opened 2024-07-10
Error on http://3.145.15.34/sa.php


ghost:
# Error Info
Error name: Error
Error message: panicked at core/src/avm1/runtime.rs:449:41:
called `Option::unwrap()` on a `None` value
Error stack:
```
Error: panicked at core/src/avm1/runtime.rs:449:41:
called `Option::unwrap()` on a `None` value
    at imports.wbg.__wbg_new_28c511d9baebfa89 (chrome-extension://donbcfbmhbcapadipfkeojnmajbakjdc/dist/69.js:3701:21)
    at ruffle_web.wasm.js_sys::Error::new::hda2cb16801956d39 (wasm://wasm/ruffle_web.wasm-0336c5d6:wasm-function[12819]:0x9df7cd)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once{{vtable.shim}}::h4efee46b7e0e8825 (wasm://wasm/ruffle_web.wasm-0336c5d6:wasm-function[9849]:0x976437)
    at ruffle_web.wasm.std::panicking::rust_panic_with_hook::h6731baa78621a747 (wasm://wasm/ruffle_web.wasm-0336c5d6:wasm-function[6498]:0x869085)
    at ruffle_web.wasm.core::panicking::panic_fmt::h5c7ce52813e94bcd (wasm://wasm/ruffle_web.wasm-0336c5d6:wasm-function[10249]:0x9a06e8)
    at ruffle_web.wasm.core::panicking::panic::h3b99507adb69386a (wasm://wasm/ruffle_web.wasm-0336c5d6:wasm-function[9991]:0x98c2ec)
    at ruffle_web.wasm.core::option::unwrap_failed::h0d060778de0f79e8 (wasm://wasm/ruffle_web.wasm-0336c5d6:wasm-function[12328]:0x9d0a17)
    at ruffle_web.wasm.ruffle_core::player::Player::tick::he18c88e31472f0a8 (wasm://wasm/ruffle_web.wasm-0336c5d6:wasm-function[884]:0x1dcb4c)
    at ruffle_web.wasm.ruffle_web::RuffleHandle::tick::hd63fea494347c202 (wasm://wasm/ruffle_web.wasm-0336c5d6:wasm-function[1082]:0x302d29)
[...]

Issue #16733, opened 2024-06-15
Error on https://flasharch.com/en/archive/play/0dc044a8fa9947f8fdbff690a781926d


a1fadi:
# Error Info
Error name: Error
Error message: panicked at core/src/avm1/runtime.rs:449:41:
called `Option::unwrap()` on a `None` value
Error stack:
```
Error: panicked at core/src/avm1/runtime.rs:449:41:
called `Option::unwrap()` on a `None` value
    at n.wbg.__wbg_new_28c511d9baebfa89 (https://cdn.jsdelivr.net/npm/@ruffle-rs/ruffle@0.1.0-nightly.2024.6.11/core.ruffle.b9e046593f0b0edf986f.js:1:79124)
    at ruffle_web.wasm.js_sys::Error::new::ha59a397dfb01052a (wasm://wasm/ruffle_web.wasm-0339d84a:wasm-function[12836]:0x9fbf3d)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once{{vtable.shim}}::hfdf01bfa88fb72ec (wasm://wasm/ruffle_web.wasm-0339d84a:wasm-function[9814]:0x98fabf)
    at ruffle_web.wasm.std::panicking::rust_panic_with_hook::h32c80a64fe4de396 (wasm://wasm/ruffle_web.wasm-0339d84a:wasm-function[6536]:0x88a909)
    at ruffle_web.wasm.core::panicking::panic_fmt::hc7427f902a13f1a9 (wasm://wasm/ruffle_web.wasm-0339d84a:wasm-function[10230]:0x9ba9e8)
    at ruffle_web.wasm.core::panicking::panic::hb157b525de3fe68d (wasm://wasm/ruffle_web.wasm-0339d84a:wasm-function[9962]:0x9a57d2)
    at ruffle_web.wasm.core::option::unwrap_failed::hb5bacfb0dd292085 (wasm://wasm/ruffle_web.wasm-0339d84a:wasm-function[12349]:0x9ed3c5)
    at ruffle_web.wasm.ruffle_core::player::Player::tick::h3d38ab5bc07c3641 (wasm://wasm/ruffle_web.wasm-0339d84a:wasm-function[876]:0x1d3ef1)
    at ruffle_web.wasm.ruffle_web::RuffleHandle::tick::hde120e3c93c66e11 (wasm://wasm/ruffle_web.wasm-0339d84a:wasm-function[1108]:0x336283)
[...]

Issue #16430, opened 2024-05-23
Error on http://www.gaiaonline.com/launch/fishing


cindytrieu:
# Error Info
Error name: Error
Error message: panicked at core/src/avm1/runtime.rs:449:41:
called `Option::unwrap()` on a `None` value
Error stack:
```
Error: panicked at core/src/avm1/runtime.rs:449:41:
called `Option::unwrap()` on a `None` value
    at J.n.wbg.__wbg_new_28c511d9baebfa89 (http://graphics.gaiaonline.com/ruffle/20240522/core.ruffle.d30afc4a2ae40b2eeb92.js:1:74791)
    at ruffle_web.wasm.js_sys::Error::new::h69c04334b4f73f11 (wasm://wasm/ruffle_web.wasm-03386dba:wasm-function[12753]:0x9f8bc8)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once{{vtable.shim}}::h939f3a1887cd9d5d (wasm://wasm/ruffle_web.wasm-03386dba:wasm-function[10512]:0x9b2660)
    at ruffle_web.wasm.std::panicking::rust_panic_with_hook::h32c80a64fe4de396 (wasm://wasm/ruffle_web.wasm-03386dba:wasm-function[6527]:0x877c5c)
    at ruffle_web.wasm.core::panicking::panic_fmt::hc7427f902a13f1a9 (wasm://wasm/ruffle_web.wasm-03386dba:wasm-function[10261]:0x9ab2f2)
    at ruffle_web.wasm.core::panicking::panic::hb157b525de3fe68d (wasm://wasm/ruffle_web.wasm-03386dba:wasm-function[9994]:0x993e9e)
    at ruffle_web.wasm.core::option::unwrap_failed::hb5bacfb0dd292085 (wasm://wasm/ruffle_web.wasm-03386dba:wasm-function[12256]:0x9ea1a2)
    at ruffle_web.wasm.ruffle_core::player::Player::tick::h9ad15f83953e88ac (wasm://wasm/ruffle_web.wasm-03386dba:wasm-function[897]:0x1dde60)
    at ruffle_web.wasm.ruffle_web::Ruffle::tick::ha1f22fdfc9b00e09 (wasm://wasm/ruffle_web.wasm-03386dba:wasm-function[1120]:0x2ff54c)
[...]

PR #19462, opened 2025-02-07
avm1: Do not unwrap parent when removing objects pending removal

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/18952
* Fixes https://github.com/ruffle-rs/ruffle/issues/17191
* Fixes https://github.com/ruffle-rs/ruffle/issues/17072
* Fixes https://github.com/ruffle-rs/ruffle/issues/16733
* Fixes https://github.com/ruffle-rs/ruffle/issues/16430
* Fixes https://github.com/ruffle-rs/ruffle/issues/16129

Turns out that sometimes objects pending removal do not have a parent.

I wasn't able to reproduce it, but it happens as we can see in error reports.

Semantically this should be fine, as we mainly operate on the parent here, but it's hard to be sure considering it's not reproducible. It's better than panicking for sure.

### DIFF
--- a/core/src/avm1/runtime.rs
+++ b/core/src/avm1/runtime.rs
@@ -444,18 +444,20 @@ impl<'gc> Avm1<'gc> {
             Self::find_display_objects_pending_removal(root_clip, &mut out);
         }
 
-        for child in out {
+        for &child in &out {
             // Get the parent of this object
-            let parent = child.parent().unwrap();
-            let parent_container = parent.as_container().unwrap();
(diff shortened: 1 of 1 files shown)


torokati44:
Is it possible that parent and child are set as "pending removal" at once, and the parent just happened to get removed first, in the same loop? 🤔 Just a blind guess...

kjarosh:
> Is it possible that parent and child are set as "pending removal" at once, and the parent just happened to get removed first, in the same loop? 🤔 Just a blind guess...

That's one of the scenarios I considered, but I failed to reproduce it :/

PR #19465, opened 2025-02-07
Move the empty and one-char strings to `CommonStrings`, and allow `istr!` to access them.

moulins:
Not entirely sure this is worth the extra complexity in macro code, but this does have a few nice consequences:
- it removes some "duplicates" strings like `"x"` or `"y"`, that were both in the one-char cache and in `CommonStrings`;
- being able to use `istr!` in more places makes code more readable, even if not that much shorter.

### DIFF
--- a/core/macros/src/lib.rs
+++ b/core/macros/src/lib.rs
@@ -192,6 +192,10 @@ pub fn enum_trait_object(args: TokenStream, item: TokenStream) -> TokenStream {
 /// istr!(context, "description");
 /// // expands to:
 /// HasStringContext::strings_ref(context).str_description;
+///
+/// istr!("A");
+/// // expands to:
+/// activation.context.strings.common().ascii_chars[65 /* 'A' */];
 /// ```
 #[proc_macro]
 pub fn atom(item: TokenStream) -> TokenStream {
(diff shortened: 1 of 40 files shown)


core/src/string/interner.rs
```diff
@@ -137,11 +105,11 @@ impl<'gc> AvmStringInterner<'gc> {
 
         // It's assumed that start<=end. This is tested later via a range check.
         if start_index == end_index {
-            return self.empty.into();
+            return self.common.str_.into();
```
Lord-McSweeney, line 108: Could you add a comment clarifying that this is the empty string?

core/macros/src/lib.rs
```diff
@@ -192,6 +192,14 @@ pub fn enum_trait_object(args: TokenStream, item: TokenStream) -> TokenStream {
...
 /// // expands to:
 /// HasStringContext::strings_ref(context).str_description;
+///
+/// istr!(context, "");
+/// // expands to:
+/// activation.context.strings.common().empty;
+///
+/// istr!(context, "A");
+/// // expands to:
+/// activation.context.strings.common().ascii_chars[65 /* 'A' */];
```
Lord-McSweeney, line 202: I think these comments should be using `HasStringContext::strings_ref`?

Lord-McSweeney (approved):

PR #19480, opened 2025-02-09
avm1: Don't use a static `AvmString` for AVM1 array's length property

moulins:
`ArrayObject` construction now requires an activation, so add `ArrayObjectBuilder` to simplify the process: it allows grabbing what we need from the activation (the interned `length` property name, and the array prototype) without running afoul of the borrow checker.

### DIFF
--- a/core/src/avm1/object/array_object.rs
+++ b/core/src/avm1/object/array_object.rs
@@ -1,8 +1,9 @@
 use crate::avm1::property::Attribute;
 use crate::avm1::{Activation, Error, Object, ObjectPtr, ScriptObject, TObject, Value};
 use crate::ecma_conversions::f64_to_wrapping_i32;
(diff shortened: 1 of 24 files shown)


Lord-McSweeney (approved):

core/src/avm1/object/array_object.rs
```diff
@@ -85,7 +92,7 @@ impl<'gc> TObject<'gc> for ArrayObject<'gc> {
         activation: &mut Activation<'_, 'gc>,
         this: Object<'gc>,
     ) -> Result<(), Error<'gc>> {
-        if &name == b"length" {
+        if name == istr!("length") {
```
Lord-McSweeney, line 95: This can remain `&name == b"length"`

torokati44:
Labels, please! 🥺

Issue #19278, opened 2025-01-19
No compatible graphics backends of any kind were available


kjarosh:
### Describe the bug

We've recently been getting error reports with the following error:

```
panicked at desktop\src\app.rs:466:14:
GUI controller should be created: No compatible graphics backends of any kind were available
```

The issues started flooding in on Nov 19th 2024, it seems that this problem is Windows-specific.

* https://github.com/ruffle-rs/ruffle/issues/18672
* https://github.com/ruffle-rs/ruffle/issues/18828
* https://github.com/ruffle-rs/ruffle/issues/18866
* https://github.com/ruffle-rs/ruffle/issues/18874
* https://github.com/ruffle-rs/ruffle/issues/18912
* https://github.com/ruffle-rs/ruffle/issues/18915
* https://github.com/ruffle-rs/ruffle/issues/19065
* https://github.com/ruffle-rs/ruffle/issues/19270
* https://github.com/ruffle-rs/ruffle/issues/19273

### Expected behavior

Ruffle does not throw the error

### Content Location

N/A

### Affected platform

Desktop app

### Operating system

Windows

adrian17:
> The issues started flooding in on Nov 19th 2024

Note: the first issue linked mentions 2024-11-17, so the regression might be earlier

> Ruffle Version
> 0.1.0-nightly (https://github.com/ruffle-rs/ruffle/commit/b7bc3d57c950256fea96d912ce8a1b03a68dd0e5 2024-11-17)

adrian17:
We bumped wgpu in 11-15: https://github.com/ruffle-rs/ruffle/releases/tag/nightly-2024-11-15

evilpie:
https://github.com/ruffle-rs/ruffle/issues/19494

adrian17:
Should be fixed with #19512 / #19714.

PR #19512, opened 2025-02-13
render: Switch to wgpu fork with backported GLES limits fix

adrian17:
Should fix hopefully all instances of https://github.com/ruffle-rs/ruffle/issues/19278 that appeared after we bumped to v23.

Same with all Android errors with this style: https://github.com/ruffle-rs/ruffle-android/issues/338

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,12 @@ license = "MIT OR Apache-2.0"
 repository = "https://github.com/ruffle-rs/ruffle"
 version = "0.1.0"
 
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -45,6 +45,12 @@ license = "MIT OR Apache-2.0"
 repository = "https://github.com/ruffle-rs/ruffle"
 version = "0.1.0"
 
+# Remove this when v25 releases
```
torokati44, line 48: nit: `TODO`, but whatever

torokati44 (approved):

PR #19517, opened 2025-02-14
avm2: Use Line Feed character for textInput newline events

### LINKED ISSUE #19482: Multiline TextField + TextEvent.TEXT_INPUT: event.text should be \n instead of \r when user press enter
### Describe the bug

On TextEvent.TEXT_INPUT event handler, when I get `event.text`:

Flash: I got `event.text` = `\n`
Ruffle: I got `event.text` = `\r`

### Expected behavior

`event.text` should return `\n`

### Content Location

There's no need a swf to reproduce the bug.

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

_No response_

### Additional information

_No response_

evilpie:
Fixes #19482


tests/tests/swfs/avm2/edittext_newline_character/Test.as
```diff
@@ -0,0 +1,40 @@
...
+            trace("onChange");
+
+            var text: String = evt.target.text;
+            for (var i: int = 0; i < text.length; i++) {
+                trace(" U+" + text.charCodeAt(i).toString(16));
+            }
+        }
+
+        public function onKeyDown(evt: KeyboardEvent): void {
+            // TODO: We must not be firing this event yet.
```
kjarosh, line 32: Why does this comment mean? Why can't you just fire the event?

kjarosh: Oh, sorry, I misunderstood the comment. I thought we cannot fire the event here somehow.

The key down/up events are separate from text input events, you need to add them to `input.json`.

evilpie: Do we not have a way of producing the TEXT_INPUT and KEY_DOWN event with just one line in `input.json`? It seems a bit pointless to test if both results are basically hardcoded. (We also don't seem to support specific a character for test KeyDown events, which is another problem)

kjarosh: > Do we not have a way of producing the TEXT_INPUT and KEY_DOWN event with just one line in `input.json`?

Nope

> It seems a bit pointless to test if both results are basically hardcoded.

Text input is inherently different from key up/down, because it goes through the OS (keyboard layout, mapping, etc.). We cannot infer one from the other, it's done outside of Ruffle.

> We also don't seem to support specifying a character for test KeyDown events, which is another problem

I agree, I remember adding key char as an optional field in the automated event. Maybe I'll find the commit

kjarosh: Oh, I remember now. I worked on redoing key events in Ruffle to properly support key mapping. I believe we should do it first, because key code can be inferred from key char (contrary to text input).

kjarosh: I think we have to add KeyDown/KeyUp events to `input.json` and we cannot combine them with `TextInput` due to how input works on Windows: https://github.com/ruffle-rs/ruffle/pull/19874

evilpie:
I honestly kind of forgot what this was about. It seems good to me though.

PR #19532, opened 2025-02-16
avm2: Implement flash.geom.PerspectiveProjection

cookie-s:
doc: https://docs.ruffle.rs/en_US/FlashPlatform/reference/actionscript/3/flash/geom/PerspectiveProjection.html

This PR implements `flash.geom.PerspectiveProjection` features. However, assignment to `flash.display.DisplayObject` through `transform.perspectiveProjection` is not yet supported. No effect to visual, either.


core/src/avm2/globals/flash/geom/perspective_projection.rs
```diff
@@ -0,0 +1,67 @@
...
+    let dobj = this
+        .get_slot(pp_slots::DISPLAY_OBJECT)
+        .as_object()
+        .and_then(|e| e.as_display_object());
+
+    match dobj {
+        // Not associated with any DO
+        None => 500.0,
+        // Stage's PerspectiveProjection
+        Some(dobj) if dobj.as_stage().is_some() => 500.0,
```
cookie-s, line 20: I don't know where those values come from. It's the output from the test result in FP.

core/src/avm2/globals/flash/geom/PerspectiveProjection.as
```diff
@@ -1,44 +1,54 @@
...
     import flash.geom.Matrix3D;
     import flash.geom.Point;
 
     public class PerspectiveProjection {
-        // Getters are stubbed with what seem to be Flash's default values
+        [Ruffle(NativeAccessible)]
+        private var displayObject: DisplayObject = null;
+
+        [Ruffle(NativeAccessible)]
+        private var fov: Number = 55.0;
```
cookie-s, line 14: From `TestTransform()` result, `fieldOfView` should be saved and `focalLength` should be computed from it.
`PP(FOV:55, FL:480.25)` assignment to `Sprite.transform` produces `PP(FOV:55, FL:288.15)` (when SWF width is 300).

emergence75:
this might be useful, take a look
https://github.com/emergence75/ruffle/blob/flash.geom-alternate/core/src/avm2/globals/flash/geom/PerspectiveProjection.as

core/src/avm2/globals/flash/geom/PerspectiveProjection.as
```diff
@@ -1,44 +1,54 @@
...
         public function set projectionCenter(value:Point) {
             stub_setter("flash.geom.PerspectiveProjection", "projectionCenter");
+            this.center = value;
         }
 
         public function toMatrix3D():Matrix3D {
-            stub_method("flash.geom.PerspectiveProjection", "toMatrix3D");
-            return new Matrix3D();
+            var fl: Number = this.focalLength;
+            return new Matrix3D(new <Number>[
```
adrian17, line 46: To be sure, this is a full matrix impl?

adrian17 (commented):
I think it all looks fine, but if you don’t mind, I wanna investigate a bit into the remaining unknowns; but I’ll only be able to do it on wednesday.

adrian17:
As for emergence75’s code, note that we did find issues with some of the changes there (see: https://github.com/ruffle-rs/ruffle/pull/18495#issuecomment-2460773494 and the preceding comment). I know that since then it was replaced by a newer PR (which I’m sorry for not commenting on), but as far as I’m aware, the listed concerns about the changes still apply :( With that, I’d be careful with treating it as a source.

core/src/avm2/globals/flash/geom/PerspectiveProjection.as
cookie-s: Yes, I believe so. For other methods, especially setters are related to rendering/visual, but I believe this `toMatrix3D` just returns the Matrix3D object.

Ah, but only if the small but not very small floating point number differences are allowed. Please see `precision` in `TestToMatrix3D` test function. There is unexplainable numerical errors which is now rounded to approximation.

Except for that, I believe all features in `toMatrix3D` are implemented.

cookie-s:
> this might be useful, take a look

Oh thank you. My PR might be better to implement the input validation for fieldOfView setter.

> I wanna investigate a bit into the remaining unknowns

Sure thank you!

core/src/avm2/globals/flash/geom/perspective_projection.rs
```diff
@@ -0,0 +1,78 @@
...
+        // root's PerspectiveProjection
+        Some(dobj) if dobj.is_root() => activation.context.stage.stage_size().0 as f64,
```
adrian17, line 23: If this has the same logic as the one below, any reason for it to be separate?

core/src/avm2/globals/flash/geom/PerspectiveProjection.as
```diff
@@ -1,44 +1,57 @@
...
     import __ruffle__.stub_method;
     import __ruffle__.stub_setter;
+    import flash.display.DisplayObject;
     import flash.geom.Matrix3D;
     import flash.geom.Point;
 
     public class PerspectiveProjection {
-        // Getters are stubbed with what seem to be Flash's default values
+        [Ruffle(NativeAccessible)]
+        private var displayObject: DisplayObject = null;
```
adrian17, line 11: Does this actually need the live reference to the DO? What if instead, only the width value was stored?

core/src/avm2/globals/flash/geom/perspective_projection.rs
cookie-s: Good point. No! Thank you, merged to the line below  at f6f2aba

core/src/avm2/globals/flash/geom/PerspectiveProjection.as
cookie-s: True, storing only `width` passes the current test, but another observation by for example this code:
```
        private function TestTransformUpdate(): void {
            var s: Sprite = new Sprite();

            trace("// Set non-null to transform.perspectiveProjection");
            s.transform.perspectiveProjection = new PerspectiveProjection();
            printProps(s.transform.perspectiveProjection);

            trace("// Set null to transform.perspectiveProjection");
            s.transform.perspectiveProjection.fieldOfView = 100;
            printProps(s.transform.perspectiveProjection);
        }
```
shows some setter in `transform.perspectiveProjection : PerspectiveProjection` should update the values in DO. 

It's the similar behavior to `transform.matrix3D` rather than `transform.matrix` in terms of what we discussed before: https://github.com/ruffle-rs/ruffle/pull/18810#issuecomment-2521639207
(although that setter behavior is not implemented to matrix3D and `transform.perspectiveProjection` setter is almost imcomplete as of now.)

Given that, we will anyway need reference to DO at some point I believe.

adrian17: Oh, I see, yeah.
```as
            var a = dobj.transform.perspectiveProjection;
            var b = dobj.transform.perspectiveProjection;
            trace(a === b); // false
            a.fieldOfView = 123;
            trace(b.fieldOfView); // 123
```

adrian17: Can you please add some TODO comments explaining the above, and maybe a known-failure test?

> although that setter behavior is not implemented to matrix3D and transform.perspectiveProjection setter is almost imcomplete as of now

Sure, I think this is an okay compromise.

adrian17 (approved):

core/src/avm2/globals/flash/geom/PerspectiveProjection.as
cookie-s: Thank you! I added a few TODO/FIXME comments.

adrian17:
Thanks!

(as for the z-order PR, I'm trying to find time to take a look at it, sorry for a delay 😅 )

cookie-s:
Thank you for merging!

(No problem at all, I know it's not a tiny PR.)

PR #19534, opened 2025-02-16
ci: Use dotnet to install wix extensions

sombraguerrero:
@danielhjacobs Proposes we hardcode to either the last general release or the RC so as to resolve the conflicting installs which are likely blocking the desktop builds. I would opt to stay with the general release for now.

### DIFF
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -121,9 +121,9 @@ jobs:
 
       - name: Install WiX
         run: |
-          dotnet tool install --global wix
-          wix extension add -g WixToolset.UI.wixext
-          wix extension add -g WixToolset.Util.wixext
+          dotnet tool install --global wix --version 5.0.2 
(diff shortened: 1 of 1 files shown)


danielhjacobs:
That was my initial suggestion, but i see you made another change which may make that unnecessary. Are we sure using dotnet for installing the extensions will require pinning at all? Maybe `dotnet add package` ignores prerelease versions just like `dotnet tool install`.

sombraguerrero:
> That was my initial suggestion, but i see you made another change which may make that unnecessary. Are we sure using dotnet for installing the extensions will require pinning at all? Maybe `dotnet add package` ignores prerelease versions just like `dotnet tool install`.

Well, that might be true, but I figure since the version arguments appear to be valid across the board, better to over-pin than under-pin :D

sombraguerrero:
@danielhjacobs Very well, I have been convinced!

danielhjacobs (approved):

torokati44:
Thanks for fixing this!

PR #19541, opened 2025-02-16
refactor: Use ActionScript names for KeyCodes

kjarosh:
This hopefully brings a bit of consistency when using `KeyCode` in code. Before that, when looking for a specific key code, it was easier to find it by its code rather than by its name.


evilpie:
Where do the ActionScript names actually come from? Because the change for AS1 from ALT to ALTERNATE seems contradictory.

kjarosh:
They come from AS3 (https://docs.ruffle.rs/en_US/FlashPlatform/reference/actionscript/3/flash/ui/Keyboard.html#ALTERNATE) as there are more key codes defined and they seem to have more descriptive names. Unfortunately the names differ a bit between AVM1 and AVM2.

I don't like the `ALTERNATE` name either, but I changed it for consistency. Although, I can change it back to `ALT`, we don't have to be consistent.

core/src/events.rs
```diff
@@ -477,18 +477,18 @@ impl KeyCode {
     pub const MOUSE_MIDDLE: KeyCode = KeyCode(4);
     pub const BACKSPACE: KeyCode = KeyCode(8);
     pub const TAB: KeyCode = KeyCode(9);
-    pub const RETURN: KeyCode = KeyCode(13);
+    pub const ENTER: KeyCode = KeyCode(13);
```
evilpie, line 480: Can you add a reference to https://docs.ruffle.rs/en_US/FlashPlatform/reference/actionscript/3/flash/ui/Keyboard.html#summaryTableConstant in the coment on line 469?

evilpie (approved):
Seems like a clear improvement. The new names match AS3. I would probably go with `ALT` still, because that seems like the commonly used name for it.

core/src/events.rs
kjarosh: Done!

kjarosh:
I agree regarding `ALT`, changed that back

PR #19542, opened 2025-02-17
Version pinning just the wix base

sombraguerrero:
Realized after looking into the approach to this more closely that just the base package should be pinned. The toolset extensions need to continue to be added globally to the cache as they were since we're not actually coding in a .NET project.


adrian17 (approved):

torokati44:
Looks like this didn't work either: https://github.com/ruffle-rs/ruffle/actions/runs/13380475153/job/37368051352

PR #19544, opened 2025-02-17
refactor: Add InputEvent

kjarosh:
This refactor simplifies code and is a first step towards tidying up and unifying input handling across platforms.

The most important improvement is the addition of `InputEvent`, which represents a mapped `PlayerEvent` related to input. It proves to be useful even now by discarding gamepad events and having click index present at all times. In the future this will make it possible to pass information about keys in `PlayerEvent` without any knowledge about Flash Player (so no `KeyCode`).

### DIFF
--- a/core/src/input.rs
+++ b/core/src/input.rs
@@ -1,6 +1,48 @@
-use crate::events::{GamepadButton, KeyCode, MouseButton, PlayerEvent, TextControlCode};
+use crate::events::{
+    GamepadButton, KeyCode, MouseButton, MouseWheelDelta, PlayerEvent, TextControlCode,
+};
 use chrono::{DateTime, TimeDelta, Utc};
 use std::collections::{HashMap, HashSet};
+
+/// An event describing input in general.
+///
+/// It's usually a processed [`PlayerEvent`].
+#[derive(Debug, Clone, Copy)]
+pub enum InputEvent {
+    KeyDown {
+        key_code: KeyCode,
+        key_char: Option<char>,
+    },
+    KeyUp {
+        key_code: KeyCode,
+        key_char: Option<char>,
+    },
+    MouseMove {
+        x: f64,
+        y: f64,
+    },
+    MouseUp {
+        x: f64,
+        y: f64,
+        button: MouseButton,
+    },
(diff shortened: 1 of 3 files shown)


core/src/input.rs
```diff
@@ -67,66 +109,92 @@ impl InputManager {
         }
     }
 
-    pub fn map_input_event(&mut self, event: PlayerEvent) -> Option<PlayerEvent> {
-        // Optionally transform gamepad button events into key events.
-        match event {
+    pub fn process_input_event(&mut self, event: PlayerEvent) -> Option<InputEvent> {
```
evilpie, line 112: I feel like the name doesn't really make much sense now that we actually have an InputEvent type. It should probably be `process_player_input` or something like that.

kjarosh: Changed to `process_event` to be as neutral as possible :switzerland:

core/src/input.rs
```diff
@@ -67,66 +109,92 @@ impl InputManager {
         }
     }
 
-    pub fn map_input_event(&mut self, event: PlayerEvent) -> Option<PlayerEvent> {
-        // Optionally transform gamepad button events into key events.
-        match event {
+    pub fn process_event(&mut self, event: PlayerEvent) -> Option<InputEvent> {
+        let event = Some(match event {
```
evilpie, line 113: Remove `Some` here.

core/src/input.rs
```diff
@@ -67,66 +109,92 @@ impl InputManager {
...
+
+            // The following are not input events.
+            PlayerEvent::FocusGained | PlayerEvent::FocusLost => return None,
+        });
+
+        if let Some(event) = event {
+            self.handle_event(&event);
         }
+
+        event
```
evilpie, line 170: Add `Some` here.

evilpie (approved):

core/src/input.rs
kjarosh: Damn, it's embarrassing

core/src/input.rs
kjarosh: :face_with_peeking_eye:

PR #19552, opened 2025-02-17
tests: Decouple `input.json` from Flash key codes

kjarosh:
This refactor removes key codes from `input.json` aiming at moving key code mapping to core and creating a common, Flash-independent interface for input.

CC @evilpie


evilpie:
Much nicer! Thank you.

tests/framework/src/runner.rs
```diff
@@ -574,3 +574,46 @@ fn assert_text_matches(ruffle: &str, flash: &str) -> Result<()> {
         Ok(())
     }
 }
+
+fn automated_key_to_key_code(automated_key: AutomatedKey) -> KeyCode {
+    match automated_key {
+        AutomatedKey::Char(ch) => KeyCode::from_code(ch.to_ascii_uppercase() as u32),
```
evilpie, line 580: I think we should assert that the char is something like a..z.

evilpie (approved):

tests/framework/src/runner.rs
```diff
@@ -235,12 +235,12 @@ impl TestRunner {
                         _ => panic!("MouseWheel: expected only one of 'lines' or 'pixels'"),
                     },
                 },
-                AutomatedEvent::KeyDown { key_code } => PlayerEvent::KeyDown {
-                    key_code: KeyCode::from_code(*key_code),
+                AutomatedEvent::KeyDown { key } => PlayerEvent::KeyDown {
+                    key_code: automated_key_to_key_code(*key),
                     key_char: None,
```
evilpie, line 240: I guess we could actually do something useful here in the future :tada:

tests/framework/src/runner.rs
```diff
@@ -574,3 +574,46 @@ fn assert_text_matches(ruffle: &str, flash: &str) -> Result<()> {
...
+            '6' => KeyCode::NUMPAD_6,
+            '7' => KeyCode::NUMPAD_7,
+            '8' => KeyCode::NUMPAD_8,
+            '9' => KeyCode::NUMPAD_9,
+            '*' => KeyCode::NUMPAD_MULTIPLY,
+            '+' => KeyCode::NUMPAD_ADD,
+            '-' => KeyCode::NUMPAD_SUBTRACT,
+            '.' | ',' => KeyCode::NUMPAD_DECIMAL,
+            '/' => KeyCode::NUMPAD_DIVIDE,
+            ch => KeyCode::from_code(ch.to_ascii_uppercase() as u32),
```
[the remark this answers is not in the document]
kjarosh: That makes sense!

tests/framework/src/runner.rs
kjarosh: Actually key codes may represent non-ASCII characters. If you try inputting a non-ASCII character you'll get its Unicode code point as the value of key code.

That's because key codes are logical in Flash, not physical.

tests/framework/src/runner.rs
kjarosh: You bet! I'm cookin' somethin'

Issue #19490, opened 2025-02-11
Error: Flash Player 11.5 or AIR 3.5 required to use this feature.


ylazy:
### Describe the bug

My swf works with Flash Player and Ruffle on desktop, but fails on the web. I got this error:

```
ERROR core/src/display_object/movie_clip.rs:2171 Got "Error: Flash Player 11.5 or AIR 3.5 required to use this feature." when constructing AVM2 side of movie clip of type main::Main
```

### Expected behavior

My swf should work properly.

### Content Location

[application.zip](https://github.com/user-attachments/files/18744919/application.zip)

### Affected platform

Desktop app

### Operating system

Windows 11 24H2

### Browser

Chrome 133.0.6943.59 (latest)

ylazy:
Hi!

Is there anything I can do to help tracking this issue?

I'm building as3 libs that support air/ruffle and I need to resolve this issue to continue.

Thanks!

ylazy:
Thank you @kjarosh 
I set `logLevel` to `info` and it worked.
I think stacktraces is an important feature and it should be enabled by default.

kjarosh:
Related:
* https://github.com/ruffle-rs/ruffle/pull/10310
* https://github.com/ruffle-rs/ruffle/pull/7786#issuecomment-1231088761

PR #7786, opened 2022-08-28
avm2: Implement Error.getStackTrace

EmperorBale:
Implements `getStackTrace` for Error. Also does a few other things:
1. Made a small fix to build_playerglobal where it generated invalid code for custom instance allocators for classes without a package name.
2. `TObject::to_string` & `TObject::to_locale_string` now take an activation, rather than mutationcontext.

PR #19571, opened 2025-02-19
avm2: Enable stack traces for Flash Player 11.5+

kjarosh:
For Flash Player 11.5 and later, stack traces are always enabled. Some content uses this behavior to verify whether the user is using Flash Player 11.5 onwards.

See <https://docs.ruffle.rs/en_US/FlashPlatform/reference/actionscript/3/Error.html#getStackTrace()>

* Fixes https://github.com/ruffle-rs/ruffle/issues/19490

Related:
* https://github.com/ruffle-rs/ruffle/pull/10310
* https://github.com/ruffle-rs/ruffle/pull/7786#issuecomment-1231088761

CC @adrian17 @Lord-McSweeney


adrian17 (approved):

n0samu (approved):

kjarosh:
> LGTM, the only thing I'm not a fan of is calling it 'player mode', since the word is super generic. But it's not exposed via CLI so I guess it's fine for now?

I agree, let me know if you have any other ideas about how to name it and we can change it. I intentionally haven't exposed it in CLI due to this reason.

PR #19574, opened 2025-02-19
Small improvements to render crates

a1phyr:
- Use array methods instead of collecting to a vector first
- Remove `Arc`s around `wgpu` objects thanks to wgpu 24 improvements


torokati44:
The first commit has no tag prefix.

adrian17:
Oh oops. Once all checks pass, I'll rebase with fixed name, then just force merge without waiting for checks.

adrian17:
Thank you!

Issue #19583, opened 2025-02-21
Error on https://www.gamenora.com/upload/games/swf/Ben%2010:%20Ben%20to%20the%20Rescue/


Vannicin:
# Error Info
Error name: Error
Error message: panicked at core/src/display_object/container.rs:1019:22:
called `Option::unwrap()` on a `None` value
Error stack:
```
Error: panicked at core/src/display_object/container.rs:1019:22:
called `Option::unwrap()` on a `None` value
    at Re.n.wbg.__wbg_new_c68d7209be747379 (https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.5b06ecac44bf6691b056.js:1:55885)
    at ruffle_web.wasm.js_sys::Error::new::h90ef3ef3139d9699 (wasm://wasm/ruffle_web.wasm-031a455e:wasm-function[12682]:0x973173)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once{{vtable.shim}}::h82cbe3b1cb3f456a (wasm://wasm/ruffle_web.wasm-031a455e:wasm-function[9882]:0x91e3d6)
    at ruffle_web.wasm.std::panicking::rust_panic_with_hook::hb39abb160cd4038c (wasm://wasm/ruffle_web.wasm-031a455e:wasm-function[6955]:0x87c493)
    at ruffle_web.wasm.core::panicking::panic_fmt::h6f4dae69dcc1a6d2 (wasm://wasm/ruffle_web.wasm-031a455e:wasm-function[10711]:0x942a0f)
    at ruffle_web.wasm.core::panicking::panic::hb7d13004db45f2e3 (wasm://wasm/ruffle_web.wasm-031a455e:wasm-function[10003]:0x921012)
    at ruffle_web.wasm.core::option::unwrap_failed::hd9229270312a0397 (wasm://wasm/ruffle_web.wasm-031a455e:wasm-function[12078]:0x965275)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once::ha7aa30e71b4891a5 (wasm://wasm/ruffle_web.wasm-031a455e:wasm-function[1447]:0x3d2480)
    at ruffle_web.wasm.ruffle_core::avm1::function::Executable::exec::h7ba382fb88d5e97c (wasm://wasm/ruffle_web.wasm-031a455e:wasm-function[988]:0x1b80e4)
[...]

Issue #19406, opened 2025-02-02
Error on https://www.gamenora.com/upload/games/swf/Ben%2010:%20Ben%20to%20the%20Rescue/


Geados:
# Error Info
Error name: Error
Error message: panicked at core/src/display_object/container.rs:1019:22:
called `Option::unwrap()` on a `None` value
Error stack:
```
Error: panicked at core/src/display_object/container.rs:1019:22:
called `Option::unwrap()` on a `None` value
    at n.wbg.__wbg_new_c68d7209be747379 (https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.ab84fba652b60e1cdb87.js:1:58331)
    at ruffle_web.wasm.js_sys::Error::new::h300c382d1c8df8b2 (wasm://wasm/ruffle_web.wasm-032a203e:wasm-function[13025]:0x9aa1e7)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once{{vtable.shim}}::h343bd8b1834d46f1 (wasm://wasm/ruffle_web.wasm-032a203e:wasm-function[9916]:0x91cd97)
    at ruffle_web.wasm.std::panicking::rust_panic_with_hook::hcc2503a68438883e (wasm://wasm/ruffle_web.wasm-032a203e:wasm-function[6951]:0x856be0)
    at ruffle_web.wasm.core::panicking::panic_fmt::h931ec2537c26fa22 (wasm://wasm/ruffle_web.wasm-032a203e:wasm-function[10836]:0x953ff3)
    at ruffle_web.wasm.core::panicking::panic::hb20c9056d85d5b5e (wasm://wasm/ruffle_web.wasm-032a203e:wasm-function[10070]:0x92e0b9)
    at ruffle_web.wasm.core::option::unwrap_failed::h7534a988a872bb9f (wasm://wasm/ruffle_web.wasm-032a203e:wasm-function[12320]:0x9854e2)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once::hfb0d75353c1e8b38 (wasm://wasm/ruffle_web.wasm-032a203e:wasm-function[1369]:0x3b5827)
    at ruffle_web.wasm.ruffle_core::avm1::function::Executable::exec::hc1b104c3f5a42fcd (wasm://wasm/ruffle_web.wasm-032a203e:wasm-function[892]:0x19a1ca)
[...]

Issue #18812, opened 2024-11-30
Error on https://www.mariogames.be/Ben_10_To_The_Rescue_Game


waleedreyaan1503:
# Error Info
Error name: Error
Error message: panicked at core/src/display_object/container.rs:1022:22:
called `Option::unwrap()` on a `None` value
Error stack:
```
Error: panicked at core/src/display_object/container.rs:1022:22:
called `Option::unwrap()` on a `None` value
    at imports.wbg.__wbg_new_70a2f23d1565c04c (chrome-extension://donbcfbmhbcapadipfkeojnmajbakjdc/dist/69.js:3887:21)
    at ruffle_web.wasm.js_sys::Error::new::hc7430be370242701 (wasm://wasm/ruffle_web.wasm-03229c4a:wasm-function[12865]:0x98663a)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once{{vtable.shim}}::h1a8fe10950133898 (wasm://wasm/ruffle_web.wasm-03229c4a:wasm-function[9790]:0x9041a4)
    at ruffle_web.wasm.std::panicking::rust_panic_with_hook::h33fe77d38d305ca3 (wasm://wasm/ruffle_web.wasm-03229c4a:wasm-function[6416]:0x82a155)
    at ruffle_web.wasm.core::panicking::panic_fmt::hde8b7aa66e2831e1 (wasm://wasm/ruffle_web.wasm-03229c4a:wasm-function[9502]:0x8fd257)
    at ruffle_web.wasm.core::panicking::panic::hcaca2598a27ec0fc (wasm://wasm/ruffle_web.wasm-03229c4a:wasm-function[9972]:0x915ae8)
    at ruffle_web.wasm.core::option::unwrap_failed::h9aa82eb71128b127 (wasm://wasm/ruffle_web.wasm-03229c4a:wasm-function[12351]:0x96344d)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once::h58d3aa18a2500e37 (wasm://wasm/ruffle_web.wasm-03229c4a:wasm-function[1341]:0x3b533c)
    at ruffle_web.wasm.ruffle_core::avm1::function::Executable::exec::hf35b4a729b8343db (wasm://wasm/ruffle_web.wasm-03229c4a:wasm-function[913]:0x1d559f)
[...]

Issue #18676, opened 2024-11-19
Error on https://www.gamenora.com/upload/games/swf/Ben%2010:%20Ben%20to%20the%20Rescue/


Aaronwilson03:
# Error Info
Error name: Error
Error message: panicked at core/src/display_object/container.rs:1022:22:
called `Option::unwrap()` on a `None` value
Error stack:
```
Error: panicked at core/src/display_object/container.rs:1022:22:
called `Option::unwrap()` on a `None` value
    at n.wbg.__wbg_new_70a2f23d1565c04c (https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.d260019924884f24f9a8.js:1:85360)
    at ruffle_web.wasm.js_sys::Error::new::hf99c185550382360 (wasm://wasm/ruffle_web.wasm-031d430a:wasm-function[12822]:0x9710cd)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once{{vtable.shim}}::h52d59cb7a23c95c3 (wasm://wasm/ruffle_web.wasm-031d430a:wasm-function[9731]:0x8f0653)
    at ruffle_web.wasm.std::panicking::rust_panic_with_hook::h33fe77d38d305ca3 (wasm://wasm/ruffle_web.wasm-031d430a:wasm-function[6386]:0x819016)
    at ruffle_web.wasm.core::panicking::panic_fmt::hde8b7aa66e2831e1 (wasm://wasm/ruffle_web.wasm-031d430a:wasm-function[9442]:0x8e96b3)
    at ruffle_web.wasm.core::panicking::panic::hcaca2598a27ec0fc (wasm://wasm/ruffle_web.wasm-031d430a:wasm-function[9911]:0x901d80)
    at ruffle_web.wasm.core::option::unwrap_failed::h9aa82eb71128b127 (wasm://wasm/ruffle_web.wasm-031d430a:wasm-function[12305]:0x94e161)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once::h1b52861b44534e10 (wasm://wasm/ruffle_web.wasm-031d430a:wasm-function[1349]:0x3a960c)
    at ruffle_web.wasm.ruffle_core::avm1::function::Executable::exec::h3c723d34f4014e13 (wasm://wasm/ruffle_web.wasm-031d430a:wasm-function[896]:0x1b26f1)
[...]

PR #19594, opened 2025-02-22
avm1: Do not panic when swapping children on depth/render list desync

kjarosh:
Unfortunately sometimes the render list and the depth list become desynchronized.  When swapping children this caused panics.

I haven't been able to reproduce those scenarios, but it's better to leave the render/depth list messed up than to panic in this case.

* Fixes https://github.com/ruffle-rs/ruffle/issues/19944
* Fixes https://github.com/ruffle-rs/ruffle/issues/19911
* Fixes https://github.com/ruffle-rs/ruffle/issues/19640
* Fixes https://github.com/ruffle-rs/ruffle/issues/19583
* Fixes https://github.com/ruffle-rs/ruffle/issues/19535
* Fixes https://github.com/ruffle-rs/ruffle/issues/19406
* Fixes https://github.com/ruffle-rs/ruffle/issues/19150
* Fixes https://github.com/ruffle-rs/ruffle/issues/19116
* Fixes https://github.com/ruffle-rs/ruffle/issues/19101
* Fixes https://github.com/ruffle-rs/ruffle/issues/18812
* Fixes https://github.com/ruffle-rs/ruffle/issues/18676
* Fixes https://github.com/ruffle-rs/ruffle/issues/18499
* Fixes https://github.com/ruffle-rs/ruffle/issues/18276
* Fixes https://github.com/ruffle-rs/ruffle/issues/17578
* Fixes https://github.com/ruffle-rs/ruffle/issues/17549
* Fixes https://github.com/ruffle-rs/ruffle/issues/17412
* Fixes https://github.com/ruffle-rs/ruffle/issues/17183
* Fixes https://github.com/ruffle-rs/ruffle/issues/17164
* Fixes https://github.com/ruffle-rs/ruffle/issues/16546
* Fixes https://github.com/ruffle-rs/ruffle/issues/16220
* Fixes https://github.com/ruffle-rs/ruffle/issues/16118
* Fixes https://github.com/ruffle-rs/ruffle/issues/14539
* Fixes https://github.com/ruffle-rs/ruffle/issues/14167
* Fixes https://github.com/ruffle-rs/ruffle/issues/13577
* Fixes https://github.com/ruffle-rs/ruffle/issues/12698
* Fixes https://github.com/ruffle-rs/ruffle/issues/9568
* Fixes https://github.com/ruffle-rs/ruffle/issues/8875
* Fixes https://github.com/ruffle-rs/ruffle/issues/8723
* Fixes https://github.com/ruffle-rs/ruffle/issues/8552
* Fixes https://github.com/ruffle-rs/ruffle/issues/6614
* Fixes https://github.com/ruffle-rs/ruffle/issues/6512
* Fixes https://github.com/ruffle-rs/ruffle/issues/5708
* Fixes https://github.com/ruffle-rs/ruffle/issues/5084
* Fixes https://github.com/ruffle-rs/ruffle/issues/5052
* Fixes https://github.com/ruffle-rs/ruffle/issues/4701

### DIFF
--- a/core/src/display_object/container.rs
+++ b/core/src/display_object/container.rs
@@ -999,31 +999,46 @@ impl<'gc> ChildContainer<'gc> {
             let prev_position = self
                 .render_list
                 .iter()
-                .position(|x| DisplayObject::ptr_eq(*x, prev_child))
-                .unwrap();
+                .position(|x| DisplayObject::ptr_eq(*x, prev_child));
             let next_position = self
                 .render_list
                 .iter()
-                .position(|x| DisplayObject::ptr_eq(*x, child))
-                .unwrap();
-            self.render_list_mut().swap(prev_position, next_position);
+                .position(|x| DisplayObject::ptr_eq(*x, child));
+
+            match (prev_position, next_position) {
+                (Some(prev_position), Some(next_position)) => {
+                    self.render_list_mut().swap(prev_position, next_position);
+                }
+                _ => {
+                    tracing::error!(
+                        "ChildContainer::swap_at_depth: A child is missing in the render list"
+                    );
+                }
+            }
         } else {
             self.depth_list.remove(&prev_depth);
 
-            let old_position = self
+            if let Some(old_position) = self
                 .render_list
                 .iter()
                 .position(|x| DisplayObject::ptr_eq(*x, child))
-                .unwrap();
-            self.render_list_mut().remove(old_position);
+            {
+                self.render_list_mut().remove(old_position);
+            }
 
             if let Some((_, below_child)) = self.depth_list.range(..depth).next_back() {
                 let new_position = self
                     .render_list
                     .iter()
                     .position(|x| DisplayObject::ptr_eq(*x, *below_child))
-                    .unwrap();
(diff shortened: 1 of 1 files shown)


core/src/display_object/container.rs
```diff
@@ -993,31 +993,41 @@ impl<'gc> ChildContainer<'gc> {
...
 
             if let Some((_, below_child)) = self.depth_list.range(..depth).next_back() {
                 let new_position = self
                     .render_list
                     .iter()
                     .position(|x| DisplayObject::ptr_eq(*x, *below_child))
-                    .unwrap();
-                self.render_list_mut().insert(new_position + 1, child);
+                    .map(|pos| pos + 1)
+                    .unwrap_or(0);
```
Lord-McSweeney, line 1029: Would it be better for this to log an error instead of inserting it at position 0 if the DO wasn't in the render list?

kjarosh: Added an error, although it won't tell us much, only that a desync happened and we should expect depth list and render list to not be in sync. We would have to add debug checks during mutation to get useful errors or panics.

I'll also be trying to refactor this in the future so that the desync is less likely / impossible to happen.

Issue #18601, opened 2024-11-14
Error on https://codeincode.io/


harrison001:
# Error Info
Error name: Error
Error message: panicked at core/src/display_object/edit_text.rs:1224:73:
called `Option::unwrap()` on a `None` value
Error stack:
```
Error: panicked at core/src/display_object/edit_text.rs:1224:73:
called `Option::unwrap()` on a `None` value
    at push.69.imports.wbg.__wbg_new_70a2f23d1565c04c (chrome-extension://donbcfbmhbcapadipfkeojnmajbakjdc/dist/69.js:3854:21)
    at ruffle_web.wasm.js_sys::Error::new::h1f24b56eea900b9e (wasm://wasm/ruffle_web.wasm-032322b2:wasm-function[12771]:0x98b959)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once{{vtable.shim}}::h97dbdb90099d7661 (wasm://wasm/ruffle_web.wasm-032322b2:wasm-function[9792]:0x91d423)
    at ruffle_web.wasm.std::panicking::rust_panic_with_hook::h33fe77d38d305ca3 (wasm://wasm/ruffle_web.wasm-032322b2:wasm-function[6361]:0x8171ff)
    at ruffle_web.wasm.core::panicking::panic_fmt::hde8b7aa66e2831e1 (wasm://wasm/ruffle_web.wasm-032322b2:wasm-function[9511]:0x90dab5)
    at ruffle_web.wasm.core::panicking::panic::hcaca2598a27ec0fc (wasm://wasm/ruffle_web.wasm-032322b2:wasm-function[9962]:0x9238cf)
    at ruffle_web.wasm.core::option::unwrap_failed::h9aa82eb71128b127 (wasm://wasm/ruffle_web.wasm-032322b2:wasm-function[12302]:0x969554)
    at ruffle_web.wasm.ruffle_core::display_object::edit_text::EditText::propagate_text_binding::hd6446a42ac6a6aef (wasm://wasm/ruffle_web.wasm-032322b2:wasm-function[2220]:0x5507fa)
    at ruffle_web.wasm.ruffle_core::display_object::edit_text::EditText::text_input::h8408d94ef8bb66e5 (wasm://wasm/ruffle_web.wasm-032322b2:wasm-function[1911]:0x4e4b84)
[...]

Issue #15466, opened 2024-03-06
Error on https://ruffle.rs/demo/


matanzvili:
# Error Info
Error name: Error
Error message: panicked at core/src/display_object/edit_text.rs:1201:73:
called `Option::unwrap()` on a `None` value
Error stack:
```
Error: panicked at core/src/display_object/edit_text.rs:1201:73:
called `Option::unwrap()` on a `None` value
    at Q.n.wbg.__wbg_new_28c511d9baebfa89 (https://ruffle.rs/demo/assets/ruffle_web-wasm_extensions-Z7gokAD-.js:2:71974)
    at js_sys::Error::new::hfa19c20c83d85194 (wasm://wasm/03a14b6a:wasm-function[13891]:0xb10d14)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::h71e7167ef760ecf5 (wasm://wasm/03a14b6a:wasm-function[11506]:0xabbe74)
    at std::panicking::rust_panic_with_hook::h1e6ac5d404b8e31b (wasm://wasm/03a14b6a:wasm-function[7053]:0x9846da)
    at core::panicking::panic_fmt::h87755523850ece9e (wasm://wasm/03a14b6a:wasm-function[11211]:0xab5724)
    at core::panicking::panic::hf4ba1575e20e9f91 (wasm://wasm/03a14b6a:wasm-function[10738]:0xa9b080)
    at ruffle_core::display_object::edit_text::EditText::propagate_text_binding::h220182f659a98513 (wasm://wasm/03a14b6a:wasm-function[2288]:0x615a6a)
    at ruffle_core::display_object::edit_text::EditText::text_input::h2c2dc58ed88411d1 (wasm://wasm/03a14b6a:wasm-function[1404]:0x456a68)
    at ruffle_core::player::Player::handle_event::hae8536ed120026e9 (wasm://wasm/03a14b6a:wasm-function[776]:0x10b2af)
    at ruffle_web::Ruffle::with_instance::h2959fd1e4331c5f0 (wasm://wasm/03a14b6a:wasm-function[1217]:0x3c3b04)
```

# Player Info
Allows script access: false
Renderer: wgpu
Adapter Backend: Gl
Adapter Name: "ANGLE (Apple, Apple M2 Max, OpenGL 4.1)"
Adapter Device Type: IntegratedGpu
Adapter Driver Name: ""
Adapter Driver Info: ""
[...]

PR #19597, opened 2025-02-22
core: Do not panic on missing parent in text variable binding

kjarosh:
Turns out that sometimes `try_bind_text_field_variable` is executed when we don't have a parent.

* Fixes https://github.com/ruffle-rs/ruffle/issues/18601
* Fixes https://github.com/ruffle-rs/ruffle/issues/15466

### DIFF
--- a/core/src/display_object/edit_text.rs
+++ b/core/src/display_object/edit_text.rs
@@ -1370,74 +1370,74 @@ impl<'gc> EditText<'gc> {
         activation: &mut Avm1Activation<'_, 'gc>,
         set_initial_value: bool,
     ) -> bool {
-        if let Some(var_path) = self.variable() {
-            let mut bound = false;
+        let Some(var_path) = self.variable() else {
+            // No variable for this text field; success by default
+            return true;
+        };
 
-            // Any previous binding should have been cleared.
-            debug_assert!(self.0.read().bound_stage_object.is_none());
+        // Any previous binding should have been cleared.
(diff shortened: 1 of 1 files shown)


core/src/display_object/edit_text.rs
```diff
@@ -1370,74 +1370,75 @@ impl<'gc> EditText<'gc> {
...
-            while parent.as_avm1_button().is_some() {
-                parent = parent.avm1_parent().unwrap();
-            }
+        // Avoid double-borrows by copying the string.
+        // TODO: Can we avoid this somehow? Maybe when we have a better string type.
+        let variable_path = WString::from_utf8(&var_path);
+        drop(var_path);
+
+        let Some(mut parent) = self.avm1_parent() else {
+            return bound;
```
evilpie, line 1389:
```suggestion
            return false;
```

Let's move `mut bound` just before `activation.run_with_child_frame_for_display_object`

core/src/display_object/edit_text.rs
```diff
@@ -1370,74 +1370,75 @@ impl<'gc> EditText<'gc> {
...
+        };
+        while parent.as_avm1_button().is_some() {
+            let Some(p) = parent.avm1_parent() else {
+                return bound;
```
evilpie, line 1393:
```suggestion
                return false;
```

evilpie (approved):

core/src/display_object/edit_text.rs
kjarosh: Done

PR #19612, opened 2025-02-23
core: Work toward removing static strings

Lord-McSweeney:


core/src/loader.rs
```diff
@@ -2533,13 +2537,13 @@ impl<'gc> Loader<'gc> {
...
                         broadcaster,
-                        uc,
-                        "broadcastMessage".into(),
+                        istr!(uc, "broadcastMessage"),
                         &[
-                            "onLoadError".into(),
+                            istr!(uc, "onLoadError").into(),
                             clip.object(),
-                            "LoadNeverCompleted".into(),
+                            istr!(uc, "LoadNeverCompleted").into(),
```
adrian17, line 2544: _This_ kind of thing could definitely stay dynamic.

core/src/backend/audio.rs
```diff
@@ -384,6 +385,9 @@ impl<'gc> AudioManager<'gc> {
         // them after running 'retain()'
         let mut event_targets = Vec::new();
 
+        // We need to declare this up here to avoid borrow-checker issues
+        let on_sound_complete_string = istr!(context, "onSoundComplete");
```
adrian17, line 389: This could stay dynamic too? (and be moved lower then?)

core/src/avm1/runtime.rs
```diff
@@ -587,10 +587,11 @@ pub fn skip_actions(reader: &mut Reader<'_>, num_actions_to_skip: u8) {
...
     match &error {
         Error::ThrownValue(value) => {
-            let message = value
-                .coerce_to_string(activation)
-                .unwrap_or_else(|_| "undefined".into());
-            activation.context.avm_trace(&message.to_utf8_lossy());
+            if let Ok(message) = value.coerce_to_string(activation) {
+                activation.context.avm_trace(&message.to_utf8_lossy());
+            } else {
+                activation.context.avm_trace("[type Object]");
```
adrian17, line 593: To be sure, is this a behavioral change?

Lord-McSweeney: Yes, it matches FP now. I'll add a comment to explain.

adrian17 (approved):

Issue #7381, opened 2022-07-09
Problem with Key.getCode on AZERTY keyboard: the layout is always QWERTY on Ruffle


Txori:
### Describe the bug

On AZERTY keyboard, pressing the A key should give the code 65. Instead, it gives the code 81 of the Q letter (QWERTY).
On Ruffle web version, the layout is always QWERTY.
On Ruffle desktop, and Flash Player, it works normally.

I made a simple test to try it out:
https://www.txori.com/stuff/ruffle/keycode_azerty.php

### Expected behavior

Pressing A on AZERTY should retrieve the A Key.getCode
And that goes for every keyboard layouts.

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

Chrome 97

Amseog:
Thank you for pointing this out, because as a German user, I have been having a similar problem:
In Germany, we use a QWERTZ layout (with "Y" and "Z" being switched), and in a lot of Flash games, typing the "Z" key returns a "Y" instead. 

Would be super-cool if this could be fixed some day!
As always: Thanks to all Ruffle contributors, I am so grateful for your work :)

hatal175:
Ruffle uses [KeyboardEvent.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) to determine which physical key was pressed. Unfortunately that ignores keyboard layout. KeyboardEvent.key doesn't but that is the actual character, as is a or A if shift is pressed or any other letter in any alphabet.

The documentation has Keyboard.getLayoutMap which is supposed to solve this problem but this isn't supported in all browsers: https://developer.mozilla.org/en-US/docs/Web/API/Keyboard/getLayoutMap.
It's actually experimental and it is not sure at all whether it will be supported in the future (security concerns: https://mozilla.github.io/standards-positions/#keyboard-map).

Toad06:
Related to #1046.

mero09-99:
any update on this?

Issue #1046, opened 2020-08-25
[Cable Capers 2] Keyboard differences vs Flash version


Toad06:
https://www.newgrounds.com/portal/view/63906

---

**Controls are different between keyboard layouts**

The game uses the following controls:
X - to shoot pipes
Z - to jump

The previous statement will always be true in Flash, whether you have a QWERTY or AZERTY keyboard.
In Ruffle, AZERTY users will always have to press the "W" key to jump (Z and W are swapped from the QWERTY layout).

---

**Arrow keys remain down**

If you press the left or right key, then press any other key while keeping the arrow key down, the character will continue to walk even after you have released the arrow key (you will need to press the same arrow key again to make him stop).

Herschel:
This only applied to the web build. Sticking keys are fixed by cab1634.

For the layout issue, we can look at `event.key` instead of `event.code` in `WebInputBackend`, but I want to research a bit more to verify the exact behavior (i.e., does `keyPress` handle this differently than `Key.getCode`?) This could also be something we have an option for the user to specify.

PR #19616, opened 2025-02-24
core: Key input overhaul

kjarosh:
The main goal of this refactor is to unify key input across all platforms and move all Flash Player related behaviors to core in order to get rid of differences across platforms and minimize bugs.

Instead of providing `KeyCode`s in `PlayerEvent`s, frontends now provide `KeyDescriptor`s, which describe the pressed key on a high level. Then, Ruffle core does all necessary mappings to `KeyCode`s (and key location).

Additionally, this PR implements support for `KeyboardEvent.keyLocation`.

* Fixes https://github.com/ruffle-rs/ruffle/issues/7381
* Fixes https://github.com/ruffle-rs/ruffle/issues/1046

CC @evilpie


desktop/src/util.rs
```diff
@@ -55,127 +61,276 @@ pub fn winit_to_ruffle_text_control(
...
-    let char = char.chars().next()?;
-
-    if char.is_ascii_alphabetic() {
-        // ASCII alphabetic characters are all mapped to
-        // their respective KeyCodes, which happen to have
-        // the same numerical value as uppercase characters.
-        return Some(KeyCode::from_code(char.to_ascii_uppercase() as u32));
+fn alpha_to_logical(ch: &str) -> LogicalKey {
+    // TODO What if we get multiple chars?
+    if let Some(ch) = ch.chars().last() {
```
evilpie, line 320: Didn't we use the first character before?

kjarosh: We weren't consistent:

https://github.com/ruffle-rs/ruffle/pull/19616/files#diff-7be54fe026a0ed3dc2fc1fe118b9028a8c46085c0e5df2a54087572ebbfa9666L219

https://github.com/ruffle-rs/ruffle/pull/19616/files#diff-62c679f6eccebf339181274aa5cd20101388c20fca85d34ccf5ef7e9c016cb77L163

I think getting the last char makes more sense in case of dead keys until we start supporting them properly.

web/src/input.rs
```diff
@@ -1,125 +1,159 @@
-use ruffle_core::events::{KeyCode, TextControlCode};
+use ruffle_core::events::{
+    KeyDescriptor, KeyLocation, LogicalKey, NamedKey, PhysicalKey, TextControlCode,
+};
+use web_sys::KeyboardEvent;
 
-/// Convert a web `KeyboardEvent.code` value into a Ruffle `KeyCode`.
```
evilpie, line 3: This comment seems at least somewhat useful with s/KeyCode/PhysicalKey/.

web/src/input.rs
```diff
@@ -1,125 +1,159 @@
...
-        "F17" => KeyCode::F17,
-        "F18" => KeyCode::F18,
-        "F19" => KeyCode::F19,
-        "F20" => KeyCode::F20,
-        "F21" => KeyCode::F21,
-        "F22" => KeyCode::F22,
-        "F23" => KeyCode::F23,
-        "F24" => KeyCode::F24,
-        _ => KeyCode::UNKNOWN,
+        "Backquote" => PhysicalKey::Backquote,
```
evilpie, line 19: Just curious: I tried to figure out where this order comes from and I think it's based on the code in e.g. https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_code_values#code_values_on_windows. So why is the first not Escape?

web/src/input.rs
```diff
@@ -128,18 +162,106 @@ pub fn web_key_to_codepoint(key: &str) -> Option<char> {
     let (c1, c2) = (chars.next(), chars.next());
     if c2.is_none() {
```
evilpie, line 163:
```suggestion
    if let None = c1 {
      LogicalKey::Unknown
    } else if let (Some(ch), None) =  (c1, c2) {
      // Single key
    
```

Just an idea

evilpie:
It is probably interesting to mention that this adds not yet used support for mapping physical keys to Flash's virtual key code.

core/src/input.rs
```diff
@@ -126,6 +128,7 @@ impl InputManager {
                     InputEvent::KeyDown {
                         key_code: *key_code,
                         key_char: None,
+                        key_location: KeyLocation::Standard,
```
evilpie, line 131: We should probably add a TODO for setting these to something reasonable.

evilpie (approved):
Great work.

I feel like If we we also used winit as an abstraction for key handling on the web we could probably unify a lot of the code between desktop and web.

kjarosh:
> I feel like If we we also used winit as an abstraction for key handling on the web we could probably unify a lot of the code between desktop and web.

That is very true. Web and winit have compatible interfaces (winit basically uses the web interface in Rust), but as long as we don't use winit on web, we have to duplicate this interface :/

core/src/input.rs
kjarosh: A good idea, although I think the standard location is very reasonable. It probably won't be left/right because those keys do not have left/right counterparts, it probably won't be numpad because duh, and it won't be dpad because it's AIR only.

web/src/input.rs
kjarosh: Oh I had no reason to use Windows-specific docs :D The order is based on the spec: https://w3c.github.io/uievents-code/#code-value-tables

First it's categorized by section, then by the key's physical location, the order should be the same as in the `PhysicalKey` enum (you can see the sections there).

web/src/input.rs
kjarosh: Added comments

web/src/input.rs
kjarosh: I like that! Fixed

core/src/input.rs
kjarosh: Done

core/src/events.rs
```diff
@@ -830,3 +828,302 @@ impl FromStr for GamepadButton {
         })
     }
 }
+
+#[derive(Debug, Clone, Copy)]
+pub struct KeyDescriptor {
+    pub physical_key: PhysicalKey,
+    pub logical_key: LogicalKey,
+    pub key_location: KeyLocation,
+}
```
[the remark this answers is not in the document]
kjarosh: IMO when it comes to a simple, custom keyboard, you should use an approach similar to one used in tests, where each key has an identifier and is mapped into `PhysicalKey`/`LogicalKey`/`KeyLocation`. For more advanced keyboards I would expect all 3 fields to come from the virtual keyboard.

Maybe we can explore the idea of using the system keyboard on Android?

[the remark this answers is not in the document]
torokati44: For pure text entry, sure, it would be great, but not for playing games, and such.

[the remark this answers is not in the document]
torokati44: Bur yeah, making each key have a proper identifier in its `tag` attribute, and carrying that string all the way into Rust, then dispatching on it there (turning it into a `KeyIdentifier`) sounds doable. But passing strings through the JNI FFI boundary is just so much more inconvenient than simple scalar values IIRC... 😩 Sire it's not impossible, just less nice...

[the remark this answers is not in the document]
kjarosh: You can always map it in Java and pass numerical values down. Not sure if that would require code duplication or not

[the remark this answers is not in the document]
torokati44: Yeah but that would just require more messing around with encoding different (sometimes optional) info about all the keys into numbers and then those into Rust enumerants... This way it would be in one place at least.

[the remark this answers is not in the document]
torokati44: Well, it would still require separating the methods to pass in the key events coming from our handmade virtual keyboard (with the string identifiers), and those coming from an actual physical keyboard connected say via Bluetooth (or USB or adb/scrcpy or whatever), in the form of an Android `KeyEvent`. Oh well, this isn't a big deal either I guess...

PR #19629, opened 2025-02-25
web: Bump vite to 6.2.0

torokati44:
This lets us drop `esbuild <0.25`, resolving https://github.com/ruffle-rs/ruffle/security/dependabot/93.
See also: https://github.com/evanw/esbuild/security/advisories/GHSA-67mh-4wv8-2f99 and https://github.com/vitejs/vite/security/advisories/GHSA-vg6x-rcgg-rjx6

Although it probably didn't actually affect us (https://github.com/vitejs/vite/pull/19389#issuecomment-2651836759), it's still nice to get rid of that annoying notice about it. And it's a dep deduplication anyway, which is always appreciated!

Changelog (nothing major, from what I can see): https://github.com/vitejs/vite/blob/v6.2.0/packages/vite/CHANGELOG.md

_Doing Dependabot's job again... 😤 https://github.com/ruffle-rs/ruffle/actions/runs/13518806812/job/37773229925_


danielhjacobs (approved):

torokati44:
@danielhjacobs Actually, https://github.com/ruffle-rs/ruffle/pull/19628 includes this, so if you think that tiny manual fix in there is fine, it might be better to go with that instead

torokati44:
These changes were merged as part of the PR linked above. Thanks anyway for looking at it!

PR #19645, opened 2025-02-27
Make the blur filter more accurate (Close #19516)

### LINKED ISSUE #19516: Multiple layered movieclips with filters looks different than in flash player
### Describe the bug

It looks like this in ruffle latest nightly:
Image

### Expected behavior

In flash player standalone:
Image

### Content Location

Code and swf file available in:
https://github.com/xrtxn/ruffle_bug_report

### Affected platform

Desktop app

### Operating system

Artix linux

### Browser

_No response_

### Additional information

For some reason in the game it looks like this, with a black background, but this may be because of a sideeffect.
Image

xrtxn:
Fixes #19516
Honestly, I got really lucky to find this. I would love to see how this affects other flash projects.

### DIFF
--- a/render/wgpu/src/filters/blur.rs
+++ b/render/wgpu/src/filters/blur.rs
@@ -254,7 +254,7 @@ impl BlurFilter {
                 // and the last one with a weight of alpha. Note that the offset is relative
                 // to the center of the next-to-last sampled pixel, in the range of 0 to 0.5.
                 let last_offset = 1.0 / ((1.0 / alpha) + 1.0);
-                let last_weight = alpha + 1.0;
+                let last_weight = full_size - 2.0 * m - alpha;
 
                 let uniform = BlurUniform {
                     direction: if horizontal {


kjarosh:
Thanks for your contribution! I think we need a test that proves this behavior (which will differentiate between 2.0 and 1.99). How did you get the value 1.99?

> Should I rewrite the tests? They look fine visually, but obviously they fail.

Depends on the test. Generally we require an output from Flash Player, but some outputs might be generated from Ruffle, not FP (e.g. shumway tests that are animated).

I'll try looking into it and proposing something.

xrtxn:
I just tested it with a color picker. I tested multiple values, 1.9 was inaccurate, 1.99 looked like a good value to me. I created an example, you can check out https://github.com/xrtxn/ruffle_bug_report master or the old branch.
What was noticable to me, even with strength 1 the middle of the blurred image is #FBFEFB while with the flash player it is #FFFFFF. And the greater the strength was, the bigger the difference. 
With this change, the ruffle behavior is #FFFFFF correctly.

torokati44:
That 2 right there is in the "spatial dimension", so to say. So I'd look for (or put) this slight adjustment somewhere else. It's probably a rounding issue, we may have to put a `floor` call somewhere...?

torokati44:
That 2 right there is in the "spatial dimension", so to say. I'd look for (or put) this slight adjustment somewhere else. Perhaps a `floor` call should be added somewhere later on the color values...?

kjarosh:
Yeah, changing 2 to 1.99 does not fully fix the issue, as it still appears for different window sizes. It's most probably a rounding / numerical stability issue somewhere.

xrtxn:
The blur filter's box kernel weights don't sum to exactly `full_size` due to a floor rounding error in the `alpha` computation. Since the shader normalizes by dividing by `full_size`, the result is always slightly dimmed. This is visible in GlowFilter as slightly different colours.

Example calculations:
full_size | radius | m | alpha (floor'd) | last_weight (old) | Total (old) | last_weight (new) | Total (new) |
|---|---|---|---|---|---|---|---|
| 6.0 | 2.5 | 2.0 | 0.498 | 1.498 | 5.996 | 1.502 | 6.000 |
| 1.5 | 0.25 | 0.0 | 0.247 | 1.247 | 1.494 | 1.253 | 1.500 |
| 5.0 | 2.0 | 1.0 | 1.0 | 2.0 | 5.000 | 2.0 | 5.000

@torokati44, @kjarosh could you take a look? I'm still getting some test errors, but these may just be that we used a different calculation before.

On the left top is the old ruffle version, bottom left is the new ruffle version and on the right is the original flash version.
Screenshot_20260624_192835

torokati44:
Ooh, nice find and analysis!
Referencing https://github.com/ruffle-rs/ruffle/pull/12225.
I can't look into it much deeper _right now_, but I like this a lot more than the 1.99 solution earlier, and will keep it in mind to review!

EDIT: _[obligatory meme](https://knowyourmeme.com/memes/bojler-elado)_

torokati44:
Maybe `last_offset` will also have to be tweaked somewhat (maybe computed back from `last_weight`), to keep the `... so that the next-to-last will end up with an effective weight of 1.0` from the comment true.

torokati44:
_(or maybe last_weight should also be rounded/quantized up/down somehow. or maybe the "imperfect" total weight should also be passed in to the shader, so it can divide by that at the end. or maybe alpha shouldn't be quantized either, but I probably put it there for a reason - sorry, these are just quick random notes before I forget them.)_

torokati44:
(I did not forget about this, just been busy, sorry!)

xrtxn:
Yes, neither did I, unfortunately I have very little clue how to do this properly, and what should it be based on.

PR #19650, opened 2025-02-27
avm2: Add System memory-related APIs

adrian17:
I wanna add a minimal test to this first.

Also docs say that some of these are from FP10.1+, but playerglobals doesn't have any version marks; need to double-check that.


adrian17:
Looks like they indeed aren't version checked. Guessing that if we really wanted to, we could gate them by FP version, but we don't do that for most other APIs so 🤷

core/src/avm2/globals/flash/system/System.as
```diff
@@ -15,5 +16,24 @@ package flash.system {
...
+        }
+
+        public static function get privateMemory(): Number {
+            stub_getter("flash.system.System", "privateMemory");
+            return 1024*1024*100;
+        }
+
+        public static function get totalMemoryNumber(): Number {
+            stub_getter("flash.system.System", "totalMemoryNumber");
+            return 1024*1024*100;
```
evilpie, line 32: This seems extremely low?

adrian17: `totalMemory` isn't system memory, but - in my understanding - memory currently reserved for FP specifically. This grows over time as FP allocates more.
For example in FP's root DO constructor, I'm getting:
```
privateMemory:
10 825 728
totalMemory:
5 398 528
totalMemoryNumber:
5 439 488
freeMemory:
2 744 320
```
So I'd say it's actually more than you'll see in an average game.

evilpie: Oh, that is on me for not reading the documentation. I do think privateMemory should probably be a bit higher than totalMemory, just based on this quote for totalMemory:
> This property does not return all memory used by an Adobe AIR application or by the application (such as a browser) containing Flash Player content. The browser or operating system may consume other memory. The System.privateMemory property reflects all memory used by an application.

core/src/avm2/globals/flash/system/System.as
```diff
@@ -15,5 +16,24 @@ package flash.system {
         public static function disposeXML(node:XML):void {
             stub_method("flash.system.System", "disposeXML");
         }
+
+        public static function get freeMemory(): Number {
+            stub_getter("flash.system.System", "freeMemory");
+            return 1024*1024*10;
```
evilpie, line 22: Can you also add comments like `// 10 MiB`. Makes it a lot easier to read.

evilpie (approved):

PR #19348, opened 2025-01-26
web: Fix extension not working in Safari 16 (close #10981)

thy486:
If there are no other changes in Safari 16+, this PR will fix #10981

PR #5017, opened 2021-08-06
web:  Reland "Simplify .wasm loading"

relrelb:
In #4922 I tried to simplify the .wasm loading by using wasm-bindgen's built-in loader instead of relying on Webpack.
This allows to no longer declare .wasm files as resource assets in each `webpack.config.js`.

However it turned out to be problematic because it stopped respecting the `publicPath` config for the .wasm location, and instead assumed it is always in the same directory as the .js, which is not always the case. So the change was reverted in #4936.

This is a redo of the change, with fixes - the `publicPath` config is taken into account in `load-ruffle.ts`, which now sets `__webpack_public_path__` instead of each package (selfhosted, extension, demo).
However this removes the `publicPaths` config, but it should be fine as long as nobody uses it.

PR #19664, opened 2025-03-01
web: Fix extension webpack import not working (close #10981)

### LINKED ISSUE #10981: Extension does not work in Safari 16
### Describe the bug

Several users have reported that the Ruffle extension does not work in Safari 16. The extension popup always says "Checking Ruffle status on current tab..." and the Ruffle player does not appear on the page. Initial investigation by @kmeisthax suggests that there was some change in Safari 16 to how `postMessage` works, which was not mentioned in the release notes.

### Expected behavior

The extension should work as it does in other browsers.

### Affected platform

Browser's extension

### Operating system

MacOS Monterey, Big Sur, and Ventura

### Browser

Safari 16

### Additional information

_No response_

thy486:
Close #10981 
Continue with PR #19348, this PR is used to fix the `ruffle.js` webpack import module.

### DIFF
--- a/web/packages/core/src/current-script.ts
+++ b/web/packages/core/src/current-script.ts
@@ -10,7 +10,7 @@ try {
         "src" in document.currentScript &&
         document.currentScript.src !== ""
     ) {
-        let src = document.currentScript.src;
+        let src = getScriptOriginalSrc(document.currentScript);
 
         // CDNs allow omitting the filename. If it's omitted, append a slash to
         // prevent the last component from being dropped.
@@ -23,4 +23,32 @@ try {
     }
 } catch (_e) {
     console.warn("Unable to get currentScript URL");
+}
+
+/**
+ *
+ * Obtain the origin src content according to the running environment of the \<script\> node
+ *
+ * @param script \<script\> node instance
+ *
+ * @returns \<script\> node origin src
+ */
+function getScriptOriginalSrc(script: HTMLScriptElement): string {
+    const scriptUrl = script.src;
+    const scriptUrlPolyfill = script.getAttribute("ruffle-src-polyfill");
+    // Reset webkit mask url should be safe
+    if (scriptUrlPolyfill && "webkit-masked-url://hidden/" === scriptUrl) {
+        try {
+            const currentPolyfillURL = new URL(".", scriptUrlPolyfill);
+            const isExtensionUrl =
+                currentPolyfillURL.protocol.includes("extension");
+            // Only apply to extension
+            if (isExtensionUrl) {
+                return scriptUrlPolyfill;
+            }
+        } catch {
+            // Fallback to itself src
+        }
+    }
+    return scriptUrl;
 }
--- a/web/packages/extension/src/ruffle.ts
+++ b/web/packages/extension/src/ruffle.ts
@@ -28,8 +28,9 @@ function handleMessage(message: Message) {
 }
 
 let ID: string | null = null;
-if (document.currentScript !== undefined && document.currentScript !== null) {
-    if ("src" in document.currentScript && document.currentScript.src !== "") {
+if (document.currentScript instanceof HTMLScriptElement) {
+    polyfillCurrentScriptWebpackPublicPath();
+    if (document.currentScript.src) {
         try {
             ID = new URL(document.currentScript.src).searchParams.get("id");
         } catch (_) {
@@ -37,7 +38,7 @@ if (document.currentScript !== undefined && document.currentScript !== null) {
         }
     }
     if (ID === null) {
-        // if `script.src` is masked, get id from attrs
+        // Fallback to get id from attrs
         const ruffleId = document.currentScript.getAttribute("ruffle-id");
         if (ruffleId) {
             ID = ruffleId;
@@ -55,6 +56,44 @@ function openInNewTab(swf: URL): void {
         },
     };
     window.postMessage(message, "*");
+}
+
+/**
+ * We are overriding the publicPath automatically configured by webpack here because the browser might mask the script's src attribute (for example, Safari 16+ may mask the script.src of an extension as webkit-masked-url://hidden/).
+ */
+function polyfillCurrentScriptWebpackPublicPath(): void {
+    const script = document.currentScript as HTMLScriptElement;
+    try {
+        const scriptUrlPolyfill = script.getAttribute("ruffle-src-polyfill");
+        if (!scriptUrlPolyfill) {
+            return;
+        }
+        const scriptUrl = script.src;
+        const scriptAutoPublicPath =
+            getWebpackPublicPathFromScriptSrc(scriptUrl);
+
+        if (
+            scriptUrl === "webkit-masked-url://hidden/" &&
+            __webpack_public_path__ === scriptAutoPublicPath
+        ) {
+            const polyfillPath =
+                getWebpackPublicPathFromScriptSrc(scriptUrlPolyfill);
+            // TODO: If there are other scripts that need to be dynamically created and executed, the current polyfill logic should also be applied.
+            __webpack_public_path__ = polyfillPath;
+        }
+        // TODO: Process other situations when mask url not webkit-masked-url://hidden/
+    } catch (_) {
+        // Continue to run
+    }
+}
+
+// Copied from Webpack: https://github.com/webpack/webpack/blob/f1bdec5cc70236083e45b665831d5d79d6485db7/lib/runtime/AutoPublicPathRuntimeModule.js#L75
+function getWebpackPublicPathFromScriptSrc(scriptUrl: string): string {
+    return scriptUrl
+        .replace(/^blob:/, "")
+        .replace(/#.*$/, "")
+        .replace(/\?.*$/, "")
+        .replace(/\/[^/]+$/, "/");
 }
 
 if (ID) {
--- a/web/packages/extension/src/content.ts
+++ b/web/packages/extension/src/content.ts
@@ -80,6 +80,9 @@ function injectScriptURL(url: string): Promise<void> {
     script.src = url;
     // safari 16+ script.src will be masked to "webkit-masked-url://hidden/"
     script.setAttribute("ruffle-id", String(ID));
+    if (url !== script.src) {
+        script.setAttribute("ruffle-src-polyfill", url);
+    }
     (document.head || document.documentElement).append(script);
     return promise;
 }
--- a/web/packages/extension/src/global.d.ts
+++ b/web/packages/extension/src/global.d.ts
@@ -1,0 +1,8 @@
+declare global {
+    /**
+     * https://webpack.js.org/guides/public-path/#on-the-fly
+     */
+    let __webpack_public_path__: string;
+}
+
+export {};


web/packages/extension/src/ruffle.ts
```diff
@@ -57,6 +58,60 @@ function openInNewTab(swf: URL): void {
...
+
+// Exclude types that do not have this prop
+function determineKnownTypeHasNoEmptyStringProp<
+    T extends object,
+    TKey extends PropertyKey,
+>(obj: T, prop: TKey): obj is T extends Record<TKey, string> ? T : never {
+    return prop in obj && (obj as Record<PropertyKey, unknown>)[prop] !== "";
+}
+
+// Copied from Webpack: https://github.com/webpack/webpack/blob/f1bdec5cc70236083e45b665831d5d79d6485db7/lib/runtime/AutoPublicPathRuntimeModule.js#L75
```
thy486, line 106: This link may be confusing.
It actually comes from [webpack automatic publicPath](https://webpack.js.org/guides/public-path/#automatic-publicpath). Specifically refers to get publicPath from the `currentScript.src`.

web/packages/extension/src/ruffle.ts
```diff
@@ -57,6 +58,60 @@ function openInNewTab(swf: URL): void {
     window.postMessage(message, "*");
 }
 
+function polyfillCurrentScript(): void {
```
thy486, line 61: I found the relevant code that has used currentScript.src in these files and make polyfills for these files.
- [web/packages/extension/src/ruffle.ts](https://github.com/ruffle-rs/ruffle/blob/85d8b2acb485bd2da6e2bc645f102db705302eb4/web/packages/extension/src/ruffle.ts#L34)
- [web/packages/core/src/current-script.ts](https://github.com/ruffle-rs/ruffle/blob/85d8b2acb485bd2da6e2bc645f102db705302eb4/web/packages/core/src/current-script.ts#L13)
- ~~[web/packages/selfhosted/js/ruffle.js](https://github.com/ruffle-rs/ruffle/blob/85d8b2acb485bd2da6e2bc645f102db705302eb4/web/packages/selfhosted/js/ruffle.js#L15)~~ (Intentionally ignored because the extension never use it)

web/packages/extension/src/ruffle.ts
```diff
@@ -57,6 +58,60 @@ function openInNewTab(swf: URL): void {
...
+        const webpackCurrentPublicPath = __webpack_public_path__;
+        const scriptAutoPublicPath =
+            getWebpackPublicPathFromScriptSrc(scriptUrl);
+        if (webpackCurrentPublicPath === scriptAutoPublicPath) {
+            applyResetWebpackPublicPath(
+                getWebpackPublicPathFromScriptSrc(scriptUrlPolyfill),
+            );
+        }
+    }
+    // TODO: Process other situations when mask url not webkit-masked-url://hidden/
```
thy486, line 95: Since the current problem is relatively easy to fix, I made a simple fix. But this may not be the best or most general solution, so I left a todo here

web/packages/extension/src/ruffle.ts
```diff
@@ -57,6 +58,60 @@ function openInNewTab(swf: URL): void {
...
+    script: HTMLScriptElement,
+): void {
+    const scriptUrlPolyfill = script.getAttribute("ruffle-src-polyfill");
+    if (!scriptUrlPolyfill) {
+        return;
+    }
+    const scriptUrl = script.src;
+    const applyResetWebpackPublicPath = (overridePublicPath: string): void => {
+        __webpack_public_path__ = overridePublicPath;
+        // TODO: If there are other scripts that need to be dynamically created and executed, the current polyfill logic should also be applied.
```
thy486, line 82: Perhaps there is a better way for us to notify downstream of such fixes?

web/packages/extension/src/ruffle.ts
```diff
@@ -57,6 +58,60 @@ function openInNewTab(swf: URL): void {
...
+    };
+    // Reset webpack public path should be safe when it is using mask url
+    if ("webkit-masked-url://hidden/" === scriptUrl) {
+        const webpackCurrentPublicPath = __webpack_public_path__;
```
thy486, line 86: The `__webpack_public_path__` runtime code is actually like `__webpack_require__.p`, so it could be set and get. But this is a webpack only logic.

[Related documents](https://webpack.js.org/guides/public-path/#on-the-fly)

web/packages/core/src/current-script.ts
```diff
@@ -24,3 +24,34 @@ try {
...
+ * @returns \<script\> node origin src
+ */
+function getScriptOriginalSrc(script: HTMLScriptElement): string {
+    const scriptUrl = script.src;
+    const scriptUrlPolyfill = script.getAttribute("ruffle-src-polyfill");
+    if (!scriptUrlPolyfill) {
+        return scriptUrl;
+    }
+    // Reset webkit mask url should be safe
+    if ("webkit-masked-url://hidden/" === scriptUrl) {
```
danielhjacobs, line 43: It feels slightly unnecessary to have the early return above when this condition can just be:

`if (scriptUrlPolyfill && "webkit-masked-url://hidden/" === scriptUrl)`

web/packages/extension/src/ruffle.ts
```diff
@@ -29,15 +29,16 @@ function handleMessage(message: Message) {
 
 let ID: string | null = null;
 if (document.currentScript !== undefined && document.currentScript !== null) {
-    if ("src" in document.currentScript && document.currentScript.src !== "") {
+    polyfillCurrentScript();
+    if (determineKnownTypeHasNoEmptyStringProp(document.currentScript, "src")) {
```
danielhjacobs, line 33: Is this extra function really necessary? Is it not just `document.currentScript?.src`?

web/packages/extension/src/ruffle.ts
```diff
@@ -57,6 +58,60 @@ function openInNewTab(swf: URL): void {
...
+    try {
+        if (determineKnownTypeHasNoEmptyStringProp(script, "src")) {
+            determineScriptIsMaskedSrcAndMakePolyfill(script);
+        }
+    } catch (_) {
+        // Continue to run
+    }
+}
+
+function determineScriptIsMaskedSrcAndMakePolyfill(
```
danielhjacobs, line 72: Can we rename this to `applyPolyfillIfNeeded` or something else that's short?

web/packages/extension/src/ruffle.ts
```diff
@@ -57,6 +58,60 @@ function openInNewTab(swf: URL): void {
     window.postMessage(message, "*");
 }
 
+function polyfillCurrentScript(): void {
+    const script = document.currentScript!;
```
danielhjacobs, line 62: Instead of using a complicated new function below (`determineKnownTypeHasNoEmptyStringProp`), couldn't we just replace this with this code?

```js
    const script = document.currentScript as HTMLScriptElement | null;
    if (!script?.src) return;
```

web/packages/extension/src/ruffle.ts
```diff
@@ -57,6 +58,60 @@ function openInNewTab(swf: URL): void {
...
+            applyResetWebpackPublicPath(
+                getWebpackPublicPathFromScriptSrc(scriptUrlPolyfill),
+            );
+        }
+    }
+    // TODO: Process other situations when mask url not webkit-masked-url://hidden/
+}
+
+// Exclude types that do not have this prop
+function determineKnownTypeHasNoEmptyStringProp<
```
danielhjacobs, line 99: I think we can remove this function, it seems unnecessary.

web/packages/extension/src/ruffle.ts
```diff
@@ -57,6 +58,60 @@ function openInNewTab(swf: URL): void {
...
+
+function determineScriptIsMaskedSrcAndMakePolyfill(
+    script: HTMLScriptElement,
+): void {
+    const scriptUrlPolyfill = script.getAttribute("ruffle-src-polyfill");
+    if (!scriptUrlPolyfill) {
+        return;
+    }
+    const scriptUrl = script.src;
+    const applyResetWebpackPublicPath = (overridePublicPath: string): void => {
```
danielhjacobs, line 80: I don't think we need an inline function here.

web/packages/extension/src/ruffle.ts
```diff
@@ -57,6 +58,60 @@ function openInNewTab(swf: URL): void {
...
+}
+
+function determineScriptIsMaskedSrcAndMakePolyfill(
+    script: HTMLScriptElement,
+): void {
+    const scriptUrlPolyfill = script.getAttribute("ruffle-src-polyfill");
+    if (!scriptUrlPolyfill) {
+        return;
+    }
+    const scriptUrl = script.src;
```
danielhjacobs, line 79: Starting here, up to the end of the function, I think we can just replace this with something much simpler:

```js
    const scriptAutoPublicPath = getWebpackPublicPathFromScriptSrc(scriptUrl);
    const polyfillPath = getWebpackPublicPathFromScriptSrc(scriptUrlPolyfill);

    if (__webpack_public_path__ === scriptAutoPublicPath) {
        // TODO: If there are other scripts that need to be dynamically created and executed, the current polyfill logic should also be applied.
        __webpack_public_path__ = polyfillPath;
    }
    // TODO: Process other situations when mask url not webkit-masked-url://hidden/
```

web/packages/extension/global.d.ts
```diff
@@ -0,0 +1,8 @@
+declare global {
+    /**
+     * https://webpack.js.org/guides/public-path/#on-the-fly
+     */
+    let __webpack_public_path__: string;
```
danielhjacobs, line 5: We used to have many more references to `__webpack_public_path__` in the codebase, but several were removed by #5017. I suppose we likely can't help adding this, but I though it worth noting that.

danielhjacobs (changes requested):
Note: My review comes from the perspective of suggesting simplications to the added code, without actually considering how much of the added code is necessary. My knowledge of webpack is unfortunately limited, so hopefully someone else can review from that perspective.

web/packages/extension/global.d.ts
```diff
@@ -0,0 +1,8 @@
+declare global {
```
danielhjacobs, line 1: Isn't this file meant to be in in the folder src, named globals.d.ts? If it is, I don't think the tsconfig.json would need changes.

web/packages/extension/src/ruffle.ts
thy486: `document.currentScript` in **TypeScript** is `HTMLOrSVGScriptElement(SVGScriptElement | HTMLScriptElement) | null`. We cannot directly access the src attribute in **TypeScript** because the type `SVGScriptElement` does not have this attribute. So I write this function according to **TypeScript** type constraints, which may be over-designed.

danielhjacobs: Can you instead do `script instanceof HTMLScriptElement && `?

web/packages/extension/src/ruffle.ts
danielhjacobs: Edited according to your other comment:

```js
    const script = document.currentScript;
    if (!(script instanceof HTMLScriptElement) || !script.src) return;
```

web/packages/extension/src/ruffle.ts
evilpie: We should definitely check to make sure that currentScript is a HTMLScriptElement, because it prevents DOM clobbering attacks. https://github.com/whatwg/html/issues/10687

web/packages/core/src/current-script.ts
```diff
@@ -24,3 +24,34 @@ try {
...
+        try {
+            const currentPolyfillURL = new URL(".", scriptUrlPolyfill);
+            const isExtensionUrl =
+                currentPolyfillURL.protocol.includes("extension");
+            // Only apply to extension
+            if (isExtensionUrl) {
+                return scriptUrlPolyfill;
+            }
+        } catch (_) {
```
evilpie, line 52:
```suggestion
        } catch {
```

thy486:
Thank you very much for the review. I have modified my code as required and re-pushed it. Please review it again when you have time

web/packages/extension/src/ruffle.ts
evilpie: I find the name confusing. We aren't polyfilling `currentScript`, but something webpack related?

web/packages/extension/src/ruffle.ts
```diff
@@ -29,15 +29,19 @@ function handleMessage(message: Message) {
 
 let ID: string | null = null;
 if (document.currentScript !== undefined && document.currentScript !== null) {
```
evilpie, line 31: Let's change this to `document.currentScript instanceof HTMLScriptElement` and remove that check from the rest of the code.

web/packages/extension/src/ruffle.ts
evilpie: Actually can you please add a comment outlining what we are doing here?

web/packages/extension/src/ruffle.ts
thy486: I think there is nothing wrong with checking null first and then determining the variable type. And in mdn, its syntax expects object as the variable to be checked.

danielhjacobs: > its syntax expects object as the variable to be checked.

That is true, but, strange as it may be, `typeof null` is `"object"`, meaning `document.currentScript` is an "object" regardless if it's `null` or not: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#typeof_null

danielhjacobs: That said, my opinion is that either way is fine. You'll get the same result either way.

Though, even if you prefer keeping the explicit null check, is there a reason not to move the `document.currentScript instanceof HTMLScriptElement` check up here after an `&&` and remove it later?

evilpie: That is actually not the reason why this works. `instanceof` with a primitive (non-object) on the left hand side just always* returns false. (* There is some Symbol.hasInstance weirdness)

https://tc39.es/ecma262/#sec-ordinaryhasinstance

web/packages/extension/src/ruffle.ts
thy486: I changed the function name `polyfillCurrentScript` to `polyfillCurrentScriptWebpackPublicPath`

danielhjacobs: Can you also add a comment as suggested? If possible, I'd suggest it describe both what we are doing **and** why we need to do it.

thy486: > Can you also add a comment as suggested? If possible, I'd suggest it describe both what we are doing **and** why we need to do it.

https://github.com/ruffle-rs/ruffle/blob/949eaf27e3c17be175856b618d7fcfc57db13e9e/web/packages/extension/src/ruffle.ts#L62

web/packages/extension/src/ruffle.ts
```diff
@@ -57,6 +58,41 @@ function openInNewTab(swf: URL): void {
...
+    try {
+        const scriptUrlPolyfill = script.getAttribute("ruffle-src-polyfill");
+        if (!scriptUrlPolyfill) {
+            return;
+        }
+        const scriptUrl = script.src;
+        const scriptAutoPublicPath =
+            getWebpackPublicPathFromScriptSrc(scriptUrl);
+
+        if (__webpack_public_path__ === scriptAutoPublicPath) {
```
danielhjacobs, line 75: Do we need this check by the way? I just got to the point of understanding the logic, and while this check is probably not harmful, do we ever expect it to be false?

danielhjacobs:
Let me see if I understand the logic:

When we do `script.src = url`, Safari masks that to `"webkit-masked-url://hidden/"`.

The previous change attempted a targeted fix: the only reason we need to know `script.src` is because the `url` has a query parameter called `ID` that needs to be used in the `postMessage`. You worked around that by setting an attribute on the script called `ruffle-id` and using that.

And we need this change because:

webpack's code for the automatic public path also uses the `script.src`. That `src` is `"webkit-masked-url://hidden/"` so that's what webpack tries to load.

So the new code says:

If `script.src = url` and then immediately `script.src !== url`, `ruffle-src-polyfill = url`. Safari doesn't mask that because it's not a `script.src`.

`getScriptOriginalSrc` gets `ruffle-src-polyfill`, which is `url`, if `script.src = "webkit-masked-url://hidden/", otherwise it gets `script.src`.

if `script.src !== url` and `__webpack_public_path__ === scriptAutoPublicPath`, `__webpack_public_path__` is updated to the real `script.src` converted to a public path (https://github.com/webpack/webpack/blob/f1bdec5cc70236083e45b665831d5d79d6485db7/lib/runtime/AutoPublicPathRuntimeModule.js#L75).

web/packages/extension/src/ruffle.ts
danielhjacobs: And shoudl we have a check be `scriptUrl === "webkit-masked-url://hidden/"`?

thy486: > 我们是否应该添加一个检查`scriptUrl === "webkit-masked-url://hidden/"`？

Yep, we need. I lost it, sorry

thy486: > Do we need this check by the way? I just got to the point of understanding the logic, and while this check is probably not harmful, do we ever expect it to be false?

If it's not using the publicPath that webpack automatically set, here is false

thy486: > Do we need this check by the way? I just got to the point of understanding the logic, and while this check is probably not harmful, do we ever expect it to be false?

If it's not using the publicPath that webpack automatically set, here is false

danielhjacobs:
Please test today's nightly with Safari.  If it works, this PR can be closed.

thy486:
> 请使用 Safari 测试今天的夜间版本。如果可以，此 PR 可以关闭。
It works. Thank you for your work

thy486:
> Please test today's nightly with Safari. If it works, this PR can be closed.

It works. Thank you for your work.

danielhjacobs:
Superseded by #20264

Issue #1778, opened 2020-11-30
[web] There are some problems with text input boxes.


jooy0224:
Hello, thanks to the ruffle team and I'm posting an issue I found about the TextInput box.

1. Some text input does not work like Korean, Chinese, and Japanese (Using IME, I guess). It seems that only English is being entered.

2. Key actions such as Ctrl+A, Ctrl+C, and Ctrl+V do not work in the Text Input Box.
(It worked normally on Adobe flash player.)

Thanks!!

ousia:
I don’t know whether #2081 might fix this.

Jmousy:
The original issue was about shortcut behavior in text fields (e.g. copy, paste) and CJK input. The shortcut behavior is now implemented and i'll leave this issue open for IME support.

kjarosh:
Implementation status:

- [x] Basic IME support https://github.com/ruffle-rs/ruffle/pull/19666
- [ ] Preediting https://github.com/ruffle-rs/ruffle/pull/19888
- [ ] Visual indicator for preediting
- [ ] Deciding when IME should be enabled

jmousy:
Hello,
IME is a feature that many Asians need, and i'm so happy and grateful that you've taken an interest in it and started improve it.

I have questions about your work and future direction and leave a comment. (I apologize if it seems like i'm trying to pressure you,):

1. Is there any chance that IME will be supported in the future, not only on the desktop, but also in self-hosted web environments? Or would technical difficulties prevent it from being implemented at this point?

2. I also found that for some specialized flash games (AVM2), as shown below, you can enter Korean text even in the current Nightly version.
The input interface seemed to work fine, although the combination characters contained in Hangul were not combined correctly (e.g., the backspace character).

for example,

When you try to type '가', you need to type 'ㄱ' and 'ㅏ'. The behavior should be 'ㄱ' -> '가', but it is entered as 'ㄱ' -> 'ㄱ가'.
Similarly, when you try to use the 'ㅇ' base for something like '강', it is entered as 'ㄱ' -> 'ㄱ가' -> 'ㄱ가강'.

![Image](https://github.com/user-attachments/assets/c04e1563-298a-4566-bb58-267b932e1c4c)

Sample File: [흥해라편의점.zip](https://github.com/user-attachments/files/19406799/default.zip)

If this could be fixed, would it be possible to have Korean input in some Flash at this time? Or is this something that would need to be structurally redesigned?


Regards,

kjarosh:
[...]

PR #19666, opened 2025-03-01
desktop: Implement basic IME support

kjarosh:
* Progresses https://github.com/ruffle-rs/ruffle/issues/1778

The basic IME support includes displaying popups in roughly the right position, and handling IME commit events without preediting yet.

Tested manually with Anthy (no automated tests yet due to lack of preediting).

IME preference is also added to the GUI (disabled by default):
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4229,6 +4229,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "ashpd",
+ "async-channel",
  "bytemuck",
  "chrono",
  "clap",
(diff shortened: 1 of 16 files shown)


kjarosh:
> Can you describe why we need the notification sender mechanism? This seems like a more generic approach compared to what we have done before with e.g. the `DesktopUiBackend`.

That's a good point. I haven't thought of it this way, and the usage is similar to opening/closing the virtual keyboard. We need to inform the frontend whether we can or cannot accept IME events.

Shall I rewrite it to use the `DesktopUiBackend` or is this approach better?

kjarosh:
I think the main advantage of having the notification mechanism is that it's easier to test, we can include expected notifications next to events from `input.json`

evilpie:
There might be some stuff we could move to this mechanism.

evilpie (approved):

PR #19670, opened 2025-03-01
avm2: Store PerspectiveProjection to DisplayObject

cookie-s:
Continued from #19532 .

DisplayObject stores `Option<PerspectiveProjection>` through `Transform` so we can remove `HAS_PERSPECTIVE_PROJECTION_STUB` flag from display objects.

The test `from_shumway/avm2/flash/geom/perspectiveprojection` passes now.

### Limitations
PerspectiveProjection is stored in Transform, but it's not used in rendering at all. (Hence, we cannot remove `stub_setter`s)


tests/tests/swfs/from_shumway/avm2/flash/geom/perspectiveprojection/test.toml
```diff
@@ -1,2 +1,8 @@
 num_ticks = 1
-known_failure = true
\ No newline at end of file
+
+[approximations]
+number_patterns = [
+# Both FP 32,0,0,465 (for Linux) and FP 32 (for Windows) produced slightly different numbers in some environment. However, the original trace from mozilla/shumway is respected in the output file.
+    'focalLength: ([0-9.]+)'
+]
+max_relative = 0.0000001
```
cookie-s, line 8: I believe the `output.txt` is from https://github.com/mozilla/shumway/blob/master/test/swfs/avm2/flash/geom/perspectiveprojection/PerspectiveProjectionClass.swf.trace .
I couldn't reproduce exactly the same result from `PerspectiveProjectionClass.as` in the same directory (after recompilation with mxmlc) or even from the `test.swf` (without recompilation).


(Fun fact: I was able to pass this test without this approximation until https://github.com/ruffle-rs/ruffle/pull/19748 was merged -- What does it mean? Anyway, I think the problem is that I cannot reproduce the `output.txt` from ActionScript or SWF.)

core/src/avm2/globals/flash/geom/perspective_projection.rs
```diff
@@ -68,10 +63,167 @@ pub fn set_focal_length<'gc>(
...
+    let (x, y) = {
+        let center = this.get_slot(pp_slots::CENTER).as_object().unwrap();
+        let x = center
+            .get_slot(point_slots::X)
+            .coerce_to_number(activation)?;
+        let y = center
+            .get_slot(point_slots::Y)
+            .coerce_to_number(activation)?;
+        (x, y)
+    };
```
Lord-McSweeney, line 223: The uses of `coerce_to_number` here can be replaced with `as_f64`

core/src/avm2/globals/flash/geom/transform.rs
```diff
@@ -391,44 +425,25 @@ pub fn get_perspective_projection<'gc>(
...
+        .base()
+        .perspective_projection()
+        .is_some()
+    {
+        let object = activation
             .avm2()
             .classes()
             .perspectiveprojection
-            .construct(activation, &[])?;
+            .construct(activation, &[])?
```
Lord-McSweeney, line 437: Hmm, does this really return a new PerspectiveProjection each time the getter is called?

core/src/context.rs
```diff
@@ -395,6 +395,7 @@ impl<'gc> UpdateContext<'gc> {
         drop(activation);
 
         root.set_depth(self.gc(), 0);
+        root.set_perspective_projection(self.gc(), None); // Set default PerspectiveProjection
```
Lord-McSweeney, line 398: Is this line necessary?

core/src/display_object/stage.rs
```diff
@@ -193,6 +194,7 @@ impl<'gc> Stage<'gc> {
             },
         ));
         stage.set_is_root(gc_context, true);
+        stage.set_perspective_projection(gc_context, None); // Set default PerspectiveProjection
```
Lord-McSweeney, line 197: Is this line necessary?

tests/tests/swfs/from_shumway/avm2/flash/geom/perspectiveprojection/test.toml
Lord-McSweeney: I think a small difference like this isn't a problem, we have similar issues with some of the AVM2 Number tests already.

Lord-McSweeney (approved):

core/src/display_object/stage.rs
cookie-s: Yes, this line actually sets `PerspectiveProjection::default()` to DO's PerspectiveProjection.
https://github.com/ruffle-rs/ruffle/blob/aa6197e2261e6ab1ed24cfc64d11ecffae307065/core/src/display_object/stage.rs#L894-L896

Technically it's fine to set `PerspectiveProjection::default()` directly at here, but for consistency with `root.set_perspective_projection()` in a different file I'm passing `None` here.
https://github.com/ruffle-rs/ruffle/blob/aa6197e2261e6ab1ed24cfc64d11ecffae307065/core/src/context.rs#L397
The `root` actually has a different value used when `None` (`null`) is provided..
https://github.com/ruffle-rs/ruffle/blob/aa6197e2261e6ab1ed24cfc64d11ecffae307065/core/src/display_object/movie_clip.rs#L2836-L2845

core/src/avm2/globals/flash/geom/transform.rs
```diff
@@ -248,6 +250,38 @@ fn object_to_matrix3d<'gc>(
...
+        .get_slot(pp_slots::FOV)
+        .coerce_to_number(activation)?;
+
+    let center = object.get_slot(pp_slots::CENTER).as_object().unwrap();
+    let x = center
+        .get_slot(point_slots::X)
+        .coerce_to_number(activation)?;
+    let y = center
+        .get_slot(point_slots::Y)
+        .coerce_to_number(activation)?;
```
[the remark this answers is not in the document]
cookie-s: Fixed!

cookie-s (commented):
Thank you for review!

Dinnerbone:
I've fixed the rebase conflict.

Thank you so much!

PR #19716, opened 2025-03-05
core: Remove all static strings, shrink Value structs

Lord-McSweeney:
`AvmString` now directly holds a `Gc<AvmStringRepr>`, which shrinks Value structs on 64-bit from 24 to 16 bytes. getlocal, setlocal, and getslot run 5-10% faster; Chess Demons starts up about 5% faster.


core/src/avm2/globals.rs
```diff
@@ -595,8 +548,6 @@ pub fn load_player_globals<'gc>(
 
     globals.set_vtable(mc, global_obj_vtable);
 
-    activation.context.avm2.toplevel_global_object = Some(globals);
```
adrian17, line 596: To be sure, why was this moved earlier?

core/src/display_object/movie_clip.rs
```diff
@@ -2330,7 +2330,10 @@ impl<'gc> MovieClip<'gc> {
                 ClipEvent::BUTTON_EVENT_METHODS
                     .iter()
                     .copied()
-                    .any(|handler| object.has_property(&mut activation, handler.into()))
+                    .any(|handler| {
+                        let handler = AvmString::new_utf8(activation.gc(), handler);
```
adrian17, line 2334: This _is_ called quite often when you even just hover the mouse in avm1.
If you don't want it to be in common strings, how about `intern_static`?

core/src/avm2/domain.rs
```diff
@@ -290,7 +290,7 @@ impl<'gc> Domain<'gc> {
                 activation.gc(),
                 &name[(start + 2)..(name.len() - 1)],
             ));
-            name = "__AS3__.vec::Vector".into();
+            name = AvmString::new_utf8(activation.gc(), "__AS3__.vec::Vector");
```
adrian17, line 293: This could still be an AvmString-pointing-to-static, right?

core/src/avm2/globals/class.rs
```diff
@@ -100,8 +103,9 @@ pub fn create_c_class<'gc>(
     let gc_context = activation.gc();
     let namespaces = activation.avm2().namespaces;
 
+    let class_name = AvmString::new_utf8(gc_context, "Class$");
```
adrian17, line 106: Managed-pointing-to-static?

adrian17 (approved):

adrian17:
So a lot of these `AvmString::new_utf8(gc_context, "Class$");` could be replaced with managed-static strings, with the only caveat being that... we don't really have an API for these, aside from `intern_static` which also implies interning?
I'm okay with some kind of `new_xyz_static(gc, s: &'static str / &'static [u8])` being out of this PR, but let's make sure this is indeed what we want in the long run.

Also after this gets merged, I'll do an instrumented test to see what dynamic strings we now allocate from our own code during normal runtime.

core/src/avm2/globals.rs
Lord-McSweeney: We have three script global objects right now- one for the builtin classes script, one for `Toplevel.as`, and one for `globals.as`. The "toplevel" global object has pointed to the builtin one for a long time, but as we moved classes out of it and into `Toplevel.as` it makes more sense to make it point to the `Toplevel.as` global object.

The hacky part was that some tests depended on the "toplevel" global object having the `parseInt`, `parseFloat`, etc properties, which, since they were defined in `Toplevel.as`, we had to copy over to the builtin global object manually. Now that the only properties left on the builtin global object are Object and Class anyway, I made the "toplevel" global object point to `Toplevel.as` and removed the copying hack.

core/src/display_object/movie_clip.rs
Lord-McSweeney: I thought this was cold- I've changed it to use the common string store.

core/src/avm1/value.rs
```diff
@@ -29,13 +29,8 @@ pub enum Value<'gc> {
 }
 
 // This type is used very frequently, so make sure it doesn't unexpectedly grow.
-// On 32-bit x86 Android, it's 12 bytes. On most other 32-bit platforms it's 16.
-#[cfg(target_pointer_width = "32")]
 const _: () = assert!(size_of::<Value<'_>>() <= 16);
 
-#[cfg(target_pointer_width = "64")]
-const _: () = assert!(size_of::<Value<'_>>() == 24);
```
adrian17, line 37: Can you keep them separate, to be more explicit? Will be also useful if we ever do NaN boxing on 32-bit.

core/src/avm2/error.rs
```diff
@@ -38,12 +38,8 @@ impl Debug for Error<'_> {
 }
 
 // This type is used very frequently, so make sure it doesn't unexpectedly grow.
-#[cfg(target_family = "wasm")]
 const _: () = assert!(size_of::<Result<Value<'_>, Error<'_>>>() == 24);
```
adrian17, line 41: I thought that the size is smaller (16?) on non-wasm due to x86 being less strict with alignment?
At least at some point it was. Do we do any testing on 32-bit?

That's why I suggest keeping them separate here too.

adrian17 (approved):

torokati44:
The failure of the assertion about the size on x86_32 (i686) still blocks updating Ruffle in the Android app:
https://github.com/ruffle-rs/ruffle-android/actions/runs/13764900529/job/38488990037#step:6:174

Issue #19687, opened 2025-03-03
Huge release binary size


niansa:
The provided ruffle release binaries are really huge (40 MB for the Linux executable, still 30 MB after stripping!). I feel like this is pretty big, specially ẃhen considering that the Unity Engine Runtime is just a bit more than 20 MB (while supporting pretty extensive 3D graphics using 4 different graphics APIs)!

While the binary size itself isn't really an issue anymore these days this might point to some build system issue perhaps leading to unused components being built. Or perhaps the executable contains dependencies like curl that the system could easily be expected to already have installed? If so, it should be made sure that there is a build option for distributors to use dynamically linked libraries instead of statically linked libraries to avoid duplicate library installations (leading to much more load on their build and package repo servers).

Hope this helps improve Ruffle!

kjarosh:
We could enable LTO when building releases, it shrinks the binary size from 40.4M to 34M. We can shrink it further to 30.6M when stripping all symbols.

niansa:
30 MB is still really big but it is an improvement. However I believe there is some underlaying issue.

Croworbit:
for reference, adobe flash player 32 standalone projector is 16M
which with the optimized version of ruffle from above is slightly less than double the size
it may or may not be able to be improved upon

from the title of the issue, I might have assumed a build had suddenly become say, 100M, ~600% bigger, which might be able to be considered huge

if you can find or have found any underlaying build issues, feel free to share your findings! 😀

niansa:
I would assume though that specially the old compilers that Adobe Flash Player was compiled with are much worse at optimizing than a modern compiler like rustc/LLVM.

Looking at the very short `ldd` output of ruffle:

        linux-vdso.so.1 (0x00007fdd488f3000)
        libasound.so.2 => /lib/x86_64-linux-gnu/libasound.so.2 (0x00007fdd46812000)
        libudev.so.1 => /lib/x86_64-linux-gnu/libudev.so.1 (0x00007fdd467cd000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fdd467a0000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fdd466ba000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fdd464c4000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fdd488f5000)
[...]

PR #19727, opened 2025-03-06
chore: Reduce binary sizes in desktop releases

kjarosh:
Stripping symbols and enabling LTO makes the compiled artifacts smaller.

Desktop Linux shrinks from 40.4MB to 30.6MB.

* Related to https://github.com/ruffle-rs/ruffle/issues/19687

### DIFF
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -100,7 +100,7 @@ jobs:
 
     env:
       PACKAGE_FILE: ${{ needs.create-nightly-release.outputs.package_prefix }}-${{ matrix.build_name }}.${{ startsWith(matrix.build_name, 'win') && 'zip' || 'tar.gz' }}
-      CARGO_BUILD_DIR: target/${{ matrix.target }}/release
+      CARGO_BUILD_DIR: target/${{ matrix.target }}/release-small
 
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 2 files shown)


torokati44:
Don't we need them for stack traces in crash reports?

torokati44:
While on web, I can _sort of_ see the point of trying to stay smaller, but on Desktop, and honestly, even in the browser extension, or on mobile, it should basically bear _no_ weight IMHO, at least until we are not in the multi-hundred-MB territory _(looks at electron apps)_.

adrian17:
We also explicitly disabled stripping in wasm-opt to preserve symbols:

https://github.com/ruffle-rs/ruffle/blob/383497e2bf0099b1ed264701fbccb859a6502e7f/web/packages/core/tools/build_wasm.ts#L6

kjarosh:
> Don't we need them for stack traces in crash reports?

I believe setting `panic = "abort"` makes us skip stack traces, and testing shows that we don't include them on desktop.

> While on web, I can _sort of_ see the point of trying to stay smaller, but on Desktop, and honestly, even in the browser extension, or on mobile, it should basically bear _no_ weight IMHO, at least until we are not in the multi-hundred-MB territory _(looks at electron apps)_.

I made this PR was to test whether web works with symbols stripped, because I kinda suspected it will break wasm bindgen (which apparently is true). According to my understanding we have no reason to have symbols on desktop release.

kjarosh:
Superseded by:
* https://github.com/ruffle-rs/ruffle/pull/23714
* https://github.com/ruffle-rs/ruffle/pull/23863

PR #19748, opened 2025-03-08
avm2: Make PerspectiveProjection.focalLength more precise

cookie-s:
Small follow up to https://github.com/ruffle-rs/ruffle/pull/19532

The previous `focalLength` getter had a certain level of difference. This change reduces the difference from Flash Player.

Also, I found Ruffle `test.toml` has the configuration section for number approximations, so move the logic from `Test.as`.


core/src/avm2/globals/flash/geom/perspective_projection.rs
```diff
@@ -40,7 +40,7 @@ pub fn get_focal_length<'gc>(
     let fov = this.get_slot(pp_slots::FOV).coerce_to_number(activation)?;
 
     let width = get_width(activation, this);
-    let focal_length = (width / 2.0) / f64::tan(fov / 2.0 * DEG2RAD);
+    let focal_length = (width / 2.0) as f32 * f64::tan((PI - fov * DEG2RAD) / 2.0) as f32;
```
cookie-s, line 43: After many trials and observations, I found this code. It doesn't seem to be perfect yet (for edge values like 1e-10), though.

* `1 / tan x` = `cot x` = `tan (pi/2 - x)`
* `_ as f32 * f32::tan(_)` is not good. This produces better result.

tests/tests/swfs/avm2/perspective_projection/test.toml
```diff
@@ -1 +1,8 @@
 num_ticks = 1
+
+[approximations]
+epsilon = 0.0000000000001
+number_patterns = [
+    'fieldOfView = ([0-9.]+)',
+    'FOV [0-9]+ ([0-9.]+)',
```
cookie-s, line 7: field of view still has some errors.

```
var pp: PerspectiveProjection = new PerspectiveProjection();
pp.fieldOfView = 179.99999;
trace(pp.fieldOfView);
```
I found the code above gives `179.99999000000003`, so `pp.fieldOfView` is not even a simple storage variable.

cookie-s:
Rebased to have all tests pass.

tests/tests/swfs/avm2/perspective_projection/output.txt
kjarosh, line 1: I get slightly different values on FP 32,0,0,465 x86_64 Linux (but they are still within the epsilon). Which FP is it from?

tests/tests/swfs/avm2/perspective_projection/test.toml
kjarosh: Is it possible that FP stores focal length instead of FOV and just converts it back and forth when setting/getting? So it would be the other way around compared to how we do it currently.

cookie-s: Interesting idea. However, I don't think so.

1. I tried it now (although it's not perfect implementaion): https://github.com/cookie-s/ruffle/compare/avm2-pp-focallength-precise...cookie-s:ruffle:avm2-pp-focallength-precise-test-fl-save
   and obtained the following diff:
   ```
   test avm2/perspective_projection       ... FAILED
   
   failures:
   
   ---- avm2/perspective_projection ----
   Approximation failed: expected 55, found 54.99999875973957. Episilon = Some(1e-13), Max Relative = None
   
   Diff < left / right > :
    // Default
      perspectiveProjection = [object PerspectiveProjection]
   <  perspectiveProjection.fieldOfView = 54.99999875973957
   >  perspectiveProjection.fieldOfView = 55
      perspectiveProjection.focalLength = 480.24554443359375
      perspectiveProjection.projectionCenter = (x=250, y=250)
    
    // Default values (stage)
      perspectiveProjection = [object PerspectiveProjection]
   <  perspectiveProjection.fieldOfView = 54.99999875973957
   >  perspectiveProjection.fieldOfView = 55
      perspectiveProjection.focalLength = 480.24554443359375
      perspectiveProjection.projectionCenter = (x=250, y=250)
   ```
   In this case, even `55` cannot be restored through FOV -> FL -> FOV process.
2. A previous attempt shows that the `.transform.perspectiveProjection` assignment preserves FOV rather than FL. https://github.com/ruffle-rs/ruffle/pull/19532#discussion_r1957233529
   This is rather a hint for guessing, than an evidence, though.

tests/tests/swfs/avm2/perspective_projection/output.txt
cookie-s: Oh, true.
I'm using Adobe Flash Player 32 debugger for Windows (`flashplayer_32_sa_debug`) through wine on Linux x86_64 machine. I tested the same SWF on my different Windows machine and it produced the same result.

But as you said, Adobe Flash Player 32,0,0,465 on the same Linux machine produces a different result.

It probably depends on FP variants instead of CPU etc.?

tests/tests/swfs/avm2/perspective_projection/test.toml
adrian17: > I found the code above gives 179.99999000000003, so even pp.fieldOfView is not a simple storage variable.

This still sounds like a simple storage variable, it could just be a floating point accuracy issue?

core/src/avm2/globals/flash/geom/perspective_projection.rs
adrian17: When we're much closer to FP inaccuracy thresholds, did you also try other tricks like replacing `* DEG2RAD` by explicit `* PI / 180.0`? Since doing the multiplication and division separately could also affect the result.

adrian17:
To be sure, I know kjarosh's results are still different, but are they still closer than before? If so, I think this can be approved (with some extra comments documenting remaining deviations) - and sorry for looking at it so late.

tests/tests/swfs/avm2/perspective_projection/test.toml
kjarosh: > This still sounds like a simple storage variable, it could just be a floating point accuracy issue?

Based on my past calculations it's not an IEEE754 accuracy issue, as it's not the closest value to 179.99999

kjarosh (approved):
> To be sure, I know kjarosh's results are still different, but are they closer than before?

I think they are closer, LGTM after documenting those differences

core/src/avm2/globals/flash/geom/perspective_projection.rs
cookie-s: I tried other cases like `_ * PI / 180.0` and `_ / 180.0 * PI`, and also both Rust `std::f64::const::PI` and ActionScript `Math.PI` defined in `Math.as`, but as far as I tested, they didn't show significant improvement...

cookie-s (commented):
Thank you for reviews!

I added a comment to `test.toml` about the difference.

cookie-s:
Rebased to pass lint

cookie-s:
oh, no. It's failing even at the current master.

kjarosh:
> oh, no. It's failing even at the current master.

Don't worry, it's a beta lint. They added some more lints and that's why it's failing, it's unrelated to this PR

cookie-s:
Now the lint test also passes!

kjarosh (approved):
Still LGTM

kjarosh:
@cookie-s I'm sorry for such long review, thank you for this PR!

cookie-s:
Thanks!

PR #7777, opened 2022-08-27
web: Add webgl backend through wgpu

Dinnerbone:
This uses wgpu API entirely on web (NOT webgpu), which translate into webgl. One less custom renderer for us to maintain.

The downside is that we stop using WebGPU (the experimental rendering tech) as this is a compile time switch.

My proposal is to try this out for a while, and if we want WebGPU back then we can build our binary twice - one for webgl and one for webgpu - and load the correct one depending on the existence of `navigator.gpu`.

Opinions and thoughts from all maintainers are welcome. Is this something we should try?

PR #19767, opened 2025-03-10
ci: Split the web build job in the nightly release workflow into extensions and demo+selfhosted+docs

torokati44:
This builds on top of https://github.com/ruffle-rs/ruffle/pull/19316.
What do y'all think?


torokati44:
Something like https://github.com/actions/runner/issues/1182 would make duplicating the first 50 or so lines of tool setup unnecessary...

.github/workflows/release_nightly.yml
```diff
@@ -486,46 +452,102 @@ jobs:
...
+          BUILD_ID: ${{ github.run_number }}
+          ENABLE_VERSION_SEAL: "true"
+          FIREFOX_EXTENSION_ID: ${{ secrets.FIREFOX_EXTENSION_ID }} # Needed to inject into manifest.json
+        run: npm run version-seal
+
+      - name: Build web
+        env:
+          BUILD_ID: ${{ github.run_number }}
+          # Build web demo with WebGPU support for testing in Chrome origin trial on ruffle.rs
+          CARGO_FEATURES: jpegxr,webgpu
```
danielhjacobs, line 518: I know this builds off my changes which had the same change but before my PR or this PR webgpu was only built for the demo, now it's built for selfhosted too, is that a problem?

danielhjacobs:
> Something like [actions/runner#1182](https://github.com/actions/runner/issues/1182) would make duplicating the first 50 or so lines of tool setup unnecessary...

Ancient issue, but some hope is given by https://github.com/actions/runner/issues/1182#issuecomment-2317953582 at least.

.github/workflows/release_nightly.yml
```diff
@@ -486,46 +452,102 @@ jobs:
...
+          FIREFOX_EXTENSION_ID: ${{ secrets.FIREFOX_EXTENSION_ID }} # Needed to inject into manifest.json
+          WASM_SOURCE: cargo_and_store
+        working-directory: web
+        shell: bash -l {0}
+        run: npm run build:repro
+
+      - name: Build web docs
+        working-directory: web
+        run: npm run docs
```
danielhjacobs, line 527: This isn't a change from the current workflows even, but while we're looking at this, do we know why both workflows build the web docs?

danielhjacobs: Seems unnecessary unless I misunderstand something.

web/README.md
```diff
@@ -105,7 +105,8 @@ In this project, you may run the following commands to build all packages:
...
-    -   `npm run build:repro` enables reproducible builds. Note that this also requires a `version_seal.json`, which is not provided in the normal Git repository - only specially-marked reproducible source archives. Running this without a version seal will generate one based on the current state of your environment.
+    -   `npm run build:repro` enables reproducible builds with the default WASM module. Note that this also requires a `version_seal.json`, which is not provided in the normal Git repository - only specially-marked reproducible source archives. Running this without a version seal will generate one based on the current state of your environment.
+    -   `npm run build:dual-wasm-repro` enables reproducible builds with both WASM modules. Note that this also requires a `version_seal.json`, which is not provided in the normal Git repository - only specially-marked reproducible source archives. Running this without a version seal will generate one based on the current state of your environment.
     -   You will also need to run `rustup component add rust-src` with either of the previous two commands since we rebuild std for the vanilla WASM module.
```
danielhjacobs, line 110: Again, this is a change I should have made in my original PR, but this should now say:
`    -   You will also need to run `rustup component add rust-src` with either of the dual-wasm commands since we rebuild std for the vanilla WASM module.`

previous two ---> dual-wasm

danielhjacobs:
Makes sense to me, aside from the questions I asked.

danielhjacobs:
Maybe add a TODO comment about the duplicated logic in both places it appears, pointing to https://github.com/actions/runner/issues/1182

.github/workflows/release_nightly.yml
danielhjacobs: As I wrote on Discord:

> On September 4, 2022 a toggle was added for webgpu and this was added: https://github.com/ruffle-rs/ruffle/pull/7777. However, on January 13, 2023, wgpu-webgl become the default: https://github.com/ruffle-rs/ruffle/pull/9134. wgpu-webgl uses both the feature wgpu uses and webgl through wgpu: https://github.com/ruffle-rs/ruffle/blob/a2c3d5e7dd54ff9a50de1f2c6bf9011237c3b73c/web/Cargo.toml#L29-L30 So is there still a reason to use the webgpu feature for the web demo?

Basically, unless I misunderstand I'm not sure we need the webgpu cargo feature at all in any workflow. Do we?

danielhjacobs: > I think if we want to make the PR refactor-only and not have any major user-facing implications, then disabling webgpu is okay.
But not being able to deploy webgpu demo-only anymore is kinda unfortunate, if we ever wanted it.

@adrian17 on Discord

.github/workflows/release_nightly.yml
```diff
@@ -406,32 +399,7 @@ jobs:
         working-directory: web
         run: npm run docs
```
danielhjacobs, line 400: I believe this should be removed from here, and kept in the place where I made my initial comment about the docs.

web/README.md
torokati44: Amended your commit, thanks!

.github/workflows/release_nightly.yml
torokati44: Removed, thank you.

.github/workflows/release_nightly.yml
torokati44: Removed the webgpu feature for now, but added a note about this.

.github/workflows/release_nightly.yml
```diff
@@ -486,46 +451,103 @@ jobs:
...
 
-          # Restore our custom swfs
-          git restore swfs swfs.json
+      # Keep the version number in sync in all workflows,
+      # and in the extension builder Dockerfile!
+      - name: Install wasm-opt
+        uses: sigoden/install-binary@v1
+        with:
+          repo: WebAssembly/binaryen
+          tag: version_123
```
danielhjacobs, line 496: One more place to update when binaryen is bumped :disappointed:

danielhjacobs (approved):
If we're okay with bumping the minimum supported Safari version for the extension specifically to 16.4 (I'm fine with it), approved.

torokati44:
Thank you for the feedback! I'm going to merge tomorrow then, unless anyone objects until then.

danielhjacobs:
I don't know how I missed this but I think you did away with build:dual-wasm-repro

danielhjacobs:
Not as an option but in the CI.

danielhjacobs:
It needs to be used for selfhosted.

.github/workflows/release_nightly.yml
```diff
@@ -486,46 +451,103 @@ jobs:
...
+        env:
+          BUILD_ID: ${{ github.run_number }}
+          # NOTE: In the future, we might want to enable some features (like `webgpu`) only in
+          # the demo build, for limited testing (like a Chrome origin trial) on ruffle.rs.
+          CARGO_FEATURES: jpegxr
+          FIREFOX_EXTENSION_ID: ${{ secrets.FIREFOX_EXTENSION_ID }} # Needed to inject into manifest.json
+          WASM_SOURCE: cargo_and_store
+        working-directory: web
+        shell: bash -l {0}
+        run: npm run build:repro
```
danielhjacobs, line 523:
```yml
        run: npm run build:dual-wasm-repro
```

Issue #19187, opened 2025-01-12
Hexiom Connect: Connection issue


waspennator:
### Describe the bug

For some reason the tiles won't connect to each other

### Expected behavior

Tiles should connect 

### Content Location

https://www.newgrounds.com/portal/view/458166?emulate=flash

### Affected platform

Browser's extension

### Operating system

Android 14

### Browser

Chrome 131

evilpie:
This SWF needs a proper implementation of `flash.display.DisplayObjectContainer.getObjectsUnderPoint`.

Issue #19083, opened 2024-12-28
Type Coercion from Sprite to Bitmap failed


JimenezLi:
### Describe the bug

When clicking puzzle slices, the game throws an error:
```
ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "mouseDown", class: flash.events::MouseEvent,s: flash.events::MouseEvent, ptr: 0x239a162cfd0 }) to handler FunctionObject(FunctionObject { ptr: 0x23999ffffc0, name: Ok("Level/puzzError #1034: Type Coercion fleClick()") }) : TypeError: Error #1034: Type Coercion failed: cannot convert flash.display::Sprite@00000000000 to flash.display.Bitmap.
        at Level/puzzleClick()
```

### Expected behavior

The puzzle slices should be dragged normally.
![Image](https://github.com/user-attachments/assets/75a224c5-7b5d-4c22-951e-33243769783a)

### Content Location

[com.arcadevoid.jigcircle.swf.zip](https://github.com/user-attachments/files/18266263/com.arcadevoid.jigcircle.swf.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

Lord-McSweeney:
The SWF expects a `getObjectsUnderPoint` call to only return Bitmaps, but for some reason in Ruffle a Sprite is in the array

PR #19793, opened 2025-03-13
avm2: Improve flash.display.DisplayObjectContainer.getObjectsUnderPoint

cookie-s:
Closes https://github.com/ruffle-rs/ruffle/issues/19187
Closes https://github.com/ruffle-rs/ruffle/issues/19083

* Fixes the order of returned values
* Fixes the `HitTestOptions` given to `hit_test_shape`
* Fixes the behavior against parents of any object with true hittest result.

I found the expected result is not obtained from `getObjectsUnderPoint` in the `from_shumway/hittesting/hittesting` test case yet. I guess I shouldn't remove the stub now.


core/src/display_object.rs
```diff
@@ -2651,6 +2651,9 @@ bitflags! {
         /// Ignore objects with the ActionScript's visibility flag turned off.
         const SKIP_INVISIBLE = 1 << 1;
 
+        /// Check only the specified object. Ignore any childrens of that object.
```
kjarosh, line 2654:
```suggestion
        /// Check only the specified object. Ignore any children of that object.
```

core/src/avm2/globals/flash/display/display_object_container.rs
```diff
@@ -590,15 +591,23 @@ pub fn get_objects_under_point<'gc>(
     };
 
     let mut under_point = Vec::new();
-    let mut children = vec![this.as_display_object().unwrap()];
-    // FIXME - what are the actual options?
-    let options = HitTestOptions::SKIP_MASK;
```
kjarosh, line 595: I'd personally leave the variable with updated options, it's more readable

kjarosh (approved):
Except those two remarks above, LGTM! Thank you!

cookie-s:
Thank you, they make sense, and fixed. oh, thank you for merging too.

PR #19822, opened 2025-03-15
web: Reload with canvas when webglcontextlost and 8+ instances exist

danielhjacobs:
Intended to fix this comment in https://github.com/ruffle-rs/ruffle/issues/1905#issuecomment-1030111132

### DIFF
--- a/web/src/lib.rs
+++ b/web/src/lib.rs
@@ -39,7 +39,7 @@ use wasm_bindgen::prelude::*;
 use web_sys::{
     AddEventListenerOptions, ClipboardEvent, Element, Event, EventTarget, FocusEvent,
     Gamepad as WebGamepad, GamepadButton as WebGamepadButton, HtmlCanvasElement, HtmlElement,
-    KeyboardEvent, Node, PointerEvent, ShadowRoot, WheelEvent, Window,
+    KeyboardEvent, Node, PointerEvent, ShadowRoot, WebGlContextEvent, WheelEvent, Window,
 };
 
 static RUFFLE_GLOBAL_PANIC: Once = Once::new();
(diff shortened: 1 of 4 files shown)


danielhjacobs:
I suppose technically, if the situation exists where the browser supports the webgl renderer but not the canvas renderer, this may infinitely loop in a way. Since preferredRenderer doesn't force the renderer so the load may fallback to webgl.

Not sure how to address that.

web/src/lib.rs
```diff
@@ -771,6 +776,21 @@ impl RuffleHandle {
...
+            // Create webglcontextlost handler.
+            instance.webglcontextlost_callback = Some(JsCallback::register(
+                &player.canvas,
+                "webglcontextlost",
+                false,
+                move |_js_event: WebGlContextEvent| {
+                    INSTANCES.with(|instances| {
+                        if instances.borrow().len() >= 8 {
+                            let _ = ruffle.remove_instance();
+                            js_player.reload_with_canvas_renderer();
```
adrian17, line 788: to be sure, how come we didn't need to `js_player.clone()` anywhere?
Does this mean a clone wasn't needed somewhere earlier already?

danielhjacobs: Not entirely sure as I'm not super familiar with Rust, but isn't it because:

1) I use the move keyword on the closures scope
2) The `reload_with_canvas_renderer` wasm-bindgen function uses it by reference.

danielhjacobs: That's a good question though, I can switch to the method used here:

https://github.com/ruffle-rs/ruffle/blob/ba677e0c40a74488be911eb8f2ba2445a1641f7b/web/src/lib.rs#L593-L599

Just not sure if I have to do so.

danielhjacobs: The `js_player.clone()` in the `mouse_down_callback` does seem to be needed, perhaps because of the `let _ = ruffle.with_instance(move |instance| {`

PR #19869, opened 2025-03-21
avm1: Only allow valid boxed primitives in `NativeObject` enum

moulins:
And fixes a corner case for the `"string".length` property.


core/src/avm1/value.rs
```diff
@@ -493,16 +493,12 @@ impl<'gc> Value<'gc> {
...
-            _ => {
-                let vbox = Gc::new(activation.gc(), Value::Undefined);
-                obj.set_native(activation.gc(), NativeObject::Value(vbox));
-                Ok(Value::Undefined)
-            }
-        };
+        match value {
+            Value::Bool(_) => drop(globals::boolean::constructor(activation, obj, &[value])),
+            Value::Number(_) => drop(globals::number::number(activation, obj, &[value])),
+            Value::String(_) => drop(globals::string::string(activation, obj, &[value])),
```
Lord-McSweeney, line 499: Do we need to `drop` these?

moulins: it's a shorter way of writing `{ globals::boolean::constructor(...); }`, this is needed because of the `_ => ()` arm that types the whole match as `()`.

If you prefer, I can instead write the last arm as `_ => Ok(Value::Undefined)`, but this value will immediately get discarded so I find it a little weird.

PR #19874, opened 2025-03-21
tests: Add tests for dead keys on Windows

kjarosh:
These tests cover the use case of inputting text using dead keys on Windows.

These tests also prove that we cannot get rid of `TextInput`, as it's possible to have `KeyDown` without a `TextInput`, and to have `TextInput` without a `KeyDown`.

Note that on Linux, dead keys use IME, so that we can see the preview of the character and `TextInput` is effectively superfluous there.

CC @evilpie


evilpie:
But doesn't the dead key sequence in Windows usually involve [CTRL]? Like these listed on https://www.computerhope.com/issues/ch000657.htm#shortcut. Or is this something Flash specific?

kjarosh:
> But doesn't the dead key sequence in Windows usually involve [CTRL]? Like these listed on https://www.computerhope.com/issues/ch000657.htm#shortcut. Or is this something Flash specific?

These seem to be Microsoft Word specific. I was testing it using the US international keyboard layout with dead keys. You can press <kid>"</kbd>,<kbd>a</kbd> to get `ä`, etc.

evilpie:
Ah. I had English: US, but I can reproduce this with English: United States - International.

evilpie (approved):

Issue #16124, opened 2024-04-26
Sushi Cat 2: Quality Button and Performance Issues


EmmyTheOptimist:
### Describe the bug

The first thing I noticed was that the game ran unusually slow in general. It got to the point where it nearly froze just by running game physics.

I then hoped to fix this by lowering the quality, but the built-in quality toggle wasn't working.

### Expected behavior

The quality button would work, and the game would speed up.

### Content Location

https://archive.org/download/armorgames/sushi-cat-2-10471f2f2.swf

### Affected platform

Desktop app

### Operating system

Windows 11

PR #19884, opened 2025-03-22
avm2: Better optimizer

Lord-McSweeney:
This implements proper state merging in the optimizer, which significantly improves performance in obfuscated code (such as in #16124).

The new optimizer algorithm also runs somewhat faster, especially on large, complex methods.

The abstract interpreter optimizes code in a separate phase now, so we can record optimization stats again. Box2d:
```
FindPropStrict -> GetScriptGlobals: 82.71%
FindPropStrict -> GetOuterScope: 9%
FindPropStrict -> GetScopeObject: 8.28%

InitProperty -> SetSlot: 8.46%
InitProperty -> SetSlotNoCoerce: 91.53%

SetProperty -> SetSlot: 9.96%
SetProperty -> SetSlotNoCoerce: 71.67%
SetProperty -> CallMethod: 2.44%

GetProperty -> GetSlot: 88.89%

CallProperty -> CallMethod: 83.26%

CallPropVoid -> CallMethod: 91.28%

Coerce -> Nop: 67.25%

CoerceD -> Nop: 90.04%

CoerceB -> Nop: 27.02%

CoerceU -> Nop: 62.28%

CoerceI -> Nop: 47%

ReturnValue -> ReturnValueNoCoerce: 83.78%
```


core/src/avm2/op.rs
```diff
@@ -136,13 +140,13 @@ pub enum Op<'gc> {
         index: u32,
     },
     GetOuterScope {
-        index: u32,
+        index: usize,
```
adrian17, line 143: why switch these to usize?

core/src/avm2/optimizer/optimize.rs
```diff
@@ -0,0 +1,1761 @@
...
+                        match vtable.get_trait(&multiname) {
+                            Some(Property::Slot { slot_id })
+                            | Some(Property::ConstSlot { slot_id }) => {
+                                optimize_op_to!(Op::ConstructSlot {
+                                    index: slot_id,
+                                    num_args
+                                });
+
+                                let mut value_class = vtable.slot_classes()[slot_id as usize];
+                                let resolved_value_class = value_class.get_class(activation)?;
```
adrian17, line 1485: This previously bailed out on failed lookup on `PropertyClass::Name`, now it throws a RustError. To be sure, is this intentional?
And to be sure, in FP `PropertyClass::Name` wouldn't exist as all the slots should have been resolved at an earlier time, right? The class refactor didn't eliminate the need for these? (I'm asking in general, not in scope of this PR)

adrian17 (approved):

core/src/avm2/op.rs
Lord-McSweeney: It prevents some casts between `u32` and `usize`, and it makes sense for these to be `usize` since they're indexing into the local/outer scopes

core/src/avm2/optimizer/optimize.rs
Lord-McSweeney: It is intentional, though this is an observable change. Testing seems to show that we should be resolving all the slot classes (and method return/parameter types, etc) when the first ClassObject corresponding to the class is created, but I'm planning to implement this correctly along with the method return type verification in a future PR.

PR #19888, opened 2025-03-22
desktop: Implement IME preediting

kjarosh:
Flash Player implements IME preediting as replacing the text with the preedit value and not dispatching events.

This implementation still lacks visual clues which text is being inputted.

This PR adds a simple test for IME based on how dead keys work on Linux with IME.

* Progresses https://github.com/ruffle-rs/ruffle/issues/1778


core/src/input.rs
```diff
@@ -163,6 +168,13 @@ impl InputManager {
                 }
             }
             PlayerEvent::KeyUp { key } => {
+                if !self.key_descriptors_down.remove(&key) {
+                    // Ignore spurious KeyUp events that may happen e.g. during IME.
+                    // We assume that in order for a key to generate KeyUp, it had to
+                    // generate KeyDown for the same exact KeyDescriptor.
```
adrian17, line 174: to be sure, this doesn't impact AS-side `keyUp` handlers at all, right?

adrian17 (approved):

core/src/input.rs
kjarosh: What do you mean? They won't be fired when the keyup event is filtered out here.

kjarosh: Chatted offline — updated the comment.

PR #19889, opened 2025-03-22
core: Use player notifications for virtual keyboard

kjarosh:
This refactor replaces `open_virtual_keyboard`/`close_virtual_keyboard methods` from `UiBackend` with `OpenVirtualKeyboard`/`CloseVirtualKeybard` player notifications.

It's a follow up to https://github.com/ruffle-rs/ruffle/pull/19666 and a PoC of how notifications can be used for other purposes too.

### DIFF
--- a/core/src/focus_tracker.rs
+++ b/core/src/focus_tracker.rs
@@ -6,6 +6,7 @@ pub use crate::display_object::{
     DisplayObject, TDisplayObject, TDisplayObjectContainer, TextSelection,
 };
 use crate::display_object::{EditText, InteractiveObject, TInteractiveObject};
+use crate::events::PlayerNotification;
 use crate::events::{ClipEvent, KeyCode};
 use crate::prelude::Avm2Value;
 use crate::Player;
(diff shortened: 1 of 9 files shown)


kjarosh:
@torokati44 is this refactor okay for Android?

core/src/focus_tracker.rs
```diff
@@ -208,15 +209,17 @@ impl<'gc> FocusTracker<'gc> {
     }
 
     fn update_virtual_keyboard(&self, context: &mut UpdateContext<'gc>) {
-        if let Some(text_field) = self.get_as_edit_text() {
+        let notification = if let Some(text_field) = self.get_as_edit_text() {
```
evilpie, line 212:
```suggestion
        let notification = self.get_as_edit_text().is_some_and(|tf| tf.is_editable())
```

n0samu:
Has anyone tested this on iOS? I'd be happy to do so if someone puts up a demo build

kjarosh:
> Has anyone tested this on iOS? I'd be happy to do so if someone puts up a demo build

Oh dang, I forgot about iOS. It won't work there because notifications are asynchronous and it requires focus to be set synchronously.

PR #19898, opened 2025-03-23
web: Fix vulnerability for upstream dependency

tsunamistate:
Applies `npm audit fix` suggestion
This only upgrades `@babel/helpers` packages, so shouldn't be a big deal

Vulnerability:

@babel/helpers  <7.26.10
Severity: moderate
Babel has inefficient RexExp complexity in generated code with .replace when transpiling named capturing groups - https://github.com/advisories/GHSA-968p-4wvh-cqc8

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -252,27 +252,27 @@
             }
         },
         "node_modules/@babel/helpers": {
-            "version": "7.26.0",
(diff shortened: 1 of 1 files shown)


evilpie:
Thank you! I agree that this shouldn't really be a problem for us.

PR #19908, opened 2025-03-24
chore: Update `getrandom` to `0.3` (and `rand` to `0.9`)

torokati44:
So far we've ignored a bunch of dep updates because of this, but let's try to upgrade them all at once.

So far, not every reference to `getrandom` `0.2` can be dropped, see:
 - https://github.com/tkaitchuck/aHash/pull/251 (https://github.com/tkaitchuck/aHash/discussions/258)
 - https://github.com/briansmith/ring/issues/2341
 - https://gitlab.redox-os.org/redox-os/users/-/blob/0f2d9faaeb04e5f74cbcc4e84d2f04f147c3da54/Cargo.toml#L14
 
 The duplication of the `objc2*` family crates to `0.5.x` and `0.6.x` is also unfortunate: https://github.com/1Password/arboard/pull/173
 So is that of `zerocopy` to `0.7.x` and `0.8.x`: https://github.com/tkaitchuck/aHash/pull/246


torokati44:
The changed output of the `avm2/stage3d_bitmap` and `avm2/stage3d_texture` visual tests are more than likely the result of https://github.com/rust-random/rand/pull/1203 (https://github.com/rust-random/rand/blob/7808f4e1f7d9578e3c08405b1f4e575720d58a51/CHANGELOG.md#reproducibility-breaking-optimisations).

adrian17:
Oh, making wasm more awkward is one thing, but what's with these reproducibility changes...? Did they really not document how to preserve old results at all? :/

torokati44:
Well, not really...:

>The documentation for seed_from_u64 says changing the algorithm is a "value-breaking change" but on the other hand the current implementation is a unintended bug so might not be protected from breaking changes.

> I'm not really convinced it's worth introducing a breaking change for this. In any case, we make zero guarantees about compatibility of SmallRng with other Xoshiro implementations;

torokati44:
The RNG is not a 1:1 match with FP anyway: https://github.com/ruffle-rs/ruffle/issues/20244

kjarosh (approved):

PR #19935, opened 2025-03-27
chore: Bump Binaryen (wasm-opt) to version 123

torokati44:
https://github.com/WebAssembly/binaryen/pull/7394:
> Given the big speedup in our official release binaries for Linux, this seems useful to get to users quickly.

https://github.com/WebAssembly/binaryen/pull/7378:
> fixes performance problems, especially on heavily multi-threaded workloads (many functions, high core count machines)

### DIFF
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -356,7 +356,7 @@ jobs:
         uses: sigoden/install-binary@v1
         with:
           repo: WebAssembly/binaryen
-          tag: version_122
+          tag: version_123
           name: wasm-opt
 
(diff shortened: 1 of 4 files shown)


torokati44:
@adrian17 I remember you often complaining about `wasm-opt` taking super long to run (especially twice) in your local builds, and opting to eliminate it somehow because of that when developing. I can recommend installing this version, it really does feel _a lot_ faster to me, at least on my multi-dozen-core machine!

adrian17:
Feels just a bit faster (21->16s), but it's still nice, thanks c:

Issue #11771, opened 2023-06-28
sushi Get Black screen, can't play.


olux997:
### Describe what you were doing

This flash player 7 game play get Black screen, can't play

title screen  :
<img width="350" alt="sushi-01" src="https://github.com/ruffle-rs/ruffle/assets/11493572/c93e72da-8da8-4393-a455-087ddb6e3e83">

Game Play screen :
<img width="350" alt="sushi-02" src="https://github.com/ruffle-rs/ruffle/assets/11493572/19f80aae-3da7-4478-bdaa-81b5ff9fb420">

Game Play screen [Ruffle]:
<img width="350" alt="sushi-ruffle" src="https://github.com/ruffle-rs/ruffle/assets/11493572/29526840-172f-470c-9da3-bf61c62e9883">

swf file:
[sushi.zip](https://github.com/ruffle-rs/ruffle/files/11888822/sushi.zip)

Ruffle no any error message.

### What does the crash message say?

Ruffle no any error message.

### Ruffle Version

Nightly 2023-06-27

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Additional information

none

n0samu:
The black screen is actually due to a movieclip called "Mdark" blocking out the entire stage. If I use the debug tools to set the movieclip to invisible, the gameplay screen appears:
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/23e366d1-fee1-47a0-b495-322fc8e03be8)

As far as I can tell, this is the relevant AS code:
```as
function FadeScreen(nSpeed, bState)
{
   if(bState == true)
   {
      this.Mdark._alpha -= nSpeed;
      if(this.Mdark._alpha <= 0)
      {
         this.Mdark.removeMovieClip();
         CTimer.Clear("FadeScreen");
      }
   }
   else
   {
      if(nSpeed == null)
      {
         nSpeed = 5;
      }
      this.createEmptyMovieClip("Mdark",100001);
      with(this.Mdark)
      {
         beginFill(0,100);
         lineTo(1000,0);
         lineTo(1000,1000);
         lineTo(0,1000);
         lineTo(0,0);
         endFill();
      }
      CTimer.Set("FadeScreen",this,40,nSpeed,true);
   }
}
```
This will need further investigation to determine what exactly is going wrong. Adding a trace statement into the FadeScreen function, I see that it is only being called once, even though it's supposed to be called at an interval (which the `CTimer.Set` call sets up).

[...]

PR #19936, opened 2025-03-27
avm1: migrate functions to `NativeObject`, and fix native object subclassing

moulins:
This is best reviewed commit by commit.

The only remanining non-`ScriptObject` variants are `SuperObject` and `StageObject`, which cannot be constructed directly by user code: this means that `NativeObject` subclasses can be constructed on top of arbitrary non-native `ScriptObject`s, and that the `TObject::create_bare_object` workaround isn't necessary anymore.

This keeps the `TObject::construct/construct_on_existing` distinction, as some builtin constructors (e.g. `BitmapData`) can return non-objects in some cases.


Toad06:
Thanks a lot for your work on this! I had some files affected by this issue, and they all work properly now.

#11771 is also fixed. I assume there are more, but it's hard to be sure.
(Oh, and this also fixes Digestomax, one of the mini-games from KadoKado. 😉)

Lord-McSweeney (approved):

PR #19939, opened 2025-03-27
web: Replace custom Map with fresh Map to fix #19758

danielhjacobs:


evilpie:
I would suggest we instead try to get an unchanged `Map` constructor from a fresh global. Can we limit this code to this strange page?

```js
var previous = Map;
if (Map is custom) {
  var iframe = document.createElement("iframe");
  iframe.style.display = "none";
  document.documentElement.append(iframe) 
  Map = iframe.contentWindow.Map;
}
// .. run translation ...
Map = previous;
```

(This is another reason why I think running WebExtension content scripts in an isolated context is useful)

danielhjacobs:
Ah, I wanted to reset to the global Map and then restore the custom one, but I couldn't think of a way to do so after the page already changed the global Map. I didn't consider a hidden iframe to get a fresh global.

> Can we limit this code to this strange page?

Probably, but I'm not sure I like that better. With my new code, I just check if the Map is custom and if not return undefined, then check if I need to restore it and if not do nothing. Granted, I do that for every translation (every call to `text(`), but it's the extra code that runs is pretty minimal if there's not a custom Map.

web/packages/core/src/js-polyfills.ts
```diff
@@ -126,6 +127,39 @@ function tryPolyfillReflect(): void {
...
+ *
+ * @returns The custom `Map` object that exists on the page, or undefined if the page uses the standard `Map`.
+ */
+export function resetCustomMap(): typeof Map | undefined {
+    if (typeof Map.prototype.set !== "function") {
+        const currentMap = Map;
+        const iframe = document.createElement("iframe");
+        iframe.style.display = "none";
+        document.documentElement.append(iframe);
+        if (iframe.contentWindow) {
```
evilpie, line 141: When is this `null`?

web/packages/core/src/js-polyfills.ts
```diff
@@ -126,6 +127,39 @@ function tryPolyfillReflect(): void {
...
+            // eslint-disable-next-line no-global-assign
+            Map = iframe.contentWindow.Map;
+            iframe.remove();
```
evilpie, line 144: We should definitely always remove the `iframe`.

web/packages/core/src/js-polyfills.ts
danielhjacobs: https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement/contentWindow#description

Because TS doesn't know if an iframe is cross-origin, it requires checking if `iframe.contentWindow` exists, because in a cross-origin case it may not.

evilpie: I don't know much about TS, but this is where you would use `!` right?

web/packages/core/src/internal/i18n.ts
```diff
@@ -20,6 +21,7 @@ const bundles: Record<string, FluentBundle> = {};
 for (const [locale, files] of Object.entries(BUNDLED_TEXTS)) {
     const bundle = new FluentBundle(locale);
     if (files) {
+        const customMap = resetCustomMap();
```
danielhjacobs, line 24: Do we think the possibility of the custom map getting defined as the loop is running makes it worth running this code inside the loop instead? The `restoreCustomMap` can still run outside the loop, since it wouldn't make sense to repeatedly restore it just to re-create a fresh Map on each iteration.

danielhjacobs: Or do we think that possibility is remote enough to not be a concern?

web/packages/core/src/js-polyfills.ts
```diff
@@ -3,6 +3,7 @@ declare global {
         Prototype?: {
             Version?: string;
         };
+        Map: typeof Map;
```
danielhjacobs, line 6: `Map` of course exists in the global namespace. That inherently makes it available on Window, but alas TypeScript does not see it that way.

evilpie (approved):

adrian17 (approved):

PR #19940, opened 2025-03-28
avm2: Refactor the interpreter loop

Lord-McSweeney:
This depends on #19884

The first commit is based off @adrian17's work on reducing the number of instructions in the op dispatch part of the interpreter. Microbenchmarks show that a `nop` runs about 20% faster and that a `getslot` runs about 5-10% faster.


core/src/avm2/activation.rs
```diff
@@ -2653,49 +2600,39 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
 
     /// Implements `Op::LookupSwitch`
-    fn op_lookup_switch(
-        &mut self,
-        default_offset: i32,
-        case_offsets: &[i32],
-    ) -> Result<FrameControl<'gc>, Error<'gc>> {
-        self.timeout_check()?;
+    fn lookup_switch(&mut self, lookup_switch: &LookupSwitch) -> i32 {
+        let index = self.pop_stack().as_i32();
```
adrian17, line 2615: Note: since we don't verify that arg is an int (FP does at verification time), this technically introduces a panic where we used to throw a RustError before. Not sure if we care, but at least worth documenting?

adrian17:
ReturnVoid -> nice catch; this reminds me that once we have emitting, we could probably remove coercions from both `ReturnValue` and `ReturnVoid` and only emit them in the optimizer if types don't 100% match.

> Store interpreter instruction pointer as usize instead of i32

Can you rename this commit, as storing the absolute offsets is the more important change?

Also, personally I'm still not convinced about usize being better than u32. Yes, u32 needs more `as` casts and it's unfortunate; but it better corresponds to limits that AS3 has. I have no clue about perf impact; on one side the `as usize` might emit an extra op, on other side u32 is... well smaller on desktop 😅 

> avm2: Fix no-op branches

If possible, can this be squashed with earlier changes?

adrian17 (approved):

core/src/avm2/activation.rs
Lord-McSweeney: I don't think we should care at all; we already also panic instead of erroring in `getslot` and `setslot` when a non-Object is passed.

Issue #19885, opened 2025-03-22
Right-click context menu appears offset to the right


ran1324:
### Describe the bug

When right-clicking in the game/interface, the context menu does not appear directly under the cursor. Instead, it is displayed noticeably to the right of the click location. As a result, the menu is easy to miss until the user moves the cursor to the offset area.

Steps to Reproduce:

1. Right-click anywhere in the interface.

2. Observe that the context menu is displayed to the right, rather than where the cursor was clicked.

### Expected behavior

The context menu should appear exactly at the location of the mouse pointer upon right-click.

### Content Location

https://www.tipost.co.il/play/our-city/

### Affected platform

Desktop app

### Operating system

Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/134.0.0.0 Safari/537.36

### Browser

Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/134.0.0.0 Safari/537.36

### Additional information

https://github.com/user-attachments/assets/1307b832-769b-4af0-b568-fbfd3df4f4ef

![Image](https://github.com/user-attachments/assets/f282662e-2b65-4b90-9f0e-aeb6cf4f503f)

torokati44:
Sounds related to https://github.com/ruffle-rs/ruffle/issues/19697.

PR #19943, opened 2025-03-28
web: Fix context menu positioning on RTL pages

kjarosh:
This patch fixes a bug where the context menu was positioned incorrectly in RTL contexts. The bug was caused by the fact that x,y coordinates in JS events are not RTL aware, whereas the `transform` CSS property is.

This patch changes CSS transform usage to top/left/right absolute positioning.

Additionally, it adds support for RTL context menus, but sets their direction permanently to LTR, because we don't have an easy way of detecting LTR/RTL browsers. (Note that the context menu should not follow the direction of the page, but the direction of the browser.)

* Fixes https://github.com/ruffle-rs/ruffle/issues/19885


WumboSpasm:
LGTM

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -282,6 +282,12 @@ export class InnerPlayer {
         this.contextMenuElement.addEventListener("contextmenu", preserveMenu);
         this.contextMenuElement.addEventListener("click", preserveMenu);
 
+        // TODO Add support for RTL context menu.
+        //   It should use the direction of the browser, not the document.
+        //   See <https://developer.mozilla.org/en-US/docs/Web/API/Navigator/language>
+        //   See <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getTextInfo>
+        this.contextMenuElement.style.direction = 'ltr';
```
danielhjacobs, line 289: https://developer.mozilla.org/en-US/docs/Web/CSS/direction says "Where possible, authors are encouraged to avoid using the direction CSS property and use the HTML [dir](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir) global attribute instead".

I know once the comment is taken care of JS may be better, but would it work to change https://github.com/ruffle-rs/ruffle/blob/master/web/packages/core/src/internal/ui/context-menu-overlay.tsx#L3 to add `dir="ltr"` for now?

danielhjacobs: Also regardless, wouldn't it make more sense to put the direction here if it's static (for now):

https://github.com/ruffle-rs/ruffle/blob/master/web/packages/core/src/internal/ui/static-styles.css#L205

kjarosh: Thanks!
1. It does make more sense to use `dir` instead of `direction`, fixed
2. It has to be `dir="ltr"` because the context menu is shown on the right of the cursor. We also should not follow the context menu content and its direction, but browser's direction. So we'll be explicitly setting ltr/rtl.
3. We'll have to detect rtl/ltr in runtime in JS, that's why I added it there. I have some basic detection prepared for a future PR

PR #19946, opened 2025-03-28
avm2: Avoid allocating when calling a method

### LINKED ISSUE #12346: Test Failure: avmplus Exception suite
Completely passing suites are not shown.

## Testsuite `from_avmplus/as3/Statements/Exceptions/MultipleCatchBlocks2`
| Test | Passes? | Expected | Actual |
| ---- | ---- | ---- | ---- |
| Testing catch block with Reference Error | :heavy_check_mark: | | |
| Testing catch block with Type Error | :heavy_check_mark: | | |
| Testing catch block with Argument Error | :x: | `ArgumentError: Error #1063` | `no error` |
| Testing catch block with URIError | :x: | `no error` | `URIError: Error #1052: Inv` |
| Testing catch block with Eval Error | :heavy_check_mark: | | |
| Testing catch block with Range Error | :heavy_check_mark: | | |
| Testing catch block with Error | :heavy_check_mark: | | |

## Testsuite `from_avmplus/as3/Statements/Exceptions/MultipleCatchBlocksArgument`
| Test | Passes? | Expected | Actual |
| ---- | ---- | ---- | ---- |
| Testing catch block with Argument Error | :x: | `ArgumentError: Error #1063` | `no error` |

Lord-McSweeney:
This makes bytecode method calls with `call_method` about 30-40% faster and any other method calls 10-20% faster.

The unchecked function fix closes #12346.

### DIFF
--- a/core/src/avm2/activation.rs
+++ b/core/src/avm2/activation.rs
@@ -255,42 +255,6 @@ impl<'a, 'gc> Activation<'a, 'gc> {
         }
     }
 
-    /// Resolve a single parameter value.
-    ///
-    /// Given an individual parameter value and the associated parameter's
-    /// configuration, return what value should be stored in the called
-    /// function's local registers (or an error, if the parameter violates the
-    /// signature of the current called method).
-    fn resolve_parameter(
-        &mut self,
-        method: Method<'gc>,
-        value: Option<&Value<'gc>>,
-        param_config: &ResolvedParamConfig<'gc>,
-        user_arguments: &[Value<'gc>],
-        bound_class: Option<Class<'gc>>,
-    ) -> Result<Value<'gc>, Error<'gc>> {
-        let arg = if let Some(value) = value {
-            value
-        } else if let Some(default_value) = &param_config.default_value {
-            default_value
-        } else if method.into_bytecode().is_some_and(|bm| bm.is_unchecked()) {
-            return Ok(Value::Undefined);
-        } else {
-            return Err(Error::AvmError(make_mismatch_error(
-                self,
-                method,
-                user_arguments,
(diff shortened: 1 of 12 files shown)


core/src/avm2/activation.rs
```diff
@@ -378,75 +439,69 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
 
-        if method.is_unchecked() {
-            let max_args = signature.len();
-            if user_arguments.len() > max_args && !method.is_variadic() {
-                user_arguments = &user_arguments[..max_args];
-            }
-        }
-
-        if user_arguments.len() > signature.len() && !has_rest_or_args {
+        if user_arguments.len() > signature.len() && !has_rest_or_args && !method.is_unchecked() {
```
adrian17, line 440: just to be sure, the `is_unchecked()` already includes the `NEED_REST` but `has_rest_or_args` also checks `NEED_ARGUMENTS | NEED_REST` separately, is this indeed correct?

adrian17: Oh, sorry, only now saw that you also changed the definition of `is_unchecked` below. But... couldn't it go the other way? As in, make `is_unchecked()` cover `NEED_ARGUMENTS | NEED_REST`? Since this appears to be the only place that uses this field anyway.

Also, either way, the docstring at `fn is_unchecked` needs updating.

Lord-McSweeney: The change you proposed makes more sense after some of the commits of #20049; I'll add it in a later PR. I've updated the documentation on `is_unchecked`.

adrian17 (approved):

PR #19950, opened 2025-03-29
web: Add support for RTL context menu

kjarosh:
This patch adds support for RTL context menu. The context menu will use RTL when the preferred language of the user is RTL.

This patch will work only for browsers supporting `Intl.Locale.getTextInfo`, which sadly does not include Firefox.

### DIFF
--- a/web/packages/core/src/internal/player/inner.tsx
+++ b/web/packages/core/src/internal/player/inner.tsx
@@ -282,11 +282,7 @@ export class InnerPlayer {
         this.contextMenuElement.addEventListener("contextmenu", preserveMenu);
         this.contextMenuElement.addEventListener("click", preserveMenu);
 
-        // TODO Add support for RTL context menu.
-        //   It should use the direction of the browser, not the document.
-        //   See <https://developer.mozilla.org/en-US/docs/Web/API/Navigator/language>
-        //   See <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getTextInfo>
-        this.contextMenuElement.dir = 'ltr';
+        this.contextMenuElement.dir = detectBrowserDirection();
 
         document.documentElement.addEventListener(
             "pointerdown",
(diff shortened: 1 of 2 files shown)


web/packages/core/src/internal/player/inner.tsx
```diff
@@ -2328,3 +2327,31 @@ function parseAllowScriptAccess(
             return null;
     }
 }
+
+/**
+ * Detect direction (LTR/RTL) of the preferred language of the browser.
+ *
+ * @returns 'ltr' or 'rtl' depending on the direction, `null` when unable to detect
```
danielhjacobs, line 2334: Would it make sense to just return 'ltr' if unable to detect, if that's the default anyway, so we can just do:

`this.contextMenuElement.dir = detectBrowserDirection()`

kjarosh: Done!

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -2328,3 +2324,31 @@ function parseAllowScriptAccess(
...
     }
 }
+
+/**
+ * Detect direction (LTR/RTL) of the preferred language of the browser.
+ *
+ * @returns 'ltr' or 'rtl' depending on the detected direction
+ */
+function detectBrowserDirection(): string {
+    const browserLocale = new Intl.Locale(navigator.language) as any;
```
danielhjacobs, line 2334: Is the `as any` necessary. What happens without it?

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -2328,3 +2324,31 @@ function parseAllowScriptAccess(
...
+
+    let textInfo = null;
```
danielhjacobs, line 2336: Unused variable

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -2328,3 +2324,31 @@ function parseAllowScriptAccess(
...
+    if ('getTextInfo' in browserLocale &&
+        typeof browserLocale.getTextInfo === 'function') {
+        textInfo = browserLocale.getTextInfo();
```
danielhjacobs, line 2339: No return

danielhjacobs:
Nevermind the latest deleted review comments, I misread some code.

web/packages/core/src/internal/player/inner.tsx
danielhjacobs: I think the code should works exactly the same without it.

danielhjacobs (approved):

danielhjacobs:
Firefox implementation bug is https://bugzilla.mozilla.org/show_bug.cgi?id=1693576

web/packages/core/src/internal/player/inner.tsx
kjarosh: You're right, removed

PR #19984, opened 2025-03-31
core: Commit pending IME text on losing focus

kjarosh:
Flash commits the IME preedit text when losing focus.


core/src/display_object/edit_text.rs
```diff
@@ -3129,6 +3144,11 @@ impl<'gc> TInteractiveObject<'gc> for EditText<'gc> {
             self.set_selection(None, context.gc());
         }
 
+        if !focused {
```
danielhjacobs, line 3147: It barely matters performance-wise, but would it make sense to move the check above to be here as:

```rs
            let is_avm1 = !self.movie().is_action_script_3();
            if is_avm1 {
                self.set_selection(None, context.gc());
            }
```

core/src/display_object/edit_text.rs
```diff
@@ -3129,6 +3144,11 @@ impl<'gc> TInteractiveObject<'gc> for EditText<'gc> {
...
+            // Commit IME on focus lost.
+            self.ensure_ime_committed(context);
```
danielhjacobs, line 3149: Immediately before this is the code:

```rs
self.set_selection(None, context.gc());
```

But `ensure_ime_finished`, which `ensure_ime_committed` calls, has this code:

```rs
self.set_selection(Some(TextSelection::for_position(ime_start)), context.gc());
```

Which selection should happen last in the event of focus being lost?

kjarosh: Added a test for this behavior and changed the order of operations. Thanks!

Issue #19958, opened 2025-03-30
프리티걸의 붕어빵 타이쿤: The game won't load. (Variable flash.globalization::StringTools is not defined.)


jmousy:
### Describe the bug

When I try to run the game "프리티걸의 붕어빵 타이쿤" in Korean on Ruffle, it freezes to a black screen.

AVM2 / Tested on Ruffle Nightly 2025-03-30

![Image](https://github.com/user-attachments/assets/9ed7a5aa-e9b8-4ee1-b234-e2376eda7f1c)

The console error shows the following error:

```
2025-03-30T03:20:51.865449Z  INFO ruffle_desktop::gui::controller: Using preferred backend Vulkan
2025-03-30T03:20:51.928500Z  INFO ruffle_desktop::gui::controller: Using graphics API vulkan on NVIDIA GeForce RTX 4070 SUPER (type: DiscreteGpu)
2025-03-30T03:20:52.049316Z  WARN egui_wgpu::renderer: Detected a linear (sRGBA aware) framebuffer Bgra8UnormSrgb. egui prefers Rgba8Unorm or Bgra8Unorm    
2025-03-30T03:20:52.053046Z  INFO ruffle_desktop::gui::controller: loaded cjk fallback font "MalgunGothic"
2025-03-30T03:20:52.064838Z  INFO ruffle_video_external::decoder::openh264: Downloading OpenH264 library
2025-03-30T03:20:52.065624Z  INFO ruffle_video_external::decoder::openh264: Using OpenH264 at "C:\\Users\\user\\AppData\\Local\\ruffle\\video\\openh264-2.4.1-win64.dll"
2025-03-30T03:20:52.102327Z  INFO ruffle_core::context: Loaded SWF version 24, resolution 800x480 @ 30 FPS
2025-03-30T03:20:52.257836Z  INFO ruffle_desktop::backends::ui: Loading device font "ArialMT" for "Arial" (italic: false, bold: false)
2025-03-30T03:20:52.258422Z  INFO ruffle_core::library: Loaded new device font "Arial" (bold: false, italic: false) from file
2025-03-30T03:20:52.463847Z ERROR ruffle_core::display_object::movie_clip: Got "ReferenceError: Error #1065: Variable flash.globalization::StringTools is not defined.
	at sys::SystemFunction$/setLocale()
	at Main/init()
	at Main()" when constructing AVM2 side of movie clip of type Main
```

[...]

PR #19995, opened 2025-04-01
avm2: Implement flash.globalization.StringTools with stubs

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/19958


Lord-McSweeney (approved):

core/src/avm2/globals/flash/globalization/StringTools.as
```diff
@@ -0,0 +1,44 @@
+package flash.globalization {
+    import __ruffle__.stub_constructor;
+    import __ruffle__.stub_getter;
+    import __ruffle__.stub_method;
+
+    [API("667")]
+    public final dynamic class StringTools {
```
Lord-McSweeney, line 7: Sorry, I just noticed that StringTools also shouldn't be `dynamic`, can you fix it too?

kjarosh: Are you sure? `describeType` returns

```xml
<type name="flash.globalization::StringTools" base="Class" isDynamic="true" isFinal="true" isStatic="true">
```

adrian17: playerglobals also doesn't seem to show them as dynamic.

kjarosh: okay, deleted `dynamic`, no idea why `describeType` describes the type as dynamic then

adrian17: Did you check if it’s possible to set dynamic props on it? In the end, I don’t know which one of us is wrong, and this’ll probably answer it for good.

kjarosh: It's not possible to set dynamic props, so the class is not `dynamic`

PR #19996, opened 2025-04-01
avm2: Stub globals.flash.globalization.Collator

kjarosh:


core/src/avm2/globals/flash/globalization/Collator.as
```diff
@@ -0,0 +1,98 @@
+package flash.globalization {
+    import __ruffle__.stub_constructor;
+    import __ruffle__.stub_getter;
+    import __ruffle__.stub_method;
+
+    [API("667")]
+    public final dynamic class Collator {
```
Lord-McSweeney, line 7: The class shouldn't be `dynamic`

kjarosh: Why? `describeType` reports

```xml
<type name="flash.globalization::Collator" base="Class" isDynamic="true" isFinal="true" isStatic="true">
```

kjarosh: removed `dynamic` as per playerglobal

torokati44 (approved):

torokati44:
(nit: I see more AS3 setters in the codebase with the `void` return type explicitly added, than without, but it doesn't matter.)

PR #20016, opened 2025-04-04
avm2: Change optimizer's block scopes and add early exit to worklist

adrian17:
As is, this makes optimizer up to 10x faster in big functions (for example, 80ms -> 10ms).

Draft as I still wanna tinker with it a bit (and it fails one test, `verify_exception_target_two_jumps`).


core/src/avm2/optimizer/blocks.rs
```diff
@@ -1,141 +1,44 @@
...
+        {
+            let block = BasicBlock {
+                start_index: current_block_start,
+                ops: &code[current_block_start..i + 1],
+            };
+
+            block_list.push(block);
+
+            current_block_start = i + 1;
         }
```
Lord-McSweeney, line 42: Since these don't correspond to basic blocks anymore, could the struct be called something else?

core/src/avm2/optimizer/optimize.rs
```diff
@@ -426,29 +414,51 @@ impl<'gc> ScopeStack<'gc> {
...
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug)]
 struct AbstractState<'gc> {
     locals: Locals<'gc>,
     stack: Stack<'gc>,
     scope_stack: ScopeStack<'gc>,
 }
 
+#[derive(Clone, Debug)]
```
Lord-McSweeney, line 424: Is there a reason to not make this `Copy` and pass it by-value?

core/src/avm2/optimizer/optimize.rs
```diff
@@ -593,98 +600,110 @@ pub fn optimize<'gc>(
...
-
-        replacement_states.insert(start_index, abstract_state);
+fn process_jump<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    target: usize,
+    abstract_states: &mut Vec<Option<AbstractState<'gc>>>,
+    current_state: &AbstractStateRef<'_, 'gc>,
+    op_index_to_block_index_table: &HashMap<usize, usize>,
+    worklist: &mut Vec<usize>,
+    do_optimize: bool,
```
Lord-McSweeney, line 664: Since `do_optimize` isn't directly linked to whether the optimizer is running anymore, could it be changed to an enum holding the two abstract interpreter phases?

core/src/avm2/optimizer/optimize.rs
```diff
@@ -593,98 +600,110 @@ pub fn optimize<'gc>(
...
+            return Ok(());
+        }
+    } else {
+        // We don't have any state in the state list yet, so we use the provided one
+        abstract_states[target_block_id] = Some(current_state.to_owned());
+    };
+
+    // FP reschedules blocks to the front of queue (for us, it'd be back of the vec).
+    // I don't know if there's any good reason for that, but not doing it is faster.
+    if let Some(_) = worklist.iter().position(|x| *x == target_block_id) {
```
Lord-McSweeney, line 686: Could this just be `worklist.iter().position(|x| *x == target_block_id).is_none()`?

core/src/avm2/optimizer/optimize.rs
```diff
@@ -593,98 +600,110 @@ pub fn optimize<'gc>(
...
+            false,
+        )?;
+    }
 
-                    worklist.push((*target_block, exception_state));
-                }
-                BlockExit::Return => {}
-            }
+    if activation.avm2().optimizer_enabled() {
+        // Now run through the ops and actually optimize them
```
Lord-McSweeney, line 636: Could this iterate through `abstract_states`, consuming it, and do `block_list[i]` instead of `abstract_states[i]`.clone, since `BasicBlock` is `Copy`?

core/src/avm2/optimizer/optimize.rs
```diff
@@ -694,14 +713,35 @@ fn abstract_interpret_ops<'gc>(
...
+        if op.get().can_throw_error() {
+            let current_idx = start_index + i;
+            for exception in method_exceptions {
+                if current_idx >= exception.from_offset as usize
+                    && current_idx < exception.to_offset as usize
+                {
+                    // When branching as a result of an exception in a catch block,
+                    // the exception target will be run starting with an empty
+                    // scope stack and a stack with a single entry on it.
+                    let mut exception_stack = abstract_states[0].as_ref().unwrap().stack.clone();
```
Lord-McSweeney, line 725: Why do this instead of using the `Stack::empty` and `ScopeStack::empty` methods to create new stacks?

adrian17: Wouldn't I need to copy these lines?
```
    let empty_stack = Stack::new(method_body.max_stack as usize);
    let empty_scope_stack =
        ScopeStack::new((method_body.max_scope_depth - method_body.init_scope_depth) as usize);
```
I wanted to avoid this duplication.

Lord-McSweeney: Sure, it just seemed very confusing to me when I read it. Then could you please add a comment explaining that these create new empty stacks?

core/src/avm2/optimizer/optimize.rs
adrian17: This does mess with `abstract_states[0]` code for empty stack :/
(and isn't really significant for perf. But if you want, I can try refactor it more aggressively.)

core/src/avm2/optimizer/optimize.rs
adrian17: Yeah, I had two reasons. First is my intuition that "more than 2*usize -> pass by ref", second is that this makes it closer to what it'd look like if it was a `AbstractState` in the first place.
Entire `AbstractStateRef` is only here to simulate the same performance as if I refactored `abstract_interpret_ops` to operate on `AbstractState` all the time and were able to pass it without constructing it repeatedly; but without _actually_ refactoring the entirety of `abstract_interpret_ops`.

Lord-McSweeney (approved):

PR #20023, opened 2025-04-05
avm2: Fix classes of objects created by newactivation and newcatch

Lord-McSweeney:


core/src/avm2/op.rs
```diff
@@ -215,8 +215,7 @@ pub enum Op<'gc> {
         num_args: u32,
     },
     NewCatch {
-        #[collect(require_static)]
-        index: Index<Exception>,
```
adrian17, line 217: Personally I still like `Index`/`u32` in ops where we know it's gonna be u32 :/ But not blocking the change either.

adrian17 (approved):

PR #20027, opened 2025-04-06
avm1: migrate `StageObject` to `NativeObject`

moulins:
The display objects (`Avm1Button`, `EditText`, `MovieClip`, `Video`) are directly stored as `NativeObject` variants of their `ScriptObject`s, and the `StageObject`-specific data (namely, the AVM1 text field binding list) is moved directly inside the DO structs.

After this, only `SuperObject` will be left as a `Object` variant.

### DIFF
--- a/core/src/avm1.rs
+++ b/core/src/avm1.rs
@@ -32,7 +32,6 @@ pub use globals::array::ArrayBuilder;
 pub use globals::context_menu::make_context_menu_state;
 pub use globals::sound::start as start_sound;
 pub use object::script_object::ScriptObject;
-pub use object::stage_object::StageObject;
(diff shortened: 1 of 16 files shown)


Lord-McSweeney:
Shouldn't the text field binding functions be passed a `DisplayObject` instead of an `&impl TDisplayObject`?

moulins:
> Shouldn't the text field binding functions be passed a `DisplayObject` instead of an `&impl TDisplayObject`?

Good point, I don't remember exactly why I did this, but clearly serves no purpose in the final version of the code ^^

Lord-McSweeney (approved):

PR #20041, opened 2025-04-07
web: Don't use .npmrc for detect_chromedriver_version

danielhjacobs:
See https://github.com/giggio/node-chromedriver/issues/535


adrian17 (approved):

torokati44:
> I'm not actually sure the config is inherited in the child package.json files though

How could we find this out?

danielhjacobs:
Looking into it it seems the package.json is not inherited. So the real question is, which package.json files need this environment variable and which do not?

torokati44:
https://github.com/ruffle-rs/ruffle/actions/runs/14863409688/job/41734002396?pr=20362:

> `npm warn Unknown env config "detect-chromedriver-version". This will stop working in the next major version of npm.`

(This is not why it failed this time though.)

torokati44:
I tried to find a roadmap or expected release date for `npm` 12, but haven't found anything.
Just to see whether this is starting to get actual yet or not really...

torokati44:
> Looking into it it seems the package.json is not inherited. So the real question is, which package.json files need this environment variable and which do not?

According to https://www.npmjs.com/package/chromedriver#configuration-methods, this config key is (only) used by `npm install`, and we only do that in the toplevel `web/` directory AFAIK, so this should be alright and sufficient as is...? Am I reading this correctly?

torokati44:
Moved the new key a bit further down in the JSON (should not matter, other than aesthetically and nitpickingally). If you're okay with it, I'd say it's time to press the green button, @danielhjacobs. It will probably be fine... :registered: :tm:

PR #20047, opened 2025-04-08
avm2: Prevent mouse picking from targeting a loader's root object

ZingBallyhoo:
Modifies the logic added in https://github.com/ruffle-rs/ruffle/pull/16334 to also consider the case where the root object of a loader is no longer in the loader's hierarchy. This is common when using `Event.COMPLETE` + `event.target.content` instead of adding the loader itself as a child.


core/src/player.rs
```diff
@@ -3044,7 +3044,9 @@ fn run_mouse_pick<'gc>(
                 if let Avm2MousePick::Hit(target) =
                     l.mouse_pick_avm2(context, *context.mouse_position, require_button_mode)
                 {
-                    // Flash Player appears to never target events at the root object
+                    // Flash Player never targets events at the root object of a stage or loader.
+                    // If we are the root of a stage, target the stage instead (by returning None).
+                    // Avm2MousePick.combine_with_parent handles the root-of-loader case.
                     if !target.as_displayobject().is_root() {
                         res = Some(target);
```
ZingBallyhoo, line 3051: Any suggestions for how to assert that this is the root of a stage?

ZingBallyhoo: Found a better approach

ZingBallyhoo:
The new logic also makes `stage.mouseChildren = false` work so I added a test for it as a bonus.

ZingBallyhoo:
cc @adrian17 

From discord,
> // The caller does not expect the stage to be the target, instead return
> // None.

For context, there is a lot of code in `handle_input_event` that interprets None to mean the stage is the pick/hover target. 
https://github.com/ruffle-rs/ruffle/blob/4c68a993246992f247a124cd0f7ee26d9cea274e/core/src/player.rs#L1394
https://github.com/ruffle-rs/ruffle/blob/4c68a993246992f247a124cd0f7ee26d9cea274e/core/src/player.rs#L1534
https://github.com/ruffle-rs/ruffle/blob/4c68a993246992f247a124cd0f7ee26d9cea274e/core/src/player.rs#L1698
https://github.com/ruffle-rs/ruffle/blob/4c68a993246992f247a124cd0f7ee26d9cea274e/core/src/player.rs#L1710
https://github.com/ruffle-rs/ruffle/blob/4c68a993246992f247a124cd0f7ee26d9cea274e/core/src/player.rs#L1743
https://github.com/ruffle-rs/ruffle/blob/4c68a993246992f247a124cd0f7ee26d9cea274e/core/src/player.rs#L1756

Maybe it would be fine to treat the stage more like a regular object in `handle_input_event` and let it flow into `context.mouse_data.hovered` etc but I am not equipped to determine if that could be correct :). So at least for now `run_mouse_pick` is set up to fiddle with the result to keep the calling logic the same.

The use of `combine_with_parent(context.stage.into())` here now instead of the previous custom logic allows the `is_root` detection to be shared + make `stage.mouseChildren = false` work now too. 

The only cases the `stage->None` logic runs is: 
mouse is over root object - root objects can never be picked -> shared combine logic gives `stage` as the target
OR
mouse is over child of the stage - `stage.mouseChildren = false` -> shared combine logic gives `stage` as the target
[both cases from this one line of code](https://github.com/ruffle-rs/ruffle/blob/c65e92497eb3678bf3890b7f18145bc0553b086b/core/src/display_object/interactive.rs#L758)

core/src/player.rs
```diff
@@ -3036,19 +3036,22 @@ fn run_mouse_pick<'gc>(
...
-                    // If we are the root of a stage, target the stage instead (by returning None).
-                    // Avm2MousePick.combine_with_parent handles the root-of-loader case.
-                    if !target.as_displayobject().is_root() {
-                        res = Some(target);
+                let pick = l
+                    .mouse_pick_avm2(context, *context.mouse_position, require_button_mode)
+                    .combine_with_parent(context.stage.into());
+
+                match pick {
+                    Avm2MousePick::Hit(target) => {
```
adrian17, line 3044: I think `if let Avm2MousePick::Hit(target) =` would be clearer here. (and reduce indentation by 1 level)

tests/tests/swfs/avm2/stage_mousechildren/test.toml
```diff
@@ -0,0 +1 @@
+num_ticks = 1
```
adrian17, line 1: Small request: can please you add a file-ending newline?
I'll try setting .editorconfig to enforce this in the future.

adrian17 (approved):
LGTM, just sent some minor code-style notes.

adrian17:
Thank you!

Issue #20045, opened 2025-04-08
New Version of Extension Blocks JS, CSS, Pictures From Loading


az-123-none:
### Describe the bug

Chrome just auto updated the extension and then all web pages broke.

### Expected behavior

Let web pages display normally.

### Content Location

Unrelated to any specific Flash file.

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Google Chrome 125

Chad90b:
came to post this. i woke up today with no websites working because of the latest nightly build of ruffle that got pushed to the google extension market and auto-updated to my browser. ruffle 0.1.0 nightly 2025-04-08 breaks the web. please roll back.

danielhjacobs:
Someone on Discord also reported updating Chrome to have fixed the issues.

danielhjacobs:
I installed Chrome 125 to try to test this, but possibly since "simply re"installing" the extension seemed to have, at least at first glance, resolved the issues it previously had.", that was not enough to reproduce the issues others are reporting.

danielhjacobs:
As an update, I attempted to reproduce this issue by downloading an outdated version of Chrome (it only appears to affect outdated versions of Chrome; as the OP reported this issue occurred on Chrome 125 which is 11 months out of date), but because uninstalling and reinstalling the extension reportedly fixes the problem, I was unable to reproduce the issue, since that fresh install of Chrome 125 did not have Ruffle installed prior to April 8th and then auto-updated, which seems to be the trigger for this issue.

All I can do is repeat the suggestion to update your browser and/or uninstall and reinstall the Ruffle extension, as everyone who has done so has reported the issue to be fixed.

danielhjacobs:
Update: I have found the likely issue, and if I'm right it will likely be fixed tonight. It came from https://github.com/webpack/webpack/issues/19394, and webpack has fixed that and we just bumped to the new version.

danielhjacobs:
I believe this is now fixed.

PR #20048, opened 2025-04-08
web: Bump webpack to 5.99.5

danielhjacobs:
5.99.0 has regressions: https://github.com/webpack/webpack/issues/19394

Unsure if it's causing issues for us, but figured an upgrade cannot hurt. Perhaps relates to #20045, but no actual idea nor any idea how to test that.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -41,7 +41,7 @@
                 "typescript": "^5.8.3",
                 "typescript-eslint": "^8.29.1",
                 "webdriverio": "^9.2.2",
-                "webpack": "^5.99.0",
+                "webpack": "^5.99.5",
                 "webpack-cli": "^6.0.1"
             }
         },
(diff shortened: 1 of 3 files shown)


torokati44 (approved):

torokati44:
It's already being cooked: https://github.com/webpack/webpack/actions/runs/14336658120

torokati44:
And it's out: https://www.npmjs.com/package/webpack/v/5.99.3

torokati44:
And it's at 5.99.5 now... 👀

danielhjacobs:
I suspect #20045 is related. I finally found a new issue that started April 8th. It affects the extension on all Chrome versions (you can see as an error by installing it via zip in Chrome). It ties to this code: https://github.com/ruffle-rs/ruffle/blob/5aa7f0276a78e9c222ff927539dce86c8876c73d/web/packages/extension/src/background.ts#L57-L59

Prior to April 8th, that line in background.js became this:
```js
            await declarativeNetRequest.updateDynamicRules({
                removeRuleIds: [ruleId],
            });
```
As of the latest build, it's this:
```js
            await utils.declarativeNetRequest.updateDynamicRules({
                removeRuleIds: [ruleId],
            });
```

There is an error "Uncaught (in promise) ReferenceError: utils is not defined"

The webpack issue, https://github.com/webpack/webpack/issues/19394, is titled "Regression on v5.99.0, [something] not defined". The imports in the `try`/`catch` are replaced properly. The import in the `finally` is the issue.

danielhjacobs:
After all, if this rule is added but not removed, what happens (rhetorical)? Answer: Every resource is blocked:

https://github.com/ruffle-rs/ruffle/blob/5aa7f0276a78e9c222ff927539dce86c8876c73d/web/packages/extension/src/background.ts#L37-L50

torokati44:
I don't immediately see the connection between this rule, and the webpack issue that got fixed, but it's okay, I believe you!

danielhjacobs:
Part of the webpack issue was not doing replacements that should be done for imports.

It's the fact that in the `finally` specifically `await utils.declarativeNetRequest.updateDynamicRules` did not get changed to `await declarativeNetRequest.updateDynamicRules`, as webpack is supposed to do. The issue isn't that the rulecondition is added, it's that it's added AND not removed

torokati44:
Ah, okay, alright, thanks! It makes sense now.

PR #20054, opened 2025-04-09
avm2: hitTestPoint should only use local coordinates for the player root

ZingBallyhoo:
The implementation in FP only appears to mirror the avm1 root-relative-coordinates thing when the root is from the player, not a loader.

Ref: https://github.com/ruffle-rs/ruffle/pull/8933, https://github.com/ruffle-rs/ruffle/pull/192


Dinnerbone (approved):
Thank you!

torokati44:
If only that third commit was squashed into the first one... 🥹 But I'm not at home to do it... 🥴

torokati44:
Thank you for granting me my most pointlessly nitpicky wishes! 🥺🫶

PR #20065, opened 2025-04-11
debug_ui: Put frame list inside egui_extras::Table

crumblingstatue:
Replaces the current egui::Grid based layout, which tanks performance due to rendering everything.

Also adds a "Track current frame" checkbox to always scroll to the current frame. Makes it more convenient to keep track of where we are.


crumblingstatue:
Here is a demo to showcase keeping track of current frame.
(Makes it easy to bypass the URL checking of the showcases flash animation)

https://github.com/user-attachments/assets/3fc7beeb-3819-44e8-b523-5dcbd8cf65d9

kjarosh (approved):
Thank you, that looks and works great!

PR #20066, opened 2025-04-12
video/external: Don't say we are downloading OpenH264 when we use the cached version

Fullmetal5:
Thought for a second there was something wrong with Ruffle saving the OpenH264 files only to realize it just always printed the message.


adrian17 (approved):
Nice catch, thanks.

PR #20088, opened 2025-04-13
swf: Add and use `[read|write]_ubits_ct` for compile-time known width fields

torokati44:
This builds on https://github.com/ruffle-rs/ruffle/pull/20044.

Instead of renaming the methods with runtime width parameters to have the `_var` prefix, like `bitstream-io` did, I opted to add a compile-time parameter variant to read simple integers, because we also have `Fixed8` and `Twips` reader/writer methods, and it would have been too much churn to rename those too, for the sake of consistency.

Other suffix ideas to bikeshed:  `_fixed` (but it colludes with `fixed8`), `_cw` for "constant width", etc...

The `sbits` variants didn't seem to ever be used with literal constant bit widths.

### DIFF
--- a/swf/src/read.rs
+++ b/swf/src/read.rs
@@ -261,6 +261,18 @@ impl<'a, 'b> BitReader<'a, 'b> {
         self.bits.read_bit()
     }
 
+    /// A variant of `read_ubits` for when the number of bits to read is
+    /// a *c*ompile-*t*ime constant. This should be more efficient.
+    #[inline]
+    fn read_ubits_ct<const NUM_BITS: u32>(&mut self) -> io::Result<u32> {
+        if NUM_BITS > 0 {
+            self.bits.read::<NUM_BITS, u32>()
+        } else {
+            Ok(0)
+        }
+    }
+
+    /// If `num_bits` is a compile-time constant, consider `read_ubits_ct` instead.
     #[inline]
     fn read_ubits(&mut self, num_bits: u32) -> io::Result<u32> {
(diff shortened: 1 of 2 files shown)


swf/src/read.rs
```diff
@@ -261,6 +261,15 @@ impl<'a, 'b> BitReader<'a, 'b> {
         self.bits.read_bit()
     }
 
+    #[inline]
```
kjarosh, line 264: I'd include docs that describe the difference between `read_ubits` and `read_ubits_ct` & elaborate on what `_ct` means. I'd do the same for `read_ubits` too and include information that it's better to use `read_ubits_ct` when applicable.

This also applies to `write_ubits_ct`

kjarosh (approved):

swf/src/read.rs
```diff
@@ -261,6 +261,18 @@ impl<'a, 'b> BitReader<'a, 'b> {
...
+    fn read_ubits_ct<const NUM_BITS: u32>(&mut self) -> io::Result<u32> {
+        if NUM_BITS > 0 {
+            self.bits.read::<NUM_BITS, u32>()
+        } else {
+            Ok(0)
+        }
+    }
+
+    /// If `num_bits` is a compile-time constant, consider `read_ubits_ct` instead.
```
torokati44, line 275: I recognize that strictly speaking this is not what a docstring is supposed to say, but... :no_mouth: ... this one is ... just ... like this! :shrug:

kjarosh (approved):

PR #20094, opened 2025-04-13
debug_ui: More frame list related improvements

crumblingstatue:
- Add scene and label selection menu buttons to the table header
- Add a top menu action to quickly open the root `MovieClip`, if it exists
  This provides quick access the "main" frame list, to quickly seek to interesting parts of a movie.

On demand, I can split this into two pull requests, but hopefully they both make sense to merge.

## Showcase video

https://github.com/user-attachments/assets/87120124-42e5-4814-9ceb-36f091a7e992

### DIFF
--- a/desktop/assets/texts/en-US/main_menu.ftl
+++ b/desktop/assets/texts/en-US/main_menu.ftl
@@ -36,6 +36,7 @@ bookmarks-menu-manage = Manage Bookmarks...
 
 debug-menu = Debug Tools
 debug-menu-open-stage = View Stage Info
+debug-menu-open-root-movie-clip = View Root MovieClip
 debug-menu-open-movie = View Movie
 debug-menu-open-movie-list = Show Known Movies
 debug-menu-open-domain-list = Show Domains
--- a/desktop/src/gui/menu_bar.rs
+++ b/desktop/src/gui/menu_bar.rs
@@ -5,6 +5,7 @@ use crate::player::LaunchOptions;
 use crate::preferences::GlobalPreferences;
 use egui::{menu, Button, Key, KeyboardShortcut, Modifiers, Widget};
 use ruffle_core::config::Letterbox;
+use ruffle_core::focus_tracker::DisplayObject;
 use ruffle_core::{Player, StageScaleMode};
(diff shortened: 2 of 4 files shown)


desktop/assets/texts/en-US/main_menu.ftl
```diff
@@ -36,6 +36,7 @@ bookmarks-menu-manage = Manage Bookmarks...
 
 debug-menu = Debug Tools
 debug-menu-open-stage = View Stage Info
+debug-menu-open-root-movie-clip = View Root MovieClip
```
crumblingstatue, line 39: I don't know the "protocol" when adding new translatable  text.
I only added the English translation, hopefully that's enough for starters.

desktop/src/gui/menu_bar.rs
```diff
@@ -135,6 +136,17 @@ impl MenuBar {
                                 player.debug_ui().queue_message(DebugMessage::TrackStage);
                             }
                         }
+                        if let Some(player) = &mut player {
```
crumblingstatue, line 139: I made this a button that's disabled if the root movie clip doesn't exist.
That makes the code for it stick out compared to the other buttons, but hopefully that's not an issue.

I can make it a "regular" button if desired, but then the user might click the button expecting something to happen,
but nothing happens if the root movie clip doesn't exist.
There could be some kind of popup message indicating that the root clip doesn't exist though.
But a disabled button is probably the cleanest solution from a UX standpoint.

Lord-McSweeney (approved):

desktop/assets/texts/en-US/main_menu.ftl
kjarosh: That's the exact protocol, rest of translations will come in through Crowdin

desktop/src/gui/menu_bar.rs
kjarosh: > That makes the code for it stick out compared to the other buttons, but hopefully that's not an issue.

It's not only not an issue but also the preferred way to do it!

You can use `on_hover_text_at_pointer` with an explanation when the button is disabled, but I think it's fine.

PR #20157, opened 2025-04-13
chore: Unify cargo versions across packages

kjarosh:
Make sure that workspace versions are used where applicable.


torokati44 (approved):
I guess, why not!

PR #20170, opened 2025-04-13
desktop: Remove unused `nightly` cfg

kjarosh:
Remove setting the `nightly` cfg option from build.rs, as it's not used.


torokati44 (approved):
I haven't found any uses of it either... :shrug:

PR #20188, opened 2025-04-14
avm2: Shrink Method to 8 bytes by unifying BytecodeMethod and NativeMethod

Lord-McSweeney:
`Method` is now a single struct (stored behind a Gc). We store separate native and bytecode method info in a `MethodKind` enum.

### DIFF
--- a/core/src/avm2/globals/Function.as
+++ b/core/src/avm2/globals/Function.as
@@ -44,9 +44,13 @@ package {
         public native function get prototype():*;
         public native function set prototype(proto:*):*;
 
+        AS3 native function apply(receiver:* = void 0, args:* = void 0):*;
         AS3 native function call(receiver:* = void 0, ...rest):*;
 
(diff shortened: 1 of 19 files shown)


core/src/avm2/globals/Function.as
```diff
@@ -44,9 +44,13 @@ package {
         public native function get prototype():*;
         public native function set prototype(proto:*):*;
 
+        AS3 native function apply(receiver:* = void 0, args:* = void 0):*;
         AS3 native function call(receiver:* = void 0, ...rest):*;
 
-        AS3 native function apply(receiver:* = void 0, args:* = void 0):*;
+        [Ruffle(NativeCallable)]
+        private static function createDummyFunction():Function {
+            return function() { };
```
adrian17, line 52: to be sure, this exists on the AS side mostly because we don't have an utility to create a FunctionObject from a native function?

adrian17 (approved):

PR #20202, opened 2025-04-16
avm2: Add peephole optimization passes

Lord-McSweeney:
This adds two peephole optimization passes to the bytecode optimizer. The first runs before block assembly and eliminates no-op `iftrue`/`iffalse`/`jump` ops. The second runs after the type-aware optimizer and eliminates more ops:
1. Removes `getlocal`+`pop`
2. Replaces `getscopeobject 0` with `getlocal 0` when possible, to enable the next optimization:
3. Removes the `getlocal 0` + `pushscope` at the start of methods when possible.

On most SWFs, this allows for removing almost all instances of the `getlocal 0` + `pushscope` sequence at the start of each function.

We also now remove no-op `getlocal 0` + `constructsuper 0` sequences in instance initializers of classes that just extend `Object`.

ChessDemons starts up ~8% faster.


core/src/avm2/optimizer/optimize.rs
```diff
@@ -1500,6 +1500,18 @@ fn abstract_interpret_ops<'gc>(
 
                 // Then receiver.
                 stack.pop(activation)?;
+
+                // Remove `super()` calls in classes that extend Object, since they
+                // are noops anyway.
+                if num_args == 0 {
+                    let object_class = activation.avm2().classes().object;
+                    if activation
+                        .bound_superclass_object()
```
adrian17, line 1526: Can you `//TODO` that a None superclass object eventually should VerifyError?

core/src/avm2/optimizer/optimize.rs
```diff
@@ -1500,6 +1500,18 @@ fn abstract_interpret_ops<'gc>(
...
+                        .is_some_and(|c| c == object_class)
+                    {
+                        optimize_op_to!(Op::Pop);
```
adrian17, line 1529: Nit: technically this doesn't check for cases like
```
pushnull
constructsuper 0
```
which AFAIK still run in FP? Though I don't know how much to care - maybe we can just `//TODO` this too.
Alternatively, we could do a different optimization and make `fn super_init()` conditionally noop at runtime (like, have the Class or constructor Method be marked with some `is_noop` flag instead of checking `type == object`. Or maybe make noop init_methods None or something. This could also allow propagating this optimization to `class C{}`'s constructor, not just Object.).

core/src/avm2/activation.rs
```diff
@@ -554,14 +554,14 @@ impl<'a, 'gc> Activation<'a, 'gc> {
     /// outer scope. If the outer scope is empty, we use the bottom
     /// of the current scope stack instead.
     ///
-    /// A return value of `None` implies that both the outer scope, and
-    /// the current scope stack were both empty.
-    pub fn global_scope(&self) -> Option<Value<'gc>> {
+    /// The verifier guarantees that there is always a global scope
+    /// when this function is called.
+    pub fn global_scope(&self) -> Value<'gc> {
```
adrian17, line 557: I kinda thought that avmplus supported this edge case, but I can also see the verifier check... weird. (not an actionable note, just thinking loudly)

core/src/avm2/optimizer/peephole.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    let simple_scope_op_positions =
+        simple_scope_structure(ops, jump_targets).filter(|_| !has_exceptions);
+
+    // Now actually run the peephole optimizer.
+    let mut last_op = None;
+
+    for (i, current_op) in ops.iter().enumerate() {
+        if jump_targets.contains(&i) {
```
adrian17, line 70: If the metadata gathering could be done earlier, then the peephole could be also run in-line (as in, together with `abstract_interpret_ops` or even inside it) and then the `jump_targets.contains` check wouldn't be needed, as the work would be per-block-slice already.

What do you think about it? It feels slightly more optimal/nicer as an idea, but is also some work to refactor and might be messier in practice.

core/src/avm2/optimizer/optimize.rs
Lord-McSweeney: I can change it to handle it by ensuring the object is guaranteed not-null before applying the optimization, which will always save this optimization for `getlocal0` + `constructsuper 0` anyway because we mark local 0 as not-null.

core/src/avm2/optimizer/peephole.rs
Lord-McSweeney: It is definitely more efficient, but I think it's cleaner to have all the peephole logic separated from the type-aware optimizer code

PR #20210, opened 2025-04-17
desktop: more fallback cjk fonts for Linux

lilydjwg:
Source Han Sans is another popular CJK font that has native font names and uses glyphs according to the locale (not sure if  ruffle supports it, but e.g. Firefox supports it so it makes things easier for me than Noto Sans CJK).

WenQuanYi Zen Hei is an old Chinese font and maybe someone still has only it.

PS: I hope ruffle could use fontconfig rather than hardcode more & more font names...

### DIFF
--- a/desktop/src/gui/controller.rs
+++ b/desktop/src/gui/controller.rs
@@ -541,7 +541,9 @@ fn load_system_fonts(
     }
     families.extend(
         [
-            Family::Name("Arial Unicode MS"), // macos
+            Family::Name("Source Han Sans"),   // linux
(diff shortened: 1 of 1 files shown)


kjarosh (approved):
In theory we should put those fonts behind the `zh` language, but TBH... the whole `load_system_fonts` method looks like it should be rewritten and I don't see how this change can break things.

kjarosh:
> PS: I hope ruffle could use fontconfig rather than hardcode more & more font names...

For the record, we use `fontdb` for that (which is not as good as `fontconfig` of course). But I believe the main issue is that we provide one font only without fallbacks.

Issue #17215, opened 2024-07-22
white screen on Tamagotchi Friends: Dream Town


LiEnby:
### Describe the bug

just like try play the game
![image](https://github.com/user-attachments/assets/0553dd57-4edf-46d2-afc5-50635633c15c)

### Expected behavior

login or play as guest screen to appear

### Content Location

https://famitama.xyz/friends/index.html

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Firefox 

### Additional information

![image](https://github.com/user-attachments/assets/0705eb3d-7d66-4fc1-a43b-de4aba0851f9)

n0samu:
We're getting a stack overflow - this also happens in the desktop app, but the game works fine in FP. Here's the SWF URL to test:
https://famitama.xyz/friends/prelauncher.swf

LiEnby:
the preloader.swf loads in another swf from https://famitama.xyz/friends/tam.cebd that is encrypted with RC4.

the swf contents are : [tam.zip](https://github.com/user-attachments/files/16327688/tam.zip)

(but i don't think it ever actually gets that far in ruffle)

Lord-McSweeney:
This is happening in the function `com.frimastudio.as3ToolKit.event::LoaderEvent/set bytesTotal`:
```
        override public function set bytesTotal(param1:Number) : void
        {
            param1 = param1 == 0 ? 1 : param1;
            super.bytesTotal = param1;
            this.ProgressPercent;
        }
```

I would guess that the `setsuper` call interacts weirdly with the override.

Lord-McSweeney:
After #20233 this needs #16426.

Lord-McSweeney:
After #21514 the error is
```
ReferenceError: Error #1069: Property uploadFromBitmapData not found on flash.display3D.textures.RectangleTexture and there is no default value.
```

PR #20233, opened 2025-04-20
avm2: Make (almost) all class signatures match FP

Lord-McSweeney:
Progresses #17215 (it needs #16426 next) and unblocks more class verification

### DIFF
--- a/core/src/avm2/function.rs
+++ b/core/src/avm2/function.rs
@@ -201,7 +201,6 @@ pub fn exec<'gc>(
             }
 
             let signature = &*method.resolved_param_config();
-            let return_type = method.resolved_return_type();
 
             // Check for too many arguments
             if arguments.len() > signature.len() && !method.is_variadic() && !method.is_unchecked()
@@ -233,14 +232,7 @@ pub fn exec<'gc>(
                 .avm2
                 .push_call(activation.gc(), method, bound_class);
 
-            let result = native_method(&mut activation, receiver, &arguments);
-            result.and_then(|v| {
-                if let Some(return_type) = return_type {
-                    v.coerce_to_type(&mut activation, return_type)
-                } else {
-                    Ok(v)
-                }
-            })
+            native_method(&mut activation, receiver, &arguments)
         }
         MethodKind::Bytecode(_) => {
             // This used to be a one step called Activation::from_method,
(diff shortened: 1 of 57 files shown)


Lord-McSweeney:
Unfortunately this uncovered an issue with our TextLine implementation. By removing a getter that wasn't supposed to be there, one of the tests ends up calling a DisplayObjectContainer method on a TextLine:
```
pub fn get_tab_children<'gc>(
    this: Value<'gc>,
    _args: &[Value<'gc>],

    if let Some(obj) = this
        .as_display_object()
        .and_then(|this| this.as_container())
    {
        Ok(Value::Bool(obj.is_tab_children(activation.context)))
    } else {
        Ok(Value::Undefined)
    }
}
```
This was returning `undefined`. As a stopgap, we now explicitly coerce the return values of native methods, so that `undefined` gets coerced to the expected `false`. This is a hack and we do need to correctly implement the TextLine display object sometime.

evilpie:
Could we test this in CI? Do you have some comparison scripts or similar?

Lord-McSweeney:
I actually wrote a comparison script to find these, but not all of our classes exactly match FP- this PR only fixes the differences for non-`final` classes, since those are the only ones where user code can actually VerifyError if it overrides methods that exist in FP's playerglobals but not ours. I don't think it's worth testing the `describeType` output for a limited set of our classes.

core/src/avm2/function.rs
```diff
@@ -231,7 +232,15 @@ pub fn exec<'gc>(
                 .context
                 .avm2
                 .push_call(activation.gc(), method, bound_class);
-            native_method(&mut activation, receiver, &arguments)
+
+            let result = native_method(&mut activation, receiver, &arguments);
+            result.and_then(|v| {
+                if let Some(return_type) = return_type {
+                    v.coerce_to_type(&mut activation, return_type)
```
adrian17, line 239: This is still the one change I'm not a fan of.
I know that the "fast call" one technically skips this, but then - since neither of these actually does a verification (and thus it's kinda arbitrary that CallMethod does but CallNative doesn't), it feels to me like it'd make more sense to just do a return type assertion in debug builds for both call types instead of adding a cost in runtime.

Aside from this, the rest of PR looks fine.

Lord-McSweeney: Sure, I'll remove the check. My main reason for keeping this in was on the basis that future native functions might accidentally return the wrong type, but thinking back on it we probably won't be adding very many more native functions now.

Issue #12685, opened 2023-08-11
fl0w: Sounds do not play


n0samu:
### Describe the bug

Download the game: [fl0w.zip](https://github.com/ruffle-rs/ruffle/files/12318879/fl0w.zip)
[Also available online here](https://ooooooooo.ooo/static/?47ad0667-9626-4f19-868f-1a3971e9a045).

Fl0w (or Flow) seems to be playable, but none of the sounds play. This seems to be because the game thinks the sounds are finished downloading before they actually are, and moves on to request the next one, leading to a flood of requests, and to Sound.start being called too early. Warnings shown are:
```
WARN run_frame: ruffle_core::stub: Encountered stub: AVM1 Sound.getBytesTotal()
WARN run_frame: ruffle_core::stub: Encountered stub: AVM1 Sound.getBytesLoaded()
WARN run_frame: ruffle_core::avm1::globals::sound: Sound.start: No sound is attached
```
Also. the Ruffle desktop app spams these warnings during sound loading:
```
WARN isahc::handler: request completed with error: unknown error`
WARN isahc::handler: request completed with error: failed to connect to the server
```
I think that may simply be because it tries to fire off more requests at a time than it can handle.

### Expected behavior

The game should load only one sound at a time and shouldn't try to play a sound that hasn't loaded yet. The game compares `getBytesLoaded()` and `getBytesTotal()` for each sound to determine if it's done loading (and the next sound can be loaded). Ruffle has these functions stubbed to always return -1, leading to the weird behavior mentioned.

### Affected platform

Desktop app

### Operating system

Windows 10

droserasprout:
I have found another affected game, Flash Guitar Hero Online. The same behavior and logs on Ruffle 0.1.0-nightly (1c8e95a061a40ccd4b2fdca30a66eb7cc2fbe12d 2024-07-19).

PR #20235, opened 2025-04-21
avm1: implement Sound getBytesTotal(), getBytesLoaded()

InvoxiPlayGames:
This PR adds more correct functionality to getBytesLoaded() and getBytesTotal() - returning the actual number of bytes loaded from the network / from disk rather than always returning 1, as well as returning undefined on getBytesLoaded when loadSound has not been called and on getBytesTotal when the total length is unknown (request not completed).

This fixes https://old.homestarrunner.com/qod.swf - with extra debug logging added to aforementioned Flash file, the trace output between Ruffle and Flash matches with this PR applied.

### DIFF
--- a/core/src/avm1/globals/sound.rs
+++ b/core/src/avm1/globals/sound.rs
@@ -43,6 +43,12 @@ struct SoundData<'gc> {
     /// Duration of the currently attached sound in milliseconds.
     duration: Cell<Option<u32>>,
 
+    /// Size of the loaded sound buffer in bytes.
+    bytes_loaded: Cell<Option<u32>>,
+
+    /// Total size of the sound in bytes.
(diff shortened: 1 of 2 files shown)


n0samu:
Could someone test #12685? This is a very exciting development, thank you!!

adrian17 (commented):
The code looks fine - that said, couldn't we mirror what AVM2 does here? As in, in avm2 `sound.bytesLoaded/bytesTotal`, we obtain the SoundHandle from the object and call `audio.get_sound_size`. Does this make any difference in behavior vs storing the state up front?

InvoxiPlayGames:
> Does this make any difference in behavior vs storing the state up front?

I'm unsure - I did it this way due to the way the function call returns `undefined` in different scenarios.

PR #20242, opened 2025-04-22
avm2: fix Loader behaviour when body is empty

### LINKED ISSUE #18297: Error on https://play.cpimagined.net
### Describe the bug

When loading the minigame "Card-Jitsu Snow" inside of Club Penguin, HTTP requests are failing because they're sending POST requests instead of GET requests, leading to the game failing to load everything, and even displaying the default Metaplace client error messages, which are normally never seen.

Image

### Expected behavior

The game loads to the player select screen successfully.

Image

### Content Location

https://media.cpimagined.net/game/mpassets//playclients/r3662/flash_client_base_fp11.swf `com.metaplace.central.net.webRequest:WebLoader/execQueuedLoad`

### Affected platform

Browser's extension

### Operating system

Windows 11 23H2 (22631.4317)

### Browser

Arc 1.24.0.54812 (Chromium 130.0.6723.44)

### Additional information

~~Until #18296 is fixed, this issue isn't the easiest to debug, as the arguments aren't being passed correctly. There isn't an easy way to bypass this because you need your login token for the API routes, and that's stored internally by the game, or externally on Redis. If you are running a local server, you can use the HTML file inside of [Snow.zip](https://github.com/user-attachments/files/17412590/Snow.zip) to skip that step.~~ As of nightly 2024-12-14, #18296 was sorted, so now you should be able to check it out without any issues. If you get to the "Already Logged In" screen (shown below), then you've likely fixed the issue (screenshot is using a proxy as a temporary fix)

Image

AllinolCP:
Currently in ruffle, when you do a URLRequest with method `POST` it will send a `POST` request regardless of the content.
This isn't the case with flash. In flash it will change a `POST` request to a `GET` when the body is null or empty.

~~This also includes another behaviour fix, in flash the header only get sent if the request is a `POST` request, will not send headers otherwise (documentation claims it only happens on browsers but it doesn't send headers for a GET request in desktop flash.)~~ Doesn't seem to be the case with FP projector.

Closes #18297


AllinolCP:
image
Nevermind. turns out this was documented.

ERmilburn02:
> This should fix #18297

Confirmed, this will fix #18297

evilpie:
Can you add a test?

core/src/avm2/globals/flash/display/loader.rs
```diff
@@ -193,23 +193,38 @@ pub fn request_from_url_request<'gc>(
...
+    // When payload is null, flash will ignore the method and do a GET request instead.
+    if body.is_none() {
+        method = NavigationMethod::Get;
+    }
+
     let mut request = Request::request(method, url.to_string(), body);
-    request.set_headers(string_headers);
+
+    // Even though the documentation said that on browsers it sends headers only through POST
+    // It also doesn't send headers on GET request on Flash 35 (Adobe Animate)
```
adrian17, line 224: I think I tested it in the past and got different results, but might have done something different. Will retest just to make sure.

adrian17 (commented):
LGTM, but wanna test FP before approving.

adrian17:
@evilpie I'm not sure we're able to test this as-is?

AllinolCP:
@evilpie I've added tests. Though it seems that in ruffle an empty bytearray sends nothing while flash it sends random characters.

core/src/avm2/globals/flash/display/loader.rs
AllinolCP: I've checked FP in Flash/Animate this is seems to be the case. But on flash projetor it actually sends headers so i'll remove this "fix".

evilpie:
Thank you for adding the test. I guess you could remove the headers from it now?

evilpie:
> Though it seems that in ruffle an empty bytearray sends nothing while flash it sends random characters.

That sound suspiciously like Flash reading uninitialized memory.

AllinolCP:
> Thank you for adding the test. I guess you could remove the headers from it now?

@evilpie removed!

evilpie (approved):
Thanks!

adrian17:
LGTM, but let's squash before/on merge.

PR #20250, opened 2025-04-23
extension: Add exclusion for TikTok

n0samu:
Problem was reported via email:
image


danielhjacobs:
Unfortunately we have some code duplication when it comes to this:

https://github.com/n0samu/ruffle/blob/patch-13/web/packages/extension/src/background.ts#L193

adrian17 (approved):

adrian17:
> Unfortunately we have some code duplication when it comes to this:
> 
> https://github.com/n0samu/ruffle/blob/patch-13/web/packages/extension/src/background.ts#L193

What do you mean?

danielhjacobs:
We have to separately specify pages on which we do not want to polyfill the plugin and pages on which we do not want to execute ruffle itself. We specify the former in the background script and the latter in the manifest. It's the same list of pages.

adrian17:
Oh, right, I didn't even notice that these are separate files :/

Then yeah, still approved, but assuming the extra line is added there too.

n0samu:
Amended, thanks!

PR #20255, opened 2025-04-24
extension: Stop injecting scripts on any browser except Firefox 127 and below

danielhjacobs:
No browser specific setting exists for Chrome but this raises the minimum version for it to 102 for the extension. Minimum Safari extension version unchanged as it was already bumped to 16.4 previously.

Edit: In the future, we should be able to remove script injection altogether, as soon as Windows 7 - 8.1 support is dropped by Mozilla.

### DIFF
--- a/web/packages/extension/src/content.ts
+++ b/web/packages/extension/src/content.ts
@@ -1,21 +1,10 @@
 /**
- * Pierce the extension sandbox by copying our code into main world.
  *
- * The isolation extension content scripts get is neat, but it causes problems
- * based on what browser you use:
- *
- * 1. On Chrome, you are explicitly banned from registering custom elements.
- * 2. On Firefox, you can register custom elements but they can't expose any
- *    useful API surface, and can't even see their own methods.
- *
- * This code exists to pierce the extension sandbox, while maintaining:
- *
- * 1. The isolation of not interfering with the page's execution environment
- *    unintentionally.
- * 2. The ability to load extension resources such as .wasm files.
- *
- * We also provide a content script message listener that proxies messages
+ * This code provides a content script message listener that proxies messages
  * into/from the main world.
+ *
+ * On older Firefox, it also pierces the extension sandbox by copying our code into the main world
+ *
  */
 
 import * as utils from "./utils";
@@ -169,9 +158,8 @@ function isXMLDocument(): boolean {
             ?.filename !== "ruffle.js"
     ) {
         injectScriptRaw("%PLUGIN_POLYFILL_SOURCE%");
+        await injectScriptURL(utils.runtime.getURL("dist/ruffle.js"));
     }
-
-    await injectScriptURL(utils.runtime.getURL(`dist/ruffle.js?id=${ID}`));
 
     window.addEventListener("message", (event) => {
         // We only accept messages from ourselves.
--- a/web/packages/extension/src/ruffle.ts
+++ b/web/packages/extension/src/ruffle.ts
@@ -1,6 +1,16 @@
 import { Setup, setCurrentScriptURL } from "ruffle-core";
 import { Message } from "./messages";
 
+/**
+ *
+ * This script runs in the MAIN ExecutionWorld for the following reasons:
+ *
+ * 1. On Chrome, you are explicitly banned from registering custom elements.
+ * 2. On Firefox, you can register custom elements but they can't expose any
+ *    useful API surface, and can't even see their own methods.
+ *
+ */
+
 // Current message ID to be included in openInNewTab
 let currentMessageId: string | null = null;
 
--- a/web/packages/extension/webpack.config.js
+++ b/web/packages/extension/webpack.config.js
@@ -104,7 +104,9 @@ export default function (/** @type {Record<string, any>} */ env, _argv) {
         },
         output: {
             path: url.fileURLToPath(new URL("assets/dist/", import.meta.url)),
-            publicPath: "auto",
+            // publicPath: "auto" throws for content scripts, which lack a script src
+            // This is changed at runtime to the full URL of the extension's /dist/ folder
+            publicPath: "/dist/",
             clean: true,
             assetModuleFilename: "assets/[name][ext][query]",
         },
--- a/web/packages/extension/src/background.ts
+++ b/web/packages/extension/src/background.ts
@@ -175,24 +175,35 @@ async function enable() {
         return;
     }
     if (!(await contentScriptRegistered())) {
+        const excludeMatches = [
+            "https://sso.godaddy.com/*", // See https://github.com/ruffle-rs/ruffle/pull/7146
+            "https://authentication.td.com/*", // See https://github.com/ruffle-rs/ruffle/issues/2158
+            "https://*.twitch.tv/*", // See https://github.com/ruffle-rs/ruffle/pull/8150
+            "https://www.tuxedocomputers.com/*", // See https://github.com/ruffle-rs/ruffle/issues/11906
+            "https://*.taobao.com/*", // See https://github.com/ruffle-rs/ruffle/pull/12650
+            "https://*.time4learning.com/*", // See https://github.com/ruffle-rs/ruffle/pull/16186
+            "https://*.edgenuity.com/*", // See https://github.com/ruffle-rs/ruffle/pull/16186
+            "https://www.chewy.com/*", // See https://github.com/ruffle-rs/ruffle/issues/18265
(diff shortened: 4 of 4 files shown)


danielhjacobs:
Future TODO:

Put the logic that can be ISOLATED in a separate isolated content script. There are parts that can't with the current logic, as the extension modifies the `window.RufflePlayer` on the page.

web/packages/extension/src/content.ts
```diff
@@ -42,8 +22,9 @@ const ID = Math.floor(Math.random() * 100000000000);
  */
 function sendMessageToPage(data: unknown): Promise<unknown> {
     const message = {
-        to: `ruffle_page${ID}`,
+        to: "ruffle_page",
         index: pendingMessages.length,
+        id: ID,
```
danielhjacobs, line 27: The ID is now passed as part of the message, as it can't be part of a script src since content scripts don't have a src.

web/packages/extension/src/content.ts
```diff
@@ -158,32 +107,16 @@ function isXMLDocument(): boolean {
...
             return;
         }
 
-        const { to, index, data } = event.data;
-        if (to === `ruffle_content${ID}`) {
+        const { to, index, data, id } = event.data;
+        if (to === "ruffle_content") {
             const request = index !== null ? pendingMessages[index] : null;
-            if (request) {
+            if (request && id === ID) {
```
danielhjacobs, line 119: Only forward messages that have the same ID as this content script (fulfills the same purpose as when previously we only looked at messages to `ruffle_content${ID}`)

web/packages/extension/src/content.ts
```diff
@@ -158,32 +107,16 @@ function isXMLDocument(): boolean {
...
                 pendingMessages[index] = null;
                 request.resolve(data);
             } else if (isMessage(data)) {
```
danielhjacobs, line 122: Ideally this would have the same `id === ID` check, but it's harder to add that for "open_url_in_player" as any use of `openInNewTab` would need to know the assigned ID. What's the worst that can happen without it (genuine question)?

web/packages/extension/src/content.ts
```diff
@@ -209,5 +142,6 @@ function isXMLDocument(): boolean {
             unmuteOverlay: options.autostart ? "hidden" : "visible",
             splashScreen: !options.autostart,
         },
+        publicPath: utils.runtime.getURL("/dist/"),
```
danielhjacobs, line 145: Pass the full URL of the extension's /dist/ folder to the ruffle content script.

web/packages/extension/webpack.config.js
```diff
@@ -99,7 +100,7 @@ export default function (/** @type {Record<string, any>} */ env, _argv) {
         },
         output: {
             path: url.fileURLToPath(new URL("assets/dist/", import.meta.url)),
-            publicPath: "auto",
+            publicPath: "/dist/",
```
danielhjacobs, line 103: This gets changed during runtime to the full URL of the extension's /dist/ folder.

web/packages/extension/src/ruffle.ts
```diff
@@ -1,9 +1,20 @@
...
+ *
+ */
+
 import { Setup } from "ruffle-core";
 import { Message } from "./messages";
 
 function handleMessage(message: Message) {
     switch (message.type) {
         case "load": {
+            __webpack_public_path__ = message.publicPath;
```
danielhjacobs, line 17: Change the webkit public path to the full URL of the extension's /dist/ folder.

web/packages/extension/src/content.ts
```diff
@@ -158,30 +107,14 @@ function isXMLDocument(): boolean {
...
     window.addEventListener("message", (event) => {
         // We only accept messages from ourselves.
         if (event.source !== window || !event.data) {
             return;
         }
 
-        const { to, index, data } = event.data;
-        if (to === `ruffle_content${ID}`) {
+        const { to, index, data, id } = event.data;
+        if (to === "ruffle_content" && id === ID) {
```
danielhjacobs, line 117: Only look at messages with the same id as was generated (fulfills the same purpose as when previously we were checking only messages to `ruffle_content${ID}`)

danielhjacobs:
This may make some of the excluded sites no longer need to be excluded. However I am unsure and none of them have flash content anyway.

danielhjacobs:
The latest commit is unfortunate, but does let this PR work on ESR 115. It adds back the script injection, but only on Firefox versions below 128. It should be reverted when Firefox officially drops support for Windows 7 and 8.1: https://support.mozilla.org/en-US/kb/firefox-users-windows-7-8-and-81-moving-extended-support

web/packages/extension/src/ruffle.ts
evilpie: Could we check that this is using an extension protocol?

evilpie:
So the whole purpose at this point is to make masked Webkit's masked URLs work? There are so many different changes and in this that it gets confusing.

danielhjacobs:
I can split the PR, but that's not my only purpose, no. My goal was moreso to stop injecting script on the page altogether and instead do as extension developers are supposed to do and use content scripts, in the MAIN ExecutionWorld as necessary. It prevents us from doing any page tampering on pages without Flash elements, therefore probably making us less likely to trip anti-tampering scripts.

If I split it, part one will just be the ruffle.ts aside from the `handleMessage` changes and the content.ts message changes. That would fix Safari on its own and be cleaner.

Part two will be the rest.

web/packages/core/src/current-script.ts
```diff
@@ -24,3 +24,13 @@ try {
 } catch (_e) {
     console.warn("Unable to get currentScript URL");
 }
+
+/**
+ * Sets the current script URL and isExtension boolean manually when using the extension, which lacks a script src.
+ *
+ * @param src The location of Ruffle's resources within the extension.
+ */
+export function setCurrentScriptURL(src: string) {
```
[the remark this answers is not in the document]
danielhjacobs: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getURL#return_value is a string though, should I `new URL` it before calling this function?

web/packages/extension/src/ruffle.ts
danielhjacobs: If I do the `new URL` here, yes. https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getURL#return_value is a string, and it's hard to check the protocol of a string.

danielhjacobs:
The changes needed for Safari have been split into #20264.

web/packages/core/src/current-script.ts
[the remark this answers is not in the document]
danielhjacobs: Resolved in #20264

web/packages/extension/src/ruffle.ts
danielhjacobs: Resolved in #20264

evilpie:
> It prevents us from doing any page tampering on pages without Flash elements, therefore probably making us less likely to trip anti-tampering scripts.

That is not really true, at least plugin-polyfill.ts still tampers with Navigator object.

web/packages/extension/webpack.config.js
```diff
@@ -104,7 +104,9 @@ export default function (/** @type {Record<string, any>} */ env, _argv) {
         },
         output: {
             path: url.fileURLToPath(new URL("assets/dist/", import.meta.url)),
-            publicPath: "auto",
+            // publicPath: "auto" throws for content scripts, which lack a script src
+            // This is changed at runtime to the full URL of the extension's /dist/ folder
+            publicPath: "/dist/",
```
evilpie, line 109: Thanks for adding this comment!

evilpie (changes requested):
I did some local testing and realized "ruffle" is missing from `scripting.unregisterContentScripts` in `disable`.

danielhjacobs:
Ah, right, I think I lost that in the rebase

danielhjacobs:
Fixed

danielhjacobs:
> That is not really true, at least plugin-polyfill.ts still tampers with the Navigator object.

Fair point. We could not do that when we don't do the element polyfill but if the longer we wait to do the plugin polyfill (by waiting to check what elements are on the page) the more likely Flash detection scripts run before the polyfill. Still, tampering less is surely better.

evilpie (approved):

PR #20263, opened 2025-04-25
desktop: Fix missing characters in UI

kjarosh:
Before this patch, we were loading only one font, based on the OS and language selected. This approach was not perfect, as the UI can have characters from multiple fonts, irrespectively of selected language.

This patch changes the way fonts are loaded, and provides multiple fallback fonts for different categories of characters (Korean, Chinese, Japanese, etc.).

This patch makes it possible to see characters from all categories at the same time in UI irrespectively of selected language. Additionally, it adds Hebrew and Arabic fonts.

Before:

image

After:

image

It's a follow-up to https://github.com/ruffle-rs/ruffle/pull/20210

### DIFF
--- a/desktop/src/gui/controller.rs
+++ b/desktop/src/gui/controller.rs
@@ -575,6 +575,22 @@ fn load_system_fonts(
         ],
     ));
 
+    // Hebrew
+    queries.push((
+        4,
+        vec![
+            Family::Name("Noto Sans Hebrew"), // Open font
+        ],
+    ));
+
+    // Arabic
+    queries.push((
+        5,
+        vec![
+            Family::Name("Noto Sans Arabic"), // Open font
+        ],
+    ));
+
     register_family(
         font_database,
         &mut fd,
--- a/desktop/src/gui/controller.rs
+++ b/desktop/src/gui/controller.rs
@@ -6,7 +6,7 @@ use crate::gui::{RuffleGui, MENU_HEIGHT};
 use crate::player::{LaunchOptions, PlayerController};
 use crate::preferences::GlobalPreferences;
 use anyhow::anyhow;
-use egui::{Context, ViewportId};
+use egui::{Context, FontData, FontDefinitions, ViewportId};
 use fontdb::{Database, Family, Query, Source};
 use ruffle_core::events::{ImeCursorArea, ImePurpose};
 use ruffle_core::{Player, PlayerEvent};
@@ -15,7 +15,6 @@ use ruffle_render_wgpu::descriptors::Descriptors;
 use ruffle_render_wgpu::utils::{format_list, get_backend_names};
 use std::sync::{Arc, MutexGuard};
 use std::time::{Duration, Instant};
-use unic_langid::LanguageIdentifier;
 use url::Url;
 use wgpu::SurfaceError;
 use winit::dpi::{PhysicalPosition, PhysicalSize};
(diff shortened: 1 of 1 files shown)


evilpie:
Shouldn't we still order the fallback order based on the current locale?

kjarosh:
> Shouldn't we still order the fallback order based on the current locale?

You mean so that e.g. if someone uses Korean, the Latin characters would be provided by the Korean font?

evilpie:
In particular I was thinking of Chinese users getting Japanese character variants were they overlap.

kjarosh:
> In particular I was thinking of Chinese users getting Japanese character variants were they overlap.

That's a fair point, fixed!

desktop/src/gui/controller.rs
```diff
@@ -511,46 +509,145 @@ fn try_wgpu_backend(backend: wgpu::Backends) -> Option<wgpu::Instance> {
...
-        "ko" => Some(Family::Name("Malgun Gothic")),
-        _ => None,
-    } {
-        families.push(windows_font);
+    locale: unic_langid::LanguageIdentifier,
+) -> egui::FontDefinitions {
+    let mut fd: FontDefinitions = egui::FontDefinitions::default();
+
+    let lang = locale.language.as_str();
+    let is_ja = lang == "ja";
```
evilpie, line 520: Would probably be simpler with `langid!`

desktop/src/gui/controller.rs
```diff
@@ -511,46 +509,145 @@ fn try_wgpu_backend(backend: wgpu::Backends) -> Option<wgpu::Instance> {
...
+        Err(e) => {
+            tracing::warn!("Failed to register {query:?} as {family}: {e}");
+            return;
+        }
+    };
+
+    tracing::info!("Registering font {name} as {family}");
+
+    fd.font_data.insert(name.clone(), fontdata.into());
+    if !fd.families.contains_key(family) {
```
evilpie, line 636: Maybe something like `fd.families.entry(family).or_insert_with(|| vec::new()).push(name)`

desktop/src/gui/controller.rs
kjarosh: What do you mean? `langid!` would check all locale properties, we're interested in language only

desktop/src/gui/controller.rs
kjarosh: Fixed, thanks!

desktop/src/gui/controller.rs
evilpie: Ah you are right, I had somehow thought that `locale == langid!("ko")` would just match the language.

desktop/src/gui/controller.rs
```diff
@@ -511,46 +509,139 @@ fn try_wgpu_backend(backend: wgpu::Backends) -> Option<wgpu::Instance> {
...
+type PrioritizedQueries<'a> = Vec<(usize, FamilyQuery<'a>)>;
+
+fn register_family(
+    font_database: &Database,
+    fd: &mut FontDefinitions,
+    family: egui::FontFamily,
+    mut queries: PrioritizedQueries<'_>,
+) {
+    queries.sort_by_key(|(priority, _)| *priority);
+    for (_, query) in queries {
```
evilpie, line 614: Maybe I a missing something, but shouldn't we start with the highest priority?
https://docs.rs/fontdb/latest/fontdb/struct.Query.html#structfield.families
This references CSS font-family, where the first has the highest priority.

evilpie: Oh, I am being stupid. We already higher priority == lower number.

kjarosh: We do. I'm not sure what you're referring to, but note we have two kinds of priorities. `PrioritizedQueries` describe queries by priority (so priority 1 is the main font, 2 is the pan-CJK font, etc.), and each (CSS-like) query has a list of font families, where the first font that matches is returned.

evilpie (approved):

PR #20265, opened 2025-04-26
desktop: Add workaround for RTL scripts in GUI

### LINKED ISSUE #11538: GUI chosen font for menus doesn't support Arabic
### Describe the bug

GUI chosen font for menus doesn't support Arabic

Screenshot_٢٠٢٣٠٦١٤_٢٠٣١٠٦

### Expected behavior

Shown correctly

### Affected platform

Desktop app

### Operating system

Fedora 38

### Browser

_No response_

### Additional information

_No response_

### LINKED ISSUE #16653: Problem in the hebrew language
### Describe the bug

The letters in the hebrew language are flipped. instead of "שלום" it writes as "םולש".
צילום מסך 2024-06-09 172625

### Expected behavior

it should be קובץ, פקדים, סימניות ext.

### Content Location

not related to a spesific flash content

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

_No response_

### Additional information

_No response_

kjarosh:
This patch implements a workaround for egui not supporting RTL scripts. RTL text runs are being reversed and some characters are being mirrored.

It does not fix the interface being LTR though, and I assume ligatures also won't be drawn properly. But at least RTL text is readable now. Also text selection will not work properly (text will be reversed), and user-inputted text will not be fixed.

* Fixes https://github.com/ruffle-rs/ruffle/issues/16653
* Fixes https://github.com/ruffle-rs/ruffle/issues/11538

| Before | After |
|:------:|:-----:|
|  |  |
|  |  |
|  |  |

### DIFF
--- a/desktop/src/gui/locale.rs
+++ b/desktop/src/gui/locale.rs
@@ -1,0 +1,65 @@
+use fluent_templates::fluent_bundle::FluentValue;
+use fluent_templates::{static_loader, Loader};
+use std::borrow::Cow;
+use std::collections::HashMap;
+use unic_langid::LanguageIdentifier;
+
+static_loader! {
+    static TEXTS = {
+        locales: "./assets/texts",
+        fallback_language: "en-US"
+    };
+}
+
+pub fn text<'a>(locale: &LanguageIdentifier, id: &'a str) -> Cow<'a, str> {
+    TEXTS
+        .try_lookup(locale, id)
+        .map(Cow::Owned)
+        .unwrap_or_else(|| {
+            tracing::error!("Unknown desktop text id '{id}'");
+            Cow::Borrowed(id)
+        })
+}
+
+pub fn optional_text(locale: &LanguageIdentifier, id: &str) -> Option<String> {
+    TEXTS
+        .lookup_single_language::<&str>(locale, id, None)
+        .inspect_err(|e| tracing::trace!("Error looking up text: {e}"))
+        .ok()
+}
+
+pub fn available_languages() -> Vec<&'static LanguageIdentifier> {
+    let mut result: Vec<_> = TEXTS.locales().collect();
+    result.sort();
+    result
+}
+
+#[allow(dead_code)]
+pub fn text_with_args(
+    locale: &LanguageIdentifier,
+    id: &'static str,
+    args: &HashMap<Cow<'static, str>, FluentValue>,
+) -> Cow<'static, str> {
+    TEXTS
+        .try_lookup_with_args(locale, id, args)
+        .map(Cow::Owned)
+        .unwrap_or_else(|| {
+            tracing::error!("Unknown desktop text id '{id}'");
+            Cow::Borrowed(id)
+        })
+}
+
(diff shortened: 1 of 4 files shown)


torokati44:
Given that RTL support in egui is closer than ever with https://github.com/emilk/egui/pull/5784, I'm not sure this hack is worth it still - but of course I can't tell the future of that PR either...

kjarosh:
I'm not sure if it's a _hack_, as something like that is done normally but on a different level.

But IMO this workaround is so simple and so loosely coupled to the existing code that we might as well do it instead of waiting for several months for RTL support in egui. Removing it will be very easy in the future.

danielhjacobs:
https://github.com/emilk/egui/blob/ba67ccc81eabe156dc0519e54571592f64d0bc20/parley-todo.md is also still somewhat of a long list.

desktop/src/gui/locale.rs
```diff
@@ -0,0 +1,131 @@
...
+/// Check is the given locale has a RTL script.
+///
+/// That's a very simplistic way of detecting RTL,
+/// but it's enough for our use.
+fn is_rtl(langid: &LanguageIdentifier) -> bool {
+    matches!(
+        langid.language.as_str(),
+        "ar" | "he" | "fa" | "ur" | "ps" | "syr" | "dv"
+    )
+}
```
torokati44, line 98: Why not https://docs.rs/unic-langid/0.9.5/unic_langid/struct.LanguageIdentifier.html#method.character_direction ?

desktop/src/gui/locale.rs
```diff
@@ -0,0 +1,131 @@
...
+        ')' => '(',
+        '[' => ']',
+        ']' => '[',
+        '{' => '}',
+        '}' => '{',
+        '<' => '>',
+        '>' => '<',
+        _ => c,
+    }
+}
```
torokati44, line 116: FWIW, there's also https://docs.rs/unicode-bidi-mirroring/0.4.0/unicode_bidi_mirroring/fn.get_mirrored.html.

desktop/src/gui/locale.rs
kjarosh: ... because somehow I missed that, thanks!

desktop/src/gui/locale.rs
kjarosh: I'm not sure if it's worth adding another dependency, as the set of strings we mirror is finite (we do that only for translated strings), but I'll add it if you think it's better \:D

torokati44: I agree, either way is fine for me.

hamzaelbarmaki1994:
there is a library for RTL https://github.com/harfbuzz/harfbuzz ,  very popular

torokati44 (approved):

jermanuts:
@kjarosh looking at the Arabic screenshot, the Arabic letters are not connected see the https://isthisarabic.com/ 2nd and 3rd example. So it is not readable and is indeed a hack.

Thanks.

PR #20266, opened 2025-04-26
web: Two small improvements

evilpie:
Review by commit.
- We can re-use exclude_matches from the manifest
- Prevent DOM clobbering problems with document.currentScript


danielhjacobs (approved):
Nice, somehow I forgot there was a method to directory get the manifest content from extension scripts.

PR #20289, opened 2025-04-28
core: Replace GcCell with Gc in Video

kjarosh:
This refactor replaces GcCell with Gc and uses interior mutability instead.

### DIFF
--- a/core/src/display_object/video.rs
+++ b/core/src/display_object/video.rs
@@ -12,7 +12,8 @@ use crate::streams::NetStream;
 use crate::tag_utils::{SwfMovie, SwfSlice};
 use crate::vminterface::{AvmObject, Instantiator};
 use core::fmt;
-use gc_arena::{Collect, GcCell, Mutation};
+use gc_arena::barrier::unlock;
+use gc_arena::{Collect, Gc, Lock, Mutation, RefLock};
 use ruffle_render::bitmap::{BitmapInfo, PixelSnapping};
 use ruffle_render::commands::CommandHandler;
 use ruffle_render::quality::StageQuality;
@@ -32,12 +33,12 @@ use swf::{DefineVideoStream, VideoCodec, VideoFrame};
 /// framerate.
 #[derive(Clone, Collect, Copy)]
 #[collect(no_drop)]
-pub struct Video<'gc>(GcCell<'gc, VideoData<'gc>>);
+pub struct Video<'gc>(Gc<'gc, VideoData<'gc>>);
 
 impl fmt::Debug for Video<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("Video")
-            .field("ptr", &self.0.as_ptr())
+            .field("ptr", &Gc::as_ptr(self.0))
             .finish()
     }
 }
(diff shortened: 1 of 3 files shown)


core/src/display_object/video.rs
```diff
@@ -33,56 +33,54 @@ use swf::{DefineVideoStream, VideoCodec, VideoFrame};
...
 #[collect(no_drop)]
 pub struct VideoData<'gc> {
-    base: DisplayObjectBase<'gc>,
+    base: RefLock<DisplayObjectBase<'gc>>,
 
-    avm1_text_field_bindings: Vec<Avm1TextFieldBinding<'gc>>,
+    avm1_text_field_bindings: RefLock<Vec<Avm1TextFieldBinding<'gc>>>,
 
     /// The source of the video data (e.g. an external file, a SWF bitstream)
-    source: GcCell<'gc, VideoSource<'gc>>,
```
moulins, line 54: Have you considered removing the `Gc` here? It would avoid an allocation for non-SWF videos:
```rs
struct VideoData<'gc> {
  // ...
  source: Lock<VideoSource<'gc>>,
}

enum VideoSource<'gc> {
  Swf(Gc<'gc, SwfVideoSource>),
  // ...
}

struct SwfVideoSource {
  // ...
}
```

core/src/display_object/video.rs
```diff
@@ -33,56 +33,54 @@ use swf::{DefineVideoStream, VideoCodec, VideoFrame};
...
     /// The self bounds for this movie.
-    size: (i32, i32),
+    size: Cell<(i32, i32)>,
 
     /// The last decoded frame in the video stream.
     ///
     /// NOTE: This is only used for SWF-source video streams.
-    #[collect(require_static)]
-    decoded_frame: Option<(u32, BitmapInfo)>,
+    decoded_frame: RefCell<Option<(u32, BitmapInfo)>>,
```
moulins, line 79: This could save a `RefCell` by using a nested struct:
```rs
struct VideoData<'gc> {
  // ...
  cell: RefCell<VideoDataMut>,
}

struct VideoDataMut {
  keyframes: BTreeSet<u32>,
  decoded_frame: Option<(u32, BitmapInfo)>,
}
```

core/src/display_object/video.rs
kjarosh: Removed `Gc` entirely. Originally I thought we need to leave the reference intact but that doesn't seem to be the case.

core/src/display_object/video.rs
kjarosh: I'm not sure I follow. The only reason would be to have one `RefCell` instead of 2? I don't think it has any noticeable perf impact

moulins: Yup, this is kind of a micro-optimization, but it would save a `RefCell` (and so an `usize` in the struct)

core/src/display_object/video.rs
moulins: That seems contradicted by the doc comment of `VideoSource::Swf`:
>  NOTE: Fields within this enum will be shared across all instances of a
> particular character. If you need to mutate the video source, consider
> reallocating a new source for your specific video instead.

In particular, I think that instantiating a partially-preloaded `Video` character shouldn't make a separate copy of the frame list.

kjarosh: I agree, that comments contradicts the code. What's more interesting, it was added after adding `frames`

core/src/display_object/video.rs
kjarosh: That looks like a premature optimization: the memory consumed by the structs >> `usize`, and it's not a free optimization, as we lose the ability to borrow multiple fields independently.

core/src/display_object/video.rs
moulins: Well, the comment *is* correct with the current setup (`GcCell<'gc, VideoSource<'gc>>`); I'm pretty sure your change breaks the `VideoFrame` tag (because the preloading only occurs on the "master" Video character, not on any videos present on stage)

kjarosh: Hmmm... are you sure about that? I'm looking at the code and I cannot see how my patch changes any behavior here.

Looking at the lifetime of the `GcCell<VideoSource>` pointer, it's allocated when creating `Video` and in `attach_netstream`. However, the pointer never escapes `Video`, so that's why I'm assuming I can just own the `VideoSource` (and thus replace a `Gc` with `RefLock`).

However, it's possible I'm not seeing something here :thinking:

moulins: The video source can escape the current instance through`TDisplayObject::instantiate`, which creates a new `Video` object by cloning `VideoData` (and thus the `GcCell`).

moulins (approved):

PR #20291, opened 2025-04-28
core: Replace GcCell with Gc in NetStream

kjarosh:
This refactor replaces GcCell with Gc and uses interior mutability instead.

### DIFF
--- a/core/src/avm1/globals/netstream.rs
+++ b/core/src/avm1/globals/netstream.rs
@@ -150,7 +150,7 @@ fn set_buffer_time<'gc>(
             .unwrap_or(Value::Undefined)
             .coerce_to_f64(activation)?;
 
-        ns.set_buffer_time(activation.gc(), buffer_time);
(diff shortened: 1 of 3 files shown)


kjarosh:
> As with the `Video` PR, all the `RefCell`s could be merged by introducing a nested struct.

Similarly to the `Video` PR I don't think it's a good idea, I would argue that having the possibility to borrow different fields independently is better (based on various "patterns" I see currently in code).

Also, I think it would be better to create a nested struct based on semantics; in this case a struct of fields that are reset when resetting buffer (see `reset_buffer`) would be better in order to prevent code duplication and possible bugs caused by that. I'll happily do that in a future PR

moulins:
> Also, I think it would be better to create a nested struct based on semantics; in this case a struct of fields that are reset when resetting buffer (see reset_buffer) would be better in order to prevent code duplication and possible bugs caused by that.

That does make a lot of sense.

moulins (approved):

Issue #19692, opened 2025-03-03
Snail Bob 2: First level softlock


waspennator:
### Describe the bug

Game seems to softlock once you beat first level, can't click on next level or level map buttons, issue also seems to be present on flashpoint version of Snail Bob 1.

### Expected behavior

Buttons should work

### Content Location

https://www.newgrounds.com/portal/view/572601

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Edge 133

n0samu:
Probably related to #11467

PR #20293, opened 2025-04-28
core: Refresh frame script on registration

### LINKED ISSUE #11467: Snail Bob: softlock, can not click on next level button
### Describe the bug

After finishing the first level, it is not possible to click on anything

image

SWF - https://data.gameflare.com/games/1278/t6FrHyCvRxjDPM.swf
Game - https://www.gameflare.com/online-game/snail-bob/

Same problem in Snail Bob 2

### Expected behavior

start second level

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

_No response_

nivkner:
fixes #11467 and #19692

while running a `goto` on a frame would normally cause its frame-script to be counted as already executed,
a newly registered frame-script does not seem to count as such.

the assignment of `None` to `queued_script_frame` is, as far as i can tell, redundant given that `last_queued_script_frame` already prevents the same script from running twice, so it was removed to allow for simpler "refresh" logic.


nivkner:
to double-check the point about `None` (and for the fun of it),
I wrote a TLA+ specification to verify that the Option is in fact redundant [here](https://gist.github.com/nivkner/5029542f809f1dad559fe8072320ae35).

what i got was that it is redundant, so long as `run_frame_internal` is called at least once before `run_frame_scripts` on a movie clip containing a frame script.

given that `run_frame_scripts` is called only during the frame script phase, while `run_frame_scripts` is called during the enter phase, it seems likely.

nivkner:
to double-check the point about `None` (and for the fun of it),
I wrote a TLA+ specification to verify that the Option is in fact redundant [here](https://gist.github.com/nivkner/5029542f809f1dad559fe8072320ae35).

what i got was that it is redundant as of 64d6bbf6bc86, so long as `run_frame_internal` is called at least once before `run_frame_scripts` on a movie clip containing a frame script.

given that `run_frame_scripts` is called only during the frame script phase, while `run_frame_scripts` is called during the enter phase, it seems likely.

nivkner:
decided to remove changes to `queued_script_frame`, to hopefully make it easier to review.

Molisson (approved):
Reviewed carefully (test files ignored) and seemed simple for me.

Dinnerbone (approved):
LGTM! Thank you!

PR #2633, opened 2021-01-18
Allow configuring log level on web

adrian17:
New default in API is "error".
Demo page uses "warn".

I'd also love to set "warn" in extension, but I couldn't actually found the config in extension's source. Any hints?

PR #20297, opened 2025-04-28
web/extension: Also add "debug" and "trace" log levels as options

torokati44:
They would have proven useful for me at least twice now during local testing.
Also, I always see them in this order, not as they currently are.


danielhjacobs:
Do we know why, after Adrian added these log levels in #2633, relrelb never added them as options in #5364?

danielhjacobs:
Were those log levels useless without avm_debug before #10310? Is that why?

kjarosh (approved):

PR #20303, opened 2025-04-29
debug_ui/display_object: Remove collapsing header for frame list

crumblingstatue:
Instead, we embed it directly in a compact way.

My guess is that the original reason it was put into a CollapsingHeader is that the original frame list tanked the performance when shown, but this has been fixed by using `TableBody::rows` to only render the visible rows.

This allows quick access to the frame list, which is helpful for quickly seeking to the parts of a movie we're interested in.

### DIFF
--- a/core/src/debug_ui/display_object.rs
+++ b/core/src/debug_ui/display_object.rs
@@ -800,126 +800,126 @@ impl DisplayObjectWindow {
                 });
                 ui.end_row();
             });
-
-        CollapsingHeader::new("Frame List")
-            .id_salt(ui.id().with("frames"))
-            .show(ui, |ui| {
+        ui.separator();
+        ui.horizontal(|ui| {
+            ui.label("Frame list");
+            ui.with_layout(egui::Layout::right_to_left(egui::Align::Center), |ui| {
                 ui.checkbox(&mut self.track_current_frame, "Track current");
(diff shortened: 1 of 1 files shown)


crumblingstatue:
Here's a preview screenshot
image

kjarosh:
I'm not sure removing the collapsing header is a good move. I get that it improves UX when operating directly on the frame list, but I don't think that's a main use case overall.

The main idea behind collapsing headers is that each page contains basic table-like overview and additional stuff is in collapsing headers. If you look at `Position` or `EditText`, you can see that there might be multiple collapsing sections.

Now, if we remove the collapsing header in `MovieClip`, we don't have a way of adding new MovieClip-related stuff here (e.g. maybe some timeline stuff?).

I think the collapsing header is a good compromise between ease of use and potential for future expansion. There's a lot of things going on in movie clips and a lot of ways to improve debugging them.

crumblingstatue:
If desired, I can slightly alter this pull request to use a `SelectableLabel` for the frame list label, which would allow adding more `SelectableLabel`s next to it (more tabs).

Example of how it would look
image

crumblingstatue:
https://github.com/user-attachments/assets/aed4854e-7d90-41ed-afbb-0c7d062ec0ba

Just a quick shameless advertisement why it's so nice to have a quickly accessible frame list for quick bug reproduction (#20326)

Issue #20191, opened 2025-04-15
Game record saving abnormality


Hiyase:
### Describe the bug

Here is a statement in advance: my question comes from the Linux desktop
When a game is saved normally, its path is like this (but it is too long):
```
/home/Hiyase/.local/share/ruffle/SharedObjects/localhost/home/Hiyase/Games/Flash/CK/CK3%20v1.2.swf/CK3save.sol
```      
This looks like an illegal path.
Warning about storage failure in a certain game:
```log
2025-04-15T04:48:30.254398Z  WARN ruffle_frontend_utils::backends::storage: Unable to read file "localhost//save_data_kakutou": Os { code: 2, kind: NotFound, message: "No such file or directory" }
```
Storing successful cases in browser extensions:
```
bf2c3ae7-400f-406d-87de-5574a38a3835//save_data_kakutou:"****************" (I hid the data)
```
It looks like the problem might be coming from [`Here`](https://github.com/ruffle-rs/ruffle/blob/master/frontend-utils/src/backends/storage.rs#L31)

I'm not sure, but I think I can provide ~/.cache/ruffle/log/ruffle.log
```log
2025-04-15T05:24:51.183544Z  INFO ruffle_desktop::gui::controller: Using preferred backend Vulkan
2025-04-15T05:24:51.209975Z  INFO ruffle_desktop::gui::controller: Using graphics API vulkan on NVIDIA GeForce GTX 970M (type: DiscreteGpu)
2025-04-15T05:24:51.281322Z  WARN egui_wgpu::renderer: Detected a linear (sRGBA aware) framebuffer Rgba8UnormSrgb. egui prefers Rgba8Unorm or Bgra8Unorm    
2025-04-15T05:24:51.285418Z  INFO ruffle_desktop::gui::controller: loaded cjk fallback font "NotoSansCJKsc-Regular"
2025-04-15T05:24:51.311918Z  WARN ruffle_video_external::decoder::openh264: Failed to load system OpenH264 library libopenh264.so.7: Error while loading OpenH264: libopenh264.so.7: cannot open shared object file: No such file or directory
2025-04-15T05:24:51.312001Z  WARN ruffle_video_external::decoder::openh264: Failed to load system OpenH264 library libopenh264.so.2.4.1: Error while loading OpenH264: libopenh264.so.2.4.1: cannot open shared object file: No such file or directory
[...]

Issue #19117, opened 2025-01-02
Interactive Buddy loaded from URL not initializing game state properly


parkerlreed:
### Describe the bug

Loading the game multiple times sometimes results in a broken save file. This happens even locally on desktop.

In game File > Clear File fixes this temporarily.

Possibly related to https://github.com/ruffle-rs/ruffle/issues/798

Working vs broken logs

[working.txt](https://github.com/user-attachments/files/18287868/working.txt)
[broken.txt](https://github.com/user-attachments/files/18287867/broken.txt)

Command for each. I kept the arguments consistent as to show it's only the load via URL that breaks it.

working
```
ruffle --proxy="http://localhost:22500/" --power=high --graphics=default /home/parker/Games/Flashpoint/Data/Games/interactive_buddy_v_1_02_by_shock_value_d6ma8m.swf
```

broken
```
ruffle --proxy="http://localhost:22500/" --power=high --graphics=default http://shock-value.deviantart.com/interactive_buddy_v_1_02_by_shock_value_d6ma8m.swf
```

### Expected behavior

Able to interact with the game

### Content Location

swf zipped [game.zip](https://github.com/user-attachments/files/18287874/game.zip)

### Affected platform

Desktop app every single launch.

Very occasionally on hosted version.

### Operating system

Arch Linux

### Additional information

Ruffle 0.1.0-nightly (1d0576766c543d3267ed6e489035e8acba677232 2025-01-01)

n0samu:
It doesn't actually seem to matter how the game is loaded. It seems that sometimes the save file breaks, causing the game to be unresponsive on next load, as in #798. It happens quite consistently for me on desktop, regardless of whether the game is loaded locally or via a URL. I also observed the problem once on web, but was not able to get it to occur again.

kjarosh:
Fixed by https://github.com/ruffle-rs/ruffle/pull/20304

Issue #19313, opened 2025-01-23
Saves not functioning in Eleanor Loving Wife (Lesson of Passion game)


GooseWhoIsLoose:
### Describe the bug

Dowloaded Lesson of Passion's "Elanor Loving Wife". When launched directly through an official Adobe flash player, it saves after every in-game day. These saves persist when the player is closed and re-opened.

When launched through Ruffle, these saves do not persist and all progress is lost if the players is closed and re-opened.

Saves _do_ persist in other games (tested Hey Wizard and Raze 2).

### Expected behavior

Expected saves to persist when the player is closed and re-opened.

### Content Location

https://drive.google.com/file/d/1SEogiZaO5krii3I-3seNyCxtFSKa2E17/view?usp=sharing

### Affected platform

Desktop app

### Operating system

Windows 10

kjarosh:
Fixed by https://github.com/ruffle-rs/ruffle/pull/20304

PR #1326, opened 2020-10-15
WGPU renderer improvements

Dinnerbone:
These should speed up rendering with bitmaps, and hopefully no longer run into issues when we use too many.
It also significantly speeds up exporting and makes exporting with dx12 viable, as dx12 takes a while to create the pipelines.

PR #20304, opened 2025-04-29
core: Flush shared objects on Player drop

kjarosh:
This patch makes sure shared objects are flushed when Player is dropped. Specifically, it fixes shared objects not being saved on desktop properly.

* Fixes https://github.com/ruffle-rs/ruffle/issues/20191

### DIFF
--- a/core/src/player.rs
+++ b/core/src/player.rs
@@ -2066,10 +2066,6 @@ impl Player {
         std::mem::swap(&mut self.storage, storage);
     }
 
-    pub fn destroy(self) -> Box<dyn RenderBackend> {
-        self.renderer
-    }
-
     pub fn ui(&self) -> &dyn UiBackend {
         &*self.ui
     }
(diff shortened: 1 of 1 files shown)


core/src/player.rs
```diff
@@ -2068,10 +2068,6 @@ impl Player {
         &mut self.storage
     }
 
-    pub fn destroy(self) -> Renderer {
```
torokati44, line 2071: What was the point of this method? 🤔 Why does it rescue the renderer, of all things? I assume it wasn't used anymore...

kjarosh: See https://github.com/ruffle-rs/ruffle/pull/1326. It was used in exporter for taking a screenshot.

torokati44: Ah, I see, thanks!

torokati44: Ah, and it got disused by https://github.com/ruffle-rs/ruffle/pull/7594.

torokati44 (approved):
Sure, I guess! :man_shrugging:

n0samu:
This may fix #19117 and/or #19313, would appreciate if someone tests them!

PR #20307, opened 2025-04-29
exporter: Phase out `--skip-unsupported` option

kjarosh:
Ruffle supports AVM2 movies pretty well now. Phasing out the `--skip-unsupported` option, but leaving it as a hidden option for backward compatibility.


kjarosh:
Of course, it's my assumption that this option is not needed anymore, discussion is welcome

PR #20310, opened 2025-04-30
core: Remove `GcCell` usage in `MovieClipShared`

moulins:
Additionally, preloading now modifies the existing `MovieClipShared` instance instead of swapping a clone on each `preload` call.

This means that duplicating a partially-preloaded `MovieClip` will return a clip sharing the preload state. This seems "more correct" to me (as it means that both clips will properly finish preloading), but I haven't tested this in FP.

EDIT: Added a further refactoring that moves most of the tag preloading methods to `MovieClipShared`, as they mostly don't need access to the actual MC instance.

### DIFF
--- a/core/src/display_object/movie_clip.rs
+++ b/core/src/display_object/movie_clip.rs
@@ -30,6 +30,7 @@ use crate::display_object::{
 use crate::drawing::Drawing;
 use crate::events::{ButtonKeyCode, ClipEvent, ClipEventResult};
 use crate::font::{Font, FontType};
+use crate::library::MovieLibrary;
 use crate::limits::ExecutionLimit;
 use crate::loader::{self, ContentType};
 use crate::loader::{LoadManager, Loader};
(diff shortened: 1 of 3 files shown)


core/src/display_object/movie_clip.rs
```diff
@@ -196,136 +195,77 @@ pub struct MovieClipData<'gc> {
...
-                avm1_text_field_bindings: Vec::new(),
-            },
-        ))
-    }
-
-    pub fn downgrade(self) -> MovieClipWeak<'gc> {
-        MovieClipWeak(GcCell::downgrade(self.0))
+        let shared = MovieClipShared::with_data(id, swf, num_frames, None);
+        let mut data = MovieClipData::new(shared, mc);
+        *data.flags.get_mut() = MovieClipFlags::PLAYING;
```
Lord-McSweeney, line 267: Is there a reason this doesn't use `data.flags.set(MovieClipFlags::PLAYING)`?

core/src/display_object/movie_clip.rs
```diff
@@ -388,76 +289,34 @@ impl<'gc> MovieClip<'gc> {
...
-            },
-        ));
+        let shared = MovieClipShared::with_data(
+            0,
+            movie.clone().into(),
+            movie.num_frames(),
+            loader_info.map(|l| l.0),
+        );
+        let mut data = MovieClipData::new(shared, activation.gc());
+        *data.flags.get_mut() = MovieClipFlags::PLAYING;
```
Lord-McSweeney, line 313: Here too

core/src/display_object/movie_clip.rs
moulins: No real reason, I can change it if you prefer.

PR #20320, opened 2025-04-30
avm2: Stub flash.system.Capabilities

kjarosh:
Add missing properties and methods along with version constraints.

### DIFF
--- a/core/src/avm2/globals/flash/system/Capabilities.as
+++ b/core/src/avm2/globals/flash/system/Capabilities.as
@@ -1,20 +1,158 @@
 package flash.system {
     import __ruffle__.stub_getter;
+    import __ruffle__.stub_method;
+
     public final class Capabilities {
-        public native static function get os(): String;
-        public native static function get playerType(): String;
-        public native static function get version(): String;
-        public native static function get screenResolutionX():Number;
-        public native static function get screenResolutionY():Number;
-        public native static function get pixelAspectRatio():Number;
-        public native static function get screenDPI():Number;
+        public static function get avHardwareDisable():Boolean {
+            stub_getter("flash.system.Capabilities", "avHardwareDisable");
+            return false;
+        }
+
+        [API("665")]
+        public static function get cpuArchitecture():String {
+            stub_getter("flash.system.Capabilities", "cpuArchitecture");
+            return "x86";
+        }
+
+        public static function get hasAccessibility():Boolean {
+            stub_getter("flash.system.Capabilities", "hasAccessibility");
+            return false;
+        }
+
+        public static function get hasAudio():Boolean {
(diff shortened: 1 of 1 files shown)


danielhjacobs:
Potentially useful in the future: #10469. Not quite sure why it was never merged, I think it may have had to do with concerns about fingerprinting. That's AVM1.

kjarosh:
> Potentially useful in the future: #10469. Not quite sure why it was never merged, I think it may have had to do with concerns about fingerprinting. That's AVM1.

I agree about fingerprinting. IMO we should either return a system-independent value or let the user choose it. But there's also the question of what those values are used for in SWFs.

core/src/avm2/globals/flash/system/Capabilities.as
```diff
@@ -1,20 +1,158 @@
...
+        }
+
+        public static function get hasMP3():Boolean {
+            stub_getter("flash.system.Capabilities", "hasMP3");
+            return true;
+        }
+
+        public static function get hasPrinting():Boolean {
+            stub_getter("flash.system.Capabilities", "hasPrinting");
+            return true;
```
danielhjacobs, line 49: Probably makes sense to be true, but since we don't yet support the PrintJob class (not even stubbing the isSupported flag to false) I imagine content that checks this will just fail later as I assume this would be checked for the purpose of safe-guarding PrintJob code.

kjarosh: That makes sense, fixed!

core/src/avm2/globals/flash/system/Capabilities.as
```diff
@@ -1,20 +1,158 @@
...
+
+        public native static function get screenDPI():Number;
+
+        public native static function get screenResolutionY():Number;
+
+        public native static function get screenResolutionX():Number;
+
+        public static function get serverString():String {
+            stub_getter("flash.system.Capabilities", "serverString");
+            return "A=t&SA=t&SV=t&EV=t&MP3=t&AE=t&VE=t&ACC=f&PR=t&SP=t&SB=f&DEB=t&V=WIN%208%2C5%2C0%2C208&M=Adobe%20Windows&R=1600x1200&DP=72&COL=color&AR=1.0&OS=Windows%20XP&L=en&PT=External&AVD=f&LFD=f&WD=f";
```
danielhjacobs, line 129: This says PR=t but now the default `hasPrinting` is `false`

danielhjacobs (approved):

core/src/avm2/globals/flash/system/Capabilities.as
kjarosh: Fixed, thanks!

adrian17:
Hey, I just noticed, I think you didn't commit removing the native implementations 😅

PR #20322, opened 2025-04-30
core: Add `NetStreamSource` for better buffer management

kjarosh:
Adding `NetStreamSource` makes it easier to reset buffer-related data. Instead of individually resetting fields, all appropriate fields are inside `NetStreamSource`, which can be replaced wholly.

### DIFF
--- a/core/src/streams.rs
+++ b/core/src/streams.rs
@@ -183,7 +183,7 @@ pub enum NetStreamType {
 
 #[derive(Clone, Debug, Collect)]
 #[collect(no_drop)]
-pub struct NetStreamData<'gc> {
+pub struct NetStreamSource {
     /// All data currently loaded in the stream.
     buffer: RefCell<Buffer>,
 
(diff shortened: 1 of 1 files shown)


moulins:
Because most of the methods that access `NetStreamSource` already have to borrow one of the `RefCell` fields, I'd actually suggest removing all interior mutability from it and changing `NetStreamData` to hold a `Gc<'gc, RefCell<NetStreamSource>>`.

kjarosh:
> most of the methods that access `NetStreamSource` already have to borrow one of the `RefCell` fields, so we don't gain much by having "fine-grained" interior mutability here.

Changed to `RefCell<NetStreamSource>` thinking that we don't need fine-grained interior mutability, but it turns out that the code _does_ require fine-grained mutability, because it's quite complex and we would have to borrow the source multiple times. In my opinion it's better to provide fine-grained mutability (and have loose coupling between methods) than to pass `RefMut` everywhere.

moulins (approved):
Sorry for the delay.

> but it turns out that the code does require fine-grained mutability, because it's quite complex and we would have to borrow the source multiple times.

After looking more closely, you're right; I think it could be done by moving some code to methods on `NetStreamSource`, but let's keep this for a potential future PR.

PR #20328, opened 2025-05-02
avm2: Speed up some native calls

Lord-McSweeney:
This allows us to optimize some calls to native functions when
1. the function has a `[Ruffle(FastCall)]` annotation,
2. the arguments are of the correct count and type, and
3. the class of the value the function is being called on is `final`.

For example, `Math.abs` runs more than twice as fast with this PR as it does on current `master`. From my testing, in most SWFs, 5-20% of calls that would be `CallMethod` are now `CallNative`.

Box2d stats:
```
ConstructSuper -> Pop: 70%

FindPropStrict -> GetScriptGlobals: 82.67%
FindPropStrict -> GetOuterScope: 8.96%
FindPropStrict -> GetScopeObject: 8.36%

InitProperty -> SetSlot: 6.28%
InitProperty -> SetSlotNoCoerce: 90.98%
InitProperty -> CallMethod: 2.73%

SetProperty -> SetSlot: 8.07%
SetProperty -> SetSlotNoCoerce: 73.88%
SetProperty -> CallMethod: 2.4%

GetProperty -> GetSlot: 88.92%

CallProperty -> CallMethod: 74.1% (-9.16%)
CallProperty -> CallNative: 9.16% (+9.16%)

CallPropVoid -> CallMethod: 91.72% (-2.25%)
CallPropVoid -> CallNative: 2.25% (+2.25%)

Coerce -> Nop: 73.18%

CoerceD -> Nop: 98.75%

CoerceB -> Nop: 97.29%

CoerceU -> Nop: 64.91%

CoerceI -> Nop: 52%

ReturnValue -> ReturnValueNoCoerce: 96.39%
```

TODO:
- [x] Record box2d optimization stats
- [x] Add documentation on FastCall to the globals README
- [ ] ~~Use a HashSet for the FastCall methods...?~~

### DIFF
--- a/core/build_playerglobal/src/lib.rs
+++ b/core/build_playerglobal/src/lib.rs
@@ -44,6 +44,9 @@ const METADATA_ABSTRACT: &str = "Abstract";
 const METADATA_NATIVE_ACCESSIBLE: &str = "NativeAccessible";
 /// Like `METADATA_NATIVE_ACCESSIBLE`, but for methods instead of slots.
 const METADATA_NATIVE_CALLABLE: &str = "NativeCallable";
+/// Indicates that this method does not read any properties of the Activation
+/// passed to it except UpdateContext fields. This is used as an optimization.
+const METADATA_FAST_CALL: &str = "FastCall";
 // The name for metadata for namespace versioning- the Flex SDK doesn't
 // strip versioning metadata, so we have to allow this metadata name
 const API_METADATA_NAME: &str = "API";
@@ -442,6 +445,7 @@ fn write_native_table(data: &[u8], out_dir: &Path) -> Result<Vec<u8>, Box<dyn st
     let mut rust_instance_allocators = vec![none_tokens.clone(); abc.classes.len()];
     let mut rust_call_handlers = vec![none_tokens.clone(); abc.classes.len()];
     let mut rust_custom_constructors = vec![none_tokens; abc.classes.len()];
+    let mut rust_fast_calls = vec![];
 
     let mut rust_accessible_slots: HashMap<String, Vec<_>> = HashMap::new();
     let mut rust_accessible_methods: HashMap<String, Vec<_>> = HashMap::new();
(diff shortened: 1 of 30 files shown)


moulins:
> Use a HashSet for the FastCall methods...?

An alternative (maybe simpler to implement?) could be sorting the static array, and use `binary_search` for O(log n) lookups.

core/build_playerglobal/src/lib.rs
```diff
@@ -726,6 +733,15 @@ fn write_native_table(data: &[u8], out_dir: &Path) -> Result<Vec<u8>, Box<dyn st
             #(#rust_custom_constructors,)*
         ];
 
+        // This is an array containing the method ids of every method marked
+        // "[Ruffle(FastCall)]". Unlike the rest, it is not indexed by method id-
+        // instead, every item in the list is a method id.
+        //
+        // FIXME: should this be some sort of hashset?
```
adrian17, line 740: If you sorted the vec before putting it in, we could use `binary_search` instead of `iter().any()` to speed up* lookups without affecting memory use at all.
Alternatively, there's always https://github.com/rust-phf/rust-phf to construct a static hashset at compile time (and thus not waste any allocations), but might be an overkill here.

(*I imagine the difference might not matter as long as we have <100 fast methods. )

core/src/avm2/optimizer/optimize.rs
```diff
@@ -1944,3 +1926,171 @@ fn abstract_interpret_ops<'gc>(
...
+
+                let mut result_op = Op::CallMethod {
+                    num_args: 0,
+                    index: disp_id,
+                    push_return_value: true,
+                };
+
+                // We can further optimize calling FastCall getters into a
+                // static native method call
+                if let MethodKind::Native {
```
adrian17: Any chance this pattern
```rs
if let MethodKind::Native {
    native_method,
    fast_call: true,
} = method.method_kind()
{
    if stack_value.class.is_some_and(|c| c.is_final()) {
        if /* check len(arguments)==N and types match */
        {
            result_op = Op::CallNative {
                method: *native_method,
                num_args: 1,
                push_return_value: false,
            };
        }
    }
```
Can be refactored to a function or macro? I see 4 variations of it here, just differing by # of args.

PR #20341, opened 2025-05-03
ci: Release dual-wasm extensions as assets

danielhjacobs:
We're building them anyway so I suppose it doesn't hurt to release them, though I'm open to opinions. Based on a user on Discord using super old hardware with a modern browser whose hardware actually doesn't allow the with extensions module to work.

### DIFF
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -541,7 +541,8 @@ jobs:
         shell: bash -l {0}
         run: npm run build:dual-wasm-repro
 
-      - name: Upload generic extension
+      # NOTE: This is done here because the job that builds the extensions for publication doesn't have the vanilla (MVP) WASM modules, but for the selfhosted package we build them both
+      - name: Upload dual-wasm generic extension
         run: |
           tag_name="${{ needs.create-nightly-release.outputs.tag_name }}"
           package_file="${{ needs.create-nightly-release.outputs.package_prefix }}-web-extension-dual-wasm.zip"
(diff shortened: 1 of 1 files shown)


torokati44:
The PR title / commit message makes it seem like the dual-wasm extensions are currently published, and this change would make them no longer be so. I suggest something like:

`ci: Release dual-wasm extensions but as assets (without publishing them)`

With that part in parentheses being optional even.

danielhjacobs:
Title changed, not at a computer but if I recall correctly squash and merge uses the title.

danielhjacobs:
I had to modify the release asset names so they wouldn't match these patterns: https://github.com/ruffle-rs/ruffle-rs.github.io/blob/master/src/app/downloads/config.tsx#L53-L54

.github/workflows/release_nightly.yml
```diff
@@ -541,6 +541,26 @@ jobs:
         shell: bash -l {0}
         run: npm run build:dual-wasm-repro
 
+      - name: Upload generic extension
```
torokati44, line 544: nit: `Upload dual-wasm generic extension`?

.github/workflows/release_nightly.yml
```diff
@@ -541,6 +541,26 @@ jobs:
...
+        run: |
+          tag_name="${{ needs.create-nightly-release.outputs.tag_name }}"
+          package_file="${{ needs.create-nightly-release.outputs.package_prefix }}-web-extension-dual-wasm.zip"
+          cp "./web/packages/extension/dist/ruffle_extension.zip" "$package_file"
+          gh release upload "$tag_name" "$package_file"
+          rm "$package_file"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Upload Firefox extension (unsigned)
```
torokati44, line 554: nit: `Upload dual-wasm Firefox extension (unsigned)`?

torokati44 (approved):
On top of the step name nits, I think a comment saying something like this, somewhere in there, would be nice: "NOTE: The reason this is done in this job is because the other job that builds the extensions for publication doesn't have the vanilla (MVP) WASM modules, but for the selfhosted package we build them both...", otherwise, LGTM!

PR #20349, opened 2025-05-05
core: Add deriveable `HasPrefixField` trait

moulins:
And use it instead of the manual layout assertions in AVM2 object structs.

This will make it easier to use `Gc` prefix casting as a part of the `GcCell` refactors on the `DisplayObject` hierarchy.

### DIFF
--- a/core/macros/src/lib.rs
+++ b/core/macros/src/lib.rs
@@ -6,8 +6,8 @@ use proc_macro2::TokenStream as TokenStream2;
 use quote::{format_ident, quote};
 use syn::parse::{Parse, ParseStream};
 use syn::{
-    parse_macro_input, parse_quote, FnArg, ImplItem, ImplItemFn, ItemEnum, ItemTrait, LitStr, Meta,
-    Pat, TraitItem, Visibility,
+    parse_macro_input, parse_quote, DeriveInput, FnArg, ImplItem, ImplItemFn, ItemEnum, ItemTrait,
+    LitStr, Meta, Pat, TraitItem, Visibility,
 };
 
 /// Define an enum whose variants each implement a trait.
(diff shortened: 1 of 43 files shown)


core/macros/src/lib.rs
```diff
@@ -177,6 +177,63 @@ pub fn enum_trait_object(args: TokenStream, item: TokenStream) -> TokenStream {
...
+                crate::utils::HasPrefixField<#field_ty>
+                for #ty #ty_generics #where_clause {
+            const ASSERT: () = {
+                ::core::assert!(::core::mem::offset_of!(Self, #field_name) == 0);
+                // Check that the field exists and has the correct type.
+                let _ = |check: &(Self,)| -> *const #field_ty {
+                    let (Self { #field_name, .. },) = check;
+                    #field_name as *const _
+                };
+            };
```
kjarosh, line 231: Can't we pull this assert out of the trait `HasPrefixField` just like it was before introducing the trait?

core/src/utils.rs
```diff
@@ -0,0 +1,26 @@
+use gc_arena::Gc;
+
+pub use ruffle_macros::HasPrefixField;
+
+/// A trait indicating that `Self` has `Inner` as an initial prefix.
```
kjarosh, line 5: nit: I'd elaborate on what that means, e.g. that it has to be a field that has the same address as the struct, is placed at the beginning in the memory, etc.

core/src/utils.rs
```diff
@@ -0,0 +1,26 @@
...
+/// - Every value of type `Self` must contain a valid `Inner` at offset 0;
+/// - `Self` must not impose additional safety invariants on the `Inner` prefix;
+/// - The methods of this trait should not be overriden.
+pub unsafe trait HasPrefixField<Inner>: Sized {
+    /// This constant should *always* be evaluated before relying on the safety
+    /// guarantees of this trait.
+    const ASSERT: () = ();
+
+    #[inline(always)]
+    fn as_prefix_gc(gc: Gc<'_, Self>) -> Gc<'_, Inner> {
```
kjarosh, line 19: I'd add some docs on why we would need this method and what can we do with it.

core/macros/src/lib.rs
kjarosh: That makes sense, thanks! In this case wouldn't it be better to assign a more unique name to the constant? If we defined two such constants in different traits, we would have a conflict when one struct implemented both, right?

kjarosh (approved):
LGTM, thank you. That looks a lot better than before!

Issue #19697, opened 2025-03-03
Context menu appears off screen in Firefox extension on Android


torokati44:
### Describe the bug

When opening the context menu on Firefox Android, it's unusable, because it's barely visible.

See:
![Image](https://github.com/user-attachments/assets/95b4c5b1-78b0-4220-9e1d-76346bcfcffd)

### Expected behavior

Context menu appears where I held down my finger.

### Content Location

https://z0r.de/7000?flash

### Affected platform

Browser's extension

### Operating system

Android 15

### Browser

Firefox 135

### Additional information

It works fine on a proper Ruffle embed, like: https://z0r.de/7000?ruffle

PR #20367, opened 2025-05-07
web: Fix context menu positioning in quirks mode

kjarosh:
When in quirks mode and the body was smaller than the content, context menu was constrained to the bounds of the body. This patch makes the context menu work properly in both quirks and no quirks modes.

* Fixes https://github.com/ruffle-rs/ruffle/issues/19697


danielhjacobs (approved):
Just to note, if I understand correctly the issue is we want the number 2 behavior from https://www.w3.org/TR/cssom-view/#dom-element-clientwidth and https://www.w3.org/TR/cssom-view/#dom-element-clientheight regardless of quirks mode or not, but in non-quirks mode that behavior only applies for the `documentElement` and in quirks mode it only applies for the `body`, so we have to use https://www.w3.org/TR/cssom-view/#dom-window-innerwidth and https://www.w3.org/TR/cssom-view/#dom-window-innerheight, which always has the same behavior.

It's worth noting the behavior of `window.innerWidth` differs from the behavior of https://www.w3.org/TR/cssom-view/#dom-element-clientwidth in non-quirks mode if a scrollbar exists, as the former includes the size of the scrollbar and the latter does not. Though I don't think it matters for use because the context menu doesn't have a scrollbar.

danielhjacobs:
Correction: as mentioned on Discord, the behavior differs when the document itself has a scrollbar, which may come up. Not sure how big a deal that may be though.

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -1641,18 +1641,17 @@ export class InnerPlayer {
         const contextMenuRect = this.contextMenuElement.getBoundingClientRect();
 
         // Keep the entire context menu inside the viewport.
-        // TODO: Allow the context menu to escape the document body while being mindful of scrollbars.
         const overflowX = Math.max(
             0,
             event.clientX +
                 contextMenuRect.width -
-                document.documentElement.clientWidth,
+                window.innerWidth,
```
danielhjacobs, line 1648: Alternative:

`(document.scrollingElement || document.body).clientWidth`

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -1641,18 +1641,17 @@ export class InnerPlayer {
...
         );
         const overflowY = Math.max(
             0,
             event.clientY +
                 contextMenuRect.height -
-                document.documentElement.clientHeight,
+                window.innerHeight,
```
danielhjacobs, line 1654: Alternative:

`(document.scrollingElement || document.body).clientHeight`

danielhjacobs:
So for the final note:

I'm fine with the current code, but it does treat scroll-able pages differently. If we want to treat them the same, the suggested code may be better. Hopefully it passes the new tests too.

kjarosh:
Fixes, could you take a second look @danielhjacobs ?

danielhjacobs (approved):
It may be better to reference this as the viewport element (viewportElement) instead of the root element (rootElement). MDN defines documentElement as the root element regardless of quirks mode: https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement. We may also wish to say in standards mode scrollingElement returns the documentElement (to avoid confusion from someone who reads the comment and thinks it returns document itself).

kjarosh:
> It may be better to reference this as the viewport element (viewportElement) instead of the root element (rootElement). MDN defines documentElement as the root element regardless of quirks mode: https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement.

I guess it depends on where you look. This page uses the term "root element" in a different way: https://developer.mozilla.org/en-US/docs/Web/API/Element/clientWidth

danielhjacobs:
The grammar is odd with the extra parentheses, but I think that page means:

> When clientWidth is used on the root element (the <html> element), or on <body> if the document is in quirks mode, the viewport's width (excluding any scrollbar) is returned.

Distinguishing the situations to say in quirks mode the behavior applies for the body **instead** of the root element.

kjarosh:
Oh, that actually makes sense. I was confused :sweat_smile:

Issue #20292, opened 2025-04-28
System fonts with multiple "full name" entries in metadata bug


1ost:
### Describe the bug

Ruffle encounters an issue when attempting to identify and load system fonts that contain multiple 'full name' entries within their metadata as according to @kjarosh (on discord). 

"the font contains multiple full names, Ruffle takes the first one and the first one in this case is a MacOS name with unsupported encoding"

Warning that gets logged into the console when one of those font files is processed :
""WARN web/src/builder.rs:449 Font /static/font/FreeSans.ttf at index 0 has no full name, using URL as font name""

### Expected behavior

"full name" of font to be properly recognized

### Content Location

Sample font that this issue is occuring with can be downloaded from this website:

https://fonts2u.com/download/free-sans.font

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Firefox

danielhjacobs:
See also: https://github.com/ruffle-rs/ruffle/pull/13361#issue-1919969490 (we should probably put this info somewhere more useful than a PR description)

PR #20368, opened 2025-05-08
web: Fix registering TTF fonts with non-Unicode names

kjarosh:
We don't support non-Unicode names, and in case such name appeared, we failed to register it properly.

* Progresses https://github.com/ruffle-rs/ruffle/issues/20292#issuecomment-2836494332

### DIFF
--- a/web/src/builder.rs
+++ b/web/src/builder.rs
@@ -438,9 +438,12 @@ impl RuffleInstanceBuilder {
         index: u32,
         player: &mut Player,
     ) {
+        // TODO A font may have multiple full names, what then?
         let full_name = face
             .names()
             .into_iter()
+            // Currently only unicode names are supported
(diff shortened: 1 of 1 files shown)


torokati44:
Doesn't this assume that a font has at least one Unicode full name? And if it doesn't, it's not registered? I get that it's still better to use a fallback then, than erroring out...

kjarosh:
> Doesn't this assume that a font has at least one Unicode full name? And if it doesn't, it's not registered? I get that it's still better to use a fallback then, than erroring out...

It is being registered with its URL as the name. Not sure how useful this is, but I'm not touching that behavior as I haven't checked how it works in Flash Player.

evilpie:
Registering with the name being a URL is probably worthless, but I guess I didn't want to push back even more in #17668.

evilpie (approved):

PR #20369, opened 2025-05-08
avm1: Remove `Gc` indirection for `SuperObject`

moulins:
Follow-up to #20204.

### DIFF
--- a/core/src/avm1/object/super_object.rs
+++ b/core/src/avm1/object/super_object.rs
@@ -8,7 +8,7 @@ use crate::avm1::function::ExecutionReason;
 use crate::avm1::object::{search_prototype, ExecutionName};
 use crate::avm1::{NativeObject, Object, Value};
 use crate::string::AvmString;
-use gc_arena::{Collect, Gc};
+use gc_arena::Collect;
 use ruffle_macros::istr;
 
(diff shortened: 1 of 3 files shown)


core/src/avm1/object/super_object.rs
```diff
@@ -18,44 +18,47 @@ use ruffle_macros::istr;
...
-pub struct SuperObject<'gc>(Gc<'gc, SuperObjectData<'gc>>);
+pub struct SuperObject<'gc> {
+    /// The object present as `this` throughout the superchain.
+    this: Object<'gc>,
+
+    /// The prototype depth of the currently-executing method.
+    depth: u8,
+
+    /// Adds a niche, so that enums contaning this type can use it for their discriminant.
+    _niche: crate::utils::ZeroU8,
```
kjarosh, line 29: How does this work exactly? Shouldn't Rust reuse this space for the discriminant anyway?

moulins: Rust will never overlap a discriminant with padding, because padding can have *any* value (it's roughly like an invisible `MaybeUninit<[u8; _]>` field); adding the `ZeroU8` field instead says that any `SuperObject` *must* have a zero byte at this offset.

PR #20380, opened 2025-05-09
chore: Reformat ActionScript code

kjarosh:
Reformat AS code: fix .editorconfig violations, remove Adobe documentation, fix obvious formatting mistakes.

Note that this doesn't supersede (or semantically conflict with) https://github.com/ruffle-rs/ruffle/pull/18439. It does fix some formatting mistakes too, but it focuses on fixing .editorconfig violations and removing Adobe documentation.


evilpie:
Should we really be removing the top comment from the autgenerated files?

kjarosh:
> Should we really be removing the top comment from the autogenerated files?

I'm not sure if they have any value. The files were originally generated, but it makes no difference on how we operate on them today. Those comments were more adequate as a commit message in my opinion.

danielhjacobs:
If we remove the messages, we may want to, just for politeness, keep some sort of acknowledgment of https://github.com/golfinq/ActionScript_Event_Builder somewhere, just to at least credit golfinq for the work.

kjarosh:
> If we remove the comments, we may want to, just for politeness, keep some sort of acknowledgment of https://github.com/golfinq/ActionScript_Event_Builder somewhere, just to at least credit golfinq for the work.

Note that the tool is acknowledged in the git history, additionally I've added a section to the AS3's README.

danielhjacobs (approved):

PR #20392, opened 2025-05-11
desktop: Implement font sorting using fontconfig

kjarosh:
This PR is the first step towards proper font/glyph fallback for device text.

Fontconfig is a library available for most platforms, but most commonly used in Linux distributions. It's a state-of-the-art mechanism of looking up and sorting fonts. By using fontconfig we get a proper font list that should be used when rendering device text. Flash Player requires fontconfig as a dependency, and it appears it uses it for device text too.

In this PR, we only supply a list of fonts to the HTML layout code, but it's not being used for glyph fallback yet.

The next steps include:
* implementing proper glyph fallback based on the sorted list of fonts,
* using fontconfig for GUI text too.


kjarosh:
I guess the main question is how to bundle fontconfig. We can link to it dynamically or load the library during runtime. Also the question is whether to include it on MacOS and Windows. Fontconfig can be available on Windows and MacOS too (although unlikely).

Currently I'm bundling it everywhere and loading libfontconfig during runtime.

It's a bummer we cannot easily set default features per platform :/

desktop/src/backends/ui.rs
```diff
@@ -401,3 +395,130 @@ impl UiBackend for DesktopUiBackend {
...
+}
+
+#[cfg(feature = "fontconfig")]
+fn fontconfig_sort_device_fonts(
+    query: &FontQuery,
+    register: &mut dyn FnMut(FontDefinition),
+) -> Vec<FontQuery> {
+    use fontconfig::{FontFormat, Pattern};
+
+    lazy_static::lazy_static! {
```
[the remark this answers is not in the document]
kjarosh: Done, thanks

kjarosh:
Also switched from our fork of fontconfig to the official crate as our issues have been fixed!

core/src/library.rs
```diff
@@ -579,6 +583,60 @@ impl<'gc> Library<'gc> {
...
+        renderer: &mut dyn RenderBackend,
+        gc_context: &Mutation<'gc>,
+    ) -> Vec<Font<'gc>> {
+        let query = FontQuery::new(FontType::Device, name.to_owned(), is_bold, is_italic);
+        if let Some(fonts) = self.font_sort_cache.get(&query) {
+            return fonts.clone();
+        }
+
+        let fonts = self.sort_device_fonts(&query, ui, renderer, gc_context);
+        self.font_sort_cache.insert(query, fonts.clone());
```
adrian17, line 636: not that it's important for perf, but: couldn't this avoid a clone if the function returned a borrow? It should be possible, right? The only user in later commits just borrows the `.first()` anyway.

core/src/library.rs
```diff
@@ -579,6 +583,60 @@ impl<'gc> Library<'gc> {
...
+        gc_context: &Mutation<'gc>,
+    ) -> Vec<Font<'gc>> {
+        // First, ask the backend to sort the fonts for us.
+        let fonts = ui.sort_device_fonts(query, &mut |definition| {
+            self.register_device_font(gc_context, renderer, definition)
+        });
+
+        let fonts: Vec<Font<'gc>> = fonts
+            .iter()
+            .filter_map(|font_query| self.device_fonts.get(font_query))
```
adrian17, line 600: I don't entirely understand why this needs to lookup the returned queries, just after we registered them.
Would passing `|query, definition|` in the `sort_device_fonts` callback and pushing them to the vec there work too?
It's probably not important to refactor this, but if this stays this way, could you leave some comment about why this is written this way?

desktop/src/backends/ui.rs
```diff
@@ -408,3 +395,127 @@ impl UiBackend for DesktopUiBackend {
...
+    if query.is_italic {
+        pattern.add_integer(fontconfig::FC_SLANT, fontconfig::FC_SLANT_ITALIC);
+    }
+
+    let font_set = pattern.sort_fonts(true);
+    let mut font_queries = Vec::new();
+    for font in font_set.iter() {
+        let is_ttf = font
+            .format()
+            .is_ok_and(|f| matches!(f, FontFormat::TrueType));
```
adrian17, line 474: do we want some info log, if our query matches a font but we happen to not be able to load it as it's not TTF?

desktop/src/backends/ui.rs
```diff
@@ -408,3 +395,127 @@ impl UiBackend for DesktopUiBackend {
...
+        )
+        else {
+            continue;
+        };
+
+        let Ok(index) = index.try_into() else {
+            continue;
+        };
+
+        let is_bold = weight >= fontconfig::FC_WEIGHT_BOLD;
```
adrian17, line 500: just to be sure, does it mean the query can return fonts that don't actually in bold-ness from the actual query?

adrian17 (approved):

desktop/src/backends/ui.rs
kjarosh: Yes, certainly. When the user has only bold fonts installed, a bold font will be returned. Also the user can configure fontconfig in such a way that a bold font will match a non-bold query. That's also something that can be done in Flash Player.

core/src/library.rs
kjarosh: I don't think I understand. The callback is for registering fonts only, you cannot assume that each call to `register` will result in a returned font and vice-versa. We'd need to look up fonts anyway, because `register_device_font` returns nothing.

I'm just using the existing mechanism from `load_device_font` here.

desktop/src/backends/ui.rs
kjarosh: Added

core/src/library.rs
kjarosh: Added a comment

core/src/library.rs
adrian17: To be honest, I _did_ kinda assume that if we call `self.register_device_font(definition)` with a font that comes with an associated query, then `self.device_fonts.get(query)` for that query will return... that font?
(And if it doesn't, because loading failed for whatever reason, then we could still return success status from `register_device_font` and detect this way)?

Anyway, it's not a big deal or perf issue, it's just this code looked extremely like the pattern
```py
for key in keys:
    map[key] = load_font(key)
result_list = []
for key in keys:
    if key in map:
        result_list.append(map[key])
```
vs
```py
result_list = []
for key in keys:
    font = load_font(key);
    map[key] = font
    result_list.append(font)
```
So I couldn't not ask about it 😅 
Gonna approve this either way.

adrian17 (approved):

core/src/library.rs
kjarosh: > To be honest, I _did_ kinda assume that if we call `self.register_device_font(definition)` with a font that comes with an associated query, then `self.device_fonts.get(query)` for that query will return... that font? (And if it doesn't, because loading failed for whatever reason, then we could still return success status from `register_device_font` and detect this way)?

It's not always true, a font won't be registered under the same query, and an existing font will be used instead. We would have to register + query anyway.

PR #20402, opened 2025-05-12
chore: Update translations

RuffleBuild:

### DIFF
--- a/web/packages/extension/assets/_locales/zh_TW/messages.json
+++ b/web/packages/extension/assets/_locales/zh_TW/messages.json
@@ -26,6 +26,12 @@
     "settings_log_level_info": {
         "message": "資訊"
     },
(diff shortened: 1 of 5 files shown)


web/packages/extension/assets/_locales/zh_TW/messages.json
```diff
@@ -26,6 +26,12 @@
     "settings_log_level_info": {
         "message": "資訊"
     },
+    "settings_log_level_debug": {
+        "message": "除錯"
+    },
+    "settings_log_level_trace": {
+        "message": "Trace"
```
torokati44, line 33: I guess this might be understood as is (some other strings are also kept in Latin script, such as "Flash", and "ActionScript"), and can always be translated (or even just transliterated at least) later...

PR #20417, opened 2025-05-16
desktop: Linux - add `StartupWMClass` to the desktop file

fiftydinar:
It's useful setting this as a fallback, if the `.desktop` file is not named the same as app ID (`rs.ruffle.Ruffle.desktop`).
As flatpaks always name their `.desktop` files as `app-id.desktop`, they are not affected.
As a bonus, another XDG desktop standard is respected.

This is useful for user-wide package managers which name their `.desktop` files differently to not collide with existing `.desktop` files (like [AM AppImage Manager](https://github.com/ivan-hc/AM)).

According to my tests, `wmclass` is the same in both Wayland & X11, so this works well.

Ending result is that Ruffle associates the window & icon correctly, even if `.desktop` file is named differently, while not causing any regressions.

### DIFF
--- a/desktop/packages/linux/rs.ruffle.Ruffle.desktop
+++ b/desktop/packages/linux/rs.ruffle.Ruffle.desktop
@@ -65,7 +65,6 @@ Comment[zh_CN]=播放 Flash 游戏和动画
 Comment[zh_TW]=播放 Flash 遊戲和動畫
 Comment=Play Flash games & movies
 Icon=rs.ruffle.Ruffle
-StartupWMClass=rs.ruffle.Ruffle
 Exec=ruffle %u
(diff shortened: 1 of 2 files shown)


fiftydinar:
> Hi, thanks for the contribution! Could you update only the .in file? The other one will update automagically.

Done

> Also nit: please add a prefix to the commit summary, e.g. `desktop: `.

I'll look on how to do that for the existing commit or force push if it's allowed.

kjarosh:
> I'll look on how to do that for the existing commit or squash/force-push if it's allowed.

Oh no worries then, I'll squash it myself

kjarosh (approved):
LGTM, thank you

PR #20429, opened 2025-05-18
web: Switch from `unload` to `pagehide` event

### LINKED ISSUE #19704: http
### Describe the bug

I have tested web script with pagespeed https://pagespeed.web.dev and have these errors:

1. Mixed Content: The page at 'https://www.example.com/' was loaded over HTTPS, but requested an insecure frame 'http://developers.google.com/#_methods=onPlusOne%2C_ready%2C_close%2C_open%2C_resizeMe%2C_renderstart%2Concircled%2Cdrefresh%2Cerefresh&id=I0_1741107425921&_gfid=I0_1741107425921&pfname=&rpctoken=15460460'. This request has been blocked; the content must be served over HTTPS.

2. Unload event listeners are deprecated and will be removed.
	[webpack://ruffle-selfhosted/../core/dist/ruffle_web.js:1074:13](https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.c16928fec031cd68ebe6.js)

### Expected behavior

   1. The Ruffle script should not trigger Mixed Content errors by attempting to load insecure resources (http://) on an HTTPS page.  All external dependencies should be loaded over HTTPS to comply with security policies and avoid browser blocking.

 2.    The script should not rely on unload event listeners, as they are deprecated and will be removed in future browser versions. The script should use modern alternatives to ensure future compatibility.

I expected the script to run without security-related warnings and deprecated feature notices in the browser console.

### Content Location

https://www.ooo.lt/zaidimas/pasirupink_suneliu/

### Affected platform

Online demo

### Operating system

Windows 11, all browsers

### Browser

_No response_

### Additional information

First of all, I want to sincerely thank you for creating such an incredible script. Ruffle has saved millions of SWF games that were developed over many years, allowing millions of people to continue playing their favorite games. Your work has made a huge impact, and I truly appreciate it!

Regarding the issue, if needed, I can provide additional logs or debugging information to help resolve it. Let me know if there's anything specific I should test or check.

Toad06:
The `unload` event is deprecated, and Chrome even plans to remove support for it entirely:
https://developer.chrome.com/docs/web-platform/deprecating-unload

The `pagehide` event is now used as a replacement. A side effect of this change, however, is that it makes pages eligible for the back/forward cache (bfcache): as a result, instances of Ruffle may no longer be recreated from scratch when navigating back to the page, but instead "revived" from their previous state. When this happens, the following message is displayed to the user:

restored

Closes #19704.

### DIFF
--- a/web/packages/core/src/internal/player/inner.tsx
+++ b/web/packages/core/src/internal/player/inner.tsx
@@ -2080,29 +2080,63 @@ export class InnerPlayer {
     /**
      * Show a dismissible message in front of the player.
      *
-     * @param message The message shown to the user.
+     * @param message The message shown to the user, which can be a string or element.
      */
-    public displayMessage(message: string): void {
-        const div = document.createElement("div");
-        div.id = "message-overlay";
-        const messageDiv = document.createElement("div");
-        messageDiv.className = "message";
-        const messageP = document.createElement("p");
-        messageP.textContent = message;
-        messageDiv.appendChild(messageP);
-        const buttonDiv = document.createElement("div");
-        const continueButton = document.createElement("button");
-        continueButton.id = "continue-btn";
-        continueButton.textContent = text("continue");
-        buttonDiv.appendChild(continueButton);
-        messageDiv.appendChild(buttonDiv);
-        div.appendChild(messageDiv);
-        this.container.prepend(div);
+    private displayMessageOrElement(message: string | HTMLDivElement): void {
+        const messageOverlay = (
+            <div id="message-overlay">
+                <div class="message">
+                    {message instanceof HTMLDivElement ? (
+                        message
+                    ) : (
+                        <p>{message}</p>
+                    )}
+                    <div>
+                        <button id="continue-btn">{text("continue")}</button>
+                    </div>
+                </div>
+            </div>
+        );
+
+        this.container.prepend(messageOverlay);
+
         (
             this.container.querySelector("#continue-btn") as HTMLButtonElement
         ).onclick = () => {
-            div.parentNode!.removeChild(div);
+            messageOverlay.parentNode!.removeChild(messageOverlay);
         };
+    }
+
+    /**
+     * Show a dismissible message in front of the player.
+     *
+     * @param message The message shown to the user.
+     */
+    public displayMessage(message: string): void {
+        this.displayMessageOrElement(message);
+    }
+
+    /**
+     * Inform the user that the browser restored the file from the back/forward cache.
+     */
+    protected displayRestoredFromBfcacheMessage(): void {
+        // Do not display the message if another one is already shown.
+        if (this.container.querySelector("#message-overlay") !== null) {
+            return;
+        }
+        const message = textAsParagraphs("message-restored-from-bfcache");
+        this.displayMessageOrElement(message);
+
+        // Remove the message element if it doesn't fit in the container, to avoid potential blocking situations.
+        const messageOverlay = this.container.querySelector(
+            "#message-overlay",
+        )! as HTMLElement;
+        if (
+            messageOverlay.scrollWidth > messageOverlay.offsetWidth ||
+            messageOverlay.scrollHeight > messageOverlay.offsetHeight
+        ) {
+            messageOverlay.parentNode!.removeChild(messageOverlay);
+        }
     }
 
     /**
--- a/web/src/lib.rs
+++ b/web/src/lib.rs
@@ -135,6 +135,7 @@ struct RuffleInstance {
     key_down_callback: Option<JsCallback<KeyboardEvent>>,
     key_up_callback: Option<JsCallback<KeyboardEvent>>,
     paste_callback: Option<JsCallback<ClipboardEvent>>,
+    pageshow_callback: Option<JsCallback<PageTransitionEvent>>,
     pagehide_callback: Option<JsCallback<PageTransitionEvent>>,
     focusin_callback: Option<JsCallback<FocusEvent>>,
     focusout_callback: Option<JsCallback<FocusEvent>>,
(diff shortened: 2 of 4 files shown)


web/packages/core/src/internal/player/inner.tsx
```diff
@@ -2034,6 +2034,34 @@ export class InnerPlayer {
...
+        continueButton.textContent = text("continue");
+        buttonDiv.appendChild(continueButton);
+        messageDiv.appendChild(buttonDiv);
+        div.appendChild(messageDiv);
+        this.container.prepend(div);
+        (
+            this.container.querySelector("#continue-btn") as HTMLButtonElement
+        ).onclick = () => {
+            div.parentNode!.removeChild(div);
+        };
```
danielhjacobs, line 2062: This whole file is a tsx component (thanks to tsx-dom), couldn't we take advantage of that? We don't do so enough.

In other words, appendChild a whole tsx component instead of doing a bunch of createElement (see #18515).

danielhjacobs: Or actually, why not just use `displayMessage`: <https://github.com/ruffle-rs/ruffle/blob/cf2995c71fef77ebf447b3289f0d5f7d83ef86cd/web/packages/core/src/internal/player/inner.tsx#L2014>

danielhjacobs: Or, if we don't want to expand the abilities of `displayMessage`:

```ts
/**
 * Show a dismissible message in front of the player.
 *
 * @param message The message shown to the user.
 */
protected displayMessageOrElement(message: string | HTMLDivElement) {
    const div = (
        <div id="message-overlay">
            <div class="message">
                {message instanceof HTMLDivElement ? (
                    message
                ) : (
                    <p>{message}</p>
                )}
                <div>
                    <button id="continue-btn">{text("continue")}</button>
                </div>
            </div>
        </div>
    );

    this.container.prepend(div);

    (this.container.querySelector("#continue-btn") as HTMLButtonElement).onclick = () => {
        div.parentNode!.removeChild();
    };
}

public displayMessage(message: string) {
    displayMessageOrElement(message);
}

    protected displayRestoredFromBfcacheMessage(): void {
        if(this.container.querySelector("#message-overlay") !== null) {
            // Do not display the message if another one is already shown - it is not that important.
            return;
        }
        const message = textAsParagraphs("message-restored-from-bfcache");
        displayMessageOrElement(message);
    }

Comment (Toad06) [reply web/packages/core/src/internal/player/inner.tsx:2062]: Thanks for the review and your suggestions — I learned something new about tsx components from you.

Do you have a preference on whether we should expand the definition of `displayMessage`?

By the way, is it expected that `npm run format` doesn't affect .tsx files? I had to fix some formatting manually. Looking at it again, I noticed there's still a small issue that didn't prevent the checks from passing — I wrote `if(` instead of `if (`.

Comment (danielhjacobs) [reply web/packages/core/src/internal/player/inner.tsx:2062]: #20448

File: web/packages/core/src/internal/player/inner.tsx [hunk 2]
```diff
@@ -2080,31 +2080,66 @@ export class InnerPlayer {
...
+     */
+    protected displayRestoredFromBfcacheMessage(): void {
+        // Do not display the message if another one is already shown.
+        if (this.container.querySelector("#message-overlay") !== null) {
+            return;
+        }
+        const message = textAsParagraphs("message-restored-from-bfcache");
+        this.displayMessageOrElement(message);
+
+        // Remove the message element if it doesn't fit in the container, to avoid potential blocking situations.
```
danielhjacobs, line 2130: I assume you mean this situation?

https://codepen.io/danieljacobs/pen/emmqEgE

Toad06: Nice trick! That said, it wouldn't be possible if the context menu were disabled.

Still, I don't think that's a big deal, since the message should appear for most files anyway.

danielhjacobs (approved):

web/src/lib.rs
```diff
@@ -782,18 +789,36 @@ impl RuffleHandle {
...
+            ));
+
+            // Create pagehide event handler.
+            instance.pagehide_callback = Some(JsCallback::register(
+                &window,
+                "pagehide",
+                false,
+                move |_js_event: PageTransitionEvent| {
                     let _ = ruffle.with_core_mut(|core| {
                         core.flush_shared_objects();
```
kjarosh, line 818: I think this shouldn't be needed, as we do that when dropping the player.

kjarosh: (that's not blocking obviously, but we could get rid of this callback altogether I think)

Toad06: The player is not dropped when leaving the page; I tested what you suggested, though, and it didn't save the data from the  `SharedObject`s.

kjarosh: Aw :/ it's interesting why the player isn't being dropped. Nevermind then

kjarosh (approved):

n0samu:
Hmm, was it really necessary to add a message "warning" the user about a behavior they'd probably prefer anyway?

https://github.com/user-attachments/assets/88df268c-0d8e-45ca-9e42-9d932c24bd51

Toad06:
This was added to emphasize that this is a browser behavior, not Ruffle's, as it still seems unexpected (and, personally, undesirable in many cases) to me that a movie or game would instantly resume from where it left off when the user navigated away.

Sure, for anything like menus or splash images, this message is certainly not useful...

But for games or movies, I think the opposite. The movie in your video also has loud audio that resumes instantly; if the user has headphones on and has turned up the volume in the meantime, I don't think they'll be happy with this browser behavior. :D

There is also a technical reason: the audio context is sometimes not properly restored and starts to "stutter". Nothing can fix this broken state until the file is completely reloaded.

But I understand that others might feel differently. In an ideal world, the extension would offer a "Do not show again in the future" checkbox.

n0samu:
Oh weird, I didn't experience the audio problem. Not sure exactly what we should do

Issue #7868, opened 2022-09-04
web: Ruffle does not support the Flash plugin's JavaScript API functions (Play, StopPlay, etc)


n0samu:
### Describe the bug

[This tutorial](https://permadi.com/tutorial/flashjscommand/) shows how to call the Flash plugin's built-in DOM methods from JavaScript, including: 
- Play
- StopPlay
- Rewind
- GotoFrame
- Zoom
- SetVariable
- GetVariable

None of these work in Ruffle currently. For example, you can't call `document.querySelector("ruffle-embed").StopPlay()`.

### Expected behavior

Calling the Flash plugin's DOM methods should work for Ruffle DOM objects. And ideally, Ruffle should also add those methods to the original `object`/`embed` tags before replacing them with `ruffle-object`/`ruffle-embed`, so JavaScript that already grabbed a reference to those elements can call those methods too.

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

Firefox 104

### Additional information

Another thing to keep in mind is that the behavior of typical JavaScript changes after Ruffle polyfills embed/object elements. The tutorial linked above uses `window.document['myFlashMovie']` to refer to the `embed` element with `name="myFlashMovie"`; this doesn't work for `ruffle-embed` elements. And it falls back to using `document.embeds` as a second method of finding the `embed` element, which again doesn't work for `ruffle-embed` elements. So I think this kind of JavaScript will inevitably need to be modified to work with Ruffle, unless we take great pains to polyfill every obscure behavior, or change how polyfill works in a significant way.

n0samu:
Here is a message about this topic from @Herschel on Discord, May 4, 2021:
> any thoughts on re-arranging RufflePlayer in this way:
```ts
class RufflePlayer extends HTMLElement{
  // ExternalInterface callbacks get exposed here
  externalInterface: any;
}

class RuffleEmbed/RuffleObject extends HTMLElement {
  player: RufflePlayer;

  // Flash plugin classic JS APIs
  void Play();
  void StopPlay();
  // etc.

  // ExternalInterface callbacks would also get mirrored here
  void SomeAsExposedFunction();
}
```
> The motivation is for implementing the classic APIs that the Flash plugin exposes on the polyfilled DOM elements, without any conflicts with Ruffle's API. The polyfill elements would wrap the RufflePlayer and expose the Flash API and any externalInterface callbacks.

Toad06:
Other related issues: #3857, #4395.
[...]

PR #20434, opened 2025-05-19
web: expose more functionality in the FlashAPI

InvoxiPlayGames:
This implements more of the JavaScript API exposed by FP (as per https://web.archive.org/web/20100710000820/http://www.adobe.com/support/flash/publishexport/scriptingwithflash/scriptingwithflash_03.html)

* CurrentFrame
* GotoFrame
* IsPlaying
* Play
* StopPlay
* TotalFrames

### DIFF
--- a/core/src/player.rs
+++ b/core/src/player.rs
@@ -2312,7 +2312,7 @@ impl Player {
                 .root_clip()
                 .and_then(|root| root.as_movie_clip())
                 .map(|clip| clip.current_frame());
-            
+
             this.total_frames = update_context
                 .stage
                 .root_clip()
--- a/core/src/player.rs
+++ b/core/src/player.rs
@@ -366,6 +366,9 @@ pub struct Player {
     /// The first frame is frame 1.
     current_frame: Option<u16>,
 
+    /// The total number of frames in the main timeline, if available.
+    total_frames: Option<u16>,
+
     /// How Ruffle should load movies.
     load_behavior: LoadBehavior,
 
@@ -836,6 +839,25 @@ impl Player {
             } else {
                 mc.play();
             }
+        }
+    }
+    fn is_playing_root_movie(context: &mut UpdateContext<'_>) -> bool {
+        if let Some(mc) = context
+            .stage
+            .root_clip()
+            .and_then(|root| root.as_movie_clip())
+        {
+            return mc.playing();
+        }
+        false
+    }
+    fn goto_frame_root_movie(context: &mut UpdateContext<'_>, frame: u16) {
+        if let Some(mc) = context
+            .stage
+            .root_clip()
+            .and_then(|root| root.as_movie_clip())
+        {
+            mc.goto_frame(context, frame, true)
         }
     }
     fn rewind_root_movie(context: &mut UpdateContext<'_>) {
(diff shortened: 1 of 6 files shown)


core/src/player.rs
```diff
@@ -2023,6 +2045,34 @@ impl Player {
...
+    pub fn total_frames(&self) -> Option<u16> {
+        self.total_frames
+    }
+
+    pub fn goto_frame(&mut self, frame: u16) {
+        self.mutate_with_update_context(|context| Self::goto_frame_root_movie(context, frame))
+    }
+
+    pub fn is_playing_movie(&mut self) -> bool {
+        self.mutate_with_update_context(|context| Self::is_playing_root_movie(context))
```
danielhjacobs, line 2057: Clippy seems to prefer this:

```rs
        self.mutate_with_update_context(Self::is_playing_root_movie)
```

danielhjacobs: https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure

danielhjacobs:
Relates to #7868

web/packages/core/src/public/player/flash.ts
```diff
@@ -3,15 +3,58 @@
...
+ * https://web.archive.org/web/20100710000820/http://www.adobe.com/support/flash/publishexport/scriptingwithflash/scriptingwithflash_03.html
+ *
  * Extra methods or properties may appear at any time, due to `ExternalInterface.addCallback()`.
  * It may even overwrite existing methods or properties.
  */
 export interface FlashAPI {
+    /**
+     * Returns the current frame index of the movie.
+     *
+     * @remarks the frame number starts from a 0 index, not 1 like in FP
```
Dinnerbone, line 15: To be clear, if this a Ruffle vs Flash thing or a Flash JavaScript vs ActionScript thing?

Is this method supposed to be 0 indexed in the Flash JS API?

(Same for other frame based methods)

InvoxiPlayGames: Flash JavaScript vs ActionScript

danielhjacobs: It might be clearer to change this (and other instances of this comment) to:

> the frame number starts from a 0 index, not 1 like in AS

Dinnerbone:
If you can please fix up the two review comments, we can take this in. In the future I'd like to revisit this with some tests, too, but it's okay to add these without for now.
Thank you so much!

weirdgore:
Is this functionality still currently planned to be implemented/these changes merged soon? The two review comments seem to be simple fixes

sombraguerrero:
Also @kjarosh pointed out in the Discord that we could use some tests on this. I can try to take a look at some of the others and see how it's done, just haven't done it before. But I'd be happy to pitch in, in an effort to get this merged.

sombraguerrero:
Upon doing some manual testing against a simple example in my Azure environment, I have verified that it works to retrieve the Flash movie object by conventional modern methods like getElementById and getElementsByName, but it does not work to retrieve it via dot syntax or performing index operations on the DOM, which is what most old Flash code that used this aspect of Flash requires. I _think_ what I'm trying to say is that Ruffle is not creating an HTML Object the way Flash once did.
[DOM Object Play Test](https://sombrastorage.z19.web.core.windows.net/ruffle/js_flash)

sombraguerrero:
After talking this through further on Discord and looking into it, the missing integration here should probably be its own TODO. I can log a separate issue so that someone can work on it. Meanwhile, I can attempt to provide a test for this one as my contribution to getting this piece of it out.

PR #20452, opened 2025-05-21
chore: Reformat more files according to .editorconfig

kjarosh:
Make more files adhere to the editorconfig formatting rules.


.gitattributes
```diff
@@ -14,5 +14,5 @@
 *.swf  binary
 *.swc  binary
 
-# Fix GH saying Ruffle is written in ActionScript 
+# Fix GH saying Ruffle is written in ActionScript
```
danielhjacobs, line 17: The comment, which predates us actually writing a lot of Ruffle in ActionScript, is a bit ironic now. Not sure if we should change it or what.

LICENSE.md
danielhjacobs, line 1: Much like my last comment, unrelated to the actual PR, but I have a strong suspicion this file is majorly outdated.

web/LICENSE.md
danielhjacobs, line 1: Same for this file

danielhjacobs (approved):

.gitattributes
kjarosh: Updated the comment

LICENSE.md
kjarosh: Moved to a discussion on the Discord server

PR #20460, opened 2025-05-22
ci: Remove lld workaround for linkme

torokati44:
This reverts https://github.com/ruffle-rs/ruffle/pull/16390 and https://github.com/ruffle-rs/ruffle/pull/16682, as https://github.com/dtolnay/linkme/issues/94 is claimed to be fixed by https://github.com/rust-lang/rust/pull/137685. See also: https://github.com/dtolnay/linkme/pull/88 and then https://github.com/dtolnay/linkme/pull/112.

### DIFF
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -79,13 +79,6 @@ jobs:
           sudo apt-get update
           sudo apt install -y libasound2-dev mesa-vulkan-drivers libudev-dev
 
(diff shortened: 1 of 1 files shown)


danielhjacobs (approved):
I think this means we may need to add this back in the future unfortunately:

https://github.com/rust-lang/rust/pull/137685/files#diff-862130c0af71054886913befc8084094485fa7d764d95f55644e033e34acd126R3396-R3397

PR #20466, opened 2025-05-23
 chore: Include attributions using cargo-attribution 

danielhjacobs:


Dinnerbone:
Why do we check this file into the repo and not just build it for distribution? Seems like a lot of noise 😅

danielhjacobs:
You mean the LICENSE.md itself? Not sure. Also wondering about why we have separately web/LICENSE.md and LICENSE.md.

This is also still a draft because I don't think .github/workflows/modify-license.yml actually works as I intended right now.

PR #20468, opened 2025-05-23
tests: Add support for multiple checks for an image & filters

kjarosh:
Multiple image checks are useful when we want to test more than one tolerance/max_outliers configurtion at one time.

For instance, we might want to check that the image has 0 outliers for tolerance 128, and 10 outliers for tolerance 5.

The configuration above would be written as:

```
[[image_comparisons.output.checks]]
tolerance = 128
max_outliers = 0

[[image_comparisons.output.checks]]
tolerance = 5
max_outliers = 10
```

Filters are useful when e.g. the given check only applies to a specific
platform. It's very common that a test output looks different on
different platforms. With filters we can write specific checks for
specific platforms instead of increasing tolerance for all of them.

An example configuration might look like this:

```
[[image_comparisons.output.checks]]
filter = 'not(arch = "aarch64")'
tolerance = 2

[[image_comparisons.output.checks]]
filter = 'arch = "aarch64"'
tolerance = 2
max_outliers = 10
```

In this configuration, we allow 10 outliers for aarch64, while not
allowing any for all other architectures.

These features will be useful for https://github.com/ruffle-rs/ruffle/pull/19259

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -628,6 +628,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "cfg-expr"
+version = "0.20.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e34e221e91c7eb5e8315b5c9cf1a61670938c0626451f954a51693ed44b37f45"
+dependencies = [
+ "smallvec",
+]
+
+[[package]]
 name = "cfg-if"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


torokati44:
I had a somewhat orthogonal idea to this: The ability to add a "tolerance map". This would essentially be the per-subpixel max of the differences of the expected and actual (per-platform) result images. What do you think about that?

torokati44:
My idea was that a tolerance map would be an option, instead of the two numbers we have now. I don't see why they would have to change more often than the expected output images, and given that they would be almost entirely black, with some smidges here and there, I expect they would compress extremely well. But I'm not so attached to this idea of course.

torokati44 (approved):

PR #20469, opened 2025-05-23
chore: Update translations

RuffleBuild:

### DIFF
--- a/web/packages/core/texts/ro-RO/messages.ftl
+++ b/web/packages/core/texts/ro-RO/messages.ftl
@@ -10,14 +10,21 @@ update-ruffle = Actualizează Ruffle
 ruffle-demo = Demo web
 ruffle-desktop = Aplicație desktop
 ruffle-wiki = Vezi wikiul Ruffle
+enable-hardware-acceleration = Se pare că accelerarea hardware este dezactivată. Deși Ruffle ar putea funcționa, va fi foarte lent. Puteți afla cum să activați accelerarea hardware accesând linkul de mai jos:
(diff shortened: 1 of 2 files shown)


web/packages/core/texts/ro-RO/messages.ftl
```diff
@@ -18,6 +18,7 @@ clipboard-message-title = Copierea și lipirea în Ruffle
 clipboard-message-copy = { " " } pentru copiere
 clipboard-message-cut = { " " } pentru decupare
 clipboard-message-paste = { " " } pentru lipire
+error-canvas-reload = Nu se poate reîncărca rendererul canvas atunci când acesta este deja utilizat.
```
torokati44, line 21: I feel like a "with" is missing here...?

colin-i: "reîncărca [cu] rendererul canvas" is "reload [with] the canvas renderer" but that is cacophony ("cacofonie"): "...ca cu", nobody is accepting that style. Instead of "cu" can be "adăugând/folosind/utilizând"(adding/using/utilizing), or instead of "reîncărca" can be "restarta"(restart). I changed it to "... utilizând ... folosit".

torokati44: I just meant (as also noted on CrowdIn) that it's not just the (canvas) renderer that's being reloaded, it's the entire player, with a different renderer (the canvas kind this time, unless it was already in use before, hence this error message). Thanks for fixing it!

colin-i: I didn't know that the entire player is reloaded.

Issue #16158, opened 2024-04-29
scanner: AVM1 Netconnection files cause the scanner to hang


danielhjacobs:
### Describe the bug

If using the command `cargo run --release --package=ruffle_scanner -- scan swf_files/ results.csv` with certain files that use AVM1 NetConnection in the folder swf_files/, the scanner hangs indefinitely and eventually the computer may freeze if you don't have a max memory limit set for the Terminal.

### Expected behavior

The files, even if they panic or have unimplemented features, should at least be able to have their errors detected and be scanned at least partially.

### Content Location

http://www.g2conline.org/network/test.swf
http://simplemapper.org/swf/map.swf

### Affected platform

Desktop app

### Operating system

Fedora Linux 39

sleepycatcoding:
This seems to be unrelated to the presence of NetConnection. Both of these SWFs have a function like this
```actionscript
function detectCpuSpeed(Void)
   {
      function fib(n)
      {
         return n <= 1 ? (n != 1 ? 0 : 1) : fib(n - 1) + fib(n - 2);
      }
      var _loc4_ = 0;
      var _loc6_ = new Date();
      var _loc5_ = _loc6_.getTime();
      while(true)
      {
         fib(12);
         var _loc2_ = new Date();
         var _loc3_ = _loc2_.getTime() - _loc5_;
         if(_loc3_ > 500)
         {
            break;
         }
         _loc4_ = _loc4_ + 1;
      }
      this._speed = _loc4_ <= 120 ? "slow" : "fast";
   }
```
But since the time returned by `new Date()` is fixed in the scanner, it gets stuck in this loop.

PR #20476, opened 2025-05-25
scanner: Make time non-deterministic

danielhjacobs:
Untested (as I made this PR on my phone with the GitHub UI) but should fix (theoretically) #16158. Can those who know more about this weigh in on whether the scanner using deterministic time may have some benefit, as I'm unsure what benefit it would have?

### DIFF
--- a/scanner/Cargo.toml
+++ b/scanner/Cargo.toml
@@ -12,7 +12,7 @@ workspace = true
 
 [dependencies]
 clap = { workspace = true }
-ruffle_core = { path = "../core", features = ["deterministic"] }
+ruffle_core = { path = "../core" }
 log = { workspace = true }
 walkdir = { workspace = true }
 serde = { workspace = true, features = ["derive"] }


husqwc:
I (and a lot of other people) ask that you keep TASing in mind. If something has to be made non-deterministic, please provide some sort of alternative which would keep determinism

danielhjacobs:
> I (and a lot of other people) ask that you keep TASing in mind. If something has to be made non-deterministic, please provide some sort of alternative which would keep determinism

For what it's worth this has nothing to do with any build of Ruffle that would actually be used to play games. The PR, as it is now, affects the scanner, a part of Ruffle that can check if swfs are valid and if they parse successfully with Ruffle. 

See https://github.com/ruffle-rs/ruffle?tab=readme-ov-file#scanner

It may also affect the exporter if expanded to do so, which is used to extract an image from an SWF:

https://github.com/ruffle-rs/ruffle?tab=readme-ov-file#exporter

Dinnerbone:
The exporter should be (or have an option to be) deterministic to be able to know if a given version of ruffle changed how a thing rendered. Scanner doesn't strictly need it, it sounds fine to be non deterministic and if needed we can add a command line flag for it.

Dinnerbone (approved):

Issue #16648, opened 2024-06-09
Crash on Hot_Air_2.swf


MavisCelus19201:
### Describe what you were doing

Launching the game causes an immediate crash.

### What does the crash message say?

panicked at core/src/library.rs:188:55:
called `Option::unwrap()` on a `None` value

### Ruffle Version

0.1.0-nightly (1d1eb6c966a17442c9e6a0730d0ba46f5d079879 2024-06-08)

### Affected platform

Desktop app

### Operating system

Manjaro 24.0.1 (Wynsdey) [64-bit]

### Additional information

Filename: Hot_Air_2.swf

### Render Info
Renderer: wgpu
Adapter Backend: Vulkan
Adapter Name: "Intel(R) HD Graphics 630 (KBL GT2)"
Adapter Device Type: IntegratedGpu
Adapter Driver Name: "Intel open-source Mesa driver"
Adapter Driver Info: "Mesa 24.0.8-manjaro1.1"
Enabled features: Features(TEXTURE_COMPRESSION_BC | FLOAT32_FILTERABLE | TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES | SHADER_UNUSED_VERTEX_OUTPUT)
[...]

Issue #16355, opened 2024-05-17
Error on https://play.penglyfe.com/


AjitForLife:
# Error Info
Error name: Error
Error message: panicked at core/src/library.rs:188:55:
called `Option::unwrap()` on a `None` value
Error stack:
```
Error: panicked at core/src/library.rs:188:55:
called `Option::unwrap()` on a `None` value
    at imports.wbg.__wbg_new_28c511d9baebfa89 (chrome-extension://donbcfbmhbcapadipfkeojnmajbakjdc/dist/69.js:3429:21)
    at ruffle_web.wasm.js_sys::Error::new::hc894352b761e2b75 (wasm://wasm/ruffle_web.wasm-0337a3be:wasm-function[12759]:0x9f6db0)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once{{vtable.shim}}::h883c9e54c2d3eed4 (wasm://wasm/ruffle_web.wasm-0337a3be:wasm-function[10527]:0x9b0c0e)
    at ruffle_web.wasm.std::panicking::rust_panic_with_hook::h32c80a64fe4de396 (wasm://wasm/ruffle_web.wasm-0337a3be:wasm-function[6533]:0x875d8e)
    at ruffle_web.wasm.core::panicking::panic_fmt::hc7427f902a13f1a9 (wasm://wasm/ruffle_web.wasm-0337a3be:wasm-function[10277]:0x9a98e8)
    at ruffle_web.wasm.core::panicking::panic::hb157b525de3fe68d (wasm://wasm/ruffle_web.wasm-0337a3be:wasm-function[10010]:0x992de9)
    at ruffle_web.wasm.core::option::unwrap_failed::hb5bacfb0dd292085 (wasm://wasm/ruffle_web.wasm-0337a3be:wasm-function[12259]:0x9e8087)
    at ruffle_web.wasm.ruffle_core::library::MovieLibrary::character_by_export_name::h72436535d599e2fd (wasm://wasm/ruffle_web.wasm-0337a3be:wasm-function[4251]:0x77e047)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once::hc75c44f1c1e6b2b6 (wasm://wasm/ruffle_web.wasm-0337a3be:wasm-function[1242]:0x36badf)
[...]

Issue #16288, opened 2024-05-11
Error on chrome-extension://pipjjbgofgieknlpefmcckdmgaaegban/player.html


bruhbruhlolol:
# Error Info
Error name: Error
Error message: panicked at core/src/library.rs:188:55:
called `Option::unwrap()` on a `None` value
Error stack:
```
Error: panicked at core/src/library.rs:188:55:
called `Option::unwrap()` on a `None` value
    at imports.wbg.__wbg_new_28c511d9baebfa89 (chrome-extension://pipjjbgofgieknlpefmcckdmgaaegban/dist/69.js:3393:21)
    at js_sys::Error::new::hdd220ef72a1ea21f (wasm://wasm/03ab1dba:wasm-function[13983]:0xb37155)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::h022c91248dc5bdaa (wasm://wasm/03ab1dba:wasm-function[11580]:0xae38b4)
    at std::panicking::rust_panic_with_hook::h9aabd906218897c3 (wasm://wasm/03ab1dba:wasm-function[7154]:0x9a8543)
    at core::panicking::panic_fmt::h3aff855fe938c13f (wasm://wasm/03ab1dba:wasm-function[11297]:0xadd371)
    at core::panicking::panic::h11a2021d92dc1cbb (wasm://wasm/03ab1dba:wasm-function[10819]:0xab0be3)
    at core::option::unwrap_failed::h0e0b231623e0d004 (wasm://wasm/03ab1dba:wasm-function[13432]:0xb25856)
    at ruffle_core::library::MovieLibrary::character_by_export_name::h4fde21fd7fdb37a4 (wasm://wasm/03ab1dba:wasm-function[5159]:0x8de0ce)
    at ruffle_core::avm1::globals::sound::attach_sound::h03374519ed5a8887 (wasm://wasm/03ab1dba:wasm-function[1223]:0x3cc3f2)
    at ruffle_core::avm1::function::Executable::exec::hc1a658ea02cf841b (wasm://wasm/03ab1dba:wasm-function[879]:0x1f83ef)
```

# Player Info
Allows script access: false
Renderer: wgpu
Adapter Backend: Gl
Adapter Name: "ANGLE (Intel, Intel(R) UHD Graphics (0x000046D1) Direct3D11 vs_5_0 ps_5_0, D3D11)"
Adapter Device Type: IntegratedGpu
Adapter Driver Name: ""
[...]

Issue #16127, opened 2024-04-26
Error on https://beatablemeat.github.io/games/hexic/


ArisuSanchez:
# Error Info
Error name: Error
Error message: panicked at core/src/library.rs:188:55:
called `Option::unwrap()` on a `None` value
Error stack:
```
Error: panicked at core/src/library.rs:188:55:
called `Option::unwrap()` on a `None` value
    at imports.wbg.__wbg_new_28c511d9baebfa89 (chrome-extension://pipjjbgofgieknlpefmcckdmgaaegban/dist/69.js:3393:21)
    at js_sys::Error::new::hf655a9c3d4762b78 (wasm://wasm/03a88812:wasm-function[13948]:0xb2e209)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::h0469bb069fdb7363 (wasm://wasm/03a88812:wasm-function[11540]:0xad7f42)
    at std::panicking::rust_panic_with_hook::h9aabd906218897c3 (wasm://wasm/03a88812:wasm-function[7112]:0x99f6c0)
    at core::panicking::panic_fmt::h3aff855fe938c13f (wasm://wasm/03a88812:wasm-function[11256]:0xad1a4b)
    at core::panicking::panic::h11a2021d92dc1cbb (wasm://wasm/03a88812:wasm-function[10777]:0xaa4c31)
    at core::option::unwrap_failed::h0e0b231623e0d004 (wasm://wasm/03a88812:wasm-function[13401]:0xb1c923)
    at ruffle_core::library::MovieLibrary::character_by_export_name::h40b16564872ab3ae (wasm://wasm/03a88812:wasm-function[5136]:0x8d4c3d)
    at ruffle_core::avm1::globals::sound::attach_sound::hda5023d16d392621 (wasm://wasm/03a88812:wasm-function[1222]:0x3c96d0)
    at ruffle_core::avm1::function::Executable::exec::h28d610c2ad9c1f4e (wasm://wasm/03a88812:wasm-function[885]:0x207806)
```

# Player Info
Allows script access: true
Renderer: wgpu
Adapter Backend: Gl
Adapter Name: "ANGLE (NVIDIA, NVIDIA GeForce RTX 3060 Ti (0x00002489) Direct3D11 vs_5_0 ps_5_0, D3D11)"
Adapter Device Type: Other
Adapter Driver Name: ""
Adapter Driver Info: ""
[...]

PR #20491, opened 2025-05-26
core: Assert that a character exists for each export

kjarosh:
This shouldn't change anything; according to my analysis, the "missing characters on looking up export" bug should be fixed. This patch only gets rid of the unwrap() and introduces better error handling in case something breaks in the future.

* Fixes https://github.com/ruffle-rs/ruffle/issues/16648
* Fixes https://github.com/ruffle-rs/ruffle/issues/16355
* Fixes https://github.com/ruffle-rs/ruffle/issues/16288
* Fixes https://github.com/ruffle-rs/ruffle/issues/16127
* Fixes https://github.com/ruffle-rs/ruffle/issues/15799

### DIFF
--- a/core/src/library.rs
+++ b/core/src/library.rs
@@ -164,6 +164,15 @@ impl<'gc> MovieLibrary<'gc> {
     /// Registers an export name for a given character ID.
     /// This character will then be instantiable from AVM1.
     pub fn register_export(&mut self, id: CharacterId, export_name: AvmString<'gc>) {
+        let character_exists = self.contains_character(id);
+        debug_assert!(character_exists);
+        if !character_exists {
+            tracing::error!(
+                "Tried to register export '{export_name}' for a non-existent character {id}"
+            );
+            return;
+        }
+
         self.export_characters.insert(export_name, id, false);
     }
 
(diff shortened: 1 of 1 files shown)


core/src/library.rs
```diff
@@ -158,6 +158,15 @@ impl<'gc> MovieLibrary<'gc> {
...
     /// This character will then be instantiable from AVM1.
     pub fn register_export(&mut self, id: CharacterId, export_name: AvmString<'gc>) {
+        let character_exists = self.contains_character(id);
+        debug_assert!(character_exists);
+        if !character_exists {
+            tracing::error!(
+                "Tried to register export '{export_name}' for a non-existent character {id}"
+            );
+            return;
+        }
```
moulins, line 168: There's nothing preventing a malformed SWF from exporting a character before defining it, so this probably shouldn't be an assert. I suspect such an export will work correctly in FP, but this would need testing.

kjarosh: I'm not sure how that works in Ruffle or FP, but it's currently impossible in Ruffle. The method `register_export` is called only when the character is registered.

Lord-McSweeney (approved):

PR #20494, opened 2025-05-27
chore: Update to wgpu 25 and egui git main

torokati44:
We may have to remove wgpu tracing (at least temporarily):
 - https://github.com/emilk/egui/pull/6744
 - https://github.com/emilk/egui/pull/6860
 - https://github.com/gfx-rs/wgpu/issues/5974
 - https://github.com/gfx-rs/wgpu/pull/5897
 - https://github.com/gfx-rs/wgpu/issues/5121

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -743,12 +743,13 @@ dependencies = [
 
 [[package]]
 name = "codespan-reporting"
-version = "0.11.1"
+version = "0.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3538270d33cc669650c4b093848450d380def10c331d38c768e34cac80576e6e"
+checksum = "fe6d2e5af09e8c8ad56c969f2157a3d4238cebc7c55f0a517728c38f7b200f81"
 dependencies = [
+ "serde",
  "termcolor",
- "unicode-width 0.1.14",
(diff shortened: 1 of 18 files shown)


torokati44:
Alternatively, we could declare a (fake) direct dependency on `wgpu-core` in the meantime, just to enable its `trace` feature. We also do the same with `profiling` and `getrandom` FWIW.

Dinnerbone:
I think I only ever used the trace feature once, in the early days when reporting wgpu bugs. It doesn't feel worth to keep it around anymore.

torokati44:
Yeah, especially since RenderDoc and Tracy and whatnot can still be used "live".

----
Oh, welcome back, @Dinnerbone ! :wave: :heart:

Dinnerbone:
Tested web for wgpu regressions and it all looks good
Testing desktop and, the file->recents menu behaves differently when there's a super long entry. Previously you'll still see newer stuff, but the long entry is long. Now you only see the long entry.

(this may be on me for having a long entry in there - or perhaps on us for not trying to shorten urls to just a filename...)

image


Not sure if this is a _blocker_ but it's awkward, my menu is unusable now unless I go clear it.

Dinnerbone (approved):
Approving, we can look at the menu in a followup

Issue #12321, opened 2023-07-26
Test Failure: avmplus Vector suite


Dinnerbone:
Completely passing suites are not shown.

## Testsuite `from_avmplus/as3/Vector/some`

| Test | Passes? | Expected | Actual |
| ---- | ---- | ---- | ---- |
| some no checker | :heavy_check_mark: | | |
| some checker not a function | :x: | `TypeError: Error #1034` | `TypeError: Error #1006` |
| some empty vector result | :heavy_check_mark: | | |
| some empty vector message empty | :heavy_check_mark: | | |
| some small vector result | :heavy_check_mark: | | |
| some small vector message | :heavy_check_mark: | | |
| some small vector result with a true | :heavy_check_mark: | | |
| some small vector message with a true | :heavy_check_mark: | | |
| Verify a value is in a Number vector | :heavy_check_mark: | | |
| Verify a value is not in a Number vector | :heavy_check_mark: | | |
| Verify a value is in a uint vector | :heavy_check_mark: | | |
| Verify a value is not in a uint vector | :heavy_check_mark: | | |
| Test thisObject param | :heavy_check_mark: | | |

## Testsuite `from_avmplus/as3/Vector/some_initializers`

| Test | Passes? | Expected | Actual |
| ---- | ---- | ---- | ---- |
| some no checker | :heavy_check_mark: | | |
| some checker not a function | :x: | `TypeError: Error #1034` | `TypeError: Error #1006` |
| some empty vector result | :heavy_check_mark: | | |
| some vector with no match | :heavy_check_mark: | | |
| some vector with match | :heavy_check_mark: | | |
| some function with no return value | :heavy_check_mark: | | |

Lord-McSweeney:
every, filter, filter_initialize, foreach, foreach_initializers, map, map_initializers, some, and some_initializers require porting Vector to ActionScript.

Lord-McSweeney:
(Technically, we could also do the coercions and argument count checking manually, but that wouldn't be very fun).

Lord-McSweeney:
The following tests are now passing:
- `Vector/initializer_ws`
- `Vector/bug_683507`
- `Vector/concat`
- `Vector/concat_with_initializers`
- `Vector/get`
- `Vector/insert_remove`
- `Vector/push`
- `Vector/put`
- `Vector/vectorIndexRangeExceptions`

Lord-McSweeney:
[...]

PR #20565, opened 2025-06-05
avm2: Vector indexing improvements

Lord-McSweeney:
These are unlikely to have any impact on real code, but they let 7 avmplus tests pass (progress on #12321). We also now have a fast path for setting integer properties on vectors.

### DIFF
--- a/core/src/avm2/object/vector_object.rs
+++ b/core/src/avm2/object/vector_object.rs
@@ -1,11 +1,11 @@
 //! Vector storage object
 
 use crate::avm2::activation::Activation;
+use crate::avm2::error::{make_error_1125, make_reference_error, Error, ReferenceErrorCode};
 use crate::avm2::object::script_object::ScriptObjectData;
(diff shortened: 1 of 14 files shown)


core/src/avm2/object/vector_object.rs
```diff
@@ -102,14 +124,76 @@ impl<'gc> TObject<'gc> for VectorObject<'gc> {
...
+                    }
                 }
             }
         }
 
-        self.base().get_property_local(name, activation)
+        // This *would* just be a call to ScriptObject::get_property_local, but
+        // Vector classes specifically pretend to be sealed when the prototype
+        // lookup fails, so we have to copy all the logic to here
+        if !name.contains_public_namespace() {
```
adrian17: The duplication could be moved to `ScriptObject`, similar to how in FP it's partially duplicated between universal `getAtomProperty` which calls `getAtomPropertyFromProtoChain` for non-dynamic.

adrian17 (approved):

Issue #20580, opened 2025-06-08
framescript execution where there shouldn't be


nivkner:
### Describe the bug

as mentioned in #19941, for the test case below, there is a mismatch between the scripts ruffle executes and those flash does:

```
Diff < left / right > :
 Main Constructor
 Spawn Constructor
 Spawn addFrameScript
 Main addFrameScript
 Main frame1
 Spawn frame1
 Container Constructor
 Container addFrameScript
 Container frame1
 Main frame2
 Spawn frame2
 Container frame2
 Main frame1
<Spawn frame1
 Container frame1
 Main frame2
<Spawn frame2
 Container frame2
 Main frame1
<Spawn frame1
 Container frame1
 Main frame2
 Main stopped
<Spawn frame2
<Spawn stopped
 Container frame2
 Container stopped
```

### Expected behavior

the framescripts of `Spawn` don't run after the first cycle of frames.

could be a flash bug, but if ruffle aims to be bug-compatible it is still a good idea to fix this.

### Content Location

[test.zip](https://github.com/user-attachments/files/20646349/test2.zip)

### Affected platform

Desktop app

### Operating system

fedora silverblue 40

### Additional information

tested against 69cd5dcc

danielhjacobs:
It may be a good idea to open a PR with this test as a known_failure.

PR #19941, opened 2025-03-28
core: Run scripts registered at frame script phase in a seperate pass

nivkner:
fixes #10443, and fixes #14629 for the version that does not include buttons.
when buttons are added to the dynamically constructed object, that is a separate issue of the frame script of the child happening *before the parent's constructor finishes*

im not sure i nailed down all the subtleties regarding the interaction between the cleanup pass and goto,
but the PR passes all the current tests on my machine, and the tests i made,
so i believe it is ready for review as is.

Issue #12443, opened 2023-07-30
Button framescript handling is incorrect


Aaron1011:
When working on improving SimpleButton support, I ended up with the following output from Flash Player in a a test case:
https://github.com/Aaron1011/ruffle/blob/ca00cf347877014679262c4b591c392938cad539/tests/tests/swfs/avm2/simplebutton_childevents/output.txt#L31C1-L34C28

```
//Down Button Shape frame 1
//MainTimeline frame 1
//Up Button Shape frame 1
//Over Button Shape frame 1
```

The framescript for the ancestor `MainTimeline` runs *between* framescripts for button states. All of the button states have already been constructed, so this isn't due to `addFrameScript` not having run yet.

The only way that this seems possible is if Flash Player is adding framescripts to some kind of hash set internally (with buttons being weird as usual).

nivkner:
adding this test case from #14629, since it seems to be more specific to buttons now: [test.zip](https://github.com/user-attachments/files/20646337/test1.zip)

```
Diff < left / right > :
 Main Constructor
 Main addFrameScript
 Main frame1
 Main frame2
 MyContainer Constructor
 MyChild Constructor
 MyChild addFrameScript
 MyChild Constructor
 MyChild addFrameScript
<MyContainer addFrameScript
 MyChild frame1
 MyChild frame1
<MyChild frame1
<MyChild frame1
<MyContainer frame1
>MyContainer addFrameScript
 MyChild frame2
 MyChild frame2
```

tested on 69cd5dcc

PR #20589, opened 2025-06-09
core: Add known_failure tests for frame script handling

nivkner:
as a followup on #19941, these are the cases of frame script execution which were not fixed,
and so are added as `known_failure` to alert the devs when the tests pass, as suggested in #20580.

in addition to `missing_frame_scripts` and `frame_script_button_order` which i identified in the PR,
i also added the failing test of @Aaron1011 mentioned in #12443.
while it seems connected to `frame_script_button_order`, it was added separately in case one is fixed, but not the other.


Dinnerbone (approved):
Thank you! New tests are always super valuable :D

Issue #14068, opened 2023-11-19
Error on https://game179258.konggames.com/games/randomnine/cardinal-quest-2/frame/50203974-0b2f-46ab-aa11-7f68ce07dbcf/


KyleChongHengKai:
# Error Info
Error name: Error
Error message: panicked at render/naga-pixelbender/src/lib.rs:1241:29:
Unimplemented opcode Step
Error stack:
```
Error: panicked at render/naga-pixelbender/src/lib.rs:1241:29:
Unimplemented opcode Step
    at H.n.wbg.__wbg_new_ab87fd305ed9004b (https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.9319559a195fae019006.js:1:56255)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::hb1c47d692dbcedf6 (wasm://wasm/0336f48a:wasm-function[9803]:0x9b0c0a)
    at std::panicking::rust_panic_with_hook::hc20eadded6bfe687 (wasm://wasm/0336f48a:wasm-function[6278]:0x89aea5)
    at core::panicking::panic_fmt::h7a368385936888dc (wasm://wasm/0336f48a:wasm-function[9603]:0x9aba14)
    at <ruffle_render_wgpu::backend::WgpuRenderBackend<T> as ruffle_render::backend::RenderBackend>::compile_pixelbender_shader::h9ef81f25215a56a4 (wasm://wasm/0336f48a:wasm-function[7928]:0x920856)
    at ruffle_core::avm2::globals::flash::display::shader_data::init::hf8d7d968a64019e5 (wasm://wasm/0336f48a:wasm-function[996]:0x377bda)
    at ruffle_core::avm2::function::Executable::exec::h031e8aa5585ebf44 (wasm://wasm/0336f48a:wasm-function[871]:0x2ffc95)
    at ruffle_core::avm2::object::TObject::call_property::h1a856e8c3ef51de1 (wasm://wasm/0336f48a:wasm-function[1263]:0x41e36d)
    at <ruffle_core::avm2::object::Object as ruffle_core::avm2::object::TObject>::call_property::h54f2da80b95b334b (wasm://wasm/0336f48a:wasm-function[2810]:0x67e9eb)
    at ruffle_core::avm2::activation::Activation::run_actions::h2dcd308c4e56efcb (wasm://wasm/0336f48a:wasm-function[568]:0x3f842)
```
AVM2 stack:
```
[...]

PR #20604, opened 2025-06-11
Implement PixelBender 'Step' opcode

Quasimondo:
This change implements the 'Step' opcode in the PixelBender shader infrastructure.

The 'Step' opcode behaves like the GLSL `step(edge, x)` function, returning 0.0 if x < edge and 1.0 if x >= edge, component-wise for vectors.

The implementation includes:
- Parsing logic for the 'Step' opcode in `render/src/pixel_bender.rs`.
- Naga IR generation for the 'Step' opcode in `render/naga-pixelbender/src/lib.rs`.
- A new test case `step_opcode_test` that defines a shader using the 'Step' opcode and verifies its compilation to Naga IR.

This addresses the issue of "Unimplemented opcode Step" when encountering shaders that utilize this instruction.

---

I am not sure how you feel about AI-coded improvements to the codebase. I used Google Jules to implement the missing "step" opcode and it does fix the problem allowing me to run an old PixelBender swf that did not work so far.

### DIFF
--- a/render/naga-pixelbender/src/lib.rs
+++ b/render/naga-pixelbender/src/lib.rs
@@ -1220,6 +1220,31 @@ impl ShaderBuilder<'_> {
                             arg2: None,
                             arg3: None,
                         }),
+                        Opcode::Step => {
+                            // dst is used as the 'edge' value in 'step(edge, x)'
+                            // src is used as the 'x' value
+                            let edge_val = self.load_src_register(&dst)?;
+                            let x_val = src; // src is already loaded from src_reg
+
+                            // Result is 1.0 if x >= edge, else 0.0
+                            // So we want `x_val < edge_val` to select 0.0, and `x_val >= edge_val` to select 1.0
+                            let comparison = self.evaluate_expr(Expression::Binary {
+                                op: BinaryOperator::Less, // x < edge
+                                left: x_val,
+                                right: edge_val,
+                            });
+
+                            let one_vec4f = self.evaluate_expr(Expression::Splat {
+                                size: VectorSize::Quad,
+                                value: self.onef32,
+                            });
+
+                            self.evaluate_expr(Expression::Select {
+                                condition: comparison, // If x < edge is true
+                                accept: self.zerovec4f, // then 0.0
+                                reject: one_vec4f,      // else 1.0
+                            })
+                        }
                         _ => {
                             panic!("Unimplemented opcode {opcode:?}");
                         }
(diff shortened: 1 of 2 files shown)


Dinnerbone:
> me to run an old PixelBender swf that did not work so far.

Can you please share that swf? There's an actual step function we can emit, rather than trying to reimplement the behaviour from scratch - but a swf to test with would be extremely valuable.

evilpie:
We have some issues for the missing Step opcode: https://github.com/ruffle-rs/ruffle/issues?q=is%3Aissue%20state%3Aopen%20%22unimplemented%20opcode%20Step%22

adrian17:
I did implement `Step` for one of these (I think it was https://github.com/ruffle-rs/ruffle/issues/14068) in the past, with something like (reproduced from memory):

```rust
                        Opcode::Step => {
                            let right = self.load_src_register(&dst)?;
                            self.evaluate_expr(Expression::Math {
                                fun: MathFunction::Step,
                                arg: src,
                                arg1: Some(right),
                                arg2: None,
                                arg3: None,
                            })
                        }
```

But it didn't render correctly and I didn't know whether that meant the implementation was wrong in some way or there was a completely unrelated bug somewhere else; so I never made a PR.
I'm also curious as to what SWF the PR author is talking about, I'm very interested to test my old snippet on other SWFs.

As for this PR, unless there's a reason why plain `MathFunction::Step` wasn't used, unfortunately I don't think this code is usable. Especially the change in `render/src/pixel_bender.rs` makes no sense, as it doesn't seem any different from the catch-all case.

Quasimondo:
> As for this PR, unless there's a reason why plain `MathFunction::Step` wasn't used, unfortunately I don't think this code is usable. Especially the change in `render/src/pixel_bender.rs` makes no sense, as it doesn't seem any different from the catch-all case.

I am not sure why one would use MathFunction inside of a shader, but then again I haven't looked into the inner workings of it - my implementation follows the principles that the implementations of other PixelBender opcodes use and none of those uses MathFunction, so I assume there is a reason for it. I also do not understand the question why a PixelBender opcode is added to the pixel_bender.rs code - it is where all the other opcode implementations are so it seems like the logical place to put it.

Lastly - if there is a catch-all it does not work as intended since when trying to run an swf that contains the step opcode inside of a PixelBender shader a panic is triggered.

All I can say is that my swf did not work before and it does work with the fix.

adrian17:
Again, can you please send us the SWF you're talking about? (if you can't share it, that's okay, just let us know)

adrian17:
Meanwhile, I merged https://github.com/ruffle-rs/ruffle/pull/20640 , so this generated PR isn't needed anymore.
If you can check if your SWF now works with upstream Ruffle, that'd be great; if it doesn't, let us know and we'll look at it some more.

PR #20621, opened 2025-06-13
web: Make default quality null

### LINKED ISSUE #20610: On mobile devices, the default quality does not change to low.
### Describe the bug

The site uses self hosted.

In the new versions of Ruffle, on mobile devices, the default quality does not change to low, and the game is played in high quality.  Because of this, the device does not have enough resources and the game crashes with an error. 

When using the old Ruffle build (2023-09-14), there is no such problem,
and by default, the quality on the desktop is high, but on mobile it is low.

What needs to be fixed so that the new Ruffle works as on the old ones?

I hope you understand me.

Here is my embed:
<embed src="//site.swf"  quality="default" allowfullscreen="true"  width="720" height="540" align="middle">

Thanks

### Expected behavior

What needs to be fixed so that with the new Ruffle, by default, the quality on the desktop is high, and on mobile it is low.

### Content Location

Warning! Adult content
https://sexflashgame.org/valentines-day/

### Affected platform

Self-hosted version

### Operating system

Android

### Browser

_No response_

### Additional information

_No response_

danielhjacobs:
This should fix #20610, because the quality will not be explicit so the builder will change it to low on mobile devices.

### DIFF
--- a/web/packages/core/src/public/config/default.ts
+++ b/web/packages/core/src/public/config/default.ts
@@ -35,7 +35,7 @@ export const DEFAULT_CONFIG: Required<BaseLoadOptions> = {
     salign: "",
     fullScreenAspectRatio: "",
     forceAlign: false,
-    quality: "high",
+    quality: null,
     scale: "showAll",
     forceScale: false,
     frameRate: null,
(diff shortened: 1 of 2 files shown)


danielhjacobs:
https://github.com/ruffle-rs/ruffle/blob/master/web/packages/core/src/internal/builder.ts#L62-L67

Dinnerbone (approved):

PR #20627, opened 2025-06-15
 avm2: Initialize the loaderInfo.loaderURL more consistently 

### LINKED ISSUE #14217: new Loader().contentLoaderInfo.loaderURL returns file:// instead of whole path
In Flash `new Loader().contentLoaderInfo.loaderURL` returns the whole path of the SWF, in Ruffle it's just `file:///`.

The value probably comes from either `SwfMovie::empty` or `SwfMovie::empty_fake_compressed_len`.

evilpie:
Fixes #14217


Dinnerbone (approved):
❤️

PR #19405, opened 2025-02-01
render: Implement reposition in Z-axis

cookie-s:
Closes #8033 . See https://github.com/ruffle-rs/ruffle/issues/8033#issuecomment-2607237748 for the difference.

This PR implements `DisplayObject.z` property feature including its rendering with the default perspective projection.

### Known limitations
* The supported perspective projection is limited to the Flash Player's default one. See `render/wgpu/src/globals.rs` (`PerspectiveProjection::default()`).
  ```
  root.transform.perspectiveProjection.fieldOfView = 55
  root.transform.perspectiveProjection.projectionCenter = (x: stage.stageWidth / 2, y: stage.stageHeight / 2)
  ```
  doc: https://airsdk.dev/docs/development/display/working-in-three-dimensions/projecting-3d-objects-onto-a-2d-view
  * Change in `root.transform.perspectiveProjection` is not effective. (In fact, even its setter is not implemented in Ruffle.)
  * `perspectiveProjection` of any ancestors is not taken into account while [the doc](https://docs.ruffle.rs/en_US/FlashPlatform/reference/actionscript/3/flash/geom/PerspectiveProjection.html) says:
    > The specified projection matrix and transformation will then apply to all the display object's three-dimensional children.
* This PR doesn't implement non-translation transformations in 3D space. e.g. rotationX, scaleZ.

The implementation might not be the good one. Especially the `Transform` struct now has the new `tz` in addition to the existing `matrix` (2D). The `tz` property will be removed at the implementation of the full 3D transformation.

PR #20633, opened 2025-06-17
avm2: Move PerspectiveProjection computation to render crate

cookie-s:
Continued from https://github.com/ruffle-rs/ruffle/pull/19670
Continue to https://github.com/ruffle-rs/ruffle/pull/19405 ?

This PR refactors flash.geom.PerspectiveProjection computation logics from ruffle_core to ruffle_render.
We can use those later from inside ruffle_render.


Dinnerbone (approved):

cookie-s:
Thank you! will update the other PR this weekend too

PR #20639, opened 2025-06-17
core: Implement glyph fallback for rendering device fonts

kjarosh:
Flash Player uses multiple device fonts to render text in order to cover as many glyphs as possible. In order to do that, it uses glyph fallback—glyphs that are not present in the main font are rendered from fallback fonts.

In Ruffle, this feature is implemented on all platforms, however it'll work only on Linux as of now, as other platforms do not provide more than one device font per resolution (font sorting is not implemented).

It's not yet perfect, I can sometimes see some characters missing (not sure why, might be something wrong with font sorting), but the fallback generally works. It's very difficult to test, as I haven't managed yet to force Flash Player to use my custom fallback fonts.

The main idea behind the implementation is to introduce a `FontLike` trait that can be implemented by both fonts and font sets. Font sets then can just search across fallback fonts when looking for a glyph. Font sets are GC'd.


core/src/font.rs
```diff
@@ -698,7 +733,9 @@ impl<'gc> Font<'gc> {
         let mut x = Twips::ZERO;
         while let Some((pos, c)) = char_indices.next() {
             let c = c.unwrap_or(char::REPLACEMENT_CHARACTER);
-            if let Some(glyph) = self.get_glyph_for_char(c) {
+            if let Some(render_data) = self.get_glyph_render_data(c) {
```
moulins, line 751: This looks wrong: `self` is a `Font` here, not a `FontSet`, so there's no possibility to do per-character fallback.

core/src/font.rs
```diff
@@ -1210,9 +1252,98 @@ impl Default for TextRenderSettings {
...
+}
+
+impl<'gc> FontSet<'gc> {
+    /// Creates a font set from a sorted list of fonts.
+    ///
+    /// The first font is the main font, the rest are fallbacks.
+    ///
+    /// Returns None when the list is empty.
+    pub fn from_fonts(mc: &Mutation<'gc>, fonts: Vec<Font<'gc>>) -> Option<Self> {
+        let (&main_font, fallback_fonts) = fonts.split_first()?;
```
moulins, line 1301: Nit: this could be simplified to
```rs
let main_font = fonts.remove(0)?;
Some(Self(Gc::new(
    mc,
    FontSetData {
        main_font,
        fallback_fonts: fonts,
    },
)))
```

core/src/font.rs
kjarosh: I don't think I follow... `self` is a `FontLike` isn't it?

core/src/font.rs
kjarosh: Hmmm... I'm not sure if it's a simplification. It requires `fonts` to be mutable and doesn't have positive performance implications, IMO it's better to operate on non-mutable data

moulins: (sorry for the long delay)
Every call to `from_fonts` passes a reference to a temporary Vec which is discarded later, so it would be slightly more efficient to take the Vec by-value and shift the contents in-place instead of reallocating a new buffer. (or if you don't want to do that, you should at least change the signature to take a `&[Font<'gc>]`, `&Vec<_>` is basically never useful)

core/src/font.rs
moulins: Let's say you have a `TextSpan` containing both arabic text and emojis: you need the arabic `char`s and the emoji `char`s to use glyphs from two different fonts in the font set (as no single font provides both). How does your code handle that?

kjarosh: The fallback logic is implemented here:

https://github.com/ruffle-rs/ruffle/blob/452e0298276aef114e9c85d9908d0cf75b50280e/core/src/font.rs#L1303-L1313

core/src/font.rs
kjarosh: You're right, it will save one allocation. I decided to change it to `&[Font<'gc>]` just because I plan to not do a vec clone for each font sort in the future.

core/src/font.rs
moulins: But this logic is only present on `FontSet`, and we don't have a `FontSet` to use inside `Font::evaluate`?

kjarosh: As I said, the method `evaluate` is defined on a trait. `Font` doesn't have any fallbacks, so `Font::evaluate` will use glyphs only from the font. But `FontSet::evaluate` will use glyph fallbacks.

core/src/font.rs
```diff
@@ -1229,9 +1271,98 @@ impl Default for TextRenderSettings {
...
+
+        for fallback_font in &self.0.fallback_fonts {
+            if let Some(glyph) = fallback_font.get_glyph_for_char(c) {
+                return Some(GlyphRenderData::new(glyph, *fallback_font));
+            }
+        }
+
+        None
+    }
+
```
Lord-McSweeney, line 1337: Could you add a comment here explaining FP's behavior when the main font's info (kerning, baseline, etc) doesn't match a fallback font's info? Or just a TODO if you haven't tested yet?

kjarosh: Added a TODO, thanks!

PR #20711, opened 2025-06-18
ci: Workaround issue by pinning cargo-nextest version

danielhjacobs:


torokati44:
The upstream fix was almost this quick, but thanks anyway! 😊

Issue #10621, opened 2023-04-09
Super Fireman game freezing when Play Game pressed


freeworldgroup:
### Describe the bug

Game loads normally and gets into the main menu. When the "Play Game" button is pressed, the whole game freezes and nothing more can be done. 

Game can be found here:
http://www.freeworldgroup.com/games9/gameindex/superfireman.htm

### Expected behavior

The game should start after pressing "Play Game"

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Chrome

Lord-McSweeney:
It's probable that these errors are relevant:
```
Error running definition tag: DefineBits, got Couldn't register bitmap
Error running definition tag: DefineShape2, got Couldn't read SWF
Error running definition tag: DefineText2, got Couldn't read SWF
Error running definition tag: DefineFont2, got Couldn't read SWF
Error running definition tag: DefineEditText, got Couldn't read SWF
Error running definition tag: DefineButton, got Couldn't read SWF
Error running definition tag: DefineBitsJpeg4, got Couldn't read SWF
Error running definition tag: DefineBitsJpeg3, got Couldn't read SWF
Error running definition tag: DefineButton2, got Couldn't read SWF
Error running definition tag: DefineEditText, got Couldn't read SWF
Error running definition tag: DefineBitsLossless2, got Couldn't read SWF
Error running definition tag: DefineText2, got Couldn't read SWF
Error running definition tag: DefineShape2, got Couldn't read SWF
```
...
```
Error running definition tag: DefineButton2, got Couldn't read SWF
Error running definition tag: DefineMorphShape2, got Couldn't read SWF
Error running definition tag: DefineText2, got Couldn't read SWF
Error running definition tag: DefineShape2, got Couldn't read SWF
Error running definition tag: DefineSprite, got Couldn't read SWF
Error running definition tag: DefineMorphShape, got Couldn't read SWF
Error running definition tag: DefineBitsLossless, got Couldn't read SWF
Error running definition tag: DefineEditText, got Couldn't read SWF
Error running definition tag: DefineShape2, got Couldn't read SWF
Error running definition tag: DefineSprite, got Couldn't read SWF
```

Still, that doesn't explain this:
`Uncaught RuntimeError: memory access out of bounds`

freeworldgroup:
hope this can be addressed at some point!

Lord-McSweeney:
[...]

PR #20749, opened 2025-06-18
core: Fix placing objects at an occupied depth

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/2536
* Fixes https://github.com/ruffle-rs/ruffle/issues/10621

Do not instantiate an object which is being placed at a depth where another object is already placed.


Dinnerbone (approved):
Neat!

PR #20757, opened 2025-06-18
core: Add avm_warning for simulating traced FP warnings

kjarosh:
Simulating traced Flash Player warnings improves our tests, because we can assert on certain code paths taken in our implementations.

### DIFF
--- a/core/src/backend/log.rs
+++ b/core/src/backend/log.rs
@@ -1,5 +1,7 @@
 pub trait LogBackend {
     fn avm_trace(&self, message: &str);
+
+    fn avm_warning(&self, message: &str);
 }
 
 /// Logging backend that just reroutes traces to the log crate
@@ -14,6 +16,10 @@ impl NullLogBackend {
 impl LogBackend for NullLogBackend {
     fn avm_trace(&self, message: &str) {
         tracing::info!(target: "avm_trace", "{}", message);
+    }
+
+    fn avm_warning(&self, message: &str) {
+        tracing::info!(target: "avm_warning", "{}", message);
     }
 }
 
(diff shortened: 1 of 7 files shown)


Dinnerbone:
Hmmm! I'm not opposed to the idea but I have a few thoughts:

1. Are the warnings the same across different FP versions, or will this complicate trying tests in FP?
2. There's _many_ warnings, I believe, that we've so far just ignored and not copied into tests/ruffle. We may start breaking tests as we implement them, which isn't a terribly bad thing but it is a bit of churn to be aware of
3. I don't know if we should call this "avm warning", as it's usually a generic flash thing, not specific to avm (flash can spit out many warnings 😅)

kjarosh:
> Are the warnings the same across different FP versions, or will this complicate trying tests in FP?

I haven't personally noticed that warnings differ specifically between versions. It's totally possible, but it's also easy to verify and I'd argue that those warnings give us more insight into the implementation anyway.

> There's _many_ warnings, I believe, that we've so far just ignored and not copied into tests/ruffle. We may start breaking tests as we implement them, which isn't a terribly bad thing but it is a bit of churn to be aware of

Yeah, if a new warning is being implemented, we have to revisit failing tests and update outputs. That's an easy task though, it's enough to just rerun FP. Note that it applies only when implementing a new warning, so it doesn't negatively influence fixes and feature impls.

> I don't know if we should call this "avm warning", as it's usually a generic flash thing, not specific to avm (flash can spit out many warnings 😅)

Do you have any suggestions?

Dinnerbone:
> Do you have any suggestions?

Nope :D Just thinking aloud really, sorry!

PR #20770, opened 2025-06-18
web: Don't show CORS error message for the Internet Archive

danielhjacobs:
Not sure how I feel about this, on the one hand it's simple and helps to stop the issues saying "why does the Internet Archive have this CORS error", but on the other hand is a single hard-coded URL. It's tough to do things any other way because the Internet Archive lies about the URL it's actually fetching due to its custom fetch code that redirects all requests to itself.

### DIFF
--- a/web/packages/core/src/internal/ui/panic.tsx
+++ b/web/packages/core/src/internal/ui/panic.tsx
@@ -185,6 +185,8 @@ function createPanicError(error: Error | null): {
 
         if (
             window.location.origin === error.swfUrl?.origin ||
+            // The Internet Archive fetches from itself despite lies about the swfUrl
+            window.location.origin === "https://web.archive.org" ||
             // The extension's internal player page is not restricted by CORS
             window.location.protocol.includes("extension")
         ) {


n0samu:
I remember I had a different idea for how to accomplish this without hardcoding the URL - let me try to dig it up in the next few days

danielhjacobs:
One way I thought of just now would be:
```js
window.location.origin === new URL((await fetch(error.swfUrl)).url).origin
```

On https://web.archive.org/web/20250119102449/https://www.dnai.org/, `window.location.origin === new URL((await fetch("not_a_real.swf")).url).origin`

Not sure I'd want to re-do a fetch just for this though.

danielhjacobs:
Working through the logic straight through the Rust code, I guess that is this:

<https://github.com/ruffle-rs/ruffle/blob/master/core/src/tag_utils.rs#L58>

I guess a more accurate swfUrl would be that.

danielhjacobs:
Last thought wasn't quite right, but it can be done by adding a new parameter for the real SWF URL that we attempt to fetch to this function: https://github.com/ruffle-rs/ruffle/blob/dc22b9916f2003b320e692ff8581f0c084b89f94/core/src/backend/ui.rs#L85

danielhjacobs:
Here, that's `error.url`:

https://github.com/ruffle-rs/ruffle/blob/dc22b9916f2003b320e692ff8581f0c084b89f94/core/src/loader.rs#L1067

Here, it's `url`:

https://github.com/ruffle-rs/ruffle/blob/dc22b9916f2003b320e692ff8581f0c084b89f94/core/src/loader.rs#L1076

Here, it's `spoofed_or_swf_url`:

https://github.com/ruffle-rs/ruffle/blob/dc22b9916f2003b320e692ff8581f0c084b89f94/core/src/loader.rs#L1098

danielhjacobs:
This should no longer be needed.

Issue #20583, opened 2025-06-09
Sift Heads Stick War Crashes in the middle


zohebnsr:
### Describe what you were doing

[sift-heads-stick-war.swf.zip](https://github.com/user-attachments/files/20652214/sift-heads-stick-war.swf.zip)

I have attached the game file. 
When exactly it crashes? - Let the game load and once you click the start button, it crashes before aactual game starts. It plays the animation fine at the start before you are in start menu. 

### What does the crash message say?

panicked at core/src/pixel_bender.rs:71:45:
called `Option::unwrap()` on a `None` value

Stack:

Error
    at push.655.Re.n.wbg.__wbg_new_8a6f238a6ece86ea (https://www.papasgaming.com/ruffle/core.ruffle.e7a8e75c10c49320b4eb.js:1:55610)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once{{vtable.shim}}::h907ab91aacbc8367 (wasm://wasm/ruffle_web.wasm-0314f8b6:wasm-function[9820]:0x913f45)
    at ruffle_web.wasm.std::panicking::rust_panic_with_hook::h9ea01914b5d92659 (wasm://wasm/ruffle_web.wasm-0314f8b6:wasm-function[6895]:0x873682)
    at ruffle_web.wasm.core::panicking::panic_fmt::hb91badace6ee8287 (wasm://wasm/ruffle_web.wasm-0314f8b6:wasm-function[10653]:0x92d308)
    at ruffle_web.wasm.core::panicking::panic::h220e5a67f5e3e06f (wasm://wasm/ruffle_web.wasm-0314f8b6:wasm-function[9939]:0x916b32)
    at ruffle_web.wasm.core::option::unwrap_failed::hbdf3643a84d1fb32 (wasm://wasm/ruffle_web.wasm-0314f8b6:wasm-function[12056]:0x95052d)
    at ruffle_web.wasm.core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &mut F>::call_once::h11d68187300f40b0 (wasm://wasm/ruffle_web.wasm-0314f8b6:wasm-function[1269]:0x34f530)
    at ruffle_web.wasm.ruffle_core::avm2::globals::flash::display::shader_job::get_shader_args::h6ac26601056671ee (wasm://wasm/ruffle_web.wasm-0314f8b6:wasm-function[3939]:0x703ee1)
[...]

Issue #15176, opened 2024-02-09
Crash on Main.swf


p0008874:
### Describe what you were doing

Play the demo
https://www.pouet.net/prod.php?which=55593

### What does the crash message say?

panicked at core\src\pixel_bender.rs:63:45:
called `Option::unwrap()` on a `None` value

### Ruffle Version

0.1.0-nightly (50a8815a5012b3ed158fa3b5b0fa69bea6fac28a 2024-02-09)

### Affected platform

Desktop app

### Operating system

Windows 10.0.19045 (Windows 10 Pro) [64-bit]

### Additional information

Filename: Main.swf

### AVM2 Callstack
```
	at flash.display::ShaderJob/flash::display::ShaderJob::start()
	at project.away3d::SpikedSphere/update()
	at project.demo.modifier::ModifierSpikedSphere/render()
	at evoTinyEngine::TinyEngine/render()
```

### Render Info
Renderer: wgpu
Adapter Backend: Vulkan
Adapter Name: "NVIDIA GeForce GTX 650"
Adapter Device Type: DiscreteGpu
Adapter Driver Name: "NVIDIA"
Adapter Driver Info: "474.64"
Enabled features: Features(FLOAT32_FILTERABLE | TEXTURE_COMPRESSION_BC | TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES | SHADER_UNUSED_VERTEX_OUTPUT)
[...]

PR #20781, opened 2025-06-19
pixelbender: Handle malformed arguments properly

kjarosh:
Malformed arguments should not crash Ruffle. Flash Player either recovers when the number of elements in an array is too small, or throws an exception when the type is invalid.

* Fixes https://github.com/ruffle-rs/ruffle/issues/15176
* Fixes https://github.com/ruffle-rs/ruffle/issues/20573
* Fixes https://github.com/ruffle-rs/ruffle/issues/20583


core/src/pixel_bender.rs
```diff
@@ -48,66 +51,90 @@ impl PixelBenderTypeExt for PixelBenderType {
...
+                    fn next_val<'gc>(
+                        activation: &mut Activation<'_, 'gc>,
+                        vals: &mut impl Iterator<Item = Option<Value<'gc>>>,
+                    ) -> Result<f64, Error<'gc>> {
+                        let Some(Some(val)) = vals.next() else {
+                            return Ok(0.0);
+                        };
+
+                        match val {
+                            Value::Number(f) => Ok(f),
```
adrian17: Instead of manually matching here, how about adding try_as_f64 to Value (like as_f64, but returns Option)?

kjarosh: Done, thanks

core/src/pixel_bender.rs
```diff
@@ -48,66 +51,88 @@ impl PixelBenderTypeExt for PixelBenderType {
...
-                            val.expect("Array with hole")
-                                .coerce_to_i32(activation)
-                                .unwrap() as i16
-                        });
                         match kind {
-                            PixelBenderTypeOpcode::TInt => {
-                                Ok(PixelBenderType::TInt(vals.next().unwrap()))
-                            }
+                            PixelBenderTypeOpcode::TInt => Ok(PixelBenderType::TInt(
+                                //
```
Lord-McSweeney, line 119: Leftover empty comment?

kjarosh: I left it intentionally to improve formatting and keep it in line with code below

torokati44: `// <this comment is intentionally left blank>` :sweat_smile:

PR #20792, opened 2025-06-20
web: Denote actual fetched URL when root movie download fails

danielhjacobs:
Unfortunately, it's another new parameter to display_root_movie_download_failed_message (see https://github.com/ruffle-rs/ruffle/pull/14715), but I think it should work as another way to accomplish https://github.com/ruffle-rs/ruffle/pull/20770 without a hard-coded URL.


core/src/loader.rs
```diff
@@ -1089,13 +1090,17 @@ impl<'gc> Loader<'gc> {
...
             let mut movie =
                 SwfMovie::from_data(&body, spoofed_or_swf_url, None).inspect_err(|_error| {
                     player
                         .lock()
                         .unwrap()
                         .ui()
-                        .display_root_movie_download_failed_message(true);
+                        .display_root_movie_download_failed_message(
+                            true,
+                            cloned_spoofed_or_swf_url,
```
n0samu: cleaner to just put `spoofed_or_swf_url.clone()` here rather than creating a new variable, no?

danielhjacobs: Not familiar enough with Rust to know if it's needed here, but the earlier clone needed to be outside the closure because the closure took ownership of the variable. Don't know if that occurs here too.

danielhjacobs: I think you're right here though, always hard to keep track of this stuff in Rust when I'm more familiar with languages without this.

danielhjacobs: Nah, it needs to be this I think:
```
   Compiling ruffle_core v0.1.0 (C:\Users\danie\Documents\danielhjacobs\ruffle\core)
error[E0382]: borrow of moved value: `spoofed_or_swf_url`
    --> core\src\loader.rs:1094:82
     |
1086 |             let spoofed_or_swf_url = player
     |                 ------------------ move occurs because `spoofed_or_swf_url` has type `std::string::String`, which does not implement the `Copy` trait
...
1094 |                 SwfMovie::from_data(&body, spoofed_or_swf_url, None).inspect_err(|_error| {
     |                                            ------------------ value moved here   ^^^^^^^^ value borrowed here after move
...
1101 |                             spoofed_or_swf_url.clone(),
     |                             ------------------ borrow occurs due to use in closure
     |
note: consider changing this parameter type in method `from_data` to borrow instead if owning the value isn't necessary
    --> core\src\tag_utils.rs:197:14
     |
195  |     pub fn from_data(
     |            --------- in this method
196  |         swf_data: &[u8],
197  |         url: String,
```

n0samu (approved):

core/src/loader.rs
kjarosh: You'd have to clone `spoofed_or_swf_url` the first time you use it, so `SwfMovie::from_data(&body, spoofed_or_swf_url.clone(), None)`, and then the second use can be moved.

danielhjacobs: Thanks, I could do this for the approach this PR took before, but I changed the approach instead. See the new code.

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -2070,9 +2073,12 @@ export class InnerPlayer {
         ) {
             this.addOpenInNewTabMessage(openInNewTab, this.swfUrl);
         } else {
+            const fetchStatusNotOk = fetchError.includes(
+                "HTTP Status is not OK:",
+            );
```
danielhjacobs, line 2078: https://github.com/ruffle-rs/ruffle/blob/fb9bddd7c8affd61594adfa91d94c9347d70adf0/core/src/loader.rs#L182-L183

danielhjacobs:
PR has been changed (commits should be squashed if this approach is taken, they are separate so I can revert to the old approach if preferred).

Now, if the `fetch` response has a status code, we display the `error-swf-fetch` message, even if the request was cross-origin.

web/packages/core/src/internal/ui/panic.tsx
```diff
@@ -185,6 +185,7 @@ function createPanicError(error: Error | null): {
 
         if (
             window.location.origin === error.swfUrl?.origin ||
+            error.statusNotOk ||
             // The extension's internal player page is not restricted by CORS
```
danielhjacobs, line 189: The other checks can probably be removed from here, but not certain.

n0samu (approved):
LGTM, thank you!!

Dinnerbone (approved):

PR #20797, opened 2025-06-20
ci: Add EditorConfig lint

kjarosh:
This PR fixes remaining EC violations and adds a linter for it.


torokati44:
Since we already have a "Lints" workflow for cargo-fmt and cargo-clippy, perhaps this workflow could be named so that it's explicitly about EditorConfig. Otherwise, it's strange that the cargo lints are not in this one (and I dont think they should be).

kjarosh:
What about "Other Lints" or "General Lints"? That workflow can include other lints in the future too.

PR #20800, opened 2025-06-20
avm1: Implement warning about undeclared variable

kjarosh:
This PR implements the `Warning: Reference to undeclared variable, '...'` warning.

It's a draft, because (1) we seem to diverge a bit from FP's behavior (see below), and (2) some tests produce a vastly different output when rerun than committed.

Examples of current discrepancies:

1. `getDepth` is found declared in Ruffle in `avm1/swf5_to_6_cross_call`, but shouldn't be
2. `/:ghi` should be found declared in `avm1/define_local_with_paths`, but it's not
3. in `avm1/mcl_loadclip_properties`, a variable resolution is done too early in Ruffle


moulins:
As a stopgap measure, we could add a new `test.toml` config option to ignore warnings, and set it to `true` for tests with discrepancies.

Issue #16212, opened 2024-05-04
Error on https://game267424.konggames.com/games/PlasticContainer/youtube-lets-player-simulator-2016/frame/165b1d68-95cf-412b-9744-7e0f6bc4795a/


JoeyWheelez:
# Error Info
Error name: Error
Error message: panicked at render/src/pixel_bender.rs:467:17:
Unexpected param type 13
Error stack:
```
Error: panicked at render/src/pixel_bender.rs:467:17:
Unexpected param type 13
    at J.n.wbg.__wbg_new_28c511d9baebfa89 (https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.1eae354ae80ec1de6482.js:1:74025)
    at ruffle_web.wasm.js_sys::Error::new::h2234a378a3d95323 (wasm://wasm/ruffle_web.wasm-03a269ba:wasm-function[14096]:0xb144ea)
    at ruffle_web.wasm.core::ops::function::FnOnce::call_once{{vtable.shim}}::hd3374b50dcd4a09d (wasm://wasm/ruffle_web.wasm-03a269ba:wasm-function[11628]:0xabfe8d)
    at ruffle_web.wasm.std::panicking::rust_panic_with_hook::h32c80a64fe4de396 (wasm://wasm/ruffle_web.wasm-03a269ba:wasm-function[7062]:0x97db45)
    at ruffle_web.wasm.core::panicking::panic_fmt::hc7427f902a13f1a9 (wasm://wasm/ruffle_web.wasm-03a269ba:wasm-function[11327]:0xab8244)
    at ruffle_web.wasm.ruffle_core::avm2::globals::flash::display::shader_data::init::h2eb014e16c44cde2 (wasm://wasm/ruffle_web.wasm-03a269ba:wasm-function[1594]:0x4c54b7)
    at ruffle_web.wasm.ruffle_core::avm2::function::Executable::exec::hb9e49ab2c0452c6c (wasm://wasm/ruffle_web.wasm-03a269ba:wasm-function[1044]:0x310ea4)
    at ruffle_web.wasm.ruffle_core::avm2::object::TObject::call_method::hac75ab3bde134a8a (wasm://wasm/ruffle_web.wasm-03a269ba:wasm-function[2851]:0x6bf69c)
[...]

PR #20808, opened 2025-06-22
pixelbender: Fix handling AVM2 parameters

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/16212
* Fixes https://github.com/ruffle-rs/ruffle/issues/19064

This PR implements boolean parameters, some conversions, and fully fixes panics and errors when parsing malformed Pixel Bender parameters.


core/src/pixel_bender.rs
```diff
@@ -34,116 +34,126 @@ impl PixelBenderTypeExt for PixelBenderType {
...
-                    fn next_val<'gc>(
-                        activation: &mut Activation<'_, 'gc>,
-                        vals: &mut impl Iterator<Item = Option<Value<'gc>>>,
-                    ) -> Result<f64, Error<'gc>> {
-                        let Some(Some(val)) = vals.next() else {
-                            return Ok(0.0);
-                        };
+        let array_storage = match value {
+            Value::Object(ref o) => o.as_array_storage(),
+            _ => None,
```
Lord-McSweeney: Since this should only be `ArrayObject`, this `match` could explicitly panic on anything else:
```
match value {
    Value::Object(Object::ArrayObject(array)) => Some(array.array_storage()),
    Value::Null => None,
    _ => unreachable!("..."),
}
```

Lord-McSweeney (approved):

core/src/pixel_bender.rs
kjarosh: Done

PR #20809, opened 2025-06-23
tests: Fix assertion comparison message order

cookie-s:
The left value is from Ruffle, and the right value is from Flash Player.
The error message should follow the same order? Isn't it confusing?

```
fn assert_text_matches(ruffle: &str, flash: &str) -> Result<()> {
```

```
        let left_pretty = PrettyString(ruffle);
        let right_pretty = PrettyString(flash);
```


torokati44:
(nit: spelling error in PR title and commit message: "comparision" -> "comparison")

cookie-s:
(oh, I didn't notice it! thank you for sharing, fixed now.)

evilpie:
Looks good to me. Can you rebase?

cookie-s:
Thank you. Rebased just now.

Issue #11727, opened 2023-06-25
The fox falls through the floor immediately when starting


olux997:
### Describe the bug

the fox falls through the floor immediately when starting....

<img width="350" alt="power_fox__1_done" src="https://github.com/ruffle-rs/ruffle/assets/11493572/1c129634-60a9-4141-9704-1956e0474774">

web site:
https://www.newgrounds.com/portal/view/212188

swf url:
https://uploads.ungrounded.net/212000/212188_powerfox.swf?1105158874

### Expected behavior

Hope my fox comes alive for a good game!

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

Google Chrome 64bit,114.0.5735.134

### Additional information

Hope my fox comes alive for a good game!

===============================================
error report:

# Player Info
Allows script access: false
Renderer: wgpu
Adapter Backend: Gl
Adapter Name: "ANGLE (Intel, Intel(R) HD Graphics 630 Direct3D11 vs_5_0 ps_5_0, D3D11)"
Adapter Device Type: IntegratedGpu
Adapter Driver Name: ""
Adapter Driver Info: ""
Enabled features: Features(TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES | PUSH_CONSTANTS)
Available features: Features(TEXTURE_COMPRESSION_BC | CLEAR_TEXTURE | MULTIVIEW)
[...]

Issue #11454, opened 2023-06-09
Player falls through the floor in Powerfox V.3


grigorig:
### Describe the bug

See: https://www.newgrounds.com/portal/view/215318

Just start the game and try to play. In a fraction of a second, the player character (the fox) falls through the floor and the game cannot be played properly.

Note: Newgrounds seems to use a Ruffle snapshot from today (2023-06-09).

### Expected behavior

Game character should move horizontally.

### Affected platform

Self-hosted version

### Operating system

Linux

### Browser

Firefox 113, Chromium 111

Toad06:
Thanks for the report.

This is unfortunately a regression introduced in commit c17da6e91a9fbd7f9fb58854b95b6067b3793bb2. The changes made to the `winding_number_line` function seem to be the cause of the issue.

The following issues are also affected: #8303, #8420.

Issue #8420, opened 2022-10-31
[BugBug in Sky Tower] Jumping above a certain height does not work.


jmousy:
### Describe the bug

In **BugBug in Sky Tower**, if you try to jump with the up arrow key on a platform of a certain height, the jump does not work.

![image](https://user-images.githubusercontent.com/75052171/199015967-b9fe9854-e615-4060-aa16-83defbbc9247.png)

Because of this, you can no longer proceed upwards from the red area in the screenshot.

Reproduced in Ruffle 2022-10-04 ~ 2022-10-30 Nightly.

[bugbug-in-sky-tower.zip](https://github.com/ruffle-rs/ruffle/files/9901322/bugbug-in-sky-tower.zip)

### Expected behavior

You should be able to jump on all platforms unless there are obstacles.

### Affected platform

Self-hosted version

### Operating system

Windows 10 22H2

### Browser

Google Chrome 107

Jmousy:
However, i have confirmed that the problem does not occur in Nightly on 2022-05-01. (I haven't figured out the exact date the problem started happening.)

n0samu:
Bisected yet again to commit c17da6e91a9fbd7f9fb58854b95b6067b3793bb2 / #7684 😭

Toad06:
Reverting the few changes related to `fn winding_number_line` in c17da6e91a9fbd7f9fb58854b95b6067b3793bb2 prevents the issue from happening (same goes for #8303). Tests still pass, unfortunately this does cause a regression in #5481.

Issue #8303, opened 2022-10-16
[Castle Cat 1-4] The character breaks through the floor and falls.


jmousy:
### Describe the bug

When the game starts, your character will break through the floor and you will be unable to proceed any further.

https://user-images.githubusercontent.com/75052171/196016142-ea7ae0bc-8bf8-4950-a5d1-f06e4e9efb0a.mp4

(Video tested on the `2022-10-15 Nightly` build)

But i had no issues with the `2022-05-01 Nightly` build. (I haven't been able to pinpoint the exact date the problem reproduced.)
It is believed that the problem started with the build of Ruffle Nightly on a specific date.

[Castle Cat 4.swf](https://cdn.flasharch.com/share/castle-cat-4.swf)

### Expected behavior

The character must be able to walk on the ground.

### Affected platform

Self-hosted version

### Operating system

Windows 10 21H2

### Browser

Google Chrome 106

Jmousy:
Initially the problem was confirmed in Castle Cat 4, but as of recent builds, I have confirmed that the problem is the same in Castle Cat 1, 2, and 3.

n0samu:
I've bisected this to commit c17da6e91a9fbd7f9fb58854b95b6067b3793bb2. That PR was tested pretty thoroughly so there must be something quite exceptional going on here.

Toad06:
This hitTest seems to cause the issue:
`platform.hitTest(_X,platform._y,true)`

For some reason, collisions work as expected when removing the shapeFlag parameter (https://www.oreilly.com/library/view/actionscript-the-definitive/1565928520/re141.html).

PR #20811, opened 2025-06-23
render: Fix winding_number on edges

cookie-s:
Resolve https://github.com/ruffle-rs/ruffle/issues/11727
Resolve https://github.com/ruffle-rs/ruffle/issues/8303
Resolve https://github.com/ruffle-rs/ruffle/issues/8420
(Re)fixes https://github.com/ruffle-rs/ruffle/issues/5481
Fixes https://github.com/ruffle-rs/ruffle/issues/11454

At least in https://github.com/ruffle-rs/ruffle/issues/11727 , the issue was the return value from `platform.hitTest(_X,platform._y,true)` where the object (`platform`) and the y value of the point in parameter (`platform._y`) were the same. This PR tries to correct the hitTest result when the point is at edges.


Lord-McSweeney:
This also fixes #11454!

Toad06:
Thank you for your work on this and for checking #5481.

I believe this PR might also fix #12618, as the game was working before commit c17da6e91a9fbd7f9fb58854b95b6067b3793bb2.

kjarosh (approved):
LGTM, thank you!

PR #20825, opened 2025-06-24
avm2: Shrink Error to pointer-size

Lord-McSweeney:


kjarosh:
Why are we doing that? I'm not saying we shouldn't, but there should be some explanation in the commit message and in the PR description.

PR #20826, opened 2025-06-24
core,avm1,avm2: Remove more uses of GcCell

Lord-McSweeney:
After this, we have six uses of GcCell left:
- AVM2 `VTable`, `Class`, `Domain`, and `TranslationUnit`;
- `BitmapDataWrapper`
- `MovieClip`

### DIFF
--- a/core/src/avm1/object/script_object.rs
+++ b/core/src/avm1/object/script_object.rs
@@ -9,7 +9,7 @@ use crate::display_object::{DisplayObject, TDisplayObject as _};
 use crate::ecma_conversions::f64_to_wrapping_i32;
 use crate::string::{AvmString, StringContext};
 use core::{fmt, mem};
-use gc_arena::{Collect, GcCell, GcWeakCell, Mutation};
+use gc_arena::{lock::RefLock, Collect, Gc, GcWeak, Mutation};
 use ruffle_macros::istr;
 
 use super::super_object::SuperObject;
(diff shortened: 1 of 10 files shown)


core/src/avm2/e4x.rs
```diff
@@ -417,57 +411,39 @@ impl<'gc> E4XNode<'gc> {
...
                 children.push(child);
             }
-            _ => {
-                // FIXME - figure out exactly when appending is allowed in FP,
-                // and throw the proper AVM error.
-                return Err(Error::RustError(
-                    format!("Cannot append child {child:?} to node {:?}", this.kind).into(),
-                ));
-            }
+            _ => panic!("append_child must be called on an Element node"),
```
kjarosh: Why are we introducing a panic here?

core/src/avm1/object/script_object.rs
```diff
@@ -53,11 +53,11 @@ impl<'gc> Watcher<'gc> {
 
 #[derive(Copy, Clone, Collect)]
 #[collect(no_drop)]
-pub struct Object<'gc>(GcCell<'gc, ObjectData<'gc>>);
+pub struct Object<'gc>(Gc<'gc, RefLock<ObjectData<'gc>>>);
```
kjarosh, line 56: That doesn't change anything semantically, right? `GcCell` is a `Gc<RefLock>`

Wouldn't it be better to leave `GcCell` for now and revisit when we can get rid of locking the whole object?

core/src/avm2/e4x.rs
Lord-McSweeney: `append_child` is only used for building the node tree and should only ever be called on `Element`s, so we don't need error handling

core/src/avm1/object/script_object.rs
kjarosh: Chatted offline—object doesn't gain that much from interior mutability

Issue #14518, opened 2023-12-25
pbjASLockingTest: Failures, crashes, and unimplemented pixelbender stuff


torokati44:
# Error Info
Error name: Error
Error message: panicked at render/naga-pixelbender/src/lib.rs:1243:29:
Unimplemented opcode RSqrt

# Player Info
Allows script access: true
Renderer: wgpu
Adapter Backend: Gl
Adapter Name: "Radeon R9 200 Series"
Adapter Device Type: Other
Adapter Driver Name: ""
Adapter Driver Info: ""
Enabled features: Features(TEXTURE_COMPRESSION_BC | TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES | PUSH_CONSTANTS | SHADER_UNUSED_VERTEX_OUTPUT)
Available features: Features(TEXTURE_COMPRESSION_ETC2 | TEXTURE_COMPRESSION_ASTC | CLEAR_TEXTURE)
Current limits: Limits { max_texture_dimension_1d: 16384, max_texture_dimension_2d: 16384, max_texture_dimension_3d: 8192, max_texture_array_layers: 256, max_bind_groups: 4, max_bindings_per_bind_group: 1000, max_dynamic_uniform_buffers_per_pipeline_layout: 8, max_dynamic_storage_buffers_per_pipeline_layout: 0, max_sampled_textures_per_shader_stage: 16, max_samplers_per_shader_stage: 16, max_storage_buffers_per_shader_stage: 0, max_storage_textures_per_shader_stage: 0, max_uniform_buffers_per_shader_stage: 11, max_uniform_buffer_binding_size: 16384, max_storage_buffer_binding_size: 0, max_vertex_buffers: 8, max_buffer_size: 268435456, max_vertex_attributes: 16, max_vertex_buffer_array_stride: 255, min_uniform_buffer_offset_alignment: 32, min_storage_buffer_offset_alignment: 256, max_inter_stage_shader_components: 31, max_compute_workgroup_storage_size: 0, max_compute_invocations_per_workgroup: 0, max_compute_workgroup_size_x: 0, max_compute_workgroup_size_y: 0, max_compute_workgroup_size_z: 0, max_compute_workgroups_per_dimension: 0, max_push_constant_size: 96, max_non_sampler_bindings: 1000000 }
Surface quality: high
Surface samples: 4
Surface size: Extent3d { width: 550, height: 400, depth_or_array_layers: 1 }

# Page Info
[...]

PR #20827, opened 2025-06-24
pixelbender: Various fixes related to returning values

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/14519
  * It fixes the crashes, does not make all tests pass.
* Progresses https://github.com/ruffle-rs/ruffle/issues/14518

FP supports shaders with less than 3 output channels, it just throws an error when trying to run such shader. When returning a 3-channel value, alpha should be set to 1. FP also supports returning a different number of channels than specified.


core/src/avm2/globals/flash/display/shader_job.rs
```diff
@@ -234,6 +235,14 @@ pub fn start<'gc>(
         }
     };
 
+    match shader_handle.0.parsed_shader().output_channels() {
+        Some(3) | Some(4) => {}
+        channels => {
+            tracing::info!("Unexpected number of shader output channels: {channels:?}");
```
Lord-McSweeney: Could this be a `warn` instead of `info`, since it's an unexpected condition?

Lord-McSweeney (approved):

core/src/avm2/globals/flash/display/shader_job.rs
kjarosh: Done, thanks! I also changed the wording from "unexpected" to "unsupported", as it's not really unexpected to see the wrong number of channels.

render/naga-pixelbender/src/lib.rs
```diff
@@ -410,9 +410,26 @@ impl ShaderBuilder<'_> {
         // We've emitted all of the opcodes into the function body, so we can now load
         // from the destination register and return it from the function.
         let dst_load = builder.load_src_register(dst)?;
-        builder.push_statement(Statement::Return {
-            value: Some(dst_load),
-        });
+
+        // Set the alpha channel to 1 if the number of channels is <4.
```
adrian17, line 414: I _think_ there's an easier way than manually composing R/G/B/1; a direct translation of `value.a = 1.0` would be something like:
```rs
            Statement::Store {
                pointer: builder.evaluate_expr(Expression::AccessIndex { base: dst_load, index: 3 })
                value: builder.onef32,
            },
```
I didn't actually test this though.

kjarosh: https://github.com/ruffle-rs/ruffle/pull/20848

PR #20834, opened 2025-06-25
ci: Rename create-nightly-release to create-release

kjarosh:
Preparing for stable releases.

### DIFF
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -12,8 +12,8 @@ env:
   RELEASE_SCRIPT: ./.github/scripts/release.py
 
 jobs:
-  create-nightly-release:
-    name: Create Nightly Release
+  create-release:
+    name: Create Release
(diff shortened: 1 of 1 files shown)


torokati44:
Shouldn't the cron schedule also be adjusted?

kjarosh:
> Shouldn't the cron schedule also be adjusted?

Not yet, I'm only changing the name of the job here to get a big, unrelated diff out of the way

torokati44:
I'm still just this :pinching_hand:  far off from being convinced that we actually do need "stable" releases at this point in time, or that they would make any sense - as in, that they would be different in any meaningful way from just picking a random nightly that "feels" less risky to use, according to ... dunno what. I'm not in any way going to stand in the way of doing them, and I don't think it's a bad idea at all, I just don't feel like putting my green tickmark on it either. :face_with_peeking_eye:

kjarosh:
Closing due to lack of interest

Issue #19400, opened 2025-02-01
Achilles - spears and arrows stop hitting the ground


TomFulp:
### Describe the bug

Reported by MeGa-wise on Newgrounds:

"some bug stops thrown spears and enemy arrows from hitting ground and makes them stay in air that disables player from picking fallen spears from ground. This removes core gameplay feature. Game usually starts without this bug but it appears after first or second enemy archer appears and attacks."

### Expected behavior

Player should be able to pick up fallen spears.

### Content Location

https://www.newgrounds.com/portal/view/419538/format/flash?emulate=flash

### Affected platform

Self-hosted version

### Operating system

Windows 11

### Browser

Chrome

Croworbit:
duplicate of https://github.com/ruffle-rs/ruffle/issues/9863

Lord-McSweeney:
Closing as duplicate, the original issue was fixed by #20837

PR #20837, opened 2025-06-25
avm1: Fix timeline scope on target clip removal (close #9863)

### LINKED ISSUE #9863: Achilles filters and floating arrows and spears issue
### Describe the bug

flash
https://user-images.githubusercontent.com/109034847/222987136-d5576db5-6249-4c16-8bff-16332df45321.mp4

notice the arrows

ruffle
https://user-images.githubusercontent.com/109034847/222987153-beb79033-909d-45ce-91f2-c856a2f12448.mp4

### Expected behavior

should be like flash

### Affected platform

Desktop app

### Operating system

Linux Mint

### Browser

_No response_

### Additional information

[419538_achilles.swf.zip](https://github.com/ruffle-rs/ruffle/files/10893012/419538_achilles.swf.zip)

ste42:
When a `removeMovieClip` removes an activation's target clip, the target clip updates to something else. The first thing it tries is to update it to the base clip. It tries to switch variable resolution to the base clip's timeline scope as well.

If the base clip is also removed (like with a `removeMovieClip(this)` that isn't inside a `tellTarget`), functions like `call(1)` target the root instead, and variable resolution switches to the root clip's timeline scope. This behavior mimics what occurs when a `tellTarget` is called with an invalid clip: the target clip becomes invalid (so certain actions target the root instead), and the timeline scope also switches to the root's.

Additionally, when the base clip is removed, scripts stop execution after a `call`.

This PR adds logic to handle these cases and a test.

I made the test with JPEXS, so unfortunately I can't provide an `fla` (I traced some comments to hopefully make it easier to see overall what the test does).

Fixes #9863 (and #19400, which is a duplicate)


Lord-McSweeney:
Also fixes #3118

adrian17 (approved):
We discussed an even more aggressive and general solution for this (involving string paths; an extension of https://github.com/ruffle-rs/ruffle/pull/20861), but this by itself feels safe and is already written, so let's get it in.

ste42:
That's great to hear!
I recently found a few other weird situations with removed clips, so a general solution sounds neat. I'll create an issue with the info for those situations to help out.

PR #20840, opened 2025-06-26
AVM1: Fix MovieClipReference laundering issue

### LINKED ISSUE #15265: Clicker Monsters: After collecting 100 coins, no more coins appear
### Describe the bug

In the game **Clicker Monsters**(AVM1), you earn coins every time you defeat a monster. (Coins are generated indefinitely, regardless of whether you clear the stage or not.)

However, after you collect a total of 100 coins, no more coins appear for defeating monsters.
If you move to a different stage or purchase an upgrade, the problem persists.

However, after restarting the game to load your saved history, the coins will appear again.

(If the video doesn't play, you can download it to watch)

https://github.com/ruffle-rs/ruffle/assets/75052171/3616b3a2-47df-456f-8785-2c0c09b21900

### Expected behavior

You should be able to earn coins every time you defeat a monster.

### Content Location

Attached file:

[clicker-monsters.zip](https://github.com/ruffle-rs/ruffle/files/14321571/clicker-monsters.zip)

### Affected platform

Desktop app

### Operating system

Windows 10 22H2

### Browser

Google Chrome 121

### Additional information

_No response_

CUB3D:
This (finally) fixes the "reference  laundering" issue present in AVM1 since the string paths fix was implemented.

Specifically this fixes two different kinds of laundering:
1: Calling a function with a reference to a removed clip, which resolved the clip to undefined before passing it on
2: Converting `MovieClipReference` -> `Object` -> `MovieClipReference` could create a reference that points to a different value due to the weirdness of the logic around the cache

Fixing the former fixes #15265, fixing the later fixes the existing known-failing test

### DIFF
--- a/core/src/display_object/movie_clip.rs
+++ b/core/src/display_object/movie_clip.rs
@@ -2243,7 +2243,7 @@ impl<'gc> MovieClip<'gc> {
     }
 
     /// Mark this `MovieClip` as being originated from resolving the given `MovieClipReference`
-    pub fn get_original_reference(
+    pub fn set_original_reference(
         &self,
         context: &mut UpdateContext<'gc>,
         mcr: MovieClipReference<'gc>,
--- a/core/src/display_object/movie_clip.rs
+++ b/core/src/display_object/movie_clip.rs
@@ -1,4 +1,7 @@
 //! `MovieClip` display object and support code.
+use super::interactive::Avm2MousePick;
+use super::BitmapClass;
+use crate::avm1::object_reference::MovieClipReference;
 use crate::avm1::Avm1;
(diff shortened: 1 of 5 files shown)


core/src/display_object/movie_clip.rs
```diff
@@ -2242,6 +2250,20 @@ impl<'gc> MovieClip<'gc> {
             self.run_goto(context, frame, false);
         }
     }
+
+    /// Mark this `MovieClip` as being originated from resolving the given `MovieClipReference`
+    pub fn get_original_reference(
```
moulins: Possible typo: should probably be `set_original_reference`.

PR #20860, opened 2025-06-28
ci: Publish release commits for nightly

kjarosh:
This PR adds publishing commits for nightly, but _only on tags_, not on the main branch.

### Why?

* Downstreams can use the proper version (instead of having always 0.1.0, currently our builds have the proper version set)
* Downstreams don't have to work around including the version in metainfo (this PR does not implement that, but it makes it possible)
* The logic of creating a release commit for both stable and nightly can be the same (assuming we'll go forward with stable releases)

By downstreams I mean mostly Flatpak (but also build-from-source AURs, etc.).

TL;DR It makes maintaining Flatpak (and other downstreams) easier & reduces differences between release binaries and Flatpak (and other downstreams).

### Before:

```
* <- nightly C
|
*
|
* <- nightly B
|
*
|
*
|
* <- nightly A
|
*
```

### After:

```
  * <- nightly C
 /
*
|
* * <- nightly B
|/
*
|
*
|
* * <- nightly A
|/
*
|
*
```

An example on how such a release would look like: https://github.com/kjarosh/ruffle/releases/tag/nightly-2025-06-28

Discussion is welcome

### DIFF
--- a/.github/scripts/release.py
+++ b/.github/scripts/release.py
@@ -28,6 +28,13 @@ def get_current_day_id():
     return f'{now.year - 2000}{day}'
 
 
+def get_tag_name():
+    now = datetime.now()
+    current_time_dashes = now.strftime('%Y-%m-%d')
+    tag_name = f'nightly-{current_time_dashes}'
+    return tag_name
+
+
 def github_output(variable, value):
     line = f'{variable}={value}'
     print(line)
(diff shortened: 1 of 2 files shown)


torokati44:
Since there is no branch created for these tagged commits, I don't see them in `gitk`, and the GH web interface also says this about them:

> :warning: This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.

The former makes bisecting between nightlies locally a lot more difficult (for me).

kjarosh:
> The former makes bisecting between nightlies locally a lot more difficult (for me).

Those tags should still be present in the repository and you can bisect between them without an issue (but git will ask to test the merge base first).

torokati44:
They are in the repo of course, but `gitk` specifically doesn't show them for me without a branch. And I usually bisect manually... 😶

Issue #14562, opened 2023-12-31
`Avm1Function`'s `base_clip` field should be a string path, not a display object


Lord-McSweeney:
### Describe the bug

In `core/src/avm1/function.rs`
```
pub struct Avm1Function<'gc> {
    ...
    base_clip: DisplayObject<'gc>, // <-- This should be a string path to the base clip
    ...
}
```
I haven't tested whether this is version-gated or not.

### Expected behavior

The function references whatever DO is at the path instead of keeping a reference to a specific DO.

### Content Location

Champion Archer (#6124) is broken due to this.

### Affected platform

Desktop app

### Operating system

Irrelevant

### Browser

Irrelevant

### Additional information

This is the root cause of #6124.

Dinnerbone:
Test for this:
[function_base_clip_readded.zip](https://github.com/ruffle-rs/ruffle/files/15370676/function_base_clip_readded.zip)

Lord-McSweeney:
Fixed by #20861

Issue #6124, opened 2022-01-24
Champion Archer bow not working after first level


FynnFreyer:
### Describe the bug

# Steps to reproduce
1. Load the game
2. Finish the first level
3. Try to shoot an arrow in the second level
4. It doesn't shoot an arrow

# SWF file
[Champion Archer](https://github.com/ruffle-rs/ruffle/files/7928208/champion_archer.zip)

# Visual problems
None

### Expected behavior

When you finish the first level, you can still use your bow afterwards.

### Affected platform

Desktop app and browser demo

### Operating system

Ubuntu 20.04

### Browser

Firefox 96.0 (64-bit) on Ubuntu

### Additional information

I did test this in the [online demo](https://ruffle.rs/demo/) and with the [desktop app](https://github.com/ruffle-rs/ruffle/releases/tag/nightly-2022-01-24) (linux-x86_64)

Toad06:
Thanks for the report.

The issue is around these lines in `\DefineSprite (157)\frame 1\PlaceObject2 (134) Depth: 5 (arms)\onClipEvent(load)`:

```
onClipEvent(load){
   if(_root.setListen != true)
   {
      click = new Object();
      click.onMouseDown = function()
      {
         trace(this._currentframe);
         trace(_currentframe);
         if(_currentframe == 1)
         {
            // code to shoot an arrow...
         }
...
```

`_currentframe` always holds a value of 1 in Flash Player, but on Ruffle the value changes to 3 when reaching the second level.

It appears the game was partially working in nightly-2021-10-13 and the older releases (it's possible to shoot arrows in the second level but there are other issues related to #5492), so it seems like the commit e9fbed7be0d33e47fe5cc522fc468ebbdfc4a2bf (PR #5465) introduced a regression.

@Herschel, maybe you have an idea of what's going on here?

FynnFreyer:
Yes, when you e.g. [play on addictinggames.com](https://www.addictinggames.com/shooting/champion-archer) (nightly 2021-04-04) there is another issue with the bow (position doesn't update anymore). Just as an example of it 'working' in an older release.

Addictinggames is where I initially found this project btw, but I thought just mentioning my test with the newest nightly version might be better for clarity. 

PS:
Even though this might not be the right place, thanks a lot for putting work into this. I have many fond memories of old flash games, and this project helps keeping those available!

dginovker:
[...]

PR #20861, opened 2025-06-29
AVM1: Avm1Function's base_clip should be a string path

CUB3D:
The base_clip of an `Avm1Function` is also accessed with string path semantics.

This fixes both #6124 and #14562


core/src/avm1/activation.rs
```diff
@@ -870,14 +870,16 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
         let func = Avm1Function::from_swf_function(
             self.gc(),
             swf_version,
             func_data,
             action,
             self.scope(),
             constant_pool,
-            self.base_clip(),
+            // `base_clip` should always be a living `MovieClip` so this can't fail
+            MovieClipReference::try_from_stage_object(self, bc).unwrap(),
```
CUB3D, line 882: I'm open for an alternative to the `unwrap`, but I've manged to convince myself it's safe

core/src/avm1/function.rs
```diff
@@ -360,14 +360,21 @@ impl<'gc> Executable<'gc> {
...
+        let base_clip = af.base_clip.coerce_to_object(activation);
+
+        let avm1_removed = base_clip
+            .and_then(|o| o.as_display_object())
+            .is_none_or(|d| d.avm1_removed());
+
+        let base_clip = if (is_closure || reason == ExecutionReason::Special) && !avm1_removed {
+            base_clip
+                .and_then(|d| d.as_display_object())
+                .unwrap_or(target)
```
CUB3D: I'm also not sure this `unwrap_or` is actually reachable in practice

Lord-McSweeney: Could you change it to an `unwrap()` in that case?

CUB3D: Sure, changed it to an expect so we can spot it a bit easier

core/src/avm1/function.rs
```diff
@@ -360,14 +360,21 @@ impl<'gc> Executable<'gc> {
...
+
+        let base_clip = if (is_closure || reason == ExecutionReason::Special) && !avm1_removed {
+            base_clip
+                .and_then(|d| d.as_display_object())
+                .expect("Somehow the base clip isn't an object or has no display object")
+        } else {
+            this_obj
+                .and_then(|this| this.as_display_object())
+                .unwrap_or(target)
+        };
```
moulins: Accessing a `MovieClipReference` will never return a removed clip, so I believe this can be simplified to the following:
```rs
let base_clip = if is_closure || reason == ExecutionReason::Special {
    af.base_clip.resolve_reference(activation).map(|_, _, dobj| dobj)
} else {
    this_obj
        .and_then(|this| this.as_display_object())
        .unwrap_or(target)
};
```

CUB3D: That would work except the base clip itself can be removed, the following works however:
```rs
        let this_do = this_obj
            .and_then(|this| this.as_display_object())
            .unwrap_or(target);

        let base_clip = if is_closure || reason == ExecutionReason::Special {
            af.base_clip
                .resolve_reference(activation)
                .map(|(_, _, dobj)| dobj)
                .unwrap_or(this_do)
        } else {
            this_do
        };
```
Do you think that's better?

moulins: LGTM

Issue #14000, opened 2023-11-13
starling - masks not working


zjk261:
### Describe the bug

https://github.com/ruffle-rs/ruffle/assets/98996990/8975b936-af83-40d2-9f4c-a9387d589432

### Expected behavior

https://github.com/ruffle-rs/ruffle/assets/98996990/63e8a2e6-a83b-4802-8eb5-f1a9f6be42e6

### Content Location

[StarlingBenchmark.zip](https://github.com/ruffle-rs/ruffle/files/13329789/StarlingBenchmark.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

Lord-McSweeney:
Not exactly sure but I'm guessing that this is unsupported `Context3D.setStencilActions()`.

PR #20875, opened 2025-06-30
avm2: Implement setStencilReferenceValue and setStencilActions on flash.display3D.Context3D

SuchAFuriousDeath:
Implements #14000 (Tested and it works properly now).

### DIFF
--- a/core/src/avm2/globals/flash/display3D/context_3d.rs
+++ b/core/src/avm2/globals/flash/display3D/context_3d.rs
@@ -293,18 +293,10 @@ pub fn set_culling<'gc>(
     if let Some(context) = this.as_context_3d() {
         let culling = args.get_string(activation, 0)?;
 
-        let culling = if &*culling == b"none" {
-            Context3DTriangleFace::None
-        } else if &*culling == b"back" {
-            Context3DTriangleFace::Back
-        } else if &*culling == b"front" {
-            Context3DTriangleFace::Front
-        } else if &*culling == b"frontAndBack" {
-            Context3DTriangleFace::FrontAndBack
-        } else {
+        let culling = Context3DTriangleFace::from_wstr(&culling).unwrap_or_else(|| {
             tracing::error!("Unknown culling {:?}", culling);
             Context3DTriangleFace::None
-        };
+        });
 
         context.set_culling(culling);
     }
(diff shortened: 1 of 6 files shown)


kjarosh:
We'll need some tests for it, probably a visual test to check that it renders properly, and a test to verify the API (it can be one test if that's easier, but usually we focus on different things there).

render/src/backend.rs
```diff
@@ -296,6 +296,22 @@ pub enum Context3DTriangleFace {
     FrontAndBack,
 }
 
+impl Context3DTriangleFace {
+    pub fn from_wstr(s: &WStr) -> Option<Self> {
+        if s == b"none" {
+            Some(Context3DTriangleFace::None)
+        } else if s == b"back" {
+            Some(Context3DTriangleFace::Back)
+        } else if s == b"Front" {
```
kjarosh: Edge cases like this one should be tested too.

SuchAFuriousDeath:
I'll put this in draft until I get the tests done.

render/src/backend.rs
```diff
@@ -429,6 +429,76 @@ impl FromWStr for Context3DCompareMode {
     }
 }
 
+#[cfg(feature = "wgpu")]
```
Lord-McSweeney: Could this be moved to the `wgpu` part of `render` to avoid having to feature-gate it?

render/src/backend.rs
```diff
@@ -429,6 +429,76 @@ impl FromWStr for Context3DCompareMode {
...
+            Ok(Context3DStencilAction::Set)
+        } else if s == b"zero" {
+            Ok(Context3DStencilAction::Zero)
+        } else {
+            Err(())
+        }
+    }
+}
+
+#[cfg(feature = "wgpu")]
```
Lord-McSweeney: Here too

render/src/backend.rs
SuchAFuriousDeath: Actually it couldn't, because of the orphan rule. In render, the Contexd3D nor the wgpu types are local to that crate :(

kjarosh: In that case you can use a named method instead of the `From` trait. I don't like the fact that it's implemented behind a feature too especially that we have a dedicated crate for its implementation.

SuchAFuriousDeath: I guess I'll create an IntoWgpu trait, otherwise it's annoying to name these functions :sweat_smile:

core/src/avm2/globals/flash/display3D/context_3d.rs
```diff
@@ -623,6 +623,53 @@ pub fn set_render_to_texture<'gc>(
...
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let this = this.as_object().unwrap();
+
+    if let Some(context) = this.as_context_3d() {
+        let triangle_face = args.get_string(activation, 0);
+
+        let triangle_face = triangle_face
+            .parse()
+            .inspect_err(|_| tracing::error!("Unknown triangle_face {:?}", triangle_face))
```
kjarosh: It's not really an error—we either (1) don't know what happens in this case (a warning stub? a TODO?) or (2) we handle the case properly (no logs).

If you're not sure whether we handle that properly go with (1), if you're sure go with (2). Of course I would recommend writing a test to check this edge case.

core/src/avm2/globals/flash/display3D/context_3d.rs
```diff
@@ -623,6 +623,53 @@ pub fn set_render_to_texture<'gc>(
...
+            .unwrap_or(Context3DTriangleFace::None);
+
+        let compare_mode = args
+            .get_string_non_null(activation, 1, "compareMode")?
+            .parse()
+            .map_err(|_| make_error_2008(activation, "compareMode"))?;
```
kjarosh, line 693: Can we add tests for these exceptions? It can be a separate, non-image test.

core/src/avm2/globals/flash/display3D/context_3d.rs
kjarosh: According to docs, we should throw an error like below, but should be easy to write a simple test for it.

render/wgpu/src/context3d/current_pipeline.rs
```diff
@@ -122,6 +165,15 @@ impl CurrentPipeline {
 
             depth_mask: true,
             pass_compare_mode: wgpu::CompareFunction::LessEqual,
+
+            stencil: wgpu::StencilState {
+                front: wgpu::StencilFaceState::IGNORE,
+                back: wgpu::StencilFaceState::IGNORE,
+                read_mask: !0,
```
kjarosh: Better to be explicit: use `u32::MAX` instead of `!0`, you don't have to worry about the type (size, signedness) and you don't have to wonder what that expression means (because it's not commonly used).

render/wgpu/src/context3d/current_pipeline.rs
```diff
@@ -64,6 +104,9 @@ pub struct CurrentPipeline {
     depth_mask: bool,
     pass_compare_mode: wgpu::CompareFunction,
 
+    stencil: wgpu::StencilState,
+    stencil_ref_value: u32,
```
kjarosh, line 110: Can you add a doc and explain what that is?

render/wgpu/src/context3d/current_pipeline.rs
kjarosh: Actually now thinking about this, it's a mask so all we care about is that all bits are set. 🤔 What do you think about this?

kjarosh:
Code LGTM, I think we should cover edge cases in tests, currently only the happy path is covered.

SuchAFuriousDeath:
Edge cases with the string parsing (working on that)? Or something more than that?

render/wgpu/src/context3d/current_pipeline.rs
SuchAFuriousDeath: I'm fine with either.

core/src/avm2/globals/flash/display3D/context_3d.rs
SuchAFuriousDeath: Done

core/src/avm2/globals/flash/display3D/context_3d.rs
SuchAFuriousDeath: I changed it to get_string_non_null like the other ones, tests included.

kjarosh:
There are some edge cases in `setStencilReferenceValue`, but I don't think they are testable without visuals? If you feel like writing a test for that, great, but I don't think it's strictly necessary, a TODO is fine.

SuchAFuriousDeath:
> There are some edge cases in `setStencilReferenceValue`, but I don't think they are testable without visuals? If you feel like writing a test for that, great, but I don't think it's strictly necessary, a TODO is fine.

https://github.com/ruffle-rs/ruffle/pull/20875/changes#diff-46759425764463d9c159e31f66487e738e5a3917b0ab94e562795703a371aa2fR683

Hopefully this is what you meant.

I think this PR already does enough :sweat_smile: Maybe I'll do it some other time, if I feel like it.

SuchAFuriousDeath:
With regards to the string parsing: Everything is case-sensitive and uses get_string_non_null, except Context3DVertexBufferFormat, which uses get_string.

Lord-McSweeney (approved):
I'm not very experienced with the wgpu code, but the AS3 part lgtm

kjarosh (approved):

PR #20896, opened 2025-06-30
exporter: Add Support for Full Auto-Export, End-on-Last-Frame Logic, and Prevent --force-play Looping

AeroXuk:
Let me know if you'd like any of these features to be placed behind new CLI arguments. I'm happy to update the PR accordingly.

Changes for the Exporter Tool:
- Support `--frames 0` to export all frames. #20930
- Add a check to exit the frame render loop once the root movie clip reaches its final frame. #20929
- Improve the `--force-play` feature to avoid restarting the timeline if it's already on the last frame. #20927
- Zero-pad filenames when exporting multiple frames to a directory, ensuring proper sort order and compatibility with external tools. #20928

### DIFF
--- a/exporter/src/main.rs
+++ b/exporter/src/main.rs
@@ -186,16 +186,10 @@ fn take_screenshot(
 fn force_root_clip_play(player: &Arc<Mutex<Player>>) {
     let mut player_guard = player.lock().unwrap();
 
-    // Check and resume if suspended
-    if !player_guard.is_playing() {
-        player_guard.set_is_playing(true);
-    }
-
-    // Also resume the root MovieClip if stopped
     player_guard.mutate_with_update_context(|ctx| {
         if let Some(root_clip) = ctx.stage.root_clip() {
             if let Some(movie_clip) = root_clip.as_movie_clip() {
-                if !movie_clip.playing() {
+                if !movie_clip.playing() && movie_clip.current_frame() < movie_clip.total_frames() {
                     movie_clip.play();
                 }
             }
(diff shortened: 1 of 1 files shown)


exporter/src/main.rs
```diff
@@ -52,7 +52,8 @@ struct Opt {
     #[clap(name = "output")]
     output_path: Option<PathBuf>,
 
-    /// Number of frames to capture per file
+    /// Number of frames to capture per file.
+    /// If set to 0, all frames in the SWF's main timeline will be captured
```
kjarosh, line 56: It makes more sense for me to pass `--frames all` instead of `--frames 0` to capture all frames. It looks non-intuitive

exporter/src/main.rs
```diff
@@ -168,23 +178,28 @@ fn take_screenshot(
         if let Some(progress) = &progress {
             progress.inc(1);
         }
+
+        if is_root_movie_clip_at_end(&player) {
+            break;
+        }
```
kjarosh, line 184: That will break some animations, it has to be another CLI option

kjarosh (commented):
Can you split that PR into multiple ones? A lot of unrelated things are going on here

exporter/src/main.rs
AeroXuk: I'll look into this but I'm still quite new to rust, so not 100% sure what the best way to implement a cli command that takes a string or number would be

AeroXuk: I've done some research and hope I've found the right way to implement this.
I need to run a few more tests, but I've raised this as #20930.

exporter/src/main.rs
AeroXuk: Made a new cli command and raised under #20929

AeroXuk:
Closing this PR as split into seperate changes as advised.

#20930
#20929
#20927
#20928

PR #20905, opened 2025-07-01
chore: Update translations

RuffleBuild:


torokati44 (approved):

desktop/assets/texts/ar-SA/volume_dialog.ftl
torokati44, line 1: Several AIs that (supposedly) speak Arabic (unlike me) found only the changes to this file problematic. However, they are almost certainly still understandable, so... They can always be improved later. I'll trust the human contributor(s) instead.

PR #20906, opened 2025-07-01
avm2: Add worker-related Rust-side objects

kjarosh:
Adding those objects in order to stub them, as their classes are abstract.


core/src/avm2/object/worker_object.rs
```diff
@@ -0,0 +1,51 @@
+use crate::avm2::activation::Activation;
+use crate::avm2::object::script_object::ScriptObjectData;
+use crate::avm2::object::{Object, ObjectPtr, TObject};
+use crate::avm2::Error;
+use crate::utils::HasPrefixField;
+use core::fmt;
+use gc_arena::{Collect, Gc, GcWeak};
+
+/// A class instance allocator that allocates Worker objects.
+#[allow(dead_code)]
```
Lord-McSweeney: We don't want allocators for these objects since they can't be allocated from AS anyway (the classes are `Abstract`). Native code should be able to construct them with some kind of `new` function, like
```
pub fn new(context: &mut UpdateContext<'gc>) -> Self {
    let class = context.avm2.classes().worker;
    WorkerObject(Gc::new(
        context.gc(),
        WorkerObjectData {
            base: ScriptObjectData::new(class),
        },
    ))
}

kjarosh: Sorry, too much copy-paste, moved them to `new` constructors

Lord-McSweeney (approved):

PR #20907, opened 2025-07-02
avm2: remove `GcCell` in `VTable`

moulins:
To avoid having to put (almost) every `VTable` field behind a `OnceLock`, this also changes the `Class/ClassObject` vtable initialization logic to fully replace the vtable object instead of mutating it in place.

### DIFF
--- a/core/src/avm2/class.rs
+++ b/core/src/avm2/class.rs
@@ -825,24 +825,28 @@ impl<'gc> Class<'gc> {
             panic!(
                 "Attempted to initialize vtable on a class that did not have its traits loaded yet"
             );
+        } else if self.0.all_interfaces.get().is_some() {
+            panic!("Attempted to initialize vtable twice");
         }
 
-        let vtable = VTable::new(
+        let write = Gc::write(context.gc(), self.0);
+
+        let interfaces = self.gather_interfaces()?;
+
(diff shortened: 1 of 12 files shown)


core/src/avm2/class.rs
```diff
@@ -144,7 +144,11 @@ pub struct ClassData<'gc> {
     /// properties that would match.
     traits: OnceLock<Box<[Trait<'gc>]>>,
 
-    vtable: VTable<'gc>,
+    /// The class' vtable.
+    ///
+    /// Newly-created `Class`es begin with a dummy vtable: the `init_vtable`
+    /// method should be called to replace it with the real vtable.
+    vtable: Lock<VTable<'gc>>,
```
adrian17: Could this use a OnceLock instead? Or is the dummy vtable actually observed by something?

moulins: Nothing ever observes the dummy vtable; I guess you could call this a micro-optimisation, as a `OnceLock` would require a check on each access.

adrian17: Not sure about perf, I was more thinking in terms of preventing issues like accidentally touching the dummy vtable; Dinnerbone just recently stumbled upon this kind of issue, with observing the dummy SwfMovie we create "temporarily".
@Lord-McSweeney what do you think?

PR #20915, opened 2025-07-03
avm2: Properly mark classes as abstract

kjarosh:
This PR goes through all classes and makes sure a proper exception is thrown when they are being instantiated.


core/src/avm2/error.rs
```diff
@@ -571,6 +571,26 @@ pub fn make_error_2008<'gc>(activation: &mut Activation<'_, 'gc>, param_name: &s
     }
 }
 
+#[inline(never)]
+#[cold]
+pub fn make_error_2012<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    class_name: impl ToString,
```
moulins: Very small nit: it would make more sense to take `impl Display` instead

kjarosh: That makes sense, thanks!

core/src/avm2/object/security_domain_object.rs
```diff
@@ -0,0 +1,49 @@
...
+        HasPrefixField::as_prefix_gc(self.0)
+    }
+
+    fn as_ptr(&self) -> *const ObjectPtr {
+        Gc::as_ptr(self.0) as *const ObjectPtr
+    }
+}
+
+impl<'gc> SecurityDomainObject<'gc> {
+    #[allow(dead_code)]
```
evilpie: Isn't it used?

kjarosh: Right, too much copy-paste

core/src/avm2/globals/flash/text/TextSnapshot.as
```diff
@@ -4,11 +4,8 @@ package flash.text {
     import __ruffle__.stub_method;
     import __ruffle__.stub_getter;
 
+    [Ruffle(Abstract)]
     public class TextSnapshot {
-        public function TextSnapshot() {
-            stub_constructor("flash.text.TextSnapshot");
```
Lord-McSweeney, line 9: nit: the `stub_constructor` import can be removed now

core/src/avm2/globals/flash/media/StageVideo.as
```diff
@@ -0,0 +1,7 @@
+package flash.media {
+    [API("670")]
+    [Ruffle(Abstract)]
+    public final class StageVideo {
```
Lord-McSweeney: Playerglobals and docs say that StageVideo isn't `final`

core/src/avm2/globals/flash/media/StageVideo.as
```diff
@@ -0,0 +1,7 @@
+package flash.media {
+    [API("670")]
```
Lord-McSweeney: At least for me, this class is accessible in SWFv10, so this should be `[API("667")]`.

Lord-McSweeney (approved):

core/src/avm2/globals/flash/text/TextSnapshot.as
kjarosh: Done

core/src/avm2/globals/flash/media/StageVideo.as
kjarosh: Removed final

core/src/avm2/globals/flash/media/StageVideo.as
kjarosh: You're right, docs are inaccurate here. Checked in FP 10,0,42,34 and it's not available there.

PR #20925, opened 2025-07-05
core: properly free memory in `BitmapData::dispose`

moulins:
`Vec::clear` doesn't actually free the backing memory, so replace with an empty vec instead.

### DIFF
--- a/core/src/bitmap/bitmap_data.rs
+++ b/core/src/bitmap/bitmap_data.rs
@@ -608,7 +608,7 @@ impl<'gc> BitmapData<'gc> {
     pub fn dispose(&mut self) {
         self.width = 0;
         self.height = 0;
-        self.pixels.clear();
+        self.pixels = Vec::new(); // free the CPU pixel buffer
         self.bitmap_handle = None;
         // There's no longer a handle to update
         self.dirty_state = DirtyState::Clean;


core/src/bitmap/bitmap_data.rs
```diff
@@ -608,10 +608,9 @@ impl<'gc> BitmapData<'gc> {
     pub fn dispose(&mut self) {
         self.width = 0;
         self.height = 0;
-        self.pixels.clear();
-        self.bitmap_handle = None;
-        // There's no longer a handle to update
-        self.dirty_state = DirtyState::Clean;
+        self.pixels = Vec::new(); // clear CPU memory
+        self.bitmap_handle = None; // clear GPU memory
```
kjarosh: Just to make sure: dropping all `BitmapHandle`s frees GPU memory?

Small nit: it won't clear any memory when someone else uses this handle.

adrian17: Yeah, BitmapHandle is an Arc and on wgpu dropping the underlying object should get wgpu to free the GPU resource automatically.

adrian17 (approved):
Diff LGTM, but yeah the comment is just slightly misleading (though in practice BitmapHandle _should_ be the only owner, I think)

PR #20927, opened 2025-07-05
exporter: Refactor force_root_clip_play to prevent force play from looping.

AeroXuk:
Split from #20896.

- Improve the `--force-play` feature to avoid restarting the timeline if it's already on the last frame.

### DIFF
--- a/exporter/src/lib.rs
+++ b/exporter/src/lib.rs
@@ -173,18 +173,10 @@ fn take_screenshot(
 }
 
 fn force_root_clip_play(player: &Arc<Mutex<Player>>) {
-    let mut player_guard = player.lock().unwrap();
-
-    // Check and resume if suspended
-    if !player_guard.is_playing() {
-        player_guard.set_is_playing(true);
-    }
-
-    // Also resume the root MovieClip if stopped
-    player_guard.mutate_with_update_context(|ctx| {
+    player.lock().unwrap().mutate_with_update_context(|ctx| {
         if let Some(root_clip) = ctx.stage.root_clip() {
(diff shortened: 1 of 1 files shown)


kjarosh:
What's the purpose of it? Will that be even necessary after https://github.com/ruffle-rs/ruffle/pull/20929 ?

AeroXuk:
> What's the purpose of it? Will that be even necessary after #20929 ?

#20929 is adding a way to exit the export loop early if we reach the end.

Using the following senario, if you request it to export 200 frames of a 100 frame animation:

- Without using `--force-play`, you will have frames 1 to 100, then 100 copies of the last frame.
- Currently with `--force-play`, you will have frames 1 to 100 twice as it loops back to the begining.

The current looping behaviour with `--force-play` was not intended, the feature was intended for bypassing "click to play" or "click to continue" prompts. Adding this bugfix, would make the output match what happens if we aren't using the `--force-play`, once we reach the end we just keep outputting the last frame.

This then makes #20929 a bonus feature to also allow the exporter to stop after the last frame was exported, regardless of how many frames are requested.

exporter/src/main.rs
```diff
@@ -173,18 +173,10 @@ fn take_screenshot(
 }
 
 fn force_root_clip_play(player: &Arc<Mutex<Player>>) {
-    let mut player_guard = player.lock().unwrap();
-
-    // Check and resume if suspended
-    if !player_guard.is_playing() {
-        player_guard.set_is_playing(true);
-    }
```
AeroXuk: It also turns out this section didn't seem to add anything that the lower section already did. So I removed the duplication.

PR #20929, opened 2025-07-05
exporter: Add --stop-on-last-frame command.

AeroXuk:
Spit from #20896

- Add a check to exit the frame render loop once the root movie clip reaches its final frame.

### DIFF
--- a/exporter/src/lib.rs
+++ b/exporter/src/lib.rs
@@ -60,6 +60,10 @@ pub struct Opt {
     #[clap(long = "skipframes", default_value = "0")]
     skipframes: u32,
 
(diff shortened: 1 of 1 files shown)


exporter/src/main.rs
```diff
@@ -97,6 +101,7 @@ fn take_screenshot(
     progress: &Option<ProgressBar>,
     size: SizeOpt,
     force_play: bool,
+    stop_on_last_frame: bool,
```
danielhjacobs: According to clippy this is too many args, but it's probably fine to just add `#[allow(clippy::too_many_arguments)]`.

PR #20930, opened 2025-07-05
Update frames cli command to accept "all" to export all frames.

AeroXuk:
Split from #20896

- Support `--frames all` to export all frames.

### DIFF
--- a/exporter/src/main.rs
+++ b/exporter/src/main.rs
@@ -18,6 +18,7 @@ use std::fs::create_dir_all;
 use std::io::{self, Write};
 use std::panic::catch_unwind;
 use std::path::{Path, PathBuf};
+use std::str::FromStr;
 use std::sync::Arc;
 use std::sync::Mutex;
 use walkdir::{DirEntry, WalkDir};
@@ -37,6 +38,32 @@ struct SizeOpt {
     height: Option<u32>,
 }
 
+#[derive(Debug, Clone, Copy)]
+enum FrameSelection {
+    All,
+    One,
+    Count(u32),
+}
+
+impl FromStr for FrameSelection {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let s_lower = s.to_ascii_lowercase();
+        if s_lower == "all" {
+            Ok(FrameSelection::All)
+        } else if let Ok(n) = s.parse::<u32>() {
+            if n == 1 {
+                Ok(FrameSelection::One)
+            } else {
+                Ok(FrameSelection::Count(n))
+            }
+        } else {
+            Err(format!("Invalid value for --frames: {s}"))
+        }
+    }
+}
+
 #[derive(Parser, Debug)]
 #[clap(name = "Ruffle Exporter", author, version)]
 struct Opt {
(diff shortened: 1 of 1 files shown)


exporter/src/main.rs
```diff
@@ -37,6 +38,32 @@ struct SizeOpt {
     height: Option<u32>,
 }
 
+#[derive(Debug, Clone, Copy)]
+enum FrameSelection {
+    All,
+    One,
+    Count(u32),
```
kjarosh: You can replace `One` and `Count(u32)` with `Count(NonZeroUsize)`

exporter/src/main.rs
```diff
@@ -52,9 +79,9 @@ struct Opt {
     #[clap(name = "output")]
     output_path: Option<PathBuf>,
 
-    /// Number of frames to capture per file
+    /// Number of frames to capture per file. Use 'all' to capture all frames."
```
kjarosh:
```suggestion
    /// Number of frames to capture per file. Use 'all' to capture all frames.
```

exporter/src/main.rs
```diff
@@ -92,7 +119,7 @@ struct Opt {
 fn take_screenshot(
     descriptors: Arc<Descriptors>,
     swf_path: &Path,
-    frames: u32,
+    frames: FrameSelection,
```
kjarosh: Instead of passing `FrameSelection`. you can pass the concrete number of frames. That way progress will still work.

kjarosh:
Your approach of parsing `FrameSelection` is the right way to do it!

exporter/src/main.rs
AeroXuk: The reason for having a seperate `One` was because there are several places in the code we check to see if we are exporting a single frame. So having a shortcut option for that makes those checks easier.
Happy to replace though if you think combing them improves it?

exporter/src/main.rs
AeroXuk: I'll have to spend sometime to figure out how to get the frame count earlier. Currently I get it via the `player` object which is constructed within `take_screenshot`.

exporter/src/main.rs
kjarosh: You can replace those check with
```
matches!(frames, FrameSelection::Count(1))
```

It doesn't make sense to separate one and all other values, as both are numerals.

exporter/src/main.rs
kjarosh: Alright, I think that's fine for now, I'd just appreciate if you added a TODO comment that mentions the issue.

AeroXuk: Added TODO notes

kjarosh:
I'm sorry for stalling your exporter PRs, I've been adding some tests for exporter so that we don't break existing use cases and we can cover future ones (https://github.com/ruffle-rs/ruffle/pull/21029, https://github.com/ruffle-rs/ruffle/pull/21042).

I think the PRs can be rebased now (the `main.rs` diff should apply to `lib.rs` now).

AeroXuk:
> I'm sorry for stalling your exporter PRs, I've been adding some tests for exporter so that we don't break existing use cases and we can cover future ones (#21029, #21042).
> 
> I think the PRs can be rebased now (the `main.rs` diff should apply to `lib.rs` now).

Thanks for the update, that's a great idea adding the test cases. I'll see if I can get some time this weekend to go through and update my still open PRs.

AeroXuk:
@kjarosh Sorry for the delay in getting back to this. I have rebased all the PRs against master now that the logic is moved into `lib.rs`

PR #20933, opened 2025-07-06
tests: Fix avm1/array_sort_random

jarca0123:


tests/tests/swfs/avm1/array_sort_random/test.toml
```diff
@@ -1,2 +1,2 @@
 num_frames = 1
-known_failure = true
+known_failure = false
```
kjarosh: You can remove this altogether

core/src/avm1/globals/array.rs
```diff
@@ -746,7 +746,7 @@ fn qsort<'gc>(
             // Find an element greater than the pivot from the left.
             while left <= high {
                 let (_, item) = &elements[left];
-                if compare_fn(activation, &pivot, item, options)?.is_le() {
+                if left >= right || compare_fn(activation, &pivot, item, options)?.is_le() {
```
kjarosh: I think this can be simplified; you can replace the while condition with `left < right`, because `right` is always less or equal to `high`.

It's the same semantically—we don't need to go into the `right` territory, because all elements are greater than the pivot there.

If I'm understanding it right, it won't affect the final order of the elements, but will prevent an additional comparison, right? In the case of `avm1/array_sort_random` it changes the final order only because the comparison function is being fed from a PRNG? (I'm talking specifically about stability of the sort.)

jarca0123: As for the while condition, you're right.

As for the additional comparison, you're also right. The condition is needed because it prevents a comparison that is not present in Flash Player output.

kjarosh: Perfect, thanks! I also think you can remove the other condition from the if, as the while condition covers all cases

PR #20934, opened 2025-07-06
tests: make avm1/from_shumway/array pass

SuchAFuriousDeath:


core/src/avm1/object/script_object.rs
```diff
@@ -146,7 +146,7 @@ impl<'gc> Object<'gc> {
     pub(super) fn get_data(
         &self,
         name: AvmString<'gc>,
-        activation: &mut Activation<'_, 'gc>,
+        activation: &Activation<'_, 'gc>,
```
kjarosh: Changes from this file seem unrelated to this PR?

SuchAFuriousDeath: Actually it is not, requiring a mutable reference makes my iterator approach not work because of the borrow checker :(

SuchAFuriousDeath: At least I think that's why I changed it.

kjarosh: Huh, we usually pass `&mut Activation` everywhere, so requiring non-mutable activation here may be problematic in the future. I'll take a look at it

SuchAFuriousDeath: I think there should be a way around it, I'll take a look after work.

moulins: I think this should be fixable by creating the `ArrayBuilder` beforehand:
```rs
let builder = ArrayBuilder::new(activation);
let elements_iter = ...:
Ok(builder.with(elements_iter).into())
```

SuchAFuriousDeath: I tried that, the issue isn't about the array builder, it's some other crazy borrow checker stuff.

danielhjacobs: Could you maybe do a `for` loop though `elements_intermediate` with a match arm instead of doing `into_iter` with a `flat_map` and `map_either`?

SuchAFuriousDeath: Yeah, that's probably the best option here

torokati44:
@SuchAFuriousDeath just a small technical note: When updating your branches with PRs opened from them, please stick to the "rebase" option instead of the "merge" one. Merge commits aren't allowed on the `master` branch, so sooner or later the branches will have to be rebased anyway, and there's a (small) chance that merge conflicts will happen then, even if merging was successful.

And in the meantime, it makes the "graphical" git history a bit less easy on the eye:
image

core/src/avm1/globals/array.rs
```diff
@@ -439,29 +440,46 @@ pub fn concat<'gc>(
...
+        .flat_map(|desc| {
+            desc.map_either(
+                // Map the object and its length to an iterator of its elements
+                |(object, length)| (0..length).map(|index| object.get_element(activation, index)),
+                // Map the value itself to an iterator
+                |&value| std::iter::once(value),
+            )
+        });
+
+    Ok(ArrayBuilder::new(activation).with(elements_iter).into())
```
danielhjacobs: I think this may work to avoid the mutability changes:

```rs
    let mut elements = vec![];

    for desc in elements_intermediate {
        match desc {
            Either::Left((object, length)) => {
                for i in 0..length {
                    elements.push(object.get_element(activation, i));
                }
            }
            Either::Right(&value) => {
                elements.push(value);
            }
        }
    }

    Ok(ArrayBuilder::new(activation).with(elements).into())
```

SuchAFuriousDeath: It would, I just wanted to avoid unnecessary allocation.

moulins (approved):

PR #20958, opened 2025-07-10
core: Finish porting `MovieClip` away from `GcCell`

moulins:
All display objects are now plain `Gc`s, which will allow future refactoring towards using `HasPrefixField` to handle the inheritance tree (as was done for AVM2 `Object`s)

### DIFF
--- a/core/src/display_object/movie_clip.rs
+++ b/core/src/display_object/movie_clip.rs
@@ -672,7 +672,7 @@ impl<'gc> MovieClip<'gc> {
     }
 
     pub fn stop(self, context: &mut UpdateContext<'gc>) {
-        self.0.write(context.gc()).stop(context)
+        self.0.read().stop(context)
     }
 
     /// Does this clip have a unload handler
@@ -1425,7 +1425,7 @@ impl<'gc> MovieClip<'gc> {
         // TODO: Move this to UpdateContext to avoid allocations.
         let mut goto_commands: Vec<GotoPlaceObject<'_>> = vec![];
 
-        self.0.write(context.gc()).stop_audio_stream(context);
+        self.0.read().stop_audio_stream(context);
 
         let is_rewind = if frame <= self.current_frame() {
             // Because we can only step forward, we have to start at frame 1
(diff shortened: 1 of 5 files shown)


core/src/display_object/movie_clip.rs
```diff
@@ -181,67 +179,83 @@ pub struct MovieClipData<'gc> {
...
+#[collect(no_drop)]
+struct MovieClipDataMut<'gc> {
+    base: InteractiveObjectBase<'gc>,
+    container: ChildContainer<'gc>,
+    frame_scripts: Vec<Option<Avm2Object<'gc>>>,
     avm1_text_field_bindings: Vec<Avm1TextFieldBinding<'gc>>,
+
+    /// List of tags queued up for the current frame.
+    #[collect(require_static)]
+    queued_tags: HashMap<Depth, QueuedTagList>,
```
Lord-McSweeney: Can this be moved into `MovieClipData` so that it can be accessed without using `Write::assume`?

moulins: It could, but that would mean an extra `RefCell`.

core/src/display_object/movie_clip.rs
```diff
@@ -337,41 +351,44 @@ impl<'gc> MovieClip<'gc> {
         is_root: bool,
         loader_info: Option<LoaderInfoObject<'gc>>,
     ) {
-        let mut mc = self.0.write(context.gc());
+        let mc = Gc::write(context.gc(), self.0);
```
Lord-McSweeney: Can this be called `write` instead of `mc`, since `mc` usually refers to a `Mutation`?

core/src/display_object/movie_clip.rs
```diff
@@ -1231,15 +1231,16 @@ impl<'gc> MovieClip<'gc> {
 
         // It is now safe to update the tag position and frame number.
         // TODO: Determine if explicit gotos override these or not.
-        let read = self.0.read();
+        let read = self.0;
```
Lord-McSweeney: Can this `let read = self.0;` be moved down or removed now?

Lord-McSweeney (approved):

PR #20959, opened 2025-07-10
core: Allow loading fonts into memory lazily

kjarosh:
This patch improves memory footprint by preventing fonts from being loaded eagerly into memory. Instead, it's possible to pass a file reference to a font, which will be used to load data from the font file when needed. This mechanism is used on desktop.

This patch also introduces a preloading mechanism, that makes sure we read from the font file once per a text fragment, and not for each glyph.


kjarosh:
Converted to draft, as I think using mmap will be a better solution

kjarosh:
Closing in favor of https://github.com/ruffle-rs/ruffle/pull/20969

PR #20965, opened 2025-07-11
desktop: Prefer non-sRGB surface formats

torokati44:
This is ~~a half measure~~ to unblock https://github.com/ruffle-rs/ruffle/pull/20898 (following https://github.com/emilk/egui/pull/7311) - the colors are probably still not right if only sRGB formats are supported.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1322,7 +1322,7 @@ checksum = "d8b14ccef22fc6f5a8f4d7d768562a182c04ce9a3b3157b91390b52ddfdf1a76"
 [[package]]
 name = "ecolor"
 version = "0.31.1"
-source = "git+https://github.com/emilk/egui.git?branch=main#78a8de2e8f8d6da82680af4701cfa6e61cfe389c"
+source = "git+https://github.com/emilk/egui.git?branch=main#f46926aaf1ea7ab5a3756c678a1d647b9e549bd6"
 dependencies = [
  "bytemuck",
  "emath",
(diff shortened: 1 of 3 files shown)


torokati44:
If you have any ideas/capacity, @Dinnerbone, I'd appreciate you helping out (or even taking over, as you please).

torokati44:
I am now on the opinion that this hasn't even worked correctly before with both sRGB and linear surface formats - the difference is just less noticeable when the mismatch is one way than when in the other (when it's darkening vs. when it's lightening).

torokati44:
See, wrt. https://github.com/ruffle-rs/ruffle/pull/21730:

| egui version / surface format | sRGB | linear |
| -------- | -------- | ------- |
| current | current_egui_srgb |current_egui_linear |
| updated | new_egui_srgb | new_egui_linear |

(Please excuse the window shadow in one of the screenshots...)
(The text may look different because of https://github.com/emilk/egui/pull/7431?)

torokati44:
So maybe this should be included in https://github.com/ruffle-rs/ruffle/pull/21730 instead of being separate...?

torokati44:
_actually_, since this has nothing to do with wgpu 26, we _could_ switch to the egui 0.32.3 release first, and have this fix included for it, (and then later, if we want wgpu 26, switch back to a later egui master, but that's an independent story).

torokati44:
Okay, added an `egui` bump to just after https://github.com/emilk/egui/pull/7311 (but before their wgpu 26 update or the text kerning change) in here, so there shouldn't be a gamma-mismatched, regressed, intermediate state.

PR #20969, opened 2025-07-11
desktop: Use mmap for fonts

kjarosh:
This patch uses mmap to load font files instead of storing them all in memory.

### DIFF
--- a/desktop/src/backends/ui.rs
+++ b/desktop/src/backends/ui.rs
@@ -15,6 +15,7 @@ use ruffle_core::backend::ui::{
     FullscreenError, LanguageIdentifier, MouseCursor, UiBackend,
 };
 use ruffle_core::{FontFileData, FontQuery};
+use std::fs::File;
 use std::path::Path;
 use std::rc::Rc;
 use std::sync::Arc;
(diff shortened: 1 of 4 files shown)


kjarosh:
Unfortunately, stumbled upon https://github.com/rust-lang/rust-clippy/issues/15252 in this PR :/

desktop/src/backends/ui.rs
```diff
@@ -429,7 +437,9 @@ fn load_fontdb_font(name: String, face: &FaceInfo) -> Result<FontDefinition<'sta
                 name,
                 is_bold,
                 is_italic,
-                data: bin.as_ref().as_ref().to_vec(),
+                // TODO remove when https://github.com/rust-lang/rust-clippy/issues/15252 is fixed
+                #[allow(clippy::unnecessary_to_owned)]
+                data: FontFileData::new(bin.as_ref().as_ref().to_vec()),
```
moulins: Couldn't this simply do `FontFileData(bin)`? (assuming we make the `FontFileData` field public) It would nicely avoid a copy, too.

kjarosh: Right, done!

desktop/src/backends/ui.rs
```diff
@@ -403,15 +404,22 @@ fn load_font_from_file(
...
-    match std::fs::read(path) {
-        Ok(data) => Ok(FontDefinition::FontFile {
-            name,
-            is_bold,
-            is_italic,
-            data,
-            index,
-        }),
-        Err(e) => Err(anyhow!("Couldn't read font file at {path:?}: {e}")),
+    match File::open(path) {
```
evilpie: Maybe use map_err here and below?

desktop/src/backends/ui.rs
```diff
@@ -403,15 +404,22 @@ fn load_font_from_file(
...
+        Ok(file) => {
+            let mmap = unsafe { memmap2::Mmap::map(&file) };
```
evilpie: I guess I would mention that we usually re-parse the whole file, so modification of the mapped file is less problematic?

evilpie (approved):

desktop/src/backends/ui.rs
kjarosh: Done

desktop/src/backends/ui.rs
kjarosh: Done, looks a lot better now

PR #20982, opened 2025-07-12
avm2: Handle negative indexes in array.lastIndexOf

SuchAFuriousDeath:


core/src/avm2/globals/array.rs
```diff
@@ -455,10 +455,17 @@ pub fn last_index_of<'gc>(
     if let Some(array) = this.as_array_storage() {
         let search_val = args.get_value(0);
         let from = args.get_i32(activation, 1)?;
+        let len = array.length();
```
kjarosh: It can be inlined

core/src/avm2/globals/array.rs
```diff
@@ -456,9 +456,15 @@ pub fn last_index_of<'gc>(
         let search_val = args.get_value(0);
         let from = args.get_i32(activation, 1)?;
 
+        let from_index = if from >= 0 {
+            from as usize
+        } else {
+            array.length().saturating_sub(-from as usize)
```
kjarosh, line 462: Also you can do `saturating_add(from)` instead of subtracting negative

SuchAFuriousDeath: from is an i32, so I don't think I can

kjarosh: Alright, that's fine

kjarosh (approved):

PR #20983, opened 2025-07-12
core: Make `Character: Copy`

moulins:
This allows returning them without having to keep a `&MovieLibrary` reference alive.


core/src/display_object/movie_clip.rs
```diff
@@ -4208,15 +4197,16 @@ impl<'gc, 'a> MovieClip<'gc> {
                                         .library
                                         .library_for_movie_mut(movie.clone());
```
Lord-McSweeney, line 4131: I think this extra `library_for_movie_mut` call can be skipped now, and the library declared above the match can be used instead?

moulins: It's prevented by the `BitmapClass::from_class_object(class_object, activation.context)` just above, which can't be easily refactored out.

adrian17: The other refactors I'm attempting (making `Library` into a `Gc`) will allow refactoring this more

adrian17 (approved):

PR #20986, opened 2025-07-12
core: Cross Domain security checks using the crossdomain.xml

evilpie:


n0samu:
I believe CORS on the modern web serves the same purpose that crossdomain.xml served for the Flash plugin, so I'm not sure it makes sense to enforce crossdomain.xml policies by default. Especially because our users have not had to worry about it for all this time, so many of them are probably not using the proper crossdomain.xml policy file anymore, meaning if we suddenly started enforcing it we would break their sites. Additionally, any mistake we make will break resource loading on some people's sites, which is a problem that people often find uniquely difficult to diagnose and debug. And there are a lot of sources of mistakes:
- Any misunderstanding of how the policies work
- Any mistake in choosing the default values
- Any incompleteness in our implementation
- Any difference in our XML parser, including handling different character encodings

evilpie:
This would not be enabled by default.

n0samu:
Okay thanks for clarifying ❤️ carry on!

PR #20998, opened 2025-07-14
avm2: Add infallible methods to get and set dynamic object properties

Lord-McSweeney:


core/src/avm2/amf.rs
```diff
@@ -313,11 +313,20 @@ pub fn deserialize_value_impl<'gc>(
...
+                // the array instead of setting a dynamic property.
+                if let Ok(index) = name.parse::<usize>() {
+                    array.set_element(activation.gc(), index, value);
+                } else {
+                    array.set_dynamic_property(
+                        AvmString::new_utf8(activation.gc(), name),
+                        value,
+                        activation.gc(),
+                    );
+                }
```
moulins: Isn't this wrong if `name` is `"length"`? Also, are we sure that `.parse::<usize>()` is correct here? (e.g., for `"001"`)

I think it would be best to move this into a dedicated method on `ArrayObject`.

Lord-McSweeney: On `master` this calls `set_property_local`, which doesn't handle `length`, only ScriptObject dynamic properties and array elements. This code is directly copied from `ArrayObject::set_property_local`- I agree that it would be better to have the string-to-usize parsing logic factored out somewhere, but I don't see a good place to do it, as `ArrayObject::set_property_local` handles multinames, not local names.

moulins: Why not add a `ArrayObject::as_array_index(name: &WStr) -> Option<usize>` method and use that?
It still duplicates the `set_element/set_dynamic_property` dichotomy, but at least the parsing logic is in its proper place.

core/src/avm2/amf.rs
```diff
@@ -316,16 +316,17 @@ pub fn deserialize_value_impl<'gc>(
                 let name = element.name();
                 let value = deserialize_value_impl(activation, element.value(), object_map)?;
 
-                // If the name of the element was numerical, we set an element on
-                // the array instead of setting a dynamic property.
-                if let Ok(index) = name.parse::<usize>() {
+                // FIXME: Can we avoid this allocation in the case of a valid
+                // index while still reusing the index-parsing logic from
+                // `ArrayObject::as_array_index`?
+                let name = AvmString::new_utf8(activation.gc(), name);
```
moulins: If `ArrayObject::as_array_index` took a `&WStr`, we could write:
```rs
let name = wstr::from_utf8(name);
if let Some(index) = ArrayObject::as_array_index(name) {
  // ...
} else {
  let name = AvmString::new(activation.gc(), name);
  // ...
}
```

Lord-McSweeney: This should be fixed now

PR #21002, opened 2025-07-14
avm2: fix coercion to function error for Vector.some and Array.some

SuchAFuriousDeath:
Completes #12321.

Also factored out some code into a separate function for the 1006 and 1034 errors as a bonus.

### DIFF
--- a/core/src/avm2/filters.rs
+++ b/core/src/avm2/filters.rs
@@ -1,4 +1,4 @@
-use crate::avm2::error::{make_error_2008, type_error};
+use crate::avm2::error::{make_error_1034, make_error_2008};
 use crate::avm2::globals::flash::display::shader_job::get_shader_args;
 use crate::avm2::globals::slots::flash_filters_bevel_filter as bevel_filter_slots;
 use crate::avm2::globals::slots::flash_filters_blur_filter as blur_filter_slots;
@@ -168,13 +168,11 @@ impl FilterAvm2Ext for Filter {
             )?));
         }
 
-        Err(Error::avm_error(type_error(
+        Err(make_error_1034(
             activation,
-            &format!(
-                "Error #1034: Type Coercion failed: cannot convert {object:?} to flash.filters.BitmapFilter."
-            ),
-            1034,
-        )?))
+            object,
+            "flash.filters.BitmapFilter",
+        ))
     }
 
     fn as_avm2_object<'gc>(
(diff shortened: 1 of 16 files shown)


Molisson:
This might fix https://github.com/ruffle-rs/ruffle/issues/12321.

SuchAFuriousDeath:
> This might fix #12321.

It does, that's the point! :D

core/src/avm2/filters.rs
```diff
@@ -498,23 +496,20 @@ fn avm2_to_displacement_map_filter<'gc>(
...
-        object.get_slot(displacement_map_filter_slots::MAP_BITMAP)
-    {
-        if let Some(bitmap) = bitmap.as_bitmap_data() {
-            Some(bitmap.bitmap_handle(activation.gc(), activation.context.renderer))
+    let map_bitmap =
+        if let Value::Object(bitmap) = object.get_slot(displacement_map_filter_slots::MAP_BITMAP) {
+            if let Some(bitmap) = bitmap.as_bitmap_data() {
+                Some(bitmap.bitmap_handle(activation.gc(), activation.context.renderer))
+            } else {
+                return Err(make_error_1034(
```
Lord-McSweeney: This `else` can't be reached because the slot is typed `BitmapData`, so the error can be replaced with an `unreachable!` or similar

core/src/avm2/filters.rs
```diff
@@ -168,13 +168,11 @@ impl FilterAvm2Ext for Filter {
             )?));
         }
 
-        Err(Error::avm_error(type_error(
```
Lord-McSweeney, line 171: In FP, attempting to allocate a custom class that extends BitmapFilter throws an ArgumentError, so if we implement the correct behavior, this error call should also be unreachable

core/src/avm2/filters.rs
SuchAFuriousDeath: Hmmm. So unwrap?

core/src/avm2/globals/vector.rs
```diff
@@ -262,55 +262,21 @@ pub fn join<'gc>(
...
-        if result {
-            return Ok(true.into());
-        }
-    }
-
-    Ok(false.into())
-}
+delegate_method_to_array!(every, super::array::every);
+delegate_method_to_array!(_some, super::array::_some);
+delegate_method_to_array!(for_each, super::array::for_each);
```
Lord-McSweeney: Since these all use `super::array`, is it possible for it to be moved inside the macro? `$method` can be an `ident`, I think, and the method body can use `super::array::$method`?

SuchAFuriousDeath: I'll try

SuchAFuriousDeath: Yup, looks like it's possible!

core/src/avm2/globals/Array.as
```diff
@@ -204,5 +206,7 @@ package {
         AS3 native function unshift(... rest):uint;
 
         public static const length:int = 1;
+
+        private native function _some(callback:Function, receiver:Object):Boolean;
```
Lord-McSweeney: Can this `_some` declaration be moved right below the declaration of `AS3 function some`?

Molisson: (Vectors too ^)

SuchAFuriousDeath: Done. The vector ones were already moved yesterday.

Molisson (approved):

PR #21003, opened 2025-07-14
avm2: fix localeCompare behaviour for swf versions <= 11

SuchAFuriousDeath:


core/src/avm2/globals/string.rs
```diff
@@ -176,7 +176,17 @@ pub fn locale_compare<'gc>(
 ) -> Result<Value<'gc>, Error<'gc>> {
     let this = this.coerce_to_string(activation)?;
 
-    let other = args.get_string(activation, 0)?;
+    let other_value = args.get_value(0);
+
+    if activation.caller_movie_or_root().version() <= 11 {
+        match other_value {
```
Lord-McSweeney, line 183: Can this use two nested if-statements rather than a match? Or at least do the following?
```
Value::Null | Value::Undefined => {
    if this.is_empty() {
        ...
    } else {
        ...
    }
}
```

SuchAFuriousDeath: Sure

SuchAFuriousDeath: Done

core/src/avm2/globals/string.rs
```diff
@@ -177,7 +177,22 @@ pub fn locale_compare<'gc>(
 ) -> Result<Value<'gc>, Error<'gc>> {
     let this = this.coerce_to_string(activation)?;
 
-    let other = args.get_string(activation, 0)?;
+    let other_value = args.get_value(0);
+
+    if activation.caller_movie_or_root().version() <= 11 {
```
Lord-McSweeney: Are you sure `activation.caller_movie_or_root().version() <= 11` is correct? In my version of FP, the special-case for null and undefined only activates in SWFv9 and SWFv10 (so `activation.caller_movie_or_root().version() <= 10`).

SuchAFuriousDeath: Done

adrian17 (approved):

PR #21014, opened 2025-07-16
avm2: Implement Dxns and DxnsLate opcodes

jarca0123:


core/src/avm2.rs
```diff
@@ -818,6 +823,25 @@ impl<'gc> Avm2<'gc> {
...
+        self.default_xml_namespace_stack.pop();
+    }
+
+    /// Get the current default XML namespace.
+    /// Returns an empty string if no default XML namespace is set.
+    pub fn current_default_xml_namespace(&self, strings: &StringContext<'gc>) -> AvmString<'gc> {
+        self.default_xml_namespace_stack
+            .last()
+            .copied()
+            .unwrap_or_else(|| AvmString::new_ascii_static(strings.gc(), b""))
```
adrian17: If this was working on Activation instead, you could just do `istr!("")` here

moulins: You can specify the StringContext manually with `istr!(strings, "")`, too.

core/src/avm2.rs
```diff
@@ -119,6 +119,10 @@ pub struct Avm2<'gc> {
     /// Scopes currently present of the scope stack.
     scope_stack: Vec<Scope<'gc>>,
 
+    /// Default XML namespace stack for E4X operations.
+    /// The last element is the current default namespace.
+    default_xml_namespace_stack: Vec<AvmString<'gc>>,
+
```
moulins: Why use a stack here? There's no `PopDxns` opcode, so a single `default_xml_namespace: Option<AvmString<'gc>>` on `Activation` would be enough.

moulins:
Note: the AVM2 specification says that `dxns/dxnslate` opcodes should be rejected at verification time if the `SET_DXNS` method flag isn't set. (relevant avmplus code: https://github.com/adobe/avmplus/blob/master/core/Verifier.cpp#L1169)

adrian17:
I don't think the stack is correct here; what FP does _kinda_ resembles the stack, in that when you call a function and set dxns, when you exit from the function the callee has its previous `dxns` back. But this alone could be implemented by just setting `dxns` as a field on the Activation - and maybe this is enough for an MVP?

But in reality, in FP this is more complex unfortunately, in avmplus (see `MethodFrame::findDxns`) it's a property of MethodFrames/MethodEnvs (kinda what our `BoundMethod`s are)/scopes. It recurses MethodFrames... presumably because _sometimes_ the current frame might inherit dxns from the parent? But in practice, with normal function calls, a function doesn't inherit and just uses its own default namespace so not sure what that's about.

The easily visible difference is that the default namespace of the `MethodEnv` depends on the environment (scope chain?) at the point in time when it was created, see the example:

```as
function f(){
    var a = <xml xmlns:a="a" xmlns:b="b"><data x="1" a:x="2" b:x="3">asdf</data></xml>;

    trace(a.data.@x); // 1

    var inline_f1 = function(){ trace(a.data.@x); };

    default xml namespace = "a";
    trace(a.data.@x); // 12

    var inline_f2 = function(){ trace(a.data.@x); };

    default xml namespace = "b";
    trace(a.data.@x); // 13

    inline_f1(); // 1
    inline_f2(); // 12

    create_custom_function("")(a); // 1
    create_custom_function("a")(a); // 12
    create_custom_function("b")(a); // 13
}
function create_custom_function(nsname){
    default xml namespace = nsname;
    return function(a){ trace(a.data.@x); };
};
```

But again, I don't exactly understand how the chained MethodFrame iteration in avmplus relates to this, I don't know how to observe it.
For MVP, maybe just setting `dxns` on the current Activation would be enough? (and make a failing test for the more complex behavior)

kjarosh:
Superseded by https://github.com/ruffle-rs/ruffle/pull/22761

PR #21017, opened 2025-07-16
avm1: Move the global execution linked list from `DisplayObjectBase` to `MovieClip`

moulins:
Only AVM1 MovieClips can execute AVM1 framescripts, so there's no reason to put non-`MovieClip` display objects in the AVM1 exec list, and no reason to have it defined as part of `DisplayObjectBase`.

### DIFF
--- a/core/src/display_object/edit_text.rs
+++ b/core/src/display_object/edit_text.rs
@@ -2134,7 +2134,7 @@ impl<'gc> EditText<'gc> {
     }
 
     /// Construct the text field's AVM1 representation.
-    fn construct_as_avm1_object(self, context: &mut UpdateContext<'gc>, _run_frame: bool) {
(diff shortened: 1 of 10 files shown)


core/src/display_object/edit_text.rs
```diff
@@ -2134,7 +2134,7 @@ impl<'gc> EditText<'gc> {
     }
 
     /// Construct the text field's AVM1 representation.
-    fn construct_as_avm1_object(self, context: &mut UpdateContext<'gc>, run_frame: bool) {
+    fn construct_as_avm1_object(self, context: &mut UpdateContext<'gc>, _run_frame: bool) {
```
Lord-McSweeney: Since it's unused, can the `run_frame` parameter be removed?

moulins: Good catch!

PR #21018, opened 2025-07-17
avm2: Implement Math methods in Number

jarca0123:
Also fixes some tests.

### DIFF
--- a/core/src/avm2/globals/Number.as
+++ b/core/src/avm2/globals/Number.as
@@ -36,6 +36,78 @@ package {
         [API("680")]
         public static const LOG10E:Number = 0.4342944819032518;
 
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function abs(x:Number):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function acos(x:Number):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function asin(x:Number):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function atan(x:Number):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function atan2(y:Number, x:Number):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function ceil(x:Number):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function cos(x:Number):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function exp(x:Number):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function floor(x:Number):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function log(x:Number):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function max(x:Number = NEGATIVE_INFINITY, y:Number = NEGATIVE_INFINITY, ...rest):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function min(x:Number = POSITIVE_INFINITY, y:Number = POSITIVE_INFINITY, ...rest):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function pow(x:Number, y:Number):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function random():Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function round(x:Number):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function sin(x:Number):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function sqrt(x:Number):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function tan(x:Number):Number;
+
         {
             prototype.toExponential = function(digits:* = 0):String {
                 var self:Number = this;
--- a/core/src/avm2/globals/math.rs
+++ b/core/src/avm2/globals/math.rs
@@ -5,6 +5,7 @@ use crate::avm2::error::type_error;
 use crate::avm2::object::Object;
 use crate::avm2::value::Value;
 use crate::avm2::{ClassObject, Error};
+use num_traits::ToPrimitive;
 use rand::Rng;
 
 macro_rules! wrap_std {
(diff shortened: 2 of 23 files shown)


Lord-McSweeney:
For `round`, `ceil`, and `floor`, I think we'd prefer to avoid the path that generates Integers instead of Numbers and instead add the logic to `impl From<f64> for Value` later so that it can apply everywhere.

core/src/avm2/globals/Number.as
```diff
@@ -36,6 +36,82 @@ package {
...
+        [Ruffle(FastCall)]
+        public static native function log(x:Number):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function max(x:Number = NegInfinity, y:Number = NegInfinity, ...rest):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function min(x:Number = PosInfinity, y:Number = PosInfinity, ...rest):Number;
```
Lord-McSweeney: Just `Infinity` seems to work in the declarations of the functions in the `Math` class, what happens if you change this line to
```
        public static native function min(x: Number = Infinity, y: Number = Infinity, ...rest):Number;
```
?

core/src/avm2/globals/math.rs
```diff
@@ -107,20 +110,44 @@ pub fn min<'gc>(
...
 pub fn pow<'gc>(
-    _activation: &mut Activation<'_, 'gc>,
+    activation: &mut Activation<'_, 'gc>,
     _this: Value<'gc>,
     args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
-    let n = args[0].as_f64();
-    let p = args[1].as_f64();
+    let n = args[0].coerce_to_number(activation)?;
+    let p = args[1].coerce_to_number(activation)?;
```
Lord-McSweeney: I think these two lines can still use `as_f64`, as the argument is already coerced to a Number by the ActionScript-side function signature

core/src/avm2/globals/math.rs
```diff
@@ -136,3 +163,65 @@ pub fn random<'gc>(
...
+    }
+}
+
+pub fn ceil<'gc>(
+    _activation: &mut Activation<'_, 'gc>,
+    _this: Value<'gc>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let input = args[0];
+    let num = input.as_f64();
```
Lord-McSweeney: nit: this can just be
```
let input = args[0].as_f64();
```

core/src/avm2/globals/math.rs
```diff
@@ -90,7 +91,9 @@ pub fn max<'gc>(
         let val = arg.coerce_to_number(activation)?;
         if val.is_nan() {
             return Ok(f64::NAN.into());
-        } else if val > cur_max {
+        } else if val > cur_max
+            || (val == cur_max && !val.is_sign_negative() && cur_max.is_sign_negative())
```
kjarosh, line 95: That's an oddly specific condition. Why does it have to be here? Can't it be somehow generalized? (Similarly the `min` one.)

core/src/avm2/globals/math.rs
```diff
@@ -107,20 +110,44 @@ pub fn min<'gc>(
...
+    activation: &mut Activation<'_, 'gc>,
     _this: Value<'gc>,
     args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
-    let n = args[0].as_f64();
-    let p = args[1].as_f64();
+    let n = args[0].coerce_to_number(activation)?;
+    let p = args[1].coerce_to_number(activation)?;
+
+    // If y is NaN, the result is NaN.
```
kjarosh: Comments refer to values as `x` and `y` whereas code refers to them as `n` and `p`. This should be unified somehow (it's better to use the names from docs if possible I guess).

core/src/avm2/globals/math.rs
```diff
@@ -107,20 +110,44 @@ pub fn min<'gc>(
...
+    if p.is_nan() {
+        return Ok(f64::NAN.into());
+    }
+
+    // Special case: If x is ±1 and y is ±Infinity, the result is NaN.
```
kjarosh: Is this case covered by existing tests? `1^inf` seems like it should return 1.

core/src/avm2/globals/math.rs
```diff
@@ -136,3 +163,65 @@ pub fn random<'gc>(
...
 }
+
+pub fn abs<'gc>(
+    _activation: &mut Activation<'_, 'gc>,
+    _this: Value<'gc>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let input = args[0];
+    match input {
+        Value::Integer(i) => Ok(i.abs().into()),
```
kjarosh, line 160: Why do we have dedicated `Integer` cases? How is that observable from AS?

core/src/avm2/globals/math.rs
jarca0123: It is covered by from_avmplus/as3/Types/Number/pow. I can add tests for Math.pow(int(1), Number.POSITIVE_INFINITY), which also returns NaN.

core/src/avm2/globals/math.rs
jarca0123: There is a pushbyte instruction in the test from_avmplus/as3/Types/Number/abs, which (if I understand it correctly) pushes a byte (and therefore an integer) to the abs function. And Ruffle does seem to separate Value::Number and Value::Integer.

core/src/avm2/globals/math.rs
jarca0123: It handles signed zero cases. In max, positive zero takes priority, while in min, the negative zero takes priority. Rust cannot compare positive zero with negative zero, so that's why it is needed.

core/src/avm2/globals/math.rs
jarca0123: However, the test verifies (when 1 is passed or when Number(-0.0) is passed), if it returns int. In other cases, it returns Number.

adrian17: I don't think this should need a special handling per-method. Just like @Lord-McSweeney said, there should probably be a "if the float is integral*, generate a Value::Integer" in `impl From<f64> for Value` instead. (for avmplus reference, see `AvmCore::doubleToAtom` which does exactly that)

(*and within `1<<28` range)

IMO adding missing methods to Number and adding new logic to core Value should go into separate PRs?

core/src/avm2/globals/math.rs
```diff
@@ -66,7 +64,10 @@ pub fn round<'gc>(
     // Note that Flash Math.round always rounds toward infinity,
     // unlike Rust f32::round which rounds away from zero.
     let ret = (x + 0.5).floor();
-    Ok(ret.into())
```
Lord-McSweeney, line 70: This should probably also return `Ok(ret.into())` directly- as Adrian said, it'd be better to keep the Number -> int optimization a general optimization and implement it in a separate PR

core/src/avm2/globals/math.rs
kjarosh: It's fine then, just making sure it's intended

kjarosh:
All of the code from this PR should be merged already, see https://github.com/ruffle-rs/ruffle/pull/21222, https://github.com/ruffle-rs/ruffle/pull/21346, https://github.com/ruffle-rs/ruffle/pull/21933

Thanks @jarca0123 for working on this!

PR #21025, opened 2025-07-18
chore: Bump `on-headers` node package to `1.1.0` fully

torokati44:
By running `npm audit fix` in `web`.

Dependabot almost managed to do this in https://github.com/ruffle-rs/ruffle/pull/21024.


torokati44:
See also: https://github.com/ruffle-rs/ruffle/security/dependabot/109

PR #21027, opened 2025-07-18
avm2: Match avmplus's array index parsing

Lord-McSweeney:


moulins (approved):

core/src/avm2/object/bytearray_object.rs
```diff
@@ -200,23 +200,6 @@ impl<'gc> TObject<'gc> for ByteArrayObject<'gc> {
         self.base().init_property_local(name, value, activation)
     }
 
-    fn delete_property_local(
```
Lord-McSweeney, line 203: `ByteArray` is a sealed class, so `delete_property_local` was unreachable anyway

PR #21029, opened 2025-07-18
exporter: Add integration tests

kjarosh:
Integration tests allow testing most of the exporter binary at once
(except CLI args parsing). The consist of test directories, each one
defines one test with input and output files. Files are compared
byte-by-byte, except for images, which are compared pixel-by-pixel.


Dinnerbone:
Neat! But, I think I'd like at least one test that uses `args`? Or any option... A single test with an empty file doesn't look great 😅 

Maybe also document the toml format as we do for the normal tests today

kjarosh:
> I think I'd like at least one test that uses `args`? Or any option... A single test with an empty file doesn't look great 😅

Yeah, sorry for that, I planned to merge the easily conflicting parts and then write some proper tests.

Should look better now!

PR #21030, opened 2025-07-18
avm2: Fix some Date methods to pass tests

SuchAFuriousDeath:


kjarosh:
Nit: it sounds like you're fixing date tests, but in fact you're fixing the date implementation to make tests pass

SuchAFuriousDeath:
Superseded by #21082 after discovering the underlying cause.

PR #21047, opened 2025-07-19
pixelbender: Remove Result usage, minor code improvements

SuchAFuriousDeath:
It looks like the usage of Result in that file was completely for nothing, no function ever returned an Err anywhere :sweat_smile:
I have even more refactors in mind, but I wanted to make the PR reviewable.

### DIFF
--- a/render/wgpu/src/pixel_bender.rs
+++ b/render/wgpu/src/pixel_bender.rs
@@ -557,7 +557,9 @@ pub(super) fn run_pixelbender_shader_impl(
 
                 let value_vec = match value {
                     PixelBenderType::TFloat(f1) => FloatOrInt::Float(smallvec![*f1, 0.0, 0.0, 0.0]),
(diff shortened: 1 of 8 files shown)


render/wgpu/src/pixel_bender.rs
```diff
@@ -541,8 +542,8 @@ pub(super) fn run_pixelbender_shader_impl(
 
                 #[derive(Debug)]
                 enum FloatOrInt {
-                    Float(Vec<f32>),
-                    Int(Vec<i32>),
+                    Float(SmallVec<[f32; 16]>),
+                    Int(SmallVec<[i32; 16]>),
```
kjarosh: I think it makes sense to use `SmallVec<[_; 4]>` as most of the time we're operating on `rgba`.

kjarosh: And in case of `Int`, we won't be getting more than 4 at all.

SuchAFuriousDeath: Done.

PR #21052, opened 2025-07-20
avm2: Reduce `TObject` boilerplate

moulins:
Some straightforward (but large) `TObject` refactorings giving a big reduction in lines of code.

- Move downcasting methods to `Object`, and define them with a macro;
  - also remove some unecessary `obj.as_foo().unwrap()` roundtrips.
- Replace all manual implementations of `as_ptr` by a default impl using `gc_base`;
- Define the `WeakObject` enum with a macro.


core/src/avm2/object/bitmapdata_object.rs
```diff
@@ -93,24 +92,24 @@ impl<'gc> BitmapDataObject<'gc> {
 
         Ok(instance)
     }
+
+    pub fn get_data(self) -> BitmapDataWrapper<'gc> {
```
Lord-McSweeney: Can these methods stay called `bitmap_data` and `init_bitmap_data`? IMO `get_data` and `init_data` sound too non-specific.

Lord-McSweeney (approved):

core/src/avm2/object.rs
```diff
@@ -711,318 +715,178 @@ pub trait TObject<'gc>: 'gc + Collect<'gc> + Debug + Into<Object<'gc>> + Clone +
...
-        None
+impl<'gc> Object<'gc> {
+    pub fn ptr_eq<T: TObject<'gc>>(a: T, b: T) -> bool {
+        std::ptr::eq(a.as_ptr(), b.as_ptr())
     }
 
-    fn as_loader_info_object(&self) -> Option<&LoaderInfoObject<'gc>> {
-        None
+    impl_downcast_methods! {
+        pub fn as_class_object for ClassObject;
```
Lord-McSweeney, line 751: Does the visibility really matter for these functions? Can't they just all be `pub`?

core/src/avm2/globals/flash/display/bitmap_data.rs
```diff
@@ -144,7 +144,9 @@ pub fn init<'gc>(
     };
 
     new_bitmap_data.init_object2(activation.gc(), this);
-    this.init_bitmap_data(activation.gc(), new_bitmap_data);
+    if let Some(bitmap_data_obj) = this.as_bitmap_data_object() {
```
adrian17: Could this be an `.expect()` instead? It _should_ always be a BitmapDataObject here after all, right?

adrian17 (approved):

core/src/avm2/globals/flash/display/bitmap_data.rs
moulins: Hm, you're right, I'll change that.

Issue #20274, opened 2025-04-27
Zombotron II: Time Machine slow


nful2:
### Describe the bug

using ruffle-nightly-2025_04_27-linux-x86_64
seems to run too slowly otherwise runs perfectly no issues other than slowness.. application doesn't use 100% full of my system resources so there is no system resource problem
in the end that is my recording problem not of ruffle!

https://github.com/user-attachments/assets/2939605c-c047-4b34-a03b-ba06318e6c3c

### Expected behavior

should run smoothly with original speed..

### Content Location

http://zombotron.com/zombotron-two-time-machine.html
i downloaded the swf from the ruffle extension in firefox

### Affected platform

Desktop app

### Operating system

Arch Linux

Lord-McSweeney:
BitmapData operations seem to be the main reason for the slowdown, these two functions show up as highest on perf:
```
  19.18%  ruffle_desktop  [.] ruffle_core::bitmap::operations::copy_on_cpu                   
  15.21%  ruffle_desktop  [.] ruffle_core::bitmap::bitmap_data::BitmapData::pixels_rgba      
```

desertking:
> BitmapData operations seem to be the main reason for the slowdown, these two functions show up as highest on perf:
> 
> ```
>   19.18%  ruffle_desktop  [.] ruffle_core::bitmap::operations::copy_on_cpu                   
>   15.21%  ruffle_desktop  [.] ruffle_core::bitmap::bitmap_data::BitmapData::pixels_rgba      
> ```

any improvements for that in the pipe ?

PR #21053, opened 2025-07-20
Avoid copying pixels, minor improvements

SuchAFuriousDeath:
In theory, this should help with #20274, but doesn't seem to, it's probably hitting the padding branch anyway.


torokati44:
ouch, that last commit message... :no_mouth:

SuchAFuriousDeath:
Yeah :( I also noticed that

PR #21057, opened 2025-07-21
chore: Move duplicate dependencies to workspace

kjarosh:


torokati44:
For reference: https://github.com/ruffle-rs/ruffle/pull/15868 (last time the "rule of  ~three~ _four_" was applied)

evilpie (approved):

PR #21058, opened 2025-07-21
pixelbender: Implement `pbasm`, a PixelBender assembler

kjarosh:
`pbasm` is a PixelBender assembler that can be used to craft kernels.


kjarosh:
After https://github.com/ruffle-rs/ruffle/pull/21203, I'm thinking I can try moving it out of Ruffle to a separate repository. But I'm thinking about leaving the assembly here, in case we need that in debug tools. As always, discussion is welcome.

adrian17:
This can be closed now, right?

kjarosh:
Yes, see also https://github.com/ruffle-rs/pbasm/pull/1

PR #21066, opened 2025-07-22
pixelbender: Treat bool parameters as int parameters

kjarosh:
Flash Player does not treat boolean parameters differently than ints, a bool param is semantically the same as an int param.


render/src/pixel_bender.rs
```diff
@@ -49,10 +49,10 @@ pub enum PixelBenderType {
     TInt3(i16, i16, i16) = 0xA,
     TInt4(i16, i16, i16, i16) = 0xB,
     TString(String) = 0xC,
-    TBool(bool) = 0xD,
-    TBool2(bool, bool) = 0xE,
-    TBool3(bool, bool, bool) = 0xF,
-    TBool4(bool, bool, bool, bool) = 0x10,
+    TBool(i16) = 0xD,
```
Lord-McSweeney: Could you add a comment here to explain why TBool uses `i16` and not `bool`?

Lord-McSweeney (approved):

render/src/pixel_bender.rs
kjarosh: Done!

PR #21067, opened 2025-07-22
pixelbender: Add validation for conditional registers & add tests

kjarosh:
This PR adds tests for `if` and `select` opcode conditions and adds validation, which fails in case a float register is used as a condition.


core/src/avm2/globals/flash/display/shader_data.rs
```diff
@@ -25,7 +26,10 @@ pub fn init<'gc>(
 
     let bytecode = args.get_object(activation, 0, "bytecode")?;
     let bytecode = bytecode.as_bytearray().unwrap();
-    let shader = parse_shader(bytecode.bytes()).expect("Failed to parse PixelBender");
+    let shader = parse_shader(bytecode.bytes(), true).map_err(|err| {
+        tracing::debug!("Failed to parse a Pixel Bender shader: {err}");
```
Lord-McSweeney: nit: Can this be `tracing::warn` or `tracing::error` instead of `tracing::debug`?

Lord-McSweeney (approved):

core/src/avm2/globals/flash/display/shader_data.rs
kjarosh: Changed to warning, I think errors should indicate that something's wrong with Ruffle and not that an exception is being thrown in the content.

PR #21068, opened 2025-07-22
pixelbender: Handle EOF properly when parsing

kjarosh:


core/src/avm2/error.rs
```diff
@@ -625,6 +625,20 @@ pub fn make_error_2027<'gc>(activation: &mut Activation<'_, 'gc>, value: i32) ->
     }
 }
 
+#[inline(never)]
+#[cold]
+pub fn make_error_2030<'gc>(activation: &mut Activation<'_, 'gc>) -> Error<'gc> {
+    let err: Result<Value<'_>, Error<'_>> = error(
```
Lord-McSweeney: Does this need to explicitly specify the type? If not, can you remove it for consistency with the rest of this file?

kjarosh: Thanks for noticing that, sometimes I missclick and vscode adds the type automatically :/

Lord-McSweeney (approved):

PR #21071, opened 2025-07-22
pixelbender: Handle various errors around defining params

kjarosh:


render/src/pixel_bender.rs
```diff
@@ -359,18 +359,24 @@ pub struct PixelBenderShader {
...
         for param in &self.params {
             if let PixelBenderParam::Normal {
                 qualifier: PixelBenderParamQualifier::Output,
                 reg,
+                param_type,
                 ..
             } = param
             {
-                return Some(reg.channels.len());
+                channels = Some((reg, *param_type));
```
adrian17: why a mutable variable instead of `return Some((reg, *param_type))`?

kjarosh: Because Flash Player uses the last defined output parameter, but now I figured I can traverse it in the reversed direction, so I'll probably try doing that 😅

kjarosh: Done

PR #21073, opened 2025-07-22
avm2: Mark classes with a builtin type enum to make it possible to identify them without an Activation

Lord-McSweeney:
This also makes it a lot easier to unify `CallNode::GlobalInit` and `CallNode::Method`


core/src/avm2/activation.rs
```diff
@@ -1196,17 +1196,15 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
         if let Some(return_type) = return_type {
-            if return_type == class_defs.void {
+            if return_type.is_builtin_void() {
                 Value::Undefined
-            } else if return_type == class_defs.int
-                || return_type == class_defs.uint
-                || return_type == class_defs.number
+            } else if return_type.is_builtin_int()
+                || return_type.is_builtin_uint()
+                || return_type.is_builtin_number()
```
moulins: This check appears often enough that it may be worth a dedicated `.is_numeric()` method.

core/src/avm2/optimizer/optimize.rs
```diff
@@ -79,54 +79,48 @@ impl<'gc> OptValue<'gc> {
...
-            return true;
+            true
+        } else {
+            // Primitives are always not-null
+            self.class.is_some_and(|c| {
+                c.is_builtin_int()
+                    || c.is_builtin_uint()
+                    || c.is_builtin_number()
+                    || c.is_builtin_boolean()
+                    || c.is_builtin_void()
```
moulins: This check appears often enough that it may be worth a dedicated `.is_primitive()` method.

core/src/avm2/class.rs
```diff
@@ -1078,6 +1086,37 @@ impl<'gc> Class<'gc> {
...
+    }
+    pub fn is_builtin_object(self) -> bool {
+        matches!(self.0.builtin_type.get(), Some(BuiltinType::Object))
+    }
+    pub fn is_builtin_string(self) -> bool {
+        matches!(self.0.builtin_type.get(), Some(BuiltinType::String))
+    }
+    pub fn is_builtin_void(self) -> bool {
+        matches!(self.0.builtin_type.get(), Some(BuiltinType::Void))
+    }
```
moulins: Nit: why not just `is_int`, `is_uint`, etc?

core/src/avm2/class.rs
```diff
@@ -1114,3 +1153,19 @@ impl<'gc> Class<'gc> {
...
+pub enum BuiltinType {
+    // `int`, `uint`, `Number`, `Boolean`, `Object`, `String`, `void` builtin
+    // classes
+    Int,
+    Uint,
+    Number,
+    Boolean,
+    Object,
+    String,
+    Void,
```
moulins: Nit: grouping all primitives types together (by putting `Void` just after `Boolean`) would lead to better codegen for primitive checks.

moulins (approved):

core/src/avm2/class.rs
Lord-McSweeney: The names seemed too vague to me without the `builtin` word- if you think they sound fine, I'll change it

core/src/avm2/script.rs
```diff
@@ -410,12 +410,7 @@ impl<'gc> Script<'gc> {
 
         let init = unit.load_method(script.init_method, false, activation)?;
 
-        // Script initializers cannot have parameters declared
```
adrian17, line 413: is this not observable?

Lord-McSweeney: This was my mistake, I forgot to increase the script initializer's `max_locals` when testing this

PR #21077, opened 2025-07-23
pixelbender: Do not panic on non-0 padding

kjarosh:
Flash Player ignores padding bytes when parsing Pixel Bender shaders.


render/src/pixel_bender.rs
```diff
@@ -403,7 +418,7 @@ const CHANNELS: [PixelBenderRegChannel; 7] = [
     PixelBenderRegChannel::M4x4,
 ];
 
-fn read_src_reg(val: u32, size: u8) -> Result<PixelBenderReg, Box<dyn std::error::Error>> {
+fn read_src_reg(val: u32, size: u8) -> Result<PixelBenderReg> {
```
SuchAFuriousDeath: Is there a reason to return a Result here at all?

render/src/pixel_bender.rs
```diff
@@ -428,7 +443,7 @@ fn read_matrix_reg(val: u16, mask: u8) -> PixelBenderReg {
     read_reg(val, vec![CHANNELS[(mask + 3) as usize]])
 }
 
-fn read_dst_reg(val: u16, mask: u8) -> Result<PixelBenderReg, Box<dyn std::error::Error>> {
+fn read_dst_reg(val: u16, mask: u8) -> Result<PixelBenderReg> {
```
SuchAFuriousDeath: Same here

render/src/pixel_bender.rs
kjarosh: Removed the Result in the dependent PR

render/src/pixel_bender.rs
kjarosh: Here as well

render/src/pixel_bender.rs
SuchAFuriousDeath: It even looks like there's a clippy lint for this: https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_wraps
It's opt-in, maybe it would bully us too much, but it's something to consider

Lord-McSweeney (approved):

PR #21081, opened 2025-07-23
chore: transition from #[allow(...)] to expect #[expect(...)]

SuchAFuriousDeath:


web/src/lib.rs
```diff
@@ -429,8 +429,12 @@ impl RuffleHandle {
         // Instance is dropped at this point.
     }
 
-    #[allow(clippy::boxed_local)] // for js_bind
-    pub fn call_exposed_callback(&self, name: &str, args: Box<[JsValue]>) -> JsValue {
+    // for js_bind
```
kjarosh: What does that comment mean? :thinking: It looks like it was meant for the allow

SuchAFuriousDeath: Yup. It's for the lint. I moved it.

PR #21082, opened 2025-07-23
avm2: Make Date.prototype an instance of Date

adrian17:
Replaces https://github.com/ruffle-rs/ruffle/pull/21030 .

Hopefully same approach could be also used to implement the same behavior for other classes doing this: `Function, MethodClosure, Error, Array, RegExp`. (note: `MethodClosure` is absent in Ruffle)

### DIFF
--- a/core/src/avm2/object/date_object.rs
+++ b/core/src/avm2/object/date_object.rs
@@ -61,6 +61,29 @@ impl<'gc> DateObject<'gc> {
         Ok(instance)
     }
 
+    pub fn for_prototype(
+        activation: &mut Activation<'_, 'gc>,
+        date_class: ClassObject<'gc>,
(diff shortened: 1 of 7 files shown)


core/src/avm2/object/date_object.rs
```diff
@@ -61,6 +61,29 @@ impl<'gc> DateObject<'gc> {
         Ok(instance)
     }
 
+    pub fn for_prototype(
+        activation: &mut Activation<'_, 'gc>,
+        date_class: ClassObject<'gc>,
+    ) -> Result<Object<'gc>, Error<'gc>> {
```
Lord-McSweeney: nit: This function never returns Err

Lord-McSweeney (approved):

PR #21083, opened 2025-07-23
chore: Update translations

RuffleBuild:
New [Crowdin](https://crowdin.com/project/ruffle) pull request with translations 🎉


desktop/packages/linux/rs.ruffle.Ruffle.metainfo.xml
```diff
@@ -228,6 +228,7 @@
         <keyword xml:lang="pl">odtwarzacz</keyword>
         <keyword xml:lang="pt-BR">reprodutor</keyword>
         <keyword xml:lang="pt-PT">leitor</keyword>
+        <keyword xml:lang="ro">cititor</keyword>
```
kjarosh: I'm not sure if that's the proper translation...

danielhjacobs: I think it should just be player: https://ro.wikipedia.org/wiki/Media_player

torokati44:
@colin-i If you are 100% sure that "cititor" is the right Romanian word for what Ruffle is, feel free to add it back - I just deleted those translations so this PR can be merged (see discussion above).

colin-i:
@torokati44 I wrote at crowdin at the comment about player/viewer/reproducer/reader . Then I changed there. Now I see this reply. I'm also happy with "reproducător". I don't think is important to add it back.

torokati44:
Alright, thank you!

PR #21087, opened 2025-07-23
tests: Port AS3 math avmplus tests

kjarosh:
This PR adapts 7 new tests from avmplus repo.

Failing tests:
* `from_avmplus/as3/Math/e15_8_2_11_rest`
* `from_avmplus/as3/Math/e15_8_2_12_rest`
* `from_avmplus/as3/Math/e15_1_2_2`


tests/tests/swfs/from_avmplus/as3/Math/e15_1_2_2/Test.as
```diff
@@ -0,0 +1,34 @@
...
+
+
+var testcases = getTestCases();
+
+function getTestCases() {
+    var array = new Array();
+    var item = 0;
+
+
+    // other js engines produce 1228503247364293600 for this testcase, and c produces 1228503247364293699
```
kjarosh, line 25: I'm not sure what to do with this case, as it fails in Flash Player (both on Linux and Windows), but passes in Ruffle :thinking:

kjarosh: Sticked with the FP output, left for someone to analyze what's going on

PR #21088, opened 2025-07-23
tests: Port AS3 RegExp tests from avmplus

kjarosh:
This PR adapts 4 new tests from avmplus repo.

Failing tests:
* `from_avmplus/as3/RegExp/replace`


kjarosh:
Those tests are interesting, because some of them are failing in FP.

CC @evilpie, you may find this interesting too

evilpie:
Interesting. I hadn't see `(?P=mygroup)` before. I suspect we don't support it.

evilpie (approved):

PR #21097, opened 2025-07-24
avm2: Don't special-case calling script initializers

Lord-McSweeney:
Also:
- Class initializers are now correctly run in interpreter mode, not JIT mode
- We pass `Option<FunctionObject>` for callee objects instead of passing a `Value`, which makes it obvious when `exec` is being provided a valid callee


core/src/avm2/object/class_object.rs
```diff
@@ -327,7 +327,22 @@ impl<'gc> ClassObject<'gc> {
     ) -> Result<Value<'gc>, Error<'gc>> {
         let scope = self.0.instance_scope.get();
         let method = self.init_method();
+
         if let Some(method) = method {
+            // Provide a callee object if necessary
+            let callee = if method.needs_arguments_object() {
```
adrian17, line 333: I'm guessing the `callee` doesn't change from call to call, so it should be stored on the class object?

Lord-McSweeney: It will change from call to call, as the FunctionObject needs to be bound to the `receiver`

core/src/avm2.rs
```diff
@@ -295,32 +297,42 @@ impl<'gc> Avm2<'gc> {
...
-        if let Err(e) = init_result {
-            // If the script initializer fails verification, we still need
-            // to properly dispose of the created stack frame.
-            init_activation.cleanup();
-            return Err(e);
+        let (method, global_object, domain) = script.init();
+
+        let scope = ScopeChain::new(domain);
+        // Script `global` classes extend Object
+        let bound_superclass = Some(activation.avm2().classes().object);
```
moulins, line 307: Why not `let bound_superclass = bound_class.and_then(|c| c.super_class())`?

Lord-McSweeney: `bound_superclass` is actually an `Option<ClassObject>`, not an `Option<Class>`, so that doesn't work here

PR #21103, opened 2025-07-25
desktop: Add Steamworks API as an optional ExternalInterface

evilpie:
This is a new optional feature, which allows SWFs running inside Ruffle to use (a currently very limited) subset of the [Steamworks API](https://partner.steamgames.com/doc/sdk/api). To make this API easily usable from both AS3 and AS2, I decided to implement this as a "fake" ExternalInterface. The current API is heavily inspired by [steamworks.js](https://github.com/ceifa/steamworks.js/).

~~The biggest missing piece is of course callback support.~~

I am not sure which APIs to prioritize.

After compiling ruffle with `cargo build --features steamworks` and including the right directory you can use the API like this:

```actionscript
package {
    import flash.display.Sprite;
    import steamworks.Client;

        function Test() {
            var client: Client = new Client(480)

            trace("appId: " + client.utils.getAppId());

            trace("name: " + client.localPlayer.getName());
            trace("level: " + client.localPlayer.getLevel());
            trace("getIpCountry: " + client.localPlayer.getIpCountry());

            client.addEventListener(Client.USER_STATS_RECEIVED, function(): void {
                trace("USER_STATS_RECEIVED");
                trace("isActived('ACH_WIN_ONE_GAME'): " + client.achievement.isActivated('ACH_WIN_ONE_GAME'));

                trace(client.achievement.names());
            });
            client.localPlayer.requestUserStats();
        }
    }
}

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4495,6 +4495,7 @@ dependencies = [
  "ruffle_render_wgpu",
  "ruffle_video_external",
  "ruffle_video_software",
+ "steamworks",
  "sys-locale",
  "thiserror 2.0.17",
  "tokio",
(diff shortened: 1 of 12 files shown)


evilpie:
So, we now have support for the most important event `userStatsReceived`. This should make it possible to use this for some basic tasks.

I am still looking for feedback for someone who actually wants to use this API.

~We should probably move the ActionScript wrapper around the ExternalInterface somewhere into the tree as well?~
The ActionScript 3 steamworks API is now located in the desktop/steamworks folder.

Fancy2209:
Just a suggestion, it'd probably make sense to also expose SteamUserStats()->ClearAchievement(), it's especially useful for testing

evilpie:
> Just a suggestion, it'd probably make sense to also expose SteamUserStats()->ClearAchievement(), it's especially useful for testing

That should already exist.

Fancy2209:
I skimmed trough the code and somehow missed Achivement.clear(), my bad!

KupoGames:
I set this up and it mostly went smoothly, but Flash was throwing up some errors over a conflict between this Utils class and Flash's existing Utils class. So I changed all the Utils references to steamworks.Utils in client.as and it worked.

Managed to unlock some achievements on Steam! And my app can still run correctly without Steam running.

image

It's slightly annoying that Ruffle.exe won't run at all without steam_api64.dll in the folder (rather than just skipping it quietly), but this seems to be a fairly common thing with other game engines, and I often see that games on other platforms will still include that Steam file.

image

Fancy2209:
I'd say appart from Achivements, Steam Input could be a good one
Especially since Ruffle doesn't have GameInput currently

PR #21104, opened 2025-07-25
avm2: Add a buggy SWF<11 integer overflow in sort comparison

kjarosh:
* Requires https://github.com/ruffle-rs/ruffle/pull/21085

Flash Player had a bug in SWF<11 where the numeric comparison function overflowed for certain numbers.  This patch replicates this behavior.

Currently it only works for Flash Player 32-bit, as Ruffle does not support the 64-bit runtime.  On 64-bit the behavior differs a bit due to different integer representations.

The test `from_avmplus/as3/Array/regress_524122` is written in such a way that it passes for both 32- and 64-bit behaviors.


core/src/avm2/globals/array.rs
```diff
@@ -904,14 +904,49 @@ pub fn compare_string_case_insensitive<'gc>(
...
     let num_a = a.coerce_to_number(activation)?;
     let num_b = b.coerce_to_number(activation)?;
 
+    if COMPAT {
+        if let (Value::Integer(a), Value::Integer(b)) = (a, b) {
+            let a = (a << 3) | 6;
+            let b = (b << 3) | 6;
+            // for SWF<11, FP relies on the following overflow
+            let diff = a.wrapping_sub(b);
+            if diff > 0 {
```
SuchAFuriousDeath: This looks like diff.cmp(0), but maybe that's less readable.

kjarosh: You're right, no idea why I haven't thought about it!

PR #21107, opened 2025-07-26
core: Upgrade quick_xml and improve error handling when parsing xml

SuchAFuriousDeath:
EDIT: This PR was repurposed to upgrade quick-xml to version 38 and adjust the code to the breaking changes. 

After https://github.com/tafia/quick-xml/issues/926 and https://github.com/tafia/quick-xml/issues/927 are implemented and after https://github.com/tafia/quick-xml/pull/924 is included in a new release, some more tests can be made to pass.

The breaking change that was introduced is that Event::GeneralRef was added, forcing us to rewrite both avm1 and avm2 logic, because text and references are now provided as separate events, but we have to merge them into one node.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4078,13 +4078,13 @@ checksum = "19f132c84eca552bf34cab8ec81f1c1dcc229b811638f9d283dceabe58c5569e"
 
 [[package]]
 name = "plist"
-version = "1.7.3"
+version = "1.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "546b279bf0638ee811d9e47de2ca5b66575a543035d79fdf83959dd2f5c3b4c3"
+checksum = "092791278e026273c1b65bbdcfbba3a300f2994c896bd01ab01da613c29c46f1"
 dependencies = [
  "base64",
  "indexmap",
- "quick-xml 0.37.5",
+ "quick-xml",
  "serde",
  "time",
 ]
@@ -4261,18 +4261,9 @@ checksum = "a993555f31e5a609f617c12db6250dedcac1b0a85076912c436e6fc9b2c8e6a3"
 
 [[package]]
 name = "quick-xml"
-version = "0.37.5"
+version = "0.39.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "331e97a1af0bf59823e6eadffe373d7b27f485be8748f71471c662c1f269b7fb"
-dependencies = [
- "memchr",
-]
-
-[[package]]
-name = "quick-xml"
-version = "0.38.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b66c2058c55a409d601666cffe35f04333cf1013010882cec174a7467cd4e21c"
+checksum = "cdcc8dd4e2f670d309a5f0e83fe36dfdc05af317008fea29144da1a2ac858e5e"
 dependencies = [
  "memchr",
 ]
@@ -4578,7 +4569,7 @@ dependencies = [
  "fnv",
  "gc-arena",
  "hashbrown 0.16.1",
- "quick-xml 0.38.4",
+ "quick-xml",
  "regress",
  "ruffle_macros",
  "ruffle_wstr",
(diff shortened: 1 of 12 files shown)


SuchAFuriousDeath:
I think I managed to fix it, but it's ugly :(

SuchAFuriousDeath:
I will reimplement this when https://github.com/tafia/quick-xml/pull/924 is merged and released.

Cargo.lock
```diff
@@ -6349,7 +6357,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "54cb1e9dc49da91950bdfd8b848c49330536d9d1fb03d4bfec8cae50caa50ae3"
 dependencies = [
  "proc-macro2",
- "quick-xml",
+ "quick-xml 0.37.5",
```
torokati44: Too bad this is not released yet...
https://github.com/Smithay/wayland-rs/pull/833

SuchAFuriousDeath: It is now :)

torokati44:
lol, I accidentally reproduced basically the same diff just now, before I discovered this...

https://github.com/torokati44/ruffle/commit/5040b22e2f5ee851e7c8463ac8ca18ecba8a24ad

Would you care to take it over and pick and choose the better parts of each?
Looks like yours fixes a few more test cases.

SuchAFuriousDeath:
I'll try to get it ready today.

torokati44:
I think you could/should take `Cargo.lock` over from my commit as is.

SuchAFuriousDeath:
I'm thinking of bumping to 0.38 first so that the changes aren't so hectic.

core/src/xml_util.rs
torokati44, line 1: Does it make sense to add this here, into a whole new file? As it is, it could as well be a local non-pub fn, or even a lambda/closure, so a new Vec (with just the entity ref in it) won't even have to be allocated each call.

Or is it because you have future work that puts more stuff in this file?

core/src/avm1/globals/xml.rs
```diff
@@ -141,37 +141,61 @@ impl<'gc> Xml<'gc> {
...
-                        | IllFormedError::UnmatchedEndTag { .. },
-                    ) => XmlStatus::MismatchedEnd,
-                    quick_xml::Error::IllFormed(IllFormedError::MissingDeclVersion(_)) => {
-                        XmlStatus::DeclNotTerminated
-                    }
-                    quick_xml::Error::InvalidAttr(AttrError::UnquotedValue(_)) => {
-                        XmlStatus::AttributeNotTerminated
+            let event = match parser.read_event() {
+                Ok(event) => event,
+                Err(error) => {
```
torokati44, line 155: Couldn't this stay a `map_err` plus `?` at the end? 🤔

core/src/xml_util.rs
SuchAFuriousDeath: When I put this here, it was used in 3 places. Maybe now it's not. I asked for a re-review a bit too early, I still have to look work on this.

core/src/html/text_format.rs
```diff
@@ -713,12 +713,22 @@ impl FormatSpans {
 
         // Helper function to decode a byte sequence returned by quick_xml into a WString.
         // TODO: use Cow<'_, WStr>?
-        let decode_to_wstr = |raw: &[u8]| -> WString {
+        let decode_to_wstr = |raw: Cow<'_, [u8]>| -> WString {
```
torokati44, line 716: Was this change to potentially spare an allocation, and instead take ownership of the given buffer?

torokati44: (Ah I see it's because of `reconstruct_entity_ref` returns a new `Vec` anyway...)

core/src/xml_util.rs
torokati44: No, you're right, it's used in two different modules... :facepalm: Somehow I missed the HTML usage just now...

torokati44:
Now I'll admit this is probably nearing nitpick territory, but there's something strange about the tiny mismatches between avm1/avm2: `match &event` / `match event`, `pending_text`/`text_buf`, `let needs_flush`/inline `if !matches!`...
There are probably good reasons for all though!

torokati44:
https://rustsec.org/advisories/RUSTSEC-2026-0195 :eyes:

PR #21114, opened 2025-07-26
avm1, avm2: Simplify iteration logic

SuchAFuriousDeath:


core/src/avm1/globals/xml_node.rs
```diff
@@ -141,7 +141,7 @@ fn get_prefix_for_namespace<'gc>(
         for ancestor in node.ancestors() {
             // Iterate attributes by their definition order, so the first matching one
             // is returned.
-            for (key, value) in ancestor.attributes().own_properties() {
+            for (key, value) in ancestor.attributes().properties().enumerable() {
```
moulins: This is vulnerable to double-borrow errors, because `coerce_to_string` (used below) can call into arbitrary AS2 code through `toString/valueOf`, which could try to modify the attributes object during the property iteration.

Unfortunately I don't think there's a way to completely avoid an allocation here, the best I can see is having a temporary Vec that's reused across iterations of the top-level loop.

SuchAFuriousDeath: Hmmm. Any suggestions on how to deal with this?.

SuchAFuriousDeath: Is it even possible to use an iterator here in any way? Was the original comment deceptive?

SuchAFuriousDeath:
This could actually be a trait, something like 
```rs
trait Ancestors {
    fn ancestors(self) -> impl Iterator<Item = Self>;
}
```

That way we could implement it directly even for foreign types, like Gc<'gc, Scope<'gc>>, allowing it to be called directly with self.ancestors() or self.scope().ancestors() and so on. 

We would also not have to duplicate documentation and make it clear in one spot that the method yields self first and then the ancestors.

Lord-McSweeney (approved):

PR #21121, opened 2025-07-27
avm2: Fix Loader.content.loaderInfo and contentLoaderInfo compatibility

jarca0123:
This isn't probably the most elegant way to go about it, but it fixes the powerbook swf from discord.


kjarosh:
What does this PR do? It's not described, neither in the PR nor in the comments.

How does this relate to the comment in the code that objects return loaders from the avm root?

In order to get a PR merged you need to convince people that's the right patch, but there are no tests and no description, so you're leaving all this work (which is a significant part of landing a patch) on maintainers/reviewers. That's why there hasn't been a review for a long time.

PR #21122, opened 2025-07-27
chore: Mark all Pixel Bender tests as flaky

kjarosh:
Unfortunately, it seems that all Pixel Bender tests are flaky :/

According to the documentation, this should mark all tests that have `pixelbender` in their name as flaky.


torokati44:
My guess has been that it's rather a lavapipe (or corresponding SW graphics driver on other platforms) issue, as it doesn't _only_ happen with PB.

torokati44 (approved):

kjarosh:
Extended my speculation from the comment

PR #21124, opened 2025-07-27
avm2: Refactor dense arrays with a start offset

jarca0123:


kjarosh:
I think this patch needs some explanation

jarca0123:
Basically this replicates avmplus behavior, and in avmplus, if you set a value to a large array index in a dense array, then it will stay dense, while currently in Ruffle, it will become sparse. I'm not sure about the performance benefits though 😕

adrian17:
@kjarosh 

IIRC we talked about this when originally refactoring Arrays in https://github.com/ruffle-rs/ruffle/pull/16051 and just didn't see it as a priority; it's not an observable behavior aside from performance, and the cases where it'd actually help also felt rare (at least I wasn't aware of any swf hitting this).

But I also said we'd happily consider it in case they wanted to implement it (https://discord.com/channels/610531541889581066/614855396468719619/1231337881029841029) :D

adrian17:
Generally if it doesn't directly impact correctness or performance, "matches avmplus" by itself isn't _that_ much of a strong argument; but I'm sure it should be possible to do some benchmarks and show a noticeable memory/time improvement for a dense offset Array?

I can try measure on your branch tomorrow. Note that if the extra logic regresses the "common path" (non-offset dense arrays) too much, this could be a blocker too.

By the way, I think this could use some more tests, as in: interacting with significantly offset arrays in various ways, just as a sanity check?

kjarosh:
> I think this patch needs some explanation

Just for the record, I meant that we should add some context to this patch (to the commit message), as it's not trivial.

Lord-McSweeney:
Hmm, I don't actually think this is working? I was benchmarking with this PR and used the following to test it:
```
            var array:Array = [];
            array[999] = 0;
            array[1000] = 4;
            array[1001] = 5;
            while(counter < 700000) {
                array[1000];
                array[1001];
                array[1002] = 5;
                array[1003] = 6;
                counter++;
            }
```
and I still see `BTreeMap` functions included in perf. Is there a specific array operation that this is supposed to help?

jarca0123:
Oops. I've been focusing on fixing timeline bugs, which I think is more important. I'll currently draft this.

PR #21125, opened 2025-07-28
avm2: Minor fixes and improvements

Lord-McSweeney:


core/src/avm2/globals/VectorObject.as
```diff
@@ -1,9 +1,7 @@
 package __AS3__.vec {
     [Ruffle(CallHandler)]
     [Ruffle(InstanceAllocator)]
-    // FIXME: This class is supposed to be final, but then we can't create any
-    // Vector.<T> (since they all extend this class)
-    internal dynamic class Vector$object {
+    internal final dynamic class Vector$object {
```
adrian17, line 4: Hmm, do our is-final checks just happen to not trigger when instantiating `Vector.<T>`?

Lord-McSweeney: We only run the is-final check for i_classes loaded from the ABC (technically global scope classes too, but that's never final anyway). I'm not sure when it was changed to be this way, but yes, no class validation is run on created `Vector.<T>` classes.

core/src/avm2/class.rs
```diff
@@ -901,12 +903,7 @@ impl<'gc> Class<'gc> {
             // Methods, getters, and setters are forbidden from appearing
             // in activation traits
             if loaded_trait.as_method().is_some() {
-                // TODO: Is this the correct error?
-                return Err(Error::avm_error(verify_error(
-                    activation,
-                    "Error #1101: Cannot verify method with unknown scope.",
-                    1101,
-                )?));
+                return Err(make_error_1107(activation));
```
kjarosh, line 922: Can we make a test for it?

core/src/avm2/globals/flash/display3D/textures/atf_jpegxr.rs
```diff
@@ -24,14 +23,22 @@ pub fn do_compressed_upload<'gc>(
...
+            3679,
+        )?));
+    }
+
+    if is_cube != atf_texture.cubemap {
+        return Err(Error::avm_error(argument_error(
+            activation,
+            "Error #3675: Texture format mismatch.",
+            3675,
+        )?));
```
kjarosh, line 41: Similarly, can we add a test for these two?

kjarosh:
Hmmm, IMO we should cover all those cases you fixed in tests

Lord-McSweeney:
I'm not sure how much value there is in testing most of these changes, but I can write tests for three of the easily observable changes.

core/src/avm2/class.rs
Lord-McSweeney: I don't think I can write a test for this, as FFDEC refuses to save activation traits with methods in them

PR #21145, opened 2025-07-30
avm2: Use ParametersExt almost everywhere possible

Lord-McSweeney:
Combined with #21125, this unblocks switching native methods to take `FunctionArgs` instead of a slice of values, which unblocks calling native methods without allocating a buffer to hold the arguments (when possible).

Most of the ParametersExt functions no longer do any coercion, too, which should make calling some native methods slightly faster, and means that we don't need to pass an Activation to most of the ParametersExt functions.


core/src/avm2/parameters.rs
```diff
@@ -3,10 +3,30 @@ use crate::avm2::Object;
...
+///
+/// `get_value`: All parameter types work
+/// `get_f64`: `Number`, `int`, or `uint` type
+/// `get_i32`: `Number`, `int`, or `uint` type
+/// `get_u32`: `Number`, `int`, or `uint` type
+/// `get_bool`: `Boolean` type only
+/// `get_string` and family: `String` type only
+/// `get_object` and family: Any non-primitive type; i.e. any type *except* the following:
+///   - `*` (aka "any") type
+///   - `Object` type
```
kjarosh: Why not `*` and `Object`? I thought that `Object` is kinda the point of this method

kjarosh: Oh, in the sense that we might get *other* types that the function will panic on? It might be useful to include that piece of information there

core/src/avm2/parameters.rs
```diff
@@ -3,10 +3,30 @@ use crate::avm2::Object;
...
+
 /// Extensions over parameters that are passed into AS-defined, Rust-implemented methods.
 ///
 /// It is expected that the AS signature is correct and you only operate on values defined from it.
 /// These values will be `expect()`ed to exist, and any method here will panic if they're missing.
+///
+/// The rules for ActionScript type coercion may be surprising. Here is a table mapping
+/// ParametersExt functions to the corresponding ActionScript types:
+///
+/// `get_value`: All parameter types work
```
kjarosh: nit: you might want to link to the methods here to make sure they exist, e.g. ``[`ParametersExt::get_value`]``

core/src/avm2/parameters.rs
```diff
@@ -20,168 +40,82 @@ pub trait ParametersExt<'gc> {
...
     /// Gets the value at the given index as an Object. It is expected that the
     /// value is either Object or Null.
     ///
     /// If the value is null, None is returned.
-    fn try_get_object(
-        &self,
-        activation: &mut Activation<'_, 'gc>,
-        index: usize,
-    ) -> Option<Object<'gc>>;
+    fn try_get_object(&self, index: usize) -> Option<Object<'gc>> {
```
kjarosh, line 55: I'm not sure if `try_` is the right prefix here... I'd expect that a method prefixed with `try_` will *not* panic. Maybe `get_object_nullable`?

kjarosh, line 55: I would argue that the term `object` in this method is also a bit misleading. Primitives are objects too. Have you thought about something like `get_non_primitive`?

(On the other hand it returns an `Object`, so I don't know...)

core/src/avm2/parameters.rs
```diff
@@ -20,168 +40,82 @@ pub trait ParametersExt<'gc> {
...
+            Value::Null => None,
+            Value::Object(o) => Some(o),
+            _ => panic!("Expected Object or null as parameter"),
+        }
+    }
 
     /// Gets the value at the given index and coerces it to an f64.
     ///
     /// If the value is null or is undefined, 0.0 is returned.
     /// If the object cannot be coerced to an f64, a TypeError 1050 is raised.
```
kjarosh: This doc isn't up to date now

core/src/avm2/parameters.rs
```diff
@@ -20,168 +40,82 @@ pub trait ParametersExt<'gc> {
...
-    /// If the value is null or is undefined, Ok(None) is returned.
-    /// If the object cannot be coerced to a string, a TypeError 1050 is raised.
-    fn try_get_string(
-        &self,
-        activation: &mut Activation<'_, 'gc>,
-        index: usize,
-    ) -> Result<Option<AvmString<'gc>>, Error<'gc>>;
-}
+    /// Gets the uint-typed value at the given index.
+    fn get_u32(&self, index: usize) -> u32 {
```
kjarosh, line 71: Have you thought about naming these methods `get_uint` and `get_int` respectively? I think it makes more sense to refer to AS types than to Rust types in their names.

core/src/avm2/parameters.rs
```diff
@@ -20,168 +40,82 @@ pub trait ParametersExt<'gc> {
...
-
-    fn get_string(
-        &self,
-        activation: &mut Activation<'_, 'gc>,
-        index: usize,
-    ) -> Result<AvmString<'gc>, Error<'gc>> {
-        self[index].coerce_to_string(activation)
+    /// Like `try_get_string`, but returns "null" for null values instead
+    /// of returning `None`.
+    fn get_string(&self, activation: &mut Activation<'_, 'gc>, index: usize) -> AvmString<'gc> {
```
kjarosh, line 104: It looks like the number of references to `get_string` and `get_string_non_null` has the same order of magnitude for both (~70 vs ~25), which means that it's reasonable to check what the behavior should be for each use case and not assume one or the other.

In this case maybe it's better to add a suffix to this method in order to make sure we don't assume that it should be used?

core/src/avm2/globals/array.rs
```diff
@@ -1035,9 +1035,12 @@ pub fn sort<'gc>(
     // FIXME avmplus does some manual argument count/type checking here,
     // should we try to match that?
     let (compare_fnc, options) = if args.len() > 1 {
+        let compare_fnc = args.get_value(0);
+        let options = args.get_value(1).coerce_to_u32(activation)?;
+
         (
-            Some(args.get_value(0)),
-            SortOptions::from_bits_truncate(args.get_u32(1) as u8),
```
kjarosh: IIUC, the first commit changes the behaviors of these methods and the succeeding commits fix the caller codes? In this case I don't think we should do it this way, as Ruffle will be broken between commits (which is an issue for bisecting, etc.). I'd go with either changing this code _before_ changing the semantics, or making commits per method and doing all refactors relating to a method at once.

If Ruffle is broken between commits from the same PR, there's no reason to split the PR into commits in the first place.

core/src/avm2/parameters.rs
Lord-McSweeney: In our and avmplus's terminology, "Object" means an `Object` pointer, not the ActionScript `Object` type. For example, `Value::as_object` gets the Object variant of value only.

core/src/avm2/parameters.rs
Lord-McSweeney: I prefer the Rust type names, and they match other names in the codebase (`as_f64`, `as_i32`, `as_u32`, `coerce_to_i32`, etc).

core/src/avm2/parameters.rs
Lord-McSweeney: I think the return type and documentation of the method make it clear enough. This was also already an issue; many places that use `get_string` should be using `get_string_non_null`, so I'd prefer to keep fixing it to a different PR.

core/src/avm2/parameters.rs
Lord-McSweeney: I changed the description to explain that `Object` can represent any primitive type except `undefined`

core/src/avm2/parameters.rs
Lord-McSweeney: This was the behavior of the method before this PR too- the "try" is only because it returns an Option instead of a Result. I think the documentation is clear enough that it expects to be passed only `Value::Object` or `Value::Null`.

kjarosh:
The first commit fails to compile with
```
error[E0599]: no method named `get_optional` found for reference `&[avm2::value::Value<'gc>]` in the current scope
    --> core/src/avm2/globals/array.rs:1046:36
     |
1046 |     } else if let Some(arg) = args.get_optional(0) {
     |                                    ^^^^^^^^^^^^
     |
help: there is a method `get_bool` with a similar name
     |
1046 -     } else if let Some(arg) = args.get_optional(0) {
1046 +     } else if let Some(arg) = args.get_bool(0) {
     |
```

(Unless we want to squash those commits, then it doesn't matter; but having a lot of commits when Ruffle cannot even be compiled between them doesn't make sense)

core/src/avm2/parameters.rs
kjarosh: > This was the behavior of the method before this PR too

I know that, I thought we could improve that when making such a refactor.

> I think the documentation is clear enough

The documentation is useful as long as the method name is adequate, but of course we can discuss this refactor outside this PR

kjarosh: I guess the bottomline is, we're referring to Rust, not AS nomenclature here

core/src/avm2/parameters.rs
kjarosh: > I think the return type and documentation of the method make it clear enough

I agree, but I'm referring to something else, i.e. which method is preferred to be used in new code.

> This was also already an issue; many places that use `get_string` should be using `get_string_non_null`

So this is yet another reason for not using a generic name like `get_string` :D

Of course doing that in a separate PR is fine

Lord-McSweeney:
Good catch, that should be fixed now

PR #21150, opened 2025-07-31
ci: Also run Rust tests on Windows on ARM

torokati44:
Re: https://github.com/ruffle-rs/ruffle/discussions/21146

Let's see what happens...

### DIFF
--- a/tests/tests/swfs/avm2/pixelbender_dithering/test.toml
+++ b/tests/tests/swfs/avm2/pixelbender_dithering/test.toml
@@ -4,7 +4,7 @@ num_ticks = 1
 
 [image_comparisons.output]
 tolerance = 1
-max_outliers = 19929
+max_outliers = 20078
 
 [player_options]
 with_renderer = { optional = false, sample_count = 1 }
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -8,3 +8,8 @@ test(avm2/stage3d_errors_atf) or \
 test(avm1/netstream_play_flv) or \
 test(avm1/netstream_play_flv_screen)"""
 retries = 4
+
+# On Windows aarch64 target: skip H264 tests
+[[profile.ci.overrides]]
+platform = 'aarch64-pc-windows-msvc'
+default-filter = 'not test(visual/video/h264)'
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -64,7 +64,7 @@ jobs:
       fail-fast: false
       matrix:
         rust_version: [stable]
-        os: [ubuntu-24.04, ubuntu-24.04-arm, windows-2025, macos-15]
+        os: [ubuntu-24.04, ubuntu-24.04-arm, windows-2025, windows-11-arm, macos-15]
         include:
           - rust_version: nightly
             os: ubuntu-24.04
@@ -307,7 +307,7 @@ jobs:
     strategy:
       matrix:
         rust_version: [stable]
-        os: [ubuntu-24.04, ubuntu-24.04-arm, windows-2025, macos-15]
+        os: [ubuntu-24.04, ubuntu-24.04-arm, windows-2025, windows-11-arm, macos-15]
         include:
           - rust_version: nightly
             os: ubuntu-24.04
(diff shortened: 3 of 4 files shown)


torokati44:
Also referencing: https://github.com/ruffle-rs/ruffle/pull/19259

torokati44:
Ouch, OpenH264 only added Windows ARM builds with 2.5.0, which is blocked on https://github.com/ruffle-rs/ruffle/pull/18581 (or https://github.com/ruffle-rs/ruffle/pull/19445, but eh). See also: https://github.com/cisco/openh264/issues/3873

torokati44:
Does this mean the whole thing is running on x86 emulation? :thinking: 
https://github.com/ruffle-rs/ruffle/actions/runs/16645092011/job/47103772107?pr=21150#step:7:114

lexcyn:
> Does this mean the whole thing is running on x86 emulation? 🤔 https://github.com/ruffle-rs/ruffle/actions/runs/16645092011/job/47103772107?pr=21150#step:7:114

Hmm, strange as the rust checks show the correct aarch64 host - are there any artifacts/output from that build test? I could test it on my ARM64 system

torokati44:
That step is just setting up the nextest test runner.

kjarosh:
Tests on Windows ARM take even longer than on Windows x86_64 :/ despite the fact that some of them are disabled

It's **15m 47s** ARM vs **12m 41s** x64 vs **4–5min** Linux / Mac

torokati44:
That could also be because nothing is cached for it yet.

tests/tests/swfs/avm2/pixelbender_dithering/test.toml
```diff
@@ -4,7 +4,7 @@ num_ticks = 1
 
 [image_comparisons.output]
 tolerance = 1
-max_outliers = 19929
+max_outliers = 20078
```
danielhjacobs, line 7: Both before and after this PR, just legitimately curious, with this many outliers how useful is this test really?

kjarosh: 20078/(1280 × 720 × 4) = 0.54% outliers

adrian17:
Can we not run this on each PR, perhaps? Only daily? (btw this question applies to more CI types, not just ARM)
Or maybe at least skip a batch of tests, like `from_avmplus`, which are <50% (but still growing fast)?

kjarosh:
> btw this question applies to more CI types, not just ARM

To be fair, only the longest job negatively impacts PR checks performance, which as of now is windows-latest, and possibly windows-arm in the future.

> Or maybe at least skip a batch of tests

I tried marking slow tests one day, but that was a failure. Maybe there's some other mechanism than features that we can use for filtering slow tests

torokati44:
Referencing: https://github.com/ruffle-rs/ruffle/pull/15950#issuecomment-3167541986

torokati44:
Yes, it's really `-2025` and `-11-arm`... :roll_eyes: 
https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

kjarosh:
1. Do we want to try filtering slow tests again for this use case?

2. What will happen if auto merge is enabled and a non-reauired check is still running?

torokati44:
1. My suggestion is no, because tests that don't run automatically and frequently, effectively don't exist. So, for this particular (admittedly not very popular - at the moment) arch/OS combo, we would have a lot less tests. Granted, at the moment, according to this philosophy, we have none at all.

2. The docs say all _required_ checks have to pass before auto-merge takes effect. (And this is what makes logical sense to me as well.)
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request#about-auto-merge
  Once it's merged, I think the remaining non-required checks still running for (and in the context of) the PR are cancelled (I think I saw this happening a few times before), and an entirely new, complete suite is started, for the pushing of the "master" branch. Those are then all run to completion, and block nothing - but still take up overall runner slots while they are running, probably making any PRs made/rebased immediately after, wait a bit longer for some of its checks to even be started.

PR #21176, opened 2025-08-02
core: Use `HasPrefixField` and interior mutability for `DisplayObjectBase`

moulins:
Follow-up to #21098, by finishing the port of the `DisplayObject` hierarchy to `Gc`-based dispatch.

Has modest code-size gains, with 32ko taken off the desktop release binary.


core/src/display_object.rs
```diff
@@ -523,43 +534,49 @@ impl<'gc> DisplayObjectBase<'gc> {
...
 
-    fn set_filters(&mut self, filters: Vec<Filter>) -> bool {
-        if filters != self.filters {
-            self.filters = filters;
+    fn set_filters(&self, filters: Box<[Filter]>) -> bool {
+        let changed;
+        {
+            let mut write = self.cell.borrow_mut();
+            changed = filters != write.filters;
+            write.filters = filters;
```
Lord-McSweeney: Can this be changed to call `drop(write)` instead of using a temporary scope? It's a little shorter and more consistent with other code.

Lord-McSweeney (approved):

torokati44:
Oh-oh: `evaluation panicked: assertion failed: align_of::<Self>() == align_of::<Inner>()` at https://github.com/ruffle-rs/ruffle-android/actions/runs/16782578059/job/47524928150#step:7:30

kjarosh:
Another regression suspected of being caused by this PR: https://github.com/ruffle-rs/ruffle/issues/21249

torokati44:
Is that also fixed by https://github.com/ruffle-rs/ruffle/pull/21238?

n0samu:
> Is that also fixed by #21238?

@torokati44 Yes, it is.

PR #21202, opened 2025-08-04
ci: Switch Windows GHA runners to windows-2025

torokati44:
This is more just to see whether anything will fail a month from now...

See: https://github.blog/changelog/2025-07-31-github-actions-new-apis-and-windows-latest-migration-notice/#windows-latest-image-label-migration


kjarosh:
(We need to update required checks though.)

PR #21207, opened 2025-08-04
avm2: Use `Lock/RefLock` instead of `GcCel`l in `Domain`

moulins:
The only remaning `GcCell` user is now `BitmapDataWrapper`.


core/src/avm2/domain.rs
```diff
@@ -46,18 +43,38 @@ struct DomainData<'gc> {
     /// to `None`. It is only optional to avoid an order-of-events problem in
     /// player globals setup (we need a global domain to put globals into, but
     /// that domain needs the bytearray global)
-    pub domain_memory: Option<ByteArrayObject<'gc>>,
+    domain_memory: Lock<Option<ByteArrayObject<'gc>>>,
+
+    default_domain_memory: Lock<Option<ByteArrayObject<'gc>>>,
```
Lord-McSweeney: Can this field be a OnceLock?

PR #21217, opened 2025-08-05
chore: Resolve new clippy manual_is_multiple_of lint

danielhjacobs:


video/software/src/decoder/vp6.rs
```diff
@@ -70,7 +70,7 @@ impl Vp6Decoder {
 }
 
 fn crop(data: &[u8], mut width: usize, to_size: (u16, u16)) -> Vec<u8> {
-    debug_assert!(data.len() % width == 0);
+    debug_assert!(data.len().is_multiple_of(width));
```
danielhjacobs, line 73: If `width == 0`, this would have previously panicked. `.is_multiple_of` never panics, which I assume doesn't matter, but I thought it worth noting this subtle behavioral change: https://github.com/rust-lang/rust-clippy/issues/15335.

It means that unlike before, the `debug_assert` is `true` if `data.len()` is 0 and so is `width`, though I guess in that case this will panic just one line later with `data.len() / width`

kjarosh:
Note that it looks like we have a lot of other instances like that

danielhjacobs:
Odd that the new lint didn't catch any others then.

kjarosh:
Oh, it's probably because this method is not available for all types, only for unsigned ones

danielhjacobs:
> If you have a real usecase for this function on signed integer types, let us know! That was not included for simplicity, but if a usecase comes up, adding it will be considered.

https://github.com/rust-lang/rust/issues/128101

Issue #8336, opened 2022-10-20
AVM1 doesn't seem to allow negative zero in floating point values


moulins:
I've been doing some experimentation with Flash floating point values, and it seems that AVM1 has no way of actually creating a `-0.0` value.

For example, running this code in Flash Player prints `Infinity`.
```as
trace(1/(-1 * 0));
```
Whereas the "correct" IEEE754-compliant result (and what Ruffle prints) is `-Infinity`.

Here's a test-case which tries creating negative zero in a few different ways: 
[no_negative_zeros.zip](https://github.com/ruffle-rs/ruffle/files/9824919/no_negative_zeros.zip)

-------

AFAIK, the only potential way to distinguish `±0.0` in AVM1 is to invert them into `±Infinity`, so I see two possible solutions to this issue:
- Filter out negative zeros when converting a raw `f64` into a `Value` (this would add a check after almost all elementary floating point opcodes);
- Allow negative zeros inside `Value`, and make sure that no operation that can distinguish them can happen (I think the only one available directly from inside AVM1 is division, but we'd need to make sure that no native computations (e.g. matrix inverses) care about signed zeros).

moulins:
Another way of distinguishing signed zeroes, courtesy of @relrelb:
```as
// only in the second argument, not the first.
trace(Math.atan2(0, 0)); // 0
trace(Math.atam2(0, -0)); // 3.141592653589793
```

PR #21223, opened 2025-08-05
avm1: Fix signed zero behavior in `Math.min`/`max`

kjarosh:
Flash Player normalizes signed zeros returned by `Math.min`/`max` to +0.


moulins:
See issue #8336.

I'm pretty sure AVM1 normalizes *all* signed zeros to +0, in *all* circumstances (maybe as part of automatic float-to-int conversions?), so maybe we should add this check in the `From<f64>` impl, as was proposed for AVM2.

PR #21237, opened 2025-08-06
avm2: Throw error when calling JSON constructor

SuchAFuriousDeath:


core/src/avm2/globals/json.rs
```diff
@@ -16,6 +16,13 @@ use serde_json::{Map as JsonObject, Value as JsonValue};
 use std::borrow::Cow;
 use std::ops::Deref;
 
+pub fn json_allocator<'gc>(
+    _class: ClassObject<'gc>,
+    activation: &mut Activation<'_, 'gc>,
+) -> Result<Object<'gc>, Error<'gc>> {
+    Err(make_error_2012(activation, "JSON"))
```
Lord-McSweeney: My version of FP throws the error
```
ArgumentError: Error #2012: JSON$ class cannot be instantiated.
```
but this code will throw the error
```
ArgumentError: Error #2012: JSON class cannot be instantiated.
```
(the dollar sign after JSON is missing)

If this is a mistake and the correct error message is the first one, I think you should just be able to mark the class as `[Ruffle(Abstract)]` and not need to edit any Rust code.

SuchAFuriousDeath: You're probably correct, I didn't actually check with Flash player for this.

kjarosh:
I'd add JSON to the abstract classes test

kjarosh (approved):

Issue #21231, opened 2025-08-06
Home Sheep Home 2: Darkened sheep on main menu


n0samu:
### Describe the bug

In [Home Sheep Home 2](https://archive.org/details/home-sheep_home-2-lost-in-london), as of Ruffle nightly 2025-08-05, the sheep on the main menu appear darkened. And after hovering the mouse cursor over a sheep, a dark oval will appear behind it.

<img width="645" height="538" alt="darkened sheep on main menu" src="https://github.com/user-attachments/assets/da06f09d-5e6a-4987-bf6b-c8f65bd9bff6" />

### Expected behavior

Before the August 5 nightly, the sheep appeared correctly, even after hovering the mouse over them:

<img width="644" height="539" alt="Sheep displaying correctly on main menu" src="https://github.com/user-attachments/assets/caeeb1e2-e8ea-4aa5-9002-ba6897556394" />

### Content Location

Page URL: https://archive.org/details/home-sheep_home-2-lost-in-london
SWF URL: https://archive.org/download/home-sheep_home-2-lost-in-london/homesheephome2lostinlondon.swf

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

This was regressed by commit e919e5770880fd0e06b0f48b9827fc05e54a1ca5 (#21176). cc @moulins

Lord-McSweeney:
Fixed by #21238

PR #21238, opened 2025-08-06
core: fix issue #21231

moulins:
A line got lost in the #21176 refactor.


Lord-McSweeney:
Can you add test coverage for this?

n0samu:
I don't wanna speak for moulins but I imagine it might be easier if someone else contributes the test (but if moulins does want to make one, that's great!)

n0samu:
This also fixes #21249. Since our users are seeing a lot of impact, let's try to merge this before tomorrow's nightly. A test can always be added later.

moulins:
Added a regression test by expanding the existing `avm1/as_transformed_flag` test to cover more properties (it only covered `_x/_y`).

PR #21253, opened 2025-08-07
A new error event has been added to the player's panic function.

falcon027:
Hey everyone, 

I couldn’t find a simple way to catch a player panic from host code.
Could Ruffle fire an `error` event when this happens? Then we could do:
``` js
player.addEventListener("error", (err) => console.error("Ruffle panic:", err));
```
I opened a small PR with a draft implementation, happy to tweak it as needed.

Love the project, thanks for all the hard work!


web/packages/core/src/internal/player/inner.tsx
```diff
@@ -152,7 +152,15 @@ export class InnerPlayer {
      * @event RufflePlayer#loadeddata
      */
     static LOADED_DATA = "loadeddata";
+    
+    /**
+     * Triggered when the ruffle player panict.
+     *
+     * @event RufflePlayer#error
+     */
+    static PLAYER_ERROR = "error";
```
kjarosh, line 161: I'd go with

```suggestion
    /**
     * Triggered when the Ruffle player panics.
     *
     * @event RufflePlayer#panicked
     */
    static PANICKED = "panicked";
```

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -2023,7 +2031,9 @@ export class InnerPlayer {
         }
 
         errorArray.push(this.getPanicData());
-
+    
+        this.element.dispatchEvent(new CustomEvent(InnerPlayer.PLAYER_ERROR, error));
```
kjarosh, line 2035: You can pass custom data in `detail`:

```suggestion
        this.element.dispatchEvent(new CustomEvent(InnerPlayer.PANICKED, { detail: { error } }));
```

kjarosh:
Thanks for the PR BTW, this feature seems reasonable for me :)

falcon027 (commented):
Looks good to me

kjarosh:
The comments still need to be addressed.

torokati44:
Is this something worth rescuing? I guess not with an actual (still relevant) use case...

PR #21254, opened 2025-08-07
chore: Only define DomainWeak struct when egui feature is enabled

Lord-McSweeney:
This should fix CI


core/src/avm2/domain.rs
```diff
@@ -25,6 +25,9 @@ use ruffle_wstr::WStr;
 #[collect(no_drop)]
 pub struct Domain<'gc>(Gc<'gc, DomainData<'gc>>);
 
+/// Weak version of the Domain struct, used to mark children of a Domain when
+/// the egui feature is enabled
+#[expect(dead_code)]
```
kjarosh: Shouldn't it be `allow` instead of `expect`? As it won't be dead_code when egui is enabled?

Issue #11935, opened 2023-07-06
Minicraft Flash: Severe performance issues


n0samu:
### Describe the bug

Link to SWF: http://files.catbox.moe/xo5lm1.swf

This is a followup to #11284. The game is working fine now, but extremely slow. It takes a long time to respond to each keypress, and "generating world" takes so long that Ruffle's max execution duration needs to be increased from the default value.

### Expected behavior

The game should be reasonably fast and responsive.

### Affected platform

Desktop app

### Operating system

Windows 10

PR #21255, opened 2025-08-07
avm2: Add an int*int -> int path to help other fast paths

adrian17:
This allows `array[i1 * i2 + i3]` to use an int-indexing fast path, instead of coercing a float to string and then parsing it back.

In https://github.com/ruffle-rs/ruffle/issues/11935 , for me in "generating world" it speeds up from ~4.5s to 2.8s.

This does slightly hurt code that just do regular float multiplies, but it shouldn't be noticeable.

### DIFF
--- a/core/src/avm2/activation.rs
+++ b/core/src/avm2/activation.rs
@@ -2158,9 +2158,21 @@ impl<'a, 'gc> Activation<'a, 'gc> {
     }
 
     fn op_multiply(&mut self) -> Result<(), Error<'gc>> {
-        let value2 = self.pop_stack().coerce_to_number(self)?;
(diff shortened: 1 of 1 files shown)


evilpie:
This might still be an useful optimization in general even after narrowing doubles to ints when possible, because it avoids two integer to double conversions. I sadly don't have any sense for how the type distribution of `op_multiply` usually looks like, I hope most int * int operations would use  `op_multiply_i`?

This also lead me to: https://github.com/ruffle-rs/ruffle/pull/21255/files#diff-8cbc05c3124304056ca5a9085dd612e6b864dae36aa34133118f7190bfdcfc14R1921 ... are we not handling integer overflow for addition?

evilpie (approved):

PR #21266, opened 2025-08-09
Add wildcard support for SocketProxy

freehuntx:
Currently we have to explicitly set host and port. In some cases we might dont know the host and port.  
In this case a wildcard could be useful.


n0samu:
Do you know of any games that use so many different hosts/ports that specifying each one is infeasible? Or is there another reason to add this feature?

freehuntx:
> Do you know of any games that use so many different hosts/ports that specifying each one is infeasible? Or is there another reason to add this feature?

I try to implement a generic flash multiplayer library, which allows for mocking network traffic via js.  
So i can revive basically any multiplayer game.  

Currently im hooking WebSocket to fake the WebSocket connection.  
But i still have the issue, i cant know what ip/port the games actually use (because it should hijack ANY traffic).

n0samu:
Ooh I see, that makes sense 👍

freehuntx:
I think there should be a way to pass the real host and port to the websocket connection somehow.  
So one could still distinguish the connection via port for example.
Maybe allow for placeholders inside of proxyUrl.

E.g.
```
window.RufflePlayer.config.socketProxy = [{
  host: '*',
  port: 0,
  proxyUrl: 'ws://example.com?host=$host&port=$port'
}]
```
where $host and $port gets replaced

freehuntx:
The more i look at it, the less it makes sense having this in the socketProxy array.
Since it becomes like a "catch-all", so its more like a "global proxy" kinda thing...

freehuntx:
What if "socketProxy" can also be "true", which causes the connection to just be "translated" into a ws connection?  
e.g. `tcp://162.71.123.123:5001` becomes (for http) `ws://162.71.123.123:5001` (for https) `wss://162.71.123.123:5001`  
I think i like that, and it allows me to intercept all connections! And even extract details from the url.

n0samu:
> What if "socketProxy" can also be "true", which causes the connection to just be "translated" into a ws connection? e.g. `tcp://162.71.123.123:5001` becomes (for http) `ws://162.71.123.123:5001` (for https) `wss://162.71.123.123:5001` I think i like that, and it allows me to intercept all connections! And even extract details from the url.

That doesn't make sense for games that still have their original server running - you can't run both the ws proxy and the main game server on the same IP/port, and if you changed what IP/port the game server runs on, you'd break legacy Flash clients that still expect that IP/port to have the game server, not the ws proxy. Hope that makes sense!

freehuntx:
Ohh true!

freehuntx:
Can you tell me, if every feature should also be "configurable" for non-web ruffle clients? If so, callbacks, methods, factory functions are a bad solution we shouldnt consider using for this. (e.g. some sort of proxyMiddlewares that can generate a ws url or skip)

E.g.
```
  socketProxyMiddlewares: [
    (host, port) => {
      if (host === '8.8.8.8') return // Skip for real servers
      return 'ws://' + host + ':' + port
    }
  ]
};
```

n0samu:
Considering that the WebSocket proxy is already a web-only feature (right?) I think something like that could be a valid thing to implement. But you'd need to talk to some more active Ruffle devs about how exactly this feature/config should be designed/implemented.

PR #21271, opened 2025-08-09
avm2: Only call init methods when necessary

Lord-McSweeney:
Split off from #20892

This uncovered a bug where `DateObject::from_date_and_time` was always returning a Date with the current date.


core/src/display_object/video.rs
```diff
@@ -452,19 +449,10 @@ impl<'gc> TDisplayObject<'gc> for Video<'gc> {
...
-                video_constr,
-                &[size.0.into(), size.1.into()],
-            ) {
-                Ok(object) => {
-                    self.set_object2(context, object.into());
-                }
-                Err(e) => tracing::error!("Got {} when constructing AVM2 side of video player", e),
-            }
+            let object =
+                Avm2StageObject::for_display_object(context.gc(), self.into(), video_constr);
```
Lord-McSweeney, line 445: At this point in the code, the size of the Video is already set, so there's no need to call the initializer to set it again using the init method. I'll add this as a comment.

PR #21276, opened 2025-08-09
avm2: Throw error 1132 for Json.parse(null) and Json.parse(undefined)

SuchAFuriousDeath:


core/src/avm2/globals/json.rs
```diff
@@ -264,7 +264,13 @@ pub fn parse<'gc>(
     _this: Value<'gc>,
     args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
-    let input = args.get_string(activation, 0)?;
+    let input = args.get_value(0);
```
kjarosh: Haven't checked it myself, but how do strings `"null"` and `"undefined"` behave here?

SuchAFuriousDeath: image
This is from the adhoc test in Ruffle. I'll test "undefined" myself.

core/src/avm2/globals/json.rs
```diff
@@ -264,7 +264,13 @@ pub fn parse<'gc>(
...
+
+    let input = match input {
+        Value::Null | Value::Undefined => return Err(make_error_1132(activation)),
```
Lord-McSweeney: `JSON.parse` accepts a string, so the input will never be `undefined` (though it can be `null`).

core/src/avm2/globals/json.rs
Lord-McSweeney: Can you use `try_get_string` instead of `get_value` to read the argument?

kjarosh (approved):
LGTM, thanks!

PR #21277, opened 2025-08-09
avm2: Correct String.split and String.lastIndexOf

SuchAFuriousDeath:


core/src/avm2/globals/string.rs
```diff
@@ -154,11 +154,19 @@ pub fn last_index_of<'gc>(
 
     let pattern = args.get_string(activation, 0);
 
-    let start_index = args.get_i32(1);
+    let position = args.get_f64(activation, 1)?;
 
-    let start_index = match usize::try_from(start_index) {
-        Ok(n) => n + pattern.len(),
-        Err(_) => return Ok((-1).into()), // Bail out on negative indices.
+    // If position is NaN or positive infinity, search the whole string
```
evilpie: It seems like it would be simpler to first handle `position < 0`?

PR #21281, opened 2025-08-09
core: Add `#[no_dynamic]` to many methods in `TDisplayObject` (and its subtraits)

moulins:
...and change the expansion of `#[no_dynamic]` to:
- be less error-prone, by triggering a compile-time error if the user tries to override a `#[no_dynamic]` method in a trait impl;
- produce better code, by moving the method body to the `enum` trait impl and delegating to it in the default method definition.

This reduces the size of the (desktop) release binary by ~150kB.

-------------------

Example of the new code generation:
```rs
#[automatically_derived]
#[doc(hidden)]
mod __TInteractiveObject_do_not_override {
    pub trait NoDyn<'no_dyn> {}
    impl NoDyn<'_> for () {}
}

pub trait TInteractiveObject<'gc>:
    'gc + Clone + Copy + Collect<'gc> + Debug + Into<InteractiveObject<'gc>>
{
    fn mouse_enabled<'no_dyn>(self) -> bool
    where
        (): __TInteractiveObject_do_not_override::NoDyn<'no_dyn>,
    {
        let mut o: InteractiveObject<'_> = (self).into();
        o.mouse_enabled()
    }
    
    // <snip>
}

impl<'gc> TInteractiveObject<'gc> for InteractiveObject<'gc> {
    fn mouse_enabled<'no_dyn>(self) -> bool
    where
        (): __TInteractiveObject_do_not_override::NoDyn<'no_dyn>,
    {
        self.raw_interactive()
            .contains_flag(InteractiveObjectFlags::MOUSE_ENABLED)
    }
    
    // <snip>
}
```

### DIFF
--- a/core/macros/src/lib.rs
+++ b/core/macros/src/lib.rs
@@ -121,10 +121,8 @@ pub fn enum_trait_object(args: TokenStream, item: TokenStream) -> TokenStream {
     let trait_methods: Vec<_> = input_trait
         .items
         .iter_mut()
-        .filter_map(|item| match item {
+        .map(|item| match item {
             TraitItem::Fn(ref mut method) => {
-                let method_name = &method.sig.ident;
-
                 let mut is_no_dynamic = false;
 
                 method.attrs.retain(|attr| match &attr.meta {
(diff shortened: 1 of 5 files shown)


core/macros/src/lib.rs
```diff
@@ -19,6 +19,9 @@ use syn::{
 /// underlying type. Additionally, `From` will be implemented for all of the variants,
 /// so an underlying type can easily be converted into the enum.
 ///
+/// Methods can be individually marked with `#[no_dynamic]`, which will exempt them from
+/// being dynamically dispatched, preventing implementots from overriding them.
```
Lord-McSweeney: typo: "implementots" -> "implementors"

Lord-McSweeney (approved):

core/macros/src/lib.rs
moulins: Thanks, fixed!

Issue #21273, opened 2025-08-09
pkgbuild file has not been updated for Arm64


coreybruce:
Hey I just wanted to let you know that the pkgbuild file hasn't been updated to support aarch64

https://github.com/ruffle-rs/ruffle/blob/master/desktop/packages/linux/aur/ruffle-nightly-bin/PKGBUILD

Here is a updated pkgbuild file with the changes needed

https://gitlab.com/-/snippets/4879001

evilpie:
@coreybruce Please consider opening  a pull request for review.

PR #21286, opened 2025-08-10
Updated PKGBUILD file to support Arm64

coreybruce:
With this change the AUR package will now support ARM64

* Fixes https://github.com/ruffle-rs/ruffle/issues/21273

### DIFF
--- a/desktop/packages/linux/aur/ruffle-nightly-bin/PKGBUILD
+++ b/desktop/packages/linux/aur/ruffle-nightly-bin/PKGBUILD
@@ -3,14 +3,16 @@ pkgname=ruffle-nightly-bin
 pkgver=@VERSION@
 pkgrel=1
 pkgdesc="A Flash Player emulator written in Rust"
-arch=('x86_64')
+arch=('x86_64' 'aarch64')
 url="https://ruffle.rs/"
 license=('Apache' 'MIT')
 depends=(zlib libxcb alsa-lib)
 provides=(ruffle)
 conflicts=(ruffle)
-source=("https://github.com/ruffle-rs/ruffle/releases/download/nightly-${pkgver//./-}/ruffle-nightly-${pkgver//./_}-linux-x86_64.tar.gz")
-sha512sums=(@SHA512SUM_x86_64@)
+source_x86_64=("https://github.com/ruffle-rs/ruffle/releases/download/nightly-${pkgver//./-}/ruffle-nightly-${pkgver//./_}-linux-x86_64.tar.gz")
+source_aarch64=("https://github.com/ruffle-rs/ruffle/releases/download/nightly-${pkgver//./-}/ruffle-nightly-${pkgver//./_}-linux-aarch64.tar.gz")
+sha512sums_x86_64=(@SHA512SUM_x86_64@)
+sha512sums_aarch64=(@SHA512SUM_aarch64@)
 
 package() {
 	cd "$srcdir/"


desktop/packages/linux/aur/ruffle-nightly-bin/PKGBUILD
```diff
@@ -3,14 +3,16 @@ pkgname=ruffle-nightly-bin
...
 license=('Apache' 'MIT')
 depends=(zlib libxcb alsa-lib)
 provides=(ruffle)
 conflicts=(ruffle)
-source=("https://github.com/ruffle-rs/ruffle/releases/download/nightly-${pkgver//./-}/ruffle-nightly-${pkgver//./_}-linux-x86_64.tar.gz")
-sha512sums=(@SHA512SUM@)
+source_x86_64=("https://github.com/ruffle-rs/ruffle/releases/download/nightly-${pkgver//./-}/ruffle-nightly-${pkgver//./_}-linux-x86_64.tar.gz")
+source_aarch64=("https://github.com/ruffle-rs/ruffle/releases/download/nightly-${pkgver//./-}/ruffle-nightly-${pkgver//./_}-linux-aarch64.tar.gz")
+sha512sums_x86_64=(@SHA512SUM@)
+sha512sums_aarch64=(@SHA512SUM@)
```
kjarosh: Aarch64 build has a different hash than x86_64, they need to be keyed differently and computed separately.

They are computed in the GHA workflow.

desktop/packages/linux/aur/ruffle-nightly-bin/PKGBUILD
```diff
@@ -3,14 +3,16 @@ pkgname=ruffle-nightly-bin
...
+arch=('x86_64' 'aarch64')
 url="https://ruffle.rs/"
 license=('Apache' 'MIT')
 depends=(zlib libxcb alsa-lib)
 provides=(ruffle)
 conflicts=(ruffle)
-source=("https://github.com/ruffle-rs/ruffle/releases/download/nightly-${pkgver//./-}/ruffle-nightly-${pkgver//./_}-linux-x86_64.tar.gz")
-sha512sums=(@SHA512SUM@)
+source_x86_64=("https://github.com/ruffle-rs/ruffle/releases/download/nightly-${pkgver//./-}/ruffle-nightly-${pkgver//./_}-linux-x86_64.tar.gz")
+source_aarch64=("https://github.com/ruffle-rs/ruffle/releases/download/nightly-${pkgver//./-}/ruffle-nightly-${pkgver//./_}-linux-aarch64.tar.gz")
```
kjarosh, line 13: Considering the URLs differ in arch only, can you use `$arch` instead of splitting `source_x86_64`/`source_aarch64`?

coreybruce: I was going to say yes but I just realized  they will need different checksums as they are different files so that will not work as it will skip or or only have 1 checksum, it's easier and better to keep it as it is to verify both files

desktop/packages/linux/aur/ruffle-nightly-bin/PKGBUILD
coreybruce: This is correct, once you update  (@SHA512SUM@) the the correct version on each then run updpkgsums it will update them to the correct checksums for each file.

Here is the latest versions checksums as a example

sha512sums_x86_64=('3898e49ecbb3ef0440063f798489d130da22d42cb296893c3ea5ea14f707785372b2175016483575dda2a2f033a77df69f11e9ad08037e7b1abc6779cdd8328b')
sha512sums_aarch64=('32357eed4424fc665cbb016eb4c77d68ba5f233e35796ee3fae79895f3d929db8ab0d6089648f0d6e3275a9c69c9e351b2321dd3a79fa6fae187dd005338b66f')

Is this what you meant @kjarosh ?

kjarosh: In Ruffle, the checksums are calculated and updated in the GHA workflow instead of using `updpkgsums`. We don't use arch to update the PKGBUILD. See here:

https://github.com/ruffle-rs/ruffle/blob/4391592bd6849d1778e8e5243128f1aeada41103/.github/workflows/release_nightly.yml#L732

We just need to calculate the aarch64 archive sum and add a separate key.

If you don't want to make those changes, that's fine, I'll add them, but please make sure that the PKGBUILD will work on aarch64 with this configuration, as I don't have a way of testing that.

coreybruce: Just out of curiosity why would you do that instead of just using the proper tool to do it the easier way? 

with Arch and a PKGBUILD file this isn't needed as It already works without doing that, that is the point of updpkgsums which will do that for you or needing to add unnecessary additional code as this is very nonstandard.

kjarosh: That decision has been made long time ago, not by me. But if I were to justify this, I'd say that it's not easier at all, because our runners are not running arch. So it's easier to calculate the hash manually and update it, than to spin up arch, set it up, and use `updpkgsums`.

coreybruce: Ahh I see, that not a very unsustainable idea that is very non standard, I would suggest that they should if they want to maintain a AUR package or at least use distrobox but fair enough.

I'll leave the that part to you also who maintains the AUR package?

kjarosh: I'm sorry, I don't think I get your point. Ruffle is a community effort, everyone's free to contribute — there's no single person that does all the work. I'm more versed in Linux stuff so I often focus on it and do reviews, like this one.

If you want to improve the way we do stuff you're welcome to do that, we appreciate every contributor. If you think that using arch specifically in this case is better, you can file a PR and improve things.

But regarding *this* PR, as I said earlier, I cannot merge something that looks untested (and the current code will **not** work). If you don't want to update the hashes, that's fine, I'll take care of it, but I need to know that this PKGBUILD will work on aarch64 with the correct hashes. (I'm unable to test it.)

kjarosh: It'd be great if you could confirm that when replacing hashes to proper ones, this PKGBUILD builds and Ruffle works on Arch-based aarch64 distro. Especially audio, video, Wayland & X11 stuff (portals, clipboard, etc.)

coreybruce: Screenshot_2025-08-23_21-59-41
Screenshot_2025-08-23_22-06-04


Replacing a hash isn't going to change it if works or not but I can confirm it does work which I already knew it did because I used to maintain the ruffle-bin package on the AUR before it got forcefully merged with ruffle-nightly-bin and compiled the binaries for x86_64 and aarch64 since 2023. All those things work on Aarch64 the same way it does on x86_64

https://gitlab.com/linuxbombay/ruffle-linux/binaries

coreybruce: I am specifically talking about the Arch side and yes I know this is community project but if you want to do Arch packages I would suggest using Arch to test it, that's like using Windows to test Linux binaries or vise versa, it cant be properly tested

Yes I can confirm it works, simply updating the version, adding aarch64 like I did on my example snippet it works

https://gitlab.com/-/snippets/4879001

kjarosh: > Replacing a hash isn't going to change it if works or not

It's not about the hash, but about the whole PKGBUILD.

> I already knew it did because I used to maintain the ruffle-bin package

That was a separate PKGBUILD, I haven't seen it so I cannot compare.

> if you want to do Arch packages I would suggest using Arch to test it, that's like using Windows to test Linux binaries or vise versa, it cant be properly tested

We aren't testing the AUR package, I was referring to the release script that updates the PKGBUILD with the new nightly version. I think it's a great idea if you'd like to set up automated tests on Arch for Ruffle.

> Yes I can confirm it works

Thanks for confirming! Here's the PR with new hashes: https://github.com/ruffle-rs/ruffle/pull/21449

kjarosh (approved):
LGTM, thanks!

coreybruce:
Awesome, happy to contribute :+1: :smile:

Issue #21285, opened 2025-08-10
latest nightly changelog shows changes from nightly-2025-08-08 to nightly-2025-08-10


zjk261:
should be from nightly-2025-08-09 to nightly-2025-08-10

torokati44:
Might be caused by https://github.com/ruffle-rs/ruffle/pull/20860.
See: https://github.com/orgs/community/discussions/5975#discussioncomment-1799743

torokati44:
We might have to add a `--notes-start-tag` arg to the `gh release create` invocation in `release.py`: https://cli.github.com/manual/gh_release_create

kjarosh:
Thanks for letting us know @zjk261 !

PR #21289, opened 2025-08-10
ci: Specify start tag manually for release notes

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/21285

When publishing release commits for nightlies, GitHub cannot properly find the start tag.

Tested on my repo and GitHub can generate notes properly after this fix.

### DIFF
--- a/.github/scripts/release.py
+++ b/.github/scripts/release.py
@@ -5,6 +5,7 @@ import subprocess
 import sys
 from datetime import datetime
 import xml.etree.ElementTree as xml
+import json
 
 SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
 REPO_DIR = os.path.realpath(os.path.join(SCRIPT_DIR, '../../'))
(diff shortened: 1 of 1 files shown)


.github/scripts/release.py
```diff
@@ -66,6 +67,23 @@ def cargo_set_version(args):
...
+        '--json=tagName',
+        '--jq', '.[] | select(.tagName | startswith("nightly-"))',
+    ])
+    for tag_json in tags_json.splitlines():
+        tag = json.loads(tag_json)
+        yield tag['tagName']
+
+
+def gh_get_last_nightly_tag():
+    return next(gh_list_nightly_tags(16), None)
```
adrian17, line 86: don't really understand why the param is here, might as well hardcode it in the function itself?

Also, can you add a comment that this would need a refactor if we ever did non-linear releases?

adrian17 (approved):

.github/scripts/release.py
kjarosh: I added the param here because it's an important assumption of `gh_get_last_nightly_tag` and not `gh_list_nightly_tags`. The assumption is that the last nightly tag will be within the last 16 releases.
 
> Also, can you add a comment that this would need a refactor if we ever did non-linear releases?

Hmmm, I don't see why it wouldn't work with non-linear releases. It should work the same as long as we release nightly releases from the main branch.

kjarosh: (I'll add a comment about the first point there)

kjarosh: Done

PR #21296, opened 2025-08-10
Put an option out of advanced & in the pop-up.

rosics-code:
Screenshot 2025-08-10 1 15 13 PM
Screenshot 2025-08-10 1 13 04 PM
makes it convenient!

### DIFF
--- a/web/packages/extension/assets/popup.html
+++ b/web/packages/extension/assets/popup.html
@@ -36,6 +36,10 @@
                 <input type="checkbox" id="swf_takeover" />
                 <label for="swf_takeover">Play SWF files in browser instead of downloading</label>
             </div>
+            <div class="option checkbox">
+                <input type="checkbox" id="ignore_optout">
+                <label for="ignore_optout" class="">Play Flash content even on sites that disallow Ruffle</label>
+            </div>
         </div>
         <div class="buttons-container">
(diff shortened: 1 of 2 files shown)


torokati44:
Just curious - have you come across many sites that "disallow Ruffle" with that opt-out data attribute?

web/packages/extension/assets/popup.html
```diff
@@ -36,6 +36,10 @@
                 <input type="checkbox" id="swf_takeover" />
                 <label for="swf_takeover">Play SWF files in browser instead of downloading</label>
             </div>
+            <div class="option checkbox">
+                <input type="checkbox" id="ignore_optout">
+                <label for="ignore_optout" class="">Play Flash content even on sites that disallow Ruffle</label>
```
torokati44: nit: Why the `class=""`?

rosics-code: That's how it is in the options page?

rosics-code: fixed

rosics-code:
> Just curious - have you come across many sites that "disallow Ruffle" with that opt-out data attribute?

list could grow

n0samu:
I don't think this is an option anyone needs quick and easy access to. It'd be on you to explain why that's not the case.

PR #21297, opened 2025-08-10
pixelbender: Implement assembly

kjarosh:
* Supersedes https://github.com/ruffle-rs/ruffle/pull/21058

Pixel Bender assembly allows creating Pixel Bender kernels easily with low level instructions.


render/pixel_bender/assembly_tests/src/runner.rs
```diff
@@ -0,0 +1,171 @@
...
+
+#[derive(Clone, Copy, Deserialize)]
+enum TestType {
+    Roundtrip,
+    Assemble,
+    Dissassemble,
+}
+
+impl TestType {
+    fn perform_assembly(self) -> bool {
```
Lord-McSweeney: nit: Can these methods be called `performs_XXX` instead of `perform_XXX`? The latter suggests that the method is doing something, while the former is a description only.

Lord-McSweeney (approved):

render/pixel_bender/assembly_tests/src/runner.rs
kjarosh: Fixed!

PR #21313, opened 2025-08-12
core: Force correct alignment for `DisplayObjectBase`

moulins:
On certain architectures (e.g. 32-bits Android), `f64`s are only 4-aligned, so we need to increase the alignment of `DisplayObjectBase` to always match with its subclasses.

### DIFF
--- a/core/src/display_object.rs
+++ b/core/src/display_object.rs
@@ -192,6 +192,8 @@ impl BitmapCache {
 
 #[derive(Clone, Collect)]
 #[collect(no_drop)]
+// Ensure this always has the same alignment as its subclasses (needed for `Gc` casts).
+#[repr(align(8))]
 pub struct DisplayObjectBase<'gc> {
     cell: RefCell<DisplayObjectBaseMut>,
     parent: Lock<Option<DisplayObject<'gc>>>,


core/src/display_object.rs
```diff
@@ -192,6 +192,7 @@ impl BitmapCache {
 
 #[derive(Clone, Collect)]
 #[collect(no_drop)]
+#[repr(align(8))]
```
torokati44, line 196: Can confirm that this change fixed the Android x86 build in my testing.

core/src/display_object.rs
```diff
@@ -229,8 +230,6 @@ pub struct DisplayObjectBase<'gc> {
     /// Values other than the default `BlendMode::Normal` implicitly cause cache-as-bitmap behavior.
     blend_mode: Cell<ExtendedBlendMode>,
 
-    #[collect(require_static)]
-
```
torokati44: I don't know how this change is related or why it's necessary, haven't tested it.

core/src/display_object.rs
torokati44: (Although maybe a comment next to it on why it's there, similar to the PR description here, would be nice.)

moulins: Comment added.

core/src/display_object.rs
moulins: Removed it, it was a drive-by change that doesn't affect anything.

PR #21334, opened 2025-08-15
avm2: Defer nextFrame calls from frame scripts for FP9 compatibility

jarca0123:
This makes the timeline/events/fp9 test pass.


core/src/display_object/movie_clip.rs
```diff
@@ -701,7 +704,13 @@ impl<'gc> MovieClip<'gc> {
     }
 
     pub fn next_frame(self, context: &mut UpdateContext<'gc>) {
-        if self.current_frame() < self.total_frames() {
```
Lord-McSweeney: Just to make sure, this behavior is only when `nextFrame` is called, not `prevFrame` etc?

Lord-McSweeney:
Thanks for working on this!

Issue #12270, opened 2023-07-23
Test failure: avm1/from_shumway/duplicatemovieclip/dontremove


Dinnerbone:
Test path: `tests/tests/swfs/avm1/from_shumway/duplicatemovieclip/dontremove`

# Expected
```
_level0.test
_level0.test1
_level0.test2
_level0.test
_level0.test1
undefined
```

# Actual
```
_level0.test
_level0.test1
_level0.test2
_level0.test
_level0.test1
_level0.test2
```

# Difference
```diff
6c6
< undefined
---
> _level0.test2
```

Issue #12145, opened 2023-07-18
Test failure: shumway's timeline/nav/ratio3


Dinnerbone:
Test path: `tests/tests/swfs/timeline/nav/ratio3`

# Expected
```
Element placed on frame 1
New object placed on frame 2: false
New object placed on frame 3: true
New object placed on frame 1: false
```

# Actual
```
Element placed on frame 1
New object placed on frame 2: false
New object placed on frame 3: true
New object placed on frame 1: true
```

# Difference
```diff
4c4
< New object placed on frame 1: false
---
> New object placed on frame 1: true
```

torokati44:
Almost exact duplicate of #12144. :stuck_out_tongue_winking_eye:

Issue #12144, opened 2023-07-18
Test failure: shumway's timeline/nav/ratio2


Dinnerbone:
Test path: `tests/tests/swfs/timeline/nav/ratio2`

# Expected
```
Element placed on frame 1
New object placed on frame 2: false
New object placed on frame 3: true
New object placed on frame 1: false
```

# Actual
```
Element placed on frame 1
New object placed on frame 2: false
New object placed on frame 3: true
New object placed on frame 1: true
```

# Difference
```
4c4
< New object placed on frame 1: false
---
> New object placed on frame 1: true
```

Issue #12143, opened 2023-07-18
Test failure: shumway's timeline/morph_shape


Dinnerbone:
Test path: `tests/tests/swfs/timeline/morph_shape`

# Expected
```
Element placed on frame 1
New object placed on frame 2: false
New object placed on frame 3: false
New object placed on frame 1: true
```

# Actual
```
Element placed on frame 1
New object placed on frame 2: false
New object placed on frame 3: false
New object placed on frame 1: false
```

# Difference
```diff
4c4
< New object placed on frame 1: true
---
> New object placed on frame 1: false
```

PR #21336, opened 2025-08-15
avm1: Adjust clip removal logic for rewinds

jarca0123:
* Fixes https://github.com/ruffle-rs/ruffle/issues/12270
* Fixes https://github.com/ruffle-rs/ruffle/issues/12143
* Fixes https://github.com/ruffle-rs/ruffle/issues/12144
* Fixes https://github.com/ruffle-rs/ruffle/issues/12145

This fixes Pink Tac Toe, Ratman Ralph, Super Sub Hero, Electricman 2, from_shumway/duplicatemovieclip/dontremove, avm1/rewind_depth and the rest of failing timeline/nav tests (ratio2, ratio3, morph_shape). I am not sure about the logic though.

### DIFF
--- a/core/src/display_object/movie_clip.rs
+++ b/core/src/display_object/movie_clip.rs
@@ -1569,22 +1569,7 @@ impl<'gc> MovieClip<'gc> {
 
             let children: SmallVec<[_; 16]> = self
                 .iter_render_list()
-                .filter(|child| {
-                    let is_candidate_for_removal = if self.movie().is_action_script_3() {
-                        child.place_frame() > frame || child.placed_by_script()
-                    } else {
-                        child.depth() < AVM_DEPTH_BIAS
-                    };
-
-                    if !is_candidate_for_removal && child.as_morph_shape().is_none() {
-                        return false;
-                    }
-                    if let Some(final_placement) = final_placements.get(&child.depth()) {
-                        !self.survives_rewind(*child, &final_placement.place_object)
-                    } else {
-                        true
-                    }
-                })
+                .filter(|child| !self.survives_rewind(*child, &final_placements, frame))
                 .collect();
 
             for child in children {
@@ -1711,9 +1696,29 @@ impl<'gc> MovieClip<'gc> {
         self.assert_expected_tag_end(hit_target_frame);
     }
 
-    fn survives_rewind(self, old_object: DisplayObject<'_>, new_params: &swf::PlaceObject) -> bool {
+    fn survives_rewind(
+        self,
+        old_object: DisplayObject<'_>,
+        final_placements: &HashMap<Depth, &GotoPlaceObject<'_>>,
+        frame: FrameNumber,
+    ) -> bool {
         // TODO [KJ] This logic is not 100% tested. It's possible it's a bit
         //    different in reality, but the spirit is there :)
+
+        let is_candidate_for_removal = if self.movie().is_action_script_3() {
+            old_object.place_frame() > frame || old_object.placed_by_script()
+        } else {
+            old_object.depth() < AVM_DEPTH_BIAS
+        };
+
+        if !is_candidate_for_removal && old_object.as_morph_shape().is_none() {
+            return true;
+        }
+        let Some(final_placement) = final_placements.get(&old_object.depth()) else {
+            return false;
+        };
+
+        let new_params = &final_placement.place_object;
 
         if !old_object.movie().is_action_script_3()
             && old_object.placed_by_avm1_script()
@@ -1721,12 +1726,14 @@ impl<'gc> MovieClip<'gc> {
         {
             return false;
         }
+
         let id_equals = match new_params.action {
             swf::PlaceObjectAction::Place(id) | swf::PlaceObjectAction::Replace(id) => {
                 old_object.id() == id
             }
             _ => false,
         };
+
         let ratio_equals = match new_params.ratio {
             Some(ratio) => old_object.ratio() == ratio,
             None => true,


n0samu:
Ahh this is super exciting!! Another one I'll need to test - I'm curious if it will affect Electricman 2 or Super Sub Hero, among others.

n0samu:
As far as I can tell, this PR (EDIT: almost) ~~completely~~ fixes Electricman 2! This is incredible news 🎉 🎉 🎉 
(No change in Super Sub Hero, though.)

n0samu:
As @p0008874 pointed out on Discord, there is still a more subtle problem remaining with Electricman 2. If you beat the Cirrian Guard and then play the same level again, only two enemies will spawn in the second time. That's a problem since you can't beat the level till you defeat 3 enemies. Also, if you play the Replicants level after that, a Cirrian Guard will spawn in in place of one of the Replicants.
GitHub won't let me upload the video demonstrating this, but here's the link to it on Discord: https://discord.com/channels/610531541889581066/704769195749146634/1406900010729345075

n0samu:
This also fixes #5219! 🎉 🎉 🎉

core/src/display_object/movie_clip.rs
```diff
@@ -1550,21 +1551,33 @@ impl<'gc> MovieClip<'gc> {
             }
         }
         let hit_target_frame = self.0.current_frame() == frame;
-
         if is_rewind {
-            // Remove all display objects that were created after the
```
Lord-McSweeney, line 1556: Can you keep the comments here, or replace them with new ones explaining what the new logic does?

n0samu:
Now that this PR is using logic from #5431, it seems to work perfectly 🎉 
Fixes #10382, #815, #5183, #9594, #5219 (can we link these issues to the PR please?)

core/src/display_object/morph_shape.rs
```diff
@@ -52,18 +51,17 @@ impl<'gc> MorphShape<'gc> {
...
     pub fn ratio(self) -> u16 {
-        self.0.ratio.get()
+        self.0.base.ratio.get()
     }
 
     pub fn set_ratio(self, ratio: u16) {
-        self.0.ratio.set(ratio);
+        self.0.base.ratio.set(ratio);
         self.invalidate_cached_bitmap();
     }
```
Lord-McSweeney: Since the `ratio` and `set_ratio` functions are now defined on TDisplayObject, I think these two functions don't need to be defined here?

core/src/display_object/morph_shape.rs
```diff
@@ -131,7 +129,7 @@ impl<'gc> TDisplayObject<'gc> for MorphShape<'gc> {
...
 
     fn render_self(self, context: &mut RenderContext) {
-        let ratio = self.0.ratio.get();
+        let ratio = self.0.base.ratio.get();
```
Lord-McSweeney: This should just be able to do `let ratio = self.ratio();`

core/src/display_object/morph_shape.rs
```diff
@@ -140,7 +138,7 @@ impl<'gc> TDisplayObject<'gc> for MorphShape<'gc> {
     }
 
     fn self_bounds(self) -> Rectangle<Twips> {
-        let ratio = self.0.ratio.get();
+        let ratio = self.0.base.ratio.get();
```
Lord-McSweeney: Does `let ratio = self.ratio();` work here?

tests/tests/swfs/avm1/from_shumway/duplicatemovieclip/dontremove/test.toml
```diff
@@ -1,4 +1,4 @@
 # Test adapted from Shumway at https://github.com/mozilla/shumway/tree/master/test/swfs/avm1/duplicatemovieclip
 
 num_frames = 4
-known_failure = true # https://github.com/ruffle-rs/ruffle/issues/12270
+
```
Lord-McSweeney: nit: can you remove this extra newline at the end of the file?

Lord-McSweeney (approved):
Thank you for working on this!

core/src/display_object/movie_clip.rs
```diff
@@ -1551,6 +1552,58 @@ impl<'gc> MovieClip<'gc> {
...
+                        && color_transform_equals
+                }
+                DisplayObject::Avm1Button(_)
+                | DisplayObject::Avm2Button(_)
+                | DisplayObject::EditText(_)
+                | DisplayObject::Bitmap(_)
+                | DisplayObject::Video(_) => {
+                    ratio_equals && old_object.id() == id && clip_depth_equals
+                }
+                _ => ratio_equals,
```
Lord-McSweeney: Could this "default" case be replaced with an explicit `DisplayObject::MovieClip(_)` to make it clear what's being checked here?

tests/tests/swfs/from_shumway/timeline/nav/ratio3/test.toml
```diff
@@ -1,4 +1,3 @@
 # Test adapted from Shumway at https://github.com/mozilla/shumway/tree/master/test/swfs/timeline/nav
 
 num_frames = 3
-known_failure = true # https://github.com/ruffle-rs/ruffle/issues/12145
```
kjarosh, line 4: Before merging I want to just make sure if those shumway tests are good. Based on my experience, shumway sometimes had wrong outputs and those got merged here

kjarosh: (Sorry, added the comment under a wrong test, but my point stands 😄)

kjarosh: The test is alright!

Lord-McSweeney:
It might be worth testing this with some SWFs mentioned in #1305 to ensure it doesn't regress any of them

jarca0123:
> It might be worth testing this with some SWFs mentioned in #1305 to ensure it doesn't regress any of them

There is probably only a regression most likely unrelated to this PR with The Bottom of the Sea (http://www.ferryhalim.com/orisinal/g3/sea.htm), where the bar showing how much you want to jump updates sporadically.

core/src/display_object.rs
```diff
@@ -2321,6 +2352,8 @@ pub trait TDisplayObject<'gc>:
                     morph_shape.set_ratio(ratio);
                 } else if let Some(video) = self.as_video() {
                     video.seek(context, ratio.into());
+                } else {
+                    self.set_ratio(ratio);
                 }
```
Lord-McSweeney: I think this code doesn't properly set the ratio on Videos?
```
                if let Some(morph_shape) = self.as_morph_shape() {
                    morph_shape.set_ratio(ratio);
                } else if let Some(video) = self.as_video() {
                    video.seek(context, ratio.into());
                } else {
                    self.set_ratio(ratio);
                }
```
should be
```
                if let Some(video) = self.as_video() {
                    video.seek(context, ratio.into());
                }
                self.set_ratio(ratio);
```

core/src/display_object/morph_shape.rs
kjarosh: Done

core/src/display_object/morph_shape.rs
kjarosh: Fixed

core/src/display_object/morph_shape.rs
kjarosh: Outdated

core/src/display_object.rs
kjarosh: Outdated

core/src/display_object/movie_clip.rs
```diff
@@ -1690,6 +1711,59 @@ impl<'gc> MovieClip<'gc> {
...
+        if !old_object.movie().is_action_script_3()
+            && old_object.placed_by_avm1_script()
+            && old_object.depth() < AVM_DEPTH_BIAS
+        {
+            return false;
+        }
+        let id = match new_params.action {
+            swf::PlaceObjectAction::Place(id) | swf::PlaceObjectAction::Replace(id) => id,
+            _ => 0,
+        };
```
moulins: Nit: this conflates `Place(0) / Replace(0)` with other actions; I know SWFs aren't supposed to mention the character #0, but I don't think we guard against that, so maybe `id` should be a `Option` here?

core/src/display_object/movie_clip.rs
```diff
@@ -1690,6 +1711,59 @@ impl<'gc> MovieClip<'gc> {
...
+        let ratio_equals = match new_params.ratio {
+            Some(ratio) => old_object.ratio() == ratio,
+            None => true,
+        };
```
moulins, line 1740: Style nit: this could be shortened to `new_params.ratio.is_none_or(|r| r == old_object.ratio())` (and similarly for the subsequent `match`es).

core/src/display_object/movie_clip.rs
```diff
@@ -1562,10 +1563,30 @@ impl<'gc> MovieClip<'gc> {
             // TODO: We want to do something like self.children.retain here,
             // but BTreeMap::retain does not exist.
             // TODO: Should AS3 children ignore GOTOs?
+
+            let final_placements: HashMap<Depth, &GotoPlaceObject<'_>> =
+                goto_commands.iter().map(|cmd| (cmd.depth(), cmd)).collect();
+
```
moulins, line 1569: As a larger refactor, we could instead have `goto_commands` be directly a `HashMap`; from what I can see it doesn't have a meaningful ordering and we already ensure no duplicate `Depth`s.

core/src/display_object/movie_clip.rs
```diff
@@ -1562,10 +1563,30 @@ impl<'gc> MovieClip<'gc> {
...
+                    if !is_candidate_for_removal && child.as_morph_shape().is_none() {
+                        return false;
+                    }
+                    if let Some(final_placement) = final_placements.get(&child.depth()) {
+                        !self.survives_rewind(*child, &final_placement.place_object)
+                    } else {
+                        true
+                    }
+                })
```
moulins: Nit: I'd prefer if the entirety of this closure was pulled in `survives_rewind`, I think.

core/src/display_object/movie_clip.rs
kjarosh: It's shorter but less readable, you'd have to think for a while what's the default value for `None`, it's not natural to read it as `is none or / is some and`.

core/src/display_object/movie_clip.rs
kjarosh: Fixed

core/src/display_object/movie_clip.rs
moulins: Idk, it seems fairly natural to me: "`new_params.ratio` is none or equal to `old_object.ratio()`"

kjarosh: The true/false default value is an equal choice, there's no reason why you would choose "is none or" vs "is some and". Matter of fact, for some properties it's the former, for some it's the latter. When trying to assess which one should be where, you have to map "is none or" & "is some and" to default values instead of reading them as-is.

Not to mention, this logic is untested, so we don't really know which one should be where.

core/src/display_object/movie_clip.rs
kjarosh: Done

kjarosh:
I think we should pull in this code and improve upon it later (with dedicated tests).

kjarosh:
@jarca0123 thank you for working on it!

Issue #21301, opened 2025-08-11
Partial Font Display Issue — Font Rendering Problem When Running AimBooster SWF on Ruffle


iansnet:
### Describe the bug

Hello,
I encountered an issue with partial font display when playing AimBooster.swf using Ruffle. The specific problem is:
The lower part of fonts on some buttons or texts in the game is missing or cut off, causing incomplete text display.
This issue does not occur in the original Flash Player, where fonts display correctly.

<img height="200" alt="Image" src="https://github.com/user-attachments/assets/1cff0b2c-6bce-4476-81d0-67da6188a1f6" />

<img height="200" alt="Image" src="https://github.com/user-attachments/assets/aa3414d7-2835-4bc5-ab70-f3a81203a930" />

<img height="200" alt="Image" src="https://github.com/user-attachments/assets/c5dc384d-2ded-4dfb-a9fe-6fc3c77369f1" />

Additionally, I created a Simplified Chinese version of this game, but when running it in Ruffle, the Chinese text in the settings options of Custom mode does not display correctly, while in the original Flash Player it displays normally.

<img height="200" alt="Image" src="https://github.com/user-attachments/assets/f81f2ee9-640f-4e2f-95b4-ed23dd56be21" />

### Expected behavior

Fonts should display completely and consistently with the original Flash Player.

<img height="200" alt="Image" src="https://github.com/user-attachments/assets/b19c21b5-f459-4793-b440-81c998449301" />

<img height="200" alt="Image" src="https://github.com/user-attachments/assets/c35091bd-8a12-4af2-b075-fda54a9c4f69" />

<img height="200" alt="Image" src="https://github.com/user-attachments/assets/aadd57e2-2ee2-49e8-bc8c-6f10b0a372ab" />

<img height="200" alt="Image" src="https://github.com/user-attachments/assets/cffef844-25c5-4902-b4a5-07b29701434d" />

### Content Location

Original game download link：
http://www.aimbooster.com/AimBooster.swf

Simplified Chinese version files：
[AimBooster_CHS.zip](https://github.com/user-attachments/files/21720045/AimBooster_CHS.zip)

Steps to Reproduce:
1）：Use the latest version of the Ruffle client to open AimBooster.swf.
2）：Enter Training mode and select any sub-mode, or choose any game mode and start it, then exit to the results screen.
3）：Notice that the lower part of the fonts is missing or not fully displayed.
[...]

PR #21338, opened 2025-08-15
debug_ui: Add layout debug information on EditTexts

kjarosh:
Makes it easier to debug device text issues such as https://github.com/ruffle-rs/ruffle/issues/21301.

### DIFF
--- a/core/src/html/layout.rs
+++ b/core/src/html/layout.rs
@@ -1378,6 +1378,10 @@ impl<'gc> LayoutBox<'gc> {
         }
     }
 
+    pub fn text_range(&self) -> Range<usize> {
+        self.start()..self.end()
+    }
+
     /// Return x-axis char bounds of the given char relative to the whole layout.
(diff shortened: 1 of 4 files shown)


core/src/html/layout.rs
```diff
@@ -1270,7 +1270,7 @@ impl<'gc> LayoutBox<'gc> {
     ///
     /// TODO It's currently unused, but will be useful when adding support for
     /// images embedded in HTML.
-    #[expect(unused)]
+    #[allow(unused)]
```
Lord-McSweeney, line 1273: Was there a reason to change this to `allow`?

Lord-McSweeney (approved):

core/src/html/layout.rs
kjarosh: Yes, it's described in the commit message:
> Looks like clippy can mistakenly fail on expect(unused) when the method
is in fact unused.

Basically clippy started complaining after this PR despite the fact that it's still unused. It looks like a bug but I haven't looked into it

PR #21339, opened 2025-08-15
Improve String.lastIndexOf and String.split accuracy

SuchAFuriousDeath:
Supersedes #21277

### DIFF
--- a/core/src/avm2/globals/string.rs
+++ b/core/src/avm2/globals/string.rs
@@ -154,11 +154,14 @@ pub fn last_index_of<'gc>(
 
     let pattern = args.get_string(activation, 0);
 
-    let start_index = args.get_i32(1);
-
-    let start_index = match usize::try_from(start_index) {
-        Ok(n) => n + pattern.len(),
-        Err(_) => return Ok((-1).into()), // Bail out on negative indices.
+    let start_index = match args.get_f64(1) {
+        float if float.is_nan() => this.len(),
+        f64::INFINITY => this.len(),
+        f64::NEG_INFINITY => return Ok((-1).into()),
+        float => match usize::try_from(float as i32) {
+            Ok(n) => n + pattern.len(),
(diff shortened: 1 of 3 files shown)


core/src/avm2/globals/string.rs
```diff
@@ -154,11 +154,14 @@ pub fn last_index_of<'gc>(
 
     let pattern = args.get_string(activation, 0);
 
-    let start_index = args.get_i32(1);
-
-    let start_index = match usize::try_from(start_index) {
-        Ok(n) => n + pattern.len(),
-        Err(_) => return Ok((-1).into()), // Bail out on negative indices.
+    let start_index = match args.get(1) {
```
Lord-McSweeney: The argument is `Number` typed, so you can access it with `args.get_f64(1)` without losing any information.

core/src/avm2/globals/string.rs
```diff
@@ -154,11 +154,14 @@ pub fn last_index_of<'gc>(
...
+        None | Some(Value::Undefined) | Some(Value::Number(f64::INFINITY)) => this.len(),
+        Some(Value::Number(float)) if float.is_nan() => this.len(),
+        Some(Value::Number(f64::NEG_INFINITY)) => return Ok((-1).into()),
+        Some(n) => match usize::try_from(n.coerce_to_i32(activation)?) {
```
Lord-McSweeney: Looking at avmplus, I think this should be using an `as i32` cast instead of the `coerce_to_i32` logic?

core/src/avm2/globals/string.rs
SuchAFuriousDeath: Done

core/src/avm2/globals/string.rs
SuchAFuriousDeath: Done. Hope I understood correctly

kjarosh (approved):

PR #21342, opened 2025-08-15
ci: Add test coverage

kjarosh:
Coverage Report computes coverage of SWF tests after each PR.

This can be used to assess how well the newly added code is covered.

I've tried using `tarpaulin` and `llvm-cov`. The output of `llvm-cov` was a lot more accurate and it took around the same time to run. Running `llvm-cov` with `nextest` caused tests to execute over 4 times slower compared to `test`. Still, coverage runs faster than tests on Windows, which remains the longest executing job for a PR.

Things to do in the future:

* add coverage for all tests; this was impossible with `tarpaulin` (due to errors), but should be possible with `llvm-cov`; it's still worth keeping SWF-tests-only coverage, as it informs us about what code is tested in terms of compatibility with FP,
* ignore known failures as they don't contribute towards coverage.

### DIFF
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -13,6 +13,9 @@ concurrency:
 env:
   FEATURES: lzma,jpegxr
   TEST_OPTS: --workspace --locked --no-fail-fast -j 4
+  LLVM_COV_OPTS: --branch --profile ci ${TEST_OPTS} --features ${FEATURES},imgtests
+
+  LINUX_DEPENDENCIES: libasound2-dev mesa-vulkan-drivers libudev-dev
 
   # This is to counteract the disabling by rust-cache.
   # See: https://github.com/Swatinem/rust-cache/issues/43
@@ -77,7 +80,7 @@ jobs:
         if: runner.os == 'Linux'
         run: |
           sudo apt-get update
-          sudo apt install -y libasound2-dev mesa-vulkan-drivers libudev-dev
+          sudo apt install -y ${{ env.LINUX_DEPENDENCIES }}
 
       - name: Enable image tests
         if: runner.os != 'macOS'
@@ -135,7 +138,7 @@ jobs:
       - name: Install Linux dependencies
         run: |
           sudo apt-get update
-          sudo apt install -y libasound2-dev mesa-vulkan-drivers libudev-dev
+          sudo apt install -y ${{ env.LINUX_DEPENDENCIES }}
 
       - name: Check formatting
         run: cargo fmt --all -- --check
@@ -152,6 +155,49 @@ jobs:
         run: cargo doc --no-deps --all-features
         env:
           RUSTDOCFLAGS: -D warnings
+
+  coverage:
+    needs: changes
+    if: needs.changes.outputs.should_run == 'true'
+    name: Coverage Report
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Install Rust toolchain
+        # We need nightly for coverage due to:
+        #  1. wayland-backend requiring that,
+        #  2. branch coverage being unstable.
+        #
+        # Ad 1.
+        #   See <https://github.com/Smithay/wayland-rs/commit/05d8307ea647053b949357eb8588dfdab394f4fe>
+        # Ad 2.
+        #   See <https://github.com/taiki-e/cargo-llvm-cov/issues/8>
+        #   See <https://github.com/rust-lang/rust/issues/79649>
+        uses: dtolnay/rust-toolchain@nightly
+        with:
+          components: llvm-tools
+
+      - name: Install llvm-cov
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-llvm-cov
+
+      - name: Install Linux dependencies
+        run: |
+          sudo apt-get update
+          sudo apt install -y ${{ env.LINUX_DEPENDENCIES }}
+
+      - name: Run llvm-cov for SWF tests
+        working-directory: tests
+        run: cargo llvm-cov --html ${{ env.LLVM_COV_OPTS }}
+
+      - name: Upload report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: llvm-cov
+          path: target/llvm-cov
 
   dependencies:
     needs: changes


Lord-McSweeney:
Is it possible to also run the image tests when testing coverage?

kjarosh:
> Is it possible to also run the image tests when testing coverage?

I thought I enabled them :sweat_smile: should be good now!

torokati44 (approved):
Neat! I'd only have some minor nits, but I'll spare you those...

Do we want some short summary (maybe the "Totals" row at the bottom) printed to the action log, or even commented under PRs? :eyes:

kjarosh:
> Neat! I'd only have some minor nits, but I'll spare you those...

Please don't hesitate!

> Do we want some short summary (maybe the "Totals" row at the bottom) printed to the action log, or even commented under PRs? 👀

That was my original plan, but then I realized that such summary without delta coverage is not that useful :/ (and computing delta coverage is not that easy, so I spared it for a future PR)

torokati44:
(This is not really related, I just remembered again: what would - still - be quite cool - and maybe even useful sometimes - is to have a live deployment of the demo page somewhere for every open PR. Basically the same thing as what `egui` does. Maybe the ruffle-regressions site could facilitate this...?)

torokati44:
Well, even if the delta is not computed, having the current totals easily accessible would help with manual comparisons...

.github/workflows/test_rust.yml
```diff
@@ -153,6 +156,53 @@ jobs:
...
+
+      - name: Install Linux dependencies
+        run: |
+          sudo apt-get update
+          sudo apt install -y ${{ env.LINUX_DEPENDENCIES }}
+
+      - name: Run llvm-cov for SWF tests
+        working-directory: tests
+        run: |
+          cargo llvm-cov --html ${{ env.LLVM_COV_OPTS }}
```
torokati44: nit: Why the newline with pipe? i get that this is consistent with the previous step, but not with other single-liners.

.github/workflows/test_rust.yml
```diff
@@ -153,6 +156,53 @@ jobs:
...
+
+      - name: Upload report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: llvm-cov
+          path: |
+            target/llvm-cov
```
torokati44: nit: Same here about the line break.

.github/workflows/test_rust.yml
```diff
@@ -153,6 +156,53 @@ jobs:
         env:
           RUSTDOCFLAGS: -D warnings
 
+  coverage:
+    needs: changes
+    if: needs.changes.outputs.should_run == 'true'
+    name: Coverage Report
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Clone Ruffle repo
```
torokati44: nit: Do we usually name this step too?

.github/workflows/test_rust.yml
```diff
@@ -153,6 +156,53 @@ jobs:
...
+          #
+          # Ad 1.
+          #   See <https://github.com/Smithay/wayland-rs/commit/05d8307ea647053b949357eb8588dfdab394f4fe>
+          # Ad 2.
+          #   See <https://github.com/taiki-e/cargo-llvm-cov/issues/8>
+          #   See <https://github.com/rust-lang/rust/issues/79649>
+          toolchain: nightly
+          components: llvm-tools
+
+      - name: Install llvm-cov
```
torokati44, line 181: nit: Isn't this just the cargo subcommand, with the llvm-cov tool itself already installed above?

.github/workflows/test_rust.yml
```diff
@@ -153,6 +156,53 @@ jobs:
...
+    needs: changes
+    if: needs.changes.outputs.should_run == 'true'
+    name: Coverage Report
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Clone Ruffle repo
+        uses: actions/checkout@v4
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
```
torokati44: nit: Why not `@nightly` then?

torokati44:
As you can see, none of these notes are particularly important... (:

.github/workflows/test_rust.yml
kjarosh: Fixed

.github/workflows/test_rust.yml
kjarosh: We do in the release workflow, that's where I originally put it. Removed the name

.github/workflows/test_rust.yml
kjarosh: Good catch, I didn't notice that I can specify versions in two places...

.github/workflows/test_rust.yml
kjarosh: No, they are separate projects. cargo-llvm-cov needs llvm-tools

.github/workflows/test_rust.yml
kjarosh: Done

kjarosh:
> Well, even if the delta is not computed, having the current totals easily accessible would help with manual comparisons...

I'll do that in a separate PR then

.github/workflows/test_rust.yml
```diff
@@ -153,6 +156,49 @@ jobs:
...
+          tool: cargo-llvm-cov
+
+      - name: Install Linux dependencies
+        run: |
+          sudo apt-get update
+          sudo apt install -y ${{ env.LINUX_DEPENDENCIES }}
+
+      - name: Run llvm-cov for SWF tests
+        working-directory: tests
+        run: cargo llvm-cov --html ${{ env.LLVM_COV_OPTS }}
```
torokati44, line 193: nit: Why is `--html` not in the envvar?

kjarosh: That's because it's a local option, not global. You can call `llvm-cov` in a different way, e.g. first generate coverage without the report and then generate the report separately. Global options wouldn't change then, but `--html` would.

torokati44:
Hmm, are we sure that adding another build (different set of enabled features and cfg flags) to the same os+toolchain+cargoconfig combo doesn't interfere with the caching of intermediate build artifacts? :thinking:
I haven't seen anything hinting at this, just thought about the possibility. And even if so, it most likely won't break anything, just slow down some checks.

torokati44:
Oh, never mind, there's no caching in the coverage job. Maybe there should be? And definitely with its own cache key then.

torokati44:
BTW, for making the report easily accessible: There's also this sort of thing:
https://github.com/ruffle-rs/ruffle-android/actions/runs/17045700377/workflow
Which does these things:
https://github.com/ruffle-rs/ruffle-android/actions/runs/17045700377
https://github.com/ruffle-rs/ruffle-android/actions/runs/17045567210?pr=532

I know this specifically is tied to junit, but I meant more to show the possibility of adding additional workflow run annotations during run and after the fact.

kjarosh:
> Oh, never mind, there's no caching in the coverage job. Maybe there should be? And definitely with its own cache key then.

Just repeating here what I said on Discord: llvm-cov currently does hard clean so it wouldn't work anyway, we'll have to check whether building incrementally won't break anything

torokati44:
> Still, coverage runs faster than tests on Windows, which remains the longest executing job for a PR.

Just a general note about this: In some cases, it's not only the max duration of jobs triggered for a PR what matters. Whenever there are multiple PRs to be merged all after one another quickly, the workflows are also running on `master` after each one. And the more job slots those workflows take up (we can only have a small number of jobs running at once per repo), and for longer (individually); the PRs rebased after can only start their check jobs that much later...

torokati44:
And since the "merge queue" doesn't work the way we would want/expect (at least the last time I checked, years ago, it didn't), someone has to shepherd them all manually - it's even better when multiple people are doing it, without coordinating.

torokati44:
> Still, coverage runs faster than tests on Windows, which remains the longest executing job for a PR.

Seems like this is no longer the case: 22m vs. 14m (N=1).
Yay many tests! :P

PR #21346, opened 2025-08-16
avm2: Implement missing Number methods

SuchAFuriousDeath:
The uncontroversial changes split off from #21018


core/src/avm2/globals/Number.as
```diff
@@ -36,6 +36,78 @@ package {
...
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function floor(x:Number):Number;
+
+        [API("680")]
+        [Ruffle(FastCall)]
+        public static native function log(x:Number):Number;
+
+        [API("680")]
```
Lord-McSweeney, line 81: Can you mention the comment in Math.as explaining why the functions `max` and `min` are marked as `FastCall`?

Lord-McSweeney (approved):

kjarosh:
If the code comes from someone else's PR, the author should be preserved, and you can add yourself as a co-author.

SuchAFuriousDeath:
Just FYI, I asked the author whether I can do this. But I think your proposal makes sense anyway. I'll redo the commit later today.

SuchAFuriousDeath:
@jarca0123

jarca0123:
I agree with the author changes.

kjarosh:
> Just FYI, I asked the author whether I can do this. But I think your proposal makes sense anyway. I'll redo the commit later today.

> I agree with the author changes.

We're talking about two different things here. The _author_ git field tracks authorship, and authorship *cannot* be transferred (compared to ownership/copyright). There's no legal framework to transfer authorship and the point of authorship is to not be transferrable.

You could create this PR without asking jarca0123 for consent, because they've already given consent by distributing their works under an MIT + Apache license, which explicitly allows doing such things (under some circumstances).

SuchAFuriousDeath:
What I was trying to convey was that I wasn't maliciously trying to appropriate anyone's contribution. Nothing about legality or anything like that.

kjarosh:
I wasn't implying that you were doing this maliciously, I'm pretty sure you weren't; I assumed it was a simple mistake :)

PR #21347, opened 2025-08-16
avm2: Correct error text in LocalConnectionObject::run_method

SuchAFuriousDeath:
This is probably not quite correct, but with a proper error constructor, there's an extra 'Error: ' in the error text, failing the test. Hopefully someone knows the proper way to do this


evilpie:
I suspect we should call error.toString or something like that. I will take a look.

evilpie:
Okay, sorry for delaying this. It seems like the `text` is always the same indeed.

We can even see this in output.txt. The `AsyncErrorEvent`'s `text` has nothing to do with the `error`.
```
AsyncError caught. Text: Error #2095: flash.net.LocalConnection was unable to invoke callback lcHandler., error: Error: fooo
```

PR #21351, opened 2025-08-16
ci: Update metainfo automatically in release commits

kjarosh:
Metainfo contains information about releases.  By updating it automatically on each release commit, we can make sure all downstreams have valid metainfo with proper releases.

It's a follow-up to https://github.com/ruffle-rs/ruffle/pull/20860

It hasn't been end-to-end tested, but the code has been adapted from stable releases branch (and it worked there), and the worst case scenario is a failed nightly release. The Python script works properly.

### DIFF
--- a/.github/scripts/release.py
+++ b/.github/scripts/release.py
@@ -48,6 +48,32 @@ def log(msg):
     print(msg, file=sys.stderr)
 
 
+def add_release_to_metainfo(path, tag_name, version):
+    url = f'https://github.com/ruffle-rs/ruffle/releases/tag/{tag_name}'
+    xml_release = xml.Element('release')
+    xml_release.set('version', version)
+    xml_release.set('date', get_current_date())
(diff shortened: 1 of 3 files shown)


.github/scripts/release.py
```diff
@@ -121,6 +147,15 @@ def bump():
     github_output('version4', version4)
 
 
+def metainfo():
+    metainfo_path1 = f'{REPO_DIR}/desktop/packages/linux/rs.ruffle.Ruffle.metainfo.xml'
+    metainfo_path2 = f'{REPO_DIR}/desktop/packages/linux/rs.ruffle.Ruffle.metainfo.xml.in'
```
adrian17, line 152: What's the `xml.in` file for?

kjarosh: That's the source we're editing, the `.xml` file is the output that contains translations from Crowdin

PR #21352, opened 2025-08-16
avm2: Improve error handling for verifier and supercalls, do typechecking in optimizer

Lord-McSweeney:
We now throw a proper AVM error in many cases where we would've panicked before


core/src/avm2/optimizer/optimize.rs
```diff
@@ -117,6 +116,24 @@ impl<'gc> OptValue<'gc> {
                     created_value.class = other.class;
                 }
             }
+        } else if let (Some(self_class), Some(other_class)) = (self.class, other.class) {
+            // Check for a common superclass.
+            // FIXME: Make this faster?
```
adrian17, line 123: avmplus temporaraily sets a `commonBase` flag when iterating supers of the first type, then iterating the second type checks for the flag. It's kinda silly, but prevents a quadratic loop and should be doable here with a `Cell<bool>` too?

Not sure it'll even speed up the average case, but definitely prevent quadratic edge cases.

Lord-McSweeney: From what I can tell, 99.9% of the time that this branch is reached it's because of merging `Number` and `int`/`uint`, so this is usually very fast. I saw `commonBase` in avmplus too but it seems like a very hacky solution for almost no improvement (this branch is almost never hit).

I think it's fine to have this implemented this way, but if we care about the possibility of a SWF intentionally using a very long inheritance chain to break this, I can switch it over to the `commonBase` approach.

adrian17 (approved):
Can you rename the PR please? Since it also extends the optimizer, making it possibly trigger more often

core/src/avm2/globals.rs
```diff
@@ -507,6 +507,9 @@ pub fn init_early_classes<'gc>(
     class_class.into_finished_class(activation);
     object_class.into_finished_class(activation);
 
+    // We don't need to validate the classes, as we already know that the
+    // `Object` and `Class` classes are valid
```
homersimpsons, line 511: Sorry if this review is not expected.

I was wondering if this "we already know" should be made explicit using `debug_assert!()`?

PR #21374, opened 2025-08-17
core: Merge a subset of NotoSansMath into the built-in NotSans font

### LINKED ISSUE #20337: vanishing of the symbole "≥" and "≤"
### Describe the bug

ruffle don't detect the symboles "< or equal" and "> or equal"  if it is between a letter and a number

### Expected behavior

showing "< or >" wherever is written in the text 

Image

### Content Location

https://bibliotheque.sesamath.net/ressource/apercevoir/1836
after first question :  https://bibliotheque.sesamath.net/ressource/apercevoir/734 

### Affected platform

Online demo

### Operating system

windows 10

### Browser

opera 

### Additional information

also desktop have same issue

evilpie:
Fixes #20337. Thanks to @jenskutilek for solving the issue with the MATH table.

**Before:**
Screenshot 2025-08-17 at 22-28-10 exo2657 swf
(Not sure why it says _undefined_ everywhere)

**After:**
Screenshot 2025-08-17 at 22-29-23 exo2657 swf


tests/tests/swfs/avm2/edittext_always_show_selection/output.expected.png
evilpie, line 1: Not sure if that is okay. Seems like the selection now extends further down... Are we using the glyph extends for rendering this?

tests/tests/swfs/avm2/stage3d_texture/output.expected.png
evilpie, line 1: Actually rendering the `=` here now :smile:

tests/tests/swfs/avm2/edittext_always_show_selection/output.expected.png
kjarosh: If you've added glyphs with bigger descent, the selection will be extended

kjarosh:
I think it makes sense to go through all of those tests and fix them, it doesn't make sense to update them each time we update the default font, not to mention the fact that it negatively impacts the repo size.

We could maybe add one test that renders various characters using the default font.

I'll try looking into that

tests/tests/swfs/avm2/stage3d_texture/output.expected.png
kjarosh: It's due to using low quality. Device text cannot be rendered on low quality according to my knowledge (not to mention the test shouldn't depend on device fonts).

evilpie:
@kjarosh Would it make sense to try this again now?

kjarosh:
I'm still working on fixes, they'll have to be merged anyway, so I'd wait with this PR

evilpie:
There are still a few differences that look almost all tiny to me. The question is if we should increase the tolerance or update the expected image.

On bigger difference is `stage3d_fractal`, which now shows an equal sign instead of something that looks like an underscore. Same with `stage3d_texture`.

kjarosh:
I'll try fixing these tests too, it's a bit more complicated, but doable

kjarosh:
(but if you prefer merging before fixing those tests up, I'd go with increasing outliers)

tests/tests/swfs/avm2/stage3d_texture/output.expected.png
evilpie: I don't think it makes sense to fix this using max_outliers, which would be something like 128_000.

hamzaelbarmaki1994:
> (Not sure why it says _undefined_ everywhere)

don't worry about it its a txt file needed to show text. in the web "hoster" the text can be loaded well.
many thanks to all contributors.

kjarosh (approved):

hamzaelbarmaki1994:
@evilpie hi, i notice that the space between lines are much bigger (it it the outliers ? ), is this will be reduced after ?because it hides the text ,like in this image 
image
the optimal result :
image
at least actual behave before this PR: 
image

Issue #10785, opened 2023-04-22
Add support for "ruffle-skip" data tag to automatically skip objects


ErzenXz:
I would like to request a new feature for the Ruffle player that would allow it to automatically skip objects with a "ruffle-skip" data tag. This would be useful for cases where certain objects in a SWF file should not be displayed or processed by Ruffle.

Currently, there is no built-in way to skip objects in this manner, so users must manually edit the SWF file or modify the HTML code to exclude the objects they do not want to display.

By adding support for the "ruffle-skip" data tag, Ruffle could automatically skip any objects in a SWF file that have this tag, making it much easier for users to control which objects are displayed and which are not.

Thank you for your consideration, and please let me know if you have any questions or feedback on this feature request.

Lord-McSweeney:
Currently the only way to do this is to add a `data-ruffle-optout` attribute to the main `<html>` tag:

https://github.com/ruffle-rs/ruffle/blob/292e068e487855d8e157e4c74ceaf582f127f820/web/packages/extension/src/content.ts#L77

ErzenXz:
Yes, i want like specific elements, idk why that even exists because if you dont want it on your entire page just remove the script tag, or maybe can it be done dynamicly? like manipulating the html with js?

danielhjacobs:
> idk why that even exists because if you dont want it on your entire page just remove the script tag, or maybe can it be done dynamicly? like manipulating the html with js?

It exists if your page does not use Ruffle but you want to stop Ruffle extension users from using Ruffle on your page.

I am looking into adding the ability to put `data-ruffle-optout` on any element on an HTML page. That element, and any of its potential children, will not load with Ruffle when Ruffle is in use, but other content on the page will still do so. I'm unsure if that matches your request though. Are you referring to adding an SWF tag for this? We'd rather not add an ability to use tags that were not in Adobe Flash Player, as an anti-goal of any software emulator is distinguishing the emulator usage.

kjarosh:
@ErzenXz could you describe an example use case of this feature and go through it step by step?

PR #21383, opened 2025-08-18
web: Allow data-ruffle-optout for any element to opt-out that section

danielhjacobs:
The purpose is so if a single page has some content that works with Ruffle and some that does not you can add this directly to the embeds that don't work, or even to some parent div of all the embeds that don't work with Ruffle. It still can be ignored with the "Play Flash content even on sites that disallow Ruffle" toggle.

Before this, you could only opt-out of Ruffle page-wide with data-ruffle-optout on the https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement of the page or iframe.

Requested on Discord.

### DIFF
--- a/web/packages/core/src/internal/player/ruffle-player-element.tsx
+++ b/web/packages/core/src/internal/player/ruffle-player-element.tsx
@@ -213,3 +213,23 @@ export function copyElement(element: Element, destination: Element): void {
         }
     }
 }
+
+/**
+ * Checks whether a given element is inside an "opt-out" zone for Ruffle.
+ *
+ * @param element The element to check.
+ * @returns true if the element or its ancestor has data-ruffle-optout and not set to ignore opt-out, else false.
+ */
+export function hasOptOut(element: Element): boolean {
+    // We could try to pass the config specific to this element all the way
+    // but only the extension can set ignoreOptout and when it does
+    // it gets added to the global RufflePlayer config, so this is simpler.
+    const globalConfig = window.RufflePlayer?.config ?? {};
+    if ("ignoreOptout" in globalConfig && globalConfig["ignoreOptout"]) {
+        return false;
+    }
+    if (element.closest("[data-ruffle-optout]")) {
+        return true;
+    }
+    return false;
+}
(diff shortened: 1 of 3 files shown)


n0samu (approved):

danielhjacobs:
Just looked at old issues, so adding:

Addresses #10785 (not called `ruffle-skip`, just re-using `data-ruffle-optout`)

Edit: Actually, I'm not sure that's what is meant by the request. They say "This would be useful for cases where certain objects in a SWF file should not be displayed or processed by Ruffle." That doesn't quite match what I'm implementing.

kjarosh:
What about the existing `checkPageOptout`? Is it still needed?

web/packages/core/src/internal/player/ruffle-player-element.tsx
```diff
@@ -213,3 +213,23 @@ export function copyElement(element: Element, destination: Element): void {
...
+/**
+ * Checks whether a given element is inside an "opt-out" zone for Ruffle.
+ *
+ * @param element The element to check.
+ * @returns true if the element or its ancestor has data-ruffle-optout and not set to ignore opt-out, else false.
+ */
+export function hasOptOut(element: Element): boolean {
+    // We could try to pass the config specific to this element all the way
+    // but only the extension can set ignoreOptout and when it does
+    // it gets added to the global RufflePlayer config, so this is simpler.
```
kjarosh, line 226: It doesn't look good to me, isn't the page able to override the extension's choice this way if it wanted to?

n0samu: No, what the comment is saying is that we're not checking the `<ruffle-player>` element's config, but that's irrelevant because that's not what the extension sets! (for this particular config option, anyway)

kjarosh: Hmmm, to me that comment suggests we could pass the config from within the code, but we aren't doing that because it's simpler to use the global config, which will also have these properties.

The issue I see here is that the global object could be modified by the website.

PR #21417, opened 2025-08-19
avm2: Return undefined for construct_prop on primitives and namespaces

SuchAFuriousDeath:


core/src/avm2/activation.rs
```diff
@@ -1692,7 +1692,20 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
+            || source
+                .as_object()
+                .is_some_and(|o| o.as_namespace_object().is_some());
+
+        // See <https://bugzilla.mozilla.org/show_bug.cgi?id=638233>
+        // See <https://github.com/adobe/avmplus/blob/858d034a3bd3a54d9b70909386435cf4aec81d21/halfmoon/hm-specializer.cpp#L512>
+        let ctor = if is_primitive_or_namespace {
+            source
+                .get_property(&multiname, self)
+                .unwrap_or(Value::Undefined)
```
Lord-McSweeney: avmplus has a `constructprop` function that, when called on a primitive, falls back to the primitive's prototype for the property lookup when the lookup fails. Can you implement that here instead of doing it this way?

SuchAFuriousDeath: Sure

core/src/avm2/value.rs
```diff
@@ -1200,6 +1200,32 @@ impl<'gc> Value<'gc> {
...
+        arguments: FunctionArgs<'_, 'gc>,
+        activation: &mut Activation<'_, 'gc>,
+    ) -> Result<Value<'gc>, Error<'gc>> {
+        let ctor = match self {
+            Value::Object(obj) if obj.as_namespace_object().is_none() => {
+                self.get_property(multiname, activation)?
+            }
+            _ => self
+                .get_property(multiname, activation)
+                .unwrap_or_else(|_| {
```
Lord-McSweeney, line 1235: This still feels like a hacky solution to me, since
1. The property not being found is not the only reason `get_property` can return an `Err`
2. avmplus's version of `construct_property` throws different errors than its `get_property` (e.g. when run on a slot that isn't callable), so we'll have to change it to work that way when we implement it correctly anyways

Also, this shouldn't be trying to construct the `proto`; it should be trying to construct the result of a dynamic lookup of the multiname on the `proto`.

Could you make this match the way avmplus does it, or even just inline the vtable lookup logic from `get_property` here (with the special case for primitives)? That would solve the issues I mentioned above.

SuchAFuriousDeath: ok

PR #21429, opened 2025-08-20
avm2: Move supercalling code into its own file

Lord-McSweeney:
It didn't really make sense for this to be in `class_object.rs`, except that a ClassObject was being used to store the super vtable. Also, the vtable for supercalls is now stored in the op itself rather than being read off the Activation. This will make it easier to implement the method binding logic in the future.

TODO: Add a test for constructsuper coercing the reciever to the super type


Lord-McSweeney:
This PR fails the test added in #21442

PR #21439, opened 2025-08-21
chore: Update package index before installing tools in l10n GHA workflows

torokati44:
To avoid these situations:
 - https://github.com/ruffle-rs/ruffle/actions/runs/17138673876/job/48620893530
 - https://github.com/ruffle-rs/ruffle/actions/runs/17138737889/job/48621278812

### DIFF
--- a/.github/workflows/download_translations.yml
+++ b/.github/workflows/download_translations.yml
@@ -19,6 +19,7 @@ jobs:
 
       - name: Install dependencies
         run: |
+          sudo apt-get update
           sudo apt-get install -y itstool gettext appstream
 
       - name: Generate desktop & metainfo sources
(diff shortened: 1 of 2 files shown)


kjarosh:
I'm not sure if that's a better approach. Won't it introduce non-determinism? We'll always depend on the newest packages which can regress something instead of the standard versions of packages provided by the image, no?

kjarosh:
Oh wait, maybe I'm misunderstanding the error. Is it caused by an outage, or that specific version of a package not existing?

torokati44:
I suspected the former, but checking the status dashboard, it might be the latter. 🤷

torokati44:
The 404 reply suggests that the archive is online, but didn't find a package with the cached version.

kjarosh (approved):

PR #21445, opened 2025-08-22
ci: Calculate delta coverage using diff-cover

kjarosh:
Now each coverage report also includes delta coverage, i.e. coverage scoped to current changes.  For PRs, it's code modified in the PR, for branches it's code since the last push.

The delta coverage report is included as a separate report and also a markdown summary is posted on GitHub Actions.


torokati44:
Huh, does this compute the coverage delta purely from the diff? What if the tests change, etc.?

e.g. if a (hypothetical) PR drops a bunch of SWF tests, can it still report increased coverage if it also happens to add tests for any new code?

kjarosh:
> Huh, does this compute the coverage delta purely from the diff? What if the tests change, etc.?

It's not "delta coverage" as in "the difference in coverage between two versions", but as in "the coverage of difference between two versions". Delta coverage is the standard coverage scoped to new/modified code.

torokati44:
Ah, so "coverage of changes" not "change in coverage"!?

kjarosh:
Yeps, this metric is a lot more useful and easier to calculate compared to the other one

.github/workflows/test_rust.yml
```diff
@@ -163,6 +164,37 @@ jobs:
...
       - uses: actions/checkout@v5
+        with:
+          # We need history for delta coverage, but we also don't want to fetch
+          # the whole repo. 42 commits seem reasonable
+          fetch-depth: 42
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
```
torokati44, line 175: nit: Yet another version to bump manually once in a while (similar to cargo-edit elsewhere)... :face_with_spiral_eyes: (I don't have a better idea though)

torokati44 (approved):
It's ... uh ... probably fine, as far as I can tell - let's see!

.github/workflows/test_rust.yml
kjarosh: I'd be against updating this version automatically, it's important that diff-cover _works_ and not that it's up to date. If it breaks, it's possible we won't notice that.

I should probably fix diff-cover's version too...

PR #21446, opened 2025-08-22
ci: Ignore known failures when running coverage

kjarosh:
Known failures do not contribute to coverage, because we do not verify their output, instead we only check if they start passing.

This PR adds the `RUFFLE_TEST_OPTS` variable, which can contain Ruffle tests specific options separate from `cargo test` (specifically `--ignore-known-failures`). In the future it can also contain e.g. test filtering capabilities (slow/fast tests).

### DIFF
--- a/tests/tests/regression_tests.rs
+++ b/tests/tests/regression_tests.rs
@@ -7,6 +7,7 @@ use crate::external_interface::tests::{external_interface_avm1, external_interfa
 use crate::shared_object::{shared_object_avm1, shared_object_avm2, shared_object_self_ref_avm1};
 use anyhow::Context;
 use anyhow::Result;
+use clap::Parser;
 use libtest_mimic::Trial;
 use ruffle_fs_tests_runner::{FsTestsRunner, TestLoaderParams};
 use ruffle_test_framework::options::TestOptions;
@@ -15,6 +16,7 @@ use ruffle_test_framework::test::Test;
 use std::borrow::Cow;
 use std::panic::{catch_unwind, resume_unwind, AssertUnwindSafe};
 use std::path::PathBuf;
+use std::sync::Arc;
 use std::thread::sleep;
 
 mod environment;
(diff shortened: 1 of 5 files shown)


tests/tests/regression_tests.rs
```diff
@@ -37,12 +47,24 @@ fn main() {
...
 
+    let ruffle_test_opts = match std::env::var("RUFFLE_TEST_OPTS") {
+        Ok(val) => val,
+        Err(std::env::VarError::NotPresent) => "".to_owned(),
+        e @ Err(_) => panic!("{e:?}"),
+    };
+    let ruffle_test_opts = Arc::new(RuffleTestOpts::parse_from(
+        std::iter::once("RUFFLE_TEST_OPTS=")
+            .chain(ruffle_test_opts.split(" ").filter(|s| !s.is_empty())),
+    ));
```
torokati44: Am I reading correctly that here `clap` will get `["RUFFLE_TEST_OPTS=", "--ignore-known-failures"]` as a source to parse, and it will understand it somehow? :thinking: (I'm not that familiar with that crate...)

SuchAFuriousDeath: Basically yes, it's kind of similar to serde.

kjarosh: > Am I reading correctly that here `clap` will get `["RUFFLE_TEST_OPTS=", "--ignore-known-failures"]` as a source to parse, and it will understand it somehow? 🤔 (I'm not that familiar with that crate...)

Yes, the first argument is the name of the program (in this case the name of the variable). It's used to show the proper help message, that's the only reason, it can be anything.

torokati44: Ah, alright, that makes sense.

kjarosh: Added a comment explaining that

torokati44 (approved):

PR #21451, opened 2025-08-23
exporter: Refactor code & add tests

kjarosh:
* Move player-related code to `player_ext.rs`
* Refactor `FrameSelection`
* Fix exporting single frame with `--frames all` & add tests for it


n0samu:
> // TODO Can we just use num_frames from the movie here?

#17019 would suggest not, I think?

kjarosh:
> > // TODO Can we just use num_frames from the movie here?
> 
> #17019 would suggest not, I think?

I don't think this applies, because we read the number of frames in advance anyway. The difference is that we try to read them from the movie clip vs from the SWF, and I'm not sure if there's a difference between those two, especially that we do it at the very beginning.

If we wanted to support https://github.com/ruffle-rs/ruffle/pull/17019, we would have to check after each frame whether we are done (which is probably a better approach, but it's a different issue).

kjarosh:
I've replaced the convoluted logic of getting total frames to getting num frames from the SWF, as after frame count refactors from MovieClip it's clear it's the same value.

I've added a TODO to use another method for that.

PR #21455, opened 2025-08-23
avm2: Store method's bound class on Method

Lord-McSweeney:
This is closer to what avmplus does, and will allow us to optimize dynamic calls to static calls in the future.


core/src/avm2/method.rs
```diff
@@ -370,16 +366,69 @@ impl<'gc> Method<'gc> {
...
+    /// Mark this method as associated with the given `MethodAssociation`.
+    pub fn associate(
+        self,
+        activation: &mut Activation<'_, 'gc>,
+        new_association: MethodAssociation<'gc>,
+    ) -> Result<(), Error<'gc>> {
+        // It is valid to mark a method as bound to a class if either:
+        //   1. It has not been associated yet
+        //   2. It has already been associated as a freestanding function (i.e.
+        //      it has a `bound_class` of `None`)
```
moulins: This is a little misleading: to me, `2.` seems to imply that re-associating a freestanding function should update the `bound_class`, but this isn't what the code does.

Lord-McSweeney: Good catch, I forgot to change this comment when merging two functions. Does it make more sense now?

PR #21468, opened 2025-08-25
tests: Clean up existing shumway tests

kjarosh:
* Remove `from_shumway/as3-loader/load_bytes_2` as it's an exact duplicate of from_shumway/as3-loader/load_bytes. (Looks like it was an error and the real test didn't get adapted.)
* Rename tests so that they match their original names.

This patch makes it less difficult to find your way around shumway tests.


kjarosh:
This PR changes case of some directories

cc @adrian17

Issue #15103, opened 2024-02-03
Layering and transparency seems incorrect when setMask is called on an object.


wearrrrr:
### Describe the bug

![image](https://github.com/ruffle-rs/ruffle/assets/99224452/bccee399-becb-4e31-b9ab-b096bac6a6dc)

This screenshots shows inconsistent layering and transparency from how the SWF is interpreted inside of Flash 32

### Expected behavior

![image](https://github.com/ruffle-rs/ruffle/assets/99224452/345dd77b-7fc2-461c-871c-13e33844d9ab)

This is the expected result, seemingly the only thing changing on this frame is an actionscript call to 
`marisa_maskee.setMask(marisa_mask);`

### Content Location

http://eye.swfchan.com/flash.asp?id=5789&n=-..-flash185--IOSYS_Border_of_Death.swf

### Affected platform

Desktop app

### Operating system

Linux

### Additional information

I am still looking out for other inconsistencies in this flash file, I can open more issues as I find them or update this one, whichever is preferred :slightly_smiling_face:

wearrrrr:
## I've decided to jump back into this, hopefully to figure it out once and for all

### Things I've noted so far
- Frame 842 in Sprite 661 (seems to be a view of all frames in the animation) is the first frame where you can see the issue visually, but Frame 838 is the first frame that the mask is loaded. 
![Frame 842](https://github.com/user-attachments/assets/6cf546d6-1be2-4692-9800-b61f68017b49)
- Frame 838 has the following actions, the contents of doAction being `marisa_maskee.setMask(marisa_mask);`
![838 actions](https://github.com/user-attachments/assets/8d2d37cb-f792-48f1-ac8a-34d0622c62fa)
- When played back in ruffle, the masking is getting applied, but the layering isn't. Keep in mind that neither swivel nor JPEXS can render it correctly, so I would have to assume that it's using some quirk of the original flash player to work.
- [Here](https://github.com/ruffle-rs/ruffle/blob/1c8e95a061a40ccd4b2fdca30a66eb7cc2fbe12d/core/src/avm1/globals/movie_clip.rs#L1216) Is where I, and the other's I asked at the time believed that the handling for setMask was, so this is a good starting point for anyone looking to debug this.

For posterity, here are the flags on `marisa_mask`
![marisa_mask flags](https://github.com/user-attachments/assets/7bab852f-a0f0-4d90-b7ca-38cdc8689d49)

And here are the flags on `marisa_maskee`
[...]

PR #21471, opened 2025-08-25
tests: Add tests for alpha & filters in masks

kjarosh:
Added 3 tests:
* `visual/cache_as_bitmap/cab_mask_alpha`
* `visual/cache_as_bitmap/cab_mask_filters`
* `visual/cache_as_bitmap/cab_mask_triangle`

They check how FP renders masks with alpha depending on the value of `cacheAsBitmap` for both mask and maskee.

Currently Ruffle fails all of these tests.

See https://github.com/ruffle-rs/ruffle/issues/15103

cc @adrian17 @wearrrrr


tests/tests/swfs/avm2/cab_mask_alpha/test.toml
```diff
@@ -0,0 +1,8 @@
+num_ticks = 1
```
kjarosh: To-do: they should probably be placed under visual/, not avm2/

kjarosh: done

PR #21479, opened 2025-08-26
core: Fix memory leak when unloading child SWFs

jarca0123:
This probably fixes all the games that also depend on other SWFs. Currently only tested on binary data.


jarca0123:
Closing in favor of wrapping MovieLibrary into a Gc.

Issue #21443, opened 2025-08-22
Issue with masks


ChrisCPI:
### Describe the bug

If a MovieClip is being masked by another MovieClip in the timeline, but then via ActionScript the first MovieClip has a mask applied to it, and that mask is the same MovieClip that's already masking it on the timeline, then it will appear invisible.

<img width="1092" height="447" alt="Image" src="https://github.com/user-attachments/assets/e0ad610b-a7ef-4ee7-b503-07fae42db021" />

### Expected behavior

The MovieClip should still appear as normal with the mask.

### Content Location

A simple test I made:

[mask_bug.zip](https://github.com/user-attachments/files/21929739/mask_bug.zip)

### Affected platform

Desktop app

### Operating system

macOS 15.5

### Additional information

I tried several other combinations with how the masks are set up, and could not replicate the bug other than how it is here.

PR #21495, opened 2025-08-27
avm1: Reset clip depth on mask when reapplying

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/21443

Clip depth should be reset on both sides of masker and maskee when setting the mask.


ChrisCPI:
Does a fix also need to be done for AVM2? Because the bug also exists there (in the case of my issue, running `rectangle_mc.mask = circle_mc`)

kjarosh:
Yes, I'll open another PR for that

Issue #18688, opened 2024-11-20
Using cacheAsBitmap for dynamic mask not works


juicelink77:
### Describe the bug

My game draw patern into mask to reveal image, here the code : 

var mcPatern:Patern = new Patern();
var patern:BitmapData = new BitmapData(mcPatern.width, mcPatern.height, true, 0);
patern.draw(mcPatern);

var matrixPatern:Matrix = new Matrix();
var maskBitmap:Bitmap = new Bitmap( new BitmapData(image.width, image.height, true, 0) );
addChild(maskBitmap);

image.cacheAsBitmap = true;
maskBitmap.cacheAsBitmap = true;
image.mask = maskBitmap;

function DrawPatern(e)
{
    if (_Stage.Game.paused) {
        return;
    }
    if (!_parent.game_play) {
        return;
    }
    matrixPatern.tx = mouseX;
    matrixPatern.ty = mouseY;
    maskBitmap.bitmapData.draw(patern, matrixPatern);

    if (finished) {
        timer.removeEventListener(TimerEvent.TIMER, valide );
        removeEventListener(MouseEvent.MOUSE_DOWN, mouseEventHandler);
        removeEventListener(MouseEvent.MOUSE_UP, mouseEventHandler);
        _parent.game_play = false;

        clearInterval(_parent.Start);
        _parent.finish(true);
    }
    e.updateAfterEvent();
}

It's works well in Flash CS6 but not in Ruffle.

### Expected behavior

Draw with mouse the mask of image and revealed it.

### Content Location

Launch game and choose training mode and select game 21 or 27.

https://www.flashkof.fr/pulpo/pulpo.swf

### Affected platform

Self-hosted version

### Operating system

Windows 11

### Browser

Chrome

kjarosh:
See also https://github.com/ruffle-rs/ruffle/issues/15103

Issue #15190, opened 2024-02-11
Transparency in masks does not work.


mike84:
### Describe the bug

In AS3 you can create transparent masks when masking a movieclip. If the movieclip and the mask are cached as bitmaps the transparency of the mask would show.
An example is from https://rubber.tube/games/timcredible/timcredible_latest_rufflejs.html
where the lights are rendered this way with a gradient mask with transparency. As the transparency is not working they appear with solid edges.

![ruffle_player_lights](https://github.com/ruffle-rs/ruffle/assets/385883/94225faf-bf59-453f-a09b-051081266cb0)

### Expected behavior

Gradient masks should be shown when the movieclip and the mask are cached as bitmap.
The expected display in this example is like this (shown in the flash player) with soft edges from the gradient mask:
![flash_player_lights](https://github.com/ruffle-rs/ruffle/assets/385883/56189d2c-8d99-4996-8e80-854af1c6323e)

### Content Location

The example is from 
https://rubber.tube/games/timcredible/timcredible_latest_rufflejs.html
the swf is hosted here:
https://rubber.tube/games/timcredible/timCredible_game.swf

Although to preview the effect you will have to either start the game and wait several minutes until night time starts, or use the in-game "dev tools" using "ctrl-alt-R" after starting a new game to show the debug options top right, and click on "night mode".

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

Chrome 120.0.6099.225/Firefox 122.0

Issue #4011, opened 2021-04-15
Mask transparency ignored


ChaseWaylon:
The square is a linear gradient where left side color is set to 0% Alpha:

![mask_transparency](https://user-images.githubusercontent.com/37155903/114871748-49c98980-9e02-11eb-9bf8-1134b2238882.png)

PR #21497, opened 2025-08-27
render: Support transparency in masks

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/18688
* Fixes https://github.com/ruffle-rs/ruffle/issues/15190
* Fixes https://github.com/ruffle-rs/ruffle/issues/4011
* Progresses https://github.com/ruffle-rs/ruffle/issues/15103
* Supersedes https://github.com/ruffle-rs/ruffle/pull/10394

When CAB is enabled on both mask and maskee, masks are rendered differently; instead of masking pixels based on hit test, pixels are masked based on the value of alpha in the mask. This is more akin to `BlendMode.ALPHA`.

This patch adds support for alpha masks (or CAB masks), which are rendered using a shader, compared to a stencil used by regular masks.

Currently it's supported only on `wgpu`.


Lord-McSweeney:
Does this fix any of the `visual/cache_as_bitmap/oversize/` tests?

kjarosh:
> Does this fix any of the `visual/cache_as_bitmap/oversize/` tests?

It does! I'll update the PR soon

PR #21517, opened 2025-08-30
desktop: Switch from winapi to windows-sys

evilpie:
We already depend on window-sys and it's more modern. We previously had a build issue with winapi in #21513.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4544,7 +4544,7 @@ dependencies = [
  "vergen",
  "webbrowser",
  "wgpu",
- "winapi",
+ "windows-sys 0.59.0",
  "winit",
 ]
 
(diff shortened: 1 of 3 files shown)


adrian17 (approved):

torokati44:
Was there any particular reason to go with `windows-sys` instead of plain `windows`? Does it matter at all? I mean from a dep tree perspective, at runtime it's probably the exact same thing...

c.f.:
 - https://docs.rs/windows-sys/0.60.2/windows_sys/Win32/System/Console/fn.AttachConsole.html
 - https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/System/Console/fn.AttachConsole.html

evilpie:
Really, just "why not". We don't need any of the fany features from windows, so I don't think we _need_ to use it. Other crates we use depend on a mix of both as well.

PR #21523, opened 2025-08-31
render: Support filters in alpha masks

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/15103

Alpha masks (used when CAB is on) take filters into account when masking objects in FP.


kjarosh:
Updated code to also match https://github.com/ruffle-rs/ruffle/pull/21624

core/src/display_object/video.rs
```diff
@@ -472,13 +472,16 @@ impl<'gc> TDisplayObject<'gc> for Video<'gc> {
         }
     }
 
-    fn render(self, context: &mut RenderContext) {
+    fn render_with_options(self, context: &mut RenderContext<'_, 'gc>, options: RenderOptions) {
         if !context.is_offscreen && !self.world_bounds().intersects(&context.stage.view_bounds()) {
```
Lord-McSweeney, line 476: From what I understand, this culling check will break if `options.apply_transform` is set to `false`, as `world_bounds` always takes the transform into account, but the rendering itself won't actually be transformed. However, the culling won't be checked anyways if `context.is_offscreen` is `true`?

So, by my understanding, `context.is_offscreen` is always `true` when `options.apply_transform` is `false`...? If this is the case, could you add an assertion for that here and anywhere else this applies?

kjarosh: > However, the culling won't be checked anyways if context.is_offscreen is true?

> So, by my understanding, context.is_offscreen is always true when options.apply_transform is false...?

Currently that's the case I believe.

> From what I understand, this culling check will break if options.apply_transform is set to false, as world_bounds always takes the transform into account, but the rendering itself won't actually be transformed.

I don't think it will necessarily break. `apply_transform` is not about translating the object, but informing that transform does not need to be applied as it's already been applied before. It's a low-level rendering option that should not influence other higher level logic. It's the responsibility of the caller to make sure `apply_transform` or `apply_matrix` can be used, and it's heavily call-site-dependent (that's why it's not a part of `context`).

kjarosh:
@adrian17 mentioned he wants to review it too, so I'm still holding back with merging

core/src/display_object.rs
```diff
@@ -194,12 +194,16 @@ impl BitmapCache {
 #[derive(Clone)]
 pub struct RenderOptions {
     pub apply_transform: bool,
+    pub skip_masks: bool,
```
adrian17, line 202: Can you please document these fields for the future? (and the existing one too, if possible)

Especially since `apply_transform` and `apply_matrix` overlap in meaning, at least partially, right?

adrian17 (approved):

core/src/display_object.rs
kjarosh: Done!

PR #21524, opened 2025-08-31
avm2: Add call handlers to several error types

SuchAFuriousDeath:


kjarosh:
Please provide a description of changes, why are you doing that, why does it fix the test?

SuchAFuriousDeath:
Sorry, I don't know how I managed to forget again :(.

The test is basically testing that for all these errors, calling Error() is equivalent to calling new Error(). Hence, add the call handlers for these classes and call the constructor in them. 

Before, the test did this: 
image

Lord-McSweeney:
Do `DefinitionError` and `UninitializedError` not have these call handlers?

SuchAFuriousDeath:
I didn't check. The test doesn't test for that.

Lord-McSweeney:
From testing both `DefinitionError` and `UninitializedError` also have call handlers on them in FP. Can you add them to those two classes too?

SuchAFuriousDeath:
Sure

Lord-McSweeney (approved):
LGTM as long as the commits are squashed before we merge

kjarosh (approved):

Issue #2301, opened 2021-01-06
Console looks is just looping continuously because "broadcastMessage is not callableEmbedded SWF (flash) into html doesn't run in Ruffle.


winston895:
Embedded SWF (flash) into html doesn't run in Ruffle, and it was suggested that I post the file(s) here for a developer to look at. (may be the missing puzzle to get Ruffle being more compatible. )

winston895:
[FluidsandElectrolytes.zip](https://github.com/ruffle-rs/ruffle/files/5774343/FluidsandElectrolytes.zip)

winston895:
The FluidsandElectrolytes.zip file contains 1 html file + 1 folder (which contains 6 files including the SWF file).

Poopooracoocoo:
@winston895 did you mean to close your issue btw?

winston895:
Sorry, I did not mean to close the issue. Any help would be appreciated.

EmperorBale:
are there any updates on this issue?

Herschel:
Haven't looked at this deeply yet, but the message itself is not necessarily a problem -- in fact, we've disabled warning logs by default in recent web builds. Calling a non-existent method happens quite often in AVM1 content. Specifically the Flash built-in BroadcasterMX component lazily initializes the broadcastMessage method, so broadcastMessage doesn't exist, which is totally fine -- the call is just a no-op. So the warning message is red herring.

I need to investigate this more, but I suspect this particular instance is another case of #1535. The AS execution stops because of a goto, when it shouldn't because the clip should be kept alive by an unload handler.

In DefineSprite 10009:
```
      if(percent > 99.5)
      {
          // This gotoAndPlay causes execution to stop in Ruffle, but it doesn't in FlashP layer
         _parent._parent.gotoAndPlay(2);
         // Execution does not continue here
         gotoAndStop("off");
         doneLoading = 1;
      }
```

CUB3D:
TLDR: This isn't due to #1535, the header in this swf tells lies.

According to the header in this SWF `frameCount = 2`, in reality there are 77 frames.
When the enterFrame event is dispatched to sprite 10009, it invokes `preload()`, this function contains the code from @Herschel comment  above, which moves the stage to frame 2. After executing frame 2, in Flash Player it then moves to frame 3, however in Ruffle it assumes that there are no more frames because it trusts the header and it loops back to frame 1.

Here's a basic example, runs 3 frames in Flash but only 2 in Ruffle: 
[...]

PR #21526, opened 2025-08-31
core: Allow frames to be played past what was defined in the header

Bale001:
This is an updated version of #17019 & accomplishes the same thing. I decided to make a new PR for this since the old PR was basically abandoned. The following conditions have been addressed:
- a movie with 1 frame does `.play(); .isPlaying == false`
- a movie with 2 header frames but 1 real frame does `.play(); .isPlaying == true` 
- a movie with 1 header frame but 2 real frames does `.play(); .isPlaying == false` but still loops (`enterFrame` event shows cycling `.currentFrame`) 
(from @adrian17)

Furthermore, 4 previously "known failure" tests are now working:
- avm2/issue_8630_placeremoveplace_scriptremove
- avm2/issue_8630_scriptremove
- avm2/movieclip_goto_scene_last_frame_label
- avm2/movieclip_goto_scene_last_frame_int


core/src/display_object/movie_clip.rs
```diff
@@ -561,11 +561,18 @@ impl<'gc> MovieClip<'gc> {
         }
 
         if is_finished {
+            if progress.cur_preload_frame.get() == 1 {
+                // If this clip did not have any show frame tags,
+                // we can just pretend that it did.
+                shared.show_frame(reader, 0).unwrap();
```
Lord-McSweeney, line 567: Just to make sure, `show_frame` gets triggered twice when this path is hit?

core/src/display_object/movie_clip.rs
```diff
@@ -561,11 +561,18 @@ impl<'gc> MovieClip<'gc> {
...
+            }
             // End-of-clip should be treated as ShowFrame
             shared.show_frame(reader, 0).unwrap();
 
             // Flag the movie as fully preloaded when we hit the end of the tag stream.
-            progress.cur_preload_frame.set(shared.total_frames + 1);
+            progress
+                .cur_preload_frame
+                .set(progress.cur_preload_frame.get() - 1);
```
Lord-McSweeney: Can you add a comment explaining why this subtracts one from `cur_preload_frame`?

core/src/display_object/movie_clip.rs
Bale001: Yes, and it's kind of hacky, i'm going to find a better way to do this.

core/src/display_object/movie_clip.rs
```diff
@@ -1156,12 +1169,25 @@ impl<'gc> MovieClip<'gc> {
...
     fn determine_next_frame(self) -> NextFrame {
-        if self.current_frame() < self.total_frames() {
+        let mc = self.0.shared.get();
+        let mut reader = mc.swf.read_from(self.0.tag_stream_pos.get());
+
+        // We ignore the frame count from the header, and instead continue
+        // until we reach the end of the stream or a `TagCode::End`.
+        // Flash Player ignores the frame count, and just executes the full
+        // tag stream before returning to the first frame.
+        if !reader.as_slice().is_empty()
```
Lord-McSweeney, line 1169: Based on the comment, shouldn't this be doing `reader.as_slice().is_empty() || reader.read_tag_code()...`? What happens if the stream doesn't end with an End tag?

core/src/display_object/movie_clip.rs
Bale001: Okay the better way was actually really easy, not sure why I overcomplicated it. Now we only call `show_frame` when we are finished preloading if the clip didn't have any ShowFrame tags. We used to **always** call `show_frame` when the clip was done preloading, but this resulted in `cur_preload_frame` being **sometimes** off by one, depending on whether or not the clip had ShowFrame tags. `

`cur_preload_frame` being off by one was not a problem previously because it would end up just getting set to whatever frame count was defined in the header. But now we can't rely on what was defined in the header, so we need to make sure `cur_preload_frame` is accurate.

core/src/display_object/movie_clip.rs
Bale001: > Based on the comment, shouldn't this be doing `reader.as_slice().is_empty() || reader.read_tag_code()...`? 

Could you elaborate? Not sure what you mean.
> What happens if the stream doesn't end with an End tag?

An SWF not ending with an End tag would indicate that it is corrupted and i'm not sure how Flash responds, would be interesting to see. Also worth mentioning that this code was copied from Aaron's PR

Lord-McSweeney: The comment says that we "instead continue until we reach the end of the stream or a `TagCode::End`". However, the way the conditional is set up is that it continues only if it hasn't reached the end of the stream (`!reader.as_slice().is_empty()`) _and_ the next tag is a `TagCode::End`. If that's the intended behavior, could you edit the comment instead to reflect that?

Bale001: I think you may have misread, it's checking !reader.as_slice().is_empty() and reader.read_tag_code().expect("Failed to read tag") **!=** TagCode::End. This is logically equivalent to the negation of reader.as_slice().is_empty() || reader.read_tag_code().expect("Failed to read tag") == TagCode::End

Lord-McSweeney: Right, sorry, my bad! That makes sense then.

Lord-McSweeney (approved):

Bale001:
Before we merge i'm going to add more regression tests, Adrian was also the one who found the edge cases in Aaron's PR, so i'll give him some time to review as well.

Lord-McSweeney:
@Bale001 does this fix #2301?

Bale001:
> @Bale001 does this fix #2301?

Yes, I just tested that swf and it appears to be working after this PR. I'm thinking since this PR is just fixing an edge case, it's probably fine to merge, and if there are more edge cases that I missed in this PR, we can just fix them on a case by case basis

PR #21536, opened 2025-09-01
tests: Remove default font usage from more tests

kjarosh:
Fixed tests:

* `avm2/bitmapdata_draw`
* `avm2/edittext_always_show_selection`
* `avm2/edittext_get_line_index_of_char`
* `visual/cache_as_bitmap/edittext_selection`
* `visual/cache_as_bitmap/text`

See also https://github.com/ruffle-rs/ruffle/pull/21374


tests/tests/swfs/visual/cache_as_bitmap/text/test.toml
```diff
@@ -1,9 +1,18 @@
 num_frames = 4
 
-[image_comparisons.output]
-tolerance = 0
+[[image_comparisons.output.checks]]
+max_outliers = 12000
```
evilpie, line 8: That seems extremly high?

tests/tests/swfs/visual/cache_as_bitmap/text/fonts.conf
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0"?>
+<!DOCTYPE fontconfig SYSTEM "fonts.dtd">
+<!-- Place in ~/.config/fontconfig/fonts.conf -->
```
evilpie: Okay, I guess this is for running the test in the Adobe Flashplayer? Not obvious at first brush.

tests/tests/swfs/visual/cache_as_bitmap/text/test.toml
kjarosh: It's 1.3%, plus it's one of 2 checks, the other being 0.17%.

The difference comes from the fact that we're rendering device fonts differently. Especially wrt. pixel snapping.

tests/tests/swfs/visual/cache_as_bitmap/text/fonts.conf
kjarosh: That's right, how do you think it should be communicated better?

tests/tests/swfs/visual/cache_as_bitmap/text/test.toml
moulins: This seems worth at least an explanatory comment, then.

tests/tests/swfs/visual/cache_as_bitmap/text/fonts.conf
evilpie: "To re-create the expected output using the flash player, place this file in ~/.config/fontconfig/fonts.conf"

tests/tests/swfs/avm2/edittext_always_show_selection/output.expected.png
evilpie, line 1: So unlike a web browser, we actually create the highlight "background" only as tall as the glyph being selected? Interesting. I feel like for the "underscore" I can't tell if it's working or not. Especiall the last line doesn't look like I expected it to.

tests/tests/swfs/visual/cache_as_bitmap/text/test.toml
kjarosh: Added a comment

tests/tests/swfs/visual/cache_as_bitmap/text/fonts.conf
kjarosh: Done

tests/tests/swfs/avm2/edittext_always_show_selection/output.expected.png
kjarosh: > So unlike a web browser, we actually create the highlight "background" only as tall as the glyph being selected?

No, we render the selection based on ascent and descent which is defined by the highest and the most protruding glyphs. Here that's the case because we have only 4 glyphs and they are the same size.

> I feel like for the "underscore" I can't tell if it's working or not. Especiall the last line doesn't look like I expected it to.

Every line has the same text, so ascent and descent switching places means that the text is being selected.

It's also important to emphasize what we're testing here. The whole ascent, descent, selection rendering thing is being tested in other places. Here we're checking specifically whether the always-on selection appears and how it looks (what color it is).

PR #21537, opened 2025-09-01
avm2: Add `[Ruffle(ConstructOnCall)]` annotation

kjarosh:
This annotation removes a lot of boilerplate code for classes that are constructed when called (i.e. `Error()` instead of `new Error()`).


Lord-McSweeney:
I think you can also add it to `Array`, `Error`, `Function`, and `Namespace`?

Lord-McSweeney (approved):

PR #21552, opened 2025-09-03
avm2: Return object if function constructor returns object

SuchAFuriousDeath:
This is the test case this fixes:

```js
function TestFunction() {
    return arguments;
}

Assert.expectEq(
      "(new TestFunction(0,1,2,3,4,5)).length",
      6,
      (new TestFunction(0,1,2,3,4,5)).length
 );
```

The fix corresponds to 9) from here:
https://262.ecma-international.org/5.1/#sec-13.2.2

### DIFF
--- a/core/src/avm2/object/function_object.rs
+++ b/core/src/avm2/object/function_object.rs
@@ -127,9 +127,15 @@ impl<'gc> FunctionObject<'gc> {
             object_class.instance_vtable(),
         );
 
-        self.call(activation, instance.into(), arguments)?;
+        let result = self.call(activation, instance.into(), arguments)?;
 
-        Ok(instance)
+        // If the constructor returns an object, use that instead of the created instance
+        // TODO: avmplus returns null here if the constructor returns null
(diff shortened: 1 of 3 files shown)


core/src/avm2/object/function_object.rs
```diff
@@ -127,9 +127,14 @@ impl<'gc> FunctionObject<'gc> {
             object_class.instance_vtable(),
         );
 
-        self.call(activation, instance.into(), arguments)?;
+        let result = self.call(activation, instance.into(), arguments)?;
 
-        Ok(instance)
+        // If the constructor returns an object, use that instead of the created instance
+        if let Value::Object(obj) = result {
+            Ok(obj)
```
Lord-McSweeney, line 135: avmplus returns `null` here if the constructor returns `null` (see `FunctionObject::construct`). Since this method doesn't support returning non-Objects yet, can you add this as a TODO here?

SuchAFuriousDeath: Done

PR #21570, opened 2025-09-05
ci: Exclude unrelated crates from coverage

kjarosh:
Currently SWF tests coverage cannot cover some files, that's why this patch excludes them from tests.


torokati44 (approved):

torokati44:
I considered asking whether a positive list of crates to include would be better instead, but probably not...

PR #21571, opened 2025-09-05
core: Clear frame script cleanup queue only on SWF10+

### LINKED ISSUE #20590: udon gallop - regression since nightly 2025-06-08
### Describe the bug

player seems to be spawning under the platform

https://github.com/user-attachments/assets/bd424a5c-273a-4425-904c-480786dc05b7

### Expected behavior

player should not fall under the platform at start

### Content Location

[gallop.zip](https://github.com/user-attachments/files/20670650/gallop.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

_No response_

nivkner:
fixes #20590

while trying to address #20590 i noticed that there are times where `construct_frame` is called outside of the construction phase,
which results in frame scripts triggering when they shouldn't, because `has_pending_script` gets reset.

added a test demonstrating this, which fails before this fix,
and also the full test reproducing the behavior of #20590, on which it is based.

EDIT: noticed the full test failed before https://github.com/ruffle-rs/ruffle/commit/2b6583a7f1fc0c21a4f7dd312f708fc761819b5f (where the regression occurred), added a new test which regresses precisely at that commit.

EDIT 2: added changes that fix the above issue altogether.


nivkner:
tried checking for differences in behavior between all of frame-script tests under SWF9 and SWF10, noticed the queue clearing behavior only appears on SWF10+.

adding version check to queue clearing.

keeping in the phase check, since some of the new `swf_9_*` test variants fail otherwise (maybe the queue clearing hid a false reset of `has_pending_script`)

nivkner:
moved queue clearing to `run_inner_goto_frame`, which already had the right SWF version handling for `goto`.

Lord-McSweeney (approved):

evilpie:
@Lord-McSweeney should we merge this?

PR #21572, opened 2025-09-05
avm2: Dont construct frame in nested 'goto' frame for SWF 9

nivkner:
noticed that, when removing the `construct_frame` from the SWF9 stub of `run_inner_goto_frame`, the test introduced in #12580 actually passes.
probably means the "odd framescript behavior" mentioned in the PR was properly implemented some time later, and now the construction is superfluous.

@Aaron1011 @jarca0123 can you confirm that the games you mentioned in the PR have not regressed with this change?


nivkner:
couldnt find `who_killed_travolta` (doesn't appear on the [flashpoint database](https://flashpointproject.github.io/flashpoint-database/search/)), but i did run [wedding troubles](https://www.superhry.cz/cnt_img/001/1692.swf), which doesn't seem to have regressed between versions (there are pre-existing issues with the `next/skip`  buttons shaking and the animation going too fast, but nothing new).

n0samu:
@nivkner I just checked Who Killed Travolta and it's working fine, thanks! 👍 
I commented on #12580 with a link to the SWF.

PR #21579, opened 2025-09-06
core/avm2: Move orphan code from `Avm2` to new `OrphanManager`

Lord-McSweeney:


adrian17:
To be sure, is this in preparation for some refactor?

Lord-McSweeney:
No, I just made this change because I thought it made sense to keep DisplayObject code separate from the AVM2.

PR #21594, opened 2025-09-08
ci: Add comment on the PR when delta coverage is low

kjarosh:
Low coverage suggests that parts of code are completely untested and prone to break in the future.  By adding a comment for coverage <90% the author and maintainers can assess the severity of low coverage.

### DIFF
--- a/.github/scripts/coverage.py
+++ b/.github/scripts/coverage.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python3
+
+import os
+import subprocess
+import sys
+from datetime import datetime
+import xml.etree.ElementTree as xml
+import json
+
+SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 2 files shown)


.github/scripts/coverage.py
```diff
@@ -0,0 +1,102 @@
...
+By adding tests you contribute to a better understanding of how Flash works \
+and you make sure the work you've done will not disappear.
+
+The code you add *will* be modified in the future by someone else, who won't \
+have the scenario you're implementing in mind.
+
+#### How to fix it?
+
+Add more tests! Currently coverage runs on SWF tests (that test compatibility \
+with Flash Player).
```
Lord-McSweeney: Can you add a link here to the instructions for creating a test? https://github.com/ruffle-rs/ruffle/blob/master/CONTRIBUTING.md#test-guidelines

kjarosh: Done!

PR #21607, opened 2025-09-09
chore: Clean up flake8 lints in release.py

torokati44:


torokati44:
(related TODO: set up Python linting with flake8 and/or black, _because our CI is not slow enough yet!_)

PR #21621, opened 2025-09-10
extension: Remove duplicate calls to `getOptions`

kjarosh:
This hopefully will improve loading time.

### DIFF
--- a/web/packages/extension/src/background.ts
+++ b/web/packages/extension/src/background.ts
@@ -1,5 +1,6 @@
 import * as utils from "./utils";
 import { isMessage } from "./messages";
+import { Options } from "./common";
 
 async function contentScriptRegistered() {
     const matchingScripts = await utils.scripting.getRegisteredContentScripts({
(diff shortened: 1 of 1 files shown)


web/packages/extension/src/background.ts
```diff
@@ -163,10 +164,10 @@ async function disableSWFTakeover() {
     }
 }
 
-async function enable() {
-    const { swfTakeover } = await utils.getOptions();
-    if (swfTakeover) {
-        await enableSWFTakeover();
+async function enable(opt: Options | null = null) {
```
danielhjacobs, line 167: Would this be equivalent (same for `enableSWFTakeover`)?
```js
async function enable(opt?: Options) {
```

danielhjacobs (approved):
This works fine and I don't see a downside and it makes sense, but in testing it seems like the speed difference is negligible.

For testing, I built Ruffle with this commit and Ruffle without this commit (checking out the prior commit), and added some logging.

Namely, in both versions I added this to the start:
```js
let startTime: number | null = null;
```
Added this inside the `if` in the IIFE:
```js
startTime = performance.now();
```
And added this in the `if (options.ruffleEnable)`/`if (ruffleEnable)` in `enableSWFTakeover`:
```js
            if (startTime !== null) {
                const elapsed = performance.now() - startTime;
                console.log(`[Perf] enableSWFTakeover reached ruffleEnable check in ${elapsed.toFixed(2)} ms`);
            }
```

Loading both into Chromium where header conditions are supported, I opened "service worker" in chrome://extensions/ for Ruffle. In the console log for background.js, the time it took to progress through those functions was within the margin of error between the two (toggling the extension on and then off then reopening the service worker log serves to give another data point, and what I mean is the time it took for one was between the min and max time it could take for the other).

Issue #11209, opened 2023-05-25
Incorrect rendering in Fireboy & Watergirl 2 and 3


n0samu:
### Describe the bug

In the second and third Fireboy & Watergirl games ([The Light Temple](https://www.newgrounds.com/portal/view/552273/format/flash?emulate=flash) and [The Ice Temple](https://www.newgrounds.com/portal/view/593216/format/flash?emulate=flash)), the levels are rendered incorrectly, making the games almost impossible to play. I've attached screenshots of the first level of each game below.

![image](https://github.com/ruffle-rs/ruffle/assets/71368227/30b59285-68ce-4a67-bb54-8bcae832de4f)
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/f3c89a4a-4e17-4fc0-9e9f-ab3cd0c4f66b)

### Expected behavior

Here is what the levels are supposed to look like:
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/01839ad3-d8fe-40ea-bffd-34a5fd6dcbf1)
![image](https://github.com/ruffle-rs/ruffle/assets/71368227/4cdd7def-3838-47af-94de-055607bdf4fd)

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

This is probably a masking issue similar to https://github.com/lightspark/lightspark/issues/1007

Aaron1011:
We're not applying the mask when drawing a Displayobject with bitmapdata.draw

Issue #11986, opened 2023-07-10
HitBox: Background color reverts after completing level


n0samu:
### Describe the bug

Link to SWF: http://i.notdoppler.com/files/hitbox.swf

In [HitBox](http://www.notdoppler.com/hitbox.php), the background color changes dynamically using ColorMatrixFilter. This has recently been implemented in Ruffle, but the background color reverts to default after completing a level:

![ruffle_hitbox_endlevel](https://github.com/ruffle-rs/ruffle/assets/71368227/62b8dd29-8974-4386-8e17-9e8d903d8f9c)

### Expected behavior

The background color should stay the same on the end-level screen.

![flashplayer_hitbox_endlevel](https://github.com/ruffle-rs/ruffle/assets/71368227/7433cb6e-a506-4a02-aa05-8c3bfe69b30c)

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

The game is taking a sort of 'screenshot' of the entire stage and drawing it to a BitmapData, which is then having the blur filter applied and being shown behind the end-level screen. According to Dinnerbone, Ruffle does not yet apply filters inside BitmapData.draw.

PR #21624, opened 2025-09-10
core: Render masks in `BitmapData.draw`

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/11209

This patch replaces `render_self` with `render` in `BitmapData.draw` in order to render masks.  It also adds `RenderOptions` in order to pass additional information, e.g. whether to skip applying transform.

`RenderOptions` will also be useful for rendering CAB masks.


Lord-McSweeney:
Does this fix #11986 and related issues?

kjarosh:
> Does this fix #11986 and related issues?

Sadly, it doesn't, because filters still don't work properly, I'll look into it quickly because I might be able to include it in this PR too.

kjarosh:
So filters don't work in `.draw()` because of 1a9bec30620ecc38ca17888f5f93eb63b2627145. I'll have to make sure we don't regress anything when enabling cache draws. It's more complicated, so we'll have to fix it in another PR.

kjarosh:
Added a test for filters that is a known failure for now.

PR #21625, opened 2025-09-10
chore: Reduce size of Tag by boxing the Shape variant

SuchAFuriousDeath:
The size of the Tag enum is now 48 bytes (down from 112).

SoundStreamHead is 16 bytes, Sound and FontV1 are both 32 bytes, so there is no reason to box them in their variants, they are not the bottleneck here.


evilpie:
I find it occasionally useful to add static assertions for the struct size to make sure the size doesn't regress: https://github.com/ruffle-rs/ruffle/blob/28997d30d58a3ec559e0696b42770f6ff96992c2/core/src/avm2/value.rs#L54 Not sure if this is used enough to warrant that though.

kjarosh (approved):

PR #21629, opened 2025-09-11
chore: Move AvmString code into a new crate

Lord-McSweeney:
This should make separating AVM2 from core slightly easier in the future, if we ever do it. This also reduces compile times slightly, especially when re-building after modifying `core` code.

The only downside of this that I can see is that without LTO, some functions from `avm_string` might not get inlined in `core`. However, those functions don't seem to be very hot, so this is probably fine.


kjarosh:
The idea of separating avm string to its own crate sounds good, but I'd say we 100% have to do benchmarks and make sure it doesn't affect performance negatively.

Lord-McSweeney:
I don't see any performance impact from this change (on microbenchmarks for string indexing and concatenation) and I see a ~2 second speedup in recompiling core, so I think this is okay on the performance side.

core/macros/src/lib.rs
```diff
@@ -342,7 +342,7 @@ pub fn atom(item: TokenStream) -> TokenStream {
 pub fn istr(item: TokenStream) -> TokenStream {
     atom_internal(item, |atom| {
         quote!(
-            crate::string::AvmString::from(#atom)
+            ruffle_common::avm_string::AvmString::from(#atom)
```
moulins: Nit: this should be an absolute path: `::ruffle_common::avm_string...`

core/macros/src/lib.rs
```diff
@@ -382,7 +382,7 @@ fn atom_internal(
 
     let mut atom = if let Some(context) = input.context {
         quote!(
-            crate::string::HasStringContext::strings_ref(#context).common().#string_ident
+            ruffle_common::avm_string::HasStringContext::strings_ref(#context).common().#string_ident
```
moulins: Same as above.

core/src/string.rs
```diff
@@ -1,19 +1,6 @@
...
-
-use repr::AvmStringRepr;
-
-pub use ruffle_wstr::*;
-
-pub use avm_string::AvmString;
-pub use common::CommonStrings;
-pub use context::{HasStringContext, StringContext};
-pub use interner::{AvmAtom, AvmStringInterner};
+use ruffle_wstr::{from_utf8, WStr, WString};
```
moulins: I'd keep the other reexports (now from `ruffle_common::avm_string`), to avoid having to update the `use` statements everywhere else.

Issue #9976, opened 2023-03-10
3D Logic 2: Cannot place colors while dragging


n0samu:
### Describe the bug

SWF URL: http://i.notdoppler.com/files/3dlogic2.swf

In [3D Logic 2](http://www.notdoppler.com/3dlogic2.php), it should be possible to fill in a line of squares with a color by clicking and dragging along the path that you want to fill. But in Ruffle, you need to click each individual square to fill it in. Clicking and dragging does not work.

### Expected behavior

It should be possible to fill in multiple squares by clicking and dragging.

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

Firefox 110

n0samu:
Here is a hacked version for easier testing: [3dlogic2_hacked.zip](https://github.com/user-attachments/files/22982916/3dlogic2_hacked.zip)
(The only change is I've made it playable offline/locally)

PR #21661, opened 2025-09-14
avm2: Dispatch hover events while dragging

uaua:
# Dispatch hover events while dragging
## Problem
During a left-button drag, hover transitions are suppressed. AVM2 listeners cannot receive `MouseEvent.MOUSE_OVER`/`MOUSE_OUT` while dragging.

## Reproduction

1. Run [the sample](https://gist.github.com/uaua/d912c516d9efe242d15946e404811eb7) and drag the red square over the green square:

2. Observe the hover behavior while the mouse button is down.

### Expected (Flash Player 32)
The green square becomes semi‑transparent as soon as the cursor enters it during the drag (hover events fire while dragging).

Video:

https://github.com/user-attachments/assets/6e52c6d9-c6c9-458f-bd33-c85ed18dc28f

### Actual (latest master with Google Chrome 140.0.7339.128)
The green square does not change while dragging because hover events aren’t dispatched (mouseOver/mouseOut do not fire during drag).

Video:

https://github.com/user-attachments/assets/50f5e195-d69c-414d-a9b4-5aca77de354a

## Root Cause
Only DragOver/DragOut were sent to the originally pressed object, so AVM2 mouseOver/mouseOut were never dispatched during a drag.

## Solution
When the hovered target changes while the left button is down, enqueue RollOut/RollOver in addition to existing DragOut/DragOver. This is only for AVM2 targets. AVM1 shouldn't be affected.

## Result

Running with Google Chrome 140.0.7339.128

https://github.com/user-attachments/assets/5c2f5f58-850a-44f6-90d9-73e0e11c1022


Lord-McSweeney:
This PR closes #13330.

core/src/player.rs
```diff
@@ -1657,6 +1657,36 @@ impl Player {
                             ));
                         }
                     }
+
+                    // While dragging, dispatch hover roll events only for AVM2 targets.
```
kjarosh, line 1660: I think the comment in line 1632 is outdated now, we aren't ignoring other objects anymore. Could you fix that up before merging?

tests/tests/swfs/avm2/mouse_over_while_dragging/output.txt
```diff
@@ -0,0 +1,4 @@
+startDrag
+OVER buttonDown=true
+OUT buttonDown=true
+stopDrag
```
kjarosh: This output seems wrong, I'm not getting the last line in Flash Player.

tests/tests/swfs/avm2/mouse_over_while_dragging/Test.as
```diff
@@ -0,0 +1,61 @@
...
+            target.mouseChildren = false;
+
+            target.addEventListener(MouseEvent.MOUSE_OVER, function(e:MouseEvent):void {
+                trace("OVER buttonDown=" + e.buttonDown);
+                sawOver = true;
+            });
+            target.addEventListener(MouseEvent.MOUSE_OUT, function(e:MouseEvent):void {
+                trace("OUT buttonDown=" + e.buttonDown);
+                if (sawOver) {
+                    fscommand("quit");
```
kjarosh, line 47: This causes the last line not to be traced. Why is this here? And most importantly, how does the test pass in Ruffle? :D

core/src/player.rs
uaua: Removed the "Other objects are ignored." comment.

tests/tests/swfs/avm2/mouse_over_while_dragging/output.txt
uaua: Thanks, removed the `stopDrag` event as it's focusing on the hover event

tests/tests/swfs/avm2/mouse_over_while_dragging/Test.as
uaua: I'm not confident about the behavior, but seems we run 1 tick and then running all events. So, we get all events which ran within 1 tick. 
The stopDrag part isn’t relevant for this test. I removed the test to avoid confusion.

kjarosh: I think you're right, the test runner has a bug and doesn't consider fscommands between events.

kjarosh (approved):
LGTM, thank you!

n0samu:
I just remembered my old issue #9976 - I wonder if AVM1 needs a similar fix?

uaua:
@n0samu I cannot open the SWF you mentioned, so I cannot test it. Why I explicitly excluded dispatching the ROLL/DRAG events because no ROLL/DRAG events emitted while dragging from outside an object. Tested with [test2.zip](https://github.com/user-attachments/files/22970708/test2.zip) [code](https://gist.github.com/uaua/d100ed5b18a990426451b71717d2e237) 
It only emits `startDrag` and `stopDrag` traces when I drag the red square across the green square with Adobe Flash Player 32. Also, the green square events are not dispatched when I drag across the green square from outside red/green square.

n0samu:
@uaua I uploaded a modified version of the SWF that allows for local testing, does that help? [3dlogic2_hacked.zip](https://github.com/user-attachments/files/22982916/3dlogic2_hacked.zip)

uaua:
@n0samu Raised a PR https://github.com/ruffle-rs/ruffle/pull/21962
Ruffle only emits the hover events to dragging object, so these cells are not colored

n0samu:
Yay thanks so much!! ❤️

PR #21678, opened 2025-09-16
debug_ui: Improve movie list display responsiveness 

anthony-hyo:
- Improve Know Movie List in debug UI by replacing Grid with TableBuilder.
- - Auto Adjust with content/row size. 
- - Better scrolling.
- Improve search and button layout by making it more responsive.
- - Auto Adjust with widow resize.

Preview


n0samu:
Hmm, isn't this a little inconsistent with the design of the other debug tools? Not that that's necessarily a bad thing but I want us to be intentional about the design of the debug tools feature as a whole.

Dinnerbone (approved):
Neat! Thank you!

PR #21689, opened 2025-09-18
avm2: Express Array.removeAt in terms of splice

SuchAFuriousDeath:
The insertremove test seems to indicate that removeAt needs the same resolve_array_hole logic that splice has.
Hence, I created splice_internal that splice, insertAt and removeAt are now expressed in terms of. 

Also, insertAt previously returned an array with the removed elements, which it seems it should not do (The function signature doesn't suggest it should and there are no tests for it). Now it just returns undefined. 

Splice_internal also collects directly into an ArrayStorage instead of an intermediate Vec as that seemed unnecessary.

### DIFF
--- a/core/src/avm2/globals/array.rs
+++ b/core/src/avm2/globals/array.rs
@@ -11,8 +11,10 @@ use crate::avm2::Error;
 use crate::string::AvmString;
 use bitflags::bitflags;
 use ruffle_macros::istr;
+use std::cell::RefMut;
 use std::cmp::{min, Ordering};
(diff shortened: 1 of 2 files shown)


kjarosh:
There were discussions about it with @adrian17. TL;DR the behavior is different for dense/sparse/simple/... arrays, FP has fast paths. We should probably add tests and cover all those cases before we do significant refactors

SuchAFuriousDeath:
I'll take a look at that.

PR #21690, opened 2025-09-18
avm2: Refactor Array.splice to ArrayStorage

kjarosh:
This refactor moves splice implementation from AVM2 to ArrayStorage. The goal was to improve performance of splice, which was only partially successful, as it works better, but the bottleneck is still the weird FP behavior with holes and the prototype.

Nonetheless, this looks like a step in the right direction.

CC @adrian17 @SuchAFuriousDeath


core/src/avm2/array.rs
```diff
@@ -527,6 +527,72 @@ impl<'gc> ArrayStorage<'gc> {
...
+        let end = start + delete_count;
+        match self {
+            ArrayStorage::Dense {
+                storage,
+                occupied_count,
+            } => {
+                let splice = storage
+                    .splice(start..end, items.iter().map(|i| Some(*i)))
+                    .collect::<Vec<_>>();
+                *occupied_count =
```
SuchAFuriousDeath: I think that using Iterator::inspect, you can calculate the occupied count at the same time as collecting into the Vec.

kjarosh: Done!

PR #21698, opened 2025-09-19
core: Fix drawing Stage into a bitmap

kjarosh:
This patch moves `Stage.render`'s logic into `Stage.render_viewport`. The reason for that is you can render Stage as any other display object, and rendering the viewport in `render` results in a completely different behavior for Stage.

This fixes an issue where rendering Stage into a bitmap resulted in the viewport being rendered.


torokati44:
Only somewhat related: I've always had some suspicion about how `render` and `render_self` is done in `Video`. I haven't seen any proof that it causes problems with regular drawing, or transforms, CAB, filters, BMD.draw, or masking or anything like that... It just felt "different" somehow, than the rendering of other DOs.
(Keep in mind that scaling/cropping if the bounds and encoded video data size doesn't match also depends on the codec being used...)

Dinnerbone (approved):

PR #21721, opened 2025-09-21
avm1: Do not reference trivially copyable objects

kjarosh:
It's more efficient and simpler to just copy trivially copyable objects instead of referencing them.

### DIFF
--- a/core/src/avm1/debug.rs
+++ b/core/src/avm1/debug.rs
@@ -35,7 +35,7 @@ impl<'a> VariableDumper<'a> {
         &self.output
     }
 
-    fn object_id(&mut self, object: &Object) -> (usize, bool) {
+    fn object_id(&mut self, object: Object) -> (usize, bool) {
         let ptr = object.as_ptr();
 
         for (i, other) in self.objects.iter().enumerate() {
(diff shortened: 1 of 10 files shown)


SuchAFuriousDeath:
Look like there's a lint for this: https://rust-lang.github.io/rust-clippy/rust-1.90.0/index.html#trivially_copy_pass_by_ref, it's just opt-in. But maybe it would do more bad than good here, not sure.

torokati44:
For the coverage report comment posting failure: https://stackoverflow.com/a/76994510/635587 ...?

Alternatively, you could try using the @RuffleBuild account's token somehow.

kjarosh:
> Look like there's a lint for this: https://rust-lang.github.io/rust-clippy/rust-1.90.0/index.html#trivially_copy_pass_by_ref, it's just opt-in. But maybe it would do more bad than good here, not sure.

I'm currently using this lint to fix this stuff, my plan was to enable it but the biggest issue is that Ruffle is being compiled for various different platforms with different word sizes. We'll see if that's problematic or not.

SuchAFuriousDeath:
> > Look like there's a lint for this: https://rust-lang.github.io/rust-clippy/rust-1.90.0/index.html#trivially_copy_pass_by_ref, it's just opt-in. But maybe it would do more bad than good here, not sure.
> 
> I'm currently using this lint to fix this stuff, my plan was to enable it but the biggest issue is that Ruffle is being compiled for various different platforms with different word sizes. We'll see if that's problematic or not.

The documentation says that the cutoff for when the lint is triggered or not is customizable, if that's of any use.

kjarosh:
> The documentation says that the cutoff for when the lint is triggered or not is customizable, if that's of any use.

That's true, but the issue is that you have some types that depend on word size, some that don't. Setting the cutoff fixes issues for fixed size types, but then breaks for types that depend on word size.

That's why my strategy is to fix obvious issues and then assess whether the lint is useful.

PR #21724, opened 2025-09-21
ci: Fix posting coverage comments

kjarosh:
Let's do some random stuff and hope it works

### DIFF
--- a/.github/workflows/test_rust.yml
+++ b/.github/workflows/test_rust.yml
@@ -164,6 +164,8 @@ jobs:
     if: needs.changes.outputs.should_run == 'true'
     name: Coverage Report
     runs-on: ubuntu-24.04
+    permissions:
+      pull-requests: write
     steps:
       - uses: actions/checkout@v5
         with:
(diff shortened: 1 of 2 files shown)


core/src/player.rs
```diff
@@ -1082,7 +1082,7 @@ impl Player {
                         );
 
                         dumper.print_variables(
-                            "Global Variables:",
+                            "Global Variables 2:",
```
torokati44, line 1085: ?

torokati44: ah, right, testing...

kjarosh: [`a test commit to trigger coverage comment`](https://github.com/ruffle-rs/ruffle/pull/21724/commits/f441a75cd82346bf8b3b912d7193c119d86db531)

torokati44:
I'm not 100% sure that secrets like these are exposed for runs done in forks (for PRs).

Dinnerbone:
> I'm not 100% sure that secrets like these are exposed for runs done in forks (for PRs).

They aren't, this would require each fork to have that secret defined - if it _did_, it would give our secret to anybody.

torokati44:
FWIW: https://github.com/emilk/egui/actions/runs/17862090972/workflow#L58

.github/workflows/test_rust.yml
```diff
@@ -271,13 +271,21 @@ jobs:
...
-          $COVERAGE_SCRIPT comment_report "target/diff-cover/report.json" "${{ github.event.pull_request.number }}" "${RUN_URL}"
+          comment=$($COVERAGE_SCRIPT comment_report "target/diff-cover/report.json" "" "${RUN_URL}")
+          echo "comment=$comment" >> $GITHUB_OUTPUT
+
+      - name: Post comment
+        if: always() && steps.generate_comment.outputs.comment != ''
+        uses: thollander/actions-comment-pull-request@v3
+        with:
+          message: ${{ steps.generate_comment.outputs.comment }}
+        comment-tag: coverage-report
```
torokati44: looks like an indentation issue...?

torokati44:
There are 20k+ lines of `ruffle_core::html::layout: Fallback font not found ... for ..., text will be missing` logged in the coverage test :skull:

torokati44:
The 

```
permissions:
  contents: write
  pull-requests: write
```

part is probably necessary? (present in the SO answer and the linked workflow as well)

torokati44:
My current guess is that in the `egui` repo it works because the workflow that posts the comment is not triggered on the PR itself, but on the completion of such a workflow run.

SuchAFuriousDeath:
> Let's do some random stuff and hope it works

Me when I'm programming:

danielhjacobs:
Per https://github.com/thollander/actions-comment-pull-request?tab=readme-ov-file#permissions

> Note that, if the PR comes from a fork, it will have only read permission despite the permissions given in the action for the `pull_request` event. In this case, you may use the `pull_request_target` event. With this event, permissions can be given without issue (the difference is that it will execute the action from the target branch and not from the origin PR).

PR #21731, opened 2025-09-22
chore: Update `toml` to `0.9.7`, `toml_edit` to `0.23.6`

torokati44:
This got unblocked by:
https://github.com/bkchr/proc-macro-crate/pull/61
https://github.com/bkchr/proc-macro-crate/pull/62

~~I hope the `serde` version requirement in `Cargo.toml` won't be left behind by Dependabot even though this bumps it in `Cargo.lock`.~~ - it did.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1410,9 +1410,9 @@ dependencies = [
 
 [[package]]
 name = "embed-resource"
-version = "3.0.4"
+version = "3.0.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0963f530273dc3022ab2bdc3fcd6d488e850256f2284a82b7413cb9481ee85dd"
+checksum = "55a075fc573c64510038d7ee9abc7990635863992f83ebc52c8b433b8411a02e"
 dependencies = [
  "cc",
  "memchr",
@@ -4048,9 +4048,9 @@ dependencies = [
 
 [[package]]
 name = "proc-macro-crate"
-version = "3.3.0"
+version = "3.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "edce586971a4dfaa28950c6f18ed55e0406c1ab88bbce2c6f6293a7aaba73d35"
+checksum = "219cb19e96be00ab2e37d6e299658a0cfa83e52429179969b0f0121b4ac46983"
 dependencies = [
  "toml_edit",
 ]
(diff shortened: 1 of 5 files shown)


torokati44:
(The motivation was the dep duplication noticed in https://github.com/ruffle-rs/ruffle-android/pull/560.)

torokati44:
Hmm, should we:
 - accept the new error message as expected output
 - report the issue upstream as "invalid key no longer reported as invalid"
 - change the test to something even more invalider
 
 (@Dinnerbone?)
 
 I don't think the TOML spec changed, even with 1.1.0, to allow keys with `~` in them.
 
 For easier access:
 
 ```
     thread 'preferences::read::tests::invalid_toml' panicked at desktop/src/preferences/read.rs:112:9:
    assertion `left == right` failed
      left: "Invalid TOML: TOML parse error at line 1, column 1\n  |\n1 | ~~INVALID~~\n  | ^\ninvalid key\n"
     right: "Invalid TOML: TOML parse error at line 1, column 12\n  |\n1 | ~~INVALID~~\n  |            ^\nkey with no value, expected `=`\n"
```

Multi-line strings expanded:
```
Invalid TOML: TOML parse error at line 1, column 1
  |
1 | ~~INVALID~~
  | ^
invalid key
```
vs.
```
Invalid TOML: TOML parse error at line 1, column 12
  |
1 | ~~INVALID~~
  |            ^
key with no value, expected `=`
```

(And there's also a similar test in `frontend-utils/src/bundle/info.rs`, but with `???` as input.)

**Comment by torokati44** [comment]
Went with the first option for now, please do tell if any of you would prefer something else (as well).

**Comment by kjarosh** [comment]
What's the worst case scenario here? How Ruffle will behave if we cannot parse preferences?

**Comment by torokati44** [comment]
I guess it's just what the test change shows - not exactly the same error message as before, for preference and bundle files with various invalid/broken TOML content.

**Comment by kjarosh** [comment]
So will Ruffle crash in this case? I think it's important not to remove any user data in this case

**Comment by torokati44** [comment]
Just tested - if `~/.config/ruffle/preferences.toml` contains borked nonsense (similar to the test case changed by this PR), it will be silently ignored, and if any preferences are changed, completely overwritten. Exactly the same way on `master` as with this PR.

**Comment by kjarosh** [comment]
Hmmm, we should probably change this behavior. Either backup the preferences or show a popup asking what to do.

I'm worried that changes in TOML parsing can cause such issues and lead to accidental user data deletion.

**Comment by torokati44** [comment]
I highly doubt that any config file written by Ruffle (being valid TOML, not edited by hand) will at any point be considered invalid, and overwritten. And in any case, this would be out of scope for this particular PR.

**Comment by kjarosh** [comment]
I would check if this key works with a value, it may be the case that just the order of validations changed. If it does, we should file an issue, if it doesn't this PR LGTM.

**Comment by moulins** [comment]
Changing the invalid TOML in the test to `~~INVALID~~ = 1` still gives an error:
```
Invalid TOML: TOML parse error at line 1, column 1
  |
1 | ~~INVALID~~ = 1
  | ^
invalid unquoted key, expected letters, numbers, `-`, `_`
```
so this looks indeed like the order of validations changed.

torokati44:
Thank you for your testing and inputs!

PR #21738, opened 2025-09-22
tests: Add visual/define_bits_jpeg2 test

kjarosh:
This covers image decoding code paths in Ruffle.


torokati44:
Now this is thorough, as always - hats off to you! :tophat:
I'd appreciate a little README though, with a list of files and what they are supposed to test and why they are there.
I had to guess that `c0`, `c5`, `c9` is for "Compression Level", `fi` is for "Fast Integer" and `fp` is "Floating Point" IDCT, and `ac` is "Arithmetic Coding", and `rm` is "Restart Markers"... - all by looking at the export dialogs of GIMP... :) But I'm still not sure what `np` could stand for...

torokati44:
One thing to consider - with JPEG having 8x8 blocks for IDCT, the two bottom rows and two rightmost columns will be on different blocks. On some images, you can even see some discoloration on the next-to-last column because of this.

torokati44:
Also, did you make the expected output by screenshotting Flash Player, or by stitching together the input files in an image editor?

tests/tests/swfs/visual/define_bits_jpeg2/test.toml
```diff
@@ -0,0 +1,18 @@
+num_ticks = 1
+
+# FIXME We're badly decoding JPEG images with 4:2:0 and 4:4:0 chroma subsampling
```
torokati44, line 3: If the test pattern was rotated by 90 degrees, perhaps some other subsampling options would (also) look wrong!

torokati44: Always those chroma sample locations...

image

https://www.pcmag.com/encyclopedia/term/chroma-subsampling

image

https://www.sciencedirect.com/topics/engineering/chroma-sample

Notice how there is a difference of horizontal and vertical chroma sample offsets in the different modes.

What I'm also saying is - if the test pattern was magnified just 2x (with nearest interpolation, or even recreated with two-pixel tall gradients) - it would probably look a _lot_ less wrong, even if there's technically a mismatch. And not a lot of real world content is like these images.

kjarosh:
Added a readme, I'll look into including other patterns too

kjarosh:
> Also, did you make the expected output by screenshotting Flash Player, or by stitching together the input files in an image editor?

My outputs are always from Flash Player. In this case I made sure it's the same on different platforms too.

torokati44:
Neat! But guess what - this is what _my_ FP outputs:

image

Difference from the expected output being:

image

PR #21741, opened 2025-09-22
extension: Add a way to set custom JSON config for any options

danielhjacobs:


n0samu:
Just pasting what I wrote on Discord:
> Now that I think about it, I realize this user experience isn't really acceptable :(
people are gonna type random shit in the box and have no idea whether it worked
best solution I can think of is having a dialog with OK and Cancel buttons, and when you clicked OK it would reject the input if invalid (meaning invalid JSON, OR invalid option names/values)
and the validation would be challenging to implement I know. so it might not be worth doing this custom JSON idea at all

Issue #21737, opened 2025-09-22
The inline spaces in texts are much more than expected


hamzaelbarmaki1994:
### Describe the bug

 i notice that the space between lines are much bigger, is this will be reduced after ? some texts get hiddenand  this is critical ,like in this image 
<img width="1656" height="926" alt="image" src="https://github.com/user-attachments/assets/3392fbca-72cb-4265-a437-fb751cec7af1" />

### Expected behavior

the optimal result (can be ignored because it is from Flashplayer and the text has 4 lines Vs 3 line here) :
<img width="913" height="458" alt="image" src="https://github.com/user-attachments/assets/c5e87d5e-30dc-4d0e-9607-122ab1afc871" />
as reference :  before PR is merged https://github.com/ruffle-rs/ruffle/issues/20337: 
<img width="1483" height="886" alt="image" src="https://github.com/user-attachments/assets/f7be63e7-8bbe-4b4e-aada-1a77276b5428" />

### Content Location

https://bibliotheque.sesamath.net/ressource/apercevoir/734

### Affected platform

Online demo

### Operating system

windows 10

### Browser

opera

### Additional information

if possible reduce the inline even more before the [PR](https://bibliotheque.sesamath.net/ressource/apercevoir/734) will be more welcome to make it close to inline in Flashplayer (only if possible)

kjarosh:
That's due to device fonts being used, see https://github.com/ruffle-rs/ruffle/wiki/Frequently-Asked-Questions-For-Users#how-do-i-change-the-fonts-used-by-flash-on-my-website for a current workaround. The needed font is `Tahoma`.

The proper fix requires implementing proper device font lookup on web, which is a pretty difficult topic (but doable).

The content should work properly on Ruffle desktop.

kjarosh:
Here's the SWF for archival purposes: [exo735.zip](https://github.com/user-attachments/files/22472527/exo735.zip)

kjarosh:
@evilpie Could you look into the default font regression? Apparently before your PR the default font had different ascent & descent.

evilpie:
I wonder if by merging in the math glyphs we somehow got bigger bounds. Need to look into that.

kjarosh:
If you've added bigger glyphs (with bigger ascent/descent) then most likely yes.

evilpie:
[...]

PR #21742, opened 2025-09-22
core: Manually fix the descent of the built-in NotoSans font

evilpie:
Merging NotoSansRegular and NotoSansMath takes the bigger descent value from the latter. That unnecessarily increases the space between lines. So we just forcefully set it back to the descent value from NotoSansRegular. This seems to work for #21737.


evilpie:
I need to figure out how to write a test for this.

evilpie:
Oh nice, this changes makes the rendering of those two reference images go back to what we had before. I guess we would call this success, even though to me the rendering looks worse again now.

hamzaelbarmaki1994:
I'm an amateur speaking here, can it be good maybe spread reduction space in half between ascent & descent (can improve quality rendering ?)

kjarosh:
> I need to figure out how to write a test for this.

This test would have to be Ruffle-specific (we should perhaps create a new directory for such tests?).

You can just enable default font in test.toml and add device text to the stage

evilpie:
I lost my previous dev environment so I can't write tests at the moment. But the test changes already show that the line height was reduced.

kjarosh:
Have you reverted the output images to the exact same files as before?

evilpie:
output difference-color-linux-Vulkan

Almost ... This is `stage3d_fractal`

kjarosh:
I think we can just revert to the original images, it should be covered by tolerance

evilpie:
> I think we can just revert to the original images, it should be covered by tolerance

This didn't quite work. I had to add some more outliers.

kjarosh (approved):

PR #21748, opened 2025-09-22
avm1: Interfaces can only be set once per object

moulins:
I'd be surprised if this affected any real-world SWF, but at least the issue (and the fix) is straightforward, in contrast to the other inaccuracies I've found while investigating the `ImplementsOp` and `InstanceOf` opcodes.

Note that the test makes use of a hand-crafted `shim.swf` that provides a function for directly invoking `ImplementsOp` from AS2 code.


Bale001 (approved):
LGTM

PR #21754, opened 2025-09-23
extension: Add per-site configuration settings

anthony-hyo:
## Frontend / UI
- [x] Global settings section
- [x] Advanced settings section
- [x] Per-site rules modal with toggles
- [ ] Review inputs & numbers & selects
- [x] Move inline JS (modal open/close & toggles) to `options.js`

## Backend / Storage
- [ ] Load per-site rules from storage (using `utils.storage`)
- [ ] Save new/edited rules to storage
- [ ] Apply per-site overrides when the player runs
- [ ] Remove or edit site rules dynamically

## Other
- [ ] Validation for input fields (domains, numbers, colors)
- [ ] Integration with actual player settings

## Preview
|  Preview 1 |  Preview 2 |
|---|---|
| Settings | Per-site Rules Modal |

### DIFF
--- a/web/packages/extension/src/options.ts
+++ b/web/packages/extension/src/options.ts
@@ -25,18 +25,18 @@ type SelectOption = {
 };
 
 type SettingGroup = {
-    settingsBasicPlayback: SettingForm[];
-    settingsDisplay: SettingForm[];
-    settingsPerformance: SettingForm[];
-    settingsPlayerConfiguration: SettingForm[];
-    settingsSecurityFeatures: SettingForm[];
-    settingsUserInterface: SettingForm[];
-    settingsFullscreenScaling: SettingForm[];
-    settingMiscellaneous: SettingForm[];
+    "settings-basic-playback": SettingForm[];
+    "settings-display": SettingForm[];
+    "settings-performance": SettingForm[];
+    "settings-player-configuration": SettingForm[];
+    "settings-security-features": SettingForm[];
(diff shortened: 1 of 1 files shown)


danielhjacobs:
I know it can be a little unclear, so in case you haven't pieced together the JS logic yet, working in reverse order, the final step for creating the config is here:

https://github.com/ruffle-rs/ruffle/blob/6e6cda085499629dc4010dcff5847d5c57914f3f/web/packages/extension/src/ruffle.ts#L24-L34

Before that, message.config is sent here:

https://github.com/ruffle-rs/ruffle/blob/6e6cda085499629dc4010dcff5847d5c57914f3f/web/packages/extension/src/content.ts#L191-L200

Explicit options uses this function:

https://github.com/ruffle-rs/ruffle/blob/6e6cda085499629dc4010dcff5847d5c57914f3f/web/packages/extension/src/utils.ts#L91-L103

And `getOptions` is here:

https://github.com/ruffle-rs/ruffle/blob/6e6cda085499629dc4010dcff5847d5c57914f3f/web/packages/extension/src/utils.ts#L81-L86

The options are instantly set on change here:

https://github.com/ruffle-rs/ruffle/blob/f46c744841214e8d1f10b994a6be1a73cca47b55/web/packages/extension/src/common.ts#L172-L176

anthony-hyo:
Some visual updates.

preview3

PR #21755, opened 2025-09-23
core: Store `PreloadProgress.next_preload_chunk` as an `Option<u64>`

Lord-McSweeney:
Previously, we were using `u64::MAX` as a sentinel value


Bale001:
I'm not sure if I entirely agree - Option\<u64\> is larger than a regular u64, and I don't think there's anything wrong with using u64::MAX as a sentinel value. Is there any other reason for this other than for style?

Issue #21773, opened 2025-09-25
Add ARM64 FJCVTZS instruction optimization for f64 to i32


CrazyboyQCD:
Implementation reference: https://github.com/oxc-project/oxc/blob/bc1d716c5da777cdc59a6e35c31140f2e176dfd7/crates/oxc_ecmascript/src/to_int_32.rs#L39-L53

PR #21780, opened 2025-09-26
core: Add ARM64 `FJCVTZS` instruction optimization for `f64` to `i32`

CrazyboyQCD:
Closes https://github.com/ruffle-rs/ruffle/issues/21773.


kjarosh:
Did you do any (reproducible) benchmarks to compare performance between using a generic conversion vs checking runtime feature and using the native conversion?

CrazyboyQCD:
> Did you do any (reproducible) benchmarks to compare performance between using a generic conversion vs checking runtime feature and using the native conversion?

There are some data from [here](https://github.com/oxc-project/oxc/pull/12823#issuecomment-3242718348)

Sorry I don't have a mac to test but there is a [benchamrk setup](https://github.com/Brooooooklyn/jsconv-benchmark).

kjarosh:
Why is the generic f64 to i32 implementation in oxc different than ours? 

Out of those 3 conversions:
* Ruffle's generic
* oxc generic
* fjcvtzs

Either some has to differ or Ruffle's and oxc's has to be equivalent, in which case it doesn't make sense why oxc's is so complex compared to ours.

(Just quickly skimming though it, didn't analyze the differences)

kjarosh:
As a side note: Flash conversions don't always have to match the spec, plus the behavior might be different between 32 and 64 bit runtimes. Do we have extensive testing of this conversion on both runtimes? I'm not sure actually

CrazyboyQCD:
> Why is the generic f64 to i32 implementation in oxc different than ours?

It's originally copied from [boa engine](https://github.com/boa-dev/boa/pull/520/commits/f55f032ef7c0f2d64a43f467fa6efaff70264501), and since both implementations pass the test, this shouldn't be a problem.

CrazyboyQCD:
> As a side note: Flash conversions don't always have to match the spec, plus the behavior might be different between 32 and 64 bit runtimes.

Since this a aarch64 instruction, 32 bit target shouldn't be affected.

kjarosh:
> since both implementations pass the test, this shouldn't be a problem.

So if both impls are equivalent, we cannot use the benchmarks done on oxc's impl, because it's (seemingly) less efficient, so it's expected we'd see better performance of `fjcvtzs` with the feature check there.

kjarosh:
> Since this a aarch64 instruction, 32 bit target shouldn't be affected.

I'm talking about Flash's 32/64 bit runtimes, because there are slight differences in behavior between them, and I recall stumbling upon those differences in conversions.

Ruffle tries to behave as the 32-bit runtime, but we're thinking about letting the user choose which runtime to emulate, so at least we have to be aware of those differences.

CrazyboyQCD:
> So if both impls are equivalent, we cannot use the benchmarks done on oxc's impl, because it's (seemingly) less efficient, so it's expected we'd see better performance of `fjcvtzs` with the feature check there.

We could just replace the implementation with our own, and someone with mac to test.

Assembly reference: 
https://godbolt.org/z/9TqEsWjcW (oxc and ruffle implementation, notice that ruffle version contains a [fmod call](https://github.com/rust-lang/compiler-builtins/blob/master/libm/src/math/generic/fmod.rs))

https://godbolt.org/z/Me1TaP79e (pure fjcvtzs function)

> Ruffle tries to behave as the 32-bit runtime, but we're thinking about letting the user choose which runtime to emulate, so at least we have to be aware of those differences.

Runtime behaviour should depend on how this conversion is used.

core/src/ecma_conversions.rs
```diff
@@ -40,9 +40,49 @@ pub fn f64_to_wrapping_u32(n: f64) -> u32 {
...
+#[allow(unused)]
+#[cfg(target_arch = "aarch64")]
+#[target_feature(enable = "jsconv")]
+/// Converts an `f64` to an `i32` with ECMAScript `ToInt32` wrapping behavior.
+/// The value will be wrapped in the range [-2^31, 2^31).
+/// Optimized for macOS aarch64 with the fjcvtzs instruction.
+unsafe fn f64_to_wrapping_int32_arm64(number: f64) -> i32 {
+    let ret: i32;
+    // SAFETY: fjcvtzs instruction is available on macOS aarch64.
+    std::arch::asm!(
```
linkmauve: You should put this call in an `unsafe {}` block, otherwise you get a warning for it.

core/src/ecma_conversions.rs
```diff
@@ -40,9 +40,49 @@ pub fn f64_to_wrapping_u32(n: f64) -> u32 {
 /// Converts an `f64` to an `i32` with ECMAScript `ToInt32` wrapping behavior.
 /// The value will be wrapped in the range [-2^31, 2^31).
 pub fn f64_to_wrapping_i32(n: f64) -> i32 {
+    #[cfg(all(target_arch = "aarch64", target_os = "macos"))]
```
linkmauve: macOS isn’t the only target which can unconditionally use `fjcvtzs`, `target_feature = "jsconv"` should also be considered here. You can test it on Linux with `RUSTFLAGS='-C target-feature=+jsconv'`, it will produce a binary which only runs on ARMv8.3+ CPUs, but that is a valid configuration if you know every CPU you will run this binary on.

This should probably work:
```rust
#[cfg(all(target_arch = "aarch64", any(target_os = "macos", target_feature = "jsconv")))]
```

Although if like you say macOS always has this feature enabled, maybe this is the same as:
```rust
#[cfg(all(target_arch = "aarch64", target_feature = "jsconv"))]
```

linkmauve:
Also why doesn’t Rust’s `core::arch::aarch64` module support `fjcvtzs` already, behind the same feature flag? It would make sense to open an issue to the stdlib.

core/src/ecma_conversions.rs
```diff
@@ -40,9 +40,49 @@ pub fn f64_to_wrapping_u32(n: f64) -> u32 {
...
+    {
+        // macOS aarch64 always has jsconv feature
+        // SAFETY: macOS aarch64 always supports jsconv
+        unsafe { f64_to_wrapping_int32_arm64(n) }
+    }
+    #[cfg(all(target_arch = "aarch64", not(target_os = "macos")))]
+    {
+        if std::arch::is_aarch64_feature_detected!("jsconv") {
```
moulins, line 46: I worry about the runtime cost of feature detection here, why cannot this be a single `#[cfg(all(target_arch = "aarch64", target_feature = "jsconv"))]`?

core/src/ecma_conversions.rs
moulins: Are we sure this is correct for `NaN` and infinities (which should become 0)? The ARM docs (https://developer.arm.com/documentation/dui0801/h/A64-Floating-point-Instructions/FJCVTZS) are silent on this.

core/src/ecma_conversions.rs
CrazyboyQCD: This is for cross compilation, the cost of detection should not be a problem [since it uses cache internally at runtime](https://github.com/rust-lang/rust/blob/master/library/std_detect/src/detect/cache.rs) and `cfg` macro at compile time.

core/src/ecma_conversions.rs
CrazyboyQCD: > Are we sure this is correct for `NaN` and infinities (which should become 0)? The ARM docs (https://developer.arm.com/documentation/dui0801/h/A64-Floating-point-Instructions/FJCVTZS) are silent on this.

V8 has a set of test for it, which claims that it works properly for Nan and infinity:  https://github.com/v8/v8/blob/3b58a432df85d51bf1c439c30efcf0752de1076f/test/cctest/test-assembler-arm64.cc#L11416C1-L11468C48

core/src/ecma_conversions.rs
```diff
@@ -40,9 +40,53 @@ pub fn f64_to_wrapping_u32(n: f64) -> u32 {
 /// Converts an `f64` to an `i32` with ECMAScript `ToInt32` wrapping behavior.
 /// The value will be wrapped in the range [-2^31, 2^31).
 pub fn f64_to_wrapping_i32(n: f64) -> i32 {
+    #[cfg(target_arch = "aarch64")]
+    #[cfg(target_feature = "jsconv")]
+    {
+        // macOS aarch64 always has jsconv feature
+        // SAFETY: macOS aarch64 always supports jsconv
```
linkmauve: You still have three mentions of macOS in this patch, these are stall comments since this patch now supports all ARMv8.3 systems equally, you should replace them with e.g. “jsconv feature already checked for.”

CrazyboyQCD: Fixed it.

CrazyboyQCD:
Can't figure out why one scrolling test fails on firefox, need help here.

torokati44:
That's flaky, nothing to do with this PR.

core/src/ecma_conversions.rs
kjarosh: 1. Why are you supporting only one-way cross compilation then?
2. I think without benchmarks we have no way of knowing how well this is going to behave. (As mentioned in the main conversation under this PR.)

CrazyboyQCD: @kjarosh 

> 1. Why are you supporting only one-way cross compilation then?

The feature check will be a compile error on non-arm machine, so `target_arch` is needed for that.

> 2\. I think without benchmarks we have no way of knowing how well this is going to behave. (As mentioned in the main conversation under this PR.)

I used ruffle's implmentation for the benchmark above for reference, overall the performance is better.
https://github.com/CrazyboyQCD/jsconv-benchmark/actions/runs/18364018981

kjarosh: > The feature check will be a compile error on non-arm machine, so `target_arch` is needed for that.

I'm referring to _your_ definition of cross compilation. When the target doesn't support jsconv, we are adding a runtime check, but when it does, we directly use the instruction.

If you compile the code on a target that does support jsconv, the code won't work on targets without jsconv. Hence my question: why one-way cross compilation?

kjarosh: > I used ruffle's implmentation for the benchmark above for reference, overall the performance is better. https://github.com/CrazyboyQCD/jsconv-benchmark/actions/runs/18364018981

Benches LGTM, could you also simulate the case of aarch64 without jsconv by flipping the runtime check?

core/src/ecma_conversions.rs
```diff
@@ -40,9 +40,52 @@ pub fn f64_to_wrapping_u32(n: f64) -> u32 {
...
     f64_to_wrapping_u32(n) as i32
 }
 
+#[allow(unused)]
+#[cfg(target_arch = "aarch64")]
+#[target_feature(enable = "jsconv")]
+/// Converts an `f64` to an `i32` with ECMAScript `ToInt32` wrapping behavior.
+/// The value will be wrapped in the range [-2^31, 2^31).
+/// Optimized for aarch64 cpu with the fjcvtzs instruction.
+unsafe fn f64_to_wrapping_int32_arm64(number: f64) -> i32 {
```
kjarosh: Nit: `arm64` is a colloquial name, please use `aarch64`

core/src/ecma_conversions.rs
CrazyboyQCD: > If you compile the code on a target that does support jsconv, the code won't work on targets without jsconv. Hence my question: why one-way cross compilation?

The stdlib doesn't provide mechanism for such situation, it doesn't know if target_feature(enable) code will be hit or not.

kjarosh: I'm sorry, I don't understand your idea (how are you planning to fall back without runtime detection?). But my point was, why even using target_feature when we want to have cross compilation? I agree that a binary should work on hardware with and without jsconv, but I don't see why you're doing it one-way only.

CrazyboyQCD: Sorry I misunderstood how target_feature(enable) works before, and currently I don't have a runtime fallback solution.

kjarosh: But what's the issue with `#[target_feature(enable = "jsconv")]`? I think it's fine to use it provided you use the runtime check. I just don't see a reason for using `cfg!(target_feature = "jsconv")`

CrazyboyQCD: @kjarosh 
I put this code behind a cargo feature because there is not easy way to solve the issue and runtime detection will introduce many complicated code.

kjarosh: But what's the issue with runtime detection?

I don't think a cargo feature will be useful for Ruffle. We usually don't compile different versions for different hardware and the same architecture. It will just end up not being used.

linkmauve: Err, the previous version was correct, the new one is super weird.

You don’t need a cargo feature for that: if the user compiles for ARMv8.3 they mean it and throwing an illegal instruction if ran on an ARMv8.2 or older CPU is fine; if they compile for generic AArch64 the runtime detection was fine too.

You can use `git reflog` to get back in your history and take the previous commit you had.

CrazyboyQCD: > But what's the issue with runtime detection?

The runtime detection is complicated and stdlib doesn't expose it, and [third party crate doesn't provide `jsconv` detection](https://crates.io/crates/cpufeatures), unless we add cpu feature detection specially which is beyond this PR.

> I don't think a cargo feature will be useful for Ruffle. We usually don't compile different versions for different hardware and the same architecture. It will just end up not being used.

Because we don't build for specific hardware(hard for management), I think this is more meaningful for user who want to build from source(haven't synced other cargo config yet).

kjarosh: > The runtime detection is complicated and stdlib doesn't expose it, and [third party crate doesn't provide jsconv detection](https://crates.io/crates/cpufeatures), unless we add cpu feature detection specially which is beyond this PR.

Aren't we literally having a discussion under a line of code that does it?

CrazyboyQCD: > Aren't we literally having a discussion under a line of code that does it?

But it doesn't solve this.

> If you compile the code on a target that does support jsconv, the code won't work on targets without jsconv. Hence my question: why one-way cross compilation?

CrazyboyQCD: > You don’t need a cargo feature for that: if the user compiles for ARMv8.3 they mean it and throwing an illegal instruction if ran on an ARMv8.2 or older CPU is fine; if they compile for generic AArch64 the runtime detection was fine too.

But do we really want to publish binary that will crash for those CPU? Or we need to explain it on `README`.

kjarosh: > > Aren't we literally having a discussion under a line of code that does it?
> 
> But it doesn't solve this.
> 
> > If you compile the code on a target that does support jsconv, the code won't work on targets without jsconv. Hence my question: why one-way cross compilation?

But I wasn't referring to the runtime check, but to the compile check. IMO the compile check doesn't make sense if we want to support cross compilation. Having only runtime check for aarch64 is fine.

CrazyboyQCD: > But I wasn't referring to the runtime check, but to the compile check. IMO the compile check doesn't make sense if we want to support cross compilation. Having only runtime check for aarch64 is fine.

Then it's fine, compile check only works for the target setup and cross compilation needs it.

linkmauve: You definitely want to ship binaries for platforms which always have `jsconv` enabled (you mentioned macOS in early versions of this PR), but for Linux and Android binaries (and maybe Windows, I don’t know which ARMv8 level it targets) you definitely want to ship a generic binary.

Cross-compilation has nothing to do with anything here, you can always decide which target features you want or don’t want to have.

Obviously if you build with `-C target-cpu=native` on an ARMv8.3+ CPU and try to run that binary on an ARMv8 CPU it will crash will illegal instructions, but that’s to be expected and you can’t protect against users shooting themselves in the foot.

kjarosh: Oh, sorry, I think it was me who misunderstood how target_festure works. I thought `-C target-cpu=native` is the default. It's fine as it was before then

CrazyboyQCD: > Benches LGTM, could you also simulate the case of aarch64 without jsconv by flipping the runtime check?

https://github.com/CrazyboyQCD/jsconv-benchmark/actions/runs/18371763313 (not sure why macos target is still faster)

linkmauve: Probably because of hardware differences, we do know that macOS only runs on ARMv8.3+ hardware, and don’t know anything (AFAIK) about the CPUs the Ubuntu and Windows runners are using, maybe they are ARMv8-ARMv8.2 and thus don’t benefit from this PR at all, and even lose a bit of performances due to the failing runtime check for the `jsconv` feature.

You might want to run `lscpu` on Ubuntu to know which core you’re using; I don’t know how to do that on Windows.

core/src/ecma_conversions.rs
```diff
@@ -40,9 +40,47 @@ pub fn f64_to_wrapping_u32(n: f64) -> u32 {
...
+#[allow(unused)]
+fn f64_to_wrapping_i32_generic(n: f64) -> i32 {
     f64_to_wrapping_u32(n) as i32
 }
 
+#[allow(unused)]
+#[cfg(target_arch = "aarch64")]
+#[target_feature(enable = "jsconv")]
+/// Converts an `f64` to an `i32` with ECMAScript `ToInt32` wrapping behavior.
+/// The value will be wrapped in the range [-2^31, 2^31).
```
linkmauve, line 63: `2**31` is out of bounds for i32:
```suggestion
/// The value will be wrapped in the range [-2³¹, 2³¹-1).
```

kjarosh: The doc does not claim that this value is in i32 range, I don't see the issue.

core/src/ecma_conversions.rs
```diff
@@ -40,9 +40,47 @@ pub fn f64_to_wrapping_u32(n: f64) -> u32 {
...
+/// Optimized for aarch64 cpu with the fjcvtzs instruction.
+unsafe fn f64_to_wrapping_int32_aarch64(number: f64) -> i32 {
+    let ret: i32;
+    // SAFETY: fjcvtzs instruction is available under jsconv feature.
+    unsafe {
+        std::arch::asm!(
+            "fjcvtzs {dst:w}, {src:d}",
+            src = in(vreg) number,
+            dst = out(reg) ret,
+        );
```
folkertdev, line 83: just FYI you get slightly better assembly by adding some options here, e.g. `options(nostack, nomem, pure)`.

https://godbolt.org/z/hsPEx96jG

core/src/ecma_conversions.rs
CrazyboyQCD: This is open interval, it‘’s correct.

core/src/ecma_conversions.rs
```diff
@@ -40,9 +40,49 @@ pub fn f64_to_wrapping_u32(n: f64) -> u32 {
...
     f64_to_wrapping_u32(n) as i32
 }
 
+#[allow(unused)]
+#[cfg(target_arch = "aarch64")]
+#[target_feature(enable = "jsconv")]
+/// Converts an `f64` to an `i32` with ECMAScript `ToInt32` wrapping behavior.
+/// The value will be wrapped in the range [-2^31, 2^31).
+/// Optimized for aarch64 cpu with the fjcvtzs instruction.
+unsafe fn f64_to_wrapping_int32_aarch64(number: f64) -> i32 {
```
kjarosh, line 74: Could you add a SAFETY comment that says `jsconv` is required?

core/src/ecma_conversions.rs
```diff
@@ -40,9 +40,49 @@ pub fn f64_to_wrapping_u32(n: f64) -> u32 {
...
+    // TODO: use core intrinsic when https://github.com/rust-lang/rust/issues/147555 is stabilized.
+    #[cfg(target_arch = "aarch64")]
+    {
+        #[cfg(target_feature = "jsconv")]
+        unsafe {
+            f64_to_wrapping_int32_aarch64(n)
+        }
+        #[cfg(not(target_feature = "jsconv"))]
+        if std::arch::is_aarch64_feature_detected!("jsconv") {
+            unsafe { f64_to_wrapping_int32_aarch64(n) }
```
kjarosh, line 48: Could you add a SAFETY comment saying that `jsconv` is guaranteed to be present due to the runtime check?

core/src/ecma_conversions.rs
```diff
@@ -40,9 +40,49 @@ pub fn f64_to_wrapping_u32(n: f64) -> u32 {
 /// Converts an `f64` to an `i32` with ECMAScript `ToInt32` wrapping behavior.
 /// The value will be wrapped in the range [-2^31, 2^31).
 pub fn f64_to_wrapping_i32(n: f64) -> i32 {
+    // TODO: use core intrinsic when https://github.com/rust-lang/rust/issues/147555 is stabilized.
+    #[cfg(target_arch = "aarch64")]
+    {
+        #[cfg(target_feature = "jsconv")]
+        unsafe {
```
kjarosh: Could you add a SAFETY comment saying that jsconv is guaranteed to be present due to the compile time check?

core/src/ecma_conversions.rs
```diff
@@ -40,9 +40,49 @@ pub fn f64_to_wrapping_u32(n: f64) -> u32 {
...
+}
+
+#[allow(unused)]
+fn f64_to_wrapping_i32_generic(n: f64) -> i32 {
     f64_to_wrapping_u32(n) as i32
 }
 
+#[allow(unused)]
+#[cfg(target_arch = "aarch64")]
+#[target_feature(enable = "jsconv")]
```
kjarosh: Could you move those macros below the doc?

kjarosh:
Do you happen to know what's the popularity of jsconv in ARM processors? I guess if it's reasonable (>50% per platform) we can merge this PR.

I think we're pretty safe with Apple hardware here, but what about ARM PCs? What about Android phones?

core/src/ecma_conversions.rs
```diff
@@ -40,9 +40,57 @@ pub fn f64_to_wrapping_u32(n: f64) -> u32 {
...
 /// The value will be wrapped in the range [-2^31, 2^31).
 pub fn f64_to_wrapping_i32(n: f64) -> i32 {
+    // TODO: use core intrinsic when https://github.com/rust-lang/rust/issues/147555 is stabilized.
+    #[cfg(target_arch = "aarch64")]
+    {
+        #[cfg(target_feature = "jsconv")]
+        // SAFETY: `jsconv` feature is checked in compile time to be existed, so it's safe to call.
+        unsafe {
+            f64_to_wrapping_int32_aarch64(n)
+        }
```
folkertdev: this branch is not needed at all: `is_aarch64_feature_detected` will expand to just `true` if the feature is already present at compile time, so just the logic below will statically pick the optimized version when it can, or else check at compile time.

CrazyboyQCD:
@kjarosh 

> Do you happen to know what's the popularity of jsconv in ARM processors? I guess if it's reasonable (>50% per platform) we can merge this PR.
> 
> I think we're pretty safe with Apple hardware here, but what about ARM PCs? What about Android phones?

I didn't find precise data on these, but as an early instruction set extension of ARM64, ARMv8.3 is supported as built-in functions by both GCC and Clang. Additionally, V8, JSC, and SpiderMonkey also use this instruction. I believe the adoption rate is likely quite high.

PR #21816, opened 2025-09-30
web: Log traces into the console during tests

kjarosh:
This should improve debuggability of tests in runners.

### DIFF
--- a/web/packages/selfhosted/test/utils.ts
+++ b/web/packages/selfhosted/test/utils.ts
@@ -106,6 +106,7 @@ export async function setupAndPlay(
             player.__ruffle_log__ = "";
             player.ruffle().traceObserver = (msg) => {
                 player.__ruffle_log__ += msg + "\n";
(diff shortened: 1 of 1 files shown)


torokati44 (approved):
Hmmm, reading `// Get the output, and replace it with an empty string for any future test` a bit below the change, I now wonder whether it's possible that in a single call, we might get multiple lines (the outputs of multiple trace calls) at once, and whether this chunking is deterministic or not...

kjarosh:
> Hmmm, reading `// Get the output, and replace it with an empty string for any future test` a bit below the change, I now wonder whether it's possible that in a single call, we might get multiple lines (the outputs of multiple trace calls) at once, and if so, whether this chunking is deterministic or not...

You're right, we can get multiple lines, and it's not deterministic, I've noticed this too. Currently each test that traces multiple times asynchronously will be flaky due to that.

But that's not the culprit of the flaky mouse wheel test :/

(I'm in the process of fixing this issue)

PR #21817, opened 2025-09-30
web: Fix duplicate console logs in tests

kjarosh:
Subscribing to logs in beforeTest caused multiple subscribers to log the same message, subscribing in beforeSuite fixes the issue.


torokati44 (approved):
I assume this is why we eventually got 20k+ log lines about missing fonts...? Or is that really traced that many times?

kjarosh:
Yeah, it's during normal tests, we log about missing fonts when fonts are truly missing, and unfortunately that happens a lot during tests, because the text is often being laid out before setting the proper font (I guess).

PR #21823, opened 2025-10-01
web: Improve asserting on traces in tests

kjarosh:
Before this patch there could be potential race conditions on traces if they were produced asynchronously by the SWF. The old logic waited for any trace and then asserted on all available traces, which meant that we could include some unintended traces or assert before all traces have been produced.

The new approach ensures that we wait for as many traces as we need.

### DIFF
--- a/web/packages/selfhosted/test/integration_tests/external_interface/test.ts
+++ b/web/packages/selfhosted/test/integration_tests/external_interface/test.ts
@@ -1,7 +1,8 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
 import {
-    getTraceOutput,
+    assertNoMoreTraceOutput,
+    expectTraceOutput,
     injectRuffleAndWait,
     openTest,
     playAndMonitor,
@@ -75,13 +76,10 @@ describe("ExternalInterface", () => {
         await openTest(browser, "integration_tests/external_interface");
         await injectRuffleAndWait(browser);
         const player = await browser.$("<ruffle-object>");
-        await playAndMonitor(
-            browser,
-            player,
-            `ExternalInterface.available: true
-ExternalInterface.objectID: "flash_name"
-`,
-        );
+        await playAndMonitor(browser, player, [
+            "ExternalInterface.available: true",
+            'ExternalInterface.objectID: "flash_name"',
+        ]);
 
         // Set up callbacks
         await browser.execute(() => {
@@ -121,10 +119,9 @@ ExternalInterface.objectID: "flash_name"
             player,
         );
 
-        const actualOutput = await getTraceOutput(browser, player);
-        expect(actualOutput).to.eql(
-            `log called with 2 arguments
-  [
+        await expectTraceOutput(browser, player, [
+            "log called with 2 arguments",
+            `  [
     "Hello world!"
     Object {
         key = "value"
@@ -133,9 +130,8 @@ ExternalInterface.objectID: "flash_name"
           "b"
         ]
       }
-  ]
-`,
-        );
+  ]`,
+        ]);
     });
 
     it("returns a value with legacy API", async () => {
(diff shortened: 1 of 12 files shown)


torokati44:
Does this potentially fix the flakiness? 👀

kjarosh:
Unfortunately no, the current flakiness is not related to traces :/

web/packages/selfhosted/test/integration_tests/external_interface/test.ts
```diff
@@ -75,13 +76,10 @@ describe("ExternalInterface", () => {
...
-            browser,
-            player,
-            `ExternalInterface.available: true
-ExternalInterface.objectID: "flash_name"
-`,
-        );
+        await playAndMonitor(browser, player, [
+            "ExternalInterface.available: true",
+            'ExternalInterface.objectID: "flash_name"',
+        ]);
```
torokati44, line 82: Ah, this looks neater.

web/packages/selfhosted/test/integration_tests/external_interface/test.ts
```diff
@@ -242,20 +230,18 @@ ExternalInterface.objectID: "flash_name"
...
+            "callMethodWithDelay called with 2 arguments",
+            `  [
     "window.RuffleTest.log"
     "Reentrant!"
-  ]
-log called with 1 argument
-  [
+  ]`,
+            "log called with 1 argument",
+            `  [
```
torokati44, line 240: This too.

torokati44:
I 100% welcome this change, just curious (and nitpicky, as usual) - has this race condition ever been observed? Do any of the current tests even make it possible? If not, perhaps "fix" is a bit too strong of a word, but "improve" is definitely warranted. Anyway, the exact phrasing is not that important...

kjarosh:
> I 100% welcome this change, just curious (and nitpicky, as usual) - has this race condition ever been observed? Do any of the current tests even make it possible? If not, perhaps "fix" is a bit too strong of a word, but "improve" is definitely warranted. Anyway, the exact phrasing is not that important...

It hasn't and I'm fairly certain it wouldn't, because tests are written without async traces.

web/packages/selfhosted/test/utils.ts
```diff
@@ -117,30 +116,66 @@ export async function setupAndPlay(
...
+
+export async function clearTraceOutput(
+    browser: WebdriverIO.Browser,
+    player: ChainablePromiseElement,
+) {
+    await browser.execute((playerElement) => {
+        const player = playerElement as Player.PlayerElement;
+        player.__ruffle_log__ = [];
+    }, player);
+}
```
torokati44, line 167: For debugging the flakiness: Is it possible that the initial call to this is sometimes too late, and drops valuable output? :thinking: (I may be way off here, apologies in that case.)

web/packages/selfhosted/test/utils.ts
```diff
@@ -117,30 +116,66 @@ export async function setupAndPlay(
...
+                (player) => {
+                    return (player as Player.PlayerElement).__ruffle_log__;
+                },
+                await player,
             );
+            return log.length >= messageCount;
         },
         {
-            timeoutMsg: "Expected Ruffle to trace a message",
+            timeoutMsg: `Expected Ruffle to trace ${messageCount} messages`,
```
torokati44, line 133: Also for flakiness: Or maybe that some timeout is not long enough somewhere for this?

web/packages/selfhosted/test/utils.ts
```diff
@@ -117,30 +116,66 @@ export async function setupAndPlay(
...
+) {
+    await browser.execute((playerElement) => {
         const player = playerElement as Player.PlayerElement;
-        const log = player.__ruffle_log__;
-        player.__ruffle_log__ = "";
-        return log;
+        const log = player.__ruffle_log__.join("\n");
+        if (log !== "") {
+            throw new Error(`Unexpected trace:\n${log}`);
+        }
```
torokati44: What if there's a single empty string in it? :thinking:

kjarosh: Nice catch, fixed!

web/packages/selfhosted/test/utils.ts
kjarosh: The flakiness is not about tracing, but about not calling the mouse wheel callback

web/packages/selfhosted/test/utils.ts
kjarosh: See https://github.com/ruffle-rs/ruffle/pull/21823#discussion_r2396151308

torokati44 (approved):

torokati44:
Cc: @Dinnerbone (it's okay if you don't have the means to look at this now, I just wanted to notify you, as IIRC this was your stuff initially)

PR #21825, opened 2025-10-01
chore: Enable `trivially_copy_pass_by_ref` clippy lint

kjarosh:
This lint will check if we're referencing trivially copyable objects.

### DIFF
--- a/core/src/avm2/specification.rs
+++ b/core/src/avm2/specification.rs
@@ -14,6 +14,8 @@ use std::fs::File;
 use std::path::Path;
 use std::process::exit;
 
+// This function is used in macros and they require such signature with &bool.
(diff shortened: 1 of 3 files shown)


core/src/avm2/specification.rs
```diff
@@ -14,6 +14,8 @@ use std::fs::File;
 use std::path::Path;
 use std::process::exit;
 
+// We have to use such signature with &bool.
+#[allow(clippy::trivially_copy_pass_by_ref)]
```
torokati44: But the reason for this is not immediately obvious, to me at least...

kjarosh: Elaborated about it in the comment

torokati44 (approved):

PR #21826, opened 2025-10-01
swf: Do not pass AVM2's `Index<T>` as reference

kjarosh:
It's trivially copyable, no idea why our lint didn't catch this.

### DIFF
--- a/swf/src/avm2/types.rs
+++ b/swf/src/avm2/types.rs
@@ -25,11 +25,18 @@ pub struct ConstantPool {
     pub multinames: Vec<Multiname>,
 }
 
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Debug, Eq, PartialEq)]
 pub struct Index<T>(pub u32, pub PhantomData<T>);
 
 // see: https://github.com/rust-lang/rust/issues/26925
-impl<T: Clone> Copy for Index<T> {}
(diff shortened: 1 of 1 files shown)


SuchAFuriousDeath:
Maybe it's the generics? This might be clippy issue material.

kjarosh:
Filed: https://github.com/rust-lang/rust-clippy/issues/15813

kjarosh:
See https://github.com/rust-lang/rust-clippy/issues/15813#issuecomment-3367557774, I wonder why `Copy` impl for `Index` requires `T: Clone`... :thinking: Either `T` should always be `Clone` or it shouldn't be `Clone` at all.

kjarosh:
Fixed the `T: Clone` issue so the clippy lint should work now properly.

CC @adrian17 as you touched the code in the past

SuchAFuriousDeath:
> See [rust-lang/rust-clippy#15813 (comment)](https://github.com/rust-lang/rust-clippy/issues/15813#issuecomment-3367557774), I wonder why `Copy` impl for `Index` requires `T: Clone`... 🤔 Either `T` should always be `Clone` or it shouldn't be `Clone` at all.

It's this: https://github.com/rust-lang/rust/issues/26925

Edit: oops, you already know that.

moulins (approved):

PR #21827, opened 2025-10-01
swf: Implement Copy for small types

kjarosh:
It's better to pass small objects as value than to reference them.

### DIFF
--- a/swf/src/avm1/types.rs
+++ b/swf/src/avm1/types.rs
@@ -176,7 +176,7 @@ pub struct GetUrl<'a> {
     pub target: &'a SwfStr,
 }
 
-#[derive(Clone, Debug, Eq, PartialEq)]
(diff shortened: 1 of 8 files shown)


swf/src/avm1/types.rs
```diff
@@ -288,7 +288,7 @@ pub struct Push<'a> {
     pub values: Vec<Value<'a>>,
 }
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq)]
 pub enum Value<'a> {
```
evilpie, line 292: I don't think this would be smaller than a pointer, if that is what you were going for.

evilpie (approved):

swf/src/avm1/types.rs
kjarosh: Right, reverted

PR #21835, opened 2025-10-03
core: Track PlaceObject's ratio on DisplayObject

kjarosh:
This patch sets PlaceObject's ratio on DisplayObject as it's also used for rewind logic. By itself this patch does not change how rewinds work.

This is pulled from https://github.com/ruffle-rs/ruffle/pull/21336 as a low-risk patch.

### DIFF
--- a/core/src/display_object.rs
+++ b/core/src/display_object.rs
@@ -238,6 +238,7 @@ pub struct DisplayObjectBase<'gc> {
     parent: Lock<Option<DisplayObject<'gc>>>,
     place_frame: Cell<u16>,
     depth: Cell<Depth>,
+    ratio: Cell<u16>,
     name: Lock<Option<AvmString<'gc>>>,
     clip_depth: Cell<Depth>,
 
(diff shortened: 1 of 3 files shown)


core/src/display_object.rs
```diff
@@ -2427,11 +2443,7 @@ pub trait TDisplayObject<'gc>:
                 }
             }
             if let Some(ratio) = place_object.ratio {
-                if let Some(morph_shape) = self.as_morph_shape() {
-                    morph_shape.set_ratio(ratio);
-                } else if let Some(video) = self.as_video() {
-                    video.seek(context, ratio.into());
-                }
+                self.set_ratio(context, ratio);
```
Lord-McSweeney, line 2446: Since this is the only place that ratio is set, and `Video` is the only DisplayObject that implements `on_ratio_changed`, could you avoid adding the new `on_ratio_changed` function and instead just call `video.seek` here directly?

Lord-McSweeney (approved):

core/src/display_object.rs
kjarosh: You mean to leave it as it was before? I could do that, but IMO the code looks cleaner when using polymorphism, it does make sense to seek on `set_ratio` instead of thinking if it's a `Video` or not when setting ratio.

Are there any drawbacks of using `on_ratio_changed` here?

Lord-McSweeney: I have no strong opinion either way, it just seemed to me that the code would be shorter without `on_ratio_changed`

PR #21850, opened 2025-10-05
web: Add possibility to run SWF tests on web

kjarosh:
This patch adds possibility to run SWF tests from //tests/tests/swfs on web.

Currently web-specific renderers like webgl and canvas are largely being untested. This will make it possible in the future to test them properly and make sure they work the same way as wgpu.

This is mostly a (working) draft, because I have to come up with a (sane) way of controlling Ruffle through web for testing (e.g. manual ticking).

Please let me know what's your thoughts about this PR.


n0samu:
How does this compare to #15243?

kjarosh:
> How does this compare to #15243?

Oh that's interesting, I didn't know about that PR.

This PR is a much simpler version of #15243 that leverages existing selfhosted tests for running selected SWF tests. I wonder why #15243 didn't get merged back in the day...

n0samu:
There are a lot of great PRs that seem to just be forgotten :((

n0samu:
Looking at old conversation on Discord, it doesn't look like anyone had any objections - people just didn't get around to reviewing the PR, as often happens

PR #21883, opened 2025-10-11
web: Only show file fetch error after fetch fails on file protocol

danielhjacobs:


web/src/navigator.rs
```diff
@@ -388,9 +368,22 @@ impl NavigatorBackend for WebNavigatorBackend {
...
-                .map_err(|_| ErrorResponse {
-                    url: url.to_string(),
-                    error: Error::FetchError("Got JS error".to_string()),
+                .map_err(|_| {
+                    if url.scheme() == "file" {
+                        match create_specific_fetch_error(
+                            "WASM target can't fetch local URL",
+                            url.as_str(),
+                            "",
+                        ) {
```
evilpie: Can you use `expect_err` here?

web/src/navigator.rs
```diff
@@ -190,17 +190,7 @@ impl NavigatorBackend for WebNavigatorBackend {
...
-                    tracing::error!(
-                        "Can't open the local URL {} on WASM target",
-                        url.to_string()
-                    );
-                    return;
-                } else {
-                    url
-                }
-            }
+            Ok(url) => url,
```
evilpie, line 193: This change is to navigation and not fetch, but I guess still fine.

danielhjacobs: Yeah, I noticed that, but still I thought similarly it doesn't make sense to error this early considering security.fileuri.strict_origin_policy exists in Firefox (whether or not it's advised to use).

web/src/navigator.rs
danielhjacobs: I got this:

```rs
the trait std::fmt::Debug is not implemented for dyn SuccessResponse
```

web/src/navigator.rs
danielhjacobs: Actually, you don't even need to use that.

If you put the ruffle JS and WASM files and the SWF file on a server with an appropriate Access-Control-Allow-Origin header, like a Github Pages site, your HTML can be on the file protocol, embedding the SWF file from the server, while using the script tag for the build of Ruffle from the server, and if you click a link in the SWF that goes to a different relative URL on the file protocol, it will hit this code path with current Ruffle, or with a build built from this PR, it will instead work fine. That is regardless of any about:config changes.

danielhjacobs: As for the safety of removing this code path, I suppose it's a matter of potential debate. I can see an argument for its similarities to https://www.mozilla.org/en-US/security/advisories/mfsa2019-21/#CVE-2019-11730, but on the other hand that CVE was more about the ability to read arbitrary files on the file protocol, not just redirect to them.

web/src/navigator.rs
evilpie: Ah okay, just leave it then.

web/src/navigator.rs
evilpie: Embedding web content from file: is not  a good idea, but I don't think we really need to worry about that? Running stuff on your computer is always more dangerous than going to a web page.

web/src/navigator.rs
```diff
@@ -388,9 +368,22 @@ impl NavigatorBackend for WebNavigatorBackend {
...
-                    error: Error::FetchError("Got JS error".to_string()),
+                .map_err(|_| {
+                    if url.scheme() == "file" {
+                        match create_specific_fetch_error(
+                            "WASM target can't fetch local URL",
+                            url.as_str(),
+                            "",
+                        ) {
+                            Err(e) => e,
+                            Ok(_) => unreachable!("create_specific_fetch_error never returns Ok"),
```
kjarosh: `create_specific_fetch_error` should return `E` instead of `Result<S, E>` then, and callers should wrap it with `Err`

PR #21891, opened 2025-10-11
exporter: Various refactors

kjarosh:
This PR tries to add some structure to exporter code, and moves some code around to cli.rs, progress.rs, and exporter.rs. It removes some code duplication and makes it easier to restructure image processing in future PRs.

### DIFF
--- a/exporter/src/exporter.rs
+++ b/exporter/src/exporter.rs
@@ -1,0 +1,131 @@
+use std::panic::catch_unwind;
+use std::path::Path;
+use std::sync::Arc;
+use std::sync::Mutex;
+
+use image::RgbaImage;
+use ruffle_core::limits::ExecutionLimit;
+use ruffle_core::tag_utils::SwfMovie;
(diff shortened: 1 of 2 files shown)


exporter/src/cli.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+impl FromStr for FrameSelection {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let s_lower = s.to_ascii_lowercase();
+        if s_lower == "all" {
+            Ok(FrameSelection::All)
+        } else if let Ok(n) = s.parse::<u32>() {
```
SuchAFuriousDeath, line 56: This might be intentional, but NonZero\<u32\> implements FromStr directly

kjarosh: Thanks, I'll fix that in a separate PR — I don't really want to include any changes here, just moving code around, it's easier to review

PR #21900, opened 2025-10-12
web: Properly classify web fonts as italic

kjarosh:
Fonts can have style regular and non-zero italic angle. This however should not classify them as italic, even if they might look like it visually.

This makes it behave the same way as on desktop.

### DIFF
--- a/web/src/builder.rs
+++ b/web/src/builder.rs
@@ -480,7 +480,8 @@ impl RuffleInstanceBuilder {
         player.register_device_font(FontDefinition::FontFile {
             name,
             is_bold: face.is_bold(),
-            is_italic: face.is_italic(),
+            // Note: do not use is_italic() here, as we don't care about italic_angle
+            is_italic: face.style() != ttf_parser::Style::Normal,
             data: FontFileData::new(bytes),
             index,
         });


evilpie (approved):

kjarosh:
Also filed an issue to ttf parser about it: https://github.com/harfbuzz/ttf-parser/issues/202

web/src/builder.rs
```diff
@@ -471,7 +471,8 @@ impl RuffleInstanceBuilder {
         player.register_device_font(FontDefinition::FontFile {
             name: name.to_string(),
             is_bold: face.is_bold(),
-            is_italic: face.is_italic(),
+            // Note: do not use is_italic() here, as we don't care about italic_angle
+            is_italic: face.style() != ttf_parser::Style::Normal,
```
torokati44, line 484: I had to look up `ttf_parser::Style` to see what the other "non-normal" styles are. As in - is "bold" in there, etc.
Just a nit of course, but do you think that checking for the (currently existing) other two variants explicitly would be nicer...?

evilpie: What would be the purpose?

torokati44: Being more self-explanatory. Not relying on other variants not mentioned there. Not prompting the reader to look up what those variants are. And again, I don't insist on this - just wanted to note that it wasn't immediately obvious (at least for me, before looking up the docs) that this is correct.

kjarosh: Can we merge it as-is? I agree that it's not super clear, but also this issue is present on desktop already and I think it would be better to merge it than wait till I get to it and fix it up everywhere.

kjarosh: I'm gonna merge it as-is to just get the fix

torokati44: Oof, sorry for not replying to this - yeah it's alright as is.

PR #21904, opened 2025-10-12
tests: Split up some giant files :D

Dinnerbone:
No functional changes, just trying to decimate some giant "do everything" files into smaller chunks that are easier to navigate and read

I'm trying to expand the capabilities of the test runner a little* and whilst I haven't finished that, I've split these changes off to at least make things a bit nicer and easier to work with here

_*My goal is to make the test runner responsible for things like saving the image failures, which needs to be different for web tests_

### DIFF
--- a/tests/framework/src/options/approximations.rs
+++ b/tests/framework/src/options/approximations.rs
@@ -31,7 +31,7 @@ impl Approximations {
             Ok(())
         } else {
             Err(anyhow!(
-                "Approximation failed: expected {}, found {}. Episilon = {:?}, Max Relative = {:?}",
+                "Approximation failed: expected {}, found {}. Epsilon = {:?}, Max Relative = {:?}",
                 expected,
                 actual,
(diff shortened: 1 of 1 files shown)


tests/framework/src/options/approximations.rs
```diff
@@ -0,0 +1,49 @@
...
+                relative_eq!(actual, expected, max_relative = max_relative)
+            }
+            (None, None) => relative_eq!(actual, expected),
+        };
+
+        if result {
+            Ok(())
+        } else {
+            Err(anyhow!(
+                "Approximation failed: expected {}, found {}. Episilon = {:?}, Max Relative = {:?}",
```
moulins: Typo: `Episilon` -> `Epsilon`

Dinnerbone: I just moved code, it's an existing typo! :D

(I'll fix it)

PR #21906, opened 2025-10-12
ci: Raise warnings on uncovered blocks of lines

kjarosh:
This will add warnings visible in GitHub for blocks of uncovered lines.

### DIFF
--- a/.github/scripts/coverage.py
+++ b/.github/scripts/coverage.py
@@ -91,6 +91,41 @@ for more info.
     else:
         print(report)
 
+def raise_warnings(diff_cover_report):
+    """
+    Raise warnings about missing coverage.
+    """
+
+    def print_warning(file, block_start, block_end):
+        if block_start == block_end:
+            message = f"Uncovered line ({block_start})"
+        else:
+            message = f"Uncovered lines ({block_start}–{block_end})"
+        print(f"::warning file={file},line={block_start},endLine={block_end},title=Coverage::{message}")
+
+    with open(diff_cover_report) as f:
+        report_json = json.load(f)
+    src_stats = report_json["src_stats"]
+
+    for file, file_stats in src_stats.items():
+        violation_lines = sorted(file_stats["violation_lines"])
+
+        block_start = None
+        block_end = None
+        for line in violation_lines:
+            if block_start is None:
+                block_start = line
+                block_end = line
+            elif block_end == line - 1:
+                block_end = line
+            else:
+                print_warning(file, block_start, block_end)
+                block_start = line
+                block_end = line
+
+        if block_start is not None:
+            print_warning(file, block_start, block_end)
+
 def main():
     cmd = sys.argv[1]
     log(f'Running command {cmd}')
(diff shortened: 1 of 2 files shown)


torokati44:
Rebased to resolve conflicts after https://github.com/ruffle-rs/ruffle/pull/21838 got merged.

.github/scripts/coverage.py
```diff
@@ -99,6 +135,9 @@ def main():
         pr_number = sys.argv[3] if len(sys.argv) > 3 else ""
         job_url = sys.argv[4] if len(sys.argv) > 4 else ""
         comment_report(diff_cover_report, pr_number, job_url)
+    elif cmd == 'raise_warnings':
+        diff_cover_report = sys.argv[2] if len(sys.argv) > 2 else ""
+        raise_warnings(diff_cover_report)
```
torokati44, line 139: Wouldn't it make sense to error out early, because the parameter to `raise_warnings` doesn't seem to be optional...? OTOH this way it's consistent with the other command invocation, and isn't affected by the internals of the called method...

.github/workflows/test_rust.yml
```diff
@@ -272,6 +272,11 @@ jobs:
         if: always()
         run: cat target/diff-cover/report.md >> $GITHUB_STEP_SUMMARY
 
+      - name: Warnings
+        if: always()
+        run: |
+          $COVERAGE_SCRIPT raise_warnings "target/diff-cover/report.json"
```
torokati44: nit: Why the `|`, why not just a single line, like the diff summary upload above?

.github/scripts/coverage.py
```diff
@@ -91,6 +91,42 @@ def comment_report(diff_cover_report, pr_number, job_url):
...
+            message = f"Uncovered lines ({block_start}–{block_end})"
+        print(f"::warning file={file},line={block_start},endLine={block_end},title=Coverage::{message}")
+
+    with open(diff_cover_report) as f:
+        report_json = json.load(f)
+    src_stats = report_json["src_stats"]
+
+    for file, file_stats in src_stats.items():
+        # We assume the list is sorted
+        violation_lines = file_stats["violation_lines"]
```
torokati44: nit: Wouldn't it be safer to just call `sort` on this? It should be "basically for free", if it is indeed already sorted.

torokati44:
Also, see: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-a-warning-message and https://github.com/kjarosh/ruffle/actions/runs/18450198287?pr=16 :)

.github/scripts/coverage.py
kjarosh: Done

.github/scripts/coverage.py
kjarosh: Done

.github/workflows/test_rust.yml
kjarosh: Fixed

PR #21924, opened 2025-10-14
avm2: Normalize Values where needed

kjarosh:
When differentiating between `Value::Number` and `Value::Integer`, the value should be normalized, because it affects the outcome.


core/src/avm2/globals/array.rs
```diff
@@ -910,7 +910,7 @@ pub fn compare_numeric<'gc, const COMPAT: bool>(
     if COMPAT {
         // See <https://bugzilla.mozilla.org/show_bug.cgi?id=524122>
         // See <https://github.com/adobe/avmplus/blob/858d034a3bd3a54d9b70909386435cf4aec81d21/core/ArrayClass.cpp#L498>
-        if let (Value::Integer(a), Value::Integer(b)) = (a, b) {
+        if let (Value::Integer(a), Value::Integer(b)) = (a.normalize(), b.normalize()) {
```
Lord-McSweeney, line 913: Could you add a test for this?

Lord-McSweeney:
Can you also add this to `amf::serialize_value` and remove the `num >= (1 << 28) || num < -(1 << 28)` check there?

core/src/avm2/globals/array.rs
kjarosh: Done! I'll try to write a test for the goto stuff too

kjarosh: Oh wait, there's no need to add a test for gotos, as it works the same way before and after. It's just more efficient now.

kjarosh:
Done + added a test for AMF

PR #21934, opened 2025-10-15
avm2: Cover special cases in Number.toExponential

kjarosh:
This covers cases like infinite values, zeros, or NaNs.


kjarosh:
Let's wait for https://github.com/ruffle-rs/ruffle/pull/21936, as this test needs fixing

kjarosh:
> Let's wait for #21936, as this test needs fixing

Lol, it was actually a different test with the same name, we're fine then :D

Issue #20244, opened 2025-04-22
Imperfect random number generation in Math.random()


Randomno:
Ruffle currently uses the rand crate for Math.random(), which does not exactly match Flash's implementation. Flash's can be found in the avmplus source: https://github.com/adobe-flash/avmplus/blob/master/core/MathUtils.cpp#L1719

We can create a Flash movie that takes random number samples and reverses the algorithm to find the seed, and manipulate RNG as desired:

![Image](https://github.com/user-attachments/assets/515d9b47-4969-4d75-84df-3adc942720c7)

The second number shown here is generated by Math.random() and will always match the first in Flash Player. In Ruffle, they will not match, unless by chance.

![Image](https://github.com/user-attachments/assets/2f1f652a-8248-4869-b3bb-018f49d2d7c6)

Testing also shows that the implementation is unchanged as far back as Flash Player 6, and is the same in both AVM1 and AVM2.

The main issue raised here is the algorithm rather than the seeding. To note though, in AVM1, the seed is set upon launching Flash Player, whereas in AVM2, it is set upon launching each SWF.

[rngtests.zip](https://github.com/user-attachments/files/19849021/rngtests.zip)

RandomnessTest.swf is a standalone AVM2 test.

RandomnessTest2.swf is a mixed test, and must be run with Helper.swf in the same directory. After running, open Gen.swf to show that the seed is unchanged in AVM1.

Credit to Koong for these tests and research into the algorithm.

Randomno:
The test SWFs also require Flash's behaviour for integer multiplication which Ruffle doesn't match, this can be fixed by replacing multiply opcodes with multiply_i.

PR #21938, opened 2025-10-16
core: Use same RNG algorithm as avmplus

zjk261:
I didn't really understand how rng function is called inside avmplus, but this does fix RandomnessTest.swf from #20244, with some multiply opcodes replaced with multiply_i
[RandomnessTest.zip](https://github.com/user-attachments/files/22940743/RandomnessTest.zip)


core/Cargo.toml
torokati44, line 1: If these are just for seeding, I'm not sure they're worth it. It probably can't be observed/tested properly anyway. And for tests, some determinism feature should be enabled instead.

torokati44:
> with some multiply opcodes replaced with multiply_i

Could you please elaborate on that?

I feel like https://github.com/ruffle-rs/ruffle/pull/21907 could be related...

zjk261:
https://github.com/ruffle-rs/ruffle/issues/20244#issuecomment-2841659420
the original swf still does not produce 2137 with this
so I replaced some multiply opcode with multiply_i with ffdec according to this comment
I know very little about this, sorry

Randomno:
The test requires integer multiplication, and Flash's AVM2 optimizer does this integer multiplication even with the regular `multiply` opcode, while Ruffle doesn't.

https://docs.google.com/presentation/d/1nUp89sZnQAtvsXFkzxyKdsdzYWsZ2OzV2BhlzTeJVsc/edit?slide=id.g2f09243411c_0_10#slide=id.g2f09243411c_0_10

torokati44:
Also, I think the `small_rng` feature of `rand` can be dropped now. Unfortunately, not the entire crate, as `flash.crypto.generateRandomBytes` uses `os_rng` and `std`.

core/src/avm_rng.rs
```diff
@@ -0,0 +1,65 @@
+use crate::locale::get_current_date_time;
+
+// https://github.com/adobe/avmplus/blob/858d034a3bd3a54d9b70909386435cf4aec81d21/core/MathUtils.cpp#L1546
+const C1: i32 = 1376312589;
+const C2: i32 = 789221;
+const C3: i32 = 15731;
+const K_RANDOM_PURE_MAX: i32 = 0x7FFFFFFF;
+
+#[derive(Debug, Clone, Copy, Default)]
```
Lord-McSweeney: Does this struct need to be `Clone`/`Copy`? If not, could you make this struct non-`Copy` and non-`Clone` and add a comment so that we don't accidentally `Copy` or `Clone` it?

Lord-McSweeney (approved):
Can you add the SWF in the issue as a test?

zjk261:
@Randomno can I use modified SWF file as test?

core/src/avm_rng.rs
zjk261: ~Maybe it can be used to reproduce the seed unchanged between swf loads behaviour in original issue...?~

Randomno:
Koong (the author of it) says yes

tests/tests/swfs/avm2/rng/output.txt
```diff
@@ -0,0 +1,3 @@
+743483584
+5402
```
Lord-McSweeney: Are these results consistent for you in Flash Player? For me, the first two numbers outputted are random with each run.

core/src/avm_rng.rs
```diff
@@ -0,0 +1,66 @@
...
+#[derive(Debug, Default)]
+pub struct AvmRng {
+    u_value: u32,
+    u_xor_mask: u32,
+}
+
+impl AvmRng {
+    fn init_with_seed(&mut self, seed: u32) {
+        self.u_value = seed;
+        self.u_xor_mask = 0x48000000;
```
Lord-McSweeney: Since this value never changes, can you make it a `const` instead of a struct field value?

tests/tests/swfs/avm2/rng/output.txt
zjk261: Not consistent

Lord-McSweeney: In that case can you modify the SWF to only output the last number? For this test I think we'd want the output to be reproducible in Flash Player.

Randomno: How about this? Removed most of the extra stuff, just traces true if the two numbers match.

However, it sometimes fails in Flash (as did the original version of the test). I think there's either an error in the algorithm, or Flash is executing stuff weirdly. Maybe should test with a pause between each step.

I tend to get 80% success rate in Flash Player 32.

[test.zip](https://github.com/user-attachments/files/23060805/test.zip)

koonggames: I fixed the consistency issue, it sometimes did not work because a double was not converted to a uint and when the polynomial used in the rng got too large values didn't match up.

[rngtestruffle.zip](https://github.com/user-attachments/files/23063085/rngtestruffle.zip)

zjk261: It seems that there is still a small chance to be false
[flashlog.txt](https://github.com/user-attachments/files/23085928/flashlog.txt)
example
target 7802
probe 312364026
doublecheck 768865056
result 4445

Randomno: Hm yeah, in 1000 attempts I got 7 failures. Target, probe, doublecheck:

7043, 1970344600, 266490418
653, 310941477, 1794949550
6349, 753898449, 1488657303
8564, 749768811, 1196820749
2950, 240060092, 1191158361
1542, 1440380734, 774071793
7838, 1007439541, 429495897

zjk261: maybe triplecheck is needed?

koonggames: As far as I can tell the reversing itself is flawless now, when it breaks it seems to be during the rng calling part. No clue why, I can look into it tomorrow.

zjk261: [Main.zip](https://github.com/user-attachments/files/23111565/Main.zip)
I think this should always work? I tested 1000 times and did not fail

zjk261: @Lord-McSweeney I think it should be consistent now?

Lord-McSweeney (approved):

Dinnerbone (approved):

Dinnerbone:
Thanks everyone for the work that went into this!

kjarosh (approved):

Randomno:
@zjk261 Thank you for implementing this

PR #21950, opened 2025-10-17
chore: Migrate from Node.js 22 to 24

torokati44:
Node.js 22 goes into maintenance mode in 4 days, and 24 takes its place in 11 days:
https://github.com/nodejs/Release/blob/5fbf89de51208b5b54ae4440cf512823de2b672e/README.md#release-schedule

### DIFF
--- a/.github/workflows/test_web.yml
+++ b/.github/workflows/test_web.yml
@@ -109,7 +109,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v5
         with:
-          node-version: 22
+          node-version: 24
           cache: npm
           cache-dependency-path: web/package-lock.json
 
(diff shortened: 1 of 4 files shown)


.github/workflows/test_web.yml
```diff
@@ -164,10 +164,10 @@ jobs:
     steps:
       - uses: actions/checkout@v5
 
-      - name: Setup Node.js 22
+      - name: Setup Node.js 24
```
danielhjacobs, line 167: Not a big deal, but interesting that the browser tests job says the Node.js version in the step name, but  that build for browser tests job does not.

torokati44: Also, sometimes it's `"22"`, other times it's `22`... Another cleanup PR would be nice to get rid of both inconsistencies...

PR #20362, opened 2025-05-06
ci: Update "current" Node.js from 23 to 24

torokati44:
It just got released: https://nodejs.org/en/blog/release/v24.0.0
https://github.com/nodejs/node/pull/57609
https://github.com/nodejs/Release/pull/1088

Node.js 23 goes EOL in about three weeks: https://github.com/nodejs/release#release-schedule

PR #18780, opened 2024-11-27
chore: Update Node.js, LTS: 20 -> 22, Active: 22 -> 23

torokati44:
See: https://nodejs.org/en/about/previous-releases

PR #21953, opened 2025-10-17
web: Test with Node.js 25 instead of 22 (in addition to 24)

torokati44:
Referencing https://github.com/ruffle-rs/ruffle/pull/18780, https://github.com/ruffle-rs/ruffle/pull/20362, and https://github.com/ruffle-rs/ruffle/pull/21950.

This will also necessitate updating the set of required checks for PRs, as usual.

### DIFF
--- a/.github/workflows/test_web.yml
+++ b/.github/workflows/test_web.yml
@@ -35,7 +35,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node_version: ["22", "24"]
+        node_version: ["24", "25"]
         os: [ubuntu-24.04, windows-2025]
 
     steps:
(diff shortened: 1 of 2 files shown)


danielhjacobs:
I don't mind this and it's how we've done updates in the past, but worth noting the odd number releases always have significantly shorter life-cycles than the even number releases. Node 22 will remain the maintenance release for over a year after Node 25 goes entirely out of maintenance: https://nodejs.org/en/about/previous-releases#release-schedule

danielhjacobs:
Still, the point of the tests is to catch breaking issues as early as possible, and the builds are still being done with even number releases, so it's probably fine. There's just the opposite argument to be had about testing for breakages with the maintenance release too. If CI resources were infinite, we'd ideally test both, but :shrug:

torokati44:
I can has merge plz

kjarosh (approved):

Issue #15128, opened 2024-02-05
Swf does not play at all


marcoexo:
### Describe the bug

The file available here https://mariofanuli.com/site_flash/main.swf does not play at all. Some errors are reported in js console

### Expected behavior

File plays correctly in flash player

### Content Location

https://mariofanuli.com/site_flash/main.swf

### Affected platform

Desktop app

### Operating system

Any

### Browser

Chrome

Lord-McSweeney:
This needs `ImportAssets`.

Use the SWF URL `http://mariofanuli.com/site_flash/main.swf` (HTTPS doesn't work, server certificate is expired)

marcoexo:
Thanks for info, but it plays correctly in local flash player on windows. Shouldn't it run in local Shuffle ? It doesn't

PR #16968, opened 2024-07-02
core: Pause preloading until an ImportAssets is finished, + test

Dinnerbone:
Fixes #15128, and likely other things that rely on a "simple" ImportAssets

PR #21954, opened 2025-10-17
core: Block preloading on ImportAssets/2 tags

Dinnerbone:
This is ultimately wrong because our preloading mechanic entirely is wrong, but, it's less wrong than not doing it at all.

The correct thing is probably to treat the ImportAssets/2 tag as the end of the stream until it's resolved

May or may not resolve issues with ImportAssets


Lord-McSweeney:
Does this fix #15128, like #16968? And does the test from that PR pass with this one?

Dinnerbone:
> Does this fix #15128, like #16968? And does the test from that PR pass with this one?

oh my god I forgot I fixed this once before

Dinnerbone:
It does indeed pass the test I made way back when, but I think maybe that other PR was the way to go, it doesn't fix the linked issue. I'll close this and redo that PR

PR #21956, opened 2025-10-17
frontend_utils: Implement bundle exporter

kjarosh:
It's a groundwork for adding a UI to export Ruffle bundles (e.g. on Desktop).


kjarosh:
Low coverage is due to the code not being used in SWF tests, it's covered by dedicated unit tests.

PR #21958, opened 2025-10-17
frontend_utils: Put desktop-specific code behind features

kjarosh:
This patch adds features to frontend-utils so that it's usable on web. Most of the crate is platform-agnostic, except backends, which can depend on filesystem or async executors.


kjarosh:
CC @torokati44 I guess this will need changes on Android (adding those features)

torokati44:
> I guess this will need changes on Android (adding those features)

Aright, will poke at it when CI over there for the bump PRs turn red. Or a couple days later...

torokati44:
There we go: https://github.com/ruffle-rs/ruffle-android/pull/581 :P

PR #21959, opened 2025-10-18
core: Run frames for AVM2 movies loaded into AVM1

### LINKED ISSUE #21007: Issue with mixed avm
### Describe the bug

To my understanding, ActionScript 3 movies being loaded inside of an ActionScript 2 movie doesn't get you very far. However, if said AS3 movie had *no scripts inside of it*, then you should still be able to do some stuff with it from the AS2 movie. Hence this issue.

In this test that I made, `main.swf` (an AS2 movie) will load `item.swf` (an AS3 movie with no scripts). There is also a button in main.swf for stopping the looping animation of item.swf and going to a random frame.

In Ruffle, the animation of item.swf, when loaded into main.swf, does not play at all. Nor does the button work.

### Expected behavior

The animation of item.swf should play, and the button to go to a random frame should be functional.

### Content Location

[as3_inside_as2.zip](https://github.com/user-attachments/files/21225739/as3_inside_as2.zip)

### Affected platform

Desktop app

### Operating system

macOS 15.5

### Browser

_No response_

### Additional information

I first noticed this issue in https://newcp.net/game#/login, where one of the clothing item sprites (an AS3 movie) kept looping infinitely, unaffected by the gotoAndStop calls. I suspect the reason why it doesn't stay at frame 1 the entire time like in this test is because of the root movie being AS3 instead of AS2, but I haven't tested that.

Lord-McSweeney:
Fixes #21007


core/src/orphan_manager.rs
```diff
@@ -85,7 +89,10 @@ impl<'gc> OrphanManager<'gc> {
                 // which have not been moved in the displaylist by ActionScript. Therefore,
                 // any orphan we see that has 'placed_by_script()' should stay on the orphan
                 // list, because it was not removed by a RemoveObject tag.
-                dobj.placed_by_script()
+                //
+                // Also, AVM2 MovieClips with an AVM1 parent are always orphans, since they
+                // would not run their frames if they weren't.
```
Lord-McSweeney: This comment is a bit confusing, I'll clean it up

core/src/display_object/movie_clip.rs
```diff
@@ -1951,6 +1951,14 @@ impl<'gc> MovieClip<'gc> {
         );
     }
 
+    /// Add this `MovieClip` to the global orphan list.
+    #[inline]
+    pub fn add_as_orphan(self, context: &mut UpdateContext<'gc>) {
+        assert!(self.is_avm2_orphan());
```
kjarosh: It would be better to rename this method to `maybe_add_as_orphan` and add the orphan check instead of asserting that it's an orphan.

core/src/display_object/movie_clip.rs
```diff
@@ -1968,7 +1976,11 @@ impl<'gc> MovieClip<'gc> {
         let object =
             Avm2StageObject::for_display_object(context.gc(), display_object, class_object);
 
-        self.set_object2(context, object);
+        self.set_object2(context.gc(), object);
+
+        if self.parent().is_none() {
```
kjarosh: Especially that e.g. here the condition is different, so in the future we'll probably get random panics due to that

Lord-McSweeney: This condition was changed to `self.is_avm2_orphan` in the next commit

core/src/display_object/movie_clip.rs
Lord-McSweeney: I initially wrote the method that way, but I'd prefer to keep it this way. I think it's better that the caller is responsible for checking if it's an orphan before adding it- for example, this way, someone looking at the code of `initialize_for_allocator` can know that the DO is already an orphan, rather than the person writing the code having to explain that the `maybe_add_as_orphan` call is actually guaranteed to add it as an orphan. It's still impossible for non-orphans to be added to the orphan list because of the assertion.

kjarosh: My point is that IMO it's a wrong usage of `assert!` that will only cause panics. The assertion is too far away from code that makes sure this situation won't happen. In case the caller does not check whether it's an orphan, is it appropriate to crash the whole application? I don't think it is, especially that it's enough _not to_ add something as an orphan.

And I'm saying that because such crashes **do** happen in real life, it looks neat now, you're aware of all the callers, etc., but that's not what it will look like in the future.

I see 2 solutions here:

1. we refactor the code to get rid of the assert—that was my original proposal, we don't have to do it this way, we could refactor it in a different way,
2. we change the assert to debug_assert, and add code path for the case when it's not an orphan.

Lord-McSweeney: In that case, what do you think about completely inlining this method into its callers (i.e. directly calling `context.orphan_manager.add_orphan_obj(self.into());`) and removing the assertion?

kjarosh: It's fine by me!

core/src/orphan_manager.rs
```diff
@@ -85,7 +89,11 @@ impl<'gc> OrphanManager<'gc> {
                 // which have not been moved in the displaylist by ActionScript. Therefore,
                 // any orphan we see that has 'placed_by_script()' should stay on the orphan
                 // list, because it was not removed by a RemoveObject tag.
-                dobj.placed_by_script()
+                //
+                // Also, we consider AVM2 MovieClips loaded into an AVM1 parent to always
+                // be orphans, since we know they were placed by AVM1 code and not by
+                // the timeline.
+                dobj.placed_by_script() || has_avm1_parent
```
kjarosh: Shouldn't such movies be placed by script anyway?

core/src/frame_lifecycle.rs
```diff
@@ -72,32 +72,41 @@ pub enum FramePhase {
 pub fn run_all_phases_avm2(context: &mut UpdateContext<'_>) {
     let stage = context.stage;
 
-    if !stage.movie().is_action_script_3() {
-        return;
-    }
+    // We may still have AVM2 orphans that we need to run frames for even though
+    // the root movie is AVM1. However, we don't run any stage phases if the
+    // root movie is AVM1.
+    let is_as3_root = stage.movie().is_action_script_3();
```
kjarosh: So when AVM1 movie has AVM2 movie as a child, the child won't get frame events?

kjarosh: Nit: could you add the word "movie" to the name? When I read this I thought it's about DOs.

E.g. `is_root_movie_as3`, or even better, as this is used for running frame events only: `should_run_frame_events`

core/src/display_object.rs
```diff
@@ -1765,6 +1765,9 @@ pub trait TDisplayObject<'gc>:
     /// Returns the dot-syntax path to this display object, e.g. `_level0.foo.clip`
     #[no_dynamic]
     fn path(self) -> WString {
+        // FIXME this method doesn't work correclty when this DO is from an
```
kjarosh: Can you add a known failure for it and refer to it here? (If it's easy to make one)

core/src/frame_lifecycle.rs
Lord-McSweeney: afaik `run_all_phases_avm2` is specifically for AVM2-observable events, I don't think anything it does is observable by AVM1 (except advancing frames)?

core/src/orphan_manager.rs
Lord-McSweeney: Hmm, from what I understand, `placed_by_script` is only used by AVM2? As in, AVM1 code won't `set_placed_by_script` on DOs that it creates?

core/src/display_object.rs
Lord-McSweeney: I fixed the function and added test coverage to the existing `avm1/mixed_avm` test

core/src/frame_lifecycle.rs
Lord-McSweeney: I've renamed it to `is_root_movie_as3`

core/src/orphan_manager.rs
kjarosh: Oh right, but in this case you have `placed_by_avm1_script`. As a side note, I think we should aim at merging these two.

Lord-McSweeney: Hmm, but actually we don't really care about `placed_by_avm1_script`? As the comment says, we consider AVM2 MovieClips loaded into an AVM1 parent to be orphans no matter what, even if the AVM1 parent was placed by the timeline, because we need that AVM2 movie to tick its frames no matter what.

kjarosh: Maybe I'm misunderstanding something, but the comment says that we do that exactly because the object couldn't have been placed by the timeline. In which case it makes sense to check `placed_by_avm1_script` instead of using this assumption, right?

I'm looking into merging `placed_by_avm1_script` and `placed_by_avm2_script` into one

core/src/frame_lifecycle.rs
kjarosh: Oh, sorry, I think I understood the code now, it's fine

core/src/display_object/movie_clip.rs
```diff
@@ -187,8 +187,6 @@ pub struct MovieClipData<'gc> {
     attached_audio: Lock<Option<NetStream<'gc>>>,
 
     /// The next MovieClip in the AVM1 execution list.
-    ///
-    /// `None` in an AVM2 movie.
```
kjarosh, line 191: Could you elaborate here when it's `None` vs `Some`? IIUC it's `Some` when either this movie clip or the root MC is from AVM1. Because without a comment I'd assume it's for AVM1 only.

core/src/display_object/movie_clip.rs
```diff
@@ -421,8 +419,8 @@ impl<'gc> MovieClip<'gc> {
 
     /// Execute all other timeline actions on this object.
     pub fn run_frame_avm1(self, context: &mut UpdateContext<'gc>) {
-        if !self.movie().is_action_script_3() {
-            // Run my load/enterFrame clip event.
+        // Run my load/enterFrame clip event.
+        if !self.movie().is_action_script_3() || !context.root_swf.is_action_script_3() {
```
kjarosh, line 423: Could you also add a comment here explaining this situation? I.e. why it is being run for an AVM2 movie

core/src/orphan_manager.rs
```diff
@@ -78,14 +82,18 @@ impl<'gc> OrphanManager<'gc> {
                 // indefinitely (if there are no remaining strong references, they will eventually
                 // be garbage collected).
                 //
-                // To detect this, we check 'placed_by_avm2_script'. This flag get set to 'true'
+                // To detect this, we check 'placed_by_script'. This flag get set to 'true'
```
kjarosh, line 85: These 2 changes should be reverted

Lord-McSweeney:
This is the wrong approach- when AVM1 loads an AVM2 movie, we probably want to interpret the loaded movie as actually being AVM1, except for the running of DoAction tags

PR #21962, opened 2025-10-18
avm1: Send DragOver/Out during left-drag; suppress hover in startDrag

uaua:
# Dispatch hover events while dragging
## Problem

In AVM1, `onDragOver` and `onDragOut` are not dispatched during a plain left-drag from outside an object

## Reproduction

Please check with [this SWF](https://github.com/ruffle-rs/ruffle/pull/21661#issuecomment-3417925124). Please go to game screen and try to clicking cells of the cube

### Expected (Flash Player 32)

Cells are colored by dragging

### Actual (0.2.0-nightly.2025.10.18)

Cells are not colored by dragging

## Root Cause

During a plain left-drag (without startDrag), Ruffle didn’t send DragOver/DragOut to hovered clips

## Solution
While left button is down (no startDrag, AVM1)
* send DragOver/DragOut to hovered and non-pressed AVM1 objects; do not send RollOver/RollOut


core/src/player.rs
```diff
@@ -1658,33 +1657,64 @@ impl Player {
...
-                                },
-                            ));
+                    // While dragging, dispatch hover roll/drag events.
+                    // Suppress RollOver/RollOut and DragOver/DragOut events to hovered objects while AVM1 startDrag is active
+                    let suppress_hover_events = context.drag_object.is_some()
+                        && !context.stage.movie().is_action_script_3();
+
+                    if !suppress_hover_events {
+                        // For AVM1, avoid roll events while dragging; only drag events should be emitted.
+                        let allow_roll_while_dragging = context.stage.movie().is_action_script_3();
```
uaua, line 1664: Not to break the current regression tests of `avm1/button_goto`

n0samu:
This would resolve #9976!

PR #21967, opened 2025-10-18
web: Mark `mouse_wheel_avm2` web integration test as flaky, allowing retries

torokati44:
Yes, this is admitting defeat. For now.

I'm not proud, but I'm also not shameful. Just given in.

https://webdriver.io/docs/retry/#rerun-suites-in-mocha
https://mochajs.org/#retry-tests
https://mochajs.org/#arrow-functions

### DIFF
--- a/web/packages/selfhosted/test/integration_tests/mouse_wheel_avm2/test.ts
+++ b/web/packages/selfhosted/test/integration_tests/mouse_wheel_avm2/test.ts
@@ -141,5 +141,5 @@ interface TestParams {
             const player = await browser.$("#objectElement");
             assertNoMoreTraceOutput(browser, player);
         });
-    });
+    }).retries(3); // TODO: figure out and fix flakiness
 });


kjarosh:
:/ my every analysis shows it's impossible, but I'll be trying

torokati44:
How ironic it is that this is failing in Edge. :melting_face: All the retries did...?

torokati44:
```

[MicrosoftEdge 141.0.3537.71 linux #0-48] 1 failing
[MicrosoftEdge 141.0.3537.71 linux #0-48]
[MicrosoftEdge 141.0.3537.71 linux #0-48] 1) Mouse Wheel AVM2, behavior: never scroll the first clip
[MicrosoftEdge 141.0.3537.71 linux #0-48] Expected Ruffle to trace 1 messages
[MicrosoftEdge 141.0.3537.71 linux #0-48] Error: Expected Ruffle to trace 1 messages
[MicrosoftEdge 141.0.3537.71 linux #0-48]     at async getTraceOutput (/home/runner/work/ruffle/ruffle/web/packages/selfhosted/test/utils.ts:122:5)
[MicrosoftEdge 141.0.3537.71 linux #0-48]     at async expectTraceOutput (/home/runner/work/ruffle/ruffle/web/packages/selfhosted/test/utils.ts:155:9)
[MicrosoftEdge 141.0.3537.71 linux #0-48]     at async Context.<anonymous> (/home/runner/work/ruffle/ruffle/web/packages/selfhosted/test/integration_tests/mouse_wheel_avm2/test.ts:92:13)
```

Are we supposed to do some additional cleanup with the trace capturing thingy when retry...? :thinking:

kjarosh:
> Are we supposed to do some additional cleanup with the trace capturing thingy when retry...? 🤔

What do you mean?

torokati44:
That maybe some state is not properly cleaned across retries ... and/if for whatever reason they are more closely related than test cases to each other, in the way they are run.

torokati44:
Or, maybe ... what if, in the cases when the test is failing, it's due to some glitched state of something larger or higher level that does not change across individual tests (connection to the browser, entire harness setup, no idea), therefore, if one try fails, all others will too, no matter how many? 😵‍💫

kjarosh:
I'm not sure about it, failures are in different test cases

torokati44:
So then is 3 just not enough? Vote on putting 5 there instead? Or 10? _It's bikeshed time!!_

torokati44:
Now that the root cause for the flakiness is known (IIUC), I'm not sure this should be done, even temporarily - at risk of it being left like this forever.

danielhjacobs:
Superseded by #21996

PR #21968, opened 2025-10-18
core: Remove all `Loader` variants (except `Movie`) in favor of using `DynamicRoot`s

moulins:
`core/src/loader.rs` is one of the oldest files in the Ruffle codebase, and was written before `gc_arena` introduced `DynamicRoot`s, requiring awkward contortions to access GC'd objects from `OwnedFuture`s.

In particular, the `Loader` variants (except `Movie`) were only useful for ferrying data into `OwnedFuture`s, and can be completely replaced with judicious `DynamicRoot`s wrappers.

To avoid inflating the already-large diff, I put most of the existing loaders into two dummy modules, `ops1` and `ops2`, keeping the code at the same indentation level.

### DIFF
--- a/core/src/loader.rs
+++ b/core/src/loader.rs
@@ -220,7 +220,7 @@ impl From<crate::avm1::Error<'_>> for Error {
 }
 
 /// Holds all in-progress loads for the player.
-pub struct LoadManager<'gc>(SlotMap<LoaderHandle, Loader<'gc>>);
+pub struct LoadManager<'gc>(SlotMap<LoaderHandle, MovieLoader<'gc>>);
 
 unsafe impl<'gc> Collect<'gc> for LoadManager<'gc> {
     fn trace<C: Trace<'gc>>(&self, cc: &mut C) {
@@ -245,7 +245,7 @@ impl<'gc> LoadManager<'gc> {
     /// invalidated). This can be done with remove_loader.
     /// Movie loaders are removed automatically after the loader status is set
     /// accordingly.
-    pub fn add_loader(&mut self, loader: Loader<'gc>) -> LoaderHandle {
+    pub fn add_loader(&mut self, loader: MovieLoader<'gc>) -> LoaderHandle {
         let handle = self.0.insert(loader);
         let loader = self.get_loader_mut(handle).unwrap();
         loader.self_handle = Some(handle);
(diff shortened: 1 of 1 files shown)


core/src/avm2/globals/flash/net/url_loader.rs
```diff
@@ -26,9 +26,9 @@ fn spawn_fetch<'gc>(
 ) -> Result<Value<'gc>, Error<'gc>> {
     let request = request_from_url_request(activation, url_request)?;
 
-    let future = activation.context.load_manager.load_data_into_url_loader(
-        activation.context.player.clone(),
-        loader_object,
+    let future = crate::loader::load_data_into_url_loader(
+        activation.context,
+        loader_object.as_script_object().unwrap(),
```
Lord-McSweeney, line 31: Could you make this pass a `Object` instead of a `ScriptObject`? The latter is technically more precise, but it adds a lot of extra code and goes through an extra useless unwrap+rewrap.

core/src/loader.rs
```diff
@@ -510,43 +508,40 @@ pub enum MovieLoaderVMData<'gc> {
...
-pub enum Loader<'gc> {
-    /// Loader that is loading a new movie into a MovieClip.
-    Movie {
-        /// The handle to refer to this loader instance.
-        #[collect(require_static)]
-        self_handle: Option<LoaderHandle>,
-
-        /// The target movie clip to load the movie into.
-        target_clip: DisplayObject<'gc>,
+pub struct Loader<'gc> {
```
Lord-McSweeney: Could you rename `Loader` to `MovieLoader` after this change (in this PR or in a follow-up)?

core/src/avm2/globals/flash/net/url_loader.rs
moulins: Passing an `Object` would require adding a `ObjectHandle` type, which itself requires either:
- double-boxing the handle, which is unstatisfying;
- adding a `FooHandle` for all objects variants;
- only support the `Object` variants which have a `Handle`, like it was done for `DisplayObjectWeak`.

IMHO, any of theses solutions is extra code and complexity that isn't warranted.

core/src/loader.rs
moulins: Done!

core/src/avm1/object/script_object.rs
```diff
@@ -73,6 +75,20 @@ impl fmt::Debug for ObjectWeak<'_> {
...
+#[derive(Clone)]
+#[expect(clippy::type_complexity)]
+pub struct ObjectHandle(DynamicRoot<Rootable![RefLock<ObjectData<'_>>]>);
+
+impl ObjectHandle {
+    pub fn stash<'gc>(context: &UpdateContext<'gc>, this: Object<'gc>) -> Self {
+        Self(context.dynamic_root.stash(context.gc(), this.0))
+    }
+
+    pub fn fetch<'gc>(&self, context: &UpdateContext<'gc>) -> Object<'gc> {
```
Lord-McSweeney, line 87: (non-blocking) Looking at this again, it might be better if the `fetch` methods were associated methods (`XXXHandle::fetch`) so that it's clear that they correspond to the `stash` operation

PR #21983, opened 2025-10-21
web: Note Ruffle does not collect user data (Firefox)

danielhjacobs:
https://extensionworkshop.com/documentation/develop/firefox-builtin-data-consent/#specifying-data-types

https://github.com/ruffle-rs/ruffle/actions/runs/18668416338/job/53224480952#step:17:31

> description: '"/browser_specific_settings/gecko/data_collection_permissions" property will be required in the future. Please add this key to the manifest.

### DIFF
--- a/web/packages/extension/webpack.config.js
+++ b/web/packages/extension/webpack.config.js
@@ -49,7 +49,7 @@ function transformManifest(content, env) {
                 id: firefoxExtensionId,
                 data_collection_permissions: {
                     required: ["none"]
-                }
+                },
             },
         };
         manifest.background = {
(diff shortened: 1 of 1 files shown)


danielhjacobs:
Note that I'm a bit worried this will make Ruffle stop working on older Firefox, especially ESR versions, as "The “data_collection_permissions” property is reserved for future usage and cannot be used in an extension at the moment." is a potential error per https://discourse.mozilla.org/t/the-data-collection-permissions-property-is-reserved/145449

evilpie:
That forum post was purely about getting an error when uploading an extension with that key to https://addons.mozilla.org, because at that point submitting extension with this new feature wasn't supported yet.

Locally when installing the extension in Firefox unknown keys should be completely ignored.

evilpie (approved):

Dinnerbone (approved):

PR #21990, opened 2025-10-21
core: Replace unmaintained `unic-segment` crate with `unicode-segmentation`

torokati44:
Due to https://rustsec.org/advisories/RUSTSEC-2025-0098.

This was surprisingly trivial. I'm not sure how well this is covered by CI, or how to test it well manually.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4455,7 +4455,7 @@ dependencies = [
  "tracing",
  "tracy-client",
  "ttf-parser",
- "unic-segment",
+ "unicode-segmentation",
  "url",
  "wasm-bindgen-futures",
  "weak-table",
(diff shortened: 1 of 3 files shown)


kjarosh:
This is not covered by tests, so needs manual testing. I'm not sure why it shows coverage though

kjarosh:
In order to test manually, you need to have a text field with words and go left/right with Ctrl and check if it works the same way as before. (It should advance by words.)

torokati44:
Some _cursory_ testing using `tests/tests/swfs/visual/edittext/edittext_border_basic/test.swf` (entering some "words with acćeňtėd text" and ctrl+arrow moving in there) showed identical behavior as far as I saw.

PR #21994, opened 2025-10-22
desktop: Add an option to export Ruffle bundles

kjarosh:
This PR adds an option `File > Export...` that allows exporting the currently playing content as a Ruffle bundle.

In case of remote content, the bundle will refer to that content, in case of local content, the export dialog will ask the user which files to include when exporting.

This is a very basic export feature, which should cover major use cases and allow users to create bundles which work OOTB. Of course this can be improved in the future by e.g. including remote files in the bundle, providing more export options, etc.

image
<br/>
image

### DIFF
--- a/desktop/assets/texts/en-US/main_menu.ftl
+++ b/desktop/assets/texts/en-US/main_menu.ftl
@@ -16,6 +16,7 @@ file-menu-recents = Recents
 file-menu-recents-empty = No recent entries
 file-menu-preferences = Preferences...
 file-menu-exit = Exit
+file-menu-export = Export...
 
 controls-menu = Controls
 controls-menu-suspend = Suspend
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -577,6 +577,10 @@ impl ApplicationHandler<RuffleEvent> for App {
                 main_window.gui.close_movie(&mut main_window.player);
             }
 
+            (Some(main_window), RuffleEvent::ExportBundle) => {
+                main_window.gui.export_bundle();
+            }
+
             (Some(main_window), RuffleEvent::EnterFullScreen) => {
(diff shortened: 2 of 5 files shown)


frontend-utils/src/bundle/exporter/helpers.rs
```diff
@@ -0,0 +1,314 @@
...
+            .first()
+            .ok_or(FilesystemHelperError::NoFilesToExport)?
+            .as_ref()
+            .parent()
+            .ok_or(FilesystemHelperError::NoCommonDirectory)?
+            .to_owned();
+
+        for file in files {
+            if !file.as_ref().is_absolute() {
+                return Err(FilesystemHelperError::NonAbsolutePaths);
```
danielhjacobs, line 63: On Windows, the tests with "/a/b/c", etc. are hitting this error, I think because that's not the format of an absolute path on Windows.

kjarosh:
@Dinnerbone any idea on how to approach the Windows issue here? IIUC Rust library cannot create a file URL we expect on Windows (e.g. `file:///a/b/c/d`) because it requires the path to be absolute (i.e. have a prefix), but when a path has a prefix it always gets into the URL.

Dinnerbone:
No immediate ideas sorry :( We've ran into similar-ish issues before and it usually leaves me questioning if we should be using Url at all :(

Perhaps try constructing the Url manually based on segments of the path?

kjarosh:
> We've ran into similar-ish issues before and it usually leaves me questioning if we should be using Url at all :(

You mean `Url` or `Path`? I don't think it's an issue with `Url` specifically, it's system independent, it's just paths that are system-dependent (especially non-unix ones). And in Rust, `Path` is always backed by a filesystem unfortunately.

> Perhaps try constructing the Url manually based on segments of the path?

Done

PR #21996, opened 2025-10-22
web: Fix flakiness of `mouse_wheel_avm2`

kjarosh:
wdio was randomly causing the browser to fire `pointerenter` and `pointerleave` events, which resulted in the test being flaky. When `pointerleave` event was sent before moving the mouse, there was no hovered object and scroll was not dispatched to the object it should have been.

This patch uses wdio's `.moveTo()` to update the hover before dispatching scroll event, which fixes this issue.

* Supersedes https://github.com/ruffle-rs/ruffle/pull/21967

I ran a suite of hundreds of iterations of these tests and all passed. Before this patch it was almost guaranteed to get a failure after 50 iterations.

### DIFF
--- a/web/packages/selfhosted/test/integration_tests/mouse_wheel_avm2/test.ts
+++ b/web/packages/selfhosted/test/integration_tests/mouse_wheel_avm2/test.ts
@@ -19,16 +19,15 @@ async function scroll(
     lines: number,
 ) {
     const canvas = await player.shadow$("canvas");
+    const canvasSize = await canvas.getSize();
+
+    const xOffset = x - canvasSize.width / 2;
+    const yOffset = y - canvasSize.height / 2;
+    await canvas.moveTo({ xOffset, yOffset });
 
     return await browser.execute(
-        (element, x, y, lines) => {
+        (element, lines) => {
             const el = element as unknown as HTMLElement;
-            el.dispatchEvent(
-                new PointerEvent("pointermove", {
-                    clientX: x as unknown as number,
-                    clientY: y as unknown as number,
-                }),
-            );
             return el.dispatchEvent(
                 new WheelEvent("wheel", {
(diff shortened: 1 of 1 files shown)


web/packages/selfhosted/test/integration_tests/mouse_wheel_avm2/test.ts
```diff
@@ -19,16 +19,15 @@ async function scroll(
     lines: number,
 ) {
     const canvas = await player.shadow$("canvas");
+    const canvasSize = await canvas.getSize();
+
+    const xOffset = x - canvasSize.width / 2;
+    const yOffset = y - canvasSize.hright / 2;
```
danielhjacobs: `hright`

kjarosh: ... how the heck does it work...

I guess the resulting y value is wrong, but it doesn't matter because we have vertical stripes

danielhjacobs: I believe this means yOffset is always NaN, so yeah, weird (if y is 3, in JS 3 - undefined / 2 is NaN)

danielhjacobs: The `y` used by the pointermove action caused by wdio's `moveTo` when `yOffset` is NaN, which is falsy, is 0, so I guess this test just works as expected with a proper `x` but a `y` of 0:

https://github.com/webdriverio/webdriverio/blob/25f398b666c97934b1bbd69c338d70b9c2f714c6/packages/webdriverio/src/commands/element/moveTo.ts#L41

kjarosh: Fixed!

Issue #15658, opened 2024-03-20
nick wacky racers very laggy and screen is frozen


markman12345:
### Describe the bug

like i said before on the top of screen

### Expected behavior

for the game not to be so laggy and frozen like

### Content Location

https://www.numuki.com/game/nick-wacky-racers-3d/

### Affected platform

Browser's extension

### Operating system

windows 11

### Browser

microsoft edge

PR #21999, opened 2025-10-23
avm2: Improve ability of optimizer to remove obfuscation in code

Lord-McSweeney:
Testing on a real-world obfuscated box2d SWF shows a ~10% reduction in AVM2 ops executed and a 1-5% performance improvement. Some obfuscated SWFs (e.g. #15658) include many always- or never-taken branches. The optimizer can now optimize these out. For example, with this PR, the optimizer can entirely remove both unreachable blocks in the following (contrived) code:
```
package {
    public class Test {
        public function Test() {
            if (this) {
                trace("reachable");
            } else {
                trace("unreachable");
            }

            var a:Boolean = false;
            if (a) {
                trace("unreachable");
            }
        }
    }
}
```

This doesn't change the amount of time spent performing optimizations by much.

### DIFF
--- a/core/src/avm2/optimizer/type_aware.rs
+++ b/core/src/avm2/optimizer/type_aware.rs
@@ -257,6 +257,11 @@ impl<'gc> OptValue<'gc> {
     }
 }
 
+// This type is used throughout the abstract interpreter, make sure it doesn't
+// grow too much
+#[cfg(target_pointer_width = "64")]
+const _: () = assert!(std::mem::size_of::<OptValue>() == 16);
+
 impl std::fmt::Debug for OptValue<'_> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
         f.debug_struct("OptValue")


core/src/avm2/optimizer/dce.rs
```diff
@@ -0,0 +1,51 @@
...
+    for (i, op) in ops.iter().enumerate() {
+        match op.get() {
+            Op::Jump { offset } => {
+                let mut is_reachable = false;
+                for position in i..offset {
+                    if jump_targets.contains(&position) {
+                        // If this position (inside the ops being jumped over) is
+                        // unreachable, we can exit
+                        is_reachable = true;
+                    }
```
moulins: Nit: this can be rewritten to
```rs
let is_reachable = (i..offset).any(|pos| jump_targets.contains(&pos);
```
which has the advantage of being short-circuiting.

core/src/avm2/optimizer/dce.rs
```diff
@@ -0,0 +1,51 @@
...
+                }
+
+                if !is_reachable {
+                    for position in i..offset {
+                        // Set all unreachable ops to Nop, including the Jump op
+                        ops[position].set(Op::Nop);
+                    }
```
moulins: Nit: why not this? it would make clippy happy without sacrificing readability.
```rs
for op in &position[i..offset] {
  op.set(Op::Nop);
}
```

core/src/avm2/optimizer/dce.rs
```diff
@@ -0,0 +1,51 @@
...
+                }
+            }
+            Op::PopJump { offset } => {
```
moulins, line 21: The code duplication between the two match arms is a little unfortunate, is there no way to merge them?

Lord-McSweeney: With your previous suggestion the duplicated code is fairly short. I think it's more readable this way than having a second `match` that determines whether to set the first op to `Pop` or not, do you think they should still be merged?

moulins: Hm, you're right, it's fine now ^^

PR #22006, opened 2025-10-23
demo: Add back ability to use query parameters to load specific sample

danielhjacobs:


web/packages/demo/src/navbar/samples.tsx
```diff
@@ -63,15 +68,38 @@ export function SampleSelection({
...
+                    const fileParam = params.get("file");
+
+                    let sampleIndex = 0;
+
+                    if (fileParam) {
+                        sampleIndex = data.swfs.findIndex(
+                            (swf) =>
+                                swf.location === fileParam ||
+                                swf.location.replace(/^swfs\//, "") ===
+                                    fileParam,
```
kjarosh: Why do we have this duality here?

danielhjacobs: I wasn't sure if we'd want ?file=bitey1.swf, this lets it work. That was the real path to the SWF 5 years ago, but now it's swfs/bitey1.swf

kjarosh: Can't we just remove the swfs/ prefix and make it work the same way as before?

danielhjacobs: As discussed on Discord, this will no longer support the old URLs without `swfs/`.

kjarosh (approved):

PR #22021, opened 2025-10-24
web: Better detection for unsupported WASM and improve error instruction

danielhjacobs:


danielhjacobs:
Essentially superseded by #22022, as now you'd have to manually install the single WASM extension to encounter this issue.

PR #22022, opened 2025-10-24
ci: Restore dual-wasm to the web extensions

danielhjacobs:
Support for WASM extensions such as SIMD relies on hardware as well as browser version. The alternative to this is #22021.

Note, this increases the size of the browser extensions submitted to the web store from **5.49 MB to 10.6 MB**.

Also note, this allows the build-browser-extensions and build-web-demo-and-selfhosted jobs to be recombined; that refactor is the second commit.

### DIFF
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -229,7 +229,7 @@ jobs:
 
   build-mac-universal-binary:
     name: Build macOS universal binary
-    needs: [create-nightly-release, build, build-browser-extensions]
+    needs: [create-nightly-release, build, build-web]
     runs-on: macos-15
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release_nightly.yml
```diff
@@ -351,11 +351,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-  # The first couple steps of this and the `build-web-demo-and-selfhosted` job
-  # (tool setup and dependency install) are identical. Hopefully in the future
-  # https://github.com/actions/runner/issues/1182 can help deduplicate them.
-  build-browser-extensions:
-    name: Build browser extensions
+  build-browser-extensions-web-demo-and-selfhosted:
```
kjarosh: Maybe `build-web` considering it encompasses all web builds? + We can elaborate in a comment

PR #22025, opened 2025-10-24
avm1: Change `TDisplayObject::object1` to return `Option<Object>`

moulins:
Reboot of (the last part of) PR #19522 (see PR #21611 for the other part).

- `TDisplayObject::object1` now returns `Option<Object>` (instead of `Value`);
- Adds a `TDisplayObject::object1_or_undef` convenience method that maintains the old behavior;

I did some code simplifications where it was clear they preserved existing behavior, but I've left most of the murkier cases as-is. (e.g. we may be doing a lot of possibly-useless and/or incorrect `.object1_or_undef().coerce_to_object()`s)

### DIFF
--- a/core/src/avm1/globals/movie_clip.rs
+++ b/core/src/avm1/globals/movie_clip.rs
@@ -809,10 +809,7 @@ fn attach_movie<'gc>(
         };
         new_clip.post_instantiation(activation.context, init_object, Instantiator::Avm1, true);
 
-        Ok(new_clip
-            .object1_or_undef()
-            .coerce_to_object(activation)
-            .into())
+        Ok(new_clip.object1_or_undef())
     } else {
         avm_warn!(activation, "Unable to attach '{}'", export_name);
         Ok(Value::Undefined)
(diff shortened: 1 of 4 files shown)


core/src/avm1/globals/movie_clip.rs
```diff
@@ -931,12 +931,11 @@ fn duplicate_movie_clip<'gc>(
 
     let new_clip = clone_sprite(movie_clip, activation.context, name, depth, init_object);
 
-    // On SWF<6 undefined is returned.
```
Lord-McSweeney, line 934: This block is a little hard to visually parse now, could you separate the SWF version check into an early-abort like how it was before?

core/src/context.rs
```diff
@@ -400,45 +398,31 @@ impl<'gc> UpdateContext<'gc> {
         root.set_depth(0);
         root.set_perspective_projection(None); // Set default PerspectiveProjection
 
-        let flashvars = if !self.root_swf.parameters().is_empty() {
-            let object = Avm1Object::new(&self.strings, None);
+        root.post_instantiation(self, None, Instantiator::Movie, false);
+        root.set_default_root_name(self);
+
+        // Set flashvars and the version parameters (AVM1 only)
```
Lord-McSweeney, line 404: Previously post-instantiation was done after flashvars were set, but with this change it's the other way around. Is this observable at all?

moulins: It shouldn't be observable, because the root is always instantiated as a bare `MovieClip` (which has an empty constructor) so no actual AVM1 code runs before the flashvars are set.

PR #22048, opened 2025-10-26
avm1: Avoid storing owned data in `Activation`

moulins:
Keep local registers and activation names as bare references in `Activation`, leaving the caller responsible for allocating them if necessary.

As many activations have static names and/or no local registers (and as such don't need to allocate), this improves performance (up to a 5% reduction in wall-time in AVM1 benchmarks).

Possible further improvements (which would require unsafe code, probably):
- Use a preallocated stack for local registers (like in AVM2);
- Also use references instead of `Gc`s for the scope linked-list; this would require some way of lazily allocating scopes when they get captured by `DefineFunction` closures.

### DIFF
--- a/core/src/avm1/activation.rs
+++ b/core/src/avm1/activation.rs
@@ -20,7 +20,6 @@ use gc_arena::{Gc, Mutation};
 use indexmap::IndexMap;
 use rand::Rng;
 use ruffle_macros::istr;
-use std::borrow::Cow;
 use std::cell::Cell;
 use std::cmp::min;
 use std::fmt;
@@ -63,7 +62,7 @@ enum FrameControl<'gc> {
 #[derive(Clone)]
 pub struct ActivationIdentifier<'a> {
     parent: Option<&'a ActivationIdentifier<'a>>,
-    name: Cow<'static, str>,
+    name: &'a str,
     depth: u16,
     function_count: u16,
     special_count: u8,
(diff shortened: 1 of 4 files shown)


core/src/avm1/activation.rs
```diff
@@ -2430,16 +2422,12 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
-                // Old FP versions do not fall back to the global register set.
-                self.context.player_version <= 10
-            }
+        if let Some(reg) = self.local_registers.get(id as usize) {
+            reg.set(value);
+            true
         } else {
-            false
+            // Old FP versions do not fall back to the global register set.
+            !self.local_registers.is_empty() && self.context.player_version <= 10
```
Lord-McSweeney, line 2430: This code no longer differentiates between an `Activation` with an empty register set and one with no provided registers, is this observable by AVM1 code?

core/src/avm1/activation.rs
```diff
@@ -162,16 +160,8 @@ pub struct Activation<'a, 'gc: 'a> {
 
     /// Local registers, if any.
     ///
-    /// None indicates a function executing out of the global register set.
-    /// Some indicates the existence of local registers, even if none exist.
-    /// i.e. None(Vec::new()) means no registers should exist at all.
```
kjarosh, line 167: Aren't we losing this representation? Was there a reason for it?

kjarosh: Just noticed Lord-McSweeney asked the same thing

core/src/avm1/activation.rs
moulins: "`Activation`s with an empty register set" were never a thing, this is most clearly visible in the version of the code before this PR: see the (now removed) `allocate_local_registers` method.

PR #22050, opened 2025-10-26
tests: Import actionscript.all tests from gnash

Dinnerbone:
Once again, way too many tests to detail - sadly.

All these tests surround a specific AS2 class, and the behaviour often changes per FP version - in these cases the output is in e.g. `fp10` subfolder.

All tests are ran on Windows using FP 9 -> 32. A `SharedObject` test was skipped as we don't have support for this in out test framework yet.
 

Summary is:
```
     Summary [  18.666s] 378 tests run: 50 passed, 328 failed
        FAIL [   0.112s] tests::tests from_gnash/actionscript.all/ASnative-v5
        FAIL [   0.134s] tests::tests from_gnash/actionscript.all/ASnative-v6
        FAIL [   0.099s] tests::tests from_gnash/actionscript.all/ASnative-v7
        FAIL [   0.144s] tests::tests from_gnash/actionscript.all/ASnative-v8
        FAIL [   0.071s] tests::tests from_gnash/actionscript.all/Accessibility-v5
        FAIL [   0.122s] tests::tests from_gnash/actionscript.all/AsBroadcaster-v5
        FAIL [   0.075s] tests::tests from_gnash/actionscript.all/AsBroadcaster-v6
        FAIL [   0.104s] tests::tests from_gnash/actionscript.all/AsBroadcaster-v7
        FAIL [   0.071s] tests::tests from_gnash/actionscript.all/AsBroadcaster-v8
        FAIL [   0.134s] tests::tests from_gnash/actionscript.all/BitmapData-v5
        FAIL [   0.081s] tests::tests from_gnash/actionscript.all/BitmapData-v6
        FAIL [   0.071s] tests::tests from_gnash/actionscript.all/BitmapData-v7
        FAIL [   0.346s] tests::tests from_gnash/actionscript.all/BitmapData-v8
        FAIL [   0.113s] tests::tests from_gnash/actionscript.all/Boolean-v5
        FAIL [   0.062s] tests::tests from_gnash/actionscript.all/Camera-v5
        FAIL [   0.108s] tests::tests from_gnash/actionscript.all/Camera-v6
        FAIL [   0.089s] tests::tests from_gnash/actionscript.all/Camera-v7
        FAIL [   0.098s] tests::tests from_gnash/actionscript.all/Camera-v8
        FAIL [   0.113s] tests::tests from_gnash/actionscript.all/Color-v5
        FAIL [   0.073s] tests::tests from_gnash/actionscript.all/ColorTransform-v5
[...]

PR #22065, opened 2025-10-27
avm1: Add super basic 'ASnative' implementation

Dinnerbone:
It's enough to make `Color-v5` from #22050 pass, at least


tests/tests/swfs/avm1/asnative/output.txt
```diff
@@ -0,0 +1,30 @@
...
+[type Function]
+
+// ASnative("onehundred", 0)
+undefined
+
+// ASnative(100, {valueOf: function() {return 4;}})("success")
+success
+
+// ASnative(100, 4.5)("success")
+success
```
moulins, line 23: Could you also add tests for `ASnative(100, 0x1_0000 + 4)` and `ASnative(100, 0x1_0000_0000 + 4)`, to check overflow behavior?

Dinnerbone: Done

Dinnerbone:
Color-v5 gnash passes, amended commit to remove its known_failure

PR #22086, opened 2025-10-30
avm2/tests: Begin adding test coverage for signatures of classes

Lord-McSweeney:
This PR adds some tests that test the output of `describeType` on builtin classes. So far we test all versions of the `flash.accessibility` classes (and pass), test the `flash.display` classes in SWFv30 (and fail), and test the `flash.display3D` classes in SWFv12, SWFv13, and SWFv30 (and pass).

I'll add tests for more builtin packages and fix more signatures in later PRs.


Lord-McSweeney:
Drafting until #22115 is merged

kjarosh:
Can you use the recently-added subtests feature?

Lord-McSweeney:
The subtests feature doesn't support different SWF versions, only different Flash Player versions. I would rather not add the ability to manipulate SWFs to the subtests feature because it makes comparing output with FP much more complicated.

kjarosh:
Oh, sorry, idk why I assumed those are fp versions, nevermind

PR #22093, opened 2025-10-30
tests: Add basic support for multi-configuration `test.toml`s

moulins:
Reproducing the newly-added section in `tests/README.md`:

## test.toml

&lt;snip&gt;

```toml
# Support for multiple configurations (see the 'Multiple tests' section for more details)
# Fields specified here will override the 'default' values provided in the rest of the document.
[configs.CONFIG_NAME]
# ...
```

## Multiple tests

Sometimes, you may want to test a given `test.swf` several times with different settings.
For this situation, a `[configs]` section can be added for each distinct configuration to be tested;
each configuration will be ran as a separate test.

For example, if `test.swf` has different output depending on the Flash Player version, the following
`test.toml` could be used:
```toml
num_ticks = 1
# other common settings...

[configs.fp9]
output_path = "output.fp9.txt"
player_options.version = 9

[configs.fp10]
output_path = "output.fp10.txt"
player_options.version = 10
```

-------------------------

I've migrated some of the FP-version-specific tests in the gnash test-suite as a proof of concept.

~~For now, only a small number of fields are supported in `[configs]` sections, but adding new ones as needed should be relatively straightforwards.~~ UPDATE: Now all fields (except arrays) are supported.


tests/framework/src/options.rs
```diff
@@ -34,9 +39,19 @@ pub struct TestOptions {
     pub default_fonts: DefaultFontsOptions,
 }
 
+// Per-config patches for `TestOptions`.
+#[derive(Clone, Default, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+struct TestOptionsConfig {
```
kjarosh: Maybe I'm missing something, but can't we just accept `TestOptions` for each config and just add a method to merge two `TestOptions` into one?

moulins: Well, I don't believe that every field should be overrideable; for example:
- a test.swf either requires a renderer or it doesn't, so overriding `required_features` doesn't make much sense;
- I have no idea of what the correct merge semantics would be for the various font-related fields, and I don't want to have to figure it out without having concrete use cases in mind.

One advantage of having two separate structs is that we can start with supporting a minimum set of functionality and expand it as concrete use-cases appear; if we ever end up with `TestOptionsConfig` having most of the `TestOptions` fields with the exact same types, we'll reconsider it then.

kjarosh: > a test.swf either requires a renderer or it doesn't, so overriding `required_features` doesn't make much sense;

I disagree with that, an example of such test that can both require and not require a renderer at the same time would be a Pixel Bender test for testing the same kernel on CPU and GPU.

> I have no idea of what the correct merge semantics would be for the various font-related fields, and I don't want to have to figure it out without having concrete use cases in mind.

I would just go with basic overriding. I wouldn't try to add advanced logic to that. Just make fields optional with default values, override them, and validate separately for each config. This way you don't have to (1) implement advanced logic, and (2) as a user know how it works.

> One advantage of having two separate structs is that we can start with supporting a minimum set of functionality and expand it as concrete use-cases appear; if we ever end up with `TestOptionsConfig` having most of the `TestOptions` fields with the exact same types, we'll reconsider it then.

I'll again disagree with that—implementing it the way I mentioned above covers all possible cases, is generic, and is simple. You don't have to maintain separate structs, you don't have to maintain separate behaviors, you don't have to duplicate code, you don't have to know what is supported in `[configs]` and how it works.

IMO it's better to support something that we don't see a reason to support (for now), and keep the logic simple and generic. If you really insist on properties not being overridable, IMO it's better to just add a check method and panic instead of creating completely separate structs that have (semantically) the same fields.

Currently skimming through those options, I see a potential reason to override >50% of them. For those that do not make sense, I would say—whatever, the logic is generic and if someone will need to use them, they will, if not, not a big deal.

moulins:
After some discussion on Discord, I've changed the merging logic to blindly work at the TOML level instead of taking into account the actual Rust `struct`s.

Pros:
- automatically works for all fields;
- has consistent behavior, no matter the merged field;

Cons:
- no support for arrays (which do not have 'obvious' merge semantics like tables);
- no awareness of Rust types, so logically indivisible fields (like `ViewportDimensions`) may end up split across multiple configs.

tests/fs-tests-runner/src/lib.rs
```diff
@@ -60,7 +60,7 @@ pub struct TestLoaderParams<'a> {
     pub test_name: &'a str,
 }
 
-pub type TestLoader = Box<dyn Fn(TestLoaderParams) -> Option<Trial>>;
+pub type TestLoader = Box<dyn Fn(TestLoaderParams, &mut dyn FnMut(Trial))>;
```
kjarosh, line 63: It would be nicer to return an iterator of `Trial` instead of operating on mutable state.

tests/README.md
```diff
@@ -179,6 +179,32 @@ typewriter = ["Test Font", "Test Font Fallback"]
 japanese_gothic = ["Test Font", "Test Font Fallback"]
 japanese_gothic_mono = ["Test Font", "Test Font Fallback"]
 japanese_mincho = ["Test Font", "Test Font Fallback"]
+
+# Support for multiple configurations (see the 'Multiple tests' section for more details)
+# Fields specified here will override the 'default' values provided in the rest of the document.
+[configs.CONFIG_NAME]
```
kjarosh: Nit: what do you think about naming it `overrides`/`subtests` instead of `configs`? Configs is very generic and it's not obvious we're talking about TestOptions/test.toml here.

tests/fs-tests-runner/src/lib.rs
moulins: Returning an iterator would require extra boxing; it would also be slightly more verbose at the use-site.

tests/README.md
moulins: Would it be enough to change "rest of the document" to "rest of the `test.toml`?"

kjarosh: My point is that it's not obvious from the test reader's POV. If you look at any `test.toml`, it's not obvious what `[configs._]` is, it can be any config. But if it was named e.g. `subtests` or `tests.overrides`, it would be more clear that we're defining multiple tests.

tests/fs-tests-runner/src/lib.rs
kjarosh: > Returning an iterator would require extra boxing

I don't think you have to worry about performance here, it's done once per test directory, and the whole IO is _orders_ of magnitude slower than boxing.

> it would also be slightly more verbose at the use-site

True, but it wouldn't require the refactor at call-site, and not relying on mutable state when it's not needed & not returning stuff through parameters is in general a better design.

It's not a big deal though

kjarosh: It actually required a refactor not only at the call-site but in the whole call chain from what I see, I really don't see any upside by getting rid of boxing through this whole refactor :thinking:

kjarosh (approved):

tests/README.md
moulins: Addressed by renaming `configs` to `subtests`.

tests/README.md
```diff
@@ -179,6 +179,32 @@ typewriter = ["Test Font", "Test Font Fallback"]
 japanese_gothic = ["Test Font", "Test Font Fallback"]
 japanese_gothic_mono = ["Test Font", "Test Font Fallback"]
 japanese_mincho = ["Test Font", "Test Font Fallback"]
+
+# Support for multiple configurations (see the 'Multiple tests' section for more details)
+# Fields specified here will override the 'default' values provided in the rest of the document.
+[subtests.CONFIG_NAME]
```
kjarosh: Nit: it should be subtest name

kjarosh (approved):

tests/README.md
moulins: Ahhh, you're totally right :D

Dinnerbone (approved):
Hurrah!

PR #22096, opened 2025-10-30
avm1: Stub TextSnapshot methods

kjarosh:
They are stubbed according to what `argstest` reports.


kjarosh:
Added some bonus tests

PR #22106, opened 2025-10-31
avm1: Introduce ParametersExt, use it for Array, AsBroadcaster, ASNative

Dinnerbone:
I was going to go more but almost every single function I look at is handling parameters wrongly, so this is becoming a big task to test them and make sure they're correct.

This fixes many small bugs with `Array` and `AsBroadcaster`. Who knows if content ever actually hit it :D

I've filled out all of the likely "get_X" methods on `ParametersExt`, so that I can split up work into multiple PRs later rather than big batches, without conflicting with eachother.

### DIFF
--- a/core/src/avm1/parameters.rs
+++ b/core/src/avm1/parameters.rs
@@ -216,7 +216,6 @@ pub trait ParametersExt<'gc> {
 
     /// Gets the value at the given index as an u32.
     /// The value will be coerced to an u32, even if it's undefined/missing.
-    #[expect(dead_code)]
     fn get_u32(
         &self,
         activation: &mut Activation<'_, 'gc>,
--- a/core/src/avm1/globals/asnative.rs
+++ b/core/src/avm1/globals/asnative.rs
@@ -1,6 +1,7 @@
 use crate::avm1::activation::Activation;
 use crate::avm1::error::Error;
 use crate::avm1::function::{FunctionObject, NativeFunction};
+use crate::avm1::parameters::ParametersExt;
 use crate::avm1::{Object, Value};
 
 pub fn asnative<'gc>(
@@ -11,12 +12,8 @@ pub fn asnative<'gc>(
     if args.len() != 2 {
         return Ok(Value::Undefined);
     }
-    let Ok(category) = args[0].coerce_to_u32(activation) else {
-        return Ok(Value::Undefined);
-    };
-    let Ok(index) = args[1].coerce_to_u32(activation) else {
-        return Ok(Value::Undefined);
-    };
+    let category = args.get_u32(activation, 0)?;
+    let index = args.get_u32(activation, 1)?;
 
     let category_lookup: Option<fn(u32) -> Option<NativeFunction>> = match category {
         100 => Some(crate::avm1::globals::get_native_function),


core/src/avm1/parameters.rs
```diff
@@ -0,0 +1,301 @@
+use crate::avm1::{Activation, Error, Object, Value};
+use crate::string::AvmString;
+use std::cmp::PartialEq;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum UndefinedBehaviour {
```
kjarosh: I think our code base uses en-US and we should stick to it for consistency.

Also, it sounds very scary :fearful:

Dinnerbone: Renamed. It is halloween, scary is on brand!

(Not sure of a better name; it's the behaviour of `undefined` :sweat_smile:)

core/src/avm1/parameters.rs
```diff
@@ -0,0 +1,301 @@
+use crate::avm1::{Activation, Error, Object, Value};
+use crate::string::AvmString;
+use std::cmp::PartialEq;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
Lord-McSweeney, line 5: Can you add a comment explaining that this enum controls the behavior when `undefined` is passed as the argument? As it is the name is a bit ambiguous.

core/src/avm1/globals/as_broadcaster.rs
```diff
@@ -117,11 +132,13 @@ fn broadcast_message<'gc>(
     this: Object<'gc>,
     args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
-    if let Some(event_name_value) = args.get(0) {
-        let event_name = event_name_value.coerce_to_string(activation)?;
+    if let Some(event_name) =
+        args.try_get_string(activation, 0, UndefinedBehavior::TreatAsPresent)?
```
Lord-McSweeney: Could you split this line up into two lines?

core/src/avm1/globals/array.rs
```diff
@@ -356,19 +355,14 @@ pub fn slice<'gc>(
...
+    let start = make_index_absolute(args.get_i32(activation, 0)?, length);
 
-    let end = args.get(1).unwrap_or(&Value::Undefined);
-    let end = if end == &Value::Undefined {
-        length
-    } else {
-        make_index_absolute(end.coerce_to_i32(activation)?, length)
-    };
+    let end =
+        if let Some(index) = args.try_get_i32(activation, 1, UndefinedBehavior::TreatAsMissing)? {
```
Lord-McSweeney: Here too

core/src/avm1/parameters.rs
Dinnerbone: Done

core/src/avm1/globals/as_broadcaster.rs
Dinnerbone: Done

core/src/avm1/globals/array.rs
Dinnerbone: Done

kjarosh:
I think `tests/tests/swfs/avm1/as_broadcaster_undef/.output.txt.kate-swp` is included by mistake?

Dinnerbone:
> I think `tests/tests/swfs/avm1/as_broadcaster_undef/.output.txt.kate-swp` is included by mistake?

It absolutely is! Thanks!

core/src/avm1/globals/array.rs
```diff
@@ -380,21 +374,27 @@ pub fn splice<'gc>(
...
-        if delete_count < 0 {
+    let start = make_index_absolute(start, length);
+
+    let delete_count = if args.len() > 1 {
+        if let Some(delete_count) =
+            args.try_get_i32(activation, 1, UndefinedBehavior::TreatAsMissing)?
+        {
+            delete_count.min(length - start)
+        } else {
             return Ok(Value::Undefined);
```
kjarosh, line 388: Maybe I'm missing something, but this branch cannot be hit? It's enough to do `if let Some(delete_count) =` without `if args.len() > 1 ` right?

Dinnerbone: `UndefinedBehavior::TreatAsMissing` makes an explicit `undefined` return `None`, which hits this branch. `length` will still be `> 1`.

This is the new `splice = original.splice(2, undefined)` case in the test - that will hit this line.

kjarosh: Oh, I see!

core/src/avm1/parameters.rs
moulins: Suggestion (shorter, and less scary :smiley:)
```rs
pub enum UndefinedAs {
  Present,
  Missing,
}
```

Dinnerbone: Hmm I like that, but maybe then I'll go Some or None too

Dinnerbone: Changed

Dinnerbone:
(that's 3 reviewers... any approvals? :D)

PR #22107, opened 2025-10-31
avm1: Improve property declarations of Sound

kjarosh:
This patch orders them properly, adds missing methods, and adds a TODO for properties that probably shouldn't be properties.


moulins (approved):

core/src/avm1/globals/sound.rs
```diff
@@ -206,23 +206,27 @@ impl<'gc> Sound<'gc> {
...
-    "start" => method(start; DONT_ENUM | DONT_DELETE | READ_ONLY);
     "stop" => method(stop; DONT_ENUM | DONT_DELETE | READ_ONLY);
+    "attachSound" => method(attach_sound; DONT_ENUM | DONT_DELETE | READ_ONLY);
+    "start" => method(start; DONT_ENUM | DONT_DELETE | READ_ONLY);
+    "getDuration" => method(duration; DONT_ENUM | DONT_DELETE | READ_ONLY | VERSION_6);
+    "setDuration" => method(set_duration; DONT_ENUM | DONT_DELETE | READ_ONLY | VERSION_6);
+    "getPosition" => method(get_position; DONT_ENUM | DONT_DELETE | READ_ONLY | VERSION_6);
+    "setPosition" => method(set_position; DONT_ENUM | DONT_DELETE | READ_ONLY | VERSION_6);
+    "loadSound" => method(load_sound; DONT_ENUM | DONT_DELETE | READ_ONLY | VERSION_6);
+    "getBytesLoaded" => method(get_bytes_loaded; DONT_ENUM | DONT_DELETE | READ_ONLY | VERSION_6);
```
Toad06, line 223: Does this imply the swf version check in the function is no longer relevant and could be removed?
```rust
fn get_bytes_loaded<'gc>(
    _this: Object<'gc>,
    _args: &[Value<'gc>],
    if activation.swf_version() >= 6 {
        avm1_stub!(activation, "Sound", "getBytesLoaded");
        Ok(1.into())
    } else {
        Ok(Value::Undefined)
    }
}
```
(same for `getBytesTotal`)

kjarosh: I think that's right! I'll remove it

kjarosh: Done

PR #22120, opened 2025-11-02
avm1: Clean up use of BitmapData in BitmapData functions

Dinnerbone:
(yay net negative code diff, even with more cases being handled!)

This is split off from my current "test every argument" stuff, which is escalating too much. No tests for the new error codes, they'll come later with the big tests. Make code less indented, make code more correct. Make gooder.

(easier to review with "hide whitespace")

### DIFF
--- a/core/src/avm1/globals/bitmap_data.rs
+++ b/core/src/avm1/globals/bitmap_data.rs
@@ -1401,6 +1401,7 @@ fn compare<'gc>(
     this: Object<'gc>,
     args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
+    // Note that these error values are specific to this function, there's no standard between functions
     const EQUIVALENT: i32 = 0;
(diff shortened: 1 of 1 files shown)


core/src/avm1/globals/bitmap_data.rs
```diff
@@ -1483,29 +1407,16 @@ fn compare<'gc>(
     const DIFFERENT_WIDTHS: i32 = -3;
     const DIFFERENT_HEIGHTS: i32 = -4;
```
Lord-McSweeney, line 1410: Could you add a comment here noting that these return values are different from the error codes that are used for most of the rest of the methods?

Dinnerbone: I was actually considering just removing these lines, I feel like they're not so helpful :sweat_smile: Every function has its own "-2 means X".

I added a comment for now though

PR #22121, opened 2025-11-02
tests: Deduplicate tests with multiple FP versions

moulins:
...by using `[subtests]` instead of separate subfolders.

---------

Also fixes a mistake in `from_gnash/misc-ming.all/PrototypeEventListeners/`


Dinnerbone:
Well, every time I try to open the github diff it crashes my browser. If this is a clever way to sneak in malware on the assumption that I trust you & am too lazy to open it in git... that's correct! Well done! :D

PR #22122, opened 2025-11-02
avm1: Implement Sound.id3

kjarosh:
This patch implements `Sound.id3` along with `onID3` handler.

The built-in property definition of `id3` is removed, as the property is added dynamically after loading the sound in Flash.

Tests are included.


hamzaelbarmaki1994:
some humour  sound
image

PR #22128, opened 2025-11-03
avm1: Add 'proper' support for `ASnative`-style methods

moulins:
- Add a new `Executable::TableNative` variant that directly stores the `ASnative` method ID instead of requiring a small Rust shim;
- Rewrite the `declare_properties!` macro to allow declaring `TableNative` methods;
- Convert classes which already used IDs internally to the new syntax (and also `Object` and part of `_global`);
- Make AVM1 `ASnative` method directly create `TableNative` function objects, and plug the compatible classes into it.
  - methods of: `Object`, `Date`, `Transform`, all `BitmapFilter` subclasses;
  - `escape`, `unescape`, `parseInt`, `parseFloat`, `trace` (these already worked).

### DIFF
--- a/core/src/avm1/globals.rs
+++ b/core/src/avm1/globals.rs
@@ -1,6 +1,5 @@
 use crate::avm1::activation::Activation;
 use crate::avm1::error::Error;
-use crate::avm1::function::NativeFunction;
 use crate::avm1::property::Attribute;
 use crate::avm1::property_decl::{DeclContext, Declaration};
 use crate::avm1::{Object, Value};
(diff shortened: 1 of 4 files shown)


core/src/avm1/globals.rs
```diff
@@ -74,39 +73,61 @@ mod xml_node;
...
-    "setInterval" => method(set_interval; DONT_ENUM);
-    "clearTimeout" => method(clear_timeout; DONT_ENUM);
-    "setTimeout" => method(set_timeout; DONT_ENUM);
-    "updateAfterEvent" => method(update_after_event; DONT_ENUM);
-    "escape" => method(escape; DONT_ENUM);
-    "unescape" => method(unescape; DONT_ENUM);
+    // 'true' builtins; what are their ordering?
     "NaN" => property(get_nan; DONT_ENUM);
     "Infinity" => property(get_infinity; DONT_ENUM);
+    "ASnative" => method(asnative::asnative; DONT_ENUM); // TODO: does it have an index?
```
Dinnerbone: Seems not - at least not within a reasonable range?

Tested with:
```actionscript
var c = 0;
onEnterFrame = function(){
	for (var i = 0; i < 10000; i++) {
		try {
			var f = ASnative(c, i);
			var r = f(100, 4); // try using the method to fetch `trace()`
			r("ASnative(" + c + ", " + i + ")"); // if it's actually `trace()`, then this will print the `ASnative` ID
		}
		catch(e) {
			trace(e)
		}
	}
	if ((c % 1000) == 0) {
		trace("tried up to " + c);
	}
	c++;
}
```

moulins: I ran your code locally just to be sure, for `ASnative(0..10000, 0..10000)`, and added a comment.

PR #22136, opened 2025-11-03
avm2: Merge default_values and property_classes fields of VTable

Lord-McSweeney:
We now completely disallow holes in the vtable slots list and slot id conflicts, matching avmplus. To implement this, VTable creation is now fallible, which will make it easier to move traits verification to the `init_vtable` in the future.

From some microbenchmarks, this makes constructing objects with lots of slots slightly faster. The main value-copying loop in `ScriptObject::custom_new` goes from
```
b0:   mov    -0x18(%r13,%rdx,1),%rdi 
      mov    -0x10(%r13,%rdx,1),%r8  
      cmp    $0x33,%rdi              
      cmove  %rsi,%rdi               
      mov    %rdi,-0x18(%r15,%rdx,1) 
      mov    %r8,-0x10(%r15,%rdx,1)  
      mov    -0x8(%r13,%rdx,1),%rdi  
      mov    0x0(%r13,%rdx,1),%r8    
      cmp    $0x33,%rdi              
      cmove  %rsi,%rdi               
      mov    %rdi,-0x8(%r15,%rdx,1)  
      mov    %r8,(%r15,%rdx,1)       
      add    $0x2,%rax               
      add    $0x20,%rdx              
      cmp    %rax,%rcx               
    ↑ jne    b0                      
```
to
```
b0:   movups (%rsi),%xmm0      
      movups %xmm0,-0x10(%rdx) 
      movups 0x28(%rsi),%xmm0  
      movups %xmm0,(%rdx)      
      add    $0x2,%rax         
      add    $0x50,%rsi        
      add    $0x20,%rdx        
      cmp    %rax,%rcx         
    ↑ jne    b0                
```


SuchAFuriousDeath (approved):

core/src/avm2/class.rs
```diff
@@ -1086,6 +1086,10 @@ impl<'gc> Class<'gc> {
             .expect("Interfaces not yet initialized!")
     }
 
+    pub fn translation_unit(self) -> Option<TranslationUnit<'gc>> {
```
adrian17, line 1089: any chance we could store the pointer on the class itself?
IIRC in the past there were other times we were kinda interested in "source of a type" (like for security features) and having it not rely on optional initializer sounds better.

But if it's too complex for now, it could be a followup.

core/src/avm2/vtable.rs
```diff
@@ -315,6 +316,21 @@ impl<'gc> VTable<'gc> {
             }
         }
 
+        if let Some(defining_tunit) = defining_class_def.translation_unit() {
+            if let Some(super_class) = defining_class_def.super_class() {
```
adrian17: this should iterate all superclasses. See avmplus `Traits::allowEarlyBinding`.

core/src/avm2/class.rs
Lord-McSweeney: I'll add this in a followup- it's not a big change, but I'd like to keep it separate from this PR

kjarosh:
Can you add tests for the fallible cases you mentioned?

PR #22138, opened 2025-11-04
core: Use __jcvt() intrinsic on AArch64 for f64→i32 conversion

linkmauve:
In #21780, an optimisation has been added to use the `fjcvtzs` ARMv8.3 instruction when available, to convert a `f64` into an `i32`.  This made me wonder why `core::arch::aarch64` didn’t have an intrinsic for this instruction, so I implemented it in [stdarch](https://github.com/rust-lang/stdarch/pull/1938), which got pulled in Rust [yesterday](https://github.com/rust-lang/rust/pull/148402) (see [the tracking issue](https://github.com/rust-lang/rust/issues/147555)).

This PR makes use of this new intrinsic to remove the unsafe `asm!()` block, and simplify the code.

### DIFF
--- a/core/src/ecma_conversions.rs
+++ b/core/src/ecma_conversions.rs
@@ -44,8 +44,10 @@ pub fn f64_to_wrapping_i32(n: f64) -> i32 {
     #[cfg(target_arch = "aarch64")]
     {
         if std::arch::is_aarch64_feature_detected!("jsconv") {
-            // SAFETY: `jsconv` feature is checked in both compile time and runtime to be existed, so it's safe to call.
-            unsafe { f64_to_wrapping_int32_aarch64(n) }
+            // Converts an `f64` to an `i32` with ECMAScript `ToInt32` wrapping behavior.
+            // The value will be wrapped in the range [-2^31, 2^31).
+            // Optimized for aarch64 cpu with the fjcvtzs instruction.
+            unsafe { std::arch::aarch64::__jcvt(n) }
         } else {
             f64_to_wrapping_i32_generic(n)
         }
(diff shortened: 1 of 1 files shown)


kjarosh:
Thanks for taking care of it! Ruffle uses stable Rust, so we have to wait until this feature is stabilized.

linkmauve:
I’ll try to move forward with the stabilization, in the meantime should I close this PR and reopen it when the intrinsic got stabilized, or should I keep it open as a draft?  I originally thought it would be better for stabilization if projects using it were actually using the intrinsic.

kjarosh:
It's fine to leave it as a draft.

folkertdev:
There are a bunch of red CI actions here, so just checking: the intrinsic works as expected? If so then I'd say we can start the stabilization process in ~ 2 months probably.

kjarosh:
In order to check whether the intrinsic works as expected someone with the right hardware would have to run it locally, or GitHub actions would have to be changed (in this PR) to use nightly instead of stable. Currently they are failing because they use stable where this feature is not available.

folkertdev:
Well yes either of those would do. I think that with that practical validation, we could move forward with stabilizing the intrinsic.

folkertdev:
Well, the arm and macos jobs succeeded, I guess that means it works?

linkmauve:
With only nightly enabled, the tests now pass on both [ubuntu-24.04-arm](https://github.com/ruffle-rs/ruffle/actions/runs/20954335927/job/60215164900?pr=22138) and [macos-15](https://github.com/ruffle-rs/ruffle/actions/runs/20954335927/job/60215164926?pr=22138)!

CrazyboyQCD:
This intrinsic [is stabilized in 1.95](https://github.com/rust-lang/rust/blob/0204aca0663ec534b773f78639b9ebe20b3da001/library/stdarch/crates/core_arch/src/aarch64/neon/generated.rs#L52-L62) and [it's out](https://blog.rust-lang.org/2026/04/16/Rust-1.95.0/).

linkmauve:
There we go, CI passes now that 1.95.0 is released!

core/src/ecma_conversions.rs
kjarosh, line 43: This looks like it could be removed?

core/src/ecma_conversions.rs
```diff
@@ -44,8 +44,10 @@ pub fn f64_to_wrapping_i32(n: f64) -> i32 {
     #[cfg(target_arch = "aarch64")]
     {
         if std::arch::is_aarch64_feature_detected!("jsconv") {
-            // SAFETY: `jsconv` feature is checked in both compile time and runtime to be existed, so it's safe to call.
```
kjarosh, line 47: The safety comment is missing now.

torokati44:
Merged as https://github.com/ruffle-rs/ruffle/pull/23937. Thank you!

PR #22142, opened 2025-11-05
avm1: Implement `ASnative(2, 0)`, a.k.a. `ASnew`.

moulins:
`ASnew` is an undocumented (and unnameable) internal method only accessible through `ASnative`. According to the [FlashCoders wiki](https://web.archive.org/web/20121014112705/http://osflash.org/flashcoders/undocumented/asnative), it seems it should be available as `_global.ASnew` on Flash Player 5, but we don't have a good way of having FP-specific properties so I didn't implement it.

This progresses the `from_gnash/actionscript.all/ASnative` tests.


core/src/avm1/globals/asnative.rs
```diff
@@ -48,3 +49,18 @@ pub fn asnative<'gc>(
 
     Ok(Value::Undefined)
 }
+
+/// Undocumented internal `ASnew` function: returns a boolean indicating whether or not the
+/// enclosing function has been called as a constructor. Note that this 'sees through' any
+/// native calls; see the `avm1/asnew` test for examples.
```
Lord-McSweeney, line 55: Can you add a TODO that this should be available in FP5?

moulins: Done!

core/src/avm1/activation.rs
```diff
@@ -217,6 +221,12 @@ impl<'a, 'gc> Activation<'a, 'gc> {
         self.global_scope().locals_cell()
     }
 
+    /// Was this activation created by a constructor call? Note that native calls don't
+    /// create activations, and so aren't taken into account for this check.
+    pub fn in_avm1_constructor(&self) -> bool {
```
Dinnerbone: `avm1` is a bit redundant here

moulins: I wanted to emphasize the fact that it doesn't care about native calls, so maybe `in_bytecode_constructor` instead?

Edit: this is because I'm considering adding a separate tracking mechanism for *native* constructors (to replace our separate `constructor` fn pointer, for the few builtins that behave differently as a constructor than as a function, e.g.`new Date()` vs `Date()`), which would then be called `activation.in_native_constructor()`.

Dinnerbone: Yeah that sounds much clearer!

PR #22163, opened 2025-11-07
chore: Move more `ruffle_core` code to new `ruffle_common` crate

Lord-McSweeney:
This moves the files `utils.rs`, `sandbox.rs`, `buffer.rs`, `tag_utils.rs`, and `xml.rs` to `ruffle_common`. These files don't contain any hot code that might not be inlined after this change, so this is probably fine performance-wise. Moving `SwfMovie` to `ruffle_common` is necessary for splitting AVM2 into its own crate.


moulins:
I'm not entirely convinced by the `tag_utils` move: I don't like how it makes `ruffle_common` depend on `ruffle_render` and `ttf_parser` just to be able to include their error types as `tag_utils::Error` variants. Couldn't we move only `SwfMovie` into `ruffle_common` and leave the rest of the `tag_utils` logic in `ruffle_core`?

PR #22167, opened 2025-11-07
core: Add custom version strings

### LINKED ISSUE #22148: Ruffle does not pass version checks
hello. I have a few swfs that do not function on ruffle because of their flash version checks.

one checks for flash player version 50 (this is in use by our AIR environment and is not an invalid version) 
if you believed that sitelocks and version locks were a thing of the past, many of our new swfs have both. boo. happy halloween. 
(we're happy with being able to spoof where necessary)

another few use and check for invalid flash player versions (37, 43, and 49) (these, I believe, were versions of flash player that never existed; they 'exist' in our 'cursed' environment, at the time of their creation a version 50 did not exist. as you can imagine, maintaining these 'versions' for our use never was and no longer is, viable.
hacked players and swfs with matching checks)
and lastly, another checks for flash player version 11.2.202.548. this one checks for the 2.202.... after the 11.

the ruffle version check modifier on the ruffle desktop program is not sufficient for bypassing these checks.

- the version number can only be set from 1 - 32 (this does not cover our invalid checks, and does not cover our player 50 swfs)
- the version number cannot be set to incremental versions beyond an integer (the 11 swf's check was to ensure only that version of 11 would be used)

I unfortunately cannot provide any of these swfs nor the players, we are under contract to not share, modify, or tamper with these files in any way that isn't 'updates' (they decide these)
we are aware of flash decompilers, we cannot modify the swfs.
all of these files have worked on windows 10 flash player.
we would continue to use flash player and we have an AIR contract, but our provider has decided to move to using ubuntu 24.04 LTS after some hysterics and frustrations from the higher ups about windows 11.

and they have have found that ruffle works more consistently across Operating systems than any flash player for ubuntu. This has worked very well so far, we have nearly 1400 of our swf 'systems' carried over, just not these.

we have tried to ask for versions of the swf files without version checks, or at least versions with more sensible checks - we have been ignored.
despite us absolutely needing a solution, we're out of options

we are stuck and we'd like to ask if you could:

- allow arbitrary version numbers (at least every number until 50)
- allow version numbers that indicate versioning

I apologise if some of this isn't coherent, this project has always been a legacy behemoth that we barely understand...
we would like to rewrite now or to have had rewritten before 2020, but that never happened and doesn't seem like it ever will. we never believed that flash player would truly have support revoked/ zombified like it has been and so it was never done.

there is no rush, we have been wading in both pools for some time now during the stall.
we apologise for any previous unprofessionality, we are a revolving door.

with exhaustion,
thank you

Dinnerbone:
This overrides `$version`/`Capabilities.version` to a custom string, if set.

Available through:
- Desktop CLI as `--custom-player-version <CUSTOM_PLAYER_VERSION>`
- Desktop UI
- Bundles as `custom_version_string`
- Web load options as `customPlayerVersionString`

This also changes AVM1 `$version`/`System.capabilities.version` to use the same logic as AVM2, where it will report `WIN` instead of `LNX` by default.

This should fix #22148

### DIFF
--- a/frontend-utils/src/bundle/README.md
+++ b/frontend-utils/src/bundle/README.md
@@ -1,34 +1,49 @@
 # Ruffle Bundle (.ruf) format specification
+
 A Ruffle Bundle is an easy way to package and share Flash games and any assets that are required to make the game work.
 
 A bundle can be a directory or a renamed zip file, and must contain at minimum a `ruffle-bundle.toml` file.
 
 <!-- TOC -->
+
 * [Ruffle Bundle (.ruf) format specification](#ruffle-bundle-ruf-format-specification)
-  * [Directory structure](#directory-structure)
-    * [`ruffle-bundle.toml` (Bundle information)](#ruffle-bundletoml-bundle-information)
-    * [`content/` (Flash content)](#content-flash-content)
-  * [`ruffle-bundle.toml` file specification](#ruffle-bundletoml-file-specification)
-    * [`[bundle]`](#bundle)
-      * [`name` - The name of the bundle](#name---the-name-of-the-bundle)
-      * [`url` - The url of the Flash content to open](#url---the-url-of-the-flash-content-to-open)
-    * [`[player]`](#player)
-      * [`[parameters]` - A list of parameters to pass to the starting movie](#parameters---a-list-of-parameters-to-pass-to-the-starting-movie)
-      * [`script_timeout` - Script execution timeout](#script_timeout---script-execution-timeout)
-      * [`base_url` - Base URL for relative file paths](#base_url---base-url-for-relative-file-paths)
-      * [`quality` - Quality that the content starts with](#quality---quality-that-the-content-starts-with)
-      * [`align` - Stage Alignment that the content starts with](#align---stage-alignment-that-the-content-starts-with)
-      * [`force_align` - Allow or disallow the content from changing its Stage Alignment](#force_align---allow-or-disallow-the-content-from-changing-its-stage-alignment)
-      * [`scale_mode` - Stage Scale Mode that the content starts with](#scale_mode---stage-scale-mode-that-the-content-starts-with)
-      * [`force_scale_mode` - Allow or disallow the content from changing its Stage Alignment](#force_scale_mode---allow-or-disallow-the-content-from-changing-its-stage-alignment)
-      * [`upgrade_http_to_https` - Whether to upgrade HTTP urls to HTTPS silently](#upgrade_http_to_https---whether-to-upgrade-http-urls-to-https-silently)
-      * [`load_behavior` - How Ruffle should load movies](#load_behavior---how-ruffle-should-load-movies)
-      * [`letterbox` - Controls visual letterboxing around the content](#letterbox---controls-visual-letterboxing-around-the-content)
-      * [`spoof_url` - URL to pretend the initial SWF is being loaded from](#spoof_url---url-to-pretend-the-initial-swf-is-being-loaded-from)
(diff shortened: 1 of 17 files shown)


frontend-utils/src/bundle/README.md
```diff
@@ -1,34 +1,47 @@
...
+            * [
+              `force_scale_mode` - Allow or disallow the content from changing its Stage Alignment](#force_scale_mode---allow-or-disallow-the-content-from-changing-its-stage-alignment)
+            * [
+              `upgrade_http_to_https` - Whether to upgrade HTTP urls to HTTPS silently](#upgrade_http_to_https---whether-to-upgrade-http-urls-to-https-silently)
+            * [`load_behavior` - How Ruffle should load movies](#load_behavior---how-ruffle-should-load-movies)
+            * [
+              `letterbox` - Controls visual letterboxing around the content](#letterbox---controls-visual-letterboxing-around-the-content)
+            * [
+              `spoof_url` - URL to pretend the initial SWF is being loaded from](#spoof_url---url-to-pretend-the-initial-swf-is-being-loaded-from)
+            * [`version` - Version of the Flash Player to emulate](#version---version-of-the-flash-player-to-emulate)
```
danielhjacobs, line 38: `custom_version_string` is missing from this part of the README

Dinnerbone: Thanks!

frontend-utils/src/bundle/README.md
```diff
@@ -1,34 +1,49 @@
...
+    * [Directory structure](#directory-structure)
+        * [`ruffle-bundle.toml` (Bundle information)](#ruffle-bundletoml-bundle-information)
+        * [`content/` (Flash content)](#content-flash-content)
+    * [`ruffle-bundle.toml` file specification](#ruffle-bundletoml-file-specification)
+        * [`[bundle]`](#bundle)
+            * [`name` - The name of the bundle](#name---the-name-of-the-bundle)
+            * [`url` - The url of the Flash content to open](#url---the-url-of-the-flash-content-to-open)
+        * [`[player]`](#player)
+            * [
+              `[parameters]` - A list of parameters to pass to the starting movie](#parameters---a-list-of-parameters-to-pass-to-the-starting-movie)
```
kjarosh, line 19: Should this be divided into two lines? Similarly lines below

moulins:
It seems somewhat unfortunate to have slightly different names between CLI/web/etc: couldn't CLI use `--custom-player-version-string` and bundles `custom_player_version_string` instead? This will probably be a rarely-used option so I wouldn't mind sacrificing conciseness here.

Dinnerbone:
I omitted "player" from the bundle because it's under `player`. "player's custom player version string" just feels too redundant.

Cli I have no string feelings on, it just felt too long.

kjarosh:
I'm still wondering if we should allow producing behaviors different than FP's. I personally liked @Lord-McSweeney's idea of (optionally) passing a 4-part version. We could also use this information to infer the u8 player version so that the user doesn't have to sync those two. Plus we make sure that the content doesn't break unexpectedly and we don't produce unexpected behaviors that are not present in FP.

Lord-McSweeney:
If we had a four-part player version, we could also observe API versioning that changed between, e.g. FP 10.0.0.32 and FP 10.1 without having to reparse the entire string.

Could we always store the player version as four u8s and make `--player-version` set it to `X.0.0.0` and `--custom-player-version `(`--player-version-full` maybe?) set it to `W.X.Y.Z`?

kjarosh:
We could also accept `a,b,c,d` as a variant of `--player-version`

PR #22170, opened 2025-11-08
core/avm1/avm2: Minor fixes to DO lookup, naming, and construction

Lord-McSweeney:
- When enumerating AVM1 `MovieClip` properties, we should be enumerating all `DisplayObject`s with a name, not just interactive `DisplayObject`s
- All accessible `DisplayObject` variants need to get their default instance name set in AVM2
- We need to disallow access to `Bitmap`, `Graphic`, `MorphShape`, and `Text` in AVM1, not just `Graphic`
- `StaticText` should be constructed at `construct_frame` time, not `post_instantiation` time

This includes one test for AVM1 and one test for AVM2.

This progresses `from_gnash/misc-ming.all/DefineTextTest`


core/src/avm1/activation.rs
```diff
@@ -2655,11 +2655,12 @@ impl<'a, 'gc> Activation<'a, 'gc> {
...
                     {
-                        // If an object doesn't have an object representation, e.g. Graphic, then trying to access it
-                        // Returns the parent instead
+                        // If an object doesn't have an object representation,
+                        // i.e. Graphic or Bitmap, then trying to access it
+                        // returns the parent instead
                         if path_has_slash {
                             child.object1_or_undef()
-                        } else if let crate::display_object::DisplayObject::Graphic(_) = child {
+                        } else if child.is_avm1_inaccessible() {
```
moulins: Isn't that mostly equivalent to `child.object1().is_none()`? The latter seems better as it'll fail gracefully if we somehow end up with an AVM2 movie in a AVM1 context.

core/src/avm1/object/stage_object.rs
```diff
@@ -39,9 +39,9 @@ pub fn get_property<'gc>(
     {
         return if is_slash_path {
             Some(child.object1_or_undef())
-        // If an object doesn't have an object representation, e.g. Graphic, then trying to access it
-        // Returns the parent instead
-        } else if let crate::display_object::DisplayObject::Graphic(_) = child {
+        // If an object doesn't have an object representation, e.g. Graphic,
+        // then trying to access it returns the parent instead
+        } else if child.is_avm1_inaccessible() {
```
moulins: Same as above.

core/src/avm1/activation.rs
Lord-McSweeney: Is there no way for AVM1 to view a not-yet-constructed object?

moulins: I'm don't see how this could happen, as AVM1 objects are initialized in `DisplayObject::post_instantiation`. (My understanding is that failing to call `post_instantiation` after instantiating an object is a bug, but maybe I'm wrong...)

moulins (approved):

PR #22178, opened 2025-11-08
Revert "web: Note Ruffle does not collect user data (Firefox)"

danielhjacobs:
Reverts ruffle-rs/ruffle#21983

https://addons.mozilla.org/en-US/firefox/addon/ruffle_rs/reviews/2533947/

### DIFF
--- a/web/packages/extension/webpack.config.js
+++ b/web/packages/extension/webpack.config.js
@@ -47,9 +47,6 @@ function transformManifest(content, env) {
         manifest.browser_specific_settings = {
             gecko: {
                 id: firefoxExtensionId,
-                data_collection_permissions: {
-                    required: ["none"],
-                },
             },
         };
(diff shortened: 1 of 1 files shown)


danielhjacobs:
Worth waiting for the uplift to ESR 115 of https://bugzilla.mozilla.org/show_bug.cgi?id=1996829

moulins:
The upstream bug has been closed as fixed, so I don't think this is necessary anymore?

danielhjacobs:
Yeah, this shouldn't be needed. Unfortunately the fix was merged right after the previous ESR minor release (https://www.firefox.com/en-US/firefox/115.30.0/releasenotes/), so this will probably not be in a release for another month, since the minor release before that was a month ago (https://www.firefox.com/en-US/firefox/115.29.0/releasenotes/).

PR #22179, opened 2025-11-08
desktop: Use `async-task` crate to manage `Player`-created futures

moulins:
Instead of manually (and unsafely!) fiddling with `Waker` vtables and implementing our own task "allocator" (a.k.a. the `SlotMap` in `AsyncExecutor`), rip everything out and use heap-allocated `PlayerRunnable = async_task::Runnable`s instead.

[`async-task`](https://crates.io/crates/async-task) takes care of all the nitty-gritty and provides an API we can directly plug into the `winit` event loop.

Notes:
- This removes the internal queue of to-be-polled tasks; instead, `PlayerRunnable`s get pushed as individual winit events directly on the main event loop.
- To prevent polling tasks belonging to stale players, we tag each task with a `PlayerId` uniquely identifying each `Player` instance, and only poll the task if its ID matches the active `Player`.


torokati44:
roger that, over

moulins:
Hm, if Android needs to adapt the code, maybe it would still be useful to keep a small abstraction around `async_task`? I guess the relevant shared part is only ~60 LoC, so it's fine for now :shrug:

torokati44:
> Hm, if Android needs to adapt the code, maybe it would still be useful to keep a small abstraction around `async_task`?

I'm not convinced...

> I guess the relevant shared part is only ~60 LoC, so it's fine for now 🤷

I'll probably manage to update...

PR #22183, opened 2025-11-09
chore: Migrate crates to Rust Edition 2024 (part 3/N)

moulins:
(part 2: #22166)

This migrates the `desktop` and `frontend-utils` crates.

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -580,18 +580,18 @@ impl ApplicationHandler<RuffleEvent> for App {
             }
 
             (Some(main_window), RuffleEvent::EnterFullScreen) => {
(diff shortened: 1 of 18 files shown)


frontend-utils/Cargo.toml
```diff
@@ -11,7 +11,7 @@ version.workspace = true
 workspace = true
 
 [features]
-cpal = ["dep:cpal", "dep:bytemuck"]
```
Dinnerbone, line 14: This took me a moment to understand, but it's a sneaky fix smuggled inside an edition upgrade :scream:

moulins: Oops, I probably should've split this into a separate commit, sorry ^^

PR #22196, opened 2025-11-10
chore: Migrate crates to Rust Edition 2024 (part 5/N)

moulins:
(part 4: #22186)

Migrates the `web/*` and `tests/*` crates; only `core` remains in edition 2021 now.


moulins:
Switching `web/packages/extension/safari` to edition 2024 causes `objc`'s `class!` macro to fire this warning:
```
warning: unexpected `cfg` condition value: `cargo-clippy`
 --> web/packages/extension/safari/src/main.rs:8:63
  |
8 |         let mut ruffle = ClassDecl::new("RuffleWebExtension", class!(NSObject))
  |                                                               ^^^^^^^^^^^^^^^^
  |
  = note: no expected values for `feature`
  = note: using a cfg inside a macro will use the cfgs from the destination crate and not the ones from the defining crate
  = help: try referring to `class` crate for guidance on how handle this unexpected cfg
  = help: the macro `class` may come from an old version of the `objc` crate, try updating your dependency with `cargo update -p objc`
  = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
  = note: `#[warn(unexpected_cfgs)]` on by default
  = note: this warning originates in the macro `class` (in Nightly builds, run with -Z macro-backtrace for more info)
  ```

`objc` is unmaintained, so let's switch to the better-maintained [`objc2`](https://github.com/madsmtm/objc2) instead.

EDIT: and now there's another failure... :c
EDIT 2: CI passes now, hopefully the safari extension builds properly... (I don't have an iPad/iPhone/iMac to test with)

Issue #22174, opened 2025-11-08
Error on https://uploads.ungrounded.net/ruffle_wrapper/ruffleembed.html


waspennator:
# Error Info
Error name: Error
Error message: panicked at render/src/utils.rs:370:40:
Invalid PNG width: TryFromIntError(())
Error stack:
```
Error: panicked at render/src/utils.rs:370:40:
Invalid PNG width: TryFromIntError(())
    at Ve.e.wbg.__wbg_new_476169e6d59f23ae (https://uploads.ungrounded.net/ruffle_wrapper/lib/core.ruffle.7e9610e9d26f4c0320d2.js:1:56964)
    at ruffle_web.wasm.__wbg_new_476169e6d59f23ae externref shim (wasm://wasm/ruffle_web.wasm-031a5ada:wasm-function[11892]:0x93d6bc)
    at ruffle_web.wasm.ruffle_web::global_init::{{closure}}::h67891f78a9137e6d (wasm://wasm/ruffle_web.wasm-031a5ada:wasm-function[10067]:0x915ebb)
    at ruffle_web.wasm.std::panicking::panic_with_hook::h5f62ad9dc4d6300a (wasm://wasm/ruffle_web.wasm-031a5ada:wasm-function[6753]:0x85ed2f)
    at ruffle_web.wasm.core::panicking::panic_fmt::ha095fedb92b6396f (wasm://wasm/ruffle_web.wasm-031a5ada:wasm-function[10979]:0x9303f1)
    at ruffle_web.wasm.core::result::unwrap_failed::h90b13c69eec28497 (wasm://wasm/ruffle_web.wasm-031a5ada:wasm-function[8090]:0x8aeb91)
    at ruffle_web.wasm.ruffle_render::utils::decode_define_bits_jpeg_dimensions::h0097c8a916d00a9b (wasm://wasm/ruffle_web.wasm-031a5ada:wasm-function[2081]:0x50d91a)
    at ruffle_web.wasm.ruffle_core::display_object::movie_clip::MovieClip::preload::hd8b98d3e471dcc8f (wasm://wasm/ruffle_web.wasm-031a5ada:wasm-function[1190]:0x2cfc26)
    at ruffle_web.wasm.ruffle_core::player::Player::preload::h2ca9c5591d55fd3d (wasm://wasm/ruffle_web.wasm-031a5ada:wasm-function[1507]:0x3dcdc8)
[...]

PR #22197, opened 2025-11-10
core: Support bitmaps larger than 65535x65535

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/22174

`DefineBitsJPEG2` can define bitmaps larger than 65535x65535, FP supports it, and some games use it for e.g. defining sprites on a very large bitmap.

This patch migrates bitmap width/height from `u16` to `u32`, as that's the maximum theoretical size of a PNG image. It also overall fits better with the current codebase and requires less conversions.


n0samu:
I wonder how such huge bitmaps will look on devices with a max texture size of 4096... 😅

PR #22198, opened 2025-11-10
avm2: Fix class signatures for all classes in `flash.display` package

Lord-McSweeney:
This also implements a lot of version-gating. Most of the diff is from new test coverage.


core/src/avm2/globals/flash/display/display_object.rs
```diff
@@ -541,14 +540,21 @@ pub fn set_rotation<'gc>(
...
     let this = this.as_object().unwrap();
 
     if let Some(dobj) = this.as_display_object() {
-        return Ok(dobj.name().unwrap_or_else(|| istr!("")).into());
+        if let Some(name) = dobj.name() {
+            return Ok(name.into());
+        } else {
+            // The `Stage` is the only DisplayObject that can be accessed in
+            // AVM2 that has a `None` name
+            assert!(matches!(dobj, DisplayObject::Stage(_)));
```
kjarosh, line 555: Can we change it to a debug assert at least? Or add a path for stage? When it happens in production we shouldn't crash Ruffle; it's not a situation where we cannot continue playing SWFs

core/src/avm2/globals/flash/display/LoaderInfo.as
```diff
@@ -27,5 +32,36 @@ package flash.display {
...
+        override public function dispatchEvent(event:Event):Boolean {
+            throw new IllegalOperationError("Error #2118: The LoaderInfo class does not implement this method.", 2118);
+        }
+
+        public static function getLoaderInfoByDefinition(object:Object):LoaderInfo {
+            // Docs say that this returns `null` when debugging isn't enabled,
+            // and (TODO) in FP it throws a SecurityError unless called in a
+            // local SWF that has been marked as "trusted", so do we really need
+            // to implement it?
+            return null;
```
kjarosh, line 49: I'd include a stub warning, when somehow this breaks some SWF, it will be easier to debug.

Issue #16745, opened 2024-06-16
Text stretched on boxcar2d


Lord-McSweeney:
### Describe the bug

The text at the bottom showing time, score, and torque, and the text at the top showing the generation and max score are both stretched horizontally.

### Expected behavior

The text is not stretched.

### Content Location

http://boxcar2d.com

[boxcar2d.v3.2.swf.zip](https://github.com/user-attachments/files/15859122/boxcar2d.v3.2.swf.zip)

### Affected platform

Desktop app

### Operating system

Ubuntu 20.04

### Additional information

This occurs both when running the SWF in the desktop app and when running it on web.

Lord-McSweeney:
I can still reproduce this issue

Issue #10365, opened 2023-03-25
subtitle size


manotroll:
### Describe the bug

[TV digital.zip](https://github.com/ruffle-rs/ruffle/files/11068858/TV.digital.zip)
flash player executable 17
[flashplayer_17_sa.zip](https://github.com/ruffle-rs/ruffle/files/11068863/flashplayer_17_sa.zip)

### Expected behavior

o texto em baixo esta maior que deveria 
the text is bigger than it should be
https://i.imgur.com/Ll1NJnH.png

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

-----

n0samu:
This is another case of #1881.

Issue #4948, opened 2021-07-29
Rendering one font incorrectly


curriouscat:
I cannot see that this is addressed.

I have an issue with ruffle's rendering of one font that makes the numbers unreadable.
I'm using the selfhosted version on Chrome/Safari/Firefox running on a Mac recent OS

I've attached a pic of the image at fault 

![image](https://user-images.githubusercontent.com/11879289/127485537-4d05dbd0-28ef-49d4-bc9a-35d0175bcd78.png)

The numbers in the blue box are the important numbers from the charts and cannot be read. They are being rendered too big.

This is what they should look like

![image](https://user-images.githubusercontent.com/11879289/127486119-5257e6df-6c7a-43d5-9112-a3e894f5a1f2.png)

I've attached a zipped copy of the swf questSampleSize and another that has the same problem sampleSize

[questSampleSize.swf.zip](https://github.com/ruffle-rs/ruffle/files/6900222/questSampleSize.swf.zip)
[sampleSize.swf.zip](https://github.com/ruffle-rs/ruffle/files/6900223/sampleSize.swf.zip)

You can see them online here https://netgm.com/netgm_tools/tools/useful_stuff/questionnaire_sample_size.html
and here
https://www.bellsite.id.au/netgm.com/netgm_tools/audit_sample_size.php

Any help would be greatly appreciated. I wrote the code for these about 2 decades ago and the maths was a nightmare. Inverting extremely long and difficult functions. I could not do it again now.
GeoffBell

danielhjacobs:
At a cursory glance, this seems like #1881, but I haven't actually looked at the source code to see if it's the same issue.

PR #22209, opened 2025-11-11
text: Fix x/y scaling of device text glyphs

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/1881
* Fixes https://github.com/ruffle-rs/ruffle/issues/4948
* Fixes https://github.com/ruffle-rs/ruffle/issues/1724
* Fixes https://github.com/ruffle-rs/ruffle/issues/10365
* Fixes https://github.com/ruffle-rs/ruffle/issues/1408
* Fixes https://github.com/ruffle-rs/ruffle/issues/16745
* Progresses https://github.com/ruffle-rs/ruffle/issues/2380
* Progresses https://github.com/ruffle-rs/ruffle/issues/22157
* Progresses (probably) https://github.com/ruffle-rs/ruffle/issues/15110

Device text in FP cannot be scaled independently in x/y. Only the y scale is applied, and the x scale is set so that the aspect ratio of characters is preserved. It's used only for calculating metrics and sizing bounds.

This PR contains a bunch of patches that bring us closer to FP's behavior. We're still not perfect, especially related to measurements, but the good news is that it's a mess in Flash Player too, just a different one.

The main issue with dependent x/y scaling is that measurements operate in unscaled coordinate space, meaning that we have to emulate the size of the unscaled glyphs. If `scaleX/scaleY == 0.5`, we have to act as if glyphs were twice as wide.

After this PR Ruffle still behaves differently, especially related to:
1. y metrics (which are "broken" in FP, but they work "properly" in Ruffle),
2. laying out text (which is broken in Ruffle, but works properly in FP),
3. negative scales (I don't even know where they are broken anymore),
4. small sizes (less than gutters, i.e. <4px, but no one uses that realistically).


Lord-McSweeney:
This also fixes #16745!

Lord-McSweeney:
With this, #22113 displays this:
image

The text is no longer stretched, but it's not in the right place anymore.

hamzaelbarmaki1994:
@kjarosh @Lord-McSweeney IMHO, i don't think is due to this PR because is already happens in this following example where text is getting behind the input field : https://mep-col.sesamath.net/dev/swf/exo657.swf?idMep=91&modeleMep=1&abreviationLangue=fr&idSwf=91&ch=n&idAide=79&isBoutonSuite=o&isBoutonAide=o
its better to test it with the extension player to be able to next questions with red buttons 
image
due to wrong HTML ALIGN value (here is "RIGHT") it must be "LEFT" as you can see in the image 
image
maybe this PR is the Solution ?  https://github.com/ruffle-rs/ruffle/pull/22428  (edit confirmed it solves it)

Lord-McSweeney (approved):

kjarosh:
Works properly for me:

image

Could have been a different issue that has been fixed.

hamzaelbarmaki1994:
@kjarosh maybe the stretching is gone but still the position an issue in position in this case : https://bibliotheque.sesamath.net/public/voir/584

before PR:
image
after PR:
image
edit: Ah, the alignment is not respected, here the value of alignment is center
image
image

kjarosh:
@hamzaelbarmaki1994 Can you file an issue for it? There's few remaining things I need to straighten out with the scaled device text

hamzaelbarmaki1994:
@kjarosh done : https://github.com/ruffle-rs/ruffle/issues/22945

PR #22221, opened 2025-11-12
avm1: Reduce usage of `coerce_to_object`

moulins:
They can often be replaced by more specific constructs.

Except for the last commit (which fixes an edge-case in the `__resolve` logic), these are all non-observable changes.


core/src/avm1/activation.rs
```diff
@@ -571,6 +567,15 @@ impl<'a, 'gc> Activation<'a, 'gc> {
         self.context.avm1.push(value);
     }
 
+    fn pop_call_args(&mut self, num_args: usize) -> Vec<Value<'gc>> {
+        (0..num_args.min(self.context.avm1.stack_len()))
+            .map(|_| {
+                let arg = self.context.avm1.pop();
```
Lord-McSweeney, line 573: It's not immediately obvious what this mapping does- can you add a comment to explain?

core/src/avm1/object.rs
```diff
@@ -452,8 +452,10 @@ pub fn find_resolve_method<'gc>(
             return Err(Error::PrototypeRecursionLimit);
         }
 
-        if let Some(value) = p.get_local_stored(istr!("__resolve"), activation, false) {
-            return Ok(Some(value.coerce_to_object(activation)));
+        let resolve = p.get_local_stored(istr!("__resolve"), activation, false);
+        // FP completely skips over primitives (but not over non-function objects).
+        if let Some(Value::Object(value)) = resolve {
```
Lord-McSweeney, line 457: Is it possible for `resolve` to be a `Value::MovieClip`?

core/src/avm1/activation.rs
moulins: Added a comment.

core/src/avm1/object.rs
moulins: `MovieClipReference`s are ignored; added a test-case.

PR #22226, opened 2025-11-13
web: Add config option to hide message when Ruffle restores from bfcache

danielhjacobs:
This adds a configuration option, `hideRestoredMessage`, which hides the message that appears when Ruffle is restored from the bfcache. This message was added when Toad06 switched from the unload event to the pagehide event, enabling use of the bfcache.

Note that the bfcache is never used with Ruffle on Chromium due to a "bug" (previously intended behavior), therefore this more-so affects Gecko (Firefox). See https://issues.chromium.org/issues/424552566. Basically, Adobe Flash Player would not work with the bfcache, so Chromium disables the bfcache on pages with embeds or objects that have SWF sources. Since modern Chromium no longer supports any plugins, including Flash, this is now unneeded code on Chromium's part.

### DIFF
--- a/web/packages/core/src/public/config/load-options.ts
+++ b/web/packages/core/src/public/config/load-options.ts
@@ -451,6 +451,13 @@ export interface BaseLoadOptions {
     favorFlash?: boolean;
 
     /**
+     * Hide the message when content is restored from the back/forward cache.
+     *
+     * @default false
+     */
+    hideRestoredMessage?: boolean;
+
+    /**
      * This is no longer used and does not affect anything.
      * It is only kept for backwards compatibility.
      *
--- a/web/packages/core/src/public/config/load-options.ts
+++ b/web/packages/core/src/public/config/load-options.ts
@@ -362,6 +362,26 @@ export enum GamepadButton {
 }
 
 /**
+ * The behavior when the bfcache takes effect
+ */
+export enum CacheBehavior {
+    /**
+     * Inform the user that the content was restore from the bfcache.
+     */
+    Inform = "inform",
+
+    /**
+     * Restore the content from the bfcache without a message.
+     */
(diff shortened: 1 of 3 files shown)


kjarosh:
Can we make it an enum that select whether to restore, reload, or ask the user?

danielhjacobs:
In the reload scenario, how would we want that to work? Calling `this.reload()` in `displayRestoredFromBfcacheMessage`?

danielhjacobs:
Added reload, but the content does briefly blink as the page itself has restored from the bfcache.

https://github.com/user-attachments/assets/85e5adea-4304-4a60-823d-eb92d9d19504

danielhjacobs:
The enum is `bfcacheBehavior`: "inform", "restore", and "reload".

web/packages/core/src/public/config/load-options.ts
```diff
@@ -361,6 +361,26 @@ export enum GamepadButton {
     DPadRight = "dpad-right",
 }
 
+/**
+ * The behavior when the bfcache takes effect
+ */
+export enum CacheBehavior {
```
kjarosh: Nit: can we include bfcache in the class name? In case there's a clash in the future with some other cache

kjarosh (approved):
LGTM, thank you! <3

web/packages/core/src/public/config/load-options.ts
danielhjacobs: Done, did `BFCacheBehavior` with that capitalization to be consistent with PascalCase, since b and f are abbreviations for words.

n0samu:
Honestly I would prefer if we removed the message entirely. I find it annoying and not useful. Or we could at least change the default behavior.

PR #22243, opened 2025-11-15
core: Make ruffle_core::font module public

kjarosh:
When adding more features and integrations to fonts (e.g. device fonts), more and more types are being used from the font module.  This patch makes it public so that symbols from it don't have to be reexported manually.


desktop/src/backends/ui.rs
```diff
@@ -14,7 +14,8 @@ use ruffle_core::backend::ui::{
     DialogLoaderError, DialogResultFuture, FileDialogResult, FileFilter, FontDefinition,
     FullscreenError, LanguageIdentifier, MouseCursor, UiBackend,
 };
-use ruffle_core::{FontFileData, FontQuery};
+use ruffle_core::font::FontFileData;
```
Lord-McSweeney: nit: Why not merge the imports
```
use ruffle_core::font::FontFileData;
use ruffle_core::font::FontQuery;
```
into
```
use ruffle_core::font::{FontFileData, FontQuery};
```
everywhere?

Lord-McSweeney (approved):

desktop/src/backends/ui.rs
kjarosh: I usually prefer the former syntax because of conflict resolution: the latter syntax will produce non-trivially-fixable conflicts when two people edit the same file :/

But I see that the second one is commonly used in Ruffle, so I'll change that

PR #22250, opened 2025-11-16
core: Always enable kerning for device fonts

kjarosh:
The kerning setting works only for embedded fonts, and it does not affect device fonts at all. Kerning is always enabled for device fonts.

Flash documentation also mentions that the kerning setting only affects embedded fonts.


tests/tests/swfs/fonts/device_font_kerning/output.expected.png
kjarosh, line 1: These rectangles are glyphs, spaces between them are kerning. Top field has kerning disabled, bottom one has kerning enabled.

PR #22251, opened 2025-11-16
tests: Mark `known_failure` tests with a `!` in the runner's output

moulins:
Useful to see at a glance which action to take when a test fails (`known_failure` tests failing are a good sign!).

Example output:
```
test [!]     from_gnash/misc-swfmill.all/zeroframe_definesprite ... ok
test [!fp11] from_gnash/actionscript.all/argstest-v8            ... ok
```

### DIFF
--- a/tests/tests/regression_tests.rs
+++ b/tests/tests/regression_tests.rs
@@ -104,9 +104,18 @@ fn load_test_dir<'a>(test_dir: &'a VfsPath, name: &'a str) -> impl Iterator<Item
 
 fn trial_for_test(opts: &RuffleTestOpts, test: Test, list_only: bool) -> Trial {
     let ignore = !test.should_run(opts.ignore_known_failures, !list_only, &NativeEnvironment);
-    let subtest_name = test.options.subtest_name.clone();
 
-    let mut trial = Trial::test(test.name.clone(), move || {
+    // Put extra info into the test 'kind' instead of appending it to the test name,
+    // to not break `cargo test some/test -- --exact` and `cargo test -- --list`.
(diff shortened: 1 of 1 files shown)


tests/tests/regression_tests.rs
```diff
@@ -104,9 +104,15 @@ fn load_test_dir<'a>(test_dir: &'a VfsPath, name: &'a str) -> impl Iterator<Item
 
 fn trial_for_test(opts: &RuffleTestOpts, test: Test, list_only: bool) -> Trial {
     let ignore = !test.should_run(opts.ignore_known_failures, !list_only, &NativeEnvironment);
-    let subtest_name = test.options.subtest_name.clone();
+    let mut subtest_name = String::new();
```
kjarosh: I would rename this to `test_kind`, as now it's not only the subtest name (+ we can put here other things in the future too).

tests/tests/regression_tests.rs
```diff
@@ -136,13 +142,10 @@ fn trial_for_test(opts: &RuffleTestOpts, test: Test, list_only: bool) -> Trial {
...
     });
-    if ignore {
-        trial = trial.with_ignored_flag(true);
-    }
-    if let Some(name) = subtest_name {
+
+    trial
+        .with_ignored_flag(ignore)
         // Put the subtest name as the test 'kind' instead of appending it to the test name,
         // to not break `cargo test some/test -- --exact` and `cargo test -- --list`.
```
kjarosh: After renaming `subtest_name` to `test_kind` it makes sense to move it over the line that adds the subtestname to the test kind.

Issue #21893, opened 2025-10-11
Texts missing and unexpected feature in 武斗纹章-时度


fsfsdfcsef:
### Describe the bug

1.The texts of character selection and backpack are not displayed 

<img width="873" height="623" alt="Image" src="https://github.com/user-attachments/assets/8aae9cce-530d-4d9c-8300-5b3ebb7dc8b5" />

(Character selection)

2.Load save that save from a different scene will give you 2 useless lv.0 emblems with no image and can't be selected. These useless emblems shouldn't be appeared in the game.

<img width="786" height="627" alt="Image" src="https://github.com/user-attachments/assets/c57abeae-1650-4594-8d3e-fe83218c6518" />
(Backpack)

### Expected behavior

<img width="701" height="499" alt="Image" src="https://github.com/user-attachments/assets/3576e4cd-60a6-4b58-9ef3-49d4e806d5b5" />

(Character selection)

<img width="700" height="497" alt="Image" src="https://github.com/user-attachments/assets/13fd3040-f50d-46a9-b7ce-d301974ad60c" />

(Backpack)

### Content Location

https://www.mediafire.com/file/s11xco5hvmbebjo/武斗纹章-时度.zip/file

### Affected platform

Desktop app

### Operating system

Windows10 LTSC 2021

### Additional information

Console output on load save (texts in angle brackets is translation for Chinese)

```
2025-11-04T04:17:29.426583Z  INFO ruffle_core::focus_tracker: Focus is now on None
2025-11-04T04:17:29.492998Z  INFO avm_trace: 添加一个纹章数据    <<Add a emblem data>>
2025-11-04T04:17:29.493065Z  INFO avm_trace: 添加一个纹章数据    <<Add a emblem data>>
2025-11-04T04:17:29.493092Z  INFO avm_trace: 存在纹章数 2    <<Aquired emblem amount: 2>>
2025-11-04T04:17:29.493997Z  INFO avm_trace: 点击 [object SavePoint]    <<Click>>
[...]

Issue #19354, opened 2025-01-27
Kerning/text-width issues in display of Japanese text


FioraAeterna:
### Describe the bug

I am trying to restore an old Japanese flash game so that it can be accessible with modern browsers. However, the text rendering is fairly off in a number of ways compared to Adobe Flash, resulting in text often overflowing the textbox, like this:

![Image](https://github.com/user-attachments/assets/9873e88d-6eb2-447e-96a1-a14f9c5430ec)

Furthermore, the kerning is very weird, which makes it a bit painful to read, even ignoring the issue with the text sometimes clipping. By contrast, Flash looks like this:

![Image](https://github.com/user-attachments/assets/1c4c69f6-e2f4-4235-86ff-584c5131dfe0)

It may not be obvious to someone who can't read Japanese, but the text in the Ruffle version is weirdly squished, with some letters horizontally shrank relative to others in a strange way.

### Expected behavior

I expect the text rendering to be similar to Adobe Flash, or at least similar to "typical Japanese text". Ideally, it shouldn't clip over the edge, as well.

### Content Location

https://fioraaeterna.github.io/ar-tonelico.jp/1avs/cosmosphere.html

### Affected platform

Browser's extension

### Operating system

macOS Sonoma 14.7 

### Browser

Chome 131

FioraAeterna:
Worth noting, I found an [English translated version of the same game](https://exapico.freetzi.com/1avs/cosmosphere.html) and it _also_ has problems with text rendering in Ruffle, resulting in cut off text. So there might actually be two issues here; one with the cut-off text rendering, and (separately) the general poor display of Japanese text.

<image height="300" src="https://github.com/user-attachments/assets/9bcf7cb7-5c48-4722-a4cf-b08d8faae249"/>

n0samu:
Yes, here is a comparison of the spacing in the English version (desktop FP vs desktop Ruffle):

**Ruffle**:
<image height="300" src="https://github.com/user-attachments/assets/7fd434de-2c35-4540-892f-00b0d80a47d7"/>

**FP**:
<image height="300" src="https://github.com/user-attachments/assets/5eace245-1d6c-4a9c-998a-dd58b68cfdd7"/>

Note that while the original game specifies the text field's font as `_ゴシック` (Gothic), the English version specifies it as `ＭＳ ゴシック` (MS Gothic)

nebulas-star:
[...]

Issue #19158, opened 2025-01-09
(NSFW) Witch Girl - missing letters and numbers


Noxious45:
### Describe the bug

(This is a multi asset game, so the SWF file alone won't work.)

![Image](https://github.com/user-attachments/assets/110a188e-0007-42ee-add4-b8e49f62c9ec)

On the top center of the screen with the star, there's no number showing the amount of special attacks you have along with the speed of how fast your character is going forward or backward.

This also happens during gameplay with the Witch Girl herself. When attacking, you can see white text boxes above her head, but there's nothing there.

https://github.com/user-attachments/assets/434ccc00-f6ee-4c92-8179-ab27bbbdcf5a

### Expected behavior

All the letters and numbers should be visible 

![Image](https://github.com/user-attachments/assets/69e89680-3f55-46cc-a7af-75f71abab885)

https://github.com/user-attachments/assets/a11b44f2-9baf-4e0c-b56b-2bedca1b14ce

### Content Location

https://www.comdotgame.com/play/witch-girl-english

This isn't the latest version (v2,30), However, the letters and numbers are showing normally

### Affected platform

Desktop app

### Operating system

Windows 11

### Additional information

I can't upload the .zip file due to its size, but this is a multi asset game with its latest version of v2.34.

kjarosh:
The number at the top isn't shown because it requires a Japanese Mincho device font (`_明朝`). Currently Ruffle uses the first one of the following fonts:
* `ヒラギノ明朝 Pro W3`,
* `MS PMincho`,
* `Noto Sans CJK JP`,
* `Arial Unicode MS`.

If any font above is present on the system, the text should be displayed, otherwise you'll get an error: `Known default device font not found: _明朝, ..., text will be missing`.

I suspect the same thing applies to the text above the Witch Girl (but I wasn't able to make it work). There's also `_ゴシック` (Japanese Gothic) used, which will use one of:
* `ヒラギノ角ゴ Pro W3`,
* `MS UI Gothic`,
* `Noto Sans CJK JP`,
* `Arial Unicode MS`.

Ruffle does not yet implement proper fallback device font selection, that's why some text is missing. Using device fonts, we should always fall back to some glyph from some font if available.

fckgithb1123:
hey you're doing better than me, I can't even see the bitch

![Image](https://github.com/user-attachments/assets/37a4b15a-a075-49b8-adf6-5c82c501e024)

Noxious45:
[...]

PR #22254, opened 2025-11-16
web: Render device fonts using offscreen canvas

kjarosh:
* Fixes on web (when enabled) https://github.com/ruffle-rs/ruffle/issues/21893
* Fixes on web (when enabled) https://github.com/ruffle-rs/ruffle/issues/21301
* Fixes on web (when enabled) https://github.com/ruffle-rs/ruffle/issues/19354
* Fixes on web (when enabled) https://github.com/ruffle-rs/ruffle/issues/19158
* Fixes on web (when enabled) https://github.com/ruffle-rs/ruffle/issues/18877
* Fixes on web (when enabled) https://github.com/ruffle-rs/ruffle/issues/18639
* Fixes on web (when enabled) https://github.com/ruffle-rs/ruffle/issues/14724
* Fixes on web (when enabled) https://github.com/ruffle-rs/ruffle/issues/13454
* Fixes on web (when enabled) https://github.com/ruffle-rs/ruffle/issues/12875
* Fixes on web (when enabled) https://github.com/ruffle-rs/ruffle/issues/1616
* Progresses on web (when enabled) https://github.com/ruffle-rs/ruffle/issues/14742
* ... probably many more

Implement device fonts support on web using canvas font renderer. Device fonts are rendered on canvas and then used as a bitmap in Ruffle.

This PR is divided roughly into 2 parts:
1. support bitmap glyphs and glyphs rendered externally in core,
2. implement canvas font renderer in web.

This solves the current issue of missing device fonts on web—Ruffle currently uses the default, embedded Noto Sans font and allows providing custom fonts as device fonts. When the "canvas" font renderer is enabled, glyphs are rendered on an offscreen canvas by the device, and stored in a bitmap.

This is currently experimental, as some things may not work properly or are missing, such as:
1. no hit test implemented yet,
2. each glyph is stored in a separate bitmap instead of being stored in an atlas,
3. each glyph is rendered with 64px height and then scaled down/up as a bitmap instead of being rendered with different resolutions,
4. default fonts are not supported well yet.

This is being put behind a setting `deviceFontRenderer` with value `canvas` (instead of the default `embedded`).

* https://github.com/ruffle-rs/ruffle/issues/10073 doesn't work with this because we're missing proper default font support.


web/src/ui/font_renderer.rs
```diff
@@ -0,0 +1,119 @@
...
+            ctx,
+            ascent,
+            descent,
+        })
+    }
+
+    fn to_font_str(italic: bool, bold: bool, size: f64, font_family: &str) -> String {
+        let italic = if italic { "italic " } else { "" };
+        let bold = if bold { "bold " } else { "" };
+        format!("{italic}{bold}{size}px {font_family}")
```
evilpie, line 63: This might missing quotes around the family?

danielhjacobs:
Is it possible to switch to using the current text rendering if for the browser in use `typeof OffscreenCanvas === "undefined"`, even if the config's `deviceFontRenderer` is "canvas"? Since we say we support Firefox 94 or higher and Safari 14.1 or higher but per https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas#browser_compatibility this is Firefox 105, Safari 16.4.

kjarosh:
I think we can fall back to a regular canvas then.

web/src/ui/font_renderer.rs
kjarosh: Done!

kjarosh:
I think this is ready for review. I added a TODO for the OffscreenCanvas compatibility issue, and I think it's fine to take care of it in follow-up PRs.

core/src/font.rs
```diff
@@ -937,26 +1067,79 @@ impl GlyphShape {
                     && ruffle_render::shape_utils::shape_hit_test(shape, point, local_matrix)
             }
             GlyphShape::Drawing(drawing) => drawing.hit_test(point, local_matrix),
+            GlyphShape::Bitmap(_) => {
+                // TODO Implement this.
+                true
```
Lord-McSweeney, line 1083: Is there a possibility that this will regress existing content, if hitTests now always succeed?

kjarosh: Yes, it will, but you'd have to enable canvas font rendering in the config (it's not default; and it's mentioned that it's experimental and some stuff doesn't work yet).

kjarosh: As a side note: implementing hit test is not super difficult, but I just didn't want to dump all the code in one PR.

kjarosh:
Somehow the kerning fix from https://github.com/ruffle-rs/ruffle/pull/22250 broke tests on firefox and only on the runner (it works locally). I'll be investigating this

kjarosh:
Issue fixed by https://github.com/ruffle-rs/ruffle/pull/22322

Issue #10924, opened 2023-05-03
Sieger - Objects gets "multiplied"


desertking:
### Describe the bug

I guess the game mechanics are broken and not correctly interpreted by ruffle as the game objects seem to become more instead of rotating? Also it's very slow, could be due to the performance problem with ruffle overall or due to the more objects coming in. https://www.newgrounds.com/portal/view/550278?emulate=flash

### Expected behavior

Objects should get destroyed not multiplied 👯 

### Affected platform

Online demo

### Operating system

All

### Browser

All

n0samu:
The only possibly relevant log message I see is:
`WARN run_frame:run_all_phases_avm2: ruffle_core::stub: Encountered stub: AVM2 flash.utils.Dictionary constructor with weak keys`
We should check this again once #10069 is rebased.

desertking:
Still multiplies the objects...

thenickman100:
When the objects are turned into fragmented pieces, Ruffle doesn't seem to be able to crop their images, so it will look like many duplicated objects.  Probably an image processing issue, not an object multiplication issue.  

Comparing the two videos, it looks like the physics is probably behaving correctly despite the graphics glitch.

desertking:
This game would work fine but still not fixxed as of today.

thenickman100:
It seems that now instead of duplication, the objects disappear when hit.

<img width="2056" height="1546" alt="Image" src="https://github.com/user-attachments/assets/24883b1b-3bcf-48d6-9b97-f5bf5946d3ca" />

PR #22266, opened 2025-11-17
core: Allow display objects with drawings as masks

### LINKED ISSUE #21641: Angry Birds Pigs Out – Object duplication after cutting
# Describe the Bug:

In Angry Birds Pigs Out, when cutting a support object, the game duplicates the object instead of removing/cutting it once. This creates multiple overlapping instances that shouldn’t exist.

# Expected behavior:

When you cut the wood/rope/etc., the object is removed once, and physics continue normally.

# Actual behavior:

After cutting, the same object is duplicated at the same time, breaking the gameplay.

Steps to reproduce:

1. Load Angry Birds Pigs Out in Ruffle.

2. Cut the first support beam in Level 1.

3. Observe that the beam (and sometimes pig) is duplicated instead of being cut once.

# Link: https://flashmuseum.org/angry-birds-pigs-out/

# Screenshots / GIFs:

Flash (correct): Image
Ruffle (buggy) (sorry internal audio don't work for unknown reason after i record it): Image

SuchAFuriousDeath:
Fixes #21641. Not 100% sure if this is the correct fix, but it does work. Without this, Ruffle skips applying a mask and that results in the behaviour obseved in the issue.


Lord-McSweeney:

tests/tests/swfs/avm1/mask_with_drawing/test.toml
```diff
@@ -0,0 +1,7 @@
+num_frames = 1
+
+[image_comparisons.output]
+tolerance = 0
+
+[player_options]
+with_renderer = { optional = true, quality = "high" }
```
kjarosh: I think it doesn't make sense to make it optional no? We aren't checking anything besides the visual output.

SuchAFuriousDeath: Done

kjarosh (approved):

SuchAFuriousDeath:
I tried to test #10924 as well btw, but the game is behaving kinda weird even with the fixes, so I'm not sure :sweat_smile:

kjarosh:
> I tried to test #10924 as well btw, but the game is behaving kinda weird even with the fixes, so I'm not sure 😅

Me too, but it seems to be a different issue. Looks like something's wrong with object tracking / hit test.

PR #22284, opened 2025-11-19
chore: Bump tower-http because currently used version got yanked

torokati44:
See: https://crates.io/crates/tower-http/0.6.5

`cargo-deny` has been complaining for a while now, we just don't care about it because of all the duplicates it also reports:
https://github.com/ruffle-rs/ruffle/actions/runs/19447660279/job/55645788740#step:5:6546
(We could allow the known ones in the config and make any remaining/new warnings fail CI...)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5741,9 +5741,9 @@ dependencies = [
 
 [[package]]
 name = "tower-http"
-version = "0.6.5"
+version = "0.6.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5cc2d9e086a412a451384326f521c8123a99a466b329941a9403696bff9b0da2"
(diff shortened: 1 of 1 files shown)


torokati44:
Note that the persisting CI failure is unrelated.

danielhjacobs:
Regarding the persisting CI failure, should we merge #22271 despite it duplicating the console dependency? We will have to bump `indicatif` to address https://rustsec.org/advisories/RUSTSEC-2025-0119, as the old version of indicatif, in addition to depending on `console 0.15.x`, depends on `number_prefix`.

torokati44:
> Regarding the persisting CI failure, should we merge #22271 despite it duplicating the console dependency? 

I'm leaning towards that, yeah...

EDIT: Similarly with wgpu+hashbrown BTW... 😫😩

Issue #22308, opened 2025-11-23
Panicked: local task dropped by a thread that didn't spawn it


kjarosh:
### Describe what you were doing

Opened `http://www.mousecity.com/files/games/VP8.swf` and closed the player while the screen was white

### What does the crash message say?

panicked at /run/build/ruffle/cargo/vendor/async-task-4.7.1/src/runnable.rs:437:17:
local task dropped by a thread that didn't spawn it

### Ruffle Version

0.2.0-nightly.2025.11.21-nightly (61af13e015ca9846bacbb615e071f14ff9a6b665 2011-11-11)

### Affected platform

Desktop app

### Operating system

Linux [64-bit]

kjarosh:
Caused by https://github.com/ruffle-rs/ruffle/pull/22179

CC @moulins

PR #22320, opened 2025-11-24
desktop: Properly shutdown the tokio runtime before exiting the winit event loop.

moulins:
Fixes issue #22308.

### DIFF
--- a/desktop/src/backends/ui.rs
+++ b/desktop/src/backends/ui.rs
@@ -145,10 +145,16 @@ impl FileDialogResult for DesktopFileDialogResult {
 }
 
 pub struct DesktopUiBackend {
+    // It's important that `clipboard`` gets dropped before `window`, dropping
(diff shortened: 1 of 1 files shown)


kjarosh:
I'm getting a segfault when reproducing the original issue with this PR

SuchAFuriousDeath:
Where is the segfault happening in for you? For me, the crash was in smithay-clipboard's background thread (wl_proxy_destroy in libwayland-client.so). passing None to Clipboard::new() in desktop/src/backends/ui.rs fixes this. 
But I am not quite sure how this relates any way to this PR :D. But the segfault does not happen on master, only with this PR.

kjarosh:
Same for me

SuchAFuriousDeath:
I know, I wrote the comment before checking Discord :sweat_smile:

PR #22321, opened 2025-11-24
avm2: Reformat remaining AS3 sources

kjarosh:
This makes AS3 code more consistent.


core/src/avm2/globals/__ruffle__/stubs.as
```diff
@@ -1,12 +1,11 @@
...
 
-    public native function stub_setter(... rest):void;
+    public native function stub_setter(...rest):void;
 
-    public native function stub_constructor(... rest):void;
+    public native function stub_constructor(...rest):void;
 
-
-    public native function log_warn(... rest):void;
+    public native function log_warn(...rest):void;
```
Lord-McSweeney, line 17: nit: this was intentionally separated because it's not a stubbing function. Could you add a comment before `log_warn` to explain (and keep it away from the above functions)?

core/src/avm2/globals/flash/display/Stage3D.as
```diff
@@ -41,7 +41,16 @@ package flash.display {
...
+            if (
+                [
+                    Context3DProfile.BASELINE,
+                    Context3DProfile.BASELINE_CONSTRAINED,
+                    Context3DProfile.BASELINE_EXTENDED,
+                    Context3DProfile.STANDARD,
+                    Context3DProfile.STANDARD_CONSTRAINED,
+                    Context3DProfile.STANDARD_EXTENDED
+                ].indexOf(profile) == -1
+            ) {
```
Lord-McSweeney: nit: This is a bit harder to read now, could you separate out the array into a variable?

Lord-McSweeney (approved):
Thank you!

core/src/avm2/globals/flash/display/Stage3D.as
kjarosh: Done

core/src/avm2/globals/__ruffle__/stubs.as
kjarosh: Done

PR #22323, opened 2025-11-25
desktop: don't panic when Ruffle is closed while an async operation is running

moulins:
Hacky alternative to PR #22320 that doesn't cause a segfault under Wayland.

Fixes issue #22308.


moulins:
Closed in favor of #22320.

PR #22349, opened 2025-11-27
web: Add device font renderer setting to the extension

kjarosh:
This allows users to change the device font renderer and try out the new canvas device font renderer.

### DIFF
--- a/web/packages/extension/assets/options.html
+++ b/web/packages/extension/assets/options.html
@@ -58,6 +58,13 @@
                 </select>
                 <label for="preferred_renderer">Preferred renderer</label>
             </div>
+            <div class="option select">
+                <select id="device_font_renderer">
+                    <option value="embedded" id="device_font_renderer_embedded">Embedded</option>
+                    <option value="canvas">Canvas</option>
+                </select>
+                <label for="device_font_renderer">Device font renderer</label>
+            </div>
             <div class="option number-input">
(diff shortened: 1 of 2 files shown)


web/packages/extension/assets/options.html
```diff
@@ -58,6 +58,13 @@
                 </select>
                 <label for="preferred_renderer">Preferred renderer</label>
             </div>
+            <div class="option select">
+                <select id="device_font_renderer">
+                    <option value="embedded">Embedded</option>
```
danielhjacobs: For i18n, should we add id `id="device_font_renderer_embedded" and translation `settings_device_font_renderer_embedded` (and canvas)?

kjarosh: I was thinking about it, and I wasn't sure if we can translate those values so that it makes sense...

kjarosh: Added translation for Embedded, but I don't think it makes sense to translate "Canvas", as it's a technical term.

danielhjacobs (approved):

hamzaelbarmaki1994:
can this be activated by default ? so the hoster that do automatique update to Ruffle don't need to activate it ?

kjarosh:
> can this be activated by default ? so the hoster that do automatique update to Ruffle don't need to activate it ?

It can't be for now, sadly. It needs a bit more work before we enable it by default.

PR #22355, opened 2025-11-28
avm2: Reduce make_error code duplication using a macro.

evilpie:


evilpie:
Maybe in the future we could also generate the "Error #XXX:" prefix automatically? For the undocumented method `Error.getErrorMessage()` we would also need a way of getting the message by its number.

Lord-McSweeney (approved):

Issue #42, opened 2019-08-23
Text rendering incorrectly in The Final Task


Herschel:
In The Final Task (https://www.newgrounds.com/portal/view/7549), the opening text that says "flash animation" renders incorrectly on both web and destkop.

Ruffle web:
![image](https://user-images.githubusercontent.com/36278/63608061-e0d83280-c588-11e9-9f93-e37fbbaca33a.png)

Flash Player:
![image](https://user-images.githubusercontent.com/36278/63608005-bedeb000-c588-11e9-8789-4bc60a90b20a.png)

PR #22360, opened 2025-11-29
avm2: Stub for Error.getErrorMessage

evilpie:
In Flash unknown error ids produce strings like `Error #42` vs known `Error #3608: Stream %1 is invalid.`.


core/src/avm2/globals/error.rs
```diff
@@ -1,9 +1,22 @@
...
+
+pub fn get_error_message<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    _this: Value<'gc>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    avm2_stub_method!(activation, "Error", "getErrorMessage");
+
+    let id = args.get_i32(0);
+    let message = format!("Error #{id}");
```
Lord-McSweeney, line 17: Does this match the behavior of Flash for negative error ids?

Lord-McSweeney (commented):
Can you add a small test for this?

core/src/avm2/globals/error.rs
evilpie: Surprisingly, yes. Flash also returns negative ids.

evilpie:
> Can you add a small test for this?

When I am back home.

PR #22366, opened 2025-11-29
tests: Make `known_failure` more precise

moulins:
(Requires PR #22348)

Instead of treating any error or panic as a success, only "expected" failures are allowed, and the kind of check that should fail must be specified:
- root-level `known_failure = true`: means that the final trace output check is expected to fail;
- root-level `known_failure.panic = "msg"`: means that the test is expected to panic (like the `#[should_panic = "msg"]` attribute in Rust unit tests);
- `image_comparisons.NAME.known_failure = true`: means that the given image check is expected to fail.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -40,7 +40,8 @@ ignore = false
 # When the test passes in the future, it'll fail and alert that it now passes.
 # This will not catch Ruffle panics; if the test is expected to panic, use
 #   `known_failure.panic = "panic message"`
-# instead.
+# instead (note that 'panicky' tests will be skipped if the test harness is run
+# with debug assertions disabled, e.g. with `--release`).
 known_failure = false
 
 # Path (relative to the directory containing test.toml) to the expected output
--- a/tests/framework/src/test.rs
+++ b/tests/framework/src/test.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 
 use crate::environment::Environment;
 use crate::options::TestOptions;
+use crate::options::known_failure::KnownFailure;
 use crate::runner::TestRunner;
 use crate::util::read_bytes;
 use anyhow::{Result, anyhow};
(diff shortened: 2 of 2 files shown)


Dinnerbone:
Tests in `--release` now fail (I do this a lot, it's faster to test in release :sweat_smile:)

It's because we expect on e.g. overflow panic message, but those don't exist in release.

```
        FAIL [   0.234s] tests::tests [!] avm1/bitmap_data_thorough/pixelDissolve
        FAIL [   0.067s] tests::tests [!] from_gnash/misc-ming.all/get_frame_number_test
        FAIL [   0.075s] tests::tests [!fp10] from_gnash/misc-swfc.all/matrix_accuracy_test1
        FAIL [   0.057s] tests::tests [!fp9] from_gnash/misc-swfc.all/matrix_accuracy_test1
```

moulins:
Added a commit to skip panicky tests if debug assertions are disabled; it's a *little* heavy-handed, as some panicky tests hit regular assertions, but we have few enough of those that I don't think it's a big deal.

tests/tests/swfs/from_shumway/acid/acid-textfield/test.toml
```diff
@@ -1,9 +1,10 @@
 # Test adapted from Shumway at https://github.com/mozilla/shumway/tree/master/test/swfs/acid
 
 num_frames = 1
-known_failure = true # https://github.com/ruffle-rs/ruffle/issues/12123
+known_failure = true # https://github.com/ruffle-rs/ruffle/issues/12118
```
Lord-McSweeney: Was there a reason for this comment to change? #12118 seems to be about `acid-shapes-testing`

tests/framework/src/options/known_failure.rs
```diff
@@ -0,0 +1,52 @@
...
+
+impl<'de> de::Visitor<'de> for KnownFailureVisitor {
+    type Value = KnownFailure;
+
+    fn expecting(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("a boolean, or `.panic = 'message'`")
+    }
+
+    fn visit_bool<E: de::Error>(self, v: bool) -> Result<Self::Value, E> {
+        Ok(match v {
```
Lord-McSweeney: nit: Could this just be an if-statement?

tests/tests/swfs/from_shumway/acid/acid-textfield/test.toml
moulins: That was a typo, good catch!

tests/framework/src/options/known_failure.rs
moulins: Changed.

PR #22369, opened 2025-11-29
avm2: Implement BitmapData.encode

### LINKED ISSUE #17141: BitmapData.encode missing
### Describe the bug

when trying to run this example:
https://www.jacksondunstan.com/articles/2143

I get the error:

Error dispatching event EventObject(EventObject { type: "complete", class: flash.events::Event, ptr: 0xbb8a70 }) to handler FunctionObject(FunctionObject { ptr: 0xb93b40, name: Ok("CompressPerformance/onLoaded()") }) : ReferenceError: Error #1069: **Property encode not found on flash.display.BitmapData** and there is no default value.

### Expected behavior

It should run OK.

### Content Location

http://files.jacksondunstan.com/articles/2143/CompressPerformance.swf

### Affected platform

Self-hosted version

### Operating system

Windows 11

### Browser

Chrome

### Additional information

_No response_

SuchAFuriousDeath:
Fixes #17141

Test upcoming, I'm putting it here just in case my willpower to finish this disappears.


core/src/avm2/globals/flash/display/bitmap_data.rs
```diff
@@ -1596,3 +1600,148 @@ pub fn merge<'gc>(
...
+                .get_public_property(istr!("fastCompression"), activation)?
+                .coerce_to_boolean();
+
+            EncodeOptions::Png { fast_compression }
+        } else {
+            avm2_stub_method!(
+                activation,
+                "flash.display.BitmapData",
+                "encode",
+                "with JPEGXREncoderOptions"
```
torokati44, line 1649: oof, so we're gonna need a JPEG-XR encoder too? 👀
does the current Rust wrapper crate expose that?

SuchAFuriousDeath: No idea, I'll check

SuchAFuriousDeath: > oof, so we're gonna need a JPEG-XR encoder too? 👀

Only if we want to implement it fully, who knows if it ever gets called by anything, BitmapData.encode seems pretty niche on its own.

torokati44: of course, I meant "ideally"

Issue #13946, opened 2023-11-10
Pre-Civilization: Marble Age: Panicked only on web - create_texture: Dimension Y is zero


MartySVK:
# Error Info
Error name: Error
Error message: panicked at /home/runner/.cargo/git/checkouts/wgpu-53e70f8674b08dd4/49d16f7/wgpu/src/backend/direct.rs:3056:5:
wgpu error: Validation Error

Caused by:
    In Device::create_texture
    Dimension Y is zero

Error stack:
```
Error: panicked at /home/runner/.cargo/git/checkouts/wgpu-53e70f8674b08dd4/49d16f7/wgpu/src/backend/direct.rs:3056:5:
wgpu error: Validation Error

Caused by:
    In Device::create_texture
    Dimension Y is zero


    at H.n.wbg.__wbg_new_ab87fd305ed9004b (https://www.zahraj.sk/emulators/ruffle/core.ruffle.d273449361a0cd066e42.js:1:56255)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::h189f598b25fe4514 (wasm://wasm/033ef5ea:wasm-function[9962]:0x9cb822)
    at std::panicking::rust_panic_with_hook::h3aa054d35a0817d7 (wasm://wasm/033ef5ea:wasm-function[6532]:0x8cf542)
    at core::panicking::panic_fmt::h3e1dd3d08288569e (wasm://wasm/033ef5ea:wasm-function[9762]:0x9c6622)
    at core::ops::function::Fn::call::hab4d293b6b61afb0 (wasm://wasm/033ef5ea:wasm-function[6715]:0x8d9da1)
    at wgpu::backend::direct::ErrorSinkRaw::handle_error::ha8d57641bba5ca8c (wasm://wasm/033ef5ea:wasm-function[5416]:0x870389)
    at <T as wgpu::context::DynContext>::device_create_texture::haf3ce0631cd54806 (wasm://wasm/033ef5ea:wasm-function[7273]:0x908fb3)
    at wgpu::Device::create_texture::hc6929010bde77107 (wasm://wasm/033ef5ea:wasm-function[5543]:0x87c73e)
    at <ruffle_render_wgpu::backend::WgpuRenderBackend<T> as ruffle_render::backend::RenderBackend>::register_bitmap::haf7fc2f99939d60c (wasm://wasm/033ef5ea:wasm-function[2038]:0x5aa8e8)
[...]

PR #22373, opened 2025-11-30
avm2: Set EditText width during creation

### LINKED ISSUE #12490: dechmog - can't enter the game
### Describe the bug

game stuck at loading

### Expected behavior

game can finish loading

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

_No response_

SuchAFuriousDeath:
Fixes #12490.

I do not fully understand this tbh and not sure if this is the correct fix.

What happens is that the game tries to cache movie clips using bitmaps and without this fix the bounds of the movieclip are absolutely massive, resulting in a bitmap so big that it is invalid. (The width of the TextEdit is set to the default 1 000 000 from the createTextLine method signature: https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/text/engine/TextBlock.html#createTextLine().

I expect that this will need a test, but I do not want to waste time on this before it is vibe-checked by someone who has a proper understanding of this code area.

### DIFF
--- a/core/src/avm2/globals/flash/text/engine/text_block.rs
+++ b/core/src/avm2/globals/flash/text/engine/text_block.rs
@@ -170,6 +170,7 @@ fn apply_format<'gc>(
 
     let measured_text = display_object.measure_text(activation.context);
 
+    display_object.set_width(activation.context, measured_text.0.to_pixels());
     display_object.set_height(activation.context, measured_text.1.to_pixels());
 
     Ok(())


SuchAFuriousDeath:
The tests fail, so this does seem quite correct. It should be noted that the doc for the width parameters says 'Specifies the desired width of the line in pixels. The actual width may be less. '. Maybe the right solution is some sort of compromise?

kjarosh:
I don't really have anything more to say than "it just needs a lot of testing and probably implementing some missing stuff too" :D

And also the title is misleading, as this touches FTE, not `EditText` per se (it's during creation of `TextLine`, not `TextField`).

SuchAFuriousDeath:
I was looking at 'Dimension Y is zero' errors and it seems that this is the same issue as https://github.com/ruffle-rs/ruffle/issues/13946

Issue #12633, opened 2023-08-08
Ruffle .deb or appimage file


alephdfilms:
**Describe the feature**
I'd think that a Ruffle appimage or .deb file would make the installation of Ruffle (at least, on Linux) easier.

Croworbit:
I agree, a .deb would be amazing +1

torokati44:
There is some recent work on AppImage (https://github.com/ruffle-rs/ruffle/pull/12402), and .deb Shouldn't Be That Difficult:tm: either: https://github.com/vasilakisfil/hello.service

alephdfilms:
Huh. Interesting.
However, it would be nice to have an official version. :)

kevin-wijnen:
> There is some recent work on AppImage (#12402)

The last time I checked, there was a bug (as far as I could see) with the intended tool I wanted to use, but I've not gotten a response since I posted an issue in their repository in July.

I intend to take another look at the end of the month or November again, but the big limitation holding me back is my lack of knowledge on Rust in general, sadly. I'll have to do another quick scan to see if the landscape with AppImage packing for Rust apps has changed at all and try to fiddle around with those.

kevin-wijnen:
> There is some recent work on AppImage (#12402), and .deb Shouldn't Be That Difficult™️ either: ~https://github.com/vasilakisfil/hello.service~ https://github.com/kornelski/cargo-deb

[Cargo Packager](https://github.com/crabnebula-dev/cargo-packager) seems like another solution that would solve both AppImage and .deb packages. I'll take a look at it myself for AppImage packaging. If I can get that to work, I'll give .deb packaging a shot as well.

fiftydinar:
There is this for the AppImage:
https://github.com/pkgforge-dev/Ruffle-AppImage

It uses Anylinux-AppImages tooling, which makes it works well in new, old or `musl` distros.

PR #22381, opened 2025-12-01
ci: Add `Makefile` and `debian` directory

donmor:
`Makefile` allows quick build/installation using `make`/`sudo make install`. `debian` directory adds basic support for building debian package using `Makefile` (target `version` updates `debian/changelog` and `deb` calls `dpkg-buildpackage -us -uc`).

Build dependencies can be installd by:
```
sudo apt install rustc cargo # add "-t trixie-backports" on trixie
sudo apt install libasound2-dev libudev-dev
```

Impl. #12633 .


debian/control
```diff
@@ -0,0 +1,24 @@
+Source: ruffle-nightly
```
kjarosh: Can you put all these files (including the Makefile) in `ruffle/desktop/packages/linux/debian`? If this produces more than one package (or has a potential to), it should be placed in a subdirectory there, e.g. `ruffle/desktop/packages/linux/debian/ruffle`, similarly to AUR

debian/changelog
```diff
@@ -0,0 +1,5 @@
+ruffle-nightly (20251201) UNRELEASED; urgency=medium
```
kjarosh: The changelog should start empty, and it should be updated by the `.github/scripts/release.py` script.

debian/control
kjarosh: How do you get the source package? Why does it have "nightly" in name, but this package doesn't?

Makefile
```diff
@@ -0,0 +1,33 @@
+.PHONY: all clean distclean install uninstall version deb
```
kjarosh: How does this Makefile should be used? Does it have to support some tooling (e.g. `DESTDIR`, target names)?

Makefile
```diff
@@ -0,0 +1,33 @@
...
+install: ruffle_desktop
+	install -d $(DESTDIR)$(prefix)/bin/
+	install -m755 $^ $(DESTDIR)$(prefix)/bin/
+
+uninstall:
+	-rm -f $(DESTDIR)$(prefix)/bin/ruffle_desktop
+
+version:
+	@echo $(VERSION)
+	@sed -i '1s/([0-9]\{8\})/($(VERSION))/' debian/changelog
```
kjarosh: This should not be needed after the changelog starts being properly updated.

Makefile
```diff
@@ -0,0 +1,33 @@
+.PHONY: all clean distclean install uninstall version deb
+
+prefix ?= /usr/local
+VERSION := $(shell echo $(notdir $(CURDIR)) | tr -cd '0-9')
+ifeq ($(VERSION),)
+	VERSION := $(shell date +%Y%m%d)
+endif
```
kjarosh: This should not be needed after the changelog starts being properly updated. (Not to mention it doesn't work properly.)

debian/changelog
kjarosh: Can't we use the version we have in Cargo.toml? Why does it say `ruffle-nightly`?

Makefile
```diff
@@ -0,0 +1,33 @@
...
+
+all: ruffle_desktop
+
+clean:
```
kjarosh, line 26: Shouldn't it also clean the ignored files in `/debian`?

desktop/packages/linux/debian/control
```diff
@@ -0,0 +1,24 @@
+Source: ruffle-nightly
+Section: misc
+Priority: optional
+Maintainer: unknown <localhost>
+Rules-Requires-Root: no
+Build-Depends:
+ debhelper-compat (= 13),
+ rustc (>= 1.86),
```
kjarosh, line 8: This will definitely desync. Maybe setting up a CI would somehow address it

debian/control
kjarosh: We'll need to have a CI set up so that this package is being built automatically. Otherwise it will decay, and merging it doesn't make sense, because we won't be supporting it properly.

It doesn't have to produce release artifacts yet, but it should be built at least once per day to make sure it doesn't start failing.

Makefile
```diff
@@ -0,0 +1,33 @@
...
+	-rm -f ruffle_desktop
+	-rm -rf target
+
+distclean: clean
+
+ruffle_desktop:
+	cargo build --release --package=$@
+	install -m755 target/release/$@ ./$@
+
+install: ruffle_desktop
```
kjarosh: I think you missed some stuff. We have a desktop file, metainfo, and an icon which should be installed too, along the binary.

kjarosh:
A general question: how this package is going to be used? Is it meant to be something we can attach as a release artifact? Do users have to build it themselves? Have you thought about just submitting this package to debian repositories instead?

debian/control
donmor: I'd reconsider this later...

debian/changelog
donmor: Without this I can't test the whole thing.

debian/control
donmor: This `debian` directory has a `3.0 (native)` format, which has to be in root directory of source code. Maybe we should use the more complicated `3.0 (quilt)` format?

Makefile
donmor: Currently `DESTDIR` and `prefix` is supported.

Makefile
donmor: Yep

debian/changelog
donmor: I initially get this from tags (which are nightly). Will change to that in Cargo.toml later.

Makefile
donmor: Will add this later

desktop/packages/linux/debian/control
donmor: Some dependencies may require a random rust toolchain version (rather than 1.85 for edition2024, etc.), I changed it to 1.86 because it complains about it. There may be a way to detect the minimum required toolchain version (hoping it won't be newer than that in sid...)

Makefile
donmor: I'll add them later.

donmor:
> A general question: how this package is going to be used? Is it meant to be something we can attach as a release artifact? Do users have to build it themselves? Have you thought about just submitting this package to debian repositories instead?

The current one is for building from source tarballs downloaded from tags (it updates changelog from extracted directory name with date, has "nightly" in source name, and is in `3.0 (native)` format). The Makefile is tend to be a  general-propose script (local building/installing, CIs, and other stuff). I'd try making something useful for CIs later.

debian/control
kjarosh: (Disclaimer: I'm not very proficient in debian packaging)

I think we have to talk through the general approach of building the package. I see two approaches: (1) build based on the repository, and (2) build from an external source.

Currently, we have a bit of a mix of both, but I would personally prefer the (2) approach. That's how we did AUR and Flatpak, and it just makes it easier for everyone, because packaging is not tightly coupled with the repo, and there's a hierarchy: the repo (upstream) decides how it should be built, and the package (downstream) builds it the way it should. This also fits with the general approach to packaging software, where the downstream is a separate repo.

So I would really prefer this approach, and IIUC it currently works a bit like that? Based on what you said about the tags and tarballs. But on the other side it doesn't, because it requires to be in a specific directory. Could you elaborate on the flow? Like a step by step instruction on how to produce a package with this PR? I think the (2) approach I described should be easily doable in `ruffle/desktop/packages/linux/debian`—the packaging code would just download the source (the tarball) and do the build as it wishes.

Makefile
kjarosh: Yeah, but does it _have to_ support it in order to be built properly?

debian/changelog
kjarosh: Using approach (1) (building using the repo), this should automatically read the version from Cargo.toml, or using approach (2) (building from an external source), the version can be updated in the release script.

desktop/packages/linux/debian/control
kjarosh: Usually, Ruffle waits several days after a new Rust version is released and then starts using it. So we generally keep up with repos such as Fedora, Arch, but I'm not sure how this works in sid.

Based on my previous interest in packaging Ruffle, we are just incompatible with most distributions, and packaging Ruffle doesn't make sense, because you would be using versions from several months ago, because dependencies cannot keep up. The only two distributions where packaging Ruffle would be doable was Fedora and Arch.

From what I see, there's rustc 1.90.0 on sid, which means the newest Ruffle version working there would be ~1 month old as of today. I think this is acceptable, but looking at the latest stable, it would be ~8 months, for the oldest stable it would be ~5 years. IMO this is unacceptable considering the current development model of Ruffle, as it would produce packages which are basically unsupported at launch. Not to mention we don't have a way of backporting bugfixes, because we don't have stable releases.

But that's also the reason we have the Flatpak app, it should work on every distro and we don't have to worry about dependencies too much.

debian/control
donmor: Debian packaging tools won't download source code; `apt source` does (it requires a source repository).

AFAIK:

`3.0 (native)` tries to build from where the `debian` directory is, and build the source as-is. It also build a source tarball outside the source directory (with local changes). Revisioning is not allowed.

`3.0 (quilt)` requires a properly named source tarball to be placed along with the source directory, and will verify the source tree with the tarball (any local file changes should be in separated patches rather than the source tree). Revisioning is required. A .dsc file and .debian.tar.xz (containing the whole debian directory with control, changelog, rules, patches, etc.) is created during build (for properly extract the source with `dpkg-source -x`)

donmor: Debian packaging tools won't download source code; `apt source` does (it requires a source repository).

AFAIK:

`3.0 (native)` tries to build from where the `debian` directory is, and build the source as-is. It also build a source tarball outside the source directory (with local changes). Revisioning is not allowed.

`3.0 (quilt)` requires a properly named source tarball to be placed along with the source directory, and will verify the source tree with the tarball (any local file changes should be in separated patches rather than the source tree). Revisioning is required. A .dsc file and .debian.tar.xz (containing the whole debian directory with control, changelog, rules, patches, etc.) is created during build (for properly extract the source with `dpkg-source -x`)

kjarosh: Hmmm, based on what I read about the difference between `(native)` and `(quilt)`, `(quilt)` would fit better for Ruffle, as it's not a Debian-native application. We're building a Debian package as a downstream, and coupling Ruffle with Debian that tightly could be problematic IMO.

But the main question is: what do we do with this debian packaging config? Do you want Ruffle to be published in Debian repos? Is it meant to be available for users to build? Do you want us to provide `.deb` packages for each release?

donmor: I'm not sure about this (I'm not a debian maintainer)... Maybe it's a good start to have debian scripts in repository.

Makefile
donmor: I think so. `Makefile` installs files into `$(DESTDIR)$(prefix)/`. `DESTDIR` points to debian's packaging root directory, `prefix` has a default value `/usr/local` (a unix tradition) and is overrided in `debian/rules`.

debian/changelog
donmor: It seems that changelogs are generated upon creating releases (where they're never been written into repository files before creation of the corresponding tag). Maybe we should write a log somewhere in repository everytime a pull request is merged?

kjarosh: I think it's fine to leave an empty changelog, we'll only be packaging releases, no?

debian/control
kjarosh: I'm asking because I don't see how we could use those scripts in the first place. I see issues with every application, but it would be useful to focus on the one we're planning to use, because we need to somehow address those issues before we can merge.

The main issue I see is that Ruffle does not have stable releases (and it's not gonna happen soon, the community doesn't want that), and is inherently incompatible with how Debian releases software and their stabilization process. Dependencies in Debian repos are old compared to what Ruffle uses, and we have _no way_ of supporting older releases. It means that Debian users (using the _latest_ Debian release) would have to use a ~8 months old Ruffle release without _any_ patch in between the release date and today. Something is broken in Ruffle on Debian? You have to wait a year before you can enjoy the fix. (And that is without the guarantee that something else doesn't get broken by the way.)

Makefile
kjarosh: Alright, thanks for the info

donmor:
Patch refactored, now:
- `make install` installs .desktop, metainfo and icon
- Moved `debian` into `desktop/packages/linux/`
- Migrated to `3.0 quilt` format
- Source package `ruffle` now builds package `ruffle-desktop`
- Now get version from `cargo metadata` (`0.1.0` at present)
- Get revision from directory name (any digits, but version is stripped out; use system date if got nothing from directory name)
- `make version` now writes a new entry in `debian/changelog`, if a new revision is used, using content from `.github/changelog.entries` (~maintained by #22390 , may require extra steps~ now generated by CIs)
- `make deb` now do:
  - create `../ruffle_<version>.orig.tar.gz` with `git archive` (if not exists; .orig.tar.xz is valid; make sure working tree is clean); or rename one from `ruffle-nightly-<date>.tar.gz` if found
  - remove any existing `debian` directory and make a new copy from `desktop/packages/linux/debian`
  - run `dpkg-buildpackage -us -uc`

Status:

- [x] `make deb` works in a clean working tree
- [x] `make deb` works in newly extracted nightly release tarballs (renames tarball automatically)
- [x] `changelog.entries` generated by CIs (no longer tracked)
- [x] Debian packages and source package built by CIs
- [x] Debian trivial apt source integrated in github releases (tested, https://github.com/donmor/ruffle/releases/tag/nightly-2025-12-04 built by https://github.com/donmor/ruffle/actions/runs/19916843478), usage: `deb [ trusted=yes ] https://github.com/donmor/ruffle/releases/download/nightly-2025-12-04 /` or `deb [ trusted=yes ] https://github.com/donmor/ruffle/releases/latest/download /` if specified latest

Makefile
donmor: Added

desktop/packages/linux/debian/control
donmor: Dependency check can be skipped in CIs as we have them installed. They're for users who build deb from source (and is definitely skippable).

donmor:
Added webext build script (Makefile, debian, CI)

donmor:
Note that some values in CI yml are changed to match my forked repo, they have to be changed back before merging

Makefile
```diff
@@ -0,0 +1,110 @@
+.PHONY: all clean distclean ruffle_desktop ruffle_webext install uninstall version deb
```
kjarosh: This makefile is debian-specific so it needs to be in the debian/ dir too

PR #22390, opened 2025-12-02
ci: Add actions to write to `changelog.entries` on each merged PRs

donmor:
This workflow runs on each merged and closed PRs, writing an entry at the beginning of `.github/changelog.entries`. Useful in CIs that require changelogs since last tag (by `git diff`)

related: https://github.com/ruffle-rs/ruffle/pull/22381#discussion_r2579197969

### DIFF
--- a/.github/workflows/update-changelog.yml
+++ b/.github/workflows/update-changelog.yml
@@ -1,0 +1,42 @@
+name: Update changelog on PR merge
+
+on:
+  pull_request:
+    types: [closed]
+    branches:
+      - master
+
+jobs:
+  update:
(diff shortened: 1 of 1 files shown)


kjarosh (commented):
We already had a discussion about updating changelog for each nightly release but we decided it's too much. I doubt doing it for each PR will be acceptable, it's effectively doubling the number of commits.

My opinion is that the changelog should be updated for each release, if someone has something to add to it, they should do it manually in their PR, but we do not have a mechanism for that, and it shouldn't affect the packaging.

We definitely don't want to modify our processes specifically for Debian packaging though.

Issue #12460, opened 2023-07-31
Error on https://flashpointproject.github.io/flashpoint-database/search/


clefspeare:
# Error Info
Error name: Error
Error message: panicked at 'ShaderJob.target is not a BitmapData (FIXME - support other types)', core/src/avm2/globals/flash/display/shader_job.rs:160:10
Error stack:
```
339/W/n.wbg.__wbg_new_d258248ed531ff54@https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.aa93540b793048da3495.js:1:51449
core::ops::function::FnOnce::call_once{{vtable.shim}}::h0e63761b8083e365@:wasm-function[9182]:0x8eb983
std::panicking::rust_panic_with_hook::h5cee2a9564faeb6d@:wasm-function[6045]:0x7fa563
core::panicking::panic_fmt::h8a19fa1eb63fbb67@:wasm-function[9026]:0x8e736b
core::option::expect_failed::hd65acbc4e2342087@:wasm-function[7939]:0x8a76f9
ruffle_core::avm2::globals::flash::display::shader_job::start::h3b851341e6b23fbf@:wasm-function[1279]:0x3eef8e
ruffle_core::avm2::function::Executable::exec::hb539235b671154fd@:wasm-function[830]:0x2bcdf6
ruffle_core::avm2::object::TObject::call_property::hcefb4fed0c103884@:wasm-function[1108]:0x385ea0
<ruffle_core::avm2::object::Object as ruffle_core::avm2::object::TObject>::call_property::h0699d031c205c2b4@:wasm-function[2768]:0x616d44
ruffle_core::avm2::activation::Activation::run_actions::hdacb60c3edcba5ee@:wasm-function[516]:0x2a8ec
ruffle_core::avm2::function::Executable::exec::hb539235b671154fd@:wasm-function[830]:0x2bcafa
<ruffle_core::avm2::object::Object as ruffle_core::avm2::object::TObject>::call::hb4519e9213a2d312@:wasm-function[984]:0x33f0bd
ruffle_core::avm2::events::dispatch_event_to_target::ha548825be82313ad@:wasm-function[917]:0x314e6f
ruffle_core::avm2::events::dispatch_event::h6e98c55e210686cb@:wasm-function[1224]:0x3cb3f2
ruffle_core::avm2::Avm2::dispatch_event::hb82daeadd8664f12@:wasm-function[1218]:0x3c8220
[...]

PR #22394, opened 2025-12-02
wgpu: Fix pixel copying when processing pixelbender output

SuchAFuriousDeath:
Fixes the game mentioned at #12460 ('The End') for me and makes it playable (without the changes, a RangeError is thrown. However, the issue also mentions a separate error that does not occur for me (PR probably upcoming on that one too).

The texture format is Rgba32Float for both 3 and 4 channels, so the channel count should only be taken into account when copying the pixels in each row. Since channels_in_raw_pixels would be a constant, I also refactored the surrounding code a little bit.


kjarosh:
I think crafting a test for this should be pretty simple

Issue #12256, opened 2023-07-23
(unknown game) Enemies not spawning or disappearing after spawning


Jmousy:
### Describe the bug

I don't know the original title of this Chinese game.

[i-dont-know-what-game-name-is.zip](https://github.com/ruffle-rs/ruffle/files/12138231/i-dont-know-what-game-name-is.zip)

In the comparison video below (Adobe Flash Player on the left, Ruffle Nightly 2023-07-23 on the right), you can see that very few enemies spawn on the right, as opposed to the left, which spawns correctly.

Also, most of the enemies that do manage to spawn disappear just before they enter the screen.

https://github.com/ruffle-rs/ruffle/assets/75052171/49719cfa-875b-49d9-80d7-c0be51cf2fe5

### Expected behavior

Enemies should spawn correctly, the same as the original.

### Affected platform

Desktop app

### Operating system

Windows 10 22H2

### Browser

Google Chrome 115

PR #22402, opened 2025-12-03
avm1: prototype and function patches

### LINKED ISSUE #6285: avm1: SWFv5 should coerce functions to 0
```
trace(typeof Array); // function
trace(3 - Array); // 3 (NaN in Ruffle)
trace(Array - Array); // 0 (NaN in Ruffle)

function t() {}
trace(typeof t); // function
trace(3 - t); // 3 (NaN in Ruffle)
trace(t - t); // 0 (NaN in Ruffle)


// This behaviour doesn't apply to objects and movie clips.

var a = new Sound();
trace(typeof a); // object
trace(3 - a); // NaN

trace(typeof _root); // movieclip
trace(3 - _root); // NaN

```
In SWFv6 and the next versions, this behaviour no longer applies as NaN is returned, so everything is correct in Ruffle.

---

A quick test shows that replacing
https://github.com/ruffle-rs/ruffle/blob/5e8d8cc4d130c7dfd67e9c8ab2b32d5aa112112e/core/src/avm1/value.rs#L151
with
```
            Value::Object(o) => {
                if activation.swf_version() < 6 && o.type_of() == "function" {
                    return 0.0;
                }
                return f64::NAN;
            }
```
fixes the issue, but there's probably a better way to do this.

zjk261:
This fixes #6285 and #12256 and some from_gnash tests
Most line changes are tests


kjarosh:
Hmmm... This PR covers a lot of things, so it cannot be squashed to one commit.

When you update tests at the very last commit instead of for each change, it means that tests fail between those commits, and effectively it's like merging one commit for bisecting/reverting purposes.

Test outputs need to be updated for each commit so that they pass.

kjarosh:
I will actually try to split it up and post PRs as I'm currently working on similar stuff

core/src/avm1/globals/as_broadcaster.rs
```diff
@@ -11,7 +11,7 @@ use gc_arena::Collect;
 use ruffle_macros::istr;
 
 const OBJECT_DECLS: StaticDeclarations = declare_static_properties! {
-    "initialize" => method(initialize; DONT_ENUM | DONT_DELETE);
+    "initialize" => function(initialize; DONT_ENUM | DONT_DELETE);
```
kjarosh, line 14: What's the difference? There doesn't seem to be any

zjk261: The `define_on` function in `create_class` assigns `None` to the prototype of method, and I found that `trace(AsBroadcaster.initialize.prototype)` prints `[object Object]` in flashplayer, so I changed it to function.

kjarosh: Oh you're right, sorry, I was analyzing a wrong decl kind.

Any idea why it's not showing up in the `avm1/global_proto_decls` test?

zjk261: Maybe because `typeof _global.AsBroadcaster` is function, not object, so recursion stops?

PR #22403, opened 2025-12-03
avm2: Construct all errors in `error.rs`

Lord-McSweeney:


core/src/avm2/globals/flash/net/FileReference.as
```diff
@@ -49,7 +49,30 @@ package flash.net {
...
+
+            // Convert the data to a ByteArray
+            var byteArray:ByteArray = new ByteArray();
+            if (data is ByteArray) {
+                byteArray.writeBytes(data);
+            } else if (data is XML) {
+                byteArray.writeUTFBytes(data.toXMLString());
+            } else {
+                byteArray.writeUTFBytes(data);
+            }
```
moulins, line 66: Moving the `ByteArray` conversion logic into AS3 code forces an extra copy of the byte data, so I think it'd be worth the extra complexity of keeping this logic in Rust.

Lord-McSweeney: FP actually calls these methods on the ByteArray, and it's easier to read (at least for me) this way. What about adding a `take` method to ByteArrayStorage that calls `mem::take` on the stored Vec and using that in `saveInternal`? That would avoid the extra copy.

moulins: That would work :)

PR #22410, opened 2025-12-04
avm1: Don't fall back to `__resolve` in 'special' method calls

moulins:
Makes `from_gnash/misc-ming.all/ResolveEventsTest` pass (`__resolve` isn't used when triggering `onXXX` MovieClip handlers).

Added a new test, checking that primitive coercions through `valueOf/toString` don't use `__resolve` either.

Other 'special' calls are untested, but it's a good bet that they behave similarly.


moulins:
For the newly-added test, the output is the same when compiled as SWFv6, v7 or v8 (MTASC can't compile to SWFv5, so I can't test this).

I don't have access to old FP versions to test with, but Dinnerbone tested all FP versions from 9 to 32 when importing the gnash test, so we at least know that the behavior of the MovieClip handlers isn't FP-dependent.

Issue #22418, opened 2025-12-05
avm1: Incorrect text alignment in the attached file — the text is centered when it should be left-aligned


Toad06:
### Describe the bug

<img width="417" height="283" alt="Image" src="https://github.com/user-attachments/assets/a997a0eb-52c4-4f25-851f-714c02ae80bb" />

### Expected behavior

<img width="417" height="283" alt="Image" src="https://github.com/user-attachments/assets/24a2de1c-6568-4995-a65f-8cd8b0ad2f73" />

---

The issue appears to be caused by the following line:
```as
_root.cadre.texte.html = true;
```

If this value is changed to `false`, the text `"<br />Hi, Ruffle dev!"` is displayed centered in both Ruffle and Flash Player.

The following code also produces the same result (text centered in both Ruffle and Flash Player):
```as
_root.cadre.texte.html = true;
_root.cadre.texte.html = false;
```

However, the following code displays `"<br />Hi, Ruffle dev!"` left-aligned in Flash Player, but centered in Ruffle:
```as
_root.cadre.texte.html = true;
_root.cadre.texte.htmlText = "<br />temp text";
_root.cadre.texte.html = false;
```

This likely indicates that the alignment defined in the `DefineEditText` tag (chid: 210) is reset in Flash Player.

### Content Location

[HERE :)](https://github.com/user-attachments/files/23964425/files.zip)
(Ignore the request to network access when the file loads; it is not required — and doesn't work anyway.)

### Affected platform

Desktop app

### Operating system

Windows 11

PR #22428, opened 2025-12-06
text: Fix default align calculation for HTML fields

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/22418

Align is treated in a special way by Flash Player.  When the text field (`DefineEditText`) is defined as HTML, align is completely ignored.  When it's not defined as HTML, but HTML is enabled dynamically, align is ignored for SWF<8.


core/src/html/text_format.rs
```diff
@@ -639,6 +645,11 @@ impl FormatSpans {
         // but its value is ignored and fields always behave as multiline.
         let is_multiline = is_multiline || swf_version <= 6;
 
+        if swf_version < 8 {
+            // For SWF<8, Flash Player always assumes left align, even if a
```
Lord-McSweeney: This comment looks like it was cut off

Lord-McSweeney (approved):

core/src/html/text_format.rs
kjarosh: Thanks, fixed!

PR #22429, opened 2025-12-06
avm1: Reorder globals to match Flash

kjarosh:
The order of global definitions matter, because it influences enumeration order.  This patch reorders AVM1 globals to match the same order Flash uses.

This allows us to improve upon gnash's argstest* tests, which enumerate over globals and check whether definitions are correct.

This also improves `from_gnash/actionscript.all/flash-v8`.


kjarosh:
Note that for some reason, this causes Ruffle's argstest output to shrink, I'm not sure why, it's probably worth debugging.

kjarosh:
So I mentioned this on Discord already, but it's a step in the right direction that we're missing those lines, because FP does not produce them either.

It's interesting why reordering fixed this though

moulins (approved):

PR #22434, opened 2025-12-06
Add support for blocking hosts in fetch

evilpie:


kjarosh:
Can we extend the compatibility rules mechanism instead?

PR #22435, opened 2025-12-07
core: Add default compatibility rule for blocking Mochiads domains

evilpie:


danielhjacobs:
Supersedes https://github.com/ruffle-rs/ruffle/pull/3511 when merged

core/src/backend/navigator.rs
```diff
@@ -63,6 +63,16 @@ pub enum FetchReason {
     Other,
 }
 
+impl FetchReason {
+    pub fn all() -> Vec<FetchReason> {
```
SuchAFuriousDeath: I think this might as well return the HashSet directly instead of a Vec.

moulins: It would be better to return a `&'static [FetchReason]` IMHO, there's no reason to allocate, and a `HashSet` of three elements is totally overkill.

SuchAFuriousDeath: Good point.

SuchAFuriousDeath: But then maybe the HashSet should not be inside the UrlBlockRule struct? It can be just a bitfield, since FetchReason is an enum.

moulins: True, but a bitfield is a lot of code to write... maybe a simpler solution could be to store a `Cow<'static, [FetchReasons]>`?

SuchAFuriousDeath: Maybe.

SuchAFuriousDeath: Actually, A quick look shows that Ruffle already uses enumset as a dependency. So it would be quite easy to do. Maybe merge this as-is, I can do a PR where I use rewrite this using enumset. There are other occurences of this as well.

kjarosh:
We've talked about this on Discord and no one was against, so let's go with it!

PR #22437, opened 2025-12-07
Minor code improvements

SuchAFuriousDeath:
Cell<[T; N]>::as_array_of_cells and PanicHookInfo::payload_as_str are both stable since 1.91.

### DIFF
--- a/core/src/avm1/globals/color_matrix_filter.rs
+++ b/core/src/avm1/globals/color_matrix_filter.rs
@@ -61,10 +61,9 @@ impl<'gc> ColorMatrixFilter<'gc> {
     }
 
     fn matrix(self, activation: &mut Activation<'_, 'gc>) -> Value<'gc> {
-        // Use `.as_slice_of_cells()` to avoid a copy out of the `Cell`.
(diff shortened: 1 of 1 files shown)


desktop/src/main.rs
```diff
@@ -81,18 +81,10 @@ fn panic_hook(info: &PanicHookInfo) {
...
-            text.trim()
-        } else {
-            text.trim()
-        }
-    } else {
-        panic_text.trim()
-    };
+    let message = info
+        .payload_as_str()
+        .expect("Panic message should always be a &str");
```
moulins: Instead of `expect`ing here (which will cause an immediate abort, IIRC), it would be better to fallback to a dummy message.

SuchAFuriousDeath: Sure. I used except because the only way to trigger it would be if some crate used panic_any, which I think is incredibly niche. But I'll change it, no problem.

SuchAFuriousDeath: Done.

moulins (approved):

PR #22439, opened 2025-12-07
avm1: Reorder properties of various objects to match Flash

kjarosh:
This makes the argstest v8 FP32 diff smaller by 23%.

This also allows tracking globals coverage in AVM1 using argstest.


core/src/avm1/globals/file_reference.rs
```diff
@@ -16,14 +16,6 @@ use gc_arena::{Collect, Gc};
 use ruffle_macros::istr;
 use url::Url;
 
-// There are two undocumented functions in FileReference: convertToPPT and deleteConvertedPPT.
-// Until further reason is given, they will be unimplemented.
-// See:
-// ASSetPropFlags(flash.net.FileReference.prototype, null, 6, 1);
-// for(var k in flash.net.FileReference.prototype) {
-// 	trace(k);
-// }
```
kjarosh, line 25: Removed this comment because it's covered by tests.

core/src/avm1/globals/matrix.rs
```diff
@@ -206,6 +206,9 @@ fn constructor<'gc>(
     if args.is_empty() {
         apply_matrix_to_object(Matrix::IDENTITY, this, activation)?;
     } else {
+        if let Some(d) = args.get(3) {
```
moulins: Are you sure this reordering is correct? Looking at the playerglobals.swf, it initializes non-empty matrices in the expected order: `a, b, c, d, tx, ty`.

core/src/avm1/globals/rectangle.rs
```diff
@@ -48,29 +48,29 @@ fn constructor<'gc>(
...
-            args.get(2).unwrap_or(&Value::Undefined).to_owned(),
+            istr!("y"),
+            args.get(1).unwrap_or(&Value::Undefined).to_owned(),
             activation,
         )?;
         this.set(
-            istr!("height"),
-            args.get(3).unwrap_or(&Value::Undefined).to_owned(),
+            istr!("x"),
+            args.get(0).unwrap_or(&Value::Undefined).to_owned(),
```
moulins: Same as above, per the playerglobals.swf, non-empty rectangles are initialized in the "expected" order.

kjarosh: Fixed, thanks

core/src/avm1/globals/matrix.rs
kjarosh: Fixed, thanks

PR #22440, opened 2025-12-07
 avm2: Store call stack for all errors and print it for primitive errors too

Lord-McSweeney:
For example, a
`throw 3`
will now log
```
Integer(3)
	at Test()
```
instead of
```
AvmError(Integer(3))
```

This makes our behavior closer to Flash's.

### DIFF
--- a/core/src/avm2/error.rs
+++ b/core/src/avm2/error.rs
@@ -1,16 +1,16 @@
 use crate::avm2::activation::Activation;
+use crate::avm2::call_stack::CallStack;
 use crate::avm2::class::Class;
 use crate::avm2::function::display_function;
 use crate::avm2::method::Method;
 use crate::avm2::multiname::Multiname;
 use crate::avm2::object::{ClassObject, ErrorObject};
 use crate::avm2::value::Value;
-use crate::string::AvmString;
+use crate::string::{AvmString, WString};
 
 use quick_xml::errors::{Error as XmlError, SyntaxError as XmlSyntaxError};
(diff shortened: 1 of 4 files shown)


core/src/avm2/error.rs
```diff
@@ -19,16 +22,34 @@ enum ErrorData<'gc> {
...
-        Error(Box::new(ErrorData::AvmError(error)))
+    fn avm_error(error: Value<'gc>, call_stack: CallStack<'gc>) -> Self {
+        Error(Box::new(ErrorData::AvmError(error, call_stack)))
+    }
+
+    pub fn from_value(activation: &mut Activation<'_, 'gc>, error: Value<'gc>) -> Self {
+        // If we're passed an `ErrorObject`, use its call stack. Otherwise,
+        // capture the current call stack.
+        let call_stack = if let Some(error) = error.as_object().and_then(|o| o.as_error_object()) {
+            error.call_stack().clone()
```
moulins: Cloning the callstack is be relatively expensive, maybe `AvmError` could be split into `ErrorData::AvmError(ErrorObject<'gc>)` and `ErrorData::AvmValue(Value<'gc>, CallStack<'gc>)` to avoid this?

Lord-McSweeney: Nice catch. I haven't put the fast-path for formatting for the `AvmError` variant in the `match` statement, as in the future we want to call `toString` on all error values to match FP anyway. What do you think about it now?

moulins: LGTM! :)

Issue #22442, opened 2025-12-07
Register the movie objects inserted by WASM-Bindgen to the JS DOM


sombraguerrero:
This was existing Flash functionality, so this is somewhere between an enhancement request and a bug report, but anyway....

Brought to light in manually testing #20434 --
We should create a bridge between WASM-BindGen and the WebIDL accessor properties so that when it's adding nodes to the tree, the movie object(s) (whether it's part of an embed or an object) are accessible via dot syntax and bracketed indexing.
Syntax such as `window.document[movieName]` or even just `document.movieName` should be valid.
Currently, because they only exist as nodes on the DOM tree, they are accessible by DOM methods like document.getElementById, but a lot of older Flash code leveraged Flash's built-in DOM registration to function, like the Sedea application in Onda Educa.

As an example, see [JS DOM Play Test](https://sombrastorage.z19.web.core.windows.net/ruffle/js_flash)

danielhjacobs:
On that page, `window.movieemb` works just fine (as that's a way to get elements by id without `getElementById` even on modern browsers). What W3C specification guarantees `window.document["movieemb"]` or `document.movieemb` to work for embeds, out of curiosity?

danielhjacobs:
Seems it's about the `name`, not the `id`.

danielhjacobs:
It's this in W3C, the rules are quite complex: https://html.spec.whatwg.org/multipage/nav-history-apis.html#named-access-on-the-window-object

> As a general rule, relying on this will lead to brittle code. Which IDs end up mapping to this API can vary over time, as new features are added to the web platform, for example.

danielhjacobs:
Most notably, the browser automatically assigns this for any [embed](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-embed-element), [form](https://html.spec.whatwg.org/multipage/forms.html#the-form-element), [img](https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element), or [object](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-object-element). A `ruffle-embed`, `ruffle-object`, and `ruffle-player` is none of those so we'd have to somehow handle this ourselves.

sombraguerrero:
and that's really kind of the crux of this whole situation, something that has come up plenty of times, especially in #7868 
I should probably just plug this build into Onda and see what happens. I know we still don't have all the properties and methods we need, but......

Toad06:
[...]

PR #22449, opened 2025-12-08
web: Polyfill document.embeds to return ruffle-embeds too

danielhjacobs:
The document.embeds lists all <embed> elements in the current document: https://developer.mozilla.org/en-US/docs/Web/API/Document/embeds. It is possible developers relied on their Flash elements appearing when it was used.

https://sombrastorage.z19.web.core.windows.net/ruffle/js_flash tries to use this or named access on document (see https://github.com/ruffle-rs/ruffle/issues/22442 and https://github.com/ruffle-rs/ruffle/pull/22465) for its getFlashMovieObject function, though it still won't work with this because it also used named access on the properties of `document.embeds`:

```js
function getFlashMovieObject()
{
  if (window.movieemb) 
  {
  let myMovie = window.movieemb; //window is standard to W3C, but only binds the name property
	document.getElementById("f1").innerHTML = "Hit " + myMovie + " directly!";
    return myMovie;
  }
  if (navigator.appName.indexOf("Microsoft Internet")==-1)
  {
	document.getElementById("f1").innerHTML = 'Hit on the embeds!';
    if (document.embeds && document.embeds["movieemb"])
      return document.embeds["movieemb"]; 
  }
  else // if (navigator.appName.indexOf("Microsoft Internet")!=-1)
  {
	  document.getElementById("f1").innerHTML = 'Hit on the ID!';
	  return document.getElementById("movieemb");
  }
}
```


kjarosh:
Can we add a polyfill test for it?

danielhjacobs:
Done

web/packages/selfhosted/test/polyfill/document_embeds/expected.html
```diff
@@ -0,0 +1,2 @@
+<li>There are 3 embeds for document.embeds</li>
```
kjarosh: Can't we do a simple assertion instead of writing this to the document and comparing HTML later?

evilpie:
This breaks documen.embeds == document.embeds and probably other invariants.

danielhjacobs:
> This breaks documen.embeds == document.embeds and probably other invariants.

Fixed in the latest commit

web/packages/selfhosted/test/polyfill/document_embeds/test.ts
```diff
@@ -0,0 +1,50 @@
...
+            await browser.$("#test-container").$("ruffle-embed#emb1"),
+        );
+        await playAndMonitor(
+            browser,
+            await browser.$("#test-container").$("ruffle-embed#emb2"),
+        );
+        await playAndMonitor(
+            browser,
+            await browser.$("#test-container").$("ruffle-embed#emb3"),
+        );
```
kjarosh: Are those needed? Even with autoplay disabled embeds should work, right?

web/packages/selfhosted/test/polyfill/document_embeds/test.ts
```diff
@@ -0,0 +1,50 @@
+import { injectRuffleAndWait, openTest, playAndMonitor } from "../../utils.js";
+import { expect } from "chai";
+
+describe("Document embeds", () => {
+    it("loads the test", async () => {
+        await openTest(browser, `polyfill/document_embeds`);
+    });
+
+    it("Accesses the right number of elements with ruffle", async () => {
```
kjarosh: nit: other cases start with a lowercase letter

web/packages/selfhosted/test/polyfill/document_embeds/test.ts
```diff
@@ -0,0 +1,50 @@
...
+            browser,
+            await browser.$("#test-container").$("ruffle-embed#emb3"),
+        );
+        const documentEmbedsIdentity = await documentEmbedsSelfIdentity();
+        expect(documentEmbedsIdentity).to.equal(true);
+        const embeds1 = await countDocumentEmbeds();
+        expect(embeds1).to.equal(3);
+        await removeEl("#emb1");
+        const embeds2 = await countDocumentEmbeds();
+        expect(embeds2).to.equal(2);
```
kjarosh: Can we dynamically add an embed at the end and check if the number went up?

web/packages/selfhosted/test/polyfill/document_embeds/test.ts
```diff
@@ -0,0 +1,50 @@
...
+            await browser.$("#test-container").$("ruffle-embed#emb2"),
+        );
+        await playAndMonitor(
+            browser,
+            await browser.$("#test-container").$("ruffle-embed#emb3"),
+        );
+        const documentEmbedsIdentity = await documentEmbedsSelfIdentity();
+        expect(documentEmbedsIdentity).to.equal(true);
+        const embeds1 = await countDocumentEmbeds();
+        expect(embeds1).to.equal(3);
```
kjarosh: nit: I think you can inline `embeds1`, `embeds2`, and `documentEmbedsIdentity`, they don't make the code more readable.

web/packages/core/src/internal/register-element.ts
```diff
@@ -86,6 +86,126 @@ export function registerElement(
...
+                        interface CachedCollection extends HTMLCollection {
+                            [CACHE_SYM]?: true;
+                        }
+                        Object.defineProperty(Document.prototype, "embeds", {
+                            get(this: Document): CachedCollection {
+                                const existing = (
+                                    this as unknown as Record<
+                                        symbol,
+                                        HTMLCollection
+                                    >
```
kjarosh: Can you extract this to a variable?

web/packages/selfhosted/test/polyfill/document_embeds/index.html
```diff
@@ -0,0 +1,16 @@
...
+        <title>DOCUMENT EMBEDS</title>
+    </head>
+
+    <body>
+        <div id="test-container">
+            <embed id="emb1" name="fl" src="/test_assets/example.swf" width="200" height="200"></embed>
+            <embed id="emb2" name="fl" src="/test_assets/example.swf" width="200" height="200"></embed>
+            <embed id="emb3" name="fl" src="/test_assets/example.swf" width="200" height="200"></embed>
+        </div>
+        <ul id="output"></ul>
```
kjarosh: I think this is not needed anymore

web/packages/core/src/internal/register-element.ts
```diff
@@ -86,6 +86,121 @@ export function registerElement(
...
+                                        CachedCollection
+                                    >;
+                                const existing = documentWithCache[CACHE_SYM];
+                                if (existing) {
+                                    return existing;
+                                }
+
+                                const nodes = (): NodeListOf<Element> =>
+                                    this.querySelectorAll(
+                                        "embed, ruffle-embed",
```
kjarosh: In case of retries, we'll be modifying the external name, will it work then?

web/packages/selfhosted/test/polyfill/document_embeds/test.ts
```diff
@@ -0,0 +1,56 @@
+import { injectRuffleAndWait, openTest } from "../../utils.js";
+import { expect } from "chai";
+
+describe("Document embeds", () => {
+    it("loads the test", async () => {
+        await openTest(browser, `polyfill/document_embeds`);
+    });
+
+    it("accesses the right number of elements with ruffle", async () => {
```
kjarosh: I also think we're missing tests for retrieving embeds, by name, by index, etc.

web/packages/core/src/internal/register-element.ts
danielhjacobs: Fair point, the failure would usually be if ruffle-embed were custom-registered, which probably should also be returned by document.embeds, so I'll do this:
```ts
                                const nodes = (): NodeListOf<Element> => {
                                    const selectors: string[] = ["embed"];

                                    for (let i = 0; i <= tries; i++) {
                                        selectors.push(i === 0 ? "ruffle-embed" : `ruffle-embed-${i}`);
                                    }

                                    return this.querySelectorAll(selectors.join(", "));
                                };
```

web/packages/core/src/internal/register-element.ts
```diff
@@ -86,6 +86,132 @@ export function registerElement(
                 continue;
             } else {
                 window.customElements.define(externalName, elementClass);
+
+                if (elementName === "ruffle-embed") {
+                    const orig = Object.getOwnPropertyDescriptor(
```
kjarosh: Can we refactor this code into its own function?

kjarosh:
LGTM, but would be great if someone else (that knows more about polyfilling) would also take a look at this

web/packages/core/src/internal/register-element.ts
```diff
@@ -39,6 +39,112 @@ export function lookupElement(elementName: string): Registration | null {
...
+ * @param tries Number of tries before this custom element was defined.
+ */
+function polyfillDocumentEmbeds(tries: number) {
+    const orig = Object.getOwnPropertyDescriptor(Document.prototype, "embeds");
+    if (orig?.get) {
+        const CACHE_SYM: unique symbol = Symbol("ruffle_embeds_cache");
+        interface CachedCollection extends HTMLCollection {
+            [CACHE_SYM]?: true;
+        }
+        Object.defineProperty(Document.prototype, "embeds", {
```
kjarosh: I guess I have one more question: why doesn't it refer back to `embeds` for unknown keys?

Let's say that a website uses another extension (called foo) that makes _some_ embeds work, and suppose it works similarly to Ruffle, changing `embed` to `foo-embed`. `foo` will also have to polyfill `Document.embeds` because of the same reasons, but by using the current approach those two polyfills will conflict with each other.

If I were to write this polyfill, I would assume that `Document.embeds` contains all non-Ruffle embeds, and then refer to it for length and unknown keys, at the same time making sure I don't polyfill it for the second time. Doesn't it make sense?

danielhjacobs: Done, but this required the addition of some post-element retrieval sorting, because unlike a single `querySelectorAll` call, which returns elements in DOM order, by falling back, first the built-in document.embeds would be listed and only later in the fake HTMLCollection would the polyfilled elements be listed, but the HTMLCollection is meant to be sorted in tree order per the spec for HTMLCollection, so I had to add post-element retrieval sorting.

danielhjacobs: Used https://developer.mozilla.org/en-US/docs/Web/API/Node/compareDocumentPosition

web/packages/core/src/internal/register-element.ts
```diff
@@ -39,6 +39,152 @@ export function lookupElement(elementName: string): Registration | null {
...
 }
 
+/**
+ * Polyfill so document.embeds will return ruffle-embeds too.
+ *
+ * @param tries Number of tries before this custom element was defined.
+ */
+function polyfillDocumentEmbeds(tries: number) {
+    const orig = Object.getOwnPropertyDescriptor(Document.prototype, "embeds");
+    if (orig?.get) {
```
kjarosh: Nit: you can invert this if and return early, thus producing smaller indentation.

web/packages/core/src/internal/register-element.ts
```diff
@@ -39,6 +39,152 @@ export function lookupElement(elementName: string): Registration | null {
...
+                        this.querySelectorAll(selectors.join(", ")),
+                    );
+
+                    const result: Element[] = [];
+
+                    let i = 0;
+                    let j = 0;
+
+                    // Per https://dom.spec.whatwg.org/#interface-htmlcollection, sorted in tree order
+                    while (i < baseEmbeds.length || j < ruffleEmbeds.length) {
```
kjarosh: This can be extracted into a generic function that merges two sorted lists into a sorted list (+ we can add some simple tests for it, e.g. based on numbers or something).

web/packages/core/src/internal/register-element.ts
```diff
@@ -39,6 +39,152 @@ export function lookupElement(elementName: string): Registration | null {
...
+                >;
+                const existing = documentWithCache[CACHE_SYM];
+                if (existing) {
+                    return existing;
+                }
+
+                const nodes = (): Element[] => {
+                    // Fallback to existing document.embeds for non ruffle-embed elements
+                    const baseEmbeds = orig.get!.call(this) as HTMLCollection;
+                    const selectors: string[] = ["ruffle-embed"];
```
kjarosh: Getting selectors can also be extracted into a function.

web/packages/core/src/internal/register-element.ts
```diff
@@ -39,6 +39,152 @@ export function lookupElement(elementName: string): Registration | null {
...
+                        return nodes().length;
+                    },
+                });
+
+                base.item = function (index: number): Element | null {
+                    return nodes()[index] ?? null;
+                };
+
+                base.namedItem = function (name: string): Element | null {
+                    const list = nodes();
```
kjarosh: I was thinking if we can do it more efficiently by first using a selector with name, and then falling back to Document.embeds, but it would produce different results in case of name duplication. Just sanity check: is this how it should behave? Get the first element from the list?

web/packages/selfhosted/test/polyfill/document_embeds/index.html
```diff
@@ -0,0 +1,15 @@
...
+    <head>
+        <meta charset="utf-8" />
+        <title>DOCUMENT EMBEDS</title>
+    </head>
+
+    <body>
+        <div id="test-container">
+            <embed id="emb1" name="alpha" src="/test_assets/example.swf" width="200" height="200"></embed>
+            <embed id="emb2" name="beta" src="/test_assets/example.swf" width="200" height="200"></embed>
+            <embed id="emb3" name="beta" src="/test_assets/example.swf" width="200" height="200"></embed>
```
kjarosh, line 13: Can we sprinkle some regular embeds that Ruffle will not polyfill here?

web/packages/core/src/internal/register-element.ts
```diff
@@ -39,6 +39,152 @@ export function lookupElement(elementName: string): Registration | null {
     }
 }
 
+/**
+ * Polyfill so document.embeds will return ruffle-embeds too.
+ *
```
kjarosh: Can you elaborate on why we're falling back to Document.embeds in the polyfill here? Basically what I said in the comment

web/packages/core/src/internal/register-element.ts
danielhjacobs: > is this how it should behave? Get the first element from the list?

Yeah, it is, if you have two embeds with name "test" and different ids, document.embed.namedItem("test").id will give the id of the first. Same is true on a poorly formed page where two embeds have the same id and different names and you get the named Item by id and check its name.

kjarosh: I guess we could make it more efficient by using an iterator instead of a list, but it's not worth it.

web/packages/core/src/internal/register-element.ts
danielhjacobs: Don't want to get too elaborate, so opting for this:

 A website may polyfill document.embeds if it adds custom embed-like elements itself.
 Therefore, we fallback to the existing document.embeds in this polyfill to not conflict.

web/packages/core/src/internal/register-element.ts
danielhjacobs: > (+ we can add some simple tests for it, e.g. based on numbers or something)

Is this covered by the tests that you suggested I expand to include basic embeds? Or do you mean testing the generic function itself? I'm not sure what framework we have to do that.

web/packages/core/src/internal/register-element.ts
```diff
@@ -39,6 +39,191 @@ export function lookupElement(elementName: string): Registration | null {
...
+ * @param a The first array by which to use for the merge.
+ * @param b The second array by which to use for the merge.
+ * @param comesBefore A function that checks if one element comes before another.
+ * @returns The merged array, sorted using comesBefore.
+ */
+function mergeSorted<T>(
+    a: readonly T[],
+    b: readonly T[],
+    comesBefore: (x: T, y: T) => boolean,
```
kjarosh: The standard way to pass a comparison function is to use a `(a,b) -> int` function that returns: `x<0` when `a < b`, `0` when `a == b`, and `x>0` when `a > b`.

```suggestion
    comparator: (x: T, y: T) => number,
```

web/packages/core/src/internal/register-element.ts
```diff
@@ -39,6 +39,191 @@ export function lookupElement(elementName: string): Registration | null {
     }
 }
 
+/**
+ * Use a merge sort to merge two arrays based on a comparison function.
```
kjarosh: Nit: it's not merge sort.


```suggestion
 * Merge two sorted arrays into another sorted array based on a comparison function.
```

web/packages/core/src/internal/register-element.ts
```diff
@@ -39,6 +39,191 @@ export function lookupElement(elementName: string): Registration | null {
...
+    const result: T[] = [];
+
+    let i = 0;
+    let j = 0;
+
+    while (i < a.length || j < b.length) {
+        const x = i < a.length ? a[i] : undefined;
+        const y = j < b.length ? b[j] : undefined;
+
+        if (!x) {
```
kjarosh: This can produce weird results if e.g. `a[i]` is falsly. To fix this, you can change the condition to `while (i < a.length && j < b.length)` and then append leftovers after the loop.

web/packages/core/src/internal/register-element.ts
kjarosh: I meant separate unit tests, like others that are in the `test` directory.

danielhjacobs: Is there a way, from the test directory, to test a non-exported function like `mergeSorted`? Or do you mean I should export it for the purpose of testing it? I don't see any other reason to export it.

kjarosh: We definitely should not export it, I'll look into it

danielhjacobs: After discussion, exported but kept internal

web/packages/core/src/internal/register-element.ts
```diff
@@ -39,6 +39,244 @@ export function lookupElement(elementName: string): Registration | null {
...
+}
+
+/**
+ * Polyfill so document.embeds will return ruffle-embeds too.
+ * A website may polyfill document.embeds if it adds custom embed-like elements itself.
+ * Therefore, we fallback to the existing document.embeds in this polyfill to not conflict.
+ *
+ * @param tries Number of tries before this custom element was defined.
+ */
+function polyfillDocumentEmbeds(tries: number) {
```
kjarosh: Now that I'm thinking: can we put it in a separate file? This would improve modularity, as this polyfill has some code in it.

kjarosh (approved):

web/packages/core/src/internal/register-element.ts
danielhjacobs: Done

kjarosh (approved):

PR #22455, opened 2025-12-08
core/avm2: Make all uncaught errors go through `Avm2::uncaught_error`

Lord-McSweeney:
In the future, this will allow us to easily implement
- calling `toString` on all uncaught errors
- `UncaughtErrorEvents`

### DIFF
--- a/core/src/external.rs
+++ b/core/src/external.rs
@@ -8,7 +8,7 @@ use crate::avm2::object::{
     ArrayObject as Avm2ArrayObject, FunctionObject as Avm2FunctionObject, Object as Avm2Object,
     ScriptObject as Avm2ScriptObject, TObject as _,
 };
-use crate::avm2::{FunctionArgs, Value as Avm2Value};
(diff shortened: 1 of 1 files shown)


core/src/avm2.rs
```diff
@@ -703,4 +689,19 @@ impl<'gc> Avm2<'gc> {
...
+
+    // Report an uncaught AVM2 error.
+    // TODO should the `display_object` parameter be optional or not?
+    #[cold]
+    #[inline(never)]
+    pub fn uncaught_error(
+        _activation: &mut Activation<'_, 'gc>,
+        _display_object: Option<DisplayObject<'gc>>,
+        error: Error<'gc>,
+        info: &str,
```
moulins, line 682: Nit: this could take `info: fmt::Arguments<'_>` (paired with `format_args!` at the callsites) to avoid having to allocate temporaries for non-static messages.

Lord-McSweeney: I don't think the extra `format_args!` everywhere is worth it- these are supposed to be error paths anyways

PR #22458, opened 2025-12-08
web: Fix navigating to URLs with query params

Dinnerbone:
This fixes an issue reported on Discord.

When we try to navigate to a url that already has a query string, it will be removed unless the method is POST.

### DIFF
--- a/web/src/navigator.rs
+++ b/web/src/navigator.rs
@@ -270,6 +270,23 @@ impl NavigatorBackend for WebNavigatorBackend {
                     form.set_target(target);
                 }
 
+                if navmethod == NavigationMethod::Get {
+                    // Browsers will clobber any query string with "the whole form input", so we need to re-add them here
+                    for (key, value) in url.query_pairs() {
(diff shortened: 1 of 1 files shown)


kjarosh:
I think we can add a web test for it, it should be pretty straightforward

n0samu:
Why do we use a hidden form submission to navigate to URL rather than something like window.location.assign()?

Dinnerbone:
> Why do we use a hidden form submission to navigate to URL rather than something like window.location.assign()?

There's a path that does that, but it seems to use this if there's a chance that it might not be GET or wants to encode additional parameters. I guess we could specialize that further if it's GET+`_self`+no arguments, but, this code path should work in any case.

Dinnerbone:
> I think we can add a web test for it, it should be pretty straightforward

I'm struggling to make a test for this, webdriverio (or specifically, the browsers themselves) do not seem to like pages navigating if not done by the driver itself.

PR #22467, opened 2025-12-09
tests: Add avm1/global_proto_decls* tests

kjarosh:
These tests verify declarations of global objects and their prototypes, so that we can track how they are defined and with what flags.


moulins:
Isn't that already covered by the `from_gnash/actionscript/all/argtests` tests?

kjarosh:
Nope, the main point of argstest is to call everything and check if it crashes, it does not verify the prototype at all. The common part is iteration order and sometimes some properties, but argstest also lacks flag checking.

kjarosh:
Marked owned properties as "own".

PR #22469, opened 2025-12-09
web: Remove npm token in favour of OIDC

Dinnerbone:
Our npm publishing will not work now. I've configured the repo over on npm side, and we need to make the workflow generate oidc tokens to proceed.

Absolutely untested, but according to the docs it's all that's needed.

https://github.blog/changelog/2025-12-09-npm-classic-tokens-revoked-session-based-auth-and-cli-token-management-now-available/

### DIFF
--- a/.github/workflows/release_nightly.yml
+++ b/.github/workflows/release_nightly.yml
@@ -517,8 +517,6 @@ jobs:
         run: npm publish --tag latest --access public --provenance
         continue-on-error: true
         working-directory: web/packages/selfhosted/dist
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
 
       - name: Package selfhosted
         run: zip -r "${{ needs.create-nightly-release.outputs.package_prefix }}-web-selfhosted.zip" .


.github/workflows/release_nightly.yml
```diff
@@ -229,7 +233,7 @@ jobs:
 
   build-mac-universal-binary:
     name: Build macOS universal binary
-    needs: [create-nightly-release, build, build-web]
+    needs: [ create-nightly-release, build, build-web ]
```
kjarosh: Why those whitespace changes?

Dinnerbone: IDE auto formatting, apparently. I'll amend.

danielhjacobs:
My experience in GitHub actions is that once some permissions are specified where previously none were specified the ones with permissive defaults get more restrictive. Just saying that in case it happens here. I don't know which may need to be specified though, if any.

danielhjacobs:
It's why this has all the permissions it does:

https://github.com/ruffle-rs/ruffle/blob/a0131d451661ea7af891f748e0643a661dd86132/.github/workflows/release_nightly.yml#L360-L367

Dinnerbone:
> permissions:

Huuuh, that already has id-token. Hm. Maybe it's enough to just remove the old token stuff then?

Dinnerbone:
Updated to just remove token

Issue #22438, opened 2025-12-07
Ruffle's 'hide this menu' uses a ruffle specific control to disable the context menu


Croworbit:
### Describe the bug

I'm making a game and one of the settings I have in the game is to enable or disable the context menu. I had a problem where hiding the menu causes you to be unable to show it again, so I thought enabling it through my game's settings menu would make it possible to show it again after clicking the 'hide' button.

instead, I found that the 'hide' is ruffle specific. it made sense to me that it was just using a `FSCommand("showMenu", false);` or `Stage.showMenu=false`, but it isn't, 
and I can't make it return with `Stage.showMenu=true`

the reason I'm doing this is because I want mobile users to click on my 'mouse controls' button when starting the game. I can automatically hide the context menu so I don't have to explain to them why the context menu is popping up everytime they long press or drag, nor tell them how to stop it from happening.

and I want my 'mouse controls' users on a PC to still have access to the context menu (and they can mix and match setting if they want it off, on a PC, or on, on a phone)

I can show and hide the context menu just fine, the code works in ruffle just fine
but...
as soon as the 'hide' button is pressed, it will never again appear, regardless of my code.

I definitely understand why that button was put there, but I'm disappointed its not hijacking the stage value.

so I might as well not have the menu be automatically be disabled and have an explanation pop up anyway, because I would need to explain the ease of the permanent removal of the menu, which is frustrating.

### Expected behavior

this is not an issue on original flash. 
flash never had such a 'hide' option

I'm not sure what the solution is. 
on mobile, (on swfs that are not mine) if a user has hidden the menu, there isn't a good way of giving them an option to show it again, and ruffle's controls are *in* the context menu, which the user wouldn't want appearing to show a single option to show again the rest.

there doesn't even need to be a way to show it again on mobile

I just hope that when I call stage menu true, it will be able to re-appear. the user can then hide it with the ruffle 'hide' again, or the setting in my game.

### Content Location

I have this swf that shows what I mean, but it is a very big pain to demonstrate or test. 
if you run this on desktop or web from a PC, the 'hide' option never appears.
if you run this SWF on mobile web, there is no keyboard to test anything.
if you run this SWF on mobile app, the app crashes.
if you run this with mobile app and choose load from URL, you have to host this swf, where you will then have a keyboard... but there is no 'hide' option.

so you will need to host this on a website, then use the extension, then open firefox's dev tools, then open the Responsive design mode, then right click and test from there.

[...]

PR #22476, opened 2025-12-10
web: Make "Hide this menu" only hide context menu on long-press

danielhjacobs:
Relates to #22438

### DIFF
--- a/web/packages/core/src/internal/player/inner.tsx
+++ b/web/packages/core/src/internal/player/inner.tsx
@@ -174,9 +174,6 @@ export class InnerPlayer {
     // so avoid shadowing it.
     private readonly contextMenuElement: HTMLElement;
 
-    // Allows the user to permanently disable the context menu.
-    private contextMenuForceDisabled = false;
-
     // Whether the most recent pointer event was from a touch (or pen).
     private isTouch = false;
     // Whether this device sends contextmenu events.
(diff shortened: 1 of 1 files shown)


web/packages/core/src/internal/player/inner.tsx
```diff
@@ -1536,12 +1533,18 @@ export class InnerPlayer {
         });
         // Give option to disable context menu when touch support is being used
         // to avoid a long press triggering the context menu. (#1972)
-        if (this.isTouch) {
+        if (
+            this.isTouch &&
+            this.loadedConfig?.contextMenu !== ContextMenu.RightClickOnly
```
kjarosh, line 1538: Why do we need to add this condition here? If it's right click only, we can't get `isTouch` to equal true, right?

danielhjacobs: This is incorrect due to devices with both mouse and keyboard. The moment one touches the Ruffle embed/object with a pointerType of "pen" or "touch" `isTouch` is set to true, and it's still true even if the next click is done via mouse.

kjarosh: That means the doc of `isTouch` is wrong as well as its name. Can you fix the doc and rename it to `hasTouch`?

PR #22479, opened 2025-12-11
avm1: Fix and improve interface logic

moulins:
- The SWF version check should happen during `ImplementsOp`, not `instanceof`;
- `ImplementsOp` should not coerce its arguments;
- "super-interfaces" are "interfaces of interfaces", not "superclasses of interfaces" (contrarily to what the AS2 syntax implies: `IFoo extends IBar`).

See the expanded test case for more details.

### DIFF
--- a/core/src/avm1/object.rs
+++ b/core/src/avm1/object.rs
@@ -320,7 +320,7 @@ impl<'gc> Object<'gc> {
         // TODO(moulins): should we guard against infinite loops here?
         // A recursive prototype and/or interface chain will hang Flash Player.
 
-        let mut interface_stack = vec![];
+        let mut interface_stack = smallvec::SmallVec::<[_; 4]>::new();
         let mut this = self;
 
         while let Value::Object(this_proto) = this.proto(activation) {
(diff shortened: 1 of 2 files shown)


core/src/avm1/object.rs
```diff
@@ -310,44 +310,31 @@ impl<'gc> Object<'gc> {
...
-    /// can't implement interfaces within interfaces (fortunately), but if you
-    /// somehow could this would support that, too.
+    /// check the prototype chain and call it a day: each step in the chain has
+    /// its own attached 'interface tree' which also needs to be checked.
     pub fn is_instance_of(
         self,
         activation: &mut Activation<'_, 'gc>,
-        constructor: Object<'gc>,
         prototype: Object<'gc>,
     ) -> Result<bool, Error<'gc>> {
```
Lord-McSweeney, line 319: Can this function be made infallible now, since it never returns an `Err`?

moulins: Hm, I'm not sure... if we want to add some kind of protection against infinite loops in the future, we'll need the error return.

PR #22485, opened 2025-12-11
web: For all tests that polyfill with Ruffle, wait for element to exist

danielhjacobs:
I can switch to `playAndMonitor`, but this should suffice and many tests do this already.

### DIFF
--- a/web/packages/selfhosted/test/polyfill/embed_default/test.ts
+++ b/web/packages/selfhosted/test/polyfill/embed_default/test.ts
@@ -12,6 +12,7 @@ describe("Embed tag", () => {
 
     it("polyfills with ruffle", async () => {
         await injectRuffleAndWait(browser);
+        await browser.$("<ruffle-embed />").waitForExist();
         const actual = await browser
             .$("#test-container")
             .getHTML({ includeSelectorTag: false, pierceShadowRoot: false });
--- a/web/packages/selfhosted/test/polyfill/embed_missing_type/test.ts
+++ b/web/packages/selfhosted/test/polyfill/embed_missing_type/test.ts
@@ -12,6 +12,7 @@ describe("Embed without type attribute", () => {
 
     it("polyfills with ruffle", async () => {
         await injectRuffleAndWait(browser);
+        await browser.$("<ruffle-embed />").waitForExist();
         const actual = await browser
             .$("#test-container")
             .getHTML({ includeSelectorTag: false, pierceShadowRoot: false });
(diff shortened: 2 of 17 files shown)


danielhjacobs:
Unsure how to deal with these, which are testing that Ruffle does **not** polyfill certain elements: 
- <https://github.com/ruffle-rs/ruffle/blob/master/web/packages/selfhosted/test/polyfill/embed_youtube/test.ts>
- <https://github.com/ruffle-rs/ruffle/blob/master/web/packages/selfhosted/test/polyfill/object_youtube/test.ts>
- <https://github.com/ruffle-rs/ruffle/blob/master/web/packages/selfhosted/test/polyfill/object_inside_audio/test.ts>
- <https://github.com/ruffle-rs/ruffle/blob/master/web/packages/selfhosted/test/polyfill/embed_inside_audio/test.ts>
- <https://github.com/ruffle-rs/ruffle/blob/master/web/packages/selfhosted/test/polyfill/embed_missing_src/test.ts>
- <https://github.com/ruffle-rs/ruffle/blob/master/web/packages/selfhosted/test/polyfill/pdf_with_get/test.ts>
- <https://github.com/ruffle-rs/ruffle/blob/master/web/packages/selfhosted/test/polyfill/pdf/test.ts>
- <https://github.com/ruffle-rs/ruffle/blob/master/web/packages/selfhosted/test/polyfill/embed_wrong_type/test.ts>
- <https://github.com/ruffle-rs/ruffle/blob/master/web/packages/selfhosted/test/polyfill/object_missing_data/test.ts>

danielhjacobs:
Realistically, the only option for the ones where we're testing if elements do not polyfill with Ruffle is `await browser.pause(5000)` I guess, but that will make them take at least 5 seconds unconditionally. A lower number may work too.

danielhjacobs:
I can't imagine this is bad to do, but neither `await browser.$("<ruffle-embed />").waitForExist();` nor `playAndMonitor` seem to be sufficient to stop `embed_unexpected_string` failing.

Edit: Unrelated issue

danielhjacobs:
Reviewer information: this is to account for the small chance that the tests run after line 134 here finishes and before line 144 finishes:

https://github.com/ruffle-rs/ruffle/blob/e31d83463d25234cdbb652d02ae5e40ccf92d4ae/web/packages/core/src/public/setup/public-api.ts#L134-L144

For the tests that Ruffle does **not** polyfill certain elements, #22490 was opened. It adds **to the Public API** a new boolean, initialized, which is set to true **after** the above lines. These tests could alternatively use the new function added there, but this could also be merged as-is. Either way works.

PR #22490, opened 2025-12-11
web: Add way to check for initialization for tests that check for no polyfilling

danielhjacobs:
This can be used for tests that check if polyfilling occur too and for those should eliminate the need to wait for ruffle-embed/ruffle-object to exist before continuing but at the same time for those tests perhaps checking those elements exist may be considered a better condition for which to wait anyway, since that's when the tests should have their expected.html anyway.


danielhjacobs:
Note: **This adds to the PublicAPI a new boolean initialized, which is set to true when Ruffle has done all its polyfilling**. That's arguably a good thing, I can see this variable being useful to certain people.

PR #22516, opened 2025-12-14
avm1: Fix `LocalConnection` properties

kjarosh:
`LocalConnection` properties are properly reordered and `isPerUser` is added.


kjarosh:
Oh wait, I can just do a read-only property. For some reason I thought it won't work.

PR #22517, opened 2025-12-14
avm1: Stub `flash.automation.*` classes

kjarosh:
This patch stubs:
* `flash.automation.Configuration`,
* `flash.automation.ActionGenerator`,
* `flash.automation.StageCapture`.


core/src/avm1/globals/automation_action_generator.rs
```diff
@@ -0,0 +1,73 @@
...
+    "generateActions" => method(generate_actions);
+    "valueOf" => method(value_of);
+    "toString" => method(to_string);
+};
+
+pub fn create_class<'gc>(
+    context: &mut DeclContext<'_, 'gc>,
+    super_proto: Object<'gc>,
+) -> SystemClass<'gc> {
+    let class = context.class(constructor, super_proto);
```
moulins: It'd be cleaner to use `context.empty_class` instead of an explicit empty constructor.

core/src/avm1/globals/automation_configuration.rs
```diff
@@ -0,0 +1,87 @@
...
+    "setDeviceConfiguration" => method(set_device_configuration);
+    "valueOf" => method(value_of);
+    "toString" => method(to_string);
+};
+
+pub fn create_class<'gc>(
+    context: &mut DeclContext<'_, 'gc>,
+    super_proto: Object<'gc>,
+) -> SystemClass<'gc> {
+    let class = context.class(constructor, super_proto);
```
moulins: Same as above.

moulins:
The `playerglobals.swf` shows the attributes as being:
- `DONT_ENUM, DONT_DELETE` for `StageCapture`;
- `DONT_ENUM, DONT_DELETE, READ_ONLY` for `ActionGenerator` and `Configuration`.

Is the discrepancy intentional?

core/src/avm1/globals/automation_action_generator.rs
kjarosh: Done

core/src/avm1/globals/automation_configuration.rs
kjarosh: Done

kjarosh:
Hmmm, I thought the attributes were already tested; I'll take a look later

kjarosh:
> The `playerglobals.swf` shows the attributes as being:
> 
>     * `DONT_ENUM, DONT_DELETE` for `StageCapture`;
> 
>     * `DONT_ENUM, DONT_DELETE, READ_ONLY` for `ActionGenerator` and `Configuration`.
> 
> 
> Is the discrepancy intentional?

@moulins I'm not sure if that's what you're referring to, but `playerglobals.swf` do not actually set these flags. It looks like they made a mistake, because
1. for `StageCapture` the flags are set for `StageCapture`'s prototype (o is not being assigned properly), and
2. `ASSetPropFlags(ActionGenerator,null,7);` and `ASSetPropFlags(Configuration,null,7);` do nothing, because `ActionGenerator` and `Configuration` are undefined.

And as mentioned above, these things are tested, so I just made sure the flags are set the same way as observed in tests.

PR #22547, opened 2025-12-19
avm1: Fix XMLSocket dropping data when TCP segments lack null bytes

### LINKED ISSUE #18702: Ruffle Desktop does not handle multiple XMLSocket TCP segments
### Describe the bug

1. Open https://xarium.cc/playtest.php, reload page and copy cookies
2. Enable debug
3. Launch Ruffle Desktop using command:
```shell
./ruffle https://xarium.cc/assets/swf/serial_experiments_lain_testruffle.swf --cookie "[cookies]" -P start_step=0 -P manual_server_selection=true --base "https://xarium.cc/"
```
4. Select server in list
5. You will get stuck on the loading banner. In the console you will see the following log (or something like that):
```
2024-11-22T01:27:26.802744Z  INFO avm_trace: OnData: der=&quot;0&quot;/>&lt;item Id=&quot;4&quot; Order=&quot;0&quot;/>&lt;item Id=&quot;3&quot; Order=&quot;0&quot;/>&lt;item Id=&quot;2&quot; Order=&quot;0&quot;/>&lt;item Id=&quot;1&quot; Order=&quot;0&quot;/>&lt;/smiles>&lt;sms NewCount=&quot;0&quot; />&lt;object_groups>&lt;item UOGId=&quot;-6297&quot; ObjectGroupId=&quot;6297&quot; ObjectGroupTypeId=&quot;101&quot; ControllerType=&quot;4&quot; Stored=&quot;0&quot; ContorollerMRId=&quot;39039&quot; />&lt;/object_groups>&lt;options>&lt;item Name=&quot;show_item_ids&quot; Value=&quot;false&quot; />&lt;item Name=&quot;whats_new_data&quot; Value=&quot;3|3352|1651945322312&quot; />&lt;item Name=&quot;pet_battle_news&quot; Value=&quot;false&quot; />&lt;/options>&lt;platform Type=&quot;0&quot; />&lt;/response>"/>
```

### Expected behavior

The game launches successfully.
In the console you will see log of full packet, **starting with** **`<cb uid="...`:** (this is a big escaped XML inside another XML)

Image

### Content Location

https://xarium.cc/assets/swf/serial_experiments_lain_testruffle.swf

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

_No response_

### Additional information

Game client connects to server at port 8000 using XMLSocket.
I modified the client to log any incoming data.
Small packets work fine:
Image
but when a large packet arrives that is split into multiple TCP segments, Ruffle handle only the last segment.

It works fine in browser.

[Wireshark dump](https://github.com/user-attachments/files/17854093/xarium_ruffle.zip)

SuchAFuriousDeath:
Fixes #18702 (To be fully transparent, I haven't tested with the actual swf, but this is 100% the issue that causes what is described there)


torokati44:
Was it investigated what happens in the inverse scenario: when there are (possibly multiple) `NUL` bytes within a TCP segment? (Sorry I haven't looked into it, this is just a drive-by quesh...)

SuchAFuriousDeath:
Not by me :D

However, that is in the if branch, my code is in the else branch. I can take a look though, it sounds interesting

kjarosh:
I haven't looked into this code much, but TCP has no concept of messages. Semantically it's a stream and it can get segmented in any way possible. So we should always look at it as a stream of bytes and never use the information where a segment ended for anything.

tests/tests/swfs/avm1/xml_socket_segmented/socket.json
```diff
@@ -0,0 +1,29 @@
+[
```
kjarosh, line 1: This definitely needs different types of segmentation. The null byte should be everywhere: at the beginning, at the end, as a separate segment, in the middle, multiple null bytes in one segment, etc.

SuchAFuriousDeath: Done.

SuchAFuriousDeath:
I reworked the data handling logic a little bit so that it is less convoluted, it should be equivalent to the old code.

core/src/socket.rs
```diff
@@ -317,52 +317,29 @@ impl<'gc> Sockets<'gc> {
...
-                                    // We had leftover bytes, so append them to XmlSocket internal read buffer,
-                                    // to be used when the next packet arrives.
-                                    xml_socket.read_buffer().extend(data);
-                                }
+                            // Prepend any buffered data from previous packets.
+                            let mut data: Vec<u8> =
+                                xml_socket.read_buffer().drain(..).chain(data).collect();
+
+                            // Process all complete messages (null-terminated).
+                            while let Some(null_pos) = data.iter().position(|&b| b == 0) {
```
kjarosh: This implementation looks correct, but doesn't look too efficient. In case of large objects (or small segments), we'll be copying and iterating through buffers all the time.

Have you tried using a ring buffer instead? You can safely assume that the current buffer doesn't contain a null byte, check if new data contains it (you need to iterate over it anyway), append the new data, and if a null byte was appended, process messages until all nulls are gone.

SuchAFuriousDeath: I'll try to cook something up. Maybe you can check https://github.com/ruffle-rs/ruffle/pull/22606 in the meantime? :sweat_smile:

SuchAFuriousDeath: Ok I tried something. It requires calling make_contiguous in one place, but it should be fine.

core/src/socket.rs
```diff
@@ -317,51 +317,36 @@ impl<'gc> Sockets<'gc> {
...
-                                        break;
+                            xml_socket.read_buffer().extend(data);
+
+                            // Process complete messages (null-terminated) one at a time.
+                            // We release the buffer borrow before each AS call to avoid
+                            // conflicts if AS code accesses the socket.
+                            loop {
+                                let message = {
+                                    let buffer = &mut *xml_socket.read_buffer();
+                                    match buffer.iter().position(|&b| b == 0) {
```
kjarosh: It's currently `O(n^2)`, but can be `O(n)` if we assume `xml_socket.read_buffer()` doesn't contain null bytes, in which case we need to iterate only over `data`.

core/src/socket.rs
```diff
@@ -317,52 +317,46 @@ impl<'gc> Sockets<'gc> {
...
-                                    .chain(data.drain(..index))
-                                    .collect::<Vec<_>>();
-
-                                // Now we loop to check for more null bytes.
+                            // The read buffer should never contain null bytes at this point,
+                            // since they are always processed immediately. Therefore, we
+                            // only need to check the new data for null bytes.
+                            if !data.contains(&0) {
+                                // No null bytes in new data, just buffer it.
+                                xml_socket.read_buffer().extend(data);
```
kjarosh: You can move it out of the if so that it's not duplicated.

SuchAFuriousDeath: Done

PR #22553, opened 2025-12-20
avm2: Simplify instantiation of ArrayStorage

kjarosh:
ArrayStorage::from_args() can use from_iter() to prevent code duplication, and from_iter() doesn't have to count occupied entries, because all entries are occupied.


SuchAFuriousDeath:
I also noticed this, would have submitted it at some point.

PR #22555, opened 2025-12-20
refactor: Convert parsing logic to use FromWStr

SuchAFuriousDeath:
This should be reviewable commit by commit.

### DIFF
--- a/core/src/avm2/globals/flash/display/graphics.rs
+++ b/core/src/avm2/globals/flash/display/graphics.rs
@@ -1012,15 +1012,10 @@ pub fn draw_path<'gc>(
     let mut drawing = this.as_drawing().unwrap();
     let commands = args.get_object(activation, 0, "commands")?;
     let data = args.get_object(activation, 1, "data")?;
-    let winding = args.get_string(activation, 2);
-
-    let fill_rule = if winding == WStr::from_units(b"nonZero") {
-        FillRule::NonZero
-    } else if winding == WStr::from_units(b"evenOdd") {
-        FillRule::EvenOdd
-    } else {
-        return Err(make_error_2008(activation, "winding"));
-    };
+    let fill_rule = args
+        .get_string(activation, 2)
+        .parse()
+        .map_err(|_| make_error_2008(activation, "winding"))?;
 
     // FIXME - implement fill behavior described in the Flash docs
     // (which is different from just running each command sequentially on `Graphics`)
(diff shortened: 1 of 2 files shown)


swf/src/types.rs
```diff
@@ -422,28 +422,41 @@ impl Display for BlendMode {
     }
 }
 
-impl FromStr for BlendMode {
+impl FromWStr for BlendMode {
```
kjarosh: Is it the right place for such conversions? If I see correctly, they are not used in swf, but in AVM only

SuchAFuriousDeath: I think that's correct (they are used in avm only). I'll move them wherever you want, it's up to you. Maybe it's not the correct place.

kjarosh: I think it would be better to place them in core, but you won't be able to use `FromWStr`, but a regular function instead. 

For stuff that is used in AS code, let's put it in AVM1 globals when it's used only by AVM1, similarly AVM2. For stuff that is used for both, it would be best to test if AVM1/AVM2 representations are truly the same, if so... I don't know where to put them :D maybe in some common place, but according to my experience parsing usually differs between AVM1 and AVM2.

SuchAFuriousDeath: What is the FromWStr trait even for, then? Wouldn't it make more sense to have an avm1 version of it and an avm2 version? And then we can put different parsing logic in each implementation.

kjarosh: It's a wide version of `FromStr`, so it should be used the same way and have the same impl. In case it's not obvious how to parse a string to an object, `FromWStr`/`FromStr` probably shouldn't be used.

SuchAFuriousDeath: I get that it's an analogue of std FromStr, but the reservations from this review apply even to the already existing implementations of the trait, which kind of calls its existence into question imo. For example:
StageDisplayState has a FromWStr implementation:

https://github.com/ruffle-rs/ruffle/blob/e31d83463d25234cdbb652d02ae5e40ccf92d4ae/core/src/display_object/stage.rs#L1064-L1078

 but avm1 code does its own parsing 
 
https://github.com/ruffle-rs/ruffle/blob/e31d83463d25234cdbb652d02ae5e40ccf92d4ae/core/src/avm1/globals/stage.rs#L142-L152


Being able to call .parse() is really convenient (and analogous to std Rust) so it would be nice to somehow figure this out so that everything is in order.

Having an avm1 version of the trait and an avm2 version makes more sense to me. We would of course have to move the parse method from WStr to an avm1 WStrExt and an avm2 WStrExt because it couldn't be one method, but that actually makes sense if you think about it. If there are 2 sets of ways to parse a WStr into something, it's weird to only have 1 trait for that and 1 method :sweat_smile:

kjarosh: > Having an avm1 version of the trait and an avm2 version makes more sense to me

The trait is AVM-agnostic, it's from wstr not core.

> If there are 2 sets of ways to parse a WStr into something, it's weird to only have 1 trait for that and 1 method

Yeah, that's my point. If there are multiple ways to parse something, don't use `From[W]Str`. You can provide custom methods. What's the difference between calling `.parse::<Type>()` and `Type::from_string_avm1(...)`? The latter is clearer, the former is used when something has a trivial string repr. Providing multiple Traits for AVM1 and AVM2 will only make the call implicit instead of explicit and I'm not sure the abstraction is needed as we're parsing it manually anyway?

SuchAFuriousDeath: Well .parse() is definitely more ergonomic, because you can chain it with other methods so you don't need intermediate variables. 
But I understand, I'll cherry-pick this to  implement the trait only for avm1-exclusive or avm2-exclusive structs and leave the rest alone.

Lord-McSweeney (approved):

PR #22556, opened 2025-12-20
avm2: Fix occupied count miscalculations in `ArrayStorage`

kjarosh:
This patch fixes some instances where occupied_count wasn't updated properly for dense arrays and adds tests for all mutable methods.

### DIFF
--- a/core/src/avm2/array.rs
+++ b/core/src/avm2/array.rs
@@ -204,8 +204,8 @@ impl<'gc> ArrayStorage<'gc> {
                 occupied_count,
             } => {
                 if let Some(i) = storage.get_mut(item) {
-                    *i = None;
-                    if *occupied_count > 0 {
+                    if i.is_some() {
+                        *i = None;
                         *occupied_count -= 1;
                         self.maybe_convert_to_sparse();
                     }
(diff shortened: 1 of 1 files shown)


core/src/avm2/array.rs
```diff
@@ -241,6 +242,18 @@ impl<'gc> ArrayStorage<'gc> {
                 storage,
                 occupied_count,
             } => {
+                // Update occupied_count
+                if let Some(slice) = storage.get(size..) {
+                    if slice.len() <= storage.len() / 2 {
```
Lord-McSweeney, line 243: IMO, the purpose of this branch isn't immediately obvious when looking at the code. Could you add a comment for explanation?

core/src/avm2/array.rs
```diff
@@ -212,9 +212,10 @@ impl<'gc> ArrayStorage<'gc> {
                 occupied_count,
             } => {
                 if let Some(i) = storage.get_mut(item) {
-                    *i = None;
-                    if *occupied_count > 0 {
-                        *occupied_count -= 1;
+                    if i.is_some() {
+                        *i = None;
+                        debug_assert!(*occupied_count > 0);
+                        *occupied_count = occupied_count.saturating_sub(1);
```
Lord-McSweeney: If we already know that `occupied_count > 0`, doesn't it make more sense to just do `occupied_count -= 1` here?

kjarosh: The goal is to assert in debug mode (so we'll see the desync if it happens), but to work in release mode in order not to crash Ruffle needlessly.

core/src/avm2/array.rs
kjarosh: Done

core/src/avm2/array.rs
kjarosh: Changed to regular subtractions as per our offline discussion

PR #22557, opened 2025-12-20
desktop: Only attach to parent console on Windows if stdout is not a pipe or a regular file

### LINKED ISSUE #9145: Cannot redirect command-line output to file on Windows
### Describe the bug

When I run Ruffle from the command line on Windows, I cannot redirect its output to a file as [explained here](https://ss64.com/nt/syntax-redirection.html). For example, typing `ruffle.exe --version > version.txt` still prints the version in the Command Prompt and does not output it to the `version.txt` file.

### Expected behavior

In very old builds of Ruffle (early 2021), this functionality worked properly.

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

_No response_

SuchAFuriousDeath:
Fixes #9145. I tested this on my Windows backup laptop and it seems to work, but somebody please double check.

The original code caused the issue by attaching to a console unconditionally, which overrode the redirection to the file.

### DIFF
--- a/desktop/src/main.rs
+++ b/desktop/src/main.rs
@@ -18,6 +18,8 @@ mod preferences;
 #[cfg(feature = "tracy")]
 mod tracy;
 mod util;
+#[cfg(windows)]
+mod windows;
 
 use crate::preferences::GlobalPreferences;
 use anyhow::{Context, Error};
@@ -56,23 +58,6 @@ static RUFFLE_VERSION: &str = concat!(
     env!("VERGEN_GIT_COMMIT_DATE"),
     ")"
 );
-
-fn init() {
-    // When linked with the windows subsystem windows won't automatically attach
-    // to the console of the parent process, so we do it explicitly. This fails
-    // silently if the parent has no console.
-    #[cfg(windows)]
-    unsafe {
-        use windows_sys::Win32::System::Console::{ATTACH_PARENT_PROCESS, AttachConsole};
-        AttachConsole(ATTACH_PARENT_PROCESS);
-    }
-
-    let prev_hook = std::panic::take_hook();
-    std::panic::set_hook(Box::new(move |info| {
-        prev_hook(info);
-        panic_hook(info);
-    }));
-}
 
 fn panic_hook(info: &PanicHookInfo) {
     CALLSTACK.with(|callstack| {
@@ -133,16 +118,15 @@ fn panic_hook(info: &PanicHookInfo) {
     }
 }
 
-fn shutdown() {
-    // Without explicitly detaching the console cmd won't redraw it's prompt.
-    #[cfg(windows)]
-    unsafe {
-        windows_sys::Win32::System::Console::FreeConsole();
-    }
-}
-
 fn main() -> Result<(), Error> {
-    init();
+    let prev_hook = std::panic::take_hook();
+    std::panic::set_hook(Box::new(move |info| {
+        prev_hook(info);
+        panic_hook(info);
+    }));
+
+    #[cfg(windows)]
+    let _console = windows::Console::attach();
 
     let opt = Opt::parse();
     let preferences = GlobalPreferences::load(opt)?;
@@ -188,7 +172,7 @@ fn main() -> Result<(), Error> {
     if let Err(error) = &result {
         eprintln!("{:?}", error)
     }
-    shutdown();
+
     result
 }
 
--- a/desktop/src/windows.rs
+++ b/desktop/src/windows.rs
@@ -1,0 +1,56 @@
+use windows_sys::Win32::Storage::FileSystem::{FILE_TYPE_DISK, FILE_TYPE_PIPE, GetFileType};
+use windows_sys::Win32::System::Console::{
+    ATTACH_PARENT_PROCESS, AttachConsole, FreeConsole, GetStdHandle, STD_OUTPUT_HANDLE,
+};
+
+/// RAII guard for the attached parent console.
+/// Frees the console on drop if it was successfully attached.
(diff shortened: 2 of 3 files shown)


sombraguerrero:
This checks out per my testing. I'd say it's good to merge.

desktop/src/main.rs
```diff
@@ -61,10 +61,32 @@ fn init() {
     // When linked with the windows subsystem windows won't automatically attach
     // to the console of the parent process, so we do it explicitly. This fails
     // silently if the parent has no console.
+    //
+    // However, if stdout/stderr are already redirected (e.g., `ruffle.exe > file.txt`),
+    // we should NOT attach to the console as that would bypass the redirection.
+    // See: https://github.com/ruffle-rs/ruffle/issues/9145
     #[cfg(windows)]
     unsafe {
```
kjarosh: The scope of this `unsafe` block is quite large and encompasses several unsafe calls, it's better to use `unsafe` near calls and add safety comments if needed.

desktop/src/main.rs
```diff
@@ -61,10 +61,32 @@ fn init() {
...
+        let stdout_handle = GetStdHandle(STD_OUTPUT_HANDLE);
+        let file_type = GetFileType(stdout_handle);
+
+        match file_type {
+            // If output is redirected to a file or pipe, don't attach to console
+            // as that would bypass the redirection
+            FILE_TYPE_DISK | FILE_TYPE_PIPE => {}
+            // Otherwise, attach to parent console for interactive use
+            _ => {
+                AttachConsole(ATTACH_PARENT_PROCESS);
```
kjarosh: We're attaching console conditionally but freeing it unconditionally. Should it be this way?

desktop/src/main.rs
```diff
@@ -61,10 +61,32 @@ fn init() {
...
+        use windows_sys::Win32::Storage::FileSystem::{
+            FILE_TYPE_DISK, FILE_TYPE_PIPE, GetFileType,
+        };
+        use windows_sys::Win32::System::Console::{
+            ATTACH_PARENT_PROCESS, AttachConsole, GetStdHandle, STD_OUTPUT_HANDLE,
+        };
+
+        // Check if stdout is already redirected to a file or pipe
+        let stdout_handle = GetStdHandle(STD_OUTPUT_HANDLE);
+        let file_type = GetFileType(stdout_handle);
```
kjarosh: For instance, according to the Microsoft docs, I can see that safety guarantees may be broken here. `GetFileType` claims to accept a handle to the file, but `GetStdHandle`'s docs mentions it can return `INVALID_HANDLE_VALUE` or `NULL`.

desktop/src/main.rs
kjarosh: Maybe it's too much for this PR, but the way I see it we could use RAII for freeing the console and ideally we would store such (non-unsafe) structure in a separate Windows-specific file.

SuchAFuriousDeath: Sure, I can do it.

desktop/src/main.rs
SuchAFuriousDeath: Probably not. I did not know about the FreeConsole call. Good catch.

desktop/src/main.rs
SuchAFuriousDeath: AFAIK, the Rust bindings do not have any proper safety invariants.

And actually the Microsoft docs claim the following if you look carefully: 

https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfiletype

image

But maybe we should err on the side of caution anyway. Up to you.

desktop/src/main.rs
SuchAFuriousDeath: I cooked something up.

desktop/src/main.rs
```diff
@@ -188,7 +172,11 @@ fn main() -> Result<(), Error> {
     if let Err(error) = &result {
         eprintln!("{:?}", error)
     }
-    shutdown();
+
+    // Drop the console only after printing errors above
+    #[cfg(windows)]
+    drop(console);
```
kjarosh: The console should be dropped automatically, just name it `_console`.

SuchAFuriousDeath: I wanted to be a bit more explicit about it, but sure, no problem.

desktop/src/main.rs
kjarosh: Sure, that leaves only `NULL` to handle I guess

SuchAFuriousDeath: Unless you consider NULL to be included in 'invalid handle'. But I will check for NULL to be safe.

desktop/src/windows.rs
```diff
@@ -0,0 +1,58 @@
...
+    // However, if stdout/stderr are already redirected (e.g., `ruffle.exe > file.txt`),
+    // we should NOT attach to the console as that would bypass the redirection.
+    // See: https://github.com/ruffle-rs/ruffle/issues/9145
+    pub(super) fn attach() -> Self {
+        use windows_sys::Win32::Storage::FileSystem::{
+            FILE_TYPE_DISK, FILE_TYPE_PIPE, GetFileType,
+        };
+        use windows_sys::Win32::System::Console::{
+            ATTACH_PARENT_PROCESS, AttachConsole, GetStdHandle, STD_OUTPUT_HANDLE,
+        };
```
kjarosh: Now as the file is Windows-only you can move those to the very top.

SuchAFuriousDeath: Done

kjarosh (approved):

PR #22559, opened 2025-12-21
core: Reduce some logging

Lord-McSweeney:


core/src/library.rs
```diff
@@ -674,8 +670,6 @@ impl<'gc> Library<'gc> {
                 if let Ok(font) =
                     Font::from_font_file(gc_context, descriptor, data, index, FontType::Device)
                 {
-                    let name = font.descriptor().name().to_owned();
-                    info!("Loaded new device font \"{name}\" (bold: {is_bold}, italic: {is_italic}) from file");
```
kjarosh, line 678: Why are we dropping these logs?

n0samu: I think it's good to have, but maybe at the debug level instead of info since it can feel a little spammy

PR #22567, opened 2025-12-21
avm1: Reproduce the infamous "o is null" bug

kjarosh:
The variable "o" is being used in Flash's globals to make referring to symbols more concise.  However, it's not being deleted at the very end, but instead set to `null`, which means that in every SWF, the variable "o" is `null` and not `undefined`.


core/src/avm1/globals.rs
```diff
@@ -646,6 +646,12 @@ pub fn create_globals<'gc>(
         // TODO: addRequestHeader
         // TODO: clearRequestHeaders
 
+        // The variable "o" is being used in Flash's globals to make referring
+        // to symbols more concise.  However, it's not being deleted at the very
+        // end, but instead set to `null`, which means that in every SWF, the
+        // variable "o" is `null` and not `undefined`.
+        "o" => null();
```
kjarosh, line 653: Doesn't look great, probably `value(null)` would be better (`value(null; DONT_ENUM)` etc.), but it would also mean we should do `value(float(5.0))`, etc.

PR #22569, opened 2025-12-22
avm2: Fix behavior when loaded MovieClip throws error in constructor

Lord-McSweeney:
In Flash, when this happens, the MovieClip isn't added as a child to the `Loader`, the `init` and `complete` events aren't fired, and the `contentLoaderInfo` becomes inaccessible.


kjarosh:
Shouldn't we e.g. propagate the error and let the code handle it more "naturally" than setting a flag and just changing behavior here and there?

Lord-McSweeney:
@kjarosh The error appears in `construct_frame`, which we can't easily return out of. I don't really see any other way for loader code to know that there's been an error in the constructor?

core/src/loader.rs
```diff
@@ -2011,36 +2015,48 @@ impl<'gc> MovieLoader<'gc> {
         }
 
         if let MovieLoaderVMData::Avm2 { loader_info, .. } = vm_data {
+            let dobj = dobj.unwrap();
```
kjarosh, line 2018: Why are we doing an unwrap here?

kjarosh: Oh, nevermind, it got moved from below.

kjarosh (approved):
I still don't like the flag for this purpose, but I don't have a better idea so LGTM

PR #22573, opened 2025-12-22
render: Remove BorrowedOrOwnedTexture in favor of std::borrow::Cow

SuchAFuriousDeath:
wgpu::Texture derives Clone, so it implements ToOwned thanks to this blanket impl: https://doc.rust-lang.org/std/borrow/trait.ToOwned.html#impl-ToOwned-for-T, meaning it can be used in std::borrow::Cow.

### DIFF
--- a/render/wgpu/src/pixel_bender.rs
+++ b/render/wgpu/src/pixel_bender.rs
@@ -261,22 +261,6 @@ impl PixelBenderWgpuShader {
     }
 }
 
-enum BorrowedOrOwnedTexture<'a> {
-    Borrowed(&'a wgpu::Texture),
-    Owned(wgpu::Texture),
-}
-
-impl std::ops::Deref for BorrowedOrOwnedTexture<'_> {
-    type Target = wgpu::Texture;
-
-    fn deref(&self) -> &Self::Target {
-        match self {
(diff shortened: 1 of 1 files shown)


render/wgpu/src/pixel_bender.rs
```diff
@@ -468,13 +450,12 @@ pub(super) fn run_pixelbender_shader_impl(
...
             PixelBenderShaderArgument::ImageInput { index, texture, .. } => {
                 let input_texture = &image_input_as_texture(descriptors, texture.as_ref().unwrap());
-                let same_source_dest =
-                    if let BorrowedOrOwnedTexture::Borrowed(input_texture) = input_texture {
-                        std::ptr::eq(*input_texture, target)
-                    } else {
-                        // When we create a fresh texture, it can never be equal to the pre-existing target
-                        false
-                    };
+                let same_source_dest: bool = if let Cow::Borrowed(input_texture) = input_texture {
```
kjarosh: Can you drop the type decl for consistency?

SuchAFuriousDeath: I split this off while I was working on something else. not sure how exactly it snuck in :sweat_smile: Done.

PR #22576, opened 2025-12-22
avm1: Stub missing Camera properties

kjarosh:
Stubbed methods:
* Camera.get(),
* Camera.setMode(),
* Camera.setQuality(),
* Camera.setKeyFrameInterval(),
* Camera.setMotionLevel(),
* Camera.setLoopback(),
* Camera.setCursor().

Stubbed properties:
* Camera.names.

### DIFF
--- a/core/src/avm1/globals/camera.rs
+++ b/core/src/avm1/globals/camera.rs
@@ -4,22 +4,104 @@ use crate::avm1::activation::Activation;
 use crate::avm1::error::Error;
 use crate::avm1::property_decl::{DeclContext, StaticDeclarations, SystemClass};
 use crate::avm1::{Object, Value};
+use crate::avm1_stub;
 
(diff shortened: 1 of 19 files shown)


core/src/avm1/globals/camera.rs
```diff
@@ -4,22 +4,104 @@ use crate::avm1::activation::Activation;
...
+    Ok(Value::Null)
+}
+
+fn get_names<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    _this: Object<'gc>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    avm1_stub!(activation, "Camera", "names");
+    activation
```
Lord-McSweeney, line 49: Shouldn't this be returning an array?

kjarosh: Yes, it should

Issue #20394, opened 2025-05-12
duplcateMovieClip can't duplicate buttons


gzudi:
### Describe the bug

In AS 2 it was possible to duplicate a button using "duplicateMovieClip(myButton, newName, depth)"
In Ruffle you don't get a new button.

### Expected behavior

In AS 2 it was possible to duplicate a button (not a movie clip) using "duplicateMovieClip(myButton, newName, depth)"
In Ruffle you don't get a new button (however, it works fine for movie clips)

### Content Location

[Test08.zip](https://github.com/user-attachments/files/20160232/Test08.zip)

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Firefox

Issue #9213, opened 2023-01-19
Epic Battle Fantasy 3: Battle menu commands unselectable


waspennator:
### Describe the bug

https://www.newgrounds.com/portal/view/548041/format/flash?emulate=flash
Game starts up and gets into the hub world fine, but as soon as you get into a battle, you quickly get stuck since you can't select any battle options. Quickest way to get into a battle to test is to start a new game. skip through the dialogue until you get to the starting area, go up two screens until you get to this screen, then get near the green shrubbery looking monsters on the top right and press space next to em to initiate combat.

![EBF3 area](https://user-images.githubusercontent.com/78768499/213351004-7029f397-cc2c-4b62-9eb1-aedfdf246936.png)

### Expected behavior

You should be able to pick actions when in battles and not get softlocked.

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Edge 109

waspennator:
[ruffle.rs-1676770188233.log](https://github.com/ruffle-rs/ruffle/files/10775508/ruffle.rs-1676770188233.log)
Sharing a console log running through the ruffle web demo site if that'll help.

waspennator:
Still seems to be an issue.

waspennator:
Still happening

waspennator:
Still present on latest nightly.

Molisson:
I did some research on this issue via modding the SWF in JPEXS, it turns out that duplicating a button causes issue and messes up the bounding box and messing up the hitTest for mouse events.

I modified the SWF to edit `Attack` and `Tactics` text to show `button0._width` and `button1._width` on chid 6539 aka `menuOptions`, where the code for battle options is stored.

Flash:
<img width="691" height="398" alt="Image" src="https://github.com/user-attachments/assets/51f8ceff-e23b-40e9-a28c-9cfb6cb737ee" />

Ruffle:
<img width="694" height="401" alt="Image" src="https://github.com/user-attachments/assets/f33a2f9a-35ca-47be-b7ba-cd6c1f7140c9" />

Molisson:
https://github.com/ruffle-rs/ruffle/issues/20394
This seems related to this issue.

Issue #22585, opened 2025-12-23
MEP Flash Exos | No validation button on the interface


web62-MEP:
### Describe the bug

Hello
My [website](https://intramaths.pages-perso.free.fr/mep_flash_exos/index.php) brings back to life more than 1500 math exercises. About 80% are ok, congratulation for that ;-)
This one (from a long series) doesn't work --> [EXAMPLE](https://intramaths.pages-perso.free.fr/mep_flash_exos/3eme/geometrie/chap1/serie3/exo5/index.php?liste) , non validation button ! You've got the file source G1s3ex5.swf on the top of the window to test with FlashPlayer.
If you cant fix it it's good.
Thanks for advance

### Expected behavior

Fix it in a new update

### Content Location

https://intramaths.pages-perso.free.fr/mep_flash_exos/3eme/geometrie/chap1/serie3/exo5/index.php?liste

### Affected platform

Self-hosted version

### Operating system

W10 & W11

### Browser

All browsers

kjarosh:
The root cause is that Ruffle does not support `CloneSprite` with anything other than a MovieClip, but it looks like it should work with buttons too.

web62-MEP:
Yes, that one is working perfectly (button visible) --> https://intramaths.pages-perso.free.fr/mep_flash_exos/6eme/numerique/chap1/serie1/exo7/index.php
Why does the button is not visible in the first example ?

kjarosh:
The second example doesn't use `CloneSprite` with buttons.

web62-MEP:
Ah ok.
Will Ruffle ever be compatible with CloneSprite or not ?

kjarosh:
Yes, that's the point of these issues, we track bugs and unimplemented stuff so that someone can take care of it if they want.

web62-MEP:
Great, i'm waiting so...
Nevertheless, congratulations for the enormous job of ruffle and thank you for you're answers.
See you

kjarosh:
After fixing `CloneSprite`, the button doesn't work because somehow right after calling `loadMovieNum(..., 100, ...)` Ruffle thinks `_level100._visible` is `true` (even if the movie doesn't get loaded at all).

web62-MEP:
So, it's not fix or will it be ok with the nightly update ?

kjarosh:
The button will appear after merging https://github.com/ruffle-rs/ruffle/pull/22588 but it won't do anything when clicked.

[...]

PR #22588, opened 2025-12-23
avm1: Make `CloneSprite` work with all display objects

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/9213
* Fixes https://github.com/ruffle-rs/ruffle/issues/20394
* Progresses https://github.com/ruffle-rs/ruffle/issues/22585

It can be inferred from the docs that `duplicateMovieClip()` works only with MovieClips, but that's not the case.  It works with all display objects (including buttons and text).

`avm1/clone_sprite_types` is a known failure due to a discrepancy in placing images on the timeline.


core/src/avm1/globals/movie_clip.rs
```diff
@@ -962,39 +973,56 @@ pub fn clone_sprite<'gc>(
...
-            .instantiate_by_id(movie_clip.id(), context.gc())
-            .unwrap()
-            .as_movie_clip()
+            .instantiate_by_id(sprite.id(), context.gc())
             .unwrap()
+    } else if sprite.as_movie_clip().is_some() {
+        // Dynamically created MovieClip; create a new empty movie clip.
+        MovieClip::new(movie, context.gc()).as_displayobject()
+    } else if let Some(et) = sprite.as_edit_text() {
+        // Dynamically created TextField; create a new text field.
```
Lord-McSweeney, line 986: nit: Can you mention that this copies the size of the text field?

core/src/avm1/globals/movie_clip.rs
```diff
@@ -962,39 +973,56 @@ pub fn clone_sprite<'gc>(
...
+        EditText::new(
+            context,
+            movie,
+            et.x().to_pixels(),
+            et.y().to_pixels(),
+            et.width(),
+            et.height(),
```
Lord-McSweeney, line 995: Does this not copy the text contents of the dynamically created text field over to the new `EditText`? Could you add a test for the behavior?

core/src/avm1/globals/movie_clip.rs
```diff
@@ -962,39 +973,56 @@ pub fn clone_sprite<'gc>(
...
+        )
+        .as_displayobject()
     } else {
```
Lord-McSweeney, line 998: Do buttons not get any properties copied?

kjarosh: That's a piece of code that instantiates objects that were not instantiated by a symbol, and I don't think you can instantiate a button dynamically in AS2?

core/src/avm1/globals/movie_clip.rs
kjarosh: So I haven't tested it yet, the main focus for this PR was the library instantiation. I can add a TODO or a panic here.

Lord-McSweeney: A TODO here sounds fine to me

Lord-McSweeney (approved):

core/src/avm1/globals/movie_clip.rs
kjarosh: Added a todo

core/src/avm1/globals/movie_clip.rs
kjarosh: Added a todo

Issue #22589, opened 2025-12-23
Chrome extension started blocking web sources.


PretentiousJackal:
### Describe the bug

The ruffle emulator is blocking a bunch of sources and files.
Chrome extension off:
<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/22b2174a-2646-49fe-8420-53193685f8a4" />
<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/686fd5f3-eb46-4e10-a19d-c774d1b00cb7" />
Chrome extension on:
<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/a3d88cc2-f815-4d67-8110-20c11a99a886" />
<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/ac093af9-6553-49ed-89b0-bfb93f6598b8" />

### Expected behavior

I expected images, videos, audio files and css elements to load properly. But with the extension on, websites became skeletons of their former selves. And it became impossible to use certain websites like Bluesky, YouTube and Discord.

### Content Location

It's not flash content. The problem is that ruffle is affecting everything regardless of it's flash, in a bad way.

### Affected platform

Browser's extension

### Operating system

ChromeOS

### Browser

Google Chrome

### Additional information

I have a chromebook with the system Alan version 126.15886.0. My chromeOS version is Version 126.0.6478.222 (Official Build) (64-bit). I'm not sure if that's the same thing as the Google Chrome browser version. My chrome extension of ruffle version is 0.2.0-nightly.2025.12.23. The problem started in the morning of December 22nd (In a timezone of the Western Hemisphere).

kjarosh:
Have you tried disabling other extensions and seeing if Ruffle works properly?

danielhjacobs:
Can you check chrome://version in your address bar? The first line of that page should say the Chromium version. I just installed Chrome 126 in case you're using that but can't reproduce the issue there.

danielhjacobs:
I will suggest, if you are able to, either updating your browser and/or uninstalling and re-installing Ruffle.

PretentiousJackal:
I keep forgetting to come back here, and say: I thought I had turned off other extensions and that I was just testing ruffle's effects but it turned out I was just unpinning other extensions, which interacted with ruffle, and not actually turning them off. I feel stupid.

PR #22592, opened 2025-12-23
extension: Make it if cleanup fails RuleCondition test does nothing

danielhjacobs:
This section of code tests if the responseHeaders RuleCondition is supported and not behind a flag, but if the RuleCondition is enabled behind a flag it will successfully register a rule which matches any request, even if it doesn't match the unsupported RuleCondition. What that means is on Chromium 121-127 a rule is registered via lines 37 - 50 that blocks every request. It shouldn't matter though, as that rule is removed by lines 57 - 59, but just in case that fails to remove the rule, it's better if the rule does nothing.

### DIFF
--- a/web/packages/extension/src/background.ts
+++ b/web/packages/extension/src/background.ts
@@ -22,7 +22,7 @@ async function isHeaderConditionSupported() {
                     action: {
                         type:
                             chrome.declarativeNetRequest.RuleActionType
-                                ?.BLOCK ?? "block",
+                                ?.ALLOW ?? "allow",
                     },
                 },
             ],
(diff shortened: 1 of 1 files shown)


danielhjacobs:
This last cropped up in #20045 due, ultimately, to a webpack bug. It is possible, though currently unknown, that it may be the cause of #22589 too.

kjarosh:
Is it somehow possible that we need cleanup even if `needsCleanup` is false?

danielhjacobs:
We shouldn't i don't think as it's only false if the rule registering throws in both trys (which should mean no rules were registered).

danielhjacobs:
The code was suggested by Rob--W of the Mozilla extension team: https://github.com/w3c/webextensions/issues/638#issuecomment-2181124486

torokati44:
(I just don't understand the PR title / commit message ... is there a word missing?)

danielhjacobs:
Make it (so) if cleanup fails (the) RuleCondition test (still) does nothing

Issue #16081, opened 2024-04-22
Error on https://www.missionjuno.swri.edu/spacecraft/juno-spacecraft


MrOrangeTan:
# Error Info
Error name: Error
Error message: panicked at core/src/avm2/globals/flash/display/shader_job.rs:151:29:
assertion `left == right` failed
  left: 732
 right: 672
Error stack:
```
Error: panicked at core/src/avm2/globals/flash/display/shader_job.rs:151:29:
assertion `left == right` failed
  left: 732
 right: 672
    at imports.wbg.__wbg_new_28c511d9baebfa89 (chrome-extension://donbcfbmhbcapadipfkeojnmajbakjdc/dist/69.js:3393:21)
    at js_sys::Error::new::hf655a9c3d4762b78 (wasm://wasm/03a8f47e:wasm-function[13925]:0xb2fe7c)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::h90bd631d8898e3d0 (wasm://wasm/03a8f47e:wasm-function[11515]:0xad978d)
    at std::panicking::rust_panic_with_hook::h9aabd906218897c3 (wasm://wasm/03a8f47e:wasm-function[7090]:0x9a0a6c)
    at core::panicking::panic_fmt::h3aff855fe938c13f (wasm://wasm/03a8f47e:wasm-function[11232]:0xad3277)
    at core::panicking::assert_failed_inner::hc4f70f3da5729a72 (wasm://wasm/03a8f47e:wasm-function[4937]:0x88ab7f)
    at core::panicking::assert_failed::h6e0fc2a8021afd31 (wasm://wasm/03a8f47e:wasm-function[11272]:0xad46cd)
    at <core::iter::adapters::flatten::FlatMap<I,U,F> as core::iter::traits::iterator::Iterator>::next::hc7f713e498acf3a0 (wasm://wasm/03a8f47e:wasm-function[5387]:0x8f8528)
    at ruffle_core::avm2::globals::flash::display::shader_job::get_shader_args::he1d6296052f6f88b (wasm://wasm/03a8f47e:wasm-function[3437]:0x75e925)
    at ruffle_core::avm2::globals::flash::display::shader_job::start::h170cc3670c528456 (wasm://wasm/03a8f47e:wasm-function[1356]:0x4320ee)
```
AVM2 stack:
```
    at flash.display::ShaderJob/flash::display::ShaderJob::start()
[...]

PR #22606, opened 2025-12-25
pixelbender: Improve shader argument validation

SuchAFuriousDeath:
Progresses #14518 and maybe #16081 (can't reproduce, but the assert values and the panic location suggests that this is the same issue).

TLDR:
The asserts for Vector and Bytearray size were wrong. Flash actually allows bigger lengths than what is necessary and for smaller ones, throws Error 2165.

This PR only focuses on inputs, pixelbender outputs also have some issues (which is what #14518 is stuck on after this) but I don't want bloat this PR, there is another one coming.

The FloatPixelData change makes the code more robust imo, using a plain Vec\<u8\> allowed partial pixels or even partial floats, now that can't happen. 

The ByteArray endianness assert should be removed as well, but that will need tests and again, I don't want to bloat this PR.


tests/tests/swfs/avm2/pixelbender_bytearray_input/Test.as
```diff
@@ -0,0 +1,82 @@
...
+            var exactFloats:int = WIDTH * HEIGHT * FLOATS_PER_PIXEL;
+            var extraFloats:int = exactFloats + 8; // 8 extra floats
+            var input:ByteArray = createInputByteArray(extraFloats);
+            runShader(input, "Too big (" + extraFloats + " floats, " + (extraFloats - exactFloats) + " extra)");
+        }
+
+        private function testTooSmall():void {
+            // Fewer floats than needed - should throw Error #2165
+            var exactFloats:int = WIDTH * HEIGHT * FLOATS_PER_PIXEL;
+            var tooFewFloats:int = exactFloats - 4; // 1 pixel short (4 channels)
```
kjarosh: What about 1 channel short?

tests/tests/swfs/avm2/pixelbender_bytearray_input/Test.as
```diff
@@ -0,0 +1,82 @@
...
+            // 2x2 pixels, 4 channels each = 16 floats = 64 bytes
+            var exactFloats:int = WIDTH * HEIGHT * FLOATS_PER_PIXEL;
+            var input:ByteArray = createInputByteArray(exactFloats);
+            runShader(input, "Exact size (" + exactFloats + " floats)");
+        }
+
+        private function testTooBig():void {
+            // More floats than needed - should still work
+            var exactFloats:int = WIDTH * HEIGHT * FLOATS_PER_PIXEL;
+            var extraFloats:int = exactFloats + 8; // 8 extra floats
```
kjarosh: What about 1 extra float?

tests/tests/swfs/avm2/pixelbender_vector_input/Test.as
```diff
@@ -0,0 +1,77 @@
+package {
```
kjarosh: Can we merge those 2 tests into 1? I see a lot of repetition

render/wgpu/src/pixel_bender.rs
```diff
@@ -287,33 +289,40 @@ fn image_input_as_texture<'a>(
...
-            // use the first 3 channels (since it was compiled with a 3-channel input),
-            // so it doesn't matter what value we choose here.
-            let padded_bytes = if *channels == 3 {
-                let mut padded_bytes = Vec::with_capacity(bytes.len() * 4 / 3);
-                for chunk in bytes.chunks_exact(12) {
-                    padded_bytes.extend_from_slice(chunk);
-                    padded_bytes.extend_from_slice(&[0, 0, 0, 0]);
+
+            let (channels, padded) = match data {
+                FloatPixelData::R(r) => (1, Cow::Borrowed(r.as_flattened())),
```
kjarosh: Can we cover different numbers of channels in tests? (1,2,3,4)

SuchAFuriousDeath:
On it :saluting_face:

SuchAFuriousDeath:
This should be ready now.

core/src/avm2/globals/flash/display/shader_job.rs
```diff
@@ -134,27 +136,65 @@ pub fn get_shader_args<'gc>(
...
                         } else if let Some(byte_array) = input.as_bytearray() {
-                            let expected_len = (width * height * input_channels) as usize
-                                * std::mem::size_of::<f32>();
-                            assert_eq!(byte_array.len(), expected_len);
                             assert_eq!(byte_array.endian(), Endian::Little);
-                            ImageInputTexture::Bytes {
+
+                            let num_pixels = (width * height) as usize;
+
+                            let (bytes, _trailing) = byte_array.bytes().as_chunks::<4>();
```
kjarosh: You can use `_` instead of `_trailing` to let the compiler know the value can be dropped right away.

core/src/avm2/globals/flash/display/shader_job.rs
```diff
@@ -134,27 +136,65 @@ pub fn get_shader_args<'gc>(
...
+                            };
+
+                            ImageInputTexture::Floats {
                                 width,
                                 height,
-                                channels: input_channels,
-                                bytes: byte_array.read_at(0, byte_array.len()).unwrap().to_vec(),
+                                data,
                             }
                         } else if let Some(vector) = input.as_vector_storage() {
```
kjarosh, line 152: Can't we somehow merge this branch with bytearray's one? (i.e. remap one to the other)

core/src/avm2/globals/flash/display/shader_job.rs
SuchAFuriousDeath: I wanted to provide a better hint for people reading the code as to what the other return value is, but sure, no problem.

core/src/avm2/globals/flash/display/shader_job.rs
SuchAFuriousDeath: I'll do my best.

SuchAFuriousDeath: Ok I tried something, I think this is the best we can do.

core/src/avm2/globals/flash/display/shader_job.rs
```diff
@@ -178,6 +172,52 @@ pub fn get_shader_args<'gc>(
...
+        )
+    }
+}
+
+fn make_float_texture<'gc, S: PixelSource>(
+    source: S,
+    width: u32,
+    height: u32,
+    input_channels: u32,
+    mut err: impl FnMut() -> Error<'gc>,
```
kjarosh: It's better to pass required arguments (activation, name) instead of a lambda, because (1) the error is the same in both cases, so no duplication, (2) if someone needs activation in the future for something here, it wouldn't be possible to borrow it twice.

It's fine to define `err` in this function.

render/wgpu/src/pixel_bender.rs
```diff
@@ -287,33 +289,40 @@ fn image_input_as_texture<'a>(
...
                 }
-                Cow::Owned(padded_bytes)
-            } else {
-                Cow::Borrowed(bytes)
+                FloatPixelData::Rgba(rgba) => (4, Cow::Borrowed(rgba.as_flattened())),
             };
 
+            let padded_bytes = bytemuck::cast_slice::<f32, u8>(padded.as_ref());
+
+            let texture_format = temporary_texture_format_for_channels(channels);
```
kjarosh: Wouldn't it make more sense to decide on texture format directly instead of mapping to channels and then to the format? This way the byte padding and format selection would be near each other instead of in two completely different places.

render/wgpu/src/pixel_bender.rs
```diff
@@ -287,33 +289,40 @@ fn image_input_as_texture<'a>(
...
-            // with zeros for the alpha channel. The PixelBender code will only ever try to
-            // use the first 3 channels (since it was compiled with a 3-channel input),
-            // so it doesn't matter what value we choose here.
-            let padded_bytes = if *channels == 3 {
-                let mut padded_bytes = Vec::with_capacity(bytes.len() * 4 / 3);
-                for chunk in bytes.chunks_exact(12) {
-                    padded_bytes.extend_from_slice(chunk);
-                    padded_bytes.extend_from_slice(&[0, 0, 0, 0]);
+
+            let (channels, padded) = match data {
```
kjarosh: Can we move it to `FloatPixelData`'s impl?

render/wgpu/src/pixel_bender.rs
SuchAFuriousDeath: It would, but then the format selection logic would be in 2 places temporary_texture_format_for_channels is also used elsewhere.

SuchAFuriousDeath: I did it anyway.

kjarosh:
Before merging, can you describe the changes in the first commit message?

SuchAFuriousDeath:
:saluting_face:

kjarosh:
Perfect, thank you!

PR #22610, opened 2025-12-26
chore: Migrate core to edition 2024

SuchAFuriousDeath:
I disabled import reordering in the core crate through core/rustfmt.toml. Without it, the diff would be ridiculous and this PR would be too overwhelming. Most of this is still style changes though.

As discussed on Discord, there are potential issues with https://doc.rust-lang.org/edition-guide/rust-2024/temporary-tail-expr-scope.html. Running 'RUSTFLAGS="-W tail_expr_drop_order" cargo check --all-features 2>&1' should find problematic code locations (According to [the migration section](https://doc.rust-lang.org/edition-guide/rust-2024/temporary-tail-expr-scope.html#migration), if I understand correctly). It found only 2 spots and those were harmless. They constitute the changes in  core/src/avm2/globals/__ruffle__.rs.

Also, some of the cargo fmt style choices seem ... questionable :sweat_smile: We might want to tweak that maybe.


SuchAFuriousDeath:
Interesting, clippy had no issues locally for me. Maybe we can temporarily disable the collapsible_if lint only for the core crate? And then reenable in a separate PR.

core/src/avm2/regexp.rs
```diff
@@ -361,7 +361,7 @@ impl<'gc> RegExp<'gc> {
                 storage.push(context.strings.substring(text, c.clone()).into());
                 if storage.length() >= limit {
                     break; // Intentional bug to match Flash.
-                           // Causes adding parts past limit.
+                    // Causes adding parts past limit.
```
Lord-McSweeney, line 364: nit: Could you move these two comments before the `break;`? The way it is right now they look like they refer to different things

core/src/avm2/error.rs
```diff
@@ -152,12 +152,12 @@ pub fn make_reference_error<'gc>(
...
 
     let msg = match code {
-        ReferenceErrorCode::AssignToMethod => format!(
-            "Error #1037: Cannot assign to a method {qualified_name} on {class_name}.",
-        ),
-        ReferenceErrorCode::InvalidWrite => format!(
-            "Error #1056: Cannot create property {qualified_name} on {class_name}.",
-        ),
+        ReferenceErrorCode::AssignToMethod => {
+            format!("Error #1037: Cannot assign to a method {qualified_name} on {class_name}.",)
```
Lord-McSweeney: Could you remove the extra commas after the quotes for the three cases that the new formatting changed (1037, 1056, and 1120)?

kjarosh:
> Interesting, clippy had no issues locally for me. Maybe we can temporarily disable the collapsible_if lint only for the core crate? And then reenable in a separate PR.

Lets do that. If possible maybe make it a warning that doesn't fail the build? I'm not sure if that's possible

core/src/avm1/globals/date.rs
```diff
@@ -10,11 +10,7 @@ use std::fmt;
 #[inline]
 fn rem_euclid_i32(lhs: f64, rhs: i32) -> i32 {
     let result = (lhs % f64::from(rhs)).clamp_to_i32();
-    if result < 0 {
```
kjarosh, line 13: Can we change formatting rules so that the old formatting is preferred in this case?

I'm not sure if that's a positive change, and if it is we can do it in a separate PR

SuchAFuriousDeath:
> > Interesting, clippy had no issues locally for me. Maybe we can temporarily disable the collapsible_if lint only for the core crate? And then reenable in a separate PR.
> 
> Lets do that. If possible maybe make it a warning that doesn't fail the build? I'm not sure if that's possible

I'm afraid it's not, the pipelines have -D warnings, so we have to use allow.

core/src/avm1/globals/date.rs
kjarosh: I meant this rule in general, not only this place

SuchAFuriousDeath: Oh :sweat_smile:

SuchAFuriousDeath: Ok I got it, turns out that style editions are a thing, so I just set it to 2021 in core/rustfmt.toml

core/src/lib.rs
```diff
@@ -1,6 +1,8 @@
 // This lint is helpful, but right now we have too many instances of it.
 // TODO: Remove this once all instances are fixed.
 #![allow(clippy::needless_pass_by_ref_mut)]
+// Temporarily allow this to ease migration to Rust 2024 edition.
```
kjarosh, line 4: Please add a TODO to remove it

core/src/socket.rs
```diff
@@ -346,7 +347,7 @@ impl<'gc> Sockets<'gc> {
 
                                     // Check if we have another null byte in the same payload.
                                     if let Some((index, _)) =
-                                        data.iter().enumerate().find(|(_, &b)| b == 0)
+                                        data.iter().enumerate().find(|&(_, &b)| b == 0)
```
kjarosh: Why do we have to add the ampersand here?

core/rustfmt.toml
```diff
@@ -0,0 +1 @@
+style_edition = "2021" # This is temporary to ease transitioning the core crate to edition 2024
```
kjarosh: Nit: put the comment above the line

core/src/html/text_format.rs
```diff
@@ -1717,7 +1717,11 @@ impl<'a> FormatState<'a> {
                 font.color.g,
                 font.color.b,
                 font.letter_spacing,
-                if font.kerning { "1" } else { "0" },
```
kjarosh, line 1734: I think this can be reverted?

core/src/socket.rs
SuchAFuriousDeath: I didn't do this manually, it must have been clippy. Now I changed the two occurrences to use .position() so the code makes more sense now.

SuchAFuriousDeath:
Should this be marked as squash on merge btw?

n0samu:
> Should this be marked as squash on merge btw?

Maybe the PR used to have multiple commits? I removed the tag now since there's just one commit (nothing to squash)

PR #22612, opened 2025-12-26
avm1: implemented getBytesLoaded and getBytesTotal

Flawake:


core/src/avm1/globals/sound.rs
```diff
@@ -71,9 +70,12 @@ struct SoundData<'gc> {
     duration: Cell<Option<u32>>,
 
     /// Whether this sound is an external streaming MP3.
-    /// This will be true if `Sound.loadSound` was called with `isStreaming` of `true`.
+    /// This will be true if `Sound.loadSound` was called with `isStreaming` on `true`.
```
kjarosh: Why "on"?

kjarosh:
> While testing I found that get_bytes_total actually is 64 bit instead of 32. But I was unable to fix that.

How did you find that? Have you tried playing a 5 GB sound?

core/src/avm1/globals/sound.rs
Flawake: Just felt better. But my English is not that great, I will change it back to of when I resolve the conflicts.

Flawake:
> > While testing I found that get_bytes_total actually is 64 bit instead of 32. But I was unable to fix that.
> 
> How did you find that? Have you tried playing a 5 GB sound?

Yes, that's exactly how I did it.
I created a 5gb mp3 using this ffmpeg command:
```
ffmpeg -f lavfi -i anoisesrc=color=white:sample_rate=44100 -t 134218 -c:a libmp3lame -b:a 320k noise.mp3
```

kjarosh:
You can add a TODO for the u64 total bytes stuff.

Flawake:
I was sleeping I think. My u64 should have been a u32 instead of the other way around. Has been changed in the last commit

core/src/avm1/globals/sound.rs
```diff
@@ -436,20 +448,33 @@ fn set_duration<'gc>(
 
 fn get_bytes_loaded<'gc>(
     activation: &mut Activation<'_, 'gc>,
-    _this: Object<'gc>,
+    this: Object<'gc>,
     _args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
-    avm1_stub!(activation, "Sound", "getBytesLoaded");
-    Ok(1.into())
+    if activation.swf_version() >= 6 {
```
kjarosh: This is VERSION_6 so this if does nothing?

core/src/avm1/globals/sound.rs
```diff
@@ -436,20 +448,33 @@ fn set_duration<'gc>(
...
 
 fn get_bytes_total<'gc>(
     activation: &mut Activation<'_, 'gc>,
-    _this: Object<'gc>,
+    this: Object<'gc>,
     _args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
-    avm1_stub!(activation, "Sound", "getBytesTotal");
-    Ok(1.into())
+    if activation.swf_version() >= 6 {
```
kjarosh: This is VERSION_6 so this if does nothing?

core/src/avm1/globals/sound.rs
```diff
@@ -436,20 +448,33 @@ fn set_duration<'gc>(
...
+        if let NativeObject::Sound(sound) = this.native() {
+            if let Some(sound_handle) = sound.sound() {
+                if let Some(length) = activation.context.audio.get_sound_size(sound_handle) {
```
kjarosh: If you invert those ifs you can reduce indentation

kjarosh:
> I was sleeping I think. My u64 should have been a u32 instead of the other way around. Has been changed in the last commit

So it shouldn't be 64-bit?

core/src/avm1/globals/sound.rs
```diff
@@ -436,20 +448,33 @@ fn set_duration<'gc>(
...
+                    return Ok(length.into());
+                }
+            }
+            return Ok(1.into());
```
kjarosh: Why does it return 1 here?

core/src/avm1/globals/sound.rs
Flawake: I reduced the indentation by one in the latest commit

Flawake:
> > I was sleeping I think. My u64 should have been a u32 instead of the other way around. Has been changed in the last commit
> 
> So it shouldn't be 64-bit?

No, should be 32 bit. I was sleeping or something I think

frontend-utils/src/backends/navigator.rs
```diff
@@ -469,6 +469,25 @@ impl<F: FutureSpawner<Error> + 'static, I: NavigatorInterface> NavigatorBackend
 
         tokio::spawn(future);
     }
+
+    /// Estimate the length of a local file for a given request.
+    fn estimate_file_length(&self, request: &Request) -> Option<u32> {
```
kjarosh, line 474: Why not use `SuccessResponse::expected_length()`?

Flawake: Because that one is only available from a future and is not guaranteed to be ready.
```rust
    Box::pin(async move {
        let fetch = player.lock().unwrap().fetch(request, FetchReason::Other);
        let mut response = fetch.await.map_err(|error| error.error)?;
...
```
from core->src->loader.rs line 1273

response is inside the async block.

While flash always has the length ready on local files.

```as2
s.loadsound("local_filesystem_file.mp3", b_streaming);
trace(s.getBytesTotal());
```

always immediatly returns the amount of bytes in flash with local files. Ruffle is not guaranteed to do so in the async block with the response.

Or is it possible to get the SuccessResponse::expected_length() outside of the async block?

kjarosh: What if the file is not local? Does it block until we get the content length header?

Flawake: No, It returns undefined on a non local file. In flash, and in Ruffle with this commit

Flawake: I have not added a file from a http request in the tests, because I am not sure if we can ensure that it will always be undefined, networking etc... Can we do that without problems?

kjarosh: > Can we do that without problems?

I'm not sure how our tests work in this regard, there might be a race condition.

If it returns undefined for remote files, it's fine to return undefined for local files too in this iteration.

Flawake: > If it returns undefined for remote files, it's fine to return undefined for local files too in this iteration.

But that's not what the flash player does. The player always returns the size for local files

kjarosh: If you want to return file size synchronously, you'd need to refactor how fetching is done in order to allow supplying this information synchronously. Currently it's not possible because a future is returned that requires an await.

There's also the question of whether it's really done synchronously, or asynchronously but Flash manages to do it before reading the value in AS. We'd need to simulate a delay to reading a file and check whether something blocks in Flash or `undefined` is returned. I'm not fully convinced it's done synchronously, because Flash was mainly a web technology and it just doesn't make sense to implement it differently for local files compared to remote files.

I feel like the effort involved outweighs the benefits we get from it, that's why I think it's fine to return `undefined` for now.

PR #22617, opened 2025-12-27
Revert "desktop: Only attach to parent console on Windows if stdout is not a pipe or a regular file"

kjarosh:
Reverts ruffle-rs/ruffle#22557

This causes issues with console output on Windows in some cases.


kjarosh:
Superseded by https://github.com/ruffle-rs/ruffle/pull/22618

PR #22618, opened 2025-12-27
desktop: Attach console when stdout handle is null

SuchAFuriousDeath:
Supersedes #22617.

#22557 caused console output to not work when GetStdHandle returned null.

### DIFF
--- a/desktop/src/windows.rs
+++ b/desktop/src/windows.rs
@@ -1,3 +1,5 @@
+use std::ptr::NonNull;
+
 use windows_sys::Win32::Storage::FileSystem::{FILE_TYPE_DISK, FILE_TYPE_PIPE, GetFileType};
 use windows_sys::Win32::System::Console::{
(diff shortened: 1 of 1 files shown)


sombraguerrero:
Tested both console output and file redirection with cmd and conhost explicitly set.
Working as expected.

desktop/src/windows.rs
```diff
@@ -20,25 +22,21 @@ impl Console {
...
         // Check if stdout is already redirected to a file or pipe
         // SAFETY: STD_OUTPUT_HANDLE is a valid standard device constant.
-        let stdout_handle = unsafe { GetStdHandle(STD_OUTPUT_HANDLE) };
+        let stdout_handle = NonNull::new(unsafe { GetStdHandle(STD_OUTPUT_HANDLE) });
 
-        // GetStdHandle can return NULL on failure
-        let attached = if stdout_handle.is_null() {
-            false
-        } else {
+        let attached = if stdout_handle.is_none_or(|mut handle| {
```
kjarosh: Can you extract this expression to a variable `should_attach`?

PR #22621, opened 2025-12-28
desktop: Add GUI for opening a directory

kjarosh:
* Progresses https://github.com/flathub/rs.ruffle.Ruffle/issues/38

There's two main reasons why we want to add an option to open a directory.

1. Not archived Ruffle bundles can be opened easily.
2. We can open multi-asset content on platforms where access to the filesystem is restricted (Linux, MacOS).

This PR makes it so that, when opening a directory, it tries to detect the content type. First, it checks for a Ruffle bundle, then it checks if there's only one SWF, if that's not the case, a second dialog will be shown to pick the root SWF from the directory.

When merged, this takes us one step closer to enabling filesystem isolation in Flatpak so that Ruffle (and SWFs) won't need to have access to the whole filesystem.

| Before | After |
| ------ | ----- |
| image | image |
| image | image |

When opening a directory with multiple SWF files, the following dialog will be shown.

image

### DIFF
--- a/desktop/src/gui/dialogs/select_path_dialog.rs
+++ b/desktop/src/gui/dialogs/select_path_dialog.rs
@@ -1,7 +1,8 @@
 use crate::gui::{LocalizableText, text};
 use egui::{Align2, Ui, Window};
+use std::cmp::Ordering;
 use std::ffi::OsStr;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use tokio::sync::oneshot::Sender;
 use unic_langid::LanguageIdentifier;
 use walkdir::WalkDir;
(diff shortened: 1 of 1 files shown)


Tiagoquix:
This is very interesting. Impressive work!

FYI, there's this too: https://github.com/flathub/org.torproject.torbrowser-launcher/commit/c40186b9341f7724ad476f42dc2c67bbd7cc7abd
(Tor Browser Flatpak version removed filesystem perm to use portal)

Not sure if it would work with Ruffle, I have not tested it yet.

kjarosh:
> FYI, there's this too: [flathub/org.torproject.torbrowser-launcher@c40186b](https://github.com/flathub/org.torproject.torbrowser-launcher/commit/c40186b9341f7724ad476f42dc2c67bbd7cc7abd) (Tor Browser Flatpak version removed filesystem perm to use portal)
> 
> Not sure if it would work with Ruffle, I have not tested it yet.

That's what Ruffle is using. It's not an issue of not using portals, but loading arbitrary files, that weren't chosen by the user.

After this PR, the user will have a way of choosing a directory with all required files.

PR #22635, opened 2025-12-29
avm1: Clone text fields properly

kjarosh:
This implements proper semantics for cloning text fields.

Tests are added for both dynamic text fields and EditTexts. Additionally, the default text format calculation from SWF tag is improved.


Lord-McSweeney (approved):

core/src/html/text_format.rs
```diff
@@ -166,25 +166,39 @@ impl TextFormat {
...
-            size: et.height().map(|h| h.to_pixels()),
-            color: et
-                .color()
-                .map(|color| swf::Color::from_rgb(color.to_rgb(), 0)),
-            align,
+            size: Some(et.height().map(|h| h.to_pixels()).unwrap_or(12.0)),
+            color: Some(
+                et.color()
+                    .map(|color| swf::Color::from_rgb(color.to_rgb(), 0))
+                    .unwrap_or_default(),
```
Lord-McSweeney: nit: Instead of using `unwrap_or_default`, could this be explicit about which color is used when none is specified?

kjarosh: Done

PR #22647, opened 2025-12-30
avm2: Improve ArrayStorage and Vector usage

SuchAFuriousDeath:
In a lot of places, counting the non-hole values (which from_storage does) can be avoided by directly collecting into an ArrayStorage, because in those places no holes were ever pushed into the Vec.

### DIFF
--- a/core/src/avm2/flv.rs
+++ b/core/src/avm2/flv.rs
@@ -1,4 +1,3 @@
-use crate::avm2::array::ArrayStorage;
 use crate::avm2::object::{ArrayObject, DateObject, ScriptObject, TObject as _};
 use crate::avm2::Value as Avm2Value;
 use crate::context::UpdateContext;
@@ -30,12 +29,11 @@ fn avm2_array_from_flv_values<'gc>(
     context: &mut UpdateContext<'gc>,
     values: Vec<FlvValue>,
 ) -> Avm2Value<'gc> {
-    let storage = ArrayStorage::from_storage(
-        values
-            .iter()
-            .map(|v| Some(v.clone().to_avm2_value(context)))
-            .collect::<Vec<Option<Avm2Value<'gc>>>>(),
-    );
+    let storage = values
+        .iter()
+        .cloned()
+        .map(|v| v.to_avm2_value(context))
+        .collect();
 
     ArrayObject::from_storage(context, storage).into()
 }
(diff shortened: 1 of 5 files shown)


core/src/avm2/flv.rs
```diff
@@ -30,12 +29,10 @@ fn avm2_array_from_flv_values<'gc>(
...
 ) -> Avm2Value<'gc> {
-    let storage = ArrayStorage::from_storage(
-        values
-            .iter()
-            .map(|v| Some(v.clone().to_avm2_value(context)))
-            .collect::<Vec<Option<Avm2Value<'gc>>>>(),
-    );
+    let storage = values
+        .iter()
+        .map(|v| v.clone().to_avm2_value(context))
```
kjarosh: I think you can do `.cloned().map(...)`

core/src/avm2/globals/flash/display/movie_clip.rs
```diff
@@ -260,27 +262,25 @@ pub fn get_scenes<'gc>(
...
+                let scene_class = activation.context.avm2.classes().scene;
+                let args = [
+                    AvmString::new_utf8(activation.gc(), scene_name).into(),
+                    scene_labels.into(),
+                    scene_length.into(),
+                ];
+
+                let scene = scene_class.construct(activation, &args)?;
+
+                Ok(scene)
```
kjarosh: Can't you just do `scene_class.construct(activation, &args)`?

core/src/avm2/globals/flash/display/movie_clip.rs
```diff
@@ -128,22 +128,24 @@ fn labels_for_scene<'gc>(
...
+        .into_iter()
+        .map(|(name, frame)| {
+            let name: Value<'gc> = AvmString::new(activation.gc(), name).into();
+            let local_frame = frame - scene_start + 1;
+            let args = [name, local_frame.into()];
+            let frame_label = frame_label_class.construct(activation, &args)?;
 
-        frame_labels.push(Some(frame_label));
-    }
+            Ok(frame_label)
```
kjarosh: Just `frame_label_class.construct(activation, &args)`?

core/src/avm2/flv.rs
```diff
@@ -30,12 +29,11 @@ fn avm2_array_from_flv_values<'gc>(
...
-    let storage = ArrayStorage::from_storage(
-        values
-            .iter()
-            .map(|v| Some(v.clone().to_avm2_value(context)))
-            .collect::<Vec<Option<Avm2Value<'gc>>>>(),
-    );
+    let storage = values
+        .iter()
+        .cloned()
+        .map(|v| FlvValueAvm2Ext::to_avm2_value(v, context))
```
kjarosh: `.map(|v| v.to_avm2_value(context))`?

SuchAFuriousDeath: Yeah, I don't know what I was smoking here :sweat_smile:

PR #22648, opened 2025-12-30
render: Disable wgpu indirect call validation

SuchAFuriousDeath:
As discussed in Discord, wgpu runs some sort of validation logic and during that runs this compute shader: https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-core/src/indirect_validation/dispatch.rs#L47-L73. The shader uses array\<u32\>, which requires the DYNAMIC_ARRAY_SIZE feature, but wgpu runs it without checking that DYNAMIC_ARRAY_SIZE is actually supported by the device, causing this error (image from Discord):

image
To the best of my knowledge, this seems like a wgpu bug. Since we don't need indirect draw validation anyway as Ruffle doesn't use it, disabling it doesn't lose us anything.

### DIFF
--- a/render/wgpu/src/backend.rs
+++ b/render/wgpu/src/backend.rs
@@ -38,6 +38,19 @@ use swf::Color;
 use tracing::instrument;
 use wgpu::SubmissionIndex;
 
+/// Returns wgpu instance flags with indirect call validation disabled.
+///
+/// wgpu's indirect call validation runs a compute shader that uses `array<u32>`,
+/// which requires the `DYNAMIC_ARRAY_SIZE` feature. However, wgpu runs this shader
+/// without first checking if the device supports that feature, causing device
+/// creation to fail on GPUs that lack it. Since Ruffle doesn't use indirect draws,
+/// disabling this validation has no functional impact.
+///
+/// See <https://github.com/gfx-rs/wgpu/issues/8799>
+pub fn get_wgpu_instance_flags() -> wgpu::InstanceFlags {
+    wgpu::InstanceFlags::default().difference(wgpu::InstanceFlags::VALIDATION_INDIRECT_CALL)
+}
+
 pub struct WgpuRenderBackend<T: RenderTarget> {
     pub(crate) descriptors: Arc<Descriptors>,
     target: T,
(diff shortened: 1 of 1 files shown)


kjarosh:
Can you file a bug/PR to wgpu too?

SuchAFuriousDeath:
I am planning to.

SuchAFuriousDeath:
Filed: https://github.com/gfx-rs/wgpu/issues/8799

render/wgpu/src/backend.rs
```diff
@@ -152,6 +156,8 @@ impl WgpuRenderBackend<crate::target::TextureTarget> {
         }
         let instance = wgpu::Instance::new(&wgpu::InstanceDescriptor {
             backends: backend,
+            flags: wgpu::InstanceFlags::default()
+                .difference(wgpu::InstanceFlags::VALIDATION_INDIRECT_CALL),
```
kjarosh: I don't like the fact that it's in 3 places, aren't they semantically the same?

SuchAFuriousDeath: Not sure what exactly you want from me here :sweat_smile:  Extract these code chunks into a new function?

kjarosh: It depends on what's semantically the same, `InstanceDescriptor`? `flags`? `InstanceDescriptor` differs, but maybe it should be the same for all?

SuchAFuriousDeath: Idk, don't ask me :sweat_smile:

render/wgpu/src/backend.rs
```diff
@@ -37,6 +37,10 @@ use swf::Color;
 use tracing::instrument;
 use wgpu::SubmissionIndex;
 
+pub fn get_wgpu_instance_flags() -> wgpu::InstanceFlags {
+    wgpu::InstanceFlags::default().difference(wgpu::InstanceFlags::VALIDATION_INDIRECT_CALL)
```
kjarosh, line 51: Can you elaborate on it in a comment and in the commit description? Currently the only information in the history is that such validation is disabled, but it's not mentioned why.

kjarosh (approved):

PR #22651, opened 2025-12-30
avm1: Implement *toXML methods in ExternalInterface

kjarosh:
These methods may be useful for validating whether objects have proper structures in tests.

The implementation is not perfect, but good enough :)


core/src/avm1/globals/external_interface.rs
```diff
@@ -214,49 +253,137 @@ pub fn call_in<'gc>(
...
+                .coerce_to_string(activation)?
+                .as_wstr();
+            let mut result = WString::from_utf8("<string>");
+            result.push_str(string);
+            result.push_utf8_bytes(b"</string>");
+            result
+        }
+        _ => {
+            let Some(object) = arg.as_object(activation) else {
+                // TODO What to do in this case?
```
Lord-McSweeney: Could you add an explicit match for `Value::MovieClip`, so it's clearer what "this case" is referring to?

kjarosh: Added a comment, but I don't think adding a match for MC would help us here, as MCs also need to be coerced into objects here. This branch is hit when coercion fails.

Lord-McSweeney: I think `as_object` always returns `Some` when passed a `Value::Object` and only sometimes returns `Some` when passed a `Value::MovieClip`. Are you referring to `coerce_to_object`?

kjarosh: No, I'm talking about `as_object` and that's exactly the case. This branch is hit when the MCR fails to be coerced into an object.

PR #22662, opened 2026-01-01
core: Enforce non-zero dimensions for creating empty textures

SuchAFuriousDeath:
Requires #22610 because of let chains.

Turns out that the enum variant is redundant, because there was already a 0 check in the only place that calls create_empty_texture.


kjarosh:
Hmm, actually let's wait a bit to make sure migration to edition 2024 won't produce new error reports.

PR #22669, opened 2026-01-02
 core: Use system CA store for reqwest &  Enable SOCKS proxy support

Xeleron:
I had issues with local proxies using self-signed certificates and SOCKS proxies weren't working

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2453,11 +2453,11 @@ dependencies = [
  "hyper",
  "hyper-util",
  "rustls",
+ "rustls-native-certs",
  "rustls-pki-types",
  "tokio",
  "tokio-rustls",
  "tower-service",
- "webpki-roots",
 ]
 
 [[package]]
@@ -3721,6 +3721,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "384b8ab6d37215f3c5301a95a4accb5d64aa607f1fcb26a11b5303878451b4fe"
 
 [[package]]
+name = "openssl-probe"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9f50d9b3dabb09ecd771ad0aa242ca6894994c130308ca3d7684634df8037391"
+
+[[package]]
 name = "option-ext"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4176,7 +4182,7 @@ dependencies = [
  "once_cell",
  "socket2",
  "tracing",
- "windows-sys 0.60.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


video/external/Cargo.toml
```diff
@@ -18,7 +18,7 @@ thiserror = { workspace = true }
 
 # Needed for OpenH264:
 libloading = { version = "0.8.9", optional = true }
-reqwest = { version = "0.12.28", default-features = false, features = ["blocking"], optional = true }
+reqwest = { version = "0.12.28", default-features = false, features = ["blocking", "rustls-tls-native-roots", "socks"], optional = true }
```
kjarosh: Isn't adding it in frontend-utils enough?

Xeleron: You are right, I made the assumption that the proxy configured in ruffle_frontend_utils was also being configured for ruffle_video_external which also would have failed for a proxied request (e.g. "Downloading OpenH264 library") if the CA wasn't trusted or a SOCKS proxy being used.

torokati44: I think maybe @kjarosh meant that a crate feature only has to be enabled in one place, it will be effective everywhere - that's how crate features work.

And that's an interesting point: Is/Should the configured proxy also be used to fetch the OpenH264 binary...?

kjarosh: @Xeleron can you describe your use case for socks here?

Xeleron: > @Xeleron can you describe your use case for socks here?

I wasn't able to use a socks4/socks5 proxy without adding it, HTTP proxied worked out of the box as long as the cert was trusted.

kjarosh: Well... I assume you're talking about the Proxy network setting. In which case...

> And that's an interesting point: Is/Should the configured proxy also be used to fetch the OpenH264 binary...?

I don't think that matters much? The binary is persisted anyway.

torokati44: I know, and _if_ the proxy is used only for caching reasons, then it doesn't matter - but if it's for access control reasons (and no connectivity is allowed unproxied), then fetching the binary won't work. (But I might not understand everything correctly...)

kjarosh: Yeah, access control is definitely an issue. We might want to disable proxying for OpenH264 then

torokati44: But what if there's internet access _only_ through the proxy, and it would let the binary through?

kjarosh: I think we should still prefer doing it through the local network. It's a system component, access control should be enforced per the user's physical location and proxy is mainly meant for content, not for everything.

The user can always use a VPN or something.

torokati44: So you wouldn't mind if users who are "forced" behind a proxy (as per school regulations, etc.) to not be able to fetch the OpenH264 binary (automatically at least)? Granted, it's a niche scenario, and most content doesn't use H264, so it shouldn't be a big deal...

kjarosh: The way proxy is configured currently is that you set it per content, so it makes sense for it to apply for content only. If you set the proxy in preferences, it would probably make sense to fetch OpenH264 through the proxy too.

torokati44: Aaah, okay, this is what was missing for me. Alright then.

PR #22670, opened 2026-01-02
core: Propagate mouse events to children in reverse order

### LINKED ISSUE #8357: Gowling - You cannot hit the crabs
### Describe the bug

In Gowling, you cannot hit the crabs, because the character will automatically swing the ball (and not hit the crabs) after you've decided the power of your shots (Weak/Strong), instead of deciding the angle right after.

http://game.freegamesjungle.com/game_0003/free/gowling.swf

### Expected behavior

After you've decided on the power of your shots (Weak/Strong), you will decide on the angle next. And then you can hit the crabs.

### Affected platform

Browser's extension

### Operating system

Windows 10

### Browser

_No response_

### Additional information

_No response_

SuchAFuriousDeath:
Fixes #8357 (tested) by resurrecting #9295 from the dead, thereby superseding it. Authorship retained.


SuchAFuriousDeath:
EDIT: I only ran avm1 and avm2 tests locally to see if I didn't break anything, but it turns out that this actually progresses some existing failing tests completely free of charge :)

tests/tests/swfs/from_gnash/misc-ming.all/PrototypeEventListeners/test.toml
```diff
@@ -8,3 +7,4 @@ player_options.version = 10
 [subtests.fp11]
 output_path = "output.fp11.txt"
 player_options.version = 11
+known_failure = true
```
kjarosh, line 10: If you modify the test in the second commit, tests will fail after the first. It's okay to add a new test in the second commit, but you have to update tests within the same commit that contains code changes that result in a different behavior.

SuchAFuriousDeath: Of course :facepalm: I just amended the second commit automatically without thinking, because it was test-related :sweat_smile:

tests/tests/swfs/avm1/clip_event_propagation_order/Makefile
```diff
@@ -0,0 +1,11 @@
+CLIPS=Main.as Child.as
```
kjarosh: Last time I tried to include compilation scripts, others were rather negative about that. (https://github.com/ruffle-rs/ruffle/pull/17605) I'm personally all for it, but my stance seems to be unpopular.

SuchAFuriousDeath: Oh :sweat_smile: Should I remove just the makefile? Or more?

core/src/display_object/interactive.rs
```diff
@@ -235,7 +236,15 @@ pub trait TInteractiveObject<'gc>:
     ) -> ClipEventResult {
         if event.propagates() {
             if let Some(container) = self.as_displayobject().as_container() {
-                for child in container.iter_render_list() {
+                // Button and key events fire in render list order (low depth to high depth).
+                // Mouse events fire in reverse order (high depth to low depth).
+                let children = if event.is_button_event() || event.is_key_event() {
```
kjarosh: What about events that are neither button, nor key, nor mouse events?

SuchAFuriousDeath: For those, event.propagates() should return false (I think). It's the first line of this method.

https://github.com/ruffle-rs/ruffle/blob/884724ae1037414b90747f4b3bb5e0512189c33f/core/src/events.rs#L368-L392

I am open to refactoring this for clarity. Maybe we could add is_mouse_event() and use it to invert the if statement?

kjarosh: Yeah, that's a good idea

SuchAFuriousDeath: Done

core/src/events.rs
```diff
@@ -391,6 +391,11 @@ impl ClipEvent<'_> {
         matches!(self, Self::KeyDown | Self::KeyUp | Self::KeyPress { .. })
     }
 
+    /// Indicates whether this is a mouse event type (mouseUp, mouseDown, mouseMove).
+    pub const fn is_mouse_event(self) -> bool {
+        matches!(self, Self::MouseUp | Self::MouseDown | Self::MouseMove)
```
kjarosh: Aren't mouse wheel / press / release events mouse events too?

SuchAFuriousDeath: Good point, I renamed it to is_propagating_mouse_event.

kjarosh: I rather thought about covering other mouse events too. That way the method could be useful for some other code, and in this case we don't particularly care. It also fits with other methods checking whether it's a button or key event. Don't you agree?

SuchAFuriousDeath: I do.

PR #22542, opened 2025-12-18
chore: Run `npm update` in web/

torokati44:
This is the "rest of" https://github.com/ruffle-rs/ruffle/pull/22528.

PR #22683, opened 2026-01-05
chore: Drop deprecated `lodash.isequal` indirect npm dependency

torokati44:
By running `npm update expect-webdriverio`, bumping to a version that replaced it with `deep-eql`.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -1871,22 +1871,22 @@
             }
         },
         "node_modules/@jest/expect-utils": {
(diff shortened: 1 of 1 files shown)


torokati44:
Oof, looks like this is a subset of https://github.com/ruffle-rs/ruffle/pull/22542 - and is also blocked by the same API/typing change as mentioned there... My apologies!

torokati44:
Although, it might make sense to do this separately?

PR #22687, opened 2026-01-05
avm2: Multiple fixes to frame script running and frame construction

### LINKED ISSUE #22677: Regression: Wedding Trouble - Cannot access a property or method of a null object reference
### Describe the bug

I can not select the level. Looks like the regression started with https://github.com/ruffle-rs/ruffle/pull/22408

```
ERROR core/src/display_object/movie_clip.rs:2389 Error occurred when running AVM2 frame script: TypeError: Error #1009: Cannot access a property or method of a null object reference.
at rutch.theStage::TheStage/frame41()
```

### Expected behavior

Select the level

### Content Location

PLAY Online - https://www.gameflare.com/online-game/wedding-trouble/
SWF URL  - https://data.gameflare.com/games/820/9DN05aTvnlcb9e.swf

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

Chrome

### Additional information

_No response_

Lord-McSweeney:
All commits except for the last are minor refactors.

The last commit fixes #22677. #21572 removed an extra `construct_frame` call in `run_inner_goto_frame`. The `construct_frame` was actually supposed to be run, though only for frame-script-queued gotos. This fully addresses Aaron's comment that that PR removed. If the `construct_frame` isn't run, queued frame scripts can observe not-yet-constructed children.

The last commit also includes another minor fix to "no-op" frame-script-queued gotos that allows for the added tests to pass.


Lord-McSweeney:
I'm going to split this into multiple PRs

PR #22688, opened 2026-01-05
avm1: Fix prototype chain for movies loaded into levels

### LINKED ISSUE #22395: avm1: Character in Sunny Day Sky falls through the platform (regression)
### Describe the bug

Start the game.
→ The character immediately falls through the platform.

### Expected behavior

The character should stand on the platform. 🙂

This issue first appeared in [nightly-2025-10-12](https://github.com/ruffle-rs/ruffle/releases/tag/nightly-2025-10-12).

Note that the game uses a loader which is a SWF version 4, while the game itself is a SWF version 8.
If you open the loader in JPEXS and change its SWF version to 7 or higher, the issue disappears. This suggests that PR #21863 is the likely cause.

### Content Location

https://www.ferryhalim.com/orisinal/g3/sunny.htm
→ Note that the website uses an outdated self-hosted version of Ruffle that does not exhibit this bug, so you may need to block `ruffle.js` from loading (e.g., via the browser console) in order to reproduce the issue.

OR

[sunny.zip](https://github.com/user-attachments/files/23889930/sunny.zip) (the files I used for testing in 2020/21, ahah - note that `00sunny.swf` is the loader)

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

_No response_

### Additional information

_No response_

SuchAFuriousDeath:
Fixes #22395

When a lower-version SWF (e.g., SWF6) loads a higher-version SWF (e.g., SWF8) via loadMovieNum, the loaded movie's _levelN clip had the wrong \_\_proto\_\_. This caused the loaded movie to be unable to see properties added to its version's MovieClip.prototype.

In loader.rs, get_or_create_level creates the level clip using the loader's movie version for the prototype. Then replace_with_movie updates the movie reference but doesn't update \_\_proto\_\_. So when SWF6 loads SWF8 into _level2, the clip ends up with SWF6's prototype instead of SWF8's.


core/src/loader.rs
```diff
@@ -1600,6 +1600,22 @@ impl<'gc> MovieLoader<'gc> {
...
+                    // Update the MovieClip's script object prototype to match the new movie's version.
+                    // This is needed because the level clip may have been created by a loader with
+                    // a different SWF version, and its script object was set up with the wrong prototype.
+                    if !movie.is_action_script_3() {
+                        if let Some(object) = mc.object1() {
+                            let new_proto = uc.avm1.prototypes(mc.swf_version()).movie_clip;
+
+                            object.define_value(
+                                uc.gc(),
+                                istr!(uc, "__proto__"),
```
kjarosh, line 1611: What happens to it if the movie is unloaded?

SuchAFuriousDeath: I didn't think about that, great question. I'll investigate.

PR #22696, opened 2026-01-05
web: Handle missing OffscreenCanvas for canvas font renderer

kjarosh:
Some browsers might not support OffscreenCanvas. This patch makes sure we handle the situation properly.


n0samu:
To clarify, do we fall back to our normal method of font rendering in this case?

kjarosh:
> To clarify, do we fall back to our normal method of font rendering in this case?

Yes, the default font is always the fallback when enabled.

danielhjacobs:
Specifically it will go to this path, right?

https://github.com/ruffle-rs/ruffle/blob/7e90df64b2c15c69d0fd53c10568efa76adcd6bf/web/src/ui.rs#L335-L340

danielhjacobs (approved):

PR #22701, opened 2026-01-06
tests: Add test for breaking words

kjarosh:
The `avm2/edittext_wrap_breaks` test checks which characters should break.


kjarosh:
CC @adrian17

kjarosh:
This test needs fixing as some characters are missing

Issue #12172, opened 2023-07-20
Dino Robot Adventure: Cannot start game


n0samu:
### Describe the bug

Link to SWF: http://www.kidszzanggame.com/minigame/swf/400.swf

Loading [Dino Robot Adventure](http://www.kidszzanggame.com/d/player.html?game_id=400&flag=minigame) (이노로봇 어드벤처 고대광산) in Ruffle shows nothing but a black screen. The error is:
```
ERROR run_frame:run_all_phases_avm2: ruffle_core::display_object::movie_clip: Got "TypeError: Error #1010: A term is undefined and has no properties.
        at global$init() [TU=starling/display/Stage]
        at starling.core::Starling()
        at DinoAdv/init()
        at DinoAdv()" when constructing AVM2 side of movie clip of type DinoAdv
```
There are also a large number of warnings, such as:
```
WARN run_frame: ruffle_core::display_object::movie_clip: Error loading ABC file: ReferenceError: Error #1065: Variable starling.rendering::BatchToken is not defined.
        at starling.display::DisplayObjectContainer$cinit()
        at global$init() [TU=starling/display/DisplayObjectContainer]
        at global$init() [TU=Game]
```

### Expected behavior

The game should start.

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

The original webpage tries to load the game through the IMA SDK, which causes the same problem as #1840.

Lord-McSweeney:
Still happening...

Lord-McSweeney:
Still happening

Lord-McSweeney:
The SWF is no longer up on the original site, but can still be accessed through the Wayback Machine: https://web.archive.org/web/20230720033845/http://www.kidszzanggame.com/minigame/swf/400.swf

Lord-McSweeney:
This has progressed a bit: you can now get to open a menu to start a game. Actually trying to start a game seems to cause the SWF to hang, though.

PR #22708, opened 2026-01-07
Add totalAscent and totalDescent stubs to flash.text.engine.TextLine

Fancy2209:
Needed for 
https://web.archive.org/web/20160502062843/http://feathersui.com/examples/components-explorer/

Place in FeathersUI' Source that uses it
https://github.com/feathersui/feathersui-starling/blob/master/source/feathers/controls/text/TextBlockTextRenderer.as

may help with
#12172

### DIFF
--- a/core/src/avm2/globals/flash/text/engine/TextLine.as
+++ b/core/src/avm2/globals/flash/text/engine/TextLine.as
@@ -53,7 +53,17 @@ package flash.text.engine {
             return 12.0;
         }
 
+        public function get totalAscent():Number {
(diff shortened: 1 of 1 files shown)


core/src/avm2/globals/flash/text/engine/TextLine.as
```diff
@@ -53,11 +53,21 @@ package flash.text.engine {
             return 12.0;
         }
 
+        public function get totalAscent():Number {
```
kjarosh, line 57: These methods should be marked as version 670

kjarosh (approved):

Issue #22075, opened 2025-10-29
Issue when creating a new Sound() with an unloaded MovieClip


ChrisCPI:
### Describe the bug

In this test, two instances of Sound() are created. The first one is with `undefined`, the second with a MovieClip that has been removed already. When setting the volume of the second sound, it affects the volume of the first one. This is not the case in Flash Player.

The code: 
```as
var test_mc = this.createEmptyMovieClip("test_mc",1);
var test_sound = new Sound(undefined);
trace(test_sound.getVolume());
test_mc.removeMovieClip();
var mc_sound = new Sound(test_mc);
mc_sound.setVolume(0);
trace(test_sound.getVolume());
```

Ruffle output:
```
100
0
```

Flash Player output:
```
100
100
```

### Expected behavior

Changing the volume of the Sound attached to the MC should not affect the volume of the first one.

### Content Location

[undefined_sound_bug.swf.zip](https://github.com/user-attachments/files/23200105/undefined_sound_bug.swf.zip)

### Affected platform

Desktop app

### Operating system

macOS 15.6.1

### Additional information

I *think* this is what is causing a bug in Club Penguin where leaving your igloo causes all audio to mute. This took a ridiculously long time to figure out.

ChrisCPI:
Interesting, in Flash Player `mc_sound.getVolume()` returns `undefined`.

PR #22711, opened 2026-01-08
avm1: Fix MovieClipReference laundering issue

### LINKED ISSUE #15265: Clicker Monsters: After collecting 100 coins, no more coins appear

SuchAFuriousDeath:
Fixes #15265 (I can't reproduce it though, maybe I'm doing something wrong, but the game works for me with or without these changes) 

This resurrects (and supersedes) #20840. Authorship retained. I barely changed anything, so not adding myself as a co-author. 

Note: this merges the 2 commits from the original into 1 because it makes more sense to me (the code got refactored since and one commit would be just a 1-line change, it doesn't make sense to split it off to me).


ChrisCPI:
Out of curiosity, would this fix #22075? It sounds very similar to the example reproduction in https://github.com/ruffle-rs/ruffle/issues/15265#issuecomment-2041207856.

SuchAFuriousDeath:
I tried and it doesn't. I can try to take a look, but not making any promises.

core/src/avm1/object_reference.rs
```diff
@@ -177,7 +184,12 @@ impl<'gc> MovieClipReference<'gc> {
 
     /// Convert this reference to an `Object`
     pub fn coerce_to_object(self, activation: &mut Activation<'_, 'gc>) -> Option<Object<'gc>> {
-        let (_, object, _) = self.resolve_reference(activation)?;
+        let (cached, object, display_object) = self.resolve_reference(activation)?;
+        // Track the original reference used to produce this clip, so we can get it back later.
+        // Only do this if the cache was invalid, as that's the state we need to preserve.
+        if !cached && let Some(mc) = display_object.as_movie_clip() {
+            mc.set_original_reference(activation.gc(), self);
```
kjarosh, line 190: Can't two different MCRs point to the same MC? In which case they would end up overriding each other in cases where storing MC interleaves with code execution.

SuchAFuriousDeath: Looking into it.

PR #22713, opened 2026-01-08
chore: reformat `core` using rustfmt 2024 style

moulins:
Follow-up to #22610; this mostly reorders imports.


core/src/avm1/globals/number.rs
```diff
@@ -85,11 +85,7 @@ fn to_string<'gc>(
         [] => 10,
         [radix, ..] => {
             let radix = radix.coerce_to_f64(activation)? as i32;
-            if (2..=36).contains(&radix) {
-                radix
-            } else {
-                10
-            }
+            if (2..=36).contains(&radix) { radix } else { 10 }
```
kjarosh, line 88: I wouldn't agree that this change is necessarily more readable

danielhjacobs: Seems like that's https://doc.rust-lang.org/edition-guide/rust-2024/rustfmt-formatting-fixes.html#format-the-last-expression-statement-as-an-expression

moulins: I don't think this is worth a `#[rustfmt::skip]`, though...

kjarosh: No, it's not. I'm wondering about this rule in general. If you have code worthy of splitting into multiple lines, you cannot use `return`s because there's a lint for that, and this formatting rule will force you to write a one liner.

danielhjacobs: I've looked into it, and it only converts the `if` to a one-liner if the one-liner is 50 characters or less. This one-liner happens to be exactly 50 characters. We could create a rustfmt.toml with `single_line_if_else_max_width = 49` to revert this line to its previous formatting if we want though.

https://github.com/rust-lang/rustfmt/blob/main/Configurations.md#single_line_if_else_max_width

torokati44: My two cents: this is fine and https://github.com/ruffle-rs/ruffle/pull/22718 is not worth starting a (non-trivial) `rustfmt.toml` over.

PR #22718, opened 2026-01-08
chore: Address style concern from #22713

danielhjacobs:


SuchAFuriousDeath:
Is this the only occurence? I think in the review process of #22610, more examples were mentioned.

moulins:
I don't think this is a good idea, for two reasons:
- it feels like a hack; we're using a global formatting setting to affect a particular line of code
- I don't even find the one-liner less readable than the expanded form (personally I actually don't like how eager rustfmt is to split things on multiple lines, so I would want to increase this setting instead ^^)

If you want to improve readability, what about rewriting the whole `let radix = match args { ... }` to something like this?
```rs
let radix = args
  .get(0)
  .map(|radix| radix.coerce_to_i32(activation))
  .transpose()?
  .filter(|radix| 2..=36.contains(&radix))
  .unwrap_or(10);
```

danielhjacobs:
> Is this the only occurence? I think in the review process of #22610, more examples were mentioned.

It's a bit more nuanced if I understand correctly. The `single_line_if_else_max_width` rule has been a thing since before Rust edition 2024. What's new in Rust edition 2024 is that "the last statement in a block which is an expression is now formatted as an expression", which means that rule now applies to the last statement in a block when it did not used to do so. Those statements that are newly linted into single-line if/else are often shorter than the ones that were already being linted even before that PR, but were not the last statements in a block.

This is the only exactly 50 character instance of `single_line_if_else_max_width`, but let's make a small list with rule set to different values.

48:

```diff
--- a/core/src/bitmap/operations.rs
+++ b/core/src/bitmap/operations.rs
@@ -655,7 +655,11 @@ pub fn scroll<'gc>(
         // x can be any sign
         (-x).max(0)
     };
-    let x_to = if reverse_x { -1 } else { width.min(width - x) };
+    let x_to = if reverse_x {
+        -1
+    } else {
+        width.min(width - x)
+    };
```

```diff
--- a/core/src/html/text_format.rs
+++ b/core/src/html/text_format.rs
@@ -311,7 +311,11 @@ impl TextFormat {
             } else {
                 None
             },
-            url: if self.url == rhs.url { self.url } else { None },
+            url: if self.url == rhs.url {
+                self.url
+            } else {
+                None
+            },
```

47:

```diff
--- a/core/build_playerglobal/src/lib.rs
+++ b/core/build_playerglobal/src/lib.rs
@@ -494,7 +494,11 @@ fn write_native_table(data: &[u8], out_dir: &Path) -> Result<Vec<u8>, Box<dyn st
                         // instance methods. Instead of subtracting 1 from it the disp-id,
                         // add 1 if it's a class method, or subtract 2 if it's
                         // an instance method.
-                        let disp_id = if is_class { disp_id + 1 } else { disp_id - 2 };
+                        let disp_id = if is_class {
+                            disp_id + 1
+                        } else {
+                            disp_id - 2
+                        };
```

```diff
--- a/core/src/system_properties.rs
+++ b/core/src/system_properties.rs
@@ -304,11 +304,19 @@ impl SystemProperties {
     }
 
     fn encode_capability(&self, cap: SystemCapabilities) -> &str {
-        if self.has_capability(cap) { "t" } else { "f" }
+        if self.has_capability(cap) {
+            "t"
+        } else {
+            "f"
+        }
     }
 
     fn encode_not_capability(&self, cap: SystemCapabilities) -> &str {
-        if self.has_capability(cap) { "f" } else { "t" }
+        if self.has_capability(cap) {
+            "f"
+        } else {
+            "t"
+        }
     }
```

46:

```diff
--- a/core/src/avm1/activation.rs
+++ b/core/src/avm1/activation.rs
@@ -1567,7 +1567,11 @@ impl<'a, 'gc> Activation<'a, 'gc> {
 
         // Index is 1-based for this opcode.
         let start = self.context.avm1.pop().coerce_to_i32(self)?;
-        let start = if start >= 1 { start as usize - 1 } else { 0 };
+        let start = if start >= 1 {
+            start as usize - 1
+        } else {
+            0
+        };
 
         let val = self.context.avm1.pop();
         let s = val.coerce_to_string(self)?;
@@ -2031,7 +2035,11 @@ impl<'a, 'gc> Activation<'a, 'gc> {
 
         // Index is 1-based for this opcode.
         let start = self.context.avm1.pop().coerce_to_i32(self)?;
-        let start = if start >= 1 { start as usize - 1 } else { 0 };
+        let start = if start >= 1 {
+            start as usize - 1
+        } else {
+            0
+        };
```

```diff
--- a/core/src/avm2/globals/date.rs
+++ b/core/src/avm2/globals/date.rs
@@ -238,7 +238,13 @@ pub fn init<'gc>(
                 .minute(arguments.get(4))?
                 .second(arguments.get(5))?
                 .millisecond(arguments.get(6))?
-                .map_year(|year| if year < 100.0 { year + 1900.0 } else { year })
+                .map_year(|year| {
+                    if year < 100.0 {
+                        year + 1900.0
+                    } else {
+                        year
+                    }
+                })
                 .apply(this);
         } else {
             let timestamp = if let Value::String(date_str) = timestamp {
@@ -912,7 +918,13 @@ pub fn utc<'gc>(
         .minute(args.get(4))?
         .second(args.get(5))?
         .millisecond(args.get(6))?
-        .map_year(|year| if year < 100.0 { year + 1900.0 } else { year })
+        .map_year(|year| {
+            if year < 100.0 {
+                year + 1900.0
+            } else {
+                year
+            }
+        })
```

```diff
--- a/core/src/avm2/object/namespace_object.rs
+++ b/core/src/avm2/object/namespace_object.rs
@@ -104,7 +104,11 @@ impl<'gc> TObject<'gc> for NamespaceObject<'gc> {
         last_index: u32,
         _activation: &mut Activation<'_, 'gc>,
     ) -> Result<u32, Error<'gc>> {
-        Ok(if last_index < 2 { last_index + 1 } else { 0 })
+        Ok(if last_index < 2 {
+            last_index + 1
+        } else {
+            0
+        })
     }
```

```diff
--- a/core/src/avm2/object/qname_object.rs
+++ b/core/src/avm2/object/qname_object.rs
@@ -135,7 +135,11 @@ impl<'gc> TObject<'gc> for QNameObject<'gc> {
         last_index: u32,
         _activation: &mut Activation<'_, 'gc>,
     ) -> Result<u32, Error<'gc>> {
-        Ok(if last_index < 2 { last_index + 1 } else { 0 })
+        Ok(if last_index < 2 {
+            last_index + 1
+        } else {
+            0
+        })
```

```diff
--- a/core/src/font.rs
+++ b/core/src/font.rs
@@ -635,7 +635,11 @@ impl<'gc> Font<'gc> {
                 metrics: FontMetrics {
                     // DefineFont3 stores coordinates at 20x the scale of DefineFont1/2.
                     // (SWF19 p.164)
-                    scale: if tag.version >= 3 { 20480.0 } else { 1024.0 },
+                    scale: if tag.version >= 3 {
+                        20480.0
+                    } else {
+                        1024.0
+                    },
```

```diff
--- a/core/src/html/layout.rs
+++ b/core/src/html/layout.rs
@@ -523,7 +523,11 @@ impl<'a, 'gc> LayoutContext<'a, 'gc> {
 
         fn describe_font(span: &TextSpan) -> String {
             let bold_suffix = if span.style.bold { ", bold" } else { "" };
-            let italic_suffix = if span.style.italic { ", italic" } else { "" };
+            let italic_suffix = if span.style.italic {
+                ", italic"
+            } else {
+                ""
+            };
```

danielhjacobs:
The one mentioned in that PR has a length of 46 and would be linted with a rule set to 45 (along with all the above).

PR #22730, opened 2026-01-10
core: Reimplement text wrapping for compatibility

adrian17:
Implements better line wrapping algorithm, with CJK character handling, parentheses handling and edge cases for SWF<=7.

Should pass the test in https://github.com/ruffle-rs/ruffle/pull/22701 .


kjarosh:
The CI is failing :/ also this will need unmarking the test from https://github.com/ruffle-rs/ruffle/pull/22701 as known failure.

core/src/font.rs
```diff
@@ -926,87 +927,286 @@ pub trait FontLike<'gc> {
         width
     }
 
+    // FP line wrapping behavior summary.
```
kjarosh: Placement of this comments seems random, shouldn't it be inside/above the wrapping methods?

core/src/font.rs
```diff
@@ -926,87 +927,286 @@ pub trait FontLike<'gc> {
...
+    // SWF <= 7 differences:
+    // - Breaks are allowed on both sides of '-'.
+    // - Any space is a valid break point.
+    // - Spaces are not collapsed; only the final space of the slice (if present)
+    //   is dropped for measurements.
+    //   (this means we can never split "a " apart, since its space never counts)
+    // - Final fallback:
+    //   if resulting length has <=1 char, yield entire line without wrapping.
+
+    fn is_cjk_like_char(c: char) -> bool {
```
kjarosh: Can we move these 3 methods to wstr::utils?

core/src/font.rs
```diff
@@ -926,87 +927,286 @@ pub trait FontLike<'gc> {
...
+            | '〟'
+        )
+    }
+
+    // IMO way more readable the way it is rn
+    #[allow(clippy::if_same_then_else)]
+    #[allow(clippy::collapsible_if)]
+    fn find_allowed_breaks(text: &WStr, swf7: bool) -> Vec<usize> {
+        let mut ret = vec![];
+        if text.is_empty() {
```
kjarosh: Instead of doing is empty check you can do `let Some(ch) = text.chars().next() else { return ret; };`

core/src/font.rs
```diff
@@ -926,87 +927,286 @@ pub trait FontLike<'gc> {
...
+            | '〜'
+            | '︺' | '﹀' | '﹂' | '﹄' | '︶'
+            | '〟'
+        )
+    }
+
+    // IMO way more readable the way it is rn
+    #[allow(clippy::if_same_then_else)]
+    #[allow(clippy::collapsible_if)]
+    fn find_allowed_breaks(text: &WStr, swf7: bool) -> Vec<usize> {
```
kjarosh: Nit: `swf7` kinda suggests we're talking about behavior for >=SWF7, not <=SWF7. We can invert this value and name it `swf8`.

core/src/font.rs
```diff
@@ -926,87 +927,286 @@ pub trait FontLike<'gc> {
...
+        let mut ret = vec![];
+        if text.is_empty() {
+            return ret;
+        }
+        // note: FP probably handles bad utf16 in some universal way
+        const FALLBACK: char = char::REPLACEMENT_CHARACTER;
+        let mut prev = text.chars().next().unwrap().unwrap_or(FALLBACK);
+        for (i, curr) in text.char_indices().skip(1) {
```
kjarosh: Have you tried `for (i, (left, right)) in chars.zip(chars.skip(1)).enumerate()`? I think this will be clearer than the imperative approach.

core/src/font.rs
```diff
@@ -926,87 +927,286 @@ pub trait FontLike<'gc> {
...
+            } else if prev == '-' || (swf7 && curr == '-') {
+                ret.push(i);
+            } else if Self::is_cjk_like_char(prev) || Self::is_cjk_like_char(curr) {
+                if !Self::is_opening_char(prev) && !Self::is_closing_char(curr) {
+                    ret.push(i);
+                }
+            }
+            prev = curr;
+        }
+        ret.push(text.len());
```
kjarosh: Why do we need to include the last position?

kjarosh:
This will need squashing right?

core/src/font.rs
adrian17: I wanted them to cover the entire block of wrapping-related code (in particular, have both swf7 and swf8 explained in one place). I can move them lower though.

adrian17:
Sure, I'll squash it at the end.

core/src/font.rs
adrian17: To me it just makes sense this way? And is similar to what `unicode_linebreak` crate does IIRC.
Each section of text we try to fit is marked by an end index, so to check whether the last "word" of `a b c` fits or not, we need `text.len()` to mark it.

core/src/font.rs
adrian17: This is slightly more awkward, as now we've got two utf16-decoding iterators in parallel, duplicating work, right?
I'm fine with it the way it is right now, personally.

kjarosh: > This is slightly more awkward, as now we've got two utf16-decoding iterators in parallel, duplicating work, right?

If that's a concern you can use https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.tuple_windows (`itertools` is already a transitive dependency).

> I'm fine with it the way it is right now, personally.

The issue is that the code is mixing up setting up an iteration with the semantics of an iteration (e.g. `prev = curr;` is duplicated and requires some thinking to get it right). By renaming `prev` and `curr` to `left` and `right`, and making sure the loop body is stateless you're simplifying code and making it obvious that we're iterating over consecutive pairs of characters in a string.

core/src/font.rs
kjarosh: The main thing I'm concerned about is that they are going to end up orphaned here in the future. They have to be anchored to some code, otherwise various refactors, moving code around will cause them to just dangle here.

You can e.g. make one method `wrap_line` that calls either `wrap_line_swf7` or `wrap_line_swf8` and add this comment above it. This would also simplify caller's code and would make it easier to parameterize unit tests that currently have swf8 hardcoded.

core/src/font.rs
kjarosh: That makes sense, but it's a bit non-intuitive as the method is named `find_allowed_breaks` (why would we break _after_ the text?). Can you include these semantics in the doc comment?

core/src/font.rs
```diff
@@ -926,87 +928,249 @@ pub trait FontLike<'gc> {
...
+    // - Any space is a valid break point.
+    // - Spaces are not collapsed; only the final space of the slice (if present)
+    //   is dropped for measurements.
+    //   (this means we can never split "a " apart, since its space never counts)
+    // - Final fallback:
+    //   if resulting length has <=1 char, yield entire line without wrapping.
+
+    // IMO way more readable the way it is rn
+    #[allow(clippy::if_same_then_else)]
+    #[allow(clippy::collapsible_if)]
```
kjarosh: Can you move those to the exact places where they are needed and use `expect` instead of `allow`?

core/src/font.rs
```diff
@@ -1659,7 +1823,7 @@ mod tests {
             let params = eval_parameters_from_parts(Twips::from_pixels(12.0), Twips::ZERO, true);
             let string = WStr::from_units(b"abcdefghijklmnopqrstuv");
             let breakpoint =
-                df.wrap_line(string, params, Twips::from_pixels(200.0), Twips::ZERO, true);
+                df.wrap_line_swf8(string, params, Twips::from_pixels(200.0), Twips::ZERO, true);
```
kjarosh: My understanding is that these tests should work for both SWF7 and SWF8 (or should have minimal differences)? We can use something like https://crates.io/crates/rstest to write parameterized tests here.

tests/tests/swfs/avm1/edittext_tag_indent/test.as
```diff
@@ -1,37 +1,83 @@
-Stage.scaleMode = "noScale";
```
kjarosh, line 1: This is an AVM2 test in `avm1` directory.

What about just adding an AVM2 version of this test (what's here basically), and making the avm1 version a known failure for now? When we fix the text culling it will start passing again.

core/src/font.rs
adrian17: To be honest, this feels even messier to me. I'd rather just move all these functions to a separate file with a top-level comment. Font.rs is already huge as-is and these functions don't actually have much to do with fonts.

kjarosh: SGTM

core/src/font.rs
adrian17: Wrote it as

```
Find the text indices delimiting (ending) non-breakable spans of text.
```

Since this implies `text.len()` must be there, to mark the last span.

core/src/font.rs
adrian17: Can we consider this out of scope? The PR is already bloating up...

kjarosh: Adding a TODO is fine too

core/src/html/line_wrapping.rs
```diff
@@ -0,0 +1,476 @@
...
+            }
+            return Some(line_end);
+        }
+    }
+
+    None
+}
+
+/// Find the text indices delimiting (ending) non-breakable spans of text.
+// IMO way more readable the way it is rn
```
kjarosh, line 157: A stray comment?

core/src/html/line_wrapping.rs
```diff
@@ -0,0 +1,476 @@
+use crate::font::{EvalParameters, FontLike};
+use crate::prelude::*;
+use crate::string::WStr;
+use ruffle_wstr::utils::{swf_is_cjk_like, swf_is_closing, swf_is_opening};
+
+use itertools::Itertools;
+
+// FP line wrapping behavior summary.
```
kjarosh, line 8: You can make it a module-level doc

core/src/html/line_wrapping.rs
```diff
@@ -0,0 +1,404 @@
...
+            // - word: "  b ", trimmed_word: "  b",
+            word.trim_end()
+        } else {
+            // The swf7 difference is that every space is a break point
+            //  and we only trim the final space,
+            // So given "a  b ", the checked words are:
+            // - word: "a ", trimmed_word: "a",
+            // - word: "  ", trimmed_word: " ",
+            // - word: " b ", trimmed_word: " b",
+            assert!(!word.is_empty(), "trying to line-break an empty string?");
```
kjarosh: I think it makes sense to check it irrespectively of swf version, and I think you can just skip the iteration if the word is empty, as it means `find_allowed_breaks` returned duplicate positions. In case of an edge-case bug that somehow ends up in production it won't crash Ruffle.

kjarosh (approved):

kjarosh (approved):

PR #22743, opened 2026-01-12
avm2: Implement flash.system.System.gc()

Fancy2209:


core/src/avm2/globals/flash/system/system.rs
```diff
@@ -5,6 +5,16 @@ use crate::avm2::activation::Activation;
 use crate::avm2::parameters::ParametersExt;
 use crate::avm2::value::Value;
 
+/// Implements `flash.system.System.setClipboard` method
```
SuchAFuriousDeath: This seems wrong.

SuchAFuriousDeath (changes requested):

core/src/player.rs
```diff
@@ -2382,6 +2384,10 @@ impl Player {
 
         // GC
         self.gc_arena.borrow_mut().collect_debt();
+        if self.needs_gc {
+            self.gc_arena.borrow_mut().finish_cycle();
+        }
+        self.needs_gc = false;
```
moulins, line 2391: There's no need to call both `collect_debt` and `finish_cycle`, so it would be cleaner to have:
```rs
        if self.needs_gc {
            self.gc_arena.borrow_mut().finish_cycle();
        } else {
            self.gc_arena.borrow_mut().collect_debt();
        }
```

core/src/context.rs
```diff
@@ -87,6 +87,9 @@ pub struct UpdateContext<'gc> {
     /// Requests that the player re-renders after this execution (e.g. due to `updateAfterEvent`).
     pub needs_render: &'gc mut bool,
 
+    // Requests that the gc finishes it's cycle this execution (e.g. due to 'flash.system.System.gc()')
```
kjarosh: 1. `its`
2. It should be a doc comment 
3. `GC`, not `gc`

kjarosh:
How will Ruffle behave when the content requests GC excessively? How will Flash behave?

Fancy2209:
> How will Ruffle behave when the content requests GC excessively? How will Flash behave?

I still need to test those cases properly, that's why this is still a draft

PR #3611, opened 2021-03-11
core: Introduce BoundsMode enum

relrelb:
This is useful for cases where bounds returned by ActionScript don't match the actual visual bounds of the object.
Inspired by https://github.com/ruffle-rs/ruffle/pull/3214#issuecomment-791685468.

Fixes #1915.

PR #3214, opened 2021-02-11
core: MorphShape improvements

relrelb:
General improvements around `MorphShape`:
* It seems that most functions (e.g. `width`, `height`, `getBounds`, `hitTest`) don't respect the ratio of `MorphShape`, and instead treat it as 0. One exception is `hit_test_shape` that takes the ratio into account.
* `calculate_shape_bounds` used to calculate only the edge bounds, which don't include strokes and are not suitable for hit testing. Now it calculates both shape bounds and edge bounds.
* If the last record is a `StyleChange` (shouldn't happen without manual intervention), it doesn't affect the bounding box of the shape.

Fixes #1915.

PR #22755, opened 2026-01-13
core: Introduce BoundsMode enum

### LINKED ISSUE #1915: Level 38 of Astro Barrier (Club Penguin) is impossible
In level 38 of Astro Barrier, you have to shoot a target at the top of the screen, but you are only given a single chance to shoot because a target expands and blocks a shot's path.  
In flash, the single shot you are allowed to fire goes past the expanding target, letting it reach the top of the screen. But in Ruffle, the expanding target destroys the shot before it can reach the top.

Link to a video with the the bug: https://youtu.be/6-bPsc4k1Vk
Link to the file (two different versions of the same game): https://drive.google.com/file/d/1xZX4J9ioGvmGG5g6GQx5-5qYKO2dFNco/view?usp=sharing https://drive.google.com/file/d/1A0kUts2NgVfki1cJ0XkZF8GpgZPhTDHA/view?usp=sharing
To replicate: Press 3 on the title screen to skip to level 30 and play until level 38

AllinolCP:
Based on #3611.

Fixes #1915.

Input test made from https://github.com/ruffle-rs/ruffle/pull/3214#issuecomment-791685468


core/src/display_object.rs
```diff
@@ -906,6 +906,20 @@ impl<'gc> DisplayObjectBase<'gc> {
     }
 }
 
+/// Indicates which kind of bounds should be returned by `self_bounds`.
+/// In most cases `BoundsMode::Engine` should be used.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
```
Lord-McSweeney: This shouldn't need to implement any of these traits except for `Debug`: there's no reason to copy or clone it, and checking if the mode is a certain mode can be done with `matches!`

core/src/display_object.rs
```diff
@@ -906,6 +906,20 @@ impl<'gc> DisplayObjectBase<'gc> {
...
+pub enum BoundsMode {
+    /// The bounds visible on the stage (e.g. takes MorphShape ratio into
+    /// account). Used for hit testing and rendering.
+    #[default]
```
Lord-McSweeney: Let's not implement `Default` for this, the `Engine` variant isn't that much more common than the `Script` variant

core/src/display_object/morph_shape.rs
```diff
@@ -55,6 +55,17 @@ impl<'gc> MorphShape<'gc> {
             },
         ))
     }
+
+    pub fn ratio_with_mode(self, mode: BoundsMode) -> u16 {
+        if mode == BoundsMode::Script {
```
Lord-McSweeney: Could you write this as a match statement to make it clearer what's happening?

core/src/display_object/morph_shape.rs
```diff
@@ -55,6 +55,17 @@ impl<'gc> MorphShape<'gc> {
...
+            if self.ratio() == 65535 {
+                return 65535;
```
Lord-McSweeney: This line is apparently not covered by any tests- do none of the tests you added check what happens when the MorphShape is fully morphed?

core/src/prelude.rs
```diff
@@ -1,5 +1,5 @@
 pub use crate::display_object::{
-    DisplayObject, HitTestOptions, TDisplayObject, TDisplayObjectContainer,
+    BoundsMode, DisplayObject, HitTestOptions, TDisplayObject, TDisplayObjectContainer,
```
Lord-McSweeney: nit: Could you avoid adding this to the prelude?

core/src/display_object.rs
SuchAFuriousDeath: I think there actually is a reason to copy, there are functions where you need to call 2 functions that both take the mode.

core/src/display_object/morph_shape.rs
```diff
@@ -120,8 +120,14 @@ impl<'gc> TDisplayObject<'gc> for MorphShape<'gc> {
...
     }
 
-    fn self_bounds(self) -> Rectangle<Twips> {
-        let ratio = self.ratio();
+    fn self_bounds(self, mode: &BoundsMode) -> Rectangle<Twips> {
+        let ratio = match mode {
+            // For getBounds(), getRect() or hitTestObject(), return start bounds (0)
+            BoundsMode::Script => 0,
+            // otherwise, use the actual interpolated ratio
+            _ => self.ratio(),
```
SuchAFuriousDeath: Nit: I think it would be better to match explicitly on the other variant in case there is another one added in the future. It also makes the code more readable imo.

AllinolCP:
Sorry for the gazillion commits, I figured since it's going to be squished anyways

core/src/display_object.rs
```diff
@@ -906,6 +906,19 @@ impl<'gc> DisplayObjectBase<'gc> {
     }
 }
 
+/// Indicates which kind of bounds should be returned by `self_bounds`.
+/// In most cases `BoundsMode::Engine` should be used.
+#[derive(Debug)]
```
SuchAFuriousDeath: Contrary to @Lord-McSweeney, I do think that Clone and Copy should be derived here to prevent having to pass references everywhere. But let's wait on him

Lord-McSweeney: Sorry, I must’ve missed the places where multiple methods are called. In that case, it’s fine for the enum to implement Copy and Clone.

SuchAFuriousDeath (approved):
LGTM

nitram779:
Tested. It fixes the issue perfectly.

core/src/display_object.rs
```diff
@@ -1296,32 +1309,38 @@ pub trait TDisplayObject<'gc>:
...
-    fn world_bounds(self) -> Rectangle<Twips> {
-        self.bounds_with_transform(&self.local_to_global_matrix())
+    fn world_bounds(self, mode: BoundsMode) -> Rectangle<Twips> {
+        self.bounds_with_transform(&self.local_to_global_matrix(), mode)
     }
 
     /// The world bounding box of this object, as reported by `Transform.pixelBounds`.
     fn pixel_bounds(self) -> Rectangle<Twips> {
-        self.world_bounds()
+        // Uses BoundsMode::Script as it's only used in ActionScript.
```
adrian17: I'd still suggest passing the `BoundsMode::Script` from the script for consistency.

adrian17:
OK, so the summary is, the places like these:

```
as2 MovieClip::getBounds()
as2 Transform::pixelBounds
as3 DO::getBounds()

core DO::pixel_bounds() // used by AS getPixelBounds()
core EditText::pixel_bounds()
core DO::set_width()
core DO::set_height()
core DO::hit_test_object()
```
should ignore MorphShape's current ratio?

Could this be summarized in one place in code somewhere?
Also, does the test cover the pixel bounds and setters?

adrian17 (approved):
Can you please rebase to squash and remove the merge commit?

I think it should be good to merge overall :)

AllinolCP:
> Can you please rebase to squash and remove the merge commit?
> 
> I think it should be good to merge overall :)

Done!

torokati44:
That final commit message is still not ideal IMHO. 😐
The PR title would have been better there as well.

AllinolCP:
> That final commit message is still not ideal IMHO. 😐 The PR title would have been better there as well.

two blunder commits in a row whoops-

PR #22756, opened 2026-01-13
core: Refactor XML unescape and error handling

SuchAFuriousDeath:
While working on #21107, I discovered that quick_xml errors do not map 1:1 with Flash errors, so make_xml_error needs to be refactored (and all the code adapted). Doing it in that PR would be too hectic (quick-xml has had major breaking changes), so I'm submitting this separately. This PR also does some minor refactors and implements a FIXME.

Everything should map to the exact same error code as before. There's quite a lot of xml tests, so if they all pass, we can be reasonably sure this does not break anything.

Context for the decoder usage removal:

The decoder doesn't actually do anything special for us, since we don't use the encoding feature. So there's no need to pass it around.
 
https://docs.rs/quick-xml/latest/quick_xml/encoding/struct.Decoder.html
> If feature [encoding](https://docs.rs/quick-xml/latest/quick_xml/index.html#encoding) is disabled, the decoder is always UTF-8 decoder: any XML declarations are ignored.

It can be seen here that without the encoding feature, std::str::from_utf8 is all the decoder does.
https://docs.rs/quick-xml/latest/src/quick_xml/encoding.rs.html#121-129


core/src/avm2/e4x.rs
```diff
@@ -765,41 +765,71 @@ impl<'gc> E4XNode<'gc> {
...
                 Ok(event) => event,
-                Err(XmlError::IllFormed(IllFormedError::MismatchedEndTag { expected, found })) => {
-                    // We must accept </a/>, </a />, and </a b="c">
-                    // TODO: Reject </a bc>, </a//>, <a //> etc.
-                    if let Some(rest) = found.strip_prefix(&expected) {
-                        if rest.starts_with([' ', '\t', '/']) {
-                            let node = open_tags.pop().unwrap();
-                            if open_tags.is_empty() {
-                                top_level.push(node);
+                Err(err) => {
```
adrian17: I don't entirely understand why this change with the inner `match`, I was about to propose wrapping it into outer `Err()` branches before I noticed that's how it originally looked like 😅

core/src/avm2/error.rs
```diff
@@ -179,48 +177,79 @@ pub fn make_reference_error<'gc>(
...
+    DuplicateAttribute = 1104,
+}
+
+#[inline(never)]
+#[cold]
+pub fn make_xml_error<'gc>(activation: &mut Activation<'_, 'gc>, code: XmlErrorCode) -> Error<'gc> {
+    make_error!(match code {
+        XmlErrorCode::ElementMalformed => type_error(
             activation,
             "Error #1090: XML parser failure: element is malformed.",
```
adrian17: if all the branches are identical sans error string, how about you only `match` to pick the string, and use the `code` itself to pass the integer error code rather than duplicate the magic number?

core/common/src/xml.rs
```diff
@@ -1,16 +1,17 @@
+use quick_xml::escape::EscapeError as XmlEscapeError;
 use regress::Regex;
 use std::{ops::Deref, sync::LazyLock};
 
-static ENTITY_REGEX: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"&[^;]*;").unwrap());
+// Matches XML entities like &amp; or &#39; or &#x1F600;
+// We exclude both ';' and '&' from the middle to avoid matching across multiple
+// ampersands - e.g. in "& &thing;" we want to match only "&thing;", not "& &thing;"
+static ENTITY_REGEX: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"&[^;&]*;").unwrap());
```
adrian17: To be sure, this changed regex can't cause a functional change?

SuchAFuriousDeath: It shouldn't, I'll check if this is covered by existing tests and add one if it's not.

core/src/avm2/e4x.rs
SuchAFuriousDeath: You're right, this is unnecessary.

core/src/avm2/error.rs
SuchAFuriousDeath: Good point, I think I wanted to do it at some point, but I guess I forgot.

PR #22762, opened 2026-01-13
avm1: Simplify declarations of values

kjarosh:
This patch simplifies declaration of values. Instead of specifying type manually, it's automatically inferred based on the expression type.

Numbers will result in `Value::Number`, strings in `Value::String`, etc.

This also makes it clear it's a value without diluting it into different types, and makes defining `null` clearer.


core/src/avm1/property_decl.rs
```diff
@@ -123,16 +123,8 @@ pub enum DeclKind<'gc> {
     Function(NativeFunction),
     /// Declares a static string value.
     String(&'static [u8]),
```
SuchAFuriousDeath, line 125: Just wondering, is there any use-case for this that could not be changed to use Value like the other variants?

kjarosh: The issue is that we can't get an `AvmString` in a const context and without a gc context. So we pass a constant `&'static [u8]` and then produce `AvmString` when executing declarations. (We actually intern it.)

PR #22769, opened 2026-01-15
desktop: Improve, update and correct Windows metadata

SuchAFuriousDeath:
Split off from #17488

This also implements the debug flag todo from the removed ruffle_desktop.rc and includes some minor refactors.

Not sure why, but CompanyName isn't displayed now and it wasn't even displayed before

Before: 
image

After:
image


desktop/build.rs
```diff
@@ -39,10 +47,50 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+        res.set_version_info(winresource::VersionInfo::FILEFLAGS, 0x1);
+    }
+
+    res.compile()?;
+
+    Ok(())
+}
+
+fn current_year() -> i32 {
+    chrono::Utc::now().year()
```
kjarosh: Not a great idea for reproducible builds :/

kjarosh: I personally would opt for either making it reproducible, hard coding the year, or removing it altogether.

kjarosh:
Where is the version being set and what about nightly releases?

desktop/build.rs
SuchAFuriousDeath: Ok, I'll look into it.

SuchAFuriousDeath:
The version is being taken from Cargo.toml, right now it's still 0.1.0 I think

kjarosh:
What about the other version?

SuchAFuriousDeath:
What other version? :sweat_smile: As for nightly, nothing special happens right now although I agree it would be reasonable to integrate it somehow.

kjarosh:
> What other version?

There are two versions on your screens, one has 4 parts.

> As for nightly, nothing special happens right now although I agree it would be reasonable to integrate it somehow.

And I think that's how it should be. In general try not to change semver, unless something literally does not support it, in which case we can think about transforming the version.

SuchAFuriousDeath:
The other version means 'file version' and I don't know what it's for, I'll research it.

SuchAFuriousDeath:
Per https://learn.microsoft.com/en-us/windows/win32/menurc/stringfileinfo-block:
> Version number of the file—for example, "3.10" or "5.00.RC2". This string is required.

I don't get it tbf :sweat_smile:

SuchAFuriousDeath:
Ok actually I think I do get it now. ProductVersion is the version of the overall product and FileVersion is the version of that particular file. I guess for us they should always be the same, since it's just 1 .exe file.

kjarosh:
> Ok actually I think I do get it now. ProductVersion is the version of the overall product and FileVersion is the version of that particular file. I guess for us they should always be the same, since it's just 1 .exe file.

Looks like a downstream version. The question is whether it's standalone or dependent on product version.

Imagine a distribution of Ruffle, they will have their own compile scripts, etc. In which case from the same version of the product, you can create different binaries. Maybe the version should be always equal to 1 then?

SuchAFuriousDeath:
I think it's mainly for when you have multiple files that are both part of one product.

desktop/build.rs
SuchAFuriousDeath: Ok so I'm thinking about first trying SOURCE_DATE_EPOCH, then falling back to getting the date from git, should I still fall back to the actual current date if both are not available, or just panic/error in that case? In theory, someone building from a source tarball will not have a git repo created.

SuchAFuriousDeath: Ok I did it, you can take a look.

SuchAFuriousDeath:
Btw probably focus on my other PRs first, they are more consequential than this one :sweat_smile:

desktop/build.rs
```diff
@@ -39,10 +49,80 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+        .ok()
+        .filter(|o| o.status.success())
+        .and_then(|o| String::from_utf8(o.stdout).ok())
+        .and_then(|s| s.trim().parse().ok())
+    {
+        return year;
+    }
+
+    // Fall back to current year for non-git builds
+    chrono::Utc::now().year()
```
kjarosh: I still think we should remove it. The build process should not depend on current time.

SuchAFuriousDeath: No problem

SuchAFuriousDeath:
Now we just have to settle on what to do about the versions.

desktop/build.rs
```diff
@@ -39,10 +49,78 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+
+    Ok(())
+}
+
+/// Get the year for copyright notice, supporting reproducible builds.
+///
+/// Priority:
+/// 1. SOURCE_DATE_EPOCH env var (standard for reproducible builds, used by distros)
+/// 2. Git commit date (for CI and local development)
+fn get_copyright_year() -> i32 {
```
kjarosh: Does it mean that if someone doesn't have git and doesn't set `SOURCE_DATE_EPOCH` they won't be able to compile Ruffle?

kjarosh: This could return `Result` instead of panicking

SuchAFuriousDeath: Yeah, basically. Any ideas?

kjarosh: I'm not sure why we're trying to use the current year as copyright year. Why not just leave it as 2020 as that was the first year of publication?

SuchAFuriousDeath: So fall back to 2020 or just hardcode it to 2020?

kjarosh: I would just hardcode it to 2020 + include `and contributors` at the end as it's a community project

SuchAFuriousDeath: Ok

desktop/Cargo.toml
```diff
@@ -63,8 +63,18 @@ ashpd = "0.11.0"
...
 vergen = { version = "8.3.2", features = ["build", "git", "gitcl", "cargo"] }
+winresource = "0.1"
+
+[package.metadata.winresource]
+# https://learn.microsoft.com/en-us/windows/win32/menurc/stringfileinfo-block
+ProductName = "Ruffle"
+FileDescription = "Adobe Flash Player emulator"
+CompanyName = "Ruffle LLC"
+LegalCopyright = "Copyright © 2020 Ruffle LLC and contributors"
+# InternalName should be the module name without extension
```
kjarosh: Module name does not have an extension, no?

desktop/build.rs
```diff
@@ -39,10 +47,37 @@ fn main() -> Result<(), Box<dyn Error>> {
...
     }
 }
+
+fn set_windows_resource() -> Result<(), Box<dyn Error>> {
+    let mut res = winresource::WindowsResource::new();
+
+    // Set language to US English (0x0409)
+    res.set_language(0x0409);
+
+    // Set the application icon (path relative to CARGO_MANIFEST_DIR)
```
kjarosh: Doesn't look like the path is relative to `CARGO_MANIFEST_DIR`, as it's being prefixed

desktop/build.rs
```diff
@@ -39,10 +47,37 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+    println!("cargo:rerun-if-changed={icon_path}");
+
+    res.set_icon(&icon_path);
+
+    // Set debug flag when building in debug mode
+    // VS_FF_DEBUG = 0x1 (from Windows SDK winver.h)
+    if let Ok(profile) = env::var("PROFILE")
+        && profile == "debug"
+    {
+        res.set_version_info(winresource::VersionInfo::FILEFLAGS, 0x1);
```
kjarosh: Aren't there any constants for the 1?

SuchAFuriousDeath: I didn't find any.

SuchAFuriousDeath: Just double-checked just to be sure and they are really not there. I plan to create a PR to the library to add the flags, seems like a worthy addition.

core/src/display_object/edit_text.rs
```diff
@@ -1051,7 +1051,10 @@ impl<'gc> EditText<'gc> {
     /// Returns the selection, but takes into account whether the selection should be rendered.
     fn visible_selection(self) -> Option<TextSelection> {
         let selection = self.0.selection.get()?;
-        #[expect(clippy::collapsible_else_if)]
+        // FIXME: Remove this #[allow] once Rust 1.94 is released.
```
kjarosh: This looks more like a TODO, can you also submit it as a separate PR or commit?

SuchAFuriousDeath: Sure

SuchAFuriousDeath:
@kjarosh https://github.com/BenjaminRi/winresource/pull/46 was merged and released as v0.1.30 and I used the new flag in the code, so this can be merged now.

Issue #22773, opened 2026-01-15
Error on https://angles365.com/game-frame/xsydayhl/old/classroom/halloween/flash/flash/101hallowbasket.swf.html


daliadiez-droid:
# Error Info
Error name: TypeError
Error message: Failed to execute 'compile' on 'WebAssembly': HTTP status code is not ok
Error stack:
```
TypeError: HTTP status code is not ok
```

# Player Info
Allows script access: false

# Page Info
Page URL: https://angles365.com/game-frame/xsydayhl/old/classroom/halloween/flash/flash/101hallowbasket.swf.html

# Browser Info
User Agent: Mozilla/5.0 (X11; CrOS x86_64 14541.0.0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36
Platform: Linux x86_64
Has touch support: true

# Ruffle Info
Version: 0.2.0-nightly.2026.1.2
Name: 0.2.0-nightly.2026.1.2
Channel: nightly
Built: 2026-01-02T00:07:45.759Z
Commit: ecd06ab30ce4657a543733d252f10ca7e276ca0d
Is extension: false

# Metadata

danielhjacobs:
The person who set up Ruffle on this site did not upload the required WASM files, they'll need to fix that. https://angles365.com/ruffle/133fbe7961bd633e1a42.wasm is missing.

PR #22774, opened 2026-01-15
web: Handle additional compile error case in panic.tsx

danielhjacobs:


danielhjacobs:
I will. It's this error message fwiw (https://github.com/ruffle-rs/ruffle/issues/22773) Chromium specific actually, on Firefox the error message mentions a text/html mime type instead and is handled by the "mime" branch of the code (this is an HTML 404 response, Chrome mentions the 404, Firefox the text/html)

Issue #22768, opened 2026-01-15
Unity swf support


Sonic853:
**Describe the feature**
I created this SWF using Unity 4.6.1p4. Currently, the SWF works correctly in Flash Player 14 (Windows EXE), but it doesn't work properly in Ruffle.

Here is the SWF file:

[testswf.zip](https://github.com/user-attachments/files/24639325/testswf.zip)

Here is the project used to build this SWF:

[Unity461Project.zip](https://github.com/user-attachments/files/24639374/Unity461Project.zip)

Here is a screenshot of the SWF running in Flash Player 14:

<img width="1026" height="820" alt="Image" src="https://github.com/user-attachments/assets/6770d852-2f4b-401f-a1ef-4dcbb4db81fc" />

SuchAFuriousDeath:
This needs setProgramConstantsFromByteArray like several other issues, I'm working on it right now.

SuchAFuriousDeath:
With setProgramConstantsFromByteArray implemented, this is spammed:

2026-01-15T17:22:08.072719Z  INFO avm_trace: TypeError: Error #1010: A term is undefined and has no properties. (accessing field: null)
        at com.unity::UnityNative$/Ext_Flash_InvokeMethodOnObject()
        at com.unity::UnityNative$/_Z23scripting_method_invokeP15ScriptingMethodP15ScriptingObjectR18ScriptingArgumentsPP18ScriptingException()
        at com.unity::UnityNative$/_ZN19ScriptingInvocation6InvokeEPP18ScriptingException()
        at com.unity::UnityNative$/_ZZN2UI23InitializeCanvasManagerEvEN18WillRenderCanvases7ForwardEv()
        at com.unity::UnityNative$/NativeExt_PlayerLoop()
        at com.unity::UnityContentInternal/playerLoop()
        at com.unity::UnityContent/onEnterFrame()

Sonic853:
<img width="1229" height="764" alt="Image" src="https://github.com/user-attachments/assets/781da9b9-1000-4167-9bd8-05b84cc52f73" />
It looks like it's working, even though it looks ridiculous. But that doesn’t take away from the fact that this is great work!

SuchAFuriousDeath:
Hmm, the colors are wrong and the shadows are missing, right? So we can't close this yet.

maarceello:
[...]

Issue #20843, opened 2025-06-26
Error dispatching event EventObject / Minion Golf / Mission ImPOPsible


Stefannag:
### Describe the bug

It is starting ok, but when the game scene should appear, it will trigger an error. The same error (or similar) is in two games.

`ERROR core/src/avm2/events.rs:411 Error dispatching event EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x9772a40 } to handler FunctionObject(FunctionObject { ptr: 0x8e75368, name: "com.unity::UnityContent/onEnterFrame()" }) : ReferenceError: Error #1069: Property setProgramConstantsFromByteArray not found on flash.display3D.Context3D and there is no default value.	at com.unity::UnityNative$/Ext_Stage3D_Context_Set_ProgramConstants()	at com.unity::UnityNative$/_ZN9GfxDevice14BeforeDrawCallEb()	at com.unity::UnityNative$/_ZN14GfxMolehillVBO7DrawVBOERKN9ShaderLab14ChannelAssignsEmm16GfxPrimitiveTypemm()	at com.unity::UnityNative$/_ZN8DrawUtil11DrawMeshRawERKN9ShaderLab14ChannelAssignsER4Meshi()	at com.unity::UnityNative$/_ZThn28_N12MeshRenderer6RenderEiRKN9ShaderLab14ChannelAssignsE()	at com.unity::UnityNative$/_ZN13BatchRenderer5FlushEv()	at com.unity::UnityNative$/_Z25DoForwardShaderRenderLoopR23ForwardShaderRenderLoopR17RenderLoopContextR13dynamic_arrayI16RenderObjectDataLj4EL18MemLabelIdentifier52EEbbP13RenderTextureb()	at com.unity::UnityNative$/_Z12DoRenderLoopR10RenderLoop13RenderingPathRSt6vectorI11VisibleNodeSaIS3_EEbP6ShaderRKSs()	at com.unity::UnityNative$/_ZN6Camera6RenderEiP6ShaderRKSs()	at com.unity::UnityNative$/_ZN13RenderManager13RenderCamerasEv()	at com.unity::UnityNative$/NativeExt_PlayerLoop()	at com.unity::UnityContentInternal/playerLoop()	at com.unity::UnityContent/onEnterFrame()`

[...]

Issue #19499, opened 2025-02-12
unity game black screen


zjk261:
### Describe the bug

![Image](https://github.com/user-attachments/assets/4cfa4df4-b950-4a2e-b693-83a46fd3eeee)

### Expected behavior

![Image](https://github.com/user-attachments/assets/ebdfa1df-fd48-464e-824b-08af3d4a92cf)

### Content Location

[66666.zip](https://github.com/user-attachments/files/18766167/66666.zip)

### Affected platform

Desktop app

### Operating system

Windows 10

torokati44:
This game uses `jpegxr`, the feature is supposed to be enabled in local builds for it to work: `--features jpegxr` (it is enabled in releases)

Even then, I get this in the output:
```
2025-02-12T10:58:08.574768Z  WARN ruffle_core::stub: Encountered stub: AVM2 flash.display3D.textures.Texture.uploadFromByteArray() with miplevel != 0
2025-02-12T10:58:11.590973Z  INFO avm_trace: flash_invoke_method called with NULL object, and scriptmethod->m_Class is NULL too.
2025-02-12T10:58:11.591097Z  INFO avm_trace: Error
        at com.unity::UnityNative$/Ext_Flash_LogCallstack()
        at com.unity::UnityNative$/_Z17DebugStringToFilePKciS0_iiii()
        at com.unity::UnityNative$/_Z23scripting_method_invokeP15ScriptingMethodP15ScriptingObjectR18ScriptingArgumentsPP18ScriptingException()
        at com.unity::UnityNative$/_ZN19ScriptingInvocation6InvokeEPP18ScriptingException()
        at com.unity::UnityNative$/_ZZN2UI23InitializeCanvasManagerEvEN18WillRenderCanvases7ForwardEv()
        at com.unity::UnityNative$/NativeExt_PlayerLoop()
        at com.unity::UnityContentInternal/playerLoop()
        at com.unity::UnityContent/onEnterFrame()
2025-02-12T10:58:11.592060Z  INFO avm_trace: flash_invoke_method called with NULL object, and scriptmethod->m_Class is NULL too.
 
(Filename:  Line: 123)

2025-02-12T10:58:11.593126Z  INFO avm_trace: TypeError: Error #1010: A term is undefined and has no properties. (accessing field: null)
        at com.unity::UnityNative$/Ext_Flash_InvokeMethodOnObject()
[...]

Issue #15018, opened 2024-01-30
Does not load - https://www.4399.com/flash/188739_1.htm


wsxarcher:
### Describe the bug

Game does not load after 4399 logo

```
ERROR core/src/avm2/events.rs:428 Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x79adf50 }) to handler FunctionObject(FunctionObject { ptr: 0x6155f50, name: Ok("com.unity::UnityContent/onFrameInitStep()") }) : RustError("A script in this movie has taken too long to execute and has been terminated.")
339.js:1133 ERROR core/src/avm2/events.rs:428 Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x4e2a060 }) to handler FunctionObject(FunctionObject { ptr: 0x6155f50, name: Ok("com.unity::UnityContent/onFrameInitStep()") }) : RustError("A script in this movie has taken too long to execute and has been terminated.")
339.js:1133 WARN core/src/stub.rs:132 Encountered stub: AVM2 flash.display.Loader.unload()
339.js:1133 INFO web/src/log_adapter.rs:18 __________________--
339.js:1133 INFO web/src/log_adapter.rs:18 ctrl skin load OK !!!!
339.js:1133 INFO web/src/log_adapter.rs:18 ______________________________________
339.js:1133 INFO web/src/log_adapter.rs:18 __________  wmCtrl start  ____________
339.js:1133 INFO web/src/log_adapter.rs:18 ______________________________________
339.js:1133 INFO web/src/log_adapter.rs:18 width = 502    height = 357
339.js:1133 INFO web/src/log_adapter.rs:18 _gameInfoStr = 800|500|188739|急速变色龙
339.js:1133 WARN core/src/library.rs:523 Unknown device font "SimSun" (bold: false, italic: false)
339.js:1133 WARN core/src/library.rs:523 Unknown device font "Arial" (bold: true, italic: false)
339.js:1133 INFO web/src/log_adapter.rs:18 doSetStyle----------->100051514
339.js:1133 INFO web/src/log_adapter.rs:18 gameWid = 800
339.js:1133 INFO web/src/log_adapter.rs:18 gameHei = 500
339.js:1133 INFO web/src/log_adapter.rs:18 This collider has some illegal parameters. Choose 'Reset' in the component popup menu to fix it.
[...]

PR #22775, opened 2026-01-15
avm2: Implement flash.display3D.setProgramConstantsFromByteArray

SuchAFuriousDeath:
The image for the tests is taken from an existing stage3d_rotating_cube test.
The test for invalid inputs also covers setProgramConstantsFromVector with an input that is too small.

Should progress #15007, #15018, #16480, #19499, #20843 (seems to work now) and #22768 (also seems to work now)


core/src/avm2/globals/flash/display3D/context_3d.rs
```diff
@@ -388,6 +393,60 @@ pub fn set_program_constants_from_vector<'gc>(
...
+
+    if let Some(context) = this.as_context_3d() {
+        let program_type = args.get_string(activation, 0);
+
+        let program_type = if &*program_type == b"vertex" {
+            ProgramType::Vertex
+        } else if &*program_type == b"fragment" {
+            ProgramType::Fragment
+        } else {
+            panic!("Unknown program type {program_type:?}");
```
Lord-McSweeney: Shouldn't this case be returning error 2008 (or another error)?

core/src/avm2/globals/flash/display3D/context_3d.rs
```diff
@@ -388,6 +393,60 @@ pub fn set_program_constants_from_vector<'gc>(
...
+            return Err(make_error_3669(activation));
+        }
+
+        let num_floats = num_registers * 4;
+        let mut raw_data = Vec::with_capacity(num_floats);
+
+        for i in 0..num_floats {
+            let float_offset = byte_offset + i * 4;
+            let value = data
+                .read_float_at(float_offset)
```
Lord-McSweeney, line 427: Does the `ByteArray`'s `position` influence where the data is read from?

core/src/avm2/globals/flash/display3D/context_3d.rs
SuchAFuriousDeath: Probably, but setProgramConstantsFromVector also panics the same way. I can change it and write tests.

core/src/avm2/globals/flash/display3D/context_3d.rs
SuchAFuriousDeath: Good question, I'll check and include it in the tests.

SuchAFuriousDeath: Seems like it's ignored, I adjusted the test to have a position and it still renders the same.

core/src/avm2/globals/flash/display3D/context_3d.rs
SuchAFuriousDeath: Done.

core/src/avm2/globals/flash/display3D/context_3d.rs
```diff
@@ -302,7 +302,7 @@ pub fn set_program_constants_from_matrix<'gc>(
         } else if &*program_type == b"fragment" {
             ProgramType::Fragment
         } else {
-            panic!("Unknown program type {program_type:?}");
+            return Err(make_error_2008(activation, "programType"));
```
kjarosh: Can we cover those uncovered lines with tests too?

SuchAFuriousDeath: I was intending to, guess I missed this one.

PR #22784, opened 2026-01-16
core: Deduplicate file dialog filter logic between web and desktop

SuchAFuriousDeath:
Just a factoring out of some common logic. Also moves the platform detection outside of the filter for loop. 
No functional change intended.

### DIFF
--- a/core/src/backend/ui.rs
+++ b/core/src/backend/ui.rs
@@ -47,6 +47,22 @@ pub struct FileFilter {
     pub mac_type: Option<String>,
 }
 
+impl FileFilter {
+    /// Returns extensions suitable for file dialogs.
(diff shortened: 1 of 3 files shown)


kjarosh (approved):

core/src/backend/ui.rs
```diff
@@ -47,6 +47,21 @@ pub struct FileFilter {
     pub mac_type: Option<String>,
 }
 
+impl FileFilter {
+    /// Returns extensions suitable for file dialogs.
+    /// On macOS, uses `mac_type` if available; otherwise uses `extensions` with wildcards stripped.
```
kjarosh: Nit: I'd say it does that when `is_mac` is true and I'd mention that it should be true when the user uses a Mac.

PR #22793, opened 2026-01-16
extension: Remove raw script injection functions

danielhjacobs:
These functions have always been a hack. Before MV3 and its support for ExecutionWorld MAIN, they were necessary to "pierce through the isolated sandbox of web extensions" (more accurately they add a script as a page object which itself has access to page globals) in order to modify global objects like `navigator.plugins` and `window.RufflePlayer`. `ExecutionWorld` `MAIN` is the true intended way of gaining this ability. This will lose us support for Firefox versions prior to 128, which notably includes the last supported Windows 7/8 version of Firefox, Firefox 115 ESR. That is why it's a draft. Firefox 115 ESR will go out of support ~~after February 2026. This date has been repeatedly pushed back, but this is likely the final date, as previous versions of https://support.mozilla.org/en-US/kb/firefox-users-windows-7-8-and-81-moving-extended-support said the date would be re-evaluated closer to it occurring, but the current version says "After this, no security updates will be provided and you are strongly encouraged to upgrade to a supported Microsoft Windows version."~~ It seems reasonable for us to end support for Firefox 115 ESR when Mozilla ends support for it.


danielhjacobs:
New information:

> Updates will be delivered through the ESR channel until the end of August 2026. It was previously announced that support would end in February 2026, but in order to ensure our users can continue to browse the web, we will extend support for Firefox for these operating systems for another six months and then re-evaluate.

Source: https://support.mozilla.org/en-US/kb/firefox-users-windows-7-8-and-81-moving-extended-support

danielhjacobs:
> Updates will be delivered through the ESR channel until March 2027. It was previously announced that support would end in August 2026, but in order to ensure our users can continue to browse the web, we will extend support for Firefox for these operating systems for another six months and then re-evaluate.

PR #22847, opened 2026-01-24
tests: Rename sample_count to quality and set "high" by default

kjarosh:
Before this patch, sample_count basically identified the quality and mapped to it based on the value.  This was a bit non-intuitive and many authors didn't know what sample_count was.

The default value is set to "high" instead of "low" because that's the default in Flash, so if someone provides output from Flash, the quality will match.  As of this patch, all tests have the quality set explicitly, so it doesn't impact any tests.

### DIFF
--- a/tests/tests/regression_tests.rs
+++ b/tests/tests/regression_tests.rs
@@ -91,7 +91,7 @@ fn main() {
 
 fn load_test_dir<'a>(test_dir: &'a VfsPath, name: &'a str) -> impl Iterator<Item = Test> + 'a {
     let options = TestOptions::read_with_subtests(&test_dir.join("test.toml").unwrap())
-        .context("Couldn't load test options")
+        .with_context(|| format!("Couldn't load test options for test {name}"))
         .unwrap();
     options.into_iter().map(move |opts| {
         Test::from_options(opts, test_dir.to_owned(), name.to_owned())


SuchAFuriousDeath:
:sweat_smile:

kjarosh:
> 😅

No worries, that was a long standing issue, a lot of people (including me) didn't know what that meant at first :D

tests/tests/swfs/fonts/embed_matching/fallback_preferences/output.actual-linux-Vulkan.png
SuchAFuriousDeath, line 1: Is this supposed to be commited?

tests/tests/swfs/fonts/embed_matching/fallback_preferences/output.difference-color-linux-Vulkan.png
SuchAFuriousDeath, line 1: And this,?

tests/tests/swfs/fonts/embed_matching/fallback_preferences/output.actual-linux-Vulkan.png
kjarosh: No, it's not!

PR #22879, opened 2026-01-29
chore: Update hashbrown to 0.16.1

SuchAFuriousDeath:
Now there are only 2 versions of hashbrown instead of 3. This also allows implementing the 'add proper entry API?' TODO easily and removing a lot of unsafe code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2044,6 +2044,7 @@ version = "0.5.3"
 source = "git+https://github.com/kyren/gc-arena.git?rev=75671ae03f53718357b741ed4027560f14e90836#75671ae03f53718357b741ed4027560f14e90836"
 dependencies = [
  "gc-arena-derive",
+ "hashbrown 0.16.1",
  "indexmap",
  "slotmap",
  "smallvec",
@@ -2356,16 +2357,6 @@ dependencies = [
  "crunchy",
  "num-traits",
  "zerocopy",
-]
-
-[[package]]
-name = "hashbrown"
-version = "0.14.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e5274423e17b7c9fc20b6e7e208532f9b19825d82dfd615708b70edd83df41f1"
-dependencies = [
- "ahash",
- "allocator-api2",
 ]
 
 [[package]]
@@ -4480,7 +4471,7 @@ version = "0.1.0"
 dependencies = [
  "fnv",
  "gc-arena",
- "hashbrown 0.14.5",
+ "hashbrown 0.16.1",
  "quick-xml",
  "regress",
  "ruffle_macros",
@@ -4518,7 +4509,7 @@ dependencies = [
  "fnv",
  "futures",
  "gc-arena",
- "hashbrown 0.14.5",
+ "hashbrown 0.16.1",
  "id3",
  "image",
  "indexmap",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,7 +99,7 @@ rayon = "1.11.0"
 sha2 = "0.10.9"
 indexmap = "2.13.0"
 fnv = "1.0.7"
-hashbrown = { version = "0.14.5", features = ["raw"] }
+hashbrown = "0.16.1"
 fluent-templates = "0.13.2"
 lzma-rs = "0.3.0"
 tempfile = "3.25.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ egui = "0.33.3"
 clap = { version = "4.5.57", features = ["derive"] }
 cpal = "0.16.0"
 anyhow = "1.0"
-gc-arena = { git = "https://github.com/kyren/gc-arena.git", rev = "75671ae03f53718357b741ed4027560f14e90836", features = ["indexmap", "smallvec", "slotmap"] }
+gc-arena = { git = "https://github.com/kyren/gc-arena.git", rev = "75671ae03f53718357b741ed4027560f14e90836", features = ["hashbrown", "indexmap", "smallvec", "slotmap"] }
 slotmap = "1.1.1"
 async-channel = "2.5.0"
 bitflags = "2.10.0"
--- a/core/common/src/avm_string/interner.rs
+++ b/core/common/src/avm_string/interner.rs
@@ -5,7 +5,7 @@ use super::repr::AvmStringRepr;
 use core::fmt;
 use gc_arena::collect::Trace;
 use gc_arena::{Collect, Gc, GcWeak, Mutation};
-use hashbrown::HashSet;
+use hashbrown::HashTable;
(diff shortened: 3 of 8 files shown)


SuchAFuriousDeath:
Forgot to run tests locally :(

torokati44:
Won't we need to jump straight to 0.16? @moulins 

Screenshot_20260130-111342

torokati44:
Or 0.16.1 even.
Screenshot_20260130-111625

torokati44:
Also `indexmap` might be useful, but I dunno.

Cargo.toml
```diff
@@ -100,7 +100,7 @@ rayon = "1.11.0"
 sha2 = "0.10.9"
 indexmap = "2.12.0"
 fnv = "1.0.7"
-hashbrown = { version = "0.14.5", features = ["raw"] }
+hashbrown = "0.16"
```
torokati44: Oh, the title is outdated then, never mind. (still this could potentially have to be 0.16.1)

SuchAFuriousDeath:
> Won't we need to jump straight to 0.16? @moulins

We do, the description is now inaccurate, because I had to upgrade to 0.16 to make the tests pass because 0.15 does not have the bucket api on HashTable.

Cargo.lock
```diff
@@ -2359,16 +2360,6 @@ dependencies = [
...
-[[package]]
-name = "hashbrown"
-version = "0.14.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e5274423e17b7c9fc20b6e7e208532f9b19825d82dfd615708b70edd83df41f1"
-dependencies = [
- "ahash",
- "allocator-api2",
-]
-
```
torokati44, line 2370: yaay

core/common/src/avm_string/interner.rs
```diff
@@ -126,82 +126,99 @@ impl<'gc> AvmStringInterner<'gc> {
...
+                        return Entry::Occupied(OccupiedEntry(strong));
+                    } else {
+                        occupied.remove();
+                    }
                 }
-            } else {
-                // The entry is stale, delete it.
-                // SAFETY: the entry has already been yielded by the iterator.
-                unsafe { raw.erase(bucket) };
+                Err(_) => return Entry::Vacant(VacantEntry { set: self, hash }),
```
moulins, line 194: This is O(n²) in the number of encountered dead entries, but I don't see a way of improving this while still removing dead entries, so oh well... 🤷

core/src/avm2/dynamic_map.rs
```diff
@@ -26,99 +27,145 @@ pub enum DynamicKey<'gc> {
...
 
 /// A HashMap designed for dynamic properties on an object.
-#[derive(Debug, Clone)]
+///
+/// Uses `HashTable` directly to expose stable bucket indices, which are
+/// needed for correct iteration when entries are added or removed mid-iteration.
+#[derive(Clone)]
 pub struct DynamicMap<K, V> {
-    values: hashbrown::HashMap<K, DynamicProperty<V>, FnvBuildHasher>,
-    // The last index that was given back to flash
```
moulins, line 32: This comment seems to have been eaten?

core/src/avm2/dynamic_map.rs
```diff
@@ -26,99 +27,145 @@ pub enum DynamicKey<'gc> {
...
+    table: HashTable<(K, DynamicProperty<V>)>,
+    hasher: FnvBuildHasher,
     public_index: Cell<usize>,
-    // The actual index that represents where an item is in the HashMap
+    // The actual bucket index that represents where an item is in the table
     real_index: Cell<usize>,
 }
 
-// `gc-arena` doesn't provide a `Collect` impl for the version of `hashbrown` we use.
 unsafe impl<'gc, K: Collect<'gc>, V: Collect<'gc>> Collect<'gc> for DynamicMap<K, V> {
```
moulins: With the latest `gc-arena` changes, this manual impl shouldn't be necessary anymore.

core/src/avm2/dynamic_map.rs
```diff
@@ -26,99 +27,145 @@ pub enum DynamicKey<'gc> {
...
+            }),
+            (true, false) => self.table.iter().for_each(|(k, _)| cc.trace(k)),
+            (false, true) => self.table.iter().for_each(|(_, v)| cc.trace(&v.value)),
+            (false, false) => {}
         }
     }
 }
 
-impl<K: Eq + PartialEq + Hash, V> Default for DynamicMap<K, V> {
+impl<K: Debug, V: Debug> Debug for DynamicMap<K, V> {
```
moulins: Why not `#[derive(Debug)]` instead?

core/src/avm2/dynamic_map.rs
```diff
@@ -26,99 +27,145 @@ pub enum DynamicKey<'gc> {
...
-        self.values.raw_table()
+        None
     }
 
     pub fn insert(&mut self, key: K, new_value: V) {
-        match self.entry(key) {
-            Entry::Occupied(mut occupied) => {
+        let hash = self.hash_key(&key);
+
+        match self.table.find_mut(hash, |(k, _)| *k == key) {
```
moulins: Why isn't this using `self.table.entry(...)`?

core/src/avm2/dynamic_map.rs
```diff
@@ -26,99 +27,145 @@ pub enum DynamicKey<'gc> {
...
-        self.values.remove(key)
+        let hash = self.hash_key(key);
+
+        match self.table.find_entry(hash, |(k, _)| k == key) {
+            Ok(occupied) => Some(occupied.remove().0.1),
+            Err(_) => None,
+        }
     }
 
     pub fn next(&self, index: usize) -> Option<usize> {
```
moulins, line 155: Note that hashbrown dropped the guarantee that removing an already-iterated element doesn't disturb existing bucket indices (see [here](https://docs.rs/hashbrown/latest/hashbrown/struct.HashTable.html#method.find_bucket_index)).

The actual implementation hasn't changed so our iteration logic should still work, but we should try to not depend on it in the long term. In the meantime, a comment mentioning this would be worthwhile, I think.

core/common/src/avm_string/interner.rs
SuchAFuriousDeath: Trust me, I tried many times :D

core/src/avm2/dynamic_map.rs
SuchAFuriousDeath: It exposes the hasher, I was assuming we don't want. AFAIK, the Debug impl on the std hashmap does not do that. But up to you.

core/src/avm2/dynamic_map.rs
SuchAFuriousDeath: I haven't checked, but doesn't flash also have this issue?

moulins (approved):

PR #22889, opened 2026-01-31
core: Fix Fullscreen reporting wrong displayState

Fancy2209:
Ruffle allows kb input, so it's FULLSCREEN_INTERACTIVE
This is used on Matt's Hidden Cats


kjarosh:
We'll need some tests for that.

PR #22896, opened 2026-02-01
avm1: Improve loading a movie into root

kjarosh:
This contains several patches that improve how Ruffle behaves when loading a movie into _root. The basis for tests constitute the loader listener events, using which we can observe various states of the movie during loading.


SuchAFuriousDeath (approved):

core/src/loader.rs
```diff
@@ -687,93 +694,136 @@ impl<'gc> MovieLoader<'gc> {
...
+        loader_url: Option<String>,
+        body: Vec<u8>,
+        url: String,
+        status: u16,
+        redirected: bool,
+    ) -> Result<(), Error> {
+        ContentType::sniff(&body).expect(ContentType::Swf)?;
+
+        let movie = SwfMovie::from_data(&body, url, loader_url)?;
+        player.lock().unwrap().mutate_with_update_context(|uc| {
```
SuchAFuriousDeath: Nit: All 3 functions do player.lock().unwrap(). Maybe it would be better to do it at the call site and pass the resulting guard into all the functions.

kjarosh: Done

PR #22902, opened 2026-02-01
refactor: Use iterator instead of collecting for known_movies

SuchAFuriousDeath:


SuchAFuriousDeath:
Is this really A-desktop? :sweat_smile:

kjarosh:
It's debug UI, which currently is only desktop :D

SuchAFuriousDeath:
Oh.

kjarosh:
In theory it could be core, but I guess it's kind of a gray zone. On one hand it's not in desktop per se, but it only affects desktop, and the point of using A- and T- labels in PRs is for release notes and blog posts.

PR #22908, opened 2026-02-01
core: Wrap Library and SwfMovie in Gc

jarca0123:
Tested on AQWorlds, and deallocates memory on repeated map opening and closing, which the current nightly does not deallocate. Well, third time's a charm.

### DIFF
--- a/core/common/src/tag_utils.rs
+++ b/core/common/src/tag_utils.rs
@@ -1,8 +1,7 @@
 use crate::sandbox::SandboxType;
 
-use gc_arena::Collect;
+use gc_arena::{Collect, Gc};
 use std::fmt::{Debug, Formatter};
-use std::sync::Arc;
 use swf::{Fixed8, HeaderExt, Rectangle, Twips};
 use url::Url;
 
@@ -332,16 +331,16 @@ impl Debug for SwfMovie {
 }
 
 /// A shared-ownership reference to some portion of an SWF datastream.
-#[derive(Debug, Clone, Collect)]
+#[derive(Debug, Clone, Copy, Collect)]
 #[collect(no_drop)]
-pub struct SwfSlice {
-    pub movie: Arc<SwfMovie>,
+pub struct SwfSlice<'gc> {
+    pub movie: Gc<'gc, SwfMovie>,
     pub start: usize,
     pub end: usize,
 }
 
-impl From<Arc<SwfMovie>> for SwfSlice {
-    fn from(movie: Arc<SwfMovie>) -> Self {
+impl<'gc> From<Gc<'gc, SwfMovie>> for SwfSlice<'gc> {
+    fn from(movie: Gc<'gc, SwfMovie>) -> Self {
         let end = movie.data().len();
 
         Self {
@@ -352,17 +351,17 @@ impl From<Arc<SwfMovie>> for SwfSlice {
     }
 }
 
-impl AsRef<[u8]> for SwfSlice {
+impl<'gc> AsRef<[u8]> for SwfSlice<'gc> {
     #[inline]
     fn as_ref(&self) -> &[u8] {
         self.data()
     }
 }
 
-impl SwfSlice {
+impl<'gc> SwfSlice<'gc> {
     /// Creates an empty SwfSlice.
     #[inline]
-    pub fn empty(movie: Arc<SwfMovie>) -> Self {
+    pub fn empty(movie: Gc<'gc, SwfMovie>) -> Self {
         Self {
             movie,
             start: 0,
@@ -373,7 +372,7 @@ impl SwfSlice {
     /// Creates an empty SwfSlice of the same movie.
     #[inline]
     pub fn copy_empty(&self) -> Self {
-        Self::empty(self.movie.clone())
+        Self::empty(self.movie)
     }
 
     /// Construct a new SwfSlice from a regular slice.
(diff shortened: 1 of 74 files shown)


core/src/avm2/globals/flash/display/Loader.as
```diff
@@ -30,14 +28,9 @@ package flash.display {
         public native function unload():void;
 
         [API("662")]
-        public function unloadAndStop(gc:Boolean = true):void {
-            stub_method("flash.display.Loader", "unloadAndStop");
-            this.unload();
-        }
+        public native function unloadAndStop(gc:Boolean = true):void;
```
kjarosh: Looks like it could be implemented in a separate PR?

core/src/context.rs
```diff
@@ -455,6 +460,12 @@ impl<'gc> UpdateContext<'gc> {
         self.gc_context
     }
 
+    /// Convenience method to retrieve the root SWF movie.
+    #[inline(always)]
+    pub fn root_movie(&self) -> Gc<'gc, SwfMovie> {
```
kjarosh: We can introduce this method in a separate PR to minimize the diff. It's also weird we have this `root_swf`/`root_movie` inconsistency, I'd just pick one name and use it in both contexts.

core/src/debug_ui.rs
```diff
@@ -115,8 +114,9 @@ impl DebugUi {
                     self.movies.insert(movie, Default::default());
                 }
                 Message::TrackTopLevelMovie => {
-                    self.movies
-                        .insert(context.root_swf.clone(), Default::default());
+                    let movie = *context.root_swf;
```
kjarosh, line 117: Can't we use the newly introduced method here?

core/common/src/tag_utils.rs
```diff
@@ -305,16 +304,16 @@ impl Debug for SwfMovie {
 }
 
 /// A shared-ownership reference to some portion of an SWF datastream.
-#[derive(Debug, Clone, Collect)]
+#[derive(Debug, Clone, Copy, Collect)]
 #[collect(no_drop)]
-pub struct SwfSlice {
-    pub movie: Arc<SwfMovie>,
+pub struct SwfSlice<'gc> {
+    pub movie: Gc<'gc, SwfMovie>,
```
kjarosh, line 337: I think it makes sense to introduce a new type for `Gc<'gc, SwfMovie>` to simplify types.

core/src/player.rs
```diff
@@ -2380,7 +2389,7 @@ impl Player {
         });
         self.update_mouse_state(EnumSet::empty(), false, &mut false);
 
-        // GC
+        // GC - with Gc<SwfMovie>, cleanup happens automatically when unreachable
```
kjarosh: What's the point of this comment here? It looks random

core/src/player.rs
```diff
@@ -2984,6 +2995,7 @@ impl PlayerBuilder {
         let language = ui.language();
 
         // Instantiate the player.
+        // Player.swf stays as Arc for external access; a Gc copy is created inside the arena.
         let fake_movie = Arc::new(SwfMovie::empty(player_version, None));
```
kjarosh: Looks like it can be removed. You'd need to refactor the code from `should_reset_highlight` to get rid of it, but can be done by accepting update context and taking the root movie from there.

I think we need to test it and check which SWF version is used for that behavior.

Nonetheless, it could be done in a separate PR.

core/src/display_object/movie_clip.rs
```diff
@@ -4435,11 +4541,22 @@ impl<'gc, 'a> MovieClip<'gc> {
...
-                let slice = self.0.shared.get().swf.to_start_and_end(
+                let swf_slice = self.0.shared.get().swf.to_start_and_end(
                     self.0.tag_stream_pos.get() as usize,
                     self.0.tag_stream_len(),
                 );
-                Some(context.start_stream(self, self.0.current_frame(), slice, stream_info))
+                // Convert SwfSlice to AudioSlice for thread-safe audio processing
+                let movie_data: Arc<[u8]> = Arc::from(swf_slice.movie.data());
+                let audio_slice =
+                    AudioSlice::new(movie_data).to_start_and_end(swf_slice.start, swf_slice.end);
```
moulins: I'm a little concerned about this: it will make a copy of the full SWF data for every `sound_stream_block` in the SWF (imagine a 50MB SWF with one hundred 10kB blocks).

Ideally we wouldn't copy any data here, but only copying the block's data (e.g. `AudioSlice::new(Arc::from(swf_slice.data()))`) would already be much better.

core/src/avm2/class.rs
```diff
@@ -88,6 +90,15 @@ impl fmt::Debug for Allocator {
     }
 }
 
+/// Link from an AVM2 class to its SymbolClass-registered library character.
+/// Uses `GcWeak` so this does NOT keep the library alive.
+#[derive(Copy, Clone, Collect)]
+#[collect(no_drop)]
+struct SymbolClassLink<'gc> {
+    character_id: CharacterId,
+    library: GcWeak<'gc, RefLock<MovieLibrary<'gc>>>,
```
Lord-McSweeney: We should not have references to `core` in the AVM2. Is there any way to go about implementing this without this change?

jarca0123: Is implementing it with "SwfMovie" (which is in the ruffle_common crate) enough? Just want to be sure.

Lord-McSweeney: Yes, that would work

PR #22911, opened 2026-02-01
chore: Update vergen to 9.1.0

SuchAFuriousDeath:
https://github.com/rustyhorde/vergen/blob/master/MIGRATING_v8_to_v9.md


kjarosh:
Just to make sure, can you test that with flatpak? https://github.com/flathub/rs.ruffle.Ruffle

SuchAFuriousDeath:
Tested it and it works.

PR #22920, opened 2026-02-02
avm1: Add a native object for Color

kjarosh:
Looks like Flash uses a native object for Color, but we're yet to discover its use.


kjarosh:
Actually, I noticed that `Color` behaves a bit differently compared to other native objects, and the issue might be related more to differences in behavior for the prototype and not native objects.

kjarosh:
Closing, see https://github.com/ruffle-rs/ruffle/pull/23130 for details.

PR #22761, opened 2026-01-13
avm2: Implement Dxns and DxnsLate opcodes

SuchAFuriousDeath:
Another specimen from the graveyard.

Fixes #13807, supersedes #21014

PR #22923, opened 2026-02-02
avm2: Fix StaticText.text output

kjarosh:
Take edge cases into account, such as missing font.


SuchAFuriousDeath:
@kjarosh I'll be fixing the bytes error in #22761, it should be getting merged soon.

Issue #22931, opened 2026-02-03
With italics, glyphs get cut from the right with Canvas device font Renderer


hamzaelbarmaki1994:
### Describe the bug

<img width="577" height="70" alt="Image" src="https://github.com/user-attachments/assets/7f8fa6d8-30a2-427e-ac69-aa3860cd3178" />

### Expected behavior

<img width="560" height="82" alt="Image" src="https://github.com/user-attachments/assets/c6734b7a-5183-452a-8dca-c73f1ffe82b8" />

### Content Location

https://bibliotheque.sesamath.net/public/voir/325
local:
[6N4s2ex4.zip](https://github.com/user-attachments/files/25045402/6N4s2ex4.zip)

### Affected platform

Browser's extension

### Operating system

windows10

### Browser

opera

### Additional information

Thanks for help

PR #22943, opened 2026-02-04
web: Support glyphs protruding horizontally in canvas font renderer

kjarosh:
* Fixes https://github.com/ruffle-rs/ruffle/issues/22931

Some glyphs can protrude outside of the bounding box of the glyph, e.g. when the text is italic.

### DIFF
--- a/web/src/ui/font_renderer.rs
+++ b/web/src/ui/font_renderer.rs
@@ -101,10 +101,15 @@ impl CanvasFontRenderer {
 
     fn render_glyph_internal(&self, character: char) -> Result<Glyph, JsValue> {
         let text = &character.to_string();
-        let width = self.calculate_width(text)?;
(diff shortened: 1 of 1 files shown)


torokati44:
Is it for sure that this can only happen horizontally? (For italic, probably, but for any other çûrsęd reason....?)

kjarosh:
> Is it for sure that this can only happen horizontally? (For italic, probably, but for any other çûrsęd reason....?)

I'll look into it

kjarosh:
> Is it for sure that this can only happen horizontally? (For italic, probably, but for any other çûrsęd reason....?)

I'm pretty sure it can happen vertically too, the canvas API even has methods dedicated for that, but it's not obvious how we should handle that, I've never seen FP handle vertical protrusions, so I'll leave that for a future PR.

PR #22953, opened 2026-02-05
tests: Convert ignored int/uint tests to known failures

danielhjacobs:


danielhjacobs:
These are the output of a diff between the Flash expected output (left) and the actual Ruffle output (right):

[uint_toprecision.txt](https://github.com/user-attachments/files/25111775/uint_toprecision.txt)
[int_toprecision.txt](https://github.com/user-attachments/files/25111776/int_toprecision.txt)
[uint_toexponential.txt](https://github.com/user-attachments/files/25111777/uint_toexponential.txt)
[int_toexponential.txt](https://github.com/user-attachments/files/25111778/int_toexponential.txt)

PR #22954, opened 2026-02-05
tests: Make ignored test Number.toString a known failure

danielhjacobs:
See https://github.com/ruffle-rs/ruffle/commit/9e62b05157ea59e0b611ecbc375bdeacd01bfe5e#diff-2f67e7761cd58716b1a25912901de27f8fa28f5316a3d8bb06869a27438c5e29


danielhjacobs:
This is the output of a diff between the Flash expected output (left) and the actual Ruffle output (right):

[number_tostring.txt](https://github.com/user-attachments/files/25112087/number_tostring.txt)

For this specific test I suggest reading https://github.com/danielhjacobs/ruffle/blob/2f8cf9bbb7a88fbb96f62a017ff69157c641aa87/tests/tests/swfs/avm2/number_tostring/test.toml

kjarosh (approved):

kjarosh:
It's possible this test is OS-specific, I would check FP's output on different platforms.

tests/tests/swfs/avm2/number_tostring/test.toml
```diff
@@ -1,2 +1,6 @@
 num_frames = 1
-ignore = true # Ignored because Flash Player adds extra x, W, and/or ° symbols randomly
+# This test has two issues per the PR that introduced it:
+# 1. Precision limits that are specific to the chosen radix are not properly emulated
+# 2. Flash produces occasional and intermittent corruption in the resulting 0 padding; usually manifesting as `x`, `W`, or `°` characters
+# Issue 2 means the output.txt does not consistently match what Flash actual produces, but unless we emulate the corruption, Ruffle's output should be consistent.
```
kjarosh: Does it mean that flash produces different output per run?

PR #22955, opened 2026-02-05
tests: Increase pixelsnapping outliers for different platforms

danielhjacobs:


torokati44:
_At the risk of committing "while you're there"-ism_, I just remembered, we're holding back lyon because is slightly alters some outlines, failing visual tests with current expected outputs/tolerances: https://github.com/torokati44/ruffle/commit/33840417a40c6f249ffc4f716980cc71dea31bf4
See also: https://github.com/ruffle-rs/ruffle/pull/21758#issuecomment-3325635230

danielhjacobs:
I committed a new draft PR to update lyon to check how many visual tests fail from that and what the differences are.

danielhjacobs:
#22961 is now ready for review for the lyon update

PR #22830, opened 2026-01-21
avm2: Throw correct ReferenceError for E4X wildcard on objects

SuchAFuriousDeath:

PR #22962, opened 2026-02-06
avm2: Refactor stack argument retrieval

SuchAFuriousDeath:


core/src/avm2/stack.rs
```diff
@@ -147,21 +147,30 @@ impl<'a, 'gc> StackFrame<'a, 'gc> {
         self.data[index].set(value);
     }
 
-    #[inline(always)]
-    pub fn get_args(&self, num_args: usize) -> FunctionArgs<'a, 'gc> {
-        let base = self.stack_pointer.get() - num_args;
+    fn get_args_internal(&self, arg_count: u32) -> &'a [Cell<Value<'gc>>] {
```
Lord-McSweeney: Can you mark this as `#[inline(always)]`?

core/src/avm2/activation.rs
```diff
@@ -567,7 +567,10 @@ impl<'a, 'gc> Activation<'a, 'gc> {
     /// Pops multiple values off the operand stack.
     #[inline]
     #[must_use]
-    pub fn pop_stack_args(&self, arg_count: u32) -> Vec<Value<'gc>> {
+    pub fn pop_stack_args<C>(&self, arg_count: u32) -> C
```
Lord-McSweeney, line 571: Since there are only two places left where this is used, could you just replace those places with calls to `self.stack.pop_args` and remove this method?

SuchAFuriousDeath: Sure.

core/src/avm2/stack.rs
SuchAFuriousDeath: Sure. I was thinking whether set_stack_pointer should also be inline.

SuchAFuriousDeath:
@Lord-McSweeney Could you also check out https://github.com/ruffle-rs/ruffle/pull/22830? You added yourself as a reviewer some time ago. It's a fairly small PR.

Lord-McSweeney (approved):

adrian17 (approved):

PR #22975, opened 2026-02-08
render: Refactor FloatOrInt to store vec4 arrays

SuchAFuriousDeath:
What was enforced by the assert can be enforced using the typesystem.

### DIFF
--- a/render/wgpu/src/pixel_bender.rs
+++ b/render/wgpu/src/pixel_bender.rs
@@ -511,69 +511,68 @@ pub(super) fn run_pixelbender_shader_impl(
 
                 #[derive(Debug)]
                 enum FloatOrInt {
-                    Float(SmallVec<[f32; 4]>),
-                    Int(SmallVec<[i32; 4]>),
+                    Float(SmallVec<[[f32; 4]; 1]>),
+                    Int([i32; 4]),
                 }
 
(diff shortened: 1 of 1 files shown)


render/wgpu/src/pixel_bender.rs
```diff
@@ -511,12 +511,12 @@ pub(super) fn run_pixelbender_shader_impl(
 
                 #[derive(Debug)]
                 enum FloatOrInt {
-                    Float(SmallVec<[f32; 4]>),
-                    Int(SmallVec<[i32; 4]>),
+                    Float(SmallVec<[[f32; 4]; 1]>),
+                    Int(SmallVec<[[i32; 4]; 1]>),
```
Lord-McSweeney: If you can't have multiple vec4s of ints in a `FloatOrInt`, could the `Int` variant be changed to just `Int([i32; 4])`?

SuchAFuriousDeath: :thinking: Good point.

SuchAFuriousDeath:
@kjarosh :( Can you maybe rerun the failing job? Maybe it will succeed.

SuchAFuriousDeath:
Oh wait, it's security vulnerabilities, so it will fail again, nevermind.

PR #22994, opened 2026-02-09
desktop: Improve logging on desktop a bit

kjarosh:
This PR (1) adds logging for the content being opened to make sure Ruffle opens the right URL and (2) reduces font registration log level to debug.

1. It's useful e.g. in flatpak where we'd like to know quickly what's the path Ruffle tried to load. Due to filesystem isolation the path can be different than the one selected. See https://github.com/flathub/rs.ruffle.Ruffle/issues/38.
2. We didn't have too many issues related to GUI fonts and these logs are a bit spammy, so set them to debug and in case they are needed, they can be enabled manually.

### DIFF
--- a/frontend-utils/src/content.rs
+++ b/frontend-utils/src/content.rs
@@ -36,6 +36,20 @@ impl ContentDescriptor {
             root_content_path,
         })
     }
+
+    pub fn describe(&self) -> String {
+        #[cfg(not(feature = "fs"))]
+        {
+            format!("{}", self.url)
+        }
+
+        #[cfg(feature = "fs")]
+        if let Some(dir) = &self.root_content_path {
+            format!("{} within {}", self.url, dir.display())
+        } else {
+            format!("{}", self.url)
+        }
+    }
 }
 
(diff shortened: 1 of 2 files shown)


frontend-utils/src/content.rs
```diff
@@ -36,6 +36,20 @@ impl ContentDescriptor {
...
+        }
+
+        #[cfg(feature = "fs")]
+        if let Some(dir) = &self.root_content_path {
+            format!("{} within {}", self.url, dir.to_string_lossy())
+        } else {
+            format!("{}", self.url)
+        }
+    }
 }
```
moulins, line 53: Nit: this could return `fmt::Arguments<'_>` instead (with `format_args!(...)`).

Also, `dir.to_string_lossy()` could be `dir.display()`.

kjarosh: Used `.display()`, but using `fmt::Arguments<'_>` is not that simple as the borrow checker complains. I left it as-is, it's printed once per main content load anyway.

PR #22996, opened 2026-02-10
tests: Port AVM1 `Math` tests to our test framework

moulins:
This was one of the last AVM1 tests implemented directly in Rust.

Changes:
- Add `approximations.bare_numbers` test config key, to disable approximations on number-only lines
- Remove SWFv5 tests, as MTASC cannot compile them (and FP doesn't appear to have any
  SWFv5-specific behavior here)
- Test more edge-cases, exposing some incorrect Ruffle output


tests/framework/src/options/approximations.rs
```diff
@@ -1,16 +1,23 @@
...
+use anyhow::bail;
 use approx::relative_eq;
 use regex::Regex;
 use serde::Deserialize;
 
 #[derive(Clone, Deserialize, Default)]
 #[serde(default, deny_unknown_fields)]
 pub struct Approximations {
+    #[serde(default = "default_for_bare_numbers")]
+    pub bare_numbers: bool,
```
kjarosh, line 10: Do we want to make it true by default? TBH I wasn't aware that the test framework will compare bare numbers in addition to patterns.

moulins: Switching it to false-by-default would mean changing a lot of existing tests; it's probably best left to a separate PR if we decide this is something we want to do.

tests/tests/swfs/avm1/math_swf8/Test.as
```diff
@@ -0,0 +1,123 @@
...
+    trace(invalid());
+    trace("// ASnative(200, 50)({ v: 3 }, { v: 2 }, { v: 1 })");
+    trace(invalid(noisy(3), noisy(2), noisy(1)));
+
+    fscommand("quit");
+  }
+
+  static var NOISY_VALUE_OF = function() {
+    trace("valueOf called: " + this.v);
+    return this.v;
```
kjarosh, line 106: Can you add test cases where this throws?

moulins: Throwing coercions are already tested in `avm1/displacementmapfilter_mappoint_throw_error`, so I don't think this is necessary (in particular, `Math` has no internal state that could be partially modified before one of the coercions throws).

kjarosh: I don't see the connection, you have setters and different impls there. I'm not talking about testing that coercions throw exceptions (we know they do), but how the given methods handle them.

How do you know how exceptions should be handled in Math? Will the second coercion run before the first is rethrown? If both throw, which one is rethrown? Does it rethrow at all?

moulins: It would be weird if it did anything else than immediately rethrow the first exception, but I guess I can add a test for that

PR #22999, opened 2026-02-10
avm1: Migrate `Math` to `ASnative`-style declarations and fix some edge cases.

moulins:
This PR is stacked on top of #22996.

Behavioral changes in `Math.*` methods:
- The first two arguments are always coerced, even when unused.
- Invalid `ASnative` method indices return NaN, not undefined.
- Math.min/max returns +/-inf when called without arguments.
- In SWFv6 and below, an invalid arg count (almost) always returns NaN.


moulins:
The `from_gnash/actionscript.all/ASnative-v*` tests now return different results on Ruffle-MacOS compared to other platforms (due to floating-point inaccuracies), so they have been updated to not check the actual output against the expected `output.ruffle.txt` on MacOS.

This necessitated some changes to the test framework `test.toml` format:
- Added an optional `filter: TestExpression` key that will ignore the test if the given expression evaluates to `false`;
- Fixed a bug in subtests config merging where some values weren't overridden properly.

tests/tests/swfs/from_gnash/actionscript.all/ASnative-v8/test.toml
```diff
@@ -1,2 +1,11 @@
 num_frames = 30
 known_failure = true
+
+[subtests.default]
+filter = 'not(os = "macos")'
+
+# Math.* methods return slightly different results on macOS, and the way
+# Gnash tests are structured doesn't allow the use of approximate checks.
+[subtests.macos]
+filter = 'os = "macos"'
```
SuchAFuriousDeath, line 10: Seems a bit weird having this right under [subtests.macos] 🤔  But other than that, the changes are fine.

PR #23000, opened 2026-02-10
avm1: Make XmlSocket.connect(null) behave like connect()

### LINKED ISSUE #22995: Issue when providing null for host in XMLSocket.connect()
### Describe the bug

Normally, according to the documentation, providing null/undefined for the host parameter in `XMLSocket.connect()` should resolve to the IP of whatever URL the SWF was loaded from. In Ruffle, this doesn't seem to (at least entirely) be the case. It will try to literally connect to "null" or "undefined" and obviously fail.

### Expected behavior

In this test, when the connection is successful, it will show a blue rectangle and trace "onConnect success: true". Otherwise, it will show a red rectangle and trace "onConnect success: false". Currently, Ruffle will always fail, but should always succeed (as long as the connection is available).

### Content Location

In this zip, along with the swf, is a Node.js script for hosting the swf file at 127.0.0.1:24105 and a net server at 127.0.0.1:24106 for the swf to connect to.

Just run `node main` in the directory for these two files (make sure they're kept in the same directory), then open the SWF from `http://127.0.0.1:24105/`.

[xmlsocket_null_host.zip](https://github.com/user-attachments/files/25197037/xmlsocket_null_host.zip)

### Affected platform

Desktop app

### Operating system

macOS Tahoe 26.2

### Browser

_No response_

### Additional information

_No response_

adrian17:
Fixes https://github.com/ruffle-rs/ruffle/issues/22995 .

Also added logging to a failure case (avm2 implementation has the same logging), but didn't change the behavior.

### DIFF
--- a/core/src/avm1/globals/xml_socket.rs
+++ b/core/src/avm1/globals/xml_socket.rs
@@ -123,26 +123,28 @@ pub fn connect<'gc>(
     args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
     if XmlSocket::cast(this.into()).is_some() {
-        let host = args
-            .get(0)
-            .copied()
-            .unwrap_or_else(|| {
-                let movie = activation.base_clip().movie();
+        let host = args.get(0).copied().unwrap_or(Value::Null);
+        let host = if matches!(host, Value::Null | Value::Undefined) {
+            let movie = activation.base_clip().movie();
 
-                if let Ok(url) = url::Url::parse(movie.url()) {
-                    if url.scheme() == "file" {
-                        istr!("localhost").into()
-                    } else if let Some(domain) = url.domain() {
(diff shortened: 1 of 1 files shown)


kjarosh:
Can we add a test for it?

ChrisCPI:
Does this also fix when `undefined` is provided instead of `null`?

Edit: Assuming yes because of the line: `let host = if matches!(host, Value::Null | Value::Undefined)`

adrian17:
> Can we add a test for it?

I don't really know how to :c

adrian17:
@kjarosh do you consider the lack of a test to be blocking?

kjarosh:
Sorry about the delay, I planned to take a look and craft a simple test. Not having a test can introduce regressions in the future, I don't think it's blocking though.

SuchAFuriousDeath:
Superseded by #23931.

PR #23002, opened 2026-02-10
core/avm2: Misc. cleanups

Lord-McSweeney:


AllinolCP:
while you're cleaning up, I also suggest to remove mouse_pick_avm1 in loader_display

core/src/avm2/object/loaderinfo_object.rs
```diff
@@ -96,8 +93,6 @@ pub struct LoaderInfoObjectData<'gc> {
 
     uncaught_error_events: Object<'gc>,
 
-    cached_avm1movie: Lock<Option<Object<'gc>>>,
```
adrian17, line 99: what was it doing before?

Lord-McSweeney: This was used before #21528 to support `AVM1Movie`

PR #23006, opened 2026-02-11
core: Optimize CPU bitmap blending and copy operations

jarca0123:
Per future CONTRIBUTING.md, I want to state that this code was generated by Claude Code, specifically Claude Opus 4.6. My workflow is iteratively profiling Ruffle with various SWFs, feeding the profiler output to the LLM, having it identify hotspots and generate optimizations, then re-profiling to verify the improvement. This is one of many PRs I intend to submit that were produced this way.

I have reviewed the code to the best of my ability, though in full transparency, I have not audited every individual line in detail.

EDIT: Here is a description of what it does: This replaces per-pixel get/set_pixel32_raw loops with row-based slice operations: memcpy for copies, and a two-lane u32 blend with div255 bit-trick for alpha compositing.

Here are some benchmarks:
| Benchmark | This PR (ms) | Nightly (ms) | Change |                                                                                                                                                                                                                                                                         
  |---|---|---|---|
  | copypixels_merge_alpha | 468 | 2459 | -81.0% |                                                                                                                                                                                                                                                                          
  | copypixels_merge_alpha_small | 33 | 168 | -80.4% |
  | colortransform_alpha_tint | 388 | 360 | +7.8% |                                                                                                                                                                                                                                                                         
  | sprite_render_pipeline | 420 | 534 | -21.3% |
  | draw_alpha_blend | 1076 | 1694 | -36.5% |
  | draw_alpha_matrix_blend | 1499 | 1639 | -8.5% |
  | particle_fade_composite | 113 | 131 | -13.7% |
  | multi_sprite_composite | 195 | 817 | -76.1% |
  | locked_buffer_composite | 32 | 158 | -79.7% |

Excerpt from my "comprehensive" SWF benchmarks (however I can share the full SWF if you want):
```actionscript                                                                                                                                                                                                                                                                                                           
  private function benchCopyPixelsMergeAlpha(n:int):void {
      var src:BitmapData = new BitmapData(256, 256, true, 0x80FF0000); // 50% alpha red                                                                                                                                                                                                                                     
      var dst:BitmapData = new BitmapData(256, 256, true, 0xFF000000); // opaque black                                                                                                                                                                                                                                      
      var rect:Rectangle = new Rectangle(0, 0, 256, 256);                                                                                                                                                                                                                                                                   
      var pt:Point = new Point(0, 0);
      for (var i:int = 0; i < n; i++) {
          dst.copyPixels(src, rect, pt, null, null, true); // mergeAlpha=true
      }
      src.dispose();
      dst.dispose();
  }
```

### DIFF
--- a/core/src/bitmap/bitmap_data.rs
+++ b/core/src/bitmap/bitmap_data.rs
@@ -45,7 +45,9 @@ impl LehmerRng {
 /// unmultiplied values. Make sure to convert the color to the correct form beforehand.
 // TODO: Maybe split the type into `PremultipliedColor(u32)` and
 //   `UnmultipliedColor(u32)`?
-#[derive(Debug, Default, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, bytemuck::NoUninit)]
+#[derive(
+    Debug, Default, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, bytemuck::Pod, bytemuck::Zeroable,
+)]
 #[repr(C, align(4))]
 pub struct Color {
     // Note: even though AS2/AS3 represent colors as (little-endian) BGRA `u32`s, this is stored
@@ -161,13 +163,19 @@ impl Color {
     /// * `self` - Must be in premultiplied form.
     /// * `source` - Must be in premultiplied form.
     #[must_use]
+    #[inline(always)]
     pub fn blend_over(&self, source: &Self) -> Self {
-        let sa = source.alpha();
+        let inv_sa = 255 - source.alpha() as u16;
 
-        let r = source.red() + ((self.red() as u16 * (255 - sa as u16)) / 255) as u8;
-        let g = source.green() + ((self.green() as u16 * (255 - sa as u16)) / 255) as u8;
-        let b = source.blue() + ((self.blue() as u16 * (255 - sa as u16)) / 255) as u8;
-        let a = source.alpha() + ((self.alpha() as u16 * (255 - sa as u16)) / 255) as u8;
+        #[inline(always)]
+        fn scale(channel: u8, factor: u16) -> u8 {
+            (channel as u16 * factor / 255) as u8
+        }
+
+        let r = source.red() + scale(self.red(), inv_sa);
+        let g = source.green() + scale(self.green(), inv_sa);
+        let b = source.blue() + scale(self.blue(), inv_sa);
+        let a = source.alpha() + scale(self.alpha(), inv_sa);
         Self::rgba(r, g, b, a)
     }
 
--- a/core/src/bitmap/operations.rs
+++ b/core/src/bitmap/operations.rs
@@ -1292,6 +1292,83 @@ pub fn apply_filter<'gc>(
     }
 }
 
+/// Blend a single pixel (src-over, premultiplied alpha, two-lane u32 trick).
+#[inline(always)]
+fn blend_pixel(d: u32, s: u32) -> u32 {
+    #[cfg(target_endian = "little")]
+    let inv_a = 255 - (s >> 24);
+    #[cfg(target_endian = "big")]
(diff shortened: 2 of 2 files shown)


core/src/bitmap/bitmap_data.rs
```diff
@@ -161,13 +163,29 @@ impl Color {
...
+        let inv_sa = 255 - source.alpha() as u16;
 
-        let r = source.red() + ((self.red() as u16 * (255 - sa as u16)) / 255) as u8;
-        let g = source.green() + ((self.green() as u16 * (255 - sa as u16)) / 255) as u8;
-        let b = source.blue() + ((self.blue() as u16 * (255 - sa as u16)) / 255) as u8;
-        let a = source.alpha() + ((self.alpha() as u16 * (255 - sa as u16)) / 255) as u8;
+        // Fast division by 255: (x + 1 + (x >> 8)) >> 8
+        // Exact for all values in 0..=65025 (255*255).
+        #[inline(always)]
+        fn div255(x: u16) -> u8 {
```
kjarosh: Why do you think this is better? It produces more complicated assembly:
```
regular_div:
        movzx   eax, di
        imul    eax, eax, 32897
        shr     eax, 23
        ret

div255:
        mov     eax, edi
        movzx   ecx, ah
        add     eax, ecx
        inc     eax
        shr     eax, 8
        ret
```

I would imagine the compiler knows how to optimize a division by 255, why is it an issue? Which architecture are you optimizing for?

core/src/bitmap/bitmap_data.rs
```diff
@@ -161,13 +163,29 @@ impl Color {
...
+            ((x + 1 + (x >> 8)) >> 8) as u8
+        }
+
+        let r = source
+            .red()
+            .wrapping_add(div255(self.red() as u16 * inv_sa));
+        let g = source
+            .green()
+            .wrapping_add(div255(self.green() as u16 * inv_sa));
```
kjarosh: Why are you changing it to `wrapping_add`? Are we expecting an overflow? The assembly is identical (regular addition is wrapping in release mode, it adds overflow checks in debug mode).

core/src/bitmap/bitmap_data.rs
jarca0123: I am optimizing for no specific architecture in mind, I want to take every target into account. However, this truly is an oversight that I am reverting.

core/src/bitmap/bitmap_data.rs
jarca0123: Ooh, didn't know that. Reverting this too.

core/src/bitmap/operations.rs
```diff
@@ -1284,6 +1284,66 @@ pub fn apply_filter<'gc>(
     }
 }
 
+/// Blend a single pixel (src-over, premultiplied alpha, two-lane u32 trick).
```
Lord-McSweeney, line 1295: Is there a difference between this method and `Color::blend_over`?

core/src/bitmap/bitmap_data.rs
kjarosh: If you really want to extract a function, you can do it in such a way that will prevent us from casting integers in every line, e.g.

```rust
let r = source.red() + scale(self.red(), inv_sa);
```

n0samu:
> (however I can share the full SWF if you want):

I do want, yes 😄

jarca0123:
> > (however I can share the full SWF if you want):
> 
> I do want, yes 😄

Well, in order to keep everything in one place, here are the benchmarks:
[test.zip](https://github.com/user-attachments/files/25320121/test.zip)

Also, I found a quicker algorithm, so I'll update the PR.

Lord-McSweeney:
@jarca0123 does this still improve perf after #24490?

PR #23024, opened 2026-02-13
tests: Make bare_numbers = false the default

kjarosh:
It's better to be explicit than implicit.  Before this patch, by default approximations were applied to lines with only a number.  Now, in order to enable this behavior, bare_numbers = true has to be set explicitly.

### DIFF
--- a/tests/framework/src/options/approximations.rs
+++ b/tests/framework/src/options/approximations.rs
@@ -6,16 +6,11 @@ use serde::Deserialize;
 #[derive(Clone, Deserialize, Default)]
 #[serde(default, deny_unknown_fields)]
 pub struct Approximations {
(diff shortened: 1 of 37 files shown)


tests/framework/src/options/approximations.rs
```diff
@@ -15,7 +15,7 @@ pub struct Approximations {
 
 // Serde requires defaults to be provided through functions
 fn default_for_bare_numbers() -> bool {
-    true
+    false
 }
```
moulins: This can be removed and replaced by a bare `#[serde(default")]`.

kjarosh: Done

moulins (approved):

PR #23032, opened 2026-02-15
avm2: In SWFv10 and above, frame-script-queued gotos can run "no-op goto" logic

Lord-McSweeney:
This is the "easy" part of #22687. Without this, the main change made in that PR caused tests to fail.

Here's the previous logic for `goto_frame`:
```
if (not going to the current frame) {
    if (currently executing a framescript) {
        queue the goto
    } else {
        execute the goto right there
    }
} else if (in AS3) {
    run no-op goto logic
}
```

And the new logic:
```
if (currently executing a framescript) {
    if (swfv9 AND is a goto to the current frame) {
        run no-op goto logic
    } else {
        queue the goto
    }
} else {
    (this logic was refactored to "goto_frame_now")
    if (not going to the current frame) {
        execute the goto right there
    } else if (in AS3) {
        run no-op goto logic
    }
}
```

The original logic was flawed because it would immediately run no-op goto logic when a goto to the current frame was called, even when a framescript was currently executing.

However, SWFv9, being weird as usual, still immediately runs no-op goto logic when a goto to the current frame is called, even if a framescript is executing.


tests/tests/swfs/avm2/movieclip_queued_noop_goto_swf10/test.toml
```diff
@@ -0,0 +1 @@
+num_frames = 2
```
kjarosh, line 1: Can you make a `movieclip_queued_noop_goto_swf9` version of this test? This way we have a direct comparison between SWF9 and SWF10.

PR #23055, opened 2026-02-17
core: GC-manage SwfMovie, implement native unloadAndStop, and fix VRAM leaks

Xeleron:
This PR refactors `SwfMovie` to be garbage-collected via `gc_arena` (idea by jarca0123), implements the native `Loader.unloadAndStop` for AVM2 (and equivalent cleanup for AVM1), and addresses the resulting GPU VRAM leaks.
(Tested on AQWorlds)

### DIFF
--- a/tests/tests/swfs/avm2/statictext_text/output.txt
+++ b/tests/tests/swfs/avm2/statictext_text/output.txt
@@ -1,8 +1,8 @@
 AB
 CGDA
 null
-null
+B
 null
 AXTD R
-null
-null
+A
+AA


kjarosh:
That's definitely too much for one PR, reviewing all this code will be really difficult. It looks like the second and later commits can be split to separate PRs?

IIRC There were some issues with the approach from the first commit, but @adrian17 and @Lord-McSweeney probably can elaborate on that.

Xeleron:
Closing in favor of #23071

PR #23059, opened 2026-02-17
render: Improve gradient rendering

### LINKED ISSUE #21593: Inaccuracies when a gradient has out-of-order ratios
### Describe the bug

(This describes the root cause of #20895)

Normally a gradient's records are ordered by the ratios, from 0-255. But when the ratios are out of order (when one record's ratio is followed by one that has a value less than it), there are inaccuracies between Flash Player and Ruffle.

## Example instances
For all of these, the records are listed as how they are actually ordered; Flash Player is shown on the left side of each image, and Ruffle on the right side.

- Red, ratio 0
- Green, ratio 128
- Blue, ratio 255
- Cyan, ratio 128
Image

- Red, ratio 0
- Green, ratio 128
- Cyan, ratio 50
- Blue, ratio 255
Image

- Red, ratio 0
- Green, ratio 128
- Cyan, ratio 50
- Magenta, ratio 100
- Blue, ratio 200
- Yellow, ratio 150
Image

### Expected behavior

The gradients should look as how they are seen above in Flash Player.

### Content Location

[gradient.zip](https://github.com/user-attachments/files/22198724/gradient.zip)

This SWF showcases the first case listed in the example instances.
For editing the gradient, I found it helpful to use FFDec's ability to export and re-import XML data.

### Affected platform

Desktop app

### Operating system

macOS 15.6.1

### Browser

_No response_

### Additional information

There should be no reason for a gradient of this nature to appear in any real-world SWF, and yet somehow a case of this appears in Club Penguin (as seen in the linked issue at the beginning). In that case, there was opacity involved with the gradient colors, which is another thing to consider as that was not included with this test case.

ChrisCPI:
Fixes #21593 .

Rather than trying to grab the record with the lowest possible ratio and drawing it, wait until the current record has been drawn up to its ratio before moving on to the next record.

Also fixes some edge cases relating to when records have the same ratio.

The tests may need some work, because:
- The Flash Player snapshots don't look correct (edges of shapes look pretty crispy), but I don't have a better way of snapshotting it.
- I'm not sure what the correct balance is for setting the tolerance/max_outliers, and I'm also not sure if having them as high as they are is acceptable (going back to the first point).


tests/tests/swfs/from_gnash/misc-ming.all/GradientFillTest/output.expected.png
kjarosh, line 1: This test will need fixing before merging.

ChrisCPI: Can this be done by just giving it Flash Player's output and adding tolerance/max_outliers?

ChrisCPI: Expected image has been updated with FP's output

ChrisCPI:
Thinking about it, do I also need to update `visual/gradient_issue_9892`'s expected output to be from Flash Player? (I'll only do this myself if the change with `from_gnash/misc-ming.all/GradientFillTest` looks good)

kjarosh:
Usually you'd do it in a separate PR and then rebase this one on top of that one. Otherwise you're breaking the test first and then fixing it.

kjarosh:
@ChrisCPI can you rebase it and update image assertions? You should be able to lower the tolerance/max_outliers in the existing test(s).

ChrisCPI:
@kjarosh Done!

tests/tests/swfs/visual/gradient_nonsequential_ratios/test.toml
```diff
@@ -0,0 +1,8 @@
+num_frames = 1
+
+[image_comparisons.output]
+tolerance = 32
+max_outliers = 1032
+
+[player_options]
+with_renderer = { optional = true, quality = "high" }
```
kjarosh: Should renderer be optional here?

ChrisCPI: Probably not, I'll change that

tests/tests/swfs/visual/gradient_nonsequential_ratios/output.expected.png
kjarosh, line 1: I'm getting different colors in FP here, looks like it was converted from a different color profile?

tests/tests/swfs/visual/gradient_radial_same_ratios/output.expected.png
kjarosh, line 1: Same here, I'm getting different colors.

tests/tests/swfs/visual/gradient_same_ratios/output.expected.png
kjarosh, line 1: Same here, I'm getting different colors.

tests/tests/swfs/visual/gradient_nonsequential_ratios/test.toml
kjarosh: Optional renderer means that the test will be run even if we cannot produce images and compare them. In this case it would be useful only if we're checking whether Ruffle crashes or not, because there's no textual output.

tests/tests/swfs/visual/gradient_nonsequential_ratios/output.expected.png
ChrisCPI: Oops, I don't think I gave those images the same fix I gave the other tests that were fixed. Will resolve in a bit.

tests/tests/swfs/visual/gradient_nonsequential_ratios/test.toml
ChrisCPI: So should I keep the optional renderer for any of them? It doesn't sound like the new tests should be optional given that they're only comparing visual output.

kjarosh: Yep, they probably shouldn't be optional! By default they are not optional, so you can just remove the `optional` option.

kjarosh (approved):
LGTM, thank you!

> The Flash Player snapshots don't look correct (edges of shapes look pretty crispy), but I don't have a better way of snapshotting it.

That's probably because FP uses its own CPU rendering and looks like they somehow didn't implement anti-aliasing for gradients. That's definitely out of scope of this PR.

> I'm not sure what the correct balance is for setting the tolerance/max_outliers, and I'm also not sure if having them as high as they are is acceptable (going back to the first point).

Your balance is great. Generally we try to set a point that constraints the image the best (i.e. has a slope coefficient close to -1 on the tolerance/outliers graph). A tip for the future: you can set multiple tolerance/outliers checks and check that e.g. for tolerance=10 there's 200 outliers, but for tolerance=50 there's 5 outliers, sometimes it's useful when a test has multiple characteristics.

PR #23060, opened 2026-02-17
tests: Add generated images to gitignore

kjarosh:
The actual and difference images are not meant to be committed.


SuchAFuriousDeath (approved):

SuchAFuriousDeath:
I was always wondering why they were not ignored :D

PR #23062, opened 2026-02-18
tests: Use oxipng to optimize generated images

kjarosh:
Before this patch, generated images were pretty big, sometimes several times larger than necessary.  This patch adds PNG optimization step that will make sure the output file size is small.

### DIFF
--- a/tests/framework/src/runner/image_test.rs
+++ b/tests/framework/src/runner/image_test.rs
@@ -283,6 +283,20 @@ where
 {
     let mut buffer = vec![];
     image.write_to(&mut Cursor::new(&mut buffer), ImageFormat::Png)?;
-    write_bytes(path, &buffer)?;
+    write_bytes(path, &optimize_png(&buffer)?)?;
     Ok(())
+}
+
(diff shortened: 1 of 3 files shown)


tests/framework/src/runner/image_test.rs
```diff
@@ -283,6 +283,20 @@ where
...
     let mut buffer = vec![];
     image.write_to(&mut Cursor::new(&mut buffer), ImageFormat::Png)?;
-    write_bytes(path, &buffer)?;
+    write_bytes(path, &optimize_png(&buffer)?)?;
     Ok(())
 }
+
+fn optimize_png(image: &[u8]) -> anyhow::Result<Vec<u8>> {
+    // Use oxipng to optimize the PNG.
+    let mut options = oxipng::Options::from_preset(6);
```
danielhjacobs: If we always want max compression should we use `oxipng::Options::max_compression();`? It's completely identical but may be clearer: https://docs.rs/oxipng/latest/src/oxipng/options.rs.html#184-186

kjarosh: Done, thanks, I missed that function

PR #23071, opened 2026-02-18
core: Fix memory leaks by making MovieLibrary garbage-collectable

Xeleron:
- **GC-Managed Libraries**: `MovieLibrary` is now a garbage-collected object (`Gc<RefLock<MovieLibrary>>`) that owns the underlying `SwfMovie`. The global registry holds only weak references to these libraries, ensuring they are collected when no longer in use.
- **Direct References**: Every `DisplayObject` holds a strong reference to its source library. When content is unloaded and the display tree is destroyed, the lack of strong references allows the library and swf data to be freed.
- **Registry Cleanup**: The global `Avm2ClassRegistry` has been removed. Symbol class links are now stored directly on `ClassDefinition` as `(MovieLibraryGc, CharacterId)` tuples, eliminating persistent global loop references.
- **Releasable GPU Resources**: Updated `BitmapCharacter` and `Font` to use mutable storage (Cell<Option>), allowing their GPU handles to be actively released during unload.
- **Loader Lifecycle**: `LoaderInfoObject::unload()` now explicitly releases GPU handles, clears cached bitmaps, unregisters the library, and dispatches unload events. Additionally, `Loader.close()` cancels in-progress loads, while `Loader.unloadAndStop()` recursively stops media playback and clears listeners.
- **Weak Event Listeners**: Event handlers now support weak references (`HandlerRef`). This prevents listeners from inadvertently keeping display objects alive, further reducing memory leaks.

### DIFF
--- a/core/src/avm1/activation.rs
+++ b/core/src/avm1/activation.rs
@@ -1145,7 +1145,7 @@ impl<'a, 'gc> Activation<'a, 'gc> {
                         }
                     } else {
                         let level = self.get_or_create_level(level_id);
-                        let future = self.context.load_manager.load_movie_into_clip(
+                        let (future, _handle) = self.context.load_manager.load_movie_into_clip(
                             self.context.player_handle(),
                             level,
                             Request::get(url.to_string()),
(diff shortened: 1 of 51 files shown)


SuchAFuriousDeath:
You could have just redone the PR and force pushed instead of opening a new one.

ChrisCPI:
Is it possible to split the addition of Loader's `unloadAndStop` and `close` methods to another PR? Those would be nice to have on their own. Even if not having its full GC functionality, `unloadAndStop`'s functionality of stopping all sounds etc. would be nice.

PR #19897, opened 2025-03-23
6075 duration issue fix

cosmincalinov:
Replaced f64 time durations in Ruffle with std::time::Duration.

PR #7616, opened 2022-08-09
Use std::time::Duration for time intervals

higumachan:
#6075

I replaced the interval declared in f64 that I could find with ruffle_core::RuffleDuration.

Please let me know if you find any omissions in the f64-declared intervals, as they are difficult to find.

PR #23077, opened 2026-02-19
core: Implement FloatDuration type

### LINKED ISSUE #6075: Use `std::time::Duration` for time intervals
Time durations in Ruffle are often stored as floats ([example](https://github.com/ruffle-rs/ruffle/blob/master/core/src/player.rs#L570-L582)), and it's difficult to know which units the values are in (secs, ms, ...).

It'd be nice to switch these over to use [`std::time::Duration`](https://doc.rust-lang.org/stable/std/time/struct.Duration.html) for clarity.

SuchAFuriousDeath:
Closes #6075
Supersedes #7616 and #19897

### DIFF
--- a/core/common/src/duration.rs
+++ b/core/common/src/duration.rs
@@ -1,0 +1,151 @@
+//! A custom Duration type for representing time intervals with f64 precision.
+//!
+//! This type is used instead of `std::time::Duration` because some calculations
+//! require higher precision than what `std::time::Duration` (nanosecond resolution)
+//! provides. For example, audio calculations may need ~16 significant figures
+//! of precision, which f64 provides.
+
+use std::ops::{Add, AddAssign, Sub, SubAssign};
+
+/// A duration of time represented in milliseconds with f64 precision.
+///
+/// This type provides type safety for time intervals while maintaining
+/// the precision needed for calculations that require f64 precision.
+#[derive(Debug, Clone, Copy, Default, PartialEq, PartialOrd)]
+pub struct FloatDuration(f64);
+
+impl FloatDuration {
+    /// A duration of zero.
+    pub const ZERO: Self = Self(0.0);
+
+    /// Creates a new `FloatDuration` from milliseconds.
+    #[inline]
+    #[must_use]
+    pub const fn from_millis(millis: f64) -> Self {
+        Self(millis)
+    }
+
+    /// Creates a new `FloatDuration` from seconds.
+    #[inline]
+    #[must_use]
+    pub fn from_secs(secs: f64) -> Self {
+        Self(secs * 1000.0)
+    }
+
+    /// Creates a `FloatDuration` from a `std::time::Duration`.
+    #[inline]
+    #[must_use]
+    pub fn from_std(duration: std::time::Duration) -> Self {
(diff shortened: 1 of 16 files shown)


kjarosh:
I think I'm missing the point why we need a floating-point duration? It's worse than fixed point, no?

SuchAFuriousDeath:
https://github.com/ruffle-rs/ruffle/issues/6075#issuecomment-1196072473

SuchAFuriousDeath:
Actually looking at #7616, it converts many more places to the newtype :thinking: I guess I should do it here too, if we want to go this route?

core/common/src/duration.rs
```diff
@@ -0,0 +1,226 @@
+//! A custom Duration type for representing time intervals with f64 precision.
```
Lord-McSweeney, line 1: This type should be in `ruffle_common`, and `ruffle_core` should re-export it

kjarosh:
Hmmm, originally I assumed we can stumble upon numerical instability somewhere when calculating frame times, but looking at the code, I don't see it being an issue.

kjarosh:
> Actually looking at #7616, it converts many more places to the newtype 🤔 I guess I should do it here too, if we want to go this route?

It's fine to do it in several phases, so I wouldn't worry about including it in this PR.

core/common/src/duration.rs
```diff
@@ -0,0 +1,226 @@
...
+use std::ops::{Add, AddAssign, Div, Mul, Sub, SubAssign};
+
+/// A duration of time represented in milliseconds with f64 precision.
+///
+/// This type provides type safety for time intervals while maintaining
+/// the precision needed for audio calculations.
+#[derive(Debug, Clone, Copy, Default, PartialEq, PartialOrd)]
+pub struct FloatDuration(f64);
+
+impl FloatDuration {
```
kjarosh, line 17: Maybe it's worth adding `#[must_use]` to methods?

Lord-McSweeney:
I think we can skip using it in most of the places it was added to in #7616, but can you add `FloatDuration` use to `get_sound_duration` in the audio backend code?

SuchAFuriousDeath:
> I think we can skip using it in most of the places it was added to in #7616, but can you add `FloatDuration` use to `get_sound_duration` in the audio backend code?

Done

core/common/src/duration.rs
```diff
@@ -0,0 +1,150 @@
...
+//! This type is used instead of `std::time::Duration` because audio calculations
+//! require ~16 significant figures of precision, which f64 provides but
+//! `std::time::Duration` (nanosecond resolution) does not.
+
+use std::ops::{Add, AddAssign, Sub, SubAssign};
+
+/// A duration of time represented in milliseconds with f64 precision.
+///
+/// This type provides type safety for time intervals while maintaining
+/// the precision needed for audio calculations.
```
kjarosh: Nit: I wouldn't say that it's made for audio specifically, but that some places need f64 precision, and that's what it can be used for; one example is audio.

SuchAFuriousDeath: I'll fix it up

PR #23113, opened 2026-02-24
avm2: Various fixes and improvements to loading constant pool entries

Lord-McSweeney:
This also removes a lot of our remaining uses of `RustError`


tests/tests/swfs/avm2/cpool_index_invalid_bytecode_3/output.txt
```diff
@@ -0,0 +1 @@
+VerifyError - 1060
```
kjarosh: You could also include the error message here, or even stack trace if useful. Same with other errors

SuchAFuriousDeath:
> This also removes a lot of our remaining uses of `RustError`

Is the goal to eventually get rid of it completely?

tests/tests/swfs/avm2/cpool_index_invalid_bytecode_3/output.txt
Lord-McSweeney: The message doesn't currently match FP, and the stack trace for VerifyErrors is already tested elsewhere, so I would prefer to leave this one as it is

kjarosh: Can't we match the message? Sounds like a straightforward thing to do

Lord-McSweeney: The messages should be matching now

PR #23124, opened 2026-02-26
core: Cleanup some `SwfSlice/buffer::Slice` code

moulins:
- Use `SwfSlice::to_subslice` instead of `to_unbounded_subslice` outside of AVM1 bytecode (which is the only place that needs to allow whole-movie access not restricted to the current tag);
- Remove unneeded arguments in `decode_tags`'s callback & `SwfSlice::resize_to_reader`;
- Implement `io::Read` on `buffer::Slice` directly (instead of `buffer::SliceCursor`);
- Remove unused code after the previous changes.

### DIFF
--- a/core/common/src/buffer.rs
+++ b/core/common/src/buffer.rs
@@ -163,30 +163,6 @@ impl Slice {
         }
     }
 
-    /// Create a subslice of this buffer slice, without bounds checking.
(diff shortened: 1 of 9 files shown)


Dinnerbone (approved):
Lgtm!

PR #23130, opened 2026-02-26
tests: Test more classes in `avm1/native_objects_swf*`.

moulins:
Of note:
- Some objects have a read-only `__proto__`, requring a call to `ASSetPropFlags` before their nativeness can be tested.
- Even with the above work-around, `NetConnection` objects refuse to have their `__proto__` modified.

Possibly related to this [comment in #22920](https://github.com/ruffle-rs/ruffle/issues/22920#issuecomment-3842812783).


tests/tests/swfs/avm1/native_objects_swf8/output.txt
```diff
@@ -15,31 +20,43 @@ new Boolean(true): native
...
+new NetConnection(): ERROR: failed to set __proto__
 new NetStream(): non-native
 new NetStream(new NetConnection()): non-native
 new PrintJob(): native
-SharedObject: non-object: function
+Selection: non-native (readonly proto)
+new SharedObject() /* not a class */: non-native
 new Sound(): native
 new Sound(2): native
 new Sound(_root): native
```
kjarosh, line 49: Based on the discussion on Discord, we could add `new Sound(unloadedMovie)` here?

PR #23151, opened 2026-03-01
tests: Use Flash output for a couple of tests

ChrisCPI:
Related to #23059


tests/tests/swfs/from_gnash/misc-ming.all/GradientFillTest/output.expected.png
kjarosh, line 1: This output looks good, but you can compress the PNG more (e.g. exporting it in GIMP results in 63K instead of 95K).

tests/tests/swfs/visual/gradient_issue_9892/output.expected.png
kjarosh, line 1: Here I'm getting slightly different colors in FP, like #9b6fad vs #9c70ad.

Which FP did you take these outputs from?

kjarosh:
Also, you probably shouldn't embed color profiles in PNGs, it will only confuse users because the PNGs will show different colors than expected.

tests/tests/swfs/visual/gradient_issue_9892/output.expected.png
ChrisCPI: I took the screenshots from Windows 10, then sent it over Discord and cropped it on my Mac, so it wouldn’t surprise me if that messed it up some. I can try a different approach later today.

kjarosh: I'm not an expert, but it looks like the PNG has been converted from one color profile to another. Also be careful when sending pictures through such services as they will often reencode the image to save space. To make sure they are exactly the same, either use checksums or send them in an archive.

PR #23160, opened 2026-03-01
tests: avm1: Add tests for jumping outside of DoAction tag

Randomno:
These tests all jump outside of a DoAction tag and execute a push+trace statement.

jump_to_unknown_tag: Jumps to a custom tag (253)
jump_to_header: Jumps to the FileLength field of the header
jump_to_image_tag: Jumps to near the start of a DefineBitsJPEG2 tag

Various content uses jumps to unknown tags because it was a common obfuscation technique, but I'm not aware of any content doing the latter two operations. jump_to_header currently fails in Ruffle.

header and image_tag were compiled with CS6. unknown_tag was made with FFDec and a hex editor.


kjarosh:
How did you add a jump outside the DoAction tag in CS6?

Randomno:
`__bytecode__("990200D4FF");`

PR #23183, opened 2026-03-04
core: Non-root movies that AVM1 loads should always be considered AVM1

### LINKED ISSUE #21007: Issue with mixed avm

Lord-McSweeney:
Closes #21007


kjarosh:
What if AVM2 loads AVM1, which loads AVM2?

Lord-McSweeney:
Good point, I'll add test coverage for that too

core/common/src/tag_utils.rs
```diff
@@ -263,10 +277,18 @@ impl SwfMovie {
...
 
-    pub fn is_action_script_3(&self) -> bool {
+    /// Whether the SWF's FileAttributes tag declares the SWF to be AS3 (rather
+    /// than AS2).
+    pub fn is_declared_action_script_3(&self) -> bool {
         self.header.is_action_script_3()
     }
 
+    /// Whether the SWF's FileAttributes tag declares the SWF to be AS3 (rather
+    /// than AS2), and `set_force_avm1` hasn't been called on this `SwfMovie`.
```
kjarosh: Nit: instead of summarizing what's the implementation of the method (which everyone can see), you can add more context, e.g.:

> Whether the movie should be interpreted as AS3.
>
> This usually is the same as `is_declared_action_script_3`, but can differ in case AVM1 loads AVM2.

core/src/display_object/movie_clip.rs
```diff
@@ -4116,6 +4112,16 @@ impl<'gc, 'a> MovieClipShared<'gc> {
         tag_code: TagCode,
         reader: &mut SwfStream<'a>,
     ) -> Result<(), Error> {
+        if !self.movie().is_action_script_3() {
+            match tag_code {
+                TagCode::DoAbc => tracing::warn!("DoABC tag in AVM1 movie"),
+                TagCode::DoAbc2 => tracing::warn!("DoABC2 tag in AVM1 movie"),
```
kjarosh, line 4118: Should we warn when `!is_action_script_3` or `!is_declared_action_script_3`?

kjarosh (approved):

kjarosh:
Just to make sure: when AVM1 loads AVM2, the AVM2 cannot load anything as tags are not executed, right?

core/src/display_object/movie_clip.rs
Lord-McSweeney: I would go with `!is_action_script_3`, as if AVM1 loads an AVM2 movie the tags will possibly unexpectedly not be executed

kjarosh: Sure, but then the warnings are a bit misleading? They suggest it's an AVM1 movie despite the fact it's declared as AVM2.

PR #23198, opened 2026-03-07
chore: Resolve beta `clippy::collapsible_match` and `clippy::unnecessary_sort_by` lints

SuchAFuriousDeath:


core/src/avm2/optimizer/type_aware.rs
```diff
@@ -2279,40 +2279,35 @@ fn optimize_call_property<'gc>(
...
-            Some(Property::Slot { slot_id }) | Some(Property::ConstSlot { slot_id }) => {
-                // Don't optimize this for `callpropvoid`
-                if push_return_value {
-                    if stack_value.not_null() {
-                        if num_args == 1 {
-                            let mut value_class =
-                                vtable.slot_class(slot_id).expect("Slot should exist");
-                            let resolved_value_class = value_class.get_class(activation)?;
+            // Don't optimize this for `callpropvoid`
+            Some(Property::Slot { slot_id }) | Some(Property::ConstSlot { slot_id })
```
Lord-McSweeney: Can we keep this one separate? It's a lot of conditions

SuchAFuriousDeath: You mean keep it and add an #[allow(..)] for the lint?

core/src/avm2/optimizer/type_aware.rs
torokati44, line 2287: Would also resolving things make it less readable?

SuchAFuriousDeath: This was the subject of the previous review, I personally would argue that both options have similar readability.

torokati44: Ah okay sorry, I haven't seen it. Alright then.

PR #23202, opened 2026-03-08
avm1: Filter non-finite width in getTextExtent

Bubby4j:
(includes llm generated code)

Fixes an issues in `TextFormat.getTextExtent`:

**Undefined width treated as NaN**: When ActionScript passes `undefined` as the width parameter, `coerce_to_f64` produces `NaN`. This was wrapped as `Some(NaN)` instead of `None`, creating a temporary `EditText` with NaN width and `word_wrap=true`, returning `textFieldWidth=0`.

**Fix**
* Use `args.try_get_f64(activation, 1, UndefinedAs::None)` to handle `undefined width` at the parameter level before coercion, matching the pattern used elsewhere in the codebase.
* Removed `#[expect(dead_code)]` from `try_get_f64` since it's now used.
 
**Testing**
Added `text_format_get_text_extent_undefined_width` test using device font.

**NOTE**: At first we had some more advanced tests which compared exact font metrics using an embedded font in the SWF instead of looking for non-0 values, but it seems like flash player does not use embedded fonts for this use case. Because of that exact parity for all the font metrics in a test isn't currently possible.


core/src/avm1/globals/text_format.rs
```diff
@@ -522,7 +522,8 @@ fn get_text_extent<'gc>(
         .get(1)
         .cloned()
         .map(|v| v.coerce_to_f64(activation))
-        .transpose()?;
+        .transpose()?
+        .filter(|w| w.is_finite());
```
kjarosh: Instead of handling the value here, you should filter out undefined earlier. We have a dedicated method for that: `try_get_f64` with `UndefinedAs::None`.

To be extra sure that it's the correct behavior, you can pass a value that has side effects while coercing to f64 and test whether you can observe those effects:

```
var noisyFloat = new Object();
noisyFloat.valueOf = function() {
  trace("coerced");
  return 42.0;
};
```

tests/tests/swfs/avm1/text_format_get_text_extent/test.as
```diff
@@ -0,0 +1,33 @@
+// Test that getTextExtent handles undefined/NaN width correctly.
```
kjarosh: If you're up to it, you can embed a font in this test and assert on values instead of comparing them to 0. I'm also not sure if this test is valid, because we have no font in this test, whereas FP always does have one. It depends on how getTextExtent works in FP without a font / glyphs.

Bubby4j: output.txt is from flash player, and the test does break when this change is removed, but I can look into doing that for a more realistic test.

kjarosh: Yes, I assume that's the case, but my point is that the environment is different. Flash Player does have a font, we don't have any fonts in tests (because they are not configured). So if Flash Player does return 0 when there's no font, the test would assert the wrong behavior.

kjarosh: And I think the easiest thing you can do is just embed a font and use an embedded font (not device font). As a bonus you can add tests for an embedded font that doesn't exist. (In case you find a bug it's fine to make the test a known failure for a different PR to fix.)

core/src/avm1/globals/text_format.rs
Bubby4j: Added

tests/tests/swfs/avm1/text_format_get_text_extent/test.as
Bubby4j: I did run into another issue, I fixed one of them (thanks for the suggestion!), but some of the other numbers are not the same as flash, so I marked it as a known failure.

core/src/avm1/globals/text_format.rs
```diff
@@ -534,6 +531,7 @@ fn get_text_extent<'gc>(
     );
 
     temp_edittext.set_autosize(AutoSizeMode::Left, activation.context);
+    temp_edittext.set_is_device_font(activation.context, false);
```
kjarosh: So I played with this a little in Flash Player and I don't think that's right; (1) metrics are non-zero when the embedded font doesn't exist, and (2) metrics don't change when there's an embedded font or not.

In which case I think that's the reason you're getting different values than FP.

To be honest... I don't know where `getTextExtent` gets the font from, as metrics don't change even if `TestFont` is provided as a device font, but I think using device fonts is a better option.

Sorry about the confusion, I assumed `getTextExtent` will use embedded fonts but that doesn't seem to be the case :/

In which case I think you can revert to the previous test, set `with_default_font = true`, and rename it to `text_format_get_text_extent_undefined_width`.

Bubby4j: No worries - it was still a good learning experience. Done!

kjarosh (approved):
LGTM, thank you!

PR #23204, opened 2026-03-08
core: Fix behavior when AVM2 focus change events are fired for AVM1 interactives

### LINKED ISSUE #18095: NewCP - Can't login, the text intputs don't work correctly
### Describe the bug

On NewCP I can get to the login screen but I can't intput any text so I can't login

### Expected behavior

It should let me type text in the input fields

### Content Location

https://newcp.net/plays?force=true#/login

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

Mozilla Firefox 130.0.1

### Additional information

image

### LINKED ISSUE #21481: Bug with text inputs
### Describe the bug

(This is a less complicated version of #20989)

When an event listener for `FocusEvent.MOUSE_FOCUS_CHANGE` is added to the `stage`, and an AS2 SWF with a text input is loaded, the event will fire when the text input is clicked.

The handler for the event in this test case runs `preventDefault()` on the event, preventing the text input from being focused on at all.

### Expected behavior

This event should not fire when the text input is clicked.

### Content Location

[text_input_bug.zip](https://github.com/user-attachments/files/21997258/text_input_bug.zip)

### Affected platform

Desktop app

### Operating system

macOS 15.6.1

### Browser

_No response_

### Additional information

This is a new issue because I managed to make a much simpler test case without incorporating UIComponent into it.

This is the cause of #18095

ChrisCPI:
Fixes #18095 (the sites linked in this issue are no longer accessible, but I can attest to the fact that this would've fixed it)
Fixes #21481

When the `target` needs to resolve to the stage, prevent the events from firing when the related object is not AVM2 or `None`.


core/src/focus_tracker.rs
```diff
@@ -253,6 +253,18 @@ impl<'gc> FocusTracker<'gc> {
...
         };
 
+        // An AVM1 interactive object should only fire the event
+        // if the current focus is a real AVM2 interactive object.
+        //
+        // TODO: Clicking a non-interactive AVM1 object should
+        // fire the focus event with the related object
+        // as the root LoaderDisplay.
+        if related_object.is_some_and(|ro| !ro.as_displayobject().movie().is_action_script_3())
+            && target.display_object().as_stage().is_some()
```
kjarosh: Why are we checking if target is the stage here?

ChrisCPI: Because if I’m understanding the code correctly, the current focus (target) resolves to the stage if the given target is not a real display object.

kjarosh (approved):

PR #23208, opened 2026-03-08
avm2: Mark TextBlock.lastLine as a stub

evilpie:


SuchAFuriousDeath (approved):

kjarosh (approved):

core/src/avm2/globals/flash/text/engine/TextBlock.as
```diff
@@ -177,6 +178,7 @@ package flash.text.engine {
         }
 
         public function get lastLine():TextLine {
+            stub_getter("flash.text.engine.TextBlock", "lastLine");
```
Lord-McSweeney, line 181: Currently, the `TextLine` implementation puts all the content into one line, so the first line is the last line- is returning `_firstLine` actually incorrect with the current implementation?

evilpie: Not really, I guess. Still a stub.

Issue #15159, opened 2024-02-07
Flex SDK Flash app doesn't work with ruffle


mtv129:
### Describe what you were doing

I finally got the Flex SDK yesterday (long story) and decided to test it on ruffle, I create a .mxml file with this code:
```xml
<?xml version="1.0" encoding="utf-8"?>
<s:Application xmlns:fx="http://ns.adobe.com/mxml/2009"
               xmlns:s="library://ns.adobe.com/flex/spark"
               minWidth="300" minHeight="200">

    <fx:Script>
        <![CDATA[
            import spark.components.Alert;
          
            private function clickMe():void {
                Alert.show("Hello World!");
            }
        ]]>
    </fx:Script>
    
    <s:Button label="click me" click="clickMe()"/>
</s:Application>
```
but then I discover that ruffle cannot work with MXML at all

https://github.com/ruffle-rs/ruffle/assets/128605704/870676ef-5f45-4653-9c03-b5e57ea76145

Flex SDK is very necessary for AS3, I know about the existence of FlexJS but we are talking about SWF, the task of ruffle is to be able to save old SWF files, this is not done to offend someone that it is not done but done to go to Flex SDK 

### What does the crash message say?

Components and swf files made with action script 3 Flex SDK do not work

### Ruffle Version

(Version: nightly 2024-02-07 | WebAssembly extensions: ON | Used renderer: wgpu-webgl)

### Affected platform

Online demo

### Operating system

Windows 10

### Additional information

I honestly don’t know, ask Apache

Lord-McSweeney:
Can you please ZIP and upload the SWF that you created?

mtv129:
> Can you please ZIP and upload the SWF that you created?

yes, I wrote the code in VS Code

[Flex_test.zip](https://github.com/ruffle-rs/ruffle/files/14200203/Flex_test.zip)

Lord-McSweeney:
The incorrect positioning of the button text is because of TLF issues. I'm not sure why the alert doesn't show- no errors are logged.

PR #23215, opened 2026-03-09
Implement TextBlock::recreateTextLine and basic multi line support

evilpie:
Improves #10173

This implements pretty basic multi-line support for FTE's TextBlock. Most of the code was written by Claude using the Weed to Wonder as a test case. 

It works by first applying the current format and then measuring the number of characters that fit into the first line. The created TextLine is then limited to the number of chracters that fit. Based on the `textLineCreationResult` being "success" (line created) or "completed" (all remaining text fit into the current line, no more lines needed) the caller (flashx) decides whether to create new lines or not.

recreateTextLine allows re-layouting a specific line after changing its width. Which means it might fit more or less text than before.

image


core/src/avm2/globals/flash/text/engine/text_block.rs
```diff
@@ -17,6 +17,94 @@ use crate::display_object::{EditText, TDisplayObject};
...
+                .coerce_to_i32(activation)? as usize;
+            let prev_len = prev
+                .get_slot(line_slots::_RAW_TEXT_LENGTH)
+                .coerce_to_i32(activation)? as usize;
+            prev_begin + prev_len
+        }
+        None => 0,
+    };
+
+    if start >= full_text.len() {
```
Lord-McSweeney, line 61: Can `start > full_text.len()` ever be true?

evilpie: So you mean we would usually assume that we either previously consumed less than or at most equal to the length of the full_text? That makes sense. But currently you can still pass random TextLines into that function, which means that property doesn't hold. I imagine when we add more checks we could maybe soften this to `==`.

Lord-McSweeney:
#15159 seems to hang with this PR

danielhjacobs:
Temporarily switched https://new.weedtowonder.org/ to be built from this branch, can confirm significant improvements.
[weedtowonder_current.zip](https://github.com/user-attachments/files/25876022/weedtowonder_current.zip)
[weedtowonder_PR.zip](https://github.com/user-attachments/files/25876021/weedtowonder_PR.zip)

evilpie:
> #15159 seems to hang with this PR

Due to the multiple lines we are now executing other code and  hanging in this code: https://github.com/apache/flex-sdk/blob/7c47bb9b10c7ca3643079be21597cde022f1bfe1/frameworks/projects/spark/src/spark/components/Label.as#L1544-L1565. Basically `getAtomIndexAtCharIndex` and `getAtomTextBlockEndIndex` always returning zero means it just got stuck inside the loop.

danielhjacobs:
Screenshot From 2026-03-11 09-41-18
For the failing test, left is Linux with this PR, middle is Adobe Flash Player, and right is the expected image in the repo right now. I think it's fair to say it's not really a regression. It seems like there's a preference for using Flash output or if the Flash output is too different marking the test a known failure and still including the Flash output in the repo, so it may be best to change the test accordingly.

danielhjacobs:
For completeness, here are the pages above (screenshots with current Ruffle and this PR) taken from Flash's output.
[weedtowonder_Flash.zip](https://github.com/user-attachments/files/25918594/weedtowonder_Flash.zip)

kjarosh:
> For the failing test, left is Linux with this PR, middle is Adobe Flash Player, and right is the expected image in the repo right now. I think it's fair to say it's not really a regression. It seems like there's a preference for using Flash output or if the Flash output is too different marking the test a known failure and still including the Flash output in the repo, so it may be best to change the test accordingly.

Maybe I'm missing something but it looks like a significant regression? The text looks cut off at the descender, which also means the line would disappear when we implement FP's text culling logic.

evilpie:
Adding the GUTTER to the calculated height seems to mostly fix this issue. This is also improves Weed to Wonder, the "g" in "All organisms exploit their environment" is not cut of anymore.

image

evilpie:
The `visual/definefont4` test is still failing because the second line is now shifted up a little bit. Visually I can't really tell which is more correct vs. real Flash.

tests/tests/swfs/visual/definefont4/output.expected.png
kjarosh, line 1: Can we fix this test in a separate PR merged before this one?

tests/tests/swfs/visual/definefont4/test.toml
```diff
@@ -2,6 +2,7 @@ num_frames = 5
 
 [image_comparisons.output]
 tolerance = 1
+max_outliers = 17500
 
 [player_options]
 with_renderer = { optional = false, quality = "low" }
```
kjarosh, line 8: You can actually set the quality to high, which will result in better assertions against FP's output.

evilpie:
Let's focus on #24313 instead. We can always cherry pick something from here in the future.

Issue #22644, opened 2025-12-30
Error on https://www.y8.com/games/earn_to_die_v1


shempc05-prog:
# Error Info
Error name: Error
Error message: panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-27.0.1/src/backend/wgpu_core.rs:1970:22:
wgpu error: Validation Error

Caused by:
  In Queue::write_texture
    Copy of 452..452 would end up overrunning the bounds of the Source buffer of size 0

Error stack:
```
Error: panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-27.0.1/src/backend/wgpu_core.rs:1970:22:
wgpu error: Validation Error

Caused by:
  In Queue::write_texture
    Copy of 452..452 would end up overrunning the bounds of the Source buffer of size 0


    at Oe.e.wbg.__wbg_new_476169e6d59f23ae (https://unpkg.com/@ruffle-rs/ruffle/core.ruffle.3dcf2eee166c86467f54.js:1:58075)
    at ruffle_web.wasm.__wbg_new_476169e6d59f23ae externref shim (wasm://wasm/ruffle_web.wasm-031f2642:wasm-function[12102]:0x94d858)
    at ruffle_web.wasm.ruffle_web::global_init::{{closure}}::h37d72e0f502d0396 (wasm://wasm/ruffle_web.wasm-031f2642:wasm-function[10126]:0x925b03)
    at ruffle_web.wasm.std::panicking::panic_with_hook::hf90632d5c7102557 (wasm://wasm/ruffle_web.wasm-031f2642:wasm-function[6771]:0x878286)
    at ruffle_web.wasm.core::panicking::panic_fmt::hcb6b2b4be1f4be38 (wasm://wasm/ruffle_web.wasm-031f2642:wasm-function[11126]:0x93feb3)
    at ruffle_web.wasm.wgpu::backend::wgpu_core::ContextWgpuCore::handle_error_inner::h6541f42b9e1baed0 (wasm://wasm/ruffle_web.wasm-031f2642:wasm-function[2428]:0x55c818)
    at ruffle_web.wasm.<wgpu::backend::wgpu_core::CoreQueue as wgpu::dispatch::QueueInterface>::write_texture::h42959e13f3c1b7ae (wasm://wasm/ruffle_web.wasm-031f2642:wasm-function[6433]:0x837062)
[...]

PR #23223, opened 2026-03-10
check bitmap data empty

Smicry:
Fixes a wgpu validation error when updating textures from disposed BitmapData with pending dirty state.

related to #22644.


n0samu:
@Smicry Were you able to reproduce the error from #22644? When does that error occur in the game?

Smicry:
> @Smicry Were you able to reproduce the error from #22644? When does that error occur in the game?

Unable to reproduce, but based on the panic information, founded it need to be checked.

kjarosh:
Can we add a test for that?

Smicry:
> Can we add a test for that?

Looked into it. The guard triggers when update_texture gets a Bitmap with non-zero width/height but empty data — a state that can't be produced through the public API, since Bitmap::new always pads data to width * height * bpp.

A naive test using Bitmap::new(0, 0, Rgba, &[]) compiles and passes, but also passes when the guard is removed, so it doesn't actually protect anything. Reproducing the real state requires a Bitmap::new_unchecked constructor that skips the length check. I prototyped that — the test works — but exposing a public "build a deliberately malformed Bitmap" API just to cover one guard feels like a worse tradeoff than the bug.

The true fix belongs upstream in core::bitmap::bitmap_data, where a proper regression test can drive the state machine directly.

kjarosh:
I guess in this case I don't understand the patch. If this branch cannot be hit while executing a SWF, why are we adding it?

n0samu:
It's supposed to fix these errors: https://github.com/ruffle-rs/ruffle/issues?q=is%3Aissue%20state%3Aopen%20Copy%20of%20%22would%20end%20up%20overrunning%20the%20bounds%20of%20the%20Source%20buffer%20of%20size%200%22
That's pretty important, since these are popular games. It's best if we can figure out how to reproduce the bug and verify that this patch fixes it, though.

kjarosh:
If this PR really fixes those games, it should be possible to create a test, as the behavior is observable.

I asked for a test because it's not obvious why we should have such check, and IMO the problem is somewhere else. Hard to tell without a test.

Smicry:
> If this PR really fixes those games, it should be possible to create a test, as the behavior is observable.
> 
> I asked for a test because it's not obvious why we should have such check, and IMO the problem is somewhere else. Hard to tell without a test.

hi, I think you're right, I've added a test for it!

kjarosh:
Oh sorry, I wasn't specific enough. I meant a SWF test: https://github.com/ruffle-rs/ruffle/blob/master/tests/README.md

Smicry:
> Oh sorry, I wasn't specific enough. I meant a SWF test: https://github.com/ruffle-rs/ruffle/blob/master/tests/README.md

Hi, I tried to write a SWF test for this guard but found it's not really feasible. To trigger bitmap.data().is_empty(), you need a 0×0 BitmapData that's been marked dirty. The only way to get there from AS3 is:
1. Embed a 0×0 bitmap via DefineBitsLossless (since new BitmapData(0,0) gets rejected)
2. Link it to a BitmapData subclass via SymbolClass to access it from AS3
3. Call setVector(new Rectangle(0,0,0,0), new Vector.<uint>()) to mark it dirty
4. Wait for the next frame to render
The blocker is step 2 — you need a properly compiled BitmapData subclass with NewClass initialization, which really requires mxmlc. Hand-writing the ABC bytecode for this is extremely fragile and I couldn't get it to work reliably.
So I kept the Rust unit test since it directly calls backend.update_texture() with empty data — simple and reliable, no need to go through the whole SWF→AS3→core→renderer chain.
I also added a fix in set_cpu_dirty to reject zero-area regions at the source, so the problem shouldn't reach the renderer in the first place. The update_texture guard stays as a safety net.

kjarosh:
What you described sounds like a reasonable edge case that we need to cover by SWF tests, especially if there are SWFs depending on it.

PR #23230, opened 2026-03-10
core: Add object1_or_null and use it when firing onSetFocus/onKillFocus

ChrisCPI:
Fixes an issue where, when an AVM1 movie listening for `Selection.onSetFocus` (or `onSetFocus`/`onKillFocus` on a display object) is loaded by AVM2, then focusing on an AVM2 interactive object would result in `undefined` instead of `null` being given in the event.


ChrisCPI:
Updated to account for individual `onSetFocus`/`onKillFocus` listeners, since they also had this bug.

core/src/focus_tracker.rs
```diff
@@ -192,11 +192,11 @@ impl<'gc> FocusTracker<'gc> {
                     istr!(context, "onSetFocus"),
                     &[
                         old.map(|o| o.as_displayobject())
-                            .map(|v| v.object1_or_undef())
```
kjarosh, line 195: Can we add a `.object1_or_null()` method instead?

PR #23235, opened 2026-03-11
chore: Bump some NPM package versions in package-lock.json

torokati44:
Just to reduce the diff toward a full `npm update` somewhat...


torokati44:
Barring the commit messages / organization, what do you think, @danielhjacobs ?

PR #23237, opened 2026-03-12
avm2: Add SecureSocket support

Xeleron:
This PR implements the `flash.net.SecureSocket` class for AVM2, enabling support for encrypted socket communication within Ruffle. The contents of this pull request were substantially written using Claude Opus 4.6. I've reviewed to the best of my ability.

#### Changes

**AVM2 (`core`)**
*   Added `flash.net.SecureSocket` ActionScript class definition (`SecureSocket.as`).
*   Implemented native Rust bindings and state management for secure sockets.
*   Updated `SocketObject` and `Sockets` manager to support TLS handshake and certificate tracking.

**Backends (`core`, `frontend-utils`)**
*   Extended `NavigatorBackend` with `connect_secure` functionality.
*   Integrated `rustls` into `frontend-utils` to provide cross-platform TLS support (web and desktop).

**Testing (`tests`)**
*   Updated the `mocket` test utility to support TLS server simulation.
*   Added a new regression test `avm2/secure_socket_connect` to verify the end-to-end secure connection flow, including handshake and data exchange.


core/src/avm2/globals/flash/net/secure_socket.rs
```diff
@@ -0,0 +1,60 @@
...
+    let this = this.as_object().unwrap();
+
+    let socket = match this.as_socket() {
+        Some(socket) => socket,
+        None => return Ok(Value::Undefined),
+    };
+
+    let host = args.get_string(activation, 0);
+    let port = args.get_i32(1);
+    if !(1..=65535).contains(&port) {
```
SuchAFuriousDeath: This and the cast would be nicer using try_from.

SuchAFuriousDeath (changes requested):

core/src/avm2/globals/flash/net/secure_socket.rs
Xeleron: Thanks for the feedback. I will look into it ASAP. I had sadly some issues with the Cargo.toml/Cargo.lock which is a bit unusual for me. Reason for this PR was that I didn't want to depend on an ancient library for TLS connections so that I could connect to secure websockets too.

Lord-McSweeney:
Have you found any content that depends on this? IMO we probably shouldn't be adding such a large amount of code to support something if nothing uses it.

Xeleron:
> Have you found any content that depends on this? IMO we probably shouldn't be adding such a large amount of code to support something if nothing uses it.

I was planning to use it perhaps for AQWorlds instead of depending on https://github.com/timkurvers/as3-crypto (AS3Crypto from Hurlant, their TLSEngine/TLSSocket classes in particular) in addition with WebSockets.
Other than that, I am not aware of any other SWF content depending on SecureSocket.
It worked fine for me locally with a custom client build and with this Ruffle branch.

MrGrymReaper:
If http and https isn’t handled separately (part of net socket and/or secure socket) then it will be needed for things like VMware vCenter 6.7 or lower, due to it connecting to a backend service. Unless something better can be devised. As well as depending on how coded the VMware ESXi hosts.

There’s other enterprise class appliance user interfaces which depend on Adobe Flash (now Ruffle) to work for management and migration activities.

For additional context on how this impacts enterprise migration pipelines (such as moving off legacy VMware environments), I've been discussing this exact network socket dependency over on the XCP-ng / Vates VMS forums. Dealing with legacy appliances that rely on backend socket communication is a major hurdle for teams dealing with unmaintained vendor software.

https://xcp-ng.org/forum/topic/12379/migrations-of-software-to-vates-vms

PR #23244, opened 2026-03-14
core: Fix focusing behavior in SWF 8- vs 9+ and in mixed AVM

### LINKED ISSUE #23182: Inaccurate behavior with AVM1 text input focusing
### Describe the bug

In Ruffle, with an AVM1 movie, when you click anywhere while a text input is currently focused, then it will unfocus. In Flash Player, this not (always) the case.

### Expected behavior

In a standalone AVM1 movie, this behavior is SWF version dependent. On version 9 and above, clicking *anywhere* does not unfocus the text input. When the version is below 9, then clicking anywhere outside the text input should unfocus it.

But then it gets weird with mixed AVM: When the AVM1 movie is loaded into AVM2, clicking on anywhere *except for an AVM1 button* will unfocus the text input. The version of the AVM1 movie *does not matter* in this case.

Also, some notes about this behavior in mixed AVM:
When an AVM2 text input is focused on, clicking an AVM1 button will still prevent the focus from being reset. Also in this case, clicking the AVM1 button will still fire the `mouseFocusChange` event with `null` as the `relatedObject`. However, since the focus hasn't actually reset, clicking the AVM1 button repeatedly will fire the event every time for as long as the text input is in focus.

### Content Location

A simple test, where a button is covering the entire stage, and a text input is on top:

[avm1_textinput_focus.swf.zip](https://github.com/user-attachments/files/25729901/avm1_textinput_focus.swf.zip)

### Affected platform

Desktop app

### Operating system

macOS Tahoe 26

### Browser

_No response_

### Additional information

_No response_

ChrisCPI:
Fixes #23182

As explained in the relevant comment in the code, when the root SWF version is 9 or higher, then clicking anything not focusable by mouse (such as an AVM1 button) does not reset the focus. Also, clicking the stage will only reset the focus in AVM2.

Also changes the focus events for AVM2 to not fire until all of AVM1's handlers (individual `onSetFocus`/`onKillFocus` handlers, and `Selection.onSetFocus` handlers) have been fired, as this seems to be the correct behavior (see the `avm2/focus_mixed_avm` test).


core/src/focus_tracker.rs
```diff
@@ -117,6 +118,10 @@ impl<'gc> FocusTracker<'gc> {
             return;
         }
 
+        if should_not_unfocus {
```
kjarosh: That doesn't look right:
1. the variable has a double negation which is just confusing,
2. it doesn't do what it says: if `new` is `Some` and focusable, it will not be focused...

core/src/player.rs
```diff
@@ -1909,13 +1909,22 @@ impl Player {
...
+            // We need to clear the focus if an object focusable by mouse was un-focused.
+            // However, there is some specific behavior here.
+            // When the SWF version is >= 9, clicking anything that is
+            // not focusable by mouse does not clear the focus, and
+            // only in AVM2 does clicking on the stage clear the focus.
+            let should_not_unfocus = context.root_swf.version() >= 9
+                && match pressed_object {
+                    None => !context.root_swf.is_action_script_3(),
+                    Some(int) => !int.is_focusable_by_mouse(context),
+                };
```
kjarosh: It looks like we can just handle this logic in `set_by_mouse`? If `new` is `None` at the end and [other conditions] we can just skip calling `set_internal`.

core/src/focus_tracker.rs
```diff
@@ -164,6 +169,20 @@ impl<'gc> FocusTracker<'gc> {
 
         let old = self.0.focus.get();
 
+        fn handle_focus_change<'gc>(
```
kjarosh, line 172: Can we split this change into a separate PR?

ChrisCPI: We can, but it would probably mean leaving out the `avm2/focus_mixed_avm` test out of this current PR (or modifying it to not use the AVM1 focus handlers), since I unintentionally came across the behavior while I was crafting that test and decided to fix it so that it would pass.

kjarosh: That makes sense, can you at least split it across two separate commits? That's gonna get handy when bisecting and looking for regressions.

core/src/player.rs
ChrisCPI: I've tried, but that doesn't appear to be correct. It's hard to articulate why, but everything I've tried results in usually the `avm2/focus_mixed_avm` test failing.

core/src/focus_tracker.rs
ChrisCPI: Would this be better?:
```rs
            let should_reset_focus = context.root_swf.version() < 9
                || match pressed_object {
                    None => context.root_swf.is_action_script_3(),
                    Some(int) => int.is_focusable_by_mouse(context),
                };
```
And then in `focus_tracker.rs`:
```rs
        if !should_reset_focus {
            return;
        }
```

core/src/player.rs
ChrisCPI: Example behavior of when I try to do it that way: Clicking an AVM2 text input and then an AVM1 button unfocuses the text input when it shouldn't.

I think it basically comes down to this:
- This should only be checked specifically when the focus needs reset to `None`.
- When the code falls into that block, its possible for `pressed_object` to still be `Some`, even though `None` is always passed to `set_by_mouse`.

core/src/focus_tracker.rs
ChrisCPI: Actually, with further testing, I found that simplifying the condition to this also seems to work:
```rs
            let should_reset_focus = context.root_swf.version() < 9
                || (pressed_object.is_none() && context.root_swf.is_action_script_3());
```
Which I think makes sense, because even if `pressed_object` was `Some`, then it would still need to pass the `is_focusable_by_mouse` check, which it would have already failed in order to get to this point in the code in the first place, so it could never return `true`(?)

ChrisCPI:
Closing this as I've decided to split it into 2 separate PRs.

The first one will have *only* the changes relating to when the focus events are fired, and will be tested by modifying the existing `avm2/focus_events_mixed_avm_edittext` test.

The second one, which will rely on the first one, will have the rest of the changes and all the new tests seen here.

PR #23248, opened 2026-03-15
avm2: Fix return font subclass in Font.enumerateFonts

### LINKED ISSUE #22979: [Font] flash.text.Font::enumerateFonts(); returns "flash.text::Font" instead of the actual font asset
### Describe the bug

This is a little bit complex but I'll try to explain it:

Basically I have `game/client/Game.26346.swf` and `game/client/FontMapper.swf`.
Ruffle loads `Game.26346.swf` and then `FontMapper.swf`: so far so good until a piece of code in the `Game.26346.swf` runs this code here `flash.text.Font::enumerateFonts()` at `Init.FontMapperInit::execute()` which returns a different value from what **Flash** returns:

Ruffle:
```as
trace("Fonts:", Font.enumerateFonts());
// "Fonts: [object Font],[object Font],[object Font],[object Font]"
```

Flash:
```as
trace("Fonts:", Font.enumerateFonts());
// "Fonts: [object DefaultFont],[object DefaultBoldFont],[object TitleFont],[object DefaultSerifFont]"
```

And the issue is that later we run `Font.registerFont(class);` at `cafe2.rad.fonts.FontLibrary::addMapping()` which gives this error (because it doesn't accept **flash.text::Font**):
Image

### Expected behavior

I expect the font to be registered without giving any errors.

### Content Location

On my github: https://github.com/Antonio225t/cityville

### Affected platform

Self-hosted version

### Operating system

Arch Linux

### Browser

LibreWolf

### Additional information

_No response_

besuper:
This fix makes `Font.enumerateFonts()` return the correct font subclass instead of always using the base `flash.text.Font` class.

The `ClassObject` in `Font.registerFont()` was not preserved, so `enumerateFonts()` created all instances with the base Font class. This broke SWFs that rely on `getQualifiedClassName()` to recover the original font class from enumerated fonts.

Fixes https://github.com/ruffle-rs/ruffle/issues/22979
I've tested CityVille with this fix, and the game now loads correctly


kjarosh:
Does enumerate fonts return the same or new instances each time it's called? What's the relation between those objects and the ones injected through symbol classes?

The behavior above should also be covered with a test(s).

besuper:
> Does enumerate fonts return the same or new instances each time it's called? What's the relation between those objects and the ones injected through symbol classes?
> 
> The behavior above should also be covered with a test(s).

Enumerate fonts always returns new instances. While testing I figured out that embedded fonts appear as base Font instances. Only fonts explicitly registered with registerFont return instances of the subclass.

I updated the fix to make embedded fonts always use the base Font class and only return the subclass for globally registered fonts. I also added a test covering both behaviors

core/src/avm2/globals/flash/text/font.rs
```diff
@@ -118,24 +118,33 @@ pub fn enumerate_fonts<'gc>(
             // TODO: EmbeddedCFF isn't supposed to show until it's been used (some kind of internal initialization method?)
             // Device is only supposed to show when arg0 is true - but that's supposed to be "all known" device fonts, not just loaded ones
             if font.has_layout() && font.font_type() == FontType::Embedded {
-                fonts.push(font);
+                fonts.push((font, default_font_class));
             }
         }
     }
 
+    for font in activation.context.library.global_fonts() {
```
kjarosh, line 126: This looks O(n^2), can we make it O(n)?

besuper: I ended up storing an Option<Class Object> directly in FontMap struct instead of duplicating fonts in a separate vec. I think it's cleaner than maintaining a second vec alongside it

core/src/library.rs
```diff
@@ -744,18 +746,25 @@ impl<'gc> Library<'gc> {
 
 #[derive(Collect, Default)]
 #[collect(no_drop)]
-struct FontMap<'gc>(FnvHashMap<FontQuery, Font<'gc>>);
+struct FontMap<'gc>(FnvHashMap<FontQuery, (Font<'gc>, Option<ClassObject<'gc>>)>);
```
Lord-McSweeney: Could we store the `Option<ClassObject>` directly on the `Font`?

besuper: Yes this is now in FontData

core/src/avm2/globals/flash/text/font.rs
```diff
@@ -107,7 +107,7 @@ pub fn enumerate_fonts<'gc>(
         );
     }
 
-    fonts.append(&mut activation.context.library.global_fonts());
+    let default_font_class = activation.avm2().classes().font;
```
Lord-McSweeney, line 110: This change means that global fonts will now be in the list after the movie's library's fonts- is that intended? Does Flash return them in a specific order?

besuper: Not really intended. But after testing with global fonts first, my test was in error. Looks like if there are 2 fonts with the same font name, the sort doesn't take count of global/embedded fonts, so it's random order. I will update the test so it doesn't take count of order but only test the font class

PR #23253, opened 2026-03-15
Fix different priority of getter if by name or property on different namespaces

### LINKED ISSUE #17960: SDT Does not play
### Describe the bug

The version used doesn't seem to matter much (tried with 1.21.1b and also with 5.22 loader version, with loader 5.45 the game fails to load entirely displaying only the gradient background for the main menu)

The game won't play, pressing the buttons on the main menu (play and options) will not do anything, and if you bypass the main menu it will just cycle through all the sprites rapidly.

### Expected behavior

Play game

### Content Location

This (NSFW) game, full name super deepthroat is available [on flashpoint](https://flashpointproject.github.io/flashpoint-database/search/#d8862e81-4b86-46a3-b8dc-1a690611d48d). 

### Affected platform

Desktop app

### Operating system

Linux

### Browser

_No response_

### Additional information

Version: Ruffle 0.1.0-nightly (4fd9dc45c5c54548941e6263d5c5f271f388ed33 2024-09-16)

wsxarcher:
When a subclass declares a field with the same local name as a private (or internal) field in a superclass, AVM2 property lookups via bracket access (this["field"]) were resolving to the wrong slot.

The priority is reversed from where they are present in the set with priority to publics.

Traced with Flashplayer 32


Lord-McSweeney:
Unfortunately I think the proper solution to this is a little more complicated. We need to change our vtable implementation so that each vtable first attempts to lookup a property on only its own map, then falls back to its parent vtable. See this SWF: [chainedvt.swf.zip](https://github.com/user-attachments/files/26010697/chainedvt.swf.zip)

It outputs
```
subclass-field1
subclass-field1
superclass-field2
superclass-field2
```
with this PR, but
```
subclass-field1
subclass-field1
subclass-field2
subclass-field2
```
with FP.

Since `Subclass`'s vtable is the top-most, in FP the property lookups all find properties on it, which means none of the property lookups get a chance to look at `Superclass`'s vtable.

wsxarcher:
Fixing the insertion order rather than the lookup seems to fix also your case without touching the vtable, I added your test.

PR #23274, opened 2026-03-18
chore: Bump wgpu to 28, switch egui to git

torokati44:
See:
https://github.com/emilk/egui/pull/7694
https://github.com/gfx-rs/wgpu/pull/8230

### DIFF
--- a/render/wgpu/src/context3d/mod.rs
+++ b/render/wgpu/src/context3d/mod.rs
@@ -122,7 +122,7 @@ impl WgpuContext3D {
         let front_buffer_raw_texture_handle = make_dummy_handle();
 
         // FIXME - determine the best chunk size for this
-        let buffer_staging_belt = StagingBelt::new(1024);
+        let buffer_staging_belt = StagingBelt::new(descriptors.device.clone(), 1024);
         let current_pipeline = CurrentPipeline::new(&descriptors);
 
         let buffer_command_encoder =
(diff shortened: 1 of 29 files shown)


torokati44:
It's unfortunate that `sctk-adwaita` still pulls in `ab_glyph`, so we'll have two independent font rasterization engines... :/

torokati44:
*bump*

Any thoughts on the double font rendering stack?

torokati44:
`ruffle_desktop` binary size in `release` config, on my Linux machine:
`master`: 40277840 B **(38.4 MiB)**
this PR: 42196088 B **(40.24 MiB)**
increment: 1918248 B **(1.83 MiB), 4.76% of current size**

torokati44:
I think we can work on trying to drop `ab_glyph` later, and live with this additional weight for a while...

render/wgpu/src/context3d/mod.rs
```diff
@@ -948,9 +946,10 @@ impl Context3D for WgpuContext3D {
                     &mut self.buffer_command_encoder,
                     buffer,
                     offset,
-                    NonZeroU64::new(std::mem::size_of_val(matrix_raw_data_column_major) as u64)
-                        .unwrap(),
-                    &self.descriptors.device,
+                    NonZeroU64::new(
```
Lord-McSweeney: What was the reason for changing away from `size_of_val`?

Lord-McSweeney (approved):

render/wgpu/src/context3d/mod.rs
torokati44: Good catch - this is just an artifact of rebasing past  https://github.com/ruffle-rs/ruffle/pull/23901, accidentally reverting a small bit of it: https://github.com/ruffle-rs/ruffle/commit/86f5df3b3701bee050032e05a00b33b7761a4af7#diff-0146f05505ec38abc6fa9c40c2b4790d379219d224c3e79734e8379e1353a2e0

PR #23276, opened 2026-03-18
avm2: Speed up Value numerical coercions

Lord-McSweeney:
In large `match`es, the compiler uses a jump table to determine what to jump to. This can result in a lot of overhead for each coercion. Instead, we now have a match for only three conditions instead, `Number`, `Integer`, and everything else. This makes the compiler generate a chain of branches instead.

From what I've seen, this makes `coerce_to_u32`, `coerce_to_i32`, and `coerce_to_number` about 30% faster, which translates to an overall improvement of ~1-2% on physics-heavy SWFs such as box2d.

We can further improve performance if we inline the functions, but this results in code size significantly increasing.

### DIFF
--- a/core/src/avm2/value.rs
+++ b/core/src/avm2/value.rs
@@ -681,25 +681,44 @@ impl<'gc> Value<'gc> {
     ///
     /// Numerical conversions occur according to ECMA-262 3rd Edition's
     /// ToNumber algorithm which appears to match AVM2.
+    ///
+    /// This function can be very hot in physics code, so we inline a fast-path
+    /// for `Value::Number` and `Value::Integer`, and fall back to a non-inlined
+    /// slow-path handling the rest of the cases if necessary.
     pub fn coerce_to_number(
         &self,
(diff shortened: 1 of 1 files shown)


core/src/avm2/value.rs
```diff
@@ -681,25 +681,45 @@ impl<'gc> Value<'gc> {
...
+        fn coerce_to_number_slow<'gc>(
+            value: &Value<'gc>,
+            activation: &mut Activation<'_, 'gc>,
+        ) -> Result<f64, Error<'gc>> {
+            Ok(match value {
+                Value::Undefined => f64::NAN,
+                Value::Null => 0.0,
+                Value::Bool(true) => 1.0,
+                Value::Bool(false) => 0.0,
+                Value::Number(_) => unreachable!("Handled by fast path"),
```
kjarosh: You can merge these to one case using `|`

kjarosh: Similarly cases below

PR #23288, opened 2026-03-19
avm1: Fix regression with action_get_url_2 and MovieClip without DisplayObject

### LINKED ISSUE #22674: Slasher Click 2: The tale of Magawa: Status ailment crash
### Describe the bug

Using any attack that causes status ailment like sandstorm kills the game to a black screen when the animation is done. Easiest way to test is to go up and farm the enemy on the left till you have 75 gold, then buy the sandstorm attack, equip it, then attack an enemy with it.

From testing, the regression seemed to start on build 2025-11-14, but reverting to 2025-11-13 makes it work.

### Expected behavior

Attacks that cause status effects shouldn't kill the game.

### Content Location

https://www.newgrounds.com/portal/view/975359?emulate=flash

### Affected platform

Browser's extension

### Operating system

Android 16

### Browser

Chrome 143

### Additional information

_No response_

evilpie:
Fixes #22674

This was a regression introduced in #22221.

Before:
https://github.com/ruffle-rs/ruffle/blob/9b3d688aaf0e1e3b5a2aac9c066c402bf25e12d2/core/src/avm1/activation.rs#L1256-L1261

After:
https://github.com/ruffle-rs/ruffle/blob/c6e0cb01586efbce690c97678bd8ec8362b0c3a2/core/src/avm1/activation.rs#L1217-L1219

The code is subtly different. Before, we used `None` if the MovieClip didn't have a DisplayObject, but afterwards, we fell through to the `else` block.

/cc @moulins


SuchAFuriousDeath (approved):

kjarosh:
Having a test would be nice. Otherwise we'll regress it again in the future.

kjarosh:
Also, see https://discord.com/channels/610531541889581066/1018988628380241931/1457721092184735917 for discussion about this.

evilpie:
While the new behavior did cause the game to break, the new behavior might actual match Flash?

```
class Test {
    static function main() {
        var mc = _root.createEmptyMovieClip("mc", 1);
        mc.removeMovieClip();
        // This should NOT unload _root.
        loadMovie("", mc);

        // Check after a delay that _root was not unloaded.
        _root.checkId = setInterval(function() {
            clearInterval(_root.checkId);
            trace("_totalframes: " + _root._totalframes);
        }, 200);
    }
}
```

This logs nothing for me in Flash.

PR #23294, opened 2026-03-20
ci: add PR-only native desktop client CI builds

### LINKED ISSUE #23269: Provide the binary file in the CI of the PR for testing
## Note: that my English might not so good If there are any errors, please forgive me, please. Thanks

Hi team,

I'm verifying if PR #23145 fixed the regression from Issue #23245 affecting multiple Flash games.

I couldn't find compiled test binaries in the workflow artifacts to test this fix. Could you provide them? I also suggest adding a step to the PR CI workflow to auto-publish test binaries for all PRs. This would allow users to validate fixes independently and help collect broader real feedback more efficientl

Thank you!

yangyangdaji:
I've set up CI builds for the native desktop client and generated the corresponding binaries. Web and Windows MSI packages were excluded, as configuration file issues broke their builds and I couldn't fix them. I've disabled the problematic feature, so
 the current script meets our needs. When uploading macOS files, a "file not found" warning appears, despite the file being confirmed in the target directory and the 
example https://github.com/yangyangdaji/ruffle/actions/runs/23324405071

final build containing the correct file.
The build also produces nested compressed archives (compressed files nested inside one another).

Image

Close https://github.com/ruffle-rs/ruffle/issues/23269

### DIFF
--- a/.github/workflows/build_pr.yml
+++ b/.github/workflows/build_pr.yml
@@ -1,0 +1,244 @@
+name: Build pr
+
+on:
+  # Allow for manual dispatch on GitHub
+  workflow_dispatch:
+
+  # Allow for PR auto trigger for build
+ # push:
+  pull_request:
+
+jobs:
+  build:
+    name: Build ${{ matrix.build_name }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - build_name: linux-x86_64
+            os: ubuntu-24.04
+
+          - build_name: linux-aarch64
+            os: ubuntu-24.04-arm
+
+          # Mac does two Rust builds to make a universal binary
+          - build_name: macos-x86_64
+            os: macos-15
+            target: x86_64-apple-darwin
+            MACOSX_DEPLOYMENT_TARGET: 10.7
+            DESKTOP_FEATURES: sandbox,jpegxr
+
+          - build_name: macos-aarch64
+            os: macos-15
+            target: aarch64-apple-darwin
+            MACOSX_DEPLOYMENT_TARGET: 11.0
+            DESKTOP_FEATURES: sandbox,jpegxr
+
+          - build_name: windows-x86_32
+            os: windows-2025
+            target: i686-pc-windows-msvc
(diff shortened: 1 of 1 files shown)


yangyangdaji:
The build also produces nested compressed archives (compressed files nested inside one another).

kjarosh:
Looks like it duplicates our release process? Two separate but identical workflows induce high maintenance burden, and our workflows are too complicated already... I would personally suggest to reuse the release workflow.

yangyangdaji:
> Looks like it duplicates our release process? Two separate but identical workflows induce high maintenance burden, and our workflows are too complicated already... I would personally suggest to reuse the release workflow.

Yes, this file is based on the release process.  I have indeed considered the points you raised, yet my knowledge of CI is only rudimentary. I attempted to make changes to the release_nightly.yml file, but my modifications rendered the file inoperable due to my limited familiarity with YAML syntax. For this reason, I had to rewrite the file entirely from scratch following the provided example, which is the root cause of the imperfections in this feature.

yangyangdaji:
> Looks like it duplicates our release process? Two separate but identical workflows induce high maintenance burden, and our workflows are too complicated already... I would personally suggest to reuse the release workflow.

I found some stable workflows here: https://github.com/ocaml-multicore/multicoretests/tree/main/.github/workflows  ; in these workflows, repetitive logic has been broken down and grouped into dedicated files,The correct solution is to split release_nightly.yml and isolate the compilation steps to reduce maintenance work, since most of the code is shared between releases and PR pipelines.,I tried to split it, but failed due to a lack of relevant documentation.If you have any suggestions, please feel free to offer them; you are also welcome to take over and process this PR directly as needed

kjarosh:
> I found some stable workflows here: https://github.com/ocaml-multicore/multicoretests/tree/main/.github/workflows

I think this is the right approach for us if we decide to go in this direction. We'd need to create a separate release workflow and call it from both nightly release and PR workflows. I'm not sure how it would look in the PR and if nightly release would even work with it. It looks like a major refactor.

PR #23303, opened 2026-03-21
core: Use arrow mouse cursor for LoaderDisplay interactive

ChrisCPI:
Previously, if a mouse pick landed on the Loader (like a loaded movie having only a graphics shape), then in Ruffle it would show the hand cursor, but Flash Player has the arrow cursor.

Unsure if this needs a test, since I'm not sure how that could be tested headless.


kjarosh:
In order to test it, I think we would have to extend the test framework and add cursor type assertions in input.json.

PR #23304, opened 2026-03-21
avm2: Handle negative indices in Array.indexOf()

tobil4sk:
Based on 8675b29ceee9b1287716cdfc228d5eff5a7aee53 (#20982).

Note, I recompiled the test swf files using `mxmlc -o test.swf -debug Test.as` (from [here](https://github.com/ruffle-rs/ruffle/blob/master/CONTRIBUTING.md#apache-flex-sdk)), and the tests pass but they also now show:

```
2026-03-21T15:56:49.859622Z ERROR ruffle_core::avm2: Error attempting to lookup AVM2 symbol class: TypeError: Error #2023: Class Test$ must inherit from Sprite to link to the root.
```
Not sure if this is something I did wrong.


n0samu:
The Test class should extend Sprite or MovieClip, we should update the example in Contributing.md. The error shouldn't cause any problems, but it's confusing.

tobil4sk:
> The Test class should extend Sprite or MovieClip, we should update the example in Contributing.md.

I opened #23306. I wonder if the `fla` files and `as` files are out of sync, because the test.as files I have seen do not inherit from Sprite but the `swf` files don't give an error.

kjarosh (approved):

PR #23305, opened 2026-03-21
avm2: Fix edge cases in String.indexOf/lastIndexOf

tobil4sk:
This fixes the following edge cases:
- Start index is larger than the string length for String.indexOf
- Pattern is `null`/`undefined`. Previously it would search for `"null"` in the string instead of returning `-1`.

These edge cases were already covered by the tests, however, the function output was not being checked properly.


tests/tests/swfs/avm2/string_indexof_lastindexof/test.as
```diff
@@ -1,46 +1,89 @@
...
+trace(s.lastIndexOf("", 5));
+trace("// s.lastIndexOf(\"\", 100)");
+trace(s.lastIndexOf("", 100));
+trace("// s.lastIndexOf()");
+trace(s.lastIndexOf());
+trace("// s.lastIndexOf(null)");
+trace(s.lastIndexOf(null));
+trace("// s.lastIndexOf(undefined)");
+trace(s.lastIndexOf(undefined));
+trace("// \"hello undefined hi\".lastIndexOf(undefined)");
```
kjarosh, line 92: Can you add a similar case for `null` (for `ibdexOf` too).

tests/tests/swfs/avm2/string_indexof_lastindexof/test.as
```diff
@@ -1,46 +1,89 @@
...
+trace("hello undefined hi".lastIndexOf(undefined));
```
kjarosh, line 93: I would personally add also `"".[...]indexOf(null)` and same with undefined.

tests/tests/swfs/avm2/string_indexof_lastindexof/test.as
tobil4sk: The string `s` has `null` in it, and we check `s.indexOf(null)` and `s.indexOf(undefined)`. Is this ok? Those checks are how I discovered the bug.

tests/tests/swfs/avm2/string_indexof_lastindexof/test.as
tobil4sk: Added these in the last push

tests/tests/swfs/avm2/string_indexof_lastindexof/test.as
kjarosh: Right, that makes sense. Didn't notice it!

kjarosh (approved):
Thank you!

PR #23308, opened 2026-03-21
avm1,avm2: Fix unicode string comparisons

tobil4sk:
String comparisons in flash happened using 16 bit code units.

A test for avm1 is probably needed as well, but I haven't been able to download mtasc.


kjarosh (approved):
Thank you!

kjarosh:
Regarding avm1 test, you can download JPEXS and craft the test there. It's fine to submit a follow-up PR with the test.

PR #23313, opened 2026-03-22
web: Open new tabs beside the current one by extension

pzhlkj6612:
Hello! Thanks for maintaining Ruffle. I can't always go back to my childhood to play old games without your hard work.

The logic about opening new tabs in the browser extension can be improved. I have so many tabs open. New ones opened by Ruffle always go to the end of the looooong tab list. I asked AI and have modified the code to let new tabs open beside the current one.

I'm not able to test the "onboard.html" page because I can't find a way to trigger the "INSTALL" event with unpacked extension. In Firefox Nightly, installing XPI file with setting `xpinstall.signatures.required` to `false` does not work. I can't figure it out.

The two changes about "player.html" have been tested, though.

Here is the commit message:

```
* Changes:
    * Open new tabs beside the current tab instead of at default position (after all tabs).
    * Remove redundant "promisify" wrapper from "tabs.create" calls.
* Test Environment:
    * Windows 11, Firefox 143.0.
    * Windows 11, Chrome 146.0.7680.80.

```

Please review, thank you!

### DIFF
--- a/web/packages/extension/src/utils.ts
+++ b/web/packages/extension/src/utils.ts
@@ -37,21 +37,6 @@ export let declarativeNetRequest:
     | typeof browser.declarativeNetRequest
     | typeof chrome.declarativeNetRequest;
 
-function promisify<T>(
-    func: (callback: (result: T) => void) => void,
-): Promise<T> {
-    return new Promise((resolve, reject) => {
-        func((result) => {
-            const error = chrome.runtime.lastError;
-            if (error) {
-                reject(error);
-            } else {
-                resolve(result);
-            }
-        });
-    });
-}
-
 if (typeof browser !== "undefined") {
     i18n = browser.i18n;
     scripting = browser.scripting as ScriptingType;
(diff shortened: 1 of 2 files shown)


evilpie (changes requested):
Instead of setting the index, you should set the `openerTabId` to the right tab. In Firefox at least this will automatically open the new tab after the current tab. This behaviour is actually controllable by the user using the about:config preference `browser.tabs.insertRelatedAfterCurrent`.

pzhlkj6612:
> Instead of setting the index, you should set the `openerTabId` to the right tab. ...

Thank you, I didn't know that parameter. Let me check a better appraoch.

pzhlkj6612:
Hello!

> ... In Firefox at least this will automatically open the new tab after the current tab. This behaviour is actually controllable by the user using the about:config preference `browser.tabs.insertRelatedAfterCurrent`.

Unfortunately, Google Chrome does not have such thing:

> Various browsers have different behavior for placing the new tab, some of them even allow changing it. Having an opener doesn't necessitate proximity in the tab strip. In your extension you need to specify `index` parameter to control the placement i.e. you can get the active tab first and add 1 to its index.
> 
> from: tabs.create always place opened tab at last even if openerTabId is set [40822765] - Chromium | https://issues.chromium.org/issues/40822765#comment2 at Feb 16, 2022 09:06PM

To make behavior consistent across browsers, I think here what we only can do is specifying the index of newly created tabs.

<br>

We can search "openerTabId" to see more discussion about it in Chromium's bug tracker: https://issues.chromium.org/issues?q=openerTabId

web/packages/extension/src/utils.ts
```diff
@@ -73,10 +58,22 @@ if (typeof browser !== "undefined") {
...
+        currentWindow: true,
+    });
+    return activeTab?.index !== undefined ? activeTab.index + 1 : undefined;
+}
+export const openPlayerPage: () => Promise<void> = async () => {
+    const index = await getAdjacentTabIndex();
+    await tabs.create({ url: "/player.html", index });
+};
+export const openOnboardPage: () => Promise<void> = async () => {
+    const index = await getAdjacentTabIndex();
```
evilpie: The onboarding page isn't really related to the current page so I don't think it really make sense to do this.

pzhlkj6612: Thank you. I indeed did some ctrl-c & ctrl-v things. Here is the fix: 9dcc2d615488f00a6238a20cf827a19a629ee873 .

PR #23322, opened 2026-03-24
avm2: Fix ByteArray utf-16 encode/decode

tobil4sk:
Decode was broken because utf-16 code points can contain null bytes, but the raw bytes were being truncated at the first null. Now it truncates at the first actual null character after decoding.

Encode was broken because `encoding_rs` always uses utf-8 instead of utf-16 for encoding: https://docs.rs/encoding_rs/latest/encoding_rs/struct.Encoding.html#method.new_encoder

> Note: The output encoding of UTF-16BE, UTF-16LE, and replacement is UTF-8. There is no encoder for UTF-16BE, UTF-16LE, and replacement themselves.


tests/tests/swfs/avm2/bytearray_string_null/Test.as
```diff
@@ -1,7 +1,8 @@
 ﻿package {
```
Lord-McSweeney, line 1: Could you remove the `test.fla` for this test if it's now outdated?

core/src/avm2/globals/flash/utils/byte_array.rs
```diff
@@ -652,7 +651,24 @@ pub fn write_multi_byte<'gc>(
         let encoder =
             Encoding::for_label(charset_label.to_utf8_lossy().as_bytes()).unwrap_or(UTF_8);
         let utf8 = string.to_utf8_lossy();
-        let (encoded_bytes, _, _) = encoder.encode(&utf8);
+
+        // encoding_rs doesn't support encoding to UTF-16, so we have to handle it manually
+        let encoded_bytes = if encoder == UTF_16LE || encoder == UTF_16BE {
+            let mut bytes = Vec::with_capacity(utf8.len() * 2);
```
CrazyboyQCD, line 657: I think BOM should be added here.
```suggestion
            let mut bytes = Vec::with_capacity(utf8.len() * 2 + 2);
            bytes.extend_from_slice(if encoder == UTF_16LE {
                &[0xFF, 0xFE]
            } else {
                &[0xFE, 0xFF]
            });
```

tobil4sk: I don't think that would match flash player:
```as
import flash.utils.ByteArray;

var bytes = new ByteArray();
bytes.writeMultiByte("A", "utf-16be");
bytes.position = 0;

trace(bytes.readByte());
trace(bytes.readByte());

var bytes = new ByteArray();
bytes.writeMultiByte("A", "utf-16le");
bytes.position = 0;

trace(bytes.readByte());
trace(bytes.readByte());
```

```
0
65
65
0
```

CrazyboyQCD: OK, we should follow FP's behavior.

evilpie:
Thank you, looks good to me.

PR #23326, opened 2026-03-25
core: Don't load from ImportAssets if url is empty

### LINKED ISSUE #22018: Issue when ImportAsset has an empty url
### Describe the bug

This test SWF has 2 frames. Only the second frame has a symbol for an imported asset. The URL in the tag for the imported asset is an empty string (`""`).

On frame 1, `nextFrame();` is called so that the movie immediately goes to frame 2. However, in Ruffle, this doesn't happen, because of the failure to load the imported asset. This is normally correct, however Flash Player doesn't seem to care when the URL for the asset is empty.

### Expected behavior

Should move to frame 2 without problems.

### Content Location

[importasset_empty_bug.swf.zip](https://github.com/user-attachments/files/23111635/importasset_empty_bug.swf.zip)

### Affected platform

Desktop app

### Operating system

macOS 15.6.1

### Browser

_No response_

### Additional information

This has caused a regression in an alternate version of one of Club Penguin's minigames, which has an ImportAssets tag with an empty URL (somehow), where it cannot move to the second frame which contains the actual gameplay.

ChrisCPI:
Fixes #22018

When the url of an `ImportAssets`/`2` tag is empty, Flash Player does not attempt to make a request for the asset.


core/src/display_object/movie_clip.rs
```diff
@@ -743,12 +743,15 @@ impl<'gc> MovieClip<'gc> {
...
         };
 
-        let mc = context.gc();
-        let library = context.library.library_for_movie_mut(self.movie());
-
         let asset_url = url.to_string_lossy(UTF_8);
 
-        let request = Request::get(asset_url);
+        // FP does not attempt to load the url if it is empty
+        if asset_url.is_empty() {
```
kjarosh, line 749: Can you split this into a separate PR with a test?

ChrisCPI:
Actually, now I'm wondering if this is the correct fix, because it actually doesn't exactly fix the test in the linked issue. This fixes other cases that I've tested, but not that one.

For context, that test SWF has two frames on the root, and the second frame has an `ImportAssets2` tag. The first frame has a `DoAction` that calls `nextFrame();`, which works in FP, but not in Ruffle even with this fix. Removing the `nextFrame()` call makes it work.

ChrisCPI:
Fixing that will probably go beyond the scope of this PR, since the same thing also happens with other cases, even when loading a valid SWF file.

PR #23329, opened 2026-03-26
core: Continue executing when loaded url from ImportAssets is not an SWF

ChrisCPI:
Originally split from #23326.

Flash Player (at least in AVM1) continues executing the loader movie when this happens.


tests/tests/swfs/import_assets/avm1_non_swf_import/test.swf
kjarosh, line 1: Can you add frame 3 and stop there? Because often logic in FP is executed within the given frame, so it's possible FP just stops at frame 2. (And this also prevents Ruffle from doing weird things after frame 2.)

ChrisCPI: Done, although it doesn't really give anything interesting; FP just moves to frame 3 after the import as would be expected, and so does Ruffle with these changes.

kjarosh (approved):
Thank you!

PR #23333, opened 2026-03-27
core: In `replace_at_depth`, don't set removed child's parent as None

### LINKED ISSUE #21500: Bug with unloading sound
### Describe the bug

In this case, there is a swf loading another swf. In the child swf, it creates a looping sound and attaches it to an empty MovieClip. That MovieClip then gets destroyed or overwritten by another new empty MovieClip due to having the same depth.

There is a button that tells the root movie to remove the entire child MovieClip. But when it does, **the sound does not stop**, presumably due to the MovieClip that held the sound already being destroyed by this point.

### Expected behavior

The sound should stop when the child movie is destroyed.

### Content Location

[sound_bug.zip](https://github.com/user-attachments/files/22018822/sound_bug.zip)

(run sound_bug_loader.swf)

### Affected platform

Desktop app

### Operating system

macOS 15.6.1

### Browser

_No response_

### Additional information

This is the cause of a bug in Club Penguin where one of the mini-games' music does not stop after leaving it.

ChrisCPI:
Fixes #21500

To the best of my knowledge, it is not possible for an AVM2 child to end up in this block, and doing it for AVM1 doesn't seem correct. It broke the audio manager's `stop_sounds_on_parent_and_children` logic when an externally loaded movie had a `Sound` created with an MC that was then replaced by depth with another.

I have not been able to create a test to prove that this is the correct fix, and so there is a chance that regressions arise from this.


kjarosh:
Would be nice to have a test for it. Have you tried listening for some sound events and tracing?

core/src/display_object/container.rs
```diff
@@ -218,11 +218,10 @@ pub trait TDisplayObjectContainer<'gc>:
         child.set_place_frame(0);
         child.set_depth(depth);
 
-        if let Some(removed_child) = removed_child {
-            if !self.raw_container().is_action_script_3() {
-                removed_child.avm1_unload(context);
-            }
-            removed_child.set_parent(context, None);
```
kjarosh, line 225: Can you add a comment here that `set_parent(None)` should not be called and describe why?

tests/tests/swfs/avm1/issue_21500/output.txt
kjarosh, line 1: Can you add a simple log on frame 3/4 to make sure we're there?

tests/tests/swfs/avm1/issue_21500/test.toml
```diff
@@ -0,0 +1,8 @@
+num_frames = 3
+
+[player_options]
+with_audio = true
+
+[audio_assertions.silence]
```
kjarosh: Can you also add a positive assertion to make sure the sound was here before?

tests/tests/swfs/avm1/depth_replacement_audio_unloading/test.toml
kjarosh, line 1: Give the test more descriptive name instead of `issue_21500`.

tests/tests/swfs/avm1/depth_replacement_audio_unloading/test.toml
```diff
@@ -0,0 +1,12 @@
+num_frames = 4
+
+[player_options]
+with_audio = true
+
+[audio_assertions.sound]
+frames = [3]
+max_amplitude = 0.68
```
kjarosh: This doesn't really assert that the sound is here, you'd want to use `min_max_amplitude`.

ChrisCPI: Semi-related question: Is the approximate calculated amplitude expected to be the same across all platforms? For me its 0.67254645, but i bumped it to 0.68 because I wasn't sure if there would be small differences between the different test runners, like how visual tests are.

kjarosh: That's a good question, I don't think it can vary that much, maybe on macos which uses a different math library. But you should be taking the lower bound and asserting it's higher (e.g. higher than 0.6).

PR #23347, opened 2026-03-30
avm2: More performance improvements

Lord-McSweeney:
A few functions that can be hot are now marked `FastCall`. I also optimized implementations of the `Matrix3D` methods- this reduces the amount of time spent in function call overhead in some 3D SWFs by up to 20%.

### DIFF
--- a/core/src/avm2/globals/flash/geom/Matrix3D.as
+++ b/core/src/avm2/globals/flash/geom/Matrix3D.as
@@ -52,18 +52,18 @@ package flash.geom {
                 ty = pivotPoint.y;
                 tz = pivotPoint.z;
             }
-            var radian = degrees * Math.PI / 180;
-            var cos = Math.cos(radian);
-            var sin = Math.sin(radian);
-            var x = axis.x;
-            var y = axis.y;
-            var z = axis.z;
-            var x2 = x * x;
-            var y2 = y * y;
-            var z2 = z * z;
-            var ls = x2 + y2 + z2;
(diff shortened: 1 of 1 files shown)


core/src/avm2/globals/flash/geom/Matrix3D.as
```diff
@@ -224,7 +224,7 @@ package flash.geom {
         [Ruffle(NativeCallable)]
         public function transpose():void {
             // Make a copy
-            var oRawData = this._rawData.AS3::concat();
+            var oRawData:Vector.<Number> = this._rawData.AS3::concat();
```
evilpie: Seems more in line with your other changes to use `this.rawData` here.

core/src/avm2/class.rs
```diff
@@ -308,6 +308,7 @@ impl<'gc> Class<'gc> {
         let mut i_class = ClassData::empty(name);
         i_class.param = Lock::new(Some(Some(param)));
         i_class.super_class = Some(object_vector_i_class);
+        i_class.attributes = Cell::new(ClassAttributes::FINAL);
```
evilpie: Can this be tested?

Lord-McSweeney: I'll move this change to another PR

core/src/avm2/globals/flash/geom/Matrix3D.as
Lord-McSweeney: Done

PR #23380, opened 2026-04-04
core: Only fire AVM2's focus events after AVM1's listeners

ChrisCPI:
Split from #23244.

AVM2's `focusOut` and `focusIn` events should only fire after AVM1's `Selection.onSetFocus` listeners and individual text `onSetFocus`/`onKillFocus` listeners.


tests/tests/swfs/avm2/focus_events_mixed_avm_edittext/test.swf
kjarosh, line 1: I think you can also add event listeners for the buttons, not only text fields here.

tests/tests/swfs/avm2/focus_events_mixed_avm_edittext/avm1.swf
kjarosh, line 1: Similarly here, we have 2 fields, but only one has event listeners. Are we focusing all 3 of them?

core/src/focus_tracker.rs
```diff
@@ -164,6 +164,20 @@ impl<'gc> FocusTracker<'gc> {
 
         let old = self.0.focus.get();
 
+        fn handle_focus_change<'gc>(
+            int: Option<InteractiveObject<'gc>>,
+            focused: bool,
```
kjarosh, line 169: Can we call it `has_focus` so that it matches the property below?

kjarosh: Actually, nevermind, other methods call it `focused` too.

tests/tests/swfs/avm2/focus_events_mixed_avm_edittext/avm1.swf
ChrisCPI: I didn't do all three because I figured that doing just one was enough to get the point across, but I can update it to be all of them.

kjarosh: I think we can also cover potential interactions between fields (e.g. the respective order of calling callbacks). Adding listeners for all of them and identifying which one the log comes from would be great.

Issue #10013, opened 2023-03-12
Can Your Pet: Soccer ball minigame not working correctly


n0samu:
### Describe the bug

[Link to SWF](https://flasharch.com/cdn/uploads/archived/2020/10/17/e0a3edb5d66f12e00e161c4e254b0c26_20201017160245.swf)

(Warning for those unaware, this is a shock/horror game, so don't play it past the point where you can see the bug, unless you're prepared for that!)

In Can Your Pet, you're supposed to be able to pass a soccer ball back and forth with your pet by clicking on it while it's in the air. But in Ruffle, the ball disappears when you try to pass it back.

https://user-images.githubusercontent.com/71368227/224529775-f7898a3b-067a-47fa-9560-0b2ad87d1a65.mp4

### Expected behavior

In Ruffle builds 2023-02-24 and before, the minigame worked correctly:

https://user-images.githubusercontent.com/71368227/224529842-0131f1dc-f5c7-49e4-8b59-b68c7a424fc0.mp4

### Affected platform

Desktop app

### Operating system

Windows 10

### Additional information

When the soccer ball disappears, this error is shown in the terminal:
```
ERROR run_frame:run_all_phases_avm2: ruffle_core::avm2::activation: TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: y)
        at framework::GameLogic/ballSky()
        at flash.events::EventDispatcher/flash::events::EventDispatcher::dispatchEvent()
        at fl.transitions::Tween/set time()
        at fl.transitions::Tween/nextFrame()
        at fl.transitions::Tween/onEnterFrame()
```
The error happens on the first line of the `ballSky` function:
`var _loc2_:Number = this.ball_mc.y;`
`ball_mc` is a MovieClip. It was originally instantiated like this:
`this.ball_mc = new Ball_MC();`
It doesn't make sense that ball_mc is null here.

Lord-McSweeney:
I've come to the conclusion that this is the affected piece of code:
```
      private function initMiniGame() : void
      {
         Particle.gravity = 0.25;
         this.ball = new Circle(65);
         this.ball.x = this.ball_mc.x;
         this.ball.y = this.ball_mc.y;
         this.ball.display = this.ball_mc;
         this.ball.addEventListener(MouseEvent.MOUSE_DOWN,this.onBallDown);
         this.ball.addEventListener("bound_right",this.onBallRightBound);
[...]

PR #23389, opened 2026-04-04
core: Fix mouse cursor not inheriting from button-mode parent MovieClips

### LINKED ISSUE #21379: Incorrect mouse cursor behavior over interactive objects
### Describe the bug

As you can see in the GIF, the mouse cursor changes back to normal when you hover over the object.

Image

Other example (with more glitching):

Image

### Expected behavior

It should actually display the hand cursor over the entire object.

### Content Location

https://www.panfu.me/play (registration and completion of tutorial required)

### Affected platform

Self-hosted version

### Operating system

Windows 11

### Browser

Google Chrome 139.0.7258.128

### Additional information

Using jsdelivr to load Ruffle, the latest nightly version (0.2.0-nightly.2025.8.17)

robinmiau:
Fixes #21379

Previously, `mouse_cursor` only checked whether the hovered object itself was a button-mode clip, so hovering over a child inside a button-mode Sprite would always show the arrow cursor.

This walks the ancestor chain to find the nearest button-mode MovieClip and uses its `useHandCursor`/`enabled` state to determine the cursor, stopping at the first match so a disabled intermediate ancestor isn't silently skipped.

### DIFF
--- a/core/src/display_object/movie_clip.rs
+++ b/core/src/display_object/movie_clip.rs
@@ -3177,13 +3177,13 @@ impl<'gc> TInteractiveObject<'gc> for MovieClip<'gc> {
         // The nearest button-mode ancestor governs the cursor, so stop at the first one.
         let mut current: Option<DisplayObject<'gc>> = Some(self.into());
         while let Some(obj) = current {
-            if let Some(mc) = obj.as_movie_clip() {
-                if mc.is_button_mode(context) {
-                    if mc.use_hand_cursor(context) && mc.enabled(context) {
-                        return MouseCursor::Hand;
-                    }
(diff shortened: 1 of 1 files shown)


n0samu:
@robinmiau clippy is complaining about a nested if statement in your code, be sure to fix that. Thanks for tracking down and fixing this, it's a significant UX improvement!

Lord-McSweeney:
This also seems to fix the mouse cursor not changing to a hand when mousing over the ball in Can Your Pet (#10013)

Benman2785:
does not fix Obama vs Zombies (Instruction and Achievements button)

PR #23390, opened 2026-04-04
core: Fix NetStream playback stalling when play() is called after EOF

turicfr:
When a NetStream naturally reaches the end of a video, it synchronously dispatches a NetStream.Play.Stop status event and then abruptly pauses itself. If a synchronous ActionScript event handler listens to this event and tries to replay the video using `stream.play()`, the stream would previously become incorrectly paused right after, because the EOF pause evaluation proceeded regardless of whether the stream was modified.

- Extract `reset_buffer()` out from the asynchronous `load_netstream` closure in `loader.rs` and execute it synchronously inside `NetStream::play()`.
- Before executing the EOF auto-pause inside `NetStream::tick()`, we now evaluate `Gc::ptr_eq` against the original `source` pointer to verify that the backing buffer hasn't been reset during the status event dispatch.

An AVM2 test case is also included.


Dinnerbone (approved):
Thank you!

PR #23395, opened 2026-04-05
web: Add background tick mode to keep player running in hidden tabs

### LINKED ISSUE #7289: Background playback
When I switch to a different tab on Newgrounds, Ruffle suspends the flash elements, whereas the original flash kept playing in background. Please bring background playback to Ruffle. Thanks in advance.

### LINKED ISSUE #10527: Ruffle should not getting Paused when user is switching browser tab.
In Original Flash Plugin The player did not getting Paused by switching tab
https://user-images.githubusercontent.com/75534590/229392632-7fa4aec7-bf44-401a-81bb-4e5b61978b4c.mp4

But in Ruffle
https://user-images.githubusercontent.com/75534590/229392867-02b08cf1-fcf5-4ab6-9dec-8e63cdcf1132.mp4

Similar project with Webassembly
https://user-images.githubusercontent.com/75534590/229393270-313279e1-3b4f-48e1-9d9d-4f825546e1e7.mp4

robinmiau:
Related to #7289
Related to #10527

Previously, when a tab was hidden, the `requestAnimationFrame` loop would stop, halting game logic and audio. This adds a background tick mode that replaces the RAF loop with a [Web Worker](https://developer.mozilla.org/docs/Web/API/Worker)-based tick loop when the tab becomes hidden, keeping the player running.

When the tab becomes visible again, the background worker is terminated, the normal RAF loop is restarted, and the `AudioContext` is explicitly resumed since browsers may auto-suspend it in background tabs.

### New configuration option

A new `backgroundExecutionMode` config option controls background playback behavior:

| Value | Behavior |
| -----  | --------- |
| `"mainThread"` | Uses the Web Worker tick loop when the tab is hidden |
| `"none"` | Disables background execution entirely (default) |

**Example usage:**

```js
  "backgroundExecutionMode": "mainThread"
};
```

### Notes

- The `visibilitychange` listener no longer passes an explicit `false` as the third argument, since that is the default and every other listener in the file omits it.
- Tested in the latest version of Chrome, where it ran reliably and smoothly in the background.

### DIFF
--- a/web/packages/core/src/internal/player/inner.tsx
+++ b/web/packages/core/src/internal/player/inner.tsx
@@ -511,8 +511,8 @@ export class InnerPlayer {
     private startBackgroundTick(): void {
         const intervalMs = 1000 / (this.metadata?.frameRate || 24);
 
-        // intervalMs is a number computed from SWF metadata, so there is no
-        // risk of code injection via the template literal below.
+        // intervalMs is always a number (dividing 1000 by anything yields a number
+        // or NaN), so no code injection via the template literal is possible.
         const workerCode = `
             const intervalMs = ${intervalMs};
             self.onmessage = () => {
--- a/web/packages/core/src/internal/player/inner.tsx
+++ b/web/packages/core/src/internal/player/inner.tsx
@@ -473,33 +473,33 @@ export class InnerPlayer {
      */
     private setupTabVisibilityHandling(): void {
         document.addEventListener("visibilitychange", () => {
-                if (!this.instance) {
-                    return;
-                }
-                const mode =
-                    this.loadedConfig?.backgroundExecutionMode ??
-                    BackgroundExecutionMode.None;
-                if (document.hidden) {
-                    this.lastActivePlayingState = this.instance.is_playing();
-                    if (mode === BackgroundExecutionMode.MainThread) {
-                        this.instance.enable_background_tick_mode();
-                        if (this.lastActivePlayingState) {
-                            this.startBackgroundTick();
-                        }
-                    } else {
-                        this.instance.pause();
+            if (!this.instance) {
+                return;
+            }
+            const mode =
+                this.loadedConfig?.backgroundExecutionMode ??
+                BackgroundExecutionMode.None;
+            if (document.hidden) {
+                this.lastActivePlayingState = this.instance.is_playing();
+                if (mode === BackgroundExecutionMode.MainThread) {
+                    this.instance.enable_background_tick_mode();
+                    if (this.lastActivePlayingState) {
+                        this.startBackgroundTick();
                     }
                 } else {
-                    if (mode === BackgroundExecutionMode.MainThread) {
-                        this.stopBackgroundTick();
-                        this.instance.restart_animation_loop();
-                    } else if (this.lastActivePlayingState) {
-                        this.instance.play();
-                    }
-                    // Browsers may auto-suspend AudioContext in background tabs.
-                    this.instance.audio_context()?.resume();
+                    this.instance.pause();
                 }
-            });
+            } else {
+                if (mode === BackgroundExecutionMode.MainThread) {
+                    this.stopBackgroundTick();
+                    this.instance.restart_animation_loop();
+                } else if (this.lastActivePlayingState) {
+                    this.instance.play();
+                }
+                // Browsers may auto-suspend AudioContext in background tabs.
+                this.instance.audio_context()?.resume();
+            }
+        });
     }
 
     /**
--- a/web/packages/core/src/internal/player/inner.tsx
+++ b/web/packages/core/src/internal/player/inner.tsx
@@ -467,9 +467,7 @@ export class InnerPlayer {
      * running in the background when the tab is hidden.
      * While hidden, the normal animation loop is replaced with a background tick
      * so audio and game logic continue running.
-     * When the tab becomes visible again, the animation loop is restarted and
-     * {@link pause} is called only if the instance was already paused before the
-     * tab was hidden.
+     * When the tab becomes visible again, the animation loop is restarted.
      *
      * See: https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API
      * @ignore
--- a/web/packages/core/src/internal/player/inner.tsx
+++ b/web/packages/core/src/internal/player/inner.tsx
@@ -194,6 +194,9 @@ export class InnerPlayer {
     private instance: RuffleHandle | null;
     private newZipWriter: (() => ZipWriter) | null;
     private lastActivePlayingState: boolean;
+    private backgroundTickChannel: MessageChannel | null;
+    private backgroundTickActive: boolean;
+    private backgroundLockRelease: (() => void) | null;
 
     metadata: MovieMetadata | null;
     _readyState: ReadyState;
@@ -338,7 +341,10 @@ export class InnerPlayer {
         this.metadata = null;
 
         this.lastActivePlayingState = false;
-        this.setupPauseOnTabHidden();
+        this.backgroundTickChannel = null;
+        this.backgroundTickActive = false;
+        this.backgroundLockRelease = null;
+        this.setupTabVisibilityHandling();
     }
 
     addFSCommandHandler(handler: (command: string, args: string) => void) {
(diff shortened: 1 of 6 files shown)


kjarosh:
Note that we still support Firefox 94 and Safari 14.1 which do not support locks. We can either drop support for older versions or handle the missing API.

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -459,34 +463,99 @@ export class InnerPlayer {
...
+                        } else {
+                            this.backgroundLockRelease = resolve;
+                        }
+                    }),
+            );
+        }
+
+        const channel = new MessageChannel();
+        this.backgroundTickChannel = channel;
+        const intervalMs = 1000 / (this.metadata?.frameRate ?? 24);
```
kjarosh: Note that FP ticks in the background at a slower rate than requested, I think it's capped to 4-15 FPS depending on platform/browser? The main issue with ticking at the same rate is resource consumption, which I'm not sure is important considering we're running on better hardware and the browser has control over the execution.

kjarosh:
It would be really nice to have that covered by tests, it's a pretty important piece of code.

web/packages/core/src/internal/player/inner.tsx
robinmiau: If we cap the FPS, couldn't that cause some side effects when the SWF expects the full FPS?

In my test using a 24 FPS SWF file, CPU usage in Chrome's Task Manager dropped from over 12% during active use to 2-3% when running in the background. RAM usage remained almost the same.

kjarosh: > If we cap the FPS, couldn't that cause some side effects when the SWF expects the full FPS?

In general, there's no guarantee FP will run the movie at the requested frame rate, it can run slower based on how it performs. That's why movies that assume the requested framerate won't run well in FP too.

> In my test using a 24 FPS SWF file, CPU usage in Chrome's Task Manager dropped from over 12% during active use to 2-3% when running in the background. RAM usage remained almost the same.

That looks good, have you checked whether the browser applies any throttling somehow (i.e. whether the real framerate is the same as requested)?

robinmiau: I added a debug log to measure the actual background tick rate. Turns out the browser limits `setTimeout` to roughly 1 call per second in background tabs regardless of the requested delay.

Since `tick()` receives the actual elapsed time, a single background tick just processes all the missed frames at once as a burst, so audio and game logic stay correct. No explicit cap needed because the browser already handles the throttling.

I just tested with a simple `setTimeout` and it works just as well, so we could probably drop the `MessageChannel` entirely and keep it simple.

kjarosh: Yeah, that makes sense. I think the only way to execute logic reliably in the background is to put it in a service worker.

robinmiau: A `Worker` + `OffscreenCanvas` would probably solve it properly but is likely a significant refactor. For now `setTimeout` should be fine, `MAX_FRAMES_PER_TICK` caps catch-up at 5 frames anyway so bursting isn't a concern.

If that's okay with you, I'll implement it that way.

kjarosh: I think it's fine to rewrite it as `setInterval` in this case. We don't have to worry about compatibility. But considering that the browser throttles the execution significantly I think we have potential for breaking content. I would add a config option that enables background execution (e.g. `backgroundExecutionMode` with values `none`/`mainThread`), leave it disabled by default for now, and let the community test it out.

robinmiau: Sounds good, should I go ahead and add that right away? And I think `setTimeout` is more reliable here than `setInterval`, so nothing stacks up.

kjarosh: What do you mean by "nothing stacks up"?

robinmiau: With `setInterval`, if a tick runs longer than the interval, the next callback is already queued and fires immediately when the current one finishes. With `setTimeout`, the next callback is only scheduled after the current one completes, so there's always exactly one pending callback and stacking is impossible by design.

kjarosh: I see, I guess that makes sense considering we're being throttled to few FPS anyway.

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -459,34 +464,96 @@ export class InnerPlayer {
...
+     * Starts a background tick loop that keeps audio and game logic running while the tab is hidden.
+     *
+     * @ignore
+     * @internal
+     */
+    private startBackgroundTick(): void {
+        this.backgroundTickActive = true;
+
+        if ("locks" in navigator) {
+            navigator.locks.request(
```
kjarosh: Do we have any evidence that suggests that while holding a lock we'll get more resources?

robinmiau: Looking into this more closely, I found that Chrome always throttles background tabs, even with a lock, unless audio is playing, which happened to be the case in my tests and explains why it worked initially.

As an alternative, I tried using a simple Worker that sends a `postMessage('tick')` inside a `setTimeout`. This keeps it running in the background even without active audio playback. Should I go with this approach?

kjarosh: Yeah, I think that makes sense. Not sure if you plan to receive that message and act upon it, but I think it makes sense for the the worker to trigger ticks in the main thread. To prevent the bursting issue you were talking about, you might also want to use two way communication.

robinmiau: I've pushed the changes and updated the initial comment to reflect them.

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -194,6 +195,7 @@ export class InnerPlayer {
     private instance: RuffleHandle | null;
     private newZipWriter: (() => ZipWriter) | null;
     private lastActivePlayingState: boolean;
+    private backgroundWorker: Worker | null;
```
kjarosh, line 202: Can you add a comment above it and describe the overview of how this works:
1. the worker is non bull only when ticking in the background,
2. the communication (tick -> ack -> sleep -> tick),
3. when it's initialized and destroyed.

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -459,34 +462,84 @@ export class InnerPlayer {
...
+    }
+
+    /**
+     * Starts a background tick loop while the tab is hidden.
+     */
+    private startBackgroundTick(): void {
+        const intervalMs = 1000 / (this.metadata?.frameRate || 24);
+
+        const workerCode = `
+            const intervalMs = ${intervalMs};
```
kjarosh, line 517: Can you add a comment above it elaborating on why there can't be any code injection?

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -459,34 +462,84 @@ export class InnerPlayer {
...
+            }
+            const mode =
+                this.loadedConfig?.backgroundExecutionMode ??
+                BackgroundExecutionMode.None;
+            if (document.hidden) {
+                this.lastActivePlayingState = this.instance.is_playing();
+                if (mode === BackgroundExecutionMode.MainThread) {
+                    this.instance.enable_background_tick_mode();
+                    if (this.lastActivePlayingState) {
+                        this.startBackgroundTick();
```
kjarosh, line 489: It's probably worth handling the situation where we fail to create a worker here and fall back to pausing the movie.

web/packages/core/src/public/config/load-options.ts
```diff
@@ -789,6 +805,13 @@ export interface BaseLoadOptions {
      * @default DeviceFontRenderer.Embedded
      */
     deviceFontRenderer?: DeviceFontRenderer;
+
+    /**
+     * Controls how Ruffle behaves when the browser tab is hidden.
+     *
+     * @default BackgroundExecutionMode.None
+     */
+    backgroundExecutionMode?: BackgroundExecutionMode;
```
kjarosh, line 814: Can we add this option to the extension too? This way not only admins, but regular users could also try it out.

Adding it to the extensions is super easy, we just need to add a GUI and name the option properly, everything else should work.

web/packages/extension/assets/options.html
```diff
@@ -48,6 +48,13 @@
                 </select>
                 <label for="log_level">Log level</label>
             </div>
+            <div class="option select">
+                <select id="background_execution_mode">
+                    <option value="none" id="background_execution_mode_none">Pause</option>
+                    <option value="mainThread" id="background_execution_mode_main_thread">Keep running in background</option>
```
kjarosh:
```suggestion
                    <option value="mainThread" id="background_execution_mode_main_thread">Run in the main thread</option>
```

It's better to leave some room for potential "Run in a worker" option, as translations are expensive

web/packages/extension/assets/options.html
```diff
@@ -48,6 +48,13 @@
                 </select>
                 <label for="log_level">Log level</label>
             </div>
+            <div class="option select">
+                <select id="background_execution_mode">
+                    <option value="none" id="background_execution_mode_none">Pause</option>
```
kjarosh, line 53: All these IDs need to be added to `web/packages/extension/assets/_locales/en/messages.json` so that they can be translated.

robinmiau: Added them!

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -459,34 +466,88 @@ export class InnerPlayer {
...
+    }
+
+    /**
+     * Starts a background tick loop while the tab is hidden.
+     */
+    private startBackgroundTick(): void {
+        const intervalMs = 1000 / (this.metadata?.frameRate || 24);
+
+        // intervalMs is a number computed from SWF metadata, so there is no
+        // risk of code injection via the template literal below.
```
kjarosh: I would mention that it's a number because a number divided by anything will produce a number. Even if `frameRate` is not a number, `intervalMs` will produce `NaN` and thus no injection. It doesn't matter that `this.metadata?.frameRate` comes from the SWF metadata in this case.

kjarosh (approved):
LGTM, I'll do some tests and I think we'll be ready to merge this

Pandalusion:
Tested this on [Habbo Flash Client](https://habbolegacy.com/flash_client) , which normally disconnects after around 50 seconds when minimised or in the background. It doesn't disconnect and continues pinging websockify, no disconnections even after several hours

kjarosh:
Thank you!

PR #23396, opened 2026-04-05
core: Fix cursor not updating in all hover cases

robinmiau:
Two cases where the cursor was not updated correctly:

1. When the mouse was stationary but the hovered object changed (e.g. an animated object moved under the cursor), `mouse_cursor()` was never called because hover events were skipped. Fixed by re-evaluating the cursor whenever the hovered object changes, regardless of mouse movement.

2. When moving the mouse within the same interactive object, `mouse_cursor()` was never called since the hovered object pointer did not change. This caused the cursor to remain stuck, most visibly in TextFields where links should show a hand cursor but did not. Fixed by re-evaluating the cursor on every mouse move over the same object.

### DIFF
--- a/core/src/player.rs
+++ b/core/src/player.rs
@@ -1762,6 +1762,7 @@ impl Player {
             // When the mouse moves within the same object, re-evaluate the cursor,
             // since different parts of an object may use different cursors (e.g. links inside a text field).
             if is_mouse_moved && InteractiveObject::option_ptr_eq(cur_over_object, new_over_object)
+            {
                 new_cursor = match new_over_object {
                     Some(obj) => obj.mouse_cursor(context),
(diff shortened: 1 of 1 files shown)


Dinnerbone:
Looks like this is failing some tests. I guess it's updating the hovered item when it shouldn't be?

Meanwhile; do you have any content that can test this change?

robinmiau:
@Dinnerbone Sorry for the long wait.

Yes, that was the issue. I changed it so stationary mouse hover changes go through the existing rollover path instead, while preserving keyboard-focus hover. The failing tab-ordering tests pass now.

I've only tested it on my website, but it requires registration, so I'm not sure if that's helpful?

PR #23398, opened 2026-04-05
core: Fix focusing behavior in SWF 8- vs 9+

### LINKED ISSUE #23182: Inaccurate behavior with AVM1 text input focusing

ChrisCPI:
The remaining changes from #23244, fixes #23182

The behavior is explained in the added comment in player.rs, which has been revised to hopefully be more clear.


ChrisCPI:
Marking this as a draft for now, since I recently came across some other weird behavior (only happening in SWF version 9 specifically) which this PR doesn't account for, so I need to look into it some.

PR #23401, opened 2026-04-06
core: Disallow code execution with SWF version 0

kjarosh:
Executing code with SWF version 0 means that something's wrong, as version 0 is used for errored out movies and non-movies, which cannot contain code.

In the past, Ruffle had some bugs that caused code execution with SWF version 0 (e.g. see 33334537f39b9b3637bb77192154f6c7b2bfe76d).

This patch fixes one instance of SWF version 0 getting into an Activation, and adds a debug assertion for the SWF version to make sure that we don't introduce more such cases.

### DIFF
--- a/core/src/avm1/activation.rs
+++ b/core/src/avm1/activation.rs
@@ -239,6 +239,7 @@ impl<'a, 'gc> Activation<'a, 'gc> {
         callee: Option<Object<'gc>>,
         local_registers: &'a [Cell<Value<'gc>>],
     ) -> Self {
+        debug_assert!(swf_version > 0, "cannot execute code with SWF version 0");
         avm_debug!(context.avm1, "START {id}");
         Self {
             context,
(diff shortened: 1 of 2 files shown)


moulins:
Currently, nothing is preventing actual SWF files from specifying their version as 0, what should happen in these cases? (I'd guess the SWF should fail to load entirely, but who knows?)

kjarosh:
Good question, based on my tests, FP fails to load the SWF entirely, so I think we're fine to add this assertion.

It's a debug assertion anyway because there may be other instances of SWF version 0 being used in production somehow.

Lord-McSweeney:
I think we do have a check for SWF version 0, in `swf::read::decompress_swf`. Maybe it should be mentioned in the commit message?

kjarosh:
Added that info to the commit message.

PR #10630, opened 2023-04-10
avm2: Use proper scale for `Graphics.beginBitmapFill` (fix #10628)

Herschel:
The scale matrix was incorrectly using the bitmap dimensions when it should be scaled based on twips (x20). Fixes #10628.

FIXME future PR: This code is mostly duplicated between AVM1 and AVM2, would be nice to move the common code somewhere.

PR #23411, opened 2026-04-07
avm2: Use proper scale for GraphicsBitmapFill and lineBitmapStyle

ZingBallyhoo:
The same fix as https://github.com/ruffle-rs/ruffle/pull/10630, but for two additional methods.

Coincidentally the bitmap used with `lineBitmapStyle` in the existing `avm2/graphics_bitmaps` test is 20x20 pixels, making it invariant to this fix. (future work: that test is comparing against ruffle output, not flash player)

| Before | After|
|--------|--------|
| output old | image |
| image | image |
| image | image |


ZingBallyhoo:
Amended the test to use integer scaling & MSAA. I realise the `1.5` scale was needlessly pessimising the output accuracy.

kjarosh (approved):
LGTM, thank you!

PR #23448, opened 2026-04-13
tests: Use Flash output for visual/definefont4 and allow outliers

evilpie:


tests/tests/swfs/visual/definefont4/output.expected.png
kjarosh, line 1: This doesn't look good: there's no anti-aliasing, but in Flash there should be. If you reused the original image that might be the reason why, as it's 2 colors only.

tests/tests/swfs/visual/definefont4/test.toml
```diff
@@ -2,6 +2,7 @@ num_frames = 5
 
 [image_comparisons.output]
 tolerance = 1
+max_outliers = 23000
```
kjarosh: After fixing up the PNG, you can up the tolerance so that the number of outliers goes down (due to anti-aliasing).

tests/tests/swfs/visual/definefont4/test.toml
```diff
@@ -2,6 +2,7 @@ num_frames = 5
...
 
 [player_options]
-with_renderer = { optional = false, quality = "low" }
+with_renderer = { optional = false }
```
kjarosh: `optional = false` is the default too

tests/tests/swfs/visual/definefont4/output.expected.png
evilpie: Okay, not sure how I managed to paste an old image here.

tests/tests/swfs/visual/definefont4/test.toml
evilpie: Increasing the tolerance doesn't really help unless I use something crazy like 200. Maybe because this is caused by the text being offset and not from anti-aliasing?

output difference-color-linux-Vulkan

tests/tests/swfs/visual/definefont4/test.toml
evilpie: Do you mean using `with_renderer = { }`?

kjarosh: I'd use either `with_renderer = {}` or `with_renderer = { quality = "high" }` as IMO it's better to explicitly list the quality than optional.

tests/tests/swfs/visual/definefont4/output.expected.png
kjarosh: I'm still getting different output. For me, the text is positioned differently, as follows.

Image

The image is identical on FP 32 Linux and FP 32 Windows. Looks like yours is slightly translated and zoomed in...

evilpie: Do you just want to update this yourself then?

evilpie: Used your image the diff is even bigger now.

PR #23462, opened 2026-04-15
core: In MovieClip's `avm1_mouse_pick`, don't check for AVM2 content

ChrisCPI:
This is theoretically not possible, because AVM2 movies loaded into AVM1 are always interpreted as AVM1 (https://github.com/ruffle-rs/ruffle/pull/23183), and `mouse_pick_avm1` is never run in AVM2.

### DIFF
--- a/core/src/display_object/movie_clip.rs
+++ b/core/src/display_object/movie_clip.rs
@@ -2984,17 +2984,7 @@ impl<'gc> TInteractiveObject<'gc> for MovieClip<'gc> {
                     }
                 } else if result.is_none() {
                     if let Some(child) = child.as_interactive() {
-                        result = if !child.as_displayobject().movie().is_action_script_3() {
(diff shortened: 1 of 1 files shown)


kjarosh (approved):
LGTM

ChrisCPI:
Turns out this isn't entirely correct; SWF 9 allows for moving AVM1 content outside of the parenting Loader. Will revert the changes to `mouse_pick_avm2` and add a comment there explaining this.

kjarosh:
Can you add a test for that case?

ChrisCPI:
Moving the avm1 movie outside the loader is already covered in the `mixed_avm/avm2_loads_avm1_v9` test, but not specifically for mouse picking (it has some inaccuracies anyway, which I’m looking into, but won’t be for this pr).

kjarosh:
We should add a test for mouse picking specifically. The previous version of the PR should fail tests.

PR #23463, opened 2026-04-15
tests: Use output from Flash Player in more tests

kjarosh:
Fixed tests:
* `from_shumway/acid/acid-clip-2`,
* `from_shumway/acid/acid-color-2`,
* `from_shumway/acid/acid-filter-2`,
* `from_shumway/acid/acid-morph`,
* `from_shumway/acid/acid-shapes`.


tests/tests/swfs/from_shumway/acid/acid-filter-2/output.expected.png
evilpie, line 1: Why are the colors so different? Is this HDR or something like that?

kjarosh: In what way different? Different from what?

evilpie: The red was much more red before. Color picker says it was #ff0202 and now is #e43a21.

kjarosh: Are you sure about that `#e43a21`? Where are you getting it from? The new image is sRGB and should be `#ff0000` and that's what I'm getting when picking color from the new image :thinking:

Dinnerbone: I get `#FF0000` colour picking the new image too, fwiw

evilpie: Ah the the previous PNG was untagged. 

Image

Anyway looks the same (before/after) to me in Xviewer.

evilpie (approved):

PR #23468, opened 2026-04-16
core: Fix MovieClip's AVM2 mouse picking for AVM1 content

ChrisCPI:
Refactors the setting of the `res` value so that if `mouse_pick_avm1` on the child returns `None`, instead of counting it as a `Miss`, it will allow the `hit_test_shape` check to happen, so that the pick can correctly propagate to the parent.

Prior to this, an AVM1 child would not respect `mouseChildren` on its parent, and landing on a non-interactive AVM1 object would just go through to the stage (or whatever would happen to be below it).

I also plan on making a follow up PR to apply a similar fix to `LoaderDisplay`.


core/src/display_object/movie_clip.rs
```diff
@@ -3093,32 +3093,35 @@ impl<'gc> TInteractiveObject<'gc> for MovieClip<'gc> {
...
 
-                let mut res = if let Some(child) = child.as_interactive() {
-                    if child.as_displayobject().movie().is_action_script_3() {
-                        child.mouse_pick_avm2(context, point, require_button_mode)
-                    } else {
-                        let avm1_result =
-                            child.mouse_pick_avm1(context, point, require_button_mode);
-                        if let Some(result) = avm1_result {
-                            Avm2MousePick::Hit(result)
+                let mut res = (|| {
```
kjarosh:
```suggestion
                let mut res = {
```

You can use a block instead of creating a lambda and calling it immediately.

ChrisCPI: This appears to behave differently? The newly added test fails when its like this.

kjarosh: Interesting. What's the difference? What does the lambda accomplish here?

kjarosh: Oh, I just noticed the return statements... It's probably worth refactoring it because it's unnecessarily complex.

ChrisCPI: Would it be better to use a labeled block like this? Probably would be the simplest way to avoid the lambda.
```rs
let mut res = 'pick: {
...
if child.as_displayobject().movie().is_action_script_3() {
    break 'pick child.mouse_pick_avm2(context, point, require_button_mode);
} else {
...
```

ChrisCPI: Actually, I think I've figured something out.

ChrisCPI:
The only downside to this new approach is that it runs `mouse_pick_avm1` twice, but not sure how I could get around that.

ChrisCPI:
Another possible option, which moves the `hit_test_shape` check to its own function and runs that twice. Would this be better?

```rs
                let hit_test_shape = |context: &mut UpdateContext<'gc>| {
                    if child.hit_test_shape(context, point, options)
                        && child
                            .masker()
                            .map(|mask| mask.hit_test_shape(context, point, options))
                            .unwrap_or(true)
                    {
                        if self.mouse_enabled() {
                            Avm2MousePick::Hit(this)
                        } else {
                            Avm2MousePick::PropagateToParent
                        }
                    } else {
                        Avm2MousePick::Miss
                    }
                };

                let mut res = if let Some(child) = child.as_interactive() {
                    if child.as_displayobject().movie().is_action_script_3() {
                        child.mouse_pick_avm2(context, point, require_button_mode)
                    } else {
                        let avm1_result =
                            child.mouse_pick_avm1(context, point, require_button_mode);
                        if let Some(result) = avm1_result {
                            Avm2MousePick::Hit(result)
                        } else {
                            hit_test_shape(context)
                        }
                    }
                } else {
                    hit_test_shape(context)
                };
```

ChrisCPI:
Sorry for going back and forth on this, but I've finally landed on a solution I think I'm happy with.

PR #23475, opened 2026-04-17
avm2: Use Value::from_usize_lossy instead of From<usize>

kjarosh:
The From impl for Value was not compliant with the expectations: the conversion from usize to Value is not always lossless.

That's why this patch removes the impl and instead moves the conversion to Value::from_usize_lossy, which makes it clear that the conversion is lossy.

In most cases however, this shouldn't have any effect, as usually usize is used in relation to memory, and the first lossy value would happen at 9 PB of memory.

### DIFF
--- a/core/src/avm2/globals/array.rs
+++ b/core/src/avm2/globals/array.rs
@@ -69,7 +69,7 @@ pub fn get_length<'gc>(
     let this = this.as_object().unwrap();
 
     if let Some(array) = this.as_array_storage() {
-        return Ok(array.length().into());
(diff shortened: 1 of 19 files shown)


moulins (approved):
Hmm, a lot of these (especially when relating to array or string lengths) feel like values not fitting in `i32` should be either impossible or forbidden... maybe we should add things like `WStr::len32(&self) -> i32`?

Still, an explicit function is an improvement over the status-quo, so :+1:.

Lord-McSweeney (approved):

PR #23478, opened 2026-04-17
avm2: Start work on implementing XML.setNotification (attribute addition/change)

### LINKED ISSUE #23446: CollectionChange event not triggered by XMLListCollection when there is a change on an xml attribute.
### Describe the bug

When adding a listener to the CollectionChange event on an XMLListCollection, you'll see different behavior than the Adobe Flash Player runtime for certain types of changes to the collection.
For example, adding a new XML node with the "addItem" method will trigger the event correctly, while changing an attribute using an E4X expression in Ruffle will not trigger an event.

I created a simple MXML application that defines a collection, adds the listener, and then makes two changes to the collection. The event handler displays an event counter and the XML content of the collection formatted as a string. In Flash Player, two alerts are correctly displayed. In Ruffle, only the one related to the change with "addItem" is displayed.

This is what you see in Ruffle:
Image

This is what you see in Adobe Flash Player:
Image

### Expected behavior

Ruffle's behavior is expected to be identical to that of Flash Player, i.e., changing the XML attribute will also trigger the event and display the alert.

### Content Location

[application.zip](https://github.com/user-attachments/files/26676988/application.zip)

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

Mozilla Firefox 149

### Additional information

_No response_

evilpie:
Fixes #23446.

This is a non-document API that is used by https://github.com/apache/royale-asjs/blob/HEAD/frameworks/projects/MXRoyaleBase/src/main/royale/mx/utils/XMLNotifier.as

As a proof-of-concept I added support for two commands.


evilpie:
Looks like the notifications actually bubble up. Fixing that now.

evilpie:
> Looks like the notifications actually bubble up. Fixing that now.

That actually makes the SWF in #23446  work.

evilpie:
I didn't actually measure the performance of my previous implementation, but this now does less work for the common case of no notification function.

Dinnerbone (approved):

PR #23479, opened 2026-04-17
core: Simplify & support non-ident common strings

kjarosh:
This PR replaces the `define_common_strings!` macro-rules with a procedural macro that automatically generates field names from the provided byte string literals. This reduces duplication in `common.rs` by removing the need to manually specify `str_` prefixes for every common string.

After refactoring `define_common_strings!` to a proc macro, we can escape non-ident characters in strings and generate identifiers dynamically. This makes it possible to use any string as common string.

Finally, the PR adds more common strings that weren't added previously due to the fact that common strings didn't support non-ident strings.


core/macros/src/lib.rs
```diff
@@ -308,6 +308,99 @@ pub fn derive_has_prefix_field(input: TokenStream) -> TokenStream {
     .into()
 }
 
+/// Define the `CommonStrings` struct and its `new` method.
+///
+/// This macro generates the `CommonStrings` struct with fields for each of the
+/// given strings, prefixed with `str_`. It also generates a `new` method that
+/// interns all of these strings.
```
Lord-McSweeney, line 315: What happens someone accidentally causes a collision, such as by adding both `"forward-back"` and `"forward_back"`? If it causes an error, can that be noted here?

kjarosh: This will not cause a collision, because non-alphanumeric characters are escaped. So they will produce `forward_2dback` and `forward_5fback` respectively.

moulins: I think collisions are still possible with control characters <= 0x8: e.g., `"\x07c"` and `"|"` (ASCII 0x7C) will both produce `_7c`. This should be fixable by always using 2 hex digits in escape sequences, with `{:02x}`.

Now to be fair, we'll probably never hit this case in practice, but still ^^

moulins:
I'm not a fan of how this moves the actual `struct CommonStrings` definition outside of the `avm_string/common.rs` file...  
Instead, what if `common_string_to_ident` was directly exposed as a proc-macro? Then `define_common_strings` can stay as a normal decl-macro, and include things like `common_string_to_ident!($field): AvmString<'gc>` in its expansion.

core/macros/src/lib.rs
kjarosh: Right, fixed!

kjarosh:
> I'm not a fan of how this moves the actual `struct CommonStrings` definition outside of the `avm_string/common.rs` file

Doesn't it make more sense for the structure to be defined in the same file it's used? I think it makes more sense for `istr!` to be in the same place as `define_common_strings!`.

moulins:
Well, ideally `istr!` would *also* be a decl-macro defined in terms of `common_string_to_ident!`, but this can't be done easily as the short form (`istr!("foo")`) isn't hygienic.

> Doesn't it make more sense for the structure to be defined in the same file it's used?

With your changes, the declaration *also* get split: e.g. `define_common_strings!` still depends on having `ASCII_CHARS_LEN` and `ASCII_CHARS` in scope, which makes understanding how the code works much more difficult imho.

Sure, this could be fixed by moving the entirety of the `common.rs` file inside the proc-macro, but that would makes thing *even* more opaque to the average reader (and would probably worsen IDE experience).

kjarosh:
> Sure, this could be fixed by moving the entirety of the `common.rs` file inside the proc-macro, but that would makes thing _even_ more opaque to the average reader

I'm not convinced that's the case, the structure and its usage will be in the same file anyway. I think we can just more the existing code as-is before the `struct`. Doesn't make sense to make the implementation detail public. Placing definitions of `istr` and `define_common_strings` in completely different crates would be more misleading.

> (and would probably worsen IDE experience).

In what way? I didn't notice anything particularly bad.

moulins:
A "go-to" definition on the `CommonStrings` type won't bring anything useful: it'll point to the `define_common_strings! { ... }` macro call, and the actual type definition (including its constructor) will be completely hidden inside the proc-macro code.

kjarosh:
Done, left the struct definition in a macro_rules.

moulins (approved):

PR #21925, opened 2025-10-14
avm2: Allow any i32 in `Value::Integer`

kjarosh:
This is a performance optimization, as it gets rid of range checks when operating on values. Since we already implement value normalization, we don't have to make sure the value is the proper type at all times, only when differentiating between Number and Integer. This allows us to drop range checks and allow any Integer and any Number even if they are not representable in Flash Player.

This should improve performance in most cases, where we don't need a range check. This introduces overflow checks in `op_add` and `op_subtract`, but they should be faster than range checks which happened after each op anyway.

PR #23491, opened 2026-04-19
avm2: Remove `pushshort` op

Lord-McSweeney:
This is possible after #21925 and unblocks some additional (future) peephole optimizations for us.

Also make `contains_valid_integer` and `contains_valid_unsigned` track those states more accurately.

### DIFF
--- a/core/src/avm2/optimizer/type_aware.rs
+++ b/core/src/avm2/optimizer/type_aware.rs
@@ -903,21 +903,17 @@ fn abstract_interpret_ops<'gc>(
             }
             Op::PushInt { value } => {
                 let mut new_value = OptValue::of_type(types.int);
-                if value < -(1 << 28) || value >= (1 << 28) {
-                    // will be coerced to Number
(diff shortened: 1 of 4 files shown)


SuchAFuriousDeath (approved):

core/src/avm2/optimizer/type_aware.rs
```diff
@@ -901,31 +901,19 @@ fn abstract_interpret_ops<'gc>(
...
-                        new_value.contains_valid_unsigned = true;
-                    }
-                }
-                stack.push(activation, new_value)?;
-            }
             Op::PushUint { value } => {
                 let mut new_value = OptValue::of_type(types.uint);
-                if value < (1 << 28) {
+                new_value.contains_valid_unsigned = true;
+                if value < (1 << 31) {
```
kjarosh, line 915: Can you use `i32::MAX` instead?

kjarosh (approved):

Issue #22861, opened 2026-01-27
Poor font rendering despite font embedding


falpi:
I'm impressed with the level of compatibility achieved by Ruffle. I've almost managed to get a complex enterprise application running.

One remaining issue concerns the rendering of bitmap fonts, which are blurry and don't adhere to the original spacing despite following the font embedding instructions. I embed the fonts by adding the following instructions.

```
<script>
   window.RufflePlayer.config = {
      autoplay: "auto",
      fontSources: ["TahomaAll.swf"],
      defaultFonts: {
         sans: ["Tahoma"]
      }
   };
</script>
```

Below I attach the zip package containing the swf file with the embedded fonts and a screenshot showing the rendering with the flash player (on the left) and with ruffle (on the right).

[TahomaAll.zip](https://github.com/user-attachments/files/24885129/TahomaAll.zip)

<img width="1768" height="464" alt="Image" src="https://github.com/user-attachments/assets/c4fc2279-4691-4b8b-9543-a54703073183" />

I hope you can improve this aspect of Ruffle, which is important for legacy enterprise applications. 
It would be great to have pixel-precise fonts again in Ruffle.

kjarosh:
Does the font change in Ruffle when you configure `TahomaAll.swf` vs when it's not configured?

kjarosh:
Also can you include the original SWF and reproduction steps?

falpi:
Hi, yes, the rendering changes and improves with font embedding. The runtime does indeed read the fonts and can render both normal and bold text, but the rendering is still inaccurate.

I can't help you reproduce the error because it's a complex enterprise application that requires a PHP web service layer at startup. However, I've sent you the SWF file for the application in question in the other issue : https://github.com/ruffle-rs/ruffle/issues/22860

kjarosh:
So the main issue is poor support for device fonts on web, it's probably going to improve when we finish implementing canvas device font renderer.

To try it out (it will still look not quite right I'm sure), you can set `deviceFontRenderer: "canvas"` (the canvas renderer won't use your provided SWF with fonts).

falpi:
I wasn't aware of the option you indicated. I can confirm that by adding it now, the rendering is still far from what I expected. 

<img width="848" height="443" alt="Image" src="https://github.com/user-attachments/assets/4909f953-a6ee-4c6c-b948-20aecd603926" />

I eagerly await the completion of this implementation.
Thanks guys, keep it up, you're doing a great job!

[...]

PR #23492, opened 2026-04-19
core: Implement a custom font renderer which use pluggable external modules

falpi:
This proposal introduces the implementation of a custom font renderer that allows rendering logic to be externalized into pluggable modules.

Plugable modules must adhere to a simple, well-defined contract, but can be developed in a variety of ways. In the proposal, I've included two working renderer examples, located in the new "font-renderers" folder.

The first module is called "gdi-font-renderer" and is based on napi-rs because it directly uses the "gdi32.dll" library to achieve rendering identical to that of classic GDI. Of course, it only works on Electron on the Windows platform.

The second module is called "swash-font-renderer" and is based on Rust code compiled with wasm-bindgen. It uses the popular "swash" library, which allows for better rendering of small fonts than what is currently implemented in Ruffle. This module can run in any browser.

The impact on Ruffle's source code is minimal, precisely because the rendering logic is delegated to pluggable modules. However, over time, I realized that it was necessary to introduce size-aware glyph caching to allow renderers that prefer to render fonts at the required size. Currently, Ruffle's code caches and renders glyphs at a fixed size of 64px and then scales the bitmap, which isn't optimal for achieving superior quality even with small fonts. This was a necessary choice for GDI rendering in particular, to allow for classic Flash Player rendering (as discussed in issue #22861).

Specifically, regarding the GDI renderer, the reason I created it was due to a strong suspicion that even after the Canvas renderer is completed, rendering in Chrome/Chromium browsers will be different and worse than other browsers like Firefox. This is because, since the removal of the "disable-directwrite" flag in 2014 (https://www.change.org/p/google-inc-bring-sharp-fonts-back-in-google-chrome-disable-directwrite), rendering on Windows, even without ClearType, is terrible, especially with small font sizes. The GDI renderer I'm providing here restores pixel-perfect rendering, at least on Chromium/Electron where I urgently needed it.


evilpie:
In you own words, why do you need this?

falpi:
> In you own words, why do you need this?

Hi, the requirement arose from the need to support a legacy Electron-based enterprise application that uses very small fonts, so that Flash Player could be abandoned and the runtime could be updated to the latest versions. This is what initially drove me to develop the GDI-based renderer, which uses napi-rs to interface with the system DLL and is therefore only usable on Electron.

I took the opportunity to generalize the implementation so as not to limit it to the use of Electron, making it possible to implement alternative renderers to the integrated ones, for example, allowing the reuse of excellent libraries like Swash.

This also limited the impact on Ruffle's source code, hopefully contributing to improved pull request acceptance. The size-sensitive cache will also be useful for improving other integrated renderers, which currently aren't as high-performance due to fixed-size rendering.

Best regards

PR #23499, opened 2026-04-20
core: Fix some corrupted audio by preventing negative volume

### LINKED ISSUE #16234: [NSFW] Dangerous audio corruption in early Super Deepthroat
### Describe the bug

Tested with Super Deepthroat [1.14.1b](http://swfchan.com/21/103470/?SuperDeepthroat1_14_1b.swf), 1.14.3b. Most likely would happen on 1.14+ versions, but those are unable to get past the title screen.

While the player is actively moving Her, a good portion of the sounds played are heavily corrupted and very loud. Muting the game within its settings prevents this, but even on low volumes it is still uncomfortably loud.

Here is a video of it in action: https://files.catbox.moe/v00bif.mp4 (NSFW)

Nothing notable in the Console.

### Expected behavior

Sounds play at a nice and comfortable level, where no one's eardrums are hurt.

### Content Location

**Run this with very low audio only.**
http://swfchan.com/21/103847/?SuperDeepthroat1_14_3b.swf
Not sure how to force swfchan to use a newer version, but the .swf can be downloaded by clicking Save underneath the game window.

### Affected platform

Desktop app

### Operating system

Windows 10 22H2

### Browser

Mozilla Firefox 125.0.3 (64-bit)

### Additional information

Tested with Ruffle Nightly 2024-05-06 in both Extension and Standalone.

iscwDu:
Fixes #16234

I'm not a rust programmer, if there is a more idiomatic way to do this, please let me know. Thanks! (Edit: I did compile and run this against SDT to confirm that it fixes the audio issues, with and without '--release' )

I also wasn't sure if negative volumes should be clipped to zero or just treated as positive. This makes more sense to me as it looks like they could arise when doing some weird audio panning.

### DIFF
--- a/core/src/display_object.rs
+++ b/core/src/display_object.rs
@@ -3162,7 +3162,7 @@ impl SoundTransform {
             left_to_right: (sound_transform.left_to_right() * 100.0) as i32,
             right_to_left: (sound_transform.right_to_left() * 100.0) as i32,
             right_to_right: (sound_transform.right_to_right() * 100.0) as i32,
-            volume: (sound_transform.volume() * 100.0) as i32,
+            volume: i32:abs((sound_transform.volume() * 100.0) as i32),
         }
     }
 


kjarosh:
Can you add a test for it? See for details:
1. https://github.com/ruffle-rs/ruffle/blob/master/CONTRIBUTING.md#test-guidelines
2. https://github.com/ruffle-rs/ruffle/blob/master/tests/README.md

We'd probably need a test that sets negative volume and then reads it back and plays audio. Then we can verify if the audio is playing. This way you'll know whether that's the right place to fix it and whether you should clamp it or abs it.

iscwDu:
Abandoning this change as it was better solved in this [commit](https://github.com/ruffle-rs/ruffle/commit/9e5d3f569c30da5a22c18f2ed99c2709ac256457) by Natalie Raine.

PR #23504, opened 2026-04-22
web: Unwrap unneccessary non-capturing RegEx group

tygyh:


Dinnerbone (approved):
Seems valid. Thank you!

PR #23506, opened 2026-04-22
avm2: Fix incorrect numbering

tygyh:


kjarosh:
License files are generated (and not up to date I guess), so we don't care about their formatting.

core/src/avm2/globals/README.md
```diff
@@ -116,8 +116,8 @@ To add versioning to an API:
 
 1. Determine the first version where it was added. This can be seen in the Flash Documentation (e.g. "Runtime Versions: Flash Player 11.4, AIR 3.4")
 2. Convert the Flash Player version to an SWF version number using [this chart](https://github.com/ruffle-rs/ruffle/wiki/SWF-version-chart)
-2. Determine the corresponding asc.jar version code for the SWF version. This can be found in avmplus in https://github.com/adobe/avmplus/blob/master/core/api-versions.as
-3. Add an `[API("VersionCode")]` metadata to the definition. In the `Event.WORKER_STATE` example,
+3. Determine the corresponding asc.jar version code for the SWF version. This can be found in avmplus in https://github.com/adobe/avmplus/blob/master/core/api-versions.as
```
kjarosh, line 119: You can use 1. everywhere instead of numbering manually.

tygyh: Is there a reason to not have it be correct?

ItzBloxxy: kjarosh stop being lazy

kjarosh: > Is there a reason to not have it be correct?

Not sure what you mean by "correct", all 3 options result in the same rendering. The reason why you'd use 1. everywhere is because you don't have to number items manually, as I said. And we can see the manual numbering already has failed.

tygyh: Correct as in the numbers incrementing in the file, not just in the rendering. I could go through and change change it all to 1 if that is the preferred style, but it seems weird. I'll wait to see if someone else has an opinion on this.

PR #23507, opened 2026-04-22
web: Add required 'lang' attributes

tygyh:
Required according to [Web Content Accessibility Guidelines 2.0](https://www.w3.org/TR/WCAG20-TECHS/)

### DIFF
--- a/web/packages/extension/assets/player.html
+++ b/web/packages/extension/assets/player.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 
-<html>
+<html lang="en-US">
     <head>
         <meta charset="utf-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />


web/packages/extension/assets/onboard.html
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 
-<html>
+<html lang="en-US">
```
kjarosh, line 3: This page is localized, setting en-US when a different language is selected will be misleading.

web/packages/selfhosted/test/polyfill/swf_object/index.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="en-US">
```
kjarosh: We don't have to add the attribute in tests, they were created without the attribute. If you think the attribute affects our behavior, you should add a new test with the attribute, but I doubt it does.

web/packages/extension/assets/onboard.html
tygyh: I don't see how it is localized as the file uses plain text instead of variables. Afaik the standard for keeping pure HTML accessible is to use different files for each language since screen readers don't work without this tag

kjarosh: Based on what I see, the JS localizes this page based on IDs.

danielhjacobs: We can't use different pure HTML files as the localization comes from Crowdin. The localized strings come from https://github.com/ruffle-rs/ruffle/tree/master/web/packages/extension/assets/_locales, then we use https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/i18n to do the localization. You can see an example usage in https://github.com/ruffle-rs/ruffle/blob/0e83bf57a3413b4a5861312349e8a27a2eb103ac/web/packages/extension/src/common.ts#L166 and https://github.com/ruffle-rs/ruffle/blob/0e83bf57a3413b4a5861312349e8a27a2eb103ac/web/packages/extension/src/options.ts#L14

danielhjacobs: I think we just want to set this in the JS of all the files doing other i18n in the extension: `document.documentElement.lang = utils.i18n?.getUILanguage?.() || "en-US";`. The player.ts change can stay for now, as it's not localized.

danielhjacobs: All but the `player.html` change have been superseded by #23519.

kjarosh (approved):

Issue #1006, opened 2020-08-15
Internationalization/localization


Justin-CB:
We have a decent amount of text in our settings dialog and page, which will need to be localized eventually. Someone suggested a platform for this on the discord, but I don't really have any experience with localization/internationalization, except for editing the `messages.json` files manually (and a little on the translator side but not enough to know what platform it was they were using or whether it was any good. I would like to know what everyone's thoughts are on this matter and whether we have anyone willing to help translate into other languages (l know a little Spanish, but I don't know what languages, if any, the other members speak since we always communicate in English).

veikk0:
I suggest [translatewiki.net](https://translatewiki.net/), a translation community for free and open-source projects. I've used it for a while as a volunteer translator and I find the interface simple and easy-to-use. It uses the [Translate](https://www.mediawiki.org/wiki/Extension:Translate) MediaWiki extension, which has a lot of features that make translators' work easier.

The extension is also in use on various Wikimedia wikis, like MediaWiki.org, Meta-Wiki, and Wikimedia Commons, so the tech & UI itself isn't used only by this single website.

[Here's their intro page for developers](https://translatewiki.net/wiki/Translating:New_project). In terms of requirements, they require commit access so that @translatewiki can regularly integrate translation updates into the project. They also prefer that projects have i18n support. They have [a tips page](https://translatewiki.net/wiki/Translating:Localisation_for_developers) for developers and software project managers.

Disclaimer: I only have recent experience with this single translation platform, so take my suggestion just as that, a suggestion.

Justin-CB:
> I think with Crowdin set up (#10766, https://crowdin.com/project/ruffle) this can now be considered solved.

You're right, thanks.

PR #23511, opened 2026-04-22
avm2: Check prototype chain in with-scope property resolution

bjornwein:
## Problem

When AS3 code uses XML methods like `name()`, `children()`, etc. inside a `with` block (`pushwith` opcode), the compiler generates `QName(PackageNamespace(""), "name")` lookups (public namespace). However, Ruffle's `search_scope_stack` only checked:

1. **vtable traits** — XML's vtable has `name()` in the AS3 namespace (`http://adobe.com/AS3/2006/builtin`), not the public namespace → no match
2. **own properties** (for with-scopes) — XML's `has_own_property` checks E4X child elements → no match

This caused `findpropstrict` to skip the XML with-scope entirely and find the property on a parent scope instead (e.g., `DisplayObject.name` getter returning a string), leading to `Error #1006: value is not a function` when trying to call the string result.

## Root Cause

In avmplus, `hasMultinameProperty` for with-scope resolution checks own properties, traits, **and the prototype chain (delegate)**. Ruffle was missing the prototype chain check. This matters because `XML.prototype.name` (and other XML methods) are defined as public-namespace dynamic properties on the prototype in `XML.as` (lines 183-190), bridging the AS3→public namespace gap.

## Fix

Added `value_has_proto_property()` to `search_scope_stack` that walks the prototype chain checking for dynamic properties. This is called for with-scopes after the existing own-property check, matching avmplus behavior.

## Testing

Tested against a real-world Flex-based SWF (hardware appliance UI) that was completely broken by this bug — the app now loads and functions correctly past login. The `Error #1006` in `updateTreeExpansion()` is eliminated.

Existing tests pass (`cargo test --release --package ruffle_core`).


evilpie:
Please add a test, thank you.

bjornwein:
Thanks for looking into it. I apologize for dropping this LLM-generated PR on you.

I can confirm that the change solves a real-world problem for me. Unfortunately I don't have the knowledge to evaluate whether this is the best way to solve the specific issue.
I hope that your review will be able to spot any obvious significant performance costs or regression risks.

I (or mostly copilot) have updated the PR with a minimal-ish regression test. The test fails with error code #1065 without the fix.

tests/tests/swfs/avm2/xml_with_scope_prototype/Test.as
```diff
@@ -0,0 +1,29 @@
+// This test verifies that XML prototype methods (name, localName, nodeKind,
```
Lord-McSweeney, line 1: Instead of accessing methods on `XML`'s prototype, why not create a class C like
```
class C {
    {
        prototype.prototypeProperty = "abc";
    }
}
```
then try to lookup `prototypeProperty` inside `with(myInstanceOfClassC)`? This would allow for avoiding manually generating a SWF from Rust.

core/src/avm2/scope.rs
```diff
@@ -354,3 +361,23 @@ fn value_has_own_property<'gc>(
...
+// hasMultinameProperty which checks own properties, traits, AND the prototype
+// chain (delegate). This is particularly important for XML/XMLList objects
+// where methods like name(), children(), etc. are accessible via public
+// namespace prototype functions, even though the vtable traits are in the
+// AS3 namespace.
+fn value_has_proto_property<'gc>(value: Value<'gc>, multiname: &Multiname<'gc>) -> bool {
+    if let Value::Object(object) = value {
+        let mut proto = object.proto();
+        while let Some(p) = proto {
+            if p.base().has_own_dynamic_property(multiname) {
```
Lord-McSweeney, line 376: Are you sure this shouldn't be `p.has_own_property` instead of `p.base().has_own_dynamic_property`? i.e. if the `prototype` is a class with non-dynamic properties, are those non-dynamic properties considered?

Lord-McSweeney:
Thank you for the contribution!

PR #23513, opened 2026-04-22
Allow compiling AVM1 tests programatically with Rascal

Dinnerbone:
This is largely a POC before I convert any other tests to using it. Do we want this? Is it useful? What can we change to make it more useful?

The benefits in my head:
- Sometimes the swf and as aren't always kept up to date (we've changed just one or the other in the past, without realizing). This prevents that.
- Sometimes we don't provide the source at all, for older stuff there's just a `fla` and you're kinda on your own for editing that. We can swap those over.
- mtasc and old Flash versions aren't super duper easy to get a hold of - and I don't like high barrier of entry for test writing. This has no barrier.
- Rascal is more correct in places than mtasc, and supports more language features (Granted - mostly obscure :D)
- Rascal allows for writing tests entirely in pcode, or embedding random pcode within a test, which reduces the need for fiddling with tests in FFDEC
- If a test is just "edit file and run test", we can easily add more test cases as we discover them - whereas today it's slightly more of a hassle


tests/framework/src/compiler.rs
```diff
@@ -0,0 +1,68 @@
...
+            .write_all(&swf)?;
+        Ok(())
+    }
+}
+
+impl Default for RascalOptions {
+    fn default() -> Self {
+        Self {
+            target: "test.swf".to_string(),
+            swf_version: 15,
```
moulins: Why version 15? That doesn't sound like a valid AVM1 SWF version (and it doesn't match with the `README.md`, which says `10`)

Dinnerbone: Any SWF version is a valid AVM1 SWF version! My copy of Flash isn't starting right now but it produces at least SWF 15, as that's what I've been testing with. I think it might go up to 23 if we cross reference CS6 (last Flash editor with AVM1 support) with our version chart (CS6 = FP12 = SWF 23).

For why 15 - I don't know, totally arbitrary. I opened some SWFs and found that I was working in 15 more than any other I guess. We can change that trivially.

moulins: Ok then! Though it'd still be best if the value mentioned in the README matched the actual default.

Dinnerbone:
Per discord discussions, I've added a binary target of `tests` which currently only allows compiling (not running) of every test
Invoke it via `cargo run -p tests -- compile`

tests/framework/src/compiler.rs
Dinnerbone: Almost every value in the readme don't match the defaults, they're example values
I'll change it though!

tests/README.md
```diff
@@ -225,6 +225,17 @@ japanese_mincho = ["Test Font", "Test Font Fallback"]
 # Fields specified here will override the 'default' values provided in the rest of the document.
 [subtests.SUBTEST_NAME]
 # ...
+
+# To make tests easier to maintain, we encourage programatic compilation of source files where possible
+# This is currently limited to pure AVM1 code
+# Multiple compiler steps can be included in a single test
```
kjarosh: Nit: missing periods at the end of sentences

tests/README.md
```diff
@@ -225,6 +225,17 @@ japanese_mincho = ["Test Font", "Test Font Fallback"]
...
+[[compilers]]
+type = "Rascal" # The only compiler currently supported
+target = "test.swf" # The SWF to compile
+scripts = ["test.as", "other.as"] # A list of AS scripts (not classes) to compile
+classes = ["foo.Bar", "SomeClass"] # A list of class names (not paths) to compile
+swf_version = 15 # The SWF version to compile for (see https://github.com/ruffle-rs/ruffle/wiki/SWF-version-chart)
+frame_rate = 24 # The frame rate to compile the swf with
```
kjarosh: Other options have comments above them, that makes more space for documentation.

tests/framework/src/compiler.rs
```diff
@@ -0,0 +1,68 @@
...
+pub struct RascalOptions {
+    pub target: String,
+    pub swf_version: u8,
+    pub frame_rate: f32,
+    pub scripts: Vec<String>,
+    pub classes: Vec<String>,
+}
+
+impl RascalOptions {
+    pub fn compile(&self, root_dir: &VfsPath) -> anyhow::Result<()> {
```
kjarosh: It's a bit weird for `RascalOptions` to be able to compile something. Have you thought about moving `compile` to `SwfCompiler`, or making `SwfCompiler` a trait with a separate enum for compiler selection?

tests/framework/src/compiler.rs
```diff
@@ -0,0 +1,68 @@
...
+            .join(&self.target)?
+            .create_file()?
+            .write_all(&swf)?;
+        Ok(())
+    }
+}
+
+impl Default for RascalOptions {
+    fn default() -> Self {
+        Self {
```
kjarosh: Does it make sense to provide defaults at all? It's always better to be explicit, even if it means adding `swf_version = 15` everywhere. It also doesn't make sense to have no scripts to compile, no?

tests/README.md
Dinnerbone: Amended, with other doc feedback too

tests/framework/src/compiler.rs
Dinnerbone: Many of our tests don't use scripts - mtasc style would use classes instead.
(script = loose code, class = strictly a class, likely with an mtasc style entry point)

As for defaults; everything else in options has default values today, that'd be a stylistic change. I'm not immediately opposed to it, but it will mean boilerplate and more noise when adding new fields. I agree that it's nice to be explicit about swf version, but almost everything is going to have an empty `scripts` OR an empty `classes`, target being `test.swf`, and a frame rate of "whatever".

tests/framework/src/compiler.rs
Dinnerbone: I gave it a super brief consideration, but given that there's only 1 thing and we don't know what a second would look like yet, I didn't want to overengineer this. We can move compile to not be part of the struct instead?

tests/tests/swfs/avm1/movieclip_lockroot/test.toml
```diff
@@ -1 +1,11 @@
 num_frames = 10
+
+[[compilers]]
+type = "Rascal"
+target = "test.swf"
```
kjarosh, line 5: The differences I see between those SWFs:
1. display rect is 100x100 instead of 550x400,
2. it's uncompressed,
3. it uses `DefineFunction` instead of `DefineFunction2`,
4. it doesn't have file attributes.

I guess it would be better if the SWF was compressed, and probably using 550x400 as the default makes more sense as it's the default in Flash (I think?).

Are 3 and 4 desirable?

Dinnerbone: 1. We can add an option for that!
2. If we compress it, we need to guarantee that the compression is deterministic. I do not know enough about the zlib compression library we're using today to say if that will be the case. Uncompressed is safest.
3. Yeah, whether or not any given swf today uses `DefineFunction` over `DefineFunction2` is "was it compiled with debugging enabled" in Flash, or mtasc which always goes for 2 I think. I haven't added the optimization pass which converts things to `DefineFunction2` yet, which is only present when _not_ compiling for debug in Flash.
4. No opinion here. We can add it? Should it be configurable?

tests/framework/src/compiler.rs
kjarosh: Maybe I misunderstood how `classes` works then. You do not pass the filename when using `classes`? What's the difference between having a `source.as` with a class named `Test` and a `Test` in `classes`?

> As for defaults; everything else in options has default values today, that'd be a stylistic change. [...]

I don't think I follow. I meant that implementing `Default` is fine, but shouldn't provide default values (e.g. `swf_version`). You can always use an `Option` and raise an error when it's not present when validating the config. Same for empty `scripts` and `classes` at the same time.

Dinnerbone: Scripts are paths to files containing loose code. In flash this is what you use if you put code on a timeline. Mtasc has no equivalent to them. No classes are allowed in scripts.

Classes are class names (not filepath) to load up. If one of them has a main method, it'll insert a script that calls the main method, to remain compatible with mtasc tests.

This separation exists because we have some tests that are "scripts" (likely all authored in actual Flash) and some that are classes only (likely all authored in mtasc) and we need a way to support both without changing the contents of the code.


For default: yeah sure, let's make swf version required then!

Dinnerbone: Made swf version required

tests/framework/src/compiler.rs
Dinnerbone: I added a `SwfCompiler` trait and cleaned up the files a bit.

tests/tests/swfs/avm1/movieclip_lockroot/test.toml
Dinnerbone: Added options for stage rect and use_local_sandbox (the only flag afaik that's relevant, in FileAttributes)

tests/tests/regression_tests.rs
```diff
@@ -58,32 +63,40 @@ fn main() {
...
-        .with_root_dir(PathBuf::from("tests/swfs"))
-        .with_test_loader(Box::new(move |params, register_trial| {
-            for test in load_test_dir(&params.test_dir, params.test_name) {
-                let trial = trial_for_test(&ruffle_test_opts, test, params.args.list);
-                register_trial(trial);
-            }
-        }));
+    let compile_mode = ruffle_test_opts.compile_mode;
+
+    {
```
kjarosh: This block can be removed.

tests/tests/regression_tests.rs
```diff
@@ -58,32 +63,40 @@ fn main() {
...
+        runner
+            .with_descriptor_name(Cow::Borrowed(TEST_TOML_NAME))
+            .with_root_dir(PathBuf::from("tests/swfs"))
+            .with_test_loader(Box::new(move |params, register_trial| {
+                for test in load_test_dir(&params.test_dir, params.test_name) {
+                    let trial = trial_for_test(
+                        NativeEnvironment { compile_mode },
```
kjarosh: Can you move out `NativeEnvironment` so that it's not instantiated separately for every test?

Dinnerbone: I tried but... not easily :( Lots of lifetimes at play here. The absolute best I can do is make it `Clone`able, I think, but I felt uneasy about that.

Dinnerbone:
Cleaned up commit history & removed all CLI stuff from this, will open a new PR for that. Sorry moulins :D

tests/README.md
```diff
@@ -256,6 +256,11 @@ frame_rate = 24
 # The display region of the stage, in pixels. You can usually get away with just specifying max, not min.
 stage_rect = { x_min = 0.0, y_min = 0.0, x_max = 550.0, y_max = 400.0 }
 
+# Whether the SWF should use the network sandbox.
+# If true, the SWF can access the network but not local files.
+# If false (default), the SWF can access local files but not the network.
+use_network_sandbox = false
```
kjarosh: I think the name is a bit misleading: it sounds like we're sandboxing the network (which is not true). I think this is supposed to refer to the security sandbox, but we actually cannot select (inside the SWF) which security sandbox is used. That's because it's the player that decides which security sandbox is being used based on several factors, one of them being whether the SWF requires network to work (useNetwork is true in attributes). In case of remote content, this attribute doesn't even matter. I think an option named `use_network` is enough with a doc that mentions the security sandbox (but skipping the reference to security sandbox altogether is fine as well).

Dinnerbone: I named it this because it's what we call that flag inside the swf and elsewhere in the codebase. I can rename it, no worries.

kjarosh: That's probably something we should improve elsewhere too :thinking: I'll have that in mind when working on security sandboxes

Dinnerbone: Renamed.

kjarosh (approved):
Thank you, that looks great! I'm excited we'll be able to automate compilation.

tests/tests/regression_tests.rs
kjarosh: We'll have to make it clonable anyway, but hopefully https://github.com/ruffle-rs/ruffle/pull/23550 will take care of it

torokati44:
Okay, this is super awesome, like the Rascal project itself! 😎

PR #23515, opened 2026-04-23
core: Make asfunction: URL prefix case insensitive (close #23514)

### LINKED ISSUE #23514: Html HREF buttons tied to actionscript functions are treated as external URLs.
### Describe the bug

### Expected behavior

### Content Location

### Affected platform

Desktop app

### Operating system

Windows 11 24H2 26100.8246

### Browser

_No response_

### Additional information

R3velation:
The `open_url` function now temporarily converts to lowercase prior to matching for asfunction.  This makes the parse for asfunction in the URL prefix case insensitive.


R3velation:
This is lower-casing everything, not just the asfunction prefix.  I need to go back and figure out a different way to do this.

PR #23517, opened 2026-04-23
core: Make asfunction and event URL prefix case insensitive (close #23514)

### LINKED ISSUE #23514: Html HREF buttons tied to actionscript functions are treated as external URLs.

R3velation:
In AVM1, executing a function from a URL is case insensitive.  The original implementation used `strip_prefix` which is case sensitive.

This patch changes the one-liner `strip_prefix` out for a string split at the point where the `asfunction:` section would end, and then matches with `eq_ignore_case`.  This selects the required prefix without clobbering the case on the address.

This pull request supersedes #23515 and will close #23514.


kjarosh:
Can you add a test for that? Might be also worth testing whether whitespace is stripped around the function, etc.

R3velation:
So both as1/asfunction and as2/textfield_event are handled case insensitively now and I've added tests to flag regressions.  As for white space,  no changes are needed.  Adding white space to the function breaks the reference in standard flash.  Doing the same to asfunction or event breaks the tag entirely and kick you out to a browser.  This is already the normal implemented behavior in Ruffle.  

Don't mind the unverified commits, that's a rebase to the current origin as recommended in CONTRIBUTING.md

This should be ready for review if there isn't anything missing.

Edit: No idea whats going on with avm1/netstream_seek_flv.  It tests good on my windows machine, even after rebase.

core/src/display_object/edit_text.rs
```diff
@@ -2468,17 +2468,27 @@ impl<'gc> EditText<'gc> {
...
-                error!("Couldn't execute URL \"{url:?}\": {e:?}");
-            }
-        } else if let Some(address) = url.strip_prefix(WStr::from_units(b"event:")) {
-            if let Some(object) = self.object2() {
-                let mut activation = Avm2Activation::from_nothing(context);
-                let text = AvmString::new(activation.gc(), address);
-                let event = Avm2EventObject::text_event(&mut activation, "link", text, true, false);
-
-                Avm2::dispatch_event(activation.context, event, object.into());
+        if let Some(prefix_end) = url.find(WStr::from_units(b":")) {
```
kjarosh: I think it would be clearer to define a `strip_url_prefix` function inside `open_url` that will strip the prefix ignoring case. Majority of the code would be left as is.

R3velation: I've moved it into a nested function as recommended.

core/src/display_object/edit_text.rs
```diff
@@ -2468,11 +2468,24 @@ impl<'gc> EditText<'gc> {
     }
 
     fn open_url(self, context: &mut UpdateContext<'gc>, url: &WStr, target: &WStr) {
-        if let Some(address) = url.strip_prefix(WStr::from_units(b"asfunction:")) {
+        fn strip_url_prefix<'a>(url: &'a WStr, prefix: &'a WStr) -> Option<&'a WStr> {
+            if let Some(prefix_end) = url.find(WStr::from_units(b":")) {
+                let (stripped_prefix, address) = url.split_at(prefix_end + 1);
```
kjarosh: You can use `split_once` instead of `find` and `split_at`.

kjarosh: As a side note, you can do it more efficiently by just getting the prefix with length of `prefix.len()` and then comparing it ignoring case with `prefix`. You wouldn't have to split (which is linear wrt the whole string). Comparing the prefix is linear wrt the prefix and fails early.

tests/tests/swfs/avm2/textfield_event/Test.as
```diff
@@ -14,6 +14,7 @@
 				"<p>Or click <a href='event:Second Test'>here</a> to call alert 2</p>" +
 				"<p><a href='event:'>this one is empty</a></p>" +
 				"<p>But <a href='event:a,b,c,d'>this one</a> has lots of args!</p>" +
+				"<p>And <a href='Event:Caps Test'>this one</a> has odd capitalization.</p>" +
```
kjarosh: Honestly, I'm a bit worried about modifying those tests, as by comparing SWFs I can see that the output differs in placement. It would be safer to create new tests—you can base them on these, just don't modify them.

core/src/display_object/edit_text.rs
R3velation: I don't see how a split is avoidable here.  `execute_avm1_asfunction` requires the address stripped of its prefix.  I don't see a way to provide that without some form of split, either in this or after the match.  I can use `WStr::slice` and defer creation of the address to after a successful match.  If I'm misunderstanding what your saying here, please do correct me.

On another note, I am going to eliminate the need to search for the delimiter.  It ends up accounting for about half of the execution time for this function at best with no upper bound on time taken.  Not to mention that the search runs twice.

kjarosh: I don't fully understand what you're saying... `WStr::slice` is not linear.

> On another note, I am going to eliminate the need to search for the delimiter.

That's my point.

tests/tests/swfs/avm2/textfield_event/Test.as
R3velation: I've moved them out of the existing tests and into their own units.

PR #12812, opened 2023-08-17
core: Implement links in EditTexts, including AVM1 asfunction: and AVM2 event: links

Dinnerbone:
Fixes #4253
Fixes #7340
Fixes #10496

probably others

PR #23518, opened 2026-04-23
fix(avm1): match asfunction: and event: URL prefixes case-insensitively (#23514)

### LINKED ISSUE #23514: Html HREF buttons tied to actionscript functions are treated as external URLs.

mvanhorn:
Closes #23514.

## Problem

Submachine HD builds use HREFs like \`<A HREF="asFunction:toggle_music">music: on</A>\` for their footer controls. The capital-F \`asFunction:\` fell through Ruffle's byte-exact \`strip_prefix(b"asfunction:")\` check and hit the \`navigate_to_url\` fallback, so clicking "music: on" opened \`asFunction:toggle_music\` in the user's web browser instead of calling the AVM1 function.

Flash Player matched both \`asfunction:\` and \`event:\` case-insensitively.

## Fix

In \`EditText::open_url\`, lowercase the URL head once and strip the prefix against the lowercased variant, then slice the original \`url\` from the back to preserve the address's original casing for downstream consumers (the function name passed to \`execute_avm1_asfunction\` and the event text).

Same treatment for \`event:\` — per the issue author's confirmation in #12812 comments that "the check should probably be made case-insensitive", this is the symmetric fix.

## Verification

- \`cargo check -p ruffle_core\` — clean
- \`cargo clippy -p ruffle_core -- -D warnings\` — clean
- \`cargo fmt --check\` — clean

No regression for the existing canonical-case paths; the only behavioural change is that mixed-case variants (\`asFunction:\`, \`ASFUNCTION:\`, \`Event:\`, etc.) now route through the AVM1 / \`link\` event paths instead of opening a browser.

### DIFF
--- a/core/src/display_object/edit_text.rs
+++ b/core/src/display_object/edit_text.rs
@@ -2468,11 +2468,11 @@ impl<'gc> EditText<'gc> {
     }
 
     fn open_url(self, context: &mut UpdateContext<'gc>, url: &WStr, target: &WStr) {
-        if let Some(address) = url.strip_prefix(WStr::from_units(b"asfunction:")) {
+        if let Some(address) = strip_link_scheme(url, WStr::from_units(b"asfunction:")) {
             if let Err(e) = self.execute_avm1_asfunction(context, address) {
                 error!("Couldn't execute URL \"{url:?}\": {e:?}");
             }
-        } else if let Some(address) = url.strip_prefix(WStr::from_units(b"event:")) {
+        } else if let Some(address) = strip_link_scheme(url, WStr::from_units(b"event:")) {
             if let Some(object) = self.object2() {
                 let mut activation = Avm2Activation::from_nothing(context);
                 let text = AvmString::new(activation.gc(), address);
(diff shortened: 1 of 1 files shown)


kjarosh:
Can you add a test for that? Might be also worth testing whether whitespace is stripped around the function, etc.

Please also adhere to the commit message guidelines.

mvanhorn:
Thanks @kjarosh - pushed 9703d256. Extracted the prefix-matching into a \`strip_link_scheme\` helper and added unit tests in \`#[cfg(test)] mod tests\` at the bottom of edit_text.rs:

- canonical \`asfunction:toggle_music\`
- mixed case \`asFunction:ToggleMusic\` (preserves address casing)
- upper case \`ASFUNCTION:play\`
- mixed case \`Event:clicked\`
- non-matching \`https://\` URL
- empty address after \`asfunction:\`
- substring-only \`asfunctionfoo\` (regression guard)

\`cargo test -p ruffle_core --lib\` passes all 7 new tests alongside the existing 65. \`cargo clippy -p ruffle_core -- -D warnings\` and \`cargo fmt --check\` both clean. Also rewrote the commit message to follow the `core: <sentence> (close #NNN)` format in CONTRIBUTING.md.

For the whitespace-around-function question - that would change observable behavior for existing \`asfunction:\` args that intentionally include leading/trailing whitespace, so I did not add that in this PR. Happy to open a follow-up if you want it to trim.

PR #23519, opened 2026-04-23
web: Add 'lang' attributes to extension pages using i18n ui_locale

danielhjacobs:
Mostly supersedes #23507, aside from the player.html change (since that page is not internationalized (see #14628)).

Uses the predefined @@ui_locale message: https://developer.chrome.com/docs/extensions/reference/api/i18n#overview-predefined.

### DIFF
--- a/web/packages/extension/src/onboard.ts
+++ b/web/packages/extension/src/onboard.ts
@@ -2,6 +2,7 @@ import * as utils from "./utils";
 import { buildInfo } from "ruffle-core";
 
 window.addEventListener("DOMContentLoaded", () => {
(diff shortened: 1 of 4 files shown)


kjarosh (approved):

danielhjacobs:
`@@ui_locale` on Chrome/Safari uses underscores apparently, while on Firefox it uses dashes, so code slightly modified to account for that.

kjarosh (approved):
Would be nice to move this logic to a utility function instead of copy-pasting it for every page.

danielhjacobs:
Done

PR #23527, opened 2026-04-24
avm2: Fix Vector.<Boolean> default value in non-script-init contexts

### LINKED ISSUE #23317: Incorrect default values for Vector.<Boolean> in some contexts
### Describe the bug

1. Compile the following action script program:
```as
package {
    import flash.display.Sprite;
        public function Test() {
            var vec:Vector.<Boolean> = new Vector.<Boolean>(1, true);
            trace(vec[0]); // false on flash, null on ruffle
        }
    }
}

var vec:Vector.<Boolean> = new Vector.<Boolean>(1, true);
trace(vec[0]); // null on both
```

```sh
mxmlc -o test.swf -debug Test.as
```

2. Run with ruffle:
```
INFO avm_trace: null
INFO avm_trace: null
```

### Expected behavior

The traces should match flash, which gives:
```
null
false
```

### Content Location

Precompiled swf: [test.zip](https://github.com/user-attachments/files/26181677/test.zip)

### Affected platform

Desktop app

### Operating system

Arch Linux

### Browser

_No response_

### Additional information

There is already a test for the top-level case: https://github.com/ruffle-rs/ruffle/blob/master/tests/tests/swfs/avm2/vector_holes/Test.as

However, this does not cover the case in the constructor, where the behaviour is different in flash for some reason.

chabeck1:
## Summary

Fixes #23317.

In Flash Player on SWF version >= 14, `new Vector.<Boolean>(n, true)` fills the new slots with `false` when called from inside a class method or constructor, but with `null` when called from a script initializer. Ruffle was returning `null` in both cases. This PR makes the non-script-init case match Flash on SWF >= 14 while leaving SWF <= 13 and the script-init case unchanged.

## Approach

`VectorStorage::default()` for `Boolean` now delegates to a `bool_default()` helper in `core/src/avm2/vector.rs` that walks the AVM2 call stack from the top, finds the first bytecode method, and checks whether it is the instance initializer of a script's `global` class (`Class::is_script_traits()`). If so, the default is `Value::Null` (preserving the existing `vector_holes` behavior). Otherwise it is `Value::Bool(false)` for SWF >= 14 and `Value::Null` below.

To enable the call-stack walk inside `VectorStorage`, `VectorStorage::new` now takes `&mut Activation`; the change is threaded through every call site (`amf.rs`, `object/vector_object.rs`, `globals/vector.rs`, two `globals/flash/*` files). A small `iter_top_down()` helper was added to `CallStack`.

## Acknowledgement of maintainer feedback

@Lord-McSweeney and @kjarosh noted on the issue that the Flash behavior here is a JIT specialization artifact, not a Vector semantic rule, and that any fix without a JIT would likely cost `Vector` construction performance. I understand this PR takes the "encode the JIT behavior as a runtime check" approach, which the issue thread explicitly cautioned against. I'm opening it anyway so the approach is on the record and reviewable; I fully understand if the preferred resolution is to close without merging. If there are changes that would make a variant of this acceptable (e.g. restricting the call-stack walk to SWF >= 14 paths only, or gating behind a config flag), I'm happy to iterate.

## Tests

- `tests/tests/swfs/avm2/vector_boolean_default_constructor/` -- SWF v14, the bug case. Expected output `null` then `false` captured from Adobe debug Flash Player.
- `tests/tests/swfs/avm2/vector_boolean_default_constructor_v13/` -- SWF v13 sanity check, expected output `null` then `null`. Pins the pre-existing behavior below the version threshold.
- Full regression suite: `4016 passed; 0 failed`. Pre-existing `vector_holes` test continues to pass.

## Out of scope

The Haxe-generated static-method case from the issue (where Flash returns `false` regardless of SWF version) is not handled here; the discriminator appears to be Haxe's bytecode pattern rather than the Vector default rule, which is a separate investigation. Noted in the issue thread.


Lord-McSweeney (changes requested):
This fix is incorrect. The `Vector` is not filled as `false` or `null` depending on whether the activation is in interpreter mode or not; rather, the value is _read_ as `false` or `null` depending on whether the activation is in interpreter mode or not. A boolean `Vector` is always filled with `null` when initialized.

Lord-McSweeney:
Was an LLM used to write any of this code?

chabeck1:
Yes, mostly LLM-generated. Sorry for not flagging it upfront. Your review also makes clear I was on the wrong code path (fill vs read). Close or iterate, whichever you prefer.

Lord-McSweeney:
You can close this PR if you'd like, but if you want to implement the correct fix it's fine to do so on this PR

Lord-McSweeney:
I'm closing this PR as it hasn't been touched in a while and I've explained why this is the wrong way to do it. Feel free to reopen it or open a new PR if you can implement the correct fix.

PR #23531, opened 2026-04-24
web: Re-format extension pages when using RTL languages

danielhjacobs:


danielhjacobs:
This screenshot was the result of adding `html:lang(en-US),` to the CSS for rtl direction temporarily, it looks weird because English is not an RTL language but it's an example of the RTL layout.
Screenshot From 2026-04-24 11-17-02

kjarosh (approved):

PR #23534, opened 2026-04-25
web: Improve software renderer detection

robinmiau:
### Problem

`Adapter Device Type: Cpu` alone doesn't catch all software renderers. WebGL/ANGLE often reports the device type as `Other` even when no hardware acceleration is active.

### Changes

Added checks for additional known software renderers alongside the existing check:

- `VirtualGpu` - virtual GPU environments (VMs, cloud)
- `Microsoft Basic Render Driver` - WARP fallback driver
- `SwiftShader` - Google's software renderer (headless Chrome, CI)
- `llvmpipe` / `softpipe` - Mesa software rasterizers on Linux

The hardware acceleration warning now triggers reliably across all common software rendering scenarios.

### DIFF
--- a/web/packages/core/src/internal/player/inner.tsx
+++ b/web/packages/core/src/internal/player/inner.tsx
@@ -753,15 +753,17 @@ export class InnerPlayer {
         // Show a hardware acceleration warning when software rendering is detected.
         // The device type is unreliable through WebGL/ANGLE (always "Other"), so we
         // also match known software renderer names (WARP, SwiftShader, Mesa llvmpipe).
-        const isSoftwareRenderer =
-            this.rendererDebugInfo.includes("Adapter Device Type: Cpu") ||
-            this.rendererDebugInfo.includes(
-                "Adapter Device Type: VirtualGpu",
-            ) ||
-            this.rendererDebugInfo.includes("Microsoft Basic Render Driver") ||
-            this.rendererDebugInfo.includes("SwiftShader") ||
-            this.rendererDebugInfo.includes("llvmpipe") ||
-            this.rendererDebugInfo.includes("softpipe");
+        const softwareRendererIndicators = [
+            "Adapter Device Type: Cpu",
+            "Adapter Device Type: VirtualGpu",
+            "Microsoft Basic Render Driver",
+            "SwiftShader",
+            "llvmpipe",
+            "softpipe",
+        ];
+        const isSoftwareRenderer = softwareRendererIndicators.some(
+            (indicator) => this.rendererDebugInfo.includes(indicator),
+        );
         if (isSoftwareRenderer) {
             this.container.addEventListener(
                 "mouseover",
--- a/web/packages/core/src/internal/player/inner.tsx
+++ b/web/packages/core/src/internal/player/inner.tsx
@@ -750,7 +750,19 @@ export class InnerPlayer {
 
         this.rendererDebugInfo = this.instance!.renderer_debug_info();
 
-        if (this.rendererDebugInfo.includes("Adapter Device Type: Cpu")) {
+        // Show a hardware acceleration warning when software rendering is detected.
+        // The device type is unreliable through WebGL/ANGLE (always "Other"), so we
+        // also match known software renderer names (WARP, SwiftShader, Mesa llvmpipe).
+        const isSoftwareRenderer =
+            this.rendererDebugInfo.includes("Adapter Device Type: Cpu") ||
+            this.rendererDebugInfo.includes(
+                "Adapter Device Type: VirtualGpu",
+            ) ||
+            this.rendererDebugInfo.includes("Microsoft Basic Render Driver") ||
+            this.rendererDebugInfo.includes("SwiftShader") ||
+            this.rendererDebugInfo.includes("llvmpipe") ||
+            this.rendererDebugInfo.includes("softpipe");
+        if (isSoftwareRenderer) {
             this.container.addEventListener(
                 "mouseover",
                 this.openHardwareAccelerationModal.bind(this),


web/packages/core/src/internal/player/inner.tsx
```diff
@@ -750,7 +750,19 @@ export class InnerPlayer {
 
         this.rendererDebugInfo = this.instance!.renderer_debug_info();
 
-        if (this.rendererDebugInfo.includes("Adapter Device Type: Cpu")) {
+        // Show a hardware acceleration warning when software rendering is detected.
+        // The device type is unreliable through WebGL/ANGLE (always "Other"), so we
+        // also match known software renderer names (WARP, SwiftShader, Mesa llvmpipe).
+        const isSoftwareRenderer =
+            this.rendererDebugInfo.includes("Adapter Device Type: Cpu") ||
```
kjarosh: Instead of copy-pasting `this.rendererDebugInfo.includes` you could define a list and iterate over it.

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -750,7 +750,19 @@ export class InnerPlayer {
...
+            this.rendererDebugInfo.includes(
+                "Adapter Device Type: VirtualGpu",
+            ) ||
+            this.rendererDebugInfo.includes("Microsoft Basic Render Driver") ||
+            this.rendererDebugInfo.includes("SwiftShader") ||
+            this.rendererDebugInfo.includes("llvmpipe") ||
+            this.rendererDebugInfo.includes("softpipe");
+        if (isSoftwareRenderer) {
```
kjarosh, line 767: This is a partially preexisting issue, but let's say someone doesn't have a GPU, or they don't want to enable it for some reason. I will have to discard the dialog every single time I open the SWF right?

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -750,7 +750,19 @@ export class InnerPlayer {
...
             this.container.addEventListener(
                 "mouseover",
                 this.openHardwareAccelerationModal.bind(this),
```
kjarosh, line 770: If I see correctly, this modal was created for Chrome specifically. I'm not familiar with this code, but it looks like it was very targeted, and these changes expand its scope a bit.

CC @danielhjacobs

web/packages/core/src/internal/player/inner.tsx
robinmiau: Maybe a "Don't show again (for this website?)" option would be useful?

web/packages/core/src/internal/player/inner.tsx
danielhjacobs: It's only targeted insofar as only Chrome seems to allow easily disabling hardware acceleration in a way that affects Ruffle. Safari does not allow disabling hardware acceleration whatsoever. Firefox allows disabling hardware acceleration, but for that browser, the behavior seems to differ in that disabling it only affects WebRender, not whether canvas rendering is hardware-backed. There may be some setting in about:config that affects whether canvas rendering is hardware-backed, but I've not been able to find it and so it's much more likely if this is showing up in Firefox the user for whom it shows is a power user who will know how to fix that.

That said, curious about why we'd need some of these checks, though not familiar enough to definitely say we don't. If the renderer includes the phrase "swiftshader" or "llvmpipe" that should already be caught by the "Adapter Device Type: Cpu" check:

https://github.com/gfx-rs/wgpu/blob/868eb17098dadee1edaf9b04839262223bdf5a75/wgpu-hal/src/gles/adapter.rs#L144-L156

@n0samu was the one who suggested that a check for just Cpu may be sufficient:

image

@adrian17 did believe we'd get false negatives though:

image

web/packages/core/src/internal/player/inner.tsx
danielhjacobs: I think that could be reasonable, if we were to add it we could probably leverage localStorage for that (with a safety check that it's available, see the save manager check):

https://github.com/ruffle-rs/ruffle/blob/f34576a961ff40bec91fcb953534032d707cb714/web/packages/core/src/internal/player/inner.tsx#L1274-L1280

Maybe clicking "Don't show again (for this website)" can store a 1 with key ruffleDisableHardwareAccelationModal, and if a 1 is stored with that key don't show it.

web/packages/core/src/internal/player/inner.tsx
danielhjacobs: See also https://github.com/Modernizr/Modernizr/wiki/Undetectables#general-undetectables. We're just adding more duck-typing guesses, which may be fine, but overall hardware acceleration remains a "General undetectable"

web/packages/core/src/internal/player/inner.tsx
danielhjacobs: The extension can theoretically use https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/local to allow "Don't show again" to work for all websites, but I don't think the self-hosted build can do that.

web/packages/core/src/internal/player/inner.tsx
n0samu: Yeah, this modal was here to solve a very specific issue at a specific point in time. All the users we spoke to and got debug info from had the Cpu adapter device type listed, hence I figured it was enough to check for just that. I'm not against expanding the feature to account for other common scenarios where users have hardware acceleration disabled, but I don't know how common any of the additional cases covered by this PR actually are.

kjarosh: @robinmiau Did you have any specific cases of software renderers not being detected properly? If you could list those cases it would be great, because we could make the decision on real data instead of our assumptions.

robinmiau: @kjarosh Only the `Microsoft Basic Render Driver` when hardware acceleration was disabled in Opera GX.

n0samu: @robinmiau Was that on real hardware or a VM?

robinmiau: @n0samu real hardware

n0samu: Okay, that one could make sense to add then. We'd need to make sure the instructions we're giving users for fixing the problem also apply to that browser though. (I remember we link to an FAQ section with instructions.)

Issue #21421, opened 2025-08-19
Context menu over one SWF does not close when right-clicking over another


WumboSpasm:
### Describe the bug

If an SWF is right-clicked with the context menu already open over a different SWF on the same page, a second one will appear while the first one remains open. This is inconsistent with the left-click behavior, where left-clicking over one SWF closes the context menu of another.

https://github.com/user-attachments/assets/5dbb2551-27d8-4bdd-b071-3fc45b8d7325

### Expected behavior

Right-clicking should always close any open context menus before creating a new one, regardless of which SWF was right-clicked.

### Content Location

[rightclick_example.zip](https://github.com/user-attachments/files/21874007/rightclick_example.zip)

### Affected platform

Browser's extension

### Operating system

Debian 12

### Browser

Mozilla Firefox 128.13.0esr

n0samu:
We should close the context menu on any pointerdown outside the content. Need to make sure this is implemented in a way that doesn't break anything, including on iOS, though.
Note: This bug also applies to right-clicking anywhere on a webpage outside of the Ruffle player, not just on 2 different Ruffle players. Flash Player closes its context menu when you invoke the browser context menu anywhere else on the page; we should do the same.

kjarosh:
I'm not sure if that's a good idea to implement this based on pointerdown. We'd have to handle a lot of edge cases and I'm pretty sure we'd miss something (similarly to how the focus was handled in Ruffle in the past).

IMO we should just close the context menu when we lose focus and everything should work as expected. That would provide a more natural context menu, but there's also the question of how FP's context menu worked on web in different browsers and versions.

n0samu:
Ah I didn't think of implementing it based on focus. That sounds good!

PR #23535, opened 2026-04-25
web: Improve context menu behavior and styling

robinmiau:
This PR:

- Adds checkmark support for context menu items with proper CSS column alignment.
- Fixes menu positioning to flip above/left of the cursor at viewport edges instead of clamping.
- Opens the native browser context menu when using Shift+right-click.
- Adds rounded corners (and slightly more padding) to the context menu.

**Old:**
image

**New:**
image

### DIFF
--- a/web/packages/core/src/internal/player/inner.tsx
+++ b/web/packages/core/src/internal/player/inner.tsx
@@ -196,8 +196,7 @@ export class InnerPlayer {
     // When set to `true`, the next context menu event will
     // not show the context menu.
     private _suppressContextMenu = false;
-    private hideContextMenuOnWheel: ((event: WheelEvent) => void) | null =
-        null;
+    private hideContextMenuOnWheel: ((event: WheelEvent) => void) | null = null;
 
     // The effective config loaded upon `.load()`.
     public loadedConfig?: URLLoadOptions | DataLoadOptions;
@@ -393,7 +392,7 @@ export class InnerPlayer {
             }
         };
 
-        modalElement.addEventListener("click", hideModal);
+        modalElement.parentNode!.addEventListener("click", hideModal);
         const modalArea = modalElement.querySelector(".modal-area");
         if (modalArea) {
             modalArea.addEventListener("click", (event) =>
@@ -458,10 +457,6 @@ export class InnerPlayer {
         volumeMuteCheckbox.checked = this.volumeSettings.isMuted;
         volumeSlider.disabled = volumeMuteCheckbox.checked;
         volumeSlider.valueAsNumber = this.volumeSettings.volume;
-        volumeSlider.style.setProperty(
-            "--volume-pct",
-            `${this.volumeSettings.volume}%`,
-        );
         volumeSliderText.textContent = volumeSlider.value + "%";
         setVolumeIcon();
 
@@ -474,10 +469,6 @@ export class InnerPlayer {
         });
         volumeSlider.addEventListener("input", () => {
             volumeSliderText.textContent = volumeSlider.value + "%";
-            volumeSlider.style.setProperty(
-                "--volume-pct",
-                `${volumeSlider.value}%`,
-            );
             this.volumeSettings.volume = volumeSlider.valueAsNumber;
             this.instance?.set_volume(this.volumeSettings.get_volume());
             setVolumeIcon();
@@ -769,17 +760,7 @@ export class InnerPlayer {
 
         this.rendererDebugInfo = this.instance!.renderer_debug_info();
 
-        // Show a hardware acceleration warning when software rendering is detected.
-        // The device type is unreliable through WebGL/ANGLE (always "Other"), so we
-        // also match known software renderer names (WARP, SwiftShader, Mesa llvmpipe).
-        const isSoftwareRenderer =
-            this.rendererDebugInfo.includes("Adapter Device Type: Cpu") ||
-            this.rendererDebugInfo.includes("Adapter Device Type: VirtualGpu") ||
-            this.rendererDebugInfo.includes("Microsoft Basic Render Driver") ||
-            this.rendererDebugInfo.includes("SwiftShader") ||
-            this.rendererDebugInfo.includes("llvmpipe") ||
-            this.rendererDebugInfo.includes("softpipe");
-        if (isSoftwareRenderer) {
+        if (this.rendererDebugInfo.includes("Adapter Device Type: Cpu")) {
             this.container.addEventListener(
                 "mouseover",
                 this.openHardwareAccelerationModal.bind(this),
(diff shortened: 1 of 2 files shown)


web/packages/core/src/internal/player/inner.tsx
```diff
@@ -1669,6 +1677,22 @@ export class InnerPlayer {
             return;
         }
 
+        // If the custom menu is already open and the user right-clicks again
+        // at roughly the same position, hide our menu and let the browser show
+        // its native context menu. A 8px threshold accounts for slight mouse drift.
```
kjarosh: Is there a reason why we'd want to support that behavior? You can open the native menu already on Shift+RMB. I feel like it would be more confusing.

robinmiau: I think that would be a good way to quickly copy or save the canvas as an image, for example. YouTube's player also allows you to right-click twice to open the browser's context menu.

Shift+RMB doesn't seem to work for me (Windows, Chrome)

kjarosh: I see, looks like Shift+RMB works only on Firefox, I thought other browsers do that too.

In this case I think it's fine to skip showing the context menu if Shift is down. I'm strongly against adding the logic of double clicking within a 8px circle, because:
1. the native menu is usually not very usable (e.g. compared to YouTube), it's not what the user wants to see when right clicking on a SWF,
2. the context menu is critical for Ruffle (that's the only GUI we have),
3. the behavior would differ compared to Flash Player, where opening a context menu is observable.

kjarosh:
I'm a big fan of the design BTW, it's also nice that now the menu folds properly when there's not enough space—that's something we should test in `web/packages/selfhosted/test/integration_tests/context_menu_position`, but I can add that case in a follow-up PR.

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -187,6 +196,7 @@ export class InnerPlayer {
     // When set to `true`, the next context menu event will
     // not show the context menu.
     private _suppressContextMenu = false;
+    private hideContextMenuOnWheel: ((event: WheelEvent) => void) | null = null;
```
kjarosh: Why do we want to hide the menu on mouse wheel? That's not how it worked in Flash Player and that's not how it works in general, the context menu should prevent mouse wheel when displayed, which is what happens now.

robinmiau: It seems to be more of a browser-specific behavior; in Flash Player, the context menu doesn't disappear when you scroll, but it does in the browser. Since we're trying to replicate the browser's behavior from Flash here, I think it would make sense to have. What do you think?

web/packages/core/src/internal/player/inner.tsx
robinmiau: Sounds fair, I'll change it. What about making a right-click anywhere else close Ruffle’s context menu? I think that would fix #21421. Should I include that as well?

kjarosh: I think the proper solution to https://github.com/ruffle-rs/ruffle/issues/21421 is to close the menu on focus loss, as that's how it works in Flash Player. It would also close the context menu when you lose focus on the window, which is how it's supposed to work—only one context menu shown in the whole system.

The decision whether to close the menu on right click is a tough one, back in the day on Windows right clicking would open up the context menu again, but we would have to see how it worked in different browsers and operating systems. The main issue here is how the SWF observes it—ideally we don't want to break anything.

I would suggest implementing the former. Nonetheless, let's keep that for a future PR.

web/packages/core/src/internal/player/inner.tsx
kjarosh: I've done my homework on this (because that behavior looked non-standard and I don't think I've ever encountered it): I checked how the context menu behaves in many places (GTK, GNOME, KDE, Windows, Firefox on GNOME, Firefox on KDE, Chromium on GNOME, MacOS, Chrome on MacOS, Edge on Windows, Pale Moon on Windows, Flash Player in Pale Moon on Windows). The only place where it disappeared was Edge on Windows, and everywhere else scrolling does not make the menu disappear, which I think is fair and would mean that it's the standard behavior, which we should replicate.

Whether the event is passed to the content or not is a different question, that differed across different environments, but it did in Flash Player actually (I originally thought it didn't because I forgot the wheel behavior is generally weird in Flash).

Nonetheless, I wouldn't change the behavior here.

robinmiau: In Chromium 87, it disappears when you scroll. I thought that was the default behavior. I've removed it!

web/packages/core/src/internal/ui/static-styles.css
```diff
@@ -204,19 +204,38 @@
 #context-menu {
     color: rgb(var(--modal-foreground-rgb));
     background-color: var(--modal-background);
-    border: 1px solid gray;
-    box-shadow: 0 5px 10px -5px black;
+    box-shadow: 0 4px 16px #0006;
```
kjarosh: Can we make it non-translated? (i.e. `0` instead of `4px`) I don't think the difference is that big in general, but this will look nicer for white backgrounds. Currently it looks off.

web/packages/core/src/internal/ui/static-styles.css
```diff
@@ -204,19 +204,38 @@
...
+    border-radius: 8px;
```
kjarosh, line 208: Can we define a variable for this value of radius (e.g. `--ruffle-border-radius`)? When we're doing rounded corners it's best to keep the radius the same throughout all components.

web/packages/core/src/internal/ui/static-styles.css
```diff
@@ -204,19 +204,38 @@
...
-        color: rgb(var(--modal-foreground-rgb));
+        padding: 7px 12px;
+    }
+
+    &.has-checkmarks .menu-item {
+        padding-left: 32px;
+        position: relative;
+    }
+
+    &.has-checkmarks .menu-item.checked::before {
```
kjarosh, line 227: The checkmark doesn't work for RTL languages, it should be on the right.

Image

web/packages/core/src/internal/ui/static-styles.css
robinmiau: Only the context menu uses `8px` border radius; unless we change the radius for other elements to the same value, I don't think a CSS variable would make much sense yet, would it?

Or should I adjust some? `.modal-button` uses `6px`, changing it to `8px` shouldn't make much of a difference.

kjarosh: Hmmm, I think we should be using the same radius for the same category of elements, but that's something we can do in a follow-up to unify it after the fact.

web/packages/core/src/internal/ui/static-styles.css
robinmiau: I switched to CSS logical properties. I haven't been able to test it, can you check if it's correct now?

kjarosh: Appears to be working, thank you! 

Image

BTW You can test it by setting `dir` to `rtl` on `#context-menu`.

web/packages/core/src/internal/player/inner.tsx
```diff
@@ -93,6 +93,12 @@ interface ContextMenuItem {
      * @default true
      */
     enabled?: boolean;
+
+    /**
+     * Whether this item has a checkmark next to it.
+     * When defined, a checkmark column is shown for all items in the group.
```
kjarosh: Nit:

```suggestion
     * When defined, a checkmark column is shown for all items in the context menu.
```

kjarosh (approved):
Code and looks LGTM, thank you!

I'm gonna leave the PR unmerged for a while so that others can voice their opinions too about the look.

torokati44:
At least the new screenshot in the OP looks good to me visually! :+1: Thanks!

PR #23536, opened 2026-04-25
core: Invalidate Bitmap cache when bitmapData changes

FinBird:
## Summary
Invalidates cached `Bitmap` renderings when `bitmapData` is updated. 

Previously, filtered Bitmaps or those using `cacheAsBitmap` would reuse the old texture if the new `BitmapData` shared the same dimensions, resulting in stale visuals (e.g., loading screens or transitions persisting incorrectly).

## Changes
Updated `Bitmap::set_bitmap_data` to force a cache invalidation after the new `BitmapData` is assigned. This ensures the display object correctly refreshes its representation. This is a general fix and applies to all SWF content.

*Note: Sorry, No SWF regression test was added as I currently lack the environment to regenerate assets via `mxmlc`.*


Dinnerbone:
> _Note: Sorry, No SWF regression test was added as I currently lack the environment to regenerate assets via `mxmlc`._

Could you point us to a real world swf, or describe roughly how to reproduce this? We can help you with making tests!

jarca0123:
Hopefully this reproduces the issue: [bitmap_data_replaced.zip](https://github.com/user-attachments/files/27088244/bitmap_data_replaced.zip)

I tested it on FP32 debug.

FinBird:
Anyone can review and merge this pull?

kjarosh:
This will need a test to get merged.

kjarosh (approved):
Added a test, so we're good to merge. Thanks!

PR #23538, opened 2026-04-25
web: Polish modal, save manager, and volume slider UI

robinmiau:
This PR:

- Switches modals to flexbox centering, removing the translate-based centering.
- Improves save manager table rows with hover highlight, custom scrollbar, and better padding. 
- Styles the volume slider as a filled range track using a CSS custom property updated on input.

### Volume slider

**Old:**
image

**New:**
image

### Save manager

**Old:**
Screenshot

**New:**
Screenshot


Dinnerbone:
I really like this, thanks so much!

Would you know why it fails our tests though? :D

(also please prefix your commits with e.g. `web:`)

danielhjacobs:
Guessing it's the removal of parentNode from modalElement.parentNode!.addEventListener("click", hideModal), I remember adding that for a reason, can't quite remember what reason but guessing it was caught by tests

robinmiau:
@danielhjacobs That seemed like the most logical reason to me, but the tests are still failing. Since these are all context menu tests, I thought updating the base branch would help, but unfortunately it didn't.

PR #23547, opened 2026-04-27
refactor: Simplify some match patterns

kjarosh:
This patch simplifies some match patterns by removing top-level duplication, the | operator can be used in subexpressions too.

### DIFF
--- a/core/src/avm2/optimizer/type_aware.rs
+++ b/core/src/avm2/optimizer/type_aware.rs
@@ -1531,8 +1531,7 @@ fn abstract_interpret_ops<'gc>(
                 if !multiname.has_lazy_component() {
                     if let Some(vtable) = stack_value.vtable() {
                         match vtable.get_trait(&multiname) {
-                            Some(Property::Slot { slot_id })
-                            | Some(Property::ConstSlot { slot_id }) => {
+                            Some(Property::Slot { slot_id } | Property::ConstSlot { slot_id }) => {
                                 // If the set value's type is the same as the type of the slot,
                                 // a SetSlotNoCoerce can be emitted. Otherwise, emit a SetSlot.
                                 let mut value_class =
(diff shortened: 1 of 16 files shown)


core/src/avm2/optimizer/type_aware.rs
```diff
@@ -1707,28 +1706,25 @@ fn abstract_interpret_ops<'gc>(
...
+                    if let Some(vtable) = stack_value.vtable()
+                        && let Some(Property::Slot { slot_id } | Property::ConstSlot { slot_id }) =
+                            vtable.get_trait(&multiname)
+                    {
+                        let mut value_class =
+                            vtable.slot_class(slot_id).expect("Slot should exist");
+                        let resolved_value_class = value_class.get_class(activation)?;
+
+                        if let Some(slot_class) = resolved_value_class {
+                            if let Some(instance_class) = slot_class.i_class() {
```
SuchAFuriousDeath: Looks like this could be joined with the upper if let Some(...)?

core/src/display_object/movie_clip.rs
```diff
@@ -1783,8 +1783,7 @@ impl<'gc> MovieClip<'gc> {
                     prev_child.apply_place_object(context, &params.place_object);
                     prev_child.set_place_frame(params.frame);
                 }
-                (PlaceObjectAction::Place(id), _, _)
-                | (swf::PlaceObjectAction::Replace(id), _, _) => {
+                (PlaceObjectAction::Place(id) | swf::PlaceObjectAction::Replace(id), _, _) => {
```
SuchAFuriousDeath: This should probably be unified to only use PlaceObjectAction or swf::PlaceObjectAction, above match branches too

SuchAFuriousDeath (changes requested):

core/src/avm2/optimizer/type_aware.rs
kjarosh: Done

core/src/display_object/movie_clip.rs
kjarosh: Done

SuchAFuriousDeath (approved):

PR #23550, opened 2026-04-27
tests: Fix pixelbender test flakiness

SuchAFuriousDeath:
Disclaimer: This is like 95% Claude Code. I reviewed it and the logic and evidence seems compelling and it seems to actually fix the flakiness for me locally (it is actually reproducible with lavapipe, which is what CI uses)
 
  ## Summary

  PixelBender (and other wgpu) tests have been flaky on CI for some time —
  they print `test ... ok` and then the process SEGVs during teardown.
  `.config/nextest.toml` papered over it with `retries = 4` and a TODO
  ("probably some unsoundness in naga, lavapipe?"). This PR fixes the
  underlying race so the workaround can drop `pixelbender` from its filter.

  ## Root cause

  The test framework caches wgpu `Descriptors` in a `static OnceLock<…>`
  (see `tests/tests/environment.rs`). Rust statics aren't dropped at
  process exit, so `vkDestroyDevice` / `vkDestroyInstance` never run.
  lavapipe creates LLVM-JIT worker threads inside `vkCreateDevice` and
  relies on `vkDestroyDevice` to join them.

  `libtest_mimic::Conclusion::exit` calls `std::process::exit`, which
  calls libc `exit` → `__cxa_finalize` → every loaded `.so`'s static
  destructors. Those destructors free driver state while the lavapipe
  worker threads are still alive → SEGV.

  GDB confirms it. After `test … ok`:

  - Thread 1: `munmap` ← driver cleanup ← `__cxa_finalize` ← `exit` ←
    `libtest_mimic::Conclusion::exit` ← `FsTestsRunner::run`
  - Thread 2: still inside `llvm::MCJIT::finalizeObject` JIT-compiling a
    pixelbender shader
  - Threads 3–14: `llvmpipe-N` workers parked in `pthread_cond_wait`
    inside `libvulkan_lvp.so`

  This is why only PixelBender tests tripped it reliably: their shaders
  are generated at runtime and lavapipe is still busy JIT-compiling
  SPIR-V → NIR → LLVM IR → native when the test exits. The static-WGSL
  visual tests finish their compile fast enough to win the race.

  Related upstream issues for the same family of teardown bug:
  [wgpu#7309](https://github.com/gfx-rs/wgpu/issues/7309),
  [wgpu#5781](https://github.com/gfx-rs/wgpu/issues/5781),
  [wgpu#8365](https://github.com/gfx-rs/wgpu/issues/8365).

  ## Fix

  Drop the cached `Arc<Descriptors>` *before* `process::exit`, so wgpu's
  normal `Drop` chain runs `vkDestroyDevice` (which joins the lavapipe
  workers) before `__cxa_finalize` touches driver globals.

  - `tests/fs-tests-runner`: add `with_before_exit(Box<dyn FnOnce()>)`
    hook, invoked after `libtest_mimic::run` returns and before
    `Conclusion::exit`.
  - `tests/tests/environment.rs`: switch `WGPU` from
    `OnceLock<Option<Arc<Descriptors>>>` to
    `LazyLock<Mutex<Option<Arc<Descriptors>>>>` and expose
    `shutdown_wgpu()` that empties it.
  - `tests/tests/regression_tests.rs`: wire
    `runner.with_before_exit(Box::new(environment::shutdown_wgpu))`
    under the `imgtests` feature.
  - `.config/nextest.toml`: drop `test(pixelbender)` from the flaky
    filter and the "naga, lavapipe?" comment.

  No production-crate changes.

  ## Reproduction & verification

  Reproduced 100% locally on Linux with Mesa lavapipe (the same
  `llvmpipe (LLVM 22.x)` driver CI uses), via
  `VK_DRIVER_FILES=/path/to/lvp_icd.json` against the test binary.

  | Configuration                                                | Before | After |
  | ---                                                          | ---    | ---   |
  | `pixelbender_sign` × 10 runs (lavapipe, single-test process) | 10/10 SEGV | 0/10 |
  | 8 PixelBender tests × 5 runs each                            | 40/40 SEGV | 0/40 |
  | All 31 PixelBender tests in one process                      | crash      | 31 passed, exit 0 |
  | Same on native NVIDIA / RADV Vulkan                          | (didn't crash) | 0/3 |
  | `cargo clippy --workspace --no-deps`                         | —          | clean |

### DIFF
--- a/exporter/integration_tests/src/runner.rs
+++ b/exporter/integration_tests/src/runner.rs
@@ -90,7 +90,7 @@ fn main() -> Result<()> {
             register_trial(load_test(params))
         }));
 
-    runner.run()
+    runner.run().exit()
 }
 
 fn load_test(params: TestLoaderParams) -> Trial {
--- a/tests/fs-tests-runner/src/lib.rs
+++ b/tests/fs-tests-runner/src/lib.rs
@@ -6,6 +6,7 @@
 //! libtest_mimic is used so that the runner is compatible with `cargo test`.
 
 use anyhow::Context;
+pub use libtest_mimic::Conclusion;
 use libtest_mimic::{Arguments, Trial};
 use regex::Regex;
 use std::borrow::Cow;
@@ -112,7 +113,7 @@ impl FsTestsRunner {
         self
     }
 
-    pub fn run(mut self) -> ! {
+    pub fn run(mut self) -> Conclusion {
         self.ensure_root_dir_exists();
 
         let args = Arguments::from_args();
@@ -157,7 +158,7 @@ impl FsTestsRunner {
 
         tests.sort_unstable_by(|a, b| a.name().cmp(b.name()));
 
-        libtest_mimic::run(&args, tests).exit()
+        libtest_mimic::run(&args, tests)
     }
 
     fn ensure_root_dir_exists(&self) {
--- a/tests/tests/environment.rs
+++ b/tests/tests/environment.rs
@@ -1,17 +1,60 @@
 use ruffle_test_framework::environment::{CompileMode, Environment};
 
-#[derive(Debug)]
+#[cfg(feature = "imgtests")]
+use ruffle_render_wgpu::descriptors::Descriptors;
+#[cfg(feature = "imgtests")]
+use std::sync::{Arc, LazyLock};
+
 pub struct NativeEnvironment {
     pub compile_mode: CompileMode,
+    #[cfg(feature = "imgtests")]
+    descriptors: LazyLock<Option<Arc<Descriptors>>>,
+}
+
+impl NativeEnvironment {
+    pub fn new(compile_mode: CompileMode) -> Self {
+        Self {
+            compile_mode,
+            #[cfg(feature = "imgtests")]
+            descriptors: LazyLock::new(renderer::build_wgpu_descriptors),
+        }
+    }
+
+    /// Wait for any pending GPU work to finish, with a `timeout`.
+    ///
+    /// Call this before `main` ends so the driver's worker threads quiesce
+    /// while the environment is still alive — otherwise libc's
+    /// `__cxa_finalize` can run while a Mesa lavapipe JIT thread is still
+    /// compiling a shader, racing the destructors and producing a
+    /// `SIGSEGV`. The wait also keeps wgpu's `Queue::Drop` (with its
+    /// hard-coded ~6.3 s budget that `panic!`s on timeout) happy on slow
+    /// software renderers like DX12-WARP on the Windows GitHub-Actions
+    /// runner.
+    pub fn flush_gpu_with_timeout(&self, #[allow(unused)] timeout: std::time::Duration) {
+        #[cfg(feature = "imgtests")]
+        {
+            // `LazyLock::get` peeks without forcing initialization, so tests
+            // that never touched the renderer don't pay the cost of building
+            // wgpu state just to flush it again at shutdown.
+            if let Some(Some(descriptors)) = LazyLock::get(&self.descriptors) {
+                use ruffle_render_wgpu::wgpu;
+
+                let _ = descriptors.device.poll(wgpu::PollType::Wait {
+                    submission_index: None,
+                    timeout: Some(timeout),
+                });
+            }
+        }
+    }
 }
 
 impl Environment for NativeEnvironment {
     #[cfg(feature = "imgtests")]
     fn is_render_supported(
         &self,
-        requirements: &ruffle_test_framework::options::RenderOptions,
+        _requirements: &ruffle_test_framework::options::RenderOptions,
     ) -> bool {
-        renderer::is_supported(requirements)
+        self.descriptors.is_some()
     }
 
     #[cfg(feature = "imgtests")]
@@ -23,7 +66,7 @@ impl Environment for NativeEnvironment {
         Box<dyn ruffle_test_framework::environment::RenderInterface>,
         Box<dyn ruffle_test_framework::environment::RenderBackend>,
     )> {
-        renderer::NativeRenderInterface::create_pair(width, height)
+        renderer::create_pair(self, width, height)
     }
 
     fn compile_mode(&self) -> CompileMode {
@@ -33,6 +76,7 @@ impl Environment for NativeEnvironment {
 
 #[cfg(feature = "imgtests")]
 mod renderer {
+    use super::NativeEnvironment;
     use image::RgbaImage;
     use ruffle_render_wgpu::backend::{
         WgpuRenderBackend, create_wgpu_instance, request_adapter_and_device,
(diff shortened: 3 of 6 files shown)


tests/tests/environment.rs
```diff
@@ -82,7 +85,20 @@ mod renderer {
...
+    /// Drop the cached wgpu descriptors so the underlying GPU device, adapter,
+    /// and instance are destroyed before the process exits.
+    ///
+    /// This avoids a teardown race with driver worker threads (notably Mesa
+    /// lavapipe's LLVM JIT threads) when `__cxa_finalize` runs static
+    /// destructors of loaded driver shared libraries while those threads are
+    /// still alive — the workers outlive the global state they reference,
+    /// producing a `SIGSEGV` after the test has already passed.
+    pub fn shutdown_wgpu() {
+        let _ = WGPU.lock().map(|mut guard| guard.take());
```
kjarosh: I think this is a bad approach. If dropping is the issue, just refactor the code so that no static state is used. Just initialize and pass the state (descriptors) to the environment and keep it as a variable in `main`. It will get dropped automatically.

SuchAFuriousDeath: This should be possible, on it :saluting_face:

tests/tests/regression_tests.rs
```diff
@@ -57,36 +59,68 @@ fn main() {
...
-    runner.with_additional_test(Trial::test("external_interface_avm1", || {
-        external_interface_avm1(&NativeEnvironment)
-    }));
-    runner.with_additional_test(Trial::test("external_interface_avm2", || {
-        external_interface_avm2(&NativeEnvironment)
-    }));
-
-    runner.run()
+    {
+        let env = env.clone();
```
kjarosh: I don't have a strong opinion on this, but if you use a different name (e.g. `env_clone`) you can drop the additional blocks and indentation.

exporter/integration_tests/src/runner.rs
```diff
@@ -74,10 +74,21 @@ impl TestOptions {
     }
 }
 
-fn main() -> Result<()> {
+fn main() -> ExitCode {
```
kjarosh: Why are we doing this change?

tests/fs-tests-runner/src/lib.rs
```diff
@@ -112,7 +113,13 @@ impl FsTestsRunner {
         self
     }
 
-    pub fn run(mut self) -> ! {
+    /// Returns an exit code suitable for returning from `main`.
+    ///
+    /// Returning the exit code (rather than calling [`std::process::exit`])
+    /// lets the runtime drop the caller's locals before libc shuts the
+    /// process down — important for callers that own GPU/driver state which
+    /// must run its `Drop` impl before driver shared-library destructors fire.
```
kjarosh: Drop the doc, it's overly specific and focuses on this PR. It's fine to include it in the commit message, not in the doc.

exporter/integration_tests/src/runner.rs
SuchAFuriousDeath: We can't call libc exit (through the exit method on 'Conclusion'), because that is what causes the whole conundrum. So instead, the exit code has to be propagated.

kjarosh: Am I tripping or are you hallucinating? 🤔 That's exporter tests and the only thing you're doing is unwrapping the `Result` and emulating how it would behave when returned by `main`.

kjarosh: What about that?

SuchAFuriousDeath: Right. This is kind of sneaky, because it's not in the diff. The last line of this main function is runner.run() and that newly returns ExitCode instead of Result<()>. So it has to be changed up here to accomodate that.
What I said was slightly misleading because I was hallucinating a little bit.

kjarosh: Ah, that makes sense, sorry! I'm not a big fan of emulating how `main` behaves... I'll try to come up with something

kjarosh: I think we can just exit manually at the end, that should be simpler.

kjarosh: > that newly returns ExitCode instead of Result<()>

Instead of `!` I think

SuchAFuriousDeath: I'll take a look tomorrow.

tests/tests/environment.rs
```diff
@@ -1,14 +1,71 @@
...
+            // `LazyLock::get` peeks without forcing initialisation, so tests
+            // that never touched the renderer don't pay the cost of building
+            // wgpu state just to flush it again at shutdown.
+            if let Some(Some(descriptors)) = LazyLock::get(&self.descriptors) {
+                use ruffle_render_wgpu::wgpu;
+                return descriptors
+                    .device
+                    .poll(wgpu::PollType::Wait {
+                        submission_index: None,
+                        timeout: Some(_timeout),
```
kjarosh: If the timeout is used conditionally just add `#[allow]` instead of prefixing it with an underscore.

tests/tests/environment.rs
```diff
@@ -1,14 +1,71 @@
...
+        Self {
+            #[cfg(feature = "imgtests")]
+            descriptors: LazyLock::new(renderer::build_wgpu_descriptors),
+        }
+    }
+
+    /// Wait for any pending GPU work to finish, with a `timeout`.
+    ///
+    /// Returns `true` if the wait succeeded, the queue was already idle, or
+    /// the descriptors were never initialised (no GPU work to flush). Returns
```
kjarosh: Nit: `initialized`, we're using en-US in code.

tests/tests/environment.rs
```diff
@@ -1,14 +1,71 @@
...
+    /// so `catch_unwind` cannot rescue us — we have to avoid running the
+    /// `Drop` chain at all in that case.
+    ///
+    /// This matters on the GitHub-Actions Windows runners, which have no
+    /// real Vulkan/D3D12 driver and fall back to DX12-WARP whose fence wait
+    /// can outlast wgpu's hard-coded timeout.
+    pub fn flush_gpu_with_timeout(&self, _timeout: std::time::Duration) -> bool {
+        #[cfg(feature = "imgtests")]
+        {
+            // `LazyLock::get` peeks without forcing initialisation, so tests
```
kjarosh: `initialization`

tests/tests/environment.rs
```diff
@@ -108,15 +170,13 @@ mod renderer {
...
+    pub(super) fn build_wgpu_descriptors() -> Option<Arc<Descriptors>> {
+        let (instance, adapter, device, queue) = create_wgpu_device()?;
+
+        Some(Arc::new(Descriptors::new(instance, adapter, device, queue)))
     }
 
-    fn descriptors() -> Option<&'static Arc<Descriptors>> {
-        WGPU.get_or_init(build_wgpu_descriptors).as_ref()
+    fn descriptors(env: &NativeEnvironment) -> Option<Arc<Descriptors>> {
+        env.descriptors.clone()
```
kjarosh: Looks like this can be inlined?

tests/tests/environment.rs
```diff
@@ -78,12 +134,10 @@ mod renderer {
         }
     }
 
-    pub fn is_supported(_requirements: &RenderOptions) -> bool {
-        descriptors().is_some()
+    pub fn is_supported(env: &NativeEnvironment, _requirements: &RenderOptions) -> bool {
+        descriptors(env).is_some()
```
kjarosh: Looks like this can be inlined?

tests/tests/regression_tests.rs
```diff
@@ -57,36 +58,70 @@ fn main() {
             .chain(ruffle_test_opts.split(" ").filter(|s| !s.is_empty())),
     );
 
+    // The environment owns the wgpu state. Holding it in `main` (instead of
+    // a `static`) means it gets dropped as `main` returns, so wgpu's `Drop`
+    // chain (which joins driver worker threads) runs before libc's
+    // `__cxa_finalize` tears the driver shared libraries down.
```
kjarosh: That's fine as a commit message, but looks out of place as a comment

tests/tests/regression_tests.rs
```diff
@@ -57,36 +58,70 @@ fn main() {
...
+
+    let env_clone = env.clone();
+    runner.with_additional_test(Trial::test("external_interface_avm2", move || {
+        external_interface_avm2(&*env_clone)
     }));
 
-    runner.run().exit()
+    let conclusion = runner.run();
+
+    // Try to flush any pending GPU work before letting wgpu's `Drop` chain
```
kjarosh: Can you preface this with a context, i.e. that it's a workaround for a bug... somewhere

tests/tests/environment.rs
```diff
@@ -1,14 +1,65 @@
...
+    ///
+    /// Call this before `main` ends so the driver's worker threads quiesce
+    /// while the environment is still alive — otherwise libc's
+    /// `__cxa_finalize` can run while a Mesa lavapipe JIT thread is still
+    /// compiling a shader, racing the destructors and producing a
+    /// `SIGSEGV`. The wait also keeps wgpu's `Queue::Drop` (with its
+    /// hard-coded ~6.3 s budget that `panic!`s on timeout) happy on slow
+    /// software renderers like DX12-WARP on the Windows GitHub-Actions
+    /// runner.
+    #[allow(unused_variables)]
```
kjarosh: You can use `#[allow(unused)]` on a parameter instead, the current code will allow unused variables in the function too I think

kjarosh:
Code LGTM, can you rebase and fix up conflicts? The onyl conflict should be an additional field in NativeEnvironment.

kjarosh (approved):

PR #23559, opened 2026-04-28
chore: Bump wasm-bindgen to 0.2.120, ...-futures to 0.4.70, js-sys/web-sys to 0.3.97

torokati44:
See:
 - https://github.com/wasm-bindgen/wasm-bindgen/blob/main/CHANGELOG.md#02119 and down to 0.2.108
 - https://github.com/wasm-bindgen/wasm-bindgen/pull/5131

### DIFF
--- a/web/src/ui/font_renderer.rs
+++ b/web/src/ui/font_renderer.rs
@@ -85,9 +85,7 @@ impl CanvasFontRenderer {
         Ok(self.ctx.measure_text(text)?.width())
     }
 
-    fn ensure_canvas_large_enough(&self, width: f64, height: f64) {
-        let width = width.ceil() as u32;
-        let height = height.ceil() as u32;
+    fn ensure_canvas_large_enough(&self, width: u32, height: u32) {
         if self.canvas.width() < width || self.canvas.height() < height {
             self.canvas.set_width(width);
             self.canvas.set_height(height);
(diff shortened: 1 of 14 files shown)


web/src/ui/font_renderer.rs
torokati44, line 1: cc: @kjarosh (I hope I'm not messing anything up here)
See: https://github.com/wasm-bindgen/wasm-bindgen/pull/1920 (and we do enable `web_sys_unstable_apis` for other reasons, hence the API change)

danielhjacobs (approved):
LGTM, requested review from kjarosh to confirm the font renderer changes though.

web/src/ui/font_renderer.rs
```diff
@@ -105,11 +103,12 @@ impl CanvasFontRenderer {
         let height = self.ascent + self.descent;
 
         let bitmap_width = metrics.actual_bounding_box_left() + metrics.actual_bounding_box_right();
-        let bitmap_width = bitmap_width.max(1.0); // TODO Support empty bitmaps.
+        let bitmap_width = bitmap_width.max(1.0).ceil() as u32; // TODO Support empty bitmaps.
```
kjarosh: If `i32` is the largest size we support, just use `i32` here and make sure it fits (panic otherwise? saturate?) and just do `as u32` below.

torokati44: Like this?

torokati44: I think if any text ends up bigger than i32::MAX pixels, we'd have other stuff breaking sooner than the typecast here. (Probably browsers won't let you even operate on canvases/images that large?)

danielhjacobs: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/canvas#maximum_canvas_size

kjarosh (approved):

PR #23565, opened 2026-04-28
web: Replace 'var' with let and const

tygyh:
ES2015 migration for increased memory safety


Lord-McSweeney:
`swfobject.js` isn't our code; it's from its own repository. We specifically ignore linting it in `eslint.config.js`. There's no need to modify the file.

PR #23574, opened 2026-04-28
core: Fix LoaderDisplay's mouse picking for AVM1 content

### LINKED ISSUE #15525: [Bug]Games that previously worked will not be able to progress in the new version.
### Describe the bug

The swf file that works fine with Adobe Flash Player and the 2023/3/8 rffle nightly does not seem to work with the latest rffle nightly.
This behavior is the same as site locking, but this file is site unlocked.

### Expected behavior

After clicking on the screen, the game should start.

### Content Location

[games.zip](https://github.com/ruffle-rs/ruffle/files/14557528/games.zip)

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

_No response_

### Additional information

_No response_

### LINKED ISSUE #21729: Issue with mouse focus change event
### Describe the bug

(possibly related to #21481???)

The AVM2 movie (main.swf) will load an AVM1 movie (avm1.swf) and add the loader as a child. There is a listener for `FocusEvent.MOUSE_FOCUS_CHANGE`, which traces the `relatedObject`. In Ruffle, this does not fire when the loader is clicked.

main.swf has a green rectangle MC behind the AVM1 movie (which overlaps it), and clicking where there is overlap just triggers `MouseEvent.MOUSE_DOWN` for the green rectangle.

### Expected behavior

When the AVM1 movie is clicked, it should trace:
`onMouseFocusChange fired! relatedObject is [object Loader]`
(NOTE: the blue rectangle is an Avm1Button. It should NOT trace the above when it is clicked.)
Also, where the loader overlaps with the green rectangle, it should not trigger the mouse down event for the green rectangle when it is clicked.

### Content Location

[focus_change_wont_fire_for_loader.zip](https://github.com/user-attachments/files/22454601/focus_change_wont_fire_for_loader.zip)

### Affected platform

Desktop app

### Operating system

macOS 15.6.1

### Browser

_No response_

### Additional information

This is probably an unrelated bug, but if you re-export `avm1.fla` as an AS3 swf (or set the "actionscript3" flag to true on the swf in FFDEC), then in main.swf, the loader will have the hand cursor everywhere it is being hovered over. This does not happen in Flash Player.

ChrisCPI:
# Description
- Fixes #15525
- Fixes #21729 

There are two main changes here:
- Similarly to #23468, when `mouse_pick_avm1` returns `None`, instead of counting it as a `Miss`, continue with the `hit_test_shape` check, allowing the hit to still propagate to the Loader.
- When `mouse_pick_avm1` *does* return `Some`, then combine the result with the Loader. Prior to this it did not take the `mouseChildren` property on the Loader into account.

# Testing

Added the `avm2/mouse_pick_loader_avm1` test to click in various places on/near a LoaderDisplay that has loaded an AVM1 movie.

# Checklist


core/src/display_object/loader_display.rs
```diff
@@ -188,12 +188,11 @@ impl<'gc> TInteractiveObject<'gc> for LoaderDisplay<'gc> {
                     } else {
                         let avm1_result = int.mouse_pick_avm1(context, point, require_button_mode);
                         if let Some(result) = avm1_result {
```
kjarosh: I think you can inline `avm1_result` and merge with `else` above, it will simplify the expression.

core/src/display_object/loader_display.rs
```diff
@@ -188,12 +188,11 @@ impl<'gc> TInteractiveObject<'gc> for LoaderDisplay<'gc> {
...
-                            return Avm2MousePick::Hit(result);
-                        } else {
-                            return Avm2MousePick::Miss;
+                            return Avm2MousePick::Hit(result).combine_with_parent(self.into());
                         }
                     }
-                } else if child.hit_test_shape(context, point, options) {
+                }
+                if child.hit_test_shape(context, point, options) {
```
kjarosh, line 195: Nit: please add an empty line above to separate those ifs visually.

kjarosh (approved):
Thank you! That's a very nice test.

PR #23583, opened 2026-04-30
core: Fix AVM2 mouse picking excluding loaded AVM1 movies

### LINKED ISSUE #21448: Issue when AS2 movie loads an AS2 clip, sandboxed inside AS3
### Describe the bug

In this test that I made, there's a tree of 3 SWFs:
as3_loader.swf, which loads loader.swf (an AS2 movie), which then loads rectangle.swf (also AS2).

loader.swf loads rectangle.swf, and when its done, it sets onPress on the clip holding it. This works fine when loader.swf runs standalone, but in Ruffle, when as3_loader.swf is run, the rectangle is unable to be clicked. It seems to still have some kind of input set on it (it acts kinda like if `useHandCursor` were false), but it doesn't run the onPress function.

### Expected behavior

The rectangle should be able to be clicked, which traces "Clicked" and rotates the rectangle some.

### Content Location

[as3_and_as2_load_bug.zip](https://github.com/user-attachments/files/21947840/as3_and_as2_load_bug.zip)

### Affected platform

Desktop app

### Operating system

macOS 15.5

### Browser

_No response_

### Additional information

I tried having onPress be set on a child on the timeline of rectangle.swf, and it worked as expected in Ruffle. So this seems to only affect when the clip itself has onPress (or onRelease and others) set on it.

This is causing two issues in Club Penguin where:
- Collectable pins in rooms cannot be clicked on
- Igloo furniture cannot be picked up when editing

ChrisCPI:
## Description

Fixes #21448

Previously, when combining an AVM2 hit result with the parent, it would incorrectly propagate a hit on an AVM1 movie in button mode loaded within AVM1, since it classifies as being the root of loaded content.

Checking `loader_info().is_none()` instead of `!is_root()` on the display object will still exclude loader roots as intended, while also preventing this case.

## Testing

Covered in the new `avm2/mouse_pick_avm1_root` test.

## Checklist


core/src/display_object/interactive.rs
```diff
@@ -802,7 +802,9 @@ impl<'gc> Avm2MousePick<'gc> {
                 // by the parent `mouseEnabled` property.
                 // However, the root object of a loader or stage is never a valid target of hit
                 // events (even if moved out of the loader's hierarchy).
```
kjarosh, line 804: Doesn't this PR disagrees with this comment? Is the comment wrong?

kjarosh: Oh wait, maybe I misunderstood the PR, the PR actually fixes the code so that the comment applies?

ChrisCPI: That is the goal, yes. The code was mostly correct before, but this PR fixes the described edge case with AVM1 movies.

kjarosh (approved):
Thank you!

PR #23592, opened 2026-04-30
avm1: Implement ASSetNative and ASSetNativeAccessor

kjarosh:
## Description

Those two functions make it easier to set native methods/properties
and implementing them improves test outputs.

## Testing

Tests are included.

## Checklist


Dinnerbone (approved):
Super cool!

core/src/avm1/globals/asnative.rs
```diff
@@ -73,3 +75,101 @@ fn asnew_method<'gc>(
...
+                        activation.gc(),
+                        Some(prop),
+                        attributes,
+                        Attribute::empty(),
+                    );
+                }
+            }
+            as_set_native_method::AS_SET_NATIVE_ACCESSOR => {
+                if let Some(f) = f.as_object(activation) {
+                    object.add_property(activation.gc(), prop, f, Some(f), Attribute::empty());
```
moulins: That looks wrong: it sets both getter and setter to the same function.

Instead, `ASSetNativeAccessor(o, 200, "foo,bar")` should result in:
```
get(foo): ASnative(200, 0)
set(foo): ASnative(200, 1)
get(bar): ASnative(200, 2)
set(bar): ASnative(200, 3)
```

kjarosh: How do you know that?

moulins: From the AVM1 playerglobals; e.g. `ColorMatrixFilter` uses:
```as2
ASSetNativeAccessor(o,1110,"8matrix",1);
```

which sets up both getter (`ASnative(1110, 1)`) and setter (`ASnative(1110, 2)`) for the `matrix` property.

kjarosh: I meant how do you know that setters and getters have different IDs, but I checked that `alpha` setter on `ConvolutionFilter` is id 18, not 9, so it checks out. I'll update the PR to fix that.

kjarosh:
Added tests that verify the assigned IDs.

moulins (approved):

PR #23595, opened 2026-04-30
avm1: Improve LoadVars.toString()

kjarosh:
## Description

LoadVars.toString() claims it returns a string in
application/x-www-form-urlencoded encoding, but that's not entirely
true. Flash does not use + and it appears to be encoding all
non-alphanumeric characters.

In addition to that, this patch fixes error handling.

## Testing

Tests included.

## Checklist


tests/tests/swfs/avm1/loadvars_tostring/test.toml
```diff
@@ -0,0 +1,2 @@
+num_ticks = 1
+# ignore = true
```
Lord-McSweeney: Is there a reason for this line to be commented out like this?

Lord-McSweeney (approved):

tests/tests/swfs/avm1/loadvars_tostring/test.toml
kjarosh: There isn't! Must have copied it over from my avm1 test template

PR #23597, opened 2026-05-01
Make FsTestRunner less dependent on libtest_mimic

Dinnerbone:
## Description

This is all of the necessary changes to the test framework to create a CLI that can navigate tests reusing the same infrastructure as our test runner - minus the actual CLI tool itself.

## Testing

Well, it passes all the existing tests :D CLI that uses it will be in a separate PR, depends on #23513

## Checklist

### DIFF
--- a/tests/tests/regression_tests.rs
+++ b/tests/tests/regression_tests.rs
@@ -7,7 +7,7 @@ use crate::external_interface::tests::{external_interface_avm1, external_interfa
 use crate::shared_object::{shared_object_avm1, shared_object_avm2, shared_object_self_ref_avm1};
 use anyhow::Context;
 use clap::Parser;
-use libtest_mimic::Trial;
+use libtest_mimic::{Arguments, Trial};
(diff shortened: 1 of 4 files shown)


tests/tests/regression_tests.rs
```diff
@@ -76,7 +76,8 @@ fn main() {
                 let trial = trial_for_test(&env_clone, &ruffle_test_opts, test, params.args.list);
                 register_trial(trial);
             }
-        }));
+        }))
+        .with_sorter(Box::new(|a, b| a.name().cmp(b.name())));
```
kjarosh: If this sorting is so popular (used 3 times), instead of duplicating the code you could implement a `FsTestsRunner<Trial>` with `sorted_by_trial_name()`.

kjarosh (approved):

tests/tests/regression_tests.rs
Dinnerbone: Great idea, added!

PR #23608, opened 2026-05-03
web: Remove deprecated attribute

tygyh:
## Description
This is deprecated and should be removed accoding to the [docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script#deprecated_attributes).

## Checklist


n0samu:
This attribute was added to prevent breakage on non-UTF8-encoded pages. You can look for the PR where we added for more information.

Dinnerbone:
"Deprecated" for web technologies largely means "we don't encourage new websites to use it, but it needs to exist for old websites". Ruffle runs on old websites and new.

PR #23620, opened 2026-05-03
core: Set loader to cancelled if target clip is avm1_removed

### LINKED ISSUE #23402: A loaded AVM1 movie will still run even after target clip has been removed
### Describe the bug

In AVM1, if you load a separate movie into an MC (in any way, like with `loadMovie()` or `MovieClipLoader`), and then remove it before it has finished loading, then the loaded movie (when it does finish loading) will still do stuff like execute scripts or play any audio attached to it.

Interestingly, it even still seems to recognize any children (and itself) as still being part of the root movie, despite not showing up in the debug tools.

### Expected behavior

The loaded movie should not run if the clip has been removed before it has finished loading.

### Content Location

Test case:
[avm1_load_remove_bug.zip](https://github.com/user-attachments/files/26518276/avm1_load_remove_bug.zip)

### Affected platform

Desktop app

### Operating system

macOS Tahoe 26.2

### Browser

_No response_

### Additional information

_No response_

ChrisCPI:
## Description

Fixes #23402

When calling `movie_loader` and `movie_loader_data`, it will check if the target clip has been `avm1_removed()`. If so, then it will set that loader's status to `Cancelled` and return early, preventing the loaded movie from executing.

Generally, `movie_loader` will detect if the clip was removed when it was removed immediately after trying to load a movie into it. Otherwise, `movie_loader_data` will detect it if the clip was removed in-between the time when the load started and when it finished.

## Testing

3 new tests have been added, but only `avm1/load_cancel_via_removemovieclip` is passing. There is likely some underlying issue with `unloadMovie` and `MovieClipLoader.unloadClip` that is causing the other two to fail.

## Checklist


Dinnerbone (approved):
Looks reasonable to me. Thanks!

PR #23659, opened 2026-05-06
web: Map pointer coords through getBoundingClientRect

evil1morty:
## Description

Pointer events arriving at the SWF stage are mis-mapped whenever the canvas (or any ancestor) is rendered at a different visual size than its layout box — most commonly under the CSS [`zoom`](https://developer.mozilla.org/en-US/docs/Web/CSS/zoom) property, but also under CSS transforms. Concretely, with `zoom: 0.5` on a parent of the player, clicks register at the wrong stage location.

The current code in `web/src/lib.rs` is:

```rust
x: js_event.offset_x() * instance.device_pixel_ratio,
y: js_event.offset_y() * instance.device_pixel_ratio,
```

Two things break:

1. The canvas backing buffer is sized from `canvas.client_width()` times `device_pixel_ratio`. Under `zoom: 0.5` the layout box is unchanged at e.g. 1000 CSS px and the backing buffer is 1000 px — but the canvas is visually 500 px on screen.
2. `MouseEvent.offsetX` on Chromium reports the position relative to the visually rendered rect, not the untransformed padding edge. Multiplying that by `device_pixel_ratio` gives a value scaled to 0..(visual width × DPR), which doesn't match the 0..(layout width × DPR) the backing buffer expects.

This PR replaces `offsetX/Y * device_pixel_ratio` with a small helper that computes coordinates from `clientX/Y` and `getBoundingClientRect()`. The rect reflects whatever CSS `zoom` / transforms / positioning the canvas ends up with, and the backing buffer is `canvas.width × canvas.height`. The ratio `(client - rect.origin) * canvas_size / rect_size` is therefore correct under any combination of:

- CSS `zoom` on the canvas or any ancestor
- CSS transforms (`transform: scale`, etc.)
- Browser zoom
- Device pixel ratio
- CSS positioning / margins / padding

Three call sites updated (`pointermove`, `pointerdown`, `pointerup`); the wheel event doesn't pass a position so it's untouched.

`Element::get_bounding_client_rect` returns a `DomRect`, so `DomRect` is added to the web-sys feature list.

## Testing

Reproducing the original bug:

1. Visit https://safariislandsgame.com/play/?lang=en in Chromium and use the in-page `-` / `+` zoom buttons — these set CSS `zoom` on a wrapper. Clicks register offset from where the user clicks.
2. Or, in DevTools on any page, set `zoom: 0.7` on a parent of `<ruffle-player>`. Same result.

After this PR, clicks land at the cursor in both cases. Also verified with browser zoom at 75% / 100% / 125% / 150%, and `transform: scale` (which already happened to map correctly via offsetX, but is now consistent with the `zoom` path through the same helper).

`cargo fmt --all` clean. `cargo clippy -p ruffle_web --target wasm32-unknown-unknown` zero warnings.

No new automated test added — couldn't find an existing harness for pointer-event mapping under DOM transforms; happy to add one if reviewers point me at a fixture pattern.

## Checklist


kjarosh:
Can you add an integration test for it under `web/packages/selfhosted/test/integration_tests` so that each callback is tested?

evil1morty:
Added an integration test under `web/packages/selfhosted/test/integration_tests/pointer_coords`:

- 200x200 SWF stage wrapped in a `transform: scale(0.5)` parent, so the canvas's visual rect is 100x100 while its backing buffer stays 200x200.
- AS3 `Test.as` listens for `MOUSE_DOWN`, `MOUSE_UP` and `MOUSE_MOVE` on `stage` and traces the rounded `(stageX, stageY)`.
- `test.ts` dispatches synthetic `PointerEvent`s on the canvas at known `clientX/Y` inside the visual rect; the trace asserts the resulting stage coords. Under the previous `offset_x * dpr` mapping each case would land off by a factor of 2; with the new ratio mapping they land on the expected coordinates.

Three cases — one per callback (pointerdown/up/move) — to satisfy the request that each one be exercised.

evil1morty:
The CI logs show the pointer-coord values were exactly right — `onMouseDown(100,100)`, `onMouseUp(0,0)`, `onMouseMove(50,150)` all traced as expected. The failure was that I missed the synthetic `MOUSE_MOVE` Ruffle dispatches before each `MOUSE_DOWN` / `MOUSE_UP` to settle the cursor at the new position (Flash semantics). My single-element expected arrays consumed the move, then the next assertion saw the down/up still queued.

Force-pushed an amend that updates each expected list to include the leading move:

- pointerdown: `["onMouseMove(100,100)", "onMouseDown(100,100)"]`
- pointerup: `["onMouseMove(0,0)", "onMouseUp(0,0)"]`
- pointermove: unchanged

That matches the trace order the test framework already saw in the failed run.

n0samu:
How did you notice this issue? Are you scaling the Ruffle canvas on your website?
(this is a question for the human behind this PR, not Claude)

evil1morty:
I noticed it while testing on [https://safariislandsgame.com/play/?lang=en](https://safariislandsgame.com/play/?lang=en)
I used the site's built-in zoom buttons (- and + in the menu) and saw the misalignment
I've now deobfuscated their zoom logic ([https://safariislandsgame.com/play/test.js](https://safariislandsgame.com/play/test.js)) here's what they actually do:

```js
$("#all").css("zoom", scale);
$("#all").css("-moz-transform", "scale(" + scale + "," + scale + ")");
$("#all").css("MozTransform", "scale(1)");
$("#all").css("MozTransformOrigin", "center");
```
So on Firefox they use transform: scale(...), and on Chromium they use the zoom property
  The transform: scale path works fine, the issue only reproduces with the [zoom](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/zoom) rule.

n0samu:
Ah that makes sense!

web/packages/selfhosted/test/integration_tests/pointer_coords/test.ts
```diff
@@ -0,0 +1,85 @@
...
+        await playAndMonitor(browser, player, ["Loaded!"]);
+    });
+
+    it("pointerdown at the visual centre maps to stage centre", async () => {
+        const player = await browser.$("#objectElement").getElement();
+        await browser.execute((el) => {
+            const canvas = el.shadowRoot!.querySelector(
+                "canvas",
+            ) as HTMLCanvasElement;
+            const rect = canvas.getBoundingClientRect();
```
kjarosh: It's best not to use what we're testing, you can hardcode coordinates in the test instead.

kjarosh (approved):
LGTM, after fixing the test, thank you! (And sorry for the delay :( )

evil1morty:
Done — dropped `getBoundingClientRect()` from the test. The page now pins the scaled wrapper to the viewport origin (`body { margin: 0 }` + `position: absolute; top/left: 0`), so the canvas's visual rect is a fixed `(0, 0, 100, 100)`. The three dispatches use hardcoded `clientX/Y` (50,50 → stage 100,100; 0,0 → 0,0; 25,75 → 50,150) instead of measuring the canvas, so the test no longer leans on the same API the fix is verifying. Force-pushed to the same commit.

kjarosh (changes requested):
I checked out the code and noticed the test passes both before and after the fix, which means either it doesn't test it or it worked before.

evil1morty:
Good catch — you're right, and the reason is subtle: `transform: scale` doesn't actually exercise this bug. For a synthetic `PointerEvent`, `offsetX` under `transform` is reported in *unscaled layout* space, so the old `offset_x * dpr` mapping already produced the correct stage coord. The whole fixture passed with and without the fix because the chosen vector was the one path that was never broken (the same thing I noted earlier about the Safari Islands site — Firefox's `transform` path worked, only Chromium's `zoom` path misbehaved).

Fixed by switching the fixture to CSS `zoom: 0.5`, which is the real bug vector. Under `zoom`, a synthetic event's `offsetX` is reported in the zoomed/visual space, so the old mapping lands at **half** the correct stage coord. Now every case fails without the fix and passes with it:

| callback | dispatch (visual) | old `offset_x * dpr` | fixed (expected) |
|---|---|---|---|
| pointerdown | (50, 50) | (50, 50) ✗ | **(100, 100)** ✓ |
| pointerup | (30, 60) | (30, 60) ✗ | **(60, 120)** ✓ |
| pointermove | (25, 75) | (25, 75) ✗ | **(50, 150)** ✓ |

Also swapped the old pointerup-at-origin case for non-zero coords — `(0, 0)` mapped to `(0, 0)` under both formulas, so it couldn't distinguish them.

On the fixed code this stays green on chrome/firefox/edge: the helper computes `(clientX − rect.left) * canvas.width / rect.width`, and Ruffle's device→stage map then divides by `canvas.width` again, so it cancels — the stage coord depends only on the visual `getBoundingClientRect` (100px wide, origin 0), not on per-browser `zoom`/`clientWidth`/DPR quirks. Verified the geometry and both mappings in Chromium.

PR #23666, opened 2026-05-08
avm2: Fix class signatures for all classes in `flash.errors` package

Lord-McSweeney:
## Description

Makes some minor fixes to `Error`'s class signature, and implements all classes in the `flash.errors` package.

## Testing

`avm2/all_classes/errors` tests added

## Checklist


core/src/avm2/globals/Error.as
```diff
@@ -4,6 +4,7 @@ package {
     public dynamic class Error {
         {
             prototype.name = "Error";
+            prototype.message = "Error";
```
evilpie, line 7: Is that actually correct? This doesn't seem to be covered by the tests.

evilpie: (I kind of expected `message = ""` like JS)

Lord-McSweeney:
```
trace(Error.prototype.message);
```
does indeed print `Error`. There are some more changes we need to make to get `Error`'s prototype correct to FP that I'm planning for another PR, so I'll add test coverage for this in that one.

core/src/avm2/globals/flash/errors/DRMManagerError.as
```diff
@@ -0,0 +1,22 @@
...
+            super(message, id);
+            this._subErrorID = subErrorID;
+        }
+
+        public function get subErrorID():int {
+            return this._subErrorID;
+        }
+
+        public function toString():String {
+            return "DRMManagerError: '" + this.message + "', error ID:'" + this.errorID.toString() + "', subErrorID:'" + this._subErrorID.toString() + "'";
```
evilpie, line 19: This also seems special, so maybe we could test this as well?

core/src/avm2/globals/flash/errors/InvalidSWFError.as
```diff
@@ -1,6 +1,6 @@
 package flash.errors {
     public dynamic class InvalidSWFError extends Error {
-        prototype.name = "InvalidSWFError";
+        InvalidSWFError.prototype.name = "InvalidSWFError"
```
evilpie: Why is this different?

core/src/avm2/globals/flash/errors/DRMManagerError.as
Lord-McSweeney: I added test coverage for DRMManagerError in `avm2/error_tostring_more`

evilpie (approved):
Thanks for the tests.

PR #23674, opened 2026-05-09
flv: Split parse_string function into parse_string_u32 and parse_string_u16

tygyh:
## Description

Makes the function signature more readable at a glance

## Checklist

### DIFF
--- a/flv/src/script.rs
+++ b/flv/src/script.rs
@@ -2,13 +2,13 @@ use crate::error::Error;
 use crate::reader::FlvReader;
 use std::io::Seek;
 
-fn parse_string<'a>(reader: &mut FlvReader<'a>, is_long_string: bool) -> Result<&'a [u8], Error> {
-    let length = if is_long_string {
-        reader.read_u32()?
-    } else {
-        reader.read_u16()? as u32
-    };
+fn parse_string_u32<'a>(reader: &mut FlvReader<'a>) -> Result<&'a [u8], Error> {
+    let length = reader.read_u32()?;
+    reader.read(length as usize)
+}
 
+fn parse_string_u16<'a>(reader: &mut FlvReader<'a>) -> Result<&'a [u8], Error> {
+    let length = reader.read_u16()?;
     reader.read(length as usize)
 }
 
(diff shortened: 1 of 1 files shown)


n0samu (approved):

flv/src/script.rs
```diff
@@ -2,13 +2,13 @@ use crate::error::Error;
 use crate::reader::FlvReader;
 use std::io::Seek;
 
-fn parse_string<'a>(reader: &mut FlvReader<'a>, is_long_string: bool) -> Result<&'a [u8], Error> {
-    let length = if is_long_string {
-        reader.read_u32()?
-    } else {
-        reader.read_u16()? as u32
-    };
+fn parse_string_u32<'a>(reader: &mut FlvReader<'a>) -> Result<&'a [u8], Error> {
```
Lord-McSweeney, line 5: nit: Could you name these `parse_string_long` and `parse_string_short`? IMO, just by looking at the names it's hard to tell what `_u32` and `_u16` mean

Lord-McSweeney (approved):

flv/src/script.rs
tygyh: I chose to use `u32` and `u16` because they align with the naming style of Rust where you don't use euphemisms like `int` and `short`.

Lord-McSweeney: I mean `long` and `short` as to describe the length of the string, not the size of the data value. As it is it's hard to understand what the two functions do from the names.

tygyh: I only used the data types as examples of the naming style. It might be hard to tell exactly how large u32 and u16 are but there is no way of knowing the size of a short or long string.

Lord-McSweeney: This isn't about the size of the string; the value variant is literally named `Value::LongString`. Would you argue that it would be more fitting for it to be called `Value::U32String` because there's no way of knowing the size of a "long" string?

tygyh: Yes, but very cautiously since this line of reasoning could lead to all arbitrary words needing to be removed. The inner function call is `read_u32`/`u16`. I wanted to not lose that information by carrying it one abstraction layer up.

PR #23397, opened 2026-04-05
tests: Add audio amplitude assertions to test framework

kjarosh:
This allows tests to verify that audio is playing (or silent) at specific frames by checking the maximum amplitude in the audio buffer.

- Add `AudioAssertion` and `FrameSelection` to test options.
- Implement `TestRunner::test_audio` to perform amplitude checks.
- Expose the audio buffer in `TestAudioBackend`.
- Update `soundchannel_stop` test to use these new assertions.

That could make it possible to write a test for https://github.com/ruffle-rs/ruffle/pull/23333.

PR #23676, opened 2026-05-09
Slice logic cleanup, take 2

moulins:
## Description

#23124, but without bugs.

(copied from the old PR)
- Use `SwfSlice::to_subslice` instead of `to_unbounded_subslice` outside of AVM1 bytecode (which is the only place that needs to allow whole-movie access not restricted to the current tag);
- Remove unneeded arguments in `decode_tags`'s callback & `SwfSlice::resize_to_reader`;
- Implement `io::Read` on `buffer::Slice` directly (instead of `buffer::SliceCursor`);
- Remove unused code after the previous changes.

## Checklist


core/src/display_object/movie_clip.rs
```diff
@@ -4542,10 +4521,9 @@ impl<'gc, 'a> MovieClip<'gc> {
             let read = self.0.shared_cell();
             if let (Some(stream_info), None) = (&read.audio_stream_info, self.0.audio_stream.get())
             {
-                let slice = self.0.shared.get().swf.to_start_and_end(
-                    self.0.tag_stream_pos.get() as usize,
-                    self.0.tag_stream_len(),
-                );
+                let mut slice = self.0.shared.get().swf.clone();
+                slice.end = slice.start + self.0.tag_stream_len();
+                slice.start += self.0.tag_stream_pos.get() as usize;
```
moulins, line 4546: In the old PR, this read:
```rs
let mut slice = self.0.shared.get().swf.clone();
slice.start = self.0.tag_stream_pos.get() as usize;
```
Which caused the broken audio playback issues.

kjarosh:
Can we add a test for the stuff that broke the last time?

moulins:
I'm not sure how, I guess it could be done with `SoundMixer.computeSpectrum`, but that sounds very annoying to setup. :(

torokati44:
Wouldn't something using this suffice?
https://github.com/ruffle-rs/ruffle/pull/23397

moulins:
I tried making a test, but unfortunately it also fails (in a different way) in current Ruffle, and audio assertions don't support `known_failure`; This'll need a separate PR.

moulins:
I rebased this PR onto PR #23861 to be able to mark failing audio assertions as `known_failure = true`.

Lord-McSweeney (approved):

PR #23677, opened 2026-05-09
Add testutils CLI

Dinnerbone:
## Description

This adds `cargo testutils` as a utility tool for working with tests. Currently, it has two subcommands:

`cargo testutils compile` can be used to compile tests easily, without needing to actually run our tests (useful when developing tests and you don't have the right output yet; the test would always fail in Ruffle).

`cargo testutils execute` can be used to execute tests in Flash Player (and compile them if needed). **This only works on Linux right now**, future OS support is pending someone who cares enough to make those work. This allows for extremely fast iteration on Linux for "basic" tests, at the cost of having a fair few limitations:
- Only Linux
- You need to configure a list of all Flash Player executables that you have
- It does not control the test environment in any way (no sandboxing, no "only run for one frame", etc)

However, if you accept those limitations, it does make the authoring of tests _incredibly convenient_.

Both commands accept the same kind of filtering as our standard test framework (either nextest or regular `cargo test`), for convenience in switching back & forth.

## Testing

Try changing a test (e.g. `movieclip_lockroot`) and see the swf change if you run `cargo testutils compile`.

To test `cargo testutils execute`, make sure you create a `.flash_players.toml` file that looks something like this:
```toml
[[players]]
version = 32
path = "/path/to/flashplayerdebugger"
```

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -212,6 +212,8 @@ swf_version = 15
 
 Then, whenever you run the tests it'll automatically build `test.swf` (version 15) with the source code of `test.as`. Consult the [full documentation](tests/README.md) for more details about this format.
 
+There is also the ability to only compile the test and not run it - use `cargo testutils compile` to compile all swfs (which takes the same filters as tests does, e.g. `cargo testutils compile avm1/movieclip_lockroot`).
+
 ### Flash authoring tool
 
 Create a new ActionScript project. Save the `.fla` file and export an `.swf` (File -> Export -> Export Movie...).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -214,6 +214,13 @@ Then, whenever you run the tests it'll automatically build `test.swf` (version 1
 
 There is also the ability to only compile the test and not run it - use `cargo testutils compile` to compile all swfs (which takes the same filters as tests does, e.g. `cargo testutils compile avm1/movieclip_lockroot`).
 
+If you want to conveniently run the test that you create in Flash Player, we also have a `cargo testutils execute` command - this only works on Linux currently, and you must configure the Flash Player paths in a `.flash_players.toml` file which looks something like:
+```toml
+[[players]]
+version = 32
+path = "/path/to/flashplayerdebugger"
+```
+
 ### Flash authoring tool
 
 Create a new ActionScript project. Save the `.fla` file and export an `.swf` (File -> Export -> Export Movie...).
--- a/tests/src/execute.rs
+++ b/tests/src/execute.rs
@@ -1,0 +1,64 @@
+use crate::flashplayer::FlashPlayer;
+use crate::flashplayer::config::FlashHarnessConfig;
+use crate::flashplayer::environment::PlayerEnvironment;
+use crate::{TestFilterOptions, find_root_dir, load_test_dir};
+use clap::Args;
+use ruffle_core::DEFAULT_PLAYER_VERSION;
+use ruffle_fs_tests_runner::FsTestsRunner;
+use std::path::Path;
+use std::time::Duration;
+
+#[derive(Args)]
+pub struct ExecuteOptions {
+    #[command(flatten)]
+    filter: TestFilterOptions,
+
+    /// Forcefully quit the test after the given amount of seconds.
+    #[arg(long, default_value_t = 3)]
+    max_duration: u32,
+}
+
+pub fn main_execute(options: ExecuteOptions) {
+    if options.filter.is_empty() {
+        panic!(
+            "Running `execute` on every single test is inadvisable. Please specify any kind of filter."
+        );
+    }
+
+    let config = FlashHarnessConfig::load().unwrap();
+    let mut runner = FsTestsRunner::new();
+    options.filter.apply(&mut runner);
+    runner
+        .with_root_dir(find_root_dir())
+        .with_canonicalize_paths(true)
+        .with_test_loader(Box::new(move |params, register_trial| {
+            for test in load_test_dir(&params.test_dir, params.test_name) {
+                register_trial((params.test_dir_real.clone().into_owned(), test));
+            }
+        }))
+        .with_sorter(Box::new(|a, b| a.1.name().cmp(b.1.name())));
+    let tests = runner.find_tests();
+    for (test_dir, test) in tests {
+        let swf_path_real = test_dir.join(test.swf_path.as_str().trim_start_matches(['/']));
+        let swf_path_real_str = swf_path_real.to_string_lossy();
+        let environment = PlayerEnvironment::new();
+        let version = test
+            .options
+            .player_options
+            .version()
+            .unwrap_or(DEFAULT_PLAYER_VERSION);
+        let definition = config.get_player(version).unwrap_or_else(|| {
+            panic!(
+                "Could not find a Flash Player debug executable for version {} in .flash_players.toml",
+                version
+            )
+        });
+        let player = FlashPlayer::new(definition);
+        let output = player.run(
+            &environment,
+            Path::new(swf_path_real_str.as_ref()),
+            Duration::from_secs(options.max_duration as u64),
+        );
+        println!("Test output:\n{}", output);
+    }
+}
--- a/tests/src/execute.rs
+++ b/tests/src/execute.rs
@@ -5,7 +5,6 @@ use crate::{TestFilterOptions, find_root_dir, load_test_dir};
 use clap::Args;
 use ruffle_core::DEFAULT_PLAYER_VERSION;
 use ruffle_fs_tests_runner::FsTestsRunner;
-use std::path::Path;
 use std::time::Duration;
 
 #[derive(Args)]
@@ -20,6 +19,10 @@ pub struct ExecuteOptions {
     /// Forcefully quit the test after the given amount of seconds of inactivity (no log output).
     #[arg(long, default_value_t = 0.5)]
     max_idle: f32,
+
+    /// Save the output to the test's "expected output" file.
+    #[arg(short, long)]
+    save_output: bool,
 }
 
 pub fn main_execute(options: ExecuteOptions) {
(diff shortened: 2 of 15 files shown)


tests/src/main.rs
```diff
@@ -0,0 +1,57 @@
...
+        .with_context(|| format!("Couldn't load test options for test {name}"))
+        .unwrap();
+    options.into_iter().map(move |opts| {
+        Test::from_options(opts, test_dir.to_owned(), name.to_owned())
+            .with_context(|| format!("Couldn't create test {name}"))
+            .unwrap()
+    })
+}
+
+pub fn find_root_dir() -> PathBuf {
```
kjarosh, line 97: I'm not a big fan of this approach. I would personally require an env variable with the tests directory, and set one in `.cargo/config.toml` for `cargo run`:

```toml
[env]
LOCAL_RUFFLE_TESTS_SWFS_DIR = { value = "tests/tests/swfs", relative = true }
```

tests/src/main.rs
```diff
@@ -16,6 +17,42 @@ struct Opt {
...
+        help = "Exactly match filters rather than by substring"
+    )]
+    pub exact: bool,
+
+    /// A list of filters. Tests whose names contain parts of any of these
+    /// filters are skipped.
+    #[arg(
+        long = "skip",
+        value_name = "FILTER",
+        help = "Skip tests whose names contain FILTER (this flag can be used multiple times)"
```
kjarosh: Nit: missing period at the end of the sentence.

tests/src/main.rs
```diff
@@ -16,6 +17,42 @@ struct Opt {
     command: Commands,
 }
 
+// Copied out of libtest_mimic to be similar enough that you can hop between actual tests & this util CLI.
+#[derive(Args)]
+struct TestFilterOptions {
+    /// If set, filters are matched exactly rather than by substring.
+    #[arg(
+        long = "exact",
+        help = "Exactly match filters rather than by substring"
```
kjarosh: Nit: missing period at the end of the sentence.

tests/src/main.rs
```diff
@@ -16,6 +17,42 @@ struct Opt {
...
+    )]
+    pub exact: bool,
+
+    /// A list of filters. Tests whose names contain parts of any of these
+    /// filters are skipped.
+    #[arg(
+        long = "skip",
```
kjarosh, line 36: We skipped `long` in `filter`, do we have to have it here?

CONTRIBUTING.md
```diff
@@ -212,6 +212,9 @@ swf_version = 15
 
 Then, whenever you run the tests it'll automatically build `test.swf` (version 15) with the source code of `test.as`. Consult the [full documentation](tests/README.md) for more details about this format.
 
+There is also the ability to only compile the test and not run it - use `cargo testutils compile` to compile all swfs (which takes the same filters as tests does, e.g. `cargo testutils compile avm1/movieclip_lockroot`).
+
+>>>>>>> conflict 1 of 1 ends
```
kjarosh: This looks like a conflict marker

tests/src/flashplayer/config.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+impl FlashHarnessConfig {
+    pub fn load(path: &Path) -> anyhow::Result<Self> {
+        let config: Self = toml::from_str(&std::fs::read_to_string(path)?)?;
+        Ok(config)
+    }
+
+    pub fn find_and_load(start: &Path) -> anyhow::Result<Self> {
+        // Iterate every directory going up from `start` until we find a `.flash_players.toml`
```
kjarosh: Again, not a big fan of this approach, it would be better to be explicit about the location.

Maybe the same approach with an env variable defined in `.cargo/config.toml` as the default for the repo?

tests/src/flashplayer/config.rs
```diff
@@ -0,0 +1,39 @@
...
+        let mut path = start;
+        while let Some(parent) = path.parent() {
+            path = parent;
+            let config_path = path.join(".flash_players.toml");
+            if config_path.is_file() {
+                return Self::load(&config_path);
+            }
+        }
+        Err(anyhow::anyhow!(
```
kjarosh: I'm wondering, maybe instead of erroring out we could generate the `.flash_players.toml` file with a template? How is the user gonna know what to put in the file?

tests/src/flashplayer/environment.rs
```diff
@@ -0,0 +1,9 @@
+#[cfg(target_os = "linux")]
```
kjarosh: I think you can use `cfg_select!` to make it clearer:

```rust
cfg_select! {
    target_os = "linux" => {
        mod linux;
        pub use linux::*;
    }
    _ => {
        mod unsupported;
        pub use unsupported::*;
    }
}
```

tests/src/flashplayer.rs
```diff
@@ -0,0 +1,43 @@
...
+
+impl<'a> FlashPlayer<'a> {
+    pub fn new(player_definition: &'a FlashPlayerDefinition) -> Self {
+        Self { player_definition }
+    }
+
+    pub fn run(
+        &self,
+        environment: &PlayerEnvironment,
+        file: &Path,
```
kjarosh: Nit: I would call it `argument`, `swf`, or something, it's not obvious what kind of file it's referring to.

tests/src/flashplayer.rs
```diff
@@ -31,9 +32,20 @@ impl<'a> FlashPlayer<'a> {
...
+        let mut last_activity = Instant::now();
+        let mut last_log_time = None;
         while child.try_wait().unwrap().is_none() {
-            std::thread::sleep(Duration::from_millis(200));
-            if start.elapsed() > max_duration {
+            std::thread::sleep(Duration::from_millis(100));
+            let log_time = environment
+                .log_file()
+                .metadata()
+                .and_then(|m| m.modified())
```
kjarosh: Can we move this impl into the environment? `log_file()` seems to be used only here, and getting the last log time looks like it could be platform-dependent in this case (the log doesn't have to be easily backed by a file? metadata might not be accurate?).

tests/src/execute.rs
```diff
@@ -24,6 +25,10 @@ pub struct ExecuteOptions {
     /// Save the output to the test's "expected output" file.
     #[arg(short, long)]
     save_output: bool,
+
+    /// Configure how tests should be compiled, if at all
```
kjarosh: Nit: missing period at the end of the sentence.

tests/src/execute.rs
```diff
@@ -24,6 +25,10 @@ pub struct ExecuteOptions {
...
+    #[arg(short, long, default_value = "compile-silently")]
```
kjarosh: Does it make sense to silently compile tests before executing? It doesn't look obvious, I'd default to "do not compile". If you have a SWF modified, executing it will suddenly modify the SWF.

tests/src/main.rs
Dinnerbone: Good idea, done

tests/src/main.rs
Dinnerbone: It's because these were copied :D Added the period.

tests/src/main.rs
Dinnerbone: No, filter stuff aren't `--filter foo` they're just `foo`. `long` makes this `--skip`.

This is the same format as libtest/libtest_mimic, and this struct is copied from libtest_mimic, for ease of sharing args between `testutils` and `cargo test`/`nextest`

CONTRIBUTING.md
Dinnerbone: Removed from that commit

tests/src/execute.rs
Dinnerbone: I originally had it not compiling by default and a "compile yes/no" option, but moulins requested this :sweat_smile:

I'll go back to the old approach and consider adding a default option in the options later on.

tests/src/flashplayer.rs
Dinnerbone: Changed to `swf`

tests/src/flashplayer/environment.rs
Dinnerbone: Oh we have that now :D

tests/src/flashplayer/config.rs
Dinnerbone: Added a template

tests/src/flashplayer/config.rs
Dinnerbone: Sure, did that

tests/src/flashplayer.rs
Dinnerbone: Sure, good shout

tests/src/main.rs
kjarosh: Ohh that's right, sorry!

CONTRIBUTING.md
kjarosh: The conflict marker is still here 🤔

tests/src/flashplayer/environment/unsupported.rs
```diff
@@ -0,0 +1,22 @@
...
+        panic!("Unsupported platform :( Feel free to add support and open a PR!");
+    }
+
+    pub fn read_log(&self) -> String {
+        panic!("Unsupported platform :( Feel free to add support and open a PR!");
+    }
+
+    pub fn log_file(&self) -> &Path {
+        panic!("Unsupported platform :( Feel free to add support and open a PR!");
+    }
```
kjarosh, line 17: This looks like a leftover.

kjarosh (approved):

CONTRIBUTING.md
Dinnerbone: Huh! Fixing, again!

tests/src/flashplayer/environment/unsupported.rs
Dinnerbone: What do you mean?

kjarosh: `log_file` was used previously, but now it should be removed as it's unused

Dinnerbone: I think github might be doing us dirty here, can you please confirm which code you are talking about

torokati44:
Incredibly useful! 🙏

PR #23683, opened 2026-05-10
flv: Extract read function

tygyh:
## Description

Extracts a helper function to deduplicate the tests.

## Checklist

### DIFF
--- a/flv/src/script.rs
+++ b/flv/src/script.rs
@@ -290,21 +290,7 @@ mod tests {
             0x08, 0x00, 0x00, 0x00, 0x02, 0x00, 0x03, 0x01, 0x02, 0x03, 0x06, 0x00, 0x03, 0x01,
             0x02, 0x03, 0x05, 0x00, 0x00, 0x09,
         ];
-        let mut reader = FlvReader::from_source(&data);
-
-        assert_eq!(
-            Value::parse(&mut reader),
-            Ok(Value::EcmaArray(vec![
-                Variable {
-                    name: &[0x01, 0x02, 0x03],
-                    data: Value::Undefined
-                },
-                Variable {
-                    name: &[0x01, 0x02, 0x03],
-                    data: Value::Null
-                }
-            ]))
-        );
+        read(&data);
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


Dinnerbone (commented):
It's kinda weird for a method named "read" to also do the testing. It just so happens that each of these tests are testing for the same values but that's not entirely by design, I think?

tygyh:
> It's kinda weird for a method named "read" to also do the testing.

I should have said I'm very open to better names for the function. Do you have any suggestions?

flv/src/script.rs
```diff
@@ -336,7 +308,11 @@ mod tests {
             0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x03, 0x01, 0x02, 0x03, 0x06, 0x00, 0x03, 0x01,
             0x02, 0x03, 0x05, 0x00, 0x00, 0x09,
         ];
-        let mut reader = FlvReader::from_source(&data);
+        read(&data);
+    }
+
+    fn read(data: &[u8]) {
```
kjarosh, line 314: What's the point of this refactor? I get that _technically_ it's code deduplication, but now instead of having given/when/then, we have given/_read_, where `read` is a mysterious function that parses the data and asserts on a value that doesn't make any sense from `read`'s perspective.

kjarosh: Sorry, I didn't notice prior comments. My concerns are similar to Dinnerbone's.

tygyh: The whole point is the deduplication. If these actually test for different things, why do they give the same result?

kjarosh: If you write a test for addition, 6+7, (-1)+14, and 15+(-2) will result in the same value: 13, despite having different inputs and testing different things.

tygyh: Then only the input data and a descriptive name is relevant. I think we should just come up with a better name for `read`

torokati44: Thank you, but sorry, I also don't think this is worth it. For one, it breaks the pattern of all the other tests, where the input and output are clearly laid out in-place. Second, it doesn't satisfy the "rule of three": https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

tygyh: It is used three times. `read_value_ecmaarray_shortlen()`, `read_value_ecmaarray()`, `read_value_ecmaarray_longlen()`

torokati44: Oh, my bad. But I still stand by my other point of this breaking the regular pattern of tests.

PR #23701, opened 2026-05-12
web/audio: Pin AudioContext to 44.1 kHz

### LINKED ISSUE #23634: Sound Slowdown/Stuttering/Distortion with 384khz Sound Card
### Describe the bug

so i'm used to it by now that 384khz sound setting can be incompatible sometimes.
in case of ruffle.rs it leads to a little sound slowdown with distortion at the start of the flash file (any flash file).

Windows 11, ruffle.rs in newest Chrome Browser 147, AMD Ryzen, Soundblaster G6 with 32bit 384khz.
lowering the setting down to for example 192khz apparently fixes it.

while this can be fixed on the user site, it probably would be even better if ruffle can just deal with 384khz, so one doesn't has to ask users to change their soundcard setting.
i know its an very rare case, since many people nowadays probably use onboard and not change their setting to 384khz in the first place, but it would be nice if for those rare cases it can be fixed.

### Expected behavior

crystal clear not distorted sound.

### Content Location

[then_pinkie_found_a_delorean_by_alexstrazse_d4fir6j.zip](https://github.com/user-attachments/files/27378404/then_pinkie_found_a_delorean_by_alexstrazse_d4fir6j.zip)

### Affected platform

Browser's extension

### Operating system

Windows 11

### Browser

Chrome 147

### Additional information

_No response_

SuchAFuriousDeath:
Fixes #23634

## Description
web/audio: Pin AudioContext to 44.1 kHz
SWF audio sources top out at 44.1 kHz, so running the Web Audio
pipeline at the device sample rate gains no fidelity and caused
stutter at startup on high-rate devices (e.g. 192/384 kHz sound
cards on Windows): each ping-pong buffer expired faster than the
JS event loop could refill it. Request a 44.1 kHz AudioContext
and let the browser resample to the device rate per Web Audio
§1.2.1, which mandates that user agents MUST resample on output
when contextOptions.sampleRate differs from the device rate.

## Testing

_How can we test this PR?_
  <details>
  <summary>Reproduction patch — forces <code>AudioContext</code> to 384 kHz</summary>

  Apply on top of this PR, build with `npm run build:debug`, open the demo in **Chrome** (Firefox caps `AudioContext` at 192 kHz). Without the fix you'll hear stutter at the start, with the fix, it's fine.

  ```diff
  --- a/web/Cargo.toml
  +++ b/web/Cargo.toml
  @@ -67,7 +67,7 @@
   [dependencies.web-sys]
   workspace = true
   features = [
  -    "AddEventListenerOptions", "AudioBuffer", "AudioBufferSourceNode", "AudioContext",
  +    "AddEventListenerOptions", "AudioBuffer", "AudioBufferSourceNode", "AudioContext", "AudioContextOptions",
       "AudioDestinationNode", "AudioNode", "AudioParam", "Blob", "BlobPropertyBag",
  --- a/web/src/audio.rs
  +++ b/web/src/audio.rs
  @@ -57,7 +57,9 @@
       const NORMAL_PROGRESS_RANGE_MAX: f64 = 0.75;

       pub fn new(log_subscriber: Arc<Layered<WASMLayer, Registry>>) -> Result<Self, JsError> {
  -        let context = AudioContext::new().into_js_result()?;
  +        let opts = web_sys::AudioContextOptions::new();
  +        opts.set_sample_rate(384_000.0);
  +        let context = AudioContext::new_with_context_options(&opts).into_js_result()?;
           let sample_rate = context.sample_rate();
  ```

  </details>

## Checklist

### DIFF
--- a/web/src/audio.rs
+++ b/web/src/audio.rs
@@ -12,7 +12,7 @@ use tracing_subscriber::Registry;
 use tracing_subscriber::layer::Layered;
 use tracing_wasm::WASMLayer;
 use wasm_bindgen::prelude::*;
-use web_sys::{AudioContext, AudioScheduledSourceNode};
+use web_sys::{AudioContext, AudioContextOptions, AudioScheduledSourceNode};
 
 pub struct WebAudioBackend {
     mixer: AudioMixer,
@@ -48,8 +48,19 @@ impl WebAudioBackend {
     const NORMAL_PROGRESS_RANGE_MAX: f64 = 0.75;
 
     pub fn new(log_subscriber: Arc<Layered<WASMLayer, Registry>>) -> Result<Self, JsError> {
-        let context = AudioContext::new().into_js_result()?;
+        // Pin the AudioContext to 44.1 kHz. SWF audio sources top out at 44.1 kHz, so
+        // there's no fidelity to gain from running our pipeline at the device rate, and
+        // running at the device rate caused stutter at startup on high-rate devices
+        // (e.g. 192/384 kHz cards) because each buffer expired faster than the JS event
+        // loop could refill it. Per Web Audio §1.2.1, the user agent MUST resample on
+        // output if the requested rate differs from the device rate.
+        // See: https://www.w3.org/TR/webaudio-1.1/#AudioContext-constructors
+        let opts = AudioContextOptions::new();
+        opts.set_sample_rate(44_100.0);
+
+        let context = AudioContext::new_with_context_options(&opts).into_js_result()?;
         let sample_rate = context.sample_rate();
+
         let mut audio = Self {
             context,
             mixer: AudioMixer::new(2, sample_rate as u32),
--- a/web/Cargo.toml
+++ b/web/Cargo.toml
@@ -67,7 +67,7 @@ features = ["audio", "mp3", "aac", "nellymoser", "default_compatibility_rules", 
 [dependencies.web-sys]
 workspace = true
 features = [
-    "AddEventListenerOptions", "AudioBuffer", "AudioBufferSourceNode", "AudioContext",
+    "AddEventListenerOptions", "AudioBuffer", "AudioBufferSourceNode", "AudioContext", "AudioContextOptions",
     "AudioDestinationNode", "AudioNode", "AudioParam", "Blob", "BlobPropertyBag",
     "ChannelMergerNode", "ChannelSplitterNode", "ClipboardEvent", "DataTransfer", "Element",
     "EventTarget", "GainNode", "Headers", "HtmlCanvasElement", "HtmlDocument", "HtmlElement", "HtmlFormElement",


torokati44:
Ideally, we should be able to get a fixed 44.1 kHz sample rate output on the JS/WASM side, and let the browser/OS/driver upsample it however it likes (in, I presume, a lot more efficient manner), see: https://github.com/WebAudio/web-audio-api/issues/232#issuecomment-24244867
EDIT: Because I assume FP (or SWF content) could never really output anything higher sample rate than that (not even with [SampleDataEvent](https://github.com/ruffle-rs/ruffle/pull/23632), I think)...
But if https://stackoverflow.com/a/14512062 still holds[^1], this is the next best thing...
From https://www.w3.org/TR/webaudio-1.1/#AudioContext-constructors:
> If contextOptions.sampleRate differs from the sample rate of the output device, the user agent MUST resample the audio output to match the sample rate of the output device.

[^1]: Look at me linking to a 13-year old SO answer in 2026!

web/src/audio.rs
```diff
@@ -50,10 +59,23 @@ impl WebAudioBackend {
     pub fn new(log_subscriber: Arc<Layered<WASMLayer, Registry>>) -> Result<Self, JsError> {
         let context = AudioContext::new().into_js_result()?;
         let sample_rate = context.sample_rate();
+
+        // Scale the buffer-size limits by a power-of-two factor based on the AudioContext sample
+        // rate so the algorithm operates over the same span of audio time on devices running at
+        // higher rates. Powers of two are required because the algorithm halves/doubles the size.
+        let scale = ((sample_rate as u32) / Self::REFERENCE_SAMPLE_RATE)
+            .max(1)
+            .next_power_of_two();
```
torokati44: Note: `scale` will be 8 for a `sample_rate` of 384000, even though 384000/44100 is something like 8.7, but it's probably good enough.

torokati44:
Code diff itself LGTM!

SuchAFuriousDeath:
> Ideally, we should be able to get a fixed 44.1 kHz sample rate output on the JS/WASM side, and let the browser/OS/driver upsample it however it likes (in, I presume, a lot more efficient manner), see: [WebAudio/web-audio-api#232 (comment)](https://github.com/WebAudio/web-audio-api/issues/232#issuecomment-24244867) EDIT: Because I assume FP (or SWF content) could never really output anything higher sample rate than that (not even with [SampleDataEvent](https://github.com/ruffle-rs/ruffle/pull/23632), I think)... But if https://stackoverflow.com/a/14512062 still holds[1](#user-content-fn-1-76de9240fcae042fc41d8143221a597b), this is the next best thing... From https://www.w3.org/TR/webaudio-1.1/#AudioContext-constructors:
> 
> > If contextOptions.sampleRate differs from the sample rate of the output device, the user agent MUST resample the audio output to match the sample rate of the output device.
> 
> ## Footnotes
> 
>     1. Look at me linking to a 13-year old SO answer in 2026! [↩](#user-content-fnref-1-76de9240fcae042fc41d8143221a597b)

I think MP3 in an SWF and an AAC in FLV allow more than 44.1 kHz, but I think doing it your way is better. Should I adjust the PR?

torokati44:
Well, I'm not sure it's possible... From a really cursory research, you can request an AudioContext with a given sample rate, but that's more of a hint than an obligation to the browser. And whatever sample rate you actually get the context with, JS/WASM will have to accept and adapt. And every node in an AudioContext runs with the same sample rate, so it's not like we could add a resampler node after the script node.

But all of this is just the impression I got from various sources, and can be wrong.

At the very least we should try requesting a 44.1kHz (maybe 48kHz at most) context (maybe we already do), but if we get one with a higher sample rate, this should still be useful.

SuchAFuriousDeath:
@torokati44 

Disclaimer: This was researched by Claude Code and fact-checked by me after the fact and it seems to make sense.

Source-code confirmation that all three engines validate the requested `AudioContext` `sampleRate` and resample to the device rate per [Web Audio §1.2.1 — Constructors](https://www.w3.org/TR/webaudio-1.1/#AudioContext-constructors):

**Chromium (Chrome/Edge/Opera/Brave/etc.)** — `audio_context.cc`, [validation L475-L486](https://github.com/chromium/chromium/blob/afdb137e994b76bd426178bd24c6094b7866210a/third_party/blink/renderer/modules/webaudio/audio_context.cc#L475-L486) → throws `NotSupportedError` for out-of-range; otherwise [passed to `RealtimeAudioDestinationNode` L619-L621](https://github.com/chromium/chromium/blob/afdb137e994b76bd426178bd24c6094b7866210a/third_party/blink/renderer/modules/webaudio/audio_context.cc#L619-L621) which resamples to device.

**WebKit (Safari)** — `AudioContext.cpp` [L123-L124](https://github.com/WebKit/WebKit/blob/a02317c7845df741fdc35eea02278e28e5dd4752/Source/WebCore/Modules/webaudio/AudioContext.cpp#L123-L124) (`isSupportedSampleRate` else `NotSupportedError "sampleRate is not in range"`). Platform destination [`AudioDestinationCocoa.cpp` L76](https://github.com/WebKit/WebKit/blob/a02317c7845df741fdc35eea02278e28e5dd4752/Source/WebCore/platform/audio/cocoa/AudioDestinationCocoa.cpp#L76) inherits `AudioDestinationResampler` (name confirms).

**Gecko (Firefox)** — `AudioContext.cpp` [L270-L278](https://searchfox.org/mozilla-central/rev/19735aaf9711306ab2c2299c27a6a2588525ce8a/dom/media/webaudio/AudioContext.cpp#270-278) — exact error string I hit at 384 kHz in Firefox when testing.

All three: out-of-range → `NotSupportedError`; in-range → resample to device. 44 100 Hz is always in range (spec mandates ≥ `[8000, 96000]` window; all engines exceed that). So requesting 44.1 kHz on `AudioContext` is reliable across every conformant engine, and the W3C `MUST resample` applies.

So we should be fine with just setting the context rate to 44.1 kHz and the browser does everything for us. No?

Not exactly sure what you mean by 'script node' tbh.

torokati44:
Indeed, I had an afterthought (which I missed to comment it seems) that in

 >If contextOptions.sampleRate differs from the sample rate of the output device, the user agent MUST resample the audio output to match the sample rate of the output device.

the "user agent" refers to the browser or the JS/WASM code running within. It is usually the former, and looks like it's that in this case too. I don't know what the SO answer was about then...

So yeah, if we accept 44.1kHz in all cases, that should be fine.

SuchAFuriousDeath:
I adjusted the PR :saluting_face:

torokati44:
> Not exactly sure what you mean by 'script node' tbh.

I was under the impression that we output using a `ScriptProcessorNode`, but no, it's `AudioBufferSourceNode`, so disregard that.

torokati44:
One last thing is that indeed I don't know whether external audio (FLV/MP3/AAC) could be played by FP in proper 48kHz/96kHz, and we'd "lose some clarity" there...

SuchAFuriousDeath:
> I don't know what the SO answer was about then...

I think it was originally posted in a time where there was no api to request a different rate.

There is actually a new answer under that post saying that there is an api to do this now: https://stackoverflow.com/a/63238079 (and it's precisely the one used in this PR)

SuchAFuriousDeath:
> One last thing is that indeed I don't know whether external audio (FLV/MP3/AAC) could be played by FP in proper 48kHz/96kHz, and we'd "lose some clarity" there...

I think that's probably really niche with Flash and also I'm not an audio guy, but can humans even tell the difference? :D

But actually if Flash can do it, maybe Ruffle should be able to as well.

SuchAFuriousDeath:
> > One last thing is that indeed I don't know whether external audio (FLV/MP3/AAC) could be played by FP in proper 48kHz/96kHz, and we'd "lose some clarity" there...
> 
> I think that's probably really niche with Flash and also I'm not an audio guy, but can humans even tell the difference? :D
> 
> But actually if Flash can do it, maybe Ruffle should be able to as well.

Although I'm not sure if we should worry about this in this PR, maybe we can file a separate issue for the future?

torokati44:
Eh, we should be safe...

torokati44 (approved):
LGTM, thanks, but please also update the PR title!

SuchAFuriousDeath:
@torokati44 Updated :saluting_face:

Issue #17204, opened 2024-07-20
Sound clipping/sound temporarily loud issue - Warning!


Kz-92:
### Describe the bug

VOLUME WARNING

when playing a lot of games using Ruffle, sounds that are a certain volume can often start playing very loud and then adjust to a lower volume, sound like they are "clipping" or are cut off, or play extremely boosted. this is not uncommon or limited to these games. headphones users beware when watching these vids, (or rather, even playing these games using ruffle...)
this issue is to collect a few examples to one issue

a non exhaustive list:

bazooka boy
bounzy
fancy pants world 3
Groar
mineblocks
super fighters
wpn fire

if people are unsure this is true, here are some videos, volume warning. The Ruffle issue is shown first, then the flash expected result is after.

bazooka boy: occasionally when firing rockets, some are louder than others. seems normal, but is not fp behaviour.

https://github.com/user-attachments/assets/1a89f550-69e7-4c3b-995e-7ecf6e1dc0ef

bounzy: nearly every shot fired "clips" the audio, slowdown worsens the sound

https://github.com/user-attachments/assets/aceacc3e-bd9d-4d0c-9aa7-a7c03ba38ce9

fpaw3: bonking your head blows out the audio. jumping from water sometimes blows out the audio.

https://github.com/user-attachments/assets/d0711eb1-459b-4e53-874e-953dcf5f47bd

Groar: the music starts off much louder, then once things load it quiets. dinosoar groans and roars can also be heard excessively loud.

https://github.com/user-attachments/assets/afdf70a6-bf2f-4cd5-ae3d-ede12aea29d5

mineblocks: animals blow out all audio. water blows out all audio. sometimes random actions can blow out audio.

https://github.com/user-attachments/assets/2ded2a40-c31a-477c-b64d-9ab8af04e616

super fighters: jumping "clips" the audio

https://github.com/user-attachments/assets/b6f60888-d332-4123-bf20-28f41a6560ff

### Expected behavior

there should be absolutely no clipping, boosting, or (abnormal) volume switching of any swf or sound in a swf.
these are not reproducible using unmodified fp in any way

### Content Location

[575163_Superfighters.zip](https://github.com/user-attachments/files/16322569/575163_Superfighters.zip)
[606312_Bazooka_Boy.zip](https://github.com/user-attachments/files/16322570/606312_Bazooka_Boy.zip)
[...]

PR #23702, opened 2026-05-12
Fix mishandling of negative volume in SoundTransform.concat()

### LINKED ISSUE #9763: The Fancy Pants Adventures: World 3: Earrape Sound
### Describe the bug

When bumping my head agaisnt a specific part of the first level, it plays a extrememly lound and earrape sound.

https://user-images.githubusercontent.com/64917206/221641935-d630bbdf-bc51-451a-9eb8-9e90ea60b0b2.mp4

### Expected behavior

The sound would play normally like on other places.

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

110.0.5481.177 

### Additional information

_No response_

### LINKED ISSUE #23700: Earrape audio bug
### Describe the bug

(Self-repost from the help channel)

When both the "deadly laser" and conveyor belt audio are playing at the same time. one gets corrupted and becomes "earrapey" (specifically the conveyor belt audio if I'm not mistaken)

https://github.com/user-attachments/assets/b91476de-1409-4db2-8dba-0ce4b73b7110

### Expected behavior

Both looping audios would play normally

### Content Location

Testing on my personal site caused this to happen. https://crabapples.xyz/

### Affected platform

Self-hosted version

### Operating system

Windows 11

### Browser

64-bit Firefox 150.0.2 (Replication is possible on Safari but version is unknown)

### Additional information

I have a theory this is something related to Ruffle not being able to manage multiple looping audio tracks

Supposedly related to #9763

n-raine:
## Description

The code in `display_object::SoundTransform.concat()` says:
```
         // It seems like Flash masks the results below to 30-bit integers:
         // * Negative values are equivalent to their absolute value (their sign bit is unset).
         // * Specifically, 0x40000000, -0x40000000 and -0x80000000 are equivalent to zero.
        const MASK: i32 = (1 << 30) - 1;

        self.volume = (i64::from(self.volume) * i64::from(other.volume) / MAX_VOLUME) as i32 & MASK;
```

But two's complement integers don't just have a "sign bit" - negative numbers are entirely inverted (and incremented by one). Masking out the top two bits results in a ridiculously large positive number, causing incredible audio distortion. Ow.

Add a `.abs()`  before the mask to fix this.

Fixes #23700.
Fixes #9763.
Unfortunately, some of the games in #17204 still have the issue even after this fix, so there must be more than one bug causing it.

## Testing

From what I understand, an automatic test can't be written for this as the current framework can only verify image output and `trace()`s. Manually verifiable by going to the third level of Red Ball 4 Volume 3 and reaching the conveyor belt area (unlike what #23700 says, this seems to happen at any conveyor belt, not just when its sound effect and the lasers' are playing simultaneously). Without the fix, this will immediately result in incredibly distorted audio.

## Checklist

[-] All of my commits are properly scoped, compile successfully, and pass all tests. (They compile and work fine, but I can't figure out how to get tests to run -- `cargo test --all` fails wildly in compilation on my machine, even before this change. I can't imagine this would affect any of the tests though.)


n-raine:
I don't get why Test Rust / Test Rust stable / windows-2025 is failing... it says it failed on "cache cargo output", but there doesn't seem to be any error output on that step. Is this an issue with the CI or with my code?

torokati44:
We can do audio tests easily now too:
https://github.com/ruffle-rs/ruffle/pull/23397

n-raine:
Unfortunately I don't have CS6 and I don't know how to create a test SWF for an audio issue without it...

ChrisCPI:
The JPEXS decompiler can allow you to import your own sounds to a blank/existing SWF, and edit the ActionScript to play them.

n-raine:
Oh good idea! Working on a test now

n-raine:
Done! While writing the test I figured out that for the bug to be obvious, the sound has to be panned in addition to the volume being negative, as the multiplication that bakes it in in `core/src/backend/audio.rs` uses a wrapping `i32` multiplication, and `100 * (x & ((1 << 30) - 1))` happens to wrap back around to a relatively small negative number.

The test swf is a minimal reproduction case that just plays an embedded MP3 of a sine wave at -0.2 volume and 0.5 panning to the right.

torokati44:
In addition (or instead), could the SoundTransform values themselves be (also) tested with a textual test? 🤔 That would also be more easily comparable to FP output.

n-raine:
I don't think the result of any `.concat()` is ever visible to the ActionScript, just the individual `SoundTransform`s.

torokati44:
Ah okay then, my bad.

Dinnerbone (commented):
Nice work!

Dinnerbone (approved):

Dinnerbone:
Nit: Please prefix commits with e.g. `core:` for core changes. I'll add it in this PR for you.

PR #23703, opened 2026-05-12
tests: Convert a bunch of AVM1 tests to Rascal

Dinnerbone:
## Description

There's so many unshaven yaks!

This is a manual process so it's slow - however, it's been pretty valuable. I've already discovered that some tests are being fooled by the compiler, it optimized away some things we wanted to check and we were instead just testing a constant value :D

I used #23677 to test every file in Flash Player whilst I'm at it. (Also it turns out that `add2` has different output on windows vs linux! Not related to this PR, just interesting.)

With the exception of `action_to_integer` which fell victim to the optimization bug, all of the tests are fully identical bytecode to the original ones, with a couple of small exceptions: 
- Some registers got shuffled, I haven't nailed which order it allocates them yet
- I added `fscommand("quit")` to each test to make them quit the FP faster
- There's an mtasc test and I don't strive for identical bytecode to mtasc, only Flash

Feel free to manually check and verify though!

## Testing

You can manually check the before and after SWFs to see if their bytecode changed, if you wish.

## Checklist


moulins:
This looks great! Looks like getting fooled by the compiled led us to not implementing `ToInteger` properly... that test will probably have to be marked as `known_failure` for now.

A small mistake: there's a stray `old.swf` file left in the `as2_super_via_manual_prototype` test.

Dinnerbone:
Removed the `old.swf` (good catch!) and fixed the integer test; it was failing because I broke the disabling of optimizations, but that's fixed now!

PR #23711, opened 2026-05-14
chore: Bump jpegxr git ref, for bindgen 0.72

torokati44:
Since https://github.com/ruffle-rs/jpegxr/pull/5 just got merged, might as well...

This will allow us to drop `rustc-hash` 1.x duplicate after https://github.com/ruffle-rs/ruffle/pull/23274.
Not that it is particularly a problem, it's just a pet goal of mine to shepherd the dependency tree.

Sorry for deleting the PR template, but for `chore` stuff I don't think it applies that well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -340,9 +340,9 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "bindgen"
-version = "0.70.1"
+version = "0.72.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f49d8fed880d473ea71efb9bf597651e77201bdd4893efe54c9e5d65ae04ce6f"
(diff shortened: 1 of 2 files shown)


Dinnerbone (approved):

torokati44:
> This will allow us to drop rustc-hash 1.x duplicate after https://github.com/ruffle-rs/ruffle/pull/23274.

Welp, this was just an assumption, but it looks like wgpu/naga 28 and 29 still depend on rustc-hash 1.x. My bad. Anyway, I don't think any harm got done, it was just mostly pointless.

Dinnerbone:
Actually the bump hopefully means we can once again build with this on Linux, which a few of us could not for a while

torokati44:
Huh, I was not aware - how come? (Granted, I'm months behind on reading at least 3 Discord channels... 🫠)

SuchAFuriousDeath:
> Actually the bump hopefully means we can once again build with this on Linux, which a few of us could not for a while

That's why I pushed it through in the first place :D

PR #23717, opened 2026-05-15
web: Create FontStyle enum

tygyh:
## Description

Extracts an enum to reduce the number of parameters. Also opens up for other styles such as strikethrough.

## Checklist

### DIFF
--- a/web/src/ui/font_renderer.rs
+++ b/web/src/ui/font_renderer.rs
@@ -19,6 +19,24 @@ pub struct CanvasFontRenderer {
     descent: f64,
 }
 
+struct FontStyle {
+    name: String,
+}
+
+impl FontStyle {
+    fn new(italic: bool, bold: bool) -> FontStyle {
+        let mut styles = Vec::new();
+        if italic {
+            styles.push("italic");
+        }
+        if bold {
+            styles.push("bold");
+        }
+        let name = styles.join(" ");
+        FontStyle { name }
+    }
+}
+
 impl CanvasFontRenderer {
     /// Render fonts with size 64px. It affects the bitmap size.
     const SIZE_PX: f64 = 64.0;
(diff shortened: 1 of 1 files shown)


Lord-McSweeney:
AFAIK FP special-cases bold and italic font lookups, so we only ever need to consider those two styles. IMO two `bool`s are much simpler than an enum that can represent all possible states of those two booleans (the code is much longer this way). Is there any other benefit to doing it this way?

n0samu:
Yeah I don't really like this and I don't see how it "opens up" for anything, especially since dealing with every combination of bools using an enum gets exponentially worse with the number of bools...

tygyh:
Since italic and bold are connected I wanted to make it explicit. I'll redo this as a struct holding the bools in a little while

web/src/ui/font_renderer.rs
```diff
@@ -19,6 +19,33 @@ pub struct CanvasFontRenderer {
     descent: f64,
 }
 
+enum FontStyle {
+    ItalicBold,
+    Italic,
+    Bold,
```
kjarosh: In theory, `bold` is not font style, it's font weight. See https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font

kjarosh:
> Also opens up for other styles such as strikethrough.

That's not useful for us because of a couple of reasons:
1. Flash does not support strikethrough (if it does, please let me know where),
2. even if it did, we are using canvas renderer for rendering glyphs, not text, and strikethrough is text decoration, it doesn't affect glyph rendering.

kjarosh:
I'm not a big fan of the refactor either, Flash doesn't really support more than italic/bold when it comes to glyph variants, and both refer to different things: font style and font weight. While we are creating a font string based on font style, weight, size, and family (see https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font), passing `italic` and `bold` separately seems reasonable.

Making enums for them (`FontStyle`, `FontWeight`) doesn't really make sense, because Flash supports only two variants (italic, non-italic; bold, non-bold).

> Since italic and bold are connected I wanted to make it explicit. I'll redo this as a struct holding the bools in a little while

`italic` and `bold` are related to each other the same way as `size` and `family` are, all of them are a set of font properties that we are passing as arguments.

tygyh:
I enjoy how petty/meticolous you are. I'm not getting away with any guesses. I'll have something substantial ready soon. Meanwhile, what are the valid font families and what range is valid for size?

PR #23723, opened 2026-05-16
 🤖 avm2: Resolve relative URLs in NetConnection.connect() (close #23201)

### LINKED ISSUE #23201: NetConnection.connect() not resolving relative URLs / dropping HTTP calls
### Describe the bug

I'm trying to use Ruffle to revive an old AS3 site (https://joshstrike.com) that relies on NetConnection AMF0 calls to load data. When the site loads, it makes an initial call to fetch preferences from a PHP script. It waits for the NetConnection callback before running the intro animation. The odd behavior I'm seeing in Ruffle is that no outbound request is ever made, and yet the callback is called anyway, with an undefined response. After the initial intro animation, when the program tries to access the data it expected to have loaded, it finds them undefined and crashes.

I put the question to Claude, and believe it found the answer. The SWF is making a call to a PHP script at a relative URI. Apparently in Ruffle, URLLoader and other methods call activation.context.navigator.resolve_url() before checking the connection scheme, but currently NetConnection does not. Since the relative path isn't recognized as https, it branches to avm2_stub_method!( ... "with non_null, non_http command")

Could NetConnection be patched to resolve relative URLs before checking the scheme?

### Expected behavior

NetConnection calls to relative paths should be treated as http calls. Calls to rtmp are always absolute pathed so this hopefully wouldn't negatively affect future compatibility with that scheme.

### Content Location

https://joshstrike.com

### Affected platform

Desktop app

### Operating system

MacOS Monterey

### Browser

_No response_

### Additional information

_No response_

MavenRain:
## Description

`NetConnection` calls to relative paths are treated as http.

Addresses #23201 

## Checklist

### DIFF
--- a/core/src/avm2/globals/flash/net/net_connection.rs
+++ b/core/src/avm2/globals/flash/net/net_connection.rs
@@ -31,16 +31,27 @@ pub fn connect<'gc>(
 
     if let Some(url) = url {
         let url_lower = url.to_ascii_lowercase();
+        let is_already_http = url_lower.starts_with(WStr::from_units(b"http://"))
+            || url_lower.starts_with(WStr::from_units(b"https://"));
 
-        if url_lower.starts_with(WStr::from_units(b"http://"))
-            || url_lower.starts_with(WStr::from_units(b"https://"))
-        {
+        // Already-absolute HTTP(S) URLs are passed through unchanged.  Other URLs
+        // are resolved against the SWF's location so that relative HTTP(S) paths
+        // reach Flash Remoting, matching URLLoader.load() behavior. Closes #23201.
+        let http_url = if is_already_http {
+            Some(url.to_string())
+        } else {
+            let url_string = url.to_string();
+            activation
+                .context
+                .navigator
+                .resolve_url(&url_string)
+                .ok()
+                .and_then(|u| matches!(u.scheme(), "http" | "https").then(|| u.to_string()))
+        };
+
+        if let Some(http_url) = http_url {
             // HTTP(S) is for Flash Remoting, which is just POST requests to the URL.
-            NetConnections::connect_to_flash_remoting(
-                activation.context,
-                connection,
-                url.to_string(),
-            );
+            NetConnections::connect_to_flash_remoting(activation.context, connection, http_url);
         } else {
             avm2_stub_method!(
                 activation,


MavenRain:
Gentle bump: this one's been green and mergeable for a while with no review yet.  @Lord-McSweeney / @danielhjacobs, would appreciate a look when there's bandwidth.

danielhjacobs:
I think this one could use a test as well. I know you said you didn't have a direct way to test Flash, but if you're willing to download Flash debug downloads, I can make recommendations to allow you to do that.

1) In tests/tests/swfs/avm2/, make a directory for this new test.
2) The test.toml should use this, and above compilers should log_fetch if you want to log fetch requests (see existing NetConnection tests)
```toml
[[compilers]]
type = "Asc"
target = "test.swf"
class = "Test" # Or whatever class the AS file uses
scripts = ["Test.as"] # Or whatever name you give the test AS file
swf_version = 9 # Or whatever SWF version you want to test (https://github.com/ruffle-rs/ruffle/wiki/SWF-version-chart)
```

3) Write your test AS file that provides one or more test cases for your fix.

4) Run `cargo testutils compile [test_file_directory]` which will compile your test AS file into an SWF using Asc.

5) Download the correct Flash Player Debugger for your OS from https://web.archive.org/web/20220401020702/https://www.adobe.com/support/flashplayer/debug_downloads.html

6) Create a file named mm.cfg in your user directory:

- Windows: C:\Users\\[Username]\
- Mac: /Library/Application Support/Macromedia/
- Linux: /home/[username]/

  Contents:
```cfg
ErrorReportingEnable=1
TraceOutputFileEnable=1
```

7) To allow your specific directory to bypass local-with-filesystem, create a directory for FlashPlayerTrust and in it make a file called tests.cfg
- Windows: %APPDATA%\Macromedia\Flash Player\\#Security\FlashPlayerTrust\
- Mac: /Users/[Username]/Library/Preferences/Macromedia/Flash Player/#Security/FlashPlayerTrust/
- Linux: /home/[username]/.macromedia/Flash_Player/#Security/FlashPlayerTrust/

  Contents:
- For Windows: C:\Users\\[Username]\path\to\ruffle\tests\tests\swfs
- For Mac: /Users/[Username]/path/to/ruffle/tests/tests/swfs
- For Linux: /home/[username]/path/to/ruffle/tests/tests/swfs

8) Open the SWF file

9) Find trace logs:

- Windows: %APPDATA%\Macromedia\Flash Player\Logs\flashlog.txt
- Mac: ~/Library/Preferences/Macromedia/Flash Player/Logs/flashlog.txt
- Linux: ~/.macromedia/Flash_Player/Logs/flashlog.txt

10) At the same time, you can use and run a Python test server like this to see the results log_fetch would print: https://github.com/danielhjacobs/ruffle/blob/828061457cc12746eb8d70a6375f1aec38a0a364/tests/tests/swfs/avm2/netconnection_serialize_arrays/server.py

11) Combine trace logs and Python server results in the correct way into output.txt in the test directory

12) Confirm Ruffle prints the same result by running `cargo test [test_file_directory]`

MavenRain:
@danielhjacobs  This is really helpful, thank you.  Yes, I'd like to take you up on it.  I'll download the Flash Player Debugger, set up mm.cfg and the FlashPlayerTrust entry as you described, and add an avm2/ test for `NetConnection.connect()` with a relative URL, using cargo testutils compile and log_fetch so the test asserts what FP actually does with the resolved URL.

One quick question: for this case the behavior I want to pin is that a relative URL gets resolved against the SWF's URL, and that a non-http resolved scheme is rejected.  Is `log_fetch` on the `NetConnection` request the right signal to assert that, or would you capture it differently?  I'll follow the existing `NetConnection` tests as the template either way.

Thanks again for the detailed write-up.

Lord-McSweeney:
Hi! This pull request hasn't been touched in a while, and there are still some unaddressed review comments on it. I'm going to close it now- feel free to reopen it or open a new PR if you can address those concerns.

PR #23726, opened 2026-05-16
avm2: Print full error details when logging caught error

### LINKED ISSUE #23687: Regression: AvmError Debug implementation outputs no information
### Describe the bug

Since commit e2bc60f75e, the `Debug` implementation for `ErrorData::AvmError` only outputs "`AvmError`", which is obviously unhelpful. This means that e.g. caught exception logging is now useless:
```
2026-05-10T19:33:19.689842Z  INFO run_frame:run_all_phases_avm2: avm_caught: Caught exception: AvmError
2026-05-10T19:33:19.689872Z  INFO run_frame:run_all_phases_avm2: avm_caught: Caught exception: AvmError
2026-05-10T19:33:19.691888Z  INFO run_frame:run_all_phases_avm2: avm_caught: Caught exception: AvmError
2026-05-10T19:33:19.691917Z  INFO run_frame:run_all_phases_avm2: avm_caught: Caught exception: AvmError
```

I'm totally new to Ruffle so I don't have enough understanding to know what the original commit was trying to do, but this was presumably unintended.

### Expected behavior

Console logs of `AvmError`s output the actual error information, as they did before e2bc60f75e:
```
2026-05-10T20:06:26.113251Z  INFO run_frame:run_all_phases_avm2: avm_caught: Caught exception: TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: X)
        at MapPlayer/updatePosition()
        at Maps/nextMap()
        at Swisher/frame94()
```

### Content Location

N/A

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

_No response_

### Additional information

_No response_

Lord-McSweeney:
This should close #23687- @n-raine does this change work for you?

### DIFF
--- a/core/src/avm2/activation.rs
+++ b/core/src/avm2/activation.rs
@@ -924,7 +924,10 @@ impl<'a, 'gc> Activation<'a, 'gc> {
 
                 if matches {
                     #[cfg(feature = "avm_debug")]
-                    original_error.log(self, "Caught exception");
+                    {
+                        let stringified = original_error.to_string(self);
+                        tracing::warn!("Caught exception: {}", stringified);
+                    }
 
                     self.reset_stack();
                     self.push_stack(error);
--- a/core/src/avm2/activation.rs
+++ b/core/src/avm2/activation.rs
@@ -924,7 +924,7 @@ impl<'a, 'gc> Activation<'a, 'gc> {
 
                 if matches {
                     #[cfg(feature = "avm_debug")]
-                    tracing::info!(target: "avm_caught", "Caught exception: {:?}", original_error);
+                    original_error.log(self, "Caught exception");
 
                     self.reset_stack();
                     self.push_stack(error);
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,10 +75,7 @@ Additionally, if you build Ruffle with `--features avm_debug` then you will acti
 
 ### Logging caught exceptions
 
-Some SWFs may catch and suppress exceptions, which can hide the fact that the SWF is trying to use an unimplemented definition. To log call caught exceptions:
-
-1. Add `avm_caught=info` to your `RUST_LOG` environment variable (e.g. `RUST_LOG=warn,avm_caught=debug`)
-2. Build ruffle with `--features avm_debug`
+Some SWFs may catch and suppress exceptions, which can hide the fact that the SWF is trying to use an unimplemented definition. To log call caught exceptions, build ruffle with `--features avm_debug`.
 
 Caught exceptions will be logged as "Caught exception: <exception object>"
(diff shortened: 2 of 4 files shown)


core/src/avm2/activation.rs
```diff
@@ -924,7 +924,7 @@ impl<'a, 'gc> Activation<'a, 'gc> {
 
                 if matches {
                     #[cfg(feature = "avm_debug")]
-                    tracing::info!(target: "avm_caught", "Caught exception: {:?}", original_error);
+                    original_error.log(self, "Caught exception");
```
kjarosh: This will log such errors using `tracing::error!` which is not entirely accurate, it's not an error from Ruffle's POV.

core/src/avm2/activation.rs
```diff
@@ -924,7 +924,7 @@ impl<'a, 'gc> Activation<'a, 'gc> {
 
                 if matches {
                     #[cfg(feature = "avm_debug")]
```
kjarosh, line 926: Have you thought about making it a runtime feature that does not require Ruffle to be rebuilt?

core/src/avm2/activation.rs
Lord-McSweeney: I don't think someone custom-building Ruffle specifically for this feature will care very much about the log level. How would `Error::log` be made to log the error at different levels?

core/src/avm2/activation.rs
Lord-McSweeney: I don't think doing so is worth it; it's very rare that debugging requires logging caught errors. We've also found in the past that logging calls anywhere near the interpreter loop can slow the interpreter down.

kjarosh: I guess it makes sense considering it's about logging all caught errors, although I'm not convinced an additional boolean check would slow down the interpreter that much. If you look at the whole method, it's kinda heavy compared to an `if`.

core/src/avm2/activation.rs
kjarosh: > I don't think someone custom-building Ruffle specifically for this feature will care very much about the log level.

Maybe they won't, I don't know about those use cases enough to be sure. But I can imagine cases where: (1) I want to differentiate between caught errors and real errors, (2) I want to filter caught errors out.

> How would Error::log be made to log the error at different levels?

You'd have to use `tracing::event!` and pass the level as an argument `level: tracing::Level`.

core/src/avm2/activation.rs
Lord-McSweeney: It's not the boolean check- the compiler generates a decent amount of code for the logging machinery, which e.g. increases the stack size, forces more spilling, etc. It's still possible to get around it with some `#[inline(never)]` and `#[cold]` application, but at that point I'd rather have that change be in a separate PR.

kjarosh (approved):

Issue #23480, opened 2026-04-18
Error on https://armorgames.com/play/5344/primary


AndrewJVella:
# Error Info
Error name: Error
Error message: panicked at core/src/avm1/activation.rs:867:65:
called `Option::unwrap()` on a `None` value
Error stack:
```
Error: panicked at core/src/avm1/activation.rs:867:65:
called `Option::unwrap()` on a `None` value
    at __wbg_new_72b49615380db768 (https://cache.armorgames.com/ruffle/core.ruffle.574040544d43249e8cb7.js:1:49554)
    at ruffle_web.wasm.js_sys::Error::new::__wbg_new_72b49615380db768::h6a877a585481203f externref shim (wasm://wasm/ruffle_web.wasm-031d9b9a:wasm-function[12261]:0x94466b)
    at ruffle_web.wasm.ruffle_web::global_init::{{closure}}::h2a569114049d3574 (wasm://wasm/ruffle_web.wasm-031d9b9a:wasm-function[10261]:0x9195e7)
    at ruffle_web.wasm.std::panicking::panic_with_hook::h77afe0ddfda2cb89 (wasm://wasm/ruffle_web.wasm-031d9b9a:wasm-function[6847]:0x872b10)
    at ruffle_web.wasm.core::panicking::panic_fmt::h6651313c3e2c6c2f (wasm://wasm/ruffle_web.wasm-031d9b9a:wasm-function[10524]:0x91f036)
    at ruffle_web.wasm.core::panicking::panic::h0149fc8f1656305a (wasm://wasm/ruffle_web.wasm-031d9b9a:wasm-function[12250]:0x944576)
    at ruffle_web.wasm.core::option::unwrap_failed::h351295eb7223eb75 (wasm://wasm/ruffle_web.wasm-031d9b9a:wasm-function[12858]:0x9554b5)
    at ruffle_web.wasm.ruffle_core::avm1::activation::Activation::action_define_function::he79bb2a0d1719eda (wasm://wasm/ruffle_web.wasm-031d9b9a:wasm-function[2464]:0x55acac)
    at ruffle_web.wasm.ruffle_core::avm1::activation::Activation::run_actions::h3b526ce1d678259e (wasm://wasm/ruffle_web.wasm-031d9b9a:wasm-function[916]:0x18c76)
[...]

PR #23730, opened 2026-05-17
avm1: Introduce AVM1 playerglobal

Dinnerbone:
## Description

Now that we have an AVM1 compiler - we can finally move to a playerglobal for AVM1.

The only thing in it right now is `trace` as proof of concept, but this will be an iterative thing for sure. I figure let's have the groundwork in sooner rather than later, and anyone can pitch in on transitioning pieces to it.

## Testing

I added `trace` to it and you can see that it passes tests. If you add any other code there manually, you can see it work right before swfs are loaded.

## Checklist


tests/tests/swfs/avm1/global_proto_decls/output.ruffle.txt
```diff
@@ -1,3 +1,5 @@
+Testing _global.trace
+  __proto__, own, type=[object]
```
Lord-McSweeney: Why are `apply`, `call`, and `constructor` no longer traced?

Dinnerbone: `apply` and `call` actually shouldn't be traced.

For `constructor`; I think it's a bug with our `ASnative` impl. It doesn't set the prototype of the method, which means we never set its `constructor` value either.

core/src/avm1/globals.rs
```diff
@@ -648,7 +677,6 @@ pub fn create_globals<'gc>(
         "unescape" => method(UNESCAPE; DONT_ENUM);
         "parseInt" => method(PARSE_INT; DONT_ENUM);
         "parseFloat" => method(PARSE_FLOAT; DONT_ENUM);
-        "trace" => method(TRACE; DONT_ENUM);
```
moulins, line 651: Maybe a dummy `null` value should be left here, to preserve the proper ordering of globals?

kjarosh: +1, this way we'll keep the order during migration instead of shuffling it around each time

Dinnerbone: Added null, that makes the test output unchanged, great idea

core/src/avm1/activation.rs
```diff
@@ -865,8 +865,11 @@ impl<'a, 'gc> Activation<'a, 'gc> {
             action,
             self.scope(),
             constant_pool,
-            // `base_clip` should always be a living `MovieClip` so this can't fail
-            MovieClipReference::try_from_stage_object(self, bc).unwrap(),
+            // `base_clip` should always be a living `MovieClip` so this can't fail during "normal" execution
+            // However, the playerglobal doesn't have any movie clips, so we want to fall back to "something" for that
+            MovieClipReference::try_from_stage_object(self, bc).unwrap_or_else(|| {
+                MovieClipReference::from_path(self.gc(), WStr::from_units(b"_root"))
```
Lord-McSweeney: There have been some cases of this unwrap being hit (e.g. #23480); maybe it would be best to add an assertion here to preserve the old behavior and so that we don't continue with unexpected behavior on "normal" SWFs?

core/build.rs
```diff
@@ -1,14 +1,18 @@
...
+
+    build_playerglobal::build_avm1_playerglobal(repo_root, &out_dir)
+        .expect("Failed to build playerglobal_avm1");
+    build_playerglobal::build_avm2_playerglobal(repo_root, &out_dir, cfg!(feature = "known_stubs"))
+        .expect("Failed to build playerglobal_avm2");
 
-    // This is overly conservative - it will cause us to rebuild playerglobal.swf
-    // if *any* files in this directory change, not just .as files.
+    // This is overly conservative - it will cause us to rebuild playerglobals
+    // if *any* files in these directory change, not just .as files.
```
Lord-McSweeney: nit: "these directory" -> "these directories"

core/src/avm1/globals/globals.as
```diff
@@ -0,0 +1 @@
+trace("Hello world!");
```
kjarosh: Nit: looks like this was introduced in a wrong commit.

core/src/avm1/globals/globals.as
```diff
@@ -4,3 +4,5 @@
 // You should not reference any display objects here; this is not going to be loaded as a "real" movie.
 
 // For convenience, the use of `#include "foo.as"` is supported.
+
+trace = ASnative(100, 4);
```
kjarosh, line 8: Just curious: shouldn't we use `var`?

Dinnerbone: I think it doesn't matter either way as this happens to be the global scope and `var` therefor is the same as without var - however, I personally find it more clear to be explicit. Not totally sold though, if you disagree!

core/src/avm1/activation.rs
Dinnerbone: I'm not sure how you'd do that; a global "is playerglobal" boolean maybe?

core/src/avm1/scope.rs
```diff
@@ -199,10 +199,7 @@ impl<'gc> Scope<'gc> {
             // Traverse the scope chain in search of the value.
             scope.set(name, value, activation)
         } else {
-            // This probably shouldn't happen -- all AVM1 code runs in reference to some MovieClip,
-            // so we should always have a MovieClip scope.
-            // Define on the top-level scope.
-            debug_assert!(false, "Scope::set: No top-level movie clip scope");
```
kjarosh, line 205: FWIW I remember this assert being triggered for some content as well

kjarosh (approved):

Dinnerbone:
I've [restored the two panics](https://github.com/ruffle-rs/ruffle/compare/728370b33f75bd6ea53ad4303336f9e9d07f7668..29cc78a2c943ed198df7af9e892191403511d569) for non-globalscope swfs (which should be everything that isn't playerglobals itself)

core/src/avm1/activation.rs
Dinnerbone: Made it panic again

kjarosh (approved):

PR #23733, opened 2026-05-17
avm1: Fix behavior of `Sound` ownership

### LINKED ISSUE #22075: Issue when creating a new Sound() with an unloaded MovieClip

ChrisCPI:
## Description

Fixes #22075

Prior to this PR, the "ownership" of a `Sound` object (whatever MC is passed into its constructor being linked to the Sound) worked in this way:
- In the constructor, the owner is resolved one time and then stored forever.
- If the owner is not valid, then fall back to controlling the global sound.

However, this isn't very accurate to how it works in FP. This PR corrects (most of) the behavior, so that it now works in this way:
- Whatever was given for the owner in the constructor is coerced to a string and stored that way (named "`target`"). Every time the `owner()` is called from a method, it will try to resolve that path.
- A sound should only control the global sound if the owner given in the constructor is explicitly `Null` or `Undefined` to begin with.
- When a method/property is called, if the `owner()` can't be resolved, *and* the sound is not set to control global sound, then it will do nothing and/or return `Undefined`.

Not everything has been fixed, mainly there's some *weird* behavior happening with `loadSound`. However, properly fixing those will likely go out of scope from the goals of this PR.

## Testing

Added several new tests to hopefully cover most cases.

## Checklist

- [x] An LLM was involved in the authoring of this code. (small feedback from Gemini for implementing a couple things, otherwise 95% of the changes are still by me)


core/src/avm1/globals/sound.rs
```diff
@@ -359,16 +405,22 @@ fn constructor<'gc>(
...
+    // If `null` or `undefined` is provided, then the sound will be set to always
+    // control the global sound. Otherwise, it gets coerced to a string path and will
+    // try to be resolved every time a method needs it.
+    // `Sound.setTransform`, `Sound.stop`, etc. will affect all sounds owned by the target clip.
+
+    let use_global_sound = args
+        .get(0)
+        .is_none_or(|v| matches!(v, Value::Null | Value::Undefined));
+
+    let target = args
```
kjarosh, line 405: Looks like `use_global_sound` is `false` and `target` is `None` at the same time when the coercion fails. Have you tried failing the coercion and seeing if this case behaves differently than global sound?

ChrisCPI: Do you mean when the argument gets coerced to string? What would cause it to fail?

kjarosh: You can throw an exception in `toString`. I would guess it will be propagated and you wouldn't be able to construct a `Sound`.

core/src/avm1/globals/sound.rs
```diff
@@ -359,16 +405,23 @@ fn constructor<'gc>(
...
+        .get(0)
+        .filter(|_| !use_global_sound)
+        .map(|value| value.coerce_to_string(activation))
+        .transpose()?;
 
-    let sound = Sound::empty(activation.gc(), owner);
+    let sound = Sound::empty(activation.gc(), target, use_global_sound);
```
kjarosh: If I understand this code correctly, you can simplify it now by removing `use_global_sound` altogether. If `target` is `None`, you can assume that you're using the global sound.

ChrisCPI: Good call, that seems to be the case.

core/src/avm1/globals/sound.rs
```diff
@@ -126,8 +128,41 @@ impl<'gc> Sound<'gc> {
...
+    }
+
+    pub fn use_global_sound(self) -> bool {
+        self.0.target.is_none()
+    }
+
+    /// Returns `true` if this sound is attached to the global sound
+    /// (initial target was `Null` or `Undefined`), or if the target
+    /// *currently* resolves to a valid display object.
+    pub fn has_valid_owner(self, activation: &mut Activation<'_, 'gc>) -> bool {
```
kjarosh, line 149: Won't this always return `true` now?

kjarosh: Oh it won't because the owner can fail resolution!

kjarosh (approved):
LGTM, thank you!

Issue #9395, opened 2023-02-02
[AS3] URLRequest/URLLoader inconsistency of headers when compared to Flash


JordanPlayz158:
### Describe the bug

When Ruffle sends a Request from (responding to the SWFs URLLoader.load(URLRequest) call might be more accurate) the SWF. It seems the `Content-Type` header is missing and that header is important in some instances (like in the case with my Flash Server made in Laravel and Laravel automatically redirects (302) invalid POST requests (doesn't contain Content-Type)), there are more minor inaccuracies like the Accept header being more lax than flash (might not be an issue) and the Accept-Encoding available encoding formats is not the same (doubt it would cause much if any problems either but good to note I think).

Here is a comparison of an mitmproxy captured request (the headers specifically) and the network tab for the ruffle request:
![5378c148b4e0077b9a276801cccbe04f81fe2c5f](https://user-images.githubusercontent.com/22566618/216462797-8a73ac3f-d1d8-4c22-b9b4-5c0556c72e09.png)

### Expected behavior

The expected behavior is that the Ruffle headers should match the headers that flash uses (for best compatibility).

### Affected platform

Browser's extension

### Operating system

Crystal Linux 6.1.8-arch1-1

### Browser

Mozilla Firefox Browser Developer Edition 110.0b8

Lord-McSweeney:
We're also missing the `User-Agent` header (which CAN be changed on FF if I remember correctly) and `x-flash-version`.

n0samu:
The URLRequest contentType setter is unimplemented: https://github.com/ruffle-rs/ruffle/blob/5e115ed0385ee03c769470ff89e6dc2b4d39ccb9/core/src/avm2/globals/flash/net/URLRequest.as#L43

JordanPlayz158:
Ah, Well I will give the (decompiled) source that results in the `application/x-www-form-urlencoded`
```as3
         var _loc1_:URLRequest = new URLRequest();
         var _loc2_:Date = new Date();
         _loc1_.url = const_199 + "?Date=" + _loc2_.getTime();
         var _loc3_:URLVariables = new URLVariables();
         _loc3_.Action = "loadAccount";
         _loc3_.Email = this.email;
         _loc3_.Pass = this.pass;
         _loc3_.ver = class_878.const_184;
         _loc1_.method = URLRequestMethod.POST;
         _loc1_.data = _loc3_;
         this.var_41 = new URLLoader();
[...]

PR #23754, opened 2026-05-19
🤖 avm2: Send X-Flash-Version header from URLRequest (close #15276)

### LINKED ISSUE #15276: Ruffle Flash Player does not send `X-Flash-Version` in the request headers in `flash.display.Loader.load` function
### Describe the bug

I was experimenting with loading an image in ActionScript 3.0 and made my local server in TypeScript running in Deno

Then I came across `X-Flash-Version` header in the request in Adobe Flash Player's case and decided to experiment denying/locking flash player versions below 32, but Ruffle doesn't seem to have the header since it fails at `NaN` for me.

The swf I'm running is only a URL running on my local server `http://localhost:8000/experiments.swf`

image

### Expected behavior

It should send `X-Flash-Version` header with a string (e.g `32,0,0,371`) and render this image

image

### Content Location

[flash experiments.zip](https://github.com/ruffle-rs/ruffle/files/14335869/flash.experiments.zip)

Since this is meant to be ran on server-side, you should be able to run with this command line, just make sure you have Deno installed
`deno run --allow-read --allow-net index.ts`

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

_No response_

### Additional information

_No response_

MavenRain:
## Description

  Closes #15276.  Adds the X-Flash-Version HTTP header to every request built
  from an AS3 URLRequest, matching Flash Player's default behavior.  The header
  value is "{player_version},0,0,0" derived from
  activation.avm2().player_version, the same source used by
  flash.system.Capabilities.version.

  The header is set inside request_from_url_request
  (core/src/avm2/globals/flash/display/loader.rs), which is the shared AS3
  URLRequest builder consumed by Loader.load, URLLoader.load, and URLStream.
  Landing the fix at the shared builder is a single-line edit at the source and
  covers all three bindings uniformly, which matches Flash Player's per-request
  semantics (the header is a property of the HTTP request, not of the AS3 class
  that initiated it).

  X-Flash-Version is inserted before the existing user-headers loop, so a
  user-supplied X-Flash-Version entry continues to win under the existing
  last-write-wins semantics for URLRequest.requestHeaders.

  #15276 is the most-reported missing header; the umbrella issue #9395 tracks
  the broader header-parity work (Content-Type via the unimplemented
  URLRequest.contentType setter, User-Agent, and Accept / Accept-Encoding
  refinements).  I'm happy to follow up on those in subsequent PRs, or to broaden
  this one if reviewers prefer.

## Testing

```
  - cargo test --package tests url_loader (avm2/url_loader)
  - cargo test --package tests urlrequest (avm2/urlrequest)
  - cargo test --package tests loader (10 AVM1 + AVM2 loader tests)
  - cargo test --package tests netconnection (5 tests, including
  netconnection_send_remote which exercises log_fetch)
  ```

## Checklist

### CODE REVIEW

**Comment by MavenRain** [comment]
Updated tests/tests/swfs/avm2/loader_load/output.txt and tests/tests/swfs/avm2/loader_method/output.txt to include the new X-Flash-Version: 32,0,0,0 header line that the rule change now emits. These two were the only fixtures with log_fetch = true whose output.txt exercises a URLRequest path; the netconnection and AVM1 fixtures use different code paths and stay clean.

File: core/src/avm2/globals/flash/display/loader.rs [hunk 1]
```diff
@@ -148,7 +148,16 @@ pub fn request_from_url_request<'gc>(
         .get_slot(url_request_slots::_REQUEST_HEADERS)
         .as_object();
 
+    // Send `X-Flash-Version` by default, matching Flash Player.  Inserted before
+    // the user-headers loop so that a user-supplied entry wins per existing
+    // last-write-wins semantics.  Closes #15276; broader header parity tracked
+    // separately in #9395.
```
Dinnerbone, line 154: Can I please have a test that proves the claim of this comment? :D

Dinnerbone (changes requested):
The two commits should be squashed as the tests will be broken as of the first commit (which us if we need to bisect in the future)

MavenRain:
@Dinnerbone Both are now addressed.

I squashed the two commits into one, so the test suite stays green across the whole history and bisection is clean.

For the comment claim, I added `tests/tests/swfs/avm2/loader_x_flash_version_override/`.  The fixture pushes three `URLRequestHeader`s (`MyBefore`, `X-Flash-Version: 99,0,1,2`, `MyAfter`) and asserts the wire carries `X-Flash-Version: 99,0,1,2` at position 0 with `MyBefore` / `MyAfter` following in their relative order.  Position 0 only happens because the default is inserted before the user-headers loop, so the test also pins that ordering: drop the default insertion and `X-Flash-Version` would land at position 1, failing the test.  All 26 loader_* tests stay green.

Thanks for the review.

MavenRain:
@Dinnerbone friendly bump: both points from your review are addressed: the two commits are squashed into one, and I added tests/swfs/avm2/loader_x_flash_version_override/ asserting the wire carries X-Flash-Version at position 0 with user headers still winning. CI is green; the only red is a 6h "Test in chrome" runner timeout (firefox+edge pass on the same commit). Mind taking another look when you have a moment?

PR #23761, opened 2026-05-20
avm2: Refactor draw_triangles_internal

SuchAFuriousDeath:
## Description

This is a step towards rendering with uvt (which I'm trying to eventually implement) and this feels like a good time to submit this chunk as a separate PR before it gets too overwhelming. The behavioral changes should be minimal, the only added thing should be validation that indices come in triples (if not, an error is thrown, I adjusted the existing test adequately). I think the code is also much more readable now.

## Testing

_How can we test this PR?_

Everything is already included in the adjusted test.

## Checklist


kjarosh:
Damn, I had some triangles refactor that was improving performance and fixing winding behavior, which will conflict with it :/

It's no big deal, but probably can help in implementing UVT.

core/src/avm2/globals/flash/display/graphics.rs
```diff
@@ -1072,6 +1073,165 @@ pub fn draw_triangles<'gc>(
...
+    ///
+    /// Returns `Ok(None)` for inputs that produce no triangles (empty
+    /// vertices, empty indices, or all indices out of bounds).
+    fn new<'gc>(
+        activation: &mut Activation<'_, 'gc>,
+        vertices: &Object<'gc>,
+        indices: Option<&Object<'gc>>,
+        uvt_data: Option<&Object<'gc>>,
+    ) -> Result<Option<Self>, Error<'gc>> {
+        if uvt_data.is_some() {
```
Lord-McSweeney: Can we keep this mention of a stub in `draw_triangles_internal`? There can be a TODO here if you want, but IMO it's better to keep stubs closer to the actual method.

core/src/avm2/globals/flash/display/graphics.rs
```diff
@@ -1092,8 +1252,9 @@ impl TriangleCulling {
         }
     }
 
-    fn cull(self, (a, b, c): Triangle) -> bool {
-        fn triangle_orientation((a, b, c): Triangle) -> i64 {
+    #[inline]
+    fn cull(self, [a, b, c]: [Point<Twips>; 3]) -> bool {
```
Lord-McSweeney: In `cull` specifically (not `triangle_orientation`) is there a reason to have this function declare the triangle parameter as `[a, b, c]` when all three components are immediately put back together? Wouldn't it be simpler to have

```
    #[inline]
    fn cull(self, points: [Point<Twips>; 3]) -> bool {
```
and
```
            Self::Positive => triangle_orientation(points) >= 0,
            Self::Negative => triangle_orientation(points) <= 0,
```
?

core/src/avm2/globals/flash/display/graphics.rs
SuchAFuriousDeath: Actually I think they should be moved one level above this and that goes for both stubs, because there are 2 callers only one of them is of Graphics.drawTriangles.

Lord-McSweeney (approved):
LGTM but kjarosh should probably get a chance to review

SuchAFuriousDeath:
@kjarosh 🥺

core/src/avm2/globals/flash/display/graphics.rs
```diff
@@ -1072,6 +1091,155 @@ pub fn draw_triangles<'gc>(
...
+                .iter()
+                .map(|pair| coerce_point(pair, activation))
+                .collect::<Result<Vec<_>, Error<'gc>>>()?;
+
+            let num_vertices = vertices.len();
+
+            // Flash stops at the first out-of-bounds index.
+            let indices = index_triples
+                .iter()
+                .map_while(|[i0, i1, i2]| {
```
kjarosh: Can we make this simpler? I have a difficult time understanding what's going on.

SuchAFuriousDeath: Sure thing

SuchAFuriousDeath: It should be a bit better now 😇

kjarosh: I don't think it changes much, it just inlines `transpose()`? Have you tried splitting it into multiple stages? (e.g. first `map`, then `map_while`, etc., maybe `filter`?)

IIUC We first want to coerce it (which is fallible), then filter it (`i < num_vertices`), then take while? Or something like that.

SuchAFuriousDeath: :thinking:

SuchAFuriousDeath: How about this?

core/src/avm2/globals/flash/display/graphics.rs
```diff
@@ -1072,6 +1091,153 @@ pub fn draw_triangles<'gc>(
...
+                .iter()
+                .map(|pair| coerce_point(pair, activation))
+                .collect::<Result<Box<_>, Error<'gc>>>()?;
+
+            let num_vertices = vertices.len();
+
+            let indices = index_triples
+                .iter()
+                .map(|[i0, i1, i2]| {
+                    let tri = [
```
kjarosh: I think it's fine to inline here

core/src/avm2/globals/flash/display/graphics.rs
```diff
@@ -1072,6 +1091,153 @@ pub fn draw_triangles<'gc>(
...
+                        i1.coerce_to_u32(activation)?,
+                        i2.coerce_to_u32(activation)?,
+                    ];
+
+                    Ok(tri)
+                })
+                // Flash stops at the first out-of-bounds index.
+                .take_while(|result| match result {
+                    Ok(tri) => tri.iter().all(|&i| (i as usize) < num_vertices),
+                    Err(_) => true,
```
kjarosh: Can we cover this in tests? Maybe it's false? 🤔

SuchAFuriousDeath: I think this is actually unreachable, looking into it

kjarosh: I think it's reachable when the coercion fails, i.e.

```
var i = {};
i.valueOf = function() { throw "hehe"; };

var triangles = [1, 2, i];
```

I don't think this is tested.

kjarosh: Lol, that's actually AVM1, but in AVM2 you can do a similar thing if only a generic type is accepted. If not, then you shouldn't do coercions (but I think it is).

SuchAFuriousDeath: Ok I'm a bit confused here. You're saying that it's possible to push your 'i' variable into a Vec\<int\>?

kjarosh: Oh, you're right, I thought it might have been an array (for some reason), not a vector. I'm wondering if it's possible to somehow insert a wrong type into a vector, but probably not.

SuchAFuriousDeath: I think in several other places, Ruffle assumes it's not possible.

core/src/avm2/globals/flash/display/graphics.rs
```diff
@@ -1072,6 +1091,143 @@ pub fn draw_triangles<'gc>(
...
+        } else {
+            let vertex_triples = vertex_pairs
+                .as_chunks_exact::<3>()
+                .ok_or_else(|| make_error_2004(activation, Error2004Type::ArgumentError))?;
+
+            let triangles = vertex_triples
+                .iter()
+                .map(|[p0, p1, p2]| {
+                    Ok([
+                        coerce_point(p0, activation)?,
```
kjarosh: You can do `as_` here as well?

core/src/avm2/globals/flash/display/graphics.rs
```diff
@@ -1072,6 +1091,143 @@ pub fn draw_triangles<'gc>(
...
+                .as_chunks_exact::<3>()
+                .ok_or_else(|| make_error_2004(activation, Error2004Type::ArgumentError))?;
+
+            if index_triples.is_empty() {
+                return Ok(None);
+            }
+
+            let vertices = vertex_pairs
+                .iter()
+                .map(|pair| coerce_point(pair, activation))
```
kjarosh: And here?

core/src/avm2/globals/flash/display/graphics.rs
SuchAFuriousDeath: Good point (pun intended)

core/src/avm2/globals/flash/display/graphics.rs
```diff
@@ -1072,6 +1091,134 @@ pub fn draw_triangles<'gc>(
...
+                        vertices[i2 as usize],
+                    ]
+                }))
+            }
+            Self::Sequential { triangles } => Either::Right(triangles.iter().copied()),
+        }
+    }
+}
+
+fn coerce_point<'gc>([x, y]: &[Value<'gc>; 2]) -> Point<Twips> {
```
kjarosh: Nit: don't call it `coerce`, because it suggests we're coercing those values, but we are not.

kjarosh (approved):

core/src/avm2/globals/flash/display/graphics.rs
SuchAFuriousDeath: Done

PR #23762, opened 2026-05-20
chore: Clippy beta fixes

danielhjacobs:
## Description

Fix clippy nightly lints, done by running `cargo +nightly clippy --fix --all --all-features --tests -- -A clippy::byte_char_slices` first, then manually applying changes for byte_char_slices (which clippy fails to apply properly).

## Testing

- `rustup update`
- `cargo +beta clippy --all --all-features --tests`

## Checklist

### DIFF
--- a/core/src/avm1/object_reference.rs
+++ b/core/src/avm1/object_reference.rs
@@ -106,6 +106,7 @@ impl<'gc> MovieClipReference<'gc> {
         // In swfv5 paths resolve to the first MovieClip parent if the target isn't a movieclip
         if activation.swf_version() <= 5 {
             while display_object.as_movie_clip().is_none() {
+                #[allow(clippy::question_mark)]
                 if let Some(p) = display_object.avm1_parent() {
                     display_object = p;
                 } else {
--- a/core/src/html/layout.rs
+++ b/core/src/html/layout.rs
@@ -153,7 +153,7 @@ impl<'a, 'gc> LayoutContext<'a, 'gc> {
 
         let params = EvalParameters::from_span(span);
 
-        for text in span_text.split(&[b'\n', b'\r', b'\t'][..]) {
+        for text in span_text.split(&b"\n\r\t"[..]) {
             let slice_start = text.offset_in(span_text).unwrap();
             let delimiter = if slice_start > 0 {
                 span_text
--- a/core/src/input.rs
+++ b/core/src/input.rs
@@ -137,6 +137,7 @@ impl InputManager {
         let event = match event {
             // Optionally transform gamepad button events into key events.
             PlayerEvent::GamepadButtonDown { button } => {
+                #[allow(clippy::question_mark)]
                 if let Some(key_code) = self.gamepad_button_mapping.get(&button) {
                     InputEvent::KeyDown {
                         key_code: *key_code,
@@ -150,6 +151,7 @@ impl InputManager {
                 }
             }
             PlayerEvent::GamepadButtonUp { button } => {
+                #[allow(clippy::question_mark)]
                 if let Some(key_code) = self.gamepad_button_mapping.get(&button) {
                     InputEvent::KeyUp {
                         key_code: *key_code,
--- a/core/src/avm1/globals/movie_clip.rs
+++ b/core/src/avm1/globals/movie_clip.rs
@@ -970,6 +970,7 @@ pub fn clone_sprite<'gc>(
     }
 
     let movie = parent.movie();
+    #[allow(clippy::question_mark)]
     let cloned_sprite = if sprite.id() != 0 {
         // Clip from SWF; instantiate a new copy.
         let library = context.library.library_for_movie(movie).unwrap();
--- a/core/src/avm1/globals/point.rs
+++ b/core/src/avm1/globals/point.rs
@@ -191,7 +191,7 @@ fn distance<'gc>(
     let b = args.get(1).unwrap_or(&Value::Undefined);
     let delta = a.call_method(
         istr!("subtract"),
-        &[b.to_owned()],
+        std::slice::from_ref(b),
         activation,
         ExecutionReason::FunctionCall,
     )?;
--- a/core/src/avm2/events.rs
+++ b/core/src/avm2/events.rs
@@ -224,7 +224,7 @@ impl<'gc> DispatchList<'gc> {
         let new_handler = EventHandler::new(handler, use_capture);
 
         if let Some(event_sheaf) = self.get_event(event) {
-            for (_other_prio, other_set) in event_sheaf.iter() {
+            for other_set in event_sheaf.values() {
                 if other_set.contains(&new_handler) {
                     return;
                 }
@@ -247,7 +247,7 @@ impl<'gc> DispatchList<'gc> {
     ) {
         let old_handler = EventHandler::new(handler, use_capture);
 
-        for (_prio, set) in self.get_event_mut(event).iter_mut() {
+        for set in self.get_event_mut(event).values_mut() {
             if let Some(pos) = set.iter().position(|h| *h == old_handler) {
                 set.remove(pos);
             }
@@ -257,7 +257,7 @@ impl<'gc> DispatchList<'gc> {
     /// Determine if there are any event listeners in this dispatch list.
     pub fn has_event_listener(&self, event: AvmString<'gc>) -> bool {
         if let Some(event_sheaf) = self.get_event(event) {
-            for (_prio, set) in event_sheaf.iter() {
+            for set in event_sheaf.values() {
                 if !set.is_empty() {
                     return true;
                 }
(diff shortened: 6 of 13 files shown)


core/src/html/layout.rs
```diff
@@ -153,7 +153,7 @@ impl<'a, 'gc> LayoutContext<'a, 'gc> {
 
         let params = EvalParameters::from_span(span);
 
-        for text in span_text.split(&[b'\n', b'\r', b'\t'][..]) {
+        for text in span_text.split(&b"\n\r\t"[..]) {
```
danielhjacobs, line 156: If clippy tries to do this itself, it generates:

`for text in span_text.split(&*b"\n\r\t"[..]) {`

And errors with `type `[u8]` cannot be dereferenced`, similar applies to all these other changes

SuchAFuriousDeath (approved):

core/src/input.rs
```diff
@@ -137,29 +137,25 @@ impl InputManager {
...
+                {
+                    let key_code = self.gamepad_button_mapping.get(&button)?;
                     InputEvent::KeyDown {
                         key_code: *key_code,
                         key_char: None,
                         // TODO what location should we use here?
                         key_location: KeyLocation::Standard,
                     }
-                } else {
-                    // Just ignore this event.
```
kjarosh, line 148: Do not remove this comment

kjarosh:
Can we wait with these fixed until the linter hits beta? It's common that lints change in nightly.

core/src/html/layout.rs
SuchAFuriousDeath, line 156: Actually here (and in similar cases below, looks like it would be more convenient to just implement Pattern for [u8; N]? [std does it](https://doc.rust-lang.org/std/str/pattern/trait.Pattern.html#impl-Pattern-for-%5Bchar;+N%5D)

core/src/avm1/object_reference.rs
```diff
@@ -98,11 +98,9 @@ impl<'gc> MovieClipReference<'gc> {
         // In swfv5 paths resolve to the first MovieClip parent if the target isn't a movieclip
         if activation.swf_version() <= 5 {
             while display_object.as_movie_clip().is_none() {
-                if let Some(p) = display_object.avm1_parent() {
+                {
+                    let p = display_object.avm1_parent()?;
                     display_object = p;
-                } else {
-                    // Somehow we have gotten an object that has no MovieClip up the chain
```
torokati44, line 112: Another removed comment here.

core/src/input.rs
danielhjacobs: Where should I put it? clippy auto-removed it because it removed the else entirely.

core/src/avm1/object_reference.rs
danielhjacobs: Where should I put it? clippy auto-removed it because it removed the else entirely.

core/src/html/layout.rs
SuchAFuriousDeath: I volunteer to do it, if you guys approve :pleading_face:

torokati44:
The changes themselves LGTM, and I wouldn't mind either if we waited until these lints are in beta. It's still nice to have the adjustments ready! And then, I think the commits should be squashed before merging.

core/src/html/layout.rs
kjarosh: Please do it ❤️

core/src/avm1/object_reference.rs
kjarosh: IMO we shouldn't remove the `else`, it's not empty

core/src/input.rs
kjarosh: I would leave the else in this case.

kjarosh: Note that if a lint is not useful we should ignore it. If it's notoriously not useful we should ignore it in Cargo.toml

torokati44:
TBF I think at least the `.clear()` and `.values()` changes (and maybe even the `&b""` ones too) look nice, even if clippy wouldn't complain about them.

core/src/input.rs
danielhjacobs: This is the question_mark lint and has been around since pre 1.29.0, so I suspect it's just not that useful in these cases: https://rust-lang.github.io/rust-clippy/master/index.html#question_mark

danielhjacobs: By changing the above to use it, we do this same return early above instead of down here, but I guess you don't like that syntax, so I can revert it.

core/src/html/layout.rs
SuchAFuriousDeath: Ok I have it, but I'm wondering about something.

wstr.starts_with([b"a", b"b"]) gets corrected by clippy to wstr.starts_with(*b"ab").
[b"a", b"b"] feels intuitively like it should be 'any of' matching, but b"ab" feels like it should be 'does the wstr start with the whole chunk'. But they are the same thing :thinking: 

This is potentially really confusing.

core/src/input.rs
```diff
@@ -137,29 +137,25 @@ impl InputManager {
         let event = match event {
             // Optionally transform gamepad button events into key events.
             PlayerEvent::GamepadButtonDown { button } => {
-                if let Some(key_code) = self.gamepad_button_mapping.get(&button) {
+                {
+                    let key_code = self.gamepad_button_mapping.get(&button)?;
```
danielhjacobs: Alternative would be to add the comment here:

`// If the button is not in the gamepad mapping, return early to just ignore this event.`

core/src/input.rs
```diff
@@ -137,29 +137,25 @@ impl InputManager {
...
                     }
-                } else {
-                    // Just ignore this event.
-                    return None;
                 }
             }
             PlayerEvent::GamepadButtonUp { button } => {
-                if let Some(key_code) = self.gamepad_button_mapping.get(&button) {
+                {
+                    let key_code = self.gamepad_button_mapping.get(&button)?;
```
danielhjacobs: Same here:

`// If the button is not in the gamepad mapping, return early to just ignore this event.`

core/src/avm1/object_reference.rs
```diff
@@ -98,11 +98,9 @@ impl<'gc> MovieClipReference<'gc> {
         // In swfv5 paths resolve to the first MovieClip parent if the target isn't a movieclip
         if activation.swf_version() <= 5 {
             while display_object.as_movie_clip().is_none() {
-                if let Some(p) = display_object.avm1_parent() {
+                {
+                    let p = display_object.avm1_parent()?;
```
danielhjacobs: And here:

`// If we somehow have gotten an object that has no MovieClip up the chain, return early.`

core/src/html/layout.rs
SuchAFuriousDeath: One solution would be to keep the impl and just disable the byte_char_slices lint

kjarosh: Yeah, this sounds weird. Having the impl anyway SGTM

torokati44: just a random idea: we could rename it to `starts_with_any_char` or something (maybe not this verbose)

torokati44: oh sorry, in that case it would have to take an &str... but this is an "implicitly converted" regex-like thing, so that name doesn't work... maybe `matches_at_start`?

torokati44:
> Can we wait with these fixed until the linter hits beta? It's common that lints change in nightly.

There we go: https://github.com/ruffle-rs/ruffle/actions/runs/26491775623/job/78010919031

core/src/html/layout.rs
danielhjacobs: Should I allow this clippy error, or are the current changes okay?

core/src/avm1/object_reference.rs
```diff
@@ -106,6 +106,7 @@ impl<'gc> MovieClipReference<'gc> {
         // In swfv5 paths resolve to the first MovieClip parent if the target isn't a movieclip
         if activation.swf_version() <= 5 {
             while display_object.as_movie_clip().is_none() {
+                #[allow(clippy::question_mark)]
```
torokati44, line 109: Agh, we tend to prefer `expect` over `allow` (https://github.com/ruffle-rs/ruffle/pull/21081), but since this isn't marked by `stable`, that won't work... :/

danielhjacobs: Rust 1.96 only just released, so this isn't likely to become marked by `stable` until Rust 1.97, due on July 9

kjarosh (approved):

torokati44:
nit: the commit message still says "nightly"

torokati44:
> nit: the commit message still says "nightly"

Changed it to the PR title in a squash auto-merge.

PR #23768, opened 2026-05-20
desktop: Remove fontconfig from Windows and add fontconfig-dlopen 

kjarosh:
## Description

fontconfig is a unix dependency, while it can be theoretically installed on Windows, no one does that and it only takes space unnecessarily in the executable.

fontconfig-dlopen allows compiling Ruffle without fontconfig/dlopen which is usually desirable for downstream.

## Testing

Locally works without `dlopen`, release not tested.

## Checklist

### DIFF
--- a/desktop/src/backends/ui.rs
+++ b/desktop/src/backends/ui.rs
@@ -335,11 +335,14 @@ impl UiBackend for DesktopUiBackend {
         query: &FontQuery,
         register: &mut dyn FnMut(FontDefinition),
     ) -> Vec<FontQuery> {
-        #[cfg(feature = "fontconfig")]
-        return fontconfig_sort_device_fonts(query, register);
-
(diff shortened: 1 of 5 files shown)


desktop/src/backends/ui.rs
```diff
@@ -335,11 +335,14 @@ impl UiBackend for DesktopUiBackend {
         query: &FontQuery,
         register: &mut dyn FnMut(FontDefinition),
     ) -> Vec<FontQuery> {
-        #[cfg(feature = "fontconfig")]
-        return fontconfig_sort_device_fonts(query, register);
-
-        #[cfg(not(feature = "fontconfig"))]
-        return Vec::new();
+        cfg_select! {
```
SuchAFuriousDeath, line 338: Bonus points for using the fancy new macro 😌

SuchAFuriousDeath (approved):

PR #23772, opened 2026-05-21
🤖 avm2: Emit StrictArray for dense AS3 Arrays in AMF0 (close #16381)

MavenRain:
## Description

  Closes #16381.

  Per Lord-McSweeney's 2025-07-14 diagnosis, Ruffle was serializing AS3 Array
  arguments as AMF0 ECMAArray (marker 0x08, with all entries as named string
  keys "0", "1", ...) instead of AMF0 StrictArray (marker 0x0A, indexed by
  position).  Flash decoders see the former as {0: "x", 1: "y"} (associative
  object) and the latter as ["x", "y"] (indexed array), which broke real-world
  Flash Remoting endpoints that round-trip AS3 arrays as call arguments (the G2C
   Online site that motivated the issue).

  The fix unifies the AMF0 and AMF3 dense/sparse split in serialize_value
  (previously the AMF3 branch did this split, the AMF0 branch dumped everything
  into the sparse list with a // TODO: is this right? comment).  For AMF0,
  dense-only arrays now emit StrictArray; mixed arrays keep ECMAArray semantics.
    AMF3 path is unchanged.

  ## Byte-level confirmation

  For new Array("G2C Online") in AMF0:

  Before:  08 00 00 00 01 00 01 30 02 00 0A G2C Online 00 00 09   (ECMAArray
  with sparse "0" key)
  After:   0A 00 00 00 01 02 00 0A G2C Online                     (StrictArray,
  length 1)
  Flash:   StrictArray(ObjectId(-1), [String("G2C Online")])      (matches
  After)

  ## Test coverage

  The existing netconnection_send_remote test exercises the AMF0 args-wrap path
  (top-level StrictArray of arguments), which was already correct before this PR
   and stays correct after.  All 5 netconnection tests plus the broader
  amf/array/bytearray suites (~120 tests) stay green.

  What we don't yet exercise is an AS3 Array passed AS an argument to
  NetConnection.call, which is the bug case.  Adding a SWF fixture requires Flex
   SDK to recompile Test.swf, which I don't have set up locally.  Happy to
  follow up with a Rust unit test in a separate PR if you'd like dedicated
  coverage; the heaviest piece there is mocking an Activation for
  serialize_value.

  ## Verification

```
  - cargo fmt --all --check
  - cargo clippy -p ruffle_core --tests --lib
  - cargo test --package tests for netconnection_*, amf*, array, bytearray (all
  green)
```

PR #23775, opened 2026-05-21
avm1/avm2: Fix serialization of arrays in NetConnection

### LINKED ISSUE #16381: Genes to Cognition Online
### Describe the bug

- Clickable circles don't load next to the label "Selected Items"
- The connected graph of circular nodes does not appear
- While clicking the Label ADHD from the home page at least loads the main content area, going directly to the link with ADHD in the URL does not do so

Screenshot from 2024-05-20 10-19-45

### Expected behavior

- There should be clickable circles next to the label "Selected Items"
- There should be connected graph of clickable circular nodes
- The main content are should load when using a URL parameter as well

Screenshot from 2024-05-20 10-18-23

### Content Location

http://www.g2conline.org/#ADHD

### Affected platform

Browser's extension

### Operating system

Fedora Linux 39

### Browser

Mozilla Firefox 126

### Additional information

_No response_

danielhjacobs:
## Description

Gemini's explanation that seems to check out: "Because serialize is shared by both NetConnection and SharedObject, it takes a shortcut: it passes the root object into new_lso (which is designed for SharedObject saves) and then forcefully wraps the result in an AmfValue::Object.

Because of this, any array passed as a top-level argument to NetConnection gets stripped of its Array identity and turned into a dictionary of key-value pairs (e.g., {"0": "G2C Online"}). Nested arrays worked fine because recursive_serialize handles them correctly, but the root argument was being destroyed."

Fixes #16381

## Testing

Build this PR and use it on http://www.g2conline.org/ and you'll see the bubbles and the correct request body

## Checklist


danielhjacobs:
MavenRain has done a fix in #23772, less targeted but possibly more correct. Since both fix this particular issue, that needs tests to see if the other changes are needed. If they are, MavenRain's PR is probably better. If not, this PR may be better. We have to find out.

core/src/avm1/globals/shared_object.rs
```diff
@@ -86,6 +87,26 @@ pub fn serialize<'gc>(activation: &mut Activation<'_, 'gc>, value: Value<'gc>) -
         Value::Number(number) => AmfValue::Number(number),
         Value::String(string) => AmfValue::String(string.to_string()),
         Value::Object(object) => {
+            // Check if the root object being serialized is an Array.
+            // If it is, we bypass the LSO writer so it stays a StrictArray instead of becoming an Object.
+            if let NativeObject::Array(_) = object.native() {
```
kjarosh, line 90: I'm strongly in favor of adding tests here, my intuition says that FP checks for the `length` property instead of for the native object. That's how it behaves in other places where we serialize stuff.

danielhjacobs: Seems to be checking for actual array in this case, just added a test.

tests/tests/swfs/avm1/netconnection_serialize_arrays/test.as
```diff
@@ -0,0 +1,23 @@
+var nc = new NetConnection();
+nc.connect("http://localhost:8000/");
+
+// Callback function
```
danielhjacobs, line 4: While formulating the test, discovered a pre-existing Ruffle bug outside the scope of this PR. When you pass null as the callback handler to NetConnection, Flash Player sends just "/". But Ruffle currently has a bug where it always auto-increments and appends the call ID, sending "/1", "/2", etc., regardless of whether a callback exists. Discovered because Gemini's initial suggestion for the test included using null as the callback handler and the returned bytes were then inaccurate compared to the Flash output, so I added a minimal callback function to avoid the bug

tests/tests/swfs/avm1/netconnection_serialize_arrays/test.as
```diff
@@ -0,0 +1,23 @@
...
+var responder = new Object();
+responder.onResult = function(result) {
+    trace("Received result");
+};
+
+// 1. Genuine Array
+var realArray = new Array();
+realArray[0] = "real_0";
+realArray[1] = "real_1";
+
```
Lord-McSweeney, line 14: What happens if there are non-array-element properties stored in the array? Do they get skipped?

core/src/avm1/globals/shared_object.rs
```diff
@@ -86,6 +87,26 @@ pub fn serialize<'gc>(activation: &mut Activation<'_, 'gc>, value: Value<'gc>) -
...
+                let length = object.length(activation).unwrap_or(0);
+                let mut elements = Vec::with_capacity(length as usize);
+
+                for i in 0..length {
+                    let elem_name =
+                        AvmString::new_utf8(activation.context.gc_context, i.to_string());
```
Lord-McSweeney: nit: Does `activation.gc()` work here?

tests/tests/swfs/avm1/netconnection_serialize_arrays/test.as
danielhjacobs: Very good call, it gets serialized as an ECMA Array instead of a Strict Array (https://en.wikipedia.org/wiki/Action_Message_Format#AMF0), see new code.

PR #23776, opened 2026-05-21
webgl: Port to glow so it can run on native platforms

Fancy2209:
## Description

_Please explain the changes made within this PR.
Why did you make this change; does it add a feature, or fix a bug?
Be sure to link to any content or issues that it affects!
Please also remember that each commit message should describe its own changes too._

This ports the WebGL renderer to glow, allowing it to eventually be usable on Desktop (which is useful due to it's smaller requirements) and more importantly the Test Harness, allowing us to test WebGL without having to make an in browser test runner.

## Testing

_How can we test this PR?_

_If you're able to, please follow our [test guidelines](https://github.com/ruffle-rs/ruffle/blob/master/CONTRIBUTING.md#test-guidelines) to make a regression test which should fail before this PR, and pass after it. This helps us protect against breaking your changes in the future. These are highly encouraged and will speed up reviewing of your pull request. Some examples of these are SWF regression tests, unit tests, or web integration tests. Describe what you added!_

_If you need help with making tests, or do not believe this change is easily automatically testable, please describe how we can verify the changes ourselves. If there is real world content that we can test on, that helps a lot!_

I wasn't really able to figure out an optimal way to get an headless gl context so I could make a glow context to allow imgtests to run with it yet, so if this is desired to be part of this PR, i'd need help with it.

## Checklist


Fancy2209:
I realized this branch currently has fallen out of date with master, I will do a clean report with a less messy git tree.

Issue #2175, opened 2020-12-31
bug


D10G00800:
I really wanted to play this game because it marked my childhood :/ but he doesn't even start...

torokati44:
You forgot to mention... *which* game you are talking about?

PR #23783, opened 2026-05-22
🤖 avm2: Implement flash.text.engine TextLine and TextBlock

doctorpangloss:
Implements flash.text.engine's TextBlock and TextLine.

createTextLine and recreateTextLine shape a line through Ruffle's text layout code and attach it to a new FteTextLine display object. The TextLine natives read metrics and atoms back from it: the atom model (getAtomBounds, getAtomCenter, getAtomIndexAtPoint, bidi level, word boundaries) and the nextLine/previousLine chain. ElementFormat color, size, tracking, kerning, baseline shift, typographic case and weight/posture are handled, including across nested GroupElement runs.

Device-font metrics match Flash Player. The base advance is pixel-snapped, with kerning and tracking applied sub-pixel on top, and pair kerning split half-and-half between the two atoms. Kerning is taken from the font's GPOS table rather than the legacy kern table, because the two can disagree (Liberation Sans kerns "AV" more tightly in the old kern table than in GPOS) and GPOS is what Flash Player and modern shapers use. The kern table stays as a fallback for fonts that ship no GPOS kerning.

createTextLine throws Error #2175 when a content element has a null ElementFormat, and returns null for an empty TextBlock. Both were checked against the Flash Player debugger.

There are 21 test SWFs under tests/tests/swfs/avm2 (textline_* and textblock_create_text_line). Every one is pinned to a trace captured from Flash Player and to an expected output image that was also rendered by Flash Player; the harness compares Ruffle's render against that Flash Player image. The layout tests draw their TextLines, and the behavior tests draw a one-line description with a self-checked PASS or FAIL.

Inline graphics, vertical writing mode and text rotation, tab stops, and justification are not implemented. The TextLine members for those return values that are correct for horizontal LTR text, as the commit message spells out.


kjarosh:
But... but... I chose the blue pill :sob:

torokati44:
Yes, please break this up into several small, reasonably scoped, individually reviewable commits, that build on top of each other. Like how all the other PRs that actually get merged are done,

PR #23789, opened 2026-05-23
core/render: Add support for loading JPEGXR images

ChrisCPI:
## Description

Fixes #23417 

Both AVM1 and AVM2 allow for loading and displaying JPEGXR images through the normal loader methods. The decoding logic was mostly taken from `core/src/avm2/globals/flash/display3D/textures/atf_jpegxr.rs`, `jpegxr_to_tiff()`.

It isn't perfect in all cases in comparison to Flash Player, but it should hopefully be good enough to cover most (if not all) real world cases.

Only remaining question: Is throwing this under the same methods as the define image tags the best idea? Flash Player doesn't seem to recognize putting JPEGXR data in there, but the core loader logic uses this, so...

## Testing

Added the `avm2/loader_jpegxr` test to cover this.

## Checklist


render/Cargo.toml
```diff
@@ -33,6 +34,7 @@ byteorder = { workspace = true }
 wgpu = { workspace = true, optional = true }
 indexmap = { workspace = true }
 naga-agal = { path = "naga-agal" }
+jpegxr = { git = "https://github.com/ruffle-rs/jpegxr", rev = "5281b4ae42be742779269a9f1a986101f101f32f", optional = true }
```
kjarosh: This dep will need to be deduplicated, you can move it to workspace dependencies and just do `jpegxr = { workspace = true, optional = true }`

render/src/utils.rs
```diff
@@ -41,6 +43,7 @@ pub fn decode_define_bits_jpeg(
     }
     match format {
         JpegTagFormat::Jpeg => decode_jpeg(data, alpha_data),
+        JpegTagFormat::JpegXr => decode_jpegxr(data),
```
kjarosh, line 46: If `decode_jpegxr` is gated behind a feature, its call should be too, I'm not sure why this passes checks 🤔

SuchAFuriousDeath: Didn't the jpegxr feature exist only because of the build failures, which are resolved now?

torokati44: I think it's nice to keep around still. We have feature flags for the video codecs too, and jpegxr decoder is more of an oddball (C), and not that commonly used IMHO.

ChrisCPI: There's two variants of the function: one where JPEGXR is enabled, and one where it is disabled (returns an Err).

render/Cargo.toml
ChrisCPI: I moved it under `[workspace.dependencies]` in the root Cargo.toml, but it gave this error:

`jpegxr is optional, but workspace dependencies cannot be optional`

kjarosh: You need to make the workspace dep non optional, and move optional to uses of the workspace dep, e.g. `jpegxr = { workspace = true, optional = true }`

render/src/utils.rs
```diff
@@ -461,6 +465,70 @@ fn decode_gif(data: &[u8]) -> Result<Bitmap<'static>, Error> {
     ))
 }
 
+#[cfg(feature = "jpegxr")]
+fn decode_jpegxr(data: &[u8]) -> Result<Bitmap<'static>, Error> {
+    // This logic is mostly taken from 'core/src/avm2/globals/flash/display3D/textures/atf_jpegxr', jpegxr_to_tiff
+
+    let mut decoder = jpegxr::ImageDecode::with_reader(Cursor::new(data))
+        .expect("Failed to decode JPEG-XR image");
```
kjarosh: Instead of panicking, I think we should just return an error. Or can't we? (Similarly below.)

render/src/utils.rs
```diff
@@ -461,6 +465,70 @@ fn decode_gif(data: &[u8]) -> Result<Bitmap<'static>, Error> {
...
+    let info = jpegxr::PixelInfo::from_format(pixel_format);
+    let stride = jpeg_width as usize * info.bits_per_pixel() / 8;
+    let size = stride * jpeg_height as usize;
+
+    // We convert the result to a TIFF - this makes the jpegxr library handle
+    // all of the weird JPEG-XR alpha formats for us. We can then use the normal
+    // `image` crate to decode the TIFF to an rgba array.
+    let mut bmp_buffer = vec![0; size];
+    decoder
+        .convert_to_tiff(&mut Cursor::new(&mut bmp_buffer))
```
kjarosh, line 490: Have you looked into `jpegxr_to_raw_pixels`? Do we have to convert it to tiff first so that we can get rgba? Maybe it's possible to skip this step.

render/src/utils.rs
ChrisCPI: Possibly, I just copied it from the aforementioned function and figured it was also good, unless there's a valid reason for that other function to panic that doesn't apply here?

kjarosh: That function does not return a Result (for some reason). We are, so we have this possibility to error out.

render/src/utils.rs
ChrisCPI: Tried to implement it with the following code:
```rs
fn decode_jpegxr(bytes: &[u8]) -> Result<Bitmap<'static>, Error> {
    let mut decoder = jpegxr::ImageDecode::with_reader(Cursor::new(bytes))
        .expect("Failed to decode JPEG-XR image");

    let pixel_format = decoder
        .get_pixel_format()
        .expect("Failed to get pixel format");
    let (jpeg_width, jpeg_height) = decoder.get_size().expect("Failed to get JPEG-XR size");

    let info = jpegxr::PixelInfo::from_format(pixel_format);
    let stride = jpeg_width as usize * info.bits_per_pixel() / 8;

    let size = stride * jpeg_height as usize;

    let mut data = vec![0; size];

    decoder
        .copy_all(&mut data, stride)
        .expect("Failed to decode");

    Ok(Bitmap::new(
        jpeg_width as u32,
        jpeg_height as u32,
        BitmapFormat::Rgba,
        data,
    ))
}
```
But it didn't seem to like that:

output actual-macos-Metal

render/src/utils.rs
ChrisCPI: Changed all the spots with `.expect()` now return a new enum of Error (`InvalidJpegXr`). I only left the image_reader decoding the TIFF with an .expect(), simply because I'm not sure what error to return there.

render/src/utils.rs
kjarosh: Hmmm, that's probably because you're ignoring pixel format and assuming RGBA. But looking at the code, supporting that would be really difficult, so better to convert it to tiff.

render/src/utils.rs
```diff
@@ -461,6 +465,68 @@ fn decode_gif(data: &[u8]) -> Result<Bitmap<'static>, Error> {
     ))
 }
 
+#[cfg(feature = "jpegxr")]
+fn decode_jpegxr(data: &[u8]) -> Result<Bitmap<'static>, Error> {
+    // This logic is mostly taken from 'core/src/avm2/globals/flash/display3D/textures/atf_jpegxr', jpegxr_to_tiff
```
kjarosh: That's fine to reference that in the commit message, but it will get outdated in code.

render/src/error.rs
```diff
@@ -18,6 +18,10 @@ pub enum Error {
     #[error("Invalid JPEG")]
     InvalidJpeg(#[from] jpeg_decoder::Error),
 
+    #[cfg(feature = "jpegxr")]
+    #[error("Invalid JPEG-XR")]
```
kjarosh: Can you display the error?


```suggestion
    #[error("Invalid JPEG-XR: {0}")]
```

render/Cargo.toml
```diff
@@ -17,6 +17,7 @@ pixel_bender = { path = "pixel_bender"}
 tracing = { workspace = true }
 gif = "0.14.2"
 png = "0.18.1"
+image = { workspace = true, features = ["tiff"] }
```
kjarosh: The `tiff` feature should probably be added by `jpegxr`

render/src/utils.rs
kjarosh: If you feel like it's important it's better to add a TODO to unify the code with 'core/src/avm2/globals/flash/display3D/textures/atf_jpegxr'.

render/src/utils.rs
```diff
@@ -461,6 +465,68 @@ fn decode_gif(data: &[u8]) -> Result<Bitmap<'static>, Error> {
...
+    decoder
+        .convert_to_tiff(&mut Cursor::new(&mut bmp_buffer))
+        .map_err(Error::InvalidJpegXr)?;
+
+    let image_reader =
+        image::ImageReader::with_format(Cursor::new(bmp_buffer), image::ImageFormat::Tiff);
+
+    let data = image_reader
+        .decode()
+        .expect("Failed to decode TIFF bitmap")
```
kjarosh: Here you can return an error as well instead of panicking. If you don't expect it failing, you can add a debug_assert

tests/tests/swfs/avm2/loader_jpegxr/test.toml
```diff
@@ -0,0 +1,11 @@
+num_frames = 3
+
+[image_comparisons.output]
+tolerance = 48
+max_outliers = 1500
```
kjarosh: This test has a large tolerance and outliers due to one of the images. I think it would be better to split this into 2 tests: one with that problematic image, the second without. Otherwise we have no idea where the discrepancy is (and code could break jpegxr rendering without us noticing).

render/src/utils.rs
ChrisCPI: Refer to my comment in your earlier review:
> I only left the image_reader decoding the TIFF with an .expect(), simply because I'm not sure what error to return there.

As in, which enum of `Error` would be best to return? Unless I make a new one for `ImageError` (which is what .decode() returns when it fails)?

render/Cargo.toml
ChrisCPI: Do you mean like this?

`jpegxr = { workspace = true, optional = true, features = ["tiff"] }`

When it's like this it gives this error:

```
package `ruffle_render` depends on `jpegxr` with feature `tiff` but `jpegxr` does not have that feature.
```

kjarosh: You need to specify it in the `jpegxr` feature:

```
jpegxr = [..., "image/tiff"]
```

render/src/utils.rs
kjarosh: You can add a new one, but make sure it's jpegxr specific (as we are not sure the tiff is not malformed).

tests/tests/swfs/avm2/loader_jpegxr/test.toml
```diff
@@ -0,0 +1,12 @@
+# logo.jxr generated with ImageMagick:
+# magick logo.svg -define jxr logo.jxr
+#
+# gradient.jxr generated with gradient.png via https://www.aconvert.com/image/png-to-jxr/
+
+num_frames = 3
```
kjarosh, line 6: Looks like the image assertion is missing here.

kjarosh (approved):

kjarosh:
Thank you!

PR #23794, opened 2026-05-23
tests: Add asc compiler for compiling AVM2 tests

kjarosh:
## Description

Leverage the existing asc.jar from our repo (originally used to compile playerglobal) to compile AVM2 tests too.

## Testing

Tested by creating a sample AVM2 test.

## Checklist


Dinnerbone:
Super cool! Tests LGTM but can I see a test converted to use it, or an example test that uses it, to play with it locally?

kjarosh:
Sure, I thought about converting existing tests, but I noticed that `asc` compiles stuff differently compared to e.g. `mxmlc`, e.g. different multinames, etc.

I'll try picking some test that doesn't care about this stuff and converting it.

kjarosh:
Done.

The one thing I'm worried about is that when playerglobal changes, the output for tests could change too, e.g. if we used a wrong type in playerglobal and then fixed it :/

Dinnerbone (approved):

danielhjacobs (approved):
Less from the code perspective, more from the perspective of I've already git switched to this branch to compile AVM2 tests.

Issue #20220, opened 2025-04-18
Incorrect sprite rendering in "Flash Toukou Matter"


FioraAeterna:
### Describe the bug

If you click through the first few lines of character interaction, you will start to see incorrect sprite rendering like this:

<img width="1065" alt="Image" src="https://github.com/user-attachments/assets/2e9cac3b-b9fa-4557-96a8-f6ff324ea8a8" />

### Expected behavior

It should look like this, which it does in original Adobe Flash:

<img width="1062" alt="Image" src="https://github.com/user-attachments/assets/58664510-229a-431c-95c0-7eb5263e35da" />

### Content Location

https://exapico.freetzi.com/TOUKOUMATTER/yume.html

### Affected platform

Desktop app

### Operating system

Mac OS Sonoma 14.7

### Browser

Mozilla Firefox 137

torokati44:
Looks like for the faces to appear normal, they have to be mirrored.
And they are, initially:

![Image](https://github.com/user-attachments/assets/ce334a7a-2c0f-4ed6-bf5e-2516ab5dec6d)

![Image](https://github.com/user-attachments/assets/900911df-6ca4-4ef6-b708-627374908295)

But then, they lose this inverted World Matrix for some reason:

![Image](https://github.com/user-attachments/assets/c70b6b37-8f96-45f1-afd6-f8c4f6165a07)

torokati44:
Looks like this is an issue related to `DisplayObjectFlags::SCALE_ROTATION_CACHED` - specifically, `DisplayObjectBase::cache_scale_rotation` not preserving negative scales.

torokati44:
This (and the issue fixed by it) might be related: https://github.com/ruffle-rs/ruffle/pull/7788

torokati44:
Hopefully this will be fixed in tomorrow's nightly.

PR #23796, opened 2026-05-23
tests: Add a failing visual test for scale_rotation_cache

torokati44:
## Description

This is meant to be a close to minimal reproducer for https://github.com/ruffle-rs/ruffle/issues/20220.

## Testing

Well, it _is_ a test.
But either run the SWF directly in FP/Ruffle, or ... run the test case. :sweat_smile:

## Checklist


kjarosh:
> Well, it _is_ a test.

If it's a test I think you're supposed to describe how you tested that this test tests what it's supposed to test... :thinking:

kjarosh (approved):

torokati44:
Well ... I tested it by testing what the test tests!

torokati44:
I think if I replace the loaded PNG with an attached BitmapData, the bug goes away... 😵

torokati44:
So, basically:
I think the comments in https://github.com/ruffle-rs/ruffle/blob/23f80bf33f792f6366a743aa321c113b195b7c01/core/src/display_object.rs#L470-L501 probably implies that we should always consider the matrix as the single source of truth, whatever the weird decomposition into xscale/yscale/rotation/(skew) will tell us.

But then, in https://github.com/ruffle-rs/ruffle/blob/23f80bf33f792f6366a743aa321c113b195b7c01/core/src/display_object.rs#L549-L577 we probably shouldn't set the property value directly, because the decomposition from the matrix might result in something different? Instead, only the matrix should be manipulated somehow and let that caching code figure out the property values somehow...??

Or maybe that comment about "surprising results" apllies only to stuff done in the Flash authoring software, not to properties set by AS.
But then the decomposition of the matrix should be fixed to produce less weird results, at least in this case.

torokati44:
Now my latest (AI-assisted) hypothesis is that the `DisplayObjectFlags::SCALE_ROTATION_CACHED` tag is wrongly set/cleared when the MC is replaced by the loaded image - because it doesn't affect a BitmapData MC.

torokati44:
@kjarosh please re-review and merge if you want.

tests/tests/swfs/visual/scale_rotation_cache/test.as
```diff
@@ -0,0 +1,61 @@
+import flash.display.BitmapData;
+
+this.createEmptyMovieClip("img_mc", this.getNextHighestDepth());
+trace("img_mc SWF version:");
+trace(img_mc.getSWFVersion());
+var mcl:MovieClipLoader = new MovieClipLoader();
+mcl.loadClip("dots.png", img_mc);
```
torokati44, line 7: I wonder if this could ever be flaky if the image loading happens unusually quickly, but perhaps we have something explicit in place to prevent that.

torokati44: (I meant the next two lines not this one, sorry, misclick)

Issue #23757, opened 2026-05-19
Scroll rect is not respected during mouse pick


n-raine:
### Describe the bug

The various `avmX_mouse_pick()` implementations do not bail out if the mouse location is outside the object's scroll rect. This is the cause of #23576, #13737, and #23678 at the least.

### Expected behavior

Scroll rect clipping is respected and mouse events do not trigger outside the bounds of an object.

### Content Location

https://www.kongregate.com/en/games/dustinaux/the-enchanted-cave

### Affected platform

Desktop app

### Operating system

Windows 11 (but it doesn't matter)

### Browser

Mozilla Firefox 150.0.3 (again, doesn't matter)

Issue #23576, opened 2026-04-28
Enchanted Cave 1 gets buggy when trying to select items in the shop, as if the cursor is elsewhere


Georgedude:
### Describe the bug

As titled, game menu works fine then the problem begins once a game is started 

### Expected behavior

The cursor should click the items it hovers over

### Content Location

Flashpoint

### Affected platform

Desktop app

### Operating system

Windows 11

Issue #13737, opened 2023-10-27
The Enchanted Cave 2, wrong cursor position in inventory


SN902:
### Describe the bug

[TheEnchantedCave2.zip](https://github.com/ruffle-rs/ruffle/files/13187050/TheEnchantedCave2.zip) is almost playable, but one issue still prevents me from completing the game completely, difficult to click on any items in the inventory to equip new armor, rings and weapons. 

![image3](https://github.com/ruffle-rs/ruffle/assets/38821724/3273e69d-b685-4699-be46-f0941f27da77)

### Expected behavior

everything in inventory should be equipped normally.

### Affected platform

Desktop app

### Operating system

Windows 10 x64

### Browser

Firefox Nightly 121.0a1

Lord-McSweeney:
This also affects the original version of the game, at least the one on Kongregate.

PR #23797, opened 2026-05-23
core: Respect scrollRect when picking AVM1/AVM2 children (close #23678)

### LINKED ISSUE #23678: Mouse event propagation does not respect clipping in Spark containers
### Describe the bug

The bug manifests itself with Spark containers (at least VGroup and BorderContainer) and consists in the fact that mouse events are also propagated to components that are not visible due to clipping of the container edges and which therefore should not receive the events.

The attached short video clearly shows how the button responds to events despite being partially out of viewport due to clipping of its containers.

https://github.com/user-attachments/assets/51965674-6f36-4f8d-87b0-4e3d374b5afe

### Expected behavior

The expected behavior is that of the normal flash player, that is, components do not receive mouse events if they are outside the clipping area.

### Content Location

I've created a minimalist application that displays three nested VGroups, with a button inside. The center VGroup clips overflowing content using the "clipAndEnableScrolling" attribute. The button shouldn't receive mouse events on its covered parts when the browser screen is resized.

[application.zip](https://github.com/user-attachments/files/27558370/application.zip)

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

Firefox 159

### Additional information

_No response_

falpi:
## Description

Fixes https://github.com/ruffle-rs/ruffle/issues/23678

Cause
mouse_pick_avm2 / mouse_pick_avm1 on MovieClip and mouse_pick_avm2 on LoaderDisplay iterate over their children without ever consulting the container's own scrollRect. Rendering applies the rect as a stencil mask (see apply_standard_mask_and_scroll in core/src/display_object.rs), but the hit-testing path does not, so a child whose visual position falls outside the clipped viewport still receives mouse events. This breaks Flex/Spark containers that rely on clipAndEnableScrolling (e.g. VGroup, BorderContainer), where buttons remain clickable on portions hidden by the container edges.

Fix
Before recursing into children, if self.scroll_rect() is set, transform the incoming global point into the container's local coordinate space (local_matrix * point) and check it against the rect. If it is outside, return Avm2MousePick::Miss / None immediately. This is applied to:

MovieClip::mouse_pick_avm2 (core/src/display_object/movie_clip.rs)
MovieClip::mouse_pick_avm1 (core/src/display_object/movie_clip.rs)
LoaderDisplay::mouse_pick_avm2 (core/src/display_object/loader_display.rs)
Because the check runs at every recursion level, nested scrollRects are handled correctly. No changes to rendering or to the public API.

## Testing
See issue #23678 for testing application

## Checklist

### DIFF
--- a/core/src/display_object/loader_display.rs
+++ b/core/src/display_object/loader_display.rs
@@ -178,6 +178,14 @@ impl<'gc> TInteractiveObject<'gc> for LoaderDisplay<'gc> {
         options.set(HitTestOptions::SKIP_MASK, self.maskee().is_none());
 
         if self.visible() {
+            // A scrollRect clips mouse events at the visible viewport, just like rendering.
+            if let Some(scroll_rect) = self.scroll_rect()
+                && let Some(local_matrix) = self.global_to_local_matrix()
+                && !scroll_rect.contains(local_matrix * point)
+            {
+                return Avm2MousePick::Miss;
+            }
+
             // A loader has at most one child.
             if let Some(child) = self.iter_render_list().next() {
                 if let Some(int) = child.as_interactive() {
--- a/core/src/display_object/loader_display.rs
+++ b/core/src/display_object/loader_display.rs
@@ -174,6 +174,14 @@ impl<'gc> TInteractiveObject<'gc> for LoaderDisplay<'gc> {
         options.set(HitTestOptions::SKIP_MASK, self.maskee().is_none());
 
         if self.visible() {
+            // A scrollRect clips mouse events at the visible viewport, just like rendering.
+            if let Some(scroll_rect) = self.scroll_rect()
+                && let Some(local_matrix) = self.global_to_local_matrix()
+                && !scroll_rect.contains(local_matrix * point)
+            {
+                return Avm2MousePick::Miss;
+            }
+
             // A loader has at most one child.
             if let Some(child) = self.iter_render_list().next() {
                 if let Some(int) = child.as_interactive() {
--- a/core/src/display_object/movie_clip.rs
+++ b/core/src/display_object/movie_clip.rs
@@ -2992,6 +2992,15 @@ impl<'gc> TInteractiveObject<'gc> for MovieClip<'gc> {
                 }
             }
 
+            // A scrollRect clips mouse events at the container's visible viewport
+            // (in the same way it clips rendering). Children whose position falls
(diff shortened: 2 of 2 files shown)


Dinnerbone:
This seems reasonable! Can you please try to create a regression test for it - one that fails before this fix, and passes after it? Please let us know if you need help.

falpi:
> This seems reasonable! Can you please try to create a regression test for it - one that fails before this fix, and passes after it? Please let us know if you need help.

You can use the application attached to the cited issue. After the patch, the problem no longer occurs.
regards

Dinnerbone:
Yes, and that's amazing - but without _automated regression tests_, we cannot prevent it from breaking again in a week or year because somebody refactored something :(

falpi:
> Yes, and that's amazing - but without _automated regression tests_, we cannot prevent it from breaking again in a week or year because somebody refactored something :(

I understand, but I'm new to Ruffle support and have no idea how to produce this kind of test, and I don't even have the time anyway. Sorry.

Dinnerbone:
You checked the box that says you added tests, which also has a link to a guide on how to make tests. I assumed, opening this PR, that it would then include tests. We are reluctant to accept things without automated tests because there's nothing to stop it from breaking in the future.

I will try to find time to add a test for you based on the issue you linked, but if you want this merged faster it truly isn't difficult to add tests. The hardest part is making a swf at all, which I believe you can do.

falpi:
> You checked the box that says you added tests, which also has a link to a guide on how to make tests. I assumed, opening this PR, that it would then include tests. We are reluctant to accept things without automated tests because there's nothing to stop it from breaking in the future.
> 
> I will try to find time to add a test for you based on the issue you linked, but if you want this merged faster it truly isn't difficult to add tests. The hardest part is making a swf at all, which I believe you can do.

I thought you meant a test application. Now it's clear. I'll try to figure out how to do it for future PR too. Thanks.

n-raine:
There are several other implementations of `TInteractiveObject::mouse_pick_avm1()`/`mouse_pick_avm2()` which I think should all have the same check:

- `EditText` (avm1 & avm2)
- `Avm1Button` (avm1)
- `Avm2Button` (avm2)

(Is just duplicating the same check in each of them the best way to do this? That seems error-prone but maybe there's really no other easy way)

This should also fix #23757, #23576, #13737, and #23678

PR #803, opened 2020-07-02
avm1: Fix issues in `with` and getting/setting StageObject properties (fix #792)

Herschel:
Investigating #792 led me down a rabbit hole fixing several issues:

 * Using GetVariable/SetVariable with `StageObject` properties like `_x` did not work correctly; `StageObject::has_property` should consider these properties.
    * `_x = 10` generally gets compiled to `GetProperty` which is why we didn't see this earlier. Using `with`, it has to fallback to `GetVariable` because `_x` can be anywhere on the scope chain.
 * `with(undefined)` or `null` should do nothing.
 * When crawling the scope or prototype chain, virtual properties should be found even if they are read-only. Previously we would often skip properties if they were read-only.
   * Fixing the above, `TObject::is_property_overwritable` doesn't seem necessary anymore, because read-only properties anywhere on the prototype/scope chain should still "capture" the set attempt; they will just bail out deeper in `Property::set`.
   * Removing `is_property_overwritable` also removes a redundant lookup in the property map. (Longer term, this makes me think we should provide an `Entry`-like API for `TObject`, that will return the property slot (vacant/occupied) along with all attributes, as opposed to having lots of these redundant accessor methods).

PR #23809, opened 2026-05-24
avm1: Do not trace an error on with(undefined)/with(null)

kjarosh:
## Description
It was introduced in https://github.com/ruffle-rs/ruffle/pull/803, however I wasn't able to reproduce it with FP 32 and 9. It's possible this could be traced under some conditions.

## Testing

Covered by tests.

## Checklist

### DIFF
--- a/core/src/avm1/activation.rs
+++ b/core/src/avm1/activation.rs
@@ -2309,13 +2309,7 @@ impl<'a, 'gc> Activation<'a, 'gc> {
         let value = self.context.avm1.pop();
         match value {
             // Undefined/null with is ignored.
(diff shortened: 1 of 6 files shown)


Dinnerbone:
Ah, ok, this is a sneaky one.

Did you know that the flash_log.txt and the Flash "console output" are two different streams with slightly different output?

It looks like "Error: A 'with' action failed because the specified object did not exist." is only in the Flash console output, not the flash log - and many many of the `.fla` tests were copy pasted from that instead of flash_log, rip.

Dinnerbone (approved):

kjarosh:
I suspected it might be something like that :D

PR #23813, opened 2026-05-24
core: Gate device-font kerning

doctorpangloss:
Flash treats embedded glyph fonts and device fonts differently when laying out classic TextField text. Embedded shape fonts can use the font kerning data when TextFormat.kerning is enabled, so Ruffle needs to apply that adjustment after snapping glyph placement into the pixel grid. Device fonts are resolved through the host font system instead, and Flash does not apply the SWF/TTF kerning table for those classic TextField runs.

This keeps the existing kerning visual as an embedded-font test, adds a separate device-font visual test, and gates the SWF kerning adjustment to embedded fonts.

Tests:
- RUSTFLAGS='-D warnings' cargo test -p tests --features imgtests --test tests device_font_kerning


kjarosh:
Please use the PR template for filing a PR.

tests/tests/swfs/fonts/device_font_kerning/Test.as
```diff
@@ -22,6 +23,7 @@ public class Test extends Sprite {
         field.height = 25;
         field.y = y;
         field.border = true;
+        field.embedFonts = true;
```
kjarosh, line 26: This doesn't make sense, the test is about device fonts. Please don't edit it—you can add a new one if you want to.

core/src/font.rs
```diff
@@ -929,8 +937,16 @@ pub trait FontLike<'gc> {
...
                 x += twips_advance;
             } else {
-                // No glyph, zero advance.  This makes it possible to use this method for purposes
-                // other than rendering the font, e.g. measurement, iterating over characters.
+                // No glyph for this character. We report zero advance so
+                // callers can use evaluate for measurement and iteration
+                // without confusing the text cursor.
+                //
+                // This does not match Flash. When a device-font lookup
+                // misses a glyph, Flash falls back to a system font and
```
kjarosh, line 945: What's the difference between device font and system font? That's the same thing from FP's perspective.

core/src/font.rs
```diff
@@ -929,8 +937,16 @@ pub trait FontLike<'gc> {
...
+                // uses that font's advance for the missing character. The
+                // requested font's .notdef glyph is never consulted.
+                // Matching Flash would mean wiring a system-fallback chain
+                // into device-font FontSets.
```
kjarosh, line 949: We do implement device font fallback, it's already wired up. That's why we have a font set, not a font.

core/src/font.rs
```diff
@@ -125,6 +125,8 @@ pub struct EvalParameters {
     /// pairs of letters, separate from the ordinary width between glyphs. This
     /// parameter allows enabling or disabling that feature.
     pub kerning: bool,
+
+    pub device_fonts_kern: bool,
```
kjarosh, line 129: Why are you trying to gate device font kerning in the first place? I guess I don't understand the PR description. Device fonts are different than embedded fonts. Did you notice somewhere that a device font did not have kerning?

PR #23816, opened 2026-05-25
ci: Enable image tests on macos

kjarosh:
## Description

Let's enable them and see what happens.

## Testing

N/A

## Checklist

### DIFF
--- a/tests/tests/swfs/from_shumway/acid/acid-text-5/test.toml
+++ b/tests/tests/swfs/from_shumway/acid/acid-text-5/test.toml
@@ -5,6 +5,7 @@ num_frames = 1
 [image_comparisons.output]
 known_failure = true # https://github.com/ruffle-rs/ruffle/issues/12122
 tolerance = 0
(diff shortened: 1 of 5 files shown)


SuchAFuriousDeath:
Looks like for most of these, you just bump the outliers and it should be fine? :thinking:

tests/tests/swfs/from_shumway/acid/acid-text-5/test.toml
torokati44, line 1: It looks a bit odd to me that a known failure is adjusted, but I'm sure you have a good reason for it. (Like this adjustment being for a part that is different from the known issue.)

kjarosh: It isn't odd, there's always differences in rendering between different backends.

torokati44 (approved):
I mean, why not! Those two edittext FIXMEs are not a huge deal...

PR #23817, opened 2026-05-25
🤖 core: Move flash.text.engine TextLine onto its own DisplayObject

doctorpangloss:
## Description

`TextBlock.createTextLine` previously allocated an `EditText` with the FTE flag set and used that as the AS3 `TextLine` display object. That made the simplest FTE cases return something non-null, but it also tied `flash.text.engine.TextLine` to the classic `TextField` / `EditText` display path.

This PR adds a dedicated `TextLine` display object, registers it with the display object and interactive object enums, and changes `createTextLine` to allocate that object for `flash.text.engine.TextLine`. This first slice is intentionally small: the object can sit in the display tree, own the AVM2 stage object, participate in hit testing as a miss, and keep the existing `EditText::new_fte` fallback for text setup while later PRs move layout and rendering behavior onto the FTE path.

The SWF regression test checks the externally observable behavior in this slice: `createTextLine` returns a non-null line, sets `textLineCreationResult` to `success`, wires `TextBlock.firstLine`, preserves the embedded-font text setup used by the existing fallback path, and still returns null for a `TextBlock` with null content.

This branch is two commits after review feedback:

- `f4389c0984` adds the dedicated display object and observable createTextLine behavior.
- `2cdb0c861d` addresses review feedback without folding it into the original commit.

## Testing

Added `tests/tests/swfs/avm2/textblock_createline_fte`.

- `CARGO_TARGET_DIR=/tmp/ruffle-fte-pr-update-target cargo test -p tests --features imgtests --test tests --no-run`
- `CARGO_TARGET_DIR=/tmp/ruffle-fte-pr-update-target cargo test -p tests --features imgtests --test tests -- textblock_createline_fte`
- `CARGO_TARGET_DIR=/tmp/ruffle-fte-pr-update-target cargo test -p tests --features imgtests --test tests`

Full suite result: `4366 passed; 0 failed; 21 ignored`.

## Checklist


core/src/display_object/fte_text_line.rs
```diff
@@ -0,0 +1,181 @@
...
+use core::fmt;
+use gc_arena::barrier::unlock;
+use gc_arena::lock::Lock;
+use gc_arena::{Collect, Gc, Mutation};
+use ruffle_common::utils::HasPrefixField;
+use std::sync::Arc;
+
+#[derive(Clone, Collect, Copy)]
+#[collect(no_drop)]
+pub struct FteTextLine<'gc>(Gc<'gc, FteTextLineData<'gc>>);
```
kjarosh: Nit: I would just call it `TextLine` I think, it will make it shorter and there's no naming conflict.

tests/tests/swfs/avm2/textblock_createline_fte/Test.as
```diff
@@ -0,0 +1,27 @@
+package {
+import flash.display.Sprite;
+import flash.text.engine.*;
+
+// TextBlock.createTextLine now returns a real flash.text.engine.TextLine
+// display object backed by FteTextLine, not the EditText hack the
+// previous build allocated. The line is null when content is null and
```
kjarosh: It's a fine commit message, but doesn't make sense as a comment on a test.

tests/tests/swfs/avm2/textblock_createline_fte/Test.as
```diff
@@ -0,0 +1,27 @@
...
+// non null otherwise, and the block's textLineCreationResult flips to
+// success.
+[SWF(width="100", height="100")]
+public class Test extends Sprite {
+    public function Test() {
+        var fd:FontDescription = new FontDescription();
+        fd.fontName = "_sans";
```
kjarosh: It's a device font by default, which means you won't have it in tests. If device/embedded doesn't matter it's best to stick with an embedded font here.

core/src/avm2/globals/flash/text/engine/text_block.rs
```diff
@@ -114,7 +98,6 @@ fn apply_format<'gc>(
     element_format: Option<Object<'gc>>,
 ) -> Result<(), Error<'gc>> {
     if let Some(element_format) = element_format {
-        // TODO: Support more ElementFormat properties
```
kjarosh, line 117: Why are you removing these TODOs?

core/src/avm2/globals/flash/text/engine/text_block.rs
```diff
@@ -12,84 +12,69 @@ use crate::avm2::globals::slots::flash_text_engine_text_line as line_slots;
...
             )?;
             return Ok(Value::Null);
         }
-        // Get the content element's text property (it's a getter).
-        // TODO: GraphicElement?
         None => {
+            let _ = element_methods::GET_TEXT;
+            let text_name = AvmString::new_utf8(activation.gc(), "text");
             let txt = content
-                .call_method(element_methods::GET_TEXT, &[], activation)
```
kjarosh, line 55: What's the point of this change?

doctorpangloss:
i've updated this to have more of the textline members we're eventually going to implement. the code coverage warnings are new but they will get addressed as more functionality is added.

core/src/avm2/globals/flash/text/engine/text_block.rs
doctorpangloss: fixed - this stuff was a mistake of taking one big PR and splitting it up into small pieces.

core/src/avm2/globals/flash/text/engine/text_block.rs
doctorpangloss: same here, my approach to breaking up the big PR into small ones bites again. it's back for now in the PR, but it will be changed to the other way later when it makes sense to. it isn't needed to be changed here.

kjarosh (approved):

PR #23819, opened 2026-05-25
render: Skip tessellation when drawing triangles

kjarosh:
## Description

The main idea behind this refactor is that drawing triangles is separate
from drawing paths. The winding behavior is different, they are not
a part of the path (when rendered in between), etc.

This is WIP.

CC @SuchAFuriousDeath 

## Testing

TBD

## Checklist


core/src/avm2/globals/flash/display/graphics.rs
```diff
@@ -1143,128 +1135,80 @@ fn draw_triangles_internal<'gc>(
...
-                read_point(vertices, activation)?,
-            ) {
-                (Some(p0), Some(p1), Some(p2)) => Ok(Some((p0, p1, p2))),
-                _ => Ok(None),
-            }
-        }
+        // In case there's no indices, generate them. The resulting tessellation
+        // structure uses indices anyway, so it's easier to generate them here
+        // instead of supporting two formats in the tessellator.
+        (0..vertices.len()).collect()
```
SuchAFuriousDeath: I'm not sure this is really necessary, you could just leave them out and only generate them when passing them to lyon. My #23761 is set up that way

core/src/drawing.rs
```diff
@@ -233,6 +238,26 @@ impl Drawing {
         self.mark_dirty()
     }
 
+    pub fn draw_triangles(&mut self, vertices: Vec<Point<Twips>>, indices: Vec<usize>) {
```
SuchAFuriousDeath: I have the indices as a Vec\<u32\>

SuchAFuriousDeath:
Overall, I think that the vertex and index collecting logic is done better in my #23761. The logic in current master was actually written by me and it's super ugly, I really don't like it :/

SuchAFuriousDeath:
> Overall, I think that the vertex and index collecting logic is done better in my #23761. The logic in current master was actually written by me and it's super ugly, I really don't like it :/

I would suggest first merging #23761 and then putting your tessellation bypass on top of that.

kjarosh:
Alright, let's do it

PR #23827, opened 2026-05-26
wgpu: Yeet the copy_srgb shader and pipeline

SuchAFuriousDeath:
## Description

Adapts and revives #12720. The `copy_srgb` shader and pipeline existed to bridge Ruffle's linear intermediate framebuffer into an sRGB-format swap-chain texture: the GPU would otherwise auto-encode linear -> sRGB on write, double-encoding values that Ruffle had deliberately kept in sRGB byte-space to match Flash's blending behaviour. The shader manually pre-applied `srgb_to_linear` so the subsequent auto-encode cancelled out.

Since #12720 was opened, wgpu has gained `SurfaceConfiguration::view_formats`, which lets us request a view of the swap-chain texture that differs from its storage format only by the sRGB suffix. We now ask the surface for a non-sRGB view directly, the GPU skips the encode step, and the raw bytes are preserved into the still-sRGB-format texture (which the compositor decodes on present, as it would have anyway).

This addresses the unresolved review concern on #12720 (toggling the sRGB-ness at the view rather than crashing on format mismatch).

Net result:
- `copy_srgb.wgsl`, `Shaders::copy_srgb_shader`, `Descriptors::copy_srgb_pipeline`, and the `actual_surface_format` plumbing through `run_copy_pipeline` are all gone.
- `SwapChainTarget` configures the surface with both the sRGB format and its linear counterpart in `view_formats`, and `get_next_texture()` returns a non-sRGB view.
- `RenderTarget::format()` for the swap chain now returns the linear format, so all the existing call sites that assumed "intermediate buffer format == target format" remain correct without a per-call branch.

## Testing

There is no automated coverage for the sRGB swap-chain path. The visual regression suite renders into `TextureTarget` (hardcoded `Rgba8Unorm`), which never exercised either the old `copy_srgb_pipeline` or the new view-format trick. The only consumer of `SwapChainTarget` in-tree is `WgpuRenderBackend::for_canvas` (web).

I didn't test this yet on all possible platforms.

## Checklist


torokati44:
It would be interesting to see whether things will keep working and appearing correctly with this, on all 8 combinations of: OpenGL/Vulkan, with egui/without egui, final window surface being srgb/linear.

torokati44:
Since `egui` is not optional in the desktop app (so the `--no-gui` argument doesn't matter in this sense, it still goes through `egui` AFAIK), I instead tested these 8 combinations on Linux:

 - windowing system: Wayland | X11
 - surface format: linear | sRGB
 - graphics API: Vulkan | OpenGL

All looked good with shape fill colors and bitmaps.

torokati44:
> The only consumer of SwapChainTarget in-tree is WgpuRenderBackend::for_canvas (web).

Oh, so this should be tested on web then? :thinking: But we used to have sRGB wrongness issues on desktop too (maybe that was before egui? :thinking: )

torokati44:
Stuff looks fine on web as well! With wgpu backend over WebGL.

render/wgpu/src/target.rs
```diff
@@ -63,10 +63,10 @@ impl SwapChainTarget {
...
         device: &wgpu::Device,
     ) -> Self {
-        // Ideally we want to use an RGBA non-sRGB surface format, because Flash colors and
-        // blending are done in sRGB space -- we don't want the GPU to adjust the colors.
-        // Some platforms may only support an sRGB surface, in which case we will draw to an
-        // intermediate linear buffer and then copy to the sRGB surface.
+        // Flash colors and blending are done in sRGB space -- we don't want the GPU to
+        // adjust the colors. We prefer a non-sRGB surface format directly, but on platforms
+        // that only expose sRGB surface formats we view the surface texture as its non-sRGB
+        // counterpart so writes skip the linear->sRGB encode step.
```
torokati44, line 69: So, I hope I remember correctly: Flash (obviously) did naive blending math on colors that were sRGB everywhere, therefore it was "incorrect". But GPUs nowadays aren't super willing to do that, so we emulate the same behavior by keeping everything linear? And the end result will be simply treated as if it was actually sRGB the whole time, and then presented as is on sRGB surfaces, and linearized otherwise, to keep the correct appearance?

SuchAFuriousDeath: Everything stated here is correct based on my understanding.

Dinnerbone (approved):

torokati44:
The corresponding function in common.wgsl became unused with this, but it's counterpart didn't - I think it's fine to leave it there, might come in handy one day.

PR #23831, opened 2026-05-26
core: Extract run_file_dialog helper to reduce boilerplate in file dialog loaders

SuchAFuriousDeath:
## Description

Pure refactor of `core/src/loader.rs`. Extracts a `run_file_dialog` helper that wraps the common pattern shared by `select_file_dialog_avm1`, `select_file_dialog_avm2` and `select_save_file_dialog`: awaiting the `DialogResultFuture`, mapping the result, and dispatching it under an `Activation` against a stashed target object.

Motivation: prerequisite cleanup for an upcoming `FileReferenceList` (multi-file dialog) implementation, which would otherwise add a fourth copy of the same scaffold.

## Testing

No functional change, so nothing to test.

## Checklist

### DIFF
--- a/core/src/loader.rs
+++ b/core/src/loader.rs
@@ -2330,6 +2330,27 @@ impl<'gc> MovieLoader<'gc> {
     }
 }
 
+fn run_file_dialog<D: 'static>(
+    uc: &UpdateContext<'_>,
+    dialog: OwnedFuture<D, Error>,
+    on_result: impl for<'gc> FnOnce(&mut UpdateContext<'gc>, Result<D, Error>) -> Result<(), Error>
+    + 'static,
+) -> OwnedFuture<(), Error> {
+    let player = uc.player_handle();
+
+    Box::pin(async move {
+        let dialog_result = dialog.await;
+
+        // Dialog is done, allow opening new dialogs.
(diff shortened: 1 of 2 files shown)


Lord-McSweeney (approved):

core/src/loader.rs
```diff
@@ -2353,96 +2412,28 @@ pub fn select_file_dialog_avm2<'gc>(
     dialog: DialogResultFuture,
 ) -> OwnedFuture<(), Error> {
     let handle = FileReferenceObjectHandle::stash(uc, target_object);
-    select_file_dialog(uc, FileReferenceHandle::Avm2(handle), dialog)
-}
-
-enum FileReferenceHandle {
-    Avm1(ObjectHandle),
-    Avm2(FileReferenceObjectHandle),
-}
```
torokati44, line 2362: How come this distinction is no longer needed? (Might be obvious to others, but I thought asking couldn't hurt...)

torokati44: Ah, I guess the separate closures passed in now take care of capturing either variants just fine, without having to be smushed into a single function parameter...?

SuchAFuriousDeath: Exactly.

torokati44 (approved):

PR #23837, opened 2026-05-27
core: Snap 1px orthogonal strokes to pixel grid (close #23814)

### LINKED ISSUE #23814: Blurry separators in Grids and inconsistency between mx:Grid and mx:AdvancedDataGrids
### Describe the bug

There are issues with the line drawing algorithms in Grids, and inconsistent behavior is also observed between mx:Grid and mx:AdvancedDataGrid.  Lines that should be rendered with a single black pixel appear blurry and at least 2px wide.

Image

### Expected behavior

This is what the same application looks like in Flash Player.

Image

### Content Location

[application.zip](https://github.com/user-attachments/files/28200640/application.zip)

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

Firefox 150

### Additional information

_No response_

falpi:
## Description

Fixes #23814.

Flash Player rasterizes `graphics.lineStyle(1, color)` + `moveTo`/`lineTo` on integer coordinates as crisp single-pixel strokes even when `pixelHinting=false` (the default). Ruffle previously centered the stroke on the coordinate, so lyon emitted a quad spanning two pixel columns at ~50% alpha. The visible result was blurry separators in Flex grids and Spark lines:

- `mx.controls.DataGrid` with `horizontalGridLines:true`
- `mx.controls.AdvancedDataGrid` (both grouped and non-grouped columns)
- standalone `<s:Line><s:stroke><s:SolidColorStroke>`

### Root cause

Two interacting issues:

1. The `cacheAsBitmap` path (used by Flex grids — `lines.cacheAsBitmap = true` is set in `mx.controls.AdvancedDataGridBaseEx`) computes the offscreen offset from `bounds.x_min`. Because `Drawing::stretch_bounds` expands stroke bounds by ±0.5 px (the stroke radius), `bounds.x_min` ends up fractional and the whole subtree gets shifted off the pixel grid inside the bitmap cache — even fill rects that were natively pixel-aligned come out blurry.

2. For pure-stroke drawings (no cache), lyon's stroke tessellator centers the quad on the path coordinate, so a vertical line at integer `X` produces a quad `(X-0.5, X+0.5)` — half a pixel spread across two columns, blurry even under MSAA.

### Fix

Three coordinated changes, each tightly gated:

1.  **`core/src/display_object.rs`** — floor-snap `bounds.x_min` / `bounds.y_min` before computing the `cacheAsBitmap` offscreen offset, **gated on whether the subtree actually contains an ActionScript-drawn 1 px orthogonal stroke** (the only condition under which `Drawing::stretch_bounds` produces a fractional `x_min`/`y_min`). Resolves the cache-shift case (fill rects in DataGrid become crisp) while leaving alone subtrees whose fractional bounds come from legitimate sources — intentionally half-pixel-placed bitmaps, filter offscreens, etc. The subtree walk uses a new read-only `Graphic::drawing()` accessor; the existing `drawing_mut()` would initialize the `OnceCell` with an empty `Drawing` and shadow `shared.shape` in `Graphic::render_self`, wiping out embedded SWF geometry.

2. **`render/src/tessellator.rs`** — pre-shift the path by `+0.5 px` perpendicular for `DrawPath::Stroke` whose `style.width() <= 1px`, `FillStyle::Color`, and all `LineTo` segments purely horizontal OR purely vertical (no diagonals, no curves, no mixed orientation). lyon then emits a quad `(X, X+1)` instead of `(X-0.5, X+0.5)`, hitting exactly one pixel column under MSAA.

3. **`core/src/drawing.rs::Drawing::render`** — for `Drawing`s made entirely of 1 px orthogonal strokes (no fill, no curves) with a pure-translation world matrix (`a=d=1, b=c=0`, with a `1e-5` tolerance to absorb FP accumulation from Spark `HorizontalLayout`/`VerticalLayout` percent-width resolvers), floor-snap `tx` / `ty`. Compensates the `+0.5` path shift when the host `Sprite` lands at a fractional X (Spark layouts frequently produce `N.5`), keeping the active pixel column stable across layout fractions.

The three changes are independent in scope (cacheAsBitmap path / stroke tessellator / pure-stroke drawings) and gated tightly enough that transformed, scaled, rotated, or mixed fill+stroke drawings are left untouched.

## Testing

No automated regression test added — the bug surfaces with MSAA + Flex/Spark composition patterns that are awkward to express in the existing SWF test suite, and the visual difference (one pixel column vs two at 50%) is below what pixel-diff harnesses currently catch reliably. Manual verification against real Flex content:

- `mx:DataGrid` default → vertical separators crisp (was already crisp, no regression).
- `mx:DataGrid` with `horizontalGridLines:true` → horizontal **and** vertical separators all crisp (was fully blurry before).
- `mx:AdvancedDataGrid` grouped and non-grouped columns → all separators crisp (was blurry before).
- `<s:Line>`  → crisp (was blurry before).
- Scaled / rotated drawings and mixed fill+stroke drawings → unchanged (the gating excludes them).

### Known edge case

A `<s:Line>` as the first child of an `<s:HGroup>` containing an `<s:Spacer width="100%"/>` (percent-width resolver) still renders blurry. The matrix appears as a pure translation at runtime but floor-snap doesn't recover crispness — there's likely a further interaction I haven't isolated. Workaround: drop the redundant `HGroup` container. Left for follow-up.

## Checklist

- [ ] I have made or updated tests where possible. *(see Testing — happy to add one on request)*


falpi:
Update on the 23 visual SWF tests originally failing on this PR's first run.

An earlier revision of the fix called `Graphic::drawing_mut()` while walking the subtree to decide whether to apply the cacheAsBitmap bounds snap. That accessor initializes the `OnceCell` with an empty `Drawing`, after which `Graphic::render_self` rendered the empty `Drawing` instead of `shared.shape` — wiping out the embedded SWF geometry. I've added a read-only `Graphic::drawing()` accessor (analogous to `MovieClip::drawing()`) and the subtree walk now uses that. With this fix, **18 of the 23 failing tests recover automatically** (all `visual/cache_as_bitmap/*`, all `visual/filters/*`, `from_shumway/acid/acid-filter`, `visual/filters/bevel_outer`).

The **5 tests that still fail locally** are:
 - `visual/drawing_api/cursor`
 - `visual/opaque_background`
 - `visual/simple_shapes/gradients/gradients`
 - `from_shumway/acid/acid-chars`
 - `from_shumway/3_joystick` (also has a pre-set `outliers: 2500` tolerance, slightly exceeded)

In all five cases the delta is concentrated on 1-pixel orthogonal strokes, which is by design the target of this PR. I compared the existing `output.expected.png` (the repo's golden, generated by pre-PR Ruffle) with the new output and, where I could spot-check against Flash Player, the **golden itself looks imperfect with respect to the real Flash Player rendering** — e.g. in `visual/opaque_background` the green rectangle in the right half is itself shifted by 1 px vertically in the golden vs Flash, and the new (post-fix) rendering of the inner red strokes lands on a different (and arguably more Flash-compatible) pixel column/row.

I don't have the bandwidth to track down which specific golden is "wrong" against Flash and reshoot them, so I'm leaving the call to the maintainers: if you judge the new rendering closer to Flash Player on those five cases, the goldens can be regenerated as part of merge; if you'd rather restrict the fix further, I'm happy to iterate based on guidance. The fix's behavior on the **real-world target** (Flex `mx:DataGrid` with `horizontalGridLines:true`, `mx:AdvancedDataGrid` grouped and non-grouped, standalone `<s:Line caps="square"/>`) has been verified by hand against Flash Player and matches.

falpi:
**Heads-up: this PR has been reworked** (force-pushed; the description above has been
updated to describe the current approach).

The previous iteration pre-shifted 1px orthogonal stroke paths at tessellation time
(+0.5px perpendicular) and compensated the world translation for pure-stroke drawings.
Testing against more Flex/Spark content exposed structural gaps that a path-space,
per-shape adjustment cannot cover: mixed fill+stroke drawings at fractional
translations, rectangle outlines (mixed orientations), and Spark GraphicElements
drawing into a shared canvas with *different* sub-pixel phases inside the same
`Drawing`. It also perturbed SWF-embedded shapes — which is where most of the golden
churn in the earlier CI runs came from.

The rework leaves the tessellator and the render backends' shape pipeline untouched and
snaps in `Drawing::render` instead, in screen space, per run of consecutive qualifying
stroke paths (paint order preserved): each run's stroke center is snapped onto the
half-pixel of the pixel containing it, computed from the final world matrix — so both
fractional containers and fractional path coordinates are corrected at once. Only
ActionScript-drawn geometry is affected; SWF-embedded shapes render exactly as
mainline.

A second, independent commit fixes the same visual symptom on a different path:
`CommandHandler::draw_line` consumers (text carets, underlines, device text field
borders) rendered blurry on the GL backend, because wherever GL runs on top of DirectX
(ANGLE on Windows — i.e. the web build in any Chromium-based host) line primitives are
translated into antialiased ~1.4px D3D11 lines. This is the same inconsistency already
worked around for Dx12, so the existing line emulation (whole-pixel-edge quads) is
extended to the wgpu GL backend, and the WebGL backend switches to the shared emulation
helpers. Vulkan/Metal are unchanged.

Effect on the test failures discussed earlier:

- `visual/opaque_background`, `visual/simple_shapes/gradients/gradients`,
  `from_shumway/acid/acid-chars` and `from_shumway/3_joystick` **now pass** — they were
  fallout of the tessellation-time shift touching embedded shapes.
- `visual/drawing_api/cursor` still differs, same reason as before: the golden records
  two coincident hairlines blended across two pixel columns; with the fix the line
  occupies a single column and the last-drawn color wins, which matches Flash Player.
  Golden update proposed.
- New: `from_shumway/acid/acid-shapes-testing` now differs from its Ruffle-specific
  `output.ruffle.png` — the fix restores the crisp 1px red rectangle borders that the
  Flash-generated `output.expected.png` shows and Ruffle previously lost (#12118), so
  the rendering moves *toward* the Flash golden. Happy to regenerate the golden from
  the CI artifacts.
- The line-emulation commit does not affect CI image tests (they run on Vulkan).

PR #23844, opened 2026-05-27
core: Rework infrastructure around the FileDialogResult trait

SuchAFuriousDeath:
## Description

Pure refactor of the `FileDialogResult` infrastructure in `core/src/backend/ui.rs` and all its consumers (`core`, `desktop`, `web`, test framework).

Before: `FileDialogResult` was a trait with an `is_cancelled()` method and nullable accessors (`file_name() -> Option<String>`). Each backend had to ship a separate `Null*FileDialogResult` type to represent cancellation, and call sites read cancellation via `if !result.is_cancelled() { ... }`.

After:

```rust
pub enum FileDialogResult {
    Selection(Box<dyn FileDialogSelection>),
    Canceled,
}

pub trait FileDialogSelection: Any {
    fn creation_time(&self) -> Option<DateTime<Utc>>;
    fn modification_time(&self) -> Option<DateTime<Utc>>;
    fn file_name(&self) -> String;
    fn size(&self) -> Option<u64>;
    fn file_type(&self) -> Option<String>;
    fn creator(&self) -> Option<String> { None }
    fn contents(&self) -> &[u8];
    fn write_and_refresh(&mut self, data: &[u8]);
}
```

Cancellation is now expressed at the type level. `file_name()` is no longer `Option<String>` because a selected file always has a name — the previous nullability only existed to support the cancellation sentinel. The `Null*FileDialogResult` boilerplate types are deleted from desktop, web and the test framework.

Motivation: prerequisite cleanup before adding multi-file dialog support (`FileReferenceList`). With this enum/trait split, a follow-up `MultiFileDialogResult { Selection(Vec<Box<dyn FileDialogSelection>>), Canceled }` reuses the same `FileDialogSelection` trait per file with no duplication. Submitting this alone first to keep the multi-file PR scoped to just the new functionality.

No behavior change — `FileReference.browse` / `FileReference.save` in AVM1 and AVM2 behave identically.

## Checklist

### DIFF
--- a/core/src/avm1/globals/file_reference.rs
+++ b/core/src/avm1/globals/file_reference.rs
@@ -8,7 +8,7 @@ use crate::avm1::globals::as_broadcaster::BroadcasterFunctions;
 use crate::avm1::property_decl::{DeclContext, StaticDeclarations, SystemClass};
 use crate::avm1::{NativeObject, Object, Value};
 use crate::avm1_stub;
-use crate::backend::ui::{FileDialogResult, FileFilter};
+use crate::backend::ui::{FileDialogSelection, FileFilter};
 use crate::string::AvmString;
 use gc_arena::barrier::unlock;
 use gc_arena::lock::Lock;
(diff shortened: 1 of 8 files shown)


kjarosh (approved):

kjarosh:
Nit: please add the PR description to the commit description as well.

SuchAFuriousDeath:
> Nit: please add the PR description to the commit description as well.

Done.

PR #23848, opened 2026-05-28
ci: Temporarily disable browser tests on Chrome

danielhjacobs:
## Description

https://github.com/actions/runner-images/issues/14169. Edge is "close enough" for now, until that issue is resolved. This PR is meant to be reverted later.

## Testing

N/A

## Checklist

### DIFF
--- a/.github/workflows/test_web.yml
+++ b/.github/workflows/test_web.yml
@@ -159,7 +159,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        browser: ["chrome", "firefox", "edge"]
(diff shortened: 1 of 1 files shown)


torokati44:
Eh, I think it's acceptable to leave it as a not required check. At least we'll see if it starts working again one day, and we won't forget to re-enable it.

But others might see it differently, as there is indeed no point in running them, and they will just get stuck, eating up runner slots, so if there is explicit disagreement by others, feel free to overrule my opinion.

Thanks for the PR though!

PR #23857, opened 2026-05-30
avm1: Lazily attach XMLSocket native on proto-chain connect()

chameleonxxl:
## Description

Some old ActionScript 1 SWFs "subclass" XMLSocket by grafting `XMLSocket.prototype` onto a custom class's `__proto__` chain and never call `new XMLSocket()` on the instance, then invoke the inherited `connect()`. A real-world example is the FFDK socket library used by the MyCokeGames / CokeMusic Flash titles:

    FFDK.Socket = function (session) {
        this._session = session;
        this._status  = "closed";
    };
    FFDK.Socket.prototype.__proto__        = XMLSocket.prototype;
    FFDK.Socket.prototype.connectXMLSocket = XMLSocket.prototype.connect;
    FFDK.Socket.prototype.connect = function (host, port) {
        this._status = "opening";
        return this.connectXMLSocket(host, port); // <-- inherited connect()
    };

Flash Player attached the underlying socket state lazily on the first `connect()` call, so this worked even though the XMLSocket constructor never ran on the instance. Ruffle only attaches the native object in the constructor, so `XmlSocket::cast(this)` returns `None` here and `connect()` silently returns `Undefined`. The SWF reads that as a failed connection, with no error logged, and stalls with no observable diagnostic.

Lazily attach an `XmlSocket` native when `this` has no native object yet, so these SWFs can connect. Instances that already carry another native (Date, MovieClip, a real XMLSocket, ...) are left untouched, so the shim cannot clobber unrelated state.

Why did you make this change; does it add a feature, or fix a bug?
It fixes a bug in ActionScript 1 movies which using a lazy attach to an XmlSocket

Be sure to link to any content or issues that it affects!
It affects all ActionScript 1 minigames of Coke Studios, you can access them here after registering: https://decibel.fun/play.php

Please also remember that each commit message should describe its own changes too._

## Testing

_How can we test this PR?_
You probably will not be able to test this PR with the games, because they need a game server connection, which is currently not released yet.

_If you're able to, please follow our [test guidelines](https://github.com/ruffle-rs/ruffle/blob/master/CONTRIBUTING.md#test-guidelines) to make a regression test which should fail before this PR, and pass after it. This helps us protect against breaking your changes in the future. These are highly encouraged and will speed up reviewing of your pull request. Some examples of these are SWF regression tests, unit tests, or web integration tests. Describe what you added!_

_If you need help with making tests, or do not believe this change is easily automatically testable, please describe how we can verify the changes ourselves. If there is real world content that we can test on, that helps a lot!_
You could probably make a ActionScript 1 script which trys to open a WebSocket connection over the fixed described way.

## Checklist

### DIFF
--- a/core/src/avm1/globals/xml_socket.rs
+++ b/core/src/avm1/globals/xml_socket.rs
@@ -122,6 +122,46 @@ pub fn connect<'gc>(
     this: Object<'gc>,
     args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
+    // Compatibility shim for old AS1 SWFs (notably the FFDK socket library
+    // used by the MyCokeGames / CokeMusic Flash titles) that hijack
+    // `XMLSocket.prototype.connect` via a `__proto__` chain on a custom class
+    // WITHOUT ever calling `new XMLSocket()` first. e.g.:
+    //
+    //     FFDK.Socket = function (session) {
+    //         this._session = session;
+    //         this._status = "closed";
+    //     };
+    //     FFDK.Socket.prototype.__proto__ = XMLSocket.prototype;
+    //     FFDK.Socket.prototype.connectXMLSocket = XMLSocket.prototype.connect;
+    //     // ...
+    //     FFDK.Socket.prototype.connect = function (host, port) {
+    //         this._status = "opening";
+    //         return this.connectXMLSocket(host, port); // <-- this call
+    //     };
+    //
+    // Native Flash Player attached the underlying socket state lazily on the
+    // first `connect()` call, so the call would still succeed even though the
+    // constructor was never run on the instance. Ruffle's constructor-based
+    // attach otherwise makes `XmlSocket::cast` below silently fail and the
+    // function returns `Undefined`, which the SWF interprets as a connection
+    // failure with no observable error.
+    //
+    // To preserve compatibility we lazily attach the native data when `this`
+    // is a plain AS1 Object (i.e. has no other native attached). Objects that
+    // already carry another native (Date, MovieClip, ...) are left alone, so
+    // this shim cannot accidentally clobber unrelated state.
+    if matches!(this.native(), NativeObject::None) {
+        let xml_socket = XmlSocket(Gc::new(
+            activation.gc(),
+            XmlSocketData {
+                handle: Cell::new(None),
+                timeout: Cell::new(20000),
+                read_buffer: RefCell::new(VecDeque::new()),
+            },
+        ));
+        this.set_native(activation.gc(), NativeObject::XmlSocket(xml_socket));
+    }
+
     if XmlSocket::cast(this.into()).is_some() {
         let host = args
             .get(0)


Lord-McSweeney:
Hello, thank you for the contribution. Please open an issue report for the site you mentioned and attach the SWFs, to ensure that they won't be lost.

core/src/avm1/globals/xml_socket.rs
```diff
@@ -122,6 +122,46 @@ pub fn connect<'gc>(
...
+    // constructor was never run on the instance. Ruffle's constructor-based
+    // attach otherwise makes `XmlSocket::cast` below silently fail and the
+    // function returns `Undefined`, which the SWF interprets as a connection
+    // failure with no observable error.
+    //
+    // To preserve compatibility we lazily attach the native data when `this`
+    // is a plain AS1 Object (i.e. has no other native attached). Objects that
+    // already carry another native (Date, MovieClip, ...) are left alone, so
+    // this shim cannot accidentally clobber unrelated state.
+    if matches!(this.native(), NativeObject::None) {
```
Lord-McSweeney, line 153: Can you add test coverage for these changes?

Lord-McSweeney, line 153: Are you sure this is correct? This would mean that any object that gets `XMLSocket.prototype.connect` called on it becomes an `XMLSocket` instance.

chameleonxxl:
Thanks for the review!

> Can you add test coverage for these changes?

Done, I pushed two tests:

* `avm1/xml_socket_proto_chain` reproduces the FFDK pattern from the affected games.
* `avm1/xml_socket_connect_plain_object` calls `XMLSocket.prototype.connect` on a plain object, which is exactly the case you asked about.

Both fail before this PR and pass after it.

> Are you sure this is correct? This would mean that any object that gets `XMLSocket.prototype.connect` called on it becomes an `XMLSocket` instance.

I checked this in Flash Player 7 (the debug standalone player shipped with Flash MX 2004) against a local server that replays the same exchange as the `socket.json` of the tests. That is exactly what Flash does. With a plain object that does not have `XMLSocket.prototype` in its prototype chain at all:

```actionscript
var obj = new Object();
obj.onConnect = function(success) {
	trace("onConnect: " + success);
	if (success) { trace("send returned: " + XMLSocket.prototype.send.call(this, "Hello!")); }
};
obj.onData = function(data) { trace("onData: " + data); };
obj.onClose = function() { trace("onClose"); };

trace("connect returned: " + XMLSocket.prototype.connect.call(obj, "localhost", 8001));
trace("instanceof XMLSocket: " + (obj instanceof XMLSocket));
```

Flash Player 7 traces:

```
connect returned: true
instanceof XMLSocket: false
onConnect: true
send returned: 
onData: <root><item /></root>
onClose
```

So the connection really is established, the callbacks are dispatched and `send()` works, while the prototype chain is left untouched, which is why `instanceof XMLSocket` stays `false`. That matches this PR: the native data is attached, the object is not re-prototyped. Restricting the attach to objects that have `XMLSocket.prototype` in their prototype chain would actually be less accurate than what Flash does.

For the record, `send()` and `close()` before the first `connect()` do nothing and return `undefined` in Flash, which is also what this PR does, since it only attaches in `connect()`. This is covered by `xml_socket_proto_chain`.

The expectation files of both tests are the traces recorded in Flash Player 7, not Ruffle output.

> Please open an issue report for the site you mentioned and attach the SWFs, to ensure that they won't be lost.

Opened as #24729, with the SWFs of four of the affected games attached.

PR #23861, opened 2026-05-31
tests: add `known_failure` support to audio assertions

moulins:
## Description

Allow marking expected audio assertion failures with `known_failure = true`, like trace outputs or image comparisons.

## Testing

An test using this feature has been added to PR #23676.

## Checklist

### DIFF
--- a/tests/framework/src/runner.rs
+++ b/tests/framework/src/runner.rs
@@ -304,7 +304,7 @@ impl TestRunner {
             return Err(anyhow!("Audio assertions require audio"));
         };
 
-        for assertion in self.audio_assertions.values() {
+        for (assertion_name, assertion) in &self.audio_assertions {
             if !assertion.frames.includes_frame(self.current_iteration) {
                 continue;
             }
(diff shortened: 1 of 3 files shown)


SuchAFuriousDeath (approved):

tests/framework/src/runner.rs
```diff
@@ -315,27 +315,34 @@ impl TestRunner {
...
+                    "Audio assertion '{assertion_name}': expected max audio amplitude to be at least {min_max_amplitude}, was {current_max} at frame {frame}",
+                ))
+            } else {
+                Ok(())
+            };
+
+            match result {
+                Err(err) if !assertion.known_failure => return Err(err),
+                Ok(()) if assertion.known_failure => {
+                    // TODO: should we allow known_failure assertions to succeed on *some* frames as long as they fail at least once?
```
Lord-McSweeney, line 339: What would happen if some code were to be changed to make a test pass on its first few frames, but not on its later frames? Wouldn't that mean that the test would "fail" regardless of whether it were marked `known_failure` or not?

moulins: Exactly; in that case you'd need to split the assertion into two, one `known_failure = false` and one `known_failure = true`.

Lord-McSweeney (approved):

PR #23869, opened 2026-05-31
core: Do not dispatch clip events not present in allEventFlags

kjarosh:
## Description

This makes Ruffle behave like Flash, when an event is not present in
allEventFlags, the handler will not be dispatched.

## Testing

Test included.

## Checklist


Lord-McSweeney:
Is this true in all SWF versions?

kjarosh:
It appears to be the case, checked SWF 6,8,15

kjarosh:
For SWF5 it appears to be the case as well

Lord-McSweeney (approved):

PR #23870, opened 2026-06-01
core: Fix BitmapData.hitTest when alpha_threshold=0 (close #19253)

### LINKED ISSUE #19253: Super Smash Flash 2 -- Collision Issues
### Describe the bug

Collision issues in Ruffle:
Image
It looks like collisions are not being detected properly.

### Expected behavior

Collision in the SSF2 launcher:
Image

### Content Location

I downloaded SSF2 Beta (1.3.1.2 beta), [Windows 64-bit (Portable)](https://cdn.supersmashflash.com/ssf2/downloads/14ac5596/SSF2BetaWindows.v.1.3.1.2.portable.zip), 
https://www.supersmashflash.com/play/ssf2/downloads/

After extracting, I opened the index.htm file with self-hosted Ruffle.

To see how the game is supposed to behave, I ran the SSF2.exe file.

### Affected platform

Self-hosted version

### Operating system

Windows 11 Home

### Browser

Edge Version 131.0.2903.146 (Official build) (64-bit)

### Additional information

_No response_

masterwebx:
## Description

Closes #19253. Fixes `BitmapData.hitTest` treating `alpha_threshold = 0` as "every pixel is opaque", causing bounding-box collision instead of pixel-perfect collision.

When `alpha_threshold` is `0`, the comparison `pixel.alpha() >= alpha_threshold` reduces to `u8 >= 0`, which is always `true` — including for fully transparent pixels. The fix is `alpha_threshold.max(1)`, which maps `0 → 1` while leaving all other threshold values unchanged.

The change is applied to `hit_test_point` and `hit_test_bitmapdata` in `core/src/bitmap/operations.rs`. `hit_test_rectangle` is left unchanged because Flash Player genuinely returns `true` for threshold=0 against a transparent rectangle (confirmed by the existing `avm1/bitmap_data_hittest` regression test — `bmd.hitTest({x:0,y:0}, 0, rect)` → `true`).

The canonical repro is Super Smash Flash 2 (`SSF2Portable.swf`), which calls `bitmapData.hitTest(firstPoint, 0, characterPoint)` (the `hit_test_point` variant) with an explicit `threshold = 0` in its `BitmapCollisionBoundary` class. With the bug, characters walk on an invisible flat ceiling at the top of the collision bitmap's bounding box instead of following the terrain's actual shape. Drop-through platforms also fail because the entire bounding rectangle is treated as solid. Flat terrain works coincidentally because the bounding-box top edge happens to match the surface.

## Testing

```
cargo test --package ruffle_core --lib bitmap::operations::tests
```

New unit tests added directly in `operations.rs` covering the regression cases: threshold=0 with a transparent pixel (must be `false`) and threshold=0 with an opaque pixel (must be `true`).

## Checklist


torokati44:
It would be interesting to see the behavior with intermediate alpha levels - perhaps its a `>` vs. `>=` thing, not a clamping of just `0`.

kjarosh:
Can you add a SWF test for it? Unit tests in such cases are not super useful, because they require assumptions about how Flash Player works.

masterwebx:
Thanks for the feedback! I've verified the behavior against a real Flash Player (Adobe AIR 30.0 / `adl.exe`) and have two updates:

**Re: `>` vs `>=` (@torokati44)**

Flash uses `>=`. I confirmed this with a boundary test: a pixel with `alpha = 0xA0` and `threshold = 0xA0` returns `true`. Our code uses `alpha >= threshold.max(1)`, which for the threshold=0 case becomes `alpha >= 1` — mathematically identical to `alpha > 0` for integers. So the semantics are correct.

**Re: SWF test (@kjarosh)**

Added `tests/tests/swfs/avm2/bitmapdata_hittest_threshold/` with:
- `test.swf` — compiled against Flash Player 30 SDK (`playerglobal.swc`) via mxmlc
- `output.txt` — golden output recorded from actual Flash Player (Adobe AIR 30.0 runtime via `adl.exe`, with `TraceOutputFileEnable=1` in `mm.cfg`)

The test covers: threshold=0 on transparent/opaque/out-of-bounds points, the `>=` boundary case (`alpha == threshold`), and bmd-vs-bmd with threshold=0.

**One correction to the original commit:** `.max(1)` should only apply to `hit_test_point`, not `hit_test_bitmapdata`. For `hit_test_bitmapdata`, Flash genuinely uses straight `>= threshold` — a threshold of 0 means *all* pixels are considered opaque (even fully transparent ones), so `bmd.hitTest(origin, 0, bmd2, origin, 0)` returns `true` even when both bitmaps are fully transparent. The new test encodes this behavior. I've reverted the `.max(1)` from `hit_test_bitmapdata` in this update.

masterwebx:
Hi I need help here. What do I need to do to get this approved?

core/src/bitmap/operations.rs
```diff
@@ -1968,3 +1968,149 @@ pub fn pixel_dissolve<'gc>(
 
     raw_perm_index as i32
 }
+
+#[cfg(test)]
+mod tests {
```
kjarosh: If you covered all these cases in SWF tests, you can remove them

masterwebx: Done, just pushed an update with the tests removed from operations.rs

masterwebx:
Checked in a minimal SWF regression test at `tests/tests/swfs/avm2/bitmapdata_hittest_threshold/` (`Test.as`, `test.swf`, `output.txt`, `test.toml`). Golden output was recorded from Adobe AIR 30.0. SSF2 was only used for manual verification — it isn't checked in because it's copyrighted.

## What the SWF tests

The SWF builds a small 5×5 bitmap with three known pixels on column 2:

- `(2,2)` — fully transparent (`alpha = 0`)
- `(2,3)` — semi-transparent (`alpha = 0xA0`)
- `(2,4)` — fully opaque (`alpha = 0xFF`)

### `hitTest(firstPoint, threshold, secondPoint)` — the case this PR fixes

| Case | Expected | Why |
|------|----------|-----|
| threshold=0, transparent pixel `(2,2)` | `false` | Main regression: threshold=0 must not treat every pixel as opaque |
| threshold=0, opaque pixel `(2,4)` | `true` | Opaque pixels must still hit with threshold=0 |
| threshold=0, out-of-bounds `(10,10)` | `false` | Unchanged behavior |

This matches the call pattern SSF2 uses in `BitmapCollisionBoundary` (`hitTest(firstPoint, 0, characterPoint)`).

### Boundary semantics (`>=` vs `>`)

| Case | Expected |
|------|----------|
| threshold=0xA0, pixel alpha=0xA0 | `true` |
| threshold=0xA1, pixel alpha=0xA0 | `false` |
| threshold=0x9F, pixel alpha=0xA0 | `true` |

Confirms Flash uses `>=`, not `>`.

### `hitTest(firstPoint, threshold, otherBmd, otherPoint, otherThreshold)` — unchanged by this PR

| Case | Expected |
|------|----------|
| Both bitmaps fully transparent, both threshold=0 | `true` |
| Both bitmaps fully opaque, both threshold=0 | `true` |
| Src opaque / test transparent, both threshold=0 | `true` |

For bitmap-vs-bitmap, Flash uses straight `alpha >= threshold` with no `max(1)` clamp, so threshold=0 still counts fully transparent pixels. This PR only changes the point variant.

### Not covered here (already covered elsewhere)

`hitTest(..., Rectangle)` with threshold=0 is unchanged and still covered by the existing `avm1/bitmap_data_hittest` test.

core/src/avm2/globals/flash/display/bitmap_data.rs
```diff
@@ -864,17 +864,15 @@ pub fn hit_test<'gc>(
         );
         let second_threshold = args.get_u32(4).clamp(0, u8::MAX.into()) as u8;
 
-        let result = operations::hit_test_bitmapdata(
```
Lord-McSweeney, line 867: Why was this changed?

core/src/bitmap/operations.rs
```diff
@@ -809,7 +809,7 @@ pub fn hit_test_point(
             .read_area(PixelRegion::for_pixel(x, y), renderer)
             .get_pixel32_raw(x, y)
             .alpha()
-            >= alpha_threshold
+            >= alpha_threshold.max(1)
```
Lord-McSweeney, line 813: This affects AVM1 too. Can you confirm that this behavior is present in AVM1 as well, and add an AVM1 test?

Lord-McSweeney, line 813: Could you add a comment explaining what this does? Something like
```
// If `alpha_threshold` is 0, consider it to be 1.
```
above?

masterwebx: Done

masterwebx: Confirmed. Both AVM1 and AVM2 call the same `operations::hit_test_point`, so the fix applies to both.

Added `tests/tests/swfs/avm1/bitmapdata_hittest_threshold/` in `dcbf1d95` with golden output matching Flash Player behavior for threshold=0 on:

- a transparent in-bounds pixel → `false`
- an opaque pixel → `true`
- an out-of-bounds point → `false`

core/src/avm2/globals/flash/display/bitmap_data.rs
masterwebx: Undid this change, was not neccesary

masterwebx:
Hello can I get some help here! Lets get this closed please 🙏

Cleod9 (approved):

Cleod9:
> Hello can I get some help here! Lets get this closed please 🙏

+1 to this, it would be huge for us on SSF2 which relies heavily on bitmap collision testing 🙇‍♂️

Lord-McSweeney (approved):
LGTM, sorry about the wait

Lord-McSweeney:
Could you rename the PR title to replace the "avm2" prefix with "core", since this affects both AVM1 and AVM2 now?

kjarosh (approved):
Thank you!

PR #23878, opened 2026-06-02
web: Support named access on document.embeds

danielhjacobs:
## Description

Add support for named accessors on document.embeds (e.g. document.embeds.movieemb or document.embeds["movieemb"]

https://sombrastorage.z19.web.core.windows.net/ruffle/js_flash tries to use this or named access on document (see https://github.com/ruffle-rs/ruffle/issues/22442 and https://github.com/ruffle-rs/ruffle/pull/22465) for its getFlashMovieObject function:
```js
function getFlashMovieObject()
{
  if (window.movieemb) 
  {
  let myMovie = window.movieemb; //window is standard to W3C, but only binds the name property
	document.getElementById("f1").innerHTML = "Hit " + myMovie + " directly!";
    return myMovie;
  }
  if (navigator.appName.indexOf("Microsoft Internet")==-1)
  {
	document.getElementById("f1").innerHTML = 'Hit on the embeds!';
    if (document.embeds && document.embeds["movieemb"])
      return document.embeds["movieemb"]; 
  }
  else // if (navigator.appName.indexOf("Microsoft Internet")!=-1)
  {
	  document.getElementById("f1").innerHTML = 'Hit on the ID!';
	  return document.getElementById("movieemb");
  }
}
```

## Testing

`npm install`
`npm run build`
`npm run wdio --workspaces --if-present -- --headless --parallel --firefox`

## Checklist

### DIFF
--- a/web/packages/core/src/internal/polyfill-document-embeds.ts
+++ b/web/packages/core/src/internal/polyfill-document-embeds.ts
@@ -185,7 +185,19 @@ export function polyfillDocumentEmbeds(tries: number) {
                         if (!Number.isNaN(index) && index >= 0) {
                             return nodes()[index];
                         }
+
+                        // Let native properties ('length', 'item', etc.) pass through normally
+                        if (Reflect.has(target, prop)) {
+                            return Reflect.get(target, prop, receiver);
+                        }
+
+                        // Fallback to named item resolution for standard dot/bracket notation
+                        const element = target.namedItem(prop);
+                        if (element) {
+                            return element;
+                        }
                     }
+
                     return Reflect.get(target, prop, receiver);
                 },
                 has(target, prop) {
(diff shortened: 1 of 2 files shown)


web/packages/core/src/internal/polyfill-document-embeds.ts
```diff
@@ -185,7 +185,19 @@ export function polyfillDocumentEmbeds(tries: number) {
                         if (!Number.isNaN(index) && index >= 0) {
                             return nodes()[index];
                         }
+
+                        // Let native properties ('length', 'item', etc.) pass through normally
+                        if (Reflect.has(target, prop)) {
```
kjarosh, line 190: IIUC `embeds["length"]` will return the length as a number, even if there's an embed called "length"?

danielhjacobs: Fair point, let me check, knowing the weirdness of JS it's probably the opposite (this should be last)

danielhjacobs: Actually yes, this is true and works with my code. Tested on modern Firefox: `<embed name="length" src="logo-anim.swf">`. Without Ruffle enabled or with this branch, this is `1`, not the embed/ruffle-embed element.

kjarosh: Can we add a test case for it?

danielhjacobs: Added

kjarosh (approved):

PR #23883, opened 2026-06-03
avm2: Add type enum for error 2006

ChrisCPI:
## Description

The `Stage3D` `x`/`y` setters (TBA) can throw this error as an ArgumentError (https://docs.ruffle.rs/en_US/FlashPlatform/reference/actionscript/3/flash/display/Stage3D.html#x). This adds the enum now so that those setters can be implemented without needing to change all of the `make_error_2006` calls then.

## Testing

No tests, everything should behaviorally be the same.

## Checklist

### DIFF
--- a/core/src/avm2/error.rs
+++ b/core/src/avm2/error.rs
@@ -1322,14 +1322,24 @@ pub fn make_error_2005<'gc>(
     ))
 }
 
+pub enum Error2006Type {
+    // Stage3D x/y setters (not yet implemented) can throw this error as an ArgumentError
+    #[expect(unused)]
+    ArgumentError,
+    RangeError,
+}
+
 #[inline(never)]
 #[cold]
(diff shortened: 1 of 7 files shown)


core/src/avm2/error.rs
```diff
@@ -1322,14 +1322,23 @@ pub fn make_error_2005<'gc>(
...
-        "Error #2006: The supplied index is out of bounds.",
-        2006,
-    ))
+pub fn make_error_2006<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    kind: Error2006Type,
+) -> Error<'gc> {
+    let message = "Error #2006: The supplied index is out of bounds.";
+    make_error!(match kind {
+        // TODO: Stage3D x/y setters can throw this error as an ArgumentError
```
kjarosh: Don't submit commented out code, just add expect(unused) when you get warnings

kjarosh (approved):

PR #7637, opened 2022-08-11
avm1: Implement flash.net.FileReferenceList

CUB3D:
This partially implements `flash.net.FileReferenceList` (#308)

Todos:
- [x] fix dialog struct namings
- [x] add tests
- [x] fix test builds
- [x] fix web builds
- [x] Horrific rebase

Issue #308, opened 2020-01-11
Actionscript Tracking: flash.net.FileReferenceList


Dinnerbone:
This is the tracking issue for **flash\.net\.FileReferenceList** in AVM1. We will check off each item of progress as appropriate, and fill in any relevant or missing information as we continue development of Ruffle.
# Legend
Each checkbox is independent of another. It's entirely possible for something to be tested but not exist yet, or for us to believe that it's completely functional but we haven't made enough tests to prove it.
## "Property Exists"
This means the item exists, but may not necessarily be fully implemented.
## "Has Test Coverage"
This means that we have believe that we have a good test coverage of this item, regardless of if those tests pass. It's okay to have tests available but not implement the item yet.
## "Completely Functional"
This means we believe that the item is completely implemented, and no more work needs to be done towards making it functional.

# flash\.net\.FileReferenceList
### `_listeners : Array`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional

### `broadcastMessage()`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional

### `fileList : Array`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional

### `FileReferenceList()`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `browse(typelist)`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `addListener(listener)`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional
### `removeListener(listener)`
- [ ] Property Exists
- [ ] Has Test Coverage
- [ ] Completely Functional

ROBERT-MCDOWELL:
bump

ROBERT-MCDOWELL:
It would be so nice to implement this class!!! I developed 15 years ago a super nice file manager to upload and download files, preview, etc.... I passed so much time on it. bug free....

SuchAFuriousDeath:
@ROBERT-MCDOWELL Can you check with latest master? I implemented it in https://github.com/ruffle-rs/ruffle/pull/23889

ROBERT-MCDOWELL:
my filemanager is loading well, but I cannot upload any multiple files since the browse event stil not working.

SuchAFuriousDeath:
[...]

PR #23889, opened 2026-06-03
avm1: Implement FileReferenceList

SuchAFuriousDeath:
## Description
Significantly progresses #308.
Supersedes #7637.

## Testing

Tests are included. 

## Checklist


danielhjacobs:
Presumably supersedes https://github.com/ruffle-rs/ruffle/pull/7637?

SuchAFuriousDeath:
> Presumably supersedes #7637?

It does indeed.

core/src/avm1/globals/utils.rs
```diff
@@ -0,0 +1,84 @@
+//! Shared utilities for AVM1 globals.
```
Lord-McSweeney: I think the preferred method across the codebase is to keep this in `file_reference.rs` (assuming that was where it originally came from) and reference the method there in `file_reference_list.rs`

SuchAFuriousDeath: Sure, no problem

core/src/avm1/globals/file_reference.rs
```diff
@@ -220,6 +224,83 @@ pub fn file_type<'gc>(
...
 }
 
+/// Helper error type so that `parse_file_filter_array` can use `?`.
+struct ReturnFalse;
+
+impl<'gc> From<Error<'gc>> for ReturnFalse {
+    fn from(_: Error<'gc>) -> Self {
+        Self
+    }
+}
```
moulins, line 241: This is very suspicious; any exception raised by things like `coerce_to_string` will be swallowed, which is almost certainly wrong.

Instead, `parse_file_filter_array` should probably return something like `Result<None<Vec<_>>, Error<'gc>>`, with the `None` case representing the AS2 method returning `false`.

SuchAFuriousDeath: I think there's a test added for all the `false` cases and the error is really never returned, it's always `false`.

kjarosh: Even if that's the case, instead of ignoring the error you can return it and ignore it in the caller (afer e.g. logging it).

SuchAFuriousDeath: Ok and what should be returned instead of the `return Err(ReturnFalse)`? There's no error there that's produced by an A method.

moulins: > I think there's a test added for all the false cases and the error is really never returned, it's always false.

But I can't find any `throw` keyword in the new tests?

For example:
```as
FileReferenceList.browse({
  // your current code will swallow the exception, is this correct?
  extension: { toString: function() { throw "oops"; } },
});
```
Another edge-case that ideally should be tested is: what happens when the `extension`/etc members etc are dynamic properties or only present on the prototype?

> Ok and what should be returned instead of the return Err(ReturnFalse)? There's no error there that's produced by an A method.

A possibility would be returning `Ok(Some(file_filters))` when the filters are valid, and `Ok(None)` otherwise.

tests/tests/swfs/avm1/file_reference_list_browse_invalid_filters/test.as
```diff
@@ -0,0 +1,53 @@
...
+        trace("// empty description");
+        trace(list.browse([emptyDesc]));
+
+        var emptyExt = new Object();
+        emptyExt.description = "txt";
+        emptyExt.extension = "";
+        trace("// empty extension");
+        trace(list.browse([emptyExt]));
+
+        // The following four cases document a pre-existing oddity inherited
```
kjarosh: What do you mean by pre-existing here? Every behavior we're testing is pre-existing

core/src/avm1/globals/file_reference_list.rs
```diff
@@ -21,5 +30,60 @@ fn constructor<'gc>(
     _this: Object<'gc>,
     _args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
+    // `FileReferenceList` instances carry no native data; matching Flash, the
+    // `__proto__` slot stays writable from ActionScript.
```
kjarosh: When is `__proto__` non-writable?

core/src/avm1/globals/file_reference.rs
SuchAFuriousDeath: I will extend the test to include this.

tests/tests/swfs/avm1/file_reference_list_browse_invalid_filters/test.as
```diff
@@ -0,0 +1,53 @@
...
+        trace(list.browse([noExt]));
+
+        var emptyDesc = new Object();
+        emptyDesc.description = "";
+        emptyDesc.extension = "*.txt";
+        trace("// empty description");
+        trace(list.browse([emptyDesc]));
+
+        var emptyExt = new Object();
+        emptyExt.description = "txt";
```
kjarosh: If you feel like it, you could also test cases where these properties have flags (e.g. non enumerable, version 10, etc.).

moulins (approved):

Lord-McSweeney (commented):
Could you change `core::avm1::globals::file_reference::constructor` to use `FileReferenceObject::new` instead of manually constructing the `FileReferenceObject`?

SuchAFuriousDeath:
> Could you change `core::avm1::globals::file_reference::constructor` to use `FileReferenceObject::new` instead of manually constructing the `FileReferenceObject`?

Done :saluting_face:

core/src/avm1/globals/file_reference.rs
```diff
@@ -220,80 +224,85 @@ pub fn file_type<'gc>(
     Ok(Value::Undefined)
 }
 
-/// Helper error type so that `parse_file_filter_array` can use `?`.
-struct ReturnFalse;
```
kjarosh: Should we squash? If not, why add and remove ReturnFalse?

SuchAFuriousDeath: I'll fix it up, sorry :(

core/src/avm1/globals/file_reference.rs
```diff
@@ -220,6 +224,88 @@ pub fn file_type<'gc>(
...
+
+/// Parse the file filter argument used by `FileReference.browse` and
+/// `FileReferenceList.browse`.
+///
+/// Returns `Ok(Some(filters))` on success, or `Ok(None)` if `browse()` should
+/// bail out and return `false` without opening a dialog. Exceptions raised
+/// during property coercion are propagated.
+pub fn parse_file_filters<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    args: &[Value<'gc>],
```
Lord-McSweeney: nit - It seems a bit odd to me to pass the entire args slice here- is there an advantage to this over this function simply taking a `Option<Value<'gc>>`, the first argument? Personally I would find it easier to understand the inputs of this function if it were taking only the first argument.

Lord-McSweeney (approved):

core/src/avm1/globals/file_reference.rs
SuchAFuriousDeath: Good point I guess, I will redo it.

ROBERT-MCDOWELL:
@SuchAFuriousDeath the browse event still not working so I cannot select multiple files for upload. about download, it works but the progress event is not fired so the progressbar never reaches 100%

PR #23892, opened 2026-06-03
avm1: Match Flash broadcaster prototype enumeration order

SuchAFuriousDeath:
## Description

Noticed during development of #23889.

AVM1 enumerates own properties in reverse insertion order. The broadcaster methods on `MovieClipLoader`, `FileReference`, `FileReferenceList`, `System.IME`, `Stage`, `Key`, `Mouse` and `Selection` were defined in the order `_listeners, addListener, removeListener, broadcastMessage`, so `for-in` yielded the reverse — the opposite of Flash, which yields `_listeners, removeListener, addListener, broadcastMessage`.

Define them in the matching reverse order so iteration lines up with Flash. Additionally, give `_listeners` the `DONT_DELETE` flag Flash assigns to it (previously `DONT_ENUM` only).

## Testing

Good luck reviewing the diff :(

It seems to be true that _listeners comes prior to broadcastMessage everywhere I looked though (which is not every changed line), so it seems this change is correct. There are other order differences with Flash, so comparing it visually is difficult.

## Checklist

### DIFF
--- a/core/src/avm1/globals/as_broadcaster.rs
+++ b/core/src/avm1/globals/as_broadcaster.rs
@@ -197,11 +197,9 @@ fn initialize_internal<'gc>(
 ) {
     broadcaster.define_value(
         context.gc(),
-        istr!(context, "_listeners"),
-        ArrayBuilder::new_with_proto(context, array_proto)
-            .with([])
-            .into(),
-        Attribute::DONT_ENUM,
+        istr!(context, "broadcastMessage"),
+        functions.broadcast_message.into(),
+        Attribute::DONT_DELETE | Attribute::DONT_ENUM,
     );
     broadcaster.define_value(
         context.gc(),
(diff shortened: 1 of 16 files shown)


SuchAFuriousDeath:
Oops, forgot some tests :( I'll try to see if I can solve the other order differences as well. Maybe the outputs can be more comparable with extra fixes.

core/src/avm1/globals/as_broadcaster.rs
```diff
@@ -195,13 +195,15 @@ fn initialize_internal<'gc>(
     functions: BroadcasterFunctions<'gc>,
     array_proto: Object<'gc>,
 ) {
+    // Define in the reverse of Flash's enumeration order, so that AVM1's
+    // newest-first `for-in` iteration yields
+    // `_listeners, removeListener, addListener, broadcastMessage`, matching
+    // Flash Player.
```
kjarosh: It's a fine commit message, but that duplicates (1) the code below, and (2) our tests, so we can drop it.

kjarosh (approved):

PR #23893, opened 2026-06-03
avm1: Define `constructor` as own property on system class objects

SuchAFuriousDeath:
## Description
Flash exposes `constructor` as an own property on every system class (e.g. `FileReference`, `MovieClipLoader`, `Function`, `Object`...), pointing to the `Function` class. Ruffle previously only registered `constructor` on prototypes, so iterating own properties of the class object itself did not show it.

Plumb a `function_class` field through `DeclContext` (`None` while `Object` and `Function` are bootstrapping; populated to `function.constr` afterwards), and have `class`, `empty_class` and `native_class_with_proto` attach `constructor` to every class they build. Register `constructor` on `Object`'s and `Function`'s own class objects retroactively, once `Function` exists.

## Testing

Test changes included.

## Checklist


core/src/avm1/property_decl.rs
```diff
@@ -1,16 +1,52 @@
...
+/// - [`Self::Last`] — `constructor` is inserted first, then `__proto__`, then
+///   `prototype`. Iteration: `prototype, __proto__, constructor`. Used by
+///   `FileReference`, `FileReferenceList`, `flash.geom.Matrix`,
+///   `flash.geom.Rectangle`, `flash.automation.Configuration`,
+///   `flash.automation.ActionGenerator`, and
+///   `System.security.PolicyFileResolver`.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum PrototypeOrder {
+    /// `prototype` inserted first → iterated last.
+    First,
```
kjarosh: Just a random thought: maybe it's better to call it `Pcp`/`Cpp`, as this enum does not only specify the order of `prototype`, but all three properties.

(I planned to implement it the same way, so I had thought about it a bit.)

SuchAFuriousDeath: I was also considering PrototypeFirst/PrototypeLast.

kjarosh: Sure, it all makes sense. I guess what I'm saying is that I think it would be nice if we listed the order of all 3 props instead of the one. And TBH, this is so obscure that I don't mind the enum names being obscure as well.

SuchAFuriousDeath: What abour the rest of the changes? :sweat_smile:

core/src/avm1/property_decl.rs
```diff
@@ -19,7 +55,76 @@ impl<'gc> HasStringContext<'gc> for DeclContext<'_, 'gc> {
...
+        let object = build_object(&mut ctx);
+        let function = build_function(&mut ctx);
+
+        ctx.function_class = function.constr;
+
+        // `PropertyMap`'s `OccupiedEntry::insert` uses `mem::replace` at the
+        // existing index, so the placeholder values written above are swapped
+        // in-place for the real `Function` class object without disturbing
+        // iteration order.
+        let constructor_name = istr!(ctx.strings, "constructor");
```
kjarosh, line 55: This comment refers to an implementation detail. Referring to implementation details across the codebase doesn't make sense, because they can change.

core/src/avm1/property_decl.rs
```diff
@@ -19,7 +55,76 @@ impl<'gc> HasStringContext<'gc> for DeclContext<'_, 'gc> {
...
+    /// `Function` classes via the supplied builders.
+    ///
+    /// During the two builder calls, `function_class` holds a placeholder
+    /// (`object_proto`) so that `build_class_object` always has a value to
+    /// write into the `constructor` slot. Once `Function` exists,
+    /// `function_class` is set to the real class object and both bootstrap
+    /// classes' `constructor` slots are rewritten in-place: `PropertyMap`'s
+    /// `OccupiedEntry::insert` uses `mem::replace` at the existing index, so
+    /// the placeholder value is swapped for the real one without disturbing
+    /// iteration order.
```
kjarosh: You're describing implementation detail in a public doc.

core/src/avm1/property_decl.rs
```diff
@@ -1,16 +1,52 @@
...
     pub object_proto: Object<'gc>,
     pub fn_proto: Object<'gc>,
+    /// The `Function` class object, used to populate `constructor` as an own
+    /// property on every system class object built through this context.
+    ///
+    /// While bootstrapping `Object` and `Function` themselves, this holds a
+    /// placeholder; [`Self::new`] rewrites the placeholder to the real
+    /// `Function` class object on both bootstrap classes once `Function` has
+    /// been built. Subsequent `class` / `native_class` / `empty_class` calls
+    /// see the real value.
```
kjarosh: I feel like it's the third time I'm reading the same thing. Duplicating docs and comments has the same impact as duplicating code.

core/src/avm1/function.rs
```diff
@@ -498,6 +499,61 @@ impl<'gc> FunctionObject<'gc> {
...
+    /// Builds a system class object. Unlike [`Self::build`], the resulting
+    /// object's own properties are inserted in Flash's order so that AVM1's
+    /// newest-first `for-in` matches the original `playerglobal` build order.
+    ///
+    /// `order` picks between the two orderings Flash actually uses:
+    ///
+    /// - [`PrototypeOrder::First`] inserts `prototype, constructor, __proto__`
+    ///   on `obj`. Iteration: `__proto__, constructor, prototype`.
+    /// - [`PrototypeOrder::Last`] inserts `constructor, __proto__, prototype`
+    ///   on `obj`. Iteration: `prototype, __proto__, constructor`.
```
kjarosh: It's better to describe it in `PrototypeOrder`'s docs

core/src/avm1/function.rs
```diff
@@ -498,6 +499,61 @@ impl<'gc> FunctionObject<'gc> {
...
+    ///
+    /// The supplied `prototype` also has its `constructor` slot defined as
+    /// the first own property (assuming the caller hasn't already added any),
+    /// matching Flash's iteration `..., __proto__, constructor`.
```
kjarosh: Sure, but that's covered by tests right? Maintenance costs also apply to documentation.

core/src/avm1/property_decl.rs
```diff
@@ -1,16 +1,52 @@
...
+/// - [`Self::First`] — `prototype` is inserted first, then `constructor`, then
+///   `__proto__`. Iteration: `__proto__, constructor, prototype`. Most classes
+///   use this (e.g. `TextSnapshot`, `PrintJob`, `MovieClipLoader`, all
+///   `flash.filters.*`, `flash.geom.Transform`, etc.).
+/// - [`Self::Last`] — `constructor` is inserted first, then `__proto__`, then
+///   `prototype`. Iteration: `prototype, __proto__, constructor`. Used by
+///   `FileReference`, `FileReferenceList`, `flash.geom.Matrix`,
+///   `flash.geom.Rectangle`, `flash.automation.Configuration`,
+///   `flash.automation.ActionGenerator`, and
+///   `System.security.PolicyFileResolver`.
```
kjarosh: Dejavu

kjarosh:
I guess the code is fine, but the docs look like AI thought process dump, which will incur maintenance costs to us later (and they are too verbose).

kjarosh:
@moulins could you take a look at it as well?

kjarosh (approved):

core/src/avm1/function.rs
```diff
@@ -498,6 +499,46 @@ impl<'gc> FunctionObject<'gc> {
...
+        fn_proto: Object<'gc>,
+        prototype: Object<'gc>,
+        function_class: Object<'gc>,
+        order: PropertyOrder,
+    ) -> Object<'gc> {
+        let obj = Object::new_without_proto(context.gc());
+        let native = NativeObject::Function(Gc::new(context.gc(), self));
+        obj.set_native(context.gc(), native);
+
+        let attr = Attribute::DONT_ENUM | Attribute::DONT_DELETE;
```
moulins, line 516: Are we sure every property is `DONT_DELETE | DONT_ENUM` here? In `fn build()` just above, `constructor` is only marked `DONT_ENUM`.

core/src/avm1/property_decl.rs
```diff
@@ -19,7 +31,51 @@ impl<'gc> HasStringContext<'gc> for DeclContext<'_, 'gc> {
     }
 }
 
-impl<'gc> DeclContext<'_, 'gc> {
+impl<'a, 'gc> DeclContext<'a, 'gc> {
+    pub fn new(
```
moulins, line 35: I'm not sure this is the cleanest way of initializing `Object/Function` now, but it's good enough I guess.

moulins (approved):

core/src/avm1/function.rs
SuchAFuriousDeath: It looks that this is correct and fn build is where it is wrong. Judging from tests/tests/swfs/avm1/global_proto_decls_delete/output.txt

SuchAFuriousDeath:
oopsie.

PR #23894, opened 2026-06-03
frontend-utils: Use native TLS instead of rustls

kjarosh:
## Description

This makes the binary smaller as it uses the system component for TLS,
without bundling it with Ruffle.

## Testing

Tested locally by loading a remote movie.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -870,7 +870,7 @@ dependencies = [
  "bitflags 1.3.2",
  "core-foundation 0.9.4",
  "core-graphics-types 0.1.3",
- "foreign-types",
+ "foreign-types 0.5.0",
  "libc",
 ]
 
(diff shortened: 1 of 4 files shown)


SuchAFuriousDeath (approved):

Cargo.lock
```diff
@@ -5164,12 +5171,12 @@ dependencies = [
 
 [[package]]
 name = "security-framework"
-version = "3.7.0"
+version = "2.11.1"
```
torokati44, line 5174: hmmm 🤔

Cargo.lock
```diff
@@ -4192,61 +4269,6 @@ dependencies = [
  "memchr",
 ]
 
-[[package]]
-name = "quinn"
```
torokati44, line 4196: this is nice

kjarosh:
Discussed on Discord that it's the right direction

torokati44:
:|

https://github.com/ruffle-rs/ruffle-android/actions/runs/27267838836/job/80529358968?pr=762

torokati44:
I suppose this could be turned into a proxy feature of the frontend_utils crate, so desktop and android could choose a different TLS backend?

Or perhaps consider not even relying on reqwest at all on Android, but that might be difficult.

torokati44:
Hmm, maybe native-tls _is_ supposed to work on Android, just not with OpenSSL, but platform-native stuff? :thinking:

PR #23896, opened 2026-06-04
ci: Install pinned Chrome to workaround issue

danielhjacobs:
## Description

Workaround https://github.com/actions/runner-images/issues/14169

## Testing

See if the Chrome tests succeed on CI (testing right now)

## Checklist

### DIFF
--- a/web/packages/selfhosted/wdio.conf.ts
+++ b/web/packages/selfhosted/wdio.conf.ts
@@ -21,14 +21,25 @@ if (chrome) {
     const args = ["--disable-gpu", "--enable-unsafe-swiftshader"];
     if (headless) {
         args.push("--headless");
+        if (process.env["CHROME_BIN"]) {
+            args.push("--no-sandbox", "--disable-dev-shm-usage");
+        }
     }
-    capabilities.push({
+    const chromeOptions: { args: string[]; binary?: string } = { args };
+    if (process.env["CHROME_BIN"]) {
+        chromeOptions.binary = process.env["CHROME_BIN"];
+    }
+    const chromeCaps: WebdriverIO.Capabilities = {
         "wdio:maxInstances": maxInstances,
         browserName: "chrome",
-        "goog:chromeOptions": {
-            args,
-        },
-    });
+        "goog:chromeOptions": chromeOptions,
+    };
+    if (process.env["CHROMEWEBDRIVER"]) {
+        chromeCaps["wdio:chromedriverOptions"] = {
+            binary: process.env["CHROMEWEBDRIVER"],
+        };
+    }
+    capabilities.push(chromeCaps);
 }
 
 if (edge) {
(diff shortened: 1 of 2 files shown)


torokati44:
I'm mostly on a similar stance with this as with https://github.com/ruffle-rs/ruffle/pull/23894 (discussed on Discord). This has a tangible benefit (makes CI green), but will have to be reverted later (and not forgotten), and it bothers me that we have to make workarounds because of a mishap upstream (on at least two levels: GHA with deployment and GOOG with regression)...
The technical part looks fine, so if anyone else drops an approve: fine by me.
Thank you for figuring out the details!
(Sorry if I'm being annoying with my indecisiveness lately.)

torokati44:
BTW is there no fixed newer version of Chrome yet?

danielhjacobs:
I don't even know if Chrome knows about the issue.

danielhjacobs:
We'd need a minimal repro to report it: https://issues.chromium.org/issues/517079992#comment9

web/packages/selfhosted/wdio.conf.ts
```diff
@@ -21,14 +21,25 @@ if (chrome) {
     const args = ["--disable-gpu", "--enable-unsafe-swiftshader"];
     if (headless) {
         args.push("--headless");
+        if (process.env["CHROME_BIN"]) {
+            args.push("--no-sandbox", "--disable-dev-shm-usage");
```
kjarosh, line 25: Why are we setting these flags when the binary is configured?

danielhjacobs: When the binary is used, it runs as root. Chrome blocks itself from launching with the sandbox enabled while logged in as root, so we have to disable the sandbox.

https://docs.travis-ci.com/user/chrome#sandboxing

Also, the /dev/shm partition is capped at 64MB, which is not enough, so we disable using it.

https://intricatecloud.io/2019/05/running-webdriverio-tests-using-headless-chrome-inside-a-container/#--disable-dev-shm-usage

danielhjacobs: I did try various versions of this commit, including the same changes but without these lines. From what I can tell this is the minimum working change.

danielhjacobs:
> I'm mostly on a similar stance with this as with #23894 (discussed on Discord). This has a tangible benefit (makes CI green), but will have to be reverted later (and not forgotten), and it bothers me that we have to make workarounds because of a mishap upstream (on at least two levels: GHA with deployment and GOOG with regression)... The technical part looks fine, so if anyone else drops an approve: fine by me. Thank you for figuring out the details! (Sorry if I'm being annoying with my indecisiveness lately.)

3 weeks later, with no fix in sight, I'm tempted to merge

kjarosh (approved):

danielhjacobs:
Will open a revert immediately and try to remember to rebase it once a week, so we know when this becomes unnecessary.

PR #23897, opened 2026-06-04
core: Memory-leak fixes for long-running ActionScript 3 workloads

falpi:
## Summary

Five coordinated patches that bring Ruffle's AS3 implementation in line with the
documented Flash semantics on three contracts that previously had no
implementation or only a no-op stub:

- **Weak-keyed `Dictionary`** &mdash; `new Dictionary(true)` must hold object
  keys weakly.
- **Weak event listeners** &mdash; the `useWeakReference` flag of
  `addEventListener` must not extend the listener's lifetime.
- **Explicit GC hint** &mdash; `flash.system.System.gc()` must request a full
  collection cycle.

These gaps do not cause functional regressions on short-lived SWFs (Flash
games, demos), but they accumulate observable leaks and progressive slow-downs
on long-running AS3 workloads &mdash; most visibly on enterprise applications
built on the Adobe Flex 4.x SDK, whose data-binding and renderer-cache
machinery is load-bearingly dependent on the three contracts above.

## Fix list

1. **Dictionary `weakKeys`** &mdash; real weak-keyed `flash.utils.Dictionary`
   storage with the enumeration overrides required for `for...in` /
   `for each` over weak entries. Also unblocks the boot of Flex applications
   that depend on `StyleManagerImpl` iterating a `Dictionary(true)`.
2. **`HandlerRef` Strong/Weak** &mdash; `addEventListener`'s
   `useWeakReference=true` is now honored via a `Strong | Weak` enum stored in
   the dispatch list; pointer identity is preserved across variants so a
   strong `removeEventListener` still removes a weak registration.
3. **`DispatchList` add-time prune** &mdash; dead weak entries are pruned at
   `addEventListener` time (not only during dispatch), so event sources that
   accept subscribers but rarely dispatch (Flex's `ResourceManager.change`,
   for example) no longer accumulate unbounded dead handler slots.
4. **`Dictionary` set-time prune** &mdash; on weak-keyed `Dictionary` writes,
   dead-key entries are dropped in the same pass that scans for an existing
   key to replace. Without this, write-mostly weak Dictionary caches
   (e.g. Flex's `EmbeddedFontRegistry.cachedFontsForObjects`) retain the
   value side of every dead entry for the lifetime of the dictionary.
5. **`System.gc()` real implementation** &mdash; the AS3 hint is wired to
   `gc-arena`'s `finish_cycle()` via an `Avm2::force_full_gc` flag consumed by
   `Player::update`. Repeated calls within a single update window are
   idempotent. This is the enabling fix: incremental tracing alone does not,
   under sustained allocation pressure, drain the references that Fix
   1&#x2013;4 free quickly enough; an explicit AS3-level trigger gives the
   application a deterministic full-cycle hook at points of its choice
   (post-teardown, navigation transitions, etc.).

The five are mutually dependent and should land as a set: any subset leaves a
measurable residual leak.

## Empirical impact

Verified on a Flex 4.6 enterprise application with a navigator-driven
form-reload workload (~100k XML rows per reload). With the five patches
applied:

- Peak working set: bounded plateau (was: unbounded growth ending in OOM after
  a handful of reload cycles).
- Per-reload cost: within an animation-frame budget (was: progressive seconds-
  scale stalls).
- Without Fix 5: out-of-memory after a handful of cycles.
- Without Fix 4: linear growth of the cached-value side of weak-keyed
  dictionaries (~+100 retained values per reload on the test workload).
- Without Fix 1 or Fix 2: application either fails at boot
  (`ReferenceError #1009` from `StyleManagerImpl`) or accumulates retained
  form trees indefinitely.

## Documentation

A full technical writeup &mdash; per-fix problem statement, root cause, code
mechanism, files touched, and verification approach &mdash; is in this pdf:
[fix-memory-leaks.pdf](https://github.com/user-attachments/files/28606576/fix-memory-leaks.pdf)

The commit `chore: adapt to Rust beta clippy lints` on top of the main commit
addresses the new lints reported by the "Lints with Rust beta" CI gate
(`clippy::for_kv_map` in our new code; `clippy::needless_return` and
`clippy::unused_async` in two unrelated upstream files that the same gate
rejects on every open PR until they're fixed).

## Checklist

### DIFF
--- a/core/src/avm2.rs
+++ b/core/src/avm2.rs
@@ -190,6 +190,12 @@ pub struct Avm2<'gc> {
     pub debug_output: bool,
 
     pub optimizer_enabled: bool,
+
+    /// Set to `true` by `flash.system.System.gc()`. Read and cleared by
+    /// `Player::update` after each frame's incremental GC pass; when set,
+    /// a full collection cycle is performed.
+    #[collect(require_static)]
+    pub force_full_gc: std::cell::Cell<bool>,
 }
 
 impl<'gc> Avm2<'gc> {
@@ -240,6 +246,8 @@ impl<'gc> Avm2<'gc> {
             debug_output: false,
 
             optimizer_enabled: true,
+
+            force_full_gc: std::cell::Cell::new(false),
         }
     }
 
--- a/core/src/avm2.rs
+++ b/core/src/avm2.rs
@@ -191,6 +191,12 @@ pub struct Avm2<'gc> {
     pub debug_output: bool,
 
     pub optimizer_enabled: bool,
+
+    /// Set to `true` by `flash.system.System.gc()`. Read and cleared by
+    /// `Player::update` after each frame's incremental GC pass; when set,
+    /// a full collection cycle is performed.
+    #[collect(require_static)]
+    pub force_full_gc: std::cell::Cell<bool>,
 }
 
 impl<'gc> Avm2<'gc> {
@@ -241,6 +247,8 @@ impl<'gc> Avm2<'gc> {
             debug_output: false,
 
             optimizer_enabled: true,
+
+            force_full_gc: std::cell::Cell::new(false),
         }
     }
 
--- a/core/src/avm2/events.rs
+++ b/core/src/avm2/events.rs
@@ -4,11 +4,11 @@ use crate::avm2::Avm2;
 use crate::avm2::activation::Activation;
 use crate::avm2::function::FunctionArgs;
 use crate::avm2::globals::slots::flash_events_event_dispatcher as slots;
-use crate::avm2::object::{EventObject, FunctionObject, Object, TObject as _};
+use crate::avm2::object::{EventObject, FunctionObject, FunctionObjectWeak, Object, TObject as _};
 use crate::display_object::TDisplayObject;
 use crate::string::AvmString;
 use fnv::FnvHashMap;
-use gc_arena::Collect;
+use gc_arena::{Collect, Gc, GcWeak, Mutation};
 use std::collections::BTreeMap;
 use std::hash::{Hash, Hasher};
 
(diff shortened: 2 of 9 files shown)


torokati44:
If there's a five item long list of separate problems fixes, it stands to reason that the PR should consist of at least five separate commits...? 🤔

EDIT: Ah, I see you(?) even wrote "Five coordinated patches" in the description, but it's not true...

docs/fix-memory-leaks.pdf
torokati44, line 1: There's no point to adding this here in this form and format.

frontend-utils/src/bundle/exporter/fs_helper.rs
torokati44, line 1: How is this change related to any of the claimed fixes? 🤔

desktop/src/gui/theme.rs
torokati44, line 1: (same here)

falpi:
This patch (like my others) is primarily engineered by an AI tool. I'm not very familiar with Rust and I'm still very limited with Ruffle's structure, so I rely on the agent's analysis, which I try to guide and test extensively. This patch saved my life because I have an enterprise application that would crash without these (and other future refinements). There may be some inaccuracies, and I hope you can fix them, but overall it seems to work because it brilliantly solved the memory leak issues after many hours of development and intense testing.

docs/fix-memory-leaks.pdf
falpi: > There's no point to adding this here in this form and format.

When I produce complex patches like these, I always ask the agent to document the causes and proposed solutions in detail, hoping to make it easier for the maintainers to check. I apologize if it's misplaced, but there's always time to delete it.

frontend-utils/src/bundle/exporter/fs_helper.rs
falpi: > How is this change related to any of the claimed fixes? 🤔

The agent proposed this fix to avoid a pre-existing formatting error in the code that would otherwise cause tests to fail. It has nothing to do with the patch.

desktop/src/gui/theme.rs
falpi: > (same here)

Same thing as above: the agent proposed this fix to avoid a pre-existing formatting error in the code that would otherwise cause tests to fail. It has nothing to do with the patch.

docs/fix-memory-leaks.pdf
falpi: I'll remove the docs folder and the PDF from the branch and attach it inline to the PR description.

frontend-utils/src/bundle/exporter/fs_helper.rs
torokati44: The change itself might be okay, and warranted even - it just doesn't seem like it belongs here. If not a separate PR, I feel like it should at least be a separate commit.

falpi: I removed the file from my commit. Please fix them yourself, otherwise all PRs will systematically fail.

desktop/src/gui/theme.rs
falpi: I removed the file from my commit. Please fix them yourself, otherwise all PRs will systematically fail.

falpi:
Hi @torokati44 for days now, the "Test in Chrome" check has been failing because it consistently shows "cancelled" on all PRs. Is anyone working on this?

torokati44:
https://github.com/actions/runner-images/issues/14169
probably not

falpi:
Hi @torokati44 ... this PR is tagged with "Changes requested". What does it mean?

torokati44:
I don't see that tag anywhere - I deliberately on left "comment" style reviews.
At any rate, since the commit message still explicitly lists 5 separate fixes, I still think (like in my first comment), it should be split into 5 distinct commits, for each.

falpi:
> I don't see that tag anywhere - I deliberately on left "comment" style reviews. At any rate, since the commit message still explicitly lists 5 separate fixes, I still think (like in my first comment), it should be split into 5 distinct commits, for each.

For now, I have to finalize several developments and the related PRs. When I have some time, I'll try to adjust the code.

Regarding the tag, I'm referring to this message I get from the PR summary screen:
immagine

PR #23899, opened 2026-06-04
avm1: Stub several classes and properties

SuchAFuriousDeath:
## Description

Just some missing stuff.

## Testing

_How can we test this PR?_

Test changes are included :heavy_check_mark: 

## Checklist


core/src/avm1/globals/asset_cache.rs
```diff
@@ -0,0 +1,37 @@
...
+    class
+}
+
+fn get_disk_usage<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    _this: Object<'gc>,
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    avm1_stub!(activation, "AssetCache", "getDiskUsage");
+    Ok(0.into())
```
kjarosh, line 27: Shouldn't this be called by argstest? 🤔 cov says it's uncovered

SuchAFuriousDeath: I'll investigate.

SuchAFuriousDeath: Ok apparently argstest does not visit static methods on the class functions. What do we do then?

SuchAFuriousDeath: Ok I take it back, this looks a little more complicated, I'll be back with more info.

kjarosh: I think it's fine. Would be nice if we had a similar test that covers those methods. (We cannot modify argstest as it's from gnash.)

SuchAFuriousDeath: Do I have to add the test to get this PR merged?

kjarosh: Nope!

kjarosh (approved):

PR #23901, opened 2026-06-04
avm2: Fix setProgramConstantsFromByteArray implementation

SuchAFuriousDeath:
## Description

Turns out Flash ignores the endianess and always reads everything as little endian. Hence, fix the BE test (which previously produced garbage output in Flash) to write data in little endian. The BE test now produces the same result as the LE one in Flash.

Bonus: I renamed set_program_constants_from_matrix to set_program_constants and Context3DCommand::SetProgramConstantsFromVector to SetProgramConstants, it's more consistent that way. And also there is no need to copy the floats into a Vec in the setProgramConstantsFromByteArray impl, but we have to pass `&[[u8; 4]]` instead of `&[f32]` around in case `byte_offset` isn't a multiple of four, because that would ~produce an unaligned slice~ panic in bytemuck::cast_slice. 

## Checklist


tests/tests/swfs/avm2/stage3d_program_constants_bytearray_be/test.swf
kjarosh, line 1: Can we add Flash Player's output to the test?

SuchAFuriousDeath: This is a visual test, the LE version also has an empty output.txt.

kjarosh: Yes, I meant the visual output

SuchAFuriousDeath: Both already do have it. It's unchanged in this PR, because it produces the same result as the LE version and the image was from there.

SuchAFuriousDeath: Hmm wait. I guess it wasn't from Flash Player, I'll try capturing it.

kjarosh: I would be very surprised if it was from Flash Player because it's low quality and 0 outliers, which is specific to Ruffle's output (FP's output would have a lot of outliers with low quality).

SuchAFuriousDeath: Yeah, that's right. I'll modify the test slightly though, because having the cube rotate makes it impossible to capture at the right time :D

tests/tests/swfs/avm2/stage3d_program_constants_bytearray_be/output.expected.png
ChrisCPI, line 1: If this is FP's output, shouldn't the Stage3D be slightly offset from the top-left, since the x/y is set to 10? Same with the other output.

SuchAFuriousDeath: Should be better now.

ChrisCPI (approved):
Output images LGTM

kjarosh (approved):

PR #23921, opened 2026-06-08
tests: Add fuzz tests for Ruffle

kjarosh:
## Description

Add a crate dedicated for fuzzing Ruffle under fuzz/.

It uses a separate workspace in order not to pollute the main lock file (that's the recommended approach).  The ruffle-fuzz crate supports both libFuzzer and AFL++ via cargo-fuzz and cargo-afl.

Currently it has only one target for both fuzzers: parse_swf, which checks for crashes when parsing SWFs.

In general, it's not trivial to use fuzzers, often you have to pass the binary path, the corpus path, or feature config.  The fuzzer tool makes it easier to list, build, and run fuzz tests irrespective of the fuzzer used (libFuzzer vs AFL++).  In addition it supports preparing the fuzz corpus by collecting all SWF tests.

This also includes a Quck Fuzz CI workflow that runs fuzz tests for a short period of time for each PR.

Things that will need improving for sure:

1. Currently fuzzing fails very quickly because Ruffle crashes a lot.
2. The corpus quality is very low, because we're just using all SWF tests for fuzzing. Ideally we would have a separate corpus for each target, but that would need a lot of binary objects and it could prove problematic without LFS.
3. No long-time fuzzing for now. Maybe when we're more stable when it comes to fuzzing.

## Testing

Yep.

## Checklist


torokati44:
I love the idea!
But I have questions (I'm not that familiar with fuzzing, bear with me):
 - Is this something that should really be run on PRs, or rather/additionally, frequently (even quasi-continuously?) on `master`?
 - If on PRs, is there any way to try guiding the fuzzers to exercise the code paths touched by the diff?
 - Only SWF is (planned to be) fuzzed, or externally loaded media as well?

kjarosh:
> Is this something that should really be run on PRs, or rather/additionally, frequently (even quasi-continuously?) on master?

Ideally both, the issue with running it continuously is mainly timeouts on GHA and difficulty in keeping the corpus/queue up to date. Without that, we would be fuzzing the same thing all over again instead of detecting new paths in code.

That's why there are projects like OSS-fuzz or ClusterFuzz that help with that.

However, I just wanted to start with something small here and then grow (e.g. when we fix all the crashes we can find in those 5 minutes).

> If on PRs, is there any way to try guiding the fuzzers to exercise the code paths touched by the diff?

That is a great question. I don't know, but that's something we can try pursuing. I don't see a trivial way of doing that however (but apparently there are ways).

> Only SWF is (planned to be) fuzzed, or externally loaded media as well?

We can fuzz whatever we want. It's enough to add a new fuzzing target. Currently there's only one for parsing SWFs, but there can be one for executing them, for parsing ABC, etc. This PR explicitly adds support for multiple targets.

torokati44:
> Ideally both, the issue with running it continuously is mainly timeouts on GHA and difficulty in keeping the corpus/queue up to date.

Sure, I was thinking something like "roughly 10 minutes every day". (Which can then be increased until GH slaps on our wrists. 😅 - Not that we were ever shy with burning their cycles...)

> Without that, we would be fuzzing the same thing all over again instead of detecting new paths in code.

I was under the impression that these things are mostly stochastic by nature, so perhaps never exactly the same thing...? (I should probably read a bit more into how the mentioned tools work.)

kjarosh:
> I was under the impression that these things are mostly stochastic by nature, so perhaps never exactly the same thing...?

Mutating input purely randomly produces a very, very, very small change of hitting new code and a crash. That's why fuzzers use :sparkles:_magic_:sparkles: (at least that's my current understanding) to detect new code paths, assess whether two inputs are similar (trigger similar code paths), etc. They do it by keeping a temporary corpus (called a _queue_ in AFL++) which contains inputs worth focusing on. Keeping that information across runs, and sometimes committing inputs to the persistent corpus, we can improve fuzzing performance.

kjarosh:
And also: I managed to fix all crashes so that the 5 min fuzz passes. :tada: (for now)

Dinnerbone:
I think I'd appreciate some readme.md with instructions on how to use it. `cargo fuzzer run` just gives a cryptic "error: no such command: `afl`", which I can piece together based off the github actions but most won't have the context of this PR.

kjarosh:
Done, added some docs and made the tool smarter

Lord-McSweeney (approved):

tools/fuzzer/src/lib.rs
```diff
@@ -0,0 +1,441 @@
...
+    for name in &names {
+        let supported = &all[name].supported_fuzzers;
+
+        let corpus = corpus_for(name);
+
+        for f in [Fuzzer::Afl, Fuzzer::Libfuzzer] {
+            if !fuzzer.is_empty() && !fuzzer.contains(&f) {
+                // Filtered out.
+                continue;
+            }
```
moulins: I'd like a comment explaining the logic here. Why not a straightforward `for f in fuzzers`? (from what I can see, it's to weed out duplicates, but this is IMHO a strange way of doing that)

kjarosh: Simplified the logic.

.github/workflows/quick_fuzz.yml
```diff
@@ -0,0 +1,66 @@
...
+        # actions/upload-artifact errors out on : in paths, which is used by
+        # AFL++, so we need to compress it manually.
+        env:
+          # Exclude the queue directory which contains the transient corpus.
+          TAR_OPTS: --exclude=fuzz/out/*/afl/*/queue
+        run: tar -czf fuzz-artifacts.tar.gz $TAR_OPTS fuzz/out/ || true
+
+      - name: Upload artifacts
+        if: always()
+        uses: actions/upload-artifact@v4
```
torokati44, line 62: This is fairly old, and GHA complains that it uses deprecated Node.js 20.
No worries, Dependabot will take care of it shortly.

torokati44:
Do these outputs (by afl) mean ... anything?

> [!] WARNING: Some test cases are huge (1.00 MB) - see docs/fuzzing_in_depth.md#i-improve-the-speed
> [!] WARNING: Some test cases look useless. Consider using a smaller set.
> [!] WARNING: You probably have far too many input files! Consider trimming down.

Also, should the fuzzer jobs perhaps be gated on changes to just Rust files, like the Rust and web tests?

kjarosh:
> Do these outputs (by afl) mean ... anything?

Yes, see point 2 in the description.

> Also, should the fuzzer jobs perhaps be gated on changes to just Rust files, like the Rust and web tests?

Yeah, probably a good idea, but doesn't matter that much because they are stochastic anyway so they do provide some value even if rust code doesn't change. (But it usually does.)

PR #23923, opened 2026-06-08
avm2: Add a field to ops instead of adding separate integral variants

Lord-McSweeney:
## Description

Applies to `Op::Add` and `Op::Subtract`. As per moulins's advice, instead of adding separate `*Integral` variants of those two ops, we can just set a field on them.

## Testing

No observable changes.

## Checklist

### DIFF
--- a/core/src/avm2/op.rs
+++ b/core/src/avm2/op.rs
@@ -11,8 +11,11 @@ use std::cell::Cell;
 #[derive(Clone, Collect, Copy, Debug)]
 #[collect(no_drop)]
 pub enum Op<'gc> {
-    Add,
-    AddIntegral,
+    Add {
+        /// Whether the two inputs to this op are both guaranteed to be
+        /// integers. This field is used in the optimizer.
+        inputs_integral: bool,
+    },
     AddI,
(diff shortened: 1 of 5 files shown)


SuchAFuriousDeath (approved):

core/src/avm2/op.rs
```diff
@@ -327,8 +328,9 @@ pub enum Op<'gc> {
     StoreLocal {
         index: u32,
     },
-    Subtract,
-    SubtractIntegral,
+    Subtract {
+        inputs_integral: bool,
```
kjarosh, line 335: Nit: maybe just using `integral` would be enough? It's shorter and I cannot think of another meaning (the operation itself would be integral then).

kjarosh: Hmm, maybe it's less obvious because you could have integral addition on non-integral inputs... Maybe that's something that could be mentioned in a doc over this field? Up to you to decide

Lord-McSweeney: Good point, I'll add a comment

kjarosh (approved):

PR #23930, opened 2026-06-09
🤖 avm2: Implement FileReference.download() (close #11168)

### LINKED ISSUE #11168: Feature to download a file download requests from a game on the demo page
Can you add the feature to let the demo page download a requested file? This would be useful if a game needed to download a file to save progress on the demo page, since the demo is useful for mobile.

MavenRain:
Implements the previously-stubbed AVM2 `flash.net.FileReference.download()` by porting Ruffle's existing, merged AVM1 download flow to AVM2 event semantics — closing the AVM1/AVM2 parity gap reported in #11168 (e.g. Scratch 2.0 AS3 games).

### What it does
- `download(request, defaultFileName)` builds the platform request via the shared `request_from_url_request` helper (so method, headers, and data are honored like `Loader.load`/`URLLoader.load`), then prompts for a save location and fetches the URL, writing the response to the chosen file.
- An empty/omitted `defaultFileName` is derived from the request URL's final path segment.
- Dispatches `select`/`open`/`progress`/`complete` on success, `cancel` on dialog cancellation, and `ioError` on failure — matching the event sequences of the existing AVM1 `download_file_dialog` implementation. (No `httpStatus` event, mirroring the AVM1 behavior; happy to add it if preferred.)
- Scoped to `download()` only; `upload()`/`uploadUnencoded()` remain stubs.

### Testing
Adds `avm2/filereference_download_{success,cancel,httperror_status_code,httperror_dns_error}`, mirroring the AVM1 `file_reference_download_*` suite and driven entirely by the test navigator/UI mocks (no Flash Player needed). All pass under `compile-and-verify`. No regressions in the existing AVM1/AVM2 FileReference, URLLoader, or loader suites.


kjarosh:
> no Flash Player needed

I don't understand. Flash Player is definitely needed, that's how you reverse engineer. You can't just mock calls and call it a day. The point is to behave the same way as FP.

core/src/avm2/globals/flash/net/file_reference.rs
```diff
@@ -281,3 +281,55 @@ pub fn save_internal<'gc>(
...
+    let this = this.as_object().unwrap();
+
+    let this = this.as_file_reference().unwrap();
+
+    let url_request = args
+        .try_get_object(0)
+        .ok_or_else(|| make_error_2007(activation, "request"))?;
+
+    // Build the platform request, honoring the URLRequest's method, headers and
+    // data exactly like `Loader.load`/`URLLoader.load` do.
```
kjarosh, line 299: Great output for an agent, poor comment. Either acknowledge it, or act upon it (deduplicate code), in any case it doesn't really add anything here.

What it says is that your agent implemented this based on those methods. How do you know it's the right way to implement it?

MavenRain:
@kjarosh   Fair point, and I should have worded that line better.  I'm not claiming to have reverse-engineered the behavior from mocks.  The event sequence here (select, open, progress, complete on success, plus the error and cancel paths) is ported directly from the existing AVM1 FileReference.download implementation in loader.rs (download_file_dialog), and the tests use the same mock-navigator pattern as the existing avm1/file_reference_download_* suite.  So what they pin down is that the AVM2 port matches the already-merged AVM1 behavior, not a fresh guess.

Where I genuinely couldn't infer behavior from the AVM1 port, I left it unasserted on purpose.  For example, I didn't add an httpStatus event, since the AVM1 path doesn't fire one and I had no way to confirm whether FP does.

That said, your point stands: for anything beyond AVM1 parity, the right way to settle it is against a real player. @danielhjacobs offered to walk me through the Flash Player Debugger setup on #23723, so I'm going to get that running, verify the download event sequence (and the httpStatus question) against FP, and update this PR's test to match what FP actually does.  Does that address the concern?

kjarosh:
Yes, in general the majority of contributions to Ruffle require testing how stuff works in Flash Player. You cannot assume how AVM2 works based on how AVM1 works, the behavior can be different, we need to explicitly test it.

And that's why we're adding SWF tests, they allow us to assert on how Flash Player works.

Lord-McSweeney:

PR #23937, opened 2026-06-09
core: Use __jcvt() intrinsic on AArch64 for f64→i32 conversion

SuchAFuriousDeath:
Supersedes #22138.

## Description
See #22138.

## Testing

_How can we test this PR?_

Good question :D. I guess someone has to have the hardware or we have to check if it's available in CI.

## Checklist

### DIFF
--- a/core/src/ecma_conversions.rs
+++ b/core/src/ecma_conversions.rs
@@ -40,49 +40,22 @@ pub fn f64_to_wrapping_u32(n: f64) -> u32 {
 /// Converts an `f64` to an `i32` with ECMAScript `ToInt32` wrapping behavior.
 /// The value will be wrapped in the range [-2^31, 2^31).
 pub fn f64_to_wrapping_i32(n: f64) -> i32 {
-    // TODO: use core intrinsic when https://github.com/rust-lang/rust/issues/147555 is stabilized.
-    #[cfg(target_arch = "aarch64")]
-    {
-        if std::arch::is_aarch64_feature_detected!("jsconv") {
-            // SAFETY: `jsconv` feature is checked in both compile time and runtime to be existed, so it's safe to call.
-            unsafe { f64_to_wrapping_int32_aarch64(n) }
-        } else {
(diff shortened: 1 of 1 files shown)


torokati44:
Do Android phones have this instruction? In that case, the app might de facto test it. 🤔

torokati44:
Looks like they do:



So I guess it will end up being used automatically.

kjarosh:
> we have to check if it's available in CI

It is available in CI, so tests cover it.

SuchAFuriousDeath:
So is there anything blocking this PR?

kjarosh:
Someone needs to review it :D

kjarosh (approved):
LGTM, we already use 1.95 features, and it's tested in CI. Comments from the previous PR have been addressed.

@SuchAFuriousDeath Please add yourself as a co-author

core/src/ecma_conversions.rs
```diff
@@ -40,51 +40,24 @@ pub fn f64_to_wrapping_u32(n: f64) -> u32 {
...
 pub fn f64_to_wrapping_i32(n: f64) -> i32 {
-    // TODO: use core intrinsic when https://github.com/rust-lang/rust/issues/147555 is stabilized.
-    #[cfg(target_arch = "aarch64")]
-    {
-        if std::arch::is_aarch64_feature_detected!("jsconv") {
-            // SAFETY: `jsconv` feature is checked in both compile time and runtime to be existed, so it's safe to call.
-            unsafe { f64_to_wrapping_int32_aarch64(n) }
-        } else {
-            f64_to_wrapping_i32_generic(n)
+    cfg_select! {
```
linkmauve, line 43: Why do you use a macro instead of the existing `#[cfg]` block here?

kjarosh: `cfg_select!` is a 1.95 feature that makes it easier to write conditional code, we don't have to duplicate the condition in both branches.

linkmauve: I see, thanks!

PR #23940, opened 2026-06-09
core: Use Vec::retain in Avm1TextFieldBinding::bind_variables

SuchAFuriousDeath:
## Description

Just a tiny refactor, std::mem::take has to be used to satisfy the borrow checker.

SIDENOTE: The TODO talks about Vec::drain_filter, which got stabilized as Vec::extract_if, but retain is the better fit here, since there is no resulting iterator we have to consume.

## Checklist


kjarosh:
Just a general note about this particular code (text variable binding): it's generally untested and buggy in Ruffle, it can easily introduce regressions.

SuchAFuriousDeath:
@kjarosh :pleading_face:

core/src/display_object.rs
```diff
@@ -3041,19 +3041,11 @@ pub struct Avm1TextFieldBinding<'gc> {
...
-            {
-                activation.context.unbound_text_fields.swap_remove(i);
-                len -= 1;
-            } else {
-                i += 1;
-            }
-        }
+        let mut text_fields = std::mem::take(activation.context.unbound_text_fields);
+
+        text_fields.retain(|field| !field.try_bind_text_field_variable(activation, false));
```
kjarosh, line 3046: This will change the order of elements, does it matter?

SuchAFuriousDeath: Good question, I didn't consider that. From looking at the code, I don't see a reason why it should matter.

SuchAFuriousDeath:
@kjarosh 🥺

core/src/display_object.rs
kjarosh: For instance, `try_bind_text_field_variable` calls `object.set` which calls watchers. Can we add at least a test that checks the order of called watchers?

SuchAFuriousDeath: I had Claude vibe out a test, see if it makes sense please, it does to me, but this isn't really an area of code I'm familiar with.

kjarosh: Is the test failing with the previous code?

SuchAFuriousDeath: No, I just checked. It doesn't depend on the order of iteration I guess.

SuchAFuriousDeath: However, I think your claim here: 

> For instance, `try_bind_text_field_variable` calls `object.set` which calls watchers.

Is wrong because `try_bind_text_field_variable` is called with `false` as the second argument and that results in watchers not being called, at least based on my understanding.

Issue #12261, opened 2023-07-23
Unable to load any swf in https://chat.kongregate.com/flash/GameShell.swf


Matrix4348:
### Describe the bug

Opening https://chat.kongregate.com/flash/GameShell.swf?api_path=%2Fflash%2FAPI.swf&game_swf=http%3A%2F%2Fchat.kongregate.com%2Fgamez%2F0000%2F9480%2Flive%2Fdoeo_kong.swf results in a black screen.

### Expected behavior

*Doeo* should be able to load inside this "game shell". Actually, **any** swf hosted on Kongregate should be able to, I only chose *Doeo* as an example.

### Affected platform

Browser's extension

### Operating system

Windows 8.1

### Browser

All

### Additional information

# Player Info
Allows script access: false
Renderer: wgpu
Adapter Backend: Gl
Adapter Name: "ANGLE (Google, Vulkan 1.3.0 (SwiftShader Device (Subzero) (0x0000C0DE)), SwiftShader driver)"
Adapter Device Type: Cpu
Adapter Driver Name: ""
Adapter Driver Info: ""
Enabled features: Features(TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES | PUSH_CONSTANTS)
Available features: Features(TEXTURE_COMPRESSION_BC | TEXTURE_COMPRESSION_ETC2 | TEXTURE_COMPRESSION_ASTC | TEXTURE_COMPRESSION_ASTC_HDR | CLEAR_TEXTURE | MULTIVIEW)
[...]

Issue #23919, opened 2026-06-07
Jinx Dark & Stormy Night -- Loading Issues


thenickman100:
### Describe the bug

The game loads and locks up here right after entering the house.  Jinx never shows up in this room and you can not click to enter any doors.  

<img width="1902" height="1207" alt="Image" src="https://github.com/user-attachments/assets/02a9e1b0-a222-444d-92cc-eda7dd4c9029" />

### Expected behavior

<img width="1041" height="746" alt="Image" src="https://github.com/user-attachments/assets/8b89c6ed-77cc-4e18-ac7f-d9319e00e67c" />

### Content Location

https://flasharch.com/en/archive/play/86d6cd4a647262d97992e91e61b28791

### Affected platform

Self-hosted version

### Operating system

Windows 11

### Browser

Microsoft Edge

ChrisCPI:
This site annoyingly has some kind of measure that prevents normal dev tools from being opened, but it does have its own output window for Ruffle logs. This is the notable output:
<img width="443" height="123" alt="Image" src="https://github.com/user-attachments/assets/6b901f22-32e5-49d6-974e-de161fe3379e" />

thenickman100:
I found a site that works and uses the Ruffle version from 2024-07-12.

I tried a handful of different versions and here are the results:

2024-07-12: Working
2026-05-16: Working
2026-05-25: Broken
2026-06-07: Broken

thenickman100:
Oh, and here is the link to the working site: https://flashga.me/embed/jinx-dark-and-stormy-night

ChrisCPI:
Looks like it broke on the May 26 nightly, presumably from #23806. 

Direct links to the SWFs for this game:
https://flashga.me/public/storage/games/multi-file-games/jinx-dark-and-stormy-night/maingame.swf
https://flashga.me/public/storage/games/multi-file-games/jinx-dark-and-stormy-night/console.swf (maingame loads this)

ChrisCPI:
(cc @kjarosh)

kjarosh:
Thanks for bisecting @ChrisCPI! https://github.com/ruffle-rs/ruffle/pull/23948 will fix this

PR #23942, opened 2026-06-10
🤖 avm1: Unify _level/_flash target parsing (close #12820)

### LINKED ISSUE #12820: avm1: loadMovie() and MovieClipLoader.loadClip() handle target parameter incorrectly
### Describe the bug

Ruffle does not handle "_level" strings as the target parameter in MovieClipLoader.loadClip(). Here is an example:
```as
var clipLoader = new MovieClipLoader();
clipLoader.loadClip("test.swf", "_level10");
```
Ruffle does not load test.swf. The same applies to the "_flash" alias for "_level" (e.g. "_flash10").

In loadMovie(), Ruffle does handle "_level" targets, but does not handle "_flash" ones. For example, this does not work:
```as
loadMovie("test.swf", "_flash10");
```

### Expected behavior

The movie should be loaded into the specified level.

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

This is causing #12261.

We should probably refactor `movie_clip_loader.load_clip()`, `action_get_url`, and `action_get_url_2` to use common string parsing logic for the target parameter. The latter two functions have the following TODO comment:
// TODO: Use `StageObject::get_level_by_path`.
But `get_level_by_path` was since refactored and renamed to `resolve_path_property`, and I'm not sure if it's viable to use as-is.

MavenRain:
Closes #12820. `MovieClipLoader.loadClip`, `getURL`, and `getURL2` (loadMovie/loadMovieNum) each carried their own ad-hoc `_levelN` target parser, none of which handled the legacy `_flash` synonym, and `loadClip` only *resolved* an existing level — so loading into a not-yet-existing `_levelN` string target silently failed (the reported bug; this also fixes the downstream #12261).

Following the refactor outlined in the issue, this routes all three call sites through one shared `parse_level_target` helper in `stage_object.rs`.

### What it does
- New `parse_level_target(target, case_sensitive) -> Option<i32>`: returns the level id only when the whole target is a complete level name (`_level5`, `_flash5`), and `None` otherwise — so a path like `_level0.child` correctly falls through to normal target-path resolution instead of being misread as level 0.
- `loadClip` now get-or-creates the level for a `_levelN`/`_flashN` string target (mirroring its numeric-target branch), fixing the silent no-op. `unloadClip`/`getProgress` stay find-only.
- `getURL`/`getURL2` use the same helper, adding `_flash` support and case-awareness; `getURL2` preserves its existing `> -1` "not a level" sentinel for non-level targets.
- Property resolution keeps the existing lenient `parse_level_id` — it operates on already-split single path segments, where leniency is the established behaviour.

### Testing
- Unit tests for `parse_level_target` (plain levels, the `_flash` synonym, negatives, the `_level0.child` path case, and case-sensitivity).
- New Rascal regression fixture `avm1/mcl_loadclip_level_target` exercising `loadClip`/`getURL`/`loadMovie` into `_level`/`_flash` levels end-to-end, plus the non-level path case.
- The full `avm1/` suite passes (706 tests) with no regressions.


kjarosh:
Coincidentally, I've been working on a similar thing related to https://github.com/ruffle-rs/ruffle/issues/23919 and https://github.com/ruffle-rs/ruffle/pull/23806.

From what I see now, this PR definitely overenginners stuff in few places. I think it's best to wait for my PR, some code will need to be removed from here because it will be duplicated (like `parse_level_target`), some code will need to be removed because it should start working OOTB.

kjarosh:
See https://github.com/ruffle-rs/ruffle/pull/23948

MavenRain:
@kjarosh 

Thanks for the heads up, and no problem waiting for your refactor.  I'd rather not duplicate work or land something that's about to be obsoleted.

A couple of questions so I know whether anything here is still worth keeping:
   - Does your change also fix the original report in #12820, where
   loadClip/loadMovie with a non-existent _levelN (or _flashN) string target
   silently fails because the level is found rather than created?  That's the
   user-visible bug, and #12261 (the Kongregate GameShell load) depends on it.
   - If your refactor makes that work out of the box, I'm happy to close this
   in favor of yours. If there's a residual piece you'd rather not fold in,
   point me at it and I can rebase down to just that.
  
Either way, happy to review your PR when it's up.

core/src/avm1/object/stage_object.rs
```diff
@@ -704,3 +750,44 @@ pub fn action_property_coerce<'gc>(
         _ => value,
     })
 }
+
+#[cfg(test)]
+mod tests {
```
kjarosh, line 755: Unit tests are not that useful for us. Sure, they test our assumptions, but they can be wrong. In this case it's better to move this testing logic to SWF tests.

tests/tests/swfs/avm1/mcl_loadclip_level_target/test.as
```diff
@@ -0,0 +1,21 @@
+var loader = new MovieClipLoader();
+
+// #12820: loading into a non-existent _levelN string target must create the
+// level and succeed (previously it silently failed and returned false).
+trace("loadClip _level10 -> " + loader.loadClip("child.swf", "_level10"));
```
kjarosh, line 5: What if you try to load a non-existing file? What if it's a network SWF? (you can test that by placing a SWF in `localhost/`)

kjarosh, line 5: What if we use a weird number like `_level-1`, `_level8327484`, is there a cap?

tests/tests/swfs/avm1/mcl_loadclip_level_target/test.as
```diff
@@ -0,0 +1,21 @@
...
+// getURL into a `_flash` level (action_get_url path).
+getURL("child.swf", "_flash14");
+trace("_level14 created -> " + (_level14 != undefined));
+
+// loadMovie into a `_flash` level (action_get_url_2 path).
+loadMovie("child.swf", "_flash16");
+trace("_level16 created -> " + (_level16 != undefined));
+
+// A non-level string target still resolves as a path, not a level.
+trace("loadClip _level0.missing -> " + loader.loadClip("child.swf", "_level0.missing"));
```
kjarosh, line 21: What does this test exactly? Why is the return value false?

tests/tests/swfs/avm1/mcl_loadclip_level_target/test.as
kjarosh, line 5: What if you try to load a MC into an existing level?

kjarosh:
The PR I mentioned is already merged so you can rebase on top of it.

tests/tests/swfs/avm1/mcl_loadclip_level_target/test.as
kjarosh, line 21: What if you load into an existing MC path?

tests/tests/swfs/avm1/mcl_loadclip_level_target/test.as
kjarosh: What if you pass some garbage after `_level`? I guess you could just list all values you want to test and iterate over them.

Lord-McSweeney:
Hi! This pull request hasn't been touched in a while, and there are still some unaddressed review comments on it. I'm going to close it now- feel free to reopen it or open a new PR if you can address those concerns.

PR #23947, opened 2026-06-10
frontend-utils: Allow switching reqwest back to rustls

torokati44:
## Description

This is necessary for the Android app, where reqwest can't use a native TLS backend, because there's no OpenSSL.

Referencing:
 - https://github.com/ruffle-rs/ruffle/pull/23894
 - https://github.com/ruffle-rs/ruffle-android/pull/762

## Testing

Infrastructure-only stuff.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2125,8 +2125,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ff2abc00be7fca6ebc474524697ae276ad847ad0a6b3faa4bcb027e9a4614ad0"
 dependencies = [
  "cfg-if",
+ "js-sys",
  "libc",
  "wasi",
+ "wasm-bindgen",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


SuchAFuriousDeath (approved):

kjarosh:
Doesn't native-tls work on Android? Is it not supported or were they any other issues?

kjarosh (approved):

torokati44:
I just saw this, it looking for OpenSSL:
https://github.com/ruffle-rs/ruffle-android/actions/runs/27267838836/job/80529358968?pr=762

I also don't see why it _couldn't_ do non-OpenSSL (so actually native) TLS on Android, but apparently it can't. 🤷‍♂️

torokati44:
No mention of anything Android-specific here, and I don't think we want OpenSSL there...

https://github.com/seanmonstar/reqwest#requirements

kjarosh:
I think the native crate might lack support for Android then. :/

torokati44:
Yep:

https://github.com/rust-native-tls/rust-native-tls
> Specifically, this crate uses SChannel on Windows (via the [schannel](https://crates.io/crates/schannel) crate), Secure Transport on macOS (via the [security-framework](https://crates.io/crates/security-framework) crate), and OpenSSL (via the [openssl](https://crates.io/crates/openssl) crate) on all other platforms.

PR #23972, opened 2026-06-13
avm1: Make Object::set accept impl Into<Value>

SuchAFuriousDeath:
## Description

This is just a simple quality of life change. It even slightly reduces binary size. 

## Checklist

### DIFF
--- a/core/src/avm1/object.rs
+++ b/core/src/avm1/object.rs
@@ -205,18 +205,29 @@ impl<'gc> Object<'gc> {
     }
 
     /// Set a named property on this object, or its prototype.
+    #[inline]
     pub fn set(
         self,
         name: impl Into<AvmString<'gc>>,
-        value: Value<'gc>,
+        value: impl Into<Value<'gc>>,
         activation: &mut Activation<'_, 'gc>,
     ) -> Result<(), Error<'gc>> {
-        let name = name.into();
+        self.set_internal(name.into(), value.into(), activation)
+    }
+
+    /// Non-generic body of `set`, kept separate to avoid duplicating this large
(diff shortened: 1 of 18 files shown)


core/src/avm1/object.rs
```diff
@@ -205,18 +205,26 @@ impl<'gc> Object<'gc> {
...
         self,
         name: impl Into<AvmString<'gc>>,
-        value: Value<'gc>,
+        value: impl Into<Value<'gc>>,
+        activation: &mut Activation<'_, 'gc>,
+    ) -> Result<(), Error<'gc>> {
+        self.set_internal(name.into(), value.into(), activation)
+    }
+
+    fn set_internal(
```
kjarosh, line 221: Why set_internal?

SuchAFuriousDeath: What do you propose instead?

kjarosh: I'm just wondering why we need this at all 🤔 is it because of a large number of specializations of `set`?

kjarosh (approved):

core/src/avm1/object.rs
SuchAFuriousDeath: That is why I did it that way, yeah

kjarosh: If that's the case I would add a comment explaining this to make sure it doesn't get inlined by someone doing refactors in the future.

Lord-McSweeney, line 221: If this method is supposed to never be inlined, it should be marked `#[inline(never)]`.

SuchAFuriousDeath: Done :saluting_face:

PR #23975, opened 2026-06-13
avm1: Improve behavior of `Sound` when `loadSound()` has been called

### LINKED ISSUE #5019: bug with sound.position issue on imported .mp3s: Conductor Volume game
Hello to the Ruffle team, Adrian17 on discord asked me to post our issues here. We are happy to post our AS2 code, .swfs, and test!

GAME ONLINE:  game: https://goclassical.org/kids/opera/volume.htm
And  .swf and .fla and .swd attached.
[VolumeGame2021.zip](https://github.com/ruffle-rs/ruffle/files/6947588/VolumeGame2021.zip)

PROBLEM: No Audio and game doesn't display 4 little volume slider controls below conductor and game does not move forward to play. The .swd debugger says the .mp3s loaded...
-- WORKS when "test movie" from Macromeda Flash Professional 8 authoring app. 
-- NOT WORKING when playing same exported .sfw on desktop (runs Macromedia Flash Player 8). 
-- NOT WORKING in html page with Ruffle on server: https://goclassical.org/kids/opera/volume.htm

EXPECTED BEHAVIOR: there's 4 little volume slider controls below conductor, setting volume for the 4 imported .mp3s: they should display and allow user to drag up-and-down, setting volume. At end of mp3s, triggers 2 buttons (repeat and continue) to display.

ISSUE: Adrian17 wrote on Discord #help: " I think we have volume control working, but not the more advanced sound features (like sound.position)" 
I can confirm that yes this game does use sound1.position, sound2.position, etc.

Thank you!

### LINKED ISSUE #22572: Sound.setVolume should not affect the whole movie if loadSound has been called
### Describe the bug

The [docs](https://open-flash.github.io/mirrors/as2-language-reference/Sound.html) for the AVM1 class `Sound` say:

> If you do not specify a target when you create a new Sound object, you can use the methods to control sound for the whole movie.

This appears to be only partially true. In Flash Player, the .setVolume method on a Sound object created without a target controls the volume for the whole movie _unless_ .loadSound has been called on the object. Ruffle does not implement this.

I ran into this when I noticed that muting BGM in [Elektra](https://321.lv/speles/elektra) also mutes all SFX under Ruffle, but not under Flash Player. I made a minimal repro case: [muting.zip](https://github.com/user-attachments/files/24282828/muting.zip)

### Expected behavior

muting.swf contains three buttons:

1. the red button in the top left makes a sound when clicked
2. the green button in the top right executes

```
   soundWithFile = new Sound();
   soundWithFile.loadSound("invalid.mp3",false);
   soundWithFile.setVolume(0);
```
3. the blue button in the bottom right executes
```
   soundWithoutFile = new Sound();
   soundWithoutFile.setVolume(0);
```

In Flash Player (and therefore the expected behavior):
1. if I press the red button, it makes a sound,
2. if I then press the green button and then the red button, the red button still makes a sound
    - this is true regardless of whether `invalid.mp3` exists
3. if I then press the blue button, the red button no longer makes a sound when pressed

In Ruffle,
1. [correct] the red button makes a sound,
2. [wrong] if I press the green button, the red button no longer makes a sound
3. [correct] if I press the blue button, the red button no longer makes a sound

### Content Location

[muting.zip](https://github.com/user-attachments/files/24282828/muting.zip)

### Affected platform

Desktop app

### Operating system

Linux, Ruffle 0.2.0-nightly.2025.12.21-nightly (fd174fbb2ee83fcd84c4be77e6bca0d15a6f5da2 2025-12-21) from AUR

### Browser

_No response_

### Additional information

I assume setPan and setTransform also only affect the loaded sound in this case, but have not tested them.

Maybe .attachSound also triggers this behavior? I have not tested it.

ChrisCPI:
## Description

Fixes #5019 
Fixes #22572 

This fixes several of the behaviors of the `Sound` object when playing external sound via `loadSound()`.

Some of the key changes:
- When `loadSound` is called on a Sound, it appears that it enters a state of being "separated" from its original owner (even when controlling global sound). This means it can no longer control any prior sounds that were added to it via `attachSound` (and can no longer attach new sounds in this way), and it inherits its own unique sound transform, not attached to any particular display object. This has now been implemented.
- Added the ability to load sounds synchronously (where execution is paused until load has completed) when loading local sounds on desktop, matching the Flash Player behavior where methods like `start()` could become immediately available in this case.
  - Additionally, the playback queue has been removed, as this was an incorrect interpretation of the behavior that was just described.
- It appears that a single Sound object can only load one sound at a time. Trying to call `loadSound()` again while another sound is being loaded will cancel the load, and `onLoad` gets called with `false` given as the success parameter (See the `avm1/sound_load_multiple_remote` test).

## Testing

Several tests added to cover this.

## Checklist

- [x] An LLM was involved in the authoring of this code. (Used in authoring the added `set_sound_transform_with_handle` function, and the logic for the load IDs to prevent multiple loads at one time on a single Sound)


core/src/avm1/globals/sound.rs
```diff
@@ -76,6 +76,19 @@ struct SoundData<'gc> {
...
+    is_stopped: Cell<bool>,
+
+    /// A dedicated sound transform for this sound.
+    /// This is `Some` when this sound has had `loadSound()` called on it,
+    /// and therefore enters a state of being "separated" from its
+    /// original owner (even if it was controlling global sound initially),
+    /// and can then only ever control externally loaded sounds.
+    /// Because of this, this can be used to check if this Sound is set
+    /// to play external sounds.
+    transform: Cell<Option<SoundTransform>>,
```
kjarosh: Instead of giving multiple meanings to this field, can't you reuse the state enum?

core/src/avm1/globals/sound.rs
```diff
@@ -76,6 +76,19 @@ struct SoundData<'gc> {
     /// This will be true if `Sound.loadSound` was called with `isStreaming` of `true`.
     /// A streaming sound can only have a single active instance.
     is_streaming: Cell<bool>,
+
+    /// Whether this sound has been stopped.
+    /// This is only used when in a mode of playing external sounds.
+    is_stopped: Cell<bool>,
```
kjarosh: Can sound be stopped before being loaded? If the value of this field is dependent on the state, it might be better to place it on the state enum

ChrisCPI: > Can sound be stopped before being loaded?

Yes, see the `avm1/sound_load_multiple_instances` test (frame 13-14, frame 14 should be silent).

I'm not sure how this could be tied into the state enum, as it seems more reserved for the state of actual audio being loaded.

The only purpose of this field is to prevent streaming sound from autoplaying when `stop()` was called prior to it loading.

core/src/avm1/globals/sound.rs
ChrisCPI: I'm not sure how I would do that exactly?

Earlier I had a separate `is_external` field in the native object, but it seemed redundant since it should always be `true` when the transform is `Some`, and always `false` when transform is `None`.

kjarosh: You could have the transform in the state enum, e.g. a separate enum entry for when the sound is external. But that depends on how it plays with the loading/loaded states.

core/src/avm1/globals/sound.rs
kjarosh: Sounds good, if you can stop irrespective of the loading state then it's fine.

core/src/avm1/globals/sound.rs
ChrisCPI: But doesn't it not make sense to have it be a part of the state enum? Because e.g. if a sound was still being loaded (in the `Loading` state), wouldn't that mean it can't access the transform being stored in a separate state, because it's not in that state at that time?

Lord-McSweeney:
If both `is_stopped` and `transform` are only valid for external sounds, I would suggest adding a new struct called `ExternalSound` or `ExternalSoundInfo` that holds both pieces of information, and adding a single `Option<ExternalSoundInfo>` field to `SoundData` rather than the two new fields.

ChrisCPI:
I can look into doing that if no one disagrees with the idea.

ChrisCPI:
> If both `is_stopped` and `transform` are only valid for external sounds, I would suggest adding a new struct called `ExternalSound` or `ExternalSoundInfo` that holds both pieces of information, and adding a single `Option<ExternalSoundInfo>` field to `SoundData` rather than the two new fields.

@kjarosh Thoughts on doing this idea? I think it could be helpful in using that to check whether the sound is external, rather than checking the transform.

kjarosh:
It depends on how this plays with the loading state, if it's unrelated to the state then yes.

ChrisCPI:
Putting this in draft for now - Need to do some refactoring, as it looks like the idea of the existence of a playback queue for loaded sounds is incorrect, and so I will likely remove it (effectively undoing most of #21163).

What actually appears to be happening is that in FP, local sounds are loaded synchronously, so that when `loadSound()` is called, it doesn't continue executing until the sound has been loaded. This is why the `start()` calls appeared to be queued, when they actually are not.

An easy way to observe this is to see that the `onLoad` listener can be fired in-between two traces before/after a `loadSound()` call. e.g.:

```as
var s = new Sound();
s.onLoad = function() { trace("onLoad"); }
trace("before");
s.loadSound("sound.mp3");
trace("after");
```
In the local projector, this outputs:
```
before
onLoad
after
```
But when loading a remote file, it becomes asynchronous, and `start()` calls have no effect before the sound is fully loaded.

core/src/avm1/globals/sound.rs
```diff
@@ -72,17 +131,20 @@ struct SoundData<'gc> {
...
-    /// This will be true if `Sound.loadSound` was called with `isStreaming` of `true`.
-    /// A streaming sound can only have a single active instance.
-    is_streaming: Cell<bool>,
+    /// Data for playing external sounds on this object.
+    /// This is `Some` when this sound has had `loadSound()` called on it,
+    /// and therefore enters a state of being "separated" from its
+    /// original owner (even if it was controlling global sound initially),
+    /// and can then only ever control externally loaded sounds.
+    // TODO: Can this be a `Lock` instead?
+    external: RefLock<Option<ExternalSound<'gc>>>,
```
Lord-McSweeney: What's blocking this from being a `Lock`?

core/src/loader.rs
```diff
@@ -1383,47 +1383,85 @@ pub fn load_sound_avm1<'gc>(
...
+        panic!("NativeObject must be Sound");
+    };
+
+    let mut activation = Activation::from_stub(uc, ActivationIdentifier::root("[Loader]"));
+
+    let external = sound
+        .external()
+        .expect("Loaded sound should have external sound data");
+
+    if external.load_id() != load_id {
```
Lord-McSweeney, line 1430: Can you add a comment here too to explain what this does?

Lord-McSweeney (approved):

Lord-McSweeney:
Thank you for working on this!

core/src/avm1/globals/sound.rs
```diff
@@ -572,20 +591,66 @@ fn load_sound<'gc>(
...
+            if is_blocking {
+                let _ = crate::loader::load_sound_avm1_blocking(
+                    activation.context,
+                    this,
+                    Request::get(request_url),
+                    load_id,
+                );
+                return Ok(Value::Undefined);
             }
         }
```
moulins, line 647: Using `#[cfg(target_family="wasm")]` as a proxy to detect whether we're in the browser or on desktop feels a little weird to me.
Ideally, the `NavigatorBackend` should advertise whether or not it supports blocking on a future, and the means to do so.

This is a somewhat minor concern so I won't block the PR on that, but IMHO that's something to keep in mind for future refactors.

core/src/avm1/globals/sound.rs
ChrisCPI: Nothing really; I haven't really tried to yet, it just took a while to get this working at all. I can look into it later

core/src/avm1/globals/sound.rs
ChrisCPI: Sounds good, I can probably add a comment in the code about this for now.

PR #23982, opened 2026-06-15
core: Use enum-map for ActionQueue priorities

SuchAFuriousDeath:
## Description

Replaces the raw `usize` priority index and fixed-size array with a   typed `ActionPriority` enum and `enum_map::EnumMap`.

Not 100% sure about the variant names, that's something I'm seeking feedback on. 

## Checklist

### DIFF
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -41,6 +41,8 @@ use crate::timer::Timers;
 use crate::vminterface::Instantiator;
 use async_channel::Sender;
 use core::fmt;
+use enum_map::Enum;
+use enum_map::EnumMap;
 use gc_arena::{Collect, Mutation};
 use ruffle_render::backend::{BitmapCacheEntry, RenderBackend};
 use ruffle_render::commands::{CommandHandler, CommandList};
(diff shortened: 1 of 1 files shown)


core/src/context.rs
```diff
@@ -485,22 +487,31 @@ pub struct QueuedAction<'gc> {
     pub is_unload: bool,
 }
 
+/// Priority bucket for queued actions. Higher variants drain first.
+#[derive(Copy, Clone, Collect, Enum)]
+#[collect(require_static)]
+pub enum ActionPriority {
+    Low = 0,
+    Normal = 1,
+    High = 2,
```
kjarosh: In general, I think we need to get our priorities straight. I'm not entirely sure 3 is enough, but I would prefer naming them based on what's being queued there.

SuchAFuriousDeath: Straight as a ruler :saluting_face:

core/src/context.rs
```diff
@@ -485,22 +487,35 @@ pub struct QueuedAction<'gc> {
     pub is_unload: bool,
 }
 
+/// Priority bucket for queued actions. Variants are listed in drain order
+/// (highest priority last) and grouped by which `ActionType`s land in them.
```
kjarosh: > (highest priority last)

That's a bit confusing, usually "high" priority actions go first, but sometimes "high priority" means "low numerical value".

SuchAFuriousDeath: Better now?

kjarosh: I still don't get it, which priority is the highest? 0 or 2?

SuchAFuriousDeath: What if I just remove the comment altogether? Would that be simpler?

kjarosh: It still wouldn't tell me which priority executes first :D

SuchAFuriousDeath: Third time's the charm? :pleading_face:

kjarosh: LGTM!

core/src/context.rs
```diff
@@ -512,22 +528,19 @@ impl<'gc> ActionQueue<'gc> {
...
-        }
+        });
     }
 
     /// Sorts and drains the actions from the queue.
     pub fn pop_action(&mut self) -> Option<QueuedAction<'gc>> {
         self.action_queue
-            .iter_mut()
-            .rev()
+            .values_mut()
```
kjarosh: I don't see `values_mut()` specifying the order in which values are returned?

SuchAFuriousDeath: I'm not really sure what to do here tbh, the iteration order comment only appears on the non-mut methods like as_slice and iter, as_slice_mut and values_mut don't have it even though internally they follow the same logic :(

kjarosh: `iter_mut`?

SuchAFuriousDeath: iter_mut doesn't have the guarantee in writing either IIRC.

SuchAFuriousDeath: Just checked, it only says 'Returns a mutable iterator over enum map.'. Just above it, iter does have the guarantee, even though they are the same method :( : 
image

kjarosh: They are not calling the same method actually, the methods just have the same name.

I think it's safe to call `iter_mut()` because by convention, it should use the same order as `iter()`. In addition to that, the derive macro actually gives us some guarantees: https://docs.rs/enum-map-derive/latest/enum_map_derive/derive.Enum.html

> When using a derive, enum maps are maintained in the order in which enum variants are declared. This is reflected in the value returned by Enum::into_usize, iterators of enum map as well as EnumMap::as_slice method.

Not sure if we can treat `values_mut()` as an iterator, the docs say it is, and it implements an iterator trait, so maybe it's safe to use `values_mut()` at the end of the day :thinking:

kjarosh (approved):

kjarosh (approved):

PR #23984, opened 2026-06-15
web: Use non-streaming WebAssembly.instantiate in Node-like environments

### LINKED ISSUE #23983: web: instantiateStreaming crashes Electron renderer with nodeIntegration (regression in 0.2.120)
## Summary

Loading the Ruffle web build in an Electron renderer with `nodeIntegration: true` crashes V8 at startup with:

```
node::wasm_web_api::StartStreamingCompilation (node_wasm_web_api.cc:175)
Assertion failed: !impl.IsEmpty()
```

The root cause is [electron/electron#36282](https://github.com/electron/electron/issues/36282): in Node-integrated renderers, `WebAssembly.instantiateStreaming` is dispatched to Node's WASM web API, which fails this assertion.

## When this regressed

This used to work. The regression in Ruffle is the combination of:

- `e01842f9` — `web: Migrate WASM initialization to use module_or_path parameter` (Apr 2025), which switched the loader from `init(response)` to `init({ module_or_path: response })`.
- `b0165148a` — `chore: Bump wasm-bindgen to 0.2.120, ...` (Apr 2026). Starting with `0.2.120`, wasm-bindgen prefers `WebAssembly.instantiateStreaming` when handed a `Response` via `module_or_path`. The earlier `0.2.108` did not.

Result: any Ruffle build from this point on crashes Electron renderers that have Node integration enabled.

## Reproduction

1. Build current `master` (`npm run build`).
2. Load the bundle in an Electron 30+ window with `webPreferences: { nodeIntegration: true }`.
3. The window crashes immediately at boot with the stack above.

## Scope of impact

Any Electron-based deployment of Ruffle that needs `nodeIntegration` (e.g. apps that bridge Flash content with Node APIs for filesystem, IPC, or native modules). The NewGrounds Player and the Terna SISNET enterprise client are two known examples.

## Why not "just disable nodeIntegration"

For deployments that genuinely need Node in the renderer (filesystem access, native modules, custom protocols), `sandbox: true` / `nodeIntegration: false` is not an option without rearchitecting around `contextBridge`. This is a Ruffle-side regression that's worth fixing on its own.

## Suggested fix

Bypass the streaming path when we know we're in a Node-like environment. PR follows.

### Expected behavior

I expect Ruffle to be able to run on Electron again.

### Content Location

Any swf will do because it always crashes on Electron

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

Electron v42

### Additional information

_No response_

falpi:
Fixes #23983

## What this does

`load-ruffle.ts` currently hands a `Response` to wasm-bindgen's `init({ module_or_path })`. Since `wasm-bindgen 0.2.120`, this dispatches to `WebAssembly.instantiateStreaming`, which crashes in Electron renderers with `nodeIntegration` (see linked issue and [electron/electron#36282](https://github.com/electron/electron/issues/36282)).

When we detect a Node-like environment, read the bytes first and call `init` with an `ArrayBuffer` instead — wasm-bindgen falls back to `WebAssembly.instantiate`, which works everywhere. In normal browsers the streaming path is preserved.

```ts
const isNodeLike =
    typeof process !== "undefined" && process.versions?.node !== undefined;
const init_arg = isNodeLike ? await response.arrayBuffer() : response;
await init({ module_or_path: init_arg });
```

## Trade-offs

- Browsers: unchanged. `instantiateStreaming` still used; full streaming-compile preserved.
- Node / Electron renderers: one extra in-memory buffer for the wasm bytes during init. For locally-hosted wasm (the common case in Electron) this costs effectively nothing.
- The `process` check is intentionally cheap and synchronous; no async detection or feature probing.

## Validated

- Browser (Chrome, Firefox): streaming path taken, no behavior change.
- Electron 42 with `nodeIntegration: true`: previously crashed at startup, now boots and runs to completion (tested against a real Flex 4.6 application — same workload that exhibited the original crash).

## Checklist

### DIFF
--- a/web/packages/core/src/load-ruffle.ts
+++ b/web/packages/core/src/load-ruffle.ts
@@ -111,7 +111,23 @@ async function fetchRuffle(
         response = wasmResponse;
     }
 
-    await init({ module_or_path: response });
+    // Electron renderers with `nodeIntegration` enabled crash inside
+    // `WebAssembly.instantiateStreaming` with
+    // `Assertion failed: !impl.IsEmpty()` from node_wasm_web_api.cc
+    // (electron/electron#36282). The streaming path is what wasm-bindgen
+    // picks when `module_or_path` is a `Response`; since the bump to
+    // wasm-bindgen 0.2.120 this is the default dispatch and any Node-like
+    // host now hits the crash. Read the bytes first when we're in a
+    // Node-like environment so wasm-bindgen falls back to the non-streaming
+    // `WebAssembly.instantiate`. Plain browsers (where `process` is
+    // undefined) keep the streaming-compile fast path.
+    const isNodeLike =
+        typeof (globalThis as { process?: { versions?: { node?: string } } })
+            .process?.versions?.node === "string";
+
+    await init({
+        module_or_path: isNodeLike ? await response.arrayBuffer() : response,
+    });
 
     return [RuffleInstanceBuilder, ZipWriter];
 }


kjarosh (changes requested):
As mentioned in https://github.com/ruffle-rs/ruffle/issues/23983, this looks like an Electron bug. Additionally, Electron is not an officially supported platform, as we maintain a desktop app instead.

falpi:
> As mentioned in #23983, this looks like an Electron bug. Additionally, Electron is not an officially supported platform, as we maintain a desktop app instead.

I apologize if I wasn't specific in opening this issue. I didn't remember Electron being a supported runtime, but I think if the fix is ​​simple, it would be a plus for the community to have this target available, assuming the issue isn't fixed upstream in the Electron repo.

PR #23985, opened 2026-06-15
core/audio: Don't pass in FLV-specific bytes to AacSubstreamDecoder

torokati44:
Instead, demux AudioSpecificConfig out-of-band, simplifying the decoder.

Add `extra_data: Option<Box<[u8]>>` to `SoundStreamInfo` to carry codec-specific init data out-of-band. The FLV handler now stores AAC sequence headers (AudioSpecificConfig) there instead of appending them to the audio substream, and strips the leading packet-type byte from AAC payloads since the enum variant already encodes that distinction.

AacSubstreamDecoder reads the config at construction time via `codec_params.with_extra_data`, removing the mid-stream packet-type dispatch that was previously in its decode loop.

## Description

This is the first step to making AAC decoding feasible from MP4 files, for a future second iteration of https://github.com/ruffle-rs/ruffle/pull/14655.
Without this, we'd have to prepend `0` and `1` bytes to the raw AAC chunks as present in MP4 samples.

No change in behavior expected.
Although this likely makes mid-stream changes to AAC config no longer work, but I have no idea whether that is valid at all, or if it's present in any actual content, or if FP accepted it. It does _feel_ somewhat wrong (or at least weird) though, even if it was ever happening.

As you can see (from the overly verbose comments and commit message), this whole thing was vibe-coded, but in the way that I found acceptable. Feel free to roast it relentlessly nevertheless.

## Testing

We have an AAC decoding test added by https://github.com/ruffle-rs/ruffle/pull/16526, and it still passes.
~~_(BTW that test could be simplified now, thanks to https://github.com/ruffle-rs/ruffle/pull/23397.)_~~ EDIT: But that would make them weaker, so maybe not worth it. Same for the similar g711 test.

The content that https://github.com/ruffle-rs/ruffle/issues/8891 is about still plays audio just fine.

## Checklist

### DIFF
--- a/core/src/backend/audio.rs
+++ b/core/src/backend/audio.rs
@@ -67,6 +67,13 @@ pub struct SoundStreamInfo {
     pub stream_format: swf::SoundFormat,
     pub num_samples_per_block: u16,
     pub latency_seek: i16,
+
+    /// Codec-specific configuration data ("decoder specific info"), demuxed out
+    /// of the container ahead of the audio data itself, if any.
+    ///
+    /// For AAC this holds the `AudioSpecificConfig` (from an FLV AAC sequence
+    /// header, or an MP4 `esds` box). Codecs that don't need it leave it `None`.
+    pub extra_data: Option<Box<[u8]>>,
 }
 
 impl From<swf::SoundStreamHead> for SoundStreamInfo {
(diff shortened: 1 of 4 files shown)


Lord-McSweeney (approved):
I'm not very familiar with this part of the codebase, but it LGTM

PR #24000, opened 2026-06-16
core: Add tilemalloc, a size-class allocator for enterprise workloads

falpi:
## What

Adds **tilemalloc**, a size-class memory allocator that replaces dlmalloc as the global
allocator on `wasm32-unknown-unknown`. It targets the small-allocation churn that
dominates long-running ActionScript workloads, turning per-call allocator cost
into O(1) push/pop on per-class intrusive free lists. The same pool counters are
then used to back the `flash.system.System` memory-introspection getters with
real numbers.

For further information, please see the attached document:
[tilemalloc.pdf](https://github.com/user-attachments/files/29025016/tilemalloc.pdf)

## Why — impact

Enterprise Flex / AS3 apps that load large XML into data grids are the worst case
for a general-purpose allocator: an `AdvancedDataGrid` bound to an E4X document
materialises hundreds of thousands of tiny, short-lived heap objects per load,
and dlmalloc's free-list traversal + coalescing cost scales with the live set.
Measured end-to-end on a real enterprise form (XML → Flex `AdvancedDataGrid`):

| Workload | dlmalloc (stock) | tilemalloc | Result |
|---|---|---|---|
| ~100 000-row XML into a grid | ~30 s first load | ~15 s | ~2× faster |
| ~250 000-row XML into a grid | did not finish (aborted after 5 min) | ~40 s | unusable → practical |
| Per-GC cost / repeated `dataProvider` refresh | heavy stutter, grows with working set | fast, flat | drastically lower |

The ~30 s → ~15 s gap appears on the **first** load against a fresh heap, which
isolates the cost as per-`malloc` free-list work rather than fragmentation. The
250k case is the headline: it changes which documents the player can open at all.

## How it works

- **alloc**: pop an intrusive free list (O(1)), else carve one tile off a bump
  pointer into the current extent.
- **dealloc**: push onto the free-list head (O(1)). **No coalescing** — freed
  tiles stay carved at their size and are reused by the next same-shaped alloc, so
  a construct/destruct workload reaches steady state with no extra `memory.grow`.
- **add_extent** (rare): grab one ~`extent_mb` region and install it as the bump
  range — two stores, no per-tile setup loop (the eager alternative would be ~1 M
  page-touching stores for the 8-byte class).
- **realloc**: in-place fast path when the new size stays in the same class (no
  alloc/copy/dealloc, no two-tiles-alive transient).
- **fallback**: anything over the largest class, or over-aligned, transparently
  goes to dlmalloc and is tracked separately.

Size classes are configured in one file (`core/src/tilemalloc/config.rs`); the
default covers 8–256 B at sub-pow2 granularity. Tile alignment is auto-derived
(greatest pow2 divisor of the size), not a knob. The allocator is `Sync` without
locking — sound on the single-threaded WASM runtime (documented; wasm-threads
would need a `Mutex`).

## Also included

- **Built-in diagnostics** — a low-overhead counter suite (per-class waste, peak
  alive, utilisation, in-place / cross-class realloc with grow/shrink split,
  optional size/align/fallback histograms gated by a build-time
  `ENABLE_HISTOGRAMS`). Exposed to JS as `window.tilemallocStats()` (JSON) and
  `window.tilemallocTable()` (single-line), auto-installed at module start. These
  are what make the class set tunable for a given workload.
- **`flash.system.System` memory getters** — `freeMemory`, `totalMemory`,
  `totalMemoryNumber`, `privateMemory` change from hard-coded stubs to real values
  derived from the WASM linear memory size, the `__heap_base` symbol, and the live
  pool counters.
- **wasm64-ready** — the memory helpers are gated on `target_family = "wasm"` with
  a separate `wasm64::memory_size` arm, so a future `wasm64-unknown-unknown` build
  works unchanged. The lone nightly gate (`simd_wasm64`) is `cfg_attr`-scoped to
  wasm64 only and inert everywhere else.

## Scope / notes

- Self-contained: new `core/src/tilemalloc/` module + a thin `#[global_allocator]`
  wrapper in `web/src/lib.rs`. No new dependencies. No added AS3 surface — the
  getters were already declared; only their bodies changed.
- Grow-only by design (extents are not returned); tune `extent_mb` and the class
  set per workload.
- Detailed design, the full statistics schema, and a "how to read the stats to
  choose the class configuration" guide are in the attached document.

## Checklist

### DIFF
--- a/core/src/avm2/globals/flash/system/System.as
+++ b/core/src/avm2/globals/flash/system/System.as
@@ -2,7 +2,6 @@ package flash.system {
     [Ruffle(Abstract)]
     public final class System {
         import __ruffle__.stub_method;
-        import __ruffle__.stub_getter;
 
         public static function gc():void {}
 
@@ -16,20 +15,11 @@ package flash.system {
             stub_method("flash.system.System", "disposeXML");
         }
 
-        public static function get freeMemory():Number {
-            stub_getter("flash.system.System", "freeMemory");
-            return 1024*1024*10; // 10MB
-        }
+        public static native function get freeMemory():Number;
 
-        public static function get privateMemory():Number {
-            stub_getter("flash.system.System", "privateMemory");
-            return 1024*1024*100; // 100MB
-        }
+        public static native function get privateMemory():Number;
 
-        public static function get totalMemoryNumber():Number {
-            stub_getter("flash.system.System", "totalMemoryNumber");
-            return 1024*1024*90; // 90MB
-        }
+        public static native function get totalMemoryNumber():Number;
 
         public static function get totalMemory():uint {
             return totalMemoryNumber as uint;
--- a/core/src/avm2/globals/flash/system/system.rs
+++ b/core/src/avm2/globals/flash/system/system.rs
@@ -26,3 +26,147 @@ pub fn set_clipboard<'gc>(
 
     Ok(Value::Undefined)
 }
+
+// ---------------------------------------------------------------------------
+// Memory introspection helpers (WASM-only, return 0 on native targets).
+//
+// The three `flash.system.System` memory getters
+// (`totalMemoryNumber`, `freeMemory`, `privateMemory`) are derived from
+// two primitive measurements of the running WASM module:
+//
+//   * `wasm_linear_memory_bytes()` — the total size of the module's
+//     linear memory at this moment (grows monotonically via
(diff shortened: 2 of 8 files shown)


falpi:
I realize this PR is too specific to hope for acceptance. Similarly, other branches I'm developing, which implement heavy optimizations appropriate for enterprise Flex applications, will no longer be submitted as PRs but will remain available on my fork.

PR #24028, opened 2026-06-21
avm2: Back FontDescription with a native object

kjarosh:
## Description

Move FontDescription's properties from ActionScript to Rust. Properties (fontName, fontWeight, fontPosture, fontLookup, renderingMode, cffHinting, locked) are now native getters/setters that read/write interior-mutable fields directly, and text_block.rs accesses them via the typed Rust API instead of raw slot access.

## Testing

Already covered by tests.

## Checklist


core/src/avm2/globals/flash/text/engine/text_block.rs
```diff
@@ -121,31 +120,13 @@ fn apply_format<'gc>(
 
         let (font, bold, italic, is_device_font) = if let Value::Object(font_description) =
             element_format.get_slot(format_slots::_FONT_DESCRIPTION)
+            && let Some(fd) = font_description.as_font_description_object()
```
Lord-McSweeney, line 123: `font_description` is guaranteed to be a `FontDescription` object, so this can just be an unwrap

Lord-McSweeney (approved):

core/src/avm2/globals/flash/text/engine/text_block.rs
kjarosh: Oh yeah, you're right. But I'll be backing ElementFormat with a native object as well, so I'll take care of it in a follow-up PR.

torokati44:
It might be obvious to you, but I'm missing the motivation behind this refactor. Some classes were migrated in the other direction, why is this class better this way?

kjarosh:
Objects in the FTE framework are tightly coupled, so this refactor will allow us to write simpler code as we won't have to worry about interfacing with AS all the time.

PR #24030, opened 2026-06-21
avm2: Back ElementFormat with a native Rust object

kjarosh:
## Description

Move ElementFormat's property storage from ActionScript slots into a new
ElementFormatObject backed by Gc-managed Rust data. Additionally,
improve behaviors for setting properties (add mapping & validation).

This will make it easier to implement FTE in Ruffle.

## Testing

Test included.

## Checklist


core/src/fte.rs
```diff
@@ -0,0 +1,299 @@
+//! Various structs related to FTE used across the whole codebase.
```
Lord-McSweeney, line 1: nit: Instead of `as_avm2_string` and `parse_avm2_string`, the methods could just be named `parse_string` and `as_string`

Lord-McSweeney (approved):

core/src/fte.rs
kjarosh: I meant to relay that the string is an AVM2 representation of the object (contrary to any other string representation, or an AVM1 one), but maybe it's too specific considering we don't have other representations.

kjarosh: Done

PR #24034, opened 2026-06-22
core: Add Avm1Enum/Avm2Enum macro derives

kjarosh:
## Description

Avm1Enum/Avm2Enum macro derives allow an enum to be annotated with
AVM1/AVM2 string variants, from which Avm1StrRepresentable or
Avm2StrRepresentable is being implemented.

These traits have methods for parsing and converting enums to AVM1/AVM2
strings.

This refactor reduces boilerplate in many places.

## Testing

Already covered by tests.

## Checklist

### DIFF
--- a/core/common/src/avm_string/common.rs
+++ b/core/common/src/avm_string/common.rs
@@ -119,6 +119,7 @@ ruffle_macros::define_common_strings! {
     "declaredBy",
     "decode",
     "default",
+    "deflate",
     "descent",
     "description",
     "device",
@@ -186,6 +187,7 @@ ruffle_macros::define_common_strings! {
     "loop",
     "lowercase",
     "lr",
+    "lzma",
     "macType",
     "matrixType",
     "menu",
@@ -357,5 +359,6 @@ ruffle_macros::define_common_strings! {
     "year",
     "yMax",
     "yMin",
+    "zlib",
     "zoom",
 }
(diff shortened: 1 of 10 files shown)


SuchAFuriousDeath (approved):

Lord-McSweeney:
Is there ever a situation where an enum has different strings used to represent it across AVM1 and AVM2?

SuchAFuriousDeath:
I imagine there might be case sensitivity differences

Lord-McSweeney:
IMO if there's no reason to have separate derives for AVM1 and AVM2 they should be united into one

kjarosh:
> Is there ever a situation where an enum has different strings used to represent it across AVM1 and AVM2?

Yes, there are few enums that differ in case-sensitivity between AVM1 and AVM2 (e.g. StageQuality), although it looks like most of them are in `render` or `swf`, which wouldn't make this macro useful, as it requres `AvmString`.

While I can't do much about `swf` without simplifying the macros, maybe it would be possible to depend on `common` in `render`.

kjarosh:
An alternative is to specify whether the match should be case-sensitive or not and whether the output should be uppercase or not instead of relying on AVM1/AVM2, but that seems more complicated.

Lord-McSweeney (approved):
It still seems a bit odd to me to mention "AVM2 enum" for enums that are only for AVM2, but this LGTM as it is

PR #24036, opened 2026-06-22
core: Set BitmapData transparency based on BitmapFormat

pngdrift:
## Description

Previously `BitmapData` loaded from `Loader` was always created with `transparent` set to true. Now it seems to be more similar to Flash behavior.

## Testing

I added `avm2/loader_bitmap_transparency` test.

## Checklist


core/src/loader.rs
```diff
@@ -1812,7 +1812,7 @@ impl<'gc> MovieLoader<'gc> {
                 // since Bitmap and BitmapData never have AVM1-side objects.
                 let bitmap = ruffle_render::utils::decode_define_bits_jpeg(data, None)?;
 
-                let transparency = true;
+                let transparency = bitmap.format() != BitmapFormat::Rgb;
```
kjarosh: What about `BitmapFormat::Yuv420p`? If that applies to all formats without transparency (which would make sense), it's better to add a method to `BitmapFormat`.

tests/tests/swfs/avm2/loader_bitmap_transparency/test.toml
```diff
@@ -0,0 +1 @@
+num_ticks = 2
```
kjarosh, line 1: Can you use the asc compiler for this test (see e.g. `avm2/textformat_display/test.toml` for how it's done, you can also consult tests' README).

core/src/loader.rs
pngdrift: now?

kjarosh: You can call it `supports_transparency` to make it grammatical.

core/src/loader.rs
```diff
@@ -1803,7 +1804,6 @@ impl<'gc> MovieLoader<'gc> {
             }
             ContentType::Gif | ContentType::Jpeg | ContentType::JpegXr | ContentType::Png => {
                 let mut activation = Avm2Activation::from_nothing(uc);
-
```
kjarosh, line 1806: Nit: unrelated change

kjarosh (approved):
LGTM, thank you and welcome!

SuchAFuriousDeath (approved):

Issue #19531, opened 2025-02-15
Axon: MP4 videos don't play


n0samu:
### Describe the bug

[Axon](http://www.notdoppler.com/axon.php)'s "How to Play" and "The Science" sections are supposed to play videos, but in the Ruffle desktop app they still open the video files in the user's browser. The videos are MP4 and use the H.264 codec so they should be supported (#14654).
![Image](https://github.com/user-attachments/assets/9c3edd38-dec8-4ea2-91c6-97d4ff79c5d8)

### Expected behavior

The videos should play if OpenH264 is enabled.

### Content Location

Page URL: http://www.notdoppler.com/axon.php
SWF URL: http://i.notdoppler.com/files/axon.swf

### Affected platform

Desktop app

### Operating system

Windows 10

danielhjacobs:
Of note, that PR currently has hardcoded support only for the H.263 codec, one of the many things making it still a draft.

PR #24038, opened 2026-06-22
core: Add support for playing F4V (MP4) files

torokati44:
## Description

Take 2 of https://github.com/ruffle-rs/ruffle/pull/14655, superseding it.

Compared to that, this relies on `re_mp4` (by the awesome Rerun.io team, who are also the main sponsors of `egui`), not on Mozilla's `mp4parse` crate. The former is more actively maintained, while the latter hasn't seen a real "release" in years, so we'd have to use it from a git source.
It is also a lot nicer in that it directly gives us the "sample" (in MP4 lingo, it's an encoded video frame or audio chunk) offsets and lengths directly, so we don't have to calculate them in convoluted ways from the `moov` box.
One downside is that it doesn't provide a BoxIter mechanism, so we'd have to either contribute it, or manually pass in only complete boxes when streaming from a remote file, otherwise the parser errors out. I think an entirely missing `mdat` box would be tolerated though, just not a truncated one. This is one of the TODOs left.

This one also adds audio (AAC only) decoding!

~~Also includes https://github.com/ruffle-rs/ruffle/pull/23985, which was motivated by this.~~ EDIT: That is now merged.

Now that streams.rs has grown a bit lengthy, in the future it might be worth it to refactor it into a "directory module", with the FLV and MP4 parts in separate submodules.

## Testing

There's a basic test included.
https://github.com/ruffle-rs/ruffle/issues/19531 isn't fixed by this alone, at least on desktop - there's a weird H.264 NALU encapsulation artifact in one of the files it plays, which is apparently accepted by WebCodecs but not by OpenH264. Fixing it is out of scope, unrelated to the MP4 format itself.
There's also something fishy on WeedToWonder.org last time I tested (frames not passed in for the video decoder in the right order, or key frames being missed?) - hence why this is also still a draft.

## Checklist


danielhjacobs:
> There's also something fishy on WeedToWonder.org last time I tested

Note for the curious, this is specifically: https://new.weedtowonder.org/ (more specifically, *.weedtowonder.org used to be this site, while the current https://weedtowonder.org/ used to only be accessible on https://old.weedtowonder.org/, but that was changed when it was found https://old.weedtowonder.org/ would work with Ruffle while, for the time being, https://new.weedtowonder.org/ was broken with Ruffle)

torokati44:
Do you have a simple way I could test some representative content from that site with the desktop player? Building and loading the extension after each change is a bit cumbersome.

torokati44:
Also, what exactly is different between the old. and new. sites?

torokati44:
(For testing: I guess just `cargo run -- 'https://new.weedtowonder.org/w2w-home.swf'` should work, as there's already an MP4 video there.)

torokati44:
This now also includes https://github.com/ruffle-rs/ruffle/pull/24098, as another prerequisite.

tests/tests/swfs/visual/video/f4v/Test.as
```diff
@@ -0,0 +1,24 @@
...
+            var nc = new NetConnection();
+            nc.connect(null);
+            var ns = new NetStream(nc);
+            ns.client = new Object();
+            ns.client.onMetaData = function(metaData:Object):void {};
+            var vid = new Video(400, 220);
+            addChild(vid);
+            vid.attachNetStream(ns);
+            ns.play("logo-anim.f4v");
+        }
```
Lord-McSweeney, line 22: If it's possible, could you add test coverage for seeking?

core/src/streams.rs
```diff
@@ -1284,6 +1397,271 @@ impl<'gc> NetStream<'gc> {
...
+                        source.preload_offset.set(buffer.len());
+                    }
+                }
+            }
+
+            if media_context.is_none() {
+                return;
+            }
+
+            let media_context = media_context.as_ref().unwrap();
```
Lord-McSweeney, line 1468: The above lines can be reduced to something like
```
let Some(ref media_context) = media_context else {
    return;
}
```
right?

core/src/streams.rs
```diff
@@ -1284,6 +1397,271 @@ impl<'gc> NetStream<'gc> {
...
+                                self.0.last_decoded_bitmap.replace(Some(frame));
+                                if let Some(mc) = self.0.attached_to.get() {
+                                    mc.invalidate_cached_bitmap();
+                                    *context.needs_render = true;
+                                }
+                            }
+                            Err(e) => {
+                                // H.264 decoders commonly have a startup delay of
+                                // a few frames before producing output. This is
+                                // expected and not a real error.
```
Lord-McSweeney, line 1575: If this is common when decoding f4v video, it's probably better to have it be under `tracing::debug!`

core/src/streams.rs
```diff
@@ -1284,6 +1397,271 @@ impl<'gc> NetStream<'gc> {
...
+                                tracing::warn!(
+                                    "F4V video decode produced no frame (may be startup delay): {}",
+                                    e
+                                );
+                            }
+                        }
+                    } // video loop
+                }
+            } // if let Some(vti)
+
```
Lord-McSweeney, line 1585: The code for decoding FLV video and audio is separated into the methods `flv_video_tag` and `flv_audio_tag`. Could you refactor the video and audio decoding logic here into their own methods too?

core/src/streams.rs
torokati44: It's not specific to MP4, it's specific to H.264. It's already a thing in FLV - a few "Decoder produced no frame" is emitted there as well for H.264. It's just that MP4 can (AFAIK) only hold H.264, not any of the older codecs that SWF can.

torokati44: In the future, at least for H.264, the "decoder feeding" and "presenting" playheads of both FLV and MP4 are supposed to be separated, at least for video (maybe for audio too), so the decoded frames are ready by the time they are supposed to be shown, and aren't slightly more delayed than in FP. The current "one encoded frame in, the same decoded frame out" API breaks for H.264 and is a bit misused.

core/src/streams.rs
torokati44: It seems like that `ref` is not even necessary, but otherwise, yeah, evidently! Thanks!

core/src/streams.rs
torokati44: More to the point: The corresponding FLV part is even: `tracing::error!("Decoding video frame {} failed: {}", frame_id, e);`

torokati44: So for now I've just matched that here...

tests/tests/swfs/visual/video/f4v/Test.as
torokati44: I did and it doesn't work... XP

torokati44: But it looks like its a separate issue, I tried seeking after the video ended, and apparently that should resume playback, but we don't...

Lord-McSweeney: Does the issue happen in FLV video?

torokati44: I plan to figure it out later, and also isolate the seek-after-end behavior from this test, by making this video longer.

PR #24050, opened 2026-06-23
Revert "ci: Install pinned Chrome to workaround issue"

danielhjacobs:
Reverts ruffle-rs/ruffle#23896

Will rebase once a week until this passes CI.

### DIFF
--- a/.github/workflows/test_web.yml
+++ b/.github/workflows/test_web.yml
@@ -177,21 +177,9 @@ jobs:
           name: selfhosted-dist
           path: web/packages/selfhosted/dist
 
-      - name: Install pinned Chrome (workaround for actions/runner-images#14169)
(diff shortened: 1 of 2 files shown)


torokati44 (approved):
Nice, thanks!

kjarosh:
Thanks @danielhjacobs for taking care of this! Maybe we should pin browser versions in general, but the question is whether it's maintainable long term.

danielhjacobs:
It may be good to pin in order to test a known working browser version, but it can be good to stay on the bleeding edge of browser testing too. There could be legitimate circumstances where browser updates break our tests, and we'd want to fix that as quickly as possible. This is especially true for things we do that are a bit obscure, like our focus or context menu handling.

PR #24062, opened 2026-06-24
 extension: Page opt out should disable polyfill 

### LINKED ISSUE #24029: navigator.plugins["Shockwave Flash"] is still present even if data-ruffle-optout is set
### Describe the bug

i have set data-ruffle-optout in my <html> tag and now ruffle doesnt override the swf with its own, however it seems

```navigator.plugins["Shockwave Flash"]``` still contains an entry for flash content; meaning things like ```DetectFlashVer``` still see flash as installed;

### Expected behavior

adding data-ruffle-optout would mean that navigator.plugins["Shockwave Flash"] is undefined and DetectFlashVer would say its not found.

### Content Location

N/A

### Affected platform

Self-hosted version

### Operating system

Windows 10

### Browser

Firefox 152

### Additional information

_No response_

### PR DESCRIPTION
**Pull request by danielhjacobs**
## Description

Fixes #24029

## Testing

Build extension and try data-ruffle-optout

## Checklist

### DIFF
--- a/web/packages/extension/src/background.ts
+++ b/web/packages/extension/src/background.ts
@@ -1,11 +1,31 @@
 import * as utils from "./utils";
 import { isMessage } from "./messages";
 
-async function contentScriptRegistered() {
+/**
+ * Returns whether the plugin polyfill content script is registered and whether
+ * its JS file matches the expected variant.
+ */
+async function getPluginPolyfillRegistration(
+    expectedScript?: string,
+): Promise<{ registered: boolean; matches: boolean }> {
     const matchingScripts = await utils.scripting.getRegisteredContentScripts({
         ids: ["plugin-polyfill"],
     });
-    return matchingScripts?.length > 0;
+
+    if (matchingScripts?.length === 0) {
+        return {
+            registered: false,
+            matches: false,
+        };
+    }
+
+    // Content script IDs are unique, so there is at most one matching script.
+    const script = matchingScripts[0];
+
+    return {
+        registered: true,
+        matches: script?.js?.[0] === expectedScript,
+    };
 }
 
 // Copied from https://github.com/w3c/webextensions/issues/638#issuecomment-2181124486
@@ -164,7 +184,7 @@ async function disableSWFTakeover() {
 }
 
 async function enable() {
-    const { swfTakeover } = await utils.getOptions();
+    const { swfTakeover, ignoreOptout } = await utils.getOptions();
     if (swfTakeover) {
         await enableSWFTakeover();
     }
@@ -174,7 +194,18 @@ async function enable() {
     ) {
         return;
     }
-    if (!(await contentScriptRegistered())) {
+    const expectedScript = ignoreOptout
+        ? "dist/pluginPolyfillIgnoreOptout.js"
+        : "dist/pluginPolyfill.js";
+
+    const { registered, matches } =
+        await getPluginPolyfillRegistration(expectedScript);
+    if (!matches) {
+        if (registered) {
+            await utils.scripting.unregisterContentScripts({
+                ids: ["ruffle", "plugin-polyfill", "4399"],
+            });
+        }
         // Reuse the exclude_matches of dist/content.js in the manifest.
         const excludeMatches =
             utils.runtime.getManifest().content_scripts![0]!.exclude_matches!;
@@ -191,7 +222,7 @@ async function enable() {
             },
             {
                 id: "plugin-polyfill",
-                js: ["dist/pluginPolyfill.js"],
+                js: [expectedScript],
                 persistAcrossSessions: true,
                 matches: ["<all_urls>"],
                 excludeMatches,
@@ -222,7 +253,8 @@ async function disable() {
     ) {
         return;
     }
-    if (await contentScriptRegistered()) {
+    const { registered } = await getPluginPolyfillRegistration();
+    if (registered) {
         await utils.scripting.unregisterContentScripts({
             ids: ["ruffle", "plugin-polyfill", "4399"],
         });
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by kjarosh** [comment]
Can you elaborate on why it works and why your initial logic was incorrect?

**Comment by n0samu** [comment]
This is the important diff. Self-explanatory
image

(Although I don't get why it couldn't just be `if (!checkPageOptout()) { installPlugin }` instead of using an anonymous function)

**Comment by kjarosh** [comment]
Yeah, I saw the diff. Maybe I'm dumb, but how do you know the attribute is parsed/available at that point in time? What's the order of operations for our polyfill and potential js from the page? I would appreciate an explanation because I'm not very experienced with writing extensions and polyfilling plugins.

**Comment by danielhjacobs** [comment]
https://groups.google.com/a/chromium.org/g/chromium-extensions/c/32Dembdnqe8?pli=1

**Comment by danielhjacobs** [comment]
> In the beginning (run_at: document_start) there is no `body` or `head` in the document yet, only `document.documentElement` exists (without any child nodes) that represents the root element i.e. <html> in an HTML document.

Initial assumption was documentElement was unavailable until the DOM loaded

**Comment by danielhjacobs** [comment]
> (Although I don't get why it couldn't just be if (!checkPageOptout() { installPlugin } instead of using an anonymous function) 

Fair

**Comment by danielhjacobs** [comment]
Ah, I'm slightly incorrect though, I needed to revert the other file changes and just check `document.documentElement.hasAttribute("data-ruffle-optout")` in the plugin polyfill script. `checkPageOptout`'s code to check frames causes the code to run later than `document_start` (which is why it "may still be too late for some Flash detection scripts), so I need to just check top-level opt-out.

**Comment by danielhjacobs** [comment]
This site has swfobject.js in its head, which has a Flash plugin detection script, and with my changes it still works, but also page opt-out disables the polyfill installation.

https://github.com/user-attachments/assets/12f19c3e-7607-4b12-ba54-264cca3a12b5

**Comment by danielhjacobs** [comment]
Spec source is probably this: https://html.spec.whatwg.org/multipage/parsing.html#the-before-html-insertion-mode

> ↪ A start tag whose tag name is "html"
>     [Create an element for the token](https://html.spec.whatwg.org/multipage/parsing.html#create-an-element-for-the-token) in the [HTML namespace](https://infra.spec.whatwg.org/#html-namespace), with the [Document](https://html.spec.whatwg.org/multipage/dom.html#document) as the intended parent. Append it to the [Document](https://html.spec.whatwg.org/multipage/dom.html#document) object. Put this element in the [stack of open elements](<https://html.spec.whatwg.org/multipage/parsing.html#stack-of-open-elements>).

https://html.spec.whatwg.org/multipage/parsing.html#stack-of-open-elements

> *The "[before html](https://html.spec.whatwg.org/multipage/parsing.html#the-before-html-insertion-mode)" [insertion mode](https://html.spec.whatwg.org/multipage/parsing.html#insertion-mode) creates the [html](https://html.spec.whatwg.org/multipage/semantics.html#the-html-element) [document element](https://dom.spec.whatwg.org/#document-element), which is then added to the stack.*

**Comment by danielhjacobs** [comment]
Meanwhile: https://developer.chrome.com/docs/extensions/reference/api/extensionTypes#type-RunAt

> Scripts are injected after any files from css, but before any other DOM is constructed or any other script is run.

This would inherently be after the "before html" insertion mode.

**Comment by danielhjacobs** [comment]
The second commit is the unfortunately complex code needed to allow the plugin polyfill to respect ignoreOptout.

**Comment by kjarosh** [comment]
Can you use imperative titles in commits and also describe what you're changing? It's not obvious _why_ we need to do this without description.

Also it looks like you can just do one commit, as the first one is trivial compared to the second one.

File: web/packages/extension/src/background.ts [hunk 1]
```diff
@@ -1,11 +1,26 @@
 import * as utils from "./utils";
 import { isMessage } from "./messages";
 
-async function contentScriptRegistered() {
+async function getPluginPolyfillRegistration(ignoreOptout?: boolean) {
```
kjarosh: Can you add a return type? It looks like it's a complex type.

web/packages/extension/src/background.ts
```diff
@@ -1,11 +1,26 @@
...
+        };
+    }
+
+    const expectedScript = ignoreOptout
+        ? "dist/pluginPolyfillIgnoreOptout.js"
+        : "dist/pluginPolyfill.js";
+
+    return {
+        registered: true,
+        matches: matchingScripts[0]?.js?.[0] === expectedScript,
```
kjarosh: Why are we checking the 0th element here? What's the logic behind "matches"? (that could be explained in the struct returned by the function or in the function's docs)

danielhjacobs: `getRegisteredContentScripts` returns an array of `scripting.RegisteredContentScript`, so with a single id in its`scripting.ContentScriptFilter`, it returns at most one value in an array. I'll add a JSDoc and update the function to be a little clearer.

web/packages/extension/src/background.ts
```diff
@@ -1,11 +1,35 @@
...
+    const expectedScript = ignoreOptout
+        ? "dist/pluginPolyfillIgnoreOptout.js"
+        : "dist/pluginPolyfill.js";
+
+    // Content script IDs are unique, so there is at most one matching script.
+    const script = matchingScripts[0];
+
+    return {
+        registered: true,
+        matches: script?.js?.[0] === expectedScript,
```
kjarosh, line 27: I guess I still don't get the idea behind "matches". We aren't passing `ignoreOptout` anywhere, so it's just "register whatever you want and check if that was the requested script"?

danielhjacobs: > We aren't passing `ignoreOptout` anywhere

Yes we are:

https://github.com/ruffle-rs/ruffle/pull/24062/changes#diff-b9e86d98769e020e68919846919692da73f818952ca6a5b93d84d449af40e916R201-R202

We're just not using it in `disable`

danielhjacobs: In `enable`, if the wrong polyfill script is registered or no polyfill script is registered (`!matches`), we need to register the right polyfill script. If any scripts were already registered (`registered`), we need to unregister them first.

danielhjacobs: But in `disable`, we should unregister whatever scripts are registered, regardless if they're the right scripts. As such, `disable` doesn't need to know `ignoreOptout`, because it only checks `registered`, not caring about `matches`.

kjarosh: Ahh gotcha, it might be better to accept `expectedScript` instad of the boolean so you don't duplicate this code

```ts
    const expectedScript = ignoreOptout
        ? "dist/pluginPolyfillIgnoreOptout.js"
        : "dist/pluginPolyfill.js";
```

kjarosh (approved):

PR #24066, opened 2026-06-25
avm2: Implement some `Worker` methods

SuchAFuriousDeath:
## Description
This implements the state transitions, but not the actual starting up of a separate thread. For some weird reason, Worker.terminate returns a Boolean, but worker.start doesn't :thinking:.

## Testing

Test included.

## Checklist


tests/tests/swfs/avm2/worker_terminate/Test.as
```diff
@@ -0,0 +1,37 @@
...
+        public function Test() {
+            trace("== primordial ==");
+            var p:Worker = Worker.current;
+            trace("isPrimordial = " + p.isPrimordial);
+            trace("state = " + p.state);
+            tryTerminate("primordial", p);
+            trace("state after = " + p.state);
+
+            trace("== created ==");
+            var w:Worker = WorkerDomain.current.createWorker(new ByteArray());
```
kjarosh, line 25: What if you pass `null`?

core/src/avm2/object/worker_object.rs
```diff
@@ -35,10 +41,109 @@ impl<'gc> TObject<'gc> for WorkerObject<'gc> {
...
+    }
+
+    fn state(self) -> WorkerState {
+        match &self.0.kind {
+            WorkerKind::Primordial => WorkerState::Running,
+            WorkerKind::Spawned { state } => state.get(),
+        }
+    }
+
+    pub fn state_as_string(self, activation: &mut Activation<'_, 'gc>) -> AvmString<'gc> {
```
kjarosh: Can you use `Avm2Enum` instead?

tests/tests/swfs/avm2/worker_terminate/Test.as
```diff
@@ -0,0 +1,37 @@
...
+            tryTerminate("second", w);
+            trace("state = " + w.state);
+        }
+
+        private function tryTerminate(label:String, w:Worker):void {
+            try {
+                var r:Boolean = w.terminate();
+                trace(label + " terminate returned " + r);
+            } catch (e:Error) {
+                trace(label + " terminate threw " + e.errorID + ": " + e.message);
```
kjarosh: It looks like you're not asserting on the type. You can use `getStackTrace()` which will print all interesting information about the error.

core/src/avm2.rs
```diff
@@ -605,6 +624,13 @@ impl<'gc> Avm2<'gc> {
             .load_script(0, &mut activation)
             .expect("Script should load");
         init_native_system_classes(&mut activation);
+
+        // The primordial worker represents the runtime running the main SWF
+        // and must exist for the entire lifetime of the player. It is created
+        // eagerly here, after all system classes (including `Worker`) are
+        // loaded, so subsequent code can rely on it always being available.
+        let primordial = WorkerObject::new_primordial(&mut activation);
+        activation.avm2().current_worker = Some(primordial);
```
Lord-McSweeney: For consistency with domain memory initialization, this would be better in `globals::load_playerglobal`. The comment is also no longer correct

core/src/avm2/globals/flash/system/worker.rs
```diff
@@ -22,12 +50,72 @@ pub fn create_message_channel<'gc>(
...
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    avm2_stub_method!(activation, "flash.system.Worker", "start");
+
+    let worker = this_worker(this);
+
+    if worker.start() {
+        let event = EventObject::bare_default_event(activation.context, "workerState");
+
+        dispatch_event(activation, worker.into(), event, false);
```
Lord-McSweeney: Is there a reason there are calls to `dispatch_event` instead of `Avm2::dispatch_event`?

core/src/avm2/object/worker_object.rs
```diff
@@ -35,10 +42,98 @@ impl<'gc> TObject<'gc> for WorkerObject<'gc> {
...
 }
 
+/// Distinguishes the primordial worker (main SWF runtime) from workers created
+/// via `WorkerDomain.createWorker`. Only `Spawned` carries lifecycle state;
+/// the primordial worker is permanently `Running`.
+#[derive(Collect)]
+#[collect(require_static)]
+pub enum WorkerKind {
+    Primordial,
+    Spawned { state: Cell<WorkerState> },
```
Lord-McSweeney, line 52: No strong opinion either way, but I feel like this might be a bit simpler if all workers could have worker state, the primordial worker was always `Running`, and whether the worker was primordial were tracked by a boolean flag.

core/src/avm2/object/worker_object.rs
```diff
@@ -35,10 +42,98 @@ impl<'gc> TObject<'gc> for WorkerObject<'gc> {
...
+                state: Cell::new(WorkerState::New),
+            },
+        )
+    }
+
+    pub fn new_primordial(context: &mut UpdateContext<'gc>) -> Self {
+        Self::construct(context, WorkerKind::Primordial)
+    }
+
+    fn construct(context: &mut UpdateContext<'gc>, kind: WorkerKind) -> Self {
```
Lord-McSweeney: nit: For consistency, could you call this method `new` instead of `construct`?

core/src/avm2/object/worker_object.rs
```diff
@@ -35,10 +42,98 @@ impl<'gc> TObject<'gc> for WorkerObject<'gc> {
...
+    /// be terminated, and the primordial worker throws `Error #3732`.
+    pub fn terminate(self, activation: &mut Activation<'_, 'gc>) -> Result<bool, Error<'gc>> {
+        let state = match &self.0.kind {
+            WorkerKind::Primordial => return Err(make_error_3732(activation)),
+            WorkerKind::Spawned { state } => state,
+        };
+
+        let mut changed = false;
+
+        state.update(|s| match s {
```
Lord-McSweeney, line 129: Does FP ignore calls to `terminate` on workers that are not yet running, or workers that have already been terminated?

core/src/avm2/object/worker_object.rs
SuchAFuriousDeath: I like it better my way tbh :D

core/src/avm2/object/worker_object.rs
SuchAFuriousDeath: Not sure tbh

kjarosh (approved):

PR #24079, opened 2026-06-29
swf: Parse `DefineFontAlignZones` tag using zerocopy

SuchAFuriousDeath:
## Description

No need to allocate memory here. Zerocopy is already a transitive dependency.

## Testing

I could make tests for this if they are deemed necessary. 

## Checklist


SuchAFuriousDeath:
Actually maybe this is problematic for the users of the swf crate? :sweat_smile: oops. I have some more of these in mind, potentially.

PR #24082, opened 2026-06-29
avm1: Serialize top-level date/xml/display object properly

danielhjacobs:
## Description

Fix top-level serialization of Date/XML/DisplayObject

## Testing

This brings an existing known_failure closer to working, though doesn't fully fix it.

## Checklist


SuchAFuriousDeath (approved):

core/src/avm1/globals/shared_object.rs
```diff
@@ -87,8 +87,17 @@ pub fn serialize<'gc>(activation: &mut Activation<'_, 'gc>, value: Value<'gc>) -
         Value::Number(number) => AmfValue::Number(number),
```
Lord-McSweeney, line 87: Is there no way to just use `serialize_value_to_writer` in this function?

danielhjacobs: `serialize_value_to_writer` requires a name because it writes a property into an object and expects a key/value mapping. Also, while `serialize_value_to_writer` uses `ObjWriter` and mutates it, `serialize` directly translates a `Value` to an `AmfValue`. Also, `serialize_value_to_writer` has to do reference tracking and caching since it's possible for you to serialize the same data multiple times when using AMF, in which case the latter data should use references, while top-level serialization does not do the same. The only logic that is shared is the match arms, but it feels like the added boilerplate outweighs the benefits there.

danielhjacobs: The main thing to keep in mind is `serialize_value_to_writer` handles writing data for local shared objects, while `serialize` just translates an AVM1 value straight into an AMF0 value.

Lord-McSweeney (approved):

PR #24090, opened 2026-06-30
core: Fix HTML entity parsing (close #24084)

### LINKED ISSUE #24084: TextField htmlText doesn't decode hex character references
### Describe the bug

The htmlText property of TextField elements decodes entities including decimal character references, but not hex character references.
Seen in AS2 and AS3. I do not have a setup to check AS1.

Image

### Expected behavior

As shown in the Flash Player portion of the above screenshot, the `&#x26;` should have been decoded to `&`.

### Content Location

AS2: [htmlTextDecodeIssueAS2.zip](https://github.com/user-attachments/files/29479372/htmlTextDecodeIssueAS2.zip)
AS3: [htmlTextDecodeIssueAS3.zip](https://github.com/user-attachments/files/29479377/htmlTextDecodeIssueAS3.zip)

### Affected platform

Desktop app

### Operating system

Windows 11

### Browser

_No response_

### Additional information

_No response_

SNBeast:
## Description

This PR rewrites the numeric HTML entity parser to fix divergences with Flash Player. 

## Testing

The test `html_entity_parsing` has been created.

## Checklist


n0samu (approved):

SuchAFuriousDeath:
It would be nice to have a test in this PR.

SNBeast:
> It would be nice to have a test in this PR.

Fortunately, there will be. Reading this made me realize that I could just read back `TextField.text`.

tests/tests/swfs/text/html_entity_parsing/test.toml
```diff
@@ -0,0 +1 @@
+num_frames = 1
```
kjarosh: Can you use asc to compile the test?

tests/tests/swfs/text/html_entity_parsing/Test.as
```diff
@@ -0,0 +1,21 @@
...
+	// all the explicit entities supported...
+	"This is an ampersand: &amp;\n" +
+	"This is a less-than sign: &lt;\n" +
+	"This is a greater-than sign: &gt;\n" +
+	"This is a quotation mark: &quot;\n" +
+	"This is an apostrophe: &apos;\n" +
+	"This is a non-breaking space: &nbsp;\n" +
+	// numeric tests...
+	"This is a decimal-encoded greater-than sign: &#62;\n" +
+	"This is a hex-encoded ampersand: &#x26;\n" +
```
kjarosh: What if:
* You don't finish the escape?
* It contains non-hex characters?
* It's empty?
* It's super long?
* It uses `X` instead of `x`?
* It uses uppercase and lowercase hex digits?

Similarly for decimal escapes.

kjarosh: Basically what you can do to make the test simpler and cover more cases, is have a list of HTML strings with different test cases and iterate over them printing before (htmlText) and after (text).

core/src/html/text_format.rs
```diff
@@ -52,17 +52,35 @@ fn process_html_entity(src: &WStr) -> Option<WString> {
...
+                    let numeric_parse_start = s[start_index..].trim();
+                    let numeral_start_index = match s.get(start_index).map(u8::try_from) {
+                        Some(Ok(b'-')) | Some(Ok(b'+')) => 1,
+                        _ => 0,
+                    };
+                    let numeral_segment = &numeric_parse_start[numeral_start_index..];
+                    let end = if radix == 16 {
+                        numeral_segment
+                            .find(|c| {
+                                !((c >= b'0' as u16 && c <= b'9' as u16)
```
kjarosh: You can use `swf_is_ascii_hexdigit` for that

tests/tests/swfs/text/html_entity_parsing/test.toml
kjarosh: It looks like you're still not using asc, you can see `tests/tests/swfs/avm2/text_element_basic/test.toml` for how it's done.

tests/tests/swfs/text/html_entity_parsing/test.as
```diff
@@ -0,0 +1,134 @@
+// Compile with:
+//  mtasc -main -header 200:150:30 Test.as -swf test.swf
+
+class Test {
+	static function main(current) {
+		_root.createTextField("tf", 0, 0, 0, 200, 150);
+		_root.tf.html = true;
+		var tests:Array = new Array(
```
kjarosh: Thanks for writing such comprehensive tests!

SNBeast:
This is ready for re-review now, I haven't thought of anything else to poke it for.

wstr/src/parse.rs
```diff
@@ -126,6 +130,36 @@ mod int_parse {
             }
         })
     }
+
+    pub fn from_wstr_radix_wrapping<T: IntParse>(s: &WStr, radix: u32) -> Option<T> {
+        assert!(
```
kjarosh: I'm wondering if we can somehow get rid of this duplication, it's 1:1 the same as `from_wstr_radix`, but using wrapping instead of regular addition.

moulins: This new method isn't needed, `from_wstr_radix::<Wrapping<T>>` should already have the correct behavior.

SNBeast: > This new method isn't needed, `from_wstr_radix::<Wrapping<T>>` should already have the correct behavior.

Yep, realized that as well (though it did not handle negative for unsigned correctly).

moulins: Hmm, you've removed the `is_neg && !T::SIGNED` check, but maybe `Wrapping<_>`'s `IntParse` impl should always advertise `SIGNED = true` instead?

SNBeast: > Hmm, you've removed the `is_neg && !T::SIGNED` check, but maybe `Wrapping<_>`'s `IntParse` impl should always advertise `SIGNED = true` instead?

I added a `WRAPPING` bool instead, so the `SIGNED` bool remains honest.

kjarosh (approved):
LGTM, thank you!

PR #24096, opened 2026-06-30
avm1: Recursively serialize arrays as StrictArray when appropriate

danielhjacobs:
## Description

NetConnection and LocalConnection, when sending AMF data, send arrays as StrictArray in AMF0 if the arrays do not have custom properties. However, when flushed, SharedObject does not serialize any arrays as StrictArray. Ultimately, this is why MavenRain added promote_dense_ecma_to_strict_array in https://github.com/ruffle-rs/ruffle/pull/23772 and didn't use it in flush. I prefer adding and passing around a (edit) struct indicating whether strict arrays should be forced to be treated as ECMAArray, true for SharedObject, false for NetConnection/LocalConnection.

## Testing

The code changes fix known failures with NetConnection and LocalConnection. SharedObject serializing regular arrays as ECMAArray is proven by the existing test https://github.com/ruffle-rs/ruffle/tree/master/tests/tests/swfs/avm1/shared_object, which begins failing without this added (edit) struct (if StrictArray is always used for purely numeric arrays)

## Checklist


SuchAFuriousDeath (approved):

core/src/avm1/globals/netconnection.rs
```diff
@@ -241,7 +241,7 @@ fn add_header<'gc>(
         .unwrap_or(&Value::Bool(true))
         .as_bool(activation.swf_version());
 
-    let value = serialize(activation, *args.get(2).unwrap_or(&Value::Null));
+    let value = serialize(activation, *args.get(2).unwrap_or(&Value::Null), false);
```
kjarosh: Where do you call `serialize` with `true`?

danielhjacobs: I call `new_lso` with `true` in `SharedObject` `flush` and `SharedObject` `getSize`, and `new_lso` calls `serialize` with the passed value.

danielhjacobs: Or wait, maybe not.

danielhjacobs: Addressed, will rebase to one commit if the build works.

kjarosh: Hmm, I'm confused, why did tests pass if the code was wrong?

danielhjacobs: Code wasn't wrong, serialize was just always called with false, therefore now that's implicit.

More specifically, LocalConnection and NetConnection are the only callers of serialize. They called it with false, and it called new_lso with the passed value (now it explicitly calls new_lso with false). SharedObject flush and SharedObject getSize call new_lso with true. new_lso calls recursive_serialize with the passed value. recursive_serialize calls serialize_value_to_writer with the passed value.

danielhjacobs: With the fact that serialize is a function in SharedObject, the expectation is probably that it's used by SharedObject, but I can confirm if you trace back its call sites, it is not.

danielhjacobs:
[amf0_sol.html](https://github.com/user-attachments/files/29602878/amf0_sol.html)

This file shows the shared_object tests serialization in plain text, as well as validating that the comments about it match its content.

core/src/avm1/globals/shared_object.rs
```diff
@@ -99,14 +105,23 @@ pub fn serialize<'gc>(activation: &mut Activation<'_, 'gc>, value: Value<'gc>) -
...
         Value::MovieClip(_) => AmfValue::Undefined,
     }
 }
 
+fn serialize_object_properties<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    object: Object<'gc>,
+) -> Vec<Element> {
+    let mut w = Amf0Writer::default();
+    recursive_serialize(activation, object, &mut w, AmfContext::Connection);
```
kjarosh: Why is it `AmfContext::Connection` if we're in `shared_object.rs`?

danielhjacobs: Because `serialize` is only called by NetConnection and LocalConnection despite its file location.

danielhjacobs: https://github.com/ruffle-rs/ruffle/pull/24096#discussion_r3502299913

danielhjacobs: Unlike AVM2, AVM1 doesn't have an amf.rs file, therefore AVM1 AMF code that doesn't specifically apply to SharedObject but applies to more than one AVM1 class (such as NetConnection and LocalConnection) has historically ended up in shared_object.rs.

core/src/avm1/globals/shared_object.rs
```diff
@@ -106,7 +106,8 @@ fn deserialize_lso<'gc>(
 
 fn new_lso<'gc>(activation: &mut Activation<'_, 'gc>, name: &str, data: Object<'gc>) -> Lso {
     let mut w = Amf0Writer::default();
-    recursive_serialize(activation, data, &mut w);
+    // Hardcoded because LSOs always use ECMA arrays
```
kjarosh, line 109: It's not obvious what the comment is referencing, you can extract the config into a variable and place the comment above it.

core/src/avm1/amf.rs
```diff
@@ -6,6 +8,39 @@ use flash_lso::types::{Element, ObjectId, Reference, Value as AmfValue};
 use std::collections::BTreeMap;
 use std::rc::Rc;
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct AmfConfig {
+    pub supports_strict_arrays: bool,
+}
+
+impl AmfConfig {
+    pub const fn prefer_strict() -> Self {
```
kjarosh: Do we need those methods? I think it makes sense to just instantiate the struct inline when using.

core/src/avm1/amf.rs
```diff
@@ -1,3 +1,5 @@
+//! Code used by more than one class that uses AMF serialization
```
kjarosh: What about

```
AVM1 AMF serialization and deserialization code.

Used in objects such as NetConnection, LocalConnection, or SharedObject.
```

core/src/avm1/amf.rs
```diff
@@ -103,19 +127,37 @@ fn serialize_value_to_writer<'gc>(
...
-                if let Some(mut aw) = aw {
-                    recursive_serialize(activation, o, &mut aw);
-
-                    // TODO: What happens if an exception is thrown here?
-                    let length = o
-                        .length(activation)
-                        .expect("Failed to get length for SharedObject array");
-
-                    aw.commit(name, length as u32);
+                if !config.supports_strict_arrays || has_custom_properties {
```
kjarosh: Can we invert this condition? I think it would be easier to read.

danielhjacobs:
PRs, in order:

1) #24156
2) #24155
3) #24161
4) #24096

danielhjacobs:
An issue remains in the test added by #24156

Due to https://github.com/ruffle-rs/rust-flash-lso/issues/115, Ruffle serializes the cyclic object with the wrong reference index.

core/src/avm1/amf.rs
```diff
@@ -43,6 +51,20 @@ fn object_class_name<'gc>(
     }
 }
 
+/// Analyze array keys to determine max length and presence of custom properties
+fn analyze_array_keys<'gc>(keys: &[AvmString<'gc>], initial_len: usize) -> (usize, bool) {
+    let mut length = initial_len;
+    let mut has_custom = false;
+
+    for key in keys {
+        match key.to_utf8_lossy().parse::<usize>() {
```
moulins, line 60: This isn't consistent with the AVM1-specific index parsing logic (see [here](https://github.com/ruffle-rs/ruffle/blob/master/core/src/avm1/object/script_object.rs#L1044)).

It should be detectable by testing how "almost-integer" keys behave, e.g. `"01"`, `" 12"`, etc.

danielhjacobs: I have not been ignoring this, it's just been very hard to align both a new test I created based on this comment and amf0_serde_suite with what this is supposed to do. Long story short, it is *correct* for this **not** to be consistent with the AVM1-specific index parsing logic, but it's currently wrong in a different way, but for NetConnection it's wrong in one way, LocalConnection is wrong in another way, and SharedObject is right with this code.

danielhjacobs: Namely: https://github.com/ruffle-rs/ruffle/pull/24096/commits/e1dd2755f6dbd10e6253ccf97f5055f0435d999b

danielhjacobs: The biggest difference is it's only a StrictArray in Flash if and only if it uses only numeric indices (no signs or whitespace), but the length still matches the result of ` parse_array_index`, except that it can be greater than the max i32, it which case it's an ECMAArray but has the correct length, possibly up to i64.

PR #24101, opened 2026-07-01
chore: Add ttf-parser advisory to ignored list in deny.toml

danielhjacobs:
## Description

`ttf-parser` is no longer maintained, but works fine for our usage.

## Testing

CI will stop complaining

## Checklist

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -91,4 +91,5 @@ github = [
 version = 2
 
 # `paste` crate is no longer maintained https://rustsec.org/advisories/RUSTSEC-2024-0436
-ignore = ["RUSTSEC-2024-0436"]
+# `ttf-parser` crate is no longer maintained https://rustsec.org/advisories/RUSTSEC-2026-0192
+ignore = ["RUSTSEC-2024-0436", "RUSTSEC-2026-0192"]


torokati44 (approved):
I've also been meaning to do something like this, thanks!
It'll be a while until we can (if we want to) remove `ttf-parser` from the dep tree, as it's not only a direct dependency, but an indirect one too, through like three different paths.

deny.toml
```diff
@@ -91,4 +91,5 @@ github = [
 version = 2
 
 # `paste` crate is no longer maintained https://rustsec.org/advisories/RUSTSEC-2024-0436
-ignore = ["RUSTSEC-2024-0436"]
+# `ttf-parser` crate is no longer maintained https://rustsec.org/advisories/RUSTSEC-2026-0192
+ignore = ["RUSTSEC-2024-0436", "RUSTSEC-2026-0192"]
```
kjarosh, line 95: You can move comments above each entry instead of a list above `ignore`.

torokati44: I thought of this too, but with only 2 elements, it felt more nitpicky than actually necessary.

kjarosh: Sure, but then it's gonna be 3 and more, and even with 2 it's not obvious which one is which

kjarosh: Oh lol I didn't notice it's already merged, welp

torokati44: Yeah, we can reformat it once it gets to 3. 😐 Or 1... 😎🤞

PR #24107, opened 2026-07-03
avm2: Make SWFs loaded with loadBytes use their loader's sandbox type

### LINKED ISSUE #18594: [Papa's Taco Mia] Pre-2014 versions do not progress past the initial loading screen
### Describe the bug

Pre-2014 versions of Papa's Taco Mia, before Flipline removed MochiAds from their games, will not progress past the initial loading screen used to decrypt the game, instead landing on a black screen.
https://github.com/user-attachments/assets/0878f09f-6108-4b25-bf78-44474c7a8c7b
The following error appears in the console upon this happening:
Image

### Expected behavior

The game should progress to the MochiGames and Flipline Studios logo screens, and then to the game as normal.

### Content Location

Earlier versions of Taco Mia can be found via the Wayback Machine: https://web.archive.org/web/20131226000000*/http://i.flipline.com/gamefiles/papastacomia/papastacomia.swf

### Affected platform

Desktop app

### Operating system

Windows 10

### Browser

_No response_

### Additional information

This error does not occur for any of the other encrypted pre-2014 Papa's games in Ruffle (Burgeria, Freezeria, Pancakeria, etc.); They all run normally.
This error also doesn't occur in earlier versions of Ruffle, such as the August version currently used on the Wayback Machine, suggesting a regression was introduced at some point in the past 3 months.

Lord-McSweeney:
## Description

Fixes #18594

Loader SWFs have full control over the tags of a SWF loaded using `Loader.loadBytes`, so if we were to use the sandbox type declared in that SWF's header, the SWF could break sandboxing. Instead, always use the sandbox type of the loader SWF, to ensure that it can't load a child SWF with different sandboxing.

## Testing

I've added a new test, `avm2/sandbox_type_inherited`.

## Checklist


core/common/src/tag_utils.rs
```diff
@@ -283,6 +283,10 @@ impl SwfMovie {
         self.is_from_bytes
     }
 
+    pub fn set_sandbox_type(&mut self, sandbox_type: SandboxType) {
+        self.sandbox_type = sandbox_type;
```
kjarosh: I'm strongly against adding this method. Security sandbox type is a critical security flag, and should be set once when the movie is created, it shouldn't be possible to change it later.

tests/tests/swfs/avm2/sandbox_type_inherited/test.toml
```diff
@@ -0,0 +1 @@
+num_frames = 4
```
kjarosh, line 1: Can you use asc for compiling? You can set useNetwork there.

core/src/loader.rs
```diff
@@ -1650,8 +1655,21 @@ impl<'gc> MovieLoader<'gc> {
...
+                    } => forced_sandbox_type,
+                    _ => None,
+                };
+
+                let mut movie = SwfMovie::from_data_and_info(
+                    data,
+                    url.clone(),
+                    loader_url.clone(),
+                    from_bytes,
+                    forced_sandbox_type,
```
kjarosh: Is there ever a case when the sandbox type is forced and the SWF is not loaded from bytes?

Lord-McSweeney: I don't think there is

kjarosh: In that case I think it would be nice to enforce that, i.e. require passing sandbox type when loading from bytes and prohibit passing it otherwise.

Lord-McSweeney: I've made some changes, does this look better?

PR #24122, opened 2026-07-05
avm2: Introduce a global list of error messages

kjarosh:
## Description

Previously we were providing messages on the spot when instantiating errors. Flash contains a list of error messages as can be observed by `Error.getErrorMessage()`. This PR implements that method and introduces a global list of errors.

This should be better (in terms of binary size) than the previous approach because it doesn't produce that much code and moves parameter substitution fully into runtime. It also unlocks some refactors we can do in the future to instantiate errors in a simpler way.

Currently this increases binary size by 0.3%, because it provides a list of all possible error messages, not only those thrown by Ruffle.

## Testing

Tests included, one text fixed (had wrong output).

## Checklist


evilpie (approved):

core/src/avm2/error_messages.rs
```diff
@@ -0,0 +1,795 @@
...
+macro_rules! error_message {
+    ($code:literal $(,)?) => {
+        $crate::avm2::error_messages::error_message::<$code, 0>([""; 0])
+    };
+    ($code:literal, $($arg:expr),+ $(,)?) => {
+        $crate::avm2::error_messages::error_message::<$code, { [$(stringify!($arg)),+].len() }>([$($arg.to_string()),+])
+    };
+}
+
+pub fn error_message<const CODE: u32, const N: usize>(args: [impl Display; N]) -> String {
```
moulins: I think it'd be better to take a `[&dyn Display; N]` here, then there's no need to call `.to_string()` on each argument in the `error_message!` macro.

kjarosh: Good idea, done

PR #24123, opened 2026-07-05
avm2: Fix ElementFormat constructor initialization order

evilpie:
## Description

The initialization order is observable because invalid values will throw.

## Testing

Includes a test that fails with the previous initialization order.

## Checklist


tests/tests/swfs/avm2/element_format_constructor_order/test.toml
```diff
@@ -0,0 +1 @@
+num_frames = 1
```
kjarosh: Can you use asc to compile this test?

tests/tests/swfs/avm2/element_format_constructor_order/Test.as
```diff
@@ -0,0 +1,42 @@
...
+                0,
+                1,
+                "auto",
+                "<invalid>");
+        } catch (e:*) {
+            trace(e);
+        }
+
+        try {
+            new ElementFormat(null,
```
kjarosh: Can you cover all cases related to the order? I think it's enough to cover cases where parameters are getting progressively less invalid in the order of initialization.

Otherwise we don't know what's the real order of initialization.

tests/tests/swfs/avm2/element_format_constructor_order/test.toml
evilpie: Ok

tests/tests/swfs/avm2/element_format_constructor_order/Test.as
evilpie: Makes sense.

tests/tests/swfs/avm2/element_format_constructor_order/output.txt
```diff
@@ -0,0 +1,19 @@
+textRotation: auto dominantBaseline: roman alignmentBaseline: useDominantBaseline kerning: on breakOpportunity: auto digitCase: default digitWidth: default ligatureLevel: common typographicCase: default
+textRotation: auto dominantBaseline: roman alignmentBaseline: useDominantBaseline kerning: on breakOpportunity: auto digitCase: default digitWidth: default ligatureLevel: common typographicCase: invalid
```
kjarosh, line 3: Huh, how is that possible? There should be validation for `typographicCase`...

tests/tests/swfs/avm2/element_format_constructor_order/Test.as
```diff
@@ -0,0 +1,63 @@
...
+            /* trackingRight */ 0,
+            /* trackingLeft */ 0,
+            /* locale */ "en",
+            breakOpportunity,
+            digitCase,
+            digitWidth,
+            ligatureLevel,
+            typographicCase);
+        } catch (e:*) {
+            trace(e);
```
kjarosh: As a bonus you can also trace stack frames (or at least the top one), because that will give us information whether setters should indeed be used in the constructor (as compared to e.g. a native constructor or setting fields directly).

tests/tests/swfs/avm2/element_format_constructor_order/output.txt
evilpie: not sure what you mean. There is an error below because this doesn't work.

kjarosh: Oh nevermind, I messed up reading the lines. Sorry for noise.

kjarosh (approved):

tests/tests/swfs/avm2/element_format_constructor_order/Test.as
evilpie: Included. (The setter is called in Flash)

kjarosh (approved):

PR #24125, opened 2026-07-05
avm2: Implement `Error.throwError()`

kjarosh:
## Description

This patch implements Error.throwError(). It uses String.replace() as
it can be seen in stack traces coming from errors thrown from toString()
methods of arguments.

## Testing

Tests included.

## Checklist


evilpie:
Nice. We could probably use this to replace a bunch of hardcoded errors in our own ActionScript code.

core/src/avm2/globals/Error.as
```diff
@@ -36,5 +36,22 @@ package {
         public native function getStackTrace():String;
 
         public static const length:int = 1;
+
+        public static function throwError(type:Class, index:uint, ...rest):* {
+            var template:String = getErrorMessage(index);
+            var message:String = template.replace(/%[0-9]/g, function(match):String {
```
evilpie:
```suggestion
            var message:String = template.replace(/%([0-9])/g, function(_, match):String {
```

core/src/avm2/globals/Error.as
```diff
@@ -36,5 +36,22 @@ package {
...
+                // `match` contains the matched pattern, e.g. `%3`
+                var param:int = match.charAt(1) - '1';
```
evilpie:
```suggestion
                var param:int = int(match) - 1;
```

core/src/avm2/globals/Error.as
evilpie: Btw, I tried this in normal Flash and it was complaining about the number of parameters for the replacer function being wrong.

core/src/avm2/globals/Error.as
kjarosh: Done

core/src/avm2/globals/Error.as
kjarosh: Done

evilpie (approved):

kjarosh:
I added a commit that patches AVM2 playerglobal and gives an empty name to anonymous methods (as discussed on Discord).

PR #24128, opened 2026-07-05
chore: Mark edittext_caret_empty as flaky

kjarosh:
See https://github.com/ruffle-rs/ruffle/actions/runs/28749040742/job/85244908286?pr=24127

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,7 +1,9 @@
 # TODO: Figure out why these specific tests are flaky on CI, fix them, then delete this file.
+# `visual/edittext/edittext_caret_empty` is flaky on Windows for some reason.
 [[profile.ci.overrides]]
 filter = """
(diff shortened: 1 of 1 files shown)


SuchAFuriousDeath (approved):

torokati44:
Do you have any guesses for why this specific test, of all, is flaky?

danielhjacobs (approved):
Approved because it's flaky, but same question as @torokati44

kjarosh:
Not entirely sure, but maybe rendering one frame takes more than a second, which causes the caret to blink.

torokati44:
Oh, hmm, okay, that makes sense. In an ideal world, that caret blinking could be made somehow deterministic in the test harness, independent of the test execution time - but alas, as we are all painfully aware, this is not one of those ideal worlds...

PR #5808, opened 2021-12-05
Add Flash Screen Video (V1 only) decoder

torokati44:
This progresses https://github.com/ruffle-rs/ruffle/issues/723.

While there is a V2 decoder as well in the decoder used as source material, I have not been able to find or create any SWFs with videos embedded in this format as of yet.
So, I just stripped that out for now, to make the code simpler and smaller.

Even for the V1 format, the only real-world example I could find was ["Hey Hey 16K"](https://archive.org/details/heyhey16k), and another sample (that could be artificial) is [an MPlayer test file](http://samples.mplayerhq.hu/FLV/flash_screen/screen.flv).

At least for V1, I can create more with ffmpeg, but for V2, I wasn't able to embed any content into an SWF, only into an FLV.

Some questions I have:
 - Should this also be put behind a feature switch? - EDIT: Yes, done.
 - Should I separate it into a separate file? (I would really not like to put just 200 lines of code into an entirely separate crate though...) EDIT: Yes, done, as well as for all other codec (glue) modules.

PR #24135, opened 2026-07-06
video: Reduce allocations/copies when decoding frames

moulins:
## Description

Currently, the `VideoDecoder::decode_frame` method returns owned `Bitmap<'static>`s, forcing decoders to allocate a fresh buffer for each decoded frame.

This PR changes the API to instead provide a short-lived buffer through a callback, and updates the decoders to take advantage of this by reusing buffers more aggressively.

## Testing

All `visual/video` tests pass on native, but I still need to do a visual check with my human eyes, and properly test the web decoder.

## Checklist

### DIFF
--- a/render/src/bitmap.rs
+++ b/render/src/bitmap.rs
@@ -117,6 +117,17 @@ pub struct Bitmap<'a> {
 }
 
 impl<'a> Bitmap<'a> {
+    // Produces an zero-sized image with the given format.
+    #[inline]
+    pub fn empty(format: BitmapFormat) -> Self {
+        Self {
+            width: 0,
+            height: 0,
+            format,
+            data: Cow::Borrowed(&[]),
+        }
+    }
+
     /// Ensures that `data` is the correct size for the given `width` and `height`.
     #[inline]
     pub fn new<D>(width: u32, height: u32, format: BitmapFormat, data: D) -> Self
@@ -168,6 +179,21 @@ impl<'a> Bitmap<'a> {
         if let Cow::Owned(buf) = data {
             buf.resize(expected_len, 0);
         }
+    }
+
+    #[inline]
+    pub fn reborrow(&self) -> Bitmap<'_> {
+        Bitmap {
+            width: self.width,
+            height: self.height,
+            format: self.format,
+            data: Cow::Borrowed(&self.data),
+        }
+    }
+
+    #[inline]
+    pub fn into_buf(self) -> Cow<'a, [u8]> {
+        self.data
     }
 
     pub fn to_rgb(mut self) -> Self {
--- a/render/src/bitmap.rs
+++ b/render/src/bitmap.rs
@@ -123,31 +123,12 @@ impl<'a> Bitmap<'a> {
     where
         D: Into<Cow<'a, [u8]>>,
     {
-        Self::new_impl(width, height, format, data.into())
-    }
+        let mut data = data.into();
 
-    fn new_impl(width: u32, height: u32, format: BitmapFormat, mut data: Cow<'a, [u8]>) -> Self {
         // If the size is incorrect, either we screwed up or the decoder screwed up.
         let expected_len = format.length_for_size(width as usize, height as usize);
         if data.len() != expected_len {
-            tracing::warn!(
-                "Incorrect bitmap data size, expected {} bytes, got {}",
-                expected_len,
-                data.len(),
-            );
-
-            // Truncate or zero-pad to the expected size.
-            if let Cow::Borrowed(slice) = &data {
-                // Allocate the owned buffer ourselves instead of using `Cow::to_mut`, to avoid
-                // a reallocation if the buffer needs to be padded.
-                let mut vec = Vec::with_capacity(expected_len);
-                vec.extend_from_slice(&slice[..expected_len]);
-                data = Cow::Owned(vec);
-            }
-            match &mut data {
-                Cow::Owned(data) => data.resize(expected_len, 0),
-                Cow::Borrowed(_) => unreachable!(),
-            }
+            Self::resize_data(&mut data, expected_len);
         }
 
         Self {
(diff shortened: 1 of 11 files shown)


torokati44:
This might even have to evolve a bit further in the future, to use a buffer pool of some sorts.
The reason is, that most video codecs fit the basic "encoded frame in, decoded frame immediately out" paradigm, but not H.264, with its bidirectional predictions. So, ideally, we'd have to make the FLV/MP4 demuxers able to feed a few encoded H.264 frames ahead of the playhead, and store the decoded frames until it's time to present them. This would make us match Flash more closely, and also unlock the OpenH264 bump (and with it, CI on Windows ARM), because the decoding delay becomes irrelevant.

moulins:
Tested various video SWF files; everything seems to be in working order.

render/src/bitmap.rs
```diff
@@ -158,6 +139,47 @@ impl<'a> Bitmap<'a> {
...
+            (Cow::Owned(buf), Some(extra)) => buf.reserve_exact(extra),
+            (Cow::Owned(_), None) => (),
+        }
+
+        if let Cow::Owned(buf) = data {
+            buf.resize(expected_len, 0);
+        }
+    }
+
+    #[inline]
```
Lord-McSweeney, line 184: Huh, somehow this method and the one below are considered uncovered despite their callers being covered?

Lord-McSweeney (approved):
I'm completely unfamiliar with this area but the code LGTM

torokati44:
I've been meaning to review this a couple times but ... uh ... life 🫥

torokati44:
## First commit:

It looked a bit weird to me that `new` was just a proxy for `new_impl`, which was not used anywhere else. This split makes more sense. Too bad https://doc.rust-lang.org/std/hint/fn.unlikely.html isn't stable yet, I think it would make more sense to instead annotate the size mismatch condition itself. The existing "workaround" of https://doc.rust-lang.org/std/hint/fn.cold_path.html doesn't look better TBH...

It took me a while to understand exactly when each arm of that `match` is going to be taken, and then why that `if` after is what it is. If you feel like, a few short comments could improve maintainability, but it looks smart and correct to me.
Also about the fact that the match only ever truncates, but also prepares the `Vec` with capacity for the optional padding right after.
Then maybe an `assert!(len != expected_len)` would help the compiler too in that method, and the reader as well, to see right there, that this method isn't (shouldn't be) called when the lengths are the same.

But these are mostly just nipicks...

## Second commit:

~~Shouldn't the `webcodecs` decoder also return the newly added `DecoderNoOutputFrame`, instead of constructing a generic `DecoderError` with the same meaning?~~
(Immediate edit: I see a later commit fixes this, so never mind.)

Also, this opens up the possibility of handing this callback into the WebCodecs decoder itself somehow, instead of awkwardly storing `last_frame` in the middle. But this is probably best left for a future refactor, for when a couple frames could be fed into the decoder in advance to undo the decoding delay (only for H.264) - which in turn will also unblock updating OpenH264, and make us more correct too!

Is `cb.take()` needed to "convert" between `FnMut`/`FnOnce`, or to protect against (or even allow/disallow) decoder implementations calling the passed in callback multiple times? I guess they can call it, but only the first will do anything, because of this, right? Could this be documented/asserted somehow? Like tracing or erroring out or even assert-failing if the local `cb` is no longer `Some` here.

## Third commit:

LGTM, nice catch with the change to `BitmapFormat::length_for_size`!
Shouldn't that added `reserve` also be `reserve_exact` instead, like in other places?

## Fourth commit:

Nice deduplication by introducing `process_pixels`! Also copying of the chroma data within the vec.

EDIT: It would be nice to try testing all the accepted `VideoPixelFormat` values (produced by various decoder implementations behind the WebCodecs API).

## Fifth commit:

LGTM, nice simplification.

## Sixth commit:

Uhm, okay. Stared at it for quite a while, and seems fine.

## Seventh commit:

It would be nice to note that `crop` _appends_ to the new `out` param (which is how the planar YUV[A] layout is built into `scratch`, by calling it 3 or 4 times). Nice optimization though!

EDIT: Also, it seems that I mangled a comment in way back in https://github.com/ruffle-rs/ruffle/commit/2ad43994b8f1e507905e45b0b8dca7c4d81eea7a :eyes:
I don't know how that happened...
Anyway, those two leftover lines:

```
        //(most commonly: unused pieces of macroblocks)
        // Bitmap at the moment does not allow these gaps, so we need to remove them.
```

should either be moved back into/after

```
        // Removing the unwanted pixels on the right edge
        // by squishing all the rows tightly next to each other.
```

or just deleted entirely.

But it's fine if you don't want to pull this minor cleanup into the scope of this PR.

torokati44:
Now that I'm done with a thorough manual review, I thought I'd ask an LLM as well. It flagged three bugs, all related to WebCodecs (rephrased into my own words):

1.: `VideoFrame` objects are leaked (`.close()` not called on them) in the rare case when returning `DecoderError` with `"Unsupported pixel format: ...`. This is probably because this had to be rebased across https://github.com/ruffle-rs/ruffle/pull/24335.

2.: Just a few lines below this, `data.reserve_exact(size_in_bytes);` will allocate an _additional_ frame's worth of bytes, when usually `data` already contains the previous frame, so it will be double the capacity of what it needs to be. Probably adding a `clear` call before `reserve_exact` would be worth it.

3.: Allegedly, when the async `output` callback hasn't fired since the previous call of `decode_frame_dyn` yet, the last decoded (stale) frame is submitted again to the (newly introduced) callback, instead of either reporting a "no output frame yet" error, or just not calling the callback with the same `DecodedFrame` again. Visually it's the same (image displayed by video doesn't change), but it's a silent "soft-failure".

And the rest is quoted directly from the machine:

> Biggest process gap: the screenvideo in-place rewrite is the riskiest change and is effectively untested — its only fixture (rufflelogo.flv) holds exactly one video tag, a keyframe, so the interframe and last_frame-reuse paths it rewrites never run.
> Worth mentioning in the commit message: Bitmap::new also fixes a latent panic on master (short borrowed slices hit an out-of-bounds range), not just an optimization. Minor: openh264's within-call scratch aliasing between decoder input and YUV output deserves a comment explaining why it's safe.

EDIT: At least the "Hey Hey 16K" animation and the FLV file linked from https://github.com/ruffle-rs/ruffle/pull/5808 both still play fine for me.

moulins:
> Is `cb.take()` needed to "convert" between `FnMut/FnOnce`, or to protect against (or even allow/disallow) decoder implementations calling the passed in callback multiple times?

Ideally, `decode_from_dyn` would take an `FnOnce` to prevent implementations from calling the callback multiple times, but dyn-compatibility prevents it (`&mut dyn FnOnce()` is mostly useless, and `Box<dyn FnOnce()>` would require an allocation), so we're stuck with `FnMut` and this Option dance.

----

Most of your remarks should be addressed:

- I added some comments to `Bitmap::resize_data` and `vp6::crop`;
- I added the `Error::DecoderMultipleOutputFrames` variant in case a decoder calls its callback multiple times per `decode_frame` call;
- I slightly refactored the webcodecs decoder to fix the issues found by the LLM.

render/src/bitmap.rs
torokati44: It's probably a reporting error due to inlining...?

torokati44 (approved):
LGTM, thanks for addressing my feedback! The perf improvements are welcome.

PR #24136, opened 2026-07-06
core/render: Implement BitmapData.generateFilterRect

falpi:
## What

Implements `BitmapData.generateFilterRect` (previously a stub returning
`sourceRect` unchanged, for both AVM1 and AVM2) and makes the wgpu
`applyFilter` write the filter's *expanded* output — the pixels a filter
produces beyond the source rect (blur margins, drop shadow distance
offset), which were previously discarded because the filter was rendered
into a texture the size of the source rect.

## Why

The two halves fix a silent rendering gap that affects, among others,
every default Flex tooltip. Flex's `mx.graphics.RectangularDropShadow`
(used by the halo `ToolTipBorder` skin, also under the Flex 4 Spark
theme) pre-renders its shadow like this:

1. draws a small round rect onto a `BitmapData`;
2. calls `generateFilterRect` to learn how far a `DropShadowFilter`
   expands it, deriving the shadow thickness on each edge;
3. calls `applyFilter` and slices the result into corner/edge bitmaps,
   which it then tiles with `beginBitmapFill`.

With `generateFilterRect` returning the source rect unchanged, every
thickness came out as 0, no slices were created, and `drawShadow()`
drew nothing — Flex tooltips lost their drop shadow (and the rounded
bottom edge, which *is* the shadow). The same mechanism is used by the
mx preloader progress bars.

The two changes must ship together: with only `generateFilterRect`
fixed, the perimeter slices would read the destination bitmap's initial
fill (opaque white by default) instead of shadow pixels, which would
look worse than the current behavior.

## How

- `Filter::calculate_dest_margins` (render) — computes how far a
  filter's output extends beyond a source of a given pixel size,
  rounding outwards to whole pixels. Single source of truth so that
  `generateFilterRect` (what we report) and `applyFilter` (what we
  write) always agree.
- `generateFilterRect` implemented natively for AVM2 and AVM1 on top of
  it. The AVM2 ShaderFilter special case (Flash reports the whole
  bitmap) is preserved, moved from ActionScript into the native method.
- wgpu `apply_filter` copies the source region into a zero-padded
  intermediate texture, runs the filter over the whole padded area, and
  shifts the destination point back by the margins. Padding is skipped
  when the margins are empty or the padded texture would exceed
  `max_texture_dimension_2d`.

## Notes

- Margins derive from `Filter::calculate_dest_rect`, whose blur
  expansion is intentionally approximate vs. Flash (see `PASS_SCALES`),
  so the reported rect may be slightly larger than Flash Player's.
  Internal consistency between the two APIs — which is what
  `RectangularDropShadow`-style code depends on — is guaranteed.
- A filter applied to a 0×0 source rect is still a noop (pre-existing
  FIXME, comment updated).

## Checklist


docs/fix-bitmapdata-filter-expansion.html
torokati44, line 1: This seems unnecessary, at least in this form.

torokati44:
Referencing (mostly for the test, as it's ancient): https://github.com/ruffle-rs/ruffle/pull/12417

torokati44:
> - [X] All of my commits are properly scoped, compile successfully, and pass all tests.

The last part doesn't look like it holds...

docs/fix-bitmapdata-filter-expansion.html
falpi: I make it a habit to include brief documentation to facilitate reviews and help the LLM retain context. I generally separate this type of file into a dedicated commit so it can easily be excluded from the PR.

falpi:
> > * [x]  All of my commits are properly scoped, compile successfully, and pass all tests.
> 
> The last part doesn't look like it holds...

My oversight. I’ve already integrated the checks into my CI/CD scripts, but I forgot the imgtests. I’ll check and fix it.

kjarosh:
> and fully understand the changes within

Can you describe, using _your own_ words, what `BitmapData.generateFilterRect` does and why it's needed at all?

falpi:
> > and fully understand the changes within
> 
> Can you describe, using _your own_ words, what `BitmapData.generateFilterRect` does and why it's needed at all?

As I explained in the HTML document included in the branch "Every default Flex tooltip renders in Ruffle without its drop shadow, and with a squarish bottom edge instead of the rounded one Flash Player shows. The shadow is not a display-object filter: Flex pre-renders it through BitmapData.generateFilterRect + applyFilter and copies it in slices. generateFilterRect was a stub in Ruffle, so the slice thicknesses came out as zero and the shadow silently vanished."

Below, I am attaching an example of how default tooltips are rendered in Flex: the top box shows Ruffle (pre-PR), while the bottom one shows FP.

immagine


I would like to point out that every time I submit a PR, I am subjected to an interrogation—as if I had committed murder. To date, not a single one of my PRs has been accepted, even though I have always provided the requested explanations. It is not a problem for me: I am implementing fixes to get a real enterprise application running on Ruffle, and thanks to the LLM, I have been able to do so very quickly. Submitting PRs seemed like a courtesy, but if it is unwelcome—presumably due to the use of the LLM—just say so, and I will stop wasting time responding and submitting them.

docs/fix-bitmapdata-filter-expansion.html
torokati44: That's nice and good, but the commits and their contents of the PR are supposed to represent what you'd like to see in the final repo itself upon merge. This just looks out of place here, but it would be perfectly fine as a comment here.

falpi: Sorry, but I didn't quite catch what you mean: what exactly do you suggest I do? If you mean copying the document inline here, it's too large. In that case, if it really doesn't belong here, I think I'll focus the documentation on a branch of my own.

falpi: I removed this and all other similar cases from my branches, consolidating all branch documentation into my fork.

torokati44:
The problem is _not at all_ the use of LLMs, it's _purely_ about the result. And this is how we can keep the quality up, by not just trusting anything an LLM spits out, but actually understanding whether it's any good. Even if that is becoming increasingly controversial...
And keep in mind, there aren't many maintainers, and they don't have all the time in the world. Many times, even perfectly good PRs made by regular contributors (or maintainers) spend a long time in review limbo because of this.

falpi:
> The problem is _not at all_ the use of LLMs, it's _purely_ about the result. And this is how we can keep the quality up, by not just trusting anything an LLM spits out, but actually understanding whether it's any good. Even if that is becoming increasingly controversial...
> And keep in mind, there aren't many maintainers, and they don't have all the time in the world. Many times, even perfectly good PRs made by regular contributors (or maintainers) spend a long time in review limbo because of this.

Okay, I understand. The reason I always try to document everything well is to make your life at least a little easier, because I realize it's difficult to keep up with the current PR pace following the LLM boom. Regards

PR #24113, opened 2026-07-04
ci: Report macOS universal binary bundle notarization failure on Discord

torokati44:
## Description

Since Apple requires us to agree to their updated developer agreement from time to time, let's try to be notified about this.

Mostly copied from the Firefox extension builder Dockerfile reproducibility tester workflow.
I think this will post into the #github-updates channel, don't know how useful that is (it tends to be spammy).
Maybe we could add some pings or something...? Or set up a different webhook for a different channel...

Note: We could consider switching from https://github.com/th0th/notify-discord/ to the more recently updated https://github.com/stegzilla/discord-notify.

## Testing

Don't agree to the new Apple agreement until tomorrow, and see if a message pops up on Discord about it.

## Checklist

- [X] I, a human, have self-reviewed this PR and fully understand the changes within.
- [ ] I have [made or updated tests](https://github.com/ruffle-rs/ruffle/blob/master/CONTRIBUTING.md#test-guidelines) where possible.
- [X] All of my commits are properly scoped, compile successfully, and pass all tests.
- [X] This PR does not make sense to split up into smaller PRs.
- [X] An LLM was involved in the authoring of this code.

PR #24145, opened 2026-07-06
ci: Switch Discord notification action to one that doesn't need Docker

torokati44:
## Description

Follow-up to https://github.com/ruffle-rs/ruffle/pull/24113, as that didn't work out: https://github.com/ruffle-rs/ruffle/actions/runs/28760270801/job/85275571829#step:13:10

This action doesn't need Docker, so it has a higher chance of working on macOS - it's also more recently updated.

I split the reporting of the Dockerfile tester into its own job, because this action works a bit differently - it doesn't take a job ID in itself, so reporting the status of the main job (which can fail for multiple reasons) seemed more logical this way.

The macOS one only checks one step, not a whole job, so I kept that one as is.

As for the "outcome", and "result" mismatch, see:
 - https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#steps-context
 - https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#needs-context

## Testing

I can trigger a manual run of the Dockerfile tester workflow after merging, and the Release workflow will run anyway.
As for before merging... That would have to be done in someone's fork and their own Discord server.

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -345,14 +345,17 @@ jobs:
           xcrun stapler staple package/Ruffle.app
 
       - name: Report notarization failure on Discord
-        uses: th0th/notify-discord@v0.4.1
+        uses: stegzilla/discord-notify@v2
         if: ${{ steps.notarize-bundle.outcome == 'failure' && github.repository == 'ruffle-rs/ruffle' }}
         continue-on-error: true
-        env:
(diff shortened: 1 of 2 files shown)


kjarosh (approved):
Can you test it after merging by running the extension workflow manually?

torokati44:
That's the plan! And the release workflow should test the other part this night (Apple agreement is not renewed yet).

PR #24049, opened 2026-06-23
avm1: Recursively serialize constructed objects as typed objects

danielhjacobs:
## Description

**Review by commit.**

- Fixes avm1/amf0_omnibus_suite
- Fixes #16356
- Allows serialization of typed objects properly in recursive_serialize

As far as I can tell, this truly fixes AMF0 serialization for NetConnection/LocalConnection.

https://github.com/danielhjacobs/ruffle/compare/add-full-amf-test-suite...danielhjacobs:ruffle:fix-amf0-serialize-object

## Testing

There are multiple tests, including a large AMF test that tests a suite of serialization methods, but also http://www.eugenicsarchive.org/eugenics/list2.pl works fully.

## Checklist

- [x] I, a human, have self-reviewed this PR and fully understand the changes within.
- [x] I have [made or updated tests](https://github.com/ruffle-rs/ruffle/blob/master/CONTRIBUTING.md#test-guidelines) where possible.
- [x] All of my commits are properly scoped, compile successfully, and pass all tests.
- [x] This PR does not make sense to split up into smaller PRs.
- [x] An LLM was involved in the authoring of this code.

PR #24155, opened 2026-07-08
avm1: Serialize top-level objects with class definitions as appropriate

danielhjacobs:
## Description

This contains just the parts of #24049 necessary to allow proper serialization of top-level typed objects (not nested typed objects, which will be handled in a follow-up).

## Testing

Check existing test, including new non-known failure.

## Checklist


danielhjacobs:
PRs, in order:

1) #24156
2) #24155
3) #24161
4) #24096

core/src/avm1/amf.rs
```diff
@@ -2,10 +2,37 @@ use crate::avm1::{Activation, Attribute, NativeObject, Object, Value};
...
 use std::rc::Rc;
 
+/// Get class alias of object
+fn object_class_alias<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    object: Object<'gc>,
+) -> Option<String> {
+    // Check for a mutated local 'constructor' property (matches Flash's forged_ctor behavior)
+    // Fall back to the hidden '__constructor__' property to bypass the prototype chain
+    // (matches Flash's forged_proto behavior by ignoring dynamic __proto__ reassignments)
```
Lord-McSweeney: What's `forged_proto`?

core/src/avm1/amf.rs
```diff
@@ -2,10 +2,37 @@ use crate::avm1::{Activation, Attribute, NativeObject, Object, Value};
...
+    let constructor_key = AvmString::new_utf8(activation.gc(), "constructor");
+    let magic_constructor_key = AvmString::new_utf8(activation.gc(), "__constructor__");
```
Lord-McSweeney: These should use `istr!`

core/src/avm1/runtime.rs
```diff
@@ -581,6 +581,31 @@ impl<'gc> Avm1<'gc> {
...
+    ) -> Option<AvmString<'gc>> {
+        let is_case_sensitive = Self::is_case_sensitive(swf_version);
+        let registry = if is_case_sensitive {
+            &self.env_case_sensitive.constructor_registry
+        } else {
+            &self.env_case_insensitive.constructor_registry
+        };
+        // Iterate through the PropertyMap to find the matching constructor reference.
+        // If multiple aliases point to the same constructor, this will grab the one
+        // based on the PropertyMap's iteration order.
```
Lord-McSweeney, line 599: Is there test coverage checking what happens if there are multiple aliases for one constructor? If not, could you add it?

core/src/avm1/amf.rs
danielhjacobs: Ah, it was mentioning a name from a specific test, but probably shouldn't do that:

https://github.com/danielhjacobs/ruffle/blob/020a0cb29fd6d1569721bbe0c5f40dcdc2e3ffc0/tests/tests/swfs/avm1/amf_serialize_typed_objects/test.as#L51

core/src/avm1/runtime.rs
danielhjacobs: https://github.com/danielhjacobs/ruffle/blob/020a0cb29fd6d1569721bbe0c5f40dcdc2e3ffc0/tests/tests/swfs/avm1/amf_serialize_typed_objects/test.as#L21-L25

core/src/avm1/amf.rs
```diff
@@ -2,10 +2,37 @@ use crate::avm1::{Activation, Attribute, NativeObject, Object, Value};
...
+    // Check for a mutated local 'constructor' property
+    // Fall back to the hidden '__constructor__' property to bypass the prototype chain
+    // (matches Flash's behavior by ignoring dynamic __proto__ reassignments)
+    let constructor_key = istr!("constructor");
+    let magic_constructor_key = istr!("__constructor__");
+    let ctor = if object.has_own_property(activation, constructor_key) {
+        object.get(constructor_key, activation)
+    } else {
+        object.get(magic_constructor_key, activation)
+    };
```
Lord-McSweeney, line 34: What happens if an object has getters for either of these two properties?

danielhjacobs: I'll add test cases to check.

danielhjacobs: Very good question. AMF serialization actually ignores AVM1 getters in the prototype chain entirely and serializes them as undefined, also meaning the check for a constructor needs to ignore getters. See the new test cases (11 and 12) and the new logic in `recursive_serialize` and the helper function `get_raw_data_property` added to `object_class_alias`.

This also means traces in those custom getters do not print.

core/src/avm1/amf.rs
```diff
@@ -2,10 +2,85 @@ use crate::avm1::{Activation, Attribute, NativeObject, Object, Value};
...
+    // Fall back to the hidden '__constructor__' property to bypass the prototype chain
+    // (matches Flash's behavior by ignoring dynamic __proto__ reassignments)
+    let constructor_key = istr!("constructor");
+    let magic_constructor_key = istr!("__constructor__");
+    // Helper to safely fetch a raw data property from the prototype chain.
+    // Explicitly bypasses virtual properties (getters/setters) to match Flash's AMF behavior.
+    let get_raw_data_property = |act: &mut Activation<'_, 'gc>,
+                                 start_obj: Object<'gc>,
+                                 key: AvmString<'gc>|
+     -> Option<Value<'gc>> {
```
danielhjacobs: I hate the way this looks, but it's what `cargo fmt --all` did 🤷

kjarosh: Can't you just create a function instead of a lambda?

danielhjacobs: Ok

core/src/avm1/amf.rs
```diff
@@ -2,10 +2,85 @@ use crate::avm1::{Activation, Attribute, NativeObject, Object, Value};
 use crate::string::AvmString;
 use flash_lso::amf0::read::AMF0Decoder;
 use flash_lso::amf0::writer::{Amf0Writer, CacheKey, ObjWriter};
-use flash_lso::types::{Element, ObjectId, Reference, Value as AmfValue};
+use flash_lso::types::{ClassDefinition, Element, ObjectId, Reference, Value as AmfValue};
+use ruffle_macros::istr;
 use std::collections::BTreeMap;
 use std::rc::Rc;
 
+/// Get class alias of object
```
kjarosh: This doc doesn't really add anything that's not already in the signature. You could describe what a class alias is in AMF instead.

core/src/avm1/runtime.rs
```diff
@@ -581,6 +581,31 @@ impl<'gc> Avm1<'gc> {
...
+        for (alias, registered_constructor) in registry.iter() {
+            // Compare the objects by pointer identity
```
kjarosh: This comment basically describes the line below, you could explain _why_ you're comparing pointers instead.

core/src/avm1/runtime.rs
```diff
@@ -581,6 +581,31 @@ impl<'gc> Avm1<'gc> {
         registry.get(symbol, is_case_sensitive).copied()
     }
 
+    /// Finds the alias string associated with a given constructor function.
+    /// This is specifically required for AMF0 TypedObject serialization.
+    pub fn get_alias_by_constructor(
```
kjarosh: Why are we calling it "alias" instead of "symbol"? I've never heard the term "alias" used in AVM1.

core/src/avm1/amf.rs
```diff
@@ -2,10 +2,85 @@ use crate::avm1::{Activation, Attribute, NativeObject, Object, Value};
...
+fn object_class_alias<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    object: Object<'gc>,
+) -> Option<String> {
+    // Check for a mutated local 'constructor' property
+    // Fall back to the hidden '__constructor__' property to bypass the prototype chain
```
kjarosh: Are those supposed to be here? They are referring to code below and it looks like they are kinda duplicated.

core/src/avm1/amf.rs
kjarosh: Note that a function can be declared within another function as well, so a lambda is required only when you're capturing variables, but I didn't see any.

danielhjacobs: Good to know, I think a separate function looks better for this anyway.

core/src/avm1/runtime.rs
danielhjacobs: alias isn't right, yeah, though I don't think symbol is either. class_name is probably most correct, will switch to that.

core/src/avm1/amf.rs
```diff
@@ -154,10 +235,20 @@ pub fn recursive_serialize<'gc>(
 ) {
     // Reversed to match flash player ordering
     for element_name in obj.get_keys(activation, false).into_iter().rev() {
-        if let Ok(elem) = obj.get(element_name, activation) {
-            let name = element_name.to_utf8_lossy();
-            serialize_value_to_writer(activation, name.as_ref(), elem, writer);
-        }
+        let elem = if is_virtual_property(activation, obj, element_name) {
```
Lord-McSweeney: Doesn't this mean that if an object defines a property getter on its prototype, but the object itself has that property as a normal field, the property will be skipped?

danielhjacobs: Does it?

`is_virtual_property` does this:

```rs
        if obj.has_own_property(activation, key) {
            return obj.has_own_virtual(activation, key);
        }
 ```
 
 `has_own_property` and `has_own_virtual` do this:
 
 ```rs
     pub(crate) fn has_own_virtual(
        self,
        name: AvmString<'gc>,
    ) -> bool {
        if let Some(zuper) = self.as_super_object() {
            // `super` forwards property membership tests to its underlying object,
            // even though it can't be enumerated.
            return zuper.this().has_own_virtual(activation, name);
        }

        self.0
            .borrow()
            .properties
            .get(name, activation.is_case_sensitive())
            .is_some_and(|property| {
                property.is_virtual() && property.allow_swf_version(activation.swf_version())
            })
    }


    /// Checks if the object has a given named property on itself (and not,
    /// say, the object's prototype or superclass)
    pub fn has_own_property(
        self,
        name: AvmString<'gc>,
    ) -> bool {
        if let Some(zuper) = self.as_super_object() {
            // `super` forwards property membership tests to its underlying object,
            // even though it can't be enumerated.
            return zuper.this().has_own_property(activation, name);
        }

        // Note that `hasOwnProperty` does NOT return true for display object properties.
        self.0
            .borrow()
            .properties
            .contains_key(name, activation.is_case_sensitive())
    }
```

danielhjacobs: Added a new test case (case 20)

core/src/avm1/runtime.rs
```diff
@@ -581,6 +581,33 @@ impl<'gc> Avm1<'gc> {
...
+    pub fn get_class_name_by_constructor(
+        &self,
+        swf_version: u8,
+        constructor: Object<'gc>,
+    ) -> Option<AvmString<'gc>> {
+        let is_case_sensitive = Self::is_case_sensitive(swf_version);
+        let registry = if is_case_sensitive {
+            &self.env_case_sensitive.constructor_registry
+        } else {
+            &self.env_case_insensitive.constructor_registry
```
Lord-McSweeney, line 595: Seems like this line isn't covered by tests?

core/src/avm1/amf.rs
```diff
@@ -2,10 +2,89 @@ use crate::avm1::{Activation, Attribute, NativeObject, Object, Value};
...
 use flash_lso::amf0::writer::{Amf0Writer, CacheKey, ObjWriter};
-use flash_lso::types::{Element, ObjectId, Reference, Value as AmfValue};
+use flash_lso::types::{ClassDefinition, Element, ObjectId, Reference, Value as AmfValue};
+use ruffle_macros::istr;
 use std::collections::BTreeMap;
 use std::rc::Rc;
 
+/// Helper to safely fetch a raw data property from the prototype chain.
+/// Explicitly bypasses virtual properties (getters/setters) to match Flash's AMF behavior.
+fn get_raw_data_property<'gc>(
```
Lord-McSweeney: Should this function and the other function `is_virtual_property` below maybe be moved to `core/src/avm1/object.rs`, or `core/src/avm1/object/script_object.rs`?

core/src/avm1/amf.rs
Lord-McSweeney: nit (non-blocking): If it's not too hard, could you move this fix (skip getters) to a separate commit? It's a bit out of the scope of the commit it's in right now.

Lord-McSweeney: By my understanding, if the property exists as a getter on an object's prototype, or prototype of prototype, etc, `is_virtual_property` would always return `true`, causing the property to be serialized as `undefined` regardless of the property on the object's own `properties`. Does Flash really do this?

core/src/avm1/amf.rs
```diff
@@ -154,10 +235,20 @@ pub fn recursive_serialize<'gc>(
...
-        }
+        let elem = if is_virtual_property(activation, obj, element_name) {
+            // Flash never evaluates getters during AMF serialization;
+            // it serializes them as `undefined` (0x06).
+            Value::Undefined
+        } else {
+            // Not a getter, safe to retrieve
+            match obj.get(element_name, activation) {
+                Ok(val) => val,
+                Err(_) => continue, // Maintain original behavior: skip on error
```
Lord-McSweeney: This "Maintain original behavior" comment makes sense in the context of the PR, but doesn't make sense on its own. Could you remove the comment?

core/src/avm1/runtime.rs
danielhjacobs: This is why I asked if there was a way to run the same test with multiple SWF versions with a single test.toml, so my one fixed test could test it too.

It was added to match this:

```rs
    pub fn get_registered_constructor(
        &self,
        swf_version: u8,
        symbol: AvmString<'gc>,
    ) -> Option<Object<'gc>> {
        let is_case_sensitive = Self::is_case_sensitive(swf_version);
        let registry = if is_case_sensitive {
            &self.env_case_sensitive.constructor_registry
        } else {
            &self.env_case_insensitive.constructor_registry
        };
        registry.get(symbol, is_case_sensitive).copied()
    }
```

I'll add a small test for it.

core/src/avm1/amf.rs
danielhjacobs: This is incorrect as far as I can tell. If `obj.has_own_property(key)`, I `return obj.has_own_virtual(activation, key);` before ever checking the `obj.proto`. I added more test cases for this: see case 20, 23, and 24. Especially test 23 proves this, as it does serialize "normal_instance_value".

core/src/avm1/runtime.rs
danielhjacobs: By adding this test case, I found that Ruffle keeps the constructor casing of the original key when registering two constructors with different cases but the same case-insensitive values, while Flash updates the casing. Had to add avm1: When updating case-insentive contructors, change their keys too for the test to work.

core/src/avm1/amf.rs
danielhjacobs: I think this particular code of `has_own_virtual` is the most relevant for what you're asking:

https://github.com/ruffle-rs/ruffle/blob/34f9289f30ad074ea87feb8c97d79d7581f1c808/core/src/avm1/object/script_object.rs#L797-L803

Lord-McSweeney (approved):

danielhjacobs:
Rust nightly issue unrelated: https://github.com/rust-lang/rust/issues/159815

PR #24156, opened 2026-07-08
tests: Add known failure to thoroughly test AVM1 AMF0 serialization 

danielhjacobs:
## Description

This contains just the added test from #24049

## Testing

It is a test.

## Checklist


danielhjacobs:
PRs, in order:

1) #24156
2) #24155
3) #24161
4) #24096

tests/tests/swfs/avm1/amf0_omnibus_suite/test.as
```diff
@@ -0,0 +1,237 @@
+// ============================================================================
+// THE OMNIBUS AVM1 / AMF0 PROTOCOL REGRESSION SUITE
```
kjarosh: What's the "omnibus"?

tests/tests/swfs/avm1/amf0_serde_suite/test.as
```diff
@@ -0,0 +1,237 @@
...
+
+// --- 2. CONSTRUCT MASTER PAYLOAD ---
+var payload = new Object();
+
+// Primitives
+payload.p_num = 1337.5;
+payload.p_nan = NaN;
+payload.p_pos_inf = Infinity;
+payload.p_neg_inf = -Infinity;
+payload.p_str = "The quick brown AMF fox";
```
kjarosh, line 126: What about strings with non-ASCII characters? Surrogate pairs? Unpaired surrogates? Non-printable characters?

tests/tests/swfs/avm1/amf0_serde_suite/test.as
```diff
@@ -0,0 +1,237 @@
...
+Object.registerClass("com.tests.RegClass", RegClass);
+Object.registerClass("WeirdAliasString", MismatchedClass);
+Object.registerClass("Alias.One", DoubleRegClass);
+Object.registerClass("Alias.Two", DoubleRegClass);
+
+// --- 2. CONSTRUCT MASTER PAYLOAD ---
+var payload = new Object();
+
+// Primitives
+payload.p_num = 1337.5;
```
kjarosh, line 117: Negative numbers? Negative/positive zeros?

tests/tests/swfs/avm1/amf0_omnibus_suite/test.as
```diff
@@ -0,0 +1,237 @@
...
+    trace(prefix + ".nested_regular_prop: " + d.nested_container.regular_prop);
+}
+
+// --- 1. SETUP REGISTRATIONS ---
+Object.registerClass("com.tests.RegClass", RegClass);
+Object.registerClass("WeirdAliasString", MismatchedClass);
+Object.registerClass("Alias.One", DoubleRegClass);
+Object.registerClass("Alias.Two", DoubleRegClass);
+
+// --- 2. CONSTRUCT MASTER PAYLOAD ---
```
kjarosh: Why this payload is the "master" one? Compared to what? Why "master"?

tests/tests/swfs/avm1/amf0_omnibus_suite/test.as
```diff
@@ -0,0 +1,237 @@
...
+payload.p_null = null;
+payload.p_undef = undefined;
+
+// Natives
+payload.n_date = new Date(1672531200000); // 2023-01-01 UTC
+payload.n_xml = new XML("<root><child id='avm1'>test</child></root>");
+
+// Problematic Types
+payload.sp_func = function(x) { return x * 42; };
+var mock:Object = _root.createEmptyMovieClip("mockMC", _root.getNextHighestDepth());
```
kjarosh: Why is it called a "mock"? It's a real DO, no?

tests/tests/swfs/avm1/amf0_serde_suite/test.as
```diff
@@ -0,0 +1,237 @@
...
+var mock:Object = _root.createEmptyMovieClip("mockMC", _root.getNextHighestDepth());
+mock.custom_var = "i_am_a_display_object";
+payload.sp_mc = mock;
+
+// Arrays
+payload.arr_strict = ["index_0", "index_1", "index_2"];
+
+var arrSparse = new Array();
+arrSparse[0] = "sparse_0";
+arrSparse[4] = "sparse_4";
```
kjarosh, line 152: If we're working on sparse arrays you can just throw in large indices, like `i32::MAX`, `i32::MAX+1`, `u32::MAX`, `u32::MAX+1`, `i64::MAX`, `i64::MAX+1`, `u64::MAX`, `u64::MAX+1`.

tests/tests/swfs/avm1/amf0_serde_suite/test.as
```diff
@@ -0,0 +1,237 @@
...
+payload.arr_sparse = arrSparse;
+
+var arrMixed = new Array();
+arrMixed.push("mixed_0");
+arrMixed["string_key"] = "string_value";
+payload.arr_mixed = arrMixed;
+
+var arrNeg = new Array();
+arrNeg[0] = "pos_0";
+arrNeg[-1] = "neg_1";
```
kjarosh, line 170: What about non-integer numbers?

tests/tests/swfs/avm1/amf0_serde_suite/test.as
```diff
@@ -0,0 +1,237 @@
...
+};
+
+// ====================================================================
+// CHANNEL A: SHAREDOBJECT (Disk Test)
+// ====================================================================
+var so = SharedObject.getLocal("avm1_omnibus", "/");
+so.data.payload = payload;
+        
+trace("SO.getSize() pre-flush: " + so.getSize());
+so.flush();
```
kjarosh, line 229: Would be nice to assert on the shared object as well. Maybe we could add a mechanism to our test framework where we could assert on stored lso?

danielhjacobs: Not a bad idea, but feels to me like it's outside the scope of this PR.

danielhjacobs: https://github.com/ruffle-rs/ruffle/pull/24198

tests/tests/swfs/avm1/amf0_serde_suite/test.as
```diff
@@ -0,0 +1,273 @@
...
+Object.registerClass("com.tests.RegClass", RegClass);
+Object.registerClass("WeirdAliasString", MismatchedClass);
+Object.registerClass("Alias.One", DoubleRegClass);
+Object.registerClass("Alias.Two", DoubleRegClass);
+
+// --- 2. CONSTRUCT PAYLOAD ---
+var payload = new Object();
+
+// Primitives
+payload.p_num = 1337.5;
```
kjarosh, line 117: In addition to float, we should also cover ints (so 1337, -1337).

tests/tests/swfs/avm1/amf0_serde_suite/test.as
kjarosh: Thank you, that's great! Maybe we could also rewrite existing sol tests to use that new format.

I think we can merge this one, and then the one with sol asserts, which could at the same time enable the test from this PR to assert on sols.

tests/tests/swfs/avm1/amf0_serde_suite/test.as
```diff
@@ -0,0 +1,273 @@
...
+payload.t_forged_ctor = forgedCtor;
+
+var forgedProto = new Object();
+forgedProto.__proto__ = RegClass.prototype;
+forgedProto.type = "forged_p";
+payload.t_forged_proto = forgedProto;
+
+payload.t_double = new DoubleRegClass();
+
+// Deep Nesting
```
kjarosh, line 199: Can you also add a case for an object cycle? E.g.

```as
var o = new Object();
o.prop = o;
```

Of course assuming FP can handle it... if not, we need a separate test

tests/tests/swfs/avm1/amf0_serde_suite/test.toml
kjarosh, line 1: Can we not call it a "regression" test? We use these tests as the basis for how FP behaves and how Ruffle should behave. Regression tests suggest we're just arbitrarily asserting on stuff to make sure we don't regress or current behavior but that's not the case.

kjarosh (approved):
LGTM after addressing comments.

tests/tests/swfs/avm1/amf0_serde_suite/test.as
danielhjacobs: I can do that, but Ruffle will end up serializing this with an incorrect reference index, even with all my current PRs, due to https://github.com/ruffle-rs/rust-flash-lso/issues/115

danielhjacobs:
Just as an FYI, now even after my fourth PR, two issues will remain in this test (unrelated to my changes, but will keep this as a `known_failure`)

1) Due to https://github.com/ruffle-rs/rust-flash-lso/issues/115, Ruffle serializes the cyclic object with the wrong reference index.
~~2) Due to https://github.com/ruffle-rs/rust-flash-lso/issues/116, the serialization of NaN is incorrect (both are valid NaN values in IEEE-754)~~

kjarosh (approved):

PR #24176, opened 2026-07-10
avm1: Remove playback queue from `Sound`

ChrisCPI:
## Description

Undoes most of #21163. The inclusion of this feature was a misinterpretation of how loaded sounds behave on the desktop Flash Player projector. It actually loads sounds synchronously, meaning that anything immediately after `loadSound()` (such as `start()`) would work, giving the illusion that a queue exists.

However, remote files (a test for which has been added) do *not* load synchronously, and plays do not appear to be queued in that case. In addition, local files do not load synchronously when running the SWF in a Flash browser.

Note that this was manually undone since changes were made to `Sound` since that PR, so it's possible there's a mistake, though I'm pretty sure there isn't.

## Testing

- Added the `avm1/sound_load_start_remote` file to test loading remote MP3s.
- The `avm1/sound_load_start` test has been marked as a known failure, and has also been modified to include traces for when `onLoad` is fired, to demonstrate how it is fired right after calling `loadSound()`.
- The `avm1/sound_multiple_load` test has been marked as a known failure.

I'm still a bit confused as to what is expected of the tests in this scenario, so let me know if anything else needs done.

## Checklist

- [x] An LLM was involved in the authoring of this code. (used in authoring `server.py` in the added test)


tests/tests/swfs/avm1/sound_multiple_load/test.toml
```diff
@@ -1,4 +1,8 @@
 num_ticks = 100
+# The FP projector loads local MP3s synchronously, which means
+# that any start() calls immediately following loadSound()
+# do succeed. This does not happen in a Flash browser.
```
kjarosh, line 4: If that does happen in flashplayer when loading remote resources, we should implement this behavior as well. Running SWFs in standalone FP was also the thing in the past.

tests/tests/swfs/avm1/sound_load_start_remote/server.py
```diff
@@ -0,0 +1,31 @@
+import os
+from http.server import HTTPServer, BaseHTTPRequestHandler
+
+class MyHandler(BaseHTTPRequestHandler):
```
kjarosh, line 4: Is there anything special in this server that prevents us from just using `python -m http.server`?

tests/tests/swfs/avm1/sound_multiple_load/test.toml
ChrisCPI: > when loading remote resources

Do you mean when loading local files? AFAIK with these changes, Ruffle should be matching FP when it comes to loading remote files.

Also, I'm open to ideas on how the synchronous behavior could be replicated, since that would seem tricky to do.

ChrisCPI:
This was included with #23975, closing this as it is now merged

PR #24191, opened 2026-07-11
avm2: Use Error.throwError() in GraphicsPath

kjarosh:
## Description

This uses an existing method instead of throwing manually similarly to
FP, which can be observed through stack traces.

## Testing

One existing test expanded (which required some asc changes), one new test added. In addition in some tests renderer was made optional.

## Checklist


Lord-McSweeney:
Do any of the changed tests make use of the change to ASC settings?

kjarosh:
Yes, the existing test used that attribute that's why I added support for it

Lord-McSweeney (approved):

PR #24198, opened 2026-07-13
 tests: Add mechanism to test framework to assert on stored lso 

danielhjacobs:
## Description

Per https://github.com/ruffle-rs/ruffle/pull/24156#discussion_r3563680153, add a mechanism to our test framework where we could assert on stored lso

## Testing

See the new test, or add ones of your own.

## Checklist


tests/README.md
```diff
@@ -68,6 +68,9 @@ output_path = "output.txt"
 # If true, all network requests will be included in the output.
 log_fetch = false
 
+# If true, an LSO included with the test will be compared to what Ruffle saves.
+log_storage = false
```
kjarosh: This suggests we're logging something (we're not right?). Maybe it's better to list sol files explicitly here, like:

```toml
[[shared_objects]]
name = "foo"
expected = "foo.sol"
```

danielhjacobs: Fair, at first I also made these use the test log runner if `log_storage`:
```rs
    fn get(&self, name: &str) -> Option<Vec<u8>> {
        self.map.get(name).cloned()
    }

    fn put(&mut self, name: &str, value: &[u8]) -> bool {
        self.map.insert(name.into(), value.to_vec());
        true
    }

    fn remove_key(&mut self, name: &str) {
        self.map.remove(name);
    }
```
But I removed that since the better method seemed to just be a direct sol file comparison

tests/framework/src/runner.rs
```diff
@@ -376,6 +379,107 @@ impl TestRunner {
...
+                <dyn std::any::Any>::downcast_ref::<TestStorageBackend>(player.storage())
+            {
+                let stored_lsos = test_storage.get_stored_data();
+
+                for (lso_name, actual_bytes) in stored_lsos {
+                    let safe_name = lso_name.replace("/", "_").replace("\\", "_");
+                    let expected_path = self.root_path.join(format!("{safe_name}.sol"))?;
+                    let failing_path = self.root_path.join(format!("{safe_name}.failing.sol"))?;
+
+                    if std::env::var("UPDATE_EXPECTS").is_ok() {
```
kjarosh: Why would we update expected state from Ruffle? The point of the expected state is that it comes from Flash Player, not Ruffle.

tests/framework/src/runner.rs
```diff
@@ -376,6 +379,107 @@ impl TestRunner {
             ));
         }
 
+        if self.options.log_storage {
```
kjarosh: The body should be extracted into a function

tests/framework/src/backends/storage.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+impl TestStorageBackend {
+    pub fn new() -> Self {
+        Self {
+            map: HashMap::new(),
+        }
+    }
+
+    /// Exposes the inner data so we can dump the LSOs to disk.
```
kjarosh:
```suggestion
    /// Exposes the inner data so we can use them in tests.
```

tests/framework/src/options.rs
```diff
@@ -83,6 +83,18 @@ pub struct AudioAssertion {
...
 }
 
+#[derive(Clone, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct SharedObjectConfig {
+    pub name: String,
+    pub expected: String,
+    /// An optional path to write Ruffle's actual output to when this test fails,
+    /// or when `known_failure` is true. Used to track exact regression state.
+    pub actual: Option<String>,
```
kjarosh: Do we have something like that in other places? If not, I'd prefer dropping this as well and just hardcoding the name.

tests/framework/src/runner.rs
```diff
@@ -349,6 +352,159 @@ impl TestRunner {
...
 
+    fn check_shared_objects(&self, update_expects: bool) -> Result<()> {
+        for so_config in &self.options.shared_objects {
+            let player = self.player.lock().unwrap();
+            if let Some(test_storage) =
+                <dyn std::any::Any>::downcast_ref::<TestStorageBackend>(player.storage())
+            {
+                let stored_lsos = test_storage.get_stored_data();
+
+                // Look for exact match first. If not found, look for a path ending in `.../name`
```
kjarosh: Why can't we just do an exact match?

danielhjacobs: I guess we can, usually it will be localhost/test.swf/[name]

tests/framework/src/runner.rs
```diff
@@ -349,6 +352,159 @@ impl TestRunner {
...
+                    }
+                } else if !is_match {
+                    let mut mismatch_info = format!(
+                        "SharedObject mismatch for {}. Actual length: {}, Expected length: {}.",
+                        so_config.name,
+                        actual_bytes.len(),
+                        expected_bytes.len()
+                    );
+
+                    let diff = actual_bytes
```
kjarosh: Looks like the diffing logic could be extracted into a separate function (potentially useful in other parts of the framework as well).

tests/framework/src/options.rs
danielhjacobs: ruffle.expected.sol?

tests/framework/src/runner.rs
```diff
@@ -376,6 +532,9 @@ impl TestRunner {
             ));
         }
 
+        let update_expects = std::env::var("UPDATE_EXPECTS").is_ok();
```
kjarosh: Déjà vu, my original question still holds

tests/framework/src/options.rs
danielhjacobs: Oh wait, but what if there are multiple shared object files?

danielhjacobs: ruffle.[expected]?

tests/framework/src/runner.rs
danielhjacobs: This only updates actual now, only if known failure is true and it's used.

danielhjacobs: Since actual is meant to be Ruffle's output

tests/framework/src/options.rs
kjarosh: Hmm, right, in other places we don't provide the extension... For instance, in image comparisons we just provide the base name. Maybe that's something we should be doing here as well, like

```toml
[[shared_objects]]
name = "name used in SWF"
expected = "foo"
```

which would read `foo.expected.sol` and we would read `foo.ruffle.sol` if it's a known failure.

Also, the question is whether it wouldn't be better to use

```toml
[shared_objects."name used in SWF"]
expected = "foo"
```

because names have to be unique.

tests/framework/src/runner.rs
kjarosh: Yeah, I think it makes sense, it's better to be explicit in tests and we can also have multiple SWFs in a test

tests/framework/src/runner.rs
kjarosh: Yeah, but we don't have such mechanism in other places, and passing an env variable looks weird especially that we have dedicated test opts. I would just drop it because the dev can always remove the file and it should regenerate and the test should fail. It's a simple solution that we're using in all other places.

tests/framework/src/runner.rs
```diff
@@ -349,6 +352,119 @@ impl TestRunner {
         Ok(())
     }
 
+    fn diff_binary(actual: &[u8], expected: &[u8], label: &str) -> Option<String> {
```
moulins: IMHO, this should live in a separate submodule `runner::shared_object`, like image comparisons.

tests/framework/src/runner.rs
```diff
@@ -349,6 +352,119 @@ impl TestRunner {
...
+
+    fn check_shared_objects(&self) -> Result<()> {
+        for (name, so_config) in &self.options.shared_objects {
+            let player = self.player.lock().unwrap();
+            if let Some(test_storage) =
+                <dyn std::any::Any>::downcast_ref::<TestStorageBackend>(player.storage())
+            {
+                let stored_lsos = test_storage.get_stored_data();
+
+                let actual_bytes = match stored_lsos.get(name) {
```
moulins: Same as above; this logic should be in a dedicated submodule.

tests/README.md
```diff
@@ -305,6 +305,17 @@ optimize = true
...
 debug = false
 
+# SharedObject assertions to perform, with the exact SharedObject path (e.g. localhost/test.swf/localName). This block is repeatable.
+[shared_objects."name used in SWF"]
+
+# Extension-less path to the expected sol file.
+# Usually a `.sol` file matching what Flash Player would generate.
+# The test runner will compare against expected + ".sol".
+expected = "foo"
+
```
moulins, line 315: Having to specify the full name of the shared object + the filename is a little redundant, what about following image comparisons and having something like this?

```toml
# Will compare against `LOCALNAME.sol` or `LOCALNAME.ruffle.sol`
[shared_objects.LOCALNAME]
# Optional. Default to `localhost/test.swf/LOCALNAME`.
full_name = "custom/my_sharedobject"

Comment (kjarosh) [reply tests/README.md:315]: I think the current approach is better because if two reasons:

1. We care about keying settings per SO name and not filename. We don't care if two SOs have the same file for assertions, it's fine. But we cannot have two different files for addertions for the same SO.
2. There's also the use case for testing deserialization. The current setup only tests serialization s because it adds support for asserts, but in the future (at least that's my plan) we would have support for loading SOs as well. In the current setup it's easy to add `load =` next to `expected =`.

File: tests/framework/src/runner.rs (hunk 9 above)
Comment (danielhjacobs) [reply tests/framework/src/runner.rs]: Can be, though @kjarosh did say:

> Looks like the diffing logic could be extracted into a separate function (potentially useful in other parts of the framework as well).

If it's ultimately used for other parts of the framework, being in a shared_object-specific submodule wouldn't make sense.

Still, probably fine there until then, and even then it will be good to include it in something like framework/src/options/utils.rs

File: tests/README.md (hunk 11 above)
Comment (danielhjacobs) [reply tests/README.md:315]: No strong preference.

File: tests/framework/src/options/shared_object.rs [hunk 12]
```diff
@@ -0,0 +1,121 @@
...
+use std::collections::HashMap;
+use std::io::{Read, Write};
+
+fn diff_binary(actual: &[u8], expected: &[u8], label: &str) -> Option<String> {
+    if actual == expected {
+        return None;
+    }
+
+    let mut message = format!(
+        "{} mismatch. Actual length: {}, Expected length: {}.",
```
kjarosh:
```suggestion
        "{label} mismatch. Actual length: {}, Expected length: {}.",
```

tests/framework/src/options/shared_object.rs
```diff
@@ -0,0 +1,121 @@
...
+    );
+
+    if let Some((idx, (a, b))) = actual
+        .iter()
+        .zip(expected.iter())
+        .enumerate()
+        .find(|(_, (a, b))| a != b)
+    {
+        message.push_str(&format!(
+            "\nFirst mismatch at index {}: Actual 0x{:02x}, Expected 0x{:02x}",
```
kjarosh:
```suggestion
            "\nFirst mismatch at index {idx}: Actual 0x{a:02x}, Expected 0x{b:02x}",
```

tests/framework/src/options/shared_object.rs
```diff
@@ -0,0 +1,121 @@
+use crate::backends::TestStorageBackend;
+use crate::options::SharedObjectConfig;
+use anyhow::Result;
+use std::collections::HashMap;
+use std::io::{Read, Write};
+
+fn diff_binary(actual: &[u8], expected: &[u8], label: &str) -> Option<String> {
```
kjarosh:
```suggestion
fn diff_binary(actual: &[u8], expected: &[u8], label: &str) -> anyhow::Result<()> {
```

tests/framework/src/options/shared_object.rs
```diff
@@ -0,0 +1,121 @@
...
+
+pub fn check_shared_objects(
+    player: &std::sync::Mutex<ruffle_core::Player>,
+    shared_objects: &HashMap<String, SharedObjectConfig>,
+    root_path: &vfs::VfsPath,
+) -> Result<()> {
+    for (name, so_config) in shared_objects {
+        let player_lock = player.lock().unwrap();
+
+        if let Some(test_storage) =
```
kjarosh: Do we expect this to fail? If not, it can be an assert. If yes, this if can be inverted.

tests/framework/src/options/shared_object.rs
danielhjacobs: Are your sure?
```
error[E0308]: mismatched types
 --> tests/framework/src/options/shared_object.rs:9:16
  |
7 | ...ry(actual: &[u8], expected: &[u8], label: &str) -> anyhow::Result<()> {
  |                                                       ------------------ expected `Result<(), anyhow::Error>` because of return type
8 | ...l == expected {
9 | ...rn None;
  |       ^^^^ expected `Result<(), Error>`, found `Option<_>`
  |
  = note: expected enum `Result<(), anyhow::Error>`
             found enum `std::option::Option<_>`

error[E0308]: mismatched types
  --> tests/framework/src/options/shared_object.rs:34:5
   |
 7 | fn diff_binary(actual: &[u8], expected: &[u8], label: &str) -> anyhow::Result<()> {
   |                                                                ------------------ expected `Result<(), anyhow::Error>` because of return type
...
34 |     Some(message)
   |     ^^^^^^^^^^^^^ expected `Result<(), Error>`, found `Option<String>`
   |
   = note: expected enum `Result<(), anyhow::Error>`
              found enum `std::option::Option<std::string::String>`

error[E0308]: mismatched types
   --> tests/framework/src/options/shared_object.rs:96:20
    |
 96 |               if let Some(mismatch) = diff_binary(
    |  ____________________^^^^^^^^^^^^^^___-
    | |                    |
    | |                    expected `Result<(), Error>`, found `Option<_>`
 97 | |                 actual_bytes,
 98 | |                 &ruffle_bytes,
 99 | |                 &format!("Known failure SharedObject '{name}'"),
100 | |             ) {
    | |_____________- this expression has type `Result<(), anyhow::Error>`
    |
    = note: expected enum `Result<(), anyhow::Error>`
               found enum `std::option::Option<_>`

error[E0308]: mismatched types
   --> tests/framework/src/options/shared_object.rs:106:23
    |
106 |           } else if let Some(mismatch) = diff_binary(
    |  _______________________^^^^^^^^^^^^^^___-
    | |                       |
    | |                       expected `Result<(), Error>`, found `Option<_>`
107 | |             actual_bytes,
108 | |             &expected_bytes,
109 | |             &format!("SharedObject '{name}'"),
110 | |         ) {
    | |_________- this expression has type `Result<(), anyhow::Error>`
    |
    = note: expected enum `Result<(), anyhow::Error>`
               found enum `std::option::Option<_>`
```

kjarosh: Yes, you need to change `None` to `Ok(())` and `Some(x)` to `Err(x)`. `Option` is about returning something optionally (e.g. `get_error_message() -> Option<String>`), `Result` is about the operation passing or failing (e.g. `check() -> Result<()>`).

tests/README.md
```diff
@@ -305,6 +305,17 @@ optimize = true
...
+# If true, this shared object is known to fail .
```
kjarosh:
```suggestion
# If true, this shared object is known to fail.
```

kjarosh (approved):

tests/README.md
danielhjacobs: Will merge it this way, we can change it in follow-up if there's a strong preference for the other version, but I agree that being explicit doesn't hurt. The test runner purposely prints a list of available shared object keys if someone puts an invalid one, so it should be easy to fix that.

PR #24211, opened 2026-07-14
avm2: Minor performance improvements

Lord-McSweeney:
## Description

Try to return `Value::Integer` instead of `Value::Number` from some operations when possible, add some fast paths, micro-optimize `coerce_to_u32` and `coerce_to_i32`, and prevent some `Activation::op_X` code from being inlined.

From my measurements, this is a tiny (0.5-1%) performance improvement on a few physics SWFs in my collection.

## Testing

No SWF-observable changes

## Checklist

### DIFF
--- a/core/src/avm2/activation.rs
+++ b/core/src/avm2/activation.rs
@@ -2023,6 +2023,8 @@ impl<'a, 'gc> Activation<'a, 'gc> {
                     ((n1 as i64 + n2 as i64) as f64).into()
                 }
             }
+            (Value::Integer(n1), Value::Number(n2)) => (n1 as f64 + n2).into(),
+            (Value::Number(n1), Value::Integer(n2)) => (n1 + n2 as f64).into(),
             (Value::Number(n1), Value::Number(n2)) => (n1 + n2).into(),
             (Value::String(s), value2) => Value::String(AvmString::concat(
                 self.gc(),
--- a/core/src/avm2/activation.rs
+++ b/core/src/avm2/activation.rs
@@ -2146,8 +2146,9 @@ impl<'a, 'gc> Activation<'a, 'gc> {
 
     fn op_decrement(&mut self) -> Result<(), Error<'gc>> {
         let value = self.pop_stack().coerce_to_number(self)?;
+        let result = Value::from(value - 1.0);
 
-        self.push_stack(value - 1.0);
+        self.push_stack(result.try_promote_number());
 
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


core/src/avm2/activation.rs
```diff
@@ -2146,8 +2146,9 @@ impl<'a, 'gc> Activation<'a, 'gc> {
 
     fn op_decrement(&mut self) -> Result<(), Error<'gc>> {
         let value = self.pop_stack().coerce_to_number(self)?;
+        let result = Value::from(value - 1.0);
 
-        self.push_stack(value - 1.0);
+        self.push_stack(result.try_promote_number());
```
kjarosh, line 2151: Isn't fast path for integers faster than a promotion? In order for the promotion to happen, the originating value had to be an integer in the first place, so this only covers cases where we were operating on a number that was an integer. The question is whether this is worth the additional cost. (Similarly increment.)

core/src/avm2/value.rs
```diff
@@ -777,7 +779,6 @@ impl<'gc> Value<'gc> {
         }
 
         match self {
-            Value::Number(n) => Ok(f64_to_wrapping_u32(*n)),
```
kjarosh, line 760: Won't that have negative perf impact? IIRC Ruffle in a lot of places can produce a Number for integers

core/src/avm2/activation.rs
Lord-McSweeney: From my testing there were enough cases across my SWFs where the input to this op (as well as to `Increment`) was a `Number`, so I did it this way. I don't see a significant performance impact either way. I don't have a strong opinion on it, I can change it back if you think it makes more sense that way.

core/src/avm2/value.rs
Lord-McSweeney: After the commit that adds promotion in `Increment`, `Divide`, and `Decrement`, I could not find any SWFs where `coerce_to_i32_slow` and `coerce_to_u32_slow` took up more than 0.2% of time total on perf. My original motivation for removing this path was that it removes an extra branch in the fast (already `Integer`) path, and it shrinks the entirety of `coerce_to_i32`'s codegen to only a few instructions.

core/src/avm2/activation.rs
kjarosh: Yeah, but the question is what % of numbers are integers. If that percentage is high, have you looked into possibly other places where we could produce an integer but we don't? My point is that this op will not produce an integer if it didn't get an integer in the first place.

Lord-McSweeney: I haven't looked into any other promotion optimizations- this was the minimum needed to mostly eliminate the slow path from being hit in `coerce_to_u32`/`coerce_to_i32` in the SWFs I used for testing. I'd prefer to leave other optimizations for future PRs, as the more important part of this PR (to me) was shrinking the fast path in the coercion methods.

kjarosh (approved):

PR #24215, opened 2026-07-14
avm2: Trace uncaught errors

kjarosh:
## Description

When an error is not caught in AVM2, Flash Player traces it. For Ruffle, this is an opportunity to enhance our testing capabilities and:

1. make sure Ruffle doesn't throw an error during test execution (currently it can be silently ignored, and it actually was in few places),
2. make sure Ruffle throws correct errors,
3. make sure we don't have to edit FP's output (some tests were manually edited),
4. if a test fails in the middle due to an exception, we can see what exception caused it and assert on it in output.ruffle.txt.

## Testing

This includes a dedicated test and updates a lot of tests. All tests that seemed real (i.e. taken from FP) were updated based on FP's output. There were few tests that were explicitly marked as not coming from Flash Player, in which case I just updated their output to what Ruffle reported, as there's an overwhelming evidence that Ruffle correctly traces uncaught errors.

There are few tests marked as known failures because Ruffle differs in errors thrown and their formats: not including debug info, some contained memory addresses, missing stack frames, etc. Some of those tests will need to be fixed in the future, the rest are genuine inconsistencies between Flash Player and Ruffle.

## Checklist


core/src/avm2.rs
```diff
@@ -682,6 +682,14 @@ impl<'gc> Avm2<'gc> {
         error: Error<'gc>,
         extra_info: &str,
     ) {
+        // Flash Player traces uncaught errors, but when trace is disabled
+        // (e.g. release mode), FP doesn't stringify the error for the second
+        // time, which is observable.
```
Lord-McSweeney, line 687: (non-blocking) Can this comment explain a bit more about how it works (e.g. mention the error popups, enabling flashlog on the debug player)?

Lord-McSweeney (approved):

core/src/avm2.rs
kjarosh: I'll do it in a follow-up PR as mentioned on Discord

PR #24220, opened 2026-07-15
ci: Define version number in macOS Info.plist for stable releases

ChrisCPI:
## Description

This allows the version to be visible in the "Get Info" menu for the app in macOS.

## Testing

Not sure if it will work in the workflow, but I did manually run the 2 commands with a production build of Ruffle, and it appeared to work.

image

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,6 +30,7 @@ jobs:
       upload_url: ${{ steps.create_release.outputs.upload_url }}
       package_prefix: ${{ steps.create_release.outputs.package_prefix }}
       tag_name: ${{ steps.commit.outputs.tag_name }}
+      version: ${{ steps.version.outputs.version }}
       version4: ${{ steps.version.outputs.version4 }}
       release_channel: ${{ steps.configure.outputs.release_channel }}
 
@@ -297,6 +298,13 @@ jobs:
           cp -r desktop/packages/macOS package/Ruffle.app
           mkdir package/Ruffle.app/Contents/MacOS
           mv package/ruffle package/Ruffle.app/Contents/MacOS/ruffle
+
+      - name: Set app version in plist
+        # This needs to be stable only because Apple doesn't support semver
+        if: needs.create-release.outputs.release_channel == 'stable'
+        run: |
+          plutil -replace CFBundleShortVersionString -string "${{ needs.create-release.outputs.version }}" package/Ruffle.app/Contents/Info.plist
+          plutil -replace CFBundleVersion -string "${{ needs.create-release.outputs.version }}" package/Ruffle.app/Contents/Info.plist
 
       - name: Compile asset catalog
         run: |


.github/workflows/release.yml
```diff
@@ -298,6 +298,11 @@ jobs:
           mkdir package/Ruffle.app/Contents/MacOS
           mv package/ruffle package/Ruffle.app/Contents/MacOS/ruffle
 
+      - name: Set app version in plist
+        run: |
+          plutil -replace CFBundleShortVersionString -string "${{ needs.create-release.outputs.version4 }}" package/Ruffle.app/Contents/Info.plist
```
kjarosh: > This key is a user-visible string for the version of the bundle. The required format is three period-separated integers, such as 10.14.1. The string can only contain numeric characters (0-9) and periods.

Source: https://developer.apple.com/documentation/bundleresources/information-property-list/cfbundleshortversionstring

`version4` contains 4 parts, not 3

.github/workflows/release.yml
```diff
@@ -298,6 +298,11 @@ jobs:
...
+          plutil -replace CFBundleVersion -string "${{ needs.create-release.outputs.version4 }}" package/Ruffle.app/Contents/Info.plist
```
kjarosh: > This key is a machine-readable string composed of one to three period-separated integers, such as 10.14.1. The string can only contain numeric characters (0-9) and periods.

Source: https://developer.apple.com/documentation/bundleresources/information-property-list/CFBundleVersion

`version4` contains 4 parts, this accepts 1-3 parts

.github/workflows/release.yml
ChrisCPI: What can be used instead?

kjarosh:
The question is: what those fields are used for in MacOS, because if we decide to put our version there after cutting off the prerelease part, we will have multiple binaries that have exactly the same version, despite them being different. They will be indistinguishable to MacOS this way.

.github/workflows/release.yml
kjarosh: See https://github.com/ruffle-rs/ruffle/pull/24220#issuecomment-4984151871, basically we have a 3-part version only for stable releases. Nightlies have a prerelease string that is not possible to put it in a 3-part version.

kjarosh: `version4` is a bridge between semver (that we use) and platforms that do not accept semver, but accept a 4-part version. Apple doesn't seem to support either.

kjarosh: I think we have two options:
1. set these values for stable builds only,
2. lie to MacOS.

Not sure what's the consequence of 2, so I would prefer 1.

ChrisCPI: Maybe `CFBundleShortVersionString` could be the actual version (like `0.3.0`), while the `CFBundleVersion` could be the date (`2026.7.15`)?

kjarosh: But the date is not a version. If we release 0.2.1 tomorrow, it should be older than 0.3.0 but it would have a newer date and we'll lie to MacOS anyway, this time about the ordering.

ChrisCPI: Welp, I guess we can have this only be for stable releases for now. How can I get just the version number like 0.3.0 in the workflow?

kjarosh: You need to add an if for channel and use `version` instead of `version4`. It's not exposed though in the job so you'll need to do it before using.

ChrisCPI: Do I do that by putting it under `outputs` in the `create-release` job?
```
version: ${{ steps.version.outputs.version }}
```

kjarosh: That's right

kjarosh (approved):

.github/workflows/release.yml
```diff
@@ -298,6 +299,12 @@ jobs:
           mkdir package/Ruffle.app/Contents/MacOS
           mv package/ruffle package/Ruffle.app/Contents/MacOS/ruffle
 
+      - name: Set app version in plist
+        if: needs.create-release.outputs.release_channel == 'stable'
```
kjarosh, line 304: Can you add a comment above that we make sure it's stable because Apple doesn't support semver?

kjarosh (approved):

PR #24222, opened 2026-07-15
avm2: Update comment about how FP handles uncaught errors

kjarosh:
## Description

This adds more context to how uncaught exceptions work in FP.

## Testing

N/A

## Checklist

### DIFF
--- a/core/src/avm2.rs
+++ b/core/src/avm2.rs
@@ -682,9 +682,20 @@ impl<'gc> Avm2<'gc> {
         error: Error<'gc>,
         extra_info: &str,
     ) {
-        // Flash Player traces uncaught errors, but when trace is disabled
-        // (e.g. release mode), FP doesn't stringify the error for the second
-        // time, which is observable.
+        // Flash Player traces uncaught errors (when tracing is enabled).
+        //
+        // When tracing is enabled, Flash Player will *always* log and stringify
+        // the exception twice.
+        //
+        // When tracing is disabled, the error won't get stringified for the
+        // second time (which is observable), so to improve compatibility in
+        // this regard, we disable this feature when emulating the release mode
+        // of FP.
+        //
+        // There is also the additional behavior of debugger exception dialogs,
+        // and if they are enabled, Flash Player will pause the execution and
+        // show the uncaught exception in a popup. If the popups are disabled,
+        // exceptions are just being traced without pausing the execution.
         if activation.context.player_mode == PlayerMode::Debug {
             let stringified = error.to_string(activation);
             activation.context.avm_trace(&stringified);


core/src/avm2.rs
```diff
@@ -682,9 +682,20 @@ impl<'gc> Avm2<'gc> {
...
-        // time, which is observable.
+        // Flash Player traces uncaught errors (when tracing is enabled).
+        //
+        // When tracing is enabled, Flash Player will *always* log and stringify
+        // the exception twice.
+        //
+        // When tracing is disabled, the error won't get stringified for the
+        // second time (which is observable), so to improve compatibility in
+        // this regard, we disable this feature when emulating the release mode
+        // of FP.
```
Lord-McSweeney, line 693: Can you mention that tracing and flashlog are the same thing?

core/src/avm2.rs
```diff
@@ -682,9 +682,20 @@ impl<'gc> Avm2<'gc> {
         error: Error<'gc>,
         extra_info: &str,
     ) {
-        // Flash Player traces uncaught errors, but when trace is disabled
-        // (e.g. release mode), FP doesn't stringify the error for the second
-        // time, which is observable.
+        // Flash Player traces uncaught errors (when tracing is enabled).
+        //
+        // When tracing is enabled, Flash Player will *always* log and stringify
+        // the exception twice.
```
Lord-McSweeney, line 688: Can you move this comment after the comment explaining exception dialogs, to make it clearer that the exception dialog is the source of the first stringification?

core/src/avm2.rs
```diff
@@ -682,9 +682,20 @@ impl<'gc> Avm2<'gc> {
...
+        //
+        // When tracing is disabled, the error won't get stringified for the
+        // second time (which is observable), so to improve compatibility in
+        // this regard, we disable this feature when emulating the release mode
+        // of FP.
+        //
+        // There is also the additional behavior of debugger exception dialogs,
```
Lord-McSweeney, line 695: Could you make it clear that debugger exception dialogs also show up in non-debug FP? As it's written, this comment seems to suggest that they're specific to the debug player.

core/src/avm2.rs
```diff
@@ -682,9 +682,20 @@ impl<'gc> Avm2<'gc> {
...
+        // and if they are enabled, Flash Player will pause the execution and
+        // show the uncaught exception in a popup. If the popups are disabled,
```
Lord-McSweeney, line 697: Could you
1. explain how to disable popups, using "dismiss all", and
2. make it clear that this tracing only happens when flashlog is also enabled?

PR #24225, opened 2026-07-16
tests: Don't test platform/arch-specific NaN serialization in AVM1/AMF0

danielhjacobs:
## Description

Serialization of NaN is platform/arch specific in Flash Player. We don't emulate that. It was suggested on Discord that we just may not want to serialize NaN in tests output

## Testing

See all the platform difference of the old test. This one is consistent as long as your timezone is UTC.

## Checklist


kjarosh (approved):
NaNs is something we can keep in tests, but don't assert on a specific serialized value, just serialize, deserialize, check. It can be a separate test.

Issue #21569, opened 2025-09-04
hitArea property is completely unimplemented for sprites / movie clips


n0samu:
### Describe the bug

This is already indicated in #263, but I think it warrants a separate issue to draw attention to it. Setting a hit area for a movie clip does not work at all in AS2 or AS3! I opened an issue about this before in #12228, but back then I somehow didn't notice that the problem is not confined to button mode movie clips in AS3. As far as I can tell, hitArea is completely unimplemented and does not work at all!

### Expected behavior

Setting a hitArea for a movie clip should affect what areas of the clip are clickable / receive mouse events.

### Content Location

See the from_shumway/avm1/hitarea test for a simple AS2 example. (Currently the test is accidentally passing because our input.json is wrong - I'll update this when that is corrected.)
Here is a version of the test SWF modified for interactive testing:
[test.zip](https://github.com/user-attachments/files/22162371/test.zip)
Notice that Flash Player shows a hand cursor and traces `rollover` when you hover over the square on the left (the clip's defined hit area), but Ruffle shows a hand cursor and traces `rollover` when you hover over the square on the right (the original clip).

### Affected platform

Desktop app

### Operating system

Windows 10

PR #24227, opened 2026-07-16
🤖 avm1: Implement MovieClip.hitArea (part of #21569)

FIM43-Redeye:
## Description

Implements the AVM1 half of #21569: `MovieClip.hitArea` was completely
unimplemented, so assignments had no effect on mouse picking, and the
long-standing `from_shumway/avm1/hitarea` test was `known_failure`
because of it.

Flash Player resolves `hitArea` as an ordinary script property read at
pick time: whenever a button-mode clip is picked, the property is
re-fetched (running any user getter), and a value that currently
resolves to a display object has its shape tested in place of the
clip's own shape, at its own stage position. The engine stores no
hit-area state of its own — an assignment just writes the ordinary
script property, and a user setter can consume it entirely. Because a
getter runs during picking, it can call `stopDrag()` mid-drag (handled
with a reentrancy guard in `update_drag`) or remove clips mid-pick (in
self-removal captures, removal takes effect on the next pick). The
commit
messages and fixture headers describe the captured cases in full, all
established against the Flash Player 32 Linux debug projector.

(An earlier revision instead mirrored assignments out of the property
machinery into the display object; review feedback and further
captures showed that model is wrong. The branch was rewritten in
place, so the old approach remains visible in the outdated review
diffs.)

On performance: the old owner-bounds early-out is gone. Hit areas may
lie outside the owner's bounds, getters are side-effectful, and
captures show FP itself re-resolves the property on every pick, so the
result cannot be cached or bounds-gated.

No SWF5 fixture is included because these script-only fixtures rely on
the drawing API introduced in SWF6. The core diff is +78/-5:
`movie_clip.rs` (the helper and pick-time logic), `player.rs` (the
reentrancy guard), and one interned string; the rest is test fixtures.

## Testing

Four SWF regression tests under `tests/tests/swfs/avm1/hitarea_*`,
down from sixteen in the previous revision per review feedback:
`hitarea_sweep` (eleven former single-shot fixtures consolidated into
one SWF and one choreography, plus transform-in-getter coverage added
during review), `hitarea_lazy_getter`,
`hitarea_remove_sibling`, and `hitarea_remove_owner_drag`. Every
fixture's expected output is real FP32 debug-projector output,
captured by scripting the fixture's `input.json` choreography with
xdotool in a virtual X server. Each fixture's header documents its
mouse path in stage coordinates (so `input.json` can be reproduced by
hand) and its reason for being a separate fixture.

`from_shumway/avm1/hitarea` now passes and its `known_failure` flag is
removed. Against master, all four fixtures fail.

- `cargo test --all` passes (regression suite: 4162 passed, 0 failed,
  332 ignored, at this branch rebased onto current master)
- `cargo fmt --all -- --check` clean; `cargo clippy --all --tests`
  zero warnings

## Checklist


core/src/avm1/object/script_object.rs
```diff
@@ -315,20 +315,31 @@ impl<'gc> Object<'gc> {
...
+                true
             }
         };
 
+        // The assignment was processed unless a read-only property rejected
+        // it; a setter consuming it still counts (see `update_hit_area`).
+        if (stored || setter.is_some())
+            && let Some(dobj) = self.native_no_super().as_display_object()
+        {
+            stage_object::update_hit_area(dobj, name, value, activation);
```
kjarosh: Why are you calling it here?

FIM43-Redeye: This is the earliest point where we know whether the assignment was actually processed, which is what Flash keys the redirect update on. Calling it at `notify_property_change` doesn't work because it runs before the setter executes and before the store attempt. A read-only property rejecting the write shouldn't update the redirect (`hitarea_readonly` pins against Flash Player, the redirect stays on the previously assigned clip), while an own `addProperty` setter consuming the value still should (`hitarea_setter`). The setter's looked up just above, but `stored` only resolves at the stsore attempt, so this is the first point where the entire condition is knowable.

This is basically the same reason it can't be a magic display property, too. Reads echo back whatever was assigned, even a non-object, so it has to stay a plain stored property with a side effect. I can expand the comment at the call site if it helps, what would work best?

kjarosh: What does "Flash keys the redirect update on" mean? What is the redirect? What do you mean by keying?

What do you mean by "A read-only property rejecting the write shouldn't update the redirect"? That's kinda the definition of a read-only property?

core/src/display_object/movie_clip.rs
kjarosh, line 1: Since you mentioned that you've used LLM to generate some code, and I can see that you're using it extensively, please describe how you used it exactly and how you made sure that the LLM didn't produce low quality code or code that is too specific to the API you're implementing. What is your contribution here except generating the code by an LLM?

FIM43-Redeye: I personally reviewed the behavior of the fix on each of the tests (mouse automation is broken under Wayland so I hand-captured those). I also verified the code locally, checked the tests for vacuousness against master (12/13 fail), and reviewed the final diff before submission. Most importantly in my opinion, I categorized the behavior by hand with Flash Player captures to begin with.

On being too specific to the API, the property hooks follow the `notify_property_change` pattern for text-field binding mirroring in the same file, the picking change reuses the storage field AVM2 already added, and no general machinery changed behavior.

I used Claude to write the Rust and test scripts, as well as to help me understand the architecture. If there is a more specific bar for what level of LLM-assisted contributions are accepted in this repository, please let me know.

core/src/avm1/object/script_object.rs
FIM43-Redeye: Sorry, unclear shorthand on my part. By "redirect" I meant the `MovieClip::hit_area` field that `mouse_pick_avm1` consults. When it's set, picking tests the assigned clip's shape instead of the owner's own, and `update_hit_area` is what writes that field.

The point I mangled about read-only was that "the value was stored" and "FP applies the hitArea side effect" are different conditions. A setter consuming the assignment stores nothing, but FP still updates the hit area (`hitarea_setter`, `hitarea_proto_setter`); a read-only property also stores nothing, and FP doesn't (`hitarea_readonly`). Hence the `stored || setter.is_some()` gate, sitting right after the store attempt, the first point where both are known. Earlier (e.g. at `notify_property_change`) it would fire on writes FP rejects.

Expanded the call-site comment to say this clearly.

kjarosh: That seems like you're overfitting the code too much, do you have evidence that hitArea is special in how it handles overriding the setter or setting flags? All other display object properties behave differently?

core/src/display_object/movie_clip.rs
kjarosh: > no general machinery changed behavior

Yes, it did. You're changing the generic code of properties for injecting behavior for a specific property, which is without precedent.

I'm worried that the code generated by Claude is just too specific for the use case, it looks like you potentially encountered some bugs, but the code was generated with this specific API in mind instead of focusing on how AVM1 works in general.

It _is_ possible there are just bugs in ruffle that cause the difference in behavior, but the proper way is to address them separately instead of hacking around a specific property. (Assuming that was the case.)

FIM43-Redeye: Thanks for correcting me on the property machinery hook- still learning the architecture here. I did some more captures on FP and found a few things worth citing:
* A getter's call counter climbs continuously during picking, even with the mouse stationary.
* An assignment consumed by a setter (nothing stored) never redirects, there's no write-time snapshot to hook.
* A getter returning a clip that was NEVER assigned DOES redirect.
So as far as I understand, hitArea is a normal lazily-read property. In view of this I changed the code to resolve at pick time with a helper shaped like get_avm1_boolean_property, so all the avm1/object*.rs files are unchanged now. Test suite is also reworked, cut some useless fixtures and added one that any write-time implementation should fail.

Found a bunch of interesting things along the way too. _droptarget ignores hitArea entirely, so I gated that. Using a TextField as hitArea kills the owner (no redirect or own-shape fallback) so I handled that too. A getter removing clips mid-pick breaks FP itself, so instead of emulating a broken state, removed clips are never pickable. There was also a crash by recursion in my implementation if a getter called stopDrag() mid-drag, leading to it reentering update_drag, which is why I put in that guard. Without the guard it recurses until the stack overflows. I patched update_drag because that function shouldn't reenter itself afaik.

Thanks for your time reviewing this. I lean on LLM assistance heavily, so I verified all the FP behavior above against real FP captures. Happy to change anything here.

core/src/avm1/object/script_object.rs
FIM43-Redeye: Addressed in the other thread.

core/src/display_object/movie_clip.rs
kjarosh: Thanks, that looks a lot better now, and I like your approach with extensive tests!

> I lean on LLM assistance heavily, so I verified all the FP behavior above against real FP captures.

That's good to know, although you have to remember that (at least in my experience) LLMs will do everything to make tests pass without thinking much whether the code makes sense. That's why it's important (not only in the case of LLMs, but in general) to have tests more complex than the solution, because otherwise instead of reverse engineering properly and finding patterns, you'd just hard code specific cases you've tested. It's also important to try to falsify any claims you had while reverse engineering.

tests/tests/swfs/avm1/hitarea_remove_owner_drag/test.toml
kjarosh, line 1: In general you don't have to test property case sensitivity for each property. It's a well known pattern that applies to all properties and it's already well tested. You should add tests for case sensitivity if there's a behavior you found strange or when something doesn't work as expected.

tests/tests/swfs/avm1/hitarea_chain/test.as
kjarosh, line 1: I see that you've added a lot of tests (and that's great!), but a large number of small tests has a big performance overhead, because for each tests we have to load playerglobal separately. It would be great to consolidate all those cases you've tested into one test, ideally with an instruction on how to produce mouse inputs that are present in input.json.

You can do that easily by just placing different MCs on the stage in such a way that the user can just move the mouse across and it would produce required input. You can even draw a line on where to move the mouse and click, that would make it super simple to reproduce.

Similarly, tests with input.json are more difficult to maintain so if you have any parts that could be tested without input.json, you could put them into a separate test without it. (But I didn't see anything obvious from glancing over the tests.)

tests/tests/swfs/avm1/hitarea_invisible/test.as
```diff
@@ -0,0 +1,47 @@
...
+btn.endFill();
+
+createEmptyMovieClip("hit", 2);
+hit.beginFill(0x00FF00);
+hit.moveTo(50, 50);
+hit.lineTo(150, 50);
+hit.lineTo(150, 150);
+hit.lineTo(50, 150);
+hit.endFill();
+hit._visible = false;
```
kjarosh: You can also test what if `hit` is visible, but `btn` isn't.

core/src/display_object/movie_clip.rs
```diff
@@ -2981,9 +3004,16 @@ impl<'gc> TInteractiveObject<'gc> for MovieClip<'gc> {
             return None;
         }
 
+        // A `hitArea` getter that ran earlier in this pick may have removed
+        // this clip (the parent iterates a snapshot of its render list).
+        // Flash Player's mouse pipeline visibly breaks in that situation, so
```
kjarosh: What is `mouse pipeline`? How does it break?

core/src/display_object/movie_clip.rs
kjarosh: BTW You're welcome to join our Discord server and chat with us, this way you could run your ideas / questions though us quickly or ask for help if you're stuck somewhere.

core/src/display_object/movie_clip.rs
```diff
@@ -2274,6 +2274,29 @@ impl<'gc> MovieClip<'gc> {
...
+    /// Resolve the AVM1 `hitArea` property for mouse picking, if set.
+    ///
+    /// Unlike AVM2's `Sprite.hitArea`, AVM1 keeps no dedicated storage for the
+    /// hit area: Flash Player reads the ordinary script property (running any
+    /// getter) each time it picks. This mirrors how sibling engine-consulted
+    /// properties like `enabled` are read (see `get_avm1_boolean_property`).
+    ///
+    /// Returns the display object the property currently resolves to; values
+    /// that are not display objects resolve to `None`. The caller decides how
+    /// each case picks (only a `MovieClip` actually redirects).
```
kjarosh: Can we not use the word "redirect"? It sounds strange and non-technical, and despite the fact you've already explained it to me, I still don't understand what it means here.

core/src/display_object/movie_clip.rs
```diff
@@ -2274,6 +2274,29 @@ impl<'gc> MovieClip<'gc> {
...
+    fn avm1_hit_area(self, context: &mut UpdateContext<'gc>) -> Option<DisplayObject<'gc>> {
+        let object = self.object1()?;
+        let mut activation = Avm1Activation::from_nothing(
+            context,
+            ActivationIdentifier::root("[AVM1 hitArea]"),
+            self.avm1_root(),
+        );
+        let value = object.get(istr!("hitArea"), &mut activation).ok()?;
```
kjarosh, line 2274: I'm not sure if you've already tested it (I don't think so), in order to make sure what to do with an error here, just throw from inside the property getter, e.g.

```as
addProperty("hitArea", function() { throw "test"; });
```

kjarosh: (it's non-blocking though, as it's a pretty niche path, that's something that can be covered in the future)

core/src/display_object/movie_clip.rs
```diff
@@ -2997,20 +3027,62 @@ impl<'gc> TInteractiveObject<'gc> for MovieClip<'gc> {
...
+                    // the matrix check below). Flash Player resolves the
+                    // property (running any getter) on every pick, but the
+                    // result only affects button-mode picking: `_droptarget`
+                    // resolution (`require_button_mode == false`) resolves and
+                    // then ignores it.
+                    let hit_area = self.avm1_hit_area(context).filter(|_| require_button_mode);
+                    // The getter may have removed this clip itself; likewise,
+                    // never pick a removed clip.
+                    if self.avm1_removed() {
+                        return None;
```
kjarosh: Is that something covered by tests?

core/src/player.rs
```diff
@@ -1463,6 +1472,12 @@ impl Player {
 
     /// Update dragged object, if any.
     pub fn update_drag(context: &mut UpdateContext<'_>) {
+        if context.mouse_data.updating_drop_target {
+            // An AVM1 `hitArea` getter running inside the mouse pick below can
+            // call `stopDrag()`, re-entering this function while the drag is
+            // still active; returning here terminates that recursion.
```
kjarosh: It's fine to keep it mentioned in the definition itself, you can just say here:

```rs
// Prevent stack overflow.
```

core/src/display_object/movie_clip.rs
```diff
@@ -2997,20 +3027,62 @@ impl<'gc> TInteractiveObject<'gc> for MovieClip<'gc> {
...
+                            // A hit area still works when its clip is invisible
+                            // (the classic invisible-hit-rect idiom), but a hit
+                            // area that is itself in use as a mask never
+                            // registers hits. `hit_test_shape` bounds-checks the
+                            // hit area at its own stage position internally.
+                            area.hit_test_shape(context, point, HitTestOptions::SKIP_MASK)
+                        }
+                        // No hit area, a non-display-object value, or a hit
+                        // area removed from the stage: the clip's own shape
+                        // picks.
```
kjarosh: I would just mention that in other cases the clip's own shape picks. Listing the exact conditions duplicates cases above and introduces maintenance cost.

kjarosh:
This is looking very good now thank you! Let's polish details and improve tests and we'll be ready to merge.

core/src/display_object/movie_clip.rs
FIM43-Redeye: Went over all my claims to break as much as I could and pulled out four of them:

- A TextField hitArea does NOT disable picking entirely. The original fixture was silent because a selectable textField above the owner stole the mouse by basic occlusion. No TextField special case, objects are just objects.
- The getter MAY NOT BE polled continuously even with the mouse stationary, that was an estimation based on manual captures and failed under script-driven capture with a virtual frame buffer. Dropped the fixture, no claims about idle-polling anywhere. I can dig deeper on this if it's worth roping into this PR.
- FP mouse handling SOMETIMES breaks after a mid-pick removal, but this isn't a universal behavior in my captures. Depends on sibling-removal vs self-removal.
- Removed clips are returned if the clip is removing itself mid-pick, addressed in more detail elsewhere. Sibling-removal captures froze so I've got nothing there.

Thank you so much, I really appreciate the help! I understand that LLMs are very good at sliding out from under you, that's why I try to verify so hard. I'm still learning the ropes of rigorous engineering so this is really useful. Also realized I joined the Ruffle server ages ago to lurk and it disappeared under other servers, so I'll be sure to pop in when I've got ideas/questions!

tests/tests/swfs/avm1/hitarea_remove_owner_drag/test.toml
FIM43-Redeye: Got it, scrapped the case fixtures. I didn't see anything interesting in my probing so it's just SWF-version case sensitivity here.

tests/tests/swfs/avm1/hitarea_invisible/test.as
FIM43-Redeye: Did so, doesn't pick, pinned as a part of the sweep test. Ruffle already had this one down pat.

core/src/display_object/movie_clip.rs
FIM43-Redeye: Went ahead and tested, it's the `btn_throw` case in `hitarea_sweep` now. In FP, throwing doesn't disturb picking, the clip's own shape is tested like if there's no hitArea at all. The only visible effect is an uncaught-exception warning in the debug projector's flashlog, not even a dialog. Ruffle's existing `.ok()` fallback already did this correctly, now it has a test.

tests/tests/swfs/avm1/hitarea_chain/test.as
FIM43-Redeye: Thanks for explaining the test situation, makes total sense. I crunched down from 16 tests (after improvements to the PR) to 4. Little list for readability:
- The three separate ones are only separate because they'd break other tests. `lazy_getter` installs its getter-setter on `MovieClip.prototype` and would eat every `clip.hitArea = x` assignment in the same SWF, therefore most of the sweep can't coexist with it.
- `remove_sibling` and `remove_owner_drag` each have getters that delete a clip in the middle of picking. In our capture sessions, real FP froze up after the sibling-removal variant. Freeze = dead, don't want that in a joint capture, this way FP misbehavior can only ruin one capture and not all 39 stops on the sweep.
- Regarding general consolidation, I wrapped 11 fixtures into `hitarea_sweep`. There's also some little guide lines on the stage so a human can trace the path of the mouse recording, makes input.json reproduction a bit easier. Every fixture header also documents its mouse path in stage coords.
- Areas where the expected response is absolutely nothing are pinned by the trace lines AROUND them. Output comparison is exact.
- Scrapped one test and three trace lines in the sweep. `stationary_poll` tested a claim about FP's idle behavior that ended up unproven and isn't mandatory for the implementation, therefore gone. The three "readback" traces (`trace(clip.hitArea)` after deleting/overwriting/removing the hitArea clip) only tested ordinary property reads. Generic and unnecessary, removed.

Recaptured the sweep's expected output from FP, should be clean but let me know if there's anything I missed.

core/src/display_object/movie_clip.rs
FIM43-Redeye: It is now. Tested, determined the check was actually WRONG. Two captures in two layouts show that a getter removing its own owner still has the owner returned from the removing pick (`_droptarget` names it), and removal only takes effect on the next pick. The check is deleted, and if it ISN'T deleted `hitarea_remove_owner_drag` fails. We lose the `dt: /btnRm` line because the old check suppresses the removing pick's result. The entry check above it is NOT discriminated by any test because removal already removes the clip from its parent's child list, and player.rs filters event delivery. We disabled said entry check and everything passed anyway, but kept it as is nonetheless. This should only matter for clips removed earlier in the SAME pick's traversal snapshot (where FP froze in our attempts) so I don't have any usable capture to pin it against.

The coverage boundary is that `_droptarget` and event delivery are pinned while hover/cursor state and context-menu selection aren't covered. Noting this because both come from the same pick result, but the uncovered elements would need test framework expansion and that is VERY out of scope. I'm trying to keep my work here to adding single coverage elements one by one.

core/src/display_object/movie_clip.rs
FIM43-Redeye: Threw this one in mostly verbatim, `// Otherwise the clip's own shape picks.`

core/src/player.rs
FIM43-Redeye: Applied verbatim.

core/src/display_object/movie_clip.rs
FIM43-Redeye: Agreed, more Claude leaking, sorry. Scrapped the word completely and replaced it with a proper explanation of the property read.

core/src/display_object/movie_clip.rs
FIM43-Redeye: Invented LLM terminology that leaked, I'm still configuring the harness to match Ruffle's conventions more strictly. Replaced with a proper explanation.

core/src/display_object/movie_clip.rs
```diff
@@ -2997,20 +3028,47 @@ impl<'gc> TInteractiveObject<'gc> for MovieClip<'gc> {
...
+                    // picks (`require_button_mode == false`: `_droptarget`
+                    // and context-menu selection) resolve the property but
+                    // ignore its value.
+                    // The getter may remove this clip itself; Flash Player
+                    // still returns it from this pick (removal takes effect
+                    // on the next pick).
+                    let hit_area = self.avm1_hit_area(context).filter(|_| require_button_mode);
+                    let hit = match hit_area {
+                        Some(area) if !area.avm1_removed() => {
+                            // Not `SKIP_INVISIBLE`: an invisible hit area
```
Lord-McSweeney: nit: This comment could be made a little more succinct, like `// NOTE: Not SKIP_INVISIBLE`

core/src/display_object/movie_clip.rs
```diff
@@ -2997,20 +3028,47 @@ impl<'gc> TInteractiveObject<'gc> for MovieClip<'gc> {
...
+                            self.hit_test_shape(context, point, options)
+                        }
+                    };
+                    if hit {
                         return Some(this);
                     }
                 }
             }
 
+            let local_matrix = self.global_to_local_matrix()?;
```
Lord-McSweeney, line 3049: To be clear, this matrix calculation happens after the `hitArea` getter is called? This is probably worth a comment + coverage in one of the existing tests (if there is none).

core/src/display_object/movie_clip.rs
```diff
@@ -2997,20 +3028,47 @@ impl<'gc> TInteractiveObject<'gc> for MovieClip<'gc> {
             // true.
             // InteractiveObject.mouseEnabled:
             // "Any children of this instance on the display list are not affected."
-            if self.mouse_enabled() && self.world_bounds(BoundsMode::Engine).contains(point) {
+
+            if self.mouse_enabled() {
```
Lord-McSweeney, line 3013: Was the `self.world_bounds(BoundsMode::Engine).contains(point)` doing anything before?

kjarosh: It was a fast check if there's a chance that hit test will succeed. Now that you can set a different hit area, this doesn't make sense; or at least the condition would be more complex.

In general this condition should be covered down the stream in hit test (which I assume it is, but worth double checking).

core/src/display_object/movie_clip.rs
FIM43-Redeye: Yup, before the matrix calculation, on purpose. `global_to_local_matrix()?` gives up for a degenerate transform, so clips with scale zero would never consult their hit areas if the order were flipped. The comment on the hit-area block should cover it, and `hitarea_sweep`'s zero-scaled-owner zone also covers it, it would fail minus the ordering. I can move/reword/etc wherever if it'd be better suited at line 3070.

core/src/display_object/movie_clip.rs
FIM43-Redeye: Double-checked, the exact same check is already downstream in `MovieClip::hit_test_shape` at line 2707. So the entire own-shape test is gated by this and the pre-check is redundant. Plus the owner's bounds can't gate the hit-area path, every hit area in `hitarea_sweep`'s layout sits at a different stage position from the owner.

core/src/display_object/movie_clip.rs
Lord-McSweeney: ActionScript can modify the transform in the `hitArea` getter, right? Is there test coverage for that?

FIM43-Redeye: Yeah, the getter's arbitrary AS1/2, so it can modify any transform mid-pick. There wasn't coverage previously, but I added two new zones in `hitarea_sweep` that cover that.

As far as I can tell, Ruffle already does this right. I took some captures with FP32: a transform mutated inside the getter takes effect on the same pick that ran the getter. The getter moving its hit area away yields no rollover on that pick; instead, the moved shape is tested at its new position, and a later visit to the NEW position rolls over. The getter moving the owner itself has the same immediate effect. This is the opposite of mid-pick removal, which lands on the next pick.

I swapped rows 4 and 6 so the owner-removal finale is choreographed last, 'cause FP sometimes just stops delivering mouse input after the mid-pick removal, same kind of freeze as the one documented in `remove_sibling`'s header. The removed-owner readback also fires on a short frame delay instead of a click now, so assertions never depend on post-removal input. Stage grew to 800x720 and ticks grew from 49 to 62, but that's about it.

Lord-McSweeney:
Seems like there's a difference in the `hitarea_remove_sibling` between Ruffle with this PR and Flash Player. In Ruffle, the blue box disappears when the mouse is pressed down (this line in `input.json`):
```
    { "type": "MouseDown", "pos": [0.0, 0.0], "btn": "Left" },
```
However, in FP, the blue box disappears when the mouse is moved back over it, at the end of the test (this line in `input.json`):
```
    { "type": "MouseMove", "pos": [100.0, 100.0] },
```

Do you know the reason for this discrepancy?

FIM43-Redeye:
I had to go hunting a bit but I found the reason, yup.
- The getter's gate was `armed && !removed_done && _root._xmouse < 175`. The MouseDown is at 0,0 and x=0 satisfies <175, so ANY pick that runs after arming at ANY position left of 175 fires the removal.
- Ruffle runs the mouse pick on every mouse event, including button press and release, and runs it AFTER the `onMouseDown` handlers. `handle_clip_event(ClipEvent::MouseDown)` at core/src/player.s:1286 runs first, setting `armed`, then `update_mouse_state` at :1393, whose `run_mouse_pick` at :1554 is unconditional whenever the mouse is in the stage. Ruffle also re-picks once per frame tick, `Player::update` calls `update_mouse_state` at :2397. So in Ruffle, the pick right after the MouseDown handler sees its conditions satisfied and removes btnY there.
- FP only runs that pick once per mouse-MOTION event and never otherwise. Doesn't happen from button presses, button releases (even with the cursor ON the button), or idle frames. Probed with a trace-only hitArea getter plus down/up/move marker with frame stamps. Run 1 (clicks over empty stage) and Run 2 (clicks over the button) both show one `get` per `move` marker and no `get`s anywhere near `down`/`up` or across 2-second idle windows. So FP's first post-arming pick is the final move to 100,100 and removal fires there instead.

The pick cadence here is longstanding Ruffle core behavior that this PR doesn't touch, the whole player.rs diff here is just the stopDrag re-entrancy guard. Before hitArea AVM1 picks couldn't run script, so the timing of picks was unobservable.

Fix is tests-only, one line of ActionScript. I tightened the gate to `_root._xmouse > 50 && _root._xmouse < 175` so it can only pass at 100,100 and not at 0,0. Now the extra picks can't fire the removal early and the fixture actually exercises mid-pick sibling removal in Ruffle (only did so in FP before because Ruffle removed btnY at the MouseDown). Recaptured the edited SWF in FP, trace output is identical to the committed output.txt and the harness tests passes. Might deserve a fix in core logic but I don't feel comfortable taking that on.

kjarosh:
> Fix is tests-only, one line of ActionScript. [...] Might deserve a fix in core logic but I don't feel comfortable taking that on.

It definitely deserves a fix in Ruffle, but you don't have to "fix" the test. Matter of fact, it's not a "fix", it's more of a "ignore" kind of thing :)

If we managed to find a discrepancy that's great, we don't want to ignore it. You don't have to fix it in this PR, but making a known failure test would be very welcome. It can be done in a follow-up PR though.

FIM43-Redeye:
> It can be done in a follow-up PR though.

Gotcha, I'll have one up after this one! For my own peace of mind I wanna have this merged first before I start doing other stuff, but I'll be sure to set up a known-failure test for the next PR.

kjarosh (approved):
Sorry about the delay! From my side this is still LGTM, I know the input loop is in a very bad shape in Ruffle atm and this PR is simpler than I would expect it to be TBH. @Lord-McSweeney any thoughts?

Lord-McSweeney (approved):
LGTM, thank you!

PR #24233, opened 2026-07-16
avm2: Use a shared cold path for no-arg error constructors

kjarosh:
## Description

`make_error_XXXX` functions for argument-less error messages were separately marked #[cold]/#[inline(never)], each duplicating the same lookup-and-construct logic and bloating the binary with near-identical cold functions.

This patch routes them through a single generic make_error_no_args helper (backed by a shared, non-generic make_error_no_args_runtime). This results in the binary having only one cold function (make_error_no_args_runtime) which is called at all call sites. Existing functions remain as #[inline(always)] to make sure the proper error type is used and to match the convention of `make_error_XXXX` functions.

This requires exposing `try_error_message`, as now the cold functions looks up the error message at runtime. An assertion for arg count is additionally added to make sure we don't look up wrong error messages by accident.

This results in the binary shrinking by 0.3% on x86_64.

## Testing

Tested compilation manually, you can't pass a wrong number of arguments. The rest is tested by existing tests.

## Checklist

### DIFF
--- a/core/src/avm2/error.rs
+++ b/core/src/avm2/error.rs
@@ -1,6 +1,7 @@
 use crate::avm2::activation::Activation;
 use crate::avm2::call_stack::CallStack;
 use crate::avm2::class::Class;
+use crate::avm2::error_messages::{assert_error_message_arg_count, try_error_message};
 use crate::avm2::function::display_function;
 use crate::avm2::method::Method;
 use crate::avm2::multiname::Multiname;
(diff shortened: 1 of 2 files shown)


Lord-McSweeney (approved):
I don’t really like how this makes some of the error creation functions look different to others, but the binary size improvement  sounds great to me

kjarosh:
You mean the macro invocations? Underneath the functions have an identical signature. I added the macro invocation because it was a lot of boilerplate code and duplication.

PR #24241, opened 2026-07-17
avm2: Check and throw correct error for Timer delay

evilpie:
## Description

We weren't throw the correct error for invalid Timer delays. For the `delay` setter we were also not checking the right value.

## Checklist


tests/tests/swfs/avm2/timer_invalid_delay/Test.as
```diff
@@ -0,0 +1,27 @@
...
+                new Timer(delay);
+            } catch (e:*) {
+                trace(e);
+            }
+
+            var timer:Timer = new Timer(100);
+            try {
+                timer.delay = delay;
+            } catch (e:*) {
+                trace(e);
```
kjarosh: Can you trace `e.getStackTrace()` instead?

evilpie: Good idea. Flash doesn't have the `checkDelay` helper.

kjarosh (approved):

PR #24242, opened 2026-07-17
chore: Ignore cargo deny DoS errors in our outdated quick-xml

evilpie:
## Description

I would argue that DoS problems aren't really that interesting for us, because there are probably tons of ways for bad Flash code to cause a DoS, so we might as well ignore it for now. 

#21107  exists to update quick-xml

In fact, I think the permanently red CI might hide other more critical advisories in the future.

## Checklist

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -89,7 +89,13 @@ github = [
 
 [advisories]
 version = 2
-
-# `paste` crate is no longer maintained https://rustsec.org/advisories/RUSTSEC-2024-0436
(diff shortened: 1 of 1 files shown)


danielhjacobs:
It was discussed when I added the second comment that once there were three they should probably be above each string in the array.

danielhjacobs:
Also typo "Ingore"

kjarosh (approved):

Issue #23033, opened 2026-02-15
Aurora - Black box in front of some layers


tommytipp:
### Describe the bug

The element is called "ramka" (frame) and for some reason it lies above some other elements in the desktop version of Ruffle.
In the web-version it seems totally fine (see link). 

<img width="330" height="472" alt="Image" src="https://github.com/user-attachments/assets/cac5339b-7067-4221-a591-22b27ca08404" />
<img width="329" height="475" alt="Image" src="https://github.com/user-attachments/assets/77731d83-7346-4c4a-b375-c1e719987297" />
<img width="331" height="481" alt="Image" src="https://github.com/user-attachments/assets/bbd48096-8325-4afe-b07a-e8a5e71d7169" />
<img width="340" height="454" alt="Image" src="https://github.com/user-attachments/assets/3ef00d28-cc43-4af3-a634-637cfa3d4392" />

### Expected behavior

<img width="334" height="478" alt="Image" src="https://github.com/user-attachments/assets/a549c9c2-12a4-41fd-8b0a-0f4d19a49a25" />
<img width="328" height="475" alt="Image" src="https://github.com/user-attachments/assets/311c8852-20af-4d0c-be6e-ff9b18717788" />
<img width="333" height="481" alt="Image" src="https://github.com/user-attachments/assets/fdc31fe4-b766-4bc0-87ac-8dbc23d711ba" />

### Content Location

https://www.kongregate.com/en/games/Pastelgames/aurora

### Affected platform

Desktop app

### Operating system

Linux Mint 22.3

kjarosh:
Hmm, I can reproduce the black box on web too.

PR #24292, opened 2026-07-25
tests: Add failing test for AVM1 using attachMovie in a frame before ExportAssets

evilpie:
## Description

This is a test that Claude came up with demonstrating the behavior difference that causes #23033

### attach_movie_export_not_yet_run

`ExportAssets` linkage names must not be resolvable before the frame carrying
the tag has run.

`test.swf` has two frames:

| frame | contents |
| ----- | -------- |
| 1 | `DefineSprite(id=1)` exists, but is **not** exported yet. Calls `attachMovie("LateExport", "fromFrame1", 1)` and traces the result. |
| 2 | `ExportAssets` registers `id=1` as `"LateExport"`, then calls `attachMovie("LateExport", "fromFrame2", 2)` and traces the result. |

Flash Player 32 traces:

```
undefined
_level0.fromFrame2
```

The frame 1 attach fails because the linkage name does not exist yet. The frame
2 attach is a control: it proves the SWF and the export are well formed, so
frame timing is the only variable between the two lines.

Ruffle currently registers `ExportAssets` names during `MovieClip::preload`,
which walks the whole timeline up front, so the frame 1 attach succeeds and
traces `_level0.fromFrame1` instead. That is recorded in `output.ruffle.txt`
and the test is marked `known_failure = true`.

This is the root cause of #23033: the game's frame 1 script calls
`attachMovie("ramka", ...)` for a symbol only exported on frame 12, so Flash
never creates the clip, while Ruffle creates a black bar that covers the bottom
150px of the stage.

#### Regenerating test.swf

`test.swf` is generated by `gen_test_swf.rs` using the `swf` crate's writer, so
the exact frame layout is explicit rather than baked by an authoring tool. To
regenerate, copy it into the `swf` crate's examples and run it:

```sh
cp tests/tests/swfs/avm1/attach_movie_export_not_yet_run/gen_test_swf.rs \
   swf/examples/gen_test_swf.rs
cargo run -p swf --example gen_test_swf -- \
   tests/tests/swfs/avm1/attach_movie_export_not_yet_run/test.swf
rm swf/examples/gen_test_swf.rs
```

## Checklist


tests/tests/swfs/avm1/attach_movie_export_not_yet_run/gen_test_swf.rs
```diff
@@ -0,0 +1,88 @@
+// Generates this test's `test.swf`.
```
kjarosh: I think the issue with maintaining this file is that it's not being actively compiled in CI and `swf`'s API is not stable, so it's gonna rot.

kjarosh: Is there anything in the SWF that's not craftable by JPEXS? Because if not we can just commit the SWF.

evilpie: I haven't really dug into JPEXS deeply before so I couldn't say. I will remove the Rust file from the commit and just post it here as a comment.

evilpie:
For future reference, this is the Rust code that was used to generate the SWF:

```rust
// Generates this test's `test.swf`.
//
// Builds a minimal SWF demonstrating that ExportAssets
// linkage names must not be resolvable before the frame carrying the tag runs.
//
// Frame 1: attachMovie("LateExport", ...) -- the symbol is NOT exported yet.
// Frame 2: ExportAssets registers "LateExport", then attachMovie again.
//
// Flash traces `undefined` then `_level0.fromFrame2`.
use swf::avm1::types::{Action, Push, Value};
use swf::avm1::write::Writer as ActionWriter;
use swf::*;

/// `attachMovie(export, name, depth); trace(result);`
///
/// Mirrors how Aurora calls it: a bare `CallFunction` whose args are pushed
/// depth-first, followed by the arg count and the function name.
fn attach_and_trace(export: &'static str, name: &'static str, depth: i32) -> Vec<Action<'static>> {
    vec![
        Action::Push(Push {
            values: vec![
                Value::Int(depth),
                Value::Str(SwfStr::from_utf8_str(name)),
                Value::Str(SwfStr::from_utf8_str(export)),
                Value::Int(3),
                Value::Str(SwfStr::from_utf8_str("attachMovie")),
            ],
        }),
        Action::CallFunction,
        Action::Trace,
    ]
}

fn assemble(actions: &[Action]) -> Vec<u8> {
    let mut buf = Vec::new();
    {
        let mut w = ActionWriter::new(&mut buf, 6);
        for a in actions {
            w.write_action(a).unwrap();
        }
        w.write_action(&Action::End).unwrap();
    }
    buf
}

fn main() {
    let out = std::env::args().nth(1).expect("output path");

    let frame1 = assemble(&attach_and_trace("LateExport", "fromFrame1", 1));
    let mut f2 = attach_and_trace("LateExport", "fromFrame2", 2);
    f2.push(Action::Stop);
    let frame2 = assemble(&f2);

    let header = Header {
        compression: Compression::None,
        version: 6,
        stage_size: Rectangle {
            x_min: Twips::ZERO,
            x_max: Twips::from_pixels(200.0),
            y_min: Twips::ZERO,
            y_max: Twips::from_pixels(200.0),
        },
        frame_rate: Fixed8::from_f32(12.0),
        num_frames: 2,
    };

    let tags = vec![
        // --- frame 1: character exists, but is not exported yet ---
        Tag::DefineSprite(Sprite {
            id: 1,
            num_frames: 1,
            tags: vec![Tag::ShowFrame],
        }),
        Tag::DoAction(&frame1),
        Tag::ShowFrame,
        // --- frame 2: NOW export it, then attach again as a control ---
        Tag::ExportAssets(vec![ExportedAsset {
            id: 1,
            name: SwfStr::from_utf8_str("LateExport"),
        }]),
        Tag::DoAction(&frame2),
        Tag::ShowFrame,
    ];

    let file = std::fs::File::create(&out).unwrap();
    swf::write_swf(&header, &tags, std::io::BufWriter::new(file)).unwrap();
    println!("wrote {out}");
}
```

PR #24298, opened 2026-07-25
avm2: Properly stub IME.isSupported

evilpie:
## Description

`IME.isSupported` returned `undefined` and not `false`.

## Checklist

### DIFF
--- a/core/src/avm2/globals/flash/system/IME.as
+++ b/core/src/avm2/globals/flash/system/IME.as
@@ -9,7 +9,6 @@ package flash.system {
     public final class IME extends EventDispatcher {
         private static var _conversionMode:String = "ALPHANUMERIC_HALF";
         private static var _enabled:Boolean;
-        private static var _isSupported:Boolean;
 
         public static function compositionAbandoned():void {
             stub_method("flash.system.IME", "compositionAbandoned");
@@ -27,8 +26,8 @@ package flash.system {
             stub_method("flash.system.IME", "setCompositionString");
         }
 
-        public function get isSupported():Boolean {
-            return _isSupported;
+        public static function get isSupported():Boolean {
+            return false;
         }
 
         public static function get enabled():Boolean {


core/src/avm2/globals/flash/system/IME.as
```diff
@@ -28,7 +27,8 @@ package flash.system {
         }
 
         public function get isSupported():Boolean {
```
Lord-McSweeney: This doesn't fix the issue- the real issue is that the function needs to be `static`.

core/src/avm2/globals/flash/system/IME.as
```diff
@@ -28,7 +27,8 @@ package flash.system {
...
-            return _isSupported;
+            stub_getter("flash.system.IME", "isSupported");
```
Lord-McSweeney: There's no reason to mark this as a stub- if it returns `false`, it's still a correct implementation (as it signals to applications that IME is unsupported by the environment).

evilpie: I guess it's a bit of a philosophical question, but considering we don't support it anywhere, marking it as a stub doesn't seem wrong.

We are a bit inconsistent on this with the other `isSupported` getters as well.

core/src/avm2/globals/flash/system/IME.as
evilpie: Oh, will look into testing...

Lord-McSweeney (approved):

tests/tests/swfs/avm2/ime/output.txt
```diff
@@ -0,0 +1,5 @@
+typeof IME.isSupported: boolean
+typeof IME.enabled: boolean
+typeof IME.conversionMode: string
+IME.isSupported: true
+IME.enabled: true
```
Lord-McSweeney: Wouldn't these change depending on the environment FP is being run in? IMO we can skip adding tests for this change, the `avm2/all_classes` tests should provide enough coverage once they're finished.

evilpie: Probably true, but even Flash on Linux apparently supported it. I will just remove the test for now.

PR #24299, opened 2026-07-25
Fix get_url() not appending local vars to navigate_to_url_normalized() call

Slavko-P:
## Description

While trying to get Poptropica AS2 running I noticed Ruffle fails to append local variables (in this case "island" and "room") to POST request bodies thus resulting in a blank screen when transitioning scenes. This commit fixes the issue and finally puts Poptropica AS2 into a playable state on Ruffle.

## Testing

1. Download Poptropica AS2 from https://flashpointarchive.org/datahub/Legacy_Poptropica_Support
2. Extract the archive
3. Run a proper web server with PHP serving support (nginx, apache or lighttpd; python -m http.server WILL NOT WORK) with ./pop/content/www.poptropica.com/ as the root directory
4. Edit the contents of ./pop/content/www.poptropica.com/getPrefix.php from "prefix=http%3A%2F%2Fwww.poptropica.com" to "prefix=http%3A%2F%2Flocalhost"
5. Open a browser and go to "http://localhost/base.php", set AdScenes to enabled (checkmark) 
6. Go to "http://localhost/base.php?room=FlashpointStart", on the error screen, at the bottom click any gender
8. Now use the mouse cursor to guide your character, and enter the blimp up above
9. Click on any island and the game should be able to switch scenes from either the map or by clicking on various spots inside the islands

Note: I only ran "npm run test" and "npm run wdio -- --firefox" for testing, with everything passing

## Checklist


tests/tests/swfs/avm1/movieclip_geturl/test.swf
ChrisCPI, line 1: Could it be possible to have this test be compiled with Rascal? That way we can have an actual ActionScript source. See the `tests` README for how to do that.

Slavko-P: Ok, added.

tests/tests/swfs/avm1/movieclip_geturl/test.as
```diff
@@ -0,0 +1,4 @@
+this.createEmptyMovieClip("test",this.getNextHighestDepth());
+test.value1 = "string";
+test.value2 = 2;
```
kjarosh, line 3: What if you add properties not on `test`: 1. in local scope, 2. in a test's child? What if the property is virtual (a getter)? What if the property has flags (line DONT_ENUM or VERSION)?

kjarosh: It might be also worth adding a second call on `this` instead of `test` and checking if it works the same way.

Issue #2017, opened 2020-12-18
Min browser version to use Ruffle


ROBERT-MCDOWELL:
Hi folks,
is there a minimum browser version limit to use ruffle?
I tried with firefox 68.x and when I load the demon page it freezes and after a while the browser shows if I want to continue or stop the script
thanks

ousia:
@ROBERT-MCDOWELL, this issue is a duplicate from #2016.

PR #24300, opened 2026-07-25
avm2: Stub flash.system.System.exit

evilpie:
## Description

On the web this method always throws "#2018: System.exit is only available in the standalone Flash Player".
Otherwise it throws "#2017: Only trusted local files may cause the Flash Player to exit.", unless you go through the process of trusting the SWF (e.g. https://stackoverflow.com/questions/17959548/flash-player-add-trusted-location-without-preferences-panel). After doing this on Linux `flashplayerdebugger` actually crashes for me...? 

The method is [documented as](https://docs.ruffle.rs/en_US/FlashPlatform/reference/actionscript/3/flash/system/System.html#exit()):
> For the standalone Flash Player debugger version only.

We currently don't support the `localTrusted` sandbox type at all, so this always throws.

## Testing

Add a test that shows the behavior with a non-trusted SWF (the default). 

## Checklist


core/src/avm2/globals/flash/system/system.rs
```diff
@@ -26,3 +29,25 @@ pub fn set_clipboard<'gc>(
...
+    }
+
+    let movie = activation
+        .caller_movie()
+        .expect("Caller movie expected for exit");
+    if !matches!(movie.sandbox_type(), SandboxType::LocalTrusted) {
+        return Err(make_error_2017(activation));
+    }
+
+    // TODO: Actually quit.
```
kjarosh, line 49: I think `System.exit()` doing nothing may be more dangerous than it throwing. SWFs may expect that code after `System.exit()` will not execute.

I think it would be better to throw anything here and leave the TODO.

kjarosh (approved):

core/src/avm2/globals/flash/system/system.rs
kjarosh: (And yes, I know it's currently unreachable, but will not be at some point.)

core/src/avm2/globals/flash/system/system.rs
```diff
@@ -26,3 +29,25 @@ pub fn set_clipboard<'gc>(
...
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    avm2_stub_method!(activation, "flash.system.System", "exit");
+
+    if cfg!(target_family = "wasm") {
+        return Err(make_error_2018(activation));
+    }
+
+    let movie = activation
+        .caller_movie()
```
Lord-McSweeney: This should use `caller_movie_or_root`

kjarosh (approved):

PR #24306, opened 2026-07-26
avm2: Small text_block.rs apply_format cleanup

evilpie:
## Description

I think this makes the code easier to understand and shorter, but opinions might differ of course.

## Testing

No functional change.

## Checklist

- [x] I, a human, have self-reviewed this PR and fully understand the changes within.
- [ ] I have [made or updated tests](https://github.com/ruffle-rs/ruffle/blob/master/CONTRIBUTING.md#test-guidelines) where possible.
- [x] All of my commits are properly scoped, compile successfully, and pass all tests.
- [x] This PR does not make sense to split up into smaller PRs.
- [ ] An LLM was involved in the authoring of this code.

PR #24307, opened 2026-07-26
avm2: Handle null `elementFormat` correctly when creating text line

Lord-McSweeney:
## Description

Previously, when calling `createTextLine` when `textBlock.content.elementFormat` was `null`, our behavior didn't match Flash's.

As a bonus, due to more validation, some code has been simplified.

This PR conflicts with #24306.

## Testing

I've added coverage to the existing `avm2/textblock_createline_errors` test.

## Checklist


core/src/avm2/globals/flash/text/engine/text_block.rs
```diff
@@ -332,18 +333,29 @@ pub fn do_create_text_line<'gc>(
 
     let width = args.get_f64(1);
 
+    let text_line = create_text_line(activation, width);
```
evilpie: Why would we do this so early?

Lord-McSweeney: Good point, I've moved it a bit lower. Does this look better?

evilpie (approved):

PR #20033, opened 2025-04-06
web: Improve navigator.plugins and navigator.mimeTypes so they can be used with JSON.stringify

evilpie:
To make JSON.stringify work with `navigator.plugins` and `navigator.mimeTypes` we have to make sure all _enumerable_ properties are non-cyclic.

It seems like in a normal browser e.g. `navigator.mimeTypes["application/pdf"]` (all names) and `navigator.plugins["PDF Viewer"]` are not enumerable. So I switched to using `Object.defineProperty` for those.

To hide our internal private variables I switched those to [ECMAScript Private Fields](https://www.typescriptlang.org/docs/handbook/classes.html#ecmascript-private-fields). This also prevents them from being enumerated of course.

`RufflePlugin` didn't use getters so I fixed this as well for consistency with the browser `Plugin` type.

Fixes #19957

PR #24312, opened 2026-07-27
Polyfilled plugin properties remain configurable

### LINKED ISSUE #24311: Web plugin polyfill should make the overriden properties remain configurable
When Ruffle's web plugin polyfill overrides `navigator.plugins`, `navigator.mimeTypes` etc, it removes the property that they are `configurable` by default. This is a problem because it would prevent *another* plugin polyfill script on the same page (for a different plugin/mimetype) from adding its own objects to `navigator.plugins` and the rest. To fix this, the property definitions (`Object.defineProperty` in `web/packages/core/src/plugin-polyfill.ts`) should include `configurable: true`.

(I discovered the problem while attempting to write a plugin polyfill of my own while having the Ruffle extension enabled.)

mincerafter42:
## Description

Fixes #24311. 

This allows additional plugin polyfills or additional code in general (from different scripts or browser extensions) to add to `navigator.plugins`, `navigator.mimeTypes` etc. after Ruffle (or to override the properties in any other way, as they normally can do).

## Testing

Tests added to `web/packages/selfhosted/test/polyfill/spoofing`.

## Checklist

### DIFF
--- a/web/packages/core/src/plugin-polyfill.ts
+++ b/web/packages/core/src/plugin-polyfill.ts
@@ -277,10 +277,12 @@ export function installPlugin(plugin: RufflePlugin): void {
     if (!("install" in navigator.plugins) || !navigator.plugins["install"]) {
         Object.defineProperty(window, "PluginArray", {
             value: RufflePluginArray,
+            configurable: true,
         });
         Object.defineProperty(navigator, "plugins", {
             value: new RufflePluginArray(navigator.plugins),
             writable: false,
+            configurable: true,
         });
     }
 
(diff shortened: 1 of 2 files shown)


evilpie:
Thanks!

Adding a test for this shouldn't be too difficult. See e.g. https://github.com/ruffle-rs/ruffle/pull/20033 how that was done.

evilpie:
Looks great. If you could follow the commit style we can merge this. 
1. Prefix with web:
2. Prefix with tests:

evilpie:
Thank you!

PR #24324, opened 2026-07-28
avm2: Clear parent when deleting element/attribute from XMLList

evilpie:
## Description

This is a speculative fix for #24323. We were not correctly clearing the parent reference from children/attributes while still removing it from the parent's child/attributes list. This is an invalid state and causes crashes, e.g. in `child_index`.

## Testing

Tests that crashed previously for the element case and showed the parent incorrectly being preserved for attributes.

## Checklist


evilpie:
`svgAddGroupIfNeeded` in Scratch has the following loop:

```
         for each(_loc4_ in _loc2_.elements())
         {
            if(_loc4_.localName() != "defs")
            {
               delete _loc2_.children()[_loc4_.childIndex()];
               _loc3_.appendChild(_loc4_);
            }
         }
```

This pretty much confirm the issue. `delete` on the XMLList leaves behind bad element nodes with a parent and the `appendChild` call right afterwards crashes because it ends up calling `child_index` via `maybe_escape_child`.

Lord-McSweeney (approved):
LGTM, I'm surprised none of the avmplus tests hit this

PR #24334, opened 2026-07-31
swf: Add dedicated error types `Avm1ParseError` and `UnexpectedEof`

moulins:
## Description

Currently, parsing AVM1 bytecode returns a `swf::error::Error` on failure, even though it can only fail in a single way (running out of input). This is a fairly chonky type (32 bytes on x64 archs) which `ruffle_core:: avm1::Error` stores inline, so passing it around can be somewhat expensive.

This PR improves the situation by introducing two new error types:
- `UnexpectedEof`, a ZST returned from the `SwfReadExt` methods; as `byteorder`'s API depends on the `io::Error` type, it is replaced with std APIs.
- `Avm1ParseError`, a 1-byte struct returned from `swf::avm1::read::Reader::parse_action`.

This brings significant (~5% and higher) performance improvements on AVM1 benchmarks, and a ~60 kB reduction in binary size on the desktop release build.

I'll probably follow this up with separate PRs to do the equivalent changes for ABC parsing and general tag parsing for consistency's sake, but I don't expect drastic performance improvements here (`ruffle_core::avm2::Error` is already boxed, and ABC/tag parsing only happen at load-time anyways).

## Testing

This has no functional changes (except slightly different error messages).

## Checklist

### DIFF
--- a/core/src/avm1/error.rs
+++ b/core/src/avm1/error.rs
@@ -22,8 +22,8 @@ pub enum Error<'gc> {
     #[error("Property recursion limit has been hit.")]
     PropertyRecursionLimit,
 
-    #[error("Couldn't parse SWF")]
-    InvalidSwf(#[from] swf::error::Error),
+    #[error("Couldn't parse AVM1 bytecode")]
+    InvalidBytecode(#[from] swf::error::Avm1ParseError),
 
     #[error("A script has thrown a custom error.")]
     ThrownValue(Value<'gc>),
--- a/swf/src/error.rs
+++ b/swf/src/error.rs
@@ -3,16 +3,13 @@ use crate::tag_code::TagCode;
 use std::{borrow, error, fmt, io};
 
 /// A `Result` from reading SWF data.
-pub type Result<T> = std::result::Result<T, Error>;
+pub type Result<T, E = Error> = std::result::Result<T, E>;
 
 #[derive(Debug)]
 pub enum Error {
     /// An error occurred while parsing an AVM1 action.
     /// This can contain sub-errors with further information (`Error::source`)
-    Avm1ParseError {
-        opcode: u8,
-        source: Option<Box<dyn error::Error + Send + Sync + 'static>>,
-    },
+    Avm1ParseError(Avm1ParseError),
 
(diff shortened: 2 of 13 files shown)


swf/src/write.rs
```diff
@@ -47,13 +46,14 @@ pub fn write_swf<W: Write>(header: &Header, tags: &[Tag<'_>], output: W) -> Resu
 /// Writes a SWF to the output stream, where the tag list has already been serialized to bytes.
 /// This still appends other header information such as stage size.
 pub fn write_swf_raw_tags<W: Write>(header: &Header, tags: &[u8], mut output: W) -> Result<()> {
-    let signature = match header.compression {
```
Lord-McSweeney, line 50: Was there a reason for this change? The way it's done now seems a bit hacky.

core/src/avm1/error.rs
```diff
@@ -23,7 +23,7 @@ pub enum Error<'gc> {
     PropertyRecursionLimit,
 
     #[error("Couldn't parse SWF")]
```
Lord-McSweeney: This error seems to suggest now that there was an issue with parsing the SWF file itself. Maybe it should be changed to `InvalidBytecode` or something like that to make it clearer?

swf/src/error.rs
```diff
@@ -168,6 +172,23 @@ impl error::Error for AbcParseError {
     }
 }
 
+#[derive(Copy, Clone, Debug)]
+pub struct UnexpectedEof(pub(crate) ());
```
Lord-McSweeney, line 176: Just curious, what does the empty tuple do?

moulins: It makes the struct unconstructible outside the crate.

swf/src/write.rs
moulins: It avoids an extra `write_all` call; if you prefer I can change it back.

core/src/avm1/error.rs
moulins: Hm, good point I'll change it.

Lord-McSweeney (approved):

PR #24335, opened 2026-07-31
video: Add support for NV12 pixel format

X1nto:
## Description

Previously, the NV12 pixel format was unsupported, logging `Unsupported pixel format: Nv12` to the console. This adds support for decoding videos with NV12 pixel format in webcodecs. 

This PR also fixes video freezes. I'm not entirely sure if I should include this fix here, but without it I couldn't even test my changes properly. Looking into the console I found ruffle was spamming these: `decoding video frame XXX failed: No output frame produced by the decoder`. As far as I noticed, when one frame failed to decode in time, instead of skipping over that frame, the decoder waited until it finished decoding and only then continued decoding subsequent frames. This caused the frames to accumulate and stall, causing videos to freeze. 

## Testing

I couldn't find a way to write tests for this, as there are no reference tests for other pixel formats. One of the suggestions was to create an HTML wrapping Flash and let Ruffle polyfill it, but tests require GPU acceleration to be disabled and NV12 doesn't work without GPU acceleration.

Here's an NV12 encoded video, extracted from one of the games I was testing, which inspired me to make this PR: [pitkEqim.flv](https://github.com/user-attachments/files/30603887/pitkEqim.zip).

I'm not familiar with ActionScript, but I can probably whip up a test script that loads the video attached above to help test this. Please let me know.

## Checklist

### DIFF
--- a/video/external/src/decoder/webcodecs.rs
+++ b/video/external/src/decoder/webcodecs.rs
@@ -133,6 +133,8 @@ impl H264Decoder {
                     error!("Unsupported pixel format: {:?}", other_format);
                 }
             };
+
+            output.close();
         };
 
         let log_subscriber_for_error = log_subscriber.clone();
--- a/video/external/src/decoder/webcodecs.rs
+++ b/video/external/src/decoder/webcodecs.rs
@@ -108,6 +108,27 @@ impl H264Decoder {
                         data,
                     )));
                 }
+                VideoPixelFormat::Nv12 => {
+                    let luma_len = visible_rect.width() as usize * visible_rect.height() as usize;
+                    let chroma_len = (visible_rect.width() as usize).div_ceil(2)
+                        * (visible_rect.height() as usize).div_ceil(2);
+                    let mut data: Vec<u8> = vec![0; luma_len + chroma_len * 2];
+                    let _ = output.copy_to_with_u8_slice(&mut data);
+                    let chroma = data.split_off(luma_len);
+                    let chroma_pairs = chroma.as_chunks::<2>().0;
+                    for uv in chroma_pairs {
+                        data.push(uv[0]);
+                    }
+                    for uv in chroma_pairs {
+                        data.push(uv[1]);
+                    }
+                    last_frame.replace(Some(DecodedFrame::new(
+                        visible_rect.width() as u32,
+                        visible_rect.height() as u32,
+                        BitmapFormat::Yuv420p,
+                        data,
+                    )));
+                }
                 other_format => {
                     error!("Unsupported pixel format: {:?}", other_format);
                 }


torokati44:
Thanks!

> Here's an NV12 encoded video ...

As far as I understand it, there is no such thing as "NV12 encoded". The encoding itself only defines that it's in a particular YUV (or Y'CbCr to be precise) colorspace, with 4:2:0 chroma subsampling. It doesn't say anything about the pixel/plane layout of the frame once decoded.
NV12 is about how the  values of the decoded frame are packed (their layout in memory). Whether you get this format or RGB or fully planar YUV (or something else), depends only on whatever happens to be convenient/performant for the particular decoder you end up using, be it any software implementation, or various hardware devices (GPUs).

So, this exact video you linked is decoded fine on my device already (both with Ruffle desktop player and the browser extension), because my decoder behind the WebCodecs API doesn't prefer NV12 layout, like yours does. Similarly, any other video could be problematic on your device that works fine on mine.

> I couldn't find a way to write tests for this, as there are no reference tests for other pixel formats. One of the suggestions was to create an HTML wrapping Flash and let Ruffle polyfill it, but tests require GPU acceleration to be disabled and NV12 doesn't work without GPU acceleration.

The above is also the reason why this is not really testable, because there's not really a way to control (AFAIK) with what frame layout we'll be handed back the decoded frames.

>  As far as I noticed, when one frame failed to decode in time, instead of skipping over that frame, the decoder waited until it finished decoding and only then continued decoding subsequent frames. This caused the frames to accumulate and stall, causing videos to freeze.

With H.264, you can't really just skip random frames, because they depend on each other. Nevertheless, the one-line fix may be correct and necessary, to not leak VideoFrame objects, it just might not actually "skip" anything - but it shouldn't anyway.

torokati44:
With all that said, overall, this still looks alright to me, just let me look at it a bit more to be sure...

torokati44 (approved):
Both commits look correct. Thank you again!

Of course, like everything, this could also be bikeshedded to oblivion (about the lack of comments, naming, and real or imagined micro-optimizations), but it's probably good enough to merge as is!

X1nto:
> Both commits look correct. Thank you again!
> 
> Of course, like everything, this could also be bikeshedded to oblivion (about the lack of comments, naming, and real or imagined micro-optimizations), but it's probably good enough to merge as is!

I didn't really think that comments were necessary, since other pixel formats don't have comments either and adding `output.close()` is semantically correct from the VideoFrame API standpoint. I'm also not too experienced with Rust, so there's a chance I wrote something stupid. I had Claude go over my changes to clean the code up a bit before committing. Either way, glad to hear this looks good!

PR #24338, opened 2026-08-01
avm2: Store content of `ContentElement` subclasses as native data

Lord-McSweeney:
## Description

This PR changes `TextElement` and `GroupElement` to store their data in a new enum, `ElementData`. This will make it easier to access in the future for use in improved FTE support.

Additionally, this PR fixes a couple minor issues with the `ContentElement.text` getter, and `GroupElement` child validation.

Unfortunately, this PR is a bit large, as everything depends on each other.

## Testing

I've added a test, `avm2/groupelement_text`.

## Checklist


core/src/avm2/object/content_element_object.rs
```diff
@@ -98,6 +128,47 @@ impl<'gc> ContentElementObject<'gc> {
...
+            element_data
+        )
+        .borrow_mut()
+    }
+
+    pub fn text(self, activation: &mut Activation<'_, 'gc>) -> Option<AvmString<'gc>> {
+        match &*self.element_data() {
+            ElementData::Text { text } => *text,
+            ElementData::Group { elements } => {
+                let mut result = None;
```
evilpie, line 149: I was curious if using `None` by default is correct and it seems so. Maybe an useful test:

```
       var group = new GroupElement();
       trace("text: " + group.text);

       group = new GroupElement(new Vector.<ContentElement>());
       trace("text: " + group.text);

       group = new GroupElement(new <ContentElement>[new TextElement()]);
       trace("text: " + group.text);

       group = new GroupElement(new <ContentElement>[new TextElement(), new TextElement()]);
       trace("text: " + group.text);

       group = new GroupElement(new <ContentElement>[new TextElement("")]);
       trace("text: " + group.text);

       group = new GroupElement(new <ContentElement>[new TextElement("foobar")]);
       trace("text: " + group.text);
```

Lord-McSweeney: Thanks! I've added some of these cases in the new test.

core/src/avm2/globals/flash/text/engine/group_element.rs
```diff
@@ -0,0 +1,226 @@
...
+    //    with `null` text *anytime after* encountering a `TextElement` that has
+    //    non-`null`, non-empty text, FP crashes.
+
+    // I haven't reproduced this crash here, but it might be necessary to bail
+    // out to avoid having to handle cases in other FTE code that FP cannot
+    // encounter.
+
+    // This crash affects at least both `replaceElements` and `setElements`.
+
+    // Remove elements first
```
evilpie: Can we do something like this:
```
let removed_elements: Vec<_> = elements
    .splice(begin_index..end_index, new_elements.into_iter())
    .map(Value::from)
    .collect();
```

evilpie (approved):
Looks good to me. I imagine we could use `text` directly in `get_text_from_content` now instead of `GET_TEXT`.

core/src/avm2/globals/flash/text/engine/group_element.rs
Lord-McSweeney: Nice catch!

PR #24344, opened 2026-08-01
avm2: Partially implement the rest of `flash.trace.Trace`

ChrisCPI:
## Description

In the non-debugger version of Flash Player, the `setLevel` and `setListener` functions are no-ops, which makes it pretty simple to implement in this way. The actual functionality of `Trace`, which is only present in the FP debugger, is not yet implemented.

## Testing

Test included (the Flash Player output can't be gotten normally, see the test.toml for details)

## Checklist


core/src/avm2/globals/flash/trace/Trace.as
```diff
@@ -8,5 +8,27 @@ package flash.trace {
 
         public static const FILE = 1;
         public static const LISTENER = 2;
+
+        private static var _fileLevel:int = 0;
+        private static var _listenerLevel:int = 0;
+        private static var _listener:Function = null;
+
+        public static function getLevel(target:int = LISTENER):int {
+            if (target <= FILE) {
```
evilpie, line 17: Maybe it would be more obvious to just return `0`? Which is what this does anyway.

ChrisCPI: I did it this way to be closer to how Trace actually functions in the debugger (there are 2 different levels that can be set), but I can simplify it to that if it would be better for now.

evilpie (approved):

core/src/avm2/globals/flash/trace/Trace.as
evilpie: I think it's ok either way.

PR #24351, opened 2026-08-02
 avm2: Create a FormatSpans for the text block's content element in createTextLine

Lord-McSweeney:
## Description

Instead of accessing the text of the content element using a getter, create a full `FormatSpans` for the content element. Currently we discard most of the useful information from it, so this PR shouldn't significantly affect real-world content much. However, this PR does also fix code that uses `GroupElement` to create lines, so it makes some content (e.g. w2w) render correctly now.

## Testing

I'm not adding tests in this PR as it's just an intermediate step to better FTE support. In my next 1-2 PRs, I'll add comprehensive tests for FTE and TLF text rendering.

## Checklist

### DIFF
--- a/core/src/avm2/globals/flash/text/engine/text_block.rs
+++ b/core/src/avm2/globals/flash/text/engine/text_block.rs
@@ -13,7 +13,7 @@ use crate::avm2::value::Value;
 use crate::display_object::{EditText, TDisplayObject, TextLine};
 use crate::fte::{
     FontLookupValue, FontPostureValue, FontWeightValue, TextBaselineValue,
-    TextLineCreationResultValue, TextLineValidity, TextRotationValue,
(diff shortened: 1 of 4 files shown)


evilpie (approved):

core/src/avm2/globals/flash/text/engine/text_block.rs
```diff
@@ -410,12 +408,17 @@ pub fn do_create_text_line<'gc>(
 
     fallback.set_text(subtext, activation.context);
 
+    let format = spans
+        .span(0)
```
evilpie, line 412: This feels quite wrong, but followup, I guess?

Lord-McSweeney: Yeah, this is all going to be changed in the future, I just didn't want the PR to get too large

Issue #1104, opened 2020-09-03
Bullets don't collide with the player in Aggressive Alpine Skiing


Herschel:
The sniper bullets will not collide with the player in Aggressive Alpine Skiing:
https://www.newgrounds.com/portal/view/219667

In Flash, they collide with the player and end the game. Seems like clip events are getting wiped out on the bullet movie clip.

PR #24352, opened 2026-08-02
avm2: Include attribute and element name in XML parser error 1104

evilpie:
## Description

The XML parser error #1104 for duplicate attributes should include the name of the attribute and element name.

## Checklist


Dinnerbone (approved):

core/src/avm2/e4x.rs
```diff
@@ -884,9 +884,6 @@ impl<'gc> E4XNode<'gc> {
                 {
                     return Err(make_error_1088(activation));
                 }
-                Err(XmlError::InvalidAttr(XmlAttrError::Duplicated(_, _))) => {
-                    return Err(make_xml_error(activation, XmlErrorCode::DuplicateAttribute));
```
Lord-McSweeney, line 888: Is this case now unreachable? Why was it removed?

evilpie: This must have always been unreachable.

Lord-McSweeney: Can you explicitly mark it `unreachable!()` and explain why then, so someone doesn't try to handle it in the future?

evilpie: Lots of errors aren't returned by this function, so I don't think we need to single this one out.

Lord-McSweeney (approved):

PR #24354, opened 2026-08-02
avm2: Minor FTE improvements

Lord-McSweeney:
## Description

Clean up some FTE code.

## Testing

I've expanded the existing test `avm2/textblock_line_changes`.

## Checklist


evilpie:
This isn't correct when inserting a new line before the last line:

```as3
var block = new TextBlock(new TextElement("a    \na    \na   \na    \na    \na", new ElementFormat()));

var line1 = block.createTextLine(null, 1000);
var line2 = block.createTextLine(line1, 1000);
var line3 = block.createTextLine(line2, 1000);

trace("block.lastLine == block.firstLine: " + (block.lastLine == block.firstLine));
trace("block.lastLine == line3: " + (block.lastLine == line3));

trace("# insert lineX after line2");

var lineX = block.createTextLine(line2, 1000);
trace("block.lastLine == lineX: " +  (block.lastLine == lineX));
trace("line3.previousLine: " + (line3.previousLine));
```

Flash:
```
block.lastLine == block.firstLine: false
block.lastLine == line3: true
# insert lineX after line2
block.lastLine == lineX: true
line3.previousLine: null
```

Ruffle:
```
block.lastLine == block.firstLine: false
block.lastLine == line3: true
# insert lineX after line2
block.lastLine == lineX: false
line3.previousLine: [object TextLine]
```

Also note the pre-existing issue with us not unlinking the previous line of the line being orphaned.

It would be interesting to know if Flash actually maintains a `lastLine` field or if they simple walk from the `firstLine` through all `nextLine` links. I guess this could be tested by timing the time it takes to get the `lastLine` in a TextBlock with hundreds of lines vs just one line.

Lord-McSweeney:
Oh, nice catch! I'd guess that Flash just follows `nextLine` repeatedly then, since doing so always seems to be able to get you to the real last line in my testing. Unfortunately our logic doesn't match Flash's yet, so implementing it that way would break our tests :(

I'll see if there's any other way we can implement it, or if we should just implement it that way and have failing tests.

Lord-McSweeney:
I've removed the actual implementation of `TextBlock.lastLine` from this PR- I'll implement it later

core/src/avm2/object/element_format_object.rs
```diff
@@ -248,23 +253,19 @@ impl<'gc> ElementFormatObject<'gc> {
...
-                Some(fd.font_weight() == FontWeightValue::Bold),
-                Some(fd.font_posture() == FontPostureValue::Italic),
-                fd.font_lookup() == FontLookupValue::Device,
-            )
-        } else {
-            (None, None, None, true)
-        };
+        let fd = self.font_description();
+
+        let font = fd.font_name().as_wstr().into();
```
evilpie, line 258: I personally think removing the variables and moving the code into the `Some` is nicer, but that is a taste question.

core/src/avm2/object/element_format_object.rs
```diff
@@ -248,23 +253,19 @@ impl<'gc> ElementFormatObject<'gc> {
...
+        let bold = fd.font_weight() == FontWeightValue::Bold;
+        let italic = fd.font_posture() == FontPostureValue::Italic;
+        let _is_device_font = fd.font_lookup() == FontLookupValue::Device;
```
evilpie, line 263: Why keep this?

evilpie (approved):

core/src/avm2/object/element_format_object.rs
Lord-McSweeney: We don't have support for setting `is_device_font` for each individual `FormatSpan` yet, but we need to support it eventually. I'll add a TODO comment here.

core/src/avm2/object/element_format_object.rs
Lord-McSweeney: IMO it's easier to understand this way, the expressions are fairly large

PR #24360, opened 2026-08-03
avm2: Emit `native_table.rs` in a deterministic order

bmwiedemann:
## Description

This change allows for reproducible builds.
See https://reproducible-builds.org/ for why this is good.

The problem was this:
`write_native_table` collected the `[NativeAccessible]` slot constants and the `[NativeCallable]` method constants into a `HashMap` keyed by module name, then iterated it to emit `pub mod slots { .. }` and `pub mod methods { .. }`.
Since HashMap iteration order differs on every run, the generated native_table.rs came out with its 45 slot modules and 6 method modules shuffled, even though the contents were always the same.

The shuffling propagated into the ruffle_desktop binary, so we got a different `ruffle` openSUSE package from every build when nothing in the inputs changed.

The solution:
Use a `BTreeMap` instead. The keys are the module names that are about to be emitted, so ordering them alphabetically both fixes the order and makes the generated file easier to read; nothing downstream depends on the previous order, as the consumers refer to the constants by path.

## Testing

_How can we test this PR?_

One could write a test to recreate the native_table.rs multiple times and check that it has the same content each time. My command of rust is not good enough for this, but I'm not even sure this is worth a test. Do you?

## Checklist

- [X] All of my commits are properly scoped, compile successfully and pass all tests.
- [X] An LLM was involved in the authoring of this code. (I still overhauled the commit message and dropped superfluous code comments)

### DIFF
--- a/core/build_playerglobal/src/avm2.rs
+++ b/core/build_playerglobal/src/avm2.rs
@@ -3,7 +3,7 @@ use proc_macro2::TokenStream;
 use quote::quote;
 use regex::RegexBuilder;
 use std::borrow::Cow;
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 use std::ffi::OsStr;
 use std::fs;
 use std::fs::File;
@@ -462,8 +462,8 @@ fn write_native_table(data: &[u8], out_dir: &Path) -> Result<Vec<u8>, Box<dyn st
     let mut rust_custom_constructors = vec![none_tokens; abc.classes.len()];
     let mut rust_fast_calls = vec![];
 
-    let mut rust_accessible_slots: HashMap<String, Vec<_>> = HashMap::new();
-    let mut rust_accessible_methods: HashMap<String, Vec<_>> = HashMap::new();
+    let mut rust_accessible_slots: BTreeMap<String, Vec<_>> = BTreeMap::new();
+    let mut rust_accessible_methods: BTreeMap<String, Vec<_>> = BTreeMap::new();
 
     let mut check_trait = |trait_: &Trait, parent: Option<Index<Multiname>>, is_class: bool| {
         match trait_.kind {


Lord-McSweeney:
AFAIK HashMap iteration order is stable across environments and builds, as we have tests that depend on it (even though they shouldn't). Our build process should be reproducible, at least for the browser extensions- Mozilla requires it to get the Firefox extension approved. What environment are you building Ruffle in? Have you tested that this fixes the issue?

torokati44:
I thought the web build can already be reproducible...? 🤔

EDIT: Ah, I was too slow. 😅

bmwiedemann:
HashMaps use randomized seeds for indexing. Claude wrote (in a part I cut from the commit msg)
> std's `RandomState` is seeded per process

btw. There is a 2nd issue affecting builds. filesystem-dependent (possibly random with default ext4 dirindex) readdir (walk) order in https://github.com/XAMPPRocky/fluent-templates/blob/fd0720b4a8b8d4bef4ea128a72f9b196ba302a2a/macros/src/lib.rs#L127

bmwiedemann:
When you check https://doc.rust-lang.org/std/collections/struct.HashMap.html and Ctrl-F "order", all the occurences are next to "arbitrary" or "unsorted"

bmwiedemann:
> What environment are you building Ruffle in? Have you tested that this fixes the issue?

I'm building for openSUSE:Factory (aka openSUSE Tumbleweed) with rust1.97 and I only started to investigate because my reproducibleopensuse scripts reported diffs with ruffle-0.4.1 and I only submitted this PR because I found that the patch helped.

https://build.opensuse.org/package/show/openSUSE:Factory/ruffle was only newly added to the distribution 10d ago.

And we package /usr/bin/ruffle which is probably not the browser-extension.

Lord-McSweeney:
Oh, interesting, it does change between runs. But as this is part of the build script, and the order of the slots doesn't affect the final build, how does this affect reproducibility? Do you also compare generated artifacts from the build?

bmwiedemann:
It does affect the final `ruffle` binary for us.

Another part I cut from Claude (maybe I should just have left it in)
> The shuffling propagates: it changes the item order of ruffle_core, hence the crate's SVH, hence the symbol hashes of everything that links against it, so an otherwise byte-identical rebuild of ruffle produces a different ruffle_desktop binary. This was found while trying to reproduce the openSUSE ruffle package bit-for-bit.

Additionally rust enables local-LTO (link-time-optiimization) by default, which hashes all inputs and that makes many symbol names vary if a single bit in the inputs varies.

kjarosh (approved):
LGTM, makes sense to generate code in a deterministic way

bmwiedemann:
If you are interested in upstreaming the 2nd (external) issue affecting ruffle, https://github.com/XAMPPRocky/fluent-templates/commit/1ee4131ce54450aee200ca5f3266d4e2c70f2829 is straight from Claude. LGTM with my limited rust, apart from the excessive verbosity.
Note: That issue is harder to reproduce. With `disorderfs` 's order vs order+reverse mode, it should be possible.

kjarosh:
@bmwiedemann TBH it would be best if you upstreamed it, since you have the context and everything, but if for some reason you don't want to, could you file an issue about it in our repo instead? This way someone will take case of it when we can. Please include reproduction steps so that we know when the build is reproducible under your definition.

kjarosh:
Thanks for the fix!

PR #24371, opened 2026-08-04
Fte/textline composition

doombubbles:
## Description

TextBlock.createTextLine was using information from the entire text block when setting rawTextLength, and it updated TextBlock.firstLine every time it created a line.

As a result, every line reported the full content length as its rawTextLength, while firstLine ended up pointing to the most recently created line. Line separation also recognized only U+000A, even though the Text Layout Framework uses U+2028 and U+2029 as line and paragraph separators.

This updates line creation so that:
- rawTextLength reflects the text covered by that line
- firstLine is set only for the first created line
- U+000A, U+2028, and U+2029 are all recognized as line breaks

## Testing

Added tests/tests/swfs/avm2/textline_raw_text_length, covering each line’s rawTextLength and textBlockBeginIndex, all three separators, and the value of firstLine.

The expected output was captured from the Flash Player 32 debugger. The test fails on master and passes with this change.

## Checklist


evilpie:
Thank you, this looks very good. Can you please rebase?

evilpie:
The firstLine issue was already fixed in https://github.com/ruffle-rs/ruffle/commit/fcaffa3de3e4833f8433255347d1ec0a6adb3fc9, so don't forget to update your commit message.

core/src/avm2/globals/flash/text/engine/text_block.rs
```diff
@@ -452,69 +420,32 @@ fn create_text_line<'gc>(activation: &mut Activation<'_, 'gc>, width: f64) -> Te
...
-    Ok(())
+    Ok(Some(text))
 }
 
 fn next_line_break(text: &WStr, start: usize) -> usize {
     let len = text[start..]
         .iter()
-        .position(|ch| ch == b'\n' as u16)
-        // Include the newline.
+        .position(|ch| matches!(ch, 0x0A | 0x2028 | 0x2029))
```
evilpie: I think we also need to handle 0x0D (`\r`) and also 0x0D 0x0A (`\r\n`). Those seem to be treated as single line breaks compared to e.g. `\n\n`.

```
        trace("Carriage Return");
        dump("ab\rcd");
        trace("");

        trace("Carriage Return + Line Feed");
        dump("ab\r\ncd");
        trace("");

        trace("Double Line Feed");
        dump("ab\n\ncd");
        trace("");
```

```
Carriage Return
  line 0: rawTextLength=3 textBlockBeginIndex=0 first=true
  line 1: rawTextLength=2 textBlockBeginIndex=3 first=false

Carriage Return + Line Feed
  line 0: rawTextLength=4 textBlockBeginIndex=0 first=true
  line 1: rawTextLength=2 textBlockBeginIndex=4 first=false

Double Line Feed
  line 0: rawTextLength=3 textBlockBeginIndex=0 first=true
  line 1: rawTextLength=1 textBlockBeginIndex=3 first=false
  line 2: rawTextLength=2 textBlockBeginIndex=4 first=false
```

evilpie (changes requested):

Lord-McSweeney (changes requested):
Hello, and thank you for the contribution! I'm actively working on improving our FTE implementation, so a lot of this will become unused or moved around in the future.

- `TextBlock.firstLine` is already correctly implemented on current `master`.
- FTE will eventually rely on `html::layout` to do layout calculations; support for separators should be added there, as the `next_line_break` function will soon be entirely removed. For now, I think that change can be dropped in this PR.

However, the test coverage and the change to fix `rawTextLength` look good to me!

doombubbles:
> FTE will eventually rely on html::layout to do layout calculations; support for separators should be added there, as the next_line_break function will soon be entirely removed. For now, I think that change can be dropped in this PR.

The `next_line_break` function change i'm adding fixes text lines' rendering in an app I'm being asked to maintain, so it would mean a lot to me to have it kept until those layout changes land

evilpie:
I think a few additional lines of code that will be replaced don't hurt anyone. Thanks for the tests.

evilpie (approved):

Lord-McSweeney (approved):
LGTM once tests are passing- the `output.ruffle.txt` file needs to be updated

evilpie:
@doombubbles please update the PR

torokati44:
Ideally test outputs should be adjusted in the same commit that causes them to change (not after the fact) to not cause intermediate commits to fail the test. No big deal though!

PR #24393, opened 2026-08-07
desktop: Coalesce cursor moves into one hit-test per tick

JuanMateusMarinho:
## Description

Every `WindowEvent::CursorMoved` synchronously ran `run_mouse_pick`, which
hit-tests the whole display list. winit delivers cursor moves at the mouse
polling rate — hundreds per second on Windows — so moving the cursor across a
busy scene fired a burst of hit-tests *in between* rendered frames. On heavy
content this is very visible: the player hitches while the cursor is moving,
and worse the more display objects are on stage.

This records the latest cursor position instead of acting on it, and flushes it
at most once per tick from `about_to_wait`. It also flushes immediately before
a mouse press, so a click still sees an up-to-date hover target and
`MouseMove` → `MouseDown` ordering is preserved. `CursorLeft` drops the pending
move. Clicks gain no latency.

**The trade-off, stated plainly:** content now sees `MouseMove` at tick rate
rather than at the OS event rate. I think that is the right call — the extra
hit-tests were computing hover state that nothing could observe until the next
frame anyway — but it is a behaviour change for content that tracks the cursor
path finely (a drawing app, say), so it's worth a maintainer's judgement rather
than mine. If you'd prefer this gated behind a flag, or flushed more than once
per tick, I'm happy to rework it.

## Testing

`cargo check -p ruffle_desktop --release` and `cargo fmt --all --check` are
clean, and clippy reports nothing against the changed file. The change is
confined to the desktop event loop, which the SWF regression suite doesn't
exercise, so I verified it by running the player rather than by adding to that
suite.

One aside I ran into while checking this, unrelated to the change:
`cargo clippy -p ruffle_desktop` currently fails on Windows on master, because
the `#[cfg(not(target_os = "linux"))]` arm of `ThemeController::get_system_theme`
is `async` with nothing to await (the Linux arm does await). CI presumably
doesn't see it because it lints on Linux. Untouched here — happy to send the
one-line fix as its own PR if that's useful.

This has had heavy real-world use rather than just a synthetic case. I maintain
a downstream Ruffle build for AdventureQuest Worlds, and this change has been
in every release of it since 2026-06-27 — thirteen releases over about six
weeks of daily play. It was originally made because click-to-move in a crowded
town stuttered badly; the game has no `MOUSE_MOVE` listener at all, so the
entire per-move cost was host-side hit-testing. No input regressions have come
back since, including from mouse-driven UI, drag interactions and text
selection.

To see the difference without that content: open any SWF with a large display
list in the desktop player and sweep the cursor across it while watching frame
pacing. Before, the burst of `run_mouse_pick` calls between frames shows up as
a stutter that tracks cursor motion; after, hover updates once per tick and the
stutter is gone.

## Checklist

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -29,6 +29,12 @@ struct MainWindow {
     player: PlayerController,
     minimized: bool,
     mouse_pos: PhysicalPosition<f64>,
+    /// Whether the cursor moved since the last tick. winit can deliver
+    /// `CursorMoved` at the mouse polling rate (hundreds of Hz), and each move
+    /// runs a full display-list hit-test (`run_mouse_pick`). We coalesce them
+    /// so that at most one hit-test runs per tick, which keeps a burst of OS
+    /// cursor events from stalling frames in a crowded scene.
+    mouse_move_pending: bool,
     modifiers: Modifiers,
     min_window_size: LogicalSize<u32>,
     max_window_size: PhysicalSize<u32>,
@@ -100,10 +106,10 @@ impl MainWindow {
                 }
 
                 self.mouse_pos = position;
-                let (x, y) = self.gui.window_to_movie_position(position);
-                let event = PlayerEvent::MouseMove { x, y };
-                self.player.handle_event(event);
-                self.check_redraw();
+                // Defer the actual hit-test to the next tick via `flush_mouse_move`.
+                // A burst of high-frequency cursor moves collapses into a single
+                // `run_mouse_pick` instead of one per OS event.
+                self.mouse_move_pending = true;
             }
             WindowEvent::DroppedFile(file) => {
                 if let Some(content_descriptor) = ContentDescriptor::new_local(&file, None) {
@@ -124,6 +130,10 @@ impl MainWindow {
                 if self.gui.is_context_menu_visible() {
                     return;
                 }
+
+                // Apply any pending cursor move first so the press/release uses an
+                // up-to-date hover target and preserves MouseMove -> MouseDown order.
+                self.flush_mouse_move();
 
                 use ruffle_core::events::MouseButton as RuffleMouseButton;
                 use winit::event::MouseButton;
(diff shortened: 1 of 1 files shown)


kjarosh (changes requested):
PR description is not meant to contain AI dumped text. If you're using an LLM _you_ are supposed to interface with it, not us. No maintainer will engage with an LLM proxy over a PR.

> it's worth a maintainer's judgement

This does not depend on a maintainer's judgement but on how Flash Player works. If Flash Player does what you're describing then sure, if it doesn't then Ruffle shouldn't as well. This can be tested by running Flash Player.

danielhjacobs:
@JuanMateusMarinho I created the following test with AI assistance:

```as
var frameCount:Number = 0;

this.onEnterFrame = function() {
    frameCount++;
    trace("[TICK " + frameCount + "] t=" + getTimer() + "ms");
};

var mouseListener:Object = new Object();
mouseListener.onMouseMove = function() {
    trace("  ├─ MOUSE_MOVE t=" + getTimer() + "ms pos=(" + _xmouse + "," + _ymouse + ")");
};
Mouse.addListener(mouseListener);
```

Compiling it, opening it in Flash player, and moving the mouse, I have confirmed that content in Flash sees MouseMove at OS event rate, not tick rate.

```
[TICK 3] t=16532ms
  ├─ MOUSE_MOVE t=16534ms pos=(170,33)
  ├─ MOUSE_MOVE t=16543ms pos=(171,33)
  ├─ MOUSE_MOVE t=16550ms pos=(172,33)
  ├─ MOUSE_MOVE t=16559ms pos=(173,34)
[TICK 4] t=16573ms
  ├─ MOUSE_MOVE t=16574ms pos=(174,34)
  ```

If you still think this could be a useful option for certain content, you'd have to put it behind a preference that defaults to not being enabled, as by default Ruffle should be matching Flash's behavior. It would then need to be discussed whether said option is worth adding.

kjarosh:
Closing this PR since the behavior doesn't match Flash. Feel free to start a discussion as suggested by Daniel.

PR #24399, opened 2026-08-08
 avm2: Implement TextBlock.releaseLines and TextBlock.lastLine

Lord-McSweeney:
## Description

Implement `TestBlock.releaseLines` and `TextBlock.lastLine`, add a ton of test coverage.

## Testing

I've expanded the existing tests `avm2/textblock_line_changes` and `avm2/textblock_recreateline`. I've also added the new test `avm2/textblock_releaselines`. Currently this is ignored as it panics within AVM2 code, which causes a double-panic in the test runner that we can't catch. I'll fix the panic and make the test `known_failure` in a follow-up PR.

Unfortunately I encountered so many edge cases while making the PR that the test commit history is a bit wonky. All commits should pass all tests; however, there are some issues with the history (e.g. the test `avm2/textblock_releaselines_errors` is only renamed to `avm2/textblock_releaselines` in the very last commit) that I don't trust myself to fix with a rebase.

## Checklist


core/src/avm2/globals/flash/text/engine/text_block.rs
```diff
@@ -323,6 +323,53 @@ pub fn get_first_line<'gc>(
...
+        .and_then(|o| o.as_text_line())
+        .expect("Guaranteed by AS signature");
+
+    // NOTE: `releaseLines` allows for swapping the order of parameters without
+    // affecting the functionality of the method!
+
+    // Both lines must have associated text blocks
+    let (Some(line_1_block), Some(line_2_block)) = (line_1.text_block(), line_2.text_block())
+    else {
+        return Err(make_error_2004(activation, Error2004Type::ArgumentError));
```
kjarosh, line 369: Looks like something we should cover (false/false, true/false, false/true, true/true).

core/src/avm2/globals/flash/text/engine/text_block.rs
```diff
@@ -323,6 +323,53 @@ pub fn get_first_line<'gc>(
...
+    };
+
+    // Both lines must belong to this block
+    if !Object::ptr_eq(line_1_block, this) || !Object::ptr_eq(line_2_block, this) {
+        return Err(make_error_2004(activation, Error2004Type::ArgumentError));
```
kjarosh, line 374: Looks like something we should cover as well (all 4 cases)

core/src/avm2/globals/flash/text/engine/TextBlock.as
```diff
@@ -141,16 +141,18 @@ package flash.text.engine {
...
-            if (start != end || end != this.firstLine) {
-                stub_method("flash.text.engine.TextBlock", "releaseLines", "with start != end or multiple lines");
-                return;
+            if (this.firstLine === null) {
+                return null;
             }
-            this.firstLine.validity = "invalid";
-            this.firstLine.setTextBlock(null);
+
+            var currentLine:TextLine = this.firstLine;
```
evilpie: With the `next_lines()` iterator, I feel like this would be nicer to write in Rust?

core/src/display_object/text_line.rs
```diff
@@ -194,6 +226,46 @@ impl<'gc> TextLine<'gc> {
...
+        core::iter::successors(self.previous_line(), |line| line.previous_line())
+    }
+
+    /// Find the lines between `self` and `line_2`, inclusive of both. This
+    /// method will work regardless of whether `self` comes before or after
+    /// `line_2` in the text block.
+    ///
+    /// This method assumes that `self` and `line_2` aren't the same line, and
+    /// that both lines are in the same text block.
+    pub fn find_lines_between(self, line_2: TextLine<'gc>) -> Vec<TextLine<'gc>> {
```
evilpie: `between` sounds wrong, maybe `through`?

core/src/avm2/globals/flash/text/engine/TextBlock.as
Lord-McSweeney: Good point. I used `lines()` instead of `next_lines()` though, the latter was a bit less elegant

core/src/avm2/globals/flash/text/engine/text_block.rs
Lord-McSweeney: I've added extra coverage for the three failure cases

core/src/avm2/globals/flash/text/engine/text_block.rs
Lord-McSweeney: I've added extra coverage for the three failure cases

core/src/display_object/text_line.rs
```diff
@@ -194,6 +226,46 @@ impl<'gc> TextLine<'gc> {
...
+        result.push(self);
+
+        for line in self.previous_lines() {
+            result.push(line);
+            if DisplayObject::ptr_eq(line.into(), line_2.into()) {
+                return result;
+            }
+        }
+
+        unreachable!("Both lines should be in the same TextBlock!");
```
kjarosh: I don't like this assertion. First of all it's not unreachable per se, I can call this method with `line_2` not being in the same `TextBlock`. Second of all, this is asserting on a condition made far away somewhere in the codebase, which is a bad practice because it can desynchronize.

core/src/display_object/text_line.rs
```diff
@@ -194,6 +226,46 @@ impl<'gc> TextLine<'gc> {
...
+        core::iter::successors(self.previous_line(), |line| line.previous_line())
+    }
+
+    /// Find the lines between `self` and `line_2`, inclusive of both. This
+    /// method will work regardless of whether `self` comes before or after
+    /// `line_2` in the text block.
+    ///
+    /// This method assumes that `self` and `line_2` aren't the same line, and
+    /// that both lines are in the same text block.
+    pub fn find_lines_through(self, line_2: TextLine<'gc>) -> Vec<TextLine<'gc>> {
```
kjarosh:
```suggestion
    pub fn find_lines_through(self, other: TextLine<'gc>) -> Vec<TextLine<'gc>> {
```

core/src/display_object/text_line.rs
```diff
@@ -194,6 +226,46 @@ impl<'gc> TextLine<'gc> {
...
+
+    pub fn previous_lines(self) -> impl Iterator<Item = TextLine<'gc>> {
+        core::iter::successors(self.previous_line(), |line| line.previous_line())
+    }
+
+    /// Find the lines between `self` and `line_2`, inclusive of both. This
+    /// method will work regardless of whether `self` comes before or after
+    /// `line_2` in the text block.
+    ///
+    /// This method assumes that `self` and `line_2` aren't the same line, and
```
kjarosh: Isn't it easier to assume they could be the same line? It's one simple condition and makes this method safer.

core/src/avm2/globals/flash/text/engine/text_line.rs
```diff
@@ -89,27 +89,14 @@ pub fn get_text_block<'gc>(
...
-        .as_text_line()
-        .unwrap();
-
-    let text_block = args
-        .try_get_object(0)
-        .and_then(|o| o.as_text_block_object());
-    this.set_text_block(text_block, activation.gc());
-    Ok(Value::Undefined)
+    if matches!(this.validity(), TextLineValidity::Static) {
+        // "static" lines always return `null` when accessing their `textBlock`
```
kjarosh: Can you elaborate in the comment on why we can't just set null and we have to pretend?

kjarosh (approved):
LGTM after addressing comments

core/src/display_object/text_line.rs
Lord-McSweeney: I've changed it to a `panic!` for now. I'll fix this code in my next PR.

PR #24402, opened 2026-08-09
fix: restore perspective Flash 3D camera rendering

Dbono-dev:
## What changed

- Implement `flash.geom.Matrix3D.pointAt` using the documented object-relative `at` and `up` axes while preserving translation and scale.
- Keep UVT/perspective triangles from being discarded by the flat 2D winding test.
- Skip projected triangles at or behind the projection plane instead of sending invalid perspective data to the shader.
- Add an AVM2 regression test covering the Away3D `+Z` forward / `-Y` up camera convention, defaults, translation, and scale preservation.

## Root cause

Number 1 uses Away3D and calls `Object3D.lookAt()` for every Shootout attack type. Ruffle's `Matrix3D.pointAt()` was a no-op, leaving the camera pointed at its previous direction while the game projected the ball into 2D coordinates. That caused the shooter or ball to appear only after it had already crossed the goal. The UVT culling and projection-plane guards address the related perspective rendering failures.

## Validation

- `avm2/matrix3d_pointat` passes in Ruffle.
- The surrounding Matrix3D suite (6 tests) passes.
- `avm2/utils3d` passes.
- WebAssembly and self-hosted web builds complete successfully.
- The deployed Firebase build's WASM hash matches the local build.


Lord-McSweeney:
Please use the PR template for opening new PRs.

Dbono-dev:
Closing at the author’s request; this fork is for personal use and is not intended as an upstream product PR.

PR #24404, opened 2026-08-09
avm2: Reimplement Matrix3D.recompose() as a native method

kjarosh:
## Description

This patch removes the AS-based Matrix3D.recompose() implementation and
makes it a native method instead. The implementation is based on the AS
one, but edge cases are handled properly.

Based on those edge cases, we should be multiplying matrices instead of
doing low-level operations on raw data (observable via e.g. NaN
propagation).

## Testing

Existing tests + test included.

## Checklist


SuchAFuriousDeath (approved):

core/src/avm2/globals/flash/geom/matrix_3d.rs
```diff
@@ -0,0 +1,182 @@
...
+    #[avm2_variant("eulerAngles")]
+    EulerAngles,
+    #[avm2_variant("axisAngle")]
+    AxisAngle,
+    #[avm2_variant("quaternion")]
+    Quaternion,
+}
+
+/// Reads the `x`, `y`, `z` and `w` components of a `Vector3D`, or `None` if it's null.
+fn read_vector3d(value: Value<'_>) -> Option<[f64; 4]> {
```
Lord-McSweeney: IMO it's nicer for this method to take `Option<Object>`- then its caller doesn't need to manifest a temporary `Value::Undefined`.

kjarosh: Done

PR #24405, opened 2026-08-09
avm2: Reimplement Matrix3D.decompose() as a native method

kjarosh:
## Description

This patch removes the AS-based Matrix3D.decompose() implementation and
makes it a native method instead.

## Testing

Existing tests.

## Checklist


core/src/avm2/globals/flash/geom/matrix_3d.rs
```diff
@@ -50,6 +64,22 @@ fn multiply(lhs: &RawData, rhs: &RawData) -> RawData {
...
+        .get_slot(matrix3d_slots::_RAW_DATA)
+        .as_object()
+        .expect("rawData is never null");
+    let raw_data = raw_data.as_vector_storage().expect("rawData is a Vector");
+
+    std::array::from_fn(|i| {
+        raw_data
+            .get_optional(i)
+            .map(|value| value.as_f64())
+            .unwrap_or(f64::NAN)
```
Lord-McSweeney, line 80: The raw data is guaranteed to have all 16 elements, right? So this doesn't need to fill missing elements with NaN.

core/src/avm2/globals/flash/geom/matrix_3d.rs
```diff
@@ -34,6 +34,20 @@ fn read_vector3d(value: Value<'_>) -> Option<[f64; 4]> {
...
+    activation: &mut Activation<'_, 'gc>,
+    components: [f64; 4],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let [x, y, z, w] = components;
+
+    activation
+        .avm2()
+        .classes()
+        .vector3d
+        .construct(activation, &[x.into(), y.into(), z.into(), w.into()])
```
Lord-McSweeney, line 48: This can do `.unwrap()` after it to avoid propagating the error around

core/src/avm2/globals/flash/geom/matrix_3d.rs
```diff
@@ -180,3 +210,123 @@ pub fn recompose<'gc>(
...
+                rotation[0] = mr[6].atan2(mr[10]);
+                rotation[2] = mr[1].atan2(mr[0]);
+            } else {
+                rotation[2] = 0.0;
+                rotation[0] = mr[4].atan2(mr[5]);
+            }
+        }
+    }
+
+    // It looks like Flash Player can leak uninitialized memory here when it
```
Lord-McSweeney: nit (my personal preferences, not blocking):
1. Could you use a different phrase than "leak...memory" here (e.g. "return uninitialized values")?
2. This is a pretty important discovery; could you mark it with a "NOTE"?

Lord-McSweeney (commented):
Seems like a lot of the implementation of `decompose` wasn't getting any coverage; to be sure, have you tested with this PR on SWFs that use it to make sure nothing broke?

core/src/avm2/globals/flash/geom/matrix_3d.rs
kjarosh: Currently it can have fewer than 16 elements, because the user might set it this way. I suspect it's a bug in Ruffle and I generally plan to get rid of `Vector` as raw storage, so I'd leave it as-is for now.

core/src/avm2/globals/flash/geom/matrix_3d.rs
kjarosh: Sorry if it's a stupid question, but why wouldn't we want to propagate the error? Things like OOM or similar errors could potentially be thrown here (if not now then in the future).

kjarosh:
> Seems like a lot of the implementation of decompose wasn't getting any coverage; to be sure, have you tested with this PR on SWFs that use it to make sure nothing broke?

I haven't found any SWFs that use it, I wrote tests but they were failing to some extent and decided to investigate in a follow-up PR. They fail the same way before and after the PR BTW.

core/src/avm2/globals/flash/geom/matrix_3d.rs
kjarosh: Done

core/src/avm2/globals/flash/geom/matrix_3d.rs
Lord-McSweeney: Both Flash and Ruffle never throw such errors; this should be a completely infallible operation. If something went wrong here it's completely unexpected, so IMO it's better to panic here than keep going on. Either way, user code calling this will likely fail because of the error.

(also, we do this everywhere else that we perform infallible constructions of classes)

Lord-McSweeney (approved):

kjarosh:
> They fail the same way before and after the PR BTW.

Of course to the extend of the stuff I fixed, which is thrown exceptions.

core/src/avm2/globals/flash/geom/matrix_3d.rs
kjarosh: Makes sense, done

PR #24407, opened 2026-08-09
avm2: Implement Matrix3D.copyColumn*() as native methods

kjarosh:
## Description

This implements Matrix3D.copyColumn*() methods as native and handles edge cases properly.

## Testing

Tests included.

## Checklist


core/src/avm2/globals/flash/geom/matrix_3d.rs
```diff
@@ -128,6 +128,64 @@ pub fn prepend<'gc>(
...
+) -> Result<Value<'gc>, Error<'gc>> {
+    let this = this.as_object().unwrap();
+    let column = args.get_u32(0);
+    let vector3d = args.get_object(activation, 1, "vector3D")?;
+
+    if column > 3 {
+        return Err(make_error_2004(activation, Error2004Type::ArgumentError));
+    }
+
+    let [x, y, z, w] = [
```
Lord-McSweeney: This should use `read_vector3d`, right?

kjarosh: It originally did, but I would have to wrap it in `Option` and `Value` and then unwrap the result, so I chose to use slots directly, similarly to setters.

Lord-McSweeney: IMO it would be better then to make `read_vector3d` take an `Object` directly and make its other callers responsible for unwrapping `Option<Value>`.

kjarosh: Added a separate function

Lord-McSweeney (approved):

PR #24408, opened 2026-08-09
avm2: Make Matrix3D.determinant/invert() native 

kjarosh:
## Description

This makes Matrix3D.determinant and Matrix3D.invert() native. This does not change their implementations.

## Testing

Already covered by tests.

## Checklist

### DIFF
--- a/core/src/avm2/globals/flash/geom/matrix_3d.rs
+++ b/core/src/avm2/globals/flash/geom/matrix_3d.rs
@@ -185,6 +185,16 @@ pub fn copy_column_from<'gc>(
     Ok(Value::Undefined)
 }
 
+/// Computes the determinant of a matrix.
+fn determinant_of(mr: &RawData) -> f64 {
+    (mr[0] * mr[5] - mr[4] * mr[1]) * (mr[10] * mr[15] - mr[14] * mr[11])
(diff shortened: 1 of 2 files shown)


core/src/avm2/globals/flash/geom/matrix_3d.rs
```diff
@@ -185,6 +185,87 @@ pub fn copy_column_from<'gc>(
...
+    _args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let this = this.as_object().unwrap();
+    let mr = raw_data_of(this);
+
+    let d = determinant_of(&mr);
+    let invertible = d.abs() > 0.00000000001;
+
+    if !invertible {
+        return Ok(false.into());
```
Lord-McSweeney, line 222: (not related to this PR) We should definitely have test coverage for this...

Lord-McSweeney (approved):

core/src/avm2/globals/flash/geom/matrix_3d.rs
kjarosh: Will do, don't worry. I started adding such tests but the fact that other Matrix methods are inaccurate, some stuff started breaking... So I'm gonna have to improve accuracy first in many cases before adding proper thresholds.

PR #24425, opened 2026-08-10
avm2: Improve the behavior of `TextBlock.releaseLines`

Lord-McSweeney:
## Description

- Make `TextBlock.releaseLines` actually match Flash
- Make the behavior of setting the validity of lines to "static" actually match Flash

## Testing

There was existing test coverage for the changes in the tests `avm2/textblock_line_changes` and `avm2/textblock_releaselines`. These tests have been made `known_failure` rather than `ignore` so that we actually run them now (previously they were panicking). I've also added more coverage in those two tests.

## Checklist


core/src/display_object/text_line.rs
```diff
@@ -160,17 +163,38 @@ impl<'gc> TextLine<'gc> {
...
+        }
+
         unlock!(Gc::write(mc, self.0), TextLineData, validity).set(validity);
     }
 
     pub fn text_block(self) -> Option<TextBlockObject<'gc>> {
         self.0.text_block.get()
     }
 
+    pub fn text_block_hide_from_script(self) -> Option<TextBlockObject<'gc>> {
```
kjarosh:
```suggestion
    pub fn text_block_from_script(self) -> Option<TextBlockObject<'gc>> {
```

The fact that you hide something is an implementation detail.

core/src/avm2/globals/flash/text/engine/text_block.rs
```diff
@@ -360,23 +360,61 @@ pub fn release_lines<'gc>(
...
         .expect("Guaranteed by AS signature");
 
-    // NOTE: `releaseLines` allows for swapping the order of parameters without
-    // affecting the functionality of the method!
-
-    // Both lines must have associated text blocks
-    let (Some(line_1_block), Some(line_2_block)) = (line_1.text_block(), line_2.text_block())
-    else {
+    // The behavior of this method is a bit unintuitive. What Flash seems to do
+    // is the following:
```
kjarosh: Can you, more importantly, include *why* you think that's the case? This way the reader can update this model and understand it better when they discover conflicting evidence.

kjarosh (approved):

PR #24437, opened 2026-08-11
avm2: Implement Matrix3D.copyRawDataFrom() properly

kjarosh:
## Description

Implement Matrix3D.copyRawDataFrom() as a native method based on tests.

## Testing

Tests included.

## Checklist


core/src/avm2/globals/flash/geom/matrix_3d.rs
```diff
@@ -347,6 +347,34 @@ pub fn copy_row_from<'gc>(
...
+        return Err(make_error_2004(activation, Error2004Type::ArgumentError));
+    }
+
+    let raw_data: RawData =
+        std::array::from_fn(|i| source.get_optional(index + i).unwrap().as_f64());
+
+    set_raw_data(activation, this, raw_data)?;
+
+    if do_transpose {
+        transpose(activation, this.into(), &[])?;
```
Lord-McSweeney, line 372: nit: I'd prefer it if code for transposition of a `RawData` were in its own function (similarly to `determinant_of`) so it could be reused in both this function and `transpose`, but this way works fine too

Lord-McSweeney (approved):

core/src/avm2/globals/flash/geom/matrix_3d.rs
kjarosh: I know, but that's gonna get rewritten anyway when I implement native storage, so not worth doing that rn.

PR #24442, opened 2026-08-12
avm2: Use native Matrix3D in AVM2

kjarosh:
## Description

This PR makes AVM2 use the native Matrix3D as data storage (instead of a `_rawData` vector), moves some methods into that class, and simplifies existing code.

## Testing

Covered by existing tests.

## Checklist

### DIFF
--- a/core/src/avm2/globals/flash/geom/Matrix3D.as
+++ b/core/src/avm2/globals/flash/geom/Matrix3D.as
@@ -5,28 +5,14 @@ package flash.geom {
 
     [Ruffle(InstanceAllocator)]
     public class Matrix3D {
-        // The 4x4 matrix data, stored in column-major order
-        // This is never null.
-        [Ruffle(NativeAccessible)]
-        private var _rawData:Vector.<Number>;
-
-        public function get rawData():Vector.<Number> {
-            return this._rawData.AS3::concat();
-        }
-
-        public function set rawData(value:Vector.<Number>):void {
-            if (value != null) {
-                this._rawData = value.AS3::concat();
-            }
-        }
-
         public function Matrix3D(v:Vector.<Number> = null) {
             if (v != null && v.length == 16) {
-                this._rawData = v.AS3::concat();
-            } else {
-                this.identity();
+                this.rawData = v;
             }
         }
+
+        public native function get rawData():Vector.<Number>;
+        public native function set rawData(value:Vector.<Number>):void;
 
         public native function identity():void;
 
@@ -86,8 +72,7 @@ package flash.geom {
 
         [API("674")]
         public function copyFrom(other:Matrix3D):void {
-            // This makes a copy of other.rawData
-            this._rawData = other.rawData;
+            this.rawData = other.rawData;
         }
 
         [API("674")]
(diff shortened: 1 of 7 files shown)


core/src/avm2/globals/flash/geom/matrix_3d.rs
```diff
@@ -53,93 +55,79 @@ fn vector3d_to_object<'gc>(
...
+    let value = value.as_vector_storage().unwrap();
 
-    std::array::from_fn(|i| {
-        raw_data
+    let raw_data: RawData = std::array::from_fn(|i| {
+        value
             .get_optional(i)
-            .map(|value| value.as_f64())
+            .map(|v| v.as_f64())
             .unwrap_or(f64::NAN)
```
Lord-McSweeney, line 95: What happens in FP when the vector is not of the correct length or is `null`?

core/src/avm2/globals/flash/geom/Matrix3D.as
```diff
@@ -95,7 +81,7 @@ package flash.geom {
                 this.transpose();
             }
 
-            var d:Vector.<Number> = this._rawData;
+            var d:Vector.<Number> = this.rawData;
```
Lord-McSweeney: This makes a copy of `this.rawData`, so this method will no longer work correctly

core/src/avm2/globals/flash/geom/Matrix3D.as
```diff
@@ -3,30 +3,17 @@
...
-        }
-
-        public function set rawData(value:Vector.<Number>):void {
-            if (value != null) {
-                this._rawData = value.AS3::concat();
-            }
-        }
-
         public function Matrix3D(v:Vector.<Number> = null) {
             if (v != null && v.length == 16) {
```
Lord-McSweeney, line 9: Have you tested whether this if-statement's checks can be removed?

render/src/matrix3d.rs
```diff
@@ -46,4 +77,96 @@ impl Matrix3D {
...
+    pub fn transpose_in_place(&mut self) {
+        self.raw_data.swap(1, 4);
+        self.raw_data.swap(2, 8);
+        self.raw_data.swap(3, 12);
+        self.raw_data.swap(6, 9);
+        self.raw_data.swap(7, 13);
+        self.raw_data.swap(11, 14);
+    }
+
+    pub fn transpose(mut self) -> Self {
```
Lord-McSweeney: Is this method being used anywhere?

Lord-McSweeney (approved):
Nice!

core/src/avm2/globals/flash/geom/matrix_3d.rs
kjarosh: FP is more strict when it comes to input, I'm working on it in a follow-up PR.

core/src/avm2/globals/flash/geom/Matrix3D.as
kjarosh: I will take care of the remaining AS methods in follow-up PRs. The plan is to make them fully native.

core/src/avm2/globals/flash/geom/Matrix3D.as
kjarosh: Ahh shoot, forgot to migrate this method. I got it confused with `copyRawDataFrom` lol

render/src/matrix3d.rs
kjarosh: Removed, I added it because I assumed it will be needed in the future but now thinking about it I'm not sure.

core/src/avm2/globals/flash/geom/Matrix3D.as
kjarosh: Wait a minute, I think I got it right, how would this break the method? `d` is not written to.

Lord-McSweeney: Oops, right, my bad!

Lord-McSweeney (approved):

Issue #18734, opened 2024-11-24
Regression in Rainbow Dash Shooter Project on nightly build 2024-11-24


SN902:
### Describe the bug

We can now see all the hidden 2D pony models on the game's title screen

![Image](https://github.com/user-attachments/assets/f8b51ed5-b60d-4927-b6ff-a1cd20365b71)

### Expected behavior

![Image](https://github.com/user-attachments/assets/c0c622b0-52a6-494a-b2c3-0dd8154854b3)

### Content Location

[Rainbow Dash Shooter Project.zip](https://github.com/user-attachments/files/17891325/Rainbow.Dash.Shooter.Project.zip)

### Affected platform

Desktop app

### Operating system

Windows 11 x64

### Browser

Firefox nightly 134.0a1

n0samu:
Fixed by #18742

PR #18692, opened 2024-11-20
wgpu: Perform blends as-needed, instead of ahead of time. Reduces memory usage for multiple blends significantly

Dinnerbone:
This fixes #18317 and should help a lot with any movie that does multiple blends on the same frame (unless they're nested). For me, it reduces VRAM of that game during loading down from **13 GB** to just shy of **500 MB**.

Graphics memory requirement goes from `(size of stage) * (number of total blends)` to just `(size of stage) * 1` for most cases.

Imagine a hypothetical draw list:
DrawRect, DrawCircle, BlendTrivial(DrawTriangle), BlendComplex(DrawStar), DrawSmilyFace

Before it would do this:
- DrawTriangle into a temp buffer A
- DrawStar into a temp buffer B
- DrawRect, DrawCircle, DrawTempBufferA all as one render pass
- DrawTempBufferB (as one render pass)
- DrawSmilyFace as one render pass

This meant for every (not nested) blend, we needed that many temp textures.

With this change, we instead do this:
- DrawRect, DrawCircle as one render pass
- DrawTriangle to TempBuffer as one render pass
- DrawTempBuffer as one render pass
- DrawStar to TempBuffer as one render pass
- DrawTempBuffer as one render pass
- DrawSmilyFace as one render pass

I'm not super happy with the code but I don't have better ideas for organization right now :(

PR #24445, opened 2026-08-13
Codex/fix 21515 lazy blends

### LINKED ISSUE #21515: Pufflescape (Club Penguin) - Sprites with the "Layer" blend mode causes huge GPU and VRAM usage.
### Describe the bug

(More detailed version of #19618)

When the game boots up, the GPU usage spikes and the VRAM usage increases significantly (depending on the resolution and quality setting).
The main reason for this is the heavy usage of the "Layer" blend mode on the level selection screen.
The lag stops as soon as the menu is unloaded.

This sprite is a perfect example (Tested on Intel iGPU):
Without blending:
Image
With blending:
Image

### Expected behavior

This game should work fine.

### Content Location

[withBlend_withoutBlend.zip](https://github.com/user-attachments/files/22058987/withBlend_withoutBlend.zip)

### Affected platform

Online demo

### Operating system

Windows 11

### Browser

Firefox 142.0.1

### Additional information

_No response_

atilapan:
## Description

Fixes #21515.

Ruffle previously rendered every blend subtree into a full-stage temporary
texture while chunking the command list. These textures remained alive until
the parent surface finished rendering. Content with many sibling `Layer`
blends, such as Pufflescape, could therefore consume substantial VRAM and GPU
resources.

This PR changes WGPU blend rendering so that:

- Blend command lists remain lazy until they are ready to be composited.
- Each temporary child texture can return to the texture pool before the next
  sibling blend is rendered.
- Peak temporary texture usage follows blend nesting depth instead of the total
  number of sibling blends.
- Complex and shader blend backdrops are captured before rendering their child
  command lists.

Capturing the backdrop first is important because rendering a child can update
the nearest layer. An earlier lazy-rendering implementation in #18692 captured
it afterward, causing #18734, and was subsequently reverted in #18742.

This PR also avoids creating an intermediate texture for a `Layer` whose
commands only contain normal drawing or nested `Layer`/`Normal` compositing.
Layers containing nontrivial builtin blends or shader blends remain isolated
and use the lazy rendering path.

Affected issues and prior changes:

- Fixes https://github.com/ruffle-rs/ruffle/issues/21515
- Regression checked: https://github.com/ruffle-rs/ruffle/issues/18734
- Earlier implementation: https://github.com/ruffle-rs/ruffle/pull/18692
- Earlier revert: https://github.com/ruffle-rs/ruffle/pull/18742

## Testing

An automatic unit test,
`layer_isolation_is_only_required_for_exposed_blends`, was added. It verifies
that empty, normal, and nested-layer command lists can be flattened while
nontrivial blends still require isolation.

The following checks were run:

```sh
cargo fmt --all --check
cargo test -p ruffle_render
cargo test -p ruffle_render_wgpu
cargo clippy -p ruffle_render_wgpu --all-targets -- -D warnings
cd web
npm run build
```

The production web build completed for the demo, self-hosted player, generic
extension, and Firefox extension.

### Manual testing for #21515

1. Download `withBlend_withoutBlend.zip` from #21515.
2. Open `withBlend.swf` using a WebGPU build from this PR.
3. Compare it with `withoutBlend.swf`.
4. Verify that both files render correctly and that `withBlend.swf` no longer
   causes the previous extreme GPU/VRAM behavior.
5. Pufflescape's level-selection screen can be used as a larger real-world
   test, especially at higher resolutions and quality settings.

### Regression testing for #18734

The original `Rainbow Dash Shooter Project.swf` attachment from #18734 was
rendered using both:

- A clean build of `upstream/master`.
- A build containing both changes in this PR.

Both were captured using the same browser, viewport, playback state, and
production WASM build process. Their 550×400 output images were pixel-identical:
0 differing pixels out of 220,000. The hidden title-screen models remained
hidden.

The real-world SWF was tested manually rather than committed as a repository
fixture because it is a large externally supplied game attachment.

## Checklist


Dinnerbone:
Welcome and thank you!

> An automatic unit test,
> `layer_isolation_is_only_required_for_exposed_blends`, was added. It verifies
> that empty, normal, and nested-layer command lists can be flattened while
> nontrivial blends still require isolation.

I do not see this test.

Dinnerbone:
Ah, I see it - can you please make a swf regression test that would fail in the other PR but passes here?

atilapan:
Done - a SWF visual regression test was added that reduces the masked Layer/cacheAsBitmap behavior from #18734.
The expected image was captured with the Adobe Flash Player 32 content debugger. With the reverted implementation, the test fails because the cached shapes incorrectly become visible. With this PR, it passes and they aren't visible. The original Rainbow Dash Shooter Project was also manually verified after clicking Start, the hidden pony models remain hidden.

Dinnerbone:
Confirmed that the test fails on the prior implementation, and passes here. Thank you so much!

Dinnerbone (approved):

Lord-McSweeney:
This caused a regression in rendering, see #24549

PR #24447, opened 2026-08-13
Context3d matrix3d features

SulivanM:
## Description

This PR implements three unrelated Flash APIs (each in its own commit) :

- **`Context3D.drawToBitmapData`** replaces the stub with a native method that reads the current back buffer and blits it into the destination `BitmapData`. Adds a `back_buffer_handle` accessor to the `Context3D` render backend trait to distinguish the front (post-`present`) and back buffers.
- **`Context3D.dispose(recreate)`** now honors the `recreate` argument. With `recreate = true` (the default) the context stays usable; only `recreate = false` clears it. Previously `dispose` always cleared the context, breaking content that disposes and reuses its Stage3D (e.g. on resize), leaving that layer black.
- **`Matrix3D.interpolate` / `interpolateTo`** implemented on top of `decompose`/`recompose`, interpolating translation and scale linearly and the rotation via slerp.

## Testing

SWF regression tests are included :

- `avm2/matrix3d_interpolate` checks translation/scale midpoints (exact), a 45° slerp midpoint, endpoints, `interpolate(A, A, t) == A`, and in-place `interpolateTo`.
- `avm2/context3d_draw_to_bitmap_data` draws a red quad over a blue clear, reads the back buffer into a green `BitmapData`, and asserts the pixels are red (needs a renderer; run with `--features imgtests`).

`Context3D.dispose(recreate)` is not covered by an automated test: its effect is visual (the Stage3D layer no longer goes black after a dispose-and-reuse).
`It can be verified on real content that reconfigures its context after `dispose()`.

## Checklist


core/src/avm2/globals/flash/geom/Matrix3D.as
```diff
@@ -105,5 +105,61 @@ package flash.geom {
 
         public native function get determinant():Number;
 
+        public function interpolateTo(toMat:Matrix3D, percent:Number):void {
+            this.rawData = Matrix3D.interpolate(this, toMat, percent).rawData;
+        }
+
+        public static function interpolate(thisMat:Matrix3D, toMat:Matrix3D, percent:Number):Matrix3D {
```
kjarosh, line 112: How did you infer the interpolation method used by Flash?

kjarosh:
This looks like 3 unrelated changes, they should be filed as separate PRs.

core/src/avm2/globals/flash/geom/Matrix3D.as
SulivanM: I built it on the existing decompose/recompose in "quaternion" orientation : translation and scale interpolated componentwise (linear), rotation via slerp of the decomposed quaternions.
 
I confirmed the method by running the attached test.swf in Flash Player 32.0.0.465 (debug projector), the traced rawData matches to 6 decimals, and output.txt is that captured Flash output.

SulivanM:
You're right, these are unrelated. I'll split them into three PRs and close this one.

SulivanM:
Splitting into separate PRs as suggested: drawToBitmapData, dispose(recreate), and Matrix3D.interpolate.

PR #24448, opened 2026-08-13
avm2: Implement Context3D.drawToBitmapData

SulivanM:
## Description

`Context3D.drawToBitmapData` replaces the stub with a native method that reads the current back buffer and blits it into the destination `BitmapData`.
Adds a `back_buffer_handle` accessor to the `Context3D` render backend trait to distinguish the front (post-`present`) and back buffers.

## Testing

- `avm2/context3d_draw_to_bitmap_data` draws a red quad over a blue clear, reads the back buffer into a green `BitmapData`, and asserts the pixels are red (needs a renderer; run with `--features imgtests`).

## Checklist


core/src/avm2/globals/flash/display3D/context_3d.rs
```diff
@@ -221,6 +225,58 @@ pub fn present<'gc>(
...
+    this: Value<'gc>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let this = this.as_object().unwrap();
+
+    if let Some(context) = this.as_context_3d() {
+        let destination = args
+            .get_object(activation, 0, "destination")?
+            .as_bitmap_data()
+            .expect("Argument is BitmapData-typed");
```
kjarosh: This should be done before the if, also you can `unwrap()`.

core/src/avm2/globals/flash/display3D/context_3d.rs
```diff
@@ -221,6 +225,58 @@ pub fn present<'gc>(
...
+pub fn draw_to_bitmap_data<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    this: Value<'gc>,
+    args: &[Value<'gc>],
+) -> Result<Value<'gc>, Error<'gc>> {
+    let this = this.as_object().unwrap();
+
+    if let Some(context) = this.as_context_3d() {
+        let destination = args
+            .get_object(activation, 0, "destination")?
```
kjarosh, line 237: This will throw a specific error when the argument is null, have you tested this case? It should be covered by the test.

tests/tests/swfs/avm2/context3d_draw_to_bitmap_data/test.toml
```diff
@@ -0,0 +1,11 @@
+num_frames = 1
```
kjarosh: Prefer `num_ticks` to `num_frames`

tests/tests/swfs/avm2/context3d_draw_to_bitmap_data/Test.as
```diff
@@ -0,0 +1,87 @@
...
+            var program:Program3D = context.createProgram();
+            program.upload(vertexAssembly.agalcode, fragmentAssembly.agalcode);
+            context.setProgram(program);
+
+            // Clear to blue, then draw the red quad on top.
+            context.clear(0, 0, 1, 1);
+            context.drawTriangles(indexBuffer, 0, 2);
+
+            // Read the back buffer into a green-filled BitmapData.
+            var bmd:BitmapData = new BitmapData(50, 50, true, 0xff00ff00);
```
kjarosh: Have you considered adding a visual test instead of reading pixels back?

tests/tests/swfs/avm2/context3d_draw_to_bitmap_data/Test.as
```diff
@@ -0,0 +1,87 @@
...
+            context.drawToBitmapData(bmd);
+
+            trace("pixel(0,0):   " + hex(bmd.getPixel32(0, 0)));
+            trace("pixel(25,25): " + hex(bmd.getPixel32(25, 25)));
+            trace("pixel(49,49): " + hex(bmd.getPixel32(49, 49)));
+        }
```
Lord-McSweeney, line 113: What happens if you call `present` and try to `drawToBitmapData` again?

core/src/avm2/globals/flash/display3D/context_3d.rs
```diff
@@ -221,6 +225,58 @@ pub fn present<'gc>(
...
+        );
+
+        let handle = destination.bitmap_handle(activation.gc(), activation.context.renderer);
+
+        let (destination, include_dirty_area) =
+            destination.overwrite_cpu_pixels_from_gpu(activation.gc());
+        let mut write = destination.borrow_mut(activation.gc());
+
+        let mut dirty_region = PixelRegion::for_whole_size(write.width(), write.height());
+        if let Some(old) = include_dirty_area {
```
Lord-McSweeney: If the dirty region includes the entire bitmap data, what's the point of this `union`?

core/src/avm2/globals/flash/display3D/context_3d.rs
```diff
@@ -221,6 +225,58 @@ pub fn present<'gc>(
...
+        let image = activation.context.renderer.render_offscreen(
+            handle,
+            commands,
+            StageQuality::High,
+            dirty_region,
+        );
+
+        if let Some(sync_handle) = image {
+            write.set_gpu_dirty(activation.gc(), sync_handle, dirty_region);
+        }
```
Lord-McSweeney, line 283: If `image` is `None`, the render backend doesn't support offscreen rendering. Currently, there are no backends that support Context3D but don't support offscreen rendering, but there may be one (or a user-provided one) in the future. Can you add an
```
} else {
    return Err("Render backend does not support Context3D.drawToBitmapData".into());
}
```
?

core/src/avm2/globals/flash/display3D/context_3d.rs
Lord-McSweeney: Doing it after the if is consistent with the rest of the methods in this file. Is there some other advantage to having it before the if?

kjarosh: Oh hmm, I thought that we always prepared args at the very beginning, but looking at the code I see it's not consistent, so nevermind.

SulivanM: Kept it after the if, switched expect -> unwrap.

core/src/avm2/globals/flash/display3D/context_3d.rs
SulivanM: Done, the test now calls drawToBitmapData(null) and asserts the #2007 error.

tests/tests/swfs/avm2/context3d_draw_to_bitmap_data/test.toml
SulivanM: Done.

tests/tests/swfs/avm2/context3d_draw_to_bitmap_data/Test.as
SulivanM: The feature is the read-back itself, so I check the pixels directly. Can add a visual test too if you prefer ?

tests/tests/swfs/avm2/context3d_draw_to_bitmap_data/Test.as
SulivanM: present() swaps front/back, so a second call reads the swapped buffer.
I'll verify against real Flash and add a test.

core/src/avm2/globals/flash/display3D/context_3d.rs
SulivanM: Removed, the region is always the full bitmap, so the union added nothing.

core/src/avm2/globals/flash/display3D/context_3d.rs
SulivanM: Added.

tests/tests/swfs/avm2/context3d_draw_to_bitmap_data/Test.as
kjarosh: Yes, but I'm talking about asserting on the bitmap. It would be nice to assert on all pixels, but it's easier when you display the bitmap and assert on the visual output.

I think a visual output would be preferred because it asserts on the whole bitmap and it's easier to see what's wrong in case it fails.

In order to do that, just add the bitmap as a child and set proper SWF dimensions via `[SWF()]`.

tests/tests/swfs/avm2/context3d_draw_to_bitmap_data/Test.as
kjarosh: You can extend this test, if you add visual output you can place resulting bitmaps next to each other.

SulivanM: Sure, thanks i go to check this !
Thanks for the review !

SulivanM: Added the visual side-by-side test: it draws a red quad and a green quad, reads each back with drawToBitmapData, and shows them next to each other (red left, green right).

output.expected.png is captured from Flash Player 32.0.0.371, and Ruffle matches it :P

tests/tests/swfs/avm2/context3d_draw_to_bitmap_data/Test.as
```diff
@@ -0,0 +1,121 @@
...
+            trace("pixel(0,0):   " + hex(bmd.getPixel32(0, 0)));
+            trace("pixel(25,25): " + hex(bmd.getPixel32(25, 25)));
+            trace("pixel(49,49): " + hex(bmd.getPixel32(49, 49)));
+
+            // A null destination throws TypeError #2007.
+            try {
+                context.drawToBitmapData(null);
+                trace("null: no error");
+            } catch (e:Error) {
+                trace("null: " + e);
```
kjarosh:
```suggestion
                trace("null: " + e.getStackTrace());
```

tests/tests/swfs/avm2/context3d_draw_to_bitmap_data/Test.as
Lord-McSweeney: Thanks for the extra test coverage, but this still doesn't test what happens if `drawToBitmapData` is called after `present`. Could you add coverage for that too?

SulivanM: Done, added coverage for drawToBitmapData after present().

SulivanM: I added this side, i tested it ;)

SulivanM:
Need to be tested from some contributor, look like perfect in my side.

SulivanM:
Rebased on master, fixed the build: drawToBitmapData now uses FunctionArgs and the new render_bitmap region (back buffer size, 1:1). Passes locally.

tests/tests/swfs/avm2/context3d_draw_to_bitmap_data/Test.as
```diff
@@ -0,0 +1,133 @@
...
+            } catch (e:Error) {
+                trace("after present: " + e.getStackTrace());
+            }
+
+            // Show both captures next to each other: red on the left, green on
+            // the right.
+            addChild(new Bitmap(bmd));
+            var right:Bitmap = new Bitmap(bmd2);
+            right.x = 50;
+            addChild(right);
```
Lord-McSweeney: Could you add test coverage for what happens when trying to draw the stage3d into a `BitmapData` that is larger than the stage size and into a `BitmapData` that is smaller than the stage size?

tests/tests/swfs/avm2/context3d_draw_to_bitmap_data/test.toml
```diff
@@ -0,0 +1,16 @@
+num_ticks = 1
+
+[image_comparisons.output]
+tolerance = 4
+max_outliers = 24
+
+[player_options]
+with_renderer = { optional = false, quality = "high" }
+viewport_dimensions = { width = 100, height = 50, scale_factor = 1 }
```
Lord-McSweeney: Just curious, why does this need `viewport_dimensions` to be set?

tests/tests/swfs/avm2/context3d_draw_to_bitmap_data/Test.as
SulivanM: Done, now reads the back buffer into a same-size, larger and smaller BitmapData.
Larger keeps its fill outside the back buffer, smaller is clipped (not scaled). Verified against Flash ;)

tests/tests/swfs/avm2/context3d_draw_to_bitmap_data/test.toml
SulivanM: Good catch yup, it just duplicated the default (viewport comes from the SWF 100×50 stage), so i removed it.

Lord-McSweeney (approved):

SulivanM:
@kjarosh Hello !
I know you're a great contributor to Ruffle, so I was wondering if you could take a look at this PR when you have some free time.
 
Best regards

tests/tests/swfs/avm2/context3d_draw_to_bitmap_data/test.toml
```diff
@@ -0,0 +1,15 @@
+num_ticks = 1
+
+[image_comparisons.output]
+tolerance = 4
+max_outliers = 24
```
kjarosh: Where does the tolerance / max outliers come from? The output looks like it's a trivial, pixel-perfect image.

SulivanM: You're right, it's pixel-perfect i dont remember to change it..
I removed them so it compares exactly (0 tolerance, 0 outliers) in latest change.

Issue #3694, opened 2021-03-17
Error on https://uploads.ungrounded.net/ruffle_wrapper/ruffleembed.html


Msr-Shift:
# Error Info
Error name: Error
Error message: panicked at 'assertion failed: `(left == right)`
  left: `true`,
 right: `false`: cannot recursively acquire mutex', /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/std/src/sys/wasm/../unsupported/mutex.rs:23:9
Error stack:
```
Error: panicked at 'assertion failed: `(left == right)`
  left: `true`,
 right: `false`: cannot recursively acquire mutex', /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/std/src/sys/wasm/../unsupported/mutex.rs:23:9
    at a.wbg.__wbg_new_df6e6ab7a65c4c4d (https://uploads.ungrounded.net/ruffle_wrapper/lib/ruffle.js?1615731147:1:6553)
    at std::sync::once::Once::call_once::{{closure}}::hce6501f6f20767e4 (https://uploads.ungrounded.net/ruffle_wrapper/lib/6cbd7506663db5f7c175101a0d35e2ea.wasm:wasm-function[1470]:0x2bf7fd)
    at std::sync::once::Once::call_inner::hd3d269a4946f2df2 (https://uploads.ungrounded.net/ruffle_wrapper/lib/6cbd7506663db5f7c175101a0d35e2ea.wasm:wasm-function[1734]:0x2f0c79)
    at core::ops::function::FnOnce::call_once{{vtable.shim}}::h6ed7686ed76cb258 (https://uploads.ungrounded.net/ruffle_wrapper/lib/6cbd7506663db5f7c175101a0d35e2ea.wasm:wasm-function[5189]:0x3f7f7d)
    at std::panicking::rust_panic_with_hook::h4f753dc70b771d8e (https://uploads.ungrounded.net/ruffle_wrapper/lib/6cbd7506663db5f7c175101a0d35e2ea.wasm:wasm-function[3949]:0x3c8b27)
    at std::panicking::begin_panic_handler::{{closure}}::h5c9fc29957309ae8 (https://uploads.ungrounded.net/ruffle_wrapper/lib/6cbd7506663db5f7c175101a0d35e2ea.wasm:wasm-function[4670]:0x3ea8a3)
[...]

PR #24449, opened 2026-08-13
avm2: Respect the recreate argument in Context3D.dispose

SulivanM:
## Description

`Context3D.dispose(recreate)` now honors the `recreate` argument. With `recreate = true` (the default) the context stays usable; only `recreate = false` clears it.

Previously `dispose` always cleared the context, breaking content that disposes and reuses its Stage3D (e.g. on resize), leaving that layer black.

## Testing

`Context3D.dispose(recreate)` is not covered by an automated test: its effect is visual (the Stage3D layer no longer goes black after a dispose-and-reuse). It can be verified on real content that reconfigures its context after `dispose()`.

## Checklist


core/src/avm2/globals/flash/display3D/context_3d.rs
```diff
@@ -828,14 +828,22 @@ pub fn set_scissor_rectangle<'gc>(
...
-        .unwrap()
-        .stage3d()
-        .clear_context3d(activation.gc());
+    // Context3D.dispose(recreate:Boolean = true). With recreate=true (the
+    // default), Flash frees the context's resources but keeps the Context3D
+    // usable - the caller reconfigures the back buffer and keeps rendering.
+    // Only recreate=false permanently disposes it. Clearing the context
+    // unconditionally (as before) killed contexts that games dispose-and-reuse
+    // (e.g. on resize/scene reload), leaving that Stage3D layer black.
+    let recreate = args.first().is_none_or(|v| v.coerce_to_boolean());
```
kjarosh: It should use `get_bool`

core/src/avm2/globals/flash/display3D/context_3d.rs
```diff
@@ -828,14 +828,22 @@ pub fn set_scissor_rectangle<'gc>(
 pub fn dispose<'gc>(
     activation: &mut Activation<'_, 'gc>,
     this: Value<'gc>,
-    _args: &[Value<'gc>],
+    args: &[Value<'gc>],
 ) -> Result<Value<'gc>, Error<'gc>> {
     let this = this.as_object().unwrap();
 
-    avm2_stub_method!(activation, "flash.display3D.Context3D", "dispose");
```
kjarosh, line 830: Why was it a stub and why can we remove the stub annotation now?

core/src/avm2/globals/flash/display3D/context_3d.rs
```diff
@@ -828,14 +828,22 @@ pub fn set_scissor_rectangle<'gc>(
...
-    this.as_context_3d()
-        .unwrap()
-        .stage3d()
-        .clear_context3d(activation.gc());
+    // Context3D.dispose(recreate:Boolean = true). With recreate=true (the
+    // default), Flash frees the context's resources but keeps the Context3D
+    // usable - the caller reconfigures the back buffer and keeps rendering.
+    // Only recreate=false permanently disposes it. Clearing the context
+    // unconditionally (as before) killed contexts that games dispose-and-reuse
```
kjarosh: There's no point in referring to the past, if we did that Ruffle's source code would be 95% comments about how something worked in the past.

kjarosh:
Can you add a test for this as well? Ideally behavior of this method with recreate=true and recreate=false.

Lord-McSweeney:
Can you provide an example of real content that this PR fixes?

core/src/avm2/globals/flash/display3D/context_3d.rs
SulivanM: Done, switched to args.get_bool(0).

core/src/avm2/globals/flash/display3D/context_3d.rs
SulivanM: Done, rewrote the comment to describe only the current behavior.

SulivanM:
> Can you add a test for this as well? Ideally behavior of this method with recreate=true and recreate=false.

Added avm2/context3d_dispose : it disposes with recreate=true on the first context3DCreate (a second one must fire) and recreate=false on the second (no third, and Stage3D.context3D becomes null).
Output verified against Flash Player 32.0.0.371 like my other PR.

core/src/avm2/globals/flash/display3D/context_3d.rs
SulivanM: It was stubbed because dispose ignored its recreate argument and just cleared the context.
It now implements the real behavior, recreate=true makes Stage3D dispatch a fresh context3DCreate with a new context, recreate=false leaves it without one, so the stub no longer applies.

SulivanM:
> Can you provide an example of real content that this PR fixes?

Deadzone (https://deadzonegame.net/play/ruffle) it disposes and reconfigures its Stage3D, and currently goes black...
The added test shows the divergence : without this change Ruffle dispatches context3DCreate only once and never recreates the context, whereas Flash dispatches it again after dispose(true).

Also we can see the issue on console.

core/src/avm2/globals/flash/display3D/context_3d.rs
```diff
@@ -828,14 +828,21 @@ pub fn set_scissor_rectangle<'gc>(
...
-        .unwrap()
-        .stage3d()
-        .clear_context3d(activation.gc());
+    // recreate=true (the default) makes Stage3D dispatch a fresh context3DCreate
+    // with a new context, so re-request one; recreate=false leaves it without a
+    // context.
+    let recreate = args.get_bool(0);
+    let context = this.as_context_3d().unwrap();
+    let stage3d = context.stage3d();
+    if recreate {
```
Lord-McSweeney, line 836: What happens to the program/shader state in the Context3D after `dispose(true)` is called? Is it really preserved?

SulivanM: It's not preserved. dispose(true) disposes the current context and Stage3D dispatches a fresh context3DCreate with a brand-new Context3D, so all program/buffer/pipeline state is gone content re-uploads everything in the new handler, matching Flash.

I verified on Flash Player 32.0.0.371 : after dispose(true), reusing a Program3D/VertexBuffer3D that was created on the disposed context draws nothing (only the clear shows), because those objects belong to the old context.

Lord-McSweeney: Right, but what happens if the original context3d is stored somewhere? Is it still usable?

SulivanM: Currently it stays usable in Ruffle I verified: after dispose(true), a stored reference to the old Context3D still works, whereas Flash throws Error #3694 ("The object was disposed by an earlier call of dispose()").

Ruffle doesn't track a disposed state on Context3D at all (that predates this PR), so making the old reference throw #3694 means adding that state and guarding the ~28 Context3D methods.

Happy to do it, would you prefer it in this PR, or as a follow-up so this one stays focused on the recreate dispatch ?

kjarosh: Could you at least add a known_failure test just to document that behavior properly?

SulivanM: Yes, i go to made it !

SulivanM: Done, added avm2/context3d_dispose_reference as a known_failure.
It keeps a Context3D reference across dispose(true) and uses it on the next context: Flash throws Error #3694 while Ruffle keeps it usable.

Output verified against Flash Player 32.0.0.371 :)
Have a good day.

tests/tests/swfs/avm2/context3d_dispose_reference/Test.as
```diff
@@ -0,0 +1,34 @@
...
+            creates++;
+            if (creates == 1) {
+                oldContext = s3d.context3D;
+                oldContext.dispose(true);
+            } else if (creates == 2) {
+                try {
+                    oldContext.createVertexBuffer(3, 3);
+                    trace("old context still usable (no error)");
+                } catch (err:Error) {
+                    trace("errorID: " + err.errorID);
```
kjarosh:
```suggestion
                    trace("error: " + err.getStackTrace());
```

SulivanM: Miss this, thanks !

SulivanM:
Test wrong, need to look why in the day.

Issue #10259, opened 2023-03-19
New York Taxi License 3D: Property setVector not found on flash.display.BitmapData and there is no default value


MartySVK:
### Describe the bug

After clicking on PLAY, there are errors:

```
ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "addedToStage", class: flash.events::Event, ptr: 0x2941f005d20 }) to handler FunctionObject(FunctionObject { ptr: 0x2941f06bd40 }) : ReferenceError: Error #1069: Property setVector not found on flash.display.BitmapData and there is no default value.
ERROR ruffle_core::avm2::events: Error dispatching event EventObject(EventObject { type: "enterFrame", class: flash.events::Event, ptr: 0x2941f142b30 }) to handler FunctionObject(FunctionObject { ptr: 0x2941f1673e0 }) : TypeError: Error #1009: Cannot access a property or method of a null object reference. (accessing field: soundEnabled)
```

SWF - https://data.gameflare.com/games/3945/EyPXBggwZOKwEB.swf

### Expected behavior

Progress the game

### Affected platform

Desktop app

### Operating system

Windows 10

MartySVK:
After setVector PR, new error is:
``Property pointAt not found on flash.geom.Matrix3D and there is no default value``

SuchAFuriousDeath:
After stubbing flash.geom.Matrix3D.pointAt, it needs flash.geom.Matrix3D.interpolateTo and after that, it needs BitmapData.draw, which is being implemented in #12559

kjarosh:
It's now better but looks like graphics are broken, probably due to missing Matrix3D features (perspective projection?).

PR #24450, opened 2026-08-13
avm2: Implement Matrix3D.interpolate and interpolateTo

### LINKED ISSUE #10009: Shadez 3: Black Screen Issue

SulivanM:
## Description

* Fixes https://github.com/ruffle-rs/ruffle/issues/10009
* Progresses https://github.com/ruffle-rs/ruffle/issues/10259

`Matrix3D.interpolate` / `interpolateTo`, matching Flash Player 32.0.0.371.
The translation is interpolated linearly and the rotation via slerp of a quaternion read straight from the raw matrix (the scale is left in, so a rotation combined with a scale does not slerp to the naive angle), the scale itself is discarded, so the result is always unit-scaled.

## Testing

- `avm2/matrix3d_interpolate` checks translation midpoints/endpoints, the scale being discarded, a 45° rotation slerp, a rotation+scale case (midpoint ~58°, not 45°), `interpolate(A, A, t) == A`, and in-place `interpolateTo`.
Every expected value is captured from Flash Player 32.0.0.371.

## Checklist


core/src/avm2/globals/flash/geom/Matrix3D.as
```diff
@@ -105,5 +105,61 @@ package flash.geom {
 
         public native function get determinant():Number;
 
+        public function interpolateTo(toMat:Matrix3D, percent:Number):void {
+            this.rawData = Matrix3D.interpolate(this, toMat, percent).rawData;
+        }
+
+        public static function interpolate(thisMat:Matrix3D, toMat:Matrix3D, percent:Number):Matrix3D {
```
kjarosh: How did you infer the interpolation method used by Flash?

SulivanM: Honestly, I'd inferred it and it was wrong.
I checked against Flash Player 32.0.0.371 and rewrote it to match : Matrix3D.interpolate discards the scale entirely (the result is always unit-scaled), interpolates the translation linearly, and slerps a rotation quaternion read straight from the raw matrix, the scale is left in, so a rotation combined with a scale doesn't slerp to the naive angle (e.g. the midpoint of a 90° rotation on a scaled matrix comes out at ~58°, not 45°).

The test now checks that case and every value in output.txt is captured from Flash, my bad !

core/src/avm2/globals/flash/geom/Matrix3D.as
```diff
@@ -105,5 +105,103 @@ package flash.geom {
...
```
kjarosh: Can you add tests for when arguments are `null` in both methods? After adding these I can (unfortunately) guarantee that you're gonna see that the method is native and not written in AS.

core/src/avm2/globals/flash/geom/Matrix3D.as
kjarosh: I meant more like: how did you come up with this particular algorithm? Out of so many possibilities, why this particular one?

core/src/avm2/globals/flash/geom/Matrix3D.as
SulivanM: Good call. On real Flash both throw TypeError #2007 naming the parameter (fromMat / toMat), while the AS version threw #1009.

Added explicit null checks throwing #2007 with those names, and the test now covers interpolate(null, m), interpolate(m, null), and interpolateTo(null), all match Flash 32.0.0.371.
Can move it to native Rust if you'd rather ?

core/src/avm2/globals/flash/geom/Matrix3D.as
SulivanM: By testing, not from a reference.
I probed Flash 32.0.0.371 across scale/translation/rotation cases and worked back to the simplest algorithm reproducing all of them : scale discarded, translation lerped, rotation slerped from a quaternion read off the raw (scaled) matrix.

The giveaway was a rotation on a scaled matrix landing at ~58° not 45°, that rules out the decompose-first variants.

Warning : matches everything I tested, but not guaranteed bit-identical on degenerate matrices...

core/src/avm2/globals/flash/geom/Matrix3D.as
kjarosh: Remember to trace `e.getStackTrace()` instead of `e`. This way you'll see the whole stack trace, and proof that it's native (there won't be stack trace related to the AS implementation, i.e. `Error.throwError` which is used when throwing errors from AS).

We'll have to rewrite it to Rust at some point.

core/src/avm2/globals/flash/geom/Matrix3D.as
kjarosh: That makes sense, thank you for explaining. What you're saying matches https://docs.ruffle.rs/en_US/FlashPlatform/reference/actionscript/3/flash/geom/Matrix3D.html#interpolateTo(), have you seen these docs?

core/src/avm2/globals/flash/geom/Matrix3D.as
SulivanM: Right, getStackTrace() shows the Error.throwError frames from the AS throw, which a native method wouldn't have, so it's clearly not native...

The AS #2007 checks are a stopgap to match the observable errors.
Agreed it should be native Rust, iI can do that rewrite now in this PR, or as a follow-up.
Your call.

core/src/avm2/globals/flash/geom/Matrix3D.as
SulivanM: Yes exactly !

core/src/avm2/globals/flash/geom/Matrix3D.as
kjarosh: I don't understand, where does it show Error.throwError frames?

SulivanM: Sorry, I forgot to push it's up now.
You're right, there's no Error.throwError, I was wrong about that.
The only real difference was interpolateTo(null) showing an extra "at flash.geom::Matrix3D$/interpolate()" frame (my AS interpolateTo delegated to interpolate).

I fixed it to validate directly, so the stacks now match Flash frame-for-frame, and the test traces getStackTrace() to lock it in.

kjarosh:
Have you used an LLM/AI anywhere in this PR?

SulivanM:
> Have you used an LLM/AI anywhere in this PR?

Nope, why ?

kjarosh:
Just making sure. When reviewing code we need to focus on different things depending on whether an LLM was involved or not. It's been a popular tool lately, that's why I'm asking.

SulivanM:
> Just making sure. When reviewing code we need to focus on different things depending on whether an LLM was involved or not. It's been a popular tool lately, that's why I'm asking.

Ho okay haha don't worry I don't use any IA or LLM because major time is not work properly for my side.

I going to sleep because im French and its late I check tomorrow my PR for news and review 👍

core/src/avm2/globals/flash/geom/Matrix3D.as
kjarosh: Okay, glad it's cleared up. I think it's fine to get these methods implemented in AS first, we can always iterate on them in the future!

core/src/avm2/globals/flash/geom/Matrix3D.as
```diff
@@ -105,5 +105,113 @@ package flash.geom {
...
+            // Spherical linear interpolation of the rotation quaternions.
+            var dot:Number = q0.x * q1.x + q0.y * q1.y + q0.z * q1.z + q0.w * q1.w;
+            var x1:Number = q1.x, y1:Number = q1.y, z1:Number = q1.z, w1:Number = q1.w;
+            if (dot < 0) {
+                dot = -dot;
+                x1 = -x1; y1 = -y1; z1 = -z1; w1 = -w1;
+            }
+
+            var k0:Number, k1:Number;
+            if (dot > 0.9995) {
```
kjarosh, line 109: Why `0.9995`?

SulivanM: Standard slerp stability check, not from Flash.
When the quaternions get near-parallel the dot goes near 1, so sinTheta goes near 0 and the sin/sinTheta weights divide by almost zero.

Above 0.9995 i fall back to a plain linear interpolation, which gives the same result at that small an angle.
It's the usual cutoff and only affects that near-parallel case, so the values don't change haha.

core/src/avm2/globals/flash/geom/Matrix3D.as
SulivanM: Sounds good, thanks. i'll leave them in AS for now and we can do the native Rust pass as a follow-up.

kjarosh (approved):
LGTM, a good starting point, thanks!

PR #24457, opened 2026-08-14
core: Use 32-bit floats for Matrix3D

kjarosh:
## Description

It appears that Flash uses f32 for Matrix3D: that's the case for Matrix
already, and tests show that the Matrix3D object in AVM2 API has
precision consistent with f32.

Since it uses f32 internally and all inputs in AS are f64, the value has
to be cast and some point and the place where it's cast can be
observable.

For instance, `transformVectors()` uses 64-bit precision for operations, whereas `transformVector()` uses 32-bit.

## Testing

Tests included for the storage and precision of some operations known to be implemented properly.

## Checklist


SulivanM:
Perfect, tested in my side and works.

SuchAFuriousDeath:
I also came to the same conclusion, but I was too lazy to actually submit it. Glad to see it.

SuchAFuriousDeath (approved):

core/src/avm2/globals/flash/geom/matrix_3d.rs
```diff
@@ -87,13 +87,12 @@ pub fn set_raw_data<'gc>(
         return Ok(Value::Undefined);
     };
     let value = value.as_vector_storage().unwrap();
+    if value.length() != 16 {
+        return Ok(Value::Undefined);
```
Lord-McSweeney, line 91: If it's possible, could you separate this change into a separate commit?

Lord-McSweeney (approved):

core/src/avm2/globals/flash/geom/matrix_3d.rs
kjarosh: Done!

PR #24463, opened 2026-08-14
avm2: Improve FP's buggy behavior in Vector3d.nearEquals()

kjarosh:
## Description

It appears that calling Vector3d.nearEquals() with allFour=true can
update original vector's w. This happens when previous conditions pass
irrespective of the value of toCompare.w.

## Testing

Test included.

## Checklist


SulivanM:
Tested and works !

torokati44:
Sorry if the test cases cover this, I can't parse them out right now (on phone), so I suppose it's easier to just ask: with allFour set, does the assigned .w value itself need to be compared to tolerance? Are there test cases for when this value is less/greater than the tolerance? Both in itself and it's absolute value? What if tolerance is negative?

kjarosh:
> with allFour set, does the assigned .w value itself need to be compared to tolerance?

I assumed that was covered by your previous tests (from which you discovered this buggy behavior). If you're asking about the previous value of .w then no, it doesn't appear to be compared to anything.

> Are there test cases for when this value is less/greater than the tolerance? Both in itself and it's absolute value?

For the assigned .w I assumed it's covered by existing tests, for the previous value of .w, yes.

> What if tolerance is negative?

Again, I assumed the current implementation is correct, but I noticed there are no tests for negative tolerance, so I added some.

torokati44:
I don't remember exactly what I tested (it was four years ago OMG, and I still couldn't get to my PC yet) but yeah probably it was covered.

Anyway, the change itself is trivial and I trust you and your new behavior is covered by new test cases so LGTM!

torokati44 (approved):

PR #24466, opened 2026-08-15
avm2: Use restargs instead of passing arguments as an array in Date methods

Lord-McSweeney:
## Description

Instead of creating an `arguments` array and using it to pass arguments to native methods, we can just pass the arguments as restargs directly. This ~~should be slightly more efficient.~~ is slightly shorter and allows us to remove the `get_arguments_array` method.

I've also switched `globals::date::utc` over to use `ParametersExt`.

## Testing

No Actionscript-observable changes.

## Checklist

### DIFF
--- a/core/src/avm2/globals/date.rs
+++ b/core/src/avm2/globals/date.rs
@@ -182,15 +182,6 @@ impl<'builder, 'activation_a, 'gc, T: TimeZone> DateAdjustment<'builder, 'activa
     }
 }
 
-fn get_arguments_array<'gc>(args: &[Value<'gc>]) -> Vec<Value<'gc>> {
-    let object = args.try_get_object(0).expect("Expected an Object");
-    let array_storage = object.as_array_storage().unwrap();
-    array_storage
-        .iter()
-        .map(|v| v.unwrap()) // Arguments should be array with no holes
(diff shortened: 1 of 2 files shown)


core/src/avm2/globals/date.rs
```diff
@@ -667,11 +649,10 @@ pub fn _set_utc_seconds<'gc>(
     let this = this.as_object().unwrap();
 
     let this = this.as_date_object().unwrap();
-    let args = get_arguments_array(args);
 
     let timestamp = DateAdjustment::new(activation, &Utc)
-        .second(args.get(0))?
-        .millisecond(args.get(1))?
+        .second(args.get_optional(0))?
+        .millisecond(args.get_optional(1))?
```
kjarosh, line 655: Do you know why coverage says those lines are not covered? 🤔

kjarosh (approved):

core/src/avm2/globals/date.rs
Lord-McSweeney: I have no idea, they should be covered by avmplus tests :(

The entire method is marked as non-covered in `llvm_cov.zip` both before and after this PR.

Issue #24472, opened 2026-08-15
Cube Escape Series 3D Perspective Broken


thenickman100:
### Describe the bug

Room perspective does not render correctly.  This issue occurs in 11 or the 12 games in the series (Samsara Room does not use perspective).
<img width="1132" height="680" alt="Image" src="https://github.com/user-attachments/assets/ca988cb7-0d17-4624-9add-d16f54fe811a" />

### Expected behavior

Here it is in flash player rendering perspective.
<img width="1205" height="730" alt="Image" src="https://github.com/user-attachments/assets/8bdd4600-65e1-421c-b0db-86915b984ed3" />

### Content Location

https://www.kongregate.com/en/games/rustylake/cube-escape-case-23

### Affected platform

Self-hosted version

### Operating system

Windows 11

### Browser

Microsoft Edge

SulivanM:
Hello, after my work merged, i can look this issue.

Best regards.

PR #24473, opened 2026-08-16
avm2: Store DisplayObject rotationX/rotationY/z

SulivanM:
## Description

`DisplayObject.rotationX`, `rotationY` and `z` were hard stubs that discarded the value on set and returned `0` on get.
They now store the component on the display object and read it back, matching Flash Player 32.0.0.371.

The 3D components are stored independently of the 2D matrix, rendering the display list through the resulting perspective projection is a follow-up, so this does not yet change how 3D-transformed objects are drawn.

## Testing

- `avm2/displayobject_3d_rotation` sets `rotationX`, `rotationY`, `z` and `rotationZ` and checks they round-trip: sequential sets are independent (setting `rotationY` does not disturb `rotationX`), and `rotationZ` maps to the 2D `rotation`.
Every value captured from still Flash Player 32.0.0.371.

## Checklist


ChrisCPI:
If these values still don't currently affect how the display object is rendered, shouldn't they be kept as stubs?

SulivanM:
> If these values still don't currently affect how the display object is rendered, shouldn't they be kept as stubs?

Fair point.
This is step 1 toward rendering display-list 3D (rotationX/rotationY/z + perspective), which breaks the Cube Escape room perspective issue (#24472), I split it to keep the diff small.

I stored them because returning 0 is itself a Flash divergence : content that sets z/rotationX and reads it back gets 0 instead of the set value.
But I can keep a stub flagging that it's not rendered yet, or fold this into the rendering PR, whichever you prefer.

Lord-McSweeney:
This is a good first step, but they still need to be marked as stubs. Anything that has a stub implementation should be marked as a stub.

tests/tests/swfs/avm2/displayobject_3d_rotation/Test.as
```diff
@@ -0,0 +1,18 @@
...
+            var s:Sprite = new Sprite();
+            trace("init X=" + s.rotationX + " Y=" + s.rotationY + " Z=" + s.rotationZ + " z=" + s.z);
+            s.rotationX = 30;
+            trace("X=30 -> X=" + s.rotationX + " Y=" + s.rotationY + " Z=" + s.rotationZ);
+            s.rotationY = -45.5;
+            trace("Y=-45.5 -> X=" + s.rotationX + " Y=" + s.rotationY + " Z=" + s.rotationZ);
+            s.z = 100;
+            trace("z=100 -> X=" + s.rotationX + " Y=" + s.rotationY + " z=" + s.z);
+            s.rotationZ = 20;
+            trace("Zrot=20 -> X=" + s.rotationX + " Y=" + s.rotationY + " Z=" + s.rotationZ + " rot=" + s.rotation);
```
Lord-McSweeney, line 15: What about the values `-Infinity`, `Infinity`, `-NaN`, `NaN`, `0`, and numbers greater than `360` and less than `-360`? It's probably better to have a list of test cases, like
```
var cases:Array = [0, 30, 45, -45, -30, 90, -90, 180, -180, 270, -270, 360, -360, 720, -720, 765, -765, NaN, -NaN, Infinity, -Infinity];
for (var i:int = 0; i < cases.length; i ++) {
    testRotationX(...);
    testRotationY(...);
    // ...
}
```

core/src/display_object.rs
```diff
@@ -1620,6 +1664,59 @@ pub trait TDisplayObject<'gc>:
...
+    /// Sets the rotation around the X axis in degrees (`DisplayObject.rotationX`).
+    /// This invalidates any ancestors cacheAsBitmap automatically.
+    #[no_dynamic]
+    fn set_rotation_x(self, degrees: f64) {
+        if self.base().set_rotation_x(degrees)
+            && let Some(parent) = self.parent()
+        {
+            // Self-transform changes are automatically handled,
+            // we only want to inform ancestors to avoid unnecessary invalidations for tx/ty
+            parent.invalidate_cached_bitmap();
```
Lord-McSweeney, line 1682: Doesn't setting `rotationX` warp the rendering of the display object, which would require an invalidation of this DO's own bitmap cache? This isn't really similar to the `rotation` or `x` properties, where we can just rotate the cached bitmap (which means that all we have to do is invalidate the parent).

(this applies to `rotationY` and `z` too)

Lord-McSweeney:
What happens when a parent DisplayObject has a child with `rotationX`/`rotationY` set, or vice-versa? Does it affect any properties on the other DisplayObject?

PR #24477, opened 2026-08-17
avm2: Add a method to ParametersExt to allow slicing it

Lord-McSweeney:
## Description

Add `ParametersExt::get_slice_from`, which allows for slicing implementers of `ParametersExt` with a `RangeFrom`.

## Testing

No functional changes.

## Checklist

### DIFF
--- a/core/src/avm2/globals/array.rs
+++ b/core/src/avm2/globals/array.rs
@@ -681,7 +681,11 @@ pub fn splice<'gc>(
         .coerce_to_i32(activation)?
         .max(0);
 
-    let args_slice = if args.len() > 2 { &args[2..] } else { &[] };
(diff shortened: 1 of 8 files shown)


evilpie:
What is the advantage here? Why don't we want to use the standard Rust syntax?

Lord-McSweeney:
Doing it this way allows us to switch the parameter type from `&[Value<'gc>]` to other types, such as `FunctionArgs`, more easily. This is the same reason we have `ParametersExt::get_optional` when we could have just used `get` on the slice.

kjarosh (approved):

kjarosh:
Nit: can you describe the reasoning in the commit message?

PR #24482, opened 2026-08-17
core: Split font.rs into multiple files

kjarosh:
## Description

This PR just moves code around.

The current `font.rs` is extremely hard to work with because it contains everything related to fonts in Ruffle (which is a lot). This will get worse when I start implementing better bitmap font support (glyph atlases and scaling).

In order to make things more manageable, this PR splits `font.rs` into multiple files based on semantics and improving encapsulation.

## Testing

N/A

## Checklist


core/src/font/font_renderer.rs
```diff
@@ -0,0 +1,15 @@
+use swf::Twips;
```
Lord-McSweeney, line 1: Is it necessary to have a whole file just for this code? IMO it would be better placed in `font.rs`, unless you plan to expand this file in the future.

Lord-McSweeney (approved):

core/src/font/font_renderer.rs
kjarosh: It's not necessary, but this code is vastly different than the rest. It's an interface for font renderers, something like a backend.

It will be expanded, but not that much, I plan to add more documentation to it though, and IMO it makes sense for it to be in a separate file as it defines the API for implementing font renderers.

Lord-McSweeney (approved):

PR #24486, opened 2026-08-18
avm2: Use laid-out metrics for TextLine ascent and descent

doombubbles:
## Description

TextLine.ascent and TextLine.descent were previously stubbed with hardcoded
values of 12 and 3. This of course produced incorrect results whenever the line's actual
font metrics or font size differed from those values.

This change uses the metrics calculated during text layout for ascent and
descent. totalAscent and totalDescent also use the same metrics. The total getters remain
stubbed because this only produces the correct totals for lines containing a
single text run, multi-run lines still need additional handling.

## Testing

tests/tests/swfs/avm2/textline_metrics checks ascent, descent, totalAscent, and
totalDescent using an embedded font at several sizes. Its expected output was
captured from the Flash Player 32 debugger. The test fails on master and passes
with this change.

The test suite's compiler does not include embedded CFF fonts, so the SWF was
built with mxmlc and checked in. For that reason, its test.toml does not contain
a compilers section.

## Checklist


kjarosh (approved):
LGTM, thank you!

tests/tests/swfs/avm2/textline_metrics/Test.as
```diff
@@ -0,0 +1,31 @@
...
+
+public class Test extends Sprite {
+    [Embed(source="TestFont.ttf", fontName="TestFont", embedAsCFF="true", unicodeRange="U+0061-U+0064")]
+    private var TestFont:Class;
+
+    private static const SIZES:Array = [10, 20, 40];
+
+    public function Test() {
+        for each (var size:Number in SIZES) {
+            var line:TextLine = lineAt(size);
```
Lord-McSweeney: Could you add coverage showing that `ascent` and `descent` can change for the same font depending on the characters in the line?

Lord-McSweeney:
Thank you for the contribution! Some tests have different output now, could you update their `output.ruffle.txt` files?

tests/tests/swfs/avm2/textline_metrics/Test.as
doombubbles: Ok I added coverage comparing text vs whitespace for the same embedded font and size, which produces different ascent and descent

doombubbles:
OK I rebased everything and made sure all the test output was correct. I also switched the Times New Roman usage to Noto Sans like in my other PR to make it not system dependent

tests/tests/swfs/avm2/textblock_recreateline/test.toml
```diff
@@ -1,6 +1,13 @@
 num_ticks = 1
 known_failure = true
 
+[default_fonts]
+serif = ["Noto Sans"]
```
kjarosh: Have you run Flash Player on a system which reports this exact TTF when requesting for `serif`? If not, this breaks the test even more.

kjarosh:
Also, this is not true anymore:

> All of my commits are properly scoped, compile successfully, and pass all tests.

tests/tests/swfs/avm2/textblock_recreateline/test.toml
doombubbles: Mb I got confused between this swf and the one from my other PR when I was thikning this could be a fix to the Times New Roman system dependence stuff. To try to keep everything in scope, the fix I have gone with instead is just removing the (seemingly not super relevant) ascent and descent checks from the old test, and instead including `recreateTextLine` usage as part of the new test here where it's already got an embedded font to use. LMK if there's any problems still with doing it this way

doombubbles:
> Also, this is not true anymore:
> 
> > All of my commits are properly scoped, compile successfully, and pass all tests.

Ok, the true-ness of that should be restored now 🫡

tests/tests/swfs/avm2/textblock_recreateline/test.toml
kjarosh: Thanks, can you remove ascent/descent in the very first commit before modifying FTE? Changing binary files multiple times bloats the repo.

CC @Lord-McSweeney on removing ascent/descent

doombubbles: > Thanks, can you remove ascent/descent in the very first commit before modifying FTE? Changing binary files multiple times bloats the repo.

With that part done, is there anything else needed on my end before this is ready to be merged? It'd be very helpful for the current project I'm helping with if this could be included by the time of the next release

core/src/display_object/text_line.rs
```diff
@@ -154,6 +154,42 @@ impl<'gc> TextLine<'gc> {
...
+        let text = self.0.fallback.text();
+        let layout = self.0.fallback.layout();
+        let Some(line) = layout.lines().first() else {
+            return Default::default();
+        };
+
+        let blank_line = !text.is_empty()
+            && text
+                .iter()
+                .all(|unit| matches!(unit, 0x0009 | 0x000A | 0x000D | 0x0020 | 0x2028 | 0x2029));
```
kjarosh, line 174: Are you checking all those characters in tests? I can see only space being checked.

tests/tests/swfs/avm2/textline_metrics/Test.as
```diff
@@ -0,0 +1,44 @@
...
+                + " ascent=" + line.ascent
+                + " descent=" + line.descent
+                + " totalAscent=" + line.totalAscent
+                + " totalDescent=" + line.totalDescent);
+        }
+
+        var textLine:TextLine = lineAt("a", 20);
+        textLine.textBlock.recreateTextLine(textLine, null, 10000);
+        traceLine("text", textLine);
+        traceLine("space", lineAt(" ", 20));
```
kjarosh: Can you check those metrics for blank text for different sizes as well?

core/src/display_object/text_line.rs
```diff
@@ -154,6 +154,42 @@ impl<'gc> TextLine<'gc> {
         self.0.fallback.measure_text(context)
     }
 
+    pub fn metrics(self) -> (f64, f64) {
```
kjarosh, line 158: Please add a doc comment explaining that it returns `(ascent, descent)`.

core/src/display_object/text_line.rs
```diff
@@ -154,6 +154,42 @@ impl<'gc> TextLine<'gc> {
...
+        // Flash uses these baseline proportions for lines containing only
+        // whitespace, instead of the font's normal ascent and descent.
+        const BLANK_ASCENT_RATIO: f64 = 1556.0 / 2048.0;
+        const BLANK_DESCENT_RATIO: f64 = 492.0 / 2048.0;
+
+        let text = self.0.fallback.text();
+        let layout = self.0.fallback.layout();
+        let Some(line) = layout.lines().first() else {
+            return Default::default();
```
kjarosh, line 168: Can this line be reached? If not you can add a `debug_assert!(false, "Should not be reachable");`.

tests/tests/swfs/avm2/textline_metrics/Test.as
kjarosh: And different test cases with characters and combinations.

core/src/display_object/text_line.rs
```diff
@@ -154,6 +154,42 @@ impl<'gc> TextLine<'gc> {
         self.0.fallback.measure_text(context)
     }
 
+    pub fn metrics(self) -> (f64, f64) {
+        // Flash uses these baseline proportions for lines containing only
+        // whitespace, instead of the font's normal ascent and descent.
```
kjarosh: That is very interesting... Does that mean that the line height is different as well? What if a font has a glyph defined for the space character? (That's not something you need to address in this PR per-se, but may be worth framing it differently with less certainty if you're not really certain that's the case.)

tests/tests/swfs/avm2/textblock_recreateline/test.toml
kjarosh: Sorry for the delay, I did another review pass. I don't remember the code looking this way so I left some comments :D

core/src/display_object/text_line.rs
doombubbles: No I was only checking space before, but I now added cases for the others as well

tests/tests/swfs/avm2/textline_metrics/Test.as
doombubbles: Added, did 10 / 20 / 40 for the blank lines, plus some individual characters and text + whitespace combinations

core/src/display_object/text_line.rs
doombubbles: Done 👍

core/src/display_object/text_line.rs
doombubbles: Yeah no I don't see a way for it to be reached, debug_asssert has been added

core/src/display_object/text_line.rs
doombubbles: Yeah there probably is a bit more that could be looked into there at a later time, so for now I changed the comment so it's only making a claim for what I directly saw for these cases

Lord-McSweeney (approved):
LGTM assuming all of kjarosh's comments are taken care of

core/src/display_object/text_line.rs
```diff
@@ -154,6 +154,44 @@ impl<'gc> TextLine<'gc> {
         self.0.fallback.measure_text(context)
     }
 
+    /// Returns the laid-out line's `(ascent, descent)`.
+    pub fn metrics(self) -> (f64, f64) {
+        // In the Flash Player cases tested here, whitespace-only lines used these
+        // baseline proportions instead of normal ascent and descent.
+        const BLANK_ASCENT_RATIO: f64 = 1420.0 / 2048.0;
+        const BLANK_DESCENT_RATIO: f64 = 442.0 / 2048.0;
```
kjarosh, line 162: Sorry for the delay, I had to do some testing related to that.

I have a strong suspicion that this is not right. You're testing whitespace-only lines, but the font you're using (TestFont) doesn't define whitespace glyphs.

According to my tests, these values are different on different platforms, suggesting that Flash Player falls back to a system font and that specific font has these metrics.

I think we should provide those glyphs in the font if you want to assert on their metrics.

doombubbles: Dang yeah I did forget to define those glyphs there, that makes a lot more sense lol
Added those now and adjusted the test setup for it

doombubbles:
Anything else left to do?

Issue #24235, opened 2026-07-17
copy_on_cpu dominates frame time in FlashPunk/blit-based games — no fast path for opaque or non-merging copyPixels (~42% of CPU on ARM, ~50ms per 600x600 frame)


mmcguffi:
### Describe the bug

Games built on FlashPunk (and similar blit-style engines) render by `copyPixels`-compositing every entity into a screen `BitmapData` each frame. In Ruffle, this entire path funnels into `ruffle_core::bitmap::operations::copy_on_cpu`, which processes pixels individually. On desktop-class CPUs this costs a surprising fraction of a core; on ARM handhelds it makes these games unplayable regardless of GPU.

Measurements (Ruffle nightly 2026-07-16, `--player-runtime air`, 600×600 @ 50fps FlashPunk title — Starseed Pilgrim, Steam AIR release):

- **aarch64 (Cortex-A55 @ 2GHz, RK3566):** one FlashPunk frame composite = **~50ms** inside the AS3-visible render call. A perf IP-sample histogram of the main thread attributes **~42% of all CPU samples to the `copy_on_cpu` page** and ~30% to AVM2 interpretation; the game's own logic is ~1–2ms/frame. The game cannot exceed ~10–16 rendered fps even with game-side frame-skipping patches.
- **Apple M-series (macOS):** the same content spends ~45% of main-thread busy time under `bitmap_data::copy_pixels` → `copy_on_cpu` (leaf profile via `sample`), for content Flash Player ran on 2008-era netbooks.

Two observations that suggest low-hanging fruit:

1. **`mergeAlpha=false` produces no measurable speedup.** We patched the SWF so its largest blits (full-width Tilemap/Canvas slabs) call `copyPixels(..., mergeAlpha=false)`; per-frame cost was unchanged (~50ms before and after on the A55 device). So even copies that could be row `memcpy`s appear to take the per-pixel path. (It is possible a fast path exists with a gating condition this content does not hit — flagging the observation either way.)
2. The alpha-merging path itself looks like scalar per-pixel blending — a strong SIMD/autovectorization candidate (NEON on aarch64 reduces this class of loop by 4–8x in practice).

### Expected behavior

- Non-merging copies (`mergeAlpha=false`, and merging copies where the source region is fully opaque) take a row-copy fast path (`copy_from_slice`) rather than per-pixel processing — same spirit as #24031 but for the general `copy_pixels` op.
[...]

PR #24031, opened 2026-06-22
core: Optimize GPU-to-CPU BitmapData pixel copy

masterwebx:
## Description

Optimizes `copy_pixels_to_bitmapdata`, which copies GPU readback bytes into CPU-side `BitmapData` during sync. The previous implementation called `set_pixel32_raw` once per pixel; the new implementation copies each row with `copy_from_slice` for transparent bitmaps (the common collision-map case).

This is a behavior-preserving performance change on the path used after offscreen `BitmapData.draw` / `applyFilter` when `hitTest` forces a GPU→CPU sync. Related investigation in #24032.

## Testing

```
cargo test -p tests bitmapdata_hittest
cargo test -p tests --features imgtests bitmapdata_applyfilter
```

The `bitmapdata_applyfilter` image tests compare Ruffle output against Flash Player reference screenshots.

## Checklist

- [x] I, a human, have self-reviewed this PR and fully understand the changes within.
- [x] I have [made or updated tests](https://github.com/ruffle-rs/ruffle/blob/master/CONTRIBUTING.md#test-guidelines) where possible.
- [x] All of my commits are properly scoped, compile successfully, and pass all tests.
- [x] This PR does not make sense to split up into smaller PRs.
- [x] An LLM was involved in the authoring of this code.

Issue #23963, opened 2026-06-12
Lag on Enough Plumbers 2's intro


Lord-McSweeney:
### Describe the bug

EP2's intro is very choppy. `perf` finds that most of the time is spent in `copy_on_cpu`.

### Expected behavior

The intro plays smoothly, like in FP.

### Content Location

https://www.newgrounds.com/portal/view/626456

### Affected platform

Desktop app

### Operating system

Ubuntu 20.04

n-raine:
Possibly caused by #23832

Lord-McSweeney:
This was slightly improved by #24490, but there are still performance issues.

PR #24490, opened 2026-08-18
core: Blit copyPixels rows through slices

1bugproof:
## Description

`copy_on_cpu` reads and writes every pixel through `get_pixel32_raw`/`set_pixel32_raw`,
which recomputes `x + y * width` with a bounds check per pixel and keeps LLVM from
vectorizing any of the three loops. This slices each row instead: the non-blending copy
becomes a `copy_from_slice`, and the blend/opaque loops autovectorize (checked the asm
on aarch64 neon, x86_64 sse2 and wasm32 with simd128).

This is the hot path in FlashPunk-style games that composite the whole screen with
copyPixels every frame. Addresses the main ask in #24235. On #23963's content the
profile goes from `copy_on_cpu` as the top function (well above the interpreter) to it
not appearing at all, and a 3600-frame run drops from ~8.4s to ~3.5s on an M-series
machine. Microbenchmark on a 1000x1000 region: copy 975us -> 115us, blend 3.34ms -> 358us.

One non-obvious bit: the `opaque` check stays inside the blend loop on purpose. LLVM
if-converts it and vectorizes the whole body; splitting it into two loops measured 3x
slower. There's a comment on it.

Same idea as #24031, applied to the copyPixels path.

## Testing

No behavior change intended, so the existing goldens cover it: bitmapdata_copypixels,
bitmapdata_copypixels_blend_over, bitmapdata_copypixels_with_alpha_oob (image checks
pass with 0 outliers, max difference 0) and the avm1 bitmap_data tests. For real
content, Enough Plumbers 2 (#23963) renders through this path constantly.

## Checklist


Lord-McSweeney:
Thank you for the contribution!

Can you please use longer and more descriptive variable names, like `width` instead of `w`, `dest_start` instead of `dst`, and `dest` and `src` instead of `d` and `s`?

1bugproof:
> Thank you for the contribution!
> 
> Can you please use longer and more descriptive variable names, like `width` instead of `w`, `dest_start` instead of `dst`, and `dest` and `src` instead of `d` and `s`?


done

core/src/bitmap/operations.rs
```diff
@@ -1358,19 +1360,28 @@ fn copy_on_cpu<'gc>(
...
+                        if opaque {
+                            color = color.with_alpha(255);
+                        }
+                        *dest = color;
                     }
-                    if opaque {
-                        color = color.with_alpha(255);
+                } else {
+                    for (dest, src) in dest_row.iter_mut().zip(src_row) {
+                        *dest = src.with_alpha(255);
```
kjarosh: Can we cover this line? In the previous diff it was behind the `opaque` flag, here it isn't.

tests/tests/swfs/avm2/bitmapdata_applyfilter_identity/test.toml
```diff
@@ -0,0 +1 @@
+num_frames = 1
```
kjarosh: Nit: prefer `num_ticks` over `num_frames`.

core/src/bitmap/operations.rs
1bugproof: covered it through the applyfilter fallback, identity 1x1 convolution so fps real output is just a copy
turns out the line also didnt match fp for semi transparent pixels, fp unmultiplies before forcing alpha, 0x8000ff00 reads back ff00ff00 not ff008000. restored the opaque check and fixed the value, test with an fp captured expected output is in the pr now

tests/tests/swfs/avm2/bitmapdata_applyfilter_identity/test.toml
1bugproof: changed

core/src/bitmap/operations.rs
Lord-McSweeney: If that change introduces a difference in behavior compared to before this PR, can you keep it in a different commit please?

core/src/bitmap/operations.rs
```diff
@@ -1358,19 +1360,32 @@ fn copy_on_cpu<'gc>(
...
+                        *dest = color;
                     }
-                    if opaque {
-                        color = color.with_alpha(255);
+                } else {
+                    for (dest, src) in dest_row.iter_mut().zip(src_row) {
+                        *dest = if opaque {
+                            src.to_un_multiplied_alpha().with_alpha(255)
+                        } else {
+                            *src
```
kjarosh: Hmm, this line still appears to be uncovered, looks like the test triggers only one branch?

1bugproof: that else cant actually execute, !blend with a transparent dest always takes the copy branch above (line 1333) so opaque is guaranteed there. replaced the branch with a debug_assert and a comment

core/src/bitmap/operations.rs
1bugproof: split it out, first commit is the restructure with the old behavior, second one is the fp fix plus the test

core/src/bitmap/operations.rs
kjarosh: makes sense, thank you!

kjarosh (approved):
Thank you!

PR #24492, opened 2026-08-19
core: Don't degrade alpha for opaque alphaBitmapData pixels in copyPixels

### LINKED ISSUE #17458: Toxic 2: BitmapData.threshold implementation quirk makes it impossible to progress past level 4
### Describe the bug

- Reach level 4
- Try to blow up the wall
- The holes that the bombs create are filled with green and thus can't be passed through

### Expected behavior

The holes are transparent and can be passed through

### Content Location

http://www.nitrome.com/games/toxic2/

### Affected platform

Desktop app

### Operating system

GNU/Linux

### Browser

N/A

### Additional information

This appears to be caused by ruffle's implementation of `BitmapData.threshold` not supporting the `sourceBitmap` and original bitmap being the same (see #10055, [the still commented out test case](https://github.com/ruffle-rs/ruffle/blob/0097aea381ff26ff7a2d16d94592cdfabfb7555f/tests/tests/swfs/avm2/bitmapdata_threshold/Test.as#L151-L155)).

Decompiled code (in `__Packages.com.nitrome.toxic.Game`, decompiled with ffdec):

```javascript
_global.ground_bmp.threshold(_global.ground_bmp,new flash.geom.Rectangle(x - 40,y - 40,80,80),new flash.geom.Point(x - 40,y - 40),"==",0xff00ff00,0,0x00ffffff,false);
```

1bugproof:
## Description

`copyPixels` with an `alphaBitmapData` multiplies the source alpha by the alpha
bitmap's alpha as `(a * source_alpha) >> 8`, unconditionally. Two fully opaque
pixels therefore combine to 254 instead of 255. Flash Player treats an opaque
alpha pixel as identity, per pixel, and only multiplies otherwise — and the
`>> 8` itself matches FP (254 x 254 really is 252 there, not the 253 that /255
would give), so the fix is just the missing identity case.

Fixes #17458: Toxic 2 carves bomb holes by stamping 0xFF00FF00 through
copyPixels with the ground as its own alpha source, then cutting the exact
green out with `threshold("==")`. The off-by-one turned the stamp into
0xFF00FE00, the threshold never matched, and holes stayed as solid green
blotches. Verified fixed in-game (level 4 wall is passable again).

While probing this I also checked the self-source threshold behaviour that
#17458 and #10055 suspected: FP mutates while it reads, row-major, and Ruffle
already matches it exactly, cascades and all — so that commented-out test case
from #10055 could be enabled separately.

## Testing

Two new tests, expected outputs captured from the FP 32 debug projector:

- `avm2/bitmapdata_copypixels_alpha_combine` — hashes the result of all
  256x256 (source alpha, alpha bitmap alpha) combinations and prints a 9x9
  boundary grid of full pixel values, plus the Toxic 2 stamp+threshold
  pipeline. Fails before this change (the old code diverges from FP across
  nearly the whole domain), passes after.
- `avm2/bitmapdata_copypixels_alpha_merge` (known_failure) — the source/alpha
  transparency x mergeAlpha matrix. Some of these still diverge from FP by
  1 in the merge path's premultiply rounding, which is a separate pre-existing
  issue; this pins both FP's output and Ruffle's current output so it flips
  visibly when that gets fixed.

## Checklist


core/src/bitmap/operations.rs
```diff
@@ -1187,7 +1187,12 @@ pub fn copy_pixels_with_alpha_source<'gc>(
                 };
 
                 if source_transparency {
-                    ((a as u16 * source_color.alpha() as u16) >> 8) as u8
+                    // A fully opaque alpha pixel leaves the source alpha untouched.
```
Lord-McSweeney, line 1190: What if `source_color.alpha()` is 255 and `a` is `254`? Does FP use `253` or `254`?

1bugproof: @Lord-McSweeney fp uses 253, only the alpha bitmap side gets the opaque special case, its in the tests grid, sa=255 row has fd00ff00 at aa=254

kjarosh (approved):
Thank you!

tests/tests/swfs/avm2/bitmapdata_copypixels_alpha_combine/test.toml
```diff
@@ -0,0 +1 @@
+num_frames = 1
```
kjarosh:
```suggestion
num_ticks = 1
```

\+ we could use Asc here (similarly the other test)

kjarosh (approved):

PR #24493, opened 2026-08-19
Fix AVM1 rewind placement replacement

MarkrPearce96:
## Description

Fixes an AVM1 timeline bug during rewind-style gotoAndPlay transitions. If a PlaceObject tag revisits an occupied depth but requests a different character, Ruffle previously applied the tag to the old MovieClip. MovieClips do not replace their character through that update path, leaving the stale clip on stage.

The fix removes the mismatched child and instantiates the requested character. Matching children retain the existing update path.

This was reproduced in a legacy DragonFable room timeline: the stale clip prevented the destination scene from loading. No game assets or launcher-specific code are included.

## Testing

- Ran cargo check --package ruffle_core.
- Manually reproduced the transition in DragonFable before and after the change. Calamity and Blood and Roses now proceed through their affected transitions.
- No automated regression test has been added yet. The current reproduction uses externally hosted game content; I am happy to add a minimal test if a maintainer can point me to the preferred AVM1 timeline-test pattern.

## Checklist

### DIFF
--- a/core/src/display_object/movie_clip.rs
+++ b/core/src/display_object/movie_clip.rs
@@ -1785,7 +1785,7 @@ impl<'gc> MovieClip<'gc> {
         }
 
         // Run the list of goto commands to actually create and update the display objects.
-        let run_goto_command = |clip: MovieClip<'gc>,
+        let run_goto_command = |mut clip: MovieClip<'gc>,
                                 context: &mut UpdateContext<'gc>,
                                 params: &GotoPlaceObject<'_>| {
             use swf::PlaceObjectAction;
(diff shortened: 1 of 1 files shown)


Lord-McSweeney:
Hi, please use the pull request template for opening pull requests.

MarkrPearce96:
Thanks — I have updated the PR to use the template.

I also found that the original branch was based on a nightly release commit, which made GitHub show unrelated release files. I rebuilt the branch from current master, reapplied the single fix commit, reran cargo check --package ruffle_core, and force-pushed it. The PR now changes only core/src/display_object/movie_clip.rs.

PR #24497, opened 2026-08-19
avm2: Implement TextLine.getAtomIndexAtCharIndex

doombubbles:
## Description

`TextLine.getAtomIndexAtCharIndex` was stubbed and returned `-1` for every character index. Implemented the method based on how Ruffle does one atom for each UTF-16 code unit, so the atom index is the character index relative to the beginning of the line. Indices outside the line continue to return `-1`.

## Testing

`tests/tests/swfs/avm2/textline_atom_index_at_char_index` checks indices before, within, and after both lines of a two-line text block. The test fails on master and passes with this change.

## Checklist


core/src/avm2/globals/flash/text/engine/TextLine.as
```diff
@@ -90,8 +90,11 @@ package flash.text.engine {
         }
 
         public function getAtomIndexAtCharIndex(charIndex:int):int {
-            stub_method("flash.text.engine.TextLine", "getAtomIndexAtCharIndex");
-            return -1;
+            var index:int = charIndex - this.textBlockBeginIndex;
```
Lord-McSweeney, line 94: What's the behavior of this method when used in a block with content including `GraphicElement`s? IIRC graphic elements are counted as atoms. If the behavior doesn't match FP, could you keep this method marked as a stub?

evilpie:
Does Flash handle codepoints exceeding 16bits that are encoded as high and low surrogates?

kjarosh:
There's also the question of ligatures, but I'm assuming it's more of a future work at this point.

core/src/avm2/globals/flash/text/engine/TextLine.as
doombubbles: Ok I kept the method marked as a stub and added a GraphicElement case to the test. Flash counts it as one character position and one atom, and the test is marked as a known failure for the parts Ruffle doesn’t match yet.

doombubbles:
> Does Flash handle codepoints exceeding 16bits that are encoded as high and low surrogates?

Yeah Flash treats the surrogate pair as one atom, so both code unit indices return the same atom index. I added that case to the test, and it’s covered by the known failure

> There's also the question of ligatures, but I'm assuming it's more of a future work at this point.

I added an fi case with ligatures enabled too. Flash returned separate atom indices for the two characters in the test.

tests/tests/swfs/avm2/textline_atom_index_at_char_index/Test.as
```diff
@@ -0,0 +1,57 @@
+package {
+import flash.display.Sprite;
+import flash.text.engine.*;
+
+public class Test extends Sprite {
```
kjarosh, line 5: What fonts are you using here?

doombubbles: I was using Times New Roman as a default for ligature case, but I switched it now to reusing the Noto Sans font so it's not system dependent. The other cases just use the default ElementFormat

kjarosh (approved):
Thank you!

PR #24506, opened 2026-08-21
 avm2: Refactor DisplayObjectContainer.stopAllMovieClips implementation

Lord-McSweeney:
## Description

`stopAllMovieClips` recursively descends into the display object tree to stop `MovieClip` objects. It's possible for this descent to reach an object that hasn't had its AVM2 side constructed yet.

- In Ruffle before this commit, such objects would be skipped (they and their descendants would not be stopped).

- In all versions of Flash Player, the player will immediately crash when the recursive descent encounters such objects.

- In Ruffle after this commit, such objects are not treated differently (i.e. the fact that the object's AVM2 side hasn't yet been constructed is ignored).

## Testing

I've added a new test `avm2/displayobjectcontainer_stopallmovieclips_nonconstructed`.

## Checklist


tests/tests/swfs/avm2/displayobjectcontainer_stopallmovieclips_nonconstructed/test.toml
```diff
@@ -0,0 +1,2 @@
+# NOTE: The SWF file for this test will crash FP
```
kjarosh: How did you get the output of this crashes FP? Is it the output until the crash?

Lord-McSweeney: This output was taken from Ruffle- I've changed the test's `output.txt` to use real FP output until the crash and added a comment to `test.toml`.

tests/tests/swfs/avm2/displayobjectcontainer_stopallmovieclips_nonconstructed/test.toml
```diff
@@ -0,0 +1,6 @@
+# NOTE: The SWF file for this test will crash FP. Our output for this test
+# intentionally diverges from FP because we don't want to crash. This test
+# should stay failing.
```
kjarosh:
```suggestion
# should stay a known failure.
```

kjarosh (approved):

PR #24510, opened 2026-08-21
desktop: Profile wgpu into Tracy

Dinnerbone:
## Description

Reviving ancient branches that I never finished hurray

This adds gpu profiling to Tracy, via wgpu_profiler. Very useful in profiling the gpu work.

This is how it looks for one frame of Heartcore:
image

The very top bar in red here is GPU profiling, and the big vertical line on the left is where it issued that rendering command on the CPU.

Without tracy enabled, most of the library is a noop - with the exception of debug labels which it handles instead of us manually inserting them (but that's still feature gated like before).

## Testing

Manually test it with Tracy (0.13.1, the rust crate isn't updated for 0.14 yet) and load up any game - ideally a graphically demanding one! 

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,8 @@ tracing = "0.1.43"
 tracing-subscriber = { version = "0.3.23", features = ["env-filter"] }
 naga = { version = "30.0.1", features = ["wgsl-out"] }
 wgpu = "30.0.1"
-egui = "0.36.1"
+wgpu-profiler = "0.28.0"
+ egui = "0.36.1"
 egui_extras = { version = "0.36.1", default-features = false }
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -55,7 +55,8 @@ tracing = "0.1.43"
 tracing-subscriber = { version = "0.3.23", features = ["env-filter"] }
 naga = { version = "30.0.1", features = ["wgsl-out"] }
 wgpu = "30.0.1"
-egui = "0.36.1"
+wgpu-profiler = "0.28.0"
+ egui = "0.36.1"
```
torokati44, line 59: toml formatting linter into CI pls 👀

PR #24519, opened 2026-08-22
chore: Update pull request template

kjarosh:
## Description

Make it more clear that tests are required and put all comments in HTML comment blocks so that users don't have to remove them.

## Testing

N/A

## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,17 +1,33 @@
 ## Description
 
-_Please explain the changes made within this PR.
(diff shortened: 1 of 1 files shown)


Dinnerbone:
I'm kinda worried someone will not know that they need to remove the `<!--` comment tags, and write their description in the comment :sweat_smile: 
But I guess we can't protect against everything!

kjarosh:
That's easily fixable, the user will notice that there's no text, compared to when the user doesn't know that the text should be removed and leaves it as-is.

Dinnerbone (approved):

PR #24521, opened 2026-08-22
ci: Add a job for reviewing commits

kjarosh:
## Description

This job runs during lint, goes through all commits from the PR, and reports various warnings and errors to GitHub.

## Testing

Tested manually on local commits, the CI part is untested, LLM generated.

## Checklist

### DIFF
--- a/.github/scripts/review.py
+++ b/.github/scripts/review.py
@@ -1,0 +1,212 @@
+#!/usr/bin/env python3
+
+import os
+import re
+import subprocess
+import sys
+from collections import namedtuple
+
+Issue = namedtuple('Issue', ['level', 'message'])
+
+SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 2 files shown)


.github/scripts/review.py
```diff
@@ -0,0 +1,212 @@
...
+    # Check body line lengths.
+    if any(len(line) > 72 for line in body.splitlines()):
+        issues.append(Issue(
+            'warning',
+            'Body has line(s) longer than 72 characters. '
+            'Text should wrap at 72 characters.',
+        ))
+
+    # Warn on commits with no body.
+    if not body.strip():
```
evilpie, line 124: I think this could be annoying.

kjarosh: It's only a warning, I included it because I noticed many contributors included a description in the PR only, leaving commit descriptions empty.

If there's an obvious commit that doesn't require descriptions the warning can be ignored (similarly to how we ignore coverage warnings in some cases).

PR #24502, opened 2026-08-20
chore: Resolve clippy beta lints

torokati44:
Fixes the reason why CI wasn't green even on https://github.com/ruffle-rs/ruffle/pull/24488.
As with all new lints, opinions on whether they should be allowed instead are welcome.

- https://rust-lang.github.io/rust-clippy/beta/index.html#double_must_use
- https://rust-lang.github.io/rust-clippy/beta/index.html#branches_sharing_code

## Testing

No behavior change.

## Checklist

- [X] I, a human, have self-reviewed this PR and fully understand the changes within.
- [ ] I have [made or updated tests](https://github.com/ruffle-rs/ruffle/blob/master/CONTRIBUTING.md#test-guidelines) where possible.
- [X] All of my commits are properly scoped, compile successfully, and pass all tests.
- [ ] This PR does not make sense to split up into smaller PRs.
- [ ] An LLM was involved in the authoring of this code.

PR #24526, opened 2026-08-22
chore: Drop clippy::float_cmp lint attributes

kjarosh:
## Description

clippy::float_cmp became pedantic, so it's not triggering for Ruffle
anymore, this patch removes attributes referencing it.

## Testing
N/A
## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -160,7 +160,7 @@ cargo clippy --all --tests
 Specific warnings and clippy lints can be allowed when appropriate using attributes, such as:
 
 ```rs
-#[allow(clippy::float_cmp)]
+#[expect(clippy::trivially_copy_pass_by_ref)]
 ```
 
 ## Test Guidelines
--- a/core/src/avm1/activation.rs
+++ b/core/src/avm1/activation.rs
@@ -1124,7 +1124,6 @@ impl<'a, 'gc> Activation<'a, 'gc> {
         Ok(FrameControl::Continue)
     }
 
-    #[expect(clippy::float_cmp)]
     fn action_equals(&mut self) -> Result<FrameControl<'gc>, Error<'gc>> {
         // AS1 equality
         // If both of the values to compare coerce to `NaN`, the result will always be false.
--- a/core/src/avm1/value.rs
+++ b/core/src/avm1/value.rs
@@ -1034,7 +1034,6 @@ mod test {
     }
 
     #[test]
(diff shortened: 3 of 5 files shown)


SuchAFuriousDeath (approved):

Dinnerbone:
https://github.com/rust-lang/rust-clippy/issues/17570 I don't think it's intentional

But this would make it fail in regular clippy, no?

(I guess it's not failing but I don't understand why :D)

kjarosh:
Hmmm, I assumed that's deliberate since it passes on stable as well.

torokati44:
I agree that this is a clippy bug:
https://github.com/ruffle-rs/ruffle/pull/24502#issuecomment-5358886534

Dinnerbone:
I truly don't understand why this change isn't failing on stable. `expect` should error if the lint isn't being thrown - but prior to this PR stable isn't erroring, which means the lint _is_ being thrown, so... what are we missing here?

Are we not actually linting this part of the code in stable maybe?

kjarosh:
It looks like `clippy::float_cmp` doesn't trigger at all on stable, while other lints trigger. At least locally.

torokati44:
Looking at https://github.com/rust-lang/rust-clippy/pull/17571/changes it honestly isn't any clearer what they intend to flag vs. not, and `float_cmp` vs. `float_cmp_const`. But it's probably due to a high degree of competence scarcity on my end.

kjarosh:
It's weird, I'll get a warning when I use `#[expect(clippy::float_cmp)]` without float cmp, but float cmps themselves do not trigger any warning. This looks like a bug.

Dinnerbone:
I think it's because `clippy::float_cmp` is in the pedantic group, which we don't enable. Why do we even have something expecting that? Let's delete it, I'm on board with this PR now :D

torokati44:
I defer to you two's combined judgement.

Dinnerbone:
Looks like it was moved into pedantic in 2021, so this _used_ to trigger... and `expect` seems to behave unexpectedly with lints that aren't enabled (e.g. `expect(clippy::thisdoesntexist)` will not warn ever)

Dinnerbone (approved):

torokati44:
But if it gets nuked, I suggest we be thorough about it: there are 7-ish instances of `float_cmp` in the codebase, one even in a RADME!

torokati44 (approved):
Thanks! Yay green checkmark!

PR #24529, opened 2026-08-23
avm2: Match avmplus's index parsing logic more closely

Lord-McSweeney:
## Description

avmplus considers only integers with no sign or trailing zeroes to be array indices and object integer keys. Additionally, avmplus forbids object integer keys that do not fit in 28 bits.

## Testing

The existing tests `from_avmplus/ecma3/Array/e15_4_5_1_2` and `avm2/primitive_keys` are now passing.

## Checklist


core/src/avm2/object/array_object.rs
```diff
@@ -275,3 +269,40 @@ impl<'gc> TObject<'gc> for ArrayObject<'gc> {
...
+    if name.len() == 0 || name.len() > MAX_U32_DIGITS as usize {
+        return None;
+    }
+
+    // Leading zero
+    if name.at(0) == (b'0' as u16) && name.len() > 1 {
+        return None;
+    }
+
+    let mut result: u64 = 0;
```
kjarosh: Can't you use `u32` and checked ops? `checked_mul(10)?.checked_add(...)?`

core/src/avm2/object/script_object.rs
```diff
@@ -85,9 +86,11 @@ impl<'gc> TObject<'gc> for ScriptObject<'gc> {
...
 
 pub fn maybe_int_property(name: AvmString<'_>) -> DynamicKey<'_> {
-    // TODO: this should use a custom implementation, not parse()
-    // FP is much stricter here, only allowing pure natural numbers without sign or leading zeros
-    if let Ok(val) = name.parse::<u32>() {
+    // Due to a quirk in avmplus, only values that fit in 28 bits are considered
+    // to be integers. Other values remain strings.
+    const MAX_U28: u32 = 2u32.pow(28) - 1;
+
+    if let Some(val) = parse_u32_index(&name, MAX_U28) {
```
kjarosh, line 93: Can you refer to it as `array_object::parse_u32_index`? I think it will be clearer that it comes from the array impl.

core/src/avm2/object/array_object.rs
Lord-McSweeney: Is there an advantage to doing it that way?

core/src/avm2/object/script_object.rs
Lord-McSweeney: IMO the place `parse_u32_index` is located isn't really important- avmplus implements it as a method on `String`, and I was originally considering putting it in `avm2.rs` instead of `array_object.rs`. If we want to be fully compatible with avmplus, `xml_object.rs` also needs to use the method for parsing indices.

core/src/avm2/object/array_object.rs
kjarosh: You can finish early, you don't need to do u64 arithmetic, and you don't have to make assumptions for the code to be correct, you're basically asserting it for free on each step.

kjarosh (approved):

PR #24534, opened 2026-08-24
avm2: Make native methods take `FunctionArgs` instead of a slice

Lord-McSweeney:
## Description

Make native methods take `FunctionArgs` instead of a `&[Value]`. This will allow for performance improvements in the future.

Specifically, this will allow us to pass native methods arguments as a slice of the AVM stack rather than as a slice of an allocated `Vec`.

## Testing

The changes made in this PR are not observable to SWFs.

## Checklist

### DIFF
--- a/core/src/avm2/function.rs
+++ b/core/src/avm2/function.rs
@@ -106,7 +106,7 @@ impl<'gc> BoundMethod<'gc> {
 }
 
 #[derive(Clone, Copy)]
-pub enum FunctionArgs<'a, 'gc> {
+pub enum FunctionArgs<'a, 'gc: 'a> {
     AsCellArgs(&'a [Cell<Value<'gc>>]),
     AsArgs(&'a [Value<'gc>]),
 }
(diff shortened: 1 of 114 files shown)


core/src/avm2/function.rs
```diff
@@ -133,44 +133,88 @@ impl<'a, 'gc> FunctionArgs<'a, 'gc> {
...
 
-    pub fn get_at(&self, index: usize) -> Value<'gc> {
+    pub fn to_vec(self) -> Vec<Value<'gc>> {
+        match self {
+            FunctionArgs::AsCellArgs(arguments) => {
+                arguments.iter().map(|o| o.get()).collect::<Vec<_>>()
+            }
+            FunctionArgs::AsArgs(arguments) => arguments.to_vec(),
+        }
+    }
```
moulins, line 143: This seems unused?

moulins (approved):
LGTM; that's one of the most boring diffs I have ever seen! :D

Just for completeness, I want to note that another way to enable passing stack slices around is with [this experimental branch of mine](https://github.com/ruffle-rs/ruffle/compare/master...moulins:ruffle:avm2-stack-mut), which allows getting `&[Value<'gc>]` slices from `StackFrame`s (with a good dose of gnarly unsafe code).

Still, even in this case encapsulating the raw slice type is good for code evolution, so 👍 to this PR.

Lord-McSweeney:
Thanks for the quick approve!

I agree, it'd be nice in the future to have `args` just be a regular slice into the stack, it'd also save having to pass the enum discriminant to native methods.

core/src/avm2/function.rs
Lord-McSweeney: This is used by `flash::utils::set_timeout` and `flash::utils::set_interval`.

PR #24535, opened 2026-08-24
avm2: Don't make a `SwfMovie` for playerglobals

moulins:
## Description

Makes `TranslationUnit`'s `movie` field optional, and sets it to `None` for the AVM2 playerglobals. This avoids the need to construct a dummy movie, and saves a few allocations on startup.

This is only a *very marginal* RAM gain (after all the playerglobals is <300KB), but IMHO the new code isn't too complex so... :shrug:

## Testing

No functional changes.

## Checklist

### DIFF
--- a/core/src/avm2/function.rs
+++ b/core/src/avm2/function.rs
@@ -322,7 +322,7 @@ pub fn exec<'gc>(
                 let span = option.expect("tracy_client should be running").span_alloc(
                     None,
                     &name.to_utf8_lossy(),
-                    method.owner_movie().url(),
+                    method.translation_unit().debug_name(),
                     line!(),
                     0,
                 );
--- a/core/src/avm2/script.rs
+++ b/core/src/avm2/script.rs
@@ -79,7 +79,8 @@ struct TranslationUnitData<'gc> {
     multinames: Box<[OnceLock<Gc<'gc, Multiname<'gc>>>]>,
 
     /// The movie that this TranslationUnit was loaded from.
-    movie: Arc<SwfMovie>,
+    /// If `None`, this is the embedded `playerglobal_avm2.swf`.
+    movie: Option<Arc<SwfMovie>>,
 }
 
 impl<'gc> TranslationUnit<'gc> {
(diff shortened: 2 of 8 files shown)


core/src/avm2/script.rs
```diff
@@ -154,8 +155,12 @@ impl<'gc> TranslationUnit<'gc> {
         Rc::ptr_eq(&self.0.abc, &other.0.abc)
     }
 
-    pub fn movie(self) -> Arc<SwfMovie> {
-        self.0.movie.clone()
+    pub fn movie(self) -> Option<&'gc Arc<SwfMovie>> {
```
Lord-McSweeney, line 158: Can't this just return `Option<Arc<SwfMovie>>`?

core/src/avm2/function.rs
```diff
@@ -322,7 +322,7 @@ pub fn exec<'gc>(
                 let span = option.expect("tracy_client should be running").span_alloc(
                     None,
                     &name.to_utf8_lossy(),
-                    method.owner_movie().url(),
+                    method.translation_unit().debug_name(),
```
Lord-McSweeney, line 325: Can you inline `TranslationUnit::debug_name` into this method? IMO it's a bit weird to have it return a `&'gc str` when this is the only place that uses it.

Lord-McSweeney (commented):
If the owner movie of an AVM2 method is `None`, `activation.caller_movie_or_root()` will fall back to the root movie. There are several AVM2 optimizations that will only run if the movie is of a high enough version, so this PR might slightly regress performance of playerglobals code in some areas.

(Also, it might break some versioning edge cases? not so sure about that though)

Would it be difficult to pass a fake `SwfMovie` that has the right header but no data to `TranslationUnit::from_abc` in `load_builtin_abc`?

core/src/avm2/script.rs
moulins: No, `Translation::debug_name` needs a reference to borrow from. (and in general, I think more methods should return `&Arc` instead or `Arc`, to avoid extra clones unless the caller actually needs it)

core/src/avm2/function.rs
moulins: I prefer not to put too much logic under `#[cfg(...)]`s, because `cargo check/clippy` will ignore it unless you remember to use the correct CLI incantation. (and indeed, when rebasing the PR I initially forgot to update this call-site because `cargo check` didn't complain...)

moulins:
Closing in favor of #24541 (which also covers AVM1).

PR #24536, opened 2026-08-25
avm2: Optimize calls to call handlers to use CallNative

Lord-McSweeney:
## Description

If we know that a `CallProperty` op is going to invoke a call handler, we can optimize the op to a `CallNative` to directly invoke the call handler rather than going through a full property resolution and `Value::call`.

This massively speeds up stuff like `String(x)` or `Boolean(x)` where the compiler decided to use a `findpropstrict` + `callproperty` rather than a `coerce_s`/`coerce_b`, though I've never seen any cases where such code is hot.

## Testing

We don't have test coverage for AVM2 optimizations.

## Checklist


Lord-McSweeney:
Turns out this doesn't really work out for several reasons

PR #24541, opened 2026-08-25
core: Playerglobal loading performance improvements

moulins:
## Description

Alternative to #24535. List of changes:

- When reading SWF extended headers, skip over irrelevant tags instead of parsing them.
- Allow `SwfMovie` to reference static data without copying.
- Use `LazyLock`s to share playerglobal movies between all Ruffle instances.

## Testing

No functional changes.

## Checklist


evilpie:
Did you measure the performance win?

kjarosh:
To measure perf win it's basically enough to compare the test suite total time, as a lot of time there is spent on loading playerglobal.

moulins:
Added a new commit that shares the parsed `AbcFile` between all playerglobal instances (by storing it in an `Arc` instead of a an `Rc`).

> To measure perf win it's basically enough to compare the test suite total time, as a lot of time there is spent on loading playerglobal.

I didn't see any changes in running times of `cargo test` locally, so I'm guessing what you're calling "playerglobal loading" is more about registering all the AVM2 `Class` objects etc, than loading/decoding the ABC data. Anyways, this PR is more aimed towards reducing memory usage than towards reducing runtime (even though the memory savings aren't *that* big, I guess...)

core/src/avm2/script.rs
```diff
@@ -144,14 +143,14 @@ impl<'gc> TranslationUnit<'gc> {
     }
 
     /// Retrieve the underlying `AbcFile` for this translation unit.
-    pub fn abc(self) -> Rc<AbcFile> {
-        self.0.abc.clone()
+    pub fn abc(self) -> &'gc Arc<AbcFile> {
```
Lord-McSweeney, line 146: Can this just return a `&AbcFile`?

Lord-McSweeney (commented):
Does the first commit really measurably improve perf anywhere? i.e. do we have any tags that are expensive to parse, so we should skip trying to parse them?

core/src/avm2/globals.rs
```diff
@@ -816,7 +817,7 @@ pub fn init_native_system_classes(activation: &mut Activation<'_, '_>) {
         ]
     );
 }
-static PLAYERGLOBAL_ABC: LazyLock<SwfSlice> = LazyLock::new(|| {
+static PLAYERGLOBAL_ABC: LazyLock<(Arc<SwfMovie>, Arc<AbcFile>)> = LazyLock::new(|| {
```
Lord-McSweeney, line 823: We eventually want to make `SwfMovie` have a `'gc` lifetime (`Gc<SwfMovie>` instead of `Arc<SwfMovie>`), so it can be collected properly, right? Would that work with this?

core/src/avm2/script.rs
moulins: Most callers don't need an owned `Arc`, so returning a reference skips some ref-count manipulations.

Edit: Sorry i misunderstood the question; it can't return `&AbcFile` either because of `Method::new`, which is the only call-site that wants to clone the Arc

core/src/avm2/globals.rs
moulins: Even with `SwfSlice<'gc>`, I think we'll still want an `Arc` around the buffer to be able to share SWF data with e.g. the audio backend, which lives on another thread and can't accept GC'd data.

So there'll probably be a way of making this work.

moulins:
> Does the first commit really measurably improve perf anywhere? i.e. do we have any tags that are expensive to parse, so we should skip trying to parse them?

Mainly any tag containing `ShapeRecord`s, I'd say (so `DefineShape`, `Text`, etc).

core/common/src/tag_utils.rs
```diff
@@ -60,14 +61,14 @@ pub struct SwfMovie {
...
     pub fn empty(swf_version: u8, loader_url: Option<String>) -> Self {
-        let url = "file:///".to_string();
+        let url = Cow::Borrowed("file:///");
         let header = HeaderExt::default_with_swf_version(swf_version);
 
-        // TODO What sandbox type should we use here?
-        let sandbox_type = SandboxType::infer(url.as_str(), &header);
         Self {
+            // TODO What sandbox type should we use here?
+            sandbox_type: SandboxType::infer(&url, &header),
```
Lord-McSweeney: nit: Can you keep this field at the bottom of the list in all constructions of the struct?

core/common/src/tag_utils.rs
```diff
@@ -17,10 +18,10 @@ pub struct SwfMovie {
     header: HeaderExt,
 
     /// Uncompressed SWF data.
-    data: Vec<u8>,
+    data: Cow<'static, [u8]>,
 
     /// The URL the SWF was downloaded from.
-    url: String,
+    url: Cow<'static, str>,
```
Lord-McSweeney: Is it really important to avoid allocating a `String` for this for `SwfMovie::empty`, `SwfMovie::fake_with_compressed_len`, etc? I don't think the perf gain is worth the extra complexity.

swf/src/read.rs
```diff
@@ -400,21 +410,24 @@ impl<'a> Reader<'a> {
...
 
-    fn read_tag_with_code(&mut self, tag_code: TagCode, length: usize) -> Result<Tag<'a>> {
-        let mut tag_reader = Reader::new(self.read_slice(length)?, self.version);
+    fn read_raw_tag(&mut self) -> Result<(u16, Reader<'a>)> {
+        let (tag_code, length) = self.read_tag_code_and_length()?;
+        let tag_reader = Reader::new(self.read_slice(length)?, self.version);
+        Ok((tag_code, tag_reader))
+    }
+
+    fn read_tag_with_code(tag_code: TagCode, mut tag_reader: Self) -> Result<Tag<'a>> {
```
kjarosh, line 469: Why doesn't this method use self?

moulins: Mainly to avoid changing `tag_reader` to `self` in all the subsequent match arms; if you prefer I can add a `let mut tag_reader = self` line instead.

kjarosh: Yeah, I think it would be better, currently the signature is a bit wonky

PR #24554, opened 2026-08-27
tests: Add test for caching callee object of AVM2 instance init method

Lord-McSweeney:
## Description

Add coverage that checks that the callee object of a class's instance init method is cached across calls. This test currently fails in AVM2 because
- We don't cache the callee object of a class's instance init method
- Our `callstatic` op implementation is totally broken and panics in this test

## Testing

Yes

## Checklist


tests/tests/swfs/avm2/init_callee_cached/test.toml
```diff
@@ -0,0 +1,2 @@
+num_frames = 1
+ignore = true # Currently panics in AVM2 code
```
kjarosh, line 2: Can we assert on the panic?

Lord-McSweeney: It's a panic from within AVM2 code, so it hits the issue I mentioned in https://discord.com/channels/610531541889581066/614855396468719619/1534664790083895538 :(

kjarosh: Ah, dang :/

kjarosh (approved):

PR #24254, opened 2026-07-20
avm2: Implement UVT-mapped bitmap triangles across render backends and move triangles outside tassellation

wsxarcher:
## Description

This PR makes AVM2 `Graphics.drawTriangles` and `GraphicsTrianglePath` first-class indexed triangle primitives. Vertices, indices, UV/UVT data, and culling are validated before rendering, while bounds, hit testing, active fill state, and Flash’s independent-triangle winding behavior are preserved.

Triangles now bypass Lyon tessellation and are uploaded directly across Canvas, WebGL, and WGPU. WebGL and WGPU use dedicated UVT vertex layouts and perspective-correct homogeneous interpolation without increasing ordinary shape vertex sizes; Canvas computes a bitmap transform per triangle. Unit and visual coverage includes UV/UVT conversion, invalid data, winding, drawing state, hit testing, and bitmap mapping.

Fixes #17639

## Checklist

- [x] I, a human, have self-reviewed this PR and fully understand the changes within.
- [x] I have [made or updated tests](https://github.com/ruffle-rs/ruffle/blob/master/CONTRIBUTING.md#test-guidelines) where possible.
- [x] All of my commits are properly scoped, compile successfully, and pass all tests.
- [x] This PR does not make sense to split up into smaller PRs.
- [x] An LLM was involved in the authoring of this code.

PR #24555, opened 2026-08-27
wgpu: Pass UV(T) to shaders, rather than always calculating via a uniform

Dinnerbone:
## Description

This should have much less overhead and save on some resource usage, but it also unlocks being able to draw from arbitrary regions of a texture (for the atlas work).

This may simplify #24254 a bunch?

## Testing

All existing tests pass, and I've ran my usual swfs to verify they look good. I'd appreciate some extra testing though.

## Checklist


kjarosh (approved):
Not an expert, but didn't see anything suspicious or weird, LGTM

PR #24557, opened 2026-08-28
avm2: Have `MovieClip.addFrameScript()` throw error 2001 for an odd number of args

ChrisCPI:
## Description

This error is thrown if there are 0 arguments provided, or if the number of arguments is odd.

## Testing

Test included.

## Checklist


tests/tests/swfs/avm2/movieclip_addframescript/Test.as
```diff
@@ -5,11 +5,28 @@
 	
 	public class Test extends MovieClip {
 		
-		
 		public function Test() {
 			addFrameScript(1, undefined);
 			addFrameScript(0, frame1, 1, frame2a, 1, frame2b, 2, frame3, 2, null, 3, frame4, 4, frame5);
 			addFrameScript(3, "primitive");
+
+			try {
```
kjarosh: Can you add a separate test for it instead of modifying this one? It could use asc then

core/src/avm2/error.rs
```diff
@@ -943,6 +943,20 @@ pub fn make_error_1508<'gc>(activation: &mut Activation<'_, 'gc>, param_name: &s
     ))
 }
 
+#[inline(never)]
+#[cold]
+pub fn make_error_2001<'gc>(
+    activation: &mut Activation<'_, 'gc>,
+    actual: u32,
```
Lord-McSweeney: If `expected` is always `actual + 1`, this function can just take `actual`.

Also, it can be passed in as a `usize`, right?

ChrisCPI: I made it this way because I wasn't sure what other functions may use this error for, but yeah, that can work for now.

Lord-McSweeney (approved):

kjarosh (approved):
Thank you!

PR #24564, opened 2026-08-29
chore: Revert lazy blends, again

### LINKED ISSUE #24549: Regression in intro logo of Sniper Team
### Describe the bug

There's been a recent regression in how the Sniper Team logo renders. Previously, the glint would render only on the graphic (or the text?), but now it renders behind everything.

Image

This was caused by #24445- pinging @atilapan as the author of the PR.

### Expected behavior

Image

### Content Location

[sniperteam-repro.swf.zip](https://github.com/user-attachments/files/31487776/sniperteam-repro.swf.zip)

Dinnerbone:
## Description

This reverts #24445 to fix #24549. The revert is partly manual because this code has changed since that PR landed. It does not revert the test added in that PR.

It seems to be masks across blends, and I can't reason enough about this code anymore to know how to fix this. Let's revert it and I intend to clean everything up here with fire and an excessive amount of energy drinks.

## Testing

A new regression test has been added.

## Checklist


tests/tests/swfs/visual/blend_across_masks_issue_24549/test.toml
```diff
@@ -0,0 +1,8 @@
+num_ticks = 1
+
+[image_comparisons.output]
+tolerance = 0
+max_outliers = 2407
+
+[player_options]
+with_renderer = { optional = false, quality = "low" }
```
kjarosh, line 8: Is low quality a good idea? The image has non-trivial edges and it looks like we could assert better with AA enabled and non-zero tolerance.

Dinnerbone: I typically do low quality because AA is very unpredictable across different machines. In this case going with high doesn't seem to drop the outliers any significant degree, even with a high threshold

Threshold 50: 1310 outliers
Threshold 100: 808 outliers

I'll play around with the test and see if I can get it more accurate in Ruffle, maybe it's the pixel snapping

kjarosh: If the shape is unrelated to the regression, you can just use simple squares on a small image (e.g. 50x50). This will make the test faster and easier to assert on.

Dinnerbone: 0 tolerance test :tada:

kjarosh: Very cool! 😌 

The new image looks like a modern art, cubistic version of the previous image, lol

kjarosh (approved):

PR #24565, opened 2026-08-29
render: Preserve strokes under collapsed transforms

### LINKED ISSUE #21776: All assets are not displayed correctly [n_v14.swf]
### Describe the bug

The N game (version 1.4) is not displayed properly on Ruffle. A lot of its assets are not fully displayed (player, bombs, door key, golds). Other than that, the game works properly.

### Expected behavior

Expected (.exe):
Image

Actual (ruffle):
Image

### Content Location

N 1.4 can be extracted from the .exe found [here](https://www.thewayoftheninja.org/n_history.html) or [here](https://archive.org/details/n_v1pc), or obtained directly in .swf form from [NReality](https://n.infunity.com/)

### Affected platform

Desktop app

### Operating system

Arch Linux

### Browser

_No response_

### Additional information

_No response_

Perdu:
## Description

We've been working on [a Tool-Assisted Speedrun for the whole N game](https://github.com/Perdu/nv14_TAS/tree/main). As we're getting closer to the end of this tremendous task, it appears more and more urgent to get [the display bug](#21776) in Ruffle fixed, so we can submit our TAS to TASvideos.org. This bug was partially fixed in #23011, but the display of the core of the body of the player itself is still incorrect. As this has been stalled for years (#21776 is not the first report), I took the liberty at firing ChatGPT at the problem. I don't know what the repo's policy around AI-generated patches is, but I hope it will at least hint maintainers at a solution.

The proposed solution does fix the visual problem. See the following screenshots (watch the ninja in the bottom left corner):
- original Flash:
Capture d’écran du 2026-08-29 15-51-47
- Current ruffle version: 
Capture d’écran du 2026-08-29 15-49-21
- with the patch: 
Capture d’écran du 2026-08-29 15-48-46

Moreover, as a test case, ChatGPT also gave me [an .swf file](https://github.com/Perdu/nv14_TAS/blob/main/volume/n_v14_graphical_issue_fix.swf) that does solve the display problem in the current (unpatched) version of ruffle. This at least constitutes a strong workaround for us. Original swf can be found [here](https://github.com/Perdu/nv14_TAS/blob/main/external/n_v14.swf).

Disclaimer: I am not knowledgable enough about flash and ruffle internals to know about any possible side-effect or performance issue brought by this patch. All I know is that I test, it does build and fix the issue.

With all of that said, here's the ChatGPT-redacted PR:
## Summary

Fix the remaining player rendering issue from #21776 without modifying
display-object transforms in core.

#23011 made very thin strokes scale-aware during tessellation and fixed most
of the missing outlines in N 1.4, but the ninja body remained invisible.

The ninja uses a one-dimensional stroked `DefineShape` under placement
matrices whose perpendicular basis is exactly zero, for example:

```text
[ 0.1171875  0 ]
[ 0           0 ]
```

WGPU/WebGL expand strokes into triangles in local space and apply the final
display-object matrix afterwards. A zero perpendicular basis therefore
collapses even a minimum-width tessellated stroke back to zero-area geometry.

Canvas does not have this problem because it transforms the path before
applying stroke width.

## Fix

`ShapeTessellator` now attaches a small `OneDimensionalHairline` metadata
record to its output mesh when all of the following are true:

- the shape's edge bounds are exactly one-dimensional;
- the shape contains only strokes; and
- every stroke is using the minimum-width rule from the tessellator.

The metadata records the line axis, an anchor point, and the scale at which
the mesh was tessellated.

WGPU and WebGL preserve this metadata on their backend-specific shape meshes.
At draw time, if the final display transform completely collapses the axis
perpendicular to that line, the backend reconstructs only that missing basis
as a perpendicular vector whose magnitude matches the tessellation scale.
Translation is compensated so the original transformed centerline remains
unchanged.

This keeps the compatibility handling in the tessellation/rendering layer
where the stroke triangles are actually collapsed. Core and Canvas are
unchanged.

## Why the reconstructed magnitude is the tessellation scale

The minimum local stroke width used by `ShapeTessellator` is `1 / scale`.
Giving the restored perpendicular basis magnitude `scale` maps that minimum
width back to exactly one device pixel.

## Scope / performance

The metadata is computed once when a shape is tessellated. The common WGPU
and WebGL draw paths pay only an `Option` check. Matrix adjustment runs only
for tagged one-dimensional minimum-width stroke meshes whose perpendicular
basis is exactly zero.

No allocations or additional draw calls are introduced.

Filled shapes, two-dimensional shapes, thicker strokes, Canvas, and ordinary
(non-collapsed) transforms are unchanged.

## Verification

The root cause was verified in both directions with N 1.4:

1. With unmodified Ruffle, changing only the ninja placement matrices from
   `scaleY = 0` to a nonzero value makes the ninja body render.
2. With the original, unmodified SWF, preserving a perpendicular basis for
   the tessellated stroke makes the ninja body render.

## Tests

Unit tests cover:

- the horizontal collapsed-axis case used by N;
- an offset/rotated horizontal line;
- the vertical equivalent;
- preservation of the original transformed centerline; and
- rejection of thick strokes and two-dimensional geometry.

Manual regression: original N 1.4 `n_v14.swf`.

Closes #21776.

Related: #23011.

## Checklist

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -4301,9 +4301,9 @@
             "license": "MIT"
         },
         "node_modules/@wdio/mocha-framework/node_modules/brace-expansion": {
-            "version": "2.1.2",
-            "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-2.1.2.tgz",
-            "integrity": "sha512-w5JZcKgdhDOgOwm8H+KgbosopHMuGcl6qbulwjtz3SM7I7P3yW1eAjzMPLrIE+NQ9vjgANKHWeMHnrT0OXW1oA==",
+            "version": "2.1.4",
+            "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-2.1.4.tgz",
+            "integrity": "sha512-hGfVzPxthbf3+2yjg/RBs60cB0FhqBS/zvdV/4wn4/BmN0bNMMHPc4V/BbFieqf1TKAGGAHnY4eSjajCl0f2Xg==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
@@ -5199,9 +5199,9 @@
             "license": "MIT"
         },
         "node_modules/archiver-utils/node_modules/brace-expansion": {
-            "version": "2.1.2",
-            "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-2.1.2.tgz",
-            "integrity": "sha512-w5JZcKgdhDOgOwm8H+KgbosopHMuGcl6qbulwjtz3SM7I7P3yW1eAjzMPLrIE+NQ9vjgANKHWeMHnrT0OXW1oA==",
+            "version": "2.1.4",
+            "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-2.1.4.tgz",
+            "integrity": "sha512-hGfVzPxthbf3+2yjg/RBs60cB0FhqBS/zvdV/4wn4/BmN0bNMMHPc4V/BbFieqf1TKAGGAHnY4eSjajCl0f2Xg==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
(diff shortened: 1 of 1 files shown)


core/src/display_object/graphic.rs
```diff
@@ -256,7 +304,18 @@ impl<'gc> TDisplayObject<'gc> for Graphic<'gc> {
         if let Some(drawing) = self.0.drawing.get() {
             drawing.borrow().render(context);
         } else if let Some(base_handle) = self.0.shared.get().render_handle.clone() {
-            let transform = context.transform_stack.transform();
+            let mut transform = context.transform_stack.transform();
+            let shared = self.0.shared.get();
+            // Canvas transforms the path before stroking it, so singular transforms
+            // do not collapse stroke thickness there. Tessellating backends expand
+            // strokes first, so a zero transform axis can collapse the stroke mesh.
```
kjarosh: What is a zero transform axis?

core/src/display_object/graphic.rs
```diff
@@ -256,7 +304,18 @@ impl<'gc> TDisplayObject<'gc> for Graphic<'gc> {
...
+            if context.renderer.name() != "canvas"
+                && transform.perspective_projection.is_none()
+                && shared.shape.styles.fill_styles.is_empty()
```
kjarosh: It there some logic behind this? This looks like a bandaid for your specific SWF.

core/src/display_object/graphic.rs
kjarosh: If tessellating before transforms is the root cause, it looks like the proper solution would be to transform before tessellate? We already have a tessellation cache, the question is whether this would be acceptable performance-wise. My guess is that by selecting proper thresholds it would be.

n0samu:
How exactly did ChatGPT generate a modified game SWF file? Do you have source code for the game or did it somehow modify the SWF file itself?

Perdu:
Apologies, I originally pushed a fix that was just a test and not supposed to be a final PR.
I pushed a more complete fix, along with unit tests, and edited the description.
I hope this is clearer than the original PR!

Perdu:
> How exactly did ChatGPT generate a modified game SWF file? Do you have source code for the game or did it somehow modify the SWF file itself?

It's not difficult to modify an SWF file with ffdec. But ChatGPT can do much more than that all by itself as of now, it's scary. We've had it rewrite the whole game engine in C, write a bruteforcer/optimiser, and it can use it all by itself to improve the TAS.

